PR #39135, opened 2015-04-28
juju 1.23

dunn:
(the Go build process remains a continual source of delight)

PR #39223, opened 2015-04-30
juju 1.23.2

sinzui:


dunn:
I've already got #39135 open, but we've run into trouble with the test-bot. I'll bump it to 1.23.2, though.

sinzui:
Think this alternate means of building juju tries to do to much, and it can break users who upgrade before the version is in the "released" streams that the Juju Release team seeds in the clouds.

But I can explain how anyone can make a tarfile just like I do for Juju QA and Release team. We use this script to make a tarfile that only contains the required deps and removes non-free or undistributable data that some of the dependencies contain. The tarfile contains everything needed to build without a network.
    http://bazaar.launchpad.net/~sinzui/juju-release-tools/trunk/view/head:/make-release-tarball.bash

The only binaries that OSX needs (and can run) are juju and juju-metadata. The backup and restore binaries are vestigial. The native "backups" feature in the juju binary is superior. The jujud only runs on Linux (and in a limited way on Windows)

Juju QA tests the OS X binaries made from the tarfile we build. When we first announce a release, the agents required to bootstrap and upgrade in the clouds are only available in the "proposed" streams. An unsuspecting user who get the client will be surprised that "juju bootstrap" no longer works. The user needs to reconfigure all the envs with "agent-stream: proposed". I create the branch for homebrew at this time, but I do not create the pull request until the agents in all the clouds are coped from "proposed" streams to "released". I make the Pull request about few minutes later.

We could have a devel stanza in the formula form the devel and proposed juju. I think we want to show a message though that explains the extra step need to use the non-default streams.

PR #39229, opened 2015-04-30
git 2.4.0

ianferguson:


Library/Formula/git.rb
```diff
@@ -1,7 +1,7 @@
 class Git < Formula
   homepage "http://git-scm.com"
```
DomT4: Can you poke this to ` https:// `? Thanks!

ianferguson: @DomT4 seems like that should be done as a separate commit/change/PR from just bumping the versions doesn't it? 

That being said, I'm happy to tack on the change to this

ianferguson: @DomT4 done.

DomT4: Na. Opportunistic changes are fine in PRs, especially plaintext => SSL/TLS link upgrades.

ianferguson: Cool, I'll keep that in mind for the future. I also just reread contributing.md and as a result squashed that change into one commit. let me know if there's anything else you spot.

DomT4: Nice! Looks great :tada:.

joshfriend:
:+1:

PR #39233, opened 2015-04-30
AWS cli 1.7.25

darend:
Updating AWS CLI to 1.7.25

### DIFF
--- a/Library/Formula/awscli.rb
+++ b/Library/Formula/awscli.rb
@@ -1,7 +1,7 @@
 class Awscli < Formula
   homepage "https://aws.amazon.com/cli/"
-  url "https://pypi.python.org/packages/source/a/awscli/awscli-1.7.18.tar.gz"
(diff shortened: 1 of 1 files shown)


ms-ati:
Hey just so I can do this correctly next time rather than asking: how do you get the full set of python dependencies to update here? Do you manually look at all the setup.py files starting at the top, or some easier way?

darend:
Truthfully, I installed via pip and looked at the set of dependencies.

ms-ati:
Ah! Well, I supposed I could do that. But might defeat the point ;) Thanks for doing it for the rest of us.

tdsmith:
[homebrew-pypi-poet](https://github.com/tdsmith/homebrew-pypi-poet) can automate the "see what pip does" process a little.

tdsmith:

ms-ati:
@tdsmith that's fascinating. I wonder if it would make sense to wrap homebrew-pypi-poet in a virtualized environment and a script, so that it became truly a one-click affair to update homebrew? Or even automated?

PR #39234, opened 2015-04-30
awscli 1.7.25

ms-ati:
Fixes #39233


jacknagel:
Dupe of #39233

ms-ati:
I believe you should take the union of our two sets up dependency upgrades -- it looks like (in #39233 commit af3b245) he missed the following: bcdoc 0.14.0

/cc @darend

ms-ati:
(nevermind, deleted last comment as I don't know how I missed it, but bcdoc 0.14.0 was definitely in there)

PR #39244, opened 2015-05-01
mitie 0.4

dunn:


Library/Formula/mitie.rb
```diff
@@ -14,27 +14,34 @@ class Mitie < Formula
 
   depends_on :python if MacOS.version <= :snow_leopard
 
+  option "with-models", "Install v0.2 models (~415MB)"
```
tdsmith: What are these? (Is this useful without the models?)

Library/Formula/mitie.rb
```diff
@@ -14,27 +14,34 @@ class Mitie < Formula
...
 
   def install
-    (share+"MITIE-models").install resource("models-english")
+    if build.with? "models"
+      (share/"MITIE-models").install resource("models-english")
+    end
+
     inreplace "mitielib/makefile" do |s|
-      s.gsub! /libmitie.so/, "libmitie.dylib"
+      s.gsub!(/libmitie.so/, "libmitie.dylib")
```
tdsmith: fwiw this is fine without parentheses

Library/Formula/mitie.rb
```diff
@@ -14,27 +14,34 @@ class Mitie < Formula
...
+    (lib/"python2.7/site-packages").install "mitielib/mitie.py"
+    (share/"mitie").install "examples", "sample_text.txt",
       "sample_text.reference-output", "sample_text.reference-output-relations"
     bin.install "ner_example", "ner_stream", "relation_extraction_example"
   end
 
   test do
-    system "#{bin}/ner_example", "#{share}/MITIE-models/english/ner_model.dat", "#{share}/mitie/sample_text.txt"
+    system ENV.cc, "-I#{include}", "-L#{lib}", "-lmitie",
+           share/"mitie/examples/C/ner/ner_example.c", "-o", "test"
```
tdsmith: I guess the models need to be installed to invoke this?

Library/Formula/mitie.rb
dunn: I'm not 100 percent sure! But I assumed something of that size shouldn't be downloaded by default.

Library/Formula/mitie.rb
dunn: Oh, emacs flagged it as "ambiguous first argument". ¯\\\_(ツ)_/¯ Should I revert it?

Library/Formula/mitie.rb
dunn: Yeah, the old test required the models.

Library/Formula/mitie.rb
dunn: Their readme just says [you need the models in order to run the provided examples](https://github.com/mit-nlp/MITIE#initial-setup).

dunn: And it installs a C library too, so there are definitely use-cases that don't require the models.

tdsmith: I think the models are useful: they let you do things like http://blog.dlib.net/2014/07/mitie-v02-released-now-includes-python.html?m=1 out of the box. I think making them optional (`option "without-models"`) is fair though.

Library/Formula/mitie.rb
tdsmith: eh, nah, maybe emacs knows better than I do

dunn:
OK, made the models recommended and combined the old and new tests.

Library/Formula/mitie.rb
```diff
@@ -14,27 +14,37 @@ class Mitie < Formula
...
     bin.install "ner_example", "ner_stream", "relation_extraction_example"
   end
 
   test do
-    system "#{bin}/ner_example", "#{share}/MITIE-models/english/ner_model.dat", "#{share}/mitie/sample_text.txt"
+    system ENV.cc, "-I#{include}", "-L#{lib}", "-lmitie",
+           share/"mitie/examples/C/ner/ner_example.c",
+           "-o", testpath/"ner_example"
+    system "./ner_example", share/"MITIE-models/english/ner_model.dat",
+           share/"mitie/sample_text.txt"
```
xu-cheng: Maybe need to check `built.with? "models"` then run the test.

dunn: I can do that, but I've been told previously not to worry about the test passing with non-default options.

tdsmith:
:metal:

PR #38759, opened 2015-04-17
Use https URLs

pdalpra:
This PR fixes some URLs to use HTTPS (more specifically, Formulae from W to Z :) )

PR #39245, opened 2015-05-01
xml-security-c: fix system openssl linkage

DomT4:
Fixes system OpenSSL linkage, bumps revision, adds test, rolls SSL/TLS, removes redundant configure option.

Re: https://github.com/Homebrew/homebrew/pull/38759#issuecomment-94099350


tdsmith:
Thanks!

PR #39252, opened 2015-05-01
utils: play nicely with rbenv

DomT4:
See discussion thread below for detail.

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -126,8 +126,8 @@ module Homebrew
     ENV["PATH"] = "#{Gem.user_dir}/bin:#{ENV["PATH"]}"
 
     unless quiet_system "gem", "list", "--installed", gem
-      safe_system "gem", "install", "--no-ri", "--no-rdoc",
-                                    "--user-install", gem
+      safe_system "/usr/bin/gem", "install", "--no-ri", "--no-rdoc",
+                                  "--user-install", gem
     end
 
     unless which executable


DomT4:
Currently, if you have `rbenv` installed and activated, you get this:

```
Dominyks-MacBook-Pro ~> brew man                                                                  (master) [/usr/local]
WARNING:  You don't have /Users/Dominyk/.gem/ruby/2.2.0/bin in your PATH,
	  gem executables will not run.
Building native extensions.  This could take a while...
Successfully installed hpricot-0.8.6
Successfully installed mustache-1.0.1
Successfully installed ronn-0.7.3
3 gems installed
Error: The 'ronn' gem is installed but couldn't find 'ronn' in the PATH:
/Users/Dominyk/.gem/ruby/2.0.0/bin:/Users/Dominyk/.nvm/versions/node/v0.12.2/bin:/usr/local/var/rbenv/shims:/usr/local/opt/curl/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/opt/go/libexec/bin:/usr/local/jsl:/usr/texbin:/Users/Dominyk/go/bin
```

This seems to happen because `brew irb` considers `Gem.user_dir` to be `/Users/Dominyk/.gem/ruby/2.0.0` but `rbenv` and similar are ahead of `/usr/bin` in the `$PATH` always, so simply doing ` gem install ` anything leads to `rbenv`'s Gem and Ruby being used.

If you just do the same `irb` check outside of Homebrew with `rbenv` installed or similar, you get `/Users/Dominyk/.gem/ruby/2.2.0`. (Or whatever version of Ruby you currently have activated).

So the gems are being installed in `/Users/Dominyk/.gem/ruby/2.2.0/bin` but Homebrew is prepending `/Users/Dominyk/.gem/ruby/2.0.0/bin` to the `$PATH`.

I was going to do a more complicated dance on this, but simple is better, and less likely to make Jack frown. In this case, if whichever Ruby/Gem is first in the `$PATH` doesn't have the required gems installed, Homebrew now defaults to using the system Ruby to install them, and consequently correctly prepends the `$PATH` as expected.

End result:

```
Fetching: hpricot-0.8.6.gem (100%)
Building native extensions.  This could take a while...
Successfully installed hpricot-0.8.6
Fetching: rdiscount-2.1.8.gem (100%)
Building native extensions.  This could take a while...
Successfully installed rdiscount-2.1.8
Fetching: mustache-1.0.1.gem (100%)
Successfully installed mustache-1.0.1
Fetching: ronn-0.7.3.gem (100%)
Successfully installed ronn-0.7.3
4 gems installed
Writing manpages to /usr/local/share/man/man1
```

CC @Homebrew/owners

Library/Homebrew/utils.rb
```diff
@@ -126,8 +126,8 @@ def self.install_gem_setup_path! gem, executable=gem
     ENV["PATH"] = "#{Gem.user_dir}/bin:#{ENV["PATH"]}"
 
     unless quiet_system "gem", "list", "--installed", gem
```
jacknagel: this needs to match the below change.

jacknagel:
Does this mean I won't be able to use a non-system ruby/rubygems for this stuff (which is what I currently do)?

Library/Homebrew/utils.rb
DomT4: It deliberately doesn't. My thinking was if someone wants to use their custom Ruby/Gem version, presuming it's the first Ruby/Gem in the `$PATH` and they have the required gem installed already, it'll work fine.

The only issue is when there's a custom Ruby/Gem prepending the `$PATH` but the required gems aren't installed, it won't install them in the correct location and can cause breakage. Essentially `/usr/bin/gem` is the fallback to the user's custom Ruby/Gem not having the required gems installed already.

DomT4:
> Does this mean I won't be able to use a non-system ruby/rubygems for this stuff (which is what I currently do)?

Hopefully my comment on your other comment helps clarify this? Essentially, if that Ruby/Gem is in front of `/usr/bin` in your `$PATH` and you have the required gems installed already, Brew will use that. Falling back on the system is just the backup.

I can probably throw in a more detailed workaround if you'd prefer, but I didn't know if you'd be happier with it being simpler and more minimal.

xu-cheng:
Sorry, I fail to understand the problem. It works fine for me (with/without rbenv activated).

DomT4:
I reproduced it on a clean VM earlier :confused:. I'll try and see if it can reproduce in Travis or something, or give you more specific steps to reproduce.

DomT4:
Bingo. Here you go: https://travis-ci.org/DomT4/Zen/builds/60991366#L1355

Mirrored here in case Travis ditches the logs: https://gist.github.com/DomT4/9d8e4418ae1584978a14

Library/Homebrew/utils.rb
jacknagel: Nak on this approach, sorry.

If you ask the `gem` on `$PATH` if a gem is installed, but then use a possibly totally different `gem` to install it, the user will be left wondering where the hell it was installed. Not something we want to do.

DomT4: Fair enough, makes sense. Interested in a more complicated but user-expectation-friendly workaround?

xu-cheng:
Does this still happen if you installed `rbenv-gem-rehash`?

DomT4:
Yup, exact same failure. Tested `zsh`, our `bash` and the system `bash` as well.

DomT4:
Will close here. We haven't had any complaints about it aside my PR, so hopefully anyone else encountering the issue is working out how to resolve it themselves. Will review other options if we suddenly encounter a rush of issues on regenerating manpages/doing strict audits.

PR #39255, opened 2015-05-01
freetds: upgrade to 0.91.112; style changes for `brew audit --strict`; other configuration changes

L2G:
I broke this up into two commits, but can merge them into one once everyone is satisfied with the changes.

The first one is a straightforward upgrade to a new version with style changes to pass `brew audit --strict`. Only one potentially controversial item: I had to specify the `version`. Even though the tarball being downloaded rolls up all of the latest patches, they don't name it with the version number used in the code (but it is evident within the tarball's file structure and in the source code).

The second commit is mainly to migrate all of the `--enable-*` options to `--with-*` options. Two important things of note in this one:
* I turned the dependency on OpenSSL into a `:recommended` dependency (thus, `--without-openssl` is a new option). To respect all open-source licenses, I think this should really be an `:optional` dependency, but I made it `:recommended` to minimize the potential for surprise. (I say this because the FreeTDS docs make the assertion that linking with OpenSSL introduces a license conflict, yet it _is_ still an option.)
* I added a clause that checks for `HOMEBREW_DEVELOPER` and enables the configuration option `--enable-option-checking=fatal` accordingly. This gave me peace of mind while I was migrating the formula options, so I decided to leave it in; if anyone hates it, I can take it out. :smile_cat:

### DIFF
--- a/Library/Formula/freetds.rb
+++ b/Library/Formula/freetds.rb
@@ -1,10 +1,7 @@
-require "formula"
-
 class Freetds < Formula
   homepage "http://www.freetds.org/"
-  url "http://mirrors.ibiblio.org/freetds/stable/freetds-0.91.tar.gz"
-  sha1 "3ab06c8e208e82197dc25d09ae353d9f3be7db52"
-  revision 2
+  url "ftp://ftp.freetds.org/pub/freetds/stable/freetds-0.91.112.tar.gz"
+  sha256 "be4f04ee57328c32e7e7cd7e2e1483e535071cec6101e46b9dd15b857c5078ed"
 
   bottle do
     sha1 "7f98f1313e932185398c72237b7ac0d408cea986" => :mavericks
@@ -13,7 +10,7 @@ class Freetds < Formula
   end
 
   head do
-    url "https://git.gitorious.org/freetds/freetds.git"
+    url "https://github.com/FreeTDS/freetds.git"
 
     depends_on "autoconf" => :build
     depends_on "automake" => :build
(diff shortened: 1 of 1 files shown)


Library/Formula/freetds.rb
```diff
@@ -21,43 +19,52 @@ class Freetds < Formula
...
   option :universal
-  option "enable-msdblib", "Enable Microsoft behavior in the DB-Library API where it diverges from Sybase's"
-  option "enable-sybase-compat", "Enable close compatibility with Sybase's ABI, at the expense of other features"
-  option "enable-odbc-wide", "Enable odbc wide, prevent unicode - MemoryError's"
-  option "enable-krb", "Enable Kerberos support"
+  option "with-msdblib",       "Enable Microsoft behavior in the DB-Library API where it diverges from Sybase's"
+  option "with-sybase-compat", "Enable close compatibility with Sybase's ABI, at the expense of other features"
+  option "with-odbc-wide",     "Enable odbc wide, prevent unicode - MemoryError's"
+  option "with-krb5",          "Enable Kerberos support"
+  option "without-openssl",    "Build without OpenSSL support (default is to use brewed OpenSSL)"
```
tdsmith: Don't align these, please.

Library/Formula/freetds.rb
```diff
@@ -21,43 +19,52 @@ class Freetds < Formula
...
       --with-tdsver=7.1
       --mandir=#{man}
     ]
 
-    if build.with? "unixodbc"
-      args << "--with-unixodbc=#{Formula["unixodbc"].prefix}"
-    end
+    # Allow Homebrew developer mode to engage turning any warnings about
+    # configuration options into fatal errors
+    args << "--enable-option-checking=fatal" if ARGV.homebrew_developer?
```
tdsmith: This might be useful in superenv but I don't think it should live in each formula.

Library/Formula/freetds.rb
```diff
@@ -21,43 +19,52 @@ class Freetds < Formula
...
 
-    if build.include? "enable-msdblib"
-      args << "--enable-msdblib"
+    # OpenSSL has historically been an absolute requirement of this formula.
+    # However, the FreeTDS docs assert that linking with OpenSSL introduces a
+    # licensing conflict.  We're splitting the difference by making it a
+    # recommended option.
```
tdsmith: This comment can be removed. Do we have to do anything to affirmatively disable openssl otherwise?

Library/Formula/freetds.rb
```diff
@@ -21,43 +19,52 @@ class Freetds < Formula
...
+  option "with-msdblib",       "Enable Microsoft behavior in the DB-Library API where it diverges from Sybase's"
+  option "with-sybase-compat", "Enable close compatibility with Sybase's ABI, at the expense of other features"
+  option "with-odbc-wide",     "Enable odbc wide, prevent unicode - MemoryError's"
+  option "with-krb5",          "Enable Kerberos support"
+  option "without-openssl",    "Build without OpenSSL support (default is to use brewed OpenSSL)"
+
+  deprecated_option "enable-msdblib"       => "with-msdblib"
+  deprecated_option "enable-sybase-compat" => "with-sybase-compat"
+  deprecated_option "enable-odbc-wide"     => "with-odbc-wide"
+  deprecated_option "enable-krb"           => "with-krb5"
```
tdsmith: (or these)

Library/Formula/freetds.rb
```diff
@@ -1,10 +1,8 @@
-require "formula"
-
 class Freetds < Formula
   homepage "http://www.freetds.org/"
-  url "http://mirrors.ibiblio.org/freetds/stable/freetds-0.91.tar.gz"
-  sha1 "3ab06c8e208e82197dc25d09ae353d9f3be7db52"
-  revision 2
+  url "ftp://ftp.freetds.org/pub/freetds/stable/freetds-patched.tar.gz"
```
tdsmith: I do have a strong preference for a tarball with a version number in the filename, so that it won't break our hash checking when it's inevitably updated.

Library/Formula/freetds.rb
L2G: I'm not sure I understand your question, but I can strike the comment.

Library/Formula/freetds.rb
L2G: In this case, I don't think there's an alternative short of mirroring our own copy of the tarball and changing its name.

Library/Formula/freetds.rb
tdsmith: Like, do we need `--disable-openssl` to keep it from building against it? (I don't know what the default is.)

L2G: The historic default of the _formula_ is to have an absolute dependency on (Homebrew's) OpenSSL. But FreeTDS itself defaults to _not_ linking against OpenSSL at all unless it's requested.

I was afraid if I just yanked it, it might break something for someone already using Homebrew's freetds.

Library/Formula/freetds.rb
L2G: Sweet, I found out where they were keeping their properly-named tarballs.

ftp://ftp.freetds.org/pub/freetds/stable/freetds-0.91.112.tar.gz

L2G:
The latest Jenkins failure is due to this failing:

    brew test --verbose pgloader

L2G:
I'm not sure how the test against pgloader is being initiated, but it makes sense since it has freetds as a dependency.

I think I found the smoking gun in the [Mavericks build log](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/25280/version=mavericks/console):

```
LOG:  Unix-domain socket path "/private/var/folders/_z/11_kwl954xvch_n92gnzdjzw0000gp/T/pgloader20150502-57592-155rz91/socket/.s.PGSQL.5432" is too long (maximum 103 bytes)
WARNING:  could not create Unix-domain socket in directory "/private/var/folders/_z/11_kwl954xvch_n92gnzdjzw0000gp/T/pgloader20150502-57592-155rz91/socket"
FATAL:  could not create any Unix-domain sockets
Error: pgloader: failed
```

tdsmith:
That's a frustrating error condition but I guess UNIX sockets are special? Some background at https://bugs.mysql.com/bug.php?id=42512

Anyway, yeah, it looks like the pgloader test is bad.

L2G:
@jacknagel Thanks. :+1:

tdsmith:
Thanks for your work here!

PR #39256, opened 2015-05-01
couchdb: bump revision for icu4c

DomT4:
Looks like there's runtime issues again. CouchDB needs bumping with each icu4c update that changes icu4c dylib versioning.

See: https://github.com/Homebrew/homebrew/issues/19641#issuecomment-98247080

### DIFF
--- a/Library/Formula/couchdb.rb
+++ b/Library/Formula/couchdb.rb
@@ -1,10 +1,8 @@
-require "formula"
-
 class Couchdb < Formula
-  homepage "http://couchdb.apache.org/"
-  url "http://www.apache.org/dyn/closer.cgi?path=/couchdb/source/1.6.1/apache-couchdb-1.6.1.tar.gz"
-  sha1 "6275f3818579d7b307052e9735c42a8a64313229"
-  revision 1
+  homepage "https://couchdb.apache.org/"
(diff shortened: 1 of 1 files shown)


jacknagel:
Hmm, any particular reason this was missed?

DomT4:
The current test doesn't pick it up. I tried writing about 10 *(literally ~10)* tests for ` couchdb ` itself but it's a nightmare, including borrowing the one below from another formula.

You get permission and user errors everywhere, even when you customise the ` pid ` file and then you recursively copy the ` etc/couchdb ` directory into the testpath and tell `couchdb` to use that instead of the usual `sysconfdir` to read from.

```ruby
    pid = fork { exec "#{bin}/couchdb -p #{testpath}/couchdb.pid" }
    sleep 3
    Process.kill("SIGINT", pid)
    Process.wait(pid)
```

Was the most promising result, but even that = nope.

natevw:
@DomT4 Have you tried using the `-a` option to push a testpath file override to the configuration stack? (Perhaps with a `-n` to clear the stack first, too…). I'm not sure if the pid file is even written without background mode, so I wonder if the errors you are hitting are instead related to CouchDB trying to update its configuration with some defaults and/or admin user accounts, and needs a user-writable file to do so?

PR #39261, opened 2015-05-01
jvgrep v4.1 (new formula)

kozo2:
`jvgrep` is grep for Japanese Vimmer.
You can search files written in multiple Japanese encodings.


tdsmith:
Thanks for the pull request! 🎉 Homebrew wouldn't work without contributions like this and we're grateful for your support.

kozo2:
thank you

PR #39266, opened 2015-05-01
magit 1.4.1

dunn:


Library/Formula/magit.rb
```diff
@@ -1,11 +1,24 @@
...
+  def satisfied?
+    major_version = `emacs --batch --eval "(princ emacs-major-version)"`
+    minor_version = `emacs --batch --eval "(princ emacs-minor-version)"`
+    (major_version + "." + minor_version).to_f >= 23.2
+  end
+
+  def message
+    "Emacs 23.2 or later is required for magit."
+  end
+end
```
xu-cheng: I wonder should we create DSL like `depends_on :emacs => "24.4"`?

dunn: Currently only two other formulae have a custom emacs requirement—[`cask.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/cask.rb) and [`mu.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mu.rb)—but there are others that could probably use it, like [`proof-general.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/proof-general.rb#L28).

xu-cheng:
Thanks. Also, we prefer the test block is after the caveats.

PR #39175, opened 2015-04-29
vault 0.1.0 (new formula)

sidick:
Add formula for vault 0.1.0

PR #37485, opened 2015-03-07
Check revisions for Git tags

mikemcquaid:
It's always bothered me a bit that we use checksums so heavily but when we check out a Git tag we're effectively just relying on a string match. Instead, let's nudge people into adding a `:revision` when they specify a `:tag` and then verify the two match. This should give us the same level of protection for Git tags as we do for their tarball equivalents.

CC @Homebrew/owners @DomT4

PR #39267, opened 2015-05-01
Vault 0.1.1 (new formula)

egrim:
Add formula for vault 0.1.0

This extends @sidick's work in pull request #39175 by building `vault` from source.

### DIFF
--- a/Library/Formula/vault.rb
+++ b/Library/Formula/vault.rb
@@ -1,0 +1,68 @@
+require "language/go"
+
+class Vault < Formula
+  homepage "https://vaultproject.io/"
+  url "https://github.com/hashicorp/vault.git",
+      :tag => "v0.1.1",
+      :revision => "074dcaadac2e0a46ae28f6d491885165b7431aae"
+
+  head "https://github.com/hashicorp/vault.git"
+
+  depends_on "go" => :build
+
+  go_resource "github.com/tools/godep" do
+    url "https://github.com/tools/godep",
+        :revision => "58d90f262c13357d3203e67a33c6f7a9382f9223", :using => :git
+  end
+
+  # godep's dependencies
+  go_resource "github.com/kr/fs" do
+    url "https://github.com/kr/fs",
+        :revision => "2788f0dbd16903de03cb8186e5c7d97b69ad387b", :using => :git
+  end
+
+  go_resource "golang.org/x/tools" do
+    url "https://go.googlesource.com/tools",
+        :revision => "96f6cfbb921ad6d191c67d09a6d4c4fd056403ae", :using => :git
+  end
+
+  go_resource "github.com/mitchellh/gox" do
+    url "https://github.com/mitchellh/gox",
+        :revision => "e8e6fd4fe12510cc46893dff18c5188a6a6dc549", :using => :git
+  end
+
+  # gox dependency
+  go_resource "github.com/mitchellh/iochan" do
+    url "https://github.com/mitchellh/iochan",
+        :revision => "b584a329b193e206025682ae6c10cdbe03b0cd77", :using => :git
+  end
+
+  def install
+    gopath = buildpath/"gopath"
+    mkdir_p gopath
+    ENV["GOPATH"] = gopath
+    ENV.prepend_create_path "PATH", gopath/"bin"
+
+    Language::Go.stage_deps resources, gopath/"src"
+
+    cd gopath/"src/github.com/tools/godep" do
+      system "go", "install"
+    end
+
+    cd gopath/"src/github.com/mitchellh/gox" do
+      system "go", "install"
+    end
+
+    mkdir_p gopath/"src/github.com/hashicorp"
+    system "git", "clone", buildpath, gopath/"src/github.com/hashicorp/vault"
+
+    cd gopath/"src/github.com/hashicorp/vault" do
+      system "make", "dev"
+      bin.install "bin/vault"
+    end
+  end
+
+  test do
+    system bin/"vault", "--version"
+  end
+end


Library/Formula/vault.rb
```diff
@@ -0,0 +1,35 @@
+require "language/go"
+
+class Vault < Formula
+  homepage "https://vaultproject.io/"
+  version "0.1.0"
+  url "https://github.com/hashicorp/vault.git"
```
DomT4: You need to actually use a tagged release, not just the master branch. The ` head ` points at this, the ` url ` pretty much never should. Use ` https://github.com/hashicorp/vault/archive/v0.1.0.tar.gz ` instead.

Library/Formula/vault.rb
```diff
@@ -0,0 +1,35 @@
...
+  depends_on "go" => :build
+
+  def install
+    gopath = buildpath/"../gopath"
+
+    mkdir_p gopath
+    ENV["GOPATH"] = gopath
+    ENV.prepend_create_path "PATH", gopath/"bin"
+
+    system "go", "get", "github.com/tools/godep"
```
DomT4: Use `go_resource`s instead of `go get`. See example [here](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/gauge.rb).

Library/Formula/vault.rb
```diff
@@ -0,0 +1,35 @@
...
+    system "go", "get", "github.com/mitchellh/gox"
+
+    mkdir_p gopath/"src/github.com/hashicorp"
+    cp_r buildpath, gopath/"src/github.com/hashicorp/vault"
+
+    cd gopath/"src/github.com/hashicorp/vault" do
+      system "make", "dev"
+      bin.install "bin/vault"
+    end
+
```
DomT4: No need for empty line here.

Library/Formula/vault.rb
```diff
@@ -0,0 +1,35 @@
...
+  homepage "https://vaultproject.io/"
+  version "0.1.0"
+  url "https://github.com/hashicorp/vault.git"
+
+  head "https://github.com/hashicorp/vault.git"
+
+  depends_on "go" => :build
+
+  def install
+    gopath = buildpath/"../gopath"
```
DomT4: `gopath = buildpath/"src/github.com/vault`

Library/Formula/vault.rb
mistydemeo: Instead of running `go get`, can you stage these as `go_resource`s? If you take a look at deisctl you can find a formula that does this, including godep.

Library/Formula/vault.rb
egrim: Unfortunately the upstream requires the `.git` directory to be intact (which the tarball doesn't have) so I'm using git instead.  I've now added the `:tag`.

DomT4: Fair enough. `:tag` is a valid use case here :+1:.

Library/Formula/vault.rb
```diff
@@ -0,0 +1,35 @@
+require "language/go"
+
+class Vault < Formula
+  homepage "https://vaultproject.io/"
+  url "https://github.com/hashicorp/vault.git",
+      :tag => "v0.1.0",
+      :revision => "c8d2dfa67886a80cb4347c25f332a74f91313d23"
```
xu-cheng: Using tartball instead.

Library/Formula/vault.rb
```diff
@@ -0,0 +1,35 @@
...
+    cp_r buildpath, gopath/"src/github.com/hashicorp/vault"
+
+    cd gopath/"src/github.com/hashicorp/vault" do
+      system "make", "dev"
+      bin.install "bin/vault"
+    end
+  end
+
+  test do
+    system bin/"vault", "--version"
```
xu-cheng: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/vault.rb
DomT4: Asked about this earlier. Response:

> Unfortunately the upstream requires the .git directory to be intact (which the tarball doesn't have) so I'm using git instead. I've now added the :tag.

xu-cheng: Fear enough.

xu-cheng:
Also can you squash the commits? Thanks.

Library/Formula/vault.rb
egrim: Gave it a shot originally but ran into problems.  I'll try harder and report back.

Library/Formula/vault.rb
egrim: Unfortunately, the build scripts for this tool require that the git repo be placed within a subdirectory of the `gopath`.  Other `go` scripts use a symlink to achieve this, but this tool's build scripts are more persnickety and I have to place the `gopath` outside of the `buildpath` and then copy it into place.  For the copy to work, the `gopath` can't be within the `buildpath` because it'd be a recursive copy.

My solution here works, but I'm happy to change it if there's a better alternative.

egrim:
All PR comments addressed (except that I don't see a good way to make the test step more substantial).

Library/Formula/vault.rb
```diff
@@ -0,0 +1,68 @@
...
+    cd gopath/"src/github.com/tools/godep" do
+      system "go", "install"
+    end
+
+    cd gopath/"src/github.com/mitchellh/gox" do
+      system "go", "install"
+    end
+
+    mkdir_p gopath/"src/github.com/hashicorp"
+    system "git", "clone", buildpath, gopath/"src/github.com/hashicorp/vault"
```
mistydemeo: How about, instead of `git clone`ing, doing this at the very beginning of the `install` method:

```ruby
    contents = Dir["*"]
    gopath = buildpath/"gopath"
    gopath.mkpath

    target = gopath/"src/github.com/hashicorp/vault"
    target.mkpath
    mv contents, target
```

I tested and this appears to work, though if there's an issue I've missed I'm glad to go with your version!

Library/Formula/vault.rb
```diff
@@ -0,0 +1,68 @@
+require "language/go"
+
+class Vault < Formula
+  homepage "https://vaultproject.io/"
+  url "https://github.com/hashicorp/vault.git",
+      :tag => "v0.1.1",
+      :revision => "074dcaadac2e0a46ae28f6d491885165b7431aae"
```
mistydemeo: No need for both `:tag` and `:revision` - `:tag` is enough to tell Homebrew to fetch the repository at the commit referenced by that tag.

xu-cheng: Actually, `:revision` is required now as a security check.

DomT4: #37485 for reference.

mistydemeo: Ah thanks, I missed that change!

Library/Formula/vault.rb
egrim: Great suggestion, @mistydemeo!  I really like heading down that path instead, but sadly your specific formula snippet causes vault's build scripts to think the code has been changed from the tag (i.e. `vault --version` reports there were `CHANGES`) because the move doesn't include top-level "dot" files.  

To fix this we'd need to glob everything in the contents directory, however my ruby is weak and thus my attempts at making the necessary change (e.g. `Dir.glob("*", File::FNM_DOTMATCH)`) end up grabbing the `.` and `..` directories - which messes everything up!  Can you suggest an idiomatic ruby way to pick up everything in the buildpath including dotfiles but excluding the `.` and `..` directories?

mistydemeo: `Dir["{*},.git"]` should do the trick - I did [something similar](https://github.com/Homebrew/homebrew/pull/39702/files#diff-b8b72c1277d585e0105dfc20d3990fe3R207) in the packer formula.

DomT4:
Please follow up on Misty and Xu's comments here, Thanks!

DomT4:
Closing due to no response.

PR #39271, opened 2015-05-02
embulk 0.6.5

guoxiao:


Library/Formula/embulk.rb
```diff
@@ -0,0 +1,19 @@
+class Embulk < Formula
+  homepage "http://www.embulk.org/"
+  url "https://bintray.com/artifact/download/embulk/maven/embulk-0.6.5.jar"
+  sha256 "fdb45887060777f3b216793c256959fa6018bab516dbcebfb95648a1a78839c8"
+
+  depends_on :java
+
+  def install
+    bin.install "embulk-0.6.5.jar" => "embulk"
+    chmod 0755, bin/"embulk"
```
guoxiao: Hmm... Seems doesn't work....

tdsmith: The cleaner is resetting the permissions because embulk doesn't have a shebang. embulk should have a shebang (`#!/bin/bash`) so that it works with non-bash-compatible shells.

An alternative is:
* install embulk to libexec/bin and chmod it like you're doing here
* add skip_clean "libexec" to the Formula class body
* call `bin.write_exec_script libexec/"bin/embulk"` to write a bash script that execs embulk

guoxiao: Thanks for your help. I've updated the formula.

tdsmith:
Thanks for your PR! :tada:

Issue #39217, opened 2015-04-30
Please do not version bump the nimrod formula.


philip-wernersbach:
I'd like to ask the Homebrew maintainers to not version bump the [nimrod formula](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/nimrod.rb) for now. The upstream Windows installer now packages [nimble](https://github.com/nim-lang/nimble), so we have to figure out how to package that with our formula. There is at least one [upstream issue](https://github.com/nim-lang/nimble/issues/80) that may be blocking this.

@Araq and @dom96: How should we proceed?

philip-wernersbach:
#39272 updated the nimrod formula to 0.11.2, but the issue of not packaging nimble with nim still stands. Need guidance from upstream.

Also, I've updated the title to reflect the current status of the nimrod formula, but apparently I can't reopen the issue.

philip-wernersbach:
Ping @Araq and @dom96

dom96:
I'm not familiar with how Homebrew operates but I think that Nimble should be a separate formula.

philip-wernersbach:
Thanks for the quick response @dom96. How can we accomplish that, with regards to nim-lang/nimble#80?

dom96:
I suppose I will need to implement https://github.com/nim-lang/nimble/issues/80.

Araq:
I have no idea what it takes to package Nimble on "Homebrew" and I have no opinion either.

PR #39272, opened 2015-05-02
update nim / nimrod to 0.11.0; fix install; add test

dcreemer:
Update to upstream nim source 0.11.0 from http://nim-lang.org/download.html
switch from .zip file to .tar.xz as the latter is about 10 times smaller
fix the install phase to install the nim binary
fix the tests to actually test correct output

### DIFF
--- a/Library/Formula/nimrod.rb
+++ b/Library/Formula/nimrod.rb
@@ -15,7 +15,6 @@ class Nimrod < Formula
     system "/bin/sh", "build.sh"
     system "/bin/sh", "install.sh", prefix
 
-    (prefix/"nim").install "compiler"
     bin.install_symlink prefix/"nim/bin/nim"
     bin.install_symlink prefix/"nim/bin/nim" => "nimrod"
   end
--- a/Library/Formula/nimrod.rb
+++ b/Library/Formula/nimrod.rb
@@ -1,7 +1,7 @@
 class Nimrod < Formula
   homepage "http://nim-lang.org/"
-  url "http://nim-lang.org/download/nim-0.11.0.tar.xz"
-  sha1 "5cbad487c94d0dd4b8cbd52486dd43b1ccf671fb"
+  url "http://nim-lang.org/download/nim-0.11.2.tar.xz"
+  sha1 "2693022e35e31196d14ec2d1fbf11a90abac34cf"
   head "https://github.com/Araq/Nim.git", :branch => "devel"
 
   bottle do
@@ -15,7 +15,6 @@ class Nimrod < Formula
     system "/bin/sh", "build.sh"
     system "/bin/sh", "install.sh", prefix
 
-    (prefix/"nim/bin").install "bin/nim"
     (prefix/"nim").install "compiler"
     bin.install_symlink prefix/"nim/bin/nim"
     bin.install_symlink prefix/"nim/bin/nim" => "nimrod"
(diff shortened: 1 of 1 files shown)


tdsmith:
ping @philip-wernersbach per #39217 (issue titled "Please do not version bump the nimrod formula.")

dcreemer:
I will of course defer to the pkg maintainer here -- my goal is just to run the [latest release](https://github.com/Araq/Nim/tree/v0.11.0)  of https://github.com/Araq/Nim on Mac OS X. Does this pull-request make the situation of not having a strategy on https://github.com/nim-lang/nimble sorted out any worse than it is with the [previous release](https://github.com/Araq/Nim/tree/v0.10.2)?

tdsmith:
I'm tempted to pull this until we get a PR that adds nimble, yes.

philip-wernersbach:
Some comments:

1. Remove Lines 7-12, brewbot will have to generate new bottles.
2. If Line 18 is needed to install the `bin` directory, then that's a pretty severe upstream regression, as `install.sh` is supposed to do it. Keep the line and file an upstream issue for it.
3. Remove Line 19, the bug it works around has been fixed upstream.
4. Looks good otherwise.
5. Love the way you've implemented an actual test.

As for #39217, I haven't heard anything from upstream yet. I do have a solution for the upstream blocker in the form of [`nimble-wrapper`](https://github.com/philip-wernersbach/nimble-wrapper), but it was [rejected last time](https://github.com/Homebrew/homebrew/pull/34693) because it would be a Homebrew only fix, which I kind of agree with.

@dcreemer and @tdsmith I'm starting to agree with you two about pulling this for now and maybe adding nimble later when the upstream stuff is fixed. It will create issues for upstream support because they'll expect people to have `nimble` installed by default, but I'm starting to agree that it's better to provide the newest version.

tdsmith:
Please don't modify the bottle block in a pull request, thanks.

Tim

dcreemer:
Thanks for the quick response. I've filed a bug -- Araq/Nim#2647 for the install.sh script. I don't think I can remove line 19 (yet) as this formula refers directly to the 0.11.0 release (and it's tarball) which still have the bug. I'll be happy to update when 0.11.1 (or whatever) is ready.

philip-wernersbach:
Line 19 works around Araq/Nim#1706.

def-:
This has been fixed in 0.11.2.

dcreemer:
I've updated to 0.11.2 as you can see in the recent commit -- this fixes the 'nim' binary install problem. However the 'compiler' directory is still not installing as part of the default script, so I have left that line in place. (Hopefully this is temporary).

dcreemer:
Ah -- I more carefully read the Araq/Nim#1706 bug and see that the compiler directory no longer needs to be installed. Removed.

dcreemer:
ping -- can this be merged?

Library/Formula/nimrod.rb
```diff
@@ -15,16 +15,15 @@ def install
...
   end
 
   test do
     (testpath/"hello.nim").write <<-EOS.undent
-      echo("Hi!")
+      echo("hello")
     EOS
-    system "#{bin}/nim", "compile", "--run", "hello.nim"
+    assert_equal "hello\n", `#{bin}/nim compile --verbosity:0 --run #{testpath}/hello.nim`
+    assert_equal 0, $?.exitstatus
```
xu-cheng: You can replace these two line with:
```ruby
assert_equal "hello", shell_output("#{bin}/nim compile --verbosity:0 --run #{testpath}/hello.nim").chomp
```

philip-wernersbach:
Yes, merge this.

Also, no word from upstream yet.

dcreemer:
thanks @xu-cheng -- I have incorporated your suggestion.

PR #39273, opened 2015-05-02
spawn-fcgi add disable-ipv6 option and test

guoxiao:


tdsmith:
Thanks for the pull request! 🎉 Homebrew wouldn't work without contributions like this and we're grateful for your support.

PR #39275, opened 2015-05-02
mongoose 5.6

guoxiao:


Library/Formula/mongoose.rb
```diff
@@ -1,16 +1,16 @@
...
-    end
-    include.install 'mongoose.h'
-    prefix.install 'examples', 'docs'
+    system ENV.cc, "-dynamiclib", "mongoose.c", "-o", "libmongoose.dylib"
+    include.install "mongoose.h"
+    lib.install "libmongoose.dylib"
+    share.install "examples", "docs", "jni"
+  end
+
+  test do
```
asparagui: :frog: :+1:

Library/Formula/mongoose.rb
```diff
@@ -1,16 +1,16 @@
...
+    system ENV.cc, "#{share}/examples/hello_world/hello_world.c", "-o", "hello_world", "-lmongoose"
```
xu-cheng: It would be better to run the test problem after this.

guoxiao: It's a http server daemon, so it's no easy to test the functionality.
But I think it should be good if it can build without errors.

xu-cheng: Is it possible to do something like this:
```ruby
pid = fork { system "./hello_world" }
sleep 1
system "curl", "localhost", ...
Process.kill 9, pid
Process.wait pid
```

guoxiao: Let me have a try...

guoxiao: I tried your suggestion. It worked, but there are several problems:
* The firewall popup a message box to ask for permission
* The program will listen on port 8080, which may be conflict with other problem

So I'm wondering whether or not it's appropriate to add such a test?

guoxiao: In my opinion, the test block in formula should be a test of formula, rather than a test of software itself. 
If we want to have a complete test, run `make test` after building may be a better choice.

xu-cheng: * There is no firewall enabled by default on OS X. 
* You can get a free port number by listening to port 0.
* The formula test is to help identify breakage caused by dependencies update. 

Nevertheless, I'm happy to merge this as it is.

guoxiao: In one of my previous PR (#38473) I met the same problem when I wanted to test a daemon. Maybe we need a general solution in class Formula.

tdsmith:
Some build options in case any of these are interesting (openssl maybe?): https://github.com/cesanta/mongoose/blob/master/docs/Embed.md#compilation-flags

xu-cheng:

PR #39285, opened 2015-05-02
syntaxerl 0.8.0 (new formula)

waisbrot:
[Syntaxerl](https://github.com/ten0s/syntaxerl) is a utility for syntax-checking Erlang files (source, header-files, and various config files). It's meant to be used in Emacs' Flymake, Vim's syntastic, and similar tools.

### DIFF
--- a/Library/Formula/syntaxerl.rb
+++ b/Library/Formula/syntaxerl.rb
@@ -1,0 +1,21 @@
+class Syntaxerl < Formula
+  homepage "https://github.com/ten0s/syntaxerl"
+  desc "Syntax-checker for Erlang code and config files"
+  url "https://github.com/ten0s/syntaxerl/archive/0.8.0.tar.gz"
+  sha256 "4c5a6390a7d28869bf44cc831ae59cb69480a8481853de8dd59672b7cf5bff7b"
+
+  depends_on "erlang"
+
+  def install
+    system "make"
+    bin.install "syntaxerl"
+  end
+
+  test do
+    (testpath/"app.config").write "[{app,[{arg1,1},{arg2,2}]}]."
+    assert_equal "", shell_output("#{bin}/syntaxerl #{testpath}/app.config")
+
+    (testpath/"invalid.config").write "]["
+    assert_match /invalid.config:1: syntax error before: '\]'/, shell_output("#{bin}/syntaxerl #{testpath}/invalid.config")
+  end
+end


Library/Formula/syntaxerl.rb
```diff
@@ -0,0 +1,18 @@
...
+  homepage "https://github.com/ten0s/syntaxerl"
+  url "https://github.com/ten0s/syntaxerl/archive/0.7.0.tar.gz"
+  sha256 "0011fa629e268200f1767b260be984ac6ce83200ac9b401df3d6565cf75ccb24"
+
+  depends_on "erlang"
+
+  def install
+    system "make"
+    mkdir bin
+    cp "syntaxerl", "#{bin}/syntaxerl"
```
tdsmith: replace these two lines with `bin.install "syntaxerl"`

waisbrot:
Corrected. Thanks, @tdsmith .

Library/Formula/syntaxerl.rb
```diff
@@ -0,0 +1,17 @@
...
+  depends_on "erlang"
+
+  def install
+    system "make"
+    bin.install "syntaxerl"
+  end
+
+  test do
+    # syntax-check its own source code
+    system "syntaxerl", "src/syntaxerl.erl"
```
DomT4: If ` src/syntaxerl.erl ` exists in the `buildpath` the test won't have access to it FYI.

waisbrot:
My test was no good because I need to look at the output rather than the exit code. Now I'm using popen3 to check the output, and I'm writing the test files on the fly.

Library/Formula/syntaxerl.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def install
+    system "make"
+    bin.install "syntaxerl"
+  end
+
+  test do
+    require "open3"
+    (testpath/"app.config").write "[{app,[{arg1,1},{arg2,2}]}]."
+    Open3.popen3("#{bin}/syntaxerl", testpath/"app.config") do |stdin, stdout, _|
```
bfontaine: You could use `shell_output("#{bin}/syntaxerl #{testpath}/app.config")` to get the output of the command and then test it, e.g.: `assert_equal "", shell_output("#{bin}/syntaxerl #{testpath}/app.config")`.

waisbrot:
Sounds good, @bfontaine . I've updated to use `assert_equal` and `assert_match`

Library/Formula/syntaxerl.rb
```diff
@@ -0,0 +1,20 @@
+class Syntaxerl < Formula
+  homepage "https://github.com/ten0s/syntaxerl"
```
mikemcquaid: Sorry, we've added a new description (`desc`) attribute. Can you add one? Thanks!

Library/Formula/syntaxerl.rb
```diff
@@ -0,0 +1,20 @@
...
+    system "make"
+    bin.install "syntaxerl"
+  end
+
+  test do
+    (testpath/"app.config").write "[{app,[{arg1,1},{arg2,2}]}]."
+    assert_equal "", shell_output("#{bin}/syntaxerl #{testpath}/app.config")
+
+    (testpath/"invalid.config").write "]["
+    assert_match /invalid.config:1: syntax error before: ']'/, shell_output("#{bin}/syntaxerl #{testpath}/invalid.config")
```
mikemcquaid: The `]` needs escaped or you should use a string instead of a regex here.

waisbrot:
Added a `desc` and fixed the regex, @mikemcquaid

mikemcquaid:

PR #39290, opened 2015-05-02
shell.fm: migrating to homebrew/boneyard

bfontaine:
Note that `shell.fm` no longer works because [last.fm stopped their radio streaming](http://nex.scrapping.cc/shell-fm/index.html).


n054:
If the pushes came through an old device no longer in service would it help test to turn old on? Try to bounce of tower (self) build inside.

n054:
Just a note, was using at a point an a FM Transmitter w/ 4(four) channels. This allowed m.device() to internet in a "cold" spot. (I have one on ().

Library/Formula/shell.fm.rb
```diff
@@ -1,20 +1,23 @@
...
-  depends_on 'pkg-config' => :build
-  depends_on 'libao'
-  depends_on 'mad'
+  depends_on "pkg-config" => :build
+  depends_on "libao"
+  depends_on "mad"
 
   def install
+    # 'make install' fails if this exists
+    rm "INSTALL"
```
jacknagel: Upstream could address this by adding `.PHONY: install` to the Makefile.

jacknagel:
Should we boneyard this, then?

bfontaine:
@jacknagel Done: Homebrew/homebrew-boneyard#46

tdsmith:
Thanks Baptiste!

PR #39294, opened 2015-05-02
binwalk: update formula, resources and HEAD build

makigumo:
Hi!

Originally I just wanted to fix the HEAD build, but that lead to updating the whole formula and resources.
So some things might be not up to standard :/

### DIFF
--- a/Library/Formula/binwalk.rb
+++ b/Library/Formula/binwalk.rb
@@ -1,11 +1,9 @@
-require 'formula'
-
 class Binwalk < Formula
-  homepage 'http://binwalk.org/'
-  revision 1
+  homepage "http://binwalk.org/"
+  revision 2
   stable do
     url "https://github.com/devttys0/binwalk/archive/v2.0.1.tar.gz"
-    sha1 "b0ec783cbf72db358c5118a3052fc1ccf446d8f3"
+    sha256 "90ee8426d71e91b62dfe4a1446c457bc7835b475b28717859e275a0494403959"
   end
 
   bottle do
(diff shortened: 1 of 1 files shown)


Library/Formula/binwalk.rb
```diff
@@ -15,57 +13,61 @@ class Binwalk < Formula
...
-    patch do
-      url "https://gist.github.com/balr0g/a917b31318016c63a72d/raw/d434c3ceaa57438d39344e760990e9268893ce5f/binwalk-head-201140712.patch"
-      sha1 "2e000654968d2c8ad17b7fc46924300cf01bbfed"
-    end
+    url "https://github.com/devttys0/binwalk.git"
   end
 
-  option 'with-matplotlib', 'Check for presence of matplotlib, which is required for entropy graphing support'
+  option "with-matplotlib", "Check for presence of matplotlib, which is required for entropy graphing support"
+  option "with-capstone", "Enable disasm options via capstone" if build.head?
```
tdsmith: I might just say "(affects --HEAD only)" instead of adding `if build.head?` since it won't show up in `brew info` otherwise.

Library/Formula/binwalk.rb
```diff
@@ -15,57 +13,61 @@ class Binwalk < Formula
...
-    system "make install"
+    if build.head?
+      ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+      system "python", *Language::Python.setup_install_args(libexec)
+      bin.install Dir["#{libexec}/bin/*"]
+      bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+    else
+      system "./configure", "--prefix=#{prefix}", "--disable-bundles"
+      system "make", "install"
+    end
```
tdsmith: Is setup.py missing from the stable release? It'd be nice if we could always use the same build system.

Library/Formula/binwalk.rb
jacknagel: For correctness, this needs to be in a head block if it only applies to head. Using `build.head?` in the DSL is deprecated.

makigumo: Ok, will do. FYI, if I put the option in a ```head do … end``` the option never shows, but does using ```if build.head?``` in ```brew info binwalk --HEAD```

Library/Formula/binwalk.rb
makigumo: There is a setup.py but there are some C files to be compiled. They were, including configure/Makefile, completely removed in HEAD. So unfortunately two separate build paths are needed.

Library/Formula/binwalk.rb
```diff
@@ -1,11 +1,9 @@
-require 'formula'
-
 class Binwalk < Formula
-  homepage 'http://binwalk.org/'
-  revision 1
+  homepage "http://binwalk.org/"
+  revision 2
```
DomT4: Why the revision bump, out of interest?

makigumo: I thought it appropriate because of the updated resources.

DomT4: It usually isn't enough, but I'll defer to whatever the maintainers think best here. Thanks for bumping everything to SHA256 by the way! :tada:.

Library/Formula/binwalk.rb
tdsmith: Thanks for explaining!

Library/Formula/binwalk.rb
tdsmith: Let's leave the revision at 1.

makigumo: Ok, reverted.

xu-cheng:

PR #39295, opened 2015-05-02
nsq: new homepage, add HEAD

dunn:
Locally the test passed but sometimes several of the processes weren't actually killed. Will that cause problems for the bots, and how can it be avoided?

### DIFF
--- a/Library/Formula/nsq.rb
+++ b/Library/Formula/nsq.rb
@@ -1,10 +1,11 @@
-require "formula"
 require "language/go"
 
 class Nsq < Formula
-  homepage "https://bitly.github.io/nsq"
+  homepage "http://nsq.io"
   url "https://github.com/bitly/nsq/archive/v0.3.5.tar.gz"
-  sha1 "a876334ad959685fa109f67b4d07231ae84bcf43"
+  sha256 "4120ad24e3700be1e65549b9a55eab5d4e744cd114d9b39779a47b6dedda0b35"
+
+  head "https://github.com/bitly/nsq.git"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Library/Formula/nsq.rb
```diff
@@ -72,6 +73,36 @@ def install
...
+    end
+    sleep 2
+    to_file = fork do
+      system bin/"nsq_to_file", "--topic=test", "--output-dir=#{testpath}",
+             "--lookupd-http-address=127.0.0.1:4161"
+    end
+    sleep 2
+    curl = fork do
+      system "curl", "-d", "'hello world 2'", "http://127.0.0.1:4151/put?topic=test"
+    end
```
xu-cheng: Is it necessary for all of these into separate forks?

Library/Formula/nsq.rb
```diff
@@ -72,6 +73,36 @@ def install
...
+    sleep 5
+    Process.kill(9, lookupd)
+    Process.kill(9, d)
+    Process.kill(9, admin)
+    Process.kill(9, to_file)
+    Process.kill(9, curl)
```
xu-cheng: For forks it would be better to do in this way:
```
begin
 pid = fork { ... }
ensure
 Process.kill 9, pid
 Process.wait pid
end
```

Library/Formula/nsq.rb
dunn: `curl` doesn't need to be in a fork, but the rest [apparently do](http://nsq.io/overview/quick_start.html)—I tried a single fork and couldn't get the test to complete.

Library/Formula/nsq.rb
```diff
@@ -72,6 +73,39 @@ def install
...
+      Process.kill(9, to_file)
+      Process.wait lookupd
+      Process.wait d
+      Process.wait admin
+      Process.wait to_file
+
+      assert_match "test",
+                   Pathname(`ls *.dat`.chomp).read
+      assert_match version.to_s,
+                   Pathname(`ls *.dat`.chomp).read
```
xu-cheng: These shouldn't be inside ensure block right. Also let do things like this `File.read(Dir["*.data"].first)`

dunn: Right, sorry about that.

xu-cheng:
Thanks @dunn

PR #39299, opened 2015-05-03
nghttp2 0.7.13

dunn:


Library/Formula/nghttp2.rb
```diff
@@ -33,28 +33,58 @@ class Nghttp2 < Formula
   depends_on "spdylay" => :recommended
 
   resource "sphinx" do
-    url "https://pypi.python.org/packages/source/S/Sphinx/Sphinx-1.2.3.tar.gz"
-    sha1 "3a11f130c63b057532ca37fe49c8967d0cbae1d5"
+    url "https://pypi.python.org/packages/source/S/Sphinx/Sphinx-1.3.1.tar.gz"
+    sha256 "1a6e5130c2b42d2de301693c299f78cc4bd3501e78b610c08e45efc70e2b5114"
+  end
+
+  resource "sphinx_rtd_theme" do
```
DomT4: Given this is a theme, does this need to stick around after compile?

Library/Formula/nghttp2.rb
```diff
@@ -33,28 +33,58 @@ class Nghttp2 < Formula
...
+    sha256 "2e24ac5d004db5714976a04ac0e80c6df6e47e98c354cb2c0d82f8879d4f8fdb"
   end
 
   resource "markupsafe" do
     url "https://pypi.python.org/packages/source/M/MarkupSafe/MarkupSafe-0.23.tar.gz"
-    sha1 "cd5c22acf6dd69046d6cb6a3920d84ea66bdf62a"
+    sha256 "a4ec1aff59b95a14b45eb2e23761a0179e98319da5a7eb76b56ea8cdc7b871c3"
+  end
+
+  resource "alabaster" do
```
DomT4: Are all these for `sphinx` documentation?

dunn: I think so, though tbh I wasn't looking too closely after the third or fourth build failure and just gave it whatever it asked for.

DomT4: Weird. A couple are themes, and themes might need to persist after installation so Sphinx can still use them. Not 100% though.

tdsmith:
Thanks! The themes shouldn't need to persist.

PR #37280, opened 2015-02-28
poppler 0.34.0

bfontaine:

PR #39300, opened 2015-05-03
poppler 0.32.0 and more

L2G:
These changes upgrade [Poppler](http://poppler.freedesktop.org/) to 0.32.0, add a `--HEAD` build, and clean up the formula for `brew audit --strict`.

I realize they are in separate commits, but I am happy to meld them into one when the time comes.

### DIFF
--- a/Library/Formula/poppler.rb
+++ b/Library/Formula/poppler.rb
@@ -1,7 +1,7 @@
 class Poppler < Formula
   homepage "http://poppler.freedesktop.org"
-  url "http://poppler.freedesktop.org/poppler-0.29.0.tar.xz"
-  sha256 "3c1e16dc8d5d846d085baf65844ed3e05cbde5d44f6682f9f1e2b120d66c62c4"
+  url "http://poppler.freedesktop.org/poppler-0.32.0.tar.xz"
+  sha256 "4963e31ba5e17530a87b16588e22928bc044e8d28d02303dded981bff6725b98"
 
   bottle do
     sha1 "b83e3b7fe032d69343367ceb481a0387e447e565" => :yosemite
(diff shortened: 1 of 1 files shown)


Library/Formula/poppler.rb
```diff
@@ -17,26 +17,31 @@ class Poppler < Formula
...
-  depends_on 'freetype'
-  depends_on 'gettext'
-  depends_on 'glib'
-  depends_on 'gobject-introspection'
-  depends_on 'jpeg'
-  depends_on 'libpng'
-  depends_on 'libtiff'
-  depends_on 'openjpeg'
+  if build.head?
+    depends_on :automake => :build
```
DomT4: ` "automake" ` and ` "autoconf" `. The `:` version of those deps are deprecated, I believe.

dunn:
#37280 is already open but it's fallen off the front page because of problems with dependents.

L2G:
Sounds like a case of too many cooks. Er, brewers. :-) I'm going to close this PR and stand out of the way.

Library/Formula/poppler.rb
L2G: So we're back to strings?

L2G: I usually look to `brew audit --strict` to catch things like that...

DomT4: I'm surprised it didn't. I could have sworn there was some mention of ` :autoconf ` and similar being deprecated somewhere. Maybe there's not an audit check, or I'm just wrong - The latter happens a lot. CC @mikemcquaid for detail on whether I'm being stupid or not :smile_cat:

Library/Formula/poppler.rb
```diff
@@ -57,8 +62,17 @@ def install
...
+    else
+      system "./configure", *args
+    end
+
+    system "make", "install"
+    resource("font-data").stage { system "make", "install", "prefix=#{prefix}" }
+  end
+
+  test do
+    system bin/"pdfinfo", HOMEBREW_PREFIX/"Library/Homebrew/test/fixtures/test.pdf"
```
dunn: For future reference you can do `test_fixtures("test.pdf")` instead!

L2G: Oh, that's good to know. I should've thrown that into #39301...

PR #39301, opened 2015-05-03
pdfgrep: HEAD-related changes

L2G:
This is mainly to move the source for the HEAD build from Gitorious.org to GitLab.com.

I also noticed that the HEAD build was broken, so I did my best to fix it; the only sticking point that is beyond the scope of this formula is an upgrade to poppler (see #37280).

### DIFF
--- a/Library/Formula/pdfgrep.rb
+++ b/Library/Formula/pdfgrep.rb
@@ -3,7 +3,12 @@ class Pdfgrep < Formula
   url "https://downloads.sourceforge.net/project/pdfgrep/1.3.1/pdfgrep-1.3.1.tar.gz"
   sha1 "8d15760af0803ccea32760d5f68abe4224169639"
 
-  head "https://gitlab.com/pdfgrep/pdfgrep.git"
+  head do
+    url "https://gitlab.com/pdfgrep/pdfgrep.git"
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
+    depends_on "asciidoc" => :build
+  end
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Library/Formula/pdfgrep.rb
```diff
@@ -15,8 +15,16 @@ class Pdfgrep < Formula
   depends_on "pkg-config" => :build
   depends_on "poppler"
 
+  if build.head?
+    depends_on :automake
+    depends_on :autoconf
+    depends_on "asciidoc" => :build
+  end
```
jacknagel: The symbol for is no longer used; use `depends_on "automake" => :build` and `depends_on "autoconf" => :build` instead.

Also, these should be in a `head` block (with the URL) rather than a conditional.

L2G: Seems like you were just telling me about those symbols a little while ago... :astonished:

Library/Formula/pdfgrep.rb
```diff
@@ -3,7 +3,12 @@ class Pdfgrep < Formula
   url "https://downloads.sourceforge.net/project/pdfgrep/1.3.1/pdfgrep-1.3.1.tar.gz"
   sha1 "8d15760af0803ccea32760d5f68abe4224169639"
 
-  head "https://git.gitorious.org/pdfgrep/pdfgrep.git"
+  head do
+    url "https://gitlab.com/pdfgrep/pdfgrep.git"
+    depends_on "automake"
+    depends_on "autoconf"
```
jacknagel: These need `=> :build`

L2G:
Whoops. So they do.

Library/Formula/pdfgrep.rb
L2G: Whoops. So they do.

xu-cheng:
Thanks @L2G

PR #39304, opened 2015-05-03
audit: Don't complain about untapped conflicts

chdiza:
If a formula conflicts with a formula that lives in a tap, but the user doesn't have that tap tapped, then `audit` will complain that it can't find the conflicting formula.  It shouldn't do that.

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -217,7 +217,9 @@ class FormulaAuditor
       begin
         Formulary.factory(c.name)
       rescue FormulaUnavailableError
-        problem "Can't find conflicting formula #{c.name.inspect}."
+        unless c.name =~ HOMEBREW_TAP_FORMULA_REGEX
+          problem "Can't find conflicting formula #{c.name.inspect}."
+        end
       end
     end
   end


chdiza:
@xu-cheng I don't know much about Homebrew internals, but this seemed to quiet `audit` down.

xu-cheng:
LGTM

Library/Homebrew/cmd/audit.rb
```diff
@@ -217,7 +217,9 @@ def audit_conflicts
       begin
         Formulary.factory(c.name)
       rescue FormulaUnavailableError
-        problem "Can't find conflicting formula #{c.name.inspect}."
+        unless c.name =~ HOMEBREW_TAP_FORMULA_REGEX
+          problem "Can't find conflicting formula #{c.name.inspect}."
+        end
```
xu-cheng: I just find that it would be better to write in this way.
```ruby
begin
  Formulary.factory(c.name)
rescue TapFormulaUnavailableError
  # Don't complain about missing cross-tap conflicts.
rescue FormulaUnavailableError
  problem "Can't find conflicting formula #{c.name.inspect}."
end
```

PR #39305, opened 2015-05-03
sdl_rtf: brew audit --strict

L2G:


Library/Formula/sdl_rtf.rb
```diff
@@ -1,16 +1,19 @@
...
   def install
     system "./configure", "--prefix=#{prefix}"
-    system "make install"
+    system "make", "install"
+  end
+
+  test do
+    # Would need to write a sample C file that can be linked to libSDL_rtf...
+    system "false"
   end
```
jacknagel: If we're not actually adding a test at this point then this should be removed.

tdsmith:
Would you be willing to write a test?

L2G:
With all due respect, no. :smile_cat:

xu-cheng:
Again, we don't merge style only change.

DomT4:
Closing here. Will happily consider a PR rebased against master here and with an added test, Thanks!

PR #39306, opened 2015-05-03
gawk 4.1.2

chdiza:


chdiza:
The "main" url, when changed to https, didn't work any more, so I made the https-ified mirror be the main url.

DomT4:
Sorry if the `audit` is unclear on this. The `mirror` (now the `url`) is the URL that should always be `https` - The `ftpmirror` url should *never* be `https`. As you noticed, wrapping the main URL in SSL/TLS = :skull:.

tdsmith:
@DomT4, what's the right thing to do here?

DomT4:
```ruby
  homepage "https://www.gnu.org/software/gawk/"
  url "http://ftpmirror.gnu.org/gawk/gawk-4.1.2.tar.xz"
  mirror "https://ftp.gnu.org/gnu/gawk/gawk-4.1.2.tar.xz"
  sha256 "ea8b53c5834ee27012fecba7273f97fca7104884975cf81ddae6f85c5581f481"
```

chdiza:
OK, I'm gonna squash.

Someone please fix that audit warning.  That's like the fourth time it's confused the heck out of me.

DomT4:
I guess I could PR something like this:

```diff
diff --git a/Library/Homebrew/cmd/audit.rb b/Library/Homebrew/cmd/audit.rb
index 3ddbdd0..e3caea8 100644
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -332,7 +332,7 @@ class FormulaAuditor

       case p
       when %r[^http://ftp\.gnu\.org/]
-        problem "ftp.gnu.org urls should be https://, not http:// (url is #{p})."
+        problem "ftp.gnu.org mirrors should be https://, not http:// (mirror is #{p})."
       when %r[^http://[^/]*\.apache\.org/]
         problem "Apache urls should be https://, not http (url is #{p})."
       when %r[^http://code\.google\.com/]
```

Changes:

` * ftp.gnu.org urls should be https://, not http:// (url is http://ftp.gnu.org/gnu/gawk/gawk-4.1.1.tar.xz). `

Into:

` * ftp.gnu.org mirrors should be https://, not http:// (mirror is http://ftp.gnu.org/gnu/gawk/gawk-4.1.1.tar.xz). `.


:question:

chdiza:
That's a bit better, but one of the things that confuses me is the `(mirror/url is blah blah)`.  I don't understand the purpose of that, and to the user's eye it looks like what it's saying is "you messed up the url; the url is (i.e., is supposed to be) blah blah".

*Now* I think I get it, but it was confusing, and I don't see what's gained by putting it in there.  If it were deleted, it would be perfectly clear that all that's wrong is that there should be an extra `s` at the end of the protocol.  If I saw *that* version of things, I would think "Oh, put an `s` in".  The way that it is now, I think "what the heck is going on here" because there's info overload.

Perhaps the clearest of all would be for it to emit:
```
Change the `mirror` from: 
   http://ftp.gnu.org/gnu/gawk/blah
into
   https://ftp.gnu.org/gnu/gawk/blah.
```

xu-cheng:

PR #36753, opened 2015-02-12
make tap symlink free

xu-cheng:
Per https://github.com/Homebrew/homebrew/issues/36731

Ping @mikemcquaid, @jacknagel, @DomT4

PR #39313, opened 2015-05-03
add Formulary::path method

xu-cheng:
This is a little code refactoring splited from #36753

The idea is to eliminate `Formula#path` outside of formulary. And I indent to deprecate `Formula#path` method when I reimplement symlink free tap function.

### DIFF
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -183,6 +183,7 @@ class Formulary
   class NullLoader < FormulaLoader
     def initialize(name)
       @name = name
+      @path = Formula.path(name)
     end
 
     def get_formula(spec)
@@ -202,6 +203,10 @@ class Formulary
 
   def self.canonical_name(ref)
     loader_for(ref).name
+  end
+
+  def self.path(ref)
+    loader_for(ref).path
   end
 
   def self.loader_for(ref)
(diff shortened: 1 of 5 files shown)


xu-cheng:
cc @Homebrew/owners

jacknagel:
I'm not sure `canonical_path` is the right name, maybe just `path`?

`canonical_name` can take several ways of referring to a formula and return the canonical form, but I'm not sure it makes sense to ask for the "canonical path" when there is only one possible path.

xu-cheng:
I think `canonical_path` is just as the same as `canonical_name`. It can take exactly the same input as `canonical_name`(i.e. formula name, alias and full qualified name).

jacknagel:
`canonical_name` canonicalizes a name, but `canonical_path` does not canonicalize a path. It canonicalizes its argument in order to determine the correct path, but that is secondary to its main function.

xu-cheng:
OK, updated the method name to `Formulary::path`.

xu-cheng:
More comment?

Library/Homebrew/formulary.rb
```diff
@@ -183,6 +183,7 @@ def get_formula(spec)
   class NullLoader < FormulaLoader
     def initialize(name)
       @name = name
+      @path = Formula.path(name)
```
jacknagel: nit, but we can just call `super name, Formula.path(name)`.

jacknagel:
one comment, otherwise lgtm

xu-cheng:
Fixed the nit and :ship:ed.

PR #39317, opened 2015-05-03
pound 2.7, plus other changes for present-day style

L2G:
I'm not sure about the change to LDFLAGS and the addition of CFLAGS, so some guidance would be appreciated there. I had a gut feeling that the old modification to LDFLAGS was no longer the right way.

### DIFF
--- a/Library/Formula/pound.rb
+++ b/Library/Formula/pound.rb
@@ -1,20 +1,20 @@
-require 'formula'
-
 class Pound < Formula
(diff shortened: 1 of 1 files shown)


jacknagel:
Pulled earlier request #37812

L2G:
There is a change I made that was not in #37812, so watch this space.

L2G:
Ugh. Fine... #39352

PR #39319, opened 2015-05-03
hive 1.1.0

kashif:
also removed conflict with spark’s beeline binary by renaming spark’s
version to spark-beeline


tdsmith:
Thanks! Please use `sha256` instead of `sha1` moving forward.

PR #39322, opened 2015-05-03
audit: flag incorrect automake/autoconf/libtool dep handling

### LINKED ISSUE #39303: audit command should catch `depends_on :autoconf` and `depends_on :automake`
Per <del>@tdsmith</del>, these dependencies are deprecated in favor of `depends_on "autoconf" => :build` and `depends_on "automake" => :build`. The `audit` command does not catch them, though.

DomT4:
Just makes the audit cough out this:
```
 * :libtool is deprecated. Usage should be "libtool"
 * :autoconf is deprecated. Usage should be "autoconf"
 * :automake is deprecated. Usage should be "automake"
```

Closes #39303.

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -586,6 +586,10 @@ class FormulaAuditor
       problem "\"#{$1}\" should be \"\#{#{$2}}\""
     end
 
+    if line =~ %r[depends_on :(automake|autoconf|libtool)]
+      problem ":#{$1} is deprecated. Usage should be \"#{$1}\""
+    end
+
     # Commented-out depends_on
     if line =~ /#\s*depends_on\s+(.+)\s*$/
       problem "Commented-out dep #{$1}"


L2G:
I'd suggest a small modification:

    * :libtool is deprecated. Usage should be: "libtool" => :build

(And so on.)

DomT4:
I thought about that, but it's not always true. A few things use normally build-time deps as ` depends_on "example" => :run ` - So we don't actually want to flag a warning in that situation. I believe there's an existing Audit rule to catch the situation where a build-time dep isn't declared either ` :build ` or `:run`.

L2G:
>  I believe there's an existing Audit rule to catch the situation where a build-time dep isn't declared either `:build` or `:run`.

I believe the only thing that catches this now is the sharp eyes of the core maintainers doing the code reviews. :grin:

DomT4:
> I believe the only thing that catches this now is the sharp eyes of the core maintainers doing the code reviews

Change wget's `autoconf => :build ` to `autoconf` and then run `brew audit` against it and you'll see this:

```
wget:
 * autoconf dependency should be
Or if it is indeed a runtime denpendency
  depends_on "autoconf" => :run

```

:smiley_cat:

L2G:
Is it possible these weren't showing up for the formula I was auditing because I was running `brew audit` *before* `brew install` and not after?

DomT4:
Shouldn't have been, no. It won't have flagged your use of `:automake` at this point, but it should have flagged it if you'd done something like `depends_on :automake` instead of `depends_on :automake => :build`.

PR #39323, opened 2015-05-03
boost 1.58.0

dunn:
closes #38743, which never got finished.

### DIFF
--- a/Library/Formula/boost.rb
+++ b/Library/Formula/boost.rb
@@ -1,7 +1,7 @@
 class Boost < Formula
   homepage "http://www.boost.org"
-  url "https://downloads.sourceforge.net/project/boost/boost/1.57.0/boost_1_57_0.tar.bz2"
-  sha1 "e151557ae47afd1b43dc3fac46f8b04a8fe51c12"
+  url "https://downloads.sourceforge.net/project/boost/boost/1.58.0/boost_1_58_0.tar.bz2"
+  sha256 "fdfc204fc33ec79c99b9a74944c3e54bd78be4f7f15e260c0e2700a36dc7d3e5"
 
   head "https://github.com/boostorg/boost.git"
 
@@ -38,7 +38,7 @@ class Boost < Formula
 
   def install
     # https://svn.boost.org/trac/boost/ticket/8841
-    if build.with? "mpi" and build.with? "single"
+    if build.with?("mpi") && build.with?("single")
       raise <<-EOS.undent
         Building MPI support for both single and multi-threaded flavors
         is not supported.  Please use "--with-mpi" together with
--- a/Library/Formula/libswiften.rb
+++ b/Library/Formula/libswiften.rb
@@ -1,11 +1,10 @@
-require 'formula'
-
 class Libswiften < Formula
-  homepage 'http://swift.im/swiften'
+  homepage "http://swift.im/swiften"
+  revision 1
 
   stable do
     url "http://swift.im/downloads/releases/swift-2.0/swift-2.0.tar.gz"
-    sha1 "b04ba098fffb1edc2ef0215957371c249458f0be"
+    sha256 "cbcdbe527dc4d112a38f3cdec5f1051d3beba0b97f8082f90debd04a5b45a41f"
 
     # Patch to include lock from boost. Taken from
     # http://comments.gmane.org/gmane.linux.redhat.fedora.extras.cvs/957411
@@ -29,13 +28,14 @@ class Libswiften < Formula
   end
 
   head do
-    url 'git://swift.im/swift'
-    depends_on 'lua' => :recommended
+    url "git://swift.im/swift"
+    depends_on "lua" => :recommended
   end
 
-  depends_on 'scons' => :build
-  depends_on 'libidn'
-  depends_on 'boost'
+  depends_on "scons" => :build
+  depends_on "boost"
+  depends_on "libidn"
+  depends_on "openssl"
 
   def install
     boost = Formula["boost"]
@@ -52,6 +52,7 @@ class Libswiften < Formula
       libidn_includedir=#{libidn.include}
       libidn_libdir=#{libidn.lib}
       SWIFTEN_INSTALLDIR=#{prefix}
+      openssl=#{Formula["openssl"].opt_prefix}
     ]
 
     if build.with? "lua"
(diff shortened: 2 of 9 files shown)


dunn:
Fixed the checks locally; should I wait before queuing the build again?

dunn:
```
libswiften:
  /usr/local/Cellar/libswiften/2.0_1/lib/libSwiften.2.0.dylib
```
I guess it didn't help?

DomT4:
Scons stuff usually requires a firm shove in the right direction. Add ` openssl=#{Formula["openssl"].opt_prefix} ` to the ` args ` list. Tested locally, works fine.

dunn:
Should I wait for comments from the maintainers before running another build?

Library/Formula/boost.rb
```diff
@@ -137,7 +137,6 @@ def caveats
       only supported on x86_64.
       EOS
     end
-
```
DomT4: This line is usually left here in the ` caveats `. Not sure if that's deliberate, but it's usually there :smiley_cat:

Library/Formula/libswiften.rb
```diff
@@ -2,7 +2,7 @@
 
 class Libswiften < Formula
   homepage 'http://swift.im/swiften'
-
+  revision 1
   stable do
```
DomT4: Make sure there's an empty line between ` revision ` and ` stable `.

DomT4:
If you wanted to, you could fix any errors raised by ` brew audit --strict ` but aside that, looks good to me. Essentially, ensure everything revisioned actually needs revisioned. I can't think of much else the maintainers would need to ping a discussion on, but I'm fairly frequently wrong :wink:.

dunn:
OK, I wasn't sure if the protocol for revisions was to leave the rest of the formula untouched.  Since libswiften needed more help, though, I'll go ahead and audit it.

Yeah, I just revised the ones you suggested in #38743.  Did those all fail that time?

DomT4:
> OK, I wasn't sure if the protocol for revisions was to leave the rest of the formula untouched

If there is, I've been breaking it as often as possible accidentally :grinning:.

> I just revised the formulae you suggested in #38743.

Forgot about my comment there. Yup, those failed :+1:.

dunn:
Cool, thanks for the help!

Library/Formula/boost.rb
dunn: Oops, you're right.

Library/Formula/rcssserver.rb
```diff
@@ -75,8 +74,7 @@ def install
   end
 
   test do
-    system "#{bin}/rcssserver help | head -1 | grep 'rcssserver-#{version}'"
-    system "#{bin}/rcsslogplayer --version | tail -1 | grep 'rcsslogplayer Version'"
-    system "#{bin}/rcssmonitor --version | tail -1 | grep 'rcssmonitor Version'"
+    assert_match version.to_s,
+                 shell_output("#{bin}/rcssserver help", 1)
```
jacknagel: Please don't break these assertions across multiple lines unless it is super long or something.

dunn:
Shit, accidentally queued two builds because I was in a rush and pushed prematurely. Many apologies.

dunn:
```
==> Installing ledger dependency: boost-python
==> Downloading https://downloads.sourceforge.net/project/boost/boost/1.58.0/boost_1_58_0.tar.bz2
Already downloaded: /Library/Caches/Homebrew/boost-python-1.58.0.tar.bz2
Error: boost was built in C++11 mode so boost-python must be built with --c++11.
```
Should `cxx11` be required for boost-python?

DomT4:
@tdsmith's PR on `ledger` passed fine a couple of days ago so that'd fairly strongly implies the status quo is fine. I'm not quite sure why Boost is being built with `c++11` on Mavericks and Yosemite in this PR - It should only build `c++11` by default on 10.8 and below.

tdsmith:
cpp-netlib has an improper (I think) unconditional dependency on `"boost" => "c++11"`.

tdsmith:
This looks like it should be a safe change: http://upstream-tracker.org/compat_reports/boost/1.57.0_to_1.58.0/abi_compat_report.html -- a little surprised we need to revision anything.

DomT4:
> cpp-netlib has an improper (I think) unconditional dependency on "boost" => "c++11".

@dunn Might be worth changing that in this PR and seeing if that kicks anything.

> a little surprised we need to revision anything.

Agreed, It looks good. Test bot had a field day breaking in the other PR though :cry:.

dunn:
cpp-netlib built fine against non-cxx11 boost locally, so that's something.

dunn:
But not on Mountain Lion.

DomT4:
Oh, you probably need to do something like this to `cpp-netlib`:

```ruby
  if MacOS.version < :mavericks
    depends_on "boost" => "c++11"
  else
    depends_on "boost"
  end
```

dunn:
Oh, yeah; I was thinking the boost formula itself might need fiddling but that looks right.

DomT4:
Both `points2grid`, `libswiften` and `freeling` failed on ML with ` Undefined symbols for architecture x86_64: ` errors. `ledger`'s test failed.

Only ML for all of these errors. Mavericks and Yosemite are :100:.

dunn:
But they passed in earlier builds, I thought?

DomT4:
They did, yeah. Might be ` c++11 ` issues again given that ML was the only failure, but :confounded:.

tdsmith:
The problem is related to how the test bot builds packages; packages built against libstdc++ need boost build against libstdc++ and packages built against libc++ need boost built against libc++. On Mavericks and newer, there's no problem since everything is build against libc++. On Mountain Lion, though, there's a mix. If a package needs a libc++ boost, `depends_on "boost" => "c++11"` means it will always get it, but there's no "without-c++11" option, so a package that needs a libstdc++ boost will sometimes fail if it's being built after a package that needed a libc++ boost.

I propose to pull this despite the Mountain Lion failures; the side effect is that this will remove the Mountain Lion bottles, but they can be rebottled separately. I assume the packages still work.

I see you added some new `boost => "c++11"` dependencies to troubleshoot; those should be removed.

DomT4:
Tim's solution sounds good. Would be good to get someone using 10.8 to use the newly generated Boost bottles to check the other affected formulae in this PR do indeed work, but who is available and willing to test on a 10.8 machine isn't entirely obvious to me.

dunn:
OK, reverted that.

tdsmith:
ty! :fish_cake:

PR #39324, opened 2015-05-03
[RFC] Facebook Path Picker

pcottle:
Hey everyone! This is an RFC (request for comments) for a project that Facebook is going to open source soon. I wish I could put this up in a better state but I'm entirely new to homebrew recipes and ruby, so apologies in advance.

Some real quick context:
* We are open sourcing a project called "PathPicker", aliased to "fpp" for short. Best context is here: http://facebook.github.io/PathPicker/
* The project is a Python **application**, aka not going to be imported at all in code but simply run from the command line.
* It has no external dependencies besides python -- it's just:
    * One bash script
    * ~10 python files that uses standard ~2.7 libraries

Installing it theoretically is quite easy. It would essentially just be:
* Unarchive the provided tar file to somewhere permanent
* Symlink the bash script into `/usr/bin/local` or wherever `$PATH` includes

The following recipe attempts to do the above but with some obvious uncommitable hacks. Hopefully some feedback can be provided -- almost all the python recipes I found were using `setup.py` and other things that were not as simple.

Anyways sorry for the low quality PR but hopefully this is easy to improve!

### DIFF
--- a/Library/Formula/fpp.rb
+++ b/Library/Formula/fpp.rb
@@ -7,7 +7,6 @@ class Fpp < Formula
   depends_on :python if MacOS.version <= :snow_leopard
 
   def install
-    puts buildpath
     # we need to copy the bash file and source python files
     libexec.install Dir["*"]
     # and then symlink the bash file
--- a/Library/Formula/fpp.rb
+++ b/Library/Formula/fpp.rb
@@ -1,6 +1,5 @@
 class Fpp < Formula
   homepage "https://facebook.github.io/PathPicker/"
-  # TODO -- change away from gh-pages link
   url "https://facebook.github.io/PathPicker/dist/fpp.0.5.0.tar.gz"
   sha256 "634cbd2f501639186561418bf2036529d3995676836378b5865e7130d34733e8"
   head "https://github.com/facebook/pathpicker.git"
--- a/Library/Formula/fpp.rb
+++ b/Library/Formula/fpp.rb
@@ -1,8 +1,8 @@
 class Fpp < Formula
   homepage "https://facebook.github.io/PathPicker/"
-  # To be changed -- just a sample tar for now
-  url "https://github.com/pcottle/example_for_pip/raw/master/dist/pcottleexample.0.0.4.tar.gz"
-  sha256 "25f3edddc28b7a353c407d48a74a92f9a364160de790122357a3785d9d9e8aeb"
+  # TODO -- change away from gh-pages link
+  url "https://facebook.github.io/PathPicker/dist/fpp.0.5.0.tar.gz"
+  sha256 "634cbd2f501639186561418bf2036529d3995676836378b5865e7130d34733e8"
   head "https://github.com/facebook/pathpicker.git"
 
   depends_on :python if MacOS.version <= :snow_leopard
@@ -12,11 +12,11 @@ class Fpp < Formula
     # we need to copy the bash file and source python files
     libexec.install Dir["*"]
     # and then symlink the bash file
-    bin.install_symlink libexec/"example"
+    bin.install_symlink libexec/"fpp"
   end
 
   test do
-    system "example --help"
+    system "fpp --help"
   end
 
 end
--- a/Library/Formula/fpp.rb
+++ b/Library/Formula/fpp.rb
@@ -16,7 +16,6 @@ class Fpp < Formula
   end
 
   test do
-    system "fpp --help"
+    system bin/"fpp", "--help"
   end
-
 end
--- a/Library/Formula/fpp.rb
+++ b/Library/Formula/fpp.rb
@@ -1,7 +1,7 @@
 class Fpp < Formula
   homepage "https://facebook.github.io/PathPicker/"
-  url "https://facebook.github.io/PathPicker/dist/fpp.0.5.0.tar.gz"
-  sha256 "634cbd2f501639186561418bf2036529d3995676836378b5865e7130d34733e8"
+  url "https://facebook.github.io/PathPicker/dist/fpp.0.5.3.tar.gz"
+  sha256 "94b77ef10a128a694f6302ce578a2d07a3fd2892299d341b22be9496abd7277d"
   head "https://github.com/facebook/pathpicker.git"
 
   depends_on :python if MacOS.version <= :snow_leopard
--- a/Library/Formula/fpp.rb
+++ b/Library/Formula/fpp.rb
@@ -15,7 +15,7 @@ class Fpp < Formula
     bin.install_symlink libexec/"example"
   end
 
-  def test
+  test do
     system "example --help"
   end
 
--- a/Library/Formula/fpp.rb
+++ b/Library/Formula/fpp.rb
@@ -1,23 +1,18 @@
 class Fpp < Formula
   homepage "https://facebook.github.io/PathPicker/"
-  head "https://github.com/facebook/pathpicker/"
-  version "0.0.3"
   # To be changed -- just a sample tar for now
-  url "https://github.com/pcottle/example_for_pip/raw/master/dist/pcottleexample.tar.gz"
-  sha256 "038753f944edf7b3d6b0dbcf0056350eead1e14db4a3c70fafb711e9a31be3d4"
+  url "https://github.com/pcottle/example_for_pip/raw/master/dist/pcottleexample.0.0.4.tar.gz"
+  sha256 "25f3edddc28b7a353c407d48a74a92f9a364160de790122357a3785d9d9e8aeb"
+  head "https://github.com/facebook/pathpicker.git"
 
-  # I have a dependency on python 2.7 -- is this ok?
-  depends_on :python => :recommended if MacOS.version <= :snow_leopard
+  depends_on :python if MacOS.version <= :snow_leopard
 
   def install
-    puts 'Unpacking PathPicker'
-    # obviously a total hack -- my two questions:
-    # -- how can I correctly access the destination of where url was downloaded?
-    # -- how can I extract this tar correctly to somewhere permanent?
-    system "tar xopf /Library/Caches/Homebrew/fpp-0.0.3.tar.gz -C /Users/pcottle/Desktop/"
-    puts 'Symlinking bash script'
-    # and this is a hack as well -- but basically i just want to symlink one file from the tar
-    ln_sf '/Users/pcottle/Desktop/example', '/usr/local/bin/fpp'
+    puts buildpath
+    # we need to copy the bash file and source python files
+    libexec.install Dir["*"]
+    # and then symlink the bash file
+    bin.install_symlink libexec/"example"
   end
 
 end
--- a/Library/Formula/fpp.rb
+++ b/Library/Formula/fpp.rb
@@ -1,0 +1,23 @@
+class Fpp < Formula
+  homepage "https://facebook.github.io/PathPicker/"
+  head "https://github.com/facebook/pathpicker/"
+  version "0.0.3"
+  # To be changed -- just a sample tar for now
+  url "https://github.com/pcottle/example_for_pip/raw/master/dist/pcottleexample.tar.gz"
+  sha256 "038753f944edf7b3d6b0dbcf0056350eead1e14db4a3c70fafb711e9a31be3d4"
+
+  # I have a dependency on python 2.7 -- is this ok?
+  depends_on :python => :recommended if MacOS.version <= :snow_leopard
+
+  def install
+    puts 'Unpacking PathPicker'
+    # obviously a total hack -- my two questions:
+    # -- how can I correctly access the destination of where url was downloaded?
+    # -- how can I extract this tar correctly to somewhere permanent?
+    system "tar xopf /Library/Caches/Homebrew/fpp-0.0.3.tar.gz -C /Users/pcottle/Desktop/"
+    puts 'Symlinking bash script'
+    # and this is a hack as well -- but basically i just want to symlink one file from the tar
+    ln_sf '/Users/pcottle/Desktop/example', '/usr/local/bin/fpp'
+  end
+
+end
--- a/Library/Formula/fpp.rb
+++ b/Library/Formula/fpp.rb
@@ -15,4 +15,8 @@ class Fpp < Formula
     bin.install_symlink libexec/"example"
   end
 
+  def test
+    system "example --help"
+  end
+
 end


Library/Formula/fpp.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  # I have a dependency on python 2.7 -- is this ok?
+  depends_on :python => :recommended if MacOS.version <= :snow_leopard
+
+  def install
+    puts 'Unpacking PathPicker'
+    # obviously a total hack -- my two questions:
+    # -- how can I correctly access the destination of where url was downloaded?
+    # -- how can I extract this tar correctly to somewhere permanent?
+    system "tar xopf /Library/Caches/Homebrew/fpp-0.0.3.tar.gz -C /Users/pcottle/Desktop/"
```
pcottle: this line needs the most feedback

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,23 @@
...
+    puts 'Symlinking bash script'
+    # and this is a hack as well -- but basically i just want to symlink one file from the tar
+    ln_sf '/Users/pcottle/Desktop/example', '/usr/local/bin/fpp'
```
pcottle: as well as this

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,23 @@
+class Fpp < Formula
+  homepage "https://facebook.github.io/PathPicker/"
+  head "https://github.com/facebook/pathpicker/"
```
DomT4: ` https://github.com/facebook/pathpicker.git `.

DomT4: (And should go below the `sha256` of the stable version).

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,23 @@
...
+  version "0.0.3"
```
DomT4: Unless the URL is versioned in a convoluted way, `brew` will be able to read the version automatically. `brew audit` can point this out to you once you actually have a versioned tarball to compare against :)

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,23 @@
...
+  # To be changed -- just a sample tar for now
+  url "https://github.com/pcottle/example_for_pip/raw/master/dist/pcottleexample.tar.gz"
+  sha256 "038753f944edf7b3d6b0dbcf0056350eead1e14db4a3c70fafb711e9a31be3d4"
+
+  # I have a dependency on python 2.7 -- is this ok?
+  depends_on :python => :recommended if MacOS.version <= :snow_leopard
```
DomT4: If you need Python 2.7 this is mandatory rather than recommended. ` depends_on :python if MacOS.version <= :snow_leopard `

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def install
+    puts 'Unpacking PathPicker'
```
DomT4: Murder all the ` puts ` :)

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,23 @@
...
+    # obviously a total hack -- my two questions:
+    # -- how can I correctly access the destination of where url was downloaded?
```
DomT4: > how can I correctly access the destination of where url was downloaded

` buildpath `.

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,23 @@
...
+    # -- how can I extract this tar correctly to somewhere permanent?
```
DomT4: You can't, per se. The extracted tarball will be removed after Brew is done with it. The tarball itself will be kept in ` HOMEBREW_CACHE ` until deleted.

Library/Formula/fpp.rb
DomT4: You don't need this at all. Brew will extract the tarball automatically to ` /tmp ` and then build from there.

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,23 @@
...
+    system "tar xopf /Library/Caches/Homebrew/fpp-0.0.3.tar.gz -C /Users/pcottle/Desktop/"
+    puts 'Symlinking bash script'
```
DomT4: Murder this :scissors:

Library/Formula/fpp.rb
DomT4: ` bin.install buildpath/"example" `.

DomT4: Actually, ` bin.install "example" ` will work fine.

pcottle:
wow thanks for the past response @DomT4 ! Let me update...

DomT4:
> ~10 python files that uses standard ~2.7 libraries

If all these need to be installed, you can do ` libexec.install Dir["*"] ` and then ` bin.install_symlink libexec/"example" `.

pcottle:
```
[pcottle:~/Desktop/example_for_pip:master]$ brew uninstall ./fpp.rb; brew install ./fpp.rb
Uninstalling /usr/local/Cellar/fpp/0.0.4... (3 files, 16K)
==> Downloading https://github.com/pcottle/example_for_pip/raw/master/dist/pcottleexample.0.0.4.tar.gz
Already downloaded: /Library/Caches/Homebrew/fpp-0.0.4.tar.gz
/private/var/folders/g1/lrh7xk295hlblw1txwzlrn7jw6g20z/T/fpp20150503-45904-1yunklq
🍺  /usr/local/Cellar/fpp/0.0.4: 3 files, 16K, built in 2 seconds
[pcottle:~/Desktop/example_for_pip:master]$ example
hello from python world
```

works like a charm now!! woohoo, thanks a ton @DomT4. I'll update with the actual versioned tarball of our project next, but one quick question -- does everything look ok in this version? I added a test and `brew audit ./fpp.rb` comes out clean now.

the main trickery was actually in our bash script -- i had to modify it to follow recursive symlinks:

```
cd $(dirname $WHEREAMI)
while [ -h "$WHEREAMI" ]; do
  DEST=$(readlink $WHEREAMI)
  cd $(dirname $DEST)
  WHEREAMI=$(dirname $WHEREAMI)"/"$DEST
done
python ./src/example.py
```

but once I had that it was perfect

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,22 @@
...
+  homepage "https://facebook.github.io/PathPicker/"
+  # To be changed -- just a sample tar for now
+  url "https://github.com/pcottle/example_for_pip/raw/master/dist/pcottleexample.0.0.4.tar.gz"
+  sha256 "25f3edddc28b7a353c407d48a74a92f9a364160de790122357a3785d9d9e8aeb"
+  head "https://github.com/facebook/pathpicker.git"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    puts buildpath
```
DomT4: You can remove this and the two `# ` comments below. It's all fairly regular Homebrew stuff that doesn't need special explanation :)

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,22 @@
...
+    # we need to copy the bash file and source python files
+    libexec.install Dir["*"]
+    # and then symlink the bash file
+    bin.install_symlink libexec/"example"
+  end
+
+  test do
+    system "example --help"
```
DomT4: Ideally this would be improved, but it looks like it needs user input (?) in which case, this may be the best we can do.

Style should be ` system bin/"example", "--help" `.

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
```
DomT4: No need for blank line here :)

DomT4:
Few style nits left to chase up, but otherwise looks good :+1: :+1:.

pcottle:
oh whoops sorry didnt see your comments, justasec

pcottle:
```
[pcottle:~/Desktop/homebrew/Library/Formula:facebookPathPicker]$ brew test ./fpp.rb 
Testing fpp
==> /usr/local/Cellar/fpp/0.5.0/bin/fpp --help
```

alright I think that's everything! Note that I actually updated the tarball with the real `0.5` release, so everything works now:
```
[pcottle:~/Desktop/homebrew/Library/Formula:facebookPathPicker]$ fpp --help
********************************************************************************

Welcome to fpp, the Facebook PathPicker! We hope your stay 
with us is enjoyable.

To get started with fpp, pipe some kind of terminal output into the program.
Examples include:

    * git status | fpp
<<truncated>>
```

you can try it out if you'd like, but its basically the demo videos over at:
http://facebook.github.io/PathPicker/

zmwangx:
This concept looks cool, but 0.5.0 doesn't work since the way the project handles paths seems wrong (I would be careful not to `cd` into the executable path because this way all paths piped to editors or whatever are all relative to the executable path instead of the current working directory).

By the way, I have a question: since this is a small Python project without dependencies, why don't you package it with [setuptools](https://setuptools.pypa.io/en/latest/) and release it on PyPI, like how most Python packages are distributed?

pcottle:
Ah yeah that's a good point -- I added all the cd trickery for homebrew
since I had to keep following symlinks. Let me see if I can throw it all
inside some $()'s which is where it used to be.

For not using pypi, a few reasons:

1) It wasn't super easy to include bash scripts and symlink them into $PATH
with pypi. It's used more for code you import rather than apps you run.

2) I actually think more people have brew installed these days, at least at
Facebook with all the js development going on.

And 3) the ability to move off Python if we need to. It's technically just
an implementation detail and if someone wrote a curses UI in React (which
would be sweet) I would maybe use that for 2.0.

Let me update that top level bash script and get back to you!

On Monday, May 4, 2015, Zhiming Wang <notifications@github.com> wrote:

> This concept looks cool, but 0.5.0 doesn't work since the way the project
> handles paths seems wrong (I would be careful not to cd into the
> executable path because this way all paths piped to editors or whatever are
> all relative to the executable path instead of the current working
> directory).
>
> By the way, I have a question: since this is a small Python project
> without dependencies, why don't you package it with setuptools
> <https://setuptools.pypa.io/en/latest/> and release it on PyPI, like how
> most Python packages are distributed?
>
> —
> <https://github.com/Homebrew/homebrew/pull/39324#issuecomment-98656020>.
>


-- 
-Sent from Gmail Mobile

DomT4:
>  I added all the cd trickery for homebrew since I had to keep following symlinks

Why the need to follow the symlinks in this case? Is that unique to Homebrew?

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,20 @@
+class Fpp < Formula
+  homepage "https://facebook.github.io/PathPicker/"
+  url "https://facebook.github.io/PathPicker/dist/fpp.0.5.0.tar.gz"
+  sha256 "634cbd2f501639186561418bf2036529d3995676836378b5865e7130d34733e8"
+  head "https://github.com/facebook/pathpicker.git"
+
+  depends_on :python if MacOS.version <= :snow_leopard
```
tdsmith: If the script works on Python 2.6 you don't need this. If the script only works on Python 2.7, make sure the script's shebang uses `/usr/bin/env python`.

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,20 @@
...
+  def install
+    puts buildpath
+    # we need to copy the bash file and source python files
+    libexec.install Dir["*"]
+    # and then symlink the bash file
+    bin.install_symlink libexec/"fpp"
+  end
+
+  test do
+    system bin/"fpp", "--help"
```
tdsmith: A test that exercises some functionality of fpp would be great, though it doesn't need to be comprehensive.

tdsmith:
When tools look for things relative to their install paths, we can use `write_exec_script` instead of `install_symlink`.

Library/Formula/fpp.rb
tdsmith: (or moral equivalent)

tdsmith: Okay, I scrolled up and I see the comments about this above, :cool:

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,20 @@
...
+  head "https://github.com/facebook/pathpicker.git"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    puts buildpath
+    # we need to copy the bash file and source python files
+    libexec.install Dir["*"]
+    # and then symlink the bash file
+    bin.install_symlink libexec/"fpp"
```
tdsmith: For clarity, my comment about exec scripts means that this can be
`bin.write_exec_script libexec/"fpp"`

which writes a script that execs fpp instead of just dropping a symlink, so relative paths from the fpp executable should be correct.

DomT4:
> When tools look for things relative to their install paths, we can use write_exec_script instead of install_symlink.

Aye, this is what I was alluding to above.

pcottle:
> Why the need to follow the symlinks in this case? Is that unique to Homebrew?

not necessarily unique to homebrew, but its the first time the script has had to follow symlinks. That's because on our Facebook dev environments, we know the absolute path of where the script normally resides (so we just hardcode it).

But no worries! I updated the recipe with a version (0.5.3) that works for any amount of symlink jumps.

Given that it works with symlinks, ill leave `bin.install_symlink` rather than changing it back.

I think this might be ready to go in! What do you all think?

pcottle:
Also of course, `brew audit --strict ./fpp.rb` was clean and build passes as well

Library/Formula/fpp.rb
DomT4: The comment above still needs addressing. Looks good otherwise :)

Library/Formula/fpp.rb
```diff
@@ -0,0 +1,20 @@
+class Fpp < Formula
+  homepage "https://facebook.github.io/PathPicker/"
+  url "https://facebook.github.io/PathPicker/dist/fpp.0.5.3.tar.gz"
+  sha256 "94b77ef10a128a694f6302ce578a2d07a3fd2892299d341b22be9496abd7277d"
+  head "https://github.com/facebook/pathpicker.git"
```
DomT4: I presume this repo is being kept private until release? Just checking the link isn't dead/mistyped/etc.

zmwangx:
@pcottle Regarding your comments on setuptools and PyPI, here are my two cents:

> It wasn't super easy to include bash scripts and symlink them into $PATH
with pypi. It's used more for code you import rather than apps you run.

You can easily include scripts with setuptools using the `scripts` keyword argument of `setuptools.setup`. A better way to do this is to translate `fpp` from Bash to Python (which is trivial), and use `console_scripts` to create an entry point to, say, `fpp.main`. This way the program is also more portable. See [Building and Distributing Packages with Setuptools](https://setuptools.pypa.io/en/latest/setuptools.html) for details.

> I actually think more people have brew installed these days, at least at
Facebook with all the js development going on.

Yeah, but I think more people have `pip` installed, and `pip` is cross platform. `brew` on the other hand is OS X specific (well, there's also linuxbrew — which I personally use on servers that I don't have root access, but it is less tested and less adopted; Windows is completely out of scope). I guess you don't want to associate this too much with Python, which is fine.

> the ability to move off Python if we need to. It's technically just
an implementation detail and if someone wrote a curses UI in React (which
would be sweet) I would maybe use that for 2.0.

Sure. It's still nice to release it on PyPI though, with the portability benefit outlined above.

Anyway, the concept is great and the project just need to follow conventions a little bit better. (Also, the project will benefit from being compatible with Python 3 — mostly a few print calls, use `2to3` to see what might need to be changed.) I hope you all the best.

Library/Formula/fpp.rb
pcottle: yeah we have to keep it private for now since moving it to public creates a newsfeed event for all our followers, but ironically the `gh-pages` website is totally open.

but yeah i double checked the head link -- works if you are part of the FB team

pcottle:
> A better way to do this is to translate fpp from Bash to Python (which is trivial),

Have you looked at the bash script? I'm not 100% sure it's trivial due to the stdin consumption and then switch to tty input (which is especially needed when using `$EDITOR` to edit the selected files). Obviously launching a vim subprocess from python doesn't present the UI to the user running the script.

> and pip is cross platform. brew on the other hand is OS X specific 

ah yeah of course! Sorry you guys can't see our README yet since the repo is still private (grr) but we also have a section on installing without brew:
screen shot 2015-05-04 at 1 51 06 pm

its a pretty simple 2-3 step process, since we want to support linux as well. We are just adding homebrew for the quick-and-easy mac workflow.

> (Also, the project will benefit from being compatible with Python 3 — mostly a few print calls, use 2to3 to see what might need to be changed.)

yeah I agree -- its on my todolist :)

@mistydemeo  / @adamv / @jacknagel let me know if there's any more context you need, think this is ready for final review

tdsmith:
lgtm pending release

DomT4:
Still need to murder the ` puts buildpath `. We don't need to let users know where the buildpath is.

tdsmith:
In general I think it's probably not productive to give package authors advice on their packaging strategies on the Homebrew issue tracker; unless it's causing problems and hard to work around, it's off-topic.

Oops, agree with Dominyk, don't `puts` in `install`.

pcottle:
> Still need to murder the puts buildpath. We don't need to let users know where the buildpath is.

gah totally unintentional, didnt catch it since there was on string literal. updated!

@tdsmith thanks for the look-over. what do you mean by "pending release"? We have a bit of an awkward distributed locking situation here, as the recipe needs to land before we can announce and the repo needs to be opened before we can use github releases :fearful:

tdsmith:
Please do not plan on having a formula available in Homebrew/homebrew before your software is publicly available.

zmwangx:
@tdsmith Sorry, the repo is currently private so I had nowhere to discuss. I'll refrain from discussing this after this one final comment. (I'm not sure about if this is unsolicited advice though. This PR is marked as RFC, so I assumed that the author wasn't sure if this should go into Homebrew and tried to suggest a potentially better strategy; also it's related to Homebrew in that if it is distributed through PyPI then it should be excluded per [Acceptable Formulae](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Acceptable-Formulae.md).)

@pcottle The final comment. I did skim the bash script without going into details. If your only problem is with redirecting stdin, I've created an [example script](https://gist.github.com/zmwangx/23806da530a39bb1c0cf) that shows how you can redirect stdin (after exhausting it) to a tty when you either want to make a function call or launch an external program.

pcottle:
> Please do not plan on having a formula available in Homebrew/homebrew before your software is publicly available.

its technically available -- anyone can go download https://facebook.github.io/PathPicker/dist/fpp.0.5.3.tar.gz if they would like which contains all the source code. the launch page is out there for everyone to see too:
facebook.github.io/PathPicker/

tdsmith:
Oh! Okay, yep, that's fine. Sorry I didn't appreciate that; excuse me while I extract my foot from my mouth. I'll pull this tonight.

@zmwangx Homebrew frequently accepts formulae for CLI applications written in Python. You're right that we don't often accept Python libraries.

pcottle:
> Sorry I didn't appreciate that; excuse me while I extract my foot from my mouth. I'll pull this tonight.

No worries! Definitely a less-than-straightforward situation.

Also thanks @zmwangx for the example script -- i filed a task which youll be able to see when the repo goes live (soon!)

tdsmith:
Thanks for the pull request! 🎉 Homebrew depends on contributions from community members like you and we're grateful for your support. Just for future reference, our favorite commit message format is like `packagename 1.2.3 (new formula)` (in a single commit) for new formulae.

pcottle:
> just for future reference, our favorite commit message format is like packagename 1.2.3 (new formula) (in a single commit) for new formulae.

For sure! Was wondering if you wanted me to squash the commits. I'll keep this in mind in case anyone else from Facebook contributes a formula in the future. Thanks again! Very excited to launch with homebrew support -- now our install plan fits into a tweet

PR #39326, opened 2015-05-03
EmacsRequirement

dunn:
@xu-cheng floated the idea in #39266, so I figured I'd put this forward as a possibility.

I'm still getting familiar with the codebase and Ruby in general, so lmk if there are any obvious mistakes.

### DIFF
--- a/Library/Homebrew/requirements/emacs_requirement.rb
+++ b/Library/Homebrew/requirements/emacs_requirement.rb
@@ -1,0 +1,29 @@
+class EmacsRequirement < Requirement
+  fatal true
+
+  default_formula "emacs"
+
+  def initialize(tags)
+    @version = tags.shift if /\d+\.*\d*/ === tags.first
+    raise "Specify a version for EmacsRequirement" unless @version
+    super
+  end
+
+  satisfy :build_env => false do
+    next unless which "emacs"
+    emacs_version = `emacs --batch --eval "(princ emacs-version)"`
+    Version.new(emacs_version) >= Version.new(@version)
+  end
+
+  env do
+    ENV.prepend_path "PATH", which("emacs").dirname
+  end
+
+  def message
+    "Emacs #{@version} or later is required."
+  end
+
+  def inspect
+    "#<#{self.class.name}: #{name.inspect} #{tags.inspect} version=#{@version.inspect}>"
+  end
+end
--- a/Library/Homebrew/dependency_collector.rb
+++ b/Library/Homebrew/dependency_collector.rb
@@ -121,6 +121,7 @@ class DependencyCollector
     when :tuntap     then TuntapDependency.new(tags)
     when :ant        then ant_dep(spec, tags)
     when :apr        then AprDependency.new(tags)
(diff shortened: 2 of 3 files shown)


dunn:
Pulling in some comments from #39266: specific Emacs requirements might be too much of a corner case.  The formulae I found that can currently benefit are just four:
- [`magit.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/magit.rb) 
- [`cask.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/cask.rb) 
- [`mu.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mu.rb)
- [`proof-general.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/proof-general.rb#L28)

But in the future this might allow fewer formula to require brewed Emacs.

Library/Homebrew/requirements/emacs_requirement.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  default_formula "emacs"
+
+  def initialize(tags)
+    @version = tags.shift if /\d+\.*\d*/ === tags.first
+    raise "Specify a version for EmacsRequirement" unless @version
+    super
+  end
+
+  satisfy do
```
xu-cheng: `satisfy :build_env => false do`

Library/Homebrew/requirements/emacs_requirement.rb
```diff
@@ -0,0 +1,20 @@
...
+    emacs_version = `emacs --batch --eval "(princ emacs-version)"`.to_f
```
xu-cheng: Need to check whether emacs is installed first.

Library/Homebrew/requirements/emacs_requirement.rb
```diff
@@ -0,0 +1,20 @@
...
+    emacs_version >= @version.to_f
```
xu-cheng: Using `Version.new(version.to_s)`

xu-cheng:
There also should be a `env` block to change PATH.

I suggest you to have look at [RubyRequirement](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/requirements/ruby_requirement.rb) as example.

dunn:
Is the `env` block to ensure that the same `emacs` that's used to test the requirement is then used in the build process?

xu-cheng:
Actually, the way requirement works is to only check first program found in the PATH. So yes, they are the same emacs. Also could you add a `inspect` method as well? Thanks.

dunn:
For `inspect`, X11 uses `min_version=` while Java and Ruby just go with `version=`. Any preference here?

xu-cheng:
I think both are OK. But let's go with `version=`

xu-cheng:
LGTM. Will let @Homebrew/owners to have a look.

xu-cheng:
@dunn :ship:ed. Can you apply this change to the formulae you listed above? Thanks.

PR #39331, opened 2015-05-04
minizinc 2.0.1

dunn:


Library/Formula/minizinc.rb
```diff
@@ -1,53 +1,29 @@
-require 'formula'
-
 class Minizinc < Formula
-  homepage 'http://www.minizinc.org'
-  url 'http://www.minizinc.org/downloads/release-1.6/minizinc-1.6-x86_64-apple-darwin.tar.gz'
-  sha1 '71f0e08962eb8bb44c463851f0144c8b006fdb80'
+  homepage "http://www.minizinc.org"
+  head "https://github.com/MiniZinc/libminizinc.git",
+       :branch => "develop"
```
jacknagel: not sure I see a reason to break this across two lines?

Library/Formula/minizinc.rb
```diff
@@ -1,53 +1,29 @@
...
+
+  stable do
+    url "http://www.minizinc.org/downloads/release-2.0.1/libminizinc-2.0.1.tar.gz"
+    sha256 "b2dedd456d71d5670b24ec4b159a05d53c00e2aae232171c0e3be31cec915aff"
```
jacknagel: blank line between sha256 and patch

Library/Formula/minizinc.rb
dunn: I'll make it a single line; I use a large font size for editing so I break lines probably more than is necessary.

tdsmith:
Thanks!

PR #39332, opened 2015-05-04
dupx: evict to boneyard

L2G:


L2G:
Can someone shine some light on what Jenkins is running into here?

tdsmith:
There's no gdb in core.

L2G:
Is there a gdb somewhere? Because I have a /usr/local/bin/gdb on my system linked to /usr/local/Cellar/gdb/7.9/bin/gdb. For the life of me I can't remember where that would have come from if not from core or at least a tap…

xu-cheng:
There is gdb in `homebrew/dupes`. However, we don't allow mandatory cross tap dependency in the core formula.

L2G:
I understand. It was just driving me crazy that I couldn't remember where it came from.

L2G:
But that raises another question… why is it in homebrew/dupes if it's not present in OS X? (Or is it present in versions other than Mavericks?)

xu-cheng:
In OS X, we should use lldb.

xu-cheng:
> It was just driving me crazy that I couldn't remember where it came from.

You could run `brew info gdb` to check the origin.

L2G:
> You could run `brew info gdb` to check the origin.

…which was giving me nothing, which was why it was driving me crazy.

L2G:
The configure script says:

> configure: error: gdb is not in the path, this utility requires GDB

…so maybe dupx should just be thrown to the boneyard.

L2G:
https://github.com/Homebrew/homebrew-boneyard/pull/47

L2G:
Ping.

L2G:
@mikemcquaid Thanks. I see this was added to the boneyard, but I don't think the removal from core has happened yet…?

mikemcquaid:
Whoops, done in 9f95d11.

PR #39334, opened 2015-05-04
dns2tcp: brew audit --strict

L2G:


Library/Formula/dns2tcp.rb
```diff
@@ -1,12 +1,14 @@
...
+  sha256 "ea9ef59002b86519a43fca320982ae971e2df54cdc54cdb35562c751704278d9"
 
   def install
     system "./configure", "--prefix=#{prefix}", "--disable-dependency-tracking"
-    system "make install"
+    system "make", "install"
+  end
+
+  test do
+    assert_match(/^dns2tcp v#{version} /, `#{bin}/dns2tcpc -h 2>&1`)
```
xu-cheng: Please use `shell_ouput` instead of backtick

L2G: The test will not work with `shell_output`.

xu-cheng: Why?

PR #39337, opened 2015-05-04
pdksh: patches + fix install name

geoff-codes:
- Add important patches from upstream.
- Install as `pdksh` instead of `ksh`, which conflicts with [AT&T] ksh.
- Install template .kshrc and .profile files in `etc`.
- Cleanup, add test.

### DIFF
--- a/Library/Formula/pdksh.rb
+++ b/Library/Formula/pdksh.rb
@@ -1,25 +1,32 @@
-require 'formula'
-
 class Pdksh < Formula
-  homepage 'http://www.cs.mun.ca/~michael/pdksh/'
-  url 'http://www.cs.mun.ca/~michael/pdksh/files/pdksh-5.2.14.tar.gz'
-  sha1 '8b022e45de7691ef330f13b0981f0cff30b4047d'
+  homepage "http://www.cs.mun.ca/~michael/pdksh/"
+  url "http://www.cs.mun.ca/~michael/pdksh/files/pdksh-5.2.14.tar.gz"
+  sha256 "ab15bcdd50f291edc632bca351b2edce5405d4f2ce3854d3d548d721ab9bbfa6"
 
-  # Use a sort command that works with Leopard and up
+  patch :p2 do # Upstream patches
+    url "http://www.cs.mun.ca/~michael/pdksh/files/pdksh-5.2.14-patches.1"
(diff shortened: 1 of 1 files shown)


tdsmith:
What do the patches do, why are they important, and when can we remove them?

I think I'd rather not change the name of the installed executable.

geoff-codes:
> What do the patches do


From [the patch preamble](http://www.cs.mun.ca/~michael/pdksh/files/pdksh-5.2.14-patches.1):
```
Here are patches for 3 significant bugs in pdksh-5.2.14:
  - set -x dumps core (shf.c);
  - output of "jobs" command is filled with ^A characters (jobs.c);
  - "typeset -r foo=bar" fails saying foo is readonly (var.c).
```
 > why are they important

They are important because they patch critical bugs. They are mentioned on the pdksh homepage.

> when can we remove them

Probably never. The newest of these patches is dated 2001; pdksh 5.2.14 was released it 1999. It is not an actively maintained project, having largely been supplanted by shells like mksh and AT&T (a.k.a. the _real_) ksh, when it was open sourced in 2005.

> I think I'd rather not change the name of the installed executable.

Well, obviously, that's your call to make, but it would make homebrew the only [package](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=294592)  [manager](http://ftp.netbsd.org/pub/pkgsrc/current/pkgsrc/shells/pdksh/PLIST) that I know of to **not** do this, since having this piece of antiquity on your path as "ksh" _will_ break stuff (it is roughly equivalent to ksk88, not ksh93 like AT&T ksh or mksh). IMO, its a useful tool for testing the "lowest common denominator" for shell script compatibility, but that's about all.

In any case, it conflicts with the [ksh formula](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/ksh.rb), so if you elect to keep it as-is, I believe it should have a `conflicts_with` added, or be made keg-only.

tdsmith:
Thanks for the history lesson; renaming sounds reasonable.

What about the other patches (i.e. the macports one)? I'm not sure what the comment refers to.

Please add newlines between the `patch` blocks.

tdsmith:
Oh, we already applied that one, durr. Okay, thanks.

geoff-codes:
@tdsmith Newlines added.

tdsmith:
Thanks again!

PR #39338, opened 2015-05-04
Debian - Various

DomT4:
It's all in the commit messages, but essentially the root cause of this PR is to move things over from plaintext Debian mirrors to SSL/TLS protected Debian mirrors.

I've been meaning to start shoving things across for a while. 10 formulae every few days shouldn't be too odious to users updating, or maintainers, hopefully.

There's a couple of version bumps in there as well, added tests and style fixes, one build tweak and `le` has its official homepage & url back.

### DIFF
--- a/Library/Formula/le.rb
+++ b/Library/Formula/le.rb
@@ -1,18 +1,14 @@
-require 'formula'
-
 class Le < Formula
-  homepage 'http://freecode.com/projects/leeditor'
-  # url 'http://ftp.yar.ru/pub/source/le/le-1.14.9.tar.xz'
-  # upstream not responding, source from debian
-  url 'http://ftp.de.debian.org/debian/pool/main/l/le/le_1.14.9.orig.tar.gz'
-  sha1 'ce85cbefb30cf1f5a7e8349dbb24ffa0f65b1fd7'
+  homepage "https://github.com/lavv17/le"
+  url "http://lav.yar.ru/download/le/le-1.15.1.tar.xz"
+  sha256 "d9895ef82c89ae9cf30946bd43fa18b294f645e6a2bacd3ed9d39a3ccf324a4f"
 
-  conflicts_with 'logentries', :because => 'both install a le binary'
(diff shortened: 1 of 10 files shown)


dunn:
If you want a second https mirror, there's <https://mirrors.ocf.berkeley.edu>.

DomT4:
> If you want a second https mirror, there's https://mirrors.ocf.berkeley.edu.

Nice! Changed them all to that mirror :+1:.

Library/Formula/le.rb
```diff
@@ -1,18 +1,15 @@
...
+  homepage "https://github.com/lavv17/le"
+  url "http://lav.yar.ru/download/le/le-1.15.1.tar.xz"
+  sha256 "d9895ef82c89ae9cf30946bd43fa18b294f645e6a2bacd3ed9d39a3ccf324a4f"
 
-  conflicts_with 'logentries', :because => 'both install a le binary'
+  conflicts_with "logentries", :because => "both install a le binary"
 
   def install
     ENV.j1
+    system "./autogen.sh" if build.head?
```
jacknagel: I don't see a `head` defined for this formula?

DomT4: Hoops. Good catch. Will kill.

xu-cheng:
Next time, please don't mix formulae with version updated with formulae which don't. It's not friendly to bottle.

Merged in 91ac2027da899410fdca21d88b7150f93a5efda9. Except `dtc.rb` formula which need to set version string.

DomT4:
That was actually fairly deliberate. Only two of the formulae have bottles - The other 8 could have used having bottles added. Since the bot generates bottles anyway, it wasn't generating any extra work on the bot to do so. Was working with a prior stated desire to bottle as much as possible, but if that's no longer a desire that's fine.

xu-cheng:
OK. Also you need to resubmit `dtc` in another PR. The version is wrong.

DomT4:
Oh, strange. Didn't hit any problems locally but maybe I just missed it. Will check it out tonight. Thanks.

PR #39340, opened 2015-05-04
ilmbase 2.2.0

dunn:


Library/Formula/ilmbase.rb
```diff
@@ -1,10 +1,8 @@
-require 'formula'
-
 class Ilmbase < Formula
-  homepage 'http://www.openexr.com/'
-  url 'http://download.savannah.gnu.org/releases/openexr/ilmbase-2.1.0.tar.gz'
-  mirror 'http://download-mirror.savannah.gnu.org/releases/openexr/ilmbase-2.1.0.tar.gz'
-  sha1 '306d76e7a2ac619c2f641f54b59dd95576525192'
+  homepage "http://www.openexr.com/"
+  url "http://download.savannah.nongnu.org/releases/openexr/ilmbase-2.2.0.tar.gz"
```
tdsmith: Do I want to ask what `nongnu` is about?

dunn: Not 100 percent sure? But it looks just like a mirror for Savannah stuff.

tdsmith: Oh, it's on the front page:

> Savannah aims to be a central point for development, maintenance and distribution of official GNU software. In addition, for projects that support free software but are not part of GNU, we provide savannah.nongnu.org.

tdsmith:
Thanks!

PR #39346, opened 2015-05-04
mariadb 10.1.4 --devel

makigumo:


tdsmith:
Thanks, especially for the comments about the patch!

PR #39352, opened 2015-05-04
pound: remove overly-paranoid LDFLAGS append

L2G:
I have tested this and found that the brewed pcre library will be linked even without appending to the value of `LDFLAGS`.


tdsmith:
Added in a65e482b8b4bb81 which probably? predated superenv so this smells right to me.

L2G:
Hmmm. I'm using Mavericks, so I don't know if this would still be necessary in Mountain Lion…

PR #39354, opened 2015-05-04
beecrypt, gptfdisk, mapnik: up revision for icu4c update

vszakats:


DomT4:
`couchdb` was done a couple days back. Doesn't need redoing.

DomT4:
Ideally, add tests so we can actually catch these things automatically in future. That'd be awesome. I've (attempted) to deal with `mpd` at least in #39360. `couchdb` doesn't seem to have a practical, workable test which is frustrating.

vszakats:
Added strict audit updates and test for `beecrypt`. I'm passing on `mapnik` for now, it's just too large to build and install.

jacknagel:
For future reference, when a formula is updated and its shared library version changes, we need to start from the assumption that *all* dependents need their revisions bumped. If they have dynamic linkage, they will be broken otherwise.

Library/Formula/beecrypt.rb
```diff
@@ -26,8 +24,34 @@ def install
...
+        sha256Process(&hash);
+        sha256Digest(&hash, digest);
+
+        printf("%s\\n", b64crc(digest, 32));
+
+        return 0;
+      }
+    EOS
+    system ENV.cc, "test.c", "-lbeecrypt", "-o", "test"
+    assert_match /ZF8D/, shell_output("./test")
```
tdsmith: Awesome test, thanks! :100:

tdsmith:
Thanks for these; one commit per formula would be great going forward.

PR #39358, opened 2015-05-04
ctags: removing merged patch

### LINKED ISSUE #39351: ctags HEAD failed to patch on 10.10
One of the patches to `ctags` fails when installing the HEAD version. To reproduce, uninstall `ctags` as necessary and reinstall with `brew install --HEAD ctags`.

[Relevant gist](https://gist.github.com/naegelejd/45b50e0c0ca86334c073)

naegelejd:
The patch to `routines.c` has been merged upstream. Fixes #39351.

### DIFF
--- a/Library/Formula/ctags.rb
+++ b/Library/Formula/ctags.rb
@@ -19,6 +19,15 @@ class Ctags < Formula
   # fixes http://sourceforge.net/tracker/?func=detail&aid=3247256&group_id=6556&atid=106556
   patch :p2, :DATA
 
+  stable do
+    # also fixes http://sourceforge.net/tracker/?func=detail&aid=3247256&group_id=6556&atid=106556
+    # merged upstream but not yet in stable
+    patch :p2 do
+      url "https://gist.githubusercontent.com/naegelejd/9a0f3af61954ae5a77e7/raw/16d981a3d99628994ef0f73848b6beffc70b5db8/Ctags%20r782"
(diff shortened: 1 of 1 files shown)


DomT4:
The patch presumably is still necessary for the ` stable ` build?

naegelejd:
Ah I suppose so. I successfully installed `ctags` with `--build-from-source` *without* that part of the patch and can't replicate the [original bug](http://sourceforge.net/tracker/?func=detail&aid=3247256&group_id=6556&atid=106556) but I'm on 10.10. The remaining portion of the patch is necessary.

DomT4:
It didn't seem to cause any failure on 10.8 which uses a form of compiler reported in the original bug (I believe), but we no longer have a 10.7 test bot to check that OS X version.

tdsmith:
Re-introducing undefined behavior is likely to lead to spooky failures; I'll host this as a gist sometime tomorrow.

naegelejd:
Make sense. Is it more sensible to:

1. Determine if this will re-introduce bugs in older version of OS X and go from there, or
2. Just put the `patch` inside a `head do ... end`?

jacknagel:
Put the patch in a `stable.. end` block, yeah.

naegelejd:
OK done.

jacknagel:
Sorry, isn't the patch that has been deleted the one we want to apply to stable?

naegelejd:
My mistake... sorry about that. I created a Gist to host the stable-only patch in the meantime.

PR #39360, opened 2015-05-04
mpd: add test & install example conf file

DomT4:
Follow up to #39353. The test should ensure we catch the need to bump for `icu4c` in future.

### DIFF
--- a/Library/Formula/mpd.rb
+++ b/Library/Formula/mpd.rb
@@ -4,7 +4,7 @@ class Mpd < Formula
 
   stable do
     url "http://www.musicpd.org/download/mpd/0.19/mpd-0.19.9.tar.xz"
-    sha1 "6683bee5f132eda318c5a61ec14b2df8d9164d60"
+    sha256 "47851423053cd38cfad65be5985b41b7cd5bdbe8d0d13378e11748a28b19f26f"
   end
 
   bottle do
@@ -70,6 +70,7 @@ class Mpd < Formula
       --disable-debug
       --disable-dependency-tracking
       --prefix=#{prefix}
+      --sysconfdir=#{etc}
       --enable-bzip2
       --enable-ffmpeg
       --enable-fluidsynth
@@ -89,7 +90,9 @@ class Mpd < Formula
     system "./configure", *args
     system "make"
     ENV.j1 # Directories are created in parallel, so let's not do that
-    system "make install"
+    system "make", "install"
+
+    (etc+"mpd").install "doc/mpdconf.example" => "mpd.conf"
   end
 
   plist_options :manual => "mpd"
@@ -115,5 +118,19 @@ class Mpd < Formula
     </dict>
     </plist>
     EOS
+  end
+
+  test do
+    pid = fork do
+      exec "#{bin}/mpd --stdout --no-daemon --no-config"
+    end
+    sleep 2
+
+    begin
+      assert_match /OK MPD/, shell_output("curl localhost:6600")
+    ensure
+      Process.kill("SIGINT", pid)
+      Process.wait(pid)
+    end
   end
 end


Library/Formula/mpd.rb
```diff
@@ -116,4 +119,11 @@ def plist; <<-EOS.undent
     </plist>
     EOS
   end
+
+  test do
+    pid = fork { exec "#{bin}/mpd --stdout --no-daemon --no-config" }
+    sleep 2
+    Process.kill("SIGINT", pid)
+    Process.wait(pid)
+  end
```
tdsmith: Can this fail? Will we get an exception back if mpd terminates abnormally?

DomT4: Should do. If it doesn't launch correctly, say because it can't find the correct linked `dylib` it shouldn't start the process, and there shouldn't be anything to kill. I'll revert my `icu4c` tonight and then install `mpd` again and then upgrade `icu4c` and run the test and see if it blows up as intended.

DomT4: Garg. Test succeeds even when it shouldn't. Will have to try and find another way, if there is one.

DomT4: Alright, I don't know. We could go ugly and obvious as in:

```ruby
  test do
    icu4c = Formula["icu4c"].opt_lib/"libicui18n.55.dylib"
    assert_match /#{icu4c}/, shell_output("otool -L #{bin}/mpd")
  end
```

But everyone involved, me, you, all of the maintainers aside you, etc are going to hate the test. It works, but I don't really want to do it. Going to cc @jacknagel in case he knows of any Ruby magic on this.

xu-cheng: It this a server or similar? If so, can it be tested by `curl`. Also `Process.kill/wait` should be inside a `ensure` block.

DomT4: Yeah, kinda.

I tried testing it with `curl` but inside Homebrew it vomits permission denied messages at me. If I run the `curl` check outside of Homebrew, it works great. Inside the `test do` block? Failure, so much failure.

And yup, my bad on forgetting the `ensure` :smiley_cat:.

DomT4: Turns out I'm an idiot. I forgot to tuck the `curl` check inside a `begin` block and that was causing the failure. This should actually fix Tim's concern earlier as well, because `curl` will fail if the server doesn't respond and that'll blow up the test, so we *should* catch future `icu4c` bumps needed - At least, that's the hope. Pushing now.

Library/Formula/mpd.rb
```diff
@@ -116,4 +119,18 @@ def plist; <<-EOS.undent
...
+    end
+    sleep 2
+
+    begin
+      assert_match /OK MPD/, shell_output("curl localhost:6600")
+    ensure
+      Process.kill("SIGINT", pid)
+      Process.wait(pid)
+    end
+  end
```
xu-cheng: Shouldn't it be:
```ruby
test do
  begin
    pid = fork do
      exec "#{bin}/mpd --stdout --no-daemon --no-config"
    end
    sleep 2
    assert_match /OK MPD/, shell_output("curl localhost:6600")
  ensure
    Process.kill("SIGINT", pid)
    Process.wait(pid)
  end
end
```

DomT4: Seemingly, It can be either and works. Homebrew's internal use tends to prefer `begin` inside the ` fork do`, formulae lean towards preferring the `begin` after the `fork do` :neutral_face:.

DomT4:
Bot decided to crash on this last night. Green now.

mikemcquaid:
Thanks again @DomT4!

PR #39371, opened 2015-05-05
Build LLVM out-of-source

### LINKED ISSUE #37167: llvm --with-lld requires an out-of-source build
As of the past ~4 days, I get the following errors when building LLVM:

```bash
brew install llvm --HEAD --with-lld                                                                                                     [130]
==> Cloning http://llvm.org/git/llvm.git
Updating /Library/Caches/Homebrew/llvm--git
==> Patching
patching file Makefile.rules
Hunk #1 succeeded at 598 (offset -1 lines).
==> Cloning http://llvm.org/git/lld.git
Updating /Library/Caches/Homebrew/llvm--lld--git
==> ./configure --prefix=/usr/local/Cellar/llvm/HEAD --enable-optimized --disable-bindings --enable-targets=host --enable-shared
==> make
llvm[4]: Building Release+Asserts Archive Library liblldCore.a
llvm[4]: Building Release+Asserts Archive Library liblldDriver.a
make[2]: *** [all] Error 1
make[1]: *** [lld/.makeall] Error 2
make: *** [all] Error 
```

Brew doctor is clean, and I can usually build LLVM with `brew install llvm --HEAD --with-all-targets --with-clang --with-lld --with-python` with no problems. Any ideas?

tamird:
Fixes #37167.

@jacknagel

### DIFF
--- a/Library/Formula/llvm.rb
+++ b/Library/Formula/llvm.rb
@@ -69,13 +69,11 @@ class Llvm < Formula
   option "with-lld", "Build LLD linker"
   option "with-lldb", "Build LLDB debugger"
   option "with-rtti", "Build with C++ RTTI"
-  option "with-all-targets", "Build all target backends"
   option "with-python", "Build Python bindings against Homebrew Python"
   option "without-shared", "Don't build LLVM as a shared library"
   option "without-assertions", "Speeds up LLVM, but provides less debug information"
 
   deprecated_option "rtti" => "with-rtti"
-  deprecated_option "all-targets" => "with-all-targets"
   deprecated_option "disable-shared" => "without-shared"
   deprecated_option "disable-assertions" => "without-assertions"
 
@@ -84,6 +82,7 @@ class Llvm < Formula
   else
     depends_on :python => :optional
   end
+  depends_on "cmake" => :build
   depends_on "swig" if build.with? "lldb"
 
   keg_only :provided_by_osx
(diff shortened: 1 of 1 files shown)


Library/Formula/llvm.rb
```diff
@@ -117,26 +116,19 @@ def install
...
+    args << "-DBUILD_SHARED_LIBS=Off" if build.without? "shared"
 
-    args << "--disable-assertions" if build.without? "assertions"
+    args << "-DLLVM_ENABLE_ASSERTIONS=On" if build.with? "assertions"
 
-    system "./configure", *args
-    system "make"
-    system "make", "install"
+    mktemp do
+      system "cmake", "-G", "Unix Makefiles", buildpath, *std_cmake_args, *args
```
tdsmith: `*foo, *bar` doesn't work in Ruby 1.8 (on Mountain Lion); use `*(foo+bar)` or similar.

tamird: Thanks! Fixed.

tdsmith:
:+1:

tamird:
@jacknagel this is green - would you mind having a look?

jacknagel:
What happened to the all-targets option?

jacknagel:
In other words, will this build all targets by default? Shouldn't we default to the host platform as before?

tamird:
the `cmake` equivalent is [`LLVM_TARGETS_TO_BUILD`](http://llvm.org/docs/CMake.html#llvm-specific-variables) and it defaults to `all` and doesn't have a `host` special case.

jacknagel:
What are the possible values? We can give it the host target.

tamird:
https://github.com/llvm-mirror/llvm/blob/master/CMakeLists.txt#L168 - how can we get the host target?

tamird:
Also, from IRC:
> [11:16:12]  \<tdsmith>	tamird: does "all" do a lot more work than whatever the "host" equivalent would be?
> [11:16:27]  \<tamird>	i dont know if it's a lot more, but it's certainly more
> [11:17:28]  \<tdsmith>	hmm it "only" took 30 minutes on the build bot which seems reasonable to me

jacknagel:
Can we see what the difference is if we just build `X86 PowerPC AArch64 ARM`?

tamird:
On my local machine:
with:
```diff
diff --git a/Library/Formula/llvm.rb b/Library/Formula/llvm.rb
index e5e1415..5cd43bf 100644
--- a/Library/Formula/llvm.rb
+++ b/Library/Formula/llvm.rb
@@ -69,11 +69,13 @@ class Llvm < Formula
   option "with-lld", "Build LLD linker"
   option "with-lldb", "Build LLDB debugger"
   option "with-rtti", "Build with C++ RTTI"
+  option "with-all-targets", "Build all target backends"
   option "with-python", "Build Python bindings against Homebrew Python"
   option "without-shared", "Don't build LLVM as a shared library"
   option "without-assertions", "Speeds up LLVM, but provides less debug information"
 
   deprecated_option "rtti" => "with-rtti"
+  deprecated_option "all-targets" => "with-all-targets"
   deprecated_option "disable-shared" => "without-shared"
   deprecated_option "disable-assertions" => "without-assertions"
 
@@ -120,6 +122,10 @@ class Llvm < Formula
       -DLLVM_OPTIMIZED_TABLEGEN=On
     ]
 
+    if build.without? "all-targets"
+      args << "-DLLVM_TARGETS_TO_BUILD=X86;PowerPC;AArch64;ARM"
+    end
+
     args << "-DBUILD_SHARED_LIBS=Off" if build.without? "shared"
 
     args << "-DLLVM_ENABLE_ASSERTIONS=On" if build.with? "assertions"
```
`real	11m27.849s`

without:
`real	13m43.070s`

jacknagel:
OK, let's go with this.

PR #39375, opened 2015-05-05
ocrad 0.25

dunn:


dunn:
Typo in the commit message, woops

xu-cheng:
Thanks @dunn

PR #39376, opened 2015-05-05
xpwn 1.0 (new formula)

b3ll:
This adds xpwn as a forumla to homebrew. xpwn is an invaluable tool (set) to iOS jailbreak developers, as it allows decryption and modification to iOS ipsw files and devices.

### DIFF
--- a/Library/Formula/xpwn.rb
+++ b/Library/Formula/xpwn.rb
@@ -1,0 +1,32 @@
+class Xpwn < Formula
+  homepage "https://github.com/planetbeing/xpwn"
+  url "https://github.com/planetbeing/xpwn.git"
+  sha256 "6c3c00ee92275c0498a52f3e2912add8d84e693acc3edf88e8febaadc8283d61"
+  version "1.0"
+
+  depends_on "cmake" => :build
+  depends_on "openssl"
+  depends_on "libpng"
+  depends_on "libusb"
+  depends_on "lzlib"
+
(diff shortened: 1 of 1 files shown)


Library/Formula/xpwn.rb
```diff
@@ -0,0 +1,32 @@
+class Xpwn < Formula
+  homepage "https://github.com/planetbeing/xpwn"
+  url "https://github.com/planetbeing/xpwn.git"
```
DomT4: Homebrew/homebrew doesn't accept `.git` URLs as the main download. It's impossible to bottle, notice when updated, etc. If there's no stable download this will need to go in ` Homebrew/head-only`.

b3ll: I can fork / tag it with a release if that makes things easier...

DomT4: Homebrew/homebrew doesn't accept unblessed forks either, alas. If you fork it, *and* upstream bless the new fork as the official place to go for support/releases/etc that's cool - But otherwise, it'll need to come from the currently existing upstream repo & maintainer.

DomT4:
Any update on this re getting upstream to tag a release, etc?

DomT4:
Closing here. If you're still interested feel free to file it at Homebrew/head-only where it may be reconsidered, or host it in a tap of your own. Thanks!

Issue #39362, opened 2015-05-04
brew install apr fails


paxos1977:
> Praxis:~ austirg$ brew install apr
> ==> Downloading http://www.apache.org/dyn/closer.cgi?path=apr/apr-1.5.1.tar.bz2
> ==> Best Mirror http://apache.mirrors.tds.net/apr/apr-1.5.1.tar.bz2

> curl: (22) The requested URL returned error: 404 Not Found
> Error: Failed to download resource "apr"
> Download failed: http://apache.mirrors.tds.net/apr/apr-1.5.1.tar.bz2

Pulling up the apache mirror in the browser shows the problem is 1.5.2 is out and 1.5.1 is not present on the mirror.

tdsmith:
Thanks for the report! Could you try and open a pull request to bump the apr version? [This guide](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged) should help and I'll be happy to walk you through any questions you run in to, either here or on IRC.

tenshibe:
I'm afraid this broke httpd24 for me. Removing httpd24 and reinstalling, some items still look in 1.5.1 directory for static libraries. Have removed apr, apr-utils, reinstalled them and then tried to install httpd24, build failed with "Cannot find the library `/usr/local/Cellar/apr/1.5.1/libexec/lib/libapr-1.la" Guess this is an issue in httpd24

tenshibe:
Removing and reinstalling apr-util with --build-from-source fixed this for me, maybe a new bottle of apr-util built against apr should be provided as well ?

DomT4:
Confirmed. Fixed in #39557.

PR #39379, opened 2015-05-05
apr 1.5.2

vszakats:
Fixes https://github.com/Homebrew/homebrew/issues/39362

### DIFF
--- a/Library/Formula/apr.rb
+++ b/Library/Formula/apr.rb
@@ -16,7 +16,7 @@ class Apr < Formula
   def install
     ENV.universal_binary if build.universal?
 
-    # https://issues.apache.org/bugzilla/show_bug.cgi?id=57359
+    # https://bz.apache.org/bugzilla/show_bug.cgi?id=57359
     # The internal libtool throws an enormous strop if we don't do...
     ENV.deparallelize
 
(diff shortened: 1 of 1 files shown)


vszakats:
```
W0505 03:17:36.864459 1888727824 logging.cpp:81] RAW: Received signal SIGTERM from process 64468 of user 502; exiting
Error: mesos: failed
Will not overwrite /private/tmp/mesos20150505-64468-hd8g4f/Library/Python/2.7/lib/python/site-packages/homebrew.pth
```

DomT4:
@tdsmith Any idea on the failure here? I thought it may just be a random bot hiccup, but seems to have reproduced. Doesn't *look* like one of those issues where a revision is needed either.

tdsmith:
Oh, the python chunk of the mesos test is obsolete, now; I pushed a change a while back that does that automatically for all tests. Fix in #39402 and then we can give this another spin.

DomT4:
Thanks Tim, appreciate the ninja-quick fix! :bow:

vszakats:
https://github.com/Homebrew/homebrew/pull/39402 fixed this PR indeed. Thanks.

xu-cheng:
Thanks @vszakats

PR #39380, opened 2015-05-05
Jython 2.7.0

danslapman:


Library/Formula/jython.rb
```diff
@@ -1,7 +1,7 @@
 class Jython < Formula
   homepage "http://www.jython.org"
-  url "https://search.maven.org/remotecontent?filepath=org/python/jython-installer/2.5.3/jython-installer-2.5.3.jar"
-  sha256 "05405966cdfa57abc8e705dd6aab92b8240097ce709fb916c8a0dbcaa491f99e"
+  url "https://search.maven.org/remotecontent?filepath=org/python/jython-installer/2.7.0/jython-installer-2.7.0.jar"
+  sha1 "b08d73b5df7d95e35e7b8dcaf1558025cf24f0f1"
```
xu-cheng: Please use `sha256`.

tdsmith:
Please also remove the `devel` block since it's now stale.

Library/Formula/jython.rb
danslapman: search.maven.org does not offer sha256 any more :(

jacknagel: We still need to use it, you just need to confirm the output of `brew fetch jython` yourself.

danslapman: Done.

tdsmith:
The `inreplace` in the `install` method is failing; it looks like it can be deleted?

danslapman:
Okay, done...

tdsmith:
Thanks for the PR!

PR #39381, opened 2015-05-05
pcre1: add alias for pcre

vszakats:
`pcre1` is the new canonical name after the release of `pcre2`.


dunn:
I don't think aliases need the `.rb` extension.

vszakats:
Corrected, apologize (it's my first).

tdsmith:
This should be a symlink, not a literal file. Do you have a reference for this? If you could squash it to a single commit that'd be great.

vszakats:
I was hoping it is not a symlink. I'll redo this outside the GitHub web UI.

vszakats:
Redone in https://github.com/Homebrew/homebrew/issues/39399

PR #39386, opened 2015-05-05
Depend on: `prefix` -> `opt_prefix`

tamird:


jacknagel:
Use `opt_prefix` instead

Library/Formula/gearman.rb
```diff
@@ -107,7 +107,7 @@ def plist; <<-EOS.undent
 +++ b/libgearman-1.0/gearman.h
 @@ -50,7 +50,11 @@
  #endif
- 
+
```
jacknagel: don't change whitespace in patches, please

tamird: it was in a separate commit - still no good?

jacknagel: whitespace is significant in patches (the first column is always either a space, a `+`, or a `-`, and trailing whitespace is present in the source lines which provide context for the patch)

tamird: oh, that's what you mean. understood.

tamird:
Changed to `opt_prefix`. Thanks!

PR #39391, opened 2015-05-05
Update to the latest release of F#

7sharp9:
There are various fixes for the latest mono in this revision which not does not include .Net2.0, 3.5, 4.0 reference assemblies


tdsmith:
Thanks for the pull request! 🎉 Homebrew depends on contributions from community members like you and we're grateful for your support. Just for future reference, our favorite commit message format is like `packagename 1.2.3` for straightforward version updates.

PR #39410, opened 2015-05-05
nss 3.19

l-hedgehog:


vszakats:
It'd be nice to update this old link in [line 71](https://github.com/Homebrew/homebrew/pull/39410/files#diff-7c58edb1cb75354b1a006f33332e43e6R71):
http://www.mozilla.org/projects/security/pki/nss/tools/certutil.html
with this one:
https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Tools/certutil

xu-cheng:

PR #39411, opened 2015-05-05
sgrep: pass strict audit with test, add SSL/TLS

vszakats:


Library/Formula/sgrep.rb
```diff
@@ -1,14 +1,19 @@
...
     system "./configure", "--prefix=#{prefix}",
                           "--mandir=#{man}",
                           "--datadir=#{share}/sgrep"
-    system "make install"
+    system "make", "install"
+  end
+
+  test do
+    input = test_fixtures("test.eps")
+    output = `#{bin}/sgrep -c '"mark"' #{input}`
```
xu-cheng: Please use `shell_output`. Thanks.

vszakats: Updated.

PR #39412, opened 2015-05-05
sha2: SSL/TLS URLs

vszakats:


DomT4:
May as well bump things to `sha256` when you change the URL/Mirror links. Two birds with one stone, etc.

vszakats:
Decided to leave those intact when updating download URLs to ensure the content is unchanged. If this is not a concern, I'm happy to update them in one pass.

DomT4:
Verify the SHA1 locally beforehand, but otherwise, seems fine to me at least. Maybe we could adjust the test-bot to verbosely print the checksum during the `fetch` stage of each PR, but I suspect that may run into some maintainer hesitation.

vszakats:
Made a small update in my local scripts so that I can verify it. For others to trust such updates, some bot magic would certainly be nice.

tdsmith:
I make a point of fetching them and running both checksums to make sure they match.

vszakats:
That is what I do. Part of my concern was that it's only the updater who can see the match.

DomT4:
It's perhaps worth noting that via the magic of `brew pull` the maintainers get both the SHA1 and SHA256 flashed at them, so a quick glance back at the PR can show whether the SHA1 changed from before.

vszakats:
Thanks for the info, it's enough for me to feel comfortable doing combined updates.

tdsmith:
Only bottles are fetched during `pull`.

DomT4:
Oh, duh. Of course. I automatically type in ` brew fetch xyz -fs ` afterwards and check, so apparently my brain started to presume that was done automatically. Sorry :sweat_smile:.

DomT4:
Might tee up a PR so the `-fs` download is done if a bottle isn't pulled/updated. It might get shot down, but no harm in asking.

PR #39419, opened 2015-05-05
ant: use SSL/TLS, sha256

vszakats:


Library/Formula/ant.rb
```diff
@@ -1,7 +1,7 @@
 class Ant < Formula
-  homepage "http://ant.apache.org/"
-  url "http://www.apache.org/dyn/closer.cgi?path=ant/binaries/apache-ant-1.9.4-bin.tar.gz"
-  sha1 "6c41481e8201f6b3f7e216146b95bb6de70208bb"
+  homepage "https://ant.apache.org/"
+  url "https://www.apache.org/dyn/closer.cgi?path=ant/binaries/apache-ant-1.9.4-bin.tar.bz2"
```
tdsmith: Don't change file type if you can, please.

vszakats: Any reason for this? There is a mixture of `.bz2` and `.gz` used (mostly the latter) at the moment for `apache.org` downloads. I took the liberty to update to `.bz2` because it's smaller and seems preferred in other formulae.

tdsmith: There's no institutional preference. It's harder to verify the hash hasn't really changed, if nothing else.

vszakats: Thanks, noted!

PR #39420, opened 2015-05-06
pulseaudio 6.0

dunn:
Apparently no movement upstream on the i386 issue, since MacPorts updated their patch.

### DIFF
--- a/Library/Formula/pulseaudio.rb
+++ b/Library/Formula/pulseaudio.rb
@@ -1,21 +1,29 @@
-require 'formula'
-
 class Pulseaudio < Formula
   homepage "http://pulseaudio.org"
-  url "http://freedesktop.org/software/pulseaudio/releases/pulseaudio-5.0.tar.xz"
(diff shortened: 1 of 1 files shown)


Library/Formula/pulseaudio.rb
```diff
@@ -1,21 +1,30 @@
...
 
   depends_on "pkg-config" => :build
   depends_on "libtool" => :build
-  depends_on "intltool" => :build if build.with? "nls"
-  depends_on "gettext" => :build if build.with? "nls"
+
+  if build.with?("nls") || build.head?
+    depends_on "intltool" => :build
+    depends_on "gettext" => :build
+  end
```
jacknagel: Repeat these deps inside the `head` block rather than the `|| build.head?`. Going forward, any use of `build.head?`, `build.devel?`, or `build.stable?` in the class body (i.e. in the DSL section, not the install method) is to be avoided.

dunn: Got it.

tdsmith:
Thanks!

PR #39423, opened 2015-05-06
dtc: use ssl/tls links

DomT4:
Follow up from #39338.


xu-cheng:
Thanks.

DomT4:
Thanks Xu, and thanks for spotting the version had gone weird on this one - No idea how I didn't notice that.

PR #39425, opened 2015-05-06
pull: fetch source checksum if no bottle added/updated

DomT4:
The thinking here is that it's useful to quickly glance & check on SHA1 => SHA256 updates (nested within URL updates, plaintext to SSL/TLS updates, etc) that the SHA1 hasn't changed, which the SHA256 change would otherwise mask.

It's also potentially useful to check the expected checksum matches on maintainer local machines prior to push on the things we don't bottle.

Opt-in for now, because the change has the potential to catch people who aren't watching the pull file for changes like a hawk off-guard.

Re discussion in #39412.

### DIFF
--- a/Library/Homebrew/cmd/pull.rb
+++ b/Library/Homebrew/cmd/pull.rb
@@ -181,6 +181,9 @@ module Homebrew
         else
           opoo "You must set BINTRAY_USER and BINTRAY_KEY to add or update bottles on Bintray!"
         end
+      else
+        formula = changed_formulae.first
+        safe_system "brew", "fetch", "--retry", "-fs", formula.name if ARGV.include? "--check-sha"
       end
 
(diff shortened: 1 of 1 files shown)


DomT4:
Not *hugely* emotionally attached to the idea. Could be useful, might not be. It was discussed in another PR, and I think everyone prefers PRs over Issues as discussion points, so here is one.

Shoot it down if it's not useful.

tdsmith:
I wouldn't find this useful.

DomT4:
Heh, from your comments in the other thread I thought you'd be pretty much the only maintainer on board with the idea, to be honest. Closing.

mikemcquaid:
My reason for not needing this is the bot will handle it for us (which it can't with the bottle checksums). Nice idea, though!

PR #37260, opened 2015-02-27
llvm 3.6.0

DomT4:
New LLVM release.

PR #39428, opened 2015-05-06
Fix `creduce --HEAD`

tamird:
When installing HEAD, `bin` doesn't refer to the correct location.

### DIFF
--- a/Library/Formula/creduce.rb
+++ b/Library/Formula/creduce.rb
@@ -94,6 +94,6 @@ class Creduce < Formula
     EOS
 
     chmod 0755, testpath/"test1.sh"
-    system "#{bin}/creduce", "test1.sh", "test1.c"
+    system "creduce", "test1.sh", "test1.c"
   end
 end


DomT4:
Where is it installed instead?

tamird:
`$(brew --prefix)/Cellar/creduce/HEAD/bin/creduce` rather than `$(brew --prefix)/Cellar/creduce/2.2.1/bin/creduce`

tamird:
The build failures here are due to upstream updates that broke creduce compatibility with LLVM 3.5 in favour of 3.6. When upstream cuts a new release we will need to update creduce and LLVM in lockstep in #37260

DomT4:
> $(brew --prefix)/Cellar/creduce/HEAD/bin/creduce rather than $(brew --prefix)/Cellar/creduce/2.2.1/bin/creduce

Did you run `brew test --verbose --HEAD ` rather than `brew test --verbose`?

tamird:
I didn't - and didn't know that was required. That does work!

DomT4:
Great! :smiley_cat:.

I believe it's necessary because `brew test` defines `f.path` as in cellar path rather than `opt` path. I could be wildly wrong on that interpretation of the logic though.

PR #39429, opened 2015-05-06
Allow HOMEBREW_CELLAR environment variable to override cellar path

L2G:
I made this change because I wanted to be able to have a development copy of Homebrew checked out while still having it use my regular system cellar (outside of the checked-out repo).

Perhaps there is a better way to do this that doesn't involve making this change.

### DIFF
--- a/Library/Homebrew/config.rb
+++ b/Library/Homebrew/config.rb
@@ -39,9 +39,13 @@ HOMEBREW_REPOSITORY = Pathname.new(HOMEBREW_BREW_FILE).realpath.dirname.parent
 HOMEBREW_LIBRARY = HOMEBREW_REPOSITORY/"Library"
 HOMEBREW_CONTRIB = HOMEBREW_REPOSITORY/"Library/Contributions"
 
-# Where we store built products; /usr/local/Cellar if it exists,
-# otherwise a Cellar relative to the Repository.
-HOMEBREW_CELLAR = if (HOMEBREW_PREFIX+"Cellar").exist?
+# Where we store built products; in decreasing order of precedence:
+#   * value of ENV["HOMEBREW_CELLAR"], if set
+#   * the Cellar subdirectory relative to Homebrew, if it exists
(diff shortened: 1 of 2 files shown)


xu-cheng:
Personally I'm :-1: for this change. Lots of existing tools relies on current `HOMEBREW_CELLAR` location. e.g. The uninstall script. At the same time, such customization would increase the difficult for future maintaining.

---

As for your scenario. Can you just symlink your dev Cellar to system Cellar?

tdsmith:
:-1:; just invoke the other `brew` script and it will install relative to its location on disk.

tdsmith:
Nevermind, misread!

L2G:
Yes, I just prefer this way over making a symlink every time I make a checkout. I don't understand what this would be breaking.

L2G:
Would it be an acceptable alternative if the value of `HOMEBREW_CELLAR` could simply default to `/usr/local/Cellar` unless the local repo has its own Cellar subdirectory? Because my reading of the docs and the source suggests that this should be the case already.

tdsmith:
No; the Cellar is positioned relative to the repository and is created as necessary. https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Installation.md#untar-anywhere-and-then-symlink-the-brew-command-elsewhere

I suggest taking advantage of git branches in your /usr/local repository.

L2G:
Is it ever possible for `HOMEBREW_PREFIX` and `HOMEBREW_REPOSITORY` to be different?

If so, when?

tdsmith:
Yes, if you clone the Homebrew repository to `/a` and link `/a/bin/brew` to `/b/bin/brew` and put `/b/bin` in PATH, `HOMEBREW_REPOSITORY` is `/a` and `HOMEBREW_PREFIX` is `/b`.

L2G:
OK. I can verify this to be correct when I have created a setup as you describe, and I change to `/a`, and I run `bin/brew` (not just `brew`).

PR #39431, opened 2015-05-06
gnustep-make 2.6.6 (also changes for brew audit --strict)

L2G:


Library/Formula/gnustep-make.rb
```diff
@@ -1,14 +1,16 @@
...
   def install
     system "./configure", "--prefix=#{prefix}",
                           "--with-config-file=#{prefix}/etc/GNUstep.conf",
                           "--enable-native-objc-exceptions"
-    system "make install"
+    system "make", "install"
+  end
+
+  test do
+    system "false"
```
dunn: Was this left over from `brew create`?

xu-cheng:
Thanks @L2G

PR #39366, opened 2015-05-04
bullet: 2.83

scpeters:
It was just released.

https://github.com/bulletphysics/bullet3/releases/tag/2.83

PR #39436, opened 2015-05-06
bullet: 2.83

scpeters:
The cmake variable for building demos changed
from BUILD_DEMOS to BUILD_BULLET2_DEMOS.
The demos are now in the examples folder
instead of the Demos folder.

Also, demos are disabled for 10.8 and less
or if shared libraries are specified.

My first attempt was #39366. The CI jobs helped me find out about the incompatibility of demos with 10.8 ([failed job](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/25445/version=mountain_lion/console) which I [reported upstream](https://github.com/bulletphysics/bullet3/issues/371)). Also, if you try to install `--with-demo` and `--with-shared` the demos won't work (see [console warnings here](https://gist.github.com/scpeters/6afc44f0cf916b11a226) if you are so inclined). The demos do work with static libraries.

### DIFF
--- a/Library/Formula/bullet.rb
+++ b/Library/Formula/bullet.rb
@@ -1,7 +1,7 @@
 class Bullet < Formula
   homepage "http://bulletphysics.org/wordpress/"
-  url "https://github.com/bulletphysics/bullet3/archive/2.82.tar.gz"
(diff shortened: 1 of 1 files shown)


scpeters:
Upstream is going to fix the 10.8 compatibility issue and release 2.83.1 shortly.

PR #39449, opened 2015-05-06
OpenMPI: 1.8.5.

dpo:
Stylistic changes and caveats about compiler flags.

v1.8.5 is a bug fix release.


xu-cheng:

PR #39450, opened 2015-05-06
rethinkdb: explicitly set cache size to auto

dalanmiller:
@coffeemug @AtnNn 

Cache size on my MBA was being automatically set without this flag to just 100MB and with the `auto` flag it goes up to 1GB which seems to perform better. Disregard if this is intentionally left out.


jacknagel:
Can you ask upstream about this directly (i.e., on their issue tracker)?

dalanmiller:
@jacknagel done! - https://github.com/rethinkdb/rethinkdb/issues/4310

PR #39458, opened 2015-05-07
bullet: 2.83.1

scpeters:
The cmake variable for building demos changed
from BUILD_DEMOS to BUILD_BULLET2_DEMOS.
The demos are now in the examples folder
instead of the Demos folder.

Also, demos are disabled if shared libraries
are specified due to a linking problem.

Previous attempts: #39366, #39436.


scpeters:
Sorry for all the noise. I keep finding problems with the upstream release.

PR #39461, opened 2015-05-07
teapot: patch in correct version number; install LyX doc file

L2G:


Library/Formula/teapot.rb
```diff
@@ -1,17 +1,23 @@
...
+      system "cmake", *args
+      system "make", "install"
     end
   end
+
+  def caveats; <<-EOS.undent
+    Teapot documentation does not get built because it requires software (LyX)
+    not included in Homebrew. For a good introduction to Teapot, see:
+
+        https://kmandla.wordpress.com/2010/08/11/how-to-use-teapot-like-a-pro/
```
xu-cheng: I wonder whether it's good to reference to an unofficial document?

L2G: It's either that or reference no document.

xu-cheng: Let's drop this (I will kill this when merging).

IMHO, caveats should only include Homebrew specific information. Users should find general information such as document on their own. And it's pretty bad to include unofficial third party document.

xu-cheng: Wait, so I just found out this PR only contains style only changed. Can you remove the caveats and add a test? Thanks.

Library/Formula/teapot.rb
```diff
@@ -1,17 +1,34 @@
...
+--- a/config.h
++++ b/config.h
+@@ -1,7 +1,7 @@
+ /* configuration values */
+
+
+-#define VERSION "2.2.1"
++#define VERSION "2.3.0"
+
+ #define HELPFILE "/usr/local/share/doc/teapot/html/index.html"
```
DomT4: This needs to be made prefix-agnostic. You could always keep the patch and then `inreplace` this part of the patch to facilitate different `HOMEBREW_PREFIX`s. As Xu said, a test would be :heart: as well.

DomT4:
@BrewTestBot test this please

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/6188a626f2f311f457b6467dc101af66ef6150e7. Thanks @L2G! :cocktail:

Ideally tests do need adding, but I couldn't see anything particularly obvious on how to test this - The documentation is a little vague.

PR #39463, opened 2015-05-07
git-latexdiff: migrate to homebrew/tex tap

L2G:


xu-cheng:

PR #39464, opened 2015-05-07
pius 2.1.1

dunn:


Library/Formula/pius.rb
```diff
@@ -1,22 +1,24 @@
...
 
   def install
-    # Replace hardcoded gpg path: https://sourceforge.net/p/pgpius/bugs/12/
-    inreplace "pius", "/usr/bin/gpg", HOMEBREW_PREFIX/"bin/gpg"
-    bin.install "pius"
-    bin.install "pius-keyring-mgr"
-    bin.install "pius-party-worksheet"
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    # Replace hardcoded gpg path (WONTFIX):
+    # https://sourceforge.net/p/pgpius/bugs/12/
```
asparagui: Thanks for this link, makes keeping track of these things much easier.

dunn: The link was already there, I just added that it wouldn't be fixed.

PR #39468, opened 2015-05-07
pgbadger 6.4

dunn:


asparagui:
+1 on adding a test, -1 on downloading random files to make said test work.  Will defer to somebody else's opinion.

xu-cheng:
Can you just create a smallest log file using `(testpath/"log").write ...`

dunn:
Obvious in retrospect. Thanks!

xu-cheng:
Thanks.

PR #39470, opened 2015-05-07
compilers: support gcc 5

ntkme:
This PR adds gcc 5 to `--cc` option.


Library/Homebrew/compilers.rb
```diff
@@ -1,6 +1,6 @@
 module CompilerConstants
-  GNU_GCC_VERSIONS = 3..9
-  GNU_GCC_REGEXP = /^gcc-(4\.[3-9])$/
+  GNU_GCC_VERSIONS = [4.3, 4.4, 4.5, 4.6, 4.7, 4.8, 4.9, 5, 5.1]
```
jacknagel: These should be strings, we don't want to use floating-point numbers for versions.

jacknagel:
Do we need to add `5.1`? My understanding of the new versioning scheme is that `5` is more like `4.9` and `5.1` is more like `4.9.1`.

mistydemeo:
We're not creating a `gcc51` package or installing under a `gcc-5.1` name - all series 5 GCCs will be `gcc5`.

PR #39485, opened 2015-05-07
audit: enforce SSL/TLS MetaCPAN urls

vszakats:
It is "real" SSL/TLS, it doesn't redirect to plaintext.


vszakats:
Restarting. Not being confident to rebase this.

PR #39486, opened 2015-05-07
audit: Check if a 'url' value is reused as 'mirror'

vszakats:
Please verify me.

Example for this kind of problem:
https://github.com/Homebrew/homebrew/blob/5cda0af3b97ecdd78412952ab161eb85a1aca8be/Library/Formula/signing-party.rb#L44-L45

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -321,6 +321,11 @@ class FormulaAuditor
       problem "\"http://ftpmirror.gnu.org\" is preferred for GNU software (url is #{u})."
     end
 
+    # Check if main url is used again as a mirror
+    if (urls & :mirrors).length
+      problem "Url should not be reused as a mirror: #{(urls & :mirrors)[0]}"
+    end
+
     # the rest of the checks apply to mirrors as well.
     urls.concat(@specs.map(&:mirrors).flatten)
 


chdiza:
I tried this, and it doesn't even emit a warning about `signing-party.rb`.  Perhaps I did something wrong.

Library/Homebrew/cmd/audit.rb
```diff
@@ -321,6 +321,11 @@ def audit_urls
       problem "\"http://ftpmirror.gnu.org\" is preferred for GNU software (url is #{u})."
     end
 
+    # Check if main url is used again as a mirror
+    if (urls & :mirrors).length
```
jacknagel: `:mirrors` is a symbol, not an array. This line will raise a TypeError. Ignoring that, this condition would always be true, because `Array#length` always returns a value.

jacknagel:
This looks like guesswork, which just creates more work for reviewers. We would appreciate if code changes were tested locally before being submitted. Thanks!

jacknagel:
I pushed a working version, but note that the URL audits are not applied to resources at this time (not on purpose, but because nobody has ported it over yet).

vszakats:
Everyone was faster. I could since start an integrated test and come up with this:
```
    # Check if there are any duplicate urls
    if urls.uniq.length != urls.length
      problem "Each URL should appear only once as url or mirror: #{urls.select{ |e| urls.count(e) > 1 }.uniq}"
    end
```
Sorry for the lousy first attempt.

PR #39489, opened 2015-05-07
librdkafka

redmar:
librdkafka

PR #39490, opened 2015-05-07
kafkacat

redmar:
kafkacat


redmar:
depends on #39489

tdsmith:
Thanks!

PR #39491, opened 2015-05-07
pdal 0.9.9

dunn:


dunn:
`ENV.cxx11` solved `'cstdint' file not found` and `'array' file not found` but now we've got new errors.

Library/Formula/pdal.rb
```diff
@@ -1,17 +1,22 @@
...
+  url "https://github.com/PDAL/PDAL/archive/0.9.9.tar.gz"
+  sha256 "d4f91478ca55b6b775980c5c2e4c23f43b6bb4e1908ae739b1605a30b57b8a83"
 
-  head 'https://github.com/PDAL/PDAL.git'
+  head "https://github.com/PDAL/PDAL.git"
 
-  depends_on 'cmake' => :build
-  depends_on 'boost'
+  depends_on "cmake" => :build
+  depends_on "boost"
```
tdsmith: probably wants `=> "c++11"` for systems older than Mavericks

xu-cheng:
Thanks @dunn

PR #39496, opened 2015-05-07
fmpp: FreeMarker-based file PreProcessor

onlynone:


Library/Formula/fmpp.rb
```diff
@@ -0,0 +1,16 @@
+class Fmpp < Formula
+  homepage "http://fmpp.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/fmpp/fmpp/0.9.15/fmpp_0.9.15.tar.gz"
+  sha256 "b893451b5450a7f35fe680e934f6903ec8143d88959dcfca5d17467fbe4142f9"
```
xu-cheng: add `head "https://github.com/freemarker/fmpp.git"` below this line

Library/Formula/fmpp.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  depends_on :java
+
+  def install
+    libexec.install "lib"
+    bin.write_jar_script libexec/"lib/fmpp.jar", "fmpp", "-Dfmpp.home=#{libexec} $FMPP_OPTS $FMPP_ARGS"
+  end
+
+  test do
+    system "fmpp", "-h"
```
xu-cheng: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

DomT4:
Please address Xu's comments here, Thanks!

DomT4:
Any update here?

Library/Formula/fmpp.rb
onlynone: It's just annoying because `fmpp` doesn't take stdin/stdout. You have to pass it a filename for the input source, and a filename for the output destination. This works from the shell, but I'm not familiar with ruby and the homebrew build environment enough to implement it properly:

    $ echo '<#assign foo="bar"/><p>${foo}</p>' > input
    $ fmpp input -o output
    *** DONE ***
    
    $ cat output
    <p>bar</p>

mikemcquaid:
```ruby
(testpath/"input").write '<#assign foo="bar"/><p>${foo}</p>'
system "fmpp", "input", "-o", "output"
assert File.exist? "output"
```
or similar

DomT4:
It'd be nice to improve the test here per Mike's suggestion.

DomT4:
@BrewTestBot test this please

DomT4:
@BrewTestBot test this please

DomT4:
Last couple failures.

```
==> audit problems
fmpp:
 * `desc`(line 4) should be put before `homepage`(line 2)
 * Description is too long. "name: desc" should be less than 80 characters (currently 92).
```

DomT4:
Any update?

onlynone:
Can this be merged?

Library/Formula/fmpp.rb
```diff
@@ -0,0 +1,21 @@
+class Fmpp < Formula
+  desc "FMPP is a text file preprocessing tool using FreeMarker templates."
+  homepage "http://fmpp.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/fmpp/fmpp/0.9.15/fmpp_0.9.15.tar.gz"
+  sha256 "b893451b5450a7f35fe680e934f6903ec8143d88959dcfca5d17467fbe4142f9"
+  head "https://github.com/freemarker/fmpp.git"
+
+  depends_on :java
```
DomT4: Any specific version of Java?

onlynone: According to their website, version 1.4 or newer. I'm not sure how to specify that.

DomT4: I'm not sure it's worth specifying a Java minimum dependency that early; you'd have to go out of your way to install Java 1.4 these days. Thoughts @xu-cheng?

xu-cheng: We can leave `depend_on :java` as it is.

dunn:
Thanks so much! Merged in 1189471b32abe12918dacd3c9bd0007b80220190.

PR #39500, opened 2015-05-07
mosquitto 1.4.2

ralight:


Library/Formula/mosquitto.rb
```diff
@@ -35,11 +36,6 @@ def caveats; <<-EOD.undent
     mosquitto has been installed with a default configuration file.
     You can make changes to the configuration by editing:
         #{etc}/mosquitto/mosquitto.conf
-
-    Python client bindings can be installed from the Python Package Index:
-        pip install mosquitto
-
-    Javascript client has been removed, see Eclipse Paho for an alternative.
```
xu-cheng: Why remove this?

Library/Formula/mosquitto.rb
```diff
@@ -14,12 +14,13 @@ class Mosquitto < Formula
...
   depends_on "cmake" => :build
   depends_on "c-ares"
+  depends_on "libwebsockets"
 
   # mosquitto requires OpenSSL >=1.0 for TLS support
   depends_on "openssl"
 
   def install
-    system "cmake", ".", *std_cmake_args
+    system "cmake", "-DWITH_WEBSOCKETS=ON", ".", *std_cmake_args
```
xu-cheng: Should it make libwebsockets recommended or optional?

Library/Formula/mosquitto.rb
ralight: The Python client is no longer part of mosquitto and the Javascript client is long gone.

Library/Formula/mosquitto.rb
databus23: The support for websockets in mosquitto is very useful in my opinion and its arguably a core feature of version 1.4. So +1 for recommended instead of optional if it should be configurable.

ralight:
I've now set libwebsockets to recommended and made it configurable.

Library/Formula/mosquitto.rb
```diff
@@ -14,12 +14,16 @@ class Mosquitto < Formula
...
 
   # mosquitto requires OpenSSL >=1.0 for TLS support
   depends_on "openssl"
 
   def install
-    system "cmake", ".", *std_cmake_args
+    args = []
+    args << "-DWITH_WEBSOCKETS=ON" if build.with? "libwebsockets"
+
+    system "cmake", args, ".", *std_cmake_args
```
xu-cheng: Let make it like this:
```ruby
args = std_cmake_args
args << ...

system "cmake", ".", *args
```

ralight: Ah yes, that's much better. Thanks for the hint.

tdsmith:
Thanks for the pull request! 🎉 Homebrew depends on contributions from community members like you and we're grateful for your support. Just for future reference, we prefer sha256 hashes over sha1 hashes.

PR #39501, opened 2015-05-07
docker 1.6.1

alex:


xu-cheng:
Need to split the PR. Looks like there is a dependency loop.

tdsmith:
The bottles build fine; gonna pull. Thanks!

PR #39517, opened 2015-05-08
karn 0.0.2 (new formula)

prydonius:
as requested in prydonius/karn#8

### DIFF
--- a/Library/Formula/karn.rb
+++ b/Library/Formula/karn.rb
@@ -1,0 +1,47 @@
+require "language/go"
+
+class Karn < Formula
+  homepage "https://github.com/prydonius/karn"
+  url "https://github.com/prydonius/karn/archive/v0.0.2.tar.gz"
+  sha256 "eac77b82f7abd9c37e70d01240e8e69893b7497d3dd5ab7b0fb2df04ae1e6529"
+
+  depends_on "go" => :build
+
+  go_resource "github.com/codegangsta/cli" do
+    url "https://github.com/codegangsta/cli.git",
+        :revision => "a889873af50a499d060097216dcdbcc26ed09e7c"
+  end
+
+  go_resource "github.com/mitchellh/go-homedir" do
+    url "https://github.com/mitchellh/go-homedir.git",
+        :revision => "1f6da4a72e57d4e7edd4a7295a585e0a3999a2d4"
+  end
+
+  go_resource "gopkg.in/yaml.v2" do
+    url "https://gopkg.in/yaml.v2",
+        :revision => "49c95bdc21843256fb6c4e0d370a05f24a0bf213", :using => :git
+  end
+
+  def install
+    ENV["GOPATH"] = buildpath
+    mkdir_p buildpath/"src/github.com/prydonius"
+    ln_s buildpath, buildpath/"src/github.com/prydonius/karn"
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    system "go", "build", "cmd/karn/karn.go"
+    bin.install "karn"
+  end
+
+  test do
+    (testpath/".karn.yml").write <<-EOS.undent
+      ---
+      #{testpath}:
+        name: Homebrew Test
+        email: test@brew.sh
+    EOS
+    system "git", "init"
+    system "git", "config", "--global", "user.name", "Test"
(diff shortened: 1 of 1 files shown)


Library/Formula/karn.rb
```diff
@@ -0,0 +1,38 @@
...
+
+class Karn < Formula
+  homepage "https://github.com/prydonius/karn"
+  url "https://github.com/prydonius/karn/archive/v0.0.2.tar.gz"
+  sha256 "eac77b82f7abd9c37e70d01240e8e69893b7497d3dd5ab7b0fb2df04ae1e6529"
+
+  depends_on "go" => :build
+
+  go_resource "github.com/codegangsta/cli" do
+    url "https://github.com/codegangsta/cli",
```
dunn: If you change this to `https://github.com/codegangsta/cli.git` (and similarly for the next one) you won't need to specify `:using => :git`; it'll be able to figure it out.

prydonius:
Thanks for the tip @dunn

prydonius:
Any update on this?

chdiza:
This formula is unlikely to be accepted, because it is self-submitted, for a project that's only six weeks old, and is at version 0.0.2.  Homebrew [explicitly says](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Acceptable-Formulae.md#niche-or-self-submitted-stuff) it doesn't want such things.

chdiza:
Note that I'm not a Homebrew maintainer, and that I don't particularly like the policy that I linked to.

Library/Formula/karn.rb
```diff
@@ -0,0 +1,38 @@
...
+    mkdir_p buildpath/"src/github.com/prydonius"
+    ln_s buildpath, buildpath/"src/github.com/prydonius/karn"
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    system "go", "build", "cmd/karn/karn.go"
+    bin.install "karn"
+  end
+
+  test do
+    system "#{bin}/karn", "--version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikemcquaid:
> This formula is unlikely to be accepted, because it is self-submitted, for a project that's only six weeks old, and is at version 0.0.2. Homebrew explicitly says it doesn't want such things.

I'm actually OK with this given it's relatively popular and, being a Go project, this makes it a bit easier to install.

Library/Formula/karn.rb
prydonius: Thanks for the suggestion, I've updated the test :)

Library/Formula/karn.rb
```diff
@@ -0,0 +1,49 @@
...
+  test do
+    File.open(File.join(ENV["HOME"], ".karn.yml"), "w") do |f|
+      f.puts <<-EOS.undent
+        ---
+        #{testpath}:
+          name: Homebrew Test
+          email: test@brew.sh
+      EOS
+    end
+    system "git", "init"
```
prydonius: Is this the correct way of calling the git command inside of the a test environment? It seems to work on my local machine.

Library/Formula/karn.rb
```diff
@@ -0,0 +1,49 @@
...
+    mkdir_p buildpath/"src/github.com/prydonius"
+    ln_s buildpath, buildpath/"src/github.com/prydonius/karn"
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    system "go", "build", "cmd/karn/karn.go"
+    bin.install "karn"
+  end
+
+  test do
+    File.open(File.join(ENV["HOME"], ".karn.yml"), "w") do |f|
```
mikemcquaid: You can just use `(testpath/".karn.yml").write <<-EOS.undent` here.

Library/Formula/karn.rb
mikemcquaid: Yup.

Library/Formula/karn.rb
prydonius: Thanks, didn't realise HOME was set to testpath!

mikemcquaid:

PR #39519, opened 2015-05-08
mu: use EmacsRequirement

dunn:


Library/Formula/mu.rb
```diff
@@ -32,7 +21,7 @@ class Mu < Formula
   depends_on "glib"
   depends_on "gmime"
   depends_on "xapian"
-  depends_on Emacs23Installed if build.with? "emacs"
+  depends_on :emacs => "23" if build.with? "emacs"
```
xu-cheng: Should be `depends_on :emacs => ["23", :optional]`

dunn: Ah, right.

Library/Formula/mu.rb
```diff
@@ -32,7 +21,7 @@ class Mu < Formula
...
 
   env :std if build.with? "emacs"
```
xu-cheng: Also drop this line.

Library/Formula/mu.rb
```diff
@@ -32,7 +21,7 @@ class Mu < Formula
   depends_on "glib"
   depends_on "gmime"
   depends_on "xapian"
-  depends_on Emacs23Installed if build.with? "emacs"
+  depends_on :emacs => ["23", :optional]
 
   env :std if build.with? "emacs"
```
xu-cheng: Can this be dropped?

dunn: Doesn't seem necessary on Yosemite. I'll try removing it.

xu-cheng:
Thanks @dunn

PR #39522, opened 2015-05-08
sslmate: fix typo after latest

vszakats:
use sha256 for another remaining resource


xu-cheng:
Thanks @vszakats :tada:

PR #39523, opened 2015-05-08
bullet: 2.83.4

scpeters:
A few changes in 2.83.4:

* The cmake variable for building demos changed from `BUILD_DEMOS` to `BUILD_BULLET2_DEMOS`.
* The demos are now in the examples folder instead of the Demos folder.
* Demos no longer require extras.
* Also, demos are disabled if shared libraries are specified due to a [linking problem](https://gist.github.com/scpeters/6afc44f0cf916b11a226) during installation.

Previous attempts: #39366 #39436 #39458 (sorry for the noise).

### DIFF
--- a/Library/Formula/bullet.rb
+++ b/Library/Formula/bullet.rb
@@ -44,18 +44,12 @@ class Bullet < Formula
 
     args << "-DUSE_DOUBLE_PRECISION=ON" if build.with? "double-precision"
 
-    with_demo = build.with? "demo"
-    if with_demo && (build.with?("shared") || build.with?("framework"))
-      # Related to the following warnings when building --with-shared --with-demo
-      # https://gist.github.com/scpeters/6afc44f0cf916b11a226
-      opoo "Installed demos don't work with shared libraries"
-      puts <<-EOS.undent
-        Demos will be disabled due to the use of shared libraries.
-        Static libraries are required for demos installed by homebrew.
-      EOS
-      with_demo = false
+    # Related to the following warnings when building --with-shared --with-demo
+    # https://gist.github.com/scpeters/6afc44f0cf916b11a226
+    if build.with?("demo") && (build.with?("shared") || build.with?("framework"))
+      raise "Demos cannot be installed with shared libraries or framework."
     end
-    args << "-DBUILD_BULLET2_DEMOS=OFF" unless with_demo
+    args << "-DBUILD_BULLET2_DEMOS=OFF" if build.without? "demo"
 
     if build.with?("extra")
       args << "-DINSTALL_EXTRA_LIBS=ON"
--- a/Library/Formula/bullet.rb
+++ b/Library/Formula/bullet.rb
@@ -1,7 +1,7 @@
 class Bullet < Formula
   homepage "http://bulletphysics.org/wordpress/"
-  url "https://github.com/bulletphysics/bullet3/archive/2.82.tar.gz"
-  sha256 "93ffcdfdd7aa67159fc18d336456945538a6602e3cd318eed9963280620b55bd"
+  url "https://github.com/bulletphysics/bullet3/archive/2.83.4.tar.gz"
+  sha256 "2cf287cead9a116c56f6d6f15f73dc8b3ed1fe407ef2ca894027d585fab07341"
   head "https://github.com/bulletphysics/bullet3.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


Library/Formula/bullet.rb
```diff
@@ -44,11 +44,20 @@ def install
 
     args << "-DUSE_DOUBLE_PRECISION=ON" if build.with? "double-precision"
 
-    args << "-DBUILD_DEMOS=OFF" if build.without? "demo"
+    with_demo = build.with? "demo"
+    if with_demo && build.with?("shared")
```
tdsmith: or with framework, it looks like

Library/Formula/bullet.rb
```diff
@@ -44,11 +44,20 @@ def install
...
+      # Related to the following warnings when building --with-shared --with-demo
+      # https://gist.github.com/scpeters/6afc44f0cf916b11a226
+      opoo "Installed demos don't work with shared libraries"
+      puts <<-EOS.undent
+      Demos will be disabled due to the use of shared libraries.
+      Static libraries are required for demos installed by homebrew.
```
tdsmith: A couple of nits:
* can you indent this text another couple of spaces?
* This isn't a restriction imposed by Homebrew so I think this text is a little confusing; can this read "Demos will be disabled because they link incorrectly against shared libraries"?

tdsmith:
> (sorry for the noise)

You can add commits to an existing PR by pushing to the same remote branch.

This approach makes sense to me; thanks!

Library/Formula/bullet.rb
scpeters: good call; added in 398fac7621b5c4ed4787a84df6caa1f8a38c6145

Library/Formula/bullet.rb
```diff
@@ -44,11 +44,20 @@ def install
...
+      # Related to the following warnings when building --with-shared --with-demo
+      # https://gist.github.com/scpeters/6afc44f0cf916b11a226
+      opoo "Installed demos don't work with shared libraries"
+      puts <<-EOS.undent
+        Demos will be disabled due to the use of shared libraries.
+        Static libraries are required for demos installed by homebrew.
+      EOS
+      with_demo = false
+    end
+    args << "-DBUILD_BULLET2_DEMOS=OFF" unless with_demo
```
xu-cheng: Instead of silently correcting the behavior, Let's error out.

```ruby
if build.with?("demo") && (build.with?("shared") || build.with?("framework"))
  raise "Installed demos don't work with shared libraries"
end

args << "-DBUILD_BULLET2_DEMOS=OFF" unless build.with? "demo"
```

Library/Formula/bullet.rb
scpeters: * I increased the indentation in 398fac7621b5c4ed4787a84df6caa1f8a38c6145
* Let me clarify what I think the error message means:

The demos are not normally installed when one builds bullet, but in this formula there's nice `prefix.install` calls to install the appropriate build folders. This bring some extra cmake files, but that's not too big of a deal. The issue is that with shared libraries, the executables are linked using the absolute path of the build folder, such that after `prefix.install`ing, the libraries can no longer be found. There is also a [warning message](https://gist.github.com/scpeters/6afc44f0cf916b11a226) during install: `Warning: Could not fix ${lib} in ${executable}`.

I had added the `installed by homebrew` line in the warning message because trying to install them seems unique to homebrew.

That said, I'm open to changing the error message as you recommend. I just wanted to clarify why I had originally phrased it that way.

I'll wait for further comments and then update with the recommended message.

Library/Formula/bullet.rb
scpeters: Sure, I'll do this. I'd be even happier if I could figure out how to fix the linking, but I didn't know how.

scpeters:
> (sorry for the noise)
>
> You can add commits to an existing PR by pushing to the same remote branch.
>
I think I closed the old PR's when I noticed a problem and then squashed
commits before re-opening, and I wasn't able to reopen them. I'll try to
avoid premature squashing in the future.

scpeters:
There are competing suggestions from @xu-cheng and @tdsmith, so I'm waiting for clarification.

Library/Formula/bullet.rb
scpeters: I followed your suggestion after not hearing any other comments.

scpeters:
I went with @xu-cheng 's suggestion. I'll squash if there are no other suggestions.

mikemcquaid:
@xu-cheng @tdsmith mind reviewing and/or merging this? Thanks!

xu-cheng:

Issue #14089, opened 2012-08-09
Depending on a tapped formula without the full path shouldn't be possible


mistydemeo:
Currently, we allow a formula to `depends_on` a tapped formula without specifying the repo name. This has the potential to encourage taps to do some confusing stuff that's hard for end-users to debug.

For instance, php53 and php54 in josegonzalez/php used to depend on an unqualified `zlib` - which only worked for users with homebrew/dupes tapped, since neither mxcl/homebrew nor josegonzalez/php include a `zlib`. This was reported by a user in IRC when `brew upgrade` failed with confusing error messages.

Suggestion: `depends_on` should not allow unqualified names when called with a tapped formula, unless the formula being depended on comes from the same tap. While I'm -1 on the added complexity, inadvertently allowing third-party taps to do Bad Things seems like it's setting the stage for more confusion.

MikeMcQuaid:
Now that every formula is in a tap: I think this no longer applies.

PR #39524, opened 2015-05-08
Symlinks free tap structure

xu-cheng:
cc @Homebrew/owners

### DIFF
--- a/Library/Homebrew/cmd/tap.rb
+++ b/Library/Homebrew/cmd/tap.rb
@@ -5,7 +5,7 @@ module Homebrew
         puts "#{user.basename}/#{repo.basename.sub("homebrew-", "")}" if (repo/".git").directory?
       end
     elsif ARGV.first == "--repair"
-      repair_taps
+      migrate_taps :force => true
     else
       opoo "Already tapped!" unless install_tap(*tap_args)
     end
@@ -26,7 +26,6 @@ module Homebrew
 
     files = []
     tapd.find_formula { |file| files << file }
-    link_tap_formula(files)
     puts "Tapped #{files.length} formula#{plural(files.length, 'e')} (#{tapd.abv})"
 
     if private_tap?(repouser, repo) then puts <<-EOS.undent
@@ -42,59 +41,12 @@ module Homebrew
     true
   end
 
-  def link_tap_formula(paths, warn_about_conflicts=true)
-    ignores = (HOMEBREW_LIBRARY/"Formula/.gitignore").read.split rescue []
-    tapped = 0
-
-    paths.each do |path|
-      to = HOMEBREW_LIBRARY.join("Formula", path.basename)
-
-      # Unexpected, but possible, lets proceed as if nothing happened
-      to.delete if to.symlink? && to.resolved_path == path
-
-      begin
-        to.make_relative_symlink(path)
-      rescue SystemCallError
-        to = to.resolved_path if to.symlink?
-        opoo <<-EOS.undent if warn_about_conflicts
-          Could not create link for #{Tty.white}#{tap_ref(path)}#{Tty.reset}, as it
-          conflicts with #{Tty.white}#{tap_ref(to)}#{Tty.reset}. You will need to use the
-          fully-qualified name when referring this formula, e.g.
-            brew install #{tap_ref(path)}
-          EOS
-      else
-        ignores << path.basename.to_s
-        tapped += 1
-      end
-    end
-
-    HOMEBREW_LIBRARY.join("Formula/.gitignore").atomic_write(ignores.uniq.join("\n"))
-
-    tapped
-  end
-
-  def repair_taps(warn_about_conflicts=true)
-    count = 0
-    # prune dead symlinks in Formula
-    Dir.glob("#{HOMEBREW_LIBRARY}/Formula/*.rb") do |fn|
-      if not File.exist? fn
-        File.delete fn
-        count += 1
-      end
-    end
-    puts "Pruned #{count} dead formula#{plural(count, 'e')}"
-
-    return unless HOMEBREW_REPOSITORY.join("Library/Taps").exist?
-
-    count = 0
-    # check symlinks are all set in each tap
-    each_tap do |user, repo|
-      files = []
-      repo.find_formula { |file| files << file }
-      count += link_tap_formula(files, warn_about_conflicts)
-    end
-
-    puts "Tapped #{count} formula#{plural(count, 'e')}"
+  # Migrate tapped formulae from symlink-based to directory-based structure.
+  def migrate_taps(options={})
+    ignore = HOMEBREW_LIBRARY/"Formula/.gitignore"
+    return unless ignore.exist? || options.fetch(:force, false)
+    (HOMEBREW_LIBRARY/"Formula").children.select(&:symlink?).each(&:unlink)
+    ignore.unlink if ignore.exist?
   end
 
   private
@@ -123,14 +75,5 @@ module Homebrew
     true
   rescue GitHub::Error
     false
-  end
-
-  def tap_ref(path)
-    case path.to_s
-    when %r{^#{Regexp.escape(HOMEBREW_LIBRARY.to_s)}/Formula}o
-      "Homebrew/homebrew/#{path.basename(".rb")}"
-    when HOMEBREW_TAP_PATH_REGEX
-      "#{$1}/#{$2.sub("homebrew-", "")}/#{path.basename(".rb")}"
-    end
   end
 end
--- a/Library/Homebrew/exceptions.rb
+++ b/Library/Homebrew/exceptions.rb
@@ -65,6 +65,25 @@ class TapFormulaUnavailableError < FormulaUnavailableError
   end
 end
 
+class TapFormulaAmbiguityError < RuntimeError
+  attr_reader :name, :paths, :formulae
+
+  def initialize name, paths
+    @name = name
+    @paths = paths
+    @formulae = paths.map do |path|
+      path.to_s =~ HOMEBREW_TAP_PATH_REGEX
+      "#{$1}/#{$2.sub("homebrew-", "")}/#{path.basename(".rb")}"
+    end
+
+    super <<-EOS.undent
+      Formulae found in multiple taps: #{formulae.map { |f| "\n       * #{f}" }.join}
+
+      Please use the fully-qualified name e.g. #{formulae.first} to refer the formula.
+    EOS
+  end
+end
+
 class OperationInProgressError < RuntimeError
   def initialize name
     message = <<-EOS.undent
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -16,6 +16,17 @@ module HomebrewArgvExtension
     @formulae ||= (downcased_unique_named - casks).map { |name| Formulary.factory(name, spec) }
   end
 
+  def resolved_formulae
+    require "formula"
+    @resolved_formulae ||= (downcased_unique_named - casks).map do |name|
+      if name.include?("/")
+        Formulary.factory(name, spec)
+      else
+        Formulary.from_rack(HOMEBREW_CELLAR/name, spec)
+      end
+    end
+  end
+
   def casks
     @casks ||= downcased_unique_named.grep HOMEBREW_CASK_TAP_FORMULA_REGEX
   end
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -51,7 +51,7 @@ module HomebrewArgvExtension
           Keg.new(linked_keg_ref.resolved_path)
         elsif dirs.length == 1
           Keg.new(dirs.first)
-        elsif (prefix = Formulary.factory(canonical_name).prefix).directory?
+        elsif (prefix = (name.include?("/") ? Formulary.factory(name) : Formulary.from_rack(rack)).prefix).directory?
           Keg.new(prefix)
         else
           raise MultipleVersionsInstalledError.new(canonical_name)
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -107,7 +107,7 @@ class Formulary
         name = name_without_version
       end
 
-      super name, Formula.path(name)
+      super name, Formulary.path(name)
     end
 
     def get_formula(spec)
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -200,6 +200,22 @@ class Formulary
     loader_for(ref).get_formula(spec)
   end
 
+  # Return a Formula instance for the given rack.
+  def self.from_rack(rack, spec=:stable)
+    kegs = rack.directory? ? rack.subdirs.map { |d| Keg.new(d) } : []
+
+    keg = kegs.detect(&:linked?) || kegs.detect(&:optlinked?) || kegs.max_by(&:version)
+    return factory(rack.basename.to_s, spec) unless keg
+
+    tap = Tab.for_keg(keg).tap
+
+    if tap.nil? || tap == "Homebrew/homebrew"
+      factory(rack.basename.to_s, spec)
+    else
+      factory("#{tap.sub("homebrew-", "")}/#{rack.basename}", spec)
+    end
+  end
+
   def self.canonical_name(ref)
     loader_for(ref).name
   rescue TapFormulaAmbiguityError
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -202,6 +202,10 @@ class Formulary
 
   def self.canonical_name(ref)
     loader_for(ref).name
+  rescue TapFormulaAmbiguityError
+    # If there are multiple tap formulae with the name of ref,
+    # then ref is the canonical name
+    ref.downcase
   end
 
   def self.path(ref)
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -182,7 +182,7 @@ class Formulary
 
   class NullLoader < FormulaLoader
     def initialize(name)
-      super name, Formula.path(name)
+      super name, Formulary.core_path(name)
     end
 
     def get_formula(spec)
@@ -222,7 +222,7 @@ class Formulary
       return FromPathLoader.new(ref)
     end
 
-    formula_with_that_name = Formula.path(ref)
+    formula_with_that_name = core_path(ref)
     if formula_with_that_name.file?
       return FormulaLoader.new(ref, formula_with_that_name)
     end
@@ -238,5 +238,9 @@ class Formulary
     end
 
     return NullLoader.new(ref)
+  end
+
+  def self.core_path(name)
+    Pathname.new("#{HOMEBREW_LIBRARY}/Formula/#{name.downcase}.rb")
   end
 end
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -232,6 +232,13 @@ class Formulary
       return AliasLoader.new(possible_alias)
     end
 
+    possible_tap_formulae = tap_paths(ref)
+    if possible_tap_formulae.size > 1
+      raise TapFormulaAmbiguityError.new(ref, possible_tap_formulae)
+    elsif possible_tap_formulae.size == 1
+      return FormulaLoader.new(ref, possible_tap_formulae.first)
+    end
+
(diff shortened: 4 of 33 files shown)


Library/Homebrew/cmd/tap.rb
```diff
@@ -42,59 +41,17 @@ def install_tap user, repo
     true
   end
 
-  def link_tap_formula(paths, warn_about_conflicts=true)
-    ignores = (HOMEBREW_LIBRARY/"Formula/.gitignore").read.split rescue []
-    tapped = 0
+  # migrate from symlink based tap structure to directories based structure
+  def migrate_taps(options={:autoskip => true})
+    ignore = HOMEBREW_LIBRARY/"Formula/.gitignore"
+    return if options[:autoskip] && !ignore.exist?
```
jacknagel: I'm not sure I understand what "autoskip" means, as opposed to just "skip"... what's automatic about it? Maybe I don't get what the option does.

Library/Homebrew/cmd/tap.rb
```diff
@@ -42,59 +41,17 @@ def install_tap user, repo
     true
   end
 
-  def link_tap_formula(paths, warn_about_conflicts=true)
-    ignores = (HOMEBREW_LIBRARY/"Formula/.gitignore").read.split rescue []
-    tapped = 0
+  # migrate from symlink based tap structure to directories based structure
+  def migrate_taps(options={:autoskip => true})
```
jacknagel: This isn't really the best way to add an option the defaults to true... if someone calls the method with a hash argument but omits the autoskip option, it will be falsy instead. I think this is more idiomatic:

```rb
def migrate_taps(options={})
  autoskip = options.fetch(:autoskip, true)
```

jacknagel:
This is looking good!

Library/Homebrew/cmd/tap.rb
jacknagel: Maybe I should phrase it this way: what are the different modes of operation of this method?

xu-cheng: I use `gitignore` file to determine whether we have migrated or not. It `gitingore` file doesn't exist, we can skip it. But when doing `brew tap --repair`, I'm thinking it should force the migration and we should scan the symlink files regardless of the `gitignore` file.

jacknagel: Maybe `force` a better name, then?

xu-cheng: Good idea

xu-cheng:
Updated

Library/Homebrew/exceptions.rb
```diff
@@ -65,6 +65,21 @@ def to_s; <<-EOS.undent
...
+  attr_reader :name, :paths
+
+  def initialize name, paths
+    @name = name
+    @paths = paths
+
+    super <<-EOS.undent
+      Multiple formulae found: #{paths.map { |f| "\n       * #{f}" }.join}
+
+      Please use the fully-qualified name to refer the formula.
```
mikemcquaid: Might want to provide an example here.

mikemcquaid:
This is great. Would love to see it merged pretty soon too (if @jacknagel has no objections). The main thing I'd advise you is to ensure that anything called from `update` is super-defensively coded as when we have errors there it makes huge numbers of users freak out (I've done that, it's not fun). Nice work!

Library/Homebrew/exceptions.rb
xu-cheng: Like this?
```
$ touch Library/Taps/homebrew/homebrew-python/latex-mk.rb
$ brew info latex-mk
Error: Multiple formulae found:
 * /Users/xucheng/Coding/Github/homebrew/Library/Taps/homebrew/homebrew-python/latex-mk.rb
 * /Users/xucheng/Coding/Github/homebrew/Library/Taps/homebrew/homebrew-tex/Formula/latex-mk.rb

Please use the fully-qualified name to refer the formula.
```

mikemcquaid: Sorry, I mean `Please use the fully-qualified name e.g. user/tap/formula to refer the formula.`

mikemcquaid: And instead of a full path it'd be good to make this just e.g. `homebrew/python/latex-mk`.

xu-cheng: Updated. Now it looks like this:
```
$ brew info latex-mk
Error: Multiple formulae found:
 * homebrew/python/latex-mk
 * homebrew/tex/latex-mk

Please use the fully-qualified name e.g. homebrew/python/latex-mk to refer the formula.
```

Library/Homebrew/exceptions.rb
```diff
@@ -65,6 +65,25 @@ def to_s; <<-EOS.undent
...
+    @paths = paths
+    @formulae = paths.map do |path|
+      path.to_s =~ HOMEBREW_TAP_PATH_REGEX
+      "#{$1}/#{$2.sub("homebrew-", "")}/#{path.basename(".rb")}"
+    end
+
+    super <<-EOS.undent
+      Multiple formulae found: #{formulae.map { |f| "\n       * #{f}" }.join}
+
+      Please use the fully-qualified name e.g. #{formulae.first} to refer the formula.
```
mikemcquaid: :heart_eyes:

jacknagel:
I think this is in good shape, though given the importance of getting this right we should probably have all @Homebrew/owners run with this locally for a few days before we roll it out to everyone.

Library/Homebrew/exceptions.rb
```diff
@@ -65,6 +65,25 @@ def to_s; <<-EOS.undent
...
+  def initialize name, paths
+    @name = name
+    @paths = paths
+    @formulae = paths.map do |path|
+      path.to_s =~ HOMEBREW_TAP_PATH_REGEX
+      "#{$1}/#{$2.sub("homebrew-", "")}/#{path.basename(".rb")}"
+    end
+
+    super <<-EOS.undent
+      Multiple formulae found: #{formulae.map { |f| "\n       * #{f}" }.join}
```
mikemcquaid: Maybe `Formulae found in multiple taps:` or `multiple locations` or something? I dunno.

mikemcquaid: Yeh, `multiple taps` makes sense given the current behaviour. I do wonder if we want to just treat core as another tap, though (perhaps in another PR)? e.g. with this PR if something is found in multiple taps then it produces this nice message but not if it's in a tap and core with the same name e.g. `omega` and `homebrew/games/omega`. In the medium-term we've got a GSoC student working on tap priorities and long-term I think we want to be able to use Homebrew without core formulae at all and use only a tap. Interested in your thoughts here.

mikemcquaid:
`brew outdated` prints this message for me now:

```
Error: Multiple formulae found: 
 * boxen/brews/gcc48
 * homebrew/versions/gcc48

Please use the fully-qualified name e.g. boxen/brews/gcc48 to refer the formula.
```

It should probably be able to tell what the right version is itself, though.

mikemcquaid:
And `brew upgrade` produces the same message.

mikemcquaid:
And `brew uninstall`.

mikemcquaid:
and `brew cleanup`. Generally I think if this code is dealing with an installed formula it should automatically resolve e.g. `gcc48` to `boxen/brews/gcc48` if that's the version that's installed (which it is, in this case). If it's not installed yet it should prompt the user.

Library/Homebrew/formulary.rb
```diff
@@ -222,7 +223,21 @@ def self.loader_for(ref)
...
+      kegs = rack.subdirs.map { |d| Keg.new(d) }
+      raise FormulaUnavailableError.new(name) if kegs.empty?
+      keg = kegs.detect(&:linked?) || kegs.detect(&:optlinked?) || kegs.max_by(&:version)
+      tap = Tab.for_keg(keg).tap
+      if tap == "Homebrew/homebrew"
+        return FormulaLoader.new(name, Formulary.core_path(name))
+      else
+        return TapLoader.new("#{tap.sub "homebrew-", ""}/#{name}")
+      end
+    end
```
xu-cheng: @jacknagel @mikemcquaid I added a new logic to allow formula to load from rack path. I think we can use this to solve the auto detect problem. Why do your guys think? I will work on to utilize this in different commands.

Library/Homebrew/exceptions.rb
xu-cheng: Yes, current implementation will prioritize core formula. But it can be easily changed if wanted.

>  I think we want to be able to use Homebrew without core formulae at all and use only a tap.

I am always confused about how we `depends_on` another formula in the same tap for a tap formula. Should we use full qualified name or not? Deep down is that we need to solve #14089 first.

mikemcquaid: Perhaps just push the prioritisation to another PR as this one better matches current behaviour. I think we should require the fully-qualified name. This will require formulae changes in taps so should be introduced with a warning for quite a while before we start erroring out.

Library/Homebrew/formulary.rb
mikemcquaid: Will try it out, thanks.

mikemcquaid: Yeh, this seems pretty great to me. Finally my `brew outdated` is useful again because it's not trying to upgrade versions of my stuff installed from taps (that are intentionally old versions).

xu-cheng:
@mikemcquaid I think I have resolved most of `TapFormulaAmbiguityError`. (I may miss some places). But I cannot solve the problems in the bottle, `brew bottle` and `brew instal <bottle file> will fail if there is a TapFormulaAmbiguityError. This is because the bottle file name doesn't contain the information of tap. (unless we extract `INSTALL_RECEIPT.json` first)

Library/Homebrew/formulary.rb
jacknagel: How many places are we calling factory and expecting to hit this logic? In the interest of reducing the number of possible ways to specify a formula (which was one of the original goals of the refactored formula loading process), I would like to see this live outside of Formulary.factory and be called explicitly when needed.

mikemcquaid:
> @mikemcquaid I think I have resolved most of TapFormulaAmbiguityError. (I may miss some places). But I cannot solve the problems in the bottle, brew bottle and brew instal <bottle file> will fail if there is a TapFormulaAmbiguityError. This is because the bottle file name doesn't contain the information of tap. (unless we extract INSTALL_RECEIPT.json first)

Yup, that seems OK, thanks. Extracting INSTALL_RECEIPT.json is a good idea for a future PR because it means we can get significantly more metadata than the current hacky way.

Library/Homebrew/formulary.rb
xu-cheng: > How many places are we calling factory and expecting to hit this logic?

I think it's all in the commits below. About 10 or so.

> I would like to see this live outside of Formulary.factory and be called explicitly when needed.

A method inside `utils.rb` with the name like `rack_to_formula`?

jacknagel: I need to think about the ambiguity problem a bit. From a design standpoint, wrapping calls to factory in more exception handling bothers me quite a bit.

jacknagel:
Is there really two problems here?

 (a) implementing taps without symlinks
 (b) formula ambiguity

Is it possible to separate the two? In other words, can we merge (a) minus the code that raises TapFormulaAmbiguityError, and then address that in a separate pull request?

I need to do more thinking about the ambiguity problem.

Library/Homebrew/cmd/info.rb
```diff
@@ -29,6 +29,13 @@ def print_info
         puts unless i == 0
         begin
           info_formula Formulary.factory(f)
+        rescue TapFormulaAmbiguityError
+          rack = HOMEBREW_CELLAR/f
+          if rack.directory?
+            info_formula Formulary.factory(rack.to_s)
+          else
```
jacknagel: It looks like most of the places where we'd need to rescue the exception are written or can be rewritten to follow the pattern:

```rb
begin
  formula = Formulary.factory(some_variable)
rescue TapFormulaAmbiguityError
  rack = HOMEBREW_CELLAR/some_variable
  if rack.directory?
    formula = Formulary.factory(rack.to_s)
  else
    raise
  end
end
```

That indicates to me that this logic should be handled inside the call to factory. That would avoid the need to rescue this exception in most cases and simplify the logic in others.

I tentatively take back what I said about moving the rack logic out of factory.

Library/Homebrew/formulary.rb
```diff
@@ -222,7 +227,21 @@ def self.loader_for(ref)
...
     end
 
-    formula_with_that_name = Formula.path(ref)
+    if ref.start_with?(HOMEBREW_CELLAR) && File.directory?(ref)
+      rack = Pathname.new(ref)
+      name = rack.basename.to_s
+      kegs = rack.subdirs.map { |d| Keg.new(d) }
+      raise FormulaUnavailableError.new(name) if kegs.empty?
+      keg = kegs.detect(&:linked?) || kegs.detect(&:optlinked?) || kegs.max_by(&:version)
+      tap = Tab.for_keg(keg).tap
```
xu-cheng: @jacknagel What's the best way to get formula origin. `Tab#tap` wouldn't work for formula installed from url or path. `Tab#source["path"]` may not exist for formula installed before 838eb29bcf4672cf7e6832d3e3cd9cd0ef1356ff.

jacknagel: If we can't detect it from the tap or path, we have to fall back to the formula in core, if there is one.

xu-cheng: I have just updated to make loading from rack an internal logic of Formulary. And under current context, it will definite not be a formula in core as TapFormulaAmbiguityError is only for taps.

jacknagel:
The one hangup I have about this is that teaching factory about installed formulae feels like magic and may produce non-obvious results.

If I have two formulae available named "gcc48", I would *expect* that informational commands like `brew info gcc48` or `brew deps gcc48` or even a action command like `brew fetch gcc48` to tell me, otherwise it seems it would be hard to discover that I even have two same-named formulae at all.

I realize that at this moment we don't have such a warning, and just pick whichever one was tapped first, but we're switching from that heuristic to the "whichever is installed" heuristic, which is different.

I dunno.

Library/Homebrew/formulary.rb
jacknagel: I see, I think we have no choice but to raise in that case.

jacknagel:
To put in another way, in otherwise-identical Homebrew installations, `Formulary.factory` can now return different results.

xu-cheng:
I just found out for every `rescue FormulaUnavailableError` in the core, I need to handle `TapFormulaAmbiguityError` as well. So there are still lots of work to do to make it work.

Basically we have two choices now:
* Teaching factory about installed formulae. I think @mikemcquaid may prefer this one.
* Enforce full-qualified name in the commands if there is an ambiguity. Put the logic of loading formula from rack into `TapFormulaAmbiguityError`. And use it case by case like this:
```ruby
begin
  Formulary.factory(name)
rescue FormulaUnavailableError
  ...
rescue TapFormulaAmbiguityError => e
  if (f = e.guess_formula_from_rack)
    ...
  else
    ...
  end
end
```

jacknagel:
So I thought about this more, and I think we're failing to consider the context in which each command is run, particularly, whether the command is operating on installed formulae or not.

Commands that always operate on installed formulae (e.g. `brew outdated`,  `brew upgrade`, `brew cleanup`, `brew missing`, etc.) should always "just work", since we have both the context and the information required to do the disambiguation for the user.

Commands that just work with information (e.g. `brew deps`) from inside the formula should probably require the user to be more specific.

So, from an implementation standpoint, clearly `Formula.installed` should always consider the on-disk state when loading formulae.

`ARGV.formulae` shouldn't, though. But since we know on a per-command basis whether or not we want to look at the on-disk state, we can just add `ARGV.installed_formulae` that does this and keep `ARGV.formulae` as-is.

xu-cheng:
@jacknagel I did a partial implementation based on your suggestion. If it's good, I will apply this to different commands.

Library/Homebrew/extend/ARGV.rb
```diff
@@ -16,6 +16,21 @@ def formulae
...
+    @installed_formulae ||= (downcased_unique_named - casks).map do |name|
+      begin
+        Formulary.factory(name, spec)
+      rescue TapFormulaAmbiguityError => e
+        if (f = e.formula_from_rack(spec))
+          f
+        else
+          raise
+        end
+      end
```
jacknagel: Then this could be rewritten to always prefer the on-disk state, too.

jacknagel:
Just to be extra clear, this is what I'm saying. Given the following:

 * a formula named “foo” in core
 * a formula named “foo” in tap A
 * an installation of “foo” from tap A

`Formulary.factory(“foo”) is always going to return “foo” from core. But in the case of `Formula.installed` or `ARGV.installed_formulae`, that’s not what we want to happen; we want “foo” from A. In those cases we always want to look at the information in the rack/keg.

mikemcquaid:
> Commands that always operate on installed formulae (e.g. brew outdated, brew upgrade, brew cleanup, brew missing, etc.) should always "just work", since we have both the context and the information required to do the disambiguation for the user.
> 
> Commands that just work with information (e.g. brew deps) from inside the formula should probably require the user to be more specific.

Yes, I totally agree with this :+1:

mikemcquaid:
@xu-cheng Let me know when you've done something like the above (or if you have already, knowing you :wink:) and I'll take another look/test.

xu-cheng:
@mikemcquaid you can test this now. I think I have handled all `ARGV.formulae` and `Formulary.factory`.

Library/Homebrew/extend/ARGV.rb
```diff
@@ -16,6 +16,17 @@ def formulae
     @formulae ||= (downcased_unique_named - casks).map { |name| Formulary.factory(name, spec) }
   end
 
+  def installed_formulae
```
xu-cheng: I think this may need a better name as it can load uninstalled formula as well.

DomT4: `def existing_formulae`? Or `def known_formulae` perhaps.

xu-cheng:
Request some comments.

mikemcquaid:
Works for me :+1:

jacknagel:
I apologize, I've been busy this week. I will review again tomorrow or Saturday.

Library/Homebrew/formulary.rb
```diff
@@ -200,8 +200,28 @@ def self.factory(ref, spec=:stable)
     loader_for(ref).get_formula(spec)
   end
 
+  # Return a Formula instance for the given rack.
+  def self.factory_from_rack(rack, spec=:stable)
```
jacknagel: Should this be `formula_from_rack` (or just `from_rack`)? `factory_from_rack` sounds a little weird to me (`factory` itself is also probably a bad name).

jacknagel:
Code-wise this looks good to me. I wonder if `installed_formulae` should raise if a non-installed formula is passed, though? It depends on how it's being used by perhaps that would be a useful behavior.

Library/Homebrew/formulary.rb
xu-cheng: Updated to use `from_rack`.

xu-cheng:
> I wonder if installed_formulae should raise if a non-installed formula is passed

Under current implementation, not raising seems be better choice. The method callers already handle the non installed formula. At the same time, it can benefit methods like `--prefix` and `--cellar` which prefer installed formula but can accept non installed formula as well. So maybe we need better method name?

Library/Homebrew/extend/ARGV.rb
mikemcquaid: Or `resolved_formulae` or `qualified_formulae`.

xu-cheng: `resolved_formulae` seems to be the best choice.

xu-cheng:
Updated to change the method name to `ARGV#resolved_formulae`

mikemcquaid:
One possible tweak for this PR: I think having `git` installed and then running `brew upgrade boxen/brews/git` should "upgrade" (or install) the latest version of `boxen/brews/git` even if the installed version is newer. Thoughts? That could be another PR.

mikemcquaid:
Actually, never mind, this is a mistake due to how Boxen packages Git.

mikemcquaid:
This is working well for me. If @jacknagel is happy then I'd love to see us merge this.

jacknagel:
lgtm

jasonkarns:
Thanks so much @xu-cheng ! homebrew's behavior with same-name formulas from different taps has prevented me from running `brew upgrade` for months. No more!

PR #39525, opened 2015-05-08
Add PHP, Python and Ruby support to GEOS

mcuelenaere:
Note: this currently hardcodes php55 as the homebrew PHP version. Is there a way to dynamically figure out which php formula is currently installed? (php53, php54, php55 or php56)

### DIFF
--- a/Library/Formula/geos.rb
+++ b/Library/Formula/geos.rb
@@ -14,6 +14,11 @@ class Geos < Formula
 
   option :universal
   option :cxx11
+  option "with-php", "Build the PHP extension"
+  option "with-python", "Build the Python extension"
+  option "with-ruby", "Build the ruby extension"
+
+  depends_on "swig" => :build if build.with?("python") || build.with?("ruby")
 
   fails_with :llvm
 
@@ -21,7 +26,16 @@ class Geos < Formula
     ENV.universal_binary if build.universal?
     ENV.cxx11 if build.cxx11?
 
-    system "./configure", "--disable-dependency-tracking", "--prefix=#{prefix}"
+    args = [
+      "--disable-dependency-tracking",
+      "--prefix=#{prefix}",
+    ]
+
+    args << "--enable-php" if build.with?("php")
+    args << "--enable-python" if build.with?("python")
+    args << "--enable-ruby" if build.with?("ruby")
+
+    system "./configure", *args
     system "make", "install"
   end
 


tdsmith:
for PHP, you'll probably want to add a block at the top that looks like:

```ruby
class PhpRequirement < Requirement
  fatal true
  satisfy { which "php" }
end
```

and then in the Formula body, write `depends_on PhpRequirement => :optional` which will add php to PATH inside `install`. I think you may need to do the same thing for Ruby, since the `:ruby` requirement requires an exact version prefix match, which may not be what you want? @xu-cheng, you've thought about requirements more than I have; does that sound right?

You shouldn't need to do this for python, since recent versions of OS X all ship python 2.7. I think you can default the python extensions to on, so write i.e. `option "without-python"` instead of `"with-python"`.

Library/Formula/geos.rb
```diff
@@ -13,6 +13,12 @@ class Geos < Formula
 
   option :universal
   option :cxx11
+  option 'with-php', 'Build the PHP extension'
+  option 'with-python', 'Build the Python extension'
+  option 'with-ruby', 'Build the ruby extension'
+  option 'use-homebrew-php', 'Use Homebrew PHP' if build.include?('with-php')
```
tdsmith: Style notes:
* we follow the [Github Ruby style guide](https://github.com/styleguide/ruby) for new commits, so please use double quotes everywhere.
* use `build.with?("php")` instead of `build.include?("with-pip")`

xu-cheng:
For this use case, the current solution is to add `env :std` (we do this in the `vim.rb`). Yeah, I would agree `Requirement` should be a better approach. So should we start to support like `depends_on :php`, `depends_on :ruby`, `depends_on :perl` etc. without version requirement? cc @mikemcquaid as you are doing requirement cleanup.

mcuelenaere:
How about this?

mcuelenaere:
ping

Library/Homebrew/requirements/php_requirement.rb
```diff
@@ -0,0 +1,11 @@
...
+  fatal true
+  satisfy { which "php" }
+
+  env do
+    if which("php") != "/usr/bin/php"
+      php_prefix = Pathname.new Utils.popen_read(which("php-config"), "--prefix").strip
+      ENV.prepend_path "PATH", "#{php_prefix}/bin"
+    end
+  end
+end
```
xu-cheng: New line here.

Library/Homebrew/requirements/php_requirement.rb
```diff
@@ -0,0 +1,11 @@
+class PhpRequirement < Requirement
+  fatal true
+  satisfy { which "php" }
+
+  env do
+    if which("php") != "/usr/bin/php"
```
xu-cheng: Why this is needed?

Library/Homebrew/dependency_collector.rb
```diff
@@ -134,6 +134,7 @@ def parse_symbol_spec(spec, tags)
       Dependency.new("libtool", tags)
     when :python2
       PythonDependency.new(tags)
+    when :php        then PhpRequirement.new(tags)
```
xu-cheng: Put this line more forward, near `when :ruby` line. Thanks.

xu-cheng:
Ping @mikemcquaid for opinion.

mikemcquaid:
All versions of OS X we support ship with PHP so we don't need the PhpRequirement here.

Library/Homebrew/requirements/php_requirement.rb
mcuelenaere: This adds `/usr/local/bin/php-config` to the `$PATH` variable that homebrew sets up. This is necessary if you want to link this against your homebrew version of PHP, otherwise it'll use the system one.

mcuelenaere:
@mikemcquaid true, but how can I make geos link with the homebrew PHP version then?

mikemcquaid:
Ok, I see.

Library/Homebrew/requirements/php_requirement.rb
mikemcquaid: We don't want this; if you want to link against a custom `php`  you should make it first in your PATH.

mcuelenaere:
The thing @xu-cheng said (using `env :std`) would probably also work, perhaps that's the way to go? (is there a way to specify this at install time?)

Library/Homebrew/requirements/php_requirement.rb
mcuelenaere: It is first in my `$PATH`, but homebrew doesn't pick that up (it sets up a `$PATH` of its own, hence this hack).

mikemcquaid: That's not the case for `Requirements`; they will example your custom PATH in this `env` block.

mcuelenaere:
How about this then? This removes the PhpRequirement, but still allows me to install it with my homebrew PHP version through `brew install geos --with-php --env=std`.

mcuelenaere:
If this is acceptable, I'll rebase + squash.

mikemcquaid:
I'm OK with the PHPRequirement it just needs to not refuse to build against the system version.

mcuelenaere:
What's still needed to get this merged?

mikemcquaid:
@mcuelenaere Default the Python bindings to disabled by default and then I'm happy to include it.

mcuelenaere:
Squashed + rebased.

mikemcquaid:

tdsmith:
> Default the Python bindings to disabled by default

Why not bottle them?

mikemcquaid:
@tdsmith I'm happy to enable them but I was under the impression that that meant building against the Homebrew Python. If you can make them usable by both: :+1:

PR #39526, opened 2015-05-08
midnight-commander: pass strict audit, add head, url updates

vszakats:
* pass strict audit
* use http instead of ftp for `mirror`
* use SSL/TLS `homepage`
* add --HEAD


xu-cheng:
Thanks.

PR #39527, opened 2015-05-08
freeradius-server 2.2.7, 3.0.8

vszakats:
* bump devel and stable versions
  * http://freeradius.org/press/index.html#3.0.8
  * http://freeradius.org/press/index.html#2.2.7
* add http mirrors
* use .bz2 package for stable too
* use SHA256 hashes


xu-cheng:
Thanks

PR #39531, opened 2015-05-08
sqlite 3.8.10

vszakats:


Library/Formula/sqlite.rb
```diff
@@ -21,6 +21,7 @@ class Sqlite < Formula
   option "with-unlock-notify", "Enable the unlock notification feature"
   option "with-icu4c", "Enable the ICU module"
   option "with-functions", "Enable more math and string functions for SQL queries"
+  option "with-dbstat", "Enable the \"dbstat\" virtual table"
```
DomT4: Why `\"dbstat\"` instead of `Enable the 'dbstat' virtual table`?

vszakats: No strong reason, I copied the sentence from the sqlite change listing and kept the double-quotes. I'll update to single ones.

tdsmith:
Can you squash these to a single commit please?

xu-cheng:
Thanks.

Issue #14374, opened 2012-08-22
Handle Formula renames and deletions


mxcl:
The git diff tells us what has been renamed. So we need to watch that and rename kegs and update symlinks. We need this for taps too.

The git diff tells us what has been deleted. So we need to watch that and look in Aliases to see if a new one is created and if so treat it as a "dupe" being deleted and do the above. Not for taps: taps don't have Aliases (yet).

When we can do this we need to delete one of

* lastfm_fplib and lastfmfpclient (dupes of each other).

When we can do this we need to rename:

* mosh -> mosh-scheme
* cmu-sphinxbase and cmu-pocketsphinx (remove the cmu prefix)
* yuicompressor -> yui-compressor or even just yui

This thread serves as official renames thread. Please research the Debian names also as this is the canonical naming DB.

guoxiao:
`stash-cli` => `atlassian-cli`
see #40909

mikemcquaid:
Support for this has been implemented in #41006. I'm going to leave this issue open until we've migrated more of the mentioned formulae in here to new names.

jhass:
That's great news, can we have `crystal` renamed to `crystal_generator` then, to reduce confusion with the much more popular Crystal language? See the discussion in #40530.

Are there plans for any process around requesting renames, especially from non-maintainers?

mikemcquaid:
@jhass We'll be considering each rename on a case-by-case basis. You can request renames but obviously we can't make any promises either way right now.

jhass:
Well, consider the above a request then.

wesbland:
As mentioned way above and in a closed ticket (#16298), `mpich2` --> `mpich`.

mikemcquaid:
Closing this now that this is implemented and we've migrated a few formulae without end-user issues. There are a few open PRs for renames but if there isn't one open for your chosen rename: please submit a pull request following the documentation in https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Rename-A-Formula.md and format in e.g. https://github.com/Homebrew/homebrew/issues/14374. Thanks @vladshablinsky!

PR #39532, opened 2015-05-08
Updated the naming of objective-caml to just ocaml everywhere

fxfactorial:
OCaml's name is not objective-caml but actually just OCaml, http://caml.inria.fr/ocaml/name.en.html. 

(This confusion of naming is an unnecessary stumbling block for ocaml beginners.)

Brew should reflect that, this pull request does that. 
I followed the trail of all the things that `brew uses ocaml` showed. If this pull request goes through then I will finish the 3 leftover ones that are in homebrew-science and in homebrew-versions.

### DIFF
--- a/Library/Aliases/ocaml
+++ b/Library/Aliases/ocaml
@@ -1,1 +1,0 @@
-../Formula/objective-caml.rb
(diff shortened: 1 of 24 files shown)


mistydemeo:
Unfortunately, for technical reasons we're not able to support formula renames at the moment - this will cause problems for users who already have it installed under the other name. Since we provide the alias under the name `ocaml`, it is possible for users to `brew install ocaml`, `brew info ocaml`, etc. Sorry!

mistydemeo:
There's a (rather old) issue open for formula renames, #14374.

fxfactorial:
Thank you, I didn't know about that previous issue. Shame that we can't rename things.

PR #39533, opened 2015-05-08
proof-general: use EmacsRequirement

dunn:
Proof General is packaged with lisp files compiled for Emacs 23.2 [but it's not even compatible with 23.2 anymore](http://proofgeneral.inf.ed.ac.uk/download), so `make install` recompiles them by default.  So a hard dependency seems appropriate.


xu-cheng:
Thanks @dunn

PR #39534, opened 2015-05-08
Update CODEOFCONDUCT.md

cheapwebmonkey:
phrasing/grammar fix in CONSIDERATE paragraph


tdsmith:
Thanks for the PR! (And "oftentimes" is a single word, so while we're here...)

PR #39537, opened 2015-05-08
fontconfig: disbale --with-add-fonts on linux

ntkme:


mistydemeo:
Can you submit this change to [Linuxbrew](https://github.com/homebrew/linuxbrew) instead, please? Thank you!

PR #39538, opened 2015-05-08
sshfs: pass strict audit

vszakats:
* use SSL/TLS homepage
* SHA256
* add test (not a very meaningful test can be done without a mockup SSH server)
* quotes, misc


xu-cheng:
Thanks, next time it would be great that if you can squash the commits.

PR #39539, opened 2015-05-08
keepassc: pass strict audit

vszakats:
* SSL/TLS homepage
* SHA256
* double space after comma (uncaught by audit)


xu-cheng:
Thanks @vszakats

PR #39548, opened 2015-05-09
Added libiomp's include dir to clang search paths.

metacollin:


xu-cheng:
So  `#include <libiomp/omp.h>` should be `#include <omp.h>` in test case. Right?

metacollin:
....yes.  I can't believe I over looked that.  Sorry, fixing now.

metacollin:
Fixed.

xu-cheng:
Tithe build failed. You may try to change the building system to cmake. (llvm.rb has the similar change recently, if you need a reference).

metacollin:
I've always favor of moving to cmake.  :D I am going to bed but I'll do it first thing tomorrow morning.  Thanks!

metacollin:
Hi, I've migrated the formula to cmake.  Unfortunately, it is not quite as elegant and adds a number of lines to the formula, the min macosx version flags are vital, or compiler-rt will attempt to cross-compile for ARM (which fails hard) etc.   But I ran brew audit and built (and am using) the new binary from the updated formula and it both finds <omp.h> in its new location and all other headers it should, and seems to be issue free (at least on my system).

But I'll see what the build bot says :)

xu-cheng:

tdsmith:
:saxophone::turtle: :+1:

PR #39550, opened 2015-05-09
td 1.2 (new formula)

Swatto:
Your todo list in your terminal: small and fast

### DIFF
--- a/Library/Formula/tvnamer.rb
+++ b/Library/Formula/tvnamer.rb
@@ -5,6 +5,13 @@ class Tvnamer < Formula
   sha256 "c28836f4c9263ee8ad6994788ad35f00e66fa1bd602e876364cd9b938f2843c8"
   head "https://github.com/dbr/tvnamer.git"
 
+  bottle do
+    cellar :any
+    sha256 "e2054cad544e9a1bc8d4eb7925c33909d126efaa7cb79a63d58e655a9d5b6e7d" => :yosemite
+    sha256 "ccbdd91361cf82fa43718a3aa43291901d7ec26864a03422b47dbb0d6ebcf71c" => :mavericks
+    sha256 "a055ef2f5c60a5e0ae4207ad85db4b9c6c961c0b0bea74505ebd47e1d331b32a" => :mountain_lion
+  end
+
   depends_on :python if MacOS.version <= :snow_leopard
 
   resource "tvdb_api" do


Library/Formula/td.rb
```diff
@@ -0,0 +1,49 @@
...
+    mkdir_p buildpath/"src/github.com/Swatto/"
+    ln_sf buildpath, buildpath/"src/github.com/Swatto/td"
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    system "go", "build", "."
+    bin.install "td-1.2.0" => "td"
+  end
+
+  test do
+    system "#{bin}/td"
```
tdsmith: Can this test be expanded to exercise some function of `td`? `$HOME` is set to a temporary directory so the test can use the home directory without interfering with a user's data.

Library/Formula/td.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def caveats; <<-EOS.undent
+    Now you need to add "TODO_DB_PATH" to your environment variable.
+    This will be use as a fallback if your directories does not contains any ".todos" files.
+
+    Find more information here:
+    #{homepage}
+    EOS
```
tdsmith: Please remove the caveats since they aren't Homebrew-specific.

Library/Formula/td.rb
```diff
@@ -0,0 +1,49 @@
+require "language/go"
+
+class Td < Formula
+  homepage "https://github.com/Swatto/td"
+  revision 1
```
tdsmith: The `revision` line can be deleted.

Library/Formula/td.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  stable do
+    url "https://github.com/Swatto/td/archive/1.2.0.tar.gz"
+    sha256 "86a84cb183e2b3700a23cc5570b65ea8d133e51c5c59e321929ace727253d55b"
+  end
```
tdsmith: This doesn't need to be in a `stable` block; these can just be at the class level.

Library/Formula/td.rb
DomT4: If this is a new formula, it doesn't have a `revision`. The `revision` is an unwritten 0.

Library/Formula/td.rb
```diff
@@ -0,0 +1,49 @@
+require "language/go"
+
+class Td < Formula
+  homepage "https://github.com/Swatto/td"
+  revision 1
+
+  stable do
```
DomT4: Probably doesn't need to be nested inside a `stable do` block unless stable has patches, or special deps/resources.

Library/Formula/td.rb
```diff
@@ -0,0 +1,49 @@
...
+    ENV["GOBIN"] = bin
+    ENV["GOPATH"] = buildpath
+    ENV["GOHOME"] = buildpath
+
+    mkdir_p buildpath/"src/github.com/Swatto/"
+    ln_sf buildpath, buildpath/"src/github.com/Swatto/td"
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    system "go", "build", "."
+    bin.install "td-1.2.0" => "td"
```
DomT4: To save others having to rewrite this with each bump, `bin.install "td-#{version}" => "td"` may be nicer.

Library/Formula/td.rb
DomT4: Oops, Sorry. Tim's comments didn't exist before I commented, they appeared once I pressed the refresh key. Sorry Tim :sweat_smile:

Library/Formula/td.rb
Swatto: Since the user needs to do an operation, what should I use?

DomT4:
Please address Tim's comments, Thanks!

PR #39557, opened 2015-05-09
apr-util: bump revision

DomT4:
Resolves build & runtime issues for some dependents. Seems likely we're going to need to do this every time `apr` is bumped out of sync with `apr-util`.


DomT4:
Bot failed to fetch one dependant to test on Yosemite. Have re-queued it :cry:. This can close #39362 as well.

PR #39560, opened 2015-05-09
Update to Riak 2.1.1.

michaelahlers:
Incidentally, [Basho's OS X installation instructions](http://docs.basho.com/riak/latest/ops/building/installing/mac-osx/) already imply 2.1.1 will be installed by Homebrew (but notes this community-maintained package may not be up-to-date).

### DIFF
--- a/Library/Formula/riak.rb
+++ b/Library/Formula/riak.rb
@@ -2,9 +2,9 @@ require "formula"
 
 class Riak < Formula
   homepage "http://basho.com/riak/"
-  url "https://s3.amazonaws.com/downloads.basho.com/riak/2.0/2.0.4/osx/10.8/riak-2.0.4-OSX-x86_64.tar.gz"
-  version "2.0.4"
-  sha256 "023627c038833765141f6af0006622dd76e3b9f42495ea796f24e40fc76edba0"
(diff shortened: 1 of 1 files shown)


asparagui:
Looks like you're going to need to add:

    depends_on "openssl"

L2G:
If I could be excused for kibitzing here… Maybe this formula needs to be moved to the homebrew/binary tap, if it's not building from source.

DomT4:
Compile looks like a fairly simple affair. Surprised we're not building from source here.

tdsmith:
Historically: bb5bead0b8

L2G:
(So to build from source, we'd have to depend on a specific version of Erlang; i.e., make a cross-tap dependency on an older version, thus contravening Homebrew policy. Is that right?)

DomT4:
Ah, Erlang fun. If we're forced to use the binary, it's unlikely we can enforce our OpenSSL in any meaningful way. We could add the dep and then env script it so every time the binary is run our OpenSSL is thrown to the front of the `$PATH` but otherwise :no_good:.

jacknagel:
That wouldn't have any effect on shared library linkage.

DomT4:
Fair point.

PR #39563, opened 2015-05-09
openal-soft: fix build scooping up undeclared dependencies

### LINKED ISSUE #38274: OpenAL-soft does not build on OSX Yosemite 10.10.2
See https://gist.github.com/59e3c48f9c973c9cecbf

DomT4:
Closes #38274.

### DIFF
--- a/Library/Formula/openal-soft.rb
+++ b/Library/Formula/openal-soft.rb
@@ -1,9 +1,7 @@
-require "formula"
-
 class OpenalSoft < Formula
   homepage "http://kcat.strangesoft.net/openal.html"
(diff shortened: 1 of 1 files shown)


DomT4:
It's not technically necessary to regenerate bottles here, as we're not changing anything that would or wouldn't already appear in the bottle, but it'll be good to do so anyway so Yosemite has a bottle for the formula.

xu-cheng:
Thanks @DomT4 :ship:ed with :baby_bottle:

DomT4:
Thanks Xu :fireworks:

PR #39564, opened 2015-05-09
Blockhash 0.1 (new formula)

bobobo1618:
Created new formula for the C implementation of http://blockhash.io, a
perceptual image hash library and command line tool.

### DIFF
--- a/Library/Formula/blockhash.rb
+++ b/Library/Formula/blockhash.rb
@@ -1,0 +1,29 @@
+class Blockhash < Formula
+  homepage "http://blockhash.io/"
+  head "https://github.com/commonsmachinery/blockhash.git"
+  url "https://github.com/commonsmachinery/blockhash/archive/0.1.tar.gz"
+  sha256 "aef300f39be2cbc1b508af15d7ddb5b851b671b27680d8b7ab1d043cc0369893"
+
+  depends_on "pkg-config" => :build
+  depends_on "imagemagick"
+
+  resource "testdata" do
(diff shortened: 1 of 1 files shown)


bobobo1618:
Sorry if the test isn't enough. I'm not sure where to get a decent image to hash and if the time it takes to hash it would be considered too long for a test.

Library/Formula/blockhash.rb
```diff
@@ -0,0 +1,19 @@
+class Blockhash < Formula
+  homepage "http://blockhash.io/"
+  head "https://github.com/commonsmachinery/blockhash.git"
+  url "https://github.com/commonsmachinery/blockhash/archive/0.1.tar.gz"
+  sha256 "aef300f39be2cbc1b508af15d7ddb5b851b671b27680d8b7ab1d043cc0369893"
+
+  depends_on 'pkg-config' => :build
```
tdsmith: nit: double quotes, please

tdsmith:
You could download something from https://github.com/commonsmachinery/blockhash/tree/master/testdata as a `resource` in the test block.

tdsmith:
`unarj.rb` has an example of the resource test pattern I mentioned above ^

bobobo1618:
Added a new test conforming roughly to that pattern (I'm using Popen and comparing the output since Blockhash doesn't return a meaningful response code) that seems to work.

tdsmith:
Pulled, with a couple tweaks to the test. Thanks for the pull request!

PR #39567, opened 2015-05-10
jbig2dec: add upstream libpng patch

DomT4:
Upstream fixed the bug report I filed on linking into now private `libpng` headers :tada:.


xu-cheng:
Thanks @DomT4

PR #39574, opened 2015-05-10
rtl-sdr 0.5.3: migrate from head-only (releases now provided on GitHub)

L2G:
From the [rtl-sdr home page](http://sdr.osmocom.org/trac/wiki/rtl-sdr):

> The rtl-sdr code can be checked out with:
> 
>     git clone git://git.osmocom.org/rtl-sdr.git
>
> It can also be browsed via [​cgit](http://cgit.osmocom.org/cgit/rtl-sdr/), and there's an official [​mirror on github](https://github.com/steve-m/librtlsdr) that also provides ​[packaged releases](https://github.com/steve-m/librtlsdr/releases).

In light of the mirror's "official" status and its holding of versioned releases, I thought maybe this formula could graduate to core.

### DIFF
--- a/Library/Formula/rtl-sdr.rb
+++ b/Library/Formula/rtl-sdr.rb
@@ -1,0 +1,22 @@
+class RtlSdr < Formula
+  homepage "http://sdr.osmocom.org/trac/wiki/rtl-sdr"
+  url "https://github.com/steve-m/librtlsdr/archive/v0.5.3.tar.gz"
+  sha256 "98fb5c34ac94d6f2235a0bb41a08f8bed7949e1d1b91ea57a7c1110191ea58de"
+  head "git://git.osmocom.org/rtl-sdr.git"
+
+  depends_on "cmake" => :build
+  depends_on "pkg-config" => :build
+  depends_on "libusb"
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    assert_match(
+      %r{^rtl_sdr, an I/Q recorder for RTL2832 }m,
+      shell_output("#{bin}/rtl_sdr 2>&1", 1)
+    )
+  end
+end


L2G:
Aware of and working on the buildbot errors.

L2G:
This is affected by #39620.

L2G:
Also, there may be some CMake complications here. It all seems to build fine on my Mavericks system, and libusb certainly looks like it's 1.0.x.

jacknagel:
This probably needs a `depends_on "pkg-config" => :build`.

Library/Formula/rtl-sdr.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    assert_match(
+      %r{^rtl_sdr, an I/Q recorder for RTL2832 based DVB-T receivers$}m,
+      shell_output("#{bin}/rtl_sdr 2>&1", 1),
```
jacknagel: I disable the rubocop rule that asked for this trailing comma, because it's a syntax error on Ruby 1.8.

jacknagel:
Should be past the rubocop weirdness now.

L2G:
```
CMake Error at CMakeLists.txt:69 (message):
  LibUSB 1.0 required to compile rtl-sdr
```

Since libusb is already 1.0, I'm not sure what to do about this.

DomT4:
Which build is that occurring in? The latest push turned blue/green on Yosemite and Mavericks bots, Mountain Lion still threw the syntax failure on the test block.

L2G:
I just added `depends_on "pkg-config" => :build`, which I think helped.

I have a hunch there is something going on with superenv not fully sanitizing the environment of references to pkg-config.

Library/Formula/rtl-sdr.rb
```diff
@@ -0,0 +1,19 @@
...
+  depends_on "pkg-config" => :build
+  depends_on "libusb"
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    assert_match %r{^rtl_sdr, an I/Q recorder for RTL2832 based DVB-T receivers$}m, shell_output("#{bin}/rtl_sdr 2>&1", 1)
```
mikemcquaid: Do a shorter match, don't need to match full lines like this. Also would be good to reduce the length of this line so it can wrap at 80 chars.

Library/Formula/rtl-sdr.rb
```diff
@@ -0,0 +1,19 @@
+class RtlSdr < Formula
+  homepage "http://sdr.osmocom.org/trac/wiki/rtl-sdr"
+  url "https://github.com/steve-m/librtlsdr/archive/v0.5.3.tar.gz"
```
mikemcquaid: Hmm, this contradicts `head` below?

DomT4:
Couple of comments here that need addressing from Mike, but otherwise looks great.

Library/Formula/rtl-sdr.rb
L2G: The official code is the HEAD url, but they do their tagged releases on their official GitHub mirror. I know it's weird. Tell them. :smile_cat:

L2G:
Closing the PR since this has been added as `librtlsdr` with `rtl-sdr` as an alias.

PR #39575, opened 2015-05-10
gsettings-desktop-schemas 3.16.2

tschoonj:
version bump


xu-cheng:

PR #39576, opened 2015-05-10
Update libdsk formula to 1.3.8

damieng:


xu-cheng:

PR #39577, opened 2015-05-10
adwaita-icon-theme 3.16.0

tschoonj:
version bump
renamed gnome-icon-theme to adwaita-icon-theme
aliases updated correspondingly

### DIFF
--- a/Library/Formula/gnome-icon-theme.rb
+++ b/Library/Formula/gnome-icon-theme.rb
@@ -1,7 +1,7 @@
 class GnomeIconTheme < Formula
   homepage "https://developer.gnome.org"
-  url "http://ftp.gnome.org/pub/GNOME/sources/adwaita-icon-theme/3.14/adwaita-icon-theme-3.14.1.tar.xz"
-  sha1 "e1d603d9cc4e4b7f83f749ba20934832d4321dd2"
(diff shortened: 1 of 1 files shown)


xu-cheng:
Sorry, but for technical reason we don't support formula rename. So please leave the formula name as it is. Thanks.

tschoonj:
Not sure what you mean here... everything appears to work fine.

Either way: it doesn't make sense to keep the name gnome-icon-theme around since it's a dead project that has been completely superseded by adwaita-icon-theme.

xu-cheng:
Formula rename would cause problem for users who already have it installed under the other name. Sorry! Here is a old reference #14374

tschoonj:
ok, taken care of...

DomT4:
There's an `alias` in place already - Anyone who wants to do `brew install adwaita-icon-theme` will get the expected end result, but at the same time, not renaming it means everyone who currently has `gnome-icon-theme` installed doesn't suddenly stop getting updates for the formula without any indication of what/why.

tschoonj:
I see now why it wouldn't work. I just thought that renaming `gnome-icon-theme.rb` to `adwaita-icon-theme.rb`, which simultaneously deleting the old `adwaita-icon-theme` alias and creating a new `gnome-icon-theme` alias that points to `awaita-icon-theme.rb` would do the trick... 

Would be nice if something like `replaces formula` could be added at some point that deals with this kind of situations...

tschoonj:
by the way: the current build is fine and doesn't do the renaming...

PR #39579, opened 2015-05-10
logrotate 3.9.1

guoxiao:


xu-cheng:
Why the patch is changed? Have you submit the new patch to upstream?

guoxiao:
The patch is basically the same, I only resolved the conflicts.

PR #39581, opened 2015-05-10
json-glib 1.0.4

tschoonj:
version bump
brew audit --strict compliance

### DIFF
--- a/Library/Formula/json-glib.rb
+++ b/Library/Formula/json-glib.rb
@@ -1,9 +1,7 @@
-require "formula"
-
 class JsonGlib < Formula
-  homepage "http://live.gnome.org/JsonGlib"
-  url "http://ftp.gnome.org/pub/gnome/sources/json-glib/1.0/json-glib-1.0.2.tar.xz"
-  sha256 "887bd192da8f5edc53b490ec51bf3ffebd958a671f5963e4f3af32c22e35660a"
+  homepage "https://live.gnome.org/JsonGlib"
+  url "http://ftp.gnome.org/pub/gnome/sources/json-glib/1.0/json-glib-1.0.4.tar.xz"
+  sha256 "80f3593cb6bd13f1465828e46a9f740e2e9bd3cd2257889442b3e62bd6de05cd"
 
   bottle do
     revision 2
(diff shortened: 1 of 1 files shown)


Library/Formula/json-glib.rb
```diff
@@ -21,6 +19,18 @@ def install
                           "--disable-dependency-tracking",
                           "--prefix=#{prefix}",
                           "--enable-introspection=yes"
-    system "make install"
+    system "make", "install"
+  end
```
xu-cheng: Please leave a line between the install and test.

Library/Formula/json-glib.rb
```diff
@@ -21,6 +19,18 @@ def install
...
+  test do
+    (testpath/"test.c").write <<-EOS.undent
+      #include <json-glib/json-glib.h>
+
+      int main(int argc, char *argv[]) {
+        JsonParser *parser = json_parser_new();
+        return 0;
+      }
+    EOS
+    system ENV.cc, "-I#{HOMEBREW_PREFIX}/include/json-glib-1.0", "-I#{HOMEBREW_PREFIX}/include/glib-2.0", "-I#{HOMEBREW_PREFIX}/lib/glib-2.0/include", "-I#{HOMEBREW_PREFIX}/opt/gettext/include", "test.c", "-L#{HOMEBREW_PREFIX}/lib", "-L#{HOMEBREW_PREFIX}/lib", "-L#{HOMEBREW_PREFIX}/opt/gettext/lib", "-ljson-glib-1.0", "-lgio-2.0", "-lgobject-2.0", "-lglib-2.0", "-lintl", "-o", "test"
```
xu-cheng:
```ruby
glib = Formula["glib"]
system ENV.cc, "test.c", "-I#{include}", "-I#{glib.opt_include}", "-L#{lib}", "-L#{glib.opt_lib}", "-l...", "-o", "test"
```

You don't need add `"-L#{HOMEBREW_PREFIX}/lib", "-L#{HOMEBREW_PREFIX}/opt/gettext/lib"`.

tschoonj:
I added an empty line before the `test do` block. You may want to add that to the `audit --strict` tests.

tschoonj:
Mountain Lion build failed for some reason... could you trigger it again? Thanks!

tschoonj:
@xu-cheng any idea how to fix this? `ENV["CPPFLAGS"]` and/or `ENV["LDFLAGS"]` returns nil causing split to fail. This works fine on my local machine though. Is it even necessary to use these environment variables? I never did so in the past and things worked without a hitch

xu-cheng:
I have updated the tool again to address this.

tschoonj:
ok thanks, json-glib and librsvg have been queued for rebuilding
Since libxml++ didn't suffer from this problem, can you close the pull request?

xu-cheng:
Thanks @tschoonj

PR #39587, opened 2015-05-10
libswiften: use SSL/TLS URLs (and an SHA256 hash)

vszakats:


DomT4:
I don't know how the maintainers feel, but it may be kinder on them to do 4-5 http => https changes in the same PR rather than having 3/4/5 open PRs which cover the same ground.

vszakats:
You're most probably right. That would raise the barrier of entry (and the number of mistakes that can be made) quite a bit from my end, I'm not sure I can follow up on it.

DomT4:
Why the barrier raising? Technical or other concerns?

vszakats:
It'd mean that formulae cannot be tacked anymore _individually_ (one formula may have multiple updates till got right), it'd also make bot failures hard to track, squashing a mess (if possible at all - I'm not sure). It'd also mean that I cannot use GitHub for editing. This one is partially a personal gripe - Mac (*nix) editors, but a roadblock nevertheless. Maybe there is a way to make this smooth, but there are so many things to mess up already, that at this point I don't feel comfortable to add a bunch more to it. Plus, one mistake with a Git command and the whole branch with all the commits has to be flushed and redone. Probably avoidable if I'd be a Git ninja, but I'm not :(

At this moment GitHub + Homebrew-Bot give a perfect view of each formulae under editing and it makes it very easy to see if there are any problems and let them get fixed quickly.

Probably and automated script would do a better job doing these updates.

DomT4:
You wouldn't squash in this case. You'd do something like this:

* Update formula one
* git add -A && git commit -m "formula: use ssl/tls links"
* Update formula two
* git add -A && git commit -m "formula: use ssl/tls links"
* Update formula three
* git add -A && git commit -m "formula: use ssl/tls links"
* git push yourrepo yourbranch

If you made a small mistake or something, you'd do:

* git rebase -i HEAD~(Number of commits here)
* change `pick` to `edit` on the commit intended
* Make the change desired
* git add -A && git commit -m "squashing"
* git rebase --continue
* git rebase -i HEAD~(Number of commits here)
* change `pick` to `squash` on the newly created fixup commit
* Squash the commit message that pops up
* git push yourbranch -f

> This one is partially a personal gripe - Mac (*nix) editors, but a roadblock nevertheless.

Github's [Atom](https://atom.io/) is pretty great. Free software, as well. Clipping it as the default editor means you wouldn't need to deal with `vim` or `nano` to make the changes. It can pretty much all be GUI.

If you're happy though, feel free to keep doing it this way. Just ping if you want any help with Git stuff though, or ping Mike if you want help with Git stuff in a more expert way than I'll ever be able to deliver :smile_cat:.

vszakats:
Thanks for these Git details. I do use most of these when I _have to_ ;), but most often I can avoid them and it makes for a smoother workflow. (saying this as a command-line person.)

Now editing is just clicking on the edit icon on github.com, making the edit, writing the commit msg and pressing `Commit`. It's as if it was _invented_ to make single file commits like with Homebrew formulae.
For URL checks, hashes and `brew audit` I go back to CLI of course.

The other thing in grouping is that it needs careful advance planning _what_ to group, if you want to stay sane while doing it. I can see it being useful when doing _truly_ robotic updates, f.e. s&r-ing the same URL in multiple formulae, but such mass editing is probably discouraged by maintainers.

I tried Atom when the beta came out, be giving it another go with the latest, there's the hope it'll feel less heavy :) I've seen Mike's "Advanced Git" talk, it was great indeed.

PR #39600, opened 2015-05-10
http-parser 2.5.0

guoxiao:


guoxiao:
@jacknagel any problem with the PR?

Library/Formula/http-parser.rb
```diff
@@ -0,0 +1,18 @@
+class HttpParser < Formula
+  homepage "https://github.com/joyent/http-parser"
+  url "https://github.com/joyent/http-parser/archive/v2.5.0.tar.gz"
+  sha256 "e3b4ba58f4e6ee5fbec781df020e5cb74c3a799a07f059e1e125127a0b801481"
+
+  def install
+    system "make", "library", "SONAME=libhttp_parser.dylib"
+    include.install "http_parser.h"
```
mikemcquaid: Is there no `make install` we can use?

guoxiao: The `make install` is only suitable for Linux. It creates a `.so.2.5.0` and soft link to `.so`, and makes `brew audit --strict` fail.

mikemcquaid: Could you submit a pull-request to make it cross-platform and add that as a patch to this formula? Thanks!

guoxiao:
@mikemcquaid Sorry. I found `make install` can work under os x, so I've updated the formula.
I found that libraries can also be `.so` in os x, so it's not necessary to replace `.so` with `.dylib`. Just put the version number before `.so` and `brew audit --strict` doesn't throw a error.

mikemcquaid:
@guoxiao Libraries should be `.dylib` on OS X.

guoxiao:
@mikemcquaid I found some explanations about this http://www.finkproject.org/doc/porting/porting.en.html#shared.lib-and-mod
> Mach-O shared libraries have the file type MH_DYLIB and carry the extension .dylib. They can be linked against with the usual static linker flags, e.g. -lfoo for libfoo.dylib. However, they can not be loaded as a module. (Side note: Shared libraries can be loaded dynamically through an API. However, that API is different from the API for bundles and the semantics make it useless for an dlopen() emulation. Most notably, shared libraries can not be unloaded.)

> Loadable modules are called "bundles" in Mach-O speak. They have the file type MH_BUNDLE. Since no component involved cares about it, they can carry any extension. The extension .bundle is recommended by Apple, but most ported software uses .so for the sake of compatibility. Bundles can be dynamically loaded and unloaded via dyld APIs, and there is a wrapper that emulates dlopen() on top of that API. It is not possible to link against bundles as if they were shared libraries. However, it is possible that a bundle is linked against real shared libraries; those will be loaded automatically when the bundle is loaded.

So i think use `.so` in ported software is also good, especially when it need to be used as loadable modules.
And I searched in my `/usr/local/lib`, there are 82 `.so` files installed by homebrew. So I think it may be also accepted?

mikemcquaid:
Are these loadable modules rather than dynamic libraries?

guoxiao:
@mikemcquaid That's a good question. :D
I found the Makefile checked the darwin platform, so I thought the author had taken into account the platform. I'll try to contact the author and figure out the answer.

guoxiao:
I've submitted a PR at joyent/http-parser#247

guoxiao:
@mikemcquaid my patch has been merged by upstream. Should this PR be accepted now?

Library/Formula/http-parser.rb
```diff
@@ -0,0 +1,22 @@
+class HttpParser < Formula
+  homepage "https://github.com/joyent/http-parser"
+  url "https://github.com/joyent/http-parser/archive/v2.5.0.tar.gz"
+  sha256 "e3b4ba58f4e6ee5fbec781df020e5cb74c3a799a07f059e1e125127a0b801481"
+
+  depends_on "coreutils" => :build
+  patch do
+    url "https://patch-diff.githubusercontent.com/raw/joyent/http-parser/pull/247.diff"
```
mikemcquaid: Use https://github.com/joyent/http-parser/commit/289622.patch as the new URL.

Library/Formula/http-parser.rb
```diff
@@ -0,0 +1,22 @@
+class HttpParser < Formula
+  homepage "https://github.com/joyent/http-parser"
+  url "https://github.com/joyent/http-parser/archive/v2.5.0.tar.gz"
+  sha256 "e3b4ba58f4e6ee5fbec781df020e5cb74c3a799a07f059e1e125127a0b801481"
+
+  depends_on "coreutils" => :build
+  patch do
```
mikemcquaid: Add a newline before this line.

guoxiao: done

mikemcquaid:

For future reference the preferred commit message format for new formulae is `http-parser 2.5.0 (new formula)`.

PR #39601, opened 2015-05-10
heroku-toolbelt 3.36.4

dunn:
Autoupdates are turned off for Keybase and I think a few other formula so I figure that's the thing to do here too.

There's a "Next Gen" version at <https://github.com/heroku/heroku-cli>, but I don't see a deprecation notice or anything on the 3.x.x repository so I wasn't sure whether to switch over.


xu-cheng:
Thanks @dunn

PR #39602, opened 2015-05-10
Show the android-sdk opt prefix instead of keg path in the logs

1zaman:
I haven't seen any IDE yet that does not follow symlinks, so there doesn't seem to be any valid reason to provide the keg path that will change with each upgrade instead of the static opt prefix. If someone knows of some actual workflow that requires the real path, then we can add a line about that.

### DIFF
--- a/Library/Formula/android-sdk.rb
+++ b/Library/Formula/android-sdk.rb
@@ -83,8 +83,7 @@ class AndroidSdk < Formula
   def caveats; <<-EOS.undent
     Now run the 'android' tool to install the actual SDK stuff.
 
-    The Android-SDK location for IDEs such as Eclipse, IntelliJ etc is:
-      #{prefix}
+    The Android-SDK is available at #{opt_prefix}
 
     You will have to install the platform-tools and docs EVERY time this formula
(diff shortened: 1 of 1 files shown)


Library/Formula/android-sdk.rb
```diff
@@ -83,8 +83,7 @@ def install
   def caveats; <<-EOS.undent
     Now run the 'android' tool to install the actual SDK stuff.
 
-    The Android-SDK location for IDEs such as Eclipse, IntelliJ etc is:
-      #{prefix}
+    The Android-SDK is available at #{prefix}
```
DomT4: You don't seem to have changed the actual `prefix` argument here?...

1zaman: Sorry, last minute editing mistake. I'll update it shortly.

1zaman:
Fixed email as well.

1zaman:
Sorry, forgot to `git add`. Updated email as well.

1zaman:
Sorry, forgot to `git add`. Updated email as well.

PR #39608, opened 2015-05-10
riak: migrate to homebrew/binary tap

L2G:
This is simply for migrating the riak formula to homebrew/binary to be with the other non-source-building formulae. (See also Homebrew/homebrew-binary#217.)


jacknagel:
Given the relative popularity of this formula, I think this is an exception to the rule.

We've known about it for a long time and not acted so I think it's fine as-is.

L2G:
What's homebrew/binary for, then?

jacknagel:
Like I said, I think this is an exception to the rule.

The issue is not that this is a binary-only install, it's that the source install complicates things for us so we've chosen to use the binary for simplicity.

L2G:
Would it be okay at least to add a note in the homebrew/binary tap so that people don't make my mistake and try to migrate it there in the future?

jacknagel:
I don't know if it makes sense to list out exceptions (or if it's even possible to enumerate all of them). There are not hard and fast rules that one can use to predict with certainty where a formula will end up. Besides applying the technical criteria (e.g. source or binary), we have to consider package popularity and community needs. These are things that can change over time, so I'm not sure that trying to maintain a list of exceptions is the best idea.

L2G:
You have to understand that from the perspective of a person trying very hard to do the right thing and make helpful changes to Homebrew, the seeming arbitrariness and lack of codification of exceptions like this is an extremely frustrating thing.

L2G:
And really I am trying very hard not to step on others' toes, and more importantly, try to help people in the future avoid the toes I'm stepping on.

PR #34165, opened 2014-11-13
src 0.7

bfontaine:
[src](http://www.catb.org/~esr/src/) is a simple revision control system “with a modern UI for single-file solo projects kept possibly more than one to a directory”. This is a new RCS-like by Eric S. Raymond, still in its early stage (first release less than a week ago).

It has been on [HN](https://news.ycombinator.com/item?id=8601727)’s homepage for a couple hours now.

PR #39619, opened 2015-05-11
reposurgeon 3.21

dunn:


dunn:
Failures across the board, but the test worked locally—not sure what the difference is.

Library/Formula/reposurgeon.rb
```diff
@@ -1,18 +1,39 @@
...
+
+  test do
+    (testpath/".gitconfig").write <<-EOS.undent
+      [user]
+        name = Real Person
+        email = notacat@hotmail.cat
+      EOS
+    system "git", "init"
+
+    require "open3"
```
xu-cheng: Let's try to avoid using `open3`. Instead you can use `pipe_output`

dunn: I don't think it allows commands to be piped in.  When I try `assert_match "found git repository", pipe_output("#{bin}/reposurgeon", "read")` I get this:
```
🐙  brew test -v reposurgeon
Testing reposurgeon
==> git init
Initialized empty Git repository in /private/tmp/reposurgeon20150510-795-1xv6t0v/.git/
==> /usr/local/Cellar/reposurgeon/3.21/bin/reposurgeon
reposurgeon: fatal exception in popen_or_die.
reposurgeon:
```

xu-cheng: That's weird. `pipe_output` uses `IO.popen` which it's not that different from `open3`.

Can you try this `pipe_output("script -q /dev/null #{bin}/reposurgeon", "...")`?

dunn: That fails to complete, as does this:
```ruby
    IO.popen("#{bin}/reposurgeon") do |pipe|
      pipe.puts "read\n"
      pipe.close_write
      assert_match "found git repository", pipe.get
    end
```

xu-cheng: fair enough.

dunn: (`pipe.get` should have been `pipe.gets`, but that doesn't change the result here.)

dunn:
looks like this is the same situation as https://github.com/Homebrew/homebrew/pull/34165#discussion_r22342214

dunn:
(for reference, the new test failure: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/25967/version=mountain_lion/testReport/junit/brew-test-bot/mountain_lion/test_reposurgeon/)

xu-cheng:
Thanks @dunn

PR #39625, opened 2015-05-11
percona-server 5.6.24-72.2

nijikon:


xu-cheng:

PR #36926, opened 2015-02-18
gcc 5.1, ISL 0.14.1, CLooG 0.18.3

Noctem:
* ISL
  * Update to 0.14.1
  * Remove obsolete comment
* CLooG
  * Update to 0.18.3
  * Remove outdated mirror
  * Modernize formula
* GCC
  * Update to 5.1.0
  * Add HEAD
  * Add option and patch for JIT
  * Remove CLooG dependency
  * Refactor language selection
* mpich
  * Bump revision
  * Use SHA256
* gnu-cobol
  * Bump revision
  * Use SHA256
  * Remove `require "formula"`

PR #39626, opened 2015-05-11
test-bot: bugs fix

xu-cheng:
Fixes the bugs revealed in https://github.com/Homebrew/homebrew/pull/36926

In long term, I plan to rewrite test-bot and split it into multiple files. So it's easier to maintain. (There are just  too many monkey patches).

### DIFF
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -386,7 +386,7 @@ module Homebrew
       formula.conflicts.map { |c| Formulary.factory(c.name) }.
         select { |f| f.installed? }.each do |conflict|
           test "brew", "unlink", conflict.name
-      end
+        end
 
       installed_gcc = false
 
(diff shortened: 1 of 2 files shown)


xu-cheng:
cc @mikemcquaid

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -406,7 +412,8 @@ def formula formula_name
         CompilerSelector.select_for(formula)
       rescue CompilerSelectionError => e
         unless installed_gcc
-          test "brew", "install", "gcc"
+          # Don't care about e.g. bottle failures
+          run_as_not_developer { test "brew", "install", "gcc" }
```
mikemcquaid: I'd say the actual fix here is to only do this if the `gcc` formula has been changed?

mikemcquaid:
> In long term, I plan to rewrite test-bot and split it into multiple files. So it's easier to maintain. (There are just too many monkey patches).

I don't think splitting into multiple files makes it easier to maintain, personally. I'm up for cleaning this up further, though. What monkey patches?

Library/Homebrew/cmd/test-bot.rb
xu-cheng: You are right.

xu-cheng:
I'm just thinking to split different commands into multiple files or methods (like ci-upload and test are irrelevant to each other).

> What monkey patches?

Some duplicated codes can be combined into method. And cleanup some variables.

mikemcquaid:
> Some duplicated codes can be combined into method. And cleanup some variables.

Ok, sorry, I've always understood monkey patching as a Ruby-specific thing where you modify the behaviour of existing classes.

xu-cheng:
Updated.

mikemcquaid:
:+1:

Issue #39584, opened 2015-05-10
Cleanup gtk relatived formulae test


xu-cheng:
The tests of below formulae require serious fix. There are lots of redundant/incorrect arguments.

* [x] `gtk`
* [x] `gtk3`
* [x] `gtkmm`
* [x] `gtkmm3`
* [x] `glibmm`
* [x] `cairomm`
* [x] `atk`

@DomT4 can you help to fix it? (I would do it, but I'm not familiar to Gtk/x11) Thanks.

xu-cheng:
A [tool](https://gist.github.com/xu-cheng/04a353ad7b5667e4f4fe) I just write should be help to cleanup these code.

xu-cheng:
@tschoonj If possible, can you update these formulae as well? Thanks.

tschoonj:
I will take care of it tomorrow or so. Thanks again for writing this great tool!

DomT4:
Let me know if you still need any help, but if @tschoonj wants it I'm happy to not step on his toes.

tschoonj:
all 7 packages have been updated but I expect them to fail building properly on the bot due to the same issue as in #39581

tschoonj:
All are done.

Found one more problem with the script, mentioned it in #39578

PR #39578, opened 2015-05-10
libxml++ 2.38.0

tschoonj:
version bump
audit --strict compliance

PR #38969, opened 2015-04-23
without-x11 option removed in cairo, cairomm, pango and pangomm

tschoonj:
@mikemcquaid @jacknagel 

Ok, this is a big one...

I went through the complete gtkmm (2 and 3) dependency graph this morning and see if something could be done about the x11/quartz issue. First thing I figured out is that it is not necessary for cairo(mm) and pango(mm) to have the --without-x11 option. The best thing to do is just build cairo and pango with all possible backends, which creates no internal conflicts at all. This does not impact the gtk+ package: whether or not it installed with or without X11 support, the cairo and pango packages do not need rebuilding.

pangomm and cairomm's functionality depends entirely on how respectively pango and cairo were configured. So I just removed the --without-x11 option I erroneously introduced yesterday.

Regarding a separate gtk-quartz tap: this will still be necessary for gtk+ (2 only), gtkmm (2 only), and every package that depends on these two. cairo(mm) and pango(mm) should never be part of this tap.

I also made some changes to gtk+, gtk+3, gtkmm and gtkmm3. I will create separate pull requests for these. Important here will be that gtk+3 no longer supports the X11 backend (change introduced in 3.16.0), due to changes in the libepoxy dependency.

PR #39628, opened 2015-05-11
gtk+ test cleanup

tschoonj:
Part of the tasks set out in #39584

### DIFF
--- a/Library/Formula/gtk+.rb
+++ b/Library/Formula/gtk+.rb
@@ -36,6 +36,7 @@ class Gtkx < Formula
     system "./configure", *args
     system "make", "install"
   end
+
   test do
     (testpath/"test.c").write <<-EOS.undent
       #include <gtk/gtk.h>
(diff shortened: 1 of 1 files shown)


Library/Formula/gtk+.rb
```diff
@@ -45,7 +46,60 @@ def install
...
+    ENV.prepend "PKG_CONFIG_PATH", OS::Mac::X11.lib/"pkgconfig"
```
xu-cheng: This has no effect. It's not used by any logic in below. Can you try to remove it and test whether it works without this.

tschoonj:
No, the line is essential: otherwise it won't find the X11 headers.

It's basically the equivalent of ENV.x11 but that won't pass `audit --strict`

I made a comment about this in #39578

xu-cheng:
You never used `PKG_CONFIG_PATH` variable as you never run `pkg-config` in your test.

tschoonj:
Good point! In that case the ENV.x11 is completely redundant too.

xu-cheng:
You can go to have a try. Otherwise, it means it should `depends_on :x11`

tschoonj:
Ok removed the line both in gtk+ and gtkmm

By the way: you cannot add `depends_on` inside of test blocks: run time error.

xu-cheng:
> By the way: you cannot add depends_on inside of test blocks: run time error.

Actually I mean this test depends on x11. So either you find another test which doesn't depend on x11, or you should add `depends_on :x11` into the dependencies list. This will also mean this formula should be moved to `homebrew/x11`.

tschoonj:
Well I think that we'll soon make gtk+, pango and cairo quartz only (#38969) to end the nightmare that is supporting both x11 and quartz backends for gtk+. gtk3 doesn't even support x11 anymore starting with version 3.16.0.

DomT4:
Moving GTK+ to Homebrew/x11 will be mayhem; so much depends on it it's not really feasible as an outright move. `depends_on :x11` is already recommended in this formula, and most `test do` blocks only support the default formula compile and settings, so relying on `x11` to be present for the test isn't a major sin for now. The end goal is to remove the x11 backend, but till then :neutral_face:.

Issue #39609, opened 2015-05-10
Updated LLVM 3.6.0 ignores "--with-rtti" option 


sblessing:
The new ```llvm 3.6.0``` formula recently merged is ignoring the option ```--with-rtti```, consequently causing link-errors for potential dependencies

```
$ brew install -v llvm --with-rtti
[...]
Building CXX object lib/LTO/CMakeFiles/LLVMLTO.dir/LTOCodeGenerator.cpp.o
cd /tmp/llvm20150511-23271-jeimnn/lib/LTO && /usr/local/Library/ENV/4.3/clang++   [...] *-fno-rtti* [...]
```

```--with-rtti``` works with the ```llvm36``` formula from ```homebrew/versions```. The main differences between the merged ```llvm``` formula from #37260 is that it uses ```cmake```, rather than a ```configure``` script.

DomT4:
The full `brew gist-logs` will be more helpful than the sole snippet. You could also try the LLVM36 in Homebrew/versions which is currently still using an autotool compile instead of a CMAKE compile and see whether the error reproduces.

sblessing:
The issue is exactly ```cmake``` vs ```configure```. I will submit a patch with #39192.

sblessing:
Fixed and tested (61baf1be3f53d0452d1666b0cef80dde3e03fca4). This can be closed with merging #39192.

DomT4:
Thanks!

PR #39629, opened 2015-05-11
fix for #39609: llvm formula ignored --with-rtti option

sblessing:


tamird:
did you mean to remove the old `ENV["REQUIRES_RTTI"] = "1" if build.with?("rtti") || build.with?("clang")`?

sblessing:
Yes sorry. I must have forgotten removing that in the second pull request. Thanks for merging this.

PR #39636, opened 2015-05-11
Fix homepage for evas

hanxue:
- Old subdomain no longer in use. Updated to correct URL

### DIFF
--- a/Library/Formula/evas.rb
+++ b/Library/Formula/evas.rb
@@ -1,7 +1,7 @@
 require "formula"
 
 class Evas < Formula
-  homepage "http://trac.enlightenment.org/e/wiki/Evas"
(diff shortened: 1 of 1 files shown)


Library/Formula/evas.rb
```diff
@@ -1,7 +1,7 @@
 require "formula"
 
 class Evas < Formula
-  homepage "http://trac.enlightenment.org/e/wiki/Evas"
+  homepage "https://docs.enlightenment.org/auto/eet/evas_main.html"
   url "http://download.enlightenment.org/releases/evas-1.7.10.tar.gz"
```
vszakats: Would you update `url` to use `https` as well?:
https://download.enlightenment.org/releases/evas-1.7.10.tar.gz

PR #39646, opened 2015-05-11
Add `ccat` formula

jingweno:
Project: https://github.com/jingweno/ccat

### DIFF
--- a/Library/Formula/ccat.rb
+++ b/Library/Formula/ccat.rb
@@ -1,0 +1,20 @@
+class Ccat < Formula
+  homepage "https://github.com/jingweno/ccat"
+  url "https://github.com/jingweno/ccat/archive/v0.1.0.tar.gz"
+  sha256 "437c7582ab39c8a1fb5e97ae9569e0096c98209a2ebb11ca67a7f9246b129c7c"
+
+  depends_on "go" => :build
+
+  def install
+    system "./script/build"
+    bin.install "ccat"
+  end
+
+  test do
+    (testpath/"test.txt").write <<-EOS.undent
+      I am a colourful cat
+    EOS
+
+    assert_match(/I am a colourful cat/, shell_output("#{bin}/ccat test.txt"))
+  end
+end


Library/Formula/ccat.rb
```diff
@@ -0,0 +1,14 @@
+require "formula"
```
DomT4: This hasn't been required for a while. Can :scissors: it.

Library/Formula/ccat.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class Ccat < Formula
+  homepage "https://github.com/jingweno/ccat"
+  url "https://github.com/jingweno/ccat/archive/v0.1.0.tar.gz"
+  sha1 "fc0a1a49ca81fc85b9d50f48d63182d507b4627c"
```
DomT4: This should be `sha256`.

Library/Formula/ccat.rb
jingweno: Done (`--amend`)

Library/Formula/ccat.rb
jingweno: Didn't know we're using `sha256` forward. How about existing formulas? Any plan to fix all of them?

jingweno: Fixed (`--amend`)

jingweno:
@DomT4 BTW, I'd want the build bot to add `bottle` as well. I assume I don't need to do anything?

Library/Formula/ccat.rb
DomT4: Yes. `sha256` is required for all new formulae, all existing `sha1` bottles are being migrated to `sha256` as they are updated and there's a strong ongoing effort to eventually deprecate `sha1` from formulae entirely.

DomT4:
> I assume I don't need to do anything?

Nope. Bot generates bottles automatically.

Library/Formula/ccat.rb
```diff
@@ -0,0 +1,12 @@
...
+  url "https://github.com/jingweno/ccat/archive/v0.1.0.tar.gz"
+  sha256 "437c7582ab39c8a1fb5e97ae9569e0096c98209a2ebb11ca67a7f9246b129c7c"
+
+  depends_on "go" => :build
+
+  def install
+    system "./script/build"
+    bin.install "ccat"
+  end
+end
```
DomT4: If there's anything you can do in regards to a test, that'd be great. Something like:

```ruby
(testpath/"test.txt").write <<-EOS.undent
  I am a colourful cat
EOS

assert_match /I am a colourful cat/, shell_output("#{bin}/ccat test.txt")
```

Should work. Test it first though with `brew test --verbose`, don't just take my word for it :wink:.

jingweno: Yes, that works. I added it (`--amend`). `ccat` is running in non terminal mode in the test so it falls back to work like `cat`.

jingweno:
Build is green. Ready to be merged :)

chdiza:
This is self-submitted, and at version 0.1.

DomT4:
It also has 941 Github stars, which is usually enough for the maintainers, which is why I didn't bother pointing it out earlier. Indifferent though, I'll defer to those with the power to make the decision.

jingweno:
@chdiza why would version 0.1 matter for a formula?

chdiza:
> why would version 0.1 matter for a formula?

[Because](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Acceptable-Formulae.md#niche-or-self-submitted-stuff) the docs say only stable packages are welcome.  A version number that low indicates that it's something very new.

I might also add, for the maintainers, that the whole project is six days old.

jingweno:
> Because the docs say only stable packages are welcome. A version number that low indicates that it's something very new.

`ccat` is pretty stable - just look at the activity of the repo and what people said on Twitter. The reason I didn't mark it as 1.0 is because it doesn't have all features I want. If 1.0 is what people would expect a stable release before starting to use a tool, I could bump it to 1.0 right now. What I'm trying to say is, `ccat` is stable and people have already been using it.

narqo:
:+1: for the merge

mikemcquaid:
This is popular enough that I think we should merge it.

mikemcquaid:

For future reference the preferred commit message format for new formulae is `ccat 0.1.0 (new formula)`.

jingweno:
@mikemcquaid :beers:

PR #39661, opened 2015-05-12
ranger 1.7.1

dunn:
It's an interactive file manager so I couldn't figure out a real test.

### DIFF
--- a/Library/Formula/ranger.rb
+++ b/Library/Formula/ranger.rb
@@ -1,11 +1,9 @@
-require 'formula'
-
 class Ranger < Formula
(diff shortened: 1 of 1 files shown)


jacknagel:
Test failing due to not being connected to a tty :/

xu-cheng:
If possible, `script -q /dev/null <commands>` can be used to fake the tty. Ref: http://stackoverflow.com/questions/1401002/trick-an-application-into-thinking-its-stdin-is-interactive-not-a-pipe

xu-cheng:
Thanks @dunn

PR #39665, opened 2015-05-12
sysdig 0.1.99

dunn:
Looks like cmake is downloading zlib and luajit during the build: https://gist.github.com/dunn/240f50f717d7ad62190d

Should we try to vendor zlib and use brewed luajit?

### DIFF
--- a/Library/Formula/sysdig.rb
+++ b/Library/Formula/sysdig.rb
@@ -1,11 +1,8 @@
-require "formula"
-
 class Sysdig < Formula
   homepage "http://www.sysdig.org/"
-  url "https://github.com/draios/sysdig/archive/0.1.98.tar.gz"
-  sha1 "a184272b9ab34a644027a699e1e7dbb1676b5265"
-
-  head "https://github.com/draios/sysdig.git"
+  url "https://github.com/draios/sysdig/archive/0.1.99.tar.gz"
+  mirror "https://mirrors.kernel.org/debian/pool/main/s/sysdig/sysdig_0.1.99.orig.tar.gz"
+  sha256 "e6f00493feaa42f05720c708ab87b7150b465306dae18f9b97f2c40334e2ca09"
 
   bottle do
     sha1 "a84b9e27e74e38c3f25c1880f0834523c8be1ba8" => :yosemite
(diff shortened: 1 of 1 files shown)


DomT4:
The download is being done from a server that isn't the official dependency upstream download link either. The `sha256` matches at least. Ideally if we can vendor rather than rely on the build script to fetch them, which won't work offline, that'd be great.

dunn:
Turns out it can be built with lua or luajit which is nice because I can't get it to work with the latter.  Also removed the HEAD since it wasn't building either way. 

Opened an issue upstream: https://github.com/draios/sysdig/issues/360

dunn:
Nope, turns out it was building locally by ignoring the given path to Lua and finding LuaJit instead.

Library/Formula/sysdig.rb
```diff
@@ -14,18 +11,37 @@ class Sysdig < Formula
...
     url "https://gist.githubusercontent.com/juniorz/9986999/raw/a3556d7e93fa890a157a33f4233efaf8f5e01a6f/sample.scap"
-    sha1 "0aa3c30b954f9fb0d7320d900d3a103ade6b1cec"
+    sha256 "efe287e651a3deea5e87418d39e0fe1e9dc55c6886af4e952468cd64182ee7ef"
   end
 
   def install
     ENV.libcxx if MacOS.version < :mavericks
 
+    resource("zlib").stage do
+      system "./configure", "--prefix=#{buildpath}/zlib"
```
dunn: The tests pass, but this should probably be in libexec or some permanent location, right?

DomT4: Usually, no. `zlib` is typically only required at build time for the times we do need to vendor it. Is there any specific reason it's refusing to use the system version? Does it give any helpful output if we try?

dunn: TBH I didn't try the system version; I guess I assumed if it was bundled there was a newer version required. But it works for me.

DomT4: That might be the case for older platforms, but I can't find any mention of that upstream and given it passed the bot here I'm happy enough.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/8467c6f2802efbfc497e38c5094eeed3dc128980. Nice work noticing the build script download issue here @dunn, Appreciated :tada:.

PR #39671, opened 2015-05-12
yelp-tools and yelp-xsl new formulas

tschoonj:
needed these packages to autogen glade, which I will update when gtk2
becomes quartz-backend-only and when the new gtk-mac-integration release
is out

### DIFF
--- a/Library/Formula/yelp-tools.rb
+++ b/Library/Formula/yelp-tools.rb
@@ -1,0 +1,23 @@
+class YelpTools < Formula
+  homepage "https://github.com/GNOME/yelp-tools"
+  desc "Tools that help create and edit Mallard or DocBook documentation."
+  url "https://download.gnome.org/sources/yelp-tools/3.16/yelp-tools-3.16.1.tar.xz"
+  sha256 "b4f66c145af1c6448dc51037d305d6844da13dc31d07729b8e29005ee4fef89c"
+
+  depends_on "itstool" => :build
+  depends_on "pkg-config" => :build
+  depends_on "libxslt" => :build
+  depends_on "libxml2" => :build
+  depends_on "yelp-xsl"
+
+  def install
+    system "./configure", "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/yelp-new", "task", "ducksinarow"
+    system "#{bin}/yelp-build", "html", "ducksinarow.page"
+    system "#{bin}/yelp-check", "validate", "ducksinarow.page"
+  end
+end
--- a/Library/Formula/yelp-xsl.rb
+++ b/Library/Formula/yelp-xsl.rb
@@ -1,0 +1,32 @@
+class YelpXsl < Formula
+  homepage "https://github.com/GNOME/yelp-xsl"
+  desc "Yelp's universal stylesheets for Mallard and DocBook"
+  url "https://download.gnome.org/sources/yelp-xsl/3.16/yelp-xsl-3.16.1.tar.xz"
+  sha256 "3295eecc4b03d2a239f7a1bdf4a1311d34c46c3055e6535c1f72bb5a49b4174a"
+
+  depends_on "libxslt"
+  depends_on "libxml2"
+  depends_on "itstool" => :build
+  depends_on "intltool" => :build
+  depends_on "gettext" => :build
+  depends_on "gtk+3"=> :run
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  def post_install
+    system "#{Formula["gtk+3"].opt_bin}/gtk3-update-icon-cache", "-f", "-t", "#{HOMEBREW_PREFIX}/share/icons/hicolor"
+  end
+
+  test do
+    assert (share/"yelp-xsl/xslt/mallard/html/mal2html-links.xsl").exist?
+    assert (share/"yelp-xsl/js/jquery.syntax.brush.smalltalk.js").exist?
+    assert (share/"yelp-xsl/icons/hicolor/24x24/status/yelp-note-warning.png").exist?
+    assert (share/"pkgconfig/yelp-xsl.pc").exist?
+  end
+end


Library/Formula/yelp-tools.rb
```diff
@@ -0,0 +1,19 @@
+class YelpTools < Formula
+  homepage "https://github.com/GNOME/yelp-tools"
+  url "http://ftp.gnome.org/pub/gnome/sources/yelp-tools/3.16/yelp-tools-3.16.1.tar.xz"
```
vszakats: Please use this `url`:
https://download.gnome.org/sources/yelp-tools/3.16/yelp-tools-3.16.1.tar.xz

Library/Formula/yelp-xsl.rb
```diff
@@ -0,0 +1,25 @@
+class YelpXsl < Formula
+  homepage "https://github.com/GNOME/yelp-xsl"
+  url "http://ftp.gnome.org/pub/gnome/sources/yelp-xsl/3.16/yelp-xsl-3.16.1.tar.xz"
```
vszakats: and this:
https://download.gnome.org/sources/yelp-xsl/3.16/yelp-xsl-3.16.1.tar.xz

tschoonj:
ok made the changes.

But why are they necessary? Is ftp.gnome.org being phased out or so?

vszakats:
Thanks. These are the new ("canonical") links gnome.org is linking to [1][2]. It also holds the promise to support SSL/TLS mirrors someday.

[1] https://www.gnome.org/getting-gnome/
[2] https://www.mail-archive.com/gnome-infrastructure%40gnome.org/msg05024.html

tschoonj:
How about merging this one in?

Library/Formula/yelp-tools.rb
```diff
@@ -0,0 +1,19 @@
+class YelpTools < Formula
+  homepage "https://github.com/GNOME/yelp-tools"
```
bfontaine: Could you please add a short description of the formula, using `desc "…"`? Thanks!

Library/Formula/yelp-xsl.rb
```diff
@@ -0,0 +1,25 @@
+class YelpXsl < Formula
+  homepage "https://github.com/GNOME/yelp-xsl"
```
bfontaine: Same here.

bfontaine:

tschoonj:
ok, descriptions added

Library/Formula/yelp-tools.rb
```diff
@@ -0,0 +1,20 @@
...
+  depends_on "pkg-config" => :build
+  depends_on "yelp-xsl" => :build
+
+  def install
+    system "./configure", "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/yelp-check", "--help"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/yelp-tools.rb
```diff
@@ -0,0 +1,20 @@
+class YelpTools < Formula
+  homepage "https://github.com/GNOME/yelp-tools"
+  desc "Tools that help create and edit Mallard or DocBook documentation."
+  url "https://download.gnome.org/sources/yelp-tools/3.16/yelp-tools-3.16.1.tar.xz"
+  sha256 "b4f66c145af1c6448dc51037d305d6844da13dc31d07729b8e29005ee4fef89c"
+
+  depends_on "itstool" => :build
+  depends_on "pkg-config" => :build
+  depends_on "yelp-xsl" => :build
```
mikemcquaid: Are you sure these are all only needed at buildtime? If so, instead of making `yelp-xsl` a separate formula download and install it into the `buildpath` as a `resource`.

Library/Formula/yelp-tools.rb
tschoonj: doubt that's possible: these executables need specifically formatted files to work on, which can be found in gnome projects.

I only added these formulas because I needed them to autogen.sh glade3, for which I intend to submit a formula when the gtk to quartz transition is finished.

mikemcquaid: Would be good to either `Pathname.write` or download as a resource one of those files.

Library/Formula/yelp-tools.rb
tschoonj: I think the yelp-xsl dependency is a runtime one too after looking at macports and the RPM spec files. Will fix this.

I think it's best to keep both packages separate since they are officially two separate projects and everybody else seems to honor this... If someone would ever add yelp (the gnome help viewer), that would have a dependency on yelp-xsl, but not on yelp-tools.

tschoonj:
By the way, according to the macports file for yelp-xsl (https://trac.macports.org/browser/trunk/dports/gnome/yelp-xsl/Portfile), there is a run-time dependency of gtk+3 due to the icon cache needing an update.

Library/Formula/yelp-tools.rb
mikemcquaid: > I think the yelp-xsl dependency is a runtime one too after looking at macports and the RPM spec files. Will fix this.

Any news on this?

Library/Formula/yelp-tools.rb
mikemcquaid: This still needs done.

tschoonj:
Sorry, looks like I had messed up that commit. This should fix things though

mikemcquaid:
@tschoonj Any chance of a better test?

tschoonj:
Not really. `yelp-tools` works on specifically formatted help-files from the gnome project while `yelp-xsl` is a bunch of XSL files and icons.

mikemcquaid:
@tschoonj Seems we should be able to download one of those help-files. I do think it'd be good if perhaps `yelp-xsl` was rolled into another project as it feels weird that anyone would want to install just that.

tschoonj:
As far as I can tell the only other package that depends on `yelp-xsl` is `yelp`, the help-viewer of gnome (which does not depends on `yelp-tools` by the way). No one has bothered to add `yelp` to homebrew for now, but this may change of course. Either way, I would still vote to keep them separate, just because everybody else seems to be doing so.

I will remove the descriptions in the files, they seem to trigger build errors.

I will look into how we can test yelp-build better. Glade needs `yelp-tools` when running autogen so I will look into how its being used there

mikemcquaid:
It probably needs rebased on `master` for the descriptions to work as we'll needed them to be added.

Does `yelp` depend on `yelp-tools`?

tschoonj:
Nope `yelp` depends on `yelp-xsl` (and many other packages).

I will try the rebase, see if it changes things

tschoonj:
I changed the test for `yelp-tools`. It uses the three executables contained in the formula now

tschoonj:
The build still produces the same error, even after rebasing. Shall I just remove the descriptions?

tschoonj:
@mikemcquaid Any idea what caused the build error?

Thanks

mikemcquaid:
I don't really understand what's going on here CC @Homebrew/owners for ideas.

xu-cheng:
Looks like a temporary bot error. I rerun the build, it works fine now.

tschoonj:
@mikemcquaid Would you mind looking at this one again? Thanks!

mikemcquaid:
@BrewTestBot test this please

mikemcquaid:
Thanks @tschoonj! For future reference we prefer only one formula changed per commit (but I fixed this up for you).

PR #39674, opened 2015-05-12
gdk-pixbuf 2.31.4

tschoonj:
version bump
officially I think this is the development version but everybody
(Macports, Ubuntu...) appears
to be using this instead of the 2.30.x series, which hasn't seen an
update for a long time.
audit --strict compliance

### DIFF
--- a/Library/Formula/gdk-pixbuf.rb
+++ b/Library/Formula/gdk-pixbuf.rb
@@ -1,9 +1,7 @@
-require 'formula'
-
 class GdkPixbuf < Formula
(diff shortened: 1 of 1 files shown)


jacknagel:
>officially I think this is the development version but everybody

Please stay on 2.30.x unless there is a compelling reason not to, thanks.

tschoonj:
All I thought was that if it's good enough for MacPorts, Ubuntu Vivid, Debian Jessie and Fedora, it must be good enough for Homebrew too 😀

PR #39676, opened 2015-05-12
add clang-omp support

xu-cheng:


xu-cheng:
cc @Homebrew/owners @Homebrew/science

Library/Homebrew/exceptions.rb
```diff
@@ -191,11 +191,14 @@ def dump
...
 class CompilerSelectionError < RuntimeError
+  attr_reader :formula
+
   def initialize(formula)
+    @formula = formula
     super <<-EOS.undent
       #{formula.name} cannot be built with any available compilers.
       To install this formula, you may need to:
-        brew install gcc
+        brew install #{formula.compiler_standards.include?(:openmp) ? "clang-omp" : "gcc"}
```
jacknagel: I want avoid directly accessing data structures owned by other objects if we can help it; it makes future changes to the implementation more difficult.

Can we do something more like:

```rb
class Formula
  def needs_compiler_feature?(feature)
    active_spec.needs_compiler_feature?(feature)
  end
end


class SoftwareSpec
  def needs_compiler_feature?(feature)
    compiler_standards.include?(feature)
  end
end
```

This way we don't need to expose `compiler_standards` as part of the API.

xu-cheng: Good call :+1:

xu-cheng:
ping @mikemcquaid for comments.

Next step, we should substitute all `compiler == :clang` to `compiler == :clang || compiler == :clang_omp` among the formulae.

mikemcquaid:
Stupid question: why do they need to be separate compilers rather than different e.g versions of the same compiler?

xu-cheng:
Because clang-omp is a third party (by Intel) fork of original clang.

xu-cheng:
Or is it possible to complete this in a less invasive but more hack way? By only substituting clang to clang-omp inside `Library/ENV/4.3/cc`?

mikemcquaid:
If it's a fork that makes a bit more sense but maybe the hacky way might be better?

sjackman:
I prefer the hacky way. On [their web page](https://clang-omp.github.io/) they say that they plan to merge these features into the clang trunk, after which the clang-omp compiler would no longer be necessary.

> We plan eventually to contribute this implementation to the Clang trunk; initial patches have already been committed. In the meantime, you are welcome to try (and contribute to!) the OpenMP 4.0 implementation hosted here.

sjackman:
It's pretty exciting to have OpenMP support for clang for Homebrew-science. There are a number of packages that build with GCC rather than clang simply because of `needs :openmp`. Will these formula now build with "clang-omp" if it's installed?

xu-cheng:
@sjackman Yes, there will be built with `clang-omp` if it's installed. Also, you may want to check #39789 where I reimplement this using hacky approach.

dpo:
I too think this will be excellent! I've already lined up a few formulas to update in prevision of your changes. Many thanks!

sjackman:
Here's a list of formula (https://github.com/Homebrew/homebrew-science/issues/1593) that fail to build with clang. Some of them may be due to `needs :openmp`.

PR #39677, opened 2015-05-12
carthage 0.7.3

robrix:
https://github.com/Carthage/Carthage/releases/tag/0.7.3

### DIFF
--- a/Library/Formula/carthage.rb
+++ b/Library/Formula/carthage.rb
@@ -1,7 +1,7 @@
 class Carthage < Formula
   homepage "https://github.com/Carthage/Carthage"
-  url "https://github.com/Carthage/Carthage.git", :tag => "0.7.2",
-                                                  :revision => "c922316638bae863be697fb930bfef1c6b034ace",
+  url "https://github.com/Carthage/Carthage.git", :tag => "0.7.3",
+                                                  :revision => "ad8fba1a091ae493281ab44953f1e8ca4e6c0405",
                                                   :shallow => false
   head "https://github.com/Carthage/Carthage.git", :shallow => false
 


mojotalantikite:
I just tried installing this version and it looks like the sha of the bottle needs to be updated as well. It still has the old sha from 0.7.2 which causes brew to throw and error and revert to building from source:
```
==> Upgrading 1 outdated package, with result:
carthage 0.7.3
==> Upgrading carthage
==> Downloading https://homebrew.bintray.com/bottles/carthage-0.7.3.yosemite.bottle.tar.gz
######################################################################## 100.0%d
Error: SHA256 mismatch
Expected: 3569536f409798e97fc0321e30e0d1ccd7a9bd503d939fc1ebab03e8cfd0e39a
Actual: 159591c3d4a8727478205a5391d655b52b2c7051aafd3abce179d8fa0ce8a328
Archive: /Library/Caches/Homebrew/carthage-0.7.3.yosemite.bottle.tar.gz
To retry an incomplete download, remove the file above.
==> Cloning https://github.com/Carthage/Carthage.git
```
I updated the sha locally and it worked fine for me.

robrix:
Curious, I didn’t have to update the bottle’s SHA in #39652.

mojotalantikite:
I just tried it again with a clean install (uninstalled, removed cache, installed again) and the same error as above happened. The fix for Carthage/Carthage#449 also doesn't seem to be working for me -- I still get the deadlock after a clean install and 'carthage update' or bootstrap -- so maybe it's something version specific on my system. I'll dig into it a bit.

jacknagel:
The bottles are updated after the pull request is merged. You can use `--build-from-source` in concert with this PR in the meantime.

mojotalantikite:
Ah cool, that makes sense. Thanks!

robrix:
Thanks!

PR #39681, opened 2015-05-12
Bump saltstack to 2015.5

farcaller:


tdsmith:
Please leave the `bottle` block in place and remove the `devel` block (since it's older than the current stable version).

farcaller:
@tdsmith: done. Wasn't sure about bottle sha sums (I suppose they should be different for the new binary, right?). Ok to squash the commits?

tdsmith:
Yep, squashing would be great, thanks! (Our bot helps us regenerate the bottle checksums when we pull the new version.)

farcaller:
Done.

tdsmith:
Thanks!

The M2Crypto dependency was just removed from their `develop` branch but that change was too late to make it into this release, so using an old swig here is still correct.

Library/Formula/saltstack.rb
```diff
@@ -1,7 +1,7 @@
 class Saltstack < Formula
   homepage "http://www.saltstack.org"
-  url "https://github.com/saltstack/salt/archive/v2014.7.1.tar.gz"
-  sha256 "5fcf2cff700d0719b419c9cb489552645ce1287a15c7b3a8745959773d9b0dd1"
+  url "https://github.com/saltstack/salt/archive/v2015.5.0.tar.gz"
+  sha256 "278142ffd4d6ec693a7d160c27d360e9fdce6ae9c7de40e3cb18805078488b71"
   head "https://github.com/saltstack/salt.git", :branch => "develop", :shallow => false
   revision 1
```
xu-cheng: Please remove the revision line.

farcaller:
Done.

xu-cheng:

PR #39683, opened 2015-05-12
llvm: shared libraries are off by default

### LINKED ISSUE #39612: LLVM shared libraries are broken
Since switching to the cmake build, the shared libraries are no longer built by default.

Applying this patch:

```diff
diff --git a/Library/Formula/llvm.rb b/Library/Formula/llvm.rb
index 0d27333..0228033 100644
--- a/Library/Formula/llvm.rb
+++ b/Library/Formula/llvm.rb
@@ -113,8 +114,7 @@ def install
       -DLLVM_OPTIMIZED_TABLEGEN=On
     ]
 
-    args << "-DBUILD_SHARED_LIBS=Off" if build.without? "shared"
-
+    args << "-DBUILD_SHARED_LIBS=On" if build.with? "shared"
     args << "-DLLVM_ENABLE_ASSERTIONS=On" if build.with? "assertions"
 
     mktemp do
```

causes the build to fail.

Upstream says that shared libraries are not recommended, but it would be nice if we could make it work, as an option.

Technically, this is a regression, since we previously packaged the shared libraries. People may depend on this.

tamird:
This changes the configuration of shared libraries to off by default but also fixes the build when shared libraries are selected.

I don't think the bots build with `--with-shared` so someone will have to test this locally. Took about 15m on my machine:
```
==> Summary
🍺  /usr/local/Cellar/llvm/3.6.0: 811 files, 52M, built in 14.9 minutes
```
Closes #39612.

### DIFF
--- a/Library/Formula/llvm.rb
+++ b/Library/Formula/llvm.rb
@@ -70,7 +70,7 @@ class Llvm < Formula
   option "with-lldb", "Build LLDB debugger"
   option "with-rtti", "Build with C++ RTTI"
   option "with-python", "Build Python bindings against Homebrew Python"
-  option "without-shared", "Don't build LLVM as a shared library"
+  option "without-shared", "Don't build LLVM shared libraries"
   option "without-assertions", "Speeds up LLVM, but provides less debug information"
 
   deprecated_option "rtti" => "with-rtti"
@@ -114,7 +114,7 @@ class Llvm < Formula
 
     args << "-DLLVM_ENABLE_RTTI=On" if build.with? "rtti"
 
-    args << "-DBUILD_SHARED_LIBS=Off" if build.without? "shared"
+    args << "-DBUILD_SHARED_LIBS=On" if build.with? "shared"
 
     args << "-DLLVM_ENABLE_ASSERTIONS=On" if build.with? "assertions"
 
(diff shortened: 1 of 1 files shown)


DomT4:
Wasn't Jack concerned this behaviour could be breaking for some people?

> Technically, this is a regression, since we previously packaged the shared libraries. People may depend on this.

Need to add some detail for the patch as well, where it's been reported upstream, why it's necessary, etc. Difficult for people who edit the formulae 2/3/4 months down the line to know what is there and why otherwise.

tamird:
You mean we want to have the shared libraries in the bottle? Sure, can do this.

tamird:
The bug is not reported upstream near as I can tell, and LLVM's contribution process is...involved.

DomT4:
From the wording in the issue I presume they aren't intended to be bottled, but to exist as a functional, working option, which this PR achieves according to your original post. I could be wrong though, Jack also seems to call not packaging the shared libraries a regression, so. I'll let him clarify his desire on this.

> and LLVM's contribution process is...involved.

Honestly, even if you dump the patch and the reason for it in an email to their mailing list and then grab the mailing list archive link and include that in the formula that'll be good enough. If we don't upstream it there's a decent chance we'll be carrying the patch for a long time, sadly.

tamird:
`Your message to llvm-commits awaits moderator approval`

tamird:
@DomT4 updated to include patch comments.

mikemcquaid:

PR #39687, opened 2015-05-12
dirac: avoid compiling with new GCCs (work in progress)

L2G:
dirac is a GNU-style build that hasn't been updated in so long that it may be a lost cause now. But I've made a branch with my attempts to make it work again.

The pull request is mainly for code-reviewing purposes; it is _NOT_ by any means ready to merge at this writing.

### DIFF
--- a/Library/Formula/dirac.rb
+++ b/Library/Formula/dirac.rb
@@ -11,6 +11,10 @@ class Dirac < Formula
     build 2334
   end
 
+  fails_with :clang do
+    cause "build uses compiler flags not known to clang"
+  end
+
   patch :DATA
 
   def install
--- a/Library/Formula/dirac.rb
+++ b/Library/Formula/dirac.rb
@@ -1,20 +1,18 @@
-require 'formula'
-
 class Dirac < Formula
-  homepage 'http://diracvideo.org/'
-  url 'http://diracvideo.org/download/dirac-research/dirac-1.0.2.tar.gz'
-  sha1 '895aaad832a54b754e58f77c87d38c0c37752b0b'
+  homepage "http://diracvideo.org/"
+  url "http://diracvideo.org/download/dirac-research/dirac-1.0.2.tar.gz"
+  sha256 "816b16f18d235ff8ccd40d95fc5b4fad61ae47583e86607932929d70bf1f00fd"
 
   fails_with :llvm do
     build 2334
   end
 
   def install
-    # BSD cp doesn't have '-d'
-    inreplace 'doc/Makefile.in', 'cp -dR', 'cp -R'
+    # BSD cp doesn't have "-d"
+    inreplace "doc/Makefile.in", "cp -dR", "cp -R"
 
     system "./configure", "--disable-debug", "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
-    system "make install"
+    system "make", "install"
   end
 end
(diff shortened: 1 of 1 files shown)


L2G:
I'm getting good results with apple-gcc42. I just need to work on the formula some more to add some form of `fails_with :clang`.

Library/Formula/dirac.rb
```diff
@@ -1,20 +1,39 @@
...
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
 
   fails_with :llvm do
     build 2334
   end
 
+  patch :DATA
```
dunn: Could this be an `inreplace` instead?

L2G: Yes, sorry, I was under the impression that patches were preferred over inreplaces.

L2G: I see the Formula Cookbook recommends using `inreplace` when it's certain or at least very unlikely that a patch would be accepted upstream.

L2G:
I see that the Jenkins build failed with `gcc` installed. On my Mavericks system, it succeeds but with `apple-gcc42` installed. I need to find a way to add a `fails_with :gcc` that still permits the right version(s) of GCC.

jacknagel:
>build uses compiler flags not known to clang

Which flags? We have the ability to [filter out flags in the compiler wrapper](https://github.com/Homebrew/homebrew/blob/716e4dc503e2912e3c492739e31e2b0317a1dfc0/Library/ENV/4.3/cc#L134) in the build environment.

L2G:
One of them was `-nologo`. I'll have to take out the automake/autoconf stuff to refresh my memory about the other one. (Right now the Clang build fails with automake/autoconf: the configure script dies with "Unsupported Compiler version".)

L2G:
Is there a way to specify known, good versions of GCC in the formula (as opposed to known, bad versions)? Now I'm seeing Jenkins give these tests a false "success" because of this:

```
==> SKIPPING: dirac
dirac cannot be built with any available compilers.
  brew install gcc
```

jacknagel:
What do the gcc compilation errors look like? Maybe we can fix them.

mistydemeo:
I was able to fix the gcc-4.9 compilation errors using this patch from their git repo: http://diracvideo.org/git?p=dirac-research.git;a=patch;h=c99cf8d2cb1e9e69b464f4df2dcae9f9baab49dc

mistydemeo:
There have been two commits in their repo since the 1.0.2 release, that one and this: http://diracvideo.org/git?p=dirac-research.git;a=patch;h=88fae6224b54ecf92ac6e3ae5083c3f352adc798

mistydemeo:
clang isn't working for me due to a bad assumption that a compiler beginning with "cl" is an MS compiler:

```sh
dnl -------------------------------------------------------------------------
dnl Check if compiling using MS VC++ compiler in msys environment on Windows
dnl -------------------------------------------------------------------------
use_msvc=no
AC_SUBST([LIBFLAGS],[])
AC_SUBST([LIBEXT],[])
case "$CXX" in
		cl*|CL*)
		CXXFLAGS="-nologo -W1 -EHsc -DWIN32"
		if test x"$enable_shared" = "xyes"; then
		    LIBEXT=".dll";
		    LIBFLAGS="-DLL -INCREMENTAL:NO"
			CXXFLAGS="$CXXFLAGS -D_WINDLL"
		else
		    LIBEXT=".lib";
		    LIBFLAGS="-lib"
		fi
		RANLIB="echo"
		use_msvc=yes
		;;
	*)
		;;
esac
```

mistydemeo:
Actually, it looks like the bad flags are only passed because it thinks it's using a Windows compiler.

mistydemeo:
Fixed - thank you so much for the report and the helpful diagnosis!

L2G:
So the other changes for `brew style` that were also in there…?

mistydemeo:
Applied as b7ca5ecfa80fad4581ce5c5f1d59714375d95772.

L2G:
Whew.

PR #39688, opened 2015-05-12
Update predictionio to latest version

petergarbers:
* Update java version to 1.8

### DIFF
--- a/Library/Formula/predictionio.rb
+++ b/Library/Formula/predictionio.rb
@@ -2,14 +2,14 @@ require 'formula'
 
 class Predictionio < Formula
   homepage 'http://prediction.io/'
-  url 'http://download.prediction.io/PredictionIO-0.9.0.tar.gz'
-  sha1 'e5724b637382a06c098bbb3617b35f5bcff1b469'
+  url 'http://download.prediction.io/PredictionIO-0.9.2.tar.gz'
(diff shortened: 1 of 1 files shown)


petergarbers:
https://docs.prediction.io/install/install-linux/ (The instructions there are for mac and linux)

Library/Formula/predictionio.rb
```diff
@@ -2,14 +2,14 @@
...
-  sha1 'e5724b637382a06c098bbb3617b35f5bcff1b469'
+  url 'http://download.prediction.io/PredictionIO-0.9.2.tar.gz'
+  sha1 '199d1c9a4b69c3bb009ed7819294183e99ced4e9'
 
   depends_on 'elasticsearch'
   depends_on 'hadoop'
   depends_on 'hbase'
   depends_on 'apache-spark'
-  depends_on :java => "1.7"
+  depends_on :java => "1.8"
```
xu-cheng: It cannot work with Java 1.7? If not, it should be ` depends_on :java => "1.7+"`

petergarbers: I don't see a reason why it wouldn't work with java 1.6. I am happy making that changve though

asparagui:
Thanks!

PR #39693, opened 2015-05-12
stlviewer: use sha256

bfontaine:
I couldn’t add a better test because `stlviewer` starts a GUI and doesn’t accept any option.


xu-cheng:
Thanks @bfontaine

PR #39695, opened 2015-05-12
flatbuffers: add git head

rw:


Library/Formula/flatbuffers.rb
```diff
@@ -9,6 +9,10 @@ class Flatbuffers < Formula
     sha1 "a7222fe66033ca2749241e4d253534b3540b1e7c" => :mavericks
     sha1 "d0de7bdea3b5fd43fddcfe20b857e4ec803a73f2" => :mountain_lion
   end
+  
+  head do
```
dunn: If there are no head-specific dependencies/patches/resources this can be simplified to `head "https://github.com/google/flatbuffers.git"`

rw:
Thanks @dunn, updated.

Library/Formula/flatbuffers.rb
```diff
@@ -9,6 +9,8 @@ class Flatbuffers < Formula
     sha1 "a7222fe66033ca2749241e4d253534b3540b1e7c" => :mavericks
     sha1 "d0de7bdea3b5fd43fddcfe20b857e4ec803a73f2" => :mountain_lion
   end
+
```
jacknagel: this line has trailing whitespace.

PR #39702, opened 2015-05-13
packer 0.7.5

mistydemeo:
This replaces the binary formula in homebrew/binary.

### DIFF
--- a/Library/Formula/packer.rb
+++ b/Library/Formula/packer.rb
@@ -1,0 +1,224 @@
+require "language/go"
+
+class Packer < Formula
+  homepage "http://packer.io"
+
+  # buildscript requires the .git directory be present
+  url "https://github.com/mitchellh/packer.git",
+    :tag => "v0.7.5", :revision => "9cd66feeacbd9cb318b72eb5ed59428c5b8c37ac"
+
+  # All of these patches come from the upstream repo and will be in the next release.
+  #
+  # Fixes reference to "BuildID" field in an atlas struct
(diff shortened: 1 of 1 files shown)


Library/Formula/packer.rb
```diff
@@ -0,0 +1,224 @@
+require "language/go"
+
+class Packer < Formula
+  homepage "http://packer.io"
```
vszakats: I'd suggest using the SSL/TLS version:
https://packer.io/

mistydemeo: Updated, thanks!

vszakats: Thank you!

vszakats: Oops, can't see it yet for some reason.

mistydemeo: I didn't update the PR (because I didn't want to regenerate the bottles, which is slow); I amended the commit as I merged. It's in 68a0786d0c5a4848d79d3faae6a1afae488e4250.

vszakats: Oh great, I haven't seen this sort of trick, yet, sorry :)

gpakosz:
Can you explain the rationale for moving it from the `binary` tap to the main formula directory?

mikemcquaid:
We prefer stuff that's built from source so it can be more easily used on platforms where the binary isn't provided.

gpakosz:
Thanks

PR #39704, opened 2015-05-13
haxe: update to 3.2

jdonaldson:
haxe just got bumped to version 3.2

I'm not sure why neko is marked optional here, it's needed for core features (macros).

### DIFF
--- a/Library/Formula/haxe.rb
+++ b/Library/Formula/haxe.rb
@@ -4,9 +4,7 @@ class Haxe < Formula
   homepage "http://haxe.org"
 
   stable do
-    url "https://github.com/HaxeFoundation/haxe.git", :tag => "3.1.3", :revision => "7be30670b2f1f9b6082499c8fb9e23c0a6df6c28"
-    # Remove the below with the next stable release
-    depends_on MaximumMacOSRequirement => :mavericks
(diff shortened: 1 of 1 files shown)


andyli:
For the neko issue, take a look at https://github.com/Homebrew/homebrew/pull/34838
Basically we need a new release of neko ([I'm asking of it](https://github.com/HaxeFoundation/neko/issues/69)) and update the neko formula first.

jdonaldson:
Hmm... ok sorry for causing noise here.  Firstly, the merge failed checks above because for some reason we must specify a tag *and* a revision number.  The latter doesn't seem like it should be necessary, but I can update it.

Secondly, I was able to install neko just fine on yosemite now.  I've completely cleared away my old version, but I'm not sure if I'm missing anything.  Are you still having problems?

As an aside, I have another related pull request for this : https://github.com/HaxeFoundation/neko/pull/45
it should allow for proper /lib location according to brew standards.

It would be great to clean this up, I would much prefer using brew to manage my haxe installation.

mikemcquaid:
> Hmm... ok sorry for causing noise here. Firstly, the merge failed checks because for some reason we must specify a tag and a revision number. The latter doesn't seem like it should be necessary, but I can update it.

That'd be good, thanks. The reason it's needed is to ensure the tag doesn't change.

PR #39707, opened 2015-05-13
glib 2.44.1

tschoonj:
version bump
due to an upstream fix, one patch is no longer necessary


xu-cheng:

PR #39708, opened 2015-05-13
pull: warning if we are not in master

xu-cheng:
I just accidentally push wrong branch into Homebrew core as I didn't realize I was not in master branch. :cry: Thankfully, the result is creating a new branch in Homebrew instead of something worse.

So here am I, adding a new warning in `brew pull` command.


mikemcquaid:
:+1:

PR #39709, opened 2015-05-13
haskell formulae: remove gmp dependencies

geoff-codes:
With a couple of other minor changes (change sha256, remove `fails_with :clang`s, as this shouldn't be necessary any more).

### DIFF
--- a/Library/Formula/cgrep.rb
+++ b/Library/Formula/cgrep.rb
@@ -5,8 +5,10 @@ class Cgrep < Formula
 
   homepage "https://github.com/awgn/cgrep"
   url "https://github.com/awgn/cgrep/archive/v6.4.12.tar.gz"
-  sha1 "4933c1ae055d5c04f567c9405339ce4f972ef62b"
+  sha256 "a38d7957854b9b6f55ed8610d88b0ba3d5061d7194e3ec13e608d7a4515371f5"
   head "https://github.com/awgn/cgrep.git"
+
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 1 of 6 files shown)


mikemcquaid:
A few things failing to install.

geoff-codes:
Yep. cless, cgrep, and iris, to be exact. It's our good friend the cabal dependency resolver: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/26063/

They all have dependency checks for base (or lens) < 4.8, and ghc 7.10.1 ships with 4.8 (exactly). So these never would have built with 7.10.1 regardless; I'm sure if you look back at commit history, these all pre-date the change from 7.8.X to 7.10.X. so I'm not really sure how to attack this one.

geoff-codes:
pandoc-citeproc built fine, but it's test block failed to produce any output. Not sure what that means? I didn't change the test, but assumedly it passed before...

mikemcquaid:
In those cases I guess I'd just omit their `revision` bump and we won't rebottle them.

geoff-codes:
@mikemcquaid  So, I propose that we loosen cabal's dependency checks, [semi-permanently](https://github.com/pullreq/homebrew/commit/5c3d8daba2686fc791544bbc9ea92f6ed3b0086c), as this fixes two of the three failures, and it seems to be the ticket [over here](https://github.com/pthariensflame/homebrew/commit/fe0f5eea803b0fe9d31348290d41010b45ee95d7#diff-54dba270edb9e5854d59fa08a34fdb38R36) as well.

I've also fixed the pandoc-citeproc test, but backed out the changes to idris entirely as it seems very screwy and there's a [binary available](http://www.idris-lang.org/download/), so maybe it would be better handled by the cask folks.

mikemcquaid:
Works for me, thanks.

mikemcquaid:
`darcs` still failing everywhere.

geoff-codes:
Yep. Looks like it has to be done case by case. I've `rebase -p`'d that change out, added it individually for cgrep and cless. Third time's a charm...

mikemcquaid:
Thanks @geoff-codes!

PR #39710, opened 2015-05-13
juju-quickstart 2.1.1

frankban:
* Fix a bug when handling bundles including services with no num_units defined.


asparagui:
Thanks!

PR #39714, opened 2015-05-13
bbcp 14.9.2 (new formula)

sylwan:
New formula for Stanford's efficient network copy bbcp

### DIFF
--- a/Library/Formula/bbcp.rb
+++ b/Library/Formula/bbcp.rb
@@ -1,0 +1,12 @@
+class   Bbcp < Formula
+  url     "http://www.slac.stanford.edu/~abh/bbcp/bin/x86_darwin_100/bbcp"
+  homepage "http://www.slac.stanford.edu/~abh/bbcp/"
+  sha256 "cc2477a6e49cd859f14a5bb3ab5a721090273863d453592bf142c060b5e88076"
+  version "14.9.2"
+  def     install
+    bin.install 'bbcp'
+  end
+  test    do
+    system "#{bin}/bbcp", "/dev/zero localhost:/dev/null"
+  end
+end


Library/Formula/bbcp.rb
```diff
@@ -0,0 +1,12 @@
+class   Bbcp < Formula
+  url     "http://www.slac.stanford.edu/~abh/bbcp/bin/x86_darwin_100/bbcp"
+  homepage "http://www.slac.stanford.edu/~abh/bbcp/"
+  sha256 "cc2477a6e49cd859f14a5bb3ab5a721090273863d453592bf142c060b5e88076"
+  version "14.9.2"
+  def     install
+    bin.install 'bbcp'
```
mikemcquaid: This should be built from source rather than a binary.

Library/Formula/bbcp.rb
```diff
@@ -0,0 +1,12 @@
+class   Bbcp < Formula
+  url     "http://www.slac.stanford.edu/~abh/bbcp/bin/x86_darwin_100/bbcp"
```
mikemcquaid: One space after `url`

Library/Formula/bbcp.rb
```diff
@@ -0,0 +1,12 @@
...
+  homepage "http://www.slac.stanford.edu/~abh/bbcp/"
+  sha256 "cc2477a6e49cd859f14a5bb3ab5a721090273863d453592bf142c060b5e88076"
+  version "14.9.2"
```
mikemcquaid: Newline after this line.

Library/Formula/bbcp.rb
```diff
@@ -0,0 +1,12 @@
...
+  def     install
```
mikemcquaid: One space after `def`

Library/Formula/bbcp.rb
mikemcquaid: Also: use double-quotes.

Library/Formula/bbcp.rb
```diff
@@ -0,0 +1,12 @@
...
+    bin.install 'bbcp'
+  end
+  test    do
```
mikemcquaid: One space after `test` and newline above this line.

jacknagel:
This was moved to homebrew-head-only less than a montha go: f09f5c4e73f1a47d2b5fc947e4e6f977345ee46c

DomT4:
Various comments here need addressing.

sylwan:
I thought the comment by Jack Nagel:
"This was moved to homebrew-head-only less than a montha go: f09f5c4"

sorted it all, am I wrong ? should I take back the pull request ?

S

DomT4:
It looks like binaries were added just after we moved it, which may mean this can be re-added, or it can be re-added at Homebrew/binary. Ideally the binary link should be versioned upstream so it doesn't just suddenly change under us though.

If you're happy with the one in `head-only` feel free to close this, but otherwise will let Jack comment on how happy he'd be to re-add this somewhere with the new binary.

DomT4:
Tentatively closing here. We have a general desire to build things from source rather than from binaries, and this was recently moved to `head-only`. If there's only a binary available for it a more suitable place to move it to would be `homebrew-binary`.

We have a strong preference for versioned links that won't change suddenly from under us, so if you wanted to ask upstream for a versioned link prior to submission that'd be awesome. Otherwise, you can obtain the formula by doing `brew tap homebrew/head-only && brew install bbcp`.

Thanks for the submission!

PR #39719, opened 2015-05-13
ghc: fix gmp resource mirrors

vszakats:


Library/Formula/ghc.rb
```diff
@@ -15,10 +15,10 @@ class Ghc < Formula
   deprecated_option "tests" => "with-tests"
 
   resource "gmp" do
-    url "http://ftpmirror.gnu.org/gmp/gmp-6.0.0a.tar.bz2"
-    mirror "ftp://ftp.gmplib.org/pub/gmp/gmp-6.0.0a.tar.bz2"
-    mirror "ftp://ftp.gnu.org/gnu/gmp/gmp-6.0.0a.tar.bz2"
-    sha256 "7f8e9a804b9c6d07164cf754207be838ece1219425d64e28cfa3e70d5c759aaf"
+    url "http://ftpmirror.gnu.org/gmp/gmp-6.0.0a.tar.xz"
+    mirror "https://ftp.gnu.org/gnu/gmp/gmp-6.0.0a.tar.xz"
```
mikemcquaid: Why change this? Would be good to keep this all consistent with the `gmp` formula.

vszakats: There are a few issues:
- In latest formula update https://github.com/Homebrew/homebrew/commit/b160b165d42f07abc912b1903095c593de61a0df#diff-e06a9de6bb362e22e1ceb9a66eaf171e `https` was reset to `ftp` which is a step down, plus it diverges from the standard `gmp` formula
- `ftp` should probably be positioned at the end of the `mirror` list, given its shortcomings over both `http` and `https`
- `.xz` is just smaller than `.bz2` (formulae use a mixture of these at the moment, I'm not sure of the policy). Probably worth the update. Did it in a separate commit to not mix the hash change with other updates.

mikemcquaid: Let's just use `bz2` here for now but otherwise :+1:

vszakats: Thanks! `.bz2` it is now.

DomT4: As a side addition: `ftp` is a nightmare to get through tight firewalls. It's the reason I've been religiously sticking `http` or `https` mirrors onto every GNU formulae I deal with :smile_cat:.

vszakats: Agreed fully on the `ftp` thing. Plus there is not even a chance for security there.

vszakats: Turns out gmplib.org downloads are also available via `https`.

jacknagel:
See my comment in https://github.com/Homebrew/homebrew/pull/39728. I'll merge this if its updated to match the existing gmp formula, though.

vszakats:
I've made the update. Thanks.

PR #39722, opened 2015-05-13
brew.1.md: Document HOMEBREW_LOGS

sjackman:


sjackman:
`share/man/man1/brew.1` needs to be regenerated. How is that done?

mikemcquaid:
> share/man/man1/brew.1 needs to be regenerated. How is that done?

Run `brew man`

sjackman:
Thanks, Mike. Done.

Issue #39724, opened 2015-05-13
expand yara.rb formula by building and installing yara-python


frbapolkosnik:
Please see the notes about building and installing yara-python at:
http://yara.readthedocs.org/en/latest/gettingstarted.html#compiling-and-installing-yara

The reason why I'm asking is that yara-python is actually present in the source code of yara, and having the versions of yara and yara-python at the same version is pretty important.
Also, yara-python depends on the yara library built in the yara formula.

jacknagel:
Closing in favor of the pull request; note that pull requests are also issues.

PR #39726, opened 2015-05-13
Update to yara.rb formula to build yara-python

frbapolkosnik:
yara-python code is distributed in the same archive as yara, so building yara-python seems the right thing to do.

This is related to https://github.com/Homebrew/homebrew/issues/39724
Thanks!

### DIFF
--- a/Library/Formula/yara.rb
+++ b/Library/Formula/yara.rb
@@ -38,7 +38,7 @@ class Yara < Formula
     # build yara-python
     system "cd", "yara-python"
     system "python", "setup.py", "build"
-    system "python", "setup.py", "install"
+    system "sudo", "python", "setup.py", "install"
     system "cd", ".."
   end
 
--- a/Library/Formula/yara.rb
+++ b/Library/Formula/yara.rb
@@ -35,7 +35,7 @@ class Yara < Formula
                           "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
     system "make", "install"
-    # build yara-python
+
     cd "yara-python" do
       system "python", *Language::Python.setup_install_args(prefix)
     end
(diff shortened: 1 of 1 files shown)


Library/Formula/yara.rb
```diff
@@ -35,6 +35,11 @@ def install
                           "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
     system "make", "install"
+    # build yara-python
+    system "cd", "yara-python"
+    system "python", "setup.py", "build"
+    system "python", "setup.py", "install"
+    system "cd", ".."
```
tdsmith: You can replace all of this with:

```ruby
cd "yara-python" do
  system "python", *Language::Python.setup_install(prefix)
end
```

If yara-python has Python dependencies, they need to be captured as `resource` blocks and installed by the formula as well.

tdsmith: (Oops, fixed a typo in the above.)

frbapolkosnik:
Thanks again for all the help. It looks like it built fine this time.

Library/Formula/yara.rb
```diff
@@ -35,6 +35,10 @@ def install
                           "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
     system "make", "install"
+    # build yara-python
```
mikemcquaid: Can remove this comment but keep an empty line.

mikemcquaid:
@tdsmith Are bindings like this linked against the system/Homebrew Python by default typically? Do we want to guard it behind an option?

frbapolkosnik:
Removed the comment, any other blockers?

tdsmith:
@mikemcquaid Importable modules shouldn't link any Python; I've added an audit check to verify that nothing in site-packages has an explicit link. Building bindings like this against system Python is fine, and the built modules will work fine with a brewed Python, and they're safe to bottle. Using `depends_on :python if MacOS.version <= :snow_leopard` is probably polite to include for 10.6 users; I'll add it myself here when I pull this tonight.

PR #39733, opened 2015-05-13
open-mpi: Add option to build with SGE support.

yesimon:


mikemcquaid:
What does this do? What would be the disadvantage (if any) in building it by default?

yesimon:
By default OpenMPI has stopped building with support for SGE (Sun Grid Engine). This is a commonly used cluster management software and is typically a use case for OpenMPI in general. See http://gridscheduler.sourceforge.net/

mikemcquaid:
Let's just enable it by default and don't provide an option unless there's a disadvantage to having it enabled.

yesimon:
I tested it to make sure it builds even on systems without SGE installed and it looks good.

DomT4:
@BrewTestBot test this please

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/50dcedbb3743bbacef8ab37359cd365068e3abd9. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

PR #39736, opened 2015-05-13
datamash 1.0.6: Move to Homebrew/science

sjackman:
Change SHA1 to SHA256
Update test


sjackman:
Should this formula be moved to Homebrew/science?

mikemcquaid:
If you think it should and nothing depends on it I'm :+1:

sjackman:
Sounds good to me. Would you like to merge this PR, then we'll move the formula in a second PR?

sjackman:
I forget how to do a move, frankly. Do I simply open one PR here to delete the formula and a second PR at Homebrew-sci to add the formula?

mikemcquaid:
Yep (and link them in each other).

sjackman:
Done. See Homebrew/homebrew-science#2261

DomT4:
Needs to be added to `tap_migrations.rb` as well, I'd presume.

sjackman:
Thanks for the tip, Dominyk. Done.

mikemcquaid:
Thanks @sjackman!

PR #39747, opened 2015-05-14
heroku-toolbelt: check OSX version for ruby dep

sometimesfood:
This pull request removes the ruby dependency for heroku-toolbelt for OS X versions ≥ Mavericks.

The dependency is needed because heroku-toolbelt needs Ruby ≥1.9, but starting with Mavericks, OS X ships with Ruby 2.0 anyway, thereby making this dependency obsolete for newer OS X versions.

### DIFF
--- a/Library/Formula/heroku-toolbelt.rb
+++ b/Library/Formula/heroku-toolbelt.rb
@@ -4,7 +4,7 @@ class HerokuToolbelt < Formula
   sha256 "008393132b6422982221245f50cf28792f9f7a298d0ae13e34273d80976e0e77"
   head "https://github.com/heroku/heroku.git"
 
-  depends_on :ruby => "1.9"
+  depends_on :ruby => "1.9" if MacOS.version < :mavericks
 
   def install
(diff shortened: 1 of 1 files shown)


mikemcquaid:
This `if` statement isn't needed as it's already handled by the dependency. `:ruby` means it'll only install Ruby if it can't find it in the `PATH`.

sometimesfood:
Interesting. For some reason, some time in the last few days a brew-installed ruby version appeared on my system and `brew uses --installed ruby` only lists `heroku-toolbelt`, so that's why I thought the heroku-toolbelt formula was to blame.

Can't seem to reproduce the behaviour right now, though. Sorry for the noise.

mikemcquaid:
Yeh, this was fixed in 02d5bcad8b44c4c12acb4b26fe7596f2c0f7d66d. No worries about noise, always prefer a PR to an issue :wink:

sometimesfood:
Ah, great, now that makes much more sense. :+1:

PR #39758, opened 2015-05-14
ENV: Add gcc-5 to compilers support C++11

ntkme:
The PR marks gcc-5 as compiler support C++11.


mikemcquaid:
CC @jacknagel @mistydemeo for review. This makes sense to me but I just want to make sure this isn't missing somewhere else.

jacknagel:
I looked, didn't find any other places.

mistydemeo:
LGTM.

PR #39767, opened 2015-05-14
frag_find: head added

bfontaine:
I also added a better test.

### DIFF
--- a/Library/Formula/frag_find.rb
+++ b/Library/Formula/frag_find.rb
@@ -1,17 +1,36 @@
-require 'formula'
-
 class FragFind < Formula
-  homepage 'https://github.com/simsong/frag_find'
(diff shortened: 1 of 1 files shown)


DomT4:
Apparently needs an `openssl` dependency and consequent revision.

bfontaine:
@DomT4 Done. Is there any way to detect this locally, before making the PR?

mikemcquaid:
@bfontaine `brew audit frag_find` _after_ installing.

DomT4:
It should be getting thrown by [this block](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/formula_cellar_checks.rb#L142-L157) as well after you install it locally. I don't know if you have to have the developer env set or build `--debug` or such to see those messages though.

PR #39768, opened 2015-05-14
ip_relay: basic test added

bfontaine:


mikemcquaid:
Thanks again @bfontaine!

PR #39771, opened 2015-05-15
kibana 4.1.0 (new formula)

auskast:
Adding a new formula to install kibana 4.0.2. This formula installs/links the config file at /usr/local/etc/kibana/kibana.yml and the plugins directory at /usr/local/var/lib/kibana/plugins/

### DIFF
--- a/Library/Formula/kibana.rb
+++ b/Library/Formula/kibana.rb
@@ -1,0 +1,63 @@
+class Kibana < Formula
+  desc "Visualization tool for elasticsearch"
+  homepage "https://www.elastic.co/products/kibana"
+  url "https://github.com/elastic/kibana/archive/v4.1.0.tar.gz"
+  sha256 "df30e084faa1a7b0b2694fade5340ebb9125b20da481c40ee2ff8c7d4f31ac7c"
+  head "https://github.com/elastic/kibana.git"
+
+  depends_on "node"
+
+  def install
+    ENV["HOME"] = buildpath/".brew_home"
+    ENV.prepend_path "PATH", "#{Formula["node"].opt_libexec}/npm/bin"
+
+    system "npm", "install"
+    system "npm", "install", "grunt-cli"
+    system "npm", "install", "bower"
+    system "./node_modules/.bin/bower", "install"
+    system "./node_modules/.bin/grunt", "build", "--force"
+
+    dist_dir = buildpath/"build/dist/kibana"
+
+    rm_f dist_dir/"bin/*.bat"
+
+    prefix.install dist_dir/"src"
+    (etc/"kibana").mkpath
+    (var/"lib/kibana/plugins").mkpath
+
+    (etc/"kibana").install dist_dir/"config/kibana.yml" unless (etc/"kibana/kibana.yml").exist?
+
+    # point to our node
+    inreplace dist_dir/"bin/kibana" do |s|
+      s.sub! /^NODE=.*$/, "NODE=#{Formula["node"].opt_bin}/node"
+    end
+
+    bin.install dist_dir/"bin/kibana"
+
+    (prefix/"config").install_symlink etc/"kibana/kibana.yml"
+    prefix.install_symlink var/"lib/kibana/plugins"
+  end
+
+  plist_options :manual => "kibana"
+
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN"
+    "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+      <dict>
+        <key>Label</key>
+        <string>#{plist_name}</string>
+        <key>Program</key>
+        <string>#{opt_bin}/kibana</string>
+        <key>RunAtLoad</key>
+        <true/>
+      </dict>
+    </plist>
+  EOS
+  end
+
+  test do
+    assert_match /#{version}/, shell_output("#{bin}/kibana -V")
+  end
+end


Library/Formula/kibana.rb
```diff
@@ -0,0 +1,71 @@
+class Kibana < Formula
+  homepage 'https://www.elastic.co/products/kibana'
```
DomT4: Double quotes everywhere.

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,71 @@
...
+  url 'https://download.elastic.co/kibana/kibana/kibana-4.0.2-darwin-x64.tar.gz'
+  version '4.0.2'
+  sha1 'ce6136d826cf8ac1da71abea419f2f5220055933'
```
DomT4: New formulae should always be ` sha256 `.

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,71 @@
...
+    prefix.install %w[ node src ]
+
+    # setup etc path
+    etc_path.mkpath
+
+    # setup var path
+    var_path.mkpath
+
+    # copy/link config
+    cp 'config/kibana.yml', conf_path unless conf_path.exist?
```
DomT4: Pretty sure you can get away with doing `(etc/"kibana").install "kibana.yml"` and that way you can drop the `etc_path`, `conf_path` logic.

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,71 @@
+class Kibana < Formula
+  homepage 'https://www.elastic.co/products/kibana'
+  url 'https://download.elastic.co/kibana/kibana/kibana-4.0.2-darwin-x64.tar.gz'
+  version '4.0.2'
+  sha1 'ce6136d826cf8ac1da71abea419f2f5220055933'
+  head 'https://github.com/elastic/kibana.git'
+
+  def install
+    # install base files
```
DomT4: Meh to the verbose comments. I'd cut them.

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,71 @@
...
+    prefix.install %w[ node src ]
+
+    # setup etc path
+    etc_path.mkpath
+
+    # setup var path
+    var_path.mkpath
```
DomT4: Why are we making these `def` blocks rather than just inlining it? `(var/"lib/kibana").mkpath` and similar.

DomT4: I guess if it's necessary to actually symlink stuff back to the `prefix` it's useful, but even then you can safely afford to cut the `def`s in half in number. Just `prefix.install_symlink var_path/"plugins"`. Don't need a def there.

auskast: I like the def blocks so that I don't have to repeat path definitions in multiple places. Makes it easier to change later if the package changes in the future

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,63 @@
+class Kibana < Formula
+  homepage "https://www.elastic.co/products/kibana"
+  url "https://download.elastic.co/kibana/kibana/kibana-4.0.2-darwin-x64.tar.gz"
```
mistydemeo: It looks like this is a prebuilt binary? Can this use Homebrew's node?

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,63 @@
...
+    (prefix/"config").install_symlink conf_path
+
+    (var_path/"plugins").mkpath
+    prefix.install_symlink var_path/"plugins"
+
+    bin.install "bin/kibana"
+  end
+
+  def caveats; <<-EOS.undent
+    Modify #{etc_path}/kibana.yml to set the elasticsearch_url to your Elasticsearch instance
```
mikemcquaid: Let's default this to the Homebrew elasticsearch location and kill this comment.

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    Plugins are installed to #{var_path}/plugins
```
mikemcquaid: Can probably kill this comment.

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,63 @@
...
+  def etc_path
+    etc/"kibana"
+  end
+
+  def var_path
+    var/"lib/kibana"
+  end
+
+  test do
+    assert_equal "4.0.2\n", `kibana -V`
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,63 @@
...
+  EOS
+  end
+
+  def etc_path
+    etc/"kibana"
+  end
+
+  def var_path
+    var/"lib/kibana"
+  end
```
mikemcquaid: Inline these methods.

Library/Formula/kibana.rb
mikemcquaid: Yeh, this needs to be built from source.

auskast: Its a packaged node app, not a binary

mikemcquaid: What about this is OS X (`darwin`) and 64-bit (`x64`) specific?

auskast: The packaged node is the only difference

mistydemeo: That's what we're asking about: is it possible to make this use Homebrew's node instead of the packaged one?

auskast: I suppose if the run script gets modified it can

mikemcquaid: So this work will as-is on Linux? Can we use the GitHub repo instead?

DomT4:
If possible an answer to Mike and Misty's questions would be :heart:, Thanks!

auskast:
Sorry, been quite busy! I changed this to directly download from the github archive.

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    (etc/"kibana").mkpath
+    (var/"lib/kibana/plugins").mkpath
+
+    conf_path = etc/"kibana/kibana.yml"
+    cp "config/kibana.yml", conf_path unless conf_path.exist?
+
+    # point to our node
+    inreplace "bin/kibana" do |s|
+      s.sub! /^NODE=.*$/, "NODE=#{HOMEBREW_PREFIX}/bin/node"
```
DomT4: This may be better as `#{Formula["node"].opt_bin}/node` so it doesn't break if someone unlinks Node.

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,54 @@
+class Kibana < Formula
+  homepage "https://www.elastic.co/products/kibana"
+  url "https://github.com/elastic/kibana/archive/v4.0.2.tar.gz"
+  version "4.0.2"
```
DomT4: Probably redundant from the URL?

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,54 @@
...
+  depends_on "node"
+
+  def install
+    prefix.install "src"
+
+    (etc/"kibana").mkpath
+    (var/"lib/kibana/plugins").mkpath
+
+    conf_path = etc/"kibana/kibana.yml"
+    cp "config/kibana.yml", conf_path unless conf_path.exist?
```
DomT4: `(etc/"kibana").install "config/kibana.yml" unless (etc/"kibana/kibana.yml").exist?` is possibly better.

Library/Formula/kibana.rb
bfontaine: Yes:

```
$ brew irb
>> Version.parse("https://github.com/elastic/kibana/archive/v4.0.2.tar.gz").to_s
=> "4.0.2"
```

DomT4:
Few more nits but we're getting there :+1:.

auskast:
Reworked this to build from source

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,63 @@
...
+  url "https://github.com/elastic/kibana/archive/v4.0.2.tar.gz"
+  sha256 "facb5b2eb31de38ebf6e30bb656c62a512e540623df98b33655f5b0da896dd27"
+  head "https://github.com/elastic/kibana.git"
+
+  depends_on "node"
+
+  def install
+    ENV.prepend_path "PATH", HOMEBREW_PREFIX/"bin"
+    system "npm", "install"
+    system "npm", "install", "grunt-cli"
```
DomT4: > ENV.prepend_path "PATH", HOMEBREW_PREFIX/"bin"

This is unlikely to work. Superenv blocks off access to the core prefix's `bin`.

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,63 @@
...
+    system "npm", "install", "bower"
```
DomT4: > system "npm", "install", "bower"

These should be kept in the `buildpath` as they aren't runtime, so ideally double check we aren't installing them in formula's bin. Thanks!

Library/Formula/kibana.rb
auskast: Hmm it worked when I ran it... How should I be accessing the npm bin?

Library/Formula/kibana.rb
auskast: I checked and the grunt build is only keeping production dependencies in the final build, but how should I do this anyway just to be complete about it?

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    # point to our node
+    inreplace dist_dir/"bin/kibana" do |s|
+      s.sub! /^NODE=.*$/, "NODE=#{Formula["node"].opt_bin}/node"
+    end
+
+    bin.install dist_dir/"bin/kibana"
+  end
+
+  def post_install
```
DomT4: What does this block achieve?

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,63 @@
...
+        <string>#{opt_bin}/kibana</string>
+        <key>RunAtLoad</key>
+        <true/>
+      </dict>
+    </plist>
+  EOS
+  end
+
+  test do
+    assert_equal "4.0.2\n", `kibana -V`
```
DomT4: Make this non-dependant on it being bumped each time. `assert_equal /#{version}/, shell_output("#{bin}/kibana -V")` should work.

Library/Formula/kibana.rb
auskast: Symlinking the config file and plugins directory so they can be preserved through upgrades

DomT4: If they're in `var` and `etc` they won't get removed across upgrades. Only the prefix is wiped in that situation, or in a `brew rm kibana` situaiton. Do they need to be symlinked back to the prefix for the binary to correctly find them?

Library/Formula/kibana.rb
DomT4: Superenv may allows overrides when the path is explicitly re-added to the `$PATH` - @jacknagel will know either way. We want to be sure if we're going to add the override that it doesn't pick up anything unexpected.

Library/Formula/kibana.rb
DomT4: You can install things interactively with `brew install --interactive`. We essentially don't want to end up with `bower` and `grunt` being installed into the formula's eventual `prefix`, because that can cause conflict if a user already has bower/grunt installed. If only the `kibana` binary and config/plugin files are getting installed, that's :100:.

Library/Formula/kibana.rb
auskast: Yeah they either need to be symlinked, or the binary needs to be modified to look elsewhere. Symlinking seems cleaner to me.

Library/Formula/kibana.rb
auskast: Looks good to me:
```
$ ls -la $(brew --prefix kibana)/src/node_modules
total 0
drwxr-xr-x  24 pfitzgerald  wheel  816 Jun  1 16:21 .
drwxr-xr-x  12 pfitzgerald  wheel  408 Jun  1 16:21 ..
drwxr-xr-x   6 pfitzgerald  wheel  204 Jun  1 16:21 .bin
drwxr-xr-x   6 pfitzgerald  wheel  204 Jun  1 16:21 ansicolors
drwxr-xr-x   7 pfitzgerald  wheel  238 Jun  1 16:21 bluebird
drwxr-xr-x   9 pfitzgerald  wheel  306 Jun  1 16:21 body-parser
drwxr-xr-x  13 pfitzgerald  wheel  442 Jun  1 16:21 bunyan
drwxr-xr-x   8 pfitzgerald  wheel  272 Jun  1 16:21 commander
drwxr-xr-x   8 pfitzgerald  wheel  272 Jun  1 16:21 compression
drwxr-xr-x   9 pfitzgerald  wheel  306 Jun  1 16:21 cookie-parser
drwxr-xr-x  14 pfitzgerald  wheel  476 Jun  1 16:21 debug
drwxr-xr-x   8 pfitzgerald  wheel  272 Jun  1 16:21 elasticsearch
drwxr-xr-x   9 pfitzgerald  wheel  306 Jun  1 16:21 express
drwxr-xr-x   9 pfitzgerald  wheel  306 Jun  1 16:21 glob
drwxr-xr-x  12 pfitzgerald  wheel  408 Jun  1 16:21 http-auth
drwxr-xr-x  15 pfitzgerald  wheel  510 Jun  1 16:21 jade
drwxr-xr-x  13 pfitzgerald  wheel  442 Jun  1 16:21 js-yaml
drwxr-xr-x   8 pfitzgerald  wheel  272 Jun  1 16:21 less-middleware
drwxr-xr-x   7 pfitzgerald  wheel  238 Jun  1 16:21 lodash
drwxr-xr-x   8 pfitzgerald  wheel  272 Jun  1 16:21 morgan
drwxr-xr-x  17 pfitzgerald  wheel  578 Jun  1 16:21 request
drwxr-xr-x   5 pfitzgerald  wheel  170 Jun  1 16:21 requirefrom
drwxr-xr-x  16 pfitzgerald  wheel  544 Jun  1 16:21 semver
drwxr-xr-x   8 pfitzgerald  wheel  272 Jun  1 16:21 serve-favicon
```

Library/Formula/kibana.rb
auskast: `assert_match` did it. Thanks!

mikemcquaid: Shout when this is pushed, thanks!

auskast:
Pushed

mikemcquaid:
@DomT4 Can you let me know when you're happy with this? I'm a bit out my depth on node stuff.

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def install
+    ENV.prepend_path "PATH", HOMEBREW_PREFIX/"bin"
+    system "npm", "install"
+    system "npm", "install", "grunt-cli"
+    system "npm", "install", "bower"
+    system "./node_modules/.bin/bower", "install"
+    system "./node_modules/.bin/grunt", "build", "--force"
+
+    dist_dir = Pathname("build/dist/kibana")
```
DomT4: You don't need to lean on `Pathname` here - Should be just `dist_dir = buildpath/"build/dist/kibana"`.

DomT4:
We need to add a line like `ENV["HOME"] = buildpath/"home"` to `def install` ideally, otherwise when `HOMEBREW_SANDBOX` is set we hit:

```
npm ERR! Error: EPERM, mkdir '/Users/Dominyk/.npm/debug'
npm ERR!     at Error (native)
npm ERR!  { [Error: EPERM, mkdir '/Users/Dominyk/.npm/debug']
npm ERR!   errno: -1,
npm ERR!   code: 'EPERM',
npm ERR!   path: '/Users/Dominyk/.npm/debug',
npm ERR!   parent: 'kibana' }
npm ERR!
npm ERR! Please try running this command again as root/Administrator.
```

And when `HOMEBREW_SANDBOX` isn't set we spam up `~/.npm` with our build cache files, which isn't ideal either. No extra modules are installed, which is :+1:, but I think attempting to modify an existing cache in user's `HOME` directories is more trouble than it's worth when we have an easy override available.

This will also fail `-fs` unless there's an active network connection because of the way `bower` fetches dependencies. This is expected, and it's not practical to vendor in every single dependency so we're good here.

When the `HOME` issue and the nit around the `Pathname` usage are fixed I'd be :+1: on inclusion.

mikemcquaid:
> We need to add a line like ENV["HOME"] = buildpath/"home" to def install ideally

I'd be in favour of adding that globally, to be honest.

auskast:
Made the ENV and buildpath changes

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,64 @@
...
+  desc "Visualization tool for elasticsearch"
+  homepage "https://www.elastic.co/products/kibana"
+  url "https://github.com/elastic/kibana/archive/v4.0.2.tar.gz"
+  sha256 "facb5b2eb31de38ebf6e30bb656c62a512e540623df98b33655f5b0da896dd27"
+  head "https://github.com/elastic/kibana.git"
+
+  depends_on "node"
+
+  def install
+    ENV["HOME"] = buildpath/"home"
```
DomT4: Team had a discussion where `home` was deemed too generic a name for this variable - It'll be changed elsewhere soon as well. Can you set it as `buildpath/".brew_home"`. Sorry for the crossed wires!

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,64 @@
...
+    ENV.prepend_path "PATH", HOMEBREW_PREFIX/"bin"
```
DomT4: This should be `ENV.prepend_path "PATH", "#{Formula["node"].opt_libexec}/npm/bin"` instead.

Library/Formula/kibana.rb
```diff
@@ -0,0 +1,64 @@
...
+    system "npm", "install"
```
DomT4: Add a blank line between lines 12 & 13. Helps separate the `ENV` tweaks from the actual install.

DomT4:
Few nits left to tidy up but this is looking very close to ready, IMO.

Library/Formula/kibana.rb
mikemcquaid: They can be symlinked in `install`, no?

auskast: Sure. Moved them.

mikemcquaid:
@DomT4 Merge this if you're happy with it, :+1: from me.

DomT4:
There's a new version out, `https://github.com/elastic/kibana/archive/v4.1.0.tar.gz`. @auskast If you want to update it to the latest before I pull here, that'd be great. Happy with this otherwise.

auskast:
What about naming of the branch? Should I create a new PR or just update this one with the latest version?

DomT4:
The branch name doesn't matter. Either commit the new version, squash and adjust the commit message or just commit the new version as a new commit in the same branch. The former is cleaner, but either works :)

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/78054b9546f7ef495046da2f4f77705524eaf2dd. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #39772, opened 2015-05-15
tinyproxy: fix var/etc placement

### LINKED ISSUE #39748: Tinyproxy fail to run due to prefix
Using the latest `tinyproxy 1.8.3`, I get this

```
$ tinyproxy -d
tinyproxy: Could not create file /usr/local/Cellar/tinyproxy/1.8.3/var/log/tinyproxy/tinyproxy.log: No such file or directory
tinyproxy: Could not create file /usr/local/Cellar/tinyproxy/1.8.3/var/run/tinyproxy/tinyproxy.pid: No such file or directory
tinyproxy: Could not create PID file.
```

Suspect `configure` option needs to be updated in the formula.

DomT4:
Closes #39748.

### DIFF
--- a/Library/Formula/tinyproxy.rb
+++ b/Library/Formula/tinyproxy.rb
@@ -1,9 +1,7 @@
-require 'formula'
-
 class Tinyproxy < Formula
   homepage "https://www.banu.com/tinyproxy/"
   url "https://www.banu.com/pub/tinyproxy/1.8/tinyproxy-1.8.3.tar.bz2"
-  sha1 "2538fbd190d3dc357a2e7c2a07ea0fbefb768a13"
+  sha256 "be559b54eb4772a703ad35239d1cb59d32f7cf8a739966742622d57df88b896e"
 
   bottle do
     sha1 "edf8eed5ce1ae02bdd09cefba5d3700b6d34c9c9" => :mavericks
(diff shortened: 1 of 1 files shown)


Library/Formula/tinyproxy.rb
```diff
@@ -11,36 +9,56 @@ class Tinyproxy < Formula
...
   # Fix linking error, via MacPorts: https://trac.macports.org/ticket/27762
   patch :p0 do
     url "https://trac.macports.org/export/83413/trunk/dports/net/tinyproxy/files/patch-configure.diff"
-    sha1 "e946269b681f3ffaa1acd120c93050cb63bfe743"
+    sha256 "414b8ae7d0944fb8d90bef708864c4634ce1576c5f89dd79539bce1f630c9c8d"
   end
 
   def install
+    (var/"log/tinyproxy").mkpath
+    (var/"run/tinyproxy").mkpath
```
mikemcquaid: Do these need to be in `postinstall` to be kept or does specifying the `localstatedir` below mean there's files in them?

DomT4: Nope. Nothing installed, they just need to exist before the binary is executed. Have pushed them into `post_install`.

Library/Formula/tinyproxy.rb
```diff
@@ -11,36 +9,58 @@ class Tinyproxy < Formula
...
+    begin
+      pid = fork do
+        exec "#{sbin}/tinyproxy"
+      end
+      sleep 2
+    ensure
+      Process.kill("SIGINT", pid)
+      Process.wait(pid)
+    end
   end
```
xu-cheng: I think this test will never fail. You may need to test the proxy with `curl`?

DomT4: A test that never fails? That sounds like something Homebrew *would* want :wink:. Have fixed it. I'm always just hesitant to trust the test-bot to get internal network connections right, It's had major issues before.

PR #39775, opened 2015-05-15
tnef: update homepage and url, use sha256

lachlancooper:


mikemcquaid:
Install is failing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/26174/version=yosemite/testReport/junit/brew-test-bot/yosemite/install_tnef/ 

Did this work for you locally? Probably needs `autoconf` run first or similar.

lachlancooper:
Thanks for the tip! It looks like the original upstream package included `configure`, but it's now missing so `autoconf` was required.

mikemcquaid:

PR #39776, opened 2015-05-15
sshguard 1.6.0

dunn:
I didn't see an announcement on the website but it looks like a proper stable release: https://sourceforge.net/p/sshguard/mailman/message/34078097/

### DIFF
--- a/Library/Formula/sshguard.rb
+++ b/Library/Formula/sshguard.rb
@@ -1,15 +1,17 @@
-require 'formula'
-
 class Sshguard < Formula
-  homepage 'http://www.sshguard.net/'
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Thanks @dunn!

Library/Formula/sshguard.rb
```diff
@@ -19,20 +21,12 @@ class Sshguard < Formula
     sha1 "68cd0910d310e4d23e7752dee1b077ccfe715c0b"
   end
```
mikemcquaid: Pulling this as-is for now but if this isn't getting merged then I think we should think about killing it.

PR #39778, opened 2015-05-15
gtkextra 3.1.3

tschoonj:
version bump
addresses @xu-cheng 's comment in
067f5fc33183be3ba22571a14f58e0ce2be33bf4

### DIFF
--- a/Library/Formula/gtkextra.rb
+++ b/Library/Formula/gtkextra.rb
@@ -1,7 +1,7 @@
 class Gtkextra < Formula
   homepage "http://gtkextra.sourceforge.net/"
-  url "https://downloads.sourceforge.net/project/gtkextra/3.1/gtkextra-3.1.2.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/gtkextra.rb
```diff
@@ -11,7 +11,7 @@ class Gtkextra < Formula
   end
 
   depends_on "gtk+"
-  depends_on "pkg-config" => :run
+  depends_on "pkg-config" => :build
```
xu-cheng: It seems `pkg-config` was only used in test. So we can remove this line completely?

(Also, please try not to ping people in the commit message. I got too many notifications from that. Thanks.)

tschoonj:
well it's also used during the build stage. configure needs it to find gtk+ and its dependencies

xu-cheng:
Thanks @tschoonj

PR #39781, opened 2015-05-15
Fix funcoeszz when using apple's sed

jaysonsantos:
This software requires gnu-sed, so I've changed it to use explicitly gsed instead of apple's sed. With this you don't have to force users changing their paths to make gnu-sed default.

### DIFF
--- a/Library/Formula/funcoeszz.rb
+++ b/Library/Formula/funcoeszz.rb
@@ -5,8 +5,18 @@ class Funcoeszz < Formula
   url 'http://funcoeszz.net/download/funcoeszz-13.2.sh'
   sha1 '33d6950dc83fd2118bc45a752c4a77be3b112573'
 
+  depends_on "gnu-sed"
+
   def install
-    prefix.install "funcoeszz-#{version}.sh" => "funcoeszz.sh"
+    # This program needs gnu sed to work, insted of forcing users
+    # to change their paths, we just change the sed callings to gsed
+    original_path = "funcoeszz-#{version}.sh"
+
+    inreplace original_path do |s|
+      s.gsub! /\bsed\b/, "gsed"
+    end
+
+    prefix.install original_path => "funcoeszz.sh"
   end
 
   def caveats; <<-EOS.undent


mikemcquaid:
Can you elaborate on how it requires it to work? Also, can you file an issue upstream asking if they can support BSD sed? Thanks!

jaysonsantos:
It fails when executing something like this `set | sed -n '/^zz[a-z0-9]/ s/ *().*//p'` but with gsed the same statement works `set | gsed -n '/^zz[a-z0-9]/ s/ *().*//p'`.

mikemcquaid:
@jaysonsantos Can you give an example of how to run that? I'm not sure I totally understand from the comment. Thanks!

jaysonsantos:
Install de current version on homebrew's master and run this:
```bash
export ZZPATH="/usr/local/opt/funcoeszz/funcoeszz.sh"
source "$ZZPATH"
zzzz
```

You will see the sed error.

DomT4:
I didn't get as far as `zzzz` using the system `sh`.

Current:
```
sh-3.2$ export ZZPATH="/usr/local/opt/funcoeszz/funcoeszz.sh"
sh-3.2$ source "$ZZPATH"
sh: /usr/local/opt/funcoeszz/funcoeszz.sh: line 1862: syntax error near unexpected token `('
sh: /usr/local/opt/funcoeszz/funcoeszz.sh: line 1862: `				paste -d"|" <(echo "$vartemp"|cut -f1 -d:|sed 's/^[[:space:]]\+//g;s/[[:space:]]\+$//g') <(echo "$vartemp"|cut -f2- -d:|sed 's/^[[:space:]]\+//g')|'
```

New:
```
sh-3.2$ source "$ZZPATH"
sh: /usr/local/opt/funcoeszz/funcoeszz.sh: line 1862: syntax error near unexpected token `('
sh: /usr/local/opt/funcoeszz/funcoeszz.sh: line 1862: `				paste -d"|" <(echo "$vartemp"|cut -f1 -d:|gsed 's/^[[:space:]]\+//g;s/[[:space:]]\+$//g') <(echo "$vartemp"|cut -f2- -d:|gsed 's/^[[:space:]]\+//g')|'
```

If I use a modern Bash, via Homebrew in this case, I get no errors at all with either BSD `sed` or GNU `sed`.

mikemcquaid:
Looks like this needs a Requirement on a newer `bash` instead.

DomT4:
CC @jaysonsantos. Can you switch to a recent `bash` and see if the `sed` error fails to reproduce for you as well? If it does, tweaking this PR to require Homebrew's `bash` would be great, and perhaps a small caveat warning of problems against the system `sh`. Thanks!

DomT4:
Ping.

PR #39789, opened 2015-05-15
add clang-omp support (new approach)

xu-cheng:
Per https://github.com/Homebrew/homebrew/pull/39676#issuecomment-102423121

This is a less invasive but more hacky approach. Request comments on which is better this or #39676

cc @Homebrew/owners @Homebrew/science


mikemcquaid:
I'm OK with this approach if other @Homebrew/owners are.

Library/Homebrew/compilers.rb
```diff
@@ -111,6 +110,14 @@ def find_compiler
           version = versions.non_apple_gcc_version(name)
           yield Compiler.new(name, version) if version
         end
+      when :clang
+        # silently substitute clang to clang-omp when openmp is required.
+        version = if @formula.needs_compiler_feature?(:openmp)
+          versions.clang_omp_build_version
```
xu-cheng: There is a flaw here. The version of clang-omp is the version of llvm it based on. Therefore it cannot be compared with the system clang build version.

xu-cheng: Any suggestion to workaround this problem? Or we are forced to go back the old approach?

mikemcquaid: Hmm, not sure.

sjackman: Not really. Can you suggest a workaround, or is it hopeless?

xu-cheng: > Can you suggest a workaround

I have one in mind is to create a pseduo build version based on system clang build version and its llvm version. However it has clearly potential harm. 

> is it hopeless?

This only means this approach is dead. We can still support clang-omp by treating it as a separate compiler.

sjackman: Does making `clang-omp` its own new compiler require modifying the formulae that `needs :openmp` to use `clang-omp`?

xu-cheng: No, however it requires to update every formulae mentioned `:clang`. Like `compiler == :clang` or `fails_with :clang`.

sjackman: 154 formulae (including Homebrew and Homebrew/science) mention `:clang`. Editing each of those seems untenable.

Library/Homebrew/os/mac.rb
```diff
@@ -120,6 +120,24 @@ def clang_build_version
...
+    # Create a pseudo build version number based on Apple Clang build version and
+    # its LLVM version. This is hack to make clang-omp's version comparable with
+    # Apple Clang build version. Please update this map with a proper build number
+    # when clang-omp gets update.
+    def clang_omp_build_version
+      @clang_omp_build_version ||=
+        case clang_omp_version
+        when "3.5" then 600
+        end
+    end
```
xu-cheng: I create a pseudo clang-omp build version here. Ask for comments.

sjackman: I have no strong complains to this approach. I also haven't yet tested the patch.

xu-cheng:
We may don't need this PR in the near future. LLVM has announced the official OpenMP support ([source](http://blog.llvm.org/2015/05/openmp-support_22.html)) and it's likely going to land in Xcode Clang.

sjackman:
Yeah! That's fantastic news.

bfontaine:
Anything new here?

xu-cheng:
I'm waiting to see whether Clang on El Capitan will be supporting OpenMP.

tdsmith:
I don't see omp.h in Xcode 7.

xu-cheng:
> I don't see omp.h in Xcode 7.

I think you still need to install `libiomp` separately

tdsmith:
Well, I'll be:
```
tim@rocketman:~$ clang -fopenmp=libiomp5 -Xclang -fopenmp=libiomp5 -o hello_clang omp_hello.c -I$(brew --prefix libiomp)/include/libiomp
tim@rocketman:~$ ./hello_clang 
Hello World from thread = 0
Hello World from thread = 3
Hello World from thread = 6
Hello World from thread = 1
Hello World from thread = 4
Hello World from thread = 5
Hello World from thread = 2
Hello World from thread = 7
Number of threads = 8
```

I learned that incantation from http://lists.llvm.org/pipermail/openmp-dev/2015-May/000532.html; I don't understand why it works yet. Simpler variations do not seem to work.

sjackman:
Wow, that's pretty darn cool.

sjackman:
What's the status of OpenMP in `clang`?

tdsmith:
unchanged, aiui

PR #39790, opened 2015-05-15
Task 2.4.4

languitar:
Updates task (taskwarrior) to 2.4.4 and adds caveats explaining how to use fish shell completions.

### DIFF
--- a/Library/Formula/task.rb
+++ b/Library/Formula/task.rb
@@ -1,8 +1,8 @@
 class Task < Formula
   homepage "http://www.taskwarrior.org/"
-  url "http://taskwarrior.org/download/task-2.4.3.tar.gz"
-  sha1 "2a0c1519b1b572c91f8d6ee489b1250a993a3e86"
-  head "https://git.tasktools.org/scm/tm/task.git", :branch => "2.4.4", :shallow => false
+  url "http://taskwarrior.org/download/task-2.4.4.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/task.rb
```diff
@@ -23,4 +23,12 @@ def install
...
     system "#{bin}/task", "--version"
   end
+
+  def caveats; <<-EOS.undent
+    In order to use the completions for the fish shell, extend $fish_complete_path with:
+      #{HOMEBREW_PREFIX}/share/doc/task/scripts/fish/
+    e.g. by putting this line into ~/.config/fish/config.fish:
+      set -x fish_complete_path #{HOMEBREW_PREFIX}/share/doc/task/scripts/fish/ $fish_complete_path
+    EOS
+  end
```
mikemcquaid: I don't think we need these caveats, thanks.

languitar: I had to read several documentation pages to find out where the completion for task resides and how to add it to fish, since there is no automatic installation possibility like for bash or zsh. So this might someone. Where should such information be out otherwise?

mikemcquaid: I just think that most people using Task aren't using Fish. I'd suggest that it be added more prominent in the Task documentation.

languitar: Ok, so I just drop that commit for now.

PR #38674, opened 2015-04-15
rust with cargo

lloeki:
This follows discussion on #33322, where it has been hypothesized that
the best course of action may very well be to include cargo in the rust
formula. Reasons include:

- upstream packages releases with cargo
- cargo has no formal versioning rules, merely binding a particular
  cargo nightly to a given rust release
- cargo needs a non feature-gated rust to build
- cargo needs to fetch cargo to bootstrap itself (at the make stage)

The most practical way to get the cargo version bound to a rust release
is to download a binary rust release and run cargo -V, which shows the
SHA1 it was built from. This is what has been done for the beta.

Alternatively, building with --HEAD relaxes all version constrains and
builds both from their respective master HEADs.

Curl is a dependency of .travis.install.deps.sh while both curl and
python are a dependency of make (via src/etc/dl-snapshot.py) to download
a cargo snapshot. Each one makes respective use of a src/rustversion.txt
and a src/snapshots.txt file, combined with some very lighteight
processing, to find the required downloads.

Thus it may be possible to include fetched downloads as resources to
improve cacheability and reduce the dependency profile, but this
requires patching of .travis.install.deps.sh, that currently
unconditionally downloads its stuff. Simply putting the cargo snapshot
in the right place ($(TARGET_ROOT)/snapshot/cargo) will certainly make
the makefile happy, as long as the proper filename unmangling is done
(see src/etc/dl-snapshot.py). Still, configure wants to probe for
python. Staying KISS, no patching has been done here: make it work, make
it right, make it fast.

PR #39792, opened 2015-05-15
rust 1.0.0

sferik:
http://blog.rust-lang.org/2015/05/15/Rust-1.0.html

In addition to updating the `rustc` source tarball, it also updates `rustc-nightly` and `cargo-nightly`.

Does anything need to be done to update the bottles or does the @BrewTestBot do that automatically?

### DIFF
--- a/Library/Formula/rust.rb
+++ b/Library/Formula/rust.rb
@@ -1,25 +1,24 @@
 class Rust < Formula
   homepage 'http://www.rust-lang.org/'
-  version "1.0.0-beta.4"
 
   stable do
-    url 'https://static.rust-lang.org/dist/rustc-1.0.0-beta.4-src.tar.gz'
-    sha256 '54e5868dd55a5c171327c72d662b5931a962b0cf160022d11c189ea232e0bd91'
+    url 'https://static.rust-lang.org/dist/rustc-1.0.0-src.tar.gz'
+    sha256 'c304cbd4f7b25d116b73c249f66bdb5c9da8645855ce195a41bda5077b995eba'
 
     resource "cargo" do
-      url "https://github.com/rust-lang/cargo.git", :revision => "83a6d0ed8208d31a1f6dab5e5183ad9eb2d65eaf", :tag => "0.2.0"
+      url "https://github.com/rust-lang/cargo.git", :revision => "a48358155c90467ed9c897930dd0da4614605dac"
     end
 
     # name includes date to satisfy cache
-    resource "cargo-nightly-2015-04-02" do
-      url "https://static-rust-lang-org.s3.amazonaws.com/cargo-dist/2015-04-02/cargo-nightly-x86_64-apple-darwin.tar.gz"
-      sha256 "18647dccb34acb6085a04b0ea1bfb9d150dc9c17f7829932ddf7e62d518df2fe"
+    resource "cargo-nightly-2015-05-14" do
+      url "https://static-rust-lang-org.s3.amazonaws.com/cargo-dist/2015-05-14/cargo-nightly-x86_64-apple-darwin.tar.gz"
+      sha256 "46403af9abb0dd0d9e3c31c20aa2508878ae1e8d0f1e7913addbfc08605b9733"
     end
 
     # name includes date to satisfy cache
-    resource "rustc-nightly-2015-04-04" do
-      url "https://static-rust-lang-org.s3.amazonaws.com/dist/2015-04-04/rustc-nightly-x86_64-apple-darwin.tar.gz"
-      sha256 "87068b325802fee65a388d249b98a184aff7670140133de78560ef375bae70a5"
+    resource "rustc-nightly-2015-05-14" do
+      url "https://static-rust-lang-org.s3.amazonaws.com/dist/2015-05-14/rustc-nightly-x86_64-apple-darwin.tar.gz"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> Does anything need to be done to update the bottles or does the @BrewTestBot do that automatically?

It'll handle them.

sferik:
> It'll handle them.

@mikemcquaid Thanks!

sferik:
Note: This pull request also reverts the `--release-channel=beta` flag, which was added in 9721953dd5ff3da7ed08809d22831323c9d39efd.

I believe this brings the Homebrew formula back in line with the official builds, but I may be wrong about that.

jonathanmarvens:
:+1::+1::+1:

Library/Formula/rust.rb
```diff
@@ -44,7 +44,6 @@ def install
     args = ["--prefix=#{prefix}"]
     args << "--disable-rpath" if build.head?
     args << "--enable-clang" if ENV.compiler == :clang
-    args << "--release-channel=beta" unless build.head?
```
steveklabnik: you may or may not want to re-add a `--release-channel=nightly" if build.head?` too

kballard: Should this not say `--release-channel=stable`? It looks like the default channel is "dev" even in the 1.0.0 tarball.

kballard: @steveklabnik Is `nightly` correct for downloading from `HEAD`? I would think `dev` is fine there.

callahad: From reading the current [configure](https://github.com/rust-lang/rust/blob/e5394240a295650b567aa406b4a0e1e3a6749a5f/configure#L645-L656) and (make.mk)[https://github.com/rust-lang/rust/blob/e5394240a295650b567aa406b4a0e1e3a6749a5f/mk/main.mk#L27-L54] scripts, it looks like `nightly` is correct for HEAD, and we need `stable` when building 1.0

Library/Formula/rust.rb
```diff
@@ -1,25 +1,25 @@
 class Rust < Formula
   homepage 'http://www.rust-lang.org/'
-  version "1.0.0-beta.4"
+  version "1.0.0"
```
bfontaine: You can leave out this line because Homebrew infers the version from the URL.

mikemcquaid:
Install is failing on all platforms, unfortunately: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/26224/version=yosemite/testReport/junit/brew-test-bot/yosemite/install_rust/

sferik:
@mikemcquaid I just pushed an update. `cargo` was locked to the wrong revision. Now `cargo` is locked to the same revision as the Rust 1.0.0 installer.

sferik:
@mikemcquaid Ping. Tests are now passing. Should be good to merge!

antstorm:
:+1:

byronyi:
When I run rustc, I got the following error information:
```
$ rustc
dyld: Library not loaded: x86_64-apple-darwin/stage1/lib/rustlib/x86_64-apple-darwin/lib/librustc_driver-4e7c5e5c.dylib
  Referenced from: /usr/local/bin/rustc
  Reason: image not found
Trace/BPT trap: 5
```

kballard:
By any chance are you running it from within the root dir of a checkout of the [rust-lang/rust](https://github.com/rust-lang/rust) repo? My recollection is that due to various reasons, attempting to run `rustc` from that folder does not work properly as it attempts to load dylibs from the stage paths instead of from `/usr/local/lib`.

FWIW, configuring rust with `--enable-rpath` (which makes it use `@rpath` when linking rustc itself) seems to fix this issue. I'm not sure offhand why rpaths are disabled by default (and when using `rustc` to build stuff yourself, it won't use rpaths for the built product unless you pass `-C rpath` as a compiler flag to `rustc`).

byronyi:
```
$ rustc --version
dyld: Library not loaded: x86_64-apple-darwin/stage1/lib/rustlib/x86_64-apple-darwin/lib/librustc_driver-4e7c5e5c.dylib
  Referenced from: /usr/local/bin/rustc
  Reason: image not found
Trace/BPT trap: 5
```

l-hedgehog:
According to the original plan (#38674), I think the 2015-04-30 nightly of cargo should be used with rust 1.0 rust-lang/rust-packaging@37686ecb5465, instead of the current 2015-05-14 one?

sferik:
@l-hedgehog My original pull request had `cargo` at revision `efb482d9ed820e26e79329d73bdd0a26c4cebf41`, which is the same version that ships with the Rust Mac installer. Unfortunately, this version of `cargo` caused the build to fail with Rust 1.0.0, so I updated to the latest revision at the time of the 1.0.0 release. You can see the console output of that failed build [here](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/26242/version=yosemite/console). In the process of compiling `term`, a dependency of `tar`, it fails with the following error:

```
/Users/brew/Jenkins/prs/version/yosemite/home/.cargo/registry/src/github.com-1ecc6299db9ec823/tar-0.2.10/src/lib.rs:186:27: 186:31 error: mismatched types:
 expected `u16`,
    found `i32`
```

If you can find a way to use this version of `cargo` without causing the build to fail, please open a pull request.

l-hedgehog:
@sferik thanks for the detailed explanation

PR #39794, opened 2015-05-15
brew.rb: complain if out-of-date

dunn:
I can't remember if I've seen discussion of doing this before; if someone else has an implementation in progress, go ahead and close this.

There's no performance impact.

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -117,8 +117,8 @@ module Homebrew
     HOMEBREW_REPOSITORY.cd { `git rev-parse --verify -q HEAD 2>/dev/null`.chuzzle }
   end
 
-  def self.git_last_commit
-    HOMEBREW_REPOSITORY.cd { `git show -s --format="%cr" HEAD 2>/dev/null`.chuzzle }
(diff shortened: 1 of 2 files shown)


jacknagel:
>There's no performance impact.

Shelling out (to git, in this case) will always have a performance impact.

A single additional call may not be noticeable for *you*, but it will be for others, and if we are not vigilant they will add up over time. Running the `brew` tool on a no-op command used to be super slow because of this, and it's only because I spent several months figuring out how to avoid shelling out that it is as fast as it is today.

I don't want to start adding this kind of stuff back.

mikemcquaid:
Perhaps a fast way of doing this would be to check the `mtime` of the `master` pointer file.

xu-cheng:
This change has potential to break the existing tools/script which invokes brew command. At least, it will print this message in a strange and confused place for these tools.

Another thing is `.git` directory may not always exist as we allow users to just download and extract the tarball. A even worse case is that git may not exist. For example, when I bootstrap linuxbrew on CentOS 5, I would install linuxbrew using curl first then use brew to install the git.

dunn:
Gotcha, all good things to know, thanks!

PR #39801, opened 2015-05-15
strongswan: Include x509 plugin

nevir:
This plugin is normally included, and required for generating x.509 certificates, but was being removed by `--disable-default`.

Also alpha-sort a few options that weren't.


mikemcquaid:

For future reference the preferred commit message format for simple version updates is `strongswan: build x509 plugin.`.

PR #39784, opened 2015-05-15
wine 1.7.43 (devel)

Dreysman:

PR #39803, opened 2015-05-15
Wine 1.7.43 (Devel)

uyjulian:


uyjulian:
#39784 beat me to it.

PR #39805, opened 2015-05-16
tinyproxy: revert 6e87b7f2d2fdbab1b69670ab2524066313b45138

DomT4:
This partially reverts https://github.com/Homebrew/homebrew/commit/6e87b7f2d2fdbab1b69670ab2524066313b45138, which conflicts with the prior merge of https://github.com/Homebrew/homebrew/commit/b71ee671d395bb54fbcf864f85c2b95e18d896ea.

Merging the second PR added a second `post_install` stage underneath the prior existing one, and the new `post_install` stage created directories which aren't compatible with correct placement of the `var` directories.

### DIFF
--- a/Library/Formula/tinyproxy.rb
+++ b/Library/Formula/tinyproxy.rb
@@ -53,28 +53,6 @@ class Tinyproxy < Formula
     (var/"run/tinyproxy").mkpath
   end
 
-  test do
-    pid = fork do
-      exec "#{sbin}/tinyproxy"
-    end
-    sleep 2
-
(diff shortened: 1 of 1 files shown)


jacknagel:
Done manually; I don't see a reason to move the test around.

DomT4:
Thanks. I tend to keep the tests as the last part of the formulae since it's the last element of the formula that we process during builds & since had to open the PR anyway though moving was harmless enough, but it's not a huge thing in the slightest.

PR #39806, opened 2015-05-16
open-mesh version up (3.1 -> 3.3)

youngtaekoh:
3.1 -> 3.3


Library/Formula/open-mesh.rb
```diff
@@ -2,8 +2,8 @@
 
 class OpenMesh < Formula
   homepage "http://openmesh.org"
-  url "http://www.openmesh.org/media/Releases/3.1/OpenMesh-3.1.tar.gz"
-  sha1 "f73fc5a072b11028b882ce1c221602040fb23be2"
+  url "www.openmesh.org/media/Releases/3.3/OpenMesh-3.3.tar.gz"
```
xu-cheng: `http://`

youngtaekoh: @xu-cheng 'http://' is now added in the following commit.

PR #39814, opened 2015-05-16
Fixed audit complaints for Sphinx about system OpenSSL if built with --mysql flag

Envek:
When Sphinx installed with `--mysql` flag `brew audit sphinx` message was:

    sphinx:
      /usr/local/Cellar/sphinx/2.2.9/bin/indexer
      /usr/local/Cellar/sphinx/2.2.9/bin/indextool
      /usr/local/Cellar/sphinx/2.2.9/bin/searchd
      /usr/local/Cellar/sphinx/2.2.9/bin/spelldump
      /usr/local/Cellar/sphinx/2.2.9/bin/wordbreaker

Brew bot doesn't detect this as it does not build (and therefore audit) Sphinx with `--mysql` or `--pgsql` flags. But one (or both) of these flags are used in most installations (IMO). So it's essential to make CI to check installation for them. But I don't know how to do that.

### DIFF
--- a/Library/Formula/sphinx.rb
+++ b/Library/Formula/sphinx.rb
@@ -2,6 +2,7 @@ require 'formula'
 
 class Sphinx < Formula
   homepage 'http://www.sphinxsearch.com'
+  revision 1
 
   stable do
     url 'http://sphinxsearch.com/files/sphinx-2.2.9-release.tar.gz'
@@ -21,13 +22,18 @@ class Sphinx < Formula
     sha256 "a71c292f774b8e449cf8775d21e7a6f41a85497e178ea6ba678b80a2f33b4ebd" => :mountain_lion
   end
 
-  option 'mysql', 'Force compiling against MySQL'
-  option 'pgsql', 'Force compiling against PostgreSQL'
-  option 'id64',  'Force compiling with 64-bit ID support'
+  option 'with-mysql',      'Force compiling against MySQL'
+  option 'with-postgresql', 'Force compiling against PostgreSQL'
+  option 'with-id64',       'Force compiling with 64-bit ID support'
+
+  deprecated_option 'mysql' => 'with-mysql'
+  deprecated_option 'pgsql' => 'with-postgresql'
+  deprecated_option 'id64'  => 'with-id64'
 
   depends_on "re2" => :optional
-  depends_on :mysql if build.include? 'mysql'
-  depends_on :postgresql if build.include? 'pgsql'
+  depends_on :mysql => :optional
+  depends_on :postgresql => :optional
+  depends_on 'openssl' if build.with?('mysql')
 
   resource 'stemmer' do
     url "https://github.com/snowballstem/snowball.git",
@@ -55,15 +61,19 @@ class Sphinx < Formula
               --localstatedir=#{var}
               --with-libstemmer]
 
-    args << "--enable-id64" if build.include? 'id64'
+    args << "--enable-id64" if build.with? 'id64'
     args << "--with-re2" if build.with? 're2'
 
-    %w{mysql pgsql}.each do |db|
-      if build.include? db
-        args << "--with-#{db}"
-      else
-        args << "--without-#{db}"
-      end
+    if build.with? 'mysql'
+      args << '--with-mysql'
+    else
+      args << '--without-mysql'
+    end
+
+    if build.with? 'postgresql'
+      args << '--with-pgsql'
+    else
+      args << '--without-pgsql'
     end
 
     system "./configure", *args


DomT4:
Probably best to do something like this:

```diff
diff --git a/Library/Formula/sphinx.rb b/Library/Formula/sphinx.rb
index 3fdde48..f2f1b25 100644
--- a/Library/Formula/sphinx.rb
+++ b/Library/Formula/sphinx.rb
@@ -13,13 +13,18 @@ class Sphinx < Formula
     sha256 "a71c292f774b8e449cf8775d21e7a6f41a85497e178ea6ba678b80a2f33b4ebd" => :mountain_lion
   end

-  option 'mysql', 'Force compiling against MySQL'
-  option 'pgsql', 'Force compiling against PostgreSQL'
-  option 'id64',  'Force compiling with 64-bit ID support'
+  option 'with-mysql', 'Force compiling against MySQL'
+  option 'with-pgsql', 'Force compiling against PostgreSQL'
+  option 'with-id64',  'Force compiling with 64-bit ID support'
+
+  deprecated_option "mysql" => "with-mysql"
+  deprecated_option "pgsql" => "with-pgsql"
+  deprecated_option "id64" => "with-id64"

   depends_on "re2" => :optional
-  depends_on :mysql if build.include? 'mysql'
-  depends_on :postgresql if build.include? 'pgsql'
+  depends_on :mysql => :optional
+  depends_on :postgresql => :optional
+  depends_on "openssl" if build.with?("mysql") || build.with?("pgsql")
```

Envek:
@DomT4 can't understand what to do in case of PostgreSQL because option is `--with-pgsql` and dependency name is `postgresql`

Should I do something like this? [Documentation](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#specifying-other-formulae-as-dependencies) is not so clear there…

    depends_on :postgresql => 'with-pgsql'

Is it really necessary to put `depends_on "openssl"` inside of `if build.with?("mysql")`?

DomT4:
Well spotted. Have updated my proposed diff above. We barely use `pgsql` as the option name, IMO we're fine to change it here, especially since the deprecation will flag it at install time.

> depends_on :postgresql => 'with-pgsql'

This wouldn't be correct, no. That kind of thing only works when the `=> "with-blah"` is an option of the referenced formula, and obviously `pgsql` isn't an option of the `postgresql` formula.

> Is it really necessary to put depends_on "openssl" inside of

No. But there will be someone out there who builds `sphinx` without either `mysql` or `postgresql` and doesn't have `openssl` installed and we shouldn't impose an unnecessary dependency on them

Envek:
We do not use `pgsql` barely as option name. [Here](https://github.com/Envek/homebrew/blob/5ecd996dc7d09029e0208abfb5a1b3b6c5c86667/Library/Formula/sphinx.rb#L53-L59) it goes to `./configure`, so I had prefer to keep option to be named `--with-pgsql` as familiar to those who compiled Sphinx from source or installed it via Homebrew earlier.

DomT4:
> We do not use pgsql barely as option name

Yes, in Homebrew more widely, we barely use `pgsql` as an option name. This format of option name:

```ruby
option "dave"

depends_on "horse" if build.include? "dave"
```
is deprecated. All options should be ` with-xyz `.

If you want to keep it as `--with-pgsql` fine, change the dep accordingly. Changing the option wording doesn't matter because the deprecation will throw a build-time warning the option has changed. It won't sneak up on people.

Either way, the discussion here is *"Let's not shove a mandatory OpenSSL dep into the formula unless the actual vanilla formula itself requires a mandatory OpenSSL dep"*.

mikemcquaid:
I'd rather we stuck to @DomT4's format. I'd accept a patch with that diff.

Envek:
Done.

Library/Formula/sphinx.rb
```diff
@@ -21,13 +22,18 @@ class Sphinx < Formula
...
+  deprecated_option 'mysql' => 'with-mysql'
+  deprecated_option 'pgsql' => 'with-postgresql'
+  deprecated_option 'id64'  => 'with-id64'
 
   depends_on "re2" => :optional
-  depends_on :mysql if build.include? 'mysql'
-  depends_on :postgresql if build.include? 'pgsql'
+  depends_on :mysql => :optional
+  depends_on :postgresql => :optional
+  depends_on 'openssl' if build.with?('mysql')
```
DomT4: `depends_on "openssl" if build.with?("mysql") || build.with("postgresql")`

Envek: It doesn't linking with OpenSSL when compiled with PostgreSQL support, so no need in OpenSSL there.

DomT4: Ah cool. From your original post I presumed it did, Sorry.

Envek: Sorry if I was unclear. I just wanted to say that CI should somehow build and audit this formula with frequently used options like `--with-mysql` and/or `--with-postgresql`.

mikemcquaid: That's not possible due to resource constraints, unfortunately.

Envek:
Is there any other notes to fix?

mikemcquaid:

For future reference the preferred commit message format is `sphinx: fix mysql/postgresql options openssl link.`.

PR #39815, opened 2015-05-16
hugo 0.14

dunn:
The new dependencies are only needed for HEAD currently but they'll also be needed in the next stable release.

### DIFF
--- a/Library/Formula/hugo.rb
+++ b/Library/Formula/hugo.rb
@@ -1,40 +1,12 @@
 require "language/go"
 
 class Hugo < Formula
+  desc "A static site generator emphasizing speed and configurability"
   homepage "https://gohugo.io/"
-  revision 1
+  url "https://github.com/spf13/hugo/archive/v0.14.tar.gz"
+  sha256 "67e0fda342e6a35490ce03f5535e30cf04bc06775c47bacf416d96d83cce4535"
 
-  stable do
-    url "https://github.com/spf13/hugo/archive/v0.13.tar.gz"
-    sha256 "4d3f1fd0df1f993ef188b2e2c2e1c4a45411d9f7b4ff6ebcad24b9288e9ff278"
-    # remove `-DEV` from version string to prevent confusion
-    patch do
-      url "https://gist.githubusercontent.com/dunn/b4a5c15815b9067d0dcf/raw/d64809d68b716e0ce3286dd685dd23eb475c18ec/hugo.diff"
-      sha256 "1b8caa1dccd2001152de08b906d7429385c570243dbf7821a0e07f1643e2c388"
-    end
-
-    # corrects repository url for osext; remove in next release
-    patch do
-      url "https://github.com/spf13/hugo/commit/967d001ebe40cfe90992d953880356a495216202.diff"
-      sha256 "5a991372a54ac049783926c1993d920f46ab52563ff1877e7aaddc40b5a79e22"
-    end
-
-    # Install version of fsnotify needed by stable
-    go_resource "gopkg.in/fsnotify.v0" do
-      url "https://github.com/go-fsnotify/fsnotify.git",
-        :branch => "v0",
-        :revision => "ea925a0a47d225b2ca7f9932b01d2ed4f3ec74f6"
-    end
-  end
-
-  # Install version of fsnotify needed by head
-  head do
-    url "https://github.com/spf13/hugo.git"
-    go_resource "gopkg.in/fsnotify.v1" do
-      url "https://github.com/go-fsnotify/fsnotify.git",
-        :revision => "6549b98005f3e4026ad9f50ef7d5011f40ba1397"
-    end
-  end
+  head "https://github.com/spf13/hugo.git"
 
   bottle do
     cellar :any
@@ -44,8 +16,6 @@ class Hugo < Formula
   end
 
   depends_on "go" => :build
-  depends_on "bazaar" => :build
-  depends_on :hg => :build
 
   go_resource "bitbucket.org/pkg/inflect" do
     url "https://bitbucket.org/pkg/inflect",
@@ -54,7 +24,7 @@ class Hugo < Formula
   end
   go_resource "github.com/BurntSushi/toml" do
     url "https://github.com/BurntSushi/toml.git",
-        :revision => "443a628bc233f634a75bcbdd71fe5350789f1afa"
+        :revision => "056c9bc7be7190eaa7715723883caffa5f8fa3e4"
   end
   go_resource "github.com/dchest/cssmin" do
     url "https://github.com/dchest/cssmin.git",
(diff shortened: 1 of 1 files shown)


DomT4:
I presume the new dependencies don't upset `stable` at all?

dunn:
Stable still builds and runs fine for me.  I think the extra deps are just ignored.

DomT4:
@BrewTestBot test this please

dunn:
0.14 just came out—let me just add that to this PR.

Library/Formula/hugo.rb
```diff
@@ -195,8 +175,10 @@ def install
     ln_sf buildpath, buildpath/"src/github.com/spf13/hugo"
     Language::Go.stage_deps resources, buildpath/"src"
 
-    system "go", "build", "main.go"
-    bin.install "main" => "hugo"
+    system "go", "build", "-o", bin/"hugo", "main.go"
+
+    bash_completion.mkpath
+    system bin/"hugo", "genautocomplete", "--completionfile=#{bash_completion}/hugo.sh"
```
dunn: Should this be in `postinstall`?

mikemcquaid: Unless it's the same for every system/install location: yes.

dunn: For me at least, when it's in `post_install` Homebrew doesn't automatically link the completion script into `/usr/local/etc/bash_completion.d`. I have to `brew unlink hugo && brew link hugo`. Is that expected?

dunn: The user _can_ install the completion script anywhere, but I figure if they're installing hugo with Homebrew it ought to be placed in `bash_completion`.

mikemcquaid: @dunn You probably want to use the `bash_completion.install` and generate it into the buildpath instead. Can you check if the script contains any hardcoded references to the full path? Thanks!

dunn: Not seeing any, no!

mikemcquaid:
Cool. Should be good to go!

DomT4:
:ship:'d in https://github.com/Homebrew/homebrew/commit/b85f1e4f45b6c6b59da84b050b99da3c0be74a41. Thanks again @dunn!

dylancarlson:
Just FYI.   This version of hugo breaks some stuff (path names in my case).  I'm not sure why upstream is making such breaking changes (also making slight variable name changes such as .BaseUrl -> .BaseURL ...) but that was a little vexing.

I had to revert to 0.13 since I don't intend to touch my content to work around this.   

As an aside, I can't fathom why 'versions' was removed from brew.

mikemcquaid:
@dylancarlson Can you provide a reproducible test case demonstrating the breakage (and we may roll back)?

> As an aside, I can't fathom why 'versions' was removed from brew.

It didn't work well and created more problems than it solved. `homebrew/versions` is the better solution.

dylancarlson:
It is covered here.   https://github.com/spf13/hugo/issues/1176

We will have to agree to disagree about homebrew/versions  :)   Maybe I can't see the rationale behind stuffing versions into a separate repo when they are already in the active repo.  It's just adding developer overhead and user confusion.  'versions' made sense, IMO

mikemcquaid:
They are already in the master repository but we are unable to fix bugs, security issues and outright errors because they are immutable once they are in the history.

Maybe we could just apply https://github.com/spf13/hugo/commit/be7404.patch as a patch? If so, could you try and open a pull request? This should help and I'm happy to walk you through anything else: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged

Thanks!

dylancarlson:
I'll follow up a bit later and open a new issue in that case since we are getting off-track in this one.   I didn't want to clean up .Url -> .URL, .BaseUrl -> .BaseURL (et al) either since I have to do it for themes on multiple sites.

Re: master repo, I still disagree.  'versions' was just a convenience feature.  I think it would have been better to modify 'versions' with a warning to people using it.  Now we're just forcing people to git log, so the only thing being done is making it harder on people who need to revert to an older version in history.  Most packages, hugo included, aren't in homebrew-versions and most probably won't be, due to the extra step and effort in maintaining them.   

There are many varied reasons why users may need to checkout an older commit, and packagers can't account for all those reasons.  Having old versions in a separate repo is just bloating the workload in all directions IMO.    I used to do a ton of packaging in FreeBSD and Gentoo Linux years ago, I am familiar with what can happen.   Maybe this is a philosophical complaint or something.  It's trivial to make 'versions' a separate tool, but I'll always see this as an intrinsic feature of the packaging tool.

mikemcquaid:
> Having old versions in a separate repo is just bloating the workload in all directions IMO.

The workload for Homebrew maintainers is significantly lower now.

> I think it would have been better to modify 'versions' with a warning to people using it. 

We did. It was ignored.

> Now we're just forcing people to git log, so the only thing being done is making it harder on people who need to revert to an older version in history.

I don't think using `git log` is particularly hard, personally. If anything it's a suitable barrier to entry for people to think about what they are doing rather than running commands blindly and breaking things (which happened relatively often).

> I'll follow up a bit later and open a new issue in that case since we are getting off-track in this one.

Thanks. Ideally a PR rather than an issue.

dylancarlson:
As I say, it might be philosophical.  :)  If it were me, I wouldn't remove features for the sake of keeping a minority of people from shooting themselves in the feet.  Screwing up is part of learning also.  I'm not sure homebrew's audience is everyone, as the App Store is for everyone.   Doing installs from a terminal implies a certain level of technical competence.

mikemcquaid:
Oh, we don't so much mind people shooting themselves into the foot but we do mind making our own jobs much harder when debugging unrelated issues (which it often did).

dylancarlson:
Fair enough, thanks for the discussion & for working on homebrew.  I have no idea what kind of workload you have.  ```brew log package``` with a simple wrapper is easy for me to hack up.

I would imagine ```brew pin``` imposes similar workload complications, albeit from a different angle.

Just a thought, if I was in that position, what I might add a 'support' argument to ```brew``` for people to attach on new issues, which is file containing a json representation complete with ```brew info``` on each package.  Feed that to a tool which computes the delta of their local state against master.  That would help quickly flag users not conforming to a supportable environment.

DomT4:
> Most packages, hugo included, aren't in homebrew-versions and most probably won't be, due to the extra step and effort in maintaining them.

Just as a heads-up, If there's a valid use case it's perfectly welcome in Versions. There's stuff in Versions that hasn't been touched in 2+ years and still works on all three Homebrew-supported OS X versions without patching/etc. It's less prone to breakage than the alternative options.

Submitting a formula to Versions that was in the core is as simple as pulling formula from the git history, removing any bottle/head blocks, fixing `brew audit --strict` and then submitting it. After that it should be good to go.

dylancarlson:
@DomT4 thanks.  I don't think this formula will come to that.  0.14 was a feature release, hugo releases every 3-4 months so we should see this fixed in the next by ~September.  If other people complain about this issue, I'll be happy to do a PR.  For now I've rolled back to 0.13 and pinned it.  

For anyone else who has the same problem...

```shell
$ brew uninstall hugo
$ cd /usr/local
$ git checkout 03a74b44ebe763991a955062408a796aa1fd982c /usr/local/Library/Formula/hugo.rb
$ brew install hugo
$ brew pin hugo
$ git stash
```

PR #39821, opened 2015-05-16
sqsh: correct homepage, test

dunn:
The current homepage 404s and from an email to the page owner it seems he's no longer the maintainer of sqsh, if he ever was—it might have just been a how-to page he wrote.


mikemcquaid:
Thanks @dunn!

PR #39826, opened 2015-05-16
abcm2ps 8.7.8 (devel)

pthariensflame:
Unfortunately, I couldn’t convert the download URLs to `https`; the site
doesn’t seem to support it.  :(


vszakats:
Thanks for looking after this @pthariensflame

mikemcquaid:
Thanks again @pthariensflame!

PR #39827, opened 2015-05-16
abcmidi 2015-05-13

pthariensflame:
Also modernized the formula slightly.


mikemcquaid:
Thanks @pthariensflame!

PR #36835, opened 2015-02-15
Enable installing completion scripts for fish.

MerelyAPseudonym:
It would be nice if Homebrew kept fish-completion scripts for formulae that provided them (e.g., youtube-dl).

N.B. the current stable release of fish does *not* use `$(brew --prefix)/share/fish/vendor_completions.d`, but HEAD does and a new release is coming out soon-ish.

PR #39828, opened 2015-05-16
Add support for installing fish completions

languitar:
This pull request introduces support for installing completions for the fish shell comparable to bash and zsh completions. I have adapted formulae where I know about existing fish completions accordingly. Would be glad if this was accepted.

### DIFF
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -48,7 +48,7 @@ class Caveats
   end
 
   def fish_completion_caveats
-    if keg and keg.completion_installed? :fish then <<-EOS.undent
+    if keg and keg.completion_installed? :fish and which("fish") then <<-EOS.undent
       fish completion has been installed to:
         #{HOMEBREW_PREFIX}/share/fish/vendor_completions.d
       EOS
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -12,6 +12,7 @@ class Caveats
     caveats << f.keg_only_text if f.keg_only? && f.respond_to?(:keg_only_text)
     caveats << bash_completion_caveats
     caveats << zsh_completion_caveats
+    caveats << fish_completion_caveats
     caveats << plist_caveats
     caveats << python_caveats
     caveats << app_caveats
@@ -42,6 +43,14 @@ class Caveats
     if keg and keg.completion_installed? :zsh then <<-EOS.undent
       zsh completion has been installed to:
         #{HOMEBREW_PREFIX}/share/zsh/site-functions
+      EOS
+    end
+  end
+
+  def fish_completion_caveats
+    if keg and keg.completion_installed? :fish then <<-EOS.undent
+      fish completion has been installed to:
+        #{HOMEBREW_PREFIX}/share/fish/vendor_completions.d
       EOS
     end
   end
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -397,13 +397,19 @@ class Formula
   # installed.
   # This is symlinked into `HOMEBREW_PREFIX` after installation or with
   # `brew link` for formulae that are not keg-only.
-  def bash_completion; prefix+'etc/bash_completion.d' end
+  def bash_completion; prefix+'etc/bash_completion.d'    end
 
   # The directory where the formula's ZSH completion files should be
   # installed.
   # This is symlinked into `HOMEBREW_PREFIX` after installation or with
   # `brew link` for formulae that are not keg-only.
-  def zsh_completion;  share+'zsh/site-functions'     end
+  def zsh_completion;  share+'zsh/site-functions'        end
+
+  # The directory where the formula's fish completion files should be
+  # installed.
+  # This is symlinked into `HOMEBREW_PREFIX` after installation or with
+  # `brew link` for formulae that are not keg-only.
+  def fish_completion; share+'fish/vendor_completions.d' end
 
   # The directory used for as the prefix for {#etc} and {#var} files on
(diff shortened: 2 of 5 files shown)


Library/Homebrew/caveats.rb
```diff
@@ -46,6 +47,14 @@ def zsh_completion_caveats
     end
   end
 
+  def fish_completion_caveats
+    if keg and keg.completion_installed? :fish then <<-EOS.undent
+      fish completion has been installed to:
+        #{HOMEBREW_PREFIX}/etc/fish/completions
+      EOS
```
mikemcquaid: Not sure what I think about this considering the relatively niche popularity of fish. Maybe worth sniffing for what the user's shell is?

languitar: On the other hand this will only pop up on a couple of packages, because not many packages offer fish completion so far.

share/doc/homebrew/Tips-N'-Tricks.md
```diff
@@ -50,6 +50,14 @@ Run in terminal (may require `sudo`):
...
 ```
 
+
+### fish
+Add to your `~/.config/fish/config.fish`:
+
+```fish
+set -x fish_complete_path (brew --prefix)/etc/fish/completions $fish_complete_path
+```
+
```
xu-cheng: Isn't this section about shell auto-compelete for brew command not formulae?

languitar:
I have removed the "wrong" documentation from the tips and tricks file.

bfontaine:
@Homebrew/owners @DomT4 Any thoughts on this?

DomT4:
I agree it's a niche addition, but I don't have a problem with it.

There is an argument for masking less mainstream shells sit behind a variable or conditional, but how we would cater that could be troublesome since not every compile plays nicely with `fish` so glancing in the `env` for Fish variables wouldn't be hugely reliable.

mikemcquaid:
We should definitely be checking for a fish installation somehow rather than showing this to all users. @mistydemeo, do you use `fish`? I had some vague recollection you did.

DomT4:
Any dice wrapping it in something like a `which("fish")` call and not printing it if we don't find Fish on the system?

mikemcquaid:
That makes sense to me.

languitar:
Shouldn't this be used canonincally then for all shells completions?

DomT4:
`zsh` and `sh`/`bash` both ship natively with OS X.

mikemcquaid:
@DomT4 :+1:

mistydemeo:
> @mistydemeo, do you use fish? I had some vague recollection you did.

I do, yes!

Doing this if fish is installed makes sense to me.

mikemcquaid:
Yeh, install unconditionally, display the caveats conditionally works for me.

languitar:
I'll have a look how to implement this during the next days.

languitar:
I've pushed a proposal. What do you think?

Library/Homebrew/caveats.rb
```diff
@@ -46,6 +47,18 @@ def zsh_completion_caveats
     end
   end
 
+  def has_fish
+      return system("fish -v > /dev/null 2> /dev/null")
+  end
+
+  def fish_completion_caveats
+    if keg and keg.completion_installed? :fish and has_fish then <<-EOS.undent
```
xu-cheng: I think `which("fish")` would be enough. And you can remove the `has_fish` method.

languitar:
Adapted according to @xu-cheng s comment

DomT4:
Local testing looks good, :+1:.

Library/Homebrew/caveats.rb
```diff
@@ -46,6 +47,14 @@ def zsh_completion_caveats
     end
   end
 
+  def fish_completion_caveats
+    if keg and keg.completion_installed? :fish and which("fish") then <<-EOS.undent
```
mikemcquaid: Let's use `&&` instead of `and` here, thanks and consider moving the `<<-EOS.undent` onto a new line.

Library/Homebrew/formula.rb
```diff
@@ -405,6 +405,12 @@ def bash_completion; prefix+'etc/bash_completion.d' end
   # `brew link` for formulae that are not keg-only.
   def zsh_completion;  share+'zsh/site-functions'     end
 
+  # The directory where the formula's fish completion files should be
+  # installed.
+  # This is symlinked into `HOMEBREW_PREFIX` after installation or with
+  # `brew link` for formulae that are not keg-only.
+  def fish_completion;  prefix+'etc/fish/completions'     end
```
mikemcquaid: Indent the `end` with the `end` from `zsh_completion` above.

mikemcquaid:
A few tweaks but otherwise :+1:

Library/Homebrew/caveats.rb
languitar: But the other completion functions also use `and`? That would be inconsistent then?

mikemcquaid: Yeh, that's fine, thanks.

MerelyAPseudonym:
@languitar how would you feel about using `share/fish/vendor_completions.d` instead of `etc/fish/completions`? [It'll be the default in the next release.](https://github.com/fish-shell/fish-shell/issues/1485)

languitar:
@MerelyAPseudonym interesting, I was actually thinking about requesting exactly this when implementing this feature for homebrew. So it's probably now a good idea to use that directory. I'll updated the branch accordingly.

languitar:
I think I addressed all comments now.

mikemcquaid:

languitar:
Thanks! Actually I think this also fixes #36835 apart from a completion installation for a formula. Maybe this could then also be resolved.

PR #39829, opened 2015-05-17
fontforge: add option to install extra tools

DomT4:
Adds an option to install the extra Font tools. These used to be part of Fontforge by default but were moved to an optional location last year. I believe they are still packaged if you do `make dist` but obviously, we don't.

Upstream requested we add the option, and it's potentially quite useful, so here's a PR.

### DIFF
--- a/Library/Formula/fontforge.rb
+++ b/Library/Formula/fontforge.rb
@@ -11,6 +11,7 @@ class Fontforge < Formula
   end
 
   option "with-giflib", "Build with GIF support"
+  option "with-extra-tools", "Build with additional font tools"
 
   deprecated_option "with-x" => "with-x11"
   deprecated_option "with-gif" => "with-giflib"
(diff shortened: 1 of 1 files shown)


DomT4:
```
==> brew audit fontforge						 FAILED
fontforge:
 * python modules have explicit framework links
These python extension modules were linked directly to a Python
framework binary. They should be linked with -undefined dynamic_lookup
instead of -lpython or -framework Python.
  /usr/local/Cellar/fontforge/20150430/lib/python2.7/site-packages/fontforge.so
  /usr/local/Cellar/fontforge/20150430/lib/python2.7/site-packages/psMat.so
```

What do I do about this @tdsmith? Not sure how we handle it.

tdsmith:
Depends! Ideally, send a patch to fontforge to fix their build and remove the :python dependency. :) I assume people sometimes want to `import fontforge`?

tdsmith:
(I'll add this to my to-do list; we can ignore this for now.)

DomT4:
> I assume people sometimes want to import fontforge?

Yeah, It's a useful upstream feature. As well as being able to `import fontforge` there's Python usage written into the application.

DomT4:
Merged per Tim's :+1: on ignoring the audit failure for now.

PR #39831, opened 2015-05-17
spoof-mac 2.0.4

dunn:


Library/Formula/spoof-mac.rb
```diff
@@ -14,15 +15,15 @@ class SpoofMac < Formula
...
     url "https://pypi.python.org/packages/source/d/docopt/docopt-0.6.2.tar.gz"
-    sha1 "224a3ec08b56445a1bd1583aad06b00692671e04"
+    sha256 "49b3a825280bd66b3aa83585ef59c4a8c82f2c8a522dbe754a8bc8d08c85c491"
   end
 
   def install
     ENV["PYTHONPATH"] = libexec/"lib/python2.7/site-packages"
     ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
 
-    resources.each do |r|
```
DomT4: I left it this broad before in case any future resources are added, that way the language just expands to fit automatically, but I've no objections to the change if maintainers are happy with it.

dunn: Yeah I can revert that; looks cleaner this way but could make it harder for the next person to update the formula.

mikemcquaid: Yeh, let's leave it as-is :+1:

PR #39836, opened 2015-05-17
Gems,-Eggs-and-Perl-Modules: link to MetaCPAN

dolmen:
Link to MetaCPAN.org instead of search.cpan.org. MetaCPAN is the community-maintained web portal to CPAN, while search.cpan.org is the historic but proprietary one.

### DIFF
--- a/share/doc/homebrew/Gems,-Eggs-and-Perl-Modules.md
+++ b/share/doc/homebrew/Gems,-Eggs-and-Perl-Modules.md
@@ -107,14 +107,14 @@ Perl CPAN Modules without sudo
 The Perl module local::lib works similarly to rbenv/RVM (although for
 modules only, not perl installations). A simple solution that only
 pollutes your /Library/Perl a little is to install
-[local::lib](http://search.cpan.org/search?query=local::lib) with sudo:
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Isn't this just a matter of personal preference?

DomT4:
I'd be happy to see the switch if only for the SSL/TLS preference. Otherwise, :neutral_face:.

dolmen:
Of course this is a matter of taste.
But as someone deeply involved in the Perl community (I've [patched](/dolmen?tab=repositories) numerous CPAN modules and I'm collaborating with the [Perl Toolchain Gang](//Perl-Toolchain-Gang)), I have the arrogance to claim that my taste has some weight.
Of course I'm biased because [I've sent some tickets and some patches to MetaCPAN.org](/CPAN-API/metacpan-web/issues?utf8=✓&q=author%3Adolmen) to improve it, but I've not done the same for search.cpan.org just because there is no possibility to do that as the code is closed and the feedback mailbox is just a back hole.

mikemcquaid:
Fair points. Merging, thanks!

PR #39842, opened 2015-05-17
ldns: also build example tools because they are very useful

Habbie:


Library/Formula/ldns.rb
```diff
@@ -4,7 +4,7 @@ class Ldns < Formula
   homepage 'http://nlnetlabs.nl/projects/ldns/'
   url 'http://nlnetlabs.nl/downloads/ldns/ldns-1.6.17.tar.gz'
```
vszakats: Could you also update both `homepage` and `url` to use `https` protocol?

Habbie:
Amended with https links

Library/Formula/ldns.rb
```diff
@@ -1,10 +1,10 @@
 require 'formula'
 
 class Ldns < Formula
-  homepage 'http://nlnetlabs.nl/projects/ldns/'
-  url 'http://nlnetlabs.nl/downloads/ldns/ldns-1.6.17.tar.gz'
+  homepage 'https://nlnetlabs.nl/projects/ldns/'
+  url 'https://nlnetlabs.nl/downloads/ldns/ldns-1.6.17.tar.gz'
   sha1 '4218897b3c002aadfc7280b3f40cda829e05c9a4'
-  revision 1
+  revision 2
```
mikemcquaid: Leave the `revision` as-is, thanks.

vszakats:
Thank you!

Habbie:
Revision reverted. Shall I also address all other (older) complaints from `style` and `audit --strict`?

mikemcquaid:
@Habbie If you're happy to: yeh, thanks!

Habbie:
All old nits addressed; added a test based on one of the example tools. Not sure about the overly long lines, comments welcome as always.

DomT4:
Are the examples installed automatically by `make install` here? They don't require any added formula dances to install?

Habbie:
Yes, just the configure changes makes the building and installation of the examples work.

DomT4:
Nice. That's relatively rare, usually we have to do a whole dance to fish the compiled examples out of buildpath :sweat_smile:.

Habbie:
Yes, I was as surprised - last time I looked (years ago) there was no `--with-examples` in ldns indeed.

mikemcquaid:

For future reference the preferred commit message format is `ldns: build example tools.`.

PR #39844, opened 2015-05-17
Updated snappystream to 0.2.1

theirix:
Updated snappystream formula from 0.1.2 to 0.2.1.

### DIFF
--- a/Library/Formula/snappystream.rb
+++ b/Library/Formula/snappystream.rb
@@ -17,11 +17,11 @@ class Snappystream < Formula
   depends_on "boost" => :optional
 
   def install
-    args = [ ".", "-DBUILD_TESTS=ON" ]
-    if build.with? "boost"
-      args << "-DWITH_BOOST_IOSTREAMS=1"
-    end
-    system "cmake", *args, *std_cmake_args
+    args = std_cmake_args
+    args << "."
+    args << "-DBUILD_TESTS=ON"
+    args << "-DWITH_BOOST_IOSTREAMS=1" if build.with? "boost"
+    system "cmake", *args
     system "make", "all", "test", "install"
   end
 
--- a/Library/Formula/snappystream.rb
+++ b/Library/Formula/snappystream.rb
@@ -14,10 +14,14 @@ class Snappystream < Formula
 
   depends_on "cmake" => :build
   depends_on "snappy"
-  depends_on "boost"
+  depends_on "boost" => :optional
 
   def install
-    system "cmake", ".", "-DBUILD_TESTS=ON", "-DWITH_BOOST_IOSTREAMS=1", *std_cmake_args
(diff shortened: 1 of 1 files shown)


Library/Formula/snappystream.rb
```diff
@@ -14,9 +14,10 @@ class Snappystream < Formula
 
   depends_on "cmake" => :build
   depends_on "snappy"
+  depends_on "boost"
```
mikemcquaid: Is this a build-time or run-time dependency? It's an annoying big one if run-time :sob:

Library/Formula/snappystream.rb
```diff
@@ -14,9 +14,10 @@ class Snappystream < Formula
...
 
   def install
-    system "cmake", ".", "-DBUILD_TESTS=ON", *std_cmake_args
+    system "cmake", ".", "-DBUILD_TESTS=ON", "-DWITH_BOOST_IOSTREAMS=1", *std_cmake_args
```
mikemcquaid: Does not having this option set remove the need for the `boost` dependency?

Library/Formula/snappystream.rb
theirix: Boost support in snappystream means that snappystream provides a few interfaces to work with boost iterators library. So this dependency is runtime. But it can be turned on/off by the WITH_BOOST_IOSTREAMS flag. Should we make it _:optional_?

Library/Formula/snappystream.rb
theirix: No, it just enables linking with boost and adds preprocessor defines to the project.

Library/Formula/snappystream.rb
mikemcquaid: Yep, please.

theirix: Just commited an optional dependency.

Library/Formula/snappystream.rb
```diff
@@ -14,9 +14,14 @@ class Snappystream < Formula
...
   depends_on "snappy"
+  depends_on "boost" => :optional
 
   def install
-    system "cmake", ".", "-DBUILD_TESTS=ON", *std_cmake_args
+    args = [ ".", "-DBUILD_TESTS=ON" ]
+    if build.with? "boost"
+      args << "-DWITH_BOOST_IOSTREAMS=1"
+    end
+    system "cmake", *args, *std_cmake_args
```
mikemcquaid: This won't work on Ruby 1.8 (Mountain Lion) sadly. Instead:

```ruby
args = std_cmake_args
args << "-DBUILD_TESTS=ON"
args << "-DWITH_BOOST_IOSTREAMS=1" if build.with? "boost"
system "cmake", *args, *std_cmake_args
```

theirix: Thank you, totally forgot about asterisk expansion. Changed suggested code a little.

PR #39845, opened 2015-05-17
[SECURITY] Update qemu to 2.3.0 and fix VENOM

Midar:
This updates qemu to 2.3.0 and applies the fix for [VENOM](http://venom.fail/).

### DIFF
--- a/Library/Formula/qemu.rb
+++ b/Library/Formula/qemu.rb
@@ -1,13 +1,20 @@
 class Qemu < Formula
   homepage "http://wiki.qemu.org"
-  url "http://wiki.qemu-project.org/download/qemu-2.2.1.tar.bz2"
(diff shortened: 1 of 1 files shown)


Library/Formula/qemu.rb
```diff
@@ -10,6 +10,12 @@ class Qemu < Formula
     sha256 "805da9cd1c6193e345245a4c33dc090fd8b541190c56d64acb927cb37e794d50" => :mountain_lion
   end
 
+  patch do
+    # Fix VENOM
```
mikemcquaid: Can you add more detail in this comment e.g. where this patch is from/applied, when it will be needed until and what VENOM is? Thanks!

Midar: Done. I think linking to venom.fail should be more then enough, considering that bug is all over the news ;).

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `qemu 2.3.0`.

PR #39850, opened 2015-05-17
gtk-doc 1.23

tschoonj:
version bump
audit --strict reports that this is a development release but this is not true!

### DIFF
--- a/Library/Formula/gtk-doc.rb
+++ b/Library/Formula/gtk-doc.rb
@@ -1,7 +1,7 @@
 class GtkDoc < Formula
   homepage "http://www.gtk.org/gtk-doc/"
-  url "https://download.gnome.org/sources/gtk-doc/1.22/gtk-doc-1.22.tar.xz"
-  sha256 "932865c912ce0d81c9480bf957d3908bae6e18c0cf2ee33014d63b920047fca8"
+  url "https://download.gnome.org/sources/gtk-doc/1.23/gtk-doc-1.23.tar.xz"
+  sha256 "0b5c23711166c10ff5c74603db80ac26c2b9a382ce778b0e795db821d50718c4"
 
   bottle do
     sha256 "654ae76d5e15f8461dbfd0be4ed8c52d67e7cc2dd4e8f4fc893fc9e5feedcd2f" => :yosemite


DomT4:
> audit --strict reports that this is a development release but this is not true!

Are you sure? GTK pretty much always number their projects so even = stable and odd = devel.

tschoonj:
Mostly true but not for gtk-doc. They released this one so quickly after the last because it was very buggy. Check out the formula history and you will see that this formula has used odd version numbers in the past



Op 17-mei-2015 om 20:38 heeft Dominyk Tiller <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


audit --strict reports that this is a development release but this is not true!

Are you sure? GTK pretty much always number their projects so even = stable and odd = devel.

DomT4:
@jacknagel Can you okay the above? It looks like we have merged odd versions of gtk-doc before now, but want to check was intentional.

tschoonj:
By the way the same problem occurs for `pygtkglext`: 1.1.0 is considered to be a stable release

jacknagel:
I added the check because we were getting a lot of unstable versions submitted as updates, and I wanted to avoid having to explain the stable/unstable pattern every time. If there are legitimate exceptions to the rule, that's fine.

It would be nice if they would follow the established pattern, though. Shouldn't bug fixes be 1.22.1 rather than 1.23?

tschoonj:
You're right of course, but for some reason they use only two version numbers for gtk-doc

DomT4:
> I added the check because we were getting a lot of unstable versions submitted as updates

Maybe we could exempt certain formulae from the check if we know they don't follow the regular GTK versioning?

Merged this in https://github.com/Homebrew/homebrew/commit/cb755db115fa82444598f8c7efdb0a90e6ad5f82. Looked over the upstream website and looks like a stable release, and we've merged odd versions before. We should try and get upstream to clarify somewhere that they permit odd numbers for stable versions though.

Thanks for your contribution to Homebrew, we appreciate it :smiley_cat:.

mikemcquaid:
> Maybe we could exempt certain formulae from the check if we know they don't follow the regular GTK versioning?

Yeh, it'd be good to whitelist certain formulae like this in the audit. We use that pattern elsewhere.

PR #39852, opened 2015-05-17
open-mpi 1.8.5

barrywardell:
Including workaround for a memory leak which affects 1.8.5.

### DIFF
--- a/Library/Formula/open-mpi.rb
+++ b/Library/Formula/open-mpi.rb
@@ -1,9 +1,8 @@
 class OpenMpi < Formula
   homepage "https://www.open-mpi.org/"
-  # Wait for 1.8.6 and skip 1.8.5 due to a severe memory leak on OS X:
-  # https://github.com/open-mpi/ompi/issues/579
-  url "https://www.open-mpi.org/software/ompi/v1.8/downloads/openmpi-1.8.4.tar.bz2"
-  sha256 "23158d916e92c80e2924016b746a93913ba7fae9fff51bf68d5c2a0ae39a2f8a"
+  url "https://www.open-mpi.org/software/ompi/v1.8/downloads/openmpi-1.8.5.tar.bz2"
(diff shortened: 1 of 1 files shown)


Library/Formula/open-mpi.rb
```diff
@@ -1,9 +1,8 @@
 class OpenMpi < Formula
   homepage "https://www.open-mpi.org/"
-  # Wait for 1.8.6 and skip 1.8.5 due to a severe memory leak on OS X:
-  # https://github.com/open-mpi/ompi/issues/579
-  url "https://www.open-mpi.org/software/ompi/v1.8/downloads/openmpi-1.8.4.tar.bz2"
-  sha256 "23158d916e92c80e2924016b746a93913ba7fae9fff51bf68d5c2a0ae39a2f8a"
+  url "https://www.open-mpi.org/software/ompi/v1.8/downloads/openmpi-1.8.5.tar.bz2"
+  sha256 "4cea06a9eddfa718b09b8240d934b14ca71670c2dc6e6251a585ce948a93fbc4"
+  revision 1
```
mikemcquaid: Please remove the `revision`; it's not needed if we have a version bump.

Library/Formula/open-mpi.rb
```diff
@@ -1,9 +1,8 @@
 class OpenMpi < Formula
   homepage "https://www.open-mpi.org/"
-  # Wait for 1.8.6 and skip 1.8.5 due to a severe memory leak on OS X:
-  # https://github.com/open-mpi/ompi/issues/579
```
mikemcquaid: I'd rather we waited for 1.8.6 instead.

barrywardell:
Apparently 1.8.6 is coming soon (according to https://github.com/open-mpi/ompi/issues/579#issuecomment-104928398), so it makes sense to wait for 1.8.6 instead.

PR #39854, opened 2015-05-17
vobcopy: retire to boneyard

L2G:
Regarding libdvdcss: while it is a dependency of libdvdread, vobcopy does not appear to link to it.

```
$ otool -L `which vobcopy`
/Users/larry/work/GitHub/homebrew/_core_/bin/vobcopy:
	/Users/larry/work/GitHub/homebrew/_core_/lib/libdvdread.4.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
```


Library/Formula/vobcopy.rb
```diff
@@ -14,6 +11,11 @@ def install
                              "--with-lfs",
                              "--with-dvdread-libs=#{HOMEBREW_PREFIX}"
     system "make"
-    system "make install"
+    system "make", "install"
+  end
+
+  test do
+    output = shell_output("#{bin}/vobcopy -h 2>&1", 1)
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

L2G: The most realistic test would be to have the user keep a movie DVD in their DVD drive. Saving that, we'd have to create a VOB file from a DVD. I don't know if you want a test file that big in the test fixtures...

L2G: I'd've done that if I could think of a more practical way to do it than to require users to keep a rippable movie DVD in their DVD drives.

mikemcquaid: Could we download a (very) small `vob` as a resource?

DomT4:
This needs rebasing on master. Downloading and using a small `vob` file for testing as Mike suggests would be cool as well.

Library/Formula/vobcopy.rb
L2G: I found [some VOB files at digital-digest.com](http://www.digital-digest.com/movies/movie_index.php?type=dolby) that look promising.

L2G: It looks like what's really needed is the complete file tree from a DVD image. It can probably be in a ZIP file, but it needs all the .vob/.nfo/&c. files.

L2G: I found something promising from [W6RZ](http://w6rz.net/): a DVD test-pattern disc image. But it's a ZIP of an ISO image. I don't know what to do to get at the ISO filesystem.

DomT4: That looks like quite a big file. If we can't find anything smaller, I'm tempted to just pull this without an improved test.

L2G: Two things:

1. I whipped up a test disc using iDVD, saved it as a file system, and tarred it up (compressed with xz). The resulting tarball is 68 MB.
2. I can make a simple test for it, but it fails with "Floating point exception: 8". That seems like a problem much deeper than a build issue. Or it could just mean that vobcopy can't handle anything iDVD produces. :cry:

L2G: I could not get a working test based on the W6RZ DVD, either. Maybe this thing has just rotted too long and needs to be retired to the boneyard…

DomT4: Well, It's been 4 years since it was last touched in Homebrew. If you're having runtime issues, I'm happy enough to see it boneyarded - Nothing in the core depends on it. CC @mikemcquaid

mikemcquaid: Yeh, boneyard.

L2G:
Copied to boneyard in Homebrew/homebrew-boneyard#55.

DomT4:
Dead in https://github.com/Homebrew/homebrew/commit/7c7529a0dd8fd9100407cb15e1be621a853f1b7a. Thanks @L2G!

PR #39855, opened 2015-05-17
jdnssec-tools 0.12

Habbie:
This is a binary (jar) formula - I tried building from source but this project uses ant instead of maven so that would mean also packaging up two dependencies which seemed like overkill to me - and to make it worse, the author forked those dependencies in an incompatible way to their originals.

### DIFF
--- a/Library/Formula/jdnssec-tools.rb
+++ b/Library/Formula/jdnssec-tools.rb
@@ -15,7 +15,15 @@ class JdnssecTools < Formula
   end
 
   test do
-    (testpath/"powerdns.com.key").write("powerdns.com.   10773 IN  DNSKEY  257 3 8 AwEAAb/+pXOZWYQ8mv9WM5dFva8WU9jcIUdDuEjldbyfnkQ/xlrJC5zA EfhYhrea3SmIPmMTDimLqbh3/4SMTNPTUF+9+U1vpNfIRTFadqsmuU9F ddz3JqCcYwEpWbReg6DJOeyu+9oBoIQkPxFyLtIXEPGlQzrynKubn04C x83I6NfzDTraJT3jLHKeW5PVc1ifqKzHz5TXdHHTA7NkJAa0sPcZCoNE 1LpnJI/wcUpRUiuQhoLFeT1E432GuPuZ7y+agElGj0NnBxEgnHrhrnZW UbULpRa/il+Cr5Taj988HqX9Xdm6FjcP4Lbuds/44U7U8du224Q8jTrZ 57Yvj4VDQKc=")
-    assert_match /D4C3D5552B8679FAEEBC317E5F048B614B2E5F607DC57F1553182D49AB2179F7/, shell_output("#{bin}/jdnssec-dstool -d 2 powerdns.com.key")
+    (testpath/"powerdns.com.key").write(
+      "powerdns.com.   10773 IN  DNSKEY  257 3 8 (AwEAAb/+pXOZWYQ8mv9WM5dFva8
+      WU9jcIUdDuEjldbyfnkQ/xlrJC5zA EfhYhrea3SmIPmMTDimLqbh3/4SMTNPTUF+9+U1vp
+      NfIRTFadqsmuU9F ddz3JqCcYwEpWbReg6DJOeyu+9oBoIQkPxFyLtIXEPGlQzrynKubn04
+      Cx83I6NfzDTraJT3jLHKeW5PVc1ifqKzHz5TXdHHTA7NkJAa0sPcZCoNE 1LpnJI/wcUpRU
+      iuQhoLFeT1E432GuPuZ7y+agElGj0NnBxEgnHrhrnZW UbULpRa/il+Cr5Taj988HqX9Xdm
+      6FjcP4Lbuds/44U7U8du224Q8jTrZ 57Yvj4VDQKc=)")
+
+    assert_match /D4C3D5552B8679FAEEBC317E5F048B614B2E5F607DC57F1553182D49AB2179F7/,
+      shell_output("#{bin}/jdnssec-dstool -d 2 powerdns.com.key")
   end
 end
--- a/Library/Formula/jdnssec-tools.rb
+++ b/Library/Formula/jdnssec-tools.rb
@@ -1,0 +1,21 @@
+class JdnssecTools < Formula
+  desc "Java command-line tools for DNSSEC"
+  homepage "http://www.verisignlabs.com/jdnssec-tools/"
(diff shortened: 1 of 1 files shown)


Habbie:
Now has a test, passes `audit --strict`. Ready to merge?

Library/Formula/jdnssec-tools.rb
```diff
@@ -0,0 +1,21 @@
...
+  def install
+    inreplace Dir["bin/*"] do |s|
+      s.gsub! /basedir=.*/, "basedir=#{libexec}"
+    end
+    bin.install Dir["bin/*"]
+    (libexec/"lib").install Dir["lib/*"]
+  end
+
+  test do
+    (testpath/"powerdns.com.key").write("powerdns.com.   10773 IN  DNSKEY  257 3 8 AwEAAb/+pXOZWYQ8mv9WM5dFva8WU9jcIUdDuEjldbyfnkQ/xlrJC5zA EfhYhrea3SmIPmMTDimLqbh3/4SMTNPTUF+9+U1vpNfIRTFadqsmuU9F ddz3JqCcYwEpWbReg6DJOeyu+9oBoIQkPxFyLtIXEPGlQzrynKubn04C x83I6NfzDTraJT3jLHKeW5PVc1ifqKzHz5TXdHHTA7NkJAa0sPcZCoNE 1LpnJI/wcUpRUiuQhoLFeT1E432GuPuZ7y+agElGj0NnBxEgnHrhrnZW UbULpRa/il+Cr5Taj988HqX9Xdm6FjcP4Lbuds/44U7U8du224Q8jTrZ 57Yvj4VDQKc=")
```
DomT4: Can we break this onto multiple lines, or shorten it? It's currently making the formula impossible to read in a glance.

Library/Formula/jdnssec-tools.rb
```diff
@@ -0,0 +1,21 @@
...
+  url "http://www.verisignlabs.com/dnssec-tools/packages/jdnssec-tools-0.12.tar.gz"
+  sha256 "3fa0bfe062051d9048c5186adbb04d8f5e176e50da3cdcf9b0d97e1c558efc22"
+
+  depends_on :java
+
+  def install
+    inreplace Dir["bin/*"] do |s|
+      s.gsub! /basedir=.*/, "basedir=#{libexec}"
+    end
+    bin.install Dir["bin/*"]
```
DomT4: What's the reason for installing directly rather than via `bin.write_jar_script` here?

Library/Formula/jdnssec-tools.rb
Habbie: Fixed locally, will push after I look at the other thing.

Library/Formula/jdnssec-tools.rb
Habbie: Don't recall whether I tried this initially. Will look into it.

Habbie: The scripts I install now (as provided by upstream) put all three .jar files in `CLASSPATH`; I don't see (mostly because I'm not very experienced with Java) how to make the calls work without including the same `CLASSPATH` either in env or in `-cp`. Looking at other formulas (like gwt) I see not using `write_jar_script` is not uncommon for these cases.

For reference, jdnssec-dstool looks like this (and the others basically differ by one word, although some don't have the ulimit block):
```bash
#! /bin/sh

thisdir=`dirname $0`
basedir=/usr/local/Cellar/jdnssec-tools/0.12/libexec

ulimit_max=`ulimit -H -n`
if [ $ulimit_max != "unlimited" ]; then
    ulimit -n $ulimit_max
fi

# set the classpath
for i in $basedir/lib/*.jar $basedir/lib/*.zip $basedir/build/lib/*.jar; do
  CLASSPATH="$CLASSPATH":"$i"
done
export CLASSPATH

exec java com.verisignlabs.dnssec.cl.DSTool "$@"
```

Library/Formula/jdnssec-tools.rb
Habbie: Pushed this now.

Habbie:
Broke up some lines. Did not change the java/bin stuff yet, see comments on outdated diff.

Library/Formula/jdnssec-tools.rb
DomT4: Cool. That works for me at least. Are there any problems setting the `basedir` in `libexec` but installing the executables directly in `bin`?

Habbie: I'm sorry, I don't understand the question. Are you asking for clarification or suggesting a change?

DomT4: Clarification. A lot of JAVA stuff roots itself in the directory, so if you move executables from `libexec` (for example) directly to `bin` and don't move the `basedir`/`home`/etc you get failure messages on run. Just wanted to check that wasn't the case here.

Habbie: No, it works fine just like this due to the `gsub` that fixes the path.

Library/Formula/jdnssec-tools.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  test do
+    (testpath/"powerdns.com.key").write(
+      "powerdns.com.   10773 IN  DNSKEY  257 3 8 (AwEAAb/+pXOZWYQ8mv9WM5dFva8
+      WU9jcIUdDuEjldbyfnkQ/xlrJC5zA EfhYhrea3SmIPmMTDimLqbh3/4SMTNPTUF+9+U1vp
+      NfIRTFadqsmuU9F ddz3JqCcYwEpWbReg6DJOeyu+9oBoIQkPxFyLtIXEPGlQzrynKubn04
+      Cx83I6NfzDTraJT3jLHKeW5PVc1ifqKzHz5TXdHHTA7NkJAa0sPcZCoNE 1LpnJI/wcUpRU
+      iuQhoLFeT1E432GuPuZ7y+agElGj0NnBxEgnHrhrnZW UbULpRa/il+Cr5Taj988HqX9Xdm
+      6FjcP4Lbuds/44U7U8du224Q8jTrZ 57Yvj4VDQKc=)")
+    assert_match /D4C3D5552B8679FAEEBC317E5F048B614B2E5F607DC57F1553182D49AB2179F7/,
```
DomT4: Ideally there should be an empty line between the block above and this one - Helps make it look like it isn't the same thing. Tiny nit though.

DomT4:
CC @Homebrew/owners for :+1: / :-1: here.

Library/Formula/jdnssec-tools.rb
Habbie: nit fixed

mikemcquaid:

PR #39857, opened 2015-05-17
hfsutils: brew audit --strict; add test that formats and mounts HFS image

L2G:


mikemcquaid:

For future reference the preferred commit message format is `hfsutils: fix audit, add test.` (fewer than 51 characters in the title).

PR #39862, opened 2015-05-18
SourceKitten 0.4.3

jpsim:
Thanks!


mistydemeo:
Looks like the build is failing on 10.10.

jpsim:
Seems to fail with the following error:

```
cp: directory /usr/local/Cellar/sourcekitten/0.4.2/bin does not exist
```

Could this failure be unrelated to this formula?

jpsim:
Nevermind, the problem is with this formula. I found the culprit and will be updating the PR with a new tag including the fix.

mistydemeo:
Thanks!

PR #39874, opened 2015-05-18
odt2txt: fix missing prefix

### LINKED ISSUE #39871: `brew install odt2txt` doesn't seem to install an executable
I ran `brew install odt2txt` this morning. It installs to `/usr/local/Cellar/odt2txt/0.5` but the contents of that dir are just the README and INSTALL_RECEIPT. I can't seem to find an executable anywhere, which is odd! I'm running Yosemite 10.10.3, with an updated brew which usually performs flawlessly, so I'm not inclined to think it's a local problem.

For reference, the final output of `brew install` is `🍺  /usr/local/Cellar/odt2txt/0.5: 2 files, 8.0K`, which would tally with just installing the two files referenced above.

DomT4:
The bottle is empty and the from-source installs directly into `/usr/local/bin`. This fixes that.

Closes #39871

### DIFF
--- a/Library/Formula/odt2txt.rb
+++ b/Library/Formula/odt2txt.rb
@@ -1,9 +1,12 @@
-require "formula"
-
 class Odt2txt < Formula
   homepage "https://github.com/dstosberg/odt2txt/"
   url "https://github.com/dstosberg/odt2txt/archive/v0.5.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Upside: The install check we added to the bot a couple of months ago would have caught this and prevented it from happening. Sadly the addition of this formula predated the check-bottle-has-contents install check.

PR #39878, opened 2015-05-18
sslh 1.17

dunn:


Library/Formula/sslh.rb
```diff
@@ -1,17 +1,27 @@
...
     ENV.j1
-    system 'make'
-    bin.install 'sslh-fork', 'sslh-select'
-    bin.install_symlink 'sslh-fork' => 'sslh'
-    man8.install 'sslh.8.gz'
+    system "make", "install", "PREFIX=#{prefix}"
+  end
+
+  test do
+    system sbin/"sslh", "-V"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

dunn: From a quick scan through the docs it wasn't clear how to make a better test; if I get time I'll look more closely.

mikemcquaid: Thanks! Will pull as-is for now.

PR #39891, opened 2015-05-19
fixed savana url

rene-schulz:
replaced savana URL following the instructions for the codehaus deprecation as listed here: http://www.codehaus.org/. Distributions are moved to  bintray.com


mikemcquaid:

For future reference the preferred commit message format is `savana: fix url.`.

PR #39892, opened 2015-05-19
ponyc:  minor fixes

pthariensflame:
Removes a now non-existent option for the llvm formula, and ensures
that `ponyc` builds in a release configuration.


xu-cheng:

PR #39893, opened 2015-05-19
simple-tiles 0.6.0

dunn:


Library/Formula/simple-tiles.rb
```diff
@@ -1,19 +1,37 @@
...
+    prefix.install "build/test"
+  end
+
+  test do
+    cp prefix/"test/api", testpath
+    resources.each do |r|
+      (testpath/"data").install r
+    end
+    system testpath/"api"
+    File.exist? "out.png"
```
xu-cheng: Any chance to create a test without external resources?

mikemcquaid: :+1:

DomT4: Any update on this @dunn?

dunn: I haven't had time to figure out a self-contained test, sorry!

DomT4: Would the example under `Overview` on [the website](https://propublica.github.io/simple-tiles/) work?

dunn: Wouldn't it still need one of the sample data packages for this line `simplet_vector_layer_t *layer = simplet_map_add_vector_layer(map, "PG:dbname=simple_tiles_roads");`?

DomT4: Ah, yes, looks like it :cry:

xu-cheng: We only need to link to the library. So this should be enough:

```c
#include <simple-tiles/simple_tiles.h>

int main(){
  simplet_map_t *map = simplet_map_new();
  simplet_map_free(map);
  return 0;
}
```

dunn: Thanks @xu-cheng

Library/Formula/simple-tiles.rb
```diff
@@ -14,6 +12,27 @@ class SimpleTiles < Formula
 
   def install
     system "./configure", "--prefix=#{prefix}"
-    system "make install"
+    system "make", "install"
+    prefix.install "build/test"
```
DomT4: Would this be better placed in `share` or `libexec`?

dunn: It's not necessary anymore since I fixed the test.  Maybe just remove it?

DomT4: :+1:

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/f4e9ebb71785327f9c48dc5655173fdb701bf71e. Thanks @dunn! :cocktail:

PR #39899, opened 2015-05-19
non_apple_gcc_version: support gcc5

xu-cheng:


mikemcquaid:
:+1:

xu-cheng:
:ship:ed. Also #36926 requires to extend the timeout. How to do that?

PR #35279, opened 2014-12-27
Sfml formula: sfml 2.1 -> 2.2

eleweek:
Resolves #35277 

Basically: unlike 2.1, sfml 2.2 builds just fine on yosemite, so I created this pull-request. Thanks to @DomT4 for the help with this pull-request!

PR #39900, opened 2015-05-19
sfml/csfml 2.3

dunn:
The last update was a bit tricky (#35279) so let me know if there's anything I overlooked here.

### DIFF
--- a/Library/Formula/sfml.rb
+++ b/Library/Formula/sfml.rb
@@ -1,18 +1,11 @@
 class Sfml < Formula
+  # Don't update SFML until there's a corresponding CSFML release
   homepage "http://www.sfml-dev.org/"
-  revision 1
+  desc "Cross-platform multimedia library"
+  url "http://www.sfml-dev.org/files/SFML-2.3-sources.zip"
+  sha256 "a1dc8b00958000628c5394bc8438ba1aa5971fbeeef91a2cf3fa3fff443de7c1"
 
-  stable do
-    url "http://www.sfml-dev.org/download/sfml/2.2/SFML-2.2-sources.zip"
-    sha1 "b21721a3dc221a790e4b81d6ba358c16cb1c1cd3"
-
-    # Patch to support removing the extlibs dir which caused install_name_tool complications.
-    # Already merged in HEAD and will be in the next release.
-    patch do
-      url "https://github.com/LaurentGomila/SFML/commit/9f2aecf9cff75307.diff"
-      sha1 "87646d34cf0315c814ca6313cf75bb0439113eef"
-    end
-  end
+  head "https://github.com/SFML/SFML.git"
 
   bottle do
     cellar :any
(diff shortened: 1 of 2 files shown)


DomT4:
I knew I'd forgotten something. They've made major changes to the deps, see https://github.com/SFML/SFML/pull/852#issuecomment-99855970.

dunn:
From that thread:
> Also on OS X we don't use the default OpenAL shipped by Apple.
Does that mean `openal-soft` for us?

DomT4:
Looking at what upstream do for other platforms, yup.

mikemcquaid:
`csfml` failing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/26400/version=yosemite/testReport/junit/brew-test-bot/yosemite/install_csfml/

dunn:
They're recommending we wait for 2.3 of CSFML

mantognini:
@dunn Thanks for updating those! :-)

Library/Formula/sfml.rb
```diff
@@ -21,26 +12,43 @@ class Sfml < Formula
...
-    # https://github.com/Homebrew/homebrew/pull/35279
-    rm_rf buildpath/"extlibs"
+    # Always remove the "extlibs" to avoid install_name_tool failure
+    # (https://github.com/Homebrew/homebrew/pull/35279) but leave the
+    # headers that were moved there in https://github.com/SFML/SFML/pull/795
+    Dir.glob("extlibs/*") do |dir|
+      if dir != "extlibs/headers"
+        rm_rf dir
+      end
+    end
```
jacknagel: `rm_rf Dir["extlibs/*"] - ["extlibs/headers"]`?

dunn: thanks!

dunn:
Is an update warning at the top of each formula appropriate?

mikemcquaid:
> Is an update warning at the top of each formula appropriate?

:+1:

dunn:
Linking CSFML fails because it has some of the same headers SFML has:
```
Could not symlink include/SFML/Audio/AlResource.hpp
Target /usr/local/include/SFML/Audio/AlResource.hpp
is a symlink belonging to sfml. You can unlink it:
  brew unlink sfml

  brew link --overwrite csfml

  brew link --overwrite --dry-run csfml

/usr/local/include/SFML/Audio/AlResource.hpp -> /usr/local/Cellar/sfml/2.3/include/SFML/Audio/AlResource.hpp
```

Would it be safe to just remove CSFML's `include/SFML` directory, @mantognini?

mantognini:
That's a strange error you got there... there is no `AlResource.hpp` in CSFML as far as I know.

DomT4:
> Linking CSFML fails because it has some of the same headers SFML has:

You seem to be using the same tarball for the two formulae? Were the two tarballs merged?

dunn:
ahahahaha, my bad.

DomT4:
Is this ready to ship? Not waiting on anything else now?

dunn:
@mantognini, should we wait until the CSFML 2.3 source is available from http://www.sfml-dev.org/download/csfml/ ?

mantognini:
For CSFML you can safely use GitHub zip archive, yes.

For SFML, we remove some files from it so it's safer to use the one from sfml-dev.org (currently we only remove .gitattributes but it used to be more than that -- don't know about the future though ;-))

DomT4:
:ship:'d in https://github.com/Homebrew/homebrew/commit/e2de0389ed51aef47f8f67c9e67c4db79dde88e2. Thanks @dunn @mantognini!

I removed the `desc` lines manually before merge as Mike requested we don't add any to existing formulae for now. Side note: Separate commits per formula (Inside the same PR) in this situation is preferable.

dunn:
Good to know, thanks.

PR #39906, opened 2015-05-19
uses/deps: bugs fix

xu-cheng:
* uses: make skip arguments work for requirement
* uses/deps: fix for recursive deps/reqs


mikemcquaid:
:+1:

ilovezfs:
It appears that this caused the regression described in https://github.com/Homebrew/homebrew/issues/50066

PR #39922, opened 2015-05-20
mergepbx 0.9 (new formula)

theoriginalgri:
script for merging XCode project files in git

### DIFF
--- a/Library/Formula/mergepbx.rb
+++ b/Library/Formula/mergepbx.rb
@@ -1,0 +1,36 @@
+class Mergepbx < Formula
+  homepage "https://github.com/simonwagner/mergepbx"
+  url "https://github.com/simonwagner/mergepbx/archive/v0.9.tar.gz"
+  sha256 "d0d859a902357f8059fca0c5fb061a9072b82cb7cce4f38bd9eb2c3198fce003"
+
+  resource "dummy_base" do
+    url "https://raw.githubusercontent.com/simonwagner/mergepbx/a9bd9d8f4a732eff989ea03fbc0d78f6f6fb594f/test/fixtures/merge/dummy/dummy1/project.pbxproj.base"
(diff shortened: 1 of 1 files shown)


Library/Formula/mergepbx.rb
```diff
@@ -0,0 +1,11 @@
...
+  homepage "https://github.com/simonwagner/mergepbx"
+  url "https://github.com/simonwagner/mergepbx/archive/v0.9.tar.gz"
+  version "0.9"
+  sha256 "d0d859a902357f8059fca0c5fb061a9072b82cb7cce4f38bd9eb2c3198fce003"
+
+  def install
+    system "./build.py"
+    bin.install "mergepbx"
+  end  
+end
```
mikemcquaid: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/mergepbx.rb
```diff
@@ -0,0 +1,11 @@
+class Mergepbx < Formula
+  homepage "https://github.com/simonwagner/mergepbx"
+  url "https://github.com/simonwagner/mergepbx/archive/v0.9.tar.gz"
+  version "0.9"
```
mikemcquaid: Pretty sure this isn't needed.

theoriginalgri:
I added the test step as you wished. The command has no version output, nor can I create a xcode project on the fly, so the only chance was the help output.

Your were right, the version was not needed and is auto-determined.

mikemcquaid:
> nor can I create a xcode project on the fly, so the only chance was the help output.

It'd be good if we could embed some sample data to write out and/or download it as a resource.

theoriginalgri:
I hope it looks better now :)

mikemcquaid:
Thats great, thanks!

mikemcquaid:

PR #37454, opened 2015-03-07
utils: support curl env

DomT4:
Self-explanatory request. Have wanted to do this for a while, but I don’t know how the maintainers feel. The general feeling towards screwing with the download mechanism too much has seemed a little reluctant, for understandable reasons, but limiting it to using a cURL means nothing else would need to change, and would be fairly predictable.

Open to outright rejection, but I live in hope.

PR #39927, opened 2015-05-20
utils: prefer keg curl over system on 10.6

rhendric:
Adds support for using a different curl binary than /usr/bin/curl
via a HOMEBREW_CURL_PATH environment variable.

Pretty self-explanatory, I hope. Discovered I needed this when trying to brew install anything sourced from https://download.gnome.org/; the version of curl shipped with 10.6 (I know, I know) doesn't play well with servers running newer versions of OpenSSL (see http://sourceforge.net/p/curl/bugs/1037/).

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -175,7 +175,12 @@ def quiet_system cmd, *args
 end
 
 def curl *args
-  curl = Pathname.new '/usr/bin/curl'
+  brewed_curl = HOMEBREW_PREFIX/"opt/curl/bin/curl"
+  curl = if MacOS.version <= "10.6" && brewed_curl.exist?
+    brewed_curl
+  else
+    Pathname.new '/usr/bin/curl'
+  end
   raise "#{curl} is not executable" unless curl.exist? and curl.executable?
 
   flags = HOMEBREW_CURL_ARGS


dunn:
Some related issues: #37454, #33235

rhendric:
Ah, poop. Maybe the maintainers' position on this would change knowing that the above is an issue, the only workaround for which would be to change /usr/bin/curl on my system?

Affected formulae are:
atk.rb
cogl.rb
gexiv2.rb
glib.rb
glibmm.rb
gnome-icon-theme.rb
goffice.rb
gsettings-desktop-schemas.rb
gtk+.rb
gtk+3.rb
gtk-doc.rb
gtkmm.rb
gtkmm3.rb
gupnp.rb
json-glib.rb
libgsf.rb
librsvg.rb
libxml++.rb

DomT4:
Yeah, I've been shot down on this before, and I'm still in favour of the ENV option but I don't expect to see it merged any time soon - The maintainers concerns about the ENV are valid, and we recently discovered another problem we could have hit.

When OS X 10.10.3 landed, Apple removed a few major (but old) certificates which caused everything linked into Homebrew's OpenSSL to fail to connect to all & any Google domains, as well as certain smaller domains.

This impacted `cURL` (optional for >10.8), `wget`, `git` (optional) and `mutt` amongst other tools. The problem still impacts LibreSSL, but we applied a patch for OpenSSL. If a Homebrewed `cURL` was built against OpenSSL or LibreSSL, and the `ENV` was set it would have suddenly started failing to download certain formulae without much clue as to why. Same problem for Homebrew's `git` built against Homebrew's OpenSSL, `git clone`s from Google servers suddenly started failing without clear reason.

I'm still in favour of the `ENV` here, but if we do roll with it I'd vote strongly for adding printing the active `HOMEBREW_CURL_PATH` to `brew config`.

jacknagel:
If we do anything, it should be scoped to 10.6, I think.

mistydemeo:
What Tigerbrew does is allow a Homebrew-installed curl to substitute for the system curl, if present. The Homebrew-installed curl is built with openssl and newer certs than the OS's, so it doesn't have these same issues.

mikemcquaid:
> What Tigerbrew does is allow a Homebrew-installed curl to substitute for the system curl, if present. The Homebrew-installed curl is built with openssl and newer certs than the OS's, so it doesn't have these same issues.

Yeh, this seems like a good idea for 10.6-only too.

rhendric:
If supportability is the argument against making this configurable, is it really better to make Homebrew automatically use the keg-only curl if present? Even scoped to just 10.6, doesn't that exacerbate the ‘suddenly failing without clear reason’ issue @DomT4 describes? At least with an `ENV`, the user knows that they're tinkering.

And another advantage of using `ENV` instead of restricting to what's in the Cellar is that if for some reason *both* the system curl and the last curl installed by Homebrew aren't working, user can still download a new curl manually and get Homebrew working again.

I quite like the idea of adding `HOMEBREW_CURL_PATH` to `brew config`, if set. We're talking about edge cases on edge cases now: user notices problem, figures out that the cause of the problem is curl, decides to set `HOMEBREW_CURL_PATH`, something *else* goes wrong later with the curl that the user selected, and the user fails to realize *this* time that it's their `HOMEBREW_CURL_PATH`? Isn't a reminder in `brew config` the appropriate level of safeguarding in this scenario?

mikemcquaid:
I think we're better to just use the keg-only curl and on 10.6 and below to reduce the chance of errors with weirdly compiled `curl`s.

rhendric:
I defer to your wisdom. The new commit does what Tigerbrew does, plus a version check, and minus the check for curl-ca-bundle.

Library/Homebrew/utils.rb
```diff
@@ -175,7 +175,17 @@ def quiet_system cmd, *args
...
+    Formula["curl"]
+  rescue FormulaUnavailableError
+    nil
+  end
+  use_curl = MacOS.version <= "10.6" && !brewed_curl.nil? && brewed_curl.installed?
+  if use_curl
+    curl = brewed_curl.opt_bin/"curl"
+  else
+    curl = Pathname.new '/usr/bin/curl'
+  end
```
jacknagel: All of the code related to homebrew's curl should be inside the `MacOS.version <= "10.6"` conditional, as there is no reason for it to ever execute otherwise.

When would we expect `Formula["curl"]` to raise an exception? The formula will always be present.

jacknagel: In fact, I think I'd rather see us just check the path (`HOMEBREW_PREFIX/"opt/curl/bin/curl"`) rather than do the formula dance here.

rhendric: The tests failed without the rescue; apparently some of them run in a context in which there is no curl formula. But sure, I can switch to a path test.

mikemcquaid:

PR #37370, opened 2015-03-03
dcmtk 3.6.1 update

kkumler:
Added stable / devel targets, building against homebrew openssl, homebrew libiconv.
Is it desirable to include a way to select system/brewed libraries (at least for libiconv) for build?

Earlier PR and discussion: Homebrew/homebrew-versions#634

PR #39932, opened 2015-05-20
Use brewed openssl, optional libiconv, and head version.

kkumler:
Bringing over cleaner changes (and closing previous PR #37370).
Since the snapshot is used in master now I left that as-is (instead of devel).

### DIFF
--- a/Library/Formula/dcmtk.rb
+++ b/Library/Formula/dcmtk.rb
@@ -1,23 +1,23 @@
 class Dcmtk < Formula
   homepage "http://dicom.offis.de/dcmtk.php.en"
+
+  # Current snapshot used for stable now.
   url "http://dicom.offis.de/download/dcmtk/snapshot/dcmtk-3.6.1_20150217.tar.gz"
   sha256 "3cf8f3e52ed8a5240a7facc3a118de411aa54bc9beccba0cf7a975735da35304"
   version "3.6.1-20150217"
 
-  bottle do
-    sha256 "0bd582c2f37ce4b9366b0a7ba2e7c6e90cddd3a6af954ebc5475938048415d5d" => :yosemite
-    sha256 "ff49c45d86930592662caf947cd87f011b55f533396b75957e55cd748bc4c7b2" => :mavericks
-    sha256 "2e06bdf73ce156ba65c95374ac5a9e6b048c02186403b66b78ea48b2b415da01" => :mountain_lion
-  end
+  head "git://git.dcmtk.org/dcmtk.git"
 
   option "with-docs", "Install development libraries/headers and HTML docs"
   option "with-openssl", "Configure DCMTK with support for OpenSSL"
+  option "with-libiconv", "Build with brewed libiconv. Dcmtk and system libiconv can have problems with utf8."
 
   depends_on "cmake" => :build
   depends_on "doxygen" => :build if build.with? "docs"
   depends_on "libpng"
   depends_on "libtiff"
   depends_on "openssl"
+  depends_on "homebrew/dupes/libiconv" => :optional
 
   def install
     ENV.m64 if MacOS.prefer_64_bit?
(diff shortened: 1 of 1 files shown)


Library/Formula/dcmtk.rb
```diff
@@ -12,19 +17,22 @@ class Dcmtk < Formula
...
   depends_on "openssl"
+  depends_on "homebrew/dupes/libiconv" => :optional
 
   def install
     ENV.m64 if MacOS.prefer_64_bit?
 
     args = std_cmake_args
     args << "-DDCMTK_WITH_DOXYGEN=YES" if build.with? "docs"
-    args << "-DDCMTK_WITH_OPENSSL=YES" if build.with? "openssl"
+    args << "-DDCMTK_WITH_OPENSSL=YES -DWITH_OPENSSLINC=#{Formula["openssl"].opt_prefix}" if build.with? "openssl"
```
DomT4: It wasn't using our OpenSSL before?

Library/Formula/dcmtk.rb
```diff
@@ -37,6 +45,6 @@ def install
   test do
     system bin/"pdf2dcm", "--verbose",
            test_fixtures("test.pdf"), testpath/"out.dcm"
-    File.exist? testpath/"out.dcm"
+    system bin/"dcmftest", testpath/"out.dcm"
```
DomT4: Why this change?

Library/Formula/dcmtk.rb
```diff
@@ -1,8 +1,13 @@
 class Dcmtk < Formula
   homepage "http://dicom.offis.de/dcmtk.php.en"
+
+  # Current snapshot used for stable now.
   url "http://dicom.offis.de/download/dcmtk/snapshot/dcmtk-3.6.1_20150217.tar.gz"
   sha256 "3cf8f3e52ed8a5240a7facc3a118de411aa54bc9beccba0cf7a975735da35304"
   version "3.6.1-20150217"
+  revision 2
```
DomT4: Why the revision? Only necessary if it wasn't actually using our OpenSSL before. Should be `revision 1` if so.

Library/Formula/dcmtk.rb
kkumler: It would build against the system OpenSSL before, even with it depending on the brewed OpenSSL.

Library/Formula/dcmtk.rb
kkumler: It gave a higher certainty that the output file was in a valid format instead of just existing.

Library/Formula/dcmtk.rb
DomT4: I can't reproduce this, either from bottle or from source.

The source install actively references the correct OpenSSL with the current formula:

```
[ 44%] Building CXX object dcmnet/libsrc/CMakeFiles/dcmnet.dir/dimdump.cc.o
cd /tmp/dcmtk20150521-5874-18qn225/dcmtk-3.6.1_20150217/build/dcmnet/libsrc && /usr/local/Library/ENV/4.3/clang++   -DDCMTK_BUILD_IN_PROGRESS -DUSE_NULL_SAFE_OFSTRING -D_BSD_COMPAT -D_BSD_SOURCE -D_OSF_SOURCE -D_REENTRANT -D_XOPEN_SOURCE_EXTENDED -Ddcmnet_EXPORTS -D_XOPEN_SOURCE_EXTENDED -D_BSD_SOURCE -D_BSD_COMPAT -D_OSF_SOURCE -D_DARWIN_C_SOURCE -I/usr/local/include -I/usr/local/opt/openssl/include -I/usr/include/libxml2
```

What's happening for you locally?

Library/Formula/dcmtk.rb
DomT4: Fair enough.

Library/Formula/dcmtk.rb
DomT4: `otool` lines up as expected as well from-source here.

```
~> otool -L /usr/local/Cellar/dcmtk/3.6.1-20150217/bin/* | grep ssl
	/usr/local/opt/openssl/lib/libssl.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
```

kkumler: My bad. I had that in for the slightly older version/PR but it does look good with this version.

mikemcquaid:

PR #39933, opened 2015-05-20
nghttp2 1.0.0

vszakats:
Words of caution:
https://nghttp2.org/blog/2015/05/16/nghttp2-v1-0-0/
https://www.youtube.com/watch?v=VN4rVnf2cR8

The API changed in incompatible ways with the 0.7.16 to 1.0.0 bump. Its only dependency inside Homebrew, `curl` will break with its current release when built against optional `nghttp2`. It will be made compatible with its next scheduled release 7.43.0, when it will require 1.0.0.

### DIFF
--- a/Library/Formula/nghttp2.rb
+++ b/Library/Formula/nghttp2.rb
@@ -1,7 +1,7 @@
 class Nghttp2 < Formula
   homepage "https://nghttp2.org"
-  url "https://github.com/tatsuhiro-t/nghttp2/releases/download/v0.7.13/nghttp2-0.7.13.tar.xz"
-  sha256 "91f2bfcad1e27472d8c96de71d9bcb37d93b131ee60c775c95b46be82a24e1db"
+  url "https://github.com/tatsuhiro-t/nghttp2/releases/download/v1.0.0/nghttp2-1.0.0.tar.xz"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Thanks @vszakats!

DomT4:
> Its only dependency inside Homebrew, curl will break with its current release when built against optional nghttp2. It will be made compatible with its next scheduled release 7.43.0, when it will require 1.0.0.

Just stumbled into this:

```
dyld: Library not loaded: /usr/local/lib/libnghttp2.5.dylib
  Referenced from: /usr/local/opt/curl/lib/libcurl.4.dylib
  Reason: image not found
Error: Failure while executing: git clone
```

Any timetable on the next upstream release? Leaving a broken option in formula for a prolonged period is asking for trouble.

evilantnie:
This has broken curl for me, if we don't have a timeline on 7.43 can we temporarily get a formula in homebrew/versions for nghttp2 with the previous 0.7.16 version? or is this a problem in the curl forumla?

mikemcquaid:
I shouldn't have merged this. Reverted in 8f0ca3f.

vszakats:
Next release is June 17, 2015.
From this page: http://curl.haxx.se/dev/release-procedure.html

PR #39934, opened 2015-05-20
Tika: add US and EU mirrors

acdha:
Tika currently cannot be installed due to the way the Apache mirrors work (see #39935) but this avoids the problem by specifying the official backup URLs as mirrors


DomT4:
If we're going to add mirrors, it should be the `https://archive.apache.org/dist/` mirror.

jacknagel:
>If we're going to add mirrors, it should be the https://archive.apache.org/dist/ mirror.

And we already add this implicitly [here](https://github.com/Homebrew/homebrew/blob/ae8827439e7664d4cad7d66e5c78ea86da21b777/Library/Homebrew/download_strategy.rb#L354).

Closing until the issues the OP is encountering are better explained in #39935.

PR #39938, opened 2015-05-20
add formula for Pillar

digital-wonderland:
Hi,

I was missing a formula for [Pillar](https://github.com/comeara/pillar), so I wrote one.

Since it is my first, I might have missed something obvious, please point it out so it can be fixed. 

Noteworthy might be that the ```test``` section is missing. This is because those tests would require a running [Cassandra](http://cassandra.apache.org/) instance (Pillar is a tool to handle migration of Cassandra schemes) - also those test normally get run before / when creating a new release, so the code should be ok.

### DIFF
--- a/Library/Formula/pillar.rb
+++ b/Library/Formula/pillar.rb
@@ -1,0 +1,26 @@
+class Pillar < Formula
+  homepage "https://github.com/comeara/pillar"
+  url "https://github.com/comeara/pillar/archive/v2.0.1.tar.gz"
+  sha256 "d5db7fbf956dae05794ec7b26c1de70aaab1d21fed774d95d16a241e9dd3952e"
+
+  depends_on :java
+  depends_on "sbt" => :build
+
+  def install
+    system "sbt", "assembly"
+
+    inreplace "src/main/bash/pillar" do |s|
+      s.gsub! "/usr/java/default", "`/usr/libexec/java_home`"
+      s.gsub! "${PILLAR_ROOT}/lib/pillar.jar", "#{libexec}/pillar-assembly-#{version}.jar"
+      s.gsub! "${PILLAR_ROOT}/conf", "#{etc}/pillar-log4j.properties"
+    end
+
+    bin.install "src/main/bash/pillar"
+    etc.install "src/main/resources/pillar-log4j.properties"
+    libexec.install "target/scala-2.10/pillar-assembly-#{version}.jar"
+  end
+
+  test do
+    assert_match /Missing parameter/, shell_output("#{bin}/pillar 2>&1", 1)
+  end
+end


Library/Formula/pillar.rb
```diff
@@ -0,0 +1,25 @@
+class Pillar < Formula
+  homepage "https://github.com/comeara/pillar"
+  url "https://github.com/comeara/pillar/archive/v2.0.1.tar.gz"
+  version "2.0.1"
```
DomT4: This'll be redundant. `brew` can read versions automatically providing the URL is relatively simple, which it is here.

Library/Formula/pillar.rb
```diff
@@ -0,0 +1,25 @@
...
+  sha256 "d5db7fbf956dae05794ec7b26c1de70aaab1d21fed774d95d16a241e9dd3952e"
+
+  depends_on "sbt" => :build
+
+  def install
+
```
DomT4: :scissors: the blank line

Library/Formula/pillar.rb
```diff
@@ -0,0 +1,25 @@
...
+	s.gsub! "/opt/pillar", "#{prefix}"
+	s.gsub! "${PILLAR_ROOT}/lib/pillar.jar", "#{libexec}/pillar-assembly-#{version}.jar"
+	s.gsub! "${PILLAR_ROOT}/conf", "#{etc}/pillar-log4j.properties"
+    end
+
+    bin.install "src/main/bash/pillar"
+    etc.install "src/main/resources/pillar-log4j.properties"
+    libexec.install "target/scala-2.10/pillar-assembly-#{version}.jar"
+  end
+
```
DomT4: :scissors: the blank line

Library/Formula/pillar.rb
digital-wonderland: should be fixed now

Library/Formula/pillar.rb
digital-wonderland: version line is removed

Library/Formula/pillar.rb
digital-wonderland: that one is removed as well :)

DomT4: Thanks for the quick nit cleanups! :+1:

digital-wonderland: thanks a lot for the prompt feedback ;)

DomT4:
Some audit comments:

```
== /usr/local/Library/Formula/pillar.rb ==
C: 12:  2: Use 2 (not -3) spaces for indentation.

Error: 7 problems in 1 formulae

==> audit problems
pillar:
 * Use `depends_on :java` to set JAVA_HOME
 * Use spaces instead of tabs for indentation
 * Don't need to interpolate "prefix" with gsub!
 * Use spaces instead of tabs for indentation
 * Use spaces instead of tabs for indentation
 * Use spaces instead of tabs for indentation
```

You can ignore the `test do` line I expect given what you said in your original post.

mikemcquaid:
> Noteworthy might be that the test section is missing. This is because those tests would require a running Cassandra instance (Pillar is a tool to handle migration of Cassandra schemes) - also those test normally get run before / when creating a new release, so the code should be ok.

Is there any way to get even `--help` or `--version` output?

digital-wonderland:
@DomT4 Thanks for the comments, I fixed the indentation and removed the prefix interpolation but I'm not sure how you want me to fix the ```JAVA_HOME``` part since adding a ```depends_on :java``` line doesn't seem to set ```JAVA_HOME```. Also Java should already be installed in some way or the other since ```sbt``` requires it. Could you please clarify that part?

@mikemcquaid Unfortunately not that I know of. If you run it with missing required or invalid arguments you get an 'how to use it' help output.

DomT4:
> but I'm not sure how you want me to fix the JAVA_HOME part since adding a depends_on :java line doesn't seem to set JAVA_HOME.

I think this might actually be a red-herring, since we're not actually using the variable. @xu-cheng wrote the implementation though. I guess there's no harm in adding `depends_on :java` either way since you do require Java for the tool.

> @mikemcquaid Unfortunately not that I know of. If you run it with missing required or invalid arguments you get an 'how to use it' help output.

We can work with this. Something like:

```ruby
test do
  assert_match /some help output/, shell_output("#{bin}/pillar")
end
```

digital-wonderland:
I added a ```depends_on :java``` line and the suggested test. The installation works but ```brew test pillar``` fails with
```
Error: pillar: failed
<0> expected but was
<1>.
```
since the return value is ```1``` (which is correct since it is reporting an incorrect usage). How can I tell it to expect ```1``` instead of ```0```?

DomT4:
Try `shell_output("#{bin}/pillar", 1)`. If that doesn't work, try `pipe_output("#{bin}/pillar 2>&1")`.

digital-wonderland:
Thanks, with ```shell_output("#{bin}/pillar 2>&1", 1)``` it seems to be working properly :)

Anything else that should be fixed?

Library/Formula/pillar.rb
```diff
@@ -0,0 +1,26 @@
...
+      s.gsub! "${PILLAR_ROOT}/conf", "#{etc}/pillar-log4j.properties"
+    end
+
+    bin.install "src/main/bash/pillar"
+    etc.install "src/main/resources/pillar-log4j.properties"
+    libexec.install "target/scala-2.10/pillar-assembly-#{version}.jar"
+  end
+
+  test do
+    assert_match /Missing parameter/, shell_output("#{bin}/pillar 2>&1", 1)
```
mikemcquaid: Could we write some sample data and then `pillar initialize myapp`?

digital-wonderland: As said above, that would require a running Cassandra instance. Also you would test only

1. that the application is properly installed & executable - this is already ensured by the current test
2. that running ```pillar initialize myapp``` has a return value of ```0``` - otherwise you would also have to check if keyspaces within Cassandra got manipulated.

Pillar already contains a test suite, which should get run during release creation, which actually verifies it is working properly. Therefore such a test wouldn't - IMHO - really add additional value while forcing a local Cassandra installation on everyone (which is seldom wanted) and making the test much more compilcated (install cassandra, configure a standalone instance, start it, wait until it is ready, ...).

mikemcquaid: Ok, cool, if that requires a running server this is fine.

mikemcquaid:

mikemcquaid:
For future reference the preferred commit message format for simple version updates is `pillar 2.0.1 (new formula)`.

PR #39948, opened 2015-05-21
pathname: Ensure symlink paths are based on real paths

elyscape:
Previously, if a symlink was to be installed into a folder that was
itself under a symlink, the calculated relative path would result in an
invalid link. This is prevented by converting the paths used to the
actual on-disk paths first.

### DIFF
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -70,10 +70,10 @@ class Pathname
   end
 
   def install_symlink_p(src, new_basename)
-    src = Pathname(src).expand_path(self)
+    src = Pathname(src).realpath
     dst = join(new_basename)
     mkpath
-    FileUtils.ln_sf(src.relative_path_from(dst.parent), dst)
+    FileUtils.ln_sf(src.relative_path_from(dst.parent.realpath), dst)
   end
   private :install_symlink_p
 


elyscape:
The build works now.

mikemcquaid:
Would like @jacknagel to :+1: this before we include it.

xu-cheng:
Will this break symlink-chain?

jacknagel:
I don't think this is correct.

The use of `expand_path(self)` means "if passed a relative path, expand it relative to self", and `realpath` does not have that context.

For example, `lib.install_symlink "foo" => "bar" will result in:

```
lib
 foo
 bar -> foo
```

will this produce the same result?

In any case, this is a behavior change, so I think it needs to be fixed in a different way (it would help to have a test case demonstrating the bug first).

jacknagel:
For example, this test passes under the current implementation, but fails on the first call to `realpath` in this PR:

```diff
diff --git a/Library/Homebrew/test/test_pathname.rb b/Library/Homebrew/test/test_pathname.rb
index caf650b..b6e250f 100644
--- a/Library/Homebrew/test/test_pathname.rb
+++ b/Library/Homebrew/test/test_pathname.rb
@@ -218,4 +218,10 @@ def test_install_symlink
     assert_predicate @dst+"bin/b.txt", :exist?
     assert_predicate (@dst+"bin").readlink, :relative?
   end
+
+  def test_install_symlink_relative
+    (@dst+"foo").write "a"
+    @dst.install_symlink "foo" => "bar"
+    assert_equal Pathname.new("foo"), (@dst+"bar").readlink
+  end
 end
```

elyscape:
The problem this is intended to solve is most easily demonstrated by doing an install of [bash-completion2](https://github.com/Homebrew/homebrew-versions/blob/master/bash-completion2.rb) and then examining the created symlink `/usr/local/share/bash-completion/completions/brew`:
```shellsession
Fruitless:~ Elyscape$ brew install bash-completion2
==> Installing bash-completion2 from homebrew/homebrew-versions
==> Downloading https://homebrew.bintray.com/bottles-versions/bash-completion2-2
Already downloaded: /Library/Caches/Homebrew/bash-completion2-2.1_2.yosemite.bottle.tar.gz
==> Pouring bash-completion2-2.1_2.yosemite.bottle.tar.gz
==> Caveats
Add the following to your ~/.bash_profile:
  if [ -f $(brew --prefix)/share/bash-completion/bash_completion ]; then
    . $(brew --prefix)/share/bash-completion/bash_completion
  fi

  Homebrew's own bash completion script has been linked into
    /usr/local/share/bash-completion/completions
  bash-completion will automatically source it when you invoke `brew`.

  Any completion scripts in /usr/local/etc/bash_completion.d
  will continue to be sourced as well.
==> Summary
🍺  /usr/local/Cellar/bash-completion2/2.1_2: 387 files, 2.6M
Fruitless:~ Elyscape$ cd /usr/local/share/bash-completion/completions/
Fruitless:completions Elyscape$ ls -l brew
lrwxr-xr-x  1 Elyscape  admin  54 May 21 20:29 brew@ -> ../../../Library/Contributions/brew_bash_completion.sh
Fruitless:completions Elyscape$ cat brew
cat: brew: No such file or directory
```
This is because the symlink isn't actually located in `/usr/local/share/bash-completions/completions/`:
```shellsession
Fruitless:completions Elyscape$ pwd -P
/usr/local/Cellar/bash-completion2/2.1_2/share/bash-completion/completions
```
While `/usr/local/share/bash-completion/completions/../../../Library/Contributions/brew_bash_completion.sh` appears at first blush to expand to `/usr/local/Library/Contributions/brew_bash_completion.sh`, which is what we want, `/usr/local/share/bash-completion` is a symlink:
```shellsession
Fruitless:share Elyscape$ ls -l bash-completion
lrwxr-xr-x  1 Elyscape  admin  54 May 21 20:29 bash-completion@ -> ../Cellar/bash-completion2/2.1_2/share/bash-completion
```
As a result, the `brew` symlink is actually evaluated as `/usr/local/Cellar/bash-completion2/2.1_2/share/bash-completion/completions/../../../Library/Contributions/brew_bash_completion.sh` and therefore expands to `/usr/local/Cellar/bash-completion2/2.1_2/Library/Contributions/brew_bash_completion.sh`, which does not exist.  By contrast, when bash-completion2 is installed with my patch:
```shellsession
Fruitless:completions Elyscape$ ls -l brew
lrwxr-xr-x  1 Elyscape  admin  63 May 21 20:36 brew@ -> ../../../../../../Library/Contributions/brew_bash_completion.sh
```
`/usr/local/Cellar/bash-completion2/2.1_2/share/bash-completion/completions/../../../../../../Library/Contributions/brew_bash_completion.sh` correctly expands to `/usr/local/Library/Contributions/brew_bash_completion.sh`.

jacknagel:
The patch breaks cases that work today, though, so it's not acceptable in its current form.

The above test will fail even after removing the first call to realpath:

```
jacknagel@haswell:test(master *=) $ git diff
diff --git a/Library/Homebrew/extend/pathname.rb b/Library/Homebrew/extend/pathname.rb
index ee8d8c9..0725da2 100644
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -73,7 +73,7 @@ def install_symlink_p(src, new_basename)
     src = Pathname(src).expand_path(self)
     dst = join(new_basename)
     mkpath
-    FileUtils.ln_sf(src.relative_path_from(dst.parent), dst)
+    FileUtils.ln_sf(src.relative_path_from(dst.parent.realpath), dst)
   end
   private :install_symlink_p

jacknagel@haswell:test(master *=) $ bin/rake test:pathname
Run options: --seed 29707

# Running:

........F.................

Finished in 0.034928s, 744.3885 runs/s, 1460.1466 assertions/s.

  1) Failure:
PathnameInstallTests#test_install_relative_symlink [/Users/jacknagel/src/homebrew/Library/Homebrew/test/test_pathname.rb:224]:
--- expected
+++ actual
@@ -1 +1 @@
-#<Pathname:foo>
+#<Pathname:../../../../../../../var/folders/76/fd0gwqj13570v2h7x093_p_w0000gn/T/d20150522-3294-17n4pec/foo>


26 runs, 51 assertions, 1 failures, 0 errors, 0 skips
rake aborted!
Command failed with status (1): [ruby -I"lib:/Users/jacknagel/src/homebrew/Library/Homebrew/test"  "/Users/jacknagel/.rubies/ruby-2.2.2/lib/ruby/2.2.0/rake/rake_test_loader.rb" "/Users/jacknagel/src/homebrew/Library/Homebrew/test/test_pathname.rb" ]

Tasks: TOP => test:pathname
(See full trace by running task with --trace)
```

elyscape:
I'll take another look and see if I can figure out what's going on.

DomT4:
What's the situation here?

elyscape:
Sorry, been really busy.  Hopefully I'll have a chance to take another look this weekend.

tdsmith:
Thanks for your work here; please submit a PR against Homebrew/brew if you find a chance to work on it.

PR #39956, opened 2015-05-21
libepoxy: build requires Python 2.7

rhendric:
Which ships with Lion, but Snow Leopard is still on 2.6.


mikemcquaid:

PR #39960, opened 2015-05-21
8.2

hutchkintoot:
BaseX 8.2


DomT4:
If you could bump the checksum to `sha256` instead of `sha1` that'd be awesome. Not mandatory though!

hutchkintoot:
Is it ok if i switch to sha256 next version?

DomT4:
Sure. As I said, not mandatory for existing formula, just nice to have.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `basex 8.2`.

PR #39962, opened 2015-05-21
Openmpi 1.8.4 (add java dependency)

davydden:
see https://github.com/Homebrew/linuxbrew/issues/365#issuecomment-104373158

### DIFF
--- a/Library/Formula/open-mpi.rb
+++ b/Library/Formula/open-mpi.rb
@@ -21,6 +21,7 @@ class OpenMpi < Formula
   conflicts_with "mpich2", :because => "both install mpi__ compiler wrappers"
   conflicts_with "lcdf-typetools", :because => "both install same set of binaries."
 
+  depends_on :java => :build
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Is this just needed at build-time?

davydden:
I guess it depends on point of view. As you you might have seen in the issue above, the error appeared at
```
libexec.install bin/'vtsetup.jar'
```
which is absent without Java. I know nothing about java, but I suppose this file may be used later on by some (java) applications.

mikemcquaid:
Let's make it a `:build` dependency for now, thanks.

davydden:
done.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/41e38cd9a13fb27e4e667e175acf84661db9f5e1. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:. Please note for non version bumps the commit format should be a simple `formula: change i made`, Thanks!

PR #39967, opened 2015-05-21
poppler 0.33.0

edporras:
Hi,

It's been a while since we've pushed an updated recipe for poppler so I'm unsure what the current status of things are. Every update since 0.29 has added core fixes for better handling of malformed documents and memory leaks so these would be nice to have.

I've tested by building these from source:

`brew install --build-from-source pdf2svg pdfgrep pdftoipe diff-pdf mat xournal`

All build successfully except:

1. `mat` still reports "MAT was built without PDF support nor GUI." and there's still no flag to enable this.
2. `frescobaldi` complains that `fontforge` "was built with a different C++ standard
library (libstdc++ from clang)." Building `fontforge` (and `gettext`) from source doesn't resolve this though.

I've recently updated to the latest Xcode and command-line tools. Could this be causing this?

### DIFF
--- a/Library/Formula/poppler.rb
+++ b/Library/Formula/poppler.rb
@@ -2,8 +2,8 @@ require 'formula'
 
 class Poppler < Formula
   homepage 'http://poppler.freedesktop.org'
-  url 'http://poppler.freedesktop.org/poppler-0.29.0.tar.xz'
-  sha1 'ba3330ab884e6a139ca63dd84d0c1c676f545b5e'
+  url 'http://poppler.freedesktop.org/poppler-0.33.0.tar.xz'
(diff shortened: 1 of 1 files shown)


mikemcquaid:
At least `pdf2htmlex` needs a `revision` increase.

mikemcquaid:
@tdsmith something weird going on with Python here? Seems it's maybe being picked up without a dependency.

tdsmith:
Yeah, it looks like the `mat` bottles are bad? I don't see any obvious problems in the formula.

dunn:
#37280 is still open, it's being held up by mat stuff too.

jacknagel:
Closing as a duplicate of #37280.

PR #39972, opened 2015-05-21
android-ndk r10e

jfirebaugh:
Updates the package to r10e.

The [homepage](https://developer.android.com/tools/sdk/ndk/index.html) no longer lists a 32-bit version, and I've confirmed that none exists at the expected URL. So it appears that only a 64-bit version is available. I've removed the `else` branch of the `MacOS.prefer_64_bit?` conditional but I'm not sure that that is the correct way to handle the situation.

### DIFF
--- a/Library/Formula/android-ndk.rb
+++ b/Library/Formula/android-ndk.rb
@@ -1,28 +1,19 @@
 class AndroidNdk < Formula
   homepage "https://developer.android.com/sdk/ndk/index.html"
+  url "https://dl.google.com/android/ndk/android-ndk-r10e-darwin-x86_64.bin"
+  sha256 "728c309e606f63101f1258c9d3d579b80ac74fe74c511ebb71f460ce5c5d084e"
 
-  if MacOS.prefer_64_bit?
-    url "https://dl.google.com/android/ndk/android-ndk-r10d-darwin-x86_64.bin"
-    sha256 "46e6d0249012d926996616709f9fdd2d4506309309c19e3ab7468f7ab04b0ddc"
-  else
-    url "https://dl.google.com/android/ndk/android-ndk-r10d-darwin-x86.bin"
-    sha256 "f1eb62d3a256f1339978b96b05bd3d7195debbf07b2d1e8a887d2f2b468d6cc7"
-  end
+  version "r10e"
 
-  version "r10d"
-
+  # As of r10e, only a 64-bit version is provided
+  depends_on :arch => :x86_64
   depends_on "android-sdk" => :recommended
 
   def install
     bin.mkpath
 
-    if MacOS.prefer_64_bit?
-      chmod 0755, "./android-ndk-#{version}-darwin-x86_64.bin"
-      system "./android-ndk-#{version}-darwin-x86_64.bin"
-    else
-      chmod 0755, "./android-ndk-#{version}-darwin-x86.bin"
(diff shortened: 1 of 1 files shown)


ljbade:
The release notes say they dropped 32-bit support for Mac.

DomT4:
Might need to scope it to make sure 32bit folk don't end up with it installed. OS X 10.6 was the first to ship x86_64, so I guess we could add `depends_on :macos => :snow_leopard`.

Library/Formula/android-ndk.rb
```diff
@@ -2,14 +2,11 @@ class AndroidNdk < Formula
   homepage "https://developer.android.com/sdk/ndk/index.html"
 
   if MacOS.prefer_64_bit?
```
DomT4: If there's no `x64` release you can drop all the `if` statements and just make the `x86_64` version the default `url` and `sha256` in the normal location.

Library/Formula/android-ndk.rb
```diff
@@ -19,9 +16,6 @@ def install
     if MacOS.prefer_64_bit?
```
DomT4: Same here. Drop the `if` and corresponding `end` and then leave the `x86_64` chmod and system call in place.

jfirebaugh:
@DomT4 Thanks, I've made those changes.

Library/Formula/android-ndk.rb
```diff
@@ -1,28 +1,20 @@
...
-    url "https://dl.google.com/android/ndk/android-ndk-r10d-darwin-x86.bin"
-    sha256 "f1eb62d3a256f1339978b96b05bd3d7195debbf07b2d1e8a887d2f2b468d6cc7"
-  end
-
-  version "r10d"
+  version "r10e"
 
+  # As of r10e, only a 64-bit version is provided, and OS X 10.6 was the
+  # first to ship x86_64.
+  depends_on :macos => :snow_leopard
```
mikemcquaid: `depends_on :arch => :x86_64`

Library/Formula/android-ndk.rb
```diff
@@ -1,28 +1,20 @@
 class AndroidNdk < Formula
   homepage "https://developer.android.com/sdk/ndk/index.html"
+  url "https://dl.google.com/android/ndk/android-ndk-r10e-darwin-x86_64.bin"
```
mikemcquaid: Any ideas if there's any way to build this from source?

Library/Formula/android-ndk.rb
DomT4: I completely forgot this was a thing. Sorry for the bad suggestion :smile_cat:.

Library/Formula/android-ndk.rb
jfirebaugh: Not that I know of, there's no mention of it on the homepage.

Library/Formula/android-ndk.rb
jfirebaugh: Thanks, updated.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/f18f61f0c40427b74207004bc8ff2538157689e7. The audit failure is unrelated, It links to the system OpenSSL but since its a binary and there's no apparent way to build this from source we're stuck with that. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #39974, opened 2015-05-21
riak: add advice about ulimit settings when too low

squarism:
`brew audit --strict riak` complains about openssl being linked but it was doing that before my change:

    ==> audit problems
    riak:
      /usr/local/Cellar/riak/2.0.4/libexec/lib/crypto-3.1/priv/lib/crypto.so

Also, a file from the basho tarball is already linked:

    otool -L /usr/local/Cellar/riak/2.0.4/libexec/lib/crypto-3.1/priv/lib/crypto.so
      /usr/local/Cellar/riak/2.0.4/libexec/lib/crypto-3.1/priv/lib/crypto.so:
      /usr/lib/libcrypto.0.9.8.dylib (compatibility version 0.9.8, current version 47.1.0)
      /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 169.3.0)

Outside that, I tested the new formula and it works.  There's some complexity around suggesting what to do about maxfiles and other limits on Yosemite (the process has changed), so I left post_install.  Hopefully that doesn't bit rot as fast as make a suggestion of what to do from the formula.

I tried to get a plist to work but launchctl is not happy with a forking Erlang process.  I can get it to boot with homebrew but not stop.  So, that's not in here.  :(  If anyone has advice to get this to autoboot (with `launchctl`), please reach out to me, I'm curious about this.

### DIFF
--- a/Library/Formula/riak.rb
+++ b/Library/Formula/riak.rb
@@ -1,8 +1,6 @@
-require "formula"
-
 class Riak < Formula
-  desc "Distributed database"
   homepage "http://basho.com/riak/"
+  desc "Riak is a distributed NoSQL key-value data store."
   url "https://s3.amazonaws.com/downloads.basho.com/riak/2.1/2.1.1/osx/10.8/riak-2.1.1-OSX-x86_64.tar.gz"
   version "2.1.1"
   sha256 "ee06193b5fc4bb56746f8f648794b732b96879369835a94f22235e0561d652d7"
@@ -29,5 +27,41 @@ class Riak < Formula
     bin.write_exec_script libexec/"bin/riak-admin"
     bin.write_exec_script libexec/"bin/riak-debug"
     bin.write_exec_script libexec/"bin/search-cmd"
+  end
+
+  def recommended_maxfiles
+    65536
+  end
+
+  def recommended_maxproc
+    2048
+  end
+
+  def caveats
+    msg = nil
+    if maxfiles_too_low? || maxprocs_too_low?
+      msg = "Your maxfiles limit or maxprocs limit is too low.  See the riak docs:\n"
+      msg << "http://docs.basho.com/riak/latest/ops/tuning/open-files-limit/"
+    end
+
+    msg
+  end
+
+  def maxfiles_too_low?
+    maxfiles = `launchctl limit maxfiles`
+    _, maxfiles_soft, maxfiles_hard = maxfiles.split
+
+    maxfiles_soft.to_i < recommended_maxfiles || maxfiles_hard.to_i < recommended_maxfiles
+  end
+
+  def maxprocs_too_low?
+    maxproc= `launchctl limit maxproc`
+    _, maxproc_soft, maxproc_hard = maxproc.split
+
+    maxproc_soft.to_i < recommended_maxproc || maxproc_hard.to_i < recommended_maxproc
+  end
+
+  test do
+    system "#{bin}/riak", "help", "0"
   end
 end


Library/Formula/riak.rb
```diff
@@ -29,4 +28,128 @@ def install
...
+      else
+        puts "Edit or create /etc/launchd.conf with the following contents:"
+        mavericks_maxfile_advice  # mavericks and older actually
+      end
+    end
+  end
+
+  def post_install
+    check_maxfiles
+    check_maxproc
```
mikemcquaid: These should be in a `caveats` method. Also, I'm not sure what issue is being addressed here, can you elaborate and ideally give a reproducible test case? Is this something Homebrew-specific or is there a section of the Riak docs that mentions this somewhere?

DomT4:
As far as I'm aware, this uses a binary download, which is why we haven't been able to enforce the OpenSSL dependency so far. The test-bot's latest run through reaffirms the incorrect linkage remains.

Library/Formula/riak.rb
squarism: I'll move it to the caveats.  Thanks.

Regarding the issue, it's from the [riak docs](http://docs.basho.com/riak/latest/ops/tuning/open-files-limit/).  Developers have to hunt this information later after getting warning messages on first startup.

Library/Formula/riak.rb
```diff
@@ -8,6 +6,7 @@ class Riak < Formula
 
   depends_on :macos => :mountain_lion
   depends_on :arch => :x86_64
+  depends_on "openssl" => :recommended
```
DomT4: You can remove this sadly. The binary ignores it so we're stuck unless we build from source.

Library/Formula/riak.rb
mikemcquaid: If it complains by default I'd rather we just pointed to the correct section of the docs in the caveats as a one-liner, thanks.

DomT4:
You seem to have removed the `openssl` dep and then pushed the original commit back over the removal so we're back where we were. If you could remove the `openssl` dep and squash that'd be great. You may need to rebase on master, as it currently thinks you're adding a `desc` rather than improving an existing one.

squarism:
Thanks for the feedback and help @DomT4, hoping this last push is all straight.  :cake:

bfontaine:
Are these caveats Homebrew-related?

mikemcquaid:
Let's just remove the caveats; Riak should warn about these itself and it's unclear as-is how you'd fix this warning.

squarism:
Riak doesn't warn clearly.  If you don't want to step into that area, I'll just close this PR.  The whole point was to make the install easier instead of decoding the (eventual) riak error into "what I need to do".  :cake:

mikemcquaid:
> decoding the (eventual) riak error into "what I need to do".

Can you give step-by-step reproduction instructions for that error and indicate what command is needed to fix it? Thanks!

squarism:
No problem.  First, I'm assuming the machine has never had riak installed or has been used.

```
brew install riak
riak start
```

You get a warning.  But it's worse than it seems.

> !!!!
> !!!! WARNING: ulimit -n is 256; 65536 is the recommended minimum.
> !!!!

Riak has actually crashed.


## System Level Test

```
grep -R "CRASH REPORT" /usr/local/var/log/riak

{error,{db_open,"IO error: /usr/local/var/lib/riak/anti_entropy/1073290264914881830555831049026020342559825461248/MANIFEST-000012: Too many open files"}} in hashtree:new_segment_store/2 line 563 in gen_server:init_it/6 line 328
```

Originally, I wanted to generate the plist files and add caveats to fix this, that was not an accepted (in this thread).  So I squashed to provide caveats/links to the docs to fix this.  It doesn't really work as it stands.

Here are the plist files that make riak work (per the docs).

`/Library/LaunchDaemons/limit.maxfiles.plist`

```
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
  <plist version="1.0">
    <dict>
      <key>Label</key>
        <string>limit.maxfiles</string>
      <key>ProgramArguments</key>
        <array>
          <string>launchctl</string>
          <string>limit</string>
          <string>maxfiles</string>
          <string>65536</string>
          <string>65536</string>
        </array>
      <key>RunAtLoad</key>
        <true/>
      <key>ServiceIPC</key>
        <false/>
    </dict>
  </plist>
```

`/Library/LaunchDaemons/limit.maxproc.plist`

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple/DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
  <plist version="1.0">
    <dict>
      <key>Label</key>
        <string>limit.maxproc</string>
      <key>ProgramArguments</key>
        <array>
          <string>launchctl</string>
          <string>limit</string>
          <string>maxproc</string>
          <string>2048</string>
          <string>2048</string>
        </array>
      <key>RunAtLoad</key>
        <true />
      <key>ServiceIPC</key>
        <false />
    </dict>
  </plist>
```

## Application Level Test

I'm assuming you have a local ruby installed etc (rvm/chruby/rbenv).  `gem install riak-client`.


Failure with default OSX limits.
```
{snip}
Connection refused - connect(2) for "127.0.0.1" port 8087
{snip}
```

Riak responds to pings but (per the above log from the system level test section) has actually crashed.

Success with ulimit set.  Set with the ulimit plist or by setting procs and files manually it will work here.
```
$ ruby -e 'require "riak"; puts Riak::Client.new.bucket("foo").exist?("bar")'

=> false
```

Another simple test is `telnet localhost 8087`.  Resource limits by default keep riak from starting.

## Manually setting ulimit

In a shell:

```
riak stop
ulimit -n 65536
ulimit -u 2048
riak start
telnet localhost 8087
# Success!
```

Flipping it back to OSX defaults:
```
riak stop
ulimit -n 256
ulimit -u 256
riak start
telnet localhost 8087
# Connection refused.  :(
```

Originally, I wanted the `riak.rb` caveats to generate the plist that is needed.  But I understand that then that's not DRY if it's in the riak docs already.

bfontaine:

Library/Formula/riak.rb
```diff
@@ -1,8 +1,6 @@
-require "formula"
-
 class Riak < Formula
-  desc "Distributed database"
   homepage "http://basho.com/riak/"
+  desc "Riak is a distributed NoSQL key-value data store."
```
bfontaine: This line should be before `homepage` and not start with the formula name (`Riak is a distributed...` → `Distributed…`).

apjanke:
I'd have some reservations about telling users to adjust their system-wide launchd resource limits, even if that is what Riak recommends.

Since these are mostly soft limits that it's constrained by, couldn't they just be set in the wrapper script for Riak itself, reducing the scope of the changes? E.g. in the environment wrapper script, or in the implementation of `riak start`, stick a `ulimit -n 65536` in there, as you did in your manual test?

(I suspect the 65536 wouldn't be the real limit, and it would be clamped by the `sysctl kern.maxfiles` or `kern.maxfilesperproc` limits. But good enough; seems like the rather low 256-file limit under `launchctl` is the real limit here.)

DomT4:
A wrapper makes sense to me personally.

squarism:
I don't think that I've pointed out that this isn't just friendly advice.  This formula won't boot without these changes.  It tries and gets halfway but iirc this formula is useless without these system-wide changes.  At least on 10.10+ (iirc ... I don't have a blank mac to test on).

apjanke:
I understand. Setting `ulimit` in the wrapper, if successful, will allow it to run, without issuing a warning or requiring system-wide changes or user intervention.

Applies to 10.9, too; it has the same 256 files limit.

squarism:
Oh, I see.  A launch wrapper.  I wasn't following.  Yeah, that's cool.

tdsmith:
This is documented here: http://docs.basho.com/riak/latest/ops/tuning/open-files-limit/ -- it seems this is harder than running ulimit. Closing since it isn't Homebrew-specific. Thanks for you work here!

PR #39975, opened 2015-05-21
pre-commit 0.5.0 (new formula)

pharaujo:
pre-commit is a tool by Yelp for managing and maintaining multi-language pre-commit hooks.

### DIFF
--- a/Library/Formula/pre-commit.rb
+++ b/Library/Formula/pre-commit.rb
@@ -62,6 +62,9 @@ class PreCommit < Formula
       end
     end
 
+    #fix aspy.yaml
+    (libexec/"vendor/lib/python2.7/site-packages/aspy/__init__.py").write ''
+
     system "python", *Language::Python.setup_install_args(libexec)
 
     bin.install Dir["#{libexec}/bin/*"]
--- a/Library/Formula/pre-commit.rb
+++ b/Library/Formula/pre-commit.rb
@@ -62,8 +62,8 @@ class PreCommit < Formula
       end
     end
 
-    #fix aspy.yaml
-    (libexec/"vendor/lib/python2.7/site-packages/aspy/__init__.py").write ''
+    # fix aspy.yaml
+    (libexec/"vendor/lib/python2.7/site-packages/aspy/__init__.py").write ""
 
     system "python", *Language::Python.setup_install_args(libexec)
 
--- a/Library/Formula/pre-commit.rb
+++ b/Library/Formula/pre-commit.rb
@@ -72,6 +72,16 @@ class PreCommit < Formula
   end
 
   test do
-    system bin/"pre-commit", "help"
+    testpath.cd do
+      system "git", "init"
+      (testpath/".pre-commit-config.yaml").write <<-EOF.undent
+      -   repo: git://github.com/pre-commit/pre-commit-hooks
+          sha: 5541a6a046b7a0feab73a21612ab5d94a6d3f6f0
+          hooks:
+          -   id: trailing-whitespace
+      EOF
+      system bin/"pre-commit", "install"
+      system bin/"pre-commit", "run", "--all-files"
+    end
   end
 end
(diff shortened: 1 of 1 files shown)


Library/Formula/pre-commit.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir["#{libexec}/bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    system bin/"pre-commit", "help"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikemcquaid:
Test is failing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/26518/version=yosemite/testReport/junit/brew-test-bot/yosemite/test_pre_commit/

pharaujo:
Thanks, I'll take a look at it.

pharaujo:
I changed the test to run a typical use case for the tool. Will this be enough?

Library/Formula/pre-commit.rb
```diff
@@ -0,0 +1,87 @@
...
+    sha256 "3c88cf7df114c32cf0651d12f979b8db556d992d647493e5bdbbe2828be40007"
+  end
+
+  def install
+    ENV["PYTHONPATH"] = libexec/"vendor/lib/python2.7/site-packages"
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+
+    res = %w[argparse aspy.yaml cached-property jsonschema nodeenv]
+    res += %w[ordereddict pyyaml simplejson virtualenv]
+    res.each do |r|
```
mistydemeo: You can skip creating the array and just do `resources.each do |r|`, since there aren't any resources that you don't want to stage this way.

Library/Formula/pre-commit.rb
```diff
@@ -0,0 +1,85 @@
...
+    ENV["PYTHONPATH"] = libexec/"vendor/lib/python2.7/site-packages"
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+
+    resources.each do |r|
+      r.stage do
+        system "python", *Language::Python.setup_install_args(libexec/"vendor")
+      end
+    end
+
+    # fix aspy.yaml
```
mikemcquaid: Can you please submit this as a patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

pharaujo: I had a quick note on the commit with the fix about this. `aspy.yaml` uses a `.pth` file that's normally parsed automatically during python's initialization but only if it's in one of the default `site-packages` locations, which is not the case with vendor'ed resources. The fix just declares the `aspy` dir as a module, which is what the `.pth` file programatically does.

Do python homebrew formulas have a standard way of dealing with vendor dependencies that use `.pth` files to declare libraries?

(More info in the first 3 paragraphs: https://docs.python.org/2/library/site.html)

mikemcquaid: Paging @tdsmith for advice here.

pharaujo: Any word on this?

mikemcquaid: Not yet CC @tdsmith

tdsmith: Sorry! This is the right thing to do.

tdsmith:
@BrewTestBot test this please

tdsmith:
Thanks for the pull request and sorry for the slow review; Homebrew thrives on contributions like yours!

PR #39994, opened 2015-05-22
Elm 0.15 (new formula)

rtfeldman:
Adds a Formula for a binary distribution of the Elm programming language (version 0.15).

### DIFF
--- a/Library/Formula/elm.rb
+++ b/Library/Formula/elm.rb
@@ -1,0 +1,23 @@
+class Elm < Formula
+  homepage "http://elm-lang.org"
+  url "https://github.com/kevva/elm-bin/raw/v1.4.1/vendor/osx/elm-platform-osx.tar.gz"
+  sha256 "ec94047ca573e66aed5e6ac21ab8656b203bf9e3262f0819dfb35780c3700623"
+  version "0.15"
+
+  def install
+    [
+      "elm",
+      "elm-doc",
+      "elm-make",
+      "elm-package",
+      "elm-reactor",
+      "elm-repl"
+    ].each do |binaryFile|
+      bin.install binaryFile
+    end
+  end
+
+  test do
+    system "elm", "make", "--help"
+  end
+end


DomT4:
Is there any way at all to [build this from source](http://elm-lang.org/Install.elm#build-from-source)?

Library/Formula/elm.rb
```diff
@@ -0,0 +1,23 @@
+class Elm < Formula
+  homepage "http://elm-lang.org"
+  url "https://github.com/kevva/elm-bin/raw/v1.4.1/vendor/osx/elm-platform-osx.tar.gz"
```
DomT4: Is this an official link? I couldn't find reference to it anywhere on the website.

rtfeldman: It's the official [`npm`](http://npmjs.com) repo: https://www.npmjs.com/package/elm

@evancz can presumably vouch for it, if need be :smiley:

rtfeldman:
@DomT4 It can be [built from source](http://elm-lang.org/Install.elm#build-from-source), but that process relies on the Haskell build toolchain, which has been notoriously prone to breaking...that's why the NPM installer just uses these precompiled binaries instead.

Library/Formula/elm.rb
DomT4: It looks okay, but a :+1: from @evancz is certainly welcome.

DomT4:
I'll let @Homebrew/owners add their say on the binary/from-source issue here. We're usually fairly strict on requiring from-source, and we have the `install_cabal_package` method which I suspect would work here, but we've made exceptions before.

mikemcquaid:
Needs to be from source or go into `homebrew/binary`

rtfeldman:
@mikemcquaid `install_cabal_package` has `depends_on "ghc"` - but that will install the latest version of GHC, namely 7.10. Building from source requires GHC 7.8; is there some way to specify this via `depends_on`? (I couldn't find one.)

It's really important to us that end users be able to `brew install elm` rather than having to go through `homebrew/binary`, so I want to make this work one way or another. If there's no way to avoid building this from source, what would the recommended way be to intentionally require end users have an older version of GHC installed so that we can build form source?

Thanks!

mikemcquaid:
@rtfeldman Sorry for the delay. We don't support building with older versions of GHC, unfortunately, so the Homebrew way of doing this is to build your own version of GHC in the formula itself (obviously not practical in this case). I think it'll have to be `homebrew/binary`. Can I ask what your objection is to being in that tap?

rtfeldman:
Separately putting the binaries there is fine, but I've specifically heard people wanting to do `brew install elm` - which I think is reasonable, as that's by far the most common Homebrew UX people have come to expect. :smiley:

mikemcquaid:
They could `brew install homebrew/binary/elm` in that case (or `brew install elm` will return that in a search).

DomT4:
What's the status on this?

DomT4:
If this requires an older `ghc` to build from source let's add it to `homebrew/binary` per Mike's suggestion to work around that. Taps are fairly easy for users to find and use, and don't require any more manual intervention than `brew install homebrew/binary/elm`.

PR #39997, opened 2015-05-22
alac: test added

bfontaine:
I also fixed the homepage and URL because the website is down (not sure if that’s temporary or not).


xu-cheng:
Thanks @bfontaine

PR #40004, opened 2015-05-23
texapp 0.6.10

dunn:


Library/Formula/texapp.rb
```diff
@@ -1,23 +1,36 @@
...
 
-    Or if you have cpanminus:
-      cpanm Term::ReadLine::TTYtter
-    EOS
+  test do
+    IO.popen("#{bin}/texapp -vcheck", "r+") do |pipe|
+      assert_match "trying to find cURL ...", pipe.gets
+      pipe.puts "^C"
+      pipe.close_write
+    end
```
xu-cheng: Doesn't work if using `pipe_output` ?

xu-cheng:
Thanks @dunn

PR #40008, opened 2015-05-23
slimerjs: remove redundant option

### LINKED ISSUE #40003: SlimerJS flag description confusing
```
--without-xulrunner
	Build with xulrunner
```

Which one is it? Also, it may be worth documenting what that does precisely, i.e. if it embeds xulrunner, and if/how that works on OS X Yosemite, since xulrunner is currently not brewable on Yosemite.

DomT4:
Closes #40003


mikemcquaid:
:+1:

DomT4:
Thanks for the :+1:. I gave this one a few minutes to close by itself before I deleted my branch, seems to have fixed the non-appearance of the `Closed by` message.

PR #40016, opened 2015-05-23
antiword: test added

bfontaine:


Library/Formula/antiword.rb
```diff
@@ -1,21 +1,24 @@
...
-
 class Antiword < Formula
-  homepage 'http://www.winfield.demon.nl/'
-  url 'http://www.winfield.demon.nl/linux/antiword-0.37.tar.gz'
-  sha1 '4364f7f99cb2d37f7d1d5bc14a335ccc0c67292e'
+  homepage "http://www.winfield.demon.nl/"
+  url "http://www.winfield.demon.nl/linux/antiword-0.37.tar.gz"
+  sha256 "8e2c000fcbc6d641b0e6ff95e13c846da3ff31097801e86702124a206888f5ac"
+
+  resource "sample.doc" do
```
DomT4: Is there any way to inline this with a `(testpath/"example.doc").write` or?

bfontaine: @DomT4 No, because it’s binary data.

xu-cheng:
Thanks @bfontaine

Library/Formula/antiword.rb
DomT4: Fair enough :+1:

PR #40018, opened 2015-05-23
brew-deps: improved --tree formatting

bfontaine:
This improves the way the tree is formatted with `brew deps --tree`.

Before:
```
$ brew deps cairo --tree
cairo
|- xz
|- pkg-config
|- freetype
|  |- libpng
|  |  |- xz
|- fontconfig
|  |- pkg-config
|  |- freetype
|  |  |- libpng
|  |  |  |- xz
|- libpng
|  |- xz
|- pixman
|  |- pkg-config
|- glib
|  |- xz
|  |- pkg-config
|  |- gettext
|  |  |- xz
|  |- libffi
```

After:
```
$ brew deps cairo --tree
cairo
├── xz
├── pkg-config
├── freetype
|   └── libpng
|       └── xz
├── fontconfig
|   ├── pkg-config
|   └── freetype
|       └── libpng
|           └── xz
├── libpng
|   └── xz
├── pixman
|   └── pkg-config
└── glib
    ├── xz
    ├── pkg-config
    ├── gettext
    |   └── xz
    └── libffi
```

Update: I added the new output after @xu-cheng’s suggestion.

### DIFF
--- a/Library/Homebrew/cmd/deps.rb
+++ b/Library/Homebrew/cmd/deps.rb
@@ -1,3 +1,4 @@
+# encoding: UTF-8
 require 'formula'
 require 'ostruct'
 
@@ -63,18 +64,25 @@ module Homebrew
   def puts_deps_tree(formulae)
     formulae.each do |f|
       puts f.full_name
-      recursive_deps_tree(f, 1)
+      recursive_deps_tree(f, "")
       puts
     end
   end
 
-  def recursive_deps_tree f, level
-    f.requirements.select(&:default_formula?).each do |req|
-      puts "|  "*(level-1) + "|- :#{req.to_dependency.name}"
+  def recursive_deps_tree f, prefix
+    reqs = f.requirements.select(&:default_formula?)
+    max = reqs.length - 1
+    reqs.each_with_index do |req, i|
+      chr = i == max ? "└──" : "├──"
+      puts prefix + "#{chr} :#{req.to_dependency.name}"
     end
-    f.deps.default.each do |dep|
-      puts "|  "*(level-1) + "|- #{dep.name}"
-      recursive_deps_tree(Formulary.factory(dep.name), level+1)
+    deps = f.deps.default
+    max = deps.length - 1
+    deps.each_with_index do |dep, i|
+      chr = i == max ? "└──" : "├──"
+      prefix_ext = i == max ? "    " : "|   "
+      puts prefix + "#{chr} #{dep.name}"
+      recursive_deps_tree(Formulary.factory(dep.name), prefix + prefix_ext)
     end
   end
 end


xu-cheng:
Maybe change `└` to `└──` and `├` to `├──`, like what `tree` command does.

:+1: from me. But will let other maintainers to decide.

DomT4:
Agree with Xu, but otherwise definitely an improvement over the status quo :+1:.

mikemcquaid:
Yeh, what @xu-cheng said would be great.

xu-cheng:
Also do we need to add `# coding: utf-8` in the first line to avoid encoding problem for old version ruby? @mistydemeo may know the answer.

bfontaine:
I fixed the PR with your suggestion @xu-cheng, plus added the encoding comment.

jacknagel:
Please use `# encoding: UTF-8`, which is what we use elsewhere, for uniformity.

chdiza:
Does this require the user's terminal to be set to UTF-8?  Mine is, but not everyone's is, and I wouldn't want `--tree` to result in gibberish for them.

Also: won't this fail to display properly, depending on what one's terminal font is?

chdiza:
I ask because, e.g., mutt employs similar tricks when displaying email threads, but it looks crappy or even unreadable depending on which font and terminal encoding one uses.  I seem to recall, for instance, that Monaco (a standard terminal font if there ever was one) botched that.

I think there's a reason why the deps tree was formatted using ASCII.

mikemcquaid:
> Does this require the user's terminal to be set to UTF-8? Mine is, but not everyone's is, and I wouldn't want --tree to result in gibberish for them.

The OS X terminal is UTF-8 by default so I think this isn't a problem. Nonetheless we can try it with the default fonts to make sure it renders correctly.

mikemcquaid:
The output looks great to me.

chdiza:
> The OS X terminal is UTF-8 by default so I think this isn't a problem. Nonetheless we can try it with the default fonts to make sure it renders correctly.

But some people change it away from the default.

>  Nonetheless we can try it with the default fonts to make sure it renders correctly.

Also, it should work with more than just the "default fonts", whatever those are.  (Surely more than "Menlo", for example).

It can look like the following in pure ASCII.  It doesn't look quite as good as the PR, but it's clearly better than the status quo.
```
cairo
|  |__xz
|  |__pkg-config
|  |__freetype
|     |__libpng
|        |__xz
|  |__fontconfig
|     |__pkg-config
|     |__freetype
|        |__libpng
|           |__xz
|  |__libpng
|     |__xz
|  |__pixman
|     |__pkg-config
|  |__glib
|     |__xz
|     |__pkg-config
|     |__gettext
|        |__xz
|     |__libffi
```

mikemcquaid:
> Also, it should work with more than just the "default fonts", whatever those are. (Surely more than "Menlo", for example).

What fonts do you think it needs to work with? Please test it with them and report back here if they don't work.

chdiza:
It works with the fonts I typically use.  Again, *I* don't see garbled output from this, but I'm worried that others might.

mikemcquaid:
I think we're OK to merge as-is.

DomT4:
^ :+1:.

bfontaine:
Thank you all for the feedback, it’s merged.

PR #40020, opened 2015-05-23
awsenv 0.4.2

Luzifer:


Library/Formula/awsenv.rb
```diff
@@ -0,0 +1,66 @@
...
+class Awsenv < Formula
+  homepage "https://github.com/Luzifer/awsenv"
+  url "https://github.com/Luzifer/awsenv/archive/v0.3.1.tar.gz"
+  head "https://github.com/Luzifer/awsenv.git"
+  sha256 "106ff6acbe72bfc5e8a95a4cfbe49c5c34b4c682775e7072fe740cae8a7d3026"
+
+  depends_on "go" => :build
+
+  go_resource "github.com/Sirupsen/logrus" do
+    url "https://github.com/Sirupsen/logrus",
```
dunn: If you add '.git' to the end of the url you don't need `:using => :git`!

DomT4: ^ :+1:

Luzifer: Used the suggested `homebrew-go-resources` tool and did not change in order to have it interchangeable without manual edit… Which variant do you prefer? The removed `:using => :git` or non-edited output from `homebrew-go-resources`? (Maybe also a change to that tooling to add `.git` to github.com URLs and remove the strategy?)

DomT4: We seem to use the `.git` format everywhere else for the `go_resource`s. If the `homebrew-go-resources` tool is suggesting this syntax it's probably worth flagging a bug report upstream to them.

Luzifer: I'm quite sure (after taking a look into the code) they are using the `:using` strategy in order not to create exceptions for single hosts… Other git hosts does not support that `.git` URLs…

DomT4: Misty wrote the `go_resource` implementation so will defer to @mistydemeo, but I believe `.git` is preferred.

Luzifer:
Lets see what the bot has to say about the `.git` substitution…

Luzifer:
Hmm looks like that works fine even on gopkg.in… In that case: Nothing said on my side ;)

dunn:
Went ahead and opened samertm/homebrew-go-resources#1

Luzifer:
@dunn PTAL again at this formula as I've rewritten it according to your suggestions.

dunn:
I'm not a maintainer but it looks pretty good! Is it possible to have a test that checks more functionality?

Luzifer:
Not without messing with the users encrypted credential storage…

mikemcquaid:
It won't mess with the users information as we override `HOME`.

Luzifer:
Added some more basic checks and updated to version 0.4.2 (A full integration test is planned but not for this PR)

mikemcquaid:

For future reference the preferred commit message format for new formulae is `awsenv 0.4.2 (new formula)`.

PR #40022, opened 2015-05-23
nvm: add version to nvm-exec

dunn:
Homebrew was using 'exec' as the version, so it wouldn't update if there was a cached copy of `nvm--nvm-exec-exec`.


DomT4:
Confirmed. Nice catch.

Merged in https://github.com/Homebrew/homebrew/commit/93b6d8c26574dd5b2981479bdd2b1d55a48638b5. Thanks for your contribution to Homebrew, we appreciate it :smiley_cat:.

PR #40025, opened 2015-05-23
audit: formula version string should have digit

xu-cheng:
Inspired by #40022. And I catched two formulae with such problem.

```
$ brew audit | grep -B 1 "without a digit"
denominator:
 * Stable: version (cli) is set to a string without a digit
--
moco:
 * Stable: version (runner) is set to a string without a digit
Error: 349 problems in 258 formulae
```

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -340,6 +340,14 @@ class FormulaAuditor
       spec.patches.select(&:external?).each { |p| audit_patch(p) }
     end
 
+    %w[Stable Devel].each do |name|
+      next unless spec = formula.send(name.downcase)
+      version = spec.version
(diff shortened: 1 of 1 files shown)


dunn:
Does this check versions of resources as well?

xu-cheng:
No , because there are lots of unversioned resources. And we allow such behavior.

dunn:
Could check resource versions when they are versioned?

jacknagel:
The version field is populated no matter what (for caching reasons), so it would result in many false positives.

xu-cheng:
> The version field is populated no matter what (for caching reasons), so it would result in many false positives.

That's right. So I only apply this audit rule to formulae version rather than resources version.

mikemcquaid:
This is great :+1:

PR #40026, opened 2015-05-23
tomcat-native: depend on Java 1.7 and up, use sha256, add mirror

### LINKED ISSUE #40023: tomcat-native Java 1.7 requirement failed
Since homebrew bumped up to the latest version of tomcat-native (1.1.33) I've been unable to upgrade it.  I'm running Mac OS X 10.10.3 with the Java JDK 1.8.0_40 as my default JVM. The previous upgrades of tomcat-native (up to and including 1.1.32) installed fine under this setup. Any suggestions on what I can do to get the formula to recognise Java 1.8 as valid? Or, alternatively, does anyone know why Java 1.7 is definitely required and 1.8 won't do?

makigumo:
This should also fix #40023.

Java 1.7 is not strictly required as per [webpage](http://tomcat.apache.org/native-doc/):
> Any JDK should work but it is advisable to use the same JVM version the JVM you use with Tomcat.

### DIFF
--- a/Library/Formula/tomcat-native.rb
+++ b/Library/Formula/tomcat-native.rb
@@ -1,9 +1,8 @@
-require 'formula'
-
 class TomcatNative < Formula
-  homepage 'https://tomcat.apache.org/native-doc/'
-  url 'https://www.apache.org/dyn/closer.cgi?path=tomcat/tomcat-connectors/native/1.1.33/source/tomcat-native-1.1.33-src.tar.gz'
-  sha1 'c7626c8e5144ee8e958175c4cd034cef90eab1ed'
+  homepage "https://tomcat.apache.org/native-doc/"
+  url "https://www.apache.org/dyn/closer.cgi?path=tomcat/tomcat-connectors/native/1.1.33/source/tomcat-native-1.1.33-src.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/tomcat-native.rb
```diff
@@ -14,20 +13,20 @@ class TomcatNative < Formula
...
-        apr_path = Formula['apr'].opt_prefix
+      if build.with? "apr"
+        apr_path = Formula["apr"].opt_prefix
       else
         apr_path = "#{MacOS.sdk_path}/usr"
       end
       system "./configure", "--prefix=#{prefix}",
                             "--with-apr=#{apr_path}",
-                            "--with-java-home=#{`/usr/libexec/java_home`.chomp}",
+                            "--with-java-home=#{`/usr/libexec/java_home -v 1.7+`.chomp}",
```
xu-cheng: If it's only used in build time, let's make it `"--with-java-home=#{ENV["JAVA_HOME"]}"`

makigumo: Hmm, is $JAVA_HOME always set in Homebrew? On my system I don't rely on it.

xu-cheng: `JAVA_HOME` will be set automatically by `depends_on :java`.

xu-cheng:

PR #40027, opened 2015-05-23
formula_assertions: unused condition removed

bfontaine:
This condition repeats the first one.


jacknagel:
No, it doesn't; the case is different `MiniTest` vs `Minitest`. This is necessary to work across ruby versions that package different minitest versions.

bfontaine:
Ah sorry, I knew I was probably missing something.

PR #40033, opened 2015-05-24
unyaffs (new formula)

dunn:
This looks like the canonical fork of `unyaffs` (it's been blessed by Debian anyway) which was currently in head-only.


dunn:
Homebrew/homebrew-head-only#129

Library/Formula/unyaffs.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  head "https://github.com/ehlers/unyaffs.git"
+
+  def install
+    system "make"
+    bin.install "unyaffs"
+  end
+
+  test do
+    assert_match version.to_s, shell_output("#{bin}/unyaffs -V")
```
tdsmith: What, you couldn't find a yaffs disk image to test against? :p

tdsmith:
Thanks!

PR #40035, opened 2015-05-24
cmd/install: check time of last update if FormulaUnavailableError

dunn:
A second attempt at #39794, addressing the problems with that PR.

This was the least invasive place I could think to put the reminder while still having it be useful.

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -91,6 +91,16 @@ module Homebrew
         puts_columns(search_formulae(query))
         puts "Searching taps..."
         puts_columns(search_taps(query))
+
+        # If they haven't updated in 48 hours (172800 seconds), that
+        # might explain the error
+        master = HOMEBREW_REPOSITORY.join(".git", "refs", "heads", "master")
+        if master.exist? && (Time.now.to_i - File.mtime(master).to_i) > 172800
+          ohai "You haven't updated Homebrew in a while."
+          puts <<-EOS.undent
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/install.rb
```diff
@@ -91,6 +91,14 @@ def install
         puts_columns(search_formulae(query))
         puts "Searching taps..."
         puts_columns(search_taps(query))
+
+        if HOMEBREW_REPOSITORY.exist? &&
+           (Time.now.strftime("%s").to_i - `stat -f "%m" #{HOMEBREW_REPOSITORY}/.git/refs/heads/master`.to_i) > 172800
```
jacknagel: `HOMEBREW_REPOSITORY` will always exist (it's not the `.git` directory). Also, you can do the stat call in ruby.

```rb
master = HOMEBREW_REPOSITORY.join(".git", "refs", "heads", "master")
if master.exist? && (Time.now.to_i - File.mtime(master).to_i) > 172800
  # ...
end
```

dunn: Oh ok, cool.

Is 48 hours a reasonable time for issuing a reminder?

mikemcquaid: Would be good to have the magic number (`172800`) broken down a bit more so it's more obvious how long this is.

dunn:
Any further thoughts on this?

mikemcquaid:
Thanks @dunn!

Issue #39983, opened 2015-05-22
Installing PolarSSL breaks compilation


Aorimn:
Hi,
Since PolarSSL changed its name to mbedTLS due to ARM rebrand, the library name changed from libpolarssl.so to libmbedtls.so.

Is it possible to add a symbolic link to the brew formula of polarssl in order not to break compatibility?

DomT4:
> Is it possible to add a symbolic link to the brew formula of polarssl in order not to break compatibility?

Can you be more explicit? As in symlink the libraries from the new names to the old names?

Aorimn:
As the mbedtls/polarssl formula provides libmbedtls.so, it should be easier to make a symlink named libpolarssl.so pointing to libmbedtls.so.

DomT4:
I'm semi-reluctant to rub up against upstream's choice here, but will let @Homebrew/owners :+1: or :-1: the idea.

Aorimn:
The problem is I'm not sure against which library I should link, and it depends on the version installed by homebrew. Another solution would be to require brew to be up-to-date on clients, but I'm not sure that's really applicable or something one would want to enforce -- I'm not, for my part.

tdsmith:
Is it a bad idea to make your build scripts smart enough to figure this out? It would be a small change but this isn't unique to Homebrew's packaging; you'd run into the same problem on other platforms.

DomT4:
Going to close here. I agree these things should be handled upstream, and I'm concerned putting in a temporary Homebrew hack is going to cause more problems than it'll solve, especially since once we add these things they tend to stick around.

PR #36775, opened 2015-02-13
polarssl => mbed-tls

DomT4:
PolarSSL is/has rebranded to ‘mbed TLS’ as part of their acquisition by ARM. This includes a name change, a change of lib names, a change of license (soonish) and a change to Git names and website addresses in the near future.

Due to the fairly awful way Git handles formulae renames, thought the best way to handle this was to first boneyard PolarSSL, and then introduce the replacement here. Have adapted the affected formulae, which are thankfully few.

It looks like one has just replaced the other, but there is two separate commits, one which removes PolarSSL, and another which introduces the replacement and fixes up the formulae. If this still breaks git let me know and I'll split them into entirely different PRs.

**References:**

* https://polarssl.org/tech-updates/releases/mbedtls-1.3.10-released
* https://community.arm.com/groups/internet-of-things/blog/2015/02/09/polarssl-is-dead-long-live-mbed-tls
* https://github.com/Homebrew/homebrew-boneyard/pull/38

PR #40036, opened 2015-05-24
polarssl/mbedtls: update url, homepage and head

makigumo:
PolarSSL is now mbed TLS, the current release already has the new name. So with reference to #39983 this PR migrates the old polarssl formula to the new name "mbedtls".
An alias is provided.
Homepage and url have been updated to their new locations.

Not sure if this is the homebrew way, please review and advise.

### DIFF
--- a/Library/Aliases/polarssl
+++ b/Library/Aliases/polarssl
@@ -1,0 +1,1 @@
+../Formula/mbedtls.rb
--- a/Library/Aliases/polarssl
+++ b/Library/Aliases/polarssl
@@ -1,1 +1,0 @@
-../Formula/mbedtls.rb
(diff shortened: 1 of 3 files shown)


DomT4:
Sadly we don't yet support formula renames. See #36775 for PolarSSL specifically, and #14374 historically.

I'm happy to accept the homepage/url/head updates, but afraid we can't carry the rename at this time.

makigumo:
Ok, I see. I reverted the renaming and alias. Thank you for the clarification.

xu-cheng:
Maybe create an alias `mbed-tls` which is pointed to `polarssl`.

DomT4:
There's already an alias from `mbedtls` to `polarssl`. It seems to be just the one solid word rather than hyphenated elsewhere.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/df7a8b64502dc3d20b31a0deae6e3089c1b86639. Thanks for your contribution to Homebrew, we appreciate it :smiley_cat:.

PR #40037, opened 2015-05-24
unrtf 0.21.9

dunn:
Right now there's only one other formula that uses an RTF in its test ([`rtf2latex2e.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/rtf2latex2e.rb)) but I couldn't get `(testpath/"test.rtf").write` treated as an RTF during the test.  `unrtf` would treat the control characters as text and output stuff like this:
```html
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
<html>
<head>
<meta http-equiv="content-type" content="text/html; charset=utf-8">
<!-- Translation from RTF performed by UnRTF, version 0.21.9 -->
</head>
<body>tf1nsi hello world</body>
</html>
```
Am I just doing something wrong?

### DIFF
--- a/Library/Formula/unrtf.rb
+++ b/Library/Formula/unrtf.rb
@@ -1,20 +1,36 @@
-require 'formula'
-
 class Unrtf < Formula
-  homepage 'http://www.gnu.org/software/unrtf/'
-  url 'http://ftpmirror.gnu.org/unrtf/unrtf-0.21.5.tar.gz'
-  mirror 'http://ftp.gnu.org/gnu/unrtf/unrtf-0.21.5.tar.gz'
-  sha1 '73d22805eb7a83edf5c3c27fe036e3c33248902d'
+  homepage "https://www.gnu.org/software/unrtf/"
+  url "http://ftpmirror.gnu.org/unrtf/unrtf-0.21.9.tar.gz"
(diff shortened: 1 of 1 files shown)


dunn:
womp womp
```
 > git checkout -f a7b1f6ccda8db35e26f0b5e293030f3f5de61235
FATAL: Could not checkout null with start point a7b1f6ccda8db35e26f0b5e293030f3f5de61235
hudson.plugins.git.GitException: Could not checkout null with start point a7b1f6ccda8db35e26f0b5e293030f3f5de61235
```

DomT4:
Gave the bot a :athletic_shoe:. It's checking again now.

xu-cheng:
> I couldn't get (testpath/"test.rtf").write treated as an RTF during the test.

You can either do this:

```ruby
<<-EOS.undent
  {\\rtf1\\ansi
  {\\b hello} world
  }
EOS
```

or this:
```ruby
<<-'EOS'.undent
  {\rtf1\ansi
  {\b hello} world
  }
EOS
```

dunn:
`<<-EOS.undent` is what I was doing; `<<-'EOS'.undent` works though. Thanks!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/77bdf0f2e7859f8fe3d614f131a7d08fda9caa5a. Thanks @dunn!

PR #40044, opened 2015-05-24
[WIP] Improve version match

xu-cheng:
* [x] Prevent `foo-bar-1.0-foo.bar` being parsed as `bar` rather than `1.0`.
* [ ] Support `foo/v1.0/bar` and `foo/1.0/bar`.
* [ ] Create testcase.
* [ ] Cleanup redundant explicit version assign?

### DIFF
--- a/Library/Homebrew/version.rb
+++ b/Library/Homebrew/version.rb
@@ -349,11 +349,11 @@ class Version
     return m.captures.first unless m.nil?
 
     # e.g. https://www.openssl.org/source/openssl-0.9.8s.tar.gz
-    m = /-v?([^-]+)/.match(stem)
+    m = /-v?((?:\d+\.)*\d+[^-]*)/.match(stem)
     return m.captures.first unless m.nil?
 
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Makes sense. The only tricky thing with stuff like this is making sure we don't accidentally change the version on existing implicit version detection. Perhaps we could/should have some way of running this against all formulae in core and making sure that none of them have different versions after this?

xu-cheng:
> making sure we don't accidentally change the version on existing implicit version detection.

I will dump all formulae version to check that.

mikemcquaid:
@xu-cheng :+1: This might be something we want to add to `brew readall` or something just as it's something I've wanted to do a few times.

bfontaine:
FYI I started a list of URLs where the current system doesn’t work a few months ago: https://github.com/bfontaine/homebrew-tools/wiki/Version-Misdetections

xu-cheng:
Thanks @bfontaine

PR #40049, opened 2015-05-24
gcc: Fix "Enabling language "jit" requires --enable-host-shared".

SaberUK:
Exactly what it says on the tin.

### DIFF
--- a/Library/Formula/gcc.rb
+++ b/Library/Formula/gcc.rb
@@ -137,7 +137,7 @@ class Gcc < Formula
       args << "--enable-multilib"
     end
 
-    args << "--enable-host-shared" if build.with? "jit"
+    args << "--enable-host-shared" if build.with?("jit") || build.with?("all-languages")
 
     # Ensure correct install names when linking against libgcc_s;
     # see discussion in https://github.com/Homebrew/homebrew/pull/34303


Library/Formula/gcc.rb
```diff
@@ -137,7 +137,7 @@ def install
       args << "--enable-multilib"
     end
 
-    args << "--enable-host-shared" if build.with? "jit"
+    args << "--enable-host-shared" if build.with? "jit" or build.with? "all-languages"
```
DomT4: ` || ` rather than ` or `.

SaberUK: Changed.

SaberUK:
The build error on Mountain Lion seems to be a Jenkins issue:

    ==> brew bottle --rb gcc						Build timed out (after 60 minutes).
    Marking the build as aborted.
    Build was aborted

DomT4:
Yeah, we set the build timeouts at 1 hour to stop any one build taking up too many resources. The build limit will need bumping to about 1hr30 to pass ML.

Jack or Mike will bump it later if they need to. Given we don't test non-default options, the build success isn't super-important though.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/e43b8040439fbb076ddfe40fd2e5512dc8ea31fc. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:. Tiny nit: commit messages should be lower case.

PR #40052, opened 2015-05-24
linenoise 1.0 (new formula)

zostay:
Basic formula for the released version of the linenoise readline lib.

This should supersede #23833 from 2013.

### DIFF
--- a/Library/Formula/linenoise.rb
+++ b/Library/Formula/linenoise.rb
@@ -1,0 +1,51 @@
+class Linenoise < Formula
+  homepage "https://github.com/antirez/linenoise"
+  url "https://github.com/antirez/linenoise/archive/1.0.tar.gz"
+  sha256 "f5054a4fe120d43d85427cf58af93e56b9bb80389d507a9bec9b75531a340014"
+
+  # Added patch to build linenoise as a dynamic library
+  # Upstream Issue: https://github.com/antirez/linenoise/issues/94
+  patch :DATA
+
+  def install
+    system "make"
+    lib.install "liblinenoise.dylib"
+    include.install "linenoise.h"
+  end
+
+  test do
+    (testpath/"test.c").write <<-EOS.undent
+    #include <stdio.h>
+    #include "linenoise.h"
+
+    int main() {
+      linenoiseHistoryAdd("test");
+      return 0;
+    }
+    EOS
+
+    system ENV.cc, "test.c", "-L#{lib}", "-llinenoise", "-o", "test"
+    system "./test"
+  end
+end
+
+__END__
+--- a/Makefile
++++ b/Makefile
+@@ -1,7 +1,15 @@
++all: liblinenoise.dylib linenoise_example
++
++liblinenoise.dylib: linenoise.o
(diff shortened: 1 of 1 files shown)


Library/Formula/linenoise.rb
```diff
@@ -0,0 +1,23 @@
+class Linenoise < Formula
+  homepage "https://github.com/antirez/linenoise"
+  url "https://github.com/antirez/linenoise/archive/1.0.tar.gz"
+  sha256 "f5054a4fe120d43d85427cf58af93e56b9bb80389d507a9bec9b75531a340014"
+
+  def install
+    system "make", "install"
```
mistydemeo: The install location needs to be specified.

zostay:
My original pull request was missing a required patch. Added.

Library/Formula/linenoise.rb
```diff
@@ -0,0 +1,50 @@
+class Linenoise < Formula
+  homepage "https://github.com/antirez/linenoise"
+  url "https://github.com/antirez/linenoise/archive/1.0.tar.gz"
+  sha256 "f5054a4fe120d43d85427cf58af93e56b9bb80389d507a9bec9b75531a340014"
+
+  def patches
+    # add dylib to the makefile
+    DATA
```
DomT4: You can just make this `patch :DATA` without wrapping it in a `def`.

zostay:
I believe I've now straightened out all the strict audit and testing issues unless someone can point out a new problem.

DomT4:
One last thing and then it looks good. The patch needs a comment, and ideally a link to an upstream report about the issue, i.e.

```ruby
# Why I needed to change this
# Link to report upstream where I reported the need to patch this element
patch :DATA
```

zostay:
Added a comment and submitted an [issue upstream](https://github.com/antirez/linenoise/issues/94) to link to as suggested.

Library/Formula/linenoise.rb
```diff
@@ -0,0 +1,51 @@
+class Linenoise < Formula
+  homepage "https://github.com/antirez/linenoise"
+  url "https://github.com/antirez/linenoise/archive/1.0.tar.gz"
+  sha256 "f5054a4fe120d43d85427cf58af93e56b9bb80389d507a9bec9b75531a340014"
+
+  # Added patch to build linenoise as a dynamic library
+  # Upstream Issue: https://github.com/antirez/linenoise/issues/94
+  patch :DATA
```
mikemcquaid: Please file this as a pull-request rather than just an issue. Until then I think we shouldn't release this as a dynamic library until upstream supports it.

zostay: This patch is OS X centric and the upstream library is not, so I don't feel that it's legit as an upstream pull-request. I don't have the expertise to build such a patch.

mikemcquaid: Please remove it then, thanks.

DomT4:
Any update here?

DomT4:
This seems to have gone dead. Closing consequently.

If you pick up interest later please feel free to resubmit as a new PR with the comments above addressed, Thanks!

Thank you for the submission!

PR #39421, opened 2015-05-06
audit: tweak ftp.gnu.org language for clarity

DomT4:
Re https://github.com/Homebrew/homebrew/pull/39306#issuecomment-99269170.

We already fire off an audit warning when the `ftpmirror` URL isn't the main URL for GNU formulae, so there's no real harm to changing the language on this line if it makes people a little happier on the clarity.

PR #40054, opened 2015-05-24
audit: simplify ssl/tls audit failure messages

DomT4:
This cuts the amount of custom code/language fairly drastically, applying the same generic syntax across the different SSL/TLS enforced areas. It also cuts some of the somewhat confusing language out, we refer to things as `urls` and a decent chunk of people are thinking *"Oh I need to change the url"* when often we actually mean to say *"Change the mirror link"*. This hopefully removes that accidental confusion to a large degree.

It also makes the failure message fit onto one terminal line for almost every formula, which looks cleaner than the multi-line message we currently print for standard-size Terminals.

Changes something like:

`ftp.gnu.org urls should be https://, not http:// (url is http://ftp.gnu.org/gnu/gawk/gawk-4.1.1.tar.xz)`

Into:

`Please use https:// for http://ftp.gnu.org/gnu/gawk/gawk-4.1.1.tar.xz`

References: #39421 #39306

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -1,7 +1,7 @@
-require 'formula'
-require 'utils'
-require 'extend/ENV'
-require 'formula_cellar_checks'
+require "formula"
+require "utils"
+require "extend/ENV"
+require "formula_cellar_checks"
 
 module Homebrew
   def audit
@@ -208,9 +208,9 @@ class FormulaAuditor
           problem "Use `depends_on :hg` instead of `depends_on 'mercurial'`"
         when "ruby"
           problem "Don't use ruby as a dependency. We allow non-Homebrew ruby installations."
-        when 'gfortran'
+        when "gfortran"
           problem "Use `depends_on :fortran` instead of `depends_on 'gfortran'`"
-        when 'open-mpi', 'mpich2'
+        when "open-mpi", "mpich2"
           problem <<-EOS.undent
             There are multiple conflicting ways to install MPI. Use an MPIDependency:
               depends_on :mpi => [<lang list>]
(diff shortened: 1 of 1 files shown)


DomT4:
CC @Homebrew/owners

chdiza:
`Please use https:// for http://ftp.gnu.org/gnu/gawk/gawk-4.1.1.tar.xz`

This is awkward still.  There's no such thing as "using a prefix for" a full URL that already has a prefix. Why not have it print:

`Please use https:// as the prefix for ftp.gnu.org/gnu/gawk/gawk-4.1.1.tar.xz`
 
or

`Please change the prefix of http://ftp.gnu.org/gnu/gawk/gawk-4.1.1.tar.xz to https://`

or (shortest)

`Please use https:// for ftp.gnu.org/gnu/gawk/gawk-4.1.1.tar.xz`

DomT4:
> as the prefix for

Yeah, I'd rather keep away from using yet more `prefix` based language. We use enough `prefix` based discussions that chucking another one on gets confusing quickly. Protocol is somewhat better, but given what HTTP stands for protocol is also a bit like "ATM Machine" :smile_cat:.

> Please use https:// for ftp.gnu.org/gnu/gawk/gawk-4.1.1.tar.xz

I like this more, and I did actually start working on an implementation before I pushed this. It's not a difficult thing to get done, It just involves duplicating `url` logic in these audit sections that we're already using heavily, and then filtering the protocol from the URL only to reapply it manually. There's more room to allow mistakes to slip through the net, which is my concern on that.

Library/Homebrew/cmd/audit.rb
```diff
@@ -870,19 +870,19 @@ def audit_urls
...
+        problem "Please use https:// for #{p}"
       when %r[^http://mirrors\.kernel\.org/]
-        problem "mirrors.kernel urls should be https://, not http (url is #{p})."
+        problem "Please use https:// for #{p}"
       when %r[^http://([^/]*\.|)bintray\.com/]
-        problem "Bintray urls should be https://, not http (url is #{p})."
+        problem "Please use https:// for #{p}"
       when %r[^http://tools\.ietf\.org/]
-        problem "ietf urls should be https://, not http (url is #{p})."
+        problem "Please use https:// for #{p}"
```
bfontaine: Couldn’t this code be factored? In Ruby, the following:

```ruby
case n
when 1, 2, 3
    puts "yep"
end
```

is the same as:

```ruby
case n
when 1
    puts "yep"
when 2
    puts "yep"
when 3
    puts "yep"
end
```

DomT4: Yes, It could be. I'd left it so far because I thought it was easier to read vertically rather than horizontally:

```ruby
    urls.each do |p|
      # Skip the main url link, as it can't be made SSL/TLS yet.
      next if p =~ %r[/ftpmirror\.gnu\.org]

      case p
      when %r[^http://ftp\.gnu\.org/], %r[^http://[^/]*\.apache\.org/], %r[^http://code\.google\.com/], %r[^http://fossies\.org/],
           %r[^http://mirrors\.kernel\.org/], %r[^http://([^/]*\.|)bintray\.com/], %r[^http://tools\.ietf\.org/]
        problem "Please use https:// for #{p}"
      when %r[^http://search\.mcpan\.org/CPAN/(.*)]i
        problem "MetaCPAN url should be `https://cpan.metacpan.org/#{$1}` (url is #{p})."
      end
    end
```

It doesn't look quite as crowded as I thought it would though.

bfontaine: Maybe like this?

```ruby
when %r[^http://ftp\.gnu\.org/],
     %r[^http://[^/]*\.apache\.org/],
     %r[^http://code\.google\.com/],
     # …
```

DomT4: Nice. That could work :+1:.

dunn:
> Please use https:// as the prefix for ftp.gnu.org/gnu/gawk/gawk-4.1.1.tar.xz

> Please use https:// for ftp.gnu.org/gnu/gawk/gawk-4.1.1.tar.xz

I can see these being misunderstood, with people mistaking `ftp.` for the prefix and trying to replace _it_ with `https://`.

chdiza:
> I can see these being misunderstood, with people mistaking ftp. for the prefix and trying to replace it with https://.

I can too, which is why I still like best of all my original suggestion, in some other thread, of doing this:
```
Replace: http://blahblah.com
with:    https://blahblah.com
```
But for some reason multiline output is frowned upon.  I don't know why it's frowned upon, though.

mikemcquaid:
> But for some reason multiline output is frowned upon. I don't know why it's frowned upon, though.

It makes `brew audit` easier to parse with lots of warnings.

DomT4:
Compacted a chunk more. Some of the homepages remain outside the new `case` because of the detailed notations around them, but anything that didn't need notations or didn't have notations has been squashed into a `case`. I think the language is a decent compromise on everything proposed, and clearer and cleaner than it is now.

Would appreciate @Homebrew/owners @bfontaine :+1: / :-1: on the changes.

bfontaine:
Great work @DomT4 :+1:

mikemcquaid:
LGTM :+1:

DomT4:
Thanks both. Will merge as is with you both happy. It's only essentially glorified style changes and code compacting rather than functionality tweaks, so I won't bug the whole team for a y/n.

PR #40061, opened 2015-05-25
asciidoc new home page and head site

apjanke:
AsciiDoc has a new "official" website, asciidoc.org, along with a change in development
organization.  They have also recently migrated from Google Code to
GitHub. This change points the `asciidoc` formula's `home` and `head` at the new locations.

Reference: https://groups.google.com/forum/#!topic/asciidoc/6p7KcFRabp4

Caveat: [Moot now; see UPDATE] When you merge this PR, `brew install asciidoc --HEAD` will start failing with the error `install: filters/latex/latex2png.py: No such file or directory`. This is not an issue with the brew formula; it's because the asciidoc build on `master` is currently broken, due to outstanding issue asciidoc/asciidoc#67. The formula was working for `--HEAD` previously because it's pulling stale code from Google Code.

UPDATE 2015-05-25: AsciiDoc fixed the build on `master`, so the "Caveat" is no longer relevant. `--HEAD` works with this PR now.

### DIFF
--- a/Library/Formula/asciidoc.rb
+++ b/Library/Formula/asciidoc.rb
@@ -1,5 +1,5 @@
 class Asciidoc < Formula
-  homepage "http://www.methods.co.nz/asciidoc"
+  homepage "http://asciidoc.org/"
   url "https://downloads.sourceforge.net/project/asciidoc/asciidoc/8.6.9/asciidoc-8.6.9.tar.gz"
   sha1 "82e574dd061640561fa0560644bc74df71fb7305"
 
@@ -12,7 +12,7 @@ class Asciidoc < Formula
   end
 
   head do
-    url "https://code.google.com/p/asciidoc/", :using => :hg
+    url "https://github.com/asciidoc/asciidoc.git"
     depends_on "autoconf" => :build
   end
 


dunn:
That PR has been sitting for two months; maybe we should just remove HEAD until they get around to dealing with it?

apjanke:
Ehh, beats me. If HEAD is broken, I think it's useful for that to be visible here. Because of the additional dependencies, I think it's still useful to have a HEAD install option here as a way of testing the main asciidoc branch, especially because it has build-time dependencies.

I also like the idea that this is a small vector of social pressure to keep the build on `master` for upstream projects working.

Personally, I'd like to leave the HEAD definition in here as a mechanism for testing asciidoc on my main Mac. My view is that`brew`'s `--HEAD` should just expose the state of what upstream is publishing, good or bad. But I'd defer to the community since I'm a Homebrew noob.

Like, should "HEAD builds and installs successfully" be a prerequisite for a `head` link in a homebrew formula? Do we want to track upstream projects' build status? Or should `head` just be a simple link to whatever the head of their development trunk is, regardless of state or quality? Doesn't seem like it should be homebrew's role to track the build state of upstreams. But maybe that's a philosophical objection.

apjanke:
(Sorry if that turned in to a bit of a rant there. I don't know if there's an established practice for `head` links wrt build state etc, and I'm happy to just make this PR conform to the existing convention.)

DomT4:
Said PR was just merged upstream.

apjanke:
Woohoo!

`brew install asciidoc --HEAD` is clean for me now, on OS X 10.9.5.

```
$ brew install asciidoc --HEAD
==> Cloning https://github.com/asciidoc/asciidoc.git
Updating /Library/Caches/Homebrew/asciidoc--git
==> autoconf
==> ./configure --prefix=/usr/local/Cellar/asciidoc/HEAD
==> make install
==> Caveats
If you intend to process AsciiDoc files through an XML stage
(such as a2x for manpage generation) you need to add something
like:

  export XML_CATALOG_FILES=/usr/local/etc/xml/catalog

to your shell rc file so that xmllint can find AsciiDoc's
catalog files.

See `man 1 xmllint' for more.
==> Summary
🍺  /usr/local/Cellar/asciidoc/HEAD: 92 files, 1.2M, built in 117 seconds
[@ in /]
$
```

I guess this doesn't address the philosophical issues of whether we should point to broken upstream heads, but it's a moot issue for this PR now.

Thanks for the heads-up.

DomT4:
> I guess this doesn't address the philosophical issues of whether we should point to broken upstream heads, but it's a moot issue for this PR now.

There's no hard rule, but the general pattern recently has been if a `head` is consistently reported broken and there's no movement upstream the `head` will be dropped, since we can't offer technical support for a moving target. Otherwise they're allowed to persist, even if they break temporarily. It's pretty much taken on a case-by-case basis.

mikemcquaid:
> There's no hard rule, but the general pattern recently has been if a head is consistently reported broken and there's no movement upstream the head will be dropped, since we can't offer technical support for a moving target. Otherwise they're allowed to persist, even if they break temporarily. It's pretty much taken on a case-by-case basis.

Yes, I agree.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/77a9d614355f7dfad5ea12baa0efe06655571606. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Issue #33731, opened 2014-10-30
--cc=gcc-4.9 gets me in an infinite loop


rhendric:
Encountered this while trying to compile wine, although I can repro it with a simple `brew sh --cc=gcc-4.9`

To wit:
```
$ brew install gcc
[...]
$ brew sh --cc=gcc-4.9
[...]
$ HOMEBREW_CC_LOG_PATH=log gcc -v # this will hang; Ctrl+C
^C
$ cat log.cc
gcc called with: -v
superenv executed: gcc-4.9 -v

gcc-4.9 called with: -v
superenv executed: gcc-4.9 -v

gcc-4.9 called with: -v
superenv executed: gcc-4.9 -v
[and so on...]
```

My `brew --config`:
```
HOMEBREW_VERSION: 0.9.5
ORIGIN: https://github.com/Homebrew/homebrew.git
HEAD: 60d4c157306c6d9a40cae4a02a094f92f18576b0
Last commit: 7 hours ago
HOMEBREW_PREFIX: /Homebrew
HOMEBREW_CELLAR: /Homebrew/Cellar
CPU: dual-core 64-bit penryn
OS X: 10.6.8-i386
Xcode: 4.2
LLVM-GCC: build 2336
Clang: 3.0 build 211
X11: 2.7.7 => /opt/X11
System Ruby: 1.8.7-358
Perl: /usr/bin/perl
Python: /Homebrew/bin/python => /Homebrew/Cellar/python/2.7.8_2/Frameworks/Python.framework/Versions/2.7/bin/python2.7
Ruby: /usr/bin/ruby => /System/Library/Frameworks/Ruby.framework/Versions/1.8/usr/bin/ruby
```

jacknagel:
I think this may have been fixed by 7c92f38b1dad961cb719007f13b2a9b15773dcdf. Please correct me if I am wrong.

rhendric:
Negative @jacknagel, I can still repro with the same steps.

jacknagel:
I'm guessing that this has something to do with the xcrun cache on 10.6, but I don't know what it is.

rhendric:
If there are things I could investigate or test, I'd be happy to help. I'm mostly just limited by not understanding what the desired behavior is here.

PR #40062, opened 2015-05-25
Fix for #33731

rhendric:
The xcrun script should be skipping the ENV directory when it scans
$PATH looking for tools to run. Unfortunately, the script compares the
paths found to the real path of the ENV directory (following symlinks),
but superenv was adding the nominal path to $PATH, not following
symlinks. As a consequence, platforms with Xcode <4.3 would get into
infinite loops when trying to call non-system versions of gcc, as xcrun
calls the ENV version of gcc-X.X which calls xcrun and so on forever.

This commit changes superenv to follow symlinks when determining the bin
path to use.

### DIFF
--- a/Library/Homebrew/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/ENV/super.rb
@@ -23,7 +23,8 @@ module Superenv
   end
 
   def self.bin
-    (HOMEBREW_REPOSITORY/"Library/ENV").subdirs.reject { |d| d.basename.to_s > MacOS::Xcode.version }.max
+    bin = (HOMEBREW_REPOSITORY/"Library/ENV").subdirs.reject { |d| d.basename.to_s > MacOS::Xcode.version }.max
+    bin.realpath unless bin.nil?
   end
 
   def reset


Library/Homebrew/extend/ENV/super.rb
```diff
@@ -23,7 +23,8 @@ def self.extended(base)
   end
 
   def self.bin
-    (HOMEBREW_REPOSITORY/"Library/ENV").subdirs.reject { |d| d.basename.to_s > MacOS::Xcode.version }.max
+    bin = (HOMEBREW_REPOSITORY/"Library/ENV").subdirs.reject { |d| d.basename.to_s > MacOS::Xcode.version }.max
+    bin.realpath unless bin.nil?
```
jacknagel: Can you change this to `(HOMEBREW_LIBRARY/"ENV").realpath.subdirs.reject { ...`

That should have the same effect, but avoids the nil check.

rhendric: It's not `Library/ENV` that's the symlink, it's `Library/ENV/4.2`.

jacknagel:
Thanks!

PR #40064, opened 2015-05-25
mlogger - Log to syslog from the command line, just like `logger` but more massive

kawanet:
New formula for mlogger: https://github.com/nbrownus/mlogger


kawanet:
sent another PR #40066

mikemcquaid:
@kawanet In future please just push to the same branch and it'll update the same PR, thanks!

PR #40066, opened 2015-05-25
mlogger-1.1.2 (new formula)

kawanet:
https://github.com/nbrownus/mlogger
Log to syslog from the command line, just like `logger` but more massive


mikemcquaid:

For future reference the preferred commit message format for simple version updates is `mlogger 1.1.2 (new formula)` (no dash).

PR #40068, opened 2015-05-25
libodfgen: SSL/TLS homepage

vszakats:


Library/Formula/libodfgen.rb
```diff
@@ -1,5 +1,5 @@
 class Libodfgen < Formula
-  homepage "http://sourceforge.net/p/libwpd/wiki/libodfgen/"
+  homepage "https://sourceforge.net/p/libwpd/wiki/libodfgen/"
```
DomT4: I'm a little :neutral_face: on this. For users who aren't logged in it automatically redirects immediately to plaintext, there's no real SSL/TLS element for unregistered users.

vszakats: This _was_ the case indeed, but it seems to stay SSL/TLS now even when not being logged in. At least that's what happens in my browser since a few weeks and now with this URL.

DomT4: I can't reproduce here. Three different browsers, two different devices. All going direct to plaintext for me. Will leave this open for the other maintainers to check.

vszakats: Let's close it, it's not that important. (it may be the 'SSL Everywhere' Safari extension in action, though it works also when I disable it. On Firefox it redirects to plaintext.)

PR #40076, opened 2015-05-25
Update logstash to version 1.5.0

sjhewitt:


Library/Formula/logstash.rb
```diff
@@ -2,19 +2,20 @@
 
 class Logstash < Formula
   homepage 'http://logstash.net/'
```
vszakats: I'd suggest updating this to this permanent redirect from above URL:
https://www.elastic.co/products/logstash

DomT4: Agree with the above. Since it's automatically redirecting, we should change it.

Library/Formula/logstash.rb
```diff
@@ -2,19 +2,20 @@
...
-  url 'https://download.elasticsearch.org/logstash/logstash/logstash-1.4.2.tar.gz'
-  sha1 'd59ef579c7614c5df9bd69cfdce20ed371f728ff'
+  url 'https://download.elasticsearch.org/logstash/logstash/logstash-1.5.0.tar.gz'
+  sha1 '9729c2d31fddaabdd3d8e94c34a6d1f61d57f94a'
```
DomT4: If you could bump this checksum to `sha256` when you fix the homepage as well, that'd be :tada: :fireworks:.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/06a72539380a3d7a090bb8fa2e75feb318180f19. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Please note the ideal commit message for simple version bumps is `logstash 1.5.0`, Thanks!

PR #40078, opened 2015-05-25
InspIRCd: add a description.

SaberUK:
I couldn't see any solid rules for the description other than length so I used the value we use for the GitHub description.


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/f6da27113edea28b4100dd5304c69d018e939c69. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Tiny nit for future: commit messages should be all lowercase. Otherwise perfect, thanks!

mikemcquaid:
For future reference we also don't want to merge description-only changes for now as we're going to bulk import them shortly.

PR #40085, opened 2015-05-26
ansible: add httplib2 package as resource

L2G:
Ansible's core `uri` module depends on the `httplib2` package from PyPI.

### DIFF
--- a/Library/Formula/ansible.rb
+++ b/Library/Formula/ansible.rb
@@ -1,5 +1,6 @@
 class Ansible < Formula
   homepage "http://www.ansible.com/home"
+  revision 1
   url "http://releases.ansible.com/ansible/ansible-1.9.1.tar.gz"
   sha256 "a6f975d565723765a4d490ff40cede96833a745f38908def4950a0075f1973f5"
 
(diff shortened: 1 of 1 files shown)


Library/Formula/ansible.rb
```diff
@@ -94,11 +95,17 @@ class Ansible < Formula
...
+
   def install
     ENV["PYTHONPATH"] = libexec/"vendor/lib/python2.7/site-packages"
     ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
 
-    res = %w[pycrypto boto pyyaml paramiko markupsafe jinja2]
+    res = %w[pycrypto boto pyyaml paramiko markupsafe jinja2 httplib2]
     res += %w[isodate xmltodict kerberos pywinrm] # windows support
     res += %w[six requests websocket-client docker-py] # docker support
     res += %w[pyasn1 python-keyczar] # accelerate support
```
xu-cheng: We could just use `resources.each do |r| ...` in blow. If all of resources are used in such way.

brakon:
when will this be added??

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/23dc1cdf6dd565f95f3a004e7ad00aa288ba7cf7. Thanks @L2G!

Please note `revision` should typically go under the `sha256` rather than under the `homepage`. Otherwise :100:.

Library/Formula/ansible.rb
DomT4: Agree with Xu here. I let it slide in this PR, but in future there's no reason to not adopt `resources.each do |r|`.

astorije:
Why was this added?
Does ansible's core rely on `httplib2` to work on a host, or is it just to have the `uri` module work as a `local_action`?

If the second option, then this sadly does not make it work:

```
failed: [default -> 127.0.0.1] => {"failed": true}
msg: httplib2 is not installed

FATAL: all hosts have already failed -- aborting
```

PR #40088, opened 2015-05-26
Add package for py2app.

pr0d1r2:


mikemcquaid:
Please fix all the `brew audit --strict py2app` warnings: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/26743/version=yosemite/testReport/junit/brew-test-bot/yosemite/audit_py2app___strict/

Library/Formula/py2app.rb
```diff
@@ -0,0 +1,14 @@
...
+  url "https://pypi.python.org/packages/source/p/py2app/py2app-0.9.tar.gz"
+  sha256 '7922672f9e99b50ed931780d43405ac134988b1532dd0659ef130b824f88c59d'
+
+  version "0.9"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", "#{lib}/python2.7/site-packages"
+    system "python", "setup.py", "install", "--prefix=#{prefix}"
```
mikemcquaid: Is this a library or a binary?

pr0d1r2: This is binary used for creating OSX apps from python scripts.

DomT4:
I'd recommend looking at [jrnl](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/jrnl.rb#L42-L66) for how Python installations should be handled. You'll need to vendor `altgraph`, `modulegraph` and `macholib` which are all dependencies.

pr0d1r2:
Pass `brew audit --strict py2app.rb`

pr0d1r2:
@DomT4 dependencies included.

Library/Formula/py2app.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir[libexec/"bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    system("touch example.py")
```
DomT4: You don't need to call the system here. You can just do `touch "example.py"`.

Library/Formula/py2app.rb
```diff
@@ -0,0 +1,43 @@
...
+    system("py2applet --make-setup example.py")
```
DomT4: This should be `system "py2applet", "--make-setup", "example.py"`

pr0d1r2:
@DomT4 Done :)

DomT4:
Thanks :tada:, looks great as far as I can see.

pr0d1r2:
Achievement unlocked: Homebrew Contributor ;)

tdsmith:
Installing py2app this way, it won't work as a setuptools command; it'll only work when it's invoked as `py2applet`, which may not be what people expect. What's the advantage of installing this with Homebrew and not with pip? (There's no blanket prohibition on Homebrew packaging things that pip also packages, but sometimes Homebrew doesn't seem like a good fit.)

pr0d1r2:
@tdsmith my initial intention was to provide `py2applet` command to experiment with building Mac OS X apps without polluting Cellar with pip installments. Perhaps changing name of package to `py2applet` will be a good idea?

tdsmith:
Passing on this, sorry; I don't think a systemwide installation of the py2app script is useful.

PR #40089, opened 2015-05-26
Aap 1.093

apjanke:
This project isn't very active recently, but the release is stable, and it is used by AsciiDoc, which has an existing formula and is about as active. (It's not needed for doing normal AsciiDoc installs, but used in development for testing and making distributions.)

Audit is clean.

```
$ brew audit aap
$
```

### DIFF
--- a/Library/Formula/aap.rb
+++ b/Library/Formula/aap.rb
@@ -1,0 +1,24 @@
+class Aap < Formula
+  desc "Aap is a make-like tool to download, build, and install software"
+  homepage "http://www.a-a-p.org"
+  url "https://downloads.sourceforge.net/project/a-a-p/Aap/1.093/aap-1.093.zip"
+  sha256 "7a6c6c4a819a8379e60c679fe0c3f93eb1b74204cd7cc1c158263f4b34943001"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    # Aap only installs "man" at top level. This moves it under share/,
+    # which OS X and Homebrew use by default.
+    # Upstream bug report: http://sourceforge.net/p/a-a-p/mailman/message/34146703/
+    inreplace "main.aap", "mandir = $PREFIX/man/man1", "mandir = $PREFIX/share/man/man1"
+
+    # Aap is designed to install using itself
+    system "./aap", "install", "PREFIX=#{prefix}"
+  end
+
+  test do
+    # A dummy target definition
+    (testpath/"main.aap").write("dummy:\n\t:print OK\n")
+    system "#{bin}/aap", "dummy"
+  end
+end


Library/Formula/aap.rb
```diff
@@ -0,0 +1,39 @@
+
```
DomT4: Can :skull: the empty line up top here.

Library/Formula/aap.rb
```diff
@@ -0,0 +1,39 @@
...
+class Aap < Formula
+  homepage "http://www.a-a-p.org"
```
DomT4: You'll need to add a `desc` since this is a new formula.

Library/Formula/aap.rb
```diff
@@ -0,0 +1,39 @@
...
+  url "https://downloads.sourceforge.net/project/a-a-p/Aap/1.093/aap-1.093.zip"
+  sha256 "7a6c6c4a819a8379e60c679fe0c3f93eb1b74204cd7cc1c158263f4b34943001"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  # AAP installs "man" at top level. This moves it under share/.
```
DomT4: Use an `inreplace` inside `def install` rather than a patch.

Library/Formula/aap.rb
```diff
@@ -0,0 +1,39 @@
...
+  patch :DATA
+
+  def install
+    # Aap is designed to install using itself
+    system "./aap", "install", "PREFIX=#{prefix}"
+  end
+
+  test do
+    # We can't just do "aap --version" as a test, because it exits with an error status.
```
DomT4: No need to explain this. The test below is better than a simple `--version` test.

apjanke:
I'll make those changes. Do I need to squash and submit a different branch as a new PR, or add an additional commit on this branch?

DomT4:
> or add an additional commit on this branch?

This is preferable. You can squash in this branch though, just commit the changes and then `git rebase -i HEAD~2` and so on.

apjanke:
Okay, I've made the changes and squashed inside the branch. PR is updated.

apjanke:
BTW, I had removed the `desc` field from the template that `brew create` gave me because the Formula Cookbook [still says Homebrew has no description field](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#fill-in-the-homepage). Should that be updated?

mikemcquaid:
@apjanke Updated: https://github.com/Homebrew/homebrew/commit/e9ae59166cd1fb78bf377c82f25ce8f8f63f386f

apjanke:
Looks good. Thanks.

Library/Formula/aap.rb
```diff
@@ -0,0 +1,22 @@
...
+  desc "Aap is a make-like tool to download, build, and install software"
+  homepage "http://www.a-a-p.org"
+  url "https://downloads.sourceforge.net/project/a-a-p/Aap/1.093/aap-1.093.zip"
+  sha256 "7a6c6c4a819a8379e60c679fe0c3f93eb1b74204cd7cc1c158263f4b34943001"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    # Aap installs "man" at top level. This moves it under share/.
+    inreplace "main.aap", "mandir = $PREFIX/man/man1", "mandir = $PREFIX/share/man/man1"
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Library/Formula/aap.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    # Aap is designed to install using itself
+    system "./aap", "install", "PREFIX=#{prefix}"
+  end
+
+  test do
+    # A dummy target definition
+    (testpath/"main.aap").write("dummy:\n\t:print OK\n")
+    system "#{bin}/aap", "dummy"
```
mikemcquaid: Nice :+1:

apjanke:
> Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

I've submitted a bug report to upstream: http://sourceforge.net/p/a-a-p/mailman/message/34146703/.

I don't think I can just submit this patch to upstream, because it's platform-specific. IMHO what upstream needs isn't just to move the man pages under `share`, but to have a configuration option to relocate them under `$PREFIX` (and maybe some autodetection) so you can get the right location depending on whether or not your platform (and installation location under it) use `share` or not.

I expanded the comment about the patch a little bit and added a link to the upstream bug report. Not sure what else to say about why it's necessary...?

apjanke:
Sooooo, it turns out that Aap development is pretty inactive right now. My bug report was the first activity on the mailing list in six months.

I think it'd still be okay to add this to Homebrew, since it seems pretty stable, and it's about the same time since release as `asciidoc`, which it kind of goes with, and is in Homebrew. But I'll defer to your judgment. (I could always just put it in my private tap for my own use, so it's not critical for me to have it in core Homebrew.)

mikemcquaid:
I think I'd probably rather hold off on this until there's some movement on that mailing list post. Happy to keep this open until then.

apjanke:
All right. Let's wait and see. I'll check back in if/when I hear something.

apjanke:
Okay, I got a response from the maintainer, and I've sent in a patch for upstream that adds explicit support for relocating the `man` subdir underneath `$PREFIX`.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `aap 1.093 (new formula)`.

apjanke:
You're welcome, and thanks! Edited title for this PR and will use that format in future.

apjanke:
Should I delete my branch now to "clean things up", or leave it around for future reference? (Sorry, I'm still learning GitHub workflows.)

mikemcquaid:
Delete it.

PR #38817, opened 2015-04-19
[WIP] update: only display updated installed formulae.

mikemcquaid:
Chances are people aren't interested in modifications to formulae that aren't installed but we've modified in the same way they might be with new or deleted formulae.

Also, print the versions of these formulae too.

CC @Homebrew/owners for review. Not done yet as I wrote it on a :airplane: so not tested.

PR #40090, opened 2015-05-26
update: only display version changed formulae

xu-cheng:
An alternative approach of #38817

cc @Homebrew/owners @DomT4 @bfontaine

### DIFF
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -164,8 +164,20 @@ class Updater
         next unless paths.any? { |p| File.dirname(p) == formula_directory }
 
         case status
-        when "A", "M", "D"
+        when "A", "D"
           map[status.to_sym] << repository.join(src)
+        when "M"
+          file = repository.join(src)
+          begin
+            require "formula_versions"
+            formula = Formulary.factory(file)
+            new_version = formula.pkg_version
+            old_version = FormulaVersions.new(formula).formula_at_revision(@initial_revision, &:pkg_version)
+            next if new_version == old_version
+          rescue LoadError, FormulaUnavailableError => e
+            onoe e if ARGV.homebrew_developer?
+          end
+          map[:M] << file
         when /^R\d{0,3}/
           map[:D] << repository.join(src) if File.dirname(src) == formula_directory
           map[:A] << repository.join(dst) if File.dirname(dst) == formula_directory


mikemcquaid:
Wow, quick work here @xu-cheng (as usual). What's the output look like; just the same but with only changes to versions shown?

xu-cheng:
> just the same but with only changes to versions shown?

Yes.

DomT4:
Does this still handle the revision bump situation?

xu-cheng:
> Does this still handle the revision bump situation?

Yes as I use `pkg_version` to do the compare.

xu-cheng:
Sorry, but do we want to show or hide formulae with revision bump? Currently I use `pkg_version` so they will be shown. If we want to hide, we can use `version` to do the compare.

DomT4:
IMO, show. Revision bumps tend to be for fairly important changes.

mikemcquaid:
:+1: to show.

mikemcquaid:
If no other @Homebrew/owners have objections I'd love to see this land in the next few days.

chdiza:
As in the previous thread, I request that users be able to do `brew update --someflag` and get the current behavior as output.  Perhaps `--showall`.

mikemcquaid:
@chdiza I recommend you use `git diff` to write your own wrapper instead. I don't believe in supporting these things with options indefinitely.

dunn:
Is the objection more to proliferation of command flags or code complexity? If the former, could `HOMEBREW_DEVELOPER=1` make it revert to the old behavior?

bfontaine:
I think @chdiza’s request is easy to fulfill, it’s only one `if` away (i.e. check the version *only if* `--showall` wasn’t passed on the command-line).

Library/Homebrew/cmd/update.rb
```diff
@@ -181,8 +181,20 @@ def report
         next unless paths.any? { |p| File.dirname(p) == formula_directory }
 
         case status
-        when "A", "M", "D"
+        when "A", "D"
           map[status.to_sym] << repository.join(src)
+        when "M"
+          file = repository.join(src)
+          begin
+            require "formula_versions"
```
jacknagel: Please move this to the top of the file.

Library/Homebrew/cmd/update.rb
```diff
@@ -181,8 +181,20 @@ def report
...
+            formula = Formulary.factory(file)
+            new_version = formula.pkg_version
+            old_version = FormulaVersions.new(formula).formula_at_revision(@initial_revision, &:pkg_version)
+            next if new_version == old_version
+          rescue Exception => e
```
jacknagel: What is this protecting against? Can we narrow it down? This will catch programming errors in any code invoked in the begin block, so we should avoid it if possible.

Library/Homebrew/cmd/update.rb
xu-cheng: I kinda want to keep this as it is to minimum the chance that we accidentally break `brew update`, which has far more impact than breaking other things.

Library/Homebrew/cmd/update.rb
xu-cheng: > What is this protecting against?

Formula loading error and possible IO error inside FormulaVersions.

> This will catch programming errors in any code invoked in the begin block, so we should avoid it if possible.

I intend to protect `brew update` command as much as possible.

Library/Homebrew/cmd/update.rb
mikemcquaid: I think it's fine to move this `require` as it'll fail super-early in that case.

Library/Homebrew/cmd/update.rb
mikemcquaid: Might want to simulate those errors and catch those specific exceptions? I think protecting is good but probably better to handle the specific failure cases (e.g. check what exceptions can be thrown, check every value for `nil`)

jacknagel: This is in the diff parsing, right? Exceptions here won't break the user's repository.

mikemcquaid: My concern isn't so much the repository being broken but an error being produced which encourages users to report it to us and they do endlessly.

jacknagel: That was more of an observation than a suggestion, I think we're in agreement based on your comments above.

jacknagel:
So, a few general comments:

* I think `rescue Exception` is acceptable in the case where you need to perform some cleanup after an error, but the exception should always be re-raised in those cases. Example: https://github.com/Homebrew/homebrew/blob/3b02430b2e52cf90ddde5c64065604baf607a08e/Library/Homebrew/cmd/update.rb#L79-L82 Otherwise I would really like to avoid it going forward.

* A rule I try to follow is to keep the minimal amount of code in a begin/rescue block. So for example, the line of code that can raise `FormulaUnavailableError` should be in its own block, which only rescues that error. Then a separate block for `FormulaVersions` that catches only what we anticipate it will raise.

  This makes it clear what the expected behavior is for people who come along later, and minimizes the chances that we will rescue errors that are hiding bugs.

  Often by actively trying to do this I've found that the code fits neatly into the `begin/rescue/else` pattern, of which I am a fan as it helps reduce the need for additional control flow (`return`, `next`).

Library/Homebrew/cmd/update.rb
xu-cheng: By loading `formula_versions`, it will load almost the entire codebase. So if something is accidentally broke, users wouldn't be able to run `brew update` to download the fix. Instead they are required to  run `git` manually. I think we should avoid that.

Library/Homebrew/cmd/update.rb
xu-cheng: OK I will narrow down the exceptions.

xu-cheng:
I just found `FormulaVersions` already handles most of exceptions. So I only need to rescue `LoadError` from `requires "formula_versions"` and `FormulaUnavailableError` from `Formula.factory`.

Library/Homebrew/cmd/update.rb
```diff
@@ -164,8 +164,20 @@ def report
...
+          file = repository.join(src)
+          begin
+            require "formula_versions"
+            formula = Formulary.factory(file)
+            new_version = formula.pkg_version
+            old_version = FormulaVersions.new(formula).formula_at_revision(@initial_revision, &:pkg_version)
+            next if new_version == old_version
+          rescue LoadError, FormulaUnavailableError => e
+            onoe e if ARGV.homebrew_developer?
+          end
```
mikemcquaid: Looks good :+1:

xu-cheng:
Shall we merge this? How about the `--showall` flag(personally I think it's fine to include this tiny feature)?

mikemcquaid:
I think we should merge this as-is. Let's hold off on `showall` and see how many people we get voicing their opinions on the change. I'd rather we didn't add another option we'll have to support forever based on a couple of voices.

xu-cheng:
OK. Merged.

telemachus:
Thanks @xu-cheng for this! It's something that's long been a little pain: seeing updates that are (only) formula changes rather than version bumps. This is great.

PR #40096, opened 2015-05-26
formula_installer: set tap source for bottles.

mikemcquaid:
This defaults to the value that's in the bottle but that isn't necessarily correct. For example, some Boxen modules will reuse our old bottles and so if they are installed from there we should be sure
to set the tab's tap to the tap we've installed from rather than the one set inside the bottle's tarball.


jacknagel:
Can we add a `tap=` method to `Tab`, I'd like to keep `source` encapsulated.

mikemcquaid:
@jacknagel Sure.

PR #40097, opened 2015-05-26
test-bot: always brew install as a non-developer.

mikemcquaid:
Bored of dealing with the bottle conflicts errors here so I don't think it's really worth working around.

CC @DomT4 @xu-cheng


xu-cheng:
If it doesn't affect the bot in http://bot.brew.sh/job/Homebrew/, :+1:

DomT4:
Thanks Mike :tada:

PR #40104, opened 2015-05-26
Add update without pulling

vladshablinsky:
`brew update` switches to the master branch and pulls from it in order to update `HOMEBREW_REPOSITORY`. But for debugging purpose it's useful to be able to `update` from local commits without pulling from remote master branch. So we need to check environment variable `HOMEBREW_UPDATE_NO_PULL` and if it's set then we don't pull from master and update from the latest commit. Basically it just sets `initial_revision` to `origin/master` and `current_revision` to `HEAD`. The revisions are set by new `set_revisions_from_local` method from `Updater` class, `initial_revision` is set by new `read_origin_local_revision` method. If `HOMEBREW_UPDATE_NO_PULL` isn't set Homebrew updates in a usual way.

### DIFF
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -26,7 +26,12 @@ module Homebrew
     report = Report.new
     master_updater = Updater.new(HOMEBREW_REPOSITORY)
     begin
-      master_updater.pull!
+      if ENV["HOMEBREW_UPDATE_NO_PULL"]
+        ohai "Updating without pulling"
+        master_updater.set_revisions_from_local
+      else
+        master_updater.pull!
+      end
     ensure
       link_tap_formula(tapped_formulae)
     end
@@ -141,6 +146,11 @@ class Updater
     @repository = repository
   end
 
+  def set_revisions_from_local
+    @initial_revision = read_origin_local_revision
+    @current_revision = read_current_revision
+  end
+
   def pull!
     safe_system "git", "checkout", "-q", "master"
 
@@ -209,6 +219,10 @@ class Updater
 
   def read_current_revision
     `git rev-parse -q --verify HEAD`.chomp
+  end
+
+  def read_origin_local_revision
+    `git rev-parse -q --verify origin/master`.chomp
   end
 
   def diff


jacknagel:
What is the use case, and what is the advantage over just using git at this point?

vladshablinsky:
I'm working on formula-renames and I want to be able to simulate the behaviour like if the formula is renamed remotely, but rename it locally.

jacknagel:
It would probably be simpler to just customize the upstream here: https://github.com/Homebrew/homebrew/blob/9eb261fb4847180234a0fe07b3fd89f5f96fcf13/Library/Homebrew/cmd/update.rb#L82

I can see how it would be useful, but I don't think it's appropriate as a permanent feature that needs to be documented and maintained over time, given the small amount of users that would utilize it.

mikemcquaid:
Heads up @jacknagel: I suggested @vladshablinsky submit this (he's one of the Summer of Code students) as it is something I've personally wanted a few times too when testing changes to `update` without an internet connection so it calls all the other methods that `update` does. If it helps we could have it additionally guarded by `HOMEBREW_DEVELOPER` too.

jacknagel:
I don't like the idea of a switch that disables most of `update` and turns it into just a diff parser. If you're testing changes to update, you should be testing the entirety of its functionality to ensure other stuff isn't broken. The easiest way to do that is to just change your origin remote to point at a local repository. I've been doing that for years.

mikemcquaid:
@jacknagel Could you detail the commands you use to switch that on and off? Thanks.

bfontaine:
You could use a local directory as your origin, e.g.:

```sh
# save the current one
OLD_ORIGIN=$(git config remote.origin.url)

# switch to the local repo
git remote set-url origin /you/local/repo

# … do stuff …

# switch back
git remote set-url origin $OLD_ORIGIN
```

mikemcquaid:
Is that what you do @jacknagel?

jacknagel:
Basically, wrapped with some shell aliases.

mikemcquaid:
Ok, closing. @vladshablinsky let's talk about this in our meeting today and see if it'll work for you. Thanks for feedback @bfontaine, @jacknagel.

PR #40105, opened 2015-05-26
imagemagick: Change libtiff to recommended

sjackman:
TIFF is a pretty common file format, particularly in printing, and would be useful to include by default. `libtiff` is not a particularly large or onerous dependency.

### DIFF
--- a/Library/Formula/imagemagick.rb
+++ b/Library/Formula/imagemagick.rb
@@ -31,11 +31,11 @@ class Imagemagick < Formula
 
   depends_on "jpeg" => :recommended
   depends_on "libpng" => :recommended
+  depends_on "libtiff" => :recommended
   depends_on "freetype" => :recommended
 
   depends_on :x11 => :optional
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Let's leave this open for a while and see if it is requested by others too (as I'm personally :-1:)

sjackman:
I'm okay with leaving it open and waiting for a :+1: from someone else, but what's your reason for being :-1:, Mike? Disk space?

sjackman:
TIFF is the only file format accepted by many scientific journals and other printing houses.

mistydemeo:
I'm +1; libtiff is bottled and only 5.5MB, and it's a common default dependency of imagemagick outside Homebrew.

tdsmith:
I personally use tiff support in Imagemagick for microscopy image processing and would be happy to see it baked in by default. :)

sjackman:
Vote | Count
---|---
:+1: | 3
:-1:  | 1

PR #39107, opened 2015-04-27
Frescobaldi updated to version 2.18

Asuranceturix:

PR #40106, opened 2015-05-26
frescobaldi 2.18.1

DomT4:
Finishes up #39107.

Didn't realise `Lilypond` was still recommended in the core, I thought this had been merged already - Apologies.

### DIFF
--- a/Library/Formula/frescobaldi.rb
+++ b/Library/Formula/frescobaldi.rb
@@ -2,15 +2,15 @@ require "formula"
 
 class Frescobaldi < Formula
   homepage "http://frescobaldi.org/"
-  url "https://github.com/wbsoft/frescobaldi/releases/download/v2.0.16/frescobaldi-2.0.16.tar.gz"
-  sha1 "6b7e72def3f93aa9521d7a1cdb72399f1a5765c5"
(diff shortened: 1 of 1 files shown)


DomT4:
CC @mikemcquaid re https://github.com/Homebrew/homebrew/pull/39107#issuecomment-97507101.

Library/Formula/frescobaldi.rb
```diff
@@ -34,17 +38,21 @@ def install
...
+  def caveats
+    s = <<-EOS.undent
+      By default, a splash screen is shown on startup; this causes the main
+      window not to show until the application icon on the dock is clicked
+      (Cmd-Tab application switching does not appear to work). You may
+      want to disable the splash screen in the preferences to solve this
+      issue. See:
+        https://github.com/wbsoft/frescobaldi/issues/428
+    EOS
+    return s
```
mikemcquaid: Don't need the `return` or the `s` declaration; in Ruby the last value is implicitly returned from any function.

mikemcquaid:
Other than the nit :+1:

Library/Formula/frescobaldi.rb
DomT4: Heh, my brain skipped over the caveats entirely :smile_cat:, Will fix.

DomT4:
Should be good now. Will let the bot trundle back over it, and merge it sometime after.

PR #40109, opened 2015-05-26
vimpager 2.04

sjackman:
The man pages have been replaced with Markdown.
The man pages can be built from the Markdown with "make docs", but requires pandoc.

### DIFF
--- a/Library/Formula/vimpager.rb
+++ b/Library/Formula/vimpager.rb
@@ -1,13 +1,18 @@
 class Vimpager < Formula
   homepage "https://github.com/rkitover/vimpager"
-  url "https://github.com/rkitover/vimpager/archive/2.02.tar.gz"
-  sha256 "d0e59b33b1c21f996992ddeeaf87a63f6494b60443847acf28e3f3376acd6a05"
+  desc "vimpager: Use Vim as PAGER"
+  url "https://github.com/rkitover/vimpager/archive/2.04.tar.gz"
+  sha256 "eefbfe178ea03be3df8bbad82ba162797ab8fb49c994b4b240d513cd0c3ef3f0"
   head "https://github.com/rkitover/vimpager.git"
 
+  depends_on "pandoc" => [:build, :optional]
+
   def install
+    system "make", "docs" if build.with? "pandoc"
     bin.install "vimcat"
     bin.install "vimpager"
-    man1.install gzip("vimpager.1")
+    doc.install "README.md", "vimcat.md", "vimpager.md"
+    man1.install "vimcat.1", "vimpager.1" if build.with? "pandoc"
   end
 
   def caveats; <<-EOS.undent


DomT4:
> The man pages can be built from the Markdown with "make docs", but requires pandoc.

Is it worth adding an option to do this?

mikemcquaid:
> Is it worth adding an option to do this?

Yeh, I think so.

sjackman:
Yes, I agree. Done.

Library/Formula/vimpager.rb
```diff
@@ -1,13 +1,18 @@
 class Vimpager < Formula
   homepage "https://github.com/rkitover/vimpager"
-  url "https://github.com/rkitover/vimpager/archive/2.02.tar.gz"
-  sha256 "d0e59b33b1c21f996992ddeeaf87a63f6494b60443847acf28e3f3376acd6a05"
+  desc "vimpager: Use ViM as PAGER"
```
MindTooth: Vim, not ViM ;-)

http://www.vim.org/about.php

sjackman: Fixed.

DomT4:
I'm not sure about the `:recommended` tag on `pandoc`. Anyone building from source, anyone who installs Homebrew outside /usr/local, etc is going to have to install both `ghc` and `cabal-install` to be able to install `pandoc`. Those are some fairly major dependencies that could take users by surprise.

xu-cheng:
It should be just `:build`.

DomT4:
I was thinking `[:build, :optional]`, which is what we do for a lot of other documentation generating dependencies.

bfontaine:
> I was thinking `[:build, :optional]`, which is what we do for a lot of other documentation generating dependencies.

:+1: for that.

mikemcquaid:
It'd be good maybe if we had a `:bottle` as well which flagged that this is an option we should enable for bottling (as we can use our binary packages).

DomT4:
We've been working around that sort of thing so far by doing things like `if build.with?("docs") || build.bottle?`. Making that sort of thing more explicit though could be welcome.

sjackman:
> It'd be good maybe if we had a :bottle as well which flagged that this is an option we should enable for bottling (as we can use our binary packages).

@mikemcquaid :+1:

@DomT4 @bfontaine I'm :-1: on `:optional`. Are we really going to make the user build from source to get the man pages? Pretty much no one's going to do that. The bottles should include the man pages.

mikemcquaid:
@sjackman Yes, we are until we've got a better solution I'm afraid.

bfontaine:
> The bottles should include the man pages.

Yes I agree with that. I’m :-1: for the `:recommended` when we’re building from the sources but unfortunately there isn’t any option to target the bottles only.

sjackman:
Changed `:recommended` to `:optional`.

sjackman:
Thanks, Tim.

PR #39968, opened 2015-05-21
Change formula to use sshuttle/sshuttle fork

verdurin:
This is a simple change to use the fork at https://github.com/sshuttle/sshuttle that includes fixes for sshuttle to work in recent releases of OS X. See the discussion at https://github.com/Homebrew/homebrew/issues/37907

PR #38581, opened 2015-04-12
sshuttle 0.7.1

uraimo:
Bump version and now pointing to the repository maintained by the community[1].
This version supports Yosemite[2] but drops support for the older OSX releases, the formula selects the release 0.7.0 when installed on older OSX.
I had to add a specific version check with an If for that, let me know if there are better ways to manage this situation.

[1] https://github.com/sshuttle/sshuttle
[2] https://github.com/sshuttle/sshuttle/pull/3

PR #40120, opened 2015-05-27
Changed repository and version of sshuttle.

nkts:
I have updated reference to repository which looks maintained and up to date.


DomT4:
Already discussed and open in #39968 and #38581. Let's not open a third PR.

PR #40121, opened 2015-05-27
Update Typesafe-Activator to 1.3.3

CraftThatBlock:
Also changed to minimal distribution, which will make original download much faster, and leave the rest to activator to setup, which is recommended.

The full install (1.3.3 not minimal) URL and sha1 are:

    https://downloads.typesafe.com/typesafe-activator/1.3.3/typesafe-activator-1.3.3-minimal.zip
    ac8cc706a9b266f82a7d372f37e32f2a2ac34a30

### DIFF
--- a/Library/Formula/typesafe-activator.rb
+++ b/Library/Formula/typesafe-activator.rb
@@ -2,8 +2,8 @@ require 'formula'
 
 class TypesafeActivator < Formula
   homepage 'https://typesafe.com/activator'
-  url 'https://downloads.typesafe.com/typesafe-activator/1.3.2/typesafe-activator-1.3.2.zip'
-  sha1 'b222c96b082fe6e74e03e455fe3103817391f003'
+  url 'https://downloads.typesafe.com/typesafe-activator/1.3.3/typesafe-activator-1.3.3-minimal.zip'
+  sha1 '4930cd73b2d5b7e263287419c75df5be293b210a'
 
   def install
     rm Dir["*.bat"] # Remove Windows .bat files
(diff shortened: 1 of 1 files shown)


Library/Formula/typesafe-activator.rb
```diff
@@ -2,8 +2,8 @@
 
 class TypesafeActivator < Formula
   homepage 'https://typesafe.com/activator'
-  url 'https://downloads.typesafe.com/typesafe-activator/1.3.2/typesafe-activator-1.3.2.zip'
-  sha1 'b222c96b082fe6e74e03e455fe3103817391f003'
+  url 'https://downloads.typesafe.com/typesafe-activator/1.3.3/typesafe-activator-1.3.3-minimal.zip'
+  sha1 '4930cd73b2d5b7e263287419c75df5be293b210a'
```
mikemcquaid: You need an explicit `version` here as it's not being autodetected correctly.

CraftThatBlock: Alright, will do. Would I need to squash commits too?

bfontaine:
Thanks @CraftThatBlock! I squashed the commits for you, please note for future reference that the preferred checksum is now `sha256` instead of `sha1`. Also, the preferred commit message is `typesafe-activator 1.3.3` :wink:

CraftThatBlock:
Alright, I wasn't sure what to use for sha because it was written as 1 but Caskroom (which I contribute to) always used 256.

The commit message was different on Cask too, byt for my future contibution to brew I will follow the preferred!
Thanks!

Issue #31156, opened 2014-07-27
Strange output about Fortran compiler


dongli:
I found the following strange output about `Fortran` compiler:
```
==> Installing armadillo dependency: arpack
==> Using Homebrew-provided fortran compiler.
This may be changed by setting the FC environment variable.
==> Building with an alternative Fortran compiler
This is unsupported.
Warning: No Fortran optimization information was provided.  You may want to consider
setting FCFLAGS and FFLAGS or pass the `--default-fortran-flags` option to
`brew install` if your compiler is compatible with GCC.

If you like the default optimization level of your compiler, ignore this
warning.
```
As you can see, `homebrew` first told me that I am using the 'Homebrew-provided fortran compiler', then it told me that it will build with an alternative Fortran compiler. What is going on? I didn't do anything special.

EDIT:

I have added a print statement in `/usr/local/Library/Homebrew/extend/ENV/shared.rb` as:
```
  def fortran
    flags = []

    print "---> CHECK fc = #{fc}\n"
    if fc
```
and the output of `brew install arpack` is:
```
$ brew install arpack
---> CHECK fc =
==> Using Homebrew-provided fortran compiler.
This may be changed by setting the FC environment variable.
---> CHECK fc = /usr/local/bin/gfortran
==> Building with an alternative Fortran compiler
This is unsupported.
Warning: No Fortran optimization information was provided.  You may want to consider
setting FCFLAGS and FFLAGS or pass the `--default-fortran-flags` option to
`brew install` if your compiler is compatible with GCC.

If you like the default optimization level of your compiler, ignore this
warning.
==> Downloading http://forge.scilab.org/index.php/p/arpack-ng/downloads/get/arpack-ng_3.1.4.tar.gz
Already downloaded: /Library/Caches/Homebrew/arpack-3.1.4.tar.gz
==> ./configure --prefix=/usr/local/Cellar/arpack/3.1.4/libexec --with-blas=-L/usr/local/Cellar/veclibfort/0.4.0/lib -lvecLibFort
==> make
==> make install
🍺  /usr/local/Cellar/arpack/3.1.4: 12 files, 1.7M, built in 15 seconds
```
It seems that method `fortran` has been called twice, and for the second time, `fc` is not empty or nil, so the warning is printed without looking at if the fortran compiler is installed by homebrew or not.

UniqMartin:
This should be fixed by #47246.

PR #40128, opened 2015-05-27
Added formula for json-fortran

zbeekman:
Please add this formula to the Homebrew repository

### DIFF
--- a/Library/Formula/json-fortran.rb
+++ b/Library/Formula/json-fortran.rb
@@ -1,0 +1,46 @@
+class JsonFortran < Formula
+  desc "A Fortran 2008 JSON API"
+  homepage "https://github.com/jacobwilliams/json-fortran"
+  url "https://github.com/jacobwilliams/json-fortran/archive/4.1.1.tar.gz"
+  sha256 "97f258d28536035ef70e9ead5c7053e654106760a12db2cc652587ed61b76124"
+
+  head "https://github.com/jacobwilliams/json-fortran.git"
+
+  option "with-unicode-support", "Build json-fortran to support unicode text in json objects and files"
+  option "without-test", "Skip running build-time tests (not recommended)"
+  option "without-robodoc", "Do not build and install ROBODoc generated documentation for json-fortran"
+
+  depends_on "robodoc" => [:recommended, :build]
+  depends_on "cmake" => :build
+  depends_on :fortran
+
+  def install
+    mkdir "build" do
+      args = std_cmake_args
+      args << "-DUSE_GNU_INSTALL_CONVENTION:BOOL=TRUE" # Use more GNU/Homebrew-like install layout
+      args << "-DENABLE_UNICODE:BOOL=TRUE" if build.with? "unicode-support"
+      args << "-DSKIP_DOC_GEN:BOOL=TRUE" if build.without? "robodoc"
+      system "cmake", "..", *args
+      system "make", "all", "test" if build.with? "test" # CMake doesn't build tests when `make test`
+      system "make", "install"
+    end
+  end
+
+  test do
+    ENV.fortran
+    (testpath/"json_test.f90").write <<-EOS.undent
+      program json_test
+        use json_module
+        use ,intrinsic :: iso_fortran_env ,only: error_unit
+        implicit none
+        call json_initialize()
+        if ( json_failed() ) then
+          call json_print_error_message(error_unit)
+          stop 2
+        endif
+      end program
+    EOS
+    system ENV.fc, "-ojson_test", "-ljsonfortran", "-I#{HOMEBREW_PREFIX}/include", testpath/"json_test.f90"
+    system "./json_test"
+  end
+end


zbeekman:
It appears that the `test do` block is failing with the message ‘failed to execute cmake’ Does this mean that I should make cmake a full dependency rather than a `:build` dependency? I would love some advice on how to fix this.

zbeekman:
trying to fix the `test do` jenkins test by specifying `cmake` as both a full and a `:build` dependency. Amended previous commit and forced push

Library/Formula/json-fortran.rb
```diff
@@ -0,0 +1,73 @@
...
+  option "with-unicode-support", "Build json-fortran to support unicode text in json objects and files"
+  option "with-keg-only", "Keg only install (Skip linking json-fortran.) Required to support multiple Fortran compilers."
+  option "without-make-test", "Skip running build-time tests (not recommended)"
+
+  depends_on "cmake" => :build
+  depends_on "cmake"
+  depends_on "robodoc" => :recommended
+  depends_on :fortran
+
+  keg_only "User selected keg only install" if build.with? "keg-only"
```
DomT4: We don't support this kind of usage of `keg_only`. Typically if a user wishes to `keg_only` their installations `brew unlink` is advised.

Library/Formula/json-fortran.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def caveats; <<-EOS.undent
```
DomT4: I'd scrap the whole caveats block here. If the worst happens and it refuses to compile because their `fortran` is too old they'll never see the caveats anyway. I'm wary to get into promises about which compilers will work and which won't because it ties whoever updates the formula into checking each time that the caveats remain true. It's also worth noting that `brew` is not *always* installed into `/usr/local`.

Library/Formula/json-fortran.rb
```diff
@@ -0,0 +1,73 @@
...
+  url "https://github.com/jacobwilliams/json-fortran/archive/4.1.1.tar.gz"
+  sha256 "97f258d28536035ef70e9ead5c7053e654106760a12db2cc652587ed61b76124"
+
+  head "https://github.com/jacobwilliams/json-fortran.git"
+
+  option "with-unicode-support", "Build json-fortran to support unicode text in json objects and files"
+  option "with-keg-only", "Keg only install (Skip linking json-fortran.) Required to support multiple Fortran compilers."
+  option "without-make-test", "Skip running build-time tests (not recommended)"
+
+  depends_on "cmake" => :build
```
DomT4: Don't dupe the deps. Either make it runtime or build time, both is redundant.

Library/Formula/json-fortran.rb
```diff
@@ -0,0 +1,73 @@
...
+  desc "A Fortran 2008 JSON API"
+  homepage "https://github.com/jacobwilliams/json-fortran"
+  url "https://github.com/jacobwilliams/json-fortran/archive/4.1.1.tar.gz"
+  sha256 "97f258d28536035ef70e9ead5c7053e654106760a12db2cc652587ed61b76124"
+
+  head "https://github.com/jacobwilliams/json-fortran.git"
+
+  option "with-unicode-support", "Build json-fortran to support unicode text in json objects and files"
+  option "with-keg-only", "Keg only install (Skip linking json-fortran.) Required to support multiple Fortran compilers."
+  option "without-make-test", "Skip running build-time tests (not recommended)"
```
DomT4: `without-check` or `without-test` is the more regular syntax.

Library/Formula/json-fortran.rb
zbeekman: OK, question though: I need CMake for `brew test json-fortran` because without it the test fails AND I need it to build json-fortran but it is not a true runtime dependency. What is the correct way to specify the `:build` dependency and the dependency for `test do …`? 

I originally had only `depends_on “cmake” => :build` because cmake is needed to build the project but the test was failing when testbot ran it. (`brew test json-fortran` was working on my local machine, I guess since I have cmake installed and on my path…) It appears adding this second line fixed the testbot failure.

Library/Formula/json-fortran.rb
```diff
@@ -0,0 +1,73 @@
...
+          call json_initialize()
+          if ( json_failed() ) then
+            call json_print_error_message(error_unit)
+            stop 2
+          endif
+        end program\n")
+      add_executable ( json_test json_test.f90 )
+      target_link_libraries ( json_test jsonfortran )
+    EOS
+    system "cmake", ".", "-DCMAKE_Fortran_COMPILER=gfortran", *(std_cmake_args)
```
zbeekman: with only having `depends_on cmake => :build` the `brew test json-fortran` fails when run by testbot under the Jenkins CI setup

Library/Formula/json-fortran.rb
zbeekman: OK, but my original rational was to allow a user to install json-fortran in such a way that other versions could be installed elsewhere on the system to support other Fortran compilers. I’ll scrap this idea entirely if you like, but additional feedback would be most welcome.

Here is some background on why I attempted to let the user choose a keg-only install:
Fortran module files, and object code are not compatible between compilers (or sometimes even compiler releases from the same vendor). To install json-fortran on a system that wishes to support multiple Fortran compilers, it is necessary to install multiple versions of the library, each one built with the Fortran compiler it supports. To achieve this, the json-fortran project/library defaults to installing as:
```
<DESTDIR | PREFIX>/jsonfortran-${CMAKE_Fortran_COMPILER_ID}-${VERSION}/
├── cmake
│   ├── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-config-version.cmake
│   ├── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-config.cmake
│   ├── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-targets-release.cmake
│   └── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-targets.cmake
└── lib
    ├── json_module.mod
    ├── libjsonfortran.${VERSION}.dylib
    ├── libjsonfortran.${VERSION_MAJOR}.${VERSION_MINOR}.dylib -> libjsonfortran.${VERSION}.dylib
    ├── libjsonfortran.a
    └── libjsonfortran.dylib -> libjsonfortran.${VERSION_MAJOR}.${VERSION_MINOR}.dylib
```
Usually this means that the project is installed in `/usr/local/jsonfortran-${CMAKE_Fortran_COMPILER_ID}-${VERSION}/` and as a consequence homebrew does not create any symbolic links in `/usr/local/lib` and /usr/local/include`. Also, since the `.mod` file is binary, and compiler specific—even in its name—AND the `lib` directory is on most compilers module search paths it is just thrown in with the static and dynamic libraries.

Note, that If the user wishes to only support one Fortran compiler (which is what we anticipate anyone installing via Homebrew will want) and use a more standard GNU install structure they may  set the CMake cache variable `USE_GNU_INSTALL_CONVENTION` to  `TRUE`. This will result in the following install structure:
```
<DESTDIR | PREFIX>/
├── include
│   └── json_module.mod
└── lib
    ├── cmake
    │   └── json-fortran-${VERSION}
    │       ├── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-config-version.cmake
    │       ├── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-config.cmake
    │       ├── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-targets-release.cmake
    │       └── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-targets.cmake
    ├── libjsonfortran.${VERSION}.dylib
    ├── libjsonfortran.${VERSION_MAJOR}.${VERSION_MINOR}.dylib
    ├── libjsonfortran.a
    └── libjsonfortran.dylib
```
The thinking was to provide an option to allow a more “normal” GNU install structure, that is consistent with the other Fortran projects I have seen on Homebrew, and that will allow the libraries and module files to get symlinked by Homebrew. This is the default for the Homebrew installation, but I thought I would give the user the opportunity to use the original install structure which will support multiple compilers. However, if multiple compiler support is desired, the user should probably install json-fortran by hand.

zbeekman:
OK, I did some more experimenting and hope that the recent changes are better and clearer. Here is a summary:
 1. got rid of the keg only business, and always use `-DUSE_GNU_INSTALL_CONVENTION` when installing with Homebrew
 2. Scrapped old caveats block, added shorter, more concise caveats about supporting multiple compilers (TL;DR uninstall Homebrew installed json-fortran and go install each version by hand if this is what is wanted)
 3. Removed duplicate `depends_on cmake`. CMake is really a `:build` dependency but it is also needed when `brew test json-fortran` is called. Setting it as a `:run` dependency allows the brewtestbot to run the test successfully. If you have a better idea, @DomT4, about how I should handle the test, please let me know
4. Switched `without-make-test` to `without-test`
5. Added better description concerning the robodoc option

Library/Formula/json-fortran.rb
zbeekman: :heavy_check_mark: got rid of keg-only stuff, only supporting installs to support a single Fortran compiler now

Library/Formula/json-fortran.rb
zbeekman: :heavy_check_mark: Scrapped old caveats block, added shorter, more concise block about supporting only one Fortran compiler with Homebrew installs of json-fortran

Library/Formula/json-fortran.rb
zbeekman: :heavy_check_mark: Dependencies un-duped. `cmake` is really only a `:build` dependency, but the brewtestbot `brew test json-fortran` fails unless it is set to be a `:run` dependency. Any advice on better ways to treat this is welcome.

Library/Formula/json-fortran.rb
zbeekman: :heavy_check_mark: switched to more regular syntax

jacknagel:
Does json-fortran really require cmake, or is it just the test? If it's the test, then let's rewrite the test to not require cmake.

zbeekman:
#### json-fortran requires cmake both to build the library/project.


In addition I used cmake in the `test do` block I created. I could rewrite the `test do` block to not use cmake at all, but one of the nice things is that json-fortran provides the cmake package stuff so other cmake projects can just do:
```cmake
find_package ( jsonfortran-${CMAKE_Fortran_COMPILER_ID} 4.1.1 REQUIRED )
include_directories ( “${jsonfortran_INCLUDE_DIRS}” )
…
# add an executable
target_link_libraries ( <exe-target> jsonfortran )
…
```
to link against json-fortran once it is installed. To run the test I took advantage of this, and running the test this way makes sure that the cmake files can be found by cmake _in addition to_ making sure the library can be linked and run against. (The cmake files that get installed are jsonfortran-config.cmake, jsonfortran-targets.cmake, etc.)

Library/Formula/json-fortran.rb
```diff
@@ -0,0 +1,60 @@
...
+  option "with-unicode-support", "Build json-fortran to support unicode text in json objects and files"
+  option "without-test", "Skip running build-time tests (not recommended)"
+  depends_on "cmake" => :run
+  option "without-robodoc", "Do not build and install ROBODoc generated documentation for json-fortran"
+  depends_on "robodoc" => [:recommended, :build]
+  depends_on :fortran
+
+  def install
+    mktemp do
+      args = ["-DUSE_GNU_INSTALL_CONVENTION:BOOL=TRUE"]
```
DomT4: What does this actually do?

Library/Formula/json-fortran.rb
```diff
@@ -0,0 +1,60 @@
...
+  homepage "https://github.com/jacobwilliams/json-fortran"
+  url "https://github.com/jacobwilliams/json-fortran/archive/4.1.1.tar.gz"
+  sha256 "97f258d28536035ef70e9ead5c7053e654106760a12db2cc652587ed61b76124"
+
+  head "https://github.com/jacobwilliams/json-fortran.git"
+
+  option "with-unicode-support", "Build json-fortran to support unicode text in json objects and files"
+  option "without-test", "Skip running build-time tests (not recommended)"
+  depends_on "cmake" => :run
+  option "without-robodoc", "Do not build and install ROBODoc generated documentation for json-fortran"
```
DomT4: Don't bunch options and dependencies together. They should be two separate blocks.

Library/Formula/json-fortran.rb
```diff
@@ -0,0 +1,60 @@
...
+  depends_on "robodoc" => [:recommended, :build]
+  depends_on :fortran
+
+  def install
+    mktemp do
+      args = ["-DUSE_GNU_INSTALL_CONVENTION:BOOL=TRUE"]
+      args << "-DENABLE_UNICODE:BOOL=TRUE" if build.with? "unicode-support"
+      args << "-DSKIP_DOC_GEN:BOOL=TRUE" if build.without? "robodoc"
+      system "cmake", buildpath, *(args + std_cmake_args)
+      system "make"
```
DomT4: Does a direct `make install` after `cmake` not work?

Library/Formula/json-fortran.rb
```diff
@@ -0,0 +1,60 @@
...
+      system "make", "test" if build.with? "test"
+      system "make", "install"
+    end
+  end
+
+  def caveats; <<-EOS.undent
```
DomT4: I'm still not sure we need these. IMO if people hit problems they'll either report them here for advice or remove the build and use the upstream one instead.

jacknagel:
>json-fortran requires cmake to build the library/project.

I understand, I was trying to suss out whether cmake was required at runtime, as that was not clear. It would appear the answer is no.

>In addition I used cmake in the test do block I created. I could rewrite the test do block to not use cmake at all

The preferred solution is:

*  `depends_on "cmake" => :build`
* the test should not depend on cmake

We try to unnecessary dependencies in tests.

Library/Formula/json-fortran.rb
zbeekman: sure thing.

The intention of my grouping was to make it clear that I was over-riding the message associated with the `--without-robodoc` install option to be less cryptic and that the effect of passing that option would be that the API documentation would not get built and would not be installed.

Library/Formula/json-fortran.rb
zbeekman: a direct `make install` works, but `make test` does not work unless `make` is run before hand. 

Would you consider that a bug in json-fortran, to be solved upstream?

As a temporary workaround I can do this:
```ruby
if build.with? “test”
  system “make”
  system “make”, “test”
end
```

Library/Formula/json-fortran.rb
zbeekman: OK, I’ll completely remove all caveats.

Library/Formula/json-fortran.rb
zbeekman: Please see my [note from before](https://github.com/Homebrew/homebrew/pull/40128#discussion_r31146419) attached to the outdated diff.

This is a CMake option to modify where json-fortran files are installed, json-fortran defaults to `FALSE` for this value.

By default (when this CMake cache variable is `FALSE`) the project installs all library archive files and the Fortran `.mod` module file to `<prefix>/jsonfortran-${CMAKE_Fortran_COMPILER_ID}-${VERSION}/lib` and the CMake files used by the `find_package` CMake mechanism get installed into `jsonfortran-${CMAKE_Fortran_COMPILER_ID}-${VERSION}/cmake` This allows multiple versions of the library to be installed for use with different Fortran compilers. (The `.mod` file is like a precompiled header, and this and the object files in the library archives are not compatible between Fortran compilers. Client code MUST link against json-fortran using the same compiler that json-fortran was compiled with.) For example you may have a gfortran compiled version in `/usr/local/jsonfortran-gnu-4.1.1/` and an Intel compiled version in `/usr/local/jsonfortran-intel-4.1.1/`. The CMake `find_package` mechanism can then automatically link against the correct version of the library based on which Fortran compiler is currently being used.

However, IMO it does not make sense to use this install structure with Homebrew for the following reasons:
 - Users should manually install json-fortran if they are planning on supporting multiple Fortran compilers
 - This directory layout does not follow the usual Homebrew layout:
   - Homebrew puts `.mod` files into `include/` AFAICT for all other Fortran packages
   - The CMake files needed by `find_package` are installed into `lib/cmake/`, or at least this what the `scalapack` package seems to do
   - If the default layout is used all files end up installed into `/usr/local/homebrew/Cellar/json-fortran/<version>/jsonfortran-${CMAKE_Fortran_COMPILER_ID}-<version>/…` rather than `/usr/local/homebrew/Cellar/json-fortran/<version>/…` and this prevents the symbolic links from being deployed when `brew link` is run.

Passing `-DUSE_GNU_INSTALL_CONVENTION:BOOL=TRUE` to CMake causes the install directory structure to be more consistent with other Homebrew Fortran packages and GNU conventions at the cost of making it hard or impossible to allow json-fortran to be installed for use with multiple Fortran compilers.

zbeekman:
> The preferred solution is:

> - `depends_on “cmake” => :build`
> - the test should not depend on cmake

>We try to avoid unnecessary dependencies in tests.

I’d rather not remove the test dependency on CMake since
 - It’s already written
 - CMake is required for the build
 - The test not only tests that the library was properly installed, but also that the files in support of CMake’s `find_package ( )` mechanism were correctly installed and are functional.

Without CMake available during testing, the testing is less complete IMO, but, if we really must, I can remove CMake from the `test do …` block.

zbeekman:
@jacknagel @DomT4 do either of you have any insight into how Fortran compilers are picked and the effect of the `superenv`? I just ran `brew update` and `brew upgrade` and now cmake defaults to using the Intel Fortran compiler, `ifort` which I have on my system and Homebrew issues a warning about non-standard fortran compilers. It seems gcc5 was also installed when I ran `brew upgrade`, but I’m not sure why. As a work around for this I’m trying to test if the `FC` environment variable is set and if not, set it to `“${Formula[“gcc”].opt_bin}/gfortran”` but I suspect you will think this hackery. Without it however, my new, CMake free, tests fail because the Formula ends up getting built with Intel (even though FC isn’t set) and then the `brew test json-fortran` fails because it tries to link against it with gfortran. (Also, how does one invoke the Fortran compiler from inside `test do …`?)

zbeekman:
I wonder if #31156 is relevant to my previous comment and issue with the fortran compiler warning. If `FC` is not set, then `std_cmake_args` should include `-DCMAKE_Fortran_COMPILER` with a path to the Homebrew sanctioned Fortran compiler (and probably C and C++ too) to prevent CMake from determining which compiler to use.

Also [here](https://gist.github.com/zbeekman/b45e2a2f914f2820e72e) is a gist of my recent hacks on this formula, in the event that you find it useful. It has more history than my fork due the squashing of commits and forced pushes.

zbeekman:
OK, disregard some of my other comments, I don’t know how my environment got messed up, but things appear to be working as expected now, except, perhaps, my mind. I’ve squashed the commits into what I think could be an adequate formula, let me know what you think @jacknagel and @DomT4. I’m including some inline comments to address some of your previous concerns.

Library/Formula/json-fortran.rb
```diff
@@ -0,0 +1,49 @@
...
+        call json_initialize()
+        if ( json_failed() ) then
+          call json_print_error_message(error_unit)
+          stop 2
+        endif
+      end program
+    EOS
+
+    fc = ENV["FC"] || "#{Formula["gcc"].opt_bin}/gfortran"
+    system fc, "-ljsonfortran", "-I#{HOMEBREW_PREFIX}/include", "#{testpath}/json_test.f90"
```
iml: This is basically what `ENV.fortran` does already, so you should be able to just use
```
ENV.fortran
system ENV.fc, "-ljsonfortran", "-I#{include}", "json_test.f90"

**Comment by zbeekman** [comment]
First of all, sorry for the flurry of activity. I greatly appreciate everyones help, and hope you continue to bear with me as this is my first Formula submission and exposure to Ruby.

Assuming the latest testbot  CI checks pass, I’m going to stop working on this and consider it finished until/unless I hear something else needs adjusting. The latest iterations implements suggestions and comments from @iml, @DomT4, @jacknagel, and @mikemcquaid from over at #40102. It is more concise and the purpose of each statement, I think, is clearer.

**Comment by zbeekman** [comment]
@iml @jacknagel @DomT4:
I’m just wondering if any of you have had a chance to look at this since it’s been at least 8 days since I’ve last heard from anyone.

Thanks

File: Library/Formula/json-fortran.rb [hunk 11]
```diff
@@ -0,0 +1,45 @@
...
+  url "https://github.com/jacobwilliams/json-fortran/archive/4.1.1.tar.gz"
+  sha256 "97f258d28536035ef70e9ead5c7053e654106760a12db2cc652587ed61b76124"
+
+  head "https://github.com/jacobwilliams/json-fortran.git"
+
+  option "with-unicode-support", "Build json-fortran to support unicode text in json objects and files"
+  option "without-test", "Skip running build-time tests (not recommended)"
+  option "without-robodoc", "Do not build and install ROBODoc generated documentation for json-fortran"
+
+  depends_on "robodoc" => :recommended
```
DomT4: Presumably you don't need this after you've generated the documentation? If so:

`depends_on "robodoc" => [:recommended, :build]`

Library/Formula/json-fortran.rb
```diff
@@ -0,0 +1,45 @@
...
+  depends_on "cmake" => :build
+  depends_on :fortran
+
+  def install
+    mkdir "build" do
+      args = ["-DUSE_GNU_INSTALL_CONVENTION:BOOL=TRUE"] # Use more GNU/Homebrew-like install layout
+      args << "-DENABLE_UNICODE:BOOL=TRUE" if build.with? "unicode-support"
+      args << "-DSKIP_DOC_GEN:BOOL=TRUE" if build.without? "robodoc"
+      system "cmake", buildpath, *(args + std_cmake_args)
```
DomT4: Why `buildpath` here?

Would possibly be better as:

```ruby
mkdir "build" do
  args = std_cmake_args
  args << "-DUSE_GNU_INSTALL_CONVENTION:BOOL=TRUE"
  args << "-DENABLE_UNICODE:BOOL=TRUE" if build.with? "unicode-support"
  args << "-DSKIP_DOC_GEN:BOOL=TRUE" if build.without? "robodoc"

  system "cmake", ".", *args
```

iml: Using the parent directory with `system "cmake", "..", *args` is the usual invocation and functionally equivalent to `buildpath` here.

DomT4: Yup, it's exactly the same, but `".."` has been our accepted style thus far.

Library/Formula/json-fortran.rb
zbeekman: ah, OK, I had that before but for some reason the output of `brew info json-fortran` led me to think that I was doing something non-standard, reverting to this now

Library/Formula/json-fortran.rb
zbeekman: @DomT4 You meant to have two periods in your first comment right? I’ll fix that to match your suggestion.

@DomT4 Also do you want me to remove the comment about `-DUSE_GNU_INSTALL_CONVENTION` or leave it?

DomT4: > You meant to have two periods in your first comment right? I’ll fix that to match your suggestion.

Yeah, sorry, just fixed. It's not a huge thing, but the double-dot style matches the primary usage of `cmake` in Homebrew formulae.

>  Also do you want me to remove the comment about -DUSE_GNU_INSTALL_CONVENTION or leave it?

I'm not sure in which way the option ensures the build is more "Homebrew-like"? What's the end result with/without it?

zbeekman: @DomT4 I explained above but it got lost in outdated diffs… Here is the relevant text:

Fortran module files, and object code are not compatible between compilers (or sometimes even compiler releases from the same vendor). To install json-fortran on a system that wishes to support multiple Fortran compilers, it is necessary to install multiple versions of the library, each one built with the Fortran compiler it supports. To achieve this, the json-fortran project/library defaults to installing as:

```
<DESTDIR | PREFIX>/jsonfortran-${CMAKE_Fortran_COMPILER_ID}-${VERSION}/
├── cmake
│   ├── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-config-version.cmake
│   ├── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-config.cmake
│   ├── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-targets-release.cmake
│   └── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-targets.cmake
└── lib
    ├── json_module.mod
    ├── libjsonfortran.${VERSION}.dylib
    ├── libjsonfortran.${VERSION_MAJOR}.${VERSION_MINOR}.dylib -> libjsonfortran.${VERSION}.dylib
    ├── libjsonfortran.a
    └── libjsonfortran.dylib -> libjsonfortran.${VERSION_MAJOR}.${VERSION_MINOR}.dylib
```

Usually this means that the project is installed in `/usr/local/jsonfortran-${CMAKE_Fortran_COMPILER_ID}-${VERSION}/` and as a consequence homebrew does not create any symbolic links in `/usr/local/lib` and `/usr/local/include`. Also, since the `.mod` file is binary, and compiler specific—even in its name—AND the `lib` directory is on most compilers’ module search paths, the `.mod` file is just thrown in with the static and dynamic libraries.

Note, that If the user wishes to support *only one* Fortran compiler (which is what we anticipate anyone installing via Homebrew will want) and use a more standard GNU install structure they may set the CMake cache variable USE_GNU_INSTALL_CONVENTION to TRUE. This will result in the following install structure:

```
<DESTDIR | PREFIX>/
├── include
│   └── json_module.mod
└── lib
    ├── cmake
    │   └── json-fortran-${VERSION}
    │       ├── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-config-version.cmake
    │       ├── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-config.cmake
    │       ├── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-targets-release.cmake
    │       └── jsonfortran-${CMAKE_Fortran_COMPILER_ID}-targets.cmake
    ├── libjsonfortran.${VERSION}.dylib
    ├── libjsonfortran.${VERSION_MAJOR}.${VERSION_MINOR}.dylib
    ├── libjsonfortran.a
    └── libjsonfortran.dylib
```

DomT4: Thanks for the explanation - I missed it previously. I think we control the `prefix` but I'm fine with keeping `-DUSE_GNU_INSTALL_CONVENTION:BOOL=TRUE` and the comment, but will defer to @jacknagel @mistydemeo on Fortran stuff.

zbeekman:
OK, I’ll force push the changes you requested, assuming you don’t see anything else to fix, although my comments about the install directory structure will evaporate again. I’ll wait for approval from @jacknagel and @mistydemeo

zbeekman:
haven’t heard anything, so pushing the latest iteration to overwrite…

mikemcquaid:

For future reference the preferred commit message format for new formulae is `json-fortran 4.1.1 (new formula)`.

zbeekman:
@mikemcquaid No problem… if you like I can amend the commit and fix the message, just let me know.

zbeekman:
awesome, thanks so much!

PR #40129, opened 2015-05-27
Add OpenMP support to ImageMagick.

adah1972:
OpenMP support increases ImageMagick's performance greatly. This patch enables OpenMP when the compiler is GCC. When the compiler is Clang, or when `--without-openmp` is specified on the command line, OpenMP is disabled.

### DIFF
--- a/Library/Formula/imagemagick.rb
+++ b/Library/Formula/imagemagick.rb
@@ -18,6 +18,7 @@ class Imagemagick < Formula
   option "with-fftw", "Compile with FFTW support"
   option "with-hdri", "Compile with HDRI support"
   option "with-jp2", "Compile with Jpeg2000 support"
+  option "with-openmp", "Compile with OpenMP support"
   option "with-perl", "enable build/install of PerlMagick"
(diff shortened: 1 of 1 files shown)


Library/Formula/imagemagick.rb
```diff
@@ -48,6 +49,8 @@ class Imagemagick < Formula
   depends_on "fftw" => :optional
   depends_on "pango" => :optional
 
+  needs :openmp
```
xu-cheng: Sorry missed this. But in your case, it should be `needs :openmp if build.with? "openmp"`

adah1972: Should I make `--with-` or `--without-openmp` the default?

xu-cheng: Since, clang doesn't support openmp yet. Let's disable openmp as default.

adah1972:
OK, check again.

mikemcquaid:

For future reference the preferred commit message format is `imagemagick: add openmp option.`.

PR #40132, opened 2015-05-27
Formula: add full_name variable

xu-cheng:

### DIFF
--- a/Library/Homebrew/cmd/uses.rb
+++ b/Library/Homebrew/cmd/uses.rb
@@ -25,8 +25,8 @@ module Homebrew
             reqs = f.recursive_requirements do |dependent, req|
               Requirement.prune if ignores.any? { |ignore| req.send(ignore) } && !dependent.build.with?(req)
             end
-            deps.any? { |dep| dep.to_formula.name == ff.name } ||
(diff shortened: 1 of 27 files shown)


DomT4:
Looks good here with some testing. Didn't hit anything unexpected.

Library/Homebrew/cmd/uses.rb
```diff
@@ -25,7 +25,7 @@ def uses
             reqs = f.recursive_requirements do |dependent, req|
               Requirement.prune if ignores.any? { |ignore| req.send(ignore) } && !dependent.build.with?(req)
             end
-            deps.any? { |dep| dep.to_formula.name == ff.name } ||
+            deps.any? { |dep| dep.to_formula.full_name == ff.full_name } ||
               reqs.any? { |req| req.name == ff.name || req.class.default_formula == ff.name }
```
xu-cheng: I'm not sure whether it should be name or full_name here. Or maybe just:
```ruby
reqs.any? { |req| req.name == ff.name || req.class.default_formula == ff.name || req.class.default_formula == ff.full_name }
```

xu-cheng:
Any other comments?

mikemcquaid:
Looks good :+1:

jacknagel:
lgtm

PR #40134, opened 2015-05-27
Remove bottle_version

xu-cheng:
This requires #40132 merged first.


Library/Homebrew/bottles.rb
```diff
@@ -39,11 +38,18 @@ def bottle_tag
...
+  receipt_file = Utils.popen_read("tar", "-xOzf", bottle_file, receipt_file_path)
+  name = receipt_file_path.split("/").first
+  tap = Tab.from_tabfile_content(receipt_file, "#{bottle_file}/#{receipt_file_path}").tap
+
+  if tap.nil? || tap == "Homebrew/homebrew"
+    full_name = name
+  else
+    full_name = "#{tap.sub("homebrew-", "")}/#{name}"
+  end
+  [name, full_name]
```
mikemcquaid: I don't think this should return two arguments without changing the function name to indicate that.

xu-cheng: The name is already changed. Maybe a better name?

mikemcquaid: `bottle_both_formula_names` or something

xu-cheng: I finally choose `bottle_resolve_formula_names`.

mikemcquaid: Still a bit yuck but can't think of anything better so :+1:

xu-cheng:
Any other comments for this PR?

mikemcquaid:
This is great, love the approach.

xu-cheng:
Tweaked a bit on the method name. I think it's good to merge.

Notice after this is merged, old `*.bottle.rb` wouldn't be recognized. So I would like to pick a time when no jobs running on ci-bot to merge this.

mikemcquaid:
Ok, cool. How does the `.bottle.rb` file change? Can you paste a before/after?

xu-cheng:
> How does the .bottle.rb file change? Can you paste a before/after?

I just add an extra comment line to indicate the formula name. Like this:
```
$ cat tree-1.7.0.yosemite.bottle.rb
# tree
  bottle do
    prefix "/Users/xucheng/Coding/Github/homebrew"
    cellar "/Users/xucheng/Coding/Github/homebrew/Cellar"
    revision 1
    sha256 "ff6596f918d28daa58903a49409767c30fa7e3aef9cc1bc561ce0f2f3ce1b840" => :yosemite
  end
```

Library/Homebrew/bottles.rb
```diff
@@ -39,11 +38,19 @@ def bottle_tag
   end
 end
 
-def bottle_filename_formula_name filename
-  path = Pathname.new filename
-  version = BottleVersion.parse(path).to_s
-  basename = path.basename.to_s
-  basename.rpartition("-#{version}").first
+def bottle_resolve_formula_names bottle_file
+  receipt_file_path = Utils.popen_read("tar", "-tzf", bottle_file, "*/*/INSTALL_RECEIPT.json").chomp
```
xu-cheng: For gnu-tar (hence linuxbrew), it requires an extract flag (`--wildcards`) to make it work. Should I add os detection here,   or put it only in linuxbrew?

mikemcquaid: Only in Linuxbrew, I think.

mikemcquaid:
@xu-cheng cool.

PR #40135, opened 2015-05-27
InfluxDB v0.9.0-rc31 (devel)

gunnaraasen:


toddboom:
It looks like there's a test failure from Brewbot, but I don't think it's got anything to do with our recipe, and I can't seem to cause it to happen locally:
```
Error: undefined local variable or method `dependent' for #<Homebrew::Test:0x007fa904a48738>
```

@mikemcquaid @jacknagel Any suggestions?

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/a5be04601740a7129013bc9ad248d549c690963a. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Couple of tiny future nits: Commit messages should be lowercase, and `sha256` is the preferred checksum. Thanks! :tada:

PR #40144, opened 2015-05-27
offlineimap: 6.5.7

MindTooth:
With added conformity.

Side note:  Best way to run a proper test?  Not used to the test portion.

### DIFF
--- a/Library/Formula/offline-imap.rb
+++ b/Library/Formula/offline-imap.rb
@@ -1,14 +1,12 @@
-require "formula"
-
 class OfflineImap < Formula
   homepage "http://offlineimap.org/"
-  url "https://github.com/OfflineIMAP/offlineimap/archive/v6.5.6.tar.gz"
-  sha1 "cced67060be2922ca3ce056839b2aa1fe6b776a1"
+  url "https://github.com/OfflineIMAP/offlineimap/archive/v6.5.7.tar.gz"
+  sha256 "b7de52c7d8995e0657bb55da13531c8d6f96d828217159477c685ae408e390a3"
 
   head "https://github.com/OfflineIMAP/offlineimap.git"
 
   def install
-    prefix.install "offlineimap.conf", "offlineimap.conf.minimal"
+    etc.install "offlineimap.conf", "offlineimap.conf.minimal"
     libexec.install "bin/offlineimap" => "offlineimap.py"
     libexec.install "offlineimap"
     bin.install_symlink libexec+"offlineimap.py" => "offlineimap"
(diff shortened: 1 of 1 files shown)


Library/Formula/offline-imap.rb
```diff
@@ -1,9 +1,8 @@
-require "formula"
-
 class OfflineImap < Formula
   homepage "http://offlineimap.org/"
-  url "https://github.com/OfflineIMAP/offlineimap/archive/v6.5.6.tar.gz"
-  sha1 "cced67060be2922ca3ce056839b2aa1fe6b776a1"
+  desc "OfflineIMAP is to dispose your mailbox(es) as a local Maildir(s)."
```
DomT4: We're not adding descriptions to existing formulae for now. It's going to be done in bulk in the near future.

MindTooth:
@DomT4 Fixed :-)

Library/Formula/offline-imap.rb
```diff
@@ -14,6 +12,10 @@ def install
     bin.install_symlink libexec+"offlineimap.py" => "offlineimap"
   end
 
+  test do
+    system bin/"offlineimap", "--version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikemcquaid: Also, can you move it below `caveats` regardless? Thanks!

MindTooth:
Going out now, but I'll look later tonight.  But honestly, not sure if I can come up with a better test.  Sorta extend my beyond scope.  But I'll at least try.

mikemcquaid:
@MindTooth Thanks!

MindTooth:
@mikemcquaid:  Really not sure how to create a proper test.  The only way I see, is having a dummy IMAP account, and print out a dummy conf-file to use for testing.  As for now, I guess as long as it runs, it's okay.  But I do see that just pulling the version isn't enough.

Thoughts?

bfontaine:
There are some formulae where we can’t really write a test better than `--version` or `--help` because they rely on external (non-http) services, and I think this is one of them. I just looked at `offlineimap -h` and even `dry-run` needs access to an IMAP account. @mikemcquaid, thoughts?

mikemcquaid:
Cool, that's fine :+1:

DomT4:
Why aren't we placing this: `prefix.install "offlineimap.conf", "offlineimap.conf.minimal"` in `etc` by the way? Oversight or?

MindTooth:
I'll update it later today.  etc.install correct?

mikemcquaid:
Yup.

MindTooth:
Does it look okay now? :-)

DomT4:
It's good :+1:. Merged in https://github.com/Homebrew/homebrew/commit/eb13fad7eb0f89426ce1e867259ae283a2ed00bd, Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

PR #40156, opened 2015-05-28
geoserver 2.7.1

waynegraham:
Updates geoserver to 2.7.1

### DIFF
--- a/Library/Formula/geoserver.rb
+++ b/Library/Formula/geoserver.rb
@@ -2,8 +2,9 @@ require 'formula'
 
 class Geoserver < Formula
   homepage 'http://geoserver.org/'
-  url 'https://downloads.sourceforge.net/project/geoserver/GeoServer/2.7.0/geoserver-2.7.0-bin.zip'
-  sha1 '04184225bd8e640642dbd275b1a87a14fdb89a8e'
+  version = '2.7.1'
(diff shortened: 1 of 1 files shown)


Library/Formula/geoserver.rb
```diff
@@ -2,8 +2,9 @@
 
 class Geoserver < Formula
   homepage 'http://geoserver.org/'
-  url 'https://downloads.sourceforge.net/project/geoserver/GeoServer/2.7.0/geoserver-2.7.0-bin.zip'
-  sha1 '04184225bd8e640642dbd275b1a87a14fdb89a8e'
+  version = '2.7.1'
```
DomT4: Please revert this to how it was. We don't permit variables in the URL like this.

Library/Formula/geoserver.rb
```diff
@@ -2,8 +2,9 @@
...
+  url "https://downloads.sourceforge.net/project/geoserver/GeoServer/#{version}/geoserver-#{version}-bin.zip"
+  sha1 '9843864a5fa3d0bf20a92bc548565e3cfab876c9'
```
DomT4: If you could update this checksum to `sha256` when you revert the version change as well, that'd be :tada:. Thanks!

waynegraham:
@DomT4 Is this better?

mikemcquaid:

PR #40162, opened 2015-05-28
negfix8 8.3 (new formula)

sbugert:
Bash script using imagemagick to convert scanned negatives to postive images.

### DIFF
--- a/Library/Formula/negfix8.rb
+++ b/Library/Formula/negfix8.rb
@@ -1,0 +1,19 @@
+class Negfix8 < Formula
+  desc "Negfix8 turns scanned negative images into positives."
+  homepage "https://sites.google.com/site/negfix"
+  url "https://sites.google.com/site/negfix/downloads/negfix8.3.tgz"
+  sha256 "2f360b0dd16ca986fbaebf5873ee55044cae591546b573bb17797cbf569515bd"
+
+  depends_on "imagemagick" => "with-quantum-depth-16"
+
+  def install
+    bin.install "negfix8"
+  end
+
+  test do
+    (testpath/".negfix8/frameprofile").write "1 1 1 1 1 1 1"
+    system "#{bin}/negfix8", "-u", "frameprofile", test_fixtures("test.tiff"),
+        "#{testpath}/output.tiff"
+    assert (testpath/"output.tiff").exist?
+  end
+end


Library/Formula/negfix8.rb
```diff
@@ -0,0 +1,12 @@
+class Negfix8 < Formula
+  desc "Negfix8 turns scanned negative images into positives."
+  homepage "https://sites.google.com/site/negfix"
+  url "https://sites.google.com/site/negfix/downloads/negfix8.3.tgz"
+  sha256 "2f360b0dd16ca986fbaebf5873ee55044cae591546b573bb17797cbf569515bd"
+
+  depends_on "imagemagick" => %w(with-libtiff with-quantum-depth-16)
+
+  def install
+    bin.install 'negfix8'
```
DomT4: This should be double quotes.

Library/Formula/negfix8.rb
```diff
@@ -0,0 +1,12 @@
...
+  desc "Negfix8 turns scanned negative images into positives."
+  homepage "https://sites.google.com/site/negfix"
+  url "https://sites.google.com/site/negfix/downloads/negfix8.3.tgz"
+  sha256 "2f360b0dd16ca986fbaebf5873ee55044cae591546b573bb17797cbf569515bd"
+
+  depends_on "imagemagick" => %w(with-libtiff with-quantum-depth-16)
+
+  def install
+    bin.install "negfix8"
+  end
```
bfontaine: Could you add a test? Thanks!

sbugert:
Is there a way to add a test image file or will running it without user input suffice (what is considered a "bad test" in the cookbook)?

mikemcquaid:
> Is there a way to add a test image file or will running it without user input suffice (what is considered a "bad test" in the cookbook)?

Yeh, check `imagemagick.rb` for an example.

Library/Formula/negfix8.rb
```diff
@@ -0,0 +1,19 @@
+class Negfix8 < Formula
+  desc "Negfix8 turns scanned negative images into positives."
+  homepage "https://sites.google.com/site/negfix"
+  url "https://sites.google.com/site/negfix/downloads/negfix8.3.tgz"
+  sha256 "2f360b0dd16ca986fbaebf5873ee55044cae591546b573bb17797cbf569515bd"
+
+  depends_on "imagemagick" => %w(with-libtiff with-quantum-depth-16)
```
tdsmith: `with-libtiff` is the default now (as of today) so it should be removed here. You can just use `depends_on "imagemagick" => "with-quantum-depth-16"`.

tdsmith:
Great test, thanks!

PR #40163, opened 2015-05-28
luyten 0.4.4 (new formula)

makigumo:


Library/Formula/luyten.rb
```diff
@@ -0,0 +1,32 @@
...
+  head do
+    url "https://github.com/deathmarine/Luyten.git"
+    depends_on "maven" => :build
+  end
+
+  depends_on :java => "1.7+"
+
+  def install
+    if build.head?
+      version = `mvn org.apache.maven.plugins:maven-help-plugin:evaluate -Dexpression=project.version | grep -v '\\['`.chomp
```
makigumo: Hmm, why is this executed for the stable build and not only for the HEAD build?

Library/Formula/luyten.rb
```diff
@@ -0,0 +1,33 @@
...
+  head do
+    url "https://github.com/deathmarine/Luyten.git"
+    depends_on "maven" => :build
+  end
+
+  depends_on :java => "1.7+"
+
+  def install
+    if build.head?
+      ver = `mvn org.apache.maven.plugins:maven-help-plugin:evaluate -Dexpression=project.version | grep -v '\\['`.chomp
```
mikemcquaid: Use Ruby rather than `grep` to find this, thanks.

makigumo: Do you mean like this?
```ruby
ver = `mvn org.apache.maven.plugins:maven-help-plugin:evaluate -Dexpression=project.version`.split.grep(/^\d+\.\d+\.\d+/).uniq.first
```

mikemcquaid: Yep, that works.

makigumo:
I changed the grepping as suggested. Thanks.

mikemcquaid:

PR #40167, opened 2015-05-28
audit: enforce gnome urls

vszakats:
`gnome.org` maintainers had tweaked the mirroring logic to have actual `https` mirrors and those even have priority over plaintext ones:
https://bugzilla.gnome.org/show_bug.cgi?id=749481#c3


Library/Homebrew/cmd/audit.rb
```diff
@@ -892,7 +892,8 @@ def audit_urls
         problem "Please use https:// for #{p}"
       when %r[^http://search\.mcpan\.org/CPAN/(.*)]i
         problem "#{p} should be `https://cpan.metacpan.org/#{$1}`"
-      end
```
bfontaine: You shouldn’t remove this `end`. Add your new `when` above it :wink:

vszakats:
Sorry, corrected. Won't squash though, it'd be a messy one :(

vszakats:
Restarting this one.

PR #40168, opened 2015-05-28
update pushpin formula. original source package had incorrect version number.

jkarneges:


DomT4:
Is there anything available upstream specifying this was the case, for reference?

jkarneges:
I'm not sure if there's a way to get the history of a tag that was recreated, but here's the commit that was missing from the original source package: https://github.com/fanout/pushpin/commit/17359462e3a126caa28d5c7800bc2cde7624f860

DomT4:
Oh, you are the upstream. Sorry, didn't notice when I checked earlier, need more :coffee:. That's cool, we just like to check in these situations to make sure nothing malicious/suspect happened with the upstream repo.

Ideally a new tag would be less confusing for users/packagers/etc, but I don't see any issue with merging this relatively shortly.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `pushpin: use new 1.2.0 with correct version.`.

Also, I really strongly recommend against retagging things. It's hard for people to tell if you've been e.g. hacked, it'll break all people trying to install your software from source until we merge this PR and from Git's perspective if I've cloned your repository it won't update the tag to the correct one without my manually deleting the old tag locally. A better solution for future would be just to tag `1.2.1` or `1.2.0.1`

jkarneges:
Thanks for accepting. And great points, Mike. If this ever happens again I'll bump the version rather than try to repair.

PR #40170, opened 2015-05-28
ucon64 2.0.0 (new formula)

parashep:


Library/Formula/ucon64.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def install
+    cd "src"
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make"
+    bin.install "ucon64"
+    ohai "You can copy/move your DAT file collection to $HOME/.ucon64/dat"
```
mistydemeo: You can move these comments into a `caveats` block.

Library/Formula/ucon64.rb
```diff
@@ -0,0 +1,22 @@
+class Ucon64 < Formula
+  desc "A ROM backup tool and emulator's Swiss Army knife program."
+  homepage "http://ucon64.sourceforge.net/"
+  url "https://downloads.sourceforge.net/ucon64/ucon64-2.0.0-src.tar.gz"
+  sha256 "62064324a1912387f84ac9d4c521c5d5e7b80f2567e9f61bf0ab3e1d976c0127"
+
+  def install
+    cd "src"
```
mistydemeo: Use the block form of `cd` please:

```ruby
cd "src" do
  system "./configure" # ...
end

**Comment by mistydemeo** [comment]
This doesn't install discmage, which gets built during the normal build process - maybe worth fixing up where it installs/checks for it? Something like this at the top of the install method:

```ruby
    # ucon64 normally installs its private discmage library into ~/,
    # and dlopens it from the path in the ~/.ucon64rc
    # We want to install it in the prefix, of course, and update the
    # default .ucon64rc to point inside the keg.
    # There is no example .ucon64rc, but its default values are compiled
    # into ucon64 so it can generate a new .ucon64rc if none exists.
    inreplace "src/ucon64_misc.c", "~/.ucon64/discmage.dylib",
                                   opt_prefix/"libexec/libdiscmage.dylib"
```

And then at the bottom:

```ruby
libexec.install "libdiscmage/discmage.so" => "libdiscmage.dylib"
```

parashep:
thanks for the help. I will make some changes and send a new commit!
I had left out discmage because I couldn't think of a simple solution (I have never used Ruby) so your advice is definitely valuable.

mistydemeo:
@BrewTestBot test this please

Library/Formula/ucon64.rb
```diff
@@ -0,0 +1,37 @@
...
+    inreplace "src/ucon64_misc.c", "~/.ucon64/discmage.dylib",
+                                   opt_prefix/"libexec/libdiscmage.dylib"
+
+    cd "src" do
+      system "./configure", "--disable-debug",
+                            "--disable-dependency-tracking",
+                            "--disable-silent-rules",
+                            "--prefix=#{prefix}"
+      system "make"
+      bin.install "ucon64"
```
mikemcquaid: Is there a `make install` we can use (even if it needs patched)?

Library/Formula/ucon64.rb
```diff
@@ -0,0 +1,37 @@
...
+      libexec.install "libdiscmage/discmage.so" => "libdiscmage.dylib"
+    end
+  end
+
+  def caveats
+    <<-EOS.undent
+      You can copy/move your DAT file collection to $HOME/.ucon64/dat
+      Be sure to check $HOME/.ucon64rc for some options after you've run uCON64 once.
+    EOS
```
mikemcquaid: Probably can remove these unless they are Homebrew-specific.

Library/Formula/ucon64.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  test do
+    system "#{bin}/ucon64", "--help", "--snes"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikemcquaid:
This may be a better fit for the `homebrew-games` tap but let's work through review on this PR first. Thanks for submitting!

Library/Formula/ucon64.rb
parashep: There is, it calls to run install.sh and is a bit of a mess which asks for root's password, and does a lot extra that is not needed (or wanted). The FreeBSD port recreates the important parts of the install so I did the same here. If it's preferred to create a diff to patch with, I can change it to do that.

Library/Formula/ucon64.rb
parashep: I couldn't think of anything, but I'm sure I could think of something better. :wink:

mikemcquaid: :grinning: Well, if you can't, no worries, justlet me know1

Library/Formula/ucon64.rb
mikemcquaid: I guess it depends if you think the patch would ever be accepted upstream. If not: feel free not to bother. Thanks!

Library/Formula/ucon64.rb
parashep: The program does all its work on binary files, and I don't think I can generate one that would work with it. If including one is out of the question, then I think this is as good as it gets!

mikemcquaid: You could create a resource and only `stage` it during the `test` block; that's essentially equivalent. Good idea!

mistydemeo: Maybe a freeware or PD ROM like Super Bat Puncher? E.g add it as a resource, then.

```ruby
test do
  testpath.stage resource("super_bat_puncher_demo")

  assert_equal "00000000  4e 45 53 1a  08 00 11 00  00 00 00 00  00 00 00 00",
               shell_output("#{bin}/ucon64 #{path_to_rom.nes}")
end
```

mistydemeo:
I think this is ready to go with a test, so migrated this to homebrew-games - https://github.com/Homebrew/homebrew-games/pull/270

mistydemeo:
Merged in Homebrew/homebrew-games@3e993dbf08f603c1682f529867caf7f0f18e6a48.

PR #40172, opened 2015-05-29
audit: enforce gnome urls

vszakats:
`gnome.org` maintainers had tweaked the mirroring logic to have actual `https` mirrors and those even have priority over plaintext ones:
https://bugzilla.gnome.org/show_bug.cgi?id=749481#c3


vszakats:
This is a restart of https://github.com/Homebrew/homebrew/pull/40167.

mikemcquaid:
Thanks @vszakats!

PR #40174, opened 2015-05-29
aurl: add aurl 0.3.1

dai0304:


tdsmith:
Thanks for the pull request! A few things:

* New formulas should have a brief description, like `desc "A widget that does a thing."`
* A test that fetches a website would be good
* I'm not certain this meets our notability guidelines. I see you're already distributing this in a tap; that might be the best solution. Do other maintainers have thoughts?

dai0304:
Thank you for your revew.
I added description and re-pushed.

A test to fetch some website is difficult because it requres oauth client informations.

As of now aurl is distributed via tap.  But I want to avoid taps.  Current [tap repository](https://github.com/classmethod-aws/homebrew-cm/blob/master/aurl.rb) will discontinued when this PR is merged.

mikemcquaid:
I'm afraid this doesn't meet our notability requirements yet. Sorry! I recommend you continue to maintain it in a tap.

dai0304:
OK.  I'll re-request when aurl become popular.
Thanks.

PR #40177, opened 2015-05-29
Moved awsenv to tap

Luzifer:


Library/Formula/awsenv.rb
```diff
@@ -2,16 +2,9 @@
...
-  url "https://github.com/Luzifer/awsenv/archive/v0.4.2.tar.gz"
+  url "https://github.com/Luzifer/awsenv/archive/v0.4.3.tar.gz"
   head "https://github.com/Luzifer/awsenv.git"
-  sha256 "10667db8365c1f74ec0a789db6e822f2d7c8342a80c0bec95113318c113c1d50"
-  bottle do
-    cellar :any
-    sha256 "a5d568e2f756fcc1b795be44a37dfb5ca9c83d1ee0e23d285b8d343339dcba51" => :yosemite
-    sha256 "62bd5d3903007fe3a101a4afe285cd620d0204f6ba379de7fbd161a14296bea1" => :mavericks
-    sha256 "aea71464130719792e0ce96e4c61136dc174ffe4c4a5e52f1be730d27953b326" => :mountain_lion
-  end
```
mikemcquaid: Please leave the bottle block as-is and it'll be updated when we pull. Thanks!

Luzifer:
Put the bottles back in place.

Library/Formula/awsenv.rb
```diff
@@ -2,9 +2,11 @@
 
 class Awsenv < Formula
   homepage "https://github.com/Luzifer/awsenv"
-  url "https://github.com/Luzifer/awsenv/archive/v0.4.2.tar.gz"
+  desc "awsenv is a credential store for people using one or more AWS account"
```
mikemcquaid: Just leave this out for now; we're going to bulk-add them later, thanks.

mikemcquaid: Oh, you moved it up, never mind!

mikemcquaid:
Sorry, I should have been a bit more diligent about pulling the original PR, I messed up, sorry. Could we move this into a tap you maintain instead? I shouldn't have pulled this as it's a bit too niche for Homebrew core.

Luzifer:
Hmm how will this affect users (if any) already having this installed from the core? (Can you see how many users pulled the bottles?)

Luzifer:
But as I only tweeted about that this should be okay. In that way I even can do a faster release-iteration… Shall I change this PR to remove the `awsenv.rb`?

mikemcquaid:
If you wouldn't mind, yeh. If you add it to `tap_migrations.rb` it'll automatically migrate people to your tap.

mikemcquaid:
Thanks for understanding :heart:

Luzifer:
@mikemcquaid PTAL at the migration. It *should* work from my tap… (Even though I'm fighting against the travis testing… ;))

Luzifer:
@mikemcquaid Ping. (Would be great if this get merged soon as there is a bugfix from `0.4.2` to `0.4.3` which re-enables on of the core-functions of `awsenv`)

tdsmith:
Pushed; thanks!

PR #40181, opened 2015-05-29
weechat 1.2

Porkepix:


bfontaine:
Thank you for your contribution to Homebrew, @Porkepix! For the record, the preferred checksum is now `sha256`.

Porkepix:
This is mainly because SHA-1 was the one provided on the download page: https://weechat.org/download/stable/
I thought it was better to use this one in this situation. But I can provide the `sha256` one by calculating it myself if you prefer so.

bfontaine:
No worries, I merged your pull request as is :wink: . In the future, you can use `brew fetch <formula>` (e.g. `brew fetch weechat`), which will download the archive and print its `sha1` and `sha256`.

PR #40185, opened 2015-05-29
sourceforge.jp is now osdn.jp

vszakats:
Plus related updates.

### DIFF
--- a/Library/Formula/julius.rb
+++ b/Library/Formula/julius.rb
@@ -1,7 +1,8 @@
 class Julius < Formula
-  homepage "http://julius.sourceforge.jp"
-  url "http://sourceforge.jp/frs/redir.php?m=iij&f=%2Fjulius%2F60273%2Fjulius-4.3.1.tar.gz"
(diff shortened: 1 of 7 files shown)


vszakats:
Error seems unrelated to this update:
```
==> brew audit mecab-unidic						 FAILED
mecab-unidic:
```

vszakats:
`mecab-unidic` build will create these four binary data files:
```
char.bin
matrix.bin
sys.dic
unk.dic
```

xu-cheng:
Looks like `mecab-unidic` will attempt to write file inside `mecab`'s Cellar.
```
May 30 17:09:59 kernel[0]: Sandbox: mkdir(98661) deny file-write-create /Users/xucheng/Coding/Github/homebrew/Cellar/mecab/0.996/lib/mecab
```

This is against Homebrew's policy. @vszakats Can you make `mecab-unidic` as a part of `mecab` formula. Otherwise we should boneyard it.

vszakats:
First I split `mecab-unidic` from this pull, so this can pass and be accepted.

mikemcquaid:
Thanks @vszakats!

PR #40186, opened 2015-05-29
binwalk: fix library lookup and installation

UniqMartin:
Replaces hard-coded path to Homebrew cellar, that is used for locating and loading dynamic libraries, with the actual Homebrew cellar. This allows using `binwalk` with non-standard Homebrew prefixes instead of being broken because mandatory dependencies are not found.

Furthermore, fixes the setup script such that it installs the compiled `libcompress42.dylib` and `libtinfl.dylib` instead of missing them. This avoids warnings at runtime, that are non-fatal, but result in degraded functionality (because of the missing libraries).

*(As far as I can tell, these fixes do not apply to HEAD, since HEAD and stable have diverged significantly.)*

### DIFF
--- a/Library/Formula/binwalk.rb
+++ b/Library/Formula/binwalk.rb
@@ -4,6 +4,18 @@ class Binwalk < Formula
   stable do
     url "https://github.com/devttys0/binwalk/archive/v2.0.1.tar.gz"
     sha256 "90ee8426d71e91b62dfe4a1446c457bc7835b475b28717859e275a0494403959"
+
+    # Fixes OS-X-specific issues no longer relevant in HEAD:
+    #
+    # * Fixes OS X bug in 'setup.py':
+    #   * See <https://github.com/devttys0/binwalk/commit/8278229cae4c2c354ffc5bfc3bcef0fd1d9bf2b3>
+    #     and <https://github.com/devttys0/binwalk/commit/76ff729a19552e6e58505ca64b1d4ce2325e7ac0>.
+    # * Fixes library lookup for non-standard Homebrew installations:
+    #   * See upstream issue <https://github.com/devttys0/binwalk/issues/130>
+    #     for the details.
+    #   * The fix is Homebrew-specific as it uses HOMEBREW_PREFIX, that is
+    #     implicitly replace with the actual Homebrew prefix, in the patch.
+    patch :DATA
   end
 
   bottle do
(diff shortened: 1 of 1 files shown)


Library/Formula/binwalk.rb
```diff
@@ -66,6 +66,13 @@ def install
       bin.install Dir["#{libexec}/bin/*"]
       bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
     else
+      # Fix hard-coded path to Homebrew cellar.
+      inreplace "src/binwalk/core/C.py", "('/usr/local/Cellar/", "('#{HOMEBREW_CELLAR}/"
```
mikemcquaid: Is this hardcoded in their source code?

Library/Formula/binwalk.rb
```diff
@@ -66,6 +66,13 @@ def install
...
+
+      # Fix hard-coded extension of dynamic libraries. Otherwise both libraries
+      # "libcompress42.dylib" and "libtinfl.dylib" will not be installed.
+      inreplace "setup.py", "\"*.so\"", "\"*.dylib\""
```
mikemcquaid: Can you please submit these patches to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Library/Formula/binwalk.rb
bfontaine: [Yes](https://github.com/devttys0/binwalk/blob/master/src/binwalk/core/C.py#L131)

UniqMartin:
Upon further investigation I see that my fix is more of a dirty workaround. The confusion about the `so`/`dylib` extension is no longer relevant in HEAD as all compiled code has been eliminated, but the hard-coded Homebrew cellar (among other search paths) is still an issue.

I'll get in contact with upstream to find out about their release policy and to discuss how to best solve these issues. Hopefully, I can report back soon.

UniqMartin:
Upstream no longer depends on any dynamic libraries (either bundled or external) in their master branch, thus both issues addressed by me are no longer relevant. According to upstream (devttys0/binwalk#130) a new stable release is forthcoming.

I could still provide a patch for the current stable 2.0.1, if that makes sense. (This pull request is only a partial fix and shouldn't be applied yet.)

mikemcquaid:
> I could still provide a patch for the current stable 2.0.1, if that makes sense. (This pull request is only a partial fix and shouldn't be applied yet.)

Yeh, that'd be good. Also, ideally use upstream commits with `.patch` appended to them where they apply cleanly instead of inlining. Thanks!

UniqMartin:
My branch now has both fixes for the stable version as an embedded patch and hopefully provides enough context in the comment before `patch :DATA`. (I pushed before you responded.)

Unfortunately, the two referenced upstream commits do not apply cleanly because of additional unrelated changes. Thus, I don't see a way to use them directly. (And the second part of the patch is specific to Homebrew anyway.)

mikemcquaid:
No worries, thanks. Is this now ready to be merged, from your perspective?

UniqMartin:
Yes, it is. :)

mikemcquaid:

UniqMartin:
Glad, that I could contribute. Thanks for all the guidance and the feedback!

PR #40192, opened 2015-05-29
add --devel option on scala

AtkinsChang:
add --devel option for milestone version of scala


bfontaine:
Thanks for your contribution to Homebrew, @AtkinsChang!

For the record the preferred commit message is “scala: add --devel option”.

tdsmith:
> For the record the preferred commit message is “scala: add --devel option”.

FWIW, here, it's `scala 2.12.0-M1 (devel)`.

Please use double quotes and sha256 hashes for new contributions.

Issue #40151, opened 2015-05-28
loading same name formulae from different taps should be possible


xu-cheng:
Since we load all formulae class inside the same namespace, we cannot load formulae with same class name. For example.
```
$ head -4 /tmp/tree.rb
require 'formula'

class Tree < Formula
  version "1.8.0"

$ brew pry
==> Interactive Homebrew Shell
Example commands available with: brew pry --examples
[1] brew(main)> f1 = Formulary.factory("tree")
=> #<Formula tree (stable) /usr/local/Library/Formula/tree.rb>
[2] brew(main)> f2 = Formulary.factory("/tmp/tree.rb")
=> #<Formula tree (stable) /tmp/tree.rb>
[3] brew(main)> f1.version
=> #<Version::FromURL:0x007f8f943029f8 @version="1.7.0">
[4] brew(main)> f2.version
=> #<Version::FromURL:0x007f8f943029f8 @version="1.7.0">
```

The solution would be loading them in separate namespace. Instead of making `Tree` class inside `Formulary::Formulae`, we could load them inside `Formulary::Formulae::NS#{sha1_of_path[0..7]}`.

Another advantage is that it will resolve the custom `Requirement` name conflicts problem.

ping @jacknagel

jacknagel:
https://github.com/Homebrew/homebrew/compare/master...jacknagel:namespace

xu-cheng:
:+1:

jacknagel:
Simplified the implementation (caching formula classes directly rather than the parent modules).

PR #40203, opened 2015-05-30
Isolate formula class namespaces

jacknagel:
Addresses #40151


xu-cheng:
:+1:

mikemcquaid:
I guess it's an implementation detail so we don't need to care but: what's used to create the `Module`; the full path to the formula/tap?

mikemcquaid:
Otherwise: LGTM

jacknagel:
In this updated implementation, the formula file is `module_eval`'d inside the module and then the formula class is cached directly. The full path is used as the key, so we only need to do one hash lookup at runtime (as opposed to a normal hash lookup plus a `const_get`, which is also a hash lookup, in the earlier implementation).

PR #40211, opened 2015-05-30
mecab: merge mecab-unidic as an optional resource

vszakats:
It's my first attempt to tackle a resource and it even has an optional twist and a caveat. Ergo: This can be completely wrong. Please be patient.

Other updates:
- mecab-unidic: update url and checksum
- mecab: add desc to pass strict audit


Library/Formula/mecab.rb
```diff
@@ -12,7 +13,31 @@ class Mecab < Formula
...
+                                "--prefix=#{prefix}"
+          system "make", "install"
+        end
+      end
+
+      def caveats; <<-EOS.undent
+        If you want to use UniDic, please rewrite "dicdir".
+           #{opt_prefix}/etc/mecabrc
+        EOS
+      end
```
tdsmith: Nesting a caveats method under the install method shouldn't do anything.

Library/Formula/mecab.rb
```diff
@@ -12,7 +13,31 @@ class Mecab < Formula
     sha1 "9747369cd4c0aa246e6a973c4f2e5652e174bae8" => :mountain_lion
   end
 
+  option "with-unidic", "UniDicとは"
```
tdsmith: 英語で書いたらほうがいいと思う、おねがいします

Library/Formula/mecab.rb
tdsmith: Can the formula take care of the configuration that's required?

tdsmith:
Adding dictionaries to the mecab formula makes sense to me.

Library/Formula/mecab.rb
vszakats: This dictionary has the required logic:
https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mecab-ko-dic.rb#L18-L20

But, what to do when multiple dictionaries are selected?

vszakats:
Abandoning this along with https://github.com/Homebrew/homebrew/pull/40238.

Issue #40152, opened 2015-05-28
Find a way to detect formulae with name conflicts inside official Taps


xu-cheng:
Since #39524 is merged, formulae with name conflicts inside official Taps would be hard to detect. We need to find a way to detect that. Maybe adding into `brew audit` or `brew readall`. The trick part is how to suppress such error during the formula migration? Any thoughts @Homebrew/owners @bfontaine @DomT4.

mikemcquaid:
I'm thinking we add a mode to `brew audit` that does online queries (and is off by default) that could work for stuff like this.

PR #40216, opened 2015-05-30
audit: audit formulae with the same name

xu-cheng:
Address #40152

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -2,6 +2,8 @@ require "formula"
 require "utils"
 require "extend/ENV"
 require "formula_cellar_checks"
+require "official_taps"
+require "cmd/search"
 
 module Homebrew
   def audit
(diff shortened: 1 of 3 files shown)


Library/Homebrew/cmd/audit.rb
```diff
@@ -161,6 +164,42 @@ def audit_class
...
+        missing_taps.each { |tap| Homebrew.install_tap "homebrew", tap }
+      end
+      same_name_formulae = Formula.full_names.group_by { |n| n.split("/").last }.fetch(name)
+      if same_name_formulae.size > 1
+        problem "Formula name is conflicted with #{(same_name_formulae - [full_name]).join ", "}"
+      end
+    ensure
+      nostdoutstderr do
+        missing_taps.each { |tap| Homebrew.uninstall_tap "homebrew", tap }
+      end
```
mikemcquaid: We definitely don't need to install each of these taps when we can use `brew search` instead.

xu-cheng: Updated to use search.

xu-cheng:
I wonder is there any way to suppress this audit rule in the event of formula migration?

mikemcquaid:
@xu-cheng Hmm. I think it's probably just fine to ignore it in that case? I'm open to any other ideas, though. Approach makes sense to me :+1:

mikemcquaid:
Oh, last thing would be checking this fails nicely if you don't have an internet connection.

xu-cheng:
> Oh, last thing would be checking this fails nicely if you don't have an internet connection.

`Homebrew.search_tap` already handled this. Currently such error would be just ignored silently.

mikemcquaid:
:+1:

PR #40217, opened 2015-05-30
mongodb: fix openssl build

### LINKED ISSUE #40136: mongodb --with-openssl failed to build on 10.10.3
https://gist.github.com/cdorer/4c31ed2f28e16bd410ad

I also *brew cleanup; brew uninstall openssl; brew uninstall go*, and then I reinstalled mongodb --with-openssl with the same exact errors.

mongodb installed fine without the --with-openssl flag

DomT4:
Checked this works locally, It does. Compiles fine.

Closes #40136.

### DIFF
--- a/Library/Formula/mongodb.rb
+++ b/Library/Formula/mongodb.rb
@@ -3,10 +3,10 @@ require "language/go"
 class Mongodb < Formula
   homepage "https://www.mongodb.org/"
 
-  depends_on "go" => :build
   stable do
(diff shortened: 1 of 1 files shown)


Library/Formula/mongodb.rb
```diff
@@ -3,10 +3,10 @@
 class Mongodb < Formula
   homepage "https://www.mongodb.org/"
 
-  depends_on "go" => :build
```
DomT4: This is duplicated further below, also in the main body, hence the removal here.

mikemcquaid:
:zap: @DomT4

PR #40223, opened 2015-05-30
yafc 1.3.6

bfontaine:
@mikemcquaid Is it ok if I replace the patch with an `inreplace` call given that it’s OSX-specific and upstream only targets Linux (see their homepage)?

### DIFF
--- a/Library/Formula/yafc.rb
+++ b/Library/Formula/yafc.rb
@@ -1,14 +1,17 @@
-require 'formula'
-
 class Yafc < Formula
-  homepage 'http://www.yafc-ftp.com/'
-  url 'http://www.yafc-ftp.com/upload/yafc-1.3.2.tar.xz'
-  sha1 'a4cd5518b84cd40c4503f2e022ba946a8bd5484a'
+  homepage "http://www.yafc-ftp.com/"
+  url "http://www.yafc-ftp.com/upload/yafc-1.3.6.tar.xz"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> @mikemcquaid Is it ok if I replace the patch with an inreplace call given that it’s OSX-specific and upstream only targets Linux (see their homepage)?

I think that's preferable to the patch but it'd be really :heart: if you could submit a PR upstream?

bfontaine:
Yep, just did it.

Library/Formula/yafc.rb
```diff
@@ -1,14 +1,19 @@
...
+
+  depends_on "readline"
+  depends_on "libssh" => :recommended
+  depends_on "pkg-config" => :build
+
+  # The bash-completion script assumes GNU sed and its 'r' option for extended
+  # regular expressions. OSX's sed calls this 'E' (they're not entirely
+  # compatible, but enough for this use case).
+  patch do
+    url "https://patch-diff.githubusercontent.com/raw/sebastinas/yafc/pull/54.diff"
```
mikemcquaid: Mind using the pull URL? Makes jumping to it a bit easier.

bfontaine: I appended `.diff` to the pull URL and it redirected me to this URL.

mikemcquaid:
> Update: upstream merged it.

:metal:

Issue #39903, opened 2015-05-19
brew rm $(brew deps pygtk) fail


davidak:
i am following this guide: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Common-Issues.md#python-segmentation-fault-11-on-import-some_python_module

```
brew rm $(brew deps pygtk)
Error: No such keg: /usr/local/Cellar/git
```

my git is not installed via brew.
maybe this command should ignore not installed packages.

i had to delete some packages from the list:

```
brew rm atk cairo fontconfig freetype gdk-pixbuf gettext glib gobject-introspection gtk+ harfbuzz icu4c jpeg libffi libpng libtiff pango pixman pkg-config py2cairo pygobject python
```

DomT4:
Should work if you do something like ```brew rm `brew deps pygtk --installed` ```

davidak:
that shows all dependencies of all installed packages. `pygtk`has no effect

```
$ brew deps --installed
atk: glib gobject-introspection pkg-config xz
brew-cask: git ruby
cairo: fontconfig freetype glib libpng pixman pkg-config xz
fontconfig: freetype pkg-config
freetype: libpng
gdbm:
gdk-pixbuf: glib gobject-introspection jpeg libpng libtiff pkg-config xz
gettext: xz
glib: gettext libffi pkg-config xz
gobject-introspection: git glib libffi pkg-config xz
gtk+: atk gdk-pixbuf gobject-introspection jasper pango pkg-config xz
harfbuzz: cairo freetype glib icu4c pkg-config
htop-osx: autoconf automake libtool
icu4c:
jpeg:
libffi:
libpng: xz
libtiff: jpeg
openssl: makedepend
pango: cairo fontconfig glib gobject-introspection harfbuzz pkg-config xz
pixman: pkg-config
pkg-config:
py2cairo: cairo pkg-config
pygobject: glib pkg-config python
pygtk: atk glib gtk+ libglade pkg-config py2cairo pygobject
python: gdbm homebrew/dupes/tcl-tk openssl pkg-config readline sqlite
readline:
sqlite: icu4c readline
vala: gettext glib pkg-config xz
```

DomT4:
Fair point. It's an ugly workaround, and I'm surprised we don't support listing only the installed deps, but `brew deps --installed | grep -e pygtk: | /usr/bin/awk '{sub(/pygtk: /, ""); print $0}' | xargs brew remove` would work.

[...]

PR #40233, opened 2015-05-31
brew deps --installed support for formulae

bfontaine:
Current situation:

```
# shows all foo's dependencies (both installed and not)
$ brew deps foo

# shows all installed formulae and their dependencies
$ brew deps --installed

# shows all installed formulae and their dependencies (again)
$ brew deps --installed foo
```

This PR fixes the last case by making `brew deps --installed foo` show only the installed dependencies for `foo`. It allows people to remove all the dependencies for a formula using `brew rm $(brew deps --installed <formula>)`.

See #39903.

### DIFF
--- a/Library/Homebrew/cmd/deps.rb
+++ b/Library/Homebrew/cmd/deps.rb
@@ -14,16 +14,17 @@ module Homebrew
 
     if mode.installed? && mode.tree?
       puts_deps_tree Formula.installed
-    elsif mode.installed?
-      puts_deps Formula.installed
     elsif mode.all?
       puts_deps Formula
     elsif mode.tree?
       raise FormulaUnspecifiedError if ARGV.named.empty?
       puts_deps_tree ARGV.formulae
+    elsif ARGV.named.empty?
+      raise FormulaUnspecifiedError unless mode.installed?
+      puts_deps Formula.installed
     else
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/deps.rb
```diff
@@ -14,16 +14,19 @@ def deps
...
       puts_deps Formula
     elsif mode.tree?
       raise FormulaUnspecifiedError if ARGV.named.empty?
       puts_deps_tree ARGV.formulae
     else
-      raise FormulaUnspecifiedError if ARGV.named.empty?
+      if ARGV.named.empty?
+        raise FormulaUnspecifiedError unless mode.installed?
+        puts_deps Formula.installed
+        return
```
xu-cheng: Maybe
```ruby
elsif ARGV.named.empty?
  ...
else
  ...
end
```
instead of
```ruby
else
  if ARGV.named.empty?
    ...
    return
  end
  ...
end
```

bfontaine: Right, fixed.

bfontaine:
cc @Homebrew/owners @DomT4

DomT4:
LGTM :+1: :star2:

mikemcquaid:
:+1:

xu-cheng:
:+1:

PR #40238, opened 2015-05-31
mecab formulae modernization and one bump

vszakats:
It is meant to be a preparation step before converting/merging dictionary formulae into mecab/mecab-ko resources. Some formulae may fail the bot test till the merge is completed.

Ref: https://github.com/Homebrew/homebrew/pull/40211

### DIFF
--- a/Library/Formula/mecab-ipadic.rb
+++ b/Library/Formula/mecab-ipadic.rb
@@ -1,17 +1,16 @@
-require 'formula'
-
 class MecabIpadic < Formula
-  homepage 'http://mecab.googlecode.com/svn/trunk/mecab/doc/index.html'
-  url 'https://downloads.sourceforge.net/project/mecab/mecab-ipadic/2.7.0-20070801/mecab-ipadic-2.7.0-20070801.tar.gz'
-  sha1 '0d9d021853ba4bb4adfa782ea450e55bfe1a229b'
+  desc "MeCab dictionary 'ipadic'"
+  homepage "https://mecab.googlecode.com/svn/trunk/mecab/doc/index.html"
+  url "https://downloads.sourceforge.net/project/mecab/mecab-ipadic/2.7.0-20070801/mecab-ipadic-2.7.0-20070801.tar.gz"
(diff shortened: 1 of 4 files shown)


vszakats:
Two mecab dictionaries failed as expected:
```
mecab-ipadic:
```

tdsmith:
I'm not sure I understand what you're shooting for here. It's a problem that the installations are empty, right? How does it help you for Homebrew to pull this?

vszakats:
I prefer smaller steps, so that I don't have to keep that many balls in the air. These updates fix some existing audit issues, while not causing any additional problems over those that existed for a long time already, so they should be safe to pull. See also my reasoning/question here: https://github.com/Homebrew/homebrew/pull/40211#commitcomment-11439477

As a next step these dicts can be more easily merged into the MeCab formulae, that's all.

tdsmith:
I'd rather not pull things that don't work; can these be fixed? (I also don't really think this work will help you add them to `mecab` as resources.)

vszakats:
:( okay.

tdsmith:
Do these dictionaries work at all? Should we boneyard them?

vszakats:
I haven't tested them, but I see no reason why they wouldn't work once their respective MeCab[-ko] projects are configured to use them. They don't pass the bot because they don't publish their files like formulae should.

I don't think we should boneyard them. The reason I started messing with these is because @makigumo  in the referred thread [wrote](https://github.com/Homebrew/homebrew/pull/40211#commitcomment-11439390) that at least two of these are useful dictionaries.

One disclaimer: I haven't tested if the version updated dictionary `mecab-ko-dic` works (1.6.1 -> 2.0.0), but it's a data file and its host project didn't change since, so chances are high that it does.

tdsmith:
> They don't pass the bot because they don't publish their files like formulae should.

What do they do instead?

vszakats:
They build their data files and copy them inside the directory structure of their respective host `mecab[-ko]` project.

vszakats:
Into `<mecab-cellar>/lib/mecab/dic/ipadic` in case of `mecab-ipadic`.

PR #40240, opened 2015-05-31
MacBrush (0.12) new formula

nils-tekampe:
Please refer to http://nils-tekampe.github.io/MacBrush for further
details.

### DIFF
--- a/Library/Formula/macbrush.rb
+++ b/Library/Formula/macbrush.rb
@@ -1,0 +1,20 @@
+class Macbrush < Formula
+  desc "Keeping OSX directories clear of temporary files"
+  homepage "http://nils-tekampe.github.io/MacBrush"
+  url "https://github.com/nils-tekampe/MacBrush/archive/0.12.tar.gz"
+  sha256 "ed08874d9c2059a81356419d29d9ff603b4810baea7742a2132dbdfaee064414"
+
+  depends_on :xcode => :build
+
+  def install
+    cd "macbrush"
+    xcodebuild  "clean"
+    xcodebuild 
+    bin.install "./build/Release"
+  end
+  test do
+  mkdir "test"
+  touch "./test/.DS_Store"
+  macbrush "--verbose --skip-observation ./test"
+  end
+end


Library/Formula/macbrush.rb
```diff
@@ -0,0 +1,20 @@
+class Macbrush < Formula
+  desc "Keeping OSX directories clear of temporary files"
+  homepage "http://nils-tekampe.github.io/MacBrush"
```
DomT4: `https://nils-tekampe.github.io/macbrush/`. Should be both `https` and lowercase, otherwise it 404s.

Library/Formula/macbrush.rb
```diff
@@ -0,0 +1,20 @@
...
+  sha256 "ed08874d9c2059a81356419d29d9ff603b4810baea7742a2132dbdfaee064414"
+
+  depends_on :xcode => :build
+
+  def install
+    cd "macbrush"
+    xcodebuild  "clean"
+    xcodebuild 
+    bin.install "./build/Release"
+  end
```
DomT4: Should be a gap between these two blocks.

Library/Formula/macbrush.rb
```diff
@@ -0,0 +1,20 @@
...
+  test do
+  mkdir "test"
```
DomT4: You'll already be in a temporary testing folder here. You're fine not making another.

DomT4:
I'm not sure this meets our [submission guidelines](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Acceptable-Formulae.md#niche-or-self-submitted-stuff) at this time sadly.

It's very new, and it's a self-submission, and described on the upstream website as "Alpha". It may be better to ship it in your own tap for a while and potentially resubmit it at a future date.

nils-tekampe:
Thanks for the comments! I'll fix the three technical comments and will do some further testing to ensure that the tool is stable. Will re-submit then.

mikemcquaid:
Thanks! Just for your reference I'd like to see about 50 GitHub Stars and/or forks and/or external issues and/or PRs before something like this gets included into core.

nils-tekampe:
Wow, that's quite a bit. I'll see whether such a special tool can reach  it....

mikemcquaid:
@nils-tekampe Probably better to just run your own tap and let one of your users submit (without prompting) if they are interested in future.

nils-tekampe:
Sounds like the best plan....

PR #40242, opened 2015-05-31
virtuoso 7.2.0

dunn:
Last week @DomT4 said there was a moratorium on new `desc`s; is that still in effect or should I go ahead and add one?

PR #40247, opened 2015-05-31
cookbook: add an explicit version if necessary

bfontaine:
I’ve notice two cases in the past few days where someone bumped a version number and Homebrew’s guessing failed. See #40242 for example. I’m not sure this is the right place to put that, I’m open to suggestions.


DomT4:
We should look at adding an audit check to see if the version matches the expected version (if we don't already). Possibly a tricky thing to ensure we don't hit false positives on though.

tdsmith:
To implement an audit check we would have to have a version-guesser that's more accurate than our version-guesser (and shouldn't we just improve our version-guesser?).

DomT4:
Heh, fair point. I wasn't thinking particularly complicated, just along the lines of *"URL seems to have 3 numbers in the tarball name, but installed formula only has one number in the version, that's not right"*. But yeah, improving the version detection is the more desirable answer.

mikemcquaid:
:+1:

Issue #38047, opened 2015-03-25
macvim failed to build on 10.6.8


axtens:
==> Upgrading macvim
==> Downloading https://github.com/macvim-dev/macvim/archive/snapshot-74.tar.gz
Already downloaded: /Library/Caches/Homebrew/macvim-7.4-74.tar.gz
==> ./configure --with-features=huge --enable-multibyte --with-macarchs=x86_64 --enable-perlinterp --enable-rubyinterp --enable-tclinterp --with-tlib
==> make
if_ruby.c:706: error: 'rb_encoding' undeclared (first use in this function)
if_ruby.c:706: error: 'enc' undeclared (first use in this function)
make[1]: *** [objects/if_ruby.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [first] Error 2

gist: https://gist.github.com/axtens/e49c260bc97b6b4bb2a2

mistydemeo:
Could you post the output of `brew gist-logs macvim`, please? Thanks!

axtens:
https://gist.github.com/09e838289f9d9d4fd8e9

Done!

mistydemeo:
Looks like Macvim isn't compatible with the system Ruby in 10.6.

axtens:
bash-4.3$ brew install ruby
Warning: ruby-2.2.1 already installed

Which version is needed?

chdiza:
If memory serves, MacVim can't be made to link to a non-system ruby (because Vim can't).  If that's right, then @axtens could work around this by commenting out the `--enable-rubyinterp` line in the MacVim formula.

@mistydemeo I could swear that MacVim *used to be* compatible with Snow Leo's ruby.  Perhaps they mucked around with ruby detection for Linux or Windows and nobody noticed that it broke things for OSX?

jacknagel:
I would `brew unlink ruby` and then try again.

sammcj:
Reproducible on Yosemite with standard system ruby `2.0.0p481 (2014-05-08 revision 45883) [universal.x86_64-darwin14]`

```
brew install macvim --with-lua --override-system-vim
==> Downloading https://github.com/macvim-dev/macvim/archive/snapshot-74.tar.gz
Already downloaded: /Library/Caches/Homebrew/macvim-7.4-74.tar.gz
==> ./configure --with-features=huge --enable-multibyte --with-macarchs=x86_64 --enable-perlinterp --en
==> make
[...]

PR #40255, opened 2015-06-01
macvim: disable ruby integration on 10.6

rhendric:
Per discussion in #38047

### DIFF
--- a/Library/Formula/macvim.rb
+++ b/Library/Formula/macvim.rb
@@ -37,12 +37,16 @@ class Macvim < Formula
       --enable-multibyte
       --with-macarchs=#{MacOS.preferred_arch}
       --enable-perlinterp
-      --enable-rubyinterp
       --enable-tclinterp
       --with-tlib=ncurses
       --with-compiledby=Homebrew
       --with-local-dir=#{HOMEBREW_PREFIX}
     ]
+
+    # MacVim uses the system ruby, but fails to compile against the ruby
+    # provided with Snow Leopard.
+    # See https://github.com/Homebrew/homebrew/issues/38047
+    args << "--enable-rubyinterp" unless MacOS.version <= :snow_leopard
 
     args << "--enable-cscope" if build.with? "cscope"
 


DomT4:
It's probably worth adding a note above the line to the issue referenced so someone doesn't attempt to reverse it later.

rhendric:
Good idea, done.

jacknagel:
Is this correct? I thought the problem was a mismatch caused by a non-system ruby being in the PATH, and macvim's configure subsequently making invalid assumptions about the system ruby?

rhendric:
I tried brewing macvim after a `brew unlink ruby` and got the same compilation error as before, so I don't think that's it.

jacknagel:
As far as I can tell, nobody ever posted full logs from a failed build *without* a non-system ruby installed. I'd like to see that before pulling something like this.

Note that for vim, we also need the file `src/auto/config.log` from the failed build, to get the actual configure log.

rhendric:
`brew gist-logs`: https://gist.github.com/anonymous/7b7ba534a056ec21203d
`src/auto/config.log`: https://gist.github.com/rhendric/bd05f7345fc581cd075c

jacknagel:
Can you try `brew install vim --env=super` and gist the resulting logs?

rhendric:
With or without Homebrew's ruby linked?

rhendric:
Oh, and did you mean `vim` or `macvim`?

jacknagel:
Sorry, `macvim`.

wrt ruby, the same conditions as however the above logs were generated.

rhendric:
`brew gist-logs`: https://gist.github.com/16cefe6952bffdc3df95
`src/auto/config.log`: https://gist.github.com/ce63e002fad7a25a97b6

jacknagel:
Well, configure is detecting the header directory as:

```
/System/Library/Frameworks/Ruby.framework/Versions/1.8/usr/lib/ruby/1.8/universal-darwin10.0
```

This is basically how configure decides this:

```sh
ruby -r mkmf -e "print RbConfig::CONFIG['rubyhdirdir'] || RbConfig::CONFIG['archdir']"
```

It seems that this may be falling back to `archdir` which is not correct. `rubyhdrdir` must not be defined.

On 10.10 this returns

```
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/include/ruby-2.0.0
```

which is a much more reasonable thing.

rhendric:
Hmm, okay.

So, you're right: `/usr/bin/ruby -r mkmf -e "print RbConfig::CONFIG['rubyhdrdir']"` returns `nil` for me. Googling indicates that `rubyhdrdir` was added in Ruby 1.9, and Snow Leopard ships with 1.8.7, so this makes sense.

Inspired by that, I tried a few things and this is the only combination that gets `--enable-rubyinterp` to work for me so far:
* Adding a `depends_on :ruby => "1.9"`
* `brew unlink ruby` before compiling
* Running in std env (as currently specified in the formula)

I tried commenting out the `env :std` line in the hopes that superenv would protect me from a linked ruby, but that failed. (It's looking for a nonexistent `/Developer/Library/Xcode/PrivatePlugIns/Xcode3Core.ideplugin/Contents/SharedSupport/Developer/Library/Xcode/Plug-ins/LLVM GCC 4.2.xcplugin/Contents/Resources/llvm-gcc` file... good grief.)

Thoughts on next things to try? Unlinking ruby is annoying; if it weren't for that I'd just slap a `depends_on :ruby => ["1.9", :recommended] if MacOS.version <= :snow_leopard` in there and be done.

jacknagel:
I think probably any breakage in configuring vim against ruby 1.8 is unintentional, but a good first step would be to confirm that with upstream (vim, as I assume that macvim is not modifying this code) and then get a fix in place on that side.

bfontaine:
Anything new on this?

rhendric:
Upstream has decided to stop supporting compiling on 10.6, so I suppose this is dead. :crying_cat_face:

PR #40258, opened 2015-06-01
Allegro 5 Formula

SiegeLord:
With this change I propose to move the Allegro 5 formula from homebrew-versions tap to the core. Allegro (version 4) used to be in the core tap, but it was moved out when Allegro 5 was released (see discussion in https://github.com/Homebrew/homebrew/pull/4753). I propose moving it back with the following justifications:

* A lot of interesting OSX-related development is currently happening on the development branch, but the homebrew-versions disallows a devel block.
* The [current](https://github.com/Homebrew/homebrew-versions/blob/master/allegro5.rb) homebrew-versions formula inappropriately (given my understanding of that tap's rules) includes a head block, which while useful, would be a better fit if it were in the core tap.
* The older version of Allegro installed by the [allegro4](https://github.com/Homebrew/homebrew-versions/blob/master/allegro4.rb) formula is no longer actively developed, and does not build on recent OSX versions. In practice, Allegro 5 version is the only one that is of consequence at this time.
* I'm proposing to keep the formula name as is to simplify the user experience, and make it clear which Allegro is being installed. Debian also calls this package allegro5. A precedent in homebrew is the sdl2 formula (SDL 1 has the same relationship to SDL 2 as Allegro 4 does to Allegro 5: similar in spirit, but backwards incompatible).

The homebrew-versions PR: https://github.com/Homebrew/homebrew-versions/pull/847

In addition to moving, the following changes were made:

* Change the homepage to http://liballeg.org. That is the official website, http://allegro.cc is just where the forums are hosted.
* head and devel sections were added with a new dependency
* The dependencies were made recommended rather than optional, as they are required for a big bulk of the library's features.
* Renamed the formula to allegro, as per @tdsmith's preference.


tdsmith:
I would like this to be "allegro" and not "allegro5" so that this formula continues to represent the newest available version of allegro if an allegro 6 is released.

Please go ahead and add the devel block.

DomT4:
> (given my understanding of that tap's rules) 

In certain cases Versions formulae are allowed `head`s. This is one of them based on it being tailored to just one branch that lines up with the existing version tree.

I'm :+1: on moving this over though. If Tim et al. are cool with this moving, please tee up a PR in Versions to remove the formula there as well. Thanks!

SiegeLord:
Sounds good. I've made the PR to the other repository, merged in the remaining changes I had in mind and renamed it to just 'allegro'.

mikemcquaid:

For future reference the preferred commit message format for new formulae is `allegro 5.0.11 (new formula)`.

PR #40264, opened 2015-06-01
ssht 0.3 (new formula)

brejoc:
Added formula to ssht version 0.3 to the Homebrew repo. ssht is a
shortcut right into a tmux session on a remote computer.

### DIFF
--- a/Library/Formula/ssht.rb
+++ b/Library/Formula/ssht.rb
@@ -1,0 +1,13 @@
+require 'formula'
+
+class Ssht < Formula
+  homepage 'https://github.com/brejoc/ssht/'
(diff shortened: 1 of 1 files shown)


bfontaine:
Unfortunately I don’t think it meets [our guidelines](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Acceptable-Formulae.md#niche-or-self-submitted-stuff) as it’s a self-submission of a relatively new program. It may be better to include it in your own tap for a while and resubmit it when it’ll be more popular (e.g. ≥50 stars).

mikemcquaid:
Passing on this for now as mentioned above. Sorry!

PR #40274, opened 2015-06-01
xsd 4.0.0 (new formula)

Asuranceturix:


Library/Formula/xsd.rb
```diff
@@ -0,0 +1,17 @@
+class Xsd < Formula
+  desc "XML Data Binding for C++"
+  homepage "http://www.codesynthesis.com/products/xsd/"
+  url "http://www.codesynthesis.com/download/xsd/4.0/macosx/i686/xsd-4.0.0-i686-macosx.tar.bz2"
+  sha256 "56571e446ae49935611cbf46484a50c1a0afd3bb7987be728b8dcdd3de34b897"
+
+  depends_on 'xerces-c'
```
bfontaine: Please use double quotes everywhere.

Library/Formula/xsd.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def install
+    bin.install "bin/xsd"
+    include.install "libxsd/xsd"
+  end
+
+  test do
+    system "#{bin}/xsd version"
```
bfontaine: Could you add a test that does something more substantial than e.g. `--version` or `--help`? Thanks!

Also, use `system "foo", "bar"` instead of `system "foo bar"`.

Asuranceturix:
Well, it's a developer tool. The other thing that it does is parse a schema and generate cxx+hxx files to parse XML files that conform to the schema. How do you suggest the test should be?

mikemcquaid:
> Well, it's a developer tool. The other thing that it does is parse a schema and generate cxx+hxx files to parse XML files that conform to the schema. How do you suggest the test should be?

Write out a schema, try to parse it, verify the files are generated.

Asuranceturix:
Ok, where (in which directory) is the test run? How do I put a schema there? How do I test the files exist from the formula? (I know nothing about Ruby)

bfontaine:
> Ok, where (in which directory) is the test run?

The test runs in a temporary directory.

> How do I put a schema there?

You can use the following code to write a file called `yourfilename.xsd`:

```ruby
(testpath/"yourfilename.xsd").write <<-EOS.undent
   here
   you can write
   anything you
   want
EOS
```

> How do I test the files exist from the formula?

You can test that a file exist with:

```ruby
assert File.exist? "thefile"
```

I don’t know how the `xsd` tool works, but let’s say it takes a `foo.xsd` file and generates `foo.cpp` and `foo.hpp`. Here is how a test could look like:


```ruby
test do
  path = (testpath/"test.xsd")
  path.write <<-EOS.undent
    <the>
      <schema>
        <goes>
          <here></here>
        </goes>
      <schema>
    </the>
  EOS

  system "#{bin}/xsd", path
  assert File.exist? "foo.cpp"
  assert File.exist? "foo.hpp"
end
```

You could even go further and compile the files to parse a simple XML file ;)
Don’t hesitate to comment here if you have any question.

Asuranceturix:
Thanks a lot, will do. :)

Asuranceturix:
Ok, wrote a simple but I hope complete enough test and force-pushed the new commit, let me know whether I need to change anything else.

Library/Formula/xsd.rb
```diff
@@ -0,0 +1,49 @@
+class Xsd < Formula
+  desc "XML Data Binding for C++"
+  homepage "http://www.codesynthesis.com/products/xsd/"
+  url "http://www.codesynthesis.com/download/xsd/4.0/macosx/i686/xsd-4.0.0-i686-macosx.tar.bz2"
```
mikemcquaid: Sorry, just noticed this: is this a compiled binary? We should build this from source or put it in `homebrew/binary`

Library/Formula/xsd.rb
```diff
@@ -0,0 +1,49 @@
...
+        return 0;
+    }
+    EOS
+    system "#{bin}/xsd", "cxx-tree", schema
+    assert File.exist? (testpath/"meaningoflife.hxx")
+    assert File.exist? (testpath/"meaningoflife.cxx")
+    system "c++", "-o", "xsdtest", "xsdtest.cxx", "meaningoflife.cxx", "-lxerces-c"
+    assert File.exist? (testpath/"xsdtest")
+    system (testpath/"xsdtest"), instance
+  end
```
mikemcquaid: The parens here (and most others in this method) aren't needed.

Library/Formula/xsd.rb
```diff
@@ -0,0 +1,49 @@
...
+    #include "meaningoflife.hxx"
+    int main (int argc, char *argv[]) {
+        assert(2==argc);
+        std::auto_ptr< ::xml_schema::positive_integer> x = XSDTest::MeaningOfLife(argv[1]);
+        assert(42==*x);
+        return 0;
+    }
+    EOS
+    system "#{bin}/xsd", "cxx-tree", schema
+    assert File.exist? (testpath/"meaningoflife.hxx")
```
mikemcquaid: `assert File.exist?(testpath/"meaningoflife.hxx")`

Library/Formula/xsd.rb
bfontaine: My fault, I used parens in my example:

```ruby
path = (testpath/"test.xsd")
```

When it can be written:

```ruby
path = testpath/"test.xsd"
```

Asuranceturix: On 2 June 2015 at 11:15, Mike McQuaid <notifications@github.com> wrote:

> In Library/Formula/xsd.rb
> <https://github.com/Homebrew/homebrew/pull/40274#discussion_r31504699>:
>
> > +    #include <cassert>
> > +    #include "meaningoflife.hxx"
> > +    int main (int argc, char *argv[]) {
> > +        assert(2==argc);
> > +        std::auto_ptr< ::xml_schema::positive_integer> x = XSDTest::MeaningOfLife(argv[1]);
> > +        assert(42==*x);
> > +        return 0;
> > +    }
> > +    EOS
> > +    system "#{bin}/xsd", "cxx-tree", schema
> > +    assert File.exist? (testpath/"meaningoflife.hxx")
> > +    assert File.exist? (testpath/"meaningoflife.cxx")
> > +    system "c++", "-o", "xsdtest", "xsdtest.cxx", "meaningoflife.cxx", "-lxerces-c"
> > +    assert File.exist? (testpath/"xsdtest")
> > +    system (testpath/"xsdtest"), instance
> > +  end
>
> The parens here (and most others in this method) aren't needed.
>

Ok, I will fix this later today.

Library/Formula/xsd.rb
Asuranceturix: On 2 June 2015 at 11:14, Mike McQuaid <notifications@github.com> wrote:

> In Library/Formula/xsd.rb
> <https://github.com/Homebrew/homebrew/pull/40274#discussion_r31504657>:
>
> > @@ -0,0 +1,49 @@
> > +class Xsd < Formula
> > +  desc "XML Data Binding for C++"
> > +  homepage "http://www.codesynthesis.com/products/xsd/"
> > +  url "http://www.codesynthesis.com/download/xsd/4.0/macosx/i686/xsd-4.0.0-i686-macosx.tar.bz2"
>
> Sorry, just noticed this: is this a compiled binary? We should build this
> from source or put it in homebrew/binary
>
It is, indeed. I'd say just put this into whatever section it needs to be
(or should I file a pull request elsewhere?) and, when I find time for it,
I will look into compiling it if it seems preferable for the project.

mikemcquaid: Yeh, ideally we'll build from source and ship our own binaries, thanks.

Asuranceturix: Ok, I have fixed the parentheses, but I have not been able to get this to compile from source yet. Can we push this to the appropriate place and when/if I manage to compile it I will post a new formula to compile from source?

tph-thuering:
Please merge :) I would be very glad :+1:

Library/Formula/xsd.rb
bfontaine: You can keep your file locally and install directly from it: `brew install ./xsd.rb`.

Asuranceturix: I know I can install this for myself without putting it into the Homebrew repository, but the whole point of contributing formulas is to make life easier for others. That's why I am suggesting that you push this formula to homebrew/binary if that's the right place for it.

bfontaine: Ok, I misunderstood you. You can make a pull-request to [Homebrew/homebrew-binary](http://github.com/Homebrew/homebrew-binary.

Asuranceturix: Ok, thanks!

Asuranceturix:
Ok, I managed to build it from source. I had some problems because it wouldn't compile cleanly with clang++; it did with g++, which was what the README called for, but then it would crash on runtime because xerces-c is built against libc++. Finally I found the solution in a mailing list; however, it looks like running against libc++ is not-so-well tested (the binary distribution looks statically linked, I guess against libstdc++): http://www.codesynthesis.com/pipermail/xsd-users/2015-February/004522.html

Library/Formula/xsd.rb
```diff
@@ -0,0 +1,78 @@
+class Xsd < Formula
+  desc "XML Data Binding for C++"
+  homepage "http://www.codesynthesis.com/products/xsd/"
+  url "http://www.codesynthesis.com/download/xsd/4.0/xsd-4.0.0+dep.tar.bz2"
+  sha256 "eca52a9c8f52cdbe2ae4e364e4a909503493a0d51ea388fc6c9734565a859817"
+  version "4.0.0"
+
+  depends_on "gcc" => :build
```
mikemcquaid: This won't actually make it build with `gcc` so can be removed.

Library/Formula/xsd.rb
```diff
@@ -0,0 +1,78 @@
...
+  depends_on "findutils" => :build
+  depends_on "pkg-config" => :build
+  depends_on "xerces-c"
+
+  patch :DATA # Version 4.0.0 has some problems with Clang (see http://www.codesynthesis.com/pipermail/xsd-users/2015-February/004522.html )
+
+  def install
+    ENV.append "LDFLAGS", `pkg-config --libs --static xerces-c`.chomp
+    inreplace "xsd/examples/cxx/tree/makefile", "find", "gfind" # find version included with OS X does not support -printf
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Library/Formula/xsd.rb
```diff
@@ -0,0 +1,78 @@
...
+  url "http://www.codesynthesis.com/download/xsd/4.0/xsd-4.0.0+dep.tar.bz2"
+  sha256 "eca52a9c8f52cdbe2ae4e364e4a909503493a0d51ea388fc6c9734565a859817"
+  version "4.0.0"
+
+  depends_on "gcc" => :build
+  depends_on "findutils" => :build
+  depends_on "pkg-config" => :build
+  depends_on "xerces-c"
+
+  patch :DATA # Version 4.0.0 has some problems with Clang (see http://www.codesynthesis.com/pipermail/xsd-users/2015-February/004522.html )
```
mikemcquaid: Can you explain a bit more about why it's needed, put the comments above this line and wrap them at 80 characters?

Library/Formula/xsd.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def install
+    ENV.append "LDFLAGS", `pkg-config --libs --static xerces-c`.chomp
```
mikemcquaid: Seems weird their Makefile doesn't do this?

mikemcquaid:
@Asuranceturix Looking great! A few tweaks to work through but it seems good.

Library/Formula/xsd.rb
Asuranceturix: Do you mean the use of gfind instead of find? Well, there is nothing to patch, it's just that the 'find' that comes with OS X is very old. Using the name 'gfind' instead of 'find' for the GNU find that comes with 'findutils' is a Homebrew-specific tweak, does this really belong upstream?

Library/Formula/xsd.rb
Asuranceturix: OK, but I can say nothing further than what the link says. If you don't include <iostream>, g++ will compile it but clang++ won't.

Library/Formula/xsd.rb
Asuranceturix: Indeed. If I don't set LDFLAGS, make fails complaining that it doesn't know how to make target "-lxerces-c".

Library/Formula/xsd.rb
Asuranceturix: Sorry, I forgot that. At the beginning, I was unable to get it to compile with clang++ and I went the gcc route until I got SIGSEGVs for mixing libc++ and libstdc++.

Library/Formula/xsd.rb
mikemcquaid: Yeh, it'd be good to at least report it or ideally submit a patch that works with the OS X version.

DomT4:
Any update here? Looks nearly complete aside the few comments by Mike above.

Asuranceturix:
No, sorry, I've been quite busy this week. I will try to fix these issues
in the next days.

Asuranceturix:
I have made the changes you guys requested, let's hope this time all is well.

Library/Formula/xsd.rb
```diff
@@ -0,0 +1,98 @@
...
+  version "4.0.0"
+  sha256 "eca52a9c8f52cdbe2ae4e364e4a909503493a0d51ea388fc6c9734565a859817"
+
+  depends_on "pkg-config" => :build
+  depends_on "xerces-c"
+
+  # As of version 4.0.0, Clang fails to compile if the <iostream> header is not
+  # explicitly included. The developers are aware of this problem, see:
+  # http://www.codesynthesis.com/pipermail/xsd-users/2015-February/004522.html
+  patch <<-EOS.undent
```
DomT4: Inline patches need to go at the end under an `__END__` header beneath the main formula. It keeps the main formula body clean and easily readable. Then just:

```ruby
# Patch 1 does xyz
# Reported Patch 1 here: 
# Patch 2 does xyz
# Reported Patch 2 here:
patch :DATA
```

Asuranceturix:
Ok, I just wanted to decouple them. Anyway, it would be a lot easier on contributors if this kind of things were reported by brew audit. It's quite annoying to check everything on the local end only to discover, hours later, that it still doesn't cut it.

DomT4:
It's a relatively difficult thing to catch in the `audit` without flagging false positives.

Acceptable combinations of patching includes `patch do`, `patch :DATA`, `patch do` with all the `p0`, `p1` etc variables, and `patch :DATA` with all the `p0`, `p1` etc variables. We also have several formulae which have their own patch layouts, which could end up being flagged accidentally.

We do raise a warning if someone includes a `patch :DATA` line but no `__END__` line.

Sometimes human intervention remains necessary, I'm afraid that is part of the Homebrew submission process.

Asuranceturix:
Never mind, I'm just a bit frustrated because it's been a month since I first tried to contribute this formula and it seems I manage to get something wrong every time. Here we go again!

Library/Formula/xsd.rb
```diff
@@ -0,0 +1,97 @@
...
+    system "make", "install", "install_prefix=#{prefix}"
+  end
+
+  def caveats; <<-EOS.undent
+    Note that this formula has been compiled with clang++ against libc++ because
+    xerces-c uses libc++ too. Apparently, this configuration is not well tested
+    and you might encounter some issues, see:
+    http://www.codesynthesis.com/pipermail/xsd-users/2015-February/004522.html
+    EOS
+  end
```
mikemcquaid: It makes me a bit uncomfortable that we're basically telling users "stuff might go wrong". That seems to indicate this isn't ready for a Homebrew formula yet (particularly given it needs patched still to be installed)?

Asuranceturix: The software is mature (version 4, I use it every day), but the authors have documented a marginal condition of failure in the test suite when using floating point with libc++, which I am forced to use because Homebrew's xerces-c is built against libc++. 

As for the patch, I did work to remove build dependency on findutils and did submit the patch upstream; the devs told me that next versions of XSD will have a new building system that will probably work better ( http://codesynthesis.com/pipermail/xsd-users/2015-July/004639.html )

If you deem more useful a non-formula, which by definition has no failures, than a software that might have some quirks, that's OK. However, the caveat was in the formula since the first version I submitted, you could have saved me some work if this was a showstopper.

mikemcquaid:

I removed the caveats as they aren't alerting to a specific issue and I think they will probably cause more concern that they alleviate. If users report issues to us we'll definitely consider adding them back.

Asuranceturix:
Always glad to help. :)

PR #40284, opened 2015-06-01
Update open-mpi.rb

janmolnar:
Added fortran compiler awareness


mikemcquaid:
What is this fixing?

janmolnar:
If you set the FC Environment Variable to eg ifort then open mpi is built against that fortan compiler

jacknagel:
What does passing it to configure accomplish that is not accomplished by it already being set in the ENV var?

DomT4:
Jack's question needs answering before this can be considered for merge, Thanks!

DomT4:
Presuming dead.

PR #40286, opened 2015-06-01
webalizer 2.23-08

dunn:


Library/Formula/webalizer.rb
```diff
@@ -1,26 +1,31 @@
-require 'formula'
-
 class Webalizer < Formula
   homepage "http://www.mrunix.net/webalizer/"
```
vszakats: Maybe better to follow the permanent redirect to this address?:
http://www.webalizer.org/

Library/Formula/webalizer.rb
```diff
@@ -1,26 +1,31 @@
...
-    system "make install"
+                          "--prefix=#{prefix}",
+                          "--mandir=#{man}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.log").write "127.0.0.1 user-identifier homebrew [10/Oct/2000:13:55:36 -0700] \"GET /beer.gif HTTP/1.0\" 200 2326"
+    system "#{bin}/webalizer", "-c", etc/"webalizer.conf.sample", testpath/"test.log"
+    File.exist? "usage.png"
```
DomT4: Should be `assert File.exist?`

dunn: OK, but without an assertion message it's not much more informative is it?

tdsmith: Yeah, this is fine; if the `test` block returns false, the test fails.

xu-cheng:
Thanks @dunn

PR #40295, opened 2015-06-02
test-bot: simplify the upload logic

xu-cheng:
If the bottle is already published, let's fail directly.
Also silent the curl output.

### DIFF
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -789,6 +789,14 @@ module Homebrew
         version = formula.pkg_version
         bintray_package = Bintray.package formula_name
 
+        if system "curl", "-I", "--silent", "--fail", "--output", "/dev/null",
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/test-bot.rb
```diff
@@ -789,6 +789,11 @@ def test_bot
         version = formula.pkg_version
         bintray_package = Bintray.package formula_name
 
+        if system "curl", "-I", "--silent", "--fail", "--output", "/dev/null",
+                  "#{BottleSpecification::DEFAULT_DOMAIN}/#{bintray_repo}/#{filename}"
+          raise "#{filename} is already published. Please remove it manually."
```
mikemcquaid: Would be good to provide the full URL to the file so people can click through there to remove it.

xu-cheng: Good idea, updated.

mikemcquaid:
:clap: LGTM

PR #40299, opened 2015-06-02
strongswan 5.3.1

tobiasbrunner:


tdsmith:
Thanks for your pull request! :tada: Homebrew thrives because of contributions like this one. Just for future reference, we prefer sha256 hashes to sha1 hashes. I've made the change myself. Thanks again!

PR #40304, opened 2015-06-02
libmtp 1.1.9

alkk:
libmtp updated to latest release 1.1.9, which extend list of supported devices (e.g. Lumia 930)


tdsmith:
Thanks for your pull request! :tada: Homebrew thrives because of contributions like this one. Just for reference, we prefer sha256 hashes to sha1 hashes. I've made that change myself. Thanks again!

PR #40307, opened 2015-06-02
update to 1.1.1

ajschumacher:
mirrors for 1.1.0 are already failing


tdsmith:
Thanks for your pull request! :tada: Homebrew thrives because of contributions like this. Just for reference, our preferred commit message header is like "somepackage 1.2.3" for version bumps like this one. We also prefer sha256 hashes to sha1 hashes. I've made those changes myself. Thanks again!

PR #40308, opened 2015-06-02
pinentry 0.9.3

chdiza:


chdiza:
I added the new flag because pinentry added an additional option that should be explicitly disabled for the same reason that we `--disable-pinentry-{qt4,gtk2}`.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/98b39ca1fd184eae25731d36ba3659dd8837b2f2. Thanks @chdiza!

PR #40322, opened 2015-06-02
kitchen-sync 0.38 (new formula)

denji:
- 0.38, OpenSSL removed from the CMakeList for OS X.

### DIFF
--- a/Library/Formula/kitchen-sync.rb
+++ b/Library/Formula/kitchen-sync.rb
@@ -1,0 +1,23 @@
+class KitchenSync < Formula
+  desc "Fast efficiently sync database without dumping & reloading"
+  homepage "https://github.com/willbryant/kitchen_sync"
+  url "https://github.com/willbryant/kitchen_sync/archive/0.38.tar.gz"
+  sha256 "786bf30caca4e58347c0880dc1ca54685cd01d0834024c425029fb5466c96041"
+  head "https://github.com/willbryant/kitchen_sync.git"
+
+  depends_on "cmake" => :build
+  depends_on "boost"
+  depends_on :mysql => :recommended
+  depends_on :postgresql => :optional
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    assert File.exist?("#{bin}/ks")
+    assert File.exist?("#{bin}/ks_mysql") if build.with? "mysql"
+    assert File.exist?("#{bin}/ks_postgresql") if build.with? "postgresql"
+  end
+end
--- a/Library/Formula/kitchen-sync.rb
+++ b/Library/Formula/kitchen-sync.rb
@@ -10,7 +10,10 @@ class KitchenSync < Formula
   depends_on :mysql => :recommended
   depends_on :postgresql => :optional
 
+  needs :cxx11
+
   def install
+    ENV.cxx11
     system "cmake", ".", *std_cmake_args
     system "make", "install"
   end


Library/Formula/kitchen-sync.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    assert File.exist?("#{bin}/ks")
+    assert File.exist?("#{bin}/ks_mysql") if build.with? "mysql"
+    assert File.exist?("#{bin}/ks_postgresql") if build.with? "postgresql"
```
bfontaine: Could you add a more substantial test than e.g. asserting files existence or `--help`/`--version`? Thanks!

denji: @bfontaine There is no argument for `-h (--help)`, we can connect to the database you want to synchronize or run the tests work with base.

bfontaine: Mmmh I see, there’s no way to test this without a local database, and no `--help`/`--version` flag. @mikemcquaid Thoughts?

Library/Formula/kitchen-sync.rb
```diff
@@ -0,0 +1,23 @@
...
+  desc "Fast efficiently sync database without dumping & reloading"
+  homepage "https://github.com/willbryant/kitchen_sync"
+  url "https://github.com/willbryant/kitchen_sync/archive/0.37.tar.gz"
+  sha256 "a78206cdd3ac4d4aa646cdfc8ae34c5df87179223c59d5dcca625760c68ae4e7"
+  head "https://github.com/willbryant/kitchen_sync.git"
+
+  depends_on "cmake" => :build
+  depends_on "boost"
+  depends_on :mysql => :recommended
+  depends_on :postgresql => :optional
```
bfontaine: It should also `depends_on "openssl"`, as mentioned in the [upstream docs](https://github.com/willbryant/kitchen_sync/blob/master/INSTALL.md#installing-kitchen-sync).

Library/Formula/kitchen-sync.rb
willbryant: Hey guys, great to see this.  I believe you can remove the openssl dependency, that was my bad - I forgot to update the installation instructions when I added direct support for Common Crypto on OS X, so now OpenSSL is only needed on Linux and so on.

Let me know if it isn't compiling without it or something - unfortunately all my machines have OpenSSL already so hard for me to test this properly.  All I know for sure is that we don't get the deprecation warnings...

bfontaine:
The test is failing because of an issue in the core that has since been resolved, could you please rebase on top of `master`?

Also, regarding the OpenSSL issue, if you don’t have Homebrew’s OpenSSL it builds against OSX’ outdated one:

```
==> Downloading https://github.com/willbryant/kitchen_sync/archive/0.37.tar.gz
Already downloaded: /Library/Caches/Homebrew/kitchen-sync-0.37.tar.gz
==> cmake . -DCMAKE_C_FLAGS_RELEASE= -DCMAKE_CXX_FLAGS_RELEASE= -DCMAKE_INSTALL_PREFIX=/usr/local/Cel
==> make install
Warning: object files were linked against system openssl
  /usr/local/Cellar/kitchen-sync/0.37/bin/ks_mysql
==> Summary
🍺  /usr/local/Cellar/kitchen-sync/0.37: 43 files, 1,6M, built in 29 seconds
```

willbryant:
Hmm, that isn't meant to happen... we need to skip the `find_package(OpenSSL REQUIRED)` on OS X I guess?

willbryant:
Pushed 0.38 with OpenSSL removed from the CMakeList for OS X.

denji:
@willbryant http://bot.brew.sh/job/Homebrew%20Pull%20Requests/27254/version=mountain_lion/

DomT4:
Looks like it needs `cxx11` probably.

denji:
@DomT4 Now it works fine!

mikemcquaid:
Thanks @denji!

mikemcquaid:
Merged in 19d2554.

PR #40325, opened 2015-06-02
remove openssl as an osquery dependency

marpaia:
osquery no longer depends on the homebrew openssl library

### DIFF
--- a/Library/Formula/osquery.rb
+++ b/Library/Formula/osquery.rb
@@ -22,7 +22,6 @@ class Osquery < Formula
   depends_on "rocksdb" => :build
   depends_on "thrift" => :build
   depends_on "yara" => :build
-  depends_on "openssl"
 
   resource "markupsafe" do
     url "https://pypi.python.org/packages/source/M/MarkupSafe/MarkupSafe-0.23.tar.gz"


DomT4:
What does it use instead?

marpaia:
@DomT4 we use the OpenSSL which is installed on OS X

bfontaine:
The OpenSSL shipped by OSX is outdated.

DomT4:
OS X doesn't ship a modern, supported OpenSSL. It's capped on the 0.9.8 branch, which doesn't support the modern and secure TLS versions, and is rarely updated by Apple. We enforce our OpenSSL for security reasons, and an exemption is unlikely to be made here.

theopolis:
The osquery build tries as hard as possible to build with the safest TLS options available, 0.9.8 or 1.0+: https://github.com/facebook/osquery/blob/master/osquery/remote/transports/tls.cpp

It looks like the Homebrew-built osquery is finding the Homebrew-installed OpenSSL in `/usr/local/opt/openssl/lib/` just fine (from test source/bottle builds). It should be fine to keep the Homebrew OpenSSL dependency but the resultant binaries cannot be shipped to another OS X install. NBD :/

DomT4:
> The osquery build tries as hard as possible to build with the safest TLS options available, 0.9.8 

Pretty much everything before TLS 1.1 has known security flaws. SSLv2 and SSLv3 are obviously the worst culprits, but TLS 1.0 has taken a fair bashing as well, and TLS 1.0 is where the 0.9.8 branch is capped. Even the strongest connection the 0.9.8 branch can make has known previous vulnerabilities and people have been trying to kill it off for a while.

Per SSLLabs:

> Just last month, the PCI Security Council deprecated SSL v3 and TLS 1.0 for commercial transactions. No new systems are allowed to use TLS 1.0 for credit card processing and existing systems must immediately begin to transition to better protocols.

https://community.qualys.com/blogs/securitylabs/2015/05/22/ssl-labs-increased-penalty-when-tls-12-is-not-supported

> but the resultant binaries cannot be shipped to another OS X install.

The answer to this is to ship the custom OpenSSL as well, which is the route others have gone down, or modifying the formula locally when you ship it to remove the dependency. Either way, I wouldn't be comfortable removing the safer OpenSSL for *all* Homebrew users.

theopolis:
Good summary @DomT4, I'm very aware! :D

The real solution is to kick Apple and make them ship a system-available OpenSSL. Embedding OpenSSL alongside an application install is just as much a sin. Though Apple suggests just that in their docs, haha!

> The answer to this is to ship the custom OpenSSL as well, which is the route others have gone down, or modifying the formula locally when you ship it to remove the dependency. Either way, I wouldn't be comfortable removing the safer OpenSSL for all Homebrew users.

Agreed, in this case deployment ease does not outweigh the benefits of a modern TLS client; as Homebrew is a simple dependency. Homebrew is saving lives as an intermediate to bundling OpenSSL and expecting an OS to ship.

marpaia:
We had some internal talk about this, juice isn't worth the squeeze. As you were :)

DomT4:
> The real solution is to kick Apple and make them ship a system-available OpenSSL.

I've had a bug report open with them for the last 4 months on OpenSSL deprecating the 0.9.8 branch at the end of 2015 and that would be a useful opportunity to either defer shipping OpenSSL to package managers permanently, or to update the shipped OpenSSL to a more modern branch (Or alternative). So far, nada.

There's been discussion about it on some of the internal dev forums as well, so it's possible the noise will reach Apple eventually, heh. Hard to know either way though.

Thanks for understanding our viewpoint, and the reasonable discussion. Appreciated! :heart:. Feel free to ping if there's anything else that's useful to you.

mikemcquaid:
> The real solution is to kick Apple and make them ship a system-available OpenSSL. Embedding OpenSSL alongside an application install is just as much a sin. Though Apple suggests just that in their docs, haha!

Just FYI here: OpenSSL is officially deprecated on OS X and the way `.app` bundles are typically generated means vendoring all your dependencies on OS X. As a result, Apple aren't going to ship a newer version.

theopolis:
Yes, deprecated since 10.7! Even though the MO on Apple and Windows is to vendor in dependencies along side bundles/installs, there *should* be considerations for critical and ubiquitous libraries such as OpenSSL. It's fairly obvious why. ;) And both platforms do include many dependent libraries! I have hope Apple will one day again include a modern OpenSSL. :dancers:

mikemcquaid:
It won't. You're meant to be using Common Crypto instead now.

PR #40326, opened 2015-06-02
Give `brew tap` a little `brew any-tap` magic!

telemachus:
`brew tap` restricts users to GitHub and checks for private repos. I created
`brew any-tap` to support a wider variety of taps.

With very small changes (and no extra flags!), this PR brings `any-tap` to `brew`.

Users can use `brew tap` with GitHub as they always have or add one extra 
argument and tap any repository of any type from anywhere.

tl;dr

    brew tap user/name              # Same as it ever was
    brew tap user/name URL          # Tap URL, whatever it happens to be

ping @mikemcquaid @mistydemeo @jacknagel

### DIFF
--- a/Library/Homebrew/cmd/tap.rb
+++ b/Library/Homebrew/cmd/tap.rb
@@ -1,11 +1,15 @@
 module Homebrew
   def tap
     if ARGV.empty?
-      show_taps
+      each_tap do |user, repo|
+        if (repo/".git").directory?
+          puts "#{user.basename}/#{repo.basename.sub("homebrew-", "")}"
+        end
+      end
     elsif ARGV.first == "--repair"
       migrate_taps :force => true
     else
-      opoo "Already tapped!" unless install_tap(*tap_args, ARGV.named[1])
+      opoo "Already tapped!" unless install_tap(*tap_args)
     end
   end
 
@@ -70,7 +74,7 @@ module Homebrew
   def tap_args(tap_name=ARGV.named.first)
     tap_name =~ HOMEBREW_TAP_ARGS_REGEX
     raise "Invalid tap name" unless $1 && $3
-    [$1, $3]
+    [$1, $3, ARGV.named[1]]
   end
 
   def private_tap?(user, repo)
@@ -79,13 +83,5 @@ module Homebrew
     true
   rescue GitHub::Error
     false
-  end
-
-  def show_taps
-    each_tap do |user, repo|
-      if (repo/".git").directory?
-        puts "#{user.basename}/#{repo.basename.sub("homebrew-", "")}"
-      end
-    end
   end
 end
--- a/Library/Homebrew/cmd/tap.rb
+++ b/Library/Homebrew/cmd/tap.rb
@@ -9,7 +9,9 @@ module Homebrew
     elsif ARGV.first == "--repair"
       migrate_taps :force => true
     else
-      opoo "Already tapped!" unless install_tap(*tap_args)
+      user, repo = tap_args
+      clone_target = ARGV.named[1]
+      opoo "Already tapped!" unless install_tap(user, repo, clone_target)
     end
   end
 
@@ -74,7 +76,7 @@ module Homebrew
   def tap_args(tap_name=ARGV.named.first)
     tap_name =~ HOMEBREW_TAP_ARGS_REGEX
     raise "Invalid tap name" unless $1 && $3
-    [$1, $3, ARGV.named[1]]
+    [$1, $3]
   end
 
   def private_tap?(user, repo)
--- a/Library/Homebrew/cmd/tap.rb
+++ b/Library/Homebrew/cmd/tap.rb
@@ -11,7 +11,7 @@ module Homebrew
     end
   end
 
-  def install_tap user, repo
+  def install_tap user, repo, clone_target=nil
     # we special case homebrew so users don't have to shift in a terminal
     repouser = if user == "homebrew" then "Homebrew" else user end
     user = "homebrew" if user == "Homebrew"
@@ -20,7 +20,11 @@ module Homebrew
     tapd = HOMEBREW_LIBRARY/"Taps/#{user.downcase}/homebrew-#{repo.downcase}"
     return false if tapd.directory?
     ohai "Tapping #{repouser}/#{repo}"
-    args = %W[clone https://github.com/#{repouser}/homebrew-#{repo} #{tapd}]
+    if clone_target
+      args = %W[clone #{clone_target} #{tapd}]
+    else
+      args = %W[clone https://github.com/#{repouser}/homebrew-#{repo} #{tapd}]
+    end
     args << "--depth=1" unless ARGV.include?("--full")
     safe_system "git", *args
 
@@ -28,14 +32,16 @@ module Homebrew
     tapd.find_formula { |file| files << file }
     puts "Tapped #{files.length} formula#{plural(files.length, 'e')} (#{tapd.abv})"
 
-    if private_tap?(repouser, repo) then puts <<-EOS.undent
-      It looks like you tapped a private repository. To avoid entering your
-      credentials each time you update, you can use git HTTP credential caching
-      or issue the following command:
+    unless clone_target
+      if private_tap?(repouser, repo) then puts <<-EOS.undent
+        It looks like you tapped a private repository. To avoid entering your
+        credentials each time you update, you can use git HTTP credential
+        caching or issue the following command:
 
-        cd #{tapd}
-        git remote set-url origin git@github.com:#{repouser}/homebrew-#{repo}.git
-      EOS
+          cd #{tapd}
+          git remote set-url origin git@github.com:#{repouser}/homebrew-#{repo}.git
+        EOS
+      end
     end
 
     true
@@ -63,10 +69,11 @@ module Homebrew
     end
   end
 
-  def tap_args(tap_name=ARGV.named.first)
+  def tap_args()
+    tap_name, clone_target = ARGV.named[0..1]
     tap_name =~ HOMEBREW_TAP_ARGS_REGEX
     raise "Invalid tap name" unless $1 && $3
-    [$1, $3]
+    [$1, $3, clone_target]
   end
 
   def private_tap?(user, repo)
--- a/Library/Homebrew/cmd/tap.rb
+++ b/Library/Homebrew/cmd/tap.rb
@@ -9,7 +9,7 @@ module Homebrew
     end
   end
 
-  def install_tap user, repo, clone_target
+  def install_tap user, repo, clone_target=nil
     # we special case homebrew so users don't have to shift in a terminal
     repouser = if user == "homebrew" then "Homebrew" else user end
     user = "homebrew" if user == "Homebrew"
--- a/Library/Homebrew/cmd/tap.rb
+++ b/Library/Homebrew/cmd/tap.rb
@@ -2,9 +2,7 @@ module Homebrew
   def tap
     if ARGV.empty?
       each_tap do |user, repo|
-        if (repo/".git").directory?
-          puts "#{user.basename}/#{repo.basename.sub("homebrew-", "")}"
-        end
+        puts "#{user.basename}/#{repo.basename.sub("homebrew-", "")}" if (repo/".git").directory?
       end
     elsif ARGV.first == "--repair"
       migrate_taps :force => true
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cmd/tap.rb
```diff
@@ -28,14 +32,16 @@ def install_tap user, repo
     tapd.find_formula { |file| files << file }
     puts "Tapped #{files.length} formula#{plural(files.length, 'e')} (#{tapd.abv})"
 
-    if private_tap?(repouser, repo) then puts <<-EOS.undent
-      It looks like you tapped a private repository. To avoid entering your
-      credentials each time you update, you can use git HTTP credential caching
-      or issue the following command:
+    unless clone_target
```
mistydemeo: Do we still want to print some form of message if the user appears to be tapping a private repository with a custom URL? Or trust the user to know what to do?

Library/Homebrew/cmd/tap.rb
```diff
@@ -63,10 +69,11 @@ def each_tap
     end
   end
 
-  def tap_args(tap_name=ARGV.named.first)
+  def tap_args()
+    tap_name, clone_target = ARGV.named[0..1]
```
xu-cheng: Don't change this method. Instead you should handle anytap situation inside `Homebrew#tap` method.

Library/Homebrew/cmd/untap.rb
```diff
@@ -5,7 +5,7 @@ def untap
     raise "Usage is `brew untap <tap-name>`" if ARGV.empty?
 
     ARGV.each do |tapname|
-      user, repo = tap_args(tapname)
+      user, repo = tap_args()
```
xu-cheng: This changes existing behaviors as it used to be able to untap multiple taps simultaneously.

Library/Homebrew/cmd/tap.rb
telemachus: I thought no: My assumption was that anyone who asks for the custom method knows what their doing. What do you think though?

Library/Homebrew/cmd/tap.rb
telemachus: That would require more branching. This struck me as an easier, simpler solution. If some method is going to change, why *not* this one?

Library/Homebrew/cmd/untap.rb
telemachus: That's true. I'll need to think about that more. (Does that explain why you don't want me to change the `tap_args` method?)

Library/Homebrew/cmd/tap.rb
xu-cheng: > I think that would require an extra flag

You can just check the size of `ARGV.named` inside `Homebrew#tap`. 

> I think some change to this tap_args method will be necessary

I would like to keep this as it is if possible, since this method is used in somewhere else. Also the difference between tap and any-tap only applies inside tap.rb. Let's keep the influence only inside tap.rb as well.

telemachus:
@xu-cheng As you suggested, this now leaves the `tap_args` method as it was before. `brew untap` can once again handle multiple arguments without a problem.

In addition, I've given `brew untap` without arguments the same behavior as `brew tap` without arguments: both show all currently installed taps. This seems better than having different behaviors for the two.

Library/Homebrew/cmd/tap.rb
```diff
@@ -1,17 +1,15 @@
...
     elsif ARGV.first == "--repair"
       migrate_taps :force => true
     else
-      opoo "Already tapped!" unless install_tap(*tap_args)
+      opoo "Already tapped!" unless install_tap(*tap_args, ARGV.named[1])
     end
   end
 
-  def install_tap user, repo
+  def install_tap user, repo, clone_target
```
xu-cheng: `def install_tap user, repo, clone_target=nil`

So it can accept two arguments as this method is also used elsewhere.

xu-cheng:
Personally, I don't see the value to let untap print all the taps. But will let other mantainers decide whether kill such feature.

Library/Homebrew/cmd/tap.rb
telemachus: Sorry I missed that, but where else is it used?

telemachus:
> Personally, I don't see the value to let untap print all the taps.

It requires nearly no additional code. Since it's been useful for other users (i.e. the users of `brew any-tap`), my thought is "What's the harm?"

Library/Homebrew/cmd/tap.rb
mikemcquaid: I think it'd be good to do that if we can for the custom method.

Library/Homebrew/cmd/untap.rb
```diff
@@ -1,29 +1,31 @@
-require 'cmd/tap' # for tap_args
+require 'cmd/tap' # for tap_args and show_taps
 
 module Homebrew
   def untap
-    raise "Usage is `brew untap <tap-name>`" if ARGV.empty?
+    if ARGV.empty?
+      show_taps
```
mikemcquaid: This feels a bit weird to me.

mikemcquaid:
> Personally, I don't see the value to let untap print all the taps. But will let other mantainers decide whether kill such feature.

Yeh, it feels a little weird to me too.

Library/Homebrew/cmd/untap.rb
telemachus: Sorry, but in what way? `brew tap` begins with the same check of empty arguments.

My default rule of thumb is this: if a command-line tool can do something useful with a no-argument call, do that instead of just throwing an error. That said, I can remove this if people feel strongly against it.

telemachus:
> Yeh, it feels a little weird to me too.

Ok, I'll make my argument one more time and then remove it if I'm outvoted.

Imagine a user who needs to remove a tap, but can't recall exactly the name of that tap. As things stand, that user must run this sequence:

    brew tap
    # get a look at the exact names
    brew untap <plug-in-proper-name-here>

What I'm proposing is that instead they could run the simpler:

    brew untap
    # see list of names
    # Up-arrow and then add proper name

In the end, it's not a **huge** difference either way, but since the change in code is minimal, why not give `brew untap` an extra ability.

As it stands, if you type `brew untap` without arguments, you just get an error. So it's not as though we would be removing something useful. We're removing something frustrating for users (an error message) and adding a new behavior that can never hurt anyone.

Library/Homebrew/cmd/tap.rb
xu-cheng: `git grep install_tap` will show them.

Library/Homebrew/cmd/tap.rb
telemachus: @mikemcquaid Ok, that's easy. I'll just remove the `if` wrapper and leave it as it was originally. Here's my argument against this version though: the whole point of `brew any-tap` was to be able to do more than `brew tap`. For someone who taps numerous private repos, won't it become annoying to get the same reminder message over and over?

mikemcquaid: Hmm, yeh, I see. Would be good if we can detect if that's necessary and if not just omit it (as you've done)

telemachus: I didn't omit it—or not intentionally at least. 

The `unless clone_target` test checks whether this is the "fancy" style of tapping where we are given an extra argument. If we are given an explicit item to clone, then we are in `any-tap` mode, and so we assume the user knows what they're doing. If we're still in `tap` mode (i.e. the way Homebrew has always done it), then we check for private repos.

tl;dr For people who continue to run only `brew tap`, this should work like it always did.

Library/Homebrew/cmd/tap.rb
```diff
@@ -1,17 +1,15 @@
...
     if ARGV.empty?
-      each_tap do |user, repo|
-        puts "#{user.basename}/#{repo.basename.sub("homebrew-", "")}" if (repo/".git").directory?
-      end
+      show_taps
     elsif ARGV.first == "--repair"
       migrate_taps :force => true
     else
-      opoo "Already tapped!" unless install_tap(*tap_args)
+      opoo "Already tapped!" unless install_tap(*tap_args, ARGV.named[1])
```
xu-cheng: This is syntax incorrect for ruby 1.8.

xu-cheng:
> What I'm proposing is that instead they could run the simpler:
```
brew untap
# see list of names
# Up-arrow and then add proper name
```
In the end, it's not a huge difference either way, but since the change in code is minimal, why not give brew untap an extra ability?

This should belong to shell complete. On the other side, letting `brew untap` print all the taps can be misleading as it feels like to untap all the taps.

>  As it stands, if you type brew untap without arguments, you just get an error. 

We can discard this error, so brew untap without arguments just does nothing.

mikemcquaid:
> This should belong to shell complete. On the other side, letting brew untap print all the taps can be misleading as it feels like to untap all the taps.

Agreed.

telemachus:
> This should belong to shell complete.

[It's already there](https://github.com/Homebrew/homebrew/blob/master/Library/Contributions/brew_bash_completion.sh#L113-L125). I didn't know that, and it sounds like maybe all three of us didn't know. Which maybe shows that having it *only* in the completion files is not optimal. But I yield. I'll remove this behavior and revert to the old behavior. No message at all feels like a step backwards in terms of usability: what will the user think happened or didn't? A usage message is better than nothing I think.

telemachus:
Ok, I think I covered everything that was requested:

+ Revert `brew tap` with no arguments to the old behavior
+ Remove the argument handling that was no good for ruby 1.8

Let me know please if there's anything else I should do to
get this merged.

Thanks.

Library/Homebrew/cmd/tap.rb
```diff
@@ -66,7 +74,7 @@ def each_tap
   def tap_args(tap_name=ARGV.named.first)
     tap_name =~ HOMEBREW_TAP_ARGS_REGEX
     raise "Invalid tap name" unless $1 && $3
-    [$1, $3]
+    [$1, $3, ARGV.named[1]]
```
xu-cheng: `ARGV.named[1]` doesn't belong to this method. I mean you should fix the syntax problem in this way:

```ruby
user, repo = tap_args
clone_target  = ARGV.named[1]
opoo "Already tapped!" unless install_tap(user, repo, clone_target)
```

telemachus: Done

Library/Homebrew/cmd/tap.rb
```diff
@@ -2,16 +2,20 @@ module Homebrew
   def tap
     if ARGV.empty?
       each_tap do |user, repo|
-        puts "#{user.basename}/#{repo.basename.sub("homebrew-", "")}" if (repo/".git").directory?
+        if (repo/".git").directory?
+          puts "#{user.basename}/#{repo.basename.sub("homebrew-", "")}"
+        end
```
jacknagel: Is there a functional change here? If so, what is it? Otherwise I'd prefer to avoid the churn.

telemachus: Nope. I found it easier to read when I wrote my version as a method, but if you like I'll go back to the old. Two secs for a PR.

Library/Homebrew/cmd/tap.rb
xu-cheng: This `unless` and next `if` should be combined into one condition statement.

telemachus: @xu-cheng I see what you mean. I would rather make that a tiny method then.

That is, I would prefer this:

    if check_private?(clone_target, repouser, repo)

to this:

    if not clone_target && private_tap?(repouser, repo)`

Any objections to that before I make the change?

telemachus:
I've simplified two branches (as @xu-cheng asked), and I think I've covered everything people suggested or asked for.

Let me know if anything else prevents this from being pulled.

mikemcquaid:
This looks good to me. Any objections to this being pulled, @Homebrew/owners?

telemachus:
@mikemcquaid Pretty funny if this ends up being merged the same day as [brew desc](https://github.com/Homebrew/homebrew/pull/39911#issuecomment-109298515).

mikemcquaid:
Thanks @telemachus! Can you queue another PR to adjust the manpage when you get a chance? Cheers.

telemachus:
@mikemcquaid My pleasure. I should be able to do it before the end of the weekend. Thanks so much!

PR #40330, opened 2015-06-02
getdns 0.2.0

ngoyal:


Library/Formula/getdns.rb
```diff
@@ -2,8 +2,8 @@
 
 class Getdns < Formula
   homepage "http://getdnsapi.net"
```
vszakats: Could you please change homepage too to https?:
https://getdnsapi.net

vszakats: Thanks!

Library/Formula/getdns.rb
```diff
@@ -21,7 +22,10 @@ class Getdns < Formula
   depends_on "libev" => :optional
 
   def install
-    args = []
+    args = [
+      "--with-ssl=#{Formula["openssl"].opt_prefix}",
+      "--with-trust-anchor=/usr/local/etc/getdns-root.key"
```
DomT4: Don't hardcode this to `/usr/local`. Should be `#{etc}/getdns-root.key`.

ngoyal: Thanks - it's good now - somehow I missed the docs @ https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#variables-for-directory-locations

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/1879f33d41ea82813c3a739c149fe886726bcd48. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Couple of small nits for future: The preferred checksum is now `sha256` and commits should be squashed. Otherwise :100:, Thanks!

PR #40332, opened 2015-06-03
swiftcov 0.1.0 (new formula)

jpsim:
Thanks! /cc @kishikawakatsumi

### DIFF
--- a/Library/Formula/swiftcov.rb
+++ b/Library/Formula/swiftcov.rb
@@ -1,5 +1,6 @@
 class Swiftcov < Formula
   homepage "https://github.com/realm/SwiftCov"
+  desc "A tool to generate test code coverage information for Swift."
   url "https://github.com/realm/SwiftCov.git", :tag => "0.1.0", :revision => "3a13f0ede4449fd3e4e07878031544d906ee19f9"
   head "https://github.com/realm/SwiftCov.git"
 
@@ -10,6 +11,19 @@ class Swiftcov < Formula
   end
 
   test do
-    system "#{bin}/swiftcov help"
+    system "#{bin}/swiftcov",
+           "generate",
+           "--output",
+           "swiftcov_test",
+           "xcodebuild",
+           "test",
+           "-project",
+           "Examples/ExampleFramework/ExampleFramework.xcodeproj",
+           "-scheme",
+           "ExampleFramework-Mac",
+           "-sdk",
+           "macosx",
+           "-configuration",
+           "Release"
   end
 end
--- a/Library/Formula/swiftcov.rb
+++ b/Library/Formula/swiftcov.rb
@@ -11,6 +11,10 @@ class Swiftcov < Formula
   end
 
   test do
+    system "cd", testpath
+    system "wget", "https://github.com/realm/SwiftCov/archive/#{version}.zip"
+    system "unzip", "#{version}.zip"
+    system "cd", "SwiftCov-#{version}"
     system "#{bin}/swiftcov",
            "generate",
            "--output",
--- a/Library/Formula/swiftcov.rb
+++ b/Library/Formula/swiftcov.rb
@@ -11,10 +11,10 @@ class Swiftcov < Formula
   end
 
   test do
-    system "cd", testpath
-    system "wget", "https://github.com/realm/SwiftCov/archive/#{version}.zip"
+    Dir.chdir testpath
+    system "curl", "-OL", "https://github.com/realm/SwiftCov/archive/#{version}.zip"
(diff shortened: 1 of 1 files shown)


Library/Formula/swiftcov.rb
```diff
@@ -0,0 +1,15 @@
...
+  head "https://github.com/realm/SwiftCov.git"
+
+  depends_on :xcode => ["6.3", :build]
+
+  def install
+    system "make", "prefix_install", "PREFIX=#{prefix}", "TEMPORARY_FOLDER=#{buildpath}/SwiftCov.dst"
+  end
+
+  test do
+    system "#{bin}/swiftcov help"
```
bfontaine: Please use `system "#{bin}/swiftcov", "help"` instead of `system "#{bin}/swiftcov help"`. Also, could you write a more substantial test than e.g. `--help` or `--version`? Thanks!

Library/Formula/swiftcov.rb
```diff
@@ -0,0 +1,15 @@
+class Swiftcov < Formula
+  homepage "https://github.com/realm/SwiftCov"
```
bfontaine: Please add a short description using `desc "…"`.

Library/Formula/swiftcov.rb
jpsim: It would be challenging to write a more substantial test here given that an Xcode project with unit tests for a Swift framework has to be available in order to generate test code coverage data.

bfontaine: Thanks, just fix the existing line and that’ll be ok :+1: 

```ruby
system "#{bin}/swiftcov", "help"
```

mikemcquaid: If you could write these out or download them as a resource that could work.

jpsim: Ok, I think I could download a resource. Good idea!

jpsim:
I've addressed your feedback @bfontaine and @mikemcquaid, but the CI jobs seem to have failed for an unrelated reason (git exception).

bfontaine:

jpsim:

bfontaine:
@jpsim The bot (re-)tests everything when you push to this PR. The “test this please” only works for maintainers :wink: 

The test is failing on Yosemite:

```
==> /usr/local/Cellar/swiftcov/0.1.0/bin/swiftcov generate --output swiftcov_test xcodebuild test -project Examples/ExampleFramework/ExampleFramework.xcodeproj -scheme ExampleFramework-Mac -sdk macosx -configuration Release

Testing failed:
	Test target ExampleFramework-MacTests encountered an error (Test session exited(-1) without checking in. If you believe this error represents a bug, please attach the log file at /var/folders/03/jqjv4qms2b18_hwt2dcgsghc0000gp/T/com.apple.dt.XCTest-status/Session-2015-06-03_17:12:40-Cz8dXz.log)
** TEST FAILED **
```

jpsim:
I figured as much when it didn't workt. The reason why attempted

jpsim:
Sorry, accidentally pressed comment there. The reason I attempted to re-test was that I believed that particular test run was a fluke and should have succeeded.

mikemcquaid:
Seems not, unfortunately.

Library/Formula/swiftcov.rb
```diff
@@ -0,0 +1,33 @@
...
+  desc "A tool to generate test code coverage information for Swift."
+  url "https://github.com/realm/SwiftCov.git", :tag => "0.1.0", :revision => "3a13f0ede4449fd3e4e07878031544d906ee19f9"
```
DomT4: `:tag` and `:revision` should be pushed onto unique lines below so the formula isn't overly wide.

Library/Formula/swiftcov.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  depends_on :xcode => ["6.3", :build]
+
+  def install
+    system "make", "prefix_install", "PREFIX=#{prefix}", "TEMPORARY_FOLDER=#{buildpath}/SwiftCov.dst"
+  end
+
+  test do
+    Dir.chdir testpath
+    system "curl", "-OL", "https://github.com/realm/SwiftCov/archive/#{version}.zip"
```
DomT4: This should be a resource rather than `curl`'d directly into the `testpath`.

jpsim:
I think we'll actually be taking this down given that Apple just released something similar. Thanks for your feedback in here.

DomT4:
Thanks for the submission nonetheless, We appreciate you contributing to Homebrew :beers:

PR #40339, opened 2015-06-03
mosh: update formula for now-generated scripts/mosh

### LINKED ISSUE #40338: mosh HEAD fails to build because of a now-generated script that Homebrew edits
I'm the currently-active mosh maintainer.

cgull:
Mosh used to have its startup script at scripts/mosh.  We now edit in
build information, so it's been moved to scripts/mosh.pl and the
Makefile generates scripts/mosh with the appropriate edits.  Homebrew
does its own edits before the build, so tell it about this.  Once
1.2.5 is released, the formula can lose the conditionals again.

Closes #40338.

### DIFF
--- a/Library/Formula/mobile-shell.rb
+++ b/Library/Formula/mobile-shell.rb
@@ -24,8 +24,13 @@ class MobileShell < Formula
 
     # teach mosh to locate mosh-client without referring
     # PATH to support launching outside shell e.g. via launcher
-    inreplace "scripts/mosh", "'mosh-client", "\'#{bin}/mosh-client"
-
+    #
+    # In HEAD, mosh is generated from mosh.pl, This will be in 1.2.5, coming soon.
+    if build.head?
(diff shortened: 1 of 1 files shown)


Library/Formula/mobile-shell.rb
```diff
@@ -24,7 +24,10 @@ def install
 
     # teach mosh to locate mosh-client without referring
     # PATH to support launching outside shell e.g. via launcher
-    inreplace "scripts/mosh", "'mosh-client", "\'#{bin}/mosh-client"
+    #
+    # In HEAD, mosh is generated from mosh.pl, This will be in 1.2.5, coming soon.
+    inreplace "scripts/mosh.pl", "'mosh-client", "\'#{bin}/mosh-client" if build.head?
```
DomT4: Let's use this format:

```ruby
if build.head?
  xxx
else
  xxx
end
```

cgull:
Fixed, squashed, rebased.

tdsmith:
Thanks!

PR #40340, opened 2015-06-03
hashpump: add optional python bindings

DomT4:


DomT4:
Is this kosher in regards to how we handle optional bindings @tdsmith?

tdsmith:
I would tend to always build the bindings and just install them to `prefix` and drop the `:python` dependency. Is that a bad idea?

Library/Formula/hashpump.rb
```diff
@@ -10,13 +10,23 @@ class Hashpump < Formula
...
 
   def install
     bin.mkpath
     system "make", "INSTALLLOCATION=#{bin}",
                    "CXX=#{ENV.cxx}",
                    "install"
+
+    if build.with? "python"
+      # The Python bindings require the headers, which are named generically and
+      # can conflict with other package's headers.
```
tdsmith: I'm not sure what this means. Does installing the python module to `prefix` install a bunch of include files to `prefix/include` or something?

DomT4: The headers include files named such as `SHA1.h` and `SHA256.h` which would likely cause some trouble if dumped in a publicly-available prefix, so the idea is to sandbox those headers whilst still allowing access to the Python bindings.

The main tool doesn't require the headers, but the Python bindings do, so I guess the actual answer to your question is "Yes" :smile_cat:

DomT4:
> I would tend to always build the bindings and just install them to prefix and drop the :python dependency. Is that a bad idea?

Because of the generic header names that could cause some issues, which is why I `libexec`'d and then symlinked selectively. I'm :+1: on removing the option and making it mandatory if you're cool with that.

Library/Formula/hashpump.rb
tdsmith: What command installs the headers?

DomT4: `system "python", *Language::Python.setup_install_args(libexec)`.

[This](https://github.com/bwall/HashPump/blob/master/setup.py) is the upstream `setup.py` file.

tdsmith: So it is. That's a really weird way to install a Python extension! Installing to libexec seems correct.

Since this supports Python 3, you can even add `option "without-python"` and `depends_on :python3 => :optional` and wrap this in Language::Python.each_python if you want.

tdsmith:
related: discussion in https://github.com/bwall/HashPump/pull/11, PR in https://github.com/bwall/HashPump/pull/14

DomT4:
Thanks Tim :birthday:. Should I wait for your PR to be merged and then use it as a patch, or proceed with the `each_python` changes now?

tdsmith:
Up to you; the only thing that'll change is `libexec` vs `prefix`.

DomT4:
I'll sit on it for a day or so, see what happens with your PR. If we can install directly into the `prefix` rather than doing the selective symlinking dance that's a nicer alternative.

DomT4:
What am I doing wrong here?

```ruby
  depends_on "openssl"
  depends_on :python3 => :optional

  # Remove on next release
  patch do
    url "https://patch-diff.githubusercontent.com/raw/bwall/HashPump/pull/14.diff"
    sha256 "47236fed281000726942740002e44ef8bb90b05f55b2e7deeb183d9f708906c1"
  end

  def install
    bin.mkpath
    system "make", "INSTALLLOCATION=#{bin}",
                   "CXX=#{ENV.cxx}",
                   "install"

    Language::Python.each_python(build) do |python, version|
      system python, *Language::Python.setup_install_args(prefix)
    end
  end
```

The optional `python3` bindings are fine, but the `python` ones are never triggered and installed.

tdsmith:
You need to add `option "without-python", "Build without python 2 support"`; `each_python` queries `build.with? "python"` to decide whether it should try and build Python support.

DomT4:
I presume this is okay to ignore from `brew style`?

```
== Library/Formula/hashpump.rb ==
W: 31: 53: Unused block argument - version. If it's necessary, use _ or _version as an argument name to indicate that it won't be used.
```

tdsmith:
:tada:

PR #40342, opened 2015-06-03
ghq 0.6 (new formula)

Decors:


Library/Formula/ghq.rb
```diff
@@ -0,0 +1,43 @@
...
+    ENV["GOPATH"] = buildpath
+    ENV.append_path "PATH", "#{ENV["GOPATH"]}/bin"
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    system "go", "build", "-o", "ghq"
+    bin.install "ghq"
+  end
+
+  test do
+    system bin/"ghq", "--version"
```
bfontaine: Could you add a test that does something more substantial than e.g. `--version` or `--help`? Thanks!

Decors:
Thanks. I've updated.

Library/Formula/ghq.rb
```diff
@@ -0,0 +1,50 @@
...
+  desc "Remote repository management made easy"
+  homepage "https://github.com/motemen/ghq"
+  url "https://github.com/motemen/ghq.git", :tag => "v0.6", :revision => "b6f7aadbeb21ae18972577173ce175af83ce239d"
+  sha256 "d8ab0ef7386647b30210aabc79bf0391ea5c2887be0d4c6a3fb58b8c31eb9b24"
+
+  head "https://github.com/motemen/ghq.git", :branch => "master"
+
+  depends_on "go" => :build
+
+  option "without-completions", "Disable zsh completions"
```
mikemcquaid: Probably just kill this option; there's no harm in installing them.

Library/Formula/ghq.rb
```diff
@@ -0,0 +1,50 @@
...
+    bin.install "ghq"
+
+    if build.with? "completions"
+      zsh_completion.install "zsh/_ghq"
+    end
+  end
+
+  test do
+    s = shell_output("#{bin}/ghq root")
+    assert_match /#{Regexp.escape("#{testpath}/.ghq")}/, s
```
mikemcquaid: You can just use `"#{testpath}/.ghq"` as a string argument to save the escape. Also, might want to just add `shell_output` on the same line.

Library/Formula/ghq.rb
```diff
@@ -0,0 +1,50 @@
...
+    ENV["GOPATH"] = buildpath
+    ENV.append_path "PATH", "#{ENV["GOPATH"]}/bin"
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    system "go", "build", "-o", "ghq"
+    bin.install "ghq"
+
+    if build.with? "completions"
+      zsh_completion.install "zsh/_ghq"
+    end
```
mikemcquaid: Even although you're removing this: in future `zsh_completion.install "zsh/_ghq" if build.with? "completions"` is a bit nicer

bfontaine:
Could someone from @Homebrew/owners gives their :+1: / :-1: here since it’s a new formula?

mikemcquaid:
Sorry, I was unclear before: let's install the ZSH completions but not offer an option to not install them (if that makes sense).

DomT4:
Can you re-add the `zsh` completions here please? Thanks!

Library/Formula/ghq.rb
```diff
@@ -0,0 +1,43 @@
+require "language/go"
+
+class Ghq < Formula
+  desc "Remote repository management made easy"
+  homepage "https://github.com/motemen/ghq"
+  url "https://github.com/motemen/ghq.git", :tag => "v0.6", :revision => "b6f7aadbeb21ae18972577173ce175af83ce239d"
+  sha256 "d8ab0ef7386647b30210aabc79bf0391ea5c2887be0d4c6a3fb58b8c31eb9b24"
+
+  head "https://github.com/motemen/ghq.git", :branch => "master"
```
DomT4: The `:branch` shouldn't be necessary here? We assume `master` is the default.

DomT4:
Nearly there on this. One last style nit from the test-bot:

```
==> brew style ghq

Error: 1 problem in 1 formula

==> audit problems
ghq:
 * `option`(line 13) should be put before `depends_on`(line 11)
```

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/20763411e16515dc2c0ce1b50565e7241ba761aa. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #38060, opened 2015-03-25
giter8 0.6.8

justjoheinz:
update version to 0.6.8

PR #40358, opened 2015-06-03
giter8 0.6.8

bosr:
Hi,

upstream was updated to 0.6.8 several months ago, and the jar is available at maven repositories. The formula works well as is.

PS: upstream was later updated to 0.6.9, but the jar does not seem to be available yet.

### DIFF
--- a/Library/Formula/giter8.rb
+++ b/Library/Formula/giter8.rb
@@ -3,8 +3,13 @@ require "formula"
 class Giter8 < Formula
   homepage "https://github.com/n8han/giter8"
   url "https://repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/sbt-launch/0.13.7/sbt-launch.jar"
-  sha1 "b407b2a76ad72165f806ac7e7ea09132b951ef53"
-  version "0.6.6"
+  sha256 "9673ca4611e6367955ae068d5888f7ae665ab013c3e8435ffe2ca94318c6d607"
+  # note: because sbt-launch dynamically downloads giter8 from the maven repos
+  #   at first run (using the launchconfig mechanism), and not when the formula
+  #   is installed, the above url and sha256 are related to sbt-launch.jar.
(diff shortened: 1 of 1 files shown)


DomT4:
Why does the URL/checksum not change, out of interest?

bosr:
Good question: it's because the sha1 is that of sbt-launcher.jar. The actual giter8 jar is downloaded at first launch using information from giter8.launchconfig. As the formula creates this launchconfig file, (in which the version is interpolated), there is almost nothing to do. 

Romain

DomT4:
Thanks for the explanation. Could you possibly add a one or two line comment above the `url` just to say that the url/checksum doesn't need to change because it's a downloader/etc? It's confused us a few times now :smile_cat:.

If you wanted to bump the checksum to `sha256` when you do that'd be awesome - `sha256` is our preferred checksum now.

bosr:
Ok, with pleasure (probably tomorrow). 
Best regards,

Romain

DomT4:
> Ok, with pleasure (probably tomorrow).

Thanks! Tomorrow is fine. No rush tonight.

As a side note, there has been a [stated interest](https://github.com/Homebrew/homebrew/pull/38060#issuecomment-87324677) in actually downloading the end-product itself rather than farming it out. I'm not super-fussed about enforcing it right now in this PR, but longer-term it'd be cool for the Homebrew formula to download whatever is necessary in-formula rather than once the user runs the executable afterwards. Not sure how feasible that is, but it'd be great to get done if you have any ideas.

bosr:
I agree, I would personally prefer that too (even if I currently have no idea about how to do this). 
Cheers,

Romain

bosr:
Hi Dominik,

updated and rebased on master

thank you for the kind request and discussion.

mikemcquaid:

bosr:
@mikemcquaid and @DomT4: warm thanks for the kind words

PR #40365, opened 2015-06-04
Pathname#version: support bottle version

### LINKED ISSUE #23184: brew cleanup making some strange mistakes
There's something very weird going on with `brew cleanup`, specifically, the way it determines which files in the cache should be thrown away as "old".

It's generally pretty accurate about this.  However, since I recently fetched *everything*, this afforded some opportunity to test `cleanup`.

There are some (to me) inexplicable trashings of files in the cache that can't possibly be thought to be "old" or "previous versions" of something else in the cache.

For instance, `brew cleanup` will throw away:

````
Would remove: /Library/Caches/Homebrew/abcmidi-2013-09-15.zip
Would remove: /Library/Caches/Homebrew/afsctool-1.6.4
Would remove: /Library/Caches/Homebrew/android-sdk--completion-.bash
Would remove: /Library/Caches/Homebrew/aws-sns-cli-2013-09-27.zip
Would remove: /Library/Caches/Homebrew/boost-build-2011.12-svn.tar.bz2
Would remove: /Library/Caches/Homebrew/crush-tools-2013-04.tar.gz
Would remove: /Library/Caches/Homebrew/daemonlogger-1.2.1
Would remove: /Library/Caches/Homebrew/denominator-4.3.2
Would remove: /Library/Caches/Homebrew/discodex-2012-01-10.tar.gz
Would remove: /Library/Caches/Homebrew/easy-git-1.7.3
Would remove: /Library/Caches/Homebrew/jsmin-2013-03-29.tar.gz
Would remove: /Library/Caches/Homebrew/kismet-2011-03-R2.tar.gz
Would remove: /Library/Caches/Homebrew/lha-1.14i-ac20050924p1.tar.gz
Would remove: /Library/Caches/Homebrew/libpano-13-2.9.18.tar.gz
Would remove: /Library/Caches/Homebrew/lorem-0.6.1
Would remove: /Library/Caches/Homebrew/neo4j-community-1.9.4-unix.tar.gz
Would remove: /Library/Caches/Homebrew/plowshare-4.GIT-1c5013f.tar.gz
Would remove: /Library/Caches/Homebrew/poco-1.4.6p1-all.tar.bz2
Would remove: /Library/Caches/Homebrew/repo-1.20
Would remove: /Library/Caches/Homebrew/rpm2cpio-1.3
Would remove: /Library/Caches/Homebrew/term-2.1
Would remove: /Library/Caches/Homebrew/tmpreaper-1.6.13_nmu1.tar.gz
Would remove: /Library/Caches/Homebrew/zpython-00-5.0.2.tar.bz2
````

Some of these make sense as "false positives", for example, the android-sdk completion file.  OK, so `cleanup` thinks that's an outdated form of `android-sdk-22.2.1.zip`.  You can't win 'em all.

But why on earth is it throwing away `denominator-4.3.2`?  There is nothing in the cache named anything remotely similar to that.  Why is it throwing a way `boost-build-2011.12-svn.tar.bz2`?  Why is it throwing away `lorem`?  `neo4jcommunity`?  `kismet`?

Most of the things on that list are just bizarre errors.  Only the android one and `aws-sns-blahblah` make any conceivable sense to me as false positives.

xu-cheng:
Partially fix #23184.

### DIFF
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -197,7 +197,16 @@ class Pathname
 
   def version
     require 'version'
-    Version.parse(self)
+    if BOTTLE_EXTNAME_RX === to_s
+      begin
+        receipt = Utils.popen_read("tar", "-tzf", to_s, "*/*/INSTALL_RECEIPT.json").chomp
+        Version.new(receipt.split("/", 3)[1])
+      rescue
+        Version.parse(self)
+      end
+    else
+      Version.parse(self)
+    end
   end
 
   def compression_type


mikemcquaid:
Feels weird to have this in `Pathname` but the approach itself looks good.

mikemcquaid:
Sorry, any chance this could be moved outside of Pathname?

xu-cheng:
I guess we can. But I'm not sure where this method has been used other than [cleanup](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/cleanup.rb#L81). So if I remove this method, I may break things.

mikemcquaid:
Sorry, I mean specifically the `if/else` you've added here. It feels like it's not really the concern of the Pathname class and could maybe go into `Version` somewhere instead?

xu-cheng:
@mikemcquaid See #40378

PR #40366, opened 2015-06-04
tmux: moved to github

dplarson:
tmux has officially moved to Github.com (from SourceForge.net).

Also, this commit modernizes the tmux formula, including:
- adding a `desc` field
- switching to sha256 checksums

NOTE: although the tmux/tmux repo includes release tarballs, they appear
to be "bad" in that they are missing a proper `configure` script. Hence
the usage of the tmux/releases repo as the tarball source.

### DIFF
--- a/Library/Formula/tmux.rb
+++ b/Library/Formula/tmux.rb
@@ -1,9 +1,7 @@
-require 'formula'
-
 class Tmux < Formula
-  homepage 'http://tmux.sourceforge.net'
-  url 'https://downloads.sourceforge.net/project/tmux/tmux/tmux-2.0/tmux-2.0.tar.gz'
-  sha1 '977871e7433fe054928d86477382bd5f6794dc3d'
+  homepage "https://tmux.github.io/"
+  url "https://github.com/tmux/tmux/releases/download/2.0/tmux-2.0.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/tmux.rb
```diff
@@ -1,9 +1,8 @@
-require 'formula'
-
 class Tmux < Formula
-  homepage 'http://tmux.sourceforge.net'
-  url 'https://downloads.sourceforge.net/project/tmux/tmux/tmux-2.0/tmux-2.0.tar.gz'
-  sha1 '977871e7433fe054928d86477382bd5f6794dc3d'
+  desc "a terminal multiplexer"
```
mikemcquaid: Please omit this for now, we're going to mass-merge them later.

dplarson:
I removed the `desc` line (per @mikemcquaid 's comment).

dunn:
Looks like they got rid of the `releases` repo, homepage now links to https://github.com/tmux/tmux/releases/download/2.0/tmux-2.0.tar.gz

dplarson:
Good catch @dunn. I've updated to use the tmux/tmux repo for the tarball.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/8c861b64d0f63d69de7673ad638465753df7ed4e. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #40367, opened 2015-06-04
zsh 5.0.8

maxteufel:
Also, update the formula to use SHA-2-256 checksums.


mikemcquaid:

abraithwaite:
Looks like their FAQ still needs to be updated:

http://zsh.sourceforge.net/FAQ/zshfaq01.html#l6

It has been updated though:

http://www.zsh.org/mla/workers/2015/msg01269.html

< being paranoid about sourceforge.

PR #40372, opened 2015-06-04
Revert "upgrade: announce usage of --all."

mikemcquaid:
This reverts commit 9032f165bec3d0c7452169093fab32578816043b.

This seems to have been universally unpopular so let's revert it. CC @Homebrew/owners for thoughts.

### DIFF
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -5,14 +5,7 @@ module Homebrew
   def upgrade
     Homebrew.perform_preinstall_checks
 
-    if ARGV.include?("--all") || ARGV.named.empty?
-      unless ARGV.include? "--all"
-        opoo <<-EOS.undent
-          brew upgrade with no arguments will change behaviour soon!
-          It currently upgrades all formula but this will soon change to require '--all'.
-          Please update any workflows, documentation and scripts!
-        EOS
-      end
+    if ARGV.named.empty?
       outdated = Homebrew.outdated_brews(Formula.installed)
       exit 0 if outdated.empty?
     elsif ARGV.named.any?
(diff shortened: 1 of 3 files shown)


bfontaine:
Could we do the opposite, by add a flag like `--safe` or something which prevents `brew upgrade` to upgrade everything if a formulae isn’t given?

mikemcquaid:
I dunno if that's be useful unfortunately.

DomT4:
At this point I kind of feel that reverting it will cause more confusion than good, and I don't believe the problems you identified previously about upgrading everything at once have changed.

Given the extremely long deprecation period you identified before the current behaviour isn't actually changing for rather a long period, there's still a lot of adjustment room.

If you feel like it's the right thing to do I'm happy with that though - I'm not super attached to the new (or old) behaviour. Whichever serves folks best.

bfontaine:
Could we use an environment variable? E.g. “use `--all` or set `HOMEBREW_WHATEVER=1`”, so people who aren’t happy with this `--all` flag can use the environment variable to use the old behavior. Or the opposite: set `HOMEBREW_SOMETHING=1` to disable `brew upgrade` (i.e. get the new behavior).

jacknagel:
FWIW, I'm in favor of this, as I wasn't a fan of the initial change.

mistydemeo:
I'm also in favour of reverting.

If there's enough demand we could look at adding a `HOMEBREW_SAFE_UPGRADE` or something to get the new behaviour, but it should probably be opt-in rather than opt-out.

xu-cheng:
> At this point I kind of feel that reverting it will cause more confusion than good

I kinda have the same feeling.

mikemcquaid:
FWIW `--all` will just be a no-op.

PR #40377, opened 2015-06-04
rabbitmq 3.5.3

bartj3:


bfontaine:
The audit fails:

```
        Error: 2 problems in 1 formulae
rabbitmq:
 * Non-executables were installed to "/usr/local/Cellar/rabbitmq/3.5.3/bin"
The offending files are:
  /usr/local/Cellar/rabbitmq/3.5.3/bin/start.boot
  /usr/local/Cellar/rabbitmq/3.5.3/lib/crypto-3.5/priv/lib/crypto.so
```

mikemcquaid:
@bfontaine At some point we need to build this from source which will fix the latter.

bartj3:
I'd love to help out but I wouldn't know what to do. I don't know enough about RabbitMQ to know what the start.boot file does and I don't know enough about Homebrew to know how building from source would work and how it would fix the OpenSSL warning.

Are there any other packages that had the OpenSSL issue where building from source fixed it? Maybe I could read up on those and see if I can help out with anything.

DomT4:
As a short term work-around until we work out if/how we build from source to resolve OpenSSL problems, we could dump this in `libexec` and then `bin.write_exec_script` each of the executables in `libexec/"bin"`. The various other `prefix` lines would need changing to match, though.

mikemcquaid:
Pulling as-is because the current formula already has these problems.

mikemcquaid:

PR #40378, opened 2015-06-04
cleanup: handle bottle version

xu-cheng:


mikemcquaid:
:heart_eyes:

jacknagel:
I don't really see a reason that `Version.parse` needs to be responsible for this. Can't we match against `BOTTLE_EXTNAME_RX` before calling it and move this logic out?

My hope was we could entirely remove the pathname crap from Version.parse eventually.

jacknagel:
For example, if we're trying to fix the cache, can't the cache code check if the filename is a bottle and then call the appropriate logic?

xu-cheng:
En, I will try to come to a new approach.

> My hope was we could entirely remove the pathname crap from Version.parse eventually.

I have the similar thought. Also I think we can merge `Version#parse` and `Version#_parse`. There seems to be no reason to have two methods since `bottle_version` is removed.

xu-cheng:
Comment on new version PR?

xu-cheng:
Any comments?

xu-cheng:
@jacknagel @mikemcquaid Good to merge?

mikemcquaid:
:+1:

PR #40381, opened 2015-06-04
Pcap_DNSProxy v0.4.2 (new formula)

blackgear:


Library/Formula/pcap_dnsproxy.rb
```diff
@@ -0,0 +1,41 @@
...
+    xcodebuild "-project", "./Source/KeyPairGenerator.xcodeproj", "-target", "KeyPairGenerator", "-configuration", "Release", "SYMROOT=build"
+    xcodebuild "-project", "./Source/Pcap_DNSProxy.xcodeproj", "-target", "Pcap_DNSProxy", "-configuration", "Release", "SYMROOT=build"
+
+    bin.install "Source/build/Release/KeyPairGenerator"
+    bin.install "Source/build/Release/Pcap_DNSProxy"
+    (etc/"pcap_DNSproxy").install Dir["Source/ExampleConfig/*.{ini,txt,plist}"]
+  end
+
+  def caveats
+    <<-EOS.undent
```
xu-cheng: We don't need such caveats. Instead you should use [`plist` method](https://github.com/Homebrew/homebrew/blob/master/Library/Contributions/example-formula.rb#L448)

Library/Formula/pcap_dnsproxy.rb
```diff
@@ -0,0 +1,41 @@
...
+  homepage "https://github.com/chengr28/Pcap_DNSProxy"
+  url "https://github.com/chengr28/Pcap_DNSProxy/archive/v0.4.2.tar.gz"
+  sha256 "0737b16283e84bc94f63066f448ba84416141b8d5a86b5e01f1f97164d9d44f2"
+  head "https://github.com/chengr28/Pcap_DNSProxy.git"
+
+  depends_on :xcode => :build
+  depends_on "libsodium" => :build
+
+  def install
+    ln_s "#{HOMEBREW_PREFIX}/lib/libsodium.a", "#{buildpath}/Source/LibSodium/LibSodium_Mac.a"
```
xu-cheng: Using `"#{Formula["libsodium"].opt_lib}/libsodium.a"` instead.

BlackGear:
This program listen port 53 by default, and it need the root privilege to do that. the plist in `~/Library/LaunchAgents/` can't listen port 53 but `/Library/LaunchDaemons` can.

The plist method will generate a default instruction: add the plist to `~/Library/LaunchAgents/` and run it with launchctl which only cause a error.

I'm not sure how to deal with this problem but only notice it in caveats.

DomT4:
^ You want a line like this:

```ruby
plist_options :startup => true, :manual => "sudo #{HOMEBREW_PREFIX}/opt/pcap_dnsproxy/bin/Pcap_DNSProxy -c #{HOMEBREW_PREFIX}/etc/pcap_dnsproxy/"
```

That'll instruct people to chown the `plist` to root and move it into `/Library/LaunchDaemons`.

BlackGear:
It works, wow! Thanks DomT4.

Library/Formula/pcap_dnsproxy.rb
```diff
@@ -0,0 +1,51 @@
...
+    xcodebuild "-project", "./Source/KeyPairGenerator.xcodeproj", "-target", "KeyPairGenerator", "-configuration", "Release", "SYMROOT=build"
+    xcodebuild "-project", "./Source/Pcap_DNSProxy.xcodeproj", "-target", "Pcap_DNSProxy", "-configuration", "Release", "SYMROOT=build"
+
+    bin.install "Source/build/Release/KeyPairGenerator"
+    bin.install "Source/build/Release/Pcap_DNSProxy"
+    (etc/"pcap_DNSproxy").install Dir["Source/ExampleConfig/*.{ini,txt}"]
+  end
+
+  test do
+    system bin/"Pcap_DNSProxy", "--version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/pcap_dnsproxy.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  plist_options :startup => true, :manual => "sudo #{HOMEBREW_PREFIX}/opt/pcap_dnsproxy/bin/Pcap_DNSProxy -c #{HOMEBREW_PREFIX}/etc/pcap_dnsproxy/"
```
mikemcquaid: Can this software be made to work without `sudo`? If so that's more desirable.

Library/Formula/pcap_dnsproxy.rb
```diff
@@ -0,0 +1,51 @@
...
+  homepage "https://github.com/chengr28/Pcap_DNSProxy"
+  url "https://github.com/chengr28/Pcap_DNSProxy/archive/v0.4.2.tar.gz"
+  sha256 "0737b16283e84bc94f63066f448ba84416141b8d5a86b5e01f1f97164d9d44f2"
+  head "https://github.com/chengr28/Pcap_DNSProxy.git"
+
+  depends_on :xcode => :build
+  depends_on "libsodium" => :build
+
+  def install
+    ln_s "#{Formula["libsodium"].opt_lib}/libsodium.a", "#{buildpath}/Source/LibSodium/LibSodium_Mac.a"
```
mikemcquaid: Is this required or is there e.g. a vendored version that we can use instead?

Library/Formula/pcap_dnsproxy.rb
BlackGear: In fact, the --version is add by author just last release since I made a issue that I need something to be an exam...

Library/Formula/pcap_dnsproxy.rb
BlackGear: This software listen a localhost port and work as a DNS server, it proxy the DNS resolve request and filter the polluted DNS response to anti DNS spooning. Then set the DNS server to 127.0.0.1, we can avoid DNS spooning.

Sure we can use it without `sudo`, but if you want to listen port < 1024, you need a root privilege.
Unfortunately you can't use port other than 53 as OS X's DNS server in system preference.

This is why the `sudo` is necessary.

Library/Formula/pcap_dnsproxy.rb
BlackGear: libsodium.a is required for it's build-in DNSCurve protocol support, which is a core function to avoid DNS spooning.

Library/Formula/pcap_dnsproxy.rb
mikemcquaid: Ok, thanks!

Library/Formula/pcap_dnsproxy.rb
xu-cheng: This should be a better test:

```ruby
system bin/"KeyPairGenerator"
assert File.exist? "KeyPair.txt"
```

Library/Formula/pcap_dnsproxy.rb
xu-cheng: It seems that upstream already vendored the [libsodium source](https://github.com/chengr28/Pcap_DNSProxy/tree/master/Source/LibSodium). Can you utilize this vendored version instead?

blackgear: @xu-cheng Only ".h" files are vendored in that path, while ".a" file is needed.

I've discussed with the author, he says "I think it's not right to vendor a pre-build static lib in my source code, the package manager system should deal with it." (see: https://github.com/chengr28/Pcap_DNSProxy/issues/33 discussed in Chinese)

xu-cheng: It just seems to be a little weird to have a static build time only dependency. However, I'm fine to merge this as it is if you update the test and rebase/squash the commit message and @mikemcquaid doesn't object this.

mikemcquaid: Why does he vendor the headers? If it's using those headers and not the headers of our formula this may break when `libsodium` is upgraded.

blackgear: So maybe I should also use the headers in homebrew's `libsodium` like:
```
    ln_s "#{Formula["libsodium"].opt_include}/*", "#{buildpath}/Source/LibSodium/"
```
 or can I specifc the version of `libsodium`?

mikemcquaid: Yeh, that'd be better. You can't specify the version of `libsodium`. I'd advise you to tell the maintainer about this too in case they want to take a different approach in future.

blackgear:
- [x] Use the headers in homebrew's `libsodium ` instead of vendored one.
- [x] A new test added to test if the DNS function really work.
- [x] Squash commit logs into one.

DomT4:
Test is failing on ML:

```
==> brew test --verbose pcap_dnsproxy					 FAILED
Testing pcap_dnsproxy
Error: pcap_dnsproxy: failed
undefined method `spawn' for #<Formulary::FormulaNamespace9eeaf48a4f0ed9e705eff3facc2d6267::PcapDnsproxy:0x10e60c818>
/usr/local/Library/Formula/pcap_dnsproxy.rb:30:in `test'
/usr/local/Library/Homebrew/formula.rb:794:in `run_test'
/usr/local/Library/Homebrew/formula.rb:790:in `run_test'
/usr/local/Library/Homebrew/test.rb:28
/System/Library/Frameworks/Ruby.framework/Versions/1.8/usr/lib/ruby/1.8/timeout.rb:67:in `timeout'
/usr/local/Library/Homebrew/test.rb:27
```

blackgear:
Oh no， spawn is introduced to ruby since 1.9, I have to use fork & exec instead...

Library/Formula/pcap_dnsproxy.rb
```diff
@@ -0,0 +1,63 @@
...
+    cp Dir[etc/"pcap_dnsproxy/*"], testpath/"pcap_dnsproxy/"
+    inreplace testpath/"pcap_dnsproxy/Config.ini" do |s|
+      s.gsub! /Hosts Only.*/, "Hosts Only = 1"
+      s.gsub! /Listen Port.*/, "Listen Port = 9999"
+    end
+    pid = fork do
+        exec bin/"Pcap_DNSProxy", "-c", testpath/"pcap_dnsproxy/", [:out, :err]=>"/dev/null"
+    end
+    system "dig", "google.com", "@127.0.0.1", "-p", "9999", "+short"
+    Process.kill "TERM", pid
```
xu-cheng: Let's do this:
```ruby
pid = fork { system bin/"Pcap_DNSProxy", "-c", testpath/"pcap_dnsproxy/" }
begin
  system "dig", "google.com", "@127.0.0.1", "-p", "9999", "+short"
ensure
  Process.kill 9, pid
  Process.wait pid
end
```

blackgear:
```
pid = fork { system bin/"Pcap_DNSProxy", "-c", testpath/"pcap_dnsproxy/" }
...
Process.kill 9, pid
```
Can't kill the Pcap_DNSProxy's process, I use:
`pid = fork { exec bin/"Pcap_DNSProxy", "-c", testpath/"pcap_dnsproxy/", [:out, :err]=>"/dev/null" }`
instead.
The new code can ensure the Pcap_DNSProxy killed whether `dig` succeed or not.

blackgear:
Now it's ready to be merge.

xu-cheng:

PR #40382, opened 2015-06-04
Make tmux formula point to GitHub instead of SourceForge

danielcotton:
tmux has been moved to GitHub, we might as well change it now before the HEADs start to deviate.


danielcotton:
Whoops, didn't see PR #40366
Closing

PR #40383, opened 2015-06-04
Update pmd formula to 5.3.2

champo:


Library/Formula/pmd.rb
```diff
@@ -2,23 +2,27 @@
...
   def install
-    rm Dir["bin/*.{bat,cmd,dll,exe}"]
+    system "mvn", "clean", "package"
 
-    doc.install "LICENSE", "NOTICE", "ReadMe.txt", "licences", Dir["docs/*"]
-    libexec.install "bin", "etc", "lib"
+    doc.install "LICENSE", "NOTICE", "README.md"
 
-    bin.install_symlink "#{libexec}/bin/run.sh" => "pmd"
+    libexec.install "pmd-dist/target/pmd-bin-5.3.2/pmd-bin-5.3.2/bin", "pmd-dist/target/pmd-bin-5.3.2/pmd-bin-5.3.2/lib"
```
DomT4: Let's de-version this so we don't have to change this line every update. `pmd-bin-#{version}` and so on.

Library/Formula/pmd.rb
```diff
@@ -2,23 +2,27 @@
...
 
-    # the run script references paths which don't account for the
-    # file structure of this brew.
+    bin.install_symlink "#{libexec}/bin/run.sh" => "pmd"
     inreplace "#{libexec}/bin/run.sh", "${script_dir}/../lib", "#{libexec}/lib"
   end
 
   def caveats; <<-EOS.undent
+    The install process needs maven3. It's not listed as a dependency since you may have a specific version installed. If you don't, do: brew install maven
```
DomT4: This is handled automatically by the `:build` tag. You can remove the caveat.

champo: Right, I wrote that comment earlier on and it's not accurate anymore.

Library/Formula/pmd.rb
```diff
@@ -2,23 +2,26 @@
...
 class Pmd < Formula
   homepage "http://pmd.sourceforge.net/"
-  url "https://downloads.sourceforge.net/project/pmd/pmd/5.1.3/pmd-src-5.1.3.zip"
-  sha1 "da6023d73e078444a7b76e8a7a0cfbeb421221e4"
+  url "https://downloads.sourceforge.net/project/pmd/pmd/5.3.2/pmd-src-5.3.2.zip"
+  sha1 "036cdb7583de3ea0d24e7486bf7b9f9f14bfb1ec"
+
+  depends_on "maven" => :build
+
+  patch :DATA
```
bfontaine: Why is this patch required? If it’s not Homebrew-specific, could you report it upstream and add a link to the upstream issue in a comment? In any case, could you add a comment to explain why is it necessary? Thanks!

Library/Formula/pmd.rb
```diff
@@ -2,23 +2,26 @@
...
+    libexec.install "pmd-dist/target/pmd-bin-#{version}/pmd-bin-#{version}/bin", "pmd-dist/target/pmd-bin-#{version}/pmd-bin-#{version}/lib"
 
-    # the run script references paths which don't account for the
-    # file structure of this brew.
+    bin.install_symlink "#{libexec}/bin/run.sh" => "pmd"
     inreplace "#{libexec}/bin/run.sh", "${script_dir}/../lib", "#{libexec}/lib"
   end
 
   def caveats; <<-EOS.undent
+    Once pmd is installed, maven isn't needed anymore.
```
bfontaine: This is already handled by the `depends_on "maven" => :build`, which means `maven` is only needed for the build.

champo: Is it? I just tried:

    $ brew uninstall maven
    Uninstalling /usr/local/Cellar/pmd/5.3.2... (39 files, 33M)
    $ brew uninstall pmd
    Uninstalling /usr/local/Cellar/maven/3.3.3... (92 files, 9.3M)
    $ brew install pmd
    ... 
    $ readlink $(which mvn)
    ../Cellar/maven/3.3.3/bin/mvn

Given this and that maven versions conflict, I think it's a fair caveat.

bfontaine:

Library/Formula/pmd.rb
```diff
@@ -2,23 +2,28 @@
...
-  url "https://downloads.sourceforge.net/project/pmd/pmd/5.1.3/pmd-src-5.1.3.zip"
-  sha1 "da6023d73e078444a7b76e8a7a0cfbeb421221e4"
+  url "https://downloads.sourceforge.net/project/pmd/pmd/5.3.2/pmd-src-5.3.2.zip"
+  sha1 "036cdb7583de3ea0d24e7486bf7b9f9f14bfb1ec"
+
+  depends_on "maven" => :build
+
+  # The mvn package target produces a .zip with all the jars needed for PMD
+  # This patch makes it produce a folder instead so installing the jars is straightforward
+  patch :DATA
```
mikemcquaid: I'd rather we just unzipped it instead of patching.

champo: I looked into that, and opted against it since there was no mention of that being possible. I tried using a resource for it, but internally it uses downloaders that assume remote paths. I believe this is the cleanest, simplest solution.

mikemcquaid: I mean we could unzip manually in the formula. I'm afraid we're not going to accept a patch like this that can never be accepted upstream.

champo: I'm not quite sure I understand the rationale behind not using the patch for this. What are patches for then?

mikemcquaid: Temporary compilation or functionality fixes that have been submitted to (and usually accepted) by upstream.

DomT4:
@mikemcquaid Are you happy with this?

Library/Formula/pmd.rb
mikemcquaid: Please remove it, thanks.

Library/Formula/pmd.rb
```diff
@@ -39,3 +42,4 @@ def caveats; <<-EOS.undent
       "rulesets/java/basic.xml", "-f", "textcolor", "-l", "java"
   end
 end
+
```
mikemcquaid: Don't need this blank line.

Library/Formula/pmd.rb
```diff
@@ -2,23 +2,26 @@
...
-    rm Dir["bin/*.{bat,cmd,dll,exe}"]
+    system "mvn", "clean", "package"
 
-    doc.install "LICENSE", "NOTICE", "ReadMe.txt", "licences", Dir["docs/*"]
-    libexec.install "bin", "etc", "lib"
+    doc.install "LICENSE", "NOTICE", "README.md"
 
-    bin.install_symlink "#{libexec}/bin/run.sh" => "pmd"
+    # The mvn package target produces a .zip with all the jars needed for PMD
+    with_system_path { safe_system 'unzip', '-qq', buildpath/"pmd-dist/target/pmd-bin-#{version}.zip" }
```
mikemcquaid: Don't need the `with_system_path` here. Also might want to use `tar` instead just to be slightly more like what we normally do.

Library/Formula/pmd.rb
```diff
@@ -2,23 +2,26 @@
...
-  url "https://downloads.sourceforge.net/project/pmd/pmd/5.1.3/pmd-src-5.1.3.zip"
-  sha1 "da6023d73e078444a7b76e8a7a0cfbeb421221e4"
+  url "https://downloads.sourceforge.net/project/pmd/pmd/5.3.2/pmd-src-5.3.2.zip"
+  sha1 "036cdb7583de3ea0d24e7486bf7b9f9f14bfb1ec"
+
+  depends_on "maven" => :build
 
   def install
-    rm Dir["bin/*.{bat,cmd,dll,exe}"]
+    system "mvn", "clean", "package"
```
mikemcquaid: Why the `clean`?

champo: It's probably unnecessary, but there's no harm in having either.

Library/Formula/pmd.rb
champo: I'm sorry, but why? It's fair caveat, and I see no harm in having it. maven users have version compatibility problems.

Library/Formula/pmd.rb
champo: I copied and pasted that line from the resource downloaders implementation. I can modify it, but this follows what you do.

jacknagel: with_system_path is only an internal thing. it's not needed inside a formula.

Library/Formula/pmd.rb
jacknagel: Because it's true of any build-time dependency for any formula. What makes this one special?

mikemcquaid: The reason was already explained, anyway. Please remove it or close this PR.

champo: @mikemcquaid the reason didn't seem reasonable since the it's not only build time. From @jacknagel answer I get that this is expected from brew. I'll remove it in light of that. Sorry if I'm getting on your nerves, and thanks for your time.

mikemcquaid: That's OK, thanks!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/7bfd9e437dd20e241645f6b02ea61f443ea9075b. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just for future reference commits should be squashed on final push, `sha256` is the preferred checksum now and the commit message for version bumps should be a simple `pmd 5.3.2`, Thanks!

PR #40387, opened 2015-06-04
Tmux moved away from sourceforge.net to github.com

jottr:
Updated the formula accordingly.


mistydemeo:
There's a PR already open, #40366. Thanks!

PR #40388, opened 2015-06-04
cdrtools devel 3.01a29

fbrosson:
Updating the formula for cdrtools 3.01a29 which was released this morning.
Switching from sha1 to sha256 and dropping the "require 'formula'" header.

Regarding the patch about the HAVE_PRAGMA_WEAK test; may I drop it for devel?
If I'm not mistaken this patch is a workaround for a bug in Clang 3.1 only and all newer releases of Clang do not have that bug.
And since Homebrew is already using LLVM 3.6.1 I guess Clang 3.1 is now far behind us.
Just drop me a line if you agree with dropping the patch for devel and I'll be happy to get rid of it ;-)
Thanks!

### DIFF
--- a/Library/Formula/cdrtools.rb
+++ b/Library/Formula/cdrtools.rb
@@ -1,15 +1,13 @@
-require 'formula'
-
 class Cdrtools < Formula
-  homepage 'http://cdrecord.org/'
+  homepage "http://cdrecord.org/"
 
   stable do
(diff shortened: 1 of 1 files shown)


DomT4:
> If I'm not mistaken this patch is a workaround for a bug in Clang 3.1 only and all newer releases of Clang do not have that bug.

Is keeping the patch useful for older systems @mistydemeo?

DomT4:
Can you rebase this against the master branch here? Thanks!

DomT4:
Ping?

PR #40389, opened 2015-06-04
conman 0.2.7

fly:
adds conman formula

### DIFF
--- a/Library/Formula/conman.rb
+++ b/Library/Formula/conman.rb
@@ -1,0 +1,24 @@
+class Conman < Formula
+  desc "ConMan: The Console Manager"
+  homepage "https://github.com/dun/conman"
+  url "https://github.com/dun/conman/archive/conman-0.2.7.tar.gz"
+  sha256 "bf96dfbb05e1592355bab3afda4fcb4766a57bca8bc98eb1dd5020f24039d39f"
+
+  depends_on :x11 => :optional
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/conman", "-V"
+    system "#{sbin}/conmand", "-V"
+
+    ENV["DISPLAY"] = "/dev/null"  # set bogus DISPLAY to fake x11 support
+    system "#{bin}/conmen", "-V"
+  end
+end


fly:
`conmen -V` test is failing with:

```
==> /usr/local/Cellar/conman/0.2.7/bin/conmen -V
ERROR: DISPLAY not set.
Error: conman: failed
```

bfontaine:
> ERROR: DISPLAY not set.

You could try setting `ENV["DISPLAY"]`.

fly:
I thought about that, but I'm curious about what the accepted method of testing an optional dependency on x11 is in homebrew. I can certainly set a bogus `DISPLAY` envvar to like `/dev/null` or something.

fly:
ok, i've set a bogus `DISPLAY` envvar to make the test pass.

anything else I need for mergerino?

Library/Formula/conman.rb
```diff
@@ -0,0 +1,24 @@
+class Conman < Formula
+  desc "ConMan: The Console Manager"
+  homepage "https://github.com/dun/conman"
+  url "https://github.com/dun/conman/archive/conman-0.2.7.tar.gz"
+  sha256 "bf96dfbb05e1592355bab3afda4fcb4766a57bca8bc98eb1dd5020f24039d39f"
+
+  depends_on :x11 => :optional
```
mikemcquaid: Does X11 need explicit disabled? This won't be enabled by default but it's installing a binary (`conmen`) that seems to need X11?

Library/Formula/conman.rb
```diff
@@ -0,0 +1,24 @@
...
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/conman", "-V"
+    system "#{sbin}/conmand", "-V"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/conman.rb
fly: i believe `conmen` is the version of `conman` with X11 support, which is why I set x11 as an optional dependency. How would you handle it in this case?

mikemcquaid: There's probably an explicit `--without-x11` flag we can add when `build.without? "x11"` is true.

DomT4:
Ping.

fly:
Sorry, haven't been able to get around to this. I'll work on it this
weekend.

DomT4:
Any update?

DomT4:
Presuming dead from lack of response. Please feel free to fix-up various comments and resubmit the PR if you're still interested, but closing here for now. If the formula requires `x11` for certain functionality it may be a better fit for `homebrew/x11`. Thanks for the submission!

PR #40399, opened 2015-06-05
Ability to compile cryptopp and log4cxx packages in C++11 mode

cawka:
I have a project that is compiled in c++11 mode and depends on these libraries.  WIthout c++11 mode, compilation fails on OSX <= 10.8, because of incompatible STL libraries.

### DIFF
--- a/Library/Formula/log4cxx.rb
+++ b/Library/Formula/log4cxx.rb
@@ -2,8 +2,8 @@ require 'formula'
 
 class Log4cxx < Formula
   desc "Library of C++ classes for flexible logging"
-  homepage 'http://logging.apache.org/log4cxx/index.html'
-  url 'http://www.apache.org/dyn/closer.cgi?path=logging/log4cxx/0.10.0/apache-log4cxx-0.10.0.tar.gz'
+  homepage 'https://logging.apache.org/log4cxx/index.html'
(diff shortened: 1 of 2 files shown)


Library/Formula/log4cxx.rb
```diff
@@ -37,25 +39,137 @@ def install
   end
 end
 __END__
+diff --git a/src/main/cpp/locationinfo.cpp b/src/main/cpp/locationinfo.cpp
+index e76ea29..2e1950c 100644
+--- a/src/main/cpp/locationinfo.cpp
++++ b/src/main/cpp/locationinfo.cpp
```
tdsmith: Can you do something to separate out the new patches from the ones already in the formula? Hosting the new patch outside the formula in a gist should be okay. Has this been submitted to the log4cxx project / has it already been committed?

cawka: Will do. I also forgot to do the same with cryptopp, will do an update soon.

As for patch to log4cxx project.  The issue tracking for log4cxx project is not working right now, but given that the project has not seen a new release since 2008, I would not expect much...

bfontaine: You could try sending a patch to the mailing list if there’s one or to the author, just in case. Last month I got a pull-request accepted on a project that was dead since 2011.

cawka:
I have verified compilation of both log4cxx and cryptopp in c++11 mode on OSX 10.10 and on OSX 10.8

mikemcquaid:

PR #40404, opened 2015-06-05
Added automatic completion install for pidcat

paour:


Library/Formula/pidcat.rb
```diff
@@ -1,6 +1,7 @@
 require "formula"
 
 class Pidcat < Formula
+  depends_on "bash-completion" => :recommended
```
DomT4: Is this necessary? Most bash completions can just be installed without any added dependencies. If it actually needs it, the dependency should go above `def install`, not at the top of the formula here.

Library/Formula/pidcat.rb
```diff
@@ -8,6 +9,10 @@ class Pidcat < Formula
 
   def install
     bin.install "pidcat.py" => "pidcat"
+
+    if build.with? "bash-completion"
+      bash_completion.install "bash_completion.d/pidcat"
```
DomT4: Likewise, if it needs the dependency make this a single line `if` statement, not a block.

paour:
Sorry, I wasn't aware `bash_completion.install` was available regardless of the bash-completion package. I've updated the pull request.

DomT4:
Thanks! Could you possibly squash your commits into one commit? Our test-bot is having trouble doing so automatically.

paour:
Abandonning

paour:
Created a new pull request: #40565

PR #40407, opened 2015-06-05
mongoose: install executable

### LINKED ISSUE #40369: Help with installation of mongoose 5.6
Hi, 

I am relatively inexperience computer user, so please keep any responses as simple as possible.

I am trying to install the web-server mongoose on a IMac running Yosemite. The main purpose of doing this is to use a bioinformatic analysis program called trinotate

I have sucessfully (apparently) installed mongoose-5.6  using homebrew

however when I use the command 'mongoose', I get: 'mongoose: command not found'

When looking at the homebrew installation via: find /usr/local/Cellar/ -name 'mongoose'

I only get:  '/usr/local/Cellar//mongoose'

Upon changing to this directory I get the following files:

INSTALL_RECEIPT.json    README.md        lib
LICENSE            include            share

There appears to be no binary files present, following the installation

Is this a general problem with this version of mongoose, a problem with the installation formula or a problem with my own system? I have run brew doctor and it has flagged up no problems

Does anyone have any (simple) advice to rectify this situation? 

Many thanks

Alan

DomT4:
Would also close #40369.

### DIFF
--- a/Library/Formula/mongoose.rb
+++ b/Library/Formula/mongoose.rb
@@ -10,14 +10,50 @@ class Mongoose < Formula
     sha256 "0ec26e8125679d0b2464a186b5d204035879804ca6ff814e50cc0eaad658fabd" => :mountain_lion
   end
 
+  depends_on "openssl" => :recommended
+
   def install
+    # No Makefile but is an expectation upstream of binary creation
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

PR #40408, opened 2015-06-05
Update Terraform to 0.5.3

dhilton:
Updating Terraform to 0.5.3 release, removing the patch (as it's now merged in master) and moving to the newly promoted AWS go sdk from the AWSLabs github organisation location.

I used `homebrew_go_resources` to create a new `go_resource` for the AWS library to enable a clean build from source (We were getting import errors due to the move from awslabs to aws Github organisations I believe).

I have `bottled` the release locally with no issues.

### DIFF
--- a/Library/Formula/terraform.rb
+++ b/Library/Formula/terraform.rb
@@ -4,7 +4,7 @@ require "language/go"
 class Terraform < Formula
   homepage "https://www.terraform.io/"
   url "https://github.com/hashicorp/terraform/archive/v0.5.3.tar.gz"
-  sha1 "2f1a743272f6b456b29112e350ee00fc21a3f4ad"
+  sha256 "be71d430df5b28deaff815ee775bbb4d8e240b145450d6c027baa6ef0860ca94"
 
   bottle do
     cellar :any
@@ -13,14 +13,12 @@ class Terraform < Formula
     sha256 "f2be116dee2e01a75f192c5088c25f40ce3a7d6f23ea83539f485701e3b907fa" => :mountain_lion
   end
 
-
   depends_on "go" => :build
 
   go_resource "github.com/awslabs/aws-sdk-go" do
     url "https://github.com/aws/aws-sdk-go.git",
         :revision => "43d7c58d0a71c01d98b7881cb9f90047f04f4acd"
   end
-
 
   terraform_deps = %w[
     github.com/Sirupsen/logrus 52919f182f9c314f8a38c5afe96506f73d02b4b2
@@ -74,7 +72,6 @@ class Terraform < Formula
       url "https://#{x}.git", :revision => y
     end
   end
-
 
   go_resource "code.google.com/p/go-uuid" do
     url "https://code.google.com/p/go-uuid", :using => :hg,
--- a/Library/Formula/terraform.rb
+++ b/Library/Formula/terraform.rb
@@ -3,8 +3,8 @@ require "language/go"
 
 class Terraform < Formula
   homepage "https://www.terraform.io/"
-  url "https://github.com/hashicorp/terraform/archive/v0.5.1.tar.gz"
-  sha1 "7df99086b2ef218c0a56841dbd43a054b19903fb"
+  url "https://github.com/hashicorp/terraform/archive/v0.5.3.tar.gz"
+  sha1 "2f1a743272f6b456b29112e350ee00fc21a3f4ad"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Library/Formula/terraform.rb
```diff
@@ -76,6 +75,7 @@ class Terraform < Formula
     end
   end
 
+
```
mikemcquaid: Can you remove this whitespace-only change? Thanks!

Library/Formula/terraform.rb
```diff
@@ -13,19 +13,18 @@ class Terraform < Formula
...
-  end
 
   depends_on "go" => :build
 
+  go_resource "github.com/awslabs/aws-sdk-go" do
+    url "https://github.com/aws/aws-sdk-go.git",
+        :revision => "43d7c58d0a71c01d98b7881cb9f90047f04f4acd"
+  end
+
+
```
mikemcquaid: Only need one empty line here, thanks!

Library/Formula/terraform.rb
```diff
@@ -3,8 +3,8 @@
 
 class Terraform < Formula
   homepage "https://www.terraform.io/"
-  url "https://github.com/hashicorp/terraform/archive/v0.5.1.tar.gz"
-  sha1 "7df99086b2ef218c0a56841dbd43a054b19903fb"
+  url "https://github.com/hashicorp/terraform/archive/v0.5.3.tar.gz"
+  sha1 "2f1a743272f6b456b29112e350ee00fc21a3f4ad"
```
mikemcquaid: Mind making this `sha256`? thanks!

dhilton:
I've read the Jenkins output - but I'm not quite sure why this failed but the other commit succeeded given that the only functionality change was sha1 to sha256. Could someone elaborate why this commit failed?

bfontaine:

bosr:
@bfontaine, @dhilton: cool, thank you, I was also trying to fix this on my own. One remark though: looking at this formula as it was submitted in the past, may be we should also do the following:

- update each dependency to the latest commit as of today: I found it useful in the previous PR to know that all dependencies where in sync.
- remove the bottle sha's because it looks like the bot adds them (I may be wrong)

Do you agree?

If so, to save you some time, I have a copy of Dan's formula with the above remarks: https://github.com/bosr/homebrew/tree/terraform-0.5.3 (builds locally on 10.10).

For updating each package, I wrote this (inelegant) script: https://gist.github.com/8ecd7e53e6ea77551de4

I leave it up to you to use it or not.
Cheers

DomT4:
> remove the bottle sha's because it looks like the bot adds them (I may be wrong)

We leave them there and update them when we pull. Removing and then re-adding them can cause unexpected issues around loss of `revision`, failure for the bottle to upload, etc, so the process is smoother overwriting the existing block when we pull.

bosr:
Thanks!

dhilton:
@bosr the reason I haven't is that I was unsure of the consequences - i.e. this release was built on certain versions of those libraries (as stated in deps/v0-5-3.json ) so we could use that to give us the correct versions of the dependencies at the point this release was crafted.

However, when I run `homebrew_go_resource` it generates a lot smaller list than those given in the formula, which leads me to believe that either:

1) There is a cleanup piece to do on this formula or
2)  That `homebrew_go_resource` doesn't work out the dependencies of a Go package (surprised if this is the case as `go get` will go and get any dependencies listed for a package to build). 

In the name of getting this merged - how about we update the hashes using the ones specified in the json deps file for that version, thus ensuring clear audit trail from the built release and it's dependencies, leaving those others 'as is' - we can then recheck the build and run the tests?

I'd feel happier with that than rather than just updating them all to HEAD - having been burnt by the aws lib move already in this packaging process.

bosr:
Hi Dan, it makes real sense. Thanks about the thorough explanation on generating the dependencies with `homebrew_go_resource` and sorry about my rant ;) Cheers

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `terraform 0.5.3`.

dhilton:
Ok... Slight issue is that the reported version is wrong:

```
danhilton@thunder.local  homebrew   master ✔ $ terraform --version                                                                                                                                                                                               09:22:25
Terraform v0.5.3-dev (6055d4f45226f306a934d4aadbedf73798c71cf2+CHANGES)

Your version of Terraform is out of date! The latest version
is 0.5.3. You can update by downloading from www.terraform.io
```

Anyone got any ideas? A small one but an annoying issue none the less.

bosr:
Hi Dan, no idea: I don't have the problem here (I'm using the latest formula you had merged)

    ~ ❯❯❯ terraform --version
    Terraform v0.5.3
    
    ~ ❯❯❯

PR #40411, opened 2015-06-05
Derrick 0.3 (new formula)

rieck:
Formula for Derrick, a simple network stream recorder based on libnids

### DIFF
--- a/Library/Formula/derrick.rb
+++ b/Library/Formula/derrick.rb
@@ -1,0 +1,26 @@
+class Derrick < Formula
+  desc "A Simple Network Stream Recorder"
+  homepage "https://github.com/rieck/derrick"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
This is maybe a little too niche for us; we generally frown on author-submitted projects unless they are really notable.

rieck:
Okay, is there any tap for network related tools?

mikemcquaid:
You could create your own tap?

rieck:
An own tap sounds like overkill for a single formula. I was hoping that there is something similar to `homebrew-science` for network analysis.

mikemcquaid:
It makes it easier for you to distribute updates for your own project as users can just `brew install rieck/derrick/derrick` or similar.

PR #40413, opened 2015-06-05
youtube-dl 2015.06.04.1

dstftw:


bfontaine:

xu-cheng:
@BrewTestBot test this please

xu-cheng:
This is duplicated with #40447 (which is already merged).

PR #40415, opened 2015-06-05
algernon 0.74 (new formula)

xyproto:
HTTP/2 web server with built-in support for Lua and Bolt DB.

### DIFF
--- a/Library/Formula/algernon.rb
+++ b/Library/Formula/algernon.rb
@@ -48,7 +48,7 @@ class Algernon < Formula
       Process.kill("HUP", algernon_pid)
 
       # Remove temporary Bolt database
-      `rm -f #{tempdb}`
+      rm_f tempdb
     end
   end
 end
--- a/Library/Formula/algernon.rb
+++ b/Library/Formula/algernon.rb
@@ -1,3 +1,5 @@
+require "language/go"
+
 class Algernon < Formula
   desc "HTTP/2 web server with Lua support"
   homepage "http://algernon.roboticoverlords.org/"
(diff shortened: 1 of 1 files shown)


xyproto:
Fails on mountain lion because `readline/keymaps.h` is missing.

DomT4:
You can use a more modern `readline` by adding `depends_on "readline"`.

xyproto:
Thanks for the tip!

Library/Formula/algernon.rb
```diff
@@ -0,0 +1,29 @@
...
+    system "go", "get", "-d"
+
+    # Build and install algernon
+    system "go", "build", "-o", "algernon"
+
+    bin.install "algernon"
+  end
+
+  test do
+    system "#{bin}/algernon", "--version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/algernon.rb
```diff
@@ -0,0 +1,29 @@
...
+  depends_on :hg => :build
+  depends_on "readline"
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    # Install Go dependencies
+    system "go", "get", "-d"
+
+    # Build and install algernon
```
mikemcquaid: Can remove the comments above.

xyproto:
Thanks for the feedback. Made the suggested changes (if the added test is considered good enough. It's running `go test`).

Library/Formula/algernon.rb
```diff
@@ -0,0 +1,26 @@
...
+  def install
+    ENV["GOPATH"] = buildpath
+
+    system "go", "get", "-d"
+    system "go", "build", "-o", "algernon"
+    bin.install "algernon"
+  end
+
+  test do
+    system "go", "test"
```
mikemcquaid: This won't work I'm afraid as the source code is not available at runtime. Perhaps you could use `fork` to create a process and then see if it's up?

xyproto:
That's a good suggestion, I'll try that.

Library/Formula/algernon.rb
```diff
@@ -0,0 +1,26 @@
+class Algernon < Formula
+  desc "HTTP/2 web server with Lua support"
+  homepage "http://algernon.roboticoverlords.org/"
+  url "https://github.com/xyproto/algernon/archive/0.74.tar.gz"
+  version "0.74"
```
DomT4: You're fine to remove this line. It's simple enough that `brew` can determine it from the URL.

xyproto:
Added a better test and removed obvious comments.

xyproto:
`spawn` is lacking on mountain lion, will find a fix.

xyproto:
Added a better test, removed obvious comments and removed the use of `spawn`, because of Ruby 1.8 / Mountain Lion.

Library/Formula/algernon.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  test do
+    begin
+      tempdb = "/tmp/_brew_test.db"
+      cport = ":45678"
+
+      # Start the server in a detached process
+      fork_pid = fork do
+        `#{bin}/algernon --httponly --server --addr #{cport} --boltdb #{tempdb}`
```
mistydemeo: If you're not capturing the output, this should be a `system` call.

Library/Formula/algernon.rb
```diff
@@ -0,0 +1,54 @@
...
+      # Check that the server is responding correctly
+      output = `curl -sIm3 -o- http://localhost#{cport}`
+      assert output.include?("Server: Algernon")
+      assert_equal 0, $?.exitstatus
+    ensure
+      # Stop the server gracefully
+      Process.kill("HUP", algernon_pid)
+
+      # Remove temporary Bolt database
+      `rm -f #{tempdb}`
```
mistydemeo: No need to shell out:

```ruby
rm_f tempdb
```

Library/Formula/algernon.rb
xyproto: I am capturing the output, in order to hide it. When using the system call, there is quite a bit of unneeded output. I tried using `spawn` as well, to hide the output on stderr and stdout, but that does not work on Ruby 1.8 / Mountain Lion. Is using `system` instead of hiding the output preferred?

Library/Formula/algernon.rb
xyproto: Thanks! Updated.

mistydemeo:
Instead of calling `go get`, can you use `go_resource`s to list its dependencies? You can see `terraform` for an example. Some users use the [homebrew-go-resources](https://github.com/samertm/homebrew-go-resources) tool to generate the list.

Library/Formula/algernon.rb
mistydemeo: How about using `Kernel.system` and redirecting stdout and stderr?

```ruby
Kernel.system "#{bin}/algernon", "--httponly", "--server", "--addr", cport, "--boltdb", tempdb, :out => "/dev/null", :err => "/dev/null
```

jacknagel: Isn't `Formula#system` silent unless `--verbose` is passed?

xyproto: @mistydemeo `Kernel.system` worked great! Thank you.

@jacknagel, you are right, `Formula#system` is silent unless `--verbose` is passed. However, when using `system`, the test just hangs, for some reason. In any case, `Kernel.system` solved it.

xyproto:
Using the `Kernel.system` line fails on Mountain Lion with: <a href="http://bot.brew.sh/job/Homebrew%20Pull%20Requests/27594/version=mountain_lion/testReport/brew-test-bot/mountain_lion/test_algernon/">`system': can't convert Hash into String (TypeError)</a>

I assume it's the `:out` and `:err` parameters that are misinterpreted by Ruby 1.8.

xyproto:
The output is now silent and it works on Mountain Lion as well.

mikemcquaid:

xyproto:
Thanks for merging and all the advice I received!

Issue #40416, opened 2015-06-05
czmq.rb:2 and euca2ools.rb:2: syntax error, unexpected end-of-input, expecting keyword_end


aarpon:
I tried running a brew update and got following error:
```
$ brew update
Error: /usr/local/Library/Formula/czmq.rb:2: invalid multibyte char (US-ASCII)
/usr/local/Library/Formula/czmq.rb:2: invalid multibyte char (US-ASCII)
/usr/local/Library/Formula/czmq.rb:2: syntax error, unexpected end-of-input, expecting keyword_end
  desc "High-level C binding for ØMQ"
                                   ^
Please report this bug:
    https://git.io/brew-troubleshooting
/usr/local/Library/Homebrew/formulary.rb:20:in `module_eval'
/usr/local/Library/Homebrew/formulary.rb:20:in `load_formula'
/usr/local/Library/Homebrew/formulary.rb:67:in `load_file'
/usr/local/Library/Homebrew/formulary.rb:58:in `klass'
/usr/local/Library/Homebrew/formulary.rb:54:in `get_formula'
/usr/local/Library/Homebrew/formulary.rb:166:in `factory'
/usr/local/Library/Homebrew/cmd/update.rb:173:in `block in report'
/usr/local/Library/Homebrew/cmd/update.rb:159:in `each_line'
/usr/local/Library/Homebrew/cmd/update.rb:159:in `report'
/usr/local/Library/Homebrew/cmd/update.rb:24:in `update'
/usr/local/Library/brew.rb:140:in `<main>'
```
When I fixed the character in czmq.rb:2, I got another error in /usr/local/Library/Formula/euca2ools.rb:2.

```
Error: /usr/local/Library/Formula/euca2ools.rb:2: invalid multibyte char (US-ASCII)
/usr/local/Library/Formula/euca2ools.rb:2: invalid multibyte char (US-ASCII)
/usr/local/Library/Formula/euca2ools.rb:2: syntax error, unexpected end-of-input, expecting keyword_end
  desc "Eucalyptus client API tools—works with Amazon EC2 and IAM"
                                     ^
Please report this bug:
    https://git.io/brew-troubleshooting
/usr/local/Library/Homebrew/formulary.rb:20:in `module_eval'
/usr/local/Library/Homebrew/formulary.rb:20:in `load_formula'
/usr/local/Library/Homebrew/formulary.rb:67:in `load_file'
/usr/local/Library/Homebrew/formulary.rb:58:in `klass'
[...]

PR #40420, opened 2015-06-05
avoid unicode in desc to fix `brew update` failures

vszakats:
Fixes: https://github.com/Homebrew/homebrew/issues/40416

```
==> brew readall --syntax						 FAILED
Error: problem in /usr/local/Library/Formula/czmq.rb
Error: problem in /usr/local/Library/Formula/ltl2ba.rb
Error: problem in /usr/local/Library/Formula/radamsa.rb
Error: problem in /usr/local/Library/Formula/shocco.rb
Error: problem in /usr/local/Library/Formula/termrec.rb
```

```
$ brew update
Error: /usr/local/Library/Formula/czmq.rb:2: invalid multibyte char (US-ASCII)
/usr/local/Library/Formula/czmq.rb:2: invalid multibyte char (US-ASCII)
/usr/local/Library/Formula/czmq.rb:2: syntax error, unexpected end-of-input, expecting keyword_end
  desc "High-level C binding for ØMQ"
                                   ^
Please report this bug:
/usr/local/Library/Homebrew/formulary.rb:20:in `module_eval'
/usr/local/Library/Homebrew/formulary.rb:20:in `load_formula'
/usr/local/Library/Homebrew/formulary.rb:67:in `load_file'
/usr/local/Library/Homebrew/formulary.rb:58:in `klass'
/usr/local/Library/Homebrew/formulary.rb:54:in `get_formula'
/usr/local/Library/Homebrew/formulary.rb:166:in `factory'
/usr/local/Library/Homebrew/cmd/update.rb:173:in `block in report'
/usr/local/Library/Homebrew/cmd/update.rb:159:in `each_line'
/usr/local/Library/Homebrew/cmd/update.rb:159:in `report'
/usr/local/Library/Homebrew/cmd/update.rb:24:in `update'
/usr/local/Library/brew.rb:140:in `<main>'
```

### DIFF
--- a/Library/Formula/czmq.rb
+++ b/Library/Formula/czmq.rb
@@ -1,5 +1,5 @@
 class Czmq < Formula
-  desc "High-level C binding for ØMQ"
+  desc "High-level C binding for ZeroMQ"
   homepage "http://czmq.zeromq.org/"
   url "http://download.zeromq.org/czmq-2.2.0.tar.gz"
   sha1 "2f4fd8de4cf04a68a8f6e88ea7657d8068f472d2"
--- a/Library/Formula/euca2ools.rb
+++ b/Library/Formula/euca2ools.rb
@@ -1,5 +1,5 @@
 class Euca2ools < Formula
-  desc "Eucalyptus client API tools—works with Amazon EC2 and IAM"
+  desc "Eucalyptus client API tools-works with Amazon EC2 and IAM"
   homepage "https://github.com/eucalyptus/euca2ools"
   url "https://github.com/eucalyptus/euca2ools/archive/v3.2.0.tar.gz"
   sha256 "4cfbae3b978312fa23e6a0329ec346568823afbbae0bc01075c8cff6707e7cb8"
--- a/Library/Formula/ltl2ba.rb
+++ b/Library/Formula/ltl2ba.rb
@@ -1,7 +1,7 @@
 require "formula"
 
 class Ltl2ba < Formula
-  desc "Translate LTL formulae to Büchi automata"
+  desc "Translate LTL formulae to Buchi automata"
   homepage "http://www.lsv.ens-cachan.fr/~gastin/ltl2ba/"
   url "http://www.lsv.ens-cachan.fr/~gastin/ltl2ba/ltl2ba-1.2b1.tar.gz"
   sha1 "74cff4914203753544bf300e041f433dbaeb3289"
--- a/Library/Formula/radamsa.rb
+++ b/Library/Formula/radamsa.rb
@@ -1,7 +1,7 @@
 require 'formula'
 
 class Radamsa < Formula
-  desc "Test case generator for robustness testing (a.k.a. a “fuzzer”)"
+  desc "Test case generator for robustness testing (a.k.a. a \"fuzzer"\)"
   homepage 'https://code.google.com/p/ouspg/wiki/Radamsa'
   url 'https://ouspg.googlecode.com/files/radamsa-0.3.tar.gz'
   sha1 '94664298b9c5c1563fe4aa7b8fc8530bb6628a51'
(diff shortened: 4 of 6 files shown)


DomT4:
Happy to ship this ASAP, but can you identify which encoding you're using to get the `brew readall` failures? I can't get this to repro on UTF-8. Would be helpful to know what we need to be wary of in future, etc.

xu-cheng:
@DomT4 you can try run`brew audit` and grep out UTF-8 problem.

vszakats:
My own system only reached into the `czmq` failure, and it uses no explicit language or codepage envvars (it's a separate admin account)

The `brew readall` failures were shown by the bot:
http://bot.brew.sh/job/Homebrew%20Pull%20Requests/27377/version=mavericks/console

DomT4:
Merged this to resolve the immediate problem with Mike's :+1:. Ideally we want to identify why it pops up and when.

vszakats:
Why isn't UTF-8 supported by default in formulae? And how can a formula be marked to use UTF-8 characters?

vszakats:
F.e. in this formula there are UTF-8 characters and they pass fine: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mecab.rb

(Though I had been scoffed to add more Japanese character into its `desc` when trying to edit this once.)

DomT4:
^ We add a [line](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mecab.rb#L1) to deal with `UTF-8` encoding in certain formulae.

bsiegel:
This introduced a syntax error in the `radamsa` formula:
`desc "Test case generator for robustness testing (a.k.a. a \"fuzzer"\)"`

As you can see, the second backslash needs to move one character to the left.

vszakats:
Ah okay, thanks. This was my educated guess, but this line turns out to have several different variations.

vszakats:
Ops, fixing this one.

xu-cheng:
I guess we can just prepend encoding magic line to every formulae when we load them. I will work on a PR tomorrow.

DomT4:
@vszakats https://github.com/Homebrew/homebrew/commit/2de0055c3606a9d0d54a9aec1eb2112a95f25154

vszakats:
@DomT4 👍 Thanks and sorry for the sloppiness.

vszakats:
@xu-cheng That would be very good.

DomT4:
> I guess we can just prepend encoding magic line to every formulae when we load them. I will work on a PR tomorrow.

I always presumed there was a specific reason we didn't to be honest. Maybe not. CC @Homebrew/owners

vszakats:
Here's current (possibly incomplete) collection of encoding markup in formulae:
```
bsdconv.rb:# encoding: UTF-8
cabocha.rb:# encoding: UTF-8
gnu-units.rb:# coding: utf-8
harfbuzz.rb:# encoding: utf-8
html-xml-utils.rb:# encoding: UTF-8
libidn.rb:# encoding: UTF-8
mecab.rb:# encoding: UTF-8
paps.rb:# encoding: UTF-8
par.rb:# encoding: UTF-8
unac.rb:# encoding: UTF-8
uni2ascii.rb:# encoding: UTF-8
yaz.rb:# encoding: UTF-8
```

jacknagel:
Note that the bot runs with the external encoding set to US-ASCII (i.e., `LANG=C`). Almost all users have it set to UTF-8.

I don't think we need to resort to injecting the magic comment string, we can just set the external encoding properly when we read in the formula file.

vszakats:
It appears that separate admin accounts have it set to `LANG=C`, so these accounts will need extra configuration to fully support Homebrew. Wouldn't it be a better solution to make sure to add `# encoding: UTF-8` markup to all formulae that use UTF-8 encoding to avoid this ambiguity?

jacknagel:
I'm not suggesting anyone configure anything, I'm suggesting that we set the encoding on the IO object when we read in the formula file.

vszakats:
Okay I think I get it.

mikemcquaid:
> I'm suggesting that we set the encoding on the IO object when we read in the formula file.

Sounds good :+1:

PR #40427, opened 2015-06-05
apache-archiva 2.2.0 (new formula)

leonardopanatta:
Added a formula to install and configure archiva.

### DIFF
--- a/Library/Formula/apache-archiva.rb
+++ b/Library/Formula/apache-archiva.rb
@@ -1,0 +1,56 @@
+class ApacheArchiva < Formula
+  desc "The Build Artifact Repository Manager"
+  homepage "https://archiva.apache.org/"
+  url "https://www.apache.org/dyn/closer.cgi?path=archiva/2.2.0/binaries/apache-archiva-2.2.0-bin.tar.gz"
+  sha256 "6af7c3c47c35584f729a9c139675a01f9a9819d0cdde292552fc783284a34cfa"
+
+  depends_on :java => "1.7+"
+
+  def install
+    libexec.install Dir["*"]
+
+    bin.install_symlink libexec/"bin/archiva"
+  end
+
+  def post_install
+    (var/"archiva/logs").mkpath
+    (var/"archiva/data").mkpath
+    (var/"archiva/temp").mkpath
+
+    cp_r libexec/"conf", var/"archiva"
+  end
+
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+      <dict>
+        <key>Label</key>
+        <string>#{plist_name}</string>
+        <key>ProgramArguments</key>
+        <array>
+          <string>#{opt_bin}/archiva</string>
+          <string>console</string>
+        </array>
+        <key>Disabled</key>
+        <false/>
+        <key>RunAtLoad</key>
+        <true/>
+        <key>UserName</key>
+        <string>archiva</string>
+        <key>StandardOutPath</key>
+        <string>#{var}/archiva/logs/launchd.log</string>
+        <key>EnvironmentVariables</key>
+        <dict>
+          <key>ARCHIVA_BASE</key>
+          <string>#{var}/archiva</string>
+        </dict>
+      </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    assert_match "was not running.", shell_output("#{bin}/archiva stop")
+  end
+end


bfontaine:
The test fails because `archiva status` exits with 1 instead of 0.

Library/Formula/apache-archiva.rb
```diff
@@ -0,0 +1,63 @@
+class ApacheArchiva < Formula
+  desc "The Build Artifact Repository Manager"
+  homepage "https://archiva.apache.org/"
+  url "http://ftp.unicamp.br/pub/apache/archiva/2.2.0/binaries/apache-archiva-2.2.0-bin.tar.gz"
```
vszakats: Please use this canonical Apache URL:
https://archive.apache.org/dist/archiva/2.2.0/binaries/apache-archiva-2.2.0-bin.tar.gz

vszakats: Please use this canonical Apache URL:
https://www.apache.org/dyn/closer.cgi?path=archiva/2.2.0/binaries/apache-archiva-2.2.0-bin.tar.gz

[UPDATED]

Library/Formula/apache-archiva.rb
```diff
@@ -0,0 +1,63 @@
...
+          <key>ARCHIVA_BASE</key>
+          <string>#{var}/archiva</string>
+        </dict>
+      </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    output = shell_output("#{bin}/archiva stop")
```
mikemcquaid: `assert_match "was not running.", shell_output("#{bin}/archiva stop")`

Library/Formula/apache-archiva.rb
vszakats:
```ruby
url "https://www.apache.org/dyn/closer.cgi?path=archiva/2.2.0/binaries/apache-archiva-2.2.0-bin.tar.gz"
```

Library/Formula/apache-archiva.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+      <dict>
+        <key>Label</key>
+        <string>#{plist_name}</string>
+
```
DomT4: Can we cut the whitespacing out of the `plist` please?

DomT4:
A whitespacing nit, and if you could squash your commits back into one that'd be great. Otherwise LGTM.

Library/Formula/apache-archiva.rb
```diff
@@ -0,0 +1,56 @@
...
+    bin.install_symlink libexec/"bin/archiva"
+  end
+
+  def post_install
+    (var/"archiva/logs").mkpath
+    (var/"archiva/data").mkpath
+    (var/"archiva/temp").mkpath
+
+    cp_r libexec/"conf", var/"archiva"
```
DomT4: Sorry, I've just noticed this line. Does it *have* to be copied, or can it actually be moved and still run fine?

If the latter, `(var/"archiva").install libexec/"conf"` should work - Thanks!

leonardopanatta: As per [Apache Archiva's doc] (http://archiva.apache.org/docs/2.2.0/adminguide/standalone.html), moving the conf folder will make the installation to fail.

DomT4: It seems like it'd work okay?

> Copy (move will fail) the configuration files from the Archiva installation (e.g. /opt/archiva-2.0/conf to the new location (e.g. /var/archiva/conf). If you've previously run Archiva, you may need to edit conf/archiva.xml to change the location of the repositories.

> (Optionally, leave a README file in the old conf location as a reminder of where you moved the config files.)
Set the environment variable ARCHIVA_BASE to the new location (e.g. /var/archiva). In bash, be sure to export the variable.

We're passing the variable to the `plist` below, so it'd work, right?

leonardopanatta: It wouldn't. I mean, it will be *"installed"* but will not run unless we change the archiva initialization script and probably the BASEDIR to find the `wrapper.conf` in another location (like `var/archiva`). Another option is moving all files but `wrapper.conf`. Either ways we're losing the proposal of separating the configuration from installation.

Is there a problem on copying folders?

DomT4: We just try not to dupe stuff as much as possible.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/41411480f42b8ba57914210250d5b3078c164a87. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #40430, opened 2015-06-05
daq 2.0.5

bfontaine:
Unfortunately I don’t know how to test this, I didn’t found any documentation regarding the C API, and `daq` is usually used with `snort` but I won’t add a dependency just for a test (plus `snort` already depends on `daq`).


mikemcquaid:
:+1:

PR #40436, opened 2015-06-05
sdelta3: migrate to boneyard

bfontaine:
The original website doesn’t exist anymore, the download URL doesn’t work and there are no mirrors. Debian doesn’t provide any package for it, like Fedora and MacPorts.

https://github.com/Homebrew/homebrew-boneyard/pull/50


mikemcquaid:
:+1:

bfontaine:
@mikemcquaid Does `brew pull` work in taps?

DomT4:
^ Yes.

bfontaine:
Thanks @DomT4, we need to use the PR URL (not just its number) for taps but it works :+1:

PR #40444, opened 2015-06-05
formula: always override user home

DomT4:
This is a rough draft of the initial idea that we always override the user's home directory. It borrows a chunk of code from further up a bit.

It meshes particularly well with the `HOMEBREW_SANDBOX` variable, reducing the number of failures associated with sandboxed formulae trying to write to `$HOME`, but it also means we no longer write files into the home `.cache`, `.config`, `.local`, `.npm`, etc folders which is consistently stuff that should be thrown away once we're done with it (Cached downloads, etc).

Tested against Node and #39771 with success.

Still need to check it doesn't conflict with [this chunk of optional test-bot code](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/test-bot.rb#L690-L694), although it shouldn't, theoretically.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -929,8 +929,17 @@ class Formula
   def stage
     active_spec.stage do
       @buildpath = Pathname.pwd
-      yield
-      @buildpath = nil
+      env_home = buildpath/".brew_home"
+      mkdir_p env_home
+
+      old_home, ENV["HOME"] = ENV["HOME"], env_home
+
+      begin
+        yield
+      ensure
+        @buildpath = nil
+        ENV["HOME"] = old_home
+      end
     end
   end
 


DomT4:

jacknagel:
I would be really worried about unintended side effects. I think I would rather let the sandbox reveal which packages misbehave, and fix them.

Side note, we definitely wouldn't want to use buildpath/home, it may conflict with real files in the package.

chdiza:
So would this mean that, e.g.,  `depends_on "~/brews/foo.rb"` will break?

DomT4:
The motivation is that it's been discussed repeatedly at various points, better control/override of the `$HOME` variable - An implementation is easier to discuss than a recurring idea. At least with an implementation we can better test practicality.

> we definitely wouldn't want to use buildpath/home

Agreed. If we wanted to push this further we could use Ruby's SecureRandom built-in library to generate a random, short hex string. The chance of that ever conflicting with the name of something else is pretty much zero. Or anything equally unlikely to come up, really.

> So would this mean that, e.g., depends_on "~/brews/foo.rb" will break?

From testing, no. Works fine here, both with and without the dep installed first.

jacknagel:
>So would this mean that, e.g., depends_on "~/brews/foo.rb" will break?

Note that this has never been officially supported.

chdiza:
> Note that this has never been officially supported.

I'd like to work on making this officially supported, unless the code is too nasty.  The ability to `depends_on` a local formula is pretty important for those of us who lean on custom formulas.

Can anyone please point me to where I should be looking?

tdsmith:
Now that the any-tap logic is in core, you can tap a local git repository like `brew tap tdsmith/test ~/test`.

chdiza:
> Now that the any-tap logic is in core, you can tap a local git repository like brew tap tdsmith/test ~/test.

But that forces me to (a) manage my custom formulas with git, and (b) rename my custom formulas that have the same names as official formulas.  The latter will spawn a zillion new cases of having to `depends_on "chdiza/myformulas/foo"` when currently I only need `depends_on "foo"`.

tdsmith:
Yes.

DomT4:
> Side note, we definitely wouldn't want to use buildpath/home

Fixed this overnight. Now using `homebrew` followed by a random numeric string generated by `rand`.

mikemcquaid:
I'd personally suggest `.home`, `.brewhome`, `.brew/home` or something. I don't think we need to go with random strings.

DomT4:
Done.

mikemcquaid:
I think we should merge this but want to check no other @Homebrew/owners disagree (particularly @jacknagel).

mikemcquaid:
Also, as a half-way solution we could set this only if `HOMEBREW_DEVELOPER` is set if that would make @jacknagel more happy.

Library/Homebrew/formula.rb
```diff
@@ -934,9 +934,14 @@ def exec_cmd(cmd, args, out, logfn)
...
   def stage
     active_spec.stage do
+      old_home = ENV["HOME"]
       @buildpath = Pathname.pwd
+      envhome = buildpath/".brew_home"
+      mkdir_p envhome
+      ENV["HOME"] = envhome
       yield
       @buildpath = nil
+      ENV["HOME"] = old_home
```
jacknagel: Since we're mutating global state, we need to use a proper cleanup idiom here:

```rb
def stage
  active_spec.stage do
    @buildpath = Pathname.pwd
    env_home = buildpath/".brew_home"
    mkdir_p env_home

    old_home, ENV["HOME"] = ENV["HOME"], env_home

    begin
      yield
    ensure
      @buildpath = nil
      ENV["HOME"] = old_home
    end
  end
end
```

DomT4: Updated.

mikemcquaid:
Let's just merge this in.

PR #40449, opened 2015-06-06
Always treat formula files as UTF-8

jacknagel:


dunn:
Do you think it would be worthwhile to add an `.editorconfig` enforcing utf-8 as well, or can that be done in the rubocop file?

xu-cheng:
You missed two formulae.
```
Library/Formula/gnu-units.rb:# coding: utf-8
Library/Formula/harfbuzz.rb:# encoding: utf-8
```

PR #40455, opened 2015-06-06
Formula-Cookbook: update download strategy example

vszakats:
The new one uses sha256 instead of md5 and 
the URLs still work and are all https, too.


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/c83a80c7b82e909ce3d3b6dae894ca2613e91245. Thanks @vszakats!

PR #40456, opened 2015-06-06
update tmux homepage and url

guoxiao:
Tmux homepage has been moved to github.


xu-cheng:

PR #40457, opened 2015-06-06
h2object brew install request

liujianping:
first version 0.0.1.beta commit

### DIFF
--- a/Library/Formula/h2object.rb
+++ b/Library/Formula/h2object.rb
@@ -1,0 +1,12 @@
+require 'formula'
+
+class H2object < Formula
+  homepage 'http://h2object.io'
+  url 'http://dl.h2object.io/h2object/version/h2object-0.0.1-beta.tar.gz'
+  sha1 'b4a5ff6edab956df4128c3bab75b45dbbec125f2'
+
+  def install
+    bin.install 'h2object'
+  end
+end
+
--- a/Library/Formula/h2object.rb
+++ b/Library/Formula/h2object.rb
@@ -1,8 +1,8 @@
-class H2object < Formula	
+class H2object < Formula
   homepage "http://h2object.io"
   url "http://dl.h2object.io/h2object/macosx/h2object-0.0.1.tar.gz"
   sha256 "262f2f491baca8c06f600f2c94506592a08cbfc6fede51e53586b58d3e8c0fe5"
-  desc "HTTP To OBJECT, the most fasting http server with themes to build web application"
+  desc "the most fasting http server with themes to build web application"
 
   def install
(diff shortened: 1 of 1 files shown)


Library/Formula/h2object.rb
```diff
@@ -0,0 +1,12 @@
+require 'formula'
```
bfontaine: Please remove this line

Library/Formula/h2object.rb
```diff
@@ -0,0 +1,12 @@
...
+
+class H2object < Formula
+  homepage 'http://h2object.io'
```
bfontaine: Use double quotes everywhere.

Library/Formula/h2object.rb
```diff
@@ -0,0 +1,12 @@
...
+  url 'http://dl.h2object.io/h2object/macosx/h2object-0.0.1-beta.tar.gz'
```
bfontaine: Is there a stable release we can use? It’ll need to go to [Homebrew/homebrew-devel-only](https://github.com/homebrew/homebrew-devel-only) if there is only a beta release.

Library/Formula/h2object.rb
```diff
@@ -0,0 +1,12 @@
...
+  sha1 '588bab3d9372b433d15616a8b8bbe6d2720f27d6'
```
bfontaine: `sha1`s are deprecated, please use `sha256`.

```
$ curl -sL http://dl.h2object.io/h2object/macosx/h2object-0.0.1-beta.tar.gz | shasum -a 256
262f2f491baca8c06f600f2c94506592a08cbfc6fede51e53586b58d3e8c0fe5  -
```

Library/Formula/h2object.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  def install
+    bin.install 'h2object'
+  end
```
bfontaine: Could you add a test? We need something to ensure it works, so it should be more substantial than e.g. `--help` or `--version`. Thanks!

tdsmith:
This may not meet our notability requirements; this seems like a very young project. Is the source code available? Software in Homebrew/homebrew must be built from source. How is it licensed?

liujianping:
source code now is not open temporarily for more readable reasons, will use GPL license recently.

DomT4:
We'll need the source code available before considering it, in all likelihood. Given the `beta` tag it may be a better fit for the `devel-only` tap temporarily.

mikemcquaid:
Passing on this until source code is available and it's more clear that this is a popular, established project.

PR #40459, opened 2015-06-06
harbour: add dependencies

vszakats:


Library/Formula/harbour.rb
```diff
@@ -17,6 +17,19 @@ class Harbour < Formula
   depends_on "pcre"
   depends_on :x11 => :optional
 
+  depends_on "allegro" => :optional
+  depends_on "cairo" => :optional
+  depends_on "curl" => :recommended
```
DomT4: Can it not use the system version?

Library/Formula/harbour.rb
```diff
@@ -17,6 +17,19 @@ class Harbour < Formula
...
+  depends_on "freeimage" => :optional
+  depends_on "gd" => :optional
+  depends_on "ghostscript" => :optional
+  depends_on "libmagic" => :optional
+  depends_on "mysql" => :optional
+  depends_on "ncurses" => :optional
+  depends_on "postgresql" => :optional
```
DomT4: The `sql` deps are usually styled:

```
depends_on :mysql
depends_on :postgresql
```

No quotes.

vszakats:
It can. Thanks for the hints, I'll revise a bit later.

DomT4:
How likely are all of these options to be widely used? Mike et al. have expressed a general desire to not slap every optional dep into formulae unless there's likely to be more-than-niche usage.

vszakats:
We don't have precise statistics on the contrib level, but contribs [most of these dependencies enable optional libraries referred to as "contrib"] are in general popular and used in various combinations. I'd estimate that `allegro` is used by very few users and there are superior alternatives, `ncurses` also have better alternatives (`gttrm`, `gtxwc`, `gtqtc` for both), so these may be dropped without much concern.

Without these contribs the project is still usable though it is unlikely to be useful for many users for real work.

If you absolutely want to nix more optional dependencies, these would be my list of second next ones: `cairo`, `freeimage`, `ghostscript`, `libmagic`.

DomT4:
@mikemcquaid - Thoughts here?

mikemcquaid:
@vszakats What options do you personally use?

vszakats:
The most important dependencies I use are also part of the system (f.e. `curl`, `openssl`) and/or "vendored" into `harbour` itself (f.e. `pcre1`/`pcre2`, `expat`, `libharu`). Besides these, I'm building against `ghostscript`, `postgresql`, `mariadb`/`mysql`, `unixodbc`, `qt5`, `x11`, and on OS X specifically I plan to replace `x11` with `qt5` (not very happy with either - and no native Cocoa solution at this time). For database dependencies and `ghostscript`, I personally don't need any of them at this moment.

[ Though, I use my [fork](https://github.com/vszakats/harbour-core) of this project, with a slightly different [formula](https://github.com/vszakats/harbour-core/blob/master/package/harbour.rb), and I'm trying to minimize dependencies as a development goal for end-user apps. And to make a final twist, I use OS X as a testing and thin-client app platform only, as far as my own/community development needs go. ]

mikemcquaid:
If you're not personally using this formula it's maybe worth just leaving it as-is (or at least just adding the options you use). Thanks!

vszakats:
I've adapted the changes made on my own copy of the formula here, and as major contributor for 15 years and lead for this project for many years, I've contributed to a lot parts that I don't personally use (including hundreds of formula here), I felt I could contribute to this too. I felt wrong.

mikemcquaid:
My apologies. We just tend to try and keep the options fairly minimal unless they are requested.

vszakats:
All of these dependencies are automatically detected and used by the build process if they are present, so if someone wants them, this is an easy workaround, even if not perfect.

PR #40468, opened 2015-06-07
installs man pages from cabal build

halloleo:
The cabal install does not install the man pages into the destination directory tree. So added this manually.


mikemcquaid:

For future reference the preferred commit message format  is `pandoc: install manpages.`.

PR #40469, opened 2015-06-07
bottle: add homepage and desc into regex

xu-cheng:
Some formulae put `desc` field behind `url` which can make bottle block go to strange place. So let's add `homepage` and `desc` into bottle regex. 

On the other hand, should we enforce a rule inside `audit.rb` about how different fields such as `desc`, `homepage` and `url` should be ordered? cc @Homebrew/owners @DomT4 @bfontaine for advice.


DomT4:
> On the other hand

IMO there's no harm in doing both. The regex tweak here works around the issue now, but long-term we can resolve the issue of misplaced lines in the initial block with `audit` enforcement.

We've been informally enforcing specific order of formulae for a while, so I'd be :+1: on making that a more formal, automated process.

mikemcquaid:
:+1: to this change. I'm potentially open to the audit change depending on how many formulae it flags. Maybe one for `audit --strict` for now?

PR #40472, opened 2015-06-07
audit: add rules on field order

xu-cheng:
Per https://github.com/Homebrew/homebrew/pull/40469#issue-85876583

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -56,6 +56,7 @@ end
 class FormulaText
   def initialize path
     @text = path.open("rb", &:read)
+    @lines = @text.lines
   end
 
   def without_patch
@@ -84,6 +85,11 @@ class FormulaText
 
   def =~ regex
     regex =~ @text
+  end
+
+  def line_number regex
+    index = @lines.index { |line| line =~ regex }
+    index ? index + 1 : nil
   end
 end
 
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/audit.rb
```diff
@@ -56,6 +56,7 @@ def audit
 class FormulaText
   def initialize path
     @text = path.open("rb", &:read)
+    @lines = @text.split("\n")
```
mikemcquaid: `@text.lines`

mikemcquaid:
Makes sense to me but I suspect other @Homebrew/owners may have opinions

Library/Homebrew/cmd/audit.rb
xu-cheng: `lines` will retain `\n` so I use split.

mikemcquaid: Mmm, still feels a bit weird. Does the presence of `\n` matter below?

xu-cheng: > Does the presence of \n matter below?

En, seems not matter. So I have changed it into `lines`.

DomT4:
Do we want to make a decision/enforcement about `version` line placement as well? We typically allow it either below the `url` or below the `sha`.

xu-cheng:
I rewrite the code to make it more scalable and add more field into the rules.

Library/Homebrew/cmd/audit.rb
```diff
@@ -141,6 +147,36 @@ def audit_file
...
+      [/^  desc ["'][\S\ ]+["']/,          "desc"          ],
+      [/^  homepage ["'][\S\ ]+["']/,      "homepage"      ],
+      [/^  url ["'][\S\ ]+["']/,           "url"           ],
+      [/^  mirror ["'][\S\ ]+["']/,        "mirror"        ],
+      [/^  version ["'][\S\ ]+["']/,       "version"       ],
+      [/^  (sha1|sha256) ["'][\S\ ]+["']/, "checksum"      ],
+      [/^  head ["'][\S\ ]+["']/,          "head"          ],
+      [/^  stable do/,                     "stable block"  ],
+      [/^  devel do/,                      "devel block"   ],
+      [/^  head do/,                       "head block"    ],
```
mikemcquaid: Maybe add the bottle block in here?

mikemcquaid:
Looks great to me.

Library/Homebrew/cmd/audit.rb
xu-cheng: Added.

mikemcquaid:
:+1:

tdsmith:
This seems to be breaking on Mountain Lion (i.e. Ruby 1.8): http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/2188/version=mountain_lion/testReport/junit/brew-test-bot/mountain_lion/audit_Homebrew_science_boost_compute___strict/

DomT4:
^ Seeing the same thing at X11 as well: http://bot.brew.sh/job/Homebrew%20X11%20Pull%20Requests/148/version=mountain_lion/console

xu-cheng:
Should be fixed in fc20f215f3bf104f7f957bb9159d8b70cc890688

mikemcquaid:
We should probably add more test runs of various commands e.g. `brew audit` in our test suite and/or `brew test-bot` at some point.

PR #40477, opened 2015-06-07
Added dex 1.0 (new formula)

yumitsu:
dex is a small and easy to use text editor. Colors and bindings can be fully customized to your liking.

### DIFF
--- a/Library/Formula/dex.rb
+++ b/Library/Formula/dex.rb
@@ -1,0 +1,27 @@
+class Dex < Formula
+  desc "Dextrous text editor"
+  homepage "https://github.com/tihirvon/dex"
+  url "https://github.com/tihirvon/dex/archive/v1.0.tar.gz"
+  sha256 "4468b53debe8da6391186dccb78288a8a77798cb4c0a00fab9a7cdc711cd2123"
+
+  head "https://github.com/tihirvon/dex.git"
+
+  depends_on "homebrew/dupes/ncurses" => :optional
+  depends_on "homebrew/dupes/libiconv" => :optional
+
+  def install
+    args = ["prefix=#{prefix}",
+            "CC=#{ENV.cc}",
+            "HOST_CC=#{ENV.cc}"]
+
+    args << "VERSION=#{version}" if build.head?
+
+    inreplace "Makefile", /-lcurses/, "-lncurses" if build.with? "ncurses"
+
+    system "make", "install", *args
+  end
+
+  test do
+    system bin/"dex", "-V"
+  end
+end


Library/Formula/dex.rb
```diff
@@ -0,0 +1,40 @@
+class Dex < Formula
+  desc "Dextrous text editor"
+  homepage "https://github.com/tihirvon/dex"
+  url "https://github.com/tihirvon/dex/archive/v1.0.tar.gz"
+  head "https://github.com/tihirvon/dex.git"
+
+  sha256 "4468b53debe8da6391186dccb78288a8a77798cb4c0a00fab9a7cdc711cd2123"
```
DomT4: This should go under the `url`.

Library/Formula/dex.rb
```diff
@@ -0,0 +1,40 @@
...
+    args = ["prefix=#{prefix}",
+            "CC=#{ENV.cc}",
+            "HOST_CC=#{ENV.cc}"]
+
+    args << "VERSION=#{version}" if build.head?
+
+    if build.with? "ncurses"
+      inreplace "Makefile", /-lcurses/, "-lncurses"
+
+      ENV.append_to_cflags "-I#{Formula["ncurses"].opt_include}"
```
DomT4: You can remove the ENV changes for `ncurses` and `libiconv` - They're both `keg_only` and when invoked `brew` will shunt them to the front of the `$PATH` and apply the variables automatically.

yumitsu:
@DomT4 :thumbsup:

Library/Formula/dex.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  depends_on "homebrew/dupes/ncurses" => :optional
+  depends_on "homebrew/dupes/libiconv" => :optional
+
+  def install
+    args = ["prefix=#{prefix}",
+            "CC=#{ENV.cc}",
+            "HOST_CC=#{ENV.cc}"]
+
+    args << "VERSION=#{version}" if build.head?
```
DomT4: Presumably, when building `head` the version won't actually be the `#{version}`?

Library/Formula/dex.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    if build.with? "ncurses"
```
DomT4: You can punt this onto the one line.

` inreplace "Makefile", /-lcurses/, "-lncurses" if build.with? "ncurses"`

Library/Formula/dex.rb
yumitsu: @DomT4 I added that to override version defined in Makefile — otherwise `dex -V` will say `v1.0` even if built from `head`.

DomT4: So this prints as `VERSION HEAD` from `dex -V`? If so, :+1:

yumitsu: @DomT4 with `brew install --HEAD dex`: 
```
$ dex -V
dex HEAD
```
with `brew install dex`:
```
$ dex -V
dex 1.0
```

DomT4: :100:

mikemcquaid:

For future reference the preferred commit message format for new formulae is `dex 1.0 (new formula)`.

PR #40483, opened 2015-06-08
Added gupnp-tools 0.8.10 (new formula)

ened:
Provides a collection of developer tools utilising GUPnP and GTK+.

### DIFF
--- a/Library/Formula/gupnp-tools.rb
+++ b/Library/Formula/gupnp-tools.rb
@@ -1,0 +1,26 @@
+class GupnpTools < Formula
+  desc "Free replacements of Intel's UPnP tools."
+  homepage "https://wiki.gnome.org/GUPnP/"
+  url "https://download.gnome.org/sources/gupnp-tools/0.8/gupnp-tools-0.8.10.tar.xz"
+  sha256 "592c53289ff1cd70e676405c56ca87b28d2da37a385d34a3bb9b944ba9108d17"
+
+  depends_on "pkg-config" => :build
+  depends_on "intltool" => :build
+  depends_on "gettext"
+  depends_on "gupnp"
+  depends_on "gupnp-av"
+  depends_on "gtk+3"
+  depends_on "gtksourceview3"
+  depends_on "ossp-uuid"
+
+  def install
+    system "./configure", "--disable-debug", "--disable-dependency-tracking",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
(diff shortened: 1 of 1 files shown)


Library/Formula/gupnp-tools.rb
```diff
@@ -0,0 +1,22 @@
...
+  depends_on "gtk+3"
+  depends_on "gtksourceview3"
+  depends_on "ossp-uuid"
+
+  def install
+    system "./configure", "--disable-debug", "--disable-dependency-tracking",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+end
```
mikemcquaid: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/gupnp-tools.rb
```diff
@@ -0,0 +1,22 @@
+require "formula"
```
mikemcquaid: This isn't needed.

mikemcquaid:
Please fix the `brew audit --strict` warnings: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/27478/version=yosemite/testReport/junit/brew-test-bot/yosemite/audit_gupnp_tools___strict/

Library/Formula/gupnp-tools.rb
ened: Difficult - as the formula is for a GUI tool. Suggestion?

mikemcquaid:

For future reference the preferred commit message format for new formulae is `gupnp-tools 0.8.10 (new formula)`.

bfontaine:
Just saw this, isn’t it the same formula as the one in [homebrew-x11](https://github.com/Homebrew/homebrew-x11/blob/master/gupnp-tools.rb)?

DomT4:
^ Yup, looks like it. We can remove it from `x11` and remove any reference to migrating it in `tap_migrations`.

It was probably only moved to X11 in the first place because GTK+3 originally had an X11 backend (It no longer does). Either that or the X11 dependency is actually optional, in which case it's cool to stay in the core as well.

DomT4:
^ Fixed by https://github.com/Homebrew/homebrew-x11/commit/b6b0c180fa4dbd3612622467a931e4b24fabed08 and https://github.com/Homebrew/homebrew/commit/411c2178448916d4005c059bc35ee780b5af53f5

PR #40488, opened 2015-06-08
Update to moco v0.10.1

schrepfler:


bfontaine:
Thank you for your contribution to Homebrew, @schrepfler!

For future reference, the preferred commit message is “moco 0.10.1”.

PR #40490, opened 2015-06-08
clang-format-vim (new formula)

hannesweisbach:
Installs clang-format.py from clang.llvm.org, which provides
clang-format integration into vim. The script is also available in the
clang source tree under clang/tools/clang-format/clang-format.py

Signed-off-by: Hannes Weisbach <hannes.weisbach@mailbox.tu-dresden.de>


hannesweisbach:
This pull request is superseded by #40497, which I think is a better solution.

PR #40493, opened 2015-06-08
Ansible: upgrade python dependencies and add Rackspace support (pyrax)

gildegoma:
I changed the order of the resources stanzas to make it clearer (imo), but I understand it can bother for the review. I hope it's okay, otherwise let me know how I should better re-phrase it.

I was also unsure about how to deal with shared dependencies, I guess that `resources` object could deal with duplicate declarations (last statement overwrites?).

### DIFF
--- a/Library/Formula/ansible.rb
+++ b/Library/Formula/ansible.rb
@@ -16,102 +16,269 @@ class Ansible < Formula
   depends_on :python if MacOS.version <= :snow_leopard
   depends_on "libyaml"
 
-  resource "docker-py" do
-    url "https://pypi.python.org/packages/source/d/docker-py/docker-py-0.6.0.tar.gz"
-    sha1 "01eb7b2cd1a607d361170041b973a0e36bb1be42"
+  #
+  # ansible (core dependencies)
+  #
+  resource "Jinja2" do
+    url "https://pypi.python.org/packages/source/J/Jinja2/Jinja2-2.7.3.tar.gz"
+    sha256 "2e24ac5d004db5714976a04ac0e80c6df6e47e98c354cb2c0d82f8879d4f8fdb"
   end
 
-  resource "requests" do
-    url "https://pypi.python.org/packages/source/r/requests/requests-2.6.0.tar.gz"
-    sha256 "1cdbed1f0e236f35ef54e919982c7a338e4fea3786310933d3a7887a04b74d75"
-  end
-
-  resource "websocket-client" do
-    url "https://pypi.python.org/packages/source/w/websocket-client/websocket-client-0.11.0.tar.gz"
-    sha1 "a38cb6072a25b18faf11d31dd415750692c36f33"
-  end
-
-  resource "six" do
-    url "https://pypi.python.org/packages/source/s/six/six-1.8.0.tar.gz"
-    sha1 "aa3b0659cbc85c6c7a91efc51f2d1007040070cd"
-  end
-
-  resource "pycrypto" do
-    url "https://pypi.python.org/packages/source/p/pycrypto/pycrypto-2.6.tar.gz"
-    sha1 "c17e41a80b3fbf2ee4e8f2d8bb9e28c5d08bbb84"
-  end
-
-  resource "boto" do
-    url "https://pypi.python.org/packages/source/b/boto/boto-2.36.0.tar.gz"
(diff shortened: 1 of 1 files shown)


tdsmith:
We should maybe just alphabetize all of the resources; order isn't important for python install_requires dependencies (the common case). I know it's annoying, but would you mind?

Shared dependencies only need to be declared once.

tdsmith:
> We should maybe just alphabetize all of the resources

I take this back; separating out the pyrax dependencies from the winrm dependencies from the docker dependencies etc is good. Alphabetizing by section would be fine though.

gildegoma:
Thanks @tdsmith. I'll go for alphabetising by section. Do you prefer that I rebase/force-update, or simply add a new commit to the branch ?

tdsmith:
Squashing and force-updating would be great, thanks!

gildegoma:
@tdsmith I force updated. I put a few cosmetic questions in code comments as I am still unsure it fits your requirements...

Library/Formula/ansible.rb
```diff
@@ -16,102 +16,269 @@ class Ansible < Formula
...
+
+  #
+  # docker-py (for Docker support)
+  #
+  resource "backports.ssl_match_hostname" do
+    url "https://pypi.python.org/packages/source/b/backports.ssl_match_hostname/backports.ssl_match_hostname-3.4.0.2.tar.gz"
+    sha256 "07410e7fb09aab7bdaf5e618de66c3dac84e2e3d628352814dc4c37de321d6ae"
+  end
+
+  resource "docker-py" do
```
gildegoma: I alphabetized the whole set of resources, but would you prefer to keep the "parent" (e.g. docker-py, pyrax or pywinrm) at the top (or bottom) of the relevant sections?

Library/Formula/ansible.rb
```diff
@@ -16,102 +16,269 @@ class Ansible < Formula
...
+    url "https://pypi.python.org/packages/source/p/pbr/pbr-1.0.1.tar.gz"
+    sha256 "fc3d19ab844647388cc13a3df403bda872d2fc16662803d0ebcc1787d3645552"
+  end
+
+  resource "PrettyTable" do
+    url "https://pypi.python.org/packages/source/P/PrettyTable/prettytable-0.7.2.tar.bz2"
+    sha256 "853c116513625c738dc3ce1aee148b5b5757a86727e67eff6502c7ca59d43c36"
+  end
+
+  resource "pyrax" do
```
gildegoma: ditto above about "parent" package position.

Library/Formula/ansible.rb
```diff
@@ -16,102 +16,269 @@ class Ansible < Formula
...
+  #
+  resource "isodate" do
+    url "https://pypi.python.org/packages/source/i/isodate/isodate-0.5.1.tar.gz"
+    sha256 "b12aed31c0e834543497e24d609a41531a800d8304c39e6665c45ca023b012fb"
   end
 
-  resource "pycrypto" do
-    url "https://pypi.python.org/packages/source/p/pycrypto/pycrypto-2.6.tar.gz"
-    sha1 "c17e41a80b3fbf2ee4e8f2d8bb9e28c5d08bbb84"
+  resource "pywinrm" do
```
gildegoma: ditto above about "parent" package position.

Library/Formula/ansible.rb
```diff
@@ -16,102 +16,269 @@ class Ansible < Formula
...
-    sha1 "01eb7b2cd1a607d361170041b973a0e36bb1be42"
+  #
+  # ansible (core dependencies)
+  #
+  resource "Jinja2" do
+    url "https://pypi.python.org/packages/source/J/Jinja2/Jinja2-2.7.3.tar.gz"
+    sha256 "2e24ac5d004db5714976a04ac0e80c6df6e47e98c354cb2c0d82f8879d4f8fdb"
+  end
+
+  resource "MarkupSafe" do
```
gildegoma: I kept the original package names from pypi (possibly case sensitive: MarkupSafe, PyYAML,...), but I think that it is fair (and clearer) to apply a "non-case sensitive" alphabetisation. Do you agree?

tdsmith:
This looks great; thanks!

PR #40497, opened 2015-06-08
Clang format: optionally install integration scripts

hannesweisbach:
This patch adds the options --with-vim, --with-emacs, --with-sublime, --with-bbedit, and --with-diff which cause their respective integration scripts to be installed into share/clang/.

### DIFF
--- a/Library/Formula/clang-format.rb
+++ b/Library/Formula/clang-format.rb
@@ -53,6 +53,7 @@ class ClangFormat < Formula
       bin.install "bin/clang-format"
     end
     bin.install "tools/clang/tools/clang-format/git-clang-format"
+    (share/"clang").install Dir["tools/clang/tools/clang-format/clang-format*"]
   end
 
   test do


bfontaine:
We should probably install them all and don’t use any option at all.

Library/Formula/clang-format.rb
```diff
@@ -52,7 +58,16 @@ def install
...
+    path = "tools/clang/tools/clang-format"
+    bin.install "#{path}/git-clang-format"
+    dest = share/"clang"
+    if build.with? "bbedit"
+      dest.install "#{path}/clang-format-bbedit.applescript"
+    end
+    dest.install "#{path}/clang-format-diff.py" if build.with? "diff"
+    dest.install "#{path}/clang-format-sublime.py" if build.with? "sublime"
+    dest.install "#{path}/clang-format.el" if build.with? "emacs"
+    dest.install "#{path}/clang-format.py" if build.with? "vim"
```
tdsmith: Let's just do `(share/"clang").install Dir["tools/clang/tools/clang-format/*"]` unless there's something in there we don't want to install.

Library/Formula/clang-format-vim.rb
```diff
@@ -0,0 +1,14 @@
+class ClangFormatVim < Formula
+  desc "Minimal clang-format vim-integration"
+  homepage "http://clang.llvm.org/docs/ClangFormat.html#vim-integration"
+  url "https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/clang-format.py"
```
tdsmith: This would need to be from a stable release, not from trunk.

Library/Formula/clang-format.rb
hannesweisbach: Unfortunately there is, see: https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/.
It is possible to glob with tools/clang/tools/clang-format/*clang-format*

Library/Formula/clang-format-vim.rb
hannesweisbach: Well *$#;!. That commit should not be on this branch. Can I simply remove it and force-push, or does that break merging for the pull request?

Library/Formula/clang-format.rb
tdsmith: Cool, that works :)

Library/Formula/clang-format-vim.rb
tdsmith: Yep, force-pushing is fine!

hannesweisbach:
I have made the changes and it installs correctly with --HEAD and --build-from-source. The new files are obviously missing from bottled archives. I suppose they have to be ... rebottled? Not sure how that works.

mikemcquaid:
https://github.com/Homebrew/homebrew/pull/40515/files

mikemcquaid:

PR #40498, opened 2015-06-08
Use -NDEBUG for gflags by default

theopolis:
gflags will append a friendly: `Debug build (NDEBUG not #defined)` on targets when a `--version` is used. By default, a gflags install should be release ready and have `-NDEBUG` defined. A brew user may build from source and include `--with-debug` to prevent this definition.


tdsmith:
Thanks for the PR!

PR #40500, opened 2015-06-08
SonarQube 5.1.1

Godin:


bfontaine:
Thank you for your contribution, @Godin, we appreciate it!

A tiny note for the future: the preferred checksum is now `sha256`.

Godin:
@bfontaine Didn't knew that, thanks for notice. To not forget in future here is immediate PR - #40502

PR #40502, opened 2015-06-08
Switch to sha256 for SonarQube

Godin:


bfontaine:
Thanks! :smile: 

Another note: the preferred commit format is `formula-name: message`.

Godin:
@bfontaine noted for the next ones ;)

PR #40503, opened 2015-06-08
sonar: add sonarqube alias

bfontaine:
From the homepage:

> SonarQube™ software (previously known as “Sonar”) is an open source
> project hosted at Codehaus.

@mikemcquaid Do we handle software renaming? e.g. moving `sonar` to `sonarqube` or `sonar-qube` and adding an alias from the old name to the new name.

Also, should I use `sonarqube` or `sonaqube` (or both)?

### DIFF
--- a/Library/Aliases/sonarqube
+++ b/Library/Aliases/sonarqube
@@ -1,0 +1,1 @@
+../Formula/sonar.rb


DomT4:
> Do we handle software renaming? e.g. moving sonar to sonarqube or sonar-qube and adding an alias from the old name to the new name.

We don't, really. Moving formula names is something that's been broken in Homebrew for a long time. The system thus far has been to set up an alias from the new name to the old name as you've done in this PR (See PolarSSL for a recent discussion on this sort of thing).

bfontaine:
Thanks @DomT4, any opinion on the `sonarqube` vs. `sonar-qube` question?

DomT4:
`sonarqube` seems fine to me.

PR #40510, opened 2015-06-09
qt 4.8.7

UniqMartin:
Bump to latest (and final) Qt 4 release. This drops the separate patch that is now included in the new release. Also moves to the HTTPS mirror recently adopted in the “qt5” formula and fixes an indentation error.

I recompiled and quickly tested a few formulae that depend on Qt 4 without noticing any breakage.


tdsmith:
Fixed the paraview test in Homebrew/homebrew-science@4ad7a11. Thanks for the PR!

PR #40517, opened 2015-06-09
Crystal v0.5.1 (new formula)

ctate:
Open source code generator for every language, library and framework.

### DIFF
--- a/Library/Formula/crystal.rb
+++ b/Library/Formula/crystal.rb
@@ -10,6 +10,8 @@ class Crystal < Formula
     libexec.install Dir["*"]
     bin.install_symlink Dir["#{libexec}/bin/*"]
     ENV.prepend_path "PATH", "#{Formula["node"].opt_libexec}/npm/bin"
+    cd libexec
+    system "npm", "install"
   end
 
   test do
--- a/Library/Formula/crystal.rb
+++ b/Library/Formula/crystal.rb
@@ -17,6 +17,30 @@ class Crystal < Formula
   end
 
   test do
-    system bin/"crystal", "help"
+    mkdir (testpath/".crystal")
+    (testpath/".crystal/config.yml").write <<-EOS.undent
+      name: test
+      version: 0.1.0
+      description: test description
+      author:
+        name: Test User
+        email: test@example.com
+        url: http://example.com
+      copyright: 2015 Test
+      modules:
+        official.readme: master
+      imports:
+        readme: ReadmeGenerator
+      outputs:
+        - generator: readme
+          spec:
+            name: $name
+            version: $version
+            description: $description
+    EOS
+    cd testpath
+    system bin/"crystal", "update"
+    system bin/"crystal", "build"
+    system bin/"crystal", "test"
   end
 end
--- a/Library/Formula/crystal.rb
+++ b/Library/Formula/crystal.rb
@@ -9,11 +9,7 @@ class Crystal < Formula
   def install
     libexec.install Dir["*"]
     bin.install_symlink Dir["#{libexec}/bin/*"]
-  end
-
-  def post_install
-    cd libexec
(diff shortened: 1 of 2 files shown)


Library/Formula/crystal.rb
```diff
@@ -0,0 +1,22 @@
...
+    bin.install_symlink Dir["#{libexec}/bin/*"]
+  end
+
+  def post_install
+    cd libexec
+    system "npm", "install"
+  end
+
+  test do
+    system bin/"crystal", "help"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/crystal.rb
```diff
@@ -0,0 +1,22 @@
...
+  depends_on "node"
+
+  def install
+    libexec.install Dir["*"]
+    bin.install_symlink Dir["#{libexec}/bin/*"]
+  end
+
+  def post_install
+    cd libexec
+    system "npm", "install"
```
mikemcquaid: What is this installing? Why does it need to be in `post_install`?

Library/Formula/crystal.rb
ctate: Added a test for updating, building and testing a Crystal project. Thank you!

Library/Formula/crystal.rb
ctate: This install Node.js dependencies. It depends on `node` which includes `npm`, but it doesn't seem to be accessible until `post_install`. Please let me know if I should be doing something differently. Thanks for your help!

mikemcquaid: CC @DomT4 for thoughts here.

DomT4: `superenv` punts `npm` out of the `$PATH`, which does cause build failure during `def install`.

So far we've worked around it by either manually re-adding `HOMEBREW_PREFIX/"bin"` to the `$PATH` or by handling it in `post_install`.

Personal point of view is that we should be using `ENV.prepend_path "PATH", "#{Formula["node"].opt_libexec}/npm/bin"` instead in applicable formulae (Or doing that automatically).

mikemcquaid: > Personal point of view is that we should be using ENV.prepend_path "PATH", "#{Formula["node"].opt_libexec}/npm/bin" instead in applicable formulae (Or doing that automatically).

Yep, we should do that.

DomT4: Automatically or manually?

mikemcquaid: Either :)

DomT4: It's not particularly widely used, and automatically would probably be best handled by a `node` requirement which we've resisted so far.

Manually may be the way to go until we hit a mass of `npm`-using stuff being submitted.

mikemcquaid: Manually it is!

@ctate Please add `ENV.prepend_path "PATH", "#{Formula["node"].opt_libexec}/npm/bin"` into `install` instead.

DomT4: We should also move this back to `def install` so the `npm install` stage is included in the bottle. If we add `npm` to the `$PATH` with the above instruction, we can stop heavily leaning on `post_install` for Node stuff and deliver a slightly smoother UX.

`ENV["HOME"] = ".brew_home"` or similar needs adding as well, otherwise we'll have sandbox or leakage issues.

ctate:
Thanks for the helpful info! Fixed @ commit a997e6bd2f146adc2f4e3db6d6682394f005a49c: https://github.com/Homebrew/homebrew/pull/40530

PR #40518, opened 2015-06-09
ocilib 4.1.0 (new formula)

vszakats:


Library/Formula/ocilib.rb
```diff
@@ -0,0 +1,41 @@
...
+    (testpath/"test.c").write <<-EOS.undent
+      #include "ocilib.h"
+
+      int main( void )
+      {
+        return OCI_Initialize( NULL, NULL, OCI_ENV_DEFAULT | OCI_ENV_CONTEXT ) ? 0 : 1;
+      }
+    EOS
+    system ENV.cc, "test.c", "-locilib", "-o", "test"
+    system "./test"
```
mikemcquaid: Nice test!

vszakats: ops, but got the formatting wrong.

vszakats: Whitespace corrected and squashed.

vszakats:
Anything else to be done here, to get accepted by maintainers?

bfontaine:
We need one of the @Homebrew/owners to :+1: / :-1: here.

Library/Formula/ocilib.rb
```diff
@@ -0,0 +1,41 @@
...
+  sha256 "d86c1e6d5ef33541299996b76497a1271498cdcf5720ee0f1248a0b78dfb3a0c"
+
+  head "http://git.code.sf.net/p/orclib/git"
+
+  depends_on OracleHomeVar
+
+  def install
+    system "./configure"
+    system "make"
+    system "make", "install"
```
xu-cheng: No `prefix`?

vszakats: Thanks @xu-cheng, I've added it along with some more required (and optional) settings.

vszakats:
Repository moved to GitHub; thanks @vrogier! `head` updated accordingly.

vszakats:
Replaced download url with new GitHub one.

Library/Formula/ocilib.rb
```diff
@@ -0,0 +1,47 @@
...
+  desc "Open source C and C++ library for accessing Oracle Databases"
+  homepage "http://www.ocilib.net/"
+  url "https://github.com/vrogier/ocilib/releases/download/v4.1.0/ocilib-4.1.0-gnu.tar.gz"
+  sha256 "d86c1e6d5ef33541299996b76497a1271498cdcf5720ee0f1248a0b78dfb3a0c"
+
+  head "https://github.com/vrogier/ocilib.git"
+
+  option "with-charset-wide", "Use wide charset with Oracle OCI libraries"
+
+  depends_on OracleHomeVar
```
mikemcquaid: Oh, just noticed this. Not sure what I think about including this in core given it can't be compiled without another third-party download.

vszakats: The same logic / dependency is present in the existing `qt5` formula, when option `with-oci` is requested:
https://github.com/Homebrew/homebrew/blob/master/Library/Formula/qt5.rb

I'm not saying this fact should resolve all concerns, but there is already a precedent.

mikemcquaid: My issue that in that case it's optional and this case it's mandatory. It means our CI bot cannot test this and it's a pain for even maintainers to do so. I think it'd be a better fit for a tap you run, instead, if it's something you can test.

vszakats: I understand and second the pain of installing the Oracle component.

Starting my own tap is probably too much trouble (not being an Oracle DB user). I've been looking into installing the Oracle client as a resource instead, but downloading it from the canonical site requires to be logged in and the license be accepted before each download. If anyone knows about an alternative way to get that component, maybe it'd be a workable alternative (also for `qt5`).

mikemcquaid: Out of interest: why submit the formula of something you aren't using?

vszakats: I've once developed a DB driver [component](https://github.com/vszakats/harbour-core/tree/master/contrib/sddoci) that depends on `ocilib` as part of [Harbour](https://github.com/vszakats/harbour-core) part out of fun and because I used Oracle DB while attending lectures. Since I still use Harbour (though not this specific part of it), I try to keep it maintained, even those parts I don't currently need. Having brew formulae for its dependencies do help to keep it maintained on OS X.

vszakats: In short, probably a bad habit! ;)

mikemcquaid: Hehe. Well, passing on this for now, sorry!

PR #40524, opened 2015-06-09
A useful task timer.

vurte:
I was using this task timer on Linux. But when I switched back to Mac I didn't found a free 
task timer with the same features like gtimer.
It has good reporting features and the storage formate is in plain text, so it
is also good to sync it (dropbox for example) with other gtimer instances on different platforms.
In my case I can use the same timer data on Linux and now when I am using my Mac. 

I think it is worthy to be added


Library/Formula/gtimer.rb
```diff
@@ -0,0 +1,84 @@
+class Gtimer < Formula
+  desc "GTimer is a GTK application for timing how you spend your time among various projects and tasks with support for generating plain text and HTML reports."
+  homepage "http://www.k5n.us/gtimer.php"
+  url "http://downloads.sourceforge.net/project/gtimer/2.0.0/gtimer-2.0.0.tar.gz"
+  version "2.0.0"
+  sha256 "a2dd70793a5b1b4d497be0c4429b8e8cef6910497b6af8fbdedff00764308458"
+
+  depends_on :x11
```
mikemcquaid: This will need submitted to `homebrew-x11` if it has a mandatory X11 dependency, I'm afraid.

Library/Formula/gtimer.rb
```diff
@@ -0,0 +1,84 @@
...
+  depends_on "gtk+"
+
+
```
mikemcquaid: Remove this newline.

Library/Formula/gtimer.rb
```diff
@@ -0,0 +1,84 @@
...
+  patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Library/Formula/gtimer.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def install
+    ENV.O1
```
mikemcquaid: Report this issue upstream and clarify why it is needed as a Ruby comment.

Library/Formula/gtimer.rb
```diff
@@ -0,0 +1,84 @@
+class Gtimer < Formula
+  desc "GTimer is a GTK application for timing how you spend your time among various projects and tasks with support for generating plain text and HTML reports."
+  homepage "http://www.k5n.us/gtimer.php"
+  url "http://downloads.sourceforge.net/project/gtimer/2.0.0/gtimer-2.0.0.tar.gz"
+  version "2.0.0"
```
bfontaine: You can remove this line, Homebrew guesses the version from the URL.

Library/Formula/gtimer.rb
```diff
@@ -0,0 +1,83 @@
+class Gtimer < Formula
+  desc "GTimer is a GTK application for timing how you spend your time among various projects and tasks with support for generating plain text and HTML reports."
```
bfontaine: Please shorten the description. The formula’s name and description should fit under 80 characters.

Library/Formula/gtimer.rb
```diff
@@ -0,0 +1,83 @@
...
+  homepage "http://www.k5n.us/gtimer.php"
+  url "http://downloads.sourceforge.net/project/gtimer/2.0.0/gtimer-2.0.0.tar.gz"
```
bfontaine: You can use `https://` here.

Library/Formula/gtimer.rb
vurte: OK Thanks. I will move it to homebrew-x11

PR #40527, opened 2015-06-09
Create & Use npm path nudge

DomT4:
Various things:

* Stops `npm`-using formulae just breaking completely when `HOMEBREW_SANDBOX` is set, and spamming up the `$HOME` cache/config when it isn't.
* Stops `npm`-using formulae being reliant on `post_install` which currently means we can't bottle the most significant part of each of these compiles.
* Sets a standard format for how to get `npm` back in the `$PATH`, since `superenv` removes it and the answer so far has been to re-add `HOMEBREW_PREFIX/"bin"` manually which risks picking up unexpected deps, or using `post_install` which isn't great UX.

### DIFF
--- a/Library/Formula/skinny.rb
+++ b/Library/Formula/skinny.rb
@@ -1,40 +1,31 @@
-require "formula"
-
-class UniversalNpm < Requirement
-  fatal true
-  satisfy { which("npm") }
-  def message
-    "npm is required. If you have installed node with `--without-npm` option, reinstall with `--with-npm`."
-  end
-end
-
 class Skinny < Formula
   desc "Full-stack web app framework built on Scalatra"
   homepage "http://skinny-framework.org/"
   url "https://github.com/skinny-framework/skinny-framework/releases/download/1.3.18/skinny-1.3.18.tar.gz"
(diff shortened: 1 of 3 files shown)


DomT4:

DomT4:
Would also impact #40517 and #39771

Library/Formula/skinny.rb
```diff
@@ -1,5 +1,3 @@
-require "formula"
-
 class UniversalNpm < Requirement
```
DomT4: I'm not completely sure why this is a requirement given we ship Node `--with-npm` by default. IMO we should scrap the block and requirement completely.

Library/Formula/tegh.rb
```diff
@@ -1,18 +1,22 @@
...
+
+  head "https://github.com/D1plo1d/tegh.git", :branch => "develop"
 
-  depends_on 'node'
+  depends_on "node"
 
   def install
+    if build.head?
+      ENV.prepend_path "PATH", "#{Formula["node"].opt_libexec}/npm/bin"
+      ENV["HOME"] = ".brew_home"
```
mikemcquaid: Probably wants to be a full path with e.g. `buildpath` in it. Otherwise :+1: on the PR

PR #40528, opened 2015-06-09
pdftoipe: move to homebrew/head-only

DomT4:
Upstream [don't want to cut a specific release and have suggested](https://github.com/otfried/ipe-tools/issues/5#issuecomment-110225266) we move it to `head-only`, so move it does.


mikemcquaid:
:+1:

PR #40529, opened 2015-06-09
x265 1.7

dunn:


Library/Formula/x265.rb
```diff
@@ -13,20 +14,17 @@ class X265 < Formula
...
+  depends_on "cmake" => :build
+  depends_on :macos => :lion
 
   def install
-
     args = std_cmake_args
-    args.delete '-DCMAKE_BUILD_TYPE=None'
-    args << '-DCMAKE_BUILD_TYPE=Release'
-    args << '-DHIGH_BIT_DEPTH=ON' if build.include? '16-bit'
+    args << "-DHIGH_BIT_DEPTH=ON" if build.include? "16-bit"
```
DomT4: `if build.with?` :wink:

dunn: Woops, thanks!

dunn: I can open a PR to add that to `audit` if you'd like.

DomT4: There's language in there already about deprecating options, but perhaps we're only applying to the new formulae and changing them as we can for now - That would be in line with our policy on SHA256, double quotes, etc.

Library/Formula/x265.rb
```diff
@@ -13,20 +14,17 @@ class X265 < Formula
     sha256 "440805e02452b80400ff428967bd4c5d24f042bea2cc6b5d76a3a37648965a60" => :mountain_lion
   end
 
-  head 'https://bitbucket.org/multicoreware/x265', :using => :hg
+  option "with-16-bit", "Build a 16-bit x265 (default: 8-bit)"
 
-  depends_on 'yasm' => :build
-  depends_on 'cmake' => :build
-  depends_on :macos => :lion
+  deprecated_option "16-bit" => "with-16-bit"
```
DomT4: Can one/any of @Homebrew/owners chime in on this? Are we content keeping the `16-bit`, `32-bit` options or do we want to migrate them to `with/without`?

asparagui: I don't think there's 16/32-bit options for this flag.  HIGH_BIT_DEPTH is somewhat experimental/cpu heavy, I don't think it should be enabled by default.  Above LGTM.

DomT4:
Nobody objected to the `16-bit` language change, so merged in https://github.com/Homebrew/homebrew/commit/0eb9bab0818371d5c0fa55d5a9a4885be4beae67. Thanks @asparagui for the review, and @dunn for the PR! :tada:

PR #40534, opened 2015-06-09
Stack v0.0.0

danpalmer:


Library/Formula/stack.rb
```diff
@@ -0,0 +1,15 @@
+class Stack < Formula
+  desc "A cross-platform program for developing Haskell projects."
+  homepage "https://github.com/commercialhaskell/stack"
+  url "https://github.com/fpco/stack/releases/download/v0.0.0-beta/stack-0.0.0-x86_64-osx.gz"
```
mistydemeo: This should build from source, rather than use prebuilt binaries.

mistydemeo:
Is this a stable release?

Library/Formula/stack.rb
danpalmer: One of the reasons for this tool is to simplify the building of Haskell applications, by bypassing the dependency hell of package and compiler versions. As the tool itself is a haskell tool, those issues come with compiling the tool.

GHC is a large download, and with all of the dependencies of this project, compiling from source would require very large downloads, a long compile time, and unless this was all done in a sandbox environment and thrown away at the end, it would also leave a large number of possibly conflicting packages lying around the system.

It appears that the developers of Stack are encouraging the use of the binaries, and it is for these reasons that I chose to use the binaries.

mistydemeo: Homebrew also distributes binaries; all submitted packages are build using our hardware in a sandbox environment. If you change this formula to build from source, then we'll add the binaries when it's merged.

danpalmer:
@mistydemeo this is a beta release, there is no stable release yet. What is the standard practice for this?

Library/Formula/stack.rb
danpalmer: Oh cool. Is that new? I thought it built everything from source! Is it *always* binaries that are distributed? Or does it try and then fall back to a local compilation?

I'm not sure when I'll have time to do this, that will be quite a bit more work. Will try and do it sometime, but someone else will likely beat me to it.

mistydemeo: We've been shipping binaries for most packages for a couple of years now. Most users get binaries delivered, provided that a) they use the default install options, b) they haven't told Homebrew to build from source, and c) the binary is compatible with their OS version and Homebrew installation location.

Thanks!

chdiza:
> this is a beta release, there is no stable release yet. What is the standard practice for this?

https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Acceptable-Formulae.md#stable-versions

mikemcquaid:
It looks like there's a `0.0.1` now. Given this seems fairly popular I wonder if we include it as-is? Is it beta as in "not for production readiness" or beta as in "there are known showstopping bugs"?

Library/Formula/stack.rb
mikemcquaid: This will need done before this can be included, thanks.

mistydemeo:
More the former; the announcements say they expect bugs, but it doesn't necessarily look like there are *known* showstoppers.

This is at 0.0.2 now (releases seem to be coming quickly); given the popularity/interest I'd be fine including this if we had a formula building from source.

mikemcquaid:
> More the former; the announcements say they expect bugs, but it doesn't necessarily look like there are known showstoppers.

I'm :+1: on including this.

mkscrg:
@danpalmer how can I help push this forward? `brew install stack` would be :+1::sparkles:

mikemcquaid:
@mkscrg Feel free to just create your own PR, btw.

danpalmer:
Apologies for not being on top of this.

Haskell development is something I do in my free time, not my job, and this isn’t even actual Haskell development, so as I’m sure you can imagine my time for doing this is quite limited.

@mkscrg by all means, go ahead and create your own PR, I don’t know when I’ll get time to finish this off, as I’d need to learn more about how Homebrew packaging works to get it building from source.

DomT4:
Closing for now. Will welcome a new from-source PR if anyone is interested.

mkscrg:
I tried building a from-source formula for `stack`, then found this in [its docs](https://github.com/commercialhaskell/stack/wiki/Downloads#os-x):

> Note: due to [GHC bug 10322](https://ghc.haskell.org/trac/ghc/ticket/10322), building stack from source fails with GHC 7.10.1. This bug will be fixed in 7.10.2, but in the meantime, we recommend using GHC 7.8.4 on OS X.

GHC 7.10.2 RC2 is out now, so a proper `stack` formula will be possible soon!

Issue #40519, opened 2015-06-09
Git package broken in 10.11 El Capitan


nrbrook:
While I managed to fix most issues I was initially seeing in El Capitan by installing the latest command line tools, running `brew update`, `brew doctor` and fixing the warnings, I can't reinstall git (which I was attempting to do to fix a different issue).

```
$ brew reinstall git
==> Reinstalling git
==> Downloading https://www.kernel.org/pub/software/scm/git/git-2.4.3.tar.xz
Already downloaded: /Library/Caches/Homebrew/git-2.4.3.tar.xz
==> make prefix=/usr/local/Cellar/git/2.4.3 sysconfdir=/usr/local/etc CC=clang C
./git-compat-util.h:219:10: fatal error: 'openssl/ssl.h' file not found
#include <openssl/ssl.h>
         ^
1 error generated.
make: *** [ll-merge.o] Error 1
```

DomT4:
We don't support `El Capitan` at this point. What build options are you using?

mohibeyki:
I have the same issue, and I'm not using any build options, what options should I use?

DomT4:
You both have the 10.11 CLT installed? Can someone do `mdfind ssl.h | grep include`? Does `/usr/include` exist for you both?

npinon:
Yes it exist : 

```
mdfind ssl.h | grep include
/usr/local/Cellar/openssl/1.0.1j/include/openssl/ssl.h
/usr/local/Cellar/openssl/1.0.2/include/openssl/ssl.h
/usr/local/Cellar/openssl/1.0.2a-1/include/openssl/ssl.h
/usr/local/Cellar/openssl/1.0.1f/include/openssl/ssl.h
/usr/local/Cellar/openssl/1.0.1h/include/openssl/ssl.h
/usr/local/Cellar/openssl/1.0.1i/include/openssl/ssl.h
```

I don't have the same issue : 

```
brew reinstall git
==> Reinstalling git
==> Downloading https://www.kernel.org/pub/software/scm/git/git-2.4.3.tar.xz
Already downloaded: /Library/Caches/Homebrew/git-2.4.3.tar.xz
==> make prefix=/usr/local/Cellar/git/2.4.3 sysconfdir=/usr/local/etc CC=clang CFLAGS= LDFLAGS= install
./git-compat-util.h:100:10: fatal error: 'unistd.h' file not found
#include <unistd.h>
         ^
1 error generated.
[...]

PR #40536, opened 2015-06-09
Extremely Tentative El Capitan Support

DomT4:
Re https://github.com/Homebrew/homebrew/issues/40519#issuecomment-110456475

I *literally* added a caveat that it is unsupported, since we've started getting heavily-commented issues already.

### DIFF
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -214,6 +214,7 @@ module OS
       "6.3"   => { :clang => "6.1", :clang_build => 602 },
       "6.3.1" => { :clang => "6.1", :clang_build => 602 },
       "6.3.2" => { :clang => "6.1", :clang_build => 602 },
+      "7.0"   => { :clang => "7.0", :clang_build => 700 },
     }
 
     def compilers_standard?
--- a/Library/Homebrew/os/mac/xcode.rb
+++ b/Library/Homebrew/os/mac/xcode.rb
@@ -18,7 +18,7 @@ module OS
         else
           # Default to newest known version of Xcode for unreleased OSX versions.
           if MacOS.version > "10.10"
-            "6.3.2"
+            "7.0"
           else
             raise "Mac OS X '#{MacOS.version}' is invalid"
           end
@@ -161,6 +161,7 @@ module OS
 
       def latest_version
         case MacOS.version
+        when "10.11" then "700.0.53"
         when "10.10" then "602.0.53"
         when "10.9"  then "600.0.57"
         when "10.8"  then "503.0.40"
--- a/Library/brew.rb
+++ b/Library/brew.rb
@@ -50,6 +50,15 @@ if OS.mac? and MacOS.version < "10.6"
   EOABORT
 end
 
+if OS.mac? && MacOS.version == "10.11"
+  opoo <<-EOS.undent
+    Although Homebrew has added initial recognition of 10.11, it is unsupported.
+    You may encounter breakage or other failure and there is no guarantee
+    Homebrew can resolve those issues until El Capitan is stable.
+
+  EOS
+end
+
 # Many Pathname operations use getwd when they shouldn't, and then throw
 # odd exceptions. Reduce our support burden by showing a user-friendly error.
 Dir.getwd rescue abort "The current working directory doesn't exist, cannot proceed."
--- a/Library/Homebrew/os/mac/version.rb
+++ b/Library/Homebrew/os/mac/version.rb
@@ -4,6 +4,7 @@ module OS
   module Mac
     class Version < ::Version
       SYMBOLS = {
+        :el_capitan    => '10.11',
         :yosemite      => '10.10',
         :mavericks     => '10.9',
         :mountain_lion => '10.8',


DomT4:

Library/Homebrew/os/mac.rb
```diff
@@ -214,6 +214,7 @@ def preferred_arch
       "6.3"   => { :clang => "6.1", :clang_build => 602 },
       "6.3.1" => { :clang => "6.1", :clang_build => 602 },
       "6.3.2" => { :clang => "6.1", :clang_build => 602 },
+      "7.0"   => { :clang => "6.1", :clang_build => 602 },
```
mistydemeo: I'm seeing `7.0.0 (clang-700.0.53)` with the CLT.

abradner:
Compiling packages breaks (for me at least) unless you are on the Xcode 7 Beta. While this is mostly obvious, I forgot to `xcode-select` to the new xcode beta and was scratching my head for a while until I spotted my mistake. 
If we're warning users about El Capitan, it might be worthwhile to have `brew doctor` complain if your xcode 7 isn't set up correctly

Library/Homebrew/os/mac/xcode.rb
```diff
@@ -15,10 +15,11 @@ def latest_version
         when "10.8"  then "5.1.1"
         when "10.9"  then "6.2"
         when "10.10" then "6.3.2"
+        when "10.11" then "7.0"
```
mikemcquaid: Might not need this if it's in the `else`

Library/Homebrew/cmd/doctor.rb
```diff
@@ -428,6 +428,14 @@ def check_tmpdir_sticky_bit
...
 end
 
+def check_supported_osx
+  if MacOS.version == "10.11" then <<-EOS.undent
+    Although Homebrew has added initial recognition of 10.11 it is unsupported.
+    You may encounter breakage or other failure and there is no guarantee
+    Homebrew can resolve those issues until El Capitan is stable.
+  EOS
+  end
+end
```
mikemcquaid: When is this printed? Might be worth printing it on literally every `brew` run.

mikemcquaid:
Would be good to confirm all these values on a VM or actual machine by one of the maintainers.

Library/Homebrew/cmd/doctor.rb
```diff
@@ -428,6 +428,14 @@ def check_tmpdir_sticky_bit
   end
 end
 
+def check_supported_osx
+  if MacOS.version == "10.11" then <<-EOS.undent
+    Although Homebrew has added initial recognition of 10.11 it is unsupported.
```
vszakats: Would there be a comma appropriate here?:
Although Homebrew has added initial recognition of 10.11**,** it is unsupported.

Library/Homebrew/os/mac.rb
DomT4: :+1:, Will tweak.

Library/Homebrew/os/mac/xcode.rb
DomT4: We made this change last year, but I'm happy to drop it?

Library/Homebrew/cmd/doctor.rb
DomT4: Every and any time a `brew doctor` is run. I can move it to `brew.rb` instead though? We already have a too-old OS X version warning in there, so a warning about too-new wouldn't be out of place.

Library/Homebrew/cmd/doctor.rb
DomT4: Yup, no problem with that suggestion.

Library/Homebrew/os/mac.rb
DomT4: Xcode has it down as `Version 7.0 beta (7A120f)`, but I'm happy to go with the CLT.

mistydemeo: Sorry, I mean the clang version I'm seeing is different from the one you reported - I assume 7.0 beta has the same version?

DomT4: Unless you mean the `clang_build` should be `700`? Apologies for the confusion :smile_cat:

mistydemeo: Yeah, exactly - the clang I see is labelled 7.0.0 build 700, so assuming it's the same in Xcode, we want `{ :clang => "7.0", :clang_build => 700 }`

DomT4: Thanks, Sorry for the confusion there. You're right, I was looking at the wrong thing:

```
/Volumes/Xcode/Xcode-beta.app/Contents/Developer/usr/bin/gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)
Target: x86_64-apple-darwin14.4.0
Thread model: posix
```

I should have been looking at:

```
/Volumes/Xcode/Xcode-beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang --version
Apple LLVM version 7.0.0 (clang-700.0.53)
Target: x86_64-apple-darwin14.4.0
Thread model: posix
```

I presume?

mistydemeo: Yeah, the latter looks correct. Kind of weird the "gcc" shim points at an older clang though.

Library/brew.rb
```diff
@@ -50,6 +50,15 @@
   EOABORT
 end
 
+if OS.mac? && MacOS.version == "10.11"
+  opoo <<-EOS.undent
+    Although Homebrew has added initial recognition of 10.11, it is unsupported.
+    You may encounter breakage or other failure and there is no guarantee
+    Homebrew can resolve those issues until El Capitan is stable.
+
```
DomT4: Blank line here to make sure the commands aren't buffered against each other, i.e.

```
~> brew info openssl
Warning: Although Homebrew has added initial recognition of 10.11, it is unsupported.
You may encounter breakage or other failure and there is no guarantee
Homebrew can resolve those issues until El Capitan is stable.

openssl: stable 1.0.2a-1 (bottled)
OpenSSL SSL/TLS cryptography library
https://openssl.org
```

Rather than:

```
~> brew info openssl
Warning: Although Homebrew has added initial recognition of 10.11, it is unsupported.
You may encounter breakage or other failure and there is no guarantee
Homebrew can resolve those issues until El Capitan is stable.
openssl: stable 1.0.2a-1 (bottled)
OpenSSL SSL/TLS cryptography library
https://openssl.org
```

DomT4:
Addressed all feedback in new push.

* Clang version corrected.
* Unsupported warning moved from Doctor to every time `brew` is executed.
* Added a comma to the initial warning sentence.

Library/Homebrew/os/mac/xcode.rb
DomT4: Killed it.

DomT4:
@Homebrew/owners Any objection to shipping this? We're getting a lot of errors because people haven't updated their Xcode to match OS X 10.11, which this should kick up a `brew doctor` warning about.

jacknagel:
I imagine enough people are going to complain that we'll have to move the 10.11 warning to a doctor-only warning, but I guess we can try it.

DomT4:
I moved it to each run per Mike's request, but I don't mind where we put it. Whatever the consensus is.

Library/brew.rb
```diff
@@ -50,6 +50,20 @@
   EOABORT
 end
 
+if OS.mac? && MacOS.version == "10.11" && MacOS::Xcode.version < "7.0"
```
mistydemeo: Need a more complex conditional here - we need to check for the cases where:

* Xcode is installed, but below 7
* Xcode isn't installed, but CLT 7 is installed
* Xcode and CLT are both installed, both out of date
* Xcode and CLT are both installed, only one is up to date

This should probably also be part of the Xcode preinstall checks in `cmd/install` rather than here: https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/install.rb#L112-L123

DomT4: I thought @mikemcquaid wanted it runtime for `brew`, rather than just on `install`, but I could have misunderstood :smile_cat:.

Agreed that the conditional is too simplistic in this situation; Should've realised that before I pushed really. Will tune it up to cover the wider scope, thanks :)

jacknagel: I believe the Xcode/CLT doctor checks handle all these cases already. I don't see the benefit of duplicating all of that logic here.

jacknagel: i.e., it doesn't really make sense for things like `brew --prefix` or `brew info` or `brew deps` to outright fail if Xcode is outdated. We already have pre-install checks for this situation.

DomT4: Maybe the best solution here is to revert it back to what it was before the latest push, merge that, and then take firmer action if the number of issues created because of non-system-appropriate-xcode/clt doesn't drop?

DomT4: Reverted.

If we want to issue a sterner warning I'm happy to use the existing Doctor methods, but attempt to make them more fatal on 10.11 (i.e. you cannot try and install xyz until your Xcode/CLT is up-to-date), but the consensus seems to lean towards seeing how much the basic support reduces the level of Issues before we go more drastic.

If that's the case, this should be good to :rocket: once it's green on the bot again. I'll let one of you pull the trigger on this if you're happy.

jacknagel:
Can we merge at least the first commit, please, so that doctor can start warning 10.11 users that they need to update Xcode/CLT?

DomT4:
I merged it whole. If we get complaints about the runtime warning being too frequent I'll move it back to `doctor`, but there seems to be a basic consensus about getting this in to reduce the level of 10.11 issues solved by updating CLT/Xcode.

PR #40542, opened 2015-06-10
elasticsearch 1.6.0

berislavbabic:
https://www.elastic.co/downloads/past-releases/elasticsearch-1-6-0

A new version of elasticsearch has been released, so we can start using
it in development.


xu-cheng:

PR #40543, opened 2015-06-10
plural: fix for n=0

xu-cheng:


bfontaine:
[Zero is plural in English](http://english.stackexchange.com/questions/38293/why-is-zero-plural)

xu-cheng:
Thanks @bfontaine that I learn something new.

PR #40552, opened 2015-06-10
aws-elasticbeanstalk-3.4.5

zabawaba99:


DomT4:
Do any of the resources need updating?

zabawaba99:
From what I saw, no. I was able to successfully install/test that aws-elasticbeanstalk after making the change.

Are there any other checks I should do as well?

DomT4:
It'd be cool to check if the resources on pypi have updates available, and if we can use those. Not sure what @tdsmith's preference on checking for updates on formula update is.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/a74669d1a8d9c29395b89cde4d2b4cc8abde013b. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:

zabawaba99:
Anytime!

PR #40572, opened 2015-06-11
proguard add proguardgui to formula

### LINKED ISSUE #40571: Add proguardgui to proguard formula
ProGuard also includes a GUI which should be enabled in the proguard formula.

justinedelson:
fixes #40571

### DIFF
--- a/Library/Formula/proguard.rb
+++ b/Library/Formula/proguard.rb
@@ -6,7 +6,9 @@ class Proguard < Formula
 
   def install
     libexec.install "lib/proguard.jar"
+    libexec.install "lib/proguardgui.jar"
     bin.write_jar_script libexec/"proguard.jar", "proguard"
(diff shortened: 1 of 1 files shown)


DomT4:
Is this something most users will want, or something that's better as an option?

justinedelson:
I can't say for sure, but in my experience, you would use the GUI to test out configuration options in an iterative approach and then use the proguard CLI to execute with those options in some kind of build environment. In fact, it might be *more* useful to use distribute the GUI via homebrew than the plain CLI as most usages would be using a plugin into a build tool like Ant or Maven.

DomT4:
Thanks for the explanation. Merged in https://github.com/Homebrew/homebrew/commit/a70c85d1c70dee230cd7adeb51539b303074c02d. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Please note for future the preferred commit message style for non-version updates is `formula: thing changed`.

Issue #40516, opened 2015-06-09
upgrade python breaks pip


tonicbupt:
when I use `brew upgrade`, python 2.7.9 is upgraded to python 2.7.10, then I found pip broken.

```
$ pip
Traceback (most recent call last):
  File "/usr/local/bin/pip", line 9, in <module>
    load_entry_point('pip==7.0.1', 'console_scripts', 'pip')()
  File "build/bdist.macosx-10.10-x86_64/egg/pkg_resources/__init__.py", line 552, in load_entry_point
  File "build/bdist.macosx-10.10-x86_64/egg/pkg_resources/__init__.py", line 2672, in load_entry_point
  File "build/bdist.macosx-10.10-x86_64/egg/pkg_resources/__init__.py", line 2345, in load
  File "build/bdist.macosx-10.10-x86_64/egg/pkg_resources/__init__.py", line 2351, in resolve
  File "/usr/local/lib/python2.7/site-packages/pip-7.0.1-py2.7.egg/pip/__init__.py", line 15, in <module>
    from pip.vcs import git, mercurial, subversion, bazaar  # noqa
  File "/usr/local/lib/python2.7/site-packages/pip-7.0.1-py2.7.egg/pip/vcs/subversion.py", line 9, in <module>
    from pip.index import Link
  File "/usr/local/lib/python2.7/site-packages/pip-7.0.1-py2.7.egg/pip/index.py", line 30, in <module>
    from pip.wheel import Wheel, wheel_ext
  File "/usr/local/lib/python2.7/site-packages/pip-7.0.1-py2.7.egg/pip/wheel.py", line 34, in <module>
    from pip._vendor.distlib.scripts import ScriptMaker
  File "/usr/local/lib/python2.7/site-packages/pip-7.0.1-py2.7.egg/pip/_vendor/__init__.py", line 92, in load_module
    raise ImportError("No module named '%s'" % (name,))
ImportError: No module named 'pip._vendor.distlib.scripts'
```

if I use `python get-pip.py`, another error occurs:

```
$ python get-pip.py
Traceback (most recent call last):
  File "get-pip.py", line 28, in <module>
    import tempfile
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/tempfile.py", line 32, in <module>
    import io as _io
[...]

PR #40573, opened 2015-06-11
pythons: fix ssl on 10.11

tdsmith:
Hint the library detection code so it always finds openssl headers, even when they're not installed system-wide.


mistydemeo:
Tested both on 10.11 :+1:

tdsmith:
ref #40516

PR #40575, opened 2015-06-11
Added support for removing given formula versions

alexreg:
Doing `uninstall formula/version` uninstalls just the keg for version 'version' of formula 'formula.
Doing `uninstall formula` behaves as before.

### DIFF
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -9,7 +9,9 @@ module Homebrew
       ARGV.kegs.each do |keg|
         keg.lock do
           puts "Uninstalling #{keg}... (#{keg.abv})"
-          keg.unlink
+          if keg.linked?
+            keg.unlink
+          end
           keg.uninstall
           rack = keg.rack
           rm_pin rack
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -35,6 +35,15 @@ module HomebrewArgvExtension
     require 'keg'
     require 'formula'
     @kegs ||= downcased_unique_named.collect do |name|
+      name_parts = name.split('/')
+      if name_parts.length == 1
+        name = name_parts[0]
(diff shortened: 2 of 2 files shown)


xu-cheng:
This breaks keg named using full-qualified name. e.g. (`brew rm homebrew/python/numpy`). If you want, you can use different annotation for version, say `formula@version`.

Ping @Homebrew/owners @bfontaine @DomT4 for opinion on whether we should this kind feature in general.

jacknagel:
I'm -1 on such a feature. It's easy to `brew unlink` and then remove specific version from the cellar.

I am really against introducing a formula+version syntax because people will expect it to work in other places. It won't, and probably never will. If we have to add something, it should be an explicit option, like `--version=<version>`.

mikemcquaid:
I agree with @jacknagel. Between `brew switch`, `brew pin`, `brew link` and `brew cleanup` there's a variety of ways to deal with removing versions. I also agree about formula versions particularly.

alexreg:
@xu-cheng Aha, I didn't think of that. Indeed, `@` would be best.

alexreg:
@jacknagel @mikemcquaid The problem is, there's no easy way to do it right now. (Many users have naturally been confused about this, as evidenced by the a number of [Google results](https://www.google.co.uk/search?q=brew+remove+version&ie=utf-8&oe=utf-8&gws_rd=cr&ei=zpx5Vaf1LYjiUYijgogJ#q=brew+remove+specific+version)). The main problem of telling users to do `rm -rf` is that it doesn't lock the keg correctly, apart from being less intuitive.

mikemcquaid:
I guess the question worth asking is: why are you trying to do this?

alexreg:
Automation within a multi-package-manager system.

DomT4:
I'm also :-1: sadly. Our multiple versions support is already quite complicated, and I'm reluctant to add to that. If anything I'd like us to lean *more* on non-core taps for versions support.

mikemcquaid:
Passing on this, sorry and thanks.

PR #40577, opened 2015-06-11
mysql: modernize syntax

tamird:


DomT4:
Apologies for this, but I managed to give you a [merge conflict](https://github.com/Homebrew/homebrew/commit/7d59e8729449f9dd689780dd4e325f60f45fcf17) by having the same train of thought on `libexec.install` after you mentioned it in Versions. If you want to rebase and reapply the rest, that's cool.

Apologies again :crying_cat_face:

tamird:
done

tamird:
Sigh, I have no idea why this is failing in post install. Any ideas?

tamird:
@DomT4 halp >.<

Library/Formula/mysql.rb
```diff
@@ -60,11 +60,11 @@ def install
       -DCMAKE_FIND_FRAMEWORK=LAST
       -DCMAKE_VERBOSE_MAKEFILE=ON
       -DMYSQL_DATADIR=#{datadir}
-      -DINSTALL_INCLUDEDIR=include/mysql
-      -DINSTALL_MANDIR=share/man
-      -DINSTALL_DOCDIR=share/doc/#{name}
-      -DINSTALL_INFODIR=share/info
-      -DINSTALL_MYSQLSHAREDIR=share/mysql
+      -DINSTALL_INCLUDEDIR=#{include}/mysql
```
DomT4: You may need to revert some of these. I've just noticed the little hidden line above the `args` which says: `# -DINSTALL_* are relative to prefix`. `include`, `man` etc provide full paths rather than relative paths, I believe. Might be the cause of the issue.

I don't have the time to check this out tonight, but I'll put it on the list to poke my head into tomorrow.

tamird:
green

mikemcquaid:
Is this fixing any issues or just cleanup?

tamird:
just cleanup, born of discussion in https://github.com/Homebrew/homebrew-versions/pull/859

mikemcquaid:
Generally we don't merge syntax-only changes but I can see an argument for having this more in-sync with the `homebrew-versions` formulae (if they are all similar). @DomT4, thoughts?

DomT4:
This PR was motivated by the `mysql55` formula trying to install something in `libexec` before actually creating `libexec` which meant certain things weren't getting installed.

In Homebrew/homebrew we addressed this towards the start of 2014 by adding [this line](https://github.com/Homebrew/homebrew/blob/11bf7623564bf33ac9f0a5f4ea008c2ec3e1b6f3/Library/Formula/mysql.rb#L128). So @tamird teed up a PR to Homebrew/versions with that addition, and then I pointed out we could just do `libexec.install`, and Tamir highlighted his proposed language actually came from the core.

So, short and sweet, I'm :+1: on harmonised syntax across taps in that it minimises contributor confusion and makes maintaining differing versions easier, but in practice it's hard to keep on top of long-term.

Library/Formula/mysql.rb
```diff
@@ -53,18 +53,18 @@ def install
...
-      -DINSTALL_INCLUDEDIR=include/mysql
-      -DINSTALL_MANDIR=share/man
-      -DINSTALL_DOCDIR=share/doc/#{name}
-      -DINSTALL_INFODIR=share/info
-      -DINSTALL_MYSQLSHAREDIR=share/mysql
+      -DINSTALL_INCLUDEDIR=#{include.relative_path_from(prefix)}/mysql
+      -DINSTALL_MANDIR=#{man.relative_path_from(prefix)}
+      -DINSTALL_DOCDIR=#{doc.relative_path_from(prefix)}
+      -DINSTALL_INFODIR=#{info.relative_path_from(prefix)}
+      -DINSTALL_MYSQLSHAREDIR=#{share.relative_path_from(prefix)}/mysql
```
jacknagel: What is the point of this change?

tamird: Technically, it's more correct. I was already here, so I looked for the "right" way to do it

jacknagel: Can you explain why it's more correct? Are the strings not ultimately identical?

tamird: Ultimately yes, but if the method `share` changes to return something other than `prefix+'share'`, this will continue to return the correct value, while the old code would not

jacknagel: >method share changes to return something other than prefix+'share'

Frankly, that's never going to happen, and this change is confusing. Let's leave it as it was, please.

tamird: Sure, done.

mikemcquaid:
Thanks @tamird!

PR #40591, opened 2015-06-11
openssl 1.0.2c

DomT4:


DomT4:
Bumped the timeout and requeued. There's going to be a lot of failure whatever, because it tests the core and installs those versions, and then it tests (for example) Versions which it then tries to link over the core formula which causes `FAILED` messages - But there's nothing in there actually related to the OpenSSL bump itself.

```
==> brew install erlang-r13						 FAILED
==> Installing erlang-r13 from Homebrew/homebrew-versions
Error: You must `brew link unixodbc' before Homebrew/versions/erlang-r13 can be installed
==> brew fetch --retry erlang-r14					 PASSED
==> brew install erlang-r14						 FAILED
==> Installing erlang-r14 from Homebrew/homebrew-versions
==> Downloading https://homebrew.bintray.com/bottles-versions/erlang-r14-R14B04_1.yosemite.bottle.tar.gz
Already downloaded: /Library/Caches/Homebrew/erlang-r14-R14B04_1.yosemite.bottle.tar.gz
==> Pouring erlang-r14-R14B04_1.yosemite.bottle.tar.gz
Could not symlink bin/dialyzer
Target /usr/local/bin/dialyzer
is a symlink belonging to erlang. You can unlink it:
  brew unlink erlang

  brew link --overwrite erlang-r14
```

So on, so forth.

Slightly longer term, the answer is to either add `conflicts_with` to absolutely everything in Versions that does conflict, or to tweak the test-bot logic to remove core formulae once it moves into a tap for testing.

DomT4:
Yup. It's all `Versions` linking failures against installed Core formulae, which are safe to ignore in this PR. Will pull once ML is done.

chdiza:
@DomT4 This edition of openssl doesn't actually fix the Equifax bug.  Mutt on a gmail account again asks me for the certs!

chdiza:
I noticed when I looked at their changelog that none of it appeared to be what is in those patches we're carrying.

chdiza:
Never mind, I think I'm an idiot :)

vszakats:
Report about ABI breakage between 1.0.2a and 1.0.2b:
https://marc.info/?l=openssl-dev&m=143407129721271&w=2

DomT4:
> Never mind, I think I'm an idiot 

The expected code is there; I spent about a half hour reading through OpenSSL source code to check before pushing the PR :smile_cat:. I think Mutt may be a Mutt issue to some degree, sadly.

> Report about ABI breakage between 1.0.2a and 1.0.2b:

Yup, have been following this overnight. It was accidental on OpenSSL's part and they are trying to determine the most appropriate response.

We'll be breaking compat anyway soon enough by killing off SSLv3 & IDEA, so if necessary we can move forwards those changes and en mass revisions, but I'd rather get the OpenSSL change in cleanly first personally.

DomT4:
Bingo:

```
Forthcoming OpenSSL releases
============================

The OpenSSL project team would like to announce the forthcoming release
of OpenSSL versions 1.0.2c, 1.0.1o.

These releases will be made available on Friday 12th June. They will
fix two specific issues: 1) an HMAC ABI incompatibility with previous
releases, and 2) make it possible to accept a zero length extension
block in a ClientHello.

These are not security releases.

Yours

The OpenSSL Project Team
```

vszakats:
1.0.2c is out.

PR #40592, opened 2015-06-11
Update mongo fomula to always gen conf from homebrew config

inlined:
We noticed that our per-user install didn't work without --build-from-source now that a keg is available. This is because the install function wasn't being called and the prebuilt mongo.conf was being installed. By moving the conf generation code into post_install, it's executed on both keg installs and clean builds. The unlink is necessary because keg installs will cp their default mongo.conf

### DIFF
--- a/Library/Formula/mongodb.rb
+++ b/Library/Formula/mongodb.rb
@@ -90,9 +90,12 @@ class Mongodb < Formula
     end
 
     scons "install", *args
+  end
 
-    (buildpath+"mongod.conf").write mongodb_conf
-    etc.install "mongod.conf"
+  def post_install
+    conf = etc + "mongod.conf"
+    conf.unlink
+    conf.write mongodb_conf
 
     (var+"mongodb").mkpath
     (var+"log/mongodb").mkpath


mikemcquaid:
This isn't correct/necessary. It will only avoid overwriting the configuration if it has been modified and will create a file with `.default` appended.

DomT4:
Seems to have caused `post_install` to fail:

```
Warning: The post-install step did not complete successfully
You can try again using `brew postinstall mongodb`
```

inlined:
@mikemcquaid Forgive my ignorance in writing formulae. Are you saying that HEAD was supposed to create a `mongo.config.default` file? I'm not seeing any `.default` files created when using HEAD and the `mongo.config` file was a static keg asset which doesn't take `var` into account like the method in the formula.
@DomT4 Any chance you have anything more specific? I can try to get ahold of another machine to try this against. I wonder if this change fixes homebrew with custom install locations at the expense of default homebrew installs.

mikemcquaid:
@inlined Can you provide a step-by-step reproducible test-case to demonstrate the problem you're having that this is fixing? Thanks!

inlined:
If you have a homebrew install in a custom directory and install this
formula from a keg, mongo.config still points to the standard config
directory.

mikemcquaid:
> If you have a homebrew install in a custom directory and install this
> formula from a keg, mongo.config still points to the standard config
> directory.

Sorry, I need reproducible step-by-step instructions with the expected and actual output and an explanation of how they differ.

DomT4:
Closing per silence. If there's an issue here we can fix we'll look at it, but we'll need some more explicit detail. Would suggest opening an Issue documenting the specific failure and desired end result and we can go from there.

PR #40597, opened 2015-06-11
Add infer formula

dambrisco:
Adds a basic formula for Facebook's BSD-licensed `infer`.

~~Caveat: Due to infer's existing [dislike of symlinks](https://github.com/facebook/infer/pull/13), I've added a not-so-nice code-modifying inreplace.~~

### DIFF
--- a/Library/Formula/infer.rb
+++ b/Library/Formula/infer.rb
@@ -1,0 +1,52 @@
+class Infer < Formula
+  desc "Infer is a static analysis tool that can analyze Java, Obj-C, and C"
+  homepage "https://fbinfer.com/"
+  url "https://github.com/facebook/infer/releases/download/v0.2.0/infer-osx-v0.2.0.tar.xz"
+  sha256 "ef755c3f01ae3246cae8f67a914518cdff72d69d55172a45242ef11e2c62812f"
+
+  def install
+    libexec.install Dir["*"]
+    bin.install_symlink Dir["#{libexec}/infer/infer/bin/infer"]
+  end
+
+  test do
+    expected = <<-EOS.undent
+      Starting analysis (Infer version v0.2.0)
+      Analysis done
+
+      1 file analyzed
+
+
+      #{libexec}/infer/examples/Hello.java:4: error: NULL_DEREFERENCE
+        object s last assigned on line 3 could be null and is dereferenced at line 4
+    EOS
+    actual = shell_output("#{bin}/infer -- javac #{libexec}/infer/examples/Hello.java").chomp
+    assert_equal expected, actual
+
+    expected = <<-EOS.undent
+      Starting analysis (Infer version v0.2.0)
+      Analysis done
+
+      1 file analyzed
+
+
+      #{libexec}/infer/examples/Hello.m:10: error: NULL_DEREFERENCE
+        pointer hello last assigned on line 9 could be null and is dereferenced at line 10, column 12
+    EOS
+    actual = shell_output("#{bin}/infer -- clang -c #{libexec}/infer/examples/Hello.m").chomp
+    assert_equal expected, actual
+
+    expected = <<-EOS.undent
+      Starting analysis (Infer version v0.2.0)
+      Analysis done
+
+      1 file analyzed
+
+
+      #{libexec}/infer/examples/hello.c:5: error: NULL_DEREFERENCE
+        pointer s last assigned on line 4 could be null and is dereferenced at line 5, column 3
+    EOS
+    actual = shell_output("#{bin}/infer -- gcc -c #{libexec}/infer/examples/hello.c").chomp
+    assert_equal expected, actual
+  end
+end


Library/Formula/infer.rb
```diff
@@ -0,0 +1,60 @@
+class Infer < Formula
+  desc "Facebook Infer is a static analysis tool - if you give Infer some Objective-C, Java, or C code, it produces a list of potential bugs."
+  homepage "https://fbinfer.com/"
+  url "https://github.com/facebook/infer/releases/download/v0.1.0/infer-osx-v0.1.0.tar.xz"
```
mistydemeo: Can this be built from source instead?

Library/Formula/infer.rb
```diff
@@ -0,0 +1,60 @@
+class Infer < Formula
+  desc "Facebook Infer is a static analysis tool - if you give Infer some Objective-C, Java, or C code, it produces a list of potential bugs."
```
bfontaine: Please use a shorter description. The formula’s name and its description should fit under 80 characters.

mistydemeo: Do you mind shortening this description? "infer: DESCRIPTION" needs to fit in 80 characters.

Library/Formula/infer.rb
```diff
@@ -0,0 +1,60 @@
...
+    bin.install_symlink prefix/"infer"
+    libexec.install Dir["*"]
+    bin.install_symlink Dir["#{libexec}/infer/infer/bin/infer"]
+    inreplace "#{libexec}/infer/infer/bin/infer", 'os.path.dirname(__file__)', 'os.path.dirname(os.path.realpath(__file__))'
+  end
+
+  test do
+    # Java test
+    expected = <<eos
+Starting analysis (Infer version v0.1.0)
```
mistydemeo: You can indent this at the same level as the parent for readability and use the `undent` method on the string:

```ruby
  expected = <<-EOS.undent
  ...
  EOS
```

Library/Formula/infer.rb
dambrisco: Must've missed that in the docs somewhere. Will correct.

Library/Formula/infer.rb
dambrisco: Most likely, although I'm currently at 10 minutes and counting for the source build (by comparison, the drop-in takes seconds). I'm happy to look into changing it over if that doesn't seem too lengthy.

Library/Formula/infer.rb
dambrisco: Thanks for the tip.

Library/Formula/infer.rb
mistydemeo: That's fine - our buildbots will produce our own binaries.

dambrisco: Working on getting that changed over, but it'll likely be a good minute before I've got it. As it stands, there might be issues related to how opam and Infer's Makefiles expect your environment to be configured. Ideally, I'll need to find a way to avoid modifying to user's `.ocamlinit` file while not breaking ocamlfind.

mistydemeo: Thanks!

The [emily formula](https://github.com/mistydemeo/homebrew-emily/blob/master/emily.rb) might be a good reference - it sets up a private opamroot within the build directory and installs all its dependencies, including ocamlfind, there.

dambrisco: Ah, awesome. Should help me get something put together for a source build sooner than later.

dambrisco:
Just an update on this: building from source is held up on one of two errors right now, depending on whether I specify a `--comp=4.01.0` flag in `opam init` or not. If specified, the init command itself fails in `make world` due to `Error: Unbound module Odoc_info`. If not specified, the command runs all the way through to `./compile-fcp.sh` due to, as far as I can tell, a mismatch of C++ libraries.

Here's both formulas in their entirety: https://gist.github.com/dambrisco/bd3009d2113e165532c0

mistydemeo:
Can you provide logs from `brew gist-logs infer`? Might be able to help diagnose.

mistydemeo:
In the `Error: Unbound module Odoc_info` case, it appears that opam is trying to build ocaml 4.01 itself before progressing, and that failed.

dambrisco:
Building ocaml 4.01.0 via `opam init` with the same flags and environment variables set (under `/private/tmp`) seems to work just fine omm, but continues to fail when run within homebrew. My next step will likely be setting up a mock forumla to test just the opam build and hopefully determine some sort of solution.

dambrisco:
Alright, so I decided to go back to the basics. `--env=std` allows `opam init` to complete successfully which led me to ENV.depar/ENV.no_opt and that also allows it to complete successfully. I'm now adding steps back one by one and, fingers crossed, will have a from-source update here later today.

Library/Formula/infer.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    1 file analyzed
+
+
+    #{libexec}/infer/examples/Hello.m:10: error: NULL_DEREFERENCE
+       pointer hello last assigned on line 9 could be null and is dereferenced at line 10, column 12
+    EOS
+    actual = shell_output("#{bin}/infer -- clang -c #{libexec}/infer/examples/Hello.m").chomp
+    assert_equal expected, actual
+    # end Obj-C test
```
mikemcquaid: Remove the comments, thanks.

Library/Formula/infer.rb
```diff
@@ -0,0 +1,60 @@
...
+    #{libexec}/infer/examples/Hello.java:4: error: NULL_DEREFERENCE
+       object s last assigned on line 3 could be null and is dereferenced at line 4
+    EOS
+    actual = shell_output("#{bin}/infer -- javac #{libexec}/infer/examples/Hello.java").chomp
+    assert_equal expected, actual
+    # end Java test
+
+    # Obj-C test
+    expected = <<-EOS.undent
+    Starting analysis (Infer version v0.1.0)
```
mikemcquaid: Indent the `EOS.undent`s contents by two characters more in each case.

Library/Formula/infer.rb
```diff
@@ -0,0 +1,60 @@
...
+  desc "Infer is a static analysis tool that can analyze Java, Obj-C, and C"
+  homepage "https://fbinfer.com/"
+  url "https://github.com/facebook/infer/releases/download/v0.1.0/infer-osx-v0.1.0.tar.xz"
+  sha256 "7a611989ceb4ece0882b8f62ed574fa0e59b48425cede20144b854a0918e1757"
+
+  def install
+    bin.install_symlink prefix/"infer"
+    libexec.install Dir["*"]
+    bin.install_symlink Dir["#{libexec}/infer/infer/bin/infer"]
+    inreplace "#{libexec}/infer/infer/bin/infer", "os.path.dirname(__file__)", "os.path.dirname(os.path.realpath(__file__))"
```
mikemcquaid: Split this onto multiple lines.

DomT4:
Ping.

dambrisco:
Accidentally closed this while working on getting the binary-only formula upgraded to 0.2.0 - re-opened with all requested changes from @mikemcquaid 

I've given up on getting infer building from source within the Homebrew environment. I'm not nearly familiar enough with clang to understand what environment variables might be causing the issue - even after diffing a bunch of printenvs.

mshibanami:
Facebook removed the binary...

dambrisco:
Well, alrighty then. Closing for now, then. I'll re-open if I can get it building from source.

PR #38267, opened 2015-04-01
Migration to GitHub, other URL fixes, lorem 0.7.4, remove iphotoexport, etc.

Noctem:
* migrate from Google Code to GitHub
  * `chibi-scheme`, `crf++`, `csshx`, `fdupes`, `git-multipush`, and `psgrep`
  * `innotop`
    * update `DBD::mysql` to latest stable release
  * `libewf`
    * Add experimental release as `--devel`
  * `lorem`
    * Update to latest stable release
    * Update test
  * `crush-tools` and `mfoc`
    * build with `AutoTools` since the configure scripts are not included in their GitHub tarballs
  * `ppss`, `uchardet`, and `uim`
    * add head
* `iphotoexport`: delete and migrate to boneyard
  * This was replaced by PhoShare in 2010, and [PhoShare ceased development in 2012](https://groups.google.com/forum/?fromgroups=#!topic/phoshare-users/moWsMcD5SdQ).
  * homebrew/homebrew-boneyard#42
* Fix HEAD URLs for `backupninja`, `depqbf`, and `screenfetch`
* Set homepage for `debianutils` to a package description instead of gitweb
* Migrate `dfu-util` URLs to SourceForge
* Modernize all the things

PR #40604, opened 2015-06-12
[WIP] uim 1.8.6

L2G:
This is not ready for prime time; I'm only making a pull request to give it visibility while it's being worked on.

The version in core right now is 1.6.0 and also does not build, at least not on my 10.9 Mavericks system.

Note that there is another, older PR (#38267) that also modifies this formula (it adds a HEAD build).

Of course this will be condensed into a single commit when it's "done."

### DIFF
--- a/Library/Formula/uim.rb
+++ b/Library/Formula/uim.rb
@@ -6,6 +6,7 @@ class Uim < Formula
 
   depends_on "pkg-config" => :build
   depends_on "gettext"
+  depends_on "intltool"
 
   def install
     system "./configure", "--disable-debug", "--disable-dependency-tracking",
--- a/Library/Formula/uim.rb
+++ b/Library/Formula/uim.rb
@@ -1,17 +1,15 @@
-require 'formula'
-
 class Uim < Formula
   desc "Multilingual input method library"
-  homepage 'http://code.google.com/p/uim/'
-  url 'https://uim.googlecode.com/files/uim-1.6.0.tar.bz2'
-  sha1 'd27f2ca8136da0702c82f0522911d06b2b8f8ea7'
+  homepage "http://code.google.com/p/uim/"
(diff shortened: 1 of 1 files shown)


Library/Formula/uim.rb
```diff
@@ -1,17 +1,17 @@
...
-  homepage 'http://code.google.com/p/uim/'
-  url 'https://uim.googlecode.com/files/uim-1.6.0.tar.bz2'
-  sha1 'd27f2ca8136da0702c82f0522911d06b2b8f8ea7'
+  homepage "http://code.google.com/p/uim/"
+  url "https://github.com/uim/uim/releases/download/uim-1.8.6/uim-1.8.6.tar.gz"
+  sha256 "0eebcf5b2192d0cd11018adbd69e358bfa062d3b5d63c1dd0b73d5f63eb7afe7"
 
-  depends_on 'pkg-config' => :build
-  depends_on 'gettext'
+  depends_on :x11
```
mikemcquaid: If this is mandatory this should be moved to `homebrew-x11`

L2G: I didn't become aware of it until I started working on it. I think it will indeed have to be moved over.

mikemcquaid: Cool, mind changing this PR to do that? Thanks!

L2G: No problem.

DomT4:
There's no other dependency usage in the core, so let's just tweak this PR to migrate this to X11 :+1:

mikemcquaid:
> There's no other dependency usage in the core, so let's just tweak this PR to migrate this to X11

:+1:

DomT4:
Ping? :)

bfontaine:
Closing due to the inactivity. Feel free to open another PR in the future!

PR #40606, opened 2015-06-12
tap_paths: prevent conflicted formulae happen in the same tap

xu-cheng:


chdiza:
I don't follow.  Does this mean that a tap `foo` is not allowed to have two formulas, each of with `conflicts_with` the other?  If so, why are we disallowing that?

xu-cheng:
No. This is preventing a tap has such invalid file structure.
```
<root>
|- foo.rb
|- Formula/foo.rb
```

mikemcquaid:
:+1:

PR #40608, opened 2015-06-12
tweak Tap class

xu-cheng:


Library/Homebrew/tap.rb
```diff
@@ -3,11 +3,19 @@ class Tap
 
   extend Enumerable
 
+  # The user name of this {Tap}.
```
mikemcquaid: Might be worth expanding in comments like this why a tap has a username.

Library/Homebrew/tap.rb
```diff
@@ -15,12 +23,18 @@ def initialize(user, repo, remote=nil)
...
+  # e.g. `https://github.com/user/homebrew-repo`
+  def remote
+    @remote ||= if (@path/".git").exist?
       @path.cd do
-        @remote = Utils.popen_read("git", "config", "--get", "remote.origin.url").chomp
+        Utils.popen_read("git", "config", "--get", "remote.origin.url").chomp
       end
     else
-      @remote = remote || "https://github.com/#{@user}/homebrew-#{@repo}"
+      @custom_remote || "https://github.com/#{@user}/homebrew-#{@repo}"
```
mikemcquaid: Hmm, I think it's probably just worth having this be an empty string or `N/A` or something rather than an (incorrect) URL.

xu-cheng: This code is also used to generate url for `brew tap` command. If you want `N/A` for unknown origin, the code would be some like this:
```ruby
if installed?
  if (@path/".git").exist?
   ... read from git ...
  else
    nil
  end
else
  @custom_remote || "https://github.com/#{@user}/homebrew-#{@repo}"
end
```

xu-cheng: Updated.

mikemcquaid: I guess it's used by `homebrew-bundle` for dumping etc.? Hmm. Could maybe return `nil` or raise an exception but either approach will need special-cased in `homebrew-bundle`.

Library/Homebrew/tap.rb
xu-cheng: I added a short explanation for tap class in general.

Library/Homebrew/tap.rb
xu-cheng: Sorry, but I mean it's used to generate the url to clone from during `brew tap homebrew/foo`. Code is [here](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/tap.rb#L20)

Library/Homebrew/tap.rb
mikemcquaid: I guess it's more a comment on inline API documentation in general. I think the variable names already explain most of these comments e.g. `user` in a `Tap` class is obviously the user name of the Tap but it's not obvious that this is a user on GitHub or used as the first part of `user/repo` or has to be populated even for a non-github user. Does that make sense? Those are the sorts of questions it's worth answering on API docs like this, I think. It takes a little longer but it means someone can jump into the `Tap` class without context and understand a lot more about what and why we're doing. Thanks for your work here, though, I hope that's not too rude to suggest this.

Library/Homebrew/tap.rb
mikemcquaid: Sure but that particular case will never be hit because it's already guarded by the tap being `installed?`, no?

xu-cheng: Now it will return `nil` if tap is installed and with unknown remote.

Library/Homebrew/tap.rb
xu-cheng: Updated a bit. I guess I am not good at writing document.

Library/Homebrew/tap.rb
```diff
@@ -1,13 +1,30 @@
+# a {Tap} is used to extend the formulae provided by Homebrew core.
+# Usually, it's synced with a remote git repository. And it's likey
+# a Github repository with the name of `user/homebrew-repo`. In such
+# case, `user/repo` will be used as the {#name} of this {Tap}, where
+# {#user} represents Github username and {#repo} represents repository
+# name without leading `homebrew-`.
```
mikemcquaid: :+1:

Library/Homebrew/tap.rb
mikemcquaid: You are good, I'm just picky. Much better now, thanks!

xu-cheng:
After some thinking, I found make remote method behavior differently based on whether tap is installed is kind of confusing.

I think it would be more clear if remote method only works on installed tap. I will try to make change on this tomorrow.

xu-cheng:
@mikemcquaid I updated again. Now `Tap#remote` will return nil for unknown origin and raise `TapUnavailableError` for uninstalled tap. I think in this way it makes more sense.

xu-cheng:
Comments before I merge this?

mikemcquaid:
Looks good :+1:

PR #40613, opened 2015-06-12
Weechat options + SHA-1 to SHA-256 change

Porkepix:
Change SHA-1 to SHA-256.
Added options to formula for most of supported languages for scripts (problems for v8/Javascript for now).

I would mention that on Linux the package is by default with all those languages for scripts supported by default. It could be nice to have the same with Homebrew, but it would require to have some dependencies not optional anymore: guile, lua, python, maybe curl.

I don't know enough how bottle building works to know if bottle's binary could have this by default.

### DIFF
--- a/Library/Formula/weechat.rb
+++ b/Library/Formula/weechat.rb
@@ -30,12 +30,13 @@ class Weechat < Formula
 
     args = std_cmake_args
 
-    args << "-DENABLE_LUA=OFF"    if build.without? "lua"
-    args << "-DENABLE_PERL=OFF"   if build.without? "perl"
-    args << "-DENABLE_RUBY=OFF"   if build.without? "ruby"
+    args << "-DENABLE_LUA=OFF" if build.without? "lua"
+    args << "-DENABLE_PERL=OFF" if build.without? "perl"
+    args << "-DENABLE_RUBY=OFF" if build.without? "ruby"
     args << "-DENABLE_ASPELL=OFF" if build.without? "aspell"
-    args << "-DENABLE_GUILE=OFF"  if build.without? "guile"
+    args << "-DENABLE_GUILE=OFF" if build.without? "guile"
     args << "-DENABLE_PYTHON=OFF" if build.without? "python"
+    args << "-DENABLE_JAVASCRIPT=OFF"
 
     mkdir "build" do
       system "cmake", "..", *args
--- a/Library/Formula/weechat.rb
+++ b/Library/Formula/weechat.rb
@@ -2,7 +2,7 @@ class Weechat < Formula
   desc "Extensible IRC client"
   homepage "https://www.weechat.org"
   url "https://weechat.org/files/src/weechat-1.2.tar.gz"
-  sha1 "84b93c3a52f8762940380edd9a4eaa3432066351"
+  sha256 "0f9b00e3fe4d0a4e864111d4231e1756f7be5c1b2b6d17da43bd785ab9f035d8"
 
   head "https://github.com/weechat/weechat.git"
 


Library/Formula/weechat.rb
```diff
@@ -30,12 +34,13 @@ def install
 
     args = std_cmake_args
 
-    args << "-DENABLE_LUA=OFF"    if build.without? "lua"
-    args << "-DENABLE_PERL=OFF"   if build.without? "perl"
-    args << "-DENABLE_RUBY=OFF"   if build.without? "ruby"
-    args << "-DENABLE_ASPELL=OFF" if build.without? "aspell"
-    args << "-DENABLE_GUILE=OFF"  if build.without? "guile"
-    args << "-DENABLE_PYTHON=OFF" if build.without? "python"
+    args << "-DENABLE_LUA=OFF"        if build.without? "lua"
```
DomT4: Reduce the spacing on these args: `args << "-DENABLE_LUA=OFF" if build.without? "lua"` is fine.

Library/Formula/weechat.rb
```diff
@@ -14,6 +14,10 @@ class Weechat < Formula
 
   option "with-perl", "Build the perl module"
   option "with-ruby", "Build the ruby module"
+  option "with-python", "Build the python module"
```
DomT4: All of these, aside `tcl` are already implicitly stated below. The `optional` dependencies are essentially `option "with-python"` - Is there an upside/downside to adding more explicit options?

Library/Formula/weechat.rb
Porkepix: Ok, so this is because I didn't canceled the added spacing: at basis, I added them because I added a line `args << "-DENABLE_JAVASCRIPT=OFF"        if build.without? "javascript"`. I had to remove this one because there is an issue to build weechat with JS for now (I can't achieve to build with homebrew's packaged in 3.x, the simplest way will probably be to wait for weechat 1.3 and support for 4.3 and more).
Is this a real problem to keep those spaces while Javascript isn't back?

Library/Formula/weechat.rb
Porkepix: It looks like by default support for those scripts were not here, for every language (which surprised me quite a bit, as stated in first message, as on Linux it's compiled with all of them by default).
That's why I added explanation for all of them to allow to compile with them (see line 38 to 44 asking to remove language support if option do not ask for this language).

By the way, what about building with them by default? Maybe can bottle build have them by default, as Linux packages do have it?

Library/Formula/weechat.rb
DomT4: We don't really care about keeping lines the exact same length, and generally frown on unnecessary whitespacing which is why I'm a bit :-1: on it.

Library/Formula/weechat.rb
DomT4: > That's why I added explanation for all of them to allow to compile with them

The process doesn't change for the end-user whether we add the explicit options or not - Currently if you want a "full" install of `weechat` you do `brew install weechat --with-perl --with-ruby --with-python --with-lua --with-guile --with-curl` - Nothing about that changes even if we make the `:optional` dependencies more explicit.

> By the way, what about building with them by default? 

We tend not to do this, because on a clean machine that would mean instead of getting 3 dependencies prior to being able to install `weechat` by default, you'd end up with 8, which is quite a bit more. We try to keep mandatory dependencies to a minimum unless there's significant popularity/demand for an option.

Porkepix:
So, removed most of the changes looking at your explanations and some other I had.
I just: added one of the line about Javascript as I discussed with weechat developers: in the current state it would be painful to build with it on OS X and could lead to some build fail depending on what cmake will detect, so the line is here to skip JS support into cmake. It's for me a security on the manual build looking at the tests I did.
Easy support with v8 homebrew package should be here with weechat 1.3 as there will have changes to the way it's handled for newer version.

I took the liberty to remove "unnecessary spaces" too due to what you and another person from homebrew team told me.

Let me know if those idea were not good, and in this case I'm just going to cancel them and let SHA change and that's all.

Library/Formula/weechat.rb
```diff
@@ -30,12 +30,13 @@ def install
...
-    args << "-DENABLE_PERL=OFF"   if build.without? "perl"
-    args << "-DENABLE_RUBY=OFF"   if build.without? "ruby"
+    args << "-DENABLE_LUA=OFF" if build.without? "lua"
+    args << "-DENABLE_PERL=OFF" if build.without? "perl"
+    args << "-DENABLE_RUBY=OFF" if build.without? "ruby"
     args << "-DENABLE_ASPELL=OFF" if build.without? "aspell"
-    args << "-DENABLE_GUILE=OFF"  if build.without? "guile"
+    args << "-DENABLE_GUILE=OFF" if build.without? "guile"
     args << "-DENABLE_PYTHON=OFF" if build.without? "python"
+    args << "-DENABLE_JAVASCRIPT=OFF" if build.without? "javascript"
```
bfontaine: You can remove the `if build.without? "javascript"` because there’s no `with-javascript` option, so it’ll always be true. The line will then be:

```ruby
args << "-DENABLE_JAVASCRIPT=OFF"
```

Porkepix:
So, this change is now done.

I will just add that, thinking again about it, support for Ruby and Perl at least should maybe be here in every case excepted if stated otherwise: this is present on OS X by default and do not need to be installed.

I can do those changes by removing corresponding lines if it would be a good idea. This can be done in another PR.
I guess that for now all is fine in this PR.

bfontaine:
Thank you for your contribution to Homebrew, @Porkepix! :tada: 

For future reference the preferred commit message format is “`weechat: change from SHA-1 to SHA-256`”

PR #40596, opened 2015-06-11
apache-spark 1.4.0

karlhigley:
Apache Spark has a new release, 1.4.0: [spark release blog post](https://spark.apache.org/news/spark-1-4-0-released.html)

PR #40616, opened 2015-06-12
apache-spark 1.4.0

ogrisel:
Amended #40596 to use an https URL.


ogrisel:
Hum, I just tested it locally and the closer.cgi URL of apache.org are not interpreted by the homebrew. Is there a good way to deal with the apache mirrors in homebrew formulae?

ogrisel:
Here is the error message I get:

```
$ brew fetch apache-spark
==> Downloading https://www.apache.org/dyn/closer.cgi/spark/spark-1.4.0/spark-1.4.0-bin-hadoop2.6.tgz
Error: Failed to download resource "apache-spark"
Download failed: Couldn't determine mirror, try again later.
```

ogrisel:
I have forced-pushed a new version of the Formula that uses `https://archive.apache.org/dist/` instead of `https://www.apache.org/dyn/closer.cgi/` and it seems to work.

There might be an issue with the handling of the apache mirror. But this problem does not happen for other apache projects such as apache-drill for instance.

karlhigley:
@ogrisel: The [Hadoop recipe](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/hadoop.rb) uses a slightly different URL format, which seems to make the mirrors work.  Updated in #40596.

DomT4:
Let's not add a third open PR on top of the other two for this formula, as much as we appreciate the enthusiasm :wink:.

ogrisel:
But mine works ;)

ogrisel:
I don't understand the report by jenkins. All tests pass. The failure on mavericks seems to be caused by a git checkout issue:

http://bot.brew.sh/job/Homebrew%20Pull%20Requests/27717/version=mavericks/console

DomT4:
^ Jenkins crashed, essentially.

Issue #40359, opened 2015-06-04
llvm 3.6.1 fails due to lldb_codesign: no identity found


mjbshaw:
Updating to the latest llvm (3.6.1) fails. `brew upgrade llvm` errors out with the message ``couldn't understand kern.osversion `14.3.0'``, but the logs show the actual error is `lldb_codesign: no identity found`.

Note that I'm building with the options `--with-lld, --with-clang, --with-lldb` (presumably it's the `--with-lldb` option that's causing the failure, but I haven't tried to narrow down the issue because building takes forever and burns up my laptop; I can do more target testing if needed, though).

[gist of the issue can be found here](https://gist.github.com/anonymous/920533dffaaa4ce1e3b5).

mjbshaw:
This might be a dupe of #39621, but I'm not 100% sure, so I'm logging it as a separate issue here.

DomT4:
Yeah, technically it's a dupe of existing issues. IMO this needs to go upstream, since I can't see there's anything we can do if there's an expectation of code-signing there.

CC @xu-cheng @tamird for prior discussions.

tamird:
+1 dupe

luzpaz:
+1 dupe

DomT4:
Closing this in favour of #39621.

Issue #39621, opened 2015-05-11
`brew install llvm --with-python --with-lldb --with-clang` fails


xu-cheng:
https://gist.github.com/xu-cheng/41ca3c4ad95a72351e39

cc @DomT4, @tamird

xu-cheng:
The real error:
```
/tmp/llvm20150511-61002-1y7lnwn/llvm-3.6.0.src/tools/lldb/tools/debugserver/source/MacOSX/dbgnub-mig.defs:13:10: fatal error: 'DNBConfig.h' file not found
#include "DNBConfig.h"
         ^
1 error generated.
```

DomT4:
Ugh. LLVM :sob:.

You found about the one clear reference that I did as well. The `debugserver` xcodeproject inside the directory it builds from contains this block:

```
/* Begin PBXShellScriptBuildPhase section */
		26CE05C7115C36870022F371 /* ShellScript */ = {
			isa = PBXShellScriptBuildPhase;
			buildActionMask = 2147483647;
			files = (
			);
			inputPaths = (
				"$(SRCROOT)/scripts/dbgnub-config.pl",
			);
			outputPaths = (
				"$(DERIVED_FILE_DIR)/DNBConfig.h",
			);
			runOnlyForDeploymentPostprocessing = 0;
			shellPath = /bin/sh;
			shellScript = "perl -x scripts/dbgnub-config.pl\n";
		};
```

Maybe that's not getting run, or maybe it's being placed in the wrong location? If I run the project inside `xcode` I get lots of `'DNBConfig.h' file not found` message as well, but could be unrelated. Presume this problem doesn't exist on the autotooled latest LLVM in Versions?

tamird:
Does this repro with just `brew install llvm --with-lldb`?

tamird:
This actually failed for me in a whole different way:
```
cd /tmp/llvm20150511-65141-166fd89/bin && CODESIGN_ALLOCATE=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/codesign_allocate codesign --force --sign lldb_codesign debugserver
lldb_codesign: no identity found
make[2]: *** [bin/debugserver] Error 1
make[1]: *** [tools/lldb/tools/debugserver/source/MacOSX/CMakeFiles/debugserver.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

tamird:
[...]

PR #40617, opened 2015-06-12
llvm: require lldb codesign verification

DomT4:
Requires the user to confirm they have an `lldb_codesign` identity available for LLDB to build successfully.

Whilst this isn't great UX it's better than allowing a ~90+ minute build to get to 90% and only error out at that point.

Workaround for #39621, #40359.

### DIFF
--- a/Library/Formula/llvm.rb
+++ b/Library/Formula/llvm.rb
@@ -1,3 +1,22 @@
+class CodesignRequirement < Requirement
+  include FileUtils
+  fatal true
+
+  satisfy do
+    mktemp do
+      touch "llvm_check.txt"
+      quiet_system "/usr/bin/codesign", "-s", "lldb_codesign", "llvm_check.txt"
+    end
+  end
+
+  def message
+    <<-EOS.undent
+      lldb_codesign identity must be available to build with LLDB.
+      See: https://llvm.org/svn/llvm-project/lldb/trunk/docs/code-signing.txt
+    EOS
+  end
+end
+
 class Llvm < Formula
   desc "llvm (Low Level Virtual Machine): a next-gen compiler infrastructure"
   homepage "http://llvm.org/"
@@ -79,7 +98,11 @@ class Llvm < Formula
     depends_on :python => :optional
   end
   depends_on "cmake" => :build
-  depends_on "swig" if build.with? "lldb"
+
+  if build.with? "lldb"
+    depends_on "swig"
+    depends_on CodesignRequirement
+  end
 
   keg_only :provided_by_osx
 
@@ -134,12 +157,8 @@ class Llvm < Formula
     end
 
     # install llvm python bindings
-    (lib+"python2.7/site-packages").install buildpath/"bindings/python/llvm"
-    (lib+"python2.7/site-packages").install buildpath/"tools/clang/bindings/python/clang" if build.with? "clang"
-  end
-
-  test do
-    system "#{bin}/llvm-config", "--version"
+    (lib/"python2.7/site-packages").install buildpath/"bindings/python/llvm"
+    (lib/"python2.7/site-packages").install buildpath/"tools/clang/bindings/python/clang" if build.with? "clang"
   end
 
   def caveats
@@ -147,5 +166,9 @@ class Llvm < Formula
       LLVM executables are installed in #{opt_bin}.
       Extra tools are installed in #{opt_share}/llvm.
     EOS
+  end
+
+  test do
+    system "#{bin}/llvm-config", "--version"
   end
 end


DomT4:
CC @Homebrew/owners @bfontaine @greymouser @tamird

xu-cheng:
Can we generate the codesign programly for the user?

tamird:
It is extremely non-trivial. The GUI gives you a bunch of defaults that aren't exposed in the command line equivalents, and the rabbit hole is deep.

LGTM.

DomT4:
You can do it via `certtool c` but it isn't easy, and predicting how long the wait would be between each `certtool` request for information would be difficult to remove the human element from.

xu-cheng:
Ok, if we cannot generate key for users, can we auto detect whether a key is present. Therefore, we can create a custom requirement. To me, I am kinda :-1: on using option in this way.

tamird:
Detecting the key's presence is also a PITA, and requires logic in homebrew to closely track logic in LLVM's build system.

xu-cheng:
@tamird actually I mean detecting the key previous building llvm

DomT4:
> can we auto detect whether a key is present

More or less:

```
security find-certificate -a ~/Library/Keychains/login.keychain
security dump-trust-settings | grep lldb_codesign
```

But that's presuming user has stuck cert in right place, and always named it consistent with upstream's demands. I'm a little hesitant to go sniffing around a user's login keychain without prior consent, to be honest.

tamird:
@DomT4 if you're OK with hardcoding `lldb_codesign` we can just try to sign a hello world program to verify.

mikemcquaid:
I think we should probably just remove the option, to be honest, if we can't programmatically detect it'll fail.

mikemcquaid:
> @DomT4 if you're OK with hardcoding lldb_codesign we can just try to sign a hello world program to verify.

This might be a good call, though, using a `Requirement`

DomT4:
> I think we should probably just remove the option, to be honest, if we can't programmatically detect it'll fail.

I tried to get some agreement on this, but there was considerable pushback.

> using a Requirement

Will look into this.

DomT4:
Alright, I've shoved this into a `Requirement`. I'm fairly sure I'm abusing the `env` block horribly here, but it works and if necessary we can focus on making it prettier now I've made it work.

CC @xu-cheng because he'll probably have a better way of writing everything I've done instantly. Sometimes I wonder why I spend time trying to write these things instead of just asking Xu for the *best* way to write them :stuck_out_tongue_winking_eye:.

Library/Formula/llvm.rb
```diff
@@ -1,3 +1,29 @@
...
+  fatal true
+
+  def setup
+    FileUtils.touch "/tmp/llvm_check.txt"
+  end
+
+  satisfy do
+    setup
+    quiet_system "/usr/bin/codesign", "-s", "lldb_codesign", "/tmp/llvm_check.txt"
+  end
```
xu-cheng:
```ruby
satisfy do
  FileUtils.mktemp do
    FileUtils.touch "llvm_check.txt"
    system "/usr/bin/codesign", "-s", "lldb_codesign", "llvm_check.txt"
  end
end
```

DomT4: Bingo! That's what I was looking for, but I couldn't remember that `mktemp` was a FileUtils thing :speak_no_evil:.

DomT4:
We may have to ignore the `audit` going forwards, or suppress this somehow, or just move it to an external requirement entirely.

```
llvm:
 * Don't need 'FileUtils.' before mktemp.
 * Don't need 'FileUtils.' before touch.
```

Library/Formula/llvm.rb
```diff
@@ -1,3 +1,21 @@
+class CodesignRequirement < Requirement
```
tamird: i'd probably make this an inner class of LLVM

xu-cheng:
> We may have to ignore the audit going forwards, or suppress this somehow, or just move it to an external requirement entirely.

Doesn't this help to fix the problem.
```ruby
class CodesignRequirement < Requirement
  include FileUtils
   ...
  satisfy do
     mktemp do
   ...
end
```

DomT4:
Just set the build running to make sure this actually works. Will report back in ~3 years or so, build should be done by then :wink:

DomT4:
> include FileUtils

Probably, but I wasn't sure how "legit" using `include` was inside formulae files?

DomT4:
Bingo. We have success:

```
==> Summary
🍺  /usr/local/Cellar/llvm/3.6.1: 1998 files, 572M, built in 59.5 minutes
```

```
~> codesign -vvvv /usr/local/Cellar/llvm/3.6.1/bin/debugserver
/usr/local/Cellar/llvm/3.6.1/bin/debugserver: valid on disk
/usr/local/Cellar/llvm/3.6.1/bin/debugserver: satisfies its Designated Requirement
```

:tada:

DomT4:
:rocket: in https://github.com/Homebrew/homebrew/commit/8da6986730e06b51a9686afa2ad9eeb241f8708a. I wanted to get this in ASAP given the level of issues we've had, and the horrible UX of getting to 92% and the build erroring out at that point.

PR #40626, opened 2015-06-12
Added libmustache in prep for adding php55-mustache.

Stubbs:
Added libmustache in prep for adding php55-mustache, the PHP Mustache template extension.

### DIFF
--- a/Library/Formula/libmustache.rb
+++ b/Library/Formula/libmustache.rb
@@ -1,0 +1,32 @@
+class Libmustache < Formula
+  desc "C++ implementation of Mustache."
+  homepage "https://github.com/jbboehr/libmustache"
+  url "https://github.com/jbboehr/libmustache/archive/v0.3.2.tar.gz"
+  sha256 "b74e5fed678572a5d8a02fced60bf1c2d9cbe8d432346f0803161280400d93ef"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+
+  def install
+    system "autoreconf", "-fvi"
+    system "./configure", "--disable-debug",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.cpp").write <<-EOS
+    #include <mustache/mustache_config.h>
+    #include <mustache/mustache.hpp>
+    int main() {
+        mustache::Mustache m;
+        mustache_version_int();
+        return 0;
+    }
+    EOS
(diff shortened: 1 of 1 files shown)


Library/Formula/libmustache.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def install
+    system "autoreconf", "-fvi"
+    system "./configure", "--disable-debug",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/mustache", "-h"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Stubbs:
With a little help from the upstream author I've added a small test source file that gets compiled, linked & executed, pretty similar to tinyxml2 ...

Stubbs:
Not really sure what to do about this last failure, when I add a "desc" field to the formula my audit throws an error `Error: undefined method `desc' for Formulary::Formulae::Libmustache:Class`, same if I use `description`.

DomT4:
Make sure your Homebrew is up-to-date and this branch is up-to-date with the Master branch.

Stubbs:
I thought I'd done that, but I hadn't then pulled those changes into my libmustache branch. I've rebased my branch and it's all gravy.

mikemcquaid:
@BrewTestBot test this please

Stubbs:
Is there anything else I need to do on this PR?

DomT4:
My concern was that it's a little niche and may be better in the `php` tap, but if another @Homebrew/owners signs off on it I'm happy to pull.

jacknagel:
True, the README does say "C++ implementation of Mustache intended mainly for use as a PHP extension".

DomT4:
@Homebrew/php Mind accepting this in `Homebrew/php` instead of the core?

alanthing:
@DomT4 If it's something that would generally only be used alongside the PHP extension and nothing else, sure.

DomT4:
@alanthing Per Jack's comment just above it seems to be intended by the developer to be useful for PHP. If someone can provide a compelling test-case of "I will use it on OSX for xyz that isn't PHP" we could consider it, but otherwise I think it's too niche for the core personally.

alanthing:
@DomT4 sounds great.

@Stubbs could you put this same formula in a PR for [Homebrew/homebrew-php](https://github.com/Homebrew/homebrew-php) ?

DomT4:
Thanks @alanthing, Appreciated. Closing here.

PR #40627, opened 2015-06-12
expat: sha256, add non-sf.net mirror

vszakats:


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/9a2cc241e917cf187ffe5ed517e976ee6da4f6d9. Thanks @vszakats!

In this case the commit message would have ideally been `expat: add mirror` - The location of the mirror/usage of sha256 isn't something we necessarily want to encourage a lot of PRs for.

vszakats:
It's so difficult to please you guys, but noted :) I was thinking that shouldn't useful (f.e. passing the actual strict audit, or security updates like a hash upgrade), be better tolerated now, that those won't show up on users' screen on a `brew update`?

mikemcquaid:
@vszakats We want to avoid unnecessary code churn or we'd have just used Rubocop to automatically fix all style violations.

PR #40628, opened 2015-06-12
Add cjdns project to homebrew, version number 16.1

cjdelisle:
This was discussed before in #22282 but now we have somewhat proper version numbers and things have reached some semblance of stability.
FTR: https://github.com/cjdelisle/cjdns https://en.wikipedia.org/wiki/Cjdns

### DIFF
--- a/Library/Formula/cjdns.rb
+++ b/Library/Formula/cjdns.rb
@@ -1,5 +1,5 @@
 class Cjdns < Formula
-  desc "Advanced mesh routing with cryptographic addressing and integrity/confidentiality protection"
+  desc "Advanced mesh routing system with cryptographic addressing"
   homepage "https://github.com/cjdelisle/cjdns/"
   url "https://github.com/cjdelisle/cjdns/archive/cjdns-v16.1.tar.gz"
   sha256 "3eebb276ee63f103c54aa9201e67d8d810ae3a88af90300dd3a3a894e73cb400"
@@ -7,7 +7,7 @@ class Cjdns < Formula
   depends_on "node" => :build
 
   def install
-    system './do'
-    bin.install 'cjdroute'
+    system "./do"
+    bin.install "cjdroute"
   end
 end
--- a/Library/Formula/cjdns.rb
+++ b/Library/Formula/cjdns.rb
@@ -9,10 +9,10 @@ class Cjdns < Formula
   def install
     system "./do"
     bin.install "cjdroute"
-    bin.install "build_darwin/test_testcjdroute_c" => "cjdroute_test"
+    (share+"test").install "build_darwin/test_testcjdroute_c" => "cjdroute_test"
   end
 
   test do
-    system bin/"cjdroute_test", "all"
+    system "#{share}/test/cjdroute_test", "all"
   end
 end
--- a/Library/Formula/cjdns.rb
+++ b/Library/Formula/cjdns.rb
@@ -9,12 +9,10 @@ class Cjdns < Formula
   def install
     system "./do"
     bin.install "cjdroute"
-    mv "./build_darwin/test_testcjdroute_c", "./cjdroute_test"
-    bin.install "cjdroute_test"
+    bin.install "build_darwin/test_testcjdroute_c" => "cjdroute_test"
   end
 
   test do
-    system "cjdroute_test all"
+    system bin/"cjdroute_test", "all"
   end
-
 end
--- a/Library/Formula/cjdns.rb
+++ b/Library/Formula/cjdns.rb
@@ -1,0 +1,13 @@
+class Cjdns < Formula
+  desc "Advanced mesh routing with cryptographic addressing and integrity/confidentiality protection"
+  homepage "https://github.com/cjdelisle/cjdns/"
+  url "https://github.com/cjdelisle/cjdns/archive/cjdns-v16.1.tar.gz"
+  sha256 "3eebb276ee63f103c54aa9201e67d8d810ae3a88af90300dd3a3a894e73cb400"
+
+  depends_on "node" => :build
+
+  def install
+    system './do'
+    bin.install 'cjdroute'
+  end
+end
--- a/Library/Formula/cjdns.rb
+++ b/Library/Formula/cjdns.rb
@@ -9,5 +9,12 @@ class Cjdns < Formula
   def install
     system "./do"
     bin.install "cjdroute"
+    mv "./build_darwin/test_testcjdroute_c", "./cjdroute_test"
+    bin.install "cjdroute_test"
   end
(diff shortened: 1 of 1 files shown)


Library/Formula/cjdns.rb
```diff
@@ -0,0 +1,13 @@
+class Cjdns < Formula
+  desc "Advanced mesh routing with cryptographic addressing and integrity/confidentiality protection"
```
DomT4: This may be too long. The `desc` is capped at 80 characters.

Library/Formula/cjdns.rb
```diff
@@ -0,0 +1,13 @@
...
+  homepage "https://github.com/cjdelisle/cjdns/"
+  url "https://github.com/cjdelisle/cjdns/archive/cjdns-v16.1.tar.gz"
+  sha256 "3eebb276ee63f103c54aa9201e67d8d810ae3a88af90300dd3a3a894e73cb400"
+
+  depends_on "node" => :build
```
DomT4: Node can be removed from the system after install and it'll run fine?

Library/Formula/cjdns.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  def install
+    system './do'
```
DomT4: Double quotes around everything.

Library/Formula/cjdns.rb
```diff
@@ -0,0 +1,13 @@
...
+    bin.install 'cjdroute'
+  end
+end
```
DomT4: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

cjdelisle:
To respond to your line notes all at once, fixing the "typos", node is the build system, cjdns is pure C (with some ejs templates but that's also build time), `./do` runs a test suite and will fail the build. Testing the TUN infrastructure's interaction with the kernel is possible if I can rely on root access, if this is the case I can write a test function.

DomT4:
> fixing the "typos", node is the build system

:+1:.

> ./do runs a test suite and will fail the build.

Appreciate this, but ideally we want a small `test do` block to ensure it'll run after installation as well. I'm afraid there's no access to root, but a simple test is better than no test.

cjdelisle:
```
  def test
    # ./do both compiles and tests         
    system "./do"
  end
```
Ok for you ?

DomT4:
> Ok for you ?

It wouldn't work, I'm afraid. The `do` script won't exist in the `testpath` - It'll be sandboxed into the `buildpath`.

cjdelisle:
right, got ahead of myself.. I'm installing the test executable which is technically internal to the build but it's a tag so it should be safe...

Library/Formula/cjdns.rb
```diff
@@ -0,0 +1,20 @@
...
+  homepage "https://github.com/cjdelisle/cjdns/"
+  url "https://github.com/cjdelisle/cjdns/archive/cjdns-v16.1.tar.gz"
+  sha256 "3eebb276ee63f103c54aa9201e67d8d810ae3a88af90300dd3a3a894e73cb400"
+
+  depends_on "node" => :build
+
+  def install
+    system "./do"
+    bin.install "cjdroute"
+    mv "./build_darwin/test_testcjdroute_c", "./cjdroute_test"
```
DomT4: You can turn the `mv` and `bin.install` into one fluid motion:

```ruby
bin.install "build_darwin/test_testcjdroute_c" => "cjdroute_test"
```

Library/Formula/cjdns.rb
```diff
@@ -0,0 +1,20 @@
...
+    bin.install "cjdroute_test"
+  end
+
+  test do
+    system "cjdroute_test all"
```
DomT4: `system bin/"cjdroute_test", "all"`

Library/Formula/cjdns.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
```
DomT4: Can kill the blank line here.

DomT4:
> I'm installing the test executable which is technically internal to the build but it's a tag so it's not going to break...

Sounds great. There's a few formulae we have to do this for, so if it works we're fine with it.

Library/Formula/cjdns.rb
```diff
@@ -0,0 +1,18 @@
...
+  homepage "https://github.com/cjdelisle/cjdns/"
+  url "https://github.com/cjdelisle/cjdns/archive/cjdns-v16.1.tar.gz"
+  sha256 "3eebb276ee63f103c54aa9201e67d8d810ae3a88af90300dd3a3a894e73cb400"
+
+  depends_on "node" => :build
+
+  def install
+    system "./do"
+    bin.install "cjdroute"
+    bin.install "build_darwin/test_testcjdroute_c" => "cjdroute_test"
```
mikemcquaid: Sorry, can you install this to `share` instead? Thanks!

cjdelisle: I don't know what that means, sorry I'm really new to OSX.. and I'm about to wipe my laptop again (hopefully last time)... Is it possible that you can patch in the change when taking the PR?

mikemcquaid: Change this line to `(share+"test").install "build_darwin/test_testcjdroute_c" => "cjdroute_test"`

Library/Formula/cjdns.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  test do
+    system bin/"cjdroute_test", "all"
```
mikemcquaid: and this line to `system "#{share}/test/cjdroute_test", "all"`

Unfortunately I can't do this for you as our CI needs to generate the binary package and I can't edit your PR.

cjdelisle: Ok, it will have to wait until at least tonight but I can paste those lines as requested.

mikemcquaid: Great, thanks. You could always edit it in the GitHub Web UI?

cjdelisle: done. Say a hail mary...

mikemcquaid:

For future reference the preferred commit message format for new formulae is `cjdns 16.1 (new formula)`.

cjdelisle:
Well on my side all I have to say is Homebrew is the most friendly project I've ever had the privilege of contributing to and I'm certainly going to think about taking your way of working back where I come from.

mikemcquaid:
Wow, that's very kind if you to say. Thank you for that!

PR #40631, opened 2015-06-13
fontforge 20150612

DomT4:


DomT4:
I've asked @tdsmith for some advice on the unusual Python situation here, so let's not merge this without his okay.

Library/Formula/fontforge.rb
```diff
@@ -33,7 +33,10 @@ class Fontforge < Formula
   depends_on "giflib" => :optional
   depends_on "libspiro" => :optional
   depends_on :x11 => :optional
-  depends_on :python if MacOS.version <= :snow_leopard
+
+  # Segfaults on import without Homebrew's Python
+  # https://github.com/fontforge/fontforge/issues/2353
+  depends_on :python
```
tdsmith: Needs to be "python" or we need to work around their detection code :/

DomT4: Welp. I tore out giant chunks of their detection code in `configure` with no luck whatsoever, I still got insta-segfault every time I tried to `import fontforge` with any Python that wasn't `/usr/bin/python`.

tdsmith: Right, that will fail until we fix the framework linkage, which isn't related to the way they sniff Homebrew.

tdsmith: Which we can do like this, maybe: https://github.com/Homebrew/homebrew/compare/master...tdsmith:fontforge-python

apjanke:
FYI: upstream patch for Python detection under homebrew may simplify this slightly: fontforge/fontforge#2383

tdsmith:
Thanks. I appreciate why they added it but I think detecting a Homebrew build is misguided; I need to submit a patch to remove the detection code and change the way they link Python on OS X (which will look roughly like the changes here: https://github.com/Homebrew/homebrew/compare/master...tdsmith:fontforge-python -- linking Python only to things that are trying to embed an interpreter), which should solve all of the problems that have come up.

apjanke:
Great. That linkage issue is beyond my level of understanding, but it'd be nice to get it fixed: I'm just starting to work with FontForge, and it'd be nice to manage it with Homebrew.

DomT4:
CC @davelab6. We're waiting partially on this, but at the moment the FontForge detection code for Homebrew's Python is causing issues for us. It has been for a while, since FontForge's Python linkage stopped being optional (Was a while ago now) and was switched to system Python there have been some `import` problems in Python.

Whilst before the most recent Homebrew Python detection patch was merged upstream at FontForge the link was semi-working, but as you've noticed now with the `HEAD` build it's gone the way of the dodo.

Part of the conflict is that Homebrew's Python methodology is to build everything against the system Python so users don't have to carry around a clunky extra Python dependency just for certain formulae (Amongst other reasons), but when FontForge detects a Homebrew environment it attempts to latch onto Homebrew's Python.

You can see some of Tim's patching in the https://github.com/Homebrew/homebrew/compare/master...tdsmith:fontforge-python change.

DomT4:
Closing. Version bump will either go in with #41670 or I'll tee up a new PR shortly after that is merged.

davelab6:
Thanks @DomT4 :)

Issue #40392, opened 2015-06-04
Broken URLs


bfontaine:
I ran `brew fetch --build-from-source` on all formulae and found the following issues:

- [x] `burp` (404) #40393
- [x] `colortail` (domain down) #40428
- [x] `daq` (404) #40430
- [x] `davix` (repo not found) #40432
- [x] `ffmpeg2theora` (SHA1 error on patch) #40757
- [x] `fits` (404) #40434
- [x] `fuseki` (404) https://github.com/Homebrew/homebrew/commit/66fdc20fa724c793f8bd71d2d7d168465581d951
- [x] `gant` (404) https://github.com/Gant/Gant/issues/3, #43326
- [x] `hh` (404) #40757
- [x] `httpry` (network is unreachable) #40757
- [x] `libopendkim` (404) #40757
- [x] `libtar` (SHA1 error) #40757
- [x] `libtomcrypt` (domain down) #40757
- [x] `libtommath` (domain down) #40757
- [x] `libtorrent` (timeout) #40460
- [x] `logentries` (SHA1 error) #41562
- [x] `luvit` (404) #42525
- [x] `makepkg` (domain down) #41562
- [x] `mitmproxy` (404)
- [x] `mpssh` (404 on patch) #40390
- [x] `openrtsp` (404) #41562
- [x] `openwsman` (SHA256 error) #41562
- [x] `owamp` (404 on patch) (I sent an email to the patch author on 2015/07/12) #43361
- [x] `par2tbb` (SHA256 error then domain down) #43360
- [x] `pdftoipe` (404) #40528
- [x] `qi` (404) #41366
- [x] `redo` (SHA1 error) [no error as of 2015/08/27]
- [x] `rocket` (SHA1 error) #41364
- [x] `rtorrent` (timeout) #40489
- [x] `sdelta3` (domain down) #40436
- [x] `sfk` (SHA256 error) #41562
- [x] `snort` (404) #40461
- [x] `ssldump` (404) #41297
- [x] `tcsh` (ftp 550) #40462
- [x] `uni2ascii` (SHA256 error) #40466
- [x] `unison` (SHA1 error) #40711

I checked four times each one of these, plus checked from another server (different ISP, different city) for the checksum errors.
[See the detailed output in this gist](https://gist.github.com/bfontaine/65b25faa1b793ff9e54f).

Last update: 2015/08/28.

xu-cheng:
:+1: Good work.

[...]

PR #40633, opened 2015-06-13
libb2: pass strict audit

vszakats:


DomT4:
We don't really do style/checksum changes only :cry:

vszakats:
Protecting a formula implementing a secure hash with a less secure hash looked odd, hence my update ;)

vszakats:
Just found this article from @sneves &ndash; author of the very source code included in this formula &ndash;:
[SHA-1 Considered Harmful Today](https://eden.dei.uc.pt/~sneves/sha1.html)

bfontaine:
@DomT4 We don’t do style-only but I think we accept sha1→sha256 -only changes. I got a lot of PRs merged just for that.

vszakats:
I got a hash update accepted just recently as well. It was in a formula where a single resource (of several) was overlooked and stuck at `sha1`. I understand that purely upping `sha1` to `sha256` with no other reason is not encouraged though.

[ it would be _nice_ to up them all in batch, like `desc` was added :o - it also wouldn't pollute the `brew update` output after c010cc8a8cadfe2e29f541570b70e62b6d735fb9 - but I may be banging open doors here, seeing the efforts done in https://github.com/Homebrew/homebrew/issues/40392 ]

vszakats:
As an aside, I contacted the author team via e-mail about an argument check typo/bug in one of the reference implementation.

If their team decide to release a new version just for that, we could do a version bump. I'm somewhat sceptic though, because the reference code is continued to be updated [here](https://github.com/BLAKE2/BLAKE2), with none of those updates retrofitted into the official download `url` as of yet. We'll see.

DomT4:
>  author of the very source code included in this formula: SHA-1 Considered Harmful Today

I'm aware SHA1 isn't great. I wouldn't have pushed the bottle change many months ago otherwise, but to date at least Mike and Xu have been reluctant to merge in checksum only changes.

Can one of @Homebrew/owners weigh in on this? Are we :+1: / :-1: on checksum-only changes?

vszakats:
Should this be a hard-rule? I understand that a landslide of hash-only updates are something not encouraged due to the load it'd put on maintainers, but is there anything fundamentally wrong (or else) with it to deny committing them in cases where it could make sense? (to recap, here we're protecting a superior hash - BLAKE2 - with an inferior one - SHA1 -, which just doesn't make much sense from a pure security POV, even if we assume that it likely won't cause any harm in practice)

DomT4:
There's likely to be a broader decision on this soon, in regards to the future of SHA1 in Homebrew's core. Please wait on that.

vszakats:
Okay 👍

DomT4:
Have :rocket: this in https://github.com/Homebrew/homebrew/commit/f3a06bba5640a2413b81f4df8577b16efb249768. We'll make a fuller decision relatively soon.

vszakats:
Thank you @DomT4, looking forward for the decision.

PR #40644, opened 2015-06-13
Add download strategy for private GitHub files

koenrh:
In our organization we'd love to use `brew` to distribute command-line interface tools internally. However, we use GitHub Releases (downloads are served from `codeload.github.com`) with **private** repositories to distribute the binaries. This doesn't work as `brew` doesn't know how to authenticate with GitHub (at least not for downloads). So, I've added a download strategy that sends the GitHub access token in requests to `codeload.github.com` if the `HOMEBREW_GITHUB_API_TOKEN` environment variable is set.

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -323,6 +323,22 @@ class CurlDownloadStrategy < AbstractFileDownloadStrategy
   end
 end
 
+# GitHubCodeLoadDownloadStrategy downloads tarballs and ZIP archives from
+# GitHub CodeLoad, the place where the downloads for your releases are stored.
+# This download strategy uses a GitHub access token (HOMEBREW_GITHUB_API_TOKEN)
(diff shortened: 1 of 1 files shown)


jacknagel:
I think it would be better for this code to live alongside the formulae that use it, rather than in core (you can see an example of this in the [homebrew-php](https://github.com/Homebrew/homebrew-php) tap, which makes heavy use of custom code for requirements, etc.). You can use the `url "foo", :using => GitHubCodeloadDownloadStrategy` syntax to enable it.

The problem I see with accepting this into core is that we become responsible for maintaining it and acting as gatekeepers for changes to it, even though we don't use it (I regret that we merged the S3 download strategy for this reason).

koenrh:
Thinking it over again, it makes sense indeed to have this code alongside the formulae that use it as this will be in a separate tap anyway. I'm going to close this pull request and move this to our own tap. Thanks for your feedback!

PR #40660, opened 2015-06-13
gobject-introspection 1.45.2

denji:


DomT4:
Aren't odd numbers for GNOME software dev releases? I thought it was stable = even, unstable = odd?

PR #40667, opened 2015-06-13
iso-codes 3.58

denji:


bfontaine:
Thank you for your contribution, @denji, we appreciate it!

PR #40672, opened 2015-06-14
imagemagick 6.9.1-5

denji:
```
2015-06-13  6.9.1-6 Glenn Randers-Pehrson <glennrp@image...>
  * Reverted change to 6.9.1-3 that skipped palette-building.

2015-06-03  6.9.1-5 Cristy  <quetzlzacatenango@image...>
  * Use correct scale when interpretting alpha (e.g. rgba(0,0,0,1)).
  * DrawGetVectorGraphics() now returns proper XML (reference
    http://www.imagemagick.org/discourse-server/viewtopic.php?f=3&t=27751).
  * Support writing EXR files with different color types (reference
    http://www.imagemagick.org/discourse-server/viewtopic.php?f=2&t=27759).
  * Prefer PKG_CHECK_MODULES() when searching for delegate libraries.
  * Throw exception if frame option bevel exceeds to the image width / height.
  * Resolve undefined behaviors (reference
    http://www.imagemagick.org/discourse-server/viewtopic.php?f=3&t=27811).
```

### DIFF
--- a/Library/Formula/imagemagick.rb
+++ b/Library/Formula/imagemagick.rb
@@ -1,9 +1,8 @@
 class Imagemagick < Formula
   desc "Tools and libraries to manipulate images in many formats"
   homepage "http://www.imagemagick.org"
-  url "http://www.imagemagick.org/download/releases/ImageMagick-6.9.1-4.tar.xz"
-  mirror "https://downloads.sourceforge.net/project/imagemagick/6.9.1-sources/ImageMagick-6.9.1-4.tar.xz"
-  sha256 "74973478058069ce44040ff76b61df713d70adcadfd207836cc0622cadf1e4bf"
+  url "http://www.imagemagick.org/download/releases/ImageMagick-6.9.1-5.tar.xz"
+  sha256 "34d59b83a2c3e526a7fad1fed30a950ea7b4ba2e936017a4c71581250c86352d"
 
   head "https://subversion.imagemagick.org/subversion/ImageMagick/trunk",
        :using => :svn


Library/Formula/imagemagick.rb
```diff
@@ -1,9 +1,8 @@
 class Imagemagick < Formula
   desc "Tools and libraries to manipulate images in many formats"
   homepage "http://www.imagemagick.org"
-  url "http://www.imagemagick.org/download/releases/ImageMagick-6.9.1-4.tar.xz"
-  mirror "https://downloads.sourceforge.net/project/imagemagick/6.9.1-sources/ImageMagick-6.9.1-4.tar.xz"
-  sha256 "74973478058069ce44040ff76b61df713d70adcadfd207836cc0622cadf1e4bf"
+  url "http://www.imagemagick.org/download/releases/ImageMagick-6.9.1-5.tar.xz"
```
bfontaine: Is there no mirror for this release?

mikemcquaid: It seems not but we should find one.

chdiza:
When I downloaded the tarball from the homepage, the changelog says it's now at version `6.9.1-6`, not `6.9.1-5`.  Is this a typo?  Or did they very quickly issue another point release?

chdiza:
Answer: There is a typo in their changelog.  `convert --version` yields 6.9.1-5.

Library/Formula/imagemagick.rb
vszakats: Here's one:
https://fossies.org/linux/misc/ImageMagick-6.9.1-5.tar.zx

denji: @vszakats sha256 |24af18437cfccdaacd4be821cec6291397a71f7743129474f9c75b3eada57926|

vszakats: Ah, sorry, I've made a typo, and not only that, the link is wrong anyway :( 
A .gz version is there, but I'm not sure that's much of a help:
https://fossies.org/linux/misc/ImageMagick-6.9.1-5.tar.gz

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/2fc6aaf0bb951c5864d1ab8574c14851572e796c. Thanks @denji!

Thomq:
Looks like it has been bumped 6.9.1.-6 as of June 19–20th.  Doesn't update because -5 isn't there anymore.

PR #40679, opened 2015-06-14
Added formula for antigen

dimitriye98:


tseemann:
The ```head URL``` doesn't look correct?

dimitriye98:
Oops, fail, fixing now. I thought I'd save time by modifying the url instead of copy pasting again and made a stupid mistake in doing so.

Library/Formula/antigen.rb
```diff
@@ -0,0 +1,22 @@
...
+  test do
+    (testpath/".zshrc").write "source antigen.zsh"
+    system "/bin/zsh", "--login", "-i", "-c", "antigen help"
+  end
+
+  def caveats; <<-EOS.undent
+    To use antigen, add the following line to your ~/.zshrc:
+    source antigen.zsh
+    EOS
+  end
```
mikemcquaid: Remove this unless it's Homebrew-specific, thanks!

dimitriye98: Well, it's a standard part of the installation, but it's a part which homebrew can't manage; it has to be done manually, so I assumed noting that to the user would be important, but ok.

Library/Formula/antigen.rb
```diff
@@ -0,0 +1,16 @@
+class Antigen < Formula
+  desc "A plugin manager for zsh, inspired by oh-my-zsh and vundle."
+  homepage "http://antigen.sharats.me/"
+  url "https://github.com/zsh-users/antigen/archive/v1.tar.gz"
+  sha256 "6d4bd7b5d7bc3e36a23ac8feb93073b06e1e09b9100eb898f66c2e8c3f4d7847"
+  head "https://github.com/zsh-users/antigen.git"
+
+  def install
+    lib.install "antigen.zsh"
```
mikemcquaid: Install this to `share` or `libexec`: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/27900/version=yosemite/testReport/junit/brew-test-bot/yosemite/audit_antigen___strict/

dimitriye98:
I tried putting it in `libexec` but brew doesn't seem to link it if I do.

mikemcquaid:

For future reference the preferred commit message format for new formulae is `antigen 1 (new formula)`.

apjanke:
This formula is installing the file directly under `#share` at `/usr/local/share/antigen.zsh`; normally it should go under `#pkgshare`, e.g. `/usr/local/share/antigen/antigen.zsh`. Would make a difference if it ever grew to be multiple files.

This would be a breaking change to make now though, since existing installations point directly at the file to load antigen from their `~/.zshrc` file.

mikemcquaid:
@apjanke Breaking change seems fine here but `revision` makes sense, I agree.

PR #40681, opened 2015-06-14
ems-flasher 0.03 (new formula)

parashep:


Library/Formula/ems-flasher.rb
```diff
@@ -0,0 +1,19 @@
...
+  depends_on "pkg-config" => :build
+  depends_on "libusb"
+
+  def install
+    system "make"
+    bin.install "ems-flasher"
+  end
+
+  test do
+    system "#{bin}/ems-flasher", "--version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/ems-flasher.rb
```diff
@@ -0,0 +1,19 @@
...
+  url "https://lacklustre.net/projects/ems-flasher/ems-flasher-0.03.tgz"
+  sha256 "d77723a3956e00a9b8af9a3545ed2c55cd2653d65137e91b38523f7805316786"
+  head "git://lacklustre.net/ems-flasher"
+
+  depends_on "pkg-config" => :build
+  depends_on "libusb"
+
+  def install
+    system "make"
+    bin.install "ems-flasher"
```
mikemcquaid: Is there a `make install` we can use (even if it needs patched)?

mikemcquaid:
This may be a better fit for the `homebrew-games` tap but let's work through review on this PR first. Thanks for submitting!

Library/Formula/ems-flasher.rb
parashep: Unfortunately not, the test would fail given anything else to do because the software relies on an external device to be plugged in.

Library/Formula/ems-flasher.rb
parashep: There is, but I wasn't sure if a patch or replicating the behaviour would have been better for Homebrew. If a patch is a better choice, then I will gladly change it to do one. I forgot the man file.

parashep:
If this software isn't a good fit for `homebrew` in *general*, I would definitely understand.

mikemcquaid:
@parashep Oh, I think we'd definitely want it in either here or that tap.

Library/Formula/ems-flasher.rb
mikemcquaid: Yeh, ideally in these cases we'll patch the upstream `Makefile` with `inreplace` or a `patch`, submit it upstream and try and get it included for future releases.

parashep: There is no `configure` included with the software, and the stable version doesn't even have a `make install` target or manfile (but the HEAD version has both). Does this mean it still would be *better* for me to write separate patches for both of them, and use the patched `make install` targets, instead of using `bin.install` and `man1.install`?

Library/Formula/ems-flasher.rb
mikemcquaid: No worries, thanks!

Library/Formula/ems-flasher.rb
mikemcquaid: Nah, it's good as-is, thanks!

mikemcquaid:
@parashep Would you mind submitting this as-is to the `homebrew-games` tap? Mention my username (I don't watch it) and I'll merge it immediately. Thanks!

PR #40682, opened 2015-06-14
uses: rescue TapFormulaAmbiguityError caused by Dependency#to_formula

### LINKED ISSUE #40636: changes to tap behavior causes problems for `brew uses`
Possibly a byproduct of #39524.  I assume this is unintended behavior:

1. `brew tap staticfloat/juliadeps` (just to create a lot of conflicts)
2. `brew uses when`

There shouldn't be any output, but instead we get a bunch of tap ambiguity errors that are outside the user's control: https://gist.github.com/anonymous/fbe00f8044d0f2f33e69

xu-cheng:
Fixes #40636.

### DIFF
--- a/Library/Homebrew/cmd/cleanup.rb
+++ b/Library/Homebrew/cmd/cleanup.rb
@@ -124,9 +124,11 @@ module Homebrew
       true
     elsif formula.opt_prefix.directory?
       # SHA records were added to INSTALL_RECEIPTS the same day as opt symlinks
-      Formula.installed.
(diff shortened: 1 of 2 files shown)


mistydemeo:
Hm, is it worth catching just the specific exception instead of all exceptions?

xu-cheng:
I think it's fine to catch the exception in general. In this context, we care about whether a formula is depended on other formula. We don't care whether this formula has some error.

mikemcquaid:
Makes sense :+1:

PR #40702, opened 2015-06-14
gnu-parallel 20150522

denji:
HEAD repository


DomT4:
Merged! Thanks again @denji!

Issue #35758, opened 2015-01-11
ChaCha20/Poly1305 support in Crypto Libraries


azet:
Hi Guys,

This is just a suggestion I'd like to discuss before opening any kind of PR to see if you're interested in it at all.

I've written to @DomT4 about this on twitter earlier. From talking to very popular CDNs (can't name names unfortunately, but it's easy to guess) as well as Google people; they all compile OpenSSL with the prominent ChaCha20/Poly1305 patch and support it for users accessing customer sites. For NSS (used by Firefox and other software for SSL/TLS) there is also a similar patch. This might be worth thinking about. 

Let me explain in more detail:
One of the reasons Google initially started supporting ChaCha20 (the patch is also written and maintained by them) is that AES-GCM might cause timing problems on mobile devices or devices that do not support the Intel AES-NI instructions. This mainly affects ARM, but might affect other architectures as well. Google seems to have had inhouse research that there is indeed a serious issue. They're not public on details as far as I know, which concerns me a bit. ChaCha20 does not have this issue since it's a constant-time cipher by design. The patch applies to TLS 1.2 and hence adds another AEAD (authenticated encryption with associated data) cipher/mode.

From an IETF perspective: this cipher is not in any standard yet - and certainly will not be added to TLS 1.2 - but is going to be officially supported by TLS 1.3 (I can link to all relevant IETF documents and discussion if really necessary, but this will take me about an hour). 

Support for this is in different Google software, for example Chrome/Chromium:
- https://www.imperialviolet.org/2013/10/07/chacha20.html
- https://code.google.com/p/chromium/issues/detail?id=310768
- https://codereview.chromium.org/59083010/

Android phones make extensive use of this patch for example. Now; modern Macbooks support AES-NI instructions. But as there is a project called `linuxbrew` this might be of interest. It also might be of interest to developers using and linking to OpenSSL/NSS from Homebrew directly. Or Mac users with pre-Westmere Intel processors (i.e. pre-2010 Macbooks).

My suggestion would not be to turn this on by default but make it an option to enable if a user or developer wants support for it. I'd certainly like that for development, for now I custom compile all of these things by hand.

- OpenSSL patch: https://git.openssl.org/gitweb/?p=openssl.git;a=commitdiff;h=9a8646510b
- NSS patch: https://bugzilla.mozilla.org/show_bug.cgi?id=917571

Thanks for your consideration and time,
Aaron

DomT4:
I'll give this a full reply later today.

[...]

PR #40704, opened 2015-06-14
Support override openssl

denji:
If we clearly do not want to use, we can try to use universal `:libssl`, provides the ability to override libressl instead of installing openssl `depends_on :libssl`.
- Homebrew/homebrew#35758

We can also do `unlink libressl` `link openssl` when it is clearly needed.

### DIFF
--- a/Library/Homebrew/dependency_collector.rb
+++ b/Library/Homebrew/dependency_collector.rb
@@ -105,6 +105,7 @@ class DependencyCollector
     when :x11        then X11Dependency.new(spec.to_s, tags)
     when :xcode      then XcodeDependency.new(tags)
     when :macos      then MinimumMacOSRequirement.new(tags)
+    when :libssl     then LibsslDependency.new(tags)
     when :mysql      then MysqlDependency.new(tags)
     when :postgresql then PostgresqlDependency.new(tags)
     when :gpg        then GPGDependency.new(tags)
--- a/Library/Homebrew/requirements.rb
+++ b/Library/Homebrew/requirements.rb
@@ -52,6 +52,13 @@ class XcodeDependency < Requirement
   end
 end
 
+class LibsslDependency < Requirement
+  fatal true
+  default_formula 'openssl'
+
+  satisfy { which 'openssl' }
+end
+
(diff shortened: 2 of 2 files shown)


DomT4:
This won't work as-is.

If we replace all of the `openssl` and `libressl` dependencies with `:libssl` as in this PR from-bottle compiles will be linked to Homebrew's OpenSSL. From-source installs will link against the system OpenSSL unless Homebrew's OpenSSL is *already* installed. That's a regression on our current behaviour.

Even if LibreSSL is at the front of the `$PATH` because a user has linked it into `/usr/local` (Which I don't think we should encourage), It'll be compiled against Homebrew's OpenSSL when that is installed because `keg_only` takes precedence over everything else in the `$PATH`.

To manipulate this you would probably have to do something like [this](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/requirements/apr_dependency.rb#L9-L15) but given both `libressl` and `openssl` are `keg_only` you'd have to work out a consistent way to make the right one appear in front of the other in the `$PATH`.

I think more broadly the idea of allowing users to *always* override OpenSSL with LibreSSL is still premature at this point in time. There are still things that can't be built against LibreSSL, or do strange things when linked against LibreSSL and it's fair to argue that LibreSSL remains a young platform. I also suspect that we would get the bug reports about the failures rather than upstream.

For now, my strong preference remains to handle this on a formula-by-formula basis.

jacknagel:
Yeah, don't think we're going down this road.

denji:
Here's the problem, you generate the invoice code inside the homebrew formulas, instead of solving problems
https://github.com/Homebrew/homebrew/blob/master/Library/Formula/curl.rb#L50-L57
```ruby
    # Throw an error if someone actually tries to rock both SSL choices.
    # Long-term, make this singular-ssl-option-only a requirement.
    if build.with?("libressl") && build.with?("openssl")
      ohai <<-EOS.undent
      --with-openssl and --with-libressl are both specified and
      curl can only use one at a time; proceeding with libressl.
      EOS
    end
```

PR #40705, opened 2015-06-14
Update repo.rb with version '1.21'

ChaseVoid:
updated URL and SHA1 to correct version.

### DIFF
--- a/Library/Formula/repo.rb
+++ b/Library/Formula/repo.rb
@@ -5,6 +5,7 @@ class Repo < Formula
   homepage 'http://source.android.com/source/git-repo.html'
   url 'https://raw.githubusercontent.com/android/tools_repo/v1.12.21/repo'
   version '1.21'
+  revision '1'
   sha1 '9421408dba2d80e3c73910567f70f2e0c91dab2f'
 
   def install
--- a/Library/Formula/repo.rb
+++ b/Library/Formula/repo.rb
@@ -5,7 +5,7 @@ class Repo < Formula
   homepage 'http://source.android.com/source/git-repo.html'
   url 'https://raw.githubusercontent.com/android/tools_repo/v1.12.21/repo'
   version '1.21'
-  revision '1'
+  revision 1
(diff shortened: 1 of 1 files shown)


Library/Formula/repo.rb
```diff
@@ -3,9 +3,9 @@
 class Repo < Formula
   desc "Repository tool for Android development"
   homepage 'http://source.android.com/source/git-repo.html'
-  url 'https://raw.githubusercontent.com/android/tools_repo/v1.12.13/repo'
+  url 'https://raw.githubusercontent.com/android/tools_repo/v1.12.21/repo'
   version '1.21'
```
DomT4: Presumably this needs bumping as well?

ChaseVoid: The stable version currently is 1.21 but the url was pointing to incorrect
version of 1.13

DomT4: I see. Maybe it's worth adding `revision 1` to the formula so people who think they're currently using version `1.21` actually are?

ChaseVoid: Will this help?

DomT4: `1` doesn't need to go inside `""`. Should just be:

```ruby
sha1 "9421408dba2d80e3c73910567f70f2e0c91dab2f"
revision 1
```

Thanks!

ChaseVoid:
Thanks. I've updated it.

Library/Formula/repo.rb
ChaseVoid: Thanks, I've updated the pull request.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/c5dcbb8d621e12bc12f67f41eea5c6b3c0a0d42c. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Small nit for future: Commit messages for non-version bumps should be styled `repo: use correct url for version` or similar, Thanks!

ChaseVoid:
Will keep that in mind. Thanks :)

PR #40711, opened 2015-06-14
Update unison SHA1 checksum

matthewlmcclure:
Addresses a remaining issue relating to #35795.

PR #40712, opened 2015-06-14
Add fsmonitor.py to files installed with unison

matthewlmcclure:
This makes `unison -repeat watch` work instead of failing with `Fatal error: No file monitoring helper program found`.

### DIFF
--- a/Library/Formula/unison.rb
+++ b/Library/Formula/unison.rb
@@ -22,5 +22,6 @@ class Unison < Formula
     system "make ./mkProjectInfo"
     system "make UISTYLE=text"
     bin.install 'unison'
+    bin.install 'fsmonitor.py'
   end
 end


DomT4:
If you merge this into #40711 you won't need to rebase it afterwards to get it to pass the test-bot.

mikemcquaid:
Does this need installed in `bin`? The name is kinda weird to be installed globally.

matthewlmcclure:
Installing to `bin` seems to be the intent, based on the following:

>
Changes in Version 2.48.3
Changes since 2.45:
...
Small fix to the way fsmonitor.py gets invoked when using the file watching functionality, suggested by Josh Berdine. Unison will now look for fsmonitor.py in the same directory where the Unison executable itself lives.
>

mikemcquaid:
This doesn't fit with how we package Homebrew, unfortunately. Generally our solution for this situation is to install to `libexec` as the prefix and write wrapper scripts for individual binaries into `bin` e.g. like `ansible.rb`

matthewlmcclure:
I assume that the pattern I should follow is the one documented at https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#specifying-gems-python-modules-go-projects-etc-as-dependencies

mikemcquaid:
@matthewlmcclure Yes, more-or-less. `typesafe-activator.rb` may be a simpler example.

DomT4:
Ping? Mile's comments need addressing, and this needs rebasing on master, Thanks!

DomT4:
Er, Mike, not Mile.

matthewlmcclure:
Not sure when I will have time to revisit this. If you'd prefer to close
the issue until someone can give it more time, I understand.

On Mon, Aug 3, 2015 at 9:55 PM, Dominyk Tiller <notifications@github.com>
wrote:

> Er, Mike, not Mile.
>
> —
> <https://github.com/Homebrew/homebrew/pull/40712#issuecomment-127453379>.
>



-- 
Matt McClure
http://matthewlmcclure.com
http://cens.io

DomT4:
Sure. If you want to resubmit it at a future point in time we're happy to look at it then, Thanks for the submission!

PR #40715, opened 2015-06-14
pgloader: use https for resources

vszakats:


bfontaine:
Thank you @vszakats, as always :wink:

vszakats:
Thanks a lot @bfontaine!

PR #40717, opened 2015-06-14
cmd/search --desc: bold formula names

dunn:
More readable output, IMO.

### DIFF
--- a/Library/Homebrew/cmd/search.rb
+++ b/Library/Homebrew/cmd/search.rb
@@ -26,7 +26,7 @@ module Homebrew
       rx = query_regexp(query)
       Formula.each do |formula|
         if formula.desc =~ rx
-          puts "#{formula.full_name}: #{formula.desc}"
+          puts "#{Tty.white}#{formula.full_name}:#{Tty.reset} #{formula.desc}"
         end
       end
     elsif ARGV.empty?


mikemcquaid:
Makes sense to me but I wonder about the inconsistency with other things? Maybe that doesn't matter.

dunn:
Oh, where are the inconsistencies?

mikemcquaid:
Just most commands not using bold in the output. I dunno. @Homebrew/owners any thoughts?

DomT4:
I have no particular objections. On a themed terminal/shell it's hard to tell the difference, but that's not something we can really control.

dunn:
An alternative would be to align the descriptions:

```
🐙  brew search --desc ftp
curl            Get a file from an HTTP, HTTPS or FTP server
git-ftp         Git-powered FTP client
ncftp           FTP client with an advanced user interface
proftpd         Highly configurable GPL-licensed FTP server software
pure-ftpd       Secure and efficient FTP server
squid           Advanced proxy caching server for HTTP, HTTPS, FTP, and Gopher
uberftp         Interactive GridFTP client
vsftpd          Secure FTP server for UNIX
wput            Tiny, wget-like FTP client for uploading files
yafc            Command-line FTP client
```

Then we wouldn't need special formatting and the output would be easier to pipe into `awk` or whatever.

chdiza:
I like the alignment much better than the bold idea.

mikemcquaid:
Yeh, alignment looks great :+1:

dunn:
I like it too, but I'll need to think how to do it right.  My first try makes searches feel a lot slower, because it has to run the whole search before displaying any output:
```ruby
    elsif ARGV.include? '--desc'
      query = ARGV.next
      matches = []
      Formula.each do |formula|
        if formula.desc =~ query_regexp(query)
          matches << formula
        end
      end
      longest = matches.max_by{|s| s.full_name.length}
      matches.each do |f|
        padding = " " * (longest.full_name.length - f.full_name.length)
        padding += "\t"
        puts "#{f.full_name}#{padding}#{f.desc}"
      end
```
Instead it should calculate the padding before running the search, so results can still be displayed as they're found.

dunn:
I'm not sure how to accomplish the alignment without that perceived performance hit.  Should I close this, or is bold better than nothing?

xu-cheng:
Maybe just make a reason assumption on the longest formula full name length.

mikemcquaid:
> Maybe just make a reasonable assumption on the longest formula full name length.

:+1:

dunn:
The problem is that the longest formula names are very long.  

My only idea right now is to just pick a fixed length and add a linebreak for those formula with long names:
```
🐙  brew search --desc tmux
clipper        Share OS X clipboard with tmux and other local and remote apps
rainbarf       CPU/RAM/battery stats chart bar for tmux (and GNU screen)
reattach-to-user-namespace
               Reattach process (e.g., tmux) to background
tmux-cssh      TMUX with a "ClusterSSH"-like behavior
tmux-mem-cpu-load
               CPU, RAM memory, and load monitor for use with tmux
wemux          Enhances tmux's to provide multiuser terminal multiplexing
```

bfontaine:
The problem with line breaks is that you can’t use traditional line-based command-line tools anymore. I don’t have any other idea, thought.

chdiza:
What about just colorizing the formula names, in green or blue?

DomT4:
The problem with colourisation is that it can clash pretty nastily with someone's existing Terminal theme.

chdiza:
> The problem with colourisation is that it can clash pretty nastily with someone's existing Terminal theme.

True: when "warnings" got turned into yellow recently, they are VERY hard to see when one has a whitish background to one's terminal.  I have considered opening an issue over it.

OTOH there's color all over e.g. `brew info`.

chdiza:
I found that piping the output through `awk -F ":" '{print $1" ---"$2}'` (for illustration's sake) is more readable that the current output.  Perhaps we just need a wider separator than ":".  It's not as nice as alignment, but there would be no performance hit and it's more readable than how it is now.

denji:
--desc and alias (-d)

mikemcquaid:
Auch, let's just go for bold and I'll merge this in.

dunn:
Done, thanks for bearing with me on this.

mikemcquaid:
Thanks @dunn!

PR #40721, opened 2015-06-15
Preliminary PR for reviewing removal of CLT/Xcode dependencies for bottles

woodruffw:
Just some notes:

* The checks `check_for_installed_developer_tools` and `check_for_bad_install_name_tool` were removed outright, as they are currently incompatible with the way my system pulls in the `cctools` formula. Once I figure out a clean way for these checks to be satisfied by that formula, I can add them back.
* There are currently a few hardcoded checks in the core against a formula's name being `cctools`, which is definitely not good form. These can also be removed with the addition of a few helper functions I'll be adding.

Please tell me what you think!

### DIFF
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -36,14 +36,12 @@ module OS
       end
     end
 
-    def locate_cctool(tool)
-      if File.executable?(path = "#{HOMEBREW_LIBRARY}/Homebrew/vendor/cctools/#{tool}")
-        Pathname.new(path)
-      elsif File.executable?(path = "#{Formula['cctools'].opt_prefix}/bin/#{tool}")
-        Pathname.new(path)
-      else
-        locate(tool)
-      end
+    def install_name_tool
+      Pathname.new("#{HOMEBREW_LIBRARY}/Homebrew/vendor/cctools/install_name_tool")
+    end
+
+    def otool
+      Pathname.new("#{HOMEBREW_LIBRARY}/Homebrew/vendor/cctools/otool")
     end
 
     def active_developer_dir
--- a/Library/Aliases/atlassian-cli
+++ b/Library/Aliases/atlassian-cli
@@ -1,0 +1,1 @@
+../Formula/stash-cli.rb
--- a/Library/Formula/algernon.rb
+++ b/Library/Formula/algernon.rb
@@ -3,8 +3,8 @@ require "language/go"
 class Algernon < Formula
   desc "HTTP/2 web server with Lua support"
   homepage "http://algernon.roboticoverlords.org/"
-  url "https://github.com/xyproto/algernon/archive/0.74.tar.gz"
-  sha256 "1341af6864643a968d85bfa63ca231604b6d1123919c6826ae179908c6c4a176"
+  url "https://github.com/xyproto/algernon/archive/0.82.tar.gz"
+  sha256 "b64a6367445b707ae272ff917d971133777f62893f75a3dbaa0a19018d02a51c"
   head "https://github.com/xyproto/algernon.git"
 
   bottle do
@@ -20,6 +20,7 @@ class Algernon < Formula
   # List of Go dependencies and hashes.
   # Generated using: https://github.com/samertm/homebrew-go-resources
   %w[
+    github.com/bkaradzic/go-lz4 4f7c2045dbd17b802370e2e6022200468abf02ba
     github.com/bobappleyard/readline a1f23ef6d7c8177f1fe3085d3bf7b40457049f0c
     github.com/boltdb/bolt 5eb31d5821750fbc84284d610c0ce85261662adb
     github.com/bradfitz/http2 f8202bc903bda493ebba4aa54922d78430c2c42f
@@ -33,14 +34,17 @@ class Algernon < Formula
     github.com/juju/ratelimit faa59ce93750e747b2997635e8b7daf30024b1ac
     github.com/mamaar/risotto 2683127f39af835e766a70b203efc6a51dd2ebe6
     github.com/mattn/go-runewidth 5890272cd41c5103531cd7b79e428d99c9e97f76
-    github.com/natefinch/pie 704bec149c2c6e52bcb2570c2d8c5e30e0ddc732
+    github.com/mitchellh/go-homedir 1f6da4a72e57d4e7edd4a7295a585e0a3999a2d4
+    github.com/mitchellh/mapstructure 2caf8efc93669b6c43e0441cdc6aed17546c96f3
+    github.com/natefinch/pie 0e26844fad24d1f3fc81b7d6a8f8589cac6ec64c
     github.com/nsf/termbox-go 598cdb8b3c49430c1c91604805461732064bde1d
     github.com/russross/blackfriday 386ef80f18233ea97960e855a54382ec446c6637
     github.com/shiena/ansicolor 8368d3b31cf6f2c2464c7a91675342c9a0ac6658
     github.com/shurcooL/sanitized_anchor_name 11a20b799bf22a02808c862eb6ca09f7fb38f84a
-    github.com/sirupsen/logrus 386ccca031649304b1b3e6db057e8cecdaabe760
-    github.com/tylerb/graceful ff20838629c6b3e602864634a75949b3685d584a
+    github.com/sirupsen/logrus 0f2a4955b11372eec5a4b95d907eaf8379b835d0
+    github.com/tylerb/graceful ac9ebe4f1ee151ac1eeeaef32957085cba64d508
     github.com/xyproto/cookie b84c85ae2aa3e21b2c7fc8c37d5a3081c0c9c83b
+    github.com/xyproto/jpath 7b9116746a7134fdcefd5802af406361087bb190
     github.com/xyproto/mime 58d5c367ee5b5e10f4662848579b8ccd759b280e
     github.com/xyproto/permissionbolt b94c45a7f4c70603f12aa1d47a325a37036fd3ba
     github.com/xyproto/permissions2 b5248dea92e87c670813edf4e68dc16eff59c05b
@@ -53,6 +57,7 @@ class Algernon < Formula
     github.com/xyproto/term c9eabb15c0681f48654ee132a8bc6608c7bed8b3
     github.com/xyproto/unzip 823950573952ff86553b26381fe7472549873cb4
     github.com/yosssi/gcss 39677598ea4f3ec1da5568173b4d43611f307edb
+    github.com/yuin/gluamapper d836955830e75240d46ce9f0e6d148d94f2e1d3a
     github.com/yuin/gopher-lua 54750d6aec1c2706990e9ec17438bd88d30231ab
   ].each_slice(2) do |resurl, rev|
     go_resource resurl do
@@ -62,12 +67,12 @@ class Algernon < Formula
 
   go_resource "golang.org/x/crypto" do
     url "https://go.googlesource.com/crypto.git",
-      :revision => "1e856cbfdf9bc25eefca75f83f25d55e35ae72e0"
+      :revision => "4ed45ec682102c643324fae5dff8dab085b6c300"
   end
 
   go_resource "golang.org/x/net" do
     url "https://go.googlesource.com/net.git",
-      :revision => "dfcbca9c45aeabb8971affa4f76b2d40f6f72328"
+      :revision => "dfe268fd2bb5c793f4c083803609fce9806c6f80"
   end
 
   def install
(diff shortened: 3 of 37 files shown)


woodruffw:
@mistydemeo - sorry - forgot to ping you in the PR description!

Library/Formula/cctools.rb
```diff
@@ -11,6 +11,7 @@ class Cctools < Formula
   end
 
   bottle do
+    does_not_need_relocation true
```
mistydemeo: Can nix the bool argument here, I think - simply `does_not_need_relocation` is clear enough on its own.

Library/Homebrew/cmd/doctor.rb
```diff
@@ -1163,12 +1163,19 @@ def check_for_unlinked_but_not_keg_only
 end
 
   def check_xcode_license_approved
-    # If the user installs Xcode-only, they have to approve the
-    # license or no "xc*" tool will work.
-    if `/usr/bin/xcrun clang 2>&1` =~ /license/ and not $?.success? then <<-EOS.undent
-      You have not agreed to the Xcode license.
-      Builds will fail! Agree to the license by opening Xcode.app or running:
-          sudo xcodebuild -license
+    if MacOS::Xcode.installed?
```
mistydemeo: Rather than branching inside the check body, it might be better to categorize a set of Xcode checks that can be replaced with this warning in the case that Xcode isn't installed.

Library/Homebrew/os/mac.rb
```diff
@@ -24,7 +24,9 @@ def locate tool
       # Give the name of the binary you look for as a string to this method
       # in order to get the full path back as a Pathname.
       (@locate ||= {}).fetch(tool) do |key|
-        @locate[key] = if File.executable?(path = "/usr/bin/#{tool}")
+        @locate[key] = if File.executable?(path = "/usr/local/opt/cctools/bin/#{tool}")
```
mistydemeo: Want to grab the path from the formula object here.

Library/Homebrew/requirements/cctools_requirement.rb
```diff
@@ -0,0 +1,8 @@
+class CctoolsRequirement < Requirement
+  fatal true
+  default_formula 'cctools'
+
+  satisfy do
+    MacOS::Xcode.installed? || MacOS::CLT.installed? || Dir.exists?('/usr/local/opt/cctools')
```
mistydemeo: You can leave the last part off - should be fine letting it proceed to the normal mechanism for determining the formula is installed.

Library/Homebrew/software_spec.rb
```diff
@@ -260,21 +264,25 @@ class BottleSpecification
...
     @collector = BottleCollector.new
+    @does_not_need_relocation = false
   end
 
   def compatible_cellar?
     cellar == :any || cellar == HOMEBREW_CELLAR.to_s
   end
 
+  def needs_relocation?
+    @does_not_need_relocation == false
```
mistydemeo: `!@does_not_need_relocation`

Library/Homebrew/os/mac.rb
```diff
@@ -24,7 +24,9 @@ def locate tool
       # Give the name of the binary you look for as a string to this method
       # in order to get the full path back as a Pathname.
       (@locate ||= {}).fetch(tool) do |key|
-        @locate[key] = if File.executable?(path = "/usr/bin/#{tool}")
+        @locate[key] = if File.executable?(path = "#{Formula['cctools'].opt_prefix}/bin/#{tool}")
```
mistydemeo: I feel like we probably want to make the case more complex here: use the Xcode tools if present, and only fall back to a Homebrew cctools otherwise. We may want to push some logic into the places that consume this method, rather than putting it into `MacOS.locate`.

Library/Formula/cctools.rb
```diff
@@ -11,6 +11,7 @@ class Cctools < Formula
   end
 
   bottle do
+    does_not_need_relocation
     cellar :any
```
xu-cheng: It would be nice if we can combine these two annotations as one like `cellar :skip_relocation` or something.

Library/Homebrew/requirements/cctools_requirement.rb
```diff
@@ -0,0 +1,8 @@
+class CctoolsRequirement < Requirement
+  fatal true
+  default_formula 'cctools'
+
+  satisfy do
+    MacOS::Xcode.installed? || MacOS::CLT.installed?
```
xu-cheng: Shouldn't this be `MacOS::Xcode.installed? || MacOS::CLT.installed? || Formula["cctools"].installed?`

Library/Homebrew/os/mac.rb
woodruffw: Can do! This particular behavior is really only needed when we attempt to call `otool` and `install_name_tool` on unarchived bottles, so I could write a wrapper function like `MacOS.locate_cctools` or simply include this logic in each place.

Library/Homebrew/requirements/cctools_requirement.rb
mistydemeo: Since `cctools` is already the default formula, I don't think we need an explicit check for it in the `satisfy` block.

Library/Formula/cctools.rb
woodruffw: I agree, that would be much nicer than two separate lines in the DSL. I'll see about changing this style :smile:

Library/Homebrew/requirements/cctools_requirement.rb
xu-cheng: Thanks for explanation @mistydemeo

Library/Formula/cctools.rb
xu-cheng: You also need to update [this line](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/bottle.rb#L197) to make an exception for `cctools`.

mikemcquaid: Or maybe invert it and have `cellar :any` and `cellar :relocatable` (the latter being the current behaviour, the prior not needing `install_name_tool`?

Library/Homebrew/cmd/install.rb
```diff
@@ -142,7 +145,12 @@ def check_cellar
   def perform_preinstall_checks
     check_ppc
     check_writable_install_location
-    check_xcode
+    if MacOS::Xcode.installed?
+      check_xcode
+    else
+      opoo "You have not installed Xcode."
+      puts "Bottles may install correctly, but builds will fail!"
```
mikemcquaid: Not all builds will fail. It'd be good to decide how to fail here rather than saying "this may fail" and then allowing it to happen?

Library/Homebrew/requirements/cctools_requirement.rb
mikemcquaid: I think it's probably good practice to add it in case this behaviour changes in future.

Library/Formula/cctools.rb
mikemcquaid: Also: this should be autodetected in `bottle.rb` or it'll be overwritten any upgrade/rebottling

mikemcquaid:
This is a great start, nice work @woodruffw!

I think it'd be good to work out another way of installing `install_name_tool`, though. I think having a formula that's listed by e.g. `brew list` and effectively a build-time dependency of every bottle is going to be a little confusing. Are you still planning on implementing a version of `install_name_tool` in pure Ruby? If not, perhaps we should just vendor a universal `install_name_tool` binary somewhere in Homebrew's Git repository instead.

Library/Formula/cctools.rb
xu-cheng: > Or maybe invert it and have cellar :any and cellar :relocatable (the latter being the current behaviour, the prior not needing install_name_tool?

We cannot do this if we want to keep forward compatibility on old bottle files.

mikemcquaid: True, good point. Maybe `cellar :any_without_relocation`

woodruffw:
Thanks @mikemcquaid!

I would still like to do a pure Ruby implementation of `install_name_tool` and `otool`, but I've been ushered away from that idea partially out of concerns of maintainability/increased complexity compared to just pulling in `cctools` and pointing commands that need it at it.
I think keeping a universal build of the cctools somewhere in the repository could work, but it might mean dropping Leopard support entirely due to changes in the mach-o standard in 10.6 onwards (probably not too big of a problem, just worth mentioning).

I'd appreciate any recommendations that you or @mistydemeo could give me on this matter - I agree that this method isn't particularly clear to the user, but I do think it's fairly simple in terms of maintenance and implementation. If you think the pure-ruby/binary in repo approaches are better in the long term, I'm happy to change directions :smiley:.

Library/Homebrew/requirements/cctools_requirement.rb
woodruffw: Will add!

mistydemeo:
>  I think keeping a universal build of the cctools somewhere in the repository could work, but it might mean dropping Leopard support entirely due to changes in the mach-o standard in 10.6 onwards (probably not too big of a problem, just worth mentioning).

Homebrew only supports 10.6 forward now, so this is fine; I'll do something else for Tigerbrew.

Library/Homebrew/cmd/install.rb
woodruffw: Alright! Would checking the formula's dependency tree for 100% bottled-ness and failing if not be a good solution?

woodruffw:
Great, then it's just a matter of creating universal binaries for 10.6+ and including them somewhere in the repo. I think `Library/Tools` would be a reasonable location, if you want me to shift over to this approach.

mistydemeo:
How about `Library/Homebrew/vendor`? This is where I place a vendored ppc sha256 binary in Tigerbrew.

woodruffw:
Sounds good! So, should I go ahead with this new approach for now? I should be able to have a working version by tonight, it would just be a matter of changing the paths and building the universal binaries.

mistydemeo:
Yes, please!

If you don't have access to a 10.6 machine and need me to build binaries for you, let me know.

woodruffw:
Yeah, I might need you to build the universal binaries. Alternatively, I could structure it like `/Library/Homebrew/vendor/<OS X version>/`, but that would mean adding quite a few more unique files and probably wouldn't be good from a git standpoint.

mistydemeo:
Built a full cctools universal bottle and emailed it to you - you can pull the individual files you need out of it.

> Alternatively, I could structure it like `/Library/Homebrew/vendor/<OS X version>/`

We can look at that if we ever package more than one version of cctools, but as long as it's the same version of cctools for every OS, I think we can avoid that.

woodruffw:
Received, thanks!

mistydemeo:
Do we need all of cctools? Maybe we can limit this to just `otool` and `install_name_tool`.

mikemcquaid:
> Do we need all of cctools? Maybe we can limit this to just otool and install_name_tool.

This seems like a good idea. I like the single binary installed into a shared location idea too.

woodruffw:
I don't think we need all of cctools, but it might be a good idea to keep `strings` and `strip` as well. `strings` is used in cmd/bottle.rb and both have potential forward uses (possibly size reduction in bottled binaries).

mikemcquaid:
I'd rather we keep it to the bare minimum, to be honest. Binary resources in version control are fairly painful so we want to make sure we keep them minimal.

woodruffw:
Alright! I'll remove them now, then.

mistydemeo:
Just checking in - what's still left to do at this point?

woodruffw:
Well, with this system in place, there are a few more things I can/need to do:

* If we're still entertaining the idea of pure Ruby implementations of `otool` and `install_name_tool`, the fact that we've vendored them in the repo itself makes the job of transitioning much easier. I can just write Ruby utilities that behave identically to their cctool equivalents and swap them.
* There are still quite a few parts of Homebrew that run checks against various Xcode/CLT facilities, which need to be branched out so that installing bottles without the CLT doesn't cause those annoying popups or constantly remind you that Xcode is missing.
* I've tested this approach on a few 100% bottled formuae so far (`git`, `bash`, `htop`, etc), but I haven't really pushed it and tried the bottled formulae environment as a whole.
* I still need to clean up `SoftwareSpec` and `BottleSpec`, possibly making those changes suggested by @xu-cheng and @mikemcquaid w/r/t the `cellar` component.

mistydemeo:
Sounds good! At this point I'd like to prioritize the order of those remaining features, and set out what you think your roadmap is for completing them.

At some point it would be good if you could get the chance to rebase this on master and clean up history a bit - then the bots will be able to start testing it.

mikemcquaid:
> If we're still entertaining the idea of pure Ruby implementations of otool and install_name_tool, the fact that we've vendored them in the repo itself makes the job of transitioning much easier. I can just write Ruby utilities that behave identically to their cctool equivalents and swap them.

Seems like a really good idea.

woodruffw:
Great! I'll rebase tonight and start on the Ruby tool implementations :smile:

mistydemeo:
I think the Ruby tools are a really good idea, but I'd like to focus on getting everything else to a working state before we begin implementing them. Since they'd be drop-in replacements for what's in there now, we can work on doing that once everything else Just Works(tm).

woodruffw:
That makes sense. For now, I guess that means changing how `cmd/install.rb` does preinstallation checks and making sure that Xcode/CLT tests are branched out when doing bottle-only installs.

On a similar note, I made it so that `MacOS.locate_cctool` is what's being used to find the vendored tools. Looking back at this I realize it might be better to just have `MacOS.otool` and `MacOS.install_name_tool` for the sake of readability. Any thoughts on that?

mistydemeo:
> Any thoughts on that?

That sounds better to me, yeah - there are only two tools, and that's clearer.

woodruffw:
That's what I get for not paying attention to the branch I'm on. I'll close this PR and create a clean one.

PR #40725, opened 2015-06-15
Formula: always print keg_only_text

### LINKED ISSUE #40724: Proposal: Brew info to display post-install information
Hi,

I propose to change `brew info` so it includes post-install information. Recently I have spent few days off line because I was unable to recompile some gems, the missing piece of the puzzle was unaccessible to me. 

And it's not the first time I have been bitten by it, but prior to this I was able to quickly find solutions on StackOverflow. And this time the information I have found was outdated so the switch didn't work. If I could see all the information about the installed formula it would save time.

Please consider this. The details are here: https://github.com/eventmachine/eventmachine/issues/602

mistydemeo:
Fixes #40724.

### DIFF
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -9,7 +9,7 @@ class Caveats
     caveats = []
     s = f.caveats.to_s
     caveats << s.chomp + "\n" if s.length > 0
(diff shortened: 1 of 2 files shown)


jacknagel:
Should we move this into caveats.rb instead?

mistydemeo:
Good thought; it doesn't really need to be on the Formula anymore in that case, either.

mikemcquaid:
:+1:

PR #40727, opened 2015-06-15
dnsmasq 2.73

nijikon:
I don't know if i'm suppose to change tha sha256 for bottle segment.

### DIFF
--- a/Library/Formula/dnsmasq.rb
+++ b/Library/Formula/dnsmasq.rb
@@ -1,14 +1,14 @@
 class Dnsmasq < Formula
   desc "Lightweight DNS forwarder and DHCP server"
   homepage "http://www.thekelleys.org.uk/dnsmasq/doc.html"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> I don't know if i'm suppose to change tha sha256 for bottle segment.

Just leave them as-is and they'll be updated when we pull, thanks!

nijikon:
@mikemcquaid so I should only change sha1 to sha256 for the package, and the bottle will automatically be updated from sha1 to sha256?

mikemcquaid:
@nijikon Yup :+1:

nijikon:
@mikemcquaid clear, thanks!

nijikon:
@mikemcquaid is this ok or there is something else that needs to be done?

mikemcquaid:
Thanks @nijikon!

Issue #40735, opened 2015-06-15
what would the command "brew man cask" do?


ajlowndes:
I am a bit of a newbie, and instead of typing "man brew-cask" I typed "brew man cask". Terminal then proceeded to dl and install "hpricot-0.8.6.gem", "rdiscount-2.1.8.gem", "mustache-1.0.1.gem" and "ronn-0.7.3.gem"... but I have no idea what these four things are or whether I need them. When I google those names they appear to be something to do with html development. Can anybody tell me what the hell I just installed?

bfontaine:
`brew man` generates the Homebrew manpage from markdown, using these gems to do the translation.

ajlowndes:
Awesome, thanks

PR #40736, opened 2015-06-15
'brew man': abort if a formula is given

bfontaine:
I’m at work, this is a quick fix for #40735.

### DIFF
--- a/Library/Homebrew/cmd/man.rb
+++ b/Library/Homebrew/cmd/man.rb
@@ -6,32 +6,30 @@ module Homebrew
   LINKED_PATH=HOMEBREW_PREFIX/"share/man/man1"
 
   def man
-    if ARGV.include?("--link") || ARGV.include?("-l")
+    abort <<-EOS.undent unless ARGV.named.empty?
+      This command updates the brew manpage and does not take formula names.
+    EOS
+
+    if ARGV.flag? "--link"
+      abort <<-EOS.undent if TARGET_PATH == LINKED_PATH
+        The target path is the same as the linked one, aborting.
+      EOS
       Dir["#{TARGET_PATH}/*.1"].each do |page|
         FileUtils.ln_s page, LINKED_PATH
         return
       end
-    end
+    else
+      Homebrew.install_gem_setup_path! "ronn"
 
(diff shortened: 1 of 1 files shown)


xu-cheng:
I think we can just check whether `HOMEBREW_DEVELOPER` is set.

At the same time, this command requires more cares. Can you help to fix these problems as well?

* `ARGV.include?("--link") || ARGV.include?("-l")` can be replaced by `ARGV.flag? "--link"`
* Linking manpage should abort if `HOMEBREW_REPOSITORY` is the same as `HOMEBREW_PREFIX`.
* `brew man --server` is broken. Maybe just remove such feature.
* It would be nicer to replace `if  ... return end` structure to `if ... else ... end`.

bfontaine:
Thanks @xu-cheng, I will look at all these when I’ll be home tonight.

mikemcquaid:
> I think we can just check whether HOMEBREW_DEVELOPER is set.

I don't think we want to do this as this is something contributors use periodically.

DomT4:
Yeah, I'd rather keep this available outside of `HOMEBREW_DEVELOPER`, even if we have to introduce logic to ensure people aren't passing arguments to the `brew man` command.

bfontaine:
@Homebrew/owners any objection to the `--server` feature removal?

mikemcquaid:
Makes sense to me; I almost removed it a few months ago.

DomT4:
No objections. Kill it with :fire:.

jacknagel:
Kill it

jacknagel:
What does `--link` do, btw?

bfontaine:
`--link` seems to be useful only if the Homebrew prefix is not the same directory as the git repository. Can this happen? I went back in the history and found the commit introducing `brew man` 5 years ago, dadcaee, and it already had this option.

jacknagel:
Oh, that makes sense. That's a supported configuration.

bfontaine:
Is everyone happy with this PR?

DomT4:
LGTM :rocket:

xu-cheng:
:+1:

mikemcquaid:
:+1:

PR #40745, opened 2015-06-15
update: multithread support

xu-cheng:
Before:
```
$ time brew update
Already up-to-date.
brew update  1.40s user 0.95s system 5% cpu 39.312 total
```

After:
```
$ time brew update
Already up-to-date.
brew update  1.47s user 1.06s system 45% cpu 5.585 total
```

More robust tests are required.  cc @Homebrew/owners for comments/test.

### DIFF
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -12,23 +12,32 @@ module Homebrew
     # ensure GIT_CONFIG is unset as we need to operate on .git/config
     ENV.delete('GIT_CONFIG')
 
-    cd HOMEBREW_REPOSITORY
-    git_init_if_necessary
+    HOMEBREW_REPOSITORY.cd { git_init_if_necessary }
 
     # migrate to new directories based tap structure
     migrate_taps
-
-    report = Report.new
-    master_updater = Updater.new(HOMEBREW_REPOSITORY)
-    master_updater.pull!
-    report.update(master_updater.report)
 
     # rename Taps directories
     # this procedure will be removed in the future if it seems unnecessasry
     rename_taps_dir_if_necessary
 
+    report = Report.new
+    tasks = Queue.new
+    semaphore = Mutex.new
+    master_updater = Updater.new(HOMEBREW_REPOSITORY)
+    thread_num = if ARGV.verbose?
+      1
+    else
+      Hardware::CPU.cores
+    end
+
+    tasks << Proc.new do
+      master_updater.pull!
+      master_updater.report
+    end
+
     Tap.each do |tap|
-      tap.path.cd do
+      tasks << Proc.new do
         updater = Updater.new(tap.path)
 
         begin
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cmd/update.rb
```diff
@@ -36,13 +45,27 @@ def update
         rescue
           onoe "Failed to update tap: #{tap}"
         else
-          report.update(updater.report) do |key, oldval, newval|
-            oldval.concat(newval)
+          updater.report
```
mikemcquaid: Probably want to report outside the thread.

Library/Homebrew/cmd/update.rb
```diff
@@ -12,23 +12,32 @@ def update
...
     # rename Taps directories
     # this procedure will be removed in the future if it seems unnecessasry
     rename_taps_dir_if_necessary
 
+    report = Report.new
+    tasks = Queue.new
+    semaphore = Mutex.new
+    master_updater = Updater.new(HOMEBREW_REPOSITORY)
+    thread_num = if ARGV.verbose?
+      1
```
mikemcquaid: This behaviour is a little unexpected, I think.

mikemcquaid:
I like the idea, it does annoy me how slow it is with a few taps tapped.

Library/Homebrew/cmd/update.rb
xu-cheng: Report inside thread is required as we want to speed up formula version checking.

Library/Homebrew/cmd/update.rb
xu-cheng: The output would be weird  if we combine the multithread and verbose output.

Library/Homebrew/cmd/update.rb
mikemcquaid: I don't think the actual output text should be inside the thread. Might need wider refactoring.

bfontaine:
I have 7 external taps + 4 `homebrew/*` ones here, it takes 16 seconds to `brew update` without this change, and only 5 seconds with it. :smile:

Library/Homebrew/cmd/update.rb
xu-cheng: `updater.report` returns a `Hash` object not text .

mikemcquaid: Sorry, confused it with `dump`. Ignore me :grinning:

xu-cheng:
This PR works fine so far. However, I found it will have weird output when handling interrupt.

```
$ brew update
^COne sec, just cleaning up
HEAD is now at 6525f03 pygame: don't apply patch to HEAD
HEAD is now at 051c9f6 small-light-module 0.6.9
Error: Failed to update tap: homebrew/python
Error: Failed to update tap: homebrew/nginx
HEAD is now at ea36f0a php5x-phalcon: update to 2.0.3 version (@baryshev)
HEAD is now at a85729e fssh HEAD (new formula)
Error: Failed to update tap: homebrew/php
Error: Failed to update tap: homebrew/head-only
HEAD is now at 5970b0e mysql55: update 5.5.44 bottle.
HEAD is now at eaabced kallisto: update 0.42.2 bottle.
HEAD is now at 883b960 lilypond: remove fontforge mandatory x11
Error: Failed to update tap: homebrew/versions
Error: Failed to update tap: homebrew/science
Error: Failed to update tap: homebrew/tex
HEAD is now at c814dc2 pcap_dnsproxy: add 0.4.2 bottle.
Error: Failure while executing: git pull -q origin refs/heads/master:refs/remotes/origin/master
```

Any idea to make it nice?

mikemcquaid:
@xu-cheng Maybe capture that output and print it later? I dunno.

xu-cheng:
I guess I have to make more wide refactoring to make it handle interrupt properly.

PR #40746, opened 2015-06-15
shadowsocks-libev v2.2.2

blackgear:


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/5678d6ec3c7f16dd83088dd2f7e8a19ab0ae6c95. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a tiny tiny nit for future: We don't use the `v` for version in commit messages, should be just `shadowsocks-libev 2.2.2`, Thanks!

PR #40747, opened 2015-06-15
pygobject3 3.16.2

tschoonj:
version bump
`make check` support removed since it will always fail: it
relies on linking executables against dynamically loadable modules which
works on Linux but is not supported on Mac OS X


tschoonj:
Strange segmentation fault when running `brew test gst-python`. I cannot reproduce it locally.

Issue #40539, opened 2015-06-10
kumofs failed to build on 10.9


L2G:
https://gist.github.com/78636b7647fbdd6c4316

mistydemeo:
Looks like this doesn't build with modern clang / libc++. Unfortunately, the upstream project appears to be defunct (no development in close to five years), so this may not be fixable.

L2G:
I ran into this because I was intentionally looking for formulae that hadn't had substantial updates in a while and might be broken.

I didn't want to suggest kicking it to the boneyard straightaway, but if it looks that bad…

mistydemeo:
Thank you! It looks like it paid off. :D

It looks like it'll take some work to fix this, and I'm not inclined to do so given the lack of upstream activity.

L2G:
I'm going to need to create a new pull request because I'm unable to convert this issue ticket to a pull request.

mikemcquaid:
@L2G Ok, closing in in favour of that. Please link this issue there so it is referenced, thanks!

PR #40756, opened 2015-06-15
kumofs: retire to boneyard

L2G:
It has serious build issues, no dependents, and no updates upstream since December 2010.

Watch for a corresponding pull request in the homebrew-boneyard tap.

See also #40539.


mikemcquaid:
Thanks @L2G!

PR #40757, opened 2015-06-15
Various

DomT4:
Various fixes related to #40392


mikemcquaid:
Looks like a failing patch but otherwise :+1:

DomT4:
Yeah, I apparently screwed up my own checksum, which is embarrassing :wink:. Will go double check it in a sec.

PR #40758, opened 2015-06-15
Fix checksum for unison

toonetown:
It appears that the checksum is not correct in the brew file.


DomT4:
See #40711. Will be merged soon. Thanks for the PR!

PR #40770, opened 2015-06-16
doctor: use Formula.installed directly

xu-cheng:
Before
```
brew doctor  9.83s user 0.52s system 76% cpu 13.563 total
```

After
```
brew doctor  0.25s user 0.16s system 9% cpu 4.332 total
```


mikemcquaid:
Wow. :+1:

chdiza:
This is great.  I noticed that with 79 packages installed, I only see a speedup the second time I run `brew doctor`.  Probably a disk caching thing?

PR #38891, opened 2015-04-21
[WIP] Cleanup requirements more

mikemcquaid:
Following on from #38815. Still need to alias the requirements so they work with the old names.

PR #40771, opened 2015-06-16
Cleanup formula requirements

mikemcquaid:
Followup from #38891 to cleanup requirements inside formulae (and fix some strict audits while we're at it).


mikemcquaid:
Fixed the errors locally before pushing.

PR #40777, opened 2015-06-16
doctor: fix __check_linked_brew

xu-cheng:
This fixes the case when multiple versions of a formula are installed.


mikemcquaid:
:+1:

jacknagel:
Pulled since the earlier work is broken without this.

DomT4:
Can confirm this fixes the breakage in #40781 and related issues.

PR #40779, opened 2015-06-16
--repository handling moved at the top of brew.rb

### LINKED ISSUE #40778: Unwanted warning on El Capitan during bash-completion
`brew install ba<TAB>Warning: Although Homebrew has added initial recognition of 10.11, it is unsupported.
You may encounter breakage or other failure and there is no guarantee
Homebrew can resolve those issues until El Capitan is stable.`

bfontaine:
The Bash completion uses this option so it must be fast.

It also fixes #40778.

cc @Homebrew/owners for review.

### DIFF
--- a/Library/Homebrew/cmd/--repository.rb
+++ b/Library/Homebrew/cmd/--repository.rb
@@ -1,5 +1,0 @@
-module Homebrew
-  def __repository
-    puts HOMEBREW_REPOSITORY
-  end
-end
--- a/Library/brew.rb
+++ b/Library/brew.rb
@@ -14,6 +14,11 @@ HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent.join("Homebrew")
 $:.unshift(HOMEBREW_LIBRARY_PATH.to_s)
 require 'global'
 
+if ARGV == %w{--repository}
+  puts HOMEBREW_REPOSITORY
+  exit 0
+end
+
 if ARGV.first == '--version'
   puts HOMEBREW_VERSION
   exit 0


xu-cheng:
Maybe we should move OS X 10.11 warning outside of `brew.rb`. `brew --repository` may not be the only command which breaks due to this warning. Potentially, it will also break other existing scripts/tools which invoke brew command.

xu-cheng:
For example, that warning will break `brew list|tap|search|outdated` etc used by shell completion.

jacknagel:
Hmm, 'd rather make loading Homebrew fast than start special casing more things. I did a lot of work to making simple commands fast last year, it's possible (likely) we've regressed since then.

xu-cheng:
I agree with @jacknagel that we should avoid such special casing.

@bfontaine I don't think this PR would fix #40778 or speed up shell completion, because there're other brew commands invoked by completion.

We should definitely revert 4fb36339eea733fb0a016f39b3889b393a5df25. And move OS X 10.11 warning to `brew install`?

jacknagel:
Anyway, how much faster can this PR really make it? The special-case logic still happens after loading the primary core code (triggered by loading global.rb)

mikemcquaid:
@xu-cheng I'd rather it's kept a bit wider than just `brew install` (which generates a lot of text and people are likely to miss it). It's cut down on the bad bug reports.

bfontaine:
Thank you both, closing this.

xu-cheng:
> @xu-cheng I'd rather it's kept a bit wider than just brew install (which generates a lot of text and people are likely to miss it). It's cut down on the bad bug reports.

Ideally only installing formulae would fail, right? So how about moving it to formula_installer along with all the checks in `brew install`?

mikemcquaid:
@xu-cheng I guess an alternative would be moving it to `formula_installer` but ensuring that we display this output before and after the build (so people stop filing bad bug reports)

xu-cheng:
We suggested the same thing. :smiley:

jacknagel:
In my installer/deps rewrite branch I have removed all of the UX-related things to the UX layer and out of the installer. I'd appreciate it if we didn't move things in the opposite direction in the meantime.

jacknagel:
Maybe we can put it in the failed build text: https://github.com/Homebrew/homebrew/blob/8dfb519ac9fd44755e6caabe376fc7590d604214/Library/Homebrew/exceptions.rb#L182-L220

and also in doctor, and then call the doctor check in the preinstall checks: https://github.com/Homebrew/homebrew/blob/8dfb519ac9fd44755e6caabe376fc7590d604214/Library/Homebrew/cmd/install.rb#L112-L123 (these are called for both install and upgrade)

mikemcquaid:
> Maybe we can put it in the failed build text: https://github.com/Homebrew/homebrew/blob/8dfb519ac9fd44755e6caabe376fc7590d604214/Library/Homebrew/exceptions.rb#L182-L220
> 
> and also in doctor, and then call the doctor check in the preinstall checks: https://github.com/Homebrew/homebrew/blob/8dfb519ac9fd44755e6caabe376fc7590d604214/Library/Homebrew/cmd/install.rb#L112-L123 (these are called for both install and upgrade)

All sound good to me :+1:

DomT4:
No objections to the text being moved. Nobody seems to have objected to actually being so visibly, regularly reminded that 10.11 is unsupported at this point, so there's validity to keeping it prominent, but as long as it cuts down on the amount of simple issues we're dealing with I'm happy.

I think our bash completion was about the only command I didn't personally test, so that'll teach me.

PR #40780, opened 2015-06-16
bgpq3 0.1.30

job:
Mostly bug fixes, compared to 0.1.27:

    - bugfix: private asns with number > 2^31 were printed as negative integers.
    - do not use ASNs reserved for documentation purposes and private use
    - allow as-path generation with BIRD output.
    - merge README.md changes
    - bugfix: incorrect asdot representation (as101. without symbols after dot) is not allowed anymore.
    - do not include routes registered for AS23456 (transition-as) by default.Use new option -2 to restore old behaviour.
    - minor changes: .spec update, non-silent failure on wrong af, more room for masklen

### DIFF
--- a/Library/Formula/bgpq3.rb
+++ b/Library/Formula/bgpq3.rb
@@ -3,8 +3,8 @@ require 'formula'
 class Bgpq3 < Formula
   desc "bgp filtering automation for Cisco and Juniper routers"
   homepage 'http://snar.spb.ru/prog/bgpq3/'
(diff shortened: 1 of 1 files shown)


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/4cb2bf2e989e9a7668b00a27ee81a334c99f7588. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Please note for future that `sha256` is the preferred checksum now, Thanks!

PR #40787, opened 2015-06-16
shocco: use https in resource

vszakats:


DomT4:
:rocket:. Thanks @vszakats!

PR #40789, opened 2015-06-16
Update smartmontools to 6.4

samm-git:
Signed-off-by: Alex Samorukov <samm@os2.kiev.ua>

### DIFF
--- a/Library/Formula/smartmontools.rb
+++ b/Library/Formula/smartmontools.rb
@@ -2,7 +2,7 @@ require 'formula'
 
 class Smartmontools < Formula
   desc "SMART hard drive monitoring"
-  homepage 'http://sourceforge.net/apps/trac/smartmontools/'
+  homepage 'https://www.smartmontools.org/'
   url 'https://downloads.sourceforge.net/project/smartmontools/smartmontools/6.4/smartmontools-6.4.tar.gz'
   sha1 '855c7d555dd405e5b392b1631dc36dd9632db8b8'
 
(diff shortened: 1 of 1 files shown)


Library/Formula/smartmontools.rb
```diff
@@ -3,8 +3,8 @@
 class Smartmontools < Formula
   desc "SMART hard drive monitoring"
   homepage 'http://sourceforge.net/apps/trac/smartmontools/'
```
vszakats: This URL is no longer valid; would you update it to its permanent redirect (+https) ?:
`homepage "https://sourceforge.net/projects/smartmontools/"`

samm-git: Correct homepage URL is http://smartmontools.org/. Sourceforge is used only as SVN server and file storage.

vszakats: Even better! I'd suggest using the https version of that as well:
`homepage "https://www.smartmontools.org"`

samm-git: I am ok with https. Should i update pull request?

vszakats: Yes, please!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/3acb81804ae0d9de04b48b6a949099e7ec41f351. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Please note for future the preferred commit message for version bumps is `smartmontools 6.4`, Thanks!

PR #40795, opened 2015-06-16
search --desc: cache the query regexp

bfontaine:
I haven’t noticed any real performance improvement for the average case, but it’s still better to avoid one function call at each iteration.


mistydemeo:
:+1:

xu-cheng:
:+1: good work

>  I haven’t noticed any real performance improvement for the average case,

One thing on my todo list is to investigate the possibility to improve the performance on formula loading speed. We only have 4000 formulae in core but requires several seconds to load all of them. In my opinion, there should be some improve space.

PR #40804, opened 2015-06-17
bash-completion: use https urls

vszakats:
also use 'commitdiff_plain' (in commented url)


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/1bc9b9259d9447bd6a95737e6a7c20bce8346b40. Thanks @vszakats!

PR #40806, opened 2015-06-17
audit: flag re-added tap migrations

DomT4:
Re https://github.com/Homebrew/homebrew/commit/411c2178448916d4005c059bc35ee780b5af53f5#commitcomment-11714539

Not going to claim this is the *most* beautiful code in the world, but it seems to work well. This is usually the part where I submit some initial code, and then Xu wanders by and tells me how it can be improved by several thousand degrees until everyone is happier with it :wink:.

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -3,6 +3,7 @@ require "utils"
 require "extend/ENV"
 require "formula_cellar_checks"
 require "official_taps"
+require "tap_migrations"
 require "cmd/search"
 
 module Homebrew
@@ -769,6 +770,20 @@ class FormulaAuditor
     end
   end
 
+  def audit_reverse_migration
+    # Only enforce for new formula being re-added to core
+    return unless @strict
+    return unless formula.core_formula?
+
+    if TAP_MIGRATIONS.has_key?(formula.name)
+      problem <<-EOS.undent
+       #{formula.name} seems to be listed in tap_migrations.rb!
+       Please remove #{formula.name} from present tap & tap_migrations.rb
+       before submitting it to Homebrew/homebrew.
+      EOS
+    end
+  end
+
   def audit_prefix_has_contents
     return unless formula.prefix.directory?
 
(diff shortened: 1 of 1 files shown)


DomT4:
CC @Homebrew/owners

Library/Homebrew/cmd/audit.rb
```diff
@@ -747,6 +748,20 @@ def audit_caveats
     end
   end
 
+  def audit_reverse_migration
+    # Only enforce for new formula being re-added to core
+    return unless @strict
+    return unless formula.tap.downcase == "homebrew/homebrew"
+
+    unless TAP_MIGRATIONS.map(&:to_s).grep(/#{formula.name}/).empty?
```
bfontaine: Shouldn’t we just check the keys? `TAP_MIGRATIONS.has_key? formula.name`

DomT4: We can, but I was also planning to pull the tap information out of `TAP_MIGRATIONS` and then print that in place of the words `present tap`. Just hadn't decided the best way to do so yet :smile_cat:. PR is still a WIP, just wanted to get discussion rolling on the general idea.

xu-cheng: I think it is best to avoid such string matching.  It's bad for performance and likeky to break if there's any change in upper layer data structure.

Library/Homebrew/cmd/audit.rb
```diff
@@ -747,6 +748,20 @@ def audit_caveats
     end
   end
 
+  def audit_reverse_migration
+    # Only enforce for new formula being re-added to core
+    return unless @strict
+    return unless formula.tap.downcase == "homebrew/homebrew"
```
xu-cheng: `fotmula.core_formula?`

Library/Homebrew/cmd/audit.rb
xu-cheng: On second thought, string matching is incorrect. It will return false result on partial matching.

mikemcquaid:
:+1: to the approach and the comments.

Library/Homebrew/cmd/audit.rb
DomT4: WFM :+1:.

DomT4:
Finally looped back around to this PR and addressed the comments. Feel free to review again.

mikemcquaid:
:+1:

PR #40807, opened 2015-06-17
Add CocoaPods 0.37.2 formula

keith:
See more conversation about this here: https://github.com/CocoaPods/CocoaPods/pull/3010

// @mikemcquaid

### DIFF
--- a/Library/Formula/cocoapods.rb
+++ b/Library/Formula/cocoapods.rb
@@ -3,8 +3,6 @@ class Cocoapods < Formula
   url "http://keith.github.io/CocoaPods/cocoapods-0.37.2.tar.gz"
   sha1 "f727a8027a747f01323ee1188c4ce654731e3e51"
 
-  depends_on "xcproj" => :recommended
-
   def install
     prefix.install "vendor"
     prefix.install "lib" => "rubylib"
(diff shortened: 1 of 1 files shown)


Library/Formula/cocoapods.rb
```diff
@@ -0,0 +1,18 @@
+class Cocoapods < Formula
+  homepage "https://github.com/keith/cocoapods/"
```
keith: I'm happy to update this homepage to the main repo if that makes more sense.

Library/Formula/cocoapods.rb
```diff
@@ -0,0 +1,18 @@
...
+  url "http://keith.github.io/CocoaPods/cocoapods-0.37.2.tar.gz"
+  sha1 "f727a8027a747f01323ee1188c4ce654731e3e51"
+
+  depends_on "xcproj" => :recommended
```
segiddins: Xcproj isn't used any more?

keith: Awesome!

mikemcquaid:
Hey @keith, it seems `brew install Caskroom/cask/cocoapods` works now so I'm wondering if we still need this. What do you think?

keith:
Personally I'm not really interested in dealing with another tool too install a GUI to install a CLI. But if you think that's the way to go I can close this.

mikemcquaid:
In this case it seems to install `pod` into `/usr/local/bin` with the single command. Dunno if that affects things?

keith:
Yea it does. I'm not a huge fan of being dependent on cask and the GUI for that though. Like I said if you think that's the way to go that's fine though.

mikemcquaid:
I think I'll pass on this in favour of that just as it's officially supported upstream. It still annoys me, for what it's worth, though :sob:

PR #40819, opened 2015-06-17
juju 1.24.0

sinzui:


DomT4:
Was merged a few days back in https://github.com/Homebrew/homebrew/commit/c528ab92e1d9de5de3d6b95217ca37b1dcbd3291. Please `brew update`.

sinzui:
That means the juju installed was broken for 4 days. The juju version in homebrew cannot be updated until I someone on my team publish the companion agents to the released simple streams.

sinzui:
Does the homebrew bot have network access when tests are run. I can write a test that will show fail if the version of juju is not in the released streams.

DomT4:
> That means the juju installed was broken for 4 days. 

What are the runtime consequences of this? What would users have experienced?

sinzui:
It is not possible to bootstrap a new environment or upgrade or an existing environment. One day your configs in environments.yaml work, then after the update they do not. Users can scale an existing environment.

When we publish a release, there is a waiting period before we publish it in the released streams for everyone to get by default. During the "proposed" period testers can choose to change their configs to
    agent-stream: proposed
to use the new juju. If a user reports a regression, we may retract the version, or immediately supersede it with another version.

 I can a to the current test to verify the juju version is in the released streams

mikemcquaid:
> I can a to the current test to verify the juju version is in the released streams

That's be great, thanks!

DomT4:
Might be worth adding a small comment at the top of the formula with something like:

```ruby
# Before updating versions please ensure https://example has xyz available
```

Or such, as well.

PR #40823, opened 2015-06-17
gtkdatabox: use https homepage

vszakats:


DomT4:
Merged this and #40821.

This isn't a team decision, but I'm generally :-1: on bumping Sourceforge homepages to `https` until Sourceforge stop routinely downgrading them to `http`. We have no way of being honest with users that the `homepage` redirects to plaintext until their browser lets them know.

vszakats:
Well, at least it stays https if one is logged in. Not much gain, that's for sure; my (light) motive for that is to clear those http hits off of grep searches.

PR #40830, opened 2015-06-17
updating fftw.rb

francispoulin:
This is a very minor change in fftw.rb to allow the user to enable openmp.


bfontaine:
Thank you for your contribution @francispoulin, we really appreciate it ! :tada: 

Tiny note for the future: the preferred commit message format is “`fftw: allow to enable openmp`”.

PR #40831, opened 2015-06-17
sparse: New Formula

veprbl:


Library/Formula/sparse.rb
```diff
@@ -0,0 +1,16 @@
+class Sparse < Formula
+  desc "Static C code analysis tool"
+  homepage "https://sparse.wiki.kernel.org/"
+  url "https://www.kernel.org/pub/software/devel/sparse/dist/sparse-0.5.0.tar.xz"
+  sha256 "921fcf918c6778d1359f3886ac8cb4cf632faa6242627bc2ae2db75e983488d5"
+  head "git://git.kernel.org/pub/scm/devel/sparse/sparse.git"
```
DomT4: Let's use `https://git.` instead of `git://git.`

Library/Formula/sparse.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def install
+    inreplace "Makefile", /PREFIX=\$\(HOME\)/, "PREFIX=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/sparse"
```
DomT4: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

veprbl:
Updated. Thanks.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `sparse 0.5.0 (new formula)`.

PR #40842, opened 2015-06-18
Pass formula objects, not strings, into the build environment

jacknagel:
This is cleanup that emerged from a larger refactoring of the build environment code.


mikemcquaid:
:+1:

PR #40846, opened 2015-06-18
rabbitmq-c 0.7.0

guoxiao:
update `rabbitmq-c` depends
* `rabbitmq-c` does not depend on `rabbitmq` server.
* `rabbitmq-c` depends on `openssl`

update to version 0.7.0

### DIFF
--- a/Library/Formula/rabbitmq-c.rb
+++ b/Library/Formula/rabbitmq-c.rb
@@ -37,4 +37,8 @@ class RabbitmqC < Formula
     system "cmake", ".", *args
     system "make", "install"
   end
+
+  test do
+    system "amqp-get", "--help"
+  end
 end
--- a/Library/Formula/rabbitmq-c.rb
+++ b/Library/Formula/rabbitmq-c.rb
@@ -1,8 +1,8 @@
 class RabbitmqC < Formula
   desc "RabbitMQ C client"
   homepage "https://github.com/alanxz/rabbitmq-c"
-  url "https://github.com/alanxz/rabbitmq-c/archive/v0.5.2.tar.gz"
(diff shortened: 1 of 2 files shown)


Library/Formula/rabbitmq-c.rb
```diff
@@ -31,10 +29,10 @@ def install
     args << "-DBUILD_API_DOCS=OFF"
 
     args << if build.with? "tools"
-      "-DBUILD_TOOLS=ON"
-    else
-      "-DBUILD_TOOLS=OFF"
-    end
+              "-DBUILD_TOOLS=ON"
```
DomT4: Indentation has slipped way off here. Suspect you've indented with two tabs instead of two spaces.

guoxiao: It's not a mistake. I aligned else with if to fix `brew audit --strict` warning.

mikemcquaid:
`simple-amqp-client` (and perhaps other `brew uses rabbitmq-c` things) need a `revision` added or increased.

guoxiao:
Some php extensions should be updated to be compatible with updated `rabbitmq-c`.
see homebrew/homebrew-php#1806

DomT4:
Is this good to :ship: or?

mikemcquaid:
Yup :+1:

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/92c66867f2914bcf50df3ad799573a89e09dc8c9. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

If you could squash your commits on final push in future that'd be appreciated. We try to keep the commit tree relatively clean and squashing locally isn't always easy when you aren't the person who wrote the commits. Thanks!

PR #40848, opened 2015-06-18
uberftp 2.8

giffels:
Hi,

version has bumped to 2.8. In addition, I fixed a problem that uberftp do not compile with globus-toolkit versions >= 6.0. Could you review and merge, please.

Thanks,
Manuel

### DIFF
--- a/Library/Formula/uberftp.rb
+++ b/Library/Formula/uberftp.rb
@@ -9,6 +9,7 @@ class Uberftp < Formula
   def install
     globus = Formula["globus-toolkit"].opt_prefix
 
+    #patches needed since location changed with globus-toolkit versions >= 6.0, patch to upstream not yet merged https://github.com/JasonAlt/UberFTP/pull/8
     inreplace "configure", "globus_location/include/globus/gcc64dbg", "globus_location/libexec/include"
     inreplace "configure", "globus_location/lib64", "globus_location/libexec/lib"
 
--- a/Library/Formula/uberftp.rb
+++ b/Library/Formula/uberftp.rb
@@ -1,23 +1,18 @@
 class Uberftp < Formula
   desc "Interactive GridFTP client"
-  homepage "http://dims.ncsa.illinois.edu/set/uberftp/"
-  url "https://github.com/JasonAlt/UberFTP/archive/Version_2_7.tar.gz"
-  sha256 "29a111a86fa70dbbc529a5d3e5a6befc1681e64e32dc019a1a6a98cd43ffb204"
+  homepage 'http://dims.ncsa.illinois.edu/set/uberftp/'
+  url 'https://github.com/JasonAlt/UberFTP/archive/Version_2_8.tar.gz'
+  sha256 '8a397d6ef02bb714bb0cbdb259819fc2311f5d36231783cd520d606c97759c2a'
 
   depends_on "globus-toolkit"
 
   def install
-    # get the flavor
     globus = Formula["globus-toolkit"].opt_prefix
 
-    core = `"#{globus}/sbin/gpt-query" globus_core`
-    flavor = case core
-    when /gcc64dbg/ then "gcc64dbg"
-    when /gcc32dbg/ then "gcc32dbg"
-    end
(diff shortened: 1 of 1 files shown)


Library/Formula/uberftp.rb
```diff
@@ -3,24 +3,33 @@
...
   def install
-    # get the flavor
     globus = Formula["globus-toolkit"].opt_prefix
+    globus_version = Formula["globus-toolkit"].version
 
-    core = `"#{globus}/sbin/gpt-query" globus_core`
-    flavor = case core
-      when /gcc64dbg/ then "gcc64dbg"
-      when /gcc32dbg/ then "gcc32dbg"
+    if globus_version.to_s[0].to_i >= 6
```
mikemcquaid: We always know what that version would be so we shouldn't need to make an `if` here?

giffels: If somebody pinned the globus version to an older version, he can't compile uberftp anymore. However, if that is not a use case I can drop the ```if``` condition here.

mikemcquaid: Yeh, I'd drop the conditional here, thanks.

giffels: Done!

Library/Formula/uberftp.rb
```diff
@@ -3,23 +3,18 @@
...
   def install
-    # get the flavor
     globus = Formula["globus-toolkit"].opt_prefix
 
-    core = `"#{globus}/sbin/gpt-query" globus_core`
-    flavor = case core
-      when /gcc64dbg/ then "gcc64dbg"
-      when /gcc32dbg/ then "gcc32dbg"
-    end
+    inreplace "configure", "globus_location/include/globus/gcc64dbg", "globus_location/libexec/include"
```
mikemcquaid: Looks like this would break 32-bit builds? Can you please submit these patches to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

giffels: ```gpt-query``` is not available anymore in globus version 6.0 and there is no differentiation between  gcc64dbg  and gcc32dbg anymore. In addition, have a look at https://github.com/Homebrew/homebrew/pull/22037#discussion_r5982281, please.
I will prepare a patch and send it to the upstream developers, for the time being this patch is fixing the uberftp compilation errors. I will add a comment to the formula file.

mikemcquaid: Ok, cool. Let's get the patch submitting upstream before we include this, thanks!

dunn:
Any movement upstream?

giffels:
There is already a pull request upstream https://github.com/JasonAlt/UberFTP/pull/8, which will solve parts of the problem. No reaction since 18th, June.

bfontaine:
Does using the proposed patch removes the need for these `inreplace` calls? If so you can use a `patch do` instead:

```
patch do
  url "https://patch-diff.githubusercontent.com/raw/JasonAlt/UberFTP/pull/8.patch"
  sha256 "..."
end
```

Library/Formula/uberftp.rb
```diff
@@ -1,23 +1,19 @@
 class Uberftp < Formula
   desc "Interactive GridFTP client"
-  homepage "http://dims.ncsa.illinois.edu/set/uberftp/"
-  url "https://github.com/JasonAlt/UberFTP/archive/Version_2_7.tar.gz"
-  sha256 "29a111a86fa70dbbc529a5d3e5a6befc1681e64e32dc019a1a6a98cd43ffb204"
+  homepage 'http://dims.ncsa.illinois.edu/set/uberftp/'
+  url 'https://github.com/JasonAlt/UberFTP/archive/Version_2_8.tar.gz'
+  sha256 '8a397d6ef02bb714bb0cbdb259819fc2311f5d36231783cd520d606c97759c2a'
```
bfontaine: Please use double quotes here.

bfontaine:
Ping?

bfontaine:
Closing due to no response. Feel free to open another PR that addresses the comments above. Thanks!

PR #40849, opened 2015-06-18
Formulary.from_rack: reduce io through caching

xu-cheng:
cc @jacknagel for opinion.

### DIFF
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -170,20 +170,25 @@ class Formulary
     loader_for(ref).get_formula(spec)
   end
 
-  # Return a Formula instance for the given rack.
-  def self.from_rack(rack, spec=:stable)
+  def self.rack_to_ref(rack)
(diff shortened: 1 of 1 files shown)


Library/Homebrew/formulary.rb
```diff
@@ -170,22 +170,27 @@ def self.factory(ref, spec=:stable)
...
     else
-      factory("#{tap.sub("homebrew-", "")}/#{rack.basename}", spec)
+      "#{tap.sub("homebrew-", "")}/#{rack.basename}"
     end
   end
 
+  # Return a Formula instance for the given rack.
+  def self.from_rack(rack, spec=:stable)
+    @@rack_to_ref ||= Hash.new { |cache, key| cache[key] = rack_to_ref(key) }
+    factory(@@rack_to_ref[rack], spec)
```
jacknagel: Same comment about class variables here.

xu-cheng:
After some test, I decide to pass this as it doesn't improve the performance as I participated.

PR #40850, opened 2015-06-18
juju-quickstart 2.2.0

frankban:
* Add support for loading uncommitted bundles on the Juju GUI.
* Allow configuring the Juju GUI so that it listens to a customized port.
* On existing environments, automatically detect the port used by the GUI server.
* Fix SSH agent handling when using uncommon shells.


xu-cheng:

PR #40851, opened 2015-06-18
orbit: modernise

DomT4:
This threw up what looked like a race condition earlier. It needed updating/bottling regardless, so I thought I'd leave off the `ENV.deparallelize` during the first run and see if the race actually reproduces.

### DIFF
--- a/Library/Formula/orbit.rb
+++ b/Library/Formula/orbit.rb
@@ -1,29 +1,31 @@
-require 'formula'
-
 class Orbit < Formula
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

DomT4:
I can't repro the race condition in about ~20 builds locally, or on the bots here, so I'm tempted to just merge this with bottle and see if we hit it anywhere else in future. It's possible the X11 Yosemite bot just threw an unrelated tantrum this morning.

mikemcquaid:
Sounds good.

PR #40855, opened 2015-06-18
Formula: cache methods' output to reduce io

xu-cheng:
cc @jacknagel for opinion.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -624,22 +624,22 @@ class Formula
 
   # an array of all core {Formula} names
   def self.core_names
-    Dir["#{HOMEBREW_LIBRARY}/Formula/*.rb"].map{ |f| File.basename f, ".rb" }.sort
+    @core_names ||= Dir["#{HOMEBREW_LIBRARY}/Formula/*.rb"].map{ |f| File.basename f, ".rb" }.sort
   end
 
   # an array of all tap {Formula} names
   def self.tap_names
-    Tap.map(&:formula_names).flatten.sort
+    @tap_names ||= Tap.map(&:formula_names).flatten.sort
   end
 
   # an array of all {Formula} names
   def self.names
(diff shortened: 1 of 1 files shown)


jacknagel:
Any benchmarks? I don't think we need to cache unless it's a hotspot. And if we do, we need to make sure that no callers depend on the value reflecting changes between calls (I don't think they do).

Usually callers cache the value of e.g. `Formula.installed`.

Library/Homebrew/formula.rb
```diff
@@ -624,22 +624,22 @@ def python(options={}, &block)
 
   # an array of all core {Formula} names
   def self.core_names
-    Dir["#{HOMEBREW_LIBRARY}/Formula/*.rb"].map{ |f| File.basename f, ".rb" }.sort
+    @@core_names ||= Dir["#{HOMEBREW_LIBRARY}/Formula/*.rb"].map{ |f| File.basename f, ".rb" }.sort
```
jacknagel: `@core_names` on the class is fine; we don't need to use class variables.

Library/Homebrew/formula.rb
```diff
@@ -624,22 +624,22 @@ def python(options={}, &block)
...
   # an array of all {Formula} names
   def self.names
-    (core_names + tap_names.map { |name| name.split("/")[-1] }).sort.uniq
+    @@names ||= (core_names + tap_names.map { |name| name.split("/")[-1] }).sort.uniq
   end
 
   # an array of all {Formula} names, which the tap formulae have the fully-qualified name
   def self.full_names
-    core_names + tap_names
+    @@full_names ||= (core_names + tap_names)
```
jacknagel: parentheses are superfluous here.

xu-cheng:
Caching these result can significantly improve the performance on `brew audit --strict` with multiple formulae (especially when auditing all of formulae). And I haven't found caller participates the changes between calls.

I have addressed your comment and update the PR.

PR #40856, opened 2015-06-18
Add revision informations to haxe version

alebianco:
will print branch and commit hash as part of the version number when calling 'haxe -v'


mikemcquaid:
Can you give example output before and after? Thanks!

alebianco:
Sure, there you go

before: Haxe Compiler 3.3.0 - (C)2005-2015 Haxe Foundation
after: Haxe Compiler 3.3.0 (git build development @ 0a43a5e) - (C)2005-2015 Haxe Foundation

mikemcquaid:
I think this should only apply when `build.head?` is true.

alebianco:
Ideally I'd like to see it always when it's built from the sources. If you think it could be a problem then yeah, at least when building from head

mikemcquaid:
Yeh, let's just do for head, thanks!

jacknagel:
```
Error: uninitialized constant SharedEnvExtension::COMPILERS
 PASSED
Please report this bug:
/usr/local/Library/Homebrew/extend/ENV/shared.rb:124:in `<module:SharedEnvExtension>'
/usr/local/Library/Homebrew/extend/ENV/shared.rb:4:in `<top (required)>'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
/usr/local/Library/Homebrew/extend/ENV.rb:2:in `<top (required)>'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
/usr/local/Library/Homebrew/requirement.rb:172:in `yielder'
/usr/local/Library/Homebrew/requirement.rb:53:in `satisfied?'
/usr/local/Library/Homebrew/cmd/test-bot.rb:344:in `block in satisfied_requirements?'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/set.rb:232:in `each_key'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/set.rb:232:in `each'
/usr/local/Library/Homebrew/dependencies.rb:68:in `each'
/usr/local/Library/Homebrew/cmd/test-bot.rb:342:in `reject'
/usr/local/Library/Homebrew/cmd/test-bot.rb:342:in `satisfied_requirements?'
/usr/local/Library/Homebrew/cmd/test-bot.rb:397:in `formula'
/usr/local/Library/Homebrew/cmd/test-bot.rb:639:in `block in run'
/usr/local/Library/Homebrew/cmd/test-bot.rb:638:in `each'
/usr/local/Library/Homebrew/cmd/test-bot.rb:638:in `run'
/usr/local/Library/Homebrew/cmd/test-bot.rb:822:in `test_bot'
/usr/local/Library/brew.rb:140:in `<main>'
````

This exception that we're seeing CI is both unrelated to this PR, and baffling.

The `COMPILERS` constant is defined in the `CompilerConstants` module, which is included in `SharedEnvExtension`. The include happens on line 5, and is successful otherwise an exception would be raised on that line. And yet, the first reference to `COMPILERS`, on line 124 raises an exception.

The weird thing is that this code is loaded in the core test suite, and in every single source build, and never results in an error.

I suspect it has something to do with how it is being invoked here (during evaluation of a requirement), but I'm unable to reproduce it locally.

**Comment by alebianco** [comment]
Hi @jacknagel sorry for the late response, I just noticed your comment!
Is there a better (safer) way to check if it's installing from head at this point?

File: Library/Formula/haxe.rb [hunk 1]
```diff
@@ -26,7 +26,7 @@ class Haxe < Formula
   def install
     # Build requires targets to be built in specific order
     ENV.deparallelize
-    system "make", "OCAMLOPT=ocamlopt.opt"
+    system "make", "OCAMLOPT=ocamlopt.opt", ("ADD_REVISION=1" if build.head?)
```
bfontaine: I don’t think this will work if `build.head?` is false because the last argument will then be `nil`:

```
>> system "ls", ("foo" if false)
TypeError: no implicit conversion of nil into String
	from (irb):1:in `system'
	from (irb):1
	from /usr/local/Library/brew.rb:124:in `<main>'
```

alebianco: as far as i can tell, it does work correctly locally

bfontaine: Could you please try the following?

```
$ brew irb
>>> system "ls", ("." if false)
```

alebianco: that gives the same RTE as you saw.
i'll make a new test tonight to launch the formula for both head and regular install and see if maybe something changed in the last month or so ...

DomT4:
Ping.

alebianco:
@bfontaine was right, it was breaking the installation of the stable version, not sure how I missed that.
these are the log of the current patched version compare to the master version https://gist.github.com/alebianco/fa93c4cda1330b4bdc30

PR #40859, opened 2015-06-18
download_strategy: don't check url if already downloaded

xu-cheng:
Better diff view: https://github.com/Homebrew/homebrew/pull/40859/files?w=1


mikemcquaid:
:+1:

PR #40860, opened 2015-06-18
dnsmasq: add -C option to sample plist to use config file from correct homebrew prefix

oker1:
Since dnsmasq is installed from keg the default config location is not the same as the homebrew prefix if you use a custom one. The patch adds the -C option to the plist to point to the correct location of the config file.


mikemcquaid:

PR #40861, opened 2015-06-18
fdk-aac-encoder 0.6.2 (new formula)

jrnewell:


Library/Formula/fdkaac.rb
```diff
@@ -0,0 +1,26 @@
+require 'formula'
```
bfontaine: You can remove this line.

Library/Formula/fdkaac.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class Fdkaac < Formula
+  desc "Command-line encoder frontend for libfdk-aac"
+  homepage "https://github.com/nu774/fdkaac"
+  url "https://github.com/nu774/fdkaac/archive/v0.6.2.tar.gz"
+  version "0.6.2"
```
bfontaine: You can remove this, Homebrew detects it from the URL.

Library/Formula/fdkaac.rb
```diff
@@ -0,0 +1,26 @@
...
+  sha256 "de758d6e651e81e9be89d2972612fc5b96cb70321234c3339f35483b636458ad"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on 'pkg-config' => :build
```
bfontaine: Please use double-quoted strings everywhere.

Library/Formula/fdkaac.rb
```diff
@@ -0,0 +1,26 @@
...
+  depends_on 'fdk-aac'
+
+  def install
+    system "autoreconf -i"
```
bfontaine: Use `system "autoreconf", "-i"` instead of `system "autoreconf -i"`.

Library/Formula/fdkaac.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}",
+                          "--mandir=#{man}"
+    system "make", "install"
+  end
```
bfontaine: Could you please add a test with a `test do` block? Thanks!

DomT4:
Test is breaking:

```
Error: fdkaac: failed
Failed executing: /usr/local/Cellar/fdkaac/0.6.2/bin/fdkaac --help
```

Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/fdkaac.rb
```diff
@@ -0,0 +1,62 @@
...
+    frequency = 440.0
+    max_amplitude = 0.2
+
+    position_in_period = 0.0
+    position_in_period_delta = frequency / sample_rate
+
+    samples = [].fill(0.0, 0, num_samples)
+
+    num_samples.times do |i|
+      samples[i] = Math::sin(position_in_period * two_pi) * max_amplitude
```
bfontaine: Use a dot here instead of `::`.

Library/Formula/fdkaac.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      position_in_period += position_in_period_delta
+
+      if(position_in_period >= 1.0)
```
bfontaine: You can drop the parentheses here.

Library/Formula/fdkaac.rb
```diff
@@ -0,0 +1,62 @@
...
+        position_in_period -= 1.0
+      end
+    end
+
+    samples.map! do |sample|
+      (sample * 32767.0).round
+    end
+
+    File.open("#{testpath}/tone.pcm", "wb") {
```
bfontaine: Use `do`/`end` for multi-line blocks

DomT4:
I'm not sure here that it's not deeply confusing to have a `fdkaac` formula that depends on `fdk-aac`. It'd be very easy to mistype that and end up with an unexpected package.

mistydemeo:
I agree, I think the name is really confusing. I came to this issue thinking, "don't we already have a formula for this?"

jrnewell:
I had the same hesitation about the naming as well, but I was trying to follow formula cookbook which says "Name the formula like the project markets the product.".  IMO, a better name for this would be something like ``fdk-aac-encoder``

As to relevancy, there are formulae for ``vorbis-tools`` and ``opus-tools`` which basically provide the same functionality as this - a CLI encoder (plus a few other tools) wrapping ``libvoribis`` and ``opus`` respectively.

DomT4:
We bend the usual naming if two formulae sound *too* similar (The two `aescrypt` formula for example are called the same thing upstream but we named one differently).

> fdk-aac-encoder

This would be fine with me at least.

Library/Formula/fdk-aac-encoder.rb
```diff
@@ -0,0 +1,62 @@
...
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}",
+                          "--mandir=#{man}"
+    system "make", "install"
+  end
+
+  def generate_tone_pcm
+    ohai "generating test audio file"
```
DomT4: Let's not use `ohai` here. It's not likely to take long enough to merit needing to bug the user about.

Library/Formula/fdk-aac-encoder.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}",
+                          "--mandir=#{man}"
+    system "make", "install"
+  end
+
+  def generate_tone_pcm
```
DomT4: Let's inline this into the test rather than using a `def` block, Thanks!

DomT4:
@BrewTestBot test this please

DomT4:
@BrewTestBot test this please

DomT4:
Nobody raised an objection here, so merged in https://github.com/Homebrew/homebrew/commit/9aa419dfa664a815097d32cde168a68b30cd7629. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a nit, but new formulae should have their commit messages styled `fdk-aac-encoder 0.6.2 (new formula)` or similar, Thanks!

PR #40862, opened 2015-06-18
wine: updated resources -gecko 2.36 & -mono 4.5.6

Arthur2e5:
In order that non-stable versions would be able to use the resources, I moved the resources to some outside block.

By the way, is it possible to specify a resource with `:optional`?
For wine, some users may want to download the extra packages manually when needed (wine will pop a window up to ask the user to install one).
If there is a possible syntax for such thing, I can add it in later pushes.

A link to the sourceforge files list of the wine project is also added so it would be easier to check for new versions.

### DIFF
--- a/Library/Formula/wine.rb
+++ b/Library/Formula/wine.rb
@@ -4,6 +4,7 @@ require 'formula'
 # too:
 #  - http://wiki.winehq.org/Gecko
 #  - http://wiki.winehq.org/Mono
+#  - http://sourceforge.net/projects/wine/files
 class Wine < Formula
   desc "Wine Is Not an Emulator"
   homepage 'https://www.winehq.org/'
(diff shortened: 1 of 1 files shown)


DomT4:
> By the way, is it possible to specify a resource with :optional?

Effectively, yes. Stage them in the main block and then in `def install` do something like:

```ruby
def install
  if build.with? "magic"
    resource("magic").stage do
      etc
    end
  end
```

jacknagel:
These resources are intentionally pinned to these versions inside the stable block, for compatibility with the stable version of wine (1.6.2).

If you look further down in the file, you will see the most recent versions: https://github.com/Homebrew/homebrew/blob/8897f7a61434c67372801c43fd40aa48c1a0afc3/Library/Formula/wine.rb#L73-L81

These are what is used for `--devel` and `--head`.

PR #40867, opened 2015-06-18
Changed Docker tag to 1.7.0

mijndert:
Has to wait for Boot2Docker before getting merged.

### DIFF
--- a/Library/Formula/docker.rb
+++ b/Library/Formula/docker.rb
@@ -4,7 +4,7 @@ class Docker < Formula
   # Boot2docker and docker are generally updated at the same time.
   # Please update the version of boot2docker too
   url "https://github.com/docker/docker.git", :tag => "v1.7.0",
-    :revision => "7c8fca2ddb58c8d2c4fb4df31c242886df7dd257"
+    :revision => "0baf60984522744eed290348f33f396c046b2f3a"
 
   bottle do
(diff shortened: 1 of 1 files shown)


DomT4:
The `revision` (commit) needs updating as well here.

ahmetalpbalkan:
@mijndert why wait for boot2docker? It's totally independent of b2d.

DomT4:
^ https://github.com/homebrew/homebrew/blob/master/Library/Formula/docker.rb#L4-L5

ahmetalpbalkan:
alright but it's not a dependency?

it's just telling maintainer to not to forget to update boot2docker as well.

DomT4:
Users generally expect `docker` and `boot2docker` to be kept in-check with each other - Differing versions can be incompatible - We try to respect that expectation. Whilst not a formal dependency it is popular enough and requested enough that we tie the two.

mijndert:
I updated the revision. Sorry about that.

dunn:
#40889 has also been opened, FYI

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/2b84adc1557bca5326344ba5a6a3fee683ab0e7f. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #40868, opened 2015-06-18
scm-manager 1.46

sdorra:


DomT4:
```
==> brew audit scm-manager						 FAILED
Error: 1 problem in 1 formula
scm-manager:
 * Use `depends_on :java` to set JAVA_HOME
```

If you could bump the updated checksums to `sha256` instead of `sha1` at the same time that would be awesome, Thanks!

Library/Formula/scm-manager.rb
```diff
@@ -1,9 +1,9 @@
 class ScmManager < Formula
   desc "Manage Git, Mercurial, and Subversion repos over HTTP"
   homepage "http://www.scm-manager.org"
```
DomT4: This can be `https://www.scm-manager.org/` instead.

Library/Formula/scm-manager.rb
```diff
@@ -1,9 +1,9 @@
...
-  url "http://maven.scm-manager.org/nexus/content/repositories/releases/sonia/scm/scm-server/1.45/scm-server-1.45-app.tar.gz"
-  version "1.45"
-  sha1 "c8a2e9c694525a570460639bbb5cb01b23595ec4"
+  url "http://maven.scm-manager.org/nexus/content/repositories/releases/sonia/scm/scm-server/1.46/scm-server-1.46-app.tar.gz"
```
DomT4: These can be `https://` rather than `http://` as well.

Library/Formula/scm-manager.rb
```diff
@@ -12,9 +12,9 @@ class ScmManager < Formula
   end
 
   resource "client" do
-    url "http://maven.scm-manager.org/nexus/content/repositories/releases/sonia/scm/clients/scm-cli-client/1.45/scm-cli-client-1.45-jar-with-dependencies.jar"
-    version "1.45"
-    sha1 "debfe596c1d959f7c1c282c747f728818e62b321"
+    url "http://maven.scm-manager.org/nexus/content/repositories/releases/sonia/scm/clients/scm-cli-client/1.46/scm-cli-client-1.46-jar-with-dependencies.jar"
```
DomT4: These can be `https://` rather than `http://` as well.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/1875ca14ff849e317806e5ca29f55ee2d6142880. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #40871, opened 2015-06-18
pango 1.37.0

makigumo:


makigumo:
My bad, didn't realize that this is a development release.

PR #40878, opened 2015-06-19
telegraf 0.1.1

toddboom:
Telegraf is a system-level agent for reporting metrics directly into InfluxDB. This is a new recipe, but it is maintained and supported by the team behind InfluxDB.

### DIFF
--- a/Library/Formula/telegraf.rb
+++ b/Library/Formula/telegraf.rb
@@ -1,0 +1,146 @@
+require "language/go"
+
+class Telegraf < Formula
+  desc "Server-level agent for InfluxDB"
+  homepage "https://influxdb.com"
+  url "https://github.com/influxdb/telegraf/archive/v0.1.1.tar.gz"
+  sha256 "75b6e22b0c8d385ca28fe82dfb730d60cc3c8522fe705828739ebedba89bc8b6"
+
+  depends_on "go" => :build
+
+  go_resource "github.com/influxdb/influxdb" do
+    url "https://github.com/influxdb/influxdb.git",
+      :revision => "9f45a9eea3017e372711d3ae5949056d53a4ba6a"
+  end
+
+  go_resource "github.com/naoina/toml" do
+    url "https://github.com/naoina/toml.git",
+      :revision => "5667c316ee9576e9d5bca793ce4ec813a88ce7d3"
+  end
+
+  go_resource "github.com/naoina/go-stringutil" do
+    url "https://github.com/naoina/go-stringutil.git",
+      :revision => "360db0db4b01d34e12a2ec042c09e7d37fece761"
+  end
+
+  go_resource "github.com/lib/pq" do
+    url "https://github.com/lib/pq.git",
+      :revision => "a8d8d01c4f91602f876bf5aa210274e8203a6b45"
+  end
+
+  go_resource "github.com/go-sql-driver/mysql" do
+    url "https://github.com/go-sql-driver/mysql.git",
+      :revision => "66b7d5c4956096efd4c945494d64ad73f1d9ec39"
+  end
+
+  go_resource "github.com/fsouza/go-dockerclient" do
+    url "https://github.com/fsouza/go-dockerclient.git",
+      :revision => "af4d4292dfcb9ebc8052c5d63bdaaa2343801f20"
+  end
+
+  go_resource "github.com/stretchr/testify" do
+    url "https://github.com/stretchr/testify.git",
+      :revision => "ddcad49ec6b8f31bc3daf3a1fbea7eac58d61ff0"
+  end
+
+  go_resource "golang.org/x/crypto" do
+    url "https://github.com/golang/crypto.git",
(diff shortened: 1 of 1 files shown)


Library/Formula/telegraf.rb
```diff
@@ -0,0 +1,141 @@
+require "formula"
```
DomT4: This is no longer required.

Library/Formula/telegraf.rb
```diff
@@ -0,0 +1,141 @@
...
+require "language/go"
+
+class Telegraf < Formula
+  desc "Server-level agent for InfluxDB"
+  homepage "https://influxdb.com"
+  url "https://github.com/influxdb/telegraf.git", :using => :git, :tag => "v0.1.1"
```
DomT4: `:using => :git` is redundant with a `.git` url. We should be using a tarball here unless there are submodules involved. If there are submodules, please use a `:tag` and a `:revision`.

DomT4:
Other audit nits:

```
==> brew style telegraf
== /usr/local/Library/Formula/telegraf.rb ==
C: 77:  3: Use 2 (not 4) spaces for indentation.
C: 82:  3: Use 2 (not 4) spaces for indentation.
C: 87:  3: Use 2 (not 4) spaces for indentation.

1 file inspected, 3 offenses detected
Error: 4 problems in 1 formula

==> audit problems
telegraf:
```

toddboom:
@DomT4 thanks for the feedback - i updated the commit. let me know if it needs any other tweaks!

Library/Formula/telegraf.rb
```diff
@@ -0,0 +1,145 @@
+require "language/go"
+
+class Telegraf < Formula
+  desc "Server-level agent for InfluxDB"
+  homepage "https://influxdb.com"
+  url "https://github.com/influxdb/telegraf/archive/v0.1.1.tar.gz"
+  sha1 "86a4d00c9f8d34be3ac5b3a5a2db8eae5f3d8a14"
```
DomT4: Should be `sha256`. `brew audit --strict` will warn you about that, which is what gets ran against new submissions.

Library/Formula/telegraf.rb
```diff
@@ -0,0 +1,145 @@
...
+    telegraf_path = buildpath/"src/github.com/influxdb/telegraf"
+    telegraf_path.install Dir["*"]
+
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    cd telegraf_path do
+      system "go", "build", "-ldflags", "-X main.Version 0.1.1 -X main.Commit 39c90dd879ce51da60102e9f5694474933607c8f", "cmd/telegraf/telegraf.go"
+    end
+
+    bin.install telegraf_path/"telegraf" => "telegraf"
```
DomT4: This can just be `bin.install telegraf_path/"telegraf"`. You only need the `=> ""` when renaming.

Library/Formula/telegraf.rb
toddboom: Just saw that in the build. Fixed and re-pushed.

Library/Formula/telegraf.rb
toddboom: Good catch, thanks!

toddboom:
@DomT4 alright - finally got a green build! need anything else to get this merged?

Library/Formula/telegraf.rb
```diff
@@ -0,0 +1,145 @@
...
+        <string>#{var}/log/telegraf.log</string>
+        <key>StandardOutPath</key>
+        <string>#{var}/log/telegraf.log</string>
+      </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    assert shell_output("telegraf -version").include?("InfluxDB Telegraf")
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikemcquaid: Also, using `assert_match` produces slightly nicer output than `include?` on failures.

toddboom:
@mikemcquaid ok, i switched to `assert_match` and added a call to our internal `-test`

toddboom:
@mikemcquaid does that look good enough for testing this time around?

pauldix:
+1

toddboom:
@mikemcquaid @DomT4 thanks for all the feedback - anything else i can do to get this tightened up before merging?

DomT4:
:rocket: in https://github.com/Homebrew/homebrew/commit/b1b5390d752cf47eafc7b12b1d2b48a77a454cd9. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a tiny nit for future: new formula submissions should be tagged with `(new formula)` in the commit message, so `telegraf 0.1.1 (new formula)` in this case.

PR #40880, opened 2015-06-19
h2o 1.3.1

DomT4:


Library/Formula/h2o.rb
```diff
@@ -22,10 +22,9 @@ class H2o < Formula
 
   def install
     args = std_cmake_args
-    args << "."
     args << "-DWITH_BUNDLED_SSL=OFF"
 
-    system "cmake", *args
+    system "cmake", ".", *args
```
mikemcquaid: Is the `.` definitely necessary? Think `std_cmake_args` may include it? Ignore me and :+1: if I'm wrong.

DomT4: We don't include it in the standard arguments:

```ruby
  def std_cmake_args
    %W[
      -DCMAKE_C_FLAGS_RELEASE=
      -DCMAKE_CXX_FLAGS_RELEASE=
      -DCMAKE_INSTALL_PREFIX=#{prefix}
      -DCMAKE_BUILD_TYPE=Release
      -DCMAKE_FIND_FRAMEWORK=LAST
      -DCMAKE_VERBOSE_MAKEFILE=ON
      -Wno-dev
    ]
  end
```

There are certain things that need `".."` rather than `"."` I believe, which is probably the justification for that. It isn't actually necessary though, no, since the `.` is basically just saying "Compile in the present directory" which is CMake's default behaviour anyway. You only technically need the dot-passing if you're in a `build` directory or such.

mikemcquaid: :+1:

DomT4: Got rid of it in a new commit. No point keeping around redundant code. It would have bothered me if I'd left it in :smile_cat:

PR #40888, opened 2015-06-19
gtk+3: adwaita-icon-theme added as resource

tschoonj:
@mikemcquaid 

Per the discussion in Homebrew/homebrew-x11#89
Adwaita icon theme (which replaces the defunct gnome-icon-theme)
is now added as a resource to the gtk+3 formula (in an opt-out manner)
in order to provide users with the default icon theme, which is no
longer shipped along with gtk+3.

The only problem is that this will not link when `gnome-icon-theme` is already installed. Not sure what we can do about that.

The adwaita-icon-theme alias and the gnome-icon-theme formula have been
deleted.
The blacklist has been updated with messages for uses trying to install
gnome-icon-theme or adwaita-icon-theme
The `homebank` formula, which depends on gnome-icon-theme, has been stripped of this dependency


tschoonj:
Once this one gets merged in, I will update the graph-tool formula in science, since it currently depends on gnome-icon-theme.

Library/Formula/gtk+3.rb
```diff
@@ -11,6 +12,8 @@ class Gtkx3 < Formula
   end
 
   option :universal
+  option "with-quartz-relocation", "Build with quartz relocation support"
```
DomT4: What's the impact of moving this from mandatory to optional?

Library/Formula/gtk+3.rb
```diff
@@ -22,6 +25,19 @@ class Gtkx3 < Formula
...
+    depends_on "gettext" => :build
+    depends_on "icon-naming-utils" => :build
+    depends_on "intltool" => :build
+    depends_on "librsvg"
+  end
+
+  resource "adwaita-icon-theme" do
+    url "https://download.gnome.org/sources/adwaita-icon-theme/3.16/adwaita-icon-theme-3.16.2.1.tar.xz"
+    sha256 "b4556dfbf555d4fac12d4d5c12f7519de0d43ec42a1b649611439a50bf7acb96"
+  end if build.with?("icons")
```
DomT4: You don't need to protectively wrap resources. They won't be used unless invoked during `def install`.

Library/Formula/gtk+3.rb
```diff
@@ -33,17 +49,22 @@ def install
...
     system "make", "install"
     # Prevent a conflict between this and Gtk+2
     mv bin/"gtk-update-icon-cache", bin/"gtk3-update-icon-cache"
+
+    resource("adwaita-icon-theme").stage do
+      system "./configure", "--disable-dependency-tracking",
+                            "--prefix=#{prefix}",
+                            "GTK_UPDATE_ICON_CACHE=#{bin}/gtk3-update-icon-cache"
+      system "make", "install"
+    end if build.with?("icons")
```
DomT4: It's a nit but we're slowly killing off this style. Recommend

```ruby
if build.with? "icons"
  resource("adwaita-icon-theme").stage do
    etc
    etc
  end
end
```

Instead

Library/Homebrew/blacklist.rb
```diff
@@ -89,5 +89,10 @@ def blacklisted? name
     A binary installer is available:
       https://www.haskell.org/platform/mac.html
     EOS
+  when 'gnome-icon-theme', 'adwaita-icon-theme' then <<-EOS.undent
+    The Adwaita icon theme (which replaces the now defunct gnome-icon-theme)
```
DomT4: I think there's an Alias available for `adwaita-icon-theme` which would need deleting as well.

Library/Formula/gtk+3.rb
tschoonj: I never understood why this option was on by default in the formula. It only makes sense if you intend to put gtk and its dependencies in an app bundle. Otherwise it could actually break a couple of things, especially if HOMEBREW_PREFIX is not /usr/local.

tschoonj:
Ok I will improve the formula a bit according to your instructions. The adwaita-icon-theme alias already gets deleted with this commit.

DomT4:
:+1:, Didn't spot that, Apologies.

If you pull this into a new branch and then `brew upgrade` do you hit any linking errors, since we don't actually remove the existing installation from the user's device?

tschoonj:
As expected, when `gnome-icon-theme` is installed at the time of installing this updated `gtk+3` formula, I get the expected message:

````
Could not symlink share/icons/Adwaita/16x16/actions/address-book-new.png
Target /usr/local/share/icons/Adwaita/16x16/actions/address-book-new.png
is a symlink belonging to gnome-icon-theme. You can unlink it:
  brew unlink gnome-icon-theme

  brew link --overwrite gtk+3

  brew link --overwrite --dry-run gtk+3

Followed by a couple thousand of conflicting files.
````

How about adding a caveat message informing people that if they see this link error message they uninstall `gnome-icon-theme` and relink `gtk+3`?

**Comment by DomT4** [comment]
> Followed by a couple thousand of conflicting files.

Heh.

Typically in this situation we `keg_only` and `revision` the now-conflicting formulae in advance, and then merge the change a few days later. That may not work well in this situation though as it'll break everyone's GTK+3 applications temporarily.

Perhaps something like:
```ruby
def install
if File.exist?(HOMEBREW_PREFIX/"opt/gnome-icon-theme")
  raise "Please `brew rm gnome-icon-theme` - It has been merged with GTK+3"
end
```

tschoonj:
ok, added that last bit and pushed for a rebuild

tdsmith:
If conflicts_with prints the conflict reason (does it?), that would be nicer than raising.

I wonder if it makes sense to point gnome-icon-theme at an empty tarball to help with the transition; it's semantically bad but avoids breaking upgrade.

DomT4:
> If conflicts_with prints the conflict reason (does it?),

It would, but we wouldn't be able to remove `gnome-icon-theme` in that situation? You can't conflict with something the core doesn't know absolutely exists I believe?

tdsmith:
Right, sorry.

DomT4:
Empty tarballing the package isn't a bad idea though - It'd be similar to how Debian uses transitional packages. We could even use a `caveat` to flag that the package had been superseded. Could also be a way to handle formulae renamed longer-term, potentially.

tschoonj:
So exactly what needs still to be done? Just keep the gnome-icon-theme formula alive but revbump it while making it point to an empty tarball? Do you have one of those online already I could use?

The caveat in gnome-icon-theme will not be necessary I think since I have already blacklisted it.

Library/Formula/gtk+3.rb
```diff
@@ -22,9 +25,26 @@ class Gtkx3 < Formula
   depends_on "pango"
   depends_on "glib"
 
+  if build.with?("icons")
+    # Adwaita icon theme dependencies
+    depends_on "gettext" => :build
+    depends_on "icon-naming-utils" => :build
+    depends_on "intltool" => :build
+    depends_on "librsvg"
```
DomT4: There's potentially a round dependency here. `librsvg` has:

```
depends_on "gtk+3" => :optional
```

tschoonj: good catch

I had a look at the librsvg source code and apparently if gtk+3 is found at configure time then a simple svg viewer called `rsvg-view-3` will get built and installed.

what's the best course of action here?

DomT4: I'm not quite sure of the best way to handle this. `gtk+3` is a relatively popular option in `librsvg` from what I've seen, so it'd probably upset people if we killed it off. I guess we could do something like:

```
depends_on "librsvg" => "without-gtk+3"
```

Or such, if that'd work.

CC @Homebrew/owners for viewpoints on what do about the circular dep potential here.

tschoonj:
If it's too complicated too include this a resource, we might just consider adding 'gnome-icon-theme' as a dependency to formulas whenever it's needed. Nowadays most gtk3 apps don't even have icons anymore anyway after it became official gtk policy to strip them from regular buttons.

mikemcquaid:
> If it's too complicated too include this a resource, we might just consider adding 'gnome-icon-theme' as a dependency to formulas whenever it's needed. Nowadays most gtk3 apps don't even have icons anymore anyway after it became official gtk policy to strip them from regular buttons.

That sounds like a good idea :+1:

tschoonj:
I am closing this one then, but I will file a PR  for easy-tag to have the `gnome-icon-theme` dependency, as well as a PR to fix the quartz-relocation option for `gtk+3`

PR #40890, opened 2015-06-19
add libiodbc formula, based on work by eugenejen.

maparent:
Good day! This is based on an earlier pull request by eugenejen: http://github.com/Homebrew/homebrew/pull/20792
I have updated to the latest iodbc version.

### DIFF
--- a/Library/Formula/libiodbc.rb
+++ b/Library/Formula/libiodbc.rb
@@ -8,7 +8,7 @@ class Libiodbc < Formula
   depends_on "autoconf" => :build
   depends_on "libtool" => :build
 
-  keg_only :provided_pre_mountain_lion
+  keg_only :provided_pre_mavericks
 
   def install
     system "sh",  "./bootstrap.sh"
--- a/Library/Formula/libiodbc.rb
+++ b/Library/Formula/libiodbc.rb
@@ -1,14 +1,13 @@
 class Libiodbc < Formula
   homepage "http://www.iodbc.org/dataspace/iodbc/wiki/iODBC/"
   url "http://sourceforge.net/projects/iodbc/files/iodbc/3.52.10/libiodbc-3.52.10.tar.gz/download"
-  sha1 "2a8328cea84a3ddb78b10a009652cdd3f1d39fb9"
+  sha256 "1568b42b4e97f36110af661d39bfea7d94ac4ff020014574b16a7199f068e11f"
 
   depends_on "automake" => :build
   depends_on "autoconf" => :build
   depends_on "libtool" => :build
 
   def install
-    # run bootstrap.sh
     system "sh",  "./bootstrap.sh"
     system "./configure", "--prefix=#{prefix}"
     system "make", "install"
--- a/Library/Formula/libiodbc.rb
+++ b/Library/Formula/libiodbc.rb
@@ -1,7 +1,7 @@
 class Libiodbc < Formula
-  desc "Independent Open DataBase Connectivity from OpenLink Software. Alternative to unixodbc."
+  desc "Database connectivity layer based on ODBC. (alternative to unixodbc)"
   homepage "http://www.iodbc.org/dataspace/iodbc/wiki/iODBC/"
-  url "http://sourceforge.net/projects/iodbc/files/iodbc/3.52.10/libiodbc-3.52.10.tar.gz/download"
+  url "https://downloads.sourceforge.net/project/iodbc/iodbc/3.52.10/libiodbc-3.52.10.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/libiodbc.rb
```diff
@@ -0,0 +1,22 @@
+require 'formula'
```
bfontaine: Please remove this line.

Library/Formula/libiodbc.rb
```diff
@@ -0,0 +1,22 @@
...
+
+class Libiodbc < Formula
+  homepage 'http://www.iodbc.org/dataspace/iodbc/wiki/iODBC/'
```
bfontaine: Use double-quoted strings everywhere.

Library/Formula/libiodbc.rb
```diff
@@ -0,0 +1,22 @@
...
+  url 'http://sourceforge.net/projects/iodbc/files/iodbc/3.52.10/libiodbc-3.52.10.tar.gz/download'
+  sha1 '2a8328cea84a3ddb78b10a009652cdd3f1d39fb9'
```
bfontaine: Use a `sha256` here.

Library/Formula/libiodbc.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  depends_on 'automake' => :build
+  depends_on 'autoconf' => :build
+  depends_on 'libtool' => :build
+
+  def install
+    # run bootstrap.sh
```
bfontaine: You can remove this comment.

Library/Formula/libiodbc.rb
bfontaine: Also, could you add a short description above the homepage using `desc "…"`? Thanks!

maparent:
Thanks a lot for the comments; I think they're all addressed now.

DomT4:
Few nits left:

```
==> brew style libiodbc

Error: 3 problems in 1 formula

==> audit problems
libiodbc:
 * Stable: Don't use /download in SourceForge urls (url is http://sourceforge.net/projects/iodbc/files/iodbc/3.52.10/libiodbc-3.52.10.tar.gz/download).
 * Stable: Use https://downloads.sourceforge.net to get geolocation (url is http://sourceforge.net/projects/iodbc/files/iodbc/3.52.10/libiodbc-3.52.10.tar.gz/download).
 * Description is too long. "name: desc" should be less than 80 characters (currently 97).
```

maparent:
The fail mentions that these files shadow system header files. Those header files disappeared in OSX 10.9, because Apple deprecated iodbc from that version. Is the test system earlier than 10.9?

DomT4:
We test the most recent 3 stable major OS X Versions, so for now 10.8, 10.9 and 10.10. Maybe worth adding something like `keg_only :provided_pre_mountain_lion` so this doesn't cause problems on older systems, although that won't work for ML itself.

maparent:
added the keg_only tag, as requested. Still fails on ML of course.

DomT4:
We can ignore the ML failure if necessary. Going to ask @mistydemeo to comment on whether kegging this is appropriate/useful for <10.8 - Misty runs Tigerbrew and has better experience of not breaking older OS X versions than I do :wink:.

maparent:
Fair. As I said, the newer (3.52.10) include files would not break linking against the older version of iodbc used by OS X (3.52.6).

DomT4:
@BrewTestBot test this please

maparent:
Hello! I would like to know what happened to this merge? I thought absence of test results would not affect the merge. Or is there some more action required on my part?
Thank you,
Marc-Antoine

DomT4:
@BrewTestBot test this please

DomT4:
@mikemcquaid How much do we care about this clashing with 10.8 system headers? We made it `keg_only` on 10.7 and below, but we don't have a `keg_only :provided_pre_mavericks` feature (Although it wouldn't be difficult to add one I guess).

mikemcquaid:
> keg_only :provided_pre_mavericks feature (Although it wouldn't be difficult to add one I guess).

Yeh, would you mind adding that?

mikemcquaid:
At very least I'd like this not to fail the audit.

DomT4:
> Yeh, would you mind adding that?

Sure :+1:

DomT4:
@mikemcquaid #42460

DomT4:
If you update your Homebrew, then rebase this branch against master, you should be able to use `keg_only provided_pre_mavericks` now.

maparent:
Done, thank you!

mikemcquaid:

For future reference the preferred commit message format for new formulae is `libiodbc 3.52.10 (new formula)`.

PR #40896, opened 2015-06-19
gcc: use patch for Xcode 7/El Capitan.

mikemcquaid:
I'm gonna kill this build immediately as it won't run on any of our machines but I think it'd be good to add this patch. CC @Homebrew/owners for any objections. If there are none I'll add it to all the `homebrew-versions` versions where it applies cleanly too.

### DIFF
--- a/Library/Formula/gcc.rb
+++ b/Library/Formula/gcc.rb
@@ -32,6 +32,15 @@ class Gcc < Formula
     sha256 "77bce635f78bc26bd01010b5ece480251af223bf2dba6d48c29af6b29b441296" => :mountain_lion
   end
 
+  if MacOS.version >= :el_capitan
+    # Fixes build with Xcode 7.
+    # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66523
+    patch do
+      url "https://gcc.gnu.org/bugzilla/attachment.cgi?id=35773"
+      sha256 "db4966ade190fff4ed39976be8d13e84839098711713eff1d08920d37a58f5ec"
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


Library/Formula/gcc.rb
```diff
@@ -32,6 +32,15 @@ def osmajor
     sha256 "77bce635f78bc26bd01010b5ece480251af223bf2dba6d48c29af6b29b441296" => :mountain_lion
   end
 
+  if MacOS::Xcode.version > "7.0" || MacOS::CLT.version > "7.0"
```
DomT4: Why not `if MacOS.version => :el_capitan`?

mikemcquaid: Seems to be Xcode version specific; people running Xcode 6 on El Capitan have reported it works (although everything else is broken). I strongly suspect people running Xcode 7 beta on Yosemite will have the same issues.

DomT4: Ah, fair enough :+1:. I forgot the use case of 10.10 users adopting the 7.0 beta.

jacknagel: This line will fail on systems where one of the versions is nil (i.e., one or the other isn't installed).

I think we can just apply the patch unconditionally, though.

jacknagel: I guess Xcode.version can't be nil. I think CLT.version can, though.

mikemcquaid: I'll just apply to `el_capitan`; I'm funny about applying patches when they aren't needed.

PR #40899, opened 2015-06-19
add libwandevent

elnappo:
requirement for libprotoident

PR #40900, opened 2015-06-19
add libprotoident

elnappo:
this pull request has an dependency on https://github.com/Homebrew/homebrew/pull/40899

### DIFF
--- a/Library/Formula/libprotoident.rb
+++ b/Library/Formula/libprotoident.rb
@@ -1,0 +1,29 @@
+class Libprotoident < Formula
+  desc "Libprotoident performs application layer protocol identification"
+  homepage "http://research.wand.net.nz/software/libprotoident.php"
+  url "http://research.wand.net.nz/software/libprotoident/libprotoident-2.0.7.tar.gz"
+  sha256 "5063497274e546b01b0606c8906a292cbe1e2ba8d6f3b6cd25de16a91fef635e"
+
+  depends_on "libflowmanager"
+  depends_on "libwandevent"
+
+  def install
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.cpp").write <<-EOS.undent
+      #include <libprotoident.h>
+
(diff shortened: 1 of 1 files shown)


Library/Formula/libprotoident.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def install
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/lpi_protoident", "-h"
```
bfontaine: Could you please add a test which does something more substantial than e.g. `--help`/`--version`? Thanks!

Library/Formula/libprotoident.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      int main() {
+        lpi_init_library();
+        return 0;
+      }
+    EOS
+    system ENV.cc, "test.cpp", "-L#{lib}", "-lprotoident", "-o", "test"
+    system "./test"
+    system "#{bin}/lpi_protoident", "-h"
+    system "#{bin}/lpi_find_unknown", "-h"
```
DomT4: Let's only use one of these rather than getting the help output from all three. IMO, I'm happy enough with the C test without using the executables at all.

elnappo:
for further testing I need a pcap file for the lpi_* tools

Library/Formula/libprotoident.rb
bfontaine: Yes, the C test is sufficient here.

DomT4:
> this pull request has an dependency on #40899

As a small note for future, in this sort of situation it's perfectly fine to file them in the same PR so our test-bot can look over them at the same time, dependency and all.

elnappo:
is there something missing?

DomT4:
Nope, It's good. Merged in https://github.com/Homebrew/homebrew/commit/fbf779f9190b13f56008866d1076e7cff31e311b. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Please note the preferred commit message for new formula is styled `libprotoident 2.0.7 (new formula)`, Thanks!

PR #40904, opened 2015-06-20
influxdb-0.9.0 (update)

gunnaraasen:
Fixes issues with the meta directory not being created.
Adds a warning for users upgrading from v0.8 that the v0.9 config file is different.
Also adds the option to build from HEAD.

### DIFF
--- a/Library/Formula/influxdb.rb
+++ b/Library/Formula/influxdb.rb
@@ -1,11 +1,13 @@
-require "formula"
 require "language/go"
 
 class Influxdb < Formula
   desc "Time series, events, and metrics database"
   homepage "http://influxdb.com"
-  url "https://github.com/influxdb/influxdb/archive/v0.9.0.tar.gz"
-  sha256 "c7cc869754d7bfb9374b0a16a9b91073d9588adedbd7bfda1d5709e22d3a3d75"
+
+  stable do
+    url "https://github.com/influxdb/influxdb/archive/v0.9.0.tar.gz"
+    sha256 "c7cc869754d7bfb9374b0a16a9b91073d9588adedbd7bfda1d5709e22d3a3d75"
+  end
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Library/Formula/influxdb.rb
```diff
@@ -81,7 +90,12 @@ def install
     Language::Go.stage_deps resources, buildpath/"src"
 
     cd influxdb_path do
-      system "go", "install", "-ldflags", "-X main.version 0.9.0 -X main.commit 471117ce5b308e59f8f247d8a88a52ead553b602", "./..."
+      if build.stable?
+        system "go", "install", "-ldflags", "-X main.version 0.9.0 -X main.commit 471117ce5b308e59f8f247d8a88a52ead553b602", "./..."
+      end
```
DomT4: Let's use an `else` statement here:

```ruby
if build.head?
  something
else
  something_else
end
```

If/When `devel` gets re-added at some point the `else` statement becomes `elsif` and it keeps the blocks a little neater.

Library/Formula/influxdb.rb
```diff
@@ -132,4 +145,13 @@ def plist; <<-EOS.undent
     </plist>
     EOS
   end
+
+  def caveats; <<-EOS.undent
+    Config files from old InfluxDB versions are incompatible with version 0.9.
+    If upgrading from a pre-0.9 version, the new configuration file will be named:
+        #{HOMEBREW_PREFIX}/etc/influxdb.conf.default
```
DomT4: This can just be `#{etc}/influxdb.conf.default`. They'll resolve to the same path.

DomT4:
A week after the 0.9.0 update was merged the ship may have sailed on warning about incompatible config files. Is not being warned about that something you're encountering a lot of noise about upstream?

toddboom:
@DomT4 yeah, it wasn't clear to some users that the old config files weren't overwritten automatically when installing the new package. we've been handling the support tickets, but it would be nice to give a heads up to anyone else that we can.

gunnaraasen:
Recommended changes have been made.

@DomT4 For the warning, as @toddboom mentioned, some people have run into issues with the config being overwritten. With `--HEAD` available, we expect more people will be upgrading. The `(var/"influxdb/meta").mkpath` addition is also an important fix as influxd trips when the directory doesn't exist.

DomT4:
Fair enough on the config issue. If not having the caveat is going to cause more issues for both you and (likely) us I'm happy to include it.

We could use `postinstall` to create an updated config file and remove or move the old one aside, but I don't know if we'll then end up with people upset we changed their config file :octopus:.

gunnaraasen:
I'd rather keep the old config and show the warning. We'll likely remove the caveat In a couple months when we're confident most people have upgraded.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/d8167e35e35ceda9f95635fa929842ea298a05c8. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a tiny note for future, for changes that don't contain version bumps the commit style should be something like `formula: thing changed`, Thanks!

PR #40905, opened 2015-06-20
acpica 20150619

Hs-Yeah:
Update acpica to 20150619

### DIFF
--- a/Library/Formula/acpica.rb
+++ b/Library/Formula/acpica.rb
@@ -3,11 +3,11 @@ class Acpica < Formula
   homepage "https://www.acpica.org/"
   head "https://github.com/acpica/acpica.git"
   url "https://acpica.org/sites/acpica/files/acpica-unix2-20150619.tar_1.gz"
-  sha1 "8c5514b1171afb40dca40289581c8ba3f17583e1"
+  sha1 "283c594c4b89db8e3bef8fca25021c52938987ad"
 
   bottle do
     cellar :any
-    sha1 "283c594c4b89db8e3bef8fca25021c52938987ad" => :yosemite
+    sha1 "dbb1b43a1702b3eb2855dd4df1ffbb81a4e5e3c1" => :yosemite
     sha1 "5c2ded46bf71f187b637fd748d541985493ddf57" => :mavericks
     sha1 "c546de5f0bd9ead75dc58baba0489e0a37eb0f60" => :mountain_lion
   end
--- a/Library/Formula/acpica.rb
+++ b/Library/Formula/acpica.rb
@@ -3,7 +3,7 @@ class Acpica < Formula
   homepage "https://www.acpica.org/"
   head "https://github.com/acpica/acpica.git"
   url "https://acpica.org/sites/acpica/files/acpica-unix2-20150619.tar_1.gz"
-  sha1 "283c594c4b89db8e3bef8fca25021c52938987ad"
+  sha256 "12d68f781c9f5c3dfb61ef280f41f0d019d7ba4e914ef916ed3e1d22712ffd52"
 
   bottle do
     cellar :any
@@ -14,8 +14,8 @@ class Acpica < Formula
 
   def install
     ENV.deparallelize
-    system "make", "HOST=_APPLE", "PREFIX=#{prefix}"
-    system "make", "install", "HOST=_APPLE", "PREFIX=#{prefix}"
+    system "make", "PREFIX=#{prefix}"
+    system "make", "install", "PREFIX=#{prefix}"
   end
 
   test do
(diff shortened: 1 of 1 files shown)


Library/Formula/acpica.rb
```diff
@@ -2,12 +2,12 @@ class Acpica < Formula
...
   homepage "https://www.acpica.org/"
   head "https://github.com/acpica/acpica.git"
-  url "https://acpica.org/sites/acpica/files/acpica-unix2-20150204.tar.gz"
+  url "https://acpica.org/sites/acpica/files/acpica-unix2-20150619.tar_1.gz"
   sha1 "8c5514b1171afb40dca40289581c8ba3f17583e1"
 
   bottle do
     cellar :any
-    sha1 "dbb1b43a1702b3eb2855dd4df1ffbb81a4e5e3c1" => :yosemite
+    sha1 "283c594c4b89db8e3bef8fca25021c52938987ad" => :yosemite
```
DomT4: Please don't update this. We'll update this line automatically when we pull the latest version in. You may have intended to update the `sha1` beneath the `url` line up top, which does need changing :wink:

Hs-Yeah: Sorry for my ignorance, corrected.

Library/Formula/acpica.rb
```diff
@@ -2,8 +2,8 @@ class Acpica < Formula
   desc "OS-independent implementation of the ACPI specification"
   homepage "https://www.acpica.org/"
   head "https://github.com/acpica/acpica.git"
-  url "https://acpica.org/sites/acpica/files/acpica-unix2-20150204.tar.gz"
-  sha1 "8c5514b1171afb40dca40289581c8ba3f17583e1"
+  url "https://acpica.org/sites/acpica/files/acpica-unix2-20150619.tar_1.gz"
+  sha1 "283c594c4b89db8e3bef8fca25021c52938987ad"
```
DomT4: Could you possibly make this `sha256 "12d68f781c9f5c3dfb61ef280f41f0d019d7ba4e914ef916ed3e1d22712ffd52"` instead? Thanks!

DomT4:
The Makefile has changed. Now simply reads:

```make
ifeq ($(OS),efi)
	BUILD_DIRECTORY_PATH = "generate/efi"
	include generate/efi/Makefile.config
	include generate/efi/Makefile.common
else
	BUILD_DIRECTORY_PATH = "generate/unix"
	include generate/unix/Makefile.config
	include generate/unix/Makefile.common
endif
```

So we should just be good to change the current `make` lines to:

```ruby
    ENV.deparallelize
    system "make", "PREFIX=#{prefix}"
    system "make", "install", "PREFIX=#{prefix}"
```

Library/Formula/acpica.rb
DomT4: No need for an apology, It's an easy mistake to make :)

Library/Formula/acpica.rb
```diff
@@ -2,8 +2,8 @@ class Acpica < Formula
   desc "OS-independent implementation of the ACPI specification"
   homepage "https://www.acpica.org/"
   head "https://github.com/acpica/acpica.git"
-  url "https://acpica.org/sites/acpica/files/acpica-unix2-20150204.tar.gz"
-  sha1 "8c5514b1171afb40dca40289581c8ba3f17583e1"
+  url "https://acpica.org/sites/acpica/files/acpica-unix2-20150619.tar_1.gz"
```
DomT4: I've just noticed our detection code can't automatically handle this tarball version. Can you add a manual version, like this:

```ruby
url "https://acpica.org/sites/acpica/files/acpica-unix2-20150619.tar_1.gz"
version "20150619"
sha256 "12d68f781c9f5c3dfb61ef280f41f0d019d7ba4e914ef916ed3e1d22712ffd52"
```

Thanks!

Hs-Yeah:
Shall we use https://acpica.org/sites/acpica/files/acpica-unix2-20150619.tar.gz for building? But it's corresponding to 20150618.

DomT4:
> Shall we use https://acpica.org/sites/acpica/files/acpica-unix2-20150619.tar.gz for building?

:+1:. It turns out the reason why the other tarball wasn't building is that tar isn't correctly unpacking it, so instead of there being the intended contents inside the unpack it was just `acpica-unix2-20150619.tar_1` in that directory.

> But it's corresponding to 20150618.

Where does it say this?

Hs-Yeah:
After unpacking, we can see the content. It's acpica-unix-20150618.

DomT4:
Heh, so it is :smile_cat:. I think that's just a typo - The `changes` document inside the tarball lists `20150619` and the tarball itself is named `20150619`. Let's drop the version and trust that the tarball name and changes document are correct and the folder name is just a typo.

Hs-Yeah:
I made a diff between the contents of these two files, and found out they are exactly the same. It's a typo indeed.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/3fbd0c6b991a5d80301a45dc2d45458871e31c6a. Thank you for your contribution to Homebrew, and your super-responsiveness in this PR, really appreciate both :star2:.

PR #37306, opened 2015-03-02
Update gnome-icon-theme to 3.14.1 and alias adwaita-icon-theme

xanderlent:
In GNOME 3.14, gnome-icon-theme and gnome-icon-theme-symbolic were merged into adwaita-icon-theme, because the symbolic icons are no longer considered optional. (See the mailing list [archives](https://mail.gnome.org/archives/desktop-devel-list/2014-April/msg00086.html).)

Moving to the latest stable version of this package fixes the lack of symbolic icons for GTK+3 in Homebrew. (See images below.)

The tests I've written are simple sanity checks that ensure certain files were created in the install process.
Someone with GTK+3 knowledge could probably write better ones that check if GTK+ properly detects the icon files.

(This PR also makes gnome-icon-theme depend on librsvg, since the symbolic icons are converted svg to png during the `make install` step, and SVGs are subsequently used by GTK+.)

The current state of things:
symbolic-icons-missing
With the changes in this pull request:
symbolic-icons-available

PR #40909, opened 2015-06-20
atlassian-cli 4.3.0 (replaced stash-cli)

guoxiao:


DomT4:
Is this actually a new project, or did the upstream project rename?

If the latter, we don't handle upstream renames this way sadly, since the way our formula code works the user won't automatically get updated from `stash-cli` to `atlassian-cli` even though they're the same thing. It's a core issue that we're aware of and have a desire to fix going forwards, but for now, can't do formula renames.

See #37306 for a recent example of how to handle it instead. (We alias the new name onto the old name).

If it's a new project entirely and the old one isn't going to get updates any more, it should be moved to the `boneyard` instead of just deleted.

guoxiao:
The upstream integrated `stash-cli` and several commands into a `atlassian-cli` package. The `stash` command provided by `stash-cli` now can be provided by `atlassian-cli`. 
So I think I should add a alias to redirect `stash-cli` to `atlassian-cli`?

DomT4:
Sorry - I meant the other way around. The current formula remains, but you update everything aside the name, and then you symlink the new name `atlassian-cli` onto the existing `stash-cli` formula.

guoxiao:
But now `atlassian-cli` is the correct name, and `stash-cli` is only a part of it. So I think aliasing `atlassian-cli` onto `stash-cli` is not appropriate.

DomT4:
Sadly it is the way we do things, otherwise users who already have `stash-cli` installed will never get the update. See #14374 for the list of things we would like to change but can't yet.

guoxiao:
Alright. I've updated the formulas.
We may need a rename when it's possible.

DomT4:
Thanks. Feel free to comment on that issue I highlighted that it needs a rename.

mikemcquaid:
@BrewTestBot test this please

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/7fee2f6ef745d08339234af25b83f3494fac4f07. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #40911, opened 2015-06-20
formula_versions: avoid disk io

xu-cheng:
cc @jacknagel. Maybe there is a better approach.

### DIFF
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -14,11 +14,15 @@ class Formulary
     FORMULAE.fetch(path)
   end
 
-  def self.load_formula(name, path)
+  def self.load_formula_from_path(name, path)
+    path.open("r") { |f| load_formula(name, path, f) }
+  end
+
+  def self.load_formula(name, path, io)
     mod = Module.new
     const_set("FormulaNamespace#{Digest::MD5.hexdigest(path.to_s)}", mod)
-    contents = path.open("r") { |f| set_encoding(f).read }
-    mod.module_eval(contents, path)
+    content = set_encoding(io).read
+    mod.module_eval(content, path.to_s)
     class_name = class_s(name)
 
     begin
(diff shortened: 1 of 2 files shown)


Library/Homebrew/formulary.rb
```diff
@@ -14,10 +14,14 @@ def self.formula_class_get(path)
...
-  def self.load_formula(name, path)
+  def self.load_formula(name, path, io=nil)
     mod = Module.new
     const_set("FormulaNamespace#{Digest::MD5.hexdigest(path.to_s)}", mod)
-    contents = path.open("r") { |f| set_encoding(f).read }
+    if io
+      contents = set_encoding(io).read
+    else
+      contents = path.open("r") { |f| set_encoding(f).read }
+    end
```
jacknagel: Since we always know the type we are passing in, I would just make a separate method rather than branching here. For example, `load_formula` could just accept an IO, and then

```rb
def self.load_formula_from_path(name, path)
  path.open("r") { |f| load_formula(name, f) }
end
```

mikemcquaid:
:+1:

xu-cheng:
The code works. However, it kinda bother me as we use path as the cached key and loading from io doesn't have one. Maybe we need to rename the variable to make it more sense or just don't cache formula in this situation.

jacknagel:
We could pass a path in. In the place where this is being used, the original formula path makes sense.

xu-cheng:
>  the original formula path makes sense

Won't this load the original formula which is cached under the key of its path? Actually, I think we don't need to cache formula in `formula_versions` since it's a temporary formula which is only used once.

jacknagel:
I mean we would still load directly from the IO, but pass in the path when instantiating the formula. We can do this if we move the caching to load_formula_from_path so that we don't cache when loading directly from the IO.

xu-cheng:
Closed in favor of #43506.

PR #40914, opened 2015-06-20
README: add PGP key

xu-cheng:
cc @Homebrew/owners

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,11 @@ Second, read the [Troubleshooting Checklist](https://github.com/Homebrew/homebre
 ## Security
 Please report security issues to security@brew.sh.
 
+This is our PGP key which is valid until June 17, 2016.
+* Key ID: `CE59E297`
+* Fingeprint: `C657 8F76 2E23 441E C879  EC5C E33A 3D3C CE59 E297`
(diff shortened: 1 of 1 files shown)


DomT4:
We should probably sign the key with our GPG keys as well for verification, and upload it to a public traditional keyserver. But it's good to :ship:

xu-cheng:
:ship:ed. Also I have upload it to MIT PGP Server. As for signing the key , can you do this @DomT4? I'm not quite familiar on that.

DomT4:
Essentially:

* `gpg2 --sign-key 0xE33A3D3CCE59E297`
* `gpg2 --send-key 0xE33A3D3CCE59E297`

Then give it about 5 minutes and do:

* `gpg2 --recv-keys 0xe33a3d3cce59e297`

And then `gpg2 --list-sigs 0xe33a3d3cce59e297` will confirm your signature underneath it. And then 20 minutes or so later once the Keyservers have bounced each other around a bit you can go to https://pgp.mit.edu/pks/lookup?op=vindex&search=0xE33A3D3CCE59E297 and your signature will appear there as well.

xu-cheng:
Thanks @DomT4

PR #40920, opened 2015-06-20
download_strategy: remove extra colon

xu-cheng:
It's kinda wierd for me to have a colon on `Download from` but without one for `Download`.
This change makes two messages have the unified form.


DomT4:
:+1:

PR #40923, opened 2015-06-20
docker-compose 1.3.0

alex:


Library/Formula/fig.rb
```diff
@@ -83,6 +83,6 @@ def install
 
   test do
     output = shell_output(bin/"docker-compose --version")
-    assert output.include? "compose 1.2.0"
+    assert output.include? "compose 1.3.0"
```
dunn: I think these two lines can be changed to just `assert_match version.to_s shell_output(bin/"docker-compose --version")`

DomT4: :+1:, or `assert_match /#{version}/, shell_output(bin/"docker-compose --version")` - There's no strong preference either way.

Zelnox:
Why is the formula still named fig? Will it be renamed eventually?

DomT4:
We don't have the ability to rename formulae yet - When we do it breaks updates on that formula for everyone who currently has it installed - So for example if you install `dog 1.7.8` and then someone renames and releases `cat 1.7.9` and we adopt that rename the person who has `dog 1.7.8` installed will never get the update, or be aware there's an update pending under a new name. It's a known issue dating back to #14374.

michaellopez:
@alex The failed checks are due to the version string having changed in docker-compose 1.3.0. It now looks like this:

```
$ docker-compose --version
docker-compose version: 1.3.0
CPython version: 2.7.9
OpenSSL version: OpenSSL 1.0.1j 15 Oct 2014
```

Following the suggestions from @dunn and @DomT4 will most likely fix the failed checks and we can all get the latest docker-compose via Homebrew. Please update the PR @alex.

michaellopez:
The build now passes but the Jenkins instance seems to have a bit of trouble talking to Github? See http://bot.brew.sh/job/Homebrew%20Pull%20Requests/28314/console

```
Setting status of 5bcbe836cb42e7f53296d9a58fa41f992cd5fbfe to SUCCESS with url http://bot.brew.sh/job/Homebrew%20Pull%20Requests/28312/ and message: 'Merged build finished. No test results found.'
FileNotFoundException means that the credentials Jenkins is using is probably wrong. Or the user account does not have write access to the repo.
java.io.FileNotFoundException: {"message":"Not Found","documentation_url":"https://developer.github.com/v3"}
...
```

Any insights on this @DomT4 @dunn ?

DomT4:
Our test-bot seems to have transitioned into its moody teenager phase. I've flagged it up to the rest of the maintainers and we'll likely re-queue builds once the issue has been flushed.

xu-cheng:

PR #32161, opened 2014-09-08
cv 0.5.1

BestPig:
Adding cv (Coreutils Viewer) in version 0.4.1.

PR #40925, opened 2015-06-20
cv 0.7.1

guoxiao:
moving from `homebrew/homebrew-headonly`

### DIFF
--- a/Library/Formula/cv.rb
+++ b/Library/Formula/cv.rb
@@ -1,0 +1,25 @@
+class Cv < Formula
+  desc "Coreutils Viewer: Show progress for coreutils"
+  homepage "https://github.com/Xfennec/cv"
+  url "https://github.com/Xfennec/cv/archive/v0.7.1.tar.gz"
+  sha256 "c8ab81b09f6026cbfdc94c9453d4a0fad7ea5e1e34efd1c1559f88f7398cf4ee"
+  head "https://github.com/Xfennec/cv.git"
+
+  def install
+    system "make", "PREFIX=#{prefix}", "install"
+  end
+
+  test do
+    system "cv", "--help"
+    pid = fork do
+      system "/bin/dd", "if=/dev/zero", "of=/dev/null", "bs=100000", "count=1000000"
+    end
+    sleep 1
+    begin
+      assert_match /dd/, shell_output("cv")
+    ensure
+      Process.kill 9, pid
+      Process.wait pid
+    end
+  end
+end


Library/Formula/cv.rb
```diff
@@ -0,0 +1,15 @@
+class Cv < Formula
+  homepage "https://github.com/BestPig/cv"
+  desc "Coreutils Viewer: Show progress for cp, rm, dd..."
```
bfontaine: Please move this line above the homepage.

bfontaine:
Please make a PR to [Homebrew/homebrew-headonly](https://github.com/Homebrew/homebrew-head-only) to remove the formula there.

Library/Formula/cv.rb
```diff
@@ -0,0 +1,15 @@
+class Cv < Formula
+  desc "Coreutils Viewer: Show progress for cp, rm, dd..."
```
DomT4: Let's not use `...` in this context - It looks like there should be something else printed afterwards in the command-line environment. Perhaps just `Progress meter for coreutils commands` or something?

Library/Formula/cv.rb
```diff
@@ -0,0 +1,15 @@
+class Cv < Formula
+  desc "Coreutils Viewer: Show progress for coreutils"
+  homepage "https://github.com/BestPig/cv"
```
nwinkler: This is still pointing to the fork... see #32161.

Shouldn't this point to the official repo instead? https://github.com/Xfennec/cv

guoxiao: I've notices it. The official repo just merged BestPig's OSX support patch, but it's not included in latest release. So it still uses BestPig's fork.

nwinkler: True - you're right!

mikemcquaid: We won't include this until we can have it pointing to the non-forked version, sorry!

guoxiao: OK I have requested the author to create a new release supporting OSX.
see Xfennec/cv#45

guoxiao:
I've updated the formula to use official release and BestPig's patch.

Library/Formula/cv.rb
```diff
@@ -0,0 +1,20 @@
...
+    url "https://patch-diff.githubusercontent.com/raw/Xfennec/cv/pull/19.diff"
+    sha256 "3b3b702ad6d51131ef235ebd3544709d28d7f0d39a0fabe833f4909202819b76"
+  end
+
+  def install
+    system "make", "PREFIX=#{prefix}", "install"
+  end
+
+  test do
+    system "cv", "--help"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/cv.rb
```diff
@@ -0,0 +1,20 @@
+class Cv < Formula
+  desc "Coreutils Viewer: Show progress for coreutils"
+  homepage "https://github.com/Xfennec/cv"
+  url "https://github.com/Xfennec/cv/archive/v0.6.tar.gz"
+  sha256 "4291ab2f0c507df9eb694bf956ed65dc12bae3c492b795235142b5b87e2e2239"
+  head "https://github.com/Xfennec/cv.git"
+
+  patch do
+    url "https://patch-diff.githubusercontent.com/raw/Xfennec/cv/pull/19.diff"
```
mikemcquaid: Mind making this `https://github.com/Xfennec/cv/pull/19.patch`? It just makes it easier to jump to the actual PR.

Library/Formula/cv.rb
guoxiao: It's difficult to write a test, because cv is a monitor of other programmes(cp, dd, etc).

mikemcquaid: Could you just call one of those other programs with e.g. copying a zero-byte file?

guoxiao: The time of copying a zero-byte file may be too short to be monitored..

mikemcquaid: You could `dd` to `/dev/null`, fork, `cv` and then kill the `dd`.

guoxiao: Good idea. I've updated the test.

Library/Formula/cv.rb
```diff
@@ -0,0 +1,30 @@
...
+  test do
+    system "cv", "--help"
+    pid = fork do
+      system "/bin/dd", "if=/dev/zero", "of=/dev/null", "bs=100000", "count=1000000"
+    end
+    sleep 1
+    begin
+      assert shell_output("cv").include?("dd")
+    ensure
+      Process.kill 9, pid
```
mikemcquaid: Does a regular `kill` not work here?

Library/Formula/cv.rb
```diff
@@ -0,0 +1,30 @@
...
+      Process.wait pid
```
mikemcquaid: Might not be worth waiting here?

Library/Formula/cv.rb
```diff
@@ -0,0 +1,30 @@
...
+  def install
+    system "make", "PREFIX=#{prefix}", "install"
+  end
+
+  test do
+    system "cv", "--help"
+    pid = fork do
+      system "/bin/dd", "if=/dev/zero", "of=/dev/null", "bs=100000", "count=1000000"
+    end
+    sleep 1
```
mikemcquaid: Why is the sleep needed? Not disagreeing, just curious.

Library/Formula/cv.rb
```diff
@@ -0,0 +1,30 @@
...
+    begin
+      assert shell_output("cv").include?("dd")
```
mikemcquaid: `assert_match /dd/, shell_output("cv")`

Library/Formula/cv.rb
guoxiao: Only for fear that `dd` is not ready when `cv` executed. It's not needed and I've removed it.

guoxiao:
The upstream has a new release supporting OS X. 
Updated to release 0.7.1.

Library/Formula/cv.rb
guoxiao: It seems that @BrewTestBot failed after `sleep 1` was removed. I re added it.

mikemcquaid:

For future reference the preferred commit message format for new formulae is `cv 0.7.1 (new formula)`.

PR #40926, opened 2015-06-20
redis-leveldb 1.4

guoxiao:
moving from `homebrew/homebrew-headonly`

### DIFF
--- a/Library/Formula/redis-leveldb.rb
+++ b/Library/Formula/redis-leveldb.rb
@@ -1,0 +1,22 @@
+class RedisLeveldb < Formula
+  desc "Redis-protocol compatible frontend to leveldb"
+  homepage "https://github.com/KDr2/redis-leveldb"
+  url "https://github.com/KDr2/redis-leveldb/archive/v1.4.tar.gz"
+  sha256 "b34365ca5b788c47b116ea8f86a7a409b765440361b6c21a46161a66f631797c"
+  head "https://github.com/KDr2/redis-leveldb.git"
+
+  depends_on "libev"
+  depends_on "gmp"
+  depends_on "leveldb"
+
+  def install
+    inreplace "src/Makefile", "../vendor/libleveldb.a", Formula["leveldb"].opt_lib+"libleveldb.a"
+    ENV.prepend "LDFLAGS", "-lsnappy"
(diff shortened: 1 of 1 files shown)


DomT4:
Please file a PR at `head-only` to remove it there as well, Thanks!

Library/Formula/redis-leveldb.rb
```diff
@@ -0,0 +1,21 @@
...
+  url "https://github.com/KDr2/redis-leveldb/archive/v1.4.tar.gz"
+  sha256 "b34365ca5b788c47b116ea8f86a7a409b765440361b6c21a46161a66f631797c"
+  head "https://github.com/KDr2/redis-leveldb.git"
+
+  depends_on "libev"
+  depends_on "gmp"
+  depends_on "leveldb"
+
+  def install
+    inreplace "src/Makefile", "../vendor/libleveldb.a", Formula["leveldb"].lib+"libleveldb.a"
```
DomT4: Use `opt_lib` here please.

Library/Formula/redis-leveldb.rb
```diff
@@ -0,0 +1,21 @@
...
+    system "make", "LDFLAGS=-lsnappy -lgmp -lev"
```
DomT4: This should be done automatically? (The LDFLAGS part)

guoxiao: The brewed `libleveldb.a` doesn't link with snappy statically. So it's needed to add -lsnappy.
To avoid inreplacing the makefile ugly, I set LDFLAGS here.

DomT4: Alright. Let's try and get rid of the other two though if they aren't necessary.

guoxiao: Because I replaced LDFLAGS set in Makefile, the other two are also necessary...

DomT4: Does `ENV.prepend "LDFLAGS", "-lev"` at the top of `def install` not work?

guoxiao: It works. Thanks.

mikemcquaid:

PR #40929, opened 2015-06-20
pwsafe: fix download URL and more

L2G:
* Replace broken download URL with current URL from homepage
* Add openssl dependency
* Add test using upstream's test db
* Other edits for `brew audit --strict`


DomT4:
Let's add a `revision` if it was previously using the system OpenSSL.

L2G:
I _knew_ I forgot something. :crying_cat_face:

xu-cheng:

PR #40947, opened 2015-06-21
folly 0.47.0

marpaia:
I took this over from #37982 and addressed the comments in the PR.

### DIFF
--- a/Library/Formula/folly.rb
+++ b/Library/Formula/folly.rb
@@ -1,0 +1,46 @@
+class Folly < Formula
+  desc "A collection of reusable C++ library artifacts developed at Facebook"
+  homepage "https://github.com/facebook/folly"
+  url "https://github.com/facebook/folly/archive/v0.47.0.tar.gz"
+  sha256 "7d3309088cae1817d60996611e5fc0825c5c8df36d244f247c1d49ebc38c35fb"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+
+  depends_on "double-conversion"
+  depends_on "glog"
+  depends_on "gflags"
+  depends_on "boost"
+  depends_on "libevent"
+
+  needs :cxx11
+  depends_on :macos => :mavericks
+
+  def install
+    ENV.cxx11
+    cd "folly" do
+      system "autoreconf", "-i"
+      system "./configure", "--prefix=#{prefix}"
+      system "make"
+      system "make", "install"
+    end
+  end
+
+  test do
+    (testpath/"test.cc").write <<-EOS.undent
+      #include <folly/FBVector.h>
+      int main() {
+        folly::fbvector<int> numbers({0, 1, 2, 3});
+        numbers.reserve(10);
+        for (int i = 4; i < 10; i++) {
+          numbers.push_back(i * 2);
+        }
+        assert(numbers[6] == 12);
+        return 0;
+      }
+    EOS
+    system ENV.cxx, "-std=c++11", "test.cc", "-L#{lib}", "-lfolly", "-o", "test"
+    system "./test"
+  end
+end


Library/Formula/folly.rb
```diff
@@ -0,0 +1,42 @@
...
+        folly::fbvector<int> numbers({0, 1, 2, 3});
+        numbers.reserve(10);
+        for (int i = 4; i < 10; i++) {
+          numbers.push_back(i * 2);
+        }
+        assert(numbers[6] == 12);
+        return 0;
+      }
+    EOS
+    system ENV.cxx, "-std=c++11", "test.cc", "-L#{lib}", "-lfolly", "-o", "test"
```
mistydemeo: Does folly itself need C++11 to build?

mistydemeo:
Thanks for picking this up!

Library/Formula/folly.rb
marpaia: yup

mistydemeo:
Instead of a 10.9 requirement, add `needs :cxx11` below the dependencies and `ENV.cxx11` at the top of the install method; that'll allow 10.8's clang to build it. It should also work on older OS X versions with a C++11-capable compiler.

marpaia:
@mistydemeo updated the commit. just `ENV.cxx11`? not `ENV.cxx11 = true` or anything?

mistydemeo:
Yep, that's right. It's a method call that sets environment variables.

marpaia:
ah, that's right, i forgot that you can omit the `()` :weary:

Library/Formula/folly.rb
```diff
@@ -0,0 +1,45 @@
+class Folly < Formula
+  desc "An open-source C++ library developed and used at Facebook."
+  homepage "https://github.com/facebook/folly"
+  url "https://github.com/facebook/folly.git", :revision => "6457c4429379011d12ee841e8325dddeebec538e"
```
guoxiao: Should the url be `https://github.com/facebook/folly/archive/v0.45.0.tar.gz` ?

marpaia: the build doesn't work on OS X, it has since been fixed in master, but a release hasn't been cut

mistydemeo: Can you use the tarball and backport the patches that fix it?

marpaia: i'm going to see if we can just cut a new release which builds on OS X.

marpaia:
@mistydemeo the build still fails on ML. I sincerely doubt that Folly maintainers have any desire whatsoever to support ML, especially given that it doesn't even receive security updates anymore. is it cool if i just require mavericks?

mistydemeo:
Yep, that's fine.

marpaia:
so we cut a new folly release at the commit hash that has previously built fine on your 10.9 and 10.10 jenkins, removed 10.8 support, building from the tar, etc. should be good to go, but i'm not sure why jenkins isn't picking up the build. perhaps i've `git commit --amend` too much?

Library/Formula/folly.rb
```diff
@@ -0,0 +1,47 @@
+class Folly < Formula
+  desc "An open-source C++ library developed and used at Facebook."
+  homepage "https://github.com/facebook/folly"
+  url "https://github.com/facebook/folly/archive/v0.47.0.tar.gz"
+  sha256 "7d3309088cae1817d60996611e5fc0825c5c8df36d244f247c1d49ebc38c35fb"
+  version "0.47.0"
```
mistydemeo: Can remove the explicit version now that you're using a tagged release; it can be inferred from the tarball name.

DomT4:
@BrewTestBot test this please

Library/Formula/folly.rb
```diff
@@ -0,0 +1,46 @@
+class Folly < Formula
+  desc "An open-source C++ library developed and used at Facebook."
```
mikemcquaid: What does it do, though?

mikemcquaid:
Happy to merge this; just need a slightly better description.

Library/Formula/folly.rb
marpaia: That's actually the "official" description of the project. I think the ambiguity is intentional, its supposed to feel like a grab bag of awesome C++ utilities which we use at FB.

mikemcquaid: Could you make it a little less ambiguous? Maybe even just `grab bag of awesome` etc.

marpaia:
@mikemcquaid i changed it based on the second sentence of the release blog post: https://www.facebook.com/notes/facebook-engineering/folly-the-facebook-open-source-library/10150864656793920

i didn't want to get too detailed, as i didn't want the desc line to get too long, but if you'd prefer something else for the description, i'm down for whatever works for you all.

mikemcquaid:
@marpaia It's great, thanks!

marpaia:
omg desc is too long now. updating.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/3ef948bb3484c2af701c8a226550d152d8636f64. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

It's a nit but just for future reference new formulae commit messages should be styled `name 1.2.3 (new formula)`, Thanks!

marpaia:
cheers, thanks!

PR #40953, opened 2015-06-21
gtk-mac-integration: add python bindings

DomT4:


Library/Formula/gtk-mac-integration.rb
```diff
@@ -7,16 +7,18 @@ class GtkMacIntegration < Formula
   depends_on "pkg-config" => :build
   depends_on "gtk+"
   depends_on "gtk+3" => :recommended
+  depends_on "gobject-introspection"
```
DomT4: If you have `gtk+` installed, which this formula requires, you also have `gobject-introspection` installed, so there's no harm in using it.

tschoonj:
How about adding an option to let the user decide if he/she wants to use python2 or python3?

mikemcquaid:
Generally we disable Python bindings by default as they are linked to the specific Python version they were built against and therefore don't work well with binary packages. CC @tdsmith in case this has changed and I am now wrong.

DomT4:
Tim's encouraged me to add bindings by default a few times, so I presumed we were okay about doing so now. Will hold off on this until he comments though.

DomT4:
Tim said as long as there aren't any audit flags, It's fine to merge, so https://github.com/Homebrew/homebrew/commit/a18ede548521121c151988b853bf6548d7567063.

Will look at the Python3 bindings, but wanted to get the Python bindings in as people are more likely to be using GTK projects with Python2.7.x.

PR #40971, opened 2015-06-22
mhash: modernize formula, add basic test

alanthing:
I wish the test could be better, but there is no library.

### DIFF
--- a/Library/Formula/mhash.rb
+++ b/Library/Formula/mhash.rb
@@ -1,14 +1,24 @@
-require 'formula'
-
 class Mhash < Formula
   desc "Uniform interface to a large number of hash algorithms"
-  homepage 'http://mhash.sourceforge.net/'
-  url 'https://downloads.sourceforge.net/project/mhash/mhash/0.9.9.9/mhash-0.9.9.9.tar.gz'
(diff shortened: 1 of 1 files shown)


Library/Formula/mhash.rb
```diff
@@ -1,14 +1,16 @@
...
 
   def install
     system "./configure", "--disable-debug", "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
-    system "make install"
+    system "make", "install"
+  end
+
+  test do
+    system "otool -L #{prefix}/lib/libmhash.2.dylib | grep libSystem"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

alanthing:
Ripped off a test from tinyxml2.rb and rebased.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/7c7bbc47ca61a4de92460db9fb4bd40401c96627. Thanks @alanthing!

PR #40980, opened 2015-06-22
osquery 1.4.7

marpaia:


marpaia:
This is failing on OS X 10.9. The reason why is becaue, after rootpipe, we (facebook) mandated that all OS X be upgraded to or past 10.10.3. We explicitly removed build support for 10.9 after that, as 10.9 is basically EOL'd. I'm going to remove mavericks support from homebrew as we don't explicitly support it as a project anymore.

marpaia:
CC @theopolis

theopolis:
More details, FWIW, for removing support for 10.9: We don't have any 10.9 hosts anymore to dogfood/canary or performance test on, likewise we have no need to support 10.9. Rootpipe is the colloquial reason why. We don't feel comfortable "officially" supporting 10.9 if we cannot make the needed regression and performance promises.

DomT4:
>  I'm going to remove mavericks support from homebrew as we don't explicitly support it as a project anymore.

This is fine, but please add a note in the formula above the `:yosemite` enforcement about it, otherwise someone will try and reverse it at some point.

marpaia:
@DomT4 :+1:

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/aa4cd8ebfa092a166e33566ea6819d57b90ab4a1. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #40986, opened 2015-06-22
added ppl-1.1

UnixJunkie:


Library/Formula/ppl.rb
```diff
@@ -0,0 +1,25 @@
+class Ppl < Formula
+  desc "Parma Polyhedra Library"
+  homepage "http://bugseng.com/products/ppl"
+  url "ftp://ftp.cs.unipr.it/pub/ppl/releases/1.1/ppl-1.1.tar.xz"
```
vszakats: http is preferred over ftp, I'd suggest to use this `url`:
http://bugseng.com/products/ppl/download/ftp/releases/1.1/ppl-1.1.tar.xz

Library/Formula/ppl.rb
```diff
@@ -0,0 +1,25 @@
...
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  def uninstall
+    system "make", "uninstall"
+  end
+
+  test do
+    system "make", "check"
```
DomT4: This won't work. The `testpath` has no access to the `buildpath`, and vice versa.

Library/Formula/ppl.rb
```diff
@@ -0,0 +1,25 @@
+class Ppl < Formula
+  desc "Parma Polyhedra Library"
+  homepage "http://bugseng.com/products/ppl"
+  url "ftp://ftp.cs.unipr.it/pub/ppl/releases/1.1/ppl-1.1.tar.xz"
+  version "1.1"
```
DomT4: URL is simple enough that `brew` can understand the version. Can scrap the manual `version` line.

Library/Formula/ppl.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  def uninstall
```
DomT4: We don't need to define this. Our removal code isn't part of formulae.

UnixJunkie:
OK, I took all comments into account

Library/Formula/ppl.rb
```diff
@@ -0,0 +1,16 @@
...
+  depends_on "gmp"
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+end
```
mikemcquaid: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/ppl.rb
```diff
@@ -0,0 +1,16 @@
+class Ppl < Formula
+  desc "Parma Polyhedra Library"
+  homepage "http://bugseng.com/products/ppl"
```
mikemcquaid: What's the license?

Library/Formula/ppl.rb
UnixJunkie: I don't know how to do this: you would have to go where the software was installed and run 'make check'.

UnixJunkie:
I added the license info in the desc field

Library/Formula/ppl.rb
mikemcquaid: Please look at the referenced formulae.

UnixJunkie:
I just added a test target inspired by the one from tinyxml2.rb

mikemcquaid:

For future reference the preferred commit message format for new formulae is `ppl 1.1 (new formula)`.

PR #40988, opened 2015-06-22
added trace2html

brentvatne:
On [React Native](https://github.com/facebook/react-native) we will be using `trace2html` to visualize the results from our profiler - it would be great to be able to just have users install via Homebrew, even though it's fairly simple.

### DIFF
--- a/Library/Formula/trace2html.rb
+++ b/Library/Formula/trace2html.rb
@@ -1,19 +1,19 @@
 class Trace2html < Formula
   desc "Utility from Google Trace Viewer to convert JSON traces to HTML"
   homepage "https://github.com/google/trace-viewer"
-  url "https://github.com/google/trace-viewer/archive/a4c4894801935627ff412c3cc6c3ddae96019cc5.zip"
-  sha256 "b9fc71e6f1a36f91ca09c1ea26ae2c38e42d3183b6470e603d1db8b1ee6e8316"
+  url "https://github.com/google/trace-viewer/archive/a4c4894801935627ff412c3cc6c3ddae96019cc5.tar.gz"
+  sha256 "40bf1613e23c0f483536667d3ab584889c50219768414f0086150b3c364269c3"
 
-  depends_on "Python"
+  depends_on :python if MacOS.version <= :snow_leopard
 
   def install
-    cp_r(buildpath, libexec)
+    libexec.install Dir["*"]
     bin.install_symlink libexec/"trace2html"
   end
 
   test do
     touch "test.json"
-    system "#{bin}/trace2html test.json"
+    system "#{bin}/trace2html", "test.json"
     assert File.exist?("test.html")
   end
 end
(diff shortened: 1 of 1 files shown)


Library/Formula/trace2html.rb
```diff
@@ -0,0 +1,19 @@
+class Trace2html < Formula
+  desc "Utility from Google Trace Viewer to convert JSON traces to HTML"
+  homepage "https://github.com/google/trace-viewer"
+  url "https://github.com/google/trace-viewer/archive/a4c4894801935627ff412c3cc6c3ddae96019cc5.zip"
+  sha256 "b9fc71e6f1a36f91ca09c1ea26ae2c38e42d3183b6470e603d1db8b1ee6e8316"
+
+  depends_on "Python"
```
DomT4: Don't hardcode Homebrew's Python. We use the system Python on OS X versions that support >2.7. You can invoke that dependency with `depends_on :python if MacOS.version <= :snow_leopard`.

Library/Formula/trace2html.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def install
+    cp_r(buildpath, libexec)
```
DomT4: `libexec.install Dir["*"]`

DomT4:
Couple of Audit nits as well:

```
==> audit problems
trace2html:
 * Stable: Use GitHub tarballs rather than zipballs (url is https://github.com/google/trace-viewer/archive/a4c4894801935627ff412c3cc6c3ddae96019cc5.zip).
 * Use `system "#{bin}/trace2html", "test.json"` instead of `system "#{bin}/trace2html test.json"` 
```

brentvatne:
Thanks @DomT4! Fixed! We're releasing a new version tomorrow and it would be great to include this install step in the release notes, but if it's not possible to merge before then, that's totally understandable.

DomT4:
We'll need an actual version or git tag from upstream to include it, rather than just a commit tarball - But if we can get that done it's not impossible to get things in relatively quickly. The tool is popular enough and supported enough to meet our requirements.

brentvatne:
Thanks again @DomT4 - I created an issue on the trace-viewer repo, as you can see above, and referenced this.

brentvatne:
@DomT4 - The trace-viewer team was very responsive and have already gone ahead and added a tagged release. This pull request has been updated accordingly :smile:

mikemcquaid:

For future reference the preferred commit message format for new formulae is `trace2html 2015-06-23 (new formula)`.

DomT4:
@brentvatne Thanks for the responsiveness on this, both from you and upstream! :beers:

brentvatne:
Thanks a lot guys! Sorry about that @mikemcquaid - I'll do that in the future

paramaggarwal:
Just landed here looking for an easy way to install `trace2html` for use in React Native - thanks @brentvatne and @mikemcquaid!

Issue #40897, opened 2015-06-19
Edge case with HTTP servers failing to respond to HEAD requests


L2G:
Here is a headache I've just discovered: when an HTTP server does not respond to HEAD requests correctly, Homebrew will give up on a download, even if the URL would otherwise allow for a successful download.

Here are the details of the particular case I discovered:

* The `pipebench` formula points to http://www.habets.pp.se/synscan/files/pipebench-0.40.tar.gz for a download.
* Homebrew starts its download attempt by calling `curl -I -L http://www.habets.pp.se/synscan/files/pipebench-0.40.tar.gz`.
* The remote web server at www.habets.pp.se apparently does not respond to HEAD requests. If I run the above `curl` command manually from the command line, it hangs indefinitely. (Homebrew times out after a sensible duration.)
* If I try `curl -LO http://www.habets.pp.se/synscan/files/pipebench-0.40.tar.gz`, the tarball is downloaded immediately and without error.

L2G:
If it would be helpful, I'll volunteer to contact the webmaster of www.habets.pp.se and see whether the issue can be corrected on the server side.

L2G:
I'm also continuing to experiment to make sure this isn't really just a bug in curl 7.30, the version that comes with Mavericks.

DomT4:
Fails here too on Yosemite, both with system `curl` and Homebrew'd `curl`. I did a `brew fetch` of about 300 formulae the other night without hitting this, and theoretically servers are *supposed* to answer HEAD requests. I'm more inclined to tag this as something upstream should fix than us, but short-term we could look at giving the formula its own download strategy.

L2G:
I tried manually generating HTTP requests via good ol' telnet. It definitely looks like the remote web server is misbehaving.

(I've seen weirder things. Once I saw a noncompliant HTTP server return 404 for a HEAD request while returning 200 for a GET request to the exact same URL! Fortunately I haven't seen that happen with any Homebrew downloads, at least not yet.)

L2G:
FYI, I've just sent a message via [the website's contact form](http://www.habets.pp.se/synscan/contact.php) reaching out to the webmaster.

DomT4:
> Once I saw a noncompliant HTTP server return 404 for a HEAD request while returning 200 for a GET request to the exact same URL

The general rule for servers seems to be "If it's possible to set them up incorrectly, It'll be done" :wink:.

[...]

PR #40993, opened 2015-06-23
pipebench: use debian as url

DomT4:
The upstream server for this formula behaves in a number of strange ways detailed in https://github.com/Homebrew/homebrew/issues/40897.

The inconsistency of handing off different files to browsers and cURL in particular, and the checksum changes that can come with that make it hard to verify which package is intended to be the "proper" one from the server.

### DIFF
--- a/Library/Formula/pipebench.rb
+++ b/Library/Formula/pipebench.rb
@@ -1,15 +1,16 @@
-require 'formula'
-
 class Pipebench < Formula
   desc "Measure the speed of STDIN/STDOUT communication"
-  homepage 'http://www.habets.pp.se/synscan/programs.php?prog=pipebench'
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

DomT4:
:rocket:. Thanks Mike!

DomT4:
Possible we just stumbled onto a bug in the bottle code. It shoved the `url`, `mirror` and `sha256` out of the way for the bottles. I haven't seen this elsewhere, so perhaps it doesn't anticipate the use of commented out `url`s? CC @xu-cheng 

Ref https://github.com/Homebrew/homebrew/commit/25d1c9bb71dc1dca7df026b34cf117fd8586a5c9

PR #40995, opened 2015-06-23
gtk+3: quartz relocation off by default

tschoonj:
introduced an option to activate it


mikemcquaid:
What does this do?

tschoonj:
This disables default quartz-relocation support, which should only be turned on when the goal is to put gtk and its dependencies in an app bundle. If not, then this option might actually break things, especially when the homebrew prefix is not /usr/local.

mikemcquaid:
Maybe it's worth not even making it an option?

tschoonj:
I guess for people who know what they are doing, this could actually turn out to be quite a useful option.

mikemcquaid:
If that's you personally I'm happy to add this. Otherwise I suspect people who don't know what they are doing might try it out and break things.

tschoonj:
I would indeed be a user of this option :-)

By the way: the same option is present in the `gtk+` formula.

mikemcquaid:
:cool:

mikemcquaid:
Thanks @tschoonj!

PR #40996, opened 2015-06-23
tbb: https homepage

vszakats:


mikemcquaid:
Thanks @vszakats!

mikemcquaid:
Although I do think it might be worth trying to write an `audit` rule that automatically detects these e.g. tries to `curl` the `https` URL and sees if it responds.

vszakats:
Here's my shell script attempt (with some updates): https://gist.github.com/vszakats/663dc8ccf1d49b903f8e 
and @bfontaine's Ruby implementation: https://github.com/bfontaine/homebrew-tools/blob/brew-fix-https/cmd/brew-fix.rb#L44-L65

None of these is 100% for `homepage` (there are cases where https do respond, but with different content).
For `url` the basic case is easier to automatize.

There is also this effort to upgrade to https on the client side: https://github.com/EFForg/https-everywhere/tree/master/src/chrome/content/rules (rules are complex). It's a long shot, just for reference.

I've added a short sentence about it to the Formula Cookbook in the meantime: https://github.com/Homebrew/homebrew/pull/40998

PR #40998, opened 2015-06-23
Formula-Cookbook: SSL/TLS homepage is preferred

vszakats:
It's a single sentence, but lots of things can go wrong even here; I appreciate comments.


mikemcquaid:
Thanks @vszakats!

PR #41001, opened 2015-06-23
neo4j 2.2.3

mindrones:
- bumped version
- updated the url as `http://neo4j.org` is a permanent redirect (301) to `http://neo4j.com`


mikemcquaid:
Thanks @mindrones!

PR #41003, opened 2015-06-23
Make warnings yellow (previously red)

radeksimko:
Similar to https://github.com/caskroom/homebrew-cask/pull/12067

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -55,7 +55,7 @@ def oh1 title
 end
 
 def opoo warning
-  $stderr.puts "#{Tty.red}Warning#{Tty.reset}: #{warning}"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Makes sense to me but other @Homebrew/owners may have thoughts.

bfontaine:
Yes that makes sense.

DomT4:
No objections. Helps create a difference between *"This could be a problem"* and *"This is a problem"* which is something we've seen confusing people before.

xu-cheng:
:+1:

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/681b438398f028eaf00c644bc4a8df64435a937a. Thanks @radeksimko!

PR #41004, opened 2015-06-23
node 0.12.5

btmills:
npm was upgraded to 2.11.2 as part of the 0.12.5 release, so I included that upgrade in this PR as well. Let me know if it should be separate.

For the blog post announcement, see http://blog.nodejs.org/2015/06/22/node-v0-12-5-stable/

### DIFF
--- a/Library/Formula/node.rb
+++ b/Library/Formula/node.rb
@@ -2,8 +2,8 @@
 class Node < Formula
   desc "Platform built on Chrome's JavaScript runtime to build network applications"
   homepage "https://nodejs.org/"
-  url "https://nodejs.org/dist/v0.12.4/node-v0.12.4.tar.gz"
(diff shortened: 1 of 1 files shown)


mikemcquaid:

welly:
There is a real problem with the homebrew node package in that it causes segmentation faults. I've been running it for a few days now and it's been consistently inconsistent. I've removed the package and trying the official node release. Will see if that resolves the problems I've been having.

DomT4:
It's better to file Issues rather than comment on the last PR - It's harder to spot the latter, and plugs people up with notifications.

I haven't experienced any Node issues with the latest package, either with an official upstream package or the one build from source via Homebrew. If you can provide a reproducible test case that'd be great, Thanks.

PR #41006, opened 2015-06-23
GitHub Summer of Code: Formula renames

vladshablinsky:
Hi! I'm a summer of code student working on "Handling formula renames" project https://github.com/Homebrew/homebrew/issues/14374. My mentor is @mikemcquaid. 

The problem with formula renames is that if we just rename the formula and its class, user who has this formula installed will never be able to upgrade it. The second problem are dependencies, if we rename formula we won't be able to install any formula that depends on old name.

The solution is to inform user about renames and perform some migrations that won't break anything. After that user can use his software, but the software will move to new directory and the formula for the software will be changed to new name.

Let's define `oldname` as an old name of the package and the formula and `newname` for new ones.

Approach
=======
The approach is to have `FORMULA_RENAMES` hash, that stores old names as keys and new names as values. The migrations are performed during the `update`, when we have information about added and deleted formulae from diff.

The approach doesn't allow us to create `oldname` formula after it gets renamed to `newname`.

Pros of this:

* No mess in versions
* Simple structure of metadata
* Issues like https://github.com/Homebrew/homebrew/issues/15776 can still be fixed.

Update
======
* Get the `formula_renames.rb` containing `FORMULA_RENAMES` from the last revision and load it.
* If we find from the `report` that the core formula was deleted and we have it in the Cellar, we try to find it in `FORMULA_RENAMES` and perform migrations using `Migrator` class, described below.

Migrator
=======
`Migrator` is a class for performing migrations for the formulae, that have `oldname`s. If the formula doesn't have `oldname`, then the exception is raised.
`Migrator#migrate` performs following steps:

1. Unlink oldname keg
2. Move `Cellar/oldname` to `Cellar/newname` if `Cellar/newname` doesn't exist already.
3. Pin `newname` keg if `oldname` was pinned.
4. Link newname keg.
6. Create `Cellar/oldname` symlink with `Cellar/newname` target.
7. Create opt link if `oldname` had it.

If one of these steps goes wrong, everything rolls back to initial state, which brings us to `migrate` command.

Migrate
======
If automatic migration fails, the user have to manually migrate his package.
 `brew migrate oldname` is used for that purpose.

Uninstalling
=========
If we uninstall `newname` after migration, we also remove `oldname` links to `newname`.

Taps
====
Taps can also be renamed.
The structure is almost the same as for core formulae.

Each tap that has renamed formulae must be provided with `formula_renames.json` file with, that stores information about renames in the following format:
```json
{
  "oldname": "newname"
}
```

TODOs
======

- [x] Establish behaviour for changed `ARGV.kegs` and rewrite it if needed.
- [x] Rename formulae from taps.
     - [x] uninstalling
     - [x] dependencies
     - [x] updating
- [x] `brew upgrade newname` now won't relink `oldname`. It's OK.
- [x] unit tests (in progress)
    - [x] tests for migrator exceptions
    - [x] tests for migrator

`brew migrate` manual tests
=====================
- [x] Migrate core formula
- [x] Migrate tap formula (canonical/fully-qualified) 
- [x] Migrate core formula when the same tap formula exists
- [x] Migrate tap formula when the same core one exists
- [x] Migrate tap formula using canonical name
- [x] Migrate tap formula using canonical name, when multiple taps have this formula
- [x] Install formula from path and try to migrate it (MigratorDifferentTapsError is raised)
- [x] Migrate formula, that depends on another
- [x] Migrate library some formula depends on (tested on zint and libpng, itstool and libxml2)
- [x] The same for library formulae 
- [x] Dependency on tap oldname :+1: both canonical and fully-qualified names work.
- [x] Dependency on core formula oldname (reinstall zint after renaming libpng)
- [x] Uninstall using oldname (canonical/fully-qualified) `Error: /usr/local/Cellar/ack is a symlink` :+1: 
- [x] Uninstall using newname (canonical/fully-qualified) OK :+1: 
- [x] Migrating keg_only works.

*NOTE*
What if there are two taps with `newack` formula? (`user/repo/newack` and `user1/repo1/newack`) and we want to `migrate` ack?
Both repositories have `"ack" => "newack"` in `TAP_FORMULA_RENAMES` hash.
`brew migrate ack` shows error (`TapFormulaWithOldnameAmbiguityError`) and says that you have to use fully-qualified name. However, if only one tap has "ack" renamed, everything is OK.

Changes
=======
* add `migrate` command.
* add `Migrator` class for handling migrations.
* `uninstall oldname` uninstalls `newname` with `oldname` symlinks.
* `uninstall newname` uninstalls package with its oldname symlinks.
* `update` perform formula renames(migrations) and tap formula renames(migrations) for renamed formulae.
* `update` changes `Tab` if formula migrates to a tap.
* `Formula#oldname` returns oldname for formula.
* `Formulary#loader_for(ref)` tries to load formula from oldname and from tap oldname.
*`TapLoader` tries to load tap formula from oldname.
* Add `Formulary.to_rack`
*  `Tap#formula_rename` loads `formula_renames.json` and stores Hash with renames.
* `brew migrate oldname` migrates `oldname` to `newname`.
* `brew audit` warns about dependencies on `oldname`.
* Add `TapFormulaWithOldnameAmbiguityError` exception that is raised when there are multiple tap formulae with given `oldname`.
* Add `test_migrator.rb` tests.
* Add tests for errors in `Migrator`.

Any feedback/suggestions on the approach and anything else are appreciated.
Thanks!

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -6,6 +6,7 @@ require "official_taps"
 require "tap_migrations"
 require "cmd/search"
 require "date"
+require "formula_renames"
 
 module Homebrew
   def audit
@@ -229,6 +230,11 @@ class FormulaAuditor
       return
     end
 
+    if FORMULA_RENAMES.key? name
+      problem "'#{name}' is reserved as the old name of #{FORMULA_RENAMES[name]}"
+      return
+    end
+
     if !formula.core_formula? && Formula.core_names.include?(name)
       problem "Formula name conflicts with existing core formula."
       return
@@ -268,6 +274,10 @@ class FormulaAuditor
         rescue TapFormulaAmbiguityError
           problem "Ambiguous dependency #{dep.name.inspect}."
           next
+        end
+
+        if FORMULA_RENAMES[dep.name] == dep_f.name
+          problem "Dependency '#{dep.name}' was renamed; use newname '#{dep_f.name}'."
         end
 
         if @@aliases.include?(dep.name)
--- a/Library/Homebrew/cmd/migrate.rb
+++ b/Library/Homebrew/cmd/migrate.rb
@@ -1,0 +1,20 @@
+require "migrator"
+require "formula_renames"
+
+module Homebrew
+  def migrate
+    raise FormulaUnspecifiedError if ARGV.named.empty?
+
+    ARGV.resolved_formulae.each do |f|
+      if f.oldname
+        unless (rack = HOMEBREW_CELLAR/f.oldname).exist? && !rack.subdirs.empty?
+          raise NoSuchKegError, f.oldname
+        end
+        raise "#{rack} is a symlink" if rack.symlink?
+      end
+
+      migrator = Migrator.new(f)
+      migrator.migrate
+    end
+  end
+end
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -1,30 +1,66 @@
 require "keg"
 require "formula"
+require "migrator"
 
 module Homebrew
   def uninstall
     raise KegUnspecifiedError if ARGV.named.empty?
 
+    # Find symlinks that can point to keg.rack
+    links = HOMEBREW_CELLAR.subdirs.select(&:symlink?)
+
     if !ARGV.force?
       ARGV.kegs.each do |keg|
         keg.lock do
           puts "Uninstalling #{keg}... (#{keg.abv})"
+
+          old_cellars = []
+          # Remove every symlink that links to keg, because it can
+          # be left by migrator
+          links.each do |link|
+            old_opt = HOMEBREW_PREFIX/"opt/#{link.basename}"
+            if link.exist? && link.realpath == keg.rack.realpath
+              old_cellars << link
+            end
+
+            if old_opt.symlink? && old_opt.realpath.to_s == keg.to_s
+              old_opt.unlink
+              old_opt.parent.rmdir_if_possible
+            end
+          end
+
           keg.unlink
           keg.uninstall
           rack = keg.rack
           rm_pin rack
+
           if rack.directory?
             versions = rack.subdirs.map(&:basename)
             verb = versions.length == 1 ? "is" : "are"
             puts "#{keg.name} #{versions.join(", ")} #{verb} still installed."
             puts "Remove them all with `brew uninstall --force #{keg.name}`."
+          else
+            # If we delete Cellar/newname, then Cellar/oldname symlink
+            # can become broken and we have to remove it.
+            old_cellars.each(&:unlink)
           end
         end
       end
     else
       ARGV.named.each do |name|
-        name = Formulary.canonical_name(name)
-        rack = HOMEBREW_CELLAR/name
+        rack = Formulary.to_rack(name)
+        name = rack.basename
+
+        links.each do |link|
+          old_opt = HOMEBREW_PREFIX/"opt/#{link.basename}"
+          if old_opt.symlink? && old_opt.exist? \
+              && old_opt.realpath.parent == rack.realpath
+            old_opt.unlink
+            old_opt.parent.rmdir_if_possible
+          end
+
+          lnk.unlink if lnk.exist? && lnk.realpath == rack.realpath
+        end
 
         if rack.directory?
           puts "Uninstalling #{name}... (#{rack.abv})"
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -1,5 +1,7 @@
 require "cmd/tap"
 require "formula_versions"
+require "migrator"
+require "formulary"
 
 module Homebrew
   def update
@@ -46,12 +48,46 @@ module Homebrew
 
     # automatically tap any migrated formulae's new tap
     report.select_formula(:D).each do |f|
-      next unless (HOMEBREW_CELLAR/f).exist?
+      next unless (dir = HOMEBREW_CELLAR/f).exist?
       migration = TAP_MIGRATIONS[f]
       next unless migration
       tap_user, tap_repo = migration.split "/"
       install_tap tap_user, tap_repo
+      # update tap for each Tab
+      tabs = dir.subdirs.each.map { |d| Tab.for_keg(Keg.new(d)) }
+      tabs.each { |tab| tab.source["tap"] = "#{tap_user}/homebrew-#{tap_repo}" }
+      tabs.each(&:write)
     end if load_tap_migrations
+
+    # Migrate installed renamed formulae from main Homebrew repository.
+    if load_formula_renames
+      report.select_formula(:D).each do |oldname|
+        newname = FORMULA_RENAMES[oldname]
+        next unless newname
+        next unless (dir = HOMEBREW_CELLAR/oldname).directory? && !dir.subdirs.empty?
+
+        begin
+          migrator = Migrator.new(Formulary.factory("homebrew/homebrew/#{newname}"))
+          migrator.migrate
+        rescue Migrator::MigratorDifferentTapsError
+        end
+      end
+    end
+
+    # Migrate installed renamed formulae from taps
+    report.select_formula(:D).each do |oldname|
+      user, repo, oldname = oldname.split("/", 3)
+      next unless user && repo && oldname
+      tap = Tap.new(user, repo)
+      next unless newname = tap.formula_renames[oldname]
+      next unless (dir = HOMEBREW_CELLAR/oldname).directory? && !dir.subdirs.empty?
+
+      begin
+        migrator = Migrator.new(Formulary.factory("#{user}/#{repo}/#{newname}"))
+        migrator.migrate
+      rescue Migrator::MigratorDifferentTapsError
+      end
+    end
 
     if report.empty?
       puts "Already up-to-date."
@@ -112,7 +148,13 @@ module Homebrew
   end
 
   def load_tap_migrations
-    require "tap_migrations"
+    load "tap_migrations"
+  rescue LoadError
+    false
+  end
+
+  def load_formula_renames
+    load "formula_renames.rb"
   rescue LoadError
     false
   end
--- a/Library/Homebrew/exceptions.rb
+++ b/Library/Homebrew/exceptions.rb
@@ -86,6 +86,26 @@ class TapFormulaAmbiguityError < RuntimeError
   end
 end
 
+class TapFormulaWithOldnameAmbiguityError < RuntimeError
+  attr_reader :name, :possible_tap_newname_formulae, :taps
+
+  def initialize(name, possible_tap_newname_formulae)
+    @name = name
+    @possible_tap_newname_formulae = possible_tap_newname_formulae
+
+    @taps = possible_tap_newname_formulae.map do |newname|
+      newname =~ HOMEBREW_TAP_FORMULA_REGEX
+      "#{$1}/#{$2}"
+    end
+
+    super <<-EOS.undent
+      Formulae with '#{name}' old name found in multiple taps: #{taps.map { |t|  "\n       * #{t}" }.join}
+
+      Please use the fully-qualified name e.g. #{taps.first}/#{name} to refer the formula or use its new name.
+    EOS
+  end
+end
+
 class TapUnavailableError < RuntimeError
   attr_reader :name
 
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -47,14 +47,14 @@ module HomebrewArgvExtension
     require "keg"
     require "formula"
     @kegs ||= downcased_unique_named.collect do |name|
-      canonical_name = Formulary.canonical_name(name)
-      rack = HOMEBREW_CELLAR/canonical_name
+      rack = Formulary.to_rack(name)
+
       dirs = rack.directory? ? rack.subdirs : []
 
-      raise NoSuchKegError.new(canonical_name) if dirs.empty?
+      raise NoSuchKegError.new(rack.basename) if dirs.empty?
 
-      linked_keg_ref = HOMEBREW_LIBRARY.join("LinkedKegs", canonical_name)
-      opt_prefix = HOMEBREW_PREFIX.join("opt", canonical_name)
+      linked_keg_ref = HOMEBREW_LIBRARY.join("LinkedKegs", rack.basename)
+      opt_prefix = HOMEBREW_PREFIX.join("opt", rack.basename)
 
       begin
         if opt_prefix.symlink? && opt_prefix.directory?
@@ -66,7 +66,7 @@ module HomebrewArgvExtension
         elsif (prefix = (name.include?("/") ? Formulary.factory(name) : Formulary.from_rack(rack)).prefix).directory?
           Keg.new(prefix)
         else
-          raise MultipleVersionsInstalledError.new(canonical_name)
+          raise MultipleVersionsInstalledError.new(rack.basename)
         end
       rescue FormulaUnavailableError
         raise <<-EOS.undent
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -10,6 +10,7 @@ require "software_spec"
 require "install_renamed"
 require "pkg_version"
 require "tap"
+require "formula_renames"
 
 # A formula provides instructions and metadata for Homebrew to install a piece
 # of software. Every Homebrew formula is a {Formula}.
@@ -221,6 +222,21 @@ class Formula
   # A named Resource for the currently active {SoftwareSpec}.
   def resource(name)
     active_spec.resource(name)
+  end
+
+  # An old name for the formula
+  def oldname
+    @oldname ||= if core_formula?
+      if FORMULA_RENAMES && FORMULA_RENAMES.value?(name)
+        FORMULA_RENAMES.to_a.rassoc(name).first
+      end
+    elsif tap?
+      user, repo = tap.split("/")
+      formula_renames = Tap.new(user, repo.sub("homebrew-", "")).formula_renames
+      if formula_renames.value?(name)
+        formula_renames.to_a.rassoc(name).first
+      end
+    end
   end
 
   # The {Resource}s for the currently active {SoftwareSpec}.
@@ -864,6 +880,7 @@ class Formula
       "full_name" => full_name,
       "desc" => desc,
       "homepage" => homepage,
+      "oldname" => oldname,
       "versions" => {
         "stable" => (stable.version.to_s if stable),
         "bottle" => bottle ? true : false,
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -195,6 +195,24 @@ class Formulary
     end
   end
 
+  def self.to_rack(ref)
+    name = canonical_name(ref)
+    rack = HOMEBREW_CELLAR/name
+
+    # Handle the case when ref is an old name and the installation
+    # hasn't been migrated or when it's a package installed from
+    # path but same name formula was renamed.
+    unless rack.directory?
+      if ref =~ HOMEBREW_TAP_FORMULA_REGEX
+        rack = HOMEBREW_CELLAR/$3
+      elsif !ref.include?("/")
+        rack = HOMEBREW_CELLAR/ref
+      end
+    end
+
+    rack
+  end
+
   def self.canonical_name(ref)
     loader_for(ref).name
   rescue TapFormulaAmbiguityError
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -1,4 +1,5 @@
 require "digest/md5"
+require "formula_renames"
 
 # The Formulary is responsible for creating instances of Formula.
 # It is not meant to be used directy from formulae.
@@ -141,6 +142,7 @@ class Formulary
     def initialize(tapped_name)
       user, repo, name = tapped_name.split("/", 3).map(&:downcase)
       @tap = Tap.new user, repo.sub(/^homebrew-/, "")
+      name = @tap.formula_renames.fetch(name, name)
       path = @tap.formula_files.detect { |file| file.basename(".rb").to_s == name }
       path ||= @tap.path/"#{name}.rb"
 
@@ -212,7 +214,13 @@ class Formulary
     when Pathname::BOTTLE_EXTNAME_RX
       return BottleLoader.new(ref)
     when HOMEBREW_CORE_FORMULA_REGEX
-      return FormulaLoader.new($1, core_path($1))
+      name = $1
+      formula_with_that_name = core_path(name)
+      if (newname = FORMULA_RENAMES[name]) && !formula_with_that_name.file?
+        return FormulaLoader.new(newname, core_path(newname))
+      else
+        return FormulaLoader.new(name, formula_with_that_name)
+      end
     when HOMEBREW_TAP_FORMULA_REGEX
       return TapLoader.new(ref)
     end
@@ -231,11 +239,31 @@ class Formulary
       return AliasLoader.new(possible_alias)
     end
 
-    possible_tap_formulae = tap_paths(ref)
+   possible_tap_formulae = tap_paths(ref)
     if possible_tap_formulae.size > 1
       raise TapFormulaAmbiguityError.new(ref, possible_tap_formulae)
     elsif possible_tap_formulae.size == 1
       return FormulaLoader.new(ref, possible_tap_formulae.first)
+    end
+
+    if newref = FORMULA_RENAMES[ref]
+      formula_with_that_oldname = core_path(newref)
+      if formula_with_that_oldname.file?
+        return FormulaLoader.new(newref, formula_with_that_oldname)
+      end
+    end
+
+    possible_tap_newname_formulae = []
+    Tap.each do |tap|
+      if newref = tap.formula_renames[ref]
+        possible_tap_newname_formulae << "#{tap.name}/#{newref}"
+      end
+    end
+
+    if possible_tap_newname_formulae.size > 1
+      raise TapFormulaWithOldnameAmbiguityError.new(ref, possible_tap_newname_formulae)
+    elsif !possible_tap_newname_formulae.empty?
+      return TapLoader.new(possible_tap_newname_formulae.first)
     end
 
     possible_cached_formula = Pathname.new("#{HOMEBREW_CACHE_FORMULA}/#{ref}.rb")
--- a/Library/Homebrew/manpages/brew.1.md
+++ b/Library/Homebrew/manpages/brew.1.md
@@ -298,6 +298,13 @@ Note that these flags should only appear after a command.
 
     If no <formulae> are given, check all installed brews.
 
+  * `migrate [--force]` <formulae>:
+    Migrate renamed packages to new name, where <formulae> are old names of
+    packages.
+
+    If `--force` is passed and installed <formulae> have nil tap, then treat
+    them like packages installed from core.
+
   * `options [--compact] [--all] [--installed]` <formula>:
     Display install options specific to <formula>.
 
--- a/Library/Homebrew/migrator.rb
+++ b/Library/Homebrew/migrator.rb
@@ -1,0 +1,303 @@
+require "formula"
+require "keg"
+require "tab"
+require "tap_migrations"
+
+class Migrator
+  class MigratorNoOldnameError < RuntimeError
+    def initialize(formula)
+      super "#{formula.name} doesn't replace any formula."
+    end
+  end
+
+  class MigratorNoOldpathError < RuntimeError
+    def initialize(formula)
+      super "#{HOMEBREW_CELLAR/formula.oldname} doesn't exist."
+    end
+  end
+
+  class MigratorDifferentTapsError < RuntimeError
+    def initialize(formula, tap)
+      if tap.nil?
+        super <<-EOS.undent
+        #{formula.name} from #{formula.tap} is given, but old name #{formula.oldname} wasn't installed from taps or core formulae
+
+        You can try `brew migrate --force #{formula.oldname}`.
+        EOS
+      else
+        user, repo = tap.split("/")
+        repo.sub!("homebrew-", "")
+        name = "fully-qualified #{user}/#{repo}/#{formula.oldname}"
+        name = formula.oldname if tap == "Homebrew/homebrew"
+        super <<-EOS.undent
+        #{formula.name} from #{formula.tap} is given, but old name #{formula.oldname} was installed from #{tap}
+
+        Please try to use #{name} to refer the formula
+        EOS
+      end
+    end
+  end
+
+  attr_reader :formula
+  attr_reader :oldname, :oldpath, :old_pin_record, :old_opt_record
+  attr_reader :old_linked_keg_record, :oldkeg, :old_tabs, :old_tap
+  attr_reader :newname, :newpath, :new_pin_record
+  attr_reader :old_pin_link_record
+
+  def initialize(formula)
+    @oldname = formula.oldname
+    @newname = formula.name
+    raise MigratorNoOldnameError.new(formula) unless oldname
+
+    @formula = formula
+    @oldpath = HOMEBREW_CELLAR/formula.oldname
+    raise MigratorNoOldpathError.new(formula) unless oldpath.exist?
+
+    @old_tabs = oldpath.subdirs.each.map { |d| Tab.for_keg(Keg.new(d)) }
+    @old_tap = old_tabs.first.tap
+    raise MigratorDifferentTapsError.new(formula, old_tap) unless from_same_taps?
+
+    @newpath = HOMEBREW_CELLAR/formula.name
+
+    if @oldkeg = get_linked_oldkeg
+      @old_linked_keg_record = oldkeg.linked_keg_record if oldkeg.linked?
+      @old_opt_record = oldkeg.opt_record if oldkeg.optlinked?
+    end
+
+    @old_pin_record = HOMEBREW_LIBRARY/"PinnedKegs"/oldname
+    @new_pin_record = HOMEBREW_LIBRARY/"PinnedKegs"/newname
+    @pinned = old_pin_record.symlink?
+    @old_pin_link_record = old_pin_record.readlink if @pinned
+  end
+
+  # Fix INSTALL_RECEIPTS for tap-migrated formula.
+  def fix_tabs
+    old_tabs.each do |tab|
+      tab.source["tap"] = formula.tap
+      tab.write
+    end
+  end
+
+  def from_same_taps?
+    if old_tap == nil && formula.core_formula? && ARGV.force?
+      true
+    elsif formula.tap == old_tap
+      true
+    # Homebrew didn't use to update tabs while performing tap-migrations,
+    # so there can be INSTALL_RECEIPT's containing wrong information about
+    # tap (tap is Homebrew/homebrew if installed formula migrates to a tap), so
+    # we check if there is an entry about oldname migrated to tap and if
+    # newname's tap is the same as tap to which oldname migrated, then we
+    # can perform migrations and the taps for oldname and newname are the same.
+    elsif TAP_MIGRATIONS && (rec = TAP_MIGRATIONS[formula.oldname]) \
+          && rec == formula.tap.sub("homebrew-", "")
+      fix_tabs
+      true
+    elsif formula.tap
+      false
+    end
+  end
+
+  def get_linked_oldkeg
+    kegs = oldpath.subdirs.map { |d| Keg.new(d) }
+    kegs.detect(&:linked?) || kegs.detect(&:optlinked?)
+  end
+
+  def pinned?
+    @pinned
+  end
+
+  def oldkeg_linked?
+    !!oldkeg
+  end
+
+  def migrate
+    if newpath.exist?
+      onoe "#{newpath} already exists; remove it manually and run brew migrate #{oldname}."
+      return
+    end
+
+    begin
+      oh1 "Migrating #{Tty.green}#{oldname}#{Tty.white} to #{Tty.green}#{newname}#{Tty.reset}"
+      unlink_oldname
+      move_to_new_directory
+      repin
+      link_newname
+      link_oldname_opt
+      link_oldname_cellar
+      update_tabs
+    rescue Interrupt
+      ignore_interrupts { backup_oldname }
+    rescue Exception => e
+      onoe "error occured while migrating."
+      puts e if ARGV.debug?
+      puts "Backuping..."
+      ignore_interrupts { backup_oldname }
+    end
+  end
+
+  # move everything from Cellar/oldname to Cellar/newname
+  def move_to_new_directory
+    puts "Moving to: #{newpath}"
+    FileUtils.mv(oldpath, newpath)
+  end
+
+  def repin
+    if pinned?
+      # old_pin_record is a relative symlink and when we try to to read it
+      # from <dir> we actually try to find file
+      # <dir>/../<...>/../Cellar/name/version.
+      # To repin formula we need to update the link thus that it points to
+      # the right directory.
+      # NOTE: old_pin_record.realpath.sub(oldname, newname) is unacceptable
+      # here, because it resolves every symlink for old_pin_record and then
+      # substitutes oldname with newname. It breaks things like
+      # Pathname#make_relative_symlink, where Pathname#relative_path_from
+      # is used to find relative path from source to destination parent and
+      # it assumes no symlinks.
+      src_oldname = old_pin_record.dirname.join(old_pin_link_record).expand_path
+      new_pin_record.make_relative_symlink(src_oldname.sub(oldname, newname))
+      old_pin_record.delete
+    end
+  end
+
+  def unlink_oldname
+    oh1 "Unlinking #{Tty.green}#{oldname}#{Tty.reset}"
+    oldpath.subdirs.each do |d|
+      keg = Keg.new(d)
+      keg.unlink
+    end
+  end
+
+  def link_newname
+    oh1 "Linking #{Tty.green}#{newname}#{Tty.reset}"
+    keg = Keg.new(formula.installed_prefix)
+
+    if formula.keg_only?
+      begin
+        keg.optlink
+      rescue Keg::LinkError => e
+        onoe "Failed to create #{formula.opt_prefix}"
+        puts e
+        raise
+      end
+      return
+    end
+
+    keg.remove_linked_keg_record if keg.linked?
+
+    begin
+      keg.link
+    rescue Keg::ConflictError => e
+      onoe "Error while executing `brew link` step on #{newname}"
+      puts e
+      puts
+      puts "Possible conflicting files are:"
+      mode = OpenStruct.new(:dry_run => true, :overwrite => true)
+      keg.link(mode)
+      raise
+    rescue Keg::LinkError => e
+      onoe "Error while linking"
+      puts e
+      puts
+      puts "You can try again using:"
+      puts "  brew link #{formula.name}"
+    rescue Exception => e
+      onoe "An unexpected error occurred during linking"
+      puts e
+      puts e.backtrace
+      ignore_interrupts { keg.unlink }
+      raise e
+    end
+  end
+
+  # Link keg to opt if it was linked before migrating.
+  def link_oldname_opt
+    if old_opt_record
+      old_opt_record.delete if old_opt_record.symlink? || old_opt_record.exist?
+      old_opt_record.make_relative_symlink(formula.installed_prefix)
+    end
+  end
+
+  # After migtaion every INSTALL_RECEIPT.json has wrong path to the formula
+  # so we must update INSTALL_RECEIPTs
+  def update_tabs
+    new_tabs = newpath.subdirs.map { |d| Tab.for_keg(Keg.new(d)) }
+    new_tabs.each do |tab|
+      tab.source["path"] = formula.path.to_s if tab.source["path"]
+      tab.write
+    end
+  end
+
+  # Remove opt/oldname link if it belongs to newname.
+  def unlink_oldname_opt
+    return unless old_opt_record
+    if old_opt_record.symlink? && formula.installed_prefix.exist? \
+              && formula.installed_prefix.realpath == old_opt_record.realpath
+      old_opt_record.unlink
+      old_opt_record.parent.rmdir_if_possible
+    end
+  end
+
+  # Remove oldpath if it exists
+  def link_oldname_cellar
+    oldpath.delete if oldpath.symlink? || oldpath.exist?
+    oldpath.make_relative_symlink(formula.rack)
+  end
+
+  # Remove Cellar/oldname link if it belongs to newname.
+  def unlink_oldname_cellar
+    if (oldpath.symlink? && !oldpath.exist?) || (oldpath.symlink? \
+          && formula.rack.exist? && formula.rack.realpath == oldpath.realpath)
+      oldpath.unlink
+    end
+  end
+
+  # Backup everything if errors occured while migrating.
+  def backup_oldname
+    unlink_oldname_opt
+    unlink_oldname_cellar
+    backup_oldname_cellar
+    backup_old_tabs
+
+    if pinned? && !old_pin_record.symlink?
+      src_oldname = old_pin_record.dirname.join(old_pin_link_record).expand_path
+      old_pin_record.make_relative_symlink(src_oldname)
+      new_pin_record.delete
+    end
+
+    if newpath.exist?
+      newpath.subdirs.each do |d|
+        newname_keg = Keg.new(d)
+        newname_keg.unlink
+        newname_keg.uninstall
+      end
+    end
+
+    if oldkeg_linked?
+      begin
+        # The keg used to be linked  and when we backup everything we restore
+        # Cellar/oldname, the target also gets restored, so we are able to
+        # create a keg using its old path
+        keg = Keg.new(Pathname.new(oldkeg.to_s))
+        keg.link
+      rescue Keg::LinkError
+        keg.unlink
+        raise
+      rescue Keg::AlreadyLinkedError
+        keg.unlink
+        retry
+      end
+    end
+  end
+
+  def backup_oldname_cellar
+    unless oldpath.exist?
+      FileUtils.mv(newpath, oldpath)
+    end
+  end
+
+  def backup_old_tabs
+    old_tabs.each(&:write)
+  end
+end
--- a/Library/Homebrew/tap.rb
+++ b/Library/Homebrew/tap.rb
@@ -1,3 +1,5 @@
+require "utils/json"
+
 # a {Tap} is used to extend the formulae provided by Homebrew core.
 # Usually, it's synced with a remote git repository. And it's likely
 # a Github repository with the name of `user/homebrew-repo`. In such
@@ -135,6 +137,15 @@ class Tap
       "command_files" => command_files.map(&:to_s),
       "pinned" => pinned?
     }
+  end
+
+  # Hash with tap formula renames
+  def formula_renames
+    @formula_renames ||= if (rename_file = path/"formula_renames.json").file?
+      Utils::JSON.load(rename_file.read)
+    else
+      {}
+    end
   end
 
   def self.each
--- a/share/doc/homebrew/Rename-A-Formula.md
+++ b/share/doc/homebrew/Rename-A-Formula.md
@@ -1,0 +1,26 @@
+# Renaming a Formula
+
+Sometimes software and formulae need to be renamed. To rename core formula
+you need:
+
+1. Rename formula file and its class to new formula. New name must meet all the rules of naming. Fix any test failures that may occur due to the stricter requirements for new formula than existing formula (e.g. brew audit --strict must pass for that formula).
+
+2. Create a pull request to the main repository deleting the formula file, adding new formula file and also add it to `Library/Homebrew/formula_renames.rb` with a commit message like `rename: ack -> newack`
+
+To rename tap formula you need to follow the same steps, but add formula to `formula_renames.json` in the root of your tap. You don't need to change `Library/Homebrew/formula_renames.rb`, because that file is for core formulae only. Use canonical name (e.g. `ack` instead of `user/repo/ack`).
+
+`Library/Homebrew/formula_renames.rb` example for core formula renames:
+
+```ruby
+FORMULA_RENAMES = {
+  "ack" => "newack"
+}
+```
+
+`formula_renames.json` example for tap:
+
+```json
+{
+  "ack": "newack"
+}
+```
--- a/Library/Aliases/caf
+++ b/Library/Aliases/caf
@@ -1,1 +1,0 @@
-../Formula/libcppa.rb
--- a/Library/Formula/caf.rb
+++ b/Library/Formula/caf.rb
@@ -1,6 +1,5 @@
-class Libcppa < Formula
-  # TODO: since libcppa has been renamed to CAF, this formula should eventually
-  # be renamed to 'caf.rb'.
+class Caf < Formula
+  # Renamed from libccpa
   desc "Implementation of the Actor Model for C++"
   homepage "http://actor-framework.org/"
   url "https://github.com/actor-framework/actor-framework/archive/0.14.0.tar.gz"
@@ -15,12 +14,12 @@ class Libcppa < Formula
     sha256 "0ca6c6a6cae249472146a938458464b275b8d409a81ab9ffa36f22c7523d365a" => :mountain_lion
   end
 
-  depends_on "cmake" => :build
-
   needs :cxx11
 
   option "with-opencl", "build with support for OpenCL actors"
   option "without-check", "skip unit tests (not recommended)"
+
+  depends_on "cmake" => :build
 
   def install
     args = %W[./configure --prefix=#{prefix} --no-examples --build-static]
--- a/Library/Homebrew/cmd/switch.rb
+++ b/Library/Homebrew/cmd/switch.rb
@@ -12,8 +12,7 @@ module Homebrew
     name = ARGV.shift
     version = ARGV.shift
 
-    canonical_name = Formulary.canonical_name(name)
-    rack = HOMEBREW_CELLAR.join(canonical_name)
+    rack = Formulary.to_rack(name)
 
     unless rack.directory?
       onoe "#{name} not found in the Cellar."
--- a/Library/Homebrew/formula_renames.rb
+++ b/Library/Homebrew/formula_renames.rb
@@ -1,2 +1,3 @@
 FORMULA_RENAMES = {
+  "libcppa" => "caf"
 }
--- a/Library/Homebrew/formula_renames.rb
+++ b/Library/Homebrew/formula_renames.rb
@@ -1,0 +1,2 @@
+FORMULA_RENAMES = {
+}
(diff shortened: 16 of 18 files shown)


asparagui:
:tada: This has been a pain point for a long, long time. I like the explanation of your approach. :)

vladshablinsky:
Thanks for feedback @asparagu!

xu-cheng:
One small suggestion, I think `brew audit` should warn if a new formula has the name of old name in `FORMULA_RENAMES`.

vladshablinsky:
Thanks for the comment @xu-cheng! 
Something like we create formula manually without `brew create`? 
```ruby
# formula_renames.rb

FORMULA_RENAMES = {
  "oldname" = > "newname"
}
```
and we want to create `oldname`?
I think we can also forbid creating `oldname` using `brew create` if
 ```ruby
FORMULA_RENAMES.keys.include? oldname
```
Did I understand you correctly?

**Comment by xu-cheng** [comment]
Actually I mean add a new rule into `brew audit` [here](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/audit.rb#L214-L244).

**Comment by vladshablinsky** [comment]
@xu-cheng done.

**Comment by mikemcquaid** [comment]
I just wanted to repeat what I've said to another student here for full disclosure:

We're getting closer to the end of the program so it might be worth focusing on one thing (e.g. the current scope of this) for now to ensure we have one merged PR for the end of the program. I'd personally consider (and unfortunately argue) for a program fail if there's no merged PR for the end of the program.

That said, this PR is looking pretty good as-is to get merged by the end of the program.

**Comment by vladshablinsky** [comment]
Seems everything is ready to be merged now.

File: Library/Homebrew/cmd/audit.rb [hunk 1]
```diff
@@ -229,6 +230,11 @@ def audit_formula_name
       return
     end
 
+    if !!FORMULA_RENAMES[name]
+      problem "Formula name conflicts with renamed formula old name."
```
dunn: This might be more informative as something like `\"#{name}\" is reserved as the old name of #{FORMULA_RENAMES[name]}`.

xu-cheng:
I would like to ask for several days to review this PR. Thanks.

vladshablinsky:
Thanks, @dunn, changed as you suggested.

@xu-cheng, sure.

Will add commits after `rename: libcppa -> caf` and when everybody is happy I'll probably rebase them to make `rename: libcppa -> caf` the last one.

Library/Homebrew/cmd/audit.rb
```diff
@@ -229,6 +230,11 @@ def audit_formula_name
       return
     end
 
+    if !!FORMULA_RENAMES[name]
```
xu-cheng: `FORMULA_RENAMES.key? name`

Library/Homebrew/tap.rb
```diff
@@ -112,6 +112,16 @@ def to_hash
     }
   end
 
+  def load_tap_formula_renames
+    begin
+      path.cd do
+        load "tap_formula_renames.rb"
```
xu-cheng: I suggest another name such as `.tap_formula_renames` so it wouldn't be treated as formula. I know you have excluded it from `Tap#formula_files`, but considering other existing tools out there it would be best to distinguish them in file name at first place.

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -9,15 +10,41 @@ def uninstall
       ARGV.kegs.each do |keg|
         keg.lock do
           puts "Uninstalling #{keg}... (#{keg.abv})"
+
+          # Find old name and old name symlinks for formula and remove them
+          begin
+            formula = Formulary.from_rack(keg.rack)
+            if formula.oldname
+              old_opt = HOMEBREW_PREFIX/"opt"/"#{formula.oldname}"
+              old_cellar = HOMEBREW_CELLAR/"#{formula.oldname}"
```
xu-cheng: How about old `LinkedKegs`?

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -9,15 +10,41 @@ def uninstall
       ARGV.kegs.each do |keg|
         keg.lock do
           puts "Uninstalling #{keg}... (#{keg.abv})"
+
+          # Find old name and old name symlinks for formula and remove them
+          begin
+            formula = Formulary.from_rack(keg.rack)
```
xu-cheng: Do we need to load formula at all. Can we do something like:

```ruby
HOMEBREW_CELLAR.subdirs.select { |keg| keg.symlink? && keg.resloved_path.to_s == keg.to_s }
```

This way seems to be more robust.

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -26,6 +53,36 @@ def uninstall
         name = Formulary.canonical_name(name)
         rack = HOMEBREW_CELLAR/name
 
+        # Cellar/oldname can be a symlink to Cellar/newname so in order not to
+        # get newname's keg as oldname's keg we check if it's a symlink.
+        if rack.symlink?
+          raise "#{rack.to_s} is a symlink"
+        end
```
xu-cheng: I don't think it's a good idea to uninstall half of files and abort. Let's either raise at the very beginning or finish the uninstall.

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -26,6 +53,36 @@ def uninstall
...
+
+        # Since we delete every keg, opt/oldname and Cellar/oldname symlinks
+        # must be removed.
+        begin
+          formula = Formulary.from_rack(rack)
```
xu-cheng: Again, I think we can avoid load formula completely. So we can have a more robust user experience.

Library/Homebrew/dependency.rb
```diff
@@ -29,7 +29,31 @@ def hash
   end
 
   def to_formula
-    formula = Formulary.factory(name)
```
xu-cheng: I don't think you should change this method. There are more underlying problems here regarding dependency resolution. (See https://github.com/Homebrew/homebrew/pull/40991#issuecomment-116948161 and https://github.com/Homebrew/homebrew/issues/14089#issuecomment-39590767).

What you need to do is updating method [`TapLoader#initialize `](https://github.com/Homebrew/homebrew/blob/9124253ae9866684f171bffebd761714d68d2a0f/Library/Homebrew/formulary.rb#L145) and [`Formulary#load_ref` (after searching alias)](https://github.com/Homebrew/homebrew/blob/9124253ae9866684f171bffebd761714d68d2a0f/Library/Homebrew/formulary.rb#L228).  The idea is to trying old name if formula is not found when invoking `Formulary#factory`.

Library/Homebrew/extend/ARGV.rb
```diff
@@ -37,6 +44,10 @@ def kegs
     @kegs ||= downcased_unique_named.collect do |name|
       canonical_name = Formulary.canonical_name(name)
       rack = HOMEBREW_CELLAR/canonical_name
+
+      # Cellar/oldname can be a symlink to Cellar/newname, so in order not to # get newname's keg as oldname's keg we check if it's a symlink.
+      raise "#{rack.to_s} is a symlink" if rack.symlink?
```
xu-cheng: When will this happen? Noted rack is located using `canonical_name`.

Library/Homebrew/formulary.rb
```diff
@@ -190,6 +191,38 @@ def self.from_rack(rack, spec=:stable)
     end
   end
 
+  # Return a Formula instance for oldname
+  def self.from_oldname(oldname)
```
xu-cheng: Referring to [my pervious comment](https://github.com/Homebrew/homebrew/pull/41006/files#r35743574), I think we don't need this method. It's the logic inside `Formulary#factory` you need to alter.

Library/Homebrew/extend/ARGV.rb
```diff
@@ -16,6 +16,13 @@ def formulae
     @formulae ||= (downcased_unique_named - casks).map { |name| Formulary.factory(name, spec) }
   end
 
+  def renamed_formulae
```
xu-cheng: I think we can remove this method as well.

Library/Homebrew/cmd/uninstall.rb
vladshablinsky: Why half of files? It doesn't uninstall anything if rack is a symlink. The same functionality was added to ARGV, so uninstall doesn't uninstall half of files.

But yes, maybe it worth raising at the very beginning:

```ruby
ARGV.named.each do |name|
  if HOMEBREW_CELLAR.join(name).symlink?
    raise "#{HOMEBREW_CELLAR/name} is a symlink"
  end
end
```

Library/Homebrew/cmd/uninstall.rb
vladshablinsky: OK, the formula here is used to get an old name from it's method. If we don't load the formula, we will need to look into the `FORMULA_RENAMES` and `TAP_FORMULA_RENAMES` to get oldnames. Or maybe I can add oldname field to `INSTALL_RECEIPT` and get oldname from tab.

Library/Homebrew/extend/ARGV.rb
vladshablinsky: This happens on uninstall when we're trying to uninstall something that is a symlink.

Library/Homebrew/dependency.rb
vladshablinsky: I'll update these methods, thanks. There were two ways to handle this thing about renames:
  1. Updating `Formulary`.
  2. Changing  `Dependency#to_formula` behaviour.

I discussed that problem with @mikemcquaid after I had implemented it and we decided to change it according to other maintainer's feedback if necessary.

Library/Homebrew/formulary.rb
vladshablinsky: OK, I'll change `Formulary#factory` logic and remove it's not needed.

Library/Homebrew/cmd/uninstall.rb
vladshablinsky: There is no entry in `LinkedKegs`, because `LinkedKegs` are used only for installed formulae and linked formula. When we rename software and perform migrations, we unlink the keg and uninstall it.

Library/Homebrew/cmd/uninstall.rb
vladshablinsky: I need formula to get `oldname`, so I either look into `FORMULA_RENAMES` and `TAP_FORMULA_RENAMES` or instantiate a formula or make `INSTALL_RECEIPT` store the oldname.

Library/Homebrew/tap.rb
vladshablinsky: I used `tap_formula_renames.rb` because the name is pretty similar to `formula_renames.rb` in core repository.

Library/Homebrew/cmd/uninstall.rb
xu-cheng: > I need formula to get old name

You can get this information directly from files system. If `Cellar/foo` is a symbol link to `Cellar/bar`, then foo is an old name.

Library/Homebrew/cmd/uninstall.rb
xu-cheng: > Why half of files? It doesn't uninstall anything

Sorry, I miss-read the codes. 

Actually, I think you don't need to check the symlink. Noted `name = Formulary.canonical_name(name)`, if you change the `Formulary` logic as we discussed, `canonical_name` should return the proper name(i.e. not old name).

Library/Homebrew/cmd/uninstall.rb
vladshablinsky: Yes, but how do I know, that it's `foo` linked to `bar`, when I have only `bar`. I need to check the whole Cellar to find that link or somehow find files that are links to `bar`.

vladshablinsky: So when I uninstall `bar` I have only bar and have no information about `foo`.

xu-cheng: > I need to check the whole Cellar to find that link or somehow find files that are links to bar.

Given there won't be too many symbol links. I think it's acceptable. Part of Homebrew's design choice is that it can work without formula(e.g. we support self build package and use `brew link` etc to manage them). IMO, using file system would be a much robust approach.

Library/Homebrew/cmd/uninstall.rb
vladshablinsky: My only fear is that we can refer to `newname` using `oldname` everywhere we instatiate a formula. There are a lot of such places in the code. With previous implementation we could get `newname` formula by its name *ONLY* in `Dependency#to_formula` and everywhere we called `from_oldname`.

Library/Homebrew/tap.rb
xu-cheng: Let's use different name. It's easier, robust and more friendly to existing tools/scripts.

Library/Homebrew/cmd/uninstall.rb
xu-cheng: I would like to treat old name as an alias of the formula. And `canonical_name` should always return **canonical** name(i.e. not old name).

I don't want to change `Dependency#to_formula` method because there are much deeper problems should be addressing. By altering that method, you make existing problems more complicate.

mikemcquaid: We explicitly don't want the old name to be an alias because we want to be able to rename formulae and replace them with other things.

xu-cheng: > because we want to be able to rename formulae and replace them with other things.

I don't think this is contrary to treating it alias. Because we only resolve the alias after no formula is found, that is if a new formula occupy this name the old name alias will be shadowed.

mikemcquaid: :ok:

vladshablinsky:
Fixed everything according to the comments. Probably some bugs appeared, will test today-tomorrow.

Would be great to have some feedback on `cmd/uninstall`.

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -9,15 +10,39 @@ def uninstall
       ARGV.kegs.each do |keg|
         keg.lock do
           puts "Uninstalling #{keg}... (#{keg.abv})"
+
+          # Find symlinks that can point to keg.rack
+          links = HOMEBREW_CELLAR.subdirs.select { |d| d.symlink? }
```
xu-cheng: We can move it outside of `ARGV.kegs.each` so you don't need to check it repeatedly and can avoid the repeat codes in below. Also you can just do `.select(&:symlink?)` instead of block.

Library/Homebrew/tap.rb
```diff
@@ -112,6 +112,16 @@ def to_hash
     }
   end
 
+  def load_tap_formula_renames
```
xu-cheng: There are several problems in this method. Since you store the information in a const `TAP_FORMULA_RENAME`, every time you load it, it will overwrite other taps' rename information. At the same, Ruby will generate warning as you are reassigning a const value. 

Also why do we have to store it in a ruby file? why not simple JSON which it's fast, reliable and more secure. 

So I would suggest you to implement it as:

```ruby
def formule_renames
  @formula_renames ||= if (rename_file = path/"formula_rename.json").file?
    Utils::JSON.load(rename_file.read)
  else
    {}
  end
end
```

vladshablinsky: Might have sense, thanks.
I used the same structure as `tap_migrations.rb` has. Btw, I know about that warnings and thought it doesn't really matter.

Library/Homebrew/formulary.rb
```diff
@@ -144,6 +145,14 @@ def initialize tapped_name
       path = @tap.formula_files.detect { |file| file.basename(".rb").to_s == name }
       path ||= @tap.path/"#{name}.rb"
 
+      unless path.exist?
+        if tap.load_tap_formula_renames && TAP_FORMULA_RENAMES[name]
+          name = TAP_FORMULA_RENAMES[name]
+          path = @tap.formula_files.detect { |file| file.basename(".rb").to_s == name }
+          path ||= @tap.path/"#{name}.rb"
+        end
+      end
```
xu-cheng: We can do this in a more simple manner:
```ruby
name = @tap.formua_renames.fetch(name, name)
path = @tap.formula_files.detect { |file| file.basename(".rb").to_s == name }
path ||= @tap.path/"#{name}.rb"
```

vladshablinsky:
A question about `from_oldname` and `ARGV#renamed_formulae`: before I rewrote `Formulary#loader_for` I had been able to `brew migrate ack` that had been installed from tap and then renamed. So I had been able to search through old names only. Now if I have `ack` installed from core and another `ack` from tap gets renamed I can't `brew migrate` it using just `ack`, I have to use fully-qualified name and I don't get any information about it if I try to `brew migrate ack`.

I think that `from_oldname` and `renamed_formulae` methods are still needed in order not to get formula with `<somename>` but ONLY get the formulae with `<somename>` as old name.

xu-cheng:
Do you mean the case where both core ack and tap ack are installed. Actually I'm not sure we should support such case. And we definitely don't want to support if one of them need to rename. There are too many edge cases here which will eventually make supporting painful. 

The reason I think you should remove from_oldname method is we support referring to a formula in many ways. Not only name and full-qualified name, but also alias, formula file path, bottle files and URL. 

What you need to use inside migrate command is `ARGV.resolved_formulae` which loads formula based on current installed files.

vladshablinsky:
Not exactly. I mean that there can exist `ack` and `user/repo/ack`, but installed one is `user/repo/ack`. Then `user/repo/ack` gets renamed and for some reason we need to migrate it manually (i.e. using `brew migrate ack` command).

Ah, yes, there is `resolved_formulae`, that can do the same as `from_oldname` in `migrate`. I used `from_oldname` only in `migrate`, I suppose now there is no need in it, so I remove it.

mikemcquaid:
so with canonical name resolution I think we want this strategy:

1. if you pass `oldname`: is there anything installed named `oldname` (that isn't a symlink to a `newname`)? if so, resolve it as `oldname`. 
2. is there an `oldname` symlink to `newname` and the corresponding migration defined? if so, resolve as `newname`.
3. is there an `oldname` formula? is so, resolve as `oldname`.
4. is there's a corresponding migration defined: resolve as `newname`.
5. if none of the above: fail to resolve `oldname`

@xu-cheng thoughts?

Library/Homebrew/cmd/update.rb
```diff
@@ -193,6 +247,7 @@ def report
 
         next unless File.extname(dst) == ".rb"
         next unless paths.any? { |p| File.dirname(p) == formula_directory }
+        next if src == ".tap_formula_renames"
```
xu-cheng: This line can be removed.

Library/Homebrew/cmd/update.rb
```diff
@@ -115,6 +163,12 @@ def load_tap_migrations
   rescue LoadError
     false
   end
+
+  def load_formula_renames
+    load 'formula_renames.rb'
```
xu-cheng: `load` -> `require`

Library/Homebrew/formulary.rb
```diff
@@ -226,6 +228,13 @@ def self.loader_for(ref)
       return AliasLoader.new(possible_alias)
     end
 
+    if newref = FORMULA_RENAMES[ref]
+      formula_with_that_oldname = core_path(newref)
+      if formula_with_that_oldname.file?
+        return FormulaLoader.new(newref, formula_with_that_oldname)
+      end
+    end
```
xu-cheng: Put these logic below tap formulae resolution.

Library/Homebrew/formulary.rb
```diff
@@ -233,6 +242,17 @@ def self.loader_for(ref)
       return FormulaLoader.new(ref, possible_tap_formulae.first)
     end
 
+    newnames = []
```
xu-cheng: Please use a better variable name like `possible_tap_oldname_formulae`

Library/Homebrew/exceptions.rb
```diff
@@ -86,6 +86,26 @@ def initialize name, paths
   end
 end
 
+class TapFormulaWithOldnameAmbiguityError < RuntimeError
+  attr_reader :name, :newnames, :taps
+
+  def initialize name, newnames
```
xu-cheng: Can you use a better variable name than `newnames` which is confusing.

share/doc/homebrew/Rename-A-Formula.md
```diff
@@ -0,0 +1,26 @@
...
+
+`Library/Homebrew/formula_renames.rb` example for core formula renames:
+
+```ruby
+FORMULA_RENAMES = {
+  "ack" => "newack"
+}
+```
+
+`tap_formula_renames.rb` example for tap:
```
xu-cheng: `rb` -> `json`

Library/Homebrew/cmd/update.rb
vladshablinsky: I'll test it again, but it can prevent FORMULA_RENAMES from updloading and we won't get new renamed formula. Hm, this is the same problem we had for `TAP_FORMULA_MIGRATIONS`. FORMULA_RENAMES can be reinitialized.

xu-cheng: Ok, in that case, you can change the `require` in `load_tap_migrations` as `load`.

vladshablinsky: I think there's no need to change it to 'load' because `tap_migrations.rb` is loaded only once and only in `update`.

Library/Homebrew/formulary.rb
vladshablinsky: OK, done.

share/doc/homebrew/Rename-A-Formula.md
```diff
@@ -0,0 +1,26 @@
...
+
+`Library/Homebrew/formula_renames.rb` example for core formula renames:
+
+```ruby
+FORMULA_RENAMES = {
+  "ack" => "newack"
+}
+```
+
+`tap_formula_renames.json` example for tap:
```
vladshablinsky: Oh, it should be `formula_renames.json` here, will change it and squash later with some other commits.

Library/Homebrew/formulary.rb
```diff
@@ -191,6 +193,27 @@ def self.from_rack(rack, spec=:stable)
...
+        end
+      end
+    when HOMEBREW_TAP_FORMULA_REGEX
+      user, repo, name = ref.split("/", 3).map(&:downcase)
+      tap = Tap.new user, repo.sub(/^homebrew-/, "")
+      if (link = (HOMEBREW_CELLAR/name).resolved_path).directory?
+        return basename if tap.formula_renames[name] == link.basename
+      end
+    end
+
```
xu-cheng: I don't think you should change this method. `canonical_name` has completely different semiotical meaning than `Formulary#from_rack` or `ARGV#resolved_formulae`. That is it should return the same name regardless current installation.

vladshablinsky: I implemented it as didn't see any reply to [this comment](https://github.com/Homebrew/homebrew/pull/41006#issuecomment-126702686) and thought you're OK with that.

The problem here is that we need to somehow refer to kegs that e.g. installed from path. If I install `ack` from path and core `ack` gets renamed to `newack` I can't uninstall that installed keg, I can't link/unlink it.

xu-cheng: Sorry, I forget to address that problem in the comment.

To solve your current problem, you should not change the `canonical_name` method. But you need to change the logic inside `ARGV#kegs` and a few similar logics in `uninstall`, `switch` etc. Instead using `canonical_name` directly to find the rack, you can do it in this way:

```ruby
canonical_name = Formulary.canonical_name(name)
rack = HOMEBREW_CELLAR/canonical_name

unless rack.directory?
  if name =~ HOMEBREW_TAP_FORMULA_REGEX
    rack = HOMEBREW_CELLAR/$3
  elsif !name.include("/“)
    rack = HOMEBREW_CELLAR/name
  end
end
```

xu-cheng: I updated last comment a bit.

xu-cheng: In fact we can create a new method `Formulary#to_rack`.
```ruby
def self.to_rack(ref)
  name = canonical_name(ref)
  rack = HOMEBREW_CELLAR/name

  # handle the case when ref is using an old name
  unless rack.directory?
    if ref =~ HOMEBREW_TAP_FORMULA_REGEX
      rack = HOMEBREW_CELLAR/$3
    elsif !ref.include?("/")
      rack = HOMEBREW_CELLAR/ref
    end
  end

  rack
end
```

vladshablinsky:
Thanks for the comments concerning `canonical_name` and idea about `Formulary#to_rack` @xu-cheng , I implemented it as you suggested.

Library/Homebrew/formulary.rb
vladshablinsky: If we put it after tap formula resolution, then `resolved_formula` returns a formula from tap if core installed and core renamed:

* Core `ack` installed
* Core `ack` renamed
* Tap `ack` available
* `brew migrate ack` outputs: ack doesn't replace any formula (it happens because tap ack is instantiated).

mikemcquaid:
@Homebrew/owners I'd like to try and get this merged later this week so if there's any last minute thoughts: please let me know.

Library/Homebrew/manpages/brew.1.md
```diff
@@ -298,6 +298,10 @@ Note that these flags should only appear after a command.
 
     If no <formulae> are given, check all installed brews.
 
+  * `migrate` [<formulae>]:
```
dunn: I haven't studied this PR closely, but it it looks like what this does is check if a rename is registered in `FORMULA_RENAMES`, then perform the necessary migrations?

This seems like how an internal tool might work, not something we're presenting to users.  When I saw there was a `brew migrate` command, I assumed it would work something like `brew migrate <oldname> <newname>`, and would do everything for me.

mikemcquaid: `migrate` is currently like `postinstall` in that users shouldn't need to run it but it's there to help us with debugging user issues. If there are none we may kill it.

dunn: OK, makes sense.

vladshablinsky: And it also can help if migrations of renamed formula fail on `update`. If they fail we might want to migrate packages manually.

Library/Homebrew/cmd/update.rb
mikemcquaid: @xu-cheng Is that :ok:?

Library/Homebrew/formulary.rb
mikemcquaid: @xu-cheng thoughts?

Library/Homebrew/cmd/install.rb
```diff
@@ -69,6 +69,10 @@ def install
           raise "No devel block is defined for #{f.full_name}"
         end
 
+        # TODO
+        # - Fix warning message if we're trying to install to directory in the
+        # Cellar, but it's a symlink to another one made by migrator while
+        # migrating.
```
mikemcquaid: @vladshablinsky is this something you plan on addressing now or merging as-is?

mikemcquaid:
This PR seems pretty solid to me now. I've pulled it and done some local testing and I'm happy with it. I'd like to merge it soon (i.e. this this week) and then address any further issues as they come up and try to migrate some more of these next week: https://github.com/Homebrew/homebrew/issues/14374

Library/Homebrew/cmd/install.rb
vladshablinsky: I tried to fin

Library/Homebrew/cmd/migrate.rb
```diff
@@ -0,0 +1,29 @@
+require "migrator"
+require "formula_renames"
+
+module Homebrew
+  def migrate
+    raise FormulaUnspecifiedError if ARGV.named.empty?
+
+    ARGV.named.each do |name|
```
xu-cheng: Why not use `ARGV#resolved_formula`?

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -1,30 +1,66 @@
...
           rack = keg.rack
           rm_pin rack
+
           if rack.directory?
             versions = rack.subdirs.map(&:basename)
             verb = versions.length == 1 ? "is" : "are"
             puts "#{keg.name} #{versions.join(", ")} #{verb} still installed."
             puts "Remove them all with `brew uninstall --force #{keg.name}`."
+          # If we delete Cellar/newname, then Cellar/oldname symlink
+          # can become broken and we have to remove it.
```
xu-cheng: This comment should be put below `else`, right?

Library/Homebrew/cmd/migrate.rb
vladshablinsky: Because if we install core `oldname` and then it goes to `newname`, but `oldname` from tap stil exists, the wrong formula will be instantiated, because we instantiate formulae this way:

1. name
2. tap name
3. oldname
4. tap oldname

Library/Homebrew/cmd/update.rb
```diff
@@ -45,13 +47,59 @@ def update
...
       install_tap tap_user, tap_repo
+      # update tap for each Tab
+      tabs = dir.subdirs.each.map { |d| Tab.for_keg(Keg.new(d)) }
+      tabs.each { |tab| tab.source["tap"] = "#{tap_user}/homebrew-#{tap_repo}" }
+      tabs.each(&:write)
     end if load_tap_migrations
 
+    # Migrate installed renamed formulae from main Homebrew repository.
+    if load_formula_renames
+      report.select_formula(:D).each do |f|
```
xu-cheng: `f` -> `oldname`

Library/Homebrew/cmd/update.rb
```diff
@@ -45,13 +47,59 @@ def update
...
+        # 1. Check if not empty directory with formula's name exists.
+        # 2. Find symlinked keg for this formula. If no found use first.
+        # 3. Skip if installed formula is not from the main repository.
+        next unless (dir = HOMEBREW_CELLAR/f).directory? && dir.children.length > 0
+        kegs = dir.subdirs.map { |d| Keg.new(d) }
+        keg = kegs.detect { |k| k.linked? || k.optlinked? } || kegs.first
+        tab = Tab.for_keg(keg) if keg
+        next unless tab && ((tap = tab.tap).nil? || tap == "mxcl/master" \
+                            || tap  == "Homebrew/homebrew")
+        next if tab.source["path"] != Formulary.core_path(f).to_s
```
xu-cheng: Why not use `Formulary#from_rack`

Library/Homebrew/cmd/update.rb
xu-cheng: :cool:

Library/Homebrew/cmd/update.rb
vladshablinsky: OK, will change it. I used `f` because there are some other lines in `cmd/update` that use `f`.

xu-cheng:
I haven't reviewed `migrator.rb` yet(And I'm afraid I will only have time to do so in the weekend). Beside this, I would recommend all of GSoC PRs to:
* Run the codes against `brew style` to eliminate style problems.
* Rebase and rewrite the commit messages, so each commit is pointed to a related feature change.

Library/Homebrew/cmd/migrate.rb
xu-cheng: `ARGV#resolved_formulae` resolves formula by current installation first. So it shouldn't matter what order we use to  resolve formula in `Formulary#factory`, right?

vladshablinsky: If name doesn't include "/" then it use `Formulary#from_rack` which use `factory` and then resolves the formula as a tap one.

Library/Homebrew/cmd/update.rb
vladshablinsky: I think, because only core formulae can be instantiated here and it also handles the case when the formula installed from path and the formula with the same name, but from core, moves to new name.

Library/Homebrew/cmd/migrate.rb
xu-cheng: Oh, I see. I think I have a better way to solve it. You can change the logic inside `Formulary#from_rack` in this way:

```ruby
    if tap.nil?
      factory(rack.basename.to_s, spec)
    elsif  tap == "Homebrew/homebrew"
      factory(Fomulary.core_path(rack.basename.to_s), spec)
    else
      factory("#{tap.sub("homebrew-", "")}/#{rack.basename}", spec)
    end
```

Library/Homebrew/cmd/uninstall.rb
vladshablinsky: I put it before the block of code that it describes.

Library/Homebrew/cmd/migrate.rb
xu-cheng: PR #41736 implements such function but without the support of rename.

Library/Homebrew/cmd/update.rb
xu-cheng: I would like to reduce the duplicated codes in the core.

xu-cheng: I think the whole block can be simplified as:
```ruby
f = Formulary.from_rack(HOMEBREW_CELLAR/name)
next unless f.any_version_installed? || !f.core_formula?
```

Library/Homebrew/cmd/update.rb
```diff
@@ -45,13 +47,59 @@ def update
...
+          user, repo, f = f.split("/")
+          next unless user && repo && f
+          next unless user == tap.user && repo == tap.repo
+          rename = tap.formula_renames[f]
+          next unless rename
+          next unless (dir = HOMEBREW_CELLAR/f).directory? && dir.children.length > 0
+          kegs = dir.subdirs.map { |d| Keg.new(d) }
+          keg = kegs.detect { |k| k.linked? || k.optlinked? } || kegs.first
+          next unless keg && Tab.for_keg(keg).tap == "#{user}/homebrew-#{repo}"
+
```
xu-cheng: Same here, you should use `Formulary.from_rack` rather than duplicating the codes of manipulating keg or tab.

Library/Homebrew/cmd/update.rb
vladshablinsky: OK, that makes sense. I think another reason explaining this block is that it had been implemented before the behaviour of `Formulary` was changed.

mikemcquaid:
> I haven't reviewed migrator.rb yet(And I'm afraid I will only have time to do so in the weekend). Beside this, I would recommend all of GSoC PRs to:
> 
> Run the codes against brew style to eliminate style problems.
> Rebase and rewrite the commit messages, so each commit is pointed to a related feature change.

Yup, agreed here. Once all comments have been addressed and you've done these things @vladshablinsky I'll do some final testing and mege.

Library/Homebrew/cmd/update.rb
vladshablinsky: @xu-cheng I think wa can encounter the same issues here as we had for https://github.com/Homebrew/homebrew/pull/41006#discussion_r36408131, so I maybe we can just use `factory("homebrew/homebrew/newname") for core formulae and then rescue `Migrator::DifferentTapsError`.
The same for taps, but use `factory("user/repo/newname")`.

xu-cheng: :+1: LGTM

vladshablinsky: `MigratorDifferentTapsError` is an exception when we try to migrate `name`, but installed `name` is not from the same tap as `name` we want to migrate.

vladshablinsky: :+1:

Library/Homebrew/cmd/install.rb
vladshablinsky: I think we can leave it as it is or remove it at all. I remember I spent some time trying to use more specified warning message, but then I saw that there were some cases when the message wasn't clear either, so I decided to left it as it was and add this TODO.

mikemcquaid: Let's just kill the TODO.

Library/Homebrew/cmd/migrate.rb
```diff
@@ -0,0 +1,22 @@
...
+    raise FormulaUnspecifiedError if ARGV.named.empty?
+
+    # Don't do anything if name is a symlink
+    ARGV.named.each do |name|
+      name = $3 if name =~ HOMEBREW_TAP_FORMULA_REGEX
+      unless (rack = HOMEBREW_CELLAR/name).exist? && !rack.subdirs.empty?
+        raise NoSuchKegError, name
+      end
+      raise "#{rack} is a symlink" if rack.symlink?
+    end
```
xu-cheng: Can you put this block inside the block of `ARGV.resolved_formulae` below?

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -1,30 +1,66 @@
...
 require "formula"
+require "migrator"
 
 module Homebrew
   def uninstall
     raise KegUnspecifiedError if ARGV.named.empty?
 
+    # Find symlinks that can point to keg.rack
+    links = HOMEBREW_CELLAR.subdirs.select(&:symlink?)
+    old_cellars = []
```
xu-cheng: Put `old_cellars = []` inside `keg.lock do` block below.

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -1,30 +1,66 @@
...
+
     if !ARGV.force?
       ARGV.kegs.each do |keg|
         keg.lock do
           puts "Uninstalling #{keg}... (#{keg.abv})"
+
+          # Remove every symlink that links to keg, because it can
+          # be left by migrator
+          links.each do |lnk|
```
xu-cheng: Use `link` instead of `lnk` please.

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -1,30 +1,66 @@
...
+            old_opt = HOMEBREW_PREFIX/"opt/#{lnk.basename}"
+            if lnk.exist? && lnk.realpath == keg.rack.realpath
+              old_cellars << lnk
+            end
+
```
xu-cheng: Put `old_opt = HOMEBREW_PREFIX/"opt/#{lnk.basename}"` here.

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -1,30 +1,66 @@
...
+            if old_opt.symlink? && Keg.new(old_opt.realpath) == keg
```
xu-cheng: `old_opt.realpath == keg.path`

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -1,30 +1,66 @@
...
         end
       end
     else
       ARGV.named.each do |name|
-        name = Formulary.canonical_name(name)
-        rack = HOMEBREW_CELLAR/name
+        rack = Formulary.to_rack(name)
+        name = rack.basename
+
+        links.each do |lnk|
```
xu-cheng: Use `link` instead of `lnk` please.

Library/Homebrew/cmd/update.rb
```diff
@@ -46,13 +48,53 @@ def update
...
+    # Migrate installed renamed formulae from main Homebrew repository.
+    if load_formula_renames
+      report.select_formula(:D).each do |oldname|
+        newname = FORMULA_RENAMES[oldname]
+        next unless newname
+        next unless (dir = HOMEBREW_CELLAR/oldname).directory? && !dir.children.empty?
+        f = Formulary.factory("homebrew/homebrew/#{newname}")
+
+        begin
+          migrator = Migrator.new(f)
```
xu-cheng: `migrator = Migrator.new(Formulary.factory("homebrew/homebrew/#{newname}"))`

Library/Homebrew/cmd/update.rb
```diff
@@ -46,13 +48,53 @@ def update
...
+      tabs.each { |tab| tab.source["tap"] = "#{tap_user}/homebrew-#{tap_repo}" }
+      tabs.each(&:write)
     end if load_tap_migrations
 
+    # Migrate installed renamed formulae from main Homebrew repository.
+    if load_formula_renames
+      report.select_formula(:D).each do |oldname|
+        newname = FORMULA_RENAMES[oldname]
+        next unless newname
+        next unless (dir = HOMEBREW_CELLAR/oldname).directory? && !dir.children.empty?
```
xu-cheng: `dir.children.empty?` -> `dir.subdirs.empty?`

Library/Homebrew/cmd/update.rb
```diff
@@ -46,13 +48,53 @@ def update
...
+          next unless (dir = HOMEBREW_CELLAR/oldname).directory? && !dir.children.empty?
+
+          begin
+            migrator = Migrator.new(Formulary.factory("#{user}/#{repo}/#{newname}"))
+            migrator.migrate
+          rescue Migrator::MigratorDifferentTapsError
+          end
+        end
+      end
+    end
```
xu-cheng: The whole block has very poor performance. You should not repeatedly enumerate `report.select_formula`.

Library/Homebrew/formula.rb
```diff
@@ -223,6 +223,28 @@ def resource(name)
     active_spec.resource(name)
   end
 
+  # An old name for the formula
+  def oldname
+    return @oldname if @oldname
+    if core_formula?
```
xu-cheng: A little nit pick. But you can write it in this way:
```ruby
def oldname
  @oldname ||= if core_formula?
    ...
  elsif tap?
    ...
  end
end
```

Library/Homebrew/cmd/uninstall.rb
vladshablinsky: It is a protected method, can do it `old_opt.realpath.to_s == keg.to_s`.

Library/Homebrew/formula.rb
```diff
@@ -223,6 +223,28 @@ def resource(name)
...
+      begin
+        load "formula_renames.rb"
+      rescue LoadError
+      end
```
xu-cheng: Please remove this and put it at the top of `formula.rb` as `require "formula_renames"`.

Library/Homebrew/formulary.rb
xu-cheng: We don't have any other problem here, right?

Library/Homebrew/cmd/uninstall.rb
xu-cheng: That's OK.

Library/Homebrew/migrator.rb
```diff
@@ -0,0 +1,327 @@
...
+    elsif formula.tap == old_tap
+      true
+    # Homebrew didn't use to update tabs while performing tap-migrations,
+    # so there can be INSTALL_RECEIPT's containing wrong information about
+    # tap (tap is Homebrew/homebrew if installed formula migrates to a tap), so
+    # we check if there is an entry about oldname migrated to tap and if
+    # newname's tap is the same as tap to which oldname migrated, then we
+    # can perform migrations and the taps for oldname and newname are the same.
+    else
+      require "tap_migrations"
```
xu-cheng: Put `require` at top of file.

Library/Homebrew/migrator.rb
```diff
@@ -0,0 +1,327 @@
...
+    @new_pin_record = HOMEBREW_LIBRARY/"PinnedKegs"/newname
+    @pinned = old_pin_record.symlink?
+    @old_pin_link_record = old_pin_record.readlink if @pinned
+  end
+
+  # Check if installed keg is from core formulae and newname is from core.
+  def both_from_core?
+    new_core = formula.core_formula?
+    old_core = (old_tap.nil? || old_tap == "Homebrew/homebrew" \
+      || old_tap == "mxcl/master")
```
xu-cheng: You don't need to handle `mxcl/master`. It's already eliminated inside `Tab#tap`.

Library/Homebrew/migrator.rb
```diff
@@ -0,0 +1,327 @@
...
+        onoe "error occured while migrating."
+        puts "Backuping..."
+        backup_oldname
+      end
+    end
+  end
+
+  # Copy everything from Cellar/oldname to Cellar/newname
+  def copy_to_new_directory
+    puts "Copying to: #{newpath}"
```
xu-cheng: Why copy instead of move?

Library/Homebrew/migrator.rb
```diff
@@ -0,0 +1,327 @@
...
+        copy_to_new_directory
+        repin
+        remove_and_unlink_oldname
+        link_newname
+        link_oldname_opt
+        link_oldname_cellar
+        update_tabs
+      # TODO
+      # - Make some output about exception.
+      rescue Exception
```
xu-cheng: You have to handle interrupt event.

xu-cheng:
Can you give some benchmark of loading formulae (e.g. `brew info <all of core + tap formulae>` before and after this PR?

vladshablinsky:
For `brew info <all of core + tap formulae>` 

Before:
  0.030000   0.040000  44.660000 ( 47.275203)
After:
  0.030000   0.040000  46.100000 ( 50.268483)

vladshablinsky:
@xu-cheng benchmark for `brew update`:

* my remote origin is a local repository
* before benchmarking I left only my test tap, which also is updated from local repository
* I removed all the core formulae and added and then deleted about 500 formulae to my tap
* I used 50 iterations for  `brew update` in benchmark

Before:
  0.010000   0.030000  95.650000 (103.188159)
After:
   0.010000   0.040000 119.290000 (128.448115)

Library/Homebrew/migrator.rb
```diff
@@ -0,0 +1,327 @@
...
+  def both_from_core?
+    new_core = formula.core_formula?
+    old_core = (old_tap.nil? || old_tap == "Homebrew/homebrew" \
+      || old_tap == "mxcl/master")
+    # Formula can be installed from path and we have to check wheter this path
+    # is the same as for core formula.
+    if (path = old_tabs.first.source["path"]) \
+        && path != Formulary.core_path(formula.oldname).to_s
+      old_core = false
+    end
```
xu-cheng: I think you can remove this condition block. Because `Formula#tap` is resolved based on `Formula#path`, it's impossible to have tap is unknown while `Tab#source["path"]` is pointed to a core formula.

Library/Homebrew/migrator.rb
```diff
@@ -0,0 +1,327 @@
...
+    @oldpath = HOMEBREW_CELLAR/formula.oldname
+    raise MigratorNoOldpathError.new(formula) unless oldpath.exist?
+
+    @old_tabs = oldpath.subdirs.each.map { |d| Tab.for_keg(Keg.new(d)) }
+    @old_tap = old_tabs.first.tap
+    raise MigratorDifferentTapsError.new(formula, old_tap) unless from_same_taps?
+
+    @newpath = HOMEBREW_CELLAR/formula.name
+    @oldname = formula.oldname
+    @newname = formula.name
```
xu-cheng: Might want to move these two lines in the beginning.

Library/Homebrew/migrator.rb
```diff
@@ -0,0 +1,327 @@
...
+      tab.source["tap"] = formula.tap
+      tab.write
+    end
+  end
+
+  def from_same_taps?
+    if both_from_core?
+      true
+    elsif formula.tap == old_tap
+      true
```
xu-cheng: I think you can remove `both_from_core?` completely. `formula.tap == old_tap || (formula.core_formula? && old_tap.nil?)` should be the same.

Library/Homebrew/migrator.rb
```diff
@@ -0,0 +1,327 @@
...
+    # Homebrew didn't use to update tabs while performing tap-migrations,
+    # so there can be INSTALL_RECEIPT's containing wrong information about
+    # tap (tap is Homebrew/homebrew if installed formula migrates to a tap), so
+    # we check if there is an entry about oldname migrated to tap and if
+    # newname's tap is the same as tap to which oldname migrated, then we
+    # can perform migrations and the taps for oldname and newname are the same.
+    else
+      require "tap_migrations"
+      if TAP_MIGRATIONS && (rec = TAP_MIGRATIONS[formula.oldname]) \
+          && rec.sub("/", "/homebrew-") == formula.tap
```
xu-cheng: It's kinda string to have `rec.sub("/", "/homebrew-")`. Maybe `rec == formula.tap.sub("homebrew-", "")`?

Library/Homebrew/migrator.rb
```diff
@@ -0,0 +1,327 @@
...
+        link_newname
+        link_oldname_opt
+        link_oldname_cellar
+        update_tabs
+      # TODO
+      # - Make some output about exception.
+      rescue Exception
+        onoe "error occured while migrating."
+        puts "Backuping..."
+        backup_oldname
```
xu-cheng: For rescue, you may need to do thing like `ignore_interrupts { <roll back> }`. So roll back operations cannot be interrupted.

Library/Homebrew/migrator.rb
```diff
@@ -0,0 +1,327 @@
...
+    @pinned
+  end
+
+  def oldkeg_linked?
+    !!oldkeg
+  end
+
+  def migrate
+    if newpath.exist?
+      opoo "#{newpath} already exists; remove it manually and run brew migrate #{oldname}."
```
xu-cheng: `opoo` -> `onoe`. It should be an error, right?

Library/Homebrew/migrator.rb
xu-cheng: It may be better to specific what exception you are expecting rather than just general `Exception`.

xu-cheng:
@vladshablinsky Thanks for the benchmark, it looks like you need to improve the performance of `brew update`. You can achieve it by removing the unnecessary `Tap.each`, like this or so:

```
report.select_formula(:D).each do |oldname|
  user, repo, oldname = oldname.split("/")
  tap = Tap.new user, repo
  next unless newname = tap.formula_renames[oldname]
  ...
end
```

Library/Homebrew/migrator.rb
vladshablinsky: I think any Exception break the whole migration and everything need to be restored after it.

vladshablinsky: @xu-cheng because I always call the same function if error occurs.

And maybe it's a silly question, but how do I handle interrupt event?

Library/Homebrew/migrator.rb
vladshablinsky: Yep, will change it.

Library/Homebrew/migrator.rb
vladshablinsky: Or it's an `ignore_interrupts` method you told about below?

xu-cheng: To catch interrupt, you need to `rescue Interrupt`. (It won't be caught by `Exception`).

`ignore_interrupts` is to prevent a second interruption breaking the restoring process.

> I think any Exception breaks the whole migration and everything need to be restored after it.

If it's possible, you can list all the exceptions similar to [what formula_versions.rb does](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/formula_versions.rb#L3-L8)

Library/Homebrew/cmd/migrate.rb
vladshablinsky: @xu-cheng In `resolved_formulae` we have formulae but not the names passed to the command, can we get the name passed to the command using the formula?

xu-cheng: Isn't the name you need `Formula.oldname`?

vladshablinsky: Ah, sure!

I thought about the case when we pass `newname` it can have `oldname` old name, but we can't know what was passed: `newname` or `oldname` since both of them are resolved as `newname`.

Library/Homebrew/migrator.rb
mikemcquaid: You can `return` here which means the rest of the function doesn't need to be in an `else`

Library/Homebrew/migrator.rb
mikemcquaid: It may just be worth doing `rescue` without naming any exceptions to handle all unconditionally. `formula_versions` names specific exceptions because those are the only ones that occur in those circumstances whereas here we really do want to catch every possible exception.

xu-cheng: Ok, in this case, we can add some debug code here to print the exception so it can benefit future diagnostic.

mikemcquaid: :+1: as long as it doesn't say to report it to us

mikemcquaid: Although another option would be to silently catch it, say you should rerun `brew migrate foo` and then we print the exception out from that.

xu-cheng: Actually I mean print more detail for `ARGV.debug?`

mikemcquaid: :+1:

Library/Homebrew/migrator.rb
vladshablinsky: @xu-cheng I'll wo it this way:

* unlink oldname
* move to new cellar
* uninstall `opt` for oldname

Library/Homebrew/migrator.rb
vladshablinsky: @xu-cheng this is handling of the case when formula installed from path, so it has different from core path and nil tap. In order not to treat package as the one for renamed formula we need to make this check.

Library/Homebrew/migrator.rb
vladshablinsky: https://github.com/Homebrew/homebrew/pull/41006#discussion_r36581939

Library/Homebrew/migrator.rb
xu-cheng: One of reason that you should not use `Tab#source["path"]` is for old bottles and users who have prefix other than `/usr/local` the path compare will definitely return false. After some thinking, I think it's almost impossible for nil tap to be a core formula. Therefore, I think you should check whether they are same tap by `formula.tap == old_tap` and considering nil tap as non core formula.

However, there is still a rare possibility that a nil tap is a core formula. To solve this, I suggest we bypass the tap origin check if `ARGV.force?`

Library/Homebrew/formulary.rb
vladshablinsky: No, I changed everything that used `canonical_name` and could cause problems, so I think everything is OK.

mikemcquaid:
Thanks @vladshablinsky, great work here and @xu-cheng for all the review. @Homebrew/owners feel free to comment on this PR and comments can be addressed in a future PR.

xu-cheng:
Some improvements can be done:

* We should alter the report from `brew update` to have a specific place for formula renames instead of `new formula` + `deleted formula`.
* For migrate, let's make `--force` bypass all of tap check, instead of only nil tap check(which it's a little strange).

vladshablinsky:
> For migrate, let's make --force bypass all of tap check, instead of only nil tap check(which it's a little strange).

We used force for those cases when there is a low possibility, that nil tap is a core formula, however for other formulae there is no need in that. Do you want to allow migrations when two formulae are definitely from different taps? I think it's not safe. What do you think @xu-cheng ?

xu-cheng:
> Do you want to allow migrations when two formulae are definitely from different taps?

Yes, we would expect `--force` is used to bypass the safe check.

Library/Homebrew/cmd/update.rb
```diff
@@ -46,13 +48,47 @@ def update
...
-      next unless (HOMEBREW_CELLAR/f).exist?
+      next unless (dir = HOMEBREW_CELLAR/f).exist?
       migration = TAP_MIGRATIONS[f]
       next unless migration
       tap_user, tap_repo = migration.split "/"
       install_tap tap_user, tap_repo
+      # update tap for each Tab
+      tabs = dir.subdirs.each.map { |d| Tab.for_keg(Keg.new(d)) }
+      tabs.each { |tab| tab.source["tap"] = "#{tap_user}/homebrew-#{tap_repo}" }
+      tabs.each(&:write)
```
xu-cheng: I may find another bug here. You should only update tab if current installation is from a core formula.

Library/Homebrew/migrator.rb
```diff
@@ -0,0 +1,303 @@
...
+      true
+    elsif formula.tap == old_tap
+      true
+    # Homebrew didn't use to update tabs while performing tap-migrations,
+    # so there can be INSTALL_RECEIPT's containing wrong information about
+    # tap (tap is Homebrew/homebrew if installed formula migrates to a tap), so
+    # we check if there is an entry about oldname migrated to tap and if
+    # newname's tap is the same as tap to which oldname migrated, then we
+    # can perform migrations and the taps for oldname and newname are the same.
+    elsif TAP_MIGRATIONS && (rec = TAP_MIGRATIONS[formula.oldname]) \
```
xu-cheng: You need add another check that is `old_tap` has to be a core formula.

vladshablinsky: Fixed.

Library/Homebrew/cmd/update.rb
vladshablinsky: Fixed.

DomT4:
@xu-cheng @mikemcquaid Is it still the case that we can't have a new formula take the old formula's name at any point in time? If so, perhaps worth adding to `brew audit` so we don't have to manually go check every PR?

vladshablinsky:
@DomT4 do you mean [this](https://github.com/vladshablinsky/homebrew/blob/adbee49077d8be43d5e7612011197c67f8942b7e/Library/Homebrew/cmd/audit.rb#L233-L236)?

vladshablinsky:
@xu-cheng so I'm going to just skip that check if `--force` is passed.
Like this:

```ruby
    unless ARGV.force? || from_same_taps?
      raise MigratorDifferentTapsError.new(formula, old_tap)
    end
```

DomT4:
> do you mean this?

I should definitely learn to read :smile_cat:. Thanks @vladshablinsky, Nice work here. Completely awesome to finally have this old Homebrew problem fixed.

vladshablinsky:
Thanks, @DomT4. I would like to thank @xu-cheng and @mikemcquaid too for giving advice and all the comments and reviews.


I'm still working on improvements and stuff with formua renames and will be working on renames from https://github.com/Homebrew/homebrew/issues/14374.
@xu-cheng, I've been thinking about new format for report and actually I don't know how to do everything in a beautiful way:

Here are some thoughts:

1. We can add change `dump` and update a `hash` with `:R` key when it is called, but we need load each tap's `formula_renames.json`.
2. We can easily get all the renames when we perform migrations of renamed formulae, but we can't put the renames into that `report` hash, because it stores paths, but not the names. So I we get the renames, I don't know how to work with them in a right way.
3. We can make `select_formula` return array with both names and paths of formulae and add something like `add_renamed` method to `Report` class, which can update the `hash`.

  In the output we need something like:

  ```
  ==> Renamed Formulae
  oldname -> newname                        user/repo/oldname -> /user/repo/newname
  ```
  so we need to store both oldnames and newnames.
4. Maybe we can store an array of renames in `Report`, which has pairs `oldname, newname` so we can filter formulae returned by `select_formula` in `dump_formula_report`?

Each of these solutions looks strange for me, and they aren't solutions, just some ideas, but I don't have anything better for now. The last one seems the best of them, though.

mikemcquaid:
> the renames into that report hash, because it stores paths, but not the names. So I we get the renames, I don't know how to work with them in a right way.

You can probably get the names from the paths?

> We can add change dump and update a hash with :R key when it is called, but we need load each tap's formula_renames.json.

This seems like it might be the best option :+1:

vladshablinsky:
>You can probably get the names from the paths?

I can, but I can't get path from the name. (I can if it's a core formula, but for tap I need to load tap and have relative path from `HOMEBREW_REPOSOTORY`, is it OK?).

>>We can add change dump and update a hash with :R key when it is called, but we need load each tap's formula_renames.json.

>This seems like it might be the best option :+1:

We can store in a `hash[:R]` array of pairs `[oldname, newname]` then.

PR #41016, opened 2015-06-23
look in HOMEBREW_PREFIX for config

sehrgut:
The global config is installed by etc.install, but the makefile isn't being told the DESTDIR, so it's hardcoding the default config path of /etc/multitail.conf. This is also a bug in the bottled formula, so brewed multitail has no chance of finding its global config.


mikemcquaid:

For future reference the preferred commit message format is `multitail: look in HOMEBREW_PREFIX for config.`.

Issue #34619, opened 2014-12-02
brew cannot download python


ejmolinelli:
Hi all,
This is a two parter!

(1)
I've removed macports and am trying to convert to Brew. I've fixed all of the "brew doctor" warnings related to my installation of R as recommended by [this post] (http://azaleasays.com/2014/08/25/homebrew-warnings-about-unbrewed-dylibs-installed-by-r/). I've backed up all the tcl and tk files that brew was complaining about, and I upgraded my xcode to 4.6.3 on my mac os 10.7.5.  

The only remaining issues from brew doctor are 'unbrewed headers' which I see can be ignored.

I'm trying to run:
``` 
brew tap homebrew/dupes
brew install tcl-tk --with-tk
```
which gives the following error:
==> Installing tcl-tk from homebrew/homebrew-dupes
==> Downloading https://downloads.sourceforge.net/project/tcl/Tcl/8.6.3/tcl8.6.3-src.tar.gz

curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 54
Error: Failed to download resource "tcl-tk"
Download failed: https://downloads.sourceforge.net/project/tcl/Tcl/8.6.3/tcl8.6.3-src.tar.gz

Any thoughts?

(2)
Ultimately, I'm trying to get brew working so i can (re-)install python, which is not working yet. I get the following error with brew install python:

==> Downloading https://www.python.org/ftp/python/2.7.8/Python-2.7.8.tgz
Already downloaded: /Library/Caches/Homebrew/python-2.7.8.tgz
==> Downloading http://bugs.python.org/file30805/issue10910-workaround.txt
Already downloaded: /Library/Caches/Homebrew/python--patch-9926640cb7c8e273e4b451469a2b13d4b9df5ba3.txt
==> Patching
patching file Include/pyport.h
Hunk #1 succeeded at 713 (offset 14 lines).
Hunk #2 succeeded at 736 (offset 14 lines).
Error: integer 4294967294 too big to convert to `int'
Please report this bug:
    http://git.io/brew-troubleshooting
/usr/local/Library/Homebrew/extend/pathname.rb:123:in `atomic_write'
/usr/local/Library/Homebrew/utils/inreplace.rb:26:in `inreplace'
/usr/local/Library/Homebrew/utils/inreplace.rb:14:in `each'
/usr/local/Library/Homebrew/utils/inreplace.rb:14:in `inreplace'
[...]

PR #41017, opened 2015-06-23
Convert incoming unsigned gid/uid to signed, as expected by File.chown()

lvillani:
Process.uid/Process.gid return an unsigned int for nobody/nogroup (-2/-1) but File.chown() expects a signed Fixnum value (which is internally converted back to its two's complement representation)

This should also fix issue #34619

### DIFF
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -107,8 +107,19 @@ class Pathname
         old_stat = default_stat
       end
 
-      uid = Process.uid
-      gid = Process.groups.delete(old_stat.gid) { Process.gid }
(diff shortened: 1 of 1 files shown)


jacknagel:
Is there a ruby bug here? Can it be reported upstream? It seems bad that this is left to the user to sort out.

lvillani:
After a bit more testing I came to the conclusion that this might really be a Ruby bug. The `RangeError` i seemingly triggered only by the use of `Tempfile` though. In both cases, however, when I pass the raw gid to File.chown() or the converted gid to Tempfile.chown() it seems to be ignored, which means that my patch "fixes" the `RangeError` but, in practice, it might be doing nothing.

I'm closing this pull request since it requires more investigation "one level up the stack" :wink:

PR #41021, opened 2015-06-23
removal of CLT/Xcode dependencies for bottles (was #40721)

woodruffw:
Replacement of https://github.com/Homebrew/homebrew/pull/40721

Tasklist:

- [ ] If we're still entertaining the idea of pure Ruby implementations of `otool` and `install_name_tool`, the fact that we've vendored them in the repo itself makes the job of transitioning much easier. I can just write Ruby utilities that behave identically to their cctool equivalents and swap them.
- [x] There are still quite a few parts of Homebrew that run checks against various Xcode/CLT facilities, which need to be branched out so that installing bottles without the CLT doesn't cause those annoying popups or constantly remind you that Xcode is missing. (Completed in Homebrew/homebrew, Homebrew/install stills prompts the user to install Xcode/CLT).
- [x] Now that bottles are installable without Xcode/CLT, the installation process should check the dependency tree before-hand and make sure that all dependencies are installable without Xcode/CLT (i.e. are either bottles or non-building). More specifically, we may want to emit a warning and continue **or** bail out entirely with an error if a non-bottled dependency is encountered.
- [x] I still need to clean up SoftwareSpec and BottleSpec, possibly making those changes suggested by @xu-cheng and @mikemcquaid w/r/t the cellar component.


DomT4:
Old PR for reference: #40721

Is this a Google Summer of Code project or independent of that?

mistydemeo:
It's a GSOC project, yep.

woodruffw:
@DomT4: It's a Google Summer of Code project, sorry if I didn't make that clear.

DomT4:
Cool, Thanks. No worries! I created a `GSoC` label to keep track of all the big projects falling under that umbrella, so just wanted to know how to label the PR :). Really glad someone took this idea up; not needing the developer tools to use Homebrew's bottles is a huge plus.

woodruffw:
Awesome, thanks!

Library/Formula/cctools.rb
```diff
@@ -12,6 +12,7 @@ class Cctools < Formula
   end
 
   bottle do
+    does_not_need_relocation
     cellar :any
```
xu-cheng: Let's make it `cellar :any_skip_relocation`.

asparagui:
Thanks for battling the compliers!  There be :dragon:s there. :3

Library/Homebrew/cmd/install.rb
```diff
@@ -143,7 +145,12 @@ def check_cellar
   def perform_preinstall_checks
     check_ppc
     check_writable_install_location
-    check_xcode
+    if MacOS::Xcode.installed?
+      check_xcode
+    else
+      opoo "You have not installed Xcode."
+      puts "Bottles may install correctly, but builds will fail!"
```
DomT4: It's potentially only worth printing this warning about builds potentially failing *if* a user is building from source. Unless there's something to warn them about, I don't think we need to run a warning on `install` every time.

mistydemeo:
I know you wrote up a list of what remains to be done in the previous issue, but do you mind adding it to the PR here for clarity? If you could mark that up using Github's [task list](https://guides.github.com/features/mastering-markdown/) format that would be great - would make it easy to keep track of what remains to be done and how close we are to this being ready to merge, as well as what the roadmap going forward is.

woodruffw:
Sure!

Copied and reformatted:

- [ ] If we're still entertaining the idea of pure Ruby implementations of `otool` and `install_name_tool`, the fact that we've vendored them in the repo itself makes the job of transitioning much easier. I can just write Ruby utilities that behave identically to their cctool equivalents and swap them.
- [ ] There are still quite a few parts of Homebrew that run checks against various Xcode/CLT facilities, which need to be branched out so that installing bottles without the CLT doesn't cause those annoying popups or constantly remind you that Xcode is missing.
- [ ] Now that bottles are installable without Xcode/CLT, the installation process should check the dependency tree before-hand and make sure that all dependencies are installable without Xcode/CLT (i.e. are either bottles or non-building). 
- [ ] I still need to clean up SoftwareSpec and BottleSpec, possibly making those changes suggested by @xu-cheng and @mikemcquaid w/r/t the cellar component.

I'm currently working on the middle 2, do you want me to check them off as I work on them or as I complete them?

Library/Homebrew/cmd/install.rb
woodruffw: Yeah, that makes sense. I'll remove it as I implement the dependency bottle checking system.

Library/Formula/cctools.rb
woodruffw: I may just remove `does_not_need_relocation` altogether, as that was mainly there for the previous system (hooking in `cctools`). Do you still want me to make this change?

mistydemeo:
As you complete them, please! Thanks very much!

woodruffw:
No problem!

mistydemeo:
If you stick that in the PR description, btw, the PR header will display your progress both here and in the PR search page. It's a handy feature!

woodruffw:
Oh, that's great! I'll move it there right now.

mistydemeo:
Thanks!

Library/Formula/cctools.rb
xu-cheng: If it isn't required now, we can pass this. Thanks.

woodruffw: No problem. I'll remove this section tonight.

Library/Homebrew/cmd/install.rb
DomT4: Awesome, Thanks.

woodruffw:
@jacknagel: Hi! @mistydemeo encouraged me to ping you on an idea I had:

When a user is installing a formula (as a bottle) on a system without build tools (Xcode/CLT), I'm currently checking the formula's dependencies for bottle-completeness and bailing out with an error message if any require building.
The last step is to check the formula itself with `pour_bottle?`, but this creates an ugly case where all dependencies are bottled but the formula itself cannot be poured (a good example is `gcc`), making the user install all dependencies before the error is finally encountered.
I can solve this changing the structure of `FormulaInstaller.install` significantly, but I'd like your go-ahead before I rewrite such a core function.

Thanks! :smile: 
\- William

jacknagel:
Looking at the changes so far, it looks like the only thing being checked for each dependency is `Formula#pour_bottle?`, but to be clear, this isn't enough to guarantee the bottle will be used.

For example, a formula somewhere higher up the tree may be passing options to the dependency, and that needs to be accounted for. Whether a bottle will be used for a particular formula isn’t known until very late:

https://github.com/Homebrew/homebrew/blob/f2f9e487285fb13efab980cb7694b3b846ccbf67/Library/Homebrew/formula_installer.rb#L351-L355

The true problem is that the dependency resolution and installation are very tightly coupled which makes giving a good UX for things like this very hard. This isn't the first enhancement made difficult by the state of this code. I've been working on a rewrite for quite some time.

I'm not sure if that's helpful, other than to reassure you that (a) the existing installation and dependency resolution code is very poor, and is making your task unnecessarily difficult, and (b) at the end of the project, there will probably still be some warts in the UX due because of that, and that's OK. Things like this take time to get right.

woodruffw:
Ah, I understand.

Would it be possible/acceptable to override (conflicting) options passed to dependencies elsewhere in the tree, if only for this particular user case (no Xcode or CLT)?

I can always move the dependency checking downwards if that isn't possible, but it might be good for me to add a system where the user is stopped *before* the installation process if they supply options/flags that make installation impossible with only bottles.

mikemcquaid:
> might be good for me to add a system where the user is stopped before the installation process if they supply options/flags that make installation impossible with only bottles.

This makes sense to me.

woodruffw:
Alright. I'll move forwards with that.

woodruffw:
Looking at `extend/ARGV.rb`, these seem to be the conflicting flags:

*  `--HEAD`
* `--devel`
* `--universal`
* `--32-bit`
* `--build-bottle`
* `--build-from-source`

Are there any other flags I should interrupt the user on?

mikemcquaid:
@woodruffw It's possible for `--devel` to install a bottle, we just don't add them yet.

woodruffw:
A quick update:

I'm currently working on the ruby implementations of what we need from `libmacho`, `install_name_tool`, and `otool`. You can see my progress here: https://github.com/woodruffw/ruby-macho-tools

The three core things needed by Homebrew w/r/t modifying mach-o files are (equivalent commands next to them):
* Listing dynamically linked libraries (`otool -L`)
* Accessing and modifying the "id" of a dylib (`install_name_tool -id`/`otool -L`)
* Changing the install names of dependent dylibs (`install_name_tool -change`)

As of right now, my implementation covers the first bullet entirely like this:

```ruby
f = MachO::MachOFile.new("your_binary")
puts f.linked_dylibs.join("\n")
```

It also partially covers the second bullet:

```ruby
f.dylib_id # => /path/to/your_dylib.dylib
f.dylib_id = "new/path" # implementation not finished
```

I should be finished with the second part of the above by tonight or tomorrow, and I'll begin work on the third bullet.

Some notes:
* I haven't added parsing of "Universal" binaries yet, so only single-architecture files are currently supported. That means x86 and x86_64 mach-o files (PPC is untested).
* I've been developing on ruby 2.x, so it might not run on the 1.8.x tree in its current state. I'll be doing that once I finish all three bullets above.
* You can inspect mach-o files yourself using `info.rb <file>`. The `install_name_tool.rb` and `otool.rb` files are stubs and don't do anything yet.

If you have any questions/comments/concerns, please tell me! Some of the code under `lib/macho.rb` is currently pretty unorganized, but I'm going to be splitting it up into separate files for maintainability soon.

- William

mikemcquaid:
My only concern/thought is that we're getting closer to the end of the program so it might be worth focusing on one thing (e.g. this PR or the Ruby `install_name_tool` and friends) for now to ensure we have one merged PR for the end of the program. I can't speak for @mistydemeo but I'd personally consider (and unfortunately argue) for a program fail if there's no merged PR for the end of the program.

woodruffw:
@mikemcquaid This part of the project is already complete, it just needs to be reviewed and merged. I began work on the other part because it can be mixed into this work at a later date (still within the program timeframe) with ease.
The methods for doing that can either be program stubs in `vendor` (which I've constructed) or inclusion of the full `MachO` module in Homebrew.

mistydemeo:
I'm going try to blitz through a review of this PR tonight.

woodruffw:
Great, thanks!

mikemcquaid:
@woodruffw Ok cool, no worries. There's a few merge conflicts that'll need fixed first and some of the smaller bits here might be able to be moved into separate PRs that are trivially mergeable (e.g. the `sha256` addition to `cctools`).

woodruffw:
@mikemcquaid Sure thing!
I'll ping you and @mistydemeo in the separate PRs and handle the merging conflicts.

Library/Homebrew/cmd/install.rb
```diff
@@ -38,6 +38,16 @@ def install
         end
       end
 
+      # if the user's flags will prevent bottle only-installations when no
+      # developer tools are available, we need to stop them early on
+      if !MacOS.can_build?
+        bf = ARGV.collect_build_flags
+
+        if !bf.empty?
+          raise BuildFlagsError.new(bf)
```
mistydemeo: Nitpick, but make this a one-liner:

```ruby
raise BuildFlagsError.new(bf) if bf.empty?
```

Library/Homebrew/cmd/install.rb
```diff
@@ -154,7 +163,9 @@ def check_cellar
   def perform_preinstall_checks
     check_ppc
     check_writable_install_location
-    check_xcode
+    if MacOS::Xcode.installed?
+      check_xcode
```
mistydemeo: Same here.

Library/Homebrew/formula_installer.rb
```diff
@@ -146,7 +146,15 @@ def install
 
     check_conflicts
 
-    compute_and_install_dependencies unless ignore_deps?
+    if !pour_bottle? && !MacOS.can_build?
+      raise BuildToolsError.new([formula]) unless MacOS.can_build?
```
mistydemeo: There's a duplicate `MacOS.can_build?` check here.

woodruffw:
Fixed all three, thanks!

Library/Homebrew/formula_installer.rb
```diff
@@ -209,20 +217,35 @@ def check_conflicts
...
 
-    if deps.empty? and only_deps?
-      puts "All dependencies for #{formula.full_name} are satisfied."
-    else
-      install_dependencies(deps)
+    deps
+  end
+
+  def check_dependencies_bottled(deps)
+    unbottled = []
```
mistydemeo: Can simplify this by using `#select`:

```ruby
unbottled = deps.select do |dep, _|
  formula = dep.to_formula
  !formula.pour_bottle && !MacOS.can_build?
end

raise BuildToolsError.new(unbottled) unless bottled.empty?
```

mistydemeo:
Just a note when cleaning things up, you may want to organize your commits so that it's easier to squash this down to atomic commits later.

Library/Homebrew/formula_installer.rb
```diff
@@ -311,12 +334,13 @@ def inherited_options_for(dep)
   end
 
   def install_dependencies(deps)
-    if deps.length > 1
+    if deps.empty? && only_deps?
+      puts "All dependencies for #{formula.full_name} are satisfied."
```
mistydemeo: Is this a debug print?

mistydemeo:
I think this is getting pretty close now. How thoroughly has this been tested?

Are you able to rebase this on current master?

Library/Homebrew/cmd/install.rb
```diff
@@ -123,11 +131,10 @@ def check_writable_install_location
   def check_xcode
     checks = Checks.new
     %w[
-      check_for_unsupported_osx
```
mistydemeo: Is there a reason this one got removed?

Library/Homebrew/cmd/reinstall.rb
```diff
@@ -2,6 +2,14 @@
 
 module Homebrew
   def reinstall
+    if !MacOS.can_build?
```
mistydemeo: This condition is also checked inside `FormulaInstaller#install`; does it need to happen here?

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -3,6 +3,14 @@
 
 module Homebrew
   def upgrade
+    if !MacOS.can_build?
```
mistydemeo: Similarly, this also calls `FormulaInstaller`; is the intention for this to bail out earlier than that point?

Library/Homebrew/formula_installer.rb
woodruffw: That print is from the original `compute_and_install_dependencies`, so it wasn't introduced by me. I can take it out if you'd like, though.

Library/Homebrew/formula_installer.rb
woodruffw: Updated!

Library/Homebrew/cmd/install.rb
woodruffw: No, I think that was a mistake during one of my rebases. Should I put it under `check_xcode` or something more general, seeing how `check_xcode` is only executed if Xcode is installed?

Library/Homebrew/cmd/reinstall.rb
woodruffw: The intention is to create an early bail-out with that specific error.

Library/Homebrew/cmd/upgrade.rb
woodruffw: Same as above

woodruffw:
I've tested it fairly thoroughly on my own setup, and I haven't encountered any problems so far. It should definitely be tested on a wider variety of systems, but I don't forsee any issues.
I'll rebase on the current master right now.

woodruffw:
By the way, we can also begin merging my work on the ruby mach-o tools in as well. If we still want to go the stub route, dropping them in `vendor` in place of the binaries should suffice. Otherwise, it might be nice to allow Homebrew developers full programmatic access to the module in which case it might be good to put it under `Library/Homebrew/mach/` or something similar.

Library/Homebrew/formula_installer.rb
mistydemeo: Oh, you're right - sorry about that!

Library/Homebrew/formula_installer.rb
mistydemeo: Thanks! Don't need to define it here before doing the `filter` in this case.

Library/Homebrew/cmd/install.rb
mistydemeo: Leave this one at the top, I think.

Library/Homebrew/cmd/reinstall.rb
mistydemeo: Very similar logic is being repeated in a few places; maybe this could be moved into a class method on `FormulaInstaller`?

Library/Homebrew/cmd/install.rb
woodruffw: Got it, fixed!

Library/Homebrew/cmd/reinstall.rb
woodruffw: Yeah, that's a good idea. I'll create `check_build_flags` in `FormulaInstaller`.

mistydemeo:
> By the way, we can also begin merging my work on the ruby mach-o tools in as well.

I'd like to get this merged first, and then follow up with the Ruby mach-o tools later.

Library/Homebrew/cmd/reinstall.rb
mistydemeo: In this case, maybe the logic can be moved out of `FormulaInstaller#install` and have the install command call it early in the same way that occurs with reinstall and upgrade?

woodruffw:
Sure thing :smile:

Library/Homebrew/cmd/reinstall.rb
woodruffw: I'm not sure which logic you're referring to in `FormulaInstaller#install` - the bottle checking? If so, I think that's possible.

mikemcquaid:
> I'd like to get this merged first, and then follow up with the Ruby mach-o tools later.

@mistydemeo My only issue with that is that the binaries will remain in the Git history forevermore after that point.

mistydemeo:
@mikemcquaid Yeah, that's true. They are fairly small, at least.

Since the Ruby implementations of the tools were originally a "bonus" in the scope of this project, I would like to see us get this finalized before moving on to completing that portion, to make sure that we do have this work completed and merged or mergeable.

Library/Homebrew/requirements/cctools_requirement.rb
```diff
@@ -0,0 +1,8 @@
+class CctoolsRequirement < Requirement
```
mistydemeo: Is this still relevant if the vendored tools are in the repository?

mistydemeo:
Do we still need the `needs_relocation?` part of this PR?

mistydemeo:
Alternately, do we want to remove the vendored tools and focus on allowing users to `brew install cctools` to provide them until the Ruby cctools are ready?

woodruffw:
If we want to avoid the binaries being left in the git history, I can revert back to the original system for providing cctools (hooking it in and looking in `opt`). If so, `needs_relocation?` and `CctoolsRequirement` are still necessary. Otherwise, I can take them out.

mikemcquaid:
> Alternately, do we want to remove the vendored tools and focus on allowing users to brew install cctools to provide them until the Ruby cctools are ready?

I think that would be a great first step :+1:. We can then decide about vendoring or using the Ruby versions depending on how they are looking in a few weeks.

woodruffw:
Just to be clear - does that mean you want the original system or a new one where the user is interrupted and prompted to manually install `cctools`? Either works at this point :smile:

mikemcquaid:
I'm OK with either; is the "original" that it's autoinstalled but not vendored?

woodruffw:
Yep, that's correct.

mikemcquaid:
:+1:

woodruffw:
Great! I think I'll use the original system - prompting the user about something we can fix automatically seems frivolous.

Library/Homebrew/os/mac.rb
```diff
@@ -36,6 +36,26 @@ def locate tool
       end
     end
 
+    def install_name_tool
+      if File.executable?(path = "#{Formula['cctools'].opt_prefix}/bin/install_name_tool")
+        Pathname.new(path)
+      else
+        locate(tool)
```
mistydemeo: There isn't a variable by this name - did you mean `locate("install_name_tool")`?

Library/Homebrew/os/mac.rb
```diff
@@ -36,6 +36,26 @@ def locate tool
...
+      end
+    end
+
+    def otool
+      if File.executable?(path = "#{Formula['cctools'].opt_prefix}/bin/otool")
+        Pathname.new(path)
+      else
+        locate(tool)
```
mistydemeo: Here too.

Library/Homebrew/os/mac.rb
woodruffw: Oops! Corrected :smiley:

mistydemeo:
Looks like there's an error running the tests - are you able to look into this?

woodruffw:
I'm looking into it now. I haven't run into this error while testing on my own system, but I'll trace through it and see if I can find the root problem.

woodruffw:
The problem is stemming from the `Formula['cctools']` call in `Mac#otool`, but I'm not sure why it's failing. `cctools.rb` is definitely present under `Formula/`, and I haven't touched any of the code responsible for resolving formula names. Is it possible there's an bot/testing environment error in this case?

mistydemeo:
The test environment runs in an isolated environment that doesn't have access to core formulae - you won't be able to load any formulae that aren't created within the testing environment.

woodruffw:
Hmm - how do we go about testing this, then? I can continue to test on my own machine, but that's clearly not ideal for making sure the changes work universally. Is it possible to override this test or create a shunt of some kind that imports the `cctools` formula into the testing environment?

mistydemeo:
Instead of instantiating a cctools Formula object, which can be a bit slower anyway, how about just checking the path directly?

```patch
diff --git a/Library/Homebrew/os/mac.rb b/Library/Homebrew/os/mac.rb
index 4618cb9..aa08188 100644
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -37,7 +37,7 @@ module OS
     end
 
     def install_name_tool
-      if File.executable?(path = "#{Formula['cctools'].opt_prefix}/bin/install_name_tool")
+      if File.executable?(path = "#{HOMEBREW_PREFIX}/opt/cctools/bin/install_name_tool")
         Pathname.new(path)
       else
         locate("install_name_tool")
@@ -45,7 +45,7 @@ module OS
     end
 
     def otool
-      if File.executable?(path = "#{Formula['cctools'].opt_prefix}/bin/otool")
+      if File.executable?(path = "#{HOMEBREW_PREFIX}/opt/cctools/bin/otool")
         Pathname.new(path)
       else
         locate("otool")
```

woodruffw:
That sounds good, I'll make that change right now.

mistydemeo:
✨

woodruffw:
:tada:

mistydemeo:
How about a few tweaks to the messaging in the new exceptions:

```patch
diff --git a/Library/Homebrew/exceptions.rb b/Library/Homebrew/exceptions.rb
index 5f0b77c..c716bd7 100644
--- a/Library/Homebrew/exceptions.rb
+++ b/Library/Homebrew/exceptions.rb
@@ -231,12 +231,45 @@ end
 # and are being installed on a system without necessary build tools
 class BuildToolsError < RuntimeError
   def initialize(formulae)
+    if formulae.length > 1
+      formula_text = "formula"
+      package_text = "binary packages"
+    else
+      formula_text = "formulae"
+      package_text = "a binary package"
+    end
+
+    if MacOS.version >= "10.10"
+      xcode_text = <<-EOS.undent
+      To continue, you must install Xcode from the App Store,
+      or the CLT by running:
+        xcode-select --install
+      EOS
+    elsif MacOS.version == "10.9"
+      xcode_text == <<-EOS
+      To continue, you must install Xcode from:
+        https://developer.apple.com/downloads/
+      Or the CLT by running:
+        xcode-select --install
+      EOS
+      EOS
+    elsif MacOS.version >= "10.7"
+      xcode_text = <<-EOS.undent
+      To continue, you must install Xcode or the CLT from:
+        https://developer.apple.com/downloads/
+      EOS
+    else:
+      xcode_text = <<-EOS.undent
+      To continue, you must install Xcode from:
+        https://developer.apple.com/downloads/
+      EOS
+    end
+
     super <<-EOS.undent
-      The following formulae:
+      The following #{formula_text}:
         #{formulae.join(', ')}
-      cannot be installed as a bottle and must be built.
-      To continue, you need to install Xcode or the CLT with:
-        xcode-select --install
+      cannot be installed as #{package_text} and must be built from source.
+      #{xcode_text}
     EOS
   end
 end
@@ -246,13 +279,44 @@ end
 # installation on a system without build tools to fail
 class BuildFlagsError < RuntimeError
   def initialize(flags)
+    if flags.length > 1
+      flag_text = "flags"
+      require_text = "require"
+    else
+      flag_text = "flag"
+      require_text = "requires"
+    end
+
+    if MacOS.version >= "10.10"
+      xcode_text = <<-EOS.undent
+      or install Xcode from the App Store, or the CLT by running:
+        xcode-select --install
+      EOS
+    elsif MacOS.version == "10.9"
+      xcode_text = <<-EOS.undent
+      or install Xcode from:
+        https://developer.apple.com/downloads/
+      Or the CLT by running:
+        xcode-select --install
+      EOS
+    elsif MacOS.version >= "10.7"
+      xcode_text = <<-EOS.undent
+      or install Xcode or the CLT from:
+        https://developer.apple.com/downloads/
+      EOS
+    else:
+      xcode_text = <<-EOS.undent
+      or install Xcode from:
+        https://developer.apple.com/downloads/
+      EOS
+    end
+
     super <<-EOS.undent
-      The following flags:
+      The following #{flag_text}:
         #{flags.join(', ')}
-      require building tools, but none are installed.
-      Either remove the flags to attempt bottle installation, or install
-      Xcode or the CLT:
-        xcode-select --install
+      #{require_text} development tools, but none are installed.
+      Either remove the #{flag_text} to attempt binary package installation,
+      #{xcode_text}
     EOS
   end
 end
```

Specifically this adjusts pluralization, reduces a bit of jargon, and branches the messaging on how to obtain the CLT/Xcode according to the OS version.

mistydemeo:
After this, I think you're ready to squash the commits down to atomic, feature-oriented commits.

woodruffw:
Sounds good! Is there any way to pull your comment off of GitHub and into a patch I can apply? I don't want to take credit for your changes :smile:

mistydemeo:
I'd prefer you squash that into the originating commit, actually! I'm happy to lose credit for that if it means a shorter history. ;)

woodruffw:
Gotcha. I'll add it right now and squash the entire thing down tonight.

mistydemeo:
Thanks! Can you focus your squashing on one commit per atomic feature? It's fine for there to be a number of small commits as long as they're atomic and there aren't bug fix/style commits for changes you made within this branch.

woodruffw:
Sure thing. I just realized that the hooking system is still invoking one of the popup-causing utilities somewhere, but once I fix that I'll squash it down to each component.

woodruffw:
I've traced the popup to a naive call to `MacOS::Xcode.version` in `Superenv.bin`, but I'm not sure how to proceed - is there a default environment that can be safely used if Xcode is not set?

mistydemeo:
Is it the call to `MacOS.clang_version` that's causing it?

woodruffw:
Turns out it was a compound problem - the popup is being caused by `Formula#file_modified?` (which executes `git` with `popen` and hides `stderr`), while the `xcode-select` messages are being caused by the execution of `xcodebuild` in `MacOS::Xcode.uncached_version`.

`MacOS.clang_version` and `MacOS.llvm_build_version` are also being called and may also be producing popups that are being hidden by the toplevel one (`git`), but I won't be able to confirm that until I eliminate the `git` and `xcodebuild` messages. 

I found a way to detect stub binaries (they all link against `/usr/lib/libxcselect.dylib`), so that might be one way to fix this situation. Would adding something like `Metadata#is_stub?` to test programs before executing them be ok?

mistydemeo:
Hmm, right. That's an interesting hitch; at this point there's a relatively baked in assumption that if `/usr/bin/git` exists then it should be usable. You may need to look at Homebrew's git wrapper as well.

mistydemeo:
> Would adding something like Metadata#is_stub? to test programs before executing them be ok?

How do you plan to implement this? It seems we have a bit of a catch-22 here, where you need cctools to determine whether they're stubs, but you can't install cctools without running into the code you're having an issue with here.

woodruffw:
Hmm, that's right. Instead of `is_stub?` (or until I can bring in my mach-o library), I guess I could just guard `file_modified?` by checking *where* `git` is, and stopping it if it's in `/usr/bin` when `MacOS.can_build?` is false.

Library/Homebrew/formula.rb
```diff
@@ -596,7 +596,13 @@ def inspect
   end
 
   def file_modified?
-    return false unless which("git")
+    git_dir = MacOS.locate("git").dirname.to_s
+
+    # /usr/bin/git is a popup stub when Xcode/CLT aren't installed, so bail out
+    if git_dir == "/usr/bin" && !MacOS.can_build?
```
mistydemeo: Nitpick, but make this a one-liner using postfix `if` please!

woodruffw: Sure!

Library/Homebrew/os/mac/xcode.rb
```diff
@@ -74,7 +74,7 @@ def uncached_version
         # This is a separate function as you can't cache the value out of a block
         # if return is used in the middle, which we do many times in here.
 
-        return "0" unless OS.mac?
+        return "0" unless OS.mac? && MacOS.can_build?
```
mistydemeo: Do you mind using `!MacOS::Xcode.installed? || !MacOS::CLT.installed?` instead? `MacOS.can_build?` is pretty unclear in its purpose here.

woodruffw: Sure, changing it now.

mistydemeo:
Are there any unusual consequences of a "0" Xcode version? Linuxbrew does this because things need to believe that Xcode is installed.

woodruffw:
I haven't experienced any unusual consequences so far.
I see that it causes a test to fail - should that be disregarded in this case, or perhaps reworked to only be considered when Xcode is actually installed?

mistydemeo:
Looking at the failures, this is I think exposing some issues with using a "0" version like Linuxbrew does.

Linuxbrew uses "0" so that Linuxbrew thinks Xcode is installed; here we explicitly want to accept the situation where it isn't. We likely want to return `nil` from that version method and, if it's not already handled, ensure that there are appropriate nil guards in places that check Xcode's version to understand it as meaning "not installed".

woodruffw:
Alright. I'll change it to `nil` and test for adverse effects :smile:

woodruffw:
Does Homebrew's implementation of `String#undent` have any known quirks, particularly with interpolated variables?

I'm seeing some strange formatting mangling with the `BuildToolsError` and `BuildFlagsError` output:

bfontaine:
@woodruffw Not that I know of. You can see its implementation here: https://github.com/Homebrew/homebrew/blob/1903c302c850fce9b5241b1991995de057b7e3c8/Library/Homebrew/extend/string.rb#L3

mistydemeo:
Oh yeah, that didn't occur to me. `undent` takes the indent level from the first line; the strings being interpolated in have a different indent level, so actual text is being removed instead of whitespace.

woodruffw:
Hmm, only `gsub`ing for a set of whitespace characters should do the trick:

```ruby
  def undent
    gsub(/^[ \t]{#{(slice(/^ +/) || '').length}}/, '')
  end
```

This seems like the kind of thing I should change in a separate PR. Should I?

mistydemeo:
Yes, please! Can you add a test for this case when you do, too?

mistydemeo:
Hm, actually, looking at the tests this may be expected behaviour:

```ruby
  def test_undent
    undented = <<-EOS.undent
    hi
....my friend over
    there
    EOS
    assert_equal "hi\nmy friend over\nthere\n", undented
  end
```

Can be argued whether it's *good* behaviour but the tests anticipate it.

woodruffw:
Hmm, that's not great for our purposes.
Perhaps a separate `String#unindent` method and a deprecation policy? Alternatively I could just call that `gsub` on our heredocs, but that's not very clear compared to a call like `undent`/`unindent`.

mistydemeo:
I'm not opposed to making that change, just noticing that it's not a "bug" so much as "questionable design".

The other possibility is to adjust the indent levels of the other strings so they all match.

woodruffw:
I'd personally prefer a new `String#unindent` method, if only because adjusting the indentation makes the style inconsistent with other error messages. I won't mark `String#undent` as deprecated, seeing as the behavior is just different and not buggy (as you said).
Does that sound good to you?

mistydemeo:
I don't think a new nearly-identical method is really necessary. If nothing depends on the existing behaviour we can change it (with the test checking for that); otherwise it would be best just to adjust the indentation level of the strings in the exceptions.

woodruffw:
By adjusting the indentation, you mean something like this, right?

```ruby
      xcode_text = <<-EOS.undent
To continue, you must install Xcode from the App Store,
or the CLT by running:
  xcode-select --install
      EOS
```

If so, that works for me. I can change `String#undent`'s behavior at a future point if desired :smile:

mistydemeo:
Yeah, like so:

```ruby
    if MacOS.version >= "10.10"
      xcode_text = <<-EOS
      To continue, you must install Xcode from the App Store,
            or the CLT by running:
              xcode-select --install
      EOS
```

E.g. reduce the indent by two spaces, to match the destination string, and don't call `undent` on the string before it's interpolated.

mistydemeo:
How has your testing of the `nil` Xcode version been going? Seem to work okay?

woodruffw:
I haven't encountered any problems with the `nil` Xcode version yet - it appears to be properly guarded everywhere with checks against Xcode installation first.

mistydemeo:
Excellent!

woodruffw:
Yep! :tada:

mistydemeo:
Great! Ready to clean up the commit history?

woodruffw:
Yeah. How atomic/feature specific do you want it? Down to individual features or more general regions of change?

mistydemeo:
Down to individual features, please! Makes it easier to isolate changes and/or revert specific commits later if needed.

woodruffw:
You got it :smiley:

mistydemeo:
Looks like there are merge conflicts; do you mind rebasing again?

Library/Homebrew/formula_installer.rb
```diff
@@ -398,7 +433,12 @@ def finish
 
     keg = Keg.new(formula.prefix)
     link(keg)
-    fix_install_names(keg) if OS.mac?
+
+    if OS.mac?
+      # this needs to be changed to a test against build_bottle? and
+      # formula.bottle.needs_relocation?
+      fix_install_names(keg) unless formula.name == 'cctools'
```
mistydemeo: Since there's `needs_relocation?` now, can you adjust this?

woodruffw: Adjusted! I also had to add a new guard around the new `keg.relocate_install_names` call in `FormulaInstaller#pour` with similar behavior.

Library/Homebrew/formula_installer.rb
```diff
@@ -625,9 +661,10 @@ def pour
       downloader.stage
     end
 
-    keg = Keg.new(formula.prefix)
-    keg.relocate_install_names Keg::PREFIX_PLACEHOLDER, HOMEBREW_PREFIX.to_s,
-      Keg::CELLAR_PLACEHOLDER, HOMEBREW_CELLAR.to_s, :keg_only => formula.keg_only?
+    if !pour_bottle? || formula.bottle.needs_relocation?
```
xu-cheng: I haven't fully reviewed this PR yet, so I may be wrong. But since I just wrote these lines early today, I have a few questions:

* Why check `pour_bottle?` inside `pour` method? This seems to be unnecessary. 
* From my understanding, here you want to stop relocating bottles which don't need to. But since we relocate all the text files for all bottles, it means all formulae need to relocate, at least for text files(e.g. `INSTALL_RECEIPT.json`)

woodruffw: Oh, you're absolutely right. That `pour_bottle?` is redundant. I'll remove it right now.

Looking at the code in `Keg#relocate_install_names`, it seems as if the only thing being done is changing the install names of binaries and no operations on any text files. Therefore, this new guard shouldn't affect the relocation of any non-binaries. Please correct me if I'm wrong :smile:

xu-cheng: Sorry but you are incorrect. It will relocate both binaries and text files. https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/keg_relocate.rb#L41-L58

Library/Formula/cctools.rb
xu-cheng: If this is still needed, I would like to see it integrated into `cellar` DSL.

Library/Homebrew/formula_installer.rb
```diff
@@ -398,7 +433,8 @@ def finish
 
     keg = Keg.new(formula.prefix)
     link(keg)
-    fix_install_names(keg)
+
+    fix_install_names(keg) if !pour_bottle? || (pour_bottle? && formula.bottle.needs_relocation?)
```
xu-cheng: It can be confusing here. Maybe use `unless @poured_bottle && formula.bottle.skip_relocation?`

Library/Homebrew/formula_installer.rb
woodruffw: Hmm, I see. Any objections to splitting that part of `relocate_install_names` into a separate method, like `relocate_text` or something similar?
Alternatively, I could bring the condition into `relocate_install_names` somehow and only block off that part, but that might be less clear.

Library/Homebrew/formula_installer.rb
woodruffw: That sounds good, thanks!

Library/Formula/cctools.rb
woodruffw: I've integrated this, thanks!

Library/Homebrew/formula_installer.rb
```diff
@@ -398,7 +433,8 @@ def finish
 
     keg = Keg.new(formula.prefix)
     link(keg)
-    fix_install_names(keg)
+
+    fix_install_names(keg) unless @pour_bottle && formula.bottle.skip_relocation?
```
xu-cheng: `poured_bottled`

Library/Homebrew/software_spec.rb
```diff
@@ -270,6 +274,7 @@ def initialize
     @prefix = DEFAULT_PREFIX
     @cellar = DEFAULT_CELLAR
     @collector = BottleCollector.new
+    @does_not_need_relocation = false
```
xu-cheng: Should remove this now. Right?

Library/Homebrew/formula_installer.rb
woodruffw: Oops! Fixed.

Library/Homebrew/software_spec.rb
woodruffw: Yeah. Removed it.

Library/Homebrew/cmd/install.rb
```diff
@@ -121,13 +125,18 @@ def check_writable_install_location
   end
 
   def check_xcode
+    # TODO: reinstate check_for_bad_install_name_tool and check_for_installed_developer_tools
+    # currently check_for_bad_install_name_tool fails because it tries to call
+    # the /usr/bin/otool stub program on systems without XCode/CLT
+    # check_for_installed_developer_tools doesn't fail, but produces a warning
+    # when one is no longer required
```
mikemcquaid: Can kill these comments now, rioght?

Library/Homebrew/exceptions.rb
```diff
@@ -226,6 +226,100 @@ def dump
...
+
+    super <<-EOS.undent
+      The following #{flag_text}:
+        #{flags.join(', ')}
+      #{require_text} building tools, but none are installed.
+      Either remove the #{flag_text} to attempt bottle installation,
+      #{xcode_text}
+    EOS
+  end
+end
```
mikemcquaid: If possible might be good to unify some of this text with that in `brew doctor`

Library/Homebrew/formula_installer.rb
```diff
@@ -55,6 +56,13 @@ def initialize(formula)
     @pour_failed   = false
   end
 
+  # called by install/reinstall/upgrade when no build tools are available
+  def self.check_build_flags
+    bf = ARGV.collect_build_flags
```
mikemcquaid: `build_flags` is a better variable name.

Library/Homebrew/os/mac.rb
```diff
@@ -36,6 +36,26 @@ def locate tool
...
+      if File.executable?(path = "#{HOMEBREW_PREFIX}/opt/cctools/bin/otool")
+        Pathname.new(path)
+      else
+        locate("otool")
+      end
+    end
+
+    def can_build?
+      Xcode.installed? || CLT.installed?
+    end
```
mikemcquaid: Trying to think of a nicer name; `has_apple_developer_tools?`?

Library/Homebrew/os/mac/xcode.rb
```diff
@@ -76,6 +76,8 @@ def uncached_version
 
         return "0" unless OS.mac?
 
+        return nil unless MacOS::Xcode.installed? && MacOS::CLT.installed?
```
mikemcquaid: Mind not using `&&` with `unless`? Maybe just me being stupid but I always find it hard to parse.

mikemcquaid:
Looking good. When we've got a final version I'd like to do some testing before we merge but :+1: on all your work here!

Library/Homebrew/cmd/install.rb
woodruffw: Yep! Removed.

Library/Homebrew/os/mac.rb
woodruffw: Sounds good to me! My only concern is that it might make these suffix conditionals very long - if that isn't a problem, I'll go ahead :smile:

Library/Homebrew/formula_installer.rb
woodruffw: Do you mean for `check_build_flags` or `collect_build_flags`?

woodruffw: Changed!

woodruffw:
Thanks! I can definitely have a version for testing by tomorrow, if that suits you :smiley:

Library/Homebrew/exceptions.rb
woodruffw: Sounds good. I'll look into both adding new checks and which can be removed now that bottle-only installations are possible.

Library/Homebrew/os/mac/xcode.rb
woodruffw: I agree - would simply changing to `if` and inverting each with `!` suffice?

mikemcquaid: Yup :+1:

woodruffw: Changed!

mikemcquaid:
@Homebrew/owners I'd like to try and get something here merged soon (this week or next week) so let us known if you have any more feedback.

Library/Homebrew/formula_installer.rb
```diff
@@ -55,6 +56,13 @@ def initialize(formula)
     @pour_failed   = false
   end
 
+  # called by install/reinstall/upgrade when no build tools are available
+  def self.check_build_flags
```
dunn: `check_build_flags` doesn't seem perfectly descriptive.  It's specifically checking that there are no non-default flags, right?

woodruffw: It specifically checks for flags that would trigger compilation of a formula, rather than pouring. It's called when neither Xcode or the CLT is present, stopping the user before they actually get to formula installation (and are bombarded with unhelpful compilation failures).

woodruffw: @dunn: Any followup thoughts on this? I think `prevent_build_flags` is a good alternative, but I defer to you on this :smile:

dunn: That seems good to me!

woodruffw:
Pinging @mistydemeo and @mikemcquaid: Anything else I can do to help prepare for the merge?
I can rebase again and squash the history down some more if that would help.

I also just wanted to give you an update on the work I've been doing with my Ruby implementation of `install_name_tool` and `otool` functionality - support for "normal" Mach-Os is almost entirely complete, and I'm working on completing support for "fat" binaries right now.

Documentation is pretty spotty at the moment, but I've made it into a gem if you want to test it out:
https://rubygems.org/gems/ruby-macho

The two main things remaining on that front are changing rpaths (which shouldn't be too difficult) and completing fat binary support.

Since we're already pretty settled with this approach to relocating files, I'd like to continue this work beyond GSOC and hopefully introduce it sometime in the near future as a replacement to the system I've created here.

Library/Homebrew/formula_installer.rb
woodruffw: Awesome, I'll make the changes right now.

Library/Homebrew/os/mac.rb
mikemcquaid: Not a problem: go ahead :+1:

mikemcquaid:
> Pinging @mistydemeo and @mikemcquaid: Anything else I can do to help prepare for the merge?
> I can rebase again and squash the history down some more if that would help.

I made one more comment and then I'm good to merge this if you can do a bit more squashing. If it helps what I tend to do is use `git reset origin/master` and then `git gui` (or a similar tool) to stage things up line-by-line, file-by-file. 

> I also just wanted to give you an update on the work I've been doing with my Ruby implementation of install_name_tool and otool functionality - support for "normal" Mach-Os is almost entirely complete, and I'm working on completing support for "fat" binaries right now.

Sounds good. We can maybe just fail out on fat binaries for now and work on getting your work as-is integrated and vendored into the repo. Would be good to have some tests in here, too! None of that will block this PR being merged, though.

mistydemeo:
> I made one more comment and then I'm good to merge this if you can do a bit more squashing.

:+1: I feel good about this now. I haven't had the chance to test this myself thoroughly, but if we mark this as a beta feature we may be able to get community members to test more.

> We can maybe just fail out on fat binaries for now and work on getting your work as-is integrated and vendored into the repo.

If it isn't going to take long to do the additional work, and it seems realistic to do in the near future, I'd like to wait on integrating that until it's feature complete.

Library/Homebrew/os/mac.rb
woodruffw: Great! Applied :fireworks:

woodruffw:
Sounds good to me. I'll squash down tonight.

Adding unit tests is definitely a good idea, but I'm not sure how I would assert our behavior here - checking for correct installation of `cctools`, raising of exceptions when invalid build flags/build-only formulae are passed?

I'm fine with waiting to integrate the ruby macho implementation - it's close to feature complete and will probably be easier to insert without having to worry about gaps in coverage (like fat binaries). It has a test suite of its own, which I can definitely merge into Homebrew's tests as well.

mistydemeo:
> It has a test suite of its own, which I can definitely merge into Homebrew's tests as well.

:+1: That would be fantastic!

Just a reminder, re: scheduling, that we should aim to have code complete by next Monday. That will leave you time to spend the last week making sure that everything is settled and to write documentation.

woodruffw:
That shouldn't be a problem, the only code remaining is tests and whatever cosmetic changes anybody else suggests.

In terms of documentation, do you mean in the source or a writeup in the wiki or something else? I'm happy to document, just point me in the right direction :smile:

mistydemeo:
I'd like to see in-source documentation - API documentation like we have in the `Formula` class, for example. If there's a need for more of an overview in addition to class and function/method documentation, that could be written in separate Markdown files, similar to how the Homebrew documentation is handled. Thanks!

woodruffw:
Sure!

mistydemeo:
Did you have a chance to do the history cleanup you planned?

mistydemeo:
Thanks for adding documentation to the new functions!

woodruffw:
Happy to! Just tell me if there are any formatting mistakes/conventions you'd prefer :smile:

mikemcquaid:
Given Monday is the informal pencils down date I think we need to think about what state we're going to merge this PR in to give other Homebrew maintainers a change to review and test it. The other two GSoC PRs have been merged and I think we need to think about doing so relatively soon on this one too.

woodruffw:
I'm completely prepared to merge today/soon - waiting for your go-ahead!

mikemcquaid:
@woodruffw Cool! What are we doing with tests and the Ruby `install_name_tool`?

woodruffw:
@mikemcquaid The Ruby `install_name_tool` implementation is still a work in progress. It could be merged, but several things `Homebrew` expects from `otool`/`install_name_tool` would have to be stubbed out and merged incrementally. As such, I'd like to get it to a point where all common `install_name_tool` operations work before introducing it. That date will probably be after the GSOC, but I'm still interested in doing it.

As for formal unit tests, I'm not sure how to best approach coverage - since we want to test new features in `FormulaInstaller` primarily, should I be augmenting the `Testball` formula to include a bottle or should I test with a relatively uniform formula like GNU hello?

mikemcquaid:
> @mikemcquaid The Ruby install_name_tool implementation is still a work in progress. It could be merged, but several things Homebrew expects from otool/install_name_tool would have to be stubbed out and merged incrementally. As such, I'd like to get it to a point where all common install_name_tool operations work before introducing it. That date will probably be after the GSOC, but I'm still interested in doing it.

Ok, let's merge this before then.
 
> As for formal unit tests, I'm not sure how to best approach coverage - since we want to test new features in FormulaInstaller primarily, should I be augmenting the Testball formula to include a bottle or should I test with a relatively uniform formula like GNU hello?

Creating a new `Testball` with a bottle included makes sense :+1:

woodruffw:
> Creating a new Testball with a bottle included makes sense :+1:

Awesome. Thoughts on augmenting the current `Testball` versus creating something like `TestballBottle` for isolation?

mikemcquaid:
> Awesome. Thoughts on augmenting the current Testball versus creating something like TestballBottle for isolation?

Create a new one; augmenting the current will require messing with a bunch of existing tests and I can see two being useful outside of this PR too.

woodruffw:
Got it!

mistydemeo:
I have a VM set up for some testing now, and noticed a minor bug. If cctools is already installed, and you `brew install` something with a bottle which requires relocation, it prints something like this:

```
==> Installing dependencies for openssl:
```

e.g. it displays that line even if no deps were added.

Library/Homebrew/test/test_formula_installer_bottle.rb
```diff
@@ -0,0 +1,50 @@
+require "testing_env"
+require "formula"
+require "compat/formula_specialties"
+require "formula_installer"
+require "keg"
+require "testball_bottle"
+
+class InstallBottleTests < Homebrew::TestCase
+	module OS
```
mistydemeo: Instead of manually stubbing this, you can use the [mocha library](https://github.com/freerange/mocha), which is already used as part of the tests. Something ike this should work:

```ruby
def test_something
  MacOS.expects(:has_apple_developer_tools?).returns(false)
  # rest of your test goes here
end
```

mistydemeo:
Looks like `brew --config` still triggers an Xcode installation popup when attempting to determine the CLT/XCode versions, due to `xcrun` being invoked.

mistydemeo:
Ping - how are the last bits coming along?

We'd originally planned the "pencils down" date for yesterday - want to make sure this gets finished soon!

woodruffw:
Pong - sorry for the late response.

The basic tests are in place, I'm just having trouble getting the `TestballBottle` formula to associate the `bottle` DSL with the stable spec. I thought this would work (in `TestballBottle#initialize`):

```ruby
  stable.bottle do
    root_url "file://#{File.expand_path("..", __FILE__)}/bottles"
    sha256 "7cf1d893838bc940beb3671f28a85de8e636489843b272772fbcba4bfb056c4a" => :yosemite
    sha256 "7cf1d893838bc940beb3671f28a85de8e636489843b272772fbcba4bfb056c4a" => :mavericks
    sha256 "7cf1d893838bc940beb3671f28a85de8e636489843b272772fbcba4bfb056c4a" => :mountain_lion
  end
```

But no luck. Any thoughts? This is probably extremely simple and I'm just overthinking it.

woodruffw:
Pong - sorry for the late response.

The basic tests are in place, I'm just having trouble getting the `TestballBottle` formula to associate the `bottle` DSL with the stable spec. I thought this would work (in `TestballBottle#initialize`):

```ruby
  stable.bottle do
    root_url "file://#{File.expand_path("..", __FILE__)}/bottles"
    sha256 "7cf1d893838bc940beb3671f28a85de8e636489843b272772fbcba4bfb056c4a" => :yosemite
    sha256 "7cf1d893838bc940beb3671f28a85de8e636489843b272772fbcba4bfb056c4a" => :mavericks
    sha256 "7cf1d893838bc940beb3671f28a85de8e636489843b272772fbcba4bfb056c4a" => :mountain_lion
  end
```

But no luck. Any thoughts? This is probably extremely simple and I'm just overthinking it.

mikemcquaid:
@woodruffw took a look but not sure. If no-one else knows I'll spend some time on this tomorrow?

woodruffw:
@mikemcquaid: That would be great, thanks!

mistydemeo:
(Sorry, closed by accident)

mistydemeo:
You should be able to fix the issue with the number of expected calls by using `MacOS.stubs` instead of `MacOS.expects`, which stubs it for a nonspecified number of invocations.

mistydemeo:
The test itself is failing due to the lack of the cctools formula; because of that it can't actually pour the bottle.

Library/Homebrew/requirements/cctools_requirement.rb
mikemcquaid: @woodruffw Ping?

Library/Homebrew/exceptions.rb
mikemcquaid: Thoughts on this?

mikemcquaid:
@woodruffw I'm afraid the firm pencils-down date is tomorrow at 8pm. If this isn't in a green and mergeable state with time for reviewers to make any last comments (and you to fix them) I don't think you'll be able to pass.

woodruffw:
Ah, that makes sense. If we can't use the cctools formula in `testing_env`, should we instead test for the absence of the correct behavior (i.e. something like `assert_raises(FormulaUnavailableError) { ... }`)?

mistydemeo:
The formula should not be missing in normal operation so I don't think testing for that is important.

woodruffw:
@mikemcquaid Understood, the only current problem is working around the fact that the `cctools` formula can't be accessed during testing. Do you have any thoughts on the solution above/exposing `cctools` to the testing env somehow?

woodruffw:
@mistydemeo In that case, simply changing it to `cellar :any_skip_relocation` will fix the problem - does that sound good?

Library/Homebrew/exceptions.rb
woodruffw: There are a bunch of checks in `brew doctor` that overlap with the information reported by `BuildToolsError` and `BuildFlagError`, but I'm not sure how I would unify them. If you want me to prune/modify old checks so they now indicate the availability of bottle-only installations, I can do that right now.

mikemcquaid: You can leave it as-is for now or maybe just add a TODO? Not blocking on merge just was commenting on the floating question :grinning:

Library/Homebrew/requirements/cctools_requirement.rb
woodruffw: @mikemcquaid We reverted to the original system of hooking in `cctools`, so this is still relevant.

Library/Homebrew/exceptions.rb
woodruffw: You got it. It would definitely be good to reduce redefinition of methods for different OS Xes, etc. That would probably go nicely with the idea for changing `String.undent` to fix the nesting problem, so I can mark this as TODO for now and open another PR.

mistydemeo:
If it's skipping relocation, I'm not clear what it's testing at this point - is it that a bottle that doesn't require relocation can be poured on a system with no dev tools?

woodruffw:
With `:any_skip_relocation`, yes. That's why I didn't do it initially - it glazes over the actual pulling in of `cctools`. However, I can't think of a way to test relocation *without* somehow introducing `cctools` to the testing environment.

mistydemeo:
OK, makes sense. That is testing new functionality, even if it's not a complete set of tests, so makes sense to retain.

woodruffw:
Great! I'm squashing down the unit test commits now.

mistydemeo:
Bottle pouring is still broken in the test:

```
InstallBottleTests#test_a_basic_bottle_install [/usr/local/Library/Homebrew/test/test_formula_installer_bottle.rb:45]:
Expected #<Pathname:/var/folders/j0/gh46bxwx1d7303pzss1_khfw0000gp/T/homebrew_tests20150820-13418-19lr5dn/cellar/testball_bottle/0.1/bin>
 to be directory?.
```

Here's the Homebrew output for the failure:

```
==> Downloading file:///usr/local/Library/Homebrew/test/bottles/testball_bottle-0.1.yosemite.bottle.tar.gz
==> Pouring testball_bottle-0.1.yosemite.bottle.tar.gz
/usr/local/Library/Homebrew/download_strategy.rb:218: warning: conflicting chdir during another chdir block
Error: /var/folders/j0/gh46bxwx1d7303pzss1_khfw0000gp/T/homebrew_tests20150820-13628-9wp6gq/cellar/testball_bottle/0.1 is not a directory
```

woodruffw:
Looking at the structure of the testball_bottle tarball, it looks like the problem is being caused by the fact that the directory is still called "testball" and not "testball_bottle". I'm pushing the fix now.

mistydemeo:
More minor test failure now:

```
InstallBottleTests#test_a_basic_bottle_install [/usr/local/Library/Homebrew/test/test_formula_installer_bottle.rb:46]:
--- expected
+++ actual
@@ -1 +1 @@
-3
+2
```

Looks like it's a different number of binaries than you expected.

woodruffw:
Yep, it's because one of the binaries is a hardlink that's been
smashed by the rename from 'testball' to 'testball_bottle'. I'm
constructing a new bottle by hand to fix it.

mistydemeo:
Perfect, thanks!

woodruffw:
`TestballBottle` is working locally, just waiting for the test results :tada:

mistydemeo:
It looks like there's a fairly minor test failure:

```

Warning: File leak is detected
InstallBottleTests#test_a_basic_bottle_install
--- expected
+++ actual
@@ -6,4 +6,5 @@
  "/logs",
  "/formula_cache",
  "/cellar",
- "/cache"]
+ "/cache",
+ "/cache/testball_bottle-0.1.mountain_lion.bottle.tar.gz"]
```

This may not be your fault, but rather an issue with the tests not accounting for the bottle being cached.

mistydemeo:
Interestingly, I can't reproduce this locally.

woodruffw:
Hmm. I suppose I could drop the file counts/checks until this is traced down (it's less important than just ensuring that installation actually occurs). Thoughts?

mikemcquaid:
> Hmm. I suppose I could drop the file counts/checks until this is traced down (it's less important than just ensuring that installation actually occurs). Thoughts?

Dropping the counts seems fine :+1:

woodruffw:
Awesome! Doing it now.

woodruffw:
Looks like this latest failure occurred while the CI was executing `git`, not during the actual testing.

mikemcquaid:
Restarted the build.

mikemcquaid:
Still failing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/31945/version=yosemite/testReport/junit/brew-test-bot/yosemite/tests/

Do you have a local Mac you can access to run `brew tests` on?

woodruffw:
The Mac I've set up as a server doesn't have Xcide for testing purposes, so I can't run `brew tests` directly. I believe @mistydemeo wasn't able to replicate this particular failure on her local setup, so the test bot may be at fault here.

mikemcquaid:
@woodruffw Even if the bot is at fault this will need to be green before we can merge it.

mikemcquaid:
@woodruffw Worst case just remove some of the failing test(s) so we have something in a mergeable state.

woodruffw:
@mikemcquaid Do you happen to know whether `Formula#clear_cache` does or doesn't fully clear all files from `cache/`? Based on the file leak report, it looks like it doesn't (and that I need a method that does).

mikemcquaid:
@woodruffw I'm not sure. You could just manually delete it in the test.

woodruffw:
@mikemcquaid Yeah, that's what I was thinking. Something like
`Dir["#{HOMEBREW_CACHE}/*"].each { |f| File.delete(f) }` should do the
trick.

mikemcquaid:
@woodruffw One last nit would be to delete it specifically by filename instead just to avoid accidentally nuking stuff that other tools may rely on.

woodruffw:
@mikemcquaid Yeah, I tweaked it in the commit to only delete files matching `testball_bottle*`

mistydemeo:
Given how close we are to the deadline, we may need to simply remove the test at this juncture if it can't be made to work. William, are you free to make that change and squash commits?

mistydemeo:
@BrewTestBot test this please

woodruffw:
I am. However, I also think that I'm close to fixing the file leak. If
you think we're getting too close to the deadline, just tell me so and
I'll drop this work and squash down again.

mistydemeo:
Excellent - it passed! :sparkles:

woodruffw:
Looks like it's working :fireworks:
I'll squash right now.

mistydemeo:
OK, let me know when you're done.

woodruffw:
@mistydemeo I've squashed the unit testing commits down, the rest is already pretty compacted from earlier squashes. Any other changes you'd like me to make before final review/merging?

mistydemeo:
Merged in af2d4fd19707827688dbbaf79a3d969f0e92dede! I squashed in another couple of commits and altered the commit messages when merging.

woodruffw:
Awesome! I'll reference this in future related PRs.

mikemcquaid:
@woodruffw Thanks again for all your work here! I was wondering how your Ruby `install_name_tool` was looking and if you'd be able to either integrate it or give us some pointers in doing so? It'd help fix a lot of bugs. Thanks!

woodruffw:
@mikemcquaid Thank you too!

[ruby-macho](https://github.com/woodruffw/ruby-macho) is still incomplete, but it could be integrated in a partial capacity in its current state (mostly reading Mach-Os and Fat binaries, some install name editing). I'm planning on adding rpath editing in the near future (the logic is similar to install name editing), but my schedule is pretty hectic right now.

If you'd like, I can open a PR with the code (possibly in `vendor`?) and make the partial changes. A few methods would still rely on the `cctools` formula being pulled in, but it would begin to lessen the load :smile:

mikemcquaid:
@woodruffw Nice! I think adding it in `vendor` would be great but we should probably wait until it's a complete `cctools` replacement. Alternatively, we could add it now but make it only enabled with an environment variable so maintainers could test it out before it's ready for mainstream?

mikemcquaid:
(and in that case: error out if it hits functionality it can't do yet)

woodruffw:
I think an environment variable sounds good - possibly something like `$HOMEBREW_RUBY_CCTOOLS` or `$HOMEBREW_RUBY_MACHO`?

I'll create a separate PR and start moving the code in!

mikemcquaid:
Either is fine with me, thanks!

woodruffw:
Awesome :smiley:

PR #41022, opened 2015-06-23
Import fetchmail from homebrew-dupes

jacknagel:
This is not available on OS X since 10.8.

See Homebrew/homebrew-dupes#434.

Not sure if it's worth making it keg-only on 10.6 & 10.7 since our support there is limited, and the corresponding homebrew-dupes formula was not keg-only anyway.


mikemcquaid:
:+1:

PR #41025, opened 2015-06-24
ffmpeg: add zmq option

dericed:


Library/Formula/ffmpeg.rb
```diff
@@ -32,6 +32,7 @@ class Ffmpeg < Formula
   option "with-x265", "Enable x265 encoder"
   option "with-libsoxr", "Enable the soxr resample library"
   option "with-webp", "Enable using libwebp to encode WEBP images"
+  option "with-zmq", "Enable using libzmq"
```
mikemcquaid: Can you elaborate on what this does or remove the option (as the text will be be autogenerated)

Library/Formula/ffmpeg.rb
```diff
@@ -69,6 +70,7 @@ class Ffmpeg < Formula
   depends_on "openssl" => :optional
   depends_on "libssh" => :optional
   depends_on "webp" => :optional
+  depends_on "zmq" => :optional
```
DomT4: Please use the full name instead of the alias.

Library/Formula/ffmpeg.rb
dericed: Added some detail and squashed.

Library/Formula/ffmpeg.rb
dericed: Added some detail and squashed.

Library/Formula/ffmpeg.rb
```diff
@@ -32,6 +32,7 @@ class Ffmpeg < Formula
   option "with-x265", "Enable x265 encoder"
   option "with-libsoxr", "Enable the soxr resample library"
   option "with-webp", "Enable using libwebp to encode WEBP images"
+  option "with-zmq", "Enable using libzmq to receive commands sent through a libzmq client"
```
DomT4: This will need to be changed as well to `zeromq`, otherwise a user is going to have to pass `brew install ffmpeg --with-zeromq --with-zmq` to `brew` to install with the desired option.

dericed: changed to zeromq and squashed, please re-review

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/131e9c981180fc2ff89ce6020907f97c51eaadff. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #41027, opened 2015-06-24
clutter-gtk formula cleanup

tschoonj:
revision bump necessary to link properly against gtk 3.16
dependency and option cleanup
audit --strict compliance


mikemcquaid:
Thanks @tschoonj!

For future reference the preferred commit message format for simple version updates is `clutter-gtk: cleanup formula.`.

PR #41033, opened 2015-06-24
vim: client-server support restored

### LINKED ISSUE #41029: "brew install vim --with-client-server" fails to build vim with +clientserver
```
$ brew info vim | head -n 7
vim: stable 7.4.712, HEAD
Vi "workalike" with many additional features
http://www.vim.org/
Conflicts with: ex-vi
/usr/local/Cellar/vim/7.4.712 (1643 files, 31M) *
  Built from source with: --with-client-server
From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/vim.rb
```
```
$ $(brew --prefix)/bin/vim --version | grep clientserver
-clientserver    -hangul_input    +netbeans_intg   +title
```
[brew gist-logs vim](https://gist.github.com/anonymous/6525ee79c3eb7fde1fd8)

tschoonj:
properly this time!
Closes #41029

### DIFF
--- a/Library/Formula/vim.rb
+++ b/Library/Formula/vim.rb
@@ -5,6 +5,7 @@ class Vim < Formula
   # This package tracks debian-unstable: http://packages.debian.org/unstable/vim
   url "https://mirrors.kernel.org/debian/pool/main/v/vim/vim_7.4.712.orig.tar.gz"
   sha256 "b334ba9f6682c605d29fcf45e7fe246b88061736b86c3e7cdfa309404a66b55c"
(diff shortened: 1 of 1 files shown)


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/74fd09cf453c6382a035405e6883a2396d09e17c. Thanks @tschoonj!

Usually we don't permit revisions for optional options, as everyone gets the bump regardless of whether they use the option or not, but I let it go here as we'd already had a couple complaints about it.

PR #41034, opened 2015-06-24
add `brew command`

xu-cheng:
It can be helpful to check whether an external command is installed. Therefore, I can do something like this in my shell configuration:
```bash
if brew command command-not-found-init > /dev/null; then eval "$(brew command-not-found-init)"; fi
```

cc @Homebrew/owners

### DIFF
--- a/Library/brew.rb
+++ b/Library/brew.rb
@@ -65,22 +65,6 @@ end
 begin
   trap("INT", std_trap) # restore default CTRL-C handler
 
-  aliases = {'ls' => 'list',
-             'homepage' => 'home',
-             '-S' => 'search',
-             'up' => 'update',
-             'ln' => 'link',
-             'instal' => 'install', # gem does the same
-             'rm' => 'uninstall',
-             'remove' => 'uninstall',
-             'configure' => 'diy',
-             'abv' => 'info',
-             'dr' => 'doctor',
-             '--repo' => '--repository',
-             'environment' => '--env',
-             '--config' => 'config',
-             }
-
   empty_argv = ARGV.empty?
   help_regex = /(-h$|--help$|--usage$|-\?$|^help$)/
   help_flag = false
@@ -96,7 +80,7 @@ begin
     end
   end
 
-  cmd = aliases[cmd] if aliases[cmd]
+  cmd = HOMEBREW_INTERNAL_COMMAND_ALIASES.fetch(cmd, cmd)
 
   sudo_check = %w[ install link pin unpin upgrade ]
 
--- a/Library/Homebrew/cmd/command.rb
+++ b/Library/Homebrew/cmd/command.rb
@@ -1,0 +1,14 @@
+module Homebrew
+  def command
+    cmd = ARGV.first
+    cmd = HOMEBREW_INTERNAL_COMMAND_ALIASES.fetch(cmd, cmd)
+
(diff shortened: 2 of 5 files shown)


Library/Homebrew/cmd/which.rb
```diff
@@ -0,0 +1,17 @@
+module Homebrew
+  def which
+    cmd = ARGV.first
+
+    if HOMEBREW_INTERNAL_COMMAND_ALIASES.has_key? cmd
+      cmd = HOMEBREW_INTERNAL_COMMAND_ALIASES[cmd]
+    end
+
+    if (path = HOMEBREW_LIBRARY_PATH/"cmd/#{cmd}.rb"; path.file?)
```
bfontaine: Does it really make sense for internal commands?

xu-cheng: Which part doesn't make sense? It works fine for all internal commands.

mikemcquaid:
:+1: on the idea. I wonder if it makes more sense to have a very simple version inside `brew.rb`? dunno.

xu-cheng:
>  I wonder if it makes more sense to have a very simple version inside brew.rb?

I think this is easier to maintain rather than mixing different functions together.

Library/Homebrew/cmd/which.rb
bfontaine: Yes, I just wonder if people will use the tool for internal commands. But it makes sense :smile: (I need more :coffee: )

xu-cheng: Although, I do wonder should we implement `bfontaine/homebrew-aliases` inside the core so `brew which` can print the alias content rather than file.

jacknagel:
`which` is probably a poor choice of name, as it used to be a command that did something completely different.

maybe tack it onto the `brew commands` command?

DomT4:
Agree with Jack that the name is quite confusing.

I notice you're supposed to hit `odie "Unknown command: #{cmd}"` if you pass something that isn't valid, but I'm hitting:

```
Error: wrong number of arguments (1 for 0)
Please report this bug:
/usr/local/Library/Homebrew/cmd/which.rb:2:in `which'
/usr/local/Library/Homebrew/cmd/which.rb:11:in `which'
/usr/local/Library/brew.rb:126:in `<main>'
```

Rather than ever reaching the error message if I pass anything that isn't a valid `cmd`.

xu-cheng:
I change the name to `brew command`.

Library/brew.rb
```diff
@@ -96,7 +80,9 @@ def require? path
     end
   end
 
-  cmd = aliases[cmd] if aliases[cmd]
+  if HOMEBREW_INTERNAL_COMMAND_ALIASES.has_key? cmd
+    cmd = HOMEBREW_INTERNAL_COMMAND_ALIASES[cmd]
+  end
```
jacknagel: `cmd = HOMEBREW_INTERNAL_COMMAND_ALIASES.fetch(cmd, cmd)`

xu-cheng:
Updated. Any more comment?

xu-cheng:
The name is still weird. Should we use `brew where` instead?

mikemcquaid:
I dunno, `brew command` makes sense to me.

DomT4:
Yeah, `brew command` or even `brew cmd` works for me.

xu-cheng:
Thanks. I will merge this tomorrow. Being busy recently on my first PhD paper. :smile:

PR #41035, opened 2015-06-24
erlang: update to otp-18.0

optikfluffel:
- [x] couchdb #41128
- [x] lfe #41121
- [x] elixir #41191
- [x] rebar #41119
- [x] geocouch - [_comment_](https://github.com/Homebrew/homebrew/pull/41035#issuecomment-122279385)
- [x] kjell #41645
- [x] mercury - [_comment_](https://github.com/Homebrew/homebrew/pull/41035#issuecomment-122512582)
- [x] syntaxerl [ten0s/syntaxerl#8](https://github.com/ten0s/syntaxerl/issues/8#issuecomment-120861613)
- [x] tsung - #41976
- [ ] wrangler [RefactoringTools/wrangler#67](https://github.com/RefactoringTools/wrangler/issues/67)
- [x] yaws #41668

PR #41036, opened 2015-06-24
erlang 18.0

alindeman:


DomT4:
Afraid you were narrowly beaten to the PR by #41035. Thanks for the submission though, we :heart: the enthusiasm!

PR #41038, opened 2015-06-24
libetonyek 0.1.3

luzpaz:


DomT4:
Install is failing:

```
checking for MDDS... no
configure: error: Package requirements (mdds) were not met:

No package 'mdds' found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.
```

luzpaz:
Closing until mdds is submitted to homebrew

DomT4:
If nothing else is likely to use `mdds` soon, you could always vendor it into the formula here as a `resource`.

PR #41041, opened 2015-06-24
libgit2-glib 0.22.8

tschoonj:
version bump

### DIFF
--- a/Library/Formula/libgit2-glib.rb
+++ b/Library/Formula/libgit2-glib.rb
@@ -1,8 +1,8 @@
 class Libgit2Glib < Formula
   desc "Glib wrapper library around libgit2 git access library"
   homepage "https://github.com/GNOME/libgit2-glib"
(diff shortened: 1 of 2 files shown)


DomT4:
Can you bump `libgit2` in the same PR here please? The both need doing and one has a dep on the other so there's less room for things to go wrong if we do them together :smile_cat:

tschoonj:
done! but there seems to be something very wrong with the buildbot...

DomT4:
We're looking at this now. Will re-queue any builds necessary once we've ironed it out. Thanks!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/d9fa71e53379ec379175540af2bddde3d27b6086 & https://github.com/Homebrew/homebrew/commit/70d232588fe131db0393342dc28be6f57617485d. Thanks @tschoonj!

PR #41045, opened 2015-06-24
Cyprit-opencl for brew

gerardbriscoe:


Library/Formula/cpyrit-opencl.rb
```diff
@@ -0,0 +1,15 @@
+class CpyritOpencl < Formula
+  desc "OpenCL support for Pyrit."
+  homepage "https://code.google.com/p/pyrit/wiki/Installation"
+  url "https://pyrit.googlecode.com/files/cpyrit-opencl-0.4.0.tar.gz"
+  sha256 "aac593bce3f00ea7fd3d558083dbd7e168332a92736e51a621a0459d1bc042fa"
+
+  depends_on :python
```
DomT4: Should be `depends_on :python if MacOS.version <= :snow_leopard`

Library/Formula/cpyrit-opencl.rb
```diff
@@ -0,0 +1,15 @@
...
+  depends_on 'pyrit'
```
DomT4: Double quotes everywhere.

Library/Formula/cpyrit-opencl.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  def install
+    system "python", "setup.py", "build"
+    system "python", "setup.py", "install", "--prefix=#{prefix}"
```
DomT4: Should be set up something like [this](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/jrnl.rb#L62-L66) if possible.

Library/Formula/cpyrit-opencl.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
```
DomT4: No need for an empty line here.

Library/Formula/cpyrit-opencl.rb
tdsmith: No, this is a library, not an application.

tdsmith:
This should be submitted to Homebrew/homebrew-python.

Library/Formula/cpyrit-opencl.rb
tdsmith: No, you're right, I'm wrong.

tdsmith:
All of the pyrit formulas could be migrated into Homebrew/homebrew from Homebrew/homebrew-python if you prefer but they should all be in the same place. Does it make sense to ship this separately from the pyrit formula or should we always bundle it?

Library/Formula/cpyrit-opencl.rb
DomT4: Ah, Sorry. I didn't know we handled python library formula differently in terms of the dep:question:

gerardbriscoe:
I am new to brew. I followed the example provided and cpyrit-cuda as much as I could.

DomT4:
> This should be submitted to Homebrew/homebrew-python.

Let's do this, Thanks!

PR #41051, opened 2015-06-24
zeromq multicast support, pgm and norm

duedal:
Currently the zeromq formular in homebrew does not have working multicast support. There was a PGM option, but it broke (```configure: WARNING: unrecognized options: --with-system-pgm```) at some point, most likely with zeromq 4.1.
This pull request fixes that, but also adds optional support for the [new norm transport](http://zeromq.org/topics:norm-protocol-transport), included with zeromq 4.1. [Norm](http://www.nrl.navy.mil/itd/ncs/products/norm) seems like a little known, but widely supported and without deps, solution to reliable multicast communication. 
Zeromq already gave the perfect [rationale](http://zeromq.org/topics:norm-protocol-transport), but from my POV it's also easier to use than PGM, especially when running multiple zeromq instances, such as on a development laptop. In fact, based on my testing, it might be the only way to do it on OS X without doing virtual network interfaces.

### DIFF
--- a/Library/Formula/norm.rb
+++ b/Library/Formula/norm.rb
@@ -1,0 +1,30 @@
+class Norm < Formula
+  desc "NACK-Oriented Reliable Multicast"
+  homepage "https://www.nrl.navy.mil/itd/ncs/products/norm"
+  url "https://downloads.pf.itd.nrl.navy.mil/norm/archive/src-norm-1.5b3.tgz"
+  sha256 "ec2014f57be12ea3ade0685faa1e6e161f168311c8975093bce33a6e1dc5e77c"
+
+  def install
+    system "./waf", "configure", "--prefix=#{prefix}"
+    system "./waf", "install"
+    include.install "include/normApi.h"
+  end
+
+  test do
+    (testpath/"test.c").write <<-EOS.undent
+      #include <assert.h>
+      #include <normApi.h>
+
+      int main()
+      {
+        NormInstanceHandle i;
+        i = NormCreateInstance(false);
+        assert(i != NORM_INSTANCE_INVALID);
+        NormDestroyInstance(i);
+        return 0;
+      }
+    EOS
+    system ENV.cc, "test.c", "-L#{lib}", "-lnorm", "-o", "test"
+    system "./test"
+  end
+end
--- a/Library/Formula/zeromq.rb
+++ b/Library/Formula/zeromq.rb
@@ -24,12 +24,14 @@ class Zeromq < Formula
 
   option :universal
   option "with-libpgm", "Build with PGM extension"
+  option "with-norm", "Build with NORM extension"
 
   deprecated_option "with-pgm" => "with-libpgm"
 
   depends_on "pkg-config" => :build
   depends_on "libpgm" => :optional
   depends_on "libsodium" => :optional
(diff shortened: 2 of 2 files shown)


Library/Formula/norm.rb
```diff
@@ -0,0 +1,30 @@
+class Norm < Formula
+  desc "NACK-Oriented Reliable Multicast"
+  homepage "http://www.nrl.navy.mil/itd/ncs/products/norm"
+  url "http://downloads.pf.itd.nrl.navy.mil/norm/archive/src-norm-1.5b3.tgz"
```
vszakats: Please use the https version of the `homepage`:
`homepage "https://www.nrl.navy.mil/itd/ncs/products/norm"`
(it's mixed content; something to fix upstream)

And the same for the `url`:
`url "https://downloads.pf.itd.nrl.navy.mil/norm/archive/src-norm-1.5b3.tgz"`

Library/Formula/zeromq.rb
```diff
@@ -48,6 +50,10 @@ def install
       args << "--without-libsodium"
     end
 
+    if build.with? "norm"
```
DomT4: Since there's no `without` option being passed to `configure` here this can be single-line:

```ruby
args << "--with-norm" if build.with? "norm"
```

duedal:
Thank you @DomT4 and @vszakats for your improvements. I've applied them and squashed the commits.

vszakats:
Thank you!

Library/Formula/norm.rb
```diff
@@ -0,0 +1,30 @@
+class Norm < Formula
+  desc "NACK-Oriented Reliable Multicast"
+  homepage "https://www.nrl.navy.mil/itd/ncs/products/norm"
+  url "https://downloads.pf.itd.nrl.navy.mil/norm/archive/src-norm-1.5b3.tgz"
```
mikemcquaid: Is there no stable version available?

duedal: All the releases are marked with 'b', so no AFAIK. zeromq requires a version >=1.5b3

I'm in touch with the maintainer, but not sure I can convince him to make a non 'b' release, is that a deal breaker for inclusion in homebrew?

mikemcquaid: If you can just check if they are considered beta or stable that'd be enough, thanks!

duedal: Sure. I've asked the maintainer @bebopagogo to give an assessment.

bebopagogo: The 'b' isn't really beta.  I should probably change my versioning number schema, perhaps changing the 'b' for "build" to an 'r' for "release" instead to reduce this recurring question.  The posted tarballs represent stable release.  I will be posting a new stable release tarball soon that includes the  One thing to note is the current stable release that is posted is version 1.5b4 (the 1.5b3 version you are using is from the "archive" of prior releases), but there have been a couple of bug fixes in the code that may benefit ZeroMQ use.  I may even increment to a 1.6 version number if I drop the 'b' delimiter in favor of something else.

mikemcquaid: Great, thanks! Just wanted to check.

mikemcquaid:

PR #41060, opened 2015-06-25
upgrade vert.x to 3.0.0

aruis:
upgrade vert.x to 3.0.0
  url 'http://dl.bintray.com/vertx/downloads/vert.x-3.0.0-full.tar.gz'
  sha1 '47b0fad8de3e9c175fca9ff70a127facdd0c0cf0'

### DIFF
--- a/Library/Formula/vert.x.rb
+++ b/Library/Formula/vert.x.rb
@@ -9,7 +9,6 @@ class VertX < Formula
   def install
     rm_f Dir["bin/*.bat"]
     libexec.install %w[bin client conf lib]
-    doc.install %w[api-docs]
     bin.install_symlink "#{libexec}/bin/vertx"
   end
 end
--- a/Library/Formula/vert.x.rb
+++ b/Library/Formula/vert.x.rb
@@ -3,7 +3,7 @@ require 'formula'
 class VertX < Formula
   desc "Application platform for the JVM"
   homepage 'http://vertx.io/'
-  url 'https://bintray.com/artifact/download/vertx/downloads/vert.x-3.0.0-full.zip'
+  url 'https://dl.bintray.com/vertx/downloads/vert.x-3.0.0-full.tar.gz'
   sha1 '47b0fad8de3e9c175fca9ff70a127facdd0c0cf0'
 
   def install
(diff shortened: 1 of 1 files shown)


aruis:
I don't know how test it.
And I use 'brew install '+my_github_url/../vert.x.rb .
But it failed.

```==> Downloading http://dl.bintray.com/vertx/downloads/vert.x-3.0.0-full.tar.gz
Already downloaded: /Library/Caches/Homebrew/vert.x-3.0.0.tar.gz
Error: No such file or directory - api-docs```

File: Library/Formula/vert.x.rb [hunk 1]
```diff
@@ -3,8 +3,8 @@
 class VertX < Formula
   desc "Application platform for the JVM"
   homepage 'http://vertx.io/'
-  url 'http://dl.bintray.com/vertx/downloads/vert.x-2.1.5.tar.gz'
-  sha1 'bac58701858462f13c0b01de99e2ada0a4df8431'
+  url 'http://dl.bintray.com/vertx/downloads/vert.x-3.0.0-full.tar.gz'
```
DomT4: This should be `https://`

Library/Formula/vert.x.rb
```diff
@@ -3,8 +3,8 @@
...
+  sha1 '47b0fad8de3e9c175fca9ff70a127facdd0c0cf0'
```
DomT4: And ideally this should be `sha256`.

DomT4:
> I don't know how test it.

The test-bot will test it here.

> Error: No such file or directory - api-docs

This probably means they've renamed or moved the directory. You'll need to download the `vert` tarball and have a look inside.

aruis:
Thank you @DomT4 , I will correct it.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/724cf3e14e3308a72fcd5e8a87a56c8af54619b4. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a couple nits for future: Ideally the checksum should be `sha256` rather than `sha1`, and the preferred commit message for version bumps is a simple `vert.x 3.0.0` or similar, Thanks!

PR #41062, opened 2015-06-25
modgit 1.0.1 (new formula)

rafaelgarrido:
Adding new formula: modgit is a shell script for deploying multiple Git repositories in root folder of any project, which is not possible with default git submodule command. A common use case would be the easy installation of Magento modules that need to be deployed in root folder.


mikemcquaid:

PR #41064, opened 2015-06-25
fetchmail: explicitly specify openssl prefix

lembacon:


DomT4:
Is it not currently picking up on Homebrew's OpenSSL correctly in some way?

lembacon:
@DomT4 On OS X El Capitan (as of beta 2), it will attempt to pick up the nonexistent system OpenSSL.

mikemcquaid:
Can you gist the error, out of interest? Thanks!

lembacon:
@mikemcquaid https://gist.github.com/1c4301ebc11897a7c182

jacknagel:
Looking at the configure script, it just loops over a list of some common places that openssl can be found (/usr/local, /usr, etc.) but since the headers are missing from 10.11 it fails.

PR #41065, opened 2015-06-25
pdf2svg: migrate back to core (from homebrew-x11)

UniqMartin:
Turns out `pdf2svg` no longer depends on GTK+ or X11 (and previously did this only by mistake).

See also the related PR Homebrew/homebrew-x11#97.


UniqMartin:
The bot isn't happy about the conflicting formula name (because it still sees `homebrew/x11/pdf2svg`), but everything else passes. Just wanted to point that out …

xu-cheng:

Issue #39870, opened 2015-05-18
openssl and libressl if certs.pem contains expired root certificate


smuething:
The openssl and libressl formulae bootstrap the cert.pem file with Apple's root certificate store. For some reason, that store contains a number of expired root certificates with the same CN as a newer, valid root certificate, e.g. for CN='GlobalSign Root CA'.

Unfortunately, both openssl and libressl insist on checking the expiry dates of both the new (valid) and old (expired) root certificate during the SSL handshake, causing connection failures. Example:
```bash
> /usr/local/opt/openssl/bin/openssl s_client -showcerts -CAfile /usr/local/etc/openssl/cert.pem -connect git.dune-project.org:443
CONNECTED(00000003)
depth=2 C = BE, O = GlobalSign nv-sa, OU = Root CA, CN = GlobalSign Root CA
verify error:num=10:certificate has expired
notAfter=Jan 28 12:00:00 2014 GMT
---
Certificate chain
 0 s:/OU=Domain Control Validated/CN=*.dune-project.org
   i:/C=BE/O=GlobalSign nv-sa/CN=GlobalSign Domain Validation CA - SHA256 - G2
-----BEGIN CERTIFICATE-----
MIIFIzCCBAugAwIBAgISESFTYVoL1B8yW7N4NEZDMmSuMA0GCSqGSIb3DQEBCwUA
MGAxCzAJBgNVBAYTAkJFMRkwFwYDVQQKExBHbG9iYWxTaWduIG52LXNhMTYwNAYD
VQQDEy1HbG9iYWxTaWduIERvbWFpbiBWYWxpZGF0aW9uIENBIC0gU0hBMjU2IC0g
RzIwHhcNMTQwNzE4MTMzNDQwWhcNMTUwOTE0MjAyODM4WjBAMSEwHwYDVQQLExhE
b21haW4gQ29udHJvbCBWYWxpZGF0ZWQxGzAZBgNVBAMMEiouZHVuZS1wcm9qZWN0
Lm9yZzCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAMaTMLlz8M7eK3sK
T7wr5s4DdthQBbi3mgtkczyc9oBHML9cCGVzG/zUcVE7aNJsd8VCjUOXNz6ihoMP
qGOv9jor6JJZaSatiYNg5p/lRSE0CnprsqweGLUPK1AIHxnR6kA63JLdd3tZY1r8
[...]

PR #41071, opened 2015-06-25
Ignore expired certificates for OpenSSL

kouno:
I implemented the fix from @smuething in https://github.com/Homebrew/homebrew/issues/39870#issue-77627443.

Let me know if we can get this merged or if I messed up something on the way :smile:.

### DIFF
--- a/Library/Formula/openssl.rb
+++ b/Library/Formula/openssl.rb
@@ -107,8 +107,22 @@ class Openssl < Formula
       /System/Library/Keychains/SystemRootCertificates.keychain
     ]
 
+    certs_list = `security find-certificate -a -p #{keychains.join(" ")}`
+    certs = certs_list.scan(
+      /-----BEGIN CERTIFICATE-----.*?-----END CERTIFICATE-----/m
+    )
+
+    valid_certs = certs.select do |cert|
+      IO.popen('openssl x509 -inform pem -checkend 0 -noout', 'w') do |openssl_io|
+        openssl_io.write(cert)
+        openssl_io.close_write
+      end
+
+      $?.success?
+    end
+
     openssldir.mkpath
-    (openssldir/"cert.pem").atomic_write `security find-certificate -a -p #{keychains.join(" ")}`
+    (openssldir/"cert.pem").atomic_write(valid_certs.join("\n"))
   end
 
   def caveats; <<-EOS.undent


Library/Formula/openssl.rb
```diff
@@ -121,6 +118,33 @@ def caveats; <<-EOS.undent
...
+
+  def get_system_certificates
+    keychains = %w[
+      /Library/Keychains/System.keychain
+      /System/Library/Keychains/SystemRootCertificates.keychain
+    ]
+
+    certs_list = `security find-certificate -a -p #{keychains.join(" ")}`
+    certs = certs_list.scan(
+      /#{Regexp.escape('-----BEGIN CERTIFICATE-----')}.*?#{Regexp.escape('-----END CERTIFICATE-----')}/m
```
jacknagel: There aren't any metacharacters here so we shouldn't need to escape the strings.

Library/Formula/openssl.rb
```diff
@@ -121,6 +118,33 @@ def caveats; <<-EOS.undent
...
+    )
+
+    prune_expired_certificates(certs)
+  end
+
+  def prune_expired_certificates(certs)
+    certs.keep_if do |cert|
```
jacknagel: `certs.select`

Library/Formula/openssl.rb
```diff
@@ -121,6 +118,33 @@ def caveats; <<-EOS.undent
...
+      IO.popen(%w[openssl x509 -inform pem -checkend 0 -noout], 'w') do |openssl_io|
```
jacknagel: Array form of IO.popen isn't available on Ruby 1.8

Library/Formula/openssl.rb
kouno: I have no idea why I thought that `-` needed to be escaped. :)

DomT4:
Why are we making the method `private` here?

Library/Formula/openssl.rb
```diff
@@ -121,6 +118,33 @@ def caveats; <<-EOS.undent
...
+  def prune_expired_certificates(certs)
+    certs.select do |cert|
+      IO.popen('openssl x509 -inform pem -checkend 0 -noout', 'w') do |openssl_io|
+        openssl_io.write(cert)
+        openssl_io.close_write
+      end
+
+      $?.success?
+    end
+  end
```
mikemcquaid: Just inline these methods as they are only called once, thanks.

kouno:
Updated after feedback.

mikemcquaid:
@kouno you missed "Just inline these methods as they are only called once, thanks."

kouno:
> @kouno you missed "Just inline these methods as they are only called once, thanks."

If you look at the diff, I removed `prune_expired_certificates`. Your comment is still showing because you commented on a line that I didn't modify.

kouno:
Ah... sorry. I just got it. You meant both of them.

DomT4:
Happy to pull this. It doesn't make a massive change to the cert list, a few hundred lines difference, but the duplication of expired/valid certificates in the Keychain is enough of a problem to merit action on our side and it's likely to help the situation.

Given Apple's long-term shift away from OpenSSL on OS X I'd be tempted to say the post-install usage of OpenSSL here should lean on the Homebrew OpenSSL, but I'll leave it for now and see how it goes.

We need to apply this to GnuTLS and LibreSSL as well. If you want to handle those with a new PR that'd be awesome. Thanks for your contribution to Homebrew, we appreciate it! :smiley_cat:.

PR #41075, opened 2015-06-25
Disable MMX. Clang in 10.11 doesn't support the inline assembly. 

shaneburrell:
Upstream has been notified.

### DIFF
--- a/Library/Formula/pixman.rb
+++ b/Library/Formula/pixman.rb
@@ -31,18 +31,15 @@ class Pixman < Formula
 
   def install
     ENV.universal_binary if build.universal?
-    if MacOS.version >= "10.11" then
-      system "./configure", "--disable-dependency-tracking",
-                            "--disable-silent-rules",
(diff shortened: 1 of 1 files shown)


mikemcquaid:
@shaneburrell Let's disable this just for 10.11 for now, thanks.

shaneburrell:
That should get it.

DomT4:
Let's not dupe the whole arguments just for 10.11. You can do something like this:

```ruby
args = %W[
  --disable-dependency-tracking
  --prefix=#{prefix}
  --another-arg
  --and-another
  --so-on
]

args <<  "--disable-mmx" if MacOS.version >= :el_capitan
```

shaneburrell:
Changed to args and optionally disable mmx if >= el_capitan.

mikemcquaid:

For future reference the preferred commit message format is `pixman: disable MMX on 10.11.`.

PR #41078, opened 2015-06-25
cutter: fix desc and modernize

### LINKED ISSUE #41073: cutter formula has incorrect description
The description reads:
desc "Disconnect routed IP connections"

however this formula is actually the unittest testing framework rather then the connection cutter.

The official description is: "a Unit Testing Framework for C and C++"

xu-cheng:
Fixes #41073.


mikemcquaid:
:metal:

PR #41079, opened 2015-06-25
ice 3.6.0

externl:
Updating to Ice 3.6.0.

### DIFF
--- a/Library/Formula/ice.rb
+++ b/Library/Formula/ice.rb
@@ -1,11 +1,8 @@
-require 'formula'
-
 class Ice < Formula
-  desc "Distributed computing platform"
-  homepage 'http://www.zeroc.com'
-  url 'http://www.zeroc.com/download/Ice/3.5/Ice-3.5.1.tar.gz'
-  sha1 '63599ea22a1e9638a49356682c9e516b7c2c454f'
-  revision 1
+  desc "A comprehensive RPC framework"
+  homepage "https://zeroc.com"
+  url "https://github.com/zeroc-ice/ice/archive/v3.6.0.tar.gz"
+  sha256 "77933580cdc7fade0ebfce517935819e9eef5fc6b9e3f4143b07404daf54e25e"
 
   bottle do
     sha1 "52d43a08202206258c6fa33750aac25d2dc38703" => :yosemite
@@ -13,26 +10,17 @@ class Ice < Formula
     sha1 "7bdb347052b8df900c2c2e6797726a79d9c0242d" => :mountain_lion
   end
 
-  option 'doc', 'Install documentation'
-  option 'demo', 'Build demos'
+  depends_on "mcpp"
+  depends_on :java  => ["1.7", :optional]
+  depends_on :macos => :mavericks
 
   resource "berkeley-db" do
-    url "http://download.oracle.com/berkeley-db/db-5.3.28.tar.gz"
-    sha1 "fa3f8a41ad5101f43d08bc0efb6241c9b6fc1ae9"
+    url "http://download.oracle.com/berkeley-db/db-5.3.28.NC.tar.gz"
+    sha256 "76a25560d9e52a198d37a31440fd07632b5f1f8f9f2b6d5438f4bc3e7c9013ef"
   end
-
-  depends_on "openssl"
-  depends_on "mcpp"
-  depends_on :python => :optional
-
-  # 1. TODO: document the first patch
-  # 2. Patch to fix build with libc++, reported upstream:
-  # http://www.zeroc.com/forums/bug-reports/6152-mavericks-build-failure-because-unexported-symbols.html
-  patch :DATA
 
   def install
     resource("berkeley-db").stage do
-
       # Fix build under Xcode 4.6
       # Double-underscore names are reserved, and __atomic_compare_exchange is now
       # a built-in, so rename this to something non-conflicting.
@@ -42,16 +30,14 @@ class Ice < Formula
 
       # BerkeleyDB dislikes parallel builds
       ENV.deparallelize
-
-      # --enable-compat185 is necessary because our build shadows
-      # the system berkeley db 1.x
       args = %W[
         --disable-debug
         --prefix=#{libexec}
         --mandir=#{libexec}/man
         --enable-cxx
-        --enable-compat185
       ]
+
+      args << "--enable-java" if build.with? "java"
 
       # BerkeleyDB requires you to build everything from the build_unix subdirectory
       cd "build_unix" do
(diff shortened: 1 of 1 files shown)


Library/Formula/ice.rb
```diff
@@ -13,26 +10,21 @@ class Ice < Formula
     sha1 "7bdb347052b8df900c2c2e6797726a79d9c0242d" => :mountain_lion
   end
 
-  option 'doc', 'Install documentation'
-  option 'demo', 'Build demos'
+  option "without-java", "Compile without Java support."
+
+  if build.with? "java"
+    depends_on :java => "1.7"
```
DomT4: This can be on one line. `depends_on :java => "1.7" if build.with? "java"`

Library/Formula/ice.rb
```diff
@@ -13,26 +10,21 @@ class Ice < Formula
...
+  end
 
   resource "berkeley-db" do
-    url "http://download.oracle.com/berkeley-db/db-5.3.28.tar.gz"
-    sha1 "fa3f8a41ad5101f43d08bc0efb6241c9b6fc1ae9"
+    url "http://download.oracle.com/berkeley-db/db-5.3.28.NC.tar.gz"
+    sha1 "8e8971fb49fff9366cf34db2f04ffbb7ec295cc2"
```
DomT4: Let's use `sha256` here, Thanks!

Library/Formula/ice.rb
```diff
@@ -1,11 +1,8 @@
-require 'formula'
-
 class Ice < Formula
-  desc "Distributed computing platform"
-  homepage 'http://www.zeroc.com'
-  url 'http://www.zeroc.com/download/Ice/3.5/Ice-3.5.1.tar.gz'
-  sha1 '63599ea22a1e9638a49356682c9e516b7c2c454f'
-  revision 1
+  desc "A comprehensive RPC framework with support for C++, .NET, Java, Python, JavaScript and more"
```
DomT4: The limit is 80 characters. Is this over that?

Library/Formula/ice.rb
```diff
@@ -13,26 +10,21 @@ class Ice < Formula
     sha1 "7bdb347052b8df900c2c2e6797726a79d9c0242d" => :mountain_lion
   end
 
-  option 'doc', 'Install documentation'
```
DomT4: Are docs/demos not available any more?

Library/Formula/ice.rb
externl: Will fix :)

Library/Formula/ice.rb
externl: Good idea, will fix.

Library/Formula/ice.rb
externl: I'll check to be sure

Library/Formula/ice.rb
externl: It's all on GitHub now, not installed

mikemcquaid:
:+1:

DomT4:
Why is OpenSSL no longer required here by the way? There still seems to be plenty of references to OpenSSL in the upstream source and documents. Are they using an internal SSL library instead now?

externl:
OpenSSL is no longer required because we switched to Secure Transport. We're currently in the process of releasing, docs are being updated :).

externl:
In case you're wondering I work for ZeroC

DomT4:
Thanks! I just found the commit removing the OpenSSL dep. All is good. Appreciate the swift reply! :+1:

Library/Formula/ice.rb
```diff
@@ -42,170 +32,101 @@ def install
...
-      You can ensure this is possible by appending the path to
-      Ice's PHP includes to your global include_path in php.ini:
+      If you installed with Java support the IceGrid Admin application was installed.
 
-          include_path=<your-original-include-path>:#{prefix}/php
-
-      However, you can also accomplish this on a script-by-script basis
-      or via .htaccess if you so desire...
-      EOS
+      Run `brew linkapps ice` to symlink it into /Applications.
```
DomT4: You shouldn't need this line. If `brew` detects an `.app` bundle in either `prefix` or `libexec` it'll throw this dialogue automatically.

externl: The `.app` bundle is installed into bin!

mikemcquaid: Please move it into the `prefix` root instead, thanks.

DomT4:
Can you squash your commits here? I'm having some trouble pulling this locally. Thanks!

Library/Formula/ice.rb
```diff
@@ -42,170 +32,97 @@ def install
...
+        #ENV.append_to_cflags("-I/System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers/")
+
+        args << "--enable-java"
+      end
+
       # BerkeleyDB requires you to build everything from the build_unix subdirectory
       cd "build_unix" do
         system "../dist/configure", *args
-        system "make", "install"
+        system "make install"
```
DomT4: This should remain `system "make", "install"`

externl: Fixed and squash merged with latest from master

Library/Formula/ice.rb
```diff
@@ -13,26 +10,19 @@ class Ice < Formula
     sha1 "7bdb347052b8df900c2c2e6797726a79d9c0242d" => :mountain_lion
   end
 
-  option 'doc', 'Install documentation'
-  option 'demo', 'Build demos'
+  option "without-java", "Compile without Java support."
+
+  depends_on :java => "1.7" if build.with? "java"
```
rleigh-dundee: You need a

```
depends_on MinimumMacOSRequirement => :mavericks
```

to prevent building on 10.8.

rleigh-dundee:
Reading through the changes, it mostly all looks good.  The major thing I noticed was the removal of support for python.  While it can be installed with pip now, is there any reason not to build it direct from the source tree we have to hand against the homebrew (or system) python?  [We are major users of ice built against homebrew python.]

externl:
Hey,

With the release of Ice 3.6.0 we've moved Python and Ruby from the main binary distribution. One reason you can't build Python and ruby from source like the last release is that there is no longer a slice2py or slice2rb binaries. These are now genetated as part of the pypi and gem bundles. 

As long as you use pip from homebrew Ice for Python will compile against your homebrew Python.

rleigh-dundee:
@externl OK, thanks.  Just done that and I'm doing our first test build against 3.6.0 (previously tested with the beta).

By the way, thanks for being so proactive in updating the homebrew formula.  I was in the process of doing this and was just about to open a PR when I noticed you'd already done it!

Library/Formula/ice.rb
DomT4: The syntax here should be `depends_on :macos => :mavericks`, or `=> :yosemite` or such.

rleigh-dundee: @DomT4 Isn't that a dependency on a specific version, rather than all versions greater than some version?

I've used the above here:
https://github.com/Homebrew/homebrew-science/pull/2326/files#diff-0cdaeb4fa14b9af05dd33275e4359ff1R24
and it worked fine there.  Not sure what the difference is here.  It's in
https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/requirements/minimum_macos_requirement.rb so not sure why it's not working here.

DomT4: Nope. `=>` in this context means "This version or any greater version".

If you look at our dependency handling you can see `MinimumMacOSRequirement` resolves to `:macos` [here](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/dependency_collector.rb#L99). There is no "pure" usage of `MinimumMacOSRequirement` in any core formula.

rleigh-dundee: Ah, OK.  Thanks for the explanation.

externl: I'll fix this in a few hours.

Library/Formula/ice.rb
```diff
@@ -42,17 +30,35 @@ def install
...
+      if build.with? "java"
+        java_home = ENV["JAVA_HOME"] = `/usr/libexec/java_home -v 1.7`.chomp
+
+        # The Oracle JDK puts jni.h into #{java_home}/include and jni_md.h into
+        # #{java_home}/include/darwin.  The original Apple/SUN JDK placed jni.h
+        # and jni_md.h into
+        # /System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers/
+        #
+        # Setup the include path with the Oracle JDK location first and the Apple JDK location second.
+        ENV.append("CPPFLAGS", "-I#{java_home}/include")
```
DomT4: Are these necessary? This should be being automatically passed by the `:java` dependency and safe to not need?

Library/Formula/ice.rb
```diff
@@ -42,17 +30,35 @@ def install
...
+        ENV.append("CPPFLAGS", "-I#{java_home}/include/darwin")
+        ENV.append("CPPFLAGS", "-I/System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers/")
+
+        # This doesn't work at present.
```
DomT4: Let's just cut out lines that don't work.

Library/Formula/ice.rb
```diff
@@ -60,152 +66,61 @@ def install
...
     ]
 
-    ENV["DB_HOME"] = "#{libexec}"
-
-    args << "CXXFLAGS=#{ENV.cflags} -Wall -D_REENTRANT"
-
-    # Unset ICE_HOME as it interferes with the build
-    ENV.delete('ICE_HOME')
+    #
+    # Setting this gets rid of the optimization level and the arch flags.
```
DomT4: Likewise here, if we're not using it, kill it.

Library/Formula/ice.rb
```diff
@@ -42,17 +30,35 @@ def install
...
       args = %W[
         --disable-debug
         --prefix=#{libexec}
         --mandir=#{libexec}/man
         --enable-cxx
-        --enable-compat185
       ]
 
+      if build.with? "java"
+        java_home = ENV["JAVA_HOME"] = `/usr/libexec/java_home -v 1.7`.chomp
```
DomT4: I think this should be handled by the `:java` dep as well. CC @xu-cheng?

Library/Formula/ice.rb
```diff
@@ -1,38 +1,26 @@
...
-  resource "berkeley-db" do
-    url "http://download.oracle.com/berkeley-db/db-5.3.28.tar.gz"
-    sha1 "fa3f8a41ad5101f43d08bc0efb6241c9b6fc1ae9"
-  end
-
-  depends_on "openssl"
+  depends_on :macos => :yosemite
   depends_on "mcpp"
-  depends_on :python => :optional
+  depends_on :java => "1.7" if build.with? "java"
```
xu-cheng: Cannot work with Java 1.8?

Library/Formula/ice.rb
xu-cheng: Yes all of this settings including `ENV.append("CPPFLAGS", ...)` are already handed by `:java`. `args << "--enable-java"` should be enough.

Library/Formula/ice.rb
xu-cheng: Also it should be `depends_on :java => ["1.7", :optional]`

externl: Java 7 is required to build Berkeley DB. This was brought up a few months ago, and as of then there is no way to patch a vendored part of a formula.

Library/Formula/ice.rb
externl: If you read the comment, the do not do it correctly!
```
# The Oracle JDK puts jni.h into #{java_home}/include and jni_md.h into
        # #{java_home}/include/darwin.  The original Apple/SUN JDK placed jni.h
        # and jni_md.h into
        # /System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers/
        #
        # Setup the include path with the Oracle JDK location first and the Apple JDK location second.
```

Library/Formula/ice.rb
externl: I'll remove the java_home part. the CPPFLAGS are still necessary.

Library/Formula/ice.rb
externl: Java should be the default

DomT4: > Java should be the default as it builds the GUI application.

`depends_on :java => ["1.7", :recommended]`

externl: :+1:

rleigh-dundee:
@externl With the latest changes, I get:

```
Error: undefined local variable or method `java_home' for #<Formula ice (stable) /usr/local/Library/Formula/ice.rb>
```

I do have java installed; the current 1.8 release from Oracle:

```
% javac -version
javac 1.8.0_31
% java -version
java version "1.8.0_31"
Java(TM) SE Runtime Environment (build 1.8.0_31-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.31-b07, mixed mode)
% which javac
/usr/bin/javac
% which java
/usr/bin/java
```

externl:
@rleigh-dundee Ok, I see. There are two issues. Do you have Java 7 installed? It should be using that. I suspect that `depends_on :java => ["1.7", :recommended]` still chooses Java 8 over 7. can @DomT4 comment on this. We require Java 7 because Berkeley DB does not compile with Java 8 without a patch.

DomT4:
`java_home` should be set automatically by the `:java` dependency.

externl:
Yes, I've removed the extra parts. Just relies on `depends_on :java => ["1.7", :recommended]` now.

rleigh-dundee:
@externl Just checked; I have 1.7.0_51 and 1.8.0_31 in /Library/Java/JavaVirtualMachines.  If 1.7 is required to build, I would expect it to degrade gracefully if 1.8 is installed and not build the java bindings (if that's possible).

externl:
@rleigh-dundee If Java 7 is not installed you will get this error:
```
ice: Java 1.7 is required to install this formula.
You can install with Homebrew Cask:
  brew install Caskroom/cask/java

You can download from:
  http://www.oracle.com/technetwork/java/javase/downloads/index.html
Error: An unsatisfied requirement failed this build.
```

You can then use the options `--without-java` to install without Java. If you only havea Java 8 installed and would rather not install 7, ZeroC has a tap [here](https://github.com/zeroc-ice/homebrew-tap) which has a patched berkeley-db53 formula.

rleigh-dundee:
Thanks, that all seems fine I think.  I just did a build of your latest push and it all seems to be working correctly for me.

externl:
@rleigh-dundee :+1:

DomT4:
@Homebrew/owners - Can someone else try to `pull` this? Getting failure here.

```
==> Patch closes issue #41079
Error: Failure while executing: /usr/bin/curl -f#LA Homebrew\ 0.9.5\ (Ruby\ 2.0.0-481;\ OS\ X\ 10.10.4) --silent --fail -o /dev/null -I https://github.com/BrewTestBot/homebrew/compare/homebrew:master...pr-41079
```

externl:
I can rebase on top of master if necessary.

DomT4:
To be honest I'm not quite sure what the failure is. Our test bot has record of the build passing, and the downstream build passing. Usually when it fails to pull one of those steps has failed, this time neither did. I'm hoping it's just a local issue for me.

externl:
Any update on getting this merged?

Library/Formula/ice.rb
```diff
@@ -1,38 +1,24 @@
...
 
   bottle do
     sha1 "52d43a08202206258c6fa33750aac25d2dc38703" => :yosemite
-    sha1 "df1279e47c8c213a57d04318443605b2871269a7" => :mavericks
-    sha1 "7bdb347052b8df900c2c2e6797726a79d9c0242d" => :mountain_lion
   end
 
-  option 'doc', 'Install documentation'
-  option 'demo', 'Build demos'
+  depends_on :macos => :yosemite
```
rleigh-dundee: Should this be `:mavericks` to allow building on 10.9 as well?

externl:
Yosemite is the only officially supported OS for Ice 3.6.0, we have not tested on mavericks.

mikemcquaid:
@externl Unless it's known to be broken on older versions (please provide issue URLs) please leave it as-is, thanks.

rleigh-dundee:
@externl I think we need to distinguish here between what ZeroC *officially* supports, and what's *functional in practice*.  For Ice 3.5.1 I patched the hardcoding in the build files so it would build on 10.8+, and it worked just fine.  I can understand 10.8 not being supported for 3.6 due to the buggy toolchain, but unless it's known to be broken with 10.9, please let it build there.  In the event there are problems, then we can identify them and patch them--as we did for the earlier Ice releases.

While I do very much appreciate the effort that ZeroC is going to in contributing to the homebrew formula, this change will be quite detrimental.  JFTR, our entire team is using the homebrew packaging for Ice across 10.8, 10.9 and 10.10 (we test and support our code on all three platforms), and this change will drop support for two of those platforms.  In part, we are using the homebrew packaging precisely *because* the Mac builds of previous Ice releases have been incompatible across the range of versions we were supporting at the time, by only supporting the latest version, while homebrew was vastly more flexible.  This is why we have contributed developer time to maintenance of the formula.


Thanks for your consideration,
Roger

externl:
The only reason I added `:mavericks` was becuase @rleigh-dundee asked earlier, maybe I misred his message :smile: . I will remove it, it's not a big deal.

rleigh-dundee:
@externl Sorry if there's been any confusion with the naming.  I can't find my original comment now.  If it won't build on 10.8, you might need the minimum version check, just changed to mavericks rather than yosemite.

externl:
@rleigh-dundee Seems like everything is green with no version check! Thanks for testing!

Library/Formula/ice.rb
```diff
@@ -1,38 +1,23 @@
...
-  revision 1
+  desc "A comprehensive RPC framework"
+  homepage "https://zeroc.com"
+  url "https://github.com/zeroc-ice/ice/archive/v3.6.0.tar.gz"
+  sha256 "77933580cdc7fade0ebfce517935819e9eef5fc6b9e3f4143b07404daf54e25e"
 
   bottle do
     sha1 "52d43a08202206258c6fa33750aac25d2dc38703" => :yosemite
-    sha1 "df1279e47c8c213a57d04318443605b2871269a7" => :mavericks
-    sha1 "7bdb347052b8df900c2c2e6797726a79d9c0242d" => :mountain_lion
```
mikemcquaid: Leave these as-is, thanks.

Library/Formula/ice.rb
```diff
@@ -1,38 +1,23 @@
...
   end
 
-  option 'doc', 'Install documentation'
-  option 'demo', 'Build demos'
+  depends_on "mcpp"
+  depends_on :java => ["1.7", :recommended]
```
mikemcquaid: Let's make this `:optional` as Java doesn't ship with OSX.

Library/Formula/ice.rb
```diff
@@ -42,17 +27,17 @@ def install
...
       args = %W[
         --disable-debug
         --prefix=#{libexec}
         --mandir=#{libexec}/man
         --enable-cxx
-        --enable-compat185
       ]
 
+      if build.with? "java"
+        args << "--enable-java"
```
mikemcquaid: `args << "--enable-java" if build.with? "java"`

Library/Formula/ice.rb
```diff
@@ -60,152 +45,56 @@ def install
       end
     end
 
-    ENV.O2
+    inreplace "cpp/src/slice2js/Makefile" do |s|
```
mikemcquaid: `inreplace "cpp/src/slice2js/Makefile", /install:/, "dontinstall:"`

mikemcquaid: And the same for the `inreplace`s below.

Library/Formula/ice.rb
externl: Whoops, my bad.

Library/Formula/ice.rb
externl: The IceGrid GUI application is a Java application and is required to administer IceGrid deployments. It would be desirable to keep it installed by default :smile:.

mikemcquaid: Homebrew is primarily for installing non-GUI applications so that's another reason to make it non-default, I'm afraid. Thanks!

DomT4:
Still dies on Mountain Lion. Since there's actual breakage on ML we're happy for you to cap it to Mavericks and above if you'd like.

externl:
@DomT4  Still, wasn't it passing before?

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/274708c9216097e07a9a88835f2a67d6d305a66b. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #41082, opened 2015-06-25
pdf2svg: use https in test file download

vszakats:


xu-cheng:
Can you replace the PDF file from remote to the one already included in Homebrew? You can use `test_fixtures("test.pdf")`. Thanks.

vszakats:
Good idea! updated and squashed.

mikemcquaid:
Thanks @vszakats!

PR #41083, opened 2015-06-25
Improve Elasticsearch’s plist by pointing it to the *persistent* configuration.

MerelyAPseudonym:


MerelyAPseudonym:
Btw, I'd guess that `/usr/local/opt/elasticsearch/config/elasticsearch.yml` is a viable alternative to `/usr/local/etc/elasticsearch/elasticsearch.yml`. It's not obvious to me that one is better than the other.

mikemcquaid:
Thanks @MerelyAPseudonym!

PR #41090, opened 2015-06-25
pythons: clean up bundled packages

tdsmith:
* Install setuptools and pip as flat packages, not eggs, to avoid problems with multiple installed versions
* Bundle wheel to allow pip to cache built packages and bump revision to avoid breaking postinstall
* Remove unnecessary settings from distutils.cfg
* Bump setuptools to 18.0.1


mikemcquaid:
:+1:

tdsmith:
@BrewTestBot test this please

PR #41095, opened 2015-06-26
xcode: update clang version

DomT4:
Confirmed in a VM:

```
~> clang --version
Apple LLVM version 7.0.0 (clang-700.0.53.3)
Target: x86_64-apple-darwin15.0.0
Thread model: posix
```

I presume we carry the very minor `.3` on the end given that's what Clang says it is?

### DIFF
--- a/Library/Homebrew/os/mac/xcode.rb
+++ b/Library/Homebrew/os/mac/xcode.rb
@@ -119,7 +119,8 @@ module OS
           when 51      then "5.1"
           when 60      then "6.0"
           when 61      then "6.1"
-          else "6.1"
+          when 70      then "7.0"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
I think so but will wait for another @Homebrew/owners to double-:+1:

jacknagel:
I don't recall ever seeing a 4th version number before, but this seems like the correct thing to do.

DomT4:
I think we hit something similar very early in the 10.10 DPs as well, but I can barely remember last week let alone last year so :wink:.

Merged in https://github.com/Homebrew/homebrew/commit/11cf300c01e3c43fc9604b87072ae12d4208a410.

PR #41104, opened 2015-06-26
fonttools 2.5 (new formula)

blackgear:
> a tool to convert OpenType and TrueType fonts to and from XML. FontTools is a library for manipulating fonts, written in Python. It supports TrueType, OpenType, AFM and to an extent Type 1 and some Mac-specific formats.

Fonttools is a powerful tool library:
- ttx: convert fonts to and from XML
- pyftmerge: merge and combine fonts
- pyftsubset: make a subset of a font
- pyftinspect: a gui inspecter for fonts

This is program that a user will regularly interact with directly and not provided by PyPi.

### DIFF
--- a/Library/Formula/fonttools.rb
+++ b/Library/Formula/fonttools.rb
@@ -1,0 +1,26 @@
+class Fonttools < Formula
+  desc "FontTools is a library for manipulating fonts"
+  homepage "https://github.com/behdad/fonttools"
+  url "https://github.com/behdad/fonttools/archive/2.5.tar.gz"
+  sha256 "c89603f9f3346f48f4a24f786422e935423554e0a9172dcd3ec8ffbd556d2159"
+  head "https://github.com/behdad/fonttools.git"
+
+  option "with-pygtk", "Build with pygtk support for pyftinspect"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on "pygtk" => :optional
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages/FontTools"
+
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir["#{libexec}/bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    cp "/Library/Fonts/Arial.ttf", testpath
+    system bin/"ttx", "Arial.ttf"
+  end
+end


Library/Formula/fonttools.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    bin.install Dir["#{libexec}/bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    cp "/Library/Fonts/Arial.ttf", testpath
+    system "ttx", "Arial.ttf"
+  end
+
```
vszakats: Minor nit: please delete this empty line.

Library/Formula/fonttools.rb
```diff
@@ -0,0 +1,27 @@
...
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir["#{libexec}/bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    cp "/Library/Fonts/Arial.ttf", testpath
+    system "ttx", "Arial.ttf"
```
xu-cheng: `bin/"ttx"`

Library/Formula/fonttools.rb
```diff
@@ -0,0 +1,27 @@
...
+  head "https://github.com/behdad/fonttools.git"
+
+  option "with-pygtk", "Build with pygtk support for pyftinspect"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on "pygtk" => :optional
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages/FontTools"
```
DomT4: You shouldn't need this line, I believe?

blackgear: @DomT4 
In the setup.py, the author use `extra_path = 'FontTools',` to put all core files in `site-packages/FontTools`, without this line, the program can't run at all.

But I found that I can remove line 14, this line is useless.

DomT4: Thanks for the explanation. Will let @tdsmith comment on whether the unusual directory is a problem, but given we point the executable at that path it should be okay I'd have thought.

tdsmith: lgtm

blackgear:
This formula is ready to be merged

DomT4:
LGTM. Please give us a few days before considering pinging us on merging PRs, we're a fairly small team covering a lot of PRs/Issues/codebase tweaks/etc and new formula are allowed to sit for a short period so the team can look over them as desired.

Merged in https://github.com/Homebrew/homebrew/commit/1ccaf548c22b73fee36994b8552a791380690330. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #41108, opened 2015-06-26
dovecot 2.2.18: add pam build option, logging in plist

dustinwilson:
Updated to version 2.2.18, added support for a PAM build option, and added logging to the plist; it doesn't log anything otherwise.


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/73a3e9fb4e51acfa6f3049cbd304a42593314309. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just for reference, when we change things alongside the version bump the commit message still prefers the version change, so `dovecot 2.2.18` would have been fine here. Thanks!

PR #41109, opened 2015-06-26
lesspipe : changend to use all kind of decompressor.

masaruyokoi:
before the configure script will check the decompressor which is installed on the build
machine. but now skip the checker in configure script.

### DIFF
--- a/Library/Formula/lesspipe.rb
+++ b/Library/Formula/lesspipe.rb
@@ -13,7 +13,7 @@ class Lesspipe < Formula
       inreplace 'configure', %q{$ifsyntax = "\L$ifsyntax";}, %q{$ifsyntax = "\Ly";}
     end
 
-    system "./configure", "--prefix=#{prefix}", "--default"
+    system "./configure", "--prefix=#{prefix}", "--yes"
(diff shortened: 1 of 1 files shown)


DomT4:
What's the upside/downside to this change?

masaruyokoi:
Hello DomT4,

The configure option "--default" in upper side will find decompressor (gzip, bzip2, xz, etc) from local PATH. if the decompressor doesn't find in the system, the feature will not be included in the lesspipe.sh by macro .
But if we set "--yes" option instead of "--default", all of decompressor will be included in lesspipe.sh. and if the decompressor is not installed in the system which it runs, lesspipe don't try to execute the decompressor. So I think the option "--yes" is better for binary distributions.

DomT4:
Alright, sounds good since it doesn't cause breakage for users even without those decompressors available. Merged in https://github.com/Homebrew/homebrew/commit/62eabd311ca355887ccb3b457319f663921a9820.

Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Please note the commit message should be <50 characters in future, Thanks!

PR #41122, opened 2015-06-27
ack: fix devel version

dunn:
Was scanned as 15.01.  I figure this is better than `2.15_01` since underscores are usually reserved for formula revisions, right?


xu-cheng:
Thanks @dunn

PR #41123, opened 2015-06-27
Add Docker HEAD version and option to enable experimental features

ahbeng:
Since `#{version}` is used in `bin.install`, for `HEAD` builds, it has to be read from the `VERSION` file, at the root of the repository, which is [used by the build scripts](https://github.com/docker/docker/blob/master/hack/make.sh#L11).

Reference for Docker experimental: [blog post](https://blog.docker.com/2015/06/experimental-binary/) and [documentation](https://github.com/docker/docker/tree/master/experimental#install-docker-experimental).


mikemcquaid:

For future reference the preferred commit message format is `docker: add HEAD version.` and `docker: add experimental features option.`.

PR #41128, opened 2015-06-27
couchdb: support Erlang/OTP-18.0 with patch from upstream

dch:
this is a pre-requisite to committing #41035.

```
$ brew audit --strict couchdb
==> brew style couchdb

```

### DIFF
--- a/Library/Formula/couchdb.rb
+++ b/Library/Formula/couchdb.rb
@@ -3,7 +3,13 @@ class Couchdb < Formula
   homepage "https://couchdb.apache.org/"
   url "https://www.apache.org/dyn/closer.cgi?path=/couchdb/source/1.6.1/apache-couchdb-1.6.1.tar.gz"
   sha256 "5a601b173733ce3ed31b654805c793aa907131cd70b06d03825f169aa48c8627"
-  revision 2
+  revision 3
+
+  stable do
+    # Support Erlang/OTP 18.0 compatibility, see upstream #95cb436
+    # It will be in the next CouchDB point release, likely 1.6.2.
+    patch :DATA
+  end
 
   bottle do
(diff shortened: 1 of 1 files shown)


makigumo:
Why include the whole patch when a link would have sufficed? More something like this.

```ruby
  stable do
    patch do
      # Support Erlang/OTP 18.0 compatibility, see upstream #95cb436
      # It will be in the next CouchDB point release, likely 1.6.2.
      url "https://github.com/apache/couchdb/commit/95cb436.diff"
      sha256 "143114b33f4f134994c4824197e297f1860aa01e8de2c0f7f5e6ca93920a90c6"
    end
  end
```

dch:
@makigumo because the commit is for the git source tree, before `make distcheck` produces the official tarball. The patch has to account for differences over and above the commit. But well spotted ;-)

makigumo:
Ah, I see now, that there is am extra section for ```configure``` at the end. Thanks for the explanation :)

Library/Formula/couchdb.rb
```diff
@@ -32,7 +38,7 @@ def install
     # in the welcome message
     inreplace "etc/couchdb/default.ini.tpl.in" do |s|
       s.gsub! "%package_author_name%", "Homebrew"
-      s.gsub! "%version%", "%version%-1"
+      s.gsub! "%version%", "%version%-3"
```
mikemcquaid: Can you use `pkg_version` here instead which gets you the version plus revision?

dch: done, produces `version = 1.6.1_3` which seems super reasonable & easy - thanks for the #protip @mikemcquaid

dch:
updated with version = 1.6.1_3 / #{pkg_version} as proposed.

dch:
`brew audit --strict` now passes. yay for CI. pushed again.

mikemcquaid:
Thanks @dch!

PR #41129, opened 2015-06-27
Add the graphite2 smart font rendering engine.

simoncozens:
[Graphite](http://scripts.sil.org/cms/scripts/page.php?site_id=projects&item_id=graphite_home) is a library for rendering complex non-Roman fonts. It's already part of Libreoffice, Firefox, Thunderbird, XeTeX and so on, etc.

Harfbuzz (which is [in Homebrew](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/harfbuzz.rb)) has Graphite2 as an optional dependency; it likes to have Graphite2 as an additional shaper to call out to in order to render more complex scripts. 

Currently there's no way to build the Harfbuzz package in Homebrew with Graphite2 support; even if Graphite2 is installed manually it's ignored. If the graphite2 formula is accepted, then I'll also send a patch adding a `--with-graphite=auto` configuration directive to Harfbuzz, so that it can pick up on the Graphite2 library if it's installed.

### DIFF
--- a/Library/Formula/graphite2.rb
+++ b/Library/Formula/graphite2.rb
@@ -19,7 +19,7 @@ class Graphite2 < Formula
 
   test do
     resource("testfont").stage do
-      shape = `gr2fonttest Simple-Graphite-Font.ttf "abcde"`
+      shape = `#{bin}/gr2fonttest Simple-Graphite-Font.ttf "abcde"`
       assert_match /67.*36.*37.*38.*71/m, shape
     end
   end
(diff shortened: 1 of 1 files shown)


Library/Formula/graphite2.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def install
+    args = std_cmake_args
+    system "cmake", ".", *args
+    system "make", "install"
+  end
+
+  test do
+    resource("testfont").stage do
+      shape = `gr2fonttest Simple-Graphite-Font.ttf "abcde"`
```
dunn: If `gr2fonttest` is something the formula installs into `bin`, then here it should be `#{bin}/gr2fonttest` to make sure we test the right one.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/06c5f15ec6eaab20258ff9b3f39b0ec69f45baa3. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a note for future: New formulae commit messages should be styled `graphite2 1.2.4 (new formula)` or similar, Thanks!

PR #41138, opened 2015-06-27
exiv2 0.25

dunn:
The HEAD install was just adapted from `bootstrap.linux` in the SVN repo, so let me know if there are any other tweaks I should make.

### DIFF
--- a/Library/Formula/exiv2.rb
+++ b/Library/Formula/exiv2.rb
@@ -1,23 +1,62 @@
-require 'formula'
-
 class Exiv2 < Formula
   desc "EXIF and IPTC metadata manipulation library and tools"
-  homepage 'http://www.exiv2.org'
-  url 'http://www.exiv2.org/exiv2-0.24.tar.gz'
-  sha1 '2f19538e54f8c21c180fa96d17677b7cff7dc1bb'
+  homepage "http://www.exiv2.org"
+  url "http://www.exiv2.org/exiv2-0.25.tar.gz"
+  sha256 "c80bfc778a15fdb06f71265db2c3d49d8493c382e516cb99b8c9f9cbde36efa4"
 
   bottle do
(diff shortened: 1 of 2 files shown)


Library/Formula/exiv2.rb
```diff
@@ -1,23 +1,61 @@
...
   end
 
   option :universal
 
   def install
     ENV.universal_binary if build.universal?
-    system "./configure", "--prefix=#{prefix}", "--disable-dependency-tracking"
-    system "make install"
+    if build.head?
+      args = %W[-DCMAKE_BUILD_TYPE=Release
```
DomT4: We pass a lot of these automatically with `std_cmake_args`. Let's use that and then add extra `args` as necessary.

Library/Formula/exiv2.rb
```diff
@@ -1,23 +1,60 @@
...
 
   option :universal
 
   def install
     ENV.universal_binary if build.universal?
-    system "./configure", "--prefix=#{prefix}", "--disable-dependency-tracking"
-    system "make install"
+    if build.head?
+      args = std_cmake_args
+      args += %W[-DEXIV2_ENABLE_SHARED=ON
```
DomT4: It's a nit, but can we shove this onto the next line so the indentation doesn't look so out of place?

```ruby
args += %W[
  -DEXIV2_ENABLE_SHARED=ON
  -DEXIV2_ENABLE_XMP=ON
  -DEXIV2_ENABLE_LIBXMP=ON
]
```

And such.

dunn: Sure!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/114ca6f98a2c6488b3b7128553d0ffb271a42ebc. Thanks again @dunn!

PR #41141, opened 2015-06-27
gnuplot 5.0.1

dunn:


Library/Formula/gnuplot.rb
```diff
@@ -45,18 +44,18 @@ class Gnuplot < Formula
 
   depends_on "pkg-config" => :build
   depends_on LuaRequirement if build.with? "lua"
```
DomT4: @Homebrew/owners - Do we still want to support the `LuaRequirement`? It was added in 2013 by Adam [here](https://github.com/Homebrew/homebrew/commit/827ca5462a6799081e72fefa4762f5d6d725bfd0) to support building against either Lua51 or Lua52.

At the time the core shipped Lua51 and Versions had Lua52, but both symlinked into the main `HOMEBREW_PREFIX` as `lua` by default.

Since we moved to supporting both Lua51 and Lua52 in the core that no longer happens for obvious reasons, and now the only `lua` binary you can get from the core is the `lua` (lua52) package and the `lua53` formula in Versions is tucked behind a `--with-default-names` option.

IMO we could just make this `depends_on "lua" => :optional` and perhaps add `depends_on "lua51" => :optional` as well to get rid of the `LuaRequirement`.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/f2696b059f843a865b47148aa8a54d3c36ca9c9a. Thanks @dunn!

PR #41146, opened 2015-06-28
help2man 1.47.1

dunn:
- replaced the gnu mirror with another https source
- ~~didn't need to deparallelize, so that might be fixed~~

### DIFF
--- a/Library/Formula/help2man.rb
+++ b/Library/Formula/help2man.rb
@@ -1,9 +1,9 @@
 class Help2man < Formula
   desc "Automatically generate simple man pages"
   homepage "https://www.gnu.org/software/help2man/"
-  url "http://ftpmirror.gnu.org/help2man/help2man-1.46.5.tar.xz"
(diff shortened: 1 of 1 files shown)


DomT4:
For better or worse we keep GNU formula pointed at the `ftpmirror` URL as that's where you get pointed at by upstream.

dunn:
Oh, gotcha

DomT4:
Eventually, once I've had a chance to chat to the GNU team about using their core `https://ftp.gnu.org` mirror more heavily and written up the code to fallback on secure mirrors when `HOMEBREW_NO_INSECURE_REDIRECT` is set the `http` will become less of an issue for those who want it to be less of an issue.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/08f48d56fa438a5746fe2c778f097c4ed5d0c3e0. Thanks Alex!

PR #41149, opened 2015-06-28
rocksdb: Add --with-lite option

theopolis:
This adds an optional `--with-lite` to the from-source build of RocksDB.

RocksDB (librocksdb) can be included as an embedded database for applications running on memory constrained platforms like iOS using the LITE version. This version is regression-tested to maintain size and in practice requires significantly less private memory. See [ROCKSDB_LITE](https://github.com/facebook/rocksdb/blob/master/ROCKSDB_LITE.md) for the complete guide/vision.

Example build:
```
brew install --build-bottle --build-from-source -v --with-lite ./Library/Formula/rocksdb.rb
```

And in the including/linking application: `CXX_FLAGS+=-DROCKSDB_LITE`.

### DIFF
--- a/Library/Formula/rocksdb.rb
+++ b/Library/Formula/rocksdb.rb
@@ -14,6 +14,8 @@ class Rocksdb < Formula
     sha256 "ada906d1b4fa26f73ce3aa4b393c6c9d7ec76bf50663e2ff748d143f6c3518bc" => :mountain_lion
   end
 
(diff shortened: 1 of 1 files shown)


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/c373965a454af3508fa0265da3cc1980ae72b21f. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #41151, opened 2015-06-28
Build gflags as a shared library by default.

alexsmac:
- By default, newer versions of gflags that use CMake, build a static
  library, whereas previously (e.g. v2.0) it was built as a shared
  library.
- Building gflags statically causes issues when glog is built as a
  shared library (its default option).  Specifically, the gflags defined
  by glog will not be visible unless glog appears explicitly prior to
  gflags in the link step for an application (i.e. there are undefined
  symbols from gflags that need to be resolved).
- In fact, on Linux the linker will refuse to build glog (0.3.4) as a
  shared library if gflags (2.1.2) is built as a static library due to
  gflags not being position independent.

### DIFF
--- a/Library/Formula/gflags.rb
+++ b/Library/Formula/gflags.rb
@@ -13,7 +13,7 @@ class Gflags < Formula
   end
 
   option "with-debug", "Build debug version"
-  option "with-static", "Build gflags as a static library."
+  option "with-static", "Build gflags as a static (instead of shared) library."
 
   depends_on "cmake" => :build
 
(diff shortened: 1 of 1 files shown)


Library/Formula/gflags.rb
```diff
@@ -13,13 +13,20 @@ class Gflags < Formula
   end
 
   option "with-debug", "Build debug version"
+  option "with-static", "Build gflags as a static library."
```
mikemcquaid: might want to add `(instead of shared)` just to be super-explicit.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/3bd2d350d8b062748b9e999837f3f57a5c140772. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #41156, opened 2015-06-28
Update Tesseract repository (moved to Github)

francoisbruneau:
A couple of URL updates for those interested in trying out the HEAD version of Tesseract and its recognition data.


Library/Formula/tesseract.rb
```diff
@@ -22,7 +22,7 @@ class Tesseract < Formula
   end
 
   head do
-    url 'https://code.google.com/p/tesseract-ocr/', :using => :git
+    url 'https://github.com/tesseract-ocr/tesseract/', :using => :git
```
DomT4: Can just make this `url "https://github.com/tesseract-ocr/tesseract.git"` now.

Library/Formula/tesseract.rb
```diff
@@ -128,6 +128,10 @@ class Tesseract < Formula
     sha1 "a559801a42be8a3dc15ae2ec27da99bd8b0b85ac"
   end
 
+  resource "tessdata-head" do
+    url 'https://github.com/tesseract-ocr/tessdata/', :using => :git
```
DomT4: Likewise. `.git` here, not `:using => :git`.

Library/Formula/tesseract.rb
francoisbruneau: Done, thanks!

Library/Formula/tesseract.rb
francoisbruneau: Fixed!

Library/Formula/tesseract.rb
```diff
@@ -128,6 +128,10 @@ class Tesseract < Formula
     sha1 "a559801a42be8a3dc15ae2ec27da99bd8b0b85ac"
   end
 
+  resource "tessdata-head" do
```
DomT4: Let's tuck this inside the `head do` block just to make it clear it only applies to the `head` logic. You should be able to drop the `resources.reject { |r| r.name == "tessdata-head" }.each` dance below as well then.

Library/Formula/tesseract.rb
```diff
@@ -128,6 +128,10 @@ class Tesseract < Formula
...
+    url 'https://github.com/tesseract-ocr/tessdata.git'
```
DomT4: And double quotes around here. Everything should be wrapped in double quotes, we've been moving across more strictly to the Ruby style guide, but new lines in particular shouldn't reinforce the single quotes :)

PR #41158, opened 2015-06-28
New formula: openvdb-3.0.0

jehutymax:
I made extensive use of inreplace instead of passing Makefile arguments along with "make", "install" because given the number of variables and the fact that some needed to be set to empty, it quickly became unwieldy.

### DIFF
--- a/Library/Formula/openvdb.rb
+++ b/Library/Formula/openvdb.rb
@@ -10,11 +10,16 @@ class Openvdb < Formula
   option "with-logging", "Requires log4cplus"
   option "with-docs", "Installs documentation"
 
-  depends_on "boost"
   depends_on "openexr"
   depends_on "ilmbase"
   depends_on "tbb"
   depends_on "jemalloc" => :recommended
+
+  if MacOS.version < :mavericks
+    depends_on "boost" => "c++11"
+  else
+    depends_on "boost"
+  end
 
   depends_on "homebrew/versions/glfw3" if build.with? "viewer"
   depends_on "cppunit" if build.with? "tests"
--- a/Library/Formula/openvdb.rb
+++ b/Library/Formula/openvdb.rb
@@ -1,96 +1,97 @@
 class Openvdb < Formula
-  desc "Sparse volume processing toolkit"
-  homepage "http://www.openvdb.org/"
-  url "https://github.com/dreamworksanimation/openvdb/archive/v3.0.0.tar.gz"
-  sha256 "6c90cfda032c54876b321031717c13ea56a6b7b15c911d3edfbb2ad7af49700e"
-  head "https://github.com/dreamworksanimation/openvdb.git"
+  desc 'Sparse volume processing toolkit'
+  homepage 'http://www.openvdb.org/'
+  url 'https://github.com/dreamworksanimation/openvdb/archive/v3.0.0.tar.gz'
+  sha256 '6c90cfda032c54876b321031717c13ea56a6b7b15c911d3edfbb2ad7af49700e'
+  head 'https://github.com/dreamworksanimation/openvdb.git'
 
-  bottle do
-    sha256 "843a6b55c84185101c1e1b379751e99cd0215f2fce7534ef3a2387a33a7c91a5" => :yosemite
-  end
+  option 'with-viewer', 'Installs the command-line tool to view OpenVDB files'
+  option 'with-tests', 'Installs the unit tests for the OpenVDB library'
+  option 'with-logging', 'Requires log4cplus'
+  option 'with-docs', 'Installs documentation'
 
-  option 'tests', 'Installs and runs the unit tests for the OpenVDB library'
-  option 'logging', 'Requires log4cplus'
-  option 'docs', 'Installs documentation'
+  depends_on 'boost'
+  depends_on 'openexr'
+  depends_on 'ilmbase'
+  depends_on 'tbb'
+  depends_on 'jemalloc' => :recommended
 
-  depends_on "boost"
-  depends_on "openexr"
-  depends_on "ilmbase"
-  depends_on "tbb"
-  depends_on "jemalloc" => :recommended
-  depends_on "homebrew/versions/glfw3" => :recommended
+  depends_on 'homebrew/versions/glfw3' if build.with? 'viewer'
+  depends_on 'cppunit' if build.with? 'tests'
+  depends_on 'doxygen' if build.with? 'docs'
+  depends_on 'log4cplus' if build.with? 'logging'
+  needs :cxx11
 
-  if build.with? "tests"
-    depends_on "cppunit"
-  end
-
-  if build.with? "docs"
-    depends_on "doxygen"
-  end
-
-  if build.with? "logging"
-    depends_on "log4cplus"
-  end
-
-  resource "test_file" do
-    url "http://www.openvdb.org/download/models/cube.vdb.zip"
-    sha256 "05476e84e91c0214ad7593850e6e7c28f777aa4ff0a1d88d91168a7dd050f922"
+  resource 'test_file' do
+    url 'http://www.openvdb.org/download/models/cube.vdb.zip'
+    sha256 '05476e84e91c0214ad7593850e6e7c28f777aa4ff0a1d88d91168a7dd050f922'
   end
 
   def install
+    ENV.cxx11
     # Adjust hard coded paths in Makefile
-    inreplace "openvdb/Makefile", "/tmp/OpenVDB", "#{prefix}"
-    inreplace "openvdb/Makefile", "BOOST_INCL_DIR := $(HT)/include", "BOOST_INCL_DIR:=#{Formula["boost"].opt_lib}/include"
-    inreplace "openvdb/Makefile", "BOOST_LIB_DIR := $(HDSO)", "BOOST_LIB_DIR:=#{Formula["boost"].opt_lib}"
-    inreplace "openvdb/Makefile", "TBB_INCL_DIR := $(HT)/include", "TBB_INCL_DIR:=#{Formula["tbb"].opt_lib}/include"
-    inreplace "openvdb/Makefile", "TBB_LIB_DIR := $(HDSO)", "TBB_LIB_DIR:=#{Formula["tbb"].opt_lib}/lib"
-    inreplace "openvdb/Makefile", "EXR_INCL_DIR := $(HT)/include", "EXR_INCL_DIR:=#{Formula["openexr"].opt_lib}/include"
-    inreplace "openvdb/Makefile", "EXR_LIB_DIR := $(HDSO)", "EXR_LIB_DIR:=#{Formula["openexr"].opt_lib}/lib"
-    inreplace "openvdb/Makefile", "-lboost_thread", "-lboost_thread-mt"
+    if build.with? 'jemalloc'
+      jemalloc = "CONCURRENT_MALLOC_LIB_DIR=#{Formula['jemalloc'].opt_lib}/lib"
+    else
+      jemalloc = "CONCURRENT_MALLOC_LIB="
+    end
 
-    inreplace "openvdb/Makefile", "PYTHON_VERSION := 2.6", "PYTHON_VERSION :="
-    inreplace "openvdb/Makefile", "NUMPY_INCL_DIR := /rel/map/generic-2013.22/include", "NUMPY_INCL_DIR :="
+    if build.with? "viewer"
+      glfw3 = "GLFW_INCL_DIR=#{Formula['homebrew/versions/glfw3'].opt_lib}/include"
+      glfw3_lib = "GLFW_LIB_DIR=#{Formula['homebrew/versions/glfw3'].opt_lib}/lib"
+      glfw3_libname =  'GLFW_LIB=-lglfw3'
+    else
+      glfw3 =  "GLFW_INCL_DIR="
+      glfw3_lib = "GLFW_LIB_DIR="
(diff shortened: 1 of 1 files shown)


Library/Formula/openvdb.rb
```diff
@@ -0,0 +1,100 @@
+class Openvdb < Formula
+  desc "Sparse volume processing toolkit"
+  homepage "http://www.openvdb.org/"
+  url "https://github.com/dreamworksanimation/openvdb/archive/v3.0.0.tar.gz"
+  sha256 "6c90cfda032c54876b321031717c13ea56a6b7b15c911d3edfbb2ad7af49700e"
+  head "https://github.com/dreamworksanimation/openvdb.git"
+
+  bottle do
```
DomT4: Please don't add a bottle. We'll do this when we pull.

Library/Formula/openvdb.rb
```diff
@@ -0,0 +1,100 @@
...
+    sha256 "843a6b55c84185101c1e1b379751e99cd0215f2fce7534ef3a2387a33a7c91a5" => :yosemite
+  end
+
+  option 'tests', 'Installs and runs the unit tests for the OpenVDB library'
```
DomT4: All options should be `with-` or `without-`

Library/Formula/openvdb.rb
```diff
@@ -0,0 +1,100 @@
...
+  option 'logging', 'Requires log4cplus'
```
DomT4: Double quotes everywhere.

Library/Formula/openvdb.rb
```diff
@@ -0,0 +1,100 @@
...
+  option 'docs', 'Installs documentation'
+
+  depends_on "boost"
+  depends_on "openexr"
+  depends_on "ilmbase"
+  depends_on "tbb"
+  depends_on "jemalloc" => :recommended
+  depends_on "homebrew/versions/glfw3" => :recommended
```
DomT4: Can't have mandatory or recommended deps from taps in core formulae. This will need to be `optional`.

Library/Formula/openvdb.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  if build.with? "tests"
+    depends_on "cppunit"
```
DomT4: This can be one line. `depends_on "cppunit" if build.with? "tests"`

Library/Formula/openvdb.rb
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  if build.with? "docs"
+    depends_on "doxygen"
```
DomT4: This can be one line.

Library/Formula/openvdb.rb
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  if build.with? "logging"
+    depends_on "log4cplus"
```
DomT4: This can be one line.

Library/Formula/openvdb.rb
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  resource "test_file" do
+    url "http://www.openvdb.org/download/models/cube.vdb.zip"
+    sha256 "05476e84e91c0214ad7593850e6e7c28f777aa4ff0a1d88d91168a7dd050f922"
+  end
+
+  def install
+    # Adjust hard coded paths in Makefile
```
DomT4: Rather than adding the `inreplace "openvdb/Makefile"` line *every* time here you can just do this:

```ruby
inreplace "openvdb/Makefile" do |s|
  s.gsub! "something", "something else"
  s.gsub! "something", "something else"
  s.gsub! "something", "something else"
end
```

And so on.

Library/Formula/openvdb.rb
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  if build.with? "glfw3"
+    inreplace "openvdb/Makefile", "-lglfw", "-lglfw3"
+  else
+    inreplace "openvdb/Makefile", "-lglfw", ""
+  end
+
+  if build.without? "doxygen"
+  inreplace "openvdb/Makefile", "DOXYGEN := doxygen", "DOXYGEN :="
```
DomT4: Indentation is wrong here.

Library/Formula/openvdb.rb
```diff
@@ -0,0 +1,100 @@
...
+    inreplace "openvdb/Makefile", "LOG4CPLUS_INCL_DIR := /rel/folio/log4cplus/log4cplus-1.0.3-latest/sys_include", "LOG4CPLUS_INCL_DIR:=#{Formula["log4cplus"].opt_lib}/include"
+    inreplace "openvdb/Makefile", "LOG4CPLUS_LIB_DIR := /rel/folio/log4cplus/log4cplus-1.0.3-latest/library", "LOG4CPLUS_LIB_DIR:=#{Formula["log4cplus"].opt_lib}/lib"
+  else
+    inreplace "openvdb/Makefile", "LOG4CPLUS_INCL_DIR := /rel/folio/log4cplus/log4cplus-1.0.3-latest/sys_include", "LOG4CPLUS_INCL_DIR:="
+  end
+
+  # Blosc is not yet supported.
+  inreplace "openvdb/Makefile", "BLOSC_INCL_DIR := $(HT)/include", "BLOSC_INCL_DIR :="
+
+  ENV.append_to_cflags "-std=c++11"
```
DomT4: This shouldn't be necessary. If it needs `cxx11` please add `needs :cxx11` to the main block and `ENV.cxx11` to `def install`.

Library/Formula/openvdb.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  ENV.append_to_cflags "-I #{buildpath}"
+
+  cd "openvdb"
```
DomT4:
```ruby
  cd "openvdb" do
    system "make", "install"
  end
end
```

DomT4:
The inreplacing here is incredibly drastic and unlikely to be acceptable in its current form. We should be able to pass some of this to `make`, or we'll need to take this upstream and ask them to provide a way to escape the harcoding without actually having to hack up various files to do so.

Library/Formula/openvdb.rb
```diff
@@ -0,0 +1,101 @@
...
+                              "BOOST_LIB_DIR=#{Formula["boost"].opt_lib}",
+                              "BOOST_THREAD_LIB=-lboost_thread-mt",
+                              "TBB_INCL_DIR=#{Formula["tbb"].opt_lib}/include",
+                              "TBB_LIB_DIR=#{Formula["tbb"].opt_lib}/lib",
+                              "EXR_INCL_DIR=#{Formula["openexr"].opt_lib}/include",
+                              "EXR_LIB_DIR=#{Formula["openexr"].opt_lib}/lib",
+                              "BLOSC_INCL_DIR=", # Blosc is not yet supported.
+                              "PYTHON_VERSION=",
+                              "NUMPY_INCL_DIR=",
+                              *jemalloc,
```
dunn: Older Ruby can't deal with anything following one of these exploded arrays.

jehutymax: Is there any workaround for that?

dunn: Can you join all those arrays into a single `args` array beforehand?

jehutymax: I'm not sure why, but when I do that, Clang receives these as parameters too, and then complains. For example,  clang: error: no such file or directory BOOST_INCL_DIR=...
And this happens for each one of them except the first, no matter what the first is. And passing them individually like this worked.

dunn: This works for me:
```diff
diff --git a/Library/Formula/openvdb.rb b/Library/Formula/openvdb.rb
index 15efc28..d7caea4 100644
--- a/Library/Formula/openvdb.rb
+++ b/Library/Formula/openvdb.rb
@@ -29,68 +29,60 @@ class Openvdb < Formula
 
   def install
     ENV.cxx11
+    args = []
     # Adjust hard coded paths in Makefile
     if build.with? "jemalloc"
-      jemalloc = "CONCURRENT_MALLOC_LIB_DIR=#{Formula["jemalloc"].opt_lib}/lib"
+      args << "CONCURRENT_MALLOC_LIB_DIR=#{Formula["jemalloc"].opt_lib}/lib"
     else
-      jemalloc = "CONCURRENT_MALLOC_LIB="
+      args << "CONCURRENT_MALLOC_LIB="
     end
 
     if build.with? "viewer"
-      glfw3 = "GLFW_INCL_DIR=#{Formula["homebrew/versions/glfw3"].opt_lib}/include"
-      glfw3_lib = "GLFW_LIB_DIR=#{Formula["homebrew/versions/glfw3"].opt_lib}/lib"
-      glfw3_libname =  "GLFW_LIB=-lglfw3"
+      args << "GLFW_INCL_DIR=#{Formula["homebrew/versions/glfw3"].opt_lib}/include"
+      args << "GLFW_LIB_DIR=#{Formula["homebrew/versions/glfw3"].opt_lib}/lib"
+      args << "GLFW_LIB=-lglfw3"
     else
-      glfw3 =  "GLFW_INCL_DIR="
-      glfw3_lib = "GLFW_LIB_DIR="
-      glfw3_libname = "GLFW_LIB="
+      args << "GLFW_INCL_DIR="
+      args << "GLFW_LIB_DIR="
+      args << "GLFW_LIB="
     end
 
     if build.with? "docs"
-      doxygen = "DOXYGEN=doxygen"
+      args << "DOXYGEN=doxygen"
     else
-      doxygen = "DOXYGEN="
+      args << "DOXYGEN="
     end
 
     if build.with? "tests"
-      cppunit = "CPPUNIT_INCL_DIR=#{Formula["cppunit"].opt_lib}/include"
-      cppunit_lib =  "CPPUNIT_LIB_DIR=#{Formula["cppunit"].opt_lib}/lib"
+      args << "CPPUNIT_INCL_DIR=#{Formula["cppunit"].opt_lib}/include"
+      args << "CPPUNIT_LIB_DIR=#{Formula["cppunit"].opt_lib}/lib"
     else
-      cppunit = "CPPUNIT_INCL_DIR="
-      cppunit_lib =  "CPPUNIT_LIB_DIR="
+      args << "CPPUNIT_INCL_DIR=" << "CPPUNIT_LIB_DIR="
     end
 
     if build.with? "logging"
-      log4cplus = "LOG4CPLUS_INCL_DIR=#{Formula["log4cplus"].opt_lib}/include"
-      log4cplus_lib =  "LOG4CPLUS_LIB_DIR=#{Formula["log4cplus"].opt_lib}/lib"
+      args << "LOG4CPLUS_INCL_DIR=#{Formula["log4cplus"].opt_lib}/include"
+      args << "LOG4CPLUS_LIB_DIR=#{Formula["log4cplus"].opt_lib}/lib"
     else
-      log4cplus = "LOG4CPLUS_INCL_DIR="
-      log4cplus_lib =  "LOG4CPLUS_LIB_DIR="
+      args << "LOG4CPLUS_INCL_DIR=" << "LOG4CPLUS_LIB_DIR="
     end
 
     ENV.append_to_cflags "-I #{buildpath}"
 
     cd "openvdb" do
-      system "make", "install", "DESTDIR=#{prefix}",
-                              "BOOST_INCL_DIR=#{Formula["boost"].opt_lib}/include",
-                              "BOOST_LIB_DIR=#{Formula["boost"].opt_lib}",
-                              "BOOST_THREAD_LIB=-lboost_thread-mt",
-                              "TBB_INCL_DIR=#{Formula["tbb"].opt_lib}/include",
-                              "TBB_LIB_DIR=#{Formula["tbb"].opt_lib}/lib",
-                              "EXR_INCL_DIR=#{Formula["openexr"].opt_lib}/include",
-                              "EXR_LIB_DIR=#{Formula["openexr"].opt_lib}/lib",
-                              "BLOSC_INCL_DIR=", # Blosc is not yet supported.
-                              "PYTHON_VERSION=",
-                              "NUMPY_INCL_DIR=",
-                              *jemalloc,
-                              *glfw3,
-                              *glfw3_lib,
-                              *glfw3_libname,
-                              *doxygen,
-                              *cppunit,
-                              *cppunit_lib,
-                              *log4cplus,
-                              *log4cplus_lib
+      system "make", "install",
+                     "DESTDIR=#{prefix}",
+                     "BOOST_INCL_DIR=#{Formula["boost"].opt_lib}/include",
+                     "BOOST_LIB_DIR=#{Formula["boost"].opt_lib}",
+                     "BOOST_THREAD_LIB=-lboost_thread-mt",
+                     "TBB_INCL_DIR=#{Formula["tbb"].opt_lib}/include",
+                     "TBB_LIB_DIR=#{Formula["tbb"].opt_lib}/lib",
+                     "EXR_INCL_DIR=#{Formula["openexr"].opt_lib}/include",
+                     "EXR_LIB_DIR=#{Formula["openexr"].opt_lib}/lib",
+                     "BLOSC_INCL_DIR=", # Blosc is not yet supported.
+                     "PYTHON_VERSION=",
+                     "NUMPY_INCL_DIR=",
+                     *args
     end
   end
 ```

DomT4: Yup, @dunn's fix should work.

jehutymax: Thanks for all the pointers @dunn and @DomT4  - I'll test this as soon as I can and commit the diff.

jehutymax:
Let me know if there's anything else I should do to this one!

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `openvdb 3.0.0 (new formula)`.

PR #41159, opened 2015-06-28
sfnt2woff 1.0 (new formula)

zacs:
sfnt2woff is a tool to convert fonts from ttf and otf into woff format, and vice versa

### DIFF
--- a/Library/Formula/sfnt2woff.rb
+++ b/Library/Formula/sfnt2woff.rb
@@ -1,0 +1,18 @@
+class Sfnt2woff < Formula
+  desc "Convert TTF and OTF fonts to WOFF format."
+  homepage "https://people.mozilla.org/~jkew/woff/"
+  url "https://people.mozilla.org/~jkew/woff/woff-code-latest.zip"
+  version "1.0"
+  sha256 "7713630d2f43320a1d92e2dbb014ca3201caab1dd4c0ab92816016824c680d96"
+
+  def install
+    system "make"
+    bin.install "sfnt2woff"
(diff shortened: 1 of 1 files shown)


Library/Formula/sfnt2woff.rb
```diff
@@ -0,0 +1,18 @@
+class Sfnt2woff < Formula
+  desc "Convert TTF and OTF fonts to WOFF format."
+  homepage "https://people.mozilla.org/~jkew/woff/"
+  url "https://people.mozilla.org/~jkew/woff/woff-code-latest.zip"
```
DomT4: We can't use a `latest` zip because it'll change underneath us without warning. We'll need an actual versioned release to include this I'm afraid.

zacs: Got it. Any guidance on how best to proceed when there isn't an official release other than latest? I'm happy to create a repo on GitHub for it, but that seems a bit artificial and not in spirit with the rule.

DomT4: The only path of action that would meet our guidelines and user expectation is for upstream to tag a versioned release. A lot of upstream devs understand the packaging issue and are willing to tag a versioned release when asked, so that is probably worthwhile doing.

DomT4:
Upstream haven't touched the release for 6 years, which makes me suspect they won't be responsive to requests to create a versioned release. I'm going to close this out consequently, but if you talk upstream into tagging a versioned release please feel free to open a new PR for this, I agree it could be useful. Thanks for the submission!

PR #41161, opened 2015-06-28
go-app-engine 1.9.23

raindev:
Upgrade both 32 and 64 bit versions


DomT4:
Could you possibly update both to use `sha256` instead of `sha1`? Thanks!

raindev:
Yes, sure

xu-cheng:

raindev:
I'm glad to help! Thank you for maintaining Homebrew, it makes our lifes much easier :)

PR #41165, opened 2015-06-28
Update colordiff to 1.0.15

drolando:
The 1.0.13 tar.gz is no longer available on their website. The newest version is 1.0.15.

Since they have updated their Makefile, I've removed part of the patch that is now useless.


mistydemeo:
Thanks!

PR #41166, opened 2015-06-28
Woff2 compress and decompress

prieb:
Compile woff2_compress and woff2_decompress from [Google Woff2 Github repository](https://github.com/google/woff2)

### DIFF
--- a/Library/Formula/woff2.rb
+++ b/Library/Formula/woff2.rb
@@ -1,0 +1,21 @@
+class Woff2 < Formula
+  desc "Woff2 compress and decompress by Google"
+  homepage "https://github.com/prieb/woff2-homebrew"
+  url "https://github.com/prieb/woff2-homebrew.git"
+  version "0.2"
+
+  def install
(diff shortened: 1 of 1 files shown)


DomT4:
We don't accept private forks. The release here would actually need to come from Google's upstream repo. Google have been good about tagging releases when requested by Homebrew, so it's worthwhile asking.

prieb:
Hi @DomT4 !
I make an repository to add test.
I've to add a font to test woff2_compress. The only arg that woff2_compress accept its a font file.

mikemcquaid:
Passing on this as it's a personal fork, sorry!

deathbaba:
So no woff2 in homebrew yet.

DomT4:
Waiting on an [upstream release](https://github.com/google/woff2/issues/40). You can use the formula in `head-only` at the moment if desired:
```
brew install homebrew/head-only/woff2
```

deathbaba:
Thanks!

PR #41170, opened 2015-06-29
ctorrent: add explicit openssl dependency and test

L2G:
This also does some cleanup for `brew audit --strict`.


DomT4:
Needs a `revision` if it previously hooked the system `OpenSSL` :wink:

L2G:
:8ball:

L2G:
`:dunce_cap:`

xu-cheng:
Thanks @L2G

PR #41172, opened 2015-06-29
ccd2iso: added test; `brew audit --strict`

L2G:
I know you all prefer tests that go deeper than a basic installation qualification, but that would require obtaining an image file in CloneCD format. Since [CloneCD](http://slysoft.com/en/clonecd.html) is a commercial product, I don't really want to spend the money just to make a test image. So I hope this will suffice in this case.

### DIFF
--- a/Library/Formula/ccd2iso.rb
+++ b/Library/Formula/ccd2iso.rb
@@ -1,14 +1,18 @@
-require 'formula'
-
 class Ccd2iso < Formula
   desc "Convert CloneCD images to ISO images"
-  homepage 'http://ccd2iso.sourceforge.net/'
-  url 'https://downloads.sourceforge.net/project/ccd2iso/ccd2iso/ccd2iso-0.3/ccd2iso-0.3.tar.gz'
-  sha1 'cbefdc086adf45138d921a4a7c0ff012dd7c9ca7'
+  homepage "http://ccd2iso.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/ccd2iso/ccd2iso/ccd2iso-0.3/ccd2iso-0.3.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/ccd2iso.rb
```diff
@@ -1,14 +1,18 @@
...
   def install
     system "./configure", "--disable-debug", "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
-    system "make install"
+    system "make", "install"
+  end
+
+  test do
+    assert_match(
+      /^#{Regexp.escape(version)}$/, shell_output("#{bin}/ccd2iso --version")
```
DomT4: You shouldn't need to escape the expression here. `assert_match /#{version}/, shell_output("#{bin}/ccd2iso --version")` should play nicely.

L2G: It makes me cringe a little to be loose with it, but I suppose the odds of the version number turning into "0 3", "0!3", "0J3" or some such during the test are vanishingly small compared to the odds of a much more spectacular failure happening. (I assume that is what you're thinking.)

DomT4: To be honest it doesn't matter a huge amount either way. If we do something that changes the version string that drastically it's likely we'd either manually change every formula, or revert the breaking change. From upstream's point of view there hasn't been a new release in 9 years, so it's relatively safe to assume we won't hit anything unusual on their end.

I'll pull it though. As I said, it's not a major thing, I just prefer not to add extra code unless it's proven necessary.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/50f910080d4ee1016f6df543d673138157b27636. Thanks @L2G!

PR #41174, opened 2015-06-29
pgroonga 0.6.0 (new formula)

cosmo0920:


Library/Formula/pgroonga.rb
```diff
@@ -0,0 +1,16 @@
...
+  depends_on "cmake" => :build
+  depends_on "pkgconfig" => :build
+  depends_on "postgresql"
+  depends_on "groonga"
+
+  def install
+    system "make"
+    system "make", "install"
+  end
+end
```
mikemcquaid: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/pgroonga.rb
```diff
@@ -0,0 +1,16 @@
...
+  sha256 "29a631f68013bba19321beac0d7b933eafce60aec1597aa8fff0b0898ff2a73d"
+
+  depends_on "cmake" => :build
+  depends_on "pkgconfig" => :build
+  depends_on "postgresql"
+  depends_on "groonga"
+
+  def install
+    system "make"
+    system "make", "install"
```
mikemcquaid: Needs a `prefix` specified somehow/

Library/Formula/pgroonga.rb
cosmo0920: Hmm..., it's a bit difficult to test.
`PGroonga` is shared object, not executables. It does not have entry point.
And now, PGroonga does not have executable example, though.

Library/Formula/pgroonga.rb
cosmo0920: `PGroonga` intends to install `${prefix}/Cellar/postgresql/9.4.4{/lib/pgroonga.so, /share/postgresql/extension/pgroonga--<version>.sql}`.
PGroonga's makefile dose not provide choice installation directory.
"No choice installation directory" is an big issue to Homebrew Formula?
If so, I will discuss about it at upstream.

cosmo0920:
PGroonga is a one of PostgreSQL extension.
And this pull request work in progress, please waiting for review.

mikemcquaid:
This will need to install into its Cellar: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/29079/version=yosemite/testReport/junit/brew-test-bot/yosemite/audit_pgroonga___strict/

Library/Formula/pgroonga.rb
```diff
@@ -0,0 +1,27 @@
...
+  homepage "https://pgroonga.github.io/"
+  url "http://packages.groonga.org/source/pgroonga/pgroonga-0.6.0.tar.gz"
+  sha256 "29a631f68013bba19321beac0d7b933eafce60aec1597aa8fff0b0898ff2a73d"
+
+  depends_on "pkg-config" => :build
+  depends_on "postgresql"
+  depends_on "groonga"
+
+  def pour_bottle?
+    # Postgres extensions must live in the Postgres prefix, which precludes
```
xu-cheng: In this case, Pgroonga should be put as a optional resource of Postgres rather than a standalone formula.

mikemcquaid: :+1:. We should also do that with `postgis` too.

cosmo0920: > In this case, Pgroonga should be put as a optional resource of Postgres rather than a standalone formula

:+1: For me.

cosmo0920:
Once I close this pull request and I will send another PR which makes a optional resource of Postgres.

PR #41175, opened 2015-06-29
nghttp2 1.0.5

fwiesel:
Update it to the revision released on the 27th of June


xu-cheng:

PR #41176, opened 2015-06-29
gnatsd 0.6.0

fwiesel:
Update gnatsd to the version released 17th of June


Library/Formula/gnatsd.rb
```diff
@@ -3,9 +3,9 @@
 class Gnatsd < Formula
   desc "Lightweight cloud messaging system"
   homepage "http://nats.io"
```
vszakats: It's mixed content, but it'd still be nice to use https here:
```ruby
homepage "https://nats.io/"
```

xu-cheng:

PR #41179, opened 2015-06-29
Release drake 1.0.0 installation script

mavericklou:
No pre-download jar needed within homebrew Formula

### DIFF
--- a/Library/Formula/drake.rb
+++ b/Library/Formula/drake.rb
@@ -2,13 +2,24 @@ require 'formula'
 
 class Drake < Formula
   desc "Data workflow tool meant to be 'make for data'"
-  homepage 'https://github.com/Factual/drake'
-  url 'https://raw.githubusercontent.com/Factual/drake/v1.0.0/bin/drake'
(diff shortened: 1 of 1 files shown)


Library/Formula/drake.rb
```diff
@@ -3,13 +3,12 @@
 class Drake < Formula
   desc "Data workflow tool meant to be 'make for data'"
   homepage 'https://github.com/Factual/drake'
-  url 'https://github.com/Factual/drake/releases/download/v0.1.7/drake.jar'
-  version '0.1.7'
-  sha1 'c3092bd4e62effa38d4ade09bc191463b1c93c6a'
+  url 'https://raw.githubusercontent.com/Factual/drake/v1.0.0/bin/drake'
```
xu-cheng: Please use `drake.jar`. Homebrew disallows self-download formula.

mavericklou: How about leiningen?

mavericklou:
Will create another pull request with a clean acceptable version

PR #41180, opened 2015-06-29
Make go-app-engine formulae use SHA-256

raindev:
As requested by @DomT4 in #41161


xu-cheng:

DomT4:
@raindev Thanks for chasing up on this! Appreciate that! :cocktail:

PR #41183, opened 2015-06-29
gegl 0.3.0

dunn:


Library/Formula/gegl.rb
```diff
@@ -1,15 +1,8 @@
-require "formula"
-
 class Gegl < Formula
   desc "Graph based image processing framework"
   homepage "http://www.gegl.org/"
-
-  stable do
-    # Use Debian because the official URL is unreliable.
-    url "https://mirrors.kernel.org/debian/pool/main/g/gegl/gegl_0.2.0.orig.tar.bz2"
```
vszakats: It'd be nice to keep this URL as a (secure) `mirror`.

dunn: Oh, right. Maybe it wasn't available when I originally committed this.

vszakats: Thank you @dunn.

Library/Formula/gegl.rb
```diff
@@ -64,4 +59,23 @@ def install
...
+        return 0;
+      }
+    EOS
+    system ENV.cc, "-I#{include}/gegl-0.3", "-L#{lib}", "-lgegl-0.3",
+           "-I#{Formula["babl"].opt_include}/babl-0.1",
+           "-I#{Formula["glib"].opt_include}/glib-2.0",
+           "-I#{Formula["glib"].opt_lib}/glib-2.0/include",
+           "-L#{Formula["glib"].opt_lib}", "-lgobject-2.0", "-lglib-2.0",
+           testpath/"test.c", "-o", testpath/"test"
+    system "test", 0
```
DomT4: Does this need to be `system "./test"`?

dunn: Ohhhhhhh, that's why I needed to add the 0.  I thought it was just something about the test code I didn't understand (I need someone to write C tests for me pls).

DomT4: > I need someone to write C tests for me pls

Tell me about it :wink:.

> Ohhhhhhh, that's why I needed to add the 0

Yeah. Depending on what shell you use `test` when called directly will either be `test` the built-in shell command, or `/bin/test`. You can either prepend the `$PATH` with the testpath, or just do `./test` - We prefer the latter because it's cleaner and doesn't mess with any ENVs.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/94da16f4d37d101a49f36d5caa2b33c9f5e78078. Thanks Alex!

PR #41184, opened 2015-06-29
Fix URLs to several Google projects

kevmoo:
Might be an interesting exercise to script through all projects and test for redirects


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/787f0e07033e82c30f510fee2cd4902fd90b3782. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a tiny nit, but generally we prefer one commit per formula. Helps make changes over time easier to track.

kevmoo:
@DomT4 good to know for next time. Thanks! :metal:

PR #41187, opened 2015-06-29
plplot 5.11.0

tschoonj:
version bump
formula cleanup

### DIFF
--- a/Library/Formula/plplot.rb
+++ b/Library/Formula/plplot.rb
@@ -1,10 +1,8 @@
-require 'formula'
-
 class Plplot < Formula
   desc "Cross-platform software package for creating scientific plots"
-  homepage 'http://plplot.sourceforge.net'
-  url 'https://downloads.sourceforge.net/project/plplot/plplot/5.10.0%20Source/plplot-5.10.0.tar.gz'
-  sha1 'ea962cb0138c9b4cbf97ecab1fac1919ea0f939f'
+  homepage "http://plplot.sourceforge.net"
+  url "https://downloads.sourceforge.net/project/plplot/plplot/5.11.0%20Source/plplot-5.11.0.tar.gz"
+  sha256 "bfa8434e6e1e7139a5651203ec1256c8581e2fac3122f907f7d8d25ed3bd5f7e"
 
   bottle do
     revision 1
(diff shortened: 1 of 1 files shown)


Library/Formula/plplot.rb
```diff
@@ -13,201 +11,62 @@ class Plplot < Formula
...
-  depends_on :x11 => :optional
+  option "with-java"
 
-  option 'with-java'
+  depends_on "cmake" => :build
+  depends_on "pkg-config" => :build
+  depends_on "pango"
+  depends_on "libtool" => :run
+  depends_on "freetype"
+  depends_on :x11 => :recommended
```
DomT4: @mikemcquaid may object to making this `:recommended`. I suspect he made this `:optional`.

tschoonj:
the funny thing is that the `:x11` option itself is actually irrelevant here: the cmake script finds the X11 headers without using `pkg-config`. So even though it was optional before, the X11 device called `xwin` was always built.

The only way to disable the X11 device is through a cmake option I now included properly here. The `xcairo` device no longer works due to the cairo to quartz transition.

I would advise to make `x11` recommended, as this way it keeps closest to the previously provided devices.
Another reason is the package `gnudatalanguage` in homebrew-science, which relies on the `xwin` device. 

By the way, as soon as this update is merged in I will update `gnudatalanguage` too as it needs additional patching to restore compatibility with this version of `plplot`.

DomT4:
@BrewTestBot test this please

tschoonj:
@DomT4 Could you give this another go? The testbot had some internal error...

Thanks!

DomT4:
It's green now. Would like @mikemcquaid's yay/nay on the X11 change before merge.

tschoonj:
Hey @mikemcquaid could you have a look at this?

Thanks!

Library/Formula/plplot.rb
mikemcquaid: Yes, please make this `:optional`

tschoonj:
@mikemcquaid done!

Library/Formula/plplot.rb
```diff
@@ -13,201 +11,62 @@ class Plplot < Formula
     sha256 "39b18e81232ae3987018662035c49a441ddd0f0fcbe9d7534620d512b7f2910c" => :mountain_lion
   end
 
-  depends_on 'cmake' => :build
-  depends_on 'pkg-config' => :build
-  depends_on 'pango'
-  depends_on :x11 => :optional
+  option "with-java"
```
mikemcquaid: For a future PR: might be nice to `depends_on :java => :optional` instead

mikemcquaid:
Thanks @tschoonj!

PR #41191, opened 2015-06-29
elixir 1.0.5

thecodeboss:
Elixir updated to 1.0.5

### DIFF
--- a/Library/Formula/elixir.rb
+++ b/Library/Formula/elixir.rb
@@ -11,7 +11,7 @@ class Erlang17Requirement < Requirement
   end
 
   def message; <<-EOS.undent
-    Erlang 17 is required to install.
+    Erlang 17+ is required to install.
 
     You can install this with:
       brew install erlang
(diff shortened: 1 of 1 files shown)


brunoadacosta:
\o nice

edude03:
Does Elixir need to be updated to OTP-18 as well or can you built it on 17 still?

thecodeboss:
I can verify it still builds fine on OTP-17. My understanding is that 1.0.5 only adds support for OTP-18, and it does not deprecate OTP-17.

DevL:
Elixir 1.0.5 supports OTP 17+.
Elixir 1.1 will most likely require OTP 18.

orendon:
:+1: for this

Prolly worth to change the message `Erlang 17 is required to install` to `Erlang 17+` to avoid any confusion, see [Formula/elixir.rb#L14](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/elixir.rb#L14)

thecodeboss:
Thanks for the suggestion @orendon, I've modified the formula with that change.

reset:
:+1: looking forward to this being merged

scohen:
:+1:  Really looking forward to having 1.0.5, it fixes behavior in Record that we need.

drewolson:
I'd also love to get this merged soon(ish).

DomT4:
Just to double check: If we pull this now it doesn't break `elixir` for current users, and if we pull this now it won't break when we fire the gun on merging the updated `erlang`?

drewolson:
@DomT4 correct on both counts.

scohen:
Correct, this is compatible with r17 and r18.

thecodeboss:
@DomT4 Those two are correct, this will not break elixir for current users, and it will not break with the updated erlang.

DomT4:
Thanks all for the reply, and to @thecodeboss for the contribution to Homebrew. Merged in https://github.com/Homebrew/homebrew/commit/f670569c419ef96a32f2a5e84376184dccf655c2.

PR #41195, opened 2015-06-29
gif2png 2.5.10 (new formula)

keithws:
A command line program to convert images from GIF format into PNG format.

### DIFF
--- a/Library/Formula/gif2png.rb
+++ b/Library/Formula/gif2png.rb
@@ -18,9 +18,6 @@ class Gif2png < Formula
   end
 
   test do
-    # read test gif from stdin so gif2png will output a PNG file in the
-    # current working directory instead of alterting the PNG test fixture
-    $stdin.reopen(test_fixtures("test.gif"))
-    system "#{bin}/gif2png", "-O"
+    pipe_output "#{bin}/gif2png -O", File.read(test_fixtures("test.gif"))
   end
 end
--- a/Library/Formula/gif2png.rb
+++ b/Library/Formula/gif2png.rb
@@ -1,0 +1,22 @@
+class Gif2png < Formula
+  homepage "http://www.catb.org/~esr/gif2png/"
+  url "http://www.catb.org/~esr/gif2png/gif2png-2.5.10.tar.gz"
+  sha256 "3a593156f335c4ea6be68e37e09994461193f31872362de4b27ef6301492d5fd"
+
+  depends_on "libpng"
+
+  def install
+    ENV.deparallelize # parallel install fails
+
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    require "base64"
+    (testpath/"test.gif").write(::Base64.decode64("R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs="))
+    system "#{bin}/gif2png", "test.gif"
+  end
+end
--- a/Library/Formula/gif2png.rb
+++ b/Library/Formula/gif2png.rb
@@ -16,8 +16,9 @@ class Gif2png < Formula
   end
 
   test do
-    require "base64"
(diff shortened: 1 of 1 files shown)


Library/Formula/gif2png.rb
```diff
@@ -0,0 +1,22 @@
+class Gif2png < Formula
+  homepage "http://www.catb.org/~esr/gif2png/"
+  url "http://www.catb.org/~esr/gif2png/gif2png-2.5.10.tar.gz"
+  sha256 "3a593156f335c4ea6be68e37e09994461193f31872362de4b27ef6301492d5fd"
+
+  depends_on "libpng"
+
+  def install
+    ENV.deparallelize # parallel install fails
```
DomT4: Please report this upstream and add a link to the bug report, Thanks!

Library/Formula/gif2png.rb
```diff
@@ -0,0 +1,22 @@
+class Gif2png < Formula
+  homepage "http://www.catb.org/~esr/gif2png/"
```
DomT4: Please add a short (80 characters or less) `desc` above the `homepage` here.

Library/Formula/gif2png.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    require "base64"
+    (testpath/"test.gif").write(::Base64.decode64("R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs="))
```
DomT4: We have a `gif` testfile you can use rather than creating your own here. There's an example use [here](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/giflossy.rb#L38).

DomT4:
Ping.

keithws:
Thanks for the ping. I went on vacation and forgot about this.

Library/Formula/gif2png.rb
keithws: The upstream author does not appear to have a bug reporting tool, so I email him instead.

Library/Formula/gif2png.rb
```diff
@@ -0,0 +1,26 @@
...
+    ENV.deparallelize
+
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    # read test gif from stdin so gif2png will output a PNG file in the
```
DomT4: You can always `cp` the test fixture into the `testpath`. We empty the testpath afterwards, so it's not a HD space hit.

keithws: Why incur the expense of copying (even if small) when the program accepts files on stdin?

DomT4: We like to keep things relatively consistent, but if none of @mikemcquaid @xu-cheng or @bfontaine object I don't object enough to not merge it.

Can probably kill the notes about it though.

Library/Formula/gif2png.rb
```diff
@@ -0,0 +1,26 @@
...
+    # current working directory instead of alterting the PNG test fixture
+    $stdin.reopen(test_fixtures("test.gif"))
+    system "#{bin}/gif2png", "-O"
```
bfontaine: You can use `pipe_output("#{bin}/gif2png -O", File.read(test_fixtures("test.gif")), 0)`.

Library/Formula/gif2png.rb
mikemcquaid: I agree with @DomT4; I don't think the "expense of copying" is worth the slight confusion this introduces. Smells of premature optimisation.

DomT4: Update this per the comments and I'll merge it.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/0030f8b851337f311fd934d6698d8018fd6f17d8. Thank you for your contribution to Homebrew, we appreciate it!

It's a nit, but please squash your commits when the PR is ready to be merged, It saves us having to do so on pull, Thanks!

keithws:
Thanks!

I am still newish to git, so thanks for the tip about squashing commits. A quick search and I learned something new.

PR #41205, opened 2015-06-30
New formula: muttvcardsearch 1.9

benjaminfrank:
A small mutt carddav search utility for your Owncloud or SOGo server.

### DIFF
--- a/Library/Formula/muttvcardsearch.rb
+++ b/Library/Formula/muttvcardsearch.rb
@@ -1,0 +1,27 @@
+class Muttvcardsearch < Formula
+  desc "A small mutt carddav search utility for Owncloud or SOGo server"
+  homepage "https://github.com/BlackIkeEagle/muttvcardsearch"
+  url "https://github.com/BlackIkeEagle/muttvcardsearch/archive/v1.9.tar.gz"
+  sha256 "4c099938dd02f577d8289bbc5e82f0af6f290a564a30f6fdcb5cc880cd02fc8d"
+
+  depends_on "cmake" => :build
+
+  # Patch required to enable CMAKE_INSTALL_PREFIX support for custom
+  # installation paths.
+  # Upstream pull request: https://github.com/BlackIkeEagle/muttvcardsearch/pull/10
+  patch do
+    url "https://patch-diff.githubusercontent.com/raw/BlackIkeEagle/muttvcardsearch/pull/10.patch"
+    sha256 "683528161e1b4f19ba75c5e367917969f2d5f995fc753659d8c59079a314d3cc"
+  end
+
+  def install
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args
+      system "make", "install"
+    end
+  end
+
+  test do
+    assert_match /invalid or missing arguments/, shell_output("#{bin}/muttvcardsearch", 1)
+  end
+end


Library/Formula/muttvcardsearch.rb
```diff
@@ -0,0 +1,27 @@
+class Muttvcardsearch < Formula
+  desc "A small mutt carddav search utility for Owncloud or SOGo server"
+  homepage "https://github.com/BlackIkeEagle/muttvcardsearch"
+  url "https://github.com/BlackIkeEagle/muttvcardsearch/archive/v1.9.tar.gz"
+  sha256 "4c099938dd02f577d8289bbc5e82f0af6f290a564a30f6fdcb5cc880cd02fc8d"
+
+  option "with-debug", "build debug version"
+  depends_on "cmake" => :build
```
DomT4: Spacing between options and deps please.

Library/Formula/muttvcardsearch.rb
```diff
@@ -0,0 +1,27 @@
...
+  depends_on "sqlite"
```
DomT4: Is the system version not recent enough?

Library/Formula/muttvcardsearch.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def install
+    mkdir "build" do
+      args = std_cmake_parameters.split
+      if build.include? "enable-debug"
```
DomT4: `if build.with?`

Library/Formula/muttvcardsearch.rb
```diff
@@ -0,0 +1,27 @@
...
+        args.delete "-DCMAKE_BUILD_TYPE=None"
```
DomT4: We don't pass this argument any more in `std_cmake_args`.

DomT4: We use `RELEASE` now.

DomT4:
@BrewTestBot test this please

Library/Formula/muttvcardsearch.rb
```diff
@@ -0,0 +1,29 @@
+class Muttvcardsearch < Formula
+  desc "A small mutt carddav search utility for Owncloud or SOGo server"
+  homepage "https://github.com/BlackIkeEagle/muttvcardsearch"
+  url "https://github.com/BlackIkeEagle/muttvcardsearch/archive/v1.9.tar.gz"
+  sha256 "4c099938dd02f577d8289bbc5e82f0af6f290a564a30f6fdcb5cc880cd02fc8d"
+
+  option "with-debug", "build debug version"
```
mikemcquaid: What's the advantage of this? We don't typical add debug options.

Library/Formula/muttvcardsearch.rb
```diff
@@ -0,0 +1,29 @@
...
+  def install
+    mkdir "build" do
+      args = std_cmake_parameters.split
+      if build.with? "debug"
+        args.delete "-DCMAKE_BUILD_TYPE=Release"
+        args << "-DCMAKE_BUILD_TYPE=Debug"
+      end
+      system "cmake", "..", *args
+      system "make"
+      bin.install "muttvcardsearch"
```
mikemcquaid: Is there no `make install` we can use?

Library/Formula/muttvcardsearch.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+  end
+
+  test do
+    # muttvcardsearch requires a valid carddav endpoint to
+    # successfully run any operation (rc == 0)
+    true
```
mikemcquaid: Even e.g. `--version`? Really would like to have something here.

Library/Formula/muttvcardsearch.rb
benjaminfrank: Not that I am aware of, I just carried it along, to be totally honest. I'll remove it.

Library/Formula/muttvcardsearch.rb
benjaminfrank: the non-zero rc was the issue here, but today I learned that there is eg assert_match to check the output of a command, so the rc is not really releveant. Sorry, I am rather new to this whole formula thing. I'll change it to something meaningfull later.

DomT4: We can also "expect" the exit code if necessary. [Example](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/sonar.rb#L56-L58).

Library/Formula/muttvcardsearch.rb
benjaminfrank: I am no cmake expert, but using 'make install' aftwards gives the following error:
==> make install
  "/tmp/muttvcardsearch20150702-4863-so96fg/muttvcardsearch-1.9/build/muttvcardsearch"
  to "/usr/bin/muttvcardsearch".
make: *** [install] Error 1

Library/Formula/muttvcardsearch.rb
benjaminfrank: Nice, thanks.

Library/Formula/muttvcardsearch.rb
DomT4: Rather than `std_cmake_parameters.split` you just want `std_cmake_args`, I believe.

DomT4:
```
        ==> brew style muttvcardsearch
== /usr/local/Library/Formula/muttvcardsearch.rb ==
C: 19:  5: Incorrect indentation detected (column 4 instead of 5).
C: 20:  3: Use 2 (not 3) spaces for indentation.

1 file inspected, 2 offenses detected
```

benjaminfrank:
Thanks, I couldn't view the Details yesterday.

DomT4:
Yeah, we had some CI issues that persisted throughout yesterday. Sorry about the delay. You can hit the same audit input as the bot locally by doing `brew audit --strict` on the formula.

benjaminfrank:
I thought I did, that's why I was puzzled there is an error...(aka lazy me)

Library/Formula/muttvcardsearch.rb
mikemcquaid: We should fix up that `make install` instead and submit a patch upstream.

benjaminfrank: Fixed & reported upstream (https://github.com/BlackIkeEagle/muttvcardsearch/pull/10).
In the meantime the formula patches the cmake file .

mikemcquaid: Please add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Also, use `https://github.com/BlackIkeEagle/muttvcardsearch/pull/10.patch` as the patch URL. Thanks!

DomT4:
@BrewTestBot test this please

Library/Formula/muttvcardsearch.rb
```diff
@@ -0,0 +1,28 @@
...
+  patch do
+    url "https://patch-diff.githubusercontent.com/raw/BlackIkeEagle/muttvcardsearch/pull/10.patch"
+    sha256 "683528161e1b4f19ba75c5e367917969f2d5f995fc753659d8c59079a314d3cc"
+  end
+
+  def install
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args
+      system "make", "install"
+      bin.install "muttvcardsearch"
```
DomT4: Is the CMakeList not installing this, even with your patch?

benjaminfrank: Sorry, my bad. Fixed it.

benjaminfrank:
Upstream seems pretty much inactive to incorporate the cmake fix. But then I am no longer using this software anyway.

DomT4:
Alright, let's pass on this if upstream haven't responded to the patch PR and the software is no longer being used by you. Thanks for the submission though! :heart:

PR #41207, opened 2015-06-30
New formula: extract_url 1.5.8

benjaminfrank:
This is a Perl script that extracts URLs from correctly-encoded MIME
email messages or plain text.

### DIFF
--- a/Library/Formula/extract_url.rb
+++ b/Library/Formula/extract_url.rb
@@ -57,7 +57,7 @@ class ExtractUrl < Formula
     end
 
     resource("URI::Find").stage do
-      system "perl", "Build.PL", "--install_base", "#{libexec}"
+      system "perl", "Build.PL", "--install_base", libexec
       system "./Build"
       system "./Build", "install"
     end
--- a/Library/Formula/extract_url.rb
+++ b/Library/Formula/extract_url.rb
@@ -4,20 +4,74 @@ class ExtractUrl < Formula
   url "https://extracturl.googlecode.com/files/extract_url-1.5.8.tar.gz"
   sha256 "58eac907cb926deba74ab81e7503a1055fd3cbe20952f011d8e6b75da12d6bcc"
 
-  depends_on "MIME::Parser" => :perl
-  depends_on "HTML::Parser" => :perl
-  depends_on "Pod::Usage" => :perl
-  depends_on "Env" => :perl
-  depends_on "Getopt::Long" => :perl
+  resource "MIME::Parser" do
+    url "https://cpan.metacpan.org/authors/id/D/DS/DSKOLL/MIME-tools-5.506.tar.gz"
+    sha256 "dbed9bf46830c4a1df9840a546824ee44d14902012870f0c34bc4f5cc86af812"
+  end
+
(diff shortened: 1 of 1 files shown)


Library/Formula/extract_url.rb
```diff
@@ -0,0 +1,23 @@
+class ExtractUrl < Formula
+  desc "Perl script to extracts URLs from emails or plain text."
+  homepage "http://www.memoryhole.net/~kyle/extract_url/"
+  url "https://extracturl.googlecode.com/files/extract_url-1.5.8.tar.gz"
+  sha256 "58eac907cb926deba74ab81e7503a1055fd3cbe20952f011d8e6b75da12d6bcc"
+
+  depends_on "MIME::Parser" => :perl
```
DomT4: We'd prefer to vendor these rather than ordering the user to install them before they can install the formula. The latter tends to be bad UX.

Library/Formula/extract_url.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def install
+    system "make", "man"
+    mv "extract_url.pl", "extract_url"
+    bin.install "extract_url"
+    man1.install "extract_url.1"
+  end
+
+  test do
+    system "#{bin}/extract_url", "--help"
```
DomT4: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

DomT4:
@BrewTestBot test this please

Library/Formula/extract_url.rb
```diff
@@ -0,0 +1,77 @@
...
+    %w[MIME::Parser HTML::Parser Pod::Usage Env Getopt::Long Curses Curses::UI].each do |r|
+      resource(r).stage do
+        system "perl", "Makefile.PL", "INSTALL_BASE=#{libexec}"
+        system "make"
+        system "make", "install"
+      end
+    end
+
+    resource("URI::Find").stage do
+      system "perl", "Build.PL", "--install_base", "#{libexec}"
```
benjaminfrank: This line creates an audit problem: 
* Don't need to interpolate "libexec" with system
However, this it seems it is needed as 'system "perl", "Build.PL", "--install_base #{libexec}"' does not work and omitting it complete also breaks the build.
Any advice welcome.

DomT4: It just means that you don't need to use the `#{}` mechanism in this case.

```ruby
system "perl", "Build.PL", "--install_base", libexec
```

Is fine here.

benjaminfrank: Nice, thanks.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/3f5f09f6fb9039e62f17aa57b33e3ff9ce1bf513. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Again, please squash in future before final review, and commit messages for new formulae should be styled `extract_url 1.5.8 (new formula)` or similar, Thanks!

PR #41208, opened 2015-06-30
New formula: vdirsyncer 0.6.0

benjaminfrank:
Synchronize calendars and contacts https://vdirsyncer.readthedocs.org/

### DIFF
--- a/Library/Formula/vdirsyncer.rb
+++ b/Library/Formula/vdirsyncer.rb
@@ -1,0 +1,92 @@
+class Vdirsyncer < Formula
+  desc "Synchronize calendars and contacts"
+  homepage "https://github.com/untitaker/vdirsyncer"
+  url "https://pypi.python.org/packages/source/v/vdirsyncer/vdirsyncer-0.6.0.tar.gz"
+  sha256 "0d2a9677b086cfbe5fe5e7cb5e55db4c1afab62cb5dc56aeaff6e95d34bf60d5"
+  head "https://github.com/untitaker/vdirsyncer.git"
+
+  option "without-keyring", "Build without python-keyring support"
+
+  depends_on :python3
+
+  resource "keyring" do
+    url "https://pypi.python.org/packages/source/k/keyring/keyring-5.4.tar.gz"
+    sha256 "45891cd0af4c4af70fbed7ec6e3964d0261c14188de9ab31030c9d02272e22d2"
+  end
+
+  resource "click" do
+    url "https://pypi.python.org/packages/source/c/click/click-5.1.tar.gz"
+    sha256 "678c98275431fad324275dec63791e4a17558b40e5a110e20a82866139a85a5a"
+  end
+
+  resource "requests" do
+    url "https://pypi.python.org/packages/source/r/requests/requests-2.7.0.tar.gz"
+    sha256 "398a3db6d61899d25fd4a06c6ca12051b0ce171d705decd7ed5511517b4bb93d"
+  end
+
+  resource "lxml" do
+    url "https://pypi.python.org/packages/source/l/lxml/lxml-3.4.4.tar.gz"
+    sha256 "b3d362bac471172747cda3513238f115cbd6c5f8b8e6319bf6a97a7892724099"
+  end
+
+  resource "requests-toolbelt" do
+    url "https://pypi.python.org/packages/source/r/requests-toolbelt/requests-toolbelt-0.4.0.tar.gz"
+    sha256 "15b74b90a63841b8430d6301e5062cd92929b1074b0c95bf62166b8239db1a96"
+  end
+
+  resource "atomicwrites" do
+    url "https://pypi.python.org/packages/source/a/atomicwrites/atomicwrites-0.1.5.tar.gz"
+    sha256 "9b16a8f1d366fb550f3d5a5ed4587022735f139a4187735466f34cf4577e4eaa"
+  end
+
+  def install
+    version = Language::Python.major_minor_version "python3"
+    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python#{version}/site-packages"
+    rs = %w[click requests lxml requests-toolbelt atomicwrites]
+    rs << "keyring" if build.with? "keyring"
+    rs.each do |r|
+      resource(r).stage do
+        system "python3", *Language::Python.setup_install_args(libexec/"vendor")
+      end
+    end
+
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python#{version}/site-packages"
+    system "python3", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir["#{libexec}/bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    ENV["LC_ALL"] = "en_US.UTF-8"
+    ENV["LANG"] = "en_US.UTF-8"
+    (testpath/".config/vdirsyncer/config").write <<-EOS.undent
+      [general]
+      status_path = #{testpath}/.vdirsyncer/status/
+      [pair contacts]
+      a = contacts_a
+      b = contacts_b
+      collections = ["from a"]
+      [storage contacts_a]
+      type = filesystem
+      path = ~/.contacts/a/
+      fileext = .vcf
+      [storage contacts_b]
+      type = filesystem
+      path = ~/.contacts/b/
+      fileext = .vcf
+    EOS
+    (testpath/".contacts/a/foo/092a1e3b55.vcf").write <<-EOS.undent
+      BEGIN:VCARD
+      VERSION:3.0
+      EMAIL;TYPE=work:username@example.org
+      FN:User Name
+      UID:092a1e3b55
+      N:Name;User
+      END:VCARD
+    EOS
+    (testpath/".contacts/b/foo/").mkpath
+    system "#{bin}/vdirsyncer", "sync"
+    assert_match /BEGIN:VCARD/, (testpath/".contacts/b/foo/092a1e3b55.vcf").read
+  end
+end


Library/Formula/vdirsyncer.rb
```diff
@@ -0,0 +1,62 @@
+class Vdirsyncer < Formula
+  desc "Synchronize calendars and contacts"
+  homepage "https://github.com/untitaker/vdirsyncer"
+  url "https://github.com/untitaker/vdirsyncer/archive/0.5.2.tar.gz"
+  sha256 "0992d4c0c98e71a0638050f3900125a82bd29e15d5be080a4520830fa48c660a"
+
+  option "without-keyring", "Build without python-keyring support"
+  depends_on :python3
```
DomT4: Blank line between the option and dependency, please.

Library/Formula/vdirsyncer.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  if build.with? "keyring"
```
DomT4: Don't wrap resources in the main block. They should be wrapped inside `def install` instead.

DomT4: i.e. 

```ruby
r = %w[click requests lxml requests-toolbelt atomicwrites]
r << "keyring" if build.with? "keyring"
```

Library/Formula/vdirsyncer.rb
```diff
@@ -0,0 +1,62 @@
...
+    system "python3", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir["#{libexec}/bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    ENV["LC_ALL"] = "en_US.UTF-8"
+    ENV["LANG"] = "en_US.UTF-8"
+    system "#{bin}/vdirsyncer", "--version"
```
DomT4: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

benjaminfrank: Not without a working {card,cal}dav endpoint to test against. I figured introducing a remote service as test dependency might not be the best idea here. But I am open for ideas.

DomT4: Cool, no worries.

DomT4:
@BrewTestBot test this please

tdsmith:
If this is only interesting to you as a dependency of that other package, then this PR can be closed.

Does this really need python3? We use system Python (meaning python 2) when possible. All instances of "python3.4" need to use the result of `Language::Python.major_minor_version "python3"` instead, like:

```ruby
xy = Language::Python.major_minor_version "python3"
... libexec/"lib/python#{xy}/site-packages" ...
```

benjaminfrank:
As already mentioned, it is first and foremost a standalone application to sync {card,cal}dav endpoints with your local filesystem. aside from that, it offers some library like API for applications to call.
As for python3: python2 does not work as there is a dependency (forgot which one) requires python3. I will include the mentioned changes,

benjaminfrank:
Should I use Language::Python.major_minor_version "python3" for python2 as well? as in Language::Python.major_minor_version "python2" ?

Library/Formula/vdirsyncer.rb
untitaker: `make install-test` will install all required test dependencies into the current virtualenv. The CardDAV/CalDAV endpoint is just another Python package.

IOW, you need to do this:

    virtualenv env
    . env/bin/activate
    make install-test
    make test

untitaker:
>python2 does not work as there is a dependency (forgot which one) requires python3

That's not true, vdirsyncer works against Python 2.7 and Python 3.3+ (Python 3.2 and 3.1 are not supported)

Library/Formula/vdirsyncer.rb
tdsmith: Our `test` block is run after the source code has been cleaned up; it's run against the installed packages, mostly to make sure that it launches and finds its dependencies okay.

untitaker: I see. Are there any requirements on running the testsuite without internet access? I'll look into adapting the Makefile to suit this usecase.

mikemcquaid: It's fine to use internet access. Ideally instead of running the test suite it'll just run an actual typical usage of the tool.

benjaminfrank:
Yes, that comment from me is rather outdated. I thought that click requires python3 (the error said something about pyhton3 and lazy me just tried pathon3 and everything was fine, so I assumed python3) but a closer look indicated another issue regarding proper LANG env settings.

Long story short, the formula does not require python2 and works with the os x system python.

untitaker:
Ok, I forgot to mention that the user experience is better with vdirsyncer running on Python 3 though. Whether that's worth the extra dependency is up to you.

On 11 July 2015 14:00:26 CEST, benjaminfrank <notifications@github.com> wrote:
>Yes, that comment from me is rather outdated. I thought that click
>requires python3 (the error said something about pyhton3 and lazy me
>just tried pathon3 and everything was fine, so I assumed python3) but a
>closer look indicated another issue regarding proper LANG env settings.
>
>Long story short, the formula does not require python2 and works with
>the os x system python.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/Homebrew/homebrew/pull/41208#issuecomment-120607849

-- 
Sent from my phone. Please excuse my brevity.

mikemcquaid:
> Ok, I forgot to mention that the user experience is better with vdirsyncer running on Python 3 though.

@untitaker Mind elaborating a bit so we can decide? Thanks!

untitaker:
Under Python 2, vdirsyncer doesn't respond to ^C properly, or only after a long
time. I personally don't think it's worth to introduce such a huge dependency.

https://github.com/untitaker/vdirsyncer/issues/219

mikemcquaid:
@untitaker Cool, thanks, Python 2 it is.
@benjaminfrank Still waiting for a better test here. No rush but just wanted you to know, thanks!

benjaminfrank:
I will try to come up with something meaningfull, but in the progress I found another issue with tls/ssl I need to figure out first.

cweagans:
Looking forward to this. I don't think it's reasonable to hold this up on a more meaningful test. The `offlineimap` formula just runs `offlineimap --version` because testing the actual application functionality is out of scope for the homebrew formula. I think the same should apply here. Can we move forward?

cweagans:
That said, this formula doesn't successfully install for me. I'm on Yosemite.

Ran `https://raw.githubusercontent.com/benjaminfrank/homebrew/vdirsyncer/Library/Formula/vdirsyncer.rb` and got this:

```
$ brew install https://raw.githubusercontent.com/benjaminfrank/homebrew/vdirsyncer/Library/Formula/vdirsyncer.rb
==> Downloading https://github.com/untitaker/vdirsyncer/archive/0.5.2.tar.gz
Already downloaded: /Library/Caches/Homebrew/vdirsyncer-0.5.2.tar.gz
==> Downloading https://pypi.python.org/packages/source/c/click/click-4.0.tar.gz
Already downloaded: /Library/Caches/Homebrew/vdirsyncer--click-4.0.tar.gz
==> python -c import setuptools... --no-user-cfg install --prefix=/usr/local/Cellar/vdirsyncer/0.5.2/libexec/vendor --single-version-externally-managed --record=installed.txt
==> Downloading https://pypi.python.org/packages/source/r/requests/requests-2.7.0.tar.gz
Already downloaded: /Library/Caches/Homebrew/vdirsyncer--requests-2.7.0.tar.gz
==> python -c import setuptools... --no-user-cfg install --prefix=/usr/local/Cellar/vdirsyncer/0.5.2/libexec/vendor --single-version-externally-managed --record=installed.txt
==> Downloading https://pypi.python.org/packages/source/l/lxml/lxml-3.4.4.tar.gz
Already downloaded: /Library/Caches/Homebrew/vdirsyncer--lxml-3.4.4.tar.gz
==> python -c import setuptools... --no-user-cfg install --prefix=/usr/local/Cellar/vdirsyncer/0.5.2/libexec/vendor --single-version-externally-managed --record=installed.txt
==> Downloading https://pypi.python.org/packages/source/r/requests-toolbelt/requests-toolbelt-0.4.0.tar.gz
Already downloaded: /Library/Caches/Homebrew/vdirsyncer--requests-toolbelt-0.4.0.tar.gz
==> python -c import setuptools... --no-user-cfg install --prefix=/usr/local/Cellar/vdirsyncer/0.5.2/libexec/vendor --single-version-externally-managed --record=installed.txt
==> Downloading https://pypi.python.org/packages/source/a/atomicwrites/atomicwrites-0.1.5.tar.gz
Already downloaded: /Library/Caches/Homebrew/vdirsyncer--atomicwrites-0.1.5.tar.gz
==> python -c import setuptools... --no-user-cfg install --prefix=/usr/local/Cellar/vdirsyncer/0.5.2/libexec/vendor --single-version-externally-managed --record=installed.txt
==> Downloading https://pypi.python.org/packages/source/k/keyring/keyring-5.3.zip
Already downloaded: /Library/Caches/Homebrew/vdirsyncer--keyring-5.3.zip
==> python -c import setuptools... --no-user-cfg install --prefix=/usr/local/Cellar/vdirsyncer/0.5.2/libexec/vendor --single-version-externally-managed --record=installed.txt
==> python -c import setuptools... --no-user-cfg install --prefix=/usr/local/Cellar/vdirsyncer/0.5.2/libexec --single-version-externally-managed --record=installed.txt
    ep.load()(self, ep.name, value)
  File "build/bdist.macosx-10.10-intel/egg/setuptools_scm/integration.py", line 17, in version_keyword
  File "build/bdist.macosx-10.10-intel/egg/setuptools_scm/__init__.py", line 56, in get_version
  File "build/bdist.macosx-10.10-intel/egg/setuptools_scm/__init__.py", line 27, in version_from_scm
LookupError: no scm found for '/private/tmp/vdirsyncer20150724-14329-1yei5so/vdirsyncer-0.5.2'

READ THIS: https://git.io/brew-troubleshooting

These open issues may also help:
New formula: vdirsyncer 0.5.2 (https://github.com/Homebrew/homebrew/pull/41208)
```

benjaminfrank:
Yes, this error is known. The issue is that the github releases do not contain neither the version.py file nor the .git infos that would be require for setuptools_scm to generate ist. 
The solution is to simply use the pypi releases. However, I was too busy to change that so far as there is another thing that I wasn't able to solve yet. vdirsyncher build fine for me but can't connect to https sites. I am not sure what the real issue is here. When I use pip, vdirsyncer installs fine and can connect to https. I did all the same steps use all the same package versions, but something seems differen. But I am real tight on time this week, so sorry for this.
I'll quikly change the release so vdirsyncer builds at least.

untitaker:
Your latest commit is called 0.2.5 when it should be 0.5.2 FWIW.

On 25 July 2015 08:29:22 CEST, benjaminfrank <notifications@github.com> wrote:
>Yes, this error is known. The issue is that the github releases do not
>contain neither the version.py file nor the .git infos that would be
>require for setuptools_scm to generate ist. 
>The solution is to simply use the pypi releases. However, I was too
>busy to change that so far as there is another thing that I wasn't able
>to solve yet. vdirsyncher build fine for me but can't connect to https
>sites. I am not sure what the real issue is here. When I use pip,
>vdirsyncer installs fine and can connect to https. I did all the same
>steps use all the same package versions, but something seems differen.
>But I am real tight on time this week, so sorry for this.
>I'll quikly change the release so vdirsyncer builds at least.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/Homebrew/homebrew/pull/41208#issuecomment-124810433

-- 
Sent from my phone. Please excuse my brevity.

benjaminfrank:
> Your latest commit is called 0.2.5 when it should be 0.5.2 FWIW.

I am sorry, but I don't get it. I can't find anything containing 0.2.5....could you be more specific so I can fix it? Thanks.

untitaker:
Eugh, seems like I misread myself, sorry for the noise.

On 25 July 2015 09:50:16 CEST, benjaminfrank <notifications@github.com> wrote:
>> Your latest commit is called 0.2.5 when it should be 0.5.2 FWIW.
>
>I am sorry, but I don't get it. I can't find anything containing
>0.2.5....could you be more specific so I can fix it? Thanks.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/Homebrew/homebrew/pull/41208#issuecomment-124817744

-- 
Sent from my phone. Please excuse my brevity.

benjaminfrank:
No problem.

Library/Formula/vdirsyncer.rb
untitaker: FWIW I'd keep the current tests like this, and start proper testing with the next release. I've changed the build scripts in current master to be more usable for your purposes, with instructions in the comments of the `Makefile`.

>Ideally instead of running the test suite it'll just run an actual typical usage of the tool.

That requires setting up a lot of state, I don't think it's viable. The testsuite ultimatively does this too.

mikemcquaid: The testsuite sets up the state? In that case we could set it up there and install it somewhere the test can use. I'd rather we had a better test before this is included.

untitaker: Right, but the problem is that the test scripts for 0.5.2 don't really make it easy to test against any other installation than one from the repo.

I think that will be easier in 0.6.0, but until then I would keep the formula like this. Meanwhile I'd like to get feedback on the current Makefile in master.

OT, How fast are the tests expected to run?

On 1 August 2015 15:35:22 CEST, Mike McQuaid <notifications@github.com> wrote:
>> +      resource(r).stage do
>> +        system "python3",
>*Language::Python.setup_install_args(libexec/"vendor")
>> +      end
>> +    end
>> +
>> +    ENV.prepend_create_path "PYTHONPATH",
>libexec/"lib/python3.4/site-packages"
>> +    system "python3", *Language::Python.setup_install_args(libexec)
>> +
>> +    bin.install Dir["#{libexec}/bin/*"]
>> +    bin.env_script_all_files(libexec/"bin", :PYTHONPATH =>
>ENV["PYTHONPATH"])
>> +  end
>> +
>> +  test do
>> +    ENV["LC_ALL"] = "en_US.UTF-8"
>> +    ENV["LANG"] = "en_US.UTF-8"
>> +    system "#{bin}/vdirsyncer", "--version"
>
>The testsuite sets up the state? In that case we could set it up there
>and install it somewhere the test can use. I'd rather we had a better
>test before this is included.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/Homebrew/homebrew/pull/41208/files#r36033634

-- 
Sent from my phone. Please excuse my brevity.

mikemcquaid: Again, we'd prefer actual usage rather than test scripts e.g. synchronise two folders whose data you populate in the `test do` block. Tests need to run in less than 30s.

untitaker: Your decision, and yes, syncing two folders (as opposed to a folder and server) is entirely doable within 30 seconds.

On 3 August 2015 14:27:43 CEST, Mike McQuaid <notifications@github.com> wrote:
>> +      resource(r).stage do
>> +        system "python3",
>*Language::Python.setup_install_args(libexec/"vendor")
>> +      end
>> +    end
>> +
>> +    ENV.prepend_create_path "PYTHONPATH",
>libexec/"lib/python3.4/site-packages"
>> +    system "python3", *Language::Python.setup_install_args(libexec)
>> +
>> +    bin.install Dir["#{libexec}/bin/*"]
>> +    bin.env_script_all_files(libexec/"bin", :PYTHONPATH =>
>ENV["PYTHONPATH"])
>> +  end
>> +
>> +  test do
>> +    ENV["LC_ALL"] = "en_US.UTF-8"
>> +    ENV["LANG"] = "en_US.UTF-8"
>> +    system "#{bin}/vdirsyncer", "--version"
>
>Again, we'd prefer actual usage rather than test scripts e.g.
>synchronise two folders whose data you populate in the `test do` block.
>Tests need to run in less than 30s.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/Homebrew/homebrew/pull/41208/files#r36079955

-- 
Sent from my phone. Please excuse my brevity.

mikemcquaid: Sounds great.

untitaker:
0.6.0 is out.

benjaminfrank:
Sorry for taking my time, but I wasn't around the last month.
Updated to 0.6.0 and added a proper test.

Library/Formula/vdirsyncer.rb
```diff
@@ -0,0 +1,91 @@
+class Vdirsyncer < Formula
+  desc "Synchronize calendars and contacts"
+  homepage "https://github.com/untitaker/vdirsyncer"
+  url "https://pypi.python.org/packages/source/v/vdirsyncer/vdirsyncer-0.6.0.tar.gz"
+  sha256 "0d2a9677b086cfbe5fe5e7cb5e55db4c1afab62cb5dc56aeaff6e95d34bf60d5"
+  head "https://github.com/untitaker/vdirsyncer.git"
+
+  option "without-keyring", "Build without python-keyring support"
+
+  depends_on :python
```
DomT4: This should be `depends_on :python if MacOS.version <= :snow_leopard` AFAIK. Otherwise we end up making everyone who installs this from a bottle install our Python first.

benjaminfrank: That is intentional as the system python has (some) broken ssl code such that you can't use https connections with vdirsyncer. at least it didn't work for me

untitaker: In that case you might as well depend on Python 3 since vdirsyncer works better with it, no?

On 5 September 2015 08:53:47 CEST, benjaminfrank <notifications@github.com> wrote:
>> @@ -0,0 +1,91 @@
>> +class Vdirsyncer < Formula
>> +  desc "Synchronize calendars and contacts"
>> +  homepage "https://github.com/untitaker/vdirsyncer"
>> +  url
>"https://pypi.python.org/packages/source/v/vdirsyncer/vdirsyncer-0.6.0.tar.gz"
>> +  sha256
>"0d2a9677b086cfbe5fe5e7cb5e55db4c1afab62cb5dc56aeaff6e95d34bf60d5"
>> +  head "https://github.com/untitaker/vdirsyncer.git"
>> +
>> +  option "without-keyring", "Build without python-keyring support"
>> +
>> +  depends_on :python
>
>That is intentional as the system python has (some) broken ssl code
>such that you can't use https connections with vdirsyncer. at least it
>didn't work for me
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/Homebrew/homebrew/pull/41208/files#r38808419

-- 
Sent from my phone. Please excuse my brevity.

benjaminfrank: I think that should be possible, not sure about other pros/cons. If no one objects I'll change that later.

DomT4: That should have been fixed for 10.10.5. Other platforms may still have issues I guess. I seem to recall @tdsmith saying the warning wasn't entirely correct/accurate, but people will get upset if they see it. I'm happy with Python3 unless Tim isn't.

mikemcquaid: Also happy with Python3 unless @tdsmith disagrees.

benjaminfrank:
Also updated the formula to use python3.

benjaminfrank:
Anything left I can do to include this into homebrew?

Library/Formula/vdirsyncer.rb
```diff
@@ -0,0 +1,91 @@
...
+    sha256 "15b74b90a63841b8430d6301e5062cd92929b1074b0c95bf62166b8239db1a96"
+  end
+
+  resource "atomicwrites" do
+    url "https://pypi.python.org/packages/source/a/atomicwrites/atomicwrites-0.1.5.tar.gz"
+    sha256 "9b16a8f1d366fb550f3d5a5ed4587022735f139a4187735466f34cf4577e4eaa"
+  end
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python3.4/site-packages"
```
DomT4: This needs changing now. Python3 is up to 3.5. Tim has recommended:

```ruby
version = Language::Python.major_minor_version "python3"
ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python#{version}/site-packages"
```
elsewhere to handle that, so that should work here I expect. You'd want:
```ruby
ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python#{version}/site-packages"
```
for the primary installation as well.

benjaminfrank:
Thanks, changed that bit and will keep it in mind for other python programs.

DomT4:
I'm happy enough here. Merged in https://github.com/Homebrew/homebrew/commit/04438fc080ed6cfb83dc6763db1114eb6ee30f0f. Thank you for sticking with this @benjaminfrank!

untitaker:
Hooray!

@benjaminfrank I see that vdirsyncer's tests are not run, but were you able to run the testsuite successfully on OS X?

benjaminfrank:
I admit I never even tried to run them as vdirsyncer just worked. I will give the tests a shot tonight.

PR #41209, opened 2015-06-30
New formula: khard v0.4.0

benjaminfrank:
Console carddav client.

### DIFF
--- a/Library/Formula/khard.rb
+++ b/Library/Formula/khard.rb
@@ -1,0 +1,66 @@
+class Khard < Formula
+  desc "Console carddav client."
+  homepage "https://github.com/scheibler/khard"
+  url "https://github.com/scheibler/khard/archive/v0.4.0.tar.gz"
+  sha256 "4140fb06a1da63a5eb26edc7d3e3c01b99c464276a9e590b57ee2451c4416f77"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  resource "vobject" do
+    url "https://pypi.python.org/packages/source/v/vobject/vobject-0.8.1c.tar.gz"
+    sha256 "594113117f2017ed837c8f3ce727616f9053baa5a5463a7420c8249b8fc556f5"
+  end
+
+  resource "configobj" do
+    url "https://pypi.python.org/packages/source/c/configobj/configobj-5.0.6.tar.gz"
+    sha256 "a2f5650770e1c87fb335af19a9b7eb73fc05ccf22144eb68db7d00cd2bcb0902"
+  end
+
+  resource "argparse" do
+    url "https://pypi.python.org/packages/source/a/argparse/argparse-1.3.0.tar.gz"
+    sha256 "b3a79a23d37b5a02faa550b92cbbbebeb4aa1d77e649c3eb39c19abf5262da04"
+  end
+
+  resource "six" do
+    url "https://pypi.python.org/packages/source/s/six/six-1.9.0.tar.gz"
+    sha256 "e24052411fc4fbd1f672635537c3fc2330d9481b18c0317695b46259512c91d5"
+  end
+
+  def install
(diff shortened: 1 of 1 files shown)


Library/Formula/khard.rb
```diff
@@ -0,0 +1,44 @@
...
+  url "https://github.com/scheibler/khard/archive/v0.4.0.tar.gz"
+  sha256 "4140fb06a1da63a5eb26edc7d3e3c01b99c464276a9e590b57ee2451c4416f77"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  resource "vobject" do
+    url "https://pypi.python.org/packages/source/v/vobject/vobject-0.8.1c.tar.gz"
+    sha256 "594113117f2017ed837c8f3ce727616f9053baa5a5463a7420c8249b8fc556f5"
+  end
+  resource "configobj" do
```
DomT4: Please add spacing between the resources. It's hard to read quickly bunched together.

Library/Formula/khard.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir["#{libexec}/bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    system "#{bin}/khard", "--version"
```
DomT4: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

benjaminfrank: This would require to write a config file to ~/.config/khard/khard.conf (and adding a valid .vcf file to some specific directory). As I am new to all the homebrew internal magic, how would I approach that config thing?

DomT4: You won't be able to write to the user's actual home directory, but the `testpath` "captures" the `HOME` environmental variable for the duration of the test, so you can just treat the `testpath` like a `HOME` directory and write config files, etc to it. Example [here](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/crystal.rb#L25-L37).

DomT4:
@BrewTestBot test this please

benjaminfrank:
I can't make much of the error from brewbot:

    <0> expected but was <2>.

Testing locally gives me no error:

    $ brew test --verbose khard.rb
    Testing khard
    ==> /usr/local/Cellar/khard/0.4.0/bin/khard -s user
    $

DomT4:
Heh, exit 2 is usually `bash` complaining about misuse of the shell. We can expect it by using:

```ruby
assert_match /Address book: default/, shell_output("#{bin}/khard -s user", 2)
```

benjaminfrank:
Ok, that's easy to fix :) But why didn't I get this error when testing locally?

benjaminfrank:
Bummer, now it is reversed...

    Testing khard
    ==> /usr/local/Cellar/khard/0.4.0/bin/khard -s user
    Error: khard: failed
    <2> expected but was
    <0>.

benjaminfrank:
```
Error: GitHub API rate limit exceeded for 80.87.131.189. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
Try again in 21 minutes 15 seconds, or create an personal access token:
https://github.com/settings/tokens
and then set the token as: HOMEBREW_GITHUB_API_TOKEN
```
Any hint on how I can resolve this issue? 
Or how I can re-trigger a build (without pushing a fake commit)?

DomT4:
We should have fixed that this afternoon. I've re-queued it for you.

benjaminfrank:
Thanks.

mikemcquaid:

For future reference the preferred commit message format for new formulae is `khard 0.4.0 (new formula)`.

PR #41210, opened 2015-06-30
New formula: khal v0.5.0

benjaminfrank:
CLI calendar application http://lostpackets.de/khal/
This formula bundles the vdirsyncer modules as vdirsyncer is not yet
available in hombrew. Once the vdirsyncer formula is availabe via
homebrew I will update this formula.

### DIFF
--- a/Library/Formula/khal.rb
+++ b/Library/Formula/khal.rb
@@ -93,6 +93,6 @@ class Khal < Formula
   end
 
   test do
-    system "#{bin}/khard", "--version"
+    system "#{bin}/khal", "--version"
   end
 end
--- a/Library/Formula/khal.rb
+++ b/Library/Formula/khal.rb
@@ -51,7 +51,6 @@ class Khal < Formula
     sha256 "1950d112ed1b717683280d54f1e7a4533564d479127162cbf247bd0fb3708983"
   end
 
-  # vdirsyncer + deps until available via homebrew
   resource "vdirsyncer" do
     url "https://pypi.python.org/packages/source/v/vdirsyncer/vdirsyncer-0.4.4.tar.gz"
     sha256 "d6f6fa7730c1c87dd10bb5bad3143515ae983987bf4154b5e53cc74aa720abd2"
--- a/Library/Formula/khal.rb
+++ b/Library/Formula/khal.rb
@@ -79,8 +79,8 @@ class Khal < Formula
 
   def install
     ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
-    %w[click configobj tzlocal icalendar urwid pyxdg pytz six python-dateutil tzlocal requests lxml requests-toolbelt atomicwrites vdirsyncer].each do |r|
-      resource(r).stage do
+    resources.each do |r|
+      r.stage do
         system "python", *Language::Python.setup_install_args(libexec/"vendor")
       end
     end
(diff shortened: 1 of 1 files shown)


Library/Formula/khal.rb
```diff
@@ -0,0 +1,98 @@
...
+  end
+
+  resource "atomicwrites" do
+    url "https://pypi.python.org/packages/source/a/atomicwrites/atomicwrites-0.1.5.tar.gz"
+    sha256 "9b16a8f1d366fb550f3d5a5ed4587022735f139a4187735466f34cf4577e4eaa"
+  end
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
+    %w[click configobj tzlocal icalendar urwid pyxdg pytz six python-dateutil tzlocal requests lxml requests-toolbelt atomicwrites vdirsyncer].each do |r|
```
DomT4: Since the list is kinda long here let's just do:

```ruby
resources.each do |r|
  r.stage do
  end
end
```

Library/Formula/khal.rb
```diff
@@ -0,0 +1,98 @@
...
+    url "https://pypi.python.org/packages/source/c/configobj/configobj-5.0.6.tar.gz"
+    sha256 "a2f5650770e1c87fb335af19a9b7eb73fc05ccf22144eb68db7d00cd2bcb0902"
+  end
+
+  resource "tzlocal" do
+    url "https://pypi.python.org/packages/source/t/tzlocal/tzlocal-1.1.3.tar.gz"
+    sha256 "1950d112ed1b717683280d54f1e7a4533564d479127162cbf247bd0fb3708983"
+  end
+
+  # vdirsyncer + deps until available via homebrew
```
DomT4: Would `vdirsyncer` be an application or a library? If the latter we don't accept Python libraries in Homebrew.

benjaminfrank: It is an application, however it seems to expose some functionality for other applications to use via an API. So to me it looks like it is both.

DomT4:
@BrewTestBot test this please

Library/Formula/khal.rb
tdsmith: If you need to use its API, this is the best way to use it.

benjaminfrank: so the python vendor'ed modules should stay in even if vdirsyncer is available as formula? there  is a strong functional dependency between them and I believe their releases are synchronized by the respective maintainers.

tdsmith: I think yes. In order to make vdirsyncer `import`able for khal, you'd have to explicitly add paths from vdirsyncer's libexec to khal's PYTHONPATH, which adds inter-formula coupling.

If vdirsyncer requires python3, probably this also requires python3?

benjaminfrank: Thanks for pointing the python3 thingy out. It seems vdirsyncer doesn't need python3 at all. There was an issue regarding the click dependency and the "test" I wrote bailing out. I will change vdirsyncer.
btw: click isn't needed for khal thus the no python3 thing.

DomT4: Given Tim's comments, let's remove the comment about including it in Homebrew so someone doesn't try and change the resource into a dependency at some point.

benjaminfrank: Done. Thanks for reminding me.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/0b1baaf8d7ada36c6fade9ab851712e28754c0e3. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

It's a minor thing, but if you could squash your commits to one when the formula is "ready" that saves us having to do so locally on pull, Thanks!

PR #41216, opened 2015-06-30
wartremover 0.13 (new formula)

jawshooah:


DomT4:
@BrewTestBot test this please

Library/Formula/wartremover.rb
```diff
@@ -0,0 +1,31 @@
+class Wartremover < Formula
+  desc "Flexible Scala code linting tool"
+  homepage "https://github.com/puffnfresh/wartremover"
+  url "https://github.com/puffnfresh/wartremover/archive/v0.13.tar.gz"
+  sha256 "65d2f9907662bc2cc892801055713dd80a7fc25ade5091cb534bfc51eaae626b"
+
+  head "https://github.com/puffnfresh/wartremover.git", :branch => "master"
```
mikemcquaid: The `:branch` isn't needed here; it defaults to `master` I believe.

jawshooah: Ah, I figured it would default to the remote's default branch (`latest-release`, in this case).

mikemcquaid: Ah, it may. You might want to test that locally, I assumed the default branch was `master` :grinning:

DomT4: Should be [master](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/download_strategy.rb#L549-L554).

jawshooah: Cool beans, I'll push the fix.

Library/Formula/wartremover.rb
```diff
@@ -0,0 +1,31 @@
...
+  sha256 "65d2f9907662bc2cc892801055713dd80a7fc25ade5091cb534bfc51eaae626b"
+
+  head "https://github.com/puffnfresh/wartremover.git"
+
+  depends_on "sbt" => :build
+
+  def install
+    system "sbt", "core/assembly"
+    libexec.install Dir["core/target/scala-*/wartremover-assembly-*.jar"]
+    bin.write_jar_script Dir[libexec/"wartremover-assembly-*.jar"][0], "wartremover"
```
DomT4: Why the need for `[0]` here?

jawshooah: `Dir[pattern]` returns a collection, and `write_jar_script` expects a single `Path` as its first argument. The globbing is necessary because we don't know what version number will replace the `*`.

PR #41218, opened 2015-06-30
rtmidi: use https homepage

vszakats:


DomT4:
:rocket:, Thanks @vszakats!

Issue #41086, opened 2015-06-25
Pango install fail on 10.11


mikestecker:
I know I'm running a beta, but wanted to report this anyway.

```
==> Installing fontforge dependency: pango
==> Downloading https://download.gnome.org/sources/pango/1.36/pango-1.36.8.tar.xz
==> Downloading from https://mirror.umd.edu/gnome/sources/pango/1.36/pango-1.36.8.tar.xz
######################################################################## 100.0%
==> ./configure --disable-silent-rules --prefix=/usr/local/Cellar/pango/1.36.8_1 --enable-man --with-html-dir=/usr/local/Cellar/pango/1.36.8_1/share/doc --enable-introspec
==> make
make[4]: *** [PangoCairo-1.0.gir] Error 1
make[3]: *** [all-recursive] Error 1
make[2]: *** [all] Error 2
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

I have the latest Xcode beta (beta 2) installed:
```
xcode-select -p
/Applications/Xcode-beta.app/Contents/Developer

xcode-select -v
xcode-select version 2342
```

Logs: https://gist.github.com/98ae04d68f1c8fa409df

tdsmith:
It looks like g-ir-scanner is choking on the system Carbon headers: https://gist.github.com/anonymous/98ae04d68f1c8fa409df#file-02-make-L322-L351

mmindenhall:
Just saw the same error with current released version of Xcode:
```
$ xcode-select -v
xcode-select version 2339.
```

mmindenhall:
This worked for me:

1.  `brew info pango`
1.  `brew uninstall --force <all listed dependencies>`
1.  `brew install pango`

DomT4:
```
Build: xz ✔, pkg-config ✔
Required: glib ✔, cairo ✔, harfbuzz ✔, fontconfig ✔, gobject-introspection ✔
```

Solution is probably to rebuild `gobject-introspection` then, I'd imagine. If that works as a solution for @mikestecker as well I'll stick it in the 10.11 tracking list as a workaround.

mikestecker:
At first I tried this based off @DomT4's suggestion:

```
brew uninstall --force object-introspection
brew install pango
```

But this still failed to install with the same errors as I originally stated.

[...]

Issue #40853, opened 2015-06-18
pango-1.36.8 failed to build on 10.10.3 (14D136)


themsay:
https://gist.github.com/b993db679b606702efea

bernatfp:
Can't build pango either.

https://gist.github.com/1c2f2df1ccd13b9d6579

DomT4:
@themsay Yours' seems to be a simple case that it can't find Cairo. Please ensure you're up-to-date with `brew update` and then `brew install cairo` - Try again afterwards. You may need to `brew upgrade` as well, we made some major GTK+ changes yesterday.

@bernatfp Your error is a little unusual. `make` isn't terribly verbose about what's going wrong there. You have a lot of `doctor` warnings that could be contributing though. I'd recommend dealing with as many of those as you can, particularly:

* https://gist.github.com/anonymous/1c2f2df1ccd13b9d6579#file-doctor-out-L39-L41
* https://gist.github.com/anonymous/1c2f2df1ccd13b9d6579#file-doctor-out-L73-L91
* https://gist.github.com/anonymous/1c2f2df1ccd13b9d6579#file-doctor-out-L120-L122

themsay:
@DomT4 <code>brew install cairo</code> return Warning: cairo-1.14.2_1 already installed and <code>brew upgrade cairo</code> also return Error: cairo 1.14.2_1 already installed.

themsay:
I sort the issue by simply running:

<code>brew uninstall cairo</code>

then

<code>brew update && brew upgrade --all</code>

it automatically install pango and cairo successfully.

DomT4:
^ :+1:. `--all` is no longer a thing by the way. It was reverted.

tko:
The error in @bernatfp case seems to be something wrong with python:
```
/bin/sh: /usr/local/Cellar/gobject-introspection/1.44.0/bin/g-ir-scanner: /usr/bin/python: bad interpreter: No such file or directory
```
https://gist.github.com/anonymous/1c2f2df1ccd13b9d6579#file-02-make-L140

I got the same error because of /usr/local/opt/python/bin/python but I suspect that's because I've uninstalled the brew python because it was messing other things up. Missing /usr/bin/python seems bad.

bernatfp:
@tko that's probably the issue. I created a symlink and that fixed the problem:

```
ln -s /usr/bin/python2.7 /usr/bin/python
```

Thanks @themsay @tko

juanfal:
[...]

PR #41220, opened 2015-06-30
pango HEAD install fails due to Gnome.org URL

webavant:
Adding ".git" to URL fixes the problem.

To reproduce the error:

    brew install pango --HEAD

Brew error *without* --interactive:

    Pango install fail on 10.11 (https://github.com/Homebrew/homebrew/issues/41086)
    pango-1.36.8 failed to build on 10.10.3 (14D136) (https://github.com/Homebrew/homebrew/issues/40853)

Brew error *with* --interactive:

    $ ./autogen.sh
    bash: ./autogen.sh: No such file or directory

running ls -la inside --interactive shell:

    drwx------   3 webavant  wheel    102 Jun 30 13:40 .
    drwxrwxrwt  45 webavant  wheel   1530 Jun 30 13:40 ..
    -rw-r--r--   1 webavant  wheel  19648 Jun 30 13:40 pango

Contents of the 'pango' file is the downloaded webpage from the HEAD url, which is https://git.gnome.org/browse/pango. Simply appending '.git' to that url fixes the problem.

### DIFF
--- a/Library/Formula/pango.rb
+++ b/Library/Formula/pango.rb
@@ -6,7 +6,7 @@ class Pango < Formula
   revision 1
 
   head do
-    url "https://git.gnome.org/browse/pango"
+    url "https://git.gnome.org/browse/pango.git"
 
     depends_on "automake" => :build
     depends_on "autoconf" => :build


DomT4:
@BrewTestBot test this please

DomT4:
Yeah, essentially in this situation unless the URL is appended with `.git` or `:using => :git` we attempt to fetch it with `curl`, which is obviously causing the failure.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/27e427d8b58811cb6e4de0904cf983507719bdc0. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a nit, but something like `pango: fix head fetch` or similar would have been best here for the commit message. Thanks!

webavant:
Thanks for the feedback. :100:

PR #41224, opened 2015-06-30
curl: use https url

vszakats:
The new URL points to GitHub, therefore I deleted the 
previous GitHub mirror kindly maintained by @DomT4.

Announcement: http://curl.haxx.se/news.html [June 29 2015]
Download page: http://curl.haxx.se/download.html

### DIFF
--- a/Library/Formula/curl.rb
+++ b/Library/Formula/curl.rb
@@ -1,8 +1,8 @@
 class Curl < Formula
   desc "Get a file from an HTTP, HTTPS or FTP server"
   homepage "http://curl.haxx.se/"
-  url "http://curl.haxx.se/download/curl-7.43.0.tar.bz2"
-  mirror "https://raw.githubusercontent.com/DomT4/LibreMirror/master/cURL/curl-7.43.0.tar.bz2"
+  url "https://github.com/bagder/curl/releases/download/curl-7_43_0/curl-7.43.0.tar.bz2"
+  mirror "http://curl.haxx.se/download/curl-7.43.0.tar.bz2"
   sha256 "baa654a1122530483ccc1c58cc112fec3724a82c11c6a389f1e6a37dc8858df9"
 
   bottle do
@@ -98,7 +98,7 @@ class Curl < Formula
     # Fetch the curl tarball and see that the checksum matches.
     # This requires a network connection, but so does Homebrew in general.
     filename = (testpath/"test.tar.gz")
-    system "#{bin}/curl", stable.url, "-o", filename
+    system "#{bin}/curl", "-L", stable.url, "-o", filename
     filename.verify_checksum stable.checksum
   end
 end


DomT4:
@BrewTestBot test this please

vszakats:
Strange bot failure given that the new URL returns 
a package with the exact same SHA256 as before.
(can't reach the bot results to verify)

xu-cheng:
Won't this cause bootstrap problem for Linuxbrew? If the system curl is not compiled with SSL support, they won't be able to use brew to install curl with SSL support.

vszakats:
It's a possibility, but I think it's very unlikely that any system curl would ever be shipped without SSL support. It would cause a lot more breakage than this one.

We can put back the original plaintext `url` as a `mirror`, which will work till `curl.haxx.se` is accessible using plaintext HTTP.

xu-cheng:
> We can put back the original plaintext url as a mirror

:+1: Make sense to me.

vszakats:
Done and squashed.

vszakats:
Bot seems to overlooked this one.

DomT4:
Our CI is down at the moment. It'll be re-queued when we have something to re-queue it with :wink:

vszakats:
Thanks @DomT4 for the swift response. Bot had some though days, I'll be patient :)

DomT4:
Failed again:

```
==> brew test --verbose curl						 FAILED
Testing curl
==> /usr/local/Cellar/curl/7.43.0/bin/curl https://github.com/bagder/curl/releases/download/curl-7_43_0/curl-7.43.0.tar.bz2 -o /private/tmp/curl20150702-78090-5qhwvk/test.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   395    0   395    0     0    722      0 --:--:-- --:--:-- --:--:--   768
Error: curl: failed
SHA256 mismatch
Expected: baa654a1122530483ccc1c58cc112fec3724a82c11c6a389f1e6a37dc8858df9
Actual: 6e818ed850698773d26312421399961fe4148a5ecc217aa73563ff0b2fdcccc0
Archive: /private/tmp/curl20150702-78090-5qhwvk/test.tar.gz
```

What happens locally for you on the test?

DomT4:
Oh, I see what's happening. Github puts a lot of its releases on AWS rather than hosting them "directly" per se. It jumps from one connection to the other in `brew fetch`, but unless you tell `curl` to do that `curl` will only make the first hop (To Github) and not onto the actual download server. Presumably we're just downloading an empty html file or similar in the test now.

You'll need to tweak the `test do` block so `curl` follows redirection.

vszakats:
I completely overlooked that the test section will want to download self. Looking into it.

DomT4:
New error:

```
Error: curl: failed
undefined local variable or method `filename' for #<Formula curl (stable) /usr/local/Library/Formula/curl.rb>
/usr/local/Library/Formula/curl.rb:101:in `block in <class:Curl>'
/usr/local/Library/Homebrew/formula.rb:797:in `block in run_test'
/usr/local/Library/Homebrew/formula.rb:793:in `run_test'
/usr/local/Library/Homebrew/test.rb:28:in `block in <main>'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/timeout.rb:66:in `timeout'
/usr/local/Library/Homebrew/test.rb:27:in `<main>'
```

It's because you removed the `filename` variable but it's still being used on the last line of the test.

vszakats:
Gosh, sorry. Correcting.

vszakats:
Should be fine now, squashed too. Thanks for your patience.

mikemcquaid:
Thanks @vszakats!

PR #41228, opened 2015-07-01
Support installing gtk+ on other OSes

paulhybryant:
Support installing gtk+ on other OSes


mistydemeo:
Can you submit this to Linuxbrew? We prefer to keep these kinds of patches out of core formulae.

paulhybryant:
Sure. I will push this to linuxbrew. Thanks.

PR #41230, opened 2015-07-01
dos2unix: Remove gettext dependency

chdiza:
Remove the gettext dependency, which is not mandatory but only goes along with NLS support.

### DIFF
--- a/Library/Formula/dos2unix.rb
+++ b/Library/Formula/dos2unix.rb
@@ -16,9 +16,9 @@ class Dos2unix < Formula
     sha256 "59cea39b181913532bf9e9c234a142c15e330d5eee145cd6b90f54becd6ec27b"
   end
 
-  option "with-nls", "Build with Native Language Support"
+  option "with-gettext", "Build with Native Language Support"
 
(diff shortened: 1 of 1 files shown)


chdiza:
I am open to making the dep optional, and only invoked if the user passes `--with-nls` or some such.

DomT4:
Same question here. Probably a good idea to leave it `:optional`.

Library/Formula/dos2unix.rb
```diff
@@ -16,17 +16,28 @@ class Dos2unix < Formula
     sha256 "59cea39b181913532bf9e9c234a142c15e330d5eee145cd6b90f54becd6ec27b"
   end
 
-  depends_on "gettext"
+  option "with-nls", "Build with Native Language Support"
+
+  depends_on "gettext" if build.with? "nls"
```
mikemcquaid: `=> :optional` and expand the `with-gettext` description, thanks.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/57dfee09d27359aaa7bc62c846806108c56ba45b. Thanks again @chdiza!

PR #41231, opened 2015-07-01
pv: Remove gettext dependency

chdiza:
Remove the gettext dependency from `pv`, since it's not mandatory.

### DIFF
--- a/Library/Formula/pv.rb
+++ b/Library/Formula/pv.rb
@@ -12,7 +12,7 @@ class Pv < Formula
 
   option "with-nls", "Build with Native Language Support"
 
-  depends_on "gettext" => :optional
+  depends_on "gettext" if build.with? "nls"
 
   fails_with :llvm do
     build 2334
--- a/Library/Formula/pv.rb
+++ b/Library/Formula/pv.rb
@@ -10,15 +10,24 @@ class Pv < Formula
     sha256 "f343368e557cb1c86173bd0c62143b34834e2b825b1a188ac2a37c23d0c685dd" => :mountain_lion
   end
 
+  option "with-nls", "Build with Native Language Support"
+
+  depends_on "gettext" => :optional
+
   fails_with :llvm do
     build 2334
   end
 
   def install
-    system "./configure", "--disable-debug",
-                          "--prefix=#{prefix}",
-                          "--mandir=#{man}",
-                          "--disable-nls"
+    args = %W[
+      --disable-debug
+      --prefix=#{prefix}
+      --mandir=#{man}
+    ]
+
+    args << "--disable-nls" if build.without? "nls"
+
+    system "./configure", *args
     system "make", "install"
   end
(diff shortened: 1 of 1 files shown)


DomT4:
Is it worth keeping around as `:optional`?

chdiza:
I don't know.  These two packages in particular seem like there's not really a need for NLS, since they don't have tons of strings that need translating.

I'll make it optional just in case.

I notice we don't have a consistent policy on NLS.  We have formulas that have it disabled by default with an optional turn-on (e.g. `shmcat`); we have formulas where it's on by default with an optional turn-off (e.g. `pulseaudio`); we have formulas where it's on and there's no way to turn it off (like `pv` here and `dos2unix`); and we have formulas where it's off and there's no way to turn it on (e.g. `cogl`).

I personally favor keeping it off unless the user requests it to be on.  There's just no reason to lug around a needless `gettext`, or to have the doctor tell one `gettext` is missing when one makes a custom formula that eliminates the NLS.

DomT4:
No, I don't think there's ever been a centralised decision made about NLS support and how default it should be.

I suspect we're deliberately taking it on a case-by-case basis depending on demand/value of the extra dependency required. If we set a policy my preference would be off by default with an optional on, but maybe this is something other @Homebrew/owners' want to weigh in on a bit.

tdsmith:
I am generally in favor of accommodating users outside of the Anglosphere by default but maybe this isn't the poster-child formula for that cause.

Library/Formula/pv.rb
```diff
@@ -10,16 +10,24 @@ class Pv < Formula
     sha256 "f343368e557cb1c86173bd0c62143b34834e2b825b1a188ac2a37c23d0c685dd" => :mountain_lion
   end
 
-  depends_on "gettext"
+  option "with-nls", "Build with Native Language Support"
+
+  depends_on "gettext" if build.with? "nls"
```
mikemcquaid: Let's just do `:optional` or `:recommended` here, thanks.

chdiza: I would've done that, but then the user has to do "--with-gettext"  instead of the infinitely more intuitive "--with-nls".

DomT4: We've been trying to kill off `if build.with?` deps. Adding new ones is a bit bleh.

Agree with Mike we should make it `=> :optional` instead; If it's helpful we can make the option `option "with-gettext", "Build with Native Language Support"` so it's more explicit what it brings to the party.

mikemcquaid: Ideally we'll add a DSL hack at some point that lets you alias `with-gettext` as `with-nls`

DomT4:
Squashed & Merged in https://github.com/Homebrew/homebrew/commit/0c4647e2534afb97ce3842c2bb9b180f9aa8c163. Thanks @chdiza!

PR #41236, opened 2015-07-01
Head build for StrongSWAN

carlhoerberg:


Library/Formula/strongswan.rb
```diff
@@ -15,6 +15,15 @@ class Strongswan < Formula
 
   depends_on 'openssl'
   depends_on 'curl' => :optional
+  depends_on 'autoconf' => :build
```
DomT4: If these aren't required for the stable build you shouldn't put them in the core formula, otherwise everyone will require them regardless of whether they build head or not. Should be:

```ruby
head do
  url "https://git.strongswan.org/strongswan.git"

```

And so on.

DomT4: And double quotes everywhere. We're moving away from single quotes.

carlhoerberg:
Ok, thank you for the feedback @DomT4, fixed.

DomT4:
Is this incapable of using the system `flex` & `bison`?

carlhoerberg:
Could use system `flex` but not `bison`

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/c8a4df2531cc8f8afca53f92f77ab384ffebfe94. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just for future reference, `strongswan: add head` or similar would have been ideal as the commit message here, Thanks!

PR #41242, opened 2015-07-01
Make audit recognise hed-only and devel-only in all taps

rkrug:
- add wildcard into expressions to identify head-only and dev-only taps

### DIFF
--- a/Library/Formula/libuv.rb
+++ b/Library/Formula/libuv.rb
@@ -6,9 +6,9 @@ class Libuv < Formula
 
   bottle do
     cellar :any
-    sha256 "36280ae768a4bad91583ae76488d46c31a1e68f6f01b684a5e01f8a88cb1ec7a" => :yosemite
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/audit.rb
```diff
@@ -375,11 +375,11 @@ def audit_homepage
   end
 
   def audit_specs
-    if head_only?(formula) && formula.tap.downcase != "homebrew/homebrew-head-only"
+    if head_only?(formula) && formula.tap.downcase != "*/homebrew-head-only"
```
jacknagel: This won't work... this is a literal string comparison, not a wildcard pattern.

DomT4:
In addition to Jack's comment there's a merge conflict here as well apparently.

Library/Homebrew/cmd/audit.rb
rkrug: Sorry - how can I make this work? How can I compare in ruby with wildcard pattern?

rkrug:
I'll supply an updated one later in a new pull request.

PR #41249, opened 2015-07-01
elixir 1.0.5

colinrymer:
Bottle SHAs need updating, but looks like you all handle that per the CONTRIBUTING docs.


colinrymer:
Thought I searched for an elixir update PR, but I was searching for issues instead. :grimacing:

PR #40917, opened 2015-06-20
Create scaleway-cli formula

nicolas-brousse:
I create a new formula to add the [scaleway-cli](https://github.com/scaleway/scaleway-cli).

It exists a binary for 32-bits OSs but I don't know how I have to precise it. Maybe it could be added.
And also I'm not sure if I have to precise bottle and how.

PR #41250, opened 2015-07-01
Initial version for scw (1.2.1)

moul:
This formula download sources using Github tag release and build using Golang environment.
The sources contain all requirements (Godeps).

As soon as this is merged, I will add a link to official homebrew formula in our source formula (https://github.com/scaleway/scaleway-cli/tree/master/contrib/builder/homebrew), and update README.md+releases with brew instructions.

Related to #40917

### DIFF
--- a/Library/Formula/scw.rb
+++ b/Library/Formula/scw.rb
@@ -1,0 +1,54 @@
+require "language/go"
+
+class Scw < Formula
+  desc "Manage BareMetal Servers from Command Line (as easily as with Docker)"
+  homepage "https://github.com/scaleway/scaleway-cli"
+  url "https://github.com/scaleway/scaleway-cli/archive/v1.2.1.tar.gz"
+  sha256 "e3ae09558a5f451935831381177f1ee3ce50aec438a1e11269dfc380e0e196c9"
+
+  head "https://github.com/scaleway/scaleway-cli.git"
+
+  depends_on "go" => :build
+
+  go_resource "github.com/tools/godep" do
+    url "https://github.com/tools/godep.git", :revision => "58d90f262c13357d3203e67a33c6f7a9382f9223"
+  end
+
+  go_resource "github.com/kr/fs" do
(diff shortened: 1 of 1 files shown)


Library/Formula/scw.rb
```diff
@@ -0,0 +1,54 @@
...
+    bin.install "scw"
+
+    bash_completion.install "contrib/completion/bash/scw"
+    zsh_completion.install "contrib/completion/zsh/_scw"
+  end
+
+  test do
+    output = shell_output(bin/"scw version")
+    assert output.include? "Client version: v1.2.1"
+    assert output.include? "Git commit (client): 3eb2cfb2478410b61992396fad5081ee80a140e9"
```
tdsmith: Can the test be chosen so that it doesn't need to be updated at each version? Running e.g. `system bin/"scw", "ps"` should be enough as long as that exits with code 0.

tdsmith:
Looks great.

Library/Formula/scw.rb
moul: Actually, the only two commands working without being authenticated are:

- `scw login` -> to get authenticated
- `scw version`

I think for now we need to stay with `scw version` as test command, but we can check a non-dynamic part of the output:

```
Client version: v1.2.0-1-g4cf44a8
Go version (client): go1.4.2
Git commit (client): 4cf44a8411998495ba00e93d95c6c8a1e888f557
OS/Arch (client): darwin/amd64
```

The **OS/Arch (client): darwin/amd64** won't change with new versions, is that better ?

tdsmith: That works for me.

DomT4:
@BrewTestBot test this please

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/36265eab1bb45e99d373abf65c7730473180ca6a. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a nit, but ideal commit message style for new formulae is `scw 1.2.1 (new formula)` or similar, Thanks!

PR #41257, opened 2015-07-01
Make HOMEBREW_GITHUB_API_TOKEN copy/paste friendly

breadtk:
A simple change that allows for easy copy/pasting. Right now if you were to double click on the text it will also inadvertently copy the trailing period. This change will alleviate the unnecessary gymnastics of not capturing the trailing period.

/most_pedantic_pull_request_ever


xu-cheng:

PR #41258, opened 2015-07-01
magit 2.1.0

dunn:


Library/Formula/magit.rb
```diff
@@ -37,22 +16,30 @@ class Magit < Formula
...
+    resource("dash").stage do
+      (share/"emacs/site-lisp").install "dash-20150701.20.el" => "dash.el"
     end
+    (buildpath/"config.mk").write <<-EOS
+      LOAD_PATH = -L #{buildpath}/lisp -L #{share}/emacs/site-lisp
+    EOS
+
+    args = %W[
+      PREFIX=#{prefix}
+      lispdir=#{share}/emacs/site-lisp
```
dunn: I changed this from `#{share}/emacs/site-lisp/magit` to match other formulae, but is that correct?

vermiculus:
Since when did Homebrew start gobbling up the work of other package managers (`package.el`, in this case)? (Honest question; last time I checked, this was frowned upon. There's no philosophical difference between this and adding every ruby gem or python module under the sun.)

dunn:
From what I can tell, Emacs packages are allowed if they're not bundled with Emacs or part of the official ELPA package repository. The maintainers can correct me, though.

mikemcquaid:
@vermiculus How do you install it with that package manager?

vermiculus:
@mikemcquaid When MELPA is added to `package-archives` (not unlike tapping in homebrew-speak), you can just `M-x package-install RET magit RET`. You can also install from `M-x list-packages` (mark to install with `i`, delete with `d`, execute marks with `x`).

You can also use Marmalade instead of MELPA, but it's not as common in my experience.

```elisp
(let ((package-archives
       (cons '("melpa" . "http://melpa.org/packages/")
             package-archives)))
  (package-refresh-contents)
  (package-install "magit"))
```

mikemcquaid:
@dunn Thoughts?

DomT4:
> Since when did Homebrew start gobbling up the work of other package managers (package.el, in this case)? 

Just FWIW, this particular package was added in 2010:

```
commit a385f9e144ead04889f6e05b822022a1dfbf55e2
Author: Adam Vandenberg <flangy@gmail.com>
Date:   Thu Jul 1 21:40:11 2010 -0700

    Add magit as separate brew at 0.8.1.
```

dunn:
If its inclusion is an anomaly then I'll understand if you'd rather boneyard it.  I've enjoyed being able to switch between stable and head builds more easily, but I can do that in my own tap if I need to.

mikemcquaid:
@dunn That's a good enough reason for me for it to be in Homebrew.

mikemcquaid:

PR #41280, opened 2015-07-02
New formula: afew 0.0.0

benjaminfrank:
afew is an initial tagging script for notmuch mail. Afew does not
provide tags/releases, therefore I took commit
162d9fa00ba09892bd8d63ec9fa182e0319000a2 and called the version
artificially 0.0.0
I ask upstream to do proper tags/releases.

### DIFF
--- a/Library/Formula/afew.rb
+++ b/Library/Formula/afew.rb
@@ -1,9 +1,6 @@
 class Afew < Formula
   desc "afew is an initial tagging script for notmuch."
   homepage "https://github.com/teythoon/afew"
-  url "https://github.com/teythoon/afew/archive/162d9fa00ba09892bd8d63ec9fa182e0319000a2.tar.gz"
-  version "0.0.0"
-  sha256 "10fbcb1d7e9029da05d8f352499196867f40487aad9e23c12034264f88559d35"
   head "https://github.com/teythoon/afew.git"
 
   depends_on :python if MacOS.version <= :snow_leopard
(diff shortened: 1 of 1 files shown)


bfontaine:
It’s probably better to have this in head-only until upstream provides stable releases.

benjaminfrank:
Does homebrew offer something to signal that a new "head" version exists, like an internal version number that I could bump whenever new commits come in?

bfontaine:
> Does homebrew offer something to signal that a new "head" version exists

It supports `HEAD` URLs, which it’ll use to pull the changes as they come.

benjaminfrank:
But subsequent brew updates would not check if the head changed, right? So an explicit remove/install would be required to "update" a head-only formula. 
What I was thinking of is more like an internal version for a head formula such that I could bump the formula and a upgrad/update would fetch the "new" head.

DomT4:
> But subsequent brew updates would not check if the head changed, right?

Correct.

> So an explicit remove/install would be required to "update" a head-only formula. 

Correct.

> What I was thinking of is more like an internal version for a head formula such that I could bump the formula and a upgrad/update would fetch the "new" head.

We don't have anything like this, because obviously depending on when you clone the `HEAD` formula you won't have installed the same "version". It can be done with `revision` (commit) usage, but then it isn't really `head-only` or stable.

benjaminfrank:
if this is now head-only, do i need to open a new PR in homebrew-head-only?

mikemcquaid:
Yeh, sorry.

benjaminfrank:
Moving over to hombrew-head-only.

PR #41288, opened 2015-07-02
add missing luarock dependency for SILE

alerque:
The SILE typesetter actually has a dependency on the `luafilesystem` library that needs to be installed as a Lua rock. This is per [upstream issue 119](https://github.com/simoncozens/sile/pull/119), and the documentation has been updated upstream to reflect this. The library is only used for certain operations so it wasn't immediately apparent that it was failing.

Now that the upstream makefile demands it this formula will error on installing from HEAD, but ever though the current 0.9.2 release didn't call for it explicitly it needs this too.

### DIFF
--- a/Library/Formula/sile.rb
+++ b/Library/Formula/sile.rb
@@ -33,6 +33,7 @@ class Sile < Formula
   def install
     system "luarocks-5.2", "install", "lpeg"
     system "luarocks-5.2", "install", "luaexpat"
+    system "luarocks-5.2", "install", "luafilesystem"
     system "./bootstrap.sh" if build.head?
     system "./configure", "--disable-debug",
                           "--disable-dependency-tracking",
--- a/Library/Formula/sile.rb
+++ b/Library/Formula/sile.rb
@@ -4,13 +4,6 @@ class Sile < Formula
   url "https://github.com/simoncozens/sile/releases/download/v0.9.2/sile-0.9.2.tar.bz2"
   sha256 "2223582818df06daa4609cee40a81e8787085ad1795d4b2ce5edbe0663b74e18"
 
-  bottle do
-    cellar :any
-    sha256 "aacf9f59beae5b6de301b7e865d8aef602d47d6aa99ac8a87615c267a90a8646" => :yosemite
-    sha256 "04bb77b8e45794a08c31ad9efea24d722d704b31180ff89d4da80d5696150b73" => :mavericks
-    sha256 "3e634157cf79e6be843c73034cd4f4d7099c760a3ed2b2d2905e42512dac7c2c" => :mountain_lion
-  end
-
   head do
     url "https://github.com/simoncozens/sile.git"
     depends_on "automake" => :build
@@ -25,15 +18,16 @@ class Sile < Formula
   depends_on "freetype"
   depends_on "lua"
 
+  depends_on "lpeg" => :lua
+  depends_on "luaexpat" => :lua
+  depends_on "luafilesystem" => :lua
+
   resource("simple.sil") do
     url "https://raw.githubusercontent.com/simoncozens/sile/v0.9.2/examples/simple.sil"
     sha256 "f788723cd984d98343c8f8dc1b93b8f769cea6894a28f9ba6e0bec6aebf78f92"
   end
 
   def install
-    system "luarocks-5.2", "install", "lpeg"
-    system "luarocks-5.2", "install", "luaexpat"
-    system "luarocks-5.2", "install", "luafilesystem"
     system "./bootstrap.sh" if build.head?
     system "./configure", "--disable-debug",
                           "--disable-dependency-tracking",


DomT4:
We're not handling the `luarocks` stuff correctly here. It's not actually ending up in the bottle at all. From-source compile should be fine, but the bottle contains no rocks.

```
/usr/local/opt/lua/bin/lua: /usr/local/Cellar/sile/0.9.2/share/sile/core/parserbits.lua:1: module 'lpeg' not found:
	no field package.preload['lpeg']
	no file 'lpeg.lua'
	no file '/usr/local/Cellar/sile/0.9.2/share/sile/lpeg.lua'
	no file '/usr/local/Cellar/sile/0.9.2/share/sile/lua-libraries/lpeg.lua'
	no file '/usr/local/Cellar/sile/0.9.2/share/sile/lua-libraries/lpeg/init.lua'
	no file 'lua-libraries/lpeg.lua'
	no file 'lua-libraries/lpeg/init.lua'
	no file '/usr/local/share/lua/5.2/lpeg.lua'
	no file '/usr/local/share/lua/5.2/lpeg/init.lua'
	no file '/usr/local/lib/lua/5.2/lpeg.lua'
	no file '/usr/local/lib/lua/5.2/lpeg/init.lua'
	no file './lpeg.lua'
```

So on, so forth.

Give me a bit, I'll look at it tonight.

alerque:
@DomT4 I asked about the lua rocks issue when I originally submitted this in #40893, but didn't get any feedback. The short version is that I couldn't find any examples of packages with dependences on lua rocks that seemed like the right model to follow, and what is in the documented sample just doesn't work. Thanks for taking a look at this!

DomT4:
Sorry, I didn't see the first PR. I try to glance over most things but we field a lot of PRs, heh. There aren't many packages in Homebrew that require Lua modules as far as I know, it's a relatively rare thing, although we do attempt to support it. What documented sample did you look at?

alerque:
I know it's easy to loose stuff in the flood here, but I please see #40893 (which you tagged btw). I linked the example I found and even included a code snipped of what I tried that didn't work.

TL;DR: [This example](https://github.com/Homebrew/homebrew/blob/master/Library/Contributions/example-formula.rb) makes it sound like adding `depends_on "luafilesystem" => :lua` should work but it doesn't.

DomT4:
> which you tagged btw

I likely tagged it as part of a batch. I tend to use Github's drop-down mass-tagger to do several PRs at once if they have `new formula` somewhere in the title, but my apologies.

Essentially this is passing the bot but not working on local machines (Unless built from source) because of the `luarocks-5.2` commands: On the test-bot it follows those commands and installs the modules you desire, and consequently when the `configure` script asks Lua if those modules are present Lua says yes they are and goes ahead with the build, but because our test-bot only packages things that end up in the formula prefix it doesn't package the modules and those modules aren't carried across to the user's end machine inside the bottle.

The mechanism we have in place with:

```ruby
  depends_on "freetype"
  depends_on "lua"

  depends_on "lpeg" => :lua
  depends_on "luaexpat" => :lua
  depends_on "luafilesystem" => :lua
```

Is actually working:

```
~> brew install sile
sile: Unsatisfied dependency: luaexpat
Homebrew does not provide Lua dependencies; install with:
  luarocks install luaexpat
Error: An unsatisfied requirement failed this build.
```

It just means we can't build bottles for it, which is expected, and not a huge deal given the compile is relatively swift.

Long-term I am looking at changing the way we deal with Lua resources, but for now we'll need to scrap the bottles and adopt the above syntax. Apologies the review came rather late.

alerque:
> Long-term I am looking at changing the way we deal with Lua resources, but for now we'll need to scrap the bottles and adopt the above syntax. Apologies the review came rather late.

No worries on the timing — in fact I edited my comment before you even replied because I felt my tone came across harsh which was totally un-called for.

I'll update this PR in a second. I do have a question though --- do I leave the `system()` callouts that install the rocks or is the depends syntax a complete replacement for that?

DomT4:
Thanks for the understanding. I know it can be frustrating to only discover these things further down the line than desired, and I didn't read anything more than understandable frustration into the comment.

> do I leave the system() callouts that install the rocks

Yeah. The test-bot may skip the formula, but it shouldn't fail, and we can ignore it if it does. The `depends_on "such" => :lua` syntax is a complete replacement. If a user tries to install the formula without having the modules installed Homebrew will halt, tell the user what to install, and then the user will be able to do `brew install sile` again and it'll work as intended.

```
checking for required Lua library lpeg... found
checking for required Lua library lxp... found
```

Library/Formula/sile.rb
```diff
@@ -33,6 +30,7 @@ class Sile < Formula
   def install
     system "luarocks-5.2", "install", "lpeg"
```
DomT4: Can kill these three lines now :)

alerque:
@DomT4 I'm slightly confused by your reply. The bit you quoted and said "yeah" to was the first option of my OR question, but the rest of your explanation makes it sound like the second option is preferred. So is 94b7ff9 right or wrong to leave the install lines?

Library/Formula/sile.rb
alerque: Gotcha. Killing with fire and `--fixup`.

DomT4:
> I'm slightly confused by your reply.

I have apparently lost the ability to write replies this afternoon. Doesn't bode well for the essay I'm currently trying to finish :smile_cat:. I quoted the wrong bit of what you said. `depends_on` is a complete replacement and the system calls to `luarocks-5.2` are good to :skull:.

alerque:
Thanks for your help with this @DomT4, and good luck with your essay. Here's a quick oneliner if you want to check for OR questions followed by boolean answers.

    sed -n '/or/{N;/(yes|no)/p}' essay.tex

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/94acf107a26677609ae9d02c89879a296af6e23f. 

Thanks for your contribution to Homebrew, and the code snippet :wink:, we appreciate it! :smiley_cat:

Issue #13197, opened 2012-07-04
brew upgrade for HEAD formulas fails with error: already installed


jasonblewis:
brew upgrade formula  on a formula that was initially installed with the --HEAD option fails with:

```
Error: formula-HEAD already installed
```

There is no clear way to upgrade a formula installed with --HEAD other than uninstalling and re-installing. Example:

```
jason@woodhouse:/usr/local/Library/Formula$ brew install --HEAD wol
==> Cloning https://github.com/kylef/wol.git
Cloning into /Users/jason/Library/Caches/Homebrew/wol--git...
remote: Counting objects: 13, done.
remote: Compressing objects: 100% (12/12), done.
remote: Total 13 (delta 2), reused 0 (delta 0)
Unpacking objects: 100% (13/13), done.
==> make
/usr/local/Cellar/wol/HEAD: 4 files, 24K, built in 16 seconds
jason@woodhouse:/usr/local/Library/Formula$ brew upgrade wol
Error: wol-HEAD already installed
jason@woodhouse:/usr/local/Library/Formula$ brew upgrade --HEAD wol
Error: wol-HEAD already installed
jason@woodhouse:/usr/local/Library/Formula$
```

Max Howell suggested that this should either be fixed or the error message improved and requested I open this ticket.

apjanke:
Closing this out as not doable for the time being, but would welcome a PR over at homebrew/brew in the future.

PR #41289, opened 2015-07-02
Make audit recognise head-only and devel-only in all taps

rkrug:
- add wildcard into expressions to identify head-only and dev-only
  taps to avoid "head-only" and "devel-only" errors when recipie is in
  *head-only or *devel-only tap


rkrug:
This one is working and tested locally.

dunn:
Looks like you accidentally committed a patch file too.

rkrug:
Envoyé de mon iPhone

> Le 2 juil. 2015 à 18:14, Alex Dunn <notifications@github.com> a écrit :
> 
> Looks like you accidentally committed a patch file too.
> 

True - sorry.

Library/Homebrew/cmd/audit.rb
```diff
@@ -421,11 +421,11 @@ def audit_homepage
   end
 
   def audit_specs
-    if head_only?(formula) && formula.tap.to_s.downcase != "homebrew/homebrew-head-only"
+    if head_only?(formula) && formula.tap.to_s.downcase !~ /.*head-only/
```
mikemcquaid: `/-head-only/` should be enough here, I believe.

Library/Homebrew/cmd/audit.rb
```diff
@@ -421,11 +421,11 @@ def audit_homepage
...
       problem "Head-only (no stable download)"
     end
 
-    if devel_only?(formula) && formula.tap.to_s.downcase != "homebrew/homebrew-devel-only"
+    if devel_only?(formula) && formula.tap.to_s.downcase !~ /.*devel-only/
```
mikemcquaid: and `/-devel-only/` here.

Library/Homebrew/cmd/0001-Make-audit-recognise-hed-only-and-devel-only-in-all-.patch
```diff
@@ -0,0 +1,31 @@
...
+     end
+ 
+-    if devel_only?(formula) && formula.tap.to_s.downcase != "homebrew/homebrew-devel-only"
++    if devel_only?(formula) && formula.tap.to_s.downcase != "*/homebrew-devel-only"
+       problem "Devel-only (no stable download)"
+     end
+ 
+-- 
+2.4.5
+
```
mikemcquaid: Please remove this file.

mikemcquaid:
@Homebrew/owners I think I prefer this approach to https://github.com/Homebrew/homebrew/issues/41413 and I think given there's a few tap authors asking for this we should merge something.

jawshooah:
@mikemcquaid This would require any head-only or devel-only formulae to be split into separate taps, right?

mikemcquaid:
@jawshooah Yes, they would. `head-only` provides a fairly shoddy user experience in Homebrew currently, unfortunately, and `audit` checks for that for that reason.

jawshooah:
Makes sense. It's probably a good convention to follow anyway, so that a user can confidently `brew install` any formulae in a tap not explicitly labeled as head/devel-only.

Shall I close the other PR, or would you like to wait for input from other maintainers?

mikemcquaid:
@jawshooah Thanks for understanding. Let's wait for input for now. Long-term I'd love to see HEAD install doing something smart using `git describe` or the Subversion revision number to calculate a `123-HEAD` version or something.

DomT4:
A while back I threw together a rough idea of storing the `.git` or `.svn` folders for the `HEAD` clone in the formula's prefix, and then running `git pull` or similar with each `brew upgrade`, but I suspected from having a bunch of stuff locally I pull at the same time the performance impact would have sucked so I dropped it.

Theoretically we could handle the performance hit by making people specify they wanted to update/upgrade their installed `HEAD` formulae :octopus:.

jawshooah:
Yeah, this issue has been discussed at length in #13197. The problems with `HEAD` upgrades haven't changed, unfortunately.

rkrug:
I just committed the required changes. Let me know if you would prefer a clean pull request with only one commit.

PR #40677, opened 2015-06-14
objective-caml (update, resolve #40676, simplify)

tseemann:
Quoting issues. https://github.com/Homebrew/homebrew/issues/40676

PR #41294, opened 2015-07-02
Ensure that OCaml is compiled with `/usr/local` as its prefix

avsm:
The compiler has a hardbaked location of its standard library that
can be queried with `ocamlc -where`.  For it to work correctly
with other formula like `camlp4`, it must not point to the Cellar.

This partially reverts c24d78416587da42ddba9f51006b0c0e9a086038
from #40677 (but should preserve the actual bug fix there)

### DIFF
--- a/Library/Formula/objective-caml.rb
+++ b/Library/Formula/objective-caml.rb
@@ -27,7 +27,7 @@ class ObjectiveCaml < Formula
     system "./configure", *args
 
     system "make", "world.opt"
-    system "make", "install"
+    system "make", "install", "PREFIX=#{prefix}"
   end
 
   test do
--- a/Library/Formula/objective-caml.rb
+++ b/Library/Formula/objective-caml.rb
@@ -32,5 +32,6 @@ class ObjectiveCaml < Formula
 
   test do
     assert_match "val x : int = 1", shell_output("echo 'let x = 1 ;;' | ocaml 2>&1")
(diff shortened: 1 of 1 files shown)


xu-cheng:
You have to make sure the actual files are installed into the cellar. Maybe add `system "make", "install", "PREFIX=#{prefix}"`.

avsm:
Installed into the Cellar in 1f8bd45187e2e5e6ab918cf60fc31af27ce9a6d4

Users will need to force a brew link now on upgrade, but that cant be helped I think.

mikemcquaid:
> Users will need to force a brew link now on upgrade, but that cant be helped I think.

Why is that, out of interest? Would be good to avoid it if at all possible.

avsm:
> Why is that, out of interest? Would be good to avoid it if at all possible

I think the ocaml package from a few changesets ago used to install directly into `/usr/local`, and so they will have stray files lying around.  This pull request does things the right way.  I may be wrong though -- I had some files directly installed in `/usr/local`, but they could have been development artefacts.

Either way, this PR now does the right thing and works with OPAM, so merging it sooner rather than later would be great so that our CI systems aren't all broken on OSX :)

mikemcquaid:
Ok, cool, thanks. If we see issues we can always have the formula delete those files.

mikemcquaid:

PR #41296, opened 2015-07-02
FORD 3.0.2 (new formula)

zbeekman:
Adds a new formula, created with the help of [homebrew-pypi-poet](https://github.com/tdsmith/homebrew-pypi-poet), for the excellent [FORD](https://github.com/cmacmackin/ford) program.

Ford is a documentation generator for *modern* Fortran programs, *a-la* Doxygen or ROBODoc, except that it understands modern Fortran, uses markdown, and produces [beautiful output](http://jacobwilliams.github.io/bspline-fortran/). It is likely that some newer Fortran projects may need/want this as a recommended build dependency.

```ruby
depends_on "FORD" => [:recommended, :build]
```

At least one project on Homebrew is moving to adopt FORD, [json-fortran](https://github.com/Homebrew/homebrew/blob/master/Library%2FFormula%2Fjson-fortran.rb), as can be seen here: jacobwilliams/json-fortran/issues/114

### DIFF
--- a/Library/Formula/ford.rb
+++ b/Library/Formula/ford.rb
@@ -1,0 +1,131 @@
+class Ford < Formula
+  desc "Automatic documentation generator for modern Fortran programs"
+  homepage "https://github.com/cmacmackin/ford/"
+  url "https://pypi.python.org/packages/source/F/FORD/FORD-3.0.2.tar.gz"
+  sha256 "30844536c73f51fcb078a090741349a3d13de708a3f3f7a6fa722532a97f3268"
+
+  head "https://github.com/cmacmackin/ford.git"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  resource "toposort" do
+    url "https://pypi.python.org/packages/source/t/toposort/toposort-1.4.tar.gz"
+    sha256 "c190b9d9a9e53ae2835f4d524130147af601fbd63677d19381c65067a80fa903"
+  end
+
+  resource "Markdown" do
+    url "https://pypi.python.org/packages/source/M/Markdown/Markdown-2.6.2.tar.gz"
+    sha256 "ee17d0d7dc091e645dd48302a2e21301cc68f188505c2069d8635f94554170bf"
+  end
+
+  resource "beautifulsoup4" do
+    url "https://pypi.python.org/packages/source/b/beautifulsoup4/beautifulsoup4-4.3.2.tar.gz"
+    sha256 "a2b29bd048ca2fe54a046b29770964738872a9747003a371344a93eedf7ad58e"
+  end
+
+  resource "markdown-include" do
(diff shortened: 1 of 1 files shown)


Library/Formula/ford.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir[libexec/"bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    system "#{bin}/ford", "-V"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

DomT4: Any update on the potential for a better test here?

zbeekman: @mikemcquaid and @DomT4 sorry for the slow reply, busy travel weekend for me…

I have a question: it seems that the the `HEAD` is installed then my test fails because somehow `’#{bin}/ford’` isn’t being found. HEAD is virtually identical to 3.0.2 so this has more to do with `#{bin}` I suspect. Should I worry about that, and is there a better way to specify the path to ford?

To address your concerns about the test: I could create a simple Fortran program with embedded documentation and then run FORD on that as the test. What do you think of that?

DomT4: Are you passing `--HEAD` to the `brew test` command?

DomT4: > I could create a simple Fortran program with embedded documentation and then run FORD on that as the test. What do you think of that?

Sounds good. We'd need to avoid compiling the Fortran program if possible, but if you can avoid that :+1:. Anything that shows functionality is good.

zbeekman:
OK, sounds good. I’m not going to actually look at or do anything with the output though since that would be a pain to verify its correctness and is beyond the scope of a Homebrew formula in my opinion.

DomT4:
If it produces a file output we can check it exists with `assert File.exist?("test.test")` and it it outputs a textual output we can do something like `assert_match /test/, shell_output("#{bin}/ford test")` and so on. Fairly short and sweet.

zbeekman:
yeah I’ll check for existence of output files.

zbeekman:
I’ve added a more elaborate test that takes the writes a small FORD project file taken almost word for word from [the example in the FORD repository](https://raw.githubusercontent.com/cmacmackin/ford/master/example-project-file.md), as well as a small Fortran program with FORD documentation markup.

mikemcquaid:

PR #41298, opened 2015-07-02
Add support to luarocks to luajit, code adapted from lua formula

archiloque:


DomT4:
You can't use this style without creating a conflict with the `lua` formula. You'll need to adapt it more fully. Last time I checked locally it refused to take `--lua-suffix=jit` as a valid argument.

archiloque:
- what should I adapt, renaming the execs to luarocksjit ?
- I tried lua-suffix locally and it worked, you can see it used in luarock's  configure file

DomT4:
> what should I adapt, renaming the execs to luarocksjit 

Yes. Something like this.

> I tried lua-suffix locally and it worked, you can see it used in luarock's configure file

Sorry, I meant the `"--lua-version=5.2"` argument wouldn't take anything aside a number, but the problem is that we can't give it a real number because `5.1`, `5.2` and `5.3` are already in use and if we re-use them it'll introduce conflict.

archiloque:
How is used LUA_VERSION beside the configuration ? I don't know enough about luarocks' internals to understand where a conflict may appear.

xpol:
It would be wonderful if luarocks install with luajit.

garthk:
@xpol et al, you can use `luarocks-5.1` from `lua51` with `luajit`:

```
brew install lua51
eval `luarocks-5.1 path --bin`
luarocks-5.1 install socket
luajit -r 'require "socket"; print(socket)'
```

bfontaine:
Ping?

apjanke:
Closing this due to inactivity/lack of response.

PR #41305, opened 2015-07-02
go: add head support for go 1.5

### LINKED ISSUE #41304: go --HEAD does not work because of changes in go 1.5
This is a ticket for discussions on how best to handle building go next. The --head option is a good place to get started on working on the kinks.

I just stepped through this process on my machine, so hopefully I can help. Go will look for an environment variable where it expects to find a working Go >=1.4 toolchain. It will then build Go 1.5 from this and install `go` `godoc` to `$GOPATH/bin`. This is the Go 1.5 binary and what needs to be provided to the end user. The `GOROOT_BOOTSTRAP` environment variable is used to locate Go.

drewwells:
Signed-off-by: Drew Wells <drew.wells00@gmail.com>

fixes #41304

Please see discussion in #41304 for comments on how to do this.

### DIFF
--- a/Library/Formula/go.rb
+++ b/Library/Formula/go.rb
@@ -8,7 +8,12 @@ class Go < Formula
   sha1 "460caac03379f746c473814a65223397e9c9a2f6"
   version "1.4.2"
 
-  head "https://github.com/golang/go", :using => :git
+  head "https://github.com/golang/go.git"
+
+  devel do
+    url "https://github.com/golang/go.git", :revision => "72970f7f52b1f8d14f10df40098175763a556ebe"
+    version "72970f7f52b1f8d14f10df40098175763a556ebe"
+  end
 
   bottle do
     revision 1
@@ -65,7 +70,7 @@ class Go < Formula
       targets = [["darwin", [""]]]
     end
 
-    if build.head?
+    if build.head? || build.devel?
       # GOROOT_FINAL must be overidden later on real Go install
       ENV["GOROOT_FINAL"] = buildpath/"gobootstrap"
 
@@ -82,7 +87,7 @@ class Go < Formula
 
     # The version check is due to:
     # http://codereview.appspot.com/5654068
-    (buildpath/"VERSION").write("default") if build.head?
+    (buildpath/"VERSION").write("default") if build.head? || build.devel?
 
     cd "src" do
       targets.each do |os, archs|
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Might want to use a binary here rather than build-from-source (like we do with `ghc`).

Library/Formula/go.rb
```diff
@@ -8,7 +8,7 @@ class Go < Formula
   sha1 "460caac03379f746c473814a65223397e9c9a2f6"
   version "1.4.2"
 
-  head "https://go.googlesource.com/go", :using => :git
+  head "https://github.com/golang/go", :using => :git
```
dunn: change the url to `https://github.com/golang/go.git` and you'll be able to drop the `:using => :git`

drewwells:
We do not use the Go binaries provided by Google. Unless we are okay to change that decision.

If we did use provided binaries, what's the test matrix? 64bit, 32bit, lion-yosemite?

mikemcquaid:
It'll run tests on 10.8 and above. This approach makes sense to me. Would be great if you could add a beta snapshot as a `devel do`, though, so we can throw CI at this. Thanks!

drewwells:
Oh, I thought test ran against head. I'll add devel.

drewwells:
There isn't a release branch for 1.5 yet. I've locked to the current master, so you can do tests. It'll need updating whenever the RC comes. 1.5 is supposed to be released in August, so RC should be coming soon...

mikemcquaid:

rasky:
After running `brew install go --devel`, I found out that the bootstrap compiler (a binary version of go 1.4.2, downloaded during formula installation) is kept in `/usr/local/Cellar/go/1.5beta2/libexec/gobootstrap`. That compiler shouldn't be needed after Go 1.5 is compiled. @drewwells

drewwells:
Bootstrap tool chain wouldn't be required either. Feel free to pr cleaning
those up!

DomT4:
^ #41912

n054:
Just installed bootstrap yesterday. Will uninstall asap. Thank,nona

drewwells:
Looks like PR was cherry picked instead of merged, deleted branch.

bfontaine:
@drewwells Yep, that’s how we work on Homebrew.

rasky:
Thanks! Since we're at it, go 1.5 doesn't require --with-cc-* anymore, since the whole runtime is recompiled transparently (lazy) as soon as a target architecture is selected (see http://dave.cheney.net/2015/03/03/cross-compilation-just-got-a-whole-lot-better-in-go-1-5).

PR #41306, opened 2015-07-02
Do not fail when xcode-select is missing

sorin-ionescu:
When xcode-select is missing, Homebrew fails even though a compiler is installed.

### DIFF
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -37,7 +37,8 @@ module OS
     end
 
     def active_developer_dir
-      @active_developer_dir ||= Utils.popen_read("/usr/bin/xcode-select", "-print-path").strip
+      @active_developer_dir ||= which("xcode-select") ?
+        Utils.popen_read("/usr/bin/xcode-select", "-print-path").strip : ""
     end
 
     def sdk_path(v = version)


mistydemeo:
Under what circumstances can `xcode-select` be missing?

DomT4:
Why would `xcode-select` *not* be present?

sorin-ionescu:
The `/usr/bin/xcode-select` installed by the Developer Tools CLI package is a shim pointing to the actual `xcode-select` shipped with Xcode, which does not exist if Xcode.app is not installed.

One might be tempted to set `xcode-select -switch /` since the Developer Tools CLI are installed in `/usr/`, but this will fail causing an infinite loop since `xcode-select` keeps calling itself. Setting it to something else will also fail since it won't find the compiler. The best approach is to `chmod ugo-x /usr/bin/xcode-select`, and Homebrew will just work.

`linuxbrew` does [something similar]( https://github.com/Homebrew/linuxbrew/blob/master/Library/Homebrew/os/mac.rb#L40) to make Homebrew work on LINUX.

tdsmith:
The correct Xcode developer directory path for the Xcode CLT is /Library/Developer/CommandLineTools. xcode-select is functional with the CLT installed.

Before CLT installation:

> Vagrants-Mac:~ vagrant$ xcode-select -p
xcode-select: error: unable to get active developer directory, use `xcode-select --switch` to set one (or see `man xcode-select`)

After CLT installation:

>Vagrants-Mac:~ vagrant$ xcode-select -p
/Library/Developer/CommandLineTools

sorin-ionescu:
@tdsmith On older systems, it was installed in `/usr`. Homebrew works just fine without `xcode-select`.

mistydemeo:
`xcode-select` comes with the OS, and it's capable of functioning (e.g. reporting the path, if present) even if Xcode isn't installed. Prior to 10.9 it was a shell script that could respond to `-switch` and `-print-path` even if there wasn't a "real" `xcode-select` binary installed from Xcode or the CLT, and in 10.9 onward it's a real `xcode-select` binary (which doesn't have the infinite loop bug).

sorin-ionescu:
Right, but Homebrew supports earlier OS X versions. This change makes it
work on OS X 10.7 and 10.8, which have that broken xcode-select, provided
the user is willing to disable it. linuxbrew also needed this change.

mistydemeo:
Homebrew works fine on CLT-only OS X 10.7 and 10.8 systems, however. Can you specify what the issue you're seeing is?

sorin-ionescu:
There were some formulae that failed to build because xcode-select was pointing to an Xcode.app that no longer exists. Setting the path to `/` failed due to the infinite loop bug. Setting the path to nothing also failed. Disabling xcode-select and changing that line in Homebrew made it work.

On this 10.7 machine, I disabled the following shims since calling any of them with the path set to `/` will cause an infinite loop. They all deal with Xcode, yet not Xcode is installed; so, they are not needed.

```
/usr/bin/xcode-select.disabled
/usr/bin/xcodebuild.disabled
/usr/bin/xcrun.disabled
/usr/bin/xed.disabled
```

mikemcquaid:
Sorry, we don't support that configuration.

mistydemeo:
Can you specify what failed when you set the path to nothing? Our supported config for 10.7/10.8 CLT-only is to unset the `xcode-select` path. If there's an issue there I'd like to get it fixed, though I don't think anything has changed there in the past few years.

PR #41314, opened 2015-07-03
pygobject3: bottle against system Python

tdsmith:
Take advantage of the lack of explicit Python linkage.


mikemcquaid:
:+1:

tdsmith:
The gst-python test failure seems to happen even without this PR; should figure out what's going on there.

PR #41315, opened 2015-07-03
ntl 9.2.0

nijikon:


Library/Formula/ntl.rb
```diff
@@ -1,16 +1,13 @@
...
-  url "http://www.shoup.net/ntl/ntl-6.2.1.tar.gz"
-  sha1 "3b9ab3bedb0b2e9b5ee322d60745be5caf1c743f"
+  url "http://www.shoup.net/ntl/ntl-9.2.0.tar.gz"
+  sha256 "2bca70534b133904dd814a1d2a111f787a7506fa6f08f2e7f3fc9288b3bfd101"
 
   depends_on "gmp" => :optional
 
   bottle do
     cellar :any
-    revision 1
```
mikemcquaid: You can leave these as-is; it's just those outside the `bottle do` we need removed.

nijikon: Clear, thanks!

mikemcquaid:
Thanks @nijikon!

PR #41290, opened 2015-07-02
git-lfs 0.5.2

gurugray:
updated version section and shasum fo source files
removed bottle section for future correct shasums

PR #41316, opened 2015-07-03
git-lfs 0.5.2

mikemcquaid:
Closes #41290.


mikemcquaid:
I've also created a PR in https://github.com/github/git-lfs/pull/458 so these new steps shouldn't be needed from 0.5.3.

PR #41322, opened 2015-07-03
qbs 1.4.1

jakepetroules:
The Java dependency is a temporary workaround, we need this in order to allow Qbs to build Java source code. It will disappear entirely in a future version when this is implemented in a more transparent fashion.

### DIFF
--- a/Library/Formula/qbs.rb
+++ b/Library/Formula/qbs.rb
@@ -3,8 +3,8 @@ require "formula"
 class Qbs < Formula
   desc "Qt Build Suite"
   homepage "https://wiki.qt.io/Qt_Build_Suite"
-  url "https://download.qt.io/official_releases/qbs/1.4.0/qbs-1.4.0.src.tar.gz"
-  sha1 "4c70247155281b9536a6fab6672cd5f53610cfa1"
+  url "https://download.qt.io/official_releases/qbs/1.4.1/qbs-src-1.4.1.tar.gz"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Can this be made optional?

jakepetroules:
Technically, but I would really rather not.

mikemcquaid:
Java isn't installed by default on OS X so I'd rather we made this `:optional`, thanks. When I was a Qt developer and using QBS in the past I personally never had any need of the Java support.

jakepetroules:
It looks like 'recommended' would be more appropriate in this case, and how do I specify Java 1.6+ recommended?

mikemcquaid:
I don't think it would be, I'm afraid, because of the hassle to install Java on OS X.

Library/Formula/qbs.rb
```diff
@@ -13,9 +13,14 @@ class Qbs < Formula
   end
 
   depends_on "qt5"
+  depends_on :java => :recommended
 
   def install
-    system "qmake", "qbs.pro", "-r"
+    args = []
+    if build.with? "java"
+      args << "CONFIG+=qbs_enable_java"
```
mikemcquaid: `args << "CONFIG+=qbs_enable_java" if build.with? "java"`

jakepetroules:
Fine, I'll concede since this is only going to be a short term presence.

mikemcquaid:
Thanks!

mikemcquaid:

PR #41327, opened 2015-07-03
add env for bottle source

justmao945:
Hi, I want to use another mirror for bottles, but I don't know where to config it. So I add two env vars to overwrite the default bottle mirror `bintray.com`

```
HOMEBREW_BOTTLE_DOMAIN
HOMEBREW_BOTTLE_ROOT_URL
```

Any comments are welcome, thank you!


mikemcquaid:
If you're building your own bottles you should use `brew bottle --root-url` to generate the correct URL. If you're not doing that, can you elaborate more on what you're doing?

PR #41328, opened 2015-07-03
add env var to config bottle source

justmao945:
The previous patch #41327 has an empty file `source.rb`, so I make a new one.

### DIFF
--- a/Library/Homebrew/cmd/--env.rb
+++ b/Library/Homebrew/cmd/--env.rb
@@ -27,7 +27,8 @@ module Homebrew
       HOMEBREW_SVN HOMEBREW_GIT
       HOMEBREW_SDKROOT HOMEBREW_BUILD_FROM_SOURCE
       MAKE GIT CPP
-      ACLOCAL_PATH PATH CPATH].select { |key| env.key?(key) }
+      ACLOCAL_PATH PATH CPATH
+      HOMEBREW_BOTTLE_DOMAIN HOMEBREW_BOTTLE_ROOT_URL].select { |key| env.key?(key) }
   end
 
   def dump_build_env env, f=$stdout
--- a/Library/Homebrew/software_spec.rb
+++ b/Library/Homebrew/software_spec.rb
@@ -258,8 +258,8 @@ end
 class BottleSpecification
   DEFAULT_PREFIX = "/usr/local".freeze
   DEFAULT_CELLAR = "/usr/local/Cellar".freeze
-  DEFAULT_DOMAIN = "https://homebrew.bintray.com".freeze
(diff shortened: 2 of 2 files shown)


mikemcquaid:
If you're building your own bottles you should use `brew bottle --root-url` to generate the correct URL. If you're not doing that, can you elaborate more on what you're doing?

Also, please push to pull request branches rather than creating new ones in future, thanks.

justmao945:
> If you're building your own bottles you should use brew bottle --root-url to generate the correct URL. If you're not doing that, can you elaborate more on what you're doing?

Yeah, I didn't mean to build bottle with another root url. I'm Chinese, and I found it's slow to download bottles from bintray.com in China. Now I have a chance to make another mirror for bottles like bintray.com, may call `homebrew.qiniu.com`. So I want users in China can have a way to set the bottle source URL to `homebrew.qiniu.com` to speed up the installation. I don't know if there's a better way to do this, anyway I made one for user to use another bottles mirror easily.

So how ?


OK, I'm sorry that I want make a clean pull request.  :smile:

justmao945:
Thank you for your quick reply ! :smile:

mikemcquaid:
Hmm; perhaps it's worth contacting Bintray about this? I guess as long as the SHA256 is the same I don't really have an objection but I'd like to hear other @Homebrew/owners thoughts

justmao945:
Bintray uses cloudfront, which is from Amazon, and the service isn't good, sometimes may be banned by the GFW in China. I do think it's great to create a mirror for Chinese in China.

DomT4:
We've had a few issues about this problem, particularly around China. My medium-term thought was to add support for routing our `fetch` through `Tor` or any other SOCKS4/5 proxy. 

It's an option already supported by `cURL` so wouldn't need a huge amount of dancing on our part. It wouldn't particularly provide anonymity either way, but it could speed up downloads if they're being artificially limited by third-party intervention.

justmao945:
Proxy may be a good way for personal usage: setup a proxy -> fetch packages through proxy -> speed up ? But I think to set the source url is a more general way for all users, especially for who is slow to download from Bintray. e.g. npm, gem,  and pip are all able to set up mirror source, so why not ?

xu-cheng:
Supporting mirror is actually on my TODO list for long time. To support it, we should change the way to handle tap bottles. Instead of `root_url`, we should resolve the binary repo based on tap. Therefore, it will allow users to overwrite `BottleSpecification::DEFAULT_DOMAIN` to make it work for both core formulae and official tap formulae.

justmao945:
@xu-cheng 

> we should resolve the binary repo based on tap.
I'm not familiar with brew tap, can you explain it ?

justmao945:
@xu-cheng  Thank you for your support :smile:

PR #41332, opened 2015-07-03
support downloading bottles through 3rd-party mirror

xu-cheng:
cc @Homebrew/owners @justmao945

### DIFF
--- a/Library/Homebrew/bottles.rb
+++ b/Library/Homebrew/bottles.rb
@@ -67,7 +67,7 @@ class Bintray
   end
 
   def self.repository(tap=nil)
-    return "bottles" if tap.to_s.empty?
+    return "bottles" if tap.nil? || tap == "Homebrew/homebrew" || tap == "mxcl/master"
     "bottles-#{tap.sub(/^homebrew\/(homebrew-)?/i, "")}"
   end
 end
--- a/Library/Homebrew/manpages/brew.1.md
+++ b/Library/Homebrew/manpages/brew.1.md
@@ -540,7 +540,10 @@ can take several different forms:
 
     Set this to force Homebrew to use a particular git binary.
 
-  * HOMEBREW_BROWSER:
+  * HOMEBREW\_BOTTLE\_DOMAIN:
+    If set, instructs Homebrew to use the given URL as a download mirror for bottles.
+
+  * HOMEBREW\_BROWSER:
     If set, uses this setting as the browser when opening project homepages,
     instead of the OS default browser.
 
--- a/Library/Homebrew/cmd/bottle.rb
+++ b/Library/Homebrew/cmd/bottle.rb
@@ -9,7 +9,7 @@ require 'extend/pathname'
 
 BOTTLE_ERB = <<-EOS
   bottle do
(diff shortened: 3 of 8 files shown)


Library/Homebrew/bottles.rb
```diff
@@ -67,7 +67,7 @@ def self.package(formula_name)
   end
 
   def self.repository(tap=nil)
-    return "bottles" if tap.to_s.empty?
+    return "bottles" if tap.nil? || tap == "Homebrew/homebrew" || tap == "mxcl/master"
```
mikemcquaid: Maybe would be good to have a constant for these values somewhere?

Library/Homebrew/cmd/doctor.rb
```diff
@@ -651,6 +651,14 @@ def check_user_curlrc
   end
 end
 
+def check_homebrew_bottle_domain
+  unless (bottle_domain = ENV["HOMEBREW_BOTTLE_DOMAIN"]).nil? then <<-EOS.undent
+    You have set HOMEBREW_BOTTLE_DOMAIN.
+    If you have trouble downloading bottles with Homebrew, then maybe this is the problem.
```
mikemcquaid: `You have set HOMEBREW_BOTTLE_DOMAIN. This may cause issues when download formulae.`

Maybe it's enough for this to just be in `brew config` though?

Library/Homebrew/manpages/brew.1.md
```diff
@@ -540,7 +540,10 @@ can take several different forms:
 
     Set this to force Homebrew to use a particular git binary.
 
-  * HOMEBREW_BROWSER:
+  * HOMEBREW\_BOTTLE\_DOMAIN:
+    If set, instructs Homebrew to use the given url as bottle download mirror.
```
mikemcquaid: `If set, instructs Homebrew to use the given URL as a download mirror for bottles`

mikemcquaid:
Few notes but :+1: from me. Would be good if could provide an official mirror, though.

Library/Homebrew/bottles.rb
xu-cheng: It makes more sense to put `tap.nil? || tap == "Homebrew/homebrew" || tap == "mxcl/master"` as a method.

However, I haven't found a good place to put the method or what's the good name. Part of reason is we have complete different variables all calling tap. It's kinda confusing.

mikemcquaid: Yeh. No worries, don't let that block this.

xu-cheng:
Request final comments before I merge this.

DomT4:
Have some concerns about facilitating unofficial mirrors, especially given those mirrors are likely to be plaintext, but barring someone offering to set up an official local mirror that we can audit/oversee/etc in areas with speed troubles we don't have many alternatives on the table.

xu-cheng:
@DomT4 This PR doesn't concern about providing actual mirror but a general change which offers such feature.

DomT4:
Yes, I know. I was just adding some concern that making it easier to use another mirror may lead to some less-than-ideal situations, but it was a concern rather than a blocking complaint. I'm happy enough to see how it unfolds.

PR #41339, opened 2015-07-03
gtk+: Apply patch from GNOME Bug 557780

dtrebbien:
This allows the Freeciv 2.5.0 gtk2 client to run without an assertion
failure.

See:
- https://bugzilla.gnome.org/show_bug.cgi?id=557780
- Homebrew/homebrew-games#274

### DIFF
--- a/Library/Formula/gtk+.rb
+++ b/Library/Formula/gtk+.rb
@@ -3,7 +3,7 @@ class Gtkx < Formula
   homepage "http://gtk.org/"
   url "https://download.gnome.org/sources/gtk+/2.24/gtk+-2.24.28.tar.xz"
   sha256 "b2c6441e98bc5232e5f9bba6965075dcf580a8726398f7374d39f90b88ed4656"
-  revision 1
+  revision 2
 
   bottle do
     revision 2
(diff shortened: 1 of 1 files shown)


Library/Formula/gtk+.rb
```diff
@@ -27,6 +27,12 @@ class Gtkx < Formula
     cause "Undefined symbols when linking"
   end
 
+  def patches
+    # Patch from Bug 557780 - Missing support for depth == 1 in gdk_image_new_for_depth()
+    # https://bugzilla.gnome.org/show_bug.cgi?id=557780
+    DATA
+  end
```
bfontaine: Use `patch :DATA` instead

dtrebbien: Neat trick!  I didn't know about that.

I have uploaded a new patch with a slightly different URL to avoid a missing diff header issue.

Library/Formula/gtk+.rb
```diff
@@ -27,6 +27,13 @@ class Gtkx < Formula
     cause "Undefined symbols when linking"
   end
 
+  # Patch to allow Freeciv's gtk2 client to run.
+  # See Homebrew/homebrew-games#278
+  patch do
```
mistydemeo: Can you include a link to the upstream bug, please? Since it hasn't been merged yet we'll want to keep an eye on it.

dtrebbien: Done

mistydemeo:
Thanks!

PR #41343, opened 2015-07-04
node.js 2.11.2 and io.js 2.3.3

mistydemeo:


DomT4:
Can we bump the internal `npm` on `Node` here please? `npm` made some changes in the last couple stable releases (2.12.1) which make it a *lot* harder to break `npm` by running it under root.

mistydemeo:
Done!

DomT4:
Thanks Misty!

PR #41344, opened 2015-07-04
Lua tweaks

DomT4:
* Version bump to Lua, and permit users to use their own configuration files again.
* Permit user configuration files for Lua51
* Remove unnecessary pointers in libquvi.

On the configuration files issue, we/I deliberately disabled them for a while because there was a potential for it to not sit well with the integrate Luarocks into Lua changes. On the back of a lack of mayhem on that front, and a request we reconsider the config forcing, I did some local testing and am happy enough reverting that individual change won't cause breakage.

Long term I want to try and do something with the way we handle Lua resources, similar to how we already handle Python resources, but baby-steps.

### DIFF
--- a/Library/Formula/libquvi.rb
+++ b/Library/Formula/libquvi.rb
@@ -1,10 +1,8 @@
-require 'formula'
-
 class Libquvi < Formula
   desc "C library to parse flash media stream properties"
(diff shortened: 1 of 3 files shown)


DomT4:
Failure was a test, which is a known issue and being worked on in another PR. Pulling.

PR #41349, opened 2015-07-04
Update Harfbuzz dependencies (optional graphite2)

simoncozens:
Now that graphite2 is available in Homebrew, allow Harfbuzz to use it for shaping if the user wants it.

### DIFF
--- a/Library/Formula/harfbuzz.rb
+++ b/Library/Formula/harfbuzz.rb
@@ -12,7 +12,7 @@ class Harfbuzz < Formula
 
   depends_on "pkg-config" => :build
   depends_on "glib"
-  depends_on "cairo" => :optional
+  depends_on "cairo"
   depends_on "icu4c" => :recommended
   depends_on "graphite2" => :optional
   depends_on "freetype"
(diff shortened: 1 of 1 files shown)


Library/Formula/harfbuzz.rb
```diff
@@ -12,8 +12,9 @@ class Harfbuzz < Formula
 
   depends_on "pkg-config" => :build
   depends_on "glib"
-  depends_on "cairo"
+  depends_on "cairo" => :optional
```
DomT4: Would prefer to keep this `:recommended` for now. Not a fan of making potentially breaking changes for end users without some revision or version bump.

simoncozens:
OK, sticking to one issue at a time - this PR just deals with harfbuzz => graphite2.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/19c803dd08fc6037dfcb932a770480e765226559. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a nit, but commit messages for non-version changes should be styled something like `harfbuzz: add optional graphite2 dependency.`, Thanks!

PR #41352, opened 2015-07-04
graphite2: use https resource and homepage

vszakats:
* homepage is redirected back to http


DomT4:
:rocket:, Thanks @vszakats!

PR #41353, opened 2015-07-04
czmq: rename executable

DomT4:
The curse of generic naming strikes again.

```
Could not symlink bin/makecert
Target /usr/local/bin/makecert
is a symlink belonging to czmq. You can unlink it:
  brew unlink czmq

  brew link --overwrite mono

  brew link --overwrite --dry-run mono

/usr/local/bin/makecert -> /usr/local/Cellar/czmq/3.0.2/bin/makecert
```

Since czmq tends to be used as a library more than for cert generation, let's rename this one.

CC @Homebrew/owners

### DIFF
--- a/Library/Formula/czmq.rb
+++ b/Library/Formula/czmq.rb
@@ -3,6 +3,7 @@ class Czmq < Formula
   homepage "http://czmq.zeromq.org/"
   url "http://download.zeromq.org/czmq-3.0.2.tar.gz"
   sha256 "8bca39ab69375fa4e981daf87b3feae85384d5b40cef6adbe9d5eb063357699a"
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


tdsmith:
I grepped through the source tree and this seems reasonable; it's hardly mentioned in the docs. There may also be a man page to rename.

I guess another option would be installing it to `share/czmq`; I don't care which.

DomT4:
It seems either we're not building the manpages or they aren't installed automatically.

```
 ~> brew ls czmq
/usr/local/Cellar/czmq/3.0.2/bin/makecert
/usr/local/Cellar/czmq/3.0.2/include/ (43 files)
/usr/local/Cellar/czmq/3.0.2/lib/libczmq.3.dylib
/usr/local/Cellar/czmq/3.0.2/lib/pkgconfig/libczmq.pc
/usr/local/Cellar/czmq/3.0.2/lib/ (2 other files)
```

PR #40179, opened 2015-05-29
protobuf 3.0.0-alpha-3

peter-edge:
This updates the protobuf --devel formula for the new alpha released
yesterday at https://github.com/google/protobuf/releases/tag/v3.0.0-alpha-3
and fixes the proto3 test which was broken:

```
test.proto:4:12: Explicit 'optional' labels are disallowed in the Proto3
syntax. To define 'optional' fields in Proto3, simply remove the
'optional' label, as fields are 'optional' by default.
```

Optional fields should not be specified in proto3.

Signed-off-by: Peter Edge <peter.edge@gmail.com>

PR #38180, opened 2015-03-29
szl: bump revision

tdsmith:
Mountain Lion bottle went skunky, from build failure in #38171.

PR #41354, opened 2015-07-04
protobuf 3.0.0-alpha-3.1 (devel)

dustinwilson:
Devel version bump to 3.00 alpha 3.1.

Also added a patch for stable to fix the unexpected identifier error when compiling software against protobuf. Google's stated they won't fix it in stable (it is fixed in the 3.0 alphas): https://github.com/google/protobuf/issues/549

### DIFF
--- a/Library/Formula/protobuf.rb
+++ b/Library/Formula/protobuf.rb
@@ -1,13 +1,20 @@
 class Protobuf < Formula
   desc "Protocol buffers (Google's data interchange format)"
   homepage "https://github.com/google/protobuf/"
-  url 'https://github.com/google/protobuf/releases/download/v2.6.1/protobuf-2.6.1.tar.bz2'
-  sha1 '6421ee86d8fb4e39f21f56991daa892a3e8d314b'
+
+  stable do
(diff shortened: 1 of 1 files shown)


DomT4:
We don't really use `stable` blocks purely for patching any more. Should be:

```ruby
stable do
  url "https://github.com/google/protobuf/releases/download/v2.6.1/protobuf-2.6.1.tar.bz2"
  sha256 "ee445612d544d885ae240ffbcbf9267faa9f593b7b101f21d58beceb92661910"

  # Fixes the unexpected identifier error when compiling software against protobuf:
  # https://github.com/google/protobuf/issues/549
  patch :p1, :DATA
end
```

dustinwilson:
Sorry about that. Should be fixed now. :)

DomT4:
Ignoring the test-bot failure here on ML. It's the same failure we've seen before in #38180 and #40179. Looks like a long-term false flag.

Merged in https://github.com/Homebrew/homebrew/commit/52de0653797395693524ec79d439256d267a51e8. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #41362, opened 2015-07-04
log4shib 1.0.9

dunn:


Library/Formula/log4shib.rb
```diff
@@ -15,6 +13,11 @@ class Log4shib < Formula
 
   def install
     system "./configure", "--prefix=#{prefix}", "--disable-debug", "--disable-dependency-tracking"
-    system "make install"
+    system "make", "install"
+  end
+
+  test do
+    assert_equal "-L#{HOMEBREW_PREFIX}/Cellar/log4shib/1.0.9/lib -llog4shib",
+                 shell_output("#{bin}/log4shib-config --libs").chomp
```
bfontaine: :+1:

bfontaine:
Thank you @dunn! :100:

Library/Formula/log4shib.rb
DomT4: Just for future reference, we prefer to use `#{Formula["log4shib"].opt_lib}` for this sort of thing because it means we don't have to bump the version number on the test every update (revision or version) to get it to pass.

dunn: ohhh, gotcha

PR #41364, opened 2015-07-05
rocket: migrating to boneyard

bfontaine:
See #40392.

Should we really keep this formula in the core? It’s broken on OS X and upstream has no plan supporting Darwin for the moment. cc @sjackman 
I’m bumping this to 0.5.1 only (the latest is 0.6.1) because the later versions don’t even compile on OS X (the build depends on Linux-only files).


DomT4:
> because the later versions don’t even compile on OS X (the build depends on Linux-only files).

Given upstream has said they aren't supporting Darwin now I'd just boneyard it. We can't support Linux-only formulae in the core.

sjackman:
I'm fine with boneyard.

bfontaine:
I’m boneyarding it https://github.com/Homebrew/homebrew-boneyard/pull/59

PR #41366, opened 2015-07-05
qi: migrating to homebrew/boneyard

bfontaine:
See https://github.com/Homebrew/homebrew-boneyard/pull/58 and #40392.


mikemcquaid:
:+1:

PR #41252, opened 2015-07-01
qt5 5.5.0

beemaster:

PR #41367, opened 2015-07-05
qt5 5.5.0 

DomT4:
While the bot is quiet I wanted to test whether the `rpath` changes were causing the build failures on the test-bot, and integrate Secure Transport/No OpenSSL. Commit history from #41252 preserved.

### DIFF
--- a/Library/Formula/qt5.rb
+++ b/Library/Formula/qt5.rb
@@ -12,23 +12,25 @@ end
 class Qt5 < Formula
   desc "Version 5 of the Qt framework"
   homepage "https://www.qt.io/"
-  url "https://download.qt.io/official_releases/qt/5.5/5.5.0/single/qt-everywhere-opensource-src-5.5.0.tar.xz"
-  mirror "https://www.mirrorservice.org/sites/download.qt-project.org/official_releases/qt/5.5/5.5.0/single/qt-everywhere-opensource-src-5.5.0.tar.xz"
-  sha256 "7ea2a16ecb8088e67db86b0835b887d5316121aeef9565d5d19be3d539a2c2af"
+  head "https://code.qt.io/qt/qt5.git", :branch => "5.5", :shallow => false
 
-  # Apple's 3.6.0svn based clang doesn't support -Winconsistent-missing-override
-  # https://bugreports.qt.io/browse/QTBUG-46833
-  # This is fixed in 5.5 branch and below patch should be removed
-  # when this formula is updated to 5.5.1
-  patch :DATA
+  stable do
+    url "https://download.qt.io/official_releases/qt/5.5/5.5.0/single/qt-everywhere-opensource-src-5.5.0.tar.xz"
+    mirror "https://www.mirrorservice.org/sites/download.qt-project.org/official_releases/qt/5.5/5.5.0/single/qt-everywhere-opensource-src-5.5.0.tar.xz"
+    sha256 "7ea2a16ecb8088e67db86b0835b887d5316121aeef9565d5d19be3d539a2c2af"
+
+    # Apple's 3.6.0svn based clang doesn't support -Winconsistent-missing-override
+    # https://bugreports.qt.io/browse/QTBUG-46833
+    # This is fixed in 5.5 branch and below patch should be removed
+    # when this formula is updated to 5.5.1
+    patch :DATA
+  end
 
   bottle do
     sha256 "1d3aee1664b44e912ddd307fc7f1eff25e835452ce44705acaa4162f79006ef7" => :yosemite
     sha256 "f32d4dde1b09d619e5046b9e5717ab48d7dc6b066b09bbde8d44f74b2ef040fb" => :mavericks
     sha256 "855e075b522199c52876f44fe2d2a63e4c4b4f9bfd5c6edb0e3dc850fd02ef34" => :mountain_lion
   end
-
-  head "https://code.qt.io/qt/qt5.git", :branch => "5.5", :shallow => false
 
   keg_only "Qt 5 conflicts Qt 4 (which is currently much more widely used)."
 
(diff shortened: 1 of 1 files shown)


Library/Formula/qt5.rb
```diff
@@ -43,20 +51,15 @@ class Qt5 < Formula
...
   depends_on OracleHomeVarRequirement if build.with? "oci"
 
   def install
     ENV.universal_binary if build.universal?
 
     args = ["-prefix", prefix,
-            "-system-zlib",
+            "-system-zlib", "-securetransport",
             "-qt-libpng", "-qt-libjpeg",
+            "-no-rpath", "-no-openssl",
```
mistydemeo: Are we breaking the rpath install names via our install name rewriting? Is that something we can fix?

DomT4: I'm not completely sure what's going on here. `rpath` [apparently became](https://github.com/Homebrew/homebrew/pull/41252#issuecomment-118557153) the default for Qt5 in this release, and the bot errors imply (Although perhaps not correctly) that the `rpath` setting wasn't linking as expected for some reason.

I did a full build locally running with the same `--build-bottle` variables as the test-bot and both the installation and bottle turned out fine, which makes me wonder if I'm getting lucky or if something about the test-bot is borking the installation in some way. We disable rpath usage in a few formulae, but it isn't widespread. There are build logs from my local install [here](https://gist.github.com/anonymous/d2047f4c8db4e58adc15) if that's any use at all.

DomT4: [This is everything we're changing the paths on after installation](https://gist.github.com/DomT4/0f962ab0e55264c0f37f).

jakepetroules:
> Are we breaking the rpath install names via our install name rewriting? Is that something we can fix?

@mistydemeo Is this something Homebrew does automatically for all binaries of installed formulae?

DomT4:
Dropping the `rpath` change has allowed everything but Qt's test to pass on Mountain Lion, so perhaps `rpath` usage is the problem:

```
==> brew test --verbose qt5						 FAILED
Testing qt5
==> /usr/local/Cellar/qt5/5.5.0/bin/qmake -project
Error: qt5: failed
execution expired
/usr/local/Library/Homebrew/formula.rb:868:in `gets'
/usr/local/Library/Homebrew/formula.rb:868:in `block in system'
/usr/local/Library/Homebrew/formula.rb:854:in `open'
/usr/local/Library/Homebrew/formula.rb:854:in `system'
/usr/local/Library/Formula/qt5.rb:130:in `block in <class:Qt5>'
/usr/local/Library/Homebrew/formula.rb:797:in `block in run_test'
/usr/local/Library/Homebrew/formula.rb:793:in `run_test'
/usr/local/Library/Homebrew/test.rb:28:in `block in <main>'
/usr/local/Library/Homebrew/test.rb:27:in `<main>'
==> brew fetch --retry pyqt5						 PASSED
==> brew install pyqt5							 PASSED
==> brew test --verbose pyqt5						 PASSED
==> brew fetch --retry qbs						 PASSED
==> brew install qbs							 PASSED
==> brew test --verbose qbs						 PASSED
```

DomT4:
Slightly different error. Our bottles are now doing this:

```
qt5/5.5.0/bin/qdoc
qt5/5.5.0/bin/qhelpconverter
qt5/5.5.0/bin/qhelpgenerator
qt5/5.5.0/bin/qlalr
qt5/5.5.0/bin/qmake
gzip: (stdin): unexpected end of file
gtar: Unexpected EOF in archive
gtar: Unexpected EOF in archive
gtar: Error is not recoverable: exiting now
```

luzpaz:
Thanks for your work on this so far

luzpaz:
checking in to see if there is anything that I can do to help this get merged ?

DomT4:
I want to do another local build to see if `qmake` does actually work, the installation worked last time but I didn't test it properly, and another full local build outside of Homebrew to see if it is our `install_name_tool` changes causing the breakages.

The issue has just been finding ~6 hours of time I don't need to do anything else on the machine for, since Qt essentially sets fire to the CPU during compile :smile_cat:

DomT4:
Alright, from-source outside-of-brew build has been started. Shall see if the end product is usable, and then do the exact same build inside `brew` and see if that is usable, and that should help determine whether our `install_name` changes are the culprit here.

DomT4:
Either the `qmake` bug exists in the from-source, outside-brew build or `qmake -project` now hangs unless there's actually a project. The latter is more likely and would explain the timeout on the `qt5` test itself at least.

Was able to build a `qmake`-using project just fine with the from-source, outside-brew Qt5 build. No issues building or running a project built against it.

Will set an inside-brew build running in a few hours and see if the same functionality exists.

luzpaz:
how's this going so far ?

luzpaz:
@DomT4 how did the inside brew build go ?

luzpaz:
Sorry I know this is annoying but.. 
bump
:disappointed:

DomT4:
Sorry. `qt5` is an absolute killer to build, and I essentially can't touch the laptop whilst it does the 3+ hour build without burning a hole in my lap (105+ degree CPU temps, woo), which is difficult when I have a lot of uni work to do. I'll try and tee it up when I go sleep later and let it build then.

luzpaz:
@DomT4 I know how that feels like since I've built it before and it's an atrocious amount of time. Thanks so much for your work on this so far!

mikemcquaid:
Maybe it's worth trying this without secure transport for now?

DomT4:
Secure Transport doesn't seem to be the breaking variable. It broke before we checked out the Secure Transport option as well. It seems to be `rpath` related, or something we do with our `install_names` after installation, but proving fun to narrow down.

luzpaz:
@DomT4 Can you post gist-logs so we can use the power of the crowd to solve this ?

DomT4:
`gist-logs` are in the earlier diff with Misty. The important `install_name` changes aren't added automatically to those logs, but I put a link to them there as well.

Currently doing a fresh install. Should be done in about ~9 years/4 hours.

DomT4:
New gist-logs [here](https://gist.github.com/anonymous/b987f06d298e299a088c).

`install_name` changes [here](https://gist.github.com/DomT4/016a4ce9ceeee2b45361). The latter is a 2500 line document :whale2:.

DomT4:
`qbs` will build fine against this `qt5` as well. It could be that `qbs` doesn't exercise every last Framework Qt ships, so I don't want us reading too hugely into the will it/won't it compile, but it is a useful baseline.

Am going to fix up Qt5's test, and then set the bot running on this again, and cross all of my fingers.

DomT4:
Alright, build is a *go*. Have fixed the test so it doesn't hang and die horribly. Now I just cross my fingers for the next 4 hours whilst the test-bot does its thing and hope the bottle that pops out of the end is usable this time.

mikemcquaid:
Looks like it built. Hooray.

mikemcquaid:
Merged this to avoid any :ghost:s ruining things.

DomT4:
The bottle will untar this time as well. I don't think I've ever been happier to get a PR passing and behaving as expected. This was a nightmare :sweat_smile:.

luzpaz:
What a trip! Thanks @DomT4 
You Rock! :+1:

PR #41378, opened 2015-07-05
libssh 0.7.1

dunn:


Library/Formula/libssh.rb
```diff
@@ -1,10 +1,10 @@
 class Libssh < Formula
   desc "C library SSHv1/SSHv2 client and server protocols"
   homepage "https://www.libssh.org/"
-  url "https://red.libssh.org/attachments/download/107/libssh-0.6.4.tar.gz"
-  sha1 "073bf53d9e02f7cfbcc5d8738ca1c9ffb2edd247"
+  url "https://red.libssh.org/attachments/download/154/libssh-0.7.1.tar.xz"
+  sha256 "2fc7ccf96d3263cbd8ab520118cb94d9a2e11714c61e22b3f761fc5352fd046d"
 
-  head "git://git.libssh.org/projects/libssh.git"
+  head "http://git.libssh.org/projects/libssh.git"
```
DomT4: This doesn't work for me. Just hangs:

```
git clone http://git.libssh.org/projects/libssh.git
Cloning into 'libssh'...
```

dunn: Same; weird. I didn't check it since it's listed as an option on the [download page](https://www.libssh.org/get-it/). I'll revert.

DomT4:
Their website rather confusingly provides a backup of the 0.6.4 release, which suggests that is latest stable, but MacPorts are carrying the 0.7.x releases so I guess we're fine here.

Merged in https://github.com/Homebrew/homebrew/commit/9ba211a347f879884cb8bff1a718b042d09660eb. Thanks Alex!

PR #41381, opened 2015-07-05
yaggo 1.5.5 (new formula)

dunn:
It's been suggested we move this formula from homebrew-science to core (Homebrew/homebrew-science#2490). If this gets the :+1: I'll update that PR to remove it from the science repo.


dunn:
Test failure is expected:
```
==> audit problems
yaggo:
 * Formula name conflicts with homebrew/science/yaggo
```

dunn:
It's being kept in homebrew-science.

PR #41382, opened 2015-07-05
cloc: 1.64

nijikon:


bfontaine:
Thank you @nijikon! :tada: 

For the record, the preferred commit message format for version bumps is “cloc 1.64” (drop the colon).

PR #41383, opened 2015-07-05
phoon: 04A (14Aug2014)

nijikon:
I did not find any version info on this release so I just bumped it to 04A, not sure if this is ok. Maybe we should version it as 14Aug2014 instead. Thoughts?

### DIFF
--- a/Library/Formula/phoon.rb
+++ b/Library/Formula/phoon.rb
@@ -1,15 +1,13 @@
-require 'formula'
-
 class Phoon < Formula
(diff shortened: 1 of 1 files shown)


DomT4:
> I just bumped it to 04A, not sure if this is ok

Seems fine to me. It's consistent with the previous versioning.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/f4a27681d10dc6cc5559ce015dc142be5fba8aae. Thanks @nijikon!

PR #38600, opened 2015-04-13
mat 0.5.3

DomT4:
Version bump.

PR #41386, opened 2015-07-05
coreutils 8.24

bfontaine:


bfontaine:
The test is failing because of `mat`, which is already causing us some issues on #37280 and #38600.

bfontaine:
As discussed, now that `mat` is gone (45c6a6c), I’m trying to build this without any revision bump. If it fails, I’ll add the revisions back (they’re in [another branch](https://github.com/bfontaine/homebrew/compare/bfontaine:coreutils-8.24...bfontaine:coreutils-8.24-with-revisions) in the meantime).

PR #41387, opened 2015-07-05
quantlib 1.6

dunn:
No Debian mirror yet.

### DIFF
--- a/Library/Formula/quantlib.rb
+++ b/Library/Formula/quantlib.rb
@@ -1,9 +1,8 @@
 class Quantlib < Formula
   desc "Library for quantitative finance"
   homepage "http://quantlib.org/"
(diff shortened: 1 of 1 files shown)


bfontaine:
Shouldn’t we wait for Debian to add a mirror?

dunn:
Fine with me, but there's sometimes a delay, isn't there?

bfontaine:
I guess so, but I don’t know how reactive they are.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/2dc63a770b3b5ebf0928dc68bec85cef1364f844. Thanks again @dunn!

There's little point waiting for Debian to update if they haven't done so relatively quickly. Either they're backporting the version rather than adding a new `orig` base tarball, or they're not going to update. The mirrors tend to be quickly updated otherwise. I added a MacPorts one.

PR #41389, opened 2015-07-05
mat: migrating to boneyard

bfontaine:
https://github.com/Homebrew/homebrew-boneyard/pull/60


Marvin91:
Why mat has been removed from brew?
it work for me

DomT4:
Newer versions don't build as expected/intended, and the existing version has some testing failures.

PR #41390, opened 2015-07-05
jed: configure using --with-slang to set proper path to S-Lang

L2G:
I was unable to build jed in a non-/usr/local installation of Homebrew without giving the configure script the `--with-slang` option to set the path to S-Lang.

### DIFF
--- a/Library/Formula/jed.rb
+++ b/Library/Formula/jed.rb
@@ -26,9 +26,8 @@ class Jed < Formula
         system "make"
       end
     end
-    system "./configure", "--disable-debug",
-                          "--disable-dependency-tracking",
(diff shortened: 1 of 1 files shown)


DomT4:
Does the problem impact existing non `/usr/local` installs? It doesn't need a revision unless it impacts existing set ups.

mikemcquaid:
> It doesn't need a revision unless it impacts existing set ups.

:+1:

mikemcquaid:
@L2G if you remove the revision this should be mergeable, thanks!

L2G:
Got it; just slow getting to this. :smile_cat:

mikemcquaid:

PR #41397, opened 2015-07-06
fetch/build formula from source when modified

### LINKED ISSUE #36068: Suggested feature: Automagically build bottled brews if their formula has been edited
Something I'd like to see and am happy to work up a PR for if'n you all like it;  I'd like to see Homebrew automatically know I don't want to use the bottle if I've edited the formula.

I'm thinking this could be a `git status` or `git diff` done on the formula at install-time, and if it `diff`ers, then prefer to build from scratch rather then use the bottle.  Add in a `--no-srsly-bottle-me` flag for over-riding the edited formula.

dunn:
Closes #36068.

Is performance something to worry about here?

### DIFF
--- a/Library/Homebrew/cmd/fetch.rb
+++ b/Library/Homebrew/cmd/fetch.rb
@@ -32,6 +32,13 @@ module Homebrew
   def fetch_bottle? f
     return true if ARGV.force_bottle? && f.bottle
     return false unless f.bottle && f.pour_bottle?
+    if f.file_modified?
+      filename = f.path.to_s.gsub("#{HOMEBREW_PREFIX}/", "")
+      opoo "Formula file is modified!"
+      puts "Fetching source because #{filename} has local changes"
+      puts "To fetch the bottle instead, run with --force-bottle"
+      return false
+    end
     return false if ARGV.build_from_source? || ARGV.build_bottle?
     return false unless f.bottle.compatible_cellar?
     return true
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -601,6 +601,14 @@ class Formula
     s << ") #{path}>"
   end
 
+  def file_modified?
+    return false unless which("git")
+    path.parent.cd do
+      diff = Utils.popen_read("git", "diff", "origin/master", "--", "#{path}")
+      !diff.empty? && $?.exitstatus == 0
+    end
+  end
+
   # Standard parameters for CMake builds.
   # Setting CMAKE_FIND_FRAMEWORK to "LAST" tells CMake to search for our
   # libraries before trying to utilize Frameworks, many of which will be from
@@ -1187,4 +1195,3 @@ class Formula
     end
   end
 end
-
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -62,7 +62,7 @@ class FormulaInstaller
 
     bottle = formula.bottle
     return true  if force_bottle? && bottle
-    return false if build_from_source? || build_bottle? || interactive?
+    return false if build_from_source? || build_bottle? || interactive? || formula.file_modified?
     return false unless options.empty?
 
     return true if formula.local_bottle_path
@@ -180,6 +180,12 @@ class FormulaInstaller
     build_bottle_preinstall if build_bottle?
 
     unless @poured_bottle
+      if formula.file_modified?
+        filename = formula.path.to_s.gsub("#{HOMEBREW_PREFIX}/", "")
+        opoo "Formula file is modified!"
+        puts "Building from source because #{filename} has local changes"
+        puts "To install from a bottle instead, run with --force-bottle"
+      end
       compute_and_install_dependencies if @pour_failed and not ignore_deps?
       build
       clean
--- a/Library/Homebrew/test/test_bottle_hooks.rb
+++ b/Library/Homebrew/test/test_bottle_hooks.rb
@@ -7,6 +7,7 @@ class BottleHookTests < Homebrew::TestCase
     def bottle; end
     def local_bottle_path; end
     def some_random_method; true; end
+    def file_modified?; end
   end
 
   def setup


dunn:
Oh, forgot to add the override flag @DylanLacey mentioned.

dunn:
Nevermind, `--force-bottle` still works to override it.

dunn:
```
  1) Error:
BottleHookTests#test_has_no_bottle:
NoMethodError: undefined method `is_modified?' for #<BottleHookTests::FormulaDouble:0x007f9203054078>
    /usr/local/Library/Homebrew/formula_installer.rb:65:in `pour_bottle?'
```
This is weird since formula_installer `requires` formula.  I must be missing something.

DylanLacey:
Oh bugger.  This'll teach me to write an implementation on a RailsCamp and forget to send a PR >.>

dunn:
I mean if yours is better I'd be happy to close this in favor of it.

Library/Homebrew/formula.rb
```diff
@@ -595,6 +595,14 @@ def inspect
     s << ") #{path}>"
   end
 
+  def modified?
+    old_wd = Pathname.pwd
+    Dir.chdir path.dirname
+    diff = Utils.popen_read("git", "diff", "--diff-filter=M", "--name-only", "#{path}")
+    Dir.chdir old_wd
```
xu-cheng: You can do:
```ruby
path.parent.cd do
  diff = ...
end
```

Another thing that you need to check is formula file may be outside git repo.

dunn: > Another thing that you need to check is formula file may be outside git repo.

That was why I added the `exitstatus` check; git should return non-zero in those cases. Are there cases where that's insufficient?

DylanLacey:
@Dunn I don't believe so;  I pushed it as https://github.com/Homebrew/homebrew/pull/41433/files.  I just was chastising myself for having and implementing a reasonable idea, then not actually going 100% through forgetfulness and lack of tests :P

tdsmith:
I do think it's important that a solution announces "Formula modified locally; building from source" or similar.

dunn:
Updated; but let me know if this should be closed in favor of @DylanLacey's.

Library/Homebrew/formula.rb
```diff
@@ -595,6 +595,13 @@ def inspect
     s << ") #{path}>"
   end
 
+  def modified?
+    path.parent.cd do
+      diff = Utils.popen_read("git", "diff", "--diff-filter=M", "--name-only", "#{path}")
+      return !diff.empty? && $?.exitstatus == 0
```
xu-cheng: `return` is redundant

Library/Homebrew/formula.rb
```diff
@@ -595,6 +595,13 @@ def inspect
     s << ") #{path}>"
   end
 
+  def modified?
```
xu-cheng: Let's add `return false unless which("git")` to check whether git is available.

mikemcquaid: Let's also call this `file_modified?`

Library/Homebrew/formula_installer.rb
```diff
@@ -62,6 +62,11 @@ def pour_bottle? install_bottle_options={:warn=>false}
 
     bottle = formula.bottle
     return true  if force_bottle? && bottle
+    if formula.modified?
+      ohai "#{formula.full_name} is modified; building from source"
+      ohai "Build from bottle with --force-bottle"
```
mikemcquaid: `Install from a bottle`

Library/Homebrew/cmd/fetch.rb
```diff
@@ -32,6 +32,11 @@ def fetch
   def fetch_bottle? f
     return true if ARGV.force_bottle? && f.bottle
     return false unless f.bottle && f.pour_bottle?
+    if formula.modified?
+      ohai "#{formula.full_name} is modified; fetching source"
+      ohai "Fetch bottle with --force-bottle"
```
mikemcquaid: `Fetch the bottle with`

dunn:
Still need help interpreting the test failure, I'm afraid.

Library/Homebrew/formula.rb
```diff
@@ -595,6 +595,14 @@ def inspect
     s << ") #{path}>"
   end
 
+  def file_modified?
+    return false unless which("git")
+    path.parent.cd do
+      diff = Utils.popen_read("git", "diff", "--diff-filter=M", "--name-only", "#{path}")
+      !diff.empty? && $?.exitstatus == 0
+    end
+  end
```
mikemcquaid: Also add this method here: https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/test/test_bottle_hooks.rb#L6-L10

Library/Homebrew/cmd/fetch.rb
```diff
@@ -32,6 +32,11 @@ def fetch
   def fetch_bottle? f
     return true if ARGV.force_bottle? && f.bottle
     return false unless f.bottle && f.pour_bottle?
+    if formula.file_modified?
+      ohai "#{formula.full_name} is modified; fetching source"
+      ohai "Fetch the bottle with --force-bottle"
```
xu-cheng: This is kinda misleading. It would be better to be `To fetch the bottle, run with --force-bottle`. Also use `puts` for the second `ohai`.

Library/Homebrew/formula_installer.rb
```diff
@@ -62,6 +62,11 @@ def pour_bottle? install_bottle_options={:warn=>false}
 
     bottle = formula.bottle
     return true  if force_bottle? && bottle
+    if formula.file_modified?
+      ohai "#{formula.full_name} is modified; building from source"
+      ohai "Install from a bottle with --force-bottle"
```
xu-cheng: Similar problem as above.

Library/Homebrew/cmd/fetch.rb
dunn: Using a `puts` means we're not restricted by length, so maybe even "To fetch the bottle instead, run with --force-bottle"?

mikemcquaid: That sounds good.

dunn:
OK, I think that covers everything.

dunn:
Not getting those test failures locally.

mikemcquaid:
I currently see this message printed twice (it should only be output called once). Once that's fixed we're good to go.

dunn:
Sorry, where are you seeing doubled output?

Also, added `.rb` to the `ohai`, since it's the file that's been modified.

mikemcquaid:
If I run `brew install wget` after modifying the formula.

dunn:
Oh, I was testing with `brew install when` and it doesn't happen for that formula. Bizarre; I'll investigate.

dunn:
It makes sense that we get the message twice for formula like libtorrent which call `pour_bottle?`, but I'm not sure why wget does it.

dunn:
Anyway, fixed!

Library/Homebrew/cmd/fetch.rb
```diff
@@ -32,6 +32,11 @@ def fetch
   def fetch_bottle? f
     return true if ARGV.force_bottle? && f.bottle
     return false unless f.bottle && f.pour_bottle?
+    if formula.file_modified?
+      ohai "#{formula.full_name}.rb is modified; fetching source"
```
mikemcquaid: `formula.path.to_s.gsub("#{HOMEBREW_PREFIX}/", '')` might be a bit better (particularly for stuff from taps).

Library/Homebrew/formula_installer.rb
```diff
@@ -180,6 +180,10 @@ def install
     build_bottle_preinstall if build_bottle?
 
     unless @poured_bottle
+      if formula.file_modified?
+        ohai "#{formula.full_name}.rb is modified; building from source"
+        puts "To install from a bottle instead, run with --force-bottle"
```
mikemcquaid: I wonder if there's somewhere you can unify the printing of this message in a method somewhere (just so we don't need to manually keep the two strings in sync).

mikemcquaid:
@dunn This is looking awesome. A few more tweaks; almost there!

Library/Homebrew/formula_installer.rb
dunn: They're not quite the same, though. Unless you'd like them to be?

Library/Homebrew/cmd/fetch.rb
dunn: On narrow tmux/screen panes the path will get cut off, but I think that's fine:
screenshot 2015-07-11 10 25 18

mikemcquaid: Maybe go for `PATH is modified!` and then putting the rest fitting on the next line? What do you think?

dunn:
Three changes:
- Fixed the `git diff` to detect committed as well as staged changes
- changed the `ohai` to `opoo` (I think it should stand out a bit from all the other `ohai`s, but either way is OK, really)
- Modified the message as below:
screenshot 2015-07-11 14 03 21

PR #41398, opened 2015-07-06
brew.1: document `install --force-bottle`

dunn:
`--force-bottle` is explained for `brew fetch` but not `install`.


DomT4:
Makes sense to me. Merged in https://github.com/Homebrew/homebrew/commit/dd5f2528ea9e2442dbe196637abdf74f9e363f07. Thanks @dunn!

PR #41402, opened 2015-07-06
[WIP] openwsman & wsmancli 2.6.0

L2G:


L2G:
Don't merge this yet. pkg-config is incorrectly reporting the installed version as 2.4.4, not 2.6.0.

L2G:
I created this ticket upstream: Openwsman/openwsman#72

Library/Formula/openwsman.rb
```diff
@@ -18,6 +18,8 @@ class Openwsman < Formula
   depends_on "pkg-config" => :build
   depends_on "openssl"
 
+  patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

L2G: I will, but since it will involve fixing up their CMake scripts, it will take a _lot_ more time than this little patch took. It may be just as well, since this patch isn't working anyway.

mikemcquaid: Yeh, we only want to use upstreamable patches, really.

apjanke:
Looks like `openwsman` has already been bumped to 2.6.0 in 82b383999ba6b08664a88ec3092edfd0313f60a4. Is this PR still needed?

nfranke:
I think something is needed since the current brew version of wsmancli gets a segmentation fault when I try to do anything with Intel's AMI stuff.

tdsmith:
Closing as mooted; please let me know if I'm wrong!

PR #41405, opened 2015-07-06
ldc: 0.15.2-beta2

John-Colvin:


John-Colvin:
What's with the missing test results?

DomT4:
They aren't missing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/29095/version=mavericks/testReport/junit/brew-test-bot/mavericks/install_ldc/

John-Colvin:
I had to authorise access otherwise it wouldn't load them. Also, I see  ```Some checks have failed — Build finished. No test results found.``` in the test/checks dialogue on this page

DomT4:
I believe you have to authorise access to the test-bot to view the logs. I think that is out of our hands. @mikemcquaid will know.

John-Colvin:
Ok, I understand and can see the build logs now.

The problem is that the currently released version of ldc (v0.5.1) doesn't build with llvm 3.6.1
Is there any way to proceed?

mikemcquaid:
> The problem is that the currently released version of ldc (v0.5.1) doesn't build with llvm 3.6.1
> Is there any way to proceed? Currently we're totally reliant on the bottle for 0.5.1

What do you think we should do; make the beta version the stable one temporarily?

John-Colvin:
We do have patches for llvm 3.6.1 support for 0.5.1, but they touch 40+ files and are likely to be less well tested than 0.5.2-beta2

temporary beta as stable then? https://github.com/Homebrew/homebrew/pull/41511

John-Colvin:
Ok, now #41511 is merged, I guess this can be closed.

PR #41155, opened 2015-06-28
giflib 4.2.3

nijikon:
This PR update giflib to 4.2.3 and all brews that uses it. Feel free to comment if I done everything correctly.

PR #41408, opened 2015-07-06
gdal: bump revision

DomT4:
Fixes https://github.com/Homebrew/homebrew/pull/41155#issuecomment-118449123 until we can get the version bump done.


DomT4:
We had the ML failure before on `points2grid` as well. It's isolated to ML & we haven't had any reports that it's actually an issue, so going to ignore it again.

PR #41413, opened 2015-07-06
Add --head-only and --devel-only flags to brew-audit

jawshooah:
This allows head-only and devel-only Formulae in custom taps to be audited without throwing spurious errors.

Use-case: I maintain two custom taps, [homebrew-pyenv](https://github.com/jawshooah/homebrew-pyenv) and [homebrew-nodenv](https://github.com/jawshooah/homebrew-nodenv), each of which contains head-only Formulae. I'd like to add a Travis configuration to each that runs `brew audit --strict --head-only ./*.rb`.

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -18,6 +18,9 @@ module Homebrew
       style
     end
 
+    head_only = ARGV.include? "--head-only"
+    devel_only = ARGV.include? "--devel-only"
+
     ENV.activate_extensions!
     ENV.setup_build_environment
 
@@ -50,7 +53,7 @@ module Homebrew
     output_header = !strict
 
     ff.each do |f|
-      fa = FormulaAuditor.new(f, :strict => strict)
+      fa = FormulaAuditor.new(f, :strict => strict, :head_only => head_only, :devel_only => devel_only)
       fa.audit
 
       unless fa.problems.empty?
@@ -131,6 +134,8 @@ class FormulaAuditor
   def initialize(formula, options={})
     @formula = formula
     @strict = !!options[:strict]
+    @head_only = !!options[:head_only]
+    @devel_only = !!options[:devel_only]
     @problems = []
     @text = FormulaText.new(formula.path)
     @specs = %w{stable devel head}.map { |s| formula.send(s) }.compact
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Related: https://github.com/Homebrew/homebrew/pull/41289

DomT4:
I prefer this mechanism over #41289. Would be cool with dropping `--head-only` and `--devel-only` in favour of `--HEAD` and `--devel` though since that's what we'd pass to `brew` to install and test the formulae involved.

mikemcquaid:
@DomT4 For me it's about: do we want to be more friendly to tap maintainers or tap users? I think this approach is more friendly to tap maintainers but it basically means those who choose to can ignore all head-only/devel-only submissions (which I think provides a bad user experience to users). Please feel free to disagree with me here, though!

xu-cheng:
@mikemcquaid personally I prefer this approach as well. Part of reason is people indent to name the tap after project name. After some development, when the project moves to stable, they are likely to use the same tap to distribute the software. So to me, it's more reasonable and flexible to use argument rather than tap name in this matter.

mikemcquaid:
@xu-cheng I can see an argument for `devel` only but not for head-only that way unfortunately. head-only provides a really, really poor user experience.

DomT4:
@mikemcquaid To some extent, I think for non-official Homebrew taps we should leave the decision about which taps have great or crappy behaviour to the end user. Sometimes unusual solutions or workarounds that we'd never touch in the core provide the experience the user wants.

Neither proposal is bad. I consider this to be more powerful than #41289 but #41289 is less likely to cause problems for users in unofficial taps whilst still giving tap maintainers a little more control over unrelated audit failures. I guess it's a dance between power and providing the best end-user experience.

If you're happier with the other PR I've no objection to it.

mikemcquaid:
@xu-cheng Do you object to the other PR?

xu-cheng:
I still would prefer this PR's approach. IMO, giving the developers much control on `brew audit` command would eventually benefit end-user experience. Because without this, non-official developing Homebrew taps would more than likely abandon `brew audit` check in their CI system.

I don't have too much objection on the other PR. It just seems to me this PR would be a better and long term approach.

mikemcquaid:
> Because without this, non-official developing Homebrew taps would more than likely abandon brew audit check in their CI system.

I think the response by the contributors in both PRs says the opposite to me. Again, I don't mind `devel` and `head-only` having different functionality but telling people to `brew install user/tap/formula` and then running `brew upgrade` to stumble upon our HEAD mess is a bad user experience and I think we should discourage that in taps.

jawshooah:
Why not allow either approach? Formulae in taps marked as head-only or devel-only will be treated as such automatically; otherwise a CLI flag is required.

DomT4:
I'm not sure it's worth accepting both and having to support both going forwards when they cover loosely the same objective & functionality.

PR #41415, opened 2015-07-06
dependency-check 1.2.11 (new formula)

colezlaw:
New formula for OWASP Dependency Check. This will link the
`dependency-check` script, links `var/dependencycheck` into
`libexec/data`, and `etc/dependencycheck` into `libexec/etc`.

### DIFF
--- a/Library/Formula/dependency-check.rb
+++ b/Library/Formula/dependency-check.rb
@@ -1,0 +1,45 @@
+class DependencyCheck < Formula
+  desc "OWASP Dependency Check"
+  homepage "https://www.owasp.org/index.php/OWASP_Dependency_Check"
+  url "https://dl.bintray.com/jeremy-long/owasp/dependency-check-1.2.11-release.zip"
+  version "1.2.11"
+  sha256 "1a622a1c79a7bff88950fa42294d634a8155958960d656d9d796785f056d88b2"
+
+  depends_on :java
+
+  def install
+    rm_f Dir["bin/*.bat"]
+
+    chmod 0755, "bin/dependency-check.sh"
+
+    prefix.install_metafiles
+    libexec.install Dir["*"]
+
+    File.rename "#{libexec}/bin/dependency-check.sh", \
+      "#{libexec}/bin/dependency-check"
+    bin.install_symlink libexec/"bin/dependency-check"
+
+    (var/"dependencycheck").mkpath
+    libexec.install_symlink var/"dependencycheck" => "data"
+
+    (etc/"dependencycheck").mkpath
+    corejar = libexec/"repo/org/owasp/dependency-check-core/#{version}/"\
+      "dependency-check-core-#{version}.jar"
+    system "unzip", "-o", corejar, "dependencycheck.properties", "-d", \
+      etc/"dependencycheck"
+    libexec.install_symlink etc/"dependencycheck" => "etc"
+  end
+
+  test do
+    output = `#{libexec}/bin/dependency-check --version`.strip
+    assert_equal("Dependency-Check Core version 1.2.11", output)
+
+    props = File.open("temp-props.properties", "w")
+    props.puts "cve.startyear=2015"
+    props.close
+
+    system "#{bin}/dependency-check", "-P", "temp-props.properties", \
+      "-f", "XML", "-a", "dc", "-s", libexec, "-d", testpath, "-o", testpath
+    assert(File.exist?(testpath/"dependency-check-report.xml"))
+  end
+end


Library/Formula/dependency-check.rb
```diff
@@ -0,0 +1,41 @@
+class DependencyCheck < Formula
+  desc "OWASP Dependency Check"
+  homepage "https://www.owasp.org/index.php/OWASP_Dependency_Check"
+  url "https://dl.bintray.com/jeremy-long/owasp/"\
```
DomT4: Just make this a single line. No need to newline the end of the URL.

colezlaw: The style guide says that lines need to be 80 characters or fewer - the URL continues on the following line.

DomT4: We don't apply the style guide to URLs, unless those URLs include a `tag` or `revision`. It's impractical to force every URL onto two lines just to strictly adhere to 80 characters.

Library/Formula/dependency-check.rb
```diff
@@ -0,0 +1,41 @@
...
+    rm_f Dir["bin/*.bat"]
+
+    chmod 0755, "bin/dependency-check.sh"
+
+    prefix.install_metafiles
+    libexec.install Dir["*"]
+
+    File.rename "#{libexec}/bin/dependency-check.sh", \
+      "#{libexec}/bin/dependency-check"
+    (HOMEBREW_PREFIX/"bin").install_symlink libexec/"bin/dependency-check"
```
DomT4: Just `bin.install_symlink libexec/"bin/dependency-check"`. We don't directly put *anything* into the common bin. It's all symlinks.

Library/Formula/dependency-check.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    # Now, link var/dependencycheck to the default data directory
+    (var/"dependencycheck").mkpath
+    ln_s var/"dependencycheck", libexec/"data"
```
DomT4: `(var/"dependencycheck").install_symlink libexec/"data"` should work here.

Library/Formula/dependency-check.rb
```diff
@@ -0,0 +1,41 @@
...
+      etc/"dependencycheck"
+    ln_s etc/"dependencycheck", libexec/"etc"
+  end
+
+  test do
+    output = `#{libexec}/bin/dependency-check --version`.strip
+    assert_equal("Dependency-Check Core version 1.2.11", output)
+  end
+end
+
```
DomT4: This line can be cut. Files end with an invisible new line.

Library/Formula/dependency-check.rb
```diff
@@ -0,0 +1,41 @@
...
+    (var/"dependencycheck").mkpath
+    ln_s var/"dependencycheck", libexec/"data"
+
+    # Extract the included properties to etc and link
+    (etc/"dependencycheck").mkpath
+    corejar = libexec/"repo/org/owasp/dependency-check-core/#{version}/"\
+      "dependency-check-core-#{version}.jar"
+    system "unzip", "-o", corejar, "dependencycheck.properties", "-d", \
+      etc/"dependencycheck"
+    ln_s etc/"dependencycheck", libexec/"etc"
```
DomT4: `(etc/"dependencycheck").install_symlink libexec/"etc"`.

Library/Formula/dependency-check.rb
colezlaw: Hrmm - doesn't seem to work on my system. No error, but the symlink doesn't get created, and brew unlink only removes one link.

Library/Formula/dependency-check.rb
colezlaw: Doesn't seem to work on my system. No error, but the symlink doesn't get created, and brew unlink only removes one link.

Library/Formula/dependency-check.rb
DomT4: Hm. Maybe we don't support it for `var` and I've just completely missed that particular memo :octopus:. The `etc` one should work, I used that locally the other night.

colezlaw: That's the reverse of what I want. I want for the `data` directory to persist between version updates, so I prefer to keep the data in `HOMEBREW_PREFX/var/dependencycheck`. Going the other direction (`(libexec/"data").install_symlink var/"dependencycheck"`) creates a link **in** `libexec/"data"` called `dependencycheck` - I want `libexec/data` to be the symlink, and for it to point to `var/dependencycheck`.

DomT4: Oh, I misread what you wrote, sorry. Doing too much right now. In that case yes, you do want `(libexec/"data").install_symlink var/"dependencycheck"`.

Library/Formula/dependency-check.rb
```diff
@@ -0,0 +1,40 @@
...
```
DomT4: Likewise, `libexec.install_symlink etc/"dependencycheck"` here.

Library/Formula/dependency-check.rb
colezlaw: That creates the symlink **in** `libexec/"data"`. I want `libexec/"data"` to **be** the symlink. Is there a way to do `libexec.install_symlink var/"dependencycheck" :named => data` or something like that?

Library/Formula/dependency-check.rb
colezlaw: As the `var` one above - I want the link to be named `etc`. Is there an option for that?

Library/Formula/dependency-check.rb
DomT4: Yes. You pretty got the method spot on there, except we don't specify `:named`. Just `libexec.install_symlink var/"dependencycheck" => "data"`.

Library/Formula/dependency-check.rb
DomT4: Same gig. `libexec.install_symlink etc/"dependencycheck" => "etc"`

Library/Formula/dependency-check.rb
colezlaw: Lucky guess ;)

Library/Formula/dependency-check.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    File.rename "#{libexec}/bin/dependency-check.sh", \
+      "#{libexec}/bin/dependency-check"
+    bin.install_symlink libexec/"bin/dependency-check"
+
+    # Now, link var/dependencycheck to the default data directory
+    (var/"dependencycheck").mkpath
+    libexec.install_symlink var/"dependencycheck" => "data"
+
+    # Extract the included properties to etc and link
```
mikemcquaid: Kill the comments above, thanks.

Library/Formula/dependency-check.rb
```diff
@@ -0,0 +1,39 @@
...
+    (etc/"dependencycheck").mkpath
+    corejar = libexec/"repo/org/owasp/dependency-check-core/#{version}/"\
+      "dependency-check-core-#{version}.jar"
+    system "unzip", "-o", corejar, "dependencycheck.properties", "-d", \
+      etc/"dependencycheck"
+    libexec.install_symlink etc/"dependencycheck" => "etc"
+  end
+
+  test do
+    output = `#{libexec}/bin/dependency-check --version`.strip
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

colezlaw:
To really test it would require that it download a whole bunch of data which takes in the neighborhood of 10 minutes, which violates other HOMEBREW standards.

mikemcquaid:
@colezlaw I'm happy with any trivial non-version test to verify a simple project (that you can write out with `Pathname.write` and `EOS.undent`)

colezlaw:
The problem is not with the size of the project, but that Dependency Check
downloads and caches all CVE data when it first runs - this is the data
that takes a long time to fetch, not the project that's being analyzed.

I could potentially drop in a fake database or something and test with that
rather than downloading. Or I could test without downloading, which will
cause the scan to fail, but it will be a "known failure".

mikemcquaid:
How long does it take to download, roughly? Thanks!

colezlaw:
Usually in the neighborhood of 10 minutes.

colezlaw:
Alright - sorry for the strain on the CI systems - the test will now limit the CVE download to 2015 and newer, so should execute in under a minute, while not hurting the user's installation. Plus the unnecessary comments removed. Passes tests, audit, style checks, etc.

mikemcquaid:

PR #41416, opened 2015-07-06
syncthing: Fixed not being able to stop process through web UI

dustinwilson:
First I removed RunAtLoad from the plist because RunAtLoad isn't needed if KeepAlive is set. Both run the process at load, but KeepAlive keeps the process alive no matter what.

Because KeepAlive was set to true if you went to the Web UI and tried to stop syncthing launchd would simply restart the process. However, if you set SuccessfulExit to false it will allow you to manually stop the process as intended through the Web UI. Also set Crashed to true so that if it does crash launchd will still restart it.


mikemcquaid:

PR #41417, opened 2015-07-06
h2object version 1.0.0 release

liujianping:
h2object version 1.0.0 release, with source code open at https://github.com/h2object/h2object .

### DIFF
--- a/Library/Formula/h2object.rb
+++ b/Library/Formula/h2object.rb
@@ -1,5 +1,5 @@
 class H2object < Formula
-  homepage "http://h2object.io"
+  homepage "https://github.com/h2object/h2object"
   url "http://dl.h2object.io/h2object/macosx/1.0.0.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/h2object.rb
```diff
@@ -0,0 +1,14 @@
+class H2object < Formula
+  desc "another fast & flexible static website generator & deployment tool"
+  homepage "https://github.com/h2object/h2object"
+  url "http://dl.h2object.io/h2object/macosx/1.0.1.tar.gz"
```
bfontaine: Is there no way to build this from source?

bfontaine:
This is a self-submission of a relatively new project which doesn’t seem to be used at the moment (0 stars on GitHub). I’m afraid it doesn’t meet our core rules. However, you can maintain the formula in your own tap and maybe make a new pull-request in the future :smiley:

PR #41424, opened 2015-07-06
GEOS disable C asserts

nathancahill:
Per the comments here: https://github.com/Toblerity/Shapely/issues/262#issuecomment-97555385 and here: https://github.com/Toblerity/Shapely/issues/288#issuecomment-118924677

### DIFF
--- a/Library/Formula/geos.rb
+++ b/Library/Formula/geos.rb
@@ -28,6 +28,7 @@ class Geos < Formula
 
     args = [
       "--disable-dependency-tracking",
+      "--disable-cassert",
       "--prefix=#{prefix}",
     ]
 


sgillies:
@nathancahill great, you beat me to it!

DomT4:
Is it likely anyone would like us to retain the existing behaviour as an `option`?

nathancahill:
Not too much online about the option besides this thread: http://blog.gmane.org/gmane.comp.gis.geos.devel/month=20060701/page=2

> --enable-cassert=yes (GEOS equivalent of --enable-debug=yes)

Unless I'm mistaken, you'd only want this if you're building from source, so we probably don't need an option for it. Thoughts @sgillies?

mistydemeo:
For reference, I did some looking into other package managers and it doesn't appear that it is common for other package managers to pass `--disable-cassert`. Is there something else different about how Homebrew is building geos that is resulting in behaviour different than what you see with, for example, Macports, Debian/Ubuntu, etc.?

nathancahill:
Right, here is latest build log from UbuntuGIS and it's no built with `--disable-cassert`: https://launchpadlibrarian.net/176927594/buildlog_ubuntu-lucid-amd64.geos_3.4.2-6~lucid1_UPLOADING.txt.gz

Sean pinged the mailing list about this, so I guess this pull should wait for feedback from there.

DomT4:
Any update from upstream?

bfontaine:
Ping?

sgillies:
The GEOS maintainer has no interest in discussing or fixing this. I've found a solution for Shapely (mentioned at top) and that is to never doubly dlopen (in a process) the GEOS shared library on OS X. In the case where I dlopen twice (Shapely and gdal or Rasterio) I can prevent an abort by disabling the asserts, but the results of geometric union operations are incorrect. In other words, disabling asserts exposes you to incorrect results. We can close this issue.

bfontaine:
Is it worth adding an option for that?

mikemcquaid:
Thanks @sgillies!

PR #41426, opened 2015-07-06
magit: fix resource dash.el

dunn:
MELPA doesn't maintain URLs to old versions of packages, so switch to the GitHub source.

### DIFF
--- a/Library/Formula/magit.rb
+++ b/Library/Formula/magit.rb
@@ -1,7 +1,6 @@
 class Magit < Formula
   desc "Emacs interface for Git"
   homepage "https://github.com/magit/magit"
-
   url "https://github.com/magit/magit/releases/download/2.1.0/magit-2.1.0.tar.gz"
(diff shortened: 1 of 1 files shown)


bfontaine:
The test is failing:

```
==> emacs -batch -l /private/tmp/magit20150706-52573-1nsk11r/test.el
Cannot open load file: No such file or directory, dash
Error: magit: failed
<0> expected but was
<255>.
```

dunn:
Hrm, can't replicate that locally. Let's try this.

dunn:
Finally managed to reproduce these errors. Specifying the path in `config.mk` is only for make—it doesn't help emacs find dash at run-time.  Adding a `vendor` folder would mean requiring users to include `(add-to-list 'load-path "#{lib}/vendor")` to their init file for every formulae, which isn't worth it.

mikemcquaid:
Thanks @dunn!

PR #41430, opened 2015-07-06
open-mpi 1.8.6

dunn:
Release notes say the memory leak has been fixed.

### DIFF
--- a/Library/Formula/open-mpi.rb
+++ b/Library/Formula/open-mpi.rb
@@ -1,17 +1,21 @@
 class OpenMpi < Formula
   desc "High performance message passing library"
   homepage "https://www.open-mpi.org/"
-  # Wait for 1.8.6 and skip 1.8.5 due to a severe memory leak on OS X:
-  # https://github.com/open-mpi/ompi/issues/579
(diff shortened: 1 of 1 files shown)


Library/Formula/open-mpi.rb
```diff
@@ -50,9 +55,11 @@ def install
     # (Fortran header) in `lib` that need to be moved to `include`.
     include.install Dir["#{lib}/*.mod"]
 
-    # Move vtsetup.jar from bin to libexec.
-    libexec.install bin/"vtsetup.jar"
-    inreplace bin/"vtsetup", "$bindir/vtsetup.jar", "$prefix/libexec/vtsetup.jar"
+    if build.stable?
```
DomT4: Does this part not need doing for `head`?

dunn: `vtsetup.jar` doesn't exist in HEAD, or at least isn't in the same place, so it causes the build to fail.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/e413ef58658e094b211034b95836c90da2c7c6f2. Thanks @dunn! :smiley_cat:

PR #41432, opened 2015-07-06
objective-caml 4.02.2

dunn:


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/65e2af27fcc4055ac5dcfd804fe69570b1eb7cf0. Thanks Alex! :beers:

ches:
If I understand everything correctly, don't OCaml version bumps need a corresponding update to the `camlp4` formula to the same release?

See #33596 and cc @avsm—for me it appears that OPAM library deps/upgrades still get wrecked when the `camlp4` version lags behind `objective-caml`, as opam starts trying to install camlp4 as a library. I assume this is part of the lingering problems Anil referred to on that issue as 'while the package ecosystem migrates to [camlp4] being "just another OCaml library"'.

DomT4:
We really need to document this mess inside the relevant formulae. You aren't going to have every contributor and maintainer memorise each and every step to take to handle basic `ocaml` bumps, and the status quo of "Did this break something? Maybe? Probably? Better check it out" is frustrating for everyone involved.

Let's work out what, if anything, is broken here, fix it however it needs fixing, and then document the checks that need to take place in future at the top of the respective formulae.

avsm:
Yes, I agree that documenting in the formula is the way to go, since Homebrew doesn't have package maintainers.  It's slightly frustrating that there's no way to be informed before an update is merged. In this case, I've been testing a fix to 4.02.2 that causes crashes in some circumstances, and it would have been nice to have the patch in along with this update: http://caml.inria.fr/mantis/view.php?id=6919

I'll submit a camlp4 and ocaml port fix shortly...

DomT4:
Thanks @avsm, Appreciate the quick response on this.

PR #41433, opened 2015-07-07
Default to not using bottles for edited formula.

### LINKED ISSUE #36068: Suggested feature: Automagically build bottled brews if their formula has been edited

DylanLacey:
Added `formula.adulterated`, which uses git in the formula's repo
directory to check for alterations (with `git status --porcelain`).

Altered `pour_bottle?` in formula_installer to check for adulterated
formula, after all commandline flags.  Adulterated formula will output a
warning and then build from source.  Forcing a bottle install with
`--force-bottle` will still work.

Also works for `fetch`.

Closes #36068, Compare with https://github.com/Homebrew/homebrew/pull/41397.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -261,11 +261,30 @@ class Formula
     (dir = installed_prefix).directory? && dir.children.length > 0
   end
 
+  # If this {Formula} has been modified by the user.  Relies on git.
+  def adulterated?
+    return false unless git_available?
+    repo = core_formula? ? HOMEBREW_LIBRARY : path_of_repository
+    repo.cd do
+      git_status = Utils.popen_read("git", "status", "-z", "--porcelain", #{path}).empty?
+      return ( git_status.empty? && $?.success? )
+    end
+    false
+  end
+
   # @deprecated
   # The `LinkedKegs` directory for this {Formula}.
   # You probably want {#opt_prefix} instead.
   def linked_keg
     Pathname.new("#{HOMEBREW_LIBRARY}/LinkedKegs/#{name}")
+  end
+
+  def path_of_repository
+    if tap?
+      Pathname.new($&).realpath
+    else
+      HOMEBREW_REPOSITORY
+    end
   end
 
   # The latest prefix for this formula. Checks for {#head}, then {#devel}
--- a/Library/Homebrew/cmd/fetch.rb
+++ b/Library/Homebrew/cmd/fetch.rb
@@ -32,7 +32,7 @@ module Homebrew
   def fetch_bottle? f
     return true if ARGV.force_bottle? && f.bottle
     return false unless f.bottle && f.pour_bottle?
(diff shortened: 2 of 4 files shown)


DomT4:
CC @Homebrew/owners

Library/Homebrew/formula.rb
```diff
@@ -261,6 +261,15 @@ def installed?
     (dir = installed_prefix).directory? && dir.children.length > 0
   end
 
+  # If this {Formula} has been modified by the user.  Relies on git.
+  def adulterated?
```
xu-cheng: Let's add `return false unless which("git")` to handle the case where git is unavailable. This can happen for Linuxbrew.

Library/Homebrew/formula.rb
```diff
@@ -261,6 +261,15 @@ def installed?
...
+    repo = core_formula? ? HOMEBREW_LIBRARY : path_of_repository
+    repo.cd do
```
xu-cheng: Why not just `path.parent.cd ...`?

Library/Homebrew/formula.rb
```diff
@@ -261,6 +261,15 @@ def installed?
...
+      return true unless `git status --porcelain #{path}`.empty?
```
xu-cheng: You need to check whether the file is under git control. Also please use `Utils.popen_read("git", "status", ...)` instead of backticks to avoid shell injection.

Library/Homebrew/formula.rb
DylanLacey: Good call!

Library/Homebrew/formula.rb
DylanLacey: I originally just targeted things in `HOMEBREW_LIBRARY` and added `path_of_repository` later, and didn't refactor down.

Library/Homebrew/formula.rb
DylanLacey: Hmm, it might be best to just merge #41397; I added this because I had it ready and lack the experience with Homebrew to know which was better.

DylanLacey:
Changes made!  I don't really like the whole `?$.success` bit because it's starting to get a bit punctuation-salady, but making another method seemed a *little* intense.

Library/Homebrew/formula.rb
mikemcquaid: Let's call this `file_modified?`

mikemcquaid:
https://github.com/Homebrew/homebrew/pull/41397 looks like a better fit for now so closing just so we don't ask you both to make the same changes.

PR #41434, opened 2015-07-07
logtalk: update to 3.00.4

FranklinChen:
Also, remove devel because 3.00 is stable now.


DomT4:
Merged in https://github.com/Homebrew/homebrew/blob/1357d95f74397d7eb712cde1b314a42a3174c2b0/Library/Formula/logtalk.rb. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a nit but version updates should have their commit styled `logtalk 3.00.4` or similar, Thanks!

PR #41440, opened 2015-07-07
Fixes wrong scons' sha256.

linjunpop:


DomT4:
Can we ask upstream why the SHA has changed here? Thanks.

tdsmith:
Looks legit; http://article.gmane.org/gmane.comp.programming.tools.scons.user/26929

DomT4:
Thanks Tim :bow:. Would be nice if they'd bump the version very slightly in these situations rather than silently changing the SHA :/.

mikemcquaid:

For future reference the preferred commit message format is `scons: fix sha256.`.

PR #41445, opened 2015-07-07
mysql-connector-odbc: remove mysql dependency, update home page.

### LINKED ISSUE #39126: mysql-connector-odbc doesn't depend on mysql
Could someone verify that mysql-connector-odbc doesn't depend on mysql? The odbc driver needs to be installed on the client and server side. ODBC will be useless if every driver needs its database installed on the client side. Cheers, Leo

apjanke:
Fixes #39126

Removes the `mysql` dependency. The MySQL Connector/ODBC is just a client-side driver, built from a standalone distribution. It does not depend on a local installation of the MySQL server, which is what the `mysql` formula is.

Updates the homepage based on a redirect found at the previous home page URL.

Adds caveats about unixODBC/iODBC and a minimal test.

I've built from source on OS X 10.9.5 using this new definition. Builds fine, and the `myodbc-installer` runs to register and deregister drivers.

### DIFF
--- a/Library/Formula/mysql-connector-odbc.rb
+++ b/Library/Formula/mysql-connector-odbc.rb
@@ -1,6 +1,6 @@
 class MysqlConnectorOdbc < Formula
   desc "Standardized database driver"
-  homepage "https://dev.mysql.com/doc/refman/5.1/en/connector-odbc.html"
+  homepage "https://dev.mysql.com/doc/connector-odbc/en/index.html"
   url "https://cdn.mysql.com/Downloads/Connector-ODBC/5.3/mysql-connector-odbc-5.3.4-src.tar.gz"
   sha256 "a5f7a490f2958f2768d18b8a57f71909f9699a8619c82776b3ad1c02b8abce0d"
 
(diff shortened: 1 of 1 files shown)


Library/Formula/mysql-connector-odbc.rb
```diff
@@ -1,6 +1,6 @@
 class MysqlConnectorOdbc < Formula
   desc "Standardized database driver"
-  homepage "https://dev.mysql.com/doc/refman/5.1/en/connector-odbc.html"
+  homepage "http://dev.mysql.com/doc/connector-odbc/en/index.html"
```
mikemcquaid: Please use an `https` URL.

apjanke: Fixed.

apjanke:
This is a mistake... `mysql-connector-odbc` does have a dependency on the client libraries which are part of the `mysql` distribution. I failed to notice them in testing because I had an old non-Homebrew MySQL installation sitting on my machine when working on this.

Closing. Do not merge.

PR #41451, opened 2015-07-07
typo fix and test fix

botanize:
Fixed typo in description. Test changed to something that could actually pass without first starting hadoop/hdfs.


mikemcquaid:
Are you sure the test was failing before?

botanize:
Sorry about that. I had thought ```hadoop fs``` had been deprecated but I see now that it defaults to local filesystem when hdfs isn't running. I had initially changed the test to ```bin/hdfs dfs -ls```, but that failing when hdfs wasn't yet running. I reverted the test. So I guess this is just a typo fix.

mikemcquaid:

For future reference the preferred commit message format is `hadoop: fix typo in description.`.

PR #41452, opened 2015-07-07
Suricata: disable broken explicit non-debug build.

barn:
Due to https://github.com/OISF/libhtp/issues/106, passing
--disable-debug will actually cause libhtp to be built in debug mode.
Until that issue is closed the fix is to just not specify
--disable-debug and let it build a non-debug version anyway.


barn:
Well I messed that up. Sorry.

PR #41463, opened 2015-07-07
OCaml 4.02.2 fix, and matching camlp4 update

avsm:
See discussion in #41432

### DIFF
--- a/Library/Formula/camlp4.rb
+++ b/Library/Formula/camlp4.rb
@@ -1,9 +1,9 @@
 class Camlp4 < Formula
   desc "Tool to write extensible parsers in OCaml"
   homepage "https://github.com/ocaml/camlp4"
-  url "https://github.com/ocaml/camlp4/archive/4.02.1+2.tar.gz"
-  sha1 "ad43587c2bdc46d5d0f028857f5ce90bc9ce7ab5"
-  version "4.02.1+2"
+  url "https://github.com/ocaml/camlp4/archive/4.02+6.tar.gz"
+  sha1 "6dd7e591dfde94c44576dba9b847279ffb92c889"
+  version "4.02.2+6"
   head "https://github.com/ocaml/camlp4.git"
 
   bottle do
--- a/Library/Formula/objective-caml.rb
+++ b/Library/Formula/objective-caml.rb
@@ -1,9 +1,23 @@
+# OCaml does not preserve binary compatibility across compiler releases,
+# so when updating it you should ensure that all dependent packages are
+# also updated by incrementing their revisions.
+#
+# Specific packages to pay attention to include:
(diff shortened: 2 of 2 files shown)


Library/Formula/camlp4.rb
```diff
@@ -1,9 +1,9 @@
 class Camlp4 < Formula
   desc "Tool to write extensible parsers in OCaml"
   homepage "https://github.com/ocaml/camlp4"
-  url "https://github.com/ocaml/camlp4/archive/4.02.1+2.tar.gz"
-  sha1 "ad43587c2bdc46d5d0f028857f5ce90bc9ce7ab5"
-  version "4.02.1+2"
+  url "https://github.com/ocaml/camlp4/archive/4.02+6.tar.gz"
+  sha1 "6dd7e591dfde94c44576dba9b847279ffb92c889"
```
DomT4: Let's use `sha256` here, Thanks! Everything is being updated en-mass soon, so it's nice to gradually reduce the amount that has to be changed first.

mikemcquaid:
If you add something to the `camlp4` test that checks the OCaml version it'll be run everything we bump it.

Library/Formula/objective-caml.rb
```diff
@@ -1,9 +1,23 @@
+# OCaml does not preserve binary compatibility across compiler releases,
+# so when updating it you should ensure that all dependent packages are
+# also updated by incrementing their revisions.
```
mikemcquaid: Is there a way we can verify this in the `test do` blocks for those packages so they fail?

avsm:
@mikemcquaid there's no hard check in camlp4 on the minor version. It actually works fine with either 4.02.1 or 4.02.2 since the AST format hasn't changed.  It just needs to be recompiled if the version of OCaml changes.

I think that in Homebrew terms this means that we have to manually bump the revision of camlp4 every time ocaml changes, right?  There's no other way to express an exact version dependency.

Library/Formula/objective-caml.rb
avsm: The only one that we can do is a local OPAM bootstrap.  I'll take a look at this.

mikemcquaid:
> I think that in Homebrew terms this means that we have to manually bump the revision of camlp4 every time ocaml changes, right? There's no other way to express an exact version dependency.

Yeh, correct. I guess I mean is there a way you could e.g. add to the `test do` block something (even if it's just Ruby) to verify what you're saying we need to do?

mikemcquaid:

Library/Formula/objective-caml.rb
mikemcquaid: That's be great to add.

PR #41474, opened 2015-07-08
Upgrade from Mapnik v2.2.0 to v3.0.0

springmeyer:
Key changes:

  - Mapnik 3.x no longer bundles python bindings. So now homebrew will just install the C++ library. Python bindings can be pip installed from https://github.com/mapnik/python-mapnik
  - New dependencies of harfbuzz and webp
  - Removed patches no longer needed
  - Now using Mapnik's `Makefile` shim instead of calling Scons directly (recommended by upstream)

More details on 3.0.0 release:

 - http://mapnik.org/news/release-3.0.0/
 - https://github.com/mapnik/mapnik/milestones/Mapnik%203.0.0

### DIFF
--- a/Library/Formula/mapnik.rb
+++ b/Library/Formula/mapnik.rb
@@ -4,7 +4,6 @@ class Mapnik < Formula
   head "https://github.com/mapnik/mapnik.git"
   url "https://s3.amazonaws.com/mapnik/dist/v3.0.0/mapnik-v3.0.0.tar.bz2"
   sha256 "c88b1ce48899ffe0d75f9e753dcd427e2b6fd3186b40e04508608b2151c0e7b0"
-  revision 1
 
   bottle do
     sha256 "1c362e8436c60b47664dfba2bfd18eab7ef281db3e0f376274676bcb1c66685d" => :yosemite
--- a/Library/Formula/mapnik.rb
+++ b/Library/Formula/mapnik.rb
@@ -24,8 +24,11 @@ class Mapnik < Formula
   depends_on "gdal" => :optional
   depends_on "postgresql" => :optional
   depends_on "cairo" => :optional
+  
+  needs :cxx11
 
   def install
+    ENV.cxx11
     icu = Formula["icu4c"].opt_prefix
     boost = Formula["boost"].opt_prefix
     proj = Formula["proj"].opt_prefix
--- a/Library/Formula/mapnik.rb
+++ b/Library/Formula/mapnik.rb
@@ -2,9 +2,9 @@ class Mapnik < Formula
   desc "Toolkit for developing mapping applications"
   homepage "http://www.mapnik.org/"
   head "https://github.com/mapnik/mapnik.git"
-  url "https://s3.amazonaws.com/mapnik/dist/v2.2.0/mapnik-v2.2.0.tar.bz2"
-  sha256 "9b30de4e58adc6d5aa8478779d0a47fdabe6bf8b166b67a383b35f5aa5d6c1b0"
-  revision 6
+  url "https://s3.amazonaws.com/mapnik/dist/v3.0.0/mapnik-v3.0.0.tar.bz2"
+  sha256 "c88b1ce48899ffe0d75f9e753dcd427e2b6fd3186b40e04508608b2151c0e7b0"
+  revision 1
 
   bottle do
     sha256 "1c362e8436c60b47664dfba2bfd18eab7ef281db3e0f376274676bcb1c66685d" => :yosemite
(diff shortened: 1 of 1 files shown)


Library/Formula/mapnik.rb
```diff
@@ -2,9 +2,9 @@ class Mapnik < Formula
   desc "Toolkit for developing mapping applications"
   homepage "http://www.mapnik.org/"
   head "https://github.com/mapnik/mapnik.git"
-  url "https://s3.amazonaws.com/mapnik/dist/v2.2.0/mapnik-v2.2.0.tar.bz2"
-  sha256 "9b30de4e58adc6d5aa8478779d0a47fdabe6bf8b166b67a383b35f5aa5d6c1b0"
-  revision 6
+  url "https://s3.amazonaws.com/mapnik/dist/v3.0.0/mapnik-v3.0.0.tar.bz2"
+  sha256 "c88b1ce48899ffe0d75f9e753dcd427e2b6fd3186b40e04508608b2151c0e7b0"
+  revision 1
```
DomT4: You can drop the `revision` completely on version bumps.

DomT4: `0` is the default revision, but it's unwritten. We only add the line when bumping to `revision 1` or above.

DomT4:
This failed on Mountain Lion. Probably `cxx11` related.

springmeyer:
@DomT4 - Thanks, I removed `revision`. Also, yes Mapnik 3.0.0 requires c++11. But it should work on 10.8. Can you share the build error?

DomT4:
[This](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/29213/version=mountain_lion/testReport/junit/brew-test-bot/mountain_lion/install_mapnik/) was the build failure.

You can get around it by adding `needs :cxx11` one blank line under the dependencies and `ENV.cxx11` to the top of `def install`.

springmeyer:
Thanks @DomT4 - made that change.

Library/Formula/mapnik.rb
```diff
@@ -12,35 +11,24 @@ class Mapnik < Formula
...
   depends_on "jpeg"
+  depends_on "webp"
   depends_on "boost"
-  depends_on "boost-python"
   depends_on "gdal" => :optional
   depends_on "postgresql" => :optional
   depends_on "cairo" => :optional
-
-  depends_on "py2cairo" if build.with? "cairo"
+
```
DomT4: Slight trailing whitespace here. Needs to be :skull:'ed

DomT4:
Boost is throwing a strop on ML now. You may need [something like this](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mal4s.rb#L32-L36).

springmeyer:
@DomT4 - looks like things are green now? Thanks for the help on this!

springmeyer:
The other question I have is bottles: how do they get updated when a new version comes out?

DomT4:
Yup, looks like all three builds went successfully :+1:.

The test-bot automatically generates bottles for all stable builds, and our `pull` mechanism adds them to the formula and then each PR submitted afterwards generate new bottles that we can `pull` to replace the existing bottles if the contents of that bottle has changed (i.e. a version bump, a revision, fixing a problem with the build, adding bash completion, so on).

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/27d24d1def1dacedb4dd6bd7a424972e8aa6ef5b. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a nit but we tend to style commit messages for version bumps simply as `mapnik 3.0.0` or similar, Thanks!

springmeyer:
Thankful for the mentoring and I'll be in great shape to make the next update more seamless.

springmeyer:
@DomT4 - one more question. Based on your bottle explanation above, do I understand right that bottles will automatically be available soon? And `brew install mapnik` will install via a binary? Right now I'm seeing a source build is still used.

DomT4:
Make sure you `brew update`. After that, unless you're using a custom option such as `--with-postgresql` or `--with-cairo` etc `brew install mapnik` should grab you the bottle.

springmeyer:
@DomT4 Still not working via a bottle. I am up to data and am not using custom options. I've noticed that freetype works from a bottle for me, so I looked in freetype.rb and saw the use of `cellar :any`. I tried that and it fixed the Mapnik via bottle install for me: https://github.com/Homebrew/homebrew/pull/41526

BTW noted the `or the bottle's cellar is not :any or equal to the current HOMEBREW_CELLAR.` at https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Bottles.md but I can't make sense of that. What determines a bottles cellar? Is the cellar different because I have homebrew at a custom path (not /usr/local)?

mikemcquaid:
> Is the cellar different because I have homebrew at a custom path (not /usr/local)?

Yes, that means you're unable to use most bottles unfortunately.

PR #41476, opened 2015-07-08
Add "with-openni2" switch to support building with OpenNI2 driver

xionglingfeng:
Add a switch to allow building `libfreenect` with `OpenNI2` driver.


Library/Formula/libfreenect.rb
```diff
@@ -28,6 +29,10 @@ def install
       args << "-DCMAKE_OSX_ARCHITECTURES=#{Hardware::CPU.universal_archs.as_cmake_arch_flags}"
     end
 
+    if build.with? "openni2"
```
DomT4: Since there's not multiple arguments to pass with this option let's just do `args << "-DBUILD_OPENNI2_DRIVER=ON" if build.with? "openni2"`, Thanks!

mikemcquaid:
Any disadvantage to enabling this by default?

xionglingfeng:
I don't think there are any disadvantage..

mikemcquaid:
Just enable it by default and remove the option in that case, thnaks!

xionglingfeng:
I set it as default.

mikemcquaid:
Thanks! Would you mind updating the `sha1` to `sha256` too?

xionglingfeng:
Hi. I already updated sha1 to sha256 for source code file. However, 'bottle' part still leave as sha1 because I don't know how to update that.

DomT4:
You'll need to rebase this. You've accidentally merged 10,000+ Homebrew commits into your branch as new commits, which is causing a fairly major merge conflict.

Don't worry about the bottle SHA1; We'll update that when we pull, Thanks!

xionglingfeng:
Hi  @DomT4 . I just redo the fork and issued another pull request. Please close this. Thank you.

PR #41478, opened 2015-07-08
Move bottle fetch from pull to a pre-push hook

tdsmith:
Publishing a bottle on Bintray is asynchronous. Instead of waiting a long time in `pull` before checking whether the bottle was published correctly, move the check to a pre-push hook.

I have a terrible feeling you're going to ask me to rewrite this in Ruby. Call it a strawman for discussion, anyway.

### DIFF
--- a/Library/Contributions/pre-push
+++ b/Library/Contributions/pre-push
@@ -25,7 +25,7 @@ def commits_between(beginning, end):
     return git("rev-list", "{}..{}".format(beginning, end)).split("\n")
 
 def push_okay(remote_name, remote_url, updates_file):
-    if "omebrew/homebrew" not in remote_url:
+    if "homebrew/homebrew" not in remote_url.lower():
         return True
     updates = [line.strip().split() for line in updates_file.readlines()]
     # remote_ref (at commit remote_sha) will  be updated by local_ref (to commit local_sha)
--- a/Library/Contributions/pre-push
+++ b/Library/Contributions/pre-push
@@ -1,0 +1,47 @@
+#!/usr/bin/python
+# Makes sure bottles are fetchable after BrewTestBot modifies a formula.
+# Only checks pushes to master on Homebrew/homebrew* remotes.
+# To use, `ln -s $(brew --repository)/Library/Contributions/pre-push
+# $(brew --repository)/.git/hooks`.
+
+import re
+import os
+import subprocess
+import sys
+
+FORMULA_RE = re.compile(r"^(Library/Formula/)?([A-Za-z0-9-]+)\.rb")
+TESTBOT_EMAILS = ["brew-test-bot@googlegroups.com"]
+
+def git(*args):
+    return subprocess.check_output(["git"] + list(args)).strip()
+
+def commit_author_email(commit_sha):
+    return git("show", "-s", "--pretty=%ae", commit_sha)
+
+def files_touched_by_commit(commit_sha):
+    return git("diff-tree", "--name-only", "-r", "--no-commit-id", commit_sha).split("\n")
+
+def commits_between(beginning, end):
+    return git("rev-list", "{}..{}".format(beginning, end)).split("\n")
+
+def push_okay(remote_name, remote_url, updates_file):
+    if "omebrew/homebrew" not in remote_url:
+        return True
+    updates = [line.strip().split() for line in updates_file.readlines()]
+    # remote_ref (at commit remote_sha) will  be updated by local_ref (to commit local_sha)
+    for local_ref, local_sha, remote_ref, remote_sha in updates:
+        if remote_ref != "refs/heads/master": continue
+        for commit_sha in commits_between(remote_sha, local_sha):
+            if commit_author_email(commit_sha) not in TESTBOT_EMAILS: continue
+            for filename in files_touched_by_commit(commit_sha):
+                match = FORMULA_RE.match(filename)
+                if match is None: continue
+                errorlevel = subprocess.call(["brew", "fetch", "--force-bottle", match.group(2)])
+                if errorlevel > 0:
+                    return False
+    return True
+
+if __name__ == "__main__":
+    if os.environ.get("HOMEBREW_SKIP_FETCH", "0") != "0":
+        sys.exit(0)
+    sys.exit(0 if push_okay(sys.argv[1], sys.argv[2], sys.stdin) else 1)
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cmd/pull.rb
```diff
@@ -189,8 +189,14 @@ def pull
...
-            sleep 20
-            safe_system "brew", "fetch", "--retry", "--force-bottle", f.full_name
+          end
+          unless (Pathname.pwd/".git/hooks/pre-push").exist?
+            opoo "Pre-push hook missing"
+            puts <<-EOS.undent
+              Please consider symlinking #{HOMEBREW_CONTRIB}/pre-push into
+              #{Pathname.pwd}/.git/hooks to add a check that bottles are published
+              correctly before pushing to master.
+            EOS
```
xu-cheng: Would it be better to just make the symlink here rather than a message?

Library/Contributions/pre-push
```diff
@@ -0,0 +1,47 @@
...
+                if match is None: continue
+                errorlevel = subprocess.call(["brew", "fetch", "--force-bottle", match.group(2)])
+                if errorlevel > 0:
+                    return False
+    return True
+
+if __name__ == "__main__":
+    if os.environ.get("HOMEBREW_SKIP_FETCH", "0") != "0":
+        sys.exit(0)
+    sys.exit(0 if push_okay(sys.argv[1], sys.argv[2], sys.stdin) else 1)
```
bfontaine: Shouldn’t you check that `len(sys.argv) > 2` before?

Library/Contributions/pre-push
```diff
@@ -0,0 +1,47 @@
...
+    return git("show", "-s", "--pretty=%ae", commit_sha)
+
+def files_touched_by_commit(commit_sha):
+    return git("diff-tree", "--name-only", "-r", "--no-commit-id", commit_sha).split("\n")
+
+def commits_between(beginning, end):
+    return git("rev-list", "{}..{}".format(beginning, end)).split("\n")
+
+def push_okay(remote_name, remote_url, updates_file):
+    if "omebrew/homebrew" not in remote_url:
```
bfontaine: Is `omebrew` a typo here?

Library/Contributions/pre-push
```diff
@@ -0,0 +1,47 @@
...
+        return True
+    updates = [line.strip().split() for line in updates_file.readlines()]
```
bfontaine: You can loop over a file to read its lines:

```
with open("foo") as f:
    for line in f:
        puts line
```

does the same thing as:

```
with open("foo") as f:
    for line in f.readlines():
        puts line
```

except that the former never loads all the lines in memory at once.

Library/Contributions/pre-push
```diff
@@ -0,0 +1,47 @@
...
+# To use, `ln -s $(brew --repository)/Library/Contributions/pre-push
+# $(brew --repository)/.git/hooks`.
+
+import re
+import os
+import subprocess
+import sys
+
+FORMULA_RE = re.compile(r"^(Library/Formula/)?([A-Za-z0-9-]+)\.rb")
+TESTBOT_EMAILS = ["brew-test-bot@googlegroups.com"]
```
bfontaine: Since you’re going to use this to test for inclusion it might be better to use a `set()`.

Library/Contributions/pre-push
tdsmith: It's intentional, to allow both "homebrew" and "Homebrew."

bfontaine: Thanks, that’s what I thought.

xu-cheng: Maybe `"homebrew/homebrew" not in remote_ur.lower()` to make it more readable.

Library/Contributions/pre-push
tdsmith: True, but the input will be a single line in the typical case and I like the list unpacking idiom (`for x, y, z in list_of_lists:`) though I'm willing to hear that I shouldn't

Library/Contributions/pre-push
tdsmith: git promises to invoke it with two arguments but that's reasonable

Library/Homebrew/cmd/pull.rb
tdsmith: Works for me but I thought it might be too surprising.

xu-cheng: To make less surprising, we could add `brew pull --install-push-hook` or something?

Library/Homebrew/cmd/pull.rb
```diff
@@ -189,8 +189,14 @@ def pull
               "-u#{bintray_user}:#{bintray_key}", "-X", "POST",
               "https://api.bintray.com/content/homebrew/#{repo}/#{package}/#{version}/publish"
             puts
-            sleep 20
-            safe_system "brew", "fetch", "--retry", "--force-bottle", f.full_name
+          end
+          unless (Pathname.pwd/".git/hooks/pre-push").exist?
```
xu-cheng: `Pathname.pwd` should be `HOMEBREW_REPOSITORY`. So we can run `brew pull` anywhere.

tdsmith: That doesn't work from taps; there's a chdir earlier in pull.

Library/Contributions/pre-push
tdsmith: That is nicer; thanks.

Library/Contributions/pre-push
```diff
@@ -0,0 +1,50 @@
...
+import re
+import os
+import subprocess
+import sys
+
+FORMULA_RE = re.compile(r"^(Library/Formula/)?([A-Za-z0-9-]+)\.rb")
+TESTBOT_EMAILS = {"brew-test-bot@googlegroups.com"}
+
+def git(*args):
+    return subprocess.check_output(["git"] + list(args)).strip()
```
bfontaine: Is there any reason for the `list()` here? Isn’t `args` already a list?

tdsmith: I felt the same way but `*args` gives you a tuple and you can't add a list and a tuple. You _can_ add two tuples, though, so this should be `("git",)` instead of `["git"]`.

mikemcquaid:
> I have a terrible feeling you're going to ask me to rewrite this in Ruby. Call it a strawman for discussion, anyway.

Yeh, sorry. Either that or maybe Bash.

mikemcquaid:
Could we consider just polling in `brew pull` instead? This feels like it's perhaps overkill considering the issue was just my pushing once impatiently and ignoring Bintray?

tdsmith:
Sounds like better solutions are in the pipeline.

PR #41486, opened 2015-07-08
Xcode 6.4

DomT4:


mikemcquaid:
:+1: bots should already be updated

DomT4:
I meant to do this the other night after you confirmed you'd updated the bot, but I completely forgot :sweat_smile:.

PR #41490, opened 2015-07-08
Updating to 2.0.4

dalanmiller:
Updating to 2.0.4 as well as updating our description.


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/313d697dfea56c48ff7270033fd7a826703a8d29. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a nit, but we're transitioning everything over to `sha256` so if you could use that in future that'd be awesome, Thanks!

PR #41492, opened 2015-07-08
Update liquibase version to 3.4.0

madorb:


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/14134c56c8125079072de7ffb2e771f45fef68ac. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Couple of future nits: We're moving everything to `sha256` so please use that in future, and commit message style for version bump should match `liquibase 3.4.0` or similar, Thanks!

PR #41493, opened 2015-07-08
README: use long key ID instead of short key ID

viktorahlstrom:
Just decided to open a PR about this because it seems a bit better to use long instead of short key-id.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Second, read the [Troubleshooting Checklist](https://github.com/Homebrew/homebre
 Please report security issues to security@brew.sh.
 
 This is our PGP key which is valid until June 17, 2016.
-* Key ID: `E33A3D3CCE59E297`
+* Key ID: `0xE33A3D3CCE59E297`
 * Fingerprint: `C657 8F76 2E23 441E C879  EC5C E33A 3D3C CE59 E297`
 * Full key: https://keybase.io/homebrew/key.asc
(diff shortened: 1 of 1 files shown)


mikemcquaid:
CC @xu-cheng @DomT4 for thoughts

DomT4:
I'd probably make it `0xE33A3D3CCE59E297`.

On the command-line you can use short key, long key, very long key, it doesn't care as long as you don't search for the email address.

The fun with website keyservers where most people will probably head for validation is that you usually can't search for keys with the short key string, `CE59E297` returns nothing, as does `E33A3D3CCE59E297`. Using the email address directly works fine.

If you want to search for the key using the key itself rather than the email address you have to go the whole `0xE33A3D3CCE59E297`, but just to be confusing, after you locate it and click through it only presents the very short key:

screen shot 2015-07-08 at 23 27 36

mikemcquaid:
> I'd probably make it 0xE33A3D3CCE59E297.

:+1:

viktorahlstrom:
@DomT4 Changed to `0xE33A3D3CCE59E297`.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/c1b48b5878f369426ae988d0d7aff6242190f896. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #41494, opened 2015-07-08
vim 7.4.769

nijikon:


DomT4:
Let's keep this using `https://mirrors.kernel.org/debian/pool/main/v/vim` for now. The Git snapshot mirror may [vanish at some point](https://github.com/vim/vim/pull/365#issuecomment-118613068). Using Debian's mirrors allows us not to stumble into that surprise.

mikemcquaid:
> Let's keep this using https://mirrors.kernel.org/debian/pool/main/v/vim for now. The Git snapshot mirror may vanish at some point. Using Debian's mirrors allows us not to stumble into that surprise.

:+1:

nijikon:
No problem, will wait for the source file at debian site.

DomT4:
Going to pass on this for now. We track https://packages.debian.org/unstable/vim which doesn't look like it's going to be updated super-soon.

PR #41498, opened 2015-07-09
szl 1.0.1 (new maintainer)

aecolley:
I fixed many bugs in szl, but Google has abandoned it and no longer
responds to bug reports or messages on szl-users. So I took it over.

I sent an earlier version of this formula to Adam Vandenberg, whom I found
in the git commit history. But he said "Please send this to the Homebrew
mailing list, not me directly; or open a ticket with this information. I
don't maintain this formula." I'll maintain it, then.

This formula is cleaner (no patches), more stable (the code is more
portable) and more useful (bugs are fixed) than the previous one.

### DIFF
--- a/Library/Formula/szl.rb
+++ b/Library/Formula/szl.rb
@@ -1,79 +1,49 @@
 class Szl < Formula
-  desc "Compiler and runtime for the Sawzall language"
-  homepage "https://code.google.com/p/szl/"
-  url "https://szl.googlecode.com/files/szl-1.0.tar.gz"
-  sha1 "e4c6d4aec1afc025257d41dd77b8f5c25ea120d4"
-  revision 5
+  desc "Google's Sawzall language for statistical aggregation of log data."
+  homepage "https://bitbucket.org/aecolley/szl"
+  url "https://bitbucket.org/aecolley/szl/downloads/szl-1.0.1.tar.gz"
+  sha1 "9a5f0f1914781a7ca8f80896124c1a39bae329f2"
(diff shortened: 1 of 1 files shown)


Library/Formula/szl.rb
```diff
@@ -1,79 +1,49 @@
...
-  # https://code.google.com/p/szl/issues/detail?id=34
-  patch :DATA
+  # We don't need gobjdump for core functionality, but only for
+  # "szl --print_code --native". We can use it if it's installed later.
+  depends_on "binutils" => :optional
 
   def install
-    ENV["OBJDUMP"] = "#{HOMEBREW_PREFIX}/bin/gobjdump"
+    if build.head?
+      system "./autogen.sh", "-f"
```
dunn: This can be just one line `system "./autogen.sh", "-f" if build.head?`

DomT4:
We're usually *very* reluctant to accept personal forks of a project unless upstream have signed off on the fork as the new development location. That's unlikely to happen with Google, which could be tricky.

We tend to be reluctant because it can surprise users to discover they thought they had one thing installed and now have a completely different fork installed with none of the original developers/etc.

Will leave it open for other owners here to review/discuss. If we choose to accept this it may end up going in under a new name rather than taking over the old one.

mikemcquaid:
> We're usually very reluctant to accept personal forks of a project unless upstream have signed off on the fork as the new development location. That's unlikely to happen with Google, which could be tricky.

That's the case here, I'm afraid.

aecolley:
That's understandable. I'd have the same response if I were in your position. I thank you for providing the facility of Taps, which makes installing my szl almost as easy as if you'd accepted this pull request.

DomT4:
Thanks for understanding, We appreciate that.

PR #41500, opened 2015-07-09
cmake: add --devel (3.3.0-rc3); patch 3.2.3 only

L2G:


Library/Formula/cmake.rb
```diff
@@ -45,12 +50,13 @@ class Cmake < Formula
...
 
+  # Fix for older bash-completion versions.  This can be removed when 3.3.0 is
+  # released.
   patch do
-    # fix for older bash-completion versions
     # Use Github because upstream git changes sha and breaks from-source download.
     url "https://github.com/Kitware/CMake/commit/2ecf168f1909.diff"
     sha256 "d3f8cd71d0b6ce23a22c55145114012da916f2e42af71cbbad35090d0aeb4f68"
-  end
+  end if build.stable?
```
DomT4: I kind of strongly, deeply dislike this format of patching. Let's do a stable block instead.

mikemcquaid: :+1:

L2G: Sorry. I had seen other formulae doing that, and for some reason I got it in my head that that was The Thing To Do. Now I see there are only about 9 other formulae doing that type of thing.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/30ecfcbc35a6a22e1644b6e0bd7f72678949df3a. Thanks @L2G.

Commit message for devel additions should ideally be `cmake 3.3.0-rc3 (devel)` or similar :)

PR #41480, opened 2015-07-08
rds-command-line-tools: install credential-file-path.template to etc directory

kyanny:
- Install `credential-file-path.template` file to etc directory
- Add explanation about this file to caveats section

PR #41503, opened 2015-07-09
Remove keys from comment

banyan:
Follow up to https://github.com/Homebrew/homebrew/pull/41480

If I set `AWS_ACCESS_KEY` and `AWS_SECRET_KEY` but without `AWS_CREDENTIAL_FILE`, it doesn't work. Eventually we have to set  `AWS_ACCESS_KEY` and `AWS_SECRET_KEY` in `AWS_CREDENTIAL_FILE`.  So this instruction seems to be redundant and useless.

### DIFF
--- a/Library/Formula/rds-command-line-tools.rb
+++ b/Library/Formula/rds-command-line-tools.rb
@@ -21,9 +21,7 @@ class RdsCommandLineTools < Formula
 
   def caveats
     <<-EOS.undent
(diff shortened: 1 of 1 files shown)


mikemcquaid:

For future reference the preferred commit message is `rds-command-line-tools: remove keys from comment.`.

banyan:
>For future reference the preferred commit message is rds-command-line-tools: remove keys from comment..

Ah I see. Thanks!

PR #41506, opened 2015-07-09
download_strategy: private method allowing extra curl options.

indrajitr:
This allows some curl options to be always passed to curl, with raw
head calls (`curl -I`) (in `CurlDownloadStrategy#actual_urls`) or with
actual `curl` (in `CurlDownloadStrategy#curl`).

This also avoid the need for overriding whole `_fetch` in a few cases
and just override `_curl_opts` to append additional options instead.

NB: As suggest by @mikemcquaid, this is initial stab at fixing [this]( https://github.com/Homebrew/homebrew/commit/7822f70c06ead5189dd0d308f3550f9924656537#commitcomment-12052731)

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -323,9 +323,18 @@ class CurlDownloadStrategy < AbstractFileDownloadStrategy
     curl @url, "-C", downloaded_size, "-o", temporary_path
   end
 
+  # Curl options to be always passed to curl,
+  # with raw head calls (`curl -I`) or with actual `fetch`.
+  def _curl_opts
+    copts = []
+    copts << "--user" << meta.fetch(:user) if meta.key?(:user)
+    copts
+  end
+
   def actual_urls
     urls = []
-    Utils.popen_read("curl", "-I", "-L", @url).scan(/^Location: (.+)$/).map do |m|
+    curl_args = _curl_opts << "-I" << "-L" << @url
+    Utils.popen_read("curl", *curl_args).scan(/^Location: (.+)$/).map do |m|
       urls << URI.join(urls.last || @url, m.first.chomp).to_s
     end
     urls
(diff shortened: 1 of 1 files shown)


xu-cheng:
cc @Homebrew/owners. LGTM :+1:

mikemcquaid:
Makes sense to me.

mikemcquaid:

indrajitr:
Wow, this was _quick_. Thank you all for running this project so efficiently!

DomT4:
Is the line on 403 breaking the SSL3 download? It seems to have removed the `'-C', downloaded_size, '-o', temporary_path` part?

Only glanced over this for 5 seconds, feel free to say if I'm being an idiot.

mikemcquaid:
@DomT4 SSL3 download? Can you give a test case?

indrajitr:
`CurlSSL3DownloadStrategy` doesn't override `#_fetch` anymore, instead it just overrides `#_curl_opts` and appends the necessary options. `#_fetch` in `CurlDownloadStrategy` does the rest.

@DomT4, any concern with the approach?

DomT4:
@mikemcquaid To be honest I'm not aware of anything that isn't at least capable of TLSv1 these days. In an ideal world I'd tag it as `@deprecated` and remove it in 6 months or something.

> any concern with the approach?

Nope, as long as it works. My concern was just a presumption that we'd deliberately kept them apart in the past for some reason, but I hadn't looked into it deeply yet.

indrajitr:
Agreed on deprecating, I too assumed there might have been an use case somewhere.

>> any concern with the approach?

> Nope, as long as it works. My concern was just a presumption that we'd deliberately kept them apart in > the past for some reason, but I hadn't looked into it deeply yet.

Ok, I just took that as a demonstrable use case where `#_curl_opts` would be useful in DRY-ing up without overriding `#_fetch`.

DomT4:
I think we're fine to just remove it now, to be honest. SSLv3 has breakage all over the shop, and we should find alternatives for servers & formulae that rely on SSLv3 to fetch exclusively (There are none in the core). Opened #41536.

indrajitr:
:+1:

PR #41507, opened 2015-07-09
qjson: add HEAD

dunn:
There hasn't been a stable release in 3 years but it's still being
developed so HEAD might be useful.


xu-cheng:
Thanks @dunn

PR #41512, opened 2015-07-09
Remove datomic to homebrew-binary

zimbatm:
There is no compilation step for this-one.

I have also submitted a PR to https://github.com/Homebrew/homebrew-binary/pull/239


mikemcquaid:
We allow binary Java packages because they work cross-platform due to Java's bytecode.

PR #41516, opened 2015-07-09
audit: add --online flag.

mikemcquaid:
As discussed. Also add it to the manpage and brew-test-bot.

CC @Homebrew/owners for review.

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -18,6 +18,8 @@ module Homebrew
       style
     end
 
+    online = ARGV.include? "--online"
+
     ENV.activate_extensions!
     ENV.setup_build_environment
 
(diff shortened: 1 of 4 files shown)


Library/Homebrew/cmd/audit.rb
```diff
@@ -423,7 +426,7 @@ def audit_homepage
   end
 
   def audit_github_repository
-    return unless @strict
+    return unless @online
```
xu-cheng: I think this would be better as `return unless @online && @strict`. The reason is these two flags have different semantical meaning.

xu-cheng:
[This part](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/audit.rb#L237-L239) should be controlled by `--online` flag as well.

Library/Homebrew/cmd/audit.rb
mikemcquaid: That feels a bit weird to me to have `--online` do nothing without `--strict` being passed. I can see us wanting to add some non-strict `--online` tests?

mikemcquaid:
:+1:

PR #41518, opened 2015-07-09
sjk 0.3.6 (new formula)

dennisoelkers:
This recipe adds the "Swiss Java Knife", which is a very handy tool for every Java developer.

### DIFF
--- a/Library/Formula/sjk.rb
+++ b/Library/Formula/sjk.rb
@@ -6,14 +6,16 @@ class Sjk < Formula
 
   depends_on :java
 
-  resource "shell_wrapper" do
-    url "https://gist.githubusercontent.com/dennisoelkers/2bda4937ab823ea54d44/raw/87af0aa7510afea4022a0296f2c93cc1c0a2bee0/sjk.sh"
-    sha256 "42ecaa01e781553c36526485efc9702c9c2138c948f328513e22e06a74962007"
-  end
-
   def install
     lib.install "sjk-plus-0.3.6.jar"
-    resource("shell_wrapper").stage { bin.install "sjk.sh" => "sjk" }
+    (bin/"sjk").write <<-EOS.undent
+      #!/bin/sh
+
+      PREFIX=`dirname $0`/../lib
+      JARFILE=$PREFIX/sjk-plus-0.3.6.jar
+
+      java -jar $JARFILE $@
+    EOS
   end
 
   test do
--- a/Library/Formula/sjk.rb
+++ b/Library/Formula/sjk.rb
@@ -7,15 +7,8 @@ class Sjk < Formula
   depends_on :java
 
   def install
-    lib.install "sjk-plus-0.3.6.jar"
-    (bin/"sjk").write <<-EOS.undent
-      #!/bin/sh
-
-      PREFIX=`dirname $0`/../lib
-      JARFILE=$PREFIX/sjk-plus-0.3.6.jar
-
-      java -jar $JARFILE $@
-    EOS
+    libexec.install "sjk-plus-#{version}.jar"
+    bin.write_jar_script "#{libexec}/sjk-plus-#{version}.jar", "sjk"
   end
 
   test do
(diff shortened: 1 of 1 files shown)


Library/Formula/sjk.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def install
+    bin.mkdir
+    lib.install "sjk-plus-0.3.6.jar"
+    resource("shell_wrapper").stage { bin.install "sjk.sh" => "sjk" }
+  end
+
+  resource "shell_wrapper" do
+	  url "https://gist.githubusercontent.com/dennisoelkers/2bda4937ab823ea54d44/raw/87af0aa7510afea4022a0296f2c93cc1c0a2bee0/sjk.sh"
+	  sha256 "42ecaa01e781553c36526485efc9702c9c2138c948f328513e22e06a74962007"
```
mikemcquaid: two space indent and move the `resource` above `install`, thanks.

dennisoelkers: I moved the resource block. Indentation looks fine in my editor though?

mikemcquaid: Perhaps tabs vs spaces.

bfontaine: Yes there’s a mix of tabs and spaces here.

Library/Formula/sjk.rb
```diff
@@ -0,0 +1,24 @@
+class Sjk < Formula
+  desc "Swiss Java Knife"
+  homepage "https://github.com/aragozin/jvm-tools"
+  url "https://bintray.com/artifact/download/aragozin/generic/sjk-plus-0.3.6.jar"
+  version "0.3.6"
```
bfontaine: Homebrew can detect the version from the URL, so you can remove this line.

Library/Formula/sjk.rb
```diff
@@ -0,0 +1,24 @@
...
+  sha256 "9420403139c1b843320fe07bac56f704b0d13715d53b5b2b5869d32103a99a47"
+
+  depends_on :java
+
+  resource "shell_wrapper" do
+    url "https://gist.githubusercontent.com/dennisoelkers/2bda4937ab823ea54d44/raw/87af0aa7510afea4022a0296f2c93cc1c0a2bee0/sjk.sh"
+    sha256 "42ecaa01e781553c36526485efc9702c9c2138c948f328513e22e06a74962007"
+  end
```
bfontaine: This is a really small file, why not including it in the formula?

```
#!/bin/sh
 
PREFIX=`dirname $0`/../lib
JARFILE=$PREFIX/sjk-plus-0.3.6.jar
 
java -jar $JARFILE $@
```

Library/Formula/sjk.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def install
+    bin.mkdir
```
bfontaine: You can remove this line, `bin` will be created by the `bin.install` call below.

Library/Formula/sjk.rb
dennisoelkers: I was thinking about that too, but this is my first formula and I didn't know how to do it. What's the best way to do it? Define an inline heredoc and then?

dennisoelkers: The added benefit of this would be that we could add the version number (or even the whole name) to $JARFILE, so the file does not need to change when the version is bumped.

bfontaine: In fact thinking about it we have a helper for that:

```
bin.write_jar_script "#{libexec}/sjk-plus-#{version}.jar", "sjk"
```

Library/Formula/sjk.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  depends_on :java
+
+  resource "shell_wrapper" do
+    url "https://gist.githubusercontent.com/dennisoelkers/2bda4937ab823ea54d44/raw/87af0aa7510afea4022a0296f2c93cc1c0a2bee0/sjk.sh"
+    sha256 "42ecaa01e781553c36526485efc9702c9c2138c948f328513e22e06a74962007"
+  end
+
+  def install
+    lib.install "sjk-plus-0.3.6.jar"
```
bfontaine: Please install in `libexec` rather than `lib`.

bfontaine: …and use `#{version}` instead of `0.3.6` to make it future-proof :wink:

dennisoelkers: Thanks for your great help!

mikemcquaid:

For future reference the preferred commit message format for new formulae is `sjk 0.3.6 (new formula)`.

PR #41519, opened 2015-07-09
shell completion: support external commands

xu-cheng:
I only test this under zsh. Would be nicer if someone tests it for bash as well. Thanks. 
cc @Homebrew/owners

### DIFF
--- a/Library/Contributions/brew_bash_completion.sh
+++ b/Library/Contributions/brew_bash_completion.sh
@@ -580,62 +580,7 @@ _brew ()
     done
 
     if [[ $i -eq $COMP_CWORD ]]; then
-        __brewcomp "
-            --cache --cellar
-            --env --prefix --repository
-            aspell-dictionaries
-            audit
-            bottle
-            cat
-            cleanup
-            commands
-            config --config
-            create
-            deps
-            diy configure
-            doctor
-            edit
-            fetch
-            gist-logs
-            help
-            home
-            info abv
-            install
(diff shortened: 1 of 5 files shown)


bfontaine:
Seems to be broken for Bash. Without these changes if I press <kbd>Tab</kbd> after `brew` it’ll show me all the possibilities. With these it only gives me `--cache`.

Library/Contributions/brew_bash_completion.sh
```diff
@@ -580,62 +580,7 @@ _brew ()
...
-            unlink
-            unlinkapps
-            unpack
-            unpin
-            untap
-            update
-            upgrade
-            uses
-            "
+        __brewcomp $(brew commands --quiet --include-aliases)
```
bfontaine: Add quotes around `$(…)`:

```sh
__brewcomp "$(brew commands --quiet --include-aliases)"
```

xu-cheng: Fixed, Thanks @bfontaine

xu-cheng:
Any other comments from other maintainers?

mikemcquaid:
If it works in both: :+1:

xu-cheng:
It works, thought I think you may have comments on new `brew commands` flags and possible manpage change.

DomT4:
As long as they work, pretty much always :+1: on expanding the completions. The new flags look fine.

PR #41521, opened 2015-07-09
add option for "Install all input-plugins"

kkdd:
option "with-all-input-plugins", "Install all input-plugins"

### DIFF
--- a/Library/Formula/mapnik.rb
+++ b/Library/Formula/mapnik.rb
@@ -24,8 +24,6 @@ class Mapnik < Formula
   depends_on "postgresql" => :optional
   depends_on "cairo" => :optional
 
-  option "with-all-input-plugins", "Install all input-plugins"
-
   if MacOS.version < :mavericks
     depends_on "boost" => "c++11"
   else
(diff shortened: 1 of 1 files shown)


Library/Formula/mapnik.rb
```diff
@@ -68,6 +70,10 @@ def install
             "CPP_TESTS=False" # too long to compile to be worth it
            ]
 
+     if build.with? "all-input-plugins""
+      args << "INPUT_PLUGINS=all"
+     end
```
bfontaine: You can write this on one line:

```ruby
args << "INPUT_PLUGINS=all" if build.with? "all-input-plugins"
```

mikemcquaid:
Any downside to installing these by default?

kkdd:
I can't find any side effects when installing all input-plugins of mapnik. It seems OK for installing these by default.

mikemcquaid:
@kkdd Let's just enable it by default then, thanks!

mikemcquaid:
(and remove the `option` entirely)

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/9e6c76196056a09c424f0626e023c877b8e0ecde. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a nit, but non-version formula changes should be styled along the lines of `mapnik: add extra dolphins` or similar, Thanks!

PR #41523, opened 2015-07-09
audit: deprecate sha1 always.

mikemcquaid:
Rather than just when --strict is set. We're asking people to do these on most PRs now anyway so feels better to let them find this out with `brew audit`. CC @Homebrew/owners for thoughts


bfontaine:
:+1:

tdsmith:
No objections.

DomT4:
:+1:

PR #41461, opened 2015-07-07
Update percona-server to 5.6.25-73.0

wolfd:
The old tarball seems to be gone.

PR #41525, opened 2015-07-09
Less fragile Percona link

jhaber:
The fixed link in #41461 is already broken again because it was using LATEST which is constantly changing. By using the versioned path the link should work more reliably

### DIFF
--- a/Library/Formula/percona-server.rb
+++ b/Library/Formula/percona-server.rb
@@ -3,7 +3,7 @@ require 'formula'
 class PerconaServer < Formula
   desc "Drop-in MySQL replacement"
   homepage 'https://www.percona.com'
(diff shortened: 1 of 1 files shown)


DomT4:
@BrewTestBot test this please

mikemcquaid:
@DomT4 Think it's ignoring us again, not sure why.

DomT4:
Hmm, yeah, it hasn't rejoined the queue.

DomT4:
@BrewTestBot test this please

mikemcquaid:
@BrewTestBot test this please

mikemcquaid:

For future reference the preferred commit message format is `percona: use less fragile link.`.

PR #41526, opened 2015-07-09
mapnik: use cellar :any

springmeyer:
Fix bottle-based install for Mapnik

### DIFF
--- a/Library/Formula/mapnik.rb
+++ b/Library/Formula/mapnik.rb
@@ -6,6 +6,7 @@ class Mapnik < Formula
   sha256 "c88b1ce48899ffe0d75f9e753dcd427e2b6fd3186b40e04508608b2151c0e7b0"
 
   bottle do
+    cellar :any
     sha256 "bcf0efd84fc80d8db0b8f78204ecc1b36221bf53994cad19d6c5ce0083668178" => :yosemite
     sha256 "160bf811a9148c076c28cf72014aadf6994e46373cf7b83fc79e143f2fd33dcd" => :mavericks
     sha256 "b60f792870191f6a7b32a628d81120429471e4b24bb589063ea057fba5ad2c5d" => :mountain_lion


mikemcquaid:
Sorry, this isn't correct. The bottle contains hardcoded references to the installed prefix so it won't work unless you have Homebrew installed in `/usr/local`.

springmeyer:
> The bottle contains hardcoded references to the installed prefix

@mikemcquaid - thanks for the details and [this clarification](https://github.com/Homebrew/homebrew/pull/41474#issuecomment-120011987). Specifically what allows the freetype.rb formula to work okay with `cellar: any`? Are there any changes we could make upstream in Mapnik to potentially enable the bottle to work on a custom path?

mikemcquaid:
@springmeyer Sure. Basically it's looking for hardcoded references to the Cellar. You can `brew install mapnik --build-bottl && brew bottle -v mapnik` to see what these references are.

springmeyer:
@mikemcquaid - thanks, I ran that command. It outputs this https://gist.github.com/springmeyer/f8f794cb5d03eb7d74ea. That looks - to me - completely harmless. So I'm missing what might be wrong with using `cellar: any`. I'm totally new to how bottles work, but it seems like there are two types of warnings:

> Warning: String '/Users/dane/.homebrew' still exists in these files:
> /Users/dane/.homebrew/Cellar/mapnik/3.0.0/INSTALL_RECEIPT.json
>  --> match '{"used_options":[],"unused_options":["--with-gdal","--with-postgresql","--with-cairo"],"built_as_bottle":true,"poured_from_bottle":false,"time":1436457250,"HEAD":"434c3fc9764f81ba383f3df3ff3b1a84cba47f85","stdlib":"libcxx","compiler":"clang","source":{"path":"/Users/dane/.homebrew/Library/Formula/mapnik.rb","tap":"Homebrew/homebrew"}}' at offset 0xffffffff

That ^^ looks like the path to the formula is being found? I don't see how that could be a problem in using the package.

> /Users/dane/.homebrew/Cellar/mapnik/3.0.0/lib/libmapnik.dylib
> --> match '/Users/dane/.homebrew/opt/boost/include/boost/property_tree/detail/xml_parser_write.hpp' at offset 0x7eaf4e
> [ ... snip ... ]

That ^^ looks spurious. A boost header surely cannot have the `'/Users/dane/.homebrew'` string inside it. Any idea why that is being flagged?

springmeyer:
@mikemcquaid - any insight you have would be greatly appreciated. I've love to help fix these issues if they are something on the Mapnik side. But unless you think otherwise they seem like non-problems and that `cellar :any` for mapnik should be fine. Am I missing anything?

tdsmith:
> A boost header surely cannot have the '/Users/dane/.homebrew' string inside it. Any idea why that is being flagged?

libmapnik.dylib contains the path to the header. It's probably harmless, but that's difficult to detect.

You can ignore the INSTALL_RECEIPT.json warning; it is not flagged when Homebrew is installed to the default prefix, which is how the bots are configured.

springmeyer:
@tdsmith thanks

> libmapnik.dylib contains the path to the header. It's probably harmless, but that's difficult to detect.

I'm positive it is harmless. I've be building binaries of libmapnik.dylib/so for years and distributing across systems and never had a problem.

> You can ignore the INSTALL_RECEIPT.json warning; it is not flagged when Homebrew is installed to the default prefix, which is how the bots are configured.

Okay, thanks.

@mikemcquaid - can you please re-consider and apply my pull?

mikemcquaid:
@springmeyer Unfortunately we don't have a way of flagging bottles like this in a "sticky" manner i.e. if we just merge this PR it'll get reverted automatically on the next bottle upgrade. We need a way of relocating that string and/or fixing our detection, I'm afraid.

@Homebrew/owners Any thoughts here? I definitely would like to work out a way we can make more things `cellar :any`.

springmeyer:
@mikemcquaid I see, okay.

So I've done more digging into where these paths are coming from. The reason is Mapnik uses boost (specifically `boost::property_tree`) and `boost::property_tree` uses `BOOST_THROW_EXCEPTION`. That macro [outputs the line number when it throws](http://www.boost.org/doc/libs/1_37_0/libs/exception/doc/BOOST_THROW_EXCEPTION.html):  I've never seen this throw, but clearly this is the cause. So it seems like any homebrew formula that uses Boost may suffer from this problem.

springmeyer:
New pull request at #41583 which fixes the problem for Mapnik.

PR #41527, opened 2015-07-09
openssl 1.0.2d

DomT4:
With the dawning of OS X 10.11 Apple has seemingly removed the OpenSSL headers from public availability. This has meant that unless we specifically point things at our OpenSSL, or our OpenSSL is in the `PATH` things are failing to build in unusual ways.

To work around this, we are removing the `keg_only` status on OpenSSL. This means from today OpenSSL will symlink into your `HOMEBREW_PREFIX` and be available for all and any thing to compile against, and presuming you have `HOMEBREW_PREFIX` at the front of your `PATH` will be used over the older system OpenSSL.

Things that compile against Homebrew's OpenSSL but don't have an OpenSSL `depends_on` should still be reported to Homebrew. We will need to fix the link so our bottles rely on our OpenSSL rather than linking the system's still, and it allows us to revision things when the security landscape dictates that it is necessary.

There are no plans to unkeg LibreSSL at this time. It conflicts with the other OpenSSL versions and at the moment there's still a lot that won't build against it. If that changes in future we aren't rigid and unbending towards useful change, but for now don't file PRs asking for LibreSSL to be unkegged, Thanks!

Please report any issues you discover to Homebrew and we'll endeavour to fix them as quickly and thoroughly as possible.

### DIFF
--- a/Library/Formula/openssl.rb
+++ b/Library/Formula/openssl.rb
@@ -17,9 +17,6 @@ class Openssl < Formula
 
   depends_on "makedepend" => :build
 
-  keg_only :provided_by_osx,
-    "Apple has deprecated use of OpenSSL in favor of its own TLS and crypto libraries"
-
   def arch_args
     {
       :x86_64 => %w[darwin64-x86_64-cc enable-ec_nistp_64_gcc_128],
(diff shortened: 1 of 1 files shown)


mikemcquaid:
The `audit` is failing but that should be sorted by https://github.com/Homebrew/homebrew/pull/41528 so feel free to merge this while red.

tdsmith:
> This has meant that unless we specifically point things at our OpenSSL, or our OpenSSL is in the PATH things are failing to build in unusual ways.

Why isn't this desirable?

DomT4:
We can't possibly find and patch everything in Homebrew that needs OpenSSL for something, and things compiled outside of Homebrew would just fail all the time without people passing flags. The general idea was that making `keg_only` go away would be nicer UX in general.

Clang disagreed, apparently.

PR #41528, opened 2015-07-09
formula_cellar_checks: ignore openssl.

mikemcquaid:
Also: check based on the beginning of the formula name so these play nicer with e.g. homebrew-versions.


DomT4:
Was just thinking about the need to do this, :+1:.

PR #41530, opened 2015-07-09
pull: wait for Bintray publish, don't sleep.

mikemcquaid:
Doesn't seem to work yet but if we can figure out why that'd be handy CC @DomT4 @tdsmith

### DIFF
--- a/Library/Homebrew/cmd/pull.rb
+++ b/Library/Homebrew/cmd/pull.rb
@@ -185,11 +185,10 @@ module Homebrew
             ohai "Publishing on Bintray:"
             package = Bintray.package f.name
             version = f.pkg_version
-            curl "--silent", "--fail",
+            curl "-w", '\n', "--silent", "--fail",
               "-u#{bintray_user}:#{bintray_key}", "-X", "POST",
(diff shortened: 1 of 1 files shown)


DomT4:
It'll probably get lost on the chat, so for reference: https://bintray.com/docs/api/#_publish_discard_uploaded_content

mikemcquaid:
@DomT4 could you point bintray to this and let them know it's not working?

DomT4:
Sure, will do.

DomT4:
screen shot 2015-07-09 at 21 50 11

mikemcquaid:
@DomT4 The script returns a `{"files":3}` but the files aren't published so the feature is working as it was before the documentation.

mikemcquaid:
Just pulled this with the sleep if it doesn't work (it never seems to)

DomT4:
Hmm. Yeah, they do publish, but still with the delay there that requires a sleep. Still trying to debug this with Bintray; Reproduced the error we see with uploads and then immediate fetches and sent them verbose logs of the upload, as well as verbose logs of the curl command we use to retrieve packages.

PR #41537, opened 2015-07-09
mono: Fix compile/runtime error on El Cap Beta 3

felixbuenemann:
This fixes the error "could not discover the mach TLS offset" which
prohibits both running and compiling mono on OS X 10.11 (15A216g).

The patch was backported from current mono/mono master. Because the
problem also affects binaries compiled before upgrading to El Capitan
the patch is *not* scoped to `MacOS.version >= :el_capitan`.

See [Xamarin Bugzilla Bug 31761](https://bugzilla.xamarin.com/show_bug.cgi?id=31761) and PR mono/mono#1919 for more info.

### DIFF
--- a/Library/Formula/mono.rb
+++ b/Library/Formula/mono.rb
@@ -14,6 +14,14 @@ class Mono < Formula
     sha256 "80bddaefcd9c136bc763797cff98f3aa07ec758e348037fd24f07d2b1b795fbb" => :mountain_lion
   end
 
+  # Fix compile and runtime error on OS X 10.11 Beta 3
+  # https://github.com/mono/mono/pull/1919
+  # https://bugzilla.xamarin.com/show_bug.cgi?id=31761
+  patch do
+    url "https://github.com/mono/mono/pull/1919.diff"
+    sha256 "53c39c2145027fdf1a2233e12bd96da2c1164e1422e631cdb50598910b39a020"
+  end
+
   resource "monolite" do
     url "http://storage.bos.xamarin.com/mono-dist-4.0.0-release/c1/c1b37f29b1a439acf7ef42a384550ab1dca5295a/monolite-117-latest.tar.gz"
     sha256 "a3bd1c826186e4896193ad1f909bf8756f66f62d1e249fe301b10bc80ebe0795"


felixbuenemann:
Is there a reason that the brew test-bot is requesting access to my github account's associated organizations and it's repos when I click on the Details link?

DomT4:
We have it set at the least amount of permissions we're told the plugins require, I believe. CC @mikemcquaid

andrewbranch:
You, sir, are a life saver.

mikemcquaid:
@felixbuenemann screen shot 2015-07-09 at 23 59 01

The only thing it requests is to see what organisation teams you're a member of (to see if you're in the Homebrew owners team).

jonlbauer:
Hi All,

So I'm a total newbie, but I just encountered this issue with the Public Beta release of El Capitan...

I have the newest mono and the 3.10.0 versions installed, and I get this error and a freeze up on both versions. BTW, I installed 4.0.2.5 from scratch, and the mono installer didn't set up a path for me either. So that's broken too...

Can someone please tell me, in plain English :-) how I can fix it?

- Thanx in advance!
- Jon

felixbuenemann:
@mikemcquaid Thanks for the clarification. I thought it was requesting access to organization repos.

felixbuenemann:
@jonlbauer The compile will fail if there is an unfixed version of mono in your path. In the normal update case the old homebrew version is unlinked before the compile is started, so that's not an issue. But if you have multiple mono versions installed that might be the problem.

jonlbauer:
Hm. I don't know how to do a compile on my own... Can you summarize the process for me? I"m guessing that I need to download the fix that is summarized here, merge it into the version that I have, and then re-compile it?

- Jon

jonlbauer:
Oh wait a sec - are you saying that I should uninstall 3.10.0 and then try the normal install method? But what about the fix that is being outlined here? I don't need to worry about that?

- Jon

felixbuenemann:
@jonlbauer I think you can install from a url, try `brew install https://raw.githubusercontent.com/felixbuenemann/homebrew/fix-mono-el-cap-tls-offset/Library/Formula/mono.rb`

jonlbauer:
That didn't work.

If I install home-brew, and then install mono through it, will it be using the fix that's been outlined here?

- Jon

Library/Formula/mono.rb
```diff
@@ -14,6 +15,10 @@ class Mono < Formula
     sha256 "80bddaefcd9c136bc763797cff98f3aa07ec758e348037fd24f07d2b1b795fbb" => :mountain_lion
   end
 
+  # Fix compile and runtime error on OS X 10.11 Beta 3
+  # See: https://github.com/mono/mono/pull/1919
+  patch :DATA
```
mikemcquaid: Please use `https://github.com/mono/mono/pull/1919` as the patch URL, thanks!

Library/Formula/mono.rb
```diff
@@ -3,6 +3,7 @@ class Mono < Formula
   homepage "http://www.mono-project.com/"
   url "http://download.mono-project.com/sources/mono/mono-4.0.2.5.tar.bz2"
   sha256 "b074584eea5bbaaf29362486a69d70abe53d0d2feb334f231fa9c841cf6fd651"
+  revision 1
```
mikemcquaid: Don't think this needs a `revision` given it wasn't installing on 10.11 before.

felixbuenemann: I bumped the revision because mono compiled eg. on yosemite will fail on el cap unless this fix is included. That's also the reason the patch is replied regardless of OS X version.

Library/Formula/mono.rb
felixbuenemann: I didn't think of that. I'll push a new rev, thanks!

felixbuenemann:
@jonlbauer You could install homebrew, uninstall your current mono and install using the command posted above or using `brew install mono` as soon as this pull request is merged. Please take some time to get yourself aquainted with how homebrew works, a pull request is for developer discussions not user support and using it as such causes a lot of noise for the devs involved. Thanks!

Library/Formula/mono.rb
mikemcquaid: I don't think it's worth forcing an upgrade on everyone for that edge-case, thanks.

felixbuenemann: OK, I'll drop the revision.

felixbuenemann:
Pushed a new commit to address @mikemcquaid's feedback.

mikemcquaid:
Thanks @felixbuenemann!

bosmacs:
FYI, you may also want to include the x86 patch for the same problem: 
~~~rb
  # and for x86
  patch do
    url "https://github.com/mono/mono/pull/1924.diff"
    sha256 "62e9d2ad134558c3fcf9c75e260c6091e390b23b4f320e79184136b76c021a39"
  end
~~~

felixbuenemann:
@bosmacs Is there a case where mono would compile as x86 on El Cap?

bosmacs:
I thought it might affect code generation for 32-bit assemblies, but in brief testing that appears not to be the case. So perhaps it’s not needed after all.

PR #41539, opened 2015-07-09
scala: ensure java dependency

indrajitr:
Now that `java` might not be available by default in newer OSX, ensure minimal dependency is met.


mikemcquaid:

PR #41544, opened 2015-07-09
python3: don't run ensurepip during build

tdsmith:
We don't want to run ensurepip, which installs an old version of pip; we
take care of this ourselves in postinstall.

On flat (i.e. non-Framework) builds, ensurepip causes an error in
postinstall because the pip-3.4 script already exists in prefix/bin.

Will resolve Homebrew/linuxbrew#469.


sjackman:
I'll merge Homebrew into Linuxbrew once this PR is merged into Homebrew.

tdsmith:
Done!

PR #41551, opened 2015-07-09
new formula: mapext

xdbr:
map files from one extension to another via a specified command

Note: test includes `|| true` since this command exits with a non-zero exit status and doesn't have an explicit `--help` switch either, rather it displays usage instructions directly if run without arguments.

### DIFF
--- a/Library/Formula/mapext.rb
+++ b/Library/Formula/mapext.rb
@@ -1,0 +1,18 @@
+class Mapext < Formula
+  desc "map files from one extension to another via a specified command"
+  homepage "http://www.acme.com/software/mapext/"
+  url "http://www.acme.com/software/mapext/mapext_10nov1995.tar.gz"
+  version "0.0.19951110"
+  sha256 "7e7a775effed89111d079d39d98b917bcdd3d678d6f434ec5243f9e1e2b41206"
+
+  def install
+    bin.mkpath
+    man1.mkpath
+
+    system "make", "all", "install", "BINDIR=#{bin}", "MANDIR=#{man1}"
+  end
+
+  test do
+    assert_match "usage:", shell_output("#{bin}/mapext 2>&1", 1)
+  end
+end


Library/mapext.rb
```diff
@@ -0,0 +1,17 @@
...
+  desc "map files from one extension to another via a specified command"
+  homepage "http://www.acme.com/software/mapext/"
+  url "http://www.acme.com/software/mapext/mapext_10nov1995.tar.gz"
+  version "10Nov1995"
+  sha256 "7e7a775effed89111d079d39d98b917bcdd3d678d6f434ec5243f9e1e2b41206"
+
+  def install
+    system "make"
+    bin.install "mapext"
+    man1.install "mapext.1"
```
mikemcquaid: Is there no `make install` we cna use?

Library/mapext.rb
```diff
@@ -0,0 +1,17 @@
+class Mapext < Formula
+  desc "map files from one extension to another via a specified command"
+  homepage "http://www.acme.com/software/mapext/"
+  url "http://www.acme.com/software/mapext/mapext_10nov1995.tar.gz"
+  version "10Nov1995"
```
mikemcquaid: Maybe `0.0.19951110`

dunn:
It's hardly better, but maybe we could have the test be something like `assert_match "Usage: [...]", shell_output("#{bin}/mapext", 1)`, where `1` is whatever non-zero exit code it uses. That might make it a little easier to see what's going on.

Library/Formula/mapext.rb
```diff
@@ -0,0 +1,18 @@
+class Mapext < Formula
+  desc "map files from one extension to another via a specified command"
+  homepage "http://www.acme.com/software/mapext/"
+  url "http://www.acme.com/software/mapext/mapext_10nov1995.tar.gz"
+  version "0.0.19951110"
+  sha256 "7e7a775effed89111d079d39d98b917bcdd3d678d6f434ec5243f9e1e2b41206"
+
+  def install
+    mkdir_p bin
```
DomT4: Let's make these `bin.mkpath` and `man1.mkpath`. Just a bit neater.

DomT4:
Ping.

xdbr:
Pong. Was sick, sorry for the delay. Will update according to your suggestions!

Library/Formula/mapext.rb
```diff
@@ -0,0 +1,18 @@
+class Mapext < Formula
+  desc "map files from one extension to another via a specified command"
+  homepage "http://www.acme.com/software/mapext/"
+  url "http://www.acme.com/software/mapext/mapext_10nov1995.tar.gz"
```
dunn: Does this mean there hasn't been a release in 20 years? We don't usually add abandoned software unless it's very widely used.

xdbr: Well, true. In the course of making my first formulas I thought why not. Feel free to close this PR, I'm not particularly keen on seeing it merged. Thought someone/the archive of formulas itself might benefit but plase just go ahead and close, fine with me!

dunn:
I guess we'll pass on this, then. But thank you for the submission! ❤️

PR #41552, opened 2015-07-09
Update vifm to v0.8

cfillion:
http://sourceforge.net/p/vifm/mailman/message/34282033/

### DIFF
--- a/Library/Formula/vifm.rb
+++ b/Library/Formula/vifm.rb
@@ -1,9 +1,9 @@
 class Vifm < Formula
   desc "Ncurses based file manager with vi like keybindings"
   homepage "http://vifm.info/"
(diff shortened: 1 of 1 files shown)


bfontaine:
The test is failing:

```
==> /usr/local/Cellar/vifm/0.8/bin/vifm -c :q
Failed to open terminal for output.
Error: vifm: failed
```

cfillion:
Fixed. I replaced the test because the original one (executing vifm's `:quit` command) cannot run without a proper terminal and faking it with `script` like in ranger's formula (#39661) doesn't suffice.

I couldn't find a better way to simulate a tty and its size in the test environment.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `vifm 0.8` (no added `v`).

PR #41553, opened 2015-07-09
(WIP) batch http->https fixes

bfontaine:
I checked all the stable URLs (using `brew fetch` and ensuring there were no errors/warnings), plus some of the homepages.

- [x] stable
- [x] devel
- [x] HEAD
- [x] Homepages

### DIFF
--- a/Library/Formula/liblqr.rb
+++ b/Library/Formula/liblqr.rb
@@ -2,8 +2,8 @@ require 'formula'
 
 class Liblqr < Formula
   desc "C/C++ seam carving library"
-  homepage 'http://liblqr.wikidot.com/'
-  url 'http://liblqr.wdfiles.com/local--files/en:download-page/liblqr-1-0.4.2.tar.bz2'
+  homepage 'https://liblqr.wikidot.com/'
+  url 'https://liblqr.wdfiles.com/local--files/en:download-page/liblqr-1-0.4.2.tar.bz2'
   version '0.4.2'
   sha1 '69639f7dc56a084f59a3198f3a8d72e4a73ff927'
 
--- a/Library/Formula/libvbucket.rb
+++ b/Library/Formula/libvbucket.rb
@@ -2,8 +2,8 @@ require 'formula'
 
 class Libvbucket < Formula
   desc "Utility library providing mapping to virtual buckets"
-  homepage 'http://couchbase.com/develop/c/current'
-  url 'http://packages.couchbase.com/clients/c/libvbucket-1.8.0.4.tar.gz'
+  homepage 'https://couchbase.com/develop/c/current'
+  url 'https://s3.amazonaws.com/packages.couchbase.com/clients/c/libvbucket-1.8.0.4.tar.gz'
   sha1 '4f24a85d251c0fca69e7705681a2170dd794492a'
 
   bottle do
--- a/Library/Formula/a2ps.rb
+++ b/Library/Formula/a2ps.rb
@@ -1,6 +1,6 @@
 class A2ps < Formula
   desc "Any-to-PostScript filter"
-  homepage "http://www.gnu.org/software/a2ps/"
+  homepage "https://www.gnu.org/software/a2ps/"
   url "http://ftpmirror.gnu.org/a2ps/a2ps-4.14.tar.gz"
   mirror "http://ftp.gnu.org/gnu/a2ps/a2ps-4.14.tar.gz"
   sha1 "365abbbe4b7128bf70dad16d06e23c5701874852"
--- a/Library/Formula/agda.rb
+++ b/Library/Formula/agda.rb
@@ -5,7 +5,7 @@ class Agda < Formula
 
   desc "Dependently typed functional programming language"
   homepage "http://wiki.portal.chalmers.se/agda/"
-  url "http://hackage.haskell.org/package/Agda-2.4.2.2/Agda-2.4.2.2.tar.gz"
+  url "https://hackage.haskell.org/package/Agda-2.4.2.2/Agda-2.4.2.2.tar.gz"
   sha1 "fbdf7df3d5a036e683210ac7ccf4f8ec0c9fea05"
   revision 1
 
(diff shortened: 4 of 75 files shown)


bfontaine:
I killed the build.

Library/Formula/pcre2.rb
```diff
@@ -1,6 +1,6 @@
 class Pcre2 < Formula
   desc "Perl compatible regular expressions library with a new API"
-  homepage "http://www.pcre.org/"
+  homepage "https://pcre.org/"
```
vszakats: The `www.`-less URL is fine, but sadly the https:// authentication fails with a domain mismatch.

Library/Formula/zeromq.rb
```diff
@@ -1,6 +1,6 @@
 class Zeromq < Formula
   desc "High-performance, asynchronous messaging library"
-  homepage "http://www.zeromq.org/"
+  homepage "https://www.zeromq.org/"
```
vszakats: Same here, it's a domain mismatch with the certificate.

bfontaine:
Thanks @vszakats, I removed these (and killed the build).

Library/Formula/ansible.rb
```diff
@@ -1,7 +1,7 @@
 class Ansible < Formula
   desc "Automate deployment, configuration, and upgrading"
-  homepage "http://www.ansible.com/home"
-  url "http://releases.ansible.com/ansible/ansible-1.9.2.tar.gz"
+  homepage "https://www.ansible.com/home"
```
vszakats: Noticed another CA mismatch in `homepage` above.

Library/Formula/flashrom.rb
```diff
@@ -1,6 +1,6 @@
 class Flashrom < Formula
   desc "Identify, read, write, verify, and erase flash chips"
-  homepage "http://flashrom.org/"
+  homepage "https://flashrom.org/"
```
vszakats: This one uses a CACert certificate and for another domain.

Library/Formula/gearman.rb
```diff
@@ -1,6 +1,6 @@
 class Gearman < Formula
   desc "Application framework to farm out work to other machines or processes"
-  homepage "http://gearman.org/"
+  homepage "https://gearman.org/"
```
vszakats: Couldn't connect to this one via `https://` just now.

Library/Formula/liblqr.rb
```diff
@@ -3,7 +3,7 @@
 class Liblqr < Formula
   desc "C/C++ seam carving library"
   homepage 'http://liblqr.wikidot.com/'
```
vszakats: This one can be updated to `https://`, which will redirect to plaintext (for non-logged-in users I guess).

Library/Formula/libvbucket.rb
```diff
@@ -3,7 +3,7 @@
 class Libvbucket < Formula
   desc "Utility library providing mapping to virtual buckets"
   homepage 'http://couchbase.com/develop/c/current'
```
vszakats: This one can be updated to `https://`.

Library/Formula/little-cms2.rb
```diff
@@ -1,6 +1,6 @@
 class LittleCms2 < Formula
   desc "Color management engine supporting ICC profiles"
-  homepage "http://www.littlecms.com/"
+  homepage "https://www.littlecms.com/"
```
vszakats: There is a certificate domain mismatch here.

Library/Formula/openconnect.rb
```diff
@@ -1,6 +1,6 @@
 class Openconnect < Formula
   desc "Open client for Cisco AnyConnect VPN"
-  homepage "http://www.infradead.org/openconnect.html"
+  homepage "https://www.infradead.org/openconnect.html"
```
vszakats: Domain mismatch and CACert certificate here.

Library/Formula/parrot.rb
```diff
@@ -2,7 +2,7 @@
 
 class Parrot < Formula
   desc "Open source virtual machine (for Perl6, et al.)"
-  homepage "http://www.parrot.org/"
+  homepage "https://www.parrot.org/"
```
vszakats: Its wildcard certificate expired in 2011.

Library/Formula/u-boot-tools.rb
```diff
@@ -1,6 +1,6 @@
 class UBootTools < Formula
   desc "Universal boot loader"
-  homepage "http://www.denx.de/wiki/U-Boot/"
+  homepage "https://www.denx.de/wiki/U-Boot/"
```
vszakats: This one uses a self-signed certificate.

bfontaine:
@vszakats Thank you for all your comments regarding the mismatched certificates, in fact I checked on my side with `curl` and when I pushed I saw you commented on all the formulae I just removed :wink: 

This PR is a first step, I’m not trying to fix *all* formulae (that can be) with it. I wrote a Ruby script to parse [HTTPS-Everywhere’s rules](https://github.com/EFForg/https-everywhere/tree/master/src/chrome/content/rules) and applied it on Homebrew but it doesn’t support all kind of rules for now, and I got a lot of false positives as you saw.

Library/Formula/libvbucket.rb
vszakats: It is mixed-content, but these are usually accepted. (should be fixed upstream eventually)

vszakats:
Using HTTPS Everywhere rules is great, thumbs up! I now understand why these were included. I was about to recommend it a while ago, only their rules seemed complex to parse.

bfontaine:
Yeah I remember someone mentionned these in a PR, maybe that was you. The rules didn’t seem hard to parse so I tried, but in practice there are a lot of weird cases and it didn’t get as smooth as I was expected.

You can find [the parser here](https://github.com/bfontaine/tohttps.rb/blob/brew/tohttps.rb) and [the script here](https://github.com/bfontaine/tohttps.rb/blob/brew/brew-fix-http.rb) (warning: these are quick & dirty).

vszakats:
Ah, that's great, nice job.

bfontaine:
Any other comment before I merge this?

mikemcquaid:
Do it :+1:

PR #41556, opened 2015-07-09
Formatting corrections to Homebrew's core

woodruffw:
Hello!

These commits correct some minor formatting issues in Homebrew's core, namely missing spacing between methods and missing parentheses in method declarations with arguments.

Each directory's formatting corrections are isolated in its own commit.

Other potential issues I didn't correct, but can if wanted:
- [ ] Remaining heredocs that haven't been converted to use `EOS.undent`
- [ ] Excessively brief method arguments (`f`, `fn`, `val`, etc)
- [ ] Indentation corrections and preference of `""` over `''` (or vice versa)
- [ ] Replacement of Perl-style special variables with Ruby-style ones

### DIFF
--- a/Library/Homebrew/blacklist.rb
+++ b/Library/Homebrew/blacklist.rb
@@ -1,4 +1,4 @@
-def blacklisted? name
+def blacklisted?(name)
   case name.downcase
   when 'screen', /^rubygems?$/ then <<-EOS.undent
     Apple distributes #{name} with OS X, you can find it in /usr/bin.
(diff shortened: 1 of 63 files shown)


bfontaine:
Unfortunately we don’t merge style-only pull-requests.

mikemcquaid:
Yup, sorry!

woodruffw:
I was originally going to include these in #41021, but was advised to keep that limited to the bottling installation system only.
Should I drop this entirely or try to include it in some other fashion? I think it would be a big benefit in terms of conformity/readability of the codebase.

mikemcquaid:
I don't think the conformity is important necessarily. The big downside is that you'll now show up for the `git blame` for these lines.

woodruffw:
Alright. I'll drop this.

mikemcquaid:
Thanks :heart:

PR #41568, opened 2015-07-10
bottle: improve relocation 

### LINKED ISSUE #41545: Bottle with HOMEBREW_PREFIX not /usr/local incorrectly flags bottle as not relocatable
When `HOMEBREW_PREFIX` is not `/usr/local`, finding the prefix in `source.path` of `INSTALL_RECEIPT.json` incorrectly flags the bottle as not being relocatable.

I've only tested this on Linuxbrew. I'm assuming that this issue also applies to Mac Homebrew.

```
$ brew bottle --rb -v mpfr
==> Determining mpfr bottle revision...
==> Bottling mpfr-3.1.3.x86_64_linux.bottle.tar.gz...
tar czf /home/linuxbrew/bottles/mpfr-3.1.3.x86_64_linux.bottle.tar.gz mpfr/3.1.3
Warning: String '/home/linuxbrew/.linuxbrew' still exists in these files:
/home/linuxbrew/.linuxbrew/Cellar/mpfr/3.1.3/INSTALL_RECEIPT.json
 --> match '{"used_options":[],"unused_options":["--32-bit"],"built_as_bottle":true,"poured_from_bottle":false,"time":1436465451,"HEAD":null,"stdlib":null,"compiler":"gcc-5","source":{"path":"/home/linuxbrew/.linuxbrew/Library/Formula/mpfr.rb","tap":"Homebrew/homebrew"}}' at offset 0x0

./mpfr-3.1.3.x86_64_linux.bottle.tar.gz
  bottle do
    prefix "/home/linuxbrew/.linuxbrew"
    cellar "/home/linuxbrew/.linuxbrew/Cellar"
    sha256 "34949fd7e27acbd51b648b28f2a12112aa7b5cf2af1c31322e049e3351c563ad" => :x86_64_linux
  end
```

xu-cheng:
Fixes #41545.

### DIFF
--- a/Library/Homebrew/cmd/bottle.rb
+++ b/Library/Homebrew/cmd/bottle.rb
@@ -52,6 +52,9 @@ module Homebrew
     result = false
 
     keg.each_unique_file_matching(string) do |file|
+      # skip document file.
+      next if Metafiles::EXTENSIONS.include? file.extname
+
       # Check dynamic library linkage. Importantly, do not run otool on static
       # libraries, which will falsely report "linkage" to themselves.
       if file.mach_o_executable? or file.dylib? or file.mach_o_bundle?
--- a/Library/Homebrew/keg_fix_install_names.rb
+++ b/Library/Homebrew/keg_fix_install_names.rb
@@ -39,6 +39,7 @@ class Keg
     end
 
     files = pkgconfig_files | libtool_files | script_files | plist_files
+    files << tab_file if tab_file.file?
 
     files.group_by { |f| f.stat.ino }.each_value do |first, *rest|
       s = first.open("rb", &:read)
@@ -206,5 +207,9 @@ class Keg
       plist_files << pn
     end
     plist_files
+  end
+
+  def tab_file
+    join(Tab::FILENAME)
   end
 end


mikemcquaid:
This isn't actually the fix we want; this file should be trivially relocatable and I'm confused why it isn't being.

mikemcquaid:
Ah, we're only looking for script files (https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/keg_fix_install_names.rb#L163-L173)

xu-cheng:
I am confusing about the code of bottle relocatable checking as well. It seems to me that we did't check whether a unexpected string is relocatable at all during `brew bottle`. `Keg#script_files` is only used in installing the bottle rather than creating the bottle.

mikemcquaid:
I think ideally we should just rewrite all text files. Just need to work out a way to detect them. Hmm.

mikemcquaid:
@xu-cheng It's relocated before we try to create it: https://github.com/xu-cheng/homebrew/blob/keg-bottle/Library/Homebrew/cmd/bottle.rb#L152-L156

mikemcquaid:
Could maybe just use `file` and see if the output includes `text`

xu-cheng:
I think we should just skip relocation for all [Metafiles](https://github.com/xu-cheng/homebrew/blob/keg-bottle/Library/Homebrew/metafiles.rb) including `INSTALL_RECEIPT.json`. For example, we don't need to and shouldn't replace the text inside document. Right? (Maybe html should be an exception?)

mikemcquaid:
If we just add exceptions we'll need to keep adding them, though. Better to have a generalised solution, I think.

xu-cheng:
Actually I mean we should just skip document files. They are unrelated to relocation and we shouldn't alter them.

mikemcquaid:
We can skip `.md` files but I think the `.json` file should be part of a more generic fix.

xu-cheng:
Updated.

* skip document for relocatable check
* relocate INSTALL_RECEIPT.json file

I don't think I can come up some more generic way to detect what text file should be relocated. Not before I see some of this kind.

I suggest we add `-v` to `brew bottle` for test-bot. So we can be aware what files stop bottles becoming `cellar :any`.

mikemcquaid:
Ok, seems fine :+1: 

> I suggest we add -v to brew bottle for test-bot. So we can be aware what files stop bottles becoming cellar :any.

Good idea.

I'll try and work on the more general case.

xu-cheng:
So good to merge this PR?

mikemcquaid:
Yup.

xu-cheng:
:ship: ed.

PR #41570, opened 2015-07-10
Explicitly pass `--ff` and `--no-rebase` to `git pull` on `brew update`

jawshooah:
This overrides any global or per-repository `pull.rebase` and `pull.ff` config settings the user may have set.

### DIFF
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -196,6 +196,7 @@ class Updater
     safe_system "git", "config", "core.autocrlf", "false"
 
     args = ["pull"]
+    args << "--ff"
     args << ((ARGV.include? "--rebase") ? "--rebase" : "--no-rebase")
     args += quiet
     args << "origin"
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -196,7 +196,7 @@ class Updater
     safe_system "git", "config", "core.autocrlf", "false"
 
     args = ["pull"]
-    args << "--rebase" if ARGV.include? "--rebase"
+    args << ((ARGV.include? "--rebase") ? "--rebase" : "--no-rebase")
     args += quiet
     args << "origin"
     # the refspec ensures that 'origin/master' gets updated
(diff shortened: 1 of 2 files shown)


mikemcquaid:
Why, out of interest?

jawshooah:
The `--ff` is to ensure that we fast-forward to the latest upstream changes if possible. The user may have set `pull.ff` to `false`, which tells Git to create a merge commit even if a fast-forward is possible.

The `--no-rebase` is for speed (rebase is *much* slower than a fast-forward merge, especially if you have a lot of taps to update) and to ensure that any unpublished commits are not silently rebased onto upstream.

mikemcquaid:
Will wait to see what other maintainers think but I don't see a need for this personally unless it's causing us issues. If people have made these settings globally they probably would rather e.g. rebasing happen (I know I do/did).

jawshooah:
I prefer rebasing as well for repositories I hack on (I have `pull.rebase` set to `preserve` globally), but I imagine that the vast majority of users just `brew tap` a repo and forget about it.

If you're hacking on a tap, you'll probably be getting upstream updates via `git pull` rather than `brew update` anyway.

jawshooah:
Regarding `--ff`, in what scenario would you want to create an extra merge commit for *every update* on *every tap*? Similar to my argument above, most users probably are not considering taps when setting `pull.ff` to `false` globally.

mikemcquaid:
I'm not disputing this is theoretically suboptimal but whether it's actually causing any issues for any of our users in practice.

jawshooah:
What's that old saying about an ounce of prevention and a pound of cure? :wink:

mikemcquaid:
Gonna pass on this for now but will revisit if we have users reporting related issues this would have fixed.

jawshooah:
Isn't the idea of providing the CLI flag that it changes default behavior? Should a user not expect that, when running `brew update` without `--rebase`, no rebase will be performed?

glenjamin:
This issue affects me - i have `merge.ff=false` set in my global git config - as I prefer this default when using `git merge`.

Homebrew using git internally for update means that when I run `brew update` I have to accept a merge comment via `vim` for every repo, which is annoying as it means I have to watch the command and perform a manual action for every tap instead of just running it and checking back later.


Rather than forcing the particular merge approach, how about just disabling the editor prompt? http://stackoverflow.com/a/12752379/173062

mikemcquaid:
Changed my mind here as @glenjamin has chimed in. If the `brew tests` failures can be fixed up I'm happy to merge this. Sorry for the back and forth @jawshooah!

jawshooah:
@mikemcquaid I believe I've fixed the broken tests. We'll see what @BrewTestBot has to say.

mikemcquaid:
Thanks @jawshooah, really appreciate you coming back and handling this :heart:!

jawshooah:
No problem! :beers:

glenjamin:
Has this regressed recently? I'm seeing editor prompts for taps when doing `brew update` - but I'm fairly sure they had stopped appearing for a while.

PR #41579, opened 2015-07-10
umobj 1.2.0 (new formula)

zbeekman:
Command line tools for interacting with Amazon S3 type ([ceph](https://github.com/ceph/ceph)) object stores developed at the University of Maryland. In my tests for my specific application of tools, they outperformed [s3cmd](https://github.com/s3tools/s3cmd) by a factor of about 2X.

### DIFF
--- a/Library/Formula/umobj.rb
+++ b/Library/Formula/umobj.rb
@@ -1,0 +1,60 @@
+class Umobj < Formula
+  desc "Command-line utilties for S3-compatible Object Storage"
+  homepage "https://github.com/UMIACS/umobj"
+  url "https://pypi.python.org/packages/source/u/umobj/umobj-1.2.0.tar.gz"
+  sha256 "cacc2efe07b069660df595b409afc16639d96856ec9954c105b95358553d3add"
+
+  head "https://github.com/UMIACS/umobj.git"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  resource "netaddr" do
+    url "https://pypi.python.org/packages/source/n/netaddr/netaddr-0.7.15.tar.gz"
+    sha256 "d5b5bb3f4e9a94f93d232bb876e567517d8fb59e5bebb5339483b44df5529d11"
+  end
+
+  resource "qav" do
+    url "https://pypi.python.org/packages/source/q/qav/qav-0.3.2.tar.gz"
+    sha256 "4e16d4544451f200689a2ae6908b14669f9f2a4a8881368d78e609196fbdee84"
+  end
+
+  resource "boto" do
+    url "https://pypi.python.org/packages/source/b/boto/boto-2.38.0.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/umobj.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir[libexec/"bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    system "#{bin}/lsobj", "--help"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikemcquaid: e.g. list a public S3 bucket

zbeekman:
The problem right now is that it only works when given full credentials. So someone's secret key would have to be exposed in the formula. I've been in contact with h th package authors to see if that can be changed or if they know of any better tests. 

CC @liammonahan

mikemcquaid:
Even for a publicly readable bucket?

zbeekman:
right now, yes. Complains if you don’t set a `OBJ_SECRET_KEY_ID` and an `OBJ_KEY_ID`. If you set one of them to garbage it doesn’t work. I think this might be due to the fact that [umobj](https://github.com/UMIACS/umobj) uses [boto](https://github.com/boto/boto) to communicate with AWS…

zbeekman:
Ah yes the good old not notable enough problem… That’s strange, when I ran `brew audit --strict` on my local machine I didn’t get that warning…

Well if you, @mikemcquaid, are against merging this before the 30 day period is up (the project is actually older and maintained originally in a [gitlab repo](https://gitlab.umiacs.umd.edu/staff/umobj)) and before it has more followers, forks etc. then perhaps a better means of testing without needing credentials will become available in the interim.

mikemcquaid:
@zbeekman You need `--strict --online` unfortunately. We can leave this open for a bit but it's just an indication that the project may not have widespread enough popularity and you might be better maintaining it in a tap yourself.

zbeekman:
@mikemcquaid its your call. My vote is either close the PR without merging and I’ll wait the 30 days and see if the github presence has garnered enough following, or merge it now. IMO there’s no point leaving this PR open just to wait and see, since it distracts homebrew maintainers from PRs and issues that need more immediate attention.

liammonahan:
Hi guys,

Sorry I'm late to the party. Let me take a look tomorrow about getting a meaningful test in place.  

Thanks,
Liam

mikemcquaid:
Will pass on this for now, thanks for understanding @zbeekman.

zbeekman:
Sounds good. I’ll resubmit once it passes `brew audit --strict --online`

PR #41580, opened 2015-07-10
Add dependency on gnutls to enable newest TLS protocols

sdebnath:
As more and more email providers are disabling SSLv3 and TLS 1.0, the need
the need for the latest and greatest TLS vesrion support is paramount.
This change adds gnutls as dependency which enables default TLS mutt
options like ssl_use_tlsv1_2 to work (see
  http://www.mutt.org/doc/devel/manual.html#ssl-use-sslv3).

Some providers that have already cut users off for using SSLv3 or TLS 1.0
are Office365 and ZohoMail (maybe others). Others are likely to enforce
this option soon as well given the Poodle vulnerability:

  https://access.redhat.com/articles/1232123

### DIFF
--- a/Library/Formula/mutt.rb
+++ b/Library/Formula/mutt.rb
@@ -46,6 +46,7 @@ class Mutt < Formula
   depends_on "automake" => :build
 
   depends_on "openssl"
+  depends_on "gnutls"
   depends_on "tokyo-cabinet"
   depends_on "s-lang" => :optional
   depends_on "gpgme" => :optional


DomT4:
OpenSSL supports TLS 1.2 and below as well. What stops Mutt using that for the protocols?

sdebnath:
You know what, now I can't repro the issue. Discussed this on machomebrew channel on irc, and it started working after installing gnutls. Removed and retried and it still works. Retracting the PR for now till I know whats going on.

DomT4:
Sure. Feel free to shout if you hit any issues. @chdiza is a heavy user of Mutt I believe.

chdiza:
> This change adds gnutls as dependency which enables default TLS mutt
options like ssl_use_tlsv1_2 to work

Those already work.  There is no need for gnutls.

See https://github.com/Homebrew/homebrew/pull/33546

chdiza:
For those not inclined to read through that whole thread: `mutt` and `openssl` each have misleading and sloppy protocol reporting which makes it *appear* that `ssl_use_tlsv1_2` and friends are not working.  But they are working.  As it happens, `gnutls` is less sloppy about the report it gives `mutt`, and thus when the user brews against `gnutls` this misleading appearance doesn't surface.

PR #41584, opened 2015-07-10
ice: multiple formula fixes

externl:
- Add Java 8 support
- Always build slice2java and slice2freezej compilers
- Add test

### DIFF
--- a/Library/Formula/ice.rb
+++ b/Library/Formula/ice.rb
@@ -10,24 +10,19 @@ class Ice < Formula
     sha256 "87a5dd2311296bea03889062d4f20ad76a7de31382a42409bfb933d8e66052eb" => :mavericks
   end
 
+  option "with-java", "Build Ice for Java and the IceGrid Admin app"
+
   depends_on "mcpp"
-  depends_on :java  => ["1.7", :optional]
+  depends_on :java => :optional
   depends_on :macos => :mavericks
 
   resource "berkeley-db" do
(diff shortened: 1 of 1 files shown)


Library/Formula/ice.rb
```diff
@@ -89,4 +81,39 @@ def install
...
+        adapter->add(new HelloI, communicator->stringToIdentity("hello"));
+        adapter->activate();
+        communicator->destroy();
+        return 0;
+      }
+    EOS
+    system "#{bin}/slice2cpp", "Hello.ice"
+    system "xcrun", "clang++", "-c", "-I#{include}", "-I.", "Hello.cpp"
+    system "xcrun", "clang++", "-c", "-I#{include}", "-I.", "Test.cpp"
+    system "xcrun", "clang++", "-L#{lib}", "-o", "test", "Test.o", "Hello.o", "-lIce", "-lIceUtil"
```
mikemcquaid: Does it fail with the `xcrun`?

externl: xcrun worked fine for me locally.

mikemcquaid: Sorry, I meant without the `xcrun`; don't think it should be necessary.

externl: Fails to link. `xcrun` adds some necessary Frameworks automatically.

mikemcquaid: Ok, thanks.

mikemcquaid:

PR #41592, opened 2015-07-11
add space around URL of pull request in output

polamjag:
With default setting of iTerm.app, selecting URL of pull request with right-click results in like this:

scr 2015-07-11 at 10 49 04
scr 2015-07-11 at 10 49 20

This pr fixes that by appending space around the URL.

### DIFF
--- a/Library/Homebrew/exceptions.rb
+++ b/Library/Homebrew/exceptions.rb
@@ -214,7 +214,7 @@ class BuildError < RuntimeError
     puts
     unless RUBY_VERSION < "1.8.7" || issues.empty?
       puts "These open issues may also help:"
-      puts issues.map{ |i| "#{i['title']}: #{i['html_url']}" }.join("\n")
+      puts issues.map{ |i| "#{i['title']} #{i['html_url']}" }.join("\n")
     end
 
     if MacOS.version >= "10.11"
--- a/Library/Homebrew/exceptions.rb
+++ b/Library/Homebrew/exceptions.rb
@@ -214,7 +214,7 @@ class BuildError < RuntimeError
     puts
     unless RUBY_VERSION < "1.8.7" || issues.empty?
       puts "These open issues may also help:"
-      puts issues.map{ |i| "#{i['title']} ( #{i['html_url']} )" }.join("\n")
+      puts issues.map{ |i| "#{i['title']}: #{i['html_url']}" }.join("\n")
     end
 
     if MacOS.version >= "10.11"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Not sure what I think about changing this just for iTerm but will let other @Homebrew/owners chime in.

Library/Homebrew/exceptions.rb
```diff
@@ -214,7 +214,7 @@ def dump
     puts
     unless RUBY_VERSION < "1.8.7" || issues.empty?
       puts "These open issues may also help:"
-      puts issues.map{ |i| "#{i['title']} (#{i['html_url']})" }.join("\n")
+      puts issues.map{ |i| "#{i['title']} ( #{i['html_url']} )" }.join("\n")
```
xu-cheng: The space is weird. I would suggest to drop parentheses. i.e. `title url`.

polamjag:
Thanks for review! How about this one: `title: url` ? (7312686e239768f5d6b24e79f00f0eb13cb392cf)

DomT4:
You could always hit the `cmd` key and then click the link in iTerm2. That should take you directly to it in your default browser without having to copy the link?

polamjag:
@DomT4 Thank you for your advice, but I always use terminal in tmux with mouse enabled, so `cmd` + click is handled by tmux :cold_sweat:

DomT4:
Hmm. Yes, I see the issue is easy to reproduce.

I'm not sure there's a particularly easy way to fix this; each terminal handles the situation slightly differently. iTerm includes the end bracket, the standard Terminal seems to copy everything including the description of the PR. It could be tricky trying to work something that fits all here.

Library/Homebrew/exceptions.rb
```diff
@@ -214,7 +214,7 @@ def dump
     puts
     unless RUBY_VERSION < "1.8.7" || issues.empty?
       puts "These open issues may also help:"
-      puts issues.map{ |i| "#{i['title']} (#{i['html_url']})" }.join("\n")
+      puts issues.map{ |i| "#{i['title']}: #{i['html_url']}" }.join("\n")
```
mikemcquaid: Final thing: let's drop the `:` and we'll just merge without the brackets, thanks.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/343d2497a1ed03dbd9d8456af2d83c35488da20b. Thank you for your contributions to Homebrew, we appreciate it! :smiley_cat:

polamjag:
Thank you too! :smile:

PR #41593, opened 2015-07-11
Merge test gitignore into root version.

mikemcquaid:
libgit2 clients struggle at parsing these otherwise because of how we set up our ignore rules.

And remove now unneeded formula ignore rule.

CC @Homebrew/owners any objections? Would make my life a bit easier (and anyone else using the GitHub GUI).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,9 @@
 !/share/man/man1/brew.1
 .DS_Store
 /Library/Homebrew/doc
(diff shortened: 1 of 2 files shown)


bfontaine:
Never had any issue with this, probably because I only use `git` on the command-line. :+1:

DomT4:
No objection here.

xu-cheng:
:+1:

PR #41595, opened 2015-07-11
percona-toolkit does not depend on MySQL

adamchainz:
There are three MySQL forks on homebrew now: `mysql`, `mariadb`, and `percona-server`. One could want to use `percona-toolkit` with any of them, or even a remote database server, so you shouldn't need to install `mysql` (which conflicts with `mariadb` and `percona-server`) for it to work.

It turns out that installing the `DBD::mysql` dependency does not require `libmysql` to be installed, so it seems fine to just remove it as a dependency. To test this I removed this line locally with `brew edit percona-toolkit`, uninstalled `mariadb`, and did `brew install --fresh percona-toolkit`. It worked, and the `pt-summary` command (which doesn't connect to MySQL at all) worked fine. Then I installed `mariadb` again, and ran `pt-mysql-summary` which also worked fine.

### DIFF
--- a/Library/Formula/percona-toolkit.rb
+++ b/Library/Formula/percona-toolkit.rb
@@ -13,7 +13,6 @@ class PerconaToolkit < Formula
     sha256 "95fec4e6b45806a11c940c48f6e811101e7bf711aba758783762b2133229b08f" => :mountain_lion
   end
 
-  depends_on :mysql
   depends_on "openssl"
 
   resource "DBD::mysql" do


mikemcquaid:
This is actually an issue not with the dependency (which does what you expect it to) but the bottling code which means that if a bottle has been built against MySQL then it'll be installed against that too. As you can see from the test it doesn't work to just remove it as you have.

mikemcquaid:
@Homebrew/owners I'm wondering if we should make a new tag e.g. `:default` for requirement dependencies which triggers the bottle behaviour as-is (e.g. it'll always need `default_formula` installed). It feels like the majority of the time this is the wrong thing to do.

adamchainz:
It seems like `mysql_config` is required to run the perl `Makefile.pl`'s tests. `mysql_config` is ofc provided by `mysql`, `mariadb`, and `percona-server` - is there not a way to have brew require the abstract idea of `mysql_config` and then that be concretely supplied by one of the three?

dunn:
If `Makefile.pl` lets you pass the correct mysql_config as an argument, you could try `which "mysql_config"`.

tdsmith:
Yes, there is a `:mysql` requirement that does what you want, but when we build a bottle (a binary distribution) of a package that depends on a requirement like :mysql, we resolve the requirement to the package which provides it by default. We identically resolve the requirement to the default package when we install a bottle on the theory that isn't in general safe to assume that a binary built against one package that satisfies a requirement may then be run successfully against a different package that also satisfies the requirement (i.e. that the specific package against which it is built may matter). Requirements can be resolved from the environment at install time by building from source (`brew install --build-from-source`).

I think that having a way to say "it is safe to attempt to resolve this requirement from the environment at bottle install time" would be useful. I think the default is safer and generally correct and should not change. I'm not sure which is appropriate in this case.

xu-cheng:
>  I'm wondering if we should make a new tag e.g. :default for requirement dependencies which triggers the bottle behaviour as-is

We already have one i.e. `:run`.

mikemcquaid:
> We already have one i.e. :run.
>  Hide all checks

:+1:

tdsmith:
Should this read `depends_on :mysql => :run` then?

mikemcquaid:
@tdsmith if it works but it sounds like something is needed any compile-time.

tdsmith:
Taking a closer look, we disabled the percona-toolkit bottle and it's a short build so I think this behavior no longer manifests. Thanks for reporting!

PR #41597, opened 2015-07-11
doctor: don't point to App Store for Xcode on 10.11.

### LINKED ISSUE #41589: Should direct 10.11 public beta users to Xcode website
It looks like 10.11 public beta users don't get Xcode 7 and it doesn't show up in the App Store. However it is available at this url. Might be good to have in the brew doctor warnings. 

https://developer.apple.com/xcode/downloads/

mikemcquaid:
@Homebrew/owners thoughts?

Closes #41589.

### DIFF
--- a/Library/Homebrew/cmd/doctor.rb
+++ b/Library/Homebrew/cmd/doctor.rb
@@ -269,13 +269,28 @@ if MacOS.version >= "10.9"
     end
   end
 
-  def check_xcode_up_to_date
-    if MacOS::Xcode.installed? && MacOS::Xcode.outdated?
-      <<-EOS.undent
-      Your Xcode (#{MacOS::Xcode.version}) is outdated
-      Please update to Xcode #{MacOS::Xcode.latest_version}.
-      Xcode can be updated from the App Store.
-      EOS
+  # TODO: remove when 10.11 is released
+  if MacOS.version >= "10.11"
+    def check_xcode_up_to_date
+      if MacOS::Xcode.installed? && MacOS::Xcode.outdated?
+        <<-EOS.undent
+        Your Xcode (#{MacOS::Xcode.version}) is outdated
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/doctor.rb
```diff
@@ -269,13 +269,28 @@ def check_for_installed_developer_tools
...
+  if MacOS.version >= "10.11"
+    def check_xcode_up_to_date
+      if MacOS::Xcode.installed? && MacOS::Xcode.outdated?
+        <<-EOS.undent
+        Your Xcode (#{MacOS::Xcode.version}) is outdated
+        Please update to Xcode #{MacOS::Xcode.latest_version}.
+        Xcode can be updated from
+          https://developer.apple.com/downloads
+        EOS
+        EOS
```
DomT4: Redundant EOS here?

DomT4:
Can we append the correct argument rather than recreating the whole block? Something like:

```ruby
def check_xcode_up_to_date
  if MacOS.version >= "10.11"
    correct_link = "Available from https://developer.apple.com/downloads"
  else
    correct_link = "Xcode can be updated from the App Store."
  end

  if MacOS::Xcode.installed? && MacOS::Xcode.outdated?
  <<-EOS.undent
    Your Xcode (#{MacOS::Xcode.version}) is outdated
    Please update to Xcode #{MacOS::Xcode.latest_version}.
    #{correct_link}
  EOS
  end
end
```

Would that work? Haven't tested it locally, so feel free to laugh if it doesn't :wink:

mikemcquaid:
Was trying to be vaguely consistent with the approach below. I'm happy to change it around though.

DomT4:
If you're happy with it as is, I'm not super-opposed to the current style, just wondered if the above was neater.

PR #29742, opened 2014-06-01
Gnome themes

rbalint:
Standard GNOME theme for GTK+2 and GTK+3 (with and without Quartz) applications

PR #41601, opened 2015-07-11
gnome-themes-standard: new formula

tschoonj:
see also #29742
This package is quite useful for Gtk+ but almost pointless for Gtk+3.
For Gtk+ it offers two themes: Adwaita and HighContrast, while these are
both included in Gtk+3 since 3.14.0.
It also offers the HighContrast icon theme, which could be used for both
Gtk+ and Gtk+3 if the users explicitly choose to do so in their settings.
Unless switched off with `--disable-gtk3-engine`, there is a hard
dependency on Gtk+3 in `configure.ac`, which appears to be completely
useless as it is nowhere used in any of the `Makefile.am`'s in a
meaningful way. I will file a bug over this.


mikemcquaid:

For future reference the preferred commit message format new formulae `gnome-themes-standard (new formula)`.

PR #41606, opened 2015-07-11
gnome-icon-theme: librsvg dependency should be build only

tschoonj:
This formula has only `:build` dependencies


mikemcquaid:
Thanks @tschoonj

tschoonj:
Thanks, saw this when I was working on the `gnome-themes-standard` formula...

PR #41607, opened 2015-07-11
audit --online: check homepage reachability

bfontaine:
Example output:

```
brew audit --online ddar gd freealut
ddar:
 * The homepage is not reachable (curl exit code 28)
 * Dependency protobuf does not define option "with-python"
 * Dependency protobuf does not define option "with-python"

gd:
 * The homepage is not reachable (curl exit code 60)

freealut:
 * The homepage is not reachable (curl exit code 6)

Error: 7 problems in 3 formulae
```

(note that we should fix those)

cc @Homebrew/owners

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -428,6 +428,13 @@ class FormulaAuditor
          %r[^http://code\.google\.com/]
       problem "Please use https:// for #{homepage}"
     end
+
+    return unless @online
+    begin
+      nostdout { curl "--connect-timeout", "15", "-IL", "-o", "/dev/null", homepage }
+    rescue ErrorDuringExecution
+      problem "The homepage is not reachable (curl exit code #{$?.exitstatus})"
+    end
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/audit.rb
```diff
@@ -428,6 +428,13 @@ def audit_homepage
          %r[^http://code\.google\.com/]
       problem "Please use https:// for #{homepage}"
     end
+
+    return unless @online
+    begin
+      nostdout { curl "--connect-timeout", "4", "-I", "-o", "/dev/null", homepage }
```
mikemcquaid: Maybe bump this to 10-15s? Also, does this need a `-L`?

mikemcquaid:
Nice work :+1:

Library/Homebrew/cmd/audit.rb
bfontaine: I just added a `-L` as well as a 15-seconds timeout limit.

PR #41614, opened 2015-07-12
audit: automatically detect secure urls

DomT4:
Took an old implementation I've had sitting around for a few weeks and polished it up to something reasonably acceptable. Should be fairly obvious what it's doing here.

Complements #41607.

CC @Homebrew/owners

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -479,6 +479,19 @@ class FormulaAuditor
     end
   end
 
+  def audit_tls_potential
+    # This only handles stable URLs at the moment. I'd like to give it some field
+    # testing on a relatively simple check before we expand it further.
+    return unless @online
+
+    url = formula.stable.url
+    unless url =~ %r[^https://]
+      if system "curl", "--connect-timeout", "3", "--silent", "-I", "https://#{url.split("http://")[1]}", "-o", "/dev/null"
+        problem "Please change #{url} to use https://"
+      end
+    end
+  end
+
   def audit_specs
     if head_only?(formula) && formula.tap.to_s.downcase !~ /-head-only$/
       problem "Head-only (no stable download)"
@@ -890,6 +903,7 @@ class FormulaAuditor
     audit_desc
     audit_homepage
     audit_github_repository
+    audit_tls_potential
     audit_deps
     audit_conflicts
     audit_options


bfontaine:
One thing we discussed on the chat would be to verify that the https URL returns the same content as the http one using the `Etag` header.

mikemcquaid:
The fallback is doing a e.g. MD5/SHA or the actual URL contents.

mikemcquaid:
You can get the `Etag` with `curl -I`

xu-cheng:
Some comments:
* The code should be moved to `ResourceAuditor`, which is responsible to all resources. 
* You should audit URL only if the download strategy is curl and URL starts with `http://`. 
* I would suggest to use `str#start_with?` instead of regex matching. I found regex matching has poor performance.

DomT4:
> The code should be moved to ResourceAuditor, which is responsible to all resources.

I was deliberately trying to keep it limited in scope at the moment, the inclusion of it outside the resource auditor was attempting to be indicative of that. I guess it can move and still be limited for a while though.

> You should audit URL only if the download strategy is curl and URL starts with http://.

The latter is already happening here, but I agree with ensuring the former.

> I would suggest to use str#start_with? instead of regex matching.

For the `unless url =~ %r[^https://]` bit? Sure, I'll look into it.

DomT4:
> One thing we discussed on the chat would be to verify that the https URL returns the same content as the http one using the Etag header, and fallback on something else if the header isn’t supported.

I was hoping to avoid sending two requests, but I guess with such a short timeout it shouldn't slow the audit too hugely. `Etag` checks sound a reasonable plan, most servers should support that.

bfontaine:
> `Etag` checks sound a reasonable plan, most servers should support that.

Yes, I checked on all core formulae’s stable URLs; 80% of them support this header.

Library/Homebrew/cmd/audit.rb
```diff
@@ -479,6 +479,19 @@ def audit_github_repository
     end
   end
 
+  def audit_tls_potential
+    # This only handles stable URLs at the moment. I'd like to give it some field
+    # testing on a relatively simple check before we expand it further.
+    return unless @online
+
+    url = formula.stable.url
+    unless url =~ %r[^https://]
```
bfontaine: You can use `url.start_with? "https://"`.

Library/Homebrew/cmd/audit.rb
```diff
@@ -479,6 +479,19 @@ def audit_github_repository
...
+      if system "curl", "--connect-timeout", "3", "--silent", "-I", "https://#{url.split("http://")[1]}", "-o", "/dev/null"
```
bfontaine: It’s probably better to use `url.sub("http://", "https://")` instead of a `split` here.

bfontaine:
I recently played with HTTPS checks to find HTTPS-able resources in formulae: https://github.com/bfontaine/homebrew-tools/blob/master/cmd/brew-https.rb

My checks might be broken but I found a lot of false positives especially for homepages with broken certificates. I also forgot to check if an URL ended with `.git` (for `HEAD`); most of those which I tested didn’t work with HTTPS.
I haven’t tried with ETags; maybe a first solution would be to try to download the resource using `https` and check that the `sha256` matches.

mikemcquaid:
@bfontaine That sounds :ok: to me :+1:

tdsmith:
How's this going?

DomT4:
Right now it's effectively dead. I am planning to move it over to the new repo at some point, but if anyone wants to leap over me I don't mind.

PR #41615, opened 2015-07-12
openssl: return to keg_only

### LINKED ISSUE #41613: OpenSSL linking issues due to removal of keg_only status
The homebrew OpenSSL recipe is no longer keg only (as of https://github.com/Homebrew/homebrew/commit/9ca3c054c070cc825870a6fc0fc5c5a22f7d8559#diff-5f8300a99f5c14b861dfa8eff2905a16) due to El Capitan no longer shipping headers.

Unfortunately, this leads to some major problems with projects (outside homebrew) that attempt to compile against OpenSSL without having the homebrew prefix added. Specifically, a project that links against OpenSSL and uses clang will see the homebrew OpenSSL headers due to the clang search path order:
```
#include <...> search starts here:
 /usr/local/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../lib/clang/6.1.0/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include
 /usr/include
 /System/Library/Frameworks (framework directory)
 /Library/Frameworks (framework directory)
```

However, when it links the clang library search path order is:
```
Library search paths:
	/usr/lib
	/usr/local/lib
Framework search paths:
	/Library/Frameworks/
	/System/Library/Frameworks/
```

So the binary compiles using the headers for 1.0.2d but links against 0.9.8. This problem typically manifests as a Symbol Not Found error at runtime.

This problem can be worked around by passing the appropriate LDFLAGS or CFLAGS but appears to represent a significant regression from the previous behavior. Is there any better workaround or is this expected?

DomT4:
I'd like us to further investigate if this is necessary, but it looks like it might be, and since the bot is nice and quiet at the moment I've booted the timeout to giggle levels and will let this run through.

Closes #41613, if we merge it.

### DIFF
--- a/Library/Formula/openssl.rb
+++ b/Library/Formula/openssl.rb
@@ -5,6 +5,7 @@ class Openssl < Formula
   mirror "https://raw.githubusercontent.com/DomT4/LibreMirror/master/OpenSSL/openssl-1.0.2d.tar.gz"
   mirror "https://www.mirrorservice.org/sites/ftp.openssl.org/source/openssl-1.0.2d.tar.gz"
   sha256 "671c36487785628a703374c652ad2cebea45fa920ae5681515df25d9f2c9a8c8"
+  revision 1
 
   bottle do
     sha256 "e9d8e2c7ab16a8516113093e999c8a6a1bc345badaa48af629963b97a7b9a1ef" => :yosemite
@@ -16,6 +17,9 @@ class Openssl < Formula
   option "without-check", "Skip build-time tests (not recommended)"
 
   depends_on "makedepend" => :build
+
+  keg_only :provided_by_osx,
+    "Apple has deprecated use of OpenSSL in favor of its own TLS and crypto libraries"
 
   def arch_args
     {
--- a/Library/Formula/bibtexconv.rb
+++ b/Library/Formula/bibtexconv.rb
@@ -3,6 +3,7 @@ class Bibtexconv < Formula
   homepage "https://www.uni-due.de/~be0001/bibtexconv/"
   url "https://www.uni-due.de/~be0001/bibtexconv/download/bibtexconv-1.1.3.tar.gz"
   sha256 "f7fd69ff99cd48a77e53aed4219defaf1f45485a07978faec01c2b9074886e03"
+  revision 1
 
   bottle do
     cellar :any
--- a/Library/Formula/bind.rb
+++ b/Library/Formula/bind.rb
@@ -5,6 +5,7 @@ class Bind < Formula
   mirror "https://fossies.org/linux/misc/dns/bind9/9.10.2-P2/bind-9.10.2-P2.tar.gz"
   version "9.10.2-P2"
   sha256 "b1e6f0af88634aaf48fb9d06bbf82968264f49b8e2685f061dd3fd4c1ab76c5f"
+  revision 1
 
   bottle do
     sha256 "63d408519604d63f2483f4e46a315a32a94347f372e69f7066d85f1d781e46a0" => :yosemite
--- a/Library/Formula/makepkg.rb
+++ b/Library/Formula/makepkg.rb
@@ -4,6 +4,7 @@ class Makepkg < Formula
   url "https://projects.archlinux.org/git/pacman.git",
       :tag => "v4.2.1",
       :revision => "068f8cec42057751f528b19cece37db13ae92541"
+  revision 1
 
   head "https://projects.archlinux.org/git/pacman.git"
 
--- a/Library/Formula/openwsman.rb
+++ b/Library/Formula/openwsman.rb
@@ -3,6 +3,7 @@ class Openwsman < Formula
   homepage "https://openwsman.github.io"
   url "https://github.com/Openwsman/openwsman/archive/v2.6.0.tar.gz"
   sha256 "449b4972dc7700918f8f674047f0a1a9c19eeee3f4ce82f47890dddfa56e823f"
+  revision 1
 
   bottle do
     sha256 "8a0fd930fef9bd3c27d87aba598559e40d2c8d1d9d13971b356e77f395766e27" => :yosemite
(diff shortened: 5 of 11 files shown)


DomT4:
@sjackman If we merge this, you may wish to merge an up-to-date Homebrew into Linuxbrew fairly quickly. See the linked issue for further detail; It'd be interesting to know if the problem impacts Linux as well as OS X.

tdsmith:
openssl has never been keg-only in Linuxbrew since Linuxbrew is never installed to /usr/local.

DomT4:
Oh, yup. Forgot Linuxbrew operates out of a different default prefix. Ignore the tired Dom.

tdsmith:
The consequence of pulling this is breaking anything (other than these packages) which depends_on openssl and has been built from source since 9ca3c054c070cc was committed at 5pm Thursday July 9, UK time. I don't know how big that exposure is.

DomT4:
Yeah. I double checked that earlier and there's breakage there for sure. I'm hoping with it being the weekend exposure will be minimal.

Worst-case scenario and we get a lot of failure reports: I was planning to deal with SSLv3 at some point in the not-completely-distant-future which would enforce a round of revisions, I could always bring that hastily forwards to deal with two breakages, one stone, etc.

DomT4:
Sigh. This old problem has returned:

```
==> Installing wsmancli dependency: openwsman
==> Downloading https://homebrew.bintray.com/bottles/openwsman-2.6.0_1.mountain_lion.bottle.tar.gz
Error: Failed to download resource "openwsman"
Download failed: https://homebrew.bintray.com/bottles/openwsman-2.6.0_1.mountain_lion.bottle.tar.gz
Error: /usr/local/opt/automake not present or broken
Please reinstall automake. Sorry :(
```

Otherwise seems fine.

Library/Formula/openssl.rb
```diff
@@ -17,6 +18,9 @@ class Openssl < Formula
 
   depends_on "makedepend" => :build
 
+  keg_only :provided_by_osx,
```
mikemcquaid: At the least this should be `keg_only :provided_pre_el_capitan` (the definition of which may need to be added) so this isn't keg-only there.

DomT4: @mistydemeo & @tdsmith seemed to raise objection to that in the chat - Will let them comment though.

mikemcquaid:
@Homebrew/owners I'd like to hold off merging this for a couple of days (even if this means blocking updates to `openssl` dependents in that time). I don't think a single user report warrants changing this back (yet). I also think the linker/include issue is perhaps a pointer to another issue. The reason why it's picking up the wrong libs in this case is because of `libssl.dylib` rather than `libssl.1.0.0.dylib` linkage. Perhaps it would be enough/sensible to kill the `libssl.dylib` symlink?

Another related piece of data is that thoughtbot/laptop (which is popular enough and intersects with our user-base enough for us to see issues related to it) has been force-linking `openssl` for a while: https://github.com/thoughtbot/laptop/blob/61b093344a6c6c1051783aedfe9fab58145d9224/mac#L157

DomT4:
If there's a way to not have to stuff OpenSSL back into its keg and at the same time ensure Clang doesn't break everything for giggles I'm all for it. I'd like to run the commands from the other thread in the 10.11 VM later today and see what the situation is there, It's probably the same but it'll help us decide on the 10.11 situation.

mikemcquaid:
Another thought (before I jump on a long flight): perhaps the weird linking situation (and adding it to the caveats) is worth the security benefits for the OS X ecosystem. Regardless, we should definitely make this not keg-only on El Capitan.

tdsmith:
I think it should remain keg-only on El Capitan. I think Misty was commenting on a different question.

Consider the case on El Capitan where a user builds something that `#include`s openssl/ssl.h and links with -lssl without further modifying the compiler search path. This cannot work, but the failure modes are different depending on whether openssl is keg-only.

If openssl is keg-only, the build fails because openssl/ssl.h can't be found. I think this is easy to understand and troubleshoot.

If openssl is linked, either: a) the build succeeds (with Homebrew's ssl configuration and a link to the system library, because /usr/lib is earlier in the library search path), which probably fails at runtime — this is particularly likely to be the case for extension modules or b) the build fails with an error about missing symbols.

I think the openssl-linked failure modes are harder to reason about if you don't know that OS X ships the openssl libraries but not the includes. The only benefit I see of linking openssl is making the `openssl` command more easily accessible, and maybe we can just link that in postinstall.

mikemcquaid:
@tdsmith Yeh, fair enough. I still wonder why we've not been seeing issues with this when OpenSSL has been linked? Are people just linking against the old version and not noticing/caring? I'm coming around to the idea that we just keep this keg-only.

DomT4:
Self-reminder: Add `bind` to the revision list here if/when we decide to go ahead with this.

mikemcquaid:
@DomT4 Let's do it (whenever you get a chance).

DomT4:
Cool. Will set a new build with bind going shortly, and merge it ASAP after that.

sjackman:
@domt4 `keg_only :provided_by_osx` is ignored by Linuxbrew, but thanks for the heads up.

DomT4:
@sjackman Yeah, Tim reminded me. Apologies for the false warning :)

sjackman:
Linuxbrew definitely has issues with the include search path and the linker search path getting out of sync, particularly when packages in their build scripts do dumb things like e.g. add `/usr/lib` to the search path but don't add `/usr/include` to the search path in the same position. I never fully realized just how crazy this situation is of the header and library search paths being independently configurable. It creates all sorts of havoc.

DomT4:
The only thing failing here is the unrelated bottle fetch for testing failure in https://github.com/Homebrew/homebrew/pull/41615#issuecomment-120687451. Pulling shortly.

PR #41617, opened 2015-07-12
nmh 1.6 (new formula)

benfb:
adds support for the nmh message handling system

http://www.nongnu.org/nmh/

### DIFF
--- a/Library/Formula/nmh.rb
+++ b/Library/Formula/nmh.rb
@@ -1,0 +1,32 @@
+class Nmh < Formula
+  desc "Powerful electronic mail handling system"
+  homepage "http://www.nongnu.org/nmh/"
+  url "http://download.savannah.nongnu.org/releases/nmh/nmh-1.6.tar.gz"
+  sha256 "29338ae2bc8722fe8a5904b7b601a63943b72b07b6fcda53f3a354edb6a64bc3"
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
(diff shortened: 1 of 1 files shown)


bfontaine:
The build is failing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/29446/version=mountain_lion/testReport/junit/brew-test-bot/mountain_lion/install_nmh/

Library/Formula/nmh.rb
```diff
@@ -0,0 +1,23 @@
...
+  url "http://download.savannah.nongnu.org/releases/nmh/nmh-1.6.tar.gz"
+  sha256 "29338ae2bc8722fe8a5904b7b601a63943b72b07b6fcda53f3a354edb6a64bc3"
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+    system "install-mh -auto"
```
dunn: This is where the test build fails. Since it's after the `make install` step, you might need to change this line to `system bin/"install-mh", "-auto"`

DomT4:
Ping.

benfb:
I looked at the package more and it looks like it's not a good fit for Homebrew...perhaps more of a meta package at some point down the line.

fakedrake:
@benfb mind explaining in what way it is not a good fit for homebrew?

PR #41625, opened 2015-07-12
Added devel version of aspell

keithpitty:
As per instructions in comment in https://github.com/Homebrew/homebrew-devel-only/pull/20.


keithpitty:
@tdsmith would you mind checking if this is ok to merge?

tdsmith:
Thanks for the PR! I added a `version` line since Homebrew was parsing the URL wrong.

PR #41626, opened 2015-07-12
stack 0.1.2.0 (new formula)

tkawachi:
stack is a cross-platform program for developing Haskell projects.
It is aimed at Haskellers both new and experienced.

### DIFF
--- a/Library/Formula/stack.rb
+++ b/Library/Formula/stack.rb
@@ -1,0 +1,10 @@
+class Stack < Formula
+  desc "The Haskell Tool Stack"
+  homepage "https://www.stackage.org/"
+  url "https://github.com/commercialhaskell/stack/releases/download/v0.1.2.0/stack-0.1.2.0-x86_64-osx.gz"
+  sha256 "6e1039d9c5144fb03dbfb1f569a830724593191305998e9be87579d985feb36c"
+
+  def install
+    bin.install "stack-0.1.2.0-x86_64-osx" => "stack"
+  end
+end


Library/Formula/stack.rb
```diff
@@ -0,0 +1,10 @@
+class Stack < Formula
+  desc "The Haskell Tool Stack"
+  homepage "https://www.stackage.org/"
+  url "https://github.com/commercialhaskell/stack/releases/download/v0.1.2.0/stack-0.1.2.0-x86_64-osx.gz"
```
bfontaine: We prefer to build from source.

tkawachi: It requires older version of GHC than what Homebrew provides. I don't know how to use older one (is it possible?).

https://github.com/commercialhaskell/stack/wiki/Downloads#os-x
> Note: due to GHC bug 10322, building stack from source fails with GHC 7.10.1. This bug will be fixed in 7.10.2, but in the meantime, we recommend using GHC 7.8.4 on OS X.

bfontaine: We keep some versioned formulae in [Homebrew/versions](https://github.com/Homebrew/homebrew-versions) but unfortunately GHC isn’t here, so we can’t use an older version.

bfontaine: You should submit this to [Homebrew/binary](https://github.com/Homebrew/homebrew-binary) if we can’t build it from source. We’ll be able to move it in the core in the next version if the bug is fixed.

DomT4:
See also: #40534

tkawachi:
Submitted the formula to homebrew-binary. https://github.com/Homebrew/homebrew-binary/pull/242

PR #41628, opened 2015-07-12
shiboken: update homepage and head url

vszakats:
Head URL points to the "unofficial mirror" now, which 
appears to be stuck at an old revision, but the previous
`gitourious.org` URL is closed now and couldn't find 
a more recent one.


bfontaine:
Thank you @vszakats! :tada:

PR #41629, opened 2015-07-12
fish 2.2.0

asparagui:

PR #41621, opened 2015-07-12
Update fish recipe for fish 2.2.0 release

ridiculousfish:
fish 2.2.0, get it while it's hot

PR #41631, opened 2015-07-12
fish 2.2.0 + formula cleanup

geoff-codes:
This PR's better than #41629 and #41621 :wink:.
CC: @ridiculousfish @zanchey @bfontaine

### DIFF
--- a/Library/Formula/fish.rb
+++ b/Library/Formula/fish.rb
@@ -1,8 +1,8 @@
 class Fish < Formula
   desc "User-friendly command-line shell for UNIX-like operating systems"
   homepage "http://fishshell.com"
-  url "http://fishshell.com/files/2.1.2/fish-2.1.2.tar.gz"
-  sha1 "f7f8d8d26721833be3458b8113c74b747296ec0b"
+  url "http://fishshell.com/files/2.2.0/fish-2.2.0.tar.gz"
+  sha256 "a76339fd14ce2ec229283c53e805faac48c3e99d9e3ede9d82c0554acfc7b77a"
 
   bottle do
     revision 1
@@ -15,11 +15,8 @@ class Fish < Formula
     url "https://github.com/fish-shell/fish-shell.git", :shallow => false
 
     depends_on "autoconf" => :build
-    # Indeed, the head build always builds documentation
     depends_on "doxygen" => :build
   end
-
-  skip_clean "share/doc"
 
   def install
     system "autoconf" if build.head?
@@ -29,20 +26,20 @@ class Fish < Formula
     system "make", "install"
   end
 
-  def post_install
-    system "pkill fishd || true"
-  end
-
-  test do
-    system "#{bin}/fish", "-c", "echo"
-  end
-
   def caveats; <<-EOS.undent
     You will need to add:
       #{HOMEBREW_PREFIX}/bin/fish
     to /etc/shells. Run:
       chsh -s #{HOMEBREW_PREFIX}/bin/fish
     to make fish your default shell.
+
+    If you are upgrading from an older version of fish, you should now run:
+      killall fishd
+    to terminate the outdated fish daemon.
     EOS
+  end
+
+  test do
+    system "#{bin}/fish", "-c", "echo"
   end
 end


Library/Formula/fish.rb
```diff
@@ -15,26 +15,19 @@ class Fish < Formula
     url "https://github.com/fish-shell/fish-shell.git", :shallow => false
 
     depends_on "autoconf" => :build
-    # Indeed, the head build always builds documentation
     depends_on "doxygen" => :build
```
geoff-codes: This comment is incorrect; it _will build and install_ without docs. But we still want to build them.

Library/Formula/fish.rb
```diff
@@ -15,26 +15,19 @@ class Fish < Formula
...
   end
 
-  skip_clean "share/doc"
```
geoff-codes: No longer necessary.

Library/Formula/fish.rb
```diff
@@ -15,26 +15,19 @@ class Fish < Formula
...
-
   def install
     system "autoconf" if build.head?
     # In Homebrew's 'superenv' sed's path will be incompatible, so
     # the correct path is passed into configure here.
     system "./configure", "--prefix=#{prefix}", "SED=/usr/bin/sed"
     system "make", "install"
-  end
 
-  def post_install
```
geoff-codes: No longer necessary.

Library/Formula/fish.rb
```diff
@@ -15,26 +15,19 @@ class Fish < Formula
...
-    system "pkill fishd || true"
-  end
-
-  test do
-    system "#{bin}/fish", "-c", "echo"
+    # Generate an inital set of manpage completions.
+    ENV["XDG_DATA_HOME"] = "#{etc}/share/completions"
```
geoff-codes: Its weird, but that's how to set the output directory.

zanchey: I don't think this will do what you want, and also I would recommend against it - may create problems in multiuser environments, for example. fish runs `fish_update_completions` in the background on first launch anyway.

kevinSuttle:
:arrow_up: Ideas, @ridiculousfish?

Library/Formula/fish.rb
```diff
@@ -15,26 +15,19 @@ class Fish < Formula
...
   def install
     system "autoconf" if build.head?
     # In Homebrew's 'superenv' sed's path will be incompatible, so
     # the correct path is passed into configure here.
     system "./configure", "--prefix=#{prefix}", "SED=/usr/bin/sed"
     system "make", "install"
-  end
 
-  def post_install
-    system "pkill fishd || true"
```
ridiculousfish: I would like to leave this in for this release. If fishd is running and hung, then users will get the "Old instances of fish are running" message until they manually kill fishd.

Library/Formula/fish.rb
ridiculousfish: Agree, this seems sketchy.

DomT4: Let's keep the test how it was.

Library/Formula/fish.rb
DomT4: Is there absolutely no other way to get around the potential problem? Killing users' running processes rubs up against Homebrew's philosophy a little.

ridiculousfish: There's probably precedent for handling this in other formulae. How does homebrew handle updating software that has a daemon?

DomT4: Often daemonised things ship a `plist`, which we instruct users to unload and then reload after update/upgrade.

ridiculousfish: To be clear you mean that the formula outputs a message to the console like "Please run launchctl unload ..." and the user manually copy-pastes and runs it?

DomT4: [Yup](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/caveats.rb#L174-L187). Theoretically, if you wanted to add a `caveat` to the existing block with something like:

```
Before you run fish you must execute:
  `pkill fishd || true`
```

I'd be cool with that, at least.

Library/Formula/fish.rb
geoff-codes: OK.

Library/Formula/fish.rb
geoff-codes: @DomT4 That's what I'll do.

ridiculousfish: I think we can omit it. It ought to be unnecessary except in unusual cases that may not even exist.

geoff-codes:
Squashed and pushed:
- Add instructions to kill any fishd processes if upgrading.
- Removed the bit I added to generate completions.

DomT4:
:boat:'ed in https://github.com/Homebrew/homebrew/commit/8a108bed154ae79a766d59949f0e753484525a76. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

kevinSuttle:
Woot! Thanks all!

sanity:
Very impressive how quickly this release got into brew - thanks!

PR #41638, opened 2015-07-12
Mutt sidebar patch

ephexeve:
Mutt with sidebar patch.

### DIFF
--- a/Library/Formula/mutt.rb
+++ b/Library/Formula/mutt.rb
@@ -51,33 +51,43 @@ class Mutt < Formula
   depends_on "s-lang" => :optional
   depends_on "gpgme" => :optional
 
-  patch do
-    url "ftp://ftp.openbsd.org/pub/OpenBSD/distfiles/mutt/trashfolder-1.5.22.diff0.gz"
-    sha1 "c597566c26e270b99c6f57e046512a663d2f415e"
-  end if build.with? "trash-patch"
+  if build.with? "trash-patch"
+    patch do
+      url "ftp://ftp.openbsd.org/pub/OpenBSD/distfiles/mutt/trashfolder-1.5.22.diff0.gz"
+      sha256 "ce964144264a7d4f121e7a2692b1ea92ebea5f03089bfff6961d485f3339c3b8"
+    end
+  end
 
   # original source for this went missing, patch sourced from Arch at
   # https://aur.archlinux.org/packages/mutt-ignore-thread/
-  patch do
-    url "https://gist.githubusercontent.com/mistydemeo/5522742/raw/1439cc157ab673dc8061784829eea267cd736624/ignore-thread-1.5.21.patch"
-    sha1 "dbcf5de46a559bca425028a18da0a63d34f722d3"
-  end if build.with? "ignore-thread-patch"
-
-  patch do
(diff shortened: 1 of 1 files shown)


Library/Formula/mutt.rb
```diff
@@ -72,6 +73,11 @@ class Mutt < Formula
     sha1 "94da52d50508d8951aa78ca4b073023414866be1"
   end if build.with? "confirm-attachment-patch"
 
+  patch do
+    url "http://lunar-linux.org/~tchan/mutt/patch-1.5.23.sidebar.20140412.txt"
+    sha1 "8b86f148d8e1a0b37b0b77268349f4c94e797561"
```
DomT4: This should be `sha256`.

Library/Formula/mutt.rb
```diff
@@ -72,6 +73,11 @@ class Mutt < Formula
...
+  end if build.with? "sidebar-patch"
```
DomT4: I know the rest of the formula is styled differently, but we're migrating away from `end if build.with? "sidebar-patch"` and such.

Should be:

```ruby
if build.with? "sidebar-patch"
  patch do
    url "http://lunar-linux.org/~tchan/mutt/patch-1.5.23.sidebar.20140412.txt"
    sha256 "1234"
  end
end
```

Thanks!

ephexeve: No worries. How about now?

DomT4: Huge :heart: for changing them all. Thank you!

ephexeve: Anytime ;-)

DomT4:
There's a little whitespacing on the lines in-between the patches apparently:

```
mutt:
 * 69: Trailing whitespace was found
 * 76: Trailing whitespace was found
 * 83: Trailing whitespace was found
 * 90: Trailing whitespace was found
```

ephexeve:
@DomT4, yup, forgot to do an audit before. Fixed now.

DomT4:
There's a comment at the top of the formula about not adding new patches to this formula, and this patch has been removed before, but there's also a decent chunk of people wandering around online asking how to install the patch on OS X.

The patch here was removed previously because it wasn't getting regular updates on new `mutt` releases and would just break. Has that changed?

CC @Homebrew/owners for Mutt patching opinions.

tdsmith:
I think the text at the head of the formula is clear. If someone wants to set up a patchy-mutt tap, that would be a useful service to the mutt community.

Thanks for the PR; let us know if you have any questions about setting up a tap. In the simplest case, you can put `mutt.rb` in a Github repository named `homebrew-patchy-mutt` or similar (`homebrew-` is important) and then install it with e.g. `brew install ephexeve/patchy-mutt/mutt`.

ephexeve:
@DomT4, well, afaik the sidebar-patch only works with mutt 1.5.23 (which is the version Homebrew is using for sometime now). Archlinux AUR also has [mutt-sidebar](https://aur.archlinux.org/packages/mutt-sidebar/) which is also using 1.5.23

MacPorts has ```mutt-devel``` which ships with mutt-sidebar as a variant.

If that comment is still valid or not, I don't know.

PR #41642, opened 2015-07-12
freeimage: use canonical domain for homepage

vszakats:
Also:
* try homepage with https first, which will be redirected 
to plaintext for non-logged-in users. Before this commit 
the ratio of https vs. http protocol access for sf.net top 
domain in formulae is: 11:41.
* styling nits


bfontaine:
Thank you @vszakats! :smile:

PR #41643, opened 2015-07-12
moreutils: use https homepage

vszakats:
(serving mixed-content sadly - it'd be enough 
to reference flattr image via https for a fix)


DomT4:
Mixed content is their bug rather than ours, so happy on this. Merged in https://github.com/Homebrew/homebrew/commit/9d2c4d46764c37980302455533e4e456f9742b93, Thanks!

PR #41645, opened 2015-07-13
kjell 0.2.6

dunn:


dunn:
```
==> script -q /dev/null /usr/local/Cellar/kjell/0.2.6/bin/kjell -sanity_check true

=ERROR REPORT==== 12-Jul-2015::18:14:37 ===
Error in process <0.35.0> with exit value: {badarg,[{erlang,register,[user,<0.36.0>],[]},{user,start_port,1,[{file,"user.erl"},{line,53}]}]}

Error: kjell: failed
execution expired
```
Weird. Locally, this only happens without `script -q /dev/null`, which is why I added it.

karlll:
Hi! Do you think you can add the flags '-nouser' and '-noinput' to the test invocation to see if it works any better?

   `/usr/local/Cellar/kjell/0.2.6/bin/kjell -sanity_check true -nouser -noinput`

karlll:
Hi! I've noticed that the I get the same behaviour as the failing CI job if I run the sanity check with the TERM env variable unset. I looked in the logs for the jobs and it seems that there are no such variable in the environment (since its running headless).

Could you perhaps try the following test:

`system %{TERM=xterm script -q /dev/null #{bin}/kjell -sanity_check true}`

bfontaine:
You can set an environment variable in the test with `ENV["TERM"] = "xterm"`.

dunn:
That was it, thanks @karlll!

bfontaine:
Thank you @dunn!

PR #41646, opened 2015-07-13
inkscape 0.91 (new formula)

dunn:
Now that inkscape doesn't depend on X11 it can return to Homebrew core.

https://github.com/Homebrew/homebrew-x11/pull/102

### DIFF
--- a/Library/Formula/inkscape.rb
+++ b/Library/Formula/inkscape.rb
@@ -1,0 +1,69 @@
+class Inkscape < Formula
+  desc "A professional vector graphics editor"
+  homepage "https://inkscape.org/"
+  url "https://inkscape.org/en/gallery/item/3854/inkscape-0.91.tar.gz"
+  mirror "https://mirrors.kernel.org/debian/pool/main/i/inkscape/inkscape_0.91.orig.tar.gz"
+  sha256 "2ca3cfbc8db53e4a4f20650bf50c7ce692a88dcbf41ebc0c92cd24e46500db20"
+
+  devel do
+    url "lp:inkscape/0.92.x", :using => :bzr
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+  end
+
+  head do
+    url "lp:inkscape", :using => :bzr
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+  end
+
+  depends_on "boost-build" => :build
+  depends_on "intltool" => :build
+  depends_on "pkg-config" => :build
+  depends_on "poppler" => :optional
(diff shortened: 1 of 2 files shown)


mistydemeo:
Looks like the `--devel` build isn't working, though that's perhaps normal since it's a floating target.

dunn:
Most of the failures are server errors; looks like launchpad is down right now.

DomT4:
I can fetch the `devel` link here, but it is *incredibly* slow at the moment.

There's a lot of stuff in X11 that no longer actually needs X11, I don't know what the plan is on how to handle those cases - Mike was talking about a GUI tap a while back. I'm cool with it returning to the core though.

dunn:
Gotcha, I'll sit on this until y'all decide where the best place for this is.

dunn:
I'm getting the impression that a GUI tap isn't imminent. Mind if I move ahead with this?

DomT4:
No objections here.

mikemcquaid:
I'd really like to keep this out of core.

mikemcquaid:
What advantages does this provide over the `brew cask` version? (FWIW I'd like to ask this question about anything we package that they do too)

dunn:
> What advantages does this provide over the `brew cask` version?

I'm not sure if the Cask version has Poppler support.  It also requires X11: https://inkscape.org/en/download/mac-os/

mistydemeo:
Our formula also has a head spec and could have a devel in the future.

mikemcquaid:
Ok, fair enough. I dunno, my main objection to formulae like this (and even `macvim`, honestly) is that they seem to break so often for users that I'm not sure we provide a good user experience. Personally I think stuff that's failing often for users should be removed from core entirely (or at least have the options pared down). Maybe that's just me, though.

mikemcquaid:
> I'm getting the impression that a GUI tap isn't imminent. Mind if I move ahead with this?

I've created one at https://github.com/Homebrew/homebrew-gui and hooked it up to CI. I guess maintainers may want to debate on what should be in it, whether `homebrew-app` for app-bundle'y things is better or what.

bfontaine:

mikemcquaid:
I'd like if this went to `homebrew-gui`'s tap.

DomT4:
The GUI tap is private at the moment. Didn't want people to start submitting stuff there before we actually decided what belongs there. Feel free to reverse that as desired.

mikemcquaid:
@dunn It'd be great to work with Cask/upstream to try and make their version better.

dunn:
Cask just has to work with whatever binaries upstream provides, right?

mikemcquaid:
Yeh. They sometimes build stuff themselves and upload that.

PR #41649, opened 2015-07-13
[formula: natalie] Compile natalie.swift for performance

ldiqual:
This commit adds the compilation of `natalie.swift` during the installation process, resulting in a greatly improved execution time. This is very important because when `natalie` is part of an app build process, having a fast build phase for generating Storyboard code greatly improves build times.

Using the original `natalie.swift` script:
```
$ time ./natalie.swift NatalieExample/
// Output
3.04 real         2.85 user         0.12 sys
```

Using a compiled version:
```
$ time natalie NatalieExample/
// Output
0.07 real         0.01 user         0.03 sys
```

@krzyzanowskim (original author of the program): Let me know what you think.

### DIFF
--- a/Library/Formula/natalie.rb
+++ b/Library/Formula/natalie.rb
@@ -8,14 +8,15 @@ class Natalie < Formula
   depends_on :xcode => "6.3"
 
   def install
-    bin.install "natalie.swift"
+    system "xcrun -sdk macosx swiftc -O natalie.swift -o natalie"
+    bin.install "natalie"
     share.install "NatalieExample"
   end
 
   test do
     example_path = "#{share}/NatalieExample"
     output_path = testpath/"Storyboards.swift"
-    generated_code = `#{bin}/natalie.swift #{example_path}`
+    generated_code = `#{bin}/natalie #{example_path}`
     output_path.write(generated_code)
     line_count = `wc -l #{output_path}`
     assert line_count.to_i > 1
--- a/Library/Formula/natalie.rb
+++ b/Library/Formula/natalie.rb
@@ -8,8 +8,8 @@ class Natalie < Formula
   depends_on :xcode => "6.3"
 
   def install
-    system *%W[mv natalie.swift natalie-script.swift]
-    system *%W[xcrun -sdk macosx swiftc -O natalie-script.swift -o natalie.swift]
+    system "mv", "natalie.swift", "natalie-script.swift"
+    system "xcrun", "-sdk", "macosx", "swiftc", "-O", "natalie-script.swift", "-o", "natalie.swift"
     bin.install "natalie.swift"
     share.install "NatalieExample"
   end
--- a/Library/Formula/natalie.rb
+++ b/Library/Formula/natalie.rb
@@ -8,15 +8,16 @@ class Natalie < Formula
   depends_on :xcode => "6.3"
 
   def install
-    system "xcrun -sdk macosx swiftc -O natalie.swift -o natalie"
-    bin.install "natalie"
+    system *%W[mv natalie.swift natalie-script.swift]
+    system *%W[xcrun -sdk macosx swiftc -O natalie-script.swift -o natalie.swift]
+    bin.install "natalie.swift"
     share.install "NatalieExample"
   end
 
   test do
     example_path = "#{share}/NatalieExample"
     output_path = testpath/"Storyboards.swift"
-    generated_code = `#{bin}/natalie #{example_path}`
+    generated_code = `#{bin}/natalie.swift #{example_path}`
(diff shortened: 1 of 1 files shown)


Library/Formula/natalie.rb
```diff
@@ -8,14 +8,15 @@ class Natalie < Formula
   depends_on :xcode => "6.3"
 
   def install
-    bin.install "natalie.swift"
+    system "xcrun -sdk macosx swiftc -O natalie.swift -o natalie"
```
tdsmith: "split", "like", "this" or use `%W[]`, please

Library/Formula/natalie.rb
```diff
@@ -8,14 +8,15 @@ class Natalie < Formula
...
+    bin.install "natalie"
```
tdsmith: let's also install `natalie.swift` so we don't break installed uses

tdsmith: (you can pass multiple arguments to install, like `bin.install "natalie", "natalie.swift"`)

krzyzanowskim:
Ok. Just please keep both versions installed (as suggested by @tdsmith)
"natalie" <- compiled
"natalie.swif" <- script version

mikemcquaid:
@krzyzanowskim What's the advantage of running the slower version?

krzyzanowskim:
@mikemcquaid surely that compiled version is faster, on the other hand this is still a script, and I want to keep it this way (at least for a while). It thought it may be good idea to have it installed this way for consistency with the current state of the project (and docs). 

Now I think that maybe nobody will ever check if executed file is binary or script so maybe not a big deal at all.

Anyway, executable shouldn't be renamed to "natalie".

mikemcquaid:
@krzyzanowskim Up to you which but I'd really rather we picked only one to install into `bin`. The other could always go in `libexec` or `share` or something.

dunn:
> Anyway, executable shouldn't be renamed to "natalie".

Why not?  It's the name of the formula, so it seems like the most natural thing IMO.

krzyzanowskim:
@dunn that would introduce unnecessary complication: why path is different when installed with brew. brew is to make the process easier, not harder.

krzyzanowskim:
ok, pls go with compiled bin/natalie.swift 

btw @ldiqual check this https://github.com/krzyzanowskim/Natalie/issues/60 for performance tip.

ldiqual:
@krzyzanowskim @tdsmith @mikemcquaid Fixed.

@mikemcquaid I tried `system "xcrun", "-sdk macosx", "swiftc", "-O", "natalie-script.swift", "-o natalie.swift"` but it failed with the following error:

```
==> xcrun -sdk macosx swiftc -O natalie-script.swift -o natalie.swift
  -k, --kill-cache            invalidate all existing cache entries
  --show-sdk-path             show selected SDK install path
  --show-sdk-version          show selected SDK version
  --show-sdk-platform-path    show selected SDK platform path
  --show-sdk-platform-version show selected SDK platform version
```

So it seems that `xcrun` doesn't support the way the arguments are passed behind the scenes, any idea why?

tdsmith:
Keep splitting: `system "xcrun", "-sdk", "macosx", "swiftc", "-O", "natalie-script.swift", "-o", "natalie.swift"` -- you have to do all the tokenization that the shell would normally do for you, and the shell splits on whitespace unless it's quoted

ldiqual:
@tdsmith Fixed. Just for my own curiosity: having a long string worked and it seemed clearer to me, but I'm sure there is a reason why it's better to do it your way. Would you mind linking me to a previous discussion or elaborate on why it's better? Again, just curious.

mikemcquaid:
> @tdsmith Fixed. Just for my own curiosity: having a long string worked and it seemed clearer to me, but I'm sure there is a reason why it's better to do it your way. Would you mind linking me to a previous discussion or elaborate on why it's better? Again, just curious.

We prefer that style because it makes it doesn't create an unnecessary subshell and prevents accidental arbitrary command execution.

Library/Formula/natalie.rb
```diff
@@ -8,6 +8,8 @@ class Natalie < Formula
   depends_on :xcode => "6.3"
 
   def install
+    system "mv", "natalie.swift", "natalie-script.swift"
```
DomT4: Shouldn't need to call `system` on `mv`.

mikemcquaid: Ping.

ldiqual:
@mikemcquaid Thanks for the explanations!

@DomT4 Fixed

dunn:
Thanks! Merged in 1220934c32742f4496c9728fc213dffe90f98071.

PR #41663, opened 2015-07-13
keg_relocate: relocate all text files.

mikemcquaid:
Work out what's text and what's not using `file`. Also, rename `keg_fix_install_names` to `keg_relocate` because that's a more accurate description of what it does now.

CC @xu-cheng

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -1,5 +1,5 @@
 require "extend/pathname"
-require "keg_fix_install_names"
+require "keg_relocate"
 require "formula_lock"
 require "ostruct"
 
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -38,10 +38,7 @@ class Keg
       end
     end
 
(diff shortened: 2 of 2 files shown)


xu-cheng:
Sorry for the late response, I'm extremely busy this week.

To me, I don't think it's a good idea to relocate all text files, especially for documents. I would prefer to add new file type when we see it stoping the relocation. Or I would suggest to add `next if Metafiles::EXTENSIONS.include? pn.extname` to skip known documents

mikemcquaid:
> I would prefer to add new file type when we see it stoping the relocation.

I disagree strongly. This just creates a lot of busywork and code when a generalised solution achieves the same.


> Or I would suggest to add next if Metafiles::EXTENSIONS.include? pn.extname to skip known documents

I'm :ok: with this.

mikemcquaid:
@xu-cheng Let me know what you think now.

xu-cheng:
:+1: LGTM

andyli:
ocaml is broken -> #41926

mikemcquaid:
Fixed the `gettext` case in b8a48d329eaf31d4e002b0f55e2eb5ae345f7b16.

PR #41669, opened 2015-07-13
Build llvm with optimizations and NDEBUG flag

chfast:
I don't know why, but std_cmake_args clears default release flags for the compiler (`-DCMAKE_CXX_FLAGS_RELEASE=`). As the result the LLVM is built without optimization (-O3) and NDEBUG macro being set (-DNDEBUG).

### DIFF
--- a/Library/Formula/llvm.rb
+++ b/Library/Formula/llvm.rb
@@ -130,6 +130,8 @@ class Llvm < Formula
     args = %w[
       -DLLVM_OPTIMIZED_TABLEGEN=On
     ]
+    
+    args << "-DCMAKE_CXX_FLAGS_RELEASE='-O3 -DNDEBUG'"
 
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> As the result the LLVM is built without optimization

This isn't actually the case as it's handled by our compiler wrappers internally rather than the CMake buildsystem. OK to add NDEBUG though.

chfast:
Is it ok now?

mikemcquaid:
@chfast Sorry, @denji isn't a maintainer and I was on vacation so will just merge the old syntax.

mikemcquaid:

chfast:
Thanks, syntax does not matter to me. I don't know ruby, so if you have any preferences about the syntax it's ok to correct me.

PR #41670, opened 2015-07-13
Don't link fontforge extension modules to libpython

tdsmith:
Let's just pull this for now since it's been discussed upstream in fontforge/fontforge#2353 and fontforge/fontforge#2356.

### DIFF
--- a/Library/Formula/fontforge.rb
+++ b/Library/Formula/fontforge.rb
@@ -50,17 +50,26 @@ class Fontforge < Formula
   end
 
   def install
-    if MacOS.version <= :snow_leopard || !build.bottle?
-      pydir = `python-config --prefix`.chomp
-    else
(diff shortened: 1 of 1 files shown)


DomT4:
Any chance you can chuck the version bump in here as well from #40631?

tdsmith:
I think I won't, but you may ;)

tdsmith:
(I'm worried that we'll have to fuss with the detection code and I should be doing other things today.)

DomT4:
Since we're not adding a revision or anything here I'll chase in your PR with the version bump once this is merged :)

tdsmith:
Oh, the detection code is already in this version; I'll go for the newer version.

DomT4:
@tdsmith :heart:

PR #41673, opened 2015-07-13
cdrtools devel 3.01a30

fbrosson:
* Updated the formula description, as cdrtools also supports burning DVDs since 1998 and Blu-ray discs since 2007.
* Switched from 3.01a28 to 3.01a30, which is the latest preview release of cdrtools.
* Dropped, for devel only, the patch that inserts a test for HAVE_PRAGMA_WEAK, as it is only required when building cdrtools with Clang 3.1 which is now quite outdated since several newer versions of Clang have been released.

### DIFF
--- a/Library/Formula/cdrtools.rb
+++ b/Library/Formula/cdrtools.rb
@@ -1,16 +1,14 @@
-require 'formula'
-
 class Cdrtools < Formula
-  desc "ISO 9660 file system and CD creation tools"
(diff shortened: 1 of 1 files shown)


bfontaine:
Thank you @fbrosson for your contribution to Homebrew! :tada: 

For the record, the preferred commit message format for devel versions is “`cdrtools 3.01a30 (devel)`”.

PR #41681, opened 2015-07-14
Create minisign.rb

reelsense:
A dead simple tool to sign files and verify signatures. Signature written by minisign can be verified using OpenBSD's signify tool: public key files and signature files are compatible.


reelsense:
I guess I forgot "test do"

PR #41683, opened 2015-07-14
cmake: change --without-docs to --with-docs

L2G:
CMake already provides thorough documentation via the command line. Building the man pages additionally requires Sphinx (and its own dependencies) to be downloaded and installed.  Since all core formulae that depend on CMake use it as a build dependency, having the man pages built by default therefore seems like overkill.

### DIFF
--- a/Library/Formula/cmake.rb
+++ b/Library/Formula/cmake.rb
@@ -28,7 +28,7 @@ class Cmake < Formula
     sha256 "e1302ea14f218c07a94560d42d6f3cafa1942821896316a5555ca78b2597caad"
   end
 
-  option "without-docs", "Don't build man pages"
+  option "with-docs", "Build man pages"
 
   depends_on :python => :build if MacOS.version <= :snow_leopard && build.with?("docs")
 


DomT4:
We deliberately build documentation by default as much as possible so that users who use bottles, which is the majority of the user base, get said documentation without having to compile to obtain a fairly basic/expected element.

Sphinx and co are never actually installed with the CMake installation; they live in the buildpath, aren't transferred to libexec and are erased at the end of compile.

L2G:
Did you note the part where I said, "CMake already provides thorough documentation via the command line"?

DomT4:
Yup. Doesn't stop Debian or Ubuntu shipping both HTML documentation and the manpages though.

tdsmith:
Thanks for the PR! I understand your rationale but I also prefer to build man pages by default because, in general, I expect packages to install them. Building the docs adds some complexity but not much time to the cmake build, and most users don't experience either penalty because cmake is bottled and `cellar :any`.

DomT4:
Passing on this, Sorry.

PR #41686, opened 2015-07-14
CHIRP 20150702 (new formula)

johanoskarsson:
CHIRP is a tool for programming an amateur radio.
They rely on daily builds instead of stable releases, hence 20150702.

### DIFF
--- a/Library/Formula/chirp.rb
+++ b/Library/Formula/chirp.rb
@@ -1,0 +1,16 @@
+class Chirp < Formula
+  desc "CHIRP is a free, open-source tool for programming your amateur radio."
+  homepage "http://chirp.danplanet.com/"
+  url "http://trac.chirp.danplanet.com/chirp_daily/daily-20150702/chirp-daily-20150702.tar.gz"
+  sha256 "41a5ae113bea27a284675ad5d95b45e4cd90f76ea2cd29050396d55bab5be433"
+
+  depends_on "python"
+  depends_on "pygtk"
+  depends_on "libxml2" => "with-python"
+
+  def install
+    prefix.install Dir["*"]
+
+    bin.install_symlink "#{prefix}/chirpw"
+  end
+end
--- a/Library/Formula/chirp.rb
+++ b/Library/Formula/chirp.rb
@@ -10,7 +10,11 @@ class Chirp < Formula
 
   def install
     prefix.install Dir["*"]
-
     bin.install_symlink "#{prefix}/chirpw"
+  end
+
+  test do
+    # unfortunately chirp returns error code 1 for --version
+    assert_match "CHIRP", shell_output("#{bin}/chirpw --version 2>&1", 1)
   end
 end


Library/Formula/chirp.rb
```diff
@@ -0,0 +1,16 @@
...
+  url "http://trac.chirp.danplanet.com/chirp_daily/daily-20150702/chirp-daily-20150702.tar.gz"
+  sha256 "41a5ae113bea27a284675ad5d95b45e4cd90f76ea2cd29050396d55bab5be433"
+
+  depends_on "python"
+  depends_on "pygtk"
+  depends_on "libxml2" => "with-python"
+
+  def install
+    prefix.install Dir["*"]
+
```
bfontaine: You can drop this empty line.

Library/Formula/chirp.rb
```diff
@@ -0,0 +1,16 @@
...
+    bin.install_symlink "#{prefix}/chirpw"
+  end
```
bfontaine: Please add a `test do` block.

bfontaine:

DomT4:
Test failing:

```
Error: chirp: failed
Expected /CHIRP/ to match "No handlers could be found for logger \"chirp.platform\"\n".
```

dunn:
Is every daily build stable? If not, maybe this should go in the devel-only tap.

DomT4:
> CHIRP is distributed as a series of automatically-generated builds. Any time we make a change to CHIRP, a build is created for it the next day. Thus, CHIRP is versioned by the date on which it was created, which makes it easy to determine if you have an older build. We don't put experimental things into CHIRP before they are ready, except where specifically called out with a warning. Thus, you do not need to worry about finding a stable version to run. You should always be on the latest build available.

I'm not sure this is necessarily a good fit for Homebrew if it's something most users are going to expect to see updated *daily*.

Library/Formula/chirp.rb
```diff
@@ -0,0 +1,20 @@
+class Chirp < Formula
+  desc "CHIRP is a free, open-source tool for programming your amateur radio."
+  homepage "http://chirp.danplanet.com/"
+  url "http://trac.chirp.danplanet.com/chirp_daily/daily-20150702/chirp-daily-20150702.tar.gz"
+  sha256 "41a5ae113bea27a284675ad5d95b45e4cd90f76ea2cd29050396d55bab5be433"
+
+  depends_on "python"
```
DomT4: We don't hardcode our Python.

```ruby
```

Library/Formula/chirp.rb
```diff
@@ -0,0 +1,20 @@
...
+  depends_on "pygtk"
+  depends_on "libxml2" => "with-python"
+
+  def install
+    prefix.install Dir["*"]
```
DomT4: Use `libexec.install` & then `bin.install_symlink libexec/"chirpw"`.

DomT4:
Going to pass on this because of the daily-update expectation as noted in https://github.com/Homebrew/homebrew/pull/41686#issuecomment-145303503. Just not sure that makes it a good fit for Homebrew, or provides a desirable user experience.

If they ever move onto more stable releases which don't expect to be updated daily, we'd certainly be happy to review this again. Thanks for the submission :heart:

Issue #35371, opened 2014-12-31
jpeg v9b (Discuss bump) (Anyone interested?)


DomT4:
There's been a new version of ` jpeg ` out for a while now, almost a year actually, but we haven't bumped to it yet.

The previous discussion on this was #17733. There was largely dread due to the bot not existing at that point, revisions being a longer rebuild process, and generally the need for maintainers to spend a long, long time sorting everything out in regards to testing changes and dealing with fallout. I guess at least, largely, issues 1 & 2 are negated by the existence of the wonderful bot these days.

One of the big hiccups is that ` jpeg ` v9 changes the ` dylib ` version name, and consequently, likely many if not all of the ` brew uses ` need to be revisioned to cope. 

```
dyld: Library not loaded: /usr/local/lib/libjpeg.8.dylib
  Reason: image not found
```

They would also need to be tested beforehand to check whether they will actually build against the newer version. We have [the previous newer version](https://github.com/Homebrew/homebrew-versions/blob/master/jpeg9.rb) in Homebrew/Versions, but that hasn't been updated to the new `.a ` bump.

The fact that, beyond the original creation of the ` jpeg9 ` formula, absolutely nobody has teed up a PR against it to either improve it or bump it to the ` .a ` release perhaps is telling on how much general urgency there is to move off the ` .8 ` release. In addition, the Core here doesn't seem to have taken much of a ` jpeg9 plz ` pounding, to say the least.

I guess we have the capacity to more easily bump the formula at this point, it's just whether or not there's the desire or even need to get it done.

For completeness, this is the current ` brew uses ` list. Apologies for the terrible formatting:

```
allegro5		gegl			imlib2			little-cms		povray36
analog			ghostscript		io			little-cms2		qemu
argyll-cms		gle			jasper			logstalgia		sane-backends
center-im		gnuplot			jp2a			mal4s			scantailor
dcraw			gource			jpeginfo		mapnik			sdl2_image
devil			gpac			jpegoptim		mapnik071		sdl_image
djvulibre		gphoto2			jpegrescan		minidlna		sfml
eet			grace			leptonica		mjpegtools		swftools
epeg			graphicsmagick		leptonica169		open-scene-graph	tiff2png
fbida			gst-plugins-good	libbpg			open-sg			ufraw
ffmpegthumbnailer	gst-plugins-good010	libgaiagraphics		openslide		vice
[...]

PR #41689, opened 2015-07-14
allegro: drop unused deps, adjust cmake build

L2G:
While tackling #35371, I noticed that Allegro was not linking to the jpeg or libpng libraries. I don't think this is a build configuration error, though; it seems as if Allegro just doesn't care about those libraries in OS X, preferring to use the standard OS X frameworks instead.

For that reason, I've dropped the dependencies on jpeg and libpng. I'm not sure if that matters for Linux, though; is linuxbrew still a consideration?

At the same time, I've added a step that creates a "build" subdirectory for CMake to do its building, since that seems to be the CMake way.

### DIFF
--- a/Library/Formula/allegro.rb
+++ b/Library/Formula/allegro.rb
@@ -30,14 +30,13 @@ class Allegro < Formula
   depends_on "libvorbis" => :recommended
   depends_on "freetype" => :recommended
   depends_on "flac" => :recommended
-  depends_on "libpng" => :recommended
-  depends_on "jpeg" => :recommended
   depends_on "physfs" => :recommended
 
   def install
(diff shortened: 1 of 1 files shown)


dunn:
Can't hurt to check with @sjackman about Linuxbrew.

tdsmith:
:+1:; as far as Homebrew/homebrew goes, Linuxbrew is a world unto itself.

sjackman:
I appreciate the heads up, Larry. Thanks.

DomT4:
I tend to take the viewpoint on the `build` directory that if the upstream CMake script doesn't error out when building directly in the tree we shouldn't bother "fixing" it. Merged this though since there's no harm. Thanks @L2G!

PR #41699, opened 2015-07-14
exercism: import from homebrew/binary

bfontaine:
See https://github.com/Homebrew/homebrew-binary/pull/244


bfontaine:
The build fails because of a name conflict with `homebrew/binary/exercism`, but that’s expected.

mistydemeo:
:+1:

PR #41704, opened 2015-07-14
ghq: use github archive url

decors:
also:
* add build flag


DomT4:
Why was it previously using the git checkout?

PR #41705, opened 2015-07-14
improve TapLoader and TapFormulaUnavailableError

xu-cheng:
* Restore the ability to load formula by `user/homebrew-repo/foo`.
* Only suggest to install tap when tap isn't installed.


xu-cheng:
cc @mikemcquaid

mikemcquaid:
:+1:

PR #41717, opened 2015-07-14
nginx 1.9.3 (devel)

nijikon:


Library/Formula/nginx.rb
```diff
@@ -121,12 +120,7 @@ def post_install
     end
   end
 
-  test do
-    system "#{bin}/nginx", "-t"
-  end
-
   def passenger_caveats; <<-EOS.undent
-
```
DomT4: This line is actually deliberate. Without it the two caveats will rub up against each other directly when both apply. The empty line ensures there's a gap between them. I'll pull it and add the line back though, no point making the bot do it again.

DomT4: (Not actually your fault on this. Our audit does flag it, and we should probably look at an alternative way to avoid them rubbing against each other).

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/31178bae68d18dcc1a602eb916e941c0fa07813a. Thanks again @nijikon! :beers:

PR #41721, opened 2015-07-14
hayai 0.1.0 (new formula)

scribblemaniac:
hayai is a C++ framework for writing benchmarks for pieces of code in the style of unit testing frameworks such as googletest. It makes it easy to properly benchmark code without cluttering up your main source files with timing functions and loops.

### DIFF
--- a/Library/Formula/hayai.rb
+++ b/Library/Formula/hayai.rb
@@ -1,0 +1,32 @@
+class Hayai < Formula
+  desc "A C++ benchmarking framework inspired by the googletest framework"
+  homepage "http://nickbruun.dk/2012/02/07/easy-cpp-benchmarking"
+  url "https://github.com/nickbruun/hayai/archive/v0.1.0.tar.gz"
+  sha256 "739d6dc5126a0b3dfe6c944999f48e7ede823bfdbf1f8801f23af25bf355db65"
+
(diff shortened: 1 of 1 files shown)


Library/Formula/hayai.rb
```diff
@@ -0,0 +1,34 @@
+class Hayai < Formula
+  desc "A C++ benchmarking framework inspired by the googletest framework"
+  homepage "http://nickbruun.dk/2012/02/07/easy-cpp-benchmarking"
+  url "https://github.com/nickbruun/hayai/archive/v0.1.0.tar.gz"
+  sha256 "739d6dc5126a0b3dfe6c944999f48e7ede823bfdbf1f8801f23af25bf355db65"
+
+  depends_on "cmake" => :build
+  depends_on "autoconf" => :build
```
DomT4: It looks like the autotools are only needed if we pass `-DTEST=true` to CMake?

scribblemaniac: That makes sense thanks. Fixed.

dunn:
Very cool, merged as 145d90a600c90dc81cbebeaaa9b81048eb4ef1ac. Thanks!

scribblemaniac:
@dunn Thank you :smile:

PR #41722, opened 2015-07-14
Fixed Xcode download location

zebMcCorkle:


DomT4:
We use the existing version in another couple places as well. Presumably, if we wanted to change one, we'd be better changing them all. Will let the other maintainers review this and we'll make a decision.

Given the current link presents as a big login screen tagged "Developer" when you can actually log in and download Xcode with any standard Apple ID, and we've seen people get confused by this, I don't mind the change to the friendlier initial Xcode page.

dunn:
Should we change them all at once?

mikemcquaid:
Yeh, let's update this globally, it's a much nicer page.

DomT4:
Other locations for this particular string in this context are `blacklist.rb`, `requirements.rb` and `doctor.rb`.

PR #41724, opened 2015-07-15
kvazaar 0.5.0 and kvazaar HEAD

kodabb:


bfontaine:
The `ffms2` build is failing on ML: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/29592/version=mountain_lion/console

```
…snip…
==> make install
  CXX      src/core/audiosource.lo
  CXX      src/core/ffms.lo
  CXX      src/core/filehandle.lo
  CXX      src/core/indexing.lo
In file included from src/core/filehandle.cpp:21:
src/core/filehandle.h:23:10: fatal error: 'cstdint' file not found
#include <cstdint>
         ^
In file included from src/core/indexing.cpp:21:
In file included from src/core/indexing.h:24:
src/core/utils.h:60:6: error: no type named 'unique_ptr' in namespace 'std'
std::unique_ptr<T> make_unique(Args&&... args) {
~~~~~^
…snip…
```

Note: It’s easier for us if you make one PR per formula.

kodabb:
ah I didn't know that, will split this up then :)

kodabb:
Hi, here is the kvazaar update

Library/Formula/kvazaar.rb
```diff
@@ -16,6 +18,11 @@ class Kvazaar < Formula
   def install
     system "make", "-C", "src"
     bin.install "src/kvazaar"
+
+    head do
```
dunn: Use `if build.head?` within the install block.

Library/Formula/kvazaar.rb
```diff
@@ -15,7 +17,9 @@ class Kvazaar < Formula
 
   def install
     system "make", "-C", "src"
-    bin.install "src/kvazaar"
+    bin.install_symlink "src/kvazaar"
```
dunn: Looks like nothing is getting installed during the stable build except a broken symlink:
```
kvazaar:
```

DomT4: Yeah. What's likely happening here is that it's symlinking the binary from the buildpath into the Cellar, and the second the buildpath stops existing, the link breaks. Maybe you wanted to do something like this instead?

```ruby
if build.head?
  system "make", "-C", "src", "install"
else
  system "make", "-C", "src"
  bin.install "src/kvazaar"
end
```

?

Although, it is likely that `install` needs a prefix argument passed to it so it doesn't dump it into whatever prefix is hardcoded upstream.

PR #41725, opened 2015-07-15
ignore this

DomT4:
Need something tiny to test Bintray changes with, and doesn't have an existing bottle. Just completely ignore this PR - It won't exist for more than a couple hours.


DomT4:
There are some remaining issues. I've taken that back upstream. Will lock this for now so nobody thinks it's actually a PR intended to be merged at any point.

PR #41741, opened 2015-07-15
docker-compose 1.3.2

moutten:
In order to upgrade to docker-compose 1.3.2, both the websocket-client and docker-py packages need to be updated as well.


xu-cheng:

PR #41746, opened 2015-07-15
kitchen-sync 0.44

denji:


dunn:
Any chance of a more comprehensive test? 😁

denji:
@dunn I think it is not very feasible because of homebrew. For in-depth tests to run mysql and postgresql once to initialize the database, install ruby gems and ruby held tests.

bfontaine:
Yeah, it doesn’t support any `-h`/`--help` option nor anything we could use without creating a database.

bfontaine:
Thank you @denji! :tada:

Issue #41740, opened 2015-07-15
QEMU: SHA256 sum mismatch on VENOM patch


mtpereira:
Hello,

I'm getting this error when installing QEMU: https://gist.github.com/anonymous/2e067a0e1c586cb8253d

I've tried curling the patchfile and I get the same "Actual:" sum from brew:

```
$ curl 'http://git.qemu.org/?p=qemu.git;a=commitdiff_plain;h=e907746266721f305d67bc0718795fedee2e824c' 2>/dev/null | sha256sum
2aaa9ff9ee492dede64df3fcb59032ae0452bf4790c2d5881e05981fa7452368  -
```

Did anyone else got this? Am I doing something wrong here?

Thank you.

bfontaine:
I can reproduce.

DomT4:
Yup, here too. We use `commitdiff_plain` because it's supposed to stop the patch breaking whenever upstream update their `git` version or similar; not sure why that hasn't prevented the SHA changing here.

dunn:
Fixed by #41754. Forgot to add that to the commit message, my bad!

PR #41754, opened 2015-07-15
Fix for issue #41740

mtpereira:
Updated SHA256 sum for VENOM patch file on qemu recipe. This fixes issue #41740.


dunn:
Will this happen again, though? Maybe we should host our own gist.

mtpereira:
You're absolutely right! How does this look? https://gist.githubusercontent.com/mtpereira/77dd144343bf24552aad/raw/0f22e76e50013f40a4c4d6dee1a6ec2a1ffea18b/qemu-venom.patch

dunn:
Looks good to me!

mtpereira:
Ok, just pushed that change.

dunn:
And merged in 62030edb65966f9097d2a9b98b60e594e4e82379. Thanks for fixing this!

PR #41758, opened 2015-07-15
mm-common 0.9.8

nijikon:


DomT4:
This is a false flag by our audit. `mm-common` doesn't follow the er, common GNOME versioning system. We should really add a whitelist to the audit when we get the chance.

Merged in https://github.com/Homebrew/homebrew/commit/9f4729ac51a3b8c799fa8b81817fb1230b338362. Thank you @nijikon!

PR #41765, opened 2015-07-16
ffmpegthumbnailer 2.0.10

L2G:
* New upstream version
* Home moved from Google Code to GitHub
* Build process changed from GNU-style configure et al. to CMake, so add new patch and inreplace to massage the build for Homebrew
* Add test
* Other changes for `brew audit --strict`

### DIFF
--- a/Library/Formula/ffmpegthumbnailer.rb
+++ b/Library/Formula/ffmpegthumbnailer.rb
@@ -1,10 +1,9 @@
-require 'formula'
-
 class Ffmpegthumbnailer < Formula
   desc "Create thumbnails for your video files"
-  homepage 'http://code.google.com/p/ffmpegthumbnailer/'
-  url 'https://ffmpegthumbnailer.googlecode.com/files/ffmpegthumbnailer-2.0.8.tar.gz'
-  sha1 '2c54ca16efd953f46547e22799cfc40bd9c24533'
+  homepage "https://github.com/dirkvdb/ffmpegthumbnailer"
+  url "https://github.com/dirkvdb/ffmpegthumbnailer/archive/2.0.10.tar.gz"
+  sha256 "68125d98d72347a676ab2f9bc93ddd3537ff39d6a81145e2a58a6de5d3958e4e"
+
   bottle do
     cellar :any
     sha1 "b33cd322e1dd892c3ff492647a9d7fc4b8766388" => :mavericks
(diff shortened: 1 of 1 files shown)


L2G:
I need to fix the formula syntax to work with Ruby 1.9 so it will work in Mountain Lion.

Library/Formula/ffmpegthumbnailer.rb
```diff
@@ -12,16 +11,52 @@ class Ffmpegthumbnailer < Formula
...
   def install
-    system "./configure", "--disable-debug", "--disable-dependency-tracking",
-                          "--prefix=#{prefix}"
-    system "make install"
+    # The CMake script does not allow fine-grained control of man directories or
+    # other installation paths beneath the prefix
+    inreplace "CMakeLists.txt", "${CMAKE_INSTALL_PREFIX}/man/man1", man1
+
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args, "-DENABLE_GIO=ON",
```
dunn: Older versions of Ruby can't have anything in a list like this after an `*exploded` array. Just swapping `*std_cmake_args` with `"-DENABLE_GIO=ON"` should fix the ML build.

dunn: Woops, looks like you already know this!

Library/Formula/ffmpegthumbnailer.rb
```diff
@@ -12,16 +11,52 @@ class Ffmpegthumbnailer < Formula
...
-  depends_on 'ffmpeg'
+  patch :DATA
 
   def install
-    system "./configure", "--disable-debug", "--disable-dependency-tracking",
-                          "--prefix=#{prefix}"
-    system "make install"
+    # The CMake script does not allow fine-grained control of man directories or
+    # other installation paths beneath the prefix
+    inreplace "CMakeLists.txt", "${CMAKE_INSTALL_PREFIX}/man/man1", man1
```
dunn: Has this been reported upstream? They can fix their build script, right?

Library/Formula/ffmpegthumbnailer.rb
```diff
@@ -12,16 +11,52 @@ class Ffmpegthumbnailer < Formula
...
+  depends_on "pkg-config" => :build
+  depends_on "jpeg"
+  depends_on "libpng"
+  depends_on "ffmpeg"
 
-  depends_on 'pkg-config' => :build
-  depends_on 'jpeg'
-  depends_on 'libpng'
-  depends_on 'ffmpeg'
+  patch :DATA
```
dunn: Can this be accomplished with `-DMACOSX_RPATH=OFF` as an argument to cmake?

Library/Formula/ffmpegthumbnailer.rb
L2G: I have a feeling they used some kind of template or generator to create their CMakeLists.txt, then never looked at it again except to bump the version number. So yeah, there is a lot of room for improvement in there, this just being the tip of the iceberg (or the mouth of the rabbit hole).

L2G:
After more careful inspection of the GitHub repo, the bzip2 tarball, and the gzip tarball, I believe the bzip2 tarball was produced by an experimental run of CPack. I don't think the CMake build is (or was intended to be) ready for prime time. I'm going to go back and try this again using the gzip tarball with Automake and Autoconf.

dunn:
Merged in 325ab8f68bec4a5c0de1be48150e99748446fbda. Thanks for figuring that out!

PR #41769, opened 2015-07-16
giti 1.0 (new formula)

kevcooper:
giti is a script that speeds up the process of initializing new git repositories

### DIFF
--- a/Library/Formula/giti.rb
+++ b/Library/Formula/giti.rb
@@ -1,0 +1,15 @@
+class Giti < Formula
+  desc "Super awesome git initialization script!"
+  homepage "https://github.com/kevcooper/giti"
+  url "https://github.com/kevcooper/giti/archive/v1.0.tar.gz"
+  head "https://github.com/kevcooper/giti.git"
+  sha256 "24216a3ce4df46f8e37455a474afd4dc3bd2cf1e78d1c942a63119044c6c5e14"
+
+  def install
+    bin.install "giti"
+  end
+
+  test do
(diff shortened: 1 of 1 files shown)


Library/Formula/giti.rb
```diff
@@ -0,0 +1,15 @@
...
+  url "https://github.com/kevcooper/giti/archive/v1.0.tar.gz"
+  head "https://github.com/kevcooper/giti.git"
+  sha256 "24216a3ce4df46f8e37455a474afd4dc3bd2cf1e78d1c942a63119044c6c5e14"
+
+  def install
+    bin.install "giti"
+  end
+
+  test do
+    system "#{bin}/giti", "--help"
```
dunn: Any chance of a test that checks more of the functionality? Git is available on every system so you can make a test repository.

dunn:
We can't include every piece of software, so very new projects and self-submissions are rarely accepted, unfortunately.

kevcooper:
Yeah I'll add some real tests :) and sorry about adding to the clutter if this is too basic for a PR

DomT4:
Yeah, It's failing two of our standard audit checks on this subject:

```
```

For now, you may wish to maintain it in your own tap, which is very easy to set up and would allow users to use Homebrew to install it still, just via `brew install kevcooper/giti/giti` instead of `brew install giti`.

Closing here per the above, but thanks for the submission and when the tool gains a little age & popularity you're welcome to resubmit the PR. Thanks!

PR #41770, opened 2015-07-16
GNU Various, Pt2

DomT4:
Part 2 of #41768.


DomT4:
Very strange on the `gnu-apl` situation here. It compiles and bottles locally inside 5 minutes, and then you can install that bottle without issue. On the test-bot, the bottle installation hangs for >50 minutes and breaks the build. Going to move it to its own PR shortly.

PR #41775, opened 2015-07-16
gobject-introspection add dependancies on bison and flex

andrewjpage:
In a clean install on Ubuntu Trusty, it says its missing bison and flex. Adding them in fixes the issue.
Andrew


bfontaine:
Is this relevant for OS X as well?

andrewjpage:
I just checked it on Mountain Lion, and nope these arent needed so I've wrapped it in an if OS.linux? block.

tdsmith:
Linux only stuff should be sent to Homebrew/linuxbrew. Keep the `if` in when you send it there. Thanks!

Issue #41633, opened 2015-07-12
gcc 5.1.0 --with-java fails to build on 10.11


aw1621107:
[gist](https://gist.github.com/anonymous/80cdf9a02894ddbd08ea)

Clean install of 10.11, XCode 7.0. Compile seems to make it past the stage 2/3 comparison, fails somewhere when dealing with gfortran? Compiles and installs fine when passed --without-fortran, ~~even with --with-java/git/nls. --with-all-languages seems to override --without-fortran, though?~~

There seems to be some errors when compiling nls stuff, around line ~37000 in 02.make, but it doesn't seem to stop the build process.

If any more info is needed, let me know. My apologies for not being able to provide more useful info; most of this is gibberish to me.

tdsmith:
I can successfully build gcc on this configuration on 10.11 PB1:
```
HOMEBREW_VERSION: 0.9.5
ORIGIN: https://github.com/Homebrew/homebrew
HEAD: 3947f85477e93e0bb0a6cc880b948204102c195c
Last commit: 7 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: dual-core 64-bit haswell
OS X: 10.11-x86_64
Xcode: N/A
CLT: 7.0.0.0.1.1436025195
Clang: 7.0 build 700
X11: N/A
System Ruby: 2.0.0-p645
Perl: /usr/bin/perl
Python: /usr/bin/python
Ruby: /usr/bin/ruby
Java: N/A
```

tdsmith:
I think this is the error that killed this build:
```
depbase=`echo java/lang/reflect/natVMProxy.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
        /bin/sh ./libtool --tag=CXX   --mode=compile /private/tmp/gcc20150712-81848-xw8vaz/gcc-5.1.0/build/./gcc/xgcc -shared
-libgcc -B/private/tmp/gcc20150712-81848-xw8vaz/gcc-5.1.0/build/./gcc -nostdinc++ -L/private/tmp/gcc20150712-81848-xw8vaz/gcc
-5.1.0/build/x86_64-apple-darwin15.0.0/i386/libstdc++-v3/src -L/private/tmp/gcc20150712-81848-xw8vaz/gcc-5.1.0/build/x86_64-a
[...]

PR #41779, opened 2015-07-16
gcc 5.2.0

denji:
GCC 5.2 official released. Fixes more than 81 bugs/regressions in the GCC5 series.

###### Fixed (patch upstream):
- Fixes build --with-java with Xcode 7: #41633, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66509
- Fixes build with Xcode 7: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66523

###### Not fixed, continue to use the patches:
- Fix for `libgccjit.so` linkage on Darwin: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64089

### DIFF
--- a/Library/Formula/gcc.rb
+++ b/Library/Formula/gcc.rb
@@ -21,32 +21,15 @@ class Gcc < Formula
 
   desc "GNU compiler collection"
   homepage "https://gcc.gnu.org"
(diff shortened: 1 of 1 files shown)


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/4e266601b2858b71a77e50cc7ba914e5e1c696a6. Thank you @denji!

PR #41786, opened 2015-07-16
collada-dom 2.4.3

mikepurvis:
This change also switches the source from the Sourceforge-hosted releases which include binaries to just using the tagged source bundle from Github. It's a smaller download and gets away from Sourceforge's hosting issues.

@rdiankov, can you confirm that your repo is the upstream for collada-dom?

### DIFF
--- a/Library/Formula/collada-dom.rb
+++ b/Library/Formula/collada-dom.rb
@@ -2,34 +2,40 @@ require 'formula'
 
 class ColladaDom < Formula
   desc "C++ library for loading and saving COLLADA data"
-  homepage 'http://www.collada.org/mediawiki/index.php/Portal:COLLADA_DOM'
-  url 'https://downloads.sourceforge.net/project/collada-dom/Collada%20DOM/Collada%20DOM%202.4/collada-dom-2.4.0.tgz'
-  sha1 '74e28d670497abc897c06a41df7d28eea2bac836'
+  homepage 'https://www.khronos.org/collada/wiki/Portal:COLLADA_DOM'
+  url 'https://github.com/rdiankov/collada-dom/archive/v2.4.3.tar.gz'
+  sha256 'ba4d273ff68b882b89a536dc9e6d057b37c651cb476a6cfe85d119b7e60b6cce'
 
   depends_on 'cmake' => :build
   depends_on 'pcre'
   depends_on 'boost'
 
-  # Fix build of minizip: quoting arguments to cmake's add_definitions doesn't work the way they thought it did.
-  patch :DATA
-
   def install
     system "cmake", ".", *std_cmake_args
     system "make install"
   end
+
+  test do
+    (testpath/'test_simple.cpp').write <<-EOS.undent
+      #include "dae.h"
+
+      int main() {
+        DAE dae;
+        dae.add("simple.dae");
+        dae.writeAll();
+        return 0;
+      }
+    EOS
+
+    system ENV.cc, "test_simple.cpp", "-I#{include}/collada-dom2.4",
+                   "-lstdc++", "-lcollada-dom2.4-dp",
+                   "-L#{Formula["boost"].opt_lib}", "-lboost_system"
+    system "./a.out"
+
+    expected = <<-EOS.undent
+      <?xml version="1.0" encoding="UTF-8"?>
+      <COLLADA xmlns="http://www.collada.org/2008/03/COLLADASchema" version="1.5.0"/>
+    EOS
+    assert_equal expected, File.open("simple.dae", "rb").read
+  end
 end
-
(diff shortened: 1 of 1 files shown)


mikepurvis:
Ack, sorry, bad committer name, will fix.

dunn:
Any chance of a test as well? Thanks!

mikepurvis:
Done.

Library/Formula/collada-dom.rb
```diff
@@ -3,33 +3,45 @@
 class ColladaDom < Formula
   desc "C++ library for loading and saving COLLADA data"
   homepage 'http://www.collada.org/mediawiki/index.php/Portal:COLLADA_DOM'
```
vszakats: Suggesting to follow the redirect and move to 'https://':
https://www.khronos.org/collada/wiki/Portal:COLLADA_DOM

Library/Formula/collada-dom.rb
```diff
@@ -3,33 +3,51 @@
...
-end
 
-__END__
-diff --git a/CMakeLists.txt b/CMakeLists.txt
-index 72b6deb..0c7f7ce 100644
---- a/CMakeLists.txt
-+++ b/CMakeLists.txt
-@@ -100,7 +100,7 @@ endif()
+  def test
+    # Test per: https://www.khronos.org/collada/wiki/DOM_guide:_Creating_documents#Creating_a_simple_document
```
dunn: We can probably get by without the comment (unless attribution is absolutely required!)

mikepurvis: Sure, I can get rid of it— just seemed reasonable to indicate to a future maintainer where the test came from (for example, if it breaks and they want to consult the same doc for the API change).

mikepurvis:
Homepage URL updated, test comment removed.

Library/Formula/collada-dom.rb
```diff
@@ -2,34 +2,50 @@
 
 class ColladaDom < Formula
   desc "C++ library for loading and saving COLLADA data"
-  homepage 'http://www.collada.org/mediawiki/index.php/Portal:COLLADA_DOM'
-  url 'https://downloads.sourceforge.net/project/collada-dom/Collada%20DOM/Collada%20DOM%202.4/collada-dom-2.4.0.tgz'
-  sha1 '74e28d670497abc897c06a41df7d28eea2bac836'
+  homepage 'https://www.khronos.org/collada/wiki/Portal:COLLADA_DOM'
+  url 'https://github.com/rdiankov/collada-dom/archive/v2.4.3.tar.gz'
+  sha1 '0eb3b9f3183de8e5183f509311aaa7e8bc2679b7'
```
DomT4: This should be `sha256`.

tdsmith:
Can we avoid using cmake in the test? Otherwise cmake will have to become a runtime dependency, not a :build dependency.

mikepurvis:
Is there a way to specify a test-only dependency? I'd prefer to test with CMake, as that verifies the find module, which is most of what I care about.

Alternatively, is there a good example of a pkgconfig-based test?

tdsmith:
There's no concept of test dependencies. Using pkg-config isn't better since we'd have to make _that_ a dependency. Our goal for the `test` block is to have a minimal smoke test that makes sure the bottles are linked correctly. If we can just build this with `clang -I#{include} test.c -L#{lib} -lfoo -o test` that would be best.

mikepurvis:
How would we manage the boost dependency in that scenario?

tdsmith:
By hand; adding `Formula["boost"].opt_include` and `Formula["boost"].opt_lib` to the respective search paths should work.

mikepurvis:
Hash updated, test updated to remove cmake dependency.

mikepurvis:
Test block updated, getting a green light from brew-bot now. Did you want this squashed before merging?

dunn:
Yes, please!

mikepurvis:
Aight, rebase complete.

dunn:
> @rdiankov, can you confirm that your repo is the upstream for collada-dom?

Has this been confirmed?

mikepurvis:
Will ping by email. In any case, it's fairly clear that the old svn repo at SF is now dead, despite being referenced on this project's wiki pages:

https://www.khronos.org/collada/wiki/DOM_guide:_Setting_up#Downloading_the_COLLADA_DOM

DomT4:
This needs rebasing against an up-to-date master branch. Any response from upstream yet on the upstream repo change?

DomT4:
Closing here for now. If we can get upstream confirmation of the move we'll review a fresh PR on these changes. Thank you for the submission nonetheless :heart:

PR #41787, opened 2015-07-16
The Haskell Tool Stack

bjpbakker:
Rationale: provide a binary install for stack so that a system wide ghc
s not necessary to install stack and have stack install ghc.

### DIFF
--- a/Library/Formula/stack.rb
+++ b/Library/Formula/stack.rb
@@ -1,0 +1,16 @@
+class Stack < Formula
+  desc "The Haskell Tool Stack"
+  homepage "https://www.stackage.org"
+  url "https://github.com/commercialhaskell/stack/releases/download/v0.1.2.0/stack-0.1.2.0-x86_64-osx.gz"
+  sha256 "6e1039d9c5144fb03dbfb1f569a830724593191305998e9be87579d985feb36c"
+
+  def install
+    bin.install "stack-#{version}-x86_64-osx" => "stack"
+  end
+
+  test do
+    system "#{bin}/stack", "new"
+    system "#{bin}/stack", "setup"
+    system "#{bin}/stack", "build"
+  end
+end


dunn:
I'm afraid if it's just a binary it'll have to go in [homebrew-binary](https://github.com/Homebrew/homebrew-binary).

DomT4:
Looks like it is possible to build from source. That could be worth a crack.

bjpbakker:
Building from source kind of defeats the purpose of installing stack through home-brew. The soul purpose to do that is to install a binary and no system-wide ghc.

If necessary I'll move for homebrew-binary.

mistydemeo:
> Building from source kind of defeats the purpose of installing stack through homebrew. The soul purpose to do that is to install a binary and no system-wide ghc.

We build and distribute our own binaries. If this builds from source we will add our own binaries that were built from it.

mistydemeo:
Note that this has previously been submitted - see #40534, #41626, and Homebrew/homebrew-binary#242.

bjpbakker:
Building from source is currently not an option as stack needs ghc-7.8 to compile and Homebrew is on 7.10. AFAICT there's no way to depend on specific formula versions, but I sincerely hope to be wrong.

Otherwise I'll move towards homebrew-binary.

dunn:
One way to specify older versions is just to include them as a resource, like [how HamsterDB uses an old libuv](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/hamsterdb.rb#L28-L31). But it would be very sub-optimal to vendor GHC.

bjpbakker:
Maybe when ghc-7.10.2 gets released `stack` can be included in homebrew and built from source. Until then I'll try to get this formula accepted in homebrew-binary.

Thanks for your time.

mistydemeo:
GHC 7.10.2 is out now.

Issue #41388, opened 2015-07-05
Have formulae install lisp files in site-lisp/#{name}


dunn:
Right now most formulae that install emacs lisp files just dump them in `share/"emacs/site-lisp"` which is symlinked to `HOMEBREW_PREFIX/"share/emacs/site-lisp"`.  Emacs package authors are pretty good about prefixing their `.el` files, but this might still cause conflicts, especially if formulae contain vendored lisp libraries (it's surprising that magit is the only formula in core that vendors `dash.el` right now).

What I'd like to do is require that formulae install their lisp files in a formula-specific directory.  This is what it would look like: https://github.com/Homebrew/homebrew/compare/Homebrew:master...dunn:site-lisp

If this is a good idea, but too large for a single PR, let me know and I'll split it up.

dunn:
Come to think of it, vendored packages should be in `lib/vendor` instead of `site-lisp`.  But I still think this is a good idea.

mikemcquaid:
Makes sense to me :+1:.

dunn:
Should I do a PR for the caveats and cellar checks first?

mikemcquaid:
Up to you!

PR #41791, opened 2015-07-16
Lisp checks

dunn:
These are the only changes that need to be made to core for #41388. If this looks good, then once it's merged I'll start updating formulae.

### DIFF
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -16,6 +16,7 @@ class Caveats
     caveats << plist_caveats
     caveats << python_caveats
     caveats << app_caveats
+    caveats << elisp_caveats
     caveats.compact.join("\n")
   end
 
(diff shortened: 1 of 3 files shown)


Library/Homebrew/caveats.rb
```diff
@@ -132,6 +133,15 @@ def app_caveats
     end
   end
 
+  def elisp_caveats
+    if keg && keg.elisp_installed?
+      <<-EOS.undent
+        Lisp files have been installed to:
```
xu-cheng: Maybe `Emacs Lisp` rather than simple `Lisp` here.

Library/Homebrew/formula_cellar_checks.rb
```diff
@@ -173,6 +173,26 @@ def check_python_framework_links lib
     EOS
   end
 
+  def check_site_lisp(share, name)
```
xu-cheng: Method would be better to name as `check_emacs_lisp` or something.

Library/Homebrew/caveats.rb
dunn: Yeah, good point.

dunn:
I'm starting to worry it's too strict a requirement. I'll think about it a bit more.

PR #41792, opened 2015-07-16
Add ‘desc’-related functionality

hivehand:
This does three principal things:

1. Introduces a `Descriptions` class with which to retrieve and pretty-present formula descriptions.
2. Adds a `--desc` option to `brew update` which makes it pretty-present name-and-description, one per line, for new and updated formulae.
3. Resurrects the original `brew desc` command (albeit, at present, without any search facility and with slightly modified result formatting). It also adds bash completion for said command.

(It also, purely incidentally, modifies `utils.rb` to determine the longest string in an array by mapping `length` onto the array, and then taking the maximum of the result, instead of sorting the lengths and taking the largest. According my test this halves the duration of the maximum-length computation.)

**Rationale for the desc changes**:
*(with a tip of the hat to [xkcd](https://xkcd.com)’s “[Workflow](https://xkcd.com/1172/)”)*

I am always looking for new and undiscovered tools, but reading through the descriptions of all 3000+ formulae at once would be overwhelming. Since I usually run `brew update` once a day, looking into the day’s updates is manageable. Back when `homebrew-desc` existed as a separate tool, I’d just select the table of updated formulae in iTerm 2, and then run `brew desc $(echo $(pbpaste) | rs -T -C" ")` to get a list of descriptions.

It would be even better (for me, anyway) if I had the option of telling `brew update` to pretty-print the descriptions for all of the new and updated formulae to begin with.

The separate `desc` command is still useful, though, for things like processing the results of a `brew update` run without the `--desc` flag, or for summarizing your own installed formula set: `brew desc $(brew list)`

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -204,8 +204,8 @@ def puts_columns items, star_items=[]
     # determine the best width to display for different console sizes
     console_width = `/bin/stty size`.chomp.split(" ").last.to_i
     console_width = 80 if console_width <= 0
-    longest = items.sort_by { |item| item.length }.last
-    optimal_col_width = (console_width.to_f / (longest.length + 2).to_f).floor
+    max_len = items.map(&:length).max
+    optimal_col_width = (console_width.to_f / (max_len + 2).to_f).floor
     cols = optimal_col_width > 1 ? optimal_col_width : 1
 
     IO.popen("/usr/bin/pr -#{cols} -t -w#{console_width}", "w"){|io| io.puts(items) }
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -204,7 +204,7 @@ def puts_columns items, star_items=[]
     # determine the best width to display for different console sizes
     console_width = `/bin/stty size`.chomp.split(" ").last.to_i
     console_width = 80 if console_width <= 0
-    max_len = items.map(&:length).max
+    max_len = items.reduce(0) { |max, item| l = item.length ; l > max ? l : max }
     optimal_col_width = (console_width.to_f / (max_len + 2).to_f).floor
     cols = optimal_col_width > 1 ? optimal_col_width : 1
 
--- a/Library/Contributions/brew_bash_completion.sh
+++ b/Library/Contributions/brew_bash_completion.sh
@@ -189,6 +189,18 @@ _brew_deps ()
     __brew_complete_formulae
 }
 
+_brew_desc ()
+{
+    local cur="${COMP_WORDS[COMP_CWORD]}"
+    case "$cur" in
+    --*)
+        __brewcomp "--search"
+        return
+        ;;
+    esac
+    __brew_complete_formulae
+}
+
 _brew_doctor () {
     local cur="${COMP_WORDS[COMP_CWORD]}"
     __brewcomp "$(brew doctor --list-checks)"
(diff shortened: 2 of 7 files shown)


hivehand:
It occurred to me shortly after opening this request that I should have updated the man page. I’m doing that now, and will push the update as soon as it’s ready.

Library/Homebrew/utils.rb
```diff
@@ -204,8 +204,8 @@ def puts_columns items, star_items=[]
     # determine the best width to display for different console sizes
     console_width = `/bin/stty size`.chomp.split(" ").last.to_i
     console_width = 80 if console_width <= 0
-    longest = items.sort_by { |item| item.length }.last
-    optimal_col_width = (console_width.to_f / (longest.length + 2).to_f).floor
+    max_len = items.map(&:length).max
```
bfontaine: Couldn’t we do it in one pass? Your implementation is still better than the original code but something like this would probably be faster / more memory-efficient:

```ruby
max_len = items.reduce(0) { |max, item| l = item.length ; l > max ? l : max }
```

DomT4:
CC @telemachus; may be interested in this as well.

bfontaine:
I’m :-1: on this, except on the idea of an import of `brew desc` in the core.
Regarding the implementation, I don’t think we need yet another file with yet another class just for two functions with no internal state. Why not writing something like `puts_columns` but called `puts_formulae_with_desc` or similar?
Regarding `--desc`, I think it’s a niche feature and we’ve closed PRs in the past that added features like this to `brew update` asking users to implement their own custom command. You can use websites like [brewformulas](http://brewformulas.org/) to discover new formulae with their description.

(note that I do *not* want to seem harsh here)

hivehand:
Points:

1. You’re absolutely right about using `reduce` instead of `map` to determine the longest string: it’s faster still. Updated.
2. Just for the record, my thinking behind the (admittedly currently-lightweight) `Description` class was that it would make incorporating some kind of [description-caching mechanism](https://github.com/Homebrew/homebrew/pull/39911#issuecomment-103697853) down the line easier. You’d tweak `get`, and everything that called it would continue operating as before, just faster.
3. I don’t think you’re being harsh at all. I am suggesting changes in the core of `brew`, and I’m fully aware that that’s not done lightly.
4. I can get (most of) what I want via `brew desc`. If you’d like to close this pull request, and have me open a new one that disposes of the `Descriptions` class while implementing `brew desc` without it, please let me know.

telemachus:
I'm definitely in favor of adding a proper `brew desc` command. I was sorry when that didn't happen during the merge, and to be honest, I still use my `brew desc` for that need.

mikemcquaid:
@telemachus Remind me why it didn't happen during the merge? I think at most this PR should add the `desc` command and completion (for reasons above)

bfontaine:
@hivehand I wasn’t aware of the caching issue, thanks. However, I’m concerned about the YAGNI principle here ([“You aren't gonna need it”](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)). If the point of `Descriptions` here is to ease the cache implementation *in the future*, why not waiting for this implementation to add the `Descriptions` class? I’ve seen a lot of situations, outside of Homebrew, where one (including me) implements something thinking it’d be useful in the future, then things change, the future feature is never implemented, and there’s unneeded code everywhere.

telemachus:
@mikemcquaid It wasn't rejected, if that's what you're wondering. It didn't happen then simply because it wasn't part of that PR at all. That PR was only about getting descriptions into each formula and integrating them into `brew info` and (with an added `--desc` flag) into search. At the time, nobody mentioned the distinct `brew desc` command.

I thought of it then, but didn't mention it simply because it was *already* a big PR with a lot of potential blockers. So I didn't want to make it harder. And I haven't done this myself simply because I didn't get around to it.

But a PR that brought in `desc` and completion would be a great addition to `brew`, I think. (Thanks again @hivehand for doing this.)

hivehand:
I’m going to close this, and will try to open a new one focusing on just `brew desc` in the near future. (That is, of course, unless @telemachus beats me to it. 😁)

PR #41793, opened 2015-07-16
Fix no such configure file when HEAD build

zchee:
Signed-off-by: Koichi Shiraishi <zcheeee@gmail.com>

### DIFF
--- a/Library/Formula/pcre.rb
+++ b/Library/Formula/pcre.rb
@@ -5,7 +5,13 @@ class Pcre < Formula
   mirror "https://www.mirrorservice.org/sites/downloads.sourceforge.net/p/pc/pcre/pcre/8.37/pcre-8.37.tar.bz2"
   sha256 "51679ea8006ce31379fb0860e46dd86665d864b5020fc9cd19e71260eef4789d"
 
-  head "svn://vcs.exim.org/pcre/code/trunk"
+  head do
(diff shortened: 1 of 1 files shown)


Library/Formula/pcre.rb
```diff
@@ -24,6 +24,7 @@ class Pcre < Formula
   def install
     ENV.universal_binary if build.universal?
 
+    system "./autogen.sh" if build.head?
```
DomT4: Presumably this needs the autotools, in which case you'll need to add a `head` block with the autotools as build dependencies.

zchee:
@DomT4 Yes, sir!!
Will fix!

zchee:
Fixed!

But, required `--env=std` flag.
If do not set this flag, return `./autogen.sh: line 33: aclocal: command not found` errors.

DomT4:
Sorry if I confused you here :). You want to do something like this:

```ruby
head do
  url "svn://vcs.exim.org/pcre/code/trunk"

end
```

```ruby
def install

```

You've already got the last bit done.

zchee:
@DomT4 Fixed.
Will that be OK?

DomT4:
@zchee Looks perfect, Thanks!

zchee:
Thanks advise :)

mikemcquaid:

For future reference the preferred commit message is `pcre: fix HEAD build.` and please don't sign-off your own commits.

PR #41795, opened 2015-07-16
dos2unix 7.2.3

nijikon:


DomT4:
Need to kill off the devel block:

```
Error: 1 problem in 1 formula
dos2unix:
 * devel version 7.2.3-beta1 is older than stable version 7.2.3
```

dunn:
Better yet, update the devel version as well, to 7.3-beta3.

DomT4:
^ :+1:

xu-cheng:

PR #41802, opened 2015-07-17
fontforge: add --skip-git to bootstrap options

L2G:
This adds `--skip-git` to the options given to the bootstrap script. Otherwise, it will still try to fetch the latest master branch for gnulib, foiling this formula's attempt to pin to a specific commit.

### DIFF
--- a/Library/Formula/fontforge.rb
+++ b/Library/Formula/fontforge.rb
@@ -85,8 +85,12 @@ class Fontforge < Formula
     ENV["ARCHFLAGS"] = "-arch #{MacOS.preferred_arch}"
 
     # Bootstrap in every build: https://github.com/fontforge/fontforge/issues/1806
+    # "--skip-git" tells the bootstrap script not to pull the latest master
+    # branch of gnulib and just use the commit we've checked out.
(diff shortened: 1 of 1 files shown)


Library/Formula/fontforge.rb
```diff
@@ -85,8 +85,12 @@ def install
     ENV["ARCHFLAGS"] = "-arch #{MacOS.preferred_arch}"
 
     # Bootstrap in every build: https://github.com/fontforge/fontforge/issues/1806
+    # "--skip-git" tells the bootstrap script not to pull the latest master
+    # branch of gnulib and just use the commit we've checked out.
     resource("gnulib").fetch
-    system "./bootstrap", "--gnulib-srcdir=#{resource("gnulib").cached_download}"
+    system "./bootstrap",
+      "--gnulib-srcdir=#{resource("gnulib").cached_download}",
+      "--skip-git"
```
DomT4: I think you're fine to wedge this onto the same line personally, but if not let's line up the arguments with the start of `./bootstrap` if possible.

tdsmith:
Thanks; good catch!

PR #41806, opened 2015-07-17
Making a new alias, golang -> go

sparrc:
The Go community generally searches for things related to the language by
'golang' rather than 'go', since searching for 'go' returns so many
unrelated results. (see http://golang.org/)

### DIFF
--- a/Library/Aliases/golang
+++ b/Library/Aliases/golang
@@ -1,0 +1,1 @@
+../Formula/go.rb


DomT4:
Huh. Colour me surprised we didn't already have this alias. I'm cool with it at least.

bfontaine:
:+1:

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/f77a33395b8680778ef0c06b3be670a4249ff629. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

It's a nit, but generally commit messages for something like this should be styled `go: alias to golang` or similar, Thanks!

PR #41808, opened 2015-07-17
allow file:/// urls to "download" software from local directory

smoofra:
This would be silly to use in real life, but it's a convenient way to test out
changes to the source of a package for debugging purposes.

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -828,6 +828,19 @@ class FossilDownloadStrategy < VCSDownloadStrategy
 
 end
 
+class LocalDirectoryDownloadStrategy < AbstractDownloadStrategy
+
+  def initialize(name, resource)
+    super
+    @url = @url.sub(%r[^file://], "")
+  end
+
+  def stage
+    ohai "cp -r #{@url}/* ."
+    cp_r(Dir[@url + "/*"], ".")
+  end
+end
+
 class DownloadStrategyDetector
   def self.detect(url, strategy=nil)
     if strategy.nil?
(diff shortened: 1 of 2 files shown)


mistydemeo:
The default CurlDownloadStrategy can fetch from `file://` URLs fine; was there a specific problem you were having?

smoofra:
yea, that wasn't working for me, because curl isn't going to recursively copy a directory from a file:// url

smoofra:
oh, but curl *will* work for a tarball.....

smoofra:
I changed it so it only applies the new download strategy in the case where curl wouldn't work (when the path is a directory)

DomT4:
FWIW, It will work for a directory in certain circumstances. FontForge use a [special head](https://github.com/fontforge/fontforge/blob/master/travis-scripts/fontforge.rb) for their Travis testing.

smoofra:
@DomT4  yea it looks like GitDownloadStrategy can handle file:// urls , but you have to explicitly do ":using => GitDownloadStrategy"

mikemcquaid:
> This would be silly to use in real life, but it's a convenient way to test out
> changes to the source of a package for debugging purposes.

You can embed it in a Formula yourself. I feel pretty strongly that this isn't something we should be maintaining in Homebrew core.

dunn:
If `:using => GitDownloadStrategy` works for your needs, that seems like the better solution.

smoofra:
@dunn I view the problem as more about not surprising the user, rather than finding a way to solve the particular problem I was dealing with at the time.   

@mikemcquaid I can understand that.   It's hard to justify the maintenance burden of a feature that's only useful for local experiments.   My argument for putting it in the core is only this:    I was surprised when a file:/// url to a directory didn't work.   If I expected it to work, then probably other people did too, so including it in the core would save them some frustration.

dunn:
In that case, maybe `LocalDirectoryDownloadStrategy` could just be an alias to `GitDownloadStrategy`.

smoofra:
but... what if the local directory isn't a git repository??

dunn:
Oh, I misunderstood. I thought it worked for `file:///` in general.

bfontaine:
A `file:///` URL, like `http://` URLs, points to a *resource*, which is usually a file. I don’t find surprising that it doesn’t work for a directory.

mikemcquaid:
Unless other maintainers object I think we should close this, sorry!

DomT4:
Don't have any issues with closing. The in-formula workarounds already possible aren't that unreasonable, and officially supporting local file links in the core could open a world of fun troubleshooting.

smoofra:
😐

PR #41815, opened 2015-07-17
docker-compose: squelch warnings

tdsmith:
Keeps docker-compose from printing an InsecurePlatformWarning.

There are real issues that this ignores, like these:
https://github.com/shazow/urllib3/issues/602#issuecomment-97054951

I don't *think* those are relevant to the typical use of docker-compose, but I welcome comment on that aspect.

The alternative would be to build and install pyOpenSSL and pyca/cryptography, but that doesn't work if we build cryptography against Homebrew's openssl because something in the docker stack is incompatible with openssl 1.0.2. We won't be able to get away with building pyca/cryptography against system openssl once OS X 10.11 is released, since Apple's yanking the headers, so let's not do it now.

I think the warning is correct and benign but it's causing issue reports to us and to the docker-compose developers.

### DIFF
--- a/Library/Formula/fig.rb
+++ b/Library/Formula/fig.rb
@@ -76,8 +76,19 @@ class Fig < Formula
 
     bash_completion.install "contrib/completion/bash/docker-compose"
 
+    # allow PYTHONWARNINGS to squelch irrelevant InsecurePlatformWarning
+    # https://github.com/Homebrew/homebrew/issues/41766
+    # https://github.com/shazow/urllib3/issues/602
+    inreplace libexec/"vendor/lib/python2.7/site-packages/requests/packages/urllib3/__init__.py" do |s|
+      s.gsub! "warnings.simplefilter('always', exceptions.SecurityWarning)",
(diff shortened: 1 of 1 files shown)


cclauss:
Why is `docker-compose` still called by it's old name, `fig` in the formula instead of its current name, `docker-compose`?

Should `class Fig` be changed to `class docker-compose`?

tdsmith:
We can't rename formulas, but `docker-compose` is aliased to `fig` so `brew install docker-compose` works.

cclauss:
Yes but `brew search docker`shows that `docker compose` is not installed when it is.  Would it be better to have `fig` aliased to `docker-compose` instead of `docker-compose` is aliased to `fig`?

dunn:
It would be better, but that's technically renaming the formula. Keep your fingers crossed for #41006, though!

JerryVerhoef:
+1 LGTM

tdsmith:
I'm rejecting this fix because I believe the warning is correct; we can fix this properly by using a current OpenSSL once boot2docker generates correct certificates.

PR #41816, opened 2015-07-17
freeswitch 1.4.20

L2G:
* New upstream version
* Add ldns dependency (and re-enable build of mod_enum)
* Add --with-* and --without-* options to customize installed sounds
* Change bootstrap script's "-j<n>" option to a simple "-j" (otherwise it will ignore the option and not put jobs in the backround at all)

### DIFF
--- a/Library/Formula/freeswitch.rb
+++ b/Library/Formula/freeswitch.rb
@@ -2,8 +2,8 @@ class Freeswitch < Formula
   desc "Telephony platform to route various communication protocols"
   homepage "https://freeswitch.org"
   url "https://freeswitch.org/stash/scm/fs/freeswitch.git",
-      :tag => "v1.4.19",
-      :revision => "ebf2df68fa6f341311620c6d90f5a8d77334606c"
+      :tag => "v1.4.20",
+      :revision => "0ae8ee7f8f13a37cf48381381b2f30906e750e19"
 
   head "https://freeswitch.org/stash/scm/fs/freeswitch.git"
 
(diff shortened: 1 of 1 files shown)


Library/Formula/freeswitch.rb
```diff
@@ -19,20 +19,24 @@ class Freeswitch < Formula
...
+  depends_on "jpeg"
+  depends_on "ldns"
   depends_on "openssl"
   depends_on "pcre"
   depends_on "speex"
   depends_on "sqlite"
 
   def install
-    system "./bootstrap.sh", "-j#{ENV.make_jobs}"
+    system "./bootstrap.sh", "-j"
```
DomT4: Is the `-j` doing anything without `#{ENV.make_jobs}`?

L2G: It does shave about a minute off of the build time on my MacBook Pro.

DomT4: `#{ENV.make_jobs}` should do that as well, I'd presume?

L2G: When I ran `./bootstrap.sh -j8` manually, I got an error message saying that the option wasn't recognized. When I looked at the script, it didn't look like it was expecting a number after `-j` at all. Running `./bootstrap.sh -j` worked and gave a measurable improvement over plain `./bootstrap.sh`.

dunn: Yeah, looks like that's true: https://freeswitch.org/stash/projects/FS/repos/freeswitch/browse/bootstrap.sh#22

dunn:
Were you planning on updating this PR to deal with #41833 at the same time, or should we merge this now?

L2G:
@dunn This can be merged now. If I thought I'd have the other one ready within a day, I'd say wait for it, but I'm not feeling that.

dunn:
Merged in 861f5a4eaae235432c86243b535afc600bcfccce, thanks so much!

PR #37403, opened 2015-03-05
C 0.09 (new ​​formula)

janten:
C is a shell script that compiles and executes C source code in the shell.

PR #41819, opened 2015-07-17
added new formula for brew-file

rcmdnk:
New formula for brew-file (Brewfile manager).
[rcmdnk/homebrew-file](https://github.com/rcmdnk/homebrew-file)

### DIFF
--- a/Library/Formula/brew-file.rb
+++ b/Library/Formula/brew-file.rb
@@ -1,25 +1,17 @@
 class BrewFile < Formula
   desc "Brewfile manager for Homebrew."
   homepage "https://github.com/rcmdnk/homebrew-file/"
-  url "https://github.com/rcmdnk/homebrew-file.git",
-    :tag => "v3.5.9",
-    :revision => "575574b9e6d270dc2f3955df0add6c03ed18b663"
-  head "https://github.com/rcmdnk/homebrew-file.git", :branch => "master"
-  if build.with? "bash"
-    url "https://github.com/rcmdnk/homebrew-file.git", :branch => "bash"
-    version "1.1.8"
-  end
+  url "https://github.com/rcmdnk/homebrew-file/archive/v3.5.9.zip"
+  sha256 "65e05b90359abb997396e7e2ab060cc577aa12da4e5dfe5e79e2b6687caf1d1d"
 
-  option "with-python", "Use python version (same as default)"
-  option "with-bash", "Use bash version"
+  head "https://github.com/rcmdnk/homebrew-file.git"
+
   option "without-completions", "Disable bash/zsh completions"
-
-  skip_clean "bin"
 
   def install
     bin.install "bin/brew-file"
     (bin+"brew-file").chmod 0755
-    (prefix+"etc").install "etc/brew-wrap"
+    etc.install "etc/brew-wrap"
     if build.with? "completions"
       bash_completion.install "etc/bash_completion.d/brew-file"
       zsh_completion.install "share/zsh/site-functions/_brew-file"
--- a/Library/Formula/brew-file.rb
+++ b/Library/Formula/brew-file.rb
@@ -1,8 +1,8 @@
 class BrewFile < Formula
   desc "Brewfile manager for Homebrew."
   homepage "https://github.com/rcmdnk/homebrew-file/"
-  url "https://github.com/rcmdnk/homebrew-file/archive/v3.5.9.zip"
-  sha256 "65e05b90359abb997396e7e2ab060cc577aa12da4e5dfe5e79e2b6687caf1d1d"
+  url "https://github.com/rcmdnk/homebrew-file/archive/v3.5.9.tar.gz"
+  sha256 "2c7028a3798fcbf5569c5e318edb34bb48f79ff3765f10beb54dcaf385863bf2"
 
   head "https://github.com/rcmdnk/homebrew-file.git"
 
--- a/Library/Formula/brew-file.rb
+++ b/Library/Formula/brew-file.rb
@@ -1,0 +1,32 @@
+class BrewFile < Formula
+  desc "Brewfile manager for Homebrew."
+  homepage "https://github.com/rcmdnk/homebrew-file/"
+  url "https://github.com/rcmdnk/homebrew-file.git",
+    :tag => "v3.5.9",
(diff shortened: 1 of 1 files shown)


Library/Formula/brew-file.rb
```diff
@@ -0,0 +1,32 @@
+class BrewFile < Formula
+  desc "Brewfile manager for Homebrew."
+  homepage "https://github.com/rcmdnk/homebrew-file/"
+  url "https://github.com/rcmdnk/homebrew-file.git",
```
DomT4: Are there git submodules here? If not we should use the archive tarball & `sha256` rather than a tag & revision.

Library/Formula/brew-file.rb
```diff
@@ -0,0 +1,32 @@
...
+    :tag => "v3.5.9",
+    :revision => "575574b9e6d270dc2f3955df0add6c03ed18b663"
+  head "https://github.com/rcmdnk/homebrew-file.git", :branch => "master"
```
DomT4: `master` is redundant. We assume that'll be the default branch.

Library/Formula/brew-file.rb
```diff
@@ -0,0 +1,32 @@
...
+  if build.with? "bash"
```
DomT4: We don't really do things like this. What's different about the `bash` branch? Or rather, what gets installed from there that wouldn't be installed otherwise?

I think what you're looking to do here is create a `resource` block, but it's not clear.

Library/Formula/brew-file.rb
```diff
@@ -0,0 +1,32 @@
...
+    url "https://github.com/rcmdnk/homebrew-file.git", :branch => "bash"
+    version "1.1.8"
+  end
+
+  option "with-python", "Use python version (same as default)"
```
DomT4: Is this being invoked at all? I can't see any reference to it below.

Library/Formula/brew-file.rb
```diff
@@ -0,0 +1,32 @@
...
+  option "with-bash", "Use bash version"
+  option "without-completions", "Disable bash/zsh completions"
+
+  skip_clean "bin"
```
DomT4: ?

Library/Formula/brew-file.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def install
+    bin.install "bin/brew-file"
+    (bin+"brew-file").chmod 0755
+    (prefix+"etc").install "etc/brew-wrap"
```
DomT4: Should just be `etc.install`.

rcmdnk:
Thanks for pointings.

Updated:
- GitHub repository -> tarball
- Removed bash version (It is obsoleted)
- Fixed install area

mikemcquaid:
So, my only objection here is that it seems to be working OK as a tap currently and it would feel a bit weird to have this in core but `homebrew-bundle` not be. @xu-cheng, thoughts?

xu-cheng:
Yes, I share the similar feeling. On the one hand, it would work better as a tap. On the other hand, the formula `brew-file` is very confusing. Especially considering `homebrew-bundle` uses the same file name but complete different syntax. This is kinda similar to [a formula](https://github.com/Homebrew/homebrew/pull/37403#issuecomment-100428280) rejected some time ago which calls `c`.

rcmdnk:
It has a `brewdler` compatible mode (i.e. homebrew-bundle compatible),.
[homebrew-file/README.DETAILS.md at master · rcmdnk/homebrew-file](https://github.com/rcmdnk/homebrew-file/blob/master/README.DETAILS.md),
so that it can live with homebrew-bundle, if necessary.

The main concept is `managing Brewfile` itself, not only managing packages by Brewfile.

xu-cheng:
Aside the relation to `homebrew-bundle`, why don't you put it inside tap? If you put `brew-file` file inside `cmd` directory, you only need to run `brew tap rcmdnk/file` to install it. After that, when ever you run `brew update`, brew-file will be updated at the same time.

In fact, we recommend use this way rather than formula to manage external brew commands.

bfontaine:
It’s really easier to maintain it in a tap, because users will get the latest version every time they run `brew update`; there’s no need to send PRs to update the formula every time there’s a new version.

rcmdnk:
I've used a formula just because I din't know the tap way when I started to develop it (following such homebrew-cask, which also uses formula).
In addition, there are set up files in etc now, for which the formula is needed.

bfontaine:
You can use whichever additional directories you want in your tap.

rcmdnk:
How can we use `etc` directory in Tap?
Currently, I use it like:

     source $(brew --prefix)/etc/brew-wrap

If it is in Tap, maybe it shoud be full path like `$(brew-prefix)/Library/Taps/rcmdnk/homebrew-file/etc/bash_completion.d/brew-file`?
Or is there any shortcut?

xu-cheng:
You can create a `brew file-shell-init` command and use it like below in the bash/zsh configuration.

```bash
if brew command file-shell-init > /dev/null; then eval "$(brew file-shell-init)"; fi
```

That's what `homebrew/command-not-found` does.

bfontaine:
You could use a command that gives the path of the wrapper. In `homebrew-command-not-found` we have a [`brew command-not-found-init`](https://github.com/Homebrew/homebrew-command-not-found/blob/master/cmd/brew-command-not-found-init.rb) custom command which prints the path of the handler script. Users then need to add this in their `~/.bashrc`:

```
if brew command command-not-found-init > /dev/null; then eval "$(brew command-not-found-init)"; fi
```

rcmdnk:
Thank you, both.
I see.
Then, I will try to follow your recommendation,
and close this request for the fomula.

mikemcquaid:
@rcmdnk Thanks! I'd also love if there was any way you'd consider merging the two projects; it'd be great if there was only one solution and we had more people working on it! Let me know what you think or what you'd need to be done for that to happen, cheers!

rcmdnk:
@mikemcquaid I also think it is good if I can merge the functions to the bundle.
I would try to start to migrate `brew-file` in Ruby and so that I can send a pull request to `homebrew-bundle`.

mikemcquaid:
@rcmdnk Great, thanks! Let me know if we can help at all.

PR #41820, opened 2015-07-17
formula_installer: remove duplicated hold_locks check

xu-cheng:
It'll be checked inside `FormulaInstaller#unlock` method.


mikemcquaid:
:+1:

Issue #41778, opened 2015-07-16
****Sourceforge is having issues*****


smarek:
Can please someone verify and update signature? Final sourceforge redirect is through HTTP, and I can't believe such download to be legit.

```
> brew upgrade swig
==> Upgrading 1 outdated package, with result:
swig 3.0.6
==> Upgrading swig
==> Downloading https://downloads.sourceforge.net/project/swig/swig/swig-3.0.6/swig-3.0.6.tar.gz
==> Downloading from http://downloads.sourceforge.net/#!/project/swig/swig/swig-3.0.6/swig-3.0.6.tar.gz
######################################################################## 100.0%
Error: SHA256 mismatch
Expected: c67f63ea11956106e4cda66416d5020330dc4ce2ee45057d39a9494ce33eca05
Actual: 818a4b8bbcb50878a8b1b9f71b4274d242ab46bf860c74676e98dec1d0248821
Archive: /Library/Caches/Homebrew/swig-3.0.6.tar.gz
To retry an incomplete download, remove the file above.
```

bfontaine:
I get the checksum mismatch as well.

bfontaine:
Sourceforge is having a problem:

```
$ url -sL https://downloads.sourceforge.net/project/swig/swig/swig-3.0.6/swig-3.0.6.tar.gz
<html><head>
<title>SourceForge</title>
<!-- <script src="/js/jquery.com/jquery-1.11.0.min.js"></script> -->
<script src="//code.jquery.com/jquery-1.11.0.min.js"></script>
<script src="//sourceforge.net/js/mirrors.js"></script>
<script src="/js/sf.js"></script>
<script>
var DR_loc = DR_parse_hash_url();
if (DR_loc) {
    DR_sf_main(DR_loc);
} else {
    window.location.href = 'http://sourceforge.net/home.html';
}
</script>
</head><body>
<noscript>
We're sorry -- the Sourceforge site is currently in Disaster Recovery mode, and currently requires
the use of javascript to function.  Please check back later.
</noscript>
</body></html>
```

smarek:
I've also tried to obtain the file from alternative mirror, where the file got third, yet unknown checksum, what the hell is going on?
```
wget -c "https://www.mirrorservice.org/sites/download.sourceforge.net/pub/sourceforge/s/sw/swig/swig/swig-3.0.6/swig-3.0.6.tar.gz"
> sha256sum  swig-3.0.6.tar.gz 
[...]

PR #41824, opened 2015-07-17
update zsh

sunny4280:
I modified checksum.
When I executed "brew install zsh --disable-etcdir" to test, these error occurred.

This is error message:
$ brew install zsh --disable-etcdir
==> Downloading https://downloads.sourceforge.net/project/zsh/zsh/5.0.8/zsh-5.0.8.tar.bz2
Already downloaded: /Users/yasuharu/Library/Caches/Homebrew/zsh-5.0.8.tar.bz2
tar: Unrecognized archive format
tar: Error exit delayed from previous errors.
Error: Failure while executing: tar xf /Users/yasuharu/Library/Caches/Homebrew/zsh-5.0.8.tar.bz2

I couldn't resolve that error.
Please advise me.


DomT4:
This isn't a "real" error *per se*. The error you're seeing is because Sourceforge is having problems, see #41778, and consequently is currently returning empty tarballs which fail the checksum check.

Issue #41823, opened 2015-07-17
bind should depend on json-c


tg90nor:
$ dig google.com
    dyld: Library not loaded: /usr/local/lib/libjson-c.2.dylib
      Referenced from: /usr/local/bin/dig
      Reason: image not found
    [1]    41615 trace trap  dig google.com

dig works fine if I brew install json-c

dunn:
Feel like [opening a pull request](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md), adding the dependency (and maybe a test)? I can help if you have questions.

PR #41830, opened 2015-07-17
Add json-c dependency to bind

tg90nor:
Fixes issue #41823

Not sure how I would go about adding a test for this.

### DIFF
--- a/Library/Formula/bind.rb
+++ b/Library/Formula/bind.rb
@@ -16,6 +16,7 @@ class Bind < Formula
   head "https://source.isc.org/git/bind9.git"
 
   depends_on "openssl"
+  depends_on "json-c"
 
   def install
     ENV.libxml2
--- a/Library/Formula/bind.rb
+++ b/Library/Formula/bind.rb
@@ -139,6 +139,11 @@ class Bind < Formula
     EOS
   end
 
+  test do
+    system bin/"dig", "-v"
+    system bin/"dig", "brew.sh"
+  end
+
   plist_options :startup => true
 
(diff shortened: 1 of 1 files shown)


dunn:
It installs a number of executables so we could just run a few, `system bin/"dig", "http://brew.sh`, and so on

tg90nor:
I added a test.

Note that bind compiles and works just fine without json-c, but since the yosemite bottle was built with it, it must be present when using the bottle.

dunn:
> Note that bind compiles and works just fine without json-c, but since the yosemite bottle was built with it, it must be present when using the bottle.

Not sure what you mean.  Are you saying if you install bind from source, `dig` works without json-c being installed?

tg90nor:
Yes. It is only needed when using the binary bottle. Bind will automatically be built with json-c if it is present at compile time, but works fine without it.

tdsmith:
Does bind's configure script accept an option that tells it not to look for json-c?

dunn:
Yeah, looks like a better fix might be `--with-libjson=no`.

Are there advantages to building with json-c?

tg90nor:
> Are there advantages to building with json-c?

It enables generating statistics in json. Might want to make it optional? I don't personally need it.

dunn:
Yeah, `depends_on "json-c" => :optional` then pass `--with-libjson=no` to the configure script otherwise.  Thanks!

dunn:
Merged in 0ec46291a52c42e26a3467671e6ee02eaefd2104, thanks for fixing this!

PR #41832, opened 2015-07-17
use ohai headers at the top of search results

dunn:
Just a visual thing that's been bugging me:

Before:
z
zzz

After:
zz
zzzz


DomT4:
Heh. No objection here.

mikemcquaid:
:100:

PR #41835, opened 2015-07-17
ttytter: fix readline support

dunn:


Library/Formula/ttytter.rb
```diff
@@ -1,24 +1,37 @@
-require 'formula'
-
 class Ttytter < Formula
   desc "Twitter command-line client"
-  homepage 'http://www.floodgap.com/software/ttytter/'
-  url 'http://www.floodgap.com/software/ttytter/dist2/2.1.00.txt'
-  sha1 'a72b2c4b7da8a370ff15c7f5a1a3ecf9056590f0'
+  homepage "http://www.floodgap.com/software/ttytter/"
+  url "http://www.floodgap.com/software/ttytter/dist2/2.1.00.txt"
+  sha1 "a72b2c4b7da8a370ff15c7f5a1a3ecf9056590f0"
```
DomT4: Let's bump this to `sha256` since we're making changes.

dunn: Ack, right you are.

PR #41836, opened 2015-07-17
Add a special SourceForge service outage advisory

L2G:
You can take this or leave it. I just wanted to put the idea out there. It relates to the ongoing SourceForge service outage (#41778), and I thought it might be a good way to get the word out.

### DIFF
--- a/Library/Homebrew/exceptions.rb
+++ b/Library/Homebrew/exceptions.rb
@@ -278,13 +278,31 @@ class ChecksumMismatchError < RuntimeError
     @expected = expected
     @hash_type = expected.hash_type.to_s.upcase
 
-    super <<-EOS.undent
+    message = <<-EOS.undent
       #{@hash_type} mismatch
       Expected: #{expected}
       Actual: #{actual}
       Archive: #{fn}
       To retry an incomplete download, remove the file above.
       EOS
+
+    # The great SourceForge outage of July 2015
+    bad_hashes = %w[
+      0103b909e19ca9c6497a7ae696c16480
+      ea60df615387f47bdfdfe4da48f6567c4ec60f74
+      818a4b8bbcb50878a8b1b9f71b4274d242ab46bf860c74676e98dec1d0248821
+    ]
+    if bad_hashes.include?(actual.to_s)
+      message += <<-EOS.undent
+        ******************************************************************************
+        * ATTENTION: This hash is consistent with an error message for SourceForge's *
+        *            continuing service outage!                                      *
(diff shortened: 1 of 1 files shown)


mistydemeo:
Desperate times call for desperate measures. I'm OK with this.

Another possibility would be finding mirrors for every formula with a Sourceforge URL but given that that's close to 600 formulae I doubt we'd be done by the time Sourceforge is back...

DomT4:
Sadly mirrors alone wouldn't be enough, I presume? The checksum failure is fatal and we don't remove the failed file & try the mirror?

mistydemeo:
Yeah, we would need to replace the url (like Homebrew/homebrew-x11@1ec56b00cf6db0f8750cc939737eccf4d777fea9) or trigger a fallback on checksum failure.

L2G:
Too bad there isn't a way to create custom "stop" words for new GitHub issue tickets; then we could just throw that SHA-256 hash in there. :smile_cat: I figured this would be the next best thing.

DomT4:
Was thinking of an implementation where users can pass `--mirror` to `brew` at download time as well. It's been something I've wanted to do for a while, but we could also use the option automatically on certain known problematic URLs (And also later, when users are invoking the `HOMEBREW_NO_INSECURE_DOWNLOAD` variable).

DomT4:
@L2G I've considered mailing Github and asking them for a way to pin issues to the top of the list :crying_cat_face:. Every time SF goes down we get 30/40 issues.

mikemcquaid:
> I've considered mailing Github and asking them for a way to pin issues to the top of the list .

Reminder that I work there :wink:. There isn't a way, unfortunately.

mikemcquaid:
This makes sense to me :+1:

DomT4:
> Reminder that I work there 

Didn't really want to interrupt your time away with a plea for features :smile_cat:.

DomT4:
My only real "Eh" about this is that we'd have to remember to revert it fairly quickly once things were working again, and commits are not always easy to find in Homebrew's tree. But I'm tired of the many, *many* issues about it, so ¯\_(ツ)_/¯.

We don't recognise MD5 though, so not sure we should bother including that.

DomT4:
[It's alive!](https://twitter.com/sfnet_ops/status/622236074052464641)

[Mandatory celbratory gif](https://media1.giphy.com/media/uaVy4auyHhXry/200.gif)

mistydemeo:
Woo!

PR #41837, opened 2015-07-17
influxdb 0.9.2-rc1 (devel)

gunnaraasen:


Library/Formula/influxdb.rb
```diff
@@ -91,7 +96,9 @@ def install
 
     cd influxdb_path do
       if build.head?
```
dunn: If you want this to apply to `devel` as well as `head`, this should be `unless build.stable?`

dunn:
You'll need to add `version 0.9.2-rc1` to the `devel` block; Homebrew can't always figure out unusual version numbers without help.

Also, the sha is different than what you've given:
```
Warning: Formula reports different SHA256: 4311c39aa72d31bc23e7082a3af436f546038d640829a30782c5717623317d90
Downloaded to: /Library/Caches/Homebrew/influxdb-1.tar.gz
SHA1: 4b712476e0eb0b30bdb4ab91a7c3d215bf265629
SHA256: 1c90462fcacb1b14b17602c06fa6deb4161b3d9bff1c1d318c743bb1982062c9
```

Library/Formula/influxdb.rb
gunnaraasen: This is should only apply to `HEAD`.

gunnaraasen:
sha updated. Weird, brew didn't complain about the checksum mismatch when I tested the install locally.

dunn:
Still need to fix the `devel` version number before this can be merged. [Homebrew thinks it's Version 1](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/29747/version=yosemite/testReport/brew-test-bot/yosemite/install___devel_influxdb/):
```
==> Summary
/usr/local/Cellar/influxdb/1: 4 files, 25M, built in 10 seconds
```

gunnaraasen:
@dunn whoops, missed that comment. Version number added!

dunn:
Thank you! Merged in f879b42248fda8113bc3d7b907fa1c07685429c6.

PR #41842, opened 2015-07-18
Updated pdf2htmlex formula with v0.13.6 tar and sha

jamiely:


dunn:
@bfontaine I know this is included in #37280, but unless you expect the fix for poppler within a week or so, maybe we should merge this even if it means a revision after not too long.

bfontaine:
Yes merge this one, we’ll use a revision in #37280.

dunn:
Merged in 2177ee7864bb4e6d0c1be27ff672671a7e806184—thanks for contributing to Homebrew!

PR #41843, opened 2015-07-18
Revert "work around missing __debug header in Xcode CLT 6.3"

tdsmith:
The need for this workaround should have expired.

This reverts commit fd5ca9ba16f5afc181364369210cb94e997a5a5e.


tdsmith:
Oops, I should keep the change from 46bf9c4c here.

mikemcquaid:
:+1:

PR #41846, opened 2015-07-18
Build OpenNI2 driver for libfreenect

xionglingfeng:
Build `libfreenect` with `OpenNI2` support by default


dunn:
Is there any downside to this? Is it something that someone might want to build libfreenect without?

xionglingfeng:
@dunn  I don't think there are any downside here. I previously added a switch for this (PR #41476), but @mikemcquaid   suggests make this default.

dunn:
Oh, I did see that discussion, but I must have forgotten. Sorry about that, but should be good, then!

dunn:
And merged in 3d904164a39692cc96ef04bfedc79ab696d7452d. Thank you!!

PR #41848, opened 2015-07-18
cereal 1.1.2

guoxiao:
Add test block

### DIFF
--- a/Library/Formula/cereal.rb
+++ b/Library/Formula/cereal.rb
@@ -1,8 +1,8 @@
 class Cereal < Formula
   desc "C++11 library for serialization"
   homepage "https://uscilab.github.io/cereal/"
-  url "https://github.com/USCiLab/cereal/archive/v1.1.0.tar.gz"
-  sha1 "a7036f3fdb43315b0b6aa6c112c5878c03f1aa9e"
+  url "https://github.com/USCiLab/cereal/archive/v1.1.2.tar.gz"
+  sha256 "45607d0de1d29e84d03bf8eecf221eb2912005b63f02314fbade9fbabfd37b8d"
 
   head "https://github.com/USCiLab/cereal.git", :branch => "develop"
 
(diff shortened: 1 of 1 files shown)


bfontaine:
The test fails on ML:

```
Testing cereal
==> /usr/bin/clang -std=c++11 -lc++ -o test test.cpp
In file included from test.cpp:1:
In file included from /usr/local/include/cereal/types/unordered_map.hpp:33:
/usr/local/include/cereal/cereal.hpp:32:10: fatal error: 'type_traits' file not found
#include <type_traits>
         ^
1 error generated.
Error: cereal: failed
```

guoxiao:
Should we add a `depends_on :macos => :mavericks`?

xu-cheng:
Try to add `ENV.cxx11` inside `install` method.

Library/Formula/cereal.rb
```diff
@@ -16,8 +16,12 @@ class Cereal < Formula
   option "with-tests", "Build and run the test suite"
 
   depends_on "cmake" => :build if build.with? "tests"
+  depends_on :macos => :mavericks
```
bfontaine: Please remove this one to see if the `ENV.cxx11` fixed the issue on ML.

guoxiao: I've tried that before adding this line and it didn't solve the issue of ML.

bfontaine: Could you try with `Env.cxx11` inside the `test do` block?

guoxiao: That's exactly what I have tried. :D

bfontaine: cc @xu-cheng

xu-cheng: `ENV.cxx11` in the test won't work as the CFLAGS are never used. Try to compile the test file in this way:

```ruby
system ENV.cc, "-std=c++11", "-stdlib=libc++", "-lc++", "-o", "test", "test.cpp"
```

guoxiao:
It seems to be working well on ML now.

xu-cheng:

PR #41855, opened 2015-07-18
gnu-apl: use secure mirror

DomT4:
Kicked over from #41770 for reasons discussed there.

### DIFF
--- a/Library/Formula/gnu-apl.rb
+++ b/Library/Formula/gnu-apl.rb
@@ -1,11 +1,9 @@
-require "formula"
-
 class GnuApl < Formula
   desc "GNU implementation of the programming language APL"
-  homepage "http://www.gnu.org/software/apl/"
+  homepage "https://www.gnu.org/software/apl/"
   url "http://ftpmirror.gnu.org/apl/apl-1.4.tar.gz"
-  mirror "http://ftp.gnu.org/gnu/apl/apl-1.4.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
@mikemcquaid Any idea why the test-bot is incapable of installing the bottle generated here? It's a tiny build and tiny bottle, and I don't have any issues locally or on another VM, but every time I run it through the bot it freezes up until timeout :confounded:.

mikemcquaid:
@DomT4 Looks like the test is freezing rather than the install?

DomT4:
Well, now I feel like an idiot. I'll check it out.

DomT4:
:eyes: 

```
==> /usr/local/Cellar/gnu-apl/1.4/bin/apl -s -f hello.apl
Executable ./Cellar/.4/bipl/1.4/bin/APs/APserver not found.
This could means that 'apl' was not installed ('make install') or that it
was started in a non-standard way. The expected location of APserver is
either the same directory as the binary 'apl' or the subdirectory 'APs' of
that directory (the directory should also be in $PATH).
::connect() to existing APserver failed: Invalid argument
Svar_DB not connected in Svar_DB::is_registered_id()
```

That `Executable` path is er, unusual.

DomT4:
If you open a new shell after installation the `executable` path error goes away and is replaced by:

```
==> /usr/local/Cellar/gnu-apl/1.4/bin/apl -s -f hello.apl
::connect() to existing APserver failed: Invalid argument
Svar_DB not connected in Svar_DB::is_registered_id()
```

Will keep looking into it.

PR #41857, opened 2015-07-18
Update libepoxy to 1.3.1

linkmauve:
Also dropping the now-merged patch disabling GLX support.


xu-cheng:

PR #41867, opened 2015-07-18
trr: put version before checksum

wkentaro:


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/d96498c6f30f34d68812e1ab671454047414579a. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

We wouldn't normally accept a simple style fix like this without some other change as it can eat up not insignificant CI time, but I'll let it pass here.

PR #41868, opened 2015-07-18
Update, fix install on linux

amenk:
* Cocoa is not available on linux
* Updated version 4.13.1
* New HEAD URL (project switched to git)
* HEAD can currently not be build because configure is missing there

see also https://github.com/Homebrew/linuxbrew/issues/488

### DIFF
--- a/Library/Formula/gwenhywfar.rb
+++ b/Library/Formula/gwenhywfar.rb
@@ -1,9 +1,9 @@
 class Gwenhywfar < Formula
   desc "Utility library required by aqbanking and related software"
   homepage "http://www.aqbanking.de/"
-  url "http://www2.aquamaniac.de/sites/download/download.php?package=01&release=77&file=01&dummy=gwenhywfar-4.13.0.tar.gz"
-  sha1 "c4f37eb7fed069f3478e06a9311193a98cc9ddbf"
-  head "http://devel.aqbanking.de/svn/gwenhywfar/trunk"
+  url "http://www.aquamaniac.de/sites/download/download.php?package=01&release=78&file=01&dummy=gwenhywfar-4.13.1.tar.gz"
+  sha256 "4beca892c1235548ea0ae30132a6d2e57911c22340746585395ccb01d84ec72b"
+  head "http://git.aqbanking.de/git/gwenhywfar.git"
 
   bottle do
     sha1 "3855ce27d2b51561e7d5fbb720bb13363051d742" => :yosemite
@@ -18,7 +18,6 @@ class Gwenhywfar < Formula
   depends_on "libgcrypt"
   depends_on "gtk+" => :optional
   depends_on "qt" => :optional
-
   option "without-cocoa", "Build without cocoa support"
   option "with-check", "Run build-time check"
 


dunn:
We generally don't add things that are only for Linuxbrew, like `OS.mac?`, to the core. That does make things more complicated for Linux users, but if we tried to make Homebrew compatible with Linux by default that would require a lot of additional complexity.

DomT4:
The Linux fix will need to stay in Linuxbrew, per our standard rules.

sjackman:
http://bot.brew.sh/job/Homebrew%20Pull%20Requests/29802/version=yosemite/testReport/junit/brew-test-bot/yosemite/audit_gwenhywfar/
```
Error: 1 problem in 1 formula
gwenhywfar:
```

sjackman:
Please change the `sha1` to `sha256`.

Library/Formula/gwenhywfar.rb
```diff
@@ -18,15 +18,18 @@ class Gwenhywfar < Formula
...
   option "with-check", "Run build-time check"
 
   def install
     guis = []
     guis << "gtk2" if build.with? "gtk+"
     guis << "qt4" if build.with? "qt"
-    guis << "cocoa" if build.with? "cocoa"
+    if OS.mac?
+      guis << "cocoa" if build.with? "cocoa"
+    end
```
bfontaine: You can do that in one line:

```ruby
guis << "cocoa" if OS.mac? && build.with? "cocoa"
```

Library/Formula/gwenhywfar.rb
```diff
@@ -18,15 +18,18 @@ class Gwenhywfar < Formula
   depends_on "libgcrypt"
   depends_on "gtk+" => :optional
   depends_on "qt" => :optional
-
-  option "without-cocoa", "Build without cocoa support"
+  if OS.mac?
+    option "without-cocoa", "Build without cocoa support"
+  end
```
bfontaine: You can do that in one line:

```ruby
option "without-cocoa", "Build without cocoa support" if OS.mac?
```

mikemcquaid:
Please make this PR just the version update and then submit a separate PR to Linuxbrew to add the `OS.mac?` stuff, thanks.

amenk:
But why? ... to much work, sorry. But you can split it yourself if you want.

amenk:
Ah I got now the point that you mean to make a separate PR to the other project (Linuxbrew)
Maybe I understand https://github.com/Homebrew/linuxbrew/blob/master/CONTRIBUTING.md wrong - but I was thinking that homebrew should be easy to merge, so isn't it better if the code for Linux and Mac is the same so using `OS.mac?` should be fine?

mikemcquaid:
> so isn't it better if the code for Linux and Mac is the same so using OS.mac? should be fine?

Sorry, we don't do that in Homebrew core. Homebrew's code is OS X only.

sjackman:
@amenk You can send the version bump to `Homebrew/hombrew` and the Mac/Linux compatibility chunks to `Homebrew/linuxbrew`.

amenk:
updated

Library/Formula/gwenhywfar.rb
```diff
@@ -18,7 +18,6 @@ class Gwenhywfar < Formula
   depends_on "libgcrypt"
   depends_on "gtk+" => :optional
   depends_on "qt" => :optional
-
```
mikemcquaid: Please don't make unnecessary changes like this, thanks!

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `gwenhywfar 4.13.1`.

PR #41871, opened 2015-07-18
cppcms 1.0.5

guoxiao:
Add cppcms formula

### DIFF
--- a/Library/Formula/cppcms.rb
+++ b/Library/Formula/cppcms.rb
@@ -1,0 +1,86 @@
+class Cppcms < Formula
+  desc "A Free High Performance Web Development Framework"
+  homepage "http://cppcms.com/wikipp/en/page/main"
+  url "https://downloads.sourceforge.net/project/cppcms/cppcms/1.0.5/cppcms-1.0.5.tar.bz2"
+  sha256 "84b685977bca97c3e997497f227bd5906adb80555066d811a7046b01c2f51865"
+
+  depends_on "cmake" => :build
+  depends_on "pcre"
+  depends_on "openssl"
+  needs :cxx11
+
+  def install
+    ENV.cxx11
+    system "cmake", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"hello.cpp").write <<-EOS.undent
+      #include <cppcms/application.h>
+      #include <cppcms/applications_pool.h>
+      #include <cppcms/service.h>
+      #include <cppcms/http_response.h>
+      #include <iostream>
+      #include <string>
+
+      class hello : public cppcms::application {
+          public:
+              hello(cppcms::service& srv): cppcms::application(srv) {}
+              virtual void main(std::string url);
+      };
+
+      void hello::main(std::string /*url*/)
+      {
+          response().out() <<
+              "<html>\\n"
+              "<body>\\n"
+              "  <h1>Hello World</h1>\\n"
+              "</body>\\n"
+              "</html>\\n";
+      }
+
+      int main(int argc,char ** argv)
+      {
+          try {
+              cppcms::service srv(argc,argv);
+              srv.applications_pool().mount(
+                cppcms::applications_factory<hello>()
+              );
+              srv.run();
+              return 0;
+          }
+          catch(std::exception const &e) {
+              std::cerr << e.what() << std::endl;
+              return -1;
+          }
+      }
+    EOS
+
+    (testpath/"config.json").write <<-EOS.undent
+      {
+          "service" : {
+              "api" : "http",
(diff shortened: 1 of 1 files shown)


Library/Formula/cppcms.rb
```diff
@@ -0,0 +1,83 @@
...
+    EOS
+    system ENV.cc, "-o", "hello", "-stdlib=libc++", "-lc++", "-lcppcms", "hello.cpp"
+    pid = fork { system "./hello", "-c", "config.json" }
+
+    sleep 1 # grace time for server start
+    begin
+      assert_match /Hello World/, shell_output("curl http://127.0.0.1:8080/hello")
+    ensure
+      Process.kill 9, pid
+      Process.kill 9, pid + 2 # kill child process
```
guoxiao: A problem now: The test will fork a child process and is not easy to kill it. Maybe a better solution is needed.

mistydemeo:
Can you add an openssl dependency?

guoxiao:
Fails on ML. Do I need to add some special flags?

guoxiao:
@BrewTestBot Test this please.

DomT4:
The test-bot ignores pings from people who don't have admin access to it I'm afraid. I'll ask it for you.

DomT4:
@BrewTestBot test this please

mikemcquaid:
Looks like it hasn't compiled properly on 10.8 as the test is failing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/30609/version=mountain_lion/testReport/junit/brew-test-bot/mountain_lion/test_cppcms/

Library/Formula/cppcms.rb
```diff
@@ -0,0 +1,84 @@
...
+  homepage "http://cppcms.com/wikipp/en/page/main"
+  url "https://downloads.sourceforge.net/project/cppcms/cppcms/1.0.5/cppcms-1.0.5.tar.bz2"
+  sha256 "84b685977bca97c3e997497f227bd5906adb80555066d811a7046b01c2f51865"
+
+  depends_on "cmake" => :build
+  depends_on "pcre"
+  depends_on "openssl"
+
+  def install
+    system "cmake", "-DCMAKE_INSTALL_PREFIX:PATH=#{prefix}", "."
```
mikemcquaid: Use `std_cmake_args` instead.

Library/Formula/cppcms.rb
```diff
@@ -0,0 +1,85 @@
+class Cppcms < Formula
+  desc "A Free High Performance Web Development Framework"
+  homepage "http://cppcms.com/wikipp/en/page/main"
+  url "https://downloads.sourceforge.net/project/cppcms/cppcms/1.0.5/cppcms-1.0.5.tar.bz2"
+  sha256 "84b685977bca97c3e997497f227bd5906adb80555066d811a7046b01c2f51865"
+
+  depends_on "cmake" => :build
+  depends_on "pcre"
+  depends_on "openssl"
+  depends_on :macos => :mavericks
```
dunn: Is there no way to get it to compile on Mountain Lion?

guoxiao: I tried several build options but couldn't get it work.

mikemcquaid: Did you try a `fails_with` or `needs :cxx11`?

guoxiao: In a previous PR #41848 @xu-cheng commented:
> ENV.cxx11 in the test won't work as the CFLAGS are never used,  Try to compile the test file in this way:
> ```
> system ENV.cc, "-std=c++11", "-stdlib=libc++", "-lc++", "-o", "test", "test.cpp"
> ```

It works well in `cereal`. However, in this formula I also added these params, but it doesn't work.

guoxiao: > Did you try a fails_with or needs :cxx11?
Would you please tell me how to use these options in formulas? Or where is the documents/examples of these options? Thanks.

mikemcquaid: I'm afraid they aren't documented but you can `grep` other formulae to see usage of `fails_with` or `needs :cxx11`

DomT4:
Status update here?

guoxiao:
`needs :cxx11` doesn't work here,  and I don't know how to use `fails_with` to handle this.

DomT4:
How much does everyone care here about making it work on 10.8 and below?

dunn:
@guoxiao did you try `needs :cxx11` above `def install`, and then `ENV.cxx11` inside the install block? Sometimes both are needed.

mikemcquaid:
Thanks @guoxiao!

PR #41877, opened 2015-07-18
Add CC and CXX compiler information to receipt

kilojoules:
Often I need to know what CC/CXX was used when installing a package. The idea here is that this information will become available through `brew info --json=v1` .

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -775,6 +775,8 @@ class Formula
           "used_options" => tab.used_options.as_flags,
           "built_as_bottle" => tab.built_bottle,
           "poured_from_bottle" => tab.poured_from_bottle
+          if ENV["HOMEBREW_CXX"]: "CXX_Compiler" => ENV["HOMEBREW_CXX"]
+          if ENV["HOMEBREW_CC"]: "CC_Compiler" => ENV["HOMEBREW_CC"]
         }
       end
 


Library/Homebrew/formula.rb
```diff
@@ -775,6 +775,8 @@ def to_hash
           "used_options" => tab.used_options.as_flags,
           "built_as_bottle" => tab.built_bottle,
           "poured_from_bottle" => tab.poured_from_bottle
+          if ENV["HOMEBREW_CXX"]: "CXX_Compiler" => ENV["HOMEBREW_CXX"]
```
mistydemeo: This isn't valid Ruby syntax. You can't include conditionals like this in the middle of a hash literal, and Ruby's `if` statements are structured differently from Python ones.

kilojoules: Thanks for the feedback. If I develop this idea and come back with correct syntax could it be accepted? Is there a reason brew doesn't include the CC/CXX info in the receipt?

mistydemeo:
It looks like you've added this to `Formula#to_hash`, not the place that tabs are written. In this case `HOMEBREW_CC` will probably not be the actual value used to build the formula, if it's defined at all.

Library/Homebrew/formula.rb
mistydemeo: I'd be fine with including it. We haven't stored this before simply because it hasn't been requested in the past, and because users typically have access to the build logs in `~/Library/Logs/Homebrew` that makes it possible to determine what compiler was used.

What is your usecase, out of curiosity?

kilojoules: I'm using boost and boost-python for a project interfacing python with C libraries which have lots of old C/C++ code involved, and I've found that to build the  I need to install boost with HOMEBREW_CXX as g++-4.9.  Other components of this project have compiler flags which only work with clang, and I often must switch to clang boost. Its hard to keep track of which boost version I currently have installed, and being able to access this info without digging through the Homebrew logs seems useful (honestly I did not think to look in `~/Library/Logs/Homebrew/boost`... Looking in there now, its not immediately clear how I can access the `$HOMEBREW_CXX` flag used).

DomT4:
Please don't push any further commits on this for now without running it past a maintainer first. The PR seems to have caused some CI environmental leakage that is killing builds across various taps - You couldn't know in advance, and we're working on it, but let's tread carefully.

kilojoules:
I'm sorry to have caused these issues. I am excited about learning more about ruby and contributing to the homebrew project. How do I run future proposed changes by a maintainer? Should I use this thread to suggest changes?

Library/Homebrew/formula.rb
mistydemeo: Something that may be useful to you is that we store the C++ stdlib which was built against in the tab - GCC's libstdc++ or clang's libc++. If that's the reason you're needing to build boost using GCC, that may give you the information you need.

DomT4:
Yeah, just leave comments in here with proposed diffs or such and one of us will review them.

kilojoules:
@mistydemeo and @DomT4 ,

Thank you for your support of this PR. I really appreciate your suggestions and patience. Here are the proposed changes. I tested these on my system and produced the intended results.

    diff -r homebrew/Library/Homebrew/formula.rb /usr/local/homebrew/Library/Homebrew/formula.rb
    785c785,787
    <           "poured_from_bottle" => tab.poured_from_bottle
    ---
    >           "poured_from_bottle" => tab.poured_from_bottle,
    >           "CC_Compiler" => tab.CC_Compiler,
    >           "CXX_Compiler" => tab.CXX_Compiler
    diff -r homebrew/Library/Homebrew/tab.rb /usr/local/homebrew/Library/Homebrew/tab.rb
    184a185,197
    >   
    >     # record CC and CXX compiler info
    >     if ENV['HOMEBREW_CC'] then
    >       cc_compiler = ENV['HOMEBREW_CC']
    >     else
    >       cc_compiler = 'clang' 
    >     end
    >     if ENV['HOMEBREW_CXX'] then
    >       cxx_compiler = ENV['HOMEBREW_CXX']
    >     else
    >       cxx_compiler = 'clang' 
    >     end
    >

kilojoules:
@DomT4 @mistydemeo -

Here are the proposed changes

    diff --git a/Library/Homebrew/formula.rb b/Library/Homebrew/formula.rb
    index bf11af0..4906f5d 100644
    --- a/Library/Homebrew/formula.rb
    +++ b/Library/Homebrew/formula.rb
    @@ -774,7 +774,9 @@ class Formula
               "version" => keg.version.to_s,
               "used_options" => tab.used_options.as_flags,
               "built_as_bottle" => tab.built_bottle,
    -          "poured_from_bottle" => tab.poured_from_bottle
    +          "poured_from_bottle" => tab.poured_from_bottle,
    +          "CC_Compiler" => tab.CC_Compiler,
    +          "CXX_Compiler" => tab.CXX_Compiler
             }
           end
     
    diff --git a/Library/Homebrew/tab.rb b/Library/Homebrew/tab.rb
    index 37b0db2..10a8994 100644
    --- a/Library/Homebrew/tab.rb
    +++ b/Library/Homebrew/tab.rb
    @@ -182,6 +182,16 @@ class Tab < OpenStruct
       end
     
       def to_json
    +  if ENV['HOMEBREW_CC'] then
    +    cc_compiler = ENV['HOMEBREW_CC']
    +  else
    +    cc_compiler = 'clang'
    +  end
    +  if ENV['HOMEBREW_CXX'] then
    +    cxx_compiler = ENV['HOMEBREW_CXX']
    +  else
    +    cxx_compiler = 'clang' 
    +  end
         attributes = {
           "used_options" => used_options.as_flags,
           "unused_options" => unused_options.as_flags,
    @@ -192,6 +202,8 @@ class Tab < OpenStruct
           "stdlib" => (stdlib.to_s if stdlib),
           "compiler" => (compiler.to_s if compiler),
           "source" => source,
    +      "CC_Compiler" => cc_compiler,
    +      "CXX_Compiler" => cxx_compiler  
         }
     
         Utils::JSON.dump(attributes)

DomT4:
I presume our lowercasing of the variables is deliberate and this should perhaps should be mimicked here. The `ENV` stuff is fine in uppercase, that's per our norm. Makes sure we use double quotes around `"clang"` and such as well - Single quotes is something we're slowly fixing over time. It's looking better though, Thanks.

Will let Misty review as well, I try to avoid Homebrew's compiler dancing as much as possible :smile_cat:.

kilojoules:
Thank you for the feedback. Here are the revised changes. 

    diff --git a/Library/Homebrew/formula.rb b/Library/Homebrew/formula.rb
    index bf11af0..1c1caf3 100644
    --- a/Library/Homebrew/formula.rb
    +++ b/Library/Homebrew/formula.rb
    @@ -774,7 +774,9 @@ class Formula
               "version" => keg.version.to_s,
               "used_options" => tab.used_options.as_flags,
               "built_as_bottle" => tab.built_bottle,
    -          "poured_from_bottle" => tab.poured_from_bottle
    +          "poured_from_bottle" => tab.poured_from_bottle,
    +          "cc_compiler" => tab.cc_compiler,
    +          "cxx_compiler" => tab.cxx_compiler
             }
           end
     
    diff --git a/Library/Homebrew/tab.rb b/Library/Homebrew/tab.rb
    index 37b0db2..de0fe11 100644
    --- a/Library/Homebrew/tab.rb
    +++ b/Library/Homebrew/tab.rb
    @@ -182,6 +182,16 @@ class Tab < OpenStruct
       end
     
       def to_json
    +  if ENV['HOMEBREW_CC'] then
    +    cc_compiler = ENV['HOMEBREW_CC']
    +  else
    +    cc_compiler = "clang"
    +  end
    +  if ENV['HOMEBREW_CXX'] then
    +    cxx_compiler = ENV['HOMEBREW_CXX']
    +  else
    +    cxx_compiler = "clang" 
    +  end
         attributes = {
           "used_options" => used_options.as_flags,
           "unused_options" => unused_options.as_flags,
    @@ -192,6 +202,8 @@ class Tab < OpenStruct
           "stdlib" => (stdlib.to_s if stdlib),
           "compiler" => (compiler.to_s if compiler),
           "source" => source,
    +      "cc_compiler" => cc_compiler,
    +      "cxx_compiler" => cxx_compiler  
         }
     
         Utils::JSON.dump(attributes)

mikemcquaid:
@kilojoules As @mistydemeo mentioned I'm 99% sure you care about the stdlib (which is already stored) rather than the compiler.

jacknagel:
the compiler is already in the receipt (e.g. "clang", "gcc48", etc). you can infer the c++ compiler from that as well.

kilojoules:
Where is the compiler in the receipt? I can't find it...


    $ brew info --json=v1 boost
    [{"name":"boost","full_name":"boost","desc":"Collection of portable C++ source libraries","homepage":"http://www.boost.org","versions":{"stable":"1.58.0","bottle":true,"devel":null,"head":"HEAD"},"revision":0,"installed":[{"version":"1.58.0","used_options":["--c++11","--with-mpi","--without-single"],"built_as_bottle":null,"poured_from_bottle":false}],"linked_keg":"1.58.0","keg_only":null,"dependencies":["icu4c"],"conflicts_with":[],"caveats":"","requirements":[{"name":"mpi","default_formula":"open-mpi","cask":null,"download":null}],"options":[{"option":"--universal","description":"Build a universal binary"},{"option":"--with-icu4c","description":"Build regexp engine with icu support"},{"option":"--without-single","description":"Disable building single-threading variant"},{"option":"--without-static","description":"Disable building static library variant"},{"option":"--with-mpi","description":"Build with MPI support"},{"option":"--c++11","description":"Build using C++11 mode"}]}]
    
    $ brew info boost
    boost: stable 1.58.0 (bottled), HEAD
    Collection of portable C++ source libraries
    http://www.boost.org
    /usr/local/Cellar/boost/1.58.0 (10668 files, 368M) *
      Built from source with: --c++11 --with-mpi --without-single
    From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/boost.rb
    ==> Dependencies
    Optional: icu4c ✘
    ==> Options
    --c++11
    	Build using C++11 mode
    --universal
    	Build a universal binary
    --with-icu4c
    	Build regexp engine with icu support
    --with-mpi
    	Build with MPI support
    --without-single
    	Disable building single-threading variant
    --without-static
    	Disable building static library variant
    --HEAD
    	Install HEAD version

mikemcquaid:
@kilojoules Can you check if the stdlib check is enough? That should be what matters rather than the compiler.

mistydemeo:
We also track the C compiler since 099ce2f83058dd3f03721d695f321dbe143f4eec - we could add that to the JSON output.

bfontaine:
Ping?

DomT4:
Closing as dead, and per Misty/Mike/Jack's comments. If you wish to add the information to the JSON dump feel free to submit a PR for that as desired. Thanks for the submission!

kilojoules:
This feature would still be useful for me. I made the changes I suggested in a local version and kind of abandoned this PR. But, @mistydemeo, I would appreciate this feature being added to homebrew. This would be an easy and intuitive way to check what compiler was used to build a package, which I have found crucial in debugging some flat namespace errors in a project.

mikemcquaid:
@kilojoules Please see above: "the compiler is already in the receipt (e.g. "clang", "gcc48", etc). you can infer the c++ compiler from that as well."

PR #41880, opened 2015-07-19
GNU Various Pt3.

DomT4:
Nearly done with these http => https moves.


mikemcquaid:
:+1:

DomT4:
A lot of the old GNU formula are some of the bigger offenders on using outdated style, not having tests, not having bottles, dependencies being set wrong, etc so I decided to try and fix everything rather than just fixing the mirror link issue. Haven't come up with a test for everything so far, but progress :sweat_smile:.

PR #41886, opened 2015-07-19
Include git sha1 in --version output

tdsmith:
HOMEBREW_VERSION doesn't change very often; the repository state is more
interesting.

Does this need more robust error handling? I can't think of a way this will fail.

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -155,6 +155,10 @@ module Homebrew
     HOMEBREW_REPOSITORY.cd { `git show -s --format="%cd" --date=short HEAD 2>/dev/null`.chuzzle }
   end
 
+  def self.homebrew_version_string
+    "#{HOMEBREW_VERSION} (git revision #{Homebrew.git_head || "unknown"}; last commit #{Homebrew.git_last_commit_date || "unknown"})"
+  end
+
   def self.install_gem_setup_path!(gem, version = nil, executable = gem)
     require "rubygems"
     ENV["PATH"] = "#{Gem.user_dir}/bin:#{ENV["PATH"]}"
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -151,6 +151,10 @@ module Homebrew
     HOMEBREW_REPOSITORY.cd { `git show -s --format="%cr" HEAD 2>/dev/null`.chuzzle }
   end
 
+  def self.git_last_commit_date
+    HOMEBREW_REPOSITORY.cd { `git show -s --format="%cd" --date=short HEAD 2>/dev/null`.chuzzle }
+  end
+
   def self.install_gem_setup_path!(gem, version = nil, executable = gem)
     require "rubygems"
     ENV["PATH"] = "#{Gem.user_dir}/bin:#{ENV["PATH"]}"
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -147,6 +147,10 @@ module Homebrew
     HOMEBREW_REPOSITORY.cd { `git rev-parse --verify -q HEAD 2>/dev/null`.chuzzle }
   end
 
+  def self.git_short_head
+    HOMEBREW_REPOSITORY.cd { `git rev-parse --short=4 --verify -q HEAD 2>/dev/null`.chuzzle }
+  end
+
   def self.git_last_commit
     HOMEBREW_REPOSITORY.cd { `git show -s --format="%cr" HEAD 2>/dev/null`.chuzzle }
   end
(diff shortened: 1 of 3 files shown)


mikemcquaid:
Can you provide some sample output? Would be good to know what happens if `.git` is deleted.

tdsmith:
```
tim@rocketman:Homebrew (version-date)$ bin/brew -v
Homebrew 0.9.5 (be1b64bfa575ebb858861514d43db6c6f4e35fcf; 2015-07-18)

tim@rocketman:homebrew (version-date)$ mv .git .git-moved
tim@rocketman:homebrew$ bin/brew -v
Homebrew 0.9.5 (; )
```

Library/Homebrew/utils.rb
```diff
@@ -121,6 +121,10 @@ def self.git_last_commit
     HOMEBREW_REPOSITORY.cd { `git show -s --format="%cr" HEAD 2>/dev/null`.chuzzle }
   end
 
+  def self.git_last_commit_date
+    HOMEBREW_REPOSITORY.cd { `git show -s --format="%cd" --date=short HEAD 2>/dev/null`.chuzzle }
+  end
```
bfontaine: Don’t we already have that with `git_last_commit`?

```
$ brew irb
>> Homebrew.git_last_commit
=> "2 hours ago"
>>
```

This is used for the `config` command output.

tdsmith: Yeah, it's relative vs. absolute time; I think absolute is probably more useful.

bfontaine: You’re right.

DomT4:
This PR is :heart_eyes:. Possibly worth only printing the commit & absolute date if we know the `.git` folder exists, but otherwise vastly better.

tdsmith:
Revised to make the missing-git condition prettier.

```
tim@rocketman:homebrew (version-date)$ mv .git .git-moved
tim@rocketman:homebrew$ bin/brew -v
Homebrew 0.9.5 (unknown; unknown)
```

dunn:
This is getting a little involved, but "unknown" might not help if they don't know _what_ is unknown.  Maybe something like this?
```ruby
version_string = "#{HOMEBREW_VERSION}"
if (HOMEBREW_REPOSITORY/".git").exist?
  version_string += " (#{Homebrew.git_head}; #{Homebrew.git_last_commit_date}"
else
  version_string += " (no commit history)"
end

if ARGV.first == '--version'
  puts version_string
  exit 0
elsif ARGV.first == '-v'
  puts "Homebrew #{version_string}"
```

tdsmith:
I want to avoid shelling out to git on every `brew` invocation but we could refactor the string assembly into a `homebrew_version_string` method.

I think the most common way we'll encounter the missing .git condition is installing from a tarball, like we recommend here: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Installation.md#untar-anywhere -- I assume update initializes a git repo when it has to.

We could also just print `0.9.5 (git revision deadbeefdeadbeefdeadbeefdeadbeef; last commit 2015-07-19)` to be generally less cryptic, which is probably nicer.

dunn:
> We could also just print 0.9.5 (git revision deadbeefdeadbeefdeadbeefdeadbeef; last commit 2015-07-19) to be generally less cryptic, which is probably nicer.

That sounds good.

mikemcquaid:
I like the look of @dunn's approach. Agreed that the tarball install is probably the most likely situation.

dunn:
This seems good; "git revision unknown; last commit unknown" should be clear enough.

Library/Homebrew/utils.rb
```diff
@@ -121,6 +121,14 @@ def self.git_last_commit
...
+
+  def self.homebrew_version_string
+    "#{HOMEBREW_VERSION} (git revision #{Homebrew.git_head || "unknown"}; last commit #{Homebrew.git_last_commit_date || "unknown"})"
```
mikemcquaid: Let's use a new `git_short_head` method which calls `git rev-parse --short=4 --verify -q HEAD 2>/dev/null` which gives the shortest possible unique SHA for the commit. Also, can we pull the `||` outside of the string (so the line isn't more than 80 characters) and not print two unknowns in the case we don't have a `git_head`; in that case we known the last commit will always be `unknown` so just printing something like `(no git repository)` is a better idea.

tdsmith:
Rebased and amended.

DomT4:
Apparently one of the tests has wandered off.

Now we've merged the GSoC stuff I think there's merit to bump the Homebrew version anyway, but either way, keeping track of the git sha in the `--version` output is always going to be more accurate a measure of up-to-date-ness. Test failure aside, remain very :+1: here.

Library/Homebrew/utils.rb
```diff
@@ -147,10 +147,28 @@ def self.git_head
...
     HOMEBREW_REPOSITORY.cd { `git show -s --format="%cr" HEAD 2>/dev/null`.chuzzle }
   end
 
+  def self.git_last_commit_date
+    HOMEBREW_REPOSITORY.cd { `git show -s --format="%cd" --date=short HEAD 2>/dev/null`.chuzzle }
+  end
+
+  def self.homebrew_version_string
+    pretty_revision = git_short_head
+    if pretty_revision
```
mikemcquaid: Could do `if pretty_revision = git_short_head` but feel free to leave as-is unless you're making other changes.

mikemcquaid:
:+1: LGTM

Issue #41883, opened 2015-07-19
Measure code coverage of `brew tests`


mikemcquaid:
It'd be good to know what the coverage is of the Formula DSL and the commands. I don't think 100% is necessary or desirable but I'd like to at least see 100% file coverage (e.g. we make use of every file in `Library/Homebrew`)

bfontaine:
See #41890.

mikemcquaid:
We did this.

PR #41890, opened 2015-07-19
'brew tests' optional coverage measure added

bfontaine:
See #41883.

This measures the code coverage of `brew tests` if `--coverage` is passed:

```
$ brew tests --coverage
Run options: --seed 6892

# Running:

...........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Finished in 5.459945s, 101.6494 runs/s, 226.7422 assertions/s.

555 runs, 1238 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Unit Tests to /usr/local/Library/Homebrew/coverage. 4154 / 6971 LOC (59.59%) covered.
```

It generates a nice report in `/usr/local/Library/Homebrew/coverage/index.html` that looks like this:

screenshot

You can then click on each file to see the coverage:

screenshot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 /Library/Homebrew/test/.bundle
 /Library/Homebrew/test/bin
 /Library/Homebrew/test/vendor
+/Library/Homebrew/test/coverage
 /Library/LinkedKegs
 /Library/PinnedKegs
 /Library/Taps
--- a/Library/Homebrew/test/testing_env.rb
+++ b/Library/Homebrew/test/testing_env.rb
@@ -1,6 +1,22 @@
 $:.unshift File.expand_path("../..", __FILE__)
 $:.unshift File.expand_path("../lib", __FILE__)
 
+# This must be at the top
+if ENV["HOMEBREW_TESTS_COVERAGE"]
+  require "simplecov"
+  SimpleCov.start do
+    tests_path = File.dirname(__FILE__)
+
+    minimum_coverage 50
+    coverage_dir File.expand_path("#{tests_path}/coverage")
+    root File.expand_path("#{tests_path}/..")
+
+    add_filter "Homebrew/test"
+    add_filter "vendor/bundle"
+    add_filter "Homebrew/vendor"
+  end
+end
+
 require "global"
 
 # Test environment setup
--- a/Library/Homebrew/cmd/tests.rb
+++ b/Library/Homebrew/cmd/tests.rb
@@ -2,6 +2,7 @@ module Homebrew
   def tests
     (HOMEBREW_LIBRARY/"Homebrew/test").cd do
       ENV["TESTOPTS"] = "-v" if ARGV.verbose?
+      ENV["HOMEBREW_TESTS_COVERAGE"] = "1" if ARGV.flag? "--coverage"
       Homebrew.install_gem_setup_path! "bundler"
       quiet_system("bundle", "check") || \
         system("bundle", "install", "--path", "vendor/bundle")
--- a/Library/Homebrew/test/Gemfile
+++ b/Library/Homebrew/test/Gemfile
@@ -3,3 +3,4 @@ source "https://rubygems.org"
 gem "mocha", "~> 1.1"
 gem "minitest", "~> 5.3"
 gem "rake", "~> 10.3"
+gem "simplecov", "~> 0.10.0", :require => false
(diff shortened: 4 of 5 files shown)


Library/Homebrew/test/test_0.rb
```diff
@@ -0,0 +1,12 @@
+# This must be the first test file
```
xu-cheng: Why a separate file rather than inside `Library/Homebrew/test/testing_env.rb`?

xu-cheng:
Also, it would be better if we upload the coverage result to https://coveralls.io

Library/Homebrew/test/test_0.rb
bfontaine: I didn’t see this file, thanks (fixed).

bfontaine:
@xu-cheng If we do that, it’s as simple as adding [two lines of code](https://github.com/bfontaine/ud/blob/master/tests/tests.rb#L4-L5) in `testing_env.rb`:

```ruby
require 'coveralls'
Coveralls.wear!
```

(and register on their website)

mikemcquaid:
:clap: Nice work here. Rather than (or as well as) using Coveralls I'd like to integrate this into Jenkins.

.gitignore
```diff
@@ -17,6 +17,7 @@
 /Library/Homebrew/test/.bundle
 /Library/Homebrew/test/bin
 /Library/Homebrew/test/vendor
+/Library/Homebrew/coverage
```
mikemcquaid: Let's use `test/coverage` for the directory.

mikemcquaid:
Might be cool to consider using `ci_reporter` to create JUnit compatible output for Jenkins too.

Library/Homebrew/test/testing_env.rb
```diff
@@ -1,6 +1,18 @@
...
 
+# This must be at the top
+if ENV["HOMEBREW_TESTS_COVERAGE"]
+  require "simplecov"
+  SimpleCov.start do
+    SimpleCov.root File.expand_path("#{File.dirname(__FILE__)}/..")
+    add_filter do |src|
+      src.filename.include?("Homebrew/test") ||
+      src.filename.include?("vendor/bundle")
+    end
```
mikemcquaid: Can just do:

```ruby
    add_filter "Homebrew/test"
    add_filter "vendor/bundle"
```

Library/Homebrew/test/testing_env.rb
```diff
@@ -1,6 +1,18 @@
...
+  end
```
mikemcquaid: Might be good to add an aspirational `# minimum_coverage 50` (or something lower or equal to the current value) to gently stop us from adding new code without tests (I'm guilty).

.gitignore
bfontaine: Yep, fixed.

Library/Homebrew/test/testing_env.rb
bfontaine: Done.

Library/Homebrew/test/testing_env.rb
bfontaine: We’re at 59.59%. Let’s use 50 then we’ll bump it to 60 when we’ll have more tests.

bfontaine:
@mikemcquaid I’m not familiar with `ci_reporter`, can it report test coverage too? According to [a blog post](https://lostechies.com/ryansvihla/2011/09/25/rail-3-1-ci-setup-with-jenkins-test-unit-simplecov-on-os-x-lion/) we can use [`simplecov-rcov`](https://github.com/fguillen/simplecov-rcov) to create an Rcov report which is understandable by Jenkins (using a plugin).

mikemcquaid:
> According to a blog post we can use simplecov-rcov to create an Rcov report which is understandable by Jenkins (using a plugin).

Yeh, that sounds good. We can do it in a later PR, though. Feel free to merge this!

Library/Homebrew/test/testing_env.rb
```diff
@@ -1,6 +1,21 @@
...
+  require "simplecov"
+  SimpleCov.start do
+    tests_path = File.dirname(__FILE__)
+
+    minimum_coverage 50
+    coverage_dir File.expand_path("#{tests_path}/coverage")
+    root File.expand_path("#{tests_path}/..")
+
+    add_filter "Homebrew/test"
+    add_filter "vendor/bundle"
```
xu-cheng: We probably want to add `add_filter "Homebrew/vendor"`.

bfontaine: Done, thanks.

PR #41901, opened 2015-07-19
check_postgres 2.22.0

nijikon:


xu-cheng:
Test fails:
```
        Error: 1 problem in 1 formula
check_postgres:
```

Also it would be extreme grateful if you can test to install this formula with `--sandbox`(i.e. `brew install check_postgres --sandbox`) to check whether the files are installed the right places. (We find postgres related formulae tend to install files in wrong place).

nijikon:
@xu-cheng 

```
balrog:local n$ brew install check_postgres --sandbox --debug
/usr/local/Library/brew.rb (Formulary::FormulaLoader): loading /usr/local/Library/Formula/check_postgres.rb
Building from source because Library/Formula/check_postgres.rb has local changes
/usr/local/Library/Homebrew/build.rb (Formulary::FromPathLoader): loading /usr/local/Library/Formula/check_postgres.rb
/usr/local/Library/Homebrew/build.rb (Formulary::FormulaLoader): loading /usr/local/Library/Formula/postgresql.rb
==> Downloading http://bucardo.org/downloads/check_postgres-2.22.0.tar.gz
Already downloaded: /Library/Caches/Homebrew/check_postgres-2.22.0.tar.gz
==> perl Makefile.PL PREFIX=/usr/local/Cellar/check_postgres/2.22.0
==> make install
/usr/local/Library/Homebrew/postinstall.rb (Formulary::FromPathLoader): loading /usr/local/Library/Formula/check_postgres.rb
🍺  /usr/local/Cellar/check_postgres/2.22.0: 5 files, 528K, built in 2 seconds
```

xu-cheng:
Thanks. 

If you like, you can set variable `HOMEBREW_SANDBOX` which will have same effect of `--sandbox` flag that is building all formulae inside an apple sandbox. This can be helpful to check whether some files are in the wrong place.

Library/Formula/check_postgres.rb
```diff
@@ -30,6 +28,6 @@ def install
     # This test verifies that check_postgres fails correctly, assuming
     # that no server is running at that port.
     output = shell_output("#{bin}/check_postgres --action=connection --port=65432", 2)
-    assert output.include? "POSTGRES_CONNECTION CRITICAL"
+    assert_match "POSTGRES_CONNECTION CRITICAL"
```
xu-cheng: This should be `assert_match /POSTGRES_CONNECTION CRITICAL/, output`.

nijikon: rubocop states ```ambiguous first argument; put parentheses or a space even after `/' operator```

nijikon:
@xu-cheng any more useful tips when rebuilding formulas?

xu-cheng:
Thanks @nijikon. That's all the tips so far.(It's not documented because it's still unstable and in developing)

PR #41905, opened 2015-07-19
GNU Various Pt4

DomT4:
Pt4 completes the death of `http://ftp.gnu.org` :tada:.

### DIFF
--- a/Library/Formula/texinfo.rb
+++ b/Library/Formula/texinfo.rb
@@ -2,8 +2,8 @@ class Texinfo < Formula
   desc "Official documentation format of the GNU project"
   homepage "https://www.gnu.org/software/texinfo/"
   url "http://ftpmirror.gnu.org/texinfo/texinfo-5.2.tar.gz"
-  mirror "http://ftp.gnu.org/gnu/texinfo/texinfo-5.2.tar.gz"
-  sha1 "dc54edfbb623d46fb400576b3da181f987e63516"
+  mirror "https://ftp.gnu.org/gnu/texinfo/texinfo-5.2.tar.gz"
+  sha256 "6b8ca30e9b6f093b54fe04439e5545e564c63698a806a48065c0bba16994cf74"
 
   bottle do
     sha1 "988fc8c195a43ad8b9dea1da2827fb24c794c200" => :yosemite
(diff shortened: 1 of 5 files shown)


Library/Formula/texinfo.rb
```diff
@@ -28,8 +28,8 @@ def install
     # However, it complains about installing epsf.tex in TEXMF/tex/generic/dvips, so let's do that...
     # This somewhat breaks the homebrew philosophy, I am sorry.
     # Also, we don't depend on tex-live, but this directory only exists if it is installed.
-    if File.exist? "#{HOMEBREW_PREFIX}/share/texmf-dist/" then
-      cp "doc/epsf.tex", "#{HOMEBREW_PREFIX}/share/texmf-dist/tex/generic/dvips/"
+    if File.exist?(HOMEBREW_PREFIX/"share/texmf-dist/")
+      cp "doc/epsf.tex", HOMEBREW_PREFIX/"share/texmf-dist/tex/generic/dvips/"
```
xu-cheng: Won't this break under sandbox, which also means we will have a bad bottle.

DomT4: Pretty sure the answer is yes here; I just wasn't sure how to fix it best. I guess we install the `epsf.tex` file to `libexec` and then move it `post_install` or something?

xu-cheng: First of all, why we need this? I'm confusing about the comments.

DomT4: It was added with the formula, 5 years ago, and (predictably) hasn't been touched since. https://github.com/Homebrew/homebrew/commit/3f514089

xu-cheng: Let's remove it. `HOMEBREW_PREFIX/"share/texmf-dist/"` won't be created by any known software. (The path inside in MacTeX is `/usr/local/texlive/<version>/texmf-dist/`). So I guess this part of codes never got run in this 5 years.

DomT4: Will do.

These kinds of situations seem to be why we're much stricter about not adding them if possible these days, because 5 years down the line the code remains there and nobody quite knows why.

PR #41908, opened 2015-07-19
cig: remove redundant resource

dunn:
cig itself was a `go_resource` because the stable version just wasn't
being staged correctly.


stevenjack:
@dunn I was basing my formula from: https://github.com/Homebrew/homebrew/blob/634a9385799b4e7c049bcfe62a8b42efc6e1f0fc/Library/Formula/docker-machine.rb#L40-L42

but what you've done in this PR makes much more sense :+1: , did seem weird how I was specifying the source and the package itself in the go dependencies.

dunn:
Yeah, docker-machine is an odd case, I haven't figured out how to get rid of the redundancy there.

PR #41913, opened 2015-07-19
Stop flagging explicit from-source modified builds

DomT4:
Follow up to https://github.com/Homebrew/homebrew/commit/1134222baf2e91d3e2581d7651494d92de513715.

Essentially at the moment we're seeing it where even when I locally tell `brew` I'll be doing `-fs` installation it throws the warning at me about having modified the file and it'll be installing from source. I know it'll be installing from source, I just told it to :wink::

screen shot 2015-07-19 at 16 13 23

It's also popping up on the [test-bot](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/29868/version=mountain_lion/console).

I'm happy to wrap the new changes in `HOMEBREW_DEVELOPER` if people are happier with that. The warning formulae have been modified is still potentially useful to people who don't regularly pull and make local changes all over the place, but on the test-bot and when developers explicitly pass `fs` it's a bit redundant and potentially confusing.

### DIFF
--- a/Library/Homebrew/cmd/fetch.rb
+++ b/Library/Homebrew/cmd/fetch.rb
@@ -32,6 +32,7 @@ module Homebrew
   def fetch_bottle? f
     return true if ARGV.force_bottle? && f.bottle
     return false unless f.bottle && f.pour_bottle?
+    return false if ARGV.build_from_source? || ARGV.build_bottle?
     if f.file_modified?
       filename = f.path.to_s.gsub("#{HOMEBREW_PREFIX}/", "")
       opoo "Formula file is modified!"
@@ -39,7 +40,6 @@ module Homebrew
       puts "To fetch the bottle instead, run with --force-bottle"
       return false
     end
-    return false if ARGV.build_from_source? || ARGV.build_bottle?
     return false unless f.bottle.compatible_cellar?
     return true
   end
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -500,7 +500,7 @@ module Homebrew
       # Don't care about e.g. bottle failures for dependencies.
       run_as_not_developer do
         test "brew", "install", "--only-dependencies", *install_args unless dependencies.empty?
-        test "brew", "install", *install_args
+        test "brew", "install", "--build-from-source", *install_args
       end
       install_passed = steps.last.passed?
       audit_args = [canonical_formula_name]
--- a/Library/Formula/a2ps.rb
+++ b/Library/Formula/a2ps.rb
@@ -29,6 +29,7 @@ class A2ps < Formula
   def install
     system "./configure", "--disable-debug", "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
+    system "echo dog"
     system "make", "install"
   end
 
--- a/Library/Formula/a2ps.rb
+++ b/Library/Formula/a2ps.rb
@@ -29,7 +29,6 @@ class A2ps < Formula
   def install
     system "./configure", "--disable-debug", "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
-    system "echo dog"
     system "make", "install"
   end
 
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -180,7 +180,7 @@ class FormulaInstaller
     build_bottle_preinstall if build_bottle?
 
     unless @poured_bottle
-      if formula.file_modified?
+      if formula.file_modified? && !build_from_source?
         filename = formula.path.to_s.gsub("#{HOMEBREW_PREFIX}/", "")
         opoo "Formula file is modified!"
         puts "Building from source because #{filename} has local changes"
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -61,11 +61,10 @@ class FormulaInstaller
     return false if @pour_failed
 
     bottle = formula.bottle
-    return true  if force_bottle? && bottle
+    return true if (force_bottle? && bottle) || formula.local_bottle_path
     return false if build_from_source? || build_bottle? || interactive? || formula.file_modified?
     return false unless options.empty?
 
-    return true if formula.local_bottle_path
     return false unless bottle && formula.pour_bottle?
 
     unless bottle.compatible_cellar?


DomT4:
CC @dunn. Feel free to argue against this tweak if you want to :)

dunn:
Oh no, I was going to do this myself but forgot. Thanks!

DomT4:
Cool, Thanks. Apologies if I jumped over you on this; I did a lot of testing this morning and seeing the message threw me off a bit.

Do you have any feelings on whether we should retain the `opoo "Formula file is modified!"` part even if building from source?

We can always wrap the other two `puts` calls in the `!ARGV.build_from_source?` conditional if we wanted to still warn about modifications, or only issue the warning block if someone doesn't have `HOMEBREW_DEVELOPER` set.

dunn:
> Do you have any feelings on whether we should retain the `opoo "Formula file is modified!"` part even if building from source?

My immediate thought was it shouldn't, but now I'm starting to think that might be a good idea.  No strong feelings either way, though.

Library/Homebrew/cmd/fetch.rb
```diff
@@ -32,7 +32,7 @@ def fetch
   def fetch_bottle? f
     return true if ARGV.force_bottle? && f.bottle
     return false unless f.bottle && f.pour_bottle?
-    if f.file_modified?
+    if f.file_modified? && !ARGV.build_from_source?
```
tdsmith: I think it makes more sense to reorder this conditional with the one below it. Does that reduce the need to modify test-bot?

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -459,7 +459,7 @@ def formula formula_name
       formula_fetch_options << "--build-bottle" unless ARGV.include? "--no-bottle"
       formula_fetch_options << "--force" if ARGV.include? "--cleanup"
       formula_fetch_options << canonical_formula_name
-      test "brew", "fetch", "--retry", *formula_fetch_options
+      test "brew", "fetch", "--retry", "-fs", *formula_fetch_options
```
tdsmith: Let's spell out --from-source if we need to use it here.

Library/Homebrew/cmd/fetch.rb
DomT4: Will check this out later today. It does look like the logical option though, without having to chuck more code on the pile.

Library/Homebrew/cmd/test-bot.rb
DomT4: :+1:

mikemcquaid: :+1:

mikemcquaid:
Looks good :+1:

DomT4:
Made a few tweaks, feel free to review again.

Library/Homebrew/cmd/fetch.rb
```diff
@@ -32,6 +32,7 @@ def fetch
   def fetch_bottle? f
     return true if ARGV.force_bottle? && f.bottle
     return false unless f.bottle && f.pour_bottle?
+    return false if ARGV.build_from_source? || ARGV.build_bottle?
```
DomT4: Moving this above the `if f.file_modified?` saves us having to pass `&& !build_from_source?` here, as Tim suggested.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -480,6 +480,7 @@ def formula formula_name
       test "brew", "install", "--build-bottle", *changed_dependences unless changed_dependences.empty?
       formula_fetch_options = []
       formula_fetch_options << "--build-bottle" unless ARGV.include? "--no-bottle"
+      formula_fetch_options << "--build-from-source" unless ARGV.include? "--build-bottle"
```
DomT4: This is intended to catch the situations where people use the test-bot externally or locally and don't pass `--build-bottle`. The situation is likely to never pop up for us, but seemed a good idea to protect the eventuality and stop the situation where someone not passing `--build-bottle` gets the `modified formula` message chucked at them when they already know said formula is modified.

Library/Homebrew/cmd/fetch.rb
mikemcquaid: :+1:

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: So, `--build-bottle` isn't a valid option here. Did you mean `if ARGV.include? "--no-bottle"`? If so, use an `if`/`else` here instead.

mikemcquaid: Oh, also, this behaviour explicitly isn't what we want given https://github.com/DomT4/homebrew/blob/soapy_water/Library/Homebrew/cmd/test-bot.rb#L705-L707

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -520,7 +521,7 @@ def formula formula_name
               test "brew", "uninstall", "--force", *unchanged_build_dependencies
               unchanged_dependencies -= unchanged_build_dependencies
             end
-            test "brew", "install", bottle_filename
+            test "brew", "install", "--force-bottle", bottle_filename
```
mikemcquaid: Don't use `--force-bottle` here.

Library/Homebrew/cmd/test-bot.rb
DomT4: Yeah, I did, sorry. Was tired.

I think I'll just scrap it. If edge cases come up we can handle them once we actually know what those edge cases are, rather than trying to guess them in advance.

mikemcquaid: np!

Library/Homebrew/cmd/test-bot.rb
DomT4: I didn't know if we'd hit issues with this. Locally I get this when I modify the `ap2s` formula:

```
brew install --build-bottle -fs ap2s
...
brew rm ap2s
...
brew install /Users/Dominyk/Downloads/a2ps-4.14.yosemite.bottle.2.tar.gz

Building from source because Library/Formula/a2ps.rb has local changes
==> Downloading http://ftpmirror.gnu.org/a2ps/a2ps-4.14.tar.gz
Already downloaded: /usr/local/var/homebrew/cache/a2ps-4.14.tar.gz
==> Downloading https://trac.macports.org/export/56498/trunk/dports/print/a2ps/files/patch-contrib_sample_Makefile.in
Already downloaded: /usr/local/var/homebrew/cache/a2ps--patch-5a34c101feb00cf52199a28b1ea1bca83608cf0a1cb123e6af2d3d8992c6011f.in
==> Downloading https://trac.macports.org/export/56498/trunk/dports/print/a2ps/files/patch-lib__xstrrpl.c
Already downloaded: /usr/local/var/homebrew/cache/a2ps--patch-89fa3c95c329ec326e2e76493471a7a974c673792725059ef121e6f9efb05bf4.c
```

Does the bot handle that automatically?

mikemcquaid: I don't think it does but I think this behaviour probably shouldn't be the case when installing from a local bottle path CC @dunn and/or should only check for uncommitted changes.

dunn: I think it'd be unintuitive for `brew install` to install from source while someone's fiddling with their formula, then revert to using the bottle when they commit their changes.

But if we can't preserve that behavior without all these problems it introduced, then it can be changed to just look for uncommitted changes.

mikemcquaid: My concern is that as-is this completely changes the behaviour of the bot and, unless I'm mistaken, stops it from verifying bottle checksums any more. I'm open to any solution that deals with that such as adding a new `--build-bottle` argument to the bot.

DomT4: Welp. You are right. The only way it verifies the local checksum with `--force-bottle` is if you do something like this:

```ruby
  bottle do
    root_url "file:///Users/Dominyk/Downloads"
    revision 2
    sha256 "2c04bd2197ededcfe295a0366c0ee4c1322708e50f9cebef084460db960f78cc" => :yosemite
  end
```

```
~> brew install -v --force-bottle a2ps
==> Downloading file:///Users/Dominyk/Downloads/a2ps-4.14.yosemite.bottle.2.tar.gz
Already downloaded: /usr/local/var/homebrew/cache/a2ps-4.14.yosemite.bottle.2.tar.gz
==> Verifying a2ps-4.14.yosemite.bottle.2.tar.gz checksum
==> Pouring a2ps-4.14.yosemite.bottle.2.tar.gz
tar xf /usr/local/var/homebrew/cache/a2ps-4.14.yosemite.bottle.2.tar.gz
==> Finishing up
```

Otherwise:

```
~> brew install -v --force-bottle ./a2ps-4.14.yosemite.bottle.2.tar.gz
==> Pouring a2ps-4.14.yosemite.bottle.2.tar.gz
tar xf /Users/Dominyk/Downloads/a2ps-4.14.yosemite.bottle.2.tar.gz
```

DomT4: Just to be helpful, the test-bot seems to actually be rejecting bottles [now](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/30039/version=yosemite/console) because of the existing structure. Hmm.

dunn: Yikes. I'll try to figure out a fix, unless @DomT4 already has one in mind.

DomT4: I've got to dash out briefly, but will try and look at it when I come back. Feel free to throw around ideas though, nothing immediately comes to mind that doesn't either alter test-bot behaviour or alter the intended behaviour of the new modified formula check.

dunn: I'm still not familiar with the test bot code, but can this be fixed in `formula_installer`?
```diff
 Library/Homebrew/formula_installer.rb | 3 +--
 1 file changed, 1 insertion(+), 2 deletions(-)

diff --git a/Library/Homebrew/formula_installer.rb b/Library/Homebrew/formula_installer.rb
index a6264ee..a4787a0 100644
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -61,11 +61,10 @@ class FormulaInstaller
     return false if @pour_failed
 
     bottle = formula.bottle
-    return true  if force_bottle? && bottle
+    return true  if (force_bottle? && bottle) || formula.local_bottle_path
     return false if build_from_source? || build_bottle? || interactive? || formula.file_modified?
     return false unless options.empty?
 
-    return true if formula.local_bottle_path
     return false unless bottle && formula.pour_bottle?
 
     unless bottle.compatible_cellar?
```

mikemcquaid: That would fix the bot case, yeh. Seems like a good call :+1:

DomT4:
Pushed @dunn's change. Seems to work locally as expected/desired. Will merge it in a bit if nobody has any objections.

mikemcquaid:
@DomT4 :+1:

PR #41919, opened 2015-07-20
Add formular kroman

cheunghy:
kroman is a command line utility that converts hanguls to romanized syllables line by line.

It's lightweight, efficient, and enjoyable.

### DIFF
--- a/Library/Formula/kroman.rb
+++ b/Library/Formula/kroman.rb
@@ -1,0 +1,16 @@
+class Kroman < Formula
+  desc "Covert Korean Hangul to Romannized Syllables"
+  homepage "https://github.com/cheunghy/kroman"
+  url "https://github.com/cheunghy/kroman/archive/v1.0.zip"
+  sha1 "1b14683829a62e61b500ab0263bac814b48cf850"
+  head "https://github.com/cheunghy/kroman.git"
+
+  def install
+    system "make kroman"
+    system "make", "install", "PREFIX=#{prefix}", "MANDIR=#{man}"
+  end
+
+  test do
+    system "#{bin}/kroman", "--version"
+  end
+end


tdsmith:
Hi! I'm sorry but this project is too new for us to consider.

Some audit notes on the formula, from the bot:
 * Stable: Use GitHub tarballs rather than zipballs (url is https://github.com/cheunghy/kroman/archive/v1.0.zip).
 * Use `system "make", "kroman"` instead of `system "make kroman"`

Instead of checking that --version works, a good test would be running some hangul through your program and making sure it gives the correct romanization back.

If you'd like to make it easy for people to install your software using Homebrew, you can consider setting up a tap. We're working on some documentation; you can find it here: https://github.com/bfontaine/homebrew/blob/doc-tap/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md

Let us know, either here or on Freenode, if you have any questions or if we can help you set up a tap.

cheunghy:
Create a new git repository to place the tap. Do I understand correctly?

cheunghy:
I set a tap at https://github.com/cheunghy/homebrew-kroman.
The test function does not work well, the syntax may be incorrect.

Basically, I want this:
```
kroman < file1 > file2
file 1's content should be "I love 한국어\n"
file 2's content should be "I love han-gug-eo\n"
```
How to achieve this?

dunn:
This works for me:
```ruby
  test do
    ENV["LC_ALL"] = "en_US.UTF-8"
    assert_equal "I Love han-gug-eo", pipe_output("#{bin}/kroman", "I Love 한국어")
  end
```
`LC_ALL` just needs to be set to some UTF8 locale.

(By the way, the way you have `MANDIR` defined in your Makefile means you should pass `MANDIR=#{prefix}/man/man1` to make, although I think you might not need to specify it explicitly at all.)

cheunghy:
Thanks! It tests fine now.

By the way, I don't understand the MANDIR problem so much. Is there anyway to improve the Makefile syntax?

dunn:
`kroman.1` needs to be installed in `share/man/man1`, not directly in `share/man`. In your formula, you accidentally did the latter.

cheunghy:
Oh I see. Thanks!

PR #41928, opened 2015-07-20
fq 0.9.1

guoxiao:


Library/Formula/fq.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  depends_on "concurrencykit"
+  depends_on "jlog"
+  depends_on "openssl"
+
+  def install
+    inreplace "Makefile", "/usr/lib/dtrace", "#{lib}/dtrace"
+    system "make", "PREFIX=#{prefix}"
+    system "make", "install", "PREFIX=#{prefix}"
+    bin.install "fqc", "fq_sndr", "fq_rcvr"
```
mikemcquaid: Why aren't these installed by `make install`?

Library/Formula/fq.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  test do
+    system "fqd", "-h"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikemcquaid:
The test is failing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/29897/version=yosemite/testReport/junit/brew-test-bot/yosemite/test_fq/

Library/Formula/fq.rb
guoxiao: The `make install` only does the install job. We must run `make` first.

guoxiao:
@mikemcquaid Are there any details of the error? I've tested on my machine and it works well.

dunn:
It's possible you just need to replace `system "fqd", "-h"` with `system bin/"fqd", "-h"`.

guoxiao:
@mikemcquaid @dunn Oh I found the `fqd` is placed in `sbin` folder, that's why it's not found.

Should the `/usr/local/sbin` folder be added to `PATH` by default?

DomT4:
> Oh I found the fqd is placed in sbin folder, that's why it's not found.

`system sbin/"fqd", "-h"`, then.

> Should the /usr/local/sbin folder be added to PATH by default?

`system bin/` and `system sbin/` are shorthand for `system /usr/local/opt/formula/bin/executable` and the same for `sbin`. Nothing is tested directly from `HOMEBREW_PREFIX`.

guoxiao:
I want to find the public IPv4 Address of the machine, so I use the following
```ruby
address = Socket.ip_address_list.find { |a| a.ipv4? && !a.ipv4_loopback? }.ip_address
```
But it doesn't work in ML. What's the better way to do this?

guoxiao:
@mikemcquaid @dunn @DomT4 Any suggestions about the problem?
I'm not very familiar with ruby and I don't have a ML machine to test on.

xu-cheng:
Why you need the public IP address? Why not just `127.0.0.1`?

Library/Formula/fq.rb
```diff
@@ -0,0 +1,31 @@
...
+    inreplace "Makefile", "/usr/lib/dtrace", "#{lib}/dtrace"
+    system "make", "PREFIX=#{prefix}"
+    system "make", "install", "PREFIX=#{prefix}"
+    bin.install "fqc", "fq_sndr", "fq_rcvr"
+  end
+
+  test do
+    require "socket"
+    address = Socket.ip_address_list.find { |a| a.ipv4? && !a.ipv4_loopback? }.ip_address
+    system sbin/"fqd", "-h"
```
xu-cheng: This line can be removed. Because below test is sufficient enough.

guoxiao:
@xu-cheng Thanks for your help. I came up with a better way to solve the problem.
The `-n` option of `fqd` doesn't accept `127.0.0.1` or `0.0.0.0`, so I thought I must get the public address of machine. However just now I found I can simply omit the `-n`, and it's listening on `127.0.0.1`.

xu-cheng:

PR #41935, opened 2015-07-20
pango: Apply patch from GNOME Bug 736697

dtrebbien:
This fixes a SegFault issue when displaying certain Unicode characters.
For example, this fixes an issue where `frama-c-gui` would crash from
attempting to display the 'ℤ' character.

### DIFF
--- a/Library/Formula/pango.rb
+++ b/Library/Formula/pango.rb
@@ -1,9 +1,22 @@
 class Pango < Formula
   desc "Framework for layout and rendering of i18n text"
   homepage "http://www.pango.org/"
-  url "https://download.gnome.org/sources/pango/1.36/pango-1.36.8.tar.xz"
-  sha256 "18dbb51b8ae12bae0ab7a958e7cf3317c9acfc8a1e1103ec2f147164a0fc2d07"
-  revision 1
+  revision 2
+
+  stable do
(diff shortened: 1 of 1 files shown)


DomT4:
Has this been merged upstream already?

dtrebbien:
Yes.  See:  https://git.gnome.org/browse/pango/commit/?id=c2a40e3b35a37f76f353655d9220dd94156fb89c

It *just* missed the 1.36.8 release; it was the first commit after the tag.

DomT4:
We'll need to make sure it isn't applied to the `head` build then, otherwise the patch will fail. You'll need to create a `stable do` block like this:

```ruby
stable do
  url "https://download.gnome.org/sources/pango/1.36/pango-1.36.8.tar.xz"
  sha256 "18dbb51b8ae12bae0ab7a958e7cf3317c9acfc8a1e1103ec2f147164a0fc2d07"

  # Patch explanation
  patch do
    url "https://bug736697.bugzilla-attachments.gnome.org/attachment.cgi?id=286234"
    sha256 "xyz"
  end
end
```

dtrebbien:
New patch uploaded

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/e0fff360cb893b2e35b201c2a37d60a74c1c9f8a. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #41936, opened 2015-07-20
Bump scw

moul:
Since `scaleway-cli` do not use `godep` for building anymore, the formula is now much simpler

Tested with stable and `--HEAD`

### DIFF
--- a/Library/Formula/scw.rb
+++ b/Library/Formula/scw.rb
@@ -35,6 +35,6 @@ class Scw < Formula
 
   test do
     output = shell_output(bin/"scw version")
-    assert output.include? "OS/Arch (client): darwin/amd64"
+    assert_match "OS/Arch (client): darwin/amd64", output
   end
 end
--- a/Library/Formula/scw.rb
+++ b/Library/Formula/scw.rb
@@ -3,8 +3,8 @@ require "language/go"
 class Scw < Formula
   desc "Manage BareMetal Servers from Command Line (as easily as with Docker)"
   homepage "https://github.com/scaleway/scaleway-cli"
-  url "https://github.com/scaleway/scaleway-cli/archive/v1.2.1.tar.gz"
-  sha256 "e3ae09558a5f451935831381177f1ee3ce50aec438a1e11269dfc380e0e196c9"
+  url "https://github.com/scaleway/scaleway-cli/archive/v1.3.0.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/scw.rb
```diff
@@ -17,22 +17,6 @@ class Scw < Formula
...
-    url "https://github.com/kr/fs.git", :revision => "2788f0dbd16903de03cb8186e5c7d97b69ad387b"
-  end
-
-  go_resource "golang.org/x/tools" do
-    url "https://github.com/golang/tools.git", :revision => "473fd854f8276c0b22f17fb458aa8f1a0e2cf5f5"
-  end
-
-  go_resource "golang.org/x/crypto" do
-    url "https://github.com/golang/crypto.git", :revision => "8b27f58b78dbd60e9a26b60b0d908ea642974b6d"
-  end
```
mikemcquaid: Are these vendored now?

Library/Formula/scw.rb
```diff
@@ -42,17 +26,17 @@ def install
...
+    system "go", "build", "-o", "scw", "."
     bin.install "scw"
 
     bash_completion.install "contrib/completion/bash/scw"
     zsh_completion.install "contrib/completion/zsh/_scw"
   end
 
+  def caveats
+    "Use `scw login` to set up the correct environment to use scaleway-cli."
+  end
```
mikemcquaid: Don't need these caveats, thanks.

Library/Formula/scw.rb
moul: These packages were dependencies for the `godep` binary, the package actually have all its dependencies vendored and godep is now useless

Library/Formula/scw.rb
moul: Ok, I based this change from https://github.com/Homebrew/homebrew/blob/master/Library/Formula/s3sync.rb#L14-L16.
Reverting now

dunn:
Just a nitpick from the test bot:
```
        Error: 1 problem in 1 formula
scw:
```
Using the former makes failure messages more informative.

moul:
This bot error is not from a line I changed, I think the bot tests are different from last run

I will switch the test from assert `...include?` to `assert_match` now

moul:
Fixed, tell me if ok and if you need a rebase for the 2 commits

dunn:
I took care of the rebase and merged it in cb5be73abe8dafa84d4f55971acfed8fd4db1a7a, thanks for the PR!

PR #41945, opened 2015-07-20
Add integration tests

mikemcquaid:
We have a reasonable unit test suite but not really any integration tests (aside from `brew-test-bot`). This isn't done yet but wanted to get it up in case anyone else was working on it or had thoughts.

### DIFF
--- a/Library/Homebrew/test/lib/config.rb
+++ b/Library/Homebrew/test/lib/config.rb
@@ -1,4 +1,5 @@
 require "tmpdir"
+require "pathname"
 
 HOMEBREW_TEMP = Pathname.new(ENV["HOMEBREW_TEMP"] || Dir.tmpdir)
 
@@ -11,7 +12,7 @@ TEST_TMPDIR = ENV.fetch("HOMEBREW_TEST_TMPDIR") { |k|
 HOMEBREW_PREFIX        = Pathname.new(TEST_TMPDIR).join("prefix")
 HOMEBREW_REPOSITORY    = HOMEBREW_PREFIX
 HOMEBREW_LIBRARY       = HOMEBREW_REPOSITORY+"Library"
-HOMEBREW_LIBRARY_PATH  = File.expand_path("../../..", __FILE__)
+HOMEBREW_LIBRARY_PATH  = Pathname.new(File.expand_path("../../..", __FILE__))
 HOMEBREW_LOAD_PATH     = [File.expand_path("..", __FILE__), HOMEBREW_LIBRARY_PATH].join(":")
 HOMEBREW_CACHE         = HOMEBREW_PREFIX.parent+"cache"
 HOMEBREW_CACHE_FORMULA = HOMEBREW_PREFIX.parent+"formula_cache"
--- a/Library/Homebrew/test/test_integration_cmds.rb
+++ b/Library/Homebrew/test/test_integration_cmds.rb
@@ -1,0 +1,81 @@
+require "testing_env"
+
+class IntegrationCommandTests < Homebrew::TestCase
+  def cmd_output *args
+    cmd_args = %W[
+     -W0
+     -I#{HOMEBREW_LIBRARY_PATH}/test/lib
+     -rconfig
+    ]
+    cmd_args << "-rsimplecov" if ENV["HOMEBREW_TESTS_COVERAGE"]
+    cmd_args << (HOMEBREW_LIBRARY_PATH/"../brew.rb").resolved_path.to_s
+    cmd_args += args
+    Bundler.with_original_env do
+      ENV["HOMEBREW_BREW_FILE"] = HOMEBREW_PREFIX/"bin/brew"
+      ENV["HOMEBREW_INTEGRATION_TEST"] = args.join " "
+      ENV["HOMEBREW_TEST_TMPDIR"] = TEST_TMPDIR
+      Utils.popen_read(RUBY_PATH, *cmd_args).chomp
+    end
+  end
+
+  def cmd *args
+    output = cmd_output(*args)
+    assert_equal 0, $?.exitstatus
+    output
+  end
+
+  def cmd_fail *args
+    output = cmd_output(*args)
+    assert_equal 1, $?.exitstatus
+    output
+  end
+
+  def testball
+    "#{File.expand_path("..", __FILE__)}/testball.rb"
+  end
+
+  def test_prefix
+    assert_equal HOMEBREW_PREFIX.to_s,
+                 cmd("--prefix")
+  end
+
+  def test_version
+    assert_equal HOMEBREW_VERSION.to_s,
+                 cmd("--version")
+  end
+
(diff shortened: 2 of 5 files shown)


Library/Homebrew/test/testball.rb
```diff
@@ -2,7 +2,7 @@ class Testball < Formula
   def initialize(name="testball", path=Pathname.new(__FILE__).expand_path, spec=:stable)
     self.class.instance_eval do
       stable.url "file://#{File.expand_path("..", __FILE__)}/tarballs/testball-0.1.tbz"
-      stable.sha1 "482e737739d946b7c8cbaf127d9ee9c148b999f5"
+      stable.sha256 "1dfb13ce0f6143fe675b525fc9e168adb2215c5d59file://#{File.expand_path("..", __FILE__)}/tarballs/testball-0.1.tbz65c9f57306bb993170914f"
```
bfontaine: I think you mistakenly pasted the `file://` URL here.

xu-cheng:
Maybe it's worth notice that `brew tests` uses a different set of `PREFIX/CELLAR/LIBRARY` etc. See [here](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/test/lib/config.rb)

Library/Homebrew/test/testball.rb
mikemcquaid: Yeh, thanks :grinning:

mikemcquaid:
> Maybe it's worth notice that brew tests uses a different set of PREFIX/CELLAR/LIBRARY etc. See here

@xu-cheng Yup, knew that but thanks for the heads up.

mikemcquaid:
Any objections to this? CC @xu-cheng and @bfontaine for thoughts here as you've been playing with tests recently.

Library/Homebrew/test/lib/config.rb
```diff
@@ -11,7 +11,7 @@
 HOMEBREW_PREFIX        = Pathname.new(TEST_TMPDIR).join("prefix")
 HOMEBREW_REPOSITORY    = HOMEBREW_PREFIX
 HOMEBREW_LIBRARY       = HOMEBREW_REPOSITORY+"Library"
-HOMEBREW_LIBRARY_PATH  = File.expand_path("../../..", __FILE__)
+HOMEBREW_LIBRARY_PATH  = Pathname.new(File.expand_path("../../..", __FILE__))
```
mikemcquaid: This is a Pathname in the other `config.rb` so making this consistent seems a no-brainer.

Library/Homebrew/test/test_integration_cmds.rb
```diff
@@ -0,0 +1,76 @@
...
+  def setup
+    ENV["HOMEBREW_TEST_LIB"] = "1"
+  end
+
+  def cleanup
+    ENV.delete "HOMEBREW_TEST_LIB"
+  end
+
+  def cmd_output *args
+    Utils.popen_read("/usr/local/bin/brew", *args).chomp
```
mikemcquaid: Whoops, this is still hardcoded but I'll update it shortly before merging.

Library/brew.rb
```diff
@@ -4,16 +4,17 @@
...
 require 'pathname'
 HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent.join("Homebrew")
 $:.unshift(HOMEBREW_LIBRARY_PATH.to_s)
+$:.unshift("#{HOMEBREW_LIBRARY_PATH}/test/lib") if ENV["HOMEBREW_TEST_LIB"]
 require 'global'
 
+if ARGV == %w{--prefix}
+  puts HOMEBREW_PREFIX
+  exit 0
+end
```
mikemcquaid: Moving this below so that it handles the `/test/lib/` include correctly.

bfontaine:
:heart: for this.

Library/brew.rb
xu-cheng: I don't like the idea to move this block. `$(brew --prefix)` is widely used in all kinds of shell scripts. Therefore, the performance is a big concern here. I don't think we should sacrifice this for test.

Library/brew.rb
```diff
@@ -4,16 +4,17 @@
...
 
-if ARGV == %w{--prefix}
-  puts File.dirname(File.dirname(HOMEBREW_BREW_FILE))
-  exit 0
-end
-
 require 'pathname'
 HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent.join("Homebrew")
 $:.unshift(HOMEBREW_LIBRARY_PATH.to_s)
+$:.unshift("#{HOMEBREW_LIBRARY_PATH}/test/lib") if ENV["HOMEBREW_TEST_LIB"]
```
xu-cheng: I kinda want to avoid injecting test code here, so there won't be case when `HOMEBREW_TEST_LIB` is accidentally set.

In fact we can inject test config when invoking brew command, like this: (If you don't think this it too hacky)
```ruby
  def cmd_output *args
    Utils.popen_read(RUBY_PATH,
                     "-W0",
                     "-I#{HOMEBREW_LIBRARY_PATH}/test/lib",
                     "-rpathname",
                     "-rconfig",
                     "#{(HOMEBREW_LIBRARY_PATH/"../brew.rb").resolved_path}",
                     *args).chomp
  end
```

Library/brew.rb
mikemcquaid: I'll benchmark the performance and see how it changes. It concerns me that as-is it's using different logic to calculate the prefix which means in some weird circumstances it could produce a different result.

Library/brew.rb
mikemcquaid: It's no longer really an integration test at that point, unfortunately, because it's no longer calling the `HOMEBREW_PREFIX/bin/brew` script. I'm definitely open to other approaches here, though.

xu-cheng:
Another thing worth to consideration is how to make these tests support coverage. Maybe move the coverage config code into `test/lib/coverage.rb`? In this way, we can load it using command line in the approach I purposed above.

Library/brew.rb
mikemcquaid: @xu-cheng Thoughts?

mikemcquaid:
@xu-cheng Do they not support coverage currently because they shell out?

mikemcquaid:
> Another thing worth to consideration is how to make these tests support coverage. Maybe move the coverage config code into test/lib/coverage.rb? In this way, we can load it using command line in the approach I purposed above.

I believe it is generating coverage information. It just doesn't seem to cover the `cmd` directory or `brew.rb` (and I'm not quite sure why).

mikemcquaid:
@xu-cheng Updated (with stuff from your branch). Let me know what you think (and @bfontaine)

xu-cheng:
:+1:

bfontaine:
I’m too late but :+1:  :smile:

PR #41952, opened 2015-07-21
Use exec instead of system in fork block

guoxiao:
Use exec instead of system to avoid forking, because we need to kill it via pid afterwards.


dunn:
The test failures are from a resource that's 404ing. I think I added it, so I'll try to find a replacement.

xu-cheng:
This PR need to be rebase.

guoxiao:
@xu-cheng done.

xu-cheng:

PR #41956, opened 2015-07-21
test_formula_installer: fix log file leak

xu-cheng:


xu-cheng:
Currently, file leak test only run [when `LOG_FS_LEAKS ` is set](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/test/testing_env.rb#L81). Should we make it run unconditionally?

mikemcquaid:
@xu-cheng Seems like a good idea.

xu-cheng:
Updated to always check file leak.

mikemcquaid:
:+1:

PR #41960, opened 2015-07-21
debianutils: fix stale url, version bump to 4.5.1

smoofra:
The urls for 4.5 are returning 404 now.

Signed-off-by: Lawrence D'Anna <larry@elder-gods.org>

### DIFF
--- a/Library/Formula/debianutils.rb
+++ b/Library/Formula/debianutils.rb
@@ -1,9 +1,9 @@
 class Debianutils < Formula
   desc "Miscellaneous utilities specific to Debian"
   homepage "https://packages.debian.org/unstable/utils/debianutils"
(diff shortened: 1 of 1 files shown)


dunn:
Could you keep using the `https` mirrors, though? Thanks!

smoofra:
The debian server doesn't support https.   I guess that's why those mirrors were used instead?

dunn:
https://mirrors.kernel.org/debian/pool/main/d/debianutils/debianutils_4.5.1.tar.xz and https://mirrors.ocf.berkeley.edu/debian/pool/main/d/debianutils/debianutils_4.5.1.tar.xz work for me.

DomT4:
^ :+1:. Let's stay secure here.

dunn:
Merged in cce3b69bc292118334a6f0067d5f29bf3ba4d37a, thanks again!

PR #41965, opened 2015-07-21
osquery: fix runtime dependencies.

AntonioMeireles:
per ...
>  otool -L /usr/local/bin/osqueryd | grep gfla                                                                                                                                                      
>  /usr/local/lib/libgflags.2.dylib (compatibility version 2.0.0, current version 2.1.2)

... gflags is actually a runtime dependency too.

### DIFF
--- a/Library/Formula/osquery.rb
+++ b/Library/Formula/osquery.rb
@@ -5,6 +5,7 @@ class Osquery < Formula
   homepage "https://osquery.io"
   # pull from git tag to get submodules
   url "https://github.com/facebook/osquery.git", :tag => "1.5.0", :revision => "ca09fdb9f80ed632b98a2a9c41a521309e14b611"
+  revision 1
 
   bottle do
     sha256 "4eb42ad401dc6ae671b435c15683272150b53d31890a20ca4cdcabe2715cfd30" => :yosemite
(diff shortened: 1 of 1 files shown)


AntonioMeireles:
//cc @theopolis @marpaia @mistydemeo @sharvilshah

theopolis:
Ah good catch, as of https://github.com/Homebrew/homebrew/pull/41151, gflags is only installs its dynamic library.

DomT4:
Since there's breakage here let's add a `revision` as well so people get the update now.

AntonioMeireles:
@DomT4 done && pushed.

Library/Formula/osquery.rb
```diff
@@ -3,6 +3,7 @@
 class Osquery < Formula
   desc "SQL powered operating system instrumentation and analytics"
   homepage "https://osquery.io"
+  revision 1
```
DomT4: Should go below the `url` (It's always the last line of the initial block). If a local `brew audit osquery` doesn't flag it, don't worry about it, I can fix it on pull.

AntonioMeireles: fixed in https://github.com/Homebrew/homebrew/commit/26d4f6e9253fd6c81c03ed40b6e30634331a49ac.

DomT4: Thanks!

sharvilshah:
Thanks for catching this @AntonioMeireles @DomT4

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/66626e1a4d048bf7e66003e12bb26241cd98d4e9. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #41967, opened 2015-07-21
acpica 20150717

Hs-Yeah:
Update acpica to 20150717


xu-cheng:

Issue #41955, opened 2015-07-21
It is impossible to install msmtp with ntlm support


cweagans:
I have installed the `gsasl` package and edited the msmtp formula to include the `--with-libgsasl` argument on the configure command. When the configure command was run as part of `brew install msmtp --build-from-source`, this was part of the output:

```
Install prefix ......... : /usr/local/Cellar/msmtp/1.6.2
TLS/SSL support ........ : yes (Library: OpenSSL)
GNU SASL support ....... : no
IDN support ............ : no
NLS support ............ : no
Libsecret support (GNOME): no
MacOS X Keychain support : yes
```

SASL support, in this case, should be "yes".

Normally, I wouldn't have opened a homebrew bug about this, but when I manually download the msmtp source from sourceforge and run `./configure --with-libgsasl` (Again with the homebrew package `gsasl` already installed), I get this:

```
Install prefix ......... : /usr/local
TLS/SSL support ........ : yes (Library: GnuTLS)
GNU SASL support ....... : yes
IDN support ............ : no
NLS support ............ : no
Libsecret support (GNOME): no
MacOS X Keychain support : yes
```

I'm guessing this has something to do with what directories homebrew tells `./configure` to search for libraries (especially since the Homebrew version uses openssl and the manual version uses GnuTLS).

Is there any way around this issue? I'm not overly familiar with how make figures out library locations...

cweagans:
(Oh, btw - it's probably a good idea to make msmtp depend on gsasl and have the `--with-libgsasl` flag always turned on if we can get this figured out. gsasl is a pretty solid library and there's no harm in not using it, and having it available makes msmtp more useful out of the box for many people, namely those who are stuck with an Exchange server that only accepts NTLM auth)

bfontaine:
Please post your edited formula so we can see what’s missing.

tzudot:
@cweagans make these changes, as an additional option to support `gsasl` while Homebrew compiles `msmtp`.
 
```diff
diff --git a/Library/Formula/msmtp.rb b/Library/Formula/msmtp.rb
index d8984c5..8616086 100644
--- a/Library/Formula/msmtp.rb
+++ b/Library/Formula/msmtp.rb
@@ -12,6 +12,9 @@ class Msmtp < Formula

   depends_on "pkg-config" => :build
   depends_on "openssl"
+  depends_on "gsasl" => :optional
+
+  option "with-gsasl"

   def install
     args = %W[
[...]

PR #41969, opened 2015-07-21
msmtp: add --with-gsasl option.

tzudot:
~~--with-gnutls
        Build with gnutls support~~
--with-gsasl
        Use GNU SASL authentication library
~~--with-libidn
        Build with libidn support
--with-libsecret
        Build with lib secret support~~

### DIFF
--- a/Library/Formula/msmtp.rb
+++ b/Library/Formula/msmtp.rb
@@ -10,8 +10,11 @@ class Msmtp < Formula
     sha256 "6e0b109c6d290ce71bc24302b0e4fd0a45a1f1b9db025afde8273e87e99777de" => :mountain_lion
   end
 
+  option "with-gsasl", "Use GNU SASL authentication library"
+
   depends_on "pkg-config" => :build
   depends_on "openssl"
+  depends_on "gsasl" => :optional
 
   def install
     args = %W[
@@ -20,6 +23,8 @@ class Msmtp < Formula
       --prefix=#{prefix}
       --with-tls=openssl
     ]
+
+    args << "--with-libsasl" if build.with? "gsasl"
 
     system "./configure", *args
     system "make", "install"


tzudot:
Fixes https://github.com/Homebrew/homebrew/issues/41955. 

@bfontaine Need to decide on wether to make `gsasl` as default compile option as `:recommended`.

Library/Formula/msmtp.rb
```diff
@@ -12,6 +12,15 @@ class Msmtp < Formula
...
   depends_on "openssl"
+  depends_on "gnutls" => :optional
+  depends_on "gsasl" => :optional
+  depends_on "libidn" => :optional
+  depends_on "libsecret" => :optional
+
+  option "with-gsasl", "Use GNU SASL authentication library"
+  option "with-libidn", "Build with support for Internationalized Domain Names"
+  option "with-libsecret"
+  option "with-gnutls", "Use GNU Transport Layer Security library"
```
bfontaine: These options are auto-generated by `depends_on "…" => :optional`. You can remove them unless you want a custom message.

bfontaine:
Are these options used in practice? We try to avoid bloating formulae with a lot of options.

Library/Formula/msmtp.rb
mikemcquaid: i.e. do you use all these yourself? Is there any reason to not enable any of these by default?

tzudot: 1. It would be useful to have SASL enabled by default. 
2. Not sure about `libidn` but is present in Homebrew.
2. As per `msmtp` docs `gnutls` is preferred TLS method. 
3. `libsecret` is already available in Homebrew, and is useful.

tzudot:
SASL is used in smtp clients for security, we should enable it by default. Let me know about others options as well.

tzudot:
@cweagans care to add your comments here?

chdiza:
The `gsasl` shouldn't be in here, unless it's `optional`.  It's not at all necessary for securely connecting or authenticating with smtp.

tzudot:
Compiling `msmtp` with `gsasl` doesn't force us to use the `gsasl` features unless we configure in `.msmtprc`. By default, the authentication library used is the built-in one but `gsasl` is a more solid library to use as it supports many other features.

chdiza:
> Compiling msmtp with gsasl doesn't force us to use the gsasl features unless we configure in .msmtprc

But it does force us to download and build `gsasl`, which should not happen unless the user actually wants it.

cweagans:
Can we just use gsasl if it's present (or if the `--with-gsasl` flag is passed to brew)?

gsasl provides a superset of authentication methods that msmtp supplies out of the box and *replaces* the msmtp built-in authentication library, so there's a chance that some configurations may not work the same after msmtp is upgraded (since a different auth lib is used). It's probably not a good idea to require this by default.

That said, there's also a case for using it by default, which is that every Exchange server currently deployed supports NTLM authentication. It's bad, but it works. If you want to send mail through Exchange without bugging your sysadmins to enable other auth methods, this is the way to do it.

I don't particularly care about whether or not it's on by default - just that it's available.

btw, thanks @tzudot for this PR. I opened an issue, went to sleep, and then woke up to my problem being solved! Thanks! :)

tzudot:
I would be adding just the `--with-gsasl` flag, then.

cweagans:
:+1: This works for me.

cweagans:
Actually, I might need the gnutls option too. Don't merge this just yet. I'm experimenting a bit and will report back shortly.

cweagans:
The gnutls option is useful if you need to specify a lower minimum diffie-hellman prime size, but other than that, openssl works well. I say leave it out if nobody else needs it.

DomT4:
Happy enough with the potential use case & status of this dependency here. Optional is fine.

Merged in https://github.com/Homebrew/homebrew/commit/eac4b0a603a840a59fdbbb3ac5977f4bf5375afa. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #41970, opened 2015-07-21
fuseki 2.3.0

depressiveRobot:
* removed LaunchAgent functionality

### DIFF
--- a/Library/Formula/fuseki.rb
+++ b/Library/Formula/fuseki.rb
@@ -2,10 +2,9 @@ require 'formula'
 
 class Fuseki < Formula
   desc "SPARQL server"
-  homepage 'https://jena.apache.org/documentation/serving_data/'
-  url 'https://www.apache.org/dist/jena/binaries/jena-fuseki1-1.1.2-distribution.tar.gz'
-  version '1.1.2'
-  sha256 '78bd92b4e32f9e918d89946d11aed9789416f4058b127af60b251b4a8636b5f0'
+  homepage 'https://jena.apache.org/documentation/fuseki2/index.html'
+  url 'https://www.apache.org/dist/jena/binaries/apache-jena-fuseki-2.3.0.tar.gz'
+  sha256 'b37a81d7b4455ccc07055e50dc13c482c9e2a729e329eec6d2d5d51fc2e724ee'
 
   def install
     # Remove windows files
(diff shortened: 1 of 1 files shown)


Library/Formula/fuseki.rb
```diff
@@ -22,67 +21,33 @@ def install
...
-      uses traditional Unix paths: please inspect the settings here first:
-      #{etc}/fuseki.ttl
+    * This version does not include the LaunchAgent version anymore!
 
-    * NOTE: Fuseki uses 1) log4j.configuration if defined, 2) 'log4j.properties' file if exists,
-      or built-in configuration. See also:
-      https://issues.apache.org/jira/browse/JENA-536
-    EOS
-  end
+    * You may want to set the FUSEKI_BASE environment variable which points to the runtime area for the server instance (defaults to 'run/' of the current directory).
```
mikemcquaid: We should set that for the user instead using a wrapper script.

mikemcquaid:
Why does the `plist` not work any more?

Library/Formula/fuseki.rb
depressiveRobot: I do not understand what you mean! Can you please explain it with more detail.

mikemcquaid: We could write a wrapper Bash script which sets the `FUSEKI_BASE` environment variable for the install.

Library/Formula/fuseki.rb
```diff
@@ -22,43 +21,32 @@ def install
...
-    etc.install 'config.ttl' => 'fuseki.ttl'
+    libexec.install 'fuseki-server.jar', 'webapp'
 
     # Create a location for dataset and log files,
     # in case we're being used in LaunchAgent mode
     (var/'fuseki').mkpath
     (var/'log/fuseki').mkpath
 
-    # Install example configs
-    prefix.install 'config-examples.ttl', 'config-inf-tdb.ttl', 'config-tdb-text.ttl', 'config-tdb.ttl'
```
dunn: Do these files no longer exist, or are they just not worth installing?

depressiveRobot: They do no longer exist.

Library/Formula/fuseki.rb
```diff
@@ -2,10 +2,9 @@
 
 class Fuseki < Formula
   desc "SPARQL server"
-  homepage 'https://jena.apache.org/documentation/serving_data/'
-  url 'https://www.apache.org/dist/jena/binaries/jena-fuseki1-1.1.2-distribution.tar.gz'
-  version '1.1.2'
-  sha256 '78bd92b4e32f9e918d89946d11aed9789416f4058b127af60b251b4a8636b5f0'
+  homepage 'https://jena.apache.org/documentation/fuseki2/index.html'
+  url 'http://mirror.arcor-online.net/www.apache.org/jena/binaries/apache-jena-fuseki-2.0.0.tar.gz'
```
dunn: Are there any `https` mirrors we could use instead?

depressiveRobot: Found some, used now.

Library/Formula/fuseki.rb
```diff
@@ -73,14 +61,21 @@ def plist; <<-EOS.undent
...
+        <dict>
+           <key>FUSEKI_BASE</key>
+           <string>#{var}/fuseki</string>
+        </dict>
         <key>ProgramArguments</key>
         <array>
-          <string>#{opt_bin}/fuseki-server</string>
-          <string>--config</string>
-          <string>/usr/local/etc/fuseki.ttl</string>
+          <string>#{bin}/fuseki-server</string>
```
DomT4: This should remain `#{opt_bin}` so the `plist` doesn't have to be replaced by the user every update.

dunn:
Looks like this needs to be rebased now.

DomT4:
Still needs rebasing, apparently.

PR #41983, opened 2015-07-21
mr 1.20150503

L2G:
* New upstream version
* Add really basic test
* Other changes for `brew audit --strict`

### DIFF
--- a/Library/Formula/mr.rb
+++ b/Library/Formula/mr.rb
@@ -1,14 +1,23 @@
-require 'formula'
-
 class Mr < Formula
   desc "Multiple Repository management tool"
-  homepage 'http://myrepos.branchable.com/'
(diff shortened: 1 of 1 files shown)


dunn:
How about this for a test?
```ruby
  test do
    system "git", "clone", "https://github.com/Homebrew/homebrew-command-not-found.git"
    cd "homebrew-command-not-found" do
      system bin/"mr", "register"
    end
    File.exist? ".mrconfig"
    system bin/"mr", "log"
  end
```

DomT4:
If we want to use `homebrew-command-not-found` let's add it as a `resource` rather than a `git clone`. Allows us to cache it.

dunn:
Merged in 034596c6a68eddbb27cb6e4e73d534a7fd5327d3, thanks again!

PR #41986, opened 2015-07-21
Add optional libntlm dependency to gsasl

cweagans:
If libntlm is installed, I should be able to build gsasl with the ntlm functionality enabled. This, along with #41969 enables NTLM authentication in msmtp.

### DIFF
--- a/Library/Formula/gsasl.rb
+++ b/Library/Formula/gsasl.rb
@@ -5,6 +5,8 @@ class Gsasl < Formula
   mirror "https://ftp.gnu.org/gsasl/gsasl-1.8.0.tar.gz"
   sha256 "310262d1ded082d1ceefc52d6dad265c1decae8d84e12b5947d9b1dd193191e5"
 
+  depends_on "libntlm" => :optional
+
   bottle do
     cellar :any
     revision 2


DomT4:
No configure options required?

cweagans:
Nope. If libntlm is available, it'll be used.

DomT4:
Can we make it so that if it is available we *don't* use it unless the user specifies they want it to be used? We've been bitten many many times by things silently picking up available libraries in the environment, those libraries changing/etc, and the tool breaking over a silent dependency.

bfontaine:
Ping?

cweagans:
It's not really something that's decided by this formula. There's no configure option for gsasl to turn ntlm support on or off - it "just works" if libntlm is present in the environment.

DomT4:
That's kind of irritating. Involuntarily picking up libraries that happen to be available is one of the bigger causes of breakages :cry:. I guess if it's already happening there's no harm in at least making that a little more obvious, but I'd appreciate it if someone wandered upstream and asked them to give us a way to flip the off-switch when users don't request it.

cweagans:
I think in this case it's okay. The amount of code in gsasl to support NTLM is very very very minimal (as all of the functionality is bundled with libntlm).

Another option is to just to outright require it on our end. libntlm is really small.

DomT4:
Unless people actively start demanding it or it is in some way critical to functionality we'd rather not enforce a new dependency on everyone by default.

Merged in https://github.com/Homebrew/homebrew/commit/2ac5dec428d4479c187a88ab718789cd0a077b00. Thank you for your contribution to Homebrew, we appreciate it!

PR #41987, opened 2015-07-21
Upstream moved tarball

QuinnyPig:
Whoops. Nothing else changed, sha256 still matches.


dunn:
Sorry, I opened #41992 before I saw this. Thanks for following up, though!

PR #41994, opened 2015-07-22
cyclops 0.1.0

nextrevision:
Adds a formula for [cyclops](https://github.com/thisendout/cyclops), a docker-powered REPL for scripting and automation.

### DIFF
--- a/Library/Formula/cyclops.rb
+++ b/Library/Formula/cyclops.rb
@@ -1,0 +1,48 @@
+require "language/go"
+
+class Cyclops < Formula
+  desc "Docker-powered REPL for scripting and automation"
+  homepage "https://github.com/thisendout/cyclops"
+  url "https://github.com/thisendout/cyclops/archive/v0.1.0.tar.gz"
+  sha256 "93e636c67f644917410d8df04c5c46fb21e9ac69b9fe4ebbdfdda00010d0cdf8"
+
+  depends_on "go" => :build
+
+  go_resource project do
+    url "https://github.com/mitchellh/gox.git", :revision => 'a5a468f84d74eb51ece602cb113edeb37167912f'
+  end
+
+  go_resource project do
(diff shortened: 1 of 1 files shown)


dunn:
We've got a test failure:
```
        Testing cyclops
Error: cyclops: failed
<"cannot connect to Docker endpoint\n"> expected but was
<"DOCKER_HOST must be set\n">.
```
This formula might also be a bit too new for Homebrew, we want to make sure they're not a flash in the pan before we take them on.

Library/Formula/cyclops.rb
```diff
@@ -0,0 +1,51 @@
...
+  cyclops_deps = %w[
+    github.com/mitchellh/gox a5a468f84d74eb51ece602cb113edeb37167912f
+    github.com/mitchellh/iochan 87b45ffd0e9581375c491fef3d32130bb15c5bd7
+  ]
+
+  cyclops_deps.each_slice(2) do |project, revision|
+    go_resource project do
+      url "https://#{project}.git", :revision => revision
+    end
+  end
```
bfontaine: Since there are only two resources, isn’t it simpler to just write two `go_resource` instead of using a loop here?

Library/Formula/cyclops.rb
```diff
@@ -0,0 +1,51 @@
...
+      system "gox", "-arch", arch,
+        "-os", "darwin",
+        "-output", "bin/cyclops-{{.Dir}}",
+        "./..."
+      bin.install "bin/cyclops-cyclops" => "cyclops"
+    end
+  end
+
+  test do
+    assert_equal "cannot connect to Docker endpoint\n", `"#{bin}/cyclops"`
```
bfontaine: Please use `shell_output("#{bin}/cyclops")` instead of `` `"#{bin}/cyclops"` ``.

PR #41995, opened 2015-07-22
libmaxminddb 1.1.1

oschwald:


Library/Formula/libmaxminddb.rb
```diff
@@ -1,8 +1,8 @@
 class Libmaxminddb < Formula
   desc "C library for the MaxMind DB file format"
   homepage "https://github.com/maxmind/libmaxminddb"
-  url "https://github.com/maxmind/libmaxminddb/releases/download/1.0.4/libmaxminddb-1.0.4.tar.gz"
-  sha1 "57548d426d43b9b43c77786b08594d48d0c88c62"
+  url "https://github.com/maxmind/libmaxminddb/releases/download/1.1.1/libmaxminddb-1.1.1.tar.gz"
+  sha1 "5ab295438ad914e704d69e3636fc2d951ab97e27"
```
dunn: We've deprecated the use of `sha1`, could you upgrade this to the `sha256`? Thanks!

oschwald: Done.

dunn:
Thanks! One more change, if you could: let's make the testfile into a resource so we can check its `sha256` as well, and also cache it.  Add a resource block above the install section:
```ruby
  resource "mmdb" do
    url "http://geolite.maxmind.com/download/geoip/database/GeoLite2-Country.mmdb.gz"
    sha256 "4819459376f61423f5fb96eff141bf486d8e6133c47acd832d2ba87a39f2a922"
  end
```
Then update the test block:
```ruby
  test do
    resource("mmdb").stage do
      system "#{bin}/mmdblookup", "-f", "GeoLite2-Country.mmdb",
                                  "-i", "8.8.8.8"
    end
  end
```

oschwald:
That file changes monthly. I didn't write that part of the formula, but my counter suggestion would be to remove the `test` section completely. The library includes a comprehensive test suite. If the `make check` works, it is unlikely that the build failed.

oschwald:
I went ahead and removed that test.

DomT4:
The problem is we use our test block for more than just compile. If `geoipupdate` was ever made a mandatory dependency, every time it was updated we'd pour `libmaxminddb`'s bottle and use our `test do` block to assert there were no problems against the dependency. We can't do that with the block removed.

oschwald:
Ah, I didn't realize that. I can change it to just do the same looked on a test database that is included with libmaxminddb.

Library/Formula/libmaxminddb.rb
```diff
@@ -44,9 +37,7 @@ def install
   end
 
   test do
-    system "curl", "-O", "http://geolite.maxmind.com/download/geoip/database/GeoLite2-Country.mmdb.gz"
-    system "gunzip", "GeoLite2-Country.mmdb.gz"
-    system "#{bin}/mmdblookup", "-f", "GeoLite2-Country.mmdb",
-                                "-i", "8.8.8.8"
+    system "#{bin}/mmdblookup", "-f", "#{buildpath}/t/maxmind-db/test-data/GeoIP2-City-Test.mmdb",
```
DomT4: You'll need to install this somewhere. The `buildpath` is only present during compile, the `testpath` is only present during the test. The two are sandboxed from each other.

Something like `(share/"examples").install buildpath/"t/maxmind-db/test-data/GeoIP2-City-Test.mmdb"` should be fine, at the end of `def install`.

oschwald: Thanks! I did this.

oschwald:
Ok, I think this is finally good to go. Thanks for all the review!

dunn:
Looks great, thanks again! Merged in c39587140e81628cb87867580a460abcdfda568b.

PR #42002, opened 2015-07-22
wellington: add alias: wt

drewwells:
Signed-off-by: Drew Wells <drew.wells00@gmail.com>

### DIFF
--- a/Library/Aliases/wt
+++ b/Library/Aliases/wt
@@ -1,0 +1,1 @@
+../Formula/wellington.rb


mikemcquaid:
I'd probably rather not add this to avoid future clashing with e.g. http://www.webtoolkit.eu/wt/doc/reference/html/InstallationUnix.html

DomT4:
Passing on this, Sorry!

PR #42009, opened 2015-07-22
test-bot: run brew tests in no-compat mode

xu-cheng:


mikemcquaid:
:+1:

mikemcquaid:
Actually, the only thing I wonder is if it would somehow be possible to run this in a single `brew tests` run and have it be smart enough to not rerun tests unnecessary. That may well be impossible though so ignore me if it is :grinning:

PR #42010, opened 2015-07-22
boot2docker: add caveats for the VM upgrade

bfontaine:
We don’t handle the VM download/upgrade; if an user upgrades `boot2docker` without these commands they’ll get an error like `Error response from daemon: client and server don't have same version (client : 1.19, server: 1.18)`.

### DIFF
--- a/Library/Formula/boot2docker.rb
+++ b/Library/Formula/boot2docker.rb
@@ -3,8 +3,8 @@ class Boot2docker < Formula
   homepage "https://github.com/boot2docker/boot2docker-cli"
   # Boot2docker and docker are generally updated at the same time.
   # Please update the version of docker too
-  url "https://github.com/boot2docker/boot2docker-cli.git", :tag => "v1.7.1",
-    :revision => "8fdc6f573bf08149b6311681800d55fda6e19e71"
+  url "https://github.com/boot2docker/boot2docker-cli.git",
+    :tag => "v1.7.1", :revision => "8fdc6f573bf08149b6311681800d55fda6e19e71"
   head "https://github.com/boot2docker/boot2docker-cli.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Generally we wouldn't add stuff like this that isn't Homebrew-specific. I'm happy to defer to you if you think it's really important, though.

bfontaine:
I think it is because without these the VM breaks at each `brew upgrade` and it makes sysadmins like @jbdalido angry because they get pinged by everyone saying Docker is broken :wink:

mikemcquaid:
Ok, cool.

xu-cheng:
any reason we don't do these for users in `post_install`?

bfontaine:
@xu-cheng if you have something working on the VM, using these commands will destroy it, so it’s probably better to let users run them.

xu-cheng:
:cool:

DomT4:
Indentation seems a bit weird in the `caveats`, but agree it's likely to reduce the level of issues we dance around on this.

bfontaine:
@DomT4 How weird? I used four spaces to match the plist auto-generated caveats.

DomT4:
Caveats usually use a two space separation, I believe, but indeed the `plist` ones use 4. I'm more inclined to think the `plist` printing is wrong, but ¯\_(ツ)_/¯.

bfontaine:
Yeah, the 4-spaces indentation bugs me too.

DomT4:
Have asked if anyone objects to changing the `plist` caveats from 4 spaces to 2 :smile_cat:

bfontaine:
I’ll wait for this decision :smile:

mikemcquaid:
> Have asked if anyone objects to changing the plist caveats from 4 spaces to 2

Do it!

DomT4:
https://github.com/Homebrew/homebrew/commit/3ffa444831c8253d0cf80a1432a229705f2d3d68

PR #42013, opened 2015-07-22
cracklib 2.9.5

nijikon:


nijikon:
I would really appreciate some feedback on this /cc @xu-cheng @bfontaine @DomT4

mikemcquaid:
This makes sense to me :+1:

DomT4:
Definitely :+1:

Library/Formula/cracklib.rb
```diff
@@ -22,5 +25,17 @@ def install
...
                           "--with-default-dict=#{HOMEBREW_PREFIX}/share/cracklib-words"
     system "make", "install"
+
+    resource("cracklib-words").stage do |r|
+      share.install "cracklib-words-#{r.version}" => "cracklib-words"
+    end
+  end
+
+  def post_install
+    system "#{bin}/cracklib-packer < #{share}/cracklib-words"
```
bfontaine: You can probably do `pipe_output("#{bin}/cracklib-packer", (share/"cracklib-words").read, 0)`.

Library/Formula/cracklib.rb
```diff
@@ -22,5 +25,17 @@ def install
...
+  end
+
+  test do
+    assert_match /password: it is based on a dictionary word/, shell_output("echo 'password' | #{bin}/cracklib-check")
```
bfontaine: You can use `pipe_output("#{bin}/cracklib-check", "password", 0)`.

Library/Formula/cracklib.rb
dunn: Out of curiosity, why does that need to be in `post_install`?

DomT4: It actually probably doesn't need to be any more, although it is the sort of thing we don't handle directly during install; if nobody else really wants to change it I don't mind where it is now.

mikemcquaid: Yeh, putting in install probably makes sense.

nijikon: ```pipe_output``` is not available in install section.

Moving code from post_install to install breaks the build.

```
==> Verifying cracklib--cracklib-words-20080507.gz checksum
==> /usr/local/Cellar/cracklib/2.9.5/bin/cracklib-packer < /usr/local/Cellar/cracklib/2.9.5/share/cracklib-words
/usr/local/share/cracklib-words.pwd: Operation not permitted
/usr/local/share/cracklib-words: Operation not permitted
/usr/local/Library/Homebrew/debrew.rb:11:in `raise'
```

xu-cheng: I have test it under the sandbox. This should be put in `post_install`.

xu-cheng:
Thanks @nijikon

PR #42020, opened 2015-07-22
openssl: Use native openssl in post_install()

docwhat:
This is faster and allows extra checking in the future if we want.

We can also pretty up the `certs.pem`.

I'll submit some more changes to add the Subject into the `cert.pem` so
you can tell if a specific certificate is there or not.

How would you guys feel about adding the work around for the jruby-openssl bug
number 56?

https://github.com/jruby/jruby-openssl/issues/56#issuecomment-123474865

### DIFF
--- a/Library/Formula/openssl.rb
+++ b/Library/Formula/openssl.rb
@@ -109,22 +109,17 @@ class Openssl < Formula
       /System/Library/Keychains/SystemRootCertificates.keychain
     ]
 
-    certs_list = `security find-certificate -a -p #{keychains.join(" ")}`
-    certs = certs_list.scan(
-      /-----BEGIN CERTIFICATE-----.*?-----END CERTIFICATE-----/m
-    )
-
-    valid_certs = certs.select do |cert|
-      IO.popen("openssl x509 -inform pem -checkend 0 -noout", "w") do |openssl_io|
-        openssl_io.write(cert)
-        openssl_io.close_write
-      end
-
-      $?.success?
-    end
+    # We filter out:
+    # * Not yet valid certificates
+    # * Expired certificates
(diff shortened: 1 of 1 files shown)


DomT4:
I'm not sure writing it this way makes it any easier for the "casual" contributor to understand what on earth we're doing, which is something I worry about.

I'm fairly against removing certificates which are not yet valid. It's not uncommon for people to not regenerate their PEM file outside of Homebrew's OpenSSL updates, which barring security problems upstream or changes downstream don't come around massively often.

If a cert comes online during that time and starts being relied on, people are going to see difficult to debug failure. Expecting people to periodically do `brew postinstall openssl` to regenerate that PEM isn't great UX, and the future ban could do that.

docwhat:
Well it's easy to remove that `reject`. 

I don't like the current UX either. If you have a suggestion, I'd happily implement it. E.g. Moving it to an external command or something. 

I feel like the pure ruby version is easier to understand because the `openssl` command line flags are as opaque as mud.

mikemcquaid:
> This is faster

How much faster?

DomT4:
> I don't like the current UX either. If you have a suggestion, I'd happily implement it.

There's not a better one really. Either people do `brew postinstall openssl` every so often, or they replace the PEM every so often manually. We replace it every time OpenSSL is updated, but I'm not going to start automatically regenerating the PEM outside of that cycle. Drop the kill "Not yet valid" certs part and I'm happy with the status quo.

I presume the current failure here is the need to add `require "openssl"` to the formula. It probably isn't naturally invoked by Ruby.

DomT4:
Going to pass on this for now. Thanks for the submission though!

Issue #42025, opened 2015-07-22
httpry depends on bsdmake, why?


vmsp:
I just tried compiling httpry with GNU make (xcode) and it worked fine.
Am I missing something?

Thank you.

mistydemeo:
It looks like bsdmake has been used since the first version in e7dd57f846852592be7fbd50449797e22c788ff0, but the reason for using bsdmake isn't documented there or in #6038. It may no longer be necessary.

DomT4:
Because the dependency was added and probably last checked ~3 years ago :wink:. Feel free to submit a PR to remove the dependency. If it builds, we'd merge it.

DomT4:
Oops, Apologies for the double maintainer post.

mistydemeo:
Fixed, thanks!

PR #42029, opened 2015-07-22
httpry, remove unnecessary dependency

vmsp:
Removes dependency on bsdmake. GNU make works fine.
Solves #42025.


vmsp:
Ups, already solved by #42027.

PR #42030, opened 2015-07-22
Update release url for ponyc source code

royfielding:
It appears that the ponyc release archive has been moved to a separate host in the same domain. The filename and hashes remain the same.


dunn:
Merged in 8fad723d1bf173ff7661c5083c9cc31b1a8ca848, thanks for catching that!

Issue #41690, opened 2015-07-14
El Capitan Update Error


vinkla:
Maybe this is to early to submit El Capitan errors but here we go.

When running `brew update` and the `brew upgrade --all` it returns the following error:

```bash
Error: Failure while executing: git pull --quiet origin refs/heads/master:refs/remotes/origin/master
```

```bash
Warning: You have uncommitted modifications to Homebrew
If this a surprise to you, then you should stash these modifications.
Stashing returns Homebrew to a pristine state but can be undone
should you later need to do so for some reason.
    cd /usr/local/Library && git stash && git clean -d -
```

Which is strange since I haven't touched those files. When running a hard reset I've no permission reset the repository files:

```bash
error: unable to unlink old '.gitignore' (Permission denied)
fatal: Could not reset index file to revision 'origin/master'.
```

Doing a reset to HEAD works fine but it doesn't get me any updates.

This is all very strange since it worked just fine yesterday. I'm using El Capitan public beta 15A215h and Homebrew 0.9.5. This is similar issues as in #41150 and #41665.

vinkla:
Solved it by running `sudo chown $(USER):admin /usr/local`.

ilkkao:
Noticed the same problem and fixed it manually as well.

mcb:
This problem also occurs on 10.10.4 and Homebrew 0.9.5. Are you sure this is something El Capitan related?

mcb:
Seems like this is more of an OSX-upgrade issue than El Capitan specific. Running the command above but with the ```-R``` option solved the issue for me as well.

vinkla:
Maybe not El Capitan specific but I first experienced it on El Capitan. Had no problem when updating to 10.10.4 though. That this has to do with the OS X upgrade makes sense.

DomT4:
Not seeing this personally on either 10.10 or 10.11.

If the workaround is `sudo chown -R $(USER):admin /usr/local` that's fairly pleasant at least. Closing as something beyond Homebrew's control, but will shunt any others reporting this problem to this Issue.

phynet:
I'm having the same problem here with 10.11.2 Mac OS X El Capitán.

DomT4:
```
sudo chown -R $(USER):admin /usr/local
```
Remains the advice whilst we continue work on an internal command so people don't have to issue a broad chown.

phynet:
Ups, thank you, with the command `brew update` I saw the command to change the permission
[...]

Issue #41665, opened 2015-07-13
Failed `brew update` on El Capitan (OS X 10.11) Beta


ngocphamm:
I know it's beta, and I shouldn't complain if anything's wrong, but it was good last night and earlier this morning. Then now I have the following output. (My `brewu` is an alias to `brew update && brew outdated`)

```
~ ❯❯❯  brewu                                                                                        
error: unable to unlink old '.gitignore' (Permission denied)                                          
Error: Failure while executing: git pull --quiet origin refs/heads/master:refs/remotes/origin/master
```

Running the command again will get me to this
```
~ ❯❯❯  brewu                                                                                        
Stashing your changes:                                                                                
 M Library/Homebrew/cmd/doctor.rb                                                                     
 D Library/Homebrew/test/.gitignore                                                                   
error: unable to unlink old '.gitignore' (Permission denied)                                         
Error: Failure while executing: git pull --quiet origin refs/heads/master:refs/remotes/origin/master  
```

If anyone is running El Capitan, can you please confirm whether this is my issue or it's universal for homebrew under the beta OS? Thank you guys!

**EDIT:**
I don't know which `.gitignore` file the first run referring to, but the one under `Library/Homebrew/test` has `-rw-------` permissions, and I'm on the same admin user created since the start of the OS (2 years a go, I think).

omega:
I get a similar problem, but for me it also fails to unlink the old `README.md` for some reason :/ After the unsuccessful attempt there are a lot of new files and modified files as well

omega:
Running `brew doctor` told me that `/usr/local` was not writable by my user. `sudo chown $(USER):admin /usr/local` fixed that, and now `brew update` works again. Dunno if it would be worthwhile for `brew update` to also contain that check maybe?

omega:
So to be clear, this is probably not a 10.11 issue, but rather an "upgrade OS X"-issue, as that resets permissions on `/usr/local`

ngocphamm:
Yeah probably, @omega. The `chown` also fixed mine. Thanks for the quick response!

aclissold:
If there are issues when updating OS X, shouldn't brew be updated to anticipate this?

tdsmith:
What do you suggest?

aclissold:
[...]

Issue #41150, opened 2015-06-28
[El Capitan beta 2] brew update error: unable to unlink old 'README.md' (Permission denied)


simonwhitaker:
```
$ brew update
error: unable to unlink old 'README.md' (Permission denied)
Error: Failure while executing: git pull -q origin refs/heads/master:refs/remotes/origin/master
$ brew --version
0.9.5
```

DomT4:
What does `ls -lh $(brew --prefix)` say?

lloeki:
My bet is on `ls -ld $(brew --prefix)`, which will probably show it's owned by root instead of the user.

simonwhitaker:
Correct, it showed `/usr/local/` owned by root. That appears to have changed on installing 10.11 beta 2, I'm pretty sure I used `brew` without issue on beta 1.

(To be clear, I'm not suggesting this is a bug in Homebrew, just giving you a heads-up in case it turns out to be a systemic issue with 10.11 rather than a glitch in the latest beta.)

lloeki:
This is systemic, as the OS X installer does it. Yosemite did that too. Also, `brew doctor`points it out.

DomT4:
Strange. I updated my VM from 10.11 DP1 to DP2 the other night and haven't had any issues with `/usr/local` being reclaimed. Can you `sudo chown -R` it back to you?

GuamBomb:
Could have something to do with "rootless," the new security feature.  I keep finding little things broken by it.  To disable it, you can use the command `<sudo nvram boot-args="rootless=0">` and then reboot.  To turn it back on, you change that 0 to a 1 and reboot again.

DomT4:
We're trying to work with rootless, rather than around it.

For better or worse we have to assume users will have rootless enabled and them doing so actually should reduce the amount of Issues Homebrew fields when people do things like symlinking Homebrew's LLVM over the system one, removing the Python in `/usr/bin`, changing `plists` in `/System` and so on.

`/usr/local` is supposed to be available to developers under rootless, per Apple's announcement.

DomT4:
Moving to tracking in #40837.

PR #42048, opened 2015-07-23
update: print advice on bad permissions

tdsmith:
Print a friendlier error message if we don't have write permission to HOMEBREW_REPOSITORY. Should address situations where OS upgrades change the permissions on /usr/local, like in #41665, #41690, and #41150.

### DIFF
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -1,5 +1,4 @@
 require 'cmd/tap'
-require 'tempfile'
 
 module Homebrew
   def update
@@ -24,7 +23,7 @@ module Homebrew
     begin
       master_updater.pull!
     rescue ErrorDuringExecution => e
-      unless homebrew_repository_writable?
+      unless HOMEBREW_REPOSITORY.writable_real?
         onoe "Can't write to #{HOMEBREW_REPOSITORY}"
         $stderr.puts <<-EOS.undent
           You may be able to fix this error by running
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/update.rb
```diff
@@ -115,6 +128,14 @@ def load_tap_migrations
...
     false
   end
+
+  def homebrew_repository_writable?
+    tf = Tempfile.new(".homebrew-write-probe", tmpdir=HOMEBREW_REPOSITORY)
+    tf.close!
+    true
+  rescue Errno::EACCES
+    false
+  end
```
tdsmith: Oh, I guess File.writable_real? does this.

xu-cheng:
I think it's better to check permissions prior to `pull!` operation. On the one hand, it can result less error message from git, on the other hand it can benefit me to implement the multi-thread version of `brew update`.

tdsmith:
Well, update can succeed if HOMEBREW_REPOSITORY isn't writable, as long as the parent directory of any changed files is writable. Maybe that's why we only see failures being reported from files in the root (.gitignore and README.md). I think we shouldn't introduce a failure if `update` has a chance of succeeding, but maybe it's okay to just print a warning message?

mikemcquaid:
> I think we shouldn't introduce a failure if update has a chance of succeeding, but maybe it's okay to just print a warning message?

I think it'd be good to be precise here; people will get concerned by warnings if they aren't necessary and if they are necessary we should just refuse to update.

mikemcquaid:
> I think it's better to check permissions prior to pull! operation.

Agreed.

aclissold:
When I manually fixed this, I didn't need the `-R` flag. Not sure if that's true in every case.

tdsmith:
We don't know a priori whether a set of permissions will cause an update to
fail because that depends on the content of the update so I don't think we
can satisfy both of those criteria for any given invocation.

mikemcquaid:
Sure, but we can definitely check a directly where we know we have files that should be Homebrew-owned.

PR #42049, opened 2015-07-23
BottleLoader: check version mismatch

xu-cheng:


Library/Homebrew/cmd/cleanup.rb
```diff
@@ -83,7 +83,9 @@ def cleanup_cache
         next
       end
 
-      if f.version > version || ARGV.switch?('s') && !f.installed? || bottle_file_outdated?(f, file)
+      file_is_stale = PkgVersion === version ? f.pkg_version > version : f.version > version
+
+      if file_is_stale || ARGV.switch?('s') && !f.installed? || bottle_file_outdated?(f, file)
```
mikemcquaid: I realise the existing logic is also hairy but might be worth splitting this out into e.g. a few `return`s so it's easier to follow.

Library/Homebrew/cmd/cleanup.rb
```diff
@@ -83,7 +83,9 @@ def cleanup_cache
         next
       end
 
-      if f.version > version || ARGV.switch?('s') && !f.installed? || bottle_file_outdated?(f, file)
+      file_is_stale = PkgVersion === version ? f.pkg_version > version : f.version > version
```
mikemcquaid: Personally I find the ternary operator hard to follow in cases like this.

mikemcquaid:
A few comments but :+1: overall!

Library/Homebrew/cmd/cleanup.rb
xu-cheng: Updated to use `if .. else ..`

Library/Homebrew/cmd/cleanup.rb
xu-cheng: I didn't change anything so far as I realize this whole method need to do a serious change. (e.g. better matching for name, handling resources etc.) So I intend to address this in future PR.

mikemcquaid:
Cool. :+1: once green.

PR #42053, opened 2015-07-23
new option: --dsym

smoofra:
EDIT: I think this is in a pullable state now.  I can build (llvm,coreutils,git,ispell,sqlite,ninja,git) with --dsym and open up the resulting binaries in lldb and everything works great.

EDIT: in case it wasn't clear, the --dsym option also preserves the source in $KEG/Source

This is a first draft.   I realize it probably needs some work.   Comments please! 

If a Formula specifies "option :dsym", and the user specifies '--dsym' on the
command line, then debug symbols will be built and included in the Keg.

This relies on the Apple "lazy dwarf scheme", which leaves the debug info in the
object files, with only links to the object files present in the final
exe/dylib.  The 'dsymutil' tool reads the final exe, finds the objects, and
links all the debug info into a .dSYM bundle.

http://wiki.dwarfstd.org/index.php?title=Apple's_%22Lazy%22_DWARF_Scheme

This also adds a new formula method 'install_dsym', which is responsible for
finding all the MACHO files in the keg and running dsymutil on them.  It
provides a default implementation, which may be overridden if need be.

Signed-off-by: Lawrence D'Anna <larry@elder-gods.org>


Library/Homebrew/build.rb
```diff
@@ -128,6 +128,10 @@ def install
 
         formula.install
 
+        if formula.build.dsym?
+          formula.install_dsym
+        end
+
```
bfontaine: You can do this in one line: `formula.install_dsym if formula.build.dsym?`

Library/Homebrew/extend/ARGV.rb
```diff
@@ -141,6 +141,10 @@ def build_universal?
     include? '--universal'
   end
 
+  def build_dsym?
+    flag? '--dsym'
```
bfontaine: You should use `include?` here because `flag?` will check for both `--dsym` *and* `-d`, which is already a short alias for `--debug`.

smoofra:
Updated to place the source directory in the Keg if --dsym is turned on, and some fixes related to that.

mikemcquaid:
Nice work here. I need to do some proper line-by-line review but the basic approach looks sound. The bigger question is what the other @Homebrew/owners think about adding this functionality (particularly @adamv).

mikemcquaid:
I'm personally relatively indifferent; I can see the usefulness but I wonder if this approach is easily generalised across formulae and if it'll cause any hard-to-debug (no pun intended) issues.

smoofra:
I'll pick some more random formulae and dsymify them to see if any issues pop up.

smoofra:
@mikemcquaid    I've added --dsym capability to a few more packages, each with their own build system.   It seems pretty easy to add --dsym to a package.   Generally all you need to do is:

* set up the build to but object files in their own temp directory, not the source directory.   If this can't be done then just "make clean" after dsyms have been built

* make sure -g is being passed to the compiler and nobody is striping your executables

* run the install_dsym method while the object files are still around

DomT4:
Please give a maintainer a heads up before pushing changes to multiple formulae at once, especially when some of those formula aren't particularly quick to compile. There's a timeout on the bot, and multiple formula built at once can hit it.

I guess my concern here is that looking over the formula files you've edited, it isn't exactly making the formula easier or kinder to understand for someone who isn't a regular contributor. Given we rely on a wide community, somewhat worried more complication is daunting.

smoofra:
oh, ok.   sorry 'bout that.

I guess it's a trade-off of having the feature vs. the extra complexity.   I had originally thought that --dsym could be completely transparent to formulae but this doesn't seem to be possible.  Fortunately if a package is based on cmake or autoconf the added complexity is pretty minimal.    But still nonzero.   

Perhaps If you decide the added complexity in the formulae is too much,  the core changes could still be merged so taps can use it, but the formulae changes could be left out to spare people the complexity?

DomT4:
I'm just wondering if there's a way to deal with this that isn't inline in each and every necessary formula, so that we can add `option :dsym` to formula and not have to add anything else. That would significantly reduce user-visible complexity.

If we have to inline it into each formula it would be nice at least to have a common style and some guidelines/documentation on that. That there are currently quite a few different ways of handling this depending on circumstance is likely to cause confusion, I fear.

mikemcquaid:
> I guess it's a trade-off of having the feature vs. the extra complexity. I had originally thought that --dsym could be completely transparent to formulae but this doesn't seem to be possible. Fortunately if a package is based on cmake or autoconf the added complexity is pretty minimal. But still nonzero.

I think `--universal` should/would be a good guide here. If we can make it that simple I'm :+1:

Library/Formula/coreutils.rb
```diff
@@ -35,8 +37,13 @@ def install
...
     ]
     args << "--without-gmp" if build.without? "gmp"
-    system "./configure", *args
-    system "make", "install"
+
+    src = Pathname.pwd
+    mktemp do
+      system "#{src}/configure", *args
+      system "make", "install"
+      install_dsym if build.dsym?
```
mikemcquaid: What happens if you just run `install_dsym if build.dsym?` as the last thing in `install` with no other code added?

Library/Formula/git.rb
```diff
@@ -137,6 +138,12 @@ def install
     # To avoid this feature hooking into the system OpenSSL, remove it.
     # If you need it, install git --with-brewed-openssl.
     rm "#{libexec}/git-core/git-imap-send" if build.without? "brewed-openssl"
+
+    if build.dsym?
+      install_dsym
+      system "make", "clean"
```
mikemcquaid: The `make clean`s shouldn't be necessary; the build directory is cleaned after the build finishes anyway.

Library/Formula/coreutils.rb
```diff
@@ -35,8 +37,13 @@ def install
       --program-prefix=g
     ]
     args << "--without-gmp" if build.without? "gmp"
-    system "./configure", *args
-    system "make", "install"
+
+    src = Pathname.pwd
+    mktemp do
```
mikemcquaid: This is already in a temporary directory.

Library/Formula/ninja.rb
```diff
@@ -20,16 +46,20 @@ class Ninja < Formula
   end
 
   def install
-    system "python", "configure.py", "--bootstrap"
+    mktemp do
```
mikemcquaid: This is already in a temporary directory.

Library/Homebrew/formula.rb
```diff
@@ -810,6 +815,35 @@ def test_fixtures(file)
   def install
   end
 
+  def install_dsym
+    # create a makefile for this so we can take advantage of multiple cores.
+    dsyms = []
+    mktemp do
```
mikemcquaid: The build directory is already a temporary directory.

Library/Homebrew/formula.rb
```diff
@@ -810,6 +815,35 @@ def test_fixtures(file)
...
+      File.open("Makefile", "w") do |f|
```
mikemcquaid: Use `Pathname.write` and `EOS.undent`

Library/Homebrew/formula.rb
```diff
@@ -923,9 +957,23 @@ def exec_cmd(cmd, args, out, logfn)
     exit! 1 # never gets here unless exec threw or failed
   end
 
+  def source_dir
+    if build.dsym?
+      prefix + "Source"
+    else
+      nil
+    end
+  end
```
mikemcquaid: Why do you have to build out of source?

Library/Formula/coreutils.rb
smoofra: All the object files would be retained in the build directory, which is retained in KEG/Source

Library/Formula/git.rb
smoofra: I made it so the build directory is retained in $KEG/Source

Library/Homebrew/formula.rb
smoofra: the way -g  and  dsymutil work on OS X is almost all of the debug information is retained in the object files, and not copied to the final executable when it is linked.   The executable only has links to the object files.   So in order to collect dsyms we need to run dsymutil before all the object files have been cleaned up.   

Also, we want to retain the source files so the debugger will work, but we don't want to retain all those object files, so I made the build directory be $KEG/Source instead of a temp dir if --dsym is selected, and it's up to the formula to make sure that it doesn't leave a bunch of object files sitting there.

Library/Formula/coreutils.rb
mikemcquaid: The object files are in the (already temporary) build directory rather than the Keg.

mikemcquaid: Ah, ok, I see. You copy across all of the source? I'm not sure we want that.

smoofra: you think people would want debug info but not source?

Library/Homebrew/formula.rb
mikemcquaid: Ok. The option (and PR description) should be updated to reflect that all the source code is also copied.

Perhaps a better approach here is to initialise a Git repository in the source directory and copy the source files using `git archive master | tar -x -C /somewhere/else` or similar. That would mean you aren't reliant on out-of-source builds.

smoofra: oh, yea that's a good idea.  I'll give that a try.

Library/Formula/coreutils.rb
mikemcquaid: Elaborated below a bit; it just wasn't obvious from the PR title/option name.

smoofra:
@mikemcquaid OK if I push a bunch of formula updates again?

smoofra:
@mikemcquaid I've updated this along the lines you suggested.    

Most "normal" formulas should be able to just add "option :dsym" now and not need to do anything else.

Library/Homebrew/formula.rb
smoofra: ?? the makefile needs a stanza for each dsym...

Library/Formula/coreutils.rb
```diff
@@ -26,13 +26,23 @@ class Coreutils < Formula
...
+
   depends_on "gmp" => :optional
 
   def install
+    # ./configure makes a diabolical un-deleteable directory called confdir-14B---
+    # Let's skip that test.
+    (buildpath/"config.cache").write(<<-END.undent)
+      gl_cv_func_getcwd_path_max='no, but it is partly working'
+      gl_cv_func_getcwd_abort_bug='yes'
+    END
```
mikemcquaid: We shouldn't need to skip tests for this.

Library/Formula/git.rb
```diff
@@ -137,6 +138,7 @@ def install
     # To avoid this feature hooking into the system OpenSSL, remove it.
     # If you need it, install git --with-brewed-openssl.
     rm "#{libexec}/git-core/git-imap-send" if build.without? "brewed-openssl"
+
```
mikemcquaid: Remove this whitespace change.

Library/Formula/coreutils.rb
```diff
@@ -26,13 +26,23 @@ class Coreutils < Formula
...
+
     system "./bootstrap" if build.head?
     args = %W[
       --prefix=#{prefix}
       --program-prefix=g
+      -C
```
mikemcquaid: What's this do?

Library/Formula/ispell.rb
```diff
@@ -22,12 +24,18 @@ def install
       s.gsub! "/usr/local", prefix
       s.gsub! "/man/man", "/share/man/man"
       s.gsub! "/lib", "/lib/ispell"
+      s.gsub! "#endif", %(#define CFLAGS "-g"\n\n#endif) if build.dsym?
+    end
+    if build.dsym?
+      inreplace ["Makefile", "deformatters/Makefile"] do |s|
+        s.gsub! /^\t\s*strip.*?\n/, ""
+      end
     end
```
mikemcquaid: Is there no option to disable stripping?

Library/Formula/llvm.rb
```diff
@@ -128,6 +129,8 @@ def install
     (buildpath/"tools/lld").install resource("lld") if build.with? "lld"
     (buildpath/"tools/lldb").install resource("lldb") if build.with? "lldb"
 
+    commit_source
```
mikemcquaid: Why does this need to be called explicitly?

Library/Homebrew/build.rb
```diff
@@ -107,9 +107,9 @@ def install
     formula.brew do
       formula.patch
 
-      if ARGV.git?
-        system "git", "init"
-        system "git", "add", "-A"
+      if ARGV.git? or formula.build.dsym?
```
mikemcquaid: Use `||`

Library/Homebrew/build.rb
```diff
@@ -124,16 +124,34 @@ def install
 
         interactive_shell(formula)
       else
+
```
mikemcquaid: Don't need this exmpty line.

Library/Homebrew/formula.rb
mikemcquaid: Ping.

Library/Homebrew/formula.rb
mikemcquaid: Sure; you can create the string you're writing in advance and use `EOS.undent` with interpolated strings.

Library/Homebrew/formula.rb
```diff
@@ -810,6 +815,56 @@ def test_fixtures(file)
...
+        f.write("\n")
+      end
+      if dsyms
+        system "make", "dsyms"
+      end
+    end
+  end
+
+  def commit_source
+    if not build.dsym?
```
mikemcquaid: `return unless build.dsym?`

Library/Homebrew/formula.rb
```diff
@@ -810,6 +815,56 @@ def test_fixtures(file)
...
+      return
+    end
+
+    safe_system "git", "add", "-A"
+
+    status = `git status --porcelain`
```
mikemcquaid: Why not use `safe_system`?

Library/Homebrew/formula.rb
```diff
@@ -810,6 +815,56 @@ def test_fixtures(file)
...
+    if $? != 0
+      raise ErrorDuringExecution.new("git", ["status", "--porcelain"])
+    end
+
+    if status == ""
```
mikemcquaid: `return if status.to_s.empty?`

Library/Homebrew/formula.rb
```diff
@@ -810,6 +815,56 @@ def test_fixtures(file)
...
+      #nothing to commit
+      return
+    end
+
+    safe_system "git", "commit", "--quiet", "-m", "Homebrew automatic commit: saving source",
+                "--author=Homebrew <brew-auto-commit@brew.sh>"
```
mikemcquaid: Probably don't want to define this email that doesn't exist.

Library/Homebrew/formula.rb
mikemcquaid: Now that you're using Git you shouldn't need this now.

mikemcquaid:
I've added a bunch of comments. Before working on them you may want to let other @Homebrew/owners discuss this as I think a few of us have reservations about merging and supporting this long-term if most formulae are going to require patches for it to work.

Library/Formula/coreutils.rb
smoofra: Enables config.cache

Library/Formula/coreutils.rb
smoofra: How should it be delt with?

Library/Formula/ispell.rb
smoofra: Nope

Library/Formula/llvm.rb
smoofra: Because we added more source code via a resource.   We dont it thrown away with the obect files

Library/Homebrew/formula.rb
smoofra: I need the stdout.

Library/Homebrew/formula.rb
smoofra: If I took this out then the debug info would all point to paths in the temporary directory, instead of where the files ultimately wind  up.   The user would have to explicitly tell the debugger where to find source.

Library/Homebrew/formula.rb
smoofra: What should the author be set to?

mikemcquaid:
I'm afraid I think we'll have to pass on this. It requires too much code in Homebrew core, weird things like building directly in the `prefix` and even then requires formulae modifications. I'm really sorry but I just think the way Homebrew is built isn't well suited to this functionality.

smoofra:
:(

smoofra:
@mikemcquaid That's disappointing.   I'm not sure how typical I am as a user but I would find this feature very valuable.   I'm willing to keep making changes until it's up to snuff.

If you change your mind, let me know.

xu-cheng:
@smoofra Since such feature needs formula modification. If you need to such feature, I suggest you support them in your own tap. You can see `homebrew/php` how to write abstract for formulae. Sorry passing this now.

tschoonj:
FWIW I am also very much in favor of adding support for debugging symbols to Formulas. Would make my life considerably easier :smile: 

@smoofra Nice work! Hope it will be merged in one day...

ahundt:
I would start using this feature immediately if it were revived!

DomT4:
If we can get this to a place where it doesn't require hacking formula up for it to work, and we can do something as simple as the current `:universal` implementation I think most of the original concerns would be addressed. There was just way too much need for formula patching here, and building things inside the prefix instead of a sandboxed buildpath isn't something we particularly want to adopt.

We'd want this to "just work", essentially, without changing significant fundamentals of the Homebrew build process to get there.

xu-cheng:
To repeat what I said above, I think this function better be implemented inside a tap as it requires certain formula patching.

mikemcquaid:
> If we can get this to a place where it doesn't require hacking formula up for it to work, and we can do something as simple as the current :universal implementation I think most of the original concerns would be addressed. There was just way too much need for formula patching here, and building things inside the prefix instead of a sandboxed buildpath isn't something we particularly want to adopt.
> 
> We'd want this to "just work", essentially, without changing significant fundamentals of the Homebrew build process to get there.

Agreed.

lwahonen:
I was googling how to make a debug build with Homebrew and was so stoked for a while. Hope this gets revived at some point.

mikemcquaid:
A question here for @smoofra: does e.g. Debian install source code or just the debug symbols? I was of the impression it was the latter. If I'm correct there: would it be easier to meet our requirements mentioned above by just installing debug symbols?

megahall:
@mikemcquaid The difficulty with "just installing debug symbols" is that OS X itself handles symbols in a fantastically broken way. OS X stores a lot of the symbols in hard-coded paths referenced from inside of symbol tables in the executables themselves.

In Debian this challenge is not an issue as you can configure a default symbols search path inside of `gdb`, `lldb` and other debugging and disassembly tools, and then place any symbol files in a particular hierarchy under that tree, and everything on the system knows where to find the symbols.

mikemcquaid:
@megahall Thanks!

PR #42063, opened 2015-07-23
Make launchd config listen safer

osabina:
Having it listen on localhost only seems a safer default choice.


osabina:
Let me resubmit this from a topic branch.

PR #42067, opened 2015-07-23
Use SHA-256 for antlr formula

alykhank:
Switch from SHA-1 to SHA-256 for `antlr` formula. This resolves the error from `brew audit antlr`:
```
antlr:

Error: 1 problem in 1 formula
```


dunn:
Merged in 586f1c7942627e10736142d3567ea8dc40315b0b, thanks!

PR #42068, opened 2015-07-23
imgur-screenshot 1.5.4 (new formula)

ftc2:
https://github.com/jomo/imgur-screenshot
The Imgur-Screenshot uploader for Linux & OS X from imgur.com/tools
"Take screenshot selection, upload to imgur. + more cool things"

### DIFF
--- a/Library/Formula/imgur-screenshot.rb
+++ b/Library/Formula/imgur-screenshot.rb
@@ -1,0 +1,19 @@
+class ImgurScreenshot < Formula
+  desc "Take screenshot selection, upload to imgur. + more cool things"
+  homepage "https://github.com/jomo/imgur-screenshot"
+  url "https://github.com/jomo/imgur-screenshot/archive/v1.5.4.tar.gz"
+  sha256 "147fe1c3b3e407d952fa8abbdb9b436192133cf7a56278edc0dd87c2a9c9c430"
+  head "https://github.com/jomo/imgur-screenshot.git"
+
+  option "with-terminal-notifier", "Needed for Mac OS X Notifications"
+
+  depends_on "terminal-notifier" => :optional
+
+  def install
+    bin.install "imgur-screenshot.sh"
+  end
+
+  test do
+    system "#{bin}/imgur-screenshot.sh", "--check"  # checks deps
+  end
+end


Library/Formula/imgur-screenshot.rb
```diff
@@ -0,0 +1,23 @@
...
+  homepage "https://github.com/jomo/imgur-screenshot"
+  head "https://github.com/jomo/imgur-screenshot.git"
+  url "https://github.com/jomo/imgur-screenshot/archive/v1.5.4.tar.gz"
+  sha256 "147fe1c3b3e407d952fa8abbdb9b436192133cf7a56278edc0dd87c2a9c9c430"
+
+  option "with-terminal-notifier", "terminal-notifier is needed for Mac OS X User Notifications"
+  depends_on "terminal-notifier" => :optional
+
+  def install
+    bin.install "imgur-screenshot.sh"
```
DomT4: Is this usually installed with the `.sh` extension left intact?

Library/Formula/imgur-screenshot.rb
```diff
@@ -0,0 +1,23 @@
+class ImgurScreenshot < Formula
+  desc "Take screenshot selection, upload to imgur. + more cool things"
+  homepage "https://github.com/jomo/imgur-screenshot"
+  head "https://github.com/jomo/imgur-screenshot.git"
+  url "https://github.com/jomo/imgur-screenshot/archive/v1.5.4.tar.gz"
+  sha256 "147fe1c3b3e407d952fa8abbdb9b436192133cf7a56278edc0dd87c2a9c9c430"
+
+  option "with-terminal-notifier", "terminal-notifier is needed for Mac OS X User Notifications"
```
DomT4: Can drop the Mac OS X part here. It's obvious enough given the platform Homebrew targets.

Library/Formula/imgur-screenshot.rb
```diff
@@ -0,0 +1,23 @@
...
+  depends_on "terminal-notifier" => :optional
```
DomT4: Please leave an empty line between the option and the dependency.

Library/Formula/imgur-screenshot.rb
ftc2: it's a single shell script, and the official documentation reflects using it as-is... do you suggest changing it?

Library/Formula/imgur-screenshot.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def install
+    bin.install "imgur-screenshot.sh"
+  end
+
+  test do
+    system "#{bin}/imgur-screenshot.sh", "--version"
```
DomT4: Let's only do one thing here rather than several small basic tests.

Library/Formula/imgur-screenshot.rb
DomT4: Nope. If upstream recommend people use it with the `.sh` extension intact, we should respect that.

Library/Formula/imgur-screenshot.rb
ftc2: I just wanted to clarify that that's referring to the User Notifications feature of Mac OS X.
If you don't think that's helpful, I'll get rid of it.  Thoughts?

DomT4: Na, It's fine. Just wasn't sure if the Mac OS X part was redundant since that's the platform all Homebrew uses will be running. Let's de-dupe the readout though, so it just reads:

`"with-terminal-notifier", "Needed for OS X User Notifications"` or something.

Library/Formula/imgur-screenshot.rb
```diff
@@ -0,0 +1,19 @@
+class ImgurScreenshot < Formula
+  desc "Take screenshot selection, upload to imgur. + more cool things"
+  homepage "https://github.com/jomo/imgur-screenshot"
+  head "https://github.com/jomo/imgur-screenshot.git"
```
dunn: The `head` line should be below `url` and `sha256`.

ftc2: k

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/348cf53fa64eb284bd0dbc314be2734c219980e7. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

jomo:
Awesome, thanks! :heart: 
*(I'm the author of imgur-screenshot)*

PR #42069, opened 2015-07-23
brew: print bug report URL for internal commands only

bfontaine:
I don’t think we should print the Homebrew bug report URL when an external command fails. We *could* print the tap if it comes from a `cmd/` directory, but some external commands don’t come from a tap.

This is tricky, because the failure might be related to Homebrew (or not).

### DIFF
--- a/Library/brew.rb
+++ b/Library/brew.rb
@@ -68,6 +68,7 @@ begin
   empty_argv = ARGV.empty?
   help_regex = /(-h$|--help$|--usage$|-\?$|^help$)/
   help_flag = false
+  internal_cmd = true
   cmd = nil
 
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

Library/brew.rb
```diff
@@ -68,6 +68,7 @@ def require? path
   empty_argv = ARGV.empty?
   help_regex = /(-h$|--help$|--usage$|-\?$|^help$)/
   help_flag = false
+  internal_cmd = true
```
bfontaine: @mikemcquaid I just added this line, because without it any exception raised between the beginning of the block and the `internal_cmd` assignment wouldn’t trigger the “please report this bug” message. Does it makes sense?

PR #42071, opened 2015-07-23
libbpg: Add optional x265 support

felixbuenemann:
This allows to compile libbpg against x265, which offers much faster encoding than the built-in tcvtc encoder at roughly the same quality.

Also added an option to completely disable tcvtc which makes x265 the default encoder.

### DIFF
--- a/Library/Formula/libbpg.rb
+++ b/Library/Formula/libbpg.rb
@@ -1,10 +1,8 @@
-require "formula"
-
 class Libbpg < Formula
   desc "Image format meant to improve on JPEG quality and file size"
   homepage "http://bellard.org/bpg/"
   url "http://bellard.org/bpg/libbpg-0.9.5.tar.gz"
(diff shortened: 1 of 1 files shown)


dunn:
Would you be so kind as to update the `sha1`s to `sha256` at the same time? Thanks!

felixbuenemann:
@dunn Updated to sha256 and removed deprecated `require "formula"`.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/ba363d2c5f953fea2067a40baad73cd1fab1a44c, Thank you @felixbuenemann!

World's smallest nit here, but would prefer one empty line between the last `args` line and the command where `args` is invoked. Makes it a little easier to glance over.

felixbuenemann:
Agree, a bit more white space couldn't have hurt. Thanks!

PR #42072, opened 2015-07-23
jxrlib: New Formula

felixbuenemann:
This formula provides the `JxrEncApp` and `JxrDecApp` utilities that allow to encode and decode [JPEG-XR](https://en.wikipedia.org/wiki/JPEG_XR) images. The utilities will also be used by imagemagick as a delegate to enable jxr support.

The formula is head only, because the latest stable release is from 2013 and can only be built on windows.

### DIFF
--- a/Library/Formula/jxrlib.rb
+++ b/Library/Formula/jxrlib.rb
@@ -1,0 +1,12 @@
+require 'formula'
+
+class Jxrlib < Formula
+  desc "Tools for JPEG-XR image encoding/decoding"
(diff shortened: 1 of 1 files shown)


mistydemeo:
We don't accept head-only formulae in this repository; can you submit this to [homebrew/head-only](https://github.com/Homebrew/homebrew-head-only) instead please? Alternately, can you request a release from the developers? Thanks!

felixbuenemann:
Oh that's fine, I didn't know about the head-only tap.

PR #42074, opened 2015-07-23
update: always report changes to branch 'master'

UniqMartin:
**(I don't think this initial version of the PR should be commited as-is. See “Problem” part below.)**

If the user's working copy is *not* on the 'master' branch, the revision recorded after the update to the 'master' branch is not the one of the updated 'master' branch but that of whatever the user's current branch is. This tends to be not very helpful.

This change records the revision *after* the update to the 'master' branch, but *before* switching back to the user's current branch prior to the update, effectively causing the report to show changes to the 'master' branch.

Since `brew update` tries very hard to retain/restore the state of the working copy if the user is *not* on the 'master' branch, I believe this behavior to be more useful.

*Problem:* Since generating the report (added, deleted, updated formulae) also relies on the state of the working copy, I see two solutions to get a report that always shows the updates to the 'master' branch:

1. Forcibly generate the report before switching back to the user's branch. (I did this in my second commit, but in a rather hackish way. This is more of a proof of concept.)
2. Teach `Updater#report` to work directly with the Git repository for both the previous and current revision instead of relying on the state of the working copy.

Do others agree that the change in behavior, that I implemented here, is desirable? If so, what would be the best/cleanest way to solve the outlined reporting problem? (I need some guidance here, as I'm coming close to the limits of my understanding of Homebrew/Ruby.)

### DIFF
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -232,6 +232,8 @@ class Updater
     map = Hash.new { |h, k| h[k] = [] }
 
     if initial_revision && initial_revision != current_revision
+      wc_revision = read_current_revision
+
       diff.each_line do |line|
         status, *paths = line.split
         src = paths.first
@@ -247,7 +249,11 @@ class Updater
           file = repository.join(src)
           begin
             formula = Formulary.factory(file)
-            new_version = formula.pkg_version
+            new_version = if wc_revision == current_revision
+              formula.pkg_version
+            else
+              FormulaVersions.new(formula).formula_at_revision(@current_revision, &:pkg_version)
+            end
             old_version = FormulaVersions.new(formula).formula_at_revision(@initial_revision, &:pkg_version)
             next if new_version == old_version
           rescue FormulaUnavailableError, *FormulaVersions::IGNORED_EXCEPTIONS => e
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -202,6 +202,8 @@ class Updater
 
     reset_on_interrupt { safe_system "git", *args }
 
+    @current_revision = read_current_revision
+
     if @initial_branch != "master" && !@initial_branch.empty?
       safe_system "git", "checkout", @initial_branch, *quiet
     end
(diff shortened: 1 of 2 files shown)


mikemcquaid:
Sorry that it's taking so long to review this; we definitely want this fix. I need to have more of a think about the best way of approaching this but thanks for the PR :+1:

mikemcquaid:
> Teach Updater#report to work directly with the Git repository for both the previous and current revision instead of relying on the state of the working copy.

This definitely seems like the best idea and this change would be desirable :+1:.

mikemcquaid:
I'd say: try to mash something out even if it's a hacky mess and we'll give you Homebrew/Ruby(/Git?) guidance. How's that sound? Thanks for all this :heart:

UniqMartin:
Thanks for the feedback! I'll try to implement my second proposal (the one favored by you and me) and either report back with an updated branch or any major questions/issues I encounter on the way.

mikemcquaid:
Thanks @UniqMartin!

UniqMartin:
Sorry for the long silence. I had to think about the best way to solve this and had not much spare time.

**About my current approach:**

Ideally, the code that checks if a formula version has changed, would construct a `Formula` instance directly from a repository blob. However, that would have meant hacking `Formulary.factory` and related code and after a few attempts I gave up. Now I'm constructing the `Formula` instance from whatever is there in the working copy, in the hopes that the formula file still exists. (This shouldn't be an issue, unless the branch someone is working on has severely diverged from master.)

Also, only if HEAD != master, does the code fetch `new_version` from the repository, thus avoiding the penalty of a second call to `git` for most users.

**One more note:**

Commit 71a4f501724840a484b1876a1a3c322d539e2343 (setting `@current_revision` after updating master and before restoring the user's working copy) looks rather innocent, but `brew update` also uses the generated report to perform tap migrations and formula renames. I guess that could fail in strange ways if `@current_revision` and HEAD diverged too much.

mikemcquaid:
I'm not quite sure what to add to your comment! Do you consider this otherwise ready-to-go? @xu-cheng thoughts?

UniqMartin:
> Do you consider this otherwise ready-to-go?

Yes, I do.

xu-cheng:
:+1: I think it's ready to be merged

mikemcquaid:
Thanks @UniqMartin!

PR #42075, opened 2015-07-24
libmemcached: enhanced coverage options

denji:
###### Enhanced coverage options, fixed value `--with-memcached-sasl`:
- Support memcached protocol (requires libevent2)
- Support load generation and benchmark tool
- Support hsieh hashing
- Static library optional, `depends_on "libmemcached" => "with-static"`

### DIFF
--- a/Library/Formula/libmemcached.rb
+++ b/Library/Formula/libmemcached.rb
@@ -1,10 +1,8 @@
-require "formula"
-
 class Libmemcached < Formula
   desc "C and C++ client library to the memcached server"
   homepage "http://libmemcached.org"
   url "https://launchpad.net/libmemcached/1.0/1.0.18/+download/libmemcached-1.0.18.tar.gz"
-  sha1 "8be06b5b95adbc0a7cb0f232e237b648caf783e1"
+  sha256 "e22c0bb032fde08f53de9ffbc5a128233041d9f33b5de022c0978a2149885f82"
   revision 1
 
   bottle do
@@ -15,7 +13,11 @@ class Libmemcached < Formula
   end
 
   option "with-sasl", "Build with sasl support"
+  option "with-static", "Build static library"
+  option "with-memaslap", "Build with load generation and benchmark tool support"
+  option "with-hsieh-hash", "Build with hsieh hashing support"
 
+  depends_on "libevent" # requires --enable-libmemcachedprotocol
   if build.with? "sasl"
     depends_on "memcached" => "enable-sasl"
   else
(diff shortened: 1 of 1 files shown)


Library/Formula/libmemcached.rb
```diff
@@ -1,11 +1,9 @@
-require "formula"
-
 class Libmemcached < Formula
   desc "C and C++ client library to the memcached server"
   homepage "http://libmemcached.org"
   url "https://launchpad.net/libmemcached/1.0/1.0.18/+download/libmemcached-1.0.18.tar.gz"
-  sha1 "8be06b5b95adbc0a7cb0f232e237b648caf783e1"
-  revision 1
+  sha256 "e22c0bb032fde08f53de9ffbc5a128233041d9f33b5de022c0978a2149885f82"
+  revision 2
```
DomT4: Don't need the revision bump here really.

Library/Formula/libmemcached.rb
```diff
@@ -15,7 +13,11 @@ class Libmemcached < Formula
   end
 
   option "with-sasl", "Build with sasl support"
+  option "with-static", "Build static library"
+  option "with-memaslap", "Build with load generation and benchmark tool support"
+  option "with-hsieh-hash", "Build with hsieh hashing support"
 
+  depends_on "libevent" # requires --enable-libmemcachedprotocol
```
DomT4: Does this have to be mandatory? Can it not be optional or recommended, or is the gain in functionality useful enough to merit the new mandatory dependency?

denji: @DomT4 Judging by the output of ./configure, it requests it anyway libevent. Besides RHEL distributions use the default libmemcachedprotocol for full coverage, such as php*-memcached includes such possible additional options.

Library/Formula/libmemcached.rb
denji: @DomT4 Thanks, indeed it would be undesirable for have already installed.

Library/Formula/libmemcached.rb
DomT4: Presumably, it isn't mandatory though, otherwise it wouldn't have built before now? I'm not against adding the dependency, just want to understand how widely it is used elsewhere - What do Debian et al. do on this? Let's kill the comment off either way.

denji: @DomT4 Example here https://github.com/php-memcached-dev/php-memcached/pull/98 if it is missing it too will only work without additional options.

tdsmith:
Should we do something here?

tdsmith:
Thanks & sorry @denji!

Issue #41833, opened 2015-07-17
freeswitch: have Homebrew (not makefile) deal with downloading sounds?


L2G:
While updating the freeswitch formula, I noticed that Freeswitch's main makefile installs sound files (voice prompts, music-on-hold, etc.) by calling curl or wget to download tarballs directly from Freeswitch's file repository.

Of course it's possible to define the sound files as formula resources so Homebrew can handle the downloading and caching itself. However, it would require defining 8 to 16 new resources within the formula.

Is that overkill, or would the benefits still outweigh the hassle?

mikemcquaid:
> would the benefits still outweigh the hassle?

They would. Go for it.

DomT4:
In an ideal world, all unverified remote code execution not initiated by Homebrew should be set on :fire: and alternatives found.

DomT4:
Entirely welcome a PR on this.

PR #42078, opened 2015-07-24
freeswitch: convert sound downloads into resources

L2G:
This avoids having the makefile do its own downloads with curl (as discussed in #41833).

The versions of sound files were chosen based on the versions that are specified in the source code and downloaded via the makefile targets previously used. (Comments in the formula indicate where to find the versions specified in the source code.)

A by-product of these changes is that _all_ bit rates for the French sound files are downloaded now. Previously, only the 48-kHz sound files were downloaded, due to omissions in the makefile.


dunn:
Thanks so much for taking care of this! Merged in 546a4611da1605d12a5f999438a1ff606ff3a6a9.

PR #42080, opened 2015-07-24
llvm: add functional tests for clang

smoofra:
add functional tests for clang

### DIFF
--- a/Library/Formula/llvm.rb
+++ b/Library/Formula/llvm.rb
@@ -173,7 +173,37 @@ class Llvm < Formula
     EOS
   end
 
+  FOO_C = <<'END'
+    #include <stdio.h>
+    int main() {
+        printf("Hello, C!\n"); return 0;
+    }
+END
+
+  FOO_CC = <<'END'
+    #include <iostream>
+    int main() {
+       std::cout << "Hello, C++!\n"; return 0;
+    }
+END
+
   test do
     system "#{bin}/llvm-config", "--version"
+
+    return if build.without? "clang"
+
+    mktemp do
+      File.open("foo.c", "w") do |f|
+        f.write(FOO_C)
+      end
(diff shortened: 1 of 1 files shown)


DomT4:
We don't test non-default options, so I'm not really sure where this would ever get used in a standard use case, to be honest. The places we have `if build.with?` elements inside `test do` tend to be because those elements are built by default with an optional off switch.

smoofra:
maybe --with-clang should actually be the default?   most people who want llvm are going to want clang as well.

DomT4:
Killed the test-bot build here, since we won't be testing the actual changes in this PR without Clang becoming the default, and the CI queue is already starting to back up.

I'm not sure what the explanation is on not enabling Clang by default. I suspect it may be to do with a few factors:

* The bottle isn't `cellar :any` which means every time we update or revision the formula everyone outside of `/usr/local` has to built it from source, which takes a bit over 40 minutes at the moment on a modernish machine.
* It already takes up an hour of bot time every single time someone files a PR.
* The above two variables would jump by usually another hour or more when the Clang element is enabled by default.
* Homebrew/Versions LLVM formulae have Clang available by default.

mikemcquaid:
> I'm not sure what the explanation is on not enabling Clang by default.

Basically because we don't want to interfere with Xcode's Clang by adding an easily, well-supported method of potentially breaking our compilation of other packages. This may be invalid at this point but was the past, cautious reasoning.

smoofra:
Doesn't being a keg-only formula protect us from that though?

DomT4:
For better or worse, people can and often do force `keg_only` packages to link. There's no mechanism to forcefully stop them and go "This will completely screw up your compiles and shouldn't be done".

DomT4:
Passing on this. We never test non-default/recommended options, and there's little hunger to enable Clang by default in the Core. Thanks for the submission though!

PR #42081, opened 2015-07-24
coreutils: add gmp support

guoxiao:


DomT4:
If we're going to create an `args` array let's put everything in it, rather than tacking it onto the end of the existing block. Looks a lot neater.

```ruby
args = %w[
  --prefix=#{prefix}
  --program-prefix=g
]

args << "--without-gmp" if build.without? "gmp"

system "./configure", *args
```

guoxiao:
@DomT4 It should be `%W` instead of `%w` to support string interpolation.

DomT4:
Yup. That's what happens when you get very little sleep and start the morning by staring at various chunks of code :smile_cat:.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/cf377380ee63d1080847944dbcebf7720f016f6c, Thanks @guoxiao!

PR #42084, opened 2015-07-24
ntl 9.3.0

nijikon:


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/09af77dbe959de5af2c5e709379e36385c542783, Thanks again @nijikon! :cocktail:.

I removed the new `test do` block on pull. We have a fairly strict policy on only accepting tests that test functionality rather than only existence and the hope is the strict audit failure will at some point nudge someone into writing a library test for it, heh. Thanks though, We do appreciate your desire to add tests for formulae & meet the strict audit.

PR #42088, opened 2015-07-24
collectd 5.5.0

nijikon:


Library/Formula/collectd.rb
```diff
@@ -25,6 +19,13 @@ class Collectd < Formula
     depends_on "autoconf" => :build
   end
 
+  # Will fail against Java 1.7
+  option "with-java", "Enable Java 1.6 support"
```
mikemcquaid: You can do this with a `depends_on :java => [:optional, "1.6"]` I believe?

Library/Formula/collectd.rb
```diff
@@ -80,4 +81,16 @@ def plist; <<-EOS.undent
...
+  test do
+    begin
+      fork { system sbin/"collectd" }
+      sleep 2
+      assert File.exist?("#{var}/run/collectd.pid")
+    ensure
+      pid = File.read("#{var}/run/collectd.pid").to_i
+      Process.kill("SIGINT", pid)
+      Process.wait(pid)
+    end
```
mikemcquaid: Test is failing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/30168/version=yosemite/testReport/junit/brew-test-bot/yosemite/test_collectd/

Library/Formula/collectd.rb
nijikon: There is already line ```depends_on :java => ["1.6", :optional]```. What do you want me to do here?

Library/Formula/collectd.rb
```diff
@@ -80,4 +81,14 @@ def plist; <<-EOS.undent
     </plist>
     EOS
   end
+
+  test do
+    begin
+      system sbin/"collectd"
+      sleep 2
+      assert File.exist?("#{var}/run/collectd.pid")
```
xu-cheng: This test will fail under sandbox. As we don't allow to write files outside of temporary directory during the test. Can you set this file to builldpath by argument or something?

nijikon:
@xu-cheng I rewrote the test, please verify.

xu-cheng:
Thanks @nijikon

PR #42089, opened 2015-07-24
agda: ghc and audit fix

nijikon:


DomT4:
What's being fixed here?

Library/Formula/agda.rb
```diff
@@ -10,13 +10,19 @@ class Agda < Formula
...
+
   bottle do
     sha256 "b5dfadfde1b45ca3d70f2ac812e98a6d72caf1d840d4cb9c4eee95e81d74ff0b" => :yosemite
     sha256 "4513aecea2fbf5d1a682e04e4cccfd795d48836bcf1b16f79c24fa3bd6482854" => :mavericks
     sha256 "d0e17808ee896cdedc499b010b67961dc3db3c430c501faef16c4b4482238447" => :mountain_lion
   end
 
-  head "https://github.com/agda/agda.git", :branch => "master"
+  head do
+    resource "stdlib" do
```
dunn: The head `url` should be placed in this block as well, to keep everything in one place.

mikemcquaid:
> What's being fixed here?

Would like some clarification on this too, thanks.

nijikon:
This just makes ```brew audit --strict``` pass without errors.

dunn:
Things that take a long time to build, like agda, aren't really worth updating for style fixes, since it still has to run through the test bots.

Library/Formula/agda.rb
```diff
@@ -30,12 +36,6 @@ class Agda < Formula
...
 
-  head do
-    resource "stdlib" do
-      url "https://github.com/agda/agda-stdlib.git", :branch => "master"
-    end
-  end
-
   resource "stdlib" do
     url "https://github.com/agda/agda-stdlib.git", :branch => "2.4.2.4"
   end
```
xu-cheng: It seems this should be put inside stable do block otherwise it will overwrite the resource for head.

nijikon:
It looks like in order for the test to pass, ghc needs to be installed. What do we want to do here?

DomT4:
Change the test, probably. I'm relying on people who actually use the software here, but apparently GHC isn't runtime, or at least isn't expected to be runtime.

mikemcquaid:
> Change the test, probably.

Yep :+1:

DomT4:
Will take a PR for an updated test that doesn't consistently break.

PR #42093, opened 2015-07-24
Rebuild libgcrypt (1.6.3) bottle to use current libgpg-error (1.19) library.

tzudot:
update libgcrypt to use current libgpg-error library.

### DIFF
--- a/Library/Formula/libgcrypt.rb
+++ b/Library/Formula/libgcrypt.rb
@@ -4,7 +4,8 @@ class Libgcrypt < Formula
   url "ftp://ftp.gnupg.org/gcrypt/libgcrypt/libgcrypt-1.6.3.tar.bz2"
   mirror "http://ftp.heanet.ie/mirrors/ftp.gnupg.org/gcrypt/libgcrypt/libgcrypt-1.6.3.tar.bz2"
   mirror "ftp://mirror.tje.me.uk/pub/mirrors/ftp.gnupg.org/libgcrypt/libgcrypt-1.6.3.tar.bz2"
-  sha1 "9456e7b64db9df8360a1407a38c8c958da80bbf1"
+  sha256 "41b4917b93ae34c6a0e2127378d7a4d66d805a2a86a09911d4f9bd871db7025f"
+  revision 1
 
   bottle do
     cellar :any
@@ -13,14 +14,14 @@ class Libgcrypt < Formula
     sha1 "3bf6ac3f6bc55a8fbd51b3fc9d7fd6677469d14e" => :mountain_lion
   end
 
-  depends_on "libgpg-error"
-
   option :universal
+
+  depends_on "libgpg-error"
 
   resource "config.h.ed" do
     url "http://trac.macports.org/export/113198/trunk/dports/devel/libgcrypt/files/config.h.ed"
     version "113198"
-    sha1 "136f636673b5c9d040f8a55f59b430b0f1c97d7a"
+    sha256 "d02340651b18090f3df9eed47a4d84bed703103131378e1e493c26d7d0c7aab1"
   end
 
   def install


tzudot:
Currently, the linked libgpg-error package is 1.18.
```sh
$ libgcrypt-config --libs                                                                                                                        
-L/usr/local/Cellar/libgcrypt/1.6.3/lib -lgcrypt -L/usr/local/Cellar/libgpg-error/1.18/lib -lgpg-error
```
This update is to revise this library to use libgpg-error 1.19.

chdiza:
I don't understand.  `libgcrypt` already depends on `libgpg-error`.  If you think your `libgpg-error` is too old, then you just need to `brew update` and rebrew `libgpg-error`.

Is there some reason why you believe that  `libgcrypt` requires a particular version of `libgpg-error` here?

Even if there is such a need (and I doubt there is), the changes you made don't accomplish that.

tzudot:
The current version of `libgpg-error => 1.19` in Homebrew. But the current bottled version of `libgcrypt => 1.6.3` is linked to `libgpg-error => 1.18` leading to failures in compilation of applications depending on `libgcrypt`.
How to fix this?

chdiza:
Can you post a simple, reproducible example of a compilation failure, following the troubleshooting steps in the docs?  That shouldn't happen, given the small difference between 1.18 and 1.19.

tzudot:
For instance, while building `wireshark`, the compiler fails to find the path for `libgpg-error => 1.18`, which is not installed anymore. Sorry, couldn't get `gist-logs`, there was a [buffer error](https://github.com/Homebrew/homebrew/issues/42100) while executing the command.

DomT4:
Wireshark is actually pulling the flags from `libgcrypt-config`, aye:

``` -I/usr/local/Cellar/libgpg-error/1.18/include -L/usr/local/Cellar/libgpg-error/1.18/lib -lgpg-error ```

Wasn't what killed my build though.

**Comment by DomT4** [comment]
Actually going to merge here. The bottle is `cellar :any` so it doesn't suck for anyone unless they actually *choose* a from source build. The incorrect flag is clearly trickling down to other formulae that use that config executable to grab paths, and may well cause build issues in certain situations. We can review this again should it happen in future.

Thank you for the contribution @tzudot, We appreciate it! :smiley_cat:

**Comment by chdiza** [comment]
What's the plan going forward?  Are we going to bump `libgcrypt`'s revision every time `libgpg-error` gets a version bump?  It smells to me like something should be done once and for all to `libgcrypt-config`.  (I wonder if it can be made to look in the /opt dir for example.)

**Comment by DomT4** [comment]
The problem is that we do set `#{Formula["libgpg-error"].opt_prefix}"` as the path here but apparently during compile `libgcrypt` resolves that symlink and instead stores the versioned path in the `libgcrypt-config` executable:

```
# File: src/libgcrypt-config.  Generated from libgcrypt-config.in by configure.

# General.
prefix="/usr/local/Cellar/libgcrypt/1.6.3_1"
exec_prefix="${prefix}"
version="1.6.3"
includedir="${prefix}/include"
libdir="${exec_prefix}/lib"
gpg_error_libs="-L/usr/local/Cellar/libgpg-error/1.19/lib -lgpg-error"
gpg_error_cflags="-I/usr/local/Cellar/libgpg-error/1.19/include"
```

A dirty hack would be to inreplace that afterwards, or to try and tweak `libgcrypt`'s make process to stop it hunting down the true location of the symlink in `opt`.

chdiza:
Experiment reveals that `libgcrypt-config` is created simply by running the `./configure` script.  I'll see if I can figure out how to force the latter to use the `opt_prefix`.

chdiza:
Turns out the "problem" is actually in `gpg-error-config` (part of `libgpg-error`).  `libgcrypt-config` simply queries it and receives the offending path.  And the reason `gpg-error-config` returns the realpath rather than `opt_prefix` is (AFAICT) because of `libgpg-error`'s configure script.  (As with `libgcrypt-config`, `gpg-error-config` is created merely by the `./configure` step of `libgpg-error`.)

I don't know enough about configure scripts, however, to determine how to tweak the configure script belonging to `libgpg-error`.

PR #42094, opened 2015-07-24
unoconv 0.7

rddesmond:
See: https://github.com/dagwieers/unoconv/blob/0.7/CHANGELOG.md

### DIFF
--- a/Library/Formula/unoconv.rb
+++ b/Library/Formula/unoconv.rb
@@ -3,8 +3,8 @@ require "formula"
 class Unoconv < Formula
   desc "Convert between any document format supported by OpenOffice"
   homepage "http://dag.wiee.rs/home-made/unoconv/"
-  url "http://dag.wieers.com/home-made/unoconv/unoconv-0.6.tar.gz"
-  sha1 "d6c3574639b9ceedcb866b84f18ba7158f25db48"
+  url "http://dag.wieers.com/home-made/unoconv/unoconv-0.7.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/unoconv.rb
```diff
@@ -3,8 +3,8 @@
 class Unoconv < Formula
   desc "Convert between any document format supported by OpenOffice"
   homepage "http://dag.wiee.rs/home-made/unoconv/"
-  url "http://dag.wieers.com/home-made/unoconv/unoconv-0.6.tar.gz"
-  sha1 "d6c3574639b9ceedcb866b84f18ba7158f25db48"
+  url "http://dag.wieers.com/home-made/unoconv/unoconv-0.7.tar.gz"
+  sha1 "c4d764f9cd8461955ca1c3564a522449ed264c19"
```
DomT4: You'll need to use `sha256` here, Thanks!

dunn:
Is it possible to add a simple test, just to check basic functionality?

rddesmond:
Unfortunately, not a very strong one.  Unoconv needs libreoffice/openoffice
installed by other means, so a test would have to have two acceptable
outcomes -- it errors because unoconv cannot find office (yet) or it
succeeds (and perhaps, outputs a version number).

dunn:
Got it, we'll do without, then. Merged in d441dea968afc0d1fb5e7e77b54f3f92225e6695, thanks for contributing!

PR #42097, opened 2015-07-24
pull: stop printing warning on test-bot

DomT4:
The `opoo` change handles [this situation](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/30177/version=mountain_lion/testReport/junit/brew-test-bot/mountain_lion/pull___clean_https___github_com_Homebrew_homebrew_pull_42091/) where the error is printed on the test-bot where it isn't particularly relevant.

The other change is to try sleeping for 5 before the initial bottle fetch, and then falling back on another 10 sleep if that fails. Every single pull has failed to fetch on the initial command so I think we might as well give up on the "Instant fetch" idea and give it a few seconds.

### DIFF
--- a/Library/Homebrew/cmd/pull.rb
+++ b/Library/Homebrew/cmd/pull.rb
@@ -99,7 +99,7 @@ module Homebrew
       branch = `git symbolic-ref --short HEAD`.strip
 
       unless branch == "master"
-        opoo "Current branch is #{branch}: do you need to pull inside master?"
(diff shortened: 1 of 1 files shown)


DomT4:
CC @mikemcquaid

DomT4:
https://github.com/Homebrew/homebrew/commit/4f511f84825b2bae99a9f5aa4c19fb24b3523de4

PR #42102, opened 2015-07-24
influxdb v0.9.2

gunnaraasen:


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/d994284d6ad5aea220e688cd2f2b153e72d579aa. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a nit, which I didn't notice till after I pulled here heh, but we don't use the `v` for version in commit messages. Should just be `influxdb 0.9.2` or such, Thanks!

PR #42103, opened 2015-07-24
qca: fix build with Qt 5.5

UniqMartin:
This fixes the `brew install qca --with-qt5 --without-qt` build with the latest Qt 5.5.0.

Also minor adjustments for `brew audit --strict` compliance.

### DIFF
--- a/Library/Formula/qca.rb
+++ b/Library/Formula/qca.rb
@@ -1,12 +1,18 @@
-require "formula"
-
 class Qca < Formula
   desc "Qt Cryptographic Architecture (QCA)"
   homepage "http://delta.affinix.com/qca/"
   url "http://delta.affinix.com/download/qca/2.0/qca-2.1.0.tar.gz"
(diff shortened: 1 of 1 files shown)


UniqMartin:
I hope I'm not being impatient, but is there something that prevents this PR from being merged in its current form?

DomT4:
> is there something that prevents this PR from being merged in its current form?

It tends to be a little quieter Friday-Sunday. Merged in https://github.com/Homebrew/homebrew/commit/b88113cd5ed468cb2cf75c97ded0c73312fbec93, Thanks!

Just a nit, but when we create `stable do` blocks the stable URL & checksum are supposed to be moved inside them as well. I fixed that on pull here.

UniqMartin:
Thanks for the merge and for moving stuff to the `stable do` block! I'll try to keep that in mind next time I create a `stable do` block from scratch.

PR #42105, opened 2015-07-24
qt5: fix broken pkg-config files

UniqMartin:
Qt 5.5.0 introduced some quoting to its `pkg-config` files that results in invalid arguments being passed to the linker (for packages that use Qt 5 via `pkg-config`). The issue is documented [upstream in QTBUG-47162](https://bugreports.qt.io/browse/QTBUG-47162) and I added my comments and my proposed fix there.

Formulae known to be affected by the bug (they link successfully after applying the proposed fix):
- `poppler --with-qt5`
- `wireshark --with-qt5`

### DIFF
--- a/Library/Formula/qt5.rb
+++ b/Library/Formula/qt5.rb
@@ -13,6 +13,7 @@ class Qt5 < Formula
   desc "Version 5 of the Qt framework"
   homepage "https://www.qt.io/"
   head "https://code.qt.io/qt/qt5.git", :branch => "5.5", :shallow => false
+  revision 1
 
   stable do
     url "https://download.qt.io/official_releases/qt/5.5/5.5.0/single/qt-everywhere-opensource-src-5.5.0.tar.xz"
@@ -24,6 +25,16 @@ class Qt5 < Formula
     # This is fixed in 5.5 branch and below patch should be removed
     # when this formula is updated to 5.5.1
     patch :DATA
+  end
+
+  # `qmake` generates broken `pkg-config` files with quoted '-framework QtCore'
+  # (and similarly for all other frameworks), that remains a single argument in
+  # projects that use Qt via `pkg-config` and thus causes breakage when linking
+  # ('-framework' and 'QtCore' need to be passed as separate arguments).
+  # https://bugreports.qt.io/browse/QTBUG-47162
+  patch do
+    url "https://gist.githubusercontent.com/UniqMartin/a54542d666be1983dc83/raw/f235dfb418c3d0d086c3baae520d538bae0b1c70/qtbug-47162.patch"
+    sha256 "e31df5d0c5f8a9e738823299cb6ed5f5951314a28d4a4f9f021f423963038432"
   end
 
   bottle do


DomT4:
Heh, you beat me to this. I've spent a couple of hours debugging Wireshark's failures this evening.

UniqMartin:
I found the problem surprisingly quickly after having issues building Poppler, but then spent waiting half a day for Qt 5 (more than three hours) and a few dependent formulae to recompile, in an attempt to make sure nothing broke in the process …

(Poor bot that will have to crunch through this just for a few removed quotes.)

DomT4:
Let's add a `revision` here actually. It's going to suck, but the `pkg-config` problem is laughably bad and anything which taps the `Framework` line of those files is going to break more often than not.

```
clang: error: unknown argument: '-framework QtCore'
clang: error: unknown argument: '-framework QtGui'
clang: error: unknown argument: '-framework QtCore'
clang: error: unknown argument: '-framework QtWidgets'
clang: error: unknown argument: '-framework QtGui'
clang: error: unknown argument: '-framework QtCore'
clang: error: unknown argument: '-framework QtPrintSupport'
clang: error: unknown argument: '-framework QtWidgets'
clang: error: unknown argument: '-framework QtGui'
clang: error: unknown argument: '-framework QtCore'
clang: error: unknown argument: '-framework QtMultimediaWidgets'
clang: error: unknown argument: '-framework QtMultimedia'
clang: error: unknown argument: '-framework QtNetwork'
clang: error: unknown argument: '-framework QtWidgets'
clang: error: unknown argument: '-framework QtGui'
clang: error: unknown argument: '-framework QtCore'
clang: error: unknown argument: '-framework QtMacExtras'
clang: error: unknown argument: '-framework QtGui'
clang: error: unknown argument: '-framework QtCore'
make[2]: *** [wireshark] Error 1
make[2]: *** Waiting for unfinished jobs....
```

So fun.

UniqMartin:
Yes, exactly the same errors I was seeing with Poppler, only not as numerous.

I bumped the revision as you suggested. It felt a bit weird not to be able to put `revision` inside the `stable do` block. (Maybe kill/unqueue the job that still refers to the old commit?)

DomT4:
Dumped a comment on the bug tracker in the hope throwing some Homebrew at it will make Qt fix the problem a little quicker, they can be kind of slow accepting & fixing problems upstream :/.

Yeah, our `revision`s work on a formula-wide basis. Have chopped the old build off the queue.

DomT4:
If we don't get any upstream response to the proposed patch by the end of tomorrow I'll merge this. If necessary once they commit a fix we can change the patch, but this seems to resolve the fairly awful issue. That they've tagged this P3 doesn't fill me with hope.

UniqMartin:
I sure wouldn't mind getting this merged, but I think it's quite unlikely to get a response from the official maintainer during the weekend. I also wouldn't be surprised if the priority gets changed, now that the issue is (hopefully) clearer and more people have chimed in.

Thanks for joining the upstream bug with the added gravity of a Homebrew maintainer!

DomT4:
Heh, sorry, I was being very pedantic about time; by tomorrow I meant Monday. Was just after 1am local time Sunday morning when I posted that comment :wink:.

> with the added gravity of a Homebrew maintainer!

Probably very microgravity. Sometimes it's useful to throw it in there as *"I help maintain xyz, and currently there's thousands of users who have this problem with the Qt5 installations on their system"* rather than *"I have this problem too"*, but eh.

DomT4:
Alright. Not thrilled leaving this this broken any longer, so merged in https://github.com/Homebrew/homebrew/commit/0873e46f6486bfd853c2bdd5e754ec44e95b4fcc. Will keep an eye on the upstream situation; we can tweak the application of the patch or use another later on if necessary, but since this is actively breaking compiles, waiting isn't great.

Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

UniqMartin:
Thanks for the merge! I would have hoped to at least see a confirmation from upstream, but well …

PR #42107, opened 2015-07-25
bear: build a univeral binary

smoofra:
Bear works by injecting its dylib into arbitrary processes.  Some of those
processes are likely to be 32-bit.

### DIFF
--- a/Library/Formula/bear.rb
+++ b/Library/Formula/bear.rb
@@ -15,6 +15,7 @@ class Bear < Formula
   depends_on "cmake" => :build
 
   def install
+    ENV.universal_binary
     mkdir "build" do
       system "cmake", "..", *std_cmake_args
       system "make", "install"


smoofra:
bump

mikemcquaid:
This should be optional, thanks.

smoofra:
Even though it's required for bear to do it's job?

It doesn't seem like something the user should have to know about or opt-in to get.   

Is there a way to make it an option but be on by default?

mikemcquaid:
OS X builds 64-bit code (not 32-bit or universal) by default now and has for quite some while.

bfontaine:
Ping re Mike’s comment above.

smoofra:
The issue is that if a build uses a 32 bit binary for some random reason, and libbear isn't built universal, then the build is going to fail under bear, and it's going to be annoying and opaque to the user as to why that happened.     64 bit may be the default, but there's still 32 bit stuff bopping around for whatever reason, and sometimes people use them in their makefiles.

mikemcquaid:
@smoofra That may be the case but it's not how we handle other universal packages so an option would be good, thanks.

bfontaine:
Ping re Mike’s comment?

smoofra:
hrm, ok i'll update the patch.

tdsmith:
Thanks for your work here; please feel free to submit a new pull request against Homebrew/homebrew-core if you'd like to move forward with this.

PR #42113, opened 2015-07-25
gambit-scheme 4.7.7

nijikon:


nijikon:
Can we reschedule? This should pass and take significantly less time to build.

DomT4:
Failed again :confused:. Builds in 10 minutes on Yosemite, but Mavericks and ML just hang.

nijikon:
@mikemcquaid any idea what might be happening? Can we get any log why it got stuck?

mikemcquaid:
No idea I'm afraid :(

nijikon:
Closed in favor of #43407

PR #42114, opened 2015-07-25
docker-swarm 0.3.0

haihappen:


Library/Formula/docker-swarm.rb
```diff
@@ -32,6 +32,6 @@ def install
 
   test do
     output = shell_output(bin/"docker-swarm --version")
-    assert output.include? "swarm version 0.2.0 (HEAD)"
+    assert output.include? "swarm version 0.3.0 (HEAD)"
```
dunn: This is why the tests failed:
```        Error: 1 problem in 1 formula
docker-swarm:
```
It's a small thing, but instead we like to do `assert_match "swarm version #{version}", output`. It makes assertion failures easier to understand. Thanks!

haihappen: Done :)

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/f6001583cf9f685695c6b1d4b3e025a3b8a95aa6. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #42117, opened 2015-07-25
freexl 1.0.2

nijikon:


Library/Formula/freexl.rb
```diff
@@ -27,4 +27,8 @@ def install
       doc.install "html"
     end
   end
+
+  test do
+    assert File.exist? "#{lib}/libfreexl.a"
```
xu-cheng: I think we should avoid check file existing in the test. They are basically meaningless.

nijikon: How we should test this then?

xu-cheng: Can you write a simple C program? You can see `tinyxml2` for example.

nijikon: No problem, so we would also have to fix some more packages that have this kind of tests.

xu-cheng: >  so we would also have to fix some more packages that have this kind of tests.

That would be extremely appreciated.

DomT4:
@BrewTestBot test this please

DomT4:
Presuming it passes again I'll merge it & remove the `exist` test here. The version bump is something that we try not to hold up too much, but if you wanted to chase in the merge with a new PR for a new, shiny test I certainly wouldn't *object* :wink:.

DomT4:
I meant to remove the test prior to merge, rather than after, but I'm tired and apparently an idiot :smiley:.

Merged in https://github.com/Homebrew/homebrew/commit/79a3233459efe1f48b8c61d6b19833e14f087d3a. Thanks @nijikon! :)

PR #42118, opened 2015-07-25
libpng: add head

vszakats:


mistydemeo:
I'd prefer to keep using the release tarballs so we don't need autotools to build the stable releases.

vszakats:
Noted. The primary `ftp://` URL is scary looking; I thought it'd be safer to tap the original vanilla source, it'd also work from behind firewalls.

mistydemeo:
The ftp URL should be temporary; this is only until they're able to start uploading to Sourceforge again.

vszakats:
Sourceforge is no less scary :)

mistydemeo:
They haven't started tampering with source tarballs, and we use Sourceforge in plenty of places. ;)

vszakats:
I've modified the patch to add head with build dependencies moved there.

mistydemeo:
Thanks!

dunn:
Merged as a194f141e315f268abe427f10c81f451c731d26c, thanks!

PR #42126, opened 2015-07-26
embulk 0.6.19

cosmo0920:


dunn:
Let's use `write_jar_script` instead of `write_exec_script`; see [`htmlcompressor.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/htmlcompressor.rb) for an example of how that works.

And in the test block, would you mind changing `system "embulk"` to `system bin/"embulk"` on each line? Thanks!

cosmo0920:
Thanks, reviewing dunn! I've reflected your advise.

dunn:
Thank you! Merged in 1bac057b4398698a77f48eec2a301243249a56b4.

cosmo0920:
Thanks!

PR #42127, opened 2015-07-26
pidgin: revise description to reflect GUI inclusion

apjanke:
Updates `pidgin` description to reflect that the formula does include a GUI by default.

Old description called it a "GUI-less chat client". But it does have a GUI, at least by default.

Also, the old description for `--without-gui` said "Build Finch instead of Pidgin", which implies that Finch is not built in the default case. The default is to build both Finch (CLI) and Pidgin (GUI). Revised description to indicate `--without-gui` is just disabling Pidgin, and isn't required to get Finch to build.

### DIFF
--- a/Library/Formula/pidgin.rb
+++ b/Library/Formula/pidgin.rb
@@ -1,5 +1,5 @@
 class Pidgin < Formula
-  desc "GUI-less chat client (e.g., Finch-only)"
+  desc "Multi-protocol chat client"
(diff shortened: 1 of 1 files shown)


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/982cd5fa6618a61694860507fe823f20ed47b92e. Thanks @apjanke! Well spotted on the desc.

PR #42129, opened 2015-07-26
bash 4.3.39

dunn:


Library/Formula/bash.rb
```diff
@@ -1,19 +1,20 @@
...
-    version "4.3.33"
+    version "4.3.39"
 
     # Vendor the patches. The mirrors are unreliable for getting the patches,
     # and the more patches there are, the more unreliable they get. Upstream
     # patches can be found in: http://git.savannah.gnu.org/cgit/bash.git
     patch do
-      url "https://gist.githubusercontent.com/jacknagel/c1cf23775c774e2b4b6d/raw/838118bf0e07dcde8a552cb8556600b7e453d4ac/bash-4.3.33.diff"
-      sha1 "71fc36bced0d15a2c221cfcbff02ea412d4bf5fa"
+      url "https://gist.githubusercontent.com/anonymous/4282245753d5348e438b/raw/8402dc95200af1edcad1528d61d07d7003301c4b/bash-4.3.39.diff"
```
DomT4: Would prefer to keep this attached to a maintainer name if possible. Helps the legitimacy side of it, less like we've accepted a mega-patch from a random contributor in blind trust.

mikemcquaid: It'd also be worth considering just building from Git instead; makes it much easier to maintain.

dunn: Oh, ok. I made it anonymous to make sure I can't accidentally delete it 👻

dunn: @mikemcquaid like so?
```ruby
  url "http://git.savannah.gnu.org/r/bash.git",
      :revision => "d2744e05f24d1116ae8dc012e4b07ff219d0d2c1"
  version "4.3.39"
```
That's simpler but the repository is 100MB 😬

DomT4: Savannah's git can be a little up-and-down, which may be why we've avoided doing so so far? If it's less grumpy these days, turning it into a git clone could be ideal.

mikemcquaid: No official or semi-official GitHub mirror? Can we build from its tarball?

dunn: Great point.

dunn:
Changed to the git snapshot, but we'll want to add a mirror.

DomT4:
Yeah, It's very likely the snapshot checksum will change at some point when they get auto-regenerated or such.

dunn:
Oh, dang. Would we be better off with the big patch, then?

DomT4:
My preference would be something like:

* Git checkout that isn't prone to falling over occasionally (Github mirror, maybe, if there is one).
* Big patch
* Snapshot URL. 

If the bottle was `cellar :any` I'd be less worried about the git checkout being semi-unreliable, but alas it isn't.

dunn:
The snapshot SHA changed between my local build and the test bot downloading it, so I think that's out.

dunn:
Since the bottle created can't be relocated, building from the git repository is probably the wrong choice; returning to the patch method.

DomT4:
Yeah, seems like the best option we have for now, sadly.

DomT4:
Feel free to :shipit: here whenever you're ready.

PR #42131, opened 2015-07-26
Update imagemagick.rb

leoj3n:
http://www.imagemagick.org/download/releases

### DIFF
--- a/Library/Formula/imagemagick.rb
+++ b/Library/Formula/imagemagick.rb
@@ -1,9 +1,9 @@
 class Imagemagick < Formula
   desc "Tools and libraries to manipulate images in many formats"
   homepage "http://www.imagemagick.org"
-  url "http://www.imagemagick.org/download/releases/ImageMagick-6.9.1-9.tar.xz"
-  mirror "http://ftp.nluug.nl/ImageMagick/ImageMagick-6.9.1-9.tar.xz"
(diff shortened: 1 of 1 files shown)


chdiza:
Is it really necessary for HB to update `imagemagick` at every single *sub-micro* patch, like we've been doing lately?  We used to not do this.

Most sub-micro patches for `imagemagick`, if one looks at the changelog, contain only one or two individual patches.  The present case contains only two.  6.9.1-9 contained only one patch atop 6.9.1-8.  In other words, updating `imagemagick` at every sub-micro release is really close to making everyone use HEAD.

Perhaps we should only bump when a sub-micro release contains a greater number of patches, or perhaps we should only bump at every new micro release.  Users who want the absolute bleeding edge have can use `--HEAD`.

DomT4:
@chdiza Yes. Check both the URLs, the old link has gone from both and does so consistently.

mikemcquaid:
@DomT4 Previously the solution was to mirror these. We could/should do that on BinTray.

DomT4:
Will look into it. If anyone wants to leap over my head with a PR, I've no objection.

For now, merged in https://github.com/Homebrew/homebrew/commit/df4c6e7c0e3f7c03f9a9f98ccbd781d1a8f741e8. Thank you for your contribution @leoj3n!

PR #42132, opened 2015-07-26
Update imagemagick.rb

dmarkrollins:
Looks like they just updated the version yesterday.


DomT4:
Closing in favour of #42132, which slightly predates this PR. Thanks for the submission though, we appreciate that!

PR #42140, opened 2015-07-26
dub 0.9.24-rc2

John-Colvin:


Library/Formula/dub.rb
```diff
@@ -10,6 +10,12 @@ class Dub < Formula
     sha256 "4afe785f25db496f8634c19211d68c7ea05778f9ed1e7e772f8aa9ac701beca1" => :mountain_lion
   end
 
+  devel do
+      url "https://github.com/D-Programming-Language/dub/archive/v0.9.24-rc.2.tar.gz"
+      version "0.9.24-rc2"
+      sha1 "43fbdbca2313b997b219c50e13ff43742ae0d302"
```
dunn: Could you update all the `sha1`s to `sha256`, please? Thanks!

John-Colvin: done. I presume this is the new standard and I should do this for all the formulae I maintain?

DomT4: Yup. SHA1 is now formally deprecated.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/7baf0e7c3fe8e5815c6fa0b3530e74912c3c665d. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

Just a nit, but commit messages for devel version updates should be tagged with `(devel)` after the version, like `ducks 1.0.1 (devel)`, Thanks!

PR #42168, opened 2015-07-27
jprefctl 0.9.3

canbican:


DomT4:
I'm afraid this is going to be too new for us a this point in time. This would have been flagged by our [audit](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/audit.rb#L462) after installation, but the special dependency of an older Java meant our CI skipped that stage.

You should be able to see the error locally if you do `brew audit --strict --online jprefctl`.

We suggest hosting it in your own Homebrew tap for now, such as `canbican/homebrew-jprefctl` - Just drop your proposed formula in there, and then you'll be able to do `brew install canbican/jprefctl/jprefctl`.

Thanks!

mikemcquaid:
> I'm afraid this is going to be too new for us at this point in time. This would have been flagged by our audit after installation, but the special dependency of an older Java meant our CI skipped that stage.

@DomT4 We should probably still audit formulae if we skip them based on a `Requirement`.

DomT4:
Yeah, we currently do audits after installation on the test-bot, but we probably need to do one beforehand as well. Do we want to just move the audit pre-installation? I presume the [formula_cellar_checks](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/formula_cellar_checks.rb) still run even if we audit pre-installation?

mikemcquaid:
@DomT4 Moving it pre-installation will stop the (useful) post-installation checks from running. We probably want to only do so in the skip case.

PR #42175, opened 2015-07-27
wgetpaste: add missing wget dependency

cardoe:
wgetpaste needs wget so add it as a dependency


cardoe:
This looks like a Jenkins failure to communicate with GitHub. Its complaining about not being able to get the HEAD ref.

dunn:
Merged in 0dabc85232d9a642b2df096817f383854d4135c0, thanks for catching that!

PR #42176, opened 2015-07-27
midnight-commander: Fix `man` invocation

petere:
mc calls `man -l` to display man page files, which is not correct on OS X, so patch out the `-l` to make it work.

### DIFF
--- a/Library/Formula/midnight-commander.rb
+++ b/Library/Formula/midnight-commander.rb
@@ -28,6 +28,7 @@ class MidnightCommander < Formula
                           "--with-screen=slang",
                           "--enable-vfs-sftp"
     system "make", "install"
+    inreplace libexec/"mc/ext.d/text.sh", "man -P cat -l ", "man -P cat "
   end
 
   test do


Library/Formula/midnight-commander.rb
```diff
@@ -28,6 +28,7 @@ def install
                           "--with-screen=slang",
                           "--enable-vfs-sftp"
     system "make", "install"
+    inreplace libexec/"mc/ext.d/text.sh", "man -P cat -l ", "man -P cat "
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

DomT4: Ping on this.

petere:
The upstream bug tracker is rejecting my ticket submissions as spam.

Meanwhile, I found a patch for the same issue in FreeBSD: https://svnweb.freebsd.org/ports/head/misc/mc/files/patch-misc__ext.d__text.sh.in?view=markup&pathrev=382039

petere:
Ah, my ticket made it: http://www.midnight-commander.org/ticket/3509

mikemcquaid:
@petere Good news. Can you please add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

DomT4:
@BrewTestBot test this please

DomT4:
Added the URL in manually after merge. Merged in https://github.com/Homebrew/homebrew/commit/49dca0468426d2c5889132c8b3fe6375fd330c3f, Thanks!

PR #42178, opened 2015-07-27
Add option for true-colors in tmux

mhartington:
This add the option to install tmux with a true color. 

```bash
brew install tmux --with-truecolor
```

Small addition, but useful for those who have terminals that [support true-color](https://gist.github.com/XVilka/8346728#now-supporting-truecolor)

### DIFF
--- a/Library/Formula/tmux.rb
+++ b/Library/Formula/tmux.rb
@@ -23,6 +23,12 @@ class Tmux < Formula
   depends_on "pkg-config" => :build
   depends_on "libevent"
 
+  option "with-truecolor", "Build with truecolor patch enabled"
+  patch do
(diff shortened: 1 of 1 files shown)


mistydemeo:
We tend not to accept unofficial patches. Is this something that's been submitted to tmux?

mhartington:
Yes, but they seem unwilling/hesitant to add the feature. 

https://github.com/tmux/tmux/issues/34#issuecomment-116167810

Again, this would be an option, so the default install wouldn't come with it.
It's small step and may encourage tmux to add this feature.

No big deal if it's not added though, got to try :grin:

mikemcquaid:
We're going to pass if `tmux` isn't interested, I'm afraid. Sorry!

mhartington:
No worries, thanks for looking

PR #42179, opened 2015-07-27
bro: Don't clean empty directories and symlinks

JustinAzoff:
The spool and logs directories are initially empty, and the
broctl-config.sh is initially a broken symlink until broctl runs for the
first time.

### DIFF
--- a/Library/Formula/bro.rb
+++ b/Library/Formula/bro.rb
@@ -22,6 +22,12 @@ class Bro < Formula
   depends_on "geoip" => :recommended
   depends_on "openssl"
 
+  # Initially empty directories
+  skip_clean "spool"
+  skip_clean "logs"
+  # Initially a broken symlink
+  skip_clean "share/broctl/scripts/broctl-config.sh"
+
   def install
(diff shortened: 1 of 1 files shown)


Library/Formula/bro.rb
```diff
@@ -22,6 +22,12 @@ class Bro < Formula
   depends_on "geoip" => :recommended
   depends_on "openssl"
 
+  # Initially empty directories
+  skip_clean "spool"
+  skip_clean "logs"
```
DomT4: `spool` and `logs` would typically belong in `var` so they wouldn't be removed anyhow. Where are they stored now?

JustinAzoff: Bro doesn't create a var/ directory anywhere by default, just $prefix +

    bin/     etc/     include/ lib/     logs/    share/   spool/

I suppose it does make sense for logs/ and spool/ to be in var/, but the fix for that is a little more involved.

mikemcquaid: The more involved fix would be better :+1:

PR #42180, opened 2015-07-27
update caveat to fix match project README

aristiden7o:
see PR for the project README: 
https://github.com/yyuu/pyenv-virtualenv/pull/93


Library/Formula/pyenv-virtualenv.rb
```diff
@@ -15,6 +15,7 @@ def install
 
   def caveats; <<-EOS.undent
     To enable auto-activation add to your profile:
+      eval "$(pyenv init -)"
```
xu-cheng: I don't this is necessary as it will show in the caveats in `pyenv`.

aristiden7o: good call! I'm closing this

PR #42184, opened 2015-07-28
cake: 0.5.1 - C# Make (new formula)

Redth:
This adds a new formula for [cake](http://cakebuild.net) (C# Make).  The NuGet package is an existing feed of binary distributions for the project so it is used.  The executable bin installed is a bash wrapper around the `mono Cake.exe [args]` command, meaning simply `cake ./build.cake` can be invoked.  The package depends conditionally on mono.  Some software installations such as Xamarin already ensure mono is installed and kept up to date, so mono should not be installed through homebrew if it already exists from another installation.  This is determined by invoking `which mono` and looking at the output (if nothing is found, mono through homebrew is required.

Tests are included which 1) run cake to make sure there is an ok exit code, and 2) build a simple script to run and check the output of.

### DIFF
--- a/Library/Formula/cake.rb
+++ b/Library/Formula/cake.rb
@@ -1,0 +1,40 @@
+class Cake < Formula
+  desc "Cake (C# Make) is a build automation system with a C# DSL."
+  homepage "http://cakebuild.net/"
+  url "https://github.com/cake-build/cake/releases/download/v0.5.1/Cake-bin-v0.5.1.zip"
+  sha256 "ff431d845575b0be9e355caa9f44a82ea015d8b849eeafcdaee0904c3fb6da53"
+
+  # Xamarin will maintain a current mono install, may not actually need it
+  depends_on "mono" => :recommended
+
+  def install
+    libexec.install Dir["*.dll"]
+    libexec.install Dir["*.exe"]
+    libexec.install Dir["*.xml"]
+
+    # write a simple bash wrapper to execute `mono Cake.exe [args]` with the cake alias
+    bin.mkpath
+    (bin/"cake").write <<-EOS.undent
+      #!/bin/bash
+      mono #{libexec}/Cake.exe "$@"
+    EOS
+  end
+
+  test do
+    test_str = "Hello Homebrew"
+    test_name = "build.cake"
+    (testpath/test_name).write <<-EOS.undent
+
+      var target = Argument ("target", "info");
+
+      Task("info").Does(() =>
+      {
+        Information ("Hello Homebrew");
+      });
+
+      RunTarget ("info");
+
+    EOS
+    assert_match test_str, shell_output("#{bin}/cake ./build.cake").strip
+  end
+end


Library/Formula/cake.rb
```diff
@@ -0,0 +1,47 @@
+class Cake < Formula
+  desc "Cake (C# Make) is a build automation system with a C# DSL."
+  homepage "http://cakebuild.net/"
+  url "https://www.nuget.org/api/v2/package/Cake/#{version}"
```
DomT4: We don't use `#{version}"` in URLs. Should just be `https://www.nuget.org/api/v2/package/Cake/0.5.1`, Thanks.

Library/Formula/cake.rb
```diff
@@ -0,0 +1,47 @@
...
+  version "0.5.1"
+  sha256 "e3620f9bb1466d47502aeb934868284e241877ad5f43531f898443d846bd7c0b"
+
+  # Xamarin (and possibly other products) will maintain a current mono installation
+  # so we may not need to actually install it, so only depend on it if necessary
+  depends_on "mono" => :recommended
+
+  def install
+    prefix.install Dir["*.dll"]
```
DomT4: All this sort of thing belongs in `libexec` really.

Library/Formula/cake.rb
Redth: Fixed, thanks

Redth:
@DomT4 moved the assert and shell_output to a single line, anything else that needs to be changed for this PR?

Library/Formula/cake.rb
```diff
@@ -0,0 +1,46 @@
...
+    bin.mkpath
+    (bin/"cake").write <<-EOS.undent
+      #!/bin/bash
+      mono #{libexec}/Cake.exe "$@"
+    EOS
+  end
+
+  test do
+    # just run cake to see the version, if we have an ok exit code it worked
+    system bin/"cake", "-version"
```
DomT4: Let's scrap this check. If the script below fails the test will exit as a failure.

Library/Formula/cake.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    # write a temporary script and run it, check the output
```
DomT4: No need for the comment here :)

Library/Formula/cake.rb
```diff
@@ -0,0 +1,46 @@
+class Cake < Formula
+  desc "Cake (C# Make) is a build automation system with a C# DSL."
+  homepage "http://cakebuild.net/"
+  url "https://www.nuget.org/api/v2/package/Cake/0.5.1"
+  version "0.5.1"
```
DomT4: What `version` does it think this package is without this line?

Library/Formula/cake.rb
```diff
@@ -0,0 +1,46 @@
+class Cake < Formula
+  desc "Cake (C# Make) is a build automation system with a C# DSL."
+  homepage "http://cakebuild.net/"
+  url "https://www.nuget.org/api/v2/package/Cake/0.5.1"
```
DomT4: Can we use `https://github.com/cake-build/cake/releases/download/v0.5.1/Cake-bin-v0.5.1.zip` or something instead? Looks like it installs the same content.

Redth:
@DomT4 I used the releases url as you suggested (otherwise homebrew was thinking it was version 0.5 instead of 0.5.1 with the old url).  I removed the explicit version declaration as well as it's parsed properly now.

Removed the first test as requested, and the excessive comment.

Anything else? Ready for merge? Thanks!

Redth:
Just to note, I do believe the Jenkins failure has nothing to do with my `cake.rb` which passes the strict audit and tests locally.... So hoping it can be merged soon :)

DomT4:
The ML test-bot is being grumpy. I've asked it to check again.

Redth:
@DomT4 thanks looks like its passing now :)

DomT4:
I'm happy enough. Merged in  https://github.com/Homebrew/homebrew/commit/fbe4ea8609e8f472cd3ab94d6ae7f46cddf340f6. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #42190, opened 2015-07-28
move creating .zshrc to interactive_shell method

xu-cheng:
cc @DomT4


DomT4:
Seems to work as intended locally :+1:.

PR #42192, opened 2015-07-28
pidgin: set tcl/tk config paths regardless of whether building gui

### LINKED ISSUE #41968: pidgin failed to build on 10.11 (Tcl development headers missing)
The `pidgin` formula fails to build on 10.11:

```
==> Installing pidgin
==> Downloading https://downloads.sourceforge.net/project/pidgin/Pidgin/2.10.11/
Already downloaded: /Library/Caches/Homebrew/pidgin-2.10.11.tar.bz2
==> ./configure --prefix=/usr/local/Cellar/pidgin/2.10.11_1 --disable-avahi --di
checking for tclConfig.sh... no
configure: error:
Tcl development headers not found.
Use --disable-tcl if you do not need Tcl scripting support.
```

Presumably this means Tcl headers are no longer shipped with 10.11.

`brew gist-logs pidgin`: https://gist.github.com/3358fb122334a3bb5890

apjanke:
Sets the Tcl/Tk configuration paths regardless of whether the GUI is being built or not. This allows the build including the GUI to work on systems that don't have the CLT installed.

Fixes #41968

Tck/Tk is used at the lower purple protocol library level, so it's used by both Finch and Pidgin and needed regardless of whether you're building the GUI or not. The fact that it's currently only being configured for the `without-gui` option seems to be a historical quirk of how the `finch` and `pidgin` formulae were merged in 7537f3663d0ca35e5848d0a9644ebc86c92d267f: the `finch` formula was designed to work without CLT, and `pidgin` wasn't; it's not an inherent requirement of on but not the other.

####   Testing Notes   ####

Tested on 10.9.5, 10.10, 10.11, all without CLT installed. Succeeds on 10.10 and 10.11. Succeeds on 10.9 if CLT is installed.

With this patch, `brew install --build-from-source pidgin` will still fail on OS X 10.9 boxes that do not have the CLT installed, citing a Python link error. I think this is an unrelated error due to how the OS X 10.9 SDK lacks Python. It is not a regression: `brew install --build-from-source` on 10.9 w/o CLT fails for me with the current `pidgin` formula definition, too.

### DIFF
--- a/Library/Formula/pidgin.rb
+++ b/Library/Formula/pidgin.rb
@@ -63,9 +63,9 @@ class Pidgin < Formula
     args << "--disable-perl" if build.without? "perl"
     args << "--enable-cyrus-sasl" if build.with? "gsasl"
 
+    args << "--with-tclconfig=#{MacOS.sdk_path}/usr/lib"
+    args << "--with-tkconfig=#{MacOS.sdk_path}/usr/lib"
     if build.without? "gui"
-      args << "--with-tclconfig=#{MacOS.sdk_path}/usr/lib"
-      args << "--with-tkconfig=#{MacOS.sdk_path}/usr/lib"
       args << "--disable-gtkui"
     else
       args << "--disable-idn"


DomT4:
No harm in adopting this, even if it isn't completely CLT-proof at this point. Merged in https://github.com/Homebrew/homebrew/commit/7f4700077decbd1ac6129adf2ecf5b67cc63f814, Thanks @apjanke.

Just a nit, but commit messages should be under 50 characters or so ideally.

apjanke:
Oh, sorry about that! I thought it was just the first paragraph that was supposed to be short. I'll keep it concise in the future.

PR #42189, opened 2015-07-28
harfbuzz 1.0.1

lembacon:

PR #42196, opened 2015-07-28
harfbuzz: update to 1.0.0

alerque:


alerque:
Jinxed by #42189. It took me a day to download new Xcode tools and get homebrew cooperating on El Capitan and that put me a bit behind the curve!

PR #42202, opened 2015-07-28
webarchiver (new formula)

smoofra:
Webarchiver allows you to create Safari .webarchive files from the command line. Webarchives are a convenient way to store a webpage and its associated files (images, css, javascript, etc) in a single file. It is very simple to use:

./webarchiver -url http://www.google.com -output google.webarchive

### DIFF
--- a/Library/Formula/webarchiver.rb
+++ b/Library/Formula/webarchiver.rb
@@ -5,6 +5,8 @@ class Webarchiver < Formula
   sha256 "8ea826038e923c72e75a4bbb1416910368140a675421f6aaa51fd0dea703f75c"
   head "https://github.com/newzealandpaul/webarchiver.git"
 
+  depends_on :macos => :mavericks
+
   def install
     xcodebuild
     bin.install "./build/Release/webarchiver"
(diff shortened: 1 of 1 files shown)


bfontaine:
The build fails on ML: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/30371/version=mountain_lion/testReport/junit/brew-test-bot/mountain_lion/install___build_from_source_webarchiver/

smoofra:
looks like its not compatible with ML, so I marked that in the formula

mikemcquaid:
If possible guarding to the Xcode version would be preferable.

smoofra:
ok.  fixed.

Library/Formula/webarchiver.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  depends_on :xcode => ["6.0.1", :build]
+
+  def install
+    xcodebuild
+    bin.install "./build/Release/webarchiver"
+  end
+
+  test do
+    mktemp do
```
mikemcquaid: You can remove this; `test do` is already run in a temporary directory.

smoofra: done

mikemcquaid:
Thanks again @smoofra!

PR #42205, opened 2015-07-28
ARGV#resolved_formulae: auto resolve spec

### LINKED ISSUE #42147: brew reinstall doesn't keep HEAD flag?
It seems that `brew reinstall` keeps all flags used except for `--HEAD`. For example, if I had an `emacs --with-x11 --HEAD` install, and ran `brew reinstall emacs`, I'd end up with a plain `emacs --with-x11` install.

xu-cheng:
For `ARGV#resolved_formulae` and `Formulary#from_rack`, it will try to resolve the formula's spec when:
* Formula is installed.
* No `--devel` or `--HEAD` in ARGV.

The codes still need some work. cc @Homebrew/owners for review.

Fixes #42147

### DIFF
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -20,9 +20,14 @@ module HomebrewArgvExtension
     require "formula"
     @resolved_formulae ||= (downcased_unique_named - casks).map do |name|
       if name.include?("/")
-        Formulary.factory(name, spec)
+        f = Formulary.factory(name, spec)
+        if spec(default=nil).nil? && f.any_version_installed?
+          installed_spec = Tab.for_formula(f).spec
+          f.set_active_spec(installed_spec) if f.send(installed_spec)
+        end
+        f
       else
-        Formulary.from_rack(HOMEBREW_CELLAR/name, spec)
+        Formulary.from_rack(HOMEBREW_CELLAR/name, spec(default=nil))
       end
     end
   end
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -190,13 +190,13 @@ module HomebrewArgvExtension
 
   private
 
-  def spec
+  def spec(default=:stable)
     if include?("--HEAD")
       :head
     elsif include?("--devel")
       :devel
     else
-      :stable
+      default
     end
   end
 
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -67,9 +67,6 @@ class Formula
   # @see #active_spec
   attr_reader :active_spec_sym
 
-  # The {PkgVersion} for this formula with version and {#revision} information.
-  attr_reader :pkg_version
-
   # Used for creating new Homebrew versions of software without new upstream
   # versions.
   # @see .revision
@@ -118,7 +115,6 @@ class Formula
       :stable
     end
     validate_attributes!
-    @pkg_version = PkgVersion.new(version, revision)
     @build = active_spec.build
     @pin = FormulaPin.new(self)
   end
@@ -205,6 +201,11 @@ class Formula
   # @see .version
   def version
     active_spec.version
+  end
+
+  # The {PkgVersion} for this formula with {version} and {#revision} information.
+  def pkg_version
+    PkgVersion.new(version, revision)
   end
 
   # A named Resource for the currently active {SoftwareSpec}.
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -274,6 +274,12 @@ class Formula
     (dir = installed_prefix).directory? && dir.children.length > 0
   end
 
+  # If at least one version of {Formula} is installed.
+  def any_version_installed?
+    require "tab"
+    rack.directory? && rack.subdirs.any? { |keg| (keg/Tab::FILENAME).file? }
+  end
+
   # @private
   # The `LinkedKegs` directory for this {Formula}.
   # You probably want {#opt_prefix} instead.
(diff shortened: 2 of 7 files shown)


xu-cheng:
I reimplement this PR with more code refactoring and test case.

Library/Homebrew/extend/ARGV.rb
```diff
@@ -190,13 +195,13 @@ def env
...
 
-  def spec
+  def spec(default=:stable)
     if include?("--HEAD")
       :head
     elsif include?("--devel")
       :devel
     else
-      :stable
+      default
```
mikemcquaid: Might want to check this is a symbol or something? I dunno.

Library/Homebrew/formula.rb
```diff
@@ -62,8 +62,10 @@ class Formula
   attr_reader :active_spec
   protected :active_spec
 
-  # The {PkgVersion} for this formula with version and {#revision} information.
-  attr_reader :pkg_version
+  # A symbol to indecate currently active {SoftwareSpec}.
```
mikemcquaid: `indicate`

Library/Homebrew/formula.rb
```diff
@@ -105,12 +107,22 @@ def initialize(name, path, spec)
     set_spec :head
 
     @active_spec = determine_active_spec(spec)
+    @active_spec_sym = if head? then :head elsif devel? then :devel else :stable end
```
mikemcquaid: Can you split this onto multiple lines? Makes it easier to read.

Library/Homebrew/formula.rb
```diff
@@ -105,12 +107,22 @@ def initialize(name, path, spec)
...
     validate_attributes!
-    @pkg_version = PkgVersion.new(version, revision)
     @build = active_spec.build
     @pin = FormulaPin.new(self)
   end
 
+  # @private
+  def set_active_spec(requested)
+    spec = send(requested)
+    raise FormulaSpecificationError, "formulae require at least a URL for spec #{requested}" unless spec
```
mikemcquaid: Hmm, we don't know that this was definitely the reason it failed, right?

Library/Homebrew/formula.rb
```diff
@@ -105,12 +107,22 @@ def initialize(name, path, spec)
...
     @active_spec = determine_active_spec(spec)
+    @active_spec_sym = if head? then :head elsif devel? then :devel else :stable end
     validate_attributes!
-    @pkg_version = PkgVersion.new(version, revision)
     @build = active_spec.build
     @pin = FormulaPin.new(self)
   end
 
+  # @private
+  def set_active_spec(requested)
```
mikemcquaid: Not obvious what `requested` means here, a better variable name would be nice.

Library/Homebrew/formula.rb
```diff
@@ -261,6 +278,12 @@ def installed?
     (dir = installed_prefix).directory? && dir.children.length > 0
   end
 
+  # If at least one version of {Formula} is installed.
+  def any_keg?
```
mikemcquaid: `any_version_installed?`?

Library/Homebrew/test/test_tab.rb
```diff
@@ -75,6 +76,7 @@ def test_from_file
     refute_predicate tab, :built_as_bottle
     assert_predicate tab, :poured_from_bottle
     assert_equal "Homebrew/homebrew", tab.tap
+    assert_equal :stable, tab.spec
```
mikemcquaid: Would be good to have a test without a `tab.spec` too just to verify that case works properly as well (for old tabs)

Library/Homebrew/formula.rb
xu-cheng: The reason that a spec is not available is definitely because lack of URL. [Code](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/formula.rb#L118-L121)

Library/Homebrew/formula.rb
xu-cheng: This method is used to switch spec between `:stable`, `:devel` and `:head`. What better name do you have in mind.

Library/Homebrew/formula.rb
mikemcquaid: Sure, I just mean from looking at this function alone it's a bit confusing.

Library/Homebrew/formula.rb
mikemcquaid: Maybe just `spec`?

Library/Homebrew/extend/ARGV.rb
xu-cheng: I think we can leave this as it is. Because this is almost a private API.

Library/Homebrew/test/test_tab.rb
xu-cheng: Updated to test both old version and new version receipt file

xu-cheng:
Addressed all of comments.

Library/Homebrew/extend/ARGV.rb
mikemcquaid: :+1:

Library/Homebrew/test/test_tab.rb
```diff
@@ -66,6 +67,22 @@ def test_other_attributes
...
+    assert_equal @unused.sort, tab.unused_options.sort
+    refute_predicate tab, :built_as_bottle
+    assert_predicate tab, :poured_from_bottle
+    assert_equal "Homebrew/homebrew", tab.tap
+    assert_equal :stable, tab.spec
+    refute_nil tab.time
+    assert_equal TEST_SHA1, tab.HEAD
+    assert_equal :clang, tab.cxxstdlib.compiler
+    assert_equal :libcxx, tab.cxxstdlib.type
+  end
```
mikemcquaid: :heart_eyes:

mikemcquaid:
ship it squirrel

xu-cheng:
:ship:ed :smiley:

PR #42211, opened 2015-07-28
juju 1.24.3

sinzui:


dunn:
Would it be possible to make the test something a little more substantial? Could it run `init` or `bootstrap` to test up a configuration/environment?

xu-cheng:

sinzui:
@dunn. We cannot test a bootstrap because that requires cloud credentials. Juju does support an OS X local provider. I am pondering a test that will set JUJU_HOME to a temp dir, run juju init, then run juju metadata validate-tools. On an open network this would confirm that Juju can find the cloud agents needed to bootstrap. This test would fail when the agents are only in proposed streams. The test would also fail on closed networks.

PR #42212, opened 2015-07-28
blacklist: remove libiconv

dunn:
Build scripts still sometimes fail to detect it, but this message never shows up since `brew info` finds the actual formula first.

### DIFF
--- a/Library/Homebrew/blacklist.rb
+++ b/Library/Homebrew/blacklist.rb
@@ -1,15 +1,7 @@
 def blacklisted? name
   case name.downcase
-  when 'screen', /^rubygems?$/ then <<-EOS.undent
-    Apple distributes #{name} with OS X, you can find it in /usr/bin.
-    EOS
-  when 'libpcap' then <<-EOS.undent
-    Apple distributes #{name} with OS X, you can find it in /usr/lib.
-    EOS
-  when 'libiconv' then <<-EOS.undent
-    Apple distributes #{name} with OS X, you can find it in /usr/lib.
-    Some build scripts fail to detect it correctly, please check existing
-    formulae for solutions.
+  when 'gem', /^rubygems?$/ then <<-EOS.undent
+    Homebrew provides gem via: `brew install ruby`.
     EOS
   when 'tex', 'tex-live', 'texlive', 'latex' then <<-EOS.undent
     Installing TeX from source is weird and gross, requires a lot of patches,


DomT4:
Isn't this also the case for `screen` and `libpcap`?

dunn:
Woops, very true. Updated.

Library/Homebrew/blacklist.rb
```diff
@@ -1,16 +1,5 @@
 def blacklisted? name
   case name.downcase
-  when 'screen', /^rubygems?$/ then <<-EOS.undent
```
xu-cheng: We don't need to remove rubygems right?

dunn: Yeah, that's a mistake.

xu-cheng: Also I'm a little hesitant on removing formulae which are available by taps. Blacklist is used to remind users that certain tools are already available in OS X, while dupes tap is used to offer alternate newer version of these tools. Since blacklist message will be auto ignored when formulae exist, I see no harm to keep them here.

dunn: Now that screen is removed, should we change that case to
```ruby
  when 'gem', /^rubygems?$/ then <<-EOS.undent
    Homebrew provides gem via: `brew install ruby`.
    EOS
```

dunn: Oh, I missed your last comment, @xu-cheng.

I would agree, except having these formulae blacklisted prevents Homebrew from informing the user that it does exist in a tap.  Currently:
```
🐙  brew install libpcap
Error: No available formula for libpcap
Apple distributes libpcap with OS X, you can find it in /usr/lib.
```
With this change:
```
🐙  brew install libpcap
Error: No available formula for libpcap
==> Searching formulae...
==> Searching taps...
homebrew/dupes/libpcap
```

xu-cheng: In that case, we can just edit the blacklist message. Append the information that you can install from tap if you need a newer version of it.

dunn: But that's true for every formula in homebrew/dupes. Why are these three in particular on the blacklist?

xu-cheng: Yeah, you are right. Maybe we should handle it the way other around. That is to inform users the tools can be found in OS X when searching taps return the result of homebrew/dupes. 

Overall, I think we should offer a better user experience and not all of users are aware what `dupes` means in the context of `homebrew/dupes`

dunn: I definitely agree. Would `search_tap` in search.rb be the correct place for the new message?

xu-cheng: No, I think we should put it where `search_tap`'s result gets print. The reason is this method is calling somewhere else e.g. audit.rb.

dunn:
My suggestion would be to merge this, then use a different PR to work on the way search results are displayed (see also #42222).

xu-cheng:
:+1:

DomT4:
No objection here.

PR #42213, opened 2015-07-28
Update example-formula.rb

mikicaivosevic:
Refactor if else


bfontaine:
I personally find the original code more understandable.

dunn:
I agree; the ternary operator isn't as readable, and in the example formula we should emphasize readability for new contributors.

DomT4:
Agree with Baptiste & Alex here.

Although `args << build.with? "example2" ? "--with-example2" : "--with-example1"` is cleaner it isn't particularly nice for Ruby/Homebrew newcomers.

dunn:
I think we're going to pass on this, but thank you for the pull request!

PR #42215, opened 2015-07-28
Update osm2pgsql to latest stable version: 0.88

bwhtmn:
Built and tested conversion of 2 GB extract. New version works fine and seems to be faster.

### DIFF
--- a/Library/Formula/osm2pgsql.rb
+++ b/Library/Formula/osm2pgsql.rb
@@ -3,7 +3,6 @@ class Osm2pgsql < Formula
   homepage "https://wiki.openstreetmap.org/wiki/Osm2pgsql"
   url "https://github.com/openstreetmap/osm2pgsql/archive/0.88.0.tar.gz"
   sha256 "45a6768e680a50c416fdae72cf4b26091644947a7ead44c8b7484f2276b2f119"
-  revision 1
 
   bottle do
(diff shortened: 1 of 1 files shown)


dunn:
The build is failing on Mountain Lion:
```
geometry-builder.cpp:188:25: error: use of undeclared identifier 'isnan'; did you mean 'std::isnan'?
                assert(!isnan(delta));
                        ^~~~~
                        std::isnan
```
Try adding `needs :cxx11` above `def install` and `ENV.cxx11` below it.

Library/Formula/osm2pgsql.rb
```diff
@@ -1,8 +1,8 @@
 class Osm2pgsql < Formula
   desc "Convert OpenStreetMap data to postGIS-enabled PostgreSQL db"
   homepage "https://wiki.openstreetmap.org/wiki/Osm2pgsql"
-  url "https://github.com/openstreetmap/osm2pgsql/archive/0.87.0.tar.gz"
-  sha1 "6f302500e52d6e42147cc8dff8f344677e3131f8"
+  url "https://github.com/openstreetmap/osm2pgsql/archive/0.88.0.tar.gz"
+  sha256 "45a6768e680a50c416fdae72cf4b26091644947a7ead44c8b7484f2276b2f119"
   revision 1
```
DomT4: `revision`s get scrapped with version updates.

PR #42220, opened 2015-07-28
geoip 1.6.6

oschwald:


dunn:
Regarding the test, could we turn http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz into a `resource`? It will to be a file that doesn't change over time since we specify a `sha256` in the resource block.

oschwald:
That file is updated monthly.

DomT4:
Yeah, I think we're stuck with that for now, although I'd like to look at alternatives at some point.

Merged in https://github.com/Homebrew/homebrew/commit/f48b1270039efd5cc4d05f69fd8746650f83dac4. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #42227, opened 2015-07-28
ffmpeg build with "--enable-opencl" by default

quicksurf:


mistydemeo:
Are there any specific requirements for building with opencl? Is there a reason to not build with this on by default?

quicksurf:
Other than being on OS X, not that I know of. I added it locally on a test machine and built it and it built successfully with no issues. I've not tried to build on other platforms to see if having it on would cause issues.  I'd imagine that non OS X machines would choke since open-cl is a Mac specific thing.

quicksurf:
Here's the output of ffmpeg after building it

ffmpeg version 2.7.2 Copyright (c) 2000-2015 the FFmpeg developers
  built with Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)
  configuration: --prefix=/usr/local/Cellar/ffmpeg/2.7.2 --enable-shared --enable-pthreads --enable-gpl --enable-version3 --enable-hardcoded-tables --enable-avresample --cc=clang --host-cflags= --host-ldflags= --enable-libx264 --enable-libmp3lame --enable-libvo-aacenc --enable-libxvid --enable-libfaac --enable-opencl --enable-nonfree --enable-vda
  libavutil      54. 27.100 / 54. 27.100
  libavcodec     56. 41.100 / 56. 41.100
  libavformat    56. 36.100 / 56. 36.100
  libavdevice    56.  4.100 / 56.  4.100
  libavfilter     5. 16.101 /  5. 16.101
  libavresample   2.  1.  0 /  2.  1.  0
  libswscale      3.  1.101 /  3.  1.101
  libswresample   1.  2.100 /  1.  2.100
  libpostproc    53.  3.100 / 53.  3.100
Hyper fast Audio and Video encoder
usage: ffmpeg [options] [[infile options] -i infile]... {[outfile options] outfile}...

Use -h to get full help or, even better, run 'man ffmpeg'

quicksurf:
The more I think about it, the more I feel it might not be a good idea to build it by default. As far as I can tell, it's a fairly recent addition, and I don't have a good way to test how it would behave on older versions of OS X.

In all honesty, I totally stumbled across the option.  I was going through FFMPEG's documentation while researching how I'm going to set up a processing pipe-line for an upcoming project where I'll have to ingest and make dailies and DI's of almost 100TB of 4K footage and was looking to see if they'd added any hardware support for any video hardware when I found it.  Not sure if it'll actually be useful for my needs yet (still gotta test it), but thought it'd be useful to at least have the option available via homebrew.

quicksurf:
Just a quick update: running 'ffmpeg -opencl_bench' results in: 

clEnqueueNDRangeKernel error 'INVALID WORK GROUP SIZE'
[OPENCLUTILS @ 0x108c8d9b8] Benchmark failed with OpenCL device Intel(R) Core(TM) i7-3720QM CPU @ 2.60GHz
platform_idx	device_idx	device_name	runtime
0	1	GeForce GT 650M	19206
0	0	HD Graphics 4000	101264

Granted this is on an older macbook pro (mid-2012), but still, probably should keep this as an option and not build it in by default.

Additionally, doing some benchmarking on a 1080p h.264 .mov file (yuvj4:2:0 color space) to rawvideo piped to /dev/null without any OpenCL or -hwaccel options results in a decode speed of ~400fps, whereas with the flags '-opencl_options platform_idx=0:device_idx=1 -hwaccel:v auto' set the same file decodes to /dev/null at ~120fps. The same if I set the device_idx to the other device (0). Kind of disappointing performance. Clearly the CPU decoding is faster if that's all you're going to do.  With the hardware acceleration, FFMPEG pegged 1 core at 100%, with CPU decoding all 8 cores got slammed to 100%, so if someone where to have a filtergraph in place as part of transcode, this could potentially free up CPU resources to dedicate to filters. More testing required.

Additionally, doing this changes the pixel format from yuvj420 to uyvy422 on the input stream, so I'd say extreme caution should be used before using this as it could really hose up any video.

tdsmith:
If adding --enable-opencl doesn't change anything unless you specifically ask for OpenCL acceleration at runtime, I kinda think it would be fine to switch it on anyway.

quicksurf:
It's up to you guys/gals.  I've run about 15 hours of footage through it with it built in and it doesn't seem to have any effect if you don't explicitly ask for it.  My only real concern would be how it effects older versions of OS X since I tend to be on the latest and greatest.

mikemcquaid:
Let's just enable it by default :+1:

quicksurf:
That should do it. Let me know if I need to do anything else.

mikemcquaid:

For future reference the preferred commit message format is `ffmpeg: build with opencl support.`.

PR #42229, opened 2015-07-28
AGREP - approximate GREP version 3.41.4

timemachine3030:


Library/Formula/agrep.rb
```diff
@@ -0,0 +1,29 @@
+require 'formula'
+
+class Agrep < Formula
+  desc "Approximate GREP for fast fuzzy string searching."
+  revision 2
```
dunn: A new formula doesn't need a revision specified—it defaults to 0.

dunn:
Thanks for the submission! I have to ask, though—why your fork?  We try to only accept formula from canonical sources.

timemachine3030:
@dunn thanks, updated.

timemachine3030:
I needed to make 2 rounds of changes to the upstream version, so I forked it:

 1. General typo clean up that prevented building across many platforms. See https://github.com/Wikinaut/agrep/pull/8
 2. The source needed additional updates to allow it to build under OSX. If/When the above is merged I will create another pull request for the OSX build changes.

As the Wikinaut/agrep hasn't been updated in 10+ months I pointed the formula to my fork; if however the 2 commits make it into the upstream version I will update the formula and submit a second revision.

mistydemeo:
Like @dunn mentioned, we don't typically accept unofficial forks like this. I think it makes more sense to submit them as pull requests and give the author a reasonable amount of time to respond, after which we might accept a formula that builds from the upstream tarball with patches applied.

PR #42232, opened 2015-07-29
m2c 0.7 (new formula)

FranklinChen:
Modula-2 to C compiler

### DIFF
--- a/Library/Formula/m2c.rb
+++ b/Library/Formula/m2c.rb
@@ -3,7 +3,10 @@ class M2c < Formula
   homepage "https://savannah.nongnu.org/projects/m2c/"
   url "http://download.savannah.gnu.org/releases/m2c/0.7/m2c-0.7.tar.gz"
   sha256 "b725ed617f376e1a321e059bf1985098e950965d5edab161c6b24526f10a59bc"
+  head "http://git.savannah.gnu.org/cgit/m2c.git"
 
+  # Hacks purely for this 0.7 release. Git head already fixes installation glitches.
+  # Will remove hacks on release of next version.
   def install
     # The config for "gcc" works for clang also.
     cp "config/generic-gcc.h", "config/generic-clang.h"
@@ -12,33 +15,29 @@ class M2c < Formula
     # Makefile is buggy!
     inreplace "Makefile", "install: all uninstall", "install: all"
     inreplace "Makefile", "mkdir", "mkdir -p"
-    mkdir_p prefix/"include"
+    include.mkpath
 
     system "make", "install", "prefix=#{prefix}", "man1dir=#{man1}"
   end
 
   test do
-    # -make mode for now needs running in source directory.
-    cd testpath do
-      hello_mod = "Hello.mod"
-      hello_exe = "Hello"
+    hello_mod = "Hello.mod"
+    hello_exe = "Hello"
 
-      (testpath/hello_mod).write <<-EOF.undent
-        MODULE Hello;
+    (testpath/hello_mod).write <<-EOF.undent
+      MODULE Hello;
 
-        FROM InOut IMPORT
-          WriteLn, WriteString;
+      FROM InOut IMPORT
+        WriteLn, WriteString;
 
-        BEGIN
-          WriteString ("Hello world!");
-          WriteLn;
-        END Hello.
-      EOF
+      BEGIN
+        WriteString ("Hello world!");
+        WriteLn;
+      END Hello.
+    EOF
 
-      system "#{bin}/m2c", "-make", hello_mod, "-o", testpath/hello_exe
+    system "#{bin}/m2c", "-make", hello_mod, "-o", hello_exe
 
-      output = shell_output(testpath/hello_exe)
-      assert_equal "Hello world!\n", output
-    end
+    assert_equal "Hello world!\n", shell_output(hello_exe)
   end
 end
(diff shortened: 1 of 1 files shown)


Library/Formula/m2c.rb
```diff
@@ -0,0 +1,17 @@
+class M2c < Formula
+  desc "Modula-2 to C compiler"
+  homepage "https://savannah.nongnu.org/projects/m2c/"
+  url "http://download.savannah.gnu.org/releases/m2c/0.7/m2c-0.7.tar.gz"
+  sha256 "b725ed617f376e1a321e059bf1985098e950965d5edab161c6b24526f10a59bc"
+
+  def install
+    system "./configure", "+cc=gcc"
+
+    # Makefile is buggy!
```
dunn: It would be great if you could you submit a patch for these changes upstream.

dunn:
Two things flagged by the test bot:
```
==> audit problems
m2c:
 * A top-level "man" directory was found
Homebrew requires that man pages live under share.
This can often be fixed by passing "--mandir=#{man}" to configure.
```

Library/Formula/m2c.rb
FranklinChen: The Git repo is already very fast-moving and has fixed some things but there is no 0.8 out yet.

Library/Formula/m2c.rb
```diff
@@ -0,0 +1,17 @@
+class M2c < Formula
+  desc "Modula-2 to C compiler"
+  homepage "https://savannah.nongnu.org/projects/m2c/"
+  url "http://download.savannah.gnu.org/releases/m2c/0.7/m2c-0.7.tar.gz"
+  sha256 "b725ed617f376e1a321e059bf1985098e950965d5edab161c6b24526f10a59bc"
+
+  def install
+    system "./configure", "+cc=gcc"
```
dunn: Use `ENV.cc` here instead of hardcoding GCC.  If it doesn't build with Clang, you can use `fails_with :clang` (see [`ape.rb`](https://github.com/Homebrew/homebrew/blob/1777fc04b5afbc1ed309361c1324cfe62ce32f62/Library/Formula/ape.rb#L16-L19) for an example)—but that should also be filed upstream.

Library/Formula/m2c.rb
```diff
@@ -0,0 +1,44 @@
+class M2c < Formula
+  desc "Modula-2 to C compiler"
+  homepage "https://savannah.nongnu.org/projects/m2c/"
+  url "http://download.savannah.gnu.org/releases/m2c/0.7/m2c-0.7.tar.gz"
+  sha256 "b725ed617f376e1a321e059bf1985098e950965d5edab161c6b24526f10a59bc"
+
+  def install
+    # The config for "gcc" works for clang also.
+    cp "config/generic-gcc.h", "config/generic-clang.h"
```
DomT4: Can we report this upstream somewhere?

Library/Formula/m2c.rb
```diff
@@ -0,0 +1,44 @@
...
+    system "./configure", "+cc=#{ENV.cc}"
+
+    # Makefile is buggy!
```
DomT4: Likewise, need to report these nits upstream.

Library/Formula/m2c.rb
```diff
@@ -0,0 +1,44 @@
...
+    inreplace "Makefile", "install: all uninstall", "install: all"
+    inreplace "Makefile", "mkdir", "mkdir -p"
+    mkdir_p prefix/"include"
```
DomT4: Can just do `include.mkpath`

Library/Formula/m2c.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    system "make", "install", "prefix=#{prefix}", "man1dir=#{man1}"
+  end
+
+  test do
+    # -make mode for now needs running in source directory.
+    cd testpath do
```
DomT4: Don't need to `cd` into it. You'll already be in a temporary testpath when the `test` block is processed.

Library/Formula/m2c.rb
```diff
@@ -0,0 +1,44 @@
...
+
+        BEGIN
+          WriteString ("Hello world!");
+          WriteLn;
+        END Hello.
+      EOF
+
+      system "#{bin}/m2c", "-make", hello_mod, "-o", testpath/hello_exe
+
+      output = shell_output(testpath/hello_exe)
```
DomT4: Just squash this into the line below, replacing `output` with `shell_output(testpath/hello_exe)`

Library/Formula/m2c.rb
FranklinChen: I mentioned, all of these things are already fixed in master branch. There has been radical reorganization since the 0.7 release.

FranklinChen:
I've made what I hope are the final cleanups for this formula.

tdsmith:
The test fails:

```
        Testing m2c
==> /usr/local/Cellar/m2c/0.7/bin/m2c -make Hello.mod -o Hello
Hello.mod - 134 lines, 8448 bytes
/tmp/m2c87737.c:13:1: warning: type specifier missing, defaults to 'int' [-Wimplicit-int]
main(argc,argv,envp)int argc;char **argv,**envp;{
^~~~
/tmp/m2c87737.c:16:1: warning: implicit declaration of function '_InOut' is invalid in C99 [-Wimplicit-function-declaration]
_InOut();
^
2 warnings generated.
==> Hello
Error: m2c: failed
No such file or directory - Hello
```

tdsmith:
Fixed by setting hello_exe to `testpath/"Hello"`. Pulled in d11bd73. Thanks for the pull request!

PR #42233, opened 2015-07-29
mariadb 10.1.6

makigumo:
* move depends after options
* add patch for mroonga, fixing compilation (https://mariadb.atlassian.net/projects/MDEV/issues/MDEV-8551)

### DIFF
--- a/Library/Formula/mariadb.rb
+++ b/Library/Formula/mariadb.rb
@@ -13,7 +13,8 @@ class Mariadb < Formula
   devel do
     url "http://ftp.osuosl.org/pub/mariadb/mariadb-10.1.6/source/mariadb-10.1.6.tar.gz"
     sha256 "492f28f0d7aee5bf0a0efd21c542ca4f291f349e66063695c5003df16e064959"
-
+    # fix compilation failure with clang in mroonga storage engine
+    # https://mariadb.atlassian.net/projects/MDEV/issues/MDEV-8551
     patch :DATA
   end
 
--- a/Library/Formula/mariadb.rb
+++ b/Library/Formula/mariadb.rb
@@ -11,13 +11,9 @@ class Mariadb < Formula
   end
 
   devel do
-    url "http://ftp.osuosl.org/pub/mariadb/mariadb-10.1.5/source/mariadb-10.1.5.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/mariadb.rb
```diff
@@ -80,6 +80,11 @@ def install
       args << "-DPLUGIN_TOKUDB=NO"
     end
 
+    # disable Mroonga on 10.1
```
dunn: Why does it need to be disabled?

(The conditional can also be made into a single line: `args << "-DWITHOUT_MROONGA_STORAGE_ENGINE=1" if build.devel?`)

makigumo: It fails with some messages like:
```
mariadb-10.1.6/storage/mroonga/vendor/groonga/lib/dat.cpp:191:3: error: cannot use 'try' with exceptions disabled
```
But I think I found a fix.

dunn: > But I think I found a fix.

Great! If it's something that can be reported upstream, it'd be great if you could do that too.

Library/Formula/mariadb.rb
```diff
@@ -11,13 +11,11 @@ class Mariadb < Formula
...
-    url "http://ftp.osuosl.org/pub/mariadb/mariadb-10.1.5/source/mariadb-10.1.5.tar.gz"
-    sha256 "af8788bfbb842338882e505612f86ef53a25968663a1519185ecf3de3b1efe83"
-  end
+    url "http://ftp.osuosl.org/pub/mariadb/mariadb-10.1.6/source/mariadb-10.1.6.tar.gz"
+    sha256 "492f28f0d7aee5bf0a0efd21c542ca4f291f349e66063695c5003df16e064959"
 
-  depends_on "cmake" => :build
-  depends_on "pidof" unless MacOS.version >= :mountain_lion
-  depends_on "openssl"
+    patch :DATA
```
DomT4: Can you add a note on what this is doing & link to the bug report/upstream commit so we can track this as it progresses? Thanks!

makigumo: Done. I first wanted to see if it works on all platforms.

tdsmith:
Thanks for the PR! :tada: Pulled in 0006636df.

PR #42023, opened 2015-07-22
Define opt_pkgshare.

dpo:
A small addition to the recent `pkgshare` location. See https://github.com/Homebrew/homebrew-science/issues/2450#issuecomment-123409103

PR #42237, opened 2015-07-29
Install examples to subdirectories

guoxiao:
Install these examples to subdirectories to avoid collisions when linked to `HOMEBREW_PREFIX/share/examples`.

### DIFF
--- a/Library/Formula/hiredis.rb
+++ b/Library/Formula/hiredis.rb
@@ -18,14 +18,14 @@ class Hiredis < Formula
     ENV["OBJARCH"] = "-arch #{MacOS.preferred_arch}"
 
     system "make", "install", "PREFIX=#{prefix}"
-    share.install "examples"
+    pkgshare.install "examples"
   end
 
   test do
     # running `./test` requires a database to connect to, so just make
     # sure it compiles
     system ENV.cc, "-I#{include}/hiredis", "-L#{lib}", "-lhiredis",
-           share/"examples/example.c", "-o", testpath/"test"
+           pkgshare/"examples/example.c", "-o", testpath/"test"
     File.exist? testpath/"test"
   end
 end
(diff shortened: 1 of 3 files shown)


dunn:
We usually only use revisions for formula changes that everyone ought to have. These changes don't seem important enough, in my opinion.

guoxiao:
@dunn But if we don't use revisions, users who install the formula via bottles will still have these examples linked to `HOMEBREW_PREFIX/share/examples`, which may cause collisions.

dunn:
True.  I was assuming this was just a precaution, though.  Are there collisions happening right now?

guoxiao:
No collisions right now. I'm working on a new formula and found this existing problem.

DomT4:
I've been trying to encourage people to stick examples in `libexec/"examples"` for a while now to handle this edge-case, but it never caught on.

Unless we can prove problematic clashes here we shouldn't bump revisions.

guoxiao:
@DomT4 All right. I've updated it as you suggest.

tdsmith:
Sorry to yank you around, but we recently defined Formula#pkgshare, which points to e.g. prefix/share/hiredis; I think keeping them somewhere under `share` is traditional and this avoids the conflicts you're seeing.

DomT4:
@tdsmith Did we? #42023 is still open.

tdsmith:
Yes; I merged `pkgshare` without adding a `opt_pkgshare`.

DomT4:
Oh, cool! :+1:

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/909f9e36983afe9b2a2afaf999945a1fa99448c6. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #42238, opened 2015-07-29
exiftool 9.99

ftc2:


xu-cheng:

PR #42246, opened 2015-07-29
unbound: 1.5.4

MindTooth:
Only tested on El Capitan.

### DIFF
--- a/Library/Formula/unbound.rb
+++ b/Library/Formula/unbound.rb
@@ -3,8 +3,8 @@ require "formula"
 class Unbound < Formula
   desc "Validating, recursive, caching DNS resolver"
   homepage "https://www.unbound.net"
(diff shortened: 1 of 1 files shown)


dunn:
Would it be possible to add a test, just to check basic functionality?

MindTooth:
I'll be away for a few days.  But I'll look into when I can.

DomT4:
Will merge for now, since the new bottle is `cellar :any` and that's a nice thing for users. If you could swing by with a PR when you get a moment to add a test though, that'd be awesome.

Thanks @MindTooth! Merged in https://github.com/Homebrew/homebrew/commit/1b3ce6b330723ab9a94aba2865c814f82cc6ac63.

MindTooth:
Really not sure other than a version check.

Similar issue here (respons): https://github.com/Homebrew/homebrew/pull/40144#issuecomment-107664647

PR #42248, opened 2015-07-29
thefuck 2.5.6

scorphus:
Hey, Homebrewers! This pull request upgrades TheFuck from 2.4 to 2.5.6. Besides some improvements and fixes, this new version brings new rules, including `tsuru_login`, `dirty_unzip`, `dirty_tar` and `dirty_untar`.

This also improves `caveats`, as [*suggested* by a user](https://github.com/nvbn/thefuck/issues/2#issuecomment-126013625). Formula tested on Yosemite.

Thanks in advance!

### DIFF
--- a/Library/Formula/thefuck.rb
+++ b/Library/Formula/thefuck.rb
@@ -1,8 +1,8 @@
 class Thefuck < Formula
   desc "Programatically correct mistyped console commands"
   homepage "https://github.com/nvbn/thefuck"
-  url "https://pypi.python.org/packages/source/t/thefuck/thefuck-2.4.tar.gz"
-  sha256 "e942e6613e1bf2619d01476c5404fae67abce7adfae88c713689f41f0f51213e"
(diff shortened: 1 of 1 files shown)


Library/Formula/thefuck.rb
```diff
@@ -58,9 +58,10 @@ def install
 
   def caveats; <<-EOS.undent
     Add the following to your .bash_profile, .bashrc or .zshrc:
-      bash: eval "$(thefuck-alias)"
 
-      Other shells: https://github.com/nvbn/thefuck/wiki/Shell-aliases
+      eval "$(thefuck-alias)"
```
dunn: It's now implied that this will work on zsh as well as bash; is that true?

scorphus: Very much so! :smiley: TheFuck can guess which shell it's running in :wink:

dunn: :clap:

dunn:
Merged in 08bb15a8a5e4aa76c61a7f9f2a68b598b3ee6a00, thanks again!

PR #42258, opened 2015-07-30
Flow v0.14.0

gabelevi:


Library/Formula/flow.rb
```diff
@@ -3,8 +3,8 @@
 class Flow < Formula
   desc "Static type checker for JavaScript"
   homepage "http://flowtype.org/"
-  url "https://github.com/facebook/flow/archive/v0.13.1.tar.gz"
-  sha1 "9595586fb30c6baab6893b0d3fcbbb8719ab41ae"
+  url "https://github.com/facebook/flow/archive/v0.14.0.tar.gz"
+  sha1 "2450941786c51d943ae2ca53b0a6fe6feb457bb1"
```
DomT4: You'll need to use `sha256` here. We've deprecated SHA1.

PR #42260, opened 2015-07-30
libfreenect2 initial commit

ybott:
Support for Kinect version 2 depth camera.

### DIFF
--- a/Library/Formula/libfreenect2.rb
+++ b/Library/Formula/libfreenect2.rb
@@ -1,0 +1,128 @@
+class Libfreenect2 < Formula
+  desc "Drivers/Example for Kinect V2"
+  homepage "https://github.com/OpenKinect/libfreenect2"
+  url "https://github.com/ybott/libfreenect2/archive/v0.0.1-alpha.tar.gz"
+  sha256 "704572b2848c38499ef934d9445d275033eb94d245aa60706eea2179ab2629c2"
+
+  depends_on "libusb"
+  depends_on "nasm" => :optional
+  depends_on "jpeg-turbo" => :optional
+  depends_on "homebrew/science/opencv" => :optional
+  depends_on "homebrew/versions/glfw3" => :optional
+  depends_on "pkg-config" => :build
+  depends_on "cmake" => :build
+
+  patch :DATA
+
+  def install
+    system "cmake", "examples/protonect", *std_cmake_args
+    system "make"
+    system "make", "install"
+  end
+end
+
+__END__
+diff --git a/examples/protonect/CMakeLists.txt b/examples/protonect/CMakeLists.txt
+index ba93a67..a38fd9f 100644
+--- a/examples/protonect/CMakeLists.txt
++++ b/examples/protonect/CMakeLists.txt
+@@ -202,4 +202,4 @@ IF(LIBFREENECT2_THREADING_TINYTHREAD)
+ ENDIF(LIBFREENECT2_THREADING_TINYTHREAD)
+ INSTALL(FILES "${PROJECT_BINARY_DIR}/freenect2Config.cmake" DESTINATION lib/cmake/freenect2/)
+ INSTALL(FILES "${PROJECT_BINARY_DIR}/freenect2.pc" DESTINATION lib/pkgconfig/)
+-  
++INSTALL(FILES "${MY_DIR}/bin/Protonect" DESTINATION bin/)
+diff --git a/examples/protonect/cmake_modules/FindLibUSB.cmake b/examples/protonect/cmake_modules/FindLibUSB.cmake
+index f3e822b..af7a16a 100644
+--- a/examples/protonect/cmake_modules/FindLibUSB.cmake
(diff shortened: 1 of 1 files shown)


Library/Formula/libfreenect2.rb
```diff
@@ -0,0 +1,143 @@
+require "formula"
```
DomT4: Hasn't been necessary for a while. Can kill it.

Library/Formula/libfreenect2.rb
```diff
@@ -0,0 +1,143 @@
...
+
+class Libfreenect2 < Formula
+  desc "Drivers/Example for Kinect V2"
+  homepage "https://github.com/OpenKinect/libfreenect2"
+  url "https://github.com/OpenKinect/libfreenect2/archive/master.zip"
```
DomT4: We'll need an actual release. The `master.zip` will update & break the checksum every time someone pushes a new commit there. Suggest you ask upstream if they're willing to tag a release for you here.

Library/Formula/libfreenect2.rb
```diff
@@ -0,0 +1,143 @@
...
+  version "0.0.1"
+  
+  head "https://github.com/OpenKinect/libfreenect2.git"
+
+  option :universal
+
+  depends_on "libusb"
+  depends_on "nasm" => :recommended
+  depends_on "jpeg-turbo" => :recommended
+  depends_on "homebrew/science/opencv" => :recommended
```
DomT4: Can't have `:recommended` or mandatory cross-tap dependencies in the core. These would have to be `:optional` at the very most.

Library/Formula/libfreenect2.rb
```diff
@@ -0,0 +1,143 @@
...
+  depends_on "homebrew/versions/glfw3" => :recommended
```
DomT4: Can't have `:recommended` or mandatory cross-tap dependencies in the core. These would have to be `:optional` at the very most.

Library/Formula/libfreenect2.rb
```diff
@@ -0,0 +1,143 @@
...
+  depends_on "pkg-config" => :build
+  depends_on "cmake" => :build
+
+  patch :DATA
+
+  def install
+    args = "-DCMAKE_INSTALL_PREFIX=/usr/local/Cellar/libfreenect2/#{version}"
```
DomT4: We don't hardcore `/usr/local`. You can put Homebrew anywhere, `/usr/local` is just our preference and where our binaries are targeted. Use `system "cmake", "examples/protonect", *std_cmake_args` instead.

Library/Formula/libfreenect2.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  depends_on "libusb"
+  depends_on "nasm" => :recommended
+  depends_on "jpeg-turbo" => :recommended
+  depends_on "homebrew/science/opencv" => :recommended
+  depends_on "homebrew/versions/glfw3" => :recommended
+  depends_on "pkg-config" => :build
+  depends_on "cmake" => :build
+
+  patch :DATA
```
DomT4: Please add an upstream link to the bug submission/commit containing this patch, and a short summary of why it is necessary.

Library/Formula/libfreenect2.rb
```diff
@@ -0,0 +1,128 @@
...
+
+  depends_on "libusb"
+  depends_on "nasm" => :optional
+  depends_on "jpeg-turbo" => :optional
+  depends_on "homebrew/science/opencv" => :optional
+  depends_on "homebrew/versions/glfw3" => :optional
+  depends_on "pkg-config" => :build
+  depends_on "cmake" => :build
+
+  patch :DATA
```
bfontaine: What is this patch for? Did you submit it upstream? Could you add a comment above this line explaining why it’s necessary? Thanks!

Library/Formula/libfreenect2.rb
```diff
@@ -0,0 +1,128 @@
...
+
+  def install
+    system "cmake", "examples/protonect", *std_cmake_args
+    system "make"
+    system "make", "install"
+  end
+end
```
bfontaine: Please add a `test do` block.

Library/Formula/libfreenect2.rb
```diff
@@ -0,0 +1,128 @@
+class Libfreenect2 < Formula
+  desc "Drivers/Example for Kinect V2"
+  homepage "https://github.com/OpenKinect/libfreenect2"
+  url "https://github.com/ybott/libfreenect2/archive/v0.0.1-alpha.tar.gz"
```
DomT4: This release *needs* to come from the upstream repository, not your own fork. Only an official upstream release is acceptable I'm afraid.

dunn:
I'm assuming you're using your fork because upstream hasn't tagged a stable release?  In these cases we ask that formula be added to our head-only repository.  There's actually already a PR open there for this: Homebrew/homebrew-head-only#153.

If you think the developers will make a stable release soon, we can leave this open. Otherwise I think we'll close it in favor of the head-only submission.

PR #42263, opened 2015-07-30
Ensure transmission runs at load

idolize:
The transmission process daemon was not running correctly when using `brew services start transmission` (or when simply restarting the computer) because the `RunAtLoad` key was not explicitly set to `true`. This fixes it and ensures it always runs correctly.

cc @adzenith


DomT4:
You'll need to change the `sha1` usage to `sha256` to get this to pass. We've formally deprecated SHA1.

idolize:
Thanks @DomT4; updated.

dunn:
Merged in 45e2840df97b2657a98c5473bb742af3465224aa, thanks for catching that!

PR #42264, opened 2015-07-30
Tap: allow enumerate non-git taps

xu-cheng:
Also add a `Tap#git?` method to indicate whether the tap is a git
repository.

### DIFF
--- a/Library/Homebrew/tap.rb
+++ b/Library/Homebrew/tap.rb
@@ -37,7 +37,7 @@ class Tap
   # e.g. `https://github.com/user/homebrew-repo`
   def remote
     @remote ||= if installed?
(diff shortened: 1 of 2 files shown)


Library/Homebrew/tap.rb
```diff
@@ -49,6 +49,11 @@ def remote
     end
   end
 
+  # True if this {Tap} is a git repository.
+  def git_remote?
```
bfontaine: I find this naming confusing because the method don’t really tests if this is a git repo with a remote, just that it’s a git repo. Why not `git?`?

xu-cheng: Good call, updated.

PR #42269, opened 2015-07-30
phraseapp 0.2.0.beta9 (new formula)

sacry-dyn:


DomT4:
```
==> audit problems
phraseapp:
````

Afraid this doesn't meet our requirements at the moment. You can run this same check locally with `brew audit --strict --online /path/to/formula`.

We'll be happy to take a look at a new PR when the requirements are met, but please note software tagged with `alpha` or `beta` should live in [devel-only](https://github.com/Homebrew/homebrew-devel-only) until stable releases are available, unless said software is overwhelmingly popular.

A future submission should build from source using Go as well. We don't use upstream binaries, we produce our own.

Thanks for the submission!

sacry-dyn:
Hi! Thanks a lot for the reply! Sad to hear. We will be working on it! :) Regards Matthias

DomT4:
Great! We're more than happy to look at this again in future should you wish to resubmit at a later date.

PR #42271, opened 2015-07-30
Fix heroku-toolbelt.rb URL

allaire:
The caveat has the wrong URL, using a different project with the same name.

### DIFF
--- a/Library/Formula/heroku-toolbelt.rb
+++ b/Library/Formula/heroku-toolbelt.rb
@@ -20,7 +20,7 @@ class HerokuToolbelt < Formula
     <<-EOS.undent
       Unlike the standalone download for Heroku Toolbelt, the Homebrew package
       does not come with Foreman. It is available via RubyGems, direct download,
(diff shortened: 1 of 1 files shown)


Library/Formula/heroku-toolbelt.rb
```diff
@@ -20,7 +20,7 @@ def caveats
     <<-EOS.undent
       Unlike the standalone download for Heroku Toolbelt, the Homebrew package
       does not come with Foreman. It is available via RubyGems, direct download,
-      and other installation methods. See theforeman.org for more info.
+      and other installation methods. See http://ddollar.github.io/foreman/ for more info.
```
DomT4: Should be `https://ddollar.github.io/foreman/`

allaire:
Fixed and rebased.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/3d1afd64eed92a3837ebb4e726f46773a897e234. Thanks for spotting this, and contributing to Homebrew! We appreciate it.

PR #42275, opened 2015-07-30
acme: fix devel url also use https and sha256

vszakats:


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/18c52510c3fe972ec137da792c22f4f350caeb66. Thanks @vszakats.

PR #42281, opened 2015-07-31
Implement `brew desc`, with a cache

hivehand:
This is an attempt to recreate @telemachus’s `brew desc` tool, using the `desc` component now built into formulae. Like the original `brew desc`, it lets you look up formulae either by their exact names, or by searching through some combination of name and description.

**Notes:**

- It follows the [suggestion] made by @mistydemeo, in the discussion over the pull request that added `desc` to formulae, and uses a cache for performance.
- Said cache tries to be “pay-as-you-go”: the cache isn't created, or updated, until you either explicitly do so with `brew desc --cache`, or perform a search. This allows people who don’t expect to use `brew desc` to avoid paying ~5 extra seconds with every `brew update`.
- A slight performance tweak to `utils.rb`’s maximum-length-determination logic, with input from @bfontaine, came along for the ride.

[suggestion]: https://github.com/Homebrew/homebrew/pull/39911#issuecomment-103697853

**Caveats:**

- I have almost no personal experience with Kegs or Casks, and so can’t guarantee that `desc` and the cache work properly with them.
- I’m not well-versed with brew’s model for handling `ARGV`, so it’s possible that I’m making a lot of unnecessary noise in the `desc.rb` command handling.

**Open Questions:**

- `brew search --desc` is exactly equivalent to `brew desc -d`. I’m not sure it makes sense to preserve both, particularly since the former was never documented, but for the time being I’ve rewritten it to take advantage of the cache.
- The cache-generation code is completely silent. I don’t know if we want to notify the user to the effect that “It’s OK! Something’s actually happening!” when the first attempt to run `brew desc` takes several seconds.

Feedback is more than welcome.


dunn:
The tests are failing on Mountain Lion—are you using any Ruby that's not valid in 1.8?

hivehand:
That’s entirely possible. I apologize. I’ll install 1.8.7 and see about finding and fixing the issue. (Off the top of my head, I'm guessing that it might be due to my use of `proc` rather than `Proc.new`, but we’ll see.)

hivehand:
Aieeee, no. Close, but the actual culprit was my using `{ <sym_name>: <value> }` instead of `{ :<sym_name> => <value> }`. I’ll fix that, and update after proper testing.

Library/Homebrew/cmd/desc.rb
```diff
@@ -0,0 +1,59 @@
...
+    brew desc (-s|-n|-d) <string>   Search specifed fields for string
+    brew desc (-s|-n|-d) /<regex>/  Search specifed fields for regex
+EOM
+
+
+module Homebrew
+  def desc
+    options = ARGV.options_only
+
+    if options.include?("--decache")
```
bfontaine: You can drop the parentheses here.

Library/Homebrew/cmd/desc.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    if options.include?("--recache")
+      Descriptions.generate_cache
+      options.delete("--recache")
+      cache_op = true
+    end
+
+    if options.empty?
+      if ARGV.named.empty?
+        print USAGE unless cache_op
```
bfontaine: Shouldn’t this line fail, as `cache_op` is not defined above?

Library/Homebrew/cmd/desc.rb
```diff
@@ -0,0 +1,59 @@
...
+
+USAGE =<<EOM
+Print formula descriptions.
+
+Usage:
+
+    brew desc --decache             Delete the cache, if present
+    brew desc --recache             [Re]generate the cache
+    brew desc name [name ...]       Describe the named formula(e)
+    brew desc (-s|-n|-d) <string>   Search specifed fields for string
```
bfontaine: Typo here and on the line below: “specifed” → “specified”

Library/Homebrew/manpages/brew.1.md
```diff
@@ -126,6 +126,26 @@ Note that these flags should only appear after a command.
     type dependencies, pass `--skip-build`. Similarly, pass `--skip-optional`
     to skip `:optional` dependencies.
 
+  * `desc` <formula>:
+    Display <formula>'s name and one-line description.
+
+  * `desc [-s|-n|-d] <pattern>`:
+    Search both name and description (`-s`), just the names (`-n`), or just  the
+    descriptions (`-d`) for `<pattern>`. `<pattern>` is by default  interpreted
```
bfontaine: There are two spaces between “default” and “interpreted” here.

mikemcquaid:
@hivehand Can you show the output of `time brew readall` and `time brew desc --cache`? Thanks!

Library/Homebrew/manpages/brew.1.md
```diff
@@ -126,6 +126,26 @@ Note that these flags should only appear after a command.
...
+    as a literal string; if flanked by slashes, it is instead interpreted as a
+    regular expression. Formula descriptions are cached, and the cache is
+    created on the first search, making it slower than succeeding ones.
```
bfontaine: I don’t think “succeeding” is the good word here :wink:

Library/Homebrew/manpages/brew.1.md
```diff
@@ -126,6 +126,26 @@ Note that these flags should only appear after a command.
...
+
+  * `desc --recache`:
+
+    This forces the [re]construction of the name/description cache. It can be
+    combined with a search, in which case the construction will take place
+    first.
+
+  * `desc --decache`:
+
+    Delete the name/description cache, if it exists.
```
bfontaine: Should we include that in `brew cleanup`?

Library/Homebrew/manpages/brew.1.md
telemachus: How about "subsequent" or "later"?

Library/Contributions/brew_bash_completion.sh
```diff
@@ -189,6 +189,18 @@ _brew_deps ()
...
+    local cur="${COMP_WORDS[COMP_CWORD]}"
+    case "$cur" in
+    --*)
+        __brewcomp "--search"
+        return
+        ;;
+    esac
+    __brew_complete_formulae
+}
+
```
xu-cheng: Can you update zsh_completion as well.

Library/Homebrew/cmd/desc.rb
xu-cheng: `--decache` feels bit jargon.

telemachus:
@hivehand Thanks for doing this!

Library/Homebrew/cmd/desc.rb
```diff
@@ -0,0 +1,59 @@
+require 'descriptions'
+require 'cmd/search'
+
+USAGE =<<EOM
```
xu-cheng: Since none of internal commands has a usage message. I think we should drop this. Manpage should be sufficiency.

Library/Homebrew/cmd/update.rb
```diff
@@ -1,4 +1,5 @@
 require 'cmd/tap'
+require 'Descriptions'
```
xu-cheng: `require 'descriptions'`

Library/Homebrew/cmd/update.rb
```diff
@@ -57,6 +58,7 @@ def update
     else
       puts "Updated Homebrew from #{master_updater.initial_revision[0,8]} to #{master_updater.current_revision[0,8]}."
       report.dump
+      Descriptions.refresh_cache
```
xu-cheng: I'm worrying this would have huge impact on the performance. Especially considering`brew update` is already slow enough.

Library/Homebrew/manpages/brew.1.md
bfontaine: Yep.

Library/Homebrew/cmd/update.rb
telemachus: I thought the only large impact would be the initial creation of the cache. Would normal updates really be that huge?

If yes, perhaps drop this from `update` and only refresh the cache when the user calls `brew desc` explicitly? That way, users who want this will wait, but nobody else will?

mikemcquaid: Can we benchmark this before deciding either way?

mikemcquaid:
I recommend this be profiled with something like `ruby-prof --printer=call_stack --file=call_stack.html -- brew desc --cache` to try and tweak performance.

hivehand:
The first benchmarks, as requested by @mikemcquaid:

```
$ time brew readall

real	0m2.721s
user	0m2.577s
sys	0m0.139s
```

```
$ time brew desc --recache

real	0m2.925s
user	0m2.775s
sys	0m0.143s
```

More to come as soon as I familiarize myself with `ruby-prof`.

I’ll be addressing and/or replying to specific comments shortly. Thank you all for the careful proofreading and detailed feedback!

Library/Homebrew/descriptions.rb
```diff
@@ -0,0 +1,103 @@
...
+  end
+
+  # Given a regex, find all formulae whose specified fields contain a match for it.
+
+  def self.matching(regex, field = :either)
+    self.ensure_cache
+
+    matchers = {
+      :name   => proc { |name, desc| name =~ regex },
+      :desc   => proc { |name, desc| desc =~ regex },
```
bfontaine: You can use `_` for the parameters you’re no using, e.g. `{ |_, desc| desc =~ regex }`.

Library/Homebrew/descriptions.rb
```diff
@@ -0,0 +1,103 @@
...
+      :either => proc { |name, desc| (name =~ regex) || (desc =~ regex) }
+    }
+
+    results = @cache.select(&(matchers[field]))
```
bfontaine: Is it really worth creating a hash with three procs then matching on it instead of doing a `case`?

Library/Homebrew/descriptions.rb
```diff
@@ -0,0 +1,103 @@
...
+    end
+  end
+
+  # Take search results -- a hash mapping formula names to descriptions -- and
+  # print them.
+
+  def self.print(descriptions)
+    blank = "#{Tty.yellow}[no description]#{Tty.reset}"
+    descriptions.keys.sort.each do |name|
+      description = descriptions[name] || blank
```
bfontaine: You can do `descriptions.fetch(name, blank)`.

Library/Homebrew/manpages/brew.1.md
hivehand: I agree that “succeeding” is not the good word. I like “subsequent”: I’ll use that.

Library/Homebrew/cmd/desc.rb
hivehand: The line works because in Ruby, any assignment to a variable inside of a conditional defines that variable for the remainder of the containing scope — even if the condition is not met. Even if the condition *cannot* be met! If the condition isn’t met, the variable is simply assigned `nil`. For instance the following, if fed into `irb`, will output `nil` and then an error, because after the `if`, `alice` is `nil` and `bob` is undefined.

```ruby
if false
  alice = "Yikes!"
end

puts alice.inspect
puts bob.inspect
```

All of that having been said, I will explicitly set `cache_op` to `false` at the beginning of the method, because relying on that sort of implicit behavior probably amounts to taunting the Elder Gods. 😁

Library/Homebrew/cmd/desc.rb
hivehand: It was meant to be either a portmanteau of “delete” and “cache”, or analogous to “decompress”. I’m open to clearer suggestions, though. `--delete_cache`?

bfontaine: I’d prefer not to mix underscores and hyphens, so maybe `--delete-cache`.

Library/Homebrew/cmd/desc.rb
bfontaine: Thanks, I didn’t know about that.

bfontaine: Isn’t `cache_op` always `false` here? The only way for it to be `true` would be that `options` include `--decache` or `--recache`, but to enter in this block it must be empty.

hivehand: We delete the `--delete-cache` (née `--decache`) and `--recache` options after acting upon them, and set `cache_op`. That way we can tell how we came to `options.empty?` case.

Was it because the user called `brew desc` without arguments? Then we whine to some degree. (Although less that we currently do — I’ll be taking @xu-cheng’s advice regarding the usage message.)

Was it because the user called `brew desc` with only cache-administration options, which we’ve already dealt with? Then we don’t whine, and just exit quietly. I hope that makes sense.

bfontaine: It does, thank you.

Library/Homebrew/manpages/brew.1.md
hivehand: Good catch. Fixed.

Library/Homebrew/cmd/desc.rb
hivehand: Yeesh. Embarrassing. I’d fix that, except that I’m just going to take @xu-cheng’s suggestion and ditch the usage message entirely.

Library/Homebrew/manpages/brew.1.md
hivehand: I have no opinion, strong or otherwise, here, as I don’t believe I’ve ever invoked `brew cleanup`. I’ll defer to others.

Library/Contributions/brew_bash_completion.sh
hivehand: Will do.

Library/Homebrew/descriptions.rb
hivehand: I may be misunderstanding what you’re proposing. My first-pass attempt at implementing it using `case` looks like this:

```ruby
results = case field
when :name
  @cache.select { |name, _| name =~ regex }
when :desc
  @cache.select { |_, desc| desc =~ regex }
when :either
  @cache.select { |name, desc| (name =~ regex) || (desc =~ regex) }
end
```

That consumes one more line and looks, to my eye, busier to boot. Please set me straight if misinterpreting your suggestion.

Library/Homebrew/descriptions.rb
hivehand: I didn’t of that. Will do.

Library/Homebrew/cmd/update.rb
hivehand: Two clarifications:

1. The act of refreshing the cache eats on the order of 5 seconds, give or take, on a reasonably-modern system. It does this on every run, because I couldn’t devise a way of incrementally updating the cache that was both simple and reliable.

2. `Descriptions.refresh_cache` only regenerates the cache if the cache already exists. Users who never invoke `brew desc` or `brew search --desc` shouldn’t see any significant changes in their `brew update` run-time. (Those who did perform a search at some point in the past, but decide that they’d rather not pay for it with every subsequent `brew update`, can use `brew desc --delete-cache` to restore the earlier behavior.)

Library/Homebrew/cmd/desc.rb
hivehand: Fair enough. I will make it behave like `brew deps` just raise a `FormulaUnspecifiedError` instead.

hivehand:
I just pushed an update. I believe that this addresses all issues raised and suggestions made, with the exception of @bfontaine’s proposal that I use `case` instead of a hash of procs when performing a search. (I’m not yet sure that I properly understand how to use `case` concisely here.)

On the profiling front,
```
ruby-prof --printer=call_stack --file=call_stack.html -- brew desc --cache
```
fails with
```
/usr/local/lib/ruby/gems/2.2.0/gems/ruby-prof-0.15.8/lib/ruby-prof/compatibility.rb:104:in `load': cannot load such file -- brew (LoadError)
	from /usr/local/lib/ruby/gems/2.2.0/gems/ruby-prof-0.15.8/lib/ruby-prof/compatibility.rb:104:in `start_script'
	from /usr/local/lib/ruby/gems/2.2.0/gems/ruby-prof-0.15.8/bin/ruby-prof:292:in `run'
	from /usr/local/lib/ruby/gems/2.2.0/gems/ruby-prof-0.15.8/bin/ruby-prof:340:in `<top (required)>'
	from /usr/local/bin/ruby-prof:23:in `load'
	from /usr/local/bin/ruby-prof:23:in `<main>’
```
which stands to reason, since `brew` is a shell script rather than Ruby. I will see if I can teach myself enough `ruby-prof` to obtain useful results; in the meantime, hints would be welcome.

xu-cheng:
@hivehand I wrote [a small script](https://github.com/xu-cheng/dotfiles/blob/master/home/bin/brew-prof.rb) to help you profile homebrew.

If you download it and put it into place where `PATH` can find. You should be able to run `brew prof desc --cache` to generate the result.

Library/Homebrew/descriptions.rb
bfontaine: Yes I was thinking of that. Here are the benchmarks on 1M calls; the first line is the hash-matching and the second one is the `case`/`when`:

```
       user     system      total        real
   8.000000   0.160000   8.160000 (  8.192045)
   5.200000   0.080000   5.280000 (  5.310700)
```

I personally prefer the second solution because it’s simpler and more understandable for new rubyists.

mikemcquaid:
Maintainers: please hold off merging this: I haven't had the chance to review it yet but would like it before we merge it.

Library/Homebrew/descriptions.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  # Create a hash mapping all formulae to their descriptions;
+  # save it for future use.
+
+  def self.generate_cache
+    @cache = {}
+    CSV.open(CACHE_FILE, 'w') do |csv|
+      Formula.map do |f|
+        name, desc = f.name, f.desc
```
xu-cheng: `f.full_name`

Library/Homebrew/cmd/desc.rb
```diff
@@ -0,0 +1,47 @@
...
+      options.delete "--create-cache"
+      cache_op = true
+    end
+
+    if options.empty?
+      if ARGV.named.empty?
+        raise FormulaUnspecifiedError unless cache_op
+        exit
+      else
+        results = Descriptions.named(ARGV.named)
```
xu-cheng: `results = Descriptions.named(ARGV.formulae.map(&:full_name)`

Library/Homebrew/cmd/desc.rb
```diff
@@ -0,0 +1,47 @@
+require 'descriptions'
+require 'cmd/search'
+
+module Homebrew
+  def desc
+    options = ARGV.options_only
+    cache_op = false
```
xu-cheng: This `cache_op` along with below logics is confusing. I suggest you use `if-else` flaw with different methods like this:

```ruby
def desc
  if ARGV.include? "--delete-cache"
    Descriptions.delete_cache
  elsif ARGV.include? "--create-cache"
    Descriptions.generate_cache
  elsif ARGV.options_only.empty?
    results = Descriptions.named(ARGV.formulae.map(&:full_name)
    Descriptions.print(results)
  else
    ...search...
  end
end
```

Library/Homebrew/descriptions.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    results = @cache.select(&(matchers[field]))
+
+    RUBY_VERSION <= "1.8.7" ? Hash[results] : results
+  end
+
+  # Take search results -- a hash mapping formula names to descriptions -- and
+  # print them.
+
+  def self.print(descriptions)
```
xu-cheng: This method doesn't match with OOP.(This class behaviors more like a module). Can we have something more intuitive, like:
```ruby
  result = Description.search regex
  result.print
```

Library/Homebrew/descriptions.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def self.delete_cache
+    File.delete(CACHE_FILE) if File.exist?(CACHE_FILE)
+  end
+
+  # Given an array of formula names, return a hash mapping those names to
+  # their descriptions. If the array is empty, return a hash of all
+  # descriptions.
+
+  def self.named(names)
```
xu-cheng: Can we have a better for this method? A name doesn't require document to understand its function.

Library/Homebrew/descriptions.rb
```diff
@@ -0,0 +1,99 @@
...
+      names.inject({}) do |descriptions, name|
+        descriptions[name] = @cache[name]
+        descriptions
+      end
+    end
+  end
+
+  # Given a regex, find all formulae whose specified fields contain a match for it.
+
+  def self.matching(regex, field = :either)
```
xu-cheng: `search` seems to be a better name than `matching`.

Library/Homebrew/descriptions.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  # Given an array of formula names, return a hash mapping those names to
+  # their descriptions. If the array is empty, return a hash of all
+  # descriptions.
+
+  def self.named(names)
+    self.ensure_cache
+
+    if names.empty?
+      @cache
```
xu-cheng: It feels strange and unexpected that it will return whole cache for non input. If you need such function, I suggest you to create another method to do this.

mikemcquaid:
> The first benchmarks, as requested by @mikemcquaid:

@xu-cheng may have just sped these up; can you rebase on the latest `origin/master` and try again? Thanks!

Library/Homebrew/cmd/desc.rb
mikemcquaid: :+1:

Library/Homebrew/cmd/update.rb
mikemcquaid: Maybe just update/create the cache on first use and use the last update compared with the cache's date to decide whether it needs to be regenerated based on the update?

mikemcquaid:
If we include this I'd like it to be enabled by default and maybe optionally disabled.

DomT4:
I have some very minor concerns about storing the cached file in `HOMEBREW_CACHE`, if only because from time to time when people don't run `brew cleanup` forever and they hit issues it's easier to tell them *"Just wipe out `HOMEBREW_CACHE`"*.

That's not an everyday use case though, so not a big deal.

mikemcquaid:
> I have some very minor concerns about storing the cached file in HOMEBREW_CACHE, if only because from time to time when people don't run brew cleanup forever and they hit issues it's easier to tell them "Just wipe out HOMEBREW_CACHE".

Personally I think that's fine as long as the cache is just regenerated when needed.

Library/Homebrew/cmd/desc.rb
hivehand: I was trying to cater to the user who wanted to delete the cache before re-creating it with a search, and do so with a single command, but you’re right. The “pick one” logic is much cleaner.

Library/Homebrew/descriptions.rb
hivehand: I’m sorry, but I’m not entirely sure I follow. The method (which I’ll revise to return an instance, as you suggested above) returns all formulae whose names are in the argument list. In conversational terms, it translates to “give me descriptions for all the formulae named…”

I can’t think of a more appropriate named than `named`, but I’m open to suggestions.

Library/Homebrew/descriptions.rb
hivehand: True. I’ll change that.

Library/Homebrew/descriptions.rb
hivehand: Actually, this is an artifact from a time when I thought that it made sense to return *all* of the descriptions when `brew desc` was called without arguments. Since that’s no longer the case, I’ll change it.

hivehand:
When we say “Just wipe out HOMEBREW_CACHE”, do we mean emptying the directory, or deleting the directory itself? `brew desc` in its present state should have no problem with the former: it’ll note that its cache file is missing and regenerate it. The latter case, on the other hand, would break it, since it currently tries to regenerate the cache without checking that the containing directory exists.

I could revise it to (a) check whether the directory exists, and bail with an error message if it doesn’t, or (b) always try to create the directory. (I’m inclined toward (a) myself, but will defer to the experts.)

Library/Homebrew/descriptions.rb
hivehand: Just for the record, I’m reverting the change, because `Hash.fetch(key, default)` returns `default` if and only if `key` doesn’t exist in the hash. The intention here is to cover the case where `key` *does* exist, but its value is `nil`.

Library/Homebrew/descriptions.rb
hivehand: True. I’ll make `named` and `search` return instances, and make `print` an instance method.

telemachus:
@hivehand I'm pretty sure it means delete all contents. Homebrew also needs `HOMEBREW_CACHE` to work. (There's a small chance, however, that Homebrew generates the directory as needed, I suppose.)

DomT4:
> (There's a small chance, however, that Homebrew generates the directory as needed, I suppose.)

Yup. If you destroy the `CACHE` directory and then do something that requires the `CACHE` Homebrew just pops up a new one for you.

telemachus:
@DomT4 Cool. Then presumably `brew desc` should just do the same, yes?

DomT4:
Yeah, I think you want something like `HOMEBREW_CACHE.mkpath` in the appropriate places. I suspect that's what we're using elsewhere.

hivehand:
Oy. I merged rather than rebased, making for an ungodly amount of noise. I will see if I can undo the damage. I apologize.

hivehand:
Sorry for the merge-rather-than-rebase noise earlier. I think everything’s back to an (arguably) sane state. I believe that I’ve incorporated all outstanding feedback, with the exception of @xu-cheng’s suggestion that I rename `Descriptions.named`, just because I’m not yet sure what I would rename it to.

@mikemcquaid, The rebased version still takes 3-6 seconds to regenerate the cache.

```
$ time brew desc --create-cache

real	0m5.867s
user	0m4.283s
sys	0m0.818s
```
```
$ time brew desc --create-cache

real	0m3.142s
user	0m2.875s
sys	0m0.175s
```

@xu-cheng, thanks very much for the brew-prof wrapper! Would you prefer that I attempt to summarize the results, or simply stash the contents of the resulting `prof` directory somewhere publicly visible, e.g. gist?

hivehand:
<Sigh>. My attempt to undo my initial mistake of merging master into redesc, rather than rebasing redesc upon master, involved historical revision and a forced push. This has apparently thwarted Jenkins’ attempts to pull the latest version. Rather than continue flailing about, I will accept suggestions from my betters as to how I should proceed. Again, I apologize.

DomT4:
If Jenkins already has one build in the queue for something, and then someone force pushes that branch, Jenkins tends to throw a tantrum. Suspect that's what happened here. Have asked it to peek again.

DomT4:
It's happy again.

hivehand:
Thank you!

mikemcquaid:
For me 3 seconds is fast enough that we should just build/rebuild the cache on first search that would use it.

xu-cheng:
I would still like to have some benchmark on `brew update`, since the cache will be regenerated completely under current implementation. And can we make it fast by only refreshing changed formulae? Also, we might need to refresh the cache during `brew (un)tap`.

mikemcquaid:
> I would still like to have some benchmark on brew update, since the cache will be regenerated completely under current implementation.

To confirm: I don't think we should regenerate on `update` but instead regenerate on use when it's detected there's been an `update`.

hivehand:
Hi, @mikemcquaid: I apologize for my delay in replying. I’ve done my best to implement @xu-cheng’s suggestions, and add the ability to incrementally revise the cache.

Rather than generate a fresh copy of the cache on every run, `brew update` now takes advantage of the `Report` derived from git output, and only updates or deletes the formulae that have been added/revised or removed, respectively. On a typical `brew update` run, `Descriptions.update_cache` takes on the order of 0.05 seconds. I’m hoping that you consider this fast enough to let the cache-updating logic in remain in `update.rb`, since the alternative would be to have the first `brew desc` lookup/search following an update take several seconds as it regenerated the cache in full.

I’ve also taken the incremental logic and tied it into `tap.rb` and `untap.rb`, so that they quickly add and remove the target tap’s formula descriptions from the cache.

Lastly, I’ve added logic to determine the staleness of the cache by comparing its mtime to the mtime of the master Homebrew repo, and any taps that might be present. If, when asked to perform a lookup or search, `brew desc` determines that the cache exists but is stale — because of a manual `git pull` in place of a `brew update`, perhaps — it will regenerate the cache from scratch.

Feedback is of course welcome. (@xu-cheng, please let me know how you’d like me to relay the results of running `brew prof desc` and/or `brew prof update`, whether via gist or something else.)

Library/Homebrew/descriptions.rb
```diff
@@ -0,0 +1,167 @@
...
+
+  CACHE_FILE = HOMEBREW_CACHE + "desc_cache"
+
+  def self.cache
+    @cache || self.load_cache
+  end
+
+  # If the cache file exists, load it into, and return, a hash; otherwise,
+  # return nil.
+
```
mikemcquaid: Mind removing the spaces between comments and functions? Makes it easier to know what function they apply fo.

Library/Homebrew/descriptions.rb
```diff
@@ -0,0 +1,167 @@
+require "formula"
+require "csv"
+
+class Descriptions
+
```
mikemcquaid: Can remove this empty line.

Library/Homebrew/manpages/brew.1.md
```diff
@@ -126,6 +126,25 @@ Note that these flags should only appear after a command.
...
+    created on the first search, making that search slower than subsequent
+    ones.
+
+  * `desc --create-cache`:
+
+    Force the [re]construction of the name/description cache.
+
+  * `desc --delete-cache`:
+
+    Delete the name/description cache, if it exists.
```
mikemcquaid: I don't think we need to document both these options (or possibly have them at all, certainly `delete-cache`

mikemcquaid:
@hivehand That approach sounds good!

Library/Homebrew/manpages/brew.1.md
hivehand: I wanted to make it easy for people to opt out of the description-cacheing mechanism if for some reason it was annoying them, as well as to be able to force the cache's regeneration if for some reason the staleness-detection code missed a corner case.

I’ll readily admit that too much minutiae and detail in the man page can be overwhelming, though. How about I omit the man-page entries, but leave the functionality in place for the time being? (I’ll also clean up the blank lines you pointed out above.)

mikemcquaid: I think we can just tell people to delete the file manually if they have issues with either of these two cases rather than adding these flags.

hivehand: I’ll remove the flags, then. Just out of curiosity, how will we tell people to delete the file manually? In the man page, in the FAQ, or on a case-by-case basis?

mikemcquaid: Will just tell people on case-by-case I think.

hivehand:
Apologies in advance if I’m being a pest, but are there any outstanding issues regarding this pull request? The changes have been working well for me, locally, for the past two weeks and change, with minimal impact on performance. I’d still be happy to share `prof` results, once I know where and in what format to submit them.

hivehand:
And, of course, a day after my last comment, I discover the newly-added support for formula renaming in `update.rb`. I’ve added the corresponding support to the description-cache updating logic.

mikemcquaid:
The diff looks good to me. If @xu-cheng is happy with it as-is then I'll do some more local testing and merge.

xu-cheng:
LGTM

Library/Homebrew/cmd/desc.rb
```diff
@@ -0,0 +1,38 @@
...
+      if ARGV.named.empty?
+        raise FormulaUnspecifiedError
+        exit
+      end
+      results = Descriptions.named(ARGV.formulae.map(&:full_name))
+    else
+      searches = {
+        '-s' => :either,
+        '-n' => :name,
+        '-d' => :desc
```
mikemcquaid: Would be nice to also have `--search`, `--name` etc. here.

Library/Contributions/brew_bash_completion.sh
```diff
@@ -190,6 +190,18 @@ _brew_deps ()
     __brew_complete_formulae
 }
 
+_brew_desc ()
+{
+    local cur="${COMP_WORDS[COMP_CWORD]}"
+    case "$cur" in
+    --*)
+        __brewcomp "--search"
```
mikemcquaid: Does this need updated based on the flags below?

mikemcquaid:
A couple of final nits but otherwise looks and works great for me :+1:

Library/Contributions/brew_bash_completion.sh
hivehand: Good point.

Library/Homebrew/cmd/desc.rb
hivehand: Added. (In the process, I replaced the somewhat convoluted Hash logic with a case statement, which should please @bfontaine. 😉)

mikemcquaid:
Thanks for the great work here @hivehand! Merged!

hivehand:
Thank you, collectively, for the many excellent suggestions and your patience!

telemachus:
@hivehand Thanks for getting this into brew.

hivehand:
@telemachus, you’re welcome. Thanks for hopelessly addicting me to formula descriptions in the first place!

mikemcquaid:
Such a great example of a feature that we couldn't have implemented well in core as a one-off that the community picked up and got it back into core as a full-fledged, improving feature. :clap: to both of you @telemachus, @hivehand.

PR #42284, opened 2015-07-31
Fixed the description (slightly inaccurate).

oubiwann:
Also: updated the test to exercise actual LFE code, since io:format is part of
the Erlang stdlib.


Library/Formula/lfe.rb
```diff
@@ -22,6 +22,6 @@ def install
   end
 
   test do
-    system bin/"lfe", "-eval", "'(io:format \"42\")'"
+    system bin/"lfe", "-eval", '"(io:format \"~p\" (list (* 2 (lists:foldl #'+/2 0 (lists:seq 1 6)))))"'
```
dunn: Something about this is causing the test bot to choke.  It might just be the unescaped single-quote in `#'+/2`; try changing that to `#\'+/2` for starters.

oubiwann:
Indeed -- just fired up a Ruby shell and tested this out. Your suggestion works; I'll push up the change shortly.

dunn:
Woops, looks like you'll need to rebase as well.

oubiwann:
Ugh. The conflicts are a disaster. After struggling with this for 30 minutes, I'm just going to give up and start over.

PR #42285, opened 2015-07-31
pugixml 1.6 (new formula)

ajshort:


Library/Formula/pugixml.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    args = std_cmake_args
+    args << "-DBUILD_SHARED_LIBS=#{shared}"
+
+    Dir.chdir("scripts")
+
+    system "cmake", ".", *args
+    system "make", "install"
+
+    (lib/"pkgconfig/pugixml.pc").write <<-EOS.undent
```
DomT4: Would prefer to make this a `def` block rather than inline in `def install`, similar to how we handle [Lua](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/lua.rb#L107-L128).

Library/Formula/pugixml.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  depends_on "cmake" => :build
+
+  def install
+    shared = (build.with? "shared") ? "ON" : "OFF"
+
+    args = std_cmake_args
+    args << "-DBUILD_SHARED_LIBS=#{shared}"
+
+    Dir.chdir("scripts")
```
DomT4: Couldn't we just

 ```ruby
cd "scripts" do
  system "cmake", ".", *args
  system "make", "install"
end
```

?

Comment (ajshort) [inline Library/Formula/pugixml.rb]: Will do - was unaware of this syntax.

File: Library/Formula/pugixml.rb [hunk 3]
```diff
@@ -0,0 +1,61 @@
+class Pugixml < Formula
+  desc "Pugixml is a light-weight C++ XML processing library"
+  homepage "http://pugixml.org"
+  url "http://github.com/zeux/pugixml/releases/download/v1.6/pugixml-1.6.tar.gz"
```
bfontaine: Please use `https` here.

ajshort: Done.

mikemcquaid:

PR #42288, opened 2015-07-31
logstash 1.5.3

nijikon:


Library/Formula/logstash.rb
```diff
@@ -20,6 +22,6 @@ def caveats; <<-EOS.undent
 
   test do
     system "#{bin}/logstash", "--version"
-    system "/bin/echo 'hello world' | #{bin}/logstash agent -e 'input { stdin { type => stdin } } output { stdout { codec => rubydebug } }'"
+    system "/bin/echo", "'hello world' | #{bin}/logstash agent -e 'input { stdin { type => stdin } } output { stdout { codec => rubydebug } }'"
```
dunn: Can this be re-written using `pipe_output`, or would that get too messy?

nijikon: We can try ;)

DomT4: It'd be nice to kick it onto two lines instead of the one. It's currently making the formula rather wide.

mikemcquaid: :+1:

DomT4: Can we get some movement on this? Seems to be the only thing halting this being merged right now. Thanks!

PR #42289, opened 2015-07-31
cabal-install 1.22.6.0

emptyflask:
another haskell-related update

### DIFF
--- a/Library/Formula/cabal-install.rb
+++ b/Library/Formula/cabal-install.rb
@@ -1,11 +1,10 @@
 class CabalInstall < Formula
   desc "Command-line interface for Cabal and Hackage"
   homepage "https://www.haskell.org/cabal/"
-  url "https://hackage.haskell.org/package/cabal-install-1.22.2.0/cabal-install-1.22.2.0.tar.gz"
-  sha256 "25bc2ea88f60bd0f19bf40984ea85491461973895480b8633d87f54aa7ae6adb"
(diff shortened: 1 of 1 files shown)


Library/Formula/cabal-install.rb
```diff
@@ -1,11 +1,10 @@
 class CabalInstall < Formula
   desc "Command-line interface for Cabal and Hackage"
   homepage "https://www.haskell.org/cabal/"
-  url "https://hackage.haskell.org/package/cabal-install-1.22.2.0/cabal-install-1.22.2.0.tar.gz"
-  sha256 "25bc2ea88f60bd0f19bf40984ea85491461973895480b8633d87f54aa7ae6adb"
+  url "https://hackage.haskell.org/package/cabal-install-1.22.6.0/cabal-install-1.22.6.0.tar.gz"
+  sha256 "9d239e385696a7faa49f068aea451882baec6a7df26fbddbd08271c9053cb0b4"
 
   bottle do
-    revision 1
```
dunn: The whole bottle block is updated automatically, so you don't need to worry about that!

DomT4:
Will take it as-is, but yeah, @dunn is right that revisions inside the `bottle do` block don't need to be removed.

Merged in https://github.com/Homebrew/homebrew/commit/ed96b43274e2d3210b73ad78acce4e5aff1d2638. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #42291, opened 2015-07-31
oauth2_proxy 2.0.1 (new formula)

ebardsley:


Library/Formula/oauth2_proxy.rb
```diff
@@ -0,0 +1,80 @@
+require "language/go"
+
+class Oauth2Proxy < Formula
+  desc "A reverse proxy that provides authentication with Google, Github or other provider"
```
bfontaine: Please use a shorter description. “`oauth2_proxy: <description>`” should fit under 80 characters.

Library/Formula/oauth2_proxy.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def install
+    mkdir_p "#{buildpath}/src/github.com/bitly"
+    ln_s buildpath, "#{buildpath}/src/github.com/bitly/oauth2_proxy"
+
+    ENV["GOPATH"] = buildpath
+
+    Language::Go.stage_deps resources, buildpath/"src"
+    system "go", "build", "-o", "oauth2_proxy"
+    bin.install "oauth2_proxy"
```
bfontaine: You can probably merge these two lines with `system "go", "build", "-o", "#{bin}/oauth2_proxy"`

Library/Formula/oauth2_proxy.rb
```diff
@@ -0,0 +1,80 @@
...
+    doc.install "README.md"
+
+    (etc/"oauth2_proxy").mkpath
+    cfg = etc/"oauth2_proxy"/"oauth2_proxy.cfg"
+    if !cfg.exist?
+      touch cfg
+    end
```
bfontaine: You can do that in one line: `touch cfg unless cfg.exist?`

Library/Formula/oauth2_proxy.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  test do
+    system "#{bin}/oauth2_proxy", "--version"
+  end
+
+  def caveats; <<-EOS.undent
+    #{etc}/oauth2_proxy/oauth2_proxy.cfg must be filled in.
+    EOS
+  end
```
bfontaine: Please move the caveats above the test block.

Library/Formula/oauth2_proxy.rb
mikemcquaid: Just use `etc.install` here instead which can create the directory and install an empty file.

Library/Formula/oauth2_proxy.rb
ebardsley: Done. (FWIW, there seemed to be a theme in the instructions about following precedent from other repositories, which is why I copied the github project 1-liner verbatim.)

Library/Formula/oauth2_proxy.rb
ebardsley: Changed to use install. Is there a better way to pull in an empty file? (It seemed ugly explicitly do something with /dev/null.)

ebardsley:
Changes done and rebased in.

Library/Formula/oauth2_proxy.rb
mikemcquaid: Your way looks nice :+1:

Library/Formula/oauth2_proxy.rb
```diff
@@ -0,0 +1,76 @@
...
+    Language::Go.stage_deps resources, buildpath/"src"
+    system "go", "build", "-o", "#{bin}/oauth2_proxy"
+    doc.install "README.md"
+
+    touch "oauth2_proxy.cfg"
+    (etc/"oauth2_proxy").install "oauth2_proxy.cfg"
+  end
+
+  def caveats; <<-EOS.undent
+    #{etc}/oauth2_proxy/oauth2_proxy.cfg must be filled in.
```
DomT4: Can we write a basic configuration file for them?

ebardsley:
@DomT4 , done. (I forgot that there was a contributed example config.)

Library/Formula/oauth2_proxy.rb
```diff
@@ -0,0 +1,74 @@
...
+    (etc/"oauth2_proxy").install "contrib/oauth2_proxy.cfg.example"
+  end
+
+  def caveats; <<-EOS.undent
+    #{etc}/oauth2_proxy/oauth2_proxy.cfg must be filled in.
+    EOS
+  end
+
+  test do
+    system "#{bin}/oauth2_proxy", "--version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

ebardsley: I don't think so. It's a long-running daemon and doesn't have a
--validate-config-and-exit style option at the moment.

Library/Formula/oauth2_proxy.rb
```diff
@@ -0,0 +1,74 @@
...
+      <dict>
+        <key>Label</key>
+        <string>#{plist_name}</string>
+        <key>RunAtLoad</key>
+        <true/>
+        <key>KeepAlive</key>
+        <true/>
+        <key>ProgramArguments</key>
+        <array>
+            <string>#{bin}/oauth2_proxy</string>
```
DomT4: This should be `opt_bin` so it doesn't break across updates.

Library/Formula/oauth2_proxy.rb
```diff
@@ -0,0 +1,74 @@
+require "language/go"
+
+class Oauth2Proxy < Formula
+  desc "reverse proxy for authenticating users via OAuth 2 providers"
```
DomT4: Start with a capital letter please.

Library/Formula/oauth2_proxy.rb
mikemcquaid: You can use `fork do` and test the forked process.

ebardsley:
@mikemcquaid I've added an end-to-end test; let me know what you think.

@DomT4 Changes made.

Library/Formula/oauth2_proxy.rb
```diff
@@ -0,0 +1,102 @@
...
+    server = TCPServer.new(0)
+    port = server.addr[1]
+    server.close
+
+    pid = fork do
+      exec "#{bin}/oauth2_proxy",
+        "--client-id=testing",
+        "--client-secret=testing",
+        "--cookie-secret=testing",
+        "--http-address=127.0.0.1:" + port.to_s,
```
mikemcquaid: `"--http-address=127.0.0.1:#{port}"`

mikemcquaid: I'd just pick a random port here though; I don't think you need to use a TCPServer as this is primary for just using on the bots.

mikemcquaid:
Nice test :+1:

ebardsley:
@mikemcquaid Changed to use string interpolation. I'd rather keep this as a unused port, as it makes the test more robust on a wide variety of systems (eg, someone else has the same "favorite random port" as me) and it also makes it safer to run two copies of this test concurrently.

mikemcquaid:

PR #42293, opened 2015-07-31
print sandbox message

xu-cheng:
The output:
```
$ brew install tree
==> Using the sandbox
==> Downloading http://mama.indstate.edu/users/ice/tree/src/tree-1.7.0.tgz
Already downloaded: /Library/Caches/Homebrew/tree-1.7.0.tgz
==> make prefix=/Users/xucheng/Coding/Github/homebrew/Cellar/tree/1.7.0 MANDIR=/Users/xucheng/Coding/
🍺  /Users/xucheng/Coding/Github/homebrew/Cellar/tree/1.7.0: 7 files, 128K, built in 4 seconds

$ brew test tree
Testing tree
==> Using the sandbox
==> /Users/xucheng/Coding/Github/homebrew/Cellar/tree/1.7.0/bin/tree /Users/xucheng/Coding/Github/hom

$ brew postinstall tree
==> Using the sandbox
```

cc @mikemcquaid @bfontaine

### DIFF
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -14,9 +14,16 @@ class Sandbox
   end
 
   def self.print_autodisable_warning
-    unless @print_autodisable_warning
+    unless @printed_autodisable_warning
       opoo "The sandbox cannot be used in debug or interactive mode."
-      @print_autodisable_warning = true
+      @printed_autodisable_warning = true
+    end
+  end
+
+  def self.print_sandbox_message
+    unless @printed_sandbox_message
+      ohai "Using the sandbox"
+      @printed_sandbox_message = true
     end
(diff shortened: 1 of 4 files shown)


Library/Homebrew/sandbox.rb
```diff
@@ -20,6 +20,13 @@ def self.print_autodisable_warning
     end
   end
 
+  def self.print_sandbox_message
+    unless @print_sandbox_message
```
bfontaine: I find this variable name confusing, especially given it has the same name as the method. Maybe something like `printed_message`?

xu-cheng: Updated.

bfontaine:
:+1:

PR #42296, opened 2015-07-31
glib: fix for sandbox

xu-cheng:


Library/Formula/glib.rb
```diff
@@ -79,6 +79,9 @@ def install
       system "ed -s - config.h <config.h.ed"
     end
 
+    # disable creating directory for GIO_MOUDLE_DIR, we will do this manually in post_install
+    inreplace "gio/Makefile", "$(mkinstalldirs) $(DESTDIR)$(GIO_MODULE_DIR)", ""
```
mikemcquaid: Would be good to submit a patch upstream for this.

xu-cheng: Technically, this is not a bug of upstream. It's the way we handle files that requires such step. So I don't think upstream would accept this.

mikemcquaid: Yeh, I guess. Sometimes worth a try anyway? Up to you.

PR #42298, opened 2015-07-31
zbackup 1.4.2

Firefishy:
Update to zbackup 1.4.2

zbackup changelog:
* File descriptors leak fixed

### DIFF
--- a/Library/Formula/zbackup.rb
+++ b/Library/Formula/zbackup.rb
@@ -3,8 +3,8 @@ require "formula"
 class Zbackup < Formula
   desc "Globally-deduplicating backup tool (based on ideas in rsync)"
   homepage "http://zbackup.org"
-  url "https://github.com/zbackup/zbackup/archive/1.4.1.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Are the patches still necessary? It says they're fine to remove on the next release, which is presumably this one?

Firefishy:
The patches are still needed. Does not compile without them. The 1.4.2 release is a minimal release to fix a major issue with zbackup leaking file descriptors, which is particularly easy to trigger on OSX as the default process file descriptor limit is only 256.

DomT4:
Alright. Could you possibly bump those to use `sha256` as well, Thanks.

Firefishy:
Updated with sha256 for the patches too.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/7f4d4ac715ad64fb08b04a9d37dd5cc37a474a55. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #42300, opened 2015-07-31
Update tee-clc formula to 14.0.1

sofaking:


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/89a7ad9a2bd04c7710b2cd73135aab90f7802165. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

The double quotes were actually correct here - We're eliminating single quotes, not double quotes, per the [Ruby style guide](https://github.com/styleguide/ruby).

PR #42302, opened 2015-07-31
enumerate all formulae by files rather than names

xu-cheng:
Before
```
$ time brew readall
brew readall  10.63s user 0.36s system 99% cpu 11.003 total
```

After
```
$ time brew readall
brew readall  5.62s user 0.24s system 99% cpu 5.859 total
```


mikemcquaid:
:clap:

PR #42309, opened 2015-07-31
Lisp checks

dunn:
Take 2 of #41791. Changes:

1. In the cellar checks, recommend, but don't insist, that the subdirectory under `site-lisp` has the same name as the formula.
2. In the caveats, since the subdir might not be named `keg.name`, just print the path to `site-lisp`
3. An advantage of having all elisp files dumped straight into `site-lisp` was that only one directory had to be added to the user's `load-path`.  But there's a way to add all subdirectories of a directory, so we recommend they do that:
```elisp
(let ((default-directory "#{HOMEBREW_PREFIX}/share/emacs/site-lisp/"))
  (normal-top-level-add-subdirs-to-load-path))
```

### DIFF
--- a/Library/Homebrew/formula_cellar_checks.rb
+++ b/Library/Homebrew/formula_cellar_checks.rb
@@ -173,6 +173,26 @@ module FormulaCellarChecks
     EOS
   end
 
+  def check_emacs_lisp(share, name)
+    return unless share.directory?
+
+    # Emacs itself can do what it wants
+    return if name == "emacs"
+
+    elisps = (share/"emacs/site-lisp").children.select { |file| %w[.el .elc].include? file.extname }
+    return if elisps.empty?
+
+    <<-EOS.undent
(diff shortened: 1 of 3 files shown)


Library/Homebrew/formula_cellar_checks.rb
```diff
@@ -173,6 +173,29 @@ def check_python_framework_links lib
     EOS
   end
 
+  def check_emacs_lisp(share, name)
+    return unless share.directory?
+
+    # Emacs itself can do what it wants
+    return if name == "emacs"
+
+    lisps = (share/"emacs/site-lisp").children.select do |file|
```
xu-cheng: `elisps = (share/"emacs/site-lisp").children.select { |file| file.extname =~ /\.elc?/ }`

xu-cheng:
:+1:

mikemcquaid:
:+1:

PR #42310, opened 2015-07-31
The Haskell Tool Stack

bjpbakker:
Build stack from source. See also #41787.

### DIFF
--- a/Library/Formula/stack.rb
+++ b/Library/Formula/stack.rb
@@ -1,0 +1,16 @@
+class Stack < Formula
+  desc "The Haskell Tool Stack"
+  homepage "https://www.stackage.org"
+  url "https://github.com/commercialhaskell/stack/releases/download/v0.1.2.0/stack-0.1.2.0-x86_64-osx.gz"
+  sha256 "6e1039d9c5144fb03dbfb1f569a830724593191305998e9be87579d985feb36c"
+
+  def install
+    bin.install "stack-#{version}-x86_64-osx" => "stack"
+  end
+
+  test do
+    system "#{bin}/stack", "new"
+    system "#{bin}/stack", "setup"
+    system "#{bin}/stack", "build"
+  end
+end
--- a/Library/Formula/stack.rb
+++ b/Library/Formula/stack.rb
@@ -8,6 +8,8 @@ class Stack < Formula
   url "https://github.com/commercialhaskell/stack/archive/v0.1.2.0.tar.gz"
   sha256 "20ff0a36f773c2993e00c6f1bffaa33e881906d20f66cde0d557133842fc464c"
 
+  head "https://github.com/commercialhaskell/stack.git"
+
   depends_on "ghc" => :build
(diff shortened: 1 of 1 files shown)


Library/Formula/stack.rb
```diff
@@ -0,0 +1,24 @@
...
+  desc "The Haskell Tool Stack"
+  homepage "https://www.stackage.org"
+  url "https://github.com/commercialhaskell/stack/archive/v0.1.2.0.tar.gz"
+  sha256 "20ff0a36f773c2993e00c6f1bffaa33e881906d20f66cde0d557133842fc464c"
+
+  depends_on "ghc" => :build
+  depends_on "cabal-install" => :build
+
+  def install
+    system "cabal", "sandbox", "init"
```
mistydemeo: We have some cabal helpers - take a look at pandoc for an example.

bjpbakker: Awesome, thanks. I just updated the formula.

mistydemeo:
Can you add the `setup_ghc_compilers` line below the dependencies too? Thanks!

bjpbakker:
Thanks, I missed that.

mistydemeo:
Test failed on one of the bots; reads like a parallelization issue when installing a copy of GHC during `stack setup`:

```
/usr/bin/install -c -m 755 -d "/private/tmp/stack20150731-57367-1jn5376/.stack/programs/x86_64-osx/ghc-7.8.4/bin"
/usr/bin/install -c -m 755 -d "/private/tmp/stack20150731-57367-1jn5376/.stack/programs/x86_64-osx/ghc-7.8.4/lib/ghc-7.8.4/include"
/usr/bin/install -c -m 755 -d "/private/tmp/stack20150731-57367-1jn5376/.stack/programs/x86_64-osx/ghc-7.8.4/lib/ghc-7.8.4/include"
install: mkdir /private/tmp/stack20150731-57367-1jn5376/.stack/programs/x86_64-osx/ghc-7.8.4: File exists
```

There are known issues with parallelized `make install` jobs of GHC; we wrap it in `ENV.deparallelize` in the GHC formula for this reason. Looks like it applies to `stack` if `MAKEFLAGS` is in the environment when it's run. You could wrap that line in an `ENV.deparallelize` block to work around it:

```ruby


  test do
    system "#{bin}/stack", "new"
    ENV.deparallelize { system "#{bin}/stack", "setup" }
    system "#{bin}/stack", "build"
  end
```

mistydemeo:
Looks like the test timed out. Since it's downloading GHC, the amount of time this takes is variable, and we limit test runs to 5 minutes. Maybe leave the test at `system "#{bin}/stack", "new"`?

mistydemeo:
Merged in 3d1c295f68e9c37440cd493062ba05c41bc75e28 with a simplified test and as `haskell-stack`. Thanks!

bjpbakker:
Thanks for your help Misty.

PR #42311, opened 2015-07-31
Added Amdatu Bootstrap

paulbakker:
This adds Amdatu Bootstrap, with the command "bootstrap". Bootstrap is a tool to make OSGi development easier. It has a web based UI that starts up when the tool is started.

Unfortunately there are no tests in the formula because the tool starts a long running shell + UI. 
Bootstrap is around for a about a year now, and has had several releases. It is actively maintained by several contributors.

### DIFF
--- a/Library/Formula/amdatu-bootstrap.rb
+++ b/Library/Formula/amdatu-bootstrap.rb
@@ -2,7 +2,7 @@ class AmdatuBootstrap < Formula
   desc "Bootstrapping OSGi development"
   homepage "http://amdatu.org/bootstrap/intro.html"
   url "https://bitbucket.org/amdatu/amdatu-bootstrap/downloads/bootstrap-r8-bin-brew.zip"
-  sha256 "386c9f69a8bc1e12908d0a9b41c40c126c0f95627bc81d3d15bd868649798f91"
+  sha256 "f2f66af5a6b9725e260509c0a0e9996c1e3b7f696a96309b0a2a399975a9a732"
 
   def install
     libexec.install %w[amdatu-bootstrap bootstrap.jar conf]
--- a/Library/Formula/amdatu-bootstrap.rb
+++ b/Library/Formula/amdatu-bootstrap.rb
@@ -1,11 +1,11 @@
-class Bootstrap < Formula
+class AmdatuBootstrap < Formula
   desc "Bootstrapping OSGi development"
   homepage "http://amdatu.org/bootstrap/intro.html"
   url "https://bitbucket.org/amdatu/amdatu-bootstrap/downloads/bootstrap-r8-bin-brew.zip"
   sha256 "386c9f69a8bc1e12908d0a9b41c40c126c0f95627bc81d3d15bd868649798f91"
 
   def install
-    libexec.install %w[bootstrap bootstrap.jar conf]
-    bin.install_symlink "#{libexec}/bootstrap" => "bootstrap"
+    libexec.install %w[amdatu-bootstrap bootstrap.jar conf]
+    bin.install_symlink "#{libexec}/amdatu-bootstrap" => "amdatu-bootstrap"
   end
 end
--- a/Library/Formula/amdatu-bootstrap.rb
+++ b/Library/Formula/amdatu-bootstrap.rb
@@ -2,10 +2,14 @@ class AmdatuBootstrap < Formula
   desc "Bootstrapping OSGi development"
   homepage "http://amdatu.org/bootstrap/intro.html"
   url "https://bitbucket.org/amdatu/amdatu-bootstrap/downloads/bootstrap-r8-bin-brew.zip"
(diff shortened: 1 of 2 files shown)


dunn:
Beyond the lack of a test, there are a few other things flagged by the bot:
```
==> brew style bootstrap
== /usr/local/Library/Formula/bootstrap.rb ==
C:  8:  4: Inconsistent indentation detected.
C:  9:  4: Use 2 (not 1) spaces for indentation.
W: 11:  3: end at 11, 2 is not aligned with def at 8, 3

1 file inspected, 3 offenses detected

==> audit problems
bootstrap:
 * Stable: version r8 is redundant with version scanned from URL
```
Could you address those? Thanks so much!

DomT4:
Ping.

paulbakker:
Sorry it took a while, holidays :) 
Fixed the issues, besides the missing test.

dunn:
I worry that 'bootstrap' is a very generic name, maybe this should be `amdatu-bootstrap.rb`?

DomT4:
I'm less concerned about the formula name (Although I agree with Alex) than the binary name. The command `bootstrap` isn't unusual; you'd only have to call `bootstrap` instead of `./bootstrap` and you'd get a rather confusing output.

mikemcquaid:
:+1: on changing the name. The most common bootstrap people think of is the CSS framework.

paulbakker:
I agree about the naming and renamed both to amdatu-bootstrap.

Library/Formula/amdatu-bootstrap.rb
```diff
@@ -0,0 +1,11 @@
...
+  desc "Bootstrapping OSGi development"
+  homepage "http://amdatu.org/bootstrap/intro.html"
+  url "https://bitbucket.org/amdatu/amdatu-bootstrap/downloads/bootstrap-r8-bin-brew.zip"
+  sha256 "386c9f69a8bc1e12908d0a9b41c40c126c0f95627bc81d3d15bd868649798f91"
+
+  def install
+    libexec.install %w[amdatu-bootstrap bootstrap.jar conf]
+    bin.install_symlink "#{libexec}/amdatu-bootstrap" => "amdatu-bootstrap"
+  end
+end
```
mikemcquaid: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/amdatu-bootstrap.rb
```diff
@@ -0,0 +1,11 @@
+class AmdatuBootstrap < Formula
+  desc "Bootstrapping OSGi development"
+  homepage "http://amdatu.org/bootstrap/intro.html"
+  url "https://bitbucket.org/amdatu/amdatu-bootstrap/downloads/bootstrap-r8-bin-brew.zip"
```
mikemcquaid: Why is this `brew`? Is there something special you're doing for Homebrew here?

paulbakker:
I had an original zip that was without the -brew with existing links to it. The -brew zip contains a startup script, that's all that's different. For future releases this will just be included in the release, and the zip will not have a -brew in it's name.

mikemcquaid:
@paulbakker :ok: :cool:.

DomT4:
Checksum is failing now:

```
==> Verifying amdatu-bootstrap-r8.zip checksum
Error: SHA256 mismatch
Expected: 386c9f69a8bc1e12908d0a9b41c40c126c0f95627bc81d3d15bd868649798f91
Actual: 6dce4bfa1029f28ce29d8519f6b05f3356c22b2be059ad14c6cf4f6c4524e8e2
Archive: /Library/Caches/Homebrew/amdatu-bootstrap-r8.zip
```

paulbakker:
Ping :) Fixed the checksum...

dunn:
Can we get a test as well? Thanks!

Library/Formula/amdatu-bootstrap.rb
DomT4: Any chance of a test here, per Mike's comment?

bfontaine:
Ping?

paulbakker:
Working on it. Currently it will start the UI automatically, which is long running and not fit for a test. I'm changing the launching code so that it can just print a version or something like that.

paulbakker:
Added a test and all checks seem to pass :)

Library/Formula/amdatu-bootstrap.rb
bfontaine: What is the difference between this archive and `bootstrap-r8-bin.zip`?

paulbakker: This one contains the amdatu-bootstrap.sh script that I added for Brew. For next releases that will be included directly in the release.

paulbakker:
Anything else that needs to happen before this is merged?

mikemcquaid:

For future reference the preferred commit message format for new formulae is `amdatu-bootstrap 8 (new formula)`.

PR #42314, opened 2015-07-31
moreutils: confict with parallel is unconditional

dunn:
Our conflict resolution logic can't currently manage conflicts that
depend on whether a certain option is used, so remove the misleading
language and pull the `conflicts_with` out of the conditional.

see also 668fd15632f92369d9f13a53a07ca9d401ee2d64


DomT4:
:+1:

bfontaine:
:+1:

PR #42315, opened 2015-07-31
Planck 1.0

derekslager:
This is a New Formula, Planck is a self-contained ClojureScript REPL.

### DIFF
--- a/Library/Formula/planck.rb
+++ b/Library/Formula/planck.rb
@@ -1,0 +1,15 @@
+class Planck < Formula
+  desc "A command line ClojureScript REPL for OS X."
+  homepage "http://planck.fikesfarm.com/"
+  url "http://planck.fikesfarm.com/planck.gz"
+  sha256 "c8f3d1b270ab1175cffda119fe186e12d5a2aa77018e212a5ad62628c13517b2"
+  version "1.0"
+
+  def install
+    bin.install "planck"
+  end
+
+  test do
+    system "#{bin}/planck", "-e", "'(- 1 1)'"
+  end
+end


Library/Formula/planck.rb
```diff
@@ -0,0 +1,15 @@
+class Planck < Formula
+  desc "A command line ClojureScript REPL for OS X."
+  homepage "http://planck.fikesfarm.com/"
+  url "http://planck.fikesfarm.com/planck.gz"
+  sha256 "c8f3d1b270ab1175cffda119fe186e12d5a2aa77018e212a5ad62628c13517b2"
+  version "1.0"
+
+  def install
+    bin.install "planck"
```
dunn: Is this a binary? If it is, and it can't be built from source, we have a [homebrew-binary](https://github.com/Homebrew/homebrew-binary) repository that might be a better fit.

Library/Formula/planck.rb
```diff
@@ -0,0 +1,15 @@
+class Planck < Formula
+  desc "A command line ClojureScript REPL for OS X."
+  homepage "http://planck.fikesfarm.com/"
+  url "http://planck.fikesfarm.com/planck.gz"
```
mistydemeo: Is there a versioned URL available?

mistydemeo:
I see this is a binary. Can this be built from source?

derekslager:
Source is available at https://github.com/mfikes/planck

I can work with the author to provide a versioned URL.

mistydemeo:
Thanks! If they tag a release, we could use the Github tarball and make the formula build from source.

derekslager:
Got it, thanks!

derekslager:
The build has some external dependencies, so in current form this is probably a better fit for `homebrew-binary`. I will relocate it there.

Assuming I wanted to put together a "build from source" build, are formulas which pull large external repositories and interact with systems such as CocoaPods discouraged?

mistydemeo:
Not necessarily, though it's easier if we can fetch those dependencies using Homebrew's mechanisms. Take a look at Go recipes like packer, for example; Homebrew manages fetching and staging the go deps instead of using `go get`.

derekslager:
Thanks. Moved to https://github.com/Homebrew/homebrew-binary/pull/256

dunn:
Just to be clear, if it _can_ be build from source, even if it's a bit involved, then it should be here, not in homebrew-binary.

derekslager:
Thanks, Alex, I'm working with the author on this.

PR #41869, opened 2015-07-18
Add fugu 1.1.1

neronplex:

PR #42323, opened 2015-08-01
Add fugu 1.1.1

neronplex:
Docker command wrapper tool.

### DIFF
--- a/Library/Formula/fugu.rb
+++ b/Library/Formula/fugu.rb
@@ -1,0 +1,39 @@
+require "language/go"
+
+class Fugu < Formula
+  desc "Docker command wrapper."
+  homepage "https://github.com/mattes/"
+  url "https://github.com/mattes/fugu/archive/v1.1.1.tar.gz"
+  sha256 "94ec61037c3afa1267ea990ffd03ba1d0d1628926b3fdde0133fece36fa81929"
+
+  depends_on "go" => :build
+  depends_on "docker" => :recommended
+
+  go_resource "github.com/tools/godep" do
+    url "https://github.com/tools/godep.git", :revision => "e2d1eb1649515318386cc637d8996ab37d6baa5e"
+  end
+
+  go_resource "github.com/kr/fs" do
+    url "https://github.com/kr/fs.git", :revision => "2788f0dbd16903de03cb8186e5c7d97b69ad387b"
+  end
+
+  def install
+    mkdir_p buildpath/"src/github.com/mattes/"
+    ln_s buildpath, buildpath/"src/github.com/mattes/fugu"
+
+    ENV["GOPATH"] = buildpath
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    cd buildpath/"src/github.com/tools/godep" do
+      system "go", "install"
+    end
+
+    cd buildpath/"fugu" do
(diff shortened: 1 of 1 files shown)


dunn:
Did you mean to open this in homebrew-binary (#41869)?

neronplex:
"Homebrew-binary" is later to know was that it is a policy that does not accept as a principle that can be compiled in the open source.Because it was confirmed that can be compiled in the "homebrew", it issued a request this time again.

dunn:
Then it will have to be actually built from source; right now the formula just installs the precompiled binary.

neronplex:
Whether you should have specified the compressed file in the source code in the Formula, but this has not been compiled with homebrew indeed?

dunn:
Yes, "built from source", means it is compiled during the installation by Homebrew, and not previously by the upstream developer.

dunn:
See [the platinum searcher](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/the_platinum_searcher.rb) for an example of a Go project that's built from source.

neronplex:
Oh, I see. Please wait...

neronplex:
Fixed formula. Please review it.

Library/Formula/fugu.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  go_resource "github.com/howeyc/gopass" do
+    url "https://github.com/howeyc/gopass.git", :revision => "10b54de414cc9693221d5ff2ae14fd2fbf1b0ac1"
+  end
+
+  go_resource "golang.org/x/crypto" do
+    url "https://github.com/golang/crypto.git", :revision => "7f27901cb9636a61b9b0b1bab2679b2a9a12eb38"
+  end
+
+  go_resource "github.com/docker/docker" do
```
dunn: You already have `depends_on "docker"` above. Is it possible to avoid duplicating the dependency here?

dunn:
The godeps file lists a lot more dependencies: https://github.com/mattes/fugu/blob/v1/Godeps/Godeps.json 

How are you getting by without them all?

neronplex:
Some If you try to use the godep it becomes error in the package of the resolution. The package I was writing all by trial and error.

neronplex:
It was modified to use the godep.

dunn:
The test bot is failing to install:
```
==> Downloading https://github.com/mattes/fugu/archive/v1.1.1.tar.gz
Already downloaded: /Library/Caches/Homebrew/fugu-1.1.1.tar.gz
==> Verifying fugu-1.1.1.tar.gz checksum
tar xf /Library/Caches/Homebrew/fugu-1.1.1.tar.gz
Error: No such file or directory - /private/tmp/fugu20150804-51418-yxpg8a/fugu-1.1.1/src/github.com/tools/godep
```

neronplex:
I'm sorry. Verification at whether hand was left cache was insufficient. In the most recent commit to being successful.

neronplex:
ping.

tdsmith:
Thanks for your pull request! :tada: Homebrew thrives because of contributions like this one. Pulled with small tweaks in b4d0783cca7dbd2.

neronplex:
Thank you for approval!

PR #42329, opened 2015-08-01
liblinear: add test

dplarson:
Add a basic test, re-using the example classification training data from the LIBSVM formula test block.


xu-cheng:

PR #42333, opened 2015-08-01
Remove indefinite article from description

afh:
The changes were made with the following shell command:
for formula in $(grep -l 'desc "[Aa]n\?\s' Library/Formula/*.rb); do
  echo $formula
  vim -c '/desc "/s/[Aa]n\? //|normal f"gU2lZZ' $formula
done

### DIFF
--- a/Library/Formula/ack.rb
+++ b/Library/Formula/ack.rb
@@ -1,5 +1,5 @@
 class Ack < Formula
-  desc "A search tool like grep, but optimized for programmers"
+  desc "Search tool like grep, but optimized for programmers"
   homepage "http://beyondgrep.com/"
   url "http://beyondgrep.com/ack-2.14-single-file"
   sha256 "1d203cfbc52ce8f49e3992be1cd3e4d7d5dfb7daa3739e8628aa9858ccc5b9df"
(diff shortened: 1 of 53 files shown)


mikemcquaid:
@afh Cancelling the CI build. Please don't submit PRs that change lots (e.g. 5+) formula without notifying us as it messes with the bot.

afh:
I'll remember it for future PRs.

How would you like to be notified?

mikemcquaid:
Create an issue, ideally. If it's something like this that's running a script paste the script and maintainers may run it locally instead.

xu-cheng:
Thanks @afh

afh:
@mikemcquaid That makes perfect sense.

@xu-cheng :+1: You're welcome. I'm uncertain what you mean by "submit[ting] a PR on `desc` audit rule."

bfontaine:
@afh What I think @xu-cheng means is you can submit a PR to add a `brew audit` rule on `desc`s starting with an indefinite article.

afh:
@bfontaine Thanks for clarifying that.

@xu-cheng Here's the PR #42404.

PR #42336, opened 2015-08-01
cowpatty: some small patches which makes correct use of stdin

timemaster5:
Now you can be able to pass output from pyrit over stdout/stdin directly to cowpatty.

### DIFF
--- a/Library/Formula/cowpatty.rb
+++ b/Library/Formula/cowpatty.rb
@@ -11,7 +11,12 @@ class Cowpatty < Formula
     sha256 "35fba0f92c5e8fb0710453d0c2c5fe5e9c64857dd53b219977871340b22c4942"
   end
 
+  def patches
+    [ "http://proton.cygnusx-1.org/~edgan/cowpatty/cowpatty-4.6-fixup16.patch" ]
+  end
+
   def install
+    system "make"
     system "make", "BINDIR=#{bin}", "install"
   end
 
--- a/Library/Formula/cowpatty.rb
+++ b/Library/Formula/cowpatty.rb
@@ -11,8 +11,9 @@ class Cowpatty < Formula
     sha256 "35fba0f92c5e8fb0710453d0c2c5fe5e9c64857dd53b219977871340b22c4942"
   end
 
-  def patches
-    [ "http://proton.cygnusx-1.org/~edgan/cowpatty/cowpatty-4.6-fixup16.patch" ]
+  patch do
+    url "http://proton.cygnusx-1.org/~edgan/cowpatty/cowpatty-4.6-fixup16.patch"
+    sha256 "2cf392c4dec543a6d77d7d420dae231938df3ba427e30f05256f6ab5baa47e5d"
   end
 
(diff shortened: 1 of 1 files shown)


Library/Formula/cowpatty.rb
```diff
@@ -11,7 +11,12 @@ class Cowpatty < Formula
     sha256 "35fba0f92c5e8fb0710453d0c2c5fe5e9c64857dd53b219977871340b22c4942"
   end
 
+  def patches
+    [ "http://proton.cygnusx-1.org/~edgan/cowpatty/cowpatty-4.6-fixup16.patch" ]
```
dunn: Has this patch been accepted upstream?

mikemcquaid: Could you try and open a pull request? This should help and I'm happy to walk you through anything else: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged

Thanks!

mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

timemaster5: Thins patch probably never will be in upstream because this project is very old and without a maintainer. Latest version is from July 03 2009 and homepages are not working. However this patch is absolutely neccessary if you want to safe half time of wpa cracking with the use of stdout from pyrit or other similar software and it is recommended way how to get cowpatty on Debian, Ubuntu and other distros. Unfortunatelly only references to that are on several blogs..

mikemcquaid: Would still like it submitted anyway, thanks.

DomT4: Should be:

```ruby
patch do
  url ""
  sha256 ""
end
```

timemaster5: Thanks, fixed..

mikemcquaid: The formula has a homepage and therefore a developer. I'd rather we didn't accept this if upstream is now unmaintained.

timemaster5: Ok, sorry for that last time web pages doesn't work. I wrote a message to the address on the end of README file, so we will see. Thanks.

apjanke: Ever get a response on this? If not, I think we'd prefer to pass on it.

timemaster5: No response at all :(

tdsmith:
It sounds like we should boneyard cowpatty.

mikemcquaid:
@tdsmith Agreed 👍

apjanke:
Cowpatty has been boneyarded, in ba4493692b9cddc68bdcbe26ade82d69bac5932b.

PR #42337, opened 2015-08-01
PEGTL 1.1.0

guoxiao:


mikemcquaid:
Given this is only a single header I wonder if it's more useful to have it in Homebrew or just have projects vendor it?

guoxiao:
So we don't accept header-only library in homebrew now?

guoxiao:
> Given this is only a single header

It's a header `pegtl.hh` and a folder of many headers.

mikemcquaid:
> So we don't accept header-only library in homebrew now?

I haven't decided either way; I'm waiting for other maintainers to give their thoughts.

DomT4:
What is the upstream `Makefile` intended to do, and why aren't we using it, out of curiosity?

guoxiao:
@DomT4 The upstream `Makefile` is used by travis-ci to build all examples and unittests.

DomT4:
Alright.

I guess I'm not fundamentally opposed to header-only libraries in Homebrew, but I'm not sure how widely used they'll ever be.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `pegtl 1.1.0 (new formula)`.

PR #42344, opened 2015-08-01
fix parallel build issue and cpu collection issue

### LINKED ISSUE #42343: collectd fails to build in parallel. In addition, cpu not working in 5.5.0.
Two issues:
* collectd often fails to build in parallel.
* cpu metric collection not working in 5.5.0 on osx. The error is `cpu plugin: processor_info returned only 4 elements`.

This patch fixes both.

references: 
* https://github.com/collectd/collectd/issues/1146 (parallel build issue)
* http://mailman.verplant.org/pipermail/collectd/2015-May/006546.html (cpu stats issue)

```
diff --git a/Library/Formula/collectd.rb b/Library/Formula/collectd.rb
index da24134..dfb4db4 100644
--- a/Library/Formula/collectd.rb
+++ b/Library/Formula/collectd.rb
@@ -19,6 +19,13 @@ class Collectd < Formula
   end
 
+  stable do
+      patch do
+        url "https://github.com/collectd/collectd/commit/e0683047a42e217c352c2419532b8e029f9f3f0a.diff"
+        sha256 "7053170a072d27465b69eed269d32190ec810bcb0db59f139a1682e71a326fdd"
+      end
+  end
+
   # Will fail against Java 1.7
   option "with-java", "Enable Java 1.6 support"
   option "with-debug", "Enable debug support"
@@ -39,6 +46,10 @@ class Collectd < Formula
   end
 
   def install
+    # collectd breaks with makejobs
+    # see: https://github.com/collectd/collectd/issues/1146
+    ENV.deparallelize
+
     args = %W[
       --disable-debug
       --disable-dependency-tracking
```

cactus:
Two issues:
*   collectd often fails to build in parallel.
*   cpu metric collection not working in 5.5.0 on osx. The error is:
    `cpu plugin: processor_info returned only 4 elements`.

This patch fixes both.

References:
*   https://github.com/collectd/collectd/issues/1146 (parallel build
    issue)
*   http://mailman.verplant.org/pipermail/collectd/2015-May/006546.html
    (cpu stats issue)

fixes #42343

### DIFF
--- a/Library/Formula/collectd.rb
+++ b/Library/Formula/collectd.rb
@@ -1,9 +1,6 @@
 class Collectd < Formula
   desc "Statistics collection and monitoring daemon"
   homepage "https://collectd.org/"
-  url "https://collectd.org/files/collectd-5.5.0.tar.bz2"
-  mirror "http://pkgs.fedoraproject.org/repo/pkgs/collectd/collectd-5.5.0.tar.bz2/c39305ef5514b44238b0d31f77e29e6a/collectd-5.5.0.tar.bz2"
-  sha256 "847684cf5c10de1dc34145078af3fcf6e0d168ba98c14f1343b1062a4b569e88"
 
   bottle do
     sha256 "b35bfeb9b9e5318e502c7e2a76eee1508dcf3739df96554b89f3ee49e38d6b48" => :yosemite
@@ -17,6 +14,17 @@ class Collectd < Formula
     depends_on "libtool" => :build
     depends_on "automake" => :build
     depends_on "autoconf" => :build
+  end
+
+  stable do
+    url "https://collectd.org/files/collectd-5.5.0.tar.bz2"
+    mirror "http://pkgs.fedoraproject.org/repo/pkgs/collectd/collectd-5.5.0.tar.bz2/c39305ef5514b44238b0d31f77e29e6a/collectd-5.5.0.tar.bz2"
+    sha256 "847684cf5c10de1dc34145078af3fcf6e0d168ba98c14f1343b1062a4b569e88"
+
+    patch do
+      url "https://github.com/collectd/collectd/commit/e0683047a42e217c352c2419532b8e029f9f3f0a.diff"
+      sha256 "7053170a072d27465b69eed269d32190ec810bcb0db59f139a1682e71a326fdd"
+    end
   end
 
   # Will fail against Java 1.7
@@ -39,6 +47,10 @@ class Collectd < Formula
   end
 
   def install
+    # collectd breaks with makejobs
+    # see: https://github.com/collectd/collectd/issues/1146
+    ENV.deparallelize
+
     args = %W[
       --disable-debug
       --disable-dependency-tracking


Library/Formula/collectd.rb
```diff
@@ -19,6 +19,13 @@ class Collectd < Formula
     depends_on "autoconf" => :build
   end
 
+  stable do
```
DomT4: Everything should go inside the `stable do` block related to stable, so:

```ruby
class Collectd < Formula
desc "Statistics collection and monitoring daemon"
homepage "https://collectd.org/"

stable do
  url "https://collectd.org/files/collectd-5.5.0.tar.bz2"
  mirror ""
  sha256 "847684cf5c10de1dc34145078af3fcf6e0d168ba98c14f1343b1062a4b569e88"

  # I fix xyz
  patch do
    url ""
    sha256 ""
  end
end
```

cactus:
note. I tried putting the patch in an `__END__` segment, but it failed to apply that way. Not sure exactly why. I opted instead to just pull the diff from github, which is a bit suboptimal but possibly not the worst thing ever.

Library/Formula/collectd.rb
cactus: done

DomT4:
>  I opted instead to just pull the diff from github, which is a bit suboptimal but possibly not the worst thing ever.

This is our preferred method. The `__END__` patches can be grumpy, and people tend to accidentally add whitespacing to them when they make formula changes. External patches are thankfully free of that issue.

cactus:
Thanks for your help. Let me know if there are any other needed changes.

cactus:
did a rebase/squash so this is a single commit, as specified in [How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged)](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md)

Library/Formula/collectd.rb
```diff
@@ -19,6 +16,17 @@ class Collectd < Formula
...
+      url "https://collectd.org/files/collectd-5.5.0.tar.bz2"
+      mirror "http://pkgs.fedoraproject.org/repo/pkgs/collectd/collectd-5.5.0.tar.bz2/c39305ef5514b44238b0d31f77e29e6a/collectd-5.5.0.tar.bz2"
+      sha256 "847684cf5c10de1dc34145078af3fcf6e0d168ba98c14f1343b1062a4b569e88"
+
+      patch do
```
DomT4: From the spacing, it looks like you've indented with tabs instead of spaces here. Should be spaces :). Some text editors (Sublime, for one, I believe) default to tabs unless told otherwise, so didn't necessarily need direct user choice.

cactus: I use vim. These are spaces. Just 4 instead of 2 -- I use `ts=4 sw=4` as my default. Does homebrew adhere to a 2 space style guide?

cactus: fixed. set to two spaces.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `collectd: fix parallel build and collection issues`.

PR #41549, opened 2015-07-09
new formula: globe

xdbr:
globe shows an ASCII graphic of the currently-lighted face of the Earth,
similar to phoon

PR #42352, opened 2015-08-02
online audit: use GET instead of HEAD

bfontaine:
I also removed the `-L` flag which is already present in `HOMEBREW_CURL_ARGS`.

The audit fails on some formulae because the server doesn’t support `HEAD` requests and return a `501 Not Implemented` error, e.g. #41549.


xu-cheng:
:+1:

PR #42354, opened 2015-08-02
Core files style updates.

mikemcquaid:
Mass update all the core files with `brew style --fix` (which uses `rubocop`). Committing as @BrewTestBot to make it a bit more obvious that it was an automatic change. Tests pass, please review ASAP, thanks!


xu-cheng:
rubocop will turn this:
```ruby
a = if foo?
  foo
else
  bar
end
```

to this:
```ruby
a = if foo?
     foo
    else
     bar
    end
```

Any chance to disable this rule?

mikemcquaid:
@xu-cheng Yeh, :+1: on changing that. Mind submitting a PR for the Rubocop rule changes (https://github.com/bbatsov/rubocop) and I'll rebase/redo this commit after that?

xu-cheng:
Also maybe disable the rule turn `$1` to `Regexp.last_match(1)`.

mikemcquaid:
@xu-cheng The reasoning for that rule it that generally using global, implicitly set variables like that is frowned upon. I don't have strong feelings either way but I do see the justification. Just submit a PR for any Rubocop checks you think should be changed or disabled and we can discuss them there. Chances are any automatic changes you want to avoid I'll be :+1: with avoiding too.

mikemcquaid:
@xu-cheng Updated.

xu-cheng:
Are you sure? It looks like to be the old version.

mikemcquaid:
Definitely sure. There's perhaps other rules we need to enable.

xu-cheng:
But the new rule I pushed doesn't take effect on your PR. I downloaded your PR and re-run rubocop locally it will work. Probably, you forget to rebase?

mikemcquaid:
@xu-cheng Was missing one file (`ENV/4.3/cc`) but otherwise up-to-date. Have pushed everything now.

xu-cheng:
This time you update formulae as well, it’s this intentional.

Either way, :+1:

mistydemeo:
How many downstream projects do we have merging in changes from Homebrew, aside from Linuxbrew and Tigerbrew? (I'm guessing not many.) This is likely to lead to a lot of merge conflicts when merging in changes, though if it's just those two projects not too many people will be affected.

mikemcquaid:
@mistydemeo Yeh, not many or none. The merge conflicts will be an annoying one-time thing but I figured it's probably better than gradual cleanup. That reminds me that once we get the core Rubocop-clean we should make sure to setup `brew-test-bot` so they don't creep in again.

mikemcquaid:
@Homebrew/linux heads up that this is coming your way. It may be some one-off pain but once we get all the rules fixed we'll add a CI job. Rubocop should be runnable for the Linuxbrew code too which may allow you to simplify both the merge and the fixing process for your own code.

sjackman:
@Homebrew/linux has one lonely member (me). Any other Homebrew/homebrew folks interested in joining the team? I'm looking at you, @xu-cheng . =)

xu-cheng:
@sjackman Although I can help to solve the problem I encountered in Linuxbrew now an then, it's unlikely for me to have extra time to maintain Linuxbrew. Sorry about this but maintaining Homebrew already eats up most of my free time. 

FWIW, I'm fully in favor of more Linuxbrew maintainers. Maybe rest of @Homebrew/owners are interesting.

PR #42356, opened 2015-08-02
zopfli: rework HEAD build

ojab:
Zopfli development was moved to github & libzopfli/libzopflipng shared libs was added.
There is an issue with formula: src/zopfli{,png}/ contains both headers & .c files, so right now .c files are also installed, which is wrong.

I'm not sure how should I deal with it, is something like 
```ruby
%w(zopfli zopflipng zopflipng/lodepng).each do |dir|
  Dir["src/#{dir}/*.h"].each { |f| include.install i => dir }
end
```
ok or there is a better way to archive the same?

### DIFF
--- a/Library/Formula/zopfli.rb
+++ b/Library/Formula/zopfli.rb
@@ -1,21 +1,27 @@
-require 'formula'
-
 class Zopfli < Formula
   desc "New zlib (gzip, deflate) compatible compressor"
-  homepage 'https://code.google.com/p/zopfli/'
-  url 'https://zopfli.googlecode.com/files/zopfli-1.0.0.zip'
-  sha1 '98ea00216e296bf3a13e241d7bc69490042ea7ce'
-  head 'https://code.google.com/p/zopfli/', :using => :git
+  homepage "https://github.com/google/zopfli"
+  url "https://zopfli.googlecode.com/files/zopfli-1.0.0.zip"
+  sha256 "e20d73b56620285e6cce5b510d8e5da6835a81940e48cdf35a69090e666f3adb"
+  head "https://github.com/google/zopfli.git"
 
   def install
-    # Makefile hardcodes gcc
-    inreplace 'makefile', 'gcc', ENV.cc
-    system 'make', '-f', 'makefile'
-    bin.install 'zopfli'
-    if build.head?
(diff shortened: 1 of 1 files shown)


Library/Formula/zopfli.rb
```diff
@@ -1,21 +1,27 @@
...
-      system 'make', '-f', 'makefile', 'zopflipng'
-      bin.install 'zopflipng'
+    if build.stable?
+      # Makefile hardcodes gcc
+      inreplace "makefile", "gcc", ENV.cc
+      system "make", "-f", "makefile"
+    else
+      inreplace "Makefile", "gcc", ENV.cc
+      inreplace "Makefile", "g++", ENV.cxx
+      inreplace "Makefile", "-soname", "-o"
```
mikemcquaid: Can you please submit these patches to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Library/Formula/zopfli.rb
```diff
@@ -1,21 +1,27 @@
...
+      system "make", "zopfli", "zopflipng", "libzopfli", "libzopflipng"
+      bin.install "zopflipng"
+      lib.install "libzopfli.so.1" => "libzopfli.dylib"
+      lib.install "libzopflipng.so.1" => "libzopflipng.dylib"
```
mikemcquaid: And these changes too, thanks. Also, is there no `make install` we can use?

Library/Formula/zopfli.rb
ojab: There is a [pull request](https://github.com/google/zopfli/pull/57) with build system overhaul and [another one](https://github.com/google/zopfli/pull/58) adding CMake support, but none of them are answered by zopfli devs. I doubt that another one pull request with Makefile fixes will be useful.
May I just drop patch to gist and use it from there?

Library/Formula/zopfli.rb
ojab: There isn't `make install`, but I can add it to the patch.

Library/Formula/zopfli.rb
mikemcquaid: If they aren't accepting patches I'm hesitant to include this software in Homebrew as it sounds unmaintained.

PR #42364, opened 2015-08-02
facebook-flint 1.0 (new formula)

nlohmann:
flint is an open-source lint program for C++ developed and used at Facebook.

### DIFF
--- a/Library/Formula/facebook-flint.rb
+++ b/Library/Formula/facebook-flint.rb
@@ -1,0 +1,36 @@
+class FacebookFlint < Formula
+  desc "Flint - An open-source lint program for C++ developed by, and used at Facebook."
+  homepage "https://github.com/facebook/flint"
+  url "https://github.com/facebook/flint.git"
+  sha256 ""
+  version "1.0"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "dmd" => :build
+  depends_on "folly" => :build
+  depends_on "gnu-sed" => :build
+  depends_on "libtool" => :build
+  
+  def install
+    ENV.deparallelize
+
+    # remove C++ parts (wait for https://github.com/facebook/flint/pull/31)
+    system "gsed -i 's/SUBDIRS = . cxx//' Makefile.am"
+    system "gsed -i 's@cxx/Makefile@@' configure.ac"
+    
+    system "autoreconf -iv"
+
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+
+    system "make", "install"
+  end
+  
+  test do
+    (testpath/"test.h").write("class C { C(int a); C(const C&&); };")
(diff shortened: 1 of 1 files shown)


Library/Formula/facebook-flint.rb
```diff
@@ -0,0 +1,36 @@
+class FacebookFlint < Formula
+  desc "Flint - An open-source lint program for C++ developed by, and used at Facebook."
+  homepage "https://github.com/facebook/flint"
+  url "https://github.com/facebook/flint.git"
```
DomT4: If there's no stable release you'll need to file this at the `head-only` tap. Otherwise, please use a tag or tagged tarball and SHA256.

DomT4: Ping. They may well be willing to tag a release if you ask them.

Library/Formula/facebook-flint.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "dmd" => :build
+  depends_on "folly" => :build
+  depends_on "gnu-sed" => :build
+  depends_on "libtool" => :build
+  
+  def install
+    ENV.deparallelize
```
mikemcquaid: Can you please submit this issue to the upstream developers of this project and add a link to the upstream issue submission in a comment in the formula file. Thanks!

Library/Formula/facebook-flint.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    # remove C++ parts (wait for https://github.com/facebook/flint/pull/31)
+    system "gsed -i 's/SUBDIRS = . cxx//' Makefile.am"
+    system "gsed -i 's@cxx/Makefile@@' configure.ac"
```
mikemcquaid: Is this all `gsed` is needed for? Would be better to apply the pull as a `patch do` with `https://github.com/facebook/flint/pull/31.patch`

mikemcquaid: Also just FYI the `inreplace` method is best for this.

nlohmann:
Thanks for the help. However, I won't be able to work on this this week. :-S

Library/Formula/facebook-flint.rb
nlohmann: I asked them: https://github.com/facebook/flint/issues/2

Library/Formula/facebook-flint.rb
nlohmann: I'll check the syntax of these calls.

mcandre:
+1, good job nlohmann!

DomT4:
Going to close here because there seems to be minimal upstream activity lately & people have been asking for a tagged release for a year and a half now without maintainer feedback.

Happy to take another look at this if you submit it as a head-only formula [here](https://github.com/Homebrew/homebrew-head-only) though. At least that way it can be used via Homebrew and we can move it to the core if a tagged release does become available.

PR #42366, opened 2015-08-02
scmpuff v0.1.1 (new formula)

mroth:
> scmpuff makes working with git from the command line quicker by substituting numeric shortcuts for filenames. https://mroth.github.io/scmpuff/

Note that this is a self-submission from the author (me), however I am doing this because it is still unfortunately quite tricky to package a Go-based project with vendored dependencies for reliable compilation in Homebrew vis-a-vis GOPATH, and a number of users who tried to set up a formula weren't able to get it working and contacted me.  I'd rather the formula that ends up in Homebrew be done in a careful way--I annotated this formula via comments to hopefully aide future formula updaters.

### DIFF
--- a/Library/Formula/scmpuff.rb
+++ b/Library/Formula/scmpuff.rb
@@ -1,0 +1,23 @@
+class Scmpuff < Formula
+  desc "Adds numbered shortcuts for common git commands"
+  homepage "https://mroth.github.io/scmpuff/"
+  url "https://github.com/mroth/scmpuff/archive/v0.1.1.tar.gz"
+  sha256 "cec3c9df41acb1735f2e8c1c9840d0481af0d996690f5a19a0a8fc4f06f97370"
+
+  depends_on "go" => :build
+
+  def install
+    mkdir_p buildpath/"src/github.com/mroth"
+    ln_s buildpath, buildpath/"src/github.com/mroth/scmpuff"
+    ENV["GOPATH"] = buildpath
+
+    # scmpuff's build script normally does version detection which depends on
+    # being checked out via git repo -- instead have homebrew specify version.
+    system "go", "build", "-o", "#{bin}/scmpuff", "-ldflags", "-X main.VERSION #{version}"
+  end
+
+  test do
+    ENV["e1"] = "abc"
+    assert_equal "abc", shell_output("#{bin}/scmpuff expand 1").strip
+  end
+end


Library/Formula/scmpuff.rb
```diff
@@ -0,0 +1,28 @@
...
+    # do a compile, manually specifying version so that scmpuff's build script
+    # does not do it's normal version detection which depends on being in a git
+    # repository -- instead let homebrew recipe specify the version number.
+    ENV["GOPATH"] = gopath
+    system "go build -o bin/scmpuff -ldflags '-X main.VERSION #{version}'"
+    bin.install "bin/scmpuff"
+  end
+
+  test do
+    assert_equal "abc", shell_output("e1=abc #{bin}/scmpuff expand 1").strip
```
bfontaine: Please use `ENV["e1"] = "abc"` to set the environment variable.

Library/Formula/scmpuff.rb
```diff
@@ -0,0 +1,28 @@
...
+    gopath = "#{buildpath}/.go"
+    mkdir_p "#{gopath}/src/github.com/mroth"
+    ln_s buildpath, "#{gopath}/src/github.com/mroth/scmpuff"
+
+    # do a compile, manually specifying version so that scmpuff's build script
+    # does not do it's normal version detection which depends on being in a git
+    # repository -- instead let homebrew recipe specify the version number.
+    ENV["GOPATH"] = gopath
+    system "go build -o bin/scmpuff -ldflags '-X main.VERSION #{version}'"
+    bin.install "bin/scmpuff"
```
bfontaine: You can merge the two lines above by using `#{bin}/scmpuff` for the target binary. Also, split the string like so:

```rb
system "go", "build", "-o", "#{bin}/scmpuff", "-ldflags", "-X main.VERSION #{version}"
```

Library/Formula/scmpuff.rb
```diff
@@ -0,0 +1,28 @@
+class Scmpuff < Formula
+  homepage "https://mroth.github.io/scmpuff/"
```
bfontaine: Please add a short description using `desc "…"` above this line.

mroth:
Thanks for the notes!  Just pushed a new diff (squashed onto existing commit) with those changes.

Library/Formula/scmpuff.rb
```diff
@@ -0,0 +1,29 @@
+class Scmpuff < Formula
+  desc "Adds numbered shortcuts for common git commands"
+  homepage "https://mroth.github.io/scmpuff/"
+  url "https://github.com/mroth/scmpuff/archive/v0.1.1.tar.gz"
+  sha256 "cec3c9df41acb1735f2e8c1c9840d0481af0d996690f5a19a0a8fc4f06f97370"
+
+  depends_on "go" => :build
+
+  def install
+    # Go assumes things being compiled are located within a specific directory
```
DomT4: We can kill the notes. We do this in enough Go formulae that the process is apparently enough.

Library/Formula/scmpuff.rb
```diff
@@ -0,0 +1,29 @@
...
+    # hierarchy in the user $GOPATH. Since it insists on absoute import paths,
+    # we need to temporarily create this scenario so the vendored package
+    # hierarchy is find-able by gotools at compile time.
+    gopath = "#{buildpath}/.go"
```
DomT4: It's a nit, but you can just use `gopath = buildpath/".go"` here.

mikemcquaid: Just use `ENV["GOPATH"] = buildpath` and then use `buildpath` elsewhere. Doesn't really matter but more consistent with other formulae.

Library/Formula/scmpuff.rb
```diff
@@ -0,0 +1,29 @@
...
+    mkdir_p "#{gopath}/src/github.com/mroth"
+    ln_s buildpath, "#{gopath}/src/github.com/mroth/scmpuff"
+
+    # do a compile, manually specifying version so that scmpuff's build script
+    # does not do it's normal version detection which depends on being in a git
+    # repository -- instead let homebrew recipe specify the version number.
+    ENV["GOPATH"] = gopath
+    system "go", "build", "-o", "#{bin}/scmpuff", "-ldflags", "-X main.VERSION #{version}"
```
mikemcquaid: Any `godeps` requirements?

mroth: All external deps are vendored into source control directly (as seems to be the evolving "Go way" of doing this esp. once 1.5+ hits), so no need for godep resources in the formula.

mroth:
Pushed new squashed diff with second round of review changes.

Library/Formula/scmpuff.rb
mikemcquaid: Great. Glad to hear Go is moving away from this.

mroth:
By the way, when moving to the more standard Homebrew way of stubbing gopath, I noticed the pattern could probably be easily extracted into a method in the Homebrew DSL for ease of use, something like _(pseudo-code!)_:

```ruby
def stub_gopath(namespace)
  tmp_gopath = File.join buildpath, "src", namespace
  mkdir_p File.dirname(tmp_gopath)
  ln_s buildpath, tmp_gopath
end
```

And then just invoke `stub_gopath("github.com/mroth/scmpuff")` or something like that.

Although with dependencies being far from a resolved story in the Go community (sadly), you might want to hold off on that so you don't have to support it forever?

Library/Formula/scmpuff.rb
DomT4: Whoops. Yup, Mike is right :see_no_evil:

mikemcquaid:
Yeh, that's not a bad idea at all, thanks.

mikemcquaid:

PR #42373, opened 2015-08-03
capnp 0.5.3

iml:
The extra `ENV` is a guess at fixing this error in Mountain Lion https://github.com/Homebrew/homebrew-science/pull/2605:
```
checking capnp/common.h usability... no
checking capnp/common.h presence... no
checking for capnp/common.h... no
configure: error: Cap'n Proto headers not found.

### DIFF
--- a/Library/Formula/capnp.rb
+++ b/Library/Formula/capnp.rb
@@ -1,8 +1,8 @@
 class Capnp < Formula
   desc "Data interchange format and capability-based RPC system"
   homepage "https://capnproto.org/"
-  url "https://capnproto.org/capnproto-c++-0.5.2.tar.gz"
-  sha256 "23145a0127c2b1629c4b72e6000e04991640e551947fa39fd06710d64bd942a8"
+  url "https://capnproto.org/capnproto-c++-0.5.3.tar.gz"
+  sha256 "cdb17c792493bdcd4a24bcd196eb09f70ee64c83a3eccb0bc6534ff560536afb"
 
   bottle do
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Library/Formula/capnp.rb [hunk 1]
```diff
@@ -14,6 +14,7 @@ class Capnp < Formula
   option "without-shared", "Disable building shared library variant"
 
   def install
+    ENV.cxx11 if MacOS.version < :mavericks
```
mikemcquaid: I'd guess a `needs :cxx11` may be needed too. Also, that should maybe be optional (or at least recommended). Can you confirm this fixes the issue or is it a guess?

DomT4: We've already added `needs :cxx11` to the formula previously in https://github.com/Homebrew/homebrew/commit/15ca054a15eff279232672d3d2b59abbae148fb6 - Just missed the `ENV` apparently.

iml: C++11 support is obligatory for this package, so I think it's fine to be explicit here, but I'm not sure whether this will above fix the issue. The "missing" headers are present in the ML bottle, but perhaps they're missing the expected functions? 

In any case, I need to make some adjustments here as upstream has moved to cmake and the `without-shared` has apparently been removed.

mikemcquaid: If it's obligatory let's just add `ENV.cxx11` without the `if`

iml:
Switched to cmake; the `disable-shared` flag has been deprecated.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/eb2f61d4bd43fc83984be6b7213743654de07058. Thanks again @iml!

PR #42376, opened 2015-08-03
boost-python: fix strict audit

cosmo0920:


mikemcquaid:

For future reference we generally don't accept `audit`-only fixes but I'll accept this one for now. Cheers!

cosmo0920:
> For future reference we generally don't accept audit-only fixes but I'll accept this one for now.

Thanks! I understand it.

PR #40393, opened 2015-06-05
burp 2.0.20

bfontaine:

PR #42377, opened 2015-08-03
burp: fix URL, since sourceforge only hosts recent releases.

apjanke:
The SourceForge page for `burp` only hosts the most recent versions, and 1.4.28 has aged off of it. This repoints it at burp's own homepage.

### DIFF
--- a/Library/Formula/burp.rb
+++ b/Library/Formula/burp.rb
@@ -3,8 +3,8 @@ require "formula"
 class Burp < Formula
   desc "Network backup and restore"
   homepage "http://burp.grke.org/"
-  url "https://downloads.sourceforge.net/project/burp/burp-1.4.28/burp-1.4.28.tar.bz2"
-  sha1 "ddd5340db6aac1fa2e58c1e66c89730d8fd9bfb5"
+  url "http://burp.grke.org/downloads/burp-1.4.28/burp-1.4.28.tar.bz2"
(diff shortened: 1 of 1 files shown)


DomT4:
What's stopping us updating here?

apjanke:
Nothing AFAIK. I just don't know `burp`. There's a patch in the formula that'd need to be adapted.

DomT4:
Would you be willing to try an update? It doesn't look like a massive amount has changed upstream in terms of build.

bfontaine:
The new version might be broken, see #40393.

DomT4:
Ah, thanks @bfontaine, managed to miss your PR.

apjanke:
If you want, I could try an update to the latest 1.x version, 1.4.40, while we're waiting on an upstream fix for 2.0.20. That might avoid the breakage with the 2.x major version. I'll have time in the next couple days.

DomT4:
Will merge it for now so people who build from source aren't getting bitten by not being able to download it.

Just a nit, but commit message was a tad long here. Should be ~50 characters max for the main commit summary, and then a longer message underneath if necessary.

Pulled in https://github.com/Homebrew/homebrew/commit/5c3754a77c3aa196afa3548dd701582e039bee6b. Thanks @apjanke!

apjanke:
Sorry about that. I'll work on being more concise with the commits and PR titles.

PR #42397, opened 2015-08-03
add passlib to ansible formula; closes #42368

### LINKED ISSUE #42368: ansible formula needs passlib
As per the [Ansible documentation](http://docs.ansible.com/ansible/faq.html#how-do-i-generate-crypted-passwords-for-the-user-module) the `passlib` Python module is necessary to use crypted passwords on OS X.

There is no `passlib` module in Homebrew.  In addition, using the system `pip` to install passlib results in failure, because Ansible cannot find `passlib` (it isn't in its `PYTHONPATH`).

I'm not sure of the best course of action here.  It seems this may be installed with `brew-pip`, but whether or not the module would be visible to Ansible even then is anyone's guess.

boneskull:
I wasn't able to add a specific test for this, because the case I'm covering requires user input. 

A manual test against a playbook such as this works:

```yaml
---
- name: Test
  vars_prompt:
    - name: password
      prompt: Enter Password
      private: yes
      encrypt: sha512_crypt
      confirm: yes
      salt_size: 7
```

...whereas before this change, it failed.

### DIFF
--- a/Library/Formula/ansible.rb
+++ b/Library/Formula/ansible.rb
@@ -282,6 +282,11 @@ class Ansible < Formula
     sha256 "f43f9f15b0b719de94cab2754dcf78ef63b40ee2a12cea296e7af788b28501bb"
   end
 
+  resource "passlib" do
+    url "https://pypi.python.org/packages/source/p/passlib/passlib-1.6.4.tar.gz"
+    sha256 "d41bd7a2d22f9bd7e19ff4eed0eea2316eb737f3ec6a7c361dde6b2785b08cdc"
+  end
+
   def install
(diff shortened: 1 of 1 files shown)


mikemcquaid:

For future reference the preferred commit message format is `ansible: add passlib resource.`.

boneskull:
> Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

More maintainers should be so courteous when PRs are submitted.  homebrew sets a great example!

mikemcquaid:
@boneskull Thanks for the kind works :sparkles:

PR #42398, opened 2015-08-03
rubocop: update more rules

xu-cheng:
cc @mikemcquaid

### DIFF
--- a/Library/.rubocop.yml
+++ b/Library/.rubocop.yml
@@ -127,12 +127,16 @@ Style/SignalException:
 Style/TrailingComma:
   Enabled: false
 
-# We prefer compact if-else-end/case-when-end alignment
+# we prefer compact if-else-end/case-when-end alignment
 Lint/EndAlignment:
   AlignWith: variable
 Style/CaseIndentation:
   IndentWhenRelativeTo: end
 
-# We prefer Perl-style regex back references
+# we prefer Perl-style regex back references
 Style/PerlBackrefs:
   Enabled: false
+
+# this is required for Ruby 1.8
+Style/DotPosition:
+  EnforcedStyle: trailing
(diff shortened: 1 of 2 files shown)


xu-cheng:
Apply the the rule to all core files locally, the test passed with ruby 1.8.

xu-cheng:
I also noticed some weird indention for `module ... extend self`(e.g `Library/Homebrew/hardware.rb`) which probably need a manual fix. Rubocop suggests turn `extend self` to `module_function` but doesn't do any auto correction.

Library/.rubocop.yml
```diff
@@ -1,3 +1,8 @@
+AllCops:
+  Exclude:
+    - 'Homebrew/vendor/**/*'
+    - 'Homebrew/test/vendor/**/*'
```
bfontaine: Would `Homebrew/**/vendor/**/*` work for both?

Library/.rubocop.yml
```diff
@@ -69,10 +74,6 @@ Style/PercentLiteralDelimiters:
         '%W': '[]'
         '%x': '()'
 
-# comma on last element is editor friendly
-Style/TrailingComma:
-    EnforcedStyleForMultiline: comma
```
mikemcquaid: Does this fail on 1.8? I do always prefer this format.

xu-cheng: I prefer this as well. However it will append comma for method calling, e.g. `foo(a, b, )`, which is syntax error.

Library/.rubocop.yml
xu-cheng: I think explicitly list them would be better.

Library/.rubocop.yml
mikemcquaid: Syntax error just in 1.8, yeh?

xu-cheng: Yes, just in 1.8

mikemcquaid:
:+1: :shipit:

PR #42401, opened 2015-08-03
daq 2.0.6

apjanke:
daq: upgrade to 2.0.6. Also fixes a 404 for the 2.0.5 download. The Snort site doesn't seem to host older versions.


mikemcquaid:
Thanks again @apjanke!

For future reference the preferred commit message format for simple version updates is `daq 2.0.6`.

apjanke:
Will do.

PR #42403, opened 2015-08-03
audit: simplify command-line spelling check

afh:


Library/Homebrew/cmd/audit.rb
```diff
@@ -360,12 +360,8 @@ def audit_desc
...
     end
 
-    if desc =~ %r[[Cc]ommandline]
-      problem "It should be \"command-line\", not \"commandline\"."
-    end
-
-    if desc =~ %r[[Cc]ommand line]
-      problem "It should be \"command-line\", not \"command line\"."
+    if desc =~ %r[([Cc]ommand ?line)]
+      problem "It should be \"command-line\", not \"#{$1}\""
```
xu-cheng: I know it's the existing logic to be blamed. But maybe `"It should be \"command-line\", not \"#{$1}\" in description"` is better, so we know such problem comes from description.

afh:
How about `"Description should use \"#{$2}ommand-line instead of \"#{$1}\""`?
`$2` would be a new regex group within the first matching `[Cc]`.

xu-cheng:
Should be `"Description should use \"command-line\" instead of \"#{$1}\""`.

bfontaine:
Please rebase this on top of master.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/9e679750e54516576ea4e1e6823c1e2a43ceca58. Thanks @afh!

PR #42404, opened 2015-08-03
audit: add check for indefinite article

afh:
at the beginning of desc.

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -367,6 +367,10 @@ class FormulaAuditor
     if desc =~ %r[[Cc]ommand line]
       problem "It should be \"command-line\", not \"command line\"."
     end
+
+    if desc =~ %r[^([Aa]n?)\s]
+      problem "Please remove the indefinite article \"#{$1}\" from the beginning of the description"
+    end
   end
 
   def audit_homepage


Library/Homebrew/cmd/audit.rb
```diff
@@ -367,6 +367,10 @@ def audit_desc
     if desc =~ %r[[Cc]ommand line]
       problem "It should be \"command-line\", not \"command line\"."
     end
+
+    if desc =~ %r[^([Aa]n?)\s]
+      problem "Description begins with indefinite article: \"#{$1}\""
```
xu-cheng: Might be better to offer the correction suggestion, such as `Please remove the indefinite article (\"#{$1}\") in the beginning of description`.

mikemcquaid: Might want to clarify `(i.e. 'a' or 'an')` just for people who don't know what an indefinite article is.

afh: @xu-cheng I was uncertain which wording to use and will happily change it.
 
@mikemcquaid That's what the `#{$1}` is for, so people know which indefinite article was used in the description.

mikemcquaid: Ah, I see.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/4ce9763593aa310cc21a7c610e56aaf376b0679f. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

Issue #41759, opened 2015-07-15
Recursive dependencies of :build dependencies are not added to paths


tdsmith:
If a formula `depends_on "cairo" => :build`, only `/usr/local/opt/cairo/lib/pkgconfig` is added to PKG_CONFIG_PATH and fontconfig, pixman, etc are not. They are present when cairo is a "normal" dependency (i.e. `depends_on "cairo"`).

The same is true for $PATH.

This was surfaced by #41707.

jacknagel:
fwiw - this was done intentionally in fdfc54961eb497b6dd854852dc604cb50c994fda. Not sure what that means here but hopefully that provides some context.

jacknagel:
By the way, if you have questions about implementation stuff, feel free to ping me. A lot of this stuff is bus factor 1 so I'm happy to answer questions.

tdsmith:
Thank you; I appreciate it! I'll try not to abuse your patience.

I agree that it should not be necessary to link against dependencies of build dependencies, and filtering them from the build environment makes sense. It seems like a formula's build dependency may occasionally have to _invoke_ its own dependencies but empirically that seems rare, and manually adding them to PATH in those situations is probably a better answer than making a global change.

Closing as not a bug.

mikemcquaid:
> By the way, if you have questions about implementation stuff, feel free to ping me. A lot of this stuff is bus factor 1 so I'm happy to answer questions.

@jacknagel Really appreciate your attitude and help here, thanks dude :heart:

PR #42405, opened 2015-08-03
0install 2.8

afb:
Another attempt, after the previous formula had several issues. Hopefully this is better...

### DIFF
--- a/Library/Formula/zero-install.rb
+++ b/Library/Formula/zero-install.rb
@@ -1,0 +1,135 @@
+class ZeroInstall < Formula
+  desc "Zero Install is a decentralised software installation system"
+  homepage "http://0install.net/"
+  url "https://downloads.sf.net/project/zero-install/0install/2.8/0install-2.8.tar.bz2"
+  sha256 "12de771be748bce9350c90bc4720029a566b078ceabd335af09386ac6a37df2b"
+
+  depends_on "pkg-config" => :build
+  depends_on "ocaml" => :build
+  depends_on "opam" => :build
+  depends_on "gtk+" => :optional
+  depends_on :gpg
+
+  # 0install.2.8 resources from https://github.com/ocaml/opam-repository.git
+
+  resource "easy-format" do # [required by yojson]
+    url "https://opam.ocaml.org/archives/easy-format.1.0.2+opam.tar.gz"
+    version "1.0.2"
+    sha256 "c2a04ab1084bc5ce1ec52a3aa320c825c22f8be527fb5f82d920a9d92a673cd3"
+  end
+
+  resource "biniou" do # [required by yojson]
+    url "https://opam.ocaml.org/archives/biniou.1.0.9+opam.tar.gz"
+    version "1.0.9"
+    sha256 "08e6a17a19fbe5e9da4c77d1e8c1bb05333f84ec8a634721a39c6c505ce51e0d"
+  end
+
+  resource "cppo" do # [required by yojson]
+    url "https://opam.ocaml.org/archives/cppo.1.1.2+opam.tar.gz"
+    version "1.1.2"
+    sha256 "35c1caddb249293c5a24f095e0dea85c670056083e30010078124b65e557a2d8"
+  end
+
+  resource "yojson" do
+    url "https://opam.ocaml.org/archives/yojson.1.1.8+opam.tar.gz"
+    version "1.1.8"
+    sha256 "20ce2c2f752b49695468b0bf66f17cf20219d7bca5a0b71a3fb89af500fb52f1"
+  end
+
+  resource "xmlm" do
+    url "https://opam.ocaml.org/archives/xmlm.1.2.0+opam.tar.gz"
+    version "1.2.0"
+    sha256 "518617e737730c89a04ca68ae927cb1510a027a9924df591df8575e386d810f0"
+  end
+
+  resource "ounit" do
+    url "https://opam.ocaml.org/archives/ounit.2.0.0+opam.tar.gz"
+    version "2.0.0"
+    sha256 "bc87d648a6186d7a708442a9dcb2197e9f0d5d8ba62f60daf18ae8b592b658ce"
+  end
+
+  resource "react" do
+    url "https://opam.ocaml.org/archives/react.1.2.0+opam.tar.gz"
+    version "1.2.0"
+    sha256 "690a1a7b18be21fa83876599639ddd3954c2a53d5cf1e13a76f2a140e1f9aa71"
+  end
+
+  resource "ppx_tools" do # [required by lwt]
+    url "https://opam.ocaml.org/archives/ppx_tools.0.99.2+opam.tar.gz"
+    version "0.99.2"
+    sha256 "0c5b9802de2005b55717ac78bffcead1ee11dd28f91fb32f7e3518a7d7d8c48a"
+  end
+
+  resource "lwt" do
+    url "https://opam.ocaml.org/archives/lwt.2.4.6+opam.tar.gz"
+    version "2.4.6"
+    sha256 "297eda984412d6cd231bea548da4ef2d8b08db5f1a0bc295d47ac993243005b5"
+  end
+
+  resource "extlib" do
+    url "https://opam.ocaml.org/archives/extlib.1.6.1+opam.tar.gz"
+    version "1.6.1"
+    sha256 "c76176916c39d4ccae82a34c33e694652c0c55d79eec8830dfddadb580b53773"
+  end
+
+  resource "ocurl" do
+    url "https://opam.ocaml.org/archives/ocurl.0.7.2+opam.tar.gz"
+    version "0.7.2"
+    sha256 "669c5142b7f4002521468b75ee254b269f0094c9a22d3381e94a440a6aa2d400"
+  end
+
+  if build.with? "gtk+"
+    resource "lablgtk" do
+      url "https://opam.ocaml.org/archives/lablgtk.2.18.3+opam.tar.gz"
+      version "2.18.3"
+      sha256 "f0b7ed0bd85f6cf4b4c5f81966f03763e76bb9f866f5172511ce48cf31fd433c"
+    end
+  end
+
+  resource "sha" do
+    url "https://opam.ocaml.org/archives/sha.1.9+opam.tar.gz"
+    version "1.9"
+    sha256 "d3cfda4cd6f79b01c6613219baa3e8548365c309952168b3539e0edce9370b40"
+  end
+
+  def install
+    opamroot = buildpath/"opamroot"
+    ENV["OPAMROOT"] = opamroot
+    ENV["OPAMYES"] = "1"
+    system "opam", "init", "--no-setup"
+    archives = opamroot/"repo/default/archives"
+    modules = []
+    resources.each do |r|
+      r.verify_download_integrity(r.fetch)
+      original_name = File.basename(r.url)
+      cp r.cached_download, archives/original_name
+      modules << "#{r.name}=#{r.version}"
+    end
+    system "opam", "install", *modules
+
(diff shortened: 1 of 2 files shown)


Library/Formula/zero-install.rb
```diff
@@ -0,0 +1,152 @@
...
+  head do
+    url "https://github.com/0install/0install.git"
+    depends_on "gettext" => :build
+  end
+
+  option 'without-gui', "Skip the 0install gui (requires GTK+)"
+
+  depends_on :gpg
+  depends_on "curl"
+  depends_on "gtk+" if build.with? "gui"
```
mikemcquaid: Use `=> :optional` here instead, thanks.

Library/Formula/zero-install.rb
```diff
@@ -0,0 +1,152 @@
...
+  depends_on "pkg-config" => :build
+  depends_on "objective-caml" => :build
+  depends_on "opam" => :build
+
+  if build.head?
+    # opam install yojson xmlm ounit react lwt extlib ocurl lablgtk sha
```
mikemcquaid: Remove this comment and use a `head do` block.

mikemcquaid: Oh, also, these should just be resources instead or remove the `head` block.

mikemcquaid:
Maybe worth just having the non-head build in the formula first? Looks like the head build is quite a bit more complicated?

afb:
Actually it is the non-head build that is complicated, due to the want of using resources...
The current build is as simple as (or it would be, if wasn't for the keg-only dependencies):
```
$ brew install pkg-config gettext gnupg gtk+ objective-caml opam
$ opam install yojson xmlm ounit react lwt extlib ocurl sha lablgtk
$ make
$ sudo make install
```

But that doesn't yield us any bottles, which means that users need to install ocaml/opam.
So as per the cookbook, the ocaml modules are locally vendored instead of dependencies.

Resources need to be versioned, however, so then it doesn't work with a HEAD build ?
But I can remove the head sections, and convert the "gui" flag over to :recommended...

mikemcquaid:
Resources could potentially be unversioned Git repositories if they are only in the `head do` block.

Library/Formula/zero-install.rb
afb: How do I translate the :optional/:recommend over to the build option of using lablgtk or not ?
Should the GUI just be made mandatory, so that there is no way to install CLI only (skip gtk+)

mikemcquaid: The option will be named `without-gtk+` instead. It should also probably default to off.

afb:
@mikemcquaid: the resources are all in the same repository, but I suppose that *could* be done too...

* https://github.com/ocaml/opam-repository

Library/Formula/zero-install.rb
afb: If I default it to off, it won't be in the bottle ? That was the main reason for leaving it on.
In other distributions, the GUI goes in a separate package: "0install" vs. "0install-core"

mikemcquaid: Yeh, we default to more minimal versions.

afb: That is unfortunate (no gui bottle), since gtk+ is one of the more annoying ones.
But at least you can install *that* + deps (and ocaml/opam + boost) from bottles...

Library/Formula/zero-install.rb
```diff
@@ -0,0 +1,131 @@
...
+    url "https://opam.ocaml.org/archives/ocurl.0.7.2+opam.tar.gz"
+    version "0.7.2"
+    sha256 "669c5142b7f4002521468b75ee254b269f0094c9a22d3381e94a440a6aa2d400"
+  end
+
+  resource "lablgtk" do
+    url "https://opam.ocaml.org/archives/lablgtk.2.18.3+opam.tar.gz"
+    version "2.18.3"
+    sha256 "f0b7ed0bd85f6cf4b4c5f81966f03763e76bb9f866f5172511ce48cf31fd433c"
+  end if build.with? "gtk+"
```
DomT4: This should be:

```ruby
if build.with? "gtk"
  resource "lablgtk" do
    url "https://opam.ocaml.org/archives/lablgtk.2.18.3+opam.tar.gz"
    version "2.18.3"
    sha256 "f0b7ed0bd85f6cf4b4c5f81966f03763e76bb9f866f5172511ce48cf31fd433c"
  end
end
```

The `end if` lines are confusing and we've been trying to wipe them out for a while.

Library/Formula/zero-install.rb
```diff
@@ -0,0 +1,131 @@
...
+    url "https://opam.ocaml.org/archives/sha.1.9+opam.tar.gz"
+    version "1.9"
+    sha256 "d3cfda4cd6f79b01c6613219baa3e8548365c309952168b3539e0edce9370b40"
+  end
+
+  def install
+    # Parallel builds fail for some of these opam libs.
+    ENV.deparallelize
+
+    opamroot=buildpath/"opamroot"
```
DomT4: Should be spacing here. `opamroot = buildpath/"opamroot"`

Library/Formula/zero-install.rb
```diff
@@ -0,0 +1,131 @@
...
+    ENV["OPAMROOT"] = opamroot
+    ENV["OPAMYES"] = "1"
+    mkdir_p opamroot/"archives"
```
DomT4: `(opamroot/"archives").mkpath` looks a little nicer.

Library/Formula/zero-install.rb
```diff
@@ -0,0 +1,131 @@
...
+    system "opam", "init", "--no-setup"
+    modules = []
+    resources.each do |r|
+      r.verify_download_integrity(r.fetch)
```
DomT4: Is this doing checksums? Why wouldn't that checksum happen when we fetch the resources?

afb: TBH it was just copied from the opam formula, seems that is indeed doing checksumming twice

afb:
Hope this isn't another one of those nit-pickings where just it ends up lying around in the end, like #28771

DomT4:
Referencing a PR that was some 15000 Issues/PRs back isn't terribly helpful. Maintainers want things to be as expected before going in otherwise at some point we end up having to fix them. If it meets maintainer's concerns & our minimum addition requirements, it will be merged.

As far as I can see, there's not much left to deal with before this is fine.

Library/Formula/zero-install.rb
```diff
@@ -0,0 +1,132 @@
+class ZeroInstall < Formula
+  desc "Zero Install is a decentralised software installation system"
+  homepage "http://0install.net/"
+  url "https://downloads.sf.net/project/zero-install/0install/2.8/0install-2.8.tar.bz2"
+  sha256 "12de771be748bce9350c90bc4720029a566b078ceabd335af09386ac6a37df2b"
+
+  depends_on :gpg
+  depends_on "curl"
```
DomT4: Is it incapable of using the system `curl`?

Library/Formula/zero-install.rb
```diff
@@ -0,0 +1,132 @@
...
+  depends_on "gtk+" => :optional
+  depends_on "pkg-config" => :build
+  depends_on "ocaml" => :build
+  depends_on "opam" => :build
```
DomT4: It's not really that important, but for future referencing sake, we prefer something like:

```ruby
depends_on "" => :build
depends_on "" => :build
depends_on ""
depends_on "" => :optional
depends_on :special_dep
```

Build deps, then mandatory deps, then optional deps, then special deps.

Library/Formula/zero-install.rb
```diff
@@ -0,0 +1,132 @@
...
+        <summary>minimal demonstration program</summary>
+
+        <implementation id="." version="0.1-pre">
+          <command name='run' path='hello.py'>
+            <runner interface='http://repo.roscidus.com/python/python'></runner>
+          </command>
+        </implementation>
+      </interface>
+    EOS
+    assert_equal "hello world\n", `#{bin}/0launch --console hello.xml`
```
DomT4: Can this use `shell_output("#{bin}/0launch --console hello.xml")` instead of backtics?

afb:
@DomT4: ok, thanks.

seems like a missed one new dep-of-a-dep, and put the archives in the wrong place...
so pushed a new commit with that fixed. was trying to get gtk+ to build just the optional plugin,
but it seems that it doesn't work as a "build" dependency (doesn't pull in the deps like pango)
so the bottle will default to console-only, and needs a formula rebuild for the gui to work.

Library/Formula/zero-install.rb
afb: Most likely. I didn't write it. Is it better/safer with that function than backticks ?

Library/Formula/zero-install.rb
afb: Hmm, don't think neither "audit" or "style" said anything about this...

Library/Formula/zero-install.rb
afb: Nope, it should be "new enough". Dropping the curl then...

Library/Formula/zero-install.rb
DomT4: It won't. We don't codify it, yet.

Library/Formula/zero-install.rb
DomT4: We prefer it over backtics inside the `test do` blocks, aye.

afb:
@DomT4: seems like it doesn't download the resources on just brew install (not sure when, "fetch")
so if you change a dependency it looks like it can go missing, just got a file-not-found at least.
trying to upgrade the formula to the later version, but also want to upgrade the ocaml modules...
and if there was a way to get the "gui_gtk.cmxs" (plugin bundle) to build, that would be great too.

afb:
i.e. the :build doesn't work, because it only adds "gtk+" to the path but not "pango" and "glib" etc etc.
tried with `ENV.append_path "PKG_CONFIG_PATH", "#{lib}/pkgconfig"`, but that hack didn't work either.

afb:
ah, normally the `stage` is doing the fetch and verify. so need to add that line back to the formula.
(these resources are unpacked by opam, better let it handle it and get the suffix it prefers etc etc)

Library/Formula/zero-install.rb
afb: Looks like it was copied from Resource.stage:
```ruby
  def stage(target = nil, &block)
    unless target || block
      raise ArgumentError, "target directory or block is required"
    end

    verify_download_integrity(fetch)
    unpack(target, &block)
  end
```

DomT4:
Funky. As long as it works, I'm happy to revert to the `opam` way mimicked previously.

afb:
It was pointed out to me that build deps having broken pkg-config is a feature: #41759
So that probably means that we cannot build the gui in the bottle, but only conditionally

DomT4:
As long as it works optionally that's fine. We very rarely build GUIs by default anyhow.

Library/Formula/zero-install.rb
```diff
@@ -0,0 +1,138 @@
...
+
+  resource "sha" do
+    url "https://opam.ocaml.org/archives/sha.1.9+opam.tar.gz"
+    version "1.9"
+    sha256 "d3cfda4cd6f79b01c6613219baa3e8548365c309952168b3539e0edce9370b40"
+  end
+
+  def install
+    # Parallel builds fail for some of these opam libs.
+    ENV.deparallelize
```
mikemcquaid: Can you report issues upstream for these? Thanks!

afb: Will do (as soon as I remember _which_ of the modules it was that failed)

talex5:
What else needs to happen for this to get merged?

mikemcquaid:

PR #42407, opened 2015-08-03
Formula files style updates.

mikemcquaid:
Mass update all the formula files with `brew fix` (which uses `rubocop` and also updates to e.g. `sha256`). Core files were done in https://github.com/Homebrew/homebrew/pull/42354. Committing as @BrewTestBot to make it a bit more obvious that it was an automatic change. Please review ASAP, thanks! Will immediately cancel the build.


mikemcquaid:
@mistydemeo @Homebrew/linux heads up on this one too.

PR #42410, opened 2015-08-03
doxygen 1.8.10

afh:
Doxygen switched from `configure` to `cmake`.

Instead of manually installing the files (`bin.install Dir["build/bin/*"`) I would have like to use
`make install`, unfortunately that fails, since documentation files are missing.

The missing documentation files are created with `make docs`, unfortunately that fails, since `epstopdf` is used when building the doxygen reference manual and I found no way on how to customise this using CMake definitions.

I find the proposed solution more straightforward than patching the appropriate `CMakeLists.txt` files.
If I find some spare time I'll report this upstream and may propose even some changes.

### DIFF
--- a/Library/Formula/doxygen.rb
+++ b/Library/Formula/doxygen.rb
@@ -1,10 +1,10 @@
 class Doxygen < Formula
   desc "Generate documentation for several programming languages"
   homepage "http://www.doxygen.org/"
+  url "http://ftp.stack.nl/pub/users/dimitri/doxygen-1.8.10.src.tar.gz"
+  mirror "https://downloads.sourceforge.net/project/doxygen/rel-1.8.10/doxygen-1.8.10.src.tar.gz"
+  sha256 "cedf78f6d213226464784ecb999b54515c97eab8a2f9b82514292f837cf88b93"
   head "https://github.com/doxygen/doxygen.git"
-  url "http://ftp.stack.nl/pub/users/dimitri/doxygen-1.8.9.1.src.tar.gz"
-  mirror "https://downloads.sourceforge.net/project/doxygen/rel-1.8.9.1/doxygen-1.8.9.1.src.tar.gz"
-  sha256 "d4ab6e28d4d45d8956cad17470aade3fbe2356e8f64b92167e738c1887feccec"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> since epstopdf is used

Could we make that a build-time dependency? The bottle is `cellar :any` so everyone can use it anyway. Also, you'll need to rebase this on `master` I'm afraid as I just pushed a change to this formula.

afh:
@mikemcquaid Sure, `epstopdf` could be made a build-time dependency, yet I find requiring a TeX distribution a bit too much for this. Especially since there are people (like me :wink:) that install TeX from elsewhere ([MacTeX](http://tug.org/mactex/) in my case).

Which build-time dependency would that be anyways? I seem to be unable to find a suitable one in homebrew...

mikemcquaid:
I've no idea I'm afraid. I agree about not requiring a TeX distribution but I wonder if it's installable standalone (even if that's part of the `doxygen` formula)

afh:
`[epstopdf](http://ftp.fau.de/ctan/support/epstopdf/epstopdf.pl)` seems to be a Perl script that utilises ghostscript. Still too much from my point of view.

Also note that `make docs` would also install the reference manual for doxygen, which isn't too useful for the typical doxygen user.

mikemcquaid:
Ok, that's fine as it is, thanks.

Library/Formula/doxygen.rb
```diff
@@ -19,56 +19,25 @@ class Doxygen < Formula
...
-    # one Makefile doesn't get generated with these, so pull
-    # them out of a known good file and cram them into the other.
-    lex = ""
-    yacc = ""
-
-    inreplace "src/libdoxycfg.t" do |s|
-      lex = s.get_make_var "LEX"
-      yacc = s.get_make_var "YACC"
+    args = std_cmake_args + %W[
+      -DCMAKE_INSTALL_PREFIX=#{prefix}
```
mikemcquaid: This is already set by `std_cmake_args`, I believe?

DomT4: Yup:

```
      -DCMAKE_C_FLAGS_RELEASE=
      -DCMAKE_CXX_FLAGS_RELEASE=
      -DCMAKE_INSTALL_PREFIX=#{prefix}
      -DCMAKE_BUILD_TYPE=Release
      -DCMAKE_FIND_FRAMEWORK=LAST
      -DCMAKE_VERBOSE_MAKEFILE=ON
      -Wno-dev
```

afh: @mikemcquaid Good catch
@DomT4 Thanks for the info.

I've updated the PR accordingly and also remove the `-Dbuild_doc=ON` option, since that isn't necessary with the manual installation in this formula (it was a left-over from previous experiments).

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/4650b30a2a7d43a179a20324112735ee29c7e9a3. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #42415, opened 2015-08-03
falcon: remove duplicate CMAKE_INSTALL_PREFIX

afh:
which is already set by `std_cmake_args`

### DIFF
--- a/Library/Formula/falcon.rb
+++ b/Library/Formula/falcon.rb
@@ -26,20 +26,19 @@ class Falcon < Formula
 
   def install
     args = std_cmake_args + %W[
-      -DCMAKE_INSTALL_PREFIX=#{prefix}
       -DFALCON_BIN_DIR=#{bin}
       -DFALCON_LIB_DIR=#{lib}
       -DFALCON_MAN_DIR=#{man1}
       -DFALCON_WITH_INTERNAL_PCRE=OFF
       -DFALCON_WITH_MANPAGES=ON]
 
-    if build.include? "editline"
+    if build.with? "editline"
       args << "-DFALCON_WITH_EDITLINE=ON"
     else
       args << "-DFALCON_WITH_EDITLINE=OFF"
     end
 
-    if build.include? "feathers"
+    if build.with? "feathers"
       args << "-DFALCON_WITH_FEATHERS=feathers"
     else
       args << "-DFALCON_WITH_FEATHERS=NO"


DomT4:
This isn't your fault, but when the options were deprecated someone missed the options below:

```
  option "with-editline", "Use editline instead of readline"
  option "with-feathers", "Include feathers (extra libraries)"

  deprecated_option "editline" => "with-editline"
  deprecated_option "feathers" => "with-feathers"
```

```
    if build.include? "editline"
      args << "-DFALCON_WITH_EDITLINE=ON"
    else
      args << "-DFALCON_WITH_EDITLINE=OFF"
    end

    if build.include? "feathers"
      args << "-DFALCON_WITH_FEATHERS=feathers"
    else
      args << "-DFALCON_WITH_FEATHERS=NO"
    end
```

Both those `build.include?` should be `build.with?`

afh:
There are some more formulae where this might be the case:

```
% grep 'build\.include?' $(grep -l deprecated_option Library/Formula/*.rb)
Library/Formula/falcon.rb:    if build.include? "editline"
Library/Formula/falcon.rb:    if build.include? "feathers"
Library/Formula/fortune.rb:      s.change_make_var! "OFFENSIVE", "0" if build.include? "no-offensive"
Library/Formula/groonga.rb:    args << "--enable-benchmark" if build.include? "enable-benchmark"
Library/Formula/minbif.rb:  patch :DATA if build.include? "pam"
Library/Formula/minbif.rb:    args << "ENABLE_IMLIB=" + ((build.include? "imlib2") ? "ON" : "OFF")
Library/Formula/minbif.rb:    args << "ENABLE_CACA=" + ((build.include? "libcaca") ? "ON" : "OFF")
Library/Formula/minbif.rb:    args << "ENABLE_PAM=" + ((build.include? "pam") ? "ON" : "OFF")
Library/Formula/python.rb:    system "make", "quicktest" if build.include? "quicktest"
Library/Formula/python3.rb:    system "make", "quicktest" if build.include? "quicktest"
Library/Formula/subversion.rb:    if build.include? "unicode-path"
Library/Formula/udunits.rb:    targets << "install-html" if build.include? "html-docs"
Library/Formula/udunits.rb:    targets << "install-pdf" if build.include? "pdf-docs"
Library/Formula/vrpn.rb:    if build.include? "clients"
```

mikemcquaid:
@afh If you felt generous submitting PRs for those would be great. If not, thanks for the `grep` anyway!

DomT4:
The `subversion` one is intended AFAIK. It's an old option that we no longer support, I believe. It may need removing at this point; I don't know if the warning is getting activated if we've scrapped the option completely.

afh:
@mikemcquaid Sure. I've been told to be cautious with PR that change more than 5 formulae for that might mess with the bot. Would you like one big PR? Single PRs?

@DomT4 I'll make sure to review formulae to check

mikemcquaid:
@afh Single PRs would be ideal but in this case these formulae look fine for a single big PR :+1:

mikemcquaid:
(or just add them to this PR)

afh:
BTW: There are many (~80) formulae that still use `build.include?`. Is that option completely deprecated or is it still being used in certain cases?

DomT4:
It's deprecated & we're killing the option off slowly as we handle version updates/etc.

afh:
If that is the case I'd like to propose the following shell snippet to change it all in one go:

```
for formula in $(grep -l 'build\.include?' Library/Formula/*.rb); do
    echo $formula
    vim -c '%s/build\.include?/build.with?/|normal ZZ' $formula
done
```

What do you think?

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/46034c2274af01a393b8c3c83f5a25d959c39fc1. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

The world's smallest nit here, but commit message would have worked as something like `falcon: remove duplicate argument` or such. We try to keep the commit messages lowercase.

> What do you think?

If you want to tee up 5-10 per PR per day we'd be happy to take a look at them. Likelihood is some of the bottles need regenerating to handle a few edge cases, as was the case a couple of weeks back when we encountered the same issue.

PR #42421, opened 2015-08-03
shell completion of official taps added to 'brew tap'

bfontaine:
cc @xu-cheng for the zsh part; I wasn’t able to test it.

I copied the file as `/usr/local/share/zsh/functions/_brew`, made sure it was in `$fpath`, but zsh doesn’t complete the commands when I type `brew <tab>`.

It works as expected in Bash.

### DIFF
--- a/Library/Contributions/brew_bash_completion.sh
+++ b/Library/Contributions/brew_bash_completion.sh
@@ -140,6 +140,7 @@ _brew_complete_tap ()
         return
         ;;
     esac
+    __brewcomp "$(brew tap --list-official)"
 }
 
 _brew_bottle ()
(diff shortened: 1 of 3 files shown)


mikemcquaid:
:+1:

bfontaine:
@mikemcquaid Does your comment mean you successfully tested the ZSH part?

mikemcquaid:
@bfontaine Nope, just that the diff looks good.

bfontaine:
Thanks, I [fixed](http://stackoverflow.com/a/26479426/735926) my `~/.zshrc` and was able to use the completion feature.

PR #42429, opened 2015-08-03
add libxml2 compile option to emacs formula

danielschranz:
Does what it says on the tin.

I created the same PR for linuxbrew: https://github.com/Homebrew/linuxbrew/pull/501

This is the first time I touched a formula so I hope I got it right - could only test it with linuxbrew though and worked fine for me there.

### DIFF
--- a/Library/Formula/emacs.rb
+++ b/Library/Formula/emacs.rb
@@ -40,7 +40,6 @@ class Emacs < Formula
   depends_on "imagemagick" => :optional
   depends_on "mailutils" => :optional
   depends_on "glib" => :optional
-  depends_on "libxml2" => :optional
 
   # https://github.com/Homebrew/homebrew/issues/37803
   if build.with? "x11"
--- a/Library/Formula/emacs.rb
+++ b/Library/Formula/emacs.rb
@@ -26,7 +26,7 @@ class Emacs < Formula
 
   option "with-cocoa", "Build a Cocoa version of emacs"
   option "with-ctags", "Don't remove the ctags executable that emacs provides"
-  option "with-libxml2", "Build with libxml2 support"
+  option "without-libxml2", "Don't build with libxml2 support"
 
   deprecated_option "cocoa" => "with-cocoa"
   deprecated_option "keep-ctags" => "with-ctags"
--- a/Library/Formula/emacs.rb
+++ b/Library/Formula/emacs.rb
@@ -26,6 +26,7 @@ class Emacs < Formula
 
   option "with-cocoa", "Build a Cocoa version of emacs"
   option "with-ctags", "Don't remove the ctags executable that emacs provides"
+  option "with-libxml2", "Build with libxml2 support"
 
   deprecated_option "cocoa" => "with-cocoa"
   deprecated_option "keep-ctags" => "with-ctags"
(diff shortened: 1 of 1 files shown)


dunn:
That test failure was my fault, should be fixed by #42436.

Library/Formula/emacs.rb
```diff
@@ -39,6 +40,7 @@ class Emacs < Formula
   depends_on "imagemagick" => :optional
   depends_on "mailutils" => :optional
   depends_on "glib" => :optional
+  depends_on "libxml2" => :optional
```
mikemcquaid: Can this not use the system `libxml2`?

danielschranz: It can and works fine, just wondering does that mean I remove this line completely what if the user doesn't have `libxml2` installed (I know that's unlikely, but I just wanted to make sure the build doesn't break for such a reason - or is this just expected behaviour in that case?)

mikemcquaid: If they are on OS X: they always have it installed.

danielschranz: and what about linuxbrew? (that's where I'm coming from :smile:)

mikemcquaid: We don't support Linuxbrew here; Linux formulae changes should be made in Linuxbrew, sorry.

danielschranz: No worries - I was doing this PR for linuxbrew before and they told me if I add this change to homebrew it gets automatically merged every week so the PR in linuxbrew has been closed again without being merged.

Following your comment I'm going to remove this line as it's not necessary for OS X...

mikemcquaid: Thanks. Can you link me to the PR in Linuxbrew?

danielschranz: Here it is: https://github.com/Homebrew/linuxbrew/pull/501

Library/Formula/emacs.rb
```diff
@@ -26,6 +26,7 @@ class Emacs < Formula
 
   option "with-cocoa", "Build a Cocoa version of emacs"
   option "with-ctags", "Don't remove the ctags executable that emacs provides"
+  option "with-libxml2", "Build with libxml2 support"
```
mikemcquaid: Sorry, can you enable this option by default (as it's on OS X by default)

sjackman:
Hi, Mike. Feel free to take this patch or not as you wish. My suggested patch for Linuxbrew is:
```ruby
depends_on "libxml2" unless OS.mac?
```

Library/Formula/emacs.rb
DomT4: Just involves changing the option to `without-libxml2` rather than `with-libxml2` if that's unclear at all.

danielschranz: Thanks for your hint, I tried to do that but when I do a `brew info emacs` it doesn't show the changed option. when I do the same thing with another package (`lua51`) I can change the option quite easily - probably doing something silly that I can't see at the moment, but I'll get this done. :smile:

danielschranz:
If people agree on @sjackman's suggestion I'll put that in as well

mikemcquaid:
@danielschranz @sjackman's suggestion should be a separate PR to Linuxbrew after this is merged, thanks.

mikemcquaid:

For future reference the preferred commit message format is `emacs: add libxml2 option.`.

PR #42431, opened 2015-08-03
mercurial: vendor pyopenssl

### LINKED ISSUE #42416: Mercurial lists /usr/bin/python even if Homebrew-installed /usr/local/bin/python is available
If you `brew install mercurial`, the shebang from /usr/local/bin/hg is `#!/usr/bin/python` instead of `#!/usr/local/bin/python`. even if you have brew installed python. This is wrong.

https://stackoverflow.com/questions/28512350/homebrew-path-env-var-is-modified-before-installation-of-formulae-mercurial-wi has more details. I suspect this is a common problem with Python Homebrew packages. But I'm not sure exactly what. I'm not ready to just to the SO conclusion that this is strictly the result of using distutils. Then again, Python packaging is wonky.

I'm a Mercurial contributor. If this is an upstream bug, I'll land a fix upstream once the cause is identified.

DomT4:
Would close #42416, maybe.

### DIFF
--- a/Library/Formula/mercurial.rb
+++ b/Library/Formula/mercurial.rb
@@ -13,10 +13,68 @@ class Mercurial < Formula
     sha256 "bab3981ab5525491cde27a0df3686d303a2e80b4f9796bbdff6681b9d4b8637f" => :mountain_lion
   end
 
+  depends_on "openssl"
+
+  resource "six" do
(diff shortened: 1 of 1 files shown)


tdsmith:
lgtm, thanks!

PR #42439, opened 2015-08-04
Adding sourceforge mirror

terah:


dunn:
Can you squash this down to a single commit? Thanks!

PR #42442, opened 2015-08-04
couchpotato 2.6.3

stuwil:


Library/Formula/couchpotatoserver.rb
```diff
@@ -54,4 +54,8 @@ def startup_script; <<-EOS.undent
   def caveats
     "CouchPotatoServer defaults to port 5050."
   end
+
+  test do
+    system "#{bin}/couchpotatoserver", "-h"
```
dunn: It's better than nothing, but it is possible to write a test that checks some more of the functionality of the program?

stuwil:
See #42466

PR #42463, opened 2015-08-04
jenkins 1.623

kassi:


dunn:
Is it possible to add a test in order to check basic functionality?

DomT4:
The only thing that'd really be feasible is:

```ruby
test do
  assert_match /#{version}/, shell_output("#{bin}/jenkins --version")
end
```

But if the machine hadn't yet had Java invoked that day, it can take a little while to get the output. Anything else would need a running server AFAIK.

mikemcquaid:
> But if the machine hadn't yet had Java invoked that day, it can take a little while to get the output. Anything else would need a running server AFAIK.

That seems fine for now :+1:

DomT4:
Added the test in a new commit after verifying locally.

Merged in https://github.com/Homebrew/homebrew/commit/604e70e7fb21bb52e230b61667b74fc8b46e1352. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:.

PR #42468, opened 2015-08-04
fswatch 1.5.0

emcrisostomo:


dunn:
Could you add a test while you're at it? Just something that checks basic functionality would be great.

DomT4:
Will pull for now since it has been a couple days. If you can add tests to formulae that don't have them in future though, that would be significantly appreciated.

Merged in https://github.com/Homebrew/homebrew/commit/9028c1f6668df5fb53912389d3f0892f559cff70. Thank you for your contribution to Homebrew!

emcrisostomo:
@dunn, @DomT4, I'm willing to add tests to the formula, sorry for the belated answer. I'll update the formula as soon as I can: is it ok to do without without bumping `fswatch` version or would it be better to wait for the next update?

DomT4:
@emcrisostomo Feel free to submit whenever! No need to wait for a new version, we always appreciate people adding functional tests to formulae at any point.

emcrisostomo:
Thank you @DomT4, that's what I'll do then.

PR #42474, opened 2015-08-05
Add OpenStack support to Ansible.

pwnall:
This essentially adds `shade` and its dependencies to the list of resources. Some dependencies overlap with `pyrax` (which provides RackSpace support).

I installed all the dependencies using pip and used homebrew-pypi-poet to generate the resource stanzas, then I updated the formula. I also tweaked the install code to match the `poet` output a bit better.

I manually checked that the ansible installation works against MIT CSAIL's OpenStack installation.

### DIFF
--- a/Library/Formula/ansible.rb
+++ b/Library/Formula/ansible.rb
@@ -15,13 +15,14 @@ class Ansible < Formula
 
   depends_on :python if MacOS.version <= :snow_leopard
   depends_on "libyaml"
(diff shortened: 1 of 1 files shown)


mikemcquaid:

For future reference the preferred commit message format is `ansible: add openstack support.`.

pwnall:
@mikemcquaid Thank you very much for merging! I'll keep the format in mind for the future!

PR #42497, opened 2015-08-05
fuseki: fix URL for 1.1.2

apjanke:
The current download URL for 404. Switches to "archive.apache.org" where old versions live on.

There's a new fuseki 1.3.0 version out, which Apache recommends switching to. That'll take a little more work since the formula has patches.


dunn:
Merged in 21bb898f6f608355fe9e6dece45d4f779cc75c5c, thanks!

PR #42499, opened 2015-08-05
glassfish: fix URL

apjanke:
The current download URL for glassfish is returning 400 Forbidden errors, for at least a couple days now. Found an alternate download on the main Glassfish site.

### DIFF
--- a/Library/Formula/glassfish.rb
+++ b/Library/Formula/glassfish.rb
@@ -3,8 +3,8 @@ require "formula"
 class Glassfish < Formula
   desc "Java EE application server"
   homepage "https://glassfish.java.net"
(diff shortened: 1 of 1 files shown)


dunn:
Mind bumping the `sha1`s to `sha256` while you're at it?

apjanke:
Sure. Should I squash/amend the current commit, or put the sha256 change in as a separate commit (on the same PR/branch)?

bfontaine:
You can squash it, thanks!

apjanke:
sha256-ified.

dunn:
Merged as e27f2950b82458d2fbb46abde9feed2cbe4dc07a 🎶

PR #42506, opened 2015-08-05
dfix 0.2.2 (new formula)

John-Colvin:
Dfix is a tool for auto-upgrading D source code that uses deprecated or otherwise "bad" syntax.

### DIFF
--- a/Library/Formula/dfix.rb
+++ b/Library/Formula/dfix.rb
@@ -1,0 +1,29 @@
+class Dfix < Formula
+  desc "Auto-upgrade tool for D source code"
+  homepage "https://github.com/Hackerpilot/dfix"
+  url "https://github.com/Hackerpilot/dfix.git",
+      :tag => "v0.2.2",
+      :revision => "ee55825b695fce94a337ab6a2009366372649d46"
+
+  head "https://github.com/Hackerpilot/dfix.git", :shallow => false
+
+  depends_on "dmd" => :build
+
+  def install
+    system "make"
+    system "make", "test"
+    bin.install "bin/dfix"
+    pkgshare.install "test/testfile_expected.d", "test/testfile_master.d"
+  end
+
+  test do
+    system "#{bin}/dfix", "--help"
+
+    cp "#{pkgshare}/testfile_master.d", "testfile.d"
+    system "#{bin}/dfix", "testfile.d"
+    system "diff", "testfile.d", "#{pkgshare}/testfile_expected.d"
+    # Make sure that running dfix on the output of dfix changes nothing.
+    system "#{bin}/dfix", "testfile.d"
+    system "diff", "testfile.d", "#{pkgshare}/testfile_expected.d"
+  end
+end


John-Colvin:
I would have liked to have put the proper test in the ```test do``` block, but it requires test files that are in the source repository.

mikemcquaid:
> I would have liked to have put the proper test in the test do block, but it requires test files that are in the source repository.

You can write one out with `Pathname.write` and `EOS.undent`.

John-Colvin:
```make test``` also uses a hardcoded path to the built executable. A symbolic link would do the trick, but overall it seems easier just to leave it in ```install``` for now.

File: Library/Formula/dfix.rb [hunk 1]
```diff
@@ -0,0 +1,21 @@
...
+  depends_on "dmd" => :build
+
+  def install
+    system "make"
+    system "make", "test"
+    bin.install "bin/dfix"
+  end
+
+  test do
+    system "dfix", "--help"
```
bfontaine: Please use `"#{bin}/dfix"`.

Library/Formula/dfix.rb
```diff
@@ -0,0 +1,21 @@
...
+      :revision => "ee55825b695fce94a337ab6a2009366372649d46"
+
+  head "https://github.com/Hackerpilot/dfix.git", :shallow => false
+
+  depends_on "dmd" => :build
+
+  def install
+    system "make"
+    system "make", "test"
+    bin.install "bin/dfix"
```
bfontaine: Is there a `make install` that we can use?

John-Colvin: no

Library/Formula/dfix.rb
John-Colvin: why? We have to be in the right folder otherwise ```make``` would fail, no?

mikemcquaid: Also, as discussed earlier, please update this test to operate on an actual file you write out.

DomT4: `bin` is the direct path to the formula's Cellar bin, i.e. `/usr/local/Cellar/ducks/1.2.3/bin/`. We use the direct path to ensure we're not calling another tool with the same name somewhere else on the system, for one. A functionality test would be preferable here though, as Mike requests.

John-Colvin: Sorry for the noise, I was looking at the wrong line.

John-Colvin:
@mikemcquaid 
> You can write one out with Pathname.write and EOS.undent.

The files I'm referring to are a bit long to have sitting as strings in the formula file: https://github.com/Hackerpilot/dfix/tree/master/test

mikemcquaid:
@John-Colvin I actually think that's fine (even if they can't be shortened which would be good to try).

John-Colvin:
@mikemcquaid The other problem is that the test files could be dependent on the version (stable vs devel vs head). I could `share.install` the text files and then read them from there?

mikemcquaid:
> @mikemcquaid The other problem is that the test files could be dependent on the version (stable vs devel vs head). I could share.install the text files and then read them from there?

:+1:

DomT4:
`pkgshare` maybe, just to be tidy.

John-Colvin:
should be good to go now.

mikemcquaid:

PR #42509, opened 2015-08-05
bitrise 0.9.7 (new formula)

viktorbenei:


viktorbenei:
Where can I see what the error was? It works on my laptop, every test and check.

```
brew audit --strict bitrise
==> brew style bitrise

```

```
brew install bitrise
==> Downloading https://github.com/bitrise-io/bitrise/archive/0.9.7.tar.gz
Already downloaded: /Library/Caches/Homebrew/bitrise-0.9.7.tar.gz
==> go build -o _bin/bitrise
🍺  /usr/local/Cellar/bitrise/0.9.7: 4 files, 4.8M, built in 2 seconds
```

```
brew test bitrise
Testing bitrise
==> /usr/local/Cellar/bitrise/0.9.7/bin/bitrise --version
```

mikemcquaid:
http://bot.brew.sh/job/Homebrew%20Pull%20Requests/30858/version=yosemite/testReport/junit/brew-test-bot/yosemite/audit_bitrise___strict___online/

viktorbenei:
Thanks @mikemcquaid - I was concerned when I clicked on the red x as it requested access to read my account's orgs, that's why I did not follow the link.

viktorbenei:
@mikemcquaid it should pass now - should I update the pull request or can you restart the PR check (as no code change is required)?

viktorbenei:
I updated the PR to start a new build - it seems it passed on two machines but it's stuck on the third for some reason @mikemcquaid

viktorbenei:
Is there anything else required?

Library/Formula/bitrise.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  depends_on "go" => :build
+
+  def install
+    mkdir_p buildpath/"src/github.com/bitrise-io"
+    ln_s buildpath, buildpath/"src/github.com/bitrise-io/bitrise"
+
+    ENV["GOPATH"] = "#{buildpath}/Godeps/_workspace:#{buildpath}"
+    Language::Go.stage_deps resources, buildpath/"src"
+    system "go", "build", "-o", "_bin/bitrise"
```
bfontaine: You can do `system "go", "build", "-o", "#{bin}/bitrise"` and remove the `bin.install` line.

Library/Formula/bitrise.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    bin.install "_bin/bitrise"
+  end
+
+  test do
+    output = shell_output(bin/"bitrise --version")
```
bfontaine: Can’t we do a better test than e.g. `--version` or `--help`?

Library/Formula/bitrise.rb
viktorbenei: I started with that solution, but unfortunately we have a folder named `bitrise` in the source folder (this is because of how Go projects with packages are structured) and so a simple `-o bitrise` would fail (as a folder with the same name already exists).

bfontaine: I’m not talking of a `-o bitrise` but a `-o #{bin}/bitrise`, which will work, because `#{bin}` is something like `/usr/local/Cellar/bitrise/0.9.7/bin`.

Library/Formula/bitrise.rb
viktorbenei: To have a proper test we should run `go test ./...` - is the code still available in the `test do` block?

I copied the structure of https://github.com/Homebrew/homebrew/blob/master/Library/Formula/docker-swarm.rb as I had trouble with compiling the project, building a Go project in a context like what `brew` provides for `install` is not trivial unfortunately.

Library/Formula/bitrise.rb
viktorbenei: Ohh, I see, sure, if you think that's a better way to do it it's certainly possible.

P.S.: I copied the structure from https://github.com/Homebrew/homebrew/blob/master/Library/Formula/docker-swarm.rb

Library/Formula/bitrise.rb
bfontaine: No the code is not available anymore. I’m not familiar with `bitrise`, but could we write a dummy `bitrise.yml` file then run `bitrise` and check it did execute the correct commands? We don’t need something complicated; we just want to check `bitrise` was correctly installed and works as expected.

viktorbenei: We can do that, but it's a bit more complicated.

A sample `bitrise.yml` like this would be enough:

```
format_version: 0.9.8
default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib
app:
  envs: []
workflows:
  test:
    title: Workflow for master branch
    envs: []
    steps:
    - script:
        title: Test
        inputs:
        - content: |-
            #!/bin/bash
            set -e
            echo "Test - OK"
```

But to run this you first have to call a `bitrise setup`, because bitrise depends on two other tools to run these configurations (https://github.com/bitrise-io/envman/ and https://github.com/bitrise-io/stepman/).

You could use `bitrise` without `stepman` but in that case it's not possible to use a Step Library manager (`stepman`), which means that the steps used in `bitrise.yml` have to be included relative to the `bitrise.yml`, similar to the setup you can find here: https://github.com/bitrise-io/bitrise/tree/master/_examples/tutorials/steps-and-workflows - which would make the test more complicated, a bit too complicated IMO.

bfontaine: Is there a reason not to install these tools in `def install`?

viktorbenei: `stepman` is not required but in a normal use case it'll be used, that's why it's part of the `bitrise setup` process to get it.

`envman` could also be optional, it's required for the current `bitrise` version but technically it's also optional; but again, in a normal setup the user will want to have it available, because `envman` is responsible for exposing environment variables between independent steps, so a subsequent step can access a previous step's exposed environments.

bfontaine: If those are required for the general case we should install them.

viktorbenei: Can we instead call `bitrise setup` in the `do test` block, and run the test after that? `bitrise` depends on specified versions of these tools and so an older version of either tool might cause issues - that's why `bitrise setup` checks the two required tool's version as well, and installs the required versions if no appropriate version found.

Both tools are simple Go binaries, the install consists of a `curl` download and a `chmod +x` for both.

bfontaine: We need to download them ourselves in order to verify their checksum and cache them. We can download a specific version for each one and we’ll update the versions when we’ll update the formula for the next releases.

viktorbenei: I'm OK with this solution, just to be sure:

1. If we do this (to download the binary or source of the other two tools) and we update `bitrise` in the future which will require newer versions of these tools it will update the tools automatically as well, right? Even if a previous version of `envman` or `stepman` is already available?
2. If we have to build the other two tools from source: can you point to another Formula which does this and is written in Go or uses another Go tool, like we do? It wasn't exactly simple to prepare a proper Go environment for the build, it would help a lot to see another Formula which builds it's Go dependency.

Thank you!

bfontaine: 1. Yes.

bfontaine: 2\. You can build them like the `bitrise` binary. Put both resources before the `def install`:

```rb
resource "envman" do
  url "..."
  sha256 "..."
end

resource "stepman" do
  url "..."
  sha256 "..."
end
```

Then in the `install` method:

```rb
resource("envman").stage do
  # use Dir.pwd to get the current directory (instead of `buildpath`)
  # ...
end

resource("stepman").stage do
  # same here
end
```

bfontaine: I don’t know of any formula which builds its dependencies like that, but there might be one. I’ll check tonight if you didn’t succeed in the meantime.

viktorbenei: One last question @bfontaine : we're working on the 0.9.8 version of `bitrise` which will automatically do a `setup` in case the user forgot to call `setup` before calling `bitrise run`, because our beta testers forgot this frequently. Should we still go with building these dependencies, like the way you described?

With `0.9.8` the `test do` should work without any further setup, because the setup check is included in the "run" action.

bfontaine: It’s probably better to build them in the formula, that way `brew install bitrise` takes care of everything; but you can still keep your `setup` for people who didn’t install it through Homebrew.

viktorbenei: All right, I'm sold, it's probably a redundancy worth the time dealing with :)

Thank you again for your help, I'll do the changes asap (in a couple of hours).

viktorbenei:
Updated with compiling both `envman` and `stepman`, as @bfontaine recommended (thank you for the code layout example!) and the `test do` block now contains a more complete test.

Library/Formula/bitrise.rb
```diff
@@ -0,0 +1,74 @@
...
+        title: Test
+        inputs:
+        - content: |-
+            #!/bin/bash
+            set -e
+            pwd
+            printf 'Test - OK' > brew.test.file
+"
+
+    testpath.cd do
```
bfontaine: You’re already in the `testpath`, no need to `cd` into it :wink:

viktorbenei: Ohh, OK, I guess I wanted to make it sure :smile: - I'll fix it tomorrow morning

Library/Formula/bitrise.rb
```diff
@@ -0,0 +1,74 @@
...
+      File.write("bitrise.yml", test_bitrise_yml_content)
```
bfontaine: You can use:

```rb
(testpath/"bitrise.yml").write <<-EOS.undent
  format_version: 0.9.8
  default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib
  # ...
EOS
```

Library/Formula/bitrise.rb
```diff
@@ -0,0 +1,74 @@
...
+      shell_output(bin/"bitrise run test_wf")
```
bfontaine: `shell_output` is used to get the output of the command. If you don’t need it, you should use `system`:

```rb
system "#{bin}/bitrise", "run", "test_wf"
```

Library/Formula/bitrise.rb
```diff
@@ -0,0 +1,74 @@
...
+      test_file_contents = File.read("brew.test.file")
+      assert_match "Test - OK", test_file_contents
```
bfontaine: You can merge both lines above with:

```rb
assert_equal "Test - OK", (testpath/"brew.test.file").read.chomp
```

Library/Formula/bitrise.rb
```diff
@@ -0,0 +1,74 @@
...
+      go_install_package.call(Dir.pwd, "stepman")
+    end
+
+    go_install_package.call(buildpath, "bitrise")
+  end
+
+  test do
+    # simple -version call
+    output = shell_output(bin/"bitrise --version")
+    assert_match "#{version}", output
```
bfontaine: You can remove this simple test since there’s a more complex one below.

Library/Formula/bitrise.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def install
+    go_install_package = lambda do |basepth, go_packagename|
+      mkdir_p "#{basepth}/src/github.com/bitrise-io"
+      ln_s basepth, "#{basepth}/src/github.com/bitrise-io/#{go_packagename}"
+
+      ENV["GOPATH"] = "#{basepth}/Godeps/_workspace:#{basepth}"
+      Language::Go.stage_deps resources, "#{basepth}/src"
+      system "go", "build", "-o", "#{bin}/#{go_packagename}"
+    end
```
bfontaine: You can move this in a separate method outside of `install`, e.g.:

```rb
def go_install_package(basepath, pkgname)
  # ...
end
```

Then call it: `go_install_package Dir.pwd, "envman".

viktorbenei: I thought it's better to have this inside `install` as it's not used anywhere else, and that way for a brew Formula noob like myself it's probably more obvious what's required in a Formula and what's just utility code.

Library/Formula/bitrise.rb
viktorbenei: Thanks, done. Quick question: I tried to search for a documentation about what's available in a Formula.rb - which gems/modules in addition to the standard Ruby ones.

Although it seems that `File` is not available by default on the Mountain Lion CI server, but is available on the other two..

Library/Formula/bitrise.rb
viktorbenei: For future reference, where can I find docs about what testing methods (like `assert_equal`) can be used in the `test do` block?

viktorbenei:
Squashed into one commit

viktorbenei:
Is there anything else I could do / should improve?

mikemcquaid:

viktorbenei:
Thank you @mikemcquaid , really appreciated!

PR #42511, opened 2015-08-05
json-fortran 4.2.0

zbeekman:
Changes from the 4.1.1 release that have required changes to the formula:

 1. CMake Makefile targets and target dependencies now emulate GNU auto tools with `make check` target
     - `all` target builds library but no longer builds tests, for faster builds
     - `check` target builds library, and tests, and runs the tests
     - `install` will build libraries prior to installing them, but won’t trigger build of tests
 1. Documentation generator switched from [ROBODoc](https://github.com/gumpu/ROBODoc) to [FORD](https://github.com/cmacmackin/ford)
     - Documentation builds have become pretty slow
     - CMake system dynamically computes FORD outputs, so FORD is run at CMake/Config time, *BUT* outputs are reused/preserved so that they don’t need to be rebuilt when running `make`, unless the source file has changed
     - Bottle installs should help make this speedier
     - In the future, a CMake option may be added to build documentation without search functionality, which is what causes FORD to be so slow—it may make sense to propagate that option into the Homebrew Formula to allow users to create documentation more quickly by removing the search functionality.

Full changes can be viewed in the [CHANGELOG](https://github.com/jacobwilliams/json-fortran/blob/master/CHANGELOG.md#420-2015-08-03) and the [latest release notes](https://github.com/jacobwilliams/json-fortran/releases/latest)

### DIFF
--- a/Library/Formula/json-fortran.rb
+++ b/Library/Formula/json-fortran.rb
@@ -1,8 +1,8 @@
 class JsonFortran < Formula
   desc "Fortran 2008 JSON API"
   homepage "https://github.com/jacobwilliams/json-fortran"
-  url "https://github.com/jacobwilliams/json-fortran/archive/4.1.1.tar.gz"
-  sha256 "97f258d28536035ef70e9ead5c7053e654106760a12db2cc652587ed61b76124"
+  url "https://github.com/jacobwilliams/json-fortran/archive/4.2.0.tar.gz"
+  sha256 "5cf12ce527809b644ae35a0a34ce330804285d3b7987b17488d1752330dd7a81"
 
   head "https://github.com/jacobwilliams/json-fortran.git"
 
@@ -15,9 +15,11 @@ class JsonFortran < Formula
 
   option "with-unicode-support", "Build json-fortran to support unicode text in json objects and files"
   option "without-test", "Skip running build-time tests (not recommended)"
-  option "without-robodoc", "Do not build and install ROBODoc generated documentation for json-fortran"
+  option "without-docs", "Do not build and install FORD generated documentation for json-fortran"
 
-  depends_on "robodoc" => [:recommended, :build]
+  deprecated_option "without-robodoc" => "without-docs"
+
+  depends_on "ford" => :build if build.with? "docs"
   depends_on "cmake" => :build
   depends_on :fortran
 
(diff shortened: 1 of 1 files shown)


Library/Formula/json-fortran.rb
```diff
@@ -15,9 +15,9 @@ class JsonFortran < Formula
 
   option "with-unicode-support", "Build json-fortran to support unicode text in json objects and files"
   option "without-test", "Skip running build-time tests (not recommended)"
-  option "without-robodoc", "Do not build and install ROBODoc generated documentation for json-fortran"
+  option "without-ford", "Do not build and install FORD generated documentation for json-fortran"
```
dunn: People who built `without-robodoc` will probably want `without-ford` too.  How about making the option `without-docs` and adding `deprecated_option "without-robodoc" => "without-docs"`?

zbeekman: Good thinking… I’m still learning the ins and outs of homebrew formula authorship…

dunn: No problem! The rest looks great.

zbeekman: @dunn, I can’t seem to determine the answer to this question from  the Formula cookbook page:

> How do I treat the ford dependency if I change `without-ford` to `without-docs`?

Would my `depends_on "ford"` line look like this:
```ruby
depends_on "ford" => ["with-docs", :recommended, :build]
```
?

dunn: I think this will work:
```ruby
depends_on "ford" => :build if build.with? "docs"
```

zbeekman: OK, this seems to do the trick. FORD is listed as a build dependency but passing `--without-docs` does not trigger FORD to be downloaded and installed.

zbeekman:
Please do me a favor and hold off merging this for a sec… there may be a patch version bump for json-fortran soon… I’m investigating. Doesn’t make much sense to merge this, at least IMO, if I’ll just need to update the SHA256 and download URL later this afternoon… I’ll report back when I know more, probably later this afternoon.

dunn:
Got it, thanks for the heads-up.

zbeekman:
The changes appear to be too minor to necessitate a patch release, so this is ready to merge AFAIAC.

:+1:

dunn:
Merged in f33ef8d5b5c4a34e0ba2527f4ee9680af1e3d11c, then! ⚡️

PR #42514, opened 2015-08-05
gauge 0.1.7

apoorvam:


Library/Formula/gauge.rb
```diff
@@ -74,6 +74,7 @@ def install
       system "go", "run", "build/make.go"
       system "go", "run", "build/make.go", "--install", "--prefix", prefix
     end
+    puts "To install Gauge plugin, run `gauge --install <plugin_name>`. Language plugins supported are `java`, `csharp` and `ruby`. Report plugins available are `html-report` and `xml-report`."
```
bfontaine: You can remove this line.

apoorvam:
A print statement after successful installation was intentionally written. Isn't this possible to have it?

bfontaine:
It depends if it’s something that’s strictly related to Homebrew or a general information. We prefer not to print the latter case since that’s something that should be in the program documentation. If it’s something that’s related to the way Homebrew install `gauges` you can use the caveats:

```rb
def caveats; <<-EOS.undent
    your text …
    …
  EOS
end
```

Library/Formula/gauge.rb
```diff
@@ -76,6 +76,11 @@ def install
     end
   end
 
+  def caveats; <<-EOS.undent
+    Installing gauge with homebrew doesn't install Gauge plugins which are required to run Gauge specs.
+    To install Gauge plugin, please explicitly run `gauge --install <plugin_name>`. Language plugins supported are `java`, `csharp` and `ruby`. Report plugins available are `html-report` and `xml-report`.
```
DomT4: If we *have* to have these caveats they should be a lot shorter, or at least broken onto multiple lines.

I'm not sure why we couldn't have options like `without-java-plugin` and then in `postinstall` do something like `gauge --install java if build.with? "java-plugin"` or such, but either way, the caveats need to be peeled back a bit.

mahendrakariya: Hi @DomT4,

Ideally, we would like to install the plugins in `postinstall`. But we leave it upto the users to select the plugins that they require. Which means we need to have interactive homebrew formula for installing the plugins in `postinstall`. Since homebrew discourages this practice, we have put in a caveat. 

So, as suggested, we will update the pull request with the caveat text broken into multiple lines.

mikemcquaid:
@apoorvam Where are plugins installed?

apoorvam:
@mikemcquaid Gauge plugins are installed in $HOME/.gauge/plugins. 

As mentioned earlier, using options like --install_with_java and installing them in `post-install` is an option for us. We'll look into this. 

We have made shorter caveats in the new commit so that this PR can be accepted.

mikemcquaid:
@apoorvam We can't write outside `HOMEBREW_PREFIX` in `post_install` (particularly to `$HOME`).

I do think these caveats aren't Homebrew-specific so shouldn't be added here but instead more prominently to your homepage/documentation.

mikemcquaid:

PR #42019, opened 2015-07-22
luvit: version bump to 2.3.0

tzudot:
luvit: version bump to 2.3.0

PR #42525, opened 2015-08-05
Update recipie to new luvi / lit / luvit system

creationix:
This builds luvi from the source tarball (with locally built luajit, libuv, and openssl statically linked).  Then using luvi, lit builds itself from zipball and then lit builds luvit from zipball.

Including openssl, luajit, and libuv statically is the default behavior in luvit, but this could be changed to depend on the homebrew versions of these dependencies.

See luvit/luvit#787 and #42019 

Thanks for the help and encouragement from @tzudot.

### DIFF
--- a/Library/Formula/lit.rb
+++ b/Library/Formula/lit.rb
@@ -1,0 +1,18 @@
+class Lit < Formula
+  desc "Toolkit for developing, sharing, and running luvit programs and libraries."
+  homepage "https://github.com/luvit/lit"
+  url "https://lit.luvit.io/packages/luvit/lit/v2.2.4.zip"
+  sha256 "4d0ad9dce10c89e2f1cf483435c4670f5b3a22f69f73f834804b8b232411791f"
+
+  depends_on "luvi" => :build
+
+  def install
+    system "luvi", buildpath, "--", "make"
+    bin.install "lit"
+  end
+
+  test do
+    system "#{bin}/lit"
+  end
+end
+
--- a/Library/Formula/luvi.rb
+++ b/Library/Formula/luvi.rb
@@ -1,0 +1,19 @@
+class Luvi < Formula
+  desc "A project in-between luv and luvit, lua packages."
+  homepage "https://github.com/luvit/luvi"
+  url "https://github.com/luvit/luvi/releases/download/v2.1.8/luvi-src-v2.1.8.tar.gz"
+  sha256 "1ef449c6a76bcf7a03c8bbcdd5423b0366d63b10c52fd463ca2ba8aaf9248781"
+
+  depends_on "cmake" => :build
+
+  def install
+    system "make", "regular-asm"
+    system "make"
+    bin.install "build/luvi"
+  end
+
+  test do
+    system "#{bin}/luvi", "samples/test.app", "--", "1", "2", "3", "4"
+  end
+end
+
--- a/Library/Formula/luvit.rb
+++ b/Library/Formula/luvit.rb
@@ -1,31 +1,18 @@
 class Luvit < Formula
-  desc "Asynchronous I/O for Lua"
-  homepage "https://luvit.io"
-  url "https://luvit.io/dist/latest/luvit-0.8.2.tar.gz"
-  sha256 "c2639348d1716c38ac3cd66ea4c4ff1c8a72f4610dbd6e50cf31426d3956c5ff"
-  head "https://github.com/luvit/luvit.git"
-  revision 1
+  desc "Asynchronous I/O for Lua (node.js style powered by luajit and libuv)"
+  homepage "https://luvit.io/"
+  url "https://lit.luvit.io/packages/luvit/luvit/v2.4.2.zip"
+  sha256 "e4173ffae63b364c6f1c37fc4a7ce28880e4a4cda65f894cf436d268055fce17"
 
-  bottle do
-    cellar :any
-    sha256 "3c7250314f67d320d8bc36f14e8d055860a2098bf5528b53eb58ff7d93244881" => :yosemite
-    sha256 "1f4616aa1be7802088900bcec2f284e8a85a4b82ee6fc9873a4c20d30ac6ac47" => :mavericks
-    sha256 "ef23e4e0cf252074738d85b1b2b2e02fd58e3a5239e00362d6f1e4dc2c58862c" => :mountain_lion
-  end
-
-  depends_on "pkg-config" => :build
-  depends_on "luajit"
-  depends_on "openssl"
+  depends_on "lit" => :build
 
   def install
-    ENV["USE_SYSTEM_SSL"] = "1"
-    ENV["USE_SYSTEM_LUAJIT"] = "1"
-    ENV["PREFIX"] = prefix
-    system "make"
-    system "make", "install"
+    system "lit", "make", buildpath
+    bin.install "luvit"
   end
 
   test do
-    system bin/"luvit", "--cflags", "--libs"
+    system "#{bin}/luvit", "-v"
   end
 end
+


Library/Formula/lit.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  depends_on "luvi" => :build
+
+  def install
+    system "luvi", buildpath, "--", "make"
+    bin.install "lit"
+  end
+
+  test do
+    system "#{BIN}/lit"
```
dunn: `bin` is lower-case.

Library/Formula/lit.rb
```diff
@@ -0,0 +1,18 @@
+class Lit < Formula
+  desc "Toolkit for developing, sharing, and running luvit/lua programs and libraries."
```
DomT4: I could be wrong but this *might* be a tad long. Limit is 80 characters, including the formula name. `brew audit --strict` would flag it if it's too long.

Library/Formula/luvi.rb
```diff
@@ -0,0 +1,19 @@
...
+  depends_on "cmake" => :build
+
+  def install
+    system "make", "regular-asm"
+    system "make"
+    bin.install "build/luvi"
+  end
+
+  test do
+    system "make", "test"
```
DomT4: This won't work as you expect. The `testpath` is a separate sandboxed entity from `buildpath`, the two never exist simultaneously.

Library/Formula/luvit.rb
```diff
@@ -1,31 +1,18 @@
...
-    ENV["PREFIX"] = prefix
-    system "make"
-    system "make", "install"
+    system "lit", "make", buildpath
+    bin.install "luvit"
   end
 
   test do
-    system bin/"luvit", "--cflags", "--libs"
+    system "#{BIN}/luvit", "-v"
```
DomT4: Same thing per @dunn's comment above - Should be `bin/` or `#{bin}/`

Library/Formula/luvi.rb
creationix: So the problem is it will build the main binary twice?  If so, I can just run the contents of the test target directly using `bin/luvi`.

DomT4: It won't do anything. The `makefile` won't be moved across to the `testpath`, so theoretically it's likely the system will just exit with `make: *** No rule to make target'.` or similar. Once the installation has finished you have access to whatever was installed; the `buildpath` is gone.

If you can write an example to the testpath and use that for testing, that'd be awesome.

creationix: So this should work, but I need to manually copy the test app to the test path as part of the build step? https://github.com/creationix/homebrew/commit/65f3ff283c5b037807260be76b0ed0a2efc6303a#diff-fa25f5d63de4da959fc7c7808ad02f02R16

dunn:
The tests are failing with some SHA mistmatches.

Library/Formula/luvi.rb
```diff
@@ -0,0 +1,19 @@
+class Luvi < Formula
+  desc "A project in-between luv and luvit, lua packages."
+  homepage "https://github.com/luvit/luvi"
+  url "https://github.com/luvit/luvi/releases/download/v2.1.8/luvi-src-v2.1.8.tar.gz"
+  sha256 "1ef449c6a76bcf7a03c8bbcdd5423b0366d63b10c52fd463ca2ba8aaf9248781"
+
+  depends_on "cmake" => :build
+
+  def install
+    system "make", "regular-asm"
```
DomT4: Where's the OpenSSL coming from here?

```
==> make regular-asm
cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Release -DWithSharedLibluv=OFF -DWithPackageSH=ON -DWithPackageTGZ=ON -DWithPackageTBZ2=ON -DWithOpenSSL=ON -DWithSharedOpenSSL=OFF -DWithOpenSSLASM=ON
```

Library/Formula/luvi.rb
```diff
@@ -0,0 +1,19 @@
...
+    system "make"
+    bin.install "build/luvi"
+  end
+
+  test do
+    system "#{bin}/luvi", "samples/test.app", "--", "1", "2", "3", "4"
```
DomT4: `samples/test.app` won't exist at this point. Buildpath and Testpath are separate entities that are never in the same place at once. You'll either need to write the various necessary files in as something like:

```ruby
test do
  (testpath/"testing.test").write <<-EOS.undent
    something to test
  EOS

  system bin/"luvi", "testing.test", "something", "another flag", "so on"
end
```

Or you can do something like this:

```ruby
def install
  system "make", "regular-asm"
  system "make"
  bin.install "build/luvi"
  (pkgshare/"examples").install Dir["samples/*"]
end
```

And then:

```ruby
test do
  system bin/"luvi", pkgshare/"examples/test.app", "--", "1", "2", "3", "4"
end
```

You may find the second way nicer/easier.

Library/Formula/luvi.rb
creationix: Yes as I stated in the original PR description.

Library/Formula/luvi.rb
creationix: Thanks, this is helpful.

Library/Formula/luvi.rb
```diff
@@ -0,0 +1,19 @@
...
+  homepage "https://github.com/luvit/luvi"
+  url "https://github.com/luvit/luvi/releases/download/v2.1.8/luvi-src-v2.1.8.tar.gz"
+  sha256 "1ef449c6a76bcf7a03c8bbcdd5423b0366d63b10c52fd463ca2ba8aaf9248781"
+
+  depends_on "cmake" => :build
+
+  def install
+    system "make", "regular-asm"
+    system "make"
+    bin.install "build/luvi"
```
mikemcquaid: Let's use the `make install` and submit a patch upstream to allow the prefix to be configurable.

Library/Formula/lit.rb
```diff
@@ -0,0 +1,18 @@
...
+  desc "Toolkit for developing, sharing, and running luvit programs and libraries."
+  homepage "https://github.com/luvit/lit"
+  url "https://lit.luvit.io/packages/luvit/lit/v2.2.4.zip"
+  sha256 "4d0ad9dce10c89e2f1cf483435c4670f5b3a22f69f73f834804b8b232411791f"
+
+  depends_on "luvi" => :build
+
+  def install
+    system "luvi", buildpath, "--", "make"
+    bin.install "lit"
```
mikemcquaid: Is there some sort of `make install` we can use?

Library/Formula/luvit.rb
```diff
@@ -1,31 +1,18 @@
...
+  depends_on "lit" => :build
 
   def install
-    ENV["USE_SYSTEM_SSL"] = "1"
-    ENV["USE_SYSTEM_LUAJIT"] = "1"
-    ENV["PREFIX"] = prefix
-    system "make"
-    system "make", "install"
+    system "lit", "make", buildpath
+    bin.install "luvit"
```
mikemcquaid: Is there some sort of `make install` we cna use?

Library/Formula/luvit.rb
```diff
@@ -1,31 +1,18 @@
...
   end
 
   test do
-    system bin/"luvit", "--cflags", "--libs"
+    system "#{bin}/luvit", "-v"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikemcquaid:
Does it make sense to have these in three formulae instead of one? Are these other formulae actually going to be used outside of `luvit`?

Library/Formula/luvi.rb
creationix: I'll take a patch if you want.

creationix:
@mikemcquaid I'd be fine with moving them to all one formula.  I just thought it was better to have more.

mikemcquaid:
@creationix One formula would be great, thanks :

Library/Formula/luvi.rb
DomT4:
```diff
LUVI_TAG=$(shell git describe)
LUVI_ARCH=$(shell uname -s)_$(shell uname -m)
LUVI_PUBLISH_USER?=luvit
LUVI_PUBLISH_REPO?=luvi

OS:=$(shell uname -s)
+ prefix = /usr/local
+ bindir = $(prefix)/bin

install: luvi
- install -p build/luvi /usr/local/bin
+ install -p build/luvi $(bindir)

uninstall:
- rm -f /usr/local/bin/luvi
+ rm -f $(bindir)/luvi
```

Should work.

rphillips: I added the prefix code in master.

bfontaine:
ping?

tzudot:
I am not sure about the current status on this PR but it seems that there is a [blocking issue](https://github.com/luvit/lit/issues/115) at the moment.

creationix:
Mostly I'm just stuck working on higher-priority tasks at the moment.  I'll happily accept PRs to luvit or changes to this recipie.

@tzudot Yes, there are issues with auto-update conflicting with homebrew's system.  As I noted in that issue, I'm not sure what the best solution is.

The easiest and least hacky solution is to simply disable lit update if it's detected to be part of a homebrew install and instead tell the user to update via homebrew.

But this will only work if the homebrew recipe is able to keep up with the development pace.  Right now there are several releases a week.

DomT4:

PR #42526, opened 2015-08-06
squid 3.5.7

stuwil:


Library/Formula/squid.rb
```diff
@@ -58,4 +58,11 @@ def plist; <<-EOS.undent
     </plist>
     EOS
   end
+
+  test do
+    system "#{sbin}/squid"
```
DomT4: If this is launching the server, we'd prefer to do something like [this](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mpd.rb#L127-L139) because that way if one stage fails the server isn't left running on the system.

If there's a problem after the server is started & before the test completes otherwise, the process is left running.

dunn:
Merged in 10a0350d792599db62552ab2d1748328c54527a9, thanks :octopus:

PR #42527, opened 2015-08-06
changed :yosemite to :mavericks

thos37:
tested this on my mavericks machine and it runs fine


dunn:
Great, merged in 4ba9d9c7c6da11ae09924295bd4d6e87ba8295c5~

PR #42528, opened 2015-08-06
Bumped dependency-check to 1.3.0

colezlaw:
This bumps the dependency-check formula to 1.3.0.

### DIFF
--- a/Library/Formula/dependency-check.rb
+++ b/Library/Formula/dependency-check.rb
@@ -1,9 +1,9 @@
 class DependencyCheck < Formula
   desc "OWASP Dependency Check"
   homepage "https://www.owasp.org/index.php/OWASP_Dependency_Check"
-  url "https://dl.bintray.com/jeremy-long/owasp/dependency-check-1.2.11-release.zip"
-  version "1.2.11"
-  sha256 "1a622a1c79a7bff88950fa42294d634a8155958960d656d9d796785f056d88b2"
+  url "https://dl.bintray.com/jeremy-long/owasp/dependency-check-1.3.0-release.zip"
+  version "1.3.0"
+  sha256 "075f48aca75000a51c1741a612f3732517c1bb4c779103656f6a403c82a93b71"
 
   depends_on :java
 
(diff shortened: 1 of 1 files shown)


Library/Formula/dependency-check.rb
```diff
@@ -25,6 +25,10 @@ def install
     (etc/"dependencycheck").mkpath
     corejar = libexec/"repo/org/owasp/dependency-check-core/#{version}/"\
       "dependency-check-core-#{version}.jar"
+    if File.exist? etc/"dependencycheck/dependencycheck.properties"
```
DomT4: Does upstream actually overwrite the configuration files if one already exists? That's fairly unusual behaviour.

colezlaw: The overwrite happens on the following line. The default configuration is a properties file in the JAR - the installer extracts it as a matter of convenience so that the user has a place to modify the configuration that's picked up by default. We unzip that file out on the following line - in the bump from 1.2.11 to 1.3.0, new properties were added, so I want to make sure we get the new options.

mikemcquaid: Please use `etc.install` which handles the modified case and overwriting automatically, thanks.

Library/Formula/dependency-check.rb
```diff
@@ -32,7 +36,7 @@ def install
 
   test do
     output = `#{libexec}/bin/dependency-check --version`.strip
-    assert_equal("Dependency-Check Core version 1.2.11", output)
+    assert_equal("Dependency-Check Core version 1.3.0", output)
```
mikemcquaid: Use `version` in here instead of hard-coding.

mikemcquaid: OR just `assert_match("Dependency-Check Core version ", output)`

colezlaw:
Using `install_symlink` to go the other direction now (unzip to `libexec` then link into `etc`). This also removes the hard-coding of the version number in the test.

Library/Formula/dependency-check.rb
```diff
@@ -26,13 +26,14 @@ def install
...
-    libexec.install_symlink etc/"dependencycheck" => "etc"
+      libexec/"etc"
+    etc.install_symlink libexec/"etc/dependencycheck.properties" => \
+      "dependencycheck/dependencycheck.properties"
   end
 
   test do
     output = `#{libexec}/bin/dependency-check --version`.strip
-    assert_equal("Dependency-Check Core version 1.2.11", output)
+    assert_match("Dependency-Check Core version ", output)
```
dunn: Could do `assert_match("Dependency-Check Core version #{version}", output)`

colezlaw: Yeah - was going to do that, but it was suggested by @mikemcquaid that I leave the version out altogether. Will put it in.

mikemcquaid:

PR #42530, opened 2015-08-06
x264: r2555 in stable and r2579 in devel

henry0312:


DomT4:
Looks like `ffmbc`, `ffmpeg` and `libav` all need their revisions bumped here.

henry0312:
Sorry, I don't understand what you mean...

DomT4:
The new versions here cause compatibility failures on several formulae that depend on `x264`. So user's installations don't break we force updates of the other tools at the same time. In the case of `ffmbc` you'd do this:

`brew edit ffmbc`
```diff
  homepage "https://code.google.com/p/ffmbc/"
  url "https://drive.google.com/uc?export=download&id=0B0jxxycBojSwTEgtbjRZMXBJREU"
  version "0.7.2"
  sha256 "caaae2570c747077142db34ce33262af0b6d0a505ffbed5c4bdebce685d72e42"
+ revision 1
```

Then commit the change with `git add -A && git commit -m "ffmbc: revision for x264 update"`

And then repeat the steps for `ffmpeg` and `libav`.

henry0312:
Thank you for your description. I see.
I'll do after my job.

henry0312:
Updated.
Thank you.

Library/Formula/ffmpeg.rb
```diff
@@ -4,9 +4,9 @@ class Ffmpeg < Formula
   url "https://ffmpeg.org/releases/ffmpeg-2.7.2.tar.bz2"
   sha256 "7ceb7550ad628c526fa6c9ff23fdfb687a62f54d90c4a730998d8c2b417b9ef2"
   head "https://github.com/FFmpeg/FFmpeg.git"
+  revision 2
```
DomT4: This should be `revision 1`. The bottle `revision` and the formula `revision` serve different purposes and bottle revisions don't impact formula revisions.

Library/Formula/ffmpeg.rb
```diff
@@ -4,9 +4,9 @@ class Ffmpeg < Formula
...
 
   bottle do
-    revision 1
```
DomT4: Likewise, it's perfectly safe to ignore the `revision` inside of `bottle do` blocks per the above.

henry0312:
updated!

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `x264 r2555, r2579 (devel)`.

PR #42532, opened 2015-08-06
subversion 1.9.0

nijikon:


DomT4:
Looks like `mesos` might need a revision.

nijikon:
@DomT4 how does the test bot know to test mesos? I would run this locally if I would know how :)

DomT4:
It will test anything that has a `depends_on "subversion"` line & has a `test do` block defined in this case.

The only way to reproduce that behaviour locally is to do something like `git checkout /my/pr/branch && brew test-bot --skip-setup subversion` AFAIK.

nijikon:
@DomT4 sweet, will build everything with this from now on.

DomT4:
Just be wary that it'll dump bottle files in whatever directory you call it from. I usually switch to `~/Downloads` or even `/tmp` before running any `brew test-bot` commands just to avoid spamming up `/usr/local`.

nijikon:
Could we do everything but not store the file when ```HOMEBREW_DEVELOPER=1```?

DomT4:
Sadly in many cases, people who have `HOMEBREW_DEVELOPER=1` set do actually want the bottle to be generated. You could pass `brew test-bot --skip-setup --no-bottle` thinking about it, that should work.

mikemcquaid:
> It will test anything that has a depends_on "subversion" line & has a test do block defined in this case.

And a bottle, just FYI

brainy:
This change is broken wrt requiring non-system APR because that makes the build mod_dav_svn and mod_authz_svn incompatible with the system httpd.

nijikon:
@brainy does it build for you with the system httpd?

brainy:
Absolutely does, but whoever updates the formula really should read the change logs. The pull request I sent for 1.9.0 does it right. As a Subversion developer I should know, right? :)

https://github.com/Homebrew/homebrew/pull/42697

nijikon:
@brainy I can see that you added, ```--enable-broken-httpd-auth``` that is why it did not build for me. I will bring it back.

brainy:
Ack, the system httpd does not have the fixes released in apache-2.4.16

nijikon:
@brainy thanks for all the feedback! If there is anything wrong with it still, feel free to comment :)

nijikon:
@DomT4 done, please verify :)

DomT4:
> The pull request I sent for 1.9.0 does it right. As a Subversion developer I should know, right? :)

Sometimes "I work on the upstream project and this is a bug that needs fixing" is helpful information to include in your PR :wink:.

> Ack, the system httpd does not have the fixes released in apache-2.4.16

OS X 10.10.5 does:

```
Server version: Apache/2.4.16 (Unix)
Server built:   Jul 22 2015 21:07:41
```

I'll probably make note of that in the formula on pull so we can cap the line to <10.10 in future.

Library/Formula/subversion.rb
```diff
@@ -100,7 +96,7 @@ def install
                 CFLAGS=#{ENV.cflags} LINKFLAGS=#{ENV.ldflags}
                 OPENSSL=#{Formula["openssl"].opt_prefix}]
 
-      unless MacOS::CLT.installed?
+      if build.with?("apr") || !MacOS::CLT.installed?
```
DomT4: This will never get triggered on systems without the CLT installed. I killed off my own CLT and:

```
scons: Reading SConscript files ...
/bin/sh: /usr/local/opt/apr: No such file or directory
OSError: '/usr/local/opt/apr --cflags --cppflags --ldflags --includes --link-ld --libs' exited 127:
  File "/private/tmp/subversion--serf20150810-80964-6wliba/serf-1.3.8/SConstruct", line 341:
    env.ParseConfig('$APR --cflags --cppflags --ldflags --includes'
  File "/usr/local/Cellar/scons/2.3.6/libexec/scons-local/SCons/Environment.py", line 1582:
    return function(self, self.backtick(command))
  File "/usr/local/Cellar/scons/2.3.6/libexec/scons-local/SCons/Environment.py", line 597:
    raise OSError("'%s' exited %d" % (command, status))
```

The current dependency resolution works because it knows when to enforce the fallback, and that happens by default. For users to get that behaviour now they'd have to pass `brew install svn --with-apr` which isn't intuitive when that hasn't never been our default behaviour.

We need to keep the `depends_on :apr => :build` line and make it work with the newly necessary flag.

nijikon:
@DomT4 Done, please verify :)

Library/Formula/subversion.rb
```diff
@@ -30,11 +24,12 @@ class Subversion < Formula
 
   resource "serf" do
     url "https://serf.googlecode.com/svn/src_releases/serf-1.3.8.tar.bz2", :using => :curl
+    mirror "https://fossies.org/linux/www/serf-1.3.8.tar.bz2"
     sha256 "e0500be065dbbce490449837bb2ab624e46d64fc0b090474d9acaa87c82b2590"
   end
 
   depends_on "pkg-config" => :build
-  depends_on :apr => :build
+  depends_on "apr" => :build
```
DomT4: This will do the opposite and *always* enforce Homebrew's `apr` :wink:. I don't think that's the worst thing in the world personally, but we try hard to avoid unnecessary system dupes and this would be one.

nijikon:
@DomT4 Hope this is final ;)

Library/Formula/subversion.rb
```diff
@@ -156,6 +151,8 @@ def install
     if MacOS::CLT.installed?
       args << "--with-apr=/usr"
       args << "--with-apr-util=/usr"
+      # Ignore authz API bug in system httpd (fixed in httpd 2.4.16)
+      args << "--enable-broken-httpd-auth" if MacOS.version < :el_capitan
```
DomT4: One last thing: Does this need fixing for Homebrew's `apr` and `apr-util` as well?

nijikon: You are right, no.

Library/Formula/subversion.rb
```diff
@@ -156,6 +151,8 @@ def install
     if MacOS::CLT.installed?
       args << "--with-apr=/usr"
       args << "--with-apr-util=/usr"
+      # Ignore authz API bug in system httpd (fixed in httpd 2.4.16)
+      args << "--enable-broken-httpd-auth" if !MacOS::CLT.installed? && MacOS.version < :el_capitan
```
DomT4: `!MacOS::CLT.installed?` will never be fulfilled inside a `if MacOS::CLT.installed?` block. I think the previous version of this line was fine - My question earlier was whether or not we needed to apply the fix to the `Formula["apr"]` usage as well in the block below, rather than just when the system `apr` was used. Apologies if that was unclear.

Library/Formula/subversion.rb
```diff
@@ -277,23 +274,23 @@ def caveats
 index 445251b..6ff4332 100755
 --- a/configure
 +++ b/configure
-@@ -25366,6 +25366,8 @@ fi
+@@ -26127,6 +26127,8 @@ fi
  SWIG_CPPFLAGS="$CPPFLAGS"
-
```
DomT4: Careful of the whitespacing changes here.

Library/Formula/subversion.rb
nijikon: Should have not touch this yesterday, was too tired. This line is not needed with brew ```apr```, it compiles without issues. @brainy am I correct on this?

Library/Formula/subversion.rb
nijikon: Magic vim command ;)

nijikon:
@DomT4 Are we done with this?

DomT4:
Was just waiting to see if @brainy confirmed the Homebrewed Apr didn't face the same problem. Will merge it soon otherwise.

nijikon:
Ok, let's wait :)

Library/Formula/subversion.rb
DomT4: I'm likely to sit on this for a few days. Apple have just released updates to Yosemite & Mavericks which remove the need to pass `--enable-broken-httpd-auth`, which means we could limit it to Mountain Lion and below which is nicer. Have asked the other maintainers if anyone objects.

brainy:
So, my 10.10 ships with httpd-2.4.9, not 2.4.16. I've never built Subversion on OSX with anything but the stock APR and HTTPD; it would be, IMO, a very bad idea to *require* Homebrewed versions of these packages.

DomT4:
We're not requiring them. Update to 10.10.5 - It has the httpd fix.

DomT4:
My reluctance to leave the workaround in for Yosemite & Mavericks is that Apple has pushed updates to fix the security flaws on those platforms & I don't love the idea of facilitating users ignoring security holes. Either way, there's no serious discussion about forcing a Homebrew Apr usage.

brainy:
I agree that leaving in security holes is bad. My best suggestion in this case is that, if we're building on a platform without the httpd security fix, just don't build mod_dav_svn and mod_authz_svn by default (i.e., pass --without-apxs to configure) and perhaps add an option for the user to explicitly request building these modules with httpd < 2.4.16.

brainy:
Unless we're using brewed APR/httpd of course

brainy:
FWIW: The --enable-broken-httpd-auth line is only applicable to httpd >= 2.4.0 and < 2.4.16; 2.2.x isn't affected (this includes osx 10.9 and older)

DomT4:
We're likely to sit on this for a week, and then merge it afterwards without the `--enable-broken-httpd-auth` line added. If people haven't updated their OS X versions on 10.9 & 10.10 to handle the security problem in that time, that's on them essentially. Facilitating people ignoring security updates now there are those updates available isn't something Homebrew should be doing.

mikemcquaid:
> We're likely to sit on this for a week, and then merge it afterwards without the --enable-broken-httpd-auth line added. If people haven't updated their OS X versions on 10.9 & 10.10 to handle the security problem in that time, that's on them essentially. Facilitating people ignoring security updates now there are those updates available isn't something Homebrew should be doing.
>  Show all checks

:+1:

brainy:
So, 10.10.5 does have httpd-2.4.16, but the headers aren't updated even with the newest CLT: ap_release.h still says 2.4.10.

brainy:
Apple acknowledged the header-version bug today. The fix is already in 10.11 beta; i expect they'll release a fix for 10.10 soon.

nijikon:
:heart:

markphip:
Where does this stand?  SVN 1.9.1 has been out for a while and 1.8.13 is still the newest package available.  Are we still waiting for Apple fixes for httpd or something?

DomT4:
There doesn't seem to be an upcoming Yosemite release with any huge urgency. We're stuck between a rock and a hard place on moving forwards on this at the moment.

brainy:
I suggest that the Subversion formula should be changed to only build the client by default and require an explicit option to build the server-side bits. If the user requires the server bits (specifically: mod_authz_svn), we can then issue a warning if we detect a version of httpd that has the known security issue.

I'm prepared to make the necessary modifications for this, but would like to start essentially from scratch. The version detection is already present in Subversion's configure scripts, but would have to be duplicated in the formula. It's not all that hard, really.

Unfortunately, as I mentioned elsewhere, this particular issue can't be detected in httpd at run-time because it depends on the presence of a new module API.

markphip:
1.8.13 has the security problem too, so why couldn't we still release an up-to-date version that continues to allow the security problem?  Although, I would also be in favor of focusing the package on the client and making the Apache modules an explicit option.  It seems counter-productive to hold up releasing the client over this issue.  FWIW, this is a somewhat BS security issue anyway in that it is confined to a specific Apache configuration scenario and something that someone that tests their configuration could see themselves.    Point being, I'd rather release with the "broken" module then not at all.

DomT4:
There's an Xcode/CLT update out now (A few hours ago, even) that alleges to fix the problem for 10.10 and 10.11, which should bring an end to this saga relatively soon.

brainy:
The saga won't end so soon. The Xcode 7 CLT contain a totally broken Ruby framework with missing header files; so, installing Subversion --with-ruby will not work.

markphip:
Will someone with a brand new Mac installing Subversion --with-ruby work regardless?  I do not see how 1.8.13 would work either.

markphip:
If the formula is not going to get updated due to these problems, can the "devel" version at least get updated to 1.9.1?  It seems like it is still 1.9.0-rc3.

DomT4:
> installing Subversion --with-ruby will not work.

Can you provide an example of the breakage? Potentially worth noting is that the Ruby bindings trigger `:std` env which uses whatever Ruby happens to be the first in your `$PATH`.

brainy:
The Ruby framework in CLT 7.0 is incomplete; some headers are missing. This is what I reported to Apple and what happens during Subversion's configure:
```
xcrun clang -o conftest -I/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/include/ruby-2.0.0/universal-darwin15 -I/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/include/ruby-2.0.0/ruby/backward -I/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/include/ruby-2.0.0 -I. -D_XOPEN_SOURCE -D_DARWIN_C_SOURCE -D_DARWIN_UNLIMITED_SELECT -D_REENTRANT    -g -Os -pipe -DHAVE_GCC_SYNC_BUILTINS conftest.c  -L. -L/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib -L. -L/usr/local/lib   -arch i386 -arch x86_64   -lruby.2.0.0  -lpthread -ldl -lobjc
In file included from conftest.c:1:
In file included from /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/include/ruby-2.0.0/ruby.h:33:
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/include/ruby-2.0.0/ruby/ruby.h:24:10: fatal error: 'ruby/config.h' file not found
#include "ruby/config.h"
         ^
```

brainy:
As to "whatever Ruby happens to be in your `$PATH`" ... it's either the OS default Ruby framework, or we should make the Subversion formula depend on Homebrew's version. I'm not too thrilled by the second option.

bfontaine:

n054:
I can not run this continuous unless I have a way to find the file. When I get to the Formula file it only shows a few files. I didn't find this one. I'll look again.

PR #42534, opened 2015-08-06
libressl 2.2.2

nijikon:


DomT4:
A few more changes need to made to reflect upstream:

* All the direct dependants need revisions. There's only one in the core by default.
* `--with-enginesdir=#{lib}/engines` has been made obsolete and can be removed.
* Lines 39-41 can be killed. The `.cnf` is installed automatically now.
* They also now install a default `.pem` which is a bit of a PITA. We probably need to actually remove that in `postinstall` and then write a new one (The latter part of this already happens).

nijikon:
Sorry, for not responding to this in time.

mikemcquaid:
No worries @nijikon, your commit was included and useful!

Issue #42533, opened 2015-08-06
Bug with new screenbrightness links


chdiza:
If I do `brew install brightness`, then `brew rm brightness`, then do e.g., `brew outdated` or almost any other brew command, I get the following: `Error: No such file or directory - /usr/local/Cellar/screenbrightness`.

Research reveals that the following files/dirs aren't getting removed by either `brew rm brightness` or `brew rm screenbrightness`:

```
/usr/local/bin/brightness
/usr/local/cellar/brightness
```

This must have something to do with aliases, but I am unable to solve.

chdiza:
`brew rm screenbrightness` gives the same error.

xu-cheng:
Please paste the output when you run these commands with `-vd`. Thanks.

chdiza:
Ah turns out this has to do with custom formulas.  I have a custom formula for `brightness` (that's what `screenbrightness` really is now, but we merely have `brightness` as an alias for `screenbrightness`).  The formula itself doesn't do anything fancy or special, or indeed anything that the now-official formula doesn't do.

If I `cd` to the dir holding my custom formula, and do `brew install -d brightness.rb`, it builds and installs and works fine.

If I then do (e.g.,) `brew -vd outdated` I get:
```
brew outdated -vd
/Applications/homebrew/Library/brew.rb (Formulary::FormulaLoader): loading /Applications/homebrew/Library/Formula/apg.rb
/Applications/homebrew/Library/brew.rb (Formulary::FormulaLoader): loading /Applications/homebrew/Library/Taps/chdiza/homebrew-diza/appswitch.rb
/Applications/homebrew/Library/brew.rb (Formulary::FormulaLoader): loading /Applications/homebrew/Library/Formula/autoconf.rb
/Applications/homebrew/Library/brew.rb (Formulary::FormulaLoader): loading /Applications/homebrew/Library/Formula/automake.rb
/Applications/homebrew/Library/brew.rb (Formulary::FormulaLoader): loading /Applications/homebrew/Library/Formula/base64.rb
/Applications/homebrew/Library/brew.rb (Formulary::FormulaLoader): loading /Applications/homebrew/Library/Formula/blueutil.rb
/Applications/homebrew/Library/brew.rb (Formulary::AliasLoader): loading /Applications/homebrew/Library/Formula/screenbrightness.rb
[...]

PR #42537, opened 2015-08-06
ARGV#resolved_formulae: use canonical_name to locate rack

xu-cheng:


bfontaine:
Is this related to #42533?

xu-cheng:
> Is this related to #42533?

Dunno. But that issue reminds me of another problem, hence this PR.

mikemcquaid:
Makes sense :+1:

PR #42539, opened 2015-08-06
make path instead of symlink for lib/R

tdsmith:
Moves us towards being able to support formulae that install R bindings,
like nonpareil in homebrew-science.

Some discussion in Homebrew/homebrew-science#2559.


mikemcquaid:
:+1:

PR #42540, opened 2015-08-06
Update rubocop

xu-cheng:
cc @mikemcquaid

### DIFF
--- a/Library/Homebrew/test/test_inreplace.rb
+++ b/Library/Homebrew/test/test_inreplace.rb
@@ -85,7 +85,7 @@ class InreplaceTest < Homebrew::TestCase
     s.sub!("f", "b")
     assert_equal "boo", s
 
-    s.gsub!("o", "e")
+    s.tr!("o", "e")
     assert_equal "bee", s
   end
 
@@ -99,7 +99,7 @@ class InreplaceTest < Homebrew::TestCase
     end
 
     assert_raises(Utils::InreplaceError) do
-      inreplace("test") { |s| s.gsub! "d", "f" }
+      inreplace("test") { |s| s.tr!("d", "f") }
     end
 
     assert_raises(Utils::InreplaceError) do
--- a/Library/Homebrew/test/test_inreplace.rb
+++ b/Library/Homebrew/test/test_inreplace.rb
@@ -85,7 +85,8 @@ class InreplaceTest < Homebrew::TestCase
     s.sub!("f", "b")
     assert_equal "boo", s
 
-    s.tr!("o", "e")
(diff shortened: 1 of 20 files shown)


Library/Homebrew/test/test_inreplace.rb
```diff
@@ -85,7 +85,7 @@ def test_sub_gsub
     s.sub!("f", "b")
     assert_equal "boo", s
 
-    s.gsub!("o", "e")
+    s.gsub!("o", "e") # rubocop:disable Performance/StringReplacement
```
mikemcquaid: Why the ignore here?

Library/Homebrew/test/test_inreplace.rb
```diff
@@ -99,7 +99,7 @@ def test_inreplace_errors
     end
 
     assert_raises(Utils::InreplaceError) do
-      inreplace("test") { |s| s.gsub! "d", "f" }
+      inreplace("test") { |s| s.gsub!("d", "f") } # rubocop:disable Performance/StringReplacement
```
mikemcquaid: and here

Library/.rubocop.yml
```diff
@@ -2,6 +2,7 @@ AllCops:
   Exclude:
     - 'Homebrew/vendor/**/*'
     - 'Homebrew/test/vendor/**/*'
+    - 'Homebrew/compat/**/*'
```
mikemcquaid: Maybe we should fix these up anyway?

Library/Homebrew/test/test_inreplace.rb
xu-cheng: Because under current context we are testing the method of `String#gsub!`, while rubocop will try to replace it to `tr!`.

Library/.rubocop.yml
xu-cheng: I add this to ignore below warning:
```
== Library/Homebrew/compat/hardware.rb ==
C:  3:  9: Rename is_32_bit? to 32_bit?.
C:  7:  9: Rename is_64_bit? to 64_bit?.
```

mikemcquaid: Ok. Can we ignore the specific rule and add a comment explaining why instead, maybe?

Library/Homebrew/test/test_inreplace.rb
mikemcquaid: Would be good to add a comment to explain that in more detail :+1:

xu-cheng:
Updated.

mikemcquaid:
:+1: :shipit:

PR #42545, opened 2015-08-06
fdroidserver 0.4.0 (new formula)

eighthave:
Tools for creating and managing Android app repositories for F-Droid
https://f-droid.org

### DIFF
--- a/Library/Formula/fdroidserver.rb
+++ b/Library/Formula/fdroidserver.rb
@@ -1,0 +1,125 @@
+class Fdroidserver < Formula
+  desc "Create and manage Android app repositories for F-Droid"
+  homepage "https://f-droid.org/"
+  url "https://pypi.python.org/packages/source/f/fdroidserver/fdroidserver-0.4.0.tar.gz"
+  sha256 "89742f2e03e082c9d185835dbb4f166dade19a0e257b247a6e0bca3b4fd9b523"
+
+  depends_on :java => "1.7+"
+  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on "freetype"
+  depends_on "jpeg"
+  depends_on "libyaml"
+  depends_on "android-sdk" => :recommended
+  depends_on "libmagic" => :recommended
+
+  resource "apache-libcloud" do
+    url "https://pypi.python.org/packages/source/a/apache-libcloud/apache-libcloud-0.17.0.tar.gz"
+    sha256 "8ac4895c5ed2fa51812237dfd587675e3cbc4b7e57d9b44722ce849eab2131c2"
+  end
+
+  resource "backports.ssl_match_hostname" do
+    url "https://pypi.python.org/packages/source/b/backports.ssl_match_hostname/backports.ssl_match_hostname-3.4.0.2.tar.gz"
+    sha256 "07410e7fb09aab7bdaf5e618de66c3dac84e2e3d628352814dc4c37de321d6ae"
+  end
+
+  resource "mwclient" do
+    url "https://pypi.python.org/packages/source/m/mwclient/mwclient-0.7.2.zip"
+    sha256 "195462115fadbe3cee73b5a0fe3cc28c8bdc0115c590cdc0f8b5083861003d1a"
+  end
+
+  resource "paramiko" do
+    url "https://pypi.python.org/packages/source/p/paramiko/paramiko-1.15.2.tar.gz"
+    sha256 "4f56a671a3eecbb76e6143e6e4ca007d503a39aa79aa9e14ade667fa53fd6e55"
+  end
+
+  resource "pillow" do
+    url "https://pypi.python.org/packages/source/P/Pillow/Pillow-2.9.0.tar.gz"
+    sha256 "0f179d7e75e7c83b6341b9595ca1f394de7081484a9e352ad66d553a1c3daa29"
+  end
+
+  resource "pyasn1" do
+    url "https://pypi.python.org/packages/source/p/pyasn1/pyasn1-0.1.7.tar.gz"
+    sha256 "e4f81d53c533f6bd9526b047f047f7b101c24ab17339c1a7ad8f98b25c101eab"
+  end
+
+  resource "pyasn1-modules" do
+    url "https://pypi.python.org/packages/source/p/pyasn1-modules/pyasn1-modules-0.0.5.tar.gz"
+    sha256 "be65f00ed28e30756f1ef39377cb382480a2368699179d646a84d79fe9349941"
+  end
+
+  resource "python-magic" do
+    url "https://pypi.python.org/packages/source/p/python-magic/python-magic-0.4.6.tar.gz"
+    sha256 "903d3d3c676e2b1244892954e2bbbe27871a633385a9bfe81f1a81a7032df2fe"
+  end
+
+  resource "PyYAML" do
+    url "https://pypi.python.org/packages/source/P/PyYAML/PyYAML-3.11.tar.gz"
+    sha256 "c36c938a872e5ff494938b33b14aaa156cb439ec67548fcab3535bb78b0846e8"
+  end
+
+  resource "requests" do
+    url "https://pypi.python.org/packages/source/r/requests/requests-2.7.0.tar.gz"
+    sha256 "398a3db6d61899d25fd4a06c6ca12051b0ce171d705decd7ed5511517b4bb93d"
+  end
+
+  def install
+    ENV["PYTHONPATH"] = Formula["libdnet"].opt_lib/"python2.7/site-packages"
+    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
+    ENV.append_to_cflags "-I#{Formula["freetype"].opt_include}/freetype2"
+    resources.each do |r|
+      r.stage do
+        system "python", *Language::Python.setup_install_args(libexec/"vendor")
+      end
+    end
+
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    system "python", *Language::Python.setup_install_args(libexec)
+    bin.install Dir["#{libexec}/bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  def caveats; <<-EOS.undent
+    In order to function, fdroidserver requires that the Android SDK's
+    "Build-tools" and "Platform-tools" are installed.  Also, it is best if the
+    base path of the Android SDK is set in the standard environment variable
+    ANDROID_HOME.  To install them from the command line, run:
+    android update sdk --no-ui --all --filter platform-tools,build-tools-23.0.1
+    EOS
+  end
+
+  test do
+    # fdroid prefers to work in a dir called 'fdroid'
+    mkdir testpath/"fdroid" do
+      mkdir "repo"
+      mkdir "metadata"
+
+      open("config.py", "w") do |f|
+        f << "gradle = 'gradle'"
+      end
+
+      open("metadata/fake.txt", "w") do |f|
+        f << "License:GPL\n"
+        f << "Summary:Yup still fake\n"
+        f << "Categories:Internet\n"
+        f << "Description:\n"
+        f << "this is fake\n"
+        f << ".\n"
+      end
+
+      system "#{bin}/fdroid", "checkupdates", "--verbose"
+      system "#{bin}/fdroid", "lint", "--verbose"
+      system "#{bin}/fdroid", "rewritemeta", "fake", "--verbose"
+      system "#{bin}/fdroid", "scanner", "--verbose"
+
+      # TODO: enable once Android SDK build-tools are reliably installed
+      # ENV["ANDROID_HOME"] = Formula["android-sdk"].opt_prefix
+      # system "#{bin}/fdroid", "readmeta", "--verbose"
+      # system "#{bin}/fdroid", "init", "--verbose"
+      # assert File.exist?("config.py")
+      # assert File.exist?("keystore.jks")
+      # system "#{bin}/fdroid", "update", "--create-metadata", "--verbose"
+      # assert File.exist?("metadata")
+      # assert File.exist?("repo/index.jar")
+    end
+  end
+end


Library/Formula/fdroidserver.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def caveats; <<-EOS.undent
+    fdroidserver requires the Java JDK and Android SDK to work. Get the JDK from Oracle:
+       http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html
+    And the Android SDK from Google (either Android Studio or Standalone SDK):
+       https://developer.android.com/sdk/installing
+    For more info:
+       https://sites.google.com/a/android.com/tools/tech-docs/configuration/osx-jdk
+    EOS
+  end
```
mikemcquaid: Just do `depends_on :java` and `depends_on "android_sdk"`

Library/Formula/fdroidserver.rb
```diff
@@ -0,0 +1,99 @@
...
+        assert File.exist?("config.py")
+        assert File.exist?("keystore.jks")
+        system "#{bin}/fdroid", "update", "--create-metadata", "--verbose"
+        assert File.exist?(fdroidrepo/"metadata")
+        assert File.exist?(fdroidrepo/"repo/index.jar")
+      end
+    else
+      puts "Tests require the env var ANDROID_HOME to point to the Android SDK!"
+      puts "Skipping the test."
+    end
```
mikemcquaid: This test should instead be modified to find the Android SDK based on `depends_on "android_sdk"` and test that. We can't rely on this variable being set.

Library/Formula/fdroidserver.rb
```diff
@@ -0,0 +1,99 @@
...
+    For more info:
+       https://sites.google.com/a/android.com/tools/tech-docs/configuration/osx-jdk
+    EOS
+  end
+
+  test do
+    if ENV["ANDROID_HOME"]
+      fdroidrepo = testpath/"fdroid"
+      fdroidrepo.mkdir()
+      fdroidrepo.cd do
```
mikemcquaid: Not clear why this needs to be done in a created subdirectory

eighthave: the fdroid tools prefer working in a directory called `fdroid`, which in turn matches how the published repo looks on the webserver.

Library/Formula/fdroidserver.rb
eighthave: The problem is that this actually depends on the `build-tools` section of the Android SDK, which is not automatically installed when doing `brew install android-sdk`.  Also, I think a lot of people get their Android SDK via the Android Studio bundle, even with brew installed, that's why I want to make it "recommended" rather than required.

Library/Formula/fdroidserver.rb
eighthave: I left out `depends_on :java` because it wasn't clear to me that that would also install the JDK.  Plus that might install the JDK6, which is tool old for `fdroidserver`.  Oracle seems to have dropped its JDK7 downloads and only has JDK8 now.

Library/Formula/fdroidserver.rb
mikemcquaid: Ok. In that case you can use `mkdir testpath/"fdroid" do`

Library/Formula/fdroidserver.rb
mikemcquaid: `depends_on "android-sdk" => :recommended` is what we want.

Library/Formula/fdroidserver.rb
mikemcquaid: `depends_on :java => "1.7"`

eighthave: Makes sense.  I couldn't find any docs on the `depends_on :java` stuff, but I think this sounds like it addresses the issues.

dunn:
We're getting a build error on all three platforms:
```
_imagingft.c:31:10: fatal error: 'ft2build.h' file not found
#include <ft2build.h>
         ^
```
And there are two audit warnings:
```
==> audit problems
fdroidserver:
Length is calculated as fdroidserver + desc. (currently 82)
 * Dependency 'libjpeg' is an alias; use the canonical name 'jpeg'.
```

eighthave:
I fixed the description and changed 'libjpeg' to 'jpeg'.  About the ft2build.h error, that one is stumping me.  Python Pillow needs freetype 2 to build, and that is what the `freetype` brew package seems to provide.  So it seems like there is a bug somewhere because the Pillow build cannot find that freetype header. I bumped the Pillow version to the latest (v2.9.0) in the hopes that it might help.  The odd thing is that it was working on my machine when I first submitted this formula here.

DomT4:
You may need to do something like 

```ruby
ENV.append_to_cflags "-I#{Formula["freetype"].opt_include}/freetype2"
```

eighthave:
@DomT4 that did it, thanks!  Now the last remaining issue is the tests.  In order to actually run the tests, I'd need to run the Android SDK installer to install the "SDK Build-tools" package, i.e.
```
android update sdk --no-ui --filter tool
```

Is that something that's OK to do in the tests?

eighthave:
I added this to the test section, and it worked!

```ruby
  test do
    exec "echo yes | android -v update sdk --no-ui --force --filter platform-tools,build-tools-22.0.1"
    ...
```

DomT4:
> android update sdk --no-ui --filter tool

It depends on how large a download it is, but as long as it isn't huge and doesn't take too long, sure.

You could do something like:

```ruby
test do
  system bin/"android", "update", "sdk", "--no-ui", "--filter tool"
  assert File.exist?("expected_downloaded_file.file")
end
```

mikemcquaid:
Yeh, I'm up for including it :+1:

mikemcquaid:
You could even add that to the `android-sdk` formula and install it by default for everyone?

eighthave:
I just found a missing dependency for apache-libcloud, `backports.ssl_match_hostname`, and added that to this recipe.  This is ready to be merged! :-)

Library/Formula/fdroidserver.rb
```diff
@@ -0,0 +1,104 @@
...
+    "Build-tools" and "Platform-tools" are installed.  Also, it is best if the
+    base path of the Android SDK is set in the standard environment variable
+    ANDROID_HOME.  To install them from the command line, run:
+    android update sdk --no-ui --filter platform-tools,build-tools-22.0.1
+    EOS
+  end
+
+  test do
+    # fdroid requires "SDK Platform-tools" and "SDK Build-tools"
+    exec "echo yes | android -v update sdk --no-ui --force --filter platform-tools,build-tools-22.0.1"
```
mikemcquaid: Can you try to use `pipe_output` instead of using a pipe here? Otherwise `system` rather than `exec` here.

Library/Formula/fdroidserver.rb
```diff
@@ -0,0 +1,104 @@
...
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  def caveats; <<-EOS.undent
+    In order to function, fdroidserver requires that the Android SDK's
+    "Build-tools" and "Platform-tools" are installed.  Also, it is best if the
+    base path of the Android SDK is set in the standard environment variable
+    ANDROID_HOME.  To install them from the command line, run:
+    android update sdk --no-ui --filter platform-tools,build-tools-22.0.1
+    EOS
```
mikemcquaid: It'd be good to fix the android-sdk formula to maybe just install the platform tools and also if you can make this formula autoinstall (as you do in `test do`) in `postinstall`

Library/Formula/fdroidserver.rb
eighthave: I can try, but I really don't know ruby at all, can't you tell? ;-)

bfontaine: @eighthave Instead of using `exec "echo something | some command here"` use `pipe_output("some command here", "something", 0)`.

Library/Formula/fdroidserver.rb
eighthave: Build-tools and Platform-tools are both required to actually run builds.  Why install the android-sdk install in `postinstall` and not as a regular dependency?

As for modifying the android-sdk package, I do not feel like I'm a good candidate since I don't do Android development on OSX and don't use that package beyond this dependency.  I don't have any info on the use cases for how people use it, so I feel like I could easily be messing things up for people who use it.

eighthave: Also, it feels somehow wrong to me for this fdroidserver formula to modify something outside of its domain, as in installing things in the Android SDK.  But maybe that's thinking too Debian-ish and brew users expect that?

mikemcquaid: Sorry, I'm saying the `android-sdk` formula could/should be modified to install these things. I really don't think it'll mess anything up.

Library/Formula/fdroidserver.rb
mikemcquaid: :+1:

Library/Formula/fdroidserver.rb
eighthave: I looked into doing this.  There are a lot of notes and caveats so its a complicated thing.  I don't currently have time to take on that project.  For example:

    # The 'android' tool insists on deleting #{prefix}/platform-tools
    # and then installing the new one. So it is impossible for us to redirect
    # the SDK location to var so that the platform-tools don't have to be
    # freshly installed EVERY DANG time the base SDK updates.

Library/Formula/fdroidserver.rb
eighthave: Ok, I updated it to use `pipe_output()`

eighthave: Well, the `pipe_output()` fails where the `exec()` worked just fine.  This worked:

bfontaine: @eighthave when you do it by hand, what’s the exit code?

```sh
echo y | android -v update sdk --no-ui --force --filter platform-tools,build-tools-22.0.1
echo $?
```

If it’s not `0`, replace the last argument of `pipe_output` with the correct number. [See here](https://github.com/Homebrew/homebrew/blob/4fbd40a4ebace7bf0809effd62a4fa3a1993d899/Library/Formula/moreutils.rb#L64) for an example.

Library/Formula/fdroidserver.rb
```diff
@@ -0,0 +1,105 @@
...
+    base path of the Android SDK is set in the standard environment variable
+    ANDROID_HOME.  To install them from the command line, run:
+    android update sdk --no-ui --filter platform-tools,build-tools-23.0.0
+    EOS
+  end
+
+  test do
+    # fdroid requires "SDK Platform-tools" and "SDK Build-tools"
+    output = pipe_output("android -v update sdk --no-ui --force --filter platform-tools,build-tools-22.0.1", "y\n", 0)
+    print output
```
bfontaine: Please remove this line.

eighthave:
This `pipe_output()` change seems to break things on the jenkins build.  This recipe builds and tests fine on my Mavericks laptop, and I can't seem to figure out how to get the output from the `android` command on the jenkins build.  How about I switch back to the `exec` call that worked?

Library/Formula/fdroidserver.rb
```diff
@@ -0,0 +1,104 @@
...
+    "Build-tools" and "Platform-tools" are installed.  Also, it is best if the
+    base path of the Android SDK is set in the standard environment variable
+    ANDROID_HOME.  To install them from the command line, run:
+    android update sdk --no-ui --all --filter platform-tools,build-tools-23.0.1
+    EOS
+  end
+
+  test do
+    # fdroid requires "SDK Platform-tools" and "SDK Build-tools"
+    pipe_output("android --verbose update sdk --no-ui --all --filter platform-tools,build-tools-23.0.1", "y\n")
```
bfontaine: Use `"#{Formula["android-sdk"].bin}/android ...` instead of `"android ...".

Library/Formula/fdroidserver.rb
```diff
@@ -0,0 +1,104 @@
...
+    # fdroid prefers to work in a dir called 'fdroid'
+    mkdir testpath/"fdroid" do
+      ENV["ANDROID_HOME"] = "/usr/local/opt/android-sdk"
```
bfontaine: Use `Formula["android-sdk"].opt_prefix`

eighthave:
I made the changes that @bfontaine suggested and now it builds on the travis 10.9 and 10.10, but the tests fail on 10.11.  So there is progress!

Looks like the `android` command failed to install *build-tools-23.0.1*, but there is no debug log in the travis output, even though the command is run using `--verbose`.  Seems the test setup might still be a bit brittle.

Perhaps switching back to `exec` will work? It did before...

mikemcquaid:
@eighthave Once #46575 is merged you should be able to just apply:

```patch
--- a/Library/Formula/fdroidserver.rb
+++ b/Library/Formula/fdroidserver.rb
@@ -88,8 +88,6 @@ class Fdroidserver < Formula
   end
 
   test do
-    # fdroid requires "SDK Platform-tools" and "SDK Build-tools"
-    pipe_output("#{Formula["android-sdk"].bin}/android --verbose update sdk --n
o-ui --all --filter platform-tools,build-tools-23.0.1", "y\n")
```
and have this work (at least it does for me locally).

bfontaine:
@mikemcquaid You seem to have a newline issue in your diff :wink:

mikemcquaid:
@bfontaine Whoops, thanks, fixed!

bfontaine:
ping re Mike’s comment?

eighthave:
Ok, made the change from mike, the travis tests failed the first time saying it couldn't find build-tools again, but then the second time they worked.  On jenkins, only the el_capitan test failed saying 

```
==> audit problems
fdroidserver:
 * The homepage is not reachable (curl exit code 22)
```

So you probably want to increase the timeout on that test.

mikemcquaid:
@eighthave Looks like the test is still failing due to the build tools being missing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/41718/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/test_fdroidserver/

Not sure the best way around that? Maybe copy the `android-sdk` installation inside the `testpath`, install the build tools and then run the test as-is?

eighthave:
How about I remove the tests and get this merged, then come back around to the tests in an update.

mikemcquaid:
@eighthave Sounds good but rather than remove replace with a `fdroid --version` or similar (or if even that isn't doable: a `File.exist? bin/"fdroid"`)

eighthave:
Ok, I added some simpler tests and they pass on travis.  I'm still getting the errand false negative result about the home page, this time on 10.9:
```
fdroidserver:

 * The homepage is not reachable (curl exit code 28)
```
Perhaps that needs multiple retries and a long timeout.  I find that CI systems often have crappy internet response.

DomT4:
Travis is ignorable on that. Seems to be having some connectivity issues this week. You passed Jenkins fine.

mikemcquaid:
Thanks for this and your patience @eighthave! Got there in the end!

PR #42550, opened 2015-08-06
add fully qualified name for core formulae

xu-cheng:
Since #41006, #41736 and #42350 all requires such feature, I'd like to merge this first.

cc @mikemcquaid @CNA-Bld @vladshablinsky


bfontaine:
Will this change the returned values of `Formula#full_name`?

xu-cheng:
No, it won't. It's only used internally to indicate `homebrew/homebrew/foo` is not a tap formula.

mikemcquaid:
:+1:

PR #42556, opened 2015-08-06
graphviz HEAD

kbrock:
Update graphviz to also build from HEAD

Also update graphviz to just use their standard configuration script, making the patch no longer necessary.

Most people will be using the bottle anyway, so it will not affect them.
But this will allow people to use custom build options and build from head.

**NOTE:** I had thought I would need to add `depends_on "gd" => :build` but brew refuses it. All built fine without it.

For testing:

I built with `--HEAD` and verified a new feature that was added.
I built without any options (but not using bottle). Verified it did converted my sample file.

Thanks all

### DIFF
--- a/Library/Formula/graphviz.rb
+++ b/Library/Formula/graphviz.rb
@@ -4,6 +4,14 @@ class Graphviz < Formula
   url "http://graphviz.org/pub/graphviz/stable/SOURCES/graphviz-2.38.0.tar.gz"
   sha256 "81aa238d9d4a010afa73a9d2a704fc3221c731e1e06577c2ab3496bdef67859e"
 
+  head do
+    url "https://github.com/ellson/graphviz.git"
+
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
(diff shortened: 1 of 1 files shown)


Library/Formula/graphviz.rb
```diff
@@ -1,6 +1,7 @@
 class Graphviz < Formula
   desc "Graph visualization software from AT&T and Bell Labs"
   homepage "http://graphviz.org/"
+  head "https://github.com/ellson/graphviz.git"
```
DomT4: You want:

```ruby
head do
  url "https://github.com/ellson/graphviz.git"

end
```

And then:

```ruby
```

In `def install`.

kbrock:
Thank you @DomT4 - you were spot on.
Let me know if you have any other pointers

dunn:
Looks good!  Merged in a4efc2a6c3fcd668903906f578dffc519cb8af49.

kbrock:
@dunn thanks for the help

PR #42564, opened 2015-08-06
enable periodic status updates, useful for bufferbloat project

kim0:


DomT4:
What does this do/change?

kim0:
This commit adds the --enable-demo compile time option, which enables the -D run time parameter. When specified -D, enables periodic reporting of results. This option is needed by https://github.com/tohojo/flent which is the main testing tool for the bufferbloat project (which measures latency degradation under highly utilized links)

kim0:
Looking into this failure .. It seems I need to patch 2 functions to remove "inline" from them. I guess this is too invasive. I got it to compile locally, but I guess I will abort this pull request

PR #42565, opened 2015-08-06
show caveats after post_install

rkitover:
So I was doing a brew -v install postgresql yesterday, and I ended up going to #machomebrew to ask why there wasn't a launchctl config for it, when this is clearly described in the caveats.

The problem was that the post_install for postgresql runs initdb, which creates the database directory and displays its own instructions for how to start the database, and I thought these were the instructions from homebrew.

So to reduce possible confusion for other users, I wrote this patch which moves the caveats display to after the post_install. More details in the commit message.

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -176,7 +176,6 @@ module Homebrew
     fi.debug               = ARGV.debug?
     fi.prelude
     fi.install
-    fi.caveats
     fi.finish
   rescue FormulaInstallationAlreadyAttemptedError
     # We already attempted to install f as part of the dependency tree of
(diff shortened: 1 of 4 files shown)


xu-cheng:
Make sense to me. Will let other maintainers take a look as well.

mikemcquaid:
:+1:

xu-cheng:
On second thought, I think it's better to put `caveats` call inside `finish`, as such:
```
def finish
  ...

  caveats

  ohai "Summary" if verbose? || show_summary_heading?
  puts summary
ensure
  unlock
end
```

The reason is we want to invoking caveats within the keg lock.

rkitover:
Xu, that is certainly more elegant, I have updated my PR.

xu-cheng:

PR #42570, opened 2015-08-06
git-multipush: new URLs, test added

bfontaine:


bfontaine:
Mmh I’ll look into this failing test.

PR #42573, opened 2015-08-06
bazaar fixes

dlh:
With the changes in a838988c4da05db3da941f1fa1f4ae0192753fc9, plugins included with bzr are copied into `var/bazaar/plugins/`. There's a few issues with this:

1. It's not necessary to copy anything into `var/bazaar/plugins/`, because bzr knows where its included plugins are.
2. `var/bazaar/plugins/` is not a default path bzr searches in for plugins, so it doesn't actually load anything from there.
3. The files copied into `var/bazaar/plugins/` will not be removed if the user uninstalls bzr.

I assume this change was made so that the bzr-fastimport formula could work, added in af04a73456ad98ad45397e20e4a3de3b60174eb4. However, it won't actually work due to Point `2`. Point `3` also applies. Currently, the test for bzr-fastimport is failing.

@bfontaine: I bet you already had bzr-fastimport installed in `~/.bazaar/plugins`, so the test passed on your system.

### DIFF
--- a/Library/Formula/bzr-fastimport.rb
+++ b/Library/Formula/bzr-fastimport.rb
@@ -3,6 +3,7 @@ class BzrFastimport < Formula
   homepage "https://launchpad.net/bzr-fastimport"
   url "https://launchpad.net/bzr-fastimport/trunk/0.13.0/+download/bzr-fastimport-0.13.0.tar.gz"
   sha256 "5e296dc4ff8e9bf1b6447e81fef41e1217656b43368ee4056a1f024221e009eb"
+  revision 1
 
   bottle do
     cellar :any
@@ -15,8 +16,8 @@ class BzrFastimport < Formula
   depends_on "bazaar"
 
   resource "python-fastimport" do
-    url "https://launchpad.net/python-fastimport/trunk/0.9.0/+download/python-fastimport-0.9.0.tar.gz"
-    sha256 "07d1d3800b1cfaa820b62c5a88c40cc7e32be9b14d9c6d3298721f32df8e1dec"
+    url "https://launchpad.net/python-fastimport/trunk/0.9.2/+download/python-fastimport-0.9.2.tar.gz"
+    sha256 "fd60f1173e64a5da7c5d783f17402f795721b7548ea3a75e29c39d89a60f261e"
   end
 
   def install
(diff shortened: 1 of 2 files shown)


bfontaine:
> I bet you already had bzr-fastimport installed in `~/.bazaar/plugins`, so the test passed on your system.

I don’t have any `~/.bazaar` directory, and the test passed on the bot as well. I’m not a regular Bazaar user, feel free to improve things if you think it’s better :wink:

dlh:
I figured out why it's failing for me. The current bzr-fastimport formula creates the symlink in `install`, when it should be doing it in `post_install`. An installation from source will succeed, but installing the bottle won't.

The tests run by the test-bot succeed, because first it builds from source and runs the test. Then, it uninstalls (which doesn't remove anything from `var`), installs from the bottle and runs the test.

Anyways, I find the usage of `BZR_PLUGIN_PATH` used in this pull request to be much cleaner.

bfontaine:
> The tests run by the test-bot succeed, because first it builds from source and runs the test. Then, it uninstalls (which doesn't remove anything from `var`), installs from the bottle and runs the test.

@mikemcquaid could we do something here?

> Anyways, I find the usage of `BZR_PLUGIN_PATH` used in this pull request to be much cleaner.

Yes, I wasn’t aware of this when I created the PR.

Library/Formula/bzr-fastimport.rb
```diff
@@ -24,32 +25,19 @@ def install
...
-    rm_rf target if target.exist?
-    ln_s libexec, target
+    (share/"bazaar/plugins/fastimport").install Dir["*"]
   end
 
   def caveats; <<-EOS.undent
     In order to use this plugin you must set your PYTHONPATH in your ~/.bashrc:
 
-      export PYTHONPATH="#{libexec}/vendor/lib/python2.7/site-packages:$PYTHONPATH"
+      export PYTHONPATH="#{opt_prefix}/libexec/vendor/lib/python2.7/site-packages:$PYTHONPATH"
```
DomT4: `opt_libexec`

DomT4:
> The files copied into var/bazaar/plugins/ will not be removed if the user uninstalls bzr.

Is this not desired in some ways, for plugins to be preserved across removal/upgrades/etc? I seem to recall we "protect" plugins like that in various formulae, but maybe it's not appropriate here for one reason or another?

mikemcquaid:
> Is this not desired in some ways, for plugins to be preserved across removal/upgrades/etc? I seem to recall we "protect" plugins like that in various formulae, but maybe it's not appropriate here for one reason or another?

It's definitely desired for plugins to be preserved across upgrades :+1:

Library/Formula/bzr-fastimport.rb
xu-cheng: I think this can be done by below code during the install:
```ruby
(lib/"python2.7/site-packages").mkpath
(lib/"python2.7/site-packages/homebrew-bzr-fastimport.pth").write "#{libexec}/vendor/lib/python2.7/site-packages\n"
```

dlh:
Updated: `PYTHONPATH` doesn't need to be set manually now (thanks @xu-cheng).

> Is this not desired in some ways, for plugins to be preserved across removal/upgrades/etc? I seem to recall we "protect" plugins like that in various formulae, but maybe it's not appropriate here for one reason or another?

Sure it's desirable for plugins to be preserved, but these plugins are *included* with bazaar and don't need to be copied to a non-standard place. We wouldn't want to protect the included plugins from being upgraded.

The default directory for a user to install custom bzr plugin to is `~/.bazaar/plugins`.

mikemcquaid:

Issue #42273, opened 2015-07-30
Need better way to handle Luarocks


alerque:
I have a nearly useless package formula and I can't figure out how to fix it because either A) Homebrew doesn't support Lua dependencies properly or B) I just can't figure out how to do it.

When I first submitted the SILE formula in #40893 I asked about how to handle these but didn't get much feedback and the formula was eventually merged with some bits like this:

    def install
        system "luarocks-5.2", "install", "lpeg"
        # ...
    end

Then later in I discovered I needed one more Lua rock and added it in #41288. At that point it was pointed out that the Lua deps were broken because the bottling process wasn't actually catching them. I was instructed to drop those lines in favor or an something like this:

    depends_on "lpeg" => :lua

This is great in theory, but in practice it means nobody that isn't already well versed in Lua or has their machine setup with it can actually install SILE. As reported in simoncozens/sile#143 users without Lua cannot even run the luarocks install command manually. It just doesn't work unless you've gotten it setup before. Because the formula aborts when it doesn't install ALL the dependencies, the user has to install Lua by itself first, then install the luarocks manually, then go on to install SILE. This is a disaster from a usability standpoint.

The biggest hold up seems to be the bottle system. It seems like I can get people a lot further by pointing people at a separate fork rather than the Homebrew version because by skipping the bottling step I can actually install the dependencies.

The current install procedure is a disaster from a usability perspective. So the question becomes:

1. Can I submit a version of this formula that does not get bottled and does install the luarocks?
2. Can the bottler be fixed so it actually includes these parts?
3. Can I create separate formulas for each needed rock and use them as dependencies (this is the way most Linux distros handle this)?
4. A better way?

alerque:
@DomT4, you [mentioned before](https://github.com/Homebrew/homebrew/pull/41288#issuecomment-118508488) that the Lua dep system needed some help. Do you have any input into this? The way things are right now users are not able to get the software working at all. It works okay for people that have previously installed it manually because their systems have the deps, but for fresh systems they just can't make out what needs to be done.

DomT4:
Hey, Thanks for the ping. There's a lot of things in play with this issue that are rubbing up against each other in less than helpful ways.

> you mentioned before that the Lua dep system needed some help.

[...]

PR #42576, opened 2015-08-06
languages: flag missing dependency requirement

DomT4:
I'm aware this is probably more complicated than it needs to be and Xu is likely to see this, drive by and go *"Or you can compress all of this down to these two easy lines"* :stuck_out_tongue_winking_eye:.

This is in part designed to handle situations described in #42273 where we tell someone to install a special dependency, but because we (rightly, IMO) resolve special dependencies first users can end up being told to execute a command on a tool that isn't yet installed and isn't immediately obvious how to install it.

In the situation raised there, with the `sile` formula people are being told to `luarocks install xyz` but we hadn't installed Lua for them first, so they just get a `command not found: luarocks` message. Perhaps it should be obvious enough how to install said tools by looking at the formula's dependencies, but it's not a huge burden on us to make life easier than that.

Essentially, when `lua` isn't already installed this changes the current message to:

```
sile: Unsatisfied dependency: lpeg
Homebrew does not provide special Lua dependencies; install with:
  luarocks-5.2 install lpeg
You may need to:
  brew install lua
Error: Unsatisfied requirements failed this build.
```

If Lua is installed but the special dependency isn't the message stays at:

```
sile: Unsatisfied dependency: lpeg
Homebrew does not provide special Lua dependencies; install with:
  luarocks-5.2 install lpeg
Error: Unsatisfied requirements failed this build.
```

The only thing it won't flag that doesn't already come shipped with OS X is the `rbx` dependency as there's no way to install that via `brew`. It won't flag Perl/Python/Ruby stuff because `/usr/bin/env` returns the exit code `127` if a tool doesn't exist, but the tools themselves return an exit code of `1` or `2` if the tool exists but the special dependency isn't installed.

I need to check a few of the remaining ones, but `127` should be a safe check as it's *supposed* to be reserved for "command not found".

### DIFF
--- a/Library/Homebrew/requirements/language_module_requirement.rb
+++ b/Library/Homebrew/requirements/language_module_requirement.rb
@@ -19,7 +19,8 @@ class LanguageModuleRequirement < Requirement
         `#{command_line} #{@module_name}`
     EOS
 
-    if $?.exitstatus == 127
+    sys_prov = :python || :perl || :ruby
+    unless @language == sys_prov
       s << <<-EOS.undent
 
       You may need to: `brew install #{@language}`
--- a/Library/Homebrew/requirements/language_module_requirement.rb
+++ b/Library/Homebrew/requirements/language_module_requirement.rb
@@ -12,11 +12,22 @@ class LanguageModuleRequirement < Requirement
 
   satisfy(:build_env => false) { quiet_system(*the_test) }
 
-  def message; <<-EOS.undent
-    Unsatisfied dependency: #{@module_name}
-    Homebrew does not provide #{@language.to_s.capitalize} dependencies; install with:
-      #{command_line} #{@module_name}
+  def message
+    s = <<-EOS.undent
+      Unsatisfied dependency: #{@module_name}
+      Homebrew does not provide special #{@language.to_s.capitalize} dependencies; install with:
+        `#{command_line} #{@module_name}`
     EOS
+
+    if $?.exitstatus == 127
+      s << <<-EOS.undent
+
+      You may need to: `brew install #{@language}`
+
+      EOS
+    end
+
+    s
   end
 
   def the_test


Library/Homebrew/requirements/language_module_requirement.rb
```diff
@@ -12,11 +12,26 @@ def initialize(language, module_name, import_name = nil)
...
+      s
+    elsif $?.exitstatus == 127
+      s += <<-EOS.undent
+        You may need to:
+          brew install #{@language}
+      EOS
+    s
+    else
+    s
+    end
```
xu-cheng: How about:
```ruby
...

if @language != :rbx && $?.exitstatus == 127
 s += <<-EOS.undent
    You may need to:
      brew install #{@language}
 EOS
end

s
```

bfontaine: In any case we should use `<<` for string concatenation instead of `+=`. The former is faster and more memory efficient because it doesn’t create a new string.

DomT4: > How about:

:+1:. I didn't know if it would be cleaner/quicker for `rbx` users if we bailed out before checking the exit status, but the existing check is hardly slow so I can't imagine any performance reduction.

> string concatenation instead of +=.

We use the latter in a bunch of formulae and I try to keep the code between core and formulae relatively harmonious, but I'm fine with changing it.

Library/Homebrew/requirements/language_module_requirement.rb
```diff
@@ -12,11 +12,23 @@ def initialize(language, module_name, import_name = nil)
...
   satisfy(:build_env => false) { quiet_system(*the_test) }
 
-  def message; <<-EOS.undent
-    Unsatisfied dependency: #{@module_name}
-    Homebrew does not provide #{@language.to_s.capitalize} dependencies; install with:
-      #{command_line} #{@module_name}
+  def message
+    s = ""
+
+    s << <<-EOS.undent
```
xu-cheng: These two lines can be combined as `s = <<-EOS.undent ...`.

DomT4: Thanks Xu. Should have occurred to me really.

Just trying to work out why this error only gets thrown sometimes and then I'll push again:

```
  1) Error:
LanguageModuleRequirementTests#test_differing_module_and_import_name:
NoMethodError: undefined method `exitstatus' for nil:NilClass
    /usr/local/Library/Homebrew/requirements/language_module_requirement.rb:24:in `message'
    /usr/local/Library/Homebrew/test/test_language_module_requirement.rb:26:in `test_differing_module_and_import_name'
```

bfontaine:
Anything new here?

DomT4:
It's on my post-Christmas list.

DomT4:
Need to fix the test up still, but a mere 150 days later followed up with a new commit to tidy the style up and simplify after Misty's changes to the languages. Need to invent a machine that hits me when I forget PRs are still pending follow-up.

DomT4:
That test passes 90% of the time and fails randomly 10% of the time, and I'm not really sure why at this point. The `exitstatus` shouldn't even come into play there because OS X ships Python and shouldn't return the "missing" exit code, hmm.

DomT4:
Pushed a new conditional which might be a little more robust than exit status checking, but it's not necessarily ideal. Have left it in a new commit so people can compare and discuss as desired.

mistydemeo:
The test that's failing is `test_differing_module_and_import_name`, which tests the method without actually executing the requirement. As a result, `$?` is undefined (but, being a global, is autovivified as `nil`).

mistydemeo:
(As for why it sometimes works, since `$?` is a global, it can still be referenced after a test has passed. This failure will only occur if `test_differing_module_and_import_name` is the first executed test; otherwise, it is actually referencing the exit status from a different test. Since test order is randomized, it only comes up sometimes.)

Library/Homebrew/requirements/language_module_requirement.rb
```diff
@@ -12,11 +12,23 @@ def initialize(language, module_name, import_name = nil)
...
-    Homebrew does not provide #{@language.to_s.capitalize} dependencies; install with:
-      #{command_line} #{@module_name}
+  def message
+    s = <<-EOS.undent
+      Unsatisfied dependency: #{@module_name}
+      Homebrew does not provide special #{@language.to_s.capitalize} dependencies; install with:
+        `#{command_line} #{@module_name}`
     EOS
+
+    sys_prov = :python || :perl || :ruby
```
mistydemeo: This will always be `:python`; maybe something like:

```ruby
unless [:python, :perl, :ruby].include? @language
  s << <<-EOS.undent
  ...
  EOS
end
```

DomT4:
@mistydemeo You are magical. Thank you for the detailed explanation as to why it fails; I'd have been chasing my tail forever on that one. I guess exit code checking is out of the window in terms of reliability in testing then, more or less (?).

Will probably have to play with the idea of declaring certain dependencies system-available and not barking about the need to `brew install` those a little more, although the current implementation here is broken as you highlighted.

DomT4:
Moved to https://github.com/Homebrew/brew/pull/194.

PR #42581, opened 2015-08-06
augeas 1.4.0

dunn:


Library/Formula/augeas.rb
```diff
@@ -12,7 +12,6 @@ class Augeas < Formula
 
   head do
     url "https://github.com/hercules-team/augeas.git"
-
```
DomT4: I'm the most anal person in the world about this, but not having an empty line between the `url` and the first `depends_on` drives me up the wall :smile_cat:. Feel free to ignore me though.

dunn: Ahaha, having the line is what bugs me.

DomT4: Whitespacing makes me feel happy about reading formulae. It feels more readable. I'll go quietly if others prefer no whitespacing though :wink:.

dunn: I'll add it back in don't worry

DomT4: :heart: :heart: :heart:

Issue #40559, opened 2015-06-10
New version of BibTool


adamliter:
`bib-tool.rb` is currently set up to install version 2.57 of [BibTool](http://www.gerd-neugebauer.de/software/TeX/BibTool/index.en.html), but there is a new version available: 2.60.

DomT4:
Would you be willing to [submit a PR](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md) to update it? You only need to change the `url`, update the `sha` (preferably to `sha256`) and remove any `revision` in the main formula.

adamliter:
@DomT4 Yes, I am working on submitting a pull request. Trying to run `brew install bib-tool` with the updated URL and sha256 leads to:

    Error: Failure while executing: tar xf /Library/Caches/Homebrew/bib-tool-2.60.tar.gz

Would this be a problem with the zipped tar file that I should report to the maintainer? If I just download the zipped file myself and try to run `tar -xvf` on it, it will unzip, but it exits with:

    tar: Error exit delayed from previous errors.

I'm not terribly familiar with how `tar` and zipping files work, but my guess is that this is a problem with the archive, right?

DomT4:
Yeah, It's refusing to uncompress here as well:

```
gzip: unknown compression format
gtar: Child returned status 1
gtar: Error is not recoverable: exiting now
```

The contents of the tarball are there, so it isn't just filled with empty data:

```
~> tar -tzf BibTool-2.60.tar.gz
BibTool/README
BibTool/install.tex
BibTool/COPYING
BibTool/Changes.tex
BibTool/makefile.unx
BibTool/makefile.dos
BibTool/makefile.ata
BibTool/makefile.ami
BibTool/ToDo
BibTool/THANKS
BibTool/pxfile.man
BibTool/sbuffer.man
BibTool/Perl/bibtool.pl
BibTool/Tcl/bibtool.tcl
BibTool/main.c
BibTool/crossref.c
BibTool/database.c
BibTool/entry.c
BibTool/error.c
BibTool/expand.c
BibTool/init.c
BibTool/key.c
BibTool/macros.c
BibTool/names.c
BibTool/parse.c
BibTool/print.c
BibTool/pxfile.c
BibTool/record.c
BibTool/rewrite.c
BibTool/rsc.c
BibTool/s_parse.c
BibTool/symbols.c
BibTool/stack.c
[...]

PR #40574, opened 2015-06-11
bib-tool 2.60

adamliter:
Fixes  #40559

PR #42590, opened 2015-08-07
bib-tool 2.61

adamliter:
The problem in the BibTool tarball from version 2.60 is gone in version 2.61, so the workaround discussed in https://github.com/Homebrew/homebrew/issues/40559 is no longer needed.

So I've deleted the special download strategy for the new version of BibTool.

Also, the pull request (https://github.com/Homebrew/homebrew/pull/40574) that I submitted to fix https://github.com/Homebrew/homebrew/issues/40559 introduced a problem when trying to install BibTool on Linux with Linuxbrew, since Linux doesn't use `bsdtar`, and the `--fast-read` option that was introduced as a workaround is only available in `bsdtar`, not `gnutar`. That was shortsighted on my part. Sorry about that.

This was mentioned in https://github.com/Homebrew/linuxbrew/issues/456 and ultimately fixed with a pull request to Linuxbrew (https://github.com/Homebrew/linuxbrew/pull/457).

In other words, the formulas for BibTool v. 2.60 diverged between Homebrew and Linuxbrew. This new formula for v. 2.61 works with both Homebrew and Linuxbrew (tested on Mac OS 10.10.4 and Ubuntu v. 14.04). I'm not sure how formulas are maintained across Homebrew and Linuxbrew, but there is no longer any need for them to diverge, so this new formula can be used in Linuxbrew, too. Should I submit a separate pull request to Homebrew/Linuxbrew? I don't think the commit can be straight-forwardly cherry-picked, since the formula in Linuxbrew is currently different.

Thanks! :smile:

### DIFF
--- a/Library/Formula/bib-tool.rb
+++ b/Library/Formula/bib-tool.rb
@@ -1,20 +1,8 @@
-# This is needed because of a problem with the tarball for 2.60
-# Hopefully, it will not be needed for future releases
-# See https://github.com/Homebrew/homebrew/issues/40559
-class BibToolDownloadStrategy < CurlDownloadStrategy
-  def stage
-    with_system_path { safe_system "tar", "xqf", cached_location, "BibTool/doc/bibtool.tex" }
-    with_system_path { safe_system "tar", "xf", cached_location, "--exclude", "BibTool/doc/bibtool.tex" }
-    chdir
-  end
-end
-
 class BibTool < Formula
   desc "Manipulates BibTeX databases"
   homepage "http://www.gerd-neugebauer.de/software/TeX/BibTool/index.en.html"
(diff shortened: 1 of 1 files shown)


dunn:
About once a week @sjackman merges the changes from Homebrew into Linuxbrew; I don't know if it would be easier for him to reconcile the formulae with a PR.

sjackman:
No pull request is necessary. I'll pick up the change on the next weekly merge.

adamliter:
Sounds good; thanks @sjackman!

DomT4:
:tada: for getting rid of the download strategy.

Merged in https://github.com/Homebrew/homebrew/commit/8253d308de5a6f1d1d91017f85a31de4977bb506. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #42593, opened 2015-08-07
only run post_install if it's defined

xu-cheng:


Library/Homebrew/cmd/postinstall.rb
```diff
@@ -2,7 +2,7 @@
 
 module Homebrew
   def postinstall
-    ARGV.resolved_formulae.each { |f| run_post_install(f) }
+    ARGV.resolved_formulae.select(&:post_install_defined?).each { |f| run_post_install(f) }
```
bfontaine: Shouldn’t that be faster as below?

```rb
ARGV.resolved_formulae.each { |f| run_post_install(f) if f.post_install_defined? }
```

I haven’t done any benchmark but that would be one loop over the array instead of two and the creation of a new array from the selected formulae.
Feel free to discard my comment if you think such micro-optimization is unnecessary here :wink:

xu-cheng: Yes, I think we can discard this as this is hardly a hotspot which requires optimization. But thanks for your comment anyway. :smile:

DomT4:
:+1:

PR #42605, opened 2015-08-07
secp256k1

shazow:
This PR adds a formula for [secp256k1](https://github.com/bitcoin/secp256k1), a cryptographic library commonly used in bitcoin-related tools (since bitcoin uses the secp256k1 ECDSA curve).

Unfortunately the source repository does not provide version tags or anything of that sort. The library has been around for a couple of years ago and is considered stable.

It is commonly known as secp256k1 (rather than libsecp256k1) and provides a `secp256k1.h` header. There are Ruby, Python, and Node bindings that use it under this name.

### DIFF
--- a/Library/Formula/secp256k1.rb
+++ b/Library/Formula/secp256k1.rb
@@ -1,0 +1,18 @@
+class Secp256k1 < Formula
+  desc "Optimized C library for EC operations on curve secp256k1"
+  homepage "https://github.com/bitcoin/secp256k1"
+  url "https://github.com/bitcoin/secp256k1.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+
+  def install
+    system "./autogen.sh"
+    system "./configure", "--disable-debug",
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> Unfortunately the source repository does not provide version tags or anything of that sort. The library has been around for a couple of years ago and is considered stable.

I'm afraid we'll need an upstream tag for this. Can you ask them for one? Thanks!

DomT4:
I'll be surprised if upstream tags anything on this. It changes quite a bit and presumably isn't hugely stable yet. I've been supporting it in a [personal tap](https://github.com/DomT4/homebrew-crypto) for a while with roughly weekly updates to the latest revision.

shazow:
Yea, I opened an issue but looks like they're not ready to commit to a stable API just yet: https://github.com/bitcoin/secp256k1/issues/286

@DomT4 Didn't know about your tap, thanks for sharing!

mikemcquaid:
Ok, passing on this for now. It would be accepted in the `homebrew-head-only` tap.

PR #42626, opened 2015-08-07
hostname 3.15 (new formula)

guoxiao:


Library/Formula/hostname.rb
```diff
@@ -0,0 +1,66 @@
+class Hostname < Formula
+  desc "Utility to set/show the host name or domain name"
+  homepage "https://packages.debian.org/sid/hostname"
+  url "http://ftp.de.debian.org/debian/pool/main/h/hostname/hostname_3.15.tar.gz"
```
DomT4: Please use one of the secure links rather than the plaintext Debian mirrors. Either https://mirrors.kernel.org/debian/pool/main/ or https://mirrors.ocf.berkeley.edu/debian/pool/main/

Library/Formula/hostname.rb
```diff
@@ -0,0 +1,65 @@
...
+  sha256 "b6d10114ed14306b21745d2cac1b9adf7a2546f16b9fd719bec14bd7cec61d20"
+
+  patch :DATA
```
DomT4: Will need submitting upstream somewhere with a link to that, and a brief explanation of what the patch is doing & why added.

mistydemeo:
Since OS X includes a `hostname(1)`, this may belong in [homebrew-dupes](https://github.com/Homebrew/homebrew-dupes).

Library/Formula/hostname.rb
```diff
@@ -0,0 +1,65 @@
+class Hostname < Formula
+  desc "Utility to set/show the host name or domain name"
+  homepage "https://packages.debian.org/sid/hostname"
+  url "https://mirrors.kernel.org/debian/pool/main/h/hostname/hostname_3.15.tar.gz"
+  sha256 "b6d10114ed14306b21745d2cac1b9adf7a2546f16b9fd719bec14bd7cec61d20"
+
+  patch :DATA
```
dunn: Has this patch been submitted to/accepted by the upstream developers?  We try to avoid floating patches in formulae.

guoxiao:
@mistydemeo this `hostname` is a different implementation. It's more like the `coreutils` formula rather than "more recent or bugfix versions".

Library/Formula/hostname.rb
guoxiao: It's maintained by debian team, who may not accept patches to port it to OSX.

mikemcquaid: We will not accept it unless it's been submitted (and probably accepted) upstream.

mikemcquaid:
I agree with @mistydemeo.

guoxiao:
All right. I'll submit it to `homebrew-dupes`

mikemcquaid:
@guoxiao Before you do: please submit the patch upstream. This applies to all formulae in all official taps.

DomT4:
@guoxiao Debian have taken several patches from us before now and accepted them, or provided helpful alternatives where they couldn't. It's definitely worth trying to chase them on that.

PR #42640, opened 2015-08-08
Enable gnome-pty-helper in vte3

oakes:
This makes the vte3 formula behave the way it used to (when vte 0.36 was being used). The current version doesn't enable gnome-pty-helper by default, which means many use-cases for VteTerminal are broken.


mikemcquaid:

For future reference the preferred commit message format is `vte3: enable gnome-pty-helper.`.

PR #42641, opened 2015-08-08
sdl_ttf: add patch for broken font rendering

tomyun:
This fixes [an old issue](https://bugzilla.libsdl.org/show_bug.cgi?id=1433) of `sdl_ttf` 2.0.11 that completely breaks some types (e.g. `TTF_RenderGlyph_Shaded`) of font rendering. It has been fixed in the latest version, 2.0.12, but only released under SDL2 (`sdl2_ttf`). There would be likely no more official release for SDL 1.x, I guess.

### DIFF
--- a/Library/Formula/sdl_ttf.rb
+++ b/Library/Formula/sdl_ttf.rb
@@ -18,8 +18,16 @@ class SdlTtf < Formula
   depends_on "sdl"
   depends_on "freetype"
 
+  # Fix broken TTF_RenderGlyph_Shaded()
+  # https://bugzilla.libsdl.org/show_bug.cgi?id=1433
+  patch do
+    url "https://gist.githubusercontent.com/tomyun/a8d2193b6e18218217c4/raw/8292c48e751c6a9939db89553d01445d801420dd/sdl_ttf-fix-1433.diff"
+    sha256 "4c2e38bb764a23bc48ae917b3abf60afa0dc67f8700e7682901bf9b03c15be5f"
+  end
+
   def install
     ENV.universal_binary if build.universal?
+    inreplace "SDL_ttf.pc.in", "@prefix@", HOMEBREW_PREFIX
 
     system "./configure", "--disable-debug",
                           "--disable-dependency-tracking",


Library/Formula/sdl_ttf.rb
```diff
@@ -18,6 +18,10 @@ class SdlTtf < Formula
   depends_on "sdl"
   depends_on "freetype"
 
+  # Fix broken TTF_RenderGlyph_Shaded()
+  # https://bugzilla.libsdl.org/show_bug.cgi?id=1433
```
mikemcquaid: It says this is fixed in the latest release? If so would be better to update than patch. Thanks!

tomyun: Yes, it's fixed in the latest version, 2.0.12, which is only compatible with SDL2. Homebrew has `sdl2_ttf` for that. The last version released for SDL1 is 2.0.11. The codebase has gone complete transition to SDL2 and it's unlikely to have another release for SDL1.

tomyun:
FYI, this issue has been addressed by other distributions: [Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=661987), [FreeBSD](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=196655), and [pkgsrc](http://pkgsrc.se/files.php?messageId=20120520100619.96EA9175DD@cvs.netbsd.org).

tomyun:
Also added missing `inreplace` block for `pkg-config`. Many other SDL-related packages (e.g. `sdl_image`, `sdl_mixer`) already have one. Not sure if it deserves a separate PR though.

Library/Formula/sdl_ttf.rb
```diff
@@ -18,8 +18,13 @@ class SdlTtf < Formula
...
+  patch :DATA
+
   def install
     ENV.universal_binary if build.universal?
+    inreplace "SDL_ttf.pc.in", "@prefix@", HOMEBREW_PREFIX
```
DomT4: Shouldn't be `HOMEBREW_PREFIX` if we have to set this. Should be just `prefix`.

(We could also probably use an `inreplace` for the `SDL_ttf.c` change as well)

tomyun: I'm also not comfortable with `HOMEBREW_PREFIX` either, but there seems to be a historical reason for this: d37ab41. I just wanted to make it consistent with other SDL packages. Do you think we better leave it as a separate PR then?

And it wouldn't be easy to patch only this specific location with `inreplace`. The exact same `for` loop statement (with legit use of both `pixmap` and `bitmap`) appears in another part of the code multiple times.

DomT4: Oh, ugh. I forgot about SDL's prefix fun. We may well be stuck with `HOMEBREW_PREFIX` then if we absolutely have to inreplace the path. It is indeed the same as we apply for `SDL2`.

> And it wouldn't be easy to patch only this specific location with inreplace.

Alright, fair enough. It might be nicer to `gist` it or something and then apply it with `patch do`. Inline patches tend to get sloppy where people accidentally add or remove whitespacing from them on other changes and don't notice.

tomyun: I'm also in favor of cleaner formula anyways. Okay, now the diff is hosted from [my gist](https://gist.github.com/tomyun/a8d2193b6e18218217c4).

Library/Formula/sdl_ttf.rb
DomT4: How do you feel about this @mikemcquaid? It's likely to never ever be fixed upstream sadly. I guess we either carry the patch, keep the slightly broken formula, or remove the formula entirely.

mistydemeo: Backporting the patch to the original `sdl_ttf` is fine, I think - it is an official patch even if it's not in this branch of the codebase. Just to make sure, this is the exact same patch as in the sdl2_ttf version?

We have a number of packages using `sdl_ttf`, so I'd rather not remove it entirely.

tomyun: `TTF_RenderGlyph_Shaded()` has been refactored and reimplemented in `sdl2_ttf` so this patch is no longer relevant there. It seems `sdl2_ttf` has diverged quite significantly since the last release of `sdl_ttf` despite 0.0.1 difference in version.

mikemcquaid: If it's literally never going to be fixed upstream I'm a bit apprehensive about carrying it. On the flip side if we're never gonna see any new versions then: meh.

DomT4:
Taking the other maintainer's comments into account, will pull this. It does seem almost certain that we'll never see another release to fix the issue sadly.

Merged in https://github.com/Homebrew/homebrew/commit/1244b1aa4d359ce8bcbbc39f26030bdc2f6c65d9. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #42648, opened 2015-08-08
ntopng: add head

Drewshg312:
please add the latest version (head would be great). Need it because of this issue: https://github.com/ntop/ntopng/issues/84


xu-cheng:

PR #42661, opened 2015-08-08
lldpd 0.7.16

vincentbernat:
Update lldpd to 0.7.16.

Also push some unrelated changes that would also fit 0.7.15:

 1. Move configuration files from `#{prefix}/etc` to `#{etc}`. Also, the local state dir, where the daemon is putting its socket, is moved from `#{prefix}/var` to `#{var}`. I am unsure if we can do that as is. Should there be a warning for configuration files? Users are very unlikely to have any configuration file.

 2. SNMP and JSON support are now autodetected. Ensure we disable them if the user didn't request them.

### DIFF
--- a/Library/Formula/lldpd.rb
+++ b/Library/Formula/lldpd.rb
@@ -1,8 +1,8 @@
 class Lldpd < Formula
   desc "Implementation library for LLDP"
   homepage "https://vincentbernat.github.io/lldpd/"
-  url "http://media.luffy.cx/files/lldpd/lldpd-0.7.15.tar.gz"
-  sha256 "c891d6d4480a6a890561ac43d8cc923bd027deb82a3999d65f37d96ca368c246"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

DomT4:
I forgot to squash here on merge, so you've ended up with two Homebrew commits for one PR :wink:. If you could squash when the PR is ready for review in future that'd be great, Thanks :).

Merged in https://github.com/Homebrew/homebrew/commit/5aeded4bdeb227a6faf3f5b215627335c08a8cbf. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

mikemcquaid:
> I forgot to squash here on merge, so you've ended up with two Homebrew commits for one PR . If you could squash when the PR is ready for review in future that'd be great, Thanks :).

@DomT4 In this case where it's two specific different changes I think it's OK as-is :+1:

PR #42663, opened 2015-08-08
Update osm2pgsql to 0.88.0

pnorman:
0.88.0 is the long-term version of 0.87.0, which was packaged before.

This version fixes the bugs which required the patch, so that can be removed

Both Protobuf and Lua support are strongly recommended, and protobuf is set to become mandatory next version. One of the more common questions people have with brew installed osm2pgsql is why they can't import protobuf OSM files.

Lua is used less, but is not a heavy dependency, and without it many new styles cannot run.

### DIFF
--- a/Library/Formula/osm2pgsql.rb
+++ b/Library/Formula/osm2pgsql.rb
@@ -1,9 +1,8 @@
 class Osm2pgsql < Formula
-  desc "Convert OpenStreetMap data to postGIS-enabled PostgreSQL db"
+  desc "OpenStreetMap data to PostgreSQL converter"
   homepage "https://wiki.openstreetmap.org/wiki/Osm2pgsql"
-  url "https://github.com/openstreetmap/osm2pgsql/archive/0.87.0.tar.gz"
-  sha256 "6f5538d098e17a578a384f70f61b335f10edb04114d5cdc5233ca97f868ad0f6"
-  revision 1
+  url "https://github.com/openstreetmap/osm2pgsql/archive/0.88.0.tar.gz"
+  sha256 "45a6768e680a50c416fdae72cf4b26091644947a7ead44c8b7484f2276b2f119"
 
   bottle do
     sha256 "6b1bc9ca723e3c8bf23c9bb5704fa99205a1037085d408102050da03ac703a5c" => :yosemite
@@ -18,14 +17,14 @@ class Osm2pgsql < Formula
   depends_on "boost"
   depends_on "geos"
   depends_on "proj"
-  depends_on "protobuf-c" => :optional
+  depends_on "protobuf-c" => :recommended
+  depends_on "lua" => :recommended
 
-  # Fixes an upstream issue:
-  #   https://github.com/openstreetmap/osm2pgsql/issues/196
-  # Remove the patch when upgrading
+  # Apply a fix from the 0.88.x branch.
+  # This will be in 0.88.1, so won't be needed when that's tagged
   patch do
-    url "https://github.com/openstreetmap/osm2pgsql/commit/943684a9b86bee46d245970b3e5870f83afc9208.diff"
-    sha256 "608b702d79fd53f2af98e9651634377126778ff91087cf35a81b65150d8cd963"
+    url "https://github.com/openstreetmap/osm2pgsql/commit/cb449abc73e5ef076f1f8b6535b7d9641a1237f7.diff"
+    sha256 "e0175a3b1de82b67d45adbd50b8194c16673e147401f25dcc22c1c0c8d44aec6"
   end
 
   def install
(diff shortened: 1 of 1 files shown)


Library/Formula/osm2pgsql.rb
```diff
@@ -1,15 +1,8 @@
 class Osm2pgsql < Formula
-  desc "Convert OpenStreetMap data to postGIS-enabled PostgreSQL db"
+  desc "OpenStreetMap data to PostgreSQL converter"
```
pnorman: wording from Github

Library/Formula/osm2pgsql.rb
```diff
@@ -18,15 +11,8 @@ class Osm2pgsql < Formula
...
-
-  # Fixes an upstream issue:
-  #   https://github.com/openstreetmap/osm2pgsql/issues/196
-  # Remove the patch when upgrading
-  patch do
-    url "https://github.com/openstreetmap/osm2pgsql/commit/943684a9b86bee46d245970b3e5870f83afc9208.diff"
-    sha256 "608b702d79fd53f2af98e9651634377126778ff91087cf35a81b65150d8cd963"
-  end
+  depends_on "protobuf-c" => :recommended
+  depends_on "lua" => :recommended
```
pnorman: I'm counting on autoconf to pick up the lua location, is there a better way?

dunn:
Looks like C++11-related errors on Mountain Lion; try `needs :cxx11 if MacOS.version < :mavericks` above the install block, and `ENV.cxx11 if MacOS.version < :mavericks` below `def install`.

pnorman:
This version doesn't require C++11 but does make use of it if available. I'm trying with those changes, but wonder if Boost is involved

pnorman:
mountain lion is now failing with `ERROR: null`, and doesn't seem to be building.

mikemcquaid:
@pnorman Restarted the build.

DomT4:
I don't know what our policy is on `needs :cxx11 if MacOS.version` and the `ENV` equivalent but as far as I know 10.9 and above will always have a `cxx11` capable compiler, so it should be harmless to specify it unconditionally? (Question intended for other maintainers here rather than suggestion for OP).

pnorman:
> know 10.9 and above will always have a cxx11 capable compiler, so it should be harmless to specify it unconditionally?

It'd be easier to specify it unconditionally - but 0.88.0 shouldn't even need C++11. Trying to diagnose remotely is difficult, but my best guess is `./configure` is detecting a C++11 environment when it either doesn't have one, or boost doesn't have C++11 support.

There is a flag to not try to detect C++11, but I don't want to unconditionally enable it, as simply compiling in C++11 mode is a performance boost.

This is all fixed in the 0.89.x branch, where C++11 is required unconditionally and we're starting to modernize old code, but this means there's not going to be any C++11 detection fixes that could be backported.

pnorman:
> @pnorman Restarted the build.

It's giving the error `stderr: error: could not lock config file .git/config: No space left on device`

mikemcquaid:
@pnorman And again :wink:

pnorman:
> @pnorman And again

And still no space again.

pnorman:
It tried to build, but [ran into linking errors](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/31347/version=mountain_lion/console). I'm trying ``--without-cxx11`` for < :mavericks.

Is there a way to print diagnostic information? In Travis builds I [print out versions of key libraries](https://github.com/openstreetmap/osm2pgsql/blob/5aa9a7569132f865fe1bb6d84b9272109d71f6e5/.travis.yml#L41-L45).

Even if ``--without-cxx11`` works, its useful information to have, as some of the libraries do change between versions

Library/Formula/osm2pgsql.rb
```diff
@@ -1,15 +1,8 @@
...
   homepage "https://wiki.openstreetmap.org/wiki/Osm2pgsql"
-  url "https://github.com/openstreetmap/osm2pgsql/archive/0.87.0.tar.gz"
-  sha256 "6f5538d098e17a578a384f70f61b335f10edb04114d5cdc5233ca97f868ad0f6"
-  revision 1
-
-  bottle do
-    sha256 "6b1bc9ca723e3c8bf23c9bb5704fa99205a1037085d408102050da03ac703a5c" => :yosemite
-    sha256 "2835cacfecdbca3f00255cfc5b95107d54fa640e1eb1ca728520f4c252da983f" => :mavericks
-    sha256 "63c279201caf3f6c0359dbe27eb1c2e2a144cdf0cfd130ef6d218b95a9d009fe" => :mountain_lion
-  end
```
mikemcquaid: Please leave the bottle block as-is.

Library/Formula/osm2pgsql.rb
```diff
@@ -35,14 +28,18 @@ def install
...
     puts args
     if build.with? "protobuf-c"
       args << "--with-protobuf-c=#{Formula["protobuf-c"].opt_prefix}"
     end
+    # Mountain Lion has some problems with C++11.
+    # This is probably going to be a fatal issue for 0.89 and 0.90 on it, but
+    # for now it can be worked around
+    if MacOS.version < :mavericks
+      args << "--without-cxx11"
+    end
```
mikemcquaid: Another option is adding `needs :cxx11` which will ensure it's using a CXX11 compiler.

mikemcquaid: Also, this could be `args << "--without-cxx11" if MacOS.version < :mavericks`

pnorman: > Another option is adding needs :cxx11 which will ensure it's using a CXX11 compiler.

This was generating linker errors beyond my ability to diagnose through Jenkins alone

mikemcquaid: Ok, that's fine, just leave it as `args << "--without-cxx11" if MacOS.version < :mavericks`

pnorman:
I think all the issues with the PR are resolved, but don't really have a way to check for sure.

dunn:
Thanks for all your work here! Merged in b828d9ded7fdd46d6a602cefbfd570297735fe67.

PR #42664, opened 2015-08-09
capnp: revert cmake build

iml:
Mea culpa. The new CMake-based build causes problems in dependent formula on Linuxbrew, though it builds successfully there and everything seems to work fine on Mac. Bumping revision to err on the side of caution.

Fixes https://github.com/Homebrew/homebrew-science/issues/2633.


DomT4:
If it's working for OS X I suspect it's going to be seen as something Linuxbrew should revert for themselves, rather than Homebrew.

iml:
This indicates the CMake support is half-baked, though, and I'd rather not rely on it if there are inconsistencies with the builds. Even if OS X is asymptomatic. The revision is probably better suited for Linuxbrew, however.

mikemcquaid:
I'd rather we leave it as-is unless it's causing OS X problems, thanks.

iml:
I'll forward this to Linuxbrew.

PR #42676, opened 2015-08-09
cd-discid 1.4

dunn:
~~The upstream repository hasn't been touched in 2 years, so I doubt my
patch of the Makefile will be accepted.~~  So I didn't use it.


mikemcquaid:
> The upstream repository hasn't been touched in 2 years, so I doubt my
> patch of the Makefile will be accepted. So I didn't use it.

It is worth noting that sometimes these things aren't touched because no-one submits patches. That's not to say you should/shouldn't but just an FYI.

dunn:
Oh yeah, I submitted a PR.  I meant I was originally going to use my PR as a patch, but it's just fixing `make install` so hardly necessary.

PR #42695, opened 2015-08-09
harfbuzz: make cairo dependency optional

khaledhosny:
Cairo is only needed for optional command line utilities that are useful
for testing fonts, but the library itself does not need it.


dunn:
If that includes `hb-shape`, you'll have to write a different test that uses the library instead.

khaledhosny:
`hb-shape` does not require Cairo, only `hb-view` does.

dunn:
Oh, nice!

dunn:
Merged in bea83f97e5d0dd36579cd6e40f10493fa9180b5e, thank you!

PR #42697, opened 2015-08-09
subversion 1.9.0

brainy:


DomT4:
See #42532. Thanks for the submission though!

PR #42702, opened 2015-08-09
rust 1.2.0

zorgiepoo:
Hey all. This attempts to update rust to 1.2. I've never done this before so I've tried to do this to best of my ability. Rust 1.2 was announced on August 6, 2015, so I chose that day for the nightly/resource URLs (don't really know what those are used for). For cargo, I put it at the latest commit on August 9th since the date for that seemed to be more arbitrary to me (latest commit is just a doc change anyway). Tests ran and all.

P.S. I had to add the remote when pushing with the SSH URL (instead of https) because I've 2 factor authentication on, and didn't want to deal with OAuth.

### DIFF
--- a/Library/Formula/rust.rb
+++ b/Library/Formula/rust.rb
@@ -3,23 +3,23 @@ class Rust < Formula
   homepage "https://www.rust-lang.org/"
 
   stable do
-    url "https://static.rust-lang.org/dist/rustc-1.1.0-src.tar.gz"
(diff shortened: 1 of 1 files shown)


zorgiepoo:
Oops. Didn't realize somebody already made a submission for this. I'll just leave it on that issue then. Should have checked before hand :|.

DomT4:
Yeah, we're hoping the issue in the other PR is resolved soon. I don't believe we've heard back from that submitter yet as to whether the problem is known upstream. Thanks for the submission though, We :heart: the enthusiasm!

PR #42703, opened 2015-08-09
dhcping, dnstracer: add mirrors

### LINKED ISSUE #42475: dhcpdump site is offline
The URL `http://www.mavetju.org/download/dhcpdump-1.8.tar.gz` for dhcpdump is not responding, and the entire `www.mavetju.org` site appears to be offline, for at least a few days now. Connections with either `wget` or a browser are failing.

Formulae affected (determined by `grep -li mavetju /usr/local/Library/Formula`):
* `dhcpdump`
* `dhcping`
* `dnstracer`

A brief search turned up no alternate project pages. And the SourceForge files for the project only go up to dhcpdump 1.5, while we're on 1.8.

The Internet Archive actually has a copy of its download pages, including the dhcpdump 1.8 tarball our formula needs. http://web.archive.org/web/20130210091931/http://www.mavetju.org/download/ Is it appropriate to download from there?

I've sent the site owner an email using the [Contacts page found in the Internet Archive](http://web.archive.org/web/20141008024157/http://www.mavetju.org/contacts.php).

apjanke:
Fixes  #42475. (#42645 added a mirror for dhcpdump, but there are two other formulae hosted on the same site.)

The mavetju.org site is back up now, but since it's a small site and has a history of going offline, may as well preemptively add the mirrors for a major host.

### DIFF
--- a/Library/Formula/dhcping.rb
+++ b/Library/Formula/dhcping.rb
@@ -2,6 +2,7 @@ class Dhcping < Formula
   desc "Perform a dhcp-request to check whether a dhcp-server is running"
   homepage "http://www.mavetju.org/unix/general.php"
   url "http://www.mavetju.org/download/dhcping-1.2.tar.gz"
+  mirror "https://mirrors.kernel.org/debian/pool/main/d/dhcping/dhcping_1.2.orig.tar.gz"
   sha256 "32ef86959b0bdce4b33d4b2b216eee7148f7de7037ced81b2116210bc7d3646a"
 
   bottle do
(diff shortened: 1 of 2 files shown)


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/5c82542594e521cade5252334dcaa9ffe2e8321c, Thanks @apjanke!

Just a tiny tiny nit, but we tend to prefer one commit per formula changed. Just makes tracking changes a bit easier down the line.

apjanke:
Sure; I'll watch for that next time.

PR #42707, opened 2015-08-09
chicken 4.10.0

dunn:


mikemcquaid:
Clang barfing before 10.10; might be too old and need GCC?

dunn:
Just a parallel issue, apparently!

PR #42714, opened 2015-08-09
sdl_mixer: add more optional deps

tomyun:
Added `fluid-synth` and `smpeg` as optional dependencies.

PR #42715, opened 2015-08-09
sdl2_mixer: add more optional deps

tomyun:
Added `fluid-synth` and `smpeg2` (#21885) as optional dependencies.

Almost same as #42714 for SDL1, but had to export `SMPEG_CONFIG` to make `smpeg2-config` picked up.

### DIFF
--- a/Library/Formula/sdl2_mixer.rb
+++ b/Library/Formula/sdl2_mixer.rb
@@ -6,17 +6,23 @@ class Sdl2Mixer < Formula
 
   head "http://hg.libsdl.org/SDL_mixer", :using => :hg
 
+  option :universal
+
   depends_on "pkg-config" => :build
   depends_on "sdl2"
   depends_on "flac" => :optional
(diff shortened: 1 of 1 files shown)


Library/Formula/sdl2_mixer.rb
```diff
@@ -6,13 +6,19 @@ class Sdl2Mixer < Formula
...
   depends_on "pkg-config" => :build
   depends_on "sdl2"
   depends_on "flac" => :optional
+  depends_on "fluid-synth" => :optional
+  depends_on "smpeg2" => :optional
   depends_on "libmikmod" => :optional
   depends_on "libvorbis" => :optional
 
-  option :universal
+  if build.with? "smpeg2"
```
mistydemeo: This should go inside the install method.

tomyun: Updated. Thanks!

mistydemeo:
Thanks!

PR #42717, opened 2015-08-10
Fix python lib finding on OS X 10.10.4

nguillermin:
While trying to install gnuradio I had some issues with python crashing ("PyThreadState_Get: no current thread Abort trap: 6"-type fatal errors) which I narrowed down to swig using the system library when everything else was linked against my homebrew python. This patch fixes that.

### DIFF
--- a/Library/Formula/swig.rb
+++ b/Library/Formula/swig.rb
@@ -12,6 +12,8 @@ class Swig < Formula
 
   option :universal
 
+  patch :DATA
+
   depends_on "pcre"
 
   def install
@@ -46,3 +48,17 @@ class Swig < Formula
     assert_equal "2", shell_output("ruby run.rb").strip
   end
 end
+__END__
+diff --git a/configure b/configure
+index 8c2c463..eb036e6 100755
+--- a/configure
++++ b/configure
+@@ -7663,7 +7663,7 @@ $as_echo_n "checking for Python lib dir... " >&6; }
+       PYLIBDIR=`($PYTHON -c "import sys; sys.stdout.write(sys.lib)") 2>/dev/null`
+       if test -z "$PYLIBDIR"; then
+         # Fedora patch Python to add sys.lib, for other distros we assume "lib".
+-        PYLIBDIR="lib"
++        PYLIBDIR="$PYPREFIX/lib"
+       fi
+       { $as_echo "$as_me:${as_lineno-$LINENO}: result: $PYLIBDIR" >&5
+ $as_echo "$PYLIBDIR" >&6; }


DomT4:
Everything should be linked against the system Python, per Homebrew's policy. The only things that should be linked against Homebrew's Python are things which are `depends_on :python` or `depends_on :python => :optional`.

nguillermin:
The gnuradio formula I found (metacollin/homebrew-gnuradio) depends_on :python. Does this mean this commit won't be accepted? Is it a better idea to introduce a patch in the gnuradio formula?

DomT4:
@tdsmith is the Python genius, I'll let him comment, but it seems unlikely that we'll be happy to accept this. The `gnuradio` formula may have less troubles if it switches to `depends_on :python if MacOS.version <= :snow_leopard` but that would need investigating.

tdsmith:
I'm surprised this has an effect and I'd like to understand why. Specifically what does this influence? What happens without this patch, and what does this patch change? Which files are affected, and why?

nguillermin:
Hi @tdsmith, and sorry, I went back and checked and that's not the only thing that I had to change. I also had to `export PYTHON=/usr/local/bin/python` in order to get it to compile against homebrew python.

In any case, prior to patch, and after exporting python environment variable, the ./configure script will give you

    checking for Python os.name... posix
    checking for Python version... python2.7
    checking for Python lib dir... lib
    checking for Python header files... -I/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/include/python2.7 -I/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/config
    checking for Python library directory...
    /usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/config
    checking for Python library... -lpython2.7

With patch you get

    checking for Python os.name... posix
    checking for Python version... python2.7
    checking for Python lib dir... /usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib
    checking for Python header files... -I/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/include/python2.7 -I/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7//usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/config
    checking for Python library directory... Not found
    checking for Python library... -lpython2.7

It took me until right now to notice that this might actually just trade off autoconf's entries for 'lib dir' and 'library directory'...

Apologies for the confusion, this is the first PR I've ever made.

Library/Formula/swig.rb
```diff
@@ -12,6 +12,8 @@ class Swig < Formula
 
   option :universal
 
+  patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

DomT4:
Closing as dead. Please do feel free to reopen if you wish to with the comments addressed. Thank you for your submission to Homebrew :heart:

PR #42719, opened 2015-08-10
PythonRequirement: don't blindly use first Python in PATH

### LINKED ISSUE #42461: homebrew attempts to use system python rather than the version it just installed when building cmake
Archives-Computer:~ charlott$ brew install mysql55
    ==> Installing mysql55 from homebrew/homebrew-versions
    ==> Installing dependencies for homebrew/versions/mysql55: cmake, pidof
    ==> Installing homebrew/versions/mysql55 dependency: cmake
    ==> Downloading http://www.cmake.org/files/v3.3/cmake-3.3.0.tar.gz
    Already downloaded: /Library/Caches/Homebrew/cmake-3.3.0.tar.gz
    ==> Downloading https://pypi.python.org/packages/source/d/docutils/docutils-0.12
    ==> python -c import setuptools... --no-user-cfg install --prefix=/private/tmp/c
       or: -c --help [cmd1 cmd2 ...]
       or: -c --help-commands
       or: -c cmd --help
    
    error: option --no-user-cfg not recognized
    
    READ THIS: https://git.io/brew-troubleshooting
    
I can fix this myself, but pretty sure it's an issue on 10.6 when system Python is 2.6 and doesn't recognize the option, unless I'm totally mis-diagnosing this issue.

mistydemeo:
PythonRequirement currently appends the PATH to the first `python` found in the PATH, but it doesn't check the version. This means that if PythonRequirement triggers a default formula installation, but that new Python isn't first in the PATH, Homebrew may proceed with an incompatible Python.

Fixes #42461.

cc @tdsmith for sanity checking.

### DIFF
--- a/Library/Homebrew/requirements/python_requirement.rb
+++ b/Library/Homebrew/requirements/python_requirement.rb
@@ -19,13 +19,17 @@ class PythonRequirement < Requirement
   end
 
   env do
-    if system_python?
-      if python_binary == "python"
-        version = python_short_version
-        ENV["PYTHONPATH"] = "#{HOMEBREW_PREFIX}/lib/python#{version}/site-packages"
-      end
-    elsif which_python
+    short_version = python_short_version
+
+    if !system_python? && short_version == Version.new("2.7")
       ENV.prepend_path "PATH", which_python.dirname
+    # Homebrew Python should take precedence over older Pythons in the PATH
+    elsif short_version != Version.new("2.7")
+      ENV.prepend_path "PATH", Formula["python"].opt_bin
+    end
+
+    if python_binary == "python"
+      ENV["PYTHONPATH"] = "#{HOMEBREW_PREFIX}/lib/python#{version}/site-packages"
     end
   end
 


tdsmith:
I think this needs to be checked if `system_python?` is true at line 22, as well. I guess this never worked?

mistydemeo:
Hmmm, right.

No, I'm pretty certain this has always been broken!

mistydemeo:
Tweaked - it unconditionally sets `PYTHONPATH` now (though that might not be a good idea? thoughts?), and I adjusted the conditional PATH setting.

tdsmith:
tl;dr: I think that's a good idea.

I think the reason that it is useful to set PYTHONPATH is that a formula with a :python dependency may depend_on "something" => "with-python", and a non-Homebrew python will need help finding it. (If they have followed the advice given in caveats and by `brew doctor`, their python will know how to find our packages, but the user won't have necessarily had a chance to do that.)

The really headache-inducing disadvantage of ever setting PYTHONPATH is that both python2 and python3 obey it and there is no way to limit its scope to a particular Python version, which means the interpreter can get incompatible packages, leading to crashes. The badness is mitigated if formulas use Language::Python.each_python, which resets PYTHONPATH correctly for each python version.

I think it's acceptable to declare/already the case that formulas that use both python2 and python3 must use each_python, while formulas that only use python2 probably don't, so I think this is okay.

tdsmith:
lgtm!

Library/Homebrew/requirements/python_requirement.rb
```diff
@@ -19,13 +19,17 @@ def pour_bottle?
   end
 
   env do
-    if system_python?
-      if python_binary == "python"
-        version = python_short_version
-        ENV["PYTHONPATH"] = "#{HOMEBREW_PREFIX}/lib/python#{version}/site-packages"
-      end
-    elsif which_python
+    version = python_short_version
```
mikemcquaid: can we do `short_version` just to make it a big more readable below?

mikemcquaid:
:+1:

PR #42724, opened 2015-08-10
clojurescript 1.7.48

dunn:
Happily, a version-shema change that doesn't cause problems for us.


mikemcquaid:

PR #42738, opened 2015-08-10
Libquantum

hanwentao:
Add a formula for libquantum, which is a quantum computing simulator library.

### DIFF
--- a/Library/Formula/libquantum.rb
+++ b/Library/Formula/libquantum.rb
@@ -4,6 +4,11 @@ class Libquantum < Formula
   url "http://www.libquantum.de/files/libquantum-1.0.0.tar.gz"
   sha256 "b0f1a5ec9768457ac9835bd52c3017d279ac99cc0dffe6ce2adf8ac762997b2c"
 
+  devel do
+    url "http://www.libquantum.de/files/libquantum-1.1.1.tar.gz"
+    sha256 "d8e3c4407076558f87640f1e618501ec85bc5f4c5a84db4117ceaec7105046e5"
+  end
+
   def install
     system "./configure", "--prefix=#{prefix}"
     system "make", "install"
--- a/Library/Formula/libquantum.rb
+++ b/Library/Formula/libquantum.rb
@@ -1,13 +1,17 @@
 class Libquantum < Formula
-  desc "libquantum is a C library for the simulation of quantum mechanics, with a special focus laid to quantum computing."
+  desc "C library for the simulation of quantum mechanics"
   homepage "http://www.libquantum.de/"
   url "http://www.libquantum.de/files/libquantum-1.0.0.tar.gz"
-  version "1.0.0"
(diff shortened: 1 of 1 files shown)


Library/Formula/libquantum.rb
```diff
@@ -0,0 +1,17 @@
...
+  sha256 "b0f1a5ec9768457ac9835bd52c3017d279ac99cc0dffe6ce2adf8ac762997b2c"
+
+  def install
+    system "./configure", "--prefix=#{prefix}"
+    system "make", "install"
+    system "make", "quobtools_install"
+  end
+
+  test do
+    # No test
```
dunn: Could you take a stab at adding a test? We like to have something that checks basic functionality to make sure the software was installed correctly.

hanwentao:
Test added

hanwentao:
Can this be merged now?

bfontaine:
@hanwentao We need to wait for the bot.

Library/Formula/libquantum.rb
```diff
@@ -0,0 +1,35 @@
+class Libquantum < Formula
+  desc "C library for the simulation of quantum mechanics"
+  homepage "http://www.libquantum.de/"
+  url "http://www.libquantum.de/files/libquantum-1.0.0.tar.gz"
```
bfontaine: Any reason to use the `1.0.0` release instead of the `1.1.1`; which has been released in 2013 according to the homepage?

Library/Formula/libquantum.rb
```diff
@@ -0,0 +1,35 @@
...
+        srand(time(0));
+        reg = quantum_new_qureg(0, 1);
+        quantum_hadamard(0, &reg);
+        result = quantum_bmeasure(0, &reg);
+        printf("The Quantum RNG returned %i!\\n", result);
+        return 0;
+      }
+    EOS
+    system ENV.cc, "-O3", "-o", "qtest", "qtest.c", "-lquantum"
+    system testpath/"qtest"
```
bfontaine: You can use `"./qtest"` here.

Library/Formula/libquantum.rb
hanwentao: 1.0.0 is the stable release, while 1.1.1 is the development release.  Refer to http://www.libquantum.de/downloads

bfontaine: Thanks, could you add a short comment above the URL to explain that? This will prevent people from submitting PRs to update the formula to 1.1.1 (see [here for example](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/globus-toolkit.rb#L4)).

hanwentao:
I improved it according to your feedback.

Library/Formula/libquantum.rb
hanwentao: Make 1.1.1 as a devel version, so keep the 2 versions simultaneously.

mikemcquaid: :+1:

mikemcquaid:

For future reference the preferred commit message format for new formulae is `libquantum 1.0.0 (new formula)`.

PR #42743, opened 2015-08-10
Bug fix for `to_rack`

xu-cheng:
cc @mikemcquaid  @vladshablinsky


vladshablinsky:
I tested this and decided that it's okay. We just need to use fully-qualified formula for taps and everything else works fine. Am I wrong?

xu-cheng:
I think the logic here is a little messy and overcomplicated. So I re-implement it, please check again.

vladshablinsky:
Works fine. :+1:

mikemcquaid:
:+1:

PR #42755, opened 2015-08-10
poppler 0.35.0

edporras:


DomT4:
Will need to ensure this is compatible with everything that depends on it. Poppler tends to break stuff. @bfontaine handled it last time, you can see the stuff to check in #37280.

edporras:
Checked dependencies, all using `--build-from-source`:

- [x] diff-pdf
- [x] pdf2htmlex
- [x] pdf2svg
- [x] pdfgrep
- [x] pdftoipe (using `--HEAD homebrew/head-only/pdftoipe`)
- [x] xournal (using `--with-poppler` option)
- [x] inkscape
- [x] pdf-tools

dunn:
We also check if already-installed versions don't break when poppler is upgraded.  It looks like diff-pdf and pdf2htmlex do break.

mikemcquaid:
Which means they need a `revision` line added.

edporras:
I'm sorry for the delay regarding this. So, looking at both `diff-pdf` and `pdf2htmlEX`, the break because they can't find `libpoppler.53.dylib`. I then found this [issue](https://github.com/Homebrew/homebrew/issues/12190) and, if I understand correct, all I need to do this just bump the version to force a package update?

dunn:
I checked the hombrew-x11 formula and they don't need revisions.

Thanks for contributing! Merged in c4dc3b4f360f7b8cca9ca660dbfcf71b7895dd86 🐝

PR #42754, opened 2015-08-10
libfixbuf 1.7.0

mpertierra:

PR #42756, opened 2015-08-10
yaf 2.7.1

mpertierra:
Depends on #42754


DomT4:
By "Depends on" do you mean if we merge this built against the existing version of `libfixbuf` prior to the recent merge of that PR it'll break?

mpertierra:
@DomT4 No, I was wrong. YAF 2.x requires libfixbuf 1.0.0 or later, so there should be no problem.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/4c76aaa73788ff6029830a2508436b538257ea59. Thank you for your contribution once more! :smiley_cat:

PR #42761, opened 2015-08-10
audit: enforce https on [*.]archive.org

vszakats:


Library/Homebrew/cmd/audit.rb
```diff
@@ -439,7 +439,8 @@ def audit_homepage
          %r{^http://[^/.]+\.tools\.ietf\.org},
          %r{^http://www\.gnu\.org/},
          %r{^http://code\.google\.com/},
-         %r{^http://bitbucket\.org/}
+         %r{^http://bitbucket\.org/},
+         %r{^http://([^/]*\.|)archive\.org}
```
bfontaine: We don’t need to capture the group, so you can use `(?:[^/]*\.)?`.

vszakats: Nice, thanks!

vszakats: Updated for `bintray.com` as well. (where I copied this regexp from)

mikemcquaid:
Thanks @vszakats!

PR #36654, opened 2015-02-08
Add ibex formula

Moisan:
Create a new formula for the ibex library. It includes a java option to create a binding for the [Choco solver](http://choco-solver.org/).

PR #42769, opened 2015-08-10
Fix the java dependancy and upgrade the version

Moisan:
Follow-up from [issue 36654](https://github.com/Homebrew/homebrew/pull/36654). Sorry for the very long delay. I had to create a new issue since I had to delete and recreate my fork of Homebrew due to merging issues.

### DIFF
--- a/Library/Formula/ibex.rb
+++ b/Library/Formula/ibex.rb
@@ -5,6 +5,7 @@ class Ibex < Formula
   sha256 "d92ff32f14d27ad7b390ae693beb311b58cf6babccac85bbdaa5f5d0b8648845"
 
   option "with-java", "Build Java bindings for Choco solver."
+
   depends_on :java => ["1.8+", :optional]
   depends_on "bison" => :build
   depends_on "flex" => :build
@@ -12,25 +13,18 @@ class Ibex < Formula
 
   def install
     if build.with? "java"
-      system "./waf", "configure",
-        "--with-jni",
-        "--prefix=#{prefix}"
+      system "./waf", "configure", "--with-jni", "--prefix=#{prefix}"
     else
-      system "./waf", "configure",
-        "--enable-shared",
-        "--prefix=#{prefix}"
+      system "./waf", "configure", "--enable-shared", "--prefix=#{prefix}"
     end
     system "./waf", "install"
 
-    # copy and compile examples and benchmark for the test
     pkgshare.install "examples"
     pkgshare.install "benchs"
+    cd "#{pkgshare}/examples"
     cxxflags = "-frounding-math -ffloat-store -I#{include} -I#{include}/ibex"
     libflags = "-L#{lib} -libex -lprim -lClp -lCoinUtils -lm"
-    cd "#{pkgshare}/examples"
-    system "make", "defaultsolver",
-      "LIBS=#{libflags}",
-      "CXXFLAGS=#{cxxflags}"
+    system "make", "defaultsolver", "LIBS=#{libflags}", "CXXFLAGS=#{cxxflags}"
   end
 
   test do
--- a/Library/Formula/ibex.rb
+++ b/Library/Formula/ibex.rb
@@ -1,0 +1,41 @@
+class Ibex < Formula
+  desc "C++ library for constraint processing over real numbers."
+  homepage "http://www.ibex-lib.org/"
+  url "http://www.ibex-lib.org/sites/default/files/ibex-2.1.16.tar.gz"
+  sha256 "d92ff32f14d27ad7b390ae693beb311b58cf6babccac85bbdaa5f5d0b8648845"
+
+  option "with-java", "Build Java bindings for Choco solver."
+  depends_on :java => ["1.8+", :optional]
+  depends_on "bison" => :build
+  depends_on "flex" => :build
+  depends_on "pkg-config" => :build
+
+  def install
+    if build.with? "java"
+      system "./waf", "configure",
+        "--with-jni",
+        "--prefix=#{prefix}"
+    else
+      system "./waf", "configure",
+        "--enable-shared",
+        "--prefix=#{prefix}"
+    end
+    system "./waf", "install"
+
+    # copy and compile examples and benchmark for the test
+    share.install "examples"
+    share.install "benchs"
+    cxxflags = "-frounding-math -ffloat-store -I#{include} -I#{include}/ibex"
+    libflags = "-L#{lib} -libex -lprim -lClp -lCoinUtils -lm"
+    cd "#{share}/examples"
+    system "make", "defaultsolver",
+      "LIBS=#{libflags}",
+      "CXXFLAGS=#{cxxflags}"
+  end
+
+  test do
+    cp_r "#{share}/examples/.", "#{testpath}"
+    cp "#{share}/benchs/cyclohexan3D.bch", "#{testpath}"
+    system "./defaultsolver", "cyclohexan3D.bch", "1e-05", "10"
+  end
+end
--- a/Library/Formula/ibex.rb
+++ b/Library/Formula/ibex.rb
@@ -33,8 +33,8 @@ class Ibex < Formula
   end
 
   test do
-    cp_r "#{pkgshare}/examples/.", "testpath"
-    cp "#{pkgshare}/benchs/cyclohexan3D.bch", "testpath"
+    cp_r "#{pkgshare}/examples/.", testpath
+    cp "#{pkgshare}/benchs/cyclohexan3D.bch", testpath
     system "./defaultsolver", "cyclohexan3D.bch", "1e-05", "10"
   end
 end
--- a/Library/Formula/ibex.rb
+++ b/Library/Formula/ibex.rb
@@ -12,24 +12,29 @@ class Ibex < Formula
   depends_on "pkg-config" => :build
 
   def install
+    args = ["--prefix=#{prefix}"]
+
     if build.with? "java"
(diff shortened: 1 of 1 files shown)


Library/Formula/ibex.rb
```diff
@@ -0,0 +1,41 @@
...
+        "--prefix=#{prefix}"
+    else
+      system "./waf", "configure",
+        "--enable-shared",
+        "--prefix=#{prefix}"
+    end
+    system "./waf", "install"
+
+    # copy and compile examples and benchmark for the test
+    share.install "examples"
```
DomT4: Both `examples` and `benchs` should be placed in `pkgshare` now. The same syntax works, just `pkgshare` instead of `share`.

Moisan: I get errors like this with pkgshare that I didn't have with share: 
```
Error: No such file or directory - /usr/local/Cellar/ibex/2.1.16/share/examples
```

 I looked at the formula cookbook but there is no mention of pkgshare. What is the difference with share?

DomT4: `pkgshare` is new. Where was the error being thrown from, runtime or during the test?

`pkgshare` = prefix/share/formula_name/
`share` = prefix/share

We've started placing examples and similar things in `pkgshare` because there's a risk for accidental conflict installing every formula's examples into the same common prefix - One formula we handled the other day installed an example file to generate just called `example.c`. Installing them into `pkgshare` ensures that conflict can never happen.

Moisan: The error was being thrown during the installation when `pkgshare` was called right after the compilation.

DomT4: You'd need to change all the other references as well, so:

```ruby
pkgshare.install "examples"
pkgshare.install "benchs"

cd "#{pkgshare}/examples"
```

And so on in the `test do` block as well.

Moisan: Wow I feel stupid for not seeing that earlier.

Library/Formula/ibex.rb
```diff
@@ -0,0 +1,41 @@
+class Ibex < Formula
+  desc "C++ library for constraint processing over real numbers."
+  homepage "http://www.ibex-lib.org/"
+  url "http://www.ibex-lib.org/sites/default/files/ibex-2.1.16.tar.gz"
+  sha256 "d92ff32f14d27ad7b390ae693beb311b58cf6babccac85bbdaa5f5d0b8648845"
+
+  option "with-java", "Build Java bindings for Choco solver."
+  depends_on :java => ["1.8+", :optional]
```
DomT4: One empty line between the `option` and the first `depends_on` please.

Library/Formula/ibex.rb
```diff
@@ -0,0 +1,41 @@
...
+  depends_on "bison" => :build
+  depends_on "flex" => :build
+  depends_on "pkg-config" => :build
+
+  def install
+    if build.with? "java"
+      system "./waf", "configure",
+        "--with-jni",
+        "--prefix=#{prefix}"
```
DomT4: This is probably fine to compress to one line, It doesn't look too long:

```ruby
system "./waf", "configure", "--with-jni", "--prefix=#{prefix}"
```

Library/Formula/ibex.rb
```diff
@@ -0,0 +1,41 @@
...
+    else
+      system "./waf", "configure",
+        "--enable-shared",
+        "--prefix=#{prefix}"
```
DomT4: Same here.

Library/Formula/ibex.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+    system "./waf", "install"
+
+    # copy and compile examples and benchmark for the test
```
DomT4: Don't worry about explaining this. It happens enough elsewhere that it's clear. Kill the comment :)

Library/Formula/ibex.rb
```diff
@@ -0,0 +1,41 @@
...
+    pkgshare.install "examples"
+    pkgshare.install "benchs"
+    cxxflags = "-frounding-math -ffloat-store -I#{include} -I#{include}/ibex"
+    libflags = "-L#{lib} -libex -lprim -lClp -lCoinUtils -lm"
+    cd "#{pkgshare}/examples"
```
DomT4: Any chance we can do this before installing to `pkgshare` or?

DomT4:
A few nits to tidy up but looks great otherwise.

DomT4:
I'm getting a build failure with Java sometimes. Possibly a race condition?

```
mkdir: /usr/local/Cellar/ibex/2.1.16/include/coin: File exists
make[3]: *** [install-includecoinHEADERS] Error 1
make[3]: *** Waiting for unfinished jobs....
 /bin/sh ../../libtool --mode=install /usr/bin/install -c  'libCoinUtils.la' '/usr/local/Cellar/ibex/2.1.16/lib/libCoinUtils.la'
/usr/bin/install -c .libs/libCoinUtils.3.9.11.dylib /usr/local/Cellar/ibex/2.1.16/lib/libCoinUtils.3.9.11.dylib
(cd /usr/local/Cellar/ibex/2.1.16/lib && { ln -s -f libCoinUtils.3.9.11.dylib libCoinUtils.3.dylib || { rm -f libCoinUtils.3.dylib && ln -s libCoinUtils.3.9.11.dylib libCoinUtils.3.dylib; }; })
(cd /usr/local/Cellar/ibex/2.1.16/lib && { ln -s -f libCoinUtils.3.9.11.dylib libCoinUtils.dylib || { rm -f libCoinUtils.dylib && ln -s libCoinUtils.3.9.11.dylib libCoinUtils.dylib; }; })
/usr/bin/install -c .libs/libCoinUtils.lai /usr/local/Cellar/ibex/2.1.16/lib/libCoinUtils.la
----------------------------------------------------------------------
Libraries have been installed in:
   /usr/local/Cellar/ibex/2.1.16/lib

If you ever happen to want to link against installed libraries
in a given directory, LIBDIR, you must either use libtool, and
specify the full pathname of the library, or use the `-LLIBDIR'
flag during linking and do at least one of the following:
   - add LIBDIR to the `DYLD_LIBRARY_PATH' environment variable
     during execution

See any operating system documentation about shared libraries for
more information, such as the ld(1) and ld.so(8) manual pages.
----------------------------------------------------------------------
make[2]: *** [install-am] Error 2
make[1]: *** [install-recursive] Error 1
Leaving '/private/tmp/ibex20150824-89447-1nrjh0d/ibex-2.1.16/3rd/Clp-1.15.6'
failed to build Clp
```

Library/Formula/ibex.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  option "with-java", "Build Java bindings for Choco solver."
+
+  depends_on :java => ["1.8+", :optional]
+  depends_on "bison" => :build
+  depends_on "flex" => :build
+  depends_on "pkg-config" => :build
+
+  def install
+    if build.with? "java"
```
DomT4: Rather than duping the whole block here let's go with something like:

```ruby
    args = ["--prefix=#{prefix}"]

    if build.with? "java"
      args << "--with-jni"
    else
      args << "--enable-shared"
    end

    system "./waf", "configure", *args
```

Library/Formula/ibex.rb
```diff
@@ -0,0 +1,35 @@
...
+      system "./waf", "configure", "--with-jni", "--prefix=#{prefix}"
+    else
+      system "./waf", "configure", "--enable-shared", "--prefix=#{prefix}"
+    end
+    system "./waf", "install"
+
+    pkgshare.install "examples"
+    pkgshare.install "benchs"
+    cd "#{pkgshare}/examples"
```
DomT4: It'd be nice to do all the compiling here in the testpath, rather than switching into the prefix and doing more building there. Something like:

```ruby
    cd "examples" do
      cxxflags = "-frounding-math -ffloat-store -I#{include} -I#{include}/ibex"
      libflags = "-L#{lib} -libex -lprim -lClp -lCoinUtils -lm"
      system "make", "defaultsolver", "LIBS=#{libflags}", "CXXFLAGS=#{cxxflags}"
    end

    pkgshare.install %w[examples benchs]
```

Should work.

Library/Formula/ibex.rb
```diff
@@ -0,0 +1,35 @@
...
+    cxxflags = "-frounding-math -ffloat-store -I#{include} -I#{include}/ibex"
+    libflags = "-L#{lib} -libex -lprim -lClp -lCoinUtils -lm"
+    system "make", "defaultsolver", "LIBS=#{libflags}", "CXXFLAGS=#{cxxflags}"
+  end
+
+  test do
+    cp_r "#{pkgshare}/examples/.", "#{testpath}"
```
DomT4: `"#{testpath}"` here can be just `testpath`. No need for the `#{}` logic in that part.

Library/Formula/ibex.rb
```diff
@@ -0,0 +1,35 @@
...
+    cp "#{pkgshare}/benchs/cyclohexan3D.bch", "#{testpath}"
```
DomT4: And here, re the above comment.

Moisan:
I covered all comments but I could not reproduce the build failure with java.

DomT4:
I didn't get it on a second attempt, so perhaps I had a one-time issue. If we see further reports of it we may need to look at the race possibility, but otherwise I'll put it down to a local problem.

Merged in https://github.com/Homebrew/homebrew/commit/c7cc760c60b3b1b9f87f556089a28e393416ac8d. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #42773, opened 2015-08-10
Fix jq 1.4 tarball location

nicowilliams:
We moved downloadables for jq out of the gh-pages branch and into github releases.  This means that the URL for the jq 1.4 tarball changed.  Sadly, github has no way to configure redirects.


dunn:
Thanks so much! Merged in 6b68976c414e30c22979fedd8f623bc8ab38bae7.

nicowilliams:
Thanks!

PR #42577, opened 2015-08-06
update hackrf to 2015.07.2

jedahan:

PR #42777, opened 2015-08-11
hackrf update to v2015.07.2

gyaresu:
I have added the correct checksum which was the problem in this other PR: https://github.com/Homebrew/homebrew/pull/42577

### DIFF
--- a/Library/Formula/hackrf.rb
+++ b/Library/Formula/hackrf.rb
@@ -1,8 +1,8 @@
 class Hackrf < Formula
   desc "Low cost software radio platform"
   homepage "https://github.com/mossmann/hackrf"
-  url "https://github.com/mossmann/hackrf/archive/v2014.08.1.tar.gz"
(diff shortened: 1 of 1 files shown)


dunn:
While you're at it, would it be possible to add a better test, preferably something that doesn't exit with `1`? Thanks!

gyaresu:
I haven't written ruby tests before but I'll have a look at some others and give it a go.

gyaresu:
Ah. I see the problem in choosing a good way to test.
Without the board plugged in you don't have many options.

coming back with a success from `hackrf_transfer` (which just outputs the help info) seems fairly logical.

Ideas?

dunn:
Oh, I see the issue.  I think we can live with it as it is.

dunn:
Merged in b38ed95461f0ee3df9b27408c3ee364d0697ee4d 📯

PR #42779, opened 2015-08-11
jenkins 1.624

stuwil:


dunn:
Might work to just add `ENV["LC_ALL"] = "en_US.UTF-8"` at the beginning of the test.

stuwil:
Did some experimenting, looks like the test just needed more time for the daemon to start up. Giving it a full minute seems to pass on all test boxes. Not sure if there's a better way to handle than sleeping... please let me know if you have any thoughts!

DomT4:
I'm a little wary of `sleep 60` but if it's not hitting the test timeout I guess it's not a huge problem. I suspect it takes that long because when Java is first invoked it will take half a lifetime to "ready" itself.

Merged in https://github.com/Homebrew/homebrew/commit/0dbc572ba58dd8ea129265224af1e683e7485c8d. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

Issue #38306, opened 2015-04-02
Need a formula for rebar3


ijt:
Rebar3 is the new version of the main build tool for Erlang. http://www.rebar3.org/v3.0/docs/faq

It would be nice to have a formula for it, but brew doesn't seem to be able to recognize the difference between rebar and rebar3.

$ brew create https://github.com/rebar/rebar3.git
Error: /usr/local/Library/Formula/rebar.rb already exists

tdsmith:
Try pointing it at a release tarball.

If this will supplant rebar, we can add a `devel` block to the rebar formula.

If this needs to exist in parallel to rebar, this can go in `homebrew-head-only` or `-devel-only` for now, or else we can wait for a stable release before adding it to core.

ijt:
There aren't any release tarballs yet, as far as I can tell, though there
are nightly builds. I'm adding a head block.

tdsmith:
They're linked here: https://github.com/rebar/rebar3/releases

DomT4:
Ideally, would be added as a ` devel ` block. ` devel-only ` is intended primarily for things that have absolutely no stable release yet, which is why it's the easiest and most low-volume repo in the world to maintain :smile_cat:.

DomT4:
Closing per Tim's earlier comment. PR welcome in line with that.

PR #42781, opened 2015-08-11
rebar3 3.0.0-beta-1

talentdeficit:
a formula for the latest rebar3 beta


dunn:
If there's no stable release, I'm afraid this needs to be submitted to our devel-only tap: https://github.com/Homebrew/homebrew-devel-only

talentdeficit:
ok. i misinterpreted the comments on #38306 i guess. reopening against devel-only

PR #42788, opened 2015-08-11
adding plist running as daemon

mulbane:
adding plist entry for cjdns formula for persistence

### DIFF
--- a/Library/Formula/cjdns.rb
+++ b/Library/Formula/cjdns.rb
@@ -19,6 +19,42 @@ class Cjdns < Formula
     (share+"test").install "build_darwin/test_testcjdroute_c" => "cjdroute_test"
   end
 
+  def caveats
+    s = <<-EOS.undent
+      !!!IMPORTANT!!!
+      Before using cjdns, you must generate a configuration file and find peers manually.
+      Once you have found eligible peers, use the following command to generate a config that is appropriate for Launchd
+        sudo bash -c "cjdroute --genconf | sed 's@\\"noBackground\\":0@\\"noBackground\\":1@' > /etc/cjdroute.conf"
+    EOS
+  end
+
+  plist_options :manual => "cjdroute"
+
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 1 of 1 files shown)


Library/Formula/cjdns.rb
```diff
@@ -19,6 +19,42 @@ def install
     (share+"test").install "build_darwin/test_testcjdroute_c" => "cjdroute_test"
   end
 
+  def caveats
+    s = <<-EOS.undent
+      !!!IMPORTANT!!!
+      Before using cjdns, you must generate a configuration file and find peers manually.
+      Once you have found eligible peers, use the following command to generate a config that is appropriate for Launchd
+        sudo bash -c "cjdroute --genconf | sed 's@\\"noBackground\\":0@\\"noBackground\\":1@' > /etc/cjdroute.conf"
```
mikemcquaid: Can this not be run automatically? Also, can the configuration file not use `HOMEBREW_PREFIX/etc`?

Library/Formula/cjdns.rb
```diff
@@ -19,6 +19,42 @@ def install
...
+    <plist version="1.0">
+      <dict>
+        <key>Label</key>
+        <string>#{plist_name}</string>
+        <key>ProgramArguments</key>
+        <array>
+          <string>sudo</string>
+          <string>bash</string>
+          <string>-c</string>
+          <string>/usr/local/bin/cjdroute &lt; /etc/cjdroute.conf</string>
```
mikemcquaid: You should use `opt_bin` rather than hardcoding `/usr/local` here.

Library/Formula/cjdns.rb
```diff
@@ -19,6 +19,42 @@ def install
...
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+      <dict>
+        <key>Label</key>
+        <string>#{plist_name}</string>
+        <key>ProgramArguments</key>
+        <array>
+          <string>sudo</string>
```
mikemcquaid: You can't rely on this command running without `sudo`. It should be a startup plist if it needs root.

DomT4:
Closing as dead. We'll happily review a new PR with the various comments addressed here, Thanks for the submission!

PR #42740, opened 2015-08-10
eigen 3.2.5

guoxiao:

PR #42792, opened 2015-08-11
eigen: fix audit warnings

guoxiao:


bfontaine:
Isn’t there anything else to add? We don’t accept style-only pull-requests.

guoxiao:
Nothing else. Only because I submitted a PR #42740 yesterday and forgot to push this.

I remembered that @BrewTestBot would check the style of formula. Does it skip this check now?

bfontaine:
It checks the style only on new formulae.

PR #42796, opened 2015-08-11
uhd 003.008.005

jiegec:


Library/Formula/uhd.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def install
+    ENV["CHEETAH_INSTALL_WITHOUT_SETUPTOOLS"] = "1"
+    ENV.prepend_create_path "PYTHONPATH", libexec+"lib/python2.7/site-packages"
+    install_args = ["setup.py", "install", "--prefix=#{libexec}"]
+
+    resource("Markdown").stage { system "python", *install_args }
+    resource("Cheetah").stage { system "python", *install_args }
+
+    chdir "host"
```
bfontaine: You can use `cd "host" do` instead.

Library/Formula/uhd.rb
```diff
@@ -0,0 +1,50 @@
...
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args
+      system "make"
+      system "make", "test"
+      system "make", "install"
+    end
+  end
+
+  test do
+    system "uhd_find_devices", "--help"
```
bfontaine: Please use `"#{bin}/uhd_find_devices"`.

Library/Formula/uhd.rb
jiegec: The CMakeLists.txt is in that directory

Library/Formula/uhd.rb
```diff
@@ -0,0 +1,50 @@
...
+  homepage "http://files.ettus.com/manual/"
+
+  stable do
+    url "https://github.com/EttusResearch/uhd/archive/release_003_008_005.tar.gz"
+    sha256 "e0a36e64bffa7d06ac948fd8fc07ff3ce321e1efc3f7e8604b53e0c3ce842989"
+  end
+
+  head do
+    url "https://github.com/EttusResearch/uhd.git"
+  end
```
bfontaine: You can use `head "https://github.com/EttusResearch/uhd.git"`.

Library/Formula/uhd.rb
```diff
@@ -0,0 +1,50 @@
+class Uhd < Formula
+  desc "The UHD™ is the hardware driver for all USRP devices."
```
bfontaine: You could shorten this to `hardware driver for all USRP devices`.

dunn:
Just a minor style issue:
```
        ==> brew style uhd
== /usr/local/Library/Formula/uhd.rb ==
C: 43: 86: Prefer !expression.nil? over expression != nil.
```

jiegec:
Uhh, I think I can't fix the gnu radio running mavericks...

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+      resource("sphinx_rtd_theme").stage { system "python", *install_args }
+      resource("alabaster").stage { system "python", *install_args }
+      resource("babel").stage { system "python", *install_args }
+      resource("docutils").stage { system "python", *install_args }
+      resource("pygments").stage { system "python", *install_args }
+      resource("jinja2").stage { system "python", *install_args }
+      resource("markupsafe").stage { system "python", *install_args }
+      resource("snowballstemmer").stage { system "python", *install_args }
+      resource("six").stage { system "python", *install_args }
+      resource("pytz").stage { system "python", *install_args }
```
bfontaine: You could do something like:

```ruby
%w[sphinx sphinx_rtd_theme alabaster …].each do |res|
  res.stage { system "python", *install_args }
end
```

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+    end
+
+    if build.with? "zeromq"
+      resource("cppzmq").stage "#{include}"
+    end
```
bfontaine: You can do that in one line:

```rb
resource("cppzmq").stage "#{include}" if build.with? "zeromq"
```

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+  depends_on "qt" if build.with? "python"
+  depends_on "qwt" if build.with? "python"
+  depends_on "pyqt" if build.with? "python"
+  depends_on "wxpython" if build.with? "wx"
+  depends_on "sdl" => :optional
+  depends_on "uhd" => :optional
+  depends_on "doxygen" => :optional
+
+  def install
+    if (build.with? "wx") && ! (build.with? "python")
```
bfontaine: Use `build.without?` instead of `! (build.with? ...)`.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+  resource "cppzmq" do
+    url "https://github.com/zeromq/cppzmq/raw/34c8e4395c94d34a89bbeaaf2b8f9c94a8293c84/zmq.hpp"
+    sha256 "6cdd9d920f4a0f9f3a3257541321bef8369d9735c88b84ba8ae0e461f53e476c"
+  end
+
+  option "with-python", "Build python support"
+  option "with-sphinx", "Build sphinx document"
+  option "with-wavelet", "Build gr-wavelet"
+  option "with-wx", "Build wx gui (requires --with-python)"
+  option "with-qt", "Build qt gui (requires --with-python)"
```
bfontaine: Can’t we just build with Python support if either `--with-python`, `--with-wx` or `-with-qt` is passed, instead of failing if e.g. `--with-wx` is passed but not `--with-python`?
Also, how common are these options?

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+  option "with-zeromq", "BUild gr-zeromq"
```
bfontaine: “BUild” → “Build”

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
+class Gnuradio < Formula
+  desc "SDK providing the signal processing runtime and processing blocks"
+  homepage "https://gnuradio.org/"
+
+  stable do
+    url "http://gnuradio.org/releases/gnuradio/gnuradio-3.7.8rc1.tar.gz"
+    sha256 "a2228c8cba0194924c7549e37085b5c3837c3282ac96f5bf2392c07cbaed0895"
+  end
```
bfontaine: You can remove the block and make the `url` and `sha256` top-level.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+
+  # cheetah starts here
+  resource "Markdown" do
```
DomT4: `resource` blocks should go beneath the `option` and `depends_on` blocks. Thanks!

Library/Formula/gnuradio.rb
DomT4: Yeah. We've been trying to kill off `if build.with?` dependencies for a while, and I'd rather not introduce a bunch of fresh ones here. I don't see a particularly special reason why we'd need to modify the default text here.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+  end
+  # sphinx ends here
+
+  resource "cppzmq" do
+    url "https://github.com/zeromq/cppzmq/raw/34c8e4395c94d34a89bbeaaf2b8f9c94a8293c84/zmq.hpp"
+    sha256 "6cdd9d920f4a0f9f3a3257541321bef8369d9735c88b84ba8ae0e461f53e476c"
+  end
+
+  option "with-python", "Build python support"
+  option "with-sphinx", "Build sphinx document"
```
DomT4: Documentation is generally built by default, since it's not the sort of thing users should have to build from source to obtain. Let's make this `without-documentation` instead.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+  option "with-zeromq", "BUild gr-zeromq"
+
+  depends_on "cmake" => :build
+  depends_on "pkg-config" => :build
+  depends_on "swig" => :build if build.with? "python"
+
+  depends_on "boost"
+  depends_on "cppunit"
+  depends_on "fftw"
+  depends_on "python"
```
DomT4: We almost never use unconditional Homebrew Python dependencies because it means when users to try install the bottle they get a Homebrew Python forced upon them, which may not be desired. If it needs 2.7 or above you can do:

```ruby
```

If it can use 2.6 you can remove the dependency entirely.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+  depends_on "homebrew/python/numpy"
```
DomT4: You can't have mandatory cross-tap dependencies in the core. This will need to be vendored or made optional.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+  depends_on "gsl" if build.with? "wavelet"
+  depends_on "qt" if build.with? "python"
```
DomT4: If this is necessary, block them all in the same block, not individually:

```ruby
if build.with? "ducks"
  depends_on "water"
  depends_on "food"
  depends_on "freedom"
end
```

Rather than:

```ruby
depends_on "water" if build.with? "ducks"
depends_on "food" if build.with? "ducks"
```

And so on.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+  depends_on "qwt" if build.with? "python"
+  depends_on "pyqt" if build.with? "python"
+  depends_on "wxpython" if build.with? "wx"
+  depends_on "sdl" => :optional
+  depends_on "uhd" => :optional
+  depends_on "doxygen" => :optional
```
DomT4:
```ruby
[:build, :recommended]
```

For doxygen.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+
+    if (build.with? "qt") && ! (build.with? "python")
+      raise "--with-qt depends on --with-python"
+    end
+
+    ENV.append "LDFLAGS", "-undefined dynamic_lookup"
+
+    ENV["CHEETAH_INSTALL_WITHOUT_SETUPTOOLS"] = "1"
+    ENV.prepend_create_path "PYTHONPATH", libexec+"lib/python2.7/site-packages"
+    install_args = ["setup.py", "install", "--prefix=#{libexec}"]
```
DomT4: Does the formula or any of the resources here reject:

```ruby
system "python", *Language::Python.setup_install_args(libexec)
```

?

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+    if build.with? "zeromq"
+      resource("cppzmq").stage "#{include}"
+    end
+
+    ENV.prepend_path "PATH", (libexec+"bin")
+
+    args = std_cmake_args
+    args << "-DENABLE_DEFAULT=OFF"
+    args << "-DENABLE_PYTHON=ON" if build.with? "python"
+    args << "-DENABLE_GR_ANALOG=ON"
```
DomT4: If some of these are enabled by default, you don't need to declare them all `ON`. Only do that when we want to change the default setting.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+    args << "-DENABLE_GR_ZEROMQ=ON" if build.with? "zeromq"
+
+    mkdir "build" do
+      system "cmake", "..", *args
+      system "make", "-j2"
+      system "make", "install"
+    end
+  end
+
+  def caveats; <<-EOS.undent
```
DomT4: Let's not add this block.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,230 @@
...
+      osx: true
+      python-support: #{build.with? "python"} --with-python
+      sphinx: #{build.with? "sphinx"} --with-sphinx
+      testing-support: true
+      volk: true
+    EOS
+  end
+
+  test do
+    !shell_output("#{bin}/gnuradio-config-info -v").chomp.index("#{version}").nil?
```
DomT4: Use an `assert_match` instead here please.

Library/Formula/uhd.rb
```diff
@@ -0,0 +1,45 @@
...
+  resource "Cheetah" do
+    url "https://pypi.python.org/packages/source/C/Cheetah/Cheetah-2.4.4.tar.gz"
+    sha256 "be308229f0c1e5e5af4f27d7ee06d90bb19e6af3059794e5fd536a6f29a9b550"
+  end
+
+  depends_on "cmake" => :build
+  depends_on "boost"
+  depends_on "libusb"
+  depends_on "python"
+  depends_on "doxygen" => :optional
```
DomT4:
```ruby
[:build, :recommended]
```

For doxygen.

Library/Formula/uhd.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def install
+    ENV["CHEETAH_INSTALL_WITHOUT_SETUPTOOLS"] = "1"
+    ENV.prepend_create_path "PYTHONPATH", libexec+"lib/python2.7/site-packages"
+    install_args = ["setup.py", "install", "--prefix=#{libexec}"]
```
DomT4:
```ruby
system "python", *Language::Python.setup_install_args(libexec)
```

Library/Formula/uhd.rb
```diff
@@ -0,0 +1,45 @@
...
+        system "cmake", "..", *std_cmake_args
+        system "make"
+        system "make", "test"
+        system "make", "install"
+      end
+    end
+  end
+
+  test do
+    !shell_output("#{bin}/uhd_find_devices --help || true").chomp.index("#{version}").nil?
```
DomT4: And an `assert_match` here is preferable as well.

Library/Formula/gnuradio.rb
jiegec: Default is ALL ON.

Library/Formula/gnuradio.rb
jiegec: Ahhh, the document takes about 500M's space.

jiegec:
Oh, why the mountaion_lion can't connect to the homepage?

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,193 @@
+class Gnuradio < Formula
+  desc "SDK providing the signal processing runtime and processing blocks"
+  homepage "https://gnuradio.org/"
+  url "http://gnuradio.org/releases/gnuradio/gnuradio-3.7.8rc1.tar.gz"
+  sha256 "a2228c8cba0194924c7549e37085b5c3837c3282ac96f5bf2392c07cbaed0895"
+
+  option "with-python", "Build python support"
+  option "without-document", "Don't build document"
```
mikemcquaid: Do you mean documentation?

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,193 @@
...
+  option "with-wx", "Build wx gui (enables --with-python)"
+  option "with-qt", "Build qt gui (enables --with-python)"
+  option "with-wavelet", "Build gr-wavelet"
+  option "with-zeromq", "Build gr-zeromq"
+  option :universal
+
+  @build_python = (build.with? "python") || (build.with? "wx") || (build.with? "qt")
```
mikemcquaid: Can you inline this variable where it's used? Thanks.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,193 @@
...
+    args << "-DENABLE_GR_NOAA=ON"
+    args << "-DENABLE_GR_CHANNELS=ON"
+    args << "-DENABLE_GR_AUDIO=ON"
+    args << "-DENABLE_GR_FCD=ON"
+    args << "-DENABLE_GR_VOCODER=ON"
+    args << "-DENABLE_GR_FEC=ON"
+    args << "-DENABLE_GR_DIGITAL=ON"
+    args << "-DENABLE_GR_DTV=ON"
+    args << "-DENABLE_GR_ATSC=ON"
+    args << "-DENABLE_GR_TRELLIS=ON"
```
mikemcquaid: Can you build an array rather than duplicating `args << ` each tie.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,193 @@
...
+    end
+
+    if build.with? "document"
+      args << "-DENABLE_DOXYGEN=ON"
+      args << "-DENABLE_SPHINX=ON"
+    end
+
+    mkdir "build" do
+      system "cmake", "..", *args
+      system "make", "-j2"
```
mikemcquaid: Why `-j2`?

Library/Formula/uhd.rb
```diff
@@ -0,0 +1,51 @@
...
+        system "cmake", "..", *std_cmake_args
+        system "make"
+        system "make", "test"
+        system "make", "install"
+      end
+    end
+  end
+
+  test do
+    assert_match "#{version}", shell_output("#{bin}/uhd_find_devices --help || true").chomp
```
mikemcquaid: Don't use `|| true`; instead pass the second argument to `shell_output` with the exit code

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,193 @@
...
+  homepage "https://gnuradio.org/"
+  url "http://gnuradio.org/releases/gnuradio/gnuradio-3.7.8rc1.tar.gz"
+  sha256 "a2228c8cba0194924c7549e37085b5c3837c3282ac96f5bf2392c07cbaed0895"
+
+  option "with-python", "Build python support"
+  option "without-document", "Don't build document"
+  option "with-wx", "Build wx gui (enables --with-python)"
+  option "with-qt", "Build qt gui (enables --with-python)"
+  option "with-wavelet", "Build gr-wavelet"
+  option "with-zeromq", "Build gr-zeromq"
```
mikemcquaid: What do these do?

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,193 @@
+class Gnuradio < Formula
+  desc "SDK providing the signal processing runtime and processing blocks"
+  homepage "https://gnuradio.org/"
+  url "http://gnuradio.org/releases/gnuradio/gnuradio-3.7.8rc1.tar.gz"
+  sha256 "a2228c8cba0194924c7549e37085b5c3837c3282ac96f5bf2392c07cbaed0895"
+
+  option "with-python", "Build python support"
```
mikemcquaid: What does this do?

jiegec:
There is still a problem with --with-python. It links the python lib and i can't find a way to remove -lpython and add -undefined dynamic_lookup

jiegec:
Ok now, fixes the python linking problem.

Library/Formula/gnuradio-osmosdr.rb
```diff
@@ -0,0 +1,87 @@
...
+    res = %w[Markdown Cheetah]
+    res.each do |r|
+      resource(r).stage do
+        system "python", *Language::Python.setup_install_args(libexec/"vendor")
+      end
+    end
+
+    args << "-DENABLE_DEFAULT=OFF"
+    enabled_componenets = %w[fcd file rtl_tcp rfspace]
+    enabled_componenets += %w[python] if build.with? "python"
```
bfontaine: Use `<< "python"` instead of `+= %w[python]` (same thing for the lines below.

Library/Formula/gnuradio-osmosdr.rb
```diff
@@ -0,0 +1,87 @@
...
+    enabled_componenets += %w[rtl] if build.with? "librtlsdr"
+    enabled_componenets += %w[uhd] if build.with? "uhd"
+    enabled_componenets += %w[hackrf] if build.with? "hackrf"
+    enabled_componenets.each do |c|
+      args << "-DENABLE_#{c.upcase.split("-").join("_")}=ON"
```
bfontaine: This `.split("-").join("_")` seems unnecessary, there isn’t any element with an hyphen.

Library/Formula/gnuradio-osmosdr.rb
```diff
@@ -0,0 +1,87 @@
...
+    mkdir "build" do
+      system "cmake", "..", *args
+      system "make"
+      system "make", "install"
+    end
+  end
+
+  test do
+    # This depends on whether the device is connected. A good way to test needed.
+  end
```
bfontaine: Please add a test.

Library/Formula/gnuradio-osmosdr.rb
```diff
@@ -0,0 +1,87 @@
...
+    if build.universal?
+      ENV.universal_binary
+      args << "-DCMAKE_OSX_ARCHITECTURES=#{Hardware::CPU.universal_archs.as_cmake_arch_flags}"
+    end
+
+    ENV["CHEETAH_INSTALL_WITHOUT_SETUPTOOLS"] = "1"
+    ENV.prepend_create_path "PYTHONPATH", libexec+"vendor/lib/python2.7/site-packages"
+
+    res = %w[Markdown Cheetah]
+    res.each do |r|
```
bfontaine: You can use `%w[Markdown Cheetah].do |r|`.

Library/Formula/gnuradio-osmosdr.rb
```diff
@@ -0,0 +1,87 @@
+class GnuradioOsmosdr < Formula
+  desc "osmocom Gnu Radio Blocks"
+  homepage "http://sdr.osmocom.org/trac/wiki/GrOsmoSDR"
+  url "http://cgit.osmocom.org/gr-osmosdr/snapshot/gr-osmosdr-0.1.4.zip"
+  sha256 "a5c439f275cf11965d2fc67957e1d221b4f985ff5d87cdc9760522ba8319215d"
+
+  option "without-python", "Don't build gnuradio's python-support component"
+
+  patch :DATA
```
bfontaine: Please add a comment explaining why it is needed.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,195 @@
+class Gnuradio < Formula
+  desc "SDK providing the signal processing runtime and processing blocks"
+  homepage "http://gnuradio.org/"
+  url "http://gnuradio.org/releases/gnuradio/gnuradio-3.7.8rc1.tar.gz"
+  sha256 "a2228c8cba0194924c7549e37085b5c3837c3282ac96f5bf2392c07cbaed0895"
+
+  patch :DATA
```
bfontaine: Please add a comment explaining why it is needed.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,195 @@
...
+
+  option "without-python", "Don't build gnuradio's python-support component"
+  option "with-documentation", "Build documentation"
+  option "with-wx", "Build wx gui (enables --with-python)"
+  option "with-qt", "Build qt gui (enables --with-python)"
+  option "with-wavelet", "Build gnuradio's gr-wavelet component"
+  option "with-zeromq", "Build gnuradio's gr-zeromq component"
```
bfontaine: How common are these needs?

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,195 @@
...
+    res = %w[Markdown Cheetah lxml]
+    res += %w[sphinx sphinx_rtd_theme alabaster babel docutils pygments
+              jinja2 markupsafe snowballstemmer six pytz] if build.with? "documentation"
+    res.each do |r|
+      resource(r).stage do
+        system "python", *Language::Python.setup_install_args(libexec/"vendor")
+      end
+    end
+
+    resource("cppzmq").stage "#{include}" if build.with? "zeromq"
```
bfontaine: Use `include.to_s` instead of `"#{include}".

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,195 @@
...
+    enabled_componenets << "gr-qtgui" if build.with? "qt"
+    enabled_componenets << "gr-video-sdl" if build.with? "sdl"
+    enabled_componenets << "gr-wavelet" if build.with? "wavelet"
+    enabled_componenets << "gr-uhd" if build.with? "uhd"
+    enabled_componenets << "gr-wxgui" if build.with? "wx"
+    enabled_componenets << "gr-zeromq" if build.with? "zeromq"
+    enabled_componenets += %w[python gr-ctrlport grc gr-utils] if build_python
+    enabled_componenets += %w[doxygen sphinx] if build.with? "documentation"
+    enabled_componenets.each do |c|
+      args << "-DENABLE_#{c.upcase.split("-").join("_")}=ON"
```
bfontaine: You can replace hyphens with underscores above and remove this `.split("-").join("_")`.

Library/Formula/uhd.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    res = %w[Markdown Cheetah]
+    res.each do |r|
+      resource(r).stage do
+        system "python", *Language::Python.setup_install_args(libexec/"vendor")
+      end
+    end
+
+    cd "host" do
+      mkdir "build" do
```
bfontaine: You might want to try with `mkdir "host/build"` and remove the `cd "host"` above.

jiegec:
@bfontaine Now how to deal with the numpy dependency? This is required. It needs to patch the numpy to build, but how can I patch a resource?

jiegec:
A temp fix:
```
    resource("numpy").stage do
      system "zsh"
      require "open-uri"
      download = open("https://gist.githubusercontent.com/jiegec/e66b13361c1f8109a1a7/raw/c1ef3026d04b0f4b190cbb1bc0a4682e5d022c33/numpy.diff")
      IO.copy_stream(download, "numpy.diff")
      system "patch", "-p1", "-i", "numpy.diff"
    end if build.with? "python"
```
Is it acceptable?

jiegec:
@bfontaine Ready to merge!

bfontaine:
@jiegec Is there any way you could split this into two commits, one per formula?

jiegec:
@bfontaine OK, in no time.

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,193 @@
+class Gnuradio < Formula
+  desc "SDK providing the signal processing runtime and processing blocks"
+  homepage "http://gnuradio.squarespace.com/"
+  url "http://gnuradio.org/releases/gnuradio/gnuradio-3.7.8rc1.tar.gz"
+  sha256 "a2228c8cba0194924c7549e37085b5c3837c3282ac96f5bf2392c07cbaed0895"
+
+  # Replace -lpython with -undefined dynamic_lookup in linker flags.
+  patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

jiegec: Done.

jiegec:
Update to gnuradio 3.7.8

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,196 @@
...
+  desc "SDK providing the signal processing runtime and processing blocks"
+  homepage "http://gnuradio.squarespace.com/"
+  url "http://gnuradio.org/releases/gnuradio/gnuradio-3.7.8.tar.gz"
+  sha256 "fe19cb54b5d77fb76dde61d5cf184c6dee7066779b45c51676bae6e6d0cd4172"
+
+  # These python extension modules were linked directly to a Python
+  #   framework binary.
+  # Replace -lpython with -undefined dynamic_lookup in linker flags.
+  # https://github.com/gnuradio/gnuradio/pull/604
+  patch :DATA
```
mikemcquaid: Does using `https://github.com/gnuradio/gnuradio/pull/604.patch` work?

jiegec: Yes, that works even if line number is not correct. Updated

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,187 @@
...
+              jinja2 markupsafe snowballstemmer six pytz] if build.with? "documentation"
+    res.each do |r|
+      resource(r).stage do
+        system "python", *Language::Python.setup_install_args(libexec/"vendor")
+      end
+    end
+
+    resource("cppzmq").stage include.to_s
+
+    ENV.prepend_path "PATH", (libexec+"vendor/bin")
```
DomT4: Is this for Sphinx et al. when documentation is built, or?

Library/Formula/gnuradio.rb
```diff
@@ -0,0 +1,187 @@
...
+    mkdir "build" do
+      system "cmake", "..", *args
+      system "make"
+      system "make", "install"
+    end
+
+    # Remove useless data files installed in #{bin}
+    delete_files = %w[ctrlport-monitorc ctrlport-monitoro
+                      perf-monitorxc perf-monitorxo]
+    delete_files.each do |f|
```
DomT4: You might be able to one-line this:

```ruby
delete_files.each { |f| rm bin/"gr-#{f}" }
```

DomT4:
One nit & one question, but this looks great. Almost ready to merge IMO.

Library/Formula/gnuradio.rb
jiegec: Yes, it lets the cmake find the sphinx executable

jiegec: That is only needed when build with documentation.

jiegec: But it doesn't have any bad effect. Remove it?

DomT4: It's fine, was just curious.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/17e50fd6fe6486ded4a2997b9721557be42fbcde & https://github.com/Homebrew/homebrew/commit/026eea7171f2c972b1daf8abbe243ebd9a4daa88. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #42800, opened 2015-08-11
rename: plt-racket -> racket

vladshablinsky:
Rename racket from https://github.com/Homebrew/homebrew/issues/14374


dunn:
Small audit nit:
```

==> audit problems
racket:
 * `version` (line 6) should be put before `checksum` (line 5)
```

mikemcquaid:
Thanks @vladshablinsky, keep 'em coming!

For future reference the preferred commit message format for renaming formulae is `racket: rename from plt-racket.`.

PR #42801, opened 2015-08-11
Formula renames improvements

vladshablinsky:
* Add `--force` to skip check whether installed and migrating formulae are from the same taps.
* Add "Renamed Formulae" block to report.

@xu-cheng @mikemcquaid this is a branch for improvements for https://github.com/Homebrew/homebrew/pull/41006.

I'm going to work on improvements on this branch until everybody is happy.

### DIFF
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -60,27 +60,18 @@ module Homebrew
       tabs.each(&:write)
     end if load_tap_migrations
 
-    # Migrate installed renamed formulae from main Homebrew repository.
-    if load_formula_renames
-      report.select_formula(:D).each do |oldname|
-        newname = FORMULA_RENAMES[oldname]
-        next unless newname
-        next unless (dir = HOMEBREW_CELLAR/oldname).directory? && !dir.subdirs.empty?
+    load_formula_renames
+    report.update_renamed
 
-        begin
-          migrator = Migrator.new(Formulary.factory("homebrew/homebrew/#{newname}"))
-          migrator.migrate
-        rescue Migrator::MigratorDifferentTapsError
-        end
+    # Migrate installed renamed formulae from core and taps.
+    report.select_formula(:R).each do |oldname, newname|
+      if oldname.include?("/")
+        user, repo, oldname = oldname.split("/", 3)
+        newname = newname.split("/", 3).last
+      else
+        user, repo = "homebrew", "homebrew"
       end
-    end
 
-    # Migrate installed renamed formulae from taps
-    report.select_formula(:D).each do |oldname|
-      user, repo, oldname = oldname.split("/", 3)
-      next unless user && repo && oldname
-      tap = Tap.new(user, repo)
-      next unless newname = tap.formula_renames[oldname]
       next unless (dir = HOMEBREW_CELLAR/oldname).directory? && !dir.subdirs.empty?
 
       begin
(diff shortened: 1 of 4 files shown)


Library/Homebrew/migrator.rb
```diff
@@ -55,7 +55,10 @@ def initialize(formula)
 
     @old_tabs = oldpath.subdirs.each.map { |d| Tab.for_keg(Keg.new(d)) }
     @old_tap = old_tabs.first.tap
-    raise MigratorDifferentTapsError.new(formula, old_tap) unless from_same_taps?
+
+    unless ARGV.force? || from_same_taps?
```
mikemcquaid: Mind using an `if` here instead; I always find `unless` with multiple arguments hard to parse mentally.

Library/Homebrew/cmd/update.rb
```diff
@@ -319,18 +319,51 @@ def empty?
...
+        if newpath
+          "#{tap}/#{path.basename(".rb")} -> #{tap}/#{newpath.basename(".rb")}"
+        else
+          "#{tap}/#{path.basename(".rb")}"
+        end
+      elsif newpath
+        "#{path.basename(".rb")} -> #{newpath.basename(".rb")}"
       else
-        path.basename(".rb").to_s
+        "#{path.basename(".rb")}"
```
mikemcquaid: Can leave this line as it was (`path.basename(".rb").to_s`)

Library/Homebrew/cmd/update.rb
```diff
@@ -319,18 +319,51 @@ def empty?
...
+    @hash[:A] -= @hash[:R].map(&:last) if @hash[:A]
+    @hash[:D] -= @hash[:R].map(&:first) if @hash[:D]
+  end
+
+  def select_formula(key)
+    fetch(key, []).map do |path, newpath|
+      if path.to_s =~ HOMEBREW_TAP_PATH_REGEX
+        tap = "#{$1}/#{$2.sub("homebrew-", "")}"
+        if newpath
+          "#{tap}/#{path.basename(".rb")} -> #{tap}/#{newpath.basename(".rb")}"
```
xu-cheng: I found it's strange to have a method named `select_formula` returning string which is not a formula name. I suggest you create a new method to do this.

xu-cheng:
You should also update https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/update.rb#L63-L91

It would be better if you can avoid duplicated computation.

Library/Homebrew/cmd/update.rb
vladshablinsky: @xu-cheng what if I make `select_formula` return `[oldname, newname]` for `:R` key and we will be able to use `select_formulae(:R)` when performing migrations.`select_formula` will still return strings that are formula names. For that we need to update `report` after all the pulls from remote repositories. For `dump_formula_report` we can check if key is `:R`.

vladshablinsky:
@xu-cheng I updated `select_formula` as I said in the [comment](https://github.com/Homebrew/homebrew/pull/42801#discussion_r36774696)

Now there is no duplicated computation when we perform migrations. `select_formula` now returns formula names, do we still need to add the method for renamed formulae?

Library/Homebrew/migrator.rb
```diff
@@ -55,7 +55,10 @@ def initialize(formula)
 
     @old_tabs = oldpath.subdirs.each.map { |d| Tab.for_keg(Keg.new(d)) }
     @old_tap = old_tabs.first.tap
-    raise MigratorDifferentTapsError.new(formula, old_tap) unless from_same_taps?
+
+    if !ARGV.force? && !from_same_taps?
+      raise MigratorDifferentTapsError.new(formula, old_tap)
```
xu-cheng: I think you need to update the message inside `MigratorDifferentTapsError`.

vladshablinsky: @xu-cheng updated.

Library/Homebrew/migrator.rb
```diff
@@ -42,6 +42,8 @@ def initialize(formula, tap)
         #{formula.name} from #{formula.tap} is given, but old name #{formula.oldname} was installed from #{tap}
 
         Please try to use #{name} to refer the formula
+
+        To force migrate use `brew migrate --force #{formula.oldname}`
```
xu-cheng: The output may need some improvement. It looks strange to have these empty lines. Maybe it's a good idea to eliminate the if-else branch as well.

In the mean time, the manpage need update as well.

PR #42805, opened 2015-08-11
test-bot: run `git gc` on cleanup.

mikemcquaid:
CC @xu-cheng @DomT4

### DIFF
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -575,6 +575,7 @@ module Homebrew
     def cleanup_before
       @category = __method__
       return unless ARGV.include? "--cleanup"
+      git "gc", "--auto"
       git "stash"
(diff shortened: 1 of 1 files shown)


xu-cheng:
:+1:

DomT4:
:+1: here as well. `git gc` can be kind of sluggish, but if the alternative is manual plumbing every couple days this is a nicer solution.

mikemcquaid:
> here as well. git gc can be kind of sluggish, but if the alternative is manual plumbing every couple days this is a nicer solution.

@DomT4 `--auto` means it'll only run if needed (i.e. relatively infrequently)

DomT4:
Yeah, I have the same setting run locally. I guess my point was more that the test-bots get to see more Git action than most machines, so it could run fairly frequently, but there's zero harm in trying and seeing how it goes.

PR #42811, opened 2015-08-11
python3 3.5rc1 (devel)

youtux:


tdsmith:
Pulled as 7aee609202; thanks!

PR #42817, opened 2015-08-11
viewvc 1.1.23 (new formula)

dlh:


Library/Formula/viewvc.rb
```diff
@@ -0,0 +1,35 @@
...
+  depends_on :python => :optional
+  if build.with? "python"
+    depends_on "subversion" => "with-python"
+  else
+    depends_on "subversion" => ["with-python", "universal"]
+  end
+
+  def install
+    system "./viewvc-install", "--prefix=#{libexec}", "--destdir="
+    (bin/"viewvc-standalone").write_env_script libexec/"bin/standalone.py",
```
DomT4: Can compress this into `bin.env_script_all_files(libexec, :PYTHONPATH => ENV["PYTHONPATH"])`

Library/Formula/viewvc.rb
```diff
@@ -0,0 +1,35 @@
...
+  desc "Browser interface for CVS and Subversion repositories"
+  homepage "http://www.viewvc.org"
+  url "http://viewvc.tigris.org/files/documents/3330/49392/viewvc-1.1.23.tar.gz"
+  sha256 "d14290b23f44e8f5301a7ca9fc946bcd0f8bb8206d9b9c111ac56e19f2588b7d"
+
+  depends_on :python => :optional
+  if build.with? "python"
+    depends_on "subversion" => "with-python"
+  else
+    depends_on "subversion" => ["with-python", "universal"]
```
mikemcquaid: Why does this need a universal Subversion?

mikemcquaid: Can't it use the system Subversion either?

DomT4: I don't quite get this block. If we're not building with Python, why would you need to enforce Subversion with Python?

Library/Formula/viewvc.rb
dlh: Updated—all the bin files get an `env_script` written. They have really generic names (`loginfo-handler`, `make-database`, `standalone.py`, etc…), so I prefixed them with `viewvc-`.

Library/Formula/viewvc.rb
dlh: Updated—

> I don't quite get this block. […]

Oops. It was meant to check if the user has Homebrew's python installed or if they're using the System python.

`Viewvc` needs the python bindings for subversion, which don't come installed with the System subversion.

I'd like to just say `depends_on "subversion" => "with-python"`, however [the subversion formula requires `--universal` to be used if you are using the System python](https://github.com/Homebrew/homebrew/blob/c8a8f325d37f68f9d64389a7e2776d66263f301e/Library/Formula/subversion.rb#L81).

dlh:
The failure from the bot is due to dependencies not being installed.

`swig` (and its dependency `pcre`) should be installed. It's a dependency of `subversion --with-python`. Issue #34554 had the same problem. See also issue #34576 for the work-a-round.

Why are the dependencies not being correctly computed? This seems like a bug in Homebrew.

Library/Formula/viewvc.rb
```diff
@@ -0,0 +1,36 @@
+class Viewvc < Formula
+  desc "Browser interface for CVS and Subversion repositories"
+  homepage "http://www.viewvc.org"
+  url "http://viewvc.tigris.org/files/documents/3330/49392/viewvc-1.1.23.tar.gz"
+  sha256 "d14290b23f44e8f5301a7ca9fc946bcd0f8bb8206d9b9c111ac56e19f2588b7d"
+
+  if Formula["python"].installed?
```
bfontaine: What happens if the user install `python`, then this formula, then remove `python`?

dlh: `viewvc-install` replaces the shebang lines in its bin scripts with whatever python it was invoked with. So, their installation of viewvc would break.

dunn:
Now subversion is failing to install due to sandboxing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/32095/version=mountain_lion/testReport/junit/brew-test-bot/mountain_lion/install___only_dependencies_viewvc/

DomT4:
I think `serf` is the failure there, which is joyous. Will look into it.

To be honest I'm pretty unhappy with the dependency logic in this formula. I think handling dependencies based on whether or not someone has something installed right at the moment is risky risky behaviour and will be prone to breakage.

If we're going to include this I'd much rather tie it to the system Python, which we do for almost all mandatory Python-using formula, and then make the dependency an unconditional `depends_on "subversion" => ["with-python", "universal"]`. If we have to enforce the universal aspect, so be it.

dlh:
Updated to only use the system python.

`audit` complains about the `swig` dependency, but I'm not sure if I should make the change to `depends_on "swig" => :run` to appease it. `:build` would make bottle installs fail (due to the linked dependency bug), and it's not really a runtime dependency.

The error:

    viewvc:
     * swig dependency should be
      depends_on "swig" => :build
    Or if it is indeed a runtime denpendency
      depends_on "swig" => :run

DomT4:
Well, `swig` isn't a `:build` dep in the `subversion` formula, so you wouldn't actually be adding any dependencies by passing `depends_on "swig" => :run`

dlh:
Ugh, [more build errors](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/32179/). `sqlite` needs to be built with `--universal`, too…

I'm just going to add an explicit dependency on Homebrew's python.

dlh:
Updated once again (with help from the fix in #43425)—now using the system python.

The test bot failed on the [mavericks configuration](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/32777/version=mavericks/console), due to an error from the `git checkout` step. I'm not sure why…

dunn:
That's just our bot acting up. I'll re-queue the build.

dunn:
Thank you for your patience, merged in 3bf878f5a6f7409a5e6bed258c6bdaf10c2adc07!

Library/Formula/viewvc.rb
```diff
@@ -0,0 +1,36 @@
...
+  sha256 "d14290b23f44e8f5301a7ca9fc946bcd0f8bb8206d9b9c111ac56e19f2588b7d"
+
+  # swig is a dependency of subversion --with-python, but due to a
+  # bug it needs to also be specified here.
+  # https://github.com/Homebrew/homebrew/issues/42915
+  depends_on "swig" => :run
+  depends_on "subversion" => "with-python"
+
+  def install
+    system "/usr/bin/python", "./viewvc-install", "--prefix=#{libexec}", "--destdir="
```
DomT4: Should `/usr/bin/python` have been hardcoded here?

dunn: @dlh is there a reason to rely on the system Python?

dunn: (Good catch, thanks Dom.

DomT4: It'll use the system Python regardless unless people install with `--env=std`, if that was the concern behind this particular line. Superenv should prevent Homebrew's Python from wandering into the `$PATH` during normal installs.

Library/Formula/viewvc.rb
```diff
@@ -0,0 +1,36 @@
...
+    Pathname.glob(libexec/"bin/*") do |f|
```
DomT4: As a random style nit that doesn't really matter at this point, this likely would have worked as:

```ruby
Dir[libexec/"bin/*"] do |f|
```
Rather than throwing out to `Pathname` directly.

PR #42827, opened 2015-08-11
Rack, a CLI for Rackspace

lvh:
Newly released. I'm a newbie Formula developer; this is only my third, and it's my first Go one.

Obvious potential review issues that I don't quite know if they're problems or not:

1. Requires `go get` to build.
2. Upstream is still calling it beta, although it is 1.0.0.

### DIFF
--- a/Library/Formula/rack.rb
+++ b/Library/Formula/rack.rb
@@ -1,0 +1,27 @@
+class Rack < Formula
+  desc "CLI for Rackspace"
+  homepage "https://github.com/rackspace/rack"
+  url "https://github.com/rackspace/rack.git",
+      :tag => "1.0.1",
+      :revision => "71a8d7c80b3652b4dd39e683bc423b8a542b0167"
+  head "https://github.com/rackspace/rack.git"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    ENV["TRAVIS_TAG"] = version
+
+    rackpath = buildpath/"src/github.com/rackspace/rack"
+    rackpath.install Dir["{*,.??*}"]
+
+    cd rackpath do
+      system "script/build", "rack"
+      bin.install "rack"
+    end
+  end
+
+  test do
+    system "#{bin}/rack"
+  end
+end


lvh:
I guess I should probably rebase this against master, since it has some random merge junk in it.

Library/Formula/rack.rb
```diff
@@ -0,0 +1,20 @@
...
+  homepage "https://github.com/rackspace/rack"
+  url "https://github.com/rackspace/rack/archive/1.0.0-beta.1.tar.gz"
+  version "1.0.0-beta.1"
+  sha256 "2ab9513fee4a23667e89c05548be26b4f15a387a8be1e1896936f6e253e05c51"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    system "go", "get", "-d"
```
mistydemeo: Instead of running `go get`, the resources should be specified as `go resource`s; take a look at terraform or packer for examples.

lvh: Thanks for your comment! It appears rack vendors all its deps with `govendor` in `internal/`; I am currently trying to figure out how to get `go build` to use them.

lvh: Sigh; `#go-nuts` was utterly unhelpful. I've asked `govendor` upstream how to do this.

dunn: Would it work to just do something like `ln_s Dir["internal/*"], buildpath`?

rgbkrk:
Is it appropriate to use one of the built go binaries that are up on https://developer.rackspace.com/docs/rack-cli/#quickstart ?

Library/Formula/rack.rb
lvh: Not quite; but manually reconstituting a GOPATH seems to work.

lvh:
It is my understanding that that is not OK by homebrew's rules, @rgbkrk.

dunn:
That's correct; we do provide our own pre-compiled binaries (the bottles) but we want to always offer users the ability to build the software from source.

rgbkrk:
I take it that means that bottles are typically provided ([as is done in the `docker` Formula](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/docker.rb#L12-L30)), but that building from source is appreciated.

mistydemeo:
That's right. The bottles are built in the same manner as building from source as a user.

Library/Formula/rack.rb
mistydemeo: Looks like this issue was fixed by a commit since the release: https://github.com/rackspace/rack/commit/833593b72c6e8ad0577844d480d75aa924744e25

mistydemeo:
Got this to work without `go get` by backporting that commit:

```patch
--- rack.old.rb	2015-08-11 21:56:49.000000000 -0700
+++ rack.rb	2015-08-11 21:57:45.000000000 -0700
@@ -7,11 +7,22 @@
 
   depends_on "go" => :build
 
+  # Fix import paths in the vendored code
+  patch do
+    url "https://github.com/rackspace/rack/commit/833593b72c6e8ad0577844d480d75aa924744e25.patch"
+    sha256 "3c3a79c17b563fbd3082e8d8876b920ab0163deda691778fe5466a37f2196048"
+  end
+
   def install
     ENV["GOPATH"] = buildpath
-    system "go", "get", "-d"
-    system "go", "build", "-o", "rack"
-    bin.install "rack"
+
+    rackpath = buildpath/"src/github.com/rackspace/rack"
+    rackpath.install Dir["{*,.git}"]
+
+    cd rackpath do
+      system "go", "build", "-o", "rack"
+      bin.install "rack"
+    end
   end
 
   test do
```

Library/Formula/rack.rb
```diff
@@ -0,0 +1,20 @@
+class Rack < Formula
+  desc "A CLI for Rackspace"
```
bfontaine: Please remove the indefinite article here (`a`).

Library/Formula/rack.rb
```diff
@@ -0,0 +1,20 @@
...
+  homepage "https://github.com/rackspace/rack"
+  url "https://github.com/rackspace/rack/archive/1.0.0-beta.1.tar.gz"
+  version "1.0.0-beta.1"
```
bfontaine: Is there a stable release we can use?

rgbkrk:
Cool, @jrperritt just fixed up that in https://github.com/rackspace/rack/commit/833593b72c6e8ad0577844d480d75aa924744e25 - that's the right commit.

lvh:
Thanks for all the help, @mistydemeo -- that appears to have fixed it. I've addressed all review comments and pushed commits to this branch.

lvh:
@bfontaine: I have removed the indefinite article. The beta release is current.

lvh:
Hm. One of the problems with patching manually is that upstream uses `git rev-parse` at compile time to figure out what version is being installed, but that suddenly becomes full of lies because it no longer matches the actual code being ran. I think the correct solution is "tag a new release"?

bfontaine:
@lvh Unfortunately, we don’t accept unstable versions (including alpha and beta ones) in the core. We’ll have to wait for a stable version.

viktorbenei:
@lvh for future reference, if you would need it, here's our Formula: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/bitrise.rb

We have a similar setup, in that our Go binary depends on two other Go projects (also maintained by us) and the Formula downloads the source and compiles it for each.

DomT4:
Feel free to submit this to `homebrew/devel-only` if you like, until something officially stable is tagged.

rgbkrk:
@lvh, 1.0.0 has been tagged and released!

lvh:
OK, I've updated the Formula, and it seems to work:

```
lvh@zygalski ~> brew install rack                                       11:11:33
Building from source because Library/Formula/rack.rb has local changes
==> Downloading https://github.com/rackspace/rack/archive/1.0.0.tar.gz
==> Downloading from https://codeload.github.com/rackspace/rack/tar.gz/1.0.0
==> go build -o rack
🍺  /usr/local/Cellar/rack/1.0.0: 2 files, 13M, built in 8 seconds
```

However, the version problem persists (it wants to be built from git checkout):

```
lvh@zygalski ~> rack version                                            11:11:44
rack version
commit:
```

It works fine, but that's not very good UX. @bfontaine: I'm assuming this isn't the first project like this; any idea how it was resolved elsewhere? @rgbkrk: Is there a way to sidestep that? It seems like it just really wants to call `git`.

dunn:
You could switch from using the archive tarball to checking out the git tag, like Docker does: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/docker.rb#L4-L5

lvh:
Thanks for the pointer! I've done this, but I'm still not seeing the version:

```
lvh@zygalski /u/l/Library (rack) [127]> brew install -v rack            13:15:33
Building from source because Library/Formula/rack.rb has local changes
==> Cloning https://github.com/rackspace/rack.git
git --git-dir /Library/Caches/Homebrew/rack--git/.git status -s
Updating /Library/Caches/Homebrew/rack--git
git config remote.origin.url https://github.com/rackspace/rack.git
git config remote.origin.fetch +refs/tags/1.0.0:refs/tags/1.0.0
git --git-dir /Library/Caches/Homebrew/rack--git/.git rev-parse -q --verify 1.0.0^{commit}
git checkout -f 1.0.0 --
HEAD is now at 11fec2b... Version 1.0.0
git reset --hard 1.0.0
HEAD is now at 11fec2b Version 1.0.0
==> Checking out tag 1.0.0
==> go build -o rack
==> Cleaning
==> Finishing up
ln -s ../Cellar/rack/1.0.0/bin/rack rack
==> Summary
🍺  /usr/local/Cellar/rack/1.0.0: 2 files, 13M, built in 4 seconds
lvh@zygalski /u/l/Library (rack)> rack                                  13:15:53
NAME:
   rack - Command-line interface to manage Rackspace Cloud resources
USAGE:
   rack <command> <subcommand> <action> [FLAGS]

VERSION:
   rack version
   commit:


COMMANDS:
   configure		Interactively create a config file for Rackspace authentication.
   init			Enable tab for command completion.
			For Linux and OS X, creates the `rack` man page and sets up
			command completion for the Bash shell. Run `man ./rack.1` to
			view the generated man page.
			For Windows, creates a `posh_autocomplete.ps1` file in the
			`$HOME/.rack` directory. You must run the file to set up
			command completion.
   version		Print the version of this binary.
   servers		Operations on cloud servers, both virtual and bare metal.
   files		Object storage for files and media.
   networks		Software-defined networking.
   block-storage	Block-level storage, exposed as volumes to mount to
			host servers. Work with volumes and their associated snapshots.
   help, h		Shows a list of commands or help for one command

lvh@zygalski /u/l/Library (rack)> rack version                          13:16:04
rack version
commit:
```

Hopefully @rgbkrk knows what's going on :)

jrperritt:
@lvh `go build` doesn't set the version and commit; `script/build` does though.

lvh:
@jrperritt Thanks! That gave me a warning, but I submitted a ticket for that: https://github.com/rackspace/rack/issues/339

rgbkrk:
Yeah, we set the version and commit at build time now as you pointed out in https://github.com/rackspace/rack/issues/339.

We could still be setting the Commit and Version when we tag a release, as right now they're undefined until build time. This was mostly to prevent every commit from setting the commit manually.

dunn:
If it's just a matter of adding flags to the `go build` command, we can replicate here what you do upstream.  Then we could switch back to using the tarball, right?

rgbkrk:
The only issue there is that use of the `=` as part of the parameters is not supported in `go < 1.5`, while it shows a pseudo-deprecation warning on go 1.5 builds. Can we assume it's always built with go 1.5 or does that depend on their homebrew installed version of `golang`?

rgbkrk:
I've put in a PR (https://github.com/rackspace/rack/pull/340) to choose what version of `LDFLAGS -X` parameters to use based on go version, since I don't see a way to suppress just this one warning at build time.

dunn:
Homebrew's Go is 1.5.1, so you can assume that, yes!

lvh:
@rgbkrk Looks like that patch still adds -dev to the end:

```
lvh@zygalski /u/l/L/Formula (rack)> brew install -v rack                                                                                                                                            10:04:56
Building from source because Library/Formula/rack.rb has local changes
==> Cloning https://github.com/rackspace/rack.git
git --git-dir /Library/Caches/Homebrew/rack--git/.git status -s
Updating /Library/Caches/Homebrew/rack--git
git config remote.origin.url https://github.com/rackspace/rack.git
git config remote.origin.fetch +refs/tags/1.0.1:refs/tags/1.0.1
git --git-dir /Library/Caches/Homebrew/rack--git/.git rev-parse -q --verify 1.0.1^{commit}
git checkout -f 1.0.1 --
HEAD is now at 71a8d7c... Version 1.0.1
git reset --hard 1.0.1
HEAD is now at 71a8d7c Version 1.0.1
==> Checking out tag 1.0.1
==> script/build rack
Building rack!
==> Cleaning
==> Finishing up
ln -s ../Cellar/rack/1.0.1/bin/rack rack
==> Summary
🍺  /usr/local/Cellar/rack/1.0.1: 2 files, 13M, built in 4 seconds
lvh@zygalski /u/l/L/Formula (rack)> ../../Cellar/rack/1.0.1/bin/rack                                                                                                                                10:05:04
NAME:
   rack - Command-line interface to manage Rackspace Cloud resources
USAGE:
   rack <command> <subcommand> <action> [FLAGS]

VERSION:
   rack version 1.0.1-dev
   commit: 71a8d7c80b3652b4dd39e683bc423b8a542b0167


COMMANDS:
   configure		Interactively create a config file for Rackspace authentication.
   init			Enable tab for command completion.
			For Linux and OS X, creates the `rack` man page and sets up
			command completion for the Bash shell. Run `man ./rack.1` to
			view the generated man page.
			For Windows, creates a `posh_autocomplete.ps1` file in the
			`$HOME/.rack` directory. You must run the file to set up
			command completion.
   version		Print the version of this binary.
   servers		Operations on cloud servers, both virtual and bare metal.
   files		Object storage for files and media.
   networks		Software-defined networking.
   block-storage	Block-level storage, exposed as volumes to mount to
			host servers. Work with volumes and their associated snapshots.
   help, h		Shows a list of commands or help for one command
```

DomT4:
Status update on this?

lvh:
Whoops, I completely dropped the ball on this.

lvh:
After talking to @rgbkrk we're going to just go with the one that appends -dev. I'll fix the CI.

lvh:
Travis gave up:

```
The job exceeded the maxmimum time limit for jobs, and has been terminated.
```

Build log is __huge__, unsure why it took this long. Might just be intermittent.

lvh:
Whoa. My upstream branch had more than I bargained for.

rgbkrk:
Yeah, I'd probably rebase this.

rgbkrk:
I've figured out why @lvh. If you want to see what's happened to the local copy of the repo from homebrew, run

```ruby
puts `git status`
```

and you'll see that the dotfiles for the repo aren't there

```
Not currently on any branch.
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	deleted:    .dockerignore
	deleted:    .gitignore
	deleted:    .travis.yml

no changes added to commit (use "git add" and/or "git commit -a")
```

resulting in this being a dirty working copy

lvh:
Good catch! Why are those files missing, though?!

rgbkrk:
No clue, because just prior the working directory is clean and just fine.

rgbkrk:
I wonder if it's the matter of copying the directory within homebrew but missing all the dotfiles...

lvh:
Huh. This sounds like something maybe one of the homebrew folks would know :)

mikemcquaid:
Sorry, I'm not sure what the question is?

lvh:
@mikemcquaid Is there a reason why dotfiles would be missing from the build directory? I'm guessing it might be `Dir["{*,.git}"]`; I'll add the other dotfiles as well.

lvh:
Of course, the obvious `Dir[*,.*]` doesn't work since that includes the directory itself... Fail fail.

```
lvh@zygalski /u/l/L/Formula (rack)> brew install rack                   16:35:52
==> Cloning https://github.com/rackspace/rack.git
Cloning into '/Library/Caches/Homebrew/rack--git'...
remote: Counting objects: 649, done.
remote: Compressing objects: 100% (524/524), done.
remote: Total 649 (delta 200), reused 363 (delta 98), pack-reused 0
Receiving objects: 100% (649/649), 448.57 KiB | 0 bytes/s, done.
Resolving deltas: 100% (200/200), done.
Checking connectivity... done.
Note: checking out '71a8d7c80b3652b4dd39e683bc423b8a542b0167'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

==> Checking out tag 1.0.1
Error: File exists - /private/tmp/rack20151027-68335-1d06n4g/src/github.com/rackspace/rack
```

I've managed to work around that, but I'd love an opinion from someone who speaks Ruby better than I do.

rgbkrk:
It might be better to use `version` like I'm doing in https://github.com/Homebrew/homebrew/pull/45399.

lvh:
Dear homebrew folks: for clarity; that's a suggestion, and I think this might be ready for inclusion right now.

Library/Formula/rack.rb
```diff
@@ -0,0 +1,27 @@
...
+    rackpath.install (Dir["{*,.*}"] - [".", ".."])
+
+    cd rackpath do
+      system "script/build", "rack"
+      bin.install "rack"
+    end
+  end
+
+  test do
+    system "#{bin}/rack"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/rack.rb
```diff
@@ -0,0 +1,27 @@
...
+  head "https://github.com/rackspace/rack.git"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    ENV["TRAVIS_TAG"] = @tag
+
+    rackpath = buildpath/"src/github.com/rackspace/rack"
+    rackpath.install (Dir["{*,.*}"] - [".", ".."])
```
mikemcquaid: Maybe just `Dir["{*,.??*}"]`?

Library/Formula/rack.rb
```diff
@@ -0,0 +1,27 @@
...
+  url "https://github.com/rackspace/rack.git",
+      :tag => "1.0.1",
+      :revision => "71a8d7c80b3652b4dd39e683bc423b8a542b0167"
+  head "https://github.com/rackspace/rack.git"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    ENV["TRAVIS_TAG"] = @tag
```
mikemcquaid: Don't rely on this being set; use `version` instead.

Library/Formula/rack.rb
lvh: I'd love to, but I'm not sure we can. This is a tool for accessing a public API; is doing that considered acceptable? (Even if it is, I'm not sure said API does anything useful before asking for authentication; although I suppose checking that you can even authenticate is already better than `--version`.)

lvh:
New version appears to work correctly. Thanks @mikemcquaid and @rgbkrk!

```
lvh@zygalski /u/l/L/Formula (rack)> rack                                                                                                                                                                                                                                10:54:04
NAME:
   rack - Command-line interface to manage Rackspace Cloud resources
USAGE:
   rack <command> <subcommand> <action> [FLAGS]

VERSION:
   rack version 1.0.1
   commit: 71a8d7c80b3652b4dd39e683bc423b8a542b0167


COMMANDS:
   configure		Interactively create a config file for Rackspace authentication.
   init			Enable tab for command completion.
			For Linux and OS X, creates the `rack` man page and sets up
			command completion for the Bash shell. Run `man ./rack.1` to
			view the generated man page.
			For Windows, creates a `posh_autocomplete.ps1` file in the
			`$HOME/.rack` directory. You must run the file to set up
			command completion.
   version		Print the version of this binary.
   servers		Operations on cloud servers, both virtual and bare metal.
   files		Object storage for files and media.
   networks		Software-defined networking.
   block-storage	Block-level storage, exposed as volumes to mount to
			host servers. Work with volumes and their associated snapshots.
   help, h		Shows a list of commands or help for one command
```

Library/Formula/rack.rb
mikemcquaid: That's fine if you can do any non-authed operations. Otherwise it's fine as-is.

lvh: Alas, the only thing I can do is install bash completions and a man page (but not separately yet, which is why I'm not adding it to this PR yet), and that doesn't sound like a *test*. Sorry :(

rgbkrk:
@lvh - you may want to rebase down to one commit with a commit message of "rack 1.0.1 (new formula)", [per the Formula Cookbook](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#commit).

lvh:
Done :)

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/5a53c146d4a4125098c1a31d0702a336f0054ac2. Thank you for your contribution to Homebrew @lvh; we appreciate it! :smiley_cat:

rgbkrk:
:tada: Thanks!

PR #42828, opened 2015-08-11
Don't explicitly enable C++11 for ZNC formula.

### LINKED ISSUE #42822: ZNC formula: issue with C++11
ZNC 1.6.0 requires compiler with C++11 support.
Because of that https://github.com/Homebrew/homebrew/blob/33aa9a09051c1260169f2de4e23834d68478ede3/Library/Formula/znc.rb#L31 explicitly adds `-std=c++11` to `CXXFLAGS`.
`./configure` thinks that no additional flag is needed for C++11, and as a result, `znc-buildmod` on the target system is broken.

See znc/znc#1023

Is there any side-effect, if to just remove `ENV.cxx11` from the formula?

DarthGandalf:
Let ./configure to figure it out.
Fix #42822

### DIFF
--- a/Library/Formula/znc.rb
+++ b/Library/Formula/znc.rb
@@ -28,7 +28,6 @@ class Znc < Formula
   depends_on "icu4c" => :optional
 
   def install
-    ENV.cxx11
     args = ["--prefix=#{prefix}"]
     args << "--enable-debug" if build.with? "debug"
 


mistydemeo:
Like I mentioned in #42828, we added this when upgrading to this version of ZNC because it wouldn't build without on 10.8.

DarthGandalf:
The link to logs in #42828 doesn't work... Let's build bot to make new build and see what was the error. If this doesn't help, we can come up with some other solution...

mistydemeo:
After some testing, I think an alternate solution is to keep `ENV.cxx11` while also explicitly setting the stdlib and std in `CXXFLAGS`:

```patch
--- a/Library/Formula/znc.rb
+++ b/Library/Formula/znc.rb
@@ -28,6 +28,9 @@ class Znc < Formula
   depends_on "icu4c" => :optional

   def install
+    ENV.append_to_cflags "-std=c++11"
+    ENV.append_to_cflags "-stdlib=libc++" if ENV.compiler == :clang
+
     ENV.cxx11
     args = ["--prefix=#{prefix}"]
     args << "--enable-debug" if build.with? "debug"
```

This allows ZNC to build on 10.8 for me, and I'm able to build a module like znc-palaver.

DarthGandalf:
When I run it locally, either removing `ENV.cxx11` and adding those 2 lines fixes znc-buildmod.

DarthGandalf:
Let's go with your solution then.
Thanks!

PR #42829, opened 2015-08-11
elftoolchain 0.6.1 (new formula)

jwatzman:
Add the elftoolchain formula, a set of tools and libraries for handling
the ELF file format. This notably includes an implementation of
libdwarf, which is generally annoying to get on OS X, and which is
generally useful. (It's a dependency of HHVM, which I intend to package
next.)

### DIFF
--- a/Library/Formula/elftoolchain.rb
+++ b/Library/Formula/elftoolchain.rb
@@ -1,0 +1,69 @@
+class Elftoolchain < Formula
+  desc "Compilation tools and libaries for handling the ELF file format"
+
+  homepage "http://elftoolchain.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/elftoolchain/Sources/elftoolchain-0.6.1/elftoolchain-0.6.1.tgz"
+  sha256 "f4dc6e2a820f146658d6b5f9a062a2e676d0715aca654f55866c60e0025561eb"
+
+  depends_on :bsdmake => :build
+  depends_on "libarchive"
+
+  keg_only <<-EOS.undent
+    The tools provided conflict with tool names from the normal toolchain, such
+    as nm, and are not replacements for them.
+  EOS
+
+  # elftoolchain is in general pretty unaware of OS X / Darwin. The next 4
+  # patches are pulled in from MacPorts to, e.g., make it aware of how to do
+  # byte swapping.
+  patch :p0 do
+    url "https://trac.macports.org/browser/trunk/dports/devel/elftoolchain/files/patch-common-elftc.diff?rev=98117&format=raw"
+    sha256 "e00505d7ffc62df5e20cb285ae4b27d3ab3c8948fe79d0f5637bed2326400214"
+  end
+
+  patch :p0 do
+    url "https://trac.macports.org/browser/trunk/dports/devel/elftoolchain/files/patch-byteorder-macros.diff?rev=98117&format=raw"
+    sha256 "a46bd66f14723aebba84932a295626374d314d1afe3ba684be7d0661c366c9de"
+  end
+
+  patch :p0 do
+    url "https://trac.macports.org/browser/trunk/dports/devel/elftoolchain/files/patch-libelf-config.diff?rev=98117&format=raw"
+    sha256 "a5e76d9383c70e870ff32bfd7fc52c8c4cdeca04410ee2f823d11bd9aba26c58"
+  end
+
+  patch :p0 do
+    url "https://trac.macports.org/browser/trunk/dports/devel/elftoolchain/files/patch-mk.diff?rev=98117&format=raw"
+    sha256 "aa7668354ae13930fde1c3c7acae98d9098a9356b998801eb65e2e4323b58cb2"
+  end
+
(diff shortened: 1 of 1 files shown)


Library/Formula/elftoolchain.rb
```diff
@@ -0,0 +1,69 @@
+class Elftoolchain < Formula
+  desc "Compilation tools and libaries for handling the ELF file format"
+
```
DomT4: No empty line necessary here.

Library/Formula/elftoolchain.rb
```diff
@@ -0,0 +1,69 @@
...
+  homepage "http://elftoolchain.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/elftoolchain/Sources/elftoolchain-0.6.1/elftoolchain-0.6.1.tgz"
+  sha256 "f4dc6e2a820f146658d6b5f9a062a2e676d0715aca654f55866c60e0025561eb"
+
+  depends_on :bsdmake => :build
```
DomT4: `"bsdmake"` - It's not a special dependency, just a normal one.

Library/Formula/elftoolchain.rb
```diff
@@ -0,0 +1,69 @@
...
+  depends_on "libarchive"
+
+  keg_only <<-EOS.undent
+    The tools provided conflict with tool names from the normal toolchain, such
+    as nm, and are not replacements for them.
+  EOS
+
+  # elftoolchain is in general pretty unaware of OS X / Darwin. The next 4
```
DomT4: We're going to need these patches reported upstream and those upstream bug reports linked in the formula. We rarely accept heavily patched new submissions as we end up carrying and supporting patches indefinitely otherwise.

Library/Formula/elftoolchain.rb
```diff
@@ -0,0 +1,69 @@
...
+      INCSDIR=#{include}
+      SHAREDIR=#{share}
+      BINOWN=#{ENV["USER"]}
+      SHAREOWN=#{ENV["USER"]}
+      STRIP=
+    ]
+
+    system "bsdmake", "MKTEX=no", *args
+
+    bin.mkpath
```
DomT4: It won't create *any* of these directories?

Library/Formula/elftoolchain.rb
```diff
@@ -0,0 +1,69 @@
...
+    inreplace "mk/elftoolchain.prog.mk", "@PREFIX@", prefix
+
+    args = %W[
+      prefix=#{prefix}
+      BINDIR=#{bin}
+      LIBDIR=#{lib}
+      SHLIBDIR=#{lib}
+      INCSDIR=#{include}
+      SHAREDIR=#{share}
+      BINOWN=#{ENV["USER"]}
```
DomT4: I'm not sure this is going to work as you hope. Given we'll likely add bottles for this formula, the `ENV["USER"]` call will end up being `brew` rather than the actual user for anyone who installs those bottles AFAIK.

Library/Formula/elftoolchain.rb
jwatzman: I can triple-check, but yeah, not only will it not create the directories, but it will install some file with the name of the directory if it isn't there. :( The install step for this package is pretty ornery.

Library/Formula/elftoolchain.rb
jwatzman: Pretty sure this will be fine in a bottle. This is actually working around what might be a bug in the bsdmake package. Whatever user bsdmake was built with is hardcoded in its config files as the default value of `BINOWN` -- namely `brew`. So when elftoolkit tries to `install` its files, passes `-o $(BINOWN)`, which expands to `-o brew`, which of course doesn't work. This is just overriding that back to a reasonable value -- the package doesn't care about the actual owner, we just need one that *works*. So when this gets packaged up as a bottle, and then expanded on another system, that user will then own the files, and that's perfectly fine.

I had a comment to this effect in the formula at some point, looks like I lost it when I rearranged things -- will add it back since this is pretty subtle.

Library/Formula/elftoolchain.rb
jwatzman: OK -- I thought since MacPorts had the patches, it would be fine. Totally understand not wanting to bring the work on yourself. I will go through these more carefully, and find which ones are really strictly necessary, and file upstream bugs about those (or find the relevant bug and link it in a comment if someone already filed it).

jwatzman:
Thanks for the review! I have some work to do here, and will update the patch when the issues you pointed out are resolved.

jwatzman:
I'm going to put this on hold for now, in favor of https://github.com/Homebrew/homebrew/pull/42869 which provides similar functionality to what I'm looking for (an implementation of libdwarf) but with dramatically less patching, and with a much better maintained upstream. I was under the impression that that one didn't work on OS X, but it appears to work just fine.

Thanks for the review here again, and sorry if I wasted your time. I don't expect to come back to this, but if I do, I'll reopen it and make the changes you requested.

PR #42830, opened 2015-08-11
New formula for I2P

devsli:
Comparing to original PR #12613:
* Version bump
* Minor fixes

### DIFF
--- a/Library/Formula/i2p.rb
+++ b/Library/Formula/i2p.rb
@@ -1,0 +1,38 @@
+class I2p < Formula
+  desc "I2P is an anonymous overlay network - a network within a network"
+  homepage "https://geti2p.net"
+  url "https://download.i2p2.de/releases/0.9.21/i2pinstall_0.9.21.jar"
+  sha256 "0238ffc6ea44099ef4fe6c20913cb4eec675c2760aea07dfe7d499addcc89cf2"
+
+  depends_on :java => "1.6+"
+
+  def install
+    (buildpath/"path.conf").write "INSTALL_PATH=#{libexec}"
+
+    system "java", "-jar", "i2pinstall_#{version}.jar", "-options", "path.conf"
+
+    wrapper_name = "i2psvc-macosx-universal-#{MacOS.prefer_64_bit? ? 64 : 32}"
+    libexec.install_symlink libexec/wrapper_name => "i2psvc"
+    bin.write_exec_script Dir["#{libexec}/{eepget,i2prouter}"]
+    man1.install Dir["#{libexec}/man/*"]
+  end
+
+  test do
+    wrapper_pid = fork do
+      exec "#{bin}/i2prouter console"
+    end
+    router_pid = 0
+    sleep 5
+
+    begin
+      status = shell_output("#{bin}/i2prouter status")
+      assert_match(/I2P Service is running/, status)
+      /PID:(\d+)/ =~ status
+      router_pid = Regexp.last_match(1)
+    ensure
+      Process.kill("SIGINT", router_pid.to_i) unless router_pid.nil?
+      Process.kill("SIGINT", wrapper_pid)
+      Process.wait(wrapper_pid)
+    end
+  end
+end


Library/Formula/i2p.rb
```diff
@@ -0,0 +1,25 @@
...
+  sha256 "0238ffc6ea44099ef4fe6c20913cb4eec675c2760aea07dfe7d499addcc89cf2"
+
+  depends_on :java => "1.6+"
+
+  def install
+    open("i2p_template.conf", "w") do |f|
+      f.write("INSTALL_PATH=#{prefix}")
+    end
+
+    system "java -jar i2pinstall_0.9.21.jar -options i2p_template.conf"
```
DomT4: Please use commas to separate each command, such as:

```ruby
system "java", "-jar", "i2pinstall_#{version}.jar", "-options", "i2p_template.conf"
```

Library/Formula/i2p.rb
```diff
@@ -0,0 +1,25 @@
...
+  desc "I2P is an anonymous overlay network - a network within a network"
+  homepage "https://geti2p.net"
+  url "https://download.i2p2.de/releases/0.9.21/i2pinstall_0.9.21.jar"
+  sha256 "0238ffc6ea44099ef4fe6c20913cb4eec675c2760aea07dfe7d499addcc89cf2"
+
+  depends_on :java => "1.6+"
+
+  def install
+    open("i2p_template.conf", "w") do |f|
+      f.write("INSTALL_PATH=#{prefix}")
```
DomT4: It may be a better option to install to `libexec`. We tend to keep `prefix` clear of things that aren't README/License/other folders/etc.

Library/Formula/i2p.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    system "java -jar i2pinstall_0.9.21.jar -options i2p_template.conf"
+    system "ln -s #{prefix}/i2psvc-macosx-universal-64 #{prefix}/i2psvc"
```
DomT4: You don't need to call `system` on any of these. Ruby's FileUtils can handle `ln`, `mv`, `mkdir` and so on. You can actually avoid them all together by doing something like:

```ruby
libexec.install_symlink libexec/"i2psvc-macosx-universal-64" => "i2psvc"
```

Library/Formula/i2p.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    system "mkdir -p #{prefix}/bin"
+    system "cp #{prefix}/eepget #{prefix}/bin/"
+    system "cp #{prefix}/i2prouter #{prefix}/bin/"
```
DomT4: You can combine these three lines into `bin.install libexec/"eepget", libexec/"i2prouter"` or even:

```ruby
bin.write_exec_script Dir["#{libexec}/*"].select { |f| File.executable? f }
```

Library/Formula/i2p.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    system "mv #{prefix}/man #{prefix}/man1"
+    system "mkdir -p #{prefix}/share/man"
+    system "mv #{prefix}/man1 #{prefix}/share/man/"
```
DomT4: And you can combine these three lines into:

```ruby
man1.install Dir["#{libexec}/share/man/man1/*"]
```

devsli:
Thank you Dominyk! I have updated formula.

Library/Formula/i2p.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  depends_on :java => "1.6+"
+
+  def install
+    open("path.conf", "w") do |f|
+      f.write("INSTALL_PATH=#{libexec}")
+    end
+
+    system "java", "-jar", "i2pinstall_#{version}.jar", "-options", "path.conf"
+    libexec.install_symlink libexec/"i2psvc-macosx-universal-64" => "i2psvc"
```
devsli: Should I use `Hardware::CPU.is_64_bit?` to separate 64-bit wrapper from 32-bit?

This feature seems deprecated according to `homebrew/Library/Homebrew/hardware.rb`:
```ruby
def bits
  64
end
```

mikemcquaid: Use `if MacOS.prefer_64_bit?`

devsli: Thank you, Mike!

Library/Formula/i2p.rb
```diff
@@ -0,0 +1,30 @@
+class I2p < Formula
+  desc "I2P is an anonymous overlay network - a network within a network"
+  homepage "https://geti2p.net"
+  url "https://download.i2p2.de/releases/0.9.21/i2pinstall_0.9.21.jar"
+  sha256 "0238ffc6ea44099ef4fe6c20913cb4eec675c2760aea07dfe7d499addcc89cf2"
+
+  depends_on :java => "1.6+"
+
+  def install
+    open("path.conf", "w") do |f|
```
mikemcquaid: `(buildpath/"path.conf").write "INSTALL_PATH=#{libexec}"`

Library/Formula/i2p.rb
```diff
@@ -0,0 +1,30 @@
...
+      f.write("INSTALL_PATH=#{libexec}")
+    end
+
+    system "java", "-jar", "i2pinstall_#{version}.jar", "-options", "path.conf"
+    wrapper_name = "i2psvc-macosx-universal-32"
+    if MacOS.prefer_64_bit?
+      wrapper_name = "i2psvc-macosx-universal-64"
+    end
```
mikemcquaid: `wrapper_name = MacOS.prefer_64_bit? ? "i2psvc-macosx-universal-64" : "i2psvc-macosx-universal-32"`

devsli: Thanks, @mikemcquaid! I'll shorten it

Library/Formula/i2p.rb
devsli: Accidentally I've missed this comment :disappointed: I'll fix it within 8 hours

Library/Formula/i2p.rb
```diff
@@ -0,0 +1,24 @@
...
+    libexec.install_symlink libexec/wrapper_name => "i2psvc"
+    bin.write_exec_script Dir["#{libexec}/{eepget,i2prouter}"]
+    man1.install Dir["#{libexec}/man/*"]
+  end
+
+  test do
+    system bin/"i2prouter", "start"
+    sleep 5
+    assert_match(/I2P Service is running/, shell_output("#{bin}/i2prouter status"))
+    system bin/"i2prouter", "stop"
```
DomT4: Let's try and use something like [this](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mpd.rb#L127-L139). At the moment with the current test if there's a problem on the `assert_match` it'll bail out and leave `i2prouter` running which isn't desirable.

Library/Formula/i2p.rb
```diff
@@ -0,0 +1,24 @@
...
+  url "https://download.i2p2.de/releases/0.9.21/i2pinstall_0.9.21.jar"
+  sha256 "0238ffc6ea44099ef4fe6c20913cb4eec675c2760aea07dfe7d499addcc89cf2"
+
+  depends_on :java => "1.6+"
+
+  def install
+    (buildpath/"path.conf").write "INSTALL_PATH=#{libexec}"
+    system "java", "-jar", "i2pinstall_#{version}.jar", "-options", "path.conf"
+    wrapper_name = "i2psvc-macosx-universal-#{MacOS.prefer_64_bit? ? '64' : '32'}"
+    libexec.install_symlink libexec/wrapper_name => "i2psvc"
```
DomT4: It's a nit, but if we could leave one blank space between the compile stage and the installation stage that'd be nice. Having chunked together `def install` blocks makes them a little unpleasant to read.

Library/Formula/i2p.rb
```diff
@@ -0,0 +1,24 @@
...
+  homepage "https://geti2p.net"
+  url "https://download.i2p2.de/releases/0.9.21/i2pinstall_0.9.21.jar"
+  sha256 "0238ffc6ea44099ef4fe6c20913cb4eec675c2760aea07dfe7d499addcc89cf2"
+
+  depends_on :java => "1.6+"
+
+  def install
+    (buildpath/"path.conf").write "INSTALL_PATH=#{libexec}"
+    system "java", "-jar", "i2pinstall_#{version}.jar", "-options", "path.conf"
+    wrapper_name = "i2psvc-macosx-universal-#{MacOS.prefer_64_bit? ? '64' : '32'}"
```
DomT4: Our CI will stop shouting at you if you make the single quotes here double quotes :). You may need to escape them.

Library/Formula/i2p.rb
devsli: Suggested method failed to kill i2prouter :) I'll fix within 12 hours

Library/Formula/i2p.rb
devsli: I'll replace it with

```ruby
wrapper_name = "i2psvc-macosx-universal-#{MacOS.prefer_64_bit? ? 64 : 32}"
```

Library/Formula/i2p.rb
devsli: I'll split it:

1. write config
2. install i2p
3. symlinks

Library/Formula/i2p.rb
devsli: i2prouter forked with PID (for example) 7440. When 7440 killed, children (7506, 7508) continue to work as if nothing had happened.

I can kill whole PGID group, but the problem is `ruby` process have same `pgid`.

Seems like only correct way is find out `wrapper pid` (it's reported by `i2prouter status`) and kill it.

```
  PID  PGID TTY          TIME COMMAND
  ...
 7438  7438 pts/3    00:00:00 ruby        # <- oh hai!
 7440  7438 pts/3    00:00:00 i2prouter   # <- kill doesn't affect children
 7506  7438 pts/3    00:00:00 wrapper     # <- should be killed
 7508  7438 pts/3    00:00:05 java
```

DomT4: Hmm. That's a pain. Perhaps we can use an `ensure` block on `system bin/"i2prouter", "stop"` or something. Not sure if that would work.

devsli: [Too awful?](https://github.com/devsli/homebrew/blob/15ed53cd076e23c8ff9be5adf22dde7110e3737b/Library/Formula/i2p.rb#L22-L30)

devsli:
```
==> brew readall							 FAILED
Error: problem in /usr/local/Library/Formula/i2p.rb
compile error
/usr/local/Library/Formula/i2p.rb:28: undefined (?...) sequence: /PID:(?<router_pid>\d+)/
==> brew uses i2p							Error: compile error
/usr/local/Library/Formula/i2p.rb:28: undefined (?...) sequence: /PID:(?<router_pid>\d+)/
 FAILED
Error: compile error
/usr/local/Library/Formula/i2p.rb:28: undefined (?...) sequence: /PID:(?<router_pid>\d+)/
```

Huh... What's wrong with my regexp?

bfontaine:
You can use `(…)` and then refer to the matched group with `$1`.

Library/Formula/i2p.rb
DomT4: If it works I have no objection to the proposed fix.

devsli:
Thanks, @bfontaine! Let's try to avoid named groups...

devsli:
Squashed pull request into single commit

mikemcquaid:

For future reference the preferred commit message format for new formulae is `i2p 0.9.21 (new formula)`.

PR #42831, opened 2015-08-11
jemalloc: remove prefix

jwatzman:
This option controls what prefix, if any, should be added to all the
function names in the library. The default on platforms other than OS X
is to not use a prefix, but for whatever reason, one is added on OS X.
This change overrides that back to the default on other platforms, which
is necessary for any code using jemalloc which wants to be portable
between OS X and Linux.

(MacPorts does this too, FWIW.)

### DIFF
--- a/Library/Formula/jemalloc.rb
+++ b/Library/Formula/jemalloc.rb
@@ -4,6 +4,7 @@ class Jemalloc < Formula
   url "http://www.canonware.com/download/jemalloc/jemalloc-3.6.0.tar.bz2"
   sha256 "e16c2159dd3c81ca2dc3b5c9ef0d43e1f2f45b04548f42db12e7c12d7bdf84fe"
   head "https://github.com/jemalloc/jemalloc.git"
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Library/Formula/jemalloc.rb
```diff
@@ -4,6 +4,7 @@ class Jemalloc < Formula
   url "http://www.canonware.com/download/jemalloc/jemalloc-3.6.0.tar.bz2"
   sha256 "e16c2159dd3c81ca2dc3b5c9ef0d43e1f2f45b04548f42db12e7c12d7bdf84fe"
   head "https://github.com/jemalloc/jemalloc.git"
+  revision 2
```
DomT4: You've bumped the revision too hard here. The default revision is `0` and is unwritten, so when bumping it gets increased to `1` rather than `2`.

mistydemeo:
Can you bump the revisions on the software that uses jemalloc, too?

Library/Formula/jemalloc.rb
jwatzman: Oops! Thanks, will fix.

jwatzman:
> Can you bump the revisions on the software that uses jemalloc, too?

I'm not sure what the homebrew policy is on doing this :) All programs which use jemalloc and work fine today should still be fine either way, without a recompile. Is the idea to actually make sure of this, or is something else going on? I'm happy to do it if needed, just want to make sure I don't do it unnecessarily :)

jwatzman:
Also, would you want the bumps of the revisions of formulas using jemalloc in a separate commit, or all in one commit with this update, or each bump its own commit?

DomT4:
If they're necessary, one formula change per commit is preferred. Something like: `r3: bump revision for jemalloc` is fine.

mikemcquaid:

PR #42839, opened 2015-08-12
caveats.rb: don't print elisp_caveats for keg-only

dunn:
Since they won't have been installed into `#{HOMEBREW_PREFIX}/share/emacs/site-lisp/`


mikemcquaid:
:+1:

PR #42842, opened 2015-08-12
packer 0.8.5

stuwil:


stuwil:
Looks like this is running out of disk space on mavericks agent only, while copying the binaries (~420MB) around within the buildpath.

From the upstream build script:
 
```
# Copy our OS/Arch to the bin/ directory
echo "==> Copying binaries for this platform..."
DEV_PLATFORM="./pkg/$(go env GOOS)_$(go env GOARCH)"
for F in $(find ${DEV_PLATFORM} -mindepth 1 -maxdepth 1 -type f); do
    cp ${F} bin/
    cp ${F} ${MAIN_GOPATH}/bin/
done
```

Any objections if I patch it to bin.install directly from the /pkg output?

mikemcquaid:
> Any objections if I patch it to bin.install directly from the /pkg output?

Yes. We'll fix the Mavericks issue.

stuwil:
Wonderful, didn't realize it was a new issue there. Thanks!

mistydemeo:
Thanks!

Issue #42349, opened 2015-08-02
Chromium CLI


dardo82:
Could we add chromium-cli (that's available [here] (https://github.com/prasmussen/chrome-cli/issues/31)) to the keg of chrome-cli (alongside chrome-canary-cli)?

mikemcquaid:
Could you try and open a pull request? This should help and I'm happy to walk you through anything else: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged

Thanks!

dardo82:
I'm not able to build the xcode project from the source (see [here] (https://github.com/prasmussen/chrome-cli/issues/31) why)
and I'm not sure how I should handle the binary tarball that homebrew
seems to use available from bintray.com to add the executable to it.

mikemcquaid:
The binary tarball is generated from the Homebrew formula. You'll need to edit that and submit a pull request to update it.

dardo82:
Ah, it's compiled on some smart cloud build server where the process doesn't fail miserably as it does on my machine? That explains something and I could try to edit the formula adding a part for compiling chromium-cli(the same way it's done for chrome-canary-cli)hoping nothing goes wrong as I would not be able to test it.

dunn:
If you want to take a stab at a PR, that'd be great!

dardo82:
I tried but i failed, let's retry... :-|

Issue #42848, opened 2015-08-12
Chromium CLI (again)


dardo82:
I've made a mistake in patching the formula for chrome-cli to add support for Chromium to the command.
(See PR https://github.com/Homebrew/homebrew/pull/42731) I am going to submit a new PR to fix it, ok?

mikemcquaid:
Yeh, you don't need to create an issue before the PR in future though, thanks!

dardo82:
Understood, didn't knew that... I will add it to my github experience... :-)

PR #42849, opened 2015-08-12
chrome-cli: REALLY add support for Chromium

dardo82:
I have made a BIG mistake in the previous PR: once the app id "com.google.Chrome" gets replaced with "com.google.Chrome.canary", you should replace that with "org.Chromium.chromium", not the other one.
That should fix https://github.com/Homebrew/homebrew/issues/42349 and https://github.com/Homebrew/homebrew/issues/42848, they are the same.

### DIFF
--- a/Library/Formula/chrome-cli.rb
+++ b/Library/Formula/chrome-cli.rb
@@ -31,7 +31,7 @@ class ChromeCli < Formula
     # Chromium builds; see:
     # https://github.com/prasmussen/chrome-cli/issues/31
     rm_rf "build"
-    inreplace "chrome-cli/App.m", "com.google.Chrome", "org.Chromium.chromium"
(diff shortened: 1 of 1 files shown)


bfontaine:
Note that you can (force-)push on your branch to edit your PR instead of creating a new one.

dardo82:
I am sorry if I have not done it the best way, i am still learning.
I can do it even if the pull request has already been accepted?

bfontaine:
> I can do it even if the pull request has already been accepted? 

No, it’s only for opened PRs. Just push on the PR branch and it’ll update it accordingly.

dardo82:
OK, I will keep it in mind for the next time... Thanks for the tip.

mikemcquaid:

DomT4:
@dardo82 I read the thread you linked to in the formula and presumed you were right as well. These things happen :)

dardo82:
Yeah, I suppose... ;-) And you surely rock more! 👍

PR #42852, opened 2015-08-12
internetarchive: fix build with newer versions of clang

geoff-codes:
The introduction of the 'speedups' dependencies in 21b7ec4 causes build failure with newer versions of clang (i.e., that included in latest Xcode betas), due to change to `-std=gnu11` as default. This is a known issue with known workaround, see: https://github.com/NixOS/nixpkgs/issues/7275

Specific error is:
```
clang -DNDEBUG -g -fwrapv -Os -Wall -Wstrict-prototypes -U__llvm__ -arch x86_64 -arch i386 -pipe -DLIBEV_EMBED=1 -DEV_COMMON= -DEV_CLEANUP_ENABLE=0 -DEV_EMBED_ENABLE=0 -DEV_PERIODIC_ENABLE=0 -Ibuild/temp.macosx-10.10-intel-2.7/libev -Ilibev -I/System/Library/Frameworks/Python.framework/Versions/2.7/include/python2.7 -c gevent/gevent.core.c -o build/temp.macosx-10.10-intel-2.7/gevent/gevent.core.o
In file included from gevent/gevent.core.c:249:
In file included from gevent/libev.h:2:
libev/ev.c:1029:42: error: '_Noreturn' keyword must precede function declarator
  ecb_inline void ecb_unreachable (void) ecb_noreturn;
                                         ^~~~~~~~~~~~
  _Noreturn
libev/ev.c:832:26: note: expanded from macro 'ecb_noreturn'
  #define ecb_noreturn   _Noreturn
                         ^
1 error generated.
error: command 'clang' failed with exit status 1
```
The culprit is the older version of `libev` being used in the `gevent` PyPi package.

Another alternative solution would be to remove `gevent` and `ujson`  as these are optional deps. I have opted to mark them as such in comments, instead.

Also have bumped dependency resources where applicable.

CC: @mistydemeo

### DIFF
--- a/Library/Formula/internetarchive.rb
+++ b/Library/Formula/internetarchive.rb
@@ -3,8 +3,8 @@ require "utils/json"
 class Internetarchive < Formula
   desc "Python wrapper for the various Internet Archive APIs"
   homepage "https://github.com/jjjake/internetarchive"
-  url "https://pypi.python.org/packages/source/i/internetarchive/internetarchive-0.8.5.tar.gz"
-  sha256 "2ba5e8db802953b1ac25a73c88d0955df2e4cd947bc664d94d6000003b91f14e"
+  url "https://pypi.python.org/packages/source/i/internetarchive/internetarchive-0.9.8.tar.gz"
+  sha256 "d89a0e04ea99e65ead8f5a00114c807e0d75108b76eaf34be4abd9f0c41bf73e"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Has this issue been reported upstream? It's unclear to me how this is a bug affecting Clang if it didn't affect any of our bots? Is it only a failure on 10.11?

geoff-codes:
@mikemcquaid It's not to do with 10.11 _per se_... in fact, I'm not running 10.11 at all. It *does* have to do with Xcode 7 (currently at beta 5), which will be released for 10.10 and 10.11. The clang included in Xcode 7, "Apple LLVM version 7.0.0 (clang-700.0.65)", is based on a version of UIUC clang 3.6, which [adopts `-std=gnu11` by default](http://llvm.org/releases/3.6.0/tools/clang/docs/ReleaseNotes.html#c-language-changes-in-clang). So its not really a bug, either. Although there's not a whole lot different between C99 and C11, this is one place they diverge; so I'd guess that once Xcode 7 is released, you may end up with a few other formulae with similar issues.

geoff-codes:
See also: http://lists.schmorp.de/pipermail/libev/2015q1/002480.html

mikemcquaid:
LGTM but will let @mistydemeo pull if she's happy :+1:

dunn:
Feel like rolling the 0.9.8 update into this PR?

geoff-codes:
> Feel like rolling the 0.9.8 update into this PR?

Done. Would be cool to see this merged in 2015. :wink:

dunn:
Merged in 78378a4661fe0f2925b4afeec7301b1019a8fe09, thanks again!

PR #42854, opened 2015-08-12
vegeta 5.8.0

tinnet:
just replaced source tar.gz, hash, and version :)

changelog:

>2015-08-11: v5.8.0
>  Change reporters quantile estimation method to match R's 8th type.

https://github.com/tsenart/vegeta/blob/master/CHANGELOG

### DIFF
--- a/Library/Formula/vegeta.rb
+++ b/Library/Formula/vegeta.rb
@@ -3,8 +3,8 @@ require "language/go"
 class Vegeta < Formula
   desc "HTTP load testing tool and library"
   homepage "https://github.com/tsenart/vegeta"
-  url "https://github.com/tsenart/vegeta/archive/v5.7.1.tar.gz"
-  sha256 "8bf49431f9eb1b36cd3f8b816f12a8e9e5b2b5b601223e45aae13d702bda9117"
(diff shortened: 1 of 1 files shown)


Library/Formula/vegeta.rb
```diff
@@ -27,7 +27,7 @@ def install
     ENV["CGO_ENABLED"] = "0"
     Language::Go.stage_deps resources, buildpath/"src"
 
-    system "go", "build", "-ldflags", "-X main.Version v5.7.1", "-o", "vegeta"
+    system "go", "build", "-ldflags", "-X main.Version v5.8.0", "-o", "vegeta"
```
mikemcquaid: Use "#{version}` in here.

tinnet: where does that variable come from? do i have to set it within the formula?

mikemcquaid: Nope, set for you by default. Comes from the URL.

tinnet:
done, do you want me to squash the commits or is it ok like this? :)

dunn:
We prefer a single commit in most cases, yeah. I took care if it here, though! Merged in adb7bfb5aca60b1e8c815e3fc0d941616d62ad02, thanks again.

PR #42857, opened 2015-08-12
fix migrator bug

vladshablinsky:
```
brew install go-app-engine-32
brew unlink go-app-engine-32
brew install go-app-engine-64
```

then rename `go-app-engine-32` to `app-engine-go-32`
and `migrate` or `update` and we have:

```
==> Migrating go-app-engine-32 to app-engine-go-32
==> Unlinking go-app-engine-32
Moving to: /usr/local/Cellar/app-engine-go-32
==> Linking app-engine-go-32
Error: Error while executing `brew link` step on app-engine-go-32
Could not symlink bin/appcfg.py
Target /usr/local/bin/appcfg.py
is a symlink belonging to go-app-engine-64. You can unlink it:
  brew unlink go-app-engine-64

  brew link --overwrite app-engine-go-32

  brew link --overwrite --dry-run app-engine-go-32

/usr/local/bin/appcfg.py -> /usr/local/Cellar/go-app-engine-64/1.9.24/bin/appcfg.py
/usr/local/bin/bulkload_client.py -> /usr/local/Cellar/go-app-engine-64/1.9.24/bin/bulkload_client.py
/usr/local/bin/bulkloader.py -> /usr/local/Cellar/go-app-engine-64/1.9.24/bin/bulkloader.py
/usr/local/bin/dev_appserver.py -> /usr/local/Cellar/go-app-engine-64/1.9.24/bin/dev_appserver.py
/usr/local/bin/download_appstats.py -> /usr/local/Cellar/go-app-engine-64/1.9.24/bin/download_appstats.py
/usr/local/bin/goapp -> /usr/local/Cellar/go-app-engine-64/1.9.24/bin/goapp
Error: error occured while migrating.
Backuping...
Error: No such file or directory - /usr/local/Cellar/go-app-engine-32
```

This happens because we try to link app-engine-go-32
 even if go-app-engine-32 wasn’t linked.

Backup fails because it uses realpath on the file that doesn't exist.

* Fix backup
* Don't link unlinked kegs.

cc @mikemcquaid @xu-cheng

### DIFF
--- a/Library/Homebrew/migrator.rb
+++ b/Library/Homebrew/migrator.rb
@@ -131,7 +131,7 @@ class Migrator
       unlink_oldname
       move_to_new_directory
       repin
-      link_newname
+      link_newname if oldkeg_linked?
       link_oldname_opt
       link_oldname_cellar
       update_tabs
@@ -182,7 +182,10 @@ class Migrator
     oh1 "Linking #{Tty.green}#{newname}#{Tty.reset}"
     keg = Keg.new(formula.installed_prefix)
 
-    if formula.keg_only?
+    # If old_keg wasn't linked then we just optlink a keg.
+    # If old keg wasn't optlinked and linked, we don't call this method at all.
+    # If formula is keg-only we also optlink it.
+    if formula.keg_only? || !old_linked_keg_record
       begin
         keg.optlink
       rescue Keg::LinkError => e
@@ -241,7 +244,7 @@ class Migrator
   # Remove opt/oldname link if it belongs to newname.
   def unlink_oldname_opt
     return unless old_opt_record
-    if (old_opt_record.symlink? && old_opt_record.exist?) \
+    if old_opt_record.symlink? && old_opt_record.exist? \
         && formula.installed_prefix.exist? \
         && formula.installed_prefix.realpath == old_opt_record.realpath
       old_opt_record.unlink
@@ -285,18 +288,27 @@ class Migrator
     end
 
     if oldkeg_linked?
-      begin
-        # The keg used to be linked  and when we backup everything we restore
-        # Cellar/oldname, the target also gets restored, so we are able to
-        # create a keg using its old path
-        keg = Keg.new(Pathname.new(oldkeg.to_s))
-        keg.link
-      rescue Keg::LinkError
-        keg.unlink
-        raise
-      rescue Keg::AlreadyLinkedError
-        keg.unlink
-        retry
+      # The keg used to be linked and when we backup everything we restore
+      # Cellar/oldname, the target also gets restored, so we are able to
+      # create a keg using its old path
+      if old_linked_keg_record
+        begin
+          oldkeg.link
+        rescue Keg::LinkError
+          oldkeg.unlink
+          raise
+        rescue Keg::AlreadyLinkedError
+          oldkeg.unlink
+          retry
+        end
+      else
+        begin
+          oldkeg.optlink
+        rescue Keg::LinkError => e
+          onoe "Failed to create #{formula.opt_prefix}"
+          puts e
+          raise
+        end
       end
     end
   end
--- a/Library/Homebrew/migrator.rb
+++ b/Library/Homebrew/migrator.rb
@@ -53,6 +53,9 @@ class Migrator
   attr_reader :newname, :newpath, :new_pin_record
   attr_reader :old_pin_link_record
 
+  # Path to new linked keg
+  attr_reader :new_keg
+
   def initialize(formula)
     @oldname = formula.oldname
     @newname = formula.name
@@ -71,6 +74,7 @@ class Migrator
     if @oldkeg = get_linked_oldkeg
       @old_linked_keg_record = oldkeg.linked_keg_record if oldkeg.linked?
       @old_opt_record = oldkeg.opt_record if oldkeg.optlinked?
+      @new_keg = HOMEBREW_CELLAR/"#{newname}/#{File.basename(oldkeg)}"
     end
 
     @old_pin_record = HOMEBREW_LIBRARY/"PinnedKegs"/oldname
@@ -180,7 +184,7 @@ class Migrator
 
   def link_newname
     oh1 "Linking #{Tty.green}#{newname}#{Tty.reset}"
-    keg = Keg.new(formula.installed_prefix)
+    keg = Keg.new(new_keg)
 
     # If old_keg wasn't linked then we just optlink a keg.
     # If old keg wasn't optlinked and linked, we don't call this method at all.
@@ -190,7 +194,6 @@ class Migrator
         keg.optlink
       rescue Keg::LinkError => e
         onoe "Failed to create #{formula.opt_prefix}"
-        puts e
         raise
       end
       return
@@ -227,7 +230,7 @@ class Migrator
   def link_oldname_opt
     if old_opt_record
       old_opt_record.delete if old_opt_record.symlink?
-      old_opt_record.make_relative_symlink(formula.installed_prefix)
+      old_opt_record.make_relative_symlink(new_keg)
     end
   end
 
@@ -245,8 +248,8 @@ class Migrator
   def unlink_oldname_opt
     return unless old_opt_record
     if old_opt_record.symlink? && old_opt_record.exist? \
-        && formula.installed_prefix.exist? \
-        && formula.installed_prefix.realpath == old_opt_record.realpath
+        && new_keg.exist? \
+        && new_keg.realpath == old_opt_record.realpath
       old_opt_record.unlink
       old_opt_record.parent.rmdir_if_possible
     end
@@ -302,13 +305,7 @@ class Migrator
           retry
         end
       else
-        begin
-          oldkeg.optlink
-        rescue Keg::LinkError => e
-          onoe "Failed to create #{formula.opt_prefix}"
-          puts e
-          raise
-        end
+        oldkeg.optlink
       end
     end
   end
--- a/Library/Homebrew/migrator.rb
+++ b/Library/Homebrew/migrator.rb
@@ -223,7 +223,7 @@ class Migrator
   # Link keg to opt if it was linked before migrating.
   def link_oldname_opt
     if old_opt_record
-      old_opt_record.delete if old_opt_record.symlink? || old_opt_record.exist?
+      old_opt_record.delete if old_opt_record.symlink?
       old_opt_record.make_relative_symlink(formula.installed_prefix)
     end
   end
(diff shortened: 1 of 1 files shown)


Library/Homebrew/migrator.rb
```diff
@@ -232,8 +235,9 @@ def update_tabs
   # Remove opt/oldname link if it belongs to newname.
   def unlink_oldname_opt
     return unless old_opt_record
-    if old_opt_record.symlink? && formula.installed_prefix.exist? \
-              && formula.installed_prefix.realpath == old_opt_record.realpath
+    if (old_opt_record.symlink? && old_opt_record.exist?) \
```
mikemcquaid: Don't need the parentheses here because you're always using `&&`

Library/Homebrew/migrator.rb
```diff
@@ -275,18 +279,29 @@ def backup_oldname
...
-        # create a keg using its old path
-        keg = Keg.new(Pathname.new(oldkeg.to_s))
-        keg.link
-      rescue Keg::LinkError
-        keg.unlink
-        raise
-      rescue Keg::AlreadyLinkedError
-        keg.unlink
-        retry
+      # The keg used to be linked  and when we backup everything we restore
```
mikemcquaid: totally pedantic but mind fixing the two spaces between `linked` and `and`?

mikemcquaid:
Looks good to me. @xu-cheng, let me know your thoughts.

xu-cheng:
About the opt_link and pin, I'm wondering how does renaming formulae affect `brew link/unlink/pin/unpin/upgrade`? @viktorahlstrom Can you elaborate a bit about the effect on all of these commands?

vladshablinsky:
@xu-cheng when we have some formula installed we need to move it to new Cellar, and therefore relink, repin, etc.

xu-cheng:
For example in the case of `upgrade`, if I have formula `foo` installed in `Cellar/foo/1.0` and that formula is renamed to `bar`. After the migration, we will have `opt/foo` and `opt/bar` pointed to `Cellar/bar/1.0` right? So under this condition, say formula is upgrade to `2.0`. It looks to me we will have `opt/bar` pointed to `Cellar/bar/2.0` but also `opt/foo` pointed to `Cellar/bar/1.0`.

vladshablinsky:
Yes, `foo`'s opt points to the keg it used to point before the name changed.

xu-cheng:
I think that we will have a problem. Especially, if I then run `brew cleanup`. We will have a dead symlink, right?

vladshablinsky:
Maybe I confused you in the previous message. `opt/oldname` points to `Cellar/newname/version` after migration. If we upgarde `newname`, `opt/oldname` still points to the `Cellar/newname/version`, however `Cellar/newname/newversion` can exist after `upgrade`.

xu-cheng:
Right, I mean "opt/oldname still points to the Cellar/newname/version" would cause problem. Because if I run `brew cleanup`. `Cellar/newname/version` will be deleted. `opt/oldname` will become a dead symlink.

vladshablinsky:
Tested this, yes, this is a problem. The link becomes broken. We need to update `opt` on upgrade, then.

xu-cheng:
So it will also affect `switch/link/unlink` as all of them will deal with `opt_link` somehow.

vladshablinsky:
switch/link/unlink deal only with intsalled kegs and their optlinks, but not with those of oldname, am I wrong?

vladshablinsky:
Ah, I got it, you mean the case when we run `cleanup` after `switch`.

xu-cheng:
I'm not sure here. I mean what will happen when we run `(un)link/switch <oldname>`? There seem to be a few cases we need to think carefully.

vladshablinsky:
`switch` uses `to_rack` so it behaves just like if there is no `oldname` symlinks. The dangerous command here is `cleanup`. 

What can we do here? I think optlink `oldname`in `Keg#link` and `Keg#optlink` can be a solution.

xu-cheng:
> switch uses to_rack so it behaves just like if there is no oldname symlinks. 

No, to_rack will also return the old name rack. So instead, it will behave like there is no new name.

vladshablinsky:
It retruns old name rack only if old name is a directory, which is not true for migrated formula.

xu-cheng:
symlink directory is still a directory.

xu-cheng:
OK, these are what I am suggesting:
* Make `to_rack` follow the symlink.
* Make `Keg#link` and `Keg#unlink` handle old name if it's applied.

vladshablinsky:
@xu-cheng still haven't got green build, but:

* tested with switch
* tested with link/unlink/pin/unpin
* tested with cleanup
* tested when migrating unlinked formula
* `Keg#initialize` rescues Exception in order not to break lots of tests, that use `Keg.new`

mikemcquaid:
@vladshablinsky @xu-cheng Green build now.

Library/Homebrew/keg.rb
```diff
@@ -101,6 +102,19 @@ def initialize(path)
...
+
+      oldname = if tap.nil? || tap == "Homebrew/homebrew"
+        Formulary.factory(name)
+      else
+        Formulary.factory("#{tap.sub("homebrew-", "")}/#{name}")
+      end.oldname
+
+      @oldname_opt_record = HOMEBREW_PREFIX.join("opt", oldname) if oldname
+    rescue Exception
+    end
```
xu-cheng: I'd like to avoid loading formula here. Instead we can just get it by reading the filesystem. Beside this, in order to improve the performance, let's make it a separate method instead of computing it in `initialize`. So in all, here is my proposal.

```ruby
def oldname_opt_record
  @oldname_opt_record ||= if (opt_dir = HOMEBREW_PREFIX/"opt").directory?
    opt_dir.subdirs.detect { |dir| dir.symlink? && path == dir.resolved_path }
  end
end

def remove_oldname_opt_record
  oldname_opt_record.unlink
  oldname_opt_record.parent.rmdir_if_possible
  @oldname_opt_record = nil
end
```

vladshablinsky: @xu-cheng I don't completely understand your idea:

* To get `oldname_opt_record` we need to:
```ruby
(opt_dir = HOMEBREW_PREFIX/"opt").directory?
  opt_dir.subdirs.detect do |dir|
    dir.symlink? && dir.resolved_path.parent == path.parent
  end
end
```
 There can be several `dir`s, so we have to exclude `opt_record` from `opt_dir.subdirs` before detecting.
* Before we can do the step above, the formula have to be migrated, so `oldname_opt_record` isn't `nil`.
Is this what you mean?

vladshablinsky: As an option we can store `oldname` in `INSTALL_RECEIPT`.

xu-cheng: I mean can we get `oldname_opt_record` only using filesystem. 

>  so we have to exclude opt_record from opt_dir.subdirs before detecting.

Sorry I missed this part. How about:

```ruby
def oldname_opt_record
  @oldname_opt_record ||= if (opt_dir = HOMEBREW_PREFIX/"opt").directory?
    opt_dir.subdirs.detect { |dir| dir.symlink? && dir != opt_record && path == dir.resolved_path }
  end
end
```

vladshablinsky: Yes, I see your idea, but this still doesn't work. We need to optlink oldname, when it's not linked (e.g. on `upgrade`. First of all we unlink linked keg in `upgrade`, then we install new keg, and then link it. If we try to optlink that new keg, there won't be any symlink in `opt` pointing to the new keg, so we have to use `path.parent == dir.resolved_path.parent` to know that it point to the right package).

vladshablinsky:
Since two formulae have been renamed already and the bug was found, we need to merge this PR before updating versions of `ocaml` and `caf`.  CC @Homebrew/owners .

mikemcquaid:
> Since two formulae have been renamed already and the bug was found, we need to merge this PR before updating versions of `ocaml` and `caf`.

CC @Homebrew/owners.

xu-cheng:
Let's slow down a bit otherwise we would create more bug. I'm ok to merge first commits. The rest of commits require more work.

xu-cheng:
Maybe better to move the last two commits to a new PR.

vladshablinsky:
OK, I'll move them as soon as I fix [this](https://github.com/vladshablinsky/homebrew/blob/3a2cc30677a85be60fad37d1a33c4286073aa27d/Library/Homebrew/migrator.rb#L183) and we can merge two first commits with that fix.

vladshablinsky:
Replace `formula.installed_prefix`(that was used to point to keg that needed to be linked after migration) with path to new linked keg according to old linked keg. Fixes error in migrator, that caused an error in outdated. https://github.com/Homebrew/homebrew/issues/42938.

Library/Homebrew/migrator.rb
```diff
@@ -71,6 +74,7 @@ def initialize(formula)
     if @oldkeg = get_linked_oldkeg
       @old_linked_keg_record = oldkeg.linked_keg_record if oldkeg.linked?
       @old_opt_record = oldkeg.opt_record if oldkeg.optlinked?
+      @new_linked_keg_path = HOMEBREW_CELLAR/"#{newname}/#{oldkeg.to_s.split("/").last}"
```
xu-cheng: `oldkeg.to_s.split("/").last` -> `File.basename(oldkeg)`

Library/Homebrew/migrator.rb
```diff
@@ -53,6 +53,9 @@ def initialize(formula, tap)
   attr_reader :newname, :newpath, :new_pin_record
   attr_reader :old_pin_link_record
 
+  # Path to new linked keg
+  attr_reader :new_linked_keg_path
```
xu-cheng: I think `new_keg` seems to be a better name.

Library/Homebrew/migrator.rb
```diff
@@ -284,18 +292,27 @@ def backup_oldname
...
+        rescue Keg::AlreadyLinkedError
+          oldkeg.unlink
+          retry
+        end
+      else
+        begin
+          oldkeg.optlink
+        rescue Keg::LinkError => e
+          onoe "Failed to create #{formula.opt_prefix}"
+          puts e
```
xu-cheng: We don't need `puts e` here, since you are going to `raise`.

mikemcquaid:
@xu-cheng Can we merge this sooner rather than later to fix the bugs and then do cleanup in a follow-up PR?

mikemcquaid:
Or if you're happy to merge just a few commits: can you merge them? Thanks!

vladshablinsky:
Updated according to comments.

Library/Homebrew/migrator.rb
```diff
@@ -284,18 +291,27 @@ def backup_oldname
...
+          raise
+        rescue Keg::AlreadyLinkedError
+          oldkeg.unlink
+          retry
+        end
+      else
+        begin
+          oldkeg.optlink
+        rescue Keg::LinkError => e
+          onoe "Failed to create #{formula.opt_prefix}"
```
xu-cheng: Final question before I merge this: why do you need to rescue here. Also `oldkeg.optlink` failure means failing to create old opt prefix rather than `formula.opt_prefix`.

I suggest you just make it:
```ruby
else
  oldkeg.optlink
end
```

vladshablinsky: Yes, rescue is unnecessary here, thanks.

mikemcquaid:
Thanks @xu-cheng, @vladshablinsky :heart:

PR #42858, opened 2015-08-12
znc: set CXXFLAGS appropriately

### LINKED ISSUE #42822: ZNC formula: issue with C++11

mistydemeo:
Fixes #42822.

### DIFF
--- a/Library/Formula/znc.rb
+++ b/Library/Formula/znc.rb
@@ -27,6 +27,8 @@ class Znc < Formula
   depends_on "openssl"
   depends_on "icu4c" => :optional
 
+  needs :cxx11
+
   def install
     ENV.cxx11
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1: when green

Issue #41766, opened 2015-07-16
Fig / docker-compose fails with recent version


beorn:
Ref fig/docker-compose bug: https://github.com/docker/compose/issues/1484 - ssl_.py warns of an InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning

I am not sure what the right way to fix this is, but sure would like to see it working.  Maybe depend on non-system python to ensure we can get a higher version?  I don't know if it affects other docker-* packages as well.

tdsmith:
This isn't a failure and the message is safe to ignore.

If you wanted to stop seeing it, it looks like we could ship PyOpenSSL with docker-compose or we could just pass Python options to squelch warnings.

richardPFisk:
"This isn't a failure and the message is safe to ignore."

Whilst not a failure, the message prevents docker-compose from working. 

Is there an easy way to downgrade docker-compose with brew?

tdsmith:
> Whilst not a failure, the message prevents docker-compose from working.

Can you clarify what isn't working?

tdsmith:
Building with pyOpenSSL exposes whatever badness is happening with OpenSSL 1.0.2 if we link against Homebrew's openssl:
```
tim@rocketman:test$ docker-compose up --allow-insecure-ssl
SSL error: [Errno bad handshake] [('SSL routines', 'ssl3_get_server_certificate', 'certificate verify failed')]
```

Previously: #38226

tdsmith:
Comments are welcome on PR #41815.

tdsmith:
I'm rejecting #41815 because I believe the warning is correct and benign (i.e. not a failure). The fix for this will be to install pyOpenSSL as a fig dependency once boot2docker/boot2docker#808 is favorably resolved.

frio:
I don't believe the warning is benign.  When attempting to run `docker-compose start`, I am issued an `InsecurePlatformWarning`, at which point `docker-compose` exits.

tdsmith:
boot2docker 1.8.0 has been released which fixes the relevant certificate issue, so we can move on this once we accept the new boot2docker.

tdsmith:
Mooted because system Python is now 2.7.10 on the newest Yosemite release.

richardPFisk:
[...]

PR #42859, opened 2015-08-12
Update fig.rb

cristobal:
Change shebang in the `libexec/"bin/docker-compose"` from the generated:

`#!/usr/bin/python` 

to: 

`#!/usr/bin/env python` 

Since otherwise *docker-composer* will use the system default python currently *Python 2.7.6* on latest *Yosemite* and not the required *python* installed with *brew*.  This resolves the annoying issue `ssl_.py warns of an InsecurePlatformWarning: A true SSLContext object is not available.`, and would resolve the issue [41766](https://github.com/Homebrew/homebrew/issues/41766)

### DIFF
--- a/Library/Formula/fig.rb
+++ b/Library/Formula/fig.rb
@@ -79,6 +79,10 @@ class Fig < Formula
 
     bin.install Dir[libexec/"bin/*"]
     bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+    
+    # change shebang from system python to env python
+    inreplace libexec/"bin/docker-compose", "#!/usr/bin/python", "#!/usr/bin/env python"
+    
     ln_s bin/"docker-compose", bin/"fig"
   end
 


Library/Formula/fig.rb
```diff
@@ -79,6 +79,10 @@ def install
 
     bin.install Dir[libexec/"bin/*"]
     bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+    
+    # change shebang from system python to env python
+    inreplace libexec/"bin/docker-compose", "#!/usr/bin/python", "#!/usr/bin/env python"
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

DomT4:
I suspect this has the potential to cause pain, since it will be built against the system Python rather than ours, and editing that shebang will potentially then pointing at a newer Homebrew Python after installation.

10.10.5 is shipping Python 2.7.10 which negates a lot of these concerns on Yosemite.

FWIW: `depends_on :python if MacOS.version <= :snow_leopard` = Use the system Python if it is at least version 2.7. Only uses Homebrew Python on Snow Leopard and below.

CC @tdsmith

cristobal:
Well it would be possible to replace the shebang with the required *Python*
installed via brew by changing it to be installed via *pip* from the
installed via the brew *Python*. Installing qs it is now using `Language::
python.setuptools_install (*)` in the formula always ends up with the
shebang being the system python. Meaning that this formula requires on
python via brew makes little sense. Since the script will always execute
via `/usr/bin/python`.

tdsmith:
Thanks for the PR. Please see https://github.com/Homebrew/homebrew/issues/41766#issuecomment-126876088.

tdsmith:
This will incidentally cause grisly failures with boot2docker < 1.8.0 exactly because Homebrew and system pythons use different versions of openssl and openssl 1.0.2 doesn't like boot2docker's certs.

cristobal:
Just came home.

Sorry earlier comment misunderstood in terms of installing using the `/usr/bin/python`. 
I get the point of using the system python, still the warning is annoying as *f...*.

Just patching my own then, it works for me ;)

tdsmith:
This issue was resolved almost simultaneously by the release of boot2docker 1.8 and Apple's migration to Python 2.7.10 in Yosemite.

PR #42861, opened 2015-08-12
git-subrepo 0.2.3

callahad:
Version bump

The previous patch referenced in the formula has been merged upstream.

### DIFF
--- a/Library/Formula/git-subrepo.rb
+++ b/Library/Formula/git-subrepo.rb
@@ -1,19 +1,9 @@
 class GitSubrepo < Formula
   desc "Git Submodule Alternative"
   homepage "https://github.com/ingydotnet/git-subrepo"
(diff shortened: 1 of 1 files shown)


Library/Formula/git-subrepo.rb
```diff
@@ -4,15 +4,8 @@ class GitSubrepo < Formula
   head "https://github.com/ingydotnet/git-subrepo.git"
 
   stable do
```
dunn: Without the patch, we can do without the stable block; `url` and `sha256` can be placed above `head` now.

callahad: Thanks for the note.

Fixed and updated force-pushed the PR. :)

mikemcquaid:

PR #42862, opened 2015-08-12
bitrise 0.9.8

viktorbenei:


viktorbenei:
Should I leave the `bottle do` block in the Formula? It seemed strange to leave it there, as it references the previous version, but checking other PRs the `bottle do` is usually not modified. Do you have a preference in this matter?

mistydemeo:
We prefer to leave it there. The bots will update that block with the new values when we merge.

Library/Formula/bitrise.rb
```diff
@@ -3,26 +3,20 @@
...
-    cellar :any
-    sha256 "8d8cdcda113f12ab497db0700a04849143e733e2cf9c35c5e028b434f694d67c" => :yosemite
-    sha256 "9eb2435bdc04f4acaf975a5adf7bbbcced64c6ac60a678696919d67e59f04ef4" => :mavericks
-    sha256 "54e66ee1699983c1bb4a9662a8ff66a2e50aac85cd0eb0c613513b6cffa004b7" => :mountain_lion
-  end
+  url "https://github.com/bitrise-io/bitrise/archive/0.9.8.tar.gz"
+  sha256 "4282da6b9ce7f28f6caf04e12339eaee67117f44f36545cf29294631d94a806e"
 
   depends_on "go" => :build
+  depends_on :macos => :mavericks
```
mistydemeo: Out of curiosity, what's the reason this is incompatible? Are supported OSs documented somewhere?

viktorbenei: The reason is that this tool is mainly used by iOS developers, where every Xcode version is only supported on the latest 2 OS X version, and so we don't test it on older OS X versions.

In case of 0.9.8 the incompatibility is caused by `xcode-select` which is only available on Mavericks and newer OS X versions.

viktorbenei:
@mistydemeo updated the PR - leaving the `bottle do` block unchanged

Library/Formula/bitrise.rb
mistydemeo: Could you make it a `depends_on :xcode => "version"` dependency instead, in that case? Thanks!

> In case of 0.9.8 the incompatibility is caused by xcode-select which is only available on Mavericks and newer OS X versions.

`xcode-select` is available in older versions of OS X, too; it was introduced in Xcode 3.0. Is there specific behaviour you're using that's only in 10.9?

viktorbenei: xcode-select is preinstalled on OS X starting with Mavericks, before that it was only part of Xcode, but not preinstalled on the OS.

The full Xcode is not required, thats why the OS version dependency is sufficient.

viktorbenei: I forgot to mention that I specifically added the osx version dependency because the first brew ci test bot check failed on mountain lion, because xcode-select was not available.

mistydemeo: I think the problem is actually something else - the `-p` shorthand switch was added in 10.9, but `xcode-select -print-path` works in OS X back to 10.5. Submitting a PR to the bitrise repo that I believe fixes this.

viktorbenei:
I'll close this PR and open a new one later today when we release 0.9.9 with the change proposed by @mistydemeo - to fix Mountain Lion support. Thank you!

PR #42869, opened 2015-08-12
dwarfutils 20150507 (new formula)

jwatzman:
Add the dwarfutils formula, a library (libdwarf) and a utility
(dwarfdump) to produce and dump DWARF debugging information in ELF
object files.

This package is named "dwarfutils" since, although the upstream website
and tarball are plastered with "libdwarf", that's only one part of what
this package provides. The Debian source package is named the same:
https://packages.debian.org/source/sid/dwarfutils

### DIFF
--- a/Library/Formula/dwarfutils.rb
+++ b/Library/Formula/dwarfutils.rb
@@ -1,0 +1,52 @@
+class Dwarfutils < Formula
+  desc "lib and utility to dump and produce DWARF debug info in ELF objects"
+  homepage "http://www.prevanders.net/dwarf.html"
+  url "http://www.prevanders.net/libdwarf-20150507.tar.gz"
+  sha256 "29aa8d07db659d7d7af7079854cf42c09bf74d303942159cbfee82d655549870"
+
+  depends_on "libelf" => :build
+
+  def install
+    system "./configure"
+    system "make"
+
+    bin.install "dwarfdump/dwarfdump"
+    man1.install "dwarfdump/dwarfdump.1"
+    lib.install "libdwarf/libdwarf.a"
+    include.install "libdwarf/dwarf.h"
+    include.install "libdwarf/libdwarf.h"
+  end
+
+  test do
+    # Extensively testing dwarfdump is difficult, since we don't have an ELF
+    # image around to play with, and generating one on OS X is hard. Just do a
+    # very basic smoke test.
+    system "#{bin}/dwarfdump", "-V"
+
+    (testpath/"test.c").write <<-EOS.undent
+      #include <dwarf.h>
+      #include <libdwarf.h>
+      #include <stdio.h>
+      #include <string.h>
+
+      int main(void) {
+        const char *out = NULL;
+        int res = dwarf_get_children_name(0, &out);
+
+        if (res != DW_DLV_OK) {
+          printf("Getting name failed\\n");
+          return 1;
+        }
+
+        if (strcmp(out, "DW_children_no") != 0) {
+          printf("Name did not match: %s\\n", out);
+          return 1;
+        }
+
+        return 0;
+      }
+    EOS
+    system ENV.cc, "-L#{lib}", "-I#{include}", "-ldwarf", "test.c", "-o", "test"
+    system "./test"
+  end
+end


Library/Formula/dwarfutils.rb
```diff
@@ -0,0 +1,28 @@
...
+    # README. Just follow the directions the readme suggests.
+    bin.install "dwarfdump/dwarfdump"
+    man1.install "dwarfdump/dwarfdump.1"
+    lib.install "libdwarf/libdwarf.a"
+    include.install "libdwarf/dwarf.h"
+    include.install "libdwarf/libdwarf.h"
+  end
+
+  test do
+    system "#{bin}/dwarfdump", "-V"
```
bfontaine: Could you please make a test more elaborated than just `-V` or `-h`?

Library/Formula/dwarfutils.rb
```diff
@@ -0,0 +1,28 @@
...
+  url "http://www.prevanders.net/libdwarf-20150507.tar.gz"
+  sha256 "29aa8d07db659d7d7af7079854cf42c09bf74d303942159cbfee82d655549870"
+
+  depends_on "libelf" => :build
+
+  def install
+    # Disable building the shared library, since the package does not understand
+    # OS X naming quirks (dylib extension) and rpath quirks, and the static
+    # library works just fine.
+    system "./configure", "--disable-shared"
```
mikemcquaid: It would be good to build a shared library (the default) and fix these quirks upstream.

jwatzman: I'll look into it, but since the static library works with so little trouble, and it useful in and of itself, is it okay to merge this formula with the shared library disabled, and update it if it gets fixed upstream?

FWIW I think `--disable-shared` is the upstream default anyways, I was just being explicit here and explaining why the shared library isn't here for anyone who wondered wandering by.

Library/Formula/dwarfutils.rb
jwatzman: An extensive test is pretty difficult, since this package is designed to work on ELF objects, which are harder to come by on OS X unless you're doing something fairly weird/complex. A somewhat better test might be a simple C program that linked against libdwarf and called some sort of init function, which I will update this diff to add.

jwatzman:
Thanks for the review! Will update later today with a better test case.

jwatzman:
Updated with a better test case.

Library/Formula/dwarfutils.rb
mikemcquaid: If it's the upstream default and you can just remove the definition here I'd be :+1:

Library/Formula/dwarfutils.rb
```diff
@@ -0,0 +1,59 @@
...
+    system "make"
+
+    # The package doesn't contain a "make install" target, as documented in its
```
DomT4: Is there any chance upstream would take a patch to enable `make install`?

Library/Formula/dwarfutils.rb
```diff
@@ -0,0 +1,59 @@
...
+    # README. Just follow the directions the readme suggests.
+    bin.install "dwarfdump/dwarfdump"
+    man1.install "dwarfdump/dwarfdump.1"
+    lib.install "libdwarf/libdwarf.a"
+    include.install "libdwarf/dwarf.h"
+    include.install "libdwarf/libdwarf.h"
+  end
+
+  test do
+    # Extensively testing dwarfdump is difficult, since we don't have an ELF
```
DomT4: Can kill off a lot of these comments :). There's enough examples of the same thing in other Homebrew formulae that it doesn't need explaining too much.

jwatzman:
Made requested changes (removed comments, removed `--disable-shared`).

As for upstream making a real `make install` target... they actually have `make install` which just prints a message saying to read the readme, and the readme goes on for a couple paragraphs about how to copy files around, all of which is less effort than just making a real `make install` target. So I suspect they don't want one for some reason.

mikemcquaid:
Pushed with comments removed, thanks!

jwatzman:
Thanks!

PR #42870, opened 2015-08-12
compcert 2.5

dunn:
Are there currently no bottles because of the [license](https://github.com/AbsInt/CompCert#license) or just because it hasn't been updated in a while?


mikemcquaid:
> just because it hasn't been updated in a while?

Probably that.

PR #42642, opened 2015-08-08
rust 1.2.0

dhda:

PR #42872, opened 2015-08-13
rust 1.2.0 / cargo 0.4.0 / El Cap Fix

felixbuenemann:
* Update rust to 1.2.0
* Update cargo to 1.4.0
* Simplify formula head/stable differences
* Add cargo Makefile fix for DYLD_LIBRARY_PATH on 10.11

This builds on the work done in #42642, but adds the above fixes.

### DIFF
--- a/Library/Formula/rust.rb
+++ b/Library/Formula/rust.rb
@@ -3,23 +3,18 @@ class Rust < Formula
   homepage "https://www.rust-lang.org/"
 
   stable do
-    url "https://static.rust-lang.org/dist/rustc-1.1.0-src.tar.gz"
-    sha256 "cb09f443b37ec1b81fe73c04eb413f9f656859cf7d00bc5088008cbc2a63fa8a"
+    url "https://static.rust-lang.org/dist/rustc-1.2.0-src.tar.gz"
+    sha256 "ea6eb983daf2a073df57186a58f0d4ce0e85c711bec13c627a8c85d51b6a6d78"
 
     resource "cargo" do
-      url "https://github.com/rust-lang/cargo.git", :revision => "b030d35d5cf6b35bf8a6bfd218ab4df9d6a86361"
+      # git required because of submodules
+      url "https://github.com/rust-lang/cargo.git", :tag => "0.4.0", :revision => "553b363bcfcf444c5bd4713e30382a6ffa2a52dd"
     end
 
     # name includes date to satisfy cache
-    resource "cargo-nightly-2015-06-25" do
-      url "https://static-rust-lang-org.s3.amazonaws.com/cargo-dist/2015-06-25/cargo-nightly-x86_64-apple-darwin.tar.gz"
-      sha256 "b2e07bbee79cb8ad1e4f91a43cc3d93603e068a46b89bbe934d01ff97bfb0060"
-    end
-
-    # name includes date to satisfy cache
-    resource "rustc-nightly-2015-06-25" do
-      url "https://static-rust-lang-org.s3.amazonaws.com/dist/2015-06-25/rustc-nightly-x86_64-apple-darwin.tar.gz"
-      sha256 "c4eb0a639b6deb3e2aceb1713afe6570118d1055bf189f1057a839238dbe7165"
+    resource "cargo-nightly-2015-08-12" do
+      url "https://static-rust-lang-org.s3.amazonaws.com/cargo-dist/2015-08-12/cargo-nightly-x86_64-apple-darwin.tar.gz"
+      sha256 "3d0ea9e20215e6450e2ae3977bbe20b9fb2bbf51ce145017ab198ea3409ffda2"
     end
   end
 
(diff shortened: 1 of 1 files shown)


felixbuenemann:
Btw. the cargo version shipping with the official mac binaries is `cargo 0.4.0-nightly (15b497b 2015-07-08) (built 2015-07-09)`. We could switch to that revision instead of the final 0.4.0 release if desired.

Library/Formula/rust.rb
```diff
@@ -3,23 +3,18 @@ class Rust < Formula
...
   stable do
-    url "https://static.rust-lang.org/dist/rustc-1.1.0-src.tar.gz"
-    sha256 "cb09f443b37ec1b81fe73c04eb413f9f656859cf7d00bc5088008cbc2a63fa8a"
+    url "https://static.rust-lang.org/dist/rustc-1.2.0-src.tar.gz"
+    sha256 "ea6eb983daf2a073df57186a58f0d4ce0e85c711bec13c627a8c85d51b6a6d78"
 
     resource "cargo" do
-      url "https://github.com/rust-lang/cargo.git", :revision => "b030d35d5cf6b35bf8a6bfd218ab4df9d6a86361"
+      # git required because of submodules
+      url "https://github.com/rust-lang/cargo.git", :tag => "0.4.0"
```
dunn: Include the `:revision` as well as the tag, please!

felixbuenemann: OK, pushed a new commit.

felixbuenemann:
Jenkins build failures could be due to out of space condition, although the errors are not very clear.
The temp folder used for the build requires at least 2GB free space.

dunn:
Oh, it gets marked as a failure when the build is aborted.  I stopped it since it has a long build time and we were going to have to run it again for the updated PR anyway.

felixbuenemann:
Ah, OK.

There is a strange error message in the yosemite build:
```
Error: 0.4.0 tag should be bb692e3888c36b87bf6f5052620e0a13fb464682
but is actually 553b363bcfcf444c5bd4713e30382a6ffa2a52dd
```
But the correct revision is 553b363bcfcf444c5bd4713e30382a6ffa2a52dd as can be seen here:
https://github.com/rust-lang/cargo/tree/0.4.0

Library/Formula/rust.rb
```diff
@@ -3,23 +3,18 @@ class Rust < Formula
...
   stable do
-    url "https://static.rust-lang.org/dist/rustc-1.1.0-src.tar.gz"
-    sha256 "cb09f443b37ec1b81fe73c04eb413f9f656859cf7d00bc5088008cbc2a63fa8a"
+    url "https://static.rust-lang.org/dist/rustc-1.2.0-src.tar.gz"
+    sha256 "ea6eb983daf2a073df57186a58f0d4ce0e85c711bec13c627a8c85d51b6a6d78"
 
     resource "cargo" do
-      url "https://github.com/rust-lang/cargo.git", :revision => "b030d35d5cf6b35bf8a6bfd218ab4df9d6a86361"
+      # git required because of submodules
+      url "https://github.com/rust-lang/cargo.git", :tag => "0.4.0", :revision => "bb692e3888c36b87bf6f5052620e0a13fb464682"
```
dunn: You might have just copied the wrong revision?  This one instead of `553b363bcfcf444c5bd4713e30382a6ffa2a52dd`?

felixbuenemann: Argh… my fault. I'll fix that asap.

yurikoles:
Please remove/comment out bottles.

Library/Formula/rust.rb
felixbuenemann: OK, pushed a new commit. The mavericks build bot still seems to run out of disk space (LLVM IO error).

felixbuenemann:
@yurikoles Why? The buildbot should auto-update the bottles.

dunn:
Yeah, don't worry about the bottle block! We take care of that.

felixbuenemann:
@mikemcquaid could you look into the diskspace problem for the mavericks build bot?

Library/Formula/rust.rb
```diff
@@ -3,23 +3,18 @@ class Rust < Formula
...
     end
 
     # name includes date to satisfy cache
-    resource "cargo-nightly-2015-06-25" do
-      url "https://static-rust-lang-org.s3.amazonaws.com/cargo-dist/2015-06-25/cargo-nightly-x86_64-apple-darwin.tar.gz"
-      sha256 "b2e07bbee79cb8ad1e4f91a43cc3d93603e068a46b89bbe934d01ff97bfb0060"
-    end
-
-    # name includes date to satisfy cache
-    resource "rustc-nightly-2015-06-25" do
```
DomT4: `rustc` is no longer required at all?

felixbuenemann: No, it was only used by the cargo resource for no apparent reason. The cargo resource now just uses the just installed rustc.

felixbuenemann:
Btw. the 10.11 cargo make fix has been merged in rust-lang/cargo#1903, so we could alternatively depend on a recent enough revision of cargo instead of patching the Makefile.in.

mikemcquaid:

felixbuenemann:
:heart:

tomjakubowski:
> Update cargo to 1.4.0 [sic]

Was 0.4.0 (commit 553b363bcfcf444c5bd4713e30382a6ffa2a52dd, from early August) chosen for a particular reason? The version that ships with the rust 1.2.0 distribution is `cargo 0.4.0-nightly (15b497b 2015-07-08) (built 2015-07-09)`. Why doesn't the Homebrew recipe use that?

PR #42877, opened 2015-08-13
openjpeg: symlink libopenjpeg1.pc > libopenjpeg.pc

### LINKED ISSUE #42450: broken ffmpeg build with openjpeg
The openjpeg formula was upgraded from 1.5.1 to 1.5.2. ffmpeg doesn't seem to support this version, so if I run `brew install ffmpeg --with-openjpeg` then I get a broken ffmpeg. 
```
ffmpeg 
dyld: Library not loaded: /usr/local/lib/libopenjpeg.1.dylib
  Referenced from: /usr/local/bin/ffmpeg
  Reason: image not found
Trace/BPT trap: 5
```

Is there a way to have ffmpeg use the non-current openjpeg lib?

dunn:
Fixes #42450.

cc @mistydemeo


mikemcquaid:
:+1:

PR #42878, opened 2015-08-13
findutils: fix test when built with the --default-names option

zmwangx:
If `findutils` is built with the `--default-names` option, then the `find` binary would be `bin/find` instead of `bin/gfind`.

### DIFF
--- a/Library/Formula/findutils.rb
+++ b/Library/Formula/findutils.rb
@@ -28,6 +28,7 @@ class Findutils < Formula
   end
 
   test do
-    system "#{bin}/gfind", "--version"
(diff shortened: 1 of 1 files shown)


zmwangx:
I just noticed that same thing happens for `gnu-sed`. I'll submit a patch for `gnu-sed` if the style here is acceptable.

dunn:
I wouldn't worry about it.  We don't usually test non-default options; it would require a lot more complication in many tests.

zmwangx:
Fair enough. Then close this?

zmwangx:
Anyway, my personal attitude is why not make it work when it's not complicated, like in this case.

dunn:
I think even here it adds a bit of complication that may be off-putting for new contributors.  I appreciate the thought, though!

PR #42880, opened 2015-08-13
oauth2_proxy: Add "head" and use string interpolation.

ebardsley:


mikemcquaid:

PR #42891, opened 2015-08-13
Don't try to link symlinks over their targets

tdsmith:
If we have lib/python3.4/site-packages, which is a symlink to
HOMEBREW_PREFIX/lib/python3.4/site-packages, link will be confused. This
only appears after unlinking and relinking because this symlink is
created in post_install, which runs after the first link.

Fixes Homebrew/linuxbrew#502.


mikemcquaid:
:+1:

PR #42897, opened 2015-08-13
bitrise 0.9.9

viktorbenei:


dunn:
Getting this test failure on all three bots:
```
        Testing bitrise
==> /usr/local/Cellar/bitrise/0.9.9/bin/bitrise run test_wf

  ██████╗ ██╗████████╗██████╗ ██╗███████╗███████╗
  ██╔══██╗██║╚══██╔══╝██╔══██╗██║██╔════╝██╔════╝
  ██████╔╝██║   ██║   ██████╔╝██║███████╗█████╗
  ██╔══██╗██║   ██║   ██╔══██╗██║╚════██║██╔══╝
  ██████╔╝██║   ██║   ██║  ██║██║███████║███████╗
  ╚═════╝ ╚═╝   ╚═╝   ╚═╝  ╚═╝╚═╝╚══════╝╚══════╝

[32;1mVersion: 0.9.9[0m

time="22:56:54" level=warning msg="\x1b[33;1mSetup was not performed for this version of bitrise, doing it now...\x1b[0m" 
time="22:56:54" level=info msg="[BITRISE_CLI] - Setup" 
time="22:56:54" level=info msg="Detected OS: darwin" 
time="22:56:54" level=info msg="Doing OS X specific setup" 
time="22:56:54" level=info msg="Checking required tools..." 

time="22:56:58" level=warning msg="brew update returned an error:" 
time="22:56:58" level=warning msg="fatal: ref HEAD is not a symbolic ref\nError: Failure while executing: git symbolic-ref --short HEAD" 
time="22:56:58" level=fatal msg="Setup failed: Homebrew not installed or has some issues. Please fix these before calling setup again. Err:Failed to: brew update" 
Error: bitrise: failed
```

viktorbenei:
@dunn so we can't call a `brew update` in the test - we'll fix this soon and submit a new version. We modified the setup process of `bitrise` - this is similar to a `brew doctor` call and performed for every version. Because `bitrise` depends on `brew` (it uses it for dynamic/on-the-fly dependency management) this "setup" process now calls a `brew update` and then a `brew doctor` for a proper, issue free setup.

Unfortunately this can't be done on your CI server, so we'll probably add a setup flag which will skip the `brew update` and `brew doctor` calls.

mikemcquaid:
> Because bitrise depends on brew (it uses it for dynamic/on-the-fly dependency management) this "setup" process now calls a brew update and then a brew doctor for a proper, issue free setup.

I'm not sure how I feel about a tool which runs Homebrew commands without asking or notifying the user being in Homebrew at all. It's also apparently a "technology preview" which points towards it being an unstable version.

viktorbenei:
@mikemcquaid we'll remove the technology preview label soon, as most of the development now goes into integration features, to make it possible to integrate bitrise with other tools, but the "user facing" features are stable and we don't plan to introduce new breaking features.

It uses Homebrew but the user specifies the dependencies - if there are no dependencies listed by the user it won't use homebrew at all.

You can find an example here: https://github.com/bitrise-io/bitrise/blob/master/_examples/experimentals/dependencies/bitrise.yml

The user or the step have to specifically list which dependencies it needs:

```
dependencies:
            - manager: brew
              name: cmake
```

mikemcquaid:
@viktorbenei There's no dependencies declared in the test so why is it calling `brew update`?

viktorbenei:
@mikemcquaid that's part of the `setup` - when you call `bitrise setup` or `bitrise run` the first time it performs a general system health check, similar to `brew doctor`, to ensure that everything's set up correctly

viktorbenei:
part of this setup is to call `brew update` and then a `brew doctor` in addition to just checking whether `brew` is installed or not

mikemcquaid:
I really don't think it should be calling `brew update` without the user authorising that action.

viktorbenei:
the problem with an optional `brew update` is that on Mountain Lion if you install `brew` but you don't have the Xcode CLT then `brew install` will break quite frequently. Actually, if you install `brew` on a clean Mountain Lion without the CLT and you try to call `brew update` it'll print an error, that `git` is not available and you should `brew install git` - but that fails, because there are no CLT available.

mikemcquaid:
I think it's probably best if you maintain this software in your own tap. I'm uncomfortable with having something which interacts with Homebrew like this in the our main repository. I'm sorry for including it in the first place, that was my mistake.

viktorbenei:
We actually added a new flag to our setup, calling `bitrise setup --minimal` will only check whether `brew` is installed but won't call `brew update` and `brew doctor`.

If you want to we can probably invert this, and don't call `brew update` & `brew doctor` by default, only if called with a `--check` or similar flag - but as I mentioned, on a new install of Mountain Lion it's not clear from a user's perspective to have to do a proper setup (that he actually needs Xcode CLT).

viktorbenei:
Removing the dependency management feature (which depends on `brew`) would defeat the purpose of `bitrise`, which is that you describe what you want to do in a `bitrise.yml` file, and if a step has a dependency it installs it automatically.

Optionally, if you agree with this, we could ask the user every time before performing a `brew install`, unless `bitrise` runs in a non-interactive mode (it's designed to be used on CI servers, in a non-interactive mode).

viktorbenei:
Of course @mikemcquaid if you're totally against a tool which uses `brew` at all we can't change that, but if there's any way to handle this issue we'd be open comply.

mikemcquaid:
`brew update` doesn't install the CLT on OS X beyond calling `git` which will trigger the pop-up so I'm not sure why you think it's necessary; if you want to download and install the CLT then you should do that instead.`brew doctor` failing doesn't indicate that a Homebrew install is broken just that it may have problems.

What happens if Homebrew isn't already installed on a system using `bitrise`?

viktorbenei:
@mikemcquaid I tested it on a fresh Mountain Lion VM and if you install `brew` it'll only print warning messages about the missing `git` and Xcode CLT. `brew update` fails with a message that you should `brew install git` but if you try to `brew install git` it'll fail because of the missing Xcode CLT.

What happens if Homebrew is not installed: `bitrise` prints a text that `brew` is not installed and a guide on how to install it, then it aborts the `setup` :  https://github.com/bitrise-io/bitrise/blob/master/bitrise/dependencies.go#L58

mikemcquaid:
> What happens if Homebrew is not installed: bitrise prints a text that brew is not installed and a guide on how to install it, then it aborts the setup : https://github.com/bitrise-io/bitrise/blob/master/bitrise/dependencies.go#L58

You should do the same thing if the CLT isn't installed or do something like https://github.com/boxen/boxen-web/blob/master/app/views/splash/script.sh.erb#L42-L83

`brew update` is not a good check as it'll actually make changes on the systems of end-users without them being aware that a) you've done so or b) that you were going to.

viktorbenei:
@mikemcquaid Thank you for the suggestion, we'll have a look at it.

My question would be, if we do the following changes, would you allow `bitrise` to be in the main `brew` Formula library:

- we'll remove `brew update` from the setup, and will never call it from `bitrise`
- we'll always ask for the user's permission to perform a `brew install`, except if `bitrise` specifically runs in a non-interactive, Continuous Integration mode (if the user calls `bitrise` with the `--ci` flag, like `bitrise --ci run ...`)

If we implement these changes would you consider to keep `bitrise` or would you still rather remove it from `brew`?

mikemcquaid:
Yup, those changes seem fine, thanks :+1:

viktorbenei:
Thanks @mikemcquaid , we already implemented these in our upstream repository, as this is actually a better approach in general - than you for the suggestions!

I'll create a new Pull Request once the new version is released.

PR #42898, opened 2015-08-13
graphene 1.2.6 (new formula)

tschoonj:


Library/Formula/graphene.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  depends_on "pkg-config" => :build
+  depends_on "glib"
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make"
```
DomT4: It'd be nice to run `"make", "check"` here as well.

tschoonj: @DomT4 I tried but there are two failures.

I submitted an upstream issue at https://github.com/ebassi/graphene/issues/41

DomT4: Looks like they added a fix & released a new version. If that works, feel free to update, add the `make check` bit and we should be golden.

tschoonj:
Just pushed the new 1.2.6 release and added the `make check` to the formula.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/b8b6fca8ebc058f9d1d8e7bfe66b671e11ccdcfc. Thanks @tschoonj!

Issue #41884, opened 2015-07-19
Use less jargon


MikeMcQuaid:
e.g. `keg`, `tap`, `bottle`, `Cellar`, etc.

chdiza:
In the end, the only thing that helps (if indeed anything helps) is documentation in an easy-to-find place or places.  I personally would like this stuff to go on the homepage and the man page.  We should just document what we already have, rather than change it only to have to carefully document it anyway.  It will save a lot of time and effort and we'll end up in the same place, namely, a place where users who read the docs will understand and the users who don't read the docs won't.  Also, it won't murder `git blame`.

I am willing to help improve our documentation of the existing terms.

mikemcquaid:
Sorry but I disagree. Documentation is a solution for when you cannot avoid using jargon (e.g. every way of explaining requires jargon). We definitely do not need to use jargon in many of the places we do.

dunn:
> users who read the docs will understand and the users who don't read the docs won't.

Yeah, I don't think those are the only possible outcomes.

chdiza:
> Yeah, I don't think those are the only possible outcomes.

They are the only two relevant outcomes.  The other two possible outcomes are read-but-not-understand and not-read-but-understand-anyway.  The latter is irrelevant to the present issue, and *once our docs are optimized*, anyone who genuinely reads these hypothetical docs and still doesn't get it certainly won't be helped by any de-jargoning.  They will have deeper issues, like not understanding what `$PATH` is or what a symlink is, or they will confuse "won't be linked" with "can't be linked".

scpeters:
I've always thought that the jargon is a fun part of the metaphor, but reducing user confusion should be a priority. Is this issue about reducing jargon in error messages or moving away from the metaphor entirely (i.e. will `/usr/local/Cellar` be changed to something like `/usr/local/HomebrewPackages`)?

mikemcquaid:
@scpeters We'd keep jargon for internal naming (for now, at least)

mxcl:
This isn’t documentation, these are the *names* of things. Names are hard. You have to pick something unique and explicit so that when you talk about it everyone understands what it is.

The reason I did not pick eg. package is because what is a package? In Package Management the word is thrown around to mean: the downloadable, or the installation, or the installation directory or the software-project.

If you must confuses tens of millions of people be renaming Homebrew *fundamentals*, at least pick words that are not *more* ambiguous. OK?

xu-cheng:
[...]

PR #42900, opened 2015-08-13
add Formula.racks

xu-cheng:
Also exclude symlink when enumerating installed racks.

cc @mikemcquaid  @viktorahlstrom

### DIFF
--- a/Library/Homebrew/cmd/cleanup.rb
+++ b/Library/Homebrew/cmd/cleanup.rb
@@ -31,12 +31,8 @@ module Homebrew
   end
 
   def cleanup_cellar
-    Formula.racks.each do |rack|
-      begin
-        cleanup_formula Formulary.from_rack(rack)
-      rescue FormulaUnavailableError, TapFormulaAmbiguityError
-        # Don't complain about directories from DIY installs
-      end
+    Formula.installed.each do |formula|
+      cleanup_formula formula
     end
   end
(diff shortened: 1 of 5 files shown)


mikemcquaid:
:+1: except for naming; I don't know if it's obvious what `racks` does or means.

xu-cheng:
The name is a little tricking. I think we should address it in https://github.com/Homebrew/homebrew/issues/41884 

But if we have a glossary in the document, `racks` seem to be the best name out there.

mikemcquaid:
I don't agree here. I don't really understand what a "rack" is and I've been working on Homebrew for 6+ years now. We don't really expose that name to users.

xu-cheng:
What name do you suggestion here? In all, I would like to avoid a too general name, so we won't end up use a whole sentence instead of a word when referring to this in the future discussion.

mikemcquaid:
`cellar_subdirs`

xu-cheng:
I think that would be even worse and more confusing name, especially we already have methods like `Formulary#from_rack` and `Formulary#to_rack`

mikemcquaid:
Ok, up to you. I'd still like to move away from `rack` as terminology as I don't think it's obvious what it means (and I think having a glossary to have to explain to Homebrew maintainers that "rack" means "Cellar subdirectories" is weird)

PR #42903, opened 2015-08-13
libgxps 0.2.3

nijikon:


Library/Formula/libgxps.rb
```diff
@@ -14,12 +15,13 @@ class Libgxps < Formula
...
     url "https://github.com/GNOME/libgxps.git"
 
-    depends_on "autoconf" => :build
-    depends_on "automake" => :build
-    depends_on "libtool" => :build
     depends_on "gnome-common" => :build
     depends_on "gtk-doc" => :build
   end
+
+  depends_on "autoconf" => :build
```
DomT4: Why are these necessary for the stable version now?

Library/Formula/libgxps.rb
```diff
@@ -1,8 +1,9 @@
 class Libgxps < Formula
   desc "GObject based library for handling and rendering XPS documents"
   homepage "https://live.gnome.org/libgxps"
-  url "http://ftp.gnome.org/pub/gnome/sources/libgxps/0.2/libgxps-0.2.2.tar.xz"
-  sha256 "39d104739bf0db43905c315de1d8002460f1a098576f4418f69294013a5820be"
+  url "https://download.gnome.org/sources/libgxps/0.2/libgxps-0.2.3.tar.xz"
+  mirror "https://github.com/GNOME/libgxps/archive/0.2.3.tar.gz"
```
dunn: This mirror has a different `sha256`: `869c8b7ca524b1357c97a7b5b8d9827e04f77a084e6fbb0e6509d8e39f6e4148`.

nijikon: You are right, I did use ```tar.gz``` for the url, but then after ```audit --strict``` I switched it to ```https://download.gnome.org```. Mirror removed.

Library/Formula/libgxps.rb
nijikon: Hehe, don't ask me ;)

```
==> make install
Making install in libgxps
 cd .. && /bin/sh /private/tmp/libgxps20150814-4381-12ebym3/libgxps-0.2.3/missing automake-1.15 --foreign libgxps/Makefile
sh: autoconf: command not found
automake-1.15: error: autoconf failed with exit status: 127
WARNING: 'automake-1.15' is missing on your system.
         You should only need it if you modified 'Makefile.am' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'automake' program is part of the GNU Automake package:
         <http://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <http://www.gnu.org/software/autoconf>
         <http://www.gnu.org/software/m4/>
         <http://www.perl.org/>
make[1]: *** [Makefile.in] Error 127
```

Library/Formula/libgxps.rb
```diff
@@ -30,6 +31,9 @@ class Libgxps < Formula
   depends_on "little-cms2" => :recommended
   depends_on "gtk+" => :optional
 
+  # Patch from developer, can be removed with the next release
+  patch :DATA
```
mikemcquaid: Can we get a link here? Thanks!

DomT4: Ping.

nijikon: Will update this to ```0.2.3.1``` it has the patch incorporated.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/e357910c9cd8dde9fae5b4bf861a48635a8ed276. Thank you @nijikon!

PR #42905, opened 2015-08-13
exim 4.86

nijikon:

PR #42911, opened 2015-08-13
Remove berkeley db4

nijikon:
This is an attempt to remove ```berkeley-db4```. I already switch ```exim``` to ```berkeley-db``` in this #42905. The only package left is ```nvi```, I was able to build it with this dependency, but I do get ```BDB0635 DB_CREATE must be specified to create databases.```, this does not happen with ```berkeley-db4```. Please share what you think about this idea.

### DIFF
--- a/Library/Formula/jigdo.rb
+++ b/Library/Formula/jigdo.rb
@@ -14,7 +14,7 @@ class Jigdo < Formula
 
   depends_on "pkg-config" => :build
   depends_on "wget" => :recommended
-  depends_on "berkeley-db4"
+  depends_on "berkeley-db"
   depends_on "gtk+"
 
   # Use MacPorts patch for compilation on 10.9; this software is no longer developed.
@@ -33,6 +33,6 @@ class Jigdo < Formula
   end
 
   test do
-    system "#{bin}/jigdo-file", "-h"
+    assert_match "version #{version}", shell_output("#{bin}/jigdo-file -v")
   end
 end
--- a/Library/Formula/nvi.rb
+++ b/Library/Formula/nvi.rb
@@ -4,7 +4,7 @@ class Nvi < Formula
   url "https://mirrors.ocf.berkeley.edu/debian/pool/main/n/nvi/nvi_1.81.6.orig.tar.gz"
   sha256 "8bc348889159a34cf268f80720b26f459dbd723b5616107d36739d007e4c978d"
 
-  depends_on "berkeley-db4"
+  depends_on "berkeley-db"
 
   # Patches per MacPorts
   # The first corrects usage of BDB flags.
@@ -21,6 +21,13 @@ class Nvi < Formula
   patch :p0 do
     url "https://raw.githubusercontent.com/Homebrew/patches/8ef45e8b/nvi/patch-ex_script.c.diff"
     sha256 "742c4578319ddc07b0b86482b4f2b86125026f200749e07c6d2ac67976204728"
+  end
+
+  # support berkeley > 4, patch extracted from debian package
+  # http://http.debian.net/debian/pool/main/n/nvi/nvi_1.81.6-11.debian.tar.gz
+  patch do
+    url "https://gist.githubusercontent.com/nijikon/c77bd170ff76dfe6cea4/raw/76369bda9d345c8d027766c4f892a5c82de0f493/nvi-db4.patch"
+    sha256 "33f707366757cc83dddf39251f64b45d6549a07ad21bc0fcc66ed6fc5d2b3964"
   end
 
   def install
--- a/Library/Formula/gnu-cobol.rb
+++ b/Library/Formula/gnu-cobol.rb
@@ -1,7 +1,7 @@
 class GnuCobol < Formula
   desc "Implements much of the COBOL 85 and COBOL 2002 standards"
   homepage "http://www.opencobol.org/"
-  revision 1
+  revision 2
 
   stable do
     url "https://downloads.sourceforge.net/project/open-cobol/gnu-cobol/1.1/gnu-cobol-1.1.tar.gz"
@@ -30,7 +30,7 @@ class GnuCobol < Formula
 
   depends_on "autoconf" => :build
   depends_on "automake" => :build
-  depends_on "berkeley-db4"
+  depends_on "berkeley-db"
   depends_on "gmp"
   depends_on "gcc"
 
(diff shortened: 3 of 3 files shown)


nijikon:
Maybe we could upgrade ```nvi``` to https://github.com/bentley/nvi2?

lembacon:
Actually, the OpenLDAP formula in [homebrew-dupes](https://github.com/Homebrew/homebrew-dupes/blob/master/openldap.rb) optionally depends on berkeley-db4.

nijikon:
Actually I can see that upgrade for ```openldap``` can't happen due to ```Berkeley DB version 6.0.20 and later uses a software license that is incompatible with LDAP technology and should not be used with OpenLDAP.```

DomT4:
Unless `berkeley-db4` is horribly broken in some way, we're unlikely to remove it. It still does ~100 bottled downloads a week on average, which whilst being low isn't nothing. As for stuff which uses `berkeley-db4`, if it can use a more recent version, I'm fine with that.

nijikon:
It's not broken. I'm curious what would be the download rate if we apply these patches. Actually do we have stats like, why it was downloaded, for example as a dependency to what?

DomT4:
No. Homebrew doesn't track that sort of thing (Or very much of anything really) and Bintray only knows "We got a request for this bottle from this IP at this time and date".

Library/Formula/python.rb
```diff
@@ -28,7 +29,7 @@ class Python < Formula
   depends_on "gdbm" => :recommended
   depends_on "openssl"
   depends_on "homebrew/dupes/tcl-tk" => :optional
-  depends_on "berkeley-db4" => :optional
+  depends_on "berkeley-db" => :optional
```
tdsmith: I'm very surprised this works. Can you explain how you tested this change?

nijikon:
```
balrog:bsddb3-6.1.0 n$ python test2.py
Running tests from /Users/n/bsddb3-6.1.0/build

-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
Berkeley DB 6.1.26: (June 16, 2015)
bsddb.db.version():   (6, 1, 26)
bsddb.db.full_version(): ('Berkeley DB 12c Release 1, library version 12.1.6.1.26: (June 16, 2015)', 12, 1, 6, 1, 26)
bsddb.db.__version__: 6.1.0
py module:            /Users/n/bsddb3-6.1.0/build/lib.macosx-10.10-x86_64-2.7/bsddb3/__init__.pyc
extension module:     /Users/n/bsddb3-6.1.0/build/lib.macosx-10.10-x86_64-2.7/bsddb3/__init__.pyc
Test working dir:     /var/folders/ph/1ry8xt293gbc8jtc639mp_9c0000gn/T/z-Berkeley_DB
python version:       2.7.10 (default, Aug 14 2015, 09:10:39) [GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] 64bit
My pid:               54223
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
----------------------------------------------------------------------
Ran 499 tests in 119.722s

OK
```

tdsmith: That's a third-party extension module. This option builds a module named `bsddb`. https://docs.python.org/2/library/bsddb.html

Library/Formula/python.rb
```diff
@@ -1,10 +1,11 @@
 class Python < Formula
   desc "Interpreted, interactive, object-oriented programming language"
   homepage "https://www.python.org"
-  head "https://hg.python.org/cpython", :using => :hg, :branch => "2.7"
   url "https://www.python.org/ftp/python/2.7.10/Python-2.7.10.tgz"
   sha256 "eda8ce6eec03e74991abb5384170e7c65fcd7522e409b8e83d7e6372add0f12a"
-  revision 2
+  revision 3
```
DomT4: Even if we do change this, per Tim's comment below, we don't need to bump the revision for a non-default option.

Library/Formula/python.rb
nijikon: I assume that we don't care that it is ```Deprecated since version 2.6: The bsddb module has been removed in Python 3.``` We still need to support this, correct?

Library/Formula/python.rb
nijikon: Clear.

bfontaine:

DomT4:
I presume Tim's comment on the Python issue may still be relevant?

nijikon:
I think so, maybe for now we will just bump what we can?

mikemcquaid:
@nijikon Makes sense to me; incremental progress is always good :+1:

nijikon:
We can merge this for now. I will handle exim in a separate PR.

Library/Formula/jigdo.rb
```diff
@@ -33,6 +33,6 @@ def install
   end
 
   test do
-    system "#{bin}/jigdo-file", "-h"
+    assert_match "version #{version}", shell_output("#{bin}/#{name} -v")
```
mikemcquaid: Would rather not use `#{name}` here; just hard-code it.

Library/Formula/nvi.rb
```diff
@@ -23,6 +23,12 @@ class Nvi < Formula
     sha256 "742c4578319ddc07b0b86482b4f2b86125026f200749e07c6d2ac67976204728"
   end
 
+  # support berkeley > 4, patch from debian
+  patch do
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

mikemcquaid: Also: the original Debian patch URL would be useful too.

bfontaine:
Ping re Mike’s comment?

Library/Formula/nvi.rb
nijikon: Patch is taken from the package, I emailed maintainers about this issue.

Library/Formula/jigdo.rb
nijikon: Done.

nijikon:
@mikemcquaid anything else?

Library/Formula/nvi.rb
mikemcquaid: Can you just add a comment saying what you've said to me here, thanks!

mikemcquaid: and include the URL

nijikon:
@mikemcquaid ping

mikemcquaid:
Sorry about the delay. Thanks again @nijikon!

PR #42912, opened 2015-08-13
md5sha1sum: explicitly set openssl paths

mistydemeo:
configure has a hardcoded set of include and library paths, and fails if it can't find openssl in any of them. This causes it to pick up the wrong openssl on older OS Xs, and to fail to pick up any on 10.11.

It actually links against the correct openssl on older OS Xs, so the revision isn't being bumped and bottles aren't being rebuilt.

### DIFF
--- a/Library/Formula/md5sha1sum.rb
+++ b/Library/Formula/md5sha1sum.rb
@@ -16,6 +16,10 @@ class Md5sha1sum < Formula
   depends_on "openssl"
 
   def install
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

DomT4:
The amount of software that bails out if it can't find an OpenSSL in `/usr` even if one is present somewhere else along the `$PATH` is a tad worrying. Hopefully we won't hit too much more of that with 10.11.

PR #42050, opened 2015-07-23
giflib: fix "ld: library not found for -lSM"

bashu:
I got this on os x 10.6.8 :
```
==> Downloading https://downloads.sourceforge.net/project/giflib/giflib-4.x/giflib-4.2.3.tar.bz2
Already downloaded: /Library/Caches/Homebrew/giflib-4.2.3.tar.bz2
==> ./configure --prefix=/usr/local/Cellar/giflib/4.2.3
==> make install
libtool: link: gcc-4.2 -dynamiclib  -o .libs/libgif.4.dylib  .libs/dgif_lib.o .libs/egif_lib.o .libs/gif_font.o .libs/gif_hash.o .libs/gifalloc.o .libs/gif_err.o   -L/usr/X11/lib -lSM -lICE -lX11  -O2   -install_name  /usr/local/Cellar/giflib/4.2.3/lib/libgif.4.dylib -compatibility_version 6 -current_version 6.7 -Wl,-single_module
ld: library not found for -lSM
collect2: ld returned 1 exit status
make[1]: *** [libgif.la] Error 1
make: *** [install-recursive] Error 1
```

PR #42921, opened 2015-08-14
giflib: disable default X11 sniffing

DomT4:
Giflib automatically sniffs for X11 on the system during compile, and can throw linking errors if it detects its availability but can't link to it.

Closes #42050.


mikemcquaid:
:+1:

PR #42924, opened 2015-08-14
gcc.rb: fix libgccjit linker flags

davidmalcolm:
This patch to homebrew adds a patch needed by libgccjit on OS X
that I've committed to upstream gcc here:
  https://gcc.gnu.org/ml/jit/2015-q3/msg00148.html

I've not personally tested it (I don't have an OS X box), but it is
identical to the one posted here, reported as necessary:
  https://gcc.gnu.org/ml/jit/2015-q3/msg00147.html

[I'm the upstream maintainer of libgccjit within gcc]


mikemcquaid:

PR #42927, opened 2015-08-14
cmake 3.3.1

Chilledheart:


afh:
The `babel` and `jinja` resources can also be updated, see 0763b7ffb300386b8a536de22327270980524e78.

dunn:
Thanks, @afh, taking care of that in #42963.

PR #42933, opened 2015-08-14
postgresql: 9.5alpha2 devel

guoxiao:


xu-cheng:

Issue #42916, opened 2015-08-13
Cannot update brew


yudaprama:
brew update
```
error: unknown option `short'
usage: git symbolic-ref [options] name [ref]

    -q, --quiet           suppress error message for non-symbolic (detached) refs
    -m <reason>           reason of the update

Error: Failure while executing: git symbolic-ref --short HEAD
```

OSX Mountain Lion

bfontaine:
What is the output of `brew doctor`?

yudaprama:
brew doctor

```
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry and just ignore them. Thanks!

Warning: A newer Command Line Tools release is available.
The standalone package can be obtained from
  https://developer.apple.com/downloads
or it can be installed via Xcode's preferences.

Warning: "config" scripts exist outside your system or Homebrew directories.
`./configure` scripts often look for *-config scripts to determine if
software packages are installed, and what additional flags to use when
compiling and linking.

Having additional scripts in your path can confuse software installed via
Homebrew if the config script overrides a system or Homebrew provided
script of the same name. We found the following "config" scripts:

  /Users/user/.node/bin/reapp-config

Warning: An outdated version of Git was detected in your PATH.
Git 1.7.10 or newer is required to perform checkouts over HTTPS from GitHub.
Please upgrade: brew upgrade git

Warning: Some installed formula are missing dependencies.
You should `brew install` the missing dependencies:

    brew install fribidi libass libvpx xvid

Run `brew missing` for more details.

Warning: Some directories in your path end in a slash.
Directories in your path should not end in a slash. This can break other
doctor checks. The following directories should be edited:
    /Users/user/.node/bin/

Warning: /usr/bin occurs before /usr/local/bin
This means that system-provided programs will be used instead of those
provided by Homebrew. The following tools exist at both paths:

    erb
    gem
    irb
    rake
    rdoc
    ri
    ruby

Consider setting your PATH so that /usr/local/bin
occurs before /usr/bin. Here is a one-liner:
    echo 'export PATH="/usr/local/bin:$PATH"' >> ~/.bash_profile

Warning: Your Xcode (5.0.2) is outdated
Please update to Xcode 5.1.1.
Xcode can be updated from
  https://developer.apple.com/xcode/downloads/

Warning: Your Homebrew is outdated.
You haven't updated for at least 24 hours. This is a long time in brewland!
[...]

PR #42934, opened 2015-08-14
doctor: suggest to install git if the system one is outdated

bfontaine:
This is related to #42916, where the user doesn’t have the brewed `git` but `brew doctor` suggest them to `upgrade` it.

### DIFF
--- a/Library/Homebrew/cmd/doctor.rb
+++ b/Library/Homebrew/cmd/doctor.rb
@@ -862,11 +862,14 @@ class Checks
     # https://help.github.com/articles/https-cloning-errors
     `git --version`.chomp =~ /git version ((?:\d+\.?)+)/
 
-    if $1 && Version.new($1) < Version.new("1.7.10") then <<-EOS.undent
-    An outdated version of Git was detected in your PATH.
-    Git 1.7.10 or newer is required to perform checkouts over HTTPS from GitHub.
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

Library/Homebrew/cmd/doctor.rb
```diff
@@ -862,10 +862,12 @@ def __check_git_version
     # https://help.github.com/articles/https-cloning-errors
     `git --version`.chomp =~ /git version ((?:\d+\.?)+)/
 
+    git_upgrade_cmd = Formula["git"].installed? ? "upgrade" : "install"
```
xu-cheng: You may want to change it to `Formula["git"].any_version_installed?`.

xu-cheng: Also, maybe better to put this line inside if condition below.

bfontaine: @xu-cheng Thanks, I didn’t know about the `any_version_installed?`.

xu-cheng: Oh I only added that method no long ago. The thing is `Formula#installed_prefix` and `Formula#installed?` are very misleading. I wonder should we deprecate them.

PR #42937, opened 2015-08-14
legit 0.2.0

bogem:


Library/Formula/legit.rb
```diff
@@ -1,8 +1,8 @@
 class Legit < Formula
   desc "Command-line interface for Git, optimized for workflow simplicity"
   homepage "http://www.git-legit.org/"
-  url "https://github.com/downloads/kennethreitz/legit/legit-v0.1.0-darwin-x86.tar.bz2"
-  version "0.1.0"
+  url "https://github.com/kennethreitz/legit/archive/v0.2.0.tar.gz"
+  version "0.2.0"
   sha256 "9b25a4c8e12d6703627ba24f4547a7d4350ca1ef9fa44415512b1bcc57b06506"
```
bfontaine: Please update the `sha256` as well.

Library/Formula/legit.rb
```diff
@@ -1,8 +1,8 @@
 class Legit < Formula
   desc "Command-line interface for Git, optimized for workflow simplicity"
   homepage "http://www.git-legit.org/"
-  url "https://github.com/downloads/kennethreitz/legit/legit-v0.1.0-darwin-x86.tar.bz2"
-  version "0.1.0"
+  url "https://github.com/kennethreitz/legit/archive/v0.2.0.tar.gz"
+  version "0.2.0"
```
bfontaine: You can probably remove this line (Homebrew detects the version from the URL).

bogem:
Ok. Will update soon

bogem:
Updated

bfontaine:
The downloaded archive doesn’t contain a `legit` executable; only a `legit` directory, which is why the install fails because it tries to install `legit` (= the directory) under `bin/`.

bogem:
I don't know, where this executable :(

dunn:
For some reason the old version of the formula was using a pre-build binary instead of compiling the `legit` executable.  I went ahead and fixed that in b6459f14d4ef56f10d20ed98d8eee9876b92627f.

Thank you for the submission! 💙

bogem:
@dunn Thank you!!! 💙

PR #42939, opened 2015-08-14
boot2docker 1.8.0

bfontaine:
This will be the last update to `boot2docker` as the project is now deprecated in favor of [Docker Machine](https://docs.docker.com/machine/).

### DIFF
--- a/Library/Formula/boot2docker.rb
+++ b/Library/Formula/boot2docker.rb
@@ -4,7 +4,7 @@ class Boot2docker < Formula
   # Boot2docker and docker are generally updated at the same time.
   # Please update the version of docker too
   url "https://github.com/boot2docker/boot2docker-cli.git",
-    :tag => "v1.7.1", :revision => "8fdc6f573bf08149b6311681800d55fda6e19e71"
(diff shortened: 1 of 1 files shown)


bfontaine:
ML:

```
      fatal: ref HEAD is not a symbolic ref
Error: Failure while executing: /usr/bin/curl -f#LA Homebrew\ 0.9.5\ (Ruby\ 1.8.7-358;\ OS\ X\ 10.8.5) https://github.com/Homebrew/homebrew/pull/42939.patch -o /Library/Caches/Homebrew/42939.patch --silent
```

md5:
This should probably be updated to 1.8.1.

md5:
Scratch that. Looks like `boot2docker/boot2docker` has a 1.8.1 tag, but not `boot2docker/boot2docker-cli`.

bfontaine:
Yeah that what I thought because the latest Docker is 1.8.1 but they haven’t tagged a 1.8.1 release. They expect people to move to `docker-machine` instead.

md5:
Thanks @bfontaine. I hadn't noticed that the `boot2docker` deprecation had finally been made official.

mikemcquaid:
:+1:

PR #42940, opened 2015-08-14
Warn when migration is needed

mikemcquaid:
If someone uses e.g. `git pull` instead of `brew update` then they can see errors from `outdated` (which bubbles to `upgrade`) when a formula hasn't been migrated yet. Instead of erroring out throw a more helpful exception telling them to manually run `migrate`.

Also, on `brew install` in the same situation: be a bit clearer if a formula hasn't been migrated rather than just saying it isn't linked.

CC @xu-cheng @vladshablinsky

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -71,7 +71,13 @@ module Homebrew
 
         if f.installed?
           msg = "#{f.full_name}-#{f.installed_version} already installed"
-          msg << ", it's just not linked" unless f.linked_keg.symlink? || f.keg_only?
+          unless f.linked_keg.symlink?
+            if Pathname.new("#{HOMEBREW_CELLAR}/#{f.oldname}").exist?
+              msg << ", it's just not migrated"
+            elsif !f.keg_only?
+              msg << ", it's just not linked"
+            end
+          end
           opoo msg
         else
           formulae << f
--- a/Library/Homebrew/cmd/outdated.rb
+++ b/Library/Homebrew/cmd/outdated.rb
@@ -1,5 +1,6 @@
 require "formula"
 require "keg"
+require "migrator"
 
 module Homebrew
   def outdated
@@ -16,6 +17,13 @@ module Homebrew
     formulae.map do |f|
       all_versions = []
       older_or_same_tap_versions = []
+
+      if f.oldname && !f.rack.exist?
+        if Pathname.new("#{HOMEBREW_CELLAR}/#{f.oldname}").exist?
+          raise Migrator::MigrationNeededError.new(f)
+        end
+      end
+
       f.rack.subdirs.each do |dir|
         keg = Keg.new dir
         version = keg.version
--- a/Library/Homebrew/migrator.rb
+++ b/Library/Homebrew/migrator.rb
@@ -4,6 +4,15 @@ require "tab"
 require "tap_migrations"
 
 class Migrator
+  class MigrationNeededError < RuntimeError
+    def initialize(formula)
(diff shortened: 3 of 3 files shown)


Library/Homebrew/cmd/outdated.rb
```diff
@@ -16,6 +17,12 @@ def outdated_brews(formulae)
     formulae.map do |f|
       all_versions = []
       older_or_same_tap_versions = []
+
+      old_rack = Pathname.new("#{HOMEBREW_CELLAR}/#{f.oldname}")
```
vladshablinsky: `f.oldname` can be `nil`.

mikemcquaid: :+1:

Library/Homebrew/cmd/install.rb
```diff
@@ -71,7 +71,13 @@ def install
 
         if f.installed?
           msg = "#{f.full_name}-#{f.installed_version} already installed"
-          msg << ", it's just not linked" unless f.linked_keg.symlink? || f.keg_only?
+          if !f.linked_keg.symlink?
```
bfontaine: Any reason to use `if !` instead of `unless` here?

mikemcquaid: Nope :grinning:

Library/Homebrew/cmd/install.rb
```diff
@@ -71,7 +71,13 @@ def install
 
         if f.installed?
           msg = "#{f.full_name}-#{f.installed_version} already installed"
-          msg << ", it's just not linked" unless f.linked_keg.symlink? || f.keg_only?
+          unless f.linked_keg.symlink?
+            if Pathname.new("#{HOMEBREW_CELLAR}/#{f.oldname}").exist?
+              msg << ", it's just not migrated"
```
xu-cheng: Revisit this PR, I think this is a dead code. The reason is if `f.installed?` returns true, the formula has to be migrated.

Library/Homebrew/cmd/outdated.rb
```diff
@@ -16,6 +17,13 @@ def outdated_brews(formulae)
     formulae.map do |f|
       all_versions = []
       older_or_same_tap_versions = []
+
+      if f.oldname && !f.rack.exist?
+        if Pathname.new("#{HOMEBREW_CELLAR}/#{f.oldname}").exist?
+          raise Migrator::MigrationNeededError.new(f)
+        end
+      end
```
xu-cheng: There are some edge cases here as there is a chance that `HOMEBREW_CELLAR/f.oldname` is the rack of another formula.

Library/Homebrew/cmd/install.rb
mikemcquaid: I added this because I've seen this code. It can happen if you `git pull` rather than `brew update`.

Library/Homebrew/cmd/outdated.rb
mikemcquaid: Yup.

Library/Homebrew/cmd/install.rb
xu-cheng: No, I don't think so. Because if `f.installed?` returns true, then `HOMEBREW_CELLAR/newname` will definitely exist. In this state, it cannot be pre-migration.

xu-cheng: If you run `git pull`, f.installed? will return false.

mikemcquaid: Ok, I'll look more.

PR #42941, opened 2015-08-14
rename app-engine formulae

vladshablinsky:

### DIFF
--- a/Library/Formula/app-engine-go-32.rb
+++ b/Library/Formula/app-engine-go-32.rb
@@ -1,4 +1,4 @@
-class GoAppEngine32 < Formula
+class AppEngineGo32 < Formula
   desc "Google App Engine SDK for Go!"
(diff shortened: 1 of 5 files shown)


xu-cheng:
:+1:

mikemcquaid:
Can this be rebased on `master`? Thanks!

vladshablinsky:
Done. `test do` block should be added for each formula in `brew audit --strict`

mikemcquaid:
Thanks @vladshablinsky!

xu-cheng:
@vladshablinsky This PR may break these formulae. I just found `name` variable is used in `install`. By changing the names, formulae's installation are changed in the process.

vladshablinsky:
@xu-cheng I can change `name` to `oldname` in `install`. What should we do in  cases like this?

PR #42942, opened 2015-08-14
beanstalkd: rename from beanstalk

vladshablinsky:

### DIFF
--- a/Library/Aliases/beanstalkd
+++ b/Library/Aliases/beanstalkd
@@ -1,1 +1,0 @@
-../Formula/beanstalk.rb
--- a/Library/Formula/beanstalkd.rb
+++ b/Library/Formula/beanstalkd.rb
@@ -1,4 +1,4 @@
-class Beanstalk < Formula
+class Beanstalkd < Formula
   desc "Generic work queue originally designed to reduce web latency"
   homepage "https://kr.github.io/beanstalkd/"
   url "https://github.com/kr/beanstalkd/archive/v1.10.tar.gz"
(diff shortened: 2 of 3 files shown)


xu-cheng:

mikemcquaid:
Here we're inconsistent with all other package managers (and have an alias already): https://kr.github.io/beanstalkd/download.html

xu-cheng:
IMHO, alias should be enough. It's simple, robust and less likely to cause troubles.

mikemcquaid:
@xu-cheng What would need to be done code-wise for you to be happy with simpler renames like this?

xu-cheng:
I won't suggest to rename like this util we sort out:
* revert formula rename.
* rename a renamed formula.

And frankly, you cannot find simpler (hence more robust) solution than alias. For example, there are some underlying problems in formula rename which are hard(or impossible) to solve:
* It's hard to determine whether a cellar is pre-migration or a complete different formula/diy install. To the best, we can only guess. This is the same reason we have `--force` flag around to bypass these guesses in case we guess wrong.
* It's hard to have robust auto migration. We can not rely on `brew update` because there are thousands of things can break auto migration. Therefore, `brew migrate` will be required in many situations. And this is bad for CI staff(especially considering Homebrew is the default package manager for Travis CI on OS X).

mikemcquaid:
:ok:. I think for me that means "we conservatively rename formulae for a while and get less conservative if we don't see any end-user bugs".

apjanke:
Ping. Want to pursue this or just close it out as "sticking with the alias"?

mikemcquaid:
@apjanke We've not seen end-uers bugs so let's persue this. @vladshablinsky could you open a new PR at homebrew-core for this? Thanks!

PR #42943, opened 2015-08-14
dbus: rename from d-bus

vladshablinsky:

### DIFF
--- a/Library/Aliases/dbus
+++ b/Library/Aliases/dbus
@@ -1,1 +1,0 @@
-../Formula/d-bus.rb
--- a/Library/Formula/dbus.rb
+++ b/Library/Formula/dbus.rb
@@ -1,4 +1,4 @@
-class DBus < Formula
+class Dbus < Formula
   # releases: even (1.8.x) = stable, odd (1.9.x) = development
   desc "Message bus system, providing inter-application communication"
   homepage "https://wiki.freedesktop.org/www/Software/dbus"
--- a/Library/Homebrew/formula_renames.rb
+++ b/Library/Homebrew/formula_renames.rb
@@ -1,4 +1,5 @@
 FORMULA_RENAMES = {
+  "d-bus" => "dbus",
   "libcppa" => "caf",
   "plt-racket" => "racket"
 }


xu-cheng:

mikemcquaid:
FWIW the Debian name: https://packages.debian.org/stable/mail/dbus

apjanke:
Ping here too. Like beanstalkd, do we want to actually do anything here?

IMHO seems like the best thing to do in both cases might be to do the rename to match Debian and the other package managers, but add an alias for the old name so no command calls break.

mikemcquaid:
Like beanstalkd: let's do the rename but @vladshablinsky could move it to homebrew-core. Thanks!

UniqMartin:
Not a show-stopper, but let's not forget that a rename has also implications for formula options that reference the renamed formula, e.g. because it is an optional/recommended dependency.

vladshablinsky:
@UniqMartin I'm not sure what you mean. I don't think we need change all the dependent formulae, it should work fine.

UniqMartin:
@vladshablinsky Thanks, that's good and what I was hoping for. What I wanted to express is that, to pick `qt5` as an example, its optional `d-bus` dependency will sooner or later be changed to `dbus` (I guess we don't want to rely on `<oldname>` resolving to `<newname>` indefinitely) at which point we would also have to add `--with-d-bus` as a deprecated option that gets mapped to `--with-dbus`. (I don't think formula renames cover this, or do they?)

What I was implying with my comment: Is getting rid of a single hyphen in proportion to these changes?

vladshablinsky:
@UniqMartin ah, I see. You're right, they don't cover the options renaming. I haven't thought about this case ever, and I'm not very good at Homebrew's options resolving process. It's a good note, though. I think it'll be possible to support resolving renamed options in a while, when renames are fully supported.

So, what do I do now? Rename the formula and add deprecated options?

UniqMartin:
I'm not sure extending the rename support to options is what we want. If too many things happen automatically and options that don't appear in the formula file are magically recognized (because they belong to a renamed dependency), I think things can become pretty confusing. But it's definitely something to keep in mind when renaming formulae because of the tight coupling between optional/recommended dependencies and options automatically created for them we have in Homebrew.

> So, what do I do now? Rename the formula and add deprecated options?

To be honest, I'm not sure if I can answer that. @mikemcquaid and @apjanke seem to be in favor of the rename, @xu-cheng seems to be against it and I'm a bit undecided, but leaning towards @xu-cheng's side. However, *if* we move forward with this rename, eventually the options will need to be adjusted too. (This currently affects `emacs`, `pulseaudio`, `qt`, and `qt5`. They all have an optional `d-bus` dependency.) Sorry that I'm not able to provide a more definitive answer!

mikemcquaid:
> So, what do I do now? Rename the formula and add deprecated options?

Yep, this is the right approach; we already have `deprecated_option` to handle this case.

xu-cheng:
Just want to point out that although we can migrate all official formulae with `deprecated_option` it would still break third party formulae.

vladshablinsky:
I've just thought about this: If it works with alias and both `--with-dbus` and `--with-d-bus` work fine, why wouldn't it work after rename?

UniqMartin:
I guess I was successful in raising awareness that renaming a formula involves more than just changing its name, though that might have been obvious to some. Other than that, I don't care enough about `d-bus` (or `dbus`) specifically to want to veto this rename.

mikemcquaid:
@UniqMartin A good point. It may be worth passing on this one then :

PR #42944, opened 2015-08-14
gperftools: rename from google-perftools

vladshablinsky:


xu-cheng:
:+1:

mikemcquaid:
Thanks @vladshablinsky!

PR #42945, opened 2015-08-14
mpich: rename from mpich2

vladshablinsky:


mikemcquaid:
Needs rebased on `master`

xu-cheng:
:+1:

PR #42946, opened 2015-08-14
offlineimap: rename from offline-imap

vladshablinsky:

### DIFF
--- a/Library/Aliases/offlineimap
+++ b/Library/Aliases/offlineimap
@@ -1,1 +1,0 @@
-../Formula/offline-imap.rb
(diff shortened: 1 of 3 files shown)


mikemcquaid:
Needs rebased on `master`

xu-cheng:

mikemcquaid:
FWIW the Debian name: https://packages.debian.org/stable/mail/offlineimap

nijikon:
@mikemcquaid is this still the case? I can do it in the referenced PR.

mikemcquaid:
@nijikon Yeh, let's rename, thanks :+1:

UniqMartin:
Closing because the formula has been renamed in #47638.

PR #42947, opened 2015-08-14
bitrise 0.9.10

viktorbenei:
test was modified to make it smaller (the test bitrise.yml file was a bit too verbose), and to use the new --minimal flag for setup, to skip calling `brew doctor`

### DIFF
--- a/Library/Formula/bitrise.rb
+++ b/Library/Formula/bitrise.rb
@@ -3,8 +3,8 @@ require "language/go"
 class Bitrise < Formula
   desc "Command-line automation tool"
   homepage "https://github.com/bitrise-io/bitrise"
-  url "https://github.com/bitrise-io/bitrise/archive/0.9.7.tar.gz"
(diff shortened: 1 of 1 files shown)


viktorbenei:
Based on our discussion with @mikemcquaid - https://github.com/Homebrew/homebrew/pull/42897 - we removed the `brew update` call from `brew setup` completely. It now only calls `brew doctor`, unless the `--minimal` flag is specified (in that case even `brew doctor` is skipped).
We now also explicitely ask the user's permission to `brew install` the dependencies defined in the user's `bitrise.yml` file, except if bitrise is running in Continuous Integration mode.

Full release changelog: https://github.com/bitrise-io/bitrise/releases/tag/0.9.10

mikemcquaid:
Great thanks @viktorbenei!

viktorbenei:
Thank you @mikemcquaid for the suggestions, these changes greatly improve the overall user experience!

mikemcquaid:
Thanks for the kind words @viktorbenei and the PR!

PR #42948, opened 2015-08-14
FoBiS.py 1.7.4 (new formula)

zbeekman:
A powerful build tool for modern Fortran

### DIFF
--- a/Library/Formula/fobis.rb
+++ b/Library/Formula/fobis.rb
@@ -1,0 +1,37 @@
+class Fobis < Formula
+  desc "KISS build tool for automaticaly building modern Fortran projects."
+  homepage "https://github.com/szaghi/FoBiS"
+  url "https://pypi.python.org/packages/source/F/FoBiS.py/FoBiS.py-1.7.4.tar.gz"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Test is failing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/31537/version=yosemite/testReport/junit/brew-test-bot/yosemite/test_fobis/

zbeekman:
oops, thanks for pinging me...

zbeekman:
To clarify why the tests were failing: I  had forgotten to specify the  gnu fortran compiler, and FoBiS.py defaults to the Intel compiler. This is installed on my system, so tests passed when I ran them locally, but is clearly not going to be installed on many systems. Since we require `:fortran` we can anticipate, AFAICT, that most users should have gfortran, and certainly testbot will. (FoBiS.py will work with any number of compilers, but in rare circumstances it is conceivable that users could NOT have gfortran installed, but use a different compiler through the `FC` environment variable. In this latter case, the test will fail on their system, due to the lack of gfortran, even though FoBiS.py is still useable.)

zbeekman:
Should be good to go now

mikemcquaid:

For future reference the preferred commit message format for new formulae is `fobis 1.7.4 (new formula)`.

PR #39868, opened 2015-05-18
Gtk+ goes quartz-only

tschoonj:
@mikemcquaid 

This branch represents an effort to remove the X11 dependency completely from gtk+ and its dependencies cairo and pango.

The rationale behind this move is:

1. It is a major pain for homebrew to support simultaneously the X11 and quartz backends of gtk+. Some formulas require the quartz backend explicitly (like gtk-mac-integration, which for this reason is still not included in homebrew). See also e.g. #38969 for a discussion about this (there are a lot more PRs and issues where this problem was brought up!)
2. In gtk+3, the X11 backend is not even supported anymore
3. X11 is very un-Mac OS X anyway and delivers an inferior user experience (IMHO)
4. ~~Are there any packages that really need the gtk+ X11 backend to work properly anyway?~~ Apparently `gtkglarea` does, it will be moved to the boneyard...

My plan is basically to:

1. Remove X11 from the dependency list of cairo, pango and gtk+. Modify the configure argument list to ensure that no X11 bindings will get built. Explicitly enable quartz support. Update the revision number of these formulas
2. Update the revision number of all formulas in Homebrew/homebrew that depend on gtk+. This will be the most work since many of these formulas will not be audit --strict compliant (missing tests).
3. Update the revision number of all formulas in all other Homebrew taps that depend on gtk+. Will create separate pull request for each tap which should be pulled in when this PR is merged in.
  1. science: Homebrew/homebrew-science#2311
  2. x11: Homebrew/homebrew-x11#83
  3. games: Homebrew/homebrew-games#268

It is possible that some formulas that depend only on cairo and/or pango will also need a revision bump if they depend explicitly on the cairo/pango dylibs that are linked to X11. Not sure how to identify these though.

TO DO list (will get longer...):

- [x] cairo
- [x] pango
- [x] gtk+
- [x] cairomm
- [x] pangomm
- [x] gtkmm
- [x] bokken: 1.7 currently does not build on Yosemite (#36052 and #32016). Since it uses pygtk, and not gtk+ directly there is no need to rebuild the package anyway. It should work as soon as the build problem is fixed.
- [x] diffuse
- [x] gabedit
- [x] gbdfed
- [x] gdmap
- [x] gnome-icon-theme: lost its gtk+ dependency by updating librsvg
- [x] goffice: lost its gtk+ dependency by updating librsvg
- [x] gqview
- [x] grsync
- [x] gst-python
- [x] gtk-engines
- [x] gtk-gnutella
[...]

PR #42949, opened 2015-08-14
cairo: enable xcb when xquartz is installed

afh:
Certain formula like [i3](https://github.com/Homebrew/homebrew-x11/blob/master/i3.rb) require cairo to be built with xcb, only when the cairo formula `depends_on :x11` does `./configure` find the necessary files to build cairo with xcb support.

Without this change `find $(brew --prefix cairo) -name cairo-xcb.h` finds nothing; with this change the header file is found and the appropriate code available in the cairo library.

### DIFF
--- a/Library/Formula/cairo.rb
+++ b/Library/Formula/cairo.rb
@@ -18,6 +18,7 @@ class Cairo < Formula
   option :universal
 
   depends_on "pkg-config" => :build
+  depends_on :x11 => :optional if MacOS.version > :leopard
   depends_on "freetype"
   depends_on "fontconfig"
   depends_on "libpng"
@@ -38,7 +39,7 @@ class Cairo < Formula
       --enable-quartz-image
     ]
 
-    args << "--enable-xcb=no" if MacOS.version <= :leopard
+    args << (build.with?("x11") ? "--enable-xcb=yes" : "--enable-xcb=no")
 
     system "./configure", *args
     system "make", "install"


Library/Formula/cairo.rb
```diff
@@ -23,6 +23,7 @@ class Cairo < Formula
   depends_on "libpng"
   depends_on "pixman"
   depends_on "glib"
+  depends_on :x11 if MacOS::XQuartz.installed?
```
DomT4: I can't imagine we'll re-add X11 support to Cairo, but particularly by default which this essentially does (Our CI have X11, so the bottles generated would be generated with X11 ties). This is actually harsher than our previous `depends_on :x11 => :recommended` dependency as this time there would be no opt-out for anyone with X11 installed.

CC @mikemcquaid @tschoonj

mistydemeo:
Maybe adding a `with-x11` option makes more sense? Something like:

```patch
diff --git a/Library/Formula/cairo.rb b/Library/Formula/cairo.rb
index cab073e..b274d7e 100644
--- a/Library/Formula/cairo.rb
+++ b/Library/Formula/cairo.rb
@@ -18,6 +18,7 @@ class Cairo < Formula
   option :universal
 
+  depends_on :x11 => :optional if MacOS.version > :leopard
   depends_on "freetype"
   depends_on "fontconfig"
   depends_on "libpng"
@@ -38,7 +39,7 @@ class Cairo < Formula
       --enable-quartz-image
     ]
 
-    args << "--enable-xcb=no" if MacOS.version <= :leopard
+    args << (build.with?("x11") ? "--enable-xcb=yes" : "--enable-xcb=no")
 
     system "./configure", *args
     system "make", "install"
```

tdsmith:
I would consider vendoring cairo in i3 in preference to adding an x11 option to the Homebrew/homebrew formula.

DomT4:
If someone volunteered to look after it, I wouldn't hate a `keg_only` X11 enabled Cairo/Pango in X11, but I suspect as was the case in Homebrew nobody really wants to keep on top of it long term.

afh:
@tdsmith How would that look? Is there already a formula that does something similar?

mikemcquaid:
The reason we removed the option is because it causes a mess for the rest of the GTK chain I believe if someone has enabled cairo with X11 support. CC @tschoonj for views here.

tschoonj:
Bringing X11 back into `cairo` is really the worst thing we could do here. The problem is that all cairo backends are compiled into a single dylib, which means that all software depending on cairo (Gtk: see #39868) will carry the X11 dependency with them.

Perhaps there is a way around this problem though: if you remove the `depends_on "cairo"` line in the `i3` formula but keep the X11 dependency, configure should pick up (untested) the cairo version that ships with XQuartz, which has xcb support. Worth a shot?

afh:
@tschoonj Good idea, unfortunately i3 depends on the pango formula, since pango isn't available from XQuartz, and the pango formula depends on cairo :disappointed:

tschoonj:
In that case perhaps cairo and pango could perhaps be included as resources in `i3`, to be compiled into static libraries only, which would then be linked into `i3`?

afh:
I may be missing something really obvious, but I don't see the problem of cairo depending on X11, iff the user has XQuartz installed. Can someone explain the issue here?

afh:
@tschoonj I'll give that a try

mikemcquaid:
> I may be missing something really obvious, but I don't see the problem of cairo depending on X11, iff the user has XQuartz installed. Can someone explain the issue here?

Conditional dependencies like these don't really work properly and, even if they did, the binary packages are built on a machine with XQuartz installed and need to be installed on machines where XQuartz isn't available.

afh:
@mikemcquaid Thanks. Yes, that makes sense.

afh:
I also like @mistydemeo's approach, what do the other collaborators think about it?

mikemcquaid:
I think it would be problematic for the reasons @tschoonj.

tschoonj:
I guess most people will end up with `cairo` on their system because they installed some formula that has `gtk+` or `gtk+3` as dependency. Now both these packages have a quartz-only backend, so there is no use for them to install `cairo` with an X11 dependency. In fact it may even be quite annoying if you plan on bundling gtk in an app bundle as it will force users of this app to install XQuartz though it's not even really going to be used. 

This is why I removed the X11 support from `cairo` a couple of months ago, as it seemed pointless to keep it around. I didn't really take into account that there would be software like `i3` that doesn't need gtk, but merely depends on cairo (and pango).

I think @mistydemeo's suggestion is probably the best as it has no impact on `gtk+` or `gtk+3`. Don't make it `:recommended` though as it would be useless for the vast majority of homebrew users.

mikemcquaid:
Ok, :cool: with me.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/342e5cb7a73ba65cc5cfc562714f607ee9d7c2c5. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #42906, opened 2015-08-13
Added formula for libspnav

justbuchanan:
Libspnav is a client library for interfacing with 3DConnexion's 3D input devices.

PR #42950, opened 2015-08-14
Added formula for spacenavd

justbuchanan:
Spacenavd is a userspace driver/daemon for interfacing with 3Dconnexion's 3d input devices.

### DIFF
--- a/Library/Formula/spacenavd.rb
+++ b/Library/Formula/spacenavd.rb
@@ -1,0 +1,46 @@
+class Spacenavd < Formula
+  desc "Userspace device driver/daemon for 3Dconnexion's 3D input devices."
+  homepage "http://spacenav.sourceforge.net/index.html"
+  url "https://downloads.sourceforge.net/project/spacenav/spacenav%20daemon/spacenavd%200.6/spacenavd-0.6.tar.gz"
+  sha256 "c2d203bf96c5a959590146a43fe5d6e5e8c5c38a8b2f55aa199d967d0d88d0ab"
+
+  option "with-x11", "Enable support for x11 communication."
+
+  depends_on :x11 => :optional
+
+  # There is a ticket filed for this project on sourceforge to fix this:
+  # https://sourceforge.net/p/spacenav/bugs/10/
+  patch :DATA
+
+  def install
+    args = %W[
+      --disable-debug
+      --disable-dependency-tracking
+      --disable-silent-rules
+      --prefix=#{prefix}
+    ]
+
+    args << "--disable-x11" if build.without? "x11"
+
+    system "./configure", *args
+    system "make", "install"
+  end
+
+  test do
+    system "spacenavd", "-h"
+  end
+end
+
+__END__
+diff --git a/src/dev_usb_darwin.c b/src/dev_usb_darwin.c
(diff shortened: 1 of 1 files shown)


Library/Formula/spacenavd.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    system "make", "install"
+  end
+
+  test do
+    system "spacenavd", "-h"
+  end
+end
+
+# There is a ticket filed for this project on sourceforge to fix this:
```
DomT4: Please move these notes to just above `patch :DATA` rather than down here.

Library/Formula/spacenavd.rb
```diff
@@ -0,0 +1,47 @@
...
+      --disable-debug
+      --disable-dependency-tracking
+      --disable-silent-rules
+      --prefix=#{prefix}
+    ]
+
+    args << "--disable-x11" if build.without? "x11"
+
+    system "./configure", *args
+
```
DomT4: No need for an empty line here.

Library/Formula/spacenavd.rb
```diff
@@ -0,0 +1,47 @@
...
+    system "make", "install"
+  end
+
+  test do
+    system "spacenavd", "-h"
```
DomT4: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

justbuchanan: It's tough to do much better here because you need to create a client to interface with this driver, but the client library is a separate package called [libspnav](https://github.com/Homebrew/homebrew/pull/42906).

Another thing that could be checked is to run the daemon and verify that it has written its pid to the right file (`/var/run/spnavd.pid`).  Unfortunately this requires `sudo`, which I assume isn't allowed in tests.

DomT4: If it's placing things in `/var` it should be set so it uses `HOMEBREW_PREFIX/"var"` instead as per our standard.

If there's a daemon you could do something like [this](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mpd.rb#L127-L138) but `test do` blocks aren't supposed to create or write any file outside of the `testpath`, so we'd need to redirect the `pid` creation if possible.

justbuchanan: Sorry for the delay, I just got done with a pretty busy weekend.

The problem with changing the file locations is that the client library used to communicate with the daemon also has these locations hard-coded.  I could modify the `libspnav` package as well as this one to change the paths, but anyone that is using the client library outside of homebrew (many packages just include the source in their repo) won't be connect to the homebrew-installed instance of `spacenavd`.  Thoughts?

mikemcquaid: It sounds like this may not be a good fit for Homebrew. I dunno.

bfontaine:
We’ll pass on this due to the limitation with `/var`; but feel free to maintain this in a tap!
https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md

PR #42951, opened 2015-08-14
opencc 1.0.3

stuwil:


mikemcquaid:
Probably need to set `ENV["LC_ALL"] = "en_US.UTF-8"` in the `test do` block to avoid the test failure.

xu-cheng:

PR #42954, opened 2015-08-14
which: don't fail on malformed paths in PATH

bfontaine:


bfontaine:
I added a failing test which shows the issue: it fails without this PR, succeed with it.

bfontaine:
I wonder if we should warn users in `brew doctor` for this.

Library/Homebrew/test/test_utils.rb
```diff
@@ -6,6 +7,30 @@ def test_put_columns_empty
     assert_silent { puts_columns [] }
   end
 
+  def test_which
+    cmd = Tempfile.new "foo"
```
xu-cheng: You don't need to use `Tempfile` in the test. Instead you can do:
```
def test_which
  dir = Pathname.new(mktmpdir)
  cmd = dir/"foo"
  cmd.chmod 0744
  assert_equal cmd, which("foo", dir)
ensure
  dir.rmtree
end
```

bfontaine: Thanks @xu-cheng, fixed. Any remark outside of that?

xu-cheng: No. You are all good :+1:

Issue #42953, opened 2015-08-14
URL rewrite on-the-fly (Insecure Redirect, ie. Sourceforge)


smarek:
Hi,

For case `HOMEBREW_NO_INSECURE_REDIRECT` set, SourceForge has like *no HTTPS/SSL enabled mirrors*, so this DownloadStrategy rewrites the `download.sourceforge.net` url to use `mirrorservice.org` (which has valid SSL certificate)

## Better idea

Because Homebrew provided should imho not prefer any specific mirror (what would be point of mirrors in that case, isn't it?) but qualified user should be able to provide custom Ruby module (not part of Homebrew github project) and on-the-fly rewrite URLs 

Given solution could be tweaked using ENV variable like `HOMEBREW_SOURCEFORGE_DOMAIN_REPLACE` which would be by default `"www.mirrorservice.org/sites/download.sourceforge.net/pub/sourceforge/"` but that is not really flexible.

## Original DownloadStrategy idea
```diff
diff --git a/Library/Homebrew/download_strategy.rb b/Library/Homebrew/download_strategy.rb
index 5243a75..dfbcf57 100644
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -388,6 +388,25 @@ class CurlApacheMirrorDownloadStrategy < CurlDownloadStrategy
   end
 end
 
+class CurlSourceForgeSecureMirrorDownloadStrategy < CurlDownloadStrategy
+  def fetch
+    if !ENV["HOMEBREW_NO_INSECURE_REDIRECT"].nil?
+      oldurl = @url
+      @url = https_mirror
+      ohai "Sourceforge HTTPS: #{oldurl}"
+      ohai "=> converted to #{@url}"
+    end
+    super
+  end
+
+  def https_mirror
+     re = /(^https?:\/\/)(.+\.net\/project\/)(.+)/
+     @url.gsub(re) do
+       $1 + "www.mirrorservice.org/sites/download.sourceforge.net/pub/sourceforge/" + $3[0] + "/" + $3[0,2] + "/" + $3
+     end
+  end
+end
+
 # Download via an HTTP POST.
 # Query parameters on the URL are converted into POST parameters
 class CurlPostDownloadStrategy < CurlDownloadStrategy
@@ -853,6 +872,8 @@ class DownloadStrategyDetector
       SubversionDownloadStrategy
[...]

PR #42955, opened 2015-08-14
SourceForge DownloadStrategy HTTPS URL rewrite proposal #42953

smarek:
As discussed in ticket https://github.com/Homebrew/homebrew/issues/42953

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -402,9 +402,8 @@ class CurlSourceForgeSecureMirrorDownloadStrategy < CurlDownloadStrategy
   end
 
   def https_mirror
-     re = /(^https?:\/\/)(.+\.net\/project\/)(.+)/
-     @url.gsub(re) do
-       $1 + "www.mirrorservice.org/sites/download.sourceforge.net/pub/sourceforge/" + $3[0] + "/" + $3[0,2] + "/" + $3
+     @url.gsub(%r{(^https?://)(?:.+)(.net/project/)(.+)}) do
+       "#{$1}www.mirrorservice.org/sites/download.sourceforge.net/pub/sourceforge/#{$3[0]}/#{$3[0,2]}/#{$3}"
      end
   end
 end
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -389,12 +389,13 @@ class CurlApacheMirrorDownloadStrategy < CurlDownloadStrategy
   end
 end
 
+# Will convert "https://downloads.sourceforge.net/project/..." URLs to mirrorservice.org
+# In case environment variable HOMEBREW_NO_INSECURE_REDIRECT is set, so the Sourceforge is not ultimately broken
 class CurlSourceForgeSecureMirrorDownloadStrategy < CurlDownloadStrategy
   def fetch
-    if !ENV["HOMEBREW_NO_INSECURE_REDIRECT"].nil?
-      oldurl = @url
+    unless ENV["HOMEBREW_NO_INSECURE_REDIRECT"].nil?
+      ohai "Sourceforge HTTPS: #{@url}"
       @url = https_mirror
-      ohai "Sourceforge HTTPS: #{oldurl}"
       ohai "=> converted to #{@url}"
     end
     super
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -389,6 +389,25 @@ class CurlApacheMirrorDownloadStrategy < CurlDownloadStrategy
   end
 end
 
+class CurlSourceForgeSecureMirrorDownloadStrategy < CurlDownloadStrategy
+  def fetch
+    if !ENV["HOMEBREW_NO_INSECURE_REDIRECT"].nil?
+      oldurl = @url
+      @url = https_mirror
+      ohai "Sourceforge HTTPS: #{oldurl}"
+      ohai "=> converted to #{@url}"
+    end
+    super
+  end
+
+  def https_mirror
+     re = /(^https?:\/\/)(.+\.net\/project\/)(.+)/
+     @url.gsub(re) do
+       $1 + "www.mirrorservice.org/sites/download.sourceforge.net/pub/sourceforge/" + $3[0] + "/" + $3[0,2] + "/" + $3
+     end
+  end
+end
+
 # Download via an HTTP POST.
 # Query parameters on the URL are converted into POST parameters
 class CurlPostDownloadStrategy < CurlDownloadStrategy
@@ -854,6 +873,8 @@ class DownloadStrategyDetector
       SubversionDownloadStrategy
     when %r{^https?://(.+?\.)?sourceforge\.net/hgweb/}
       MercurialDownloadStrategy
+    when %r{^https://(.+?\.)?sourceforge\.net}
+      CurlSourceForgeSecureMirrorDownloadStrategy
     else
       CurlDownloadStrategy
     end
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -394,9 +394,7 @@ end
 class CurlSourceForgeSecureMirrorDownloadStrategy < CurlDownloadStrategy
   def fetch
     unless ENV["HOMEBREW_NO_INSECURE_REDIRECT"].nil?
-      ohai "Sourceforge HTTPS: #{@url}"
-      @url = https_mirror
-      ohai "=> converted to #{@url}"
+      mirrors.unshift *https_mirror
     end
     super
   end


Library/Homebrew/download_strategy.rb
```diff
@@ -388,6 +388,25 @@ def _fetch
   end
 end
 
+class CurlSourceForgeSecureMirrorDownloadStrategy < CurlDownloadStrategy
+  def fetch
+    if !ENV["HOMEBREW_NO_INSECURE_REDIRECT"].nil?
```
bfontaine: Use `unless…` instead of `if !…`.

Library/Homebrew/download_strategy.rb
```diff
@@ -388,6 +388,25 @@ def _fetch
...
+      oldurl = @url
+      @url = https_mirror
+      ohai "Sourceforge HTTPS: #{oldurl}"
+      ohai "=> converted to #{@url}"
```
bfontaine: `ohai` already prefixes its message by a fat arrow (`==>`), it might be better to use a symbol other than another arrow.

```irb
>> ohai "=> foo"
==> => foo
```

Library/Homebrew/download_strategy.rb
```diff
@@ -388,6 +388,25 @@ def _fetch
...
+    end
+    super
+  end
+
+  def https_mirror
+     re = /(^https?:\/\/)(.+\.net\/project\/)(.+)/
```
bfontaine: You can use `%r{...}` instead of `/.../` to avoid having to escape all slashes (e.g. `%r{/}` instead of `/\//`).

Library/Homebrew/download_strategy.rb
smarek: Commited fix

Library/Homebrew/download_strategy.rb
```diff
@@ -388,6 +388,25 @@ def _fetch
...
+     @url.gsub(re) do
+       $1 + "www.mirrorservice.org/sites/download.sourceforge.net/pub/sourceforge/" + $3[0] + "/" + $3[0,2] + "/" + $3
```
bfontaine: Use string interpolation instead of concatenation, e.g. `"#{$1}www...#{$3[0]}..."` instead of `$1 + "www..." + $3[0] + "..."`.

Library/Homebrew/download_strategy.rb
bfontaine: Also, use `(?:...)` when you don’t want to capture a group.

Library/Homebrew/download_strategy.rb
smarek: Oh, that was not very clear in colored view
screen shot 2015-08-14 at 23 33 36

Also the message is not very clear about what has been done and why

Library/Homebrew/download_strategy.rb
smarek: Implemented and tested, Good tip, thank you

Library/Homebrew/download_strategy.rb
```diff
@@ -388,6 +388,25 @@ def _fetch
...
+# In case environment variable HOMEBREW_NO_INSECURE_REDIRECT is set, so the Sourceforge is not ultimately broken
+class CurlSourceForgeSecureMirrorDownloadStrategy < CurlDownloadStrategy
+  def fetch
+    unless ENV["HOMEBREW_NO_INSECURE_REDIRECT"].nil?
+      ohai "Sourceforge HTTPS: #{@url}"
+      @url = https_mirror
+      ohai "=> converted to #{@url}"
+    end
+    super
+  end
```
xu-cheng: How about:
```ruby
def fetch
  mirrors.unshift *https_mirrors
  super
end
```

To make it work, you need to add: 
```ruby
class CurlDownloadStrategyHTTPSRedirectError < CurlDownloadStrategyError
  def to_s
    super + "\nHTTPS to HTTP redirect detected & HOMEBREW_NO_INSECURE_REDIRECT is set."
  end
end
```
in the `exceptions.rb`. And change `raise "HTTPS to HTTP redirect detected & HOMEBREW_NO_INSECURE_REDIRECT is set."` to `raise CurlDownloadStrategyHTTPSRedirectError.new(@url)`.

smarek: That sounds as better integration. What do you think @DomT4 @bfontaine ? Thanks

Library/Homebrew/download_strategy.rb
```diff
@@ -388,6 +388,25 @@ def _fetch
...
 end
 
+# Will convert "https://downloads.sourceforge.net/project/..." URLs to mirrorservice.org
+# In case environment variable HOMEBREW_NO_INSECURE_REDIRECT is set, so the Sourceforge is not ultimately broken
+class CurlSourceForgeSecureMirrorDownloadStrategy < CurlDownloadStrategy
+  def fetch
+    unless ENV["HOMEBREW_NO_INSECURE_REDIRECT"].nil?
+      ohai "Sourceforge HTTPS: #{@url}"
+      @url = https_mirror
+      ohai "=> converted to #{@url}"
```
mikemcquaid: Just match this output format: https://github.com/smarek/homebrew/blob/sourceforge-download-strategy/Library/Homebrew/download_strategy.rb#L271-L276

smarek: I'm calling `super` so these outputs are always appended, I wanted to notify that the URL was being rewritten, however as  @xu-cheng suggested, maybe pushing the URL into mirrors would be more intuitive.

Library/Homebrew/download_strategy.rb
DomT4: If I'm understanding Xu correctly, his proposal will get in the way less when we address the broader issue later.

Library/Homebrew/download_strategy.rb
mikemcquaid: I mean I think it's enough to just print the formula URL and then the actual download URL in exactly that format.

Library/Homebrew/download_strategy.rb
```diff
@@ -388,6 +388,25 @@ def _fetch
...
+    end
+    super
+  end
+
+  def https_mirror
+     @url.gsub(%r{(^https?://)(?:.+)(.net/project/)(.+)}) do
+       "#{$1}www.mirrorservice.org/sites/download.sourceforge.net/pub/sourceforge/#{$3[0]}/#{$3[0,2]}/#{$3}"
```
vszakats: Wouldn't `dl.sourceforge.net` be a better option instead of `download.sourceforge.net`?
Content seems to be identical but the former URL is shorter.

Library/Homebrew/download_strategy.rb
xu-cheng: @smarek Since #43020 is merged. Can you update the PR by using `mirrors`? Thanks.

smarek:
Updated and tested it works as intended
```
==> Installing sox dependency: mad
==> Downloading https://downloads.sourceforge.net/project/mad/libmad/0.15.1b/libmad-0.15.1b.tar.gz
==> Downloading from http://vorboss.dl.sourceforge.net/project/mad/libmad/0.15.1b/libmad-0.15.1b.tar.gz
HTTPS to HTTP redirect detected & HOMEBREW_NO_INSECURE_REDIRECT is set.
Trying a mirror...
==> Downloading https://www.mirrorservice.org/sites/download.sourceforge.net/pub/sourceforge/m/ma/mad/libmad/0.15.1b/libmad-0.15.1b.tar.gz
*   Trying 212.219.56.184...
* Connected to www.mirrorservice.org (212.219.56.184) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
* Server certificate: www.mirrorservice.org
* Server certificate: TERENA SSL CA 2
* Server certificate: USERTrust RSA Certification Authority
* Server certificate: AddTrust External CA Root
> GET /sites/download.sourceforge.net/pub/sourceforge/m/ma/mad/libmad/0.15.1b/libmad-0.15.1b.tar.gz HTTP/1.1
> Host: www.mirrorservice.org
> User-Agent: Homebrew 0.9.5 (Ruby 2.0.0-481; OS X 10.10.5)
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Tue, 18 Aug 2015 21:13:47 GMT
< Server: Apache/2.4.7 (Ubuntu)
< Last-Modified: Wed, 18 Feb 2004 04:03:39 GMT
< ETag: "7aa6b-3d398891cfcc0"
< Accept-Ranges: bytes
< Content-Length: 502379
< Content-Type: application/x-gzip
< 
{ [16384 bytes data]
######################################################################## 100.0%* Connection #0 to host www.mirrorservice.org left intact

==> ./configure --disable-debugging --enable-fpm=64bit --prefix=/usr/local/Cellar/mad/0.15.1b
==> make CFLAGS= LDFLAGS= install
🍺  /usr/local/Cellar/mad/0.15.1b: 10 files, 356K, built in 36 seconds
```

smarek:
Also I've came across this problem, which is ie. `lame` package

```
==> Installing sox dependency: lame
==> Downloading https://downloads.sourceforge.net/sourceforge/lame/lame-3.99.5.tar.gz
==> Downloading from http://netix.dl.sourceforge.net/project/lame/lame/3.99/lame-3.99.5.tar.gz
HTTPS to HTTP redirect detected & HOMEBREW_NO_INSECURE_REDIRECT is set.
Trying a mirror...
==> Downloading https://downloads.sourceforge.net/sourceforge/lame/lame-3.99.5.tar.gz
==> Downloading from http://garr.dl.sourceforge.net/project/lame/lame/3.99/lame-3.99.5.tar.gz
HTTPS to HTTP redirect detected & HOMEBREW_NO_INSECURE_REDIRECT is set.
Error: Failed to download resource "lame"
Download failed: https://downloads.sourceforge.net/sourceforge/lame/lame-3.99.5.tar.gz
```

Original URL is `lame/lame-3.99.5.tar.gz` which is clear, however the redirect leads to another folder, specifically `lame/lame/3.99/lame-3.99.5.tar.gz`, can anyone tip me if it's possible to capture the final redirect URL, so we can rewrite it onto HTTPS mirror? Thanks

Library/Homebrew/download_strategy.rb
smarek: I observe difference between `download.sourceforge.net`, `downloads.sourceforge.net` and `dl.sourceforge.net`, I think they need to be handled separately, see the comment https://github.com/Homebrew/homebrew/pull/42955#issuecomment-132358467

Library/Homebrew/download_strategy.rb
```diff
@@ -389,6 +389,23 @@ def _fetch
   end
 end
 
+# Will convert "https://downloads.sourceforge.net/project/..." URLs to mirrorservice.org
+# In case environment variable HOMEBREW_NO_INSECURE_REDIRECT is set, so the Sourceforge is not ultimately broken
+class CurlSourceForgeSecureMirrorDownloadStrategy < CurlDownloadStrategy
+  def fetch
+    unless ENV["HOMEBREW_NO_INSECURE_REDIRECT"].nil?
```
xu-cheng: I don't think this is necessary. You can just change mirrors unconditionally

smarek: I would rather not modify processing of all sourceforge formulas always, adding mirror is meaningful only in this specific case, unless we introduce something like "HOMEBREW_DISABLE_INSECURE_DOWNLOADS" which would require all formulas to be provided through SSL enabled address.

xu-cheng: You only append mirrors, so it won't alter the normal downloading processing. In the mean time, it will help to mitigate the downloading problems, say sourceforge is down.

smarek: Well is this change the only thing holding this PR from being merged? Also one more question, should I squash commits using rebase, so I don't spam the Homebrew git tree a lot. Thanks

xu-cheng: Have you fixed the problem in https://github.com/Homebrew/homebrew/pull/42955#issuecomment-132358467 ?

Yes it would be great for you to rebase the commits.

tdsmith:
Waiting for a reply to @xu-cheng's last comment.

apjanke:

UniqMartin:
> and re-open on the new [`homebrew/core repo`](https://github.com/Homebrew/homebrew-core).

Minor correction: The [Homebrew/brew](https://github.com/Homebrew/brew) repository.

PR #42959, opened 2015-08-14
kibana 4.1.1

auskast:
upgraded from 4.1.0


xu-cheng:

PR #42969, opened 2015-08-15
asio 1.10.6, devel 1.11.0 (new formula)

guoxiao:


dunn:
Looks like the test needs boost.  Since boost is an optional dependency we might need a different test:
```
        Testing asio
==> /usr/bin/clang++ /usr/local/Cellar/asio/1.10.6/share/asio/examples/cpp03/timers/time_t_timer.cpp -o timer
In file included from /usr/local/Cellar/asio/1.10.6/share/asio/examples/cpp03/timers/time_t_timer.cpp:11:
In file included from /usr/local/include/asio.hpp:18:
In file included from /usr/local/include/asio/async_result.hpp:18:
/usr/local/include/asio/detail/config.hpp:26:11: fatal error: 'boost/config.hpp' file not found
# include <boost/config.hpp>
          ^
1 error generated.
```

mikemcquaid:
> Since boost is an optional dependency we might need a different test:

:+1:

dunn:
Failing to build on Mountain Lion as well: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/31598/version=mountain_lion/testReport/junit/brew-test-bot/mountain_lion/install_asio/

guoxiao:
Updated.

I found that `depends_on "boost" => :test` doesn't work, and I found that `option :cxx11` doesn't have a option to enable it by default. 

Maybe some improvements can be done here.

mikemcquaid:
> I found that depends_on "boost" => :test doesn't work, and I found that option :cxx11 doesn't have a option to enable it by default.

Yep, both of those are intentional.

dunn:
New failure:
```
checking boost/noncopyable.hpp usability... no
checking boost/noncopyable.hpp presence... no
checking for boost/noncopyable.hpp... no
Can't find boost headers. Please check the location of the boost
distribution and rerun configure using the --with-boost=DIR option.
```

Library/Formula/asio.rb
```diff
@@ -0,0 +1,61 @@
...
+  url "https://downloads.sourceforge.net/project/asio/asio/1.10.6%20%28Stable%29/asio-1.10.6.tar.bz2"
+  sha256 "e0d71c40a7b1f6c1334008fb279e7361b32a063e020efd21e40d9d8ff037195e"
+  head "https://github.com/chriskohlhoff/asio.git"
+
+  devel do
+    url "https://downloads.sourceforge.net/project/asio/asio/1.11.0%20%28Development%29/asio-1.11.0.tar.bz2"
+    sha256 "4f7e13260eea67412202638ec111cb5014f44bdebe96103279c60236874daa50"
+  end
+
+  option "with-boost-coroutine", "use Boost.Coroutine to implement stackful coroutines"
```
DomT4: Start the option explanation with a capital letter please, `"Use Boost.Coroutine to implement stackful coroutines"` in this case.

Library/Formula/asio.rb
```diff
@@ -0,0 +1,61 @@
...
+  option :cxx11
+  needs :cxx11 if build.cxx11?
```
DomT4: Empty line between the `option` and `needs` please.

Library/Formula/asio.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "boost" => :optional
+  depends_on "boost" unless build.cxx11?
+  depends_on "boost" if build.with? "boost-coroutine"
```
DomT4: Can we try to combine these three different `boost` dependencies somehow?

Library/Formula/asio.rb
```diff
@@ -0,0 +1,61 @@
...
+  depends_on "openssl"
+
+  def install
+    if build.head?
+      cd "asio"
+      system "./autogen.sh"
+    else
+      system "autoconf"
+    end
+    ENV.cxx11 if build.cxx11?
```
DomT4: `ENV.cxx11 if build.cxx11?` should be the first thing in `def install`.

Library/Formula/asio.rb
```diff
@@ -0,0 +1,61 @@
...
+    args = %W[
+      --disable-dependency-tracking
+      --disable-silent-rules
+      --prefix=#{prefix}
+      --with-boost=#{(build.with?("boost") || build.with?("boost-coroutine") || !build.cxx11?) ? Formula["boost"].opt_include : "no"}
+    ]
+    args << "--enable-boost-coroutine" if build.with? "boost-coroutine"
+    system "./configure", *args
```
DomT4: Add an empty line between last `args` and the `system "./configure", *args`.

Library/Formula/asio.rb
```diff
@@ -0,0 +1,61 @@
...
+    system "make", "install"
+    pkgshare.install "src/examples"
+  end
+
+  test do
+    if build.cxx11?
```
DomT4: We don't test non-default options in `test do`, so just test whatever the default build is.

Library/Formula/asio.rb
```diff
@@ -0,0 +1,61 @@
...
+      httpserver = pkgshare/"examples/cpp11/http/server/http_server"
+    else
+      httpserver = pkgshare/"examples/cpp03/http/server/http_server"
+    end
+    pid = fork do
+      exec httpserver, "127.0.0.1", "8080", "."
+    end
+    sleep 1
+    begin
+      assert_match /404 Not Found/, shell_output("curl http://127.0.0.1:8080 2>/dev/null")
```
DomT4: Why does this need redirecting to `/dev/null`, out of curiosity?

guoxiao: It's just a personal habit...

Library/Formula/asio.rb
guoxiao: How could I combine `:optional` with `if/unless`? I didn't find a good approach.

Library/Formula/asio.rb
guoxiao: If we remove the test for cxx11, the test will fail if users install it in cxx11 mode and try to run the test.
The test block is not only for TestBot but also for users IMHO.

Library/Formula/asio.rb
guoxiao: Merged the `if/unless`.

Library/Formula/asio.rb
guoxiao: Having removed that.

Library/Formula/asio.rb
DomT4: We only support testing default options. We're not prepared to open the vast floodgates to testing non-default options. The primary purpose of `test do` is to serve the test-bot and ensure the formula functions as expected both now & long-term.

guoxiao: Updated.

Library/Formula/asio.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  option "with-boost-coroutine", "Use Boost.Coroutine to implement stackful coroutines"
+  option :cxx11
+
+  needs :cxx11 if build.cxx11?
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "boost" => :optional
+  depends_on "boost" if !build.cxx11? || build.with?("boost-coroutine")
```
mikemcquaid: This should probably be combined with the `:optional` above?

guoxiao: How should I do to combine them?

mikemcquaid: Something like:
```ruby
boost_args = []
boost_args << :optional if build.cxx11? && build.with?("boost-coroutine")
depends_on "boost" => boost_args
```

guoxiao: I changed these lines to multiline if-else to make it more readable.

guoxiao:
Ping.

guoxiao:
Could this PR be merged now?

dunn:
@BrewTestBot test this please

dunn:
Thanks so much, and sorry for the delay! Merged in f5f942cd8aa155dc0c7d17f240188d45e28e73db.

Issue #42967, opened 2015-08-15
`brew cleanup --force -s` deletes all caches


cc941201:
I ran `brew cleanup --force -s` today, and it deleted all cached downloads from my `/Library/Caches/Homebrew`. It is an undocumented change of behavior. Previously, only caches of deleted packages would be deleted.

I looked through commits from last few days, and the only commit related to cleanup is this one:
https://github.com/Homebrew/homebrew/commit/066bdb0d10533ed86e5d3580fe6ab4669d238a48

xu-cheng:
066bdb0 has nothing to do with it. Related commit is 17eee232838d4639b25f863aa342b1dda61b81bc

cc941201:
Well, that makes more sense.

mikemcquaid:
`--force` deleted all cached downloads before this change. That's what the flag does.

cc941201:
But according to the manual, `--force` deletes keg-only packages which is consistent with my experience before. This is the only time it deletes all caches.

UniqMartin:
I have to second @cc941201's observation. The manual says otherwise and my experience is also that before the change, `--force` would *only* remove outdated keg-only packages while `--prune=<days>` would *only* affect removal of cached downloads.

UniqMartin:
I hope I'm not stepping on anyone's toes, but I created PR #42970 in an attempt to make everyone happy.

mikemcquaid:
> But according to the manual, --force deletes keg-only packages which is consistent with my experience before. This is the only time it deletes all caches.

According to the manual: I'm an idiot and apologise. Sorry for my misunderstanding here; #42970 is looking much better.

UniqMartin:
The issue was addressed in 9ac7f8c7620d2ca3b9090530d46fbfe5badad6ec and slightly tweaked in subsequent commits. `brew update` and feel free to close this issue, if it works again as expected.

PR #42970, opened 2015-08-15
cleanup: make --force less aggressive

UniqMartin:
Change behavior for `brew cleanup` as follows:

- If `--force` is supplied, remove only outdated keg-only packages.
- If `--prune=<days>` is supplied, remove both logs and cached downloads older than the specified number of days. Use `--prune=all` to remove all logs and cached downloads irrespective of age.
- By default, remove logs after 14 days and cached downloads never.

Also centralizes handling of `--prune`, thus removing duplicate logic.

This is motivated by commit 17eee232838d4639b25f863aa342b1dda61b81bc that made `--force` much more aggressive and made it override whatever was specified via `--prune`, completely removing all:

- outdated keg-only packages
- cached downloads irrespective of age
- logs irrespective of age

This made it impossible to remove outdated keg-only packages without also deleting all cached downloads, which is at least inconvenient for people with limited bandwidth wanting to rebuild packages later.

**Additional notes:**

- This addresses issue #42967.
- The manual should probably be updated to reflect this change.
- Now that `--prune` affects both logs and cached downloads, it might be beneficial to have both `--prune[-cache]` and `--prune-log` (not implemented here).

### DIFF
--- a/Library/Homebrew/cmd/cleanup.rb
+++ b/Library/Homebrew/cmd/cleanup.rb
@@ -19,14 +19,8 @@ module Homebrew
 
   def cleanup_logs
     return unless HOMEBREW_LOGS.directory?
-    prune = ARGV.value "prune"
-    if prune
-      time = Time.now - 60 * 60 * 24 * prune.to_i
-    else
-      time = Time.now - 60 * 60 * 24 * 7 * 2 # two weeks
-    end
     HOMEBREW_LOGS.subdirs.each do |dir|
-      cleanup_path(dir) { dir.rmtree } if ARGV.force? || (dir.mtime < time)
+      cleanup_path(dir) { dir.rmtree } if prune?(:logs, dir.mtime)
     end
   end
 
@@ -61,10 +55,8 @@ module Homebrew
 
   def cleanup_cache
     return unless HOMEBREW_CACHE.directory?
-    prune = ARGV.value "prune"
-    time = Time.now - 60 * 60 * 24 * prune.to_i
     HOMEBREW_CACHE.children.each do |path|
-      if ARGV.force? || (prune && path.mtime < time)
+      if prune?(:cache, path.mtime)
         if path.file?
           cleanup_path(path) { path.unlink }
         elsif path.directory? && path.to_s.include?("--")
@@ -127,6 +119,29 @@ module Homebrew
             map { |p| HOMEBREW_PREFIX/p }.select(&:exist?)
     args = paths.map(&:to_s) + %w[-name .DS_Store -delete]
     quiet_system "find", *args
+  end
+
+  def prune?(kind, time)
+    unless @cleanup_prune_limit
+      # Infer and cache prune limits for cleanup: Use '--prune' or fall back to
+      # defaults (never prune download cache and prune logs after two weeks).
+      days = ARGV.value "prune"
+      offset = Time.now
+      @cleanup_prune_limit = {}
+      { :cache => nil, :logs => 14 }.each do |k, days_default|
+        @cleanup_prune_limit[k] = if days == "all"
+          "all"
+        elsif days
+          offset - 60 * 60 * 24 * days.to_i
+        elsif days_default
+          offset - 60 * 60 * 24 * days_default
+        end
+      end
+    end
+
+    if limit = @cleanup_prune_limit[kind]
+      limit == "all" || time < limit
+    end
   end
 
   def eligible_for_cleanup?(formula)


DomT4:
> Use --prune=all to remove all logs and cached downloads irrespective of age.

To keep in line with the using number of days, I'd be happier with `--prune=0` personally.

UniqMartin:
> To keep in line with the using number of days, I'd be happier with `--prune=0` personally.

Thanks for the feedback! My inspiration came from `git gc`, which also uses `--prune=all` and which seems more user-friendly to me. But `--prune=0` would of course continue to work, though with the subtle (and rather academic) difference of retaining files from the future. Are you suggesting to drop `--prune=all` entirely or to make `--prune=0` behave like `--prune=all` in my current patch?

Library/Homebrew/cmd/cleanup.rb
```diff
@@ -19,14 +19,9 @@ def cleanup
 
   def cleanup_logs
     return unless HOMEBREW_LOGS.directory?
-    prune = ARGV.value "prune"
-    if prune
-      time = Time.now - 60 * 60 * 24 * prune.to_i
-    else
-      time = Time.now - 60 * 60 * 24 * 7 * 2 # two weeks
-    end
+    prune = prune_flag 14 # two weeks unless overridden by argument
```
mikemcquaid: `prune = prune_flag(14)`

DomT4:
Apologies - For some reason my mind informed me `git gc` used `--prune=0`. If that isn't the case, ignore me :wink:. The inspiration for our wording on this came from `git` as well, so we might as well be consistent in the syntax.

Library/Homebrew/cmd/cleanup.rb
```diff
@@ -19,14 +19,9 @@ def cleanup
...
     HOMEBREW_LOGS.subdirs.each do |dir|
-      cleanup_path(dir) { dir.rmtree } if ARGV.force? || (dir.mtime < time)
+      cleanup_path(dir) { dir.rmtree } if prune?(dir.mtime, prune)
```
mikemcquaid: `if cleanup?(dir.mtime, prune_time)`

Library/Homebrew/cmd/cleanup.rb
```diff
@@ -61,10 +56,9 @@ def cleanup_keg(keg)
 
   def cleanup_cache
     return unless HOMEBREW_CACHE.directory?
-    prune = ARGV.value "prune"
-    time = Time.now - 60 * 60 * 24 * prune.to_i
+    prune = prune_flag # never unless overridden by argument
```
mikemcquaid: `prune_time = prune_time_from_argv` and can skip the comment.

Library/Homebrew/cmd/cleanup.rb
```diff
@@ -61,10 +56,9 @@ def cleanup_keg(keg)
...
     HOMEBREW_CACHE.children.each do |path|
-      if ARGV.force? || (prune && path.mtime < time)
+      if prune?(path.mtime, prune)
```
mikemcquaid: `if cleanup?(dir.mtime, prune_time)`

Library/Homebrew/cmd/cleanup.rb
mikemcquaid: Actually, no, just handle the `prune_time` in the `cleanup?` method.

Library/Homebrew/cmd/cleanup.rb
```diff
@@ -129,6 +123,24 @@ def rm_DS_Store
...
+      "all"
+    elsif prune
+      Time.now - 60 * 60 * 24 * prune.to_i
+    elsif default
+      Time.now - 60 * 60 * 24 * default
+    end
+  end
+
+  def prune?(time, flag)
+    flag && ((flag == "all") || (time < flag))
```
mikemcquaid: Use an `if` statement and have this function take a single argument, get the value from `ARGV` unless it's set and call it `cleanup`

Library/Homebrew/cmd/cleanup.rb
UniqMartin: What is the difference between `prune = ARGV.value "prune"` (old code, still there in my patch in a different place) and `prune = prune_flag 14` (new code)? Sorry if I'm being blind, but what is the underlying rule and is it documented anywhere? I also kept the style like that because `brew style` didn't complain.

UniqMartin: The reason for a separate method is that it is referenced in two different places and I wanted to avoid duplicating the logic. I like the suggested name change, though; will push it in a bit.

mikemcquaid: Both methods make use of `prune?` though, no?

mikemcquaid: So it should be possible to handle the logic inside the `prune?` method I believe?

UniqMartin:
Many thanks for the detailed feedback, @mikemcquaid! I just pushed the updated code. Here are some thoughts on why I haven't acted (yet) on some of the suggestions:

- As `prune?` quite nicely matches with the `--prune` argument it is ultimately working with, I'm somewhat hesitant to rename it to `cleanup?`.
- I think having `prune_time_from_argv` as a separate method instead of folding it into `prune?` has some value:
  - `Time.now` is only called once, providing a fixed time reference for all following deletions.
  - Avoids repeated processing of `ARGV` while iterating over a potentially long list of files/directories.

I would very much appreciate comments on these two points.

mikemcquaid:
> As prune? quite nicely matches with the --prune argument it is ultimately working with, I'm somewhat hesitant to rename it to cleanup?.

That's cool, I'm easy either way.

> I think having prune_time_from_argv as a separate method instead of folding it into prune? has some value:
> Time.now is only called once, providing a fixed time reference for all following deletions.
> Avoids repeated processing of ARGV while iterating over a potentially long list of files/directories.
> I would very much appreciate comments on these two points.

Still think even if it's not rolled into one method that it should be called by `prune?`. Perhaps a better solution here is to cache the value with a class variable so it's only ever calculated once (even with multiple calls)?

UniqMartin:
> Perhaps a better solution here is to cache the value with a class variable so it's only ever calculated once (even with multiple calls)?

Nice solution! Done. I moved the entire logic into `prune?` and actually am liking the result.

mikemcquaid:
@UniqMartin Thanks for all your work here! I merged it and added my own tweaks in c1084a3 (just to be able to get it in quickly).

UniqMartin:
Thanks for the merge! Your tweaks turned out to be quite extensive. ;) I now have a much better understanding of what you were expecting.

Unfortunately, the logic in the very last line is now exactly backwards. (I'm not sure if you get notified about my comment on your commit, so I'm mentioning it here as well.)

mikemcquaid:
@UniqMartin Ugh, fixed in a3af9ee. This is not my week :sob:. Thanks again for your work here!

PR #42973, opened 2015-08-15
kitchen-sync: use brewed libyaml-cpp

tdsmith:
Otherwise kitchen-sync vendors and installs libyaml-cpp, which causes
conflicts.


mikemcquaid:
:+1:

PR #42976, opened 2015-08-15
let tests pass on Xcode-only systems

tdsmith:
The test failed with this output on an Xcode-only system because the test didn't expect the possibility that CMAKE_PREFIX_PATH could contain additional prefixes.

http://pastebin.com/raw.php?i=QdCma4NX


mikemcquaid:
:+1:

Issue #21412, opened 2013-07-23
GTK+ searches for modules in the wrong path


philippelatulippe:
I installed the package gtk-engines so that GTK applications installed with homebrew don't look ugly, but GTK searches for modules in its Cellar path.  Of course, the engines from the gtk-engines package are not present at that path.

The simplest fix is to set libdir to #{HOMEBREW_PREFIX}/lib when calling configure, but I'm not sure if that's acceptable.

See #11445, which has more details and a link to the code that determines the search path.

ipeacocks:
@DomT4 

Tried to export variable in such way:

export GTK_PATH=/usr/local/lib/gtk-2.0

And everything is cool now. Thank you!

mikemcquaid:
I'm closing this out as people seem to have various solutions and Homebrew itself probably doesn't want to change the current behaviour.

PR #42979, opened 2015-08-15
thewidgetfactory 0.2.1

afb:
Classic showcase for gtk+ themes (only works with gtk2 not gtk3 but that is OK), try with "gtk-engines" (Clearlooks) and "gnome-themes-standard" (Adwaita) which both look better than the default (Raleigh)

### DIFF
--- a/Library/Formula/thewidgetfactory.rb
+++ b/Library/Formula/thewidgetfactory.rb
@@ -1,0 +1,68 @@
+class Thewidgetfactory < Formula
+  desc "The Widget Factory is a showcase tool for GTK+ widgets"
+  homepage "http://www.stellingwerff.com/?page_id=10"
+  url "http://www.stellingwerff.com/TheWidgetFactory/thewidgetfactory-0.2.1.tar.gz"
+  sha256 "10bbbe50dc23a0cd47762749c6bb402559721951f3c6af40366258593bcb1ce4"
+
+  depends_on "pkg-config" => :build
+  depends_on "gtk+"
+
+  # Avoid warning due to the form being created with an obsolete version of Glade:
+  # "GtkSpinButton: setting an adjustment with non-zero page size is deprecated"
+  patch :DATA
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}",
+                          "--datadir=#{HOMEBREW_PREFIX}/share"
+    system "make", "install"
+  end
+
+  test do
+    # make sure to find all themes, not only those in gtk+'s Cellar
+    ENV["GTK_EXE_PREFIX"] = HOMEBREW_PREFIX # gtk_rc_get_module_dir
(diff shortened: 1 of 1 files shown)


Library/Formula/thewidgetfactory.rb
```diff
@@ -0,0 +1,66 @@
+class Thewidgetfactory < Formula
+  desc "The Widget Factory is a showcase tool for GTK+ widgets"
+  homepage "http://www.stellingwerff.com/?page_id=10"
+  url "http://www.stellingwerff.com/TheWidgetFactory/thewidgetfactory-0.2.1.tar.gz"
+  sha256 "10bbbe50dc23a0cd47762749c6bb402559721951f3c6af40366258593bcb1ce4"
+
+  depends_on "pkg-config" => :build
+  depends_on "gtk+"
+
+  patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

afb: There is no (active) upstream, patch only avoids a runtime warning. Documented it a bit more.
Upstream has moved towards gtk3, so need to change from "twf" to "awf" if that is needed...

afb:
There's also a workaround in "test", for the broken (Cellar-only) modules/themes of gtk+ (#21412)

Library/Formula/thewidgetfactory.rb
DomT4: If there's no active upstream to take the patch it's unlikely we'll accept this into Homebrew. We're pretty :-1: on carrying patches in new formula, let alone if we'll need to carry them for a very prolonged period.

afb: OK, so you would rather have it ugly (gtk+ and twf) than doing any patching downstream ?
I suppose that "works", but it does look a little grey and it does output some deprecations.

afb:
Checks failed, due to them running a GUI program and timing out. Maybe that was a bad test...

afb:
I suppose this isn't an acceptable formula then, considering it's an old abandoned gui program... ;-)
Anyway, here was the outcome: http://imgur.com/a/3X2pD. Hope that you get gtk+ fixed, eventually.

Library/Formula/thewidgetfactory.rb
DomT4: We're happy to do patching downstream. We're less happy to carry patches forever that upstream aren't aware of and may never accept, essentially because we end up becoming upstream at that point in keeping the software functioning as expected when upstream have stopped doing so.

It isn't really uncommon for GTK projects to throw deprecation notices, pretty much everything I use that has a GTK GUI does so.

afb: Fair enough. And I suppose 9 years is a while (prolonged period) :-) Even for such a simple patch.
And what you say about GTK+ is very true, especially after GNOME took over all (i.e. gtk+3)

PR #42981, opened 2015-08-15
symlinks 1.4.3

atdt:


Library/Formula/symlinks.rb
```diff
@@ -0,0 +1,20 @@
+class Symlinks < Formula
+  desc "symbolic link maintenance utility"
+  homepage "https://github.com/brandt/symlinks"
+  url "https://github.com/brandt/symlinks/archive/v1.4.3.tar.gz"
+  head "https://github.com/brandt/symlinks.git"
+  version "1.4.3"
```
bfontaine: You can remove this line.

dunn:
```
==> audit problems
symlinks:
 * `checksum` (line 6) should be put before `head` (line 5)
```
It also doesn't look like it's being actively developed.  Unless another maintainer objects I think we'll pass on this.

mikemcquaid:
Passing on this, sorry!

PR #42982, opened 2015-08-15
qt5: restore rpath usage

DomT4:
Doing some more testing on this. There are various upstream patches available for `macdeployqt`, but none for the `pkgconfig` error yet. None applied here, want to play with the `rpath` a bit more. Will do a local build as well; want to check the bot & I produce roughly the same bottles in terms of content.


DomT4:
Giving up on this particular PR for now.

PR #42983, opened 2015-08-15
swiftgen 0.4.1 (new formula)

AliSoftware:


Library/Formula/swiftgen.rb
```diff
@@ -0,0 +1,19 @@
+class Swiftgen < Formula
+  desc "A collection of Swift tools to generate Swift code (enums for your assets, storyboards, Localizable.strings, …)"
```
DomT4: This is going to be *way* too long. The limit is 80 characters, including the name. Will be highlighted by `brew audit --strict`

Library/Formula/swiftgen.rb
```diff
@@ -0,0 +1,19 @@
...
+  homepage "https://github.com/AliSoftware/SwiftGen"
+
```
DomT4: Don't need the empty line here.

Library/Formula/swiftgen.rb
```diff
@@ -0,0 +1,19 @@
...
+  url "https://github.com/AliSoftware/SwiftGen/archive/0.4.1.tar.gz"
+  sha256 "e572631bdac5ebff49314f56cd4bd25a0eae8efc2390602e30d6cac1f196f53b"
+  head "https://github.com/AliSoftware/SwiftGen.git", :branch => "master"
+
+  def install
+    system "rake", "all[#{prefix},/bin]"
```
DomT4: I think just `rake "all[#{prefix},/bin]"` would work, without calling `system`.

Library/Formula/swiftgen.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  test do
+    system "#{bin}/swiftgen-assets", "--version"
```
DomT4: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

AliSoftware:
Thanks for the comments, this is my first formula so this is new to me ;-)

May I append commits to this PR (to fix what you point out in your comments), or should I amend and force-push instead (since Homebrew seem to have a "one formula, one commit" rule)?

DomT4:
General standard is to fix the issues & then squash back down to one commit & force push, aye.

AliSoftware:
I'm having trouble with writing the `test` block:

* I already have a lot of Unit Tests written in my Xcode project that builds the binaries, so my binaries are tested
* But those tests require big fixtures to work (fixtures that I can't re-write on the fly in my formula's test block like libxml2 does for example)
* And I can't run `rake tests` to make Homebrew run my Unit Tests either, because even if the fixtures are present in my tarsal, once the binaries are installed in the Cellar, the fixtures are not available anymore

Any suggestion to help me on this one?

DomT4:
> even if the fixtures are present in my tarsal, once the binaries are installed in the Cellar, the fixtures are not available anymore

This one is easily solved at least. Just `pkgshare.install "Tests"` or similar.

What about something like [this](https://github.com/AliSoftware/SwiftGen#uicolor) for the test?

You can write a short `colors.txt` file to the `testpath` and then use `swiftgen-color` to generate the appropriate Swift code which is presumably outputted to a file?

If so, you can use `assert File.exist?("")` to ensure that file has been created.

Library/Formula/swiftgen.rb
```diff
@@ -0,0 +1,15 @@
+class Swiftgen < Formula
+  desc "Collection of Swift tools to generate Swift code"
+  homepage "https://github.com/AliSoftware/SwiftGen"
+  url "https://github.com/AliSoftware/SwiftGen/archive/0.4.1.tar.gz"
+  sha256 "a38b4bbd27eb183e73d2c1d5245f3134b8d688bc3095b538e52ed1ca0b85180b"
+  head "https://github.com/AliSoftware/SwiftGen.git", :branch => "master"
```
DomT4: You don't need to specify `master` on this. We assume `master` is the default branch and grab that unless told otherwise.

Library/Formula/swiftgen.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  def install
+    rake "all[#{prefix},/bin]"
```
DomT4: If you're installing here, you can call `#{bin}` directly rather than calling `#{prefix}` first, i.e.

```ruby
rake "all[#{bin}]
```

AliSoftware: Well given how my rake task is created, which use both xcodebuild's `DSTROOT` and `INSTALL_ROOT`, it is kinda easier for me to have the path split in two parts like I did — my rake task doesn't support having only one argument so we would still need to `rake "all[#{bin},/]" in that case anyway, so…

DomT4: Can you not do something like `rake "install", "INSTALL_PATH=#{bin}"` or similar? Is that syntax unsupported?

AliSoftware: Well that's not how my `Rakefile` works no :-/

AliSoftware:
@DomT4 finally got the tests working thanks to your pointers with `pkgshare.install` :+1: 

`brew test swiftgen` works now and `brew audit --strict` seems ok.

AliSoftware:
Ok, everything is good now, with proper tests (that would properly fail on error, not like before!), usage of `rake "install[${bin},/]"` and versioning system of my binaries fixed :smiley:

DomT4:
This is incompatible with anything older than the Xcode-Beta 7.0?

```
== swiftgen-l10n ==
rake aborted!

[!!!] You need to have Xcode 7.x installed to compile the SwiftGen tools
```

AliSoftware:
Yes it is (it's explained in the README), because it uses (and generates) Swift 2.0 code.

mikemcquaid:
> Yes it is (it's explained in the README), because it uses (and generates) Swift 2.0 code.

We can't include this until Xcode 7 is a stable version.

esttorhe:
Since `Xcode 7` is `GM` it is stable now.

Would it be ok to merge this now?

mikemcquaid:
@esttorhe It's not about it being GM but it is released on Yosemite so we can rebuild this and include it.

mikemcquaid:
@BrewTestBot test this please

DomT4:
You need a `depends_on :xcode => "7.0"` line here presumably.

There's an audit nit as well:
```
== /usr/local/Library/Formula/swiftgen.rb ==
C: 22:  7: Put a comma after the last parameter of a multiline method call.

```

And a test issue:
```
Error: swiftgen: failed
No such file or directory - /usr/local/Cellar/swiftgen/0.4.1/bin/swiftgen-assets
```

AliSoftware:
@DomT4 I've fixed it by adding `depends_on :code => "7.0"` and adding the trailing comma. I've run `brew audit swiftgen` and `brew test swiftgen` and everything seems OK locally.

As SwiftGen is now at version 0.4.4, i've created a different PR #44186 which overrules this one.

dunn:
You're welcome to update the existing PR and force-push the new version, too.  We don't mind!

AliSoftware:
@dunn Ok, too late, but good to know for next time ;)

PR #42985, opened 2015-08-16
deisctl 1.9.0

felixbuenemann:
This updates `deisctl` to the latest upstream version and adds `--HEAD` support.

I also added a `post_install` step that downloads some mandatory data into `~/.deis/units`.

### DIFF
--- a/Library/Formula/deisctl.rb
+++ b/Library/Formula/deisctl.rb
@@ -3,8 +3,8 @@ require "language/go"
 class Deisctl < Formula
   desc "Deis Control Utility"
   homepage "http://deis.io/"
-  url "https://github.com/deis/deis/archive/v1.1.0.tar.gz"
-  sha256 "7f4fc7ddddf3157d59ed9e726a5be5510a4b208a936227e65f9041ceb3627a14"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> I also added a post_install step that downloads some mandatory data into ~/.deis/units.

We shouldn't be interacting with `$HOME` even in `post_install`, I'm afraid.

mikemcquaid:
`fetch` failing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/31630/version=mavericks/testReport/junit/brew-test-bot/mavericks/fetch_deisctl/

felixbuenemann:
Dunno why the wrong tag wasn't causing a failure locally. I've removed it along with the post_install and head, because whe can't ensure proper godeps for head.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/5fcb9ea6e9110e757c45df027b227a44fcc6dc1f. Thank you @felixbuenemann for your contribution! :smiley_cat:

felixbuenemann:
Thanks, Dominyk!

PR #42993, opened 2015-08-16
Update go.rb

syhily:
The repository url on google is inaccessible to chinese users. So please use origin Github url

### DIFF
--- a/Library/Formula/go.rb
+++ b/Library/Formula/go.rb
@@ -31,7 +31,7 @@ class Go < Formula
   deprecated_option "cross-compile-common" => "with-cc-common"
 
   resource "gotools" do
-    url "https://github.com/golang/tools",
+    url "https://github.com/golang/tools.git",
     :revision => "9f2124fb70150373b1fa070cb5547a7d8d6f2930"
   end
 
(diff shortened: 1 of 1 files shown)


Library/Formula/go.rb
```diff
@@ -31,8 +31,8 @@ class Go < Formula
   deprecated_option "cross-compile-common" => "with-cc-common"
 
   resource "gotools" do
-    url "https://go.googlesource.com/tools.git",
-    :revision => "69db398fe0e69396984e3967724820c1f631e971"
+    url "https://github.com/golang/tools",
```
bfontaine: Please add `.git` at the end of the URL.

syhily: @bfontaine OK, finished it. XD

DomT4:
I seem to recall that we've actually rejected this change in the past because it differs from the official upstream links, but maybe we're fine with doing so now.

mikemcquaid:
The revision shouldn't need to change. Can we add it as a `mirror` instead?

DomT4:
We'll consider a mirror on a rebased formula, if that functionality works in this situation, Thanks.

PR #42994, opened 2015-08-16
awf 1.3.0

afb:
Instead of #42979

### DIFF
--- a/Library/Formula/awf.rb
+++ b/Library/Formula/awf.rb
@@ -1,0 +1,30 @@
+class Awf < Formula
+  desc "awf (A Widget Factory) is a theme preview application for gtk2 and gtk3"
+  homepage "https://github.com/valr/awf"
+  url "https://github.com/valr/awf/archive/v1.3.0.tar.gz"
+  sha256 "911ded8a307beecbb8ee1dab489fa7b1aa9d7965cb34bd938482220505814fa4"
+  head "https://github.com/valr/awf.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "pkg-config" => :build
+  depends_on "gtk+"
+  depends_on "gtk+3"
+
+  def install
+    inreplace "src/awf.c", "/usr/share/themes", "#{HOMEBREW_PREFIX}/share/themes"
+    system "./autogen.sh"
+    rm "README.md" # let's not have two copies of README
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    assert (bin/"awf-gtk2").exist?
+    assert (bin/"awf-gtk3").exist?
+  end
+end


afb:
Screenshots: http://imgur.com/a/7frX7

Need to set $GTK_EXE_PREFIX/$GTK_DATA_PREFIX for gtk+ and install "gnome-icon-theme" for gtk+3
But those issues should be fixed in `gtk+` and `gtk+3`, and not in every other application using them...

Library/Formula/awf.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "pkg-config" => :build
+  depends_on "gtk+"
+  depends_on "gtk+3"
+
+  def install
+    inreplace "src/awf.c", "/usr/share", "#{HOMEBREW_PREFIX}/share"
```
bfontaine: Use `share` instead of `"#{HOMEBREW_PREFIX}/share"`.

Library/Formula/awf.rb
```diff
@@ -0,0 +1,30 @@
...
+    system "./autogen.sh"
+    rm "README.md" # lets not have two copies of README
```
bfontaine: let’s

Library/Formula/awf.rb
```diff
@@ -0,0 +1,30 @@
...
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    assert (bin/"awf-gtk2").exist?
+    assert (bin/"awf-gtk3").exist?
```
bfontaine: Please add a test that actually tests the program. Testing that files exist is not sufficient.

Library/Formula/awf.rb
afb: Doesn't that end up in the Cellar, where it is dark and damp and no other live ?

Library/Formula/awf.rb
afb: serves me right for copying that text

Library/Formula/awf.rb
afb: I tried to test the `twf` program before, but the gui only ended up and timeout/failed the check.

Do you mean that it should save a screenshot to some file - or something advanced like that ?

bfontaine: We can’t test a GUI, is there some simple flag that doesn’t start it like `--version` or `--help`?

afb: Sadly, no... Unless there is some magic gtk+ default arguments, it does call `gtk_init` from main

Library/Formula/awf.rb
DomT4: Everything is symlinked back to the common higher prefix, in this case usually `/usr/local/share`.

Library/Formula/awf.rb
xu-cheng: If you cannot find a better test, please remove the test. Asserting file existence is meaningless.

mikemcquaid: @xu-cheng I disagree, I'd rather have a rubbish test than no test at all so `brew audit --strict` isnt' failing consistently.

xu-cheng: Sorry, but Jack and I share the opinion on this matter. I think such test is equal to `test do true end`. Not only it fails to test anything, it hides the problems of lacking test. Which makes hard to verify which formulae require proper tests.

Library/Formula/awf.rb
afb: I don't see how that would work. Are you saying that /usr/local/share gets a matching symlink in every keg ? For me, I only get a link *from* the higher brew prefix *to* the corresponding keg.
Which means that if I want to find the global themes, I need to look in the global directory rather than the datadir in my own little installation where no-one else will install any themes...

Incidentally, this is the same bug as in gtk+ (#21412) and why it only finds "Raleigh" and friends.
Hardcoding to "/usr/share/theme" is lame anyway, better would be to use `gtk_rc_get_theme_dir`. Except in homebrew, where that is broken and needs environment variables to be set in order to work. But normally, one wants to avoid hardcoding paths to prefixes (relocatable software).

Library/Formula/awf.rb
afb: But the bot requires it (``* A `test do` test block should be added``), and everyone else is doing it ?
Or perhaps you mean I should keep the block, but leave it empty ? That I can do, no problem...

mikemcquaid: It's clearly not equal to that; if the buildsystem changes and fails to install these executables any more then this will show that. We have a bunch of other similar tests and there's some formula where it's impossible to do better (because they just install unexecutable data files)

afb: @mikemcquaid: so re-add the existance-test again then ? (I did copy it from "similar tests")

I personally don't think that `--version` checks much at all, except the linking of gtk+ perhaps,
but it is not possible to run the executables without "something" closing the gui down again.

DomT4: I'm with Mike on this. Some things just can't practically be tested. Obviously if there's an executable or library we should always enforce testing that, but where neither of those things exist our hands are fairly tied.

If we leave out an existence test here, where do we go in terms of testing in 3/6/9 months time? Unless we have plans & the capability to do automatic GUI testing no better test is going to come up & we should accept these sort of things.

afb: Adding it back then.

mikemcquaid:

PR #42995, opened 2015-08-16
Added option to install portaudio with C++ bindings.

ColinKinloch:


mikemcquaid:
Maybe just build and install these by default and not have an option?

ColinKinloch:
Is this better? #43004

mikemcquaid:
@ColinKinloch Could you use a single PR rather than two? I guess what I'm trying to understand: why would you not want to build the bindings? If there's no real reason: let's not make it optional at all.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/6c5a3d2c3c9e24c2e094599b1f36d952c9956e7b. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

Issue #42965, opened 2015-08-15
Formula renames bug [Tracking]


xu-cheng:
* [x] Link problem **critical bug** #42857
* [x] Better update output #42801
* [x] `--force` should bypass tap origin check #42801
* [x] Migration and any other operations which manipulate keg(old or new) (e.g. uninstall) should lock keg. **critical bug** #43025
* [x] `Formulary#to_rack` should follow the symlink. f857815178885c7c82127d0b0a9f4639d0b2e035
* [x] Old opt_link should be handled for `brew switch/(un)link/upgrade`. **critical bug** #42998
* [ ] Performance issue for `brew update` and potentially other brew commands. We can use `brew prof`(available from `brew tap homebrew/dev-tools`) to do some analysis.
* [x] The variable names inside `migrator.rb` should have uniform naming style. For example, `foo_bar` vs `foobar`. All the variables and methods would better to have proper document as comment. Variable names should reflect their purpose, e.g.:
  * We should use `new_cellar` instead of `new_path`.
  * We should use `old_linked_keg` instead of `old_keg`.
  * We should make variables named after `xxx_keg` a `Keg` object rather than `Pathname` object.
  * We should avoid variable name as `keg`, which is confusing as we don't know it's old or new.
* [x] Possible bugs regarding migration advice for `install/outdated`. https://github.com/Homebrew/homebrew/pull/42940#discussion_r37194059 https://github.com/Homebrew/homebrew/pull/42940#discussion_r37194427
* [x] If renamed formula has new DSL, `brew update` can crash.  #43353

xu-cheng:
cc @mikemcquaid @vladshablinsky

xu-cheng:
I suggest to fix each problem per PR.

vladshablinsky:
>  Old opt_link should be handled for brew switch/(un)link/upgrade. **critical bug**

Ho do we implement it? There was a little discussion about that.

* I tried to get `oldname` using `INSTALL_RECEIPT` and loading the formula, but we decided, that it wasn't the best idea.

* You suggested to get `oldname` from the symlinks to the new name keg in `opt`.
  * We are able to do this after the migration occurs if `oldname` was optlinked.
  * If there is a link in opt pointing to some of `newname`'s keg and the name of that link is not `newname`, then we can say, that `oldname` is a name of that link.

According to approach you suggested, we don't need to remove `Migrator#link_oldname_opt` and `Migrator#unlink_oldname_opt`. What we need is to 

[...]

PR #42998, opened 2015-08-16
handle old opt_link for brew switch/(un)link/upgrade

vladshablinsky:
Fix opl opt link from https://github.com/Homebrew/homebrew/issues/42965

* brew upgrade/switch relinks old name opt to new keg


xu-cheng:
Thanks @vladshablinsky

PR #43007, opened 2015-08-16
sickrage 4.0.50 (new formula)

petemcw:
I'd like to add this new formula to Homebrew. Thanks for taking a look.

### DIFF
--- a/Library/Formula/sickrage.rb
+++ b/Library/Formula/sickrage.rb
@@ -1,0 +1,113 @@
+class Sickrage < Formula
+  desc "Automatic video library manager for TV Shows"
+  homepage "https://www.sickrage.tv"
+  url "https://github.com/SiCKRAGETV/SickRage/archive/v4.0.66.tar.gz"
+  sha256 "37e74190b1393c4cd2fafc72581ced6f9188e5c90ad3af9192706e2173e1db6f"
+  head "https://github.com/SiCKRAGETV/SickRage.git"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on "openssl"
+
+  resource "Cheetah" do
+    url "https://pypi.python.org/packages/source/C/Cheetah/Cheetah-2.4.4.tar.gz"
+    sha256 "be308229f0c1e5e5af4f27d7ee06d90bb19e6af3059794e5fd536a6f29a9b550"
+  end
+
+  resource "Markdown" do
+    url "https://pypi.python.org/packages/source/M/Markdown/Markdown-2.6.2.tar.gz"
+    sha256 "ee17d0d7dc091e645dd48302a2e21301cc68f188505c2069d8635f94554170bf"
+  end
+
+  resource "pyOpenSSL" do
+    url "https://pypi.python.org/packages/source/p/pyOpenSSL/pyOpenSSL-0.15.1.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Need an `openssl` dependency on this.

The installation hangs pretty badly so I've aborted it for now. Does the whole `brew install` + `brew test --verbose` work for you locally as expected?

petemcw:
@DomT4 I'll check it out later this evening. The `brew install` + `brew test` does work for me locally, but honestly the test is probably not ideal. I was hoping to get something more substantial than the "version" or "help" test but it may not be an option for SickRage.

petemcw:
I updated the dependencies but I also had to remove the test block. I determined that the test wasn't actually working for me. Unfortunately I don't think I can do even a "bad test" for the version or help because SickRage returns a non-zero exit code even when invoking help.

dunn:
We can handle non-zero exit codes. Something like this should work:
```ruby
assert_match version.to_s, shell_output("#{bin}/sickrage --version", 1)
```
Change `1` to whatever the exit code is.

Library/Formula/sickrage.rb
```diff
@@ -0,0 +1,113 @@
...
+      export PYTHONPATH="#{libexec}/vendor/lib/python2.7/site-packages:$PYTHONPATH"
+      python "#{libexec}/SickBeard.py"\
+        "--pidfile=#{var}/run/sickrage.pid"\
+        "--datadir=#{etc}/sickrage"\
+        "$@"
+    EOS
+  end
+
+  test do
+    assert_match //, shell_output("#{bin}/sickrage --help", 1)
```
DomT4: Can we actually match something here, Thanks.

petemcw: @DomT4 I'm not sure that we can. I've been browsing through the source code and in all cases where the script exits, output is printed but not technically returned. So the `shell_output` doesn't register anything but the non-zero exit status. Originally I tried to write a test that would actually launch the service, check for a PID or config file, then terminate the process. But that was the test that kept hanging.

bfontaine: > output is printed but not technically returned

I think you mean it gets printed on stderr instead of stdout. Add `2>&1` at the end of the command to change that.

dunn: If output is printed to `stderr`, you can capture it with `sickrage --help 2>&1`

petemcw:
Would you guys check this out again? All checks look like they are passing.

Library/Formula/sickrage.rb
```diff
@@ -0,0 +1,113 @@
...
+      #!/bin/bash
+      export PYTHONPATH="#{libexec}/vendor/lib/python2.7/site-packages:$PYTHONPATH"
+      python "#{libexec}/SickBeard.py"\
+        "--pidfile=#{var}/run/sickrage.pid"\
+        "--datadir=#{etc}/sickrage"\
+        "$@"
+    EOS
+  end
+
+  test do
```
dunn: Can you make the test block the very last item? Thanks!

petemcw:
@dunn I've moved the test block to the bottom.

dunn:
@BrewTestBot test this please

dunn:
Thanks for your patience! Merged in e48380297d80a08cd6fb9a7b82c33bda2a76a271.

PR #43014, opened 2015-08-17
enable sandbox on test-bot

xu-cheng:


tdsmith:
is setting `ENV[“HOMEBREW_SANDBOX”]` near the top of test-bot.rb sufficient?

xu-cheng:
Right, I just realize we use several `run_as_not_developer` in the test-bot.rb. Updated.

PR #43016, opened 2015-08-17
improve zsh_completion

xu-cheng:


bfontaine:
typo: “imporve” → “improve”

xu-cheng:
Ouch but thanks.

PR #43018, opened 2015-08-17
cloc: fix working with homebrew-perl

denji:
```
$ perl --version
This is perl 5, version 22, subversion 0 (v5.22.0) built for darwin-thread-multi-2level

$ cloc .
Can't use 'defined(@array)' (Maybe you should just omit the defined()?) at /usr/local/bin/cloc line 1284.

$ brew reinstall https://raw.githubusercontent.com/denji/homebrew/9c2a8d6/Library/Formula/cloc.rb

```

### DIFF
--- a/Library/Formula/cloc.rb
+++ b/Library/Formula/cloc.rb
@@ -11,6 +11,12 @@ class Cloc < Formula
     sha256 "a4c159ceb7f5f67d7925864a8e4f31e2c43fe2dee356afdf931ebc1000ca11a3" => :mountain_lion
   end
 
+  # remove deprecated defined(@array) to support perl 5.22 (bug 135)
+  # remove "no warnings 'deprecated';" -- no longer relevant
+  # don't write language definition for "(unknown)"
+  # http://sourceforge.net/p/cloc/bugs/135/
+  patch :DATA
+
   def install
     bin.install "cloc-#{version}.pl" => "cloc"
   end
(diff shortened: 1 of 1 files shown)


Library/Formula/cloc.rb
```diff
@@ -11,6 +11,12 @@ class Cloc < Formula
     sha256 "a4c159ceb7f5f67d7925864a8e4f31e2c43fe2dee356afdf931ebc1000ca11a3" => :mountain_lion
   end
 
+  # remove deprecated defined(@array) to support perl 5.22 (bug 135)
+  # remove "no warnings 'deprecated';" -- no longer relevant
+  # don't write language definition for "(unknown)"
+  # http://sourceforge.net/p/cloc/bugs/135/
+  patch :DATA
```
bfontaine: Can’t you use the linked patch? https://bugs.archlinux.org/task/45494?getfile=13174

denji: @bfontaine I think it won't work with standard patch `-p1`

```
--- /usr/bin/cloc	2015-06-29 01:53:16.000000000 +0800
+++ ./cloc	2015-06-29 02:11:09.849082768 +0800
```

DomT4:
I have some slight hesitation here because we don't even pretend to support using Homebrew's Perl for other Homebrew formulae.

denji:
@DomT4 Your case, if there is a new perl in any case it will stop working.

mikemcquaid:
I think we'll pass on this for the reason @DomT4 has mentioned.

PR #43020, opened 2015-08-17
allowing rescue https redirect error

xu-cheng:
It can be useful when implementing custom https mirror strategy such as #42955.


DomT4:
:+1:. I needed to write something like this for the plan for `HOMEBREW_NO_INSECURE_REDIRECT` going forwards, so you beat me to it :smiley:.

PR #43024, opened 2015-08-17
uninstall: simplify logic

xu-cheng:
Also reduce unnecessary disk operations

cc @vladshablinsky

### DIFF
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -6,23 +6,10 @@ module Homebrew
   def uninstall
     raise KegUnspecifiedError if ARGV.named.empty?
 
-    # Find symlinks that can point to keg.rack
-    links = HOMEBREW_CELLAR.subdirs.select(&:symlink?)
-
     if !ARGV.force?
       ARGV.kegs.each do |keg|
         keg.lock do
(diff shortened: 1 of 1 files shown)


vladshablinsky:
:+1:

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -64,6 +43,10 @@ def uninstall
   rescue MultipleVersionsInstalledError => e
     ofail e
     puts "Use `brew remove --force #{e.name}` to remove all versions."
+  ensure
+    # If we delete Cellar/newname, then Cellar/oldname symlink
+    # can become broken and we have to remove it.
+    HOMEBREW_CELLAR.children.select(&:symlink?).reject(&:resolved_path_exists?).each(&:unlink)
```
bfontaine: I don’t know if it’s worth it, but you might want to optimize this:

```rb
HOMEBREW_CELLAR.children.each do |c|
  c.unlink if c.symlink? && !c. resolved_path_exists?
end
```

xu-cheng: I think it's actually worth to do this optimization here as one can install formulae as many as possible. Thanks for the advice here.

PR #43025, opened 2015-08-17
lock keg during migration

xu-cheng:
cc @vladshablinsky


vladshablinsky:
Ah, now I see the right way to make locks.

There are some leaks in `test_migrator` when running `brew tests`.

PR #43028, opened 2015-08-17
unnecessary calls to .select simplified

bfontaine:
These are minor perf optimizations.

* `.select(…).each` = “create a new array by looping over the original one then loop over the new one”
* `.each` with a condition in the block = “loop over the original array without creating a new one”


bfontaine:
cc @xu-cheng

xu-cheng:
:+1:

PR #43030, opened 2015-08-17
create tomcat-runner Formula

ryenus:
A simple formula based on jetty-runner.

Thank you jetty-runner :-)

### DIFF
--- a/Library/Formula/tomcat-runner.rb
+++ b/Library/Formula/tomcat-runner.rb
@@ -1,0 +1,18 @@
+class TomcatRunner < Formula
+  desc "Run Tomcat without an installed distribution"
+  homepage "https://github.com/jsimone/webapp-runner"
+  url "https://repo1.maven.org/maven2/com/github/jsimone/webapp-runner/8.0.24.0/webapp-runner-8.0.24.0.jar"
+  sha256 "906ca6902dd481ad1b7e0c5287daefc81e111b70b6f4ec37c1bf03b48e56099d"
+
+  def install
+    libexec.install Dir["*"]
+
+    bin.write_jar_script libexec/"webapp-runner-#{version}.jar", "tomcat-runner"
+  end
+
+  test do
+    usage = shell_output("#{bin}/tomcat-runner --help", 1)
+    assert_match /--path/, usage
+    assert_match /--port/, usage
+  end
+end


Library/Formula/tomcat-runner.rb
```diff
@@ -0,0 +1,14 @@
+class TomcatRunner < Formula
+  desc "Run Tomcat without an installed distribution"
+  homepage "https://github.com/jsimone/webapp-runner"
+  url "http://repo1.maven.org/maven2/com/github/jsimone/webapp-runner/8.0.24.0/webapp-runner-8.0.24.0.jar"
```
DomT4: Should be able to use `https` here rather than `http`

Library/Formula/tomcat-runner.rb
```diff
@@ -0,0 +1,14 @@
...
+  version "8.0.24.0"
```
DomT4: I suspect this will be detected alright automatically.

Library/Formula/tomcat-runner.rb
```diff
@@ -0,0 +1,14 @@
...
+  sha256 "906ca6902dd481ad1b7e0c5287daefc81e111b70b6f4ec37c1bf03b48e56099d"
+
+  def install
+    libexec.install Dir["*"]
+
+    bin.mkpath
```
DomT4: Don't need this line.

Library/Formula/tomcat-runner.rb
ryenus: tried with `https` and it works too, so yes, thank you @DomT4 :-)

Library/Formula/tomcat-runner.rb
ryenus: yes, `brew create <jar_url>` correctly identified the version number, I guess the `url` string could be interpolated with `#{version}` to avoid duplication.

DomT4: We don't use `#{version}` in URL strings. For the sake of clarity we prefer to use the full manual URL.

Library/Formula/tomcat-runner.rb
ryenus: removed, thanks!

ryenus:
Oops, `tomcat-runner --help` exiit with status code `1` :-(, probably a bug?

DomT4:
> probably a bug of itself in terms of exit code logic?

Unlikely. Some projects deliberately throw "bad" exit codes on checks that don't actually do anything with the tool, i.e. `help` and `version` and such. Is there anything we can do to test functionality here at all?

ryenus:
I think we can capture the output of `tomcat-runner --help`, then check if `--path` and/or `--port` appears, these are the key options that the usage info should always include.

@DomT4 any example test alike?

DomT4:
> any example test alike?

```ruby
test do
  assert_match /something to look for/, shell_output("#{bin}/"tomcat-runner --help", 1)
end
```

Something like that should work. You may or may not need the `1` at the end there.

ryenus:
Ah, the test is failing due to the jar artifact is not available in the test environment/sandbox.

How to correct this? HELP WANTED :-)

> ==> brew test --verbose tomcat-runner					 FAILED
Testing tomcat-runner
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20150817-33289-1b5hudy.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Formula/tomcat-runner.rb --verbose
==> /usr/local/Cellar/tomcat-runner/8.0.24.0/bin/tomcat-runner --help
**Error: Unable to access jarfile /usr/local/Cellar/tomcat-runner/8.0.24.0/libexec/tomcat-runner-8.0.24.0.jar**

ryenus:
it works now! @DomT4, thank you very much~

Library/Formula/tomcat-runner.rb
```diff
@@ -0,0 +1,18 @@
+class TomcatRunner < Formula
```
mikemcquaid: Should be `WebappRunner`

Library/Formula/tomcat-runner.rb
```diff
@@ -0,0 +1,18 @@
...
+  desc "Run Tomcat without an installed distribution"
+  homepage "https://github.com/jsimone/webapp-runner"
+  url "https://repo1.maven.org/maven2/com/github/jsimone/webapp-runner/8.0.24.0/webapp-runner-8.0.24.0.jar"
+  sha256 "906ca6902dd481ad1b7e0c5287daefc81e111b70b6f4ec37c1bf03b48e56099d"
+
+  def install
+    libexec.install Dir["*"]
+
```
mikemcquaid: Don't need this empty line.

Library/Formula/tomcat-runner.rb
```diff
@@ -0,0 +1,18 @@
...
+    bin.write_jar_script libexec/"webapp-runner-#{version}.jar", "tomcat-runner"
```
mikemcquaid: Should be called `webapp-runner`

ryenus: @mikemcquaid, thanks for looking into this!

Yeah, I would name it `webapp-runner` at the first sight, but clearly this tool only support tomcat, which is explicit in its README:

> Webapp runner is designed to allow you to launch an exploded or compressed war that is on your filesystem into a tomcat container with a simple java -jar command. It supports the following version of Tomcat:

> Tomcat 7: tomcat7 branch
> Tomcat 8: master branch

Also, isn't `webapp-runner` too broad? considering there's also jetty-runner which support jetty, and a lot more different webapps/platforms beyond the Java ecosystem.

bfontaine:
Ping re Mike’s comments.

Library/Formula/tomcat-runner.rb
bfontaine: We use the name upstream provides, so this should be `webapp-runner` even if it only supports Tomcat.

bfontaine:

PR #43042, opened 2015-08-17
Libmill: Go-style concurrency in C

benjolitz:
Create a recipe that will allow for building of libmill (http://libmill.org/) on OSX that passes tests and is suitable for development.

### DIFF
--- a/Library/Formula/libmill.rb
+++ b/Library/Formula/libmill.rb
@@ -1,0 +1,49 @@
+class Libmill < Formula
+  desc "Go-style concurrency in C"
+  homepage "http://libmill.org/"
+  url "http://libmill.org/libmill-0.16-beta.tar.gz"
+  sha256 "c6d68d8291dadf2ee111f6ceed057f2fd267e7dc334bef8d15b6316c35b6a0ca"
+  head "https://github.com/sustrik/libmill.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "cmake" => :build
+
+  def install
+    # Issue 21 at https://github.com/sustrik/libmill/issues/21
+    # Due to a bug in Clang's optimizer, -O2 and above will
+    # use the same address for mill_val in distinct and seperate coroutines,
+    # which breaks everything.
+    ENV.O1 if ENV.compiler == :clang
+
+    system "./autogen.sh" if build.head?
+    system "./configure", "--prefix=#{prefix}"
+    system "make", "all", "check", "install"
+  end
+
+  test do
+    (testpath/"test.c").write <<-EOS.undent
+      #include <stdio.h>
+      #include <libmill.h>
+
+      void worker(int count, const char *text) {
+          int i;
+          for(i = 0; i != count; ++i) {
+              printf("%s\\n", text);
+              msleep(10);
+          }
+      }
+
+      int main() {
+          go(worker(4, "a"));
+          go(worker(2, "b"));
+          go(worker(3, "c"));
+          msleep(100);
+          return 0;
+      }
+    EOS
+    system ENV.cc, "test.c", "-L#{lib}", "-lmill", "-o", "test"
+    system "./test"
+  end
+end


Library/Formula/libmill.rb
```diff
@@ -0,0 +1,29 @@
+class Libmill < Formula
+  desc "Go-style concurrency in C"
+  homepage "http://libmill.org/"
+  head "https://github.com/sustrik/libmill.git"
+  url "http://libmill.org/libmill-0.16-beta.tar.gz"
+  sha256 'c6d68d8291dadf2ee111f6ceed057f2fd267e7dc334bef8d15b6316c35b6a0ca'
```
DomT4: Double quotes everywhere.

Library/Formula/libmill.rb
```diff
@@ -0,0 +1,29 @@
+class Libmill < Formula
+  desc "Go-style concurrency in C"
+  homepage "http://libmill.org/"
+  head "https://github.com/sustrik/libmill.git"
```
DomT4: `head` goes beneath the `sha256` line.

Library/Formula/libmill.rb
```diff
@@ -0,0 +1,29 @@
...
+  url "http://libmill.org/libmill-0.16-beta.tar.gz"
+  sha256 'c6d68d8291dadf2ee111f6ceed057f2fd267e7dc334bef8d15b6316c35b6a0ca'
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "cmake" => :build
+  env :std
```
DomT4: We are extremely unlikely to accept this if it requires `:std` env to build. We're phasing that support out. Can you elaborate on the build issues under superenv?

Library/Formula/libmill.rb
```diff
@@ -0,0 +1,29 @@
...
+    if ENV.compiler == :clang
+      ENV.append_to_cflags "-O1"
+    end
+    system "./configure", "--prefix=#{prefix}"
+    system "make"
+    system "make", "check"
+    system "make", "install"
+  end
+
+  test do
```
DomT4: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

benjolitz: make check runs libmill's unit tests. Those unit tests must be run from within the project directory.

I did not see a way to run 'make check' from test do.

Library/Formula/libmill.rb
benjolitz: the superenv causes the injection of the CFLAG '-O2' which provokes an optimization bug in libmill.

Setting env :std allows libmill to successfully set it's own CFLAGS on detection of Clang. 

I could not make it build without env :std. 

Perhaps someone who better understands homebrew can figure out what magic is needed to keep it from setting anything other than -O1 as a compilation flag.

mistydemeo: You can pass `ENV.O1` to explicitly set the optimization level.

Library/Formula/libmill.rb
DomT4: Sorry, you misunderstood me on this. I meant double quotes everywhere is the style we use, rather than change all the double quotes to single quotes :smiley:

Library/Formula/libmill.rb
DomT4: You can't. `buildpath` and `testpath` never exist at the same time. You'll need to write a test. We use `test do` blocks to ensure installation didn't break anything after the `make check` stage and when dependencies are updated to ensure this formula still functions with them.

Library/Formula/libmill.rb
benjolitz: Thanks. That did it.

Library/Formula/libmill.rb
benjolitz: Corrected.

Library/Formula/libmill.rb
benjolitz: corrected in 3b175ead2adaa966193c968408a6cbac4aa3bb3f

Library/Formula/libmill.rb
benjolitz: Test added in 23c3c85 from libmill's test vectors after looking at tinyxml.rb as an example.

Library/Formula/libmill.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "cmake" => :build
+
+  def install
+    ENV.O1
+    if build.head?
+      system "./autogen.sh"
```
DomT4: Squash this onto one line:

```ruby
system "make"
```

Etc.

DomT4:
Just the one nit here and then I'm happy at least. If you could squash your commits back down to one I'd be appreciative.

benjolitz:
@DomT4 the make commands have been folded into one line and the commits are squashed.

benjolitz:
corrected a most unfortunate warning with ```brew audit```. Apparently "make all check install" must be "make", "check", "all", "install".

Oh well, pushed.

benjolitz:
hello? Seems everything  has passed checks. Anything else?

Library/Formula/libmill.rb
```diff
@@ -0,0 +1,45 @@
...
+  sha256 "c6d68d8291dadf2ee111f6ceed057f2fd267e7dc334bef8d15b6316c35b6a0ca"
+  head "https://github.com/sustrik/libmill.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "cmake" => :build
+
+  def install
+    ENV.O1
```
mikemcquaid: Can you please submit this as an issue to the upstream developers of this project and add a link to the upstream issue submission as in a comment in the formula file. Thanks!

Library/Formula/libmill.rb
```diff
@@ -0,0 +1,45 @@
+class Libmill < Formula
+  desc "Go-style concurrency in C"
+  homepage "http://libmill.org/"
+  url "http://libmill.org/libmill-0.16-beta.tar.gz"
```
mikemcquaid: We need to use a non-beta version for now.

benjolitz: Not possible. If you look at http://libmill.org/download.html you would notice that there are no '-rc' or '-beta' free releases. That is because the project is still in version 0.x releases, which by many interpretations of semantic versioning means things are subject to change.

Library/Formula/libmill.rb
benjolitz: The issue was known before I made this recipe. That was why I used env :std and switched to ENV.O1 (for a superenv).

See https://github.com/sustrik/libmill/issues/21

A comment has been added to the recipe. The fault is a Clang bug where it reuses a stack allocated address (which shouldn't happen and does not happen in other compilers)

Library/Formula/libmill.rb
mikemcquaid: Maybe this should be in the `homebrew-devel-only` tap instead?

Library/Formula/libmill.rb
```diff
@@ -0,0 +1,45 @@
...
+  sha256 "c6d68d8291dadf2ee111f6ceed057f2fd267e7dc334bef8d15b6316c35b6a0ca"
+  head "https://github.com/sustrik/libmill.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "libtool" => :build
+  depends_on "cmake" => :build
+
+  def install
+    ENV.O1  # Issue 21 at https://github.com/sustrik/libmill/issues/21
```
mikemcquaid: Can you put this comment above and add a short description of the issue rather than just a link? Thanks!

benjolitz: Added. Also made it conditional if and only if the compiler is clang.

Library/Formula/libmill.rb
benjolitz: Made a pull request there.

PR #43045, opened 2015-08-17
samba: install .pc files

### LINKED ISSUE #43040: samba pkg-config complains about missing devel files.
`pkg-config --libs --cflags smbclient`

gives

```
Package smbclient was not found in the pkg-config search path.
Perhaps you should add the directory containing `smbclient.pc'
to the PKG_CONFIG_PATH environment variable
No package 'smbclient' found
```

mistydemeo:
Fixes #43040.


DomT4:
You beat me to this, so I removed this bit from the PR I was going to file and handled the other stuff that was bothering me in #43046 :)

PR #43046, opened 2015-08-17
samba: install default conf

DomT4:
Mimics the way Debian does this, and is generally more Homebrewy these days than skipping over directories inside the prefix. Also install an example configuration file rather than touching an empty one.

Tested changes by running `smbd` with success.


mistydemeo:
Worth adding a `post_install` to migrate users' existing data?

DomT4:
Oh sure, that's a neat idea. Will check it out.

mistydemeo:
:shipit:

PR #43049, opened 2015-08-17
install: don't warn if modified formulae don’t have bottles

bfontaine:
I also re-ordered the conditions to have the `file_modified?` check done last.

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -180,7 +180,7 @@ class FormulaInstaller
     build_bottle_preinstall if build_bottle?
 
     unless @poured_bottle
-      if formula.file_modified? && !build_from_source?
+      if formula.bottle && !build_from_source? && formula.file_modified?
         filename = formula.path.to_s.gsub("#{HOMEBREW_PREFIX}/", "")
         opoo "Formula file is modified!"
         puts "Building from source because #{filename} has local changes"


dunn:
Thanks for picking this up, I've been having trouble nailing down the problem behaviors.  Would adding `!build_bottle?` to the conditions be redundant?

mikemcquaid:
This is good but we also need to handle the bot's case somehow; even if it's just setting a stupid environment variable there to disable this check. Eventually I'd also like this to not get trigger if I e.g. checkout old versions of formulae.

bfontaine:
@dunn the condition already has `!build_from_source?`, is it possible for `!build_from_source? && !build_bottle?` to be true?

mikemcquaid:
Another failure case: shallow clones.

DomT4:
Another niche failure case: local testing of `def pour_bottle` blocks.

```ruby
  bottle do
    sha256 "17ed246b275ce29226ea244eb46b91a5d79979368323dcbc4a143083291af84d" => :yosemite
  end

  def pour_bottle?
    return false unless which("python") == HOMEBREW_PREFIX/"bin/python"
    return false unless which("ruby") == HOMEBREW_PREFIX/"bin/ruby"
    return false unless Formula["perl"].installed?
  end
```

```
~> brew install -v vim
Building from source because Library/Formula/vim.rb has local changes
```

You can't pass `--force-bottle` in this situation because it installs the things necessary to fulfil the `pour_bottle` conditions, which means you never get the chance to test the `return false` conditions:

```
==> Installing dependencies for vim: perl
==> Installing vim dependency: perl
==> Downloading https://homebrew.bintray.com/bottles/perl-5.22.0.yosemite.bottle.tar.gz
```

PR #41930, opened 2015-07-20
ghc: try allowing `-Wno-unicode`. See #41777

geoff-codes:
Discussion in #41777.

Issue #41777, opened 2015-07-16
ghc formula can't be used as a stage0 compiler for GHC itself


rodlogic:
The issue has been reported [here](https://ghc.haskell.org/trac/ghc/ticket/10543#comment:3) and is caused by the following line: https://github.com/Homebrew/homebrew/commit/b160b165d42f07abc912b1903095c593de61a0df#diff-e06a9de6bb362e22e1ceb9a66eaf171eR72

The above line doesn't seem to be needed in this specific case, since GHC build system will select the right default for each platform. The question is why was this line needed in the first place. Can it be removed?

Here are the steps to reproduce on OS X 10.10:
```bash
brew update
brew install ghc
git clone --recursive git://git.haskell.org/ghc.git
cd ghc
./validate
```

And this is the error:
```
libraries/Cabal/Cabal/Distribution/Simple/BuildTarget.hs:392:24:
     error: \u used with no following hex digits; treating as '\' followed by identifier [-Werror,-Wunicode]
matchBuildTarget pkg = \utarget fexists ->
                       ^
1 error generated.
make[1]: *** [utils/ghc-cabal/dist/build/tmp/ghc-cabal] Error 1
make: *** [all] Error 2
```

mikemcquaid:
@geoff-codes Thoughts?
@rodlogic Could you submit a PR to remove it? Thanks!

geoff-codes:
So, the line referenced above was added to override https://github.com/ghc/ghc/blob/master/aclocal.m4#L2263, as not all the supported versions of clang (i.e., the ones that Apple provides) support these `-Wno-`s. We may still have been targeting 10.7 still at the time — I don't remember precisely — so I'm more than happy to take the line out and see what @BrewTestBot makes of it. But if we get build failures, a decision will have to be made.

[...]

PR #43053, opened 2015-08-18
Rev bump ghc due to #41777?

geoff-codes:
I just realized it might be warranted to rev-bump ghc after #41930? Because the issue in #41777 will not be resolved without a reinstall. I'll leave this up to you maintainers, feel free to just close this if you don't think its necessary.

### DIFF
--- a/Library/Formula/ghc.rb
+++ b/Library/Formula/ghc.rb
@@ -4,6 +4,8 @@ class Ghc < Formula
   url "https://downloads.haskell.org/~ghc/7.10.2/ghc-7.10.2-src.tar.xz"
   sha256 "54cd73755b784d78e2f13d5eb161bfa38d3efee9e8a56f7eb6cd9f2d6e2615f5"
 
+  revision 1
+
   bottle do
     revision 1
     sha256 "cc097abdfd1d392ffe6f8193627168279441fb868555823ac5aa75c41f2d9e74" => :yosemite


Library/Formula/ghc.rb
```diff
@@ -4,6 +4,8 @@ class Ghc < Formula
   url "https://downloads.haskell.org/~ghc/7.10.2/ghc-7.10.2-src.tar.xz"
   sha256 "54cd73755b784d78e2f13d5eb161bfa38d3efee9e8a56f7eb6cd9f2d6e2615f5"
 
+  revision 1
```
geoff-codes: Should this be `revision 2`?

Library/Formula/ghc.rb
```diff
@@ -4,6 +4,8 @@ class Ghc < Formula
...
+
   bottle do
     revision 1
```
geoff-codes: Because of this? What's the difference between a bottle revision and a formula revision?

Library/Formula/ghc.rb
DomT4: Nope. Default is `0`, which is unwritten.

Library/Formula/ghc.rb
DomT4: Formula revisions force the update on everyone. Bottle revisions are just an internal mechanism to refresh the bottle for whatever reason (Updated resources, non-urgent patching, configuration changes, sha1 => sha256, etc etc). Despite the wording they are essentially two different functions.

geoff-codes: Got it.

geoff-codes:
I'm probably gonna close this in a couple hours unless anyone feels strongly otherwise.

geoff-codes:
Closing. For the record, the build failure here is due to exceeding the 2 hour time limit, nothing more.

PR #43054, opened 2015-08-18
Update osm2pgsql to 0.88.1

pnorman:
0.88.1 includes the patch which was being cherry-picked, so the the build is simpler.


bfontaine:
Thank you for your contribution, @pnorman, we appreciate it! :tada: 

For the record the preferred commit message format for version bumps is “`osm2pgsql 0.88.1`”.

PR #42988, opened 2015-08-16
jq 1.5

nicowilliams:
Update to jq 1.5.

PR #43055, opened 2015-08-18
jq: segregate build dependencies

tzudot:
Execute `autoreconf`only when built with options, HEAD or devel. 
This has been changed in PR [#42988] (https://github.com/Homebrew/homebrew/pull/42988). Reverting to original behavior.

### DIFF
--- a/Library/Formula/jq.rb
+++ b/Library/Formula/jq.rb
@@ -13,16 +13,17 @@ class Jq < Formula
 
   head do
     url "https://github.com/stedolan/jq.git"
+
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
   end
 
   depends_on "oniguruma"  # jq depends > 1.5
-  depends_on "autoconf" => :build
-  depends_on "automake" => :build
-  depends_on "libtool" => :build
   depends_on "bison" => :build # jq depends on bison > 2.5
 
   def install
-    system "autoreconf", "-iv"
+    system "autoreconf", "-iv" unless build.stable?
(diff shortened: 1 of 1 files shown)


Library/Formula/jq.rb
```diff
@@ -22,7 +22,7 @@ class Jq < Formula
   depends_on "bison" => :build # jq depends on bison > 2.5
 
   def install
-    system "autoreconf", "-iv"
+    system "autoreconf", "-iv" if build.devel? || build.head?
```
DomT4: There is no `devel`, so the first part of this is redundant. If it only applies to `head`, can the autotool dependencies be moved into a `head do` block?

tzudot: Haven't changed the `if` condition to reflect the removal of `devel` block as wouldn't matter, and might be required in future. But yes, `build.devel?` is redundant here. Should I remove it?

Yes, `jq` builds with `autoconf`, `automake` and `libtool` in `head` block.

DomT4: Unless it exists now, we don't need it. It's easily added in future if it becomes necessary.

Great, let's tuck those autotool deps in `head do` with the `head` url inside that block then.

tzudot:
@DomT4 checks for this PR are done.

Library/Formula/jq.rb
bfontaine: We could use `unless build.stable?` to be future-proof.

dunn:
Thanks for catching that! Merged in 2ddeff0e602a6959a8599e9527d4388eaa9924c4.

PR #43060, opened 2015-08-18
Recommend setuid for mtr

charliesome:
mtr drops privileges immediately after starting up and opening raw sockets:

https://github.com/traviscross/mtr/blob/7dbc602cd558d8a0835a323511e4997b8d5d34c1/mtr.c#L562-L579

It should be fairly safe to setuid.

cc @mikemcquaid

### DIFF
--- a/Library/Formula/mtr.rb
+++ b/Library/Formula/mtr.rb
@@ -36,14 +36,15 @@ class Mtr < Formula
   end
 
   def caveats; <<-EOS.undent
-    mtr requires superuser privileges. You can either run the program
-    via `sudo`, or change its ownership to root and set the setuid bit:
+    mtr requires superuser privileges for raw sockets access. You can either
+    run the program via `sudo`, or change its ownership to root and set the
+    setuid bit:
 
       sudo chown root:wheel #{sbin}/mtr
       sudo chmod u+s #{sbin}/mtr
 
-    In any case, you should be certain that you trust the software you
-    are executing with elevated privileges.
+    mtr drops privileges immediately after starting up and should be fairly
+    safe to setuid.
     EOS
   end
 end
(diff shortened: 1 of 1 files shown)


Library/Formula/mtr.rb
```diff
@@ -36,8 +36,15 @@ def install
   end
 
   def caveats; <<-EOS.undent
-    mtr requires root privileges so you will need to run `sudo mtr`.
-    You should be certain that you trust any software you grant root privileges.
+    mtr requires superuser privileges for raw sockets access. You can either
+    run the program via `sudo`, or change its ownership to root and set the
+    setuid bit:
+
+      sudo chown root:wheel #{sbin}/mtr
```
bfontaine: Wouldn’t this break `brew rm mtr`?

DomT4: Yes, almost certainly. I think the current caveat is fine, personally.

charliesome: Deleting a file only requires write permission to the containing directory, the permissions on the file itself don't matter.

DomT4: Ah, yes. Apologies - I misread and thought you were chowing the whole directory, which would cause issues.

mikemcquaid:
@charliesome I'd probably rather not do this not because it isn't safe but because it will encourage other formulae (which aren't safe) to do the same. My concern here is also if there's a bug in `mtr` that stops it dropping privileges or e.g. a buffer overflow which allows code execution then having it be `root` is more concerning.

tl;dr: the people who I trust with `setuid` will already know how to set it.

DomT4:
Going to pass on this per Mike's comment. If you're comfortable setting it yourself though, by all means, feel free. Thanks for the PR!

charliesome:
No worries :sunglasses:

PR #43062, opened 2015-08-18
glew 1.13.0

xpicox:
Updated the link to the latest version of glew following the guidelines.

### DIFF
--- a/Library/Formula/glew.rb
+++ b/Library/Formula/glew.rb
@@ -1,8 +1,8 @@
 class Glew < Formula
   desc "OpenGL Extension Wrangler Library"
   homepage "http://glew.sourceforge.net/"
-  url "https://downloads.sourceforge.net/project/glew/glew/1.12.0/glew-1.12.0.tgz"
-  sha256 "af58103f4824b443e7fa4ed3af593b8edac6f3a7be3b30911edbc7344f48e4bf"
+  url "https://downloads.sourceforge.net/project/glew/glew/1.13.0/glew-1.13.0.tgz"
+  sha256 "aa25dc48ed84b0b64b8d41cdd42c8f40f149c37fa2ffa39cd97f42c78d128bc7"
 
   bottle do
     cellar :any


dunn:
The test for glew itself and two of its dependencies are failing since we're running them in a sandbox now and they're trying to write outside it: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/31775/

Would you mind seeing if you can update the tests, too? Thanks!

xpicox:
How can I emulate your sandbox on my computer?

dunn:
Add the `--sandbox` flag when testing (`brew test -v --sandbox glew`), or set the environmental variable `HOMEBREW_SANDBOX` to `1`.

xpicox:
The sandbox.test.log is empty, does it mean that the test wasn't trying to write outside the sandbox
?

dunn:
Not sure. Does the test fail locally, and if so what is the output of the `brew test`?

xpicox:
The output is:
Testing glew
==> /usr/local/Cellar/glew/1.13.0/bin/glewinfo

xpicox:
On your sandbox the glewinfo test was failing because was unable to create an OpenGL context.

DomT4:
@BrewTestBot test this please

xu-cheng:
I think the test fails because the headless env rather than sandbox. Rerun it now. If it's sandbox to blame,   it should show something in logs.

mikemcquaid:
@xpicox Can you try to replace the test with a simpler one that doesn't use an OpenGL context? Thanks!

bfontaine:
Ping?

xpicox:
An OpenGL context is required for this library. It would be nice if the test would give us more details about the failure, but i'm not the maintainer of the library. Moreover, on my machine i can't reproduce the error.

dunn:
This seems to work as a test, but it would mean an X11 dependency:
```
  test do
    (testpath/"test.c").write <<-EOS.undent
      #include <GL/glew.h>
      #include <GL/glut.h>

      int main(int argc, char** argv) {
        glutInit(&argc, argv);
        glutCreateWindow("GLEW Test");
        GLenum err = glewInit();
        if (GLEW_OK != err) {
          return 1;
        }
        return 0;
      }
    EOS
    system ENV.cc, "-I/opt/X11/include", "-L/opt/X11/lib",
           "-lglew", "-lglut", testpath/"test.c", "-o", "test"
    system "./test"
  end
```

mikemcquaid:
@xpicox Could you try to rebase this on `master`? Thanks!

Issue #42031, opened 2015-07-22
Latest update to poppler, 0.34.0 breaks gdal --with-complete


JesseCrocker:
Relevant portion of gdal build log https://gist.github.com/JesseCrocker/d6da469ec906c1890e75

Reverting to poppler 0.29.0 fixes the issue.

DomT4:
Will probably be fixed when someone submits a version update to the [latest version](https://trac.osgeo.org/gdal/wiki/DownloadSource). Updating `gdal` tends to be quite involved, which is the delay. If you want to take a crack at it, we'd welcome a PR!

Otherwise we'll need to dump an older `poppler` in `libexec` or something.

PR #43071, opened 2015-08-18
gdal: switch PDF driver linking to libpodofo

dakcarto:
Addresses #42031, where `gdal` fails to build against latest `poppler`.

As per the [Geospatial PDF](http://www.gdal.org/frmt_pdf.html) driver page...

>Starting with GDAL 1.9.0, as an alternative, the PDF driver can be compiled against libpodofo (LGPL-licensed) to avoid the libpoppler dependency. This is sufficient to get the georeferencing and vector information. However, for getting the imagery, the pdftoppm utility that comes with the poppler distribution must be available in the system PATH. ...

So this PR adds `podofo` and sets it to be linked against, while still requiring `poppler` for its command line utility. I looked into setting up a patch from the v2.0 branch, but IMO too much has changed for it to be worth maintaining a patch for 1.11 branch.

I did not do extensive testing, but it compiles on 10.9 and `gdalinfo` reads a geospatial PDF using that driver.

### DIFF
--- a/Library/Formula/gdal.rb
+++ b/Library/Formula/gdal.rb
@@ -3,7 +3,7 @@ class Gdal < Formula
   homepage "http://www.gdal.org/"
   url "http://download.osgeo.org/gdal/1.11.2/gdal-1.11.2.tar.gz"
   sha256 "66bc8192d24e314a66ed69285186d46e6999beb44fc97eeb9c76d82a117c0845"
-  revision 2
+  revision 3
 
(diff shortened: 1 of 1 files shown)


mikemcquaid:
I think this probably warrants a `revision` bump?

dakcarto:
Indeed. Bumped the revision.

mikemcquaid:
Thanks @dakcarto!

Issue #42957, opened 2015-08-14
llvm formula's clang++ fails with -L $(brew --prefix llvm)/lib


jwatzman:
On a system with clang from the llvm formula (`brew install llvm --with-clang`), consider the following code:

```c++
#include <stdexcept>

int main(void) {
  throw new std::length_error("blah");
  return 0;
}
```

When compiled as one normally would do:

```sh
$ $(brew --prefix llvm)/bin/clang++ test.cpp
$
```

everything is fine. However, when compiled like this, it fails:

```sh
$ $(brew --prefix llvm)/bin/clang++ test.cpp -L $(brew --prefix llvm)/lib
Undefined symbols for architecture x86_64:
  "typeinfo for std::length_error", referenced from:
      typeinfo for std::length_error* in test-aeb26f.o
  "vtable for std::length_error", referenced from:
      _main in test-aeb26f.o
  NOTE: a missing vtable usually means the first non-inline virtual member function has no definition.
ld: symbol(s) not found for architecture x86_64
clang-3.6: error: linker command failed with exit code 1 (use -v to see invocation)
$
```

The cause for this is, as far as I can tell, the `-L` option causing `ld` to find `libc++.dylib` in `$(brew --prefix llvm)/lib` instead of `/usr/lib`. I would expect this to work, even be the default, but it does not appear to work. I'm not sure what's wrong with `$(brew --prefix llvm)/lib/libc++.dylib`, but that's the root cause, I suspect?

This wouldn't normally be a huge deal -- that `-L` option is pretty weird and superfluous -- except that brew's superenv causes this to happen for any package that has a `depends_on "llvm" => "with-clang"` and uses that `clang` to build, since superenv injects its own `ld` wrapper which adds `-L` options for every keg-only formula such as `llvm`.

mikemcquaid:
We don't support building in superenv using the LLVM formula's `clang`.

jwatzman:
- Although I hit this due to superenv, since you can still repro it outside of superenv, I think it's still a problem with the `libc++.dylib` shipped with llvm+clang?
- If you don't support using LLVM's `clang` to build in superenv, should I use the standard env instead? The project I'm working on a formula for, HHVM, doesn't build otherwise -- OS X's clang doesn't have TLS support, and building with an upstream gcc (4.9 or something) causes ld to trip an assert for unclear reasons.
- Out of curiosity, why isn't building in superenv with LLVM's `clang` supported?

[...]

PR #43074, opened 2015-08-18
hhvm 3.9.0 (new formula)

jwatzman:
Add a formula for HHVM, Facebook's JIT compiler/runtime for the PHP and
Hack languages.

### DIFF
--- a/Library/Formula/hhvm.rb
+++ b/Library/Formula/hhvm.rb
@@ -1,0 +1,133 @@
+class Hhvm < Formula
+  desc "JIT compiler and runtime for the PHP and Hack languages"
+  homepage "http://hhvm.com/"
+  url "http://dl.hhvm.com/source/hhvm-3.9.0.tar.bz2"
+  sha256 "99e6e04c88e09442be5c40fd74e9eb2d816b9adfd94c76c30ed9ae1587198503"
+
+  head "https://github.com/facebook/hhvm.git"
+
+  option "with-debug", <<-EOS.undent
+    Make an unoptimized build with assertions enabled. This will run PHP and
+    Hack code dramatically slower than a release build, and is suitable mostly
+    for debugging HHVM itself.
+  EOS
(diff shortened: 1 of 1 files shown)


jwatzman:
In the interest of full disclosure,
- I'm one of the upstream maintainers of HHVM. The formula cookbook suggests not submitting software for which you are a maintainer, unless it's quite popular, and I think HHVM more than qualifies for this.
- We don't build with any compiler Apple ships. It was pointed out this was likely a problem for getting the formula into core in discussion on #42957. Figured I'd give it a shot anyways, can always move this into our own tap if I have to, but I'd much prefer it be in core if you'll take it.

Thanks!

mikemcquaid:
What's the expected build time here?

jwatzman:
Long. It takes like 10-20m-ish on my 12-core Mac Pro, and the build parallelizes pretty well... so like 2-3 hours or so on a single core?

<3 C++ :\

jwatzman:
Is there some possibility this would be merged, or should I get a tap set up under the official HHVM project? I'd love to have it merged into core to make it easier for our users to set up, but if that's not going to happen, I'd also love to announce an official tap :)

denji:
@jwatzman Not a good idea to stir here hhvm (main repository), there still is and extensions as php.
You need to put in a separate repository-tap. Am I right? PHP exactly [`Homebrew/homebrew-php`](https://github.com/Homebrew/homebrew-php)

jwatzman:
@denji This isn't a PHP extension, and so doesn't belong in `Homebrew/homebrew-php`. It belongs either here, or in a tap I'm going to set up under the HHVM project.

mikemcquaid:
I think we'll need to pass on this as it's struggling to build on our machines. Sorry!

jwatzman:
Thanks for letting me know, I'll set up an HHVM-official tap then :)

Issue #43073, opened 2015-08-18
CGO problems building go binary from source


sparrc:
I'm trying to get my project to build via homebrew, but am having a CGO-related issue that I can't quite work out. After it fails, the build works completely fine from the debug shell prompt.

There error I see is this:
__$WORK/github.com/influxdb/telegraf/plugins/system/ps/cpu/_obj//_cgo_.o: decoding dwarf section info at offset 0x0: too short__

This is the full output when trying to `brew install telegraf`

```
$ brew install telegraf --debug
/usr/local/Library/brew.rb (Formulary::FormulaLoader): loading /usr/local/Library/Formula/telegraf.rb
/usr/local/Library/brew.rb (Formulary::FormulaLoader): loading /usr/local/Library/Formula/go.rb
Warning: Formula file is modified!
Building from source because Library/Formula/telegraf.rb has local changes
To install from a bottle instead, run with --force-bottle
/usr/local/Library/Homebrew/build.rb (Formulary::FromPathLoader): loading /usr/local/Library/Formula/telegraf.rb
/usr/local/Library/Homebrew/build.rb (Formulary::FormulaLoader): loading /usr/local/Library/Formula/go.rb
==> Downloading https://github.com/influxdb/telegraf/archive/v0.1.4.tar.gz
Already downloaded: /Library/Caches/Homebrew/telegraf-0.1.4.tar.gz
==> go env
==> make build
/private/tmp/telegraf20150818-35390-kiqvdc/telegraf-0.1.4/bin/godep go install ./...
# github.com/influxdb/telegraf/plugins/system/ps/cpu
cannot load DWARF output from $WORK/github.com/influxdb/telegraf/plugins/system/ps/cpu/_obj//_cgo_.o: decoding dwarf section info at offset 0x0: too short
godep: go exit status 2
make: *** [prepare] Error 1
/usr/local/Library/Homebrew/debrew.rb:11:in `raise'
BuildError: Failed executing: make build
1. raise
2. ignore
3. backtrace
4. irb
5. shell
```

But when I enter a shell in debug mode, the build works fine:

```
Choose an action: 5
When you exit this shell, you will return to the menu.
tars% pwd
/private/tmp/telegraf20150818-35390-kiqvdc/telegraf-0.1.4/src/github.com/influxdb/telegraf
tars% make build
fatal: Not a git repository (or any of the parent directories): .git
go get github.com/tools/godep
[...]

PR #43076, opened 2015-08-18
telegraf: updating Formula to build from vendored dependencies

sparrc:
/cc @mistydemeo thanks for your help! This appears to have fixed issues seen in https://github.com/Homebrew/homebrew/issues/43073

### DIFF
--- a/Library/Formula/telegraf.rb
+++ b/Library/Formula/telegraf.rb
@@ -1,167 +1,34 @@
 require "language/go"
 
 class Telegraf < Formula
-  desc "Server-level agent for InfluxDB"
+  desc "Server-level metric gathering agent for InfluxDB"
   homepage "https://influxdb.com"
-  url "https://github.com/influxdb/telegraf/archive/v0.1.3.tar.gz"
-  sha256 "ed719e317e9918b314f92cdca09ab7e34fa12c357800fe31cb1c8978a2128f40"
-
-  bottle do
-    cellar :any
-    sha256 "6cb91d8c1db1d84ff45dcda761ac5cb5d6da11d652618a1952555dda5869d846" => :yosemite
-    sha256 "e145923cd453a2d7ed19a2a5912858b5926d78cc95161dfc6caacd487911d819" => :mavericks
-    sha256 "2ced74471933b8877527934828d6c1881c1f306ea8f5d0d5013774b5b721eed1" => :mountain_lion
-  end
+  url "https://github.com/influxdb/telegraf/archive/v0.1.4.tar.gz"
+  sha256 "df586d0ada7e808b62ff476efd87c5a35e69a98bc45b3bf9823751cbccecc017"
 
   depends_on "go" => :build
 
-  go_resource "github.com/influxdb/influxdb" do
-    url "https://github.com/influxdb/influxdb.git",
-      :revision => "714b47718357b7f8e5a03f0be5c98f15d9169902"
-  end
-
-  go_resource "github.com/naoina/toml" do
-    url "https://github.com/naoina/toml.git",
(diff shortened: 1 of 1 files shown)


Library/Formula/telegraf.rb
```diff
@@ -1,167 +1,34 @@
...
 
     Language::Go.stage_deps resources, buildpath/"src"
 
+    cd "src/github.com/tools/godep" do
+      system "go", "install"
+    end
+
     cd telegraf_path do
-      system "go", "build", "-ldflags", "-X main.Version 0.1.3 -X main.Commit 120218f9c6b45bc909c5b9eaa6825d5ace6713be", "cmd/telegraf/telegraf.go"
+      system("$GOPATH/bin/godep go build -o telegraf "\
```
mistydemeo: Can you split this into separate arguments? This is currently being passed to a shell, which we prefer to avoid; splitting it into separate arguments ensures it bypasses shell splitting. For example:

```ruby
system "#{buildpath}/bin/godep", "go", "build", "-o", "telegraf",
       "-ldflags", "-X main.Version #{version}",
       "cmd/telegraf/telegraf.go"
```

mistydemeo:
I didn't realize the Godeps workspace held vendored deps! Nice catch, this simplifies the formula nicely. :+1:

Library/Formula/telegraf.rb
sparrc: Okay, just pushed a change to fix this, thanks!

Library/Formula/telegraf.rb
```diff
@@ -1,167 +1,34 @@
...
 
     Language::Go.stage_deps resources, buildpath/"src"
 
+    cd "src/github.com/tools/godep" do
+      system "go", "install"
+    end
+
     cd telegraf_path do
-      system "go", "build", "-ldflags", "-X main.Version 0.1.3 -X main.Commit 120218f9c6b45bc909c5b9eaa6825d5ace6713be", "cmd/telegraf/telegraf.go"
+      system "#{buildpath}/bin/godep", "go", "build", "-o", "telegraf",
```
bfontaine: Can’t you build without `godep`?

sparrc: godep manages the vendored dependencies in the Godeps/ folder, it is required for the build to work.

bfontaine: True, I forgot about the path rewriting; I use `godep` everyday and build with `go build`, but it works because `godep` rewrites my import paths.

sparrc:
I don't quite understand what happened in this test:

http://bot.brew.sh/job/Homebrew%20Pull%20Requests/31827/version=yosemite/testReport/brew-test-bot/yosemite/test_telegraf/

Some sort of ASCII character problem...? Why would it have worked on Mountain Lion?

http://bot.brew.sh/job/Homebrew%20Pull%20Requests/31827/version=mountain_lion/testReport/brew-test-bot/mountain_lion/

mistydemeo:
Mt. Lion is still using Ruby 1.8, so encodings didn't exist yet.

sparrc:
alright, this is all good to go now!

mistydemeo:
Thanks!

Library/Formula/telegraf.rb
```diff
@@ -202,7 +69,6 @@ def plist; <<-EOS.undent
   end
 
   test do
-    assert_match /InfluxDB Telegraf/, shell_output("telegraf -version")
-    assert_match /localhost:8086/, shell_output("telegraf -sample-config")
+    assert_match /Telegraf/, shell_output("telegraf -version")
```
dunn: Any chance of a more substantive test while we're at it?

PR #43078, opened 2015-08-19
ruby 2.2.3

mistydemeo:


mistydemeo:
This fails if the user already has a `/usr/local/lib/ruby`, so looking into other options.

mistydemeo:
The sitedir and vendordir are empty directory structures by default, so reproducing them shouldn't be too hard. However, Ruby doesn't have a builtin mechanism to install them one place while internally looking them up in another, so will need to hack things up a bit.

It would be easy if I could just change the values in the places in which they're embedded after building, but they're embedded into a few binaries so it's not changeable.

mistydemeo:
@BrewTestBot test this please

PR #43079, opened 2015-08-19
Adding mirror-displays

jasonm23:
Provides a terminal utility to toggle, directly set or query OSX multiple display mirroring.

### DIFF
--- a/Library/Formula/mirror-displays.rb
+++ b/Library/Formula/mirror-displays.rb
@@ -1,11 +1,8 @@
 class MirrorDisplays < Formula
-
-  homepage "https://github.com/fcanas/mirror-displays"
   desc "Toggle/set/query OSX display mirroring"
+  homepage "https://github.com/fcanas/mirror-displays"
   url "https://github.com/fcanas/mirror-displays/archive/v1.1.tar.gz"
-
   sha256 "84d4de59da3cdb655d2b38d59fc0d38c89abf6f21acf1f8fc1bff4ab9f37fbee"
-
   head "https://github.com/fcanas/mirror-displays.git"
 
   depends_on :xcode => :build
@@ -17,7 +14,9 @@ class MirrorDisplays < Formula
   end
 
   test do
-    system "#{bin}/mirror -h"
+    help_text = "Toggle mirroring (default behavior)"
+    out = system "#{bin}/mirror -h"
+    assert true, out.include?(help_text)
   end
 
 end
--- a/Library/Formula/mirror-displays.rb
+++ b/Library/Formula/mirror-displays.rb
@@ -1,0 +1,23 @@
+class MirrorDisplays < Formula
+
+  homepage "https://github.com/fcanas/mirror-displays"
+  desc "Toggle/set/query OSX display mirroring"
+  url "https://github.com/fcanas/mirror-displays/archive/v1.1.tar.gz"
+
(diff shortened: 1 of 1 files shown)


Library/Formula/mirror-displays.rb
```diff
@@ -0,0 +1,23 @@
+class MirrorDisplays < Formula
+
```
bfontaine: Please remove this empty line.

Library/Formula/mirror-displays.rb
```diff
@@ -0,0 +1,23 @@
...
+  homepage "https://github.com/fcanas/mirror-displays"
+  desc "Toggle/set/query OSX display mirroring"
```
bfontaine: Please move the `desc` above the `homepage`.

Library/Formula/mirror-displays.rb
```diff
@@ -0,0 +1,23 @@
...
+  url "https://github.com/fcanas/mirror-displays/archive/v1.1.tar.gz"
+
```
bfontaine: Please remove this empty line as well.

Library/Formula/mirror-displays.rb
```diff
@@ -0,0 +1,23 @@
...
+  depends_on :xcode => :build
+
+  def install
+    # Set to build with SDK=macosx10.6, but it doesn't actually need 10.6
+    xcodebuild "SDKROOT=", "SYMROOT=build"
+    bin.install "build/Release/mirror"
+  end
+
+  test do
+    system "#{bin}/mirror -h"
```
bfontaine: Could you write a test that does something more substantial than e.g. `-h`/`--help`/`--version`? Thanks!

Also, use `system "foo", "bar"` instead of `system "foo bar"`.

jasonm23:
Note that mirror only does something useful when a secondary display is attached.

It would be helpful if you make a concrete suggestion for the test.  Is there anything I can assume about the display setup of a test machine? I would think not.

I'll flush the blanks.

jasonm23:
Predictable output (otherwise system dependent) is only available from help, I can check text includes substring?

Library/Formula/mirror-displays.rb
```diff
@@ -0,0 +1,22 @@
...
+  def install
+    # Set to build with SDK=macosx10.6, but it doesn't actually need 10.6
+    xcodebuild "SDKROOT=", "SYMROOT=build"
+    bin.install "build/Release/mirror"
+  end
+
+  test do
+    help_text = "Toggle mirroring (default behavior)"
+    out = system "#{bin}/mirror -h"
+    assert true, out.include?(help_text)
```
bfontaine: Use `assert_match help_text, shell_output("#{bin}/mirror -h")`

mikemcquaid:
This repo isn't notable for us to include in Homebrew yet, I'm afraid. I'd suggest creating a tap instead.

jasonm23:
It's very useful, why would it need to be notable?

bfontaine:
@jasonm23 Experience shows that small projects that aren’t well known are more likely to be abandoned at some point. Adding a formula to the core means we’ll have to maintain it; we’ll get bug reports when it’ll break. A niche formula will also get less PRs.
Adding it [in a tap](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) doesn’t really affect the user experience and move the work into the tap maintainer’s hands instead of ours.

jasonm23:
Fair enough. I'll see that a tap is set up.

PR #43094, opened 2015-08-19
mono 4.0.3

natemcmaster:
Update the mono formula to pull the latest public release of mono.

http://www.mono-project.com/docs/about-mono/releases/4.0.3/

### DIFF
--- a/Library/Formula/mono.rb
+++ b/Library/Formula/mono.rb
@@ -1,8 +1,8 @@
 class Mono < Formula
   desc "Cross platform, open source .NET development framework"
   homepage "http://www.mono-project.com/"
-  url "http://download.mono-project.com/sources/mono/mono-4.0.2.5.tar.bz2"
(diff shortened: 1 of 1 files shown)


dunn:
The [build failures](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/31838/) aren't really your fault.  It's testing mono's dependencies to make sure there aren't breaking changes, and scriptcs is failing because we recently started running tests within a sandbox, and the scriptcs test is trying to write outside of the sandbox.  If you have the time to tweak that test, that'd be fantastic!

xu-cheng:
The failure is from a bug inside mono itself. For unknown reason, it doesn't respect `HOME` path overwrite. I filed a bug upstream: https://bugzilla.xamarin.com/show_bug.cgi?id=33151

natemcmaster:
Closing in favor of a mono 4.2.0 instead

PR #43095, opened 2015-08-19
Go 1.5 release

antonypp:


calebthompson:
Should this also update the various tools and other stuff still on 1.4.2?

andygrunwald:
This was released today. See https://blog.golang.org/go1.5
Would love to see this as soon as possible.

mistydemeo:
The code that deals with the bootstrap go needs to be adjusted to work with the stable release as well in this version.

mistydemeo:
@andygrunwald This was only submitted less than ten minutes ago - please be patient, thanks!

antonypp:
@calebthompson change

calebthompson:
We _probably_ don't want the `devel` block anymore as `1.5.0-rc1` has been superseded by the full 1.5 release?

In any case, this looks good.

DomT4:
We also need to look at the situation re [this comment](https://github.com/Homebrew/homebrew/pull/41305#issuecomment-122707705).

mistydemeo:
The `if build.head? || build.devel?` block in the install method needs to be made unconditional.

daMupfel:
:+1:

ahmetalpbalkan:
:+1:

mistydemeo:
We don't need additional +1s, please - thanks!

Library/Formula/go.rb
```diff
@@ -16,72 +16,39 @@ class Go < Formula
   end
 
   devel do
```
DomT4: You no longer require this block, at all.

Library/Formula/go.rb
```diff
@@ -16,72 +16,39 @@ class Go < Formula
...
+    (buildpath/"gobootstrap").install resource("gobootstrap")
+    cd "#{buildpath}/gobootstrap/src" do
+      system "./make.bash", "--no-clean"
     end
+    # This should happen after we build the test Go, just in case
+    # the bootstrap toolchain is aware of this variable too.
+    ENV["GOROOT_BOOTSTRAP"] = ENV["GOROOT_FINAL"]
 
     cd "src" do
       targets.each do |os, archs|
```
DomT4: You've removed all the logic for `targets` so `targets.each do |os, archs|` is pointing at nothing. You'll need to tweak this line for whatever is necessary now.

DomT4: Actually, this whole block from L54-L61 needs changing.

PR #43099, opened 2015-08-19
miller (new formula)

bokmann:
Miller is a great little unix slice-and-dice tool that lets you operate on csv data as name value pairs, with features equivalent to tools like sed and awk. 

http://johnkerl.org/miller/doc/index.html

When miller was featured recently on Hacker News, several people added and commented on a ticket to add the project into homebrew.

https://github.com/johnkerl/miller/issues/1

I appreciate the opportunity to contribute to homebrew, a tool I use virtually every day.

### DIFF
--- a/Library/Formula/miller.rb
+++ b/Library/Formula/miller.rb
@@ -1,0 +1,26 @@
+class Miller < Formula
+  desc "Miller is like sed, awk, cut, join, etc. for name-indexed data."
+  homepage "http://johnkerl.org/miller/doc/index.html"
+  url "https://github.com/johnkerl/miller/archive/v1.0.0.tar.gz"
+  sha256 "8e00d1e55609e2deb27537d385646ee4b3a4227070bdb1375ba7eff2a7e92708"
+
+  def install
+    # Upon improving miller's build system this souldn't be necessary.
+    # Will handle upstream.
+    ENV.deparallelize
+
+    system "make"
+    bin.mkpath
+
+    # workaround for hardcoded make file. Project is eventually moving
+    # to ./configure so this will no longer be necessary:
+    # https://github.com/johnkerl/miller/issues/9
+    system "HOME=\"#{prefix}\"; make install"
+  end
+
+  test do
+    # in lieu of any -v we could test a version from, lets just assert that
+    # there is available help output
+    assert_match(/abs/, pipe_output("#{bin}/mlr -f", ""))
+  end
+end


dunn:
Gonna have to shorten the description, I'm afraid:
```
==> audit problems
miller:
Length is calculated as miller + desc. (currently 87)
```

Library/Formula/miller.rb
```diff
@@ -0,0 +1,25 @@
+class Miller < Formula
+  desc "Miller is like sed, awk, cut, join, and sort for name-indexed data such as CSV."
+  homepage "http://johnkerl.org/miller/doc/index.html"
+  url "https://github.com/johnkerl/miller/archive/v1.0.0.tar.gz"
+  sha256 "8e00d1e55609e2deb27537d385646ee4b3a4227070bdb1375ba7eff2a7e92708"
+
+  depends_on "flex"
```
DomT4: Incapable of using the system version?

Library/Formula/miller.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def install
+    ENV.deparallelize
```
DomT4: Can we report this upstream if they aren't aware of it please? Ideally it shouldn't be necessary to deparallelize the build process.

Library/Formula/miller.rb
```diff
@@ -0,0 +1,23 @@
...
+    system "make"
+    # makefile hardcodes path to install the mlr executable as of 1.0.0.
+    # This is some creative wankery to get around that, in two steps:
+    # Step 1 - create the bin dir the makefile is expecting.
+    mkdir bin
```
mistydemeo: Do you think you could work with the developer to improve the makefile? Ideally we'd prefer not to keep weird workarounds like this longterm.

Library/Formula/miller.rb
```diff
@@ -0,0 +1,25 @@
...
+  homepage "http://johnkerl.org/miller/doc/index.html"
+  url "https://github.com/johnkerl/miller/archive/v1.0.0.tar.gz"
+  sha256 "8e00d1e55609e2deb27537d385646ee4b3a4227070bdb1375ba7eff2a7e92708"
+
+  depends_on "flex"
+
+  def install
+    ENV.deparallelize
+    system "make"
+    # makefile hardcodes path to install the mlr executable as of 1.0.0.
```
DomT4: This isn't that unusual. Can scrap these comments here.

Library/Formula/miller.rb
```diff
@@ -0,0 +1,25 @@
...
+    # This is some creative wankery to get around that, in two steps:
+    # Step 1 - create the bin dir the makefile is expecting.
+    mkdir bin
```
DomT4: `bin.mkpath` looks a bit neater.

Library/Formula/miller.rb
```diff
@@ -0,0 +1,25 @@
...
+    # Step 2 - temporarily manipulate the value of $HOME to the cellar prefix homebrew provides.
+    system "HOME=\"#{prefix}\"; make install"
```
DomT4: We can't mess with the `HOME` variable in-formula; Homebrew already captures and uses that during compile prevent installations messing involuntarily or deliberately with the user's home directory.

Library/Formula/miller.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  test do
+    # in lieu of any -v we could test a version from, lets just assert that there
```
DomT4: Can we do an actual functionality test? I notice upstream has some documentation/examples available here: https://github.com/johnkerl/miller/tree/master/doc

If those aren't automatically installed, we could install them manually and use one of those for testing purposes.

Library/Formula/miller.rb
DomT4: (My comment is a separate remark from Misty's; still ask upstream about improving the Makefile on this)

bokmann:
Comments addressed and PR squashed as per contribution instructions.  Appropriate issues are being tracked upstream; the existing build process wasn't quite mature enough.

Library/Formula/miller.rb
bokmann: I copied the testing strategy here from the tmux recipe, and i'm hoping to add a '-v' to miller to tighten this up.  I figured homebrew was more concerned about making sure the binary is in place and responding to commands, not verifying functionality that ideally should be tested in the build for the binary itself.

Let me know if this adequately answers the question or if you'd still like a different testing strategy.

dunn:
A bit of a nitpick from the bot:
```
==> audit problems
miller:
 * 11: Trailing whitespace was found
 * 14: Trailing whitespace was found
```

Library/Formula/miller.rb
```diff
@@ -0,0 +1,26 @@
...
+    # Will handle upstream.
+    ENV.deparallelize
+
+    system "make"
+    bin.mkpath
+
+    # workaround for hardcoded make file. Project is eventually moving
+    # to ./configure so this will no longer be necessary:
+    # https://github.com/johnkerl/miller/issues/9
+    system "HOME=\"#{prefix}\"; make install"
```
DomT4: I'm still uncomfortable with the manipulation of `HOME` here as a variable. As I said previously we actually use this variable during installations. Is there no other way?

Library/Formula/miller.rb
DomT4: `tmux` has a fairly vanilla test because there's not really much you can do with `tmux` that doesn't involve launching it, and throwing a new window is problematic in terms of testing.

Things can (and have) passed a `make check` stage just fine and then broken after installation for various reasons. The `test do` block exists to make sure that doesn't happen, and if one of the dependencies is upgraded that this can still build and run as expected. Ideally, we would use a functionality test everywhere, but vanilla tests should only be used when there's very little other option.

bokmann: ok. I'll look at the upstream examples and bake in a functionality test.

Library/Formula/miller.rb
johnkerl: make -C c -e INSTALLDIR=/usr/local/bin

mikemcquaid: Agreed. Would be good to submit a patch there.

Library/Formula/miller.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  test do
+    # in lieu of any -v we could test a version from, lets just assert that
+    # there is available help output
+    assert_match(/abs/, pipe_output("#{bin}/mlr -f", ""))
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikemcquaid: e.g. running on actual data

johnkerl:
https://github.com/johnkerl/miller/releases/tag/v1.0.1 has the INSTALLDIR option in `c/Makefile`

DomT4:
Thanks! If we can get this PR updated to 1.0.1 & escape having to tamper with the `HOME` variable, then add a functionality test, we should be nearly ready to get this in.

bokmann:
I'll get to that this weekend

johnkerl:
Awesome. LMK if there is more I need to do on my end.

DomT4:
@bokmann Any update on this?

kristoo:
@bokmann Need any help? Would like to see this merged in. :smiley:

johnkerl:
See also https://github.com/Homebrew/homebrew/pull/43658 ... just ran across it by accident.

johnkerl:
There is now https://github.com/johnkerl/miller/tree/v2.2.1 with autoconfig support, if that helps.

DomT4:
Things seem to have halted here in terms of progress, so I've opened #44307 to get this done. Thanks for the submission though; we appreciate the willingness to contribute!

PR #43103, opened 2015-08-20
supersonic: fix build

### LINKED ISSUE #43083: supersonic fails to build on OS X 10.10
Discovered this in #42960 (boost 1.59.0) but the failure doesn't look boost-related. Logs: https://gist.github.com/anonymous/8eeaabcd0ce25c1f8f55#file-01-configure-L127-L136

DomT4:
Supersonic expects to be able to pick up all of its dependencies via pkg-config which is something of a problem as `gflags` no longer builds or ships `.pc` files and consequently, detection fails.

Fixed by passing suggested ENVs instead.

Closes #43083.

### DIFF
--- a/Library/Formula/supersonic.rb
+++ b/Library/Formula/supersonic.rb
@@ -17,15 +17,20 @@ class Supersonic < Formula
     depends_on "protobuf"
     depends_on "boost"
   end
-  depends_on "glog"
+
   depends_on "pkg-config" => :build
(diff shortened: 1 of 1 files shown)


tdsmith:
:100:, good sleuthing

mikemcquaid:
:+1:

DomT4:
:boat:'ed.

PR #43107, opened 2015-08-20
Requirement: support tap formula as default_formula

xu-cheng:


mikemcquaid:
:+1: but, just out of interest, what's the tap formula you're wanting to use here?

xu-cheng:
> what's the tap formula you're wanting to use here?

I don't have one. I only submit this PR to support the things we are missing and pave the way to solve the dependency resolution problems in #42350

mikemcquaid:
ok, cool.

PR #43111, opened 2015-08-20
New formula: rl (randomize lines)

fresnik:


dunn:
A few minor things flagged by the audit:
```
==> audit problems
rl:
 * Stable: version 0.2.7 is redundant with version scanned from URL
Length is calculated as rl + desc. (currently 137)
```

chdiza:
The homepage for this software announces its own quasi-death, and explicitly tells users to use something else:

```
Another Warning: I have mostly lost interest in this project and am not planning on doing much development on it any more. I will however still accept patches and fix important bugs (this more or less contradicts the previous warning).

Note: Users are recommended to use shuf from GNU coreutils instead which has been included since coreutils 6.0.
```

mikemcquaid:
I think @chdiza's point is enough for me to :-1: on this. Thanks @chdiza for doing the research there.

DomT4:
Going to pass on this per the quasi-death. We'd only end up supporting it long-term more than upstream does, and we don't have the human-power to support that kind of thing. Thanks for the submission though!

chdiza:
@fresnik Don't worry, you can just use your formula locally, with `brew install path/to/rl.rb`.

PR #43117, opened 2015-08-20
fleetctl 0.11.4

madeddie:
I could add a simple --version test. 
Since this application is a utility to talk to a service, I'd have to write a mock service to test more of it, which might be overkill.

If you want the --version test, reply to this comment and I'll update the PR. If you want more extensive testing, could you point me to examples doing client -> server testing with a mock server?

### DIFF
--- a/Library/Formula/fleetctl.rb
+++ b/Library/Formula/fleetctl.rb
@@ -1,8 +1,8 @@
 class Fleetctl < Formula
   desc "Distributed init system"
   homepage "https://github.com/coreos/fleet"
-  url "https://github.com/coreos/fleet/archive/v0.11.2.tar.gz"
-  sha256 "85aa7c30b835b0145efe67b1cb4ecc3236d625a120a1b5c75b22feb3ca06f818"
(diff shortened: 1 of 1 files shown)


DomT4:
>  If you want more extensive testing, could you point me to examples doing client -> server testing with a mock server?

This would be great! There's an example of server testing [here](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mpd.rb#L127-L139), if that's useful.

madeddie:
Ah, sorry, I meant it the other way around, fleetctl is more like mpc. So it'd need a server to test against. There's some Go code in the package that's able to emulate a server, but compiling that just to test the client seems like overkill.

mikemcquaid:
@madeddie A `--version` test would be great.

DomT4:
Apologies for the misunderstanding! My misread on that. Checking the version is fine then, per Mike's comment.

madeddie:
Added simplest --version test.

mikemcquaid:

PR #43129, opened 2015-08-20
jenkins 1.625

asaph:


dunn:
The test failed because we're running tests in a sandbox now, and the test tries to write outside of the sandbox.  Not your fault, but would you mind seeing if you can tweak the test?

DomT4:
Looks like if we set `ENV["JENKINS_HOME"] = testpath` in `test do` it should work.

asaph:
@DomT4 Yep, that seems to have worked. Thanks!

xu-cheng:

PR #43130, opened 2015-08-20
libuv 1.7.1

### LINKED ISSUE #38138: libuv builds but does not pass all tests
libuv has a `make check` set of tests when you build it from http://github.com/libuv/libuv . I built and tested libuv on OS X Mavericks (10.9.5) and ran into this test error: https://github.com/libuv/libuv/issues/30 where `fs_event_watch_dir` fails.

There is a 'test' target in the homebrew system, but it doesn't do the exhaustive set of tests, so the homebrew test for libuv passes.

DomT4:
Closes #38138.

### DIFF
--- a/Library/Formula/libuv.rb
+++ b/Library/Formula/libuv.rb
@@ -1,8 +1,8 @@
 class Libuv < Formula
   desc "Multi-platform support library with a focus on asynchronous I/O"
   homepage "https://github.com/libuv/libuv"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

DomT4:
Was worried this would happen; the `make check` bit of `libuv` relies on pretty much perfect network connection at that point. Any slightly slow response leads to the test failing. I'll just make it optional, which is a bit of a shame, but more tolerable than the test only passing every so often.

PR #43139, opened 2015-08-20
mydumper: MySQL Data Dumper and Loader

onlynone:


Library/Formula/mydumper.rb
```diff
@@ -0,0 +1,24 @@
+class Mydumper < Formula
+  desc "How MySQL DBA & support engineer would imagine 'mysqldump' ;-)"
+  homepage "https://launchpad.net/mydumper"
+  url "https://launchpad.net/mydumper/0.6/0.6.2/+download/mydumper-0.6.2.tar.gz"
+  version "0.6.2"
```
DomT4: Don't need this line.

Library/Formula/mydumper.rb
```diff
@@ -0,0 +1,24 @@
...
+  sha256 "fa28563e8967752828954c5d81e26ef50aad9083d50a977bf5733833b23e3330"
+
+  depends_on "cmake" => :build
+  depends_on "sphinx" => :python
```
DomT4: Don't do this. You can use resources, which is a much nicer experience for the user.

Barring the unrelated comment on circular deps in [this block](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/cmake.rb#L19-L87), you can pretty much copy that directly, including the `option "without-docs", "Don't build man pages"` line above.

Library/Formula/mydumper.rb
```diff
@@ -0,0 +1,24 @@
...
+  depends_on "pkgconfig"
+  depends_on "openssl"
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  def test
+    system "mydumper" "--help"
```
DomT4: Should be `system bin/"mydumper", "--help"`.

If it's possible to write a functionality test, we heavily prefer those.

onlynone:
So, I can't do much in the way of tests since it requires a running mysql instance.

Not sure why the build failed, I just get a 404 when I click on Details.

dunn:
@BrewTestBot test this please

DomT4:
```
==> audit problems
mydumper:
 * Dependency 'pkgconfig' is an alias; use the canonical name 'pkg-config'.
 * Use new-style test definitions (test do)
 * mydumper seems to be listed in tap_migrations.rb!
Please remove mydumper from present tap & tap_migrations.rb
before submitting it to Homebrew/homebrew.
```
```
==> mydumper--help
Failed to execute: mydumper--help
Error: mydumper: failed
Failed executing: mydumper--help 
```

bfontaine:
Ping?

DomT4:
Closing as dead. Please feel free to resubmit as fresh with the comments above addressed, if you'd like to. Thanks for the submission to Homebrew!

PR #43140, opened 2015-08-20
nawk 20121220 (new formula)

rkitover:
This is the newest version available of Brian Kernighan's original nawk
awk interpreter. It is of interest to awk programmers testing their
programs on different versions of awk. It is installed as 'nawk' and the
man page as 'nawk.1' and so does not conflict with anything.

The github homepage promises to collect patches and track new versions.

### DIFF
--- a/Library/Formula/nawk.rb
+++ b/Library/Formula/nawk.rb
@@ -1,0 +1,31 @@
+class Nawk < Formula
+  desc "Brian Kernighan's pattern scanning and processing language"
+  homepage "http://www.cs.princeton.edu/~bwk/btl.mirror/"
+  url "http://www.cs.princeton.edu/~bwk/btl.mirror/awk.tar.gz"
+  version "20121220"
+  sha256 "8dc092165c5a4e1449f964286483d06d0dbfba4b0bd003cb5dab30de8f6d9b83"
+  head "https://github.com/junghans/nawk.git", :branch => "gentoo"
+
+  def install
+    # have to run bison first or parallel build fails
+    system "make", "YACC=bison -d -y", "ytab.o"
+    system "make"
+    bin.install "a.out" => "nawk"
+
+    # refer to 'nawk' instead of 'awk' in man page, and add a couple of
+    # SEE ALSO items
+    [
+      ["awk", "nawk"], ["AWK", "NAWK"], ["Awk", "Nawk"],
+      [/^.IR sed \(1\)$/, ".IR sed (1),\n.IR awk (1),\n.IR gawk (1)"]
+    ].each do |s|
+      inreplace "awk.1", *s
+    end
+
+    man1.install "awk.1" => "nawk.1"
+  end
+
+  test do
+    assert_match(/^Working!$/,
+      shell_output("#{bin}/nawk 'BEGIN { print \"Working!\"; exit(0); }'"))
+  end
+end


Library/Formula/nawk.rb
```diff
@@ -0,0 +1,22 @@
...
+  desc "Brian Kernighan's pattern scanning and processing language"
+  homepage "https://github.com/junghans/nawk"
+  url "https://github.com/junghans/nawk/archive/20121220.tar.gz"
+  sha256 "50b7a138242f0d2f5b63b72ddfc48d94e60edc495fc4b6ee26a25051b08e5127"
+  head "https://github.com/junghans/nawk.git"
+
+  def install
+    make_args = ["YACC=bison -d -y"]
+    # have to run bison first or parallel build fails
+    system "make", *make_args, "ytab.o"
```
mikemcquaid: You can't `*` other than the last argument on Ruby 1.8 (on Mountain Lion, which we support)

mikemcquaid:
The repo doesn't seem to be a very official upstream?

rkitover:
Hi Mike, thanks for looking at this.

I updated my commit:

- removed the star syntax
- set homepage and url to Brian's site

I cannot vouch for the github repo, it does not seem to have any patches in it although it promises to collect them. It seems to be the only repo for this. I actually sent them a PR yesterday but did not receive a response.

I left head pointing to that repo, but if you prefer I will remove it.

rkitover:
Apparently the owner of the github repo is doing dev in the gentoo branch and not master, but for now there are only a couple of minor makefile fixes there.

I updated the PR to set the gentoo branch as head.

rkitover:
Is this ok to be merged now?

Library/Formula/nawk.rb
```diff
@@ -0,0 +1,22 @@
...
+  sha256 "8dc092165c5a4e1449f964286483d06d0dbfba4b0bd003cb5dab30de8f6d9b83"
+  head "https://github.com/junghans/nawk.git", :branch => "gentoo"
+
+  def install
+    # have to run bison first or parallel build fails
+    system "make", "YACC=bison -d -y", "ytab.o"
+    system "make"
+    bin.install "a.out" => "nawk"
+    prefix.join("share/man/man1").mkpath
+    prefix.install "awk.1" => "share/man/man1/nawk.1"
```
xu-cheng: These two lines can be replaced as `man1.install "awk.1"`

rkitover:
Fixed, thanks!

rkitover:
Updated formula to fix up man page.

mikemcquaid:
Is this software actively maintained? If not we might need to pass. Sorry it took me so long to get around to saying this :sob:

rkitover:
The github user junghans maintains it for gentoo, the last release from Brian is from 2012, I don't know if he will make any more releases.

junghans agreed to make tags for his gentoo branch so I can use them in the formula, but there aren't really any major functionality changes there currently.

Beyond that, this program isn't really meant to be maintained as such, it is a reference implementation of awk, which is also /usr/xpg4/bin/awk on Solaris.

mikemcquaid:
For me I'm not sure given it's not really maintained sure but I'll let other maintainers give their thoughts.

mistydemeo:
We currently have a formula for this in homebrew/dupes as [awk](https://github.com/Homebrew/homebrew-dupes/blob/master/awk.rb).

mikemcquaid:
Yeh, I think that formula should take priority.

rkitover:
Honestly I like my formula better.

That formula conflicts with gawk, but the whole point of having multiple awks installed is to test on different ones, this awk is not the best version to use a daily driver.

Also you already have "mawk" in main, so I don't see why this would be different.

DomT4:
Going to pass here. It's essentially a dupe of the version in `dupes`, except with an unofficial `head` location that supports the project. If you want to support this project in a tap that's an option; there's some documentation on doing so [here](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md).

Thank you for the submission nonetheless! Appreciate the willingness to contribute to Homebrew.

PR #43142, opened 2015-08-20
smlnj: fix 10.11 build

### LINKED ISSUE #43100: smlnj failed to build on 10.11
==> config/install.sh
    config/install.sh: Installation directory is /usr/local/Cellar/smlnj/110.78/SMLNJ_HOME.
    config/install.sh: Installing version 110.78.
    config/install.sh: URL of source archive is http://smlnj.cs.uchicago.edu/dist/working/110.78/.
    config/install.sh: !!! Script /usr/local/Cellar/smlnj/110.78/SMLNJ_HOME/bin/.arch-n-opsys fails on this machine.
    config/install.sh: !!! You must patch this by hand and repeat the installation.

Logs at https://gist.github.com/427d45c6cca60c987861

DomT4:
Closes #43100.

### DIFF
--- a/Library/Formula/smlnj.rb
+++ b/Library/Formula/smlnj.rb
@@ -2,8 +2,8 @@ class Smlnj < Formula
   desc "Standard ML of New Jersey"
   homepage "http://www.smlnj.org/"
   url "http://smlnj.cs.uchicago.edu/dist/working/110.78/config.tgz"
-  sha256 "e2dd00b39b00ad892f182ce3f824d1540b0e350f2aee748ca971d44b5d340c05"
   version "110.78"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

chmaynard:
@DomT4 I'm new to this process. When will the fix show up in brew update?

FYI, Apple has a rule for Radar bug reports that I think works pretty well. A report (issue) is closed only after the originator is satisfied that the fix is successful.

DomT4:
> When will the fix show up in brew update?

Run `brew update` and it'll be there.

> A report (issue) is closed only after the originator is satisfied that the fix is successful.

Apple has the humanpower to keep reports open indefinitely in the vague hope the original reporter of an issue gets back to them with confirmation and chase when they don't. We really don't, sadly.

PR #43147, opened 2015-08-21
Update gnu-apl.rb

eccstartup:
This [issue](http://lists.gnu.org/archive/html/bug-apl/2015-03/msg00059.html) and `clang` support have been fixed in gnu svn rev 662 or earlier, but not in `apl-1.5.tar.gz`. So this `gnu-apl-1.5` will work with svn.

It is the first time that I make such pull request, and I have gone through [this tutorial](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-%28and-get-it-merged%29.md#how-to-open-a-homebrew-pull-request-and-get-it-merged).

I hope it is fine. Please help me fix the mistakes.

### DIFF
--- a/Library/Formula/gnu-apl.rb
+++ b/Library/Formula/gnu-apl.rb
@@ -1,9 +1,9 @@
 class GnuApl < Formula
+  version "1.5"
   desc "GNU implementation of the programming language APL"
   homepage "https://www.gnu.org/software/apl/"
-  url "http://ftpmirror.gnu.org/apl/apl-1.4.tar.gz"
-  mirror "https://ftp.gnu.org/gnu/apl/apl-1.4.tar.gz"
-  sha256 "69da31db180e6ae17116122758c266e19f62e256f04e3c4959f6f2b224a1893a"
+  url "svn://svn.sv.gnu.org/apl/trunk", :revision => "662"
+  head "svn://svn.sv.gnu.org/apl/trunk"
 
   bottle do
     revision 1
@@ -14,12 +14,11 @@ class GnuApl < Formula
   # GNU Readline is required; libedit won't work.
   depends_on "readline"
   depends_on :macos => :mavericks
+  depends_on "postgresql" => :recommended
 
   def install
-    system "./configure", "--disable-debug",
-                          "--disable-dependency-tracking",
-                          "--disable-silent-rules",
-                          "--prefix=#{prefix}"
+    system "./configure", "--prefix=#{prefix}"
+    system "make"
     system "make", "install"
   end
 


Library/Formula/gnu-apl.rb
```diff
@@ -1,9 +1,9 @@
 class GnuApl < Formula
+  version "1.5"
   desc "GNU implementation of the programming language APL"
   homepage "https://www.gnu.org/software/apl/"
-  url "http://ftpmirror.gnu.org/apl/apl-1.4.tar.gz"
-  mirror "https://ftp.gnu.org/gnu/apl/apl-1.4.tar.gz"
-  sha256 "69da31db180e6ae17116122758c266e19f62e256f04e3c4959f6f2b224a1893a"
+  url "svn://svn.sv.gnu.org/apl/trunk", :revision => "662"
```
mikemcquaid: We need a tagged version here and preferably a tarball.

Library/Formula/gnu-apl.rb
```diff
@@ -14,12 +14,11 @@ class GnuApl < Formula
   # GNU Readline is required; libedit won't work.
   depends_on "readline"
   depends_on :macos => :mavericks
+  depends_on "postgresql" => :recommended
```
mikemcquaid: Can you keep this at `:optional`? Thanks!

Library/Formula/gnu-apl.rb
```diff
@@ -14,12 +14,11 @@ class GnuApl < Formula
...
 
   def install
-    system "./configure", "--disable-debug",
-                          "--disable-dependency-tracking",
-                          "--disable-silent-rules",
-                          "--prefix=#{prefix}"
+    system "./configure", "--prefix=#{prefix}"
+    system "make"
```
mikemcquaid: Is `make install` not enough?

mikemcquaid:
Thanks for the PR!

Library/Formula/gnu-apl.rb
eccstartup: It is fine.

mikemcquaid: Cool, let's just leave it as that then, thanks!

Library/Formula/gnu-apl.rb
eccstartup: This may cause warnings. Some functions will be unavailable. So, I am afraid not.

mikemcquaid: Is that a 1.5-specific change or does it also apply to 1.4?

Library/Formula/gnu-apl.rb
eccstartup: I tried. The tagged version 1.5 won't work.

http://svn.savannah.gnu.org/viewvc/tags/apl-1.5/?root=apl

mikemcquaid: We'd want to use http://ftp.heanet.ie/mirrors/gnu/apl/apl-1.5.tar.gz. You can backport individual patches using `patch do` but we'll need to have an official tagged version rather than using a trunk revision.

Library/Formula/gnu-apl.rb
eccstartup: I think it also applies to 1.4 because it is probably required to use sql. I will change it to `:optional`.

Library/Formula/gnu-apl.rb
eccstartup: I have trouble writing `patch do` and finding out the patch files.

eccstartup: I created a diff file between `rev 579` and `rev 662`

`svn diff -r 579:662 svn://svn.sv.gnu.org/apl/trunk`
or
`svn diff svn://svn.sv.gnu.org/apl/tags/apl-1.5 svn://svn.sv.gnu.org/apl/trunk`

They just won't patch perfectly.

eccstartup:
Since I cannot find a perfect patch, I think I have to close it myself. Thanks @mikemcquaid

PR #43153, opened 2015-08-21
nifi 0.2.1 (new formula)

apiri:
Providing a new formula to support Apache NiFi release version 0.2.1.

### DIFF
--- a/Library/Formula/nifi.rb
+++ b/Library/Formula/nifi.rb
@@ -11,8 +11,7 @@ class Nifi < Formula
 
     ENV["NIFI_HOME"] = libexec
 
-    bin.install libexec/"bin/nifi.sh"
-    mv bin/"nifi.sh", bin/"nifi"
+    bin.install libexec/"bin/nifi.sh" => "nifi"
     bin.env_script_all_files libexec/"bin/", :NIFI_HOME => libexec
   end
 
--- a/Library/Formula/nifi.rb
+++ b/Library/Formula/nifi.rb
@@ -1,0 +1,25 @@
+class Nifi < Formula
+  desc "An easy to use, powerful, and reliable system to process and distribute data."
+  homepage "https://nifi.apache.org"
+  url "https://www.apache.org/dyn/closer.cgi?path=/nifi/0.2.1/nifi-0.2.1-bin.tar.gz"
+  sha256 "e151dab553a8ea466f7462d75145e2aa08ced938499aef184850aa4d3209c605"
+
+  depends_on :java => "1.7+"
+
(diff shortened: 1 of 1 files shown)


Library/Formula/nifi.rb
```diff
@@ -0,0 +1,25 @@
+class Nifi < Formula
+  desc "Easy to use, powerful, and reliable system to process and distribute data."
+  homepage "https://nifi.apache.org"
+  url "https://www.apache.org/dyn/closer.cgi?path=/nifi/0.2.1/nifi-0.2.1-bin.tar.gz"
+  sha256 "e151dab553a8ea466f7462d75145e2aa08ced938499aef184850aa4d3209c605"
+
+  depends_on :java => "1.7+"
+
+  # Wrap our nifi executable to point to the appropriate location within the brew environment
+  def script; <<-EOS.undent
```
DomT4: We can do this with something like [this](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/chapel.rb#L30-L31) rather than needing to give it its own `def` block here.

Library/Formula/nifi.rb
```diff
@@ -0,0 +1,25 @@
...
+    EOS
+  end
+
+  def install
+    libexec.install Dir["*"]
+    (bin/"nifi").write script
+  end
+
+  test do
+    system "nifi", "status"
```
DomT4: This should be `system bin/"nifi", "status"`

apiri:
Thanks for the feedback.  I reworked the areas you mentioned and have updated the associated branch.

Library/Formula/nifi.rb
```diff
@@ -0,0 +1,22 @@
...
+  sha256 "e151dab553a8ea466f7462d75145e2aa08ced938499aef184850aa4d3209c605"
+
+  depends_on :java => "1.7+"
+
+  def install
+    libexec.install Dir["*"]
+
+    ENV["NIFI_HOME"] = libexec
+
+    bin.install libexec/"bin/nifi.sh"
```
DomT4: Rather than doing the mv below to rename this you can just make this line:

```ruby
bin.install libexec/"bin/nifi.sh" => "nifi"
```

apiri:
@DomT4 Thanks for the additional tip on renaming my binary and your assistance throughout.  Made the associated update, tested, and pushed.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/c6b263a0a96273b2ab99253d1718db4d90481547. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #43158, opened 2015-08-21
document HOMEBREW_CC

tdsmith:


mistydemeo:
:+1:

mikemcquaid:
I dunno how I feel about documenting this. I don't know if I want to encourage people to set this in their profiles?

tdsmith:
I think I agree.

Issue #41501, opened 2015-07-09
Formulae that break under Sandboxing [Tracking]


DomT4:
At some point we probably want to turn Sandboxing on by default, but there's some formulae that die horrible deaths inside the Sandbox so tracking them here.

Ideally, we need to report failures upstream if it's something upstream can handle, but otherwise decide if to introduce a new variable like we do with `def pour_bottle?` that stops relevant from-source compiles happening when `HOMEBREW_SANDBOX` is active, or removing the formula from Homebrew, or patching it ourselves as necessary.

* <del>Git - #38978 (Seems to be potentially a 10.10 problem).</del>

* <del>Glib - https://gist.github.com/anonymous/2b4de19a79616a716ba0</del> Fixed by #42296

* <del>Postgis - #41962</del> Fixed by #45611

* <del>mecab-ipadic - https://github.com/Homebrew/homebrew/issues/41501#issuecomment-126634120</del> Fixed by #43324

mistydemeo:
With glib, the formula explicitly passes `--with-gio-module-dir=#{HOMEBREW_PREFIX}/lib/gio/modules` to configure; I believe the sandbox intentionally disallows writing to the prefix?

DomT4:
Yeah, that error is "on" us. `def install` permits writing to [these places](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/formula_installer.rb#L483-L491). You can only write to the broader Homebrew prefix in `postinstall`. Theoretically we could "whitelist" various directories for certain formula.

xu-cheng:
I just tested building git under sandbox in OS X El Captain VM. It works fine. No temporary files left.

mikemcquaid:
@DomT4 I think that error is actually a good thing as it makes us aware of when e.g. bottling will break the formula. Better to install the modules somewhere else and then move them in `post_install`, I think.

bfontaine:
`mecab-dict` fails in the sandbox, it tries to create directories under `mecab`’s prefix:

```
$ brew install --sandbox mecab-ipadic
==> Downloading https://downloads.sourceforge.net/project/mecab/mecab-ipadic/2.7.0-20070801/mecab-ipadic-2.7.0-20070801.tar.gz
Already downloaded: /Library/Caches/Homebrew/mecab-ipadic-2.7.0-20070801.tar.gz
[...]

PR #43157, opened 2015-08-21
tarsnap 1.0.36.1

dunn:

PR #43165, opened 2015-08-22
glib-networking: fix sandbox

DomT4:
Retains the exact presently existing behaviour, just in a way we can sandbox during installation and start bottling this formula. Everything still ends up in `HOMEBREW_PREFIX/lib/gio/modules`.

Ref: #41501

### DIFF
--- a/Library/Formula/glib-networking.rb
+++ b/Library/Formula/glib-networking.rb
@@ -5,12 +5,6 @@ class GlibNetworking < Formula
   mirror "https://mirrors.kernel.org/debian/pool/main/g/glib-networking/glib-networking_2.44.0.orig.tar.xz"
   sha256 "8f8a340d3ba99bfdef38b653da929652ea6640e27969d29f7ac51fbbe11a4346"
 
-  def pour_bottle?
-    # This formula installs files directly into the top-level gio modules
-    # directory, so it can't be bottled.
-    false
-  end
-
   depends_on "pkg-config" => :build
   depends_on "intltool" => :build
   depends_on "gettext"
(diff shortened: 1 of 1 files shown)


DomT4:
Oh, heh. This is getting killed by the:

```
==> brew install ./glib-networking-2.44.0.yosemite.bottle.tar.gz	 FAILED
Error: undefined method `skip_relocation?' for nil:NilClass
```

issue @dunn saw in #43157 as well.

xu-cheng:
I don't use these formulae. But I have some questions:

* Do we have to copy all these files? Why can't we install them into `lib`, then symlink them? If you have to copy these files, can you confirm which files are used in action?(since you are altering install path, I wonder do you change another hard code searching path in the process?)
* The cache file doesn't appear to be used in this formula. I think the right course is to delete it in install and regenerate in postinstall.

DomT4:
> Why can't we install them into lib, then symlink them?

I did it this way because if we switch to symlinks and then someone does `brew rm glib-networking` the entire `HOMEBREW_PREFIX/lib/gio` directory is removed, which doesn't seem to be desirable as `glib` and `glib-networking` both attempt to create it during installation and we create it `post_install`.

It seems desired to keep that directory intact, even after `glib-networking` is removed and that's the behaviour we've supported to date.

> (since you are altering install path, I wonder do you change another hard code searching path in the process?)

There's almost nothing about the `glib-networking` formula, heh. It's these set of libs, and some locales. There doesn't seem to be any paths to bork, as long as we put the stuff in the same place for other formulae/software to find it.

> The cache file doesn't appear to be used only in this formula. I think the right course is to delete it in install and regenerate in postinstall.

Sure, we can do that, but we still need to handle the rest of the objects it dumps in that prefix that would otherwise cause linking errors. The cache is only a wee part of the linking failure.

xu-cheng:
> I did it this way because if we switch to symlinks and then someone does brew rm glib-networking the entire HOMEBREW_PREFIX/lib/gio directory is removed, which doesn't seem to be desirable as glib and glib-networking both attempt to create it during installation and we create it post_install.

> It seems desired to keep that directory intact, even after glib-networking is removed and that's the behaviour we've supported to date.

This can be fixed by https://github.com/xu-cheng/homebrew/commit/5dcfbe44ed168890c7824ef36c25ddc7956d2d28

PR #43167, opened 2015-08-22
bottle: preserve :any_skip_relocation when creating new bottle

xu-cheng:


xu-cheng:
Or should we find a way to detect any_skip_relocation?

mistydemeo:
We can detect this by seeing if any dynamic links link to anything in `HOMEBREW_PREFIX`. If not (for example, because they only link to system libraries) they won't need relocation.

xu-cheng:
OK, will do more investigation on that approach.

PR #43088, opened 2015-08-19
arangodb 2.6.5

fceller:
Updated to ArangoDB 2.6.5

PR #43170, opened 2015-08-22
arangodb 2.6.5

fceller:
Updated to ArangoDB 2.6.5


DomT4:
Was merged in https://github.com/Homebrew/homebrew/pull/43088. If you're not seeing it yet it should be available after a `brew update`. Thanks for the submission though!

PR #43172, opened 2015-08-22
glib-networking: fix for sandbox

xu-cheng:
Closes #43165

cc @DomT4 This seems to be a better approach

### DIFF
--- a/Library/Formula/glib-networking.rb
+++ b/Library/Formula/glib-networking.rb
@@ -5,12 +5,6 @@ class GlibNetworking < Formula
   mirror "https://mirrors.kernel.org/debian/pool/main/g/glib-networking/glib-networking_2.44.0.orig.tar.xz"
   sha256 "8f8a340d3ba99bfdef38b653da929652ea6640e27969d29f7ac51fbbe11a4346"
 
-  def pour_bottle?
-    # This formula installs files directly into the top-level gio modules
-    # directory, so it can't be bottled.
-    false
-  end
-
   depends_on "pkg-config" => :build
   depends_on "intltool" => :build
   depends_on "gettext"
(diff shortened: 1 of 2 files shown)


DomT4:
This doesn't actually solve the problem though. The cache is only one tiny part of the linking problem:

```
Could not symlink lib/gio/modules/libgiognomeproxy.a
Target /usr/local/lib/gio/modules/libgiognomeproxy.a
already exists. You may want to remove it:
  rm '/usr/local/lib/gio/modules/libgiognomeproxy.a'

  brew link --overwrite glib-networking

  brew link --overwrite --dry-run glib-networking

/usr/local/lib/gio/modules/libgiognomeproxy.a
/usr/local/lib/gio/modules/libgiognomeproxy.so
/usr/local/lib/gio/modules/libgiognutls.a
/usr/local/lib/gio/modules/libgiognutls.so
```

xu-cheng:
I'm not too worried about that. In fact, I say it's a necessary transition to fix a previous broken formula. The reason I would prefer symlink instead of copy is you cannot do `brew (un)link/switch` if you just the copy the files. Even worse, you cannot even uninstall formula properly.

BTW, I hope we can do similar transition(i.e. copying files to symlink) for node.

DomT4:
> I'm not too worried about that. In fact, I say it's a necessary transition to fix a previous broken formula. 

Mike's comment in the other thread pretty heavily disagreed that it's desirable behaviour, and to be honest, I'm prone to agree. These sorts of error messages aren't very friendly UX to newer users of Homebrew.

> Even worse, you cannot even uninstall formula properly.

DomT4:
> I'm not too worried about that. In fact, I say it's a necessary transition to fix a previous broken formula. 

Mike's comment in the other thread pretty heavily disagreed that it's desirable behaviour, and to be honest, I'm prone to agree. These sorts of error messages aren't very friendly UX to newer users of Homebrew.

> Even worse, you cannot even uninstall formula properly.

As I said in the other thread, this has been the expected behaviour of `glib` and `glib-networking` to date, and both `glib` and `glib-networking` expect that directory to exist. Per Homebrew's logic, if we removed the symlinked contents from that directory, it abolishes the directory entirely. Even if we're fine to remove the contents, removing the directory seems to be unexpected in terms of `glib`'s default compile behaviour.

> BTW, I hope we can do similar transition(i.e. copying files to symlink) for node.

I'm pretty reluctant to screw with Node further, to be honest.

It's such a delicate ecosystem and we already hack up the expected build process to suit Homebrew pretty significantly. We've managed a decent run without getting reports about Homebrew's Node being borked every 5 minutes, and I'm wary about changing that unless there's some clear user benefit.

At some point I think making everything differ from user's expectations as dictated by upstream isn't actually a positive step, and ends up creating more work for us either explaining it to users, or fixing edge cases as they arise.

PR #43184, opened 2015-08-22
djbdns: add clt workaround

### LINKED ISSUE #42478: djbdns failed to build on 10.9, 10.11 without CLT
https://gist.github.com/c120570ffb3c8cf1f226

Formula `djbdns` failed when I was going through building everything on 10.11. Referencing nonexistent `/usr/include/errno.h`. It also failed with the same error when I tried to build it on a clean 10.9 machine without CLT installed. Looks like it's depending on the CLT for a header.

```
./compile dnscache-conf.c
In file included from <built-in>:329:
<command line>:1:10: fatal error: '/usr/include/errno.h' file not found
#include "/usr/include/errno.h"
         ^
1 error generated.
make: *** [dnscache-conf.o] Error 1
```

DomT4:
Rewrote most of the formula to avoid echoing stuff into files and just rewrite them ourselves, since calling the system on that sort of thing is pretty redundant & slow.

Closes #42478.

### DIFF
--- a/Library/Formula/daemontools.rb
+++ b/Library/Formula/daemontools.rb
@@ -10,4 +10,8 @@ class Daemontools < Formula
       bin.install Dir["command/*"]
     end
   end
+
+  test do
+    assert_match /Homebrew/, shell_output("#{bin}/softlimit -t 1 echo 'Homebrew'")
+  end
 end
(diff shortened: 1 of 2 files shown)


mikemcquaid:
:+1:

PR #43193, opened 2015-08-23
duck 4.8.18026

dunn:
The duck tarball is ~275MB so it's silly to use it for the test.

### DIFF
--- a/Library/Formula/duck.rb
+++ b/Library/Formula/duck.rb
@@ -1,8 +1,8 @@
 class Duck < Formula
   desc "Command-line interface for Cyberduck (a multi-protocol file transfer tool)"
   homepage "https://duck.sh/"
-  url "https://dist.duck.sh/duck-src-4.7.2.18007.tar.gz"
-  sha256 "2b30071ef58aed42805724f318b1e3f013749cc534e54265ee305dfc2ec86d69"
+  url "https://dist.duck.sh/duck-src-4.8.18026.tar.gz"
+  sha256 "ff75b2b0a4df30aef4f6cd2ba20fb3fa1198760fb7d8216abc62e032df4aa1e4"
   head "https://svn.cyberduck.io/trunk/"
 
   bottle do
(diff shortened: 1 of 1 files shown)


dkocher:
How does it come that you update the formula to some random snapshot build? Version 4.8 is *not* released.

dkocher:
Please revert the formula to 4.7.2. For users that want to run the latest snapshot build we provide [Cyberduck CLI Homebrew Tap for Snapshot Builds](https://github.com/iterate-ch/homebrew-cyberduck).

mikemcquaid:
@dkocher Please submit a pull request to do that and add a comment explaining how we can tell if a build is stable or a snapshot.

dkocher:
@mikemcquaid We publish the stable formula at https://version.cyberduck.io/duck.source.rb and snapshot formula at https://version.cyberduck.io/nightly/duck.source.rb if that is of any help.

mikemcquaid:
@dkocher Not really. We need a comment in the formula to verify how a release is stable or isn't. Most formulae are managed by the community so this is an unusual case. If you'd rather we can migrate this formula to a tap you control instead.

dkocher:
@mikemcquaid I would rather not want to spend time on this revert. We will push releases as usual to Homebrew using our CI which works without my manual intervention.

dkocher:
@mikemcquaid I don't see the problem or difference to other formulas.

mikemcquaid:
@dkocher If you want it reverted, you need to submit a pull request. If you don't want to spend time on it that's up to you but then it won't be reverted as quickly. We're a community run project and you are part of that community. If you wish to more tightly control the experience you should run a tap yourself.

dunn:
I took the latest tarball from your distribution page because there wasn't anything to indicate that some were development releases.  The section on manual installation on the wiki (https://trac.cyberduck.io/wiki/help/en/howto/cli#Manualinstallation) also simply points to https://dist.duck.sh.  It might help if users were first pointed to the changelog, and from there were able to download the sources.

dunn:
Reverted in 25db9abad0c9e2cc843be16b3fc40ff8c1245c3f.

PR #43197, opened 2015-08-23
Revert "mercurial: vendor pyopenssl"

tdsmith:
These modules are optionally used by some third-party Mercurial commands
but not by Mercurial core and the InsecurePlatformWarning people were
complaining about no longer appears on current OS versions.

This reverts commit 57873691a76cbe54be7bd981292871dfa5e229a9.

cc @DomT4; what do you think?


DomT4:
If you're happy that it's no longer necessary/desired, I've no objections at all.

mikemcquaid:
:+1: if it works

Issue #42210, opened 2015-07-28
aws-cloudsearch is broken


rgrinberg:
Running `$ cs-import-documents` will always give the following error:

```
Error loading configuration file lib/cs-clt.properties : File not found
```

We can workaround this by creating a symlink to where aws-cloudsearch expects this config file to be:

```
$ cd /usr/local/Cellar/aws-cloudsearch/2.0.1.0-2014.10.27/libexec
$ ln -s jars/lib lib
```

dunn:
That's actually something that we do in the formula: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/aws-cloudsearch.rb#L12

Just above that line we do an `inreplace` on the `cs-cmd` command to point it in the right direction, but we probably need to perform that substitution on `cs-import-documents` and perhaps other commands. Would you be interested in opening a pull request?

dunn:
OK, I think I know what the issue is.  Run `brew edit aws-cloudsearch` and try replacing
```ruby
    inreplace "bin/cs-cmd", 'LIBDIR="${CS_HOME}/lib"', 'LIBDIR="${CS_HOME}/jars/lib"'
```
with
```
    inreplace "bin/cs-cmd", 'LIBDIR="${CS_HOME}/lib"', 'LIBDIR="${CS_HOME}/libexec/jars/lib"'
```
Then save and `brew reinstall -fs aws-cloudsearch`.  Then see if it works.

dunn:
Hopefully fixed in 24a4f5c0c475e326744fe79206e2d66668ebf12f; could you upgrade aws-cloudsearch and let me know? Thanks!

PR #43198, opened 2015-08-23
aws-cloudsearch: remove lib path rewrite

dunn:
A little leery about this fix for #42210, since the `inreplace` has been in the formula since it was introduced in #11722; but there's no explanation for why and this seems to work.


mikemcquaid:
:+1:

PR #43201, opened 2015-08-23
add link_overwrite DSL 

xu-cheng:
cc @DomT4 @mikemcquaid @tdsmith

### DIFF
--- a/Library/Formula/glib-networking.rb
+++ b/Library/Formula/glib-networking.rb
@@ -5,12 +5,6 @@ class GlibNetworking < Formula
   mirror "https://mirrors.kernel.org/debian/pool/main/g/glib-networking/glib-networking_2.44.0.orig.tar.xz"
   sha256 "8f8a340d3ba99bfdef38b653da929652ea6640e27969d29f7ac51fbbe11a4346"
 
-  def pour_bottle?
-    # This formula installs files directly into the top-level gio modules
-    # directory, so it can't be bottled.
-    false
-  end
-
   depends_on "pkg-config" => :build
   depends_on "intltool" => :build
   depends_on "gettext"
@@ -19,11 +13,20 @@ class GlibNetworking < Formula
   depends_on "gsettings-desktop-schemas"
 
   def install
+    # Install files to `lib` instead of `HOMEBREW_PREFIX/lib`.
+    inreplace "configure", "$($PKG_CONFIG --variable giomoduledir gio-2.0)", lib/"gio/modules"
     system "./configure", "--disable-dependency-tracking",
                           "--disable-silent-rules",
                           "--prefix=#{prefix}",
                           "--with-ca-certificates=#{etc}/openssl/cert.pem"
     system "make", "install"
+
+    # Delete the cache, will regenerate it in post_install
+    rm lib/"gio/modules/giomodule.cache"
+  end
+
+  def post_install
+    system Formula["glib"].opt_bin/"gio-querymodules", HOMEBREW_PREFIX/"lib/gio/modules"
   end
 
   test do
--- a/Library/Formula/glib-networking.rb
+++ b/Library/Formula/glib-networking.rb
@@ -12,6 +12,8 @@ class GlibNetworking < Formula
   depends_on "gnutls"
   depends_on "gsettings-desktop-schemas"
 
+  link_overwrite "lib/gio/modules"
+
   def install
     # Install files to `lib` instead of `HOMEBREW_PREFIX/lib`.
     inreplace "configure", "$($PKG_CONFIG --variable giomoduledir gio-2.0)", lib/"gio/modules"
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -11,6 +11,7 @@ require "install_renamed"
 require "pkg_version"
 require "tap"
 require "formula_renames"
+require "keg"
 
 # A formula provides instructions and metadata for Homebrew to install a piece
 # of software. Every Homebrew formula is a {Formula}.
@@ -330,7 +331,6 @@ class Formula
   # The currently installed version for this formula. Will raise an exception
   # if the formula is not installed.
   def installed_version
-    require "keg"
     Keg.new(installed_prefix).version
   end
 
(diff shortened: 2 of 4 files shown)


Library/Formula/glib-networking.rb
```diff
@@ -18,12 +12,23 @@ def pour_bottle?
   depends_on "gnutls"
   depends_on "gsettings-desktop-schemas"
 
+  link_overwrite "lib/gio/modules/*"
```
mikemcquaid: Maybe `lib/gio/modules` is enough here? Dunno why but the wildcard feels weird.

Library/Homebrew/formula.rb
```diff
@@ -307,6 +308,12 @@ def any_version_installed?
   end
 
   # @private
+  # If any old version of {Formula} is installed.
```
mikemcquaid: Maybe just `any_version_installed?`

Library/Homebrew/formula.rb
```diff
@@ -657,6 +663,29 @@ def skip_clean?(path)
     self.class.skip_clean_paths.include? to_check
   end
 
+  # Sometimes we accidentally install files outside prefix. After we fix that,
+  # users will get nasty link conflict error. So we create a whitelist here to
+  # allow overwriteing certain files. e.g.
```
mikemcquaid: `overwriting`

Library/Homebrew/formula.rb
```diff
@@ -657,6 +663,29 @@ def skip_clean?(path)
...
+  #   link_overwrite "bin/foo", "lib/bar"
+  #   link_overwrite "share/man/man1/baz-*"
+  def link_overwrite?(path)
+    # Don't overwrite files for a fresh installation.
+    return false unless any_old_version_installed?
```
mikemcquaid: I don't agree with this check as it means a `brew uninstall; brew install` (or even `brew reinstall`) will fail.

Library/Homebrew/formula.rb
```diff
@@ -657,6 +663,29 @@ def skip_clean?(path)
...
+    # Don't overwrite files not created by Homebrew.
+    return false unless path.stat.uid == File.stat(HOMEBREW_BREW_FILE).uid
+    # Don't overwrite files belong to other keg.
+    begin
+      Keg.for(path)
+    rescue NotAKegError, Errno::ENOENT
+    else
```
mikemcquaid: Should this `else` be here?

Library/Homebrew/formula_installer.rb
```diff
@@ -589,9 +589,16 @@ def link(keg)
       keg.remove_linked_keg_record
     end
 
+    link_overwrite_backup = []
+
     begin
       keg.link
     rescue Keg::ConflictError => e
+      if formula.link_overwrite?(e.dst) && !link_overwrite_backup.include?(e.dst)
+        e.dst.rename "#{e.dst}.bak"
```
mikemcquaid: I don't think we need to be backing up files. It just creates more manual work for the user to cleanup and it'll happen every time you `brew upgrade glib-networking`.

Library/Homebrew/formula_installer.rb
```diff
@@ -616,10 +623,21 @@ def link(keg)
...
+      ignore_interrupts do
+        keg.unlink
+        link_overwrite_backup.each { |p| FileUtils.mv "#{p}.bak", p }
+      end
       Homebrew.failed = true
       raise
     end
+
+    unless link_overwrite_backup.empty?
+      opoo "These files are overwritten during `brew link` step:"
```
mikemcquaid: `are overwritten` -> `were overwritten`

Library/Homebrew/formula.rb
xu-cheng: `any_old_version_installed?` and `any_version_installed?` is two different methods.

Library/Homebrew/formula.rb
xu-cheng: `brew reinstall` will not fail. `brew uninstall; brew install` will. The idea is if we find conflict when we install this formula first time, we can pretty sure conflicted files are from outside of Homebrew. And I feel uncomfortable to nuke users' files.

Library/Homebrew/formula.rb
xu-cheng: Yes, it has to be here.

Library/Homebrew/formula.rb
mikemcquaid: Ok, I think I see what this is doing now but it's not super obvious for the name. Maybe `upgradable_version_installed?` or something?

mikemcquaid: Naming is hard :sob:

Library/Homebrew/formula_installer.rb
xu-cheng: > it'll happen every time you brew upgrade glib-networking

No it will only happen once.

Another reason that we want to backup is we need restore them during unexpected error https://github.com/xu-cheng/homebrew/blob/link/Library/Homebrew/formula_installer.rb#L626-L629

Library/Homebrew/formula.rb
mikemcquaid: `brew uninstall; brew install` failing is bad. The point of the whitelist is we only whitelist files we are happy with deleting. Do you think users are going to miss `gio` modules with identical names to those they have just requested be installed by Homebrew? If we really want that behaviour it should be opt-in.

Library/Homebrew/formula.rb
mikemcquaid: Ah, I see. Maybe add a comment under the `rescue`?

Library/Homebrew/formula.rb
xu-cheng: I think users may install some thrid-party software combo which contains `gio` modules and they are not aware of. If we just delete the conflict, that software may be broken.

But on the other hand, we can just replace them if you are happy with the backup logic.

mikemcquaid: I think we should definitely replace them and I don't think we really need backup logic either. If we do insist on adding the backup logic they should be put in another directory rather than just have an extension appended and be cleaned up by `brew cleanup`. Making the users manually delete them isn't nice (and may actually break some software that iterates over the contents of a directory for e.g. plugins).

xu-cheng: Ok, let me see if I can put backup files into Cache directory.

mikemcquaid: :+1:

mikemcquaid: Not to overcomplicate but could maybe even `.tar.gz` them for space usage.

xu-cheng:
Updated to address all the problems so far.

Library/Homebrew/formula.rb
```diff
@@ -657,6 +657,30 @@ def skip_clean?(path)
...
+      Keg.for(path)
+    rescue NotAKegError, Errno::ENOENT
+      # file doesn't belong to any keg.
+    else
+      return false
+    end
+    to_check = path.relative_path_from(HOMEBREW_PREFIX).to_s
+    self.class.link_overwrite_paths.any? do |p|
+      p == to_check ||
+      to_check.start_with?(p.chomp("/") + "/") ||
```
mikemcquaid: Might want to indent this line just to make it more obvious it's chained.

mikemcquaid:
We're probably all surprised to hear: I'm happy with this as-is. Backing up to `HOMEBREW_CACHE` seems to be a happy middle-ground between not backing up and cluttering up `HOMEBREW_PREFIX`. Nice work :+1:

mikemcquaid:
I think it would have been good to wait for @DomT4's review here (sorry, should have said earlier)

xu-cheng:
Sorry about that.

DomT4:
No worries. Pretty busy today so not around a huge amount.

Line 645 of `formula_installer.rb` should probably be `puts "They have been backed up in #{backup_dir}"` rather than `puts "They are backup in #{backup_dir}"`, but it's a nit. Mechanism looks fine though.

xu-cheng:
Thanks. Fixed the word in https://github.com/Homebrew/homebrew/commit/98dd2995e430aa29c313b80c4b7ccaa7b7efe077

PR #42960, opened 2015-08-14
boost: 1.59.0

dplarson:
This commit also removes the boost-python patch as the issue was fixed upstream (see https://github.com/boostorg/build/pull/78 for more info).

PR #43212, opened 2015-08-23
quantlib 1.6.1

tdsmith:
Provides Boost 1.59 compatibility (#42960)


dunn:
`--with-lispdir=#{share}/emacs/site-lisp/quantlib`

tdsmith:
if that's the usual autotools invocation should we add that to the audit message? I wouldn't have known what to do. Maybe it doesn't matter; I know now. :p

dunn:
I can add it, but it might then confuse people when they get the message on a non-autotools formula.

PR #43215, opened 2015-08-24
cairo: added support for xlib to --with-x11 switch

retrography:
Several formulas including "R" need cairo's xlib support to compile with x11 bindings. Now --with-x11 switch installs both xlib and xcb interfaces. Previously it only installed xcb.

### DIFF
--- a/Library/Formula/cairo.rb
+++ b/Library/Formula/cairo.rb
@@ -34,12 +34,14 @@ class Cairo < Formula
       --enable-gobject=yes
       --enable-svg=yes
       --enable-tee=yes
-      --enable-xlib=no
-      --enable-xlib-xrender=no
       --enable-quartz-image
     ]
 
-    args << (build.with?("x11") ? "--enable-xcb=yes" : "--enable-xcb=no")
+    if build.with? "x11"
+      args << "--enable-xcb=yes" << "--enable-xlib=yes" << "--enable-xlib-xrender=yes"
+    else
+      args << "--enable-xcb=no" << "--enable-xlib=no" << "--enable-xlib-xrender=no"
+    end
 
     system "./configure", *args
     system "make", "install"


Library/Formula/cairo.rb
```diff
@@ -34,12 +34,10 @@ def install
       --enable-gobject=yes
       --enable-svg=yes
       --enable-tee=yes
-      --enable-xlib=no
-      --enable-xlib-xrender=no
       --enable-quartz-image
     ]
 
-    args << (build.with?("x11") ? "--enable-xcb=yes" : "--enable-xcb=no")
+    args += (build.with?("x11") ? ["--enable-xcb=yes", "--enable-xlib=yes", "--enable-xlib-xrender=yes"] : ["--enable-xcb=no", "--enable-xlib=no", "--enable-xlib-xrender=no"])
```
mistydemeo: Let's expand this from a one-liner.

retrography: @mistydemeo what is the consensus? case or if?

mikemcquaid:
CC @tschoonj on this to make sure there won't be additional adverse effects

tschoonj:
I have no issue with this. Looks harmless to me.

Could you change the commit title though? It would be much clearer if it explicitly mentioned the Formula name (i.e. cairo). Thanks!

retrography:
@tschoonj done!

Library/Formula/cairo.rb
```diff
@@ -34,12 +34,22 @@ def install
       --enable-gobject=yes
       --enable-svg=yes
       --enable-tee=yes
-      --enable-xlib=no
-      --enable-xlib-xrender=no
       --enable-quartz-image
     ]
 
-    args << (build.with?("x11") ? "--enable-xcb=yes" : "--enable-xcb=no")
+    args += if build.with?("x11")
```
DomT4: To keep this consistent with everything else in Homebrew I'd rather use:

```ruby
if build.with? "x11"
  args << "--enable-xcb=yes" << "--enable-xlib=yes" << "--enable-xlib-xrender=yes"
else
  args << "--enable-xcb=no" << "--enable-xlib=no" << "--enable-xlib-xrender=no"
end
```

It doesn't make the formula any wider than the URL, it uses existing style & means we can squash the arrays. Win win win, IMO.

retrography: @DomT4 Agreed, and done.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/44be044ff4e60ffe225700d1afa4a6774285edf8. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #43219, opened 2015-08-24
unibilium 1.2.0

dholm:


Library/Formula/unibilium.rb
```diff
@@ -0,0 +1,30 @@
+class Unibilium < Formula
+  desc "Unibilium is a very basic terminfo library."
```
bfontaine: You can remove the name from the `desc`: `Very basic terminfo library`.

dholm:
@bfontaine Fixed!

Library/Formula/unibilium.rb
```diff
@@ -0,0 +1,30 @@
+class Unibilium < Formula
+  desc "A very basic terminfo library."
```
DomT4: `* Please remove the indefinite article "A" from the beginning of the description`

xu-cheng:

PR #43221, opened 2015-08-24
qbs 1.4.2

jakepetroules:
`CONFIG+=qbs_enable_java` is gone since this option no longer exists (the functionality is now on by default since the previous javac build-time dependency is no longer necessary).


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/9760e279f5dac4b7a15ddb731aac374d4653d071. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #43223, opened 2015-08-24
Tesseract: add with-training-tools and with-opencl options

ryanfb:


Library/Formula/tesseract.rb
```diff
@@ -24,10 +25,21 @@ class Tesseract < Formula
   end
 
   option "all-languages", "Install recognition data for all languages"
+  option "with-training-tools", "Install OCR training tools"
+  option "with-opencl", "Enable OpenCL support"
 
   depends_on "libtiff" => :recommended
   depends_on "leptonica"
 
+  if build.with? "training-tools"
```
DomT4: How widely used is this likely to be? I'm a little bleh about introducing so many optional dependencies here unless it's going to prove popular/widely useful/etc.

Library/Formula/tesseract.rb
```diff
@@ -17,6 +17,7 @@ class Tesseract < Formula
     depends_on "automake" => :build
     depends_on "libtool" => :build
     depends_on "pkg-config" => :build
+    depends_on "cairo" => :build
```
DomT4: Why is this necessary now, but not previously?

Library/Formula/tesseract.rb
```diff
@@ -58,8 +70,16 @@ def install
     ENV.cxx11
 
     system "./autogen.sh" if build.head?
-    system "./configure", "--disable-dependency-tracking", "--prefix=#{prefix}"
+    if build.with? "opencl"
```
DomT4: Just use an `args` array here, rather than duplicating the whole block.

```ruby
args = %W[
  --disable-dependency-tracking
  --prefix=#{prefix}
]

args << "--enable-opencl" if build.with? "opencl"

system "./configure", *args
```

Library/Formula/tesseract.rb
```diff
@@ -58,8 +70,16 @@ def install
...
+      system "./configure", "--disable-dependency-tracking", "--prefix=#{prefix}", "--enable-opencl"
+    else
+      system "./configure", "--disable-dependency-tracking", "--prefix=#{prefix}"
+    end
     system "make", "install"
+    if build.with? "training-tools"
+      system "make training"
```
DomT4: `system "make", "training"`

Library/Formula/tesseract.rb
```diff
@@ -58,8 +70,16 @@ def install
...
+      system "make training-install"
```
DomT4: `system "make", "training-install"`

tdsmith:
OpenCL is supported on 10.6+ so let's turn it on by default.

ryanfb:
@tdsmith While OpenCL is supported in OS X, I would say it's more experimental in Tesseract from my experience, so I'd rather sacrifice it as an option than turn it on by default.

Library/Formula/tesseract.rb
ryanfb: There appears to be a change in Tesseract HEAD somewhere along the way (since 3.04.00) that introduced a warning about Cairo during the default `./configure` step (I believe it's only used for training-tools though).

Library/Formula/tesseract.rb
ryanfb: If you want to use Tesseract with anything other than pre-built `.traineddata` etc. files, you need the training tools. I don't have a sense for the general need for that, but a lot of the discussion I see on [the Tesseract-OCR Google Group](https://groups.google.com/forum/#!forum/tesseract-ocr) is about custom OCR that needs training tools.

The other issue may be that due to the way some of the training tools use Pango/Cairo font rendering there seem to be a lot of issues with them on OS X vs. Linux. Part of why I'd like to have this as an option is so that people on OS X who *do* need to do OCR training can start to report those issues upstream in Tesseract and start getting them fixed.

Library/Formula/tesseract.rb
tdsmith: Cairo is unlikely to be a :build dependency; the shared libraries are probably needed once the tools are installed.

Library/Formula/tesseract.rb
ryanfb: Done and squashed/pushed.

Library/Formula/tesseract.rb
ryanfb: Done and squashed/pushed.

Library/Formula/tesseract.rb
ryanfb: Done and squashed/pushed.

Library/Formula/tesseract.rb
ryanfb: Removed and squashed/pushed - seems like the warning only occurs during a `./configure` issue with default clang + OpenMP on current Tesseract HEAD (see tesseract-ocr/tesseract#76).

Library/Formula/tesseract.rb
DomT4: I see. I'm not super-opposed to the option, I guess, it's just the rather heavy dependencies involved.

Sadly, the way `brew` handles `if build.with?` dependencies is less than ideal and consequently we've been trying to get rid of `if build.with?` outside of `def install`. Essentially the option looks like it doesn't require anything extra under `brew info tesseract`:

```
==> Dependencies
Required: leptonica ✘
Recommended: libtiff ✔
==> Options
--all-languages
	Install recognition data for all languages
--with-opencl
	Enable OpenCL support
--with-training-tools
	Install OCR training tools
```

But obviously, there's actually a bunch of "silent" dependencies there when the option is invoked. I'll let the other maintainers weigh in on this particular use case and see what the consensus is.

ryanfb: Ah, ok. It wasn't clear to me from [the Homebrew example formula](https://github.com/Homebrew/homebrew/blob/master/Library/Contributions/example-formula.rb) if there was a better way of doing `depends_on` for options.

mikemcquaid:
Let's leave this open for a while to see if people :+1:

callmewhy:
`--with-training-tools` is really necessary if you want to train your own traineddata. I look for this anwser for a long time...Hope it can be mergeed soon : )

mikemcquaid:

PR #43224, opened 2015-08-24
Fix Magit’s version reporting.

MerelyAPseudonym:
Without this change, the “version” is an empty string.


dunn:
Good catch! Merged in 756fd429a4a57c485aecb006f89554ac010687c4.

PR #43145, opened 2015-08-21
test-bot: tap TapDependency recursively

xu-cheng:
This should help to solve the problem in https://github.com/Homebrew/homebrew-science/pull/2654

cc @mikemcquaid @tdsmith

PR #43229, opened 2015-08-24
gst-plugins-bad: can optionally use opus, opencv

ploxiln:
gst-plugins-bad finds the pkgconfig files and works.

```
[pierce@plo-pro local]$ gst-inspect-1.0 opus
Plugin Details:
  Name                     opus
  Description              OPUS plugin library
  Filename                 /usr/local/lib/gstreamer-1.0/libgstopus.so
  Version                  1.4.5
  License                  LGPL
  Source module            gst-plugins-bad
  Source release date      2014-12-18
  Binary package           GStreamer Bad Plug-ins source release
  Origin URL               Unknown package origin

  opusenc: Opus audio encoder
  opusdec: Opus audio decoder
  opusparse: Opus audio parser
  rtpopusdepay: RTP Opus packet depayloader
  rtpopuspay: RTP Opus payloader

  5 features:
  +-- 5 elements

[pierce@plo-pro local]$ gst-inspect-1.0 opencv
Plugin Details:
  Name                     opencv
  Description              GStreamer OpenCV Plugins
  Filename                 /usr/local/lib/gstreamer-1.0/libgstopencv.so
  Version                  1.4.5
  License                  LGPL
  Source module            gst-plugins-bad
  Source release date      2014-12-18
  Binary package           GStreamer Bad Plug-ins source release
  Origin URL               Unknown package origin

  cvdilate: cvdilate
  cvequalizehist: cvequalizehist
  cverode: cverode
  cvlaplace: cvlaplace
  cvsmooth: cvsmooth
  cvsobel: cvsobel
  edgedetect: edgedetect
  faceblur: faceblur
  facedetect: facedetect
  motioncells: motioncells
  pyramidsegment: pyramidsegment
  templatematch: templatematch
  opencvtextoverlay: opencvtextoverlay
  handdetect: handdetect
  skindetect: skindetect
  retinex: Retinex image colour enhacement
  segmentation: Foreground/background video sequence segmentation
  grabcut: Grabcut-based image FG/BG segmentation
  disparity: Stereo image disparity (depth) map calculation

  19 features:
  +-- 19 elements

```

PR #43237, opened 2015-08-24
test-bot: don't consider optional deps for compiler selection

tdsmith:
Fixes test-bot mishandling of optional cross-tap dependencies. Without
this change, since formula.stable.deps includes the optional dependencies but
formula.recursive_dependencies does not, test-bot was trying to select a
compiler for an untapped formula, which raised an error.

Our suspect handling of optional dependencies was exposed by #43145 /
7184348e822d6745d8a103de104ff9626536bf1e.

### DIFF
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -410,7 +410,7 @@ module Homebrew
       end
 
       begin
-        formula.recursive_dependencies
+        deps.each { |d| d.to_formula.recursive_dependencies }
       rescue TapFormulaUnavailableError => e
         raise if e.tap.installed?
         safe_system "brew", "tap", e.tap.name
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -398,15 +398,15 @@ module Homebrew
       if formula.stable
         return unless satisfied_requirements?(formula, :stable)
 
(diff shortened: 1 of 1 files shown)


tdsmith:
casts summon on @xu-cheng

tdsmith:
related to build failure in #43229

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -398,15 +398,15 @@ def formula(formula_name)
       if formula.stable
         return unless satisfied_requirements?(formula, :stable)
 
-        deps |= formula.stable.deps.to_a
-        reqs |= formula.stable.requirements.to_a
+        deps |= formula.stable.deps.to_a.select { |d| !d.optional? }
```
xu-cheng: You can use `reject(&:optional?)`. :smile:

xu-cheng:
This reminds me that we should do recursive_dependencies multiple times for different specs : stable devel and HEAD

tdsmith:
thanks, will update tomorrow. i'm not sure i understand how the current code handles HEAD builds.

mikemcquaid:
> i'm not sure i understand how the current code handles HEAD builds.

Probably: it doesn't!

tdsmith:
I changed up the strategy a little, which should prevent the need for calling recursive_dependencies for different specs.

mikemcquaid:
:+1:

xu-cheng:
:+1: for this approach.

PR #43239, opened 2015-08-24
install: fix handling of "not migrated" message.

mikemcquaid:
CC @xu-cheng, thoughts? Still not totally sure I understand before so feel free to tear this to bits :grinning:

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -75,13 +75,7 @@ module Homebrew
 
         if f.installed?
           msg = "#{f.full_name}-#{f.installed_version} already installed"
-          unless f.linked_keg.symlink?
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Note that this was written on a plane and was thus hasn't been really tested yet.

xu-cheng:
Some thoughts:
* Maybe it's better to put logic outside formulae already installed check. They are independent/unrelated to each other. 
* I think at the best we can warning people they need to migrate rather than error out. The reason is they are high chance old name rack is another formula. And we shouldn't stop users to install formulae which happen to have that old name.

mikemcquaid:
@xu-cheng Decided to just revert the initial commit.

xu-cheng:
I think https://github.com/Homebrew/homebrew/pull/43269 is an acceptable solution.

mikemcquaid:
Cool, didn't see that yet. Closing!

PR #43240, opened 2015-08-24
pull: do fetch after everything else, sleep more.

mikemcquaid:
Use an exponential back-off on the sleep and try the fetch more times.

Hopefully this will mean that we spent less time waiting for Bintray.

CC @DomT4 for thoughts. This was written on a plane so hasn't been tested really yet.

### DIFF
--- a/Library/Homebrew/cmd/pull.rb
+++ b/Library/Homebrew/cmd/pull.rb
@@ -57,6 +57,8 @@ module Homebrew
       odie "You meant `git pull --rebase`."
     end
 
+    bintray_fetch_formulae =[]
+
     ARGV.named.each do |arg|
       if arg.to_i > 0
         url = "https://github.com/Homebrew/homebrew/pull/#{arg}"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Tested and it works so merging (without `--retry` which will fetch twice as often as we want)

DomT4:
Sorry, a little under the weather the last couple days so haven't been terribly quick on anything.

It would be nice to intercept and hide failure messages on the fetch, aside the last scheduled one. Not sure how easy it is to be that selective though, and only catch the initial failures.

That pondering aside this is great. Bintray doesn't seem to be getting any quicker on the publishing, but this is a better system than one short sleep followed by a longer one and bust mechanism & this is more forgiving of things like temporary network glitches.

mikemcquaid:
No worries @DomT4, get well soon. Yeh, not sure it's worth the effort to be selective here. Hopefully this is enough to make it a big quicker.

DomT4:
Thanks Mike :)

PR #43241, opened 2015-08-24
More API documentation.

mikemcquaid:
And remove the documented stuff from the `example-formula.rb`.

I won't merge the removal from `example-formula` but I'm more interested int the thoughts from others on whether this format will be able to make us kill `example-formula`.


bfontaine:
The example formula and the API docs don’t have the same goals IMHO. The latter is an exhaustive reference documentation while the former is useful when you want to quickly understand how to write a formula. In an ideal world we’d need both API docs *and* tutorials.

mikemcquaid:
@bfontaine I think from the length of `example-formula.rb` I wouldn't use the word "quickly" there :wink:. I personally think it'd be better to point to a few actual formulae for good examples in particular cases e.g. a C program, Swift program, Python program etc.

bfontaine:
@mikemcquaid Good point.

Library/Homebrew/formula.rb
```diff
@@ -537,19 +560,23 @@ def fish_completion
...
   end
 
+  # The directory where the formula's installation logs will be written.
+  # @private
   def logs
     HOMEBREW_LOGS+name
   end
 
-  # override this to provide a plist
+  # This method can be
```
DomT4: `This method can be`?

mikemcquaid: Correct, it can. :trollface: Will fix that!

DomT4:
People reference the `example-formula.rb` surprisingly heavily, but I agree it's a beast of a document that's difficult to keep up-to-date. In many ways I'd rather encourage a practice of looking at recently merged core Homebrew formula, which we do tend to be able to keep on top of in terms of style et al.

mikemcquaid:
So, the current state looks like this:

screen shot 2015-08-26 at 16 50 18

Any thoughts?

DomT4:
:+1:. Are we currently publishing API documentation somewhere?

bfontaine:
@DomT4 It’s available on [rubydoc.info](http://www.rubydoc.info/github/Homebrew/homebrew) but not published by us.

mikemcquaid:
@bfontaine I set that up, actually.

bfontaine:
Oh. I thought someone added Homebrew just because it’s a large/well known repo.

mikemcquaid:
`example-formula` has now been removed and interleaved into the API documentation (which we now link to more prominently). I'm more than happy to tweak this further and add/remove changes from here but just wanted to get it merged to avoid conflicts.

PR #43242, opened 2015-08-24
New option to disallow support for Java in the libbluray formula

Frizlab:
This is useful mainly in order to drop the need for `ant`, and thus `java`, when building.

### DIFF
--- a/Library/Formula/libbluray.rb
+++ b/Library/Formula/libbluray.rb
@@ -19,10 +19,12 @@ class Libbluray < Formula
     depends_on "libtool" => :build
   end
 
+  option "without-ant", "Disable Support for BD Java"
+
   depends_on "pkg-config" => :build
   depends_on "freetype" => :recommended
   depends_on "fontconfig"
-  depends_on "ant" => :build
+  depends_on "ant" => [:build, :optional]
 
   def install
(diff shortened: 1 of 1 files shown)


Library/Formula/libbluray.rb
```diff
@@ -19,10 +19,12 @@ class Libbluray < Formula
     depends_on "libtool" => :build
   end
 
+  option "without-bdjava", "Disable Support for BD Java"
+
   depends_on "pkg-config" => :build
   depends_on "freetype" => :recommended
   depends_on "fontconfig"
-  depends_on "ant" => :build
+  depends_on "ant" => :build if build.with? "bdjava"
```
mikemcquaid: `depends_on "ant" => [:build, :optional]`

Library/Formula/libbluray.rb
```diff
@@ -19,10 +19,12 @@ class Libbluray < Formula
     depends_on "libtool" => :build
   end
 
+  option "without-bdjava", "Disable Support for BD Java"
```
mikemcquaid: `option "without-ant", "Disable Support for BD Java"`

Frizlab:
Understood and pushed.

PR #43245, opened 2015-08-25
osquery 1.5.2

theopolis:


tdsmith:
Is osquery dropping support for openssl? libressl is a new dependency for most users.

theopolis:
osquery will drop the Apple-provided OpenSSL for either a brew provided OpenSSL or LibreSSL (eventually). It seems that bottle builds of OpenSSL still include AVX instructions so we opted for LibreSSL for the time being.

Building with LibreSSL is an opt-in using the ENV added `BUILD_USE_LIBRESSL`. This is only absolutely needed on 10.11, but it does add TLS1.2 capabilities to 10.9/10.10, overall a plus!

tdsmith:
> It seems that bottle builds of OpenSSL still include AVX instructions

This is news to me and sounds like a bug. I see what might be an allusion to this in #35343. Do you know if we're tracking this somewhere?

theopolis:
That's a bit older and was ultimately referencing the optimizations from lz4/rocksdb and libsnappy-- all of which have been resolved. OpenSSL still seems outstanding: https://github.com/facebook/osquery/pull/1425

I'm one step out of the loop here, but I am happy to help where ever needed. I can research what needs patching in OpenSSL's build/configure scripts to add the `-no-avx` flags, just let me know.

tdsmith:
Thanks; I'll open a ticket for this tomorrow. I looked at openssl's configure script and earned a headache for my troubles. superenv makes sure the compiler gets `-march=core2` for our bottle builds, which ISTM ought to (??) imply `-mno-avx`, but openssl contains a lot of assembly and it's not immediately obvious to me what's trying to detect what and whether that happens at build time vs. runtime.

tdsmith:
(openblas is an example of a library which produces a single .dylib that dynamically switches on different arch-specific optimizations depending on the system where it finds itself running, which is what makes me think that the presence of those opcodes might not be strictly incriminating)

tdsmith:
Meant to add: any chance there's a virtualization issue interfering with feature detection, or...??

mikemcquaid:
@theopolis Is this when using `--build-bottle` or using the binary package we provide? Can you reproduce any failures with it? Is there any chance it's doing e.g. runtime feature detection? If OpenSSL was broken for people on older hardware we'd definitely have heard about it by now; I'd really rather we left this as `depends_on "openssl"` unless we have a reproducible issue here.

theopolis:
Take a look at our issue, which discusses OpenSSL and the AVX instructions present within binary distribution and as a result of a local bottle build.

AFAIK the presence of AVX opcodes in OpenSSL's shared libraries have not generated any SIGILLs in practice. It's very possible their library, or any similar library, may perform runtime checks and decide to take a non-AVX-use branch. But that responsibility and capability is on the shared library, it would almost never be the library user whom could make those runtime decisions.

Also, the AVX instructions are just one reason we're linking Libre. We have a responsible developer on their announce list and ready to account for any bugs/vulnerabilities affecting downstream usage. There are several personal and feel-good reasons the project has discussed, which has led us to the Libre choice. 

From a package manager perspective, it's good to ask these questions! But shouldn't we ease pressure from developers using compatible/reasonable OpenSSL alternatives (OpenSSL being a variable in this meaning)? Requiring users to download a new dependency sounds like a pill Apple already forces us to swallow.

mikemcquaid:
> From a package manager perspective, it's good to ask these questions! But shouldn't we ease pressure from developers using compatible/reasonable OpenSSL alternatives (OpenSSL being a variable in this meaning)? Requiring users to download a new dependency sounds like a pill Apple already forces us to swallow.

If you as `osquery` wish to not support OpenSSL (or at least recommend LibreSSL instead) we're happy to follow your recommendations. I just wanted to verify that this change isn't Homebrew-specific based on an unconfirmed (and possibly non-existing) bug in our OpenSSL package. Hope that makes sense, thanks!

theopolis:
Yay! Give me a few hours this morning and I'll get back to you on the OpenSSL+AVX practical concerns issue. I will do a quick review and see if there's a real concern. 

Either way, we'd like to stick with Libre. :)

tdsmith:
Works for me; let us know. Pulled in 21f97b3a7fb86b861f03b9a102ce97983b7df0ba. Thanks!

PR #43253, opened 2015-08-25
sqoop 1.4.6

stuwil:


Library/Formula/sqoop.rb
```diff
@@ -34,4 +34,10 @@ def caveats; <<-EOS.undent
     for Sqoop to work.
     EOS
   end
+
+  test do
+    ENV["JAVA_HOME"] = "$(/usr/libexec/java_home)"
```
bfontaine: This won’t work, use `/usr/libexec/java_home`.chomp instead.

Library/Formula/sqoop.rb
```diff
@@ -34,4 +34,9 @@ def caveats; <<-EOS.undent
     for Sqoop to work.
     EOS
   end
+
+  test do
+    ENV["JAVA_HOME"] = `/usr/libexec/java_home`.chomp
```
xu-cheng: We can set JAVA_HOME for users in install:
`bin.env_script_all_files(libexec/"bin", Language::Java.java_home_env)`

stuwil:
@bfontaine Would you like me to add your changes from #43262?

bfontaine:
@stuwil Yes I think they’ll be necessary. Thanks!

Library/Formula/sqoop.rb
```diff
@@ -22,7 +24,17 @@ def spoop_envs
   def install
     libexec.install %w[bin conf lib]
     libexec.install Dir["*.jar"]
+
+    # use coreutils' readlink instead of the OS X one to get the '-f' option
```
mikemcquaid: Can you submit an issue upstream about this? Would be nice if it wasn't needed.

mikemcquaid:
Test is failing (presumably it's trying to write to `~`). Any ideas? http://bot.brew.sh/job/Homebrew%20Pull%20Requests/32163/version=yosemite/testReport/junit/brew-test-bot/yosemite/test_sqoop/

Library/Formula/sqoop.rb
```diff
@@ -1,14 +1,16 @@
...
   desc "Transfer bulk data between Hadoop and structured datastores"
   homepage "https://sqoop.apache.org/"
-  url "https://www.apache.org/dyn/closer.cgi?path=sqoop/1.4.5/sqoop-1.4.5.bin__hadoop-2.0.4-alpha.tar.gz"
-  version "1.4.5"
-  sha256 "2f36ba52ae64f2f674780984aa4ed53d43565098f208a4fcbd800af664b1def9"
+  url "https://www.apache.org/dyn/closer.cgi?path=sqoop/1.4.6/sqoop-1.4.6.bin__hadoop-2.0.4-alpha.tar.gz"
+  version "1.4.6"
+  sha256 "d582e7968c24ff040365ec49764531cb76dfa22c38add5f57a16a57e70d5d496"
 
+  depends_on :java
```
DomT4: Any specific version requirement here or?

stuwil:
Working on a patch for upstream...

stuwil:
Submitted a patch upstream. https://issues.apache.org/jira/browse/SQOOP-2531

Library/Formula/sqoop.rb
stuwil: 1.6-1.7 it seems, it gets executed via Hadoop. Note 1.8 isn't supported - any way to exclude or specify a range?

mikemcquaid:
@stuwil Could you add a link to that in the formula and apply the patch? Thanks!

stuwil:
@mikemcquaid Added, but note we're using Java binaries here - the patch I submitted upstream is for the build script that generates the affected scripts. I've had to apply the same change to a number of different files, but please let me know if there's a better approach.

Haven't looked into it too much, but their ant build process isn't playing nicely with OS X...

mikemcquaid:

PR #43258, opened 2015-08-25
packer 0.8.6

stuwil:


Library/Formula/packer.rb
```diff
@@ -199,6 +199,7 @@ def install
     packerpath = buildpath/"src/github.com/mitchellh/packer"
     packerpath.install Dir["{*,.git}"]
     Language::Go.stage_deps resources, buildpath/"src"
+    mkdir_p buildpath/"bin"
```
DomT4: Why is this necessary now?

dunn:
Merged in 44fa0937e6a269d8c6420442ed378d2e2ca41d68, thank you!

Library/Formula/packer.rb
dunn: Sorry, somehow missed this.

DomT4: No worries, not a big deal at all, was just curious.

stuwil: Didn't spend too much time trying to track it down, but the upstream build script was trying to copy to GOPATH/bin, which didn't exist in the buildpath... not sure when that change came in.

PR #43259, opened 2015-08-25
bash completion for tap-pin/tap-unpin added

bfontaine:
Not tested for now, which is why I made a PR instead of directly committing.


xu-cheng:
:+1:

PR #43262, opened 2015-08-25
sqoop 1.4.6

bfontaine:
I added a `coreutils` dependency because scripts use `readlink`’s `-f` option, not available on OS X. This issue was already present with the previous version but we didn’t see it because of the lack of a test.


bfontaine:
Oops, didn’t see #43253.

PR #43269, opened 2015-08-25
outdated/install fixes

vladshablinsky:
Fix https://github.com/Homebrew/homebrew/pull/42940 from https://github.com/Homebrew/homebrew/issues/42965

cc @mikemcquaid @xu-cheng

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -75,14 +75,15 @@ module Homebrew
 
         if f.installed?
           msg = "#{f.full_name}-#{f.installed_version} already installed"
-          unless f.linked_keg.symlink?
-            if f.oldname && (HOMEBREW_CELLAR/f.oldname).exist?
-              msg << ", it's just not migrated"
-            elsif !f.keg_only?
-              msg << ", it's just not linked"
-            end
-          end
+          msg << ", it's just not linked" unless f.linked_keg.symlink? || f.keg_only?
           opoo msg
+        elsif f.oldname && (dir = HOMEBREW_CELLAR/f.oldname).exist? && !dir.subdirs.empty? \
+            && f.tap == Tab.for_keg(dir.subdirs.first).tap && !ARGV.force?
+          # Check if the formula we try to install is the same as installed
+          # but not migrated one. If --force passed then install anyway.
+          opoo "#{f.oldname} already installed, it's just not migrated"
+          puts "You can migrate formula with `brew migrate #{f}`"
+          puts "Or you can force install it with `brew install #{f} --force`"
         else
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cmd/install.rb
```diff
@@ -75,14 +75,10 @@ def install
...
-          unless f.linked_keg.symlink?
-            if f.oldname && (HOMEBREW_CELLAR/f.oldname).exist?
-              msg << ", it's just not migrated"
-            elsif !f.keg_only?
-              msg << ", it's just not linked"
-            end
-          end
+          msg << ", it's just not linked" unless f.linked_keg.symlink? || f.keg_only?
           opoo msg
+        elsif f.oldname && (dir = HOMEBREW_CELLAR/f.oldname).exist? && !dir.subdirs.empty?
```
xu-cheng: We need to handle the case that oldname is another formula. To do that, I suggest:
* check tap origin 
* allow --force to bypass the migration message.

Library/Homebrew/cmd/install.rb
```diff
@@ -75,14 +75,13 @@ def install
...
-              msg << ", it's just not linked"
-            end
-          end
+          msg << ", it's just not linked" unless f.linked_keg.symlink? || f.keg_only?
           opoo msg
+        elsif f.oldname && (dir = HOMEBREW_CELLAR/f.oldname).exist? && !dir.subdirs.empty? \
+            && f.tap == Tab.for_keg(dir.subdirs.first).tap && !ARGV.force?
+          # Check if the formula we try to install is the same as installed
+          # but not migrated one. If --force passed then install anyway.
+          opoo "#{f.oldname} already installed, it's just not migrated"
```
xu-cheng: I think we should mention --force in the message.

vladshablinsky: Added
```ruby
puts "You can force install it with `brew install #{f} --force`"
```

mikemcquaid:
:+1: from me. @xu-cheng?

Library/Homebrew/cmd/install.rb
xu-cheng: Maybe it's better to have:
```ruby
puts "You can migrate formula with `brew migrate #{f}`"
puts "Or you can force install it with `brew install #{f} --force`"
```

xu-cheng:
Need some word tweak, otherwise :+1: from me.

Library/Homebrew/cmd/install.rb
vladshablinsky: :+1:

mikemcquaid:
Thanks @vladshablinsky!

PR #43232, opened 2015-08-24
SwiftLint 0.1.2

jpsim:
Thanks!

PR #43162, opened 2015-08-21
Update xctool to 0.2.5.

ExtremeMan:
https://github.com/facebook/xctool/releases/tag/v0.2.5

PR #43276, opened 2015-08-25
sandbox: permit /var/tmp & DerivedData

DomT4:
Long term it would be nice to sandbox everything that writes to DerivedData but it is essentially a cache directory of sorts.

The downside of allowing stuff to write there particularly is that DerivedData is notoriously bad at getting cleaned up, so if you do a lot of Xcode-using installations very quickly, you can chew your disk space up.

### DIFF
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -57,6 +57,7 @@ class Sandbox
 
   def allow_write_temp_and_cache
     allow_write_path "/private/tmp"
+    allow_write_path "/private/var/tmp"
     allow_write "^/private/var/folders/[^/]+/[^/]+/[C,T]/", :type => :regex
     allow_write_path HOMEBREW_TEMP
     allow_write_path HOMEBREW_CACHE
@@ -66,6 +67,11 @@ class Sandbox
     allow_write_path formula.rack
     allow_write_path formula.etc
     allow_write_path formula.var
+  end
+
+  # Xcode projects expect access to certain cache/archive dirs.
+  def allow_write_xcode
+    allow_write_path "/Users/#{ENV["USER"]}/Library/Developer/Xcode/DerivedData/"
   end
 
   def allow_write_log(formula)
--- a/Library/Homebrew/cmd/postinstall.rb
+++ b/Library/Homebrew/cmd/postinstall.rb
@@ -31,6 +31,7 @@ module Homebrew
         sandbox.allow_write_temp_and_cache
         sandbox.allow_write_log(formula)
         sandbox.allow_write_cellar(formula)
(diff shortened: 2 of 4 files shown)


DomT4:
CC @mikemcquaid @xu-cheng

Library/Homebrew/sandbox.rb
```diff
@@ -68,6 +69,11 @@ def allow_write_cellar(formula)
     allow_write_path formula.var
   end
 
+  # Xcode projects expect access to certain cache/archive dirs.
+  def allow_write_xcode
+    allow_write_path "/Users/#{ENV["USER"]}/Library/Developer/Xcode/DerivedData/"
```
xu-cheng: Why do we need this? Shouldn't HOME var overwrite cover this?

DomT4: Apparently, no. The `HOME` overwrite is getting ignored by every Xcode project we've ran up against in the last few days. See #43232 and #43162.

xu-cheng: :cool:

mikemcquaid: :+1:

DomT4: I think the pointer as to where the Xcode directories are is provided by Xcode itself sadly. It looks possible to [overwrite](https://developer.apple.com/library/ios/recipes/xcode_help-locations_preferences/DerivedData/DerivedData.html) either in Xcode or via the command-line, but whether or not we want to enforce that on every project right away is probably premature.

mikemcquaid: Given we have an `xcodebuild` helper it'd be cool if we could integrate using that?

DomT4: Oh, that's a neat idea. It wouldn't catch everything but it'd reduce an awful lot of the formulae that do wander. Will check it out.

DomT4: Ah, problem:

```ruby
  def xcodebuild(*args)
    removed = ENV.remove_cc_etc
    system "xcodebuild", *args, "-derivedDataPath", "#{buildpath}/.brew_home/DerivedData"
  ensure
    ENV.update(removed)
  end
```

```
xcodebuild: error: The flag -scheme is required when specifying -derivedDataPath.
```

There are only two formula which use the `-scheme` flag, and that flag isn't applicable to every project.

mikemcquaid: No worries, let's just permit it

DomT4: Maybe worth an audit check or something, if `scheme` is set they should set `derivedDataPath` as well. Don't know.

PR #43285, opened 2015-08-26
gdal 1.11.2_4

smithjlrhit:
Install gdal.jar along with the native JNI bindings.

Native JNI bindings are typically very specific to the java wrapper.  This installs the associated
jar file to be used with the native JNI shared objects.

### DIFF
--- a/Library/Formula/jenkins-job-builder.rb
+++ b/Library/Formula/jenkins-job-builder.rb
@@ -6,9 +6,9 @@ class JenkinsJobBuilder < Formula
 
   bottle do
     cellar :any
-    sha256 "30bee8cc99c0a246d403439c44e67ff1a223d1ac837163dcdc487b30e6429be5" => :yosemite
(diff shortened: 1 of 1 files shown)


Library/Formula/gdal.rb
```diff
@@ -3,7 +3,7 @@ class Gdal < Formula
   homepage "http://www.gdal.org/"
   url "http://download.osgeo.org/gdal/1.11.2/gdal-1.11.2.tar.gz"
   sha256 "66bc8192d24e314a66ed69285186d46e6999beb44fc97eeb9c76d82a117c0845"
-  revision 3
+  revision 4
```
DomT4: Given the change you're making is optional, let's not bump the `revision`.

smithjlrhit: No problem.

smithjlrhit: Sorry I didn't mean to close this.

mikemcquaid: :+1:

PR #43257, opened 2015-08-25
vim 7.4.826

guoxiao:

PR #43286, opened 2015-08-26
vim 7.4.838

saketkc:


guoxiao:
The comment says 
`# This package tracks debian-unstable: https://packages.debian.org/unstable/vim`. 
So in my #43257 I picked the v7.4.826 used by debian.

saketkc:
Indeed @guoxiao. Sorry, I missed that!
But it looks like the previous versions aren't maintained there.
For example `7.4.712`(https://github.com/saketkc/homebrew/commit/f50c7a6bd7b2dad3bd250065676db3bddc2814cb) seems to `404`.

Is there a motivation to follow debian-unstable?

bfontaine:
We probably won’t want to update the formula for each patch as there are too many of them (e.g. 9 releases yesterday); see [`youtube-dl`](https://github.com/Homebrew/homebrew/blob/b40dc4bbc0364ed0a54678766ab180204b8f674c/Library/Formula/youtube-dl.rb#L1-L3) as an example. cc @mikemcquaid for an opinion on this.

DomT4:
> Is there a motivation to follow debian-unstable?

It's stable in terms of release schedule, and means we don't have to update Vim on a daily basis which when there's no bottle available becomes tedious rather quickly. We do a similar thing with `youtube-dl` which publishes very frequently - We only pick up the releases which make it onto `pypi`.

Prior to very recently when Vim decided to move over permanently, the Github repo had a great big warning attached to it that it could vanish at any moment.

mikemcquaid:
I agree following Debian unstable is a good call.

DomT4:
Alright. Closing this out. We're going to stick to tracking Debian unstable for the time being. Thanks for the submission though!

nelstrom:
I've just submitted a PR to [upgrade to Vim 7.4.861](https://github.com/Homebrew/homebrew/pull/43753). I thought it would be a good idea to use the `.tar.gz` hosted by GitHub and I've been pointed towards this discussion here.

I'd like to see more frequent updates to Vim in homebrew. Updating the formula every time a Vim patch is released would be too frequent, but getting a new release every week would be nice. (If you look at the activity on [github/vim](https://github.com/vim/vim/commits/master) you can tell which days Bram is doing his 20% time on Vim!). If I want to make this happen, would the best way to do it be to make a homebrew tap and customise the Vim formula there?

PR #43298, opened 2015-08-26
Warn re OS X curl on <10.7

### LINKED ISSUE #43283: glib/glibmm fails to download/build on 10.6.8
"brew doctor" is empty, no errors/warnings.

```
$ brew upgrade glibmm
==> Upgrading 1 outdated package, with result:
glibmm 2.44.1
==> Upgrading glibmm
==> Installing dependencies for glibmm: glib
==> Installing glibmm dependency: glib
==> Downloading https://download.gnome.org/sources/glib/2.44/glib-2.44.1.tar.xz
curl: (35) error:14077458:SSL routines:SSL23_GET_SERVER_HELLO:reason(1112)
Error: Failed to download resource "glib"
Download failed: https://download.gnome.org/sources/glib/2.44/glib-2.44.1.tar.xz

$ curl --version
curl 7.19.7 (universal-apple-darwin10.0) libcurl/7.19.7 OpenSSL/0.9.8y zlib/1.2.3
Protocols: tftp ftp telnet dict ldap http file https ftps 
Features: GSS-Negotiate IPv6 Largefile NTLM SSL libz 

```

DomT4:
The `curl` on <10.7 is more or less incapable of fetching any recently-changed GNOME project (i.e. the ones we've moved across to `https`) and various other places. We've been softly using our curl on 10.6 and below for a while, but never communicating that en-mass to users.

Closes #43283.

### DIFF
--- a/Library/Homebrew/cmd/doctor.rb
+++ b/Library/Homebrew/cmd/doctor.rb
@@ -653,13 +653,23 @@ class Checks
     end
   end
 
+  def check_for_bad_curl
+    if MacOS.version <= "10.6" && !Formula["curl"].installed? then <<-EOS.undent
+      The system curl on 10.6 and below is often incapable of supporting
+      modern secure connections & will fail on fetching formulae.
+      We recommend you:
+        brew install curl
+      EOS
+    end
+  end
+
   def check_user_curlrc
     if %w[CURL_HOME HOME].any? { |key| ENV[key] && File.exist?("#{ENV[key]}/.curlrc") } then <<-EOS.undent
     You have a curlrc file
     If you have trouble downloading packages with Homebrew, then maybe this
     is the problem? If the following command doesn't work, then try removing
     your curlrc:
-      curl http://github.com
+      curl https://github.com
     EOS
     end
   end


DomT4:
CC @mistydemeo @mikemcquaid

Library/Homebrew/utils.rb
```diff
@@ -221,9 +221,8 @@ def quiet_system(cmd, *args)
 end
 
 def curl(*args)
-  brewed_curl = HOMEBREW_PREFIX/"opt/curl/bin/curl"
-  curl = if MacOS.version <= "10.6" && brewed_curl.exist?
-    brewed_curl
+  curl = if MacOS.version <= "10.6" && Formula["curl"].installed?
+    Formula["curl"].opt_bin/"curl"
```
mistydemeo: The previous way of doing this was intended to avoid instantiating formula objects (in particular because it could cause issues during the tests, where the `curl` formula is unavailable). Mind leaving it the old way?

DomT4: Ah, okay, makes sense. Will poke this commit out of existence.

mikemcquaid:
LGTM :+1:

PR #43302, opened 2015-08-26
gdal 1.11.2_3 (gdal.jar)

smithjlrhit:
Install dal.jar along with the native JNI bindings.

Native JNI bindings are typically very specific to the java wrapper.  This installs the associated
jar file to be used with the native JNI shared objects.


mikemcquaid:

For future reference the preferred commit message format is `gdal: install JAR and JNI bindings.`.

PR #43303, opened 2015-08-26
Improved nesting behavior for `String#undent`

woodruffw:
In #41021 I realized that Homebrew's current `String#undent` implementation does not play nicely
with nested heredocs. In particular, `String#undent` chops off the first <number of spaces> characters from a nested heredoc.

This commit changes this behavior by only `gsub`ing for whitespace characters (`\s`) instead of matching **all** characters (`.`). I've updated `test_undent` in `test_string.rb` to reflect this behavior, and added `test_undent_nested` to reflect the new expected behavior.

As far as I know, no heredocs in Homebrew rely on the **current** behavior of `String#undent` w/r/t `gsub`ing anything up to the indentation level. `BuildToolsError` and `BuildFlagsError` do have to be changed to remove workarounds for the old behavior, which I'll do presently.

\- William

### DIFF
--- a/Library/Homebrew/extend/string.rb
+++ b/Library/Homebrew/extend/string.rb
@@ -1,6 +1,6 @@
 class String
   def undent
-    gsub(/^.{#{(slice(/^ +/) || '').length}}/, "")
+    gsub(/^[ \t]{#{(slice(/^[ \t]+/) || '').length}}/, "")
   end
 
   # eg:
--- a/Library/Homebrew/exceptions.rb
+++ b/Library/Homebrew/exceptions.rb
@@ -273,25 +273,25 @@ class BuildToolsError < RuntimeError
     if MacOS.version >= "10.10"
       xcode_text = <<-EOS.undent
         To continue, you must install Xcode from the App Store,
-              or the CLT by running:
-                xcode-select --install
+        or the CLT by running:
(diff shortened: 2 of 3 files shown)


Library/Homebrew/extend/string.rb
```diff
@@ -1,6 +1,6 @@
 class String
   def undent
-    gsub(/^.{#{(slice(/^ +/) || '').length}}/, "")
+    gsub(/^\s{#{(slice(/^\s+/) || '').length}}/, "")
```
bfontaine: Note that `\s` matches newlines, which we probably don’t want here.

woodruffw: I see. How about `[ \t]`?

bfontaine: Yes, that’d work.

woodruffw: Changed!

mikemcquaid:
Works for me :+1:

woodruffw:
Awesome! Just waiting for the test results to come back to make sure I didn't make any silly mistakes.

xu-cheng:
Thanks @woodruffw

woodruffw:
Thanks!

PR #43304, opened 2015-08-26
fileio - formula for the first version of this utility

nicholaspufal:
I wrote a utility for [file.io](https://file.io) - an ephemeral file sharing service. 

It's a simple shell script and here is the formula to install it.


mikemcquaid:
This is a bit too niche and new for Homebrew currently. Sorry!

PR #43307, opened 2015-08-26
libgpg-error 1.20

DomT4:


DomT4:
Can't repro this locally:

```
==> Summary
🍺  /usr/local/Cellar/gpgme/1.6.0: 21 files, 1.9M, built in 83 seconds
```

Tried the build 3 times in a row. No failure. Have re-queued it on the bot in case there was some hangover from a previous build or something.

DomT4:
Thinking race condition here since I can't repro at all on two machines at `-j4`. Going to try `ENV.deparallelize`.

DomT4:
Going to strip `gpgme` out here for now. Still can't repro locally, which is frustrating.

PR #43308, opened 2015-08-27
nailgun: add head version

rhendric:


Library/Formula/nailgun.rb
```diff
@@ -1,8 +1,19 @@
...
+    url "https://github.com/martylamb/nailgun/archive/nailgun-all-0.9.1.tar.gz"
+    sha256 "c487735b07f3d65e4c4d9bfa9aaef86d0d78128e4c055c6c24da818a4a47b2ab"
+
+    resource "nailgun-jar" do
+      url "http://central.maven.org/maven2/com/martiansoftware/nailgun-server/0.9.1/nailgun-server-0.9.1.jar"
+      sha256 "4518faa6bf4bd26fccdc4d85e1625dc679381a08d56872d8ad12151dda9cef25"
+    end
+
+    # The makefile is not prefix aware: https://github.com/martylamb/nailgun/pull/45
+    patch :DATA
```
mikemcquaid: Does `https://github.com/martylamb/nailgun/pull/45.patch` work as a `patch do` `url`?

bfontaine: @mikemcquaid Shouldn’t it be `https://github.com/martylamb/nailgun/pull/45.diff` (`.diff` instead of `.patch`)?

mikemcquaid: We can do either now that GitHub has removed the version from the patch output.

rhendric: Oh, nice! I was looking for a way to do this but didn't find it. I'll change both patches.

mikemcquaid:

PR #43316, opened 2015-08-27
d-bus 1.10.0

stuwil:
Removed 1.9 devel version - 1.11 is the new devel branch, no releases yet. The O_CLOEXEC patch made it into 1.10, so removing that as well.


bfontaine:
Thank you for your contribution @stuwil, we appreciate it! :smile:

PR #43320, opened 2015-08-27
tig: create & install zsh completion which source bash completion file

wkentaro:


dunn:
This seems like the sort of thing that should be offered directly to the tig developers; then non-Homebrew users get it too and our formula file doesn't get too complicated.

wkentaro:
Okay, thanks for your comment.

PR #43324, opened 2015-08-27
Sandbox fix for Mecab family

xu-cheng:


Library/Formula/mecab-ipadic.rb
```diff
@@ -16,9 +18,24 @@ def install
...
+       dicdir = #{HOMEBREW_PREFIX}/lib/mecab/dic/ipadic
+    EOS
+  end
+
+  test do
+    (testpath/"mecabrc").write <<-EOS.undent
+      dicdir = #{HOMEBREW_PREFIX}/lib/mecab/dic/ipadic
+    EOS
+
+    pipe_output("mecab --rcfile=#{testpath}/mecabrc", "すもももももももものうち\n")
```
bfontaine: `pipe_output` [doesn’t assert the exit code by default](https://github.com/Homebrew/homebrew/blob/a70e68d4ee45485585df57768c9c5f4140a773ec/Library/Homebrew/formula_assertions.rb#L43); add `, 0` at the end to ensure the command exits with `0`.

xu-cheng: Should we instead make it assert by default?

bfontaine: We didn’t not to break retro-compatibility but I guess it’s fine if we do.

mikemcquaid: Yeh, I think it'd be good to add :+1:

Library/Formula/mecab-ipadic.rb
```diff
@@ -16,9 +18,24 @@ def install
...
       --with-charset=#{charset}
+      --with-dicdir=#{lib}/mecab/dic/ipadic
     ]
 
     system "./configure", *args
     system "make", "install"
   end
+
+  def caveats; <<-EOS.undent
+     To enable dictionary, rewrite #{HOMEBREW_PREFIX}/etc/mecabrc as
```
mikemcquaid: `To enable the mecab-ipadic dictionary add to #{HOMEBREW_PREFIX}/etc/mecabrc:`

mikemcquaid:
:+1:

PR #43325, opened 2015-08-27
sandbox: better log output

xu-cheng:
* use syslog filter instead of grep.
* output sandbox log to stdout when verbose and failed.
* output nothing if sandbox log is empty.

cc @mikemcquaid

### DIFF
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -32,7 +32,7 @@ class Sandbox
   end
 
   def record_log(file)
-    @log = file
+    @logfile = file
   end
 
   def add_rule(rule)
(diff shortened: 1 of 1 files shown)


Library/Homebrew/sandbox.rb
```diff
@@ -91,24 +98,36 @@ def exec(*args)
...
+      unless logs.empty?
+        @log.write Sandbox.log_warning_message
+        @log.write logs
+
+        if @failed && ARGV.verbose?
+          ohai "Sandbox log:"
+          puts Sandbox.log_warning_message
+          puts logs
+          ohai "Sandbox profile:"
+          puts @profile.dump
```
mikemcquaid: Do you find this message useful? If so we can keep it otherwise let's :skull:

Library/Homebrew/sandbox.rb
```diff
@@ -91,24 +98,36 @@ def exec(*args)
...
         -o
         -k Time ge #{@start.to_i}
+        -k Message S deny
         -k Sender sandboxd
       ]
-      quiet_system "syslog #{syslog_args * " "} | grep deny > #{@log}"
+      logs = Utils.popen_read("syslog", *syslog_args)
+      unless logs.empty?
+        @log.write Sandbox.log_warning_message
+        @log.write logs
```
mikemcquaid: Can we `grep` this for any string or is there no prefix?

Library/Homebrew/sandbox.rb
xu-cheng: It's useful to people who are unfamiliar to homebrew sandbox's rule. Otherwise, it's not as they are almost constant.

Library/Homebrew/sandbox.rb
xu-cheng: You mean the log format? 

Currently it's `$((Time)(local)) $(Sender)[$(PID)]: $(Message)`. e.g. `Aug 27 21:19:32 sandboxd[295]: ruby(93659) deny file-write-create /1`. We can tweak it if you want.

Library/Homebrew/sandbox.rb
mikemcquaid: Yeh, I guess I personally find this output kinda noise and hard to follow. Maybe we could add a few newlines or separators in here to split it up a bit if we leave it in?

xu-cheng: Or we remove this. And add it to document, say `sandbox.md`?

Library/Homebrew/sandbox.rb
mikemcquaid: :cool: looks good :+1:

Library/Homebrew/sandbox.rb
mikemcquaid: Yeh, maybe just remove it? A document will get outdated compared to the code.

xu-cheng:
Rewrote PR to address the comments.

Library/Homebrew/sandbox.rb
```diff
@@ -32,7 +32,13 @@ def initialize
...
   def record_log(file)
-    @log = file
+    @logfile = file
+  end
+
+  def self.log_warning_message; <<-EOS.undent
+      We use time to filter sandbox log. Therefore, unrelated logs may be recorded in below.
+
+    EOS
   end
```
mikemcquaid: I think we can probably skip printing this message unless it's causing confusion.

mikemcquaid:
Looks good to me!

Library/Homebrew/sandbox.rb
xu-cheng: How about only writing it to log file? Usually there won't be unrelated logs for bot. But on users computer, the chance is high which can cause confusion.

mikemcquaid: Yep, if we write it maybe after we print the log then that makes sense to me. Equally if we didn't write it if `HOMEBREW_DEVELOPER` is set or something? Up to you!

xu-cheng: Finally, I decide to only write this message to the bottom of file.

mikemcquaid:
:+1:

PR #43326, opened 2015-08-27
gant: migrating to boneyard

bfontaine:
The URL doesn’t work anymore; I noticed upstream two months ago, got no response since.

See #40392, https://github.com/Homebrew/homebrew-boneyard/pull/64.


mikemcquaid:
:+1:

PR #43336, opened 2015-08-27
speex: version bump to 1.2rc3

tzudot:
speex: version bump to 1.2rc3


tzudot:
Seems like there is a deviation in library name, it's now called speexdsp in .pc file.

PR #43338, opened 2015-08-27
sip: strip :python dependency

tdsmith:
Inspired by comment on 4ca15c7. sip memorizes a bunch of things but I
think they're mostly just relevant to understanding how to build against
whichever python and don't materially influence sip's output.

### DIFF
--- a/Library/Formula/sip.rb
+++ b/Library/Formula/sip.rb
@@ -13,7 +13,8 @@ class Sip < Formula
 
   head "http://www.riverbankcomputing.co.uk/hg/sip", :using => :hg
 
-  depends_on :python => :recommended
+  option "without-python", "Build without python2 support"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

Library/Formula/sip.rb
```diff
@@ -13,7 +13,8 @@ class Sip < Formula
 
   head "http://www.riverbankcomputing.co.uk/hg/sip", :using => :hg
 
-  depends_on :python => :recommended
+  option "without-python", "Build without python2 support"
```
xu-cheng: I wonder can we make it `depends_on :python => [:recommended, :run]`? It seems to be better regarding DSL.

mikemcquaid: :+1: if it works.

PR #43339, opened 2015-08-27
mongrel2 1.9.3

jkarneges:


jkarneges:
The build succeeds but running tests fails. It looks like the response_tests program writes to /dev/zero, and I wonder if this is inaccessible in the jenkins environment (it works fine for me locally).

dunn:
Yeah, we have a sandbox restricting where the test can write to.  Maybe `dev/zero` should be on the whitelist, though.

jkarneges:
Must be a new thing then? This test code was the same in the previous version.

I think it belongs on the whitelist. Along with /dev/null, if it isn't already there.

dunn:
Yep, we just turned it on a couple weeks ago.

dunn:
OK, whitelisted `/dev/zero`, let's kick off a new build.

jkarneges:
Thanks! Build failed again, but it seems to come from a pushpin test rather than anything from mongrel2. Does the build do some kind of reverse dependency testing? (pushpin depends on mongrel2)

dunn:
Yeah, we test the dependents to make sure there aren't breaking changes. But this looks like a sandbox issue. We'd be _super_ grateful if you could fix that up in this PR too.

jkarneges:
See #43349.

jkarneges:
Alright it looks like the pushpin issue was fixed and now the reverse deps tests succeed.

I hope this can be merged now.

xu-cheng:

PR #43352, opened 2015-08-28
tap: rmdir when tap failed or interrupted

### LINKED ISSUE #43280: Mistyping tap name creates local tap directory anyway
I was trying to tap `caskroom/cask` but mistyped it as `vaskroom/cask` and hit ENTER before I realized my mistake. But looking under `/usr/local/Library/Taps` I see that a `vaskroom` folder was created anyway.

xu-cheng:
Fixes #43280.


mikemcquaid:
:+1:

PR #43353, opened 2015-08-28
update: secure formula loading in rename

xu-cheng:
For update, all formula loading need to be secured by `rescue`.
Otherwise it's easy to crash `brew update` for vasty reasons, e.g.
new Formula DSL.

cc @mikemcquaid @vladshablinsky


mikemcquaid:
:+1:

PR #43363, opened 2015-08-28
Tips-'N-Tricks

dunn:
specific stuff:
- removed the stuff about `sudo` since it suggests we support it
- removed `$` at the beginning of commands except for multiple-line examples
- removed shell-specific highlighting where inappropriate
- added my homebrew-mode to the list of editor plugins

### DIFF
--- a/share/doc/homebrew/Tips-N'-Tricks.md
+++ b/share/doc/homebrew/Tips-N'-Tricks.md
@@ -120,10 +120,21 @@ export HOMEBREW_NO_EMOJI=1
 
 This sets the HOMEBREW_NO_EMOJI environment variable, causing homebrew to hide all emoji.
 
-## Sublime text: Syntax for formulae including the diff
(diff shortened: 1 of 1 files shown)


share/doc/homebrew/Tips-N'-Tricks.md
```diff
@@ -1,20 +1,25 @@
...
 
-```zsh
+```
 brew install https://raw.github.com/dsr/homebrew/9b22d42f50fcbc5e52c764448b3ac002bc153bd7/Library/Formula/python3.rb
 ```
 
 ## Quickly remove something from /usr/local
 
-```bash
+```
```
bfontaine: You can use `sh` everywhere instead.

dunn: Oh, yeah.

tdsmith:
:ship:

PR #43368, opened 2015-08-28
generate-json-schema v2.1.1 (new formula)

apeschel:
Submitting request for inclusion in standard Homebrew formulas.

### DIFF
--- a/Library/Formula/generate-json-schema.rb
+++ b/Library/Formula/generate-json-schema.rb
@@ -1,0 +1,115 @@
+class GenerateJsonSchema < Formula
+  desc "Generate a JSON Schema from Sample JSON"
+  homepage "https://github.com/Nijikokun/generate-schema"
+  url "https://github.com/Nijikokun/generate-schema/archive/v2.1.1.tar.gz"
+  sha256 "bf43a7e616419876293e7de738dece27b3deb719cf5e8cc99bb309a5fb179af0"
+
+  head "https://github.com/Nijikokun/generate-schema.git"
+
+  depends_on "node"
+
+  def install
+    ENV.prepend_path "PATH", "#{Formula["node"].opt_libexec}/npm/bin"
+
+    system "npm", "install"
+    libexec.install Dir["*"]
+    bin.install_symlink Dir["#{libexec}/bin/*"]
+  end
+
+  test do
+    require "open3"
+
+    input = <<-EOS.undent
+      {
+          "id": 2,
+          "name": "An ice sculpture",
+          "price": 12.50,
+          "tags": ["cold", "ice"],
+          "dimensions": {
+              "length": 7.0,
+              "width": 12.0,
+              "height": 9.5
+          },
+          "warehouseLocation": {
+              "latitude": -78.75,
+              "longitude": 20.4
+          }
+      }
+    EOS
+
+    output = <<-EOS.undent.chomp
+      Welcome to Generate Schema 2.1.1
+
+        Mode: json
+
+      * Example Usage:
+        > {a:'b'}
+        { a: { type: 'string' } }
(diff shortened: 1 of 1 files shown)


DomT4:
Some audit nits here:

```
== /usr/local/Library/Formula/generate-json-schema.rb ==
C: 20: 13: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

Error: 5 problems in 1 formula

==> audit problems
generate-json-schema:
 * 42: Trailing whitespace was found
 * 44: Trailing whitespace was found
 * 45: Trailing whitespace was found
 * 48: Trailing whitespace was found
 * 50: Trailing whitespace was found
```

apeschel:
@DomT4 Thanks for the feedback!
audit didn't show me the quote warning for some reason. I've fixed it now though.

The whitespace is something that cannot be fixed, unfortunately, as the output string for the test is expected to have those whitespaces. Removing them breaks the test.

Library/Formula/generate-json-schema.rb
```diff
@@ -0,0 +1,112 @@
...
+              "height": 9.5
+          },
+          "warehouseLocation": {
+              "latitude": -78.75,
+              "longitude": 20.4
+          }
+      }
+    EOS
+
+    output =
```
DomT4: Wrap this in an `<<-EOS.undent` as you did above and then use a two-space indent from underneath that point. The bot will ignore things inside the `EOS` shield.

apeschel: @DomT4 Hi Dom, I did not use a HEREDOC block since I couldn't come up with a clean way to remove the newline from the end of the string.

dunn: `output.chomp` should strip the trailing newline for you.

apeschel:
@DomT4 I switched to using a heredoc for the expected output, but I'm still getting the whitespace warnings from the audit.

Library/Formula/generate-json-schema.rb
apeschel: @dunn Thanks! I had forgot about that.

DomT4:
Which warnings are you seeing now?

apeschel:
```
$ brew audit generate-json-schema
generate-json-schema:
 * 42: Trailing whitespace was found
 * 44: Trailing whitespace was found
 * 45: Trailing whitespace was found
 * 48: Trailing whitespace was found
 * 50: Trailing whitespace was found
 * 97: Trailing whitespace was found

Error: 6 problems in 1 formula
```

apeschel:
@DomT4 Just in case you were not pinged previously.

DomT4:
@xu-cheng Wasn't there a way to ignore whitespacing inside heredoc blocks? Perhaps I misremembered that & it doesn't exist.

xu-cheng:
@apeschel Maybe you can try a different testcase, such as:
```ruby
  (testpath/"test.js").write <<-EOS.undent
    var GenerateSchema = require('generate-schema')
    var schema = GenerateSchema.json('Product', [ ... ])
    ...
  EOS
  system Formula["node"].opt_bin/"node", "test.js"
```

bfontaine:
Ping?

apeschel:
@xu-cheng I'm not sure I'm following your example. How will that help with the whitespace problems?

apeschel:
Barring a response from @xu-cheng, what needs to be done for this to get merged in?

DomT4:
We just need to wrestle the test free of audit failures really. That seems to be it.

apeschel:
Commit has been updated. The formula is passing both the audit and the tests now.

mikemcquaid:

Issue #43356, opened 2015-08-28
python integration breaks when Python upgrades


hacpai:
When I open anything.py with vim, it prompt

    E887: Sorry, this command is disabled, the Python's site module could not be loaded.

Then I run `:python`

    E471: Argument required

`:python print 'anything'`

    E887: Sorry, this command is disabled, the Python's site module could not be loaded.

Recently homebrew upgrades Python so this might be relevant

DomT4:
Try `brew reinstall vim`.

Vim is one of the few formula in Homebrew with semi-mandatory Python dependencies that aren't tied to the system's Python (because it uses `:std` env to build we essentially can't force it to pick one or the other). It seems to retain whatever paths it was compiled against.

hacpai:
@DomT4 Thanks. It works.

DomT4:
Thanks for the confirmation, closing here.

Issue #42171, opened 2015-07-27
Vim python functionaly breaks when python is updated


ptzz:
I'm using homebrew python as my default python and with homebrew vim I currently see `[...] -F/usr/local/Cellar/python/2.7.10_2/Frameworks  -framework Python [...]` when doing `vim --version`. Mostly/always vim plugins using python (e.g. `pymode`) breaks ("Python's site module could not be loaded" etc.) after upgrading python and `vim --reinstall´ is required to make it well-behaved again.

Does vim rely on the version of python available during install being the same as during runtime? If yes, would it make sense to force reinstall of vim if python is upgraded? It's not always obvious that a python version mismatch is the problem as some vim plugins will just start failing silently.

DomT4:
We build vim under `std`, which means we adopt the user's `$PATH` exactly how it was before compile.

Relevant to your problem, we use `depends_on :python` rather than `depends_on :python if MacOS.version <= :snow_leopard`. The latter would normally mean *"Use the system Python unless building on 10.6 or below"*, but enforcing that wouldn't be effective here.

Instead we use `depends_on :python` which roughly means *"Use the first Python in the `$PATH`"*. In this case, because you have Homebrew's Python installed, Vim will link to that on compile, and will likely need rebuilding when you change something that was in your `$PATH` at the time of compile, whether that's Ruby or Python version.

>  If yes, would it make sense to force reinstall of vim if python is upgraded

Sadly, this is also difficult. We can't assume every user has Homebrew's Python installed, so revisioning the formula to force a rebuilt would be worthless for everyone who has Vim built against the system Python.

I'm afraid there's little we can do on these situations; We're aware of the various Vim complications and are working over time to try and handle them, but for now, you'll need to manually `brew reinstall vim` when things like this happen. Sorry! Thanks for filing the issue.

PR #43371, opened 2015-08-28
requirements: add perl requirement

DomT4:
This is still *a bit rough* around the edges in terms of polishing all the changed code here, and maybe we choose not to go down this path, but I got bored so:

* Introduces a Perl requirement.
* Moves Vim from building under `:std` to superenv, whilst still preserving from-source builds being able to build against whichever Perl & Ruby you have in your `$PATH` locally, providing it meets the min reqs. Python switches to using the system Python, which may/may not be controversial - It should however stop every update/revision/etc of our Python breaking people's Vim installations.
* Can also bottle Vim but may not want to until we sort out the issues discussed in #41595.

Required Perl version sourced from [here](https://github.com/vim/vim/blob/master/src/configure.in#L907-L913) and cross-referenced against the [Perl website](http://www.cpan.org/src/README.html).

* Build logs against a non-brewed, non-system Perl [here](https://gist.github.com/anonymous/3f8b2b11021bb17a4c04).
* Build logs against the default system Perl [here](https://gist.github.com/anonymous/649ffdb8731c69a988a7).

Also confirmed the Perl requirement works as expected, which you can do locally by pulling and changing the `depends_on :perl => "5.3"` line to `depends_on :perl => "5.19"` or such.


DomT4:
Mountain Lion disapproves of the call to `Gem::` apparently. Will look at it.

Library/Homebrew/requirements/perl_requirement.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  satisfy :build_env => false do
+    next unless which "perl"
+    version = perl_version
+    next unless version
+    Version.new(version.to_s) >= Version.new(@version)
+  end
+
+  env do
+    if Gem::Version.new(perl_version) >= Gem::Version.new(@version)
```
xu-cheng: Why not just use `Version` here.

Library/Homebrew/requirements/perl_requirement.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def perl_version
+    @perl_version ||= Utils.popen_read("perl", "--version")[/(\d\.\d(\d)?)/, 1]
+  end
+
+  satisfy :build_env => false do
+    next unless which "perl"
+    version = perl_version
+    next unless version
+    Version.new(version.to_s) >= Version.new(@version)
```
xu-cheng: The variable name is a little confused here. Maybe
```
next unless which "perl"
next unless perl_version
Version.new(perl_version) >= Version.new(@version)
```

Library/Homebrew/requirements/perl_requirement.rb
DomT4: As in `if Gem::Version.new(version) >= Gem::Version.new(@version)`? Happy either way.

xu-cheng: I mean just `if Version.new(perl_version) >= Version.new(@version)`. The `Version` is from `version.rb`. So you don't need to `require "rubygems"`.

Library/Formula/vim.rb
```diff
@@ -74,11 +74,6 @@ def install
       opts << "--enable-luainterp"
     end
 
-    # XXX: Please do not submit a pull request that hardcodes the path
-    # to ruby: vim can be compiled against 1.8.x or 1.9.3-p385 and up.
-    # If you have problems with vim because of ruby, ensure a compatible
-    # version is first in your PATH when building vim.
```
chdiza: How come this section is being removed?

Library/Homebrew/requirements/perl_requirement.rb
DomT4: Oh, you mean we don't need to call `Gem::` on these. Yeah, you should be right. I'll peek at it in the morning.

Library/Homebrew/requirements/perl_requirement.rb
DomT4: :+1:

Library/Formula/vim.rb
DomT4: The Ruby dependency handles it. It'll build bottles against Homebrew's Ruby (I haven't decided whether to actually pull a bottle on this yet) but if you build it from source it'll build against the first valid Ruby it finds in the `$PATH`. The inclusion of a `depends_on :ruby` should make it obvious enough hardcoding would be invalid, theoretically.

chdiza: > inclusion of a depends_on :ruby should make it obvious enough hardcoding would be invalid, theoretically.

I confess it's not obvious to me, because when I see `depends_on` followed by a colon-prepended package name, I think, "not an ordinary ruby dep, but one that involves some conditional kung-fu depending on what my system is like, so maybe I can get HB to look at my special ruby location if I bug them".

In any case, the part of the section that says you can build from source against the first ruby in `$PATH` should be left in, because without that note, it really is totally unobvious.

DomT4: > "not an ordinary ruby dep, but one that involves some conditional kung-fu depending on what my system is like, so maybe I can get HB to look at my special ruby location if I bug them".

Well, this is basically what's going on when you build from source. There's just no need to bug us about it at this point, as we'll use whatever happens to be first in the $PATH. Nonetheless, I've added a small explanatory comment further down.

chdiza: > Nonetheless, I've added a small explanatory comment further down.

Perfect!

tdsmith:
vim embeds python so it matters which python is used, unfortunately; building against system python will confuse people trying to use popular stuff like powerline because the python module search path won't be what they expect

DomT4:
I "lost" a discussion with Tim about changing to using the system Python on this, so it's back to using whatever Python happens to be in the `$PATH`. My thoughts along the system Python change were to handle situations like #43356 & #42171 but I'll leave those errors to Tim's considerably better expertise.

Not a huge deal-breaker for me; I decided not to bother bottling this anyhow as it'll install all the `default_formula`, so which Python we use doesn't change anything in that regards. My end goal here was primarily to get one of the last few formulae off `:std` env (and all the problems there are with `:std`) and into superenv, which this still does.

Unless any @Homebrew/owners objects here, I'd like to get this in soonish. The Perl & Ruby special dependencies mean users shouldn't (and I don't locally) see any change in current behaviour; it'll still scoop up whatever Perl/Ruby/Python you have in your `$PATH`, it just no longer requires `:std` to pull that magic off.

xu-cheng:
A thought come to my mind, actually we can bottle it if we do something like this:

* If we wan to create bottle aginst brew Python
```ruby
depends_on :python => :recommended

def pour_bottle?
  Formula["python"].installed?
end
```
* Or, if we want to bottle against system Python
```ruby
depends_on :python => [:recommended, :run]

def pour_bottle?
  which("python") == "/usr/bin/python"
end
```
```

**Comment by tdsmith** [comment]
those should both be `which` checks; homebrew python might be installed but not linked.

the name `:run` really confuses me. i can write an explanation of what it does but i find it hard to reason about.

**Comment by xu-cheng** [comment]
I just find `Requirement` has `pour_bottle?` method but it seems to be never used in `formula_installer`. Also, we obviously overwrite this method in `PythonRequirement` (so as `Python3Requirement` due to inheritance) but I fail to understand its purpose.

**Comment by DomT4** [comment]
> A thought comes to my mind, actually we can bottle it if we do something like this:

Whilst this handles Python it doesn't handle the `default_formula` pulls of Ruby and Perl, which would be installed with the bottle as well. I guess we *could* check if all three formula were installed.

File: Library/Formula/vim.rb [hunk 4]
```diff
@@ -35,6 +35,15 @@ class Vim < Formula
...
     :because => "vim and ex-vi both install bin/ex and bin/view"
 
+  # Default requirements mean installing the bottle here requires
+  # installing our Perl, Python & Ruby. To prevent that for everyone
+  # just pour the bottle when requirements already satisfied.
+  def pour_bottle?
+    (which("python") == HOMEBREW_PREFIX/"bin/python") &&
+      (which("ruby") == HOMEBREW_PREFIX/"bin/ruby") &&
+      Formula["perl"].installed?
+  end
```
mikemcquaid: Hmm, are you sure this works how you expect (e.g. overriding the bottle `default_formula` behaviour)? If so I think it's probably an accident.

mikemcquaid: Ah, I see what you're doing here actually, ignore me, I was being stupid. Still, it feels a bit weird that this bottle will only be used if you've installed three system duplicates. I'd be tempted to say it should be the other way around (e.g. build against system versions by default and don't use the bottle if you're using non-system versions).

mikemcquaid:
I'm :cool: with adding this I just think we maybe need to work out some higher-level approach for handling how packages like this respond to bottling rather than having every package behave slightly differently.

mikemcquaid:
Maybe that means deciding we always build against system versions and doing the inverse of what we have here? That feels a bit more intuitive.

DomT4:
The slightly amusing thing about needing to install the Perl bottle for Vim's bottle is that the Perl on OS X versions that we bottle for is more than new enough to fulfil the requirement, so thanks to the requirement handling and Perl being `keg_only` we only feed it into the `$PATH` when the requirement isn't met by the system. Consequently, there's no linkage to our `keg_only` Perl in any of the bottles.

The `default_formula` needs to be more flexible than it is at this point. I don't know if you saw my half essay on the chat the other day, but my thoughts were more or less along the same lines as yours there with the default behaviour being to build against the system if the system meets the requirement.

tdsmith:
This still seems useful I think?

DomT4:
Yeah, I just need to split it into the various components now formula/core are no longer one. I'll do that later today. Per the discussion that killed the PR here last time we'll likely need to keep `vim` unbottled still but we might be able to get it building inside `superenv` at last.

DomT4:
Currently laughing at myself because 8 months after I opened this I've just noticed that we missed a spectacular problem with the implementation I chose at the time.

This worked because when the Perl requirement wasn't fulfilled it overwrote `superenv`'s handy shunting of `keg_only` deps to the front of the `$PATH` by shunting the first `perl` in the `$PATH` to the front instead, but in reality this almost universally meant when this requirement was invoked `/usr/bin` was pushed to the front of the `$PATH`.

It sounds bad in theory and gets worse in practice when you realise it'll happily override other requirements...
```
checking for ruby... /usr/bin/ruby
```
Rather than what the Ruby requirement actually permits, which is:
```
checking for ruby... /usr/local/opt/ruby/bin/ruby
```

Whoops. Looking into the broader problem further it seems it wasn't _really_ designed so that `default_formula` could ever be `keg_only`. The two mechanisms fundamentally tend to tread on each other.

Part of me is tempted to just resolve this in the simplest way by reverting `perl`'s `keg_only` status. It is the only language I can think of that we've made any effort to guard: Python, Ruby, Elisp are all available to use or call from the standard `$PATH` after you've `brew install`'d them.

mikemcquaid:
> Whoops. Looking into the broader problem further it seems it wasn't really designed so that default_formula could ever be keg_only. The two mechanisms fundamentally tend to tread on each other.

Yep, sounds right. Wasn't intentional but wasn't considered.



> Part of me is tempted to just resolve this in the simplest way by reverting perl's keg_only status. It is the only language I can think of that we've made any effort to guard: Python, Ruby, Elisp are all available to use or call from the standard $PATH after you've brew install'd them.

👍 to doing that to Perl.

DomT4:
https://github.com/Homebrew/homebrew-core/pull/531

DomT4:
With https://github.com/Homebrew/brew/pull/184 & https://github.com/Homebrew/homebrew-core/pull/784 open this is finally being closed. Cheers to everyone for the input on this.

PR #43374, opened 2015-08-28
fontforge: remove app

### LINKED ISSUE #43321: FontForge installs but does not run on Mavericks
The first attempt to install it gives a warning:
"Warning: fontforge dependency gettext was built with a different C++ standard library (libc++ from clang). This may cause problems at runtime."
Launching fontforge from command line starts python and waits
So I supposed it had something to do with the warning and reinstalled it with:
brew install fontforge --with-extra-tools --with-giflib --with-libspiro
but with the same result.
Copying the app folder to /Applications shows a Contents/Resources/opt/local folder with an etc folder only, so when launching 
/Applications/FontForge.app/Contents/MacOS/FontForge
it answers with:
PKGBASE: /Applications/FontForge.app
/Applications/FontForge.app/Contents/MacOS/FontForge: line 135: /Applications/FontForge.app/Contents/Resources/opt/local/bin/fontforge: No such file or directory

DomT4:
The App hasn't actually been fully supported for a long long time. Launching the GUI from the command-line worked until recently but that support passed with Cairo/Pango's mandatory X11 support.

If you want the App/GUI, I recommend checking out the Cask.

Closes #43321.

### DIFF
--- a/Library/Formula/fontforge.rb
+++ b/Library/Formula/fontforge.rb
@@ -93,6 +93,10 @@ class Fontforge < Formula
     system "make"
     system "make", "install"
 
+    # The app here is not functional. You should install Fontforge
+    # via the Cask if you want GUI/App support.
+    (share/"fontforge/osx/FontForge.app").rmtree
+
     if build.with? "extra-tools"
       cd "contrib/fonttools" do
         system "make"


DomT4:
People are getting confused about the app's presence, so easiest way to handle this is making the app depart.

PR #43377, opened 2015-08-29
wt: version 0.9.2

drewwells:
update to version 0.9.2 and remove libsass, no longer a dependency requirement

### DIFF
--- a/Library/Formula/wellington.rb
+++ b/Library/Formula/wellington.rb
@@ -5,10 +5,8 @@ class Wellington < Formula
   homepage "https://github.com/wellington/wellington"
 
   stable do
-    url "https://github.com/wellington/wellington/archive/v0.9.1.tar.gz"
-    sha256 "133a3d698f98139808c4a86955a9cdc0d9e91a0ed886c0851aceaed4595d8022"
-
-    depends_on "libsass"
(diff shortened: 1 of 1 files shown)


drewwells:
Why does the build links require github account access now? 

"brew test-bot by @Homebrew would like to modify its permission to access your account"

screen shot 2015-08-28 at 9 53 53 pm

DomT4:
As we've said previously in places, we don't actually control the permissions level. Those requests are not our choice, per se.

Nonetheless, the new permissions seem to be less reaching than the previous set of permissions, so if you've authorised the test-bot before the permissions should show themselves as changing to a lesser level. The build links have required some account access for as long as I've been around Homebrew.

mikemcquaid:
> Why do the build links require github account access now?

They always did.

> Nonetheless, the new permissions seem to be less reaching than the previous set of permissions, so if you've authorised the test-bot before the permissions should show themselves as changing to a lesser level. The build links have required some account access for as long as I've been around Homebrew.

This is correct, yes. We don't have permissions to do anything beyond check if you're in the Homebrew org.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/9164f2cc39fe2b9f855325dffb167322d25733f3. Thank you for your contribution @drewwells!

PR #43401, opened 2015-08-30
legit: fix wrong version specification

wkentaro:
I don't know why this had passed the test, but anyway the correct version is 0.2.0 for legit.


dunn:
Woops, thanks for catching that! Merged in b7f28f4aa9a696ff6ffac107b8c4daadab1b5a98.

PR #43402, opened 2015-08-30
legit: install completion files

wkentaro:


bfontaine:
We should probably add a `revision 1` here.

wkentaro:
Updated.

xu-cheng:
I suggest you use single PR to update the same formula. Otherwise, it's easy to confuse our ci bot.

xu-cheng:

PR #43406, opened 2015-08-30
legit: add head version

wkentaro:
Closes kennethreitz/legit/issues/47

### DIFF
--- a/Library/Formula/legit.rb
+++ b/Library/Formula/legit.rb
@@ -4,6 +4,7 @@ class Legit < Formula
   url "https://github.com/kennethreitz/legit/archive/v0.2.0.tar.gz"
   version "0.1.0"
   sha256 "dce86a16d9c95e2a7d93be75f1fc17c67d3cd2a137819fa498e179bf21daf39e"
+  head "https://github.com/kennethreitz/legit.git", :branch => "develop"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Library/Formula/legit.rb
```diff
@@ -55,6 +56,10 @@ def install
 
     bin.install Dir["#{libexec}/bin/*"]
     bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+
+    if build.head?
+      man1.install "extra/man/legit.1"
+    end
```
bfontaine: You can inline this: `man1.install "extra/man/legit.1" if build.head?`. Why is it done only when building `HEAD`?

wkentaro: Because it has not yet released.

wkentaro:
Updated.

wkentaro:
revision 2?

dunn:
We only add revisions for changes to the stable version, otherwise the majority of users (who run stable) would get an unnecessary upgrade.

xu-cheng:
Why the formula has version 0.2.0 in URL while it's explicitly set as 0.1.0

dunn:
@xu-cheng that was a mistake, fixed in b7f28f4aa9a696ff6ffac107b8c4daadab1b5a98

xu-cheng:

PR #43407, opened 2015-08-30
gambit-scheme 4.7.8

nijikon:


DomT4:
```
gambit-scheme:
 * Emacs Lisp files were linked directly to /usr/local/share/emacs/site-lisp

This may cause conflicts with other packages; install to a subdirectory instead, such as
/usr/local/Cellar/gambit-scheme/4.7.8/share/emacs/site-lisp/gambit-scheme

The offending files are:
  /usr/local/Cellar/gambit-scheme/4.7.8/share/emacs/site-lisp/gambit.el
```

There should be a configure flag for this. @dunn's the Emacs guru :smiley:

dunn:
Looks like it's set in the makefile.  This should work:
```diff
diff --git a/Library/Formula/gambit-scheme.rb b/Library/Formula/gambit-scheme.rb
index df89310..8c29a88 100644
--- a/Library/Formula/gambit-scheme.rb
+++ b/Library/Formula/gambit-scheme.rb
@@ -44,11 +44,11 @@ class GambitScheme < Formula
     end
 
     system "./configure", *args
-    system "make check" if build.with? "check"
+    system "make", "check" if build.with? "check"
 
     ENV.j1
     system "make"
-    system "make", "install"
+    system "make", "install", "emacsdir=#{share}/emacs/site-lisp/gambit-scheme"
   end
 
   test do
```

nijikon:
thanks @dunn & @DomT4

DomT4:
The build timed out here, but passed on Yosemite. I'll bump the timeout and set it running again.

nijikon:
@DomT4 can you bump the timeout so we rerun and can close this?

DomT4:
I bumped the timeout to over two hours the other night & it still failed :/. I'll reschedule it this evening when there's less bot activity.

nijikon:
Awesome!

DomT4:
@BrewTestBot test this please

DomT4:
Timeout is now at 4 hours. If you hit it again, it probably isn't going to pass :smile_cat:

nijikon:
It did not pass, what are we going to do with this?

mikemcquaid:
Not sure. Does this update often? If not we could build a new version as a one-off.

nijikon:
@mikemcquaid updates are not regular, depends, from the past I can see that there is a new version every 2 months. https://github.com/feeley/gambit/releases

nijikon:
How can I switch the compiler to gcc? So we don't build this with clang but with gcc.

mikemcquaid:
Use `fails_with`

nijikon:
OK, let's try to build this with gcc.

nijikon:
@mikemcquaid LOL, it worked!

nijikon:
I will remove the comment about clang compiler. It is not needed as we mark it as fails with clang. Is that ok?

DomT4:
Go for it. Kind of curious to see if the bot does the build in 20 minutes again, after timing out at 4+ hours previously :smile_cat:.

nijikon:
@DomT4 Took 24 min on master :dancer:

Library/Formula/gambit-scheme.rb
```diff
@@ -32,7 +34,7 @@ def install
     # Don't enable this when using clang, per configure warning.
     args << "--enable-single-host" unless ENV.compiler == :clang
 
-    args << "--enable-shared" if build.include? "enable-shared"
+    args << "--enable-shared" if build.with? "shared"
 
     if ENV.compiler == :clang
```
dunn: Can we remove this block now, and the `unless ENV.compiler == :clang` on line 35?

servercrunch: Yes, that is exactly what I was asking for, updating now.

nijikon: Yes, that is exactly what I was asking for, updating now.

dunn: Oh, I missed @nijikon's comment above.

Thanks for sticking with this, and figuring it out!

DomT4: You removed slightly too much here :wink:. We need to keep the:

```ruby
args << "--enable-single-host" unless ENV.compiler == :clang
```
line & explanation, as obviously, we'll always be using something that isn't `:clang` now.

If you want to you can compress it into the `args = %W[` block.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/7c8344be9f3249f4c0ea32272f981fbc38639ffe. Thank you for sticking with us on this bumpy ride @nijikon! :smile_cat:

PR #43413, opened 2015-08-30
cask: make emacs a run-time dependency

### LINKED ISSUE #43400: cask (emacs package management tool) dependancy issue
The formula for [cask](https://github.com/cask/cask), a project management tool for emacs (not to be confused with homebrew-cask) recently changed to include a dependancy for emacs. See c7e7ce42ed55020ca22fa4649c05b7e8d71bc6c0.

I use the [Yamamoto Mitsuharu port](https://github.com/railwaycat/mirror-emacs-mac) of emacs and not the homebrew version of emacs.

However when trying to update cask with homebrew it now wants to install the hombrew version of emacs. How do I get around this?

dunn:
Fixes #43400.

### DIFF
--- a/Library/Formula/cask.rb
+++ b/Library/Formula/cask.rb
@@ -13,7 +13,7 @@ class Cask < Formula
     sha256 "8262e37f32afc254ff6a25baabaa48537961bec0094e5b01aae73a9cc4baf87c" => :mountain_lion
   end
 
-  depends_on :emacs => "24"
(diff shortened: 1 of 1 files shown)


dunn:
@DomT4 was it actually necessary to update the bottles here?

DomT4:
If there wasn't linkage from the installed tools to the existing Homebrewed Emacs, you should be okay not updating the bottles. No harm either way though.

PR #43421, opened 2015-08-30
tig: add zsh completion installation for head version

wkentaro:
See also https://github.com/jonas/tig/pull/433

### DIFF
--- a/Library/Formula/tig.rb
+++ b/Library/Formula/tig.rb
@@ -3,6 +3,18 @@ class Tig < Formula
   homepage "http://jonas.nitro.dk/tig/"
   url "http://jonas.nitro.dk/tig/releases/tig-2.1.1.tar.gz"
   sha256 "50c5179fd564b829b6b2cec087e66f10cf8799601de19350df0772ae77e4852f"
+  head "https://github.com/jonas/tig.git"
+
+  stable do
+    url "http://jonas.nitro.dk/tig/releases/tig-2.1.1.tar.gz"
+    sha256 "50c5179fd564b829b6b2cec087e66f10cf8799601de19350df0772ae77e4852f"
+
+    # Merged in HEAD; remove in next stable release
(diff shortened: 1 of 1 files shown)


dunn:
Does the completion rely on any HEAD-only functionality or can it be back-ported to stable?

wkentaro:
the completion file

2015年8月31日月曜日、Alex Dunn<notifications@github.com>さんは書きました:

> Does the completion rely on any HEAD-only functionality or can it be
> back-ported to stable?
>
> —
> <https://github.com/Homebrew/homebrew/pull/43421#issuecomment-136249144>.
>


-- 
和田 健太郎 / Kentaro Wada
http://wkentaro.com <http://www.wkentaro.com>

wkentaro:
the completion file is there on HEAD-only.

dunn:
We can take the commit and use it as a patch in a stable block:
```ruby
  stable do
    url "http://jonas.nitro.dk/tig/releases/tig-2.1.1.tar.gz"
    sha256 "50c5179fd564b829b6b2cec087e66f10cf8799601de19350df0772ae77e4852f"

    # Merged in HEAD; remove in next stable release
    patch do
      url "https://github.com/jonas/tig/commit/91912eb97da4f6907015dab41ef9bba315730854.diff"
      sha256 "263e86b25f788eb158bdd667e112bc839debe9e3fe788cbc39cc9654b65b6c8a"
    end
  end
```

wkentaro:
instead of  
```
head "https://github.com/jonas/tig.git"
```
?

dunn:
Oh no, in addition. That way head and stable builds both get zsh completion.

wkentaro:
I see thanks. I added.

dunn:
Thanks so much! Merged in 472a7f002d55783111a0a7b508cd0e55d49b5d8a.

PR #40437, opened 2015-06-05
subversion: add non-homebrew universal python fail

DomT4:
Borrows my change from https://github.com/Homebrew/homebrew-versions/pull/779

Closes #34119

PR #43425, opened 2015-08-30
subversion: remove universal python fail

dlh:
Remove `--universal` requirement when using `--with-python` and a universal Python.

The previous fix's pull request: #40437

Using `ARCHFLAGS` is [the documented way to handle building a Python extension for a single architecture](https://developer.apple.com/library/mac/releasenotes/OpenSource/PerlExtensionsRelNotes/#//apple_ref/doc/uid/TP40006659-CH1-DontLinkElementID_3).

### DIFF
--- a/Library/Formula/subversion.rb
+++ b/Library/Formula/subversion.rb
@@ -72,19 +72,6 @@ class Subversion < Formula
   end
 
   def install
-    # OS X's Python is built universally and can't link with Homebrew's deps
-    # unless Homebrew's deps are universal as well.
-    # https://github.com/Homebrew/homebrew-versions/issues/777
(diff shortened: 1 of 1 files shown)


dlh:
Ping.

By the way, [the commit summary](/dlh/homebrew/commit/b02225093b9267ca75aa3c7fd9df957e4d7851a5) wasn't meant to be snarky (if you interpret it that way). It's derivative of the previous pull request's summary, *subversion: add non-homebrew universal python fail*.

mikemcquaid:
Sorry for the delay. Makes sense to me and don't worry about the snark, wasn't taken that way! Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #43428, opened 2015-08-31
v8 4.4.63.31

### LINKED ISSUE #41091: Errors when linking to bottled v8
I'm attempting to get [Google JS Test](https://github.com/google/gjstest) working with Homebrew (again, now that re2 has a non-head-only formula). The problem is that it no longer compiles with the bottle of v8 in homebrew:

```
% git checkout v8-4.1
% make clean && make
[...]
Undefined symbols for architecture x86_64:
  "std::_Rb_tree_decrement(std::_Rb_tree_node_base*)", referenced from:
      std::_Rb_tree<v8::Isolate*, std::pair<v8::Isolate* const, std::queue<v8::Task*, std::deque<v8::Task*, std::allocator<v8::Task*> > > >, std::_Select1st<std::pair<v8::Isolate* const, std::queue<v8::Task*, std::deque<v8::Task*, std::allocator<v8::Task*> > > > >, std::less<v8::Isolate*>, std::allocator<std::pair<v8::Isolate* const, std::queue<v8::Task*, std::deque<v8::Task*, std::allocator<v8::Task*> > > > > >::_M_insert_unique(std::_Rb_tree_iterator<std::pair<v8::Isolate* const, std::queue<v8::Task*, std::deque<v8::Task*, std::allocator<v8::Task*> > > > >, std::pair<v8::Isolate* const, std::queue<v8::Task*, std::deque<v8::Task*, std::allocator<v8::Task*> > > > const&) in libv8_libplatform.a(default-platform.o)
[...]
```

Note that this is against v8 installed from `v8-4.1.0.27.yosemite.bottle.tar.gz`.

As far as I can tell [this is due to](http://stackoverflow.com/questions/24718996/linker-error-undefined-symbols-stdstringc-str-const-on-macos-with-libboo) the version of v8 installed by homebrew having been built with g++, and gjstest compiling with clang. Supposedly you can work around this by setting `-stdlib=libstdc++`, but this doesn't seem to work for me due to C++11 features not being available that way.

Since g++ isn't around on most Macs these days, would it make sense to bottle v8 using clang instead? Or am I just misconfiguring something?

DomT4:
Closes #41091.

### DIFF
--- a/Library/Formula/v8.rb
+++ b/Library/Formula/v8.rb
@@ -1,12 +1,10 @@
 # Track Chrome stable.
-# https://github.com/v8/v8-git-mirror/commits/901b67916
 # https://omahaproxy.appspot.com/
-
 class V8 < Formula
   desc "Google's JavaScript engine"
   homepage "https://code.google.com/p/v8/"
-  url "https://github.com/v8/v8-git-mirror/archive/4.1.0.27.tar.gz"
-  sha256 "88bafa0bf80154f8f00e9808acd90a9233c0589c5da46ac4ebe3489ce914b87a"
+  url "https://github.com/v8/v8-git-mirror/archive/4.4.63.31.tar.gz"
+  sha256 "f384461a492243cd4ceac8f144b32c9b2f89b5935afccf4f2dc4479127dc019b"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


dunn:
Where is it listed what version corresponds to Chrome stable?

Library/Formula/v8.rb
```diff
@@ -24,24 +22,44 @@ class V8 < Formula
...
   resource "gtest" do
-    url "http://googletest.googlecode.com/svn/trunk", :revision => 700
-    version "700"
+    url "https://chromium.googlesource.com/external/googletest.git",
+        :revision => "be1868139ffe0ccd0e8e3b37292b84c821d9c8ad"
   end
 
   def install
+    # Bully GYP into correctly linking with c++11
+    ENV.libcxx
```
mistydemeo: Is it C++11 or libc++ you want? If it's the former, use `ENV.cxx11`, which will also set the appropriate flags when building with GCC.

DomT4:
https://omahaproxy.appspot.com/

The first table on the page. Scroll to the far right and you'll see the V8 versions.

Library/Formula/v8.rb
DomT4: Pretty sure I meant the former and just had a 3am brain failure, Sorry :weary:. Will re-do the build locally just to make sure.

DomT4: Fixed and pushed. Thanks for spotting this.

dunn:
Oh, neat, thanks!

PR #43433, opened 2015-08-31
cmd/search: fix filtering of aliases in results

dunn:
By directly modifying the results array with
`results[i] = "#{name} (installed)"`, it appeared on successive
iterations that the canonical name was no longer in the array, so
aliases were not removed.

See https://github.com/Homebrew/homebrew/commit/9efe5b554ce9cf6626d9794173725f5e063e5806#commitcomment-12969631

### DIFF
--- a/Library/Homebrew/cmd/search.rb
+++ b/Library/Homebrew/cmd/search.rb
@@ -143,15 +143,16 @@ module Homebrew
     aliases = Formula.aliases
     results = (Formula.full_names+aliases).grep(rx).sort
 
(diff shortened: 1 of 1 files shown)


dunn:
It's late, so I don't trust myself to just push this.

mikemcquaid:
@dunn Just for future reference if it's too late to push it's probably just worth pushing a revert of your previous commit.

dunn:
Ah, ok. Will do.

PR #43434, opened 2015-08-31
legit: add stable block to install zsh completion

wkentaro:


Library/Formula/legit.rb
```diff
@@ -6,6 +6,17 @@ class Legit < Formula
   head "https://github.com/kennethreitz/legit.git", :branch => "develop"
   revision 1
 
+  stable do
+    url "https://github.com/kennethreitz/legit/archive/v0.2.0.tar.gz"
```
DomT4: You don't need two versions of the url and checksum. Having the url/checksum tucked inside the stable block and not at the top of the formula is fine.

wkentaro:
Updated.

mikemcquaid:

PR #43437, opened 2015-08-31
evince 3.16.1 (new formula)

tschoonj:
new formula
libspectre is a dependency too, but was not available in Homebrew
libsecret is also a dependency, but was outdated

### DIFF
--- a/Library/Formula/evince.rb
+++ b/Library/Formula/evince.rb
@@ -1,0 +1,43 @@
+class Evince < Formula
+  desc "GNOME document viewer"
+  homepage "https://wiki.gnome.org/Apps/Evince"
+  url "https://download.gnome.org/sources/evince/3.16/evince-3.16.1.tar.xz"
+  sha256 "06ff75065b2a30cd588c402f6bd2ea88ee3166181805e0cc00bd54f71dbf6130"
+
+  depends_on "pkg-config" => :build
+  depends_on "intltool" => :build
+  depends_on "itstool" => :build
+  depends_on "poppler"
+  depends_on "libxml2" => "with-python"
+  depends_on "gtk+3"
+  depends_on "hicolor-icon-theme"
+  depends_on "gnome-icon-theme"
+  depends_on "libsecret"
+  depends_on "libspectre"
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    # forces use of gtk3-update-icon-cache instead of gtk-update-icon-cache. No bugreport should
+    # be filed for this since it only occurs because Homebrew renames gtk+3's gtk-update-icon-cache
+    # to gtk3-update-icon-cache in order to avoid a collision between gtk+ and gtk+3.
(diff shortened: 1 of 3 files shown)


Library/Formula/evince.rb
```diff
@@ -0,0 +1,60 @@
...
+  depends_on "hicolor-icon-theme"
+  depends_on "gnome-icon-theme"
+  depends_on "libsecret"
+  depends_on "libspectre"
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  # forces use of gtk3-update-icon-cache instead of gtk-update-icon-cache. No bugreport should
+  # be filed for this since it only occurs because Homebrew renames gtk+3's gtk-update-icon-cache
+  # to gtk3-update-icon-cache in order to avoid a collision between gtk+ and gtk+3.
+  patch :DATA
```
bfontaine: We can use `inreplace` instead since it’s a small change and Homebrew-specific.

tschoonj:
done

Library/Formula/libspectre.rb
```diff
@@ -0,0 +1,36 @@
+class Libspectre < Formula
+  desc "Small library for rendering Postscript documents"
+  homepage "https://wiki.freedesktop.org/www/Software/libspectre/"
+  url "http://libspectre.freedesktop.org/releases/libspectre-0.2.7.tar.gz"
+  sha256 "e81b822a106beed14cf0fec70f1b890c690c2ffa150fa2eee41dc26518a6c3ec"
+
+  depends_on "ghostscript"
+
+  def install
+    ENV.append "CFLAGS", "-I#{Formula["ghostscript"].opt_include}/ghostscript"
```
DomT4: Can it not find these with `pkg-config` at all?

tschoonj: nope. `ghostscript` doesn't have a pkg-config file.

DomT4: I forget there are things that still don't. Any chance we can feed it a prefix then or are we stuck with flags?

tschoonj: Unfortunately this appears to be the only way to pass the ghostscript flags. Strictly speaking they are not necessary when the brew prefix is /usr/local, but I added them to take care of users with non-default brew prefixes

DomT4: Ah, alright. I'll take it as is then. If you're feeling kind it might be cool to ask upstream to work in a `--with-gs=$DIR` option or similar.

tschoonj: Thanks for pulling this one in (and the others you did this morning)!

DomT4:
Signed, Sealed & Delivered in https://github.com/Homebrew/homebrew/commit/036dd0fba9d16fcd17563dc450b66f22a5e58d19, https://github.com/Homebrew/homebrew/commit/2e9befeb8c2ab68a9033f2af5eb0043066e33df5 & https://github.com/Homebrew/homebrew/commit/17c076a63feb484f1aa8011500d1efb503a6d27c. Thanks Tom!

PR #43439, opened 2015-08-31
id3lib fix unicode support

tschoonj:
as recommended by the easy-tag developers
See also Homebrew/homebrew-x11#123

### DIFF
--- a/Library/Formula/id3lib.rb
+++ b/Library/Formula/id3lib.rb
@@ -1,6 +1,7 @@
 class Id3lib < Formula
   desc "ID3 tag manipulation"
   homepage "http://id3lib.sourceforge.net/"
+  revision 1
 
   stable do
     url "https://downloads.sourceforge.net/project/id3lib/id3lib/3.8.3/id3lib-3.8.3.tar.gz"
@@ -50,6 +51,12 @@ class Id3lib < Formula
     url "https://raw.githubusercontent.com/DomT4/scripts/c24f2952/Homebrew_Resources/MacPorts_Import/id3lib/r112430/boolcheck.patch"
     mirror "https://trac.macports.org/export/112430/trunk/dports/audio/id3lib/files/boolcheck.patch"
     sha256 "a7881dc25665f284798934ba19092d1eb45ca515a34e5c473accd144aa1a215a"
+  end
+
+  # fixes Unicode display problem in easytag: see Homebrew/homebrew-x11#123
+  patch do
+    url "https://git.gnome.org/browse/easytag/plain/src/tags/id3lib/patch_id3lib_3.8.3_UTF16_writing_bug.diff"
+    sha256 "71c79002d9485965a3a93e87ecbd7fed8f89f64340433b7ccd263d21385ac969"
   end
 
   fails_with :llvm do


Library/Formula/id3lib.rb
```diff
@@ -52,6 +53,12 @@ class Id3lib < Formula
     sha256 "a7881dc25665f284798934ba19092d1eb45ca515a34e5c473accd144aa1a215a"
   end
 
+  # fixes Unicode display problem in easytag: see Homebrew/homebrew-x11#123
```
robotvert: it's not so much a display problem as it is an encoding issue: the tags get filled with erroneous sequences and opening the mp3s in an unrelated player such as VLC then displays mojibake'd info.

Library/Formula/id3lib.rb
```diff
@@ -52,6 +53,12 @@ class Id3lib < Formula
...
+  patch do
```
DomT4: Has this already been applied upstream? If so you'll need to tuck this inside `stable do` to stop `head` build conflicts.

tschoonj:
As far as I can tell id3lib is dead. There hasn't been a release for more than a decade and no commits in the last five years. This explains the many patches necessary to get it compiled.

The distinction between stable and HEAD seems kind of pointless honestly.

DomT4:
Oh dear. Normally at this point where we're having to pile on patch after patch to keep a dead project alive we'd just get rid of it, but this still seems reasonably popular which makes that tricky. I guess I've no objections to one more patch on the pile. Is there anywhere, even futilely, that we can report the patch upstream to?

tschoonj:
@robotvert Could you please find out what the easy-tag people did with this patch? Perhaps they submitted it upstream?

robotvert:
@tschoonj As you pointed out id3lib is unmaintained so even if they submitted that patch it's not been merged, right. I suppose one way to deal with this if adding a patch to the existing pile feels wrong is to find some C/C++ hacker to fork the project, merge the various patches and have homebrew use that version but as you can imagine...


On a side note I've not looked at easytag's code but it started to behave that way only from v2.4. Could it be that up until now they would "silently" workaround these limitations? Pure speculation at this point.
@amigadave, any thoughts?

amigadave:
The patch in question has a link to the id3lib bug report: http://sourceforge.net/tracker/index.php?func=detail&aid=1016290&group_id=979&atid=300979
There has been no response since 2006. The Fedora and Debian packages also contain the same fix: http://pkgs.fedoraproject.org/cgit/id3lib.git/tree/id3lib-3.8.3-io_helpers-163101.patch http://anonscm.debian.org/viewvc/collab-maint/deb-maint/id3lib/trunk/debian/patches/30-fix-utf16.patch?revision=14890&view=markup

This may only have shown up recently in EasyTAG because the default ID3v2 mode was changed from ID3v2.4 (using UTF-8) to ID3v2.3 (using UTF-16) during the 2.3 development series.

robotvert:
@amigadave thanks for the explanation! and btw, thanks for easy-tag!!! :+1: :+1: :+1:

DomT4:
If other packagers are carrying it my objection is lessened, although I have some doubts how long we can keep patching this to support it without an upstream.

For now though, merged in https://github.com/Homebrew/homebrew/commit/29b6a9fcfc0ed61ea2829047dcb5b7f947f11504. Thank you @tschoonj!

PR #43449, opened 2015-08-31
Remove references to bin/examples

gabelevi:
Flow used to stick some examples in the bin directory, but we stopped doing that as of v0.14.0 (https://github.com/facebook/flow/commit/b904adf1aaefa5b58990819bc03b01e7b618cb32). However the homebrew formula was still referencing it and that would cause errors.

Fixes facebook/flow#765

### DIFF
--- a/Library/Formula/flow.rb
+++ b/Library/Formula/flow.rb
@@ -12,20 +12,23 @@ class Flow < Formula
     sha256 "5cf57fcca47de49144872145c9bb8b004c649f0ff12b1927573ddf821710275a" => :mountain_lion
   end
 
-  depends_on "objective-caml" => :build
+  depends_on "ocaml" => :build
 
   def install
     system "make"
     bin.install "bin/flow"
-    (share/"flow").install "bin/examples"
 
     bash_completion.install "resources/shell/bash-completion" => "flow-completion.bash"
     zsh_completion.install_symlink bash_completion/"flow-completion.bash" => "_flow"
   end
 
   test do
-    output = `#{bin}/flow single #{share}/flow/examples/01_HelloWorld`
-    assert_match(/This type is incompatible with/, output)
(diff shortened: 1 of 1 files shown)


dunn:
Would it be possible to replace the test with another that checks the functionality a bit more than `--version` does?

DomT4:
It'd be nice to try and use `shell_output` or `pipe_output` rather than backtics if possible as well.

dunn:
FYI also
```
flow:
 * Dependency 'objective-caml' was renamed; use newname 'ocaml'.
 * Dependency 'objective-caml' was renamed; use newname 'ocaml'.

**Comment by gabelevi** [comment]
Sure, I'll address the feedback.

As for tests, I'll see what's available when `brew test` runs. If I have access to Flow's source code, I can probably just run `make test`. Otherwise, I'll see if I can do something simple like

```Bash
mkdir test
cd test
flow init test
echo "var x: string = 123;" > test/test.js
flow check test
rm -r test
```

I think I can infer how to test this locally by looking at Jenkin's console log.

DomT4:
You won't have access to anything in the `buildpath` - The `testpath` is a separate entity. You also don't need to create or remove top-level temporary directories inside `test do`; we handle that automatically.

gabelevi:
Whoops

gabelevi:
Can someone merge this? It should be good to go

dunn:
Sorry for the delay! Could you squash into one commit, though? Thanks!

gabelevi:
Squashed!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/2cb9f9c0be6cef34ff2925871b6fdaee54253ece. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #43457, opened 2015-09-01
gjstest 1.0.1 (new formula)

jacobsa:
There was a gjstest formula in the past, but it stopped building due to
v8 issues and was removed. As of the v8 upgrade in 9dae089, this can
work again.

For google/gjstest#3.

### DIFF
--- a/Library/Formula/gjstest.rb
+++ b/Library/Formula/gjstest.rb
@@ -1,0 +1,34 @@
+class Gjstest < Formula
+  desc "Fast javascript unit testing framework that runs on the V8 engine."
+  homepage "https://github.com/google/gjstest"
+  url "https://github.com/google/gjstest/archive/v1.0.1.tar.gz"
+  sha256 "c64d1b8d153b9afaa17ce521d9a7d07acbf13838dab7b2e57822f0bf046f80e5"
+  head "https://github.com/google/gjstest.git"
+
+  depends_on :macos => :mavericks
+
+  depends_on "gflags"
+  depends_on "glog"
+  depends_on "libxml2"
+  depends_on "protobuf"
+  depends_on "re2"
+  depends_on "v8"
+
+  def install
+    system "make", "PREFIX=#{prefix}", "install"
+  end
+
+  test do
+    (testpath/"sample_test.js").write <<-EOF
+      function SampleTest() {
+      }
+      registerTestSuite(SampleTest);
+
+      addTest(SampleTest, function twoPlusTwoEqualsFour() {
+        expectEq(4, 2+2);
+      });
+    EOF
+
+    system "#{bin}/gjstest", "--js_files", "#{testpath}/sample_test.js"
+  end
+end


dunn:
Any reason to not include `head "https://github.com/google/gjstest.git"`?

jacobsa:
Only that I didn't realize that was a thing you could/should do. Done.

dunn:
:ok_woman:

jacobsa:
@dunn: Are you able to pull this? I don't know how to see who has commit permissions.

dunn:
Yep, once the build passes.

jacobsa:
Ah I see, sorry. I had assumed it was stuck because it's been this way for awhile, and I can't find a link to show progress (all I see is "Build triggered. sha1 is merged."). Is there some way to see progress or troubleshoot issues?

dunn:
There's a big queue for the bots right now so it hasn't even started 🙀 the Jenkins interface is far from intuitive, but it's here: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/

jacobsa:
@dunn: The build above appears to have failed because the bot was using an old version of v8; not the one at `HEAD` in homebrew. Are the builds not hermetic, using the latest dependencies? What can I do about this?

dunn:
Yeah, everything should be up-to-date for each test build.  The failure is only happening on Mountain Lion, too.  Try adding `ENV.libcxx` at the beginning of the install block.

jacobsa:
Ah you're right, I had misdiagnosed the problem; it seems to just be C++11 support. I've added that, fingers crossed.

jacobsa:
@dunn: That still failed, with another build issue related to RE2. I don't actually care about supporting two versions back, so I've restricted the formula to Mavericks and above. Please take another look.

DomT4:
@BrewTestBot test this please

DomT4:
Just merged a V8 update (Chrome released a new stable) so putting this through again to double check prior to merge.

dunn:
Whoever merges this should add `"google-js-test" => "gjstest"` to `formula_renames.rb` and remove the former from head-only.

DomT4:
> Whoever merges this should add "google-js-test" => "gjstest"

Does doing renames for a formula that won't necessarily exist on a user's machine work/not throw weird errors?

dunn:
Good question. cc @vladshablinsky

jacobsa:
@DomT4: The build [failed](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/32563/version=mavericks/testReport/junit/brew-test-bot/mavericks/fetch_cmake_gflags_git_glog_libxml2_protobuf_python_re2_v8/) on Mavericks with differing SHA-256 errors for v8. Any idea what is going on?

dunn:
I get the "wrong" SHA as well, fwiw.

DomT4:
Yosemite is okay:

```
Downloaded to: /usr/local/var/homebrew/cache/v8-4.5.103.29.yosemite.bottle.tar.gz
SHA1: 9b569368fd39e16710c73d878ccd93687add21b7
SHA256: 739f6745233c716427c48e3e1a45ea431ce9dbd2113a5fb3a77bc1a95fdfe92a
==> Verifying v8-4.5.103.29.yosemite.bottle.tar.gz checksum
```

Wonder what happened with Mavericks. Confirmed though:
```
e60a8df1beff9601f2120966c44caa4e4b459106225ecef6f1c9a49216d1ba9e  v8-4.5.103.29.mavericks.bottle.tar.gz
```

jacobsa:
Given that this appears to be a problem with v8 and not this formula, does it make sense to pull this now?

vladshablinsky:
>>Whoever merges this should add "google-js-test" => "gjstest"

>Does doing renames for a formula that won't necessarily exist on a user's machine work/not throw weird errors?

<strike>@DomT4 Which one won't necessarily exist?
`google-js-test` shouldn't exist, but we get `gjstest` from the update with `google-js-test` => `gjstest`. Does it answer your question?</strike>

I read the comments again. @dunn are you suggesting removing a formula from tap? If yes, then it won't work. `formula_renames.rb` is for core formula migrations **only** and `formula_renames.json`s in taps are **only** for tap formula renames.

AFAIU, you're suggesting something like `homebrew/science/google-js-test -> homebrew/homebrew/gjstest` which is migration from one tap to another, actually (from science to core) and we can't do that.

jacobsa:
For what it's worth, the old formula hasn't built correctly in quite a long time. If it requires bending over backwards to handle it gracefully, it may not be worth it?

DomT4:
@BrewTestBot test this please

DomT4:
@vladshablinsky Yeah, the current formula as such is currently in `head-only` as `google-js-test`, and we can't assume everyone in brewland has the `head-only` tap tapped.

> AFAIU, you're suggesting something like 
homebrew/science/google-js-test -> homebrew/homebrew/gjstest which is migration from one tap to another, actually (from science to core) and we can't do that.

:+1:. We'll just remove the old formula in `head-only` then and let people find the new one manually. Thanks for the quick help here Vlad! Appreciated.

@jacobsa Still happy to accept this PR here, just per the caveat above that people will have to hunt down the new name manually. I've requeued it now I've merged fresh V8 bottles.

dunn:
:fireworks: d02d72f830e3b1cdf97ae9270da6b7d030e6236b :fireworks:

dunn:
Removed google-js-test from head-only, so everything's :100:

jacobsa:
Thank you!

PR #43465, opened 2015-09-01
pdns 3.4.6

justinclift:


dunn:
Any chance of adding a test while you're at it? We'd be very grateful!

justinclift:
Hmmmm, it could be tricky in this instance.  At a guess, the daemon would need to start up binding to an unused port, and a separate process would then need to confirm it's working and kill that daemon.

Not really sure if I can be bothered (I'm not a Ruby coder, and have no wish to be).

DomT4:
Merged as-is in https://github.com/Homebrew/homebrew/commit/e162363394f267d70a4bdc3ee66dcdb78da9533e. It'd be great to try and drum up a test in a subsequent PR if possible. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #43466, opened 2015-09-01
FORD 4.2.3

zbeekman:
FORD 4.2.3

Major enhancement:
 - Call graphs
 - Module dependency graphs
 - Type/class graph/diagram

Also, I have reordered the python package dependencies/resources now that [Homebrew PyPi Poet](https://github.com/tdsmith/homebrew-pypi-poet) sorts them in alphabetical order. This will ease maintenance in the future. When the formula for FORD was first authored, `poet -f ford` produced resource stanzas that were neither sorted, nor in a deterministic order. This made inspecting the stanzas for changed versions more time consuming.

### DIFF
--- a/Library/Formula/ford.rb
+++ b/Library/Formula/ford.rb
@@ -1,8 +1,8 @@
 class Ford < Formula
   desc "Automatic documentation generator for modern Fortran programs"
   homepage "https://github.com/cmacmackin/ford/"
-  url "https://pypi.python.org/packages/source/F/FORD/FORD-4.1.0.tar.gz"
-  sha256 "ec04b08468f75ac0acbda4fd6a15a432eb267e9c28098b5937125a95d2c6fdff"
+  url "https://pypi.python.org/packages/source/F/FORD/FORD-4.2.3.tar.gz"
+  sha256 "8dac3abec4ea88ff68218738bbd0407c2a0b1714c644e28883b1e36611c609b4"
 
   head "https://github.com/cmacmackin/ford.git"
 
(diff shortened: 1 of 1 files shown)


dunn:
The test is failling on all three builds:
```
==> /usr/local/Cellar/ford/4.2.3/bin/ford /private/tmp/ford20150901-70562-8dt6h6/test-project.md
  File "/usr/local/Cellar/ford/4.2.3/libexec/bin/ford", line 5, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 2793, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 673, in require
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 576, in resolve
    raise DistributionNotFound(req)
pkg_resources.DistributionNotFound: graphviz
Error: ford: failed

zbeekman:
ah, I forgot to add the new python graphviz binding dependency to the `resource(r).stage do` and also add graphviz itself as a dependency

zbeekman:
@dunn I can't seem to determine if my latest commits have been queued up by the testbot or not... It's been a while since I pushed the fixes...

dunn:
If you pushed again, it's been queued, but we've still got a bit of a backlog so the delay isn't surprising.

zbeekman:
ah OK, I tried to search the queue but couldn't find any evidence of it... It seems that all is well now, and the tests passed

dunn:
Looks good! Merged in 2a5243d28e2a3719f385d4c11f6073e6d0be2878, thanks!

PR #43468, opened 2015-09-01
update: report tap updates and improve wording

UniqMartin:
I feel like the information printed by `brew update` about the pulled commits could be slightly more detailed. These are the scenarios I had in mind when creating this change (and what would have been previously printed):

- The main repository and all taps were already up-to-date (no change):
  ```
  Already up-to-date.
  ```

- The main repository and/or taps had changes, but without user-visible changes to formulae:
  ```
  Updated Homebrew from e4920e76 to a6fc8812.
  No changes to formulae.
  ```
  (Previously, this would only show `Already up-to-date.` without any indication of a change.)

- The main repository and/or taps had changes, including user-visible changes to formulae:
  ```
  Updated Homebrew from 51a477de to 353e3573.
  Updated tap 'homebrew/science' from 7c00dbc9 to 37f6c041.
  Updated tap 'homebrew/versions' from c450e482 to 857638c6.
  Updated tap 'homebrew/x11' from 40f5a4d8 to 43f5c761.
  [the usual report about new/updated/renamed/deleted formulae]
  ```
  (Previously, there would be no indication that taps were updated unless that also involved a user-visible change to a formula. `Updated Homebrew from X to Y.` would be printed even if X and Y were the same commit, i.e. only taps changed.)

What do others think about this proposal?

### DIFF
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -39,12 +39,18 @@ module Homebrew
     report = Report.new
     master_updater = Updater.new(HOMEBREW_REPOSITORY)
     master_updater.pull!
+    master_updated = master_updater.updated?
+    if master_updated
+      puts "Updated Homebrew from #{master_updater.initial_revision[0, 8]} " \
+           "to #{master_updater.current_revision[0, 8]}."
+    end
     report.update(master_updater.report)
 
     # rename Taps directories
     # this procedure will be removed in the future if it seems unnecessasry
     rename_taps_dir_if_necessary
 
+    updated_taps = []
     Tap.each do |tap|
       next unless tap.git?
 
@@ -56,12 +62,18 @@ module Homebrew
         rescue
           onoe "Failed to update tap: #{tap}"
         else
+          updated_taps << tap.name if updater.updated?
           report.update(updater.report) do |_key, oldval, newval|
             oldval.concat(newval)
           end
         end
       end
     end
+    unless updated_taps.empty?
+      puts "Updated #{updated_taps.size} tap#{plural(updated_taps.size)} " \
+           "(#{updated_taps.join(", ")})."
+    end
+    puts "Already up-to-date." unless master_updated || !updated_taps.empty?
 
     Tap.clear_cache
 
@@ -109,9 +121,8 @@ module Homebrew
     end
 
     if report.empty?
-      puts "Already up-to-date."
+      puts "No changes to formulae." if master_updated || !updated_taps.empty?
     else
-      puts "Updated Homebrew from #{master_updater.initial_revision[0, 8]} to #{master_updater.current_revision[0, 8]}."
       report.dump
     end
     Descriptions.update_cache(report)
@@ -315,6 +326,10 @@ class Updater
     end
 
     map
+  end
+
+  def updated?
+    initial_revision && initial_revision != current_revision
   end
 
   private


Library/Homebrew/cmd/update.rb
```diff
@@ -92,10 +95,25 @@ def update
       end
     end
 
-    if report.empty?
+    # Report changes to Homebrew itself and tapped Git repositories.
+    repo_changes.reject! { |_key, r| r[0] == r[1] }
```
bfontaine: It would be faster to check for this *before* adding a change in the hash instead of adding everything then removing unneeded stuff.

bfontaine: Or you could skip them in the `.each` just below:

```irb
repo_changes.each do |tap, change|
  next if change[0] == change[1]
```

xu-cheng:
I think the first and second scenarios can be helpful, but I would worry the third one being too spammed especially for users who install dozen of taps.

Library/Homebrew/cmd/update.rb
UniqMartin: Thanks for the feedback! I believe the penalty paid by using `.reject!` is negligible, given that the number of taps will always be a moderate number and given that `pull!` and `report` from `Updater` are much more heavyweight (spawning `git` several times, network access, etc.). Thus, I'd like to keep it for the sake of code clarity and elegance.

The first option (checking before adding an unwanted change) would work, but would make the code uglier, as I would have to replicate the logic in two places and would have to store the result of `.revision_range` in a temporary variable.

I cannot use the second option (skipping inside `.each`) as I rely on `repo_changes.empty?` later on.

UniqMartin:
Thanks for the feedback, @xu-cheng! I think the situation for people with a lot of taps is not that bad:
- There will only be an output if the tap was actually updated.
- An updated tap will almost always mean additional entries in the changed formulae report printed below, making those first few lines probably negligible.

If the consensus turns out to be that this is indeed too verbose, maybe this behavior could be made optional (activated by `--verbose`) and the default would be to print something like `Updated 3 taps.`?

Library/Homebrew/cmd/update.rb
bfontaine: Right.

UniqMartin:
I pushed an update with improved code (better variable names, slightly more terse).

But more importantly, the updated code uses an array instead of a hash to track the changes. There's really no need for a hash here and the output now always matches the order of the `git pull` operations (with `Updated Homebrew from X to Y.` always being printed first).

xu-cheng:
I would still like to have more compact output. How about this:
* When everything is update to date, it shows `Already up-to-date.`
* When Hombrew is update to date but not taps, it shows `Updated 3 taps.`
* When taps are update to date but not homebrew, it shows `Updated Homebrew from e4920e76 to a6fc8812.`
* Otherwise, it shows `Updated Homebrew from e4920e76 to a6fc8812. (Also updated 3 taps)`

Library/Homebrew/cmd/update.rb
```diff
@@ -22,9 +22,11 @@ def update
     migrate_taps
 
     report = Report.new
+    repo_changes = []
     master_updater = Updater.new(HOMEBREW_REPOSITORY)
     master_updater.pull!
     report.update(master_updater.report)
+    repo_changes << ["homebrew/homebrew", *master_updater.revision_range]
```
bfontaine: I don’t think it’s necessary to expand the second part, i.e. you can have `[name, [a, b]]` instead of `[name, a, b]`.

`a, b, c = [1, [2, 3]]` doesn’t work as expected in Ruby, but it’s better not to use multiple assignments anyway because they create an array behind the scene.

Library/Homebrew/cmd/update.rb
```diff
@@ -91,10 +94,23 @@ def update
...
 
+    # Report changes to Homebrew itself and tapped Git repositories.
+    repo_changes.reject! { |change| change[1] == change[2] }
+    repo_changes.each do |change|
+      tap_name, rev_from, rev_to = change
+      tap_label = if tap_name == "homebrew/homebrew"
+        "Homebrew"
+      else
+        "tap '#{tap_name}'"
+      end
```
bfontaine: It’s probably nitpicking, but you can use a ternary operator here.

DomT4:
I'm with Xu on this. Having a line output for every tap tapped is going to get a little tedious quite quickly.

Library/Homebrew/cmd/update.rb
UniqMartin: Thanks for the hint. I was trying to stay below my personal limit of 80 columns (though Homebrew frequently breaks this limit), otherwise I would have used the ternary operator. I'm really not sure what the more readable variant is …

Library/Homebrew/cmd/update.rb
UniqMartin: I would have expected `a, b, c = [1, [2, 3]]` to become `a = 1`, `b = [2, 3]`, and `c = nil` and apparently that's also the case. Anyway, thanks for the hint. I'll try to stay away from multiple assignment and find a cleaner solution once I update the PR.

Library/Homebrew/cmd/update.rb
bfontaine: If the ternary operator breaks the limit then stay on the `if`/`else`/`end`.

UniqMartin:
@xu-cheng, @DomT4 Are you in support of a solution that would produce the following output?

- A mention of updated taps while still being terse (suggested by Xu) for `brew update`.
- Detailed report of every updated repository (suggested by me) for `brew update --verbose`.

I'm asking because the code for supporting both cases would be quite different from the code that would support the terse version only.

Just a final note: Besides making the output a bit clearer and acknowledging the existence of taps, my secondary aim was to output the old/new commit hashes of all updated repositories, so that interested users could feed them to `git log`, `git diff` et al. for further inspection. (I know that this information is also available via `git reflog`.)

bfontaine:
> Just a final note: Besides making the output a bit clearer and acknowledging the existence of taps, my secondary aim was to output the old/new commit hashes of all updated repositories

Should we hide that behind a `--verbose` option?

DomT4:
@UniqMartin Xu's suggestion seems reasonable to me, and generally in line with what you want to achieve. I'm fine with the `--verbose` option including the lengthier output.

I think we need to work on the wording a bit, somehow, but I'm not really sure how to make the non-verbose wording sound neither clunky or throwaway. Maybe something like `Updated Homebrew from e4920e76 to a6fc8812. All taps updated.` but that's not great either.

UniqMartin:
So I pushed an update, that implements both the terse and the detailed output. For the terse output I went with the following wording (depending on the situation):

- `Updated Homebrew and N taps.`
- `Updated Homebrew.`
- `Updated N taps.`
- `Already up-to-date.`

My initial idea of using `--verbose` to trigger the more detailed output didn't really work as it is already used to enable all kinds of other output (e.g. the output of all those `git` commands issued behind the scenes). Thus I picked `--detailed` for now, but maybe someone will come up with a better name. :)

Finally, I restructured the code quite a bit (again). I believe it to be much cleaner now, but since I'm still relatively new to Ruby, I'm more than happy about feedback.

xu-cheng:
Can we keep the revision change for Homebrew update because it usually reflects the core library change.

UniqMartin:
I removed the revision change from the terse output because I personally disliked the imbalance in verbosity between the master repository and taps (previously not mentioned at all, now only as a count). It also made the wording more awkward. After all the point of this PR was to make taps more visible.

But I can totally see that it would be useful to keep the revision change for Homebrew, as it is not only a repository of formulae, but also the repository with the code working behind the scenes.

---

Here is my updated suggestion of how the output could look like for the various scenarios:

- All repositories already up-to-date:
  ```
  Already up-to-date.
  ```

- Only updates to the master repository:
  ```
  Updated Homebrew from 51a477de to 353e3573.
  ```

- Only updates to one or more taps:
  ```
  Updated 3 taps.
  ```

- Updates to both the master repository and one or more taps:
  ```
  Updated Homebrew from 51a477de to 353e3573.
  Updated 3 taps.
  ```

If detailed output is enabled, `Updated 3 taps.` would be replaced with:

```
Updated tap 'homebrew/science' from 7c00dbc9 to 37f6c041.
Updated tap 'homebrew/versions' from c450e482 to 857638c6.
Updated tap 'homebrew/x11' from 40f5a4d8 to 43f5c761.
```

Unless all repositories were already up-to-date, this would be followed by either `No changes to formulae.` or the usual report about new/updated/renamed/deleted formulae.

mikemcquaid:
Personally I liked the original format. One of the things I really hate about `brew update` is that it churns for such a long time if you have many taps without any output.

mikemcquaid:
The newly proposed output is definitely an improvement though. @UniqMartin are these all output at the end or in between updates? It'd be nice if there was some incremental output after each stage (if that's not too gnarly).

UniqMartin:
> Are these all output at the end or in between updates? It'd be nice if there was some incremental output after each stage (if that's not too gnarly).

@mikemcquaid Currently, all the output is happening at the very end, right before dumping the formula report. But now that you mention it, it would make a lot of sense to output those messages as soon as the update has happened, i.e. to be more precise:

- Directly after the master repository has been updated: `Updated Homebrew from A to B.`
- In terse mode, directly after all taps have been updated: `Updated N taps.` (Assuming it is okay to “waste” another line in terse mode to report updates taps in a separate line. See my most recent proposal and previous iterations.)
- In detailed mode, immediately after a tap has been updated: `Updated tap '<tap>' from A to B.` 

All of this could happen before performing tap migrations and migrations due to renames. Code-wise, I don't think it is a big problem. It would certainly improve the perceived responsiveness of `brew update`.

I think I want to wait a day or two for more feedback to avoid too many complete rewrites of my code. :wink:

mikemcquaid:
> I think I want to wait a day or two for more feedback to avoid too many complete rewrites of my code. 

Good idea :+1:. Approach sounds great to me.

xu-cheng:
> Currently, all the output is happening at the very end, right before dumping the formula report. But now that you mention it, it would make a lot of sense to output those messages as soon as the update has happened, i.e. to be more precise:

@UniqMartin FYI, I plan to make `brew update` supporting multithread. If you are interesting, the on-going work is [here](https://github.com/Homebrew/homebrew/pull/40745). However, I haven't figure the output (especially during interrupt) under multithread mode yet. Another problem is that `brew update` could be most critical brew command and I'm afraid that multithread would make it more vulnerable.

UniqMartin:
Sadly, no more feedback. Here's the new version where the output happens as soon as possible. This change actually simplified the code. This time I also updated the man page and Bash completion.

@xu-cheng Thanks for pointing that out! Your multithreading work is interesting and I'll have a closer look as soon as I find the time. As for the output, I guess one would have to capture it in every thread and then use a queue (or some other thread-safe data structure) to push meaningful chunks of it to the main thread for printing on screen.

UniqMartin:
Like before, except for an updated `zsh` completion. Anything else I can do to move this PR forward?

UniqMartin:
@bfontaine @DomT4 @mikemcquaid @xu-cheng Sorry for pestering you with an explicit mention, but I thought it warranted due to the long silence (no progress in over a month). Is there anything I can do to move this PR forward or do I simply have to be more patient? Thanks!

mikemcquaid:
@UniqMartin No worries. I've been on vacation but I'll try to test this out and merge  (if none of the others object) or provide further comments later this week.

mikemcquaid:
Implementation and local testing look great to me. Last thing I think would be to kill the `--detailed` argument. Although I think it should be the default it seems I'm in the minority and I doubt (m)any people (myself included) are going to manually run `brew update --detailed`, unfortunately. After that I'll :ship:.

UniqMartin:
I agree that adding `--detailed` to every invocation of `brew update` doesn't sound like something many people would remember to do, even if `bash`/`zsh` completion significantly helps with this. Would it be more acceptable to use an environment variable (e.g. `HOMEBREW_UPDATE_VERBOSE`) instead of the argument to trigger the detailed output? That way people like me, who prefer such a level of detail, could enable it once in their `.bashrc` and then forget about it.

Alternatively, a compromise between terseness and usefulness might be to output something like `Updated 3 taps (homebrew/science, homebrew/versions, homebrew/x11).` I think that's inferior to my first suggestion, but still better than having no way to see which taps were updated at all.

I'd be truly sad to kill the functionality behind `--detailed` completely. I believe just mentioning the number of updated taps is almost as useless as the current state of not mentioning tap updates at all. But if I see the taps that have changed during `brew update` I can simply switch to one that interests me and inspect its recent history with `git log` and friends.

mikemcquaid:
> Alternatively, a compromise between terseness and usefulness might be to output something like Updated 3 taps (homebrew/science, homebrew/versions, homebrew/x11). I think that's inferior to my first suggestion, but still better than having no way to see which taps were updated at all.

That seems like a nice middle ground to me. @xu-cheng @bfontaine @DomT4 thoughts?

DomT4:
I agree people aren't going to remember to use `--detailed` a lot of the time. The verbose output for `update` already outputs quite a bit and I'm not sure it wouldn't be confusing to have two verbose options, but it would be nice to retain the functionality currently available under `--detailed` if there's an easy and agreeable way to do that.

> That seems like a nice middle ground to me

:+1:

mikemcquaid:
Sounds to me no-one objects to that option, let's go for it!

UniqMartin:
Pushed the “nice middle ground” version. Now the output is something like:

```
Updated Homebrew from 51a477de to 353e3573.
Updated 3 taps (homebrew/science, homebrew/versions, homebrew/x11).
==> New Formulae
[…]
```

Not exactly the outcome I was hoping for, but oh well … It is still an improvement.

mikemcquaid:
Thanks @UniqMartin!

PR #43470, opened 2015-09-01
fluent-bit 0.1.0 (new formula)

cosmo0920:


Library/Formula/fluent-bit.rb
```diff
@@ -0,0 +1,20 @@
+class FluentBit < Formula
+  desc "Data Collector for IoT"
+  homepage "https://github.com/fluent/fluent-bit"
+  url "https://github.com/fluent/fluent-bit/archive/0.1.0.tar.gz"
+  sha256 "5d77c75b1261695ca7f7b4752f5563b80c63c9d68f89fa6f5a1bfcd197a2a785"
+
+  head "https://github.com/fluent/fluent-bit.git", :branch => "master"
```
DomT4: You shouldn't need the `, :branch => "master"` bit here? We assume `master` is the upstream branch unless told otherwise.

cosmo0920: Ah..., Yes.
This repo's default branch is` master`.
So, I'll remove it.

Library/Formula/fluent-bit.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", ".", "-DWITH_IN_MEM=OFF", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    output = shell_output("fluent-bit --version")
```
DomT4: Any chance we can improve this test at all?

cosmo0920: Ah, yes we can.
I rewrite test which use `fork { .. }` and sending TERM signal.

Library/Formula/fluent-bit.rb
```diff
@@ -0,0 +1,22 @@
...
+  head "https://github.com/fluent/fluent-bit.git"
+
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", ".", "-DWITH_IN_MEM=OFF", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
```
DomT4: Can you write this like [this](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mpd.rb#L127-L139)? The `begin` and `ensure` parts are useful, Thanks.

xu-cheng: Also you need to add some test after server is spawned, otherwise the test will always success.

cosmo0920: Unfortunately,  `Fluent Bit` can not communicate via HTTP for now.
I rewrote test with `IO.popen` and confirming version from its output log.

Library/Formula/fluent-bit.rb
```diff
@@ -0,0 +1,23 @@
...
+    io = IO.popen("#{bin}/fluent-bit --input stdin --output stdout")
+    sleep 5
+    Process.kill("SIGINT", io.pid)
+    Process.wait(io.pid)
+    io.read =~ /Fluent-Bit v#{version}/
```
mikemcquaid: Please use `assert_match` here, thanks!

mikemcquaid: Also, does it need 5 seconds to output the version?

cosmo0920: Ah, no it dosen't.
I shortened sleep time here.

PR #43473, opened 2015-09-01
fix style problems in migrator

vladshablinsky:
Fix https://github.com/Homebrew/homebrew/issues/42965 style problems.

cc @mikemcquaid @xu-cheng


xu-cheng:
:+1:

xu-cheng:
Thanks.

PR #43477, opened 2015-09-01
consul-web-ui 0.5.2 (new formula).

srikalyan:
There is a cask consul-web-ui but the problems with it are
1. consul-web-ui is zip not a mac app so it should not be a cask formula
2. Does not provide a symlink to the one that is going to be used. This is
useful for writing scripts to perform automatic installation and configuration
of consul and consul-web-ui.


srikalyan:
I've never written ruby before so please let me know if there are any issues with it.

srikalyan:
Fixed the build issue.

srikalyan:
@dunn, please remove the label "new formula" as @tdsmith suggested to me irc that it should be part of the consul formula so I've added it to the consul formula and removed the consul-web-ui formula.

Library/Formula/consul.rb
```diff
@@ -13,6 +13,8 @@ class Consul < Formula
     sha256 "67052157266d536598ba7eb5af31a10fd3e609ca812be6d50869d0fe80a583f9" => :mountain_lion
   end
 
+  option "with-web-ui"
```
dunn: Could you add a second parameter with a description of what the option does?  It can me something simple; like:
```ruby
option "with-web-ui", "Build GUI"
```

srikalyan:
Done. Sorry, I am really new to ruby and homebrew land and the example I saw had no description so did not add one. It is added now.

Please let me know if you need anything else.

Library/Formula/consul.rb
srikalyan: Done. Please let me know if you need anything else.

Library/Formula/consul.rb
```diff
@@ -13,6 +13,8 @@ class Consul < Formula
     sha256 "67052157266d536598ba7eb5af31a10fd3e609ca812be6d50869d0fe80a583f9" => :mountain_lion
   end
 
+  option "with-web-ui", "Installs the consul web ui (to /usr/local/share/consul directory)"
```
DomT4: `/usr/local/share/consul `

This won't always be true. Although we encourage the use of `/usr/local` & our bottles are built for there you can actually stick Homebrew in any directory.

Library/Formula/consul.rb
```diff
@@ -174,5 +184,7 @@ def install
     end
     sleep 3
     system "#{bin}/consul", "leave"
+
+    assert_equal true, File.exist?("#{pkgshare}/index.html") if build.with? "web-ui"
```
DomT4: You can just do `assert File.exist` here. I think it's safe to cut the line entirely though; we never test non-default options.

srikalyan: Ok. I will remove it.

Library/Formula/consul.rb
srikalyan: This is a html/js/css app and it was suggested that it should be installed to share directory so I did pkgshare.install.

DomT4: The use of `pkgshare` is fine. It's just that `HOMEBREW_PREFIX` can be anything, it's usually `/usr/local` but you can make it `/opt/brew` or `~/homebrew` or `/Library/Shiny/brew` or whatever you like really, so the path could potentially be `/opt/brew/share/consul`.

It's probably best to leave the path out, we don't really include them in options.

srikalyan: I just wanted to give them that additional bit of info so that they don't have to read the formula to figure it out. Is it ok I say something like this "<prefex>/share/consul"?

srikalyan: or "HOMEBREW_PREFIX/share/consul"?
If you insist then I won't add it.

DomT4: My gut feeling is that it's easy enough to work out where it has been installed without needing to include that in the option, but I'll let some of the other maintainers comment on this if they'd like to.

srikalyan: I've removed it and added it to caveat and I hope it makes everyone happy :).

tdsmith:
I suggested that the caveat text should not be conditional on `build.with? "web-ui"` because that makes it much harder to access before and after installation (e.g. through `info`).

DomT4:
I'm cool with it if Tim is. Merged in https://github.com/Homebrew/homebrew/commit/eee136f5bec5dcbdf9cc4235050b940fbf19cfe5. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

srikalyan:
Thank you guys (esp @tdsmith) for helping me out. I did not knew much about inner workings of homebrew before this (as I quite new to mac world) but I could easily get things done with your help both on IRC and documentation. It definitely encourages me to contribute more.

PR #43480, opened 2015-09-01
Add missing build step to boost formula

aw1621107:
See: http://lists.boost.org/boost-users/2014/12/83414.php and followup

--HEAD --with-mpi --without-single builds were failing with a missing
header sometime when compiling the threads component of the library.
Turns out another command was added to the build sequence, but the docs
weren't updated. Now, instead of

  ./bootstrap.sh
  ./b2

the proper build commands should be

  ./bootstrap.sh
  ./b2 headers
  ./b2


aw1621107:
```
==> brew uses boost
Error: undefined method `has_apple_developer_tools?' for OS::Mac:Module
```
Well that's fun... Anything I could do about that?

DomT4:
@aw1621107 You may need to ensure the Homebrew your branch was based off of is completely up-to-date.

aw1621107:
Alright, updated, I think.

If the merge commit is a problem I can resubmit a clean PR. Just thought since PRs are not directly merged that the merge commit wouldn't be much of an issue, and it lets me keep using this PR instead of making a dupe. If not, let me know.

Library/Formula/boost.rb
```diff
@@ -119,6 +119,7 @@ def install
     end
 
     system "./bootstrap.sh", *bootstrap_args
+    system "./b2 headers"
```
mikemcquaid: Should be `system "./b2", "headers"`

mikemcquaid: Actually, I'll just fix it up locally.

mikemcquaid:

mikemcquaid:

For future reference the preferred commit message format is `boost: add missing build step.`.

aw1621107:
Should I still split up the "./b2 headers" line? Doesn't seem it's been split in the official repo

DomT4:
I'll fix it in a sec.

DomT4:
https://github.com/Homebrew/homebrew/commit/93d7446483660f375271a469a09cb485b13ef4c2

PR #43482, opened 2015-09-01
Update ImageMagick head URL to official git repo

aw1621107:
The svn url no longer seems to work, failing with an HTTP 500 internal
server error and/or HTTP 403 forbidden. Cannot check out the latest
revision both by hand and through brew.

Site says that https://git.imagemagick.org/repos/ImageMagick is the
official source repo, so this points brew thataway.

http://www.imagemagick.org/script/install-source.php


xu-cheng:

PR #43486, opened 2015-09-01
Rocker compose 0.1.0 (new formula)

ybogdanov:


Library/Formula/rocker-compose.rb
```diff
@@ -0,0 +1,15 @@
+class RockerCompose < Formula
+  desc "Docker tool for deploying apps composed of multiple containers."
+  homepage "https://github.com/grammarly/rocker-compose"
+  url "https://github.com/grammarly/rocker-compose/releases/download/0.1.0/rocker-compose-0.1.0_darwin_amd64.tar.gz"
```
DomT4: This needs to be built from source, not using the binary. You can see an example of building Go projects from source [here](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/gauge.rb).

DomT4:
Presuming this is dead from the lack of response, but please feel free to resubmit a fresh from-source PR if still interested. Thanks for the submission!

PR #43487, opened 2015-09-01
Make openal-soft keg only

tomyun:
OS X already provides `OpenAL.framework` and many existing packages pick that up without much problem. But `openal-soft` has same (with some extensions?) header and library signatures that precede the framework. For example, `sfml` can be successfully built without `openal-soft`, using native `OpenAL.framework`.

I suggest make `openal-soft` keg only and any dependency on that optional.

### DIFF
--- a/Library/Formula/openal-soft.rb
+++ b/Library/Formula/openal-soft.rb
@@ -3,6 +3,7 @@ class OpenalSoft < Formula
   homepage "http://kcat.strangesoft.net/openal.html"
   url "http://kcat.strangesoft.net/openal-releases/openal-soft-1.16.0.tar.bz2"
   sha256 "2f3dcd313fe26391284fbf8596863723f99c65d6c6846dccb48e79cadaf40d5f"
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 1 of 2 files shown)


dunn:
Looks like the test will need to be tweaked:
```
==> /usr/bin/clang test.c -lopenal
test.c:1:10: fatal error: 'AL/al.h' file not found
#include "AL/al.h"
         ^
1 error generated.

tomyun:
Forgot to add keg-only `include` and `lib` for the test. It's now updated.

DomT4:
Neither of these changes can be made without revisioning both formula, otherwise you'll cause breakage.

tomyun:
@DomT4 Good point. Now `revision` added for both formulae.

DomT4:
World's smallest nit, but please leave the version number off any non-version changes. Without any version bump should just be `dave: we had tea` or similar, Thanks :speak_no_evil:.

Merged in https://github.com/Homebrew/homebrew/commit/e6db6dae1df27e1b2991b1358baffbb0ca780fe7. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #43488, opened 2015-09-01
cdrtools 3.01 (stable)

fbrosson:
Update cdrtools (stable) to 3.01.
Drop devel until a version newer than 3.01 (stable) is released. This seems to be required to pass "brew audit cdrtools".


dunn:
Thanks! Merged in 0d0f3cb8aff36d3d0d39b34e7e0775b6c128d0b5 🎈

PR #43489, opened 2015-09-01
change oniguruma url

msurovcak:
As original URL is not available at a time I guess switching to Fedora's "[mirror](http://pkgs.fedoraproject.org/cgit/oniguruma.git/)" is safe (same hashsum) and more robust than geocities.jp


dunn:
Is the original URL permanently down? If it's not then maybe we should leave it and use the Fedora source as a `mirror`.

DomT4:
Closing in favour of #43581.

PR #43490, opened 2015-09-01
dcd 0.7.0

cspencer51:
Update DCD to latest version.


dunn:
```
        Error: 1 problem in 1 formula
dcd:
 * devel version 0.7.0-rc1 is older than stable version 0.7.0
```

You can remove the entire `devel` block, unless there's a newer devel version.

cspencer51:
Ah, that makes sense. Thanks for letting me know. First time contributing to Homebrew; still learning the ropes.

xu-cheng:

PR #39979, opened 2015-05-22
nghttp2: optionally enable python3 bindings

fwiesel:
This optionally enables the python 3 bindings of the nghttp2 library.
It should still work with python2, when with-docs are requested. 

The patch is required to detect the linker flags properly under Mac OS X and python 3.4.
I'll try to upstream it, as soon as I find out, where in the automake chain that actually gets generated.

PR #43491, opened 2015-09-01
nghttp2: fix Python bindings

tdsmith:
Mixing python2 (which we need for docs) and python3 (for which we
install bindings) is too weird, so tell nghttp2 not to worry about the
bindings and build them ourselves.

Also works around a bad call to setup.py install in the nghttp2
Makefile.

Fixes https://github.com/Homebrew/homebrew/pull/39979#issuecomment-136885203.


tdsmith:
thanks for the nudge, @RandomDSdevel; opening new issues is usually best.

RandomDSdevel:
@tdsmith:  Thanks, and you're welcome!

PR #43499, opened 2015-09-02
llvm 3.7.0

lembacon:


tamird:
looks this needs a creduce update in lockstep

bfontaine:
We need to check the following dependencies:

- [ ] cctools
- [ ] creduce
- [ ] emacs-clang-complete-async
- [ ] ldc
- [ ] ponyc
- [ ] rtags
- [ ] homebrew/python/numba
- [ ] homebrew/science/pocl

DomT4:
Yeah, a revision on the `creduce` formula isn't likely to cut it this time; major version bumps tend to need an actual upstream new release.

zeux:
Does it make sense to do homebrew/versions/llvm37 first?

Code that was compiled with LLVM 3.6 is unlikely to compile with LLVM 3.7 unless it used LLVM-C API, there are several breaking changes that will likely affect any frontend in existence.

chfast:
If the dependencies require LLVM 3.6 they should depend on llvm36, right?

dunn:
> If the dependencies require LLVM 3.6 they should depend on llvm36, right?

We don't do cross-tap dependencies in core, though (i.e., `depends_on "homebrew/versions/llvm36"`).  So adding `llvm37.rb` to homebrew-versions while we nudge developers to fix their software sounds like a good plan.

zeux:
PR for llvm37: https://github.com/Homebrew/homebrew-versions/pull/955

chfast:
@dunn, nice idea but that will simply not work in case of LLVM. Upgrading LLVM dependency is not "fixing software".

DomT4:
The discussion in https://github.com/rust-lang/rust/issues/28830 is interesting. Looks like it's worth waiting on 3.7.1 even if the dependents are happy.

DomT4:
Going to close here. Current status is more or less waiting for 3.7.1 to land. I may merge the Versions formula, with a caveat about the accidental API break, in the mean time for people desperate to use that. Thank you for the submission to Homebrew @lembacon :heart:

PR #43506, opened 2015-09-02
formula_versions: avoid disk io

xu-cheng:
Another take of #40911


xu-cheng:
Created `test_formulary.rb`, now the coverage for `formulary.rb` is 89.78 % (used to be 60.59 %)

xu-cheng:
cc @Homebrew/owners for comments.

mikemcquaid:
:+1:

PR #43507, opened 2015-09-02
daemon: Fix hosting for strlcpy patch.

jacobsa:
I've tried all day and haven't been able to get trac.macports.org to respond a single time, which means this formula won't install. Host it on GitHub instead.

I couldn't find a pristine copy of this patch, so I reconstructed it by hand from an HTML view in Google's [cache][] of the macports page. I guess I got it right byte for byte because the SHA-256 hash checks
out.

[cache]: http://webcache.googleusercontent.com/search?q=cache:V75bzw8DTMAJ:trac.macports.org/attachment/ticket/42845/daemon-0.6.4-ignore-strlcpy-strlcat.patch&hl=en&gl=au&strip=0&vwsrc=1

### DIFF
--- a/Library/Formula/daemon.rb
+++ b/Library/Formula/daemon.rb
@@ -6,7 +6,7 @@ class Daemon < Formula
 
   # fixes for mavericks strlcpy/strlcat: https://trac.macports.org/ticket/42845
   patch do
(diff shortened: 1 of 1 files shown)


dunn:
Nice work! Merged in d2d094e6372e5b4545aa2802c041907515fda0a8.

PR #43508, opened 2015-09-02
Added Formula for Qore Language

temnoregg:
Fixed and improved formula as defined here - https://github.com/Homebrew/homebrew/pull/42133

### DIFF
--- a/Library/Formula/qore.rb
+++ b/Library/Formula/qore.rb
@@ -1,8 +1,7 @@
 class Qore < Formula
   desc "Multi-threaded embeddable scripting language"
-  homepage ""
+  homepage "http://qore.org"
   url "https://github.com/qorelanguage/qore/releases/download/release-0.8.11.1/qore-0.8.11.1.tar.bz2"
-  version "0.8.11.1"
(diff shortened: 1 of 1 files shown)


dunn:
A few things caught by the audit; some you can fix, but the notability issue is one we'll have to think about:
```
==> audit problems
qore:
 * Stable: version 0.8.11.1 is redundant with version scanned from URL
 * The homepage should start with http or https (URL is ).
 * The homepage is not reachable (curl exit code 3)

temnoregg:
Updated as suggested by audit

bfontaine:
> but the notability issue is one we'll have to think about

FWIW the Wikipedia page about the language was deleted two years ago because it didn’t meet their notability guidelines (which aren’t the same as ours). It’s in the [OpenSUSE official repo](http://qore.org/47-news/208-qore-in-opensuse-official-repository) but not in Debian nor Fedora.

mikemcquaid:
Passing on this for the previously mentioned notability reasons, sorry. I suggest you create your own tap.

PR #43513, opened 2015-09-02
Add formula for puddletag.

chincheta0815:


mistydemeo:
It looks like the test is running:

> 2015-09-02 17:35:34.253 Python[31835:1c0b] Persistent UI failed to open file file://localhost/Users/brew/Library/Saved%20Application%20State/org.python.python.savedState/window_3.data: No such file or directory (2)

chincheta0815:
puddletag is tools for changing mp3 tag in a guy. Since puddle tag does not have any test routine I added  system "puddletag" in the test part of the formula.
Running from the command line brew test puddle tag simply opens this gui.

I interpret the test result as there is no possible way to open that gui.

Do you have a suggestion for me where to get some info on solving the issue?

chincheta0815:
Now runs without errors.
The test was changed to not trying to open a gui.

chincheta0815:
Sorry, stupid me. Wrong klick...

PR #43517, opened 2015-09-02
caveats: wrap elisp caveats

dlh:


Library/Homebrew/caveats.rb
```diff
@@ -136,11 +136,12 @@ def app_caveats
   def elisp_caveats
     return if f.keg_only?
     if keg && keg.elisp_installed?
-      <<-EOS.undent
+      <<-EOS.undent_________________________________________________________72
```
DomT4: Can we not just manually do this without adding the giant `_________________________________________________________72` line?

dlh: 🆗

xu-cheng:

PR #43518, opened 2015-09-02
always use opt_prefix in install_name

tdsmith:
Opening this for feedback; are there downsides?

dylib install_names are presently rewritten relative to HOMEBREW_PREFIX,
except for keg-only formulas, which are rewritten relative to
opt_prefix.

If we make a formula keg_only, all of its dependents need to be bumped,
since it will no longer exist in HOMEBREW_PREFIX and all of the links
will be written incorrectly.

If we always use opt_prefix, we avoid that dilemma, and make installed
packages more robust if a dependency is `brew unlink`ed. (Is that a feature or a bug?)

### DIFF
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -137,13 +137,7 @@ class Keg
     # the basename of the file itself.
     basename = File.basename(file.dylib_id)
     relative_dirname = file.dirname.relative_path_from(path)
-    shortpath = HOMEBREW_PREFIX.join(relative_dirname, basename)
-
-    if shortpath.exist? && !options[:keg_only]
-      shortpath.to_s
-    else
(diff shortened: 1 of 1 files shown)


DomT4:
I presumed there was a reason we didn't do this, but if there isn't, I'm in favour. The OpenSSL issue a couple months back was a prominent reminder of how much pain can be caused by having to make things `keg_only` and building bottles against the non-kegged version beforehand.

mistydemeo:
I think this makes a lot of sense, yeah. It should also make things more robust against users accidentally overwriting the `/usr/local/lib` libs with something they've installed manually.

xu-cheng:
:+1:

mikemcquaid:
Makes sense to me :+1:

xu-cheng:
any news?

UniqMartin:
I can see a lot of benefits and no downsides. :+1:

PR #43521, opened 2015-09-02
The ASF now has a preferred .lua mirror script on top of the .cgi file

janl:


mikemcquaid:
Thanks @janl :sparkling_heart:!

PR #43524, opened 2015-09-03
git-annex: add head clause

FiloSottile:


bfontaine:
Thanks a lot @FiloSottile! :tada:

Issue #43528, opened 2015-09-03
Unable to install newt - trac.macports.com down


ktaragorn:
I need to update `byobu` right now, since I am facing issues with it after updating to python 2.7.10. Updating that requires me to update `newt` and updating `newt` hits https://trac.macports.org/export/132914/trunk/dports/devel/libnewt/files/patch-Makefile.in.diff which has been down for atleast the last one day and seems like this has happened before according to this site -  http://comments.gmane.org/gmane.os.apple.macports.user/37211

It is odd that brew is depending on macports like this, and effort should be made to make this process more reliable..

dunn:
Trac has been down for almost a week now.  If you can find a copy of the patch (maybe from a Google cache), we'd be happy to replace it with a gist-ed version.

ktaragorn:
Hey, I dont know enough to understand what is this or how it is used, but a google got me https://github.com/dstrctrng/macports/blob/master/rsync.macports.org/release/tarballs/ports/devel/libnewt/files/patch-Makefile.in.diff . I am not sure if this is a static file or changes with every update..

ktaragorn:
Tried replacing.. didnt work.

ktaragorn:
Found ftp://ftp.ca.freebsd.org/MacPorts/release/ports/devel/libnewt/files/patch-Makefile.in.diff gonna make a pr

dunn:
Awesome, thanks!

I didn't know Trac has a history of outages—we should try to move away from using patches hosted there.

ktaragorn:
PR made. http://www.filewatcher.com/m/patch-Makefile.in.diff.1393-0.html gave me the list of alternatives, in case this mirror also goes down in the future..

dunn:
Nice, I didn't know about FileWatcher.

ktaragorn:
I didnt either.. Google is great :P . Very handy in this case.. considering I still dont know what this patch is or how often it changes.. All the other results were 2-4 years ago and likely wrong.

PR #43529, opened 2015-09-03
Fix newt install formula - trac.macports.com down

ktaragorn:
More information in #43528

### DIFF
--- a/Library/Formula/newt.rb
+++ b/Library/Formula/newt.rb
@@ -19,7 +19,7 @@ class Newt < Formula
   # build dylibs with -dynamiclib; version libraries
   patch :p0 do
     url "ftp://ftp.ca.freebsd.org/MacPorts/release/ports/devel/libnewt/files/patch-Makefile.in.diff"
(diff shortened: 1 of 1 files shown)


ktaragorn:
Got this url from http://www.filewatcher.com/m/patch-Makefile.in.diff.1393-0.html.

Tested on my local, `sha1` matches and `newt` is able to install.

bfontaine:
Thank you! Could you please upgrade the `sha1` to a `sha256`?

ktaragorn:
Done @bfontaine

DomT4:
Fixed in https://github.com/Homebrew/homebrew/commit/79b0774dbf22853ffeedd36fabb2f6e974aa8400.

You missed the other SHA1 at the top of the formula which needed updating as well and I was reluctant to push it through another bot run given how backed up we've been the last few days. Apologies & Thanks for the submission and willingness to fix :heart:

ktaragorn:
thanks for fixing @DomT4 glad i could be of help

PR #43533, opened 2015-09-03
Install the latest stable version of npm

davepwsmith:


Library/Formula/node.rb
```diff
@@ -31,7 +31,7 @@ class Node < Formula
   end
 
   resource "npm" do
-    url "https://registry.npmjs.org/npm/-/npm-2.12.1.tgz"
+    url "https://registry.npmjs.org/npm/-/npm-2.14.1.tgz"
     sha256 "6b6512c6f9097da193dfe046053d6d0483b5c5658dc0a763c1ba5609b6bbc16c"
```
bfontaine: Please update the  `sha256` as well.

davepwsmith:
Ultimately would it not be more sensible to split npm from the node package and install separately as a dependency? This way both node and npm could be kept up to date using brew. I'm sure this discussion has been had before...

davepwsmith:
Anyhow, for now I have added a revised and more accurately named pull request.

PR #43535, opened 2015-09-03
[WIP] dependency prioritization 

xu-cheng:
TODO list:

* [ ] perfromance evaluation 
* [ ] add test case

### DIFF
--- a/Library/Homebrew/dependency.rb
+++ b/Library/Homebrew/dependency.rb
@@ -2,9 +2,21 @@ require "dependable"
 
 # A dependency on another Homebrew formula.
 class Dependency
+  class DependencyMergeConflict < RuntimeError
+    def initialize(formulae)
+      super <<-EOS.undent
+        Cannot merge dependencies, multiple conflicting formalae are requested.
+         * #{formulae.join("\n * ")}
+      EOS
+    end
+  end
+
   include Dependable
 
   attr_reader :name, :tags, :env_proc, :option_name
+
+  # @private
+  attr_accessor :owner_tap, :formula
 
   DEFAULT_ENV_PROC = proc {}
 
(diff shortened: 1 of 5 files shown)


xu-cheng:
cc @CNA-Bld

xu-cheng:
I just realize this approach has tons of problems. Investigating better solution.

Library/Homebrew/dependency.rb
```diff
@@ -2,9 +2,22 @@
 
 # A dependency on another Homebrew formula.
 class Dependency
+  class DependencyMergeConflict < RuntimeError
+    def initialize(formulae)
+      super <<-EOS.undent
+        Cannot merge dependencies, multiple conflicted formluae are requested.
```
mikemcquaid: `formluae` -> `formulae`

mikemcquaid: Also probably `conflicted` -> `conflicting`

mikemcquaid:
Approach makes sense to me so far :+1:

davydden:
@xu-cheng do you think you will have a chance to look into this PR in some near future? Would be a nice feature to have...

xu-cheng:
@davydden this is on my todo list. Current PR has performance issue. And I'm working another PR to address this issue. So it will take a period of time before this finally landing.

davydden:
@xu-cheng thanks for the prompt reply and the status update.

UniqMartin:
Could you please summarize what the current behavior is (i.e., prior to this PR) and how and why this PR changes this? (I think this should go into the first post.) I would like to understand more of the dependency resolution logic, but all of this stuff is not exactly self-documenting. Thanks!

xu-cheng:
## Current Homebrew's Behaviors 

* `depends_on` only invokes `Formulary.factory`. Therefore, for `depends_on "foo"`, it will look for core formula first, then in taps. If multiple tap formulae are found, `TapFormulaAmbiguityError` will be raised.
* `brew tap-pin` is only used to prioritize certain taps over core formula in `ARGV.formulae`. i.e. only help to offer shortcut for user's input. **Notice** this is the desired behavior. Because we believe formula resolution order by `depends_on` should be decided by formula authors instead of users.

## Desired Homebrew's Behaviors 

We need to change `depends_on "foo"`  to use following prioritization order:
* Look for formula inside its own tap. i.e. `homebrew/homebrew/foo` for core formulae and `some/tap/foo` for tap formulae.
* Look for formula inside core. This is mainly for tap formulae.
* Look for formula anywhere. A warning message will be shown.

## What need to be done(What holds this PR)

Current implementation has poor performance because we lack of tap information. I'm still working on passing tap information during loading formula code. And I'm still not satisfied with them. I'm hoping to find a delicate way to get formula file path during loading.

UniqMartin:
Thanks for this overview/status, @xu-cheng! That's massively helpful to me (and I'm sure to others, too).

So the reason the current system works quite well is that at least the official taps take care to use unique formula names and thus the search order usually doesn't matter. Will there still be a possibility for a tap formula like `some/tap/formula` to depend on a core formula if there's also one with the same name in the tap, e.g., by specifying `homebrew/homebrew/dependency` so that it is chosen before `some/tap/dependency` from the same tap?

mikemcquaid:
Agreed with the desired behavkours :+1:

MikeMcQuaid:
Can this be ported over to the Homebrew/brew repo when it's ready to 🚢? Thanks!

PR #43537, opened 2015-09-03
node: update npm and add revision.

davepwsmith:


Library/Formula/node.rb
```diff
@@ -31,8 +32,8 @@ class Node < Formula
   end
 
   resource "npm" do
-    url "https://registry.npmjs.org/npm/-/npm-2.12.1.tgz"
-    sha256 "6b6512c6f9097da193dfe046053d6d0483b5c5658dc0a763c1ba5609b6bbc16c"
+    url "https://registry.npmjs.org/npm/-/npm-2.14.1.tgz"
```
dunn: Is this the version shipped with 0.12.7?

DomT4: We don't care really. `npm` moves faster than `node` but remains compatible with Node & itself (at least across the 2.x releases). We apply newer `npm` versions from time to time for important fixes. I'm fine accepting this because there's a security issue in play, but we should use `2.14.2` instead.

davepwsmith:
Duly updated.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/562a5b1aed62525e96dc3788c620f98386031a4c. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #43541, opened 2015-09-03
Added --universal option for libdc1394

bleppie:
libdc1394 is often needed for both i386 and x86_64 on OSX. Different libraries, for example Cinder, openframeworks, and EMGU.opencv, use libdc1394 and might be compiled for different architectures.


dunn:
Thanks! Merged in 8278b897b5749e9ac8a365e5284c059668c60d77.

PR #43561, opened 2015-09-03
Added --dry-run functionality to unlink

SnoringFrog:
Added the ability to use `--dry-run`/`-n` with unlink to preview what links would be removed.

Came about from me trying to find a way to determine an easy way to figure out the commands associated with a formula. During that process I tried to use an `unlink --dry-run` only to be surprised when it actually unlinked everything lol.


DomT4:
You seem to have indented everything in this PR with tabs? We use a two-space indent rather than tabs.

SnoringFrog:
Whoops, thought I'd fixed that. Should be good now.

xu-cheng:

PR #43562, opened 2015-09-03
wimlib 1.8.2

cprecioso:
Updated version and changed to the new homepage


xu-cheng:

Issue #43082, opened 2015-08-19
InfluxDB 0.9.3 will attempt to write to /var/opt


victorhooi:
In the newer versions of InfluxDB (0.9.3 and upwards), the default configuration file generated will have a section like this:

```
  # The following WAL settings are for the b1 storage engine used in 0.9.2. They won't
  # apply to any new shards created after upgrading to a version > 0.9.3.
  max-wal-size = 104857600 # Maximum size the WAL can reach before a flush. Defaults to 100MB.
  wal-flush-interval = "10m" # Maximum time data can sit in WAL before a flush.
  wal-partition-flush-delay = "2s" # The delay time between each WAL partition being flushed.

  # These are the WAL settings for the storage engine >= 0.9.3
  wal-dir = "/var/opt/influxdb/wal"
  wal-enable-logging = true
```

Note the reference to `/var/opt` - in my case, when I try to actually write points to InfluxDB, I get:
```
[write] 2015/08/19 14:23:31 write failed for shard 1 on node 1: load metadata index: mkdir /var/opt: permission denied
```
I think the receipe simply needs to be adapter to regex out this path as well.

mikemcquaid:
Could you try and open a pull request? This should help and I'm happy to walk you through anything else: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged

Thanks!

DomT4:
Closed via https://github.com/Homebrew/homebrew/commit/f5e92cc8f9bf0935e9e3db238e89ef9641763a7b.

PR #43572, opened 2015-09-04
influxdb: inreplace config 'wal-dir' to user writable path

tzudot:
Addresses issue in #43082.

### DIFF
--- a/Library/Formula/influxdb.rb
+++ b/Library/Formula/influxdb.rb
@@ -105,6 +105,7 @@ class Influxdb < Formula
       s.gsub! "/var/opt/influxdb/data", "#{var}/influxdb/data"
       s.gsub! "/var/opt/influxdb/meta", "#{var}/influxdb/meta"
       s.gsub! "/var/opt/influxdb/hh", "#{var}/influxdb/hh"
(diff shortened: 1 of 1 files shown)


Library/Formula/influxdb.rb
```diff
@@ -105,6 +105,7 @@ def install
       s.gsub! "/var/opt/influxdb/data", "#{var}/influxdb/data"
       s.gsub! "/var/opt/influxdb/meta", "#{var}/influxdb/meta"
       s.gsub! "/var/opt/influxdb/hh", "#{var}/influxdb/hh"
+      s.gsub! "/var/opt/influxdb/wal", "#{var}/influxdb/wal"
```
DomT4: Presumably we want to add `(var/"influxdb/wal").mkpath` further down as well?

tzudot: Yup, missed that. I'll update it.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/f5e92cc8f9bf0935e9e3db238e89ef9641763a7b. Thank you for your contribution to Homebrew, we appreciate it! :smiley_cat:

PR #43578, opened 2015-09-04
lsyncd fails on OS X 10.11

DavidStaron:
https://gist.github.com/DavidStaron/4f40121d0e116316cfd6

### DIFF
--- a/Library/Formula/lsyncd.rb
+++ b/Library/Formula/lsyncd.rb
@@ -40,8 +40,8 @@ class Lsyncd < Formula
     "10.10.1"  => ["xnu-2782.1.97.tar.gz", "18fd93155c53fa66c48c2c876313311bba55cff260ea10e7b67dd7ed1f4b945c"],
     "10.10.2"  => ["xnu-2782.10.72.tar.gz", "0725dfc77ce245e37b57d226445217c17d0a7750db099d3ca69a4ad1c7f39356"],
     "10.10.3"  => ["xnu-2782.20.48.tar.gz", "75effef91dbeb7ab59da7a62936e37e30c2a2feca9ee653d3245b6a5b17ad2d7"],
-    "10.10.4"  => ["xnu-2782.1.97.tar.gz", "18fd93155c53fa66c48c2c876313311bba55cff260ea10e7b67dd7ed1f4b945c"],
-    "10.10.5"  => ["xnu-2782.1.97.tar.gz", "18fd93155c53fa66c48c2c876313311bba55cff260ea10e7b67dd7ed1f4b945c"],
+    "10.10.4"  => ["xnu-2782.20.48.tar.gz", "75effef91dbeb7ab59da7a62936e37e30c2a2feca9ee653d3245b6a5b17ad2d7"],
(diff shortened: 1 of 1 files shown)


DomT4:
It will be interesting to see if this doesn't cause problems. OS X 10.11 is actually currently `xnu-3247.1.106~5` which is rather a hike from `2782.1.97`.

Library/Formula/lsyncd.rb
```diff
@@ -38,7 +38,8 @@ class Lsyncd < Formula
     "10.9.5" => ["xnu-2422.115.4.tar.gz", "48207e250422be7e78d238cd8b4d741ac98856df"],
     "10.10"  => ["xnu-2782.1.97.tar.gz", "c99cf8ec04c29d40b771652241dd325e4977d92b"],
     "10.10.1"  => ["xnu-2782.1.97.tar.gz", "c99cf8ec04c29d40b771652241dd325e4977d92b"],
-    "10.10.2"  => ["xnu-2782.1.97.tar.gz", "c99cf8ec04c29d40b771652241dd325e4977d92b"]
+    "10.10.2"  => ["xnu-2782.1.97.tar.gz", "c99cf8ec04c29d40b771652241dd325e4977d92b"],
+    "10.11"  => ["xnu-2782.1.97.tar.gz", "c99cf8ec04c29d40b771652241dd325e4977d92b"]
```
mikemcquaid: Mind adding a `,` at the end of the line? Just makes future `diff`s cleaner.

mikemcquaid:
Would like to hold off on this until we have a 10.11 bot (should be soon).

xu-cheng:
Would be great to update all sha1 to sha256.

mikemcquaid:
> Would be great to update all sha1 to sha256.

:+1:

DomT4:
10.10.2 & 10.10.3 now have their `xnu` version published and tarballs available, so we can actually update a couple more point releases in this list as well now. Should be done for 10.10.2-10.10.5.

10.10.2 here is an assumed number because Apple was slow publishing the new tarballs, so that needs to become `xnu-2782.10.72`. For OS X point releases Apple hasn't published the open-source elements of yet just use the `xnu` release for the prior point release.

DomT4:
Side note: Apple's Open Source website is insecure at the moment, so things will fail to fetch here.

DomT4:
@mikemcquaid Any objection to this going in, minus the 10.11 addition for now if necessary? It currently doesn't build where there's not a `xnu` defined:

```
==> Patching
/usr/bin/patch -g 0 -f -p1 -i /private/tmp/lsyncd--patch20150912-29318-s6481d/cve-2014-8990.diff
patching file default-rsyncssh.lua
Error: lsyncd does not define resource "xnu"
```

DavidStaron:
Would it make any sense to leave out the 10.11 addition? I mean it's currently not building at all on 10.11, so it can't get any worse than that...?

mikemcquaid:
Sorry for the delay! Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #43583, opened 2015-09-04
Gprof2dot

zbeekman:
Two new formulae:
 1. [gprof2dot](https://github.com/jrfonseca/gprof2dot) - Python script to convert many profiling tools outputs into annotated call graphs
 1. [xdot](https://github.com/jrfonseca/xdot) - Interactive gui viewer for graphviz files.

N.B. I don't know how to write a good test for `xdot` since it is a GUI program and has no non-interactive uses.

### DIFF
--- a/Library/Formula/gprof2dot.rb
+++ b/Library/Formula/gprof2dot.rb
@@ -1,0 +1,180 @@
+class Gprof2dot < Formula
+  desc "Convert the output from many profilers into a Graphviz dot graph."
+  homepage "https://github.com/jrfonseca/gprof2dot"
+  url "https://pypi.python.org/packages/source/g/gprof2dot/gprof2dot-2015.02.03.tar.gz"
+  sha256 "9bdebb71a7c3daba1430de4486ef620e418a43bb051b30875d02f657f9c69d65"
+
+  head "https://github.com/jrfonseca/gprof2dot.git"
+
+  depends_on "graphviz" => :recommended
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir[libexec/"bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    (testpath/"gprof.output").write <<-EOS.undent
+      Flat profile:
+
+      Each sample counts as 0.01 seconds.
+       no time accumulated
+
+        %   cumulative   self              self     total
+       time   seconds   seconds    calls  Ts/call  Ts/call  name
+        0.00      0.00     0.00        2     0.00     0.00  manager
+        0.00      0.00     0.00        2     0.00     0.00  worker1
+        0.00      0.00     0.00        2     0.00     0.00  worker2
+        0.00      0.00     0.00        1     0.00     0.00  project1
+        0.00      0.00     0.00        1     0.00     0.00  project2
+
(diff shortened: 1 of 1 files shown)


Library/Formula/gprof2dot.rb
```diff
@@ -0,0 +1,1339 @@
...
+                      0.00    0.00    5232/9480        multi_glob [96]
+      [87]     0.0    0.00    0.00    9480         ar_name [87]
+      -----------------------------------------------
+                      0.00    0.00    9213/9213        hash_find_slot [11]
+      [88]     0.0    0.00    0.00    9213         dirfile_hash_cmp [88]
+      -----------------------------------------------
+                      0.00    0.00       1/9025        enter_command_line_file [82]
+                      0.00    0.00       1/9025        construct_vpath_list [79]
+                      0.00    0.00       5/9025        main [1]
+                      0.
```
bfontaine: Can’t we make a shorter test? :smile:

bfontaine:
Note that we generally prefer one PR per (new) formula.

Library/Formula/gprof2dot.rb
DomT4: Heh, I checked this thinking "How long can it possibly be". Now I know :wink:

zbeekman:
Sure, why don't I split into 2 PRs and find a shorter example...

zbeekman:
OK, I've split out xdot into its own PR (#43587) and used a shorter test. Still longish, but that is due to gprof's lengthy output.

zbeekman:
now only ~153 lines of gprof input

mikemcquaid:
Needs some audit tweaks: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/32676/version=yosemite/testReport/junit/brew-test-bot/yosemite/audit_gprof2dot___strict___online/

zbeekman:
gah, sorry... It should be resolved now... I'll try to keep my eye on the test result... I wish github could add failed tests to your notifications.

mikemcquaid:
Test is still failing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/32845/version=yosemite/testReport/junit/brew-test-bot/yosemite/test_gprof2dot/ Maybe needs to be told not to output UTF-8 or something?

zbeekman:
strange, that doesn't happen on my system... very annoying. I'll look into this some more

zbeekman:
OK, I've pushed what I think is a valid fix. This passes the audit and the test on my machine... Hopefully it will work on the testbot too.

zbeekman:
@mikemcquaid any idea why this might pass on mountain lion but not mavericks or yosemite? My home system is Yosemite and never has this problem...

zbeekman:
I know very little ruby, so I've taken a shot in the dark at fixing these errors by trying to force UTF-8 encoding for the shell ouput of the gprof2dot test. Fingers crossed this works. :pray:

zbeekman:
@dunn @mikemcquaid @DomT4 So, i'm completely stumped on why this fails, some of the time...

dunn:
`force_encoding` isn't available on Ruby 1.8; maybe you could use `iconv` instead?

zbeekman:
OK, new tact for the test: just use `assert File.exist?(...)` to check that the graphviz dot file was written, rather than producing the contents to stdout and then trying to match them to some expected output.

zbeekman:
Thank goodness... hope that test looks ok to you guys.

DomT4:
This seems fine to me now. Merged in https://github.com/Homebrew/homebrew/commit/cc31f79fbe3aa44fa337fb177a7a62380e55d962. Thank you for your contribution to Homebrew; we appreciate it! :smiley_cat:

zbeekman:
:bow: thanks!

PR #43588, opened 2015-09-04
tidy-html5 5.0.0

zmwangx:
Also updated test because 5.0.0 introduced a breaking change: `tidy5` has been renamed to `tidy`.

By the way, after the rename, the executable and libraries now override system `tidy`, which is the old HTML 4 `tidy` from 2006 (see `/usr/bin/tidy --version`). Still, `tidy-html5` should be mostly (if not entirely) backwards compatible, per https://github.com/htacg/tidy-html5/issues/221. So, do we need to do anything about this? Like making it keg-only or moving it to homebrew/dupes?

### DIFF
--- a/Library/Formula/tidy-html5.rb
+++ b/Library/Formula/tidy-html5.rb
@@ -1,8 +1,8 @@
 class TidyHtml5 < Formula
   desc "Granddaddy of HTML tools, with support for modern standards"
   homepage "http://www.html-tidy.org/"
-  url "https://github.com/htacg/tidy-html5/archive/4.9.35.tar.gz"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> By the way, after the rename, the executable and libraries now override system tidy, which is the old HTML 4 tidy from 2006 (see /usr/bin/tidy --version). Still, tidy-html5 should be mostly (if not entirely) backwards compatible, per htacg/tidy-html5#221. So, do we need to do anything about this? Like making it keg-only or moving it to homebrew/dupes?

Thanks for raising this. If it's backwards compatible I'm less worried. Let's merge as-is for now.

mikemcquaid:

PR #43591, opened 2015-09-04
qBittorrent's Homebrew formula.

ainstushar:
My 1st formula. For qBittorrent. Please let me know if I've screwed up.

Big thanks to @mistydemeo for spoon-feeding me these steps.

### DIFF
--- a/Library/Formula/qbittorrent.rb
+++ b/Library/Formula/qbittorrent.rb
@@ -1,0 +1,36 @@
+class Qbittorrent < Formula
+  desc "Qt based free and reliable P2P Bittorrent client"
+  homepage 'http://www.qbittorrent.org'
+  url 'https://github.com/qbittorrent/qBittorrent/archive/release-3.2.3.tar.gz'
+  sha256 "816d611a0af50c469f52946c94b88c003f908ee031ee0e55a45273e4c64897c5"
+  head 'https://github.com/qbittorrent/qBittorrent.git',
+    :branch => ENV['QBT_BRANCH'] || "master"
+
+  depends_on "pkg-config" => :build
+  depends_on "qt" => :build
+  depends_on "libtorrent-rasterbar" => :build
+  depends_on "boost" => :build
+  depends_on "gnu-sed" => :build
+
+  resource "geoIP" do
+    url "http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz"
+    sha256 "c03b4b8e10d67626fb3cd67cb338d186ce49273be60519dc4ad2ff073437bb87"
+  end
+
+  def install
+    (buildpath/"src/gui/geoIP").install resource("geoIP") if build.stable?
+
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--prefix=#{prefix}",
+                          "--disable-qt-dbus"
+    system "make"
+    system "macdeployqt src/qBittorrent.app"
+    prefix.install "src/qBittorrent.app"
+  end
+
+  test do
+    system "#{prefix}/qBittorrent.app/Contents/MacOS/qbittorrent", "-v"
+  end
+
+end
--- a/Library/Formula/qbittorrent.rb
+++ b/Library/Formula/qbittorrent.rb
@@ -1,10 +1,10 @@
 class Qbittorrent < Formula
   desc "Qt based free and reliable P2P Bittorrent client"
-  homepage 'http://www.qbittorrent.org'
-  url 'https://github.com/qbittorrent/qBittorrent/archive/release-3.2.3.tar.gz'
+  homepage "http://www.qbittorrent.org"
+  url "https://github.com/qbittorrent/qBittorrent/archive/release-3.2.3.tar.gz"
   sha256 "816d611a0af50c469f52946c94b88c003f908ee031ee0e55a45273e4c64897c5"
-  head 'https://github.com/qbittorrent/qBittorrent.git',
-    :branch => ENV['QBT_BRANCH'] || "master"
+  head "https://github.com/qbittorrent/qBittorrent.git",
+    :branch => ENV["QBT_BRANCH"] || "master"
 
   depends_on "pkg-config" => :build
   depends_on "qt" => :build
--- a/Library/Formula/qbittorrent.rb
+++ b/Library/Formula/qbittorrent.rb
@@ -3,8 +3,7 @@ class Qbittorrent < Formula
   homepage "http://www.qbittorrent.org"
   url "https://github.com/qbittorrent/qBittorrent/archive/release-3.2.3.tar.gz"
   sha256 "816d611a0af50c469f52946c94b88c003f908ee031ee0e55a45273e4c64897c5"
-  head "https://github.com/qbittorrent/qBittorrent.git",
-    :branch => ENV["QBT_BRANCH"] || "master"
+  head "https://github.com/qbittorrent/qBittorrent.git"
 
   depends_on "pkg-config" => :build
   depends_on "qt" => :build
--- a/Library/Formula/qbittorrent.rb
+++ b/Library/Formula/qbittorrent.rb
@@ -31,5 +31,4 @@ class Qbittorrent < Formula
   test do
     system "#{prefix}/qBittorrent.app/Contents/MacOS/qbittorrent", "-v"
   end
-
 end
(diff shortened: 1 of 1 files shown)


Library/Formula/qbittorrent.rb
```diff
@@ -0,0 +1,36 @@
+class Qbittorrent < Formula
+  desc "Qt based free and reliable P2P Bittorrent client"
+  homepage 'http://www.qbittorrent.org'
```
bfontaine: Please use double-quotes everywhere.

Library/Formula/qbittorrent.rb
```diff
@@ -0,0 +1,36 @@
...
+  url 'https://github.com/qbittorrent/qBittorrent/archive/release-3.2.3.tar.gz'
+  sha256 "816d611a0af50c469f52946c94b88c003f908ee031ee0e55a45273e4c64897c5"
+  head 'https://github.com/qbittorrent/qBittorrent.git',
+    :branch => ENV['QBT_BRANCH'] || "master"
```
bfontaine: I don’t think we should support this.

Library/Formula/qbittorrent.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def install
+    (buildpath/"src/gui/geoIP").install resource("geoIP") if build.stable?
+
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--prefix=#{prefix}",
+                          "--disable-qt-dbus"
+    system "make"
+    system "macdeployqt src/qBittorrent.app"
```
bfontaine: Please split this into one string per argument: `system "macdeployqt", "src/qBittorrent.app"`.

Library/Formula/qbittorrent.rb
```diff
@@ -0,0 +1,36 @@
...
+    :branch => ENV['QBT_BRANCH'] || "master"
+
+  depends_on "pkg-config" => :build
+  depends_on "qt" => :build
+  depends_on "libtorrent-rasterbar" => :build
+  depends_on "boost" => :build
+  depends_on "gnu-sed" => :build
+
+  resource "geoIP" do
+    url "http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz"
```
DomT4: This changes enormously frequently. It'll break on from-source builds. We'll need to find a less frequent mirror; Debian have one.

Library/Formula/qbittorrent.rb
```diff
@@ -0,0 +1,36 @@
...
+                          "--disable-qt-dbus"
+    system "make"
+    system "macdeployqt src/qBittorrent.app"
+    prefix.install "src/qBittorrent.app"
+  end
+
+  test do
+    system "#{prefix}/qBittorrent.app/Contents/MacOS/qbittorrent", "-v"
+  end
+
```
DomT4: Don't need the empty line here.

Library/Formula/qbittorrent.rb
```diff
@@ -0,0 +1,36 @@
+class Qbittorrent < Formula
+  desc "Qt based free and reliable P2P Bittorrent client"
+  homepage 'http://www.qbittorrent.org'
+  url 'https://github.com/qbittorrent/qBittorrent/archive/release-3.2.3.tar.gz'
+  sha256 "816d611a0af50c469f52946c94b88c003f908ee031ee0e55a45273e4c64897c5"
+  head 'https://github.com/qbittorrent/qBittorrent.git',
+    :branch => ENV['QBT_BRANCH'] || "master"
+
+  depends_on "pkg-config" => :build
+  depends_on "qt" => :build
```
DomT4: This isn't likely to be a buildtime only dependency.

Library/Formula/qbittorrent.rb
```diff
@@ -0,0 +1,36 @@
...
+  depends_on "libtorrent-rasterbar" => :build
+  depends_on "boost" => :build
```
DomT4: Don't think this is either, from the last time I build QBT from source.

mistydemeo: `macdeployqt` embeds all of the libraries into the built app bundle.

DomT4: I had some problems with this before; I briefly supplied a few test `dmg` packages for upstream & the deploy process didn't quite work as it should have. Perhaps the root cause has been fixed upstream, or I just really sucked :smile:

Library/Formula/qbittorrent.rb
ainstushar: Yeah I discussed that with @mistydemeo and you are right. Maybe finding a different source would be helpful. Discussed not supplying sha256 for it. Any other suggestions might be helpful.

Library/Formula/qbittorrent.rb
ainstushar: @bfontaine, just curiosity, why is it not a good idea to support branch thingy?

Library/Formula/qbittorrent.rb
DomT4: It's difficult, to be honest. Debian hack around this in their `geoip` package by doing this in the bootstrap:

```
mkdir -p $DIR/data
if [ ! -f $DIR/data/GeoIP.dat  ]; then
      curl http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz | gzip -d > $DIR/data/GeoIP.dat
fi
```

Not ideal, really. My gut feeling would be for Homebrew to mirror semi-up-to-date GeoIP.dat "releases", but the attribution element of the license requirement makes that slightly tricky.

ainstushar: They are updating their db only once a month. TBH I don't really have a problem with leaving this in there. I am sure qBittorrent devs can include attribution in their "About" section in app. That should cover it. Just to make sure, omitting sha256 is a big deal? (I am new at this so sorry in advance). I mean I could just have a script running on 1st Tuesday of each month updating hash...

DomT4: Kind of a big deal. We warn every time if a checksum isn't attached to a resource in a formula, unless that resource is from a git checkout or such.

```
==> Downloading http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz
Warning: Cannot verify integrity of qbittorrent--geoIP-3.2.3.gz
A checksum was not provided for this resource
For your reference the SHA256 is: c03b4b8e10d67626fb3cd67cb338d186ce49273be60519dc4ad2ff073437bb87
```

ainstushar: OK, so updating this formula on 1st Tuesday of each month with appropriate checksum is fine right?

DomT4: I guess so, if nobody else objects.

dunn: Still a little leery.  There'll be a short period once a month when installations fail.  And what if you forget?

ainstushar: I was thinking about just running a cron job that monitors if the file changes and then automatically update the formula. But I see that if a user doesn't update formula, then they are screwed. I can't just keep a copy on github and distribute it due to their licensing. Other option is to keep an outdated copy in the code so we don't have to worry about adding it using formula.

bfontaine: > But I see that if a user doesn't update formula, then they are screwed

They have a local copy that was downloaded by the formula so why would they be screwed?

sledgehammer999: Guys, just so you know that url stays always the same. And that file updates only once a month.
Anyway, as discussed in our bugtracker from v3.3.0 you don't need to include geoip db.

mikemcquaid: We need to work something out that doesn't rely on someone updating this once a month.

mikemcquaid:
I think this would be a better fit for Homebrew Cask given that upstream provides a binary installer for QBittorrent.

ainstushar:
@mikemcquaid There's already a cask formula available if user wants to install qBittorrent without compiling it. I made this formula so that people who wanted to test out the latest code on github can do so with ``--HEAD.`` As a matter of fact, I talked with some folks in homebrew cask irc and they recommended that I create a formula here. :) Was I told wrong?

mikemcquaid:
I wouldn't say you're told wrong but I think I probably have a different opinion there. I'm OK with another maintainer merging this, though.

mikemcquaid:
Maintainers: this may be a better fit for https://github.com/Homebrew/homebrew-gui if we decide to go ahead with that. @ainstushar Just ignore this for now.

DomT4:
Going to pass on this *for now*, per:

> Anyway, as discussed in our bugtracker from v3.3.0 you don't need to include geoip db.

Once we reach that point and don't have to vendor in a file where the checksum is going to change and we have to require people to be very on-point in updating it every 30 days we'll review this again, if desired.

PR #43597, opened 2015-09-05
Adds HEAD for ponyc

colindean:
Bug in 0.1.7 causing segfaults during compliation fixed in master.

### DIFF
--- a/Library/Formula/ponyc.rb
+++ b/Library/Formula/ponyc.rb
@@ -3,6 +3,7 @@ class Ponyc < Formula
   homepage "http://www.ponylang.org"
   url "http://releases.ponylang.org/source/ponyc-0.1.7.tar.bz2"
   sha256 "fc6f783f65cd6708a80bdea71f414cada801528143ea22d9bb13957cb7061eb6"
+  head "https://github.com/CausalityLtd/ponyc.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


dunn:
Can the commit in question be applied to the stable build as a patch?  Something like
```ruby
class Ponyc < Formula
  desc "Object-oriented, actor-model, capabilities-secure programming language"
  homepage "http://www.ponylang.org"
  head "https://github.com/CausalityLtd/ponyc.git"

  stable do
    url "http://releases.ponylang.org/source/ponyc-0.1.7.tar.bz2"
    sha256 "fc6f783f65cd6708a80bdea71f414cada801528143ea22d9bb13957cb7061eb6"

    patch do
      # the commit hash instead of '123'
      url "https://github.com/CausalityLtd/ponyc/commit/123.patch"
      sha256 "123"
    end
  end
```

colindean:
I didn't really take the time to figure out exactly *which* commit fixes it. I only know that what's at HEAD of master works and that the formula didn't previously have a `head`. When I added it and reinstalled with `--head`, my compile worked.

mikemcquaid:
Would be great if you could figure that out but pulling this as-is for now.

For future reference the preferred commit message format is `ponyc: add HEAD.`.

PR #43602, opened 2015-09-05
readall: add --aliases

### LINKED ISSUE #43600: Broken symlinks in repo
I wonder if anyone can reproduce this, but in two separate clones on my machine, the following files are symlinks whose targets don't exist.

First, some Homebrew aliases:
```
$HOMEBREW_REPOSITORY/Library/Aliases/actor-framework
$HOMEBREW_REPOSITORY/Library/Aliases/o-caml
$HOMEBREW_REPOSITORY/Library/Aliases/shell-fm
```

I suppose with those that someone who engaged in a rename or a deletion of a formula forgot to handle all its aliases.

But with the next two cases, I dunno what's up:

```
$HOMEBREW_REPOSITORY/Library/ENV/pkgconfig/fuse/fuse.pc
$HOMEBREW_REPOSITORY/Library/ENV/pkgconfig/fuse/osxfuse.pc
```

Running `file` on those gives `broken symbolic link to /usr/local/lib/pkgconfig/fuse.pc`.  That doesn't make sense to me, as I have never brewed fuse/osxfuse.

I haven't made a PR because I want to first make sure these five things really are supposed to be deleted.

xu-cheng:
Also fix broken alias and syntax error.

Closes #43600

### DIFF
--- a/Library/Homebrew/cmd/readall.rb
+++ b/Library/Homebrew/cmd/readall.rb
@@ -17,17 +17,19 @@ module Homebrew
       end
 
       failed = false
-      workers = (0...Hardware::CPU.cores).map do
-        Thread.new do
-          begin
-            while rb = ruby_files.pop(true)
-              nostdout { failed = true unless system RUBY_PATH, "-c", "-w", rb }
+      nostdout do
+        workers = (0...Hardware::CPU.cores).map do
+          Thread.new do
+            begin
+              while rb = ruby_files.pop(true)
+                failed = true unless system RUBY_PATH, "-c", "-w", rb
+              end
+            rescue ThreadError # ignore empty queue error
             end
-          rescue ThreadError # ignore empty queue error
           end
         end
+        workers.map(&:join)
       end
-      workers.map(&:join)
       Homebrew.failed = failed
     end
 
--- a/Library/Homebrew/cmd/readall.rb
+++ b/Library/Homebrew/cmd/readall.rb
@@ -31,6 +31,15 @@ module Homebrew
       Homebrew.failed = failed
     end
 
+    if ARGV.delete("--aliases")
+      Pathname.glob("#{HOMEBREW_LIBRARY}/Aliases/*").each do |f|
+        next unless f.symlink?
+        next if f.file?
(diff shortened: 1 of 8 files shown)


chdiza:
Ah, good.  I was about to suggest making `brew prune` look for broken symlinks among the aliases, but this is better.

xu-cheng:
cc @Homebrew/owners for comments before I merge this.

Library/Homebrew/cmd/readall.rb
```diff
@@ -17,20 +17,31 @@ def readall
...
     end
 
+    if ARGV.delete("--aliases")
+      Pathname.glob("#{HOMEBREW_LIBRARY}/Aliases/*").each do |f|
+        next unless f.symlink?
+        next if f.file?
+        onoe "Broken alias: #{f}"
+        Homebrew.failed = true
+      end
+    end
```
mikemcquaid: I think we should probably just do this by default.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -573,7 +573,7 @@ def homebrew
       return if ARGV.include? "--skip-homebrew"
       test "brew", "tests"
       test "brew", "tests", "--no-compat"
-      readall_args = []
+      readall_args = ["--aliases"]
```
mikemcquaid: :+1: nice.

Library/Homebrew/cmd/readall.rb
mikemcquaid: But given you've added it to the `test-bot` below I'm OK with omitting it if you'd rather.

mikemcquaid:
:+1:

Library/Homebrew/cmd/readall.rb
xu-cheng: Since we support `brew readall <some tap>`, it would be weird to have a tap readall to check core aliases. Therefore, I will keep this flag.

mikemcquaid: :+1: :shipit:

PR #43615, opened 2015-09-05
remove Formula#file_modified?

### LINKED ISSUE #43607: Third-party diff tools will lead to deadlock
I use `brew edit go` to modify formula and rebuild golang, but I found that `brew upgrade` will be blocked. Then I checked and found that brew will detect if there's local changes in formulas and maybe it runs `git diff`, I use `vimdiff` as default diff tool for `git diff`, so brew will wait for vimdiff to exit and vimdiff will never exit because it runs background... So I was wondering if maybe we can do something to enhance the behavior for checking local changes...

dunn:
It's been more trouble than it's worth.

Fixes #43607.
Closes #43049.

### DIFF
--- a/Library/Homebrew/cmd/fetch.rb
+++ b/Library/Homebrew/cmd/fetch.rb
@@ -33,13 +33,6 @@ module Homebrew
     return true if ARGV.force_bottle? && f.bottle
     return false unless f.bottle && f.pour_bottle?
     return false if ARGV.build_from_source? || ARGV.build_bottle?
-    if f.file_modified?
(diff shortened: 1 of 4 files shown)


dunn:
cc @Homebrew/owners

mistydemeo:
:+1:

mikemcquaid:
:+1: :sob:

DomT4:
It'd be nice to try and work something like this at some point, as it is a useful feature in certain circumstances, but the edge cases are troublesome to handle. :+1: from me.

xu-cheng:
:+1:

PR #43620, opened 2015-09-05
fq 0.9.2, HEAD

dunn:


mikemcquaid:
:+1: once green

DomT4:
Sandbox failure, I believe:

```
Opening configdb /usr/local/Cellar/fq/0.9.2/var/lib/fq/fqd.sqlite
... failed to open /usr/local/Cellar/fq/0.9.2/var/lib/fq/fqd.sqlite: unable to open database file
```

dunn:
It works locally with `--sandbox.`  I'll poke a little more.

dunn:
Oh, @DomT4 was right.  The sandbox is preventing fqd.sqlite from being created; I had already created it on a previous non-sandbox test, so that's why I wasn't seeing the error.

PR #43622, opened 2015-09-05
osxutils: Remove conflict line referencing non-existant googlecl

timsutton:
This formula doesn't install because brew looks for the now-removed `googlecl` formula. This PR removes the reference to it, and cleans up the formula to pass audit.


xu-cheng:

PR #43319, opened 2015-08-27
activemq 5.12.0

stuwil:

PR #43634, opened 2015-09-06
activemq 5.11.2

nijikon:


xu-cheng:
Dupe with #43319 ?

nijikon:
@xu-cheng I saw that, I assumed that this can go in until that one is closed.

PR #43638, opened 2015-09-06
ccache 3.2.3

nijikon:


nijikon:
If added ```--with-bundled-zlib```, cause when I was building against system zlib, I did get this error ```make: *** No rule to make target `-lz', needed by `ccache'.  Stop.```

mikemcquaid:
Thanks @nijikon!

PR #43640, opened 2015-09-06
libelf 0.8.14

nijikon:


Library/Formula/libelf.rb
```diff
@@ -19,6 +19,8 @@ def install
                           "--disable-compat"
     # Use separate steps; there is a race in the Makefile.
     system "make"
+    mkdir_p lib
```
mikemcquaid: Probably prefer `lib.mkpath` and `(include+name).mkpath`

DomT4:
Switched it over to `mkpath` after pull. Merged in https://github.com/Homebrew/homebrew/commit/7063c65967015ed8d01ac4cc23dcffc1616b51cb, Thanks as ever @nijikon!

PR #43647, opened 2015-09-06
mutt: remove bad patches

dunn:
trash-patch and pgp-verbose-mime-patch no longer apply cleanly.

Also update sha1 -> sha256 and avoid trailing block conditionals.

### DIFF
--- a/Library/Formula/mutt.rb
+++ b/Library/Formula/mutt.rb
@@ -34,10 +34,8 @@ class Mutt < Formula
     :because => "both install mmdf.5 and mbox.5 man pages"
 
   option "with-debug", "Build with debug option enabled"
-  option "with-trash-patch", "Apply trash folder patch"
   option "with-s-lang", "Build against slang instead of ncurses"
   option "with-ignore-thread-patch", "Apply ignore-thread patch"
-  option "with-pgp-verbose-mime-patch", "Apply PGP verbose mime patch"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

mistydemeo:
:+1:

dunn:
Posting my commit message here, if people come to this issue wondering what happened to the patches and why:

> The trash-patch and pgp-verbose-mime-patch no longer apply cleanly, and
> are being removed.
> 
> Maintaining the many (often out-of-date) patches that mutt users rely on
> is not something we have the time or expertise to do.  We do however
> have a very flexible tap system that the mutt community can use to
> maintain their own version of this formula with better support for
> patches: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md

x-way:
Being a user of the trash_folder patch, I did as you explained and created a tap with the trash_folder patch for Mutt 1.5.24: https://github.com/x-way/homebrew-mutt

emmanuelbernard:
Here is another formluae based on @x-way 's

- based on Mutt 1.5.24
- sidebar patch
- trash patch
- indexcolor patch

https://emmanuelbernard.com/blog/2016/03/15/brew-formulae-mutt-sidebar-trash/
https://github.com/emmanuelbernard/homebrew-mutt

PR #43648, opened 2015-09-06
add brew update-test

xu-cheng:
To prevent https://github.com/Homebrew/homebrew/issues/42553 happening again, add `brew update-test` which will simulate brew update in test bot.

cc @Homebrew/owners


mikemcquaid:
Great work here. The only thing is I'm not sure if this should be in `cmd` as it's a strange thing for users to be able to run.

xu-cheng:
Because we need to manipulate Homebrew core file, I found it's hard to put it into `brew tests` as it use complete different file structure. Let me do more test to be really sure whether that is possible.

xu-cheng:
Also, I kinda want to avoid running this for tap test because I think it should has certain performance impact .

mikemcquaid:
I think it's fine to avoid it for a tap test.

mikemcquaid:
Tap tests probably shouldn't even run `brew tests` imo.

xu-cheng:
> The only thing is I'm not sure if this should be in cmd as it's a strange thing for users to be able to run.

I found putting it into `brew tests` is a bad idea for following reasons:
* It requires some hacks (Bundle env, get repo real path, etc.)
* The result of `brew tests` is constant with code, while result of `brew update-test` depends on current commit state.
* We won't get detail output when update-test failed if we put it in tests.

I think making standalone cmd seems to be the best approach so far.

> Tap tests probably shouldn't even run brew tests imo.

I have applied such change.

mikemcquaid:
@xu-cheng I'm OK with it being a standalone command, I just think it should probably live in another location than `cmd`.

xu-cheng:
How about creating a `dev-cmd` folder which it's only visible when Homebrew developer is set?

mikemcquaid:
:+1:

DomT4:
Could we do something as simple as:
```ruby
return unless !ENV["HOMEBREW_DEVELOPER"].nil?
```
?

mikemcquaid:
@DomT4 I'd rather it wasn't exposed as a command really.

DomT4:
Fair enough :+1:.

xu-cheng:
Another slight related questions: what other commands we want to move them to `dev-cmd`?

xu-cheng:
OK created `dev-cmd` as discussed

xu-cheng:
Any other comments before I merge it. 

Since we are creating dev-cmd folder. I wonder:
* We should move more cmds there, such as test-bot tests etc. 
* Should we support similar feature in the tap?

bfontaine:
> We should move more cmds there, such as test-bot tests etc.

This might break some scripts for people who use these commands and don’t have the `HOMEBREW_DEVELOPER` variable set.

> Should we support similar feature in the tap?

I’m personally :+1: on that.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -573,11 +573,16 @@ def formula(formula_name)
     def homebrew
       @category = __method__
       return if @skip_homebrew
-      test "brew", "tests"
-      test "brew", "tests", "--no-compat" if @tap
-      readall_args = ["--aliases"]
-      readall_args << "--syntax" if MacOS.version >= :mavericks
-      test "brew", "readall", *readall_args
+      if @tap
+        test "brew", "readall", @tap
```
mikemcquaid: Last time we talked about this I said I thought it made sense to run `brew tests` for taps so they can see if there are upstream issues. Do you disagree?

xu-cheng: Sorry I forget about this part. Fixed in 4c6e764f95cbe1074731b4ff3b31a23a7aeb4a42

mikemcquaid: Thanks :heart:

PR #43655, opened 2015-09-07
gammaray 2.3.0, HEAD

dunn:
Garbage new test because there's no longer anything installed outside
the app bundle and the bundled executables have no command-line uses.


mikemcquaid:
:+1:

PR #43659, opened 2015-09-07
takt 0.309 (new formula)

JikanBae:
Takt is a text-based music programming language designed by Satoshi
Nishimura. It has the following features:

* It allows a concise description of note and chord sequences like
  `c d [e g]`.
* Expressive information such as dynamics or tempo rubato can be added
  to such note/chord sequences.
* It provides rich programming features including object-oriented
  support with C-like syntax. Such features enable seamless fusion of
  note-by-note description and algorithmic composition.
* Music transformation (such as transposition) can be applied to
  musical phrases described with the above means. Many predefined
  transformation modules are available; in addition; users can develop
  their own transformation modules with the language. Transformations
  can also be applied to standard MIDI files or events from a MIDI
  input device in a reactive way.

(above lines were taken from the Takt website)

### DIFF
--- a/Library/Formula/takt.rb
+++ b/Library/Formula/takt.rb
@@ -1,0 +1,18 @@
+class Takt < Formula
+  desc "a text-based music programming language"
+  homepage "http://takt.sourceforge.net"
+  url "http://sourceforge.net/projects/takt/files/takt-0.309-src.tar.gz"
+  version "0.309"
+  sha256 "44fca667d610fba15deb441dd44e5b87dabbe5619d7fa8a13f5c8c1f054dd509"
+
+  depends_on "readline"
+
+  def install
+    system "./configure", "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system bin/"takt", "-o etude1.mid", pkgshare/"examples/etude1.takt"
+  end
+end
--- a/Library/Formula/takt.rb
+++ b/Library/Formula/takt.rb
@@ -1,14 +1,13 @@
 class Takt < Formula
-  desc "a text-based music programming language"
+  desc "text-based music programming language"
   homepage "http://takt.sourceforge.net"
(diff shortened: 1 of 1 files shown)


Library/Formula/takt.rb
```diff
@@ -0,0 +1,18 @@
+class Takt < Formula
+  desc "a text-based music programming language"
```
DomT4: No need for the `a` here. Just `Text-based music programming language` is fine.

Library/Formula/takt.rb
```diff
@@ -0,0 +1,18 @@
...
+  homepage "http://takt.sourceforge.net"
+  url "http://sourceforge.net/projects/takt/files/takt-0.309-src.tar.gz"
```
DomT4: This should use a `https://downloads.sourceforge.net` URL.

Library/Formula/takt.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  depends_on "readline"
+
+  def install
+    system "./configure", "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system bin/"takt", "-o etude1.mid", pkgshare/"examples/etude1.takt"
```
DomT4: You'll likely need to install the examples into `pkgshare` manually. Not many things install examples by default.

JikanBae: Hi, thanks for the comments. I really appreciate them, since it's my first time here for the pull request. Do you mean `pkgshare` instead of `pkgshare/examples`?

DomT4: > (I might be misunderstanding your point. Takt's make install automatically installs the example files in pkgshare/examples.)

This was what I was asking about, aye :). Cool, that's pretty unusual! Thanks for the clarification.

mikemcquaid: :+1:

mikemcquaid:

JikanBae:
Thank you all for kind comments and supports!

PR #43663, opened 2015-09-07
audit: cache formula names during audit_formula_name

xu-cheng:
It speeds up the process when auditing multiple formulae. For example, audit all of formulae:
* Before `audit_formula_name: 15.094845999999986`
* After `audit_formula_name: 0.5617120000000043`


mikemcquaid:
:heart:

PR #43667, opened 2015-09-07
phraseapp 1.0.1 (new formula)

sacry-dyn:
New formula for Homebrew, created a new pull request for PhraseApp. See here for an older (rejected) pull request https://github.com/Homebrew/homebrew/pull/42269

### DIFF
--- a/Library/Formula/phraseapp.rb
+++ b/Library/Formula/phraseapp.rb
@@ -1,0 +1,16 @@
+class Phraseapp < Formula
+  desc "PhraseApp CLI Tool - API client binary"
+  homepage "https://phraseapp.com/cli"
+  url "https://github.com/phrase/phraseapp-client/releases/download/1.0.1/phraseapp_macosx_amd64.tar.gz"
+  version "1.0.1"
+  sha256 "4b1aa4c36f7c92618904d6882817e8459e3420f146e4864555dbe6ab9b65a192"
+
+  def install
+    bin.install "phraseapp"
+  end
+
+  test do
+    output = `#{bin}/phraseapp info`
+    assert_match /Built at .+/, output
+  end
+end


DomT4:
From the previous thread:

> A future submission should build from source using Go as well. We don't use upstream binaries, we produce our own.

:)

Library/Formula/phraseapp.rb
```diff
@@ -0,0 +1,16 @@
...
+  url "https://github.com/phrase/phraseapp-client/releases/download/1.0.1/phraseapp_macosx_amd64.tar.gz"
+  version "1.0.1"
+  sha256 "4b1aa4c36f7c92618904d6882817e8459e3420f146e4864555dbe6ab9b65a192"
+
+  def install
+    bin.install "phraseapp"
+  end
+
+  test do
+    output = `#{bin}/phraseapp info`
```
DomT4: Should be able to use `shell_output` or `pipe_output` rather than backtics here.

sacry-dyn:
Hi @DomT4 , 

thank you for the fast reply and feedback. We will find something out for the binary. The building of this client is a little bit more complicated because we generate code for several repositories and then make the binary (some of the repositories are private (the one that generates the code)) and others public. As for the test we will be on it. Does this PR stay open or do we need to make another pull request?

Regards

DomT4:
Feel free to leave this one open whilst you work on it. If you  keep us semi-updated it won't be closed unless there's nothing we can do to get this into Homebrew or doing so becomes too convoluted.

sacry-dyn:
Hey,

just to give you a small heads up. We are working on it, go 1.5 building gives us a small headache. We are looking into resolving this though.

Regards,
Matthias

DomT4:
Thanks for the update!

DomT4:
How's this going?

sacry-dyn:
@DomT4 sorry for the late reply (holidays). We are still working on it.  This is a bit on "hold" as we need to evaluate our options on how to migrate to go 1.5, as it has still a few things we do not like about it and go 1.4 is more reasonable to use as of now. This will change soon but it can take a bit more time. We hope this is okay.

DomT4:
No worries! As long as we're aware the PR isn't dead or likely to hang indefinitely it can stay open.

bfontaine:
Anything new here?

bfontaine:
Ping?

Library/Formula/phraseapp.rb
```diff
@@ -0,0 +1,16 @@
+class Phraseapp < Formula
+  desc "PhraseApp CLI Tool - API client binary"
+  homepage "https://phraseapp.com/cli"
+  url "https://github.com/phrase/phraseapp-client/releases/download/1.0.1/phraseapp_macosx_amd64.tar.gz"
```
bfontaine: We need to build from source.

bfontaine:
Anything new? The latest release is now [1.1.8](https://github.com/phrase/phraseapp-client/releases/tag/1.1.8).

mikemcquaid:
Let's get a newer release here and include it in a new PR.

PR #43670, opened 2015-09-07
ffmpeg: add libbs2b option

Nyx0uf:


Library/Formula/ffmpeg.rb
```diff
@@ -33,6 +33,7 @@ class Ffmpeg < Formula
   option "with-libsoxr", "Enable the soxr resample library"
   option "with-webp", "Enable using libwebp to encode WEBP images"
   option "with-zeromq", "Enable using libzeromq to receive commands sent through a libzeromq client"
+  option "with-libbs2b", "Enable using libbs2b"
```
bfontaine: I don’t think this line is necessary; Homebrew will generate a similar option text if we don’t explicitly add it.

bfontaine:
How common is this need to build with `libbs2b`?

Nyx0uf:
mpv player removed its libbs2b support in favor of the ffmpeg one through lavfi filters. and it's optional anyway so…

see https://github.com/mpv-player/mpv/commit/d04d2380e35ba016e3816ab5d7c282997df74168

bfontaine:
@Nyx0uf We try to limit the number of options on a formula so we have three possibilities here:

- either the option will be widely used; so we should build with `libbs2b` by default (and thus remove the option)
- either it’s really niche and will be rarely used; so we shouldn’t support it
- neither of the previous possibilities; so we should include the option

Which one fits here?

Nyx0uf:
@Nyx0uf We try to limit the number of options on a formula so we have three possibilities here:

Just curious, why ? if it's optional it doesn't bother people who don't want it or don't know what it is. It's really inconvenient to build something from sources when all others dependencies are handled with Homebrew.

And to answer your question, I said the third. It's clearly not widely used, but I know several people using bs2b filters in their media players

mikemcquaid:
> Just curious, why ? if it's optional it doesn't bother people who don't want it or don't know what it is. It's really inconvenient to build something from sources when all others dependencies are handled with Homebrew.

Basically because they aren't tested so if/when they break we don't know.

Library/Formula/ffmpeg.rb
bfontaine: Please remove the line. With it `brew info ffmpeg` shows this:

```
…snip…
--with-libbluray
	Build with libbluray support
--with-libbs2b
	Enable using libbs2b
--with-libcaca
	Build with libcaca support
--with-libquvi
	Build with libquvi support
…snip…
```

Without it:

```
…snip…
--with-libbluray
	Build with libbluray support
--with-libbs2b
	Build with libbs2b support
--with-libcaca
	Build with libcaca support
--with-libquvi
	Build with libquvi support
…snip…
```

I thus don’t think it’s necessary.

Nyx0uf: should be good

gtm87x:
News about this?
With the last version of mpv released today, they completely dropped the internal bs2b implementation in favour of the libavfilter's one, and this filter has a quite affectionate following.

Nyx0uf:
agreed, it's should be good for merge, what's holding it ?

mikemcquaid:
Thanks for your contribution to Homebrew (and sorry for the delay)! Without people like you submitting PRs we couldn't run this project. You rock!

PR #43673, opened 2015-09-07
gom 0.3.1 (new formula)

tschoonj:
triggers the development release warning with `brew --strict audit`,
but this should be ignored as there isn't a version 1.0.0 yet.

### DIFF
--- a/Library/Formula/gom.rb
+++ b/Library/Formula/gom.rb
@@ -1,0 +1,81 @@
+class Gom < Formula
+  desc "GObject wrapper around SQLite"
+  homepage "https://wiki.gnome.org/Projects/Gom"
+  url "https://download.gnome.org/sources/gom/0.3/gom-0.3.1.tar.xz"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
@tschoonj Can you suggest how to fix the rule to avoid the false positive? Should it not apply for versions below 1.0 or something? Thanks!

tschoonj:
Hi Mike,

Like you said, I think it would be good to exclude all GNOME project releases with version number less than 1.0.0 from this rule.

DomT4:
Long-term we need some grand whitelist for various audit checks, but I'm tempted to just throw together a very basic list of exemptions from this development version issue to fix it for now.

tschoonj:
@DomT4 Could you restart this one?

Thanks!

DomT4:
@BrewTestBot test this please

tschoonj:
@DomT4 and this is the last one :smile:

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/e15fa4e6211dc6ad78ca5c85b14eeacfdd0f1eb3. Thank you again :smile_cat:

PR #43677, opened 2015-09-07
puddletag 1.0.5 (new feature: app bundle as starter)

chincheta0815:
Added a routine for an app bundle for starting the puddletag gui by just a click.
I hope this meets the conventions.
If not: Please provide info on where I can get info for my intent.

### DIFF
--- a/Library/Formula/puddletag.rb
+++ b/Library/Formula/puddletag.rb
@@ -70,7 +70,7 @@ class Puddletag < Formula
     cp buildpath/"Info.plist", "Puddletag.app/Contents"
     system "sips", "-s", "format", "icns", buildpath/"puddletag.png", "--out", "Puddletag.app/Contents/Resources/Puddletag.icns"
     cp prefix/"bin/puddletag", "Puddletag.app/Contents/MacOS/Puddletag"
-    chmod "+x", "Puddletag.app/Contents/MacOS/Puddletag"
+    chmod "755", "Puddletag.app/Contents/MacOS/Puddletag"
     prefix.install "Puddletag.app"
   end
 
--- a/Library/Formula/puddletag.rb
+++ b/Library/Formula/puddletag.rb
@@ -4,13 +4,6 @@ class Puddletag < Formula
   url "https://github.com/keithgg/puddletag/archive/v1.0.5.tar.gz"
   sha256 "f94ebcc4ed31389574c187197b99256bec1f96e1e7d4dd61730e88f79deeaba2"
 
-  bottle do
-    cellar :any
-    sha256 "a68e0d8951475db3151e8bbd91a66028200ea2bd18363fcd37a6d9191e693633" => :yosemite
-    sha256 "64fbfbe641417db9cf8544360628c15894e633a48b81765033f0f2f404876419" => :mavericks
-    sha256 "c5ff96058a4f5262822327108318549a76beeaef4c39715944a22b30bdd19280" => :mountain_lion
-  end
-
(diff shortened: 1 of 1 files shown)


Library/Formula/puddletag.rb
```diff
@@ -4,13 +4,6 @@ class Puddletag < Formula
   url "https://github.com/keithgg/puddletag/archive/v1.0.5.tar.gz"
   sha256 "f94ebcc4ed31389574c187197b99256bec1f96e1e7d4dd61730e88f79deeaba2"
 
-  bottle do
```
dunn: You can leave the bottle block as-is; our test-bot takes care of it!

chincheta0815: Do I have to recommit or is it okay for now?

Library/Formula/puddletag.rb
```diff
@@ -45,7 +38,40 @@ def install
...
+      </plist>
+    EOS
+
+    mkdir_p "Puddletag.app/Contents/MacOS"
+    mkdir_p "Puddletag.app/Contents/Resources"
+    cp buildpath/"Info.plist", "Puddletag.app/Contents"
+    system "sips", "-s", "format", "icns", buildpath/"puddletag.png", "--out", "Puddletag.app/Contents/Resources/Puddletag.icns"
+    cp prefix/"bin/puddletag", "Puddletag.app/Contents/MacOS/Puddletag"
+    chmod "+x", "Puddletag.app/Contents/MacOS/Puddletag"
+    prefix.install "Puddletag.app"
```
mikemcquaid: I'd rather we didn't do this in the Homebrew formula; it should be done in the upstream buildsystem insted.

chincheta0815: Up to now they don't have that in their code. They just provide everything for linux (as of today...)

mikemcquaid: Sure, I just don't think creating an app bundle is Homebrew's responsibility.

chincheta0815: I totally understand.
The app bundle I wrote is more or less a "windows link" not the app itself. It just contains an icon and the wrapper script made by homebrew for starting. Nothing else, no libs, not anything else.
Is there a mechanism in hombrew which would provide the thing I need?

mikemcquaid: Sorry, I don't agree. Please submit it upstream rather than to Homebrew.

chincheta0815: Okay, I will.
Just some question on how you would accept it from upstream (sorry for being nasty):
Is it okay to give them a script containing a routine (bash script; the routine as is in the forumula) that can be run during the brew install process for making the bundle (which means acutally just creating predefined two files from the bash script routine, nothing more)?
Or is it necessary to push the whole bundle structure?

For clarity I could provide the contents of this script.

Using a script would be the easiest. Then the name of the executing script would just have to be patched/changed/set and the respective bash script would be more usable for upstream. Maintenance would be at upstream.

I hope I am not too nasty, but I think having your thoughts up front will provoke less effort for all.

Thanks in advance.

If there's another way to discuss that (private message, other means) pleaae let know. I won't bother the build system or so...

mikemcquaid: If it's included in the upstream repository a script sounds fine, yeh. Thanks!

chincheta0815: Okay. I talk to the upstream maintainers.
I leave this pull request open until I get news.

PR #43680, opened 2015-09-07
proxmark3: add 2.2.0 (new formula)

chrisfu:
Formula compiles both old and new flashers (useful when upgrading the FPGA from the old HID bootloader the newer CDC bootloader), client, and the ARM-based FPGA and bootloader images.


Library/Formula/proxmark3.rb
```diff
@@ -0,0 +1,39 @@
...
+  head "https://github.com/Proxmark/proxmark3.git"
+
+  depends_on "automake" => :build
+  depends_on "readline"
+  depends_on "p7zip" => :build
+  depends_on "libusb"
+  depends_on "libusb-compat"
+  depends_on "pkg-config" => :build
+  depends_on "wget"
+  depends_on "nitsky/stm32/arm-none-eabi-gcc" => :build
```
DomT4: You can't have mandatory cross-tap dependencies, and in the core we probably won't support even optional cross-tap dependencies to an unofficial tap.

Library/Formula/proxmark3.rb
```diff
@@ -0,0 +1,39 @@
...
+    bin.mkpath
+    bin.install "client/flasher" => "proxmark3-flasher"
+    bin.install "client/hid-flasher/flasher" => "proxmark3-hid-flasher"
+    bin.install "client/proxmark3" => "proxmark3"
+    bin.install "client/fpga_compress" => "fpga_compress"
+    share.mkpath
+    (share/"firmware").mkpath
+    (share/"firmware").install "armsrc/obj/fullimage.elf" => "fullimage.elf"
+    (share/"firmware").install "bootrom/obj/bootrom.elf" => "bootrom.elf"
+    ohai "Install success! Upgrade devices on HID firmware with proxmark3-hid-flasher, or devices on more modern firmware with proxmark3-flasher."
```
DomT4: Please don't `ohai` in formulae.

Library/Formula/proxmark3.rb
```diff
@@ -0,0 +1,39 @@
...
+  depends_on "readline"
+  depends_on "p7zip" => :build
+  depends_on "libusb"
+  depends_on "libusb-compat"
+  depends_on "pkg-config" => :build
+  depends_on "wget"
+  depends_on "nitsky/stm32/arm-none-eabi-gcc" => :build
+
+  def install
+    ENV.deparallelize
```
DomT4: The need to do this should be reported upstream and a link to that bug report added here.

Library/Formula/proxmark3.rb
mikemcquaid: Yup.

chrisfu:
Thanks for the feedback guys. I'll report the bug regarding deparallelize being required to the Proxmark3 devs and see what they have to say.

Removing the ohai is trivial, so I'll sort that today.

With regards to the ARM toolchain, I believe I came across a compliant-looking formula that builds the chain from source, so I'll try and encourage the author to submit a pull request. If not I'll do it myself.

Cheers!

mikemcquaid:
This seems pretty niche for Homebrew core; might be better off in a tap (where you can depend on whatever you want)?

PR #43683, opened 2015-09-07
libelf: bump revision

DomT4:
It took us a while to catch this in terms of how rapidly Homebrew moves, so rather than leaving people with a decade old `libelf` I think it's worthwhile bumping the revision a touch to fix it. The bottle is `cellar :any`.

Roughly ~430 bottle downloads in the last 24 hours.

Ref: https://github.com/Homebrew/homebrew/commit/7063c65967015ed8d01ac4cc23dcffc1616b51cb


mikemcquaid:
:metal:

Issue #42864, opened 2015-08-12
Adding Swift to the Compiler Version Database


beltex:
First off, forgot to say this before. Thank you Homebrew team for such a great project! :)

I was wondering what you thought about adding Swift to the [Compiler Version Database](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Xcode.md#compiler-version-database) in the Xcode document file. If yes, I’d be happy to submit a pull request.

Sincerely,

beltex

denji:
Why in this case there is no Objective-C/C++ ?

Actually Swift is not a compiler, it's a front-end for the LLVM.

beltex:
Hi @denji!

> Why in this case there is no Objective-C/C++ ?

[Clang](http://clang.llvm.org) is listed in the table, which is a C, C++, Objective-C and Objective C++ front-end for LLVM. Thus I thought it be appropriate to add Swift as well, which is also a front-end for LLVM as you mentioned. In addition, because it comes bundled with Xcode as with the rest of the compilers listed in the document.

My understanding as to the purpose of the document is as a reference for which version of Xcode to assert (depend on) in a formula, if the project in question requires a compiler that Xcode will be providing. So for example, knowing that Xcode 6.4 comes with the Swift 1.2 compiler, while Xcode 7 has Swift 2.0 (which has breaking language changes).

mikemcquaid:
I don't see the use-case for now as we don't have any support for Homebrew to depend on a particular version of Swift currently. We'd accept a PR for that if you also changed a couple of Swift formulae to make use of it, too. Thanks!

beltex:
Hi @mikemcquaid!

Sorry about the late response!

> I don't see the use-case for now as we don't have any support for Homebrew to depend on a particular version of Swift currently.

Currently this is done indirectly by depending on Xcode version. Some example formulae:

- [Carthage](https://github.com/Homebrew/homebrew/blob/a5fc72d4e97ef507e6374b1fa7c71cc2653ff3f3/Library/Formula/carthage.rb)
- [SourceKitten](https://github.com/Homebrew/homebrew/blob/1777fc04b5afbc1ed309361c1324cfe62ce32f62/Library/Formula/sourcekitten.rb)
- [SwiftLint](https://github.com/Homebrew/homebrew/blob/1777fc04b5afbc1ed309361c1324cfe62ce32f62/Library/Formula/swiftlint.rb)

[...]

PR #43684, opened 2015-09-08
SourceKitten 0.5.0

jpsim:
Hopefully Homebrew supports formulas requiring Xcode 7 Beta 6. Thanks!

### DIFF
--- a/Library/Formula/sourcekitten.rb
+++ b/Library/Formula/sourcekitten.rb
@@ -1,7 +1,7 @@
 class Sourcekitten < Formula
   desc "Framework and command-line tool for interacting with SourceKit"
   homepage "https://github.com/jpsim/SourceKitten"
-  url "https://github.com/jpsim/SourceKitten.git", :tag => "0.4.5", :revision => "35508f8c95ecd220719db45b3955d1821f523b7d"
+  url "https://github.com/jpsim/SourceKitten.git", :tag => "0.5.0", :revision => "4106ae36b4fc7a98406ed9bdd1d64074a04a35be"
   head "https://github.com/jpsim/SourceKitten.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> Hopefully Homebrew supports formulas requiring Xcode 7 Beta 6.

Not until it's a stable version, unfortunately.

jpsim:
That's understandable. Should I keep this PR open until then, or close it?

mikemcquaid:
Up to you, we're happy for you to leave it open

jpsim:
Does the Xcode 7 GM release count as a "stable version"?

mikemcquaid:
@jpsim Unfortunately not. It will when it starts getting delivered through the App Store.

apjanke:
That `depends_on :xcode => ["6.4", :build]` doesn't seem to actually catch the version mismatch. I ran the current 0.4.5 formula on OS X 10.11 with Xcode 7 GM, and `brew` didn't complain about the Xcode version; it just ran the build, which failed with an obscure-sounding compiler error. Is this expected? Can `brew` actually check the Xcode version, or is that just documentation?

beltex:
Hello @apjanke!

> I ran the current 0.4.5 formula on OS X 10.11 with Xcode 7 GM, and brew didn't complain about the Xcode version; it just ran the build, which failed with an obscure-sounding compiler error.

I believe `depends_on :xcode` checks that there is a version that is greater than or equal to the one listed, hence the check passes. The compiler error occurs because Xcode 6.4 comes with the Swift 1.2 compiler (which version 0.4.5 of SourceKitten required), while Xcode 7 comes with the Swift 2.0 compiler which has breaking changes.

The other thing is `depends_on :xcode` doesn't check if the Xcode version in question is active, if it is set via `xcode-select -s` that is, thus you would still hit the compiler error even if you had Xcode 6.4 installed as well, but Xcode 7 was active.

The long term solution would be to instead depend on Swift compiler version. Once Swift goes open-source later this year, this can be done. See https://github.com/Homebrew/homebrew/issues/42864#issuecomment-133078788 for more.

jpsim:
@mikemcquaid now that Xcode 7 is officially released in the App Store, this PR should be good to go once your CI infrastructure is in place!

mikemcquaid:
@BrewTestBot test this please

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/23df9fc70fc56e4882751540ad60f2d8855651c1. Thank you for your contribution to Homebrew @jpsim!

Issue #43679, opened 2015-09-07
Re-creating branch from #40640


nnutter:
I would have liked to have commented on #40640 with easy instructions to restore the branch from the PR but @dunn locked it.  For the record I find that frustrating from the perspective of someone trying to help.

@knakayama could checkout #40640 [locally][1] to recreate the branch:

```
# remotes: origin is Homebrew, knakayama is fork
$ git fetch origin pull/40640/head:add-procmail-formula
$ git push knakayama add-procmail-formula
```

[1]: https://help.github.com/articles/checking-out-pull-requests-locally/

DomT4:
I suspect Alex locked it because people were pinging Jack, who is no longer involved with the project and we try to respect that by not nudging him.

You don't really need to recreate the branch. It's easier in this case where the PR has been closed for a while just to create a new branch locally &

```
curl -O https://raw.githubusercontent.com/knakayama/homebrew/a5016ebba27e166f72bc88cd0a032410ca78c1e4/Library/Formula/procmail.rb
```

And then push + resubmit it.

We welcome a PR on this, as Mike said.

DomT4:
Closed via https://github.com/Homebrew/homebrew/pull/40640#issuecomment-138357040.

dunn:
Sorry about that @nnutter. I was trying to encourage a fresh PR, rather than trying to discourage discussion. Old PRs get buried and are harder to keep track of, so we don't mind people opening new issues.

knakayama:
Sorry for my bad manner. I re-submit PR(#43686).

PR #43686, opened 2015-09-08
procmail: 3.22 (new formula)

knakayama:
Looks like procmail is being removed in [El Capitan.](http://www.shirt-pocket.com/blog/index.php/shadedgrey/comments/oh_yeah_well_lock_this/) And there is some demand for this PR(#43679), so I re-submit PR.

### DIFF
--- a/Library/Formula/procmail.rb
+++ b/Library/Formula/procmail.rb
@@ -19,6 +19,14 @@ class Procmail < Formula
   end
 end
 
+# The getline() used in procmail original code is conflict with Mac OS X
+# system function, so need to replace "getline" with other name(i.e: get_line).
+# Please see the following links for more info.
+# error message:
+# https://trac.macports.org/ticket/30353
+# about getline():
+# https://developer.apple.com/library/prerelease/mac/documentation/Darwin/Reference/ManPages/man3/getline.3.html
+
 __END__
 diff -Naur procmail-3.22-getline/src/fields.c procmail-3.22/src/fields.c
 --- procmail-3.22-getline/src/fields.c	2015-06-13 14:43:11.000000000 +0900
--- a/Library/Formula/procmail.rb
+++ b/Library/Formula/procmail.rb
@@ -7,6 +7,8 @@ class Procmail < Formula
 
   depends_on "autoconf" => :build
 
+  keg_only :provided_pre_el_capitan
+
   patch :p1, :DATA
 
   def install
--- a/Library/Formula/procmail.rb
+++ b/Library/Formula/procmail.rb
@@ -1,5 +1,5 @@
 class Procmail < Formula
-  desc "procmail: sort incoming mail into separate folders/files"
+  desc "Sort incoming mail into separate folders/files"
   homepage "http://www.procmail.org/"
   url "ftp://ftp.procmail.net/pub/procmail/procmail-3.22.tar.gz"
   mirror "ftp://ftp.ucsb.edu/pub/mirrors/procmail/procmail-3.22.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/procmail.rb
```diff
@@ -0,0 +1,78 @@
+class Procmail < Formula
+  desc "procmail: sort incoming mail into separate folders/files"
+  homepage "http://www.procmail.org/"
+  url "ftp://ftp.procmail.net/pub/procmail/procmail-3.22.tar.gz"
+  mirror "ftp://ftp.ucsb.edu/pub/mirrors/procmail/procmail-3.22.tar.gz"
+  sha256 "087c75b34dd33d8b9df5afe9e42801c9395f4bf373a784d9bc97153b0062e117"
+
+  depends_on "autoconf" => :build
+
+  patch :p1, :DATA
```
bfontaine: Please add a comment explaining why this patch is necessary as well as a link to the upstream submission. Thanks!

DomT4:
@Homebrew/owners Is it worth creating a `keg_only :provided_pre_el_capitan` here?

It's not any newer or older than the system version, there hasn't been a release in 14 years, but don't know if we want to prevent it overriding the system version on <10.11 regardless.

mikemcquaid:
> @Homebrew/owners Is it worth creating a keg_only :provided_pre_el_capitan here?

:+1:

DomT4:
https://github.com/Homebrew/homebrew/commit/ad1fcfc77c0871fe02366687be32c766bf89df20

@knakayama If you rebase your PR against the master branch here you should be able to use:

```ruby
keg_only :provided_pre_el_capitan
```

in the formula, one empty line underneath the `depends_on` line. Thanks!

knakayama:
I add comments about patch add `keg_only` line. My English skill is terrible, so it may be a little strange...

> as well as a link to the upstream submission

@bfontaine 
It looks like procmail has not been [developed](http://www.procmail.org/procmail.HISTORY.html) (3.22 is released in 2001). So I have not submitted patch because it will not be merged and Macports procmail has [the same approach](https://github.com/dstrctrng/macports/blob/master/rsync.macports.org/release/tarballs/ports/mail/procmail/files/getline.patch).

mikemcquaid:
> It looks like procmail has not been developed (3.22 is released in 2001). So I have not submitted patch because it will not be merged and Macports procmail has the same approach.

We are different to MacPorts and require all patches to be submitted upstream.

knakayama:
> We are different to MacPorts and require all patches to be submitted upstream.

OK, I appreciate the homebrew way. But I have no time to submit patches for now.
Is there anyone who submit patches? If no one, I don't mind that this PR is rejected. Thanks.

bmike:
I'd like to help move this forward. What's the path forward?

bfontaine:
@bmike we need to submit the patch to upstream.

Library/Formula/procmail.rb
```diff
@@ -0,0 +1,88 @@
+class Procmail < Formula
+  desc "procmail: sort incoming mail into separate folders/files"
```
DomT4: You don't need the formula name in here. It'll be prepended automatically.

```ruby
desc "Sort incoming mail into separate folders/files"
```

is fine.

knakayama:
Thanks. I modify the description.

afh:
How about using the sources that Apple provides at [http://opensource.apple.com](http://www.opensource.apple.com/tarballs/procmail) and that were used for previous versions of Mac OS X (see #44105)?

knakayama:
> How about using the sources that Apple provides at http://opensource.apple.com and that were used for previous versions of Mac OS X (see #44105)?

@Homebrew/owners
If use this source, does not need **upstream submission**?

mikemcquaid:
> If use this source, does not need upstream submission?

If it's Apple's: nope.

afh:
I'd like to mention that there's a procmail formula pending in #44105. It has functional tests and compiles to `brew audit --strict`. An issue one might have with it is that is uses a different (Apple's) version number (14) than the procmail upstream (3.22). Just sayin'… :wink:

PR #43690, opened 2015-09-08
Git related improvement

xu-cheng:


mikemcquaid:
:+1:. The only thing you might want to verify is if e.g. `git` is installed by Homebrew it can still get picked up after that.

xu-cheng:
>  The only thing you might want to verify is if e.g. git is installed by Homebrew it can still get picked up after that.

Right, I almost forget this part.

xu-cheng:
> The only thing you might want to verify is if e.g. git is installed by Homebrew it can still get picked up after that.

Fixed in https://github.com/Homebrew/homebrew/commit/540d61edfe8de16667de4eb72a777b449bbe9ea2

mikemcquaid:
Bit hacky but I don't see a better approach :+1:

xu-cheng:
I seems accidentally nuke my PR during rebase.

xu-cheng:
resubmit one in #43693

PR #43691, opened 2015-09-08
quiet_system: don't print cmd even when verbose mode

xu-cheng:
It's kinda annoying that quiet_system wasn't really quiet.


mikemcquaid:
:+1:

PR #43693, opened 2015-09-08
git related improvement

xu-cheng:


Library/Homebrew/cmd/doctor.rb
```diff
@@ -914,12 +906,11 @@ def check_git_newline_settings
   end
 
   def check_git_origin
-    return unless git? && (HOMEBREW_REPOSITORY/".git").exist?
+    return unless Utils.git_available? && (HOMEBREW_REPOSITORY/".git").exist?
```
mikemcquaid: Just a nit (that you didn't do): I always find `unless ... &&` hard to parse compared to `if ! ...`

mikemcquaid:
:+1:

Library/Homebrew/cmd/doctor.rb
xu-cheng: Fixed and shipped.

PR #43700, opened 2015-09-08
Git Town 0.7.3 (new formula)

kevgo:
We have been running on [our own tap](https://github.com/Originate/homebrew-gittown) for a while. Git Town is now stable and in widespread use, with external contributors, so we wanted to make this available on the main tap for easier installation.

Any guidance on improving the formula is appreciated! Right now the audit fails with 

```
git-town:
 * Non-executables were installed to "/usr/local/Cellar/git-town/0.7.3/bin"
The offending files are:
  /usr/local/Cellar/git-town/0.7.3/bin/helpers
```

Git Town is a Bash script that loads additional code from `src/helpers` and `src/drivers`. Is there a better way to do this?

Thanks for all the amazing work on HomeBrew, we are honored to contribute Git Town!

### DIFF
--- a/Library/Formula/git-town.rb
+++ b/Library/Formula/git-town.rb
@@ -1,0 +1,40 @@
+class GitTown < Formula
+  desc "High-level command-line interface for Git"
+  homepage "http://www.git-town.com"
+  url "https://github.com/Originate/git-town/archive/v0.7.3.tar.gz"
+  sha256 "947fd6dac603dcb02467cc2969ef26db1ccdc72312b6b0d70fb8d11179eb09ba"
+
+  depends_on "dialog" => :recommended
+
+  def install
+    # Install the source
+    libexec.install Dir["src/*"]
+
+    # Symlink the executables
+    bin.install_symlink Dir["#{libexec}/git-*"]
+    bin.install_symlink "#{libexec}/helpers"
+    bin.install_symlink "#{libexec}/drivers"
+
+    # Install the man pages
+    man1.install Dir["man/man1/*"]
+  end
+
+  def caveats
+    <<-EOS.undent
+      To install the Fish shell autocompletions,
+      run "git town install-fish-autocompletion"
+      in the terminal.
+
+      To install the completions manually, make
+      #{libexec}/autocomplete/git.fish
+      available as ~/.config/fish/completions/git.fish.
+      In a standard setup, this looks like:
+      mkdir -p ~/.config/fish/completions
+      ln -s #{libexec}/autocomplete/git.fish ~/.config/fish/completions/git.fish
+    EOS
+  end
+
+  test do
+    system "#{bin}/git-town", "version"
+  end
+end
--- a/Library/Formula/git-town.rb
+++ b/Library/Formula/git-town.rb
@@ -7,15 +7,8 @@ class GitTown < Formula
   depends_on "dialog" => :recommended
 
   def install
-    # Install the source
     libexec.install Dir["src/*"]
-
-    # Symlink the executables
-    bin.install_symlink Dir["#{libexec}/git-*"]
-    bin.install_symlink "#{libexec}/helpers"
-    bin.install_symlink "#{libexec}/drivers"
-
-    # Install the man pages
+    bin.write_exec_script Dir["#{libexec}/git-*"]
     man1.install Dir["man/man1/*"]
   end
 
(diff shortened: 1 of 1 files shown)


Library/Formula/git-town.rb
```diff
@@ -0,0 +1,40 @@
...
+    bin.install_symlink "#{libexec}/drivers"
+
+    # Install the man pages
+    man1.install Dir["man/man1/*"]
+  end
+
+  def caveats
+    <<-EOS.undent
+      To install the Fish shell autocompletions,
+      run "git town install-fish-autocompletion"
```
mikemcquaid: Can these be installed automatically?

Library/Formula/git-town.rb
```diff
@@ -0,0 +1,40 @@
...
+      in the terminal.
+
+      To install the completions manually, make
+      #{libexec}/autocomplete/git.fish
+      available as ~/.config/fish/completions/git.fish.
+      In a standard setup, this looks like:
+      mkdir -p ~/.config/fish/completions
+      ln -s #{libexec}/autocomplete/git.fish ~/.config/fish/completions/git.fish
```
mikemcquaid: Would be good to just do this automatically.

Library/Formula/git-town.rb
```diff
@@ -0,0 +1,40 @@
...
+  def install
+    # Install the source
+    libexec.install Dir["src/*"]
+
+    # Symlink the executables
+    bin.install_symlink Dir["#{libexec}/git-*"]
+    bin.install_symlink "#{libexec}/helpers"
+    bin.install_symlink "#{libexec}/drivers"
+
+    # Install the man pages
```
mikemcquaid: Can scrap these comments.

Library/Formula/git-town.rb
```diff
@@ -0,0 +1,40 @@
...
+  depends_on "dialog" => :recommended
+
+  def install
+    # Install the source
+    libexec.install Dir["src/*"]
+
+    # Symlink the executables
+    bin.install_symlink Dir["#{libexec}/git-*"]
+    bin.install_symlink "#{libexec}/helpers"
+    bin.install_symlink "#{libexec}/drivers"
```
mikemcquaid: These names are too generic for `bin`.

mikemcquaid: (`helpers` and `drivers`)

Library/Formula/git-town.rb
```diff
@@ -0,0 +1,40 @@
...
+      #{libexec}/autocomplete/git.fish
+      available as ~/.config/fish/completions/git.fish.
+      In a standard setup, this looks like:
+      mkdir -p ~/.config/fish/completions
+      ln -s #{libexec}/autocomplete/git.fish ~/.config/fish/completions/git.fish
+    EOS
+  end
+
+  test do
+    system "#{bin}/git-town", "version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/git-town.rb
DomT4: Instead of dumping these all in the `bin`, if you do this instead it should solve the audit warning and leave the executables available:


```ruby
bin.write_exec_script Dir["#{libexec}/git-*"]
```

The tools will still technically live in `libexec/"bin"` so they'll know where to find the helpers/drivers in the directory they expect.

Library/Formula/git-town.rb
kevgo: Its possible to install this automatically, but I'd be hesitant to do that without user confirmation, because (a) this only applies to Fish shell users, and (b) this possibly overrides existing Git autocompletion set up for Fish shell. I'm happy to remove this notice altogether and rely on the installation instructions on the website, though, if that helps keep this formula clean.

kevgo: We have simplified this to list only a few pointers to optional additional install scripts. Hope this addresses this issue.  :)

Library/Formula/git-town.rb
kevgo: Removed these steps.

Library/Formula/git-town.rb
kevgo: Done

Library/Formula/git-town.rb
kevgo: Great suggestion, thanks! Implemented using `write_exec_script`

Library/Formula/git-town.rb
kevgo: Good point. We are now running a real Git Town command that loads the whole tool, and calculates and displays real data. Thanks for the feedback!

kevgo:
@mikemcquaid @DomT4 All comments addressed. Want to take another look?

DomT4:
Seems fine to me. I tweaked the test a bit from https://github.com/Originate/homebrew/commit/4efbb3efaa263f51086ca353e2a821540a29c5d6 for the same reasons I did in https://github.com/Homebrew/homebrew/commit/902d7668bf7d5466682d953d65d04a38541244e8, but that aside.

Merged in https://github.com/Homebrew/homebrew/commit/07266c680985c0430173f4122bf3486e5a24db99 & https://github.com/Homebrew/homebrew/commit/f0e9ff5741435d1f77bd08802fa32ce43ae2d959. Thank you for your contribution to Homebrew; we appreciate it! :smiley_cat:

kevgo:
Thanks!!  :)

PR #43706, opened 2015-09-08
CoreLocationCLI 1.0.0 (new formula)

adaofeliz:
Command line program to print location information from CoreLocation.

More info: https://github.com/fulldecent/corelocationcli

### DIFF
--- a/Library/Formula/core-location-cli.rb
+++ b/Library/Formula/core-location-cli.rb
@@ -1,0 +1,13 @@
+class CoreLocationCli < Formula
+  desc "Prints location information from CoreLocation"
+  homepage "https://github.com/fulldecent/corelocationcli"
+  url "https://github.com/fulldecent/corelocationcli/archive/v1.0.0.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/core-location-cli.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  depends_on :xcode => :build
+
+  def install
+    xcodebuild "-project", "CoreLocationCLI.xcodeproj", "SYMROOT=build", "-sdk", "macosx#{MacOS.version}"
+    bin.install "build/Release/CoreLocationCLI"
+  end
+
+  test do
+    system "#{bin}/CoreLocationCLI", "-h"
```
bfontaine: Can’t we write a test more substantial than e.g. `-h`/`--help`/`--version`?

adaofeliz: Not really, in this case we can only test that CoreLocationCLI is installed.
CoreLocation framework needs a user's interaction to allow being used: https://dl.dropboxusercontent.com/spa/y32fwqxnzbyqfvp/zd1l7ui1.png

Issue #43713, opened 2015-09-08
update blob/master/Library/Formula/node.rb to Node v4.0.0 stable


cdbattags:
Node v4.0.0 was just released! Yay!

ScottSmith95:
:+1:

DomT4:
Please file PRs, not issues. Thanks.

PR #36369, opened 2015-01-29
Graduate iojs out of keg-only

bcomnes:
This is an alternative to https://github.com/Homebrew/homebrew/pull/36343 that brings the following qualities:

### This PR:
- Graduates iojs out of keg-only
- Conflicts iojs and node (this seems to be what the iojs team wants)
- installs npm from the tarball every time node or iojs is installed.
 - No patching is performed. Don't merge until compatibility is reached.

### Issues that still need resolution
- [x] Move npm install code to `Library/Homebrew/language/javascript.rb` and share between node and iojs formula.  (Anyone feel free to hop in on this one!  I don't have time right now)
- [ ] Wait for `node-gyp` patch to land 

### Why this arrangement works well
- users can either brew install node OR brew install iojs and everything "just works".
- the npm install is basically identical between the two ~~(except for the current patch)~~.
- nuking the npm install during updates isn't a big deal.
- npm can still update itself.
- ~~npm updating itself wont break patch~~

### Wackiness
- ~~Massive DRY violation.  Can the `npm` install code be separated out somehow?~~
- ~~Patching resources is a tad funky.  This would be cleaned up by https://github.com/Homebrew/homebrew/issues/31508#issuecomment-70918737~~
- ~~Patching one formula but not the other breaks the linking/unlinking dance with both formula installed.~~  
- npm installs on all node/iojs installs+updates and nukes whatever is in `/usr/local/lib/node_modules/npm`

#### Current iojs formula discussions
- ~~https://github.com/Homebrew/homebrew/pull/36357 Adds flag to patch npm install for iojs compatibility~~
- ~~https://github.com/Homebrew/homebrew/pull/36343 Node, iojs and npm formula resolution~~
- https://github.com/Homebrew/homebrew/pull/36369 Graduate iojs out of keg-only

#### working iojs taps
- https://github.com/aredridel/homebrew-iojs

#### Past iojs formula discussions
- https://github.com/Homebrew/homebrew/pull/35853
- https://github.com/Homebrew/homebrew/pull/36039
- https://github.com/Homebrew/homebrew/pull/36193

####  npm formula issues:
- https://github.com/Homebrew/homebrew/issues/3762
- https://github.com/Homebrew/homebrew/issues/22408
- https://github.com/Homebrew/homebrew/pull/27479
- https://github.com/Homebrew/homebrew/pull/28075

PR #43714, opened 2015-09-08
Upgrade Node.js to v4.0.0

killswitch:
Today the Node.js team has released `v4.0.0` so this formula takes from the `io.js` formula since `v4.0.0` is coming from the `io.js` codebase.

https://nodejs.org/en/blog/release/v4.0.0/

### DIFF
--- a/Library/Formula/node.rb
+++ b/Library/Formula/node.rb
@@ -1,130 +1,23 @@
-# Note that x.even are stable releases, x.odd are devel releases
 class Node < Formula
   desc "Platform built on Chrome's JavaScript runtime to build network applications"
   homepage "https://nodejs.org/"
-  url "https://nodejs.org/dist/v0.12.7/node-v0.12.7.tar.gz"
-  sha256 "b23d64df051c9c969b0c583f802d5d71de342e53067127a5061415be7e12f39d"
-  head "https://github.com/nodejs/node.git", :branch => "v0.12"
-  revision 1
-
-  bottle do
-    sha256 "15689cc474a79975eaa6d791b24e6fa021494839c9b691ac307d74acefc5f834" => :yosemite
-    sha256 "a7a7d37c6e5088ed3f58b867d4d246851715d3a4f2f3b4b3c40cc7452ff6728c" => :mavericks
-    sha256 "374f3c5b576e4173590b8413e9941df121a84f46bd48161fc758e1f7d42e0402" => :mountain_lion
-  end
+  url "https://nodejs.org/dist/v4.0.0/node-v4.0.0.tar.gz"
+  sha256 "e110e5a066f3a6fe565ede7dd66f3727384b9b5c5fbf46f8db723d726e2f5900"
 
   option "with-debug", "Build with debugger hooks"
-  option "without-npm", "npm will not be installed"
-  option "without-completion", "npm bash completion will not be installed"
+  option "with-icu4c", "Build with Intl (icu4c) support"
 
-  deprecated_option "enable-debug" => "with-debug"
-
-  depends_on :python => :build if MacOS.version <= :snow_leopard
   depends_on "pkg-config" => :build
-  depends_on "openssl" => :optional
-
-  # https://github.com/nodejs/node-v0.x-archive/issues/7919
-  # https://github.com/Homebrew/homebrew/issues/36681
   depends_on "icu4c" => :optional
-
-  fails_with :llvm do
-    build 2326
-  end
-
-  resource "npm" do
-    url "https://registry.npmjs.org/npm/-/npm-2.14.2.tgz"
-    sha256 "592029e3406cbbaf249135e18212fab91db1601f991f61b4b2a03580311a066e"
-  end
+  depends_on :python => :build if MacOS.version <= :snow_leopard
 
   def install
-    args = %W[--prefix=#{prefix} --without-npm]
+    args = %W[--prefix=#{prefix}]
     args << "--debug" if build.with? "debug"
     args << "--with-intl=system-icu" if build.with? "icu4c"
 
-    if build.with? "openssl"
-      args << "--shared-openssl"
-    else
-      args << "--without-ssl2" << "--without-ssl3"
-    end
-
     system "./configure", *args
     system "make", "install"
-
-    if build.with? "npm"
-      resource("npm").stage buildpath/"npm_install"
-
-      # make sure npm can find node
-      ENV.prepend_path "PATH", bin
-      # set log level temporarily for npm's `make install`
-      ENV["NPM_CONFIG_LOGLEVEL"] = "verbose"
-
-      cd buildpath/"npm_install" do
-        system "./configure", "--prefix=#{libexec}/npm"
-        system "make", "install"
-      end
-
-      if build.with? "completion"
-        bash_completion.install \
-          buildpath/"npm_install/lib/utils/completion.sh" => "npm"
-      end
-    end
-  end
-
-  def post_install
-    return if build.without? "npm"
-
-    node_modules = HOMEBREW_PREFIX/"lib/node_modules"
-    node_modules.mkpath
-    npm_exec = node_modules/"npm/bin/npm-cli.js"
-    # Kill npm but preserve all other modules across node updates/upgrades.
-    rm_rf node_modules/"npm"
-
-    cp_r libexec/"npm/lib/node_modules/npm", node_modules
-    # This symlink doesn't hop into homebrew_prefix/bin automatically so
-    # remove it and make our own. This is a small consequence of our bottle
-    # npm make install workaround. All other installs **do** symlink to
-    # homebrew_prefix/bin correctly. We ln rather than cp this because doing
-    # so mimics npm's normal install.
-    ln_sf npm_exec, "#{HOMEBREW_PREFIX}/bin/npm"
-
-    # Let's do the manpage dance. It's just a jump to the left.
-    # And then a step to the right, with your hand on rm_f.
-    ["man1", "man3", "man5", "man7"].each do |man|
-      # Dirs must exist first: https://github.com/Homebrew/homebrew/issues/35969
-      mkdir_p HOMEBREW_PREFIX/"share/man/#{man}"
-      rm_f Dir[HOMEBREW_PREFIX/"share/man/#{man}/{npm.,npm-,npmrc.}*"]
-      ln_sf Dir[libexec/"npm/lib/node_modules/npm/man/#{man}/npm*"], HOMEBREW_PREFIX/"share/man/#{man}"
-    end
-
-    npm_root = node_modules/"npm"
-    npmrc = npm_root/"npmrc"
-    npmrc.atomic_write("prefix = #{HOMEBREW_PREFIX}\n")
-  end
-
-  def caveats
-    s = ""
-
-    if build.without? "npm"
-      s += <<-EOS.undent
-        Homebrew has NOT installed npm. If you later install it, you should supplement
-        your NODE_PATH with the npm module folder:
-          #{HOMEBREW_PREFIX}/lib/node_modules
-      EOS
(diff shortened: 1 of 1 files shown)


mmwtsn:
:tada:

This fixes #43713

vinkla:
finally

DomT4:
This should be applied to the `iojs` formula in the mean time, yes. The Node formula is remaining on the LTS 0.x.x releases until what was `iojs` no longer requires a patched `npm` to run. Likewise, because it requires a patched `npm` still it is unlikely we'll be adding `npm` support to formerly known as `iojs` **at this point in time**.

killswitch:
Copying @DomT4's comment from https://github.com/Homebrew/homebrew/pull/43716#issuecomment-138699910

> It is likely that the Node formula is going away, or rather being moved to homebrew/versions, and iojs will be renamed nodejs or similar - Once we have an npm that we can just fetch as a tarball and use the same stable mechanism we've been using in node for a long period now. This is not set in stone yet, by any means.

> npm cannot live inside the Cellar path. We tried that before and the end result was incredibly troublesome. Unless that has changed and npm can now live in one place whilst installed modules live and persist somewhere else globally that won't be removed across updates/reinstalls/etc the position on that can't change.

> Essentially, no npm that we cannot control the build location of can be used. On top of that if people install a different node version from versions the installation of the npm shipped with that formula should not break the main node-esque formula in the core. If people update or downgrade their npm version locally it should not break the node-esque formula.

> Package managing another package manager (npm) is difficult and delicate, and we appreciate this is frustrating. We absolutely will move forwards as quickly as we can continue to provide stability to end users, and appreciate patience in that process.

killswitch:
I never thought about that stuff, good thinking. I have this formula installing Node.js `v4.0.0` right now and don't seem to see any real issues yet. I had stuff already installed globally from when I was using `v0.12.7` and all those still seem to work.

inuscript:
:+1:

DomT4:
Please don't add thumbs to this conversation. It'll be merged when all parties involved are happy. Some extra discussion quoted by @killswitch is in #43716 and I'll reply to the latest point there at some point in the next few hours probably.

nvartolomei:
:tada:

mourner:
Great points! Were they brought up to NPM maintainers in some public ticket? The team there is pretty nice and responsive. I'd love to see what they have to say on the issue.

Library/Formula/node.rb
```diff
@@ -1,130 +1,23 @@
...
-    sha256 "a7a7d37c6e5088ed3f58b867d4d246851715d3a4f2f3b4b3c40cc7452ff6728c" => :mavericks
-    sha256 "374f3c5b576e4173590b8413e9941df121a84f46bd48161fc758e1f7d42e0402" => :mountain_lion
-  end
+  url "https://nodejs.org/dist/v4.0.0/node-v4.0.0.tar.gz"
+  sha256 "e110e5a066f3a6fe565ede7dd66f3727384b9b5c5fbf46f8db723d726e2f5900"
 
   option "with-debug", "Build with debugger hooks"
-  option "without-npm", "npm will not be installed"
-  option "without-completion", "npm bash completion will not be installed"
+  option "with-icu4c", "Build with Intl (icu4c) support"
```
geoff-codes: Why not simply remove this option as well, and make this the default? If it adds no additional _homebrew_ dependency, we're basically just getting i18n support "for free" by linking to the system libicucore.A.dylib, right? Why would one _not_ want to do this?

geoff-codes:
Otherwise, concur with @hmalphettes that this is the way to go over #43725. I would give it an upward thumb, but apparently that is not allowed.

Library/Formula/node.rb
```diff
@@ -1,130 +1,23 @@
...
 
-  deprecated_option "enable-debug" => "with-debug"
-
-  depends_on :python => :build if MacOS.version <= :snow_leopard
   depends_on "pkg-config" => :build
-  depends_on "openssl" => :optional
-
-  # https://github.com/nodejs/node-v0.x-archive/issues/7919
-  # https://github.com/Homebrew/homebrew/issues/36681
   depends_on "icu4c" => :optional
```
geoff-codes: I.e. (in re: https://github.com/Homebrew/homebrew/pull/43714#discussion_r39027144) one would remove this line as well, so the homebrew icu4c formula doesn't override the system ICU lib.

DomT4:
Left a comment in the other thread very early this morning along the lines of the blockers on this at the moment: https://github.com/Homebrew/homebrew/pull/43716#issuecomment-138777973

> Were they brought up to NPM maintainers in some public ticket?

It was discussed heavily in various Homebrew tickets, including #3762, #22408, #27479, #28075 & #36369. Isaacs & Othiym23 have both had a lot of immensely valuable input that has been really useful to us in resolving various sticky points over the long-term.

We try to loop the `npm` team in on various discussions as obviously Homebrew has been a driver of issues to them in the past and an open relationship between the two teams helps reduce both our maintenance burdens.

Library/Formula/node.rb
```diff
@@ -1,130 +1,23 @@
...
-  url "https://nodejs.org/dist/v0.12.7/node-v0.12.7.tar.gz"
-  sha256 "b23d64df051c9c969b0c583f802d5d71de342e53067127a5061415be7e12f39d"
-  head "https://github.com/nodejs/node.git", :branch => "v0.12"
-  revision 1
-
-  bottle do
-    sha256 "15689cc474a79975eaa6d791b24e6fa021494839c9b691ac307d74acefc5f834" => :yosemite
-    sha256 "a7a7d37c6e5088ed3f58b867d4d246851715d3a4f2f3b4b3c40cc7452ff6728c" => :mavericks
-    sha256 "374f3c5b576e4173590b8413e9941df121a84f46bd48161fc758e1f7d42e0402" => :mountain_lion
-  end
```
mikemcquaid: Leave the bottle-block as-is, it'll be updated when we pull.

Library/Formula/node.rb
mikemcquaid: As mentioned: I believe this uses the system `icu4c` without this option. Is there an advantage to using this option over that?

Library/Formula/node.rb
```diff
@@ -1,130 +1,23 @@
...
+  sha256 "e110e5a066f3a6fe565ede7dd66f3727384b9b5c5fbf46f8db723d726e2f5900"
 
   option "with-debug", "Build with debugger hooks"
-  option "without-npm", "npm will not be installed"
-  option "without-completion", "npm bash completion will not be installed"
+  option "with-icu4c", "Build with Intl (icu4c) support"
 
-  deprecated_option "enable-debug" => "with-debug"
-
-  depends_on :python => :build if MacOS.version <= :snow_leopard
```
mikemcquaid: Are you sure this doesn't need Python 2.7 any more? That seems surprising.

Library/Formula/node.rb
```diff
@@ -1,130 +1,23 @@
...
-  end
+  url "https://nodejs.org/dist/v4.0.0/node-v4.0.0.tar.gz"
+  sha256 "e110e5a066f3a6fe565ede7dd66f3727384b9b5c5fbf46f8db723d726e2f5900"
 
   option "with-debug", "Build with debugger hooks"
-  option "without-npm", "npm will not be installed"
-  option "without-completion", "npm bash completion will not be installed"
+  option "with-icu4c", "Build with Intl (icu4c) support"
 
-  deprecated_option "enable-debug" => "with-debug"
```
mikemcquaid: Leave this as-is.

Library/Formula/node.rb
```diff
@@ -1,130 +1,23 @@
...
-
-  depends_on :python => :build if MacOS.version <= :snow_leopard
   depends_on "pkg-config" => :build
-  depends_on "openssl" => :optional
```
mikemcquaid: This should almost certainly be left as-is if not changed to `:recommended` or default.

Library/Formula/node.rb
mikemcquaid: Ah, I see, it just moved. Generally moving stuff around like that makes it harder to review the PR.

Library/Formula/node.rb
```diff
@@ -1,130 +1,23 @@
...
-
-  # https://github.com/nodejs/node-v0.x-archive/issues/7919
-  # https://github.com/Homebrew/homebrew/issues/36681
   depends_on "icu4c" => :optional
-
-  fails_with :llvm do
-    build 2326
-  end
```
mikemcquaid: Have you tested with this `llvm` version and verified it works?

mikemcquaid:
As a general comment: please don't move stuff around or change stuff in version bumps like this unless they are necessary for it to build (and in those cases: please let us know in the PR description). Otherwise there will be more back-and-forth and it'll take longer to get merged. Thanks!

Library/Formula/node.rb
DomT4: It doesn't, AFAIK. We ran into problems building against the system `icu4c` (or the bits of it OS X ships natively) in the past. If that has changed, I don't mind dropping the dependency, but we need to test this before.

Library/Formula/node.rb
DomT4: It's fine with `:optional` IMO. If Nodejs is going to ensure more timely security releases than `node` managed the reason I put the dependency there becomes semi-moot. I know enough people who use this option though, so I'd like to keep it around.

mikemcquaid: :cool: ignore me!

Library/Formula/node.rb
mikemcquaid: :cool: ignore me!

killswitch:
@mikemcquaid understood. This is the first I've done with any formulas, I just wanted node v4 asap, so I copied the iojs formula and used it for node since node is now using iojs' codebase. 

This works fine for me currently, but I have no issues in someone else taking this over who understands the underlining stuff better as @DomT4 pointed out.

cameronjroe:
Looking forward to this merge. Would love to get Node v4

mikemcquaid:
Just to let people know: we won't be merging this until we have a stable, unpatched `npm` release that supports NodeJS 4.0.0

iffyuva:
this is sad. is there any discussion thread about new unpatched `npm` release which we can follow?

francip:
@mikemcquaid seems npm v2.14.3 will be able to support node v4 unpatched. The current npm in node v4 is  v2.14.2, which still includes a [single patch!](https://github.com/npm/io.js/commit/1ff911c860e5c3f943015744dcfba949d61f4b69) to make the node tests go green

chrmoritz:
@francip The patches you are looking for are https://github.com/nodejs/node/commit/b8341e8a4ef3c29d0a2456ec4fb4528361678de5 and https://github.com/nodejs/node/commit/22097a227d7d0e7dc9ebf9c4c27a1c3cb9d93a62 which are bumping node-gyp to 3.0.0 and 3.0.1.

npm 2.14.4 was release today which officially bundles node-gyp 3.0.1. So we only have to wait one more week for a stable npm which is fully compatible with all possible Node.js release lines without a patch.

killswitch:
Closing as a much better PR exists at https://github.com/Homebrew/homebrew/pull/43973

bcomnes:
In the meantime, we have a working node4 formula in this tap: https://github.com/aredridel/homebrew-iojs

DomT4:
Thanks for the patience & understanding in this PR @killswitch; appreciate the willingness to work with us constructively.

killswitch:
@DomT4 No problem. Thanks for the great work on homebrew. I'd go crazy without it.

jeromedecoster:
@mikemcquaid node 4.1.0 is out today, including updated npm

bcomnes:
@jeromedecoster we are waiting on npm@2.14.4 in https://github.com/Homebrew/homebrew/pull/43973

PR #43716, opened 2015-09-08
node 4.0.0

chrmoritz:
Node.js v4.0.0 was released with new features and performance/memory improvements (see [this blog post](https://nodejs.org/en/blog/release/v4.0.0/)).
This version of node requires node-gyp v3 to work (install native modules). Sadly it will take ~2 more weeks until a official stable version of npm with node-gyp will be released. Thats why a floating patch to upgrade node-gyp is needed until then. I can think of two possibilities how we can handle this patch in the formula:
* download npm as before as a resource and apply the floating patch manually (as there is no patch support for resources yet)
* use the already patched version of npm which ships with node v4 but install it in a separate step to libexec/npm as before

I've used the second approve fore this PR because I see no advantage in manually patching a resource to only get an exact copy of what is already stored on disk in the node sources.
I've updated the test to install a native addon to test node-gyp, removed the npm update test because it would install a broken npm atm and removed the without-ssl arguments because ssl support was dropped in v4.

### DIFF
--- a/Library/Formula/node.rb
+++ b/Library/Formula/node.rb
@@ -2,10 +2,9 @@
 class Node < Formula
   desc "Platform built on Chrome's JavaScript runtime to build network applications"
   homepage "https://nodejs.org/"
-  url "https://nodejs.org/dist/v0.12.7/node-v0.12.7.tar.gz"
-  sha256 "b23d64df051c9c969b0c583f802d5d71de342e53067127a5061415be7e12f39d"
-  head "https://github.com/nodejs/node.git", :branch => "v0.12"
-  revision 1
+  url "https://nodejs.org/dist/v4.0.0/node-v4.0.0.tar.gz"
+  sha256 "e110e5a066f3a6fe565ede7dd66f3727384b9b5c5fbf46f8db723d726e2f5900"
+  head "https://github.com/nodejs/node.git"
 
   bottle do
     sha256 "15689cc474a79975eaa6d791b24e6fa021494839c9b691ac307d74acefc5f834" => :yosemite
@@ -31,11 +30,6 @@ class Node < Formula
     build 2326
   end
 
-  resource "npm" do
-    url "https://registry.npmjs.org/npm/-/npm-2.14.2.tgz"
-    sha256 "592029e3406cbbaf249135e18212fab91db1601f991f61b4b2a03580311a066e"
-  end
-
   def install
     args = %W[--prefix=#{prefix} --without-npm]
     args << "--debug" if build.with? "debug"
@@ -43,29 +37,25 @@ class Node < Formula
 
     if build.with? "openssl"
       args << "--shared-openssl"
-    else
-      args << "--without-ssl2" << "--without-ssl3"
     end
 
     system "./configure", *args
     system "make", "install"
 
     if build.with? "npm"
-      resource("npm").stage buildpath/"npm_install"
-
       # make sure npm can find node
       ENV.prepend_path "PATH", bin
       # set log level temporarily for npm's `make install`
       ENV["NPM_CONFIG_LOGLEVEL"] = "verbose"
 
-      cd buildpath/"npm_install" do
+      cd "deps/npm" do
         system "./configure", "--prefix=#{libexec}/npm"
         system "make", "install"
       end
 
       if build.with? "completion"
         bash_completion.install \
-          buildpath/"npm_install/lib/utils/completion.sh" => "npm"
+          "deps/npm/lib/utils/completion.sh" => "npm"
       end
     end
   end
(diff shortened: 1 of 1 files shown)


Tonkpils:
Dupe of #43714

chrmoritz:
Ah someone else was a little bit faster. Nevertheless the move of npm out of libexec would cause a lot of trouble. There were valid reasons why it was moved there.

killswitch:
@chrmoritz I cloned the `iojs` formula and just upgraded it for nodejs.

DomT4:
This should be applied to the `iojs` formula in the mean time, yes. The Node formula is remaining on the LTS 0.x.x releases until what was `iojs` no longer requires a patched `npm` to run.

chrmoritz:
@DomT4 Node.js v4 will become the new LTS release starting on the 1st October. Also there is a major difference between the old floating npm patches and the new one: The new npm patch upgrades node-gyp to a official release version (node-gyp 3.0.1) which only needs 2-3 more npm release cycles to land in a stable npm version.
Putting Node.js v4 to the iojs formula could be somehow confusing to, because I would except that this formula contains actuall iojs releases like the upcoming iojs 3.3.1.
Closing this now and reopening an updated version in ~2 week after a stable npm version with node-gyp 3 was released.

DomT4:
It is likely that the Node formula is going away, or rather being moved to `homebrew/versions`, and `iojs` will be renamed `nodejs` or similar - Once we have an `npm` that we can just fetch as a tarball and use the same stable mechanism we've been using in `node` for a long period now. This is not set in stone yet, by any means.

`npm` *cannot* live inside the Cellar path. We tried that before and the end result was incredibly troublesome. Unless that has changed and `npm` can now live in one place whilst installed modules live and persist somewhere else globally that won't be removed across updates/reinstalls/etc the position on that can't change.

Essentially, no `npm` that we cannot control the build location of can be used. On top of that if people install a different `node` version from `versions` the installation of the `npm` shipped with that formula *should not* break the main `node`-esque formula in the core. If people update or downgrade their `npm` version locally it should not break the `node`-esque formula.

Package managing another package manager (npm) is difficult and delicate, and we appreciate this is frustrating. We absolutely will move forwards as quickly as we can continue to provide stability to end users, and appreciate patience in that process.

chrmoritz:
Wouldn't it be better to base the node.js v4 formula on the current node formula, because the only difference between the iojs and node formula is that the node formula has all the stuff for separately building npm, bottling it in libexec and extracting it to HOMEBREW_PREFIX/lib/node_modules already implemented and this would be needed once a stable npm with node-gyp is released and can be used in a node v4 formula.

>On top of that if people install a different node version from versions the installation of the npm shipped with that formula should not break the main node-esque formula in the core.
But this will alway be the case because older npm version will never be upwards compatible to never node versions. Installing and to old version of npm will alway break node v4 the same way as it is the case with very old npm versions and node v0.12. The only way this can be partial fixed is by updating all npm version in homebrew versions as soon as a npm with node-gyp 3 is released.

DomT4:
> Wouldn't it be better to base the node.js v4 formula on the current node formula

Honestly, I don't particularly mind either way on the name issue. Whether we repurpose the Node formula or the `iojs` formula, it's not a blocker. Whichever way we move on this, `node012` will be created in `homebrew/versions` as per standard procedure with the Nodes.

If there's going to be a LTS and rapid-development versions similar to the prior Node/iojs situation the `iojs` formula may end up being removed and merged into the `node` formula as a `devel` block now the two projects are operating under the same banner.

What is a blocker is the following, as laid out in the prior comment:

* `npm` must continue living under `HOMEBREW_PREFIX` using the same mechanism we currently use in the Node formula, *unless* `npm` has enabled a workable feature where you can install `npm` in say `/usr/local/Cellar/node` but keep anything the user installs with `npm install` *directly* in `/usr/local` and `/usr/local/lib/node_modules`. User-installed modules persisting cross-update is something we need to keep.

* `npm` must be able to be user-updated without borking the installation. We went through a long period with people breaking the Node formula by updating/downgrading `npm` and it was chaos.

* The `npm` used with `node` must remain interchangeable with the `npm` used by other Node formulae. Yes, if this means we have to install the `npm` release in a couple of weeks into the `versions` formulae and bumping the revisions, we'll do that. To some extent, If we keep the Homebrew side of this relationship stable I'm not going to try and fix every edge case a user could possibly walk into, but people switching Node versions in Homebrew is not an uncommon use case and is something we can control & keep usable.

mikemcquaid:
Agreed with all @DomT4 has said. Between the two of us we've spent a very long time fighting with `npm` to get it installed in a way that works consistently and upstream find acceptable. I'd really rather we didn't change that unless it's fixing reported user issues.

PR #43718, opened 2015-09-08
texpretty 0.02 (new formula)

valkjsaaa:
new formula

### DIFF
--- a/Library/Formula/texpretty.rb
+++ b/Library/Formula/texpretty.rb
@@ -1,0 +1,22 @@
+class Texpretty < Formula
+  desc "prettyprint TeX files"
+  homepage "http://ftp.math.utah.edu/pub/texpretty/"
+  url "http://ftp.math.utah.edu/pub/texpretty/texpretty-0.02.tar.bz2"
+  sha256 "d8e07c29cad7eb7eb8159fd3d03ac0a746ea5f380062af9f1e52061691152526"
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}",
+                          "--mandir=#{man}"
+    system "make"
+    man1.mkpath
+    bin.mkpath
+    system "make", "install"
+  end
+
+  test do
+    system "texpretty", "-a"
+  end
+end


Library/Formula/texpretty.rb
```diff
@@ -0,0 +1,23 @@
...
+
+
+  def install
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}",
+                          "--mandir=#{man}"
+    system "make"
+    mkpath("#{man1}")
```
DomT4: You can just do `man1.mkpath` here. No need for the `#{}` dance.

Library/Formula/texpretty.rb
```diff
@@ -0,0 +1,23 @@
...
+    mkpath("#{bin}")
```
DomT4: Same thing here with `bin.mkpath`

Library/Formula/texpretty.rb
```diff
@@ -0,0 +1,23 @@
+class Texpretty < Formula
+  desc "texpretty - prettyprint TeX files"
```
DomT4: The `desc` output will include the formula name already, so this would be fine with `desc "Prettyprint TeX files"` :)

chdiza:
This software, according to the cited homepage, hasn't been updated since November of 2004.

DomT4:
To be honest, unless it requires patching to compile or has security issues attached to it, I don't really mind. We've older packages than this, including the likely to be added soon #43686.

chdiza:
I'm just looking at our own Acceptable Formulae page, which explicitly requires both that the software be maintained and that there be a stable version.

This package is so obscure that it's not even on CTAN, despite having an author who has plenty of other stuff there (nearly all older than `texpretty`).

Plus I was wrong about how old it is.  The Changelog has the most recent version appearing in 2001, not 2004.

DomT4:
From some of the comments from other maintainers lately we've seemed happy enough to accept effectively abandoned software as long as it isn't patched and still works as expected. As for the stable requirement the website notes the release is stable.

I don't really mind either way here. If another maintainer wants to block this, by all means, I won't stand against them.

chdiza:
> From some of the comments from other maintainers lately we've seemed happy enough to accept effectively abandoned software as long as it isn't patched and still works as expected. 

In that case, the Acceptable Formulae page really needs to be changed.

If this package isn't obscure enough to count as an un-Acceptable Formula, then nothing is.

mikemcquaid:
> In that case, the Acceptable Formulae page really needs to be changed.

As I've said before when we've talked about such things: we'll change it when our behaviour changes. The documentation cannot always 100% reflect the current opinions of the current maintainers.

chdiza:
@DomT4 I think this should go in the TeX tap if it goes anywhere.

DomT4:
We've been keeping the `TeX` tap for things that actually require a `TeX` installation to function. If you can use this formula to just download `TeX` files from the web & play with them without any other `TeX` utility on the system I'm fine with it staying here, *if we accept it*.

I really don't care either way on this. If another maintainer wants to block it or endorse merging it, they're welcome to.

valkjsaaa:
I found this package on Internet for my poster in Tex. It seems that there is no Formula for this package in Homebrew and I make one for it. I'm not really familiar with this package, but I just want to help whoever what to install this package on Mac in the future. Sorry to make you guys bother.

chdiza:
@valkjsaaa You've done nothing wrong, and there's no need to apologize :)

DomT4:
@valkjsaaa No worries at all. Whatever the decision is we really appreciate your willingness to submit formulae to Homebrew!

valkjsaaa:
Thanks!

valkjsaaa:
That was a mistake...

DomT4:
Am going to pass on this for now, but if people come looking for this PR we may reconsider it a bit. Thank you for the submission regardless @valkjsaaa; we appreciate the willingness to contribute to Homebrew.

PR #43719, opened 2015-09-08
Merge the fsharp recipe into the mono recipe

### LINKED ISSUE #36286: F# 3.1 extension paths wrong
I'm opening an issue here, to get a suggestion for how to solve

https://github.com/fsharp/fsharp/issues/328

This is the way these directories look:

```
➜  xbuild git:(upgrade/mono) tree /usr/local/Cellar/mono/3.12.0/lib/mono/xbuild/
/usr/local/Cellar/mono/3.12.0/lib/mono/xbuild/
├── 12.0
│   └── bin
│       ├── MSBuild
│       │   ├── Microsoft.Build.CommonTypes.xsd
│       │   └── Microsoft.Build.Core.xsd
│       ├── Microsoft.Build.Engine.dll -> ../../../gac/Microsoft.Build.Engine/12.0.0.0__b03f5f7f11d50a3a/Microsoft.Build.Engine.dll
│       ├── Microsoft.Build.Framework.dll -> ../../../gac/Microsoft.Build.Framework/12.0.0.0__b03f5f7f11d50a3a/Microsoft.Build.Framework.dll
│       ├── Microsoft.Build.Tasks.v12.0.dll -> ../../../gac/Microsoft.Build.Tasks.v12.0/12.0.0.0__b03f5f7f11d50a3a/Microsoft.Build.Tasks.v12.0.dll
│       ├── Microsoft.Build.Utilities.v12.0.dll -> ../../../gac/Microsoft.Build.Utilities.v12.0/12.0.0.0__b03f5f7f11d50a3a/Microsoft.Build.Utilities.v12.0.dll
│       ├── Microsoft.Build.dll -> ../../../gac/Microsoft.Build/12.0.0.0__b03f5f7f11d50a3a/Microsoft.Build.dll
│       ├── Microsoft.Build.xsd
│       ├── Microsoft.CSharp.targets
│       ├── Microsoft.Common.targets
│       ├── Microsoft.Common.tasks
│       ├── Microsoft.VisualBasic.targets
│       ├── Mono.XBuild.Tasks.dll -> ../../../gac/Mono.XBuild.Tasks/4.0.0.0__0738eb9f132ed756/Mono.XBuild.Tasks.dll
│       ├── xbuild.exe
│       ├── xbuild.exe.config
│       ├── xbuild.exe.mdb
│       └── xbuild.rsp
└── Microsoft
    ├── Portable
    │   ├── Microsoft.Portable.Core.props
    │   ├── Microsoft.Portable.Core.targets
    │   ├── v4.0
    │   │   ├── Microsoft.Portable.CSharp.targets
    │   │   └── Microsoft.Portable.Common.targets
    │   └── v4.5
    │       ├── Microsoft.Portable.CSharp.targets
    │       └── Microsoft.Portable.Common.targets
    └── VisualStudio
        ├── v10.0
        │   └── WebApplications
        │       └── Microsoft.WebApplication.targets
        ├── v11.0
        │   └── WebApplications
        │       └── Microsoft.WebApplication.targets
        └── v9.0
            └── WebApplications
                └── Microsoft.WebApplication.targets

14 directories, 26 files
➜  xbuild git:(upgrade/mono) tree /usr/local/Cellar/fsharp/3.1.1.25/lib/mono/xbuild/
/usr/local/Cellar/fsharp/3.1.1.25/lib/mono/xbuild/
└── Microsoft
    └── VisualStudio
        ├── v
        │   └── FSharp
        │       ├── Microsoft.FSharp.Targets
        │       └── Microsoft.Portable.FSharp.Targets
        └── v12.0
            └── FSharp
                ├── Microsoft.FSharp.Targets
                └── Microsoft.Portable.FSharp.Targets

6 directories, 4 files
```

Is it advisable to let the `fsharp` formula symlink to the `mono` formula, like is suggested?

Should a PR be sent to the mono project to support the new extensions? Or should @rneatherway or @dsyme or someone with push access to the F# repo add in build extensions for mono?

rneatherway:
For reasons related to the way the mono runtime works, the installations
must be "on top" of each other for the MSBuild compilation
infrastructure to function properly. This change matches the official
Mono installer.

Fixes #36286


Library/Formula/mono.rb
```diff
@@ -39,6 +47,15 @@ def install
     # mono-gdb.py and mono-sgen-gdb.py are meant to be loaded by gdb, not to be
     # run directly, so we move them out of bin
     libexec.install bin/"mono-gdb.py", bin/"mono-sgen-gdb.py"
+
+    # Now build and install fsharp as well
+    resource("fsharp").stage() do
+      ENV.prepend_path "PATH", bin
+      ENV["PKG_CONFIG_PATH"] = prefix/"lib/pkgconfig"
```
DomT4: Rather than overwriting the entire `PKG_CONFIG_PATH` here do something like:

```ruby
ENV.prepend_path "PKG_CONFIG_PATH", lib/"pkgconfig"
```
:)

Library/Formula/mono.rb
```diff
@@ -72,6 +89,39 @@ def install
       </Project>
     EOS
     shell_output "#{bin}/xbuild test.csproj"
+
+    # Test that fsharpi is working
+    ENV.prepend_path "PATH", bin
+    output = shell_output %(echo 'printfn "#{test_str}"; exit 0' | #{bin}/fsharpi)
+    assert output.include? test_str
```
DomT4: Can we use an `assert_match` here? We're trying to phase out `output.include?` because the failure message isn't very clear or helpful. Should work as something like:

```ruby
assert_match test_str, output
```
?

Library/Formula/mono.rb
```diff
@@ -72,6 +89,39 @@ def install
...
+          <Reference Include="System" />
+          <Reference Include="FSharp.Core" />
+        </ItemGroup>
+      </Project>
+    EOS
+    (testpath/"Main.fs").write <<-EOS.undent
+      [<EntryPoint>]
+      let main _ = printfn "#{test_str}"; 0
+    EOS
+    shell_output "#{bin}/xbuild test.fsproj"
```
DomT4: Why `shell_output` instead of `system` here?

DomT4:
Few nits, but looks solid and should help reduce the level of pain we've had on this issue for a while now.

dunn:
Gonna have to bring over the pkg-config dependency from F#:
```
==> Cloning https://github.com/fsharp/fsharp.git
Updating /Library/Caches/Homebrew/mono--fsharp--git
git config remote.origin.url https://github.com/fsharp/fsharp.git
git config remote.origin.fetch +refs/tags/3.1.2.5:refs/tags/3.1.2.5
git checkout -f 3.1.2.5 --
HEAD is now at c5e345b... Merge pull request #450 from ctaggart/xamarin
git reset --hard 3.1.2.5
HEAD is now at c5e345b Merge pull request #450 from ctaggart/xamarin
==> Checking out tag 3.1.2.5
==> ./autogen.sh --prefix=/usr/local/Cellar/mono/4.2.0.179
checking whether make sets $(MAKE)... yes
checking for pkg-config... no
configure: error: You need to install pkg-config
```

rneatherway:
Thanks for the review comments, I think I've addressed them.

Library/Formula/mono.rb
```diff
@@ -39,6 +48,15 @@ def install
     # mono-gdb.py and mono-sgen-gdb.py are meant to be loaded by gdb, not to be
     # run directly, so we move them out of bin
     libexec.install bin/"mono-gdb.py", bin/"mono-sgen-gdb.py"
+
+    # Now build and install fsharp as well
+    resource("fsharp").stage() do
+      ENV.prepend_path "PATH", bin
+      ENV.prepend_path "PKG_CONFIG_PATH", prefix/"lib/pkgconfig"
```
DomT4: Does `lib/"pkgconfig"` not work here?

rneatherway: Oh sorry I misread your previous comment and I haven't tried that. I can do so tonight.

rneatherway: That worked fine, so updated.

dunn:
Forgive me if this has already been discussed, but should there be an option `--with-fsharp`, rather than it being included in every mono installation?

DomT4:
If I understand the original post here correctly Mono ships with `fsharp` in the native installer? In which case, an option like `--without-fsharp` would be alright, IMO, but it'd be nice to match the upstream default.

rneatherway:
That is correct, Mono includes fsharp in the Xamarin installer.

DomT4:
At the moment people with `fsharp` already installed here when they upgrade will get linking failures. We could perhaps use `link_overwrite` as a workaround on that. If you install the current `fsharp` formula (Presuming it's possible) and then install your PR locally do you run into any problems linking?

rneatherway:
I can give it a go later or tomorrow, see what happens. Ideally people
would remove the old fsharp recipe first. Could we mark this as conflicting
somehow with an explanation?

DomT4:
>  Could we mark this as conflicting somehow with an explanation?

Sadly, No. `conflicts_with` only works when the other formula still exists. And sadly we can't assume everyone is watching this thread, so we need to hack around problems on their behalf.

dunn:
I think adding a note to the caveats would be appropriate here.

rneatherway:
The output I get is:

```
Could not symlink bin/fsharpc
Target /usr/local/bin/fsharpc
is a symlink belonging to fsharp. You can unlink it:
  brew unlink fsharp

  brew link --overwrite mono

  brew link --overwrite --dry-run mono

/usr/local/bin/fsharpc -> /usr/local/Cellar/fsharp/3.1.2.4/bin/fsharpc
/usr/local/bin/fsharpi -> /usr/local/Cellar/fsharp/3.1.2.4/bin/fsharpi
...
<many files listed here>
```

I've added a corresponding note to the caveats.

DomT4:
How big a conflict list are we talking about?

rneatherway:
At a quick glance around a 100. But the caveat comes out the end offering a resolution. I won't have Mac access for a week from tomorrow so I'll only be able to make cosmetic adjustments. What needs to be done to get this merged?

xu-cheng:
FYI, `link_overwrite` support subdir path and wildcards.

xu-cheng:
For example, you can do `link_overwrite "bin/fsharp*"

DomT4:
> FYI, link_overwrite support subdir path and wildcards.

:+1:. This is what I was wondering about.

Library/Formula/mono.rb
```diff
@@ -39,6 +48,15 @@ def install
     # mono-gdb.py and mono-sgen-gdb.py are meant to be loaded by gdb, not to be
     # run directly, so we move them out of bin
     libexec.install bin/"mono-gdb.py", bin/"mono-sgen-gdb.py"
+
+    # Now build and install fsharp as well
+    resource("fsharp").stage() do
```
xu-cheng: `stage()` -> `stage`

Library/Formula/mono.rb
```diff
@@ -71,12 +89,48 @@ def install
         <Import Project="$(MSBuildBinPath)\\Microsoft.CSharp.targets" />
       </Project>
     EOS
-    shell_output "#{bin}/xbuild test.csproj"
+    system "#{bin}/xbuild test.csproj"
```
xu-cheng: `system "#{bin}/xbuild", "test.csproj"`

Library/Formula/mono.rb
```diff
@@ -71,12 +89,48 @@ def install
...
+
+    # Test that fsharpi is working
+    ENV.prepend_path "PATH", bin
+    output = shell_output %(echo 'printfn "#{test_str}"; exit 0' | #{bin}/fsharpi)
```
xu-cheng: `output = pipe_output("#{bin}/fsharpi", "printfn \"#{test_str}\"; exit 0")`

Library/Formula/mono.rb
```diff
@@ -71,12 +89,48 @@ def install
...
+          <Reference Include="System" />
+          <Reference Include="FSharp.Core" />
+        </ItemGroup>
+      </Project>
+    EOS
+    (testpath/"Main.fs").write <<-EOS.undent
+      [<EntryPoint>]
+      let main _ = printfn "#{test_str}"; 0
+    EOS
+    system "#{bin}/xbuild test.fsproj"
```
xu-cheng: `system "#{bin}/xbuild", "test.fsproj"`

xu-cheng:
I think we should add an option to build without fsharp.

DomT4:
> I think we should add an option to build without fsharp.

`--without-fsharp` seems reasonable; given upstream ships Mono the way this PR does, it seems fair to include it by default with an optional out.

rneatherway:
@xu-cheng thanks for your review comments, I've updated accordingly.

rneatherway:
Hmm, something seems to be going wrong with the build. Is there anything I can do to help?

DomT4:
Not your fault here. If it doesn't loop back around to it I'll stick it back in the queue later. There's a bit of a backlog at the moment; we had some pretty fun CI issues over the last few days.

rneatherway:
Hmm, it looks like the build has failed due to a Jenkins configuration issue:

```
ERROR: Error cloning remote repo 'origin'
hudson.plugins.git.GitException: Could not init /usr/local
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$4.execute(CliGitAPIImpl.java:606)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$2.execute(CliGitAPIImpl.java:462)
	at hudson.plugins.git.GitSCM.retrieveChanges(GitSCM.java:1003)
	at hudson.plugins.git.GitSCM.checkout(GitSCM.java:1043)
	at hudson.scm.SCM.checkout(SCM.java:484)
	at hudson.model.AbstractProject.checkout(AbstractProject.java:1274)
	at hudson.model.AbstractBuild$AbstractBuildExecution.defaultCheckout(AbstractBuild.java:609)
	at jenkins.scm.SCMCheckoutStrategy.checkout(SCMCheckoutStrategy.java:86)
	at hudson.model.AbstractBuild$AbstractBuildExecution.run(AbstractBuild.java:531)
	at hudson.model.Run.execute(Run.java:1738)
	at hudson.matrix.MatrixBuild.run(MatrixBuild.java:301)
	at hudson.model.ResourceController.execute(ResourceController.java:98)
	at hudson.model.Executor.run(Executor.java:381)
Caused by: hudson.plugins.git.GitException: Command "git init /usr/local" returned status code 69:
stdout: 
stderr:
Agreeing to the Xcode/iOS license requires admin privileges, please re-run as root via sudo.



	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:1640)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:1616)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:1612)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommand(CliGitAPIImpl.java:1254)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$4.execute(CliGitAPIImpl.java:604)
	... 12 more
ERROR: null
```

dunn:
Ah, yeah. Re-queued.

rneatherway:
OK, I've tested locally with the `link_overwrite` change, and there are still errors:

```
Could not symlink bin/fsharpc
Target /usr/local/bin/fsharpc
is a symlink belonging to fsharp. You can unlink it:
  brew unlink fsharp
```

Have I made a mistake in my use of this directive?

DomT4:
Any idea why the above wouldn't work @xu-cheng?

xu-cheng:
That is because [current implementation](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/formula.rb#L808-L815) doesn't allow to overwrite files from other keg.

Maybe we should extend the DSL a bit, something like:
```ruby
link_overwrite "bin/fsharpc", :allow_keg => ["fsharp"]
```

rneatherway:
I'm really keen to get this merged in some form or another. Even if the caveat section tells people to relink, surely that is better than the current situation where the `fsharp` recipe is just broken?

xu-cheng:
@rneatherway We only need to update the logic of `link_overwite?` I'm thinking something like this:

```ruby
    # Don't overwrite files belong to other keg except when that keg's formula is deleted.
    begin
      keg = Keg.for(path)
    rescue NotAKegError, Errno::ENOENT
      # file doesn't belong to any keg.
    else
      tab = Tab.for_keg(keg)
      tap = tab.tap
      return false if tap.nil? # this keg doesn't below to any core/tap formula, most likely coming from a DIY install.
      begin
        Formulary.factory("#{tap}/#{keg.name}")
      rescue FormulaAmbiguityError
        # formula is unavailable
      else
        return false # this keg belongs to another formula
      end
    end
```

rneatherway:
@xu-cheng thanks for the suggestion, I will try to find time to test this tonight.

rneatherway:
@xu-cheng: I tried your suggestion (with a couple of `return true` inserted), but I haven't yet been successful. I am installing with `brew install --verbose --debug --build-from-source mono`, which with the latest commit gives:

```
...
rmdir /usr/local/etc/mono/2.0
rmdir /usr/local/etc/mono
/usr/local/Library/brew.rb (Formulary::FormulaLoader): loading /usr/local/Library/Formula/fsharp.rb
Error: No available formula for fsharp
==> Searching formulae...
==> Searching taps...
```

It seems like perhaps the `TapFormulaAmbiguityError` exception is not the right one?

I am using `--build-from-source` because otherwise the bottle is used, which doesn't contain F# and so doesn't try to overwrite the problematic links. However, this takes 20 minutes for each test, which is part of the reason this is taking so long. Is there a way to use a "local bottle" or something like that to speed this up?

I also noticed that `brew link` does not respect `link_overwrite` settings in the formula. Is this deliberate?

Library/Homebrew/formula.rb
```diff
@@ -805,13 +805,23 @@ def skip_clean?(path)
   def link_overwrite?(path)
     # Don't overwrite files not created by Homebrew.
     return false unless path.stat.uid == File.stat(HOMEBREW_BREW_FILE).uid
-    # Don't overwrite files belong to other keg.
+     # Don't overwrite files belong to other keg except when that
+     # keg's formula is deleted.
```
xu-cheng: Can you fix the indention, thanks.

Library/Homebrew/formula.rb
```diff
@@ -805,13 +805,23 @@ def skip_clean?(path)
...
     rescue NotAKegError, Errno::ENOENT
       # file doesn't belong to any keg.
     else
-      return false
+      tab = Tab.for_keg(keg)
+      tap = tab.tap
+      return false if tap.nil? # this keg doesn't below to any core/tap formula, most likely coming from a DIY install.
+      begin
+        Formulary.factory("#{tap}/#{keg.name}")
+      rescue TapFormulaAmbiguityError
```
xu-cheng: Sorry, I miss typed in the comment. It should be `rescue FormulaUnavailableError`

Library/Homebrew/formula.rb
```diff
@@ -805,13 +805,23 @@ def skip_clean?(path)
...
+        return true # formula for this keg is deleted
```
xu-cheng: You should not `return true` here, because there are other check in the below.

Library/Homebrew/formula.rb
```diff
@@ -805,13 +805,23 @@ def skip_clean?(path)
...
+      keg = Keg.for(path)
     rescue NotAKegError, Errno::ENOENT
       # file doesn't belong to any keg.
     else
-      return false
+      tab = Tab.for_keg(keg)
+      tap = tab.tap
+      return false if tap.nil? # this keg doesn't below to any core/tap formula, most likely coming from a DIY install.
+      begin
+        Formulary.factory("#{tap}/#{keg.name}")
```
xu-cheng: After some thought, it should be `Formulary.factory(keg.name)` here. (The reason is to handle the case when formula is migrated from core to tap)

rneatherway:
Right, I'm happy to report that this now works!

rneatherway:
@xu-cheng ping!

xu-cheng:
LGTM. @rneatherway Mind rebase the commits?

Any objection @DomT4?

rneatherway:
@xu-cheng no problem, do you mean to rebase on top of master and/or squash down to a single commit?

xu-cheng:
You can squash it to two commits, one for formula and one for core library.

rneatherway:
@xu-cheng done

DomT4:
> Any objection @DomT4?

Looks good to me. No objections if you're happy here.

xu-cheng:

haf:
Kudos to @rneatherway for making the F# ecosystem better!

PR #43721, opened 2015-09-08
let-me-in 0.0.6 (new formula)

rlister:


Library/Formula/let-me-in.rb
```diff
@@ -0,0 +1,21 @@
...
+  desc "Add my IP to AWS security group(s)"
+  homepage "https://github.com/rlister/let-me-in"
+  url "https://github.com/rlister/let-me-in/archive/v0.0.3.tar.gz"
+  sha256 "538ada5748d513a3a926d7795ddcbe13bf96c99e0921cdc4148ff38f0d196121"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    system "go", "get", "github.com/aws/aws-sdk-go/aws"
```
DomT4: Please use `go_resource`s rather than `go get`. You can find an example [here](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/aptly.rb#L105-L135).

Library/Formula/let-me-in.rb
```diff
@@ -0,0 +1,21 @@
...
+    system "go", "get", "github.com/aws/aws-sdk-go/aws/awserr"
+    system "go", "get", "github.com/aws/aws-sdk-go/service/ec2"
+    system "go", "build", "let-me-in.go"
+    bin.install "let-me-in"
+  end
+
+  test do
+    ## I should write some tests
```
DomT4: `## I should write some tests`

:smile_cat:

Library/Formula/let-me-in.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    system "go", "get", "github.com/aws/aws-sdk-go/aws"
+    system "go", "get", "github.com/aws/aws-sdk-go/aws/awserr"
+    system "go", "get", "github.com/aws/aws-sdk-go/service/ec2"
+    system "go", "build", "let-me-in.go"
+    bin.install "let-me-in"
```
bfontaine: Use `system "go", "build", "-o", "#{bin}/let-me-in", "let-me-in.go"`.

rlister: @bfontaine Trying this I get:

```
==> go build -o /usr/local/Cellar/let-me-in/0.0.3/bin/let-me-in let-me-in.go
Error: No such file or directory - let-me-in
```

Tried adding `mkdir_p bin` to no avail ...

Library/Formula/let-me-in.rb
rlister: @DomT4 What is the correct way to use `go_resource` for `go get` of repo sub-directories? Please see latest commit, this doesn't install the dependency for the `service/ec2` subdir of `aws-sdk-go`.

Library/Formula/let-me-in.rb
```diff
@@ -0,0 +1,36 @@
...
+  desc "Add my IP to AWS security group(s)"
+  homepage "https://github.com/rlister/let-me-in"
+  url "https://github.com/rlister/let-me-in/archive/v0.0.3.tar.gz"
+  sha256 "538ada5748d513a3a926d7795ddcbe13bf96c99e0921cdc4148ff38f0d196121"
+
+  depends_on "go" => :build
+
+  ## this works
+  go_resource "github.com/aws/aws-sdk-go/aws" do
+    url "github.com/aws/aws-sdk-go/aws.git", revision: 'f096b7d61df3d7d6d97f0e701f92616d1ea5420d'
```
DomT4: This needs to be written:

```ruby
url "github.com/aws/aws-sdk-go/aws.git", :revision => "f096b7d61df3d7d6d97f0e701f92616d1ea5420d"
```
The current syntax isn't being recognised.

Library/Formula/let-me-in.rb
```diff
@@ -0,0 +1,36 @@
+class LetMeIn < Formula
```
DomT4: You need:

```ruby
require "language/go"
```

At the top of this formula, one empty line above the `class` name.

Library/Formula/let-me-in.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  ## this DOES NOT work
+  go_resource "github.com/aws/aws-sdk-go/service/ec2" do
+    url "github.com/aws/aws-sdk-go/aws.git", revision: 'f096b7d61df3d7d6d97f0e701f92616d1ea5420d'
+  end
+
+  def install
+    ENV["GOPATH"] = buildpath
+    # system "go", "get", "github.com/aws/aws-sdk-go/aws"
```
DomT4: You need to stage the resources in the buildpath as well. They won't automatically know where they're supposed to be and hop over. Something like this is fairly usual:

```ruby
    ENV["GOPATH"] = buildpath
```

Library/Formula/let-me-in.rb
bfontaine: You have to remove the `bin.install "let-me-in"` line; this is that one which causes the error.

rlister:
@DomT4 This wasn't working with `go_resource` and deps in subdirs, so I have vendored the deps and now build with `godep`. This seems to be reasonably idiomatic (see e.g. `forego.rb`, `telegraf.rb`).

Library/Formula/let-me-in.rb
rlister: @DomT4 I added a test. Not a brilliant one, but better than nothing for now.

Library/Formula/let-me-in.rb
```diff
@@ -0,0 +1,33 @@
...
+      system "go", "install"
+    end
+
+    system "./bin/godep", "go", "build", "-ldflags", "-X main.VERSION=#{version}", "-o", "let-me-in"
+    bin.install "let-me-in"
+  end
+
+  ## not a great test, but better than nothing
+  test do
+    assert `#{bin}/let-me-in -v` =~ /let-me-in #{version}/
```
DomT4: Use `assert_match /let-me-in #{version}/, shell_output("#{bin}/let-me-in -v")` rather than backtics here. It'd be nice to have a more complicated test, if at all possible, but if not feel free to cut the commented out line above :)

rlister: @DomT4 Updated. Agreed on needing a better test, but that will require mocking AWS responses, which I will add in a future version.

bfontaine:

DomT4:
Afraid we'll need to pass on this for now per:

>  * GitHub repository not notable enough (<10 forks, <10 watchers and <20 stars)

Once it has passed that threshold we'd be happy to take another look at a fresh PR. You may wish to host this in your own tap for now, documentation for which you can find [here](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md). Thanks for the submission! :heart:

rlister:
@DomT4 No worries. I already have a tap at https://github.com/rlister/homebrew-let-me-in. Will resubmit when the repo gets more stuff.

PR #43724, opened 2015-09-09
don't set PYTHONPATH in PythonRequirement env

tdsmith:
Alex noticed that setting PYTHONPATH causes weirdness if we depend_on something which
may be optionally built --with-python3; PYTHONPATH unexpectedly contains
python3 modules in the depending formula if the formula upon which it
depends was built --with-python3 even though the depending formula may
only use python2.

The consequence of this is if something needs to use a Python module
installed by a Homebrew formula, the depending formula will need to
configure PYTHONPATH manually.

### DIFF
--- a/Library/Homebrew/requirements/python_requirement.rb
+++ b/Library/Homebrew/requirements/python_requirement.rb
@@ -23,8 +23,6 @@ class PythonRequirement < Requirement
     elsif short_version != Version.new("2.7")
       ENV.prepend_path "PATH", Formula["python"].opt_bin
     end
-
-    ENV["PYTHONPATH"] = "#{HOMEBREW_PREFIX}/lib/python#{short_version}/site-packages"
   end
 
   def python_short_version
@@ -60,9 +58,7 @@ class Python3Requirement < PythonRequirement
 
   satisfy(:build_env => false) { which_python }
 
-  env do
-    ENV["PYTHONPATH"] = "#{HOMEBREW_PREFIX}/lib/python#{python_short_version}/site-packages"
-  end
+  env {}
 
   def python_binary
     "python3"


Library/Homebrew/requirements/python_requirement.rb
```diff
@@ -60,9 +58,7 @@ class Python3Requirement < PythonRequirement
 
   satisfy(:build_env => false) { which_python }
 
-  env do
-    ENV["PYTHONPATH"] = "#{HOMEBREW_PREFIX}/lib/python#{python_short_version}/site-packages"
-  end
+  env {}
```
xu-cheng: You can just remove this line. `env` DSL wouldn't be inherited.

mikemcquaid:
> The consequence of this is if something needs to use a Python module
> installed by a Homebrew formula, the depending formula will need to
> configure PYTHONPATH manually.

This is a pretty big regression; all formula relying on that behaviour of `depends_on :python` are going to break. I think we need to work out another solution here.

mikemcquaid:
Alternatively, `ENV["PYTHONPATH"] =nil` in the Python 3 requirement.

xu-cheng:
Maybe we can do something like this:
```ruby
env do
  if ENV["PYTHONPATH"].nil? || ENV["PYTHONPATH"] == python_path
    ENV["PYTHONPATH"] = python_path
  else
    # we already set PYTHONPATH somewhere else(it usually means formula requires python 2 and python 3)
    # delete the env variable to avoid conflict.
    ENV.delete "PYTHONPATH"
  end
end

def python_path
  "#{HOMEBREW_PREFIX}/lib/python#{python_short_version}/site-packages"
end
```

tdsmith:
Conditionally clearing PYTHONPATH bothers me; never setting it is the only really consistent behavior.

Reverting #43453 and fixing treeline is a smaller change.

mikemcquaid:
@tdsmith Fine with me :+1:

PR #43725, opened 2015-09-09
Upgrade to nodejs-v4.0.0

hmalphettes:
Cheers!

### DIFF
--- a/Library/Formula/node.rb
+++ b/Library/Formula/node.rb
@@ -22,7 +22,6 @@ class Node < Formula
 
   depends_on :python => :build if MacOS.version <= :snow_leopard
   depends_on "pkg-config" => :build
-  depends_on "icu4c" => :optional
 
   # https://github.com/nodejs/node-v0.x-archive/issues/7919
   # https://github.com/Homebrew/homebrew/issues/36681
--- a/Library/Formula/node.rb
+++ b/Library/Formula/node.rb
@@ -2,9 +2,9 @@
 class Node < Formula
   desc "Platform built on Chrome's JavaScript runtime to build network applications"
   homepage "https://nodejs.org/"
-  url "https://nodejs.org/dist/v0.12.7/node-v0.12.7.tar.gz"
-  sha256 "b23d64df051c9c969b0c583f802d5d71de342e53067127a5061415be7e12f39d"
-  head "https://github.com/nodejs/node.git", :branch => "v0.12"
+  url "https://nodejs.org/dist/v4.0.0/node-v4.0.0.tar.gz"
+  sha256 "e110e5a066f3a6fe565ede7dd66f3727384b9b5c5fbf46f8db723d726e2f5900"
+  head "https://github.com/nodejs/node.git", :branch => "v4.x"
   revision 1
 
   bottle do
@@ -14,6 +14,7 @@ class Node < Formula
   end
 
   option "with-debug", "Build with debugger hooks"
+  option "with-icu4c", "Build with Intl (icu4c) support"
   option "without-npm", "npm will not be installed"
   option "without-completion", "npm bash completion will not be installed"
 
@@ -21,7 +22,7 @@ class Node < Formula
 
   depends_on :python => :build if MacOS.version <= :snow_leopard
   depends_on "pkg-config" => :build
-  depends_on "openssl" => :optional
+  depends_on "icu4c" => :optional
 
   # https://github.com/nodejs/node-v0.x-archive/issues/7919
   # https://github.com/Homebrew/homebrew/issues/36681
@@ -40,12 +41,6 @@ class Node < Formula
     args = %W[--prefix=#{prefix} --without-npm]
     args << "--debug" if build.with? "debug"
     args << "--with-intl=system-icu" if build.with? "icu4c"
-
-    if build.with? "openssl"
-      args << "--shared-openssl"
-    else
-      args << "--without-ssl2" << "--without-ssl3"
-    end
 
     system "./configure", *args
     system "make", "install"


asparagui:
See also #43714.

kwo:
+1

Library/Formula/node.rb
```diff
@@ -14,6 +14,7 @@ class Node < Formula
   end
 
   option "with-debug", "Build with debugger hooks"
+  option "with-icu4c", "Build with Intl (icu4c) support"
```
geoff-codes: Why has this been added as a separate option?

Library/Formula/node.rb
```diff
@@ -21,7 +22,7 @@ class Node < Formula
 
   depends_on :python => :build if MacOS.version <= :snow_leopard
   depends_on "pkg-config" => :build
-  depends_on "openssl" => :optional
+  depends_on "icu4c" => :optional
```
geoff-codes: And why do we have this line now, twice? (lines 25, 29)

Library/Formula/node.rb
```diff
@@ -21,7 +22,7 @@ class Node < Formula
 
   depends_on :python => :build if MacOS.version <= :snow_leopard
   depends_on "pkg-config" => :build
-  depends_on "openssl" => :optional
```
geoff-codes: Why the removal of this option? My understanding is that node (by default) statically links its own version of OpenSSL, complete with a compiled-in CA list.

geoff-codes:
I appreciate that node has moved to semver, and a lot of work has been put in, but maybe lets try to get this formula right before committing ...?

Library/Formula/node.rb
```diff
@@ -2,9 +2,9 @@
 class Node < Formula
   desc "Platform built on Chrome's JavaScript runtime to build network applications"
   homepage "https://nodejs.org/"
-  url "https://nodejs.org/dist/v0.12.7/node-v0.12.7.tar.gz"
-  sha256 "b23d64df051c9c969b0c583f802d5d71de342e53067127a5061415be7e12f39d"
-  head "https://github.com/nodejs/node.git", :branch => "v0.12"
+  url "https://nodejs.org/dist/v4.0.0/node-v4.0.0.tar.gz"
+  sha256 "e110e5a066f3a6fe565ede7dd66f3727384b9b5c5fbf46f8db723d726e2f5900"
+  head "https://github.com/nodejs/node.git", :branch => "v4.x"
```
geoff-codes: The [4.0.0 release notes](https://nodejs.org/en/blog/release/v4.0.0/) seem to indicate that future development will be on a (yet to be created) v5.x branch, and that the v4.x branch will be sort of LTS/"patches only", in line with their move to semantic versioning.

geoff-codes: The [4.0.0 release notes](https://nodejs.org/en/blog/release/v4.0.0/) seem to indicate that the 'v4.x' branch will now be "LTS"/"patches only" and development will de done on a (yet to be created) 'v5.x' branch, in line with the move to semantic versioning.

mikemcquaid:
@geoff-codes I appreciate that you deleted your comment but everyone watching this thread still saw it. Please read our Code of Conduct: https://github.com/Homebrew/homebrew/blob/master/CODEOFCONDUCT.md

geoff-codes:
@mikemcquaid All I can really say is that,
1. I am only human, as are those whom I criticized
2.  I removed it as soon as I realized what I said (within minutes), but, alas, the internet has no true 'delete' key
3. I apologize

mikemcquaid:
@geoff-codes That's :ok:, we appreciate the apology, thanks.

geoff-codes:
Thanks. Having had a proper good refresher on said [Code of Conduct](https://github.com/Homebrew/homebrew/blob/master/CODEOFCONDUCT.md), I quite appreciate your criticism as laid out therein:
> Our community is based on mutual respect, tolerance, and encouragement, and **we are working to help each other live up to these principles.**

I think If everyone held each other to account as well as you have with me just now, I'd reckon the internet would be a much more civilized place. :+1: :wink: :beers:

hmalphettes:
@asparagui I apologize for the duplicate effort: somehow I missed #43714 in the list of PRs.
@geoff-codes thanks for the review.

Library/Formula/node.rb
hmalphettes: It reflects this choice when configuring node:
https://github.com/nodejs/node#intl-ecma-402-support

I ported what we were doing for iojs here: https://github.com/aredridel/homebrew-iojs/blob/master/Formula/iojs.rb

Library/Formula/node.rb
hmalphettes: Thanks for noting that one. Removing the duplicate.

Library/Formula/node.rb
hmalphettes: I tried leaving it in place and it complained that `--without-ssl2` is not a supported parameter.

Library/Formula/node.rb
geoff-codes: > It reflects this choice when configuring node:
https://github.com/nodejs/node#intl-ecma-402-support

I see. I like that there are a few different options now.

> I ported what we were doing for iojs here: https://github.com/aredridel/homebrew-iojs/blob/master/Formula/iojs.rb

So if I understanding correctly, the "default" is becoming ` --with-intl=system-icu`, (a decision which I heartily concur with; I think its the best option), and the homebrew install flag '`-with-icu4c` is to force the build to link with homebrew's 'icu4c' package? May I suggest then that the option be changed to `--with-brewed-icu4c`, a la the `git.rb` formula's options for openssl, ssl, and curl?

hmalphettes:
Reviewing the earlier PR by @killswitch my feeling is that the discussions and reviews are more advanced over there.
I'm closing this one and will focus on #43714

Library/Formula/node.rb
geoff-codes: So (again, please pardon any ignorance) but this leaves me inclined to believe that (most likely) `--without-ssl2` is now default (i.e., not an option any more, when configuring node). So, I don't see why the _homebrew_ option needs to be changed; simply stop passing this argument to the configure script?

geoff-codes: Although, again, I personally would advocate that this option shouldn't really be necessary, since the default configuration is perfectly fine. If it _is_ left as an option, `--with-brewed-openssl` would be clearer.

Library/Formula/node.rb
geoff-codes: Or simply remove this as an option altogether, since I don't see much reason why one should wan't to **not** link to the system lib, as it's perfectly adequate?

DomT4:
@hmalphettes Thanks for closing this one up, and submitting the PR. Do appreciate the submission.

I'm going to lock this thread as keeping tabs on two active discussion threads is enough for everyone, let alone three, but welcome politely-toned discussion in the other threads.

PR #43726, opened 2015-09-09
Description related improvement

xu-cheng:
cc @hivehand

### DIFF
--- a/Library/Homebrew/descriptions.rb
+++ b/Library/Homebrew/descriptions.rb
@@ -42,20 +42,26 @@ class Descriptions
   # Return true if the cache exists, and neither Homebrew nor any of the Taps
   # repos were updated more recently than it was.
   def self.cache_fresh?
-    if CACHE_FILE.exist?
-      cache_date = File.mtime(CACHE_FILE)
+    return false unless CACHE_FILE.exist?
+    cache_mtime = File.mtime(CACHE_FILE)
+    ref_master = ".git/refs/heads/master"
 
-      ref_master = ".git/refs/heads/master"
-      master = HOMEBREW_REPOSITORY/ref_master
+    master = HOMEBREW_REPOSITORY/ref_master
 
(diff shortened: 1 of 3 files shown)


Library/Homebrew/descriptions.rb
```diff
@@ -1,8 +1,8 @@
 require "formula"
-require "csv"
+require "formula_versions"
 
 class Descriptions
-  CACHE_FILE = HOMEBREW_CACHE + "desc_cache"
+  CACHE_FILE = HOMEBREW_CACHE + "desc_cache.json"
```
mikemcquaid: Will this destroy the existing cache? If so, maybe migrate it?

mikemcquaid:
One nit otherwise :+1:

Library/Homebrew/descriptions.rb
xu-cheng: For many reasons, (to prevent multiple instance of brew writing cache file, make sure we finish the writing, etc.) We need to use `atomic_write` when saving the cache. And due to lack of `dump` in csv module, I choose to use json. Therefore I have to change the cache filename as  the format isn't compatible. 

>  If so, maybe migrate it?

I don't think we need this as the cost to regenerate cache is considerable small.

mikemcquaid: Ok, if the format has changed: :cool:, skipped over that below.

hivehand:
Thanks for the cc! These changes look eminently reasonable, when not downright elegant, to me.

One random request: I realized shortly after the original merge that `Descriptions::generate_cache` currently uses `Formula.map` where it really should just be using `Formula.each`. If you wouldn’t mind folding that change into this pull request, I’d be grateful. (If folks would rather I submit that as a separate pull request, I’ll do so; it struck me as potential overkill for a one-line change.)

xu-cheng:
@hivehand Thanks for comment. Added the `map -> each` change.

Issue #43568, opened 2015-09-04
Unable to install JRuby on OSX- Homebrew cannot find proper Java version


elifoster:
Running Mac OS X 10.9.5, I am unable to install JRuby via Homebrew with Java 6, 7, or 8 installed.
```
$ java -version
java version "1.8.0_31"
Java(TM) SE Runtime Environment (build 1.8.0_31-b15)
Java HotSpot(TM) 64-Bit Server VM (build 25.31-b07, mixed mode)

$ brew install jruby
jruby: Java 1.7+ is required to install this formula.
You can install with Homebrew Cask:
  brew install Caskroom/cask/java

You can download from:
  http://www.oracle.com/technetwork/java/javase/downloads/index.html
Error: An unsatisfied requirement failed this build.
```

```
$ java -version
openjdk version "1.7.0-internal"
OpenJDK Runtime Environment (build 1.7.0-internal-root_2014_04_03_15_15-b00)
OpenJDK 64-Bit Server VM (build 24.0-b26, mixed mode)

$ brew install jruby
jruby: Java 1.7+ is required to install this formula.
You can install with Homebrew Cask:
  brew install Caskroom/cask/java

You can download from:
  http://www.oracle.com/technetwork/java/javase/downloads/index.html
Error: An unsatisfied requirement failed this build.
```

As you can see, it cannot find the Java installations that it should be trying to find. I assume it is instead looking at Apple's version of Java, which it should not be using unless there are no other alternatives. It should actually just be using Terminal's version of Java (/usr/bin/java typically).

tzudot:
Can you supply your `brew config`, thanks.

elifoster:
```
$ brew config
HOMEBREW_VERSION: 0.9.5
ORIGIN: https://github.com/Homebrew/homebrew
HEAD: a2373af9c153d68511778b446387024e1f3caa8f
Last commit: 4 weeks ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: dual-core 64-bit penryn
OS X: 10.9.5-x86_64
Xcode: 6.2
CLT: 6.2.0.0.1.1424975374
GCC-4.2: build 5666
Clang: 6.0 build 600
MacPorts/Fink: /sw/bin/fink
X11: 2.7.7 => /opt/X11
System Ruby: 2.0.0-p481
[...]

PR #43730, opened 2015-09-09
config: show all installed JDK

xu-cheng:
Per https://github.com/Homebrew/homebrew/issues/43568#issuecomment-138808054

### DIFF
--- a/Library/Homebrew/cmd/config.rb
+++ b/Library/Homebrew/cmd/config.rb
@@ -1,5 +1,6 @@
 require "hardware"
 require "software_spec"
+require "rexml/document"
 
 module Homebrew
   def config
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

mikemcquaid:
Although it'd be good to ensure the first in the PATH is the first displayed maybe?

xu-cheng:
The question is do we want to show all JDK. (I'd like to considering there are formulae requiring specific java version)

Otherwise we can do `/usr/libexec/java_home --exec java -version` to get JDK version.

mikemcquaid:
Showing all versions sounds good :+1:

xu-cheng:
> Although it'd be good to ensure the first in the PATH is the first displayed maybe?

It seems it is this case already. It's ordered by version which will be the same order to determine java_home. c.f. http://stackoverflow.com/a/20994919

mikemcquaid:
:+1: :shipit:

PR #43744, opened 2015-09-09
Don't set PYTHONPATH for :python3

tdsmith:
Reduces the impact of #43724 by not changing anything in the python2 case.

### DIFF
--- a/Library/Homebrew/requirements/python_requirement.rb
+++ b/Library/Homebrew/requirements/python_requirement.rb
@@ -24,9 +24,7 @@ class PythonRequirement < Requirement
       ENV.prepend_path "PATH", Formula["python"].opt_bin
     end
 
-    if python_binary == "python"
-      ENV["PYTHONPATH"] = "#{HOMEBREW_PREFIX}/lib/python#{short_version}/site-packages"
-    end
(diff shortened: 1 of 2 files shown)


Library/Homebrew/requirements/python_requirement.rb
```diff
@@ -24,7 +24,9 @@ class PythonRequirement < Requirement
       ENV.prepend_path "PATH", Formula["python"].opt_bin
     end
 
-    ENV["PYTHONPATH"] = "#{HOMEBREW_PREFIX}/lib/python#{short_version}/site-packages"
+    if python_binary == "python"
```
xu-cheng: I think this will always be true.

mikemcquaid:
:+1: CC @dunn

Library/Homebrew/requirements/python_requirement.rb
mikemcquaid: The `env` isn't inherited, right? That's probably the cause of the confusion.

xu-cheng: Right, all of DSL wouldn't be able to be inherited by subclass.

mikemcquaid: Maybe it's worth refactoring this into a method and calling it from both pythons `env do` block. I believe that will sort all the problems as well as ensuring it's at the front of the `PATH`.

PR #43753, opened 2015-09-09
Vim 7.4.865

nelstrom:
Vim's source code recently migrated from Google Code to GitHub. Github generates [a `.tar.gz` for every patch release](https://github.com/vim/vim/releases). I propose using the github archive of releases instead of [debian-unstable](https://mirrors.kernel.org/debian/pool/main/v/vim/) in this formula.

### DIFF
--- a/Library/Formula/vim.rb
+++ b/Library/Formula/vim.rb
@@ -1,9 +1,8 @@
 class Vim < Formula
   desc "Vi \"workalike\" with many additional features"
   homepage "http://www.vim.org/"
-  # This package tracks debian-unstable: https://packages.debian.org/unstable/vim
-  url "https://mirrors.kernel.org/debian/pool/main/v/vim/vim_7.4.826.orig.tar.gz"
-  sha256 "02f07b60eff53f45d58686e43b72e83aa8f24a94acfa69b95fa84dc020671a38"
+  url "https://github.com/vim/vim/archive/v7.4.861.tar.gz"
+  sha256 "fc24e318073db486bf99204eca8cdfd71174810c327c42cc20aad941858d1636"
   head "https://github.com/vim/vim.git"
 
   # We only have special support for finding depends_on :python, but not yet for
--- a/Library/Formula/vim.rb
+++ b/Library/Formula/vim.rb
@@ -1,8 +1,8 @@
 class Vim < Formula
   desc "Vi \"workalike\" with many additional features"
   homepage "http://www.vim.org/"
-  url "https://github.com/vim/vim/archive/v7.4.861.tar.gz"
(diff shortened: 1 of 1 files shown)


guoxiao:
see #43286

nelstrom:
@guoxiao Thanks for pointing me towards that discussion.

nelstrom:
As far as I'm concerned, [patch 7.4.858](https://github.com/vim/vim/commit/aa23b379421aa214e6543b06c974594a25799b09) is a big deal. I've been eager for this functionality to be merged into core Vim since [the patch first appeared](https://groups.google.com/forum/#!topic/vim_dev/dfyt-G6SMec) in February 2013.

I would love to get this new functionality out to more Vim users and I thought I'd start by updating the homebrew Vim formula. It would be better still if we could get this update out to all debian users! How would I go about suggesting to the maintainers of debian unstable that this would be a good time to create a new `.tar.gz`?

mikemcquaid:
@nelstrom I'm OK with it being updated once a week given we wait to make sure it's not gonna get immediately updated the day after.

DomT4:
I'm fine with weekly updates. That experience will get nicer once bottles are sorted here. We'll need a note in the formula to explain the new release schedule.

nelstrom:
@mikemcquaid cool. And what do you know, this Formula is already out of date now that [Vim 7.4.865 is out](https://github.com/vim/vim/releases/tag/v7.4.865)! If we're going to go for a weekly release, perhaps Monday is a good day to do it. From the [punch card](https://github.com/vim/vim/graphs/punch-card), it looks to be one of the less active days for Vim development.

nelstrom:
@DomT4 where can I read more about bottles? Homebrew's terminology makes me dizzy :dizzy_face:

bfontaine:
@nelstrom see https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Bottles.md

DomT4:
Bottle = CI-produced binary package, if that's easier to understand.

nelstrom:
@DomT4 so when you say:

> That experience will get nicer once bottles are sorted here

You're referring to the experience for people running `brew install vim`, who will be able to install the binary built by the Brew test bot, rather than building from source. Is that right?

DomT4:
> Is that right?

Yes. It won't be quite as unconditional as that due to the fun fun way Vim builds, but at least the option for people who meet the preconditions (still being ironed out) will be available.

Vim isn't a huge compile by any means, but there's time, download speed & CPU usage advantages to pouring bottles instead of building from source every single update.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/c82c7058c3aaf80b9610bb137e867a58fae8e271. Thank you for your contribution to Homebrew @nelstrom; we appreciate it! :smiley_cat:

PR #43755, opened 2015-09-09
saltstack 2015.5.5 (updated to Jinja 2.8)

dneades:
This PR updates the existing saltstack 2015.5.5 formula to use Jinja 2.8, rather than the elderly Jinja 2.7.3. (The equivalent saltstack package on FreeBSD already uses Jinja 2.8, so this brings the OS X Homebrew formula into line with that, thus making formulas using Jinja compatible between the platforms.)

Jinja 2.8 introduces useful additional features such as the [`equalto` test](http://jinja.pocoo.org/docs/dev/templates/#equalto).


bfontaine:
Please add a `revision 1` so the formula will be rebuilt.

dneades:
@bfontaine, done, please try that…

bfontaine:
Thanks a lot @dneades! :tada:

PR #43757, opened 2015-09-09
ldc 0.16.0-alpha2

tomyun:
It's not a stable release yet, but maybe more stable than 0.15.2-beta-2? Or we may want this back to be `devel` (#41405).

### DIFF
--- a/Library/Formula/ldc.rb
+++ b/Library/Formula/ldc.rb
@@ -13,6 +13,12 @@ class Ldc < Formula
     sha256 "8ea1ad6cbd93ca1b909fdce9ff136fe56ada447cd535840f2684088dcab0fea2" => :mountain_lion
   end
 
+  devel do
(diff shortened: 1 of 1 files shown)


DomT4:
Any thoughts here @John-Colvin?

John-Colvin:
I think 0.15.2-beta-2 should be stable (followed by any later betas, then 0.15.2 when released), 0.16.0-alpha-2 should be devel.
It's bad enough having a beta as stable, but an alpha would be even worse.

tomyun:
Okay, it's now changed to `devel`.

DomT4:
Thanks for the review here @John-Colvin.

Merged in https://github.com/Homebrew/homebrew/commit/2eafe73ffa6767121c66882719c64eca7264f68a. Thank you for your contribution to Homebrew @tomyun; we appreciate it! :smiley_cat:

PR #43760, opened 2015-09-10
scamper: use https urls

vszakats:


bfontaine:
Thank you @vszakats! :smile:

PR #43761, opened 2015-09-10
Updated libbtbb to 2015-09-R2

bkonosky:
Updated libbtbb to 2015-09-R2 to keep things up to date.

### DIFF
--- a/Library/Formula/libbtbb.rb
+++ b/Library/Formula/libbtbb.rb
@@ -11,6 +11,7 @@ class Libbtbb < Formula
 
   depends_on "cmake" => :build
   depends_on "python"
+  depends_on "libpcap"
 
   def install
     args = std_cmake_args
--- a/Library/Formula/libbtbb.rb
+++ b/Library/Formula/libbtbb.rb
@@ -1,18 +1,11 @@
 class Libbtbb < Formula
   desc "Bluetooth baseband decoding library"
   homepage "https://github.com/greatscottgadgets/libbtbb"
-  url "https://github.com/greatscottgadgets/libbtbb/archive/2014-02-R4.tar.gz"
-  sha256 "6e79a2d8530596e34ad13fcce43dcf8b30b165d4629d5bf9c3eb7f8817980524"
-  version "2014-02-R4"
+  url "https://github.com/greatscottgadgets/libbtbb/archive/2015-09-R2.tar.gz"
+  sha256 "35ce44636649163f0d9a4de5905f686470f54c79b408959760db8c8182853161"
+  version "2015-09-R2"
 
   head "https://github.com/greatscottgadgets/libbtbb.git"
-
-  bottle do
-    cellar :any
-    sha1 "e1870ab678e92b9524d3f1ea05862fedc3d196eb" => :yosemite
-    sha1 "52071cec084c13a10a08102e391f457d176ecd1d" => :mavericks
-    sha1 "3069f205340202284cd8e0ec902d056c4f851bba" => :mountain_lion
-  end
 
   option :universal
 
--- a/Library/Formula/libbtbb.rb
+++ b/Library/Formula/libbtbb.rb
@@ -11,7 +11,15 @@ class Libbtbb < Formula
 
   depends_on "cmake" => :build
   depends_on "python"
-  depends_on "libpcap"
+  depends_on "homebrew/dupes/libpcap"
+
+
+  bottle do
+    cellar :any
+    sha1 "e1870ab678e92b9524d3f1ea05862fedc3d196eb" => :yosemite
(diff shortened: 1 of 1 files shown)


bkonosky:
Didn't notice the libpcap requirement that was missing.

Library/Formula/libbtbb.rb
```diff
@@ -1,23 +1,17 @@
...
-    sha1 "e1870ab678e92b9524d3f1ea05862fedc3d196eb" => :yosemite
-    sha1 "52071cec084c13a10a08102e391f457d176ecd1d" => :mavericks
-    sha1 "3069f205340202284cd8e0ec902d056c4f851bba" => :mountain_lion
-  end
-
   option :universal
 
   depends_on "cmake" => :build
   depends_on "python"
+  depends_on "libpcap"
```
bfontaine: You have to use `homebrew/dupes/libpcap` here.

mikemcquaid: And this needs to be an optional dependency.

Library/Formula/libbtbb.rb
```diff
@@ -1,23 +1,17 @@
...
-  url "https://github.com/greatscottgadgets/libbtbb/archive/2014-02-R4.tar.gz"
-  sha256 "6e79a2d8530596e34ad13fcce43dcf8b30b165d4629d5bf9c3eb7f8817980524"
-  version "2014-02-R4"
+  url "https://github.com/greatscottgadgets/libbtbb/archive/2015-09-R2.tar.gz"
+  sha256 "35ce44636649163f0d9a4de5905f686470f54c79b408959760db8c8182853161"
+  version "2015-09-R2"
 
   head "https://github.com/greatscottgadgets/libbtbb.git"
 
-  bottle do
```
DomT4: Please don't kill these bottle blocks on update. We pull a new bottle set regardless of whether one exists or not.

bkonosky: Thanks, I'm putting it back in.

Library/Formula/libbtbb.rb
bkonosky: bfontaine: Thank you, this is my first ever contribution to homebrew, so all the help on getting it right is welcome.

mikemcquaid: It's not optional, the latest version of libbtbb requires libpcap now, unless I'm missing a switch somewhere.

DomT4: OS X ships `libpcap` natively, It shouldn't require ours directly:

```
/usr/bin/pcap-config
/usr/include/pcap.h
/usr/include/pcap-namedb.h
/usr/include/pcap-bpf.h
/usr/include/pcap/usb.h
/usr/include/pcap/sll.h
/usr/include/pcap/pcap.h
/usr/include/pcap/pcap-util.h
/usr/include/pcap/namedb.h
/usr/include/pcap/bpf.h
/usr/lib/libpcap.A.dylib
```

bkonosky:
Thank you for the help, it looks like it's building properly now. Next up formula for Ubertooth.

DomT4:
You need to try and use the system `libpcap` here. We don't permit mandatory cross-tap dependencies.

bkonosky:
It won't compile without the Homebrew version of libpcap installed, the OSX version doesn't have the Bluetooth bits that the new version of libbtbb needs in it. So should libbtbb be removed from the main branch and put into a separate tap? If so I'll be willing to create a tap with both libbtbb and ubertooth in it away from the main branch.

mikemcquaid:
Another option is just to vendor a `libpcap` in here and statically link against it?

DomT4:
Yeah. Dump it in `libexec/vendor` and then point this towards it, IMO. You'd either need to feed install a direct path or do something like `ENV.prepend_path "PATH", libexec/"vendor/bin"`.

bkonosky:
I've been trying to figure out how to do the vendor parts, and still don't know how to do it that way. I thought my Google-fu was strong, but it's weak in this area. How about I just close this pull request and make a tap instead, since libbtbb and ubertooth aren't really core items anyways?

DomT4:
```ruby
depends_on "cmake" => :build
depends_on "whatever"

resource "kittens" do
  url "https://kittens.com/kittens/kittens-1.0.4.tar.gz"
  sha256 "fc615c623bfc65b957b51902918de7d5ab70d48c690c26a4c6ecd96f30d3311f"
end

def install
  resource("kittens").stage do
    system "./configure", "--prefix=#{libexec}/vendor", "--other-option",
                          "--and-another"
    system "make", "install"
  end

  ENV.prepend_path "PATH", libexec/"vendor/bin"
  # continue libbtbb install normally here
end
```

bkonosky:
Thank you DonT4. I'm working on implementing that now.

PR #43766, opened 2015-09-10
bottle: allow to keep old bottle specification

xu-cheng:
cc @mikemcquaid


xu-cheng:
I just found I forget to make revision etc to sync. Will have a try to find a solution.

xu-cheng:
Added field checking when keeping old bottle. And set revision as the same in that situation.

mikemcquaid:
LGTM :+1:

xu-cheng:
@BrewTestBot test this please

PR #43782, opened 2015-09-10
Gpgme and dependencies

jeroenooms:
- Update libgpgme to the latest version. 
 - Enable static libraries for the dependencies of libgpgme, so that we can actually use the libgpgme static library.

### DIFF
--- a/Library/Formula/libassuan.rb
+++ b/Library/Formula/libassuan.rb
@@ -18,7 +18,8 @@ class Libassuan < Formula
   def install
     system "./configure", "--disable-dependency-tracking",
                           "--disable-silent-rules",
-                          "--prefix=#{prefix}"
+                          "--prefix=#{prefix}",
+                          "--enable-static"
     system "make", "install"
   end
 
--- a/Library/Formula/libassuan.rb
+++ b/Library/Formula/libassuan.rb
@@ -5,6 +5,7 @@ class Libassuan < Formula
   mirror "ftp://ftp.gnupg.org/gcrypt/libassuan/libassuan-2.3.0.tar.bz2"
   mirror "https://www.mirrorservice.org/sites/ftp.gnupg.org/gcrypt/libassuan/libassuan-2.3.0.tar.bz2"
   sha256 "87c999f572047fa22a79ab5de4c8a1a5a91f292561b69573965cac7751320452"
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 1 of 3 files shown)


Library/Formula/libassuan.rb
```diff
@@ -18,7 +18,8 @@ class Libassuan < Formula
   def install
     system "./configure", "--disable-dependency-tracking",
                           "--disable-silent-rules",
-                          "--prefix=#{prefix}"
+                          "--prefix=#{prefix}",
+                          "--enable-static"
```
dunn: This probably warrants a `revision 1` cc @DomT4

DomT4:
Heh, I hadn't forgotten about this @jeroenooms :smiley:; Just caught up lately on finishing up my current Uni module. The other reason I hadn't moved on this just yet is because the recent GPGME build kept failing to build (See #43307). Hopefully the error I saw there doesn't keep reproducing.

I've no objection here.
My gut feeling was to make it optional, but providing everything works, `¯\_(ツ)_/¯`.

Library/Formula/libassuan.rb
DomT4: I think it'd be okay, but all the bottles are `cellar :any` so we can play it better safe than sorry.

jeroenooms:
It would be really nice if `--enable-static` is the default config so that the static libraries are included with the bottles, as they are for libgpgme. The same is done for most linux distributions.

DomT4:
I don't really object enough to making it the default to stop it being so. If people see problems we may have to look at it again, but there are things we build both for so it's not unprecedented. Bump the revisions on `libassuan` and `libgpg-error` and I'll merge this. It'd be nice to use this download link for GPGME as well:

```
https://gnupg.org/ftp/gcrypt/gpgme/gpgme-1.6.0.tar.bz2
```

jeroenooms:
OK added the mirror URL. Where do I bump the revision? I see no such field in the recipe?

DomT4:
It would be under the `sha256` line if it already existed. The default revision is `0` which we don't write. So in this case, just:

```diff
 url "blah"
 mirror "blah"
 sha256 "blah"
+revision 1
```

jeroenooms:
ok done, thank you!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/c2c71a24c1869e65d67fea997e3205036e7733df, https://github.com/Homebrew/homebrew/commit/61ddb9fcc5fae7bd3ee82831b5385ea737802d4a & https://github.com/Homebrew/homebrew/commit/a6a93d08a6f810fdcd8d531ffca60a724fb9a33c. Thank you for your contribution to Homebrew @jeroenooms!

PR #43789, opened 2015-09-10
Improve launchdns linking example

josh:
With the old instructions, if /etc/resolver link does not exist the link will be created as expected. However, if the link or directory already exists this will result in a nested directory like /etc/resolver/resolver.

-
To: @mikemcquaid


mikemcquaid:
Thanks @josh!

Issue #40837, opened 2015-06-17
10.11 [Tracking Issue]


DomT4:
This is essentially a triage of various issues and progress on them. Did the same thing for 10.10 last year and it worked relatively well at keeping everyone in the loop.

This thread will be comment-policed fairly hard; Please keep comments related to failures/issues/etc. If you can reproduce an issue providing a `brew gist-logs FORMULA` is preferred over a simple :+1: - Thanks!

**New issues should still be filed as new issues. This is only a centralised tracking thread rather than a solve-everything-here thread.**

Core Fixes:
* Basic recognition of 10.11 was added to Homebrew - https://github.com/Homebrew/homebrew/commit/62381a65d1aa98e00d3fc2889363d8d44fdf6b58
* `pkg-config` files were added for 10.11 - https://github.com/Homebrew/homebrew/commit/4a8fcd199df144b1d3d209c21f9443d369f4bf5b
* Better detection and warning of unsupported OS X versions was committed - https://github.com/Homebrew/homebrew/commit/baaa295515ce877557dcd69cc89c7e441a6185b2.
* Clang version update for 10.11 DP2 added in https://github.com/Homebrew/homebrew/commit/11cf300c01e3c43fc9604b87072ae12d4208a410.
* Clang version recognition fixed in https://github.com/Homebrew/homebrew/commit/0599646af6e6cf775081afb4fcb9af99d6b1088c.
* Clang version updated for 10.11 DP4 in https://github.com/Homebrew/homebrew/commit/df15afd28b9d7b7778c4f0d84dd8317e5369bacd.

Potential Issues:
* Rootless: Bans writing to [/System, /usr, /bin, /sbin](http://arstechnica.com/apple/2015/06/preview-os-x-el-capitans-first-beta-is-a-promising-heap-of-refinements/4/) - So far `/usr/local` isn't write-protected.
* Permissions can wander off and prevent README overwriting - #41150.
* System OpenSSL: Headers have seemingly been removed from the CLT and Xcode - https://github.com/Homebrew/homebrew/issues/40519#issuecomment-110590553.

Formulae Issues Outstanding:
* Pixman fails to build - #41056 - Can't find somewhere reliable to report upstream.
* Python can fail to build core extension modules - #41085.
* Google-sparsehash fails to build on DP2 - #41087 - [Upstream report](https://code.google.com/p/sparsehash/issues/detail?id=111).
* Fltk fails to build on DP3 - #41581.
[...]

PR #43802, opened 2015-09-10
zookeeper: Apply htonll patch to El Capitan builds as well

wfarr:
This fixes the zookeeper build on 10.11

I've tested this on the El Cap GM.

ref: https://github.com/Homebrew/homebrew/issues/40837


DomT4:
Our CI is snowed at the moment, so rather than making you sit on a fix here for another 4 hours I've done it in https://github.com/Homebrew/homebrew/commit/e00142fde31c4873c691e8d22881e68e91e5d343. Thanks for the submission and testing this! :heart:

PR #43787, opened 2015-09-10
python3 3.5.0rc4 (devel)

zoidbergwill:
rc4 came out on the 9th of September. [Changelog](https://docs.python.org/3.5/whatsnew/changelog.html#python-3-5-0-release-candidate-4)

PR #43804, opened 2015-09-10
Python3: 3.5.0rc4 for --devel

tehmaze:
Release candidate 4 was released https://www.python.org/downloads/release/python-350rc4/


tdsmith:
Duplicate of #43787; thanks.

PR #43811, opened 2015-09-10
Alter patch to make http-parser bottle relocatable

staticfloat:
I'm not sure if you guys would prefer this patch rolled into the current one (as I have it right now) or as a separate one on top of the current patch.  The only reason I rolled it into the current one is that it touches some of the same lines.

This patch changes the `ln` command to create a relative symlink rather than an absolute one.

### DIFF
--- a/Library/Formula/http-parser.rb
+++ b/Library/Formula/http-parser.rb
@@ -13,8 +13,8 @@ class HttpParser < Formula
   depends_on "coreutils" => :build
 
   patch do
-    url "https://patch-diff.githubusercontent.com/raw/joyent/http-parser/pull/247.diff"
-    sha256 "0f081d0cbf44575b2db2889b318245e692706b756725ba64d1277f571c8b921f"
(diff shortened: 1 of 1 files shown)


staticfloat:
I have a PR open to upstream about this here: https://github.com/joyent/http-parser/pull/264

xu-cheng:
I'm thinking instead patching individual formulae we should handle this inside relocate logic.

staticfloat:
That is also a good idea; I'm working with upstream to get this patched, but relocation logic to remap symlinks would be nice as well!

staticfloat:
Great; so now that #43816 has been merged, can `http-parser` be changed to simply be `Cellar :any`, or does it need to be rebottled?

xu-cheng:
It requires rebottle. I have queue a job here http://bot.brew.sh/job/Homebrew%20Testing/880/

xu-cheng:
Fixed by https://github.com/Homebrew/homebrew/commit/01a0d4b5555298c2bf61abd48f061f0cf1dda38d

staticfloat:
Thanks @xu-cheng!  I appreciate it!

PR #43813, opened 2015-09-11
update git-lfs to v0.6.0

technoweenie:


Library/Formula/git-lfs.rb
```diff
@@ -1,9 +1,9 @@
 class GitLfs < Formula
   desc "Git extension for versioning large files"
   homepage "https://github.com/github/git-lfs"
-  url "https://github.com/github/git-lfs/archive/v0.5.4-homebrew.tar.gz"
-  version "0.5.4"
-  sha256 "cf860d56d86cda40a5b883b81f4d991617788c13d5e09be6b35c8bb9ac878ad7"
+  url "https://github.com/github/git-lfs/archive/v0.6.0.tar.gz"
+  version "0.6.0"
```
DomT4: You can cut this line now. Was only needed because the previous version had the `-homebrew` appended to it.

technoweenie:
Ah, thanks @DomT4 :)

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/78783da2087c26c6ee0990f327de0cb28af4fbd2. Thank you for your contribution to Homebrew @technoweenie!

gurugray:
thanks!

but I don't get it — why it's an error, and not warning with this case :)

PR #43816, opened 2015-09-11
keg_relocate: fix absolute symlink

xu-cheng:
Per https://github.com/Homebrew/homebrew/pull/43811#issuecomment-139458280

cc @staticfloat  @mikemcquaid


mikemcquaid:
:+1:

flipphillips:
I've just started having problems making things relocatable, traced back to keg_relocate https://github.com/Homebrew/homebrew/issues/44030 - any chance this fix might be responsible?

PR #43821, opened 2015-09-11
duplicity: gnupg dependency

bruno-:
This PR "relaxes" the `duplicity` formula dependency on `gnupg` if `gnupg2` is installed.

`duplicity` formula works just fine with `gnupg2`.

### DIFF
--- a/Library/Formula/duplicity.rb
+++ b/Library/Formula/duplicity.rb
@@ -20,7 +20,7 @@ class Duplicity < Formula
 
   depends_on :python if MacOS.version <= :snow_leopard
   depends_on "librsync"
-  depends_on "gnupg"
+  depends_on :gpg => :run
 
   # generated with homebrew-pypi-poet from
   # for i in boto pyrax dropbox mega.py paramiko pycrypto


Library/Formula/duplicity.rb
```diff
@@ -20,7 +20,7 @@ class Duplicity < Formula
 
   depends_on :python if MacOS.version <= :snow_leopard
   depends_on "librsync"
-  depends_on "gnupg"
+  depends_on "gnupg" unless Formula["gnupg2"].any_version_installed?
```
xu-cheng: You can just use `depends_on :gpg`

DomT4:
Do it like this instead:

```ruby
depends_on :gpg
```

It'll do the same thing when you build from source in not installing `gnupg` if you already have `gnupg2` installed locally.

Library/Formula/duplicity.rb
DomT4: Xu beat me to the same comment I made below :smiley_cat:

xu-cheng:
Actually, if gpg is only runtime dependency, we should make it `depends_on :gpg => :run`.

bruno-:
Thank you for the feedback guys, I've made the update to this: `depends_on :gpg => :run`

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/c4b7eeeddfb2c4f447a15f2cf6a8c81639d49cc8. Thank you for your contribution to Homebrew @bruno-!

PR #43823, opened 2015-09-11
git-lfs 0.6.0

gurugray:
New major version with fetch/checkout/pull commands and batch API support.


DomT4:
See #43813. Will be merged once that version issue is fixed there.

PR #43826, opened 2015-09-11
ARGV: cache the results

xu-cheng:


Library/Homebrew/extend/ARGV.rb
```diff
@@ -93,82 +93,100 @@ def value(arg)
   end
 
   def force?
-    flag? "--force"
+    return @force if instance_variable_defined?(:@force)
+    @force = flag? "--force"
```
bfontaine: Why didn’t you use `@force ||= flag? "--force"` here?

bfontaine:
Good idea!

Library/Homebrew/extend/ARGV.rb
xu-cheng: Because that logic cannot cache false result.

bfontaine: That’s right.

mikemcquaid:
Do you have any before/after benchmarks for this? I'm struggling to believe this is a performance bottleneck.

xu-cheng:
Yes it won't make any huge performance improvement. I do this to simplify code writing, so we can just have `ARGV.foo? ... ARGV.foo?` instead of `foo = AGRV.foo? ... <use foo afterwards>`.

mikemcquaid:
I think it makes the code much harder to read. I also don't think you need to cache `ARGV` usage like that given it's just an array lookup.

PR #43828, opened 2015-09-11
ice 3.6.1

externl:
Update to Ice 3.6.1 and fix Java to require at least java 7 (6 does not work.) and Berkeley DB to build Java 7 compatible code.

### DIFF
--- a/Library/Formula/ice.rb
+++ b/Library/Formula/ice.rb
@@ -1,8 +1,9 @@
 class Ice < Formula
   desc "Comprehensive RPC framework"
   homepage "https://zeroc.com"
-  url "https://github.com/zeroc-ice/ice/archive/v3.6.0.tar.gz"
-  sha256 "77933580cdc7fade0ebfce517935819e9eef5fc6b9e3f4143b07404daf54e25e"
+  url "https://github.com/zeroc-ice/ice/archive/v3.6.1-el_capitan.tar.gz"
+  sha256 "4a348ba24daceb7694bc23ee91994e2653c5d869918e44b2b1f0d49a360e93fb"
+  version "3.6.1"
 
   bottle do
     revision 2
@@ -13,7 +14,7 @@ class Ice < Formula
   option "with-java", "Build Ice for Java and the IceGrid Admin app"
 
   depends_on "mcpp"
-  depends_on :java => :optional
+  depends_on :java => [ "1.7+", :optional]
   depends_on :macos => :mavericks
 
   resource "berkeley-db" do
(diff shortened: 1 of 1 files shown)


Library/Formula/ice.rb
```diff
@@ -32,7 +32,10 @@ def install
         --enable-cxx
       ]
 
-      args << "--enable-java" if build.with? "java"
+      if build.with? "java"
+        args << "--enable-java"
+        inreplace "dist/Makefile.in", "@JAVACFLAGS@",  "@JAVACFLAGS@ -source 1.7 -target 1.7"
```
dunn: Should this be submitted as a patch upstream?

externl: No, Oracle is not doing anything with DB 5 anymore.

mikemcquaid: It should still be reported.

externl: TBO I'm not really sure it's an issue with the Berkeley DB build system. The distros which ship it just build with Java 7.

The issue is that if I have Java 7 and 8 installed on my system and I want to build a formula, there's no way to specify a version (unless I missed it in the docs). Setting JAVA_HOME does not work. Brew will always choose Java 8.

xu-cheng: > Brew will always choose Java 8

`depends_on :java => "1.7"`

externl: It does not depend on Java 1.7 though, Java 1.8 builds fine. The issue is if I build with Java 8 (the default if both are installed) then want to use it with Java 7 only project. Which is what this replace fixes.

externl: I will still report this to Oracle. That aside is everyone okay with this in the meantime?

externl:
As a side node I have another commit I was going to make which fixes the build on El Capitan. Would like to get this one in first though.

Library/Formula/ice.rb
externl: Anyone?

DomT4: Can you include a link to that bug submission in the formula? If we can get that done, I'm happy at least.

externl: It's an internal bug submission system, there is no public access. Also, the version of db used by Ice comes from ZeroC, which already includes various patches. I don't see why this one line patch is such a big deal? Especially considering Oracle is no longer supported 5.3 and this version is specific just for ZeroC Ice.

externl: Suggestions? I can just make it a part of our Berkeley DB zip?

externl: Would love to get this in so I can submit an El Capitan fix! :)

dunn: Can you add a comment about having submitted a ticket upstream, so it doesn't end up as a floating patch?

externl: Will do.

externl:
I've added the comment for Berkeley DB. I've also added the various things needed for El Capitan support.

dunn:
Thanks for your patience here; merged in e81b5b03bdb0045c9589093afbaf09136a208d28 ❄️

PR #43840, opened 2015-09-12
llvm: add runtime compiler support libraries

aw1621107:
Some of clang's features, such as adding instrumentation or runtime
sanity checks, did not work with the previous build. This is because
these features relied on runtime support libraries that were not
included in the build.

Even though clang mostly works without the libraries, I still
unconditionally build these libraries  because the errors that clang
emits when these libraries were needed but not available do not do
that good of a job at pointing out the missing libraries. For example,
the errors I got when trying to use clang's address sanitizer were of
the form:
    "missing symbol ___asan_init_, referenced from
      ..."
Someone more experienced than me might quickly realize that these
libraries were supposed to be linked during compilation, but I thought
might as well save users the trouble.

### DIFF
--- a/Library/Formula/llvm.rb
+++ b/Library/Formula/llvm.rb
@@ -49,6 +49,11 @@ class Llvm < Formula
       url "http://llvm.org/releases/3.6.2/clang-tools-extra-3.6.2.src.tar.xz"
       sha256 "6a0ec627d398f501ddf347060f7a2ccea4802b2494f1d4fd7bda3e0442d04feb"
     end
+
+    resource "compiler-rt" do
+      url "http://llvm.org/releases/3.6.2/compiler-rt-3.6.2.src.tar.xz"
+      sha256 "0f2ff37d80a64575fecd8cf0d5c50f7ac1f837ddf700d1855412bb7547431d87"
+    end
   end
 
   bottle do
@@ -79,6 +84,10 @@ class Llvm < Formula
 
     resource "clang-tools-extra" do
       url "http://llvm.org/git/clang-tools-extra.git"
+    end
+
+    resource "compiler-rt" do
+      url "http://llvm.org/git/compiler-rt.git"
     end
   end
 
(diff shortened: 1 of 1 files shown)


aw1621107:
This *might* be a bit too soon. I cannot get the build to complete
consistently, but I think it's my computer, since I can run the build
two times back-to-back, fail the first time, and succeed the second.

Library/Formula/llvm.rb
```diff
@@ -111,6 +120,37 @@ class Llvm < Formula
...
+  # time (11 September 2015). This is not the only incident of its kind; see:
+  #     https://forums.developer.apple.com/thread/13795
+  # so it is possible that this can be removed if/when XCode and/or the CLT get
+  # fixed.
+  #
+  # Builds without these support libraries completed normally, which is why
+  # this change was not required earlier. In addition, the libraries are only
+  # built when clang is built
+
+  env :std if build.with? "clang"
```
DomT4: I'm pretty :-1: on the idea of an new `env :std` addition right off the bat. We're trying to reduce places where that occurs. It'd be nicer to work out the problem with `superenv` here and see if we can tweak that either globally or in-formula to resolve the issue instead. `:std` is a bit of a hack.

aw1621107: Hmmm, ok. I'll take a closer look when I can. Here's what I (think I) know so far:

None of the makefiles I looked at had mixed platforms in their commands, so if it's a bad command it probably isn't local to that folder or surrounding folders. I didn't look at the ENV folder because I thought it just had symlinks. Didn't know it did its own processing. Don't know if that could have messed up; stdout looked ok, with no mixups, but I don't know if what is printed is before or after the command is started.

I'll do some messing around later. And just wondering, what gets passed to the ENV folder that requires so much processing? Symlinking that to the regular binary might help reveal the issue, but that's only if it's in a "normal" format.

aw1621107: Should I open a new issue for this, or is tracking things here ok?

DomT4: Feel free to track here if you like, since it's directly related to the PR.

aw1621107: Is this supposed to happen in the debug shell?
```
Choose an action: 5
When you exit this shell, you will return to the menu.
rp% which ld
/usr/local/Library/ENV/4.3/ld
rp% which clang
/usr/local/Library/ENV/4.3/clang
rp% clang
Undefined symbols for architecture x86_64:
  "_main", referenced from:
     implicit entry/start for main executable
rp% ld
ld: warning: -arch not specified
ld: warning: -macosx_version_min not specified, assuming 10.9
ld: no object files specified for inferred architecture x86_64
rp% clang++
Undefined symbols for architecture x86_64:
  "_main", referenced from:
     implicit entry/start for main executable
rp%
```

aw1621107: Also, would it be problematic if the build bypasses ENV? Looks like llvm is using hardcoded paths from xcode-select or something

aw1621107: All right, managed to get past the error in the formula.

It seems clang++ is being invoked with both `-arch x86_64` and `arch i386`. While I don't think this is normally an issue, this breaks linking iOSSim because it's i386 only. Deleting the x86-64 flag from both the link.txt and flags.make files allowed compilation to continue

aw1621107: Never mind, worked once and didn't work again...

aw1621107: See: bottom of [this](https://forums.developer.apple.com/thread/13795) apple dev thread

Seems the linker wasn't getting all the search paths it needed. In my case at least, it was $(llvm_root)/projects/compiler-rt/lib/asan/CMakeFiles/clang_rt.asan_iossim_dynamic.dir that was causing the linker error. Adding the libraries as additional -L flags to link.txt allowed compilation to advance.

The missing libraries for me were /Application/Xcode-beta.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator9.1.sdk/usr/lib and the system folder inside that folder.

Unfortunately, I get a new error: unknown symbol "_wordexp". wordexp.h is a standard header, and could be easily found in multiple places on my system. It is also included in source files properly. However, none of the dylibs I looked in (all of the ones in Xcode-beta.app and the ones in /usr) had a _wordexp symbol that worked.

aw1621107: @DomT4 Got a piece of the puzzle. It seems setting `HOMEBREW_CCCFG` to anything including the `O` flag is causing only x86_64 architectures for part of the build to be built, instead of both x86_64 and i386. The commands printed in that section have `-arch x86_64 -arch i386` in them, so it wasn't the commands that were incorrect. Dropping into a shell in that section and prefixing `HOMEBREW_CCCFG=sag` to the compile command resulted in both architectures being built.

I'm not sure what the flag is normally; it's just `O` right now, but I think in previous runs it was `Osag`

Thoughts?

mikemcquaid: It'll only build 64-bit binaries unless universal is set.

aw1621107: Hmmm... Thing is that some earlier linker errors complained about building for iOS simulator (i386) but trying to link against libraries for OS X (x86_64). And that worked fine when pointed to the iOS simulator dylibs. I believe those are i386, but I'll check next time I'm at my computer.

I also remember testing unsuccessfully with --universal set once, but don't remember if it was that particular bit that killed the build. I'll check when I can

aw1621107: Yep, you were right. Leapt to a conclusion a bit fast there...

It looks like the iOS simulator libraries require both 32- and 64-bit, though, so I think that'll be set in the formula anyways...

aw1621107:

PR #43841, opened 2015-09-12
llvm: add support for building with polly

aw1621107:
For reference:
    http://polly.llvm.org/get_started.html
    http://polly.llvm.org/example_load_Polly_into_clang.html

Polly is an optimization infrastructure for llvm that performs memory
access optimizations.

Currently set up to support --HEAD-only builds (I think), because
according to the website Polly is not really that stable.


aw1621107:

PR #43842, opened 2015-09-12
llvm: add option to run regression tests

aw1621107:
For reference:
    http://llvm.org/docs/TestingGuide.html

There is a larger test suite available, but that's going in a different
commit because it requires the installation of a separate tool (lnt).


DomT4:
Just FYI, please don't queue up three PRs for the same formula in a row. Combine as much as you possibly can. Queuing up PRs for the same formula like this not only chews up CI time but upsets the bottle logic, sadly.

aw1621107:
Ah, ok. Sorry for that. Yeah, now that I think about it, it should have been 1 PR and 3 commits... I'll keep that in mind

aw1621107:

PR #43849, opened 2015-09-12
bottle: allow updating :any to :any_skip_relocation when keep old

xu-cheng:
cc @mikemcquaid


mikemcquaid:
:+1: but I think it'd also be worth handling the change from default to `:any` too.

xu-cheng:
> I think it'd also be worth handling the change from default to :any too.

We can't. In this situation, it usually means we now relocate more files than used to be(usually text files). But the old bottles still need rebuild.

mikemcquaid:
Ok, cool.

Issue #43854, opened 2015-09-12
Homemade bottles not installable if name doesn't match anything in core/taps


chdiza:
If I have a formula called `foo`, and there is no `foo` in core or any of the taps, then I cannot `brew install` a homemade bottle of foo.

If I do:
```
brew install --build-bottle /path/to/foo.rb
brew bottle foo
```
then everything is fine.  When I then do `brew install a/path/to/foo-1.0.yosemite.bottle.tar.gz`, things fail thus:
```
Error: No available formula for foo
==> Searching formulae...
==> Searching taps...
```

Brew errors like this even if the formula foo.rb, the homemade bottle, and the current dir of my terminal prompt, are in the very same directory.

I note that making homemade bottles for formulas named the same as something in core/taps seems to be working fine, now that that `tar`-related bug has been fixed.

mikemcquaid:
We've said before that we don't really support the `/path/to/foo.rb` format intentionally. I'd suggest putting it in a tap.

chdiza:
But we do support `file:///path/to/foo.rb`, right?  (See 0303a442db03bf5723).

xu-cheng:
> But we do support file:///path/to/foo.rb, right? 

We support loading formula by path and URL. But we DO NOT support bottles created by formulae from path or URL.

And since we already support anytap, there isn't a reason you cannot use tap.

mikemcquaid:
I wouldn't rely on that behaviour being supported forever.

chdiza:
> I wouldn't rely on that behaviour being supported forever.

What behavior?  `file:///path/to/foo.rb`?  Or `/path/to/foo.rb` (without the protocol prefix)?

mikemcquaid:
Either.

PR #43858, opened 2015-09-12
Document what we don't support re local bottles

chdiza:
See #43854.


mikemcquaid:
Thanks for the PR but I'd rather keep this document lean unless more than one person asks about this.

PR #43867, opened 2015-09-13
mitmproxy 0.12.1: download url update

detunized:
https://mitmproxy.org/download/mitmproxy-0.12.1.tar.gz is no longer available, so let's use Github instead.


DomT4:
It's no longer available because there's an update to `0.13` pending. Let's update instead here.

https://mitmproxy.org/download/mitmproxy-0.13.tar.gz

detunized:
I'm afraid I don't know how to update to 0.13. I'm not sure where all those dependency versions come from and how to get their SHAs. I thought this was an quick and easy fix to make at least 0.12.1 work. Currently installing `mitmproxy` not from a bottle is not possible.

PR #43876, opened 2015-09-13
python3 3.5

alex:


DomT4:
You need to remove the `devel` block here and revision everything with a mandatory or recommended `python3` dependency, I believe. CC @tdsmith

alex:
@DomT4 thanks for the quick review! Is there a good way to find everything that depends on python3, or just "grep"?

DomT4:
Tim wrote a blog [here](http://blog.tim-smith.us/2015/08/python-35-transition/) which has a list:

> (The complete list of those packages is: circlator, gubbins, iva, keepassc, lensfun, libsigrokdecode, mypy, pastebinit, ponysay, pulseview, py3cairo, pymummer, pyqt5, retext, and xonsh.)

alex:
Slick, thanks!

alex:
Bumped the revisions on a bunch of those, however (allowing for the possibility that I'm very dumb), I couldn't find the formulas for some of those.

DomT4:
Some of them will be in taps like `science` and `python`. Those will have to be done very shortly after this PR is merged into the core.

alex:
Ah. Thanks.

tdsmith:
We added `treeline` recently and I didn't update my post.

Current list is:
```
$ find -f /usr/local/Library/Formula -f /usr/local/Library/Taps -- -name \*.rb -exec grep -E 'depends_on [:"]python3' {} + | grep -v -e :build -e :optional | cut -f1 -d: | xargs brew info --json=v1 | jq -r '.[] | .full_name'
keepassc
lensfun
mypy
pastebinit
ponysay
py3cairo
pyqt5
treeline
xonsh
homebrew/python/pymummer
homebrew/python/retext
homebrew/science/circlator
homebrew/science/gubbins
homebrew/science/iva
homebrew/science/libsigrokdecode
homebrew/science/pulseview
```

alex:
@tdsmith Add `revision` for `treeline` as well.

alex:
Tests appear to be failing, but they don't _look_ like it's py3 related, is it possible these packages were broken on master?

tdsmith:
Sure is. Some of these packages are running into problems with our sandboxed builds and will need some effort to fix. pyqt5 should work before we pull this.

DomT4:
`keepassc` works fine locally on the current version, so I presume that did break here somehow.

`mypy` is breaking because the `PYTHONPATH` is hardcoded to look for the `3.4` version.

`pastebinit` can't be build by the bot, so that one isn't your fault.

`ponysay` is trying to write to `/usr/local/Cellar/ponythink` and is consequently failing inside the sandbox.

`py3cairo` has been a problem for a while because it hooks the Framework directly.

`pyqt5` is trying to write a directory into `qt5`'s prefix, which fails sandboxing:
```
mkdir: /usr/local/Cellar/qt5/5.5.0/plugins/PyQt5: Operation not permitted
make[1]: *** [install_target] Error 1
make: *** [sub-qmlscene-install_subtargets-ordered] Error 2
```

alex:
Ok, I believe I fixed `keepassc` and `mypy`.

tdsmith:
I can work on this later if you'd like but can you construct PYTHONPATH like this instead?

```
pyver = Language::Python.major_minor_version "python3"
... "... python#{pyver}/site-packages ..." ...
```

alex:
yup, will do for both `keepassc` and `mypy`

alex:
Done!

tdsmith:
ponysay will work if we install with --without-shell-completion; it's doing some unclever string replacement that I'm not in the mood to fix correctly

alex:
@tdsmith Not know anything about that package, is that something that could go into it's own separate PR?

DomT4:
I get this problem if I pass `--without-shell-completion` to `ponysay`:

```
  File "./setup.py", line 1188, in <module>
    Setup()
  File "./setup.py", line 292, in __init__
    self.install(conf)
  File "./setup.py", line 612, in install
    cmdarr = ['install-info', '--entry=Miscellaneous', '--description=' + conf['info-install'], '--dir-file=' + conf['info'] + '/dir', dest]
UnboundLocalError: local variable 'dest' referenced before assignment
```

tdsmith:
that's what I get for not actually testing a solution I propose. I'll take care of ponysay elsewhere.

alex:
coolio, thanks

DomT4:
These are the sandbox issues with `pyqt5`:
```
Sep 13 19:32:29 sandboxd[26550]: install(13474) deny file-write-create /usr/local/Cellar/qt5/5.5.0/plugins/designer/libpyqt5.dylib
Sep 13 19:32:29 sandboxd[26550]: mkdir(13480) deny file-write-create /usr/local/Cellar/qt5/5.5.0/plugins/PyQt5
```
Off the top of my head, I'm not really sure how to get around those, given the `plugins` directory inside `qt5` isn't actually made public anywhere outside the `qt5` Cellar.

alex:
@DomT4 That'd be an issue on master as well right?

DomT4:
Yes. Tim has opened #43885 which may well resolve the issue.

alex:
Woo, @tdsmith is on top of stuff! I see there's also a PR for ponysay up.

alex:
@tdsmith now that the `pyqt5` and `ponysay` patches are landed, do you want me to merge `master` in?

tdsmith:
nah, they shouldn't conflict; I'll just kick off a new build. @BrewTestBot test this please

tdsmith:
Thanks for your help, Alex & Dominyk!

PR #43884, opened 2015-09-13
carthage 0.9.2

guidomb:


mdiep:
0.9.1 was just released to fix a crash in 0.9. It'd be nice update the homebrew formula directly to 0.9.1.

guidomb:
I have amended the commit bumping the version to 0.9.1

gfontenot:
Any word on this? Getting this in would mean Great Things for using Carthage on CI servers.

DomT4:
The test needs fixing. We run under a sandbox now, which you can reproduce locally for testing with `brew test --sandbox xyz`:

```
==> /usr/local/Cellar/carthage/0.9.1/bin/carthage update
A shell task failed with exit code 255:
error: cannot open FETCH_HEAD: Operation not permitted

*** Fetching xcconfigs
Error: carthage: failed
Failed executing: /usr/local/Cellar/carthage/0.9.1/bin/carthage update
```

The test may need to be tweaked accordingly.

guidomb:
Well actually running the `brew test --sandbox carthage` for Carthage 0.8 fails in my local machine. Also the [details](bot.brew.sh/job/Homebrew%20Pull%20Requests/33127/) in the PR status section seems to be broken.

guidomb:
Do I need to manually install the latest version of Carthage in order to run brew test against it?

guidomb:
@DomT4 Could it be a temporary error? Is it possible to run the tests again?

guidomb:
OK I finally could run the tests against the latest version of carthage. The problem seems to be `--sandbox` as you said. For me little understanding of how Carthage works ( @mdiep may be you can help us here ) the problem seems to be related with Carthage trying to checkout the git repository in `~/Library/Caches/org.carthage.CarthageKit/dependencies/` and for some reason related with sandboxing the process is not allowed to write in that directory.  

@DomT4 Is it possible to tell brew to let the process write to that directory?
@mdiep Is it possible to tell carthage to checkout the dependencies in a different location (maybe the current working directory)?

mikemcquaid:
@mdiep How is Carthage getting that location? We override `$HOME` for tests so we need to coax it to writing somewhere else somehow?

DomT4:
> Do I need to manually install the latest version of Carthage in order to run brew test against it?

Yes. I expect the error would reproduce with the last version as well though.

>  Could it be a temporary error?

It won't be a temporary error.

>  Is it possible to tell brew to let the process write to that directory?

It is, sure. But we use the same basic sandbox for all formulae building, so permitting carthage to write to that directory will allow everything installed to write to that directory, which semi-ruins the point of sandboxing :smile_cat:.

We need to override the location here, if possible, per Mike's comment above. If it is ignoring the `$HOME` variable in finding that path I suspect it may be hardcoded as a default somewhere.

guidomb:
@DomT4 Yeah whitelisting that directory just for carthage makes no sense.

@mdiep The problem seems to be [here](https://github.com/Carthage/Carthage/blob/d6211444e6c30ebccd9d3ad2c676ac4eb6770274/Source/CarthageKit/Project.swift#L17), isn't it?

For what I read in Apple docs about `NSUserDomainMask`
> The user’s home directory—the place to install user’s personal items (~). 
This is probably not using the `$HOME` env variable to expand the user's home directory path.

mikemcquaid:
@guidomb @mdiep To be fair to us that's probably our/Apple's bad for not changing that directory accordingly. If there's a better way to override stuff like that I'm all :ear:s.

mdiep:
What does `~/Library/Caches/org.carthage.CarthageKit/dependencies/` resolve to in the sandbox? Is it in the home directory? What are the permissions there?

We might need to find the directory in a different way, add some better testing around the fallback, or add a better fallback. I don't know enough of the details around what's happening.

mikemcquaid:
@mdiep The short version of the issue is that when you run `brew test` we set `$HOME` to somewhere in `/tmp` but this is still trying to write to the current user's home directory (`~`/`/Users/something`) rather than under the new `$HOME` (`/tmp/something`).

DomT4:
This is the more verbose readout, if it helps:
```
==> /usr/local/Cellar/carthage/0.8/bin/carthage update
Failed to write to /Users/Dominyk/Library/Caches/org.carthage.CarthageKit/dependencies: Error Domain=NSCocoaErrorDomain Code=513 "You don’t have permission to save the file “dependencies” in the folder “org.carthage.CarthageKit”." UserInfo={NSFilePath=/Users/Dominyk/Library/Caches/org.carthage.CarthageKit/dependencies, NSUnderlyingError=0x7faacbd28d30 {Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted"}}
==> Sandbox log
Sep 16 14:21:15 sandboxd[97326]: carthage(37022) deny file-write-create /Users/Dominyk/Library/Caches/org.carthage.CarthageKit
```

guidomb:
@mdiep `		NSFileManager.defaultManager().URLForDirectory(NSSearchPathDirectory.CachesDirectory, inDomain: NSSearchPathDomainMask.UserDomainMask, appropriateForURL: nil, create: true, error: error)
` expands (in my case) to `/Users/guidomb/Library/Caches/org.carthage.CarthageKit/dependencies/` which is fine but it does it by ignoring the actual value of `$HOME`. Homebrew in `--sandbox` mode overrides `$HOME` but that value gets ignored by Apple's (or at least that is my conclusion).

The fallback URL is not being used because the directory exists the problem is that carthage (when being tested in sandbox mode) does not have write permissions.

guidomb:
@mdiep A hacky solution would be to somehow detect that Carthage is being tested by homebrew (probably by using an env variable, something like `RUNNING_HOMEBREW_TEST`) and construct the URL using the `$HOME` variable.

mdiep:
@guidomb I think that code could use some improvement. We should check for write permissions for the actual Carthage cache directory and then fall back to the other URL. This seems like a legitimate Carthage issue.

If, after getting Carthage to try the fallback URL, we still don't have the right directory, then we should try to use `$HOME` for the callback.

guidomb:
@mdiep I could try to that later today or maybe tomorrow if you want

mdiep:
That would be great!

guidomb:
@mdiep I have submitted a PR to fix this issue. I've run hombrew test locally and it passes. Once that PR is merged you can release a new patch version and I can update this PR for Carthage `0.9.2`.

dreid93:
This would be great :+1: The disappointing thing is I was just about to submit a PR for this. Lesson learnt, check open PRs first.

rastersize:
@guidomb Carthage [0.9.2 has been released](https://github.com/Carthage/Carthage/releases/tag/0.9.2) :smiley: 
Revision `c82c3d1cb6f35a146793a914f1fae0e14d95ce0c`.

rastersize:
@guidomb The build failed as `:tag => "0.9.1"` and not `.2`.

guidomb:
Yeah my bad sorry about that

rastersize:
No need to be sorry, big thanks for updating the formula! :smiley:

guidomb:
OK the problem seems to be that travis is using xcode7 but carthage requires xcode6

guidomb:
```
==> Configuration

HOMEBREW_VERSION: 0.9.5

ORIGIN: https://github.com/Homebrew/homebrew.git

HEAD: 5e1cf3cd653ccb7b5605195fd3115e503459c583

Last commit: 7 minutes ago

HOMEBREW_PREFIX: /usr/local

HOMEBREW_REPOSITORY: /usr/local

HOMEBREW_CELLAR: /usr/local/Cellar

HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com

CPU: single-core 64-bit nehalem

OS X: 10.10.5-x86_64

Xcode: 7.0

CLT: 6.4.0.0.1.1435007323

Clang: 7.0 build 700

X11: 2.7.7 => /opt/X11

System Ruby: 2.0.0-p481

Perl: /usr/bin/perl

Python: /usr/bin/python

Ruby: /usr/bin/ruby => /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby

Java: 1.8.0_60
```

mikemcquaid:
Unfortunately we can only build with the current stable Xcode release :cry:.

mdiep:
:sob:

guidomb:
@mikemcquaid I've just run `./bin/brew test --sandbox carthage` locally with XCode 6.4 and everything works fine. Would it be possible to release a new version of carthage even if hombrew tests fail in travis?

@mdiep Something to take into account when running hombrew tests locally is that the folder `~/Library/Caches/org.carthage.CarthageKit` already exists carthage fails because the code I patched is not even run and something is trying to write to this folder. I was about to freak out but I deleted the folder and run hombrew tests locally and they work fine. I don't think this should be a problem with travis because travis should start with a clean directory.

mikemcquaid:
@guidomb We actively nag Homebrew users to upgrade Xcode so if it's broken in Travis it'll be most likely broken for most Homebrew users too unless we can build binaries.

guidomb:
@mikemcquaid I see. Would it be possible to distribute a precompiled version?
@mdiep Worst case scenario we can use hombrew cask and distribute the dmg. What do you think about that?

mikemcquaid:
@guidomb As a one-off for this release: perhaps. Is there any ETA on when Xcode 7 will be supported?

guidomb:
@mdiep May be able to answer that question.

mdiep:
Distributing the `.pkg` seems like a good idea if that's acceptable. We're going to be tied to specific Xcode versions because of the very tight coupling between Swift and Xcode.

This might be an issue with Xcode 7.1 when it's released too. There have been multiple incompatible versions of Swift in the past 14 months. Each requires all of our dependencies to update newer versions of Swift.

Xcode 7 / Swift 2 support is being worked on. But https://github.com/Carthage/Commandant needs to be updated first, and that's stalled because of problems getting those unit tests to run on Travis with Swift 2. It's probably 1-2 weeks out.

ReDetection:
@mikemcquaid either Xcode7 is not stable enough, or homebrew failed at that tactic. This is unbelievable that Xcode update causes the software build failures. Just frustrating.
Check how many active formulaes fails to build now.

mdiep:
@ReDetection Don't take it out on @mikemcquaid; it's Apple's fault. :relaxed:

ReDetection:
@mdiep yea, I agree, but since they are not likely to fix this, shouldn't we adopt to it?

mikemcquaid:
@ReDetection If you've installed Xcode through the App Store it'll autoupdate anyway so, yeh: blame Apple.

gfontenot:
@mikemcquaid Is there a reason that Carthage shouldn't distribute a pre-built binary? I'd be concerned about Carthage needing to stay in sync with Swift/Xcode changes every couple of months just for Homebrew compatibility.

dunn:
We could move Carthage to homebrew-binary if you think this won't be resolved soon..

mdiep:
It will be resolved soon for some definition of soon.

The issue is that we'll face this problem with each new X.X release of Xcode as Swift changes in incompatible ways.

DomT4:
> The issue is that we'll face this problem with each new X.X release of Xcode as Swift changes in incompatible ways.

Just to be explicit on this, we'll only see this on something like 6.4 => 7.0 Xcode updates?

mdiep:
It depends on how often Apple changes Swift. Xcode 6.X had Swift 1.0, 1.1, and 1.2 last year, which were all incompatible. It's unclear how many incompatible versions of Swift we'll see this year.

DomT4:
When Apple open-sources Swift this year, I wonder if it'd be possible to vendor Carthage a specific Swift to use. That way we'd be able to update freely and you wouldn't have to be constantly on top of Swift changes to get the latest into Homebrew. Not an immediate solution obviously, but perhaps an option at some point.

PunKeel:
:+1: :(

mikemcquaid:
I'm OK with this formula being changed to just install a binary for now.

kkpoon:
when will the pull request be merged?

DomT4:
We're happy to accept a binary for now pending investigation of further possibilities going forwards but not sure what the status is on this PR itself.

guidomb:
I don't know much about homebrew nor how to make a formula install a precompiled binary. If someone wants to give it a try go for it. Should I close this PR and let somebody else try to make a new formula?

ReDetection:
Because I have Xcode 6 installed and it's current selected Xcode in my system, I would rather install Carthage by building it from source with formula from this pull request: 
```
brew uninstall && brew install https://raw.githubusercontent.com/guidomb/homebrew/6a097e7b97c9104e694e33b3878ed0f29da201c8/Library/Formula/carthage.rb
```
I believe it's the most proper way to install latest Carthage until it support swift2. 

@guidomb no, I don't think you should close it. Let others know there is a problem and somebody working on it

mikemcquaid:
Probably the best thing for now is to submit Carthage to Homebrew Cask.

ReDetection:
sure, it's the fastest workaround to get it working for everyone

BryanJBryce:
Is there a version of Carthage that can be installed using brew that will still work?

abbeycode:
Hey all, I'm anxiously awaiting this Brew release of Carthage to use on CI builds. What's the next step involved?

mdiep:
The Swift 2 branch of Carthage was merged recently. I'm planning to create a new release in the next day and then this PR can be updated.

mdiep:
I just released Carthage [0.10: 7 Levels of Xcode](https://github.com/Carthage/Carthage/releases/tag/0.10) which uses Swift 2.

@guidomb Would you be up for updating this PR for that release?

ikesyo:
`depends_on :xcode` should be also updated to `"7.1"` for 0.10, I think. :smiley:

bfontaine:
#45747 was just merged; Closing this one.

PR #43890, opened 2015-09-13
Add surfraw build dependencies

keith:
Previously installing surfraw with `--HEAD` failed since it needs autoconf and automake to build from source.

### DIFF
--- a/Library/Formula/surfraw.rb
+++ b/Library/Formula/surfraw.rb
@@ -4,7 +4,12 @@ class Surfraw < Formula
   url "https://surfraw.alioth.debian.org/dist/surfraw-2.2.9.tar.gz"
   sha256 "aa97d9ac24ca4299be39fcde562b98ed556b3bf5ee9a1ae497e0ce040bbcc4bb"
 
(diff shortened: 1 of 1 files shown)


Library/Formula/surfraw.rb
```diff
@@ -12,6 +12,9 @@ class Surfraw < Formula
     sha1 "b17ead2b3f8030e659d0af4b44332f76397607f5" => :mountain_lion
   end
 
+  depends_on "autoconf" => :build
```
DomT4: If these are just needed for `head` installations you should restrict the dependencies to `head`.

```ruby
head do
  url "git://git.debian.org/surfraw/surfraw.git", :shallow => false

end
```

keith: Awesome, updated

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/ea0f1246bfd9fdf3558ee84145aaf2b142a1e633. Thank you for your contribution to Homebrew @keith!

keith:
Thanks!

PR #43892, opened 2015-09-13
dshb v0.1.0 (new formula)

beltex:
Hello Homebrew! :beers: 

This is a self-submission of a new formula, [dshb](https://github.com/beltex/dshb), my sincerest apologies for breaking the rules! :) Being a Swift based tool I thought I’d go ahead and write the initial formula to kick things off incase any difficulties were to come up in the process. I hope this is alright, if not, by all means close this.

### What

An OS X system monitor in Swift, inspired by <a href="https://en.wikipedia.org/wiki/Top_(software)">top</a> & [htop](https://github.com/hishamhm/htop)

### Notes

* Passes `brew audit dshb --online --strict`

* Used [SourceKitten](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/sourcekitten.rb), [Carthage](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/carthage.rb), (both Swift based tools) and [htop-osx](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/htop-osx.rb) formulae as reference

* `depends_on :xcode => ["6.3", :build]` is used because Xcode 6.3 (up to 6.4) comes with the Swift 1.2 compiler. See https://github.com/Homebrew/homebrew/issues/42864#issuecomment-133078788 for more

* Since `dshb` uses multiple Git submodules, I thought it be best to point to the source code archive such that the checksum holds more value. This is a custom source code archive, as the ones generated from GitHub do not include submodules (see https://github.com/beltex/dshb/issues/29 for more detail)

### Questions

* Can `depends_on` specify a version range? Xcode 6.3 to 6.4 (but not higher) have the Swift 1.2 compiler that is required

* Can `depends_on` also warn that the Xcode version in question is active or not? As in `xcode-select -s`. Currently, if a user has Xcode 6.3 and 7 installed (with 7 being active) for example, the `depends_on` check will pass, but the build will fail since when `swiftc` (the Swift compiler) is called it will link to version 2.0 of the compiler that has breaking changes.

* I'm not sure how to express the fact that `dshb` supports OS X 10.9, yet Mavericks can't officially build it because Xcode 6.3 (and above) doesn't support it (really it's Xcode.app that doesn't, the Swift compiler that comes bundled should work fine). The `dshb` binary we get is _“universal”_ in that it works on 10.9 to 10.11. Two options I can see for this. One is when Swift goes open source later this year, folks can get there hands on Swift independent of Xcode and thus the formula could instead depend on the Swift compiler version instead of Xcode (see https://github.com/Homebrew/homebrew/issues/42864#issuecomment-133078788 for more). Alternatively, could have the bottle be used in that case, with no option to build from source.

* Not much we can do for the test requirement, did the same thing that `htop-osx` formula does, booting and passing `q` as input to quit and asserting success. Beyond that, all I can think of is running `-v` or `-h` and checking the output. Is there something better that can be done?

Reference issue: https://github.com/beltex/dshb/issues/1

### DIFF
--- a/Library/Formula/dshb.rb
+++ b/Library/Formula/dshb.rb
@@ -1,0 +1,19 @@
+class Dshb < Formula
+  desc "OS X system monitor in Swift"
+  homepage "https://github.com/beltex/dshb"
+  url "https://github.com/beltex/dshb/releases/download/v0.1.0/dshb-0.1.0-source.zip"
+  sha256 "efed42a2be0dbc6de3b22b314b582b1d6517922b72e08e063f3d1d3810a782f9"
+
+  depends_on :xcode => ["7.0", :build]
+
+  def install
+    system "make", "release"
+    bin.install "bin/dshb"
+    man1.install "doc/dshb.1"
+  end
+
+  test do
+    ENV["TERM"] = "xterm"
+    pipe_output("#{bin}/dshb", "q", 0)
+  end
+end


tdsmith:
We don't have a way to specify a maximum Xcode version or to target a particular Swift version, we don't have a mechanism to indicate binary platform support independently from build support, and we don't generally support anything but the newest Xcode version on any given platform. For those reasons I'm not sure that distributing a source build through Homebrew is a good solution for you.

Hosting a tap of your own that installs a binary distribution might be a useful alternative. We have [a document](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) on creating and maintaining taps.

It looks like a neat project and I'd be happy to see it in core once the swift ecosystem stabilizes a little.

Library/Formula/dshb.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def install
+    system "make", "release"
+    bin.install "bin/dshb"
+    man1.install "doc/dshb.1"
+  end
+
+  test do
+    pipe_output("#{bin}/dshb", "q")
+    assert $?.success?
```
tdsmith: You can pass a third argument to pipe_output which is the exit status to assert, like `pipe_output("#{bin}/dshb", "q", 0)`

beltex:
Hi @tdsmith!

> We don't have a way to specify a maximum Xcode version

>  we don't have a mechanism to indicate binary platform support independently from build support

> we don't generally support anything but the newest Xcode version on any given platform.

Thanks for the clarifications!

> or to target a particular Swift version

Indeed! This was something I discussed with @mikemcquaid recently (see https://github.com/Homebrew/homebrew/issues/42864#issuecomment-133078788). Once Swift goes open-source later this year, this can be done.

> For those reasons I'm not sure that distributing a source build through Homebrew is a good solution for you.

> I'd be happy to see it in core once the swift ecosystem stabilizes a little.

Thats totally fair, though I would say that formulae like [Carthage](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/carthage.rb), [SourceKitten](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/sourcekitten.rb), and [SwiftLint](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/swiftlint.rb) are all examples of Swift based tools that face all these same issues. Since Homebrew defaults to bottles I think it’s been smooth for them (not that building from source is a problem), though of course Mavericks users are left out. I thought I’d raise the points as I hadn’t seen discussion on them previously.

> Hosting a tap of your own that installs a binary distribution might be a useful alternative. We have a document on creating and maintaining taps.

Cool, thanks! Definitely seems like the way to go for the Mavericks case till Swift goes open-source.

> It looks like a neat project

Oh thanks! :)

**P.S.** Any additional logs indicating why the build failed?

Library/Formula/dshb.rb
beltex: Fixed! Thanks!

bfontaine:

beltex:
Just a friendly update! :)

I’ve updated this PR to the latest version of dshb (`0.1.0`) that uses Xcode 7 and Swift 2.0.

@bfontaine As for the test failure, not sure whats happening, `brew test --verbose dshb` is running fine on my machine (Yosemite). Any thoughts?

DomT4:
It looks like it's trying to open a terminal and doesn't understand what terminal it is?
```
==> /usr/local/Cellar/dshb/0.1.0/bin/dshb
Error opening terminal: unknown.
Error: dshb: failed
<0> expected but was
<1>.
```
Something like setting the `TERM` as `xterm` or [this sort of thing](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/ne.rb#L23) may work here.

beltex:
@DomT4 Great catch, setting `TERM` fixed it! Strangely now just the Yosemite build is failing with no logs :(

bfontaine:

beltex:
Thanks for kicking off another build!

Here’s the error:

```
** INTERNAL ERROR: Uncaught exception **
Exception: Unable to lookup com.apple.CoreSimulator.CoreSimulatorService in the bootstrap.  This can happen if running with a sandbox profile.  When running with a sandbox profile, make sure that com.apple.CoreSimulator.CoreSimulatorService.xpc is owned by root, not group writable, and not world writable.  See <rdar://problem/22142915>.
```

Strange. Looks like SourceKitten (also a Swift based CLI tool) ran into this recently in #47539. As mentioned in that PR, seems odd that the iOS simulator is being invoked for an OS X CLI tool. Seems that there was an issue with the Yosemite CI machine that caused it.

DomT4:
@BrewTestBot test this please

beltex:
All green! Awesome! :)

mikemcquaid:

For future reference the preferred commit message format for new formulae is `dshb 0.1.0 (new formula)` (no `v`; sorry for extreme pedantry).

beltex:
np! :) Thank you so much Homebrew team! :beers: 

> For future reference the preferred commit message format for new formulae is `dshb 0.1.0 (new formula)`

ah, noted!

> (no v; sorry for extreme pedantry).

:laughing:

Not at all, totally understand!

PR #43895, opened 2015-09-14
fossil: don't pass --with-tcl unless CLT available

### LINKED ISSUE #42493: fossil failed to build on 10.11
https://gist.github.com/967082d2abb9a0ef0304

On OS X 10.11 PB4, no CLT. It's not finding Tcl.

```
Error: Cannot find Tcl on your system
Try: 'configure --help' for options
```

DomT4:
This seems to resolve the problem in #42493, but since there's no explicit `--without-tcl` option, not 110% sure.

Closes #42493.

### DIFF
--- a/Library/Formula/fossil.rb
+++ b/Library/Formula/fossil.rb
@@ -1,9 +1,10 @@
 class Fossil < Formula
   desc "Distributed software configuration management"
   homepage "https://www.fossil-scm.org/"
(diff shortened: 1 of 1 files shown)


apjanke:
Works for me: I tested on a no-CLT 10.9.5. system and the build succeeded.

DomT4:
Thanks for checking @apjanke! :star2:

PR #43902, opened 2015-09-14
fix update-test

xu-cheng:
Previously, `brew update-test` is run against master branch of local
repo. However, we test PR using a detached branch in `brew test-bot`.
The result is `brew update-test` will always be up-to-date in `test-bot`.

To fix it, we create two local copies of git repo, and set master branch
to start and end sha1 correspondingly. After that, `brew update` will be
run to simulate the change between start and end sha1.

cc @mikemcquaid


mikemcquaid:
If it works: :+1:.

Issue #43906, opened 2015-09-14
libtomcrypt failed to build on 10.11


apjanke:
https://gist.github.com/5e9b2308816dd4e01b73

Same failure on 10.10. Works on 10.9.

Upstream issue: same build error happens when building outside Homebrew. Upstream issue: libtom/libtomcrypt#82.

mikemcquaid:
It's been broken for 2 major OS X releases so: boneyard time.

apjanke:
Doesn't build on last two OS X releases; current 1.17 release is from 2010; upstream bug open for five months. Yeah, boneyard time.

kimsama:
Got the following error on OSX 10.11.4 when it installs 'libtomcrypt'.

```
==> make library
Last 15 lines from /Users/kims/Library/Logs/Homebrew/libtomcrypt/01.make:
      :"0" (word),"I" (i));
                       ^
./src/headers/tomcrypt_macros.h:277:24: error: constraint 'I' expects an integer constant expression
      :"0" (word),"I" (i));
                       ^
./src/headers/tomcrypt_macros.h:368:24: error: constraint 'J' expects an integer constant expression
      :"0" (word),"J" (i));
                       ^
./src/headers/tomcrypt_macros.h:376:24: error: constraint 'J' expects an integer constant expression
      :"0" (word),"J" (i));
                       ^
4 errors generated.
make: *** [src/ciphers/aes/aes.o] Error 1
make: *** Waiting for unfinished jobs....
make: *** [src/ciphers/aes/aes_enc.o] Error 1

READ THIS: https://git.io/brew-troubleshooting
libtomcrypt was moved to homebrew-boneyard because it has unfixable issues.
Please do not file any issues about this. Sorry!
```

mikemcquaid:
> libtomcrypt was moved to homebrew-boneyard because it has unfixable issues.
> Please do not file any issues about this. Sorry!

@kimsama We don't support `libtomcrypt` any more, sorry.

kimsama:
ok, thanks, anyway. :-)

PR #43907, opened 2015-09-14
libtommath, libtomcrypt: new website for libtom project

apjanke:
STATUS: Don't merge yet; status of upstream sites is unclear.

Per their website, now at www.libtom.net:

> This website is back, and libtom has found new maintainers.

### DIFF
--- a/Library/Formula/libtomcrypt.rb
+++ b/Library/Formula/libtomcrypt.rb
@@ -1,7 +1,7 @@
 class Libtomcrypt < Formula
   desc "Modular and portable cryptographic toolkit"
-  homepage "http://www.libtom.net/?page=features&whatfile=crypt"
-  url "http://www.libtom.net/files/crypt-1.17.tar.bz2"
+  homepage "http://www.libtom.net/LibTomCrypt/"
+  url "https://github.com/libtom/libtomcrypt/releases/download/1.17/crypt-1.17.tar.bz2"
   mirror "https://distfiles.macports.org/libtomcrypt/crypt-1.17.tar.bz2"
   sha256 "e33b47d77a495091c8703175a25c8228aff043140b2554c08a3c3cd71f79d116"
   head "https://github.com/libtom/libtomcrypt.git"
--- a/Library/Formula/libtommath.rb
+++ b/Library/Formula/libtommath.rb
@@ -1,10 +1,10 @@
 class Libtommath < Formula
   desc "C library for number theoretic multiple-precision integers"
-  # homepage/url down since ~May 2015
-  homepage "http://libtom.org/?page=features&newsitems=5&whatfile=ltm"
-  url "http://libtom.org/files/ltm-0.42.0.tar.bz2"
+  homepage "http://www.libtom.net/LibTomMath/"
+  url "https://github.com/libtom/libtommath/releases/download/v0.42.0/ltm-0.42.0.tar.bz2"
   mirror "https://distfiles.macports.org/libtommath/ltm-0.42.0.tar.bz2"
   sha256 "7b5c258304c34ac5901cfddb9f809b9b3b8ac7d04f700cf006ac766a923eb217"
+  head "https://github.com/libtom/libtommath.git"
 
   bottle do
     cellar :any_skip_relocation


Library/Formula/libtomcrypt.rb
```diff
@@ -1,10 +1,10 @@
 class Libtomcrypt < Formula
   desc "Modular and portable cryptographic toolkit"
-  # homepage/url down since ~May 2015
-  homepage "http://libtom.org/?page=features&whatfile=crypt"
-  url "http://libtom.org/files/crypt-1.17.tar.bz2"
+  homepage "http://www.libtom.net/?page=features&newsitems=5&whatfile=crypt"
+  url "http://www.libtom.net/files/crypt-1.17.tar.bz2"
   mirror "https://distfiles.macports.org/libtomcrypt/crypt-1.17.tar.bz2"
   sha256 "e33b47d77a495091c8703175a25c8228aff043140b2554c08a3c3cd71f79d116"
+  head "https://github.com/libtom/libtomcrypt"
```
DomT4: Need a `.git` on the end of this URL.

Library/Formula/libtommath.rb
```diff
@@ -1,10 +1,10 @@
 class Libtommath < Formula
   desc "C library for number theoretic multiple-precision integers"
-  # homepage/url down since ~May 2015
-  homepage "http://libtom.org/?page=features&newsitems=5&whatfile=ltm"
-  url "http://libtom.org/files/ltm-0.42.0.tar.bz2"
+  homepage "http://www.libtom.net/?page=features&newsitems=5&whatfile=ltm"
+  url "http://www.libtom.net/files/ltm-0.42.0.tar.bz2"
   mirror "https://distfiles.macports.org/libtommath/ltm-0.42.0.tar.bz2"
   sha256 "7b5c258304c34ac5901cfddb9f809b9b3b8ac7d04f700cf006ac766a923eb217"
+  head "https://github.com/libtom/libtommath"
```
DomT4: Need a `.git` on the end of this URL.

apjanke:
Repo URLs fixed.

`brew install --HEAD libtomcrypt` from this new URL is failing for me on 10.11, but it looks like the same error as the latest actual release in #43906.

apjanke:
Bump... can this be pulled in? Previous issues are fixed and checks pass.

dunn:
😰 http://bot.brew.sh/job/Homebrew%20Pull%20Requests/34227/

apjanke:
D'oh! I'll check that out.

apjanke:
Known issue: #43906; it's failing on 10.10 and 10.11.

Can we just merge this website fix since it's not a regression, and deal with the build failure under the other issue?

sjaeckel:
Not a good idea to update that, as the owner of libtom.net is currently not reachable and I don't know when this will change.
I hope to get libtom.org back up running.

apjanke:
How about now? It looks like libtom.net has the same contents as libtom.org now, and libtom.org is back online. Should we update this with links in to the new libtom.org site layout?

sjaeckel:
Yes, libtom.net and libtom.org are functional now.
As I've only control over libtom.net you should probably go for that domain.
Download URLs are currently only from github.

sjaeckel:
Regarding the site layout, yes that will be the one you want to update your links to.
You may ask youself why libtom.net is back - long story short, I was able to transfer it to me

apjanke:
Cool. I've revised and rebased this PR to use libtom.net and reflect the new site structure and GitHub releases. Works for me locally. Look good?

Download URLs from GitHub are fine for us, as long as they're for tagged releases, like you're doing here.

The new `libtom.net` web site is still referencing `libtom.org`, both in the site text (e.g. "Welcome to the LibTom Projects home here at libtom.org" on the front page) and in link URLs (e.g. in the navigation bar). Is there a repo for the web site where we can put bug reports/PRs for this?

If these new URLs work out, I'll try to add a couple little tests as part of this PR so the formulae pass our CI.

mikemcquaid:
@apjanke LGTM when :green_heart:

DomT4:
Would be good to pull bottles here when you merge. The `libtomcrypt` is currently lacking an El Cap one.

apjanke:
No bottles (at least for libtomcrypt), due to outstanding issue #43906, which breaks the build on 10.10 and 10.11.

I think the thing to do is just merge this anyway and deal with the build failure separately: this URL fix will at least let brew-using devs download the files and attempt the build so they can debug and fix it.

DomT4:
Ah, alright. Feel free in that case!

PR #43909, opened 2015-09-14
lha: add head and test

tomyun:


Library/Formula/lha.rb
```diff
@@ -1,14 +1,32 @@
...
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+  end
+
   def install
+    if build.head?
+      system "aclocal"
+      system "automake", "-a"
+      system "autoheader"
+      system "autoconf"
```
DomT4: Is there not a `./autogen.sh` or `autoreconf` step that handles all of this? Most upstreams put something convenient in place.

tomyun: That's true. I updated it with `autoreconf` as indicated by [README](https://osdn.jp/projects/lha/scm/git/lha/).

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/b0338f7c5f3237279ea488e72d96b734fdf35d4e. Thank you once more @tomyun!

PR #43911, opened 2015-09-14
pythons: pip, setuptools

tdsmith:
Didn't want to block python3 on adding this but it'll be good to freshen up a little bit.


xu-cheng:
I suggest you just merge this and use [Homebrew Testing](http://bot.brew.sh/job/Homebrew%20Testing/) to generate bottles. So it will build for El Capitan

tdsmith:
works for me. thanks for suggesting!

tdsmith:
http://bot.brew.sh/job/Homebrew%20Testing/884/

PR #43932, opened 2015-09-14
openrtsp 2015.08.07

apjanke:
Fixes another 404 on the previous version.

### DIFF
--- a/Library/Formula/openrtsp.rb
+++ b/Library/Formula/openrtsp.rb
@@ -1,8 +1,8 @@
 class Openrtsp < Formula
   desc "Command-line RTSP client"
   homepage "http://www.live555.com/openRTSP"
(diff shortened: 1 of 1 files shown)


dunn:
Any chance of tossing a test in there while you're at it? :heart_eyes_cat:

apjanke:
Not for a couple days. There don't seem to be any usable "no-op" operations like `--help` or `--version`, and it would take a little bit to put together an actual input-driven test.

DomT4:
Will merge as-is, but if you wanted to chase this in at some point with a test that'd certainly be great. Merged in https://github.com/Homebrew/homebrew/commit/c14e56c065ebda0218d96df2be044c35c4022e0e. Thank you @apjanke!

PR #43934, opened 2015-09-14
gpg-agent 2.0.29

chdiza:
Looks like we forgot to bump this when we bumped `gnupg2`.  I think this means that anyone who's brewed the latest `gnupg2` has an outdated agent.  But I don't *think* this means we should make a revision bump, because I don't *think* anything significant about the 2.0.29 release had to do with the agent.

I could however be totally wrong.  Paging doctor @DomT4!

### DIFF
--- a/Library/Formula/gpg-agent.rb
+++ b/Library/Formula/gpg-agent.rb
@@ -1,9 +1,9 @@
 class GpgAgent < Formula
   desc "GPG key agent"
   homepage "https://www.gnupg.org/"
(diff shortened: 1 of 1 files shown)


DomT4:
I think we should be fine without a revision bump. 2.0.29 was mostly bug fixes, so 2.0.28 & 2.0.29 should be broadly compatible. Thanks for catching this missing update!

DomT4:
Yosemite CI threw a strop. Have re-queued this.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/e180904d82adcafba6370e35bc3d5da44ac80ae8. Thank you @chdiza!

PR #43935, opened 2015-09-14
add `bottle :disable` DSL

xu-cheng:
cc @Homebrew/owners

### DIFF
--- a/Library/Homebrew/cmd/bottle.rb
+++ b/Library/Homebrew/cmd/bottle.rb
@@ -145,6 +145,12 @@ module Homebrew
       return ofail "Formula not installed or up-to-date: #{f.full_name}"
     end
 
+    if f.bottle_disabled?
+      ofail "Formula has disabled bottle: #{f.full_name}"
+      puts f.bottle_disable_reason
+      return
+    end
+
     unless built_as_bottle? f
       return ofail "Formula not installed with '--build-bottle': #{f.full_name}"
     end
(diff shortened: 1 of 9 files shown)


mikemcquaid:
Can we go for `:unneeded` perhaps? We could maybe add a `:disabled:` too for the case where the bottling process is known to be broken for that package.

mikemcquaid:
Code changes look good :+1:

mikemcquaid:
Also perhaps having the second argument be a reason wouldn't be a bad thing.

xu-cheng:
> Can we go for :unneeded perhaps? 

Done.

> Also perhaps having the second argument be a reason wouldn't be a bad thing.

I'm thinking `bottle :unneeded` strictly means no compile is required, so we can use it for no Xcode/CLT system. We can consider add reason for `bottle :disable`.

mikemcquaid:
> I'm thinking bottle :unneeded strictly means no compile is required, so we can use it for no Xcode/CLT system. We can consider add reason for bottle :disable.

Yep, that sounds good :+1:

mikemcquaid:
`:disable` works for me :+1:

xu-cheng:
I will work on this maybe this weekend.

xu-cheng:
Updated. The DSL will be looked like these:
```ruby
bottle :unneeded # doesn't need compiling, useful for non-Xcode/CLT install.

bottle :disable, "the reason"
```

@mikemcquaid, could you help to double check that this PR won't break the bottle hooks used by Boxen?

mikemcquaid:
Looks good to me :+1:

PR #40773, opened 2015-06-16
add payara formula

nagaseyasuhito:
Payara is Java EE Application Server forked from GlassFish.
It just replaced name and url and sha1 from glassfish.rb fomula.

thanks!

PR #43950, opened 2015-09-14
Updated glassfish and added payara

trajano:
Glassfish and Payara now install something in `bin`

Added note that Glassfish will not run without root priviledges from
asadmin.  Payara does not have the same restriction.

Supercedes PR #40773

### DIFF
--- a/Library/Formula/glassfish.rb
+++ b/Library/Formula/glassfish.rb
@@ -1,26 +1,61 @@
-require "formula"
-
 class Glassfish < Formula
   desc "Java EE application server"
   homepage "https://glassfish.java.net"
   url "http://download.java.net/glassfish/4.1.1/release/glassfish-4.1.1.zip"
   sha256 "1b20d42b7d97d0282b90b3f6ec958138f1e81a500468f391ff5aa06afb599b9c"
 
+  depends_on :java => "1.7+"
+
+  conflicts_with "payara", :because => "Both install `asadmin` binaries"
+
   def install
     rm_rf Dir["bin/*.bat"]
+    rm "glassfish/config/asenv.bat"
+
+    # Change to point to alternate domains and nodes directory
+    inreplace "glassfish/config/asenv.conf" do |s|
+      s.gsub! %r{AS_DEF_DOMAINS_PATH="../domains"}, "AS_DEF_DOMAINS_PATH=#{var}/glassfish/domains"
+      s.gsub! %r{AS_DEF_NODES_PATH="../nodes"}, "AS_DEF_NODES_PATH=#{var}/glassfish/nodes"
+    end
+
     libexec.install Dir["*", ".org.opensolaris,pkg"]
+    (bin/"asadmin").write <<-EOS.undent
+      #!/bin/sh
+      AS_INSTALL="#{libexec}/glassfish"
+      AS_INSTALL_LIB="$AS_INSTALL/lib"
+      . "${AS_INSTALL}/config/asenv.conf"
+      JAVA=java
+      #Depends upon Java from ../config/asenv.conf
+      if [ ${AS_JAVA} ]; then
+        JAVA=${AS_JAVA}/bin/java
+      fi
+      exec "$JAVA" -jar "$AS_INSTALL_LIB/client/appserver-cli.jar" "$@"
+    EOS
+  end
+
+  def post_install
+    (var/"glassfish/nodes").mkpath
+    # Make a copy of the domains folder if they do not exist yet
+    unless (var/"glassfish/domains").exist?
+      mv libexec/"glassfish/domains", "#{var}/glassfish"
+    end
   end
 
   def caveats; <<-EOS.undent
     The home of GlassFish Application Server 4 is:
       #{opt_libexec}
 
-    You may want to add the following to your .bash_profile:
-      export GLASSFISH_HOME=#{opt_libexec}
-      export PATH=${PATH}:${GLASSFISH_HOME}/bin
+    Starting up GlassFish requires root privileges as per
+    https://java.net/jira/browse/GLASSFISH-21343.  The work around is
+    to use
+      sudo asadmin start-domain
 
     Note: The support scripts used by GlassFish Application Server 4
     are *NOT* linked to bin.
   EOS
+  end
+
+  test do
+    system bin/"asadmin", "--help"
   end
 end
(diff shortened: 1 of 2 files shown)


Library/Formula/glassfish.rb
```diff
@@ -6,21 +6,41 @@ class Glassfish < Formula
   url "http://download.java.net/glassfish/4.1/release/glassfish-4.1.zip"
   sha256 "3edc5fc72b8be241a53eae83c22f274479d70e15bdfba7ba2302da5260f23e9d"
 
+  depends_on :java
```
DomT4: Any particular version?

Library/Formula/glassfish.rb
```diff
@@ -6,21 +6,41 @@ class Glassfish < Formula
...
+  conflicts_with "payara", :because => "Both install `asadmin` binaries"
```
DomT4: Please leave an empty line between `depends_on` and `conflicts_with`, Thanks!

Library/Formula/glassfish.rb
```diff
@@ -6,21 +6,41 @@ class Glassfish < Formula
...
+    to use
+      sudo asadmin start-domain
 
     Note: The support scripts used by GlassFish Application Server 4
     are *NOT* linked to bin.
   EOS
   end
+
+  test do
+    system "asadmin", "--help"
```
DomT4: This needs to be a direct path, i.e. `system bin/"asadmin", "--help"`

Library/Formula/payara.rb
```diff
@@ -0,0 +1,40 @@
+class Payara < Formula
+  desc "Payara is 24/7 software support for GlassFish Server Open Source Edition"
+  homepage "http://www.payara.co.uk/"
+  url "https://github.com/payara/Payara/releases/download/payara-server-4.1.153/payara-4.1.153.zip"
+  sha256 "52736b251b424280b1a4f19dd3522eb1fd1be12d0e2020a074c4872b898ef55c"
+
+  depends_on :java
```
DomT4: Same comments as above for both these lines.

Library/Formula/payara.rb
```diff
@@ -0,0 +1,40 @@
...
+    The home of Payara Application Server is:
+      #{opt_libexec}
+
+    Note: The support scripts used by Payara Application Server
+    are *NOT* linked to bin.
+  EOS
+  end
+
+  test do
+    system "asadmin", "start-domain"
```
DomT4: And here, per the above.

Library/Formula/glassfish.rb
trajano: How can you define the version for Java?

trajano: It should be 7 or 8

Library/Formula/glassfish.rb
trajano: For future reference, where are these extra guidelines specified?  `brew audit` does not complain.

Library/Formula/glassfish.rb
dunn: `depends_on :java => "1.7+"`

trajano:
Is there a more detailed guide that I have to follow for future reference? I used `brew audit` to verify the new formula I write and tried to follow whatever I can glean from https://github.com/Homebrew/homebrew/blob/master/CONTRIBUTING.md and existing formulae.

I am planning to write another for WebSphere Liberty

trajano:
In addition, is there some formula that can give me an example for supporting "different editions" for example glassfish 4 has full profile and web profile, payara has a bit more.

Library/Formula/glassfish.rb
DomT4: `brew audit` doesn't complain here because there is sometimes a reason not to call `bin`, `libexec` or `sbin` on the command, like pipes. There's some API documentation here: http://www.rubydoc.info/github/Homebrew/homebrew/Formula.test

DomT4:
API documentation (roughly) is here: http://www.rubydoc.info/github/Homebrew/homebrew/Formula.test
Formula Cookbook is here: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md

trajano:
Methinks http://bot.brew.sh/job/Homebrew%20Pull%20Requests/33213/ is stuck

DomT4:
Yes, It looks like it was. Looking into it. Thanks.

trajano:
Thanks, looks like the next one is also stuck.  http://bot.brew.sh/job/Homebrew%20Pull%20Requests/33214/ Perhaps it's out of disk space?

DomT4:
Expect stuckness for a bit. We're working on it.

trajano:
For payara testing, I make sure that the domains that are provided work correctly.  However I got 

      Exception in thread "main" java.lang.RuntimeException: the domain directory is not writable.

Is there something I need to change on the test?  I can also switch it to a more simplistic test like what was done on glassfish which is just to see if help shows up.

DomT4:
It is probably trying to write to a directory we don't permit in testing. You can check locally with `brew test --verbose --sandbox payara`

trajano:
It seems the tests do not allow writing to `/usr/local/var`

trajano:
I have removed the more elaborate test and replaced it with a simpler one.

trajano:
Is this still waiting on anything?

DomT4:
@BrewTestBot test this please

Library/Formula/glassfish.rb
```diff
@@ -1,26 +1,61 @@
...
+
+  conflicts_with "payara", :because => "Both install `asadmin` binaries"
+
   def install
     rm_rf Dir["bin/*.bat"]
+    rm "glassfish/config/asenv.bat"
+
+    # Change to point to alternate domains and nodes directory
+    inreplace "glassfish/config/asenv.conf" do |s|
+      s.gsub! %r{AS_DEF_DOMAINS_PATH="../domains"}, "AS_DEF_DOMAINS_PATH=#{var}/payara/domains"
```
DomT4: Why are these `#{var}/payara` instead of `#{var}/glassfish` in the `glassfish` formula, out of interest?

trajano: Because it's wrong.  I'll fix it right away.

bfontaine:

apjanke:
The test failure was just the CI server hitting the GitHub rate limit. If you can rebase this on master, we should be good to go here.

bfontaine:
Ping?

trajano:
I no longer have access to a Mac at the moment. Someone else needs to take
this task over.

mikemcquaid:
People cannot take over PRs unfortunately.

PR #43951, opened 2015-09-14
launchdns 1.0.3

josh:
Updates launchdns formula with a smoke test to ensure `launch.h` support was compiled correctly. Should be a good test to ensure any automated binaries work.

-
To: @mikemcquaid

### DIFF
--- a/Library/Formula/launchdns.rb
+++ b/Library/Formula/launchdns.rb
@@ -1,9 +1,9 @@
 class Launchdns < Formula
   desc "Mini DNS server designed soely to route queries to localhost"
   homepage "https://github.com/josh/launchdns"
-  url "https://github.com/josh/launchdns/archive/v1.0.1.tar.gz"
+  url "https://github.com/josh/launchdns/archive/v1.0.2.tar.gz"
   head "https://github.com/josh/launchdns.git"
-  sha256 "e96d1b92819a294f1e325df629ae4bf202fd137b8504cf4ddd00cda7e47f7099"
+  sha256 "a67d54096055532af25f6dc9319c0dad6da87658d4d4a2d636ab015b393f011f"
 
   bottle do
     cellar :any
@@ -21,6 +21,7 @@ class Launchdns < Formula
   end
 
   test do
+    system "#{bin}/launchdns --version | grep -vq 'without socket activation'"
     system "#{bin}/launchdns", "-p0", "-t1"
   end
 
--- a/Library/Formula/launchdns.rb
+++ b/Library/Formula/launchdns.rb
@@ -21,7 +21,7 @@ class Launchdns < Formula
   end
 
   test do
-    system "#{bin}/launchdns --version | grep -vq 'without socket activation'"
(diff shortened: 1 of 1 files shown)


Library/Formula/launchdns.rb
```diff
@@ -21,6 +21,7 @@ def install
   end
 
   test do
+    system "#{bin}/launchdns --version | grep -vq 'without socket activation'"
```
dunn: How likely is it that this new test fails while the one below passes?

mikemcquaid: This is testing specific functionality so :+1:

bfontaine: Can’t we use something like `assert shell_output("#{bin}/launchdns --version") !~ /without socket activation/`?

mikemcquaid: Yeh, we can.

mikemcquaid: But `assert_not_match` or something may be nicer.

mikemcquaid:
@josh @dunn @bfontaine Actually, I've submitted two PRs upstream that are probably preferable that will allow us to make `configure` error-out and ensure that the include paths are found when using Xcode without the CLT: https://github.com/josh/launchdns/pull/5 https://github.com/josh/launchdns/pull/6

Library/Formula/launchdns.rb
josh: Rad, I didn't realize unit test helpers existed in this scope!

josh:
Updated to launchdns 1.0.3 with @mikemcquaid's upstream changes.

Library/Formula/launchdns.rb
```diff
@@ -15,12 +15,14 @@ class Launchdns < Formula
...
   def install
     ENV["PREFIX"] = prefix
+    system "./configure", "--with-launch-h", "--with-launch-h-activate-socket"
     system "make", "install"
 
     (prefix+"etc/resolver/dev").write("nameserver 127.0.0.1\nport 55353\n")
   end
 
   test do
+    assert_not_match /without socket activation/, shell_output("#{bin}/launchdns --version")
```
xu-cheng: Unfortunately `assert_not_match` is not available in Ruby 1.8. You can use `assert_no_match` instead.

mikemcquaid:
Nice, those errors are proving useful. @josh is it expected that that won't be found in the header on 10.8/10.9?

josh:
>  is it expected that that won't be found in the header on 10.8/10.9?

Yeah, they don't support the activate socket api.

Should we maybe consider marking this Yosemite and higher?

mikemcquaid:
@josh If you consider this not useful without the socket API: :+1:

josh:
You can still use it without those APIs, it just has to use a different plist format. I'm indifferent on supporting it.

Just let me know what to do to fix the tests in either case.

mikemcquaid:
@josh Add a `depends_on :macos => :yosemite`

josh:
Done. Thanks!

PR #43963, opened 2015-09-15
avro-cpp: patch for Boost 1.59

tdsmith:


tdsmith:
unblocks #42960 by applying https://github.com/apache/avro/pull/51

PR #43964, opened 2015-09-15
test-bot: delete local ref

xu-cheng:
cc @mikemcquaid


xu-cheng:
I guess this is useless.

PR #43972, opened 2015-09-15
Watchman 3.8.0

wez:
Tested via:

```
brew install watchman
brew audit watchman
brew test watchman
```

### DIFF
--- a/Library/Formula/watchman.rb
+++ b/Library/Formula/watchman.rb
@@ -1,10 +1,8 @@
 class Watchman < Formula
   desc "Watch files and take action when they change"
   homepage "https://github.com/facebook/watchman"
-  url "https://github.com/facebook/watchman/archive/v3.7.0-brew.tar.gz"
-  version "3.7.0"
-  sha256 "55a3ea1ee3990a5e5c11f1a37ad5bafbb63e8002f48f449c083e598f6f332154"
-
+  url "https://github.com/facebook/watchman/archive/v3.8.0.tar.gz"
+  sha256 "10d1e134e6ff110044629a517e7c69050fb9e4a26f21079b267989119987b40d"
   head "https://github.com/facebook/watchman.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


wez:
Any idea what happened on the mountain lion run?

mikemcquaid:
@BrewTestBot test this please

mikemcquaid:
@wez Sorry, we've been having some CI issues. I'm rerunning this

DomT4:
It's trying to break out of the sandbox we now run on our CI. Can you take a look at this, Thanks!

```
==> /usr/local/Cellar/watchman/3.8.0/bin/watchman shutdown-server
2015-09-18T14:46:29,537: [1939804928] failed to create /usr/local/Cellar/watchman/3.8.0/var/run/watchman/brew-state: Operation not permitted
Error: watchman: failed
Failed executing: /usr/local/Cellar/watchman/3.8.0/bin/watchman shutdown-server
```

wez:
How does the sandbox work?  Watchman launches itself via launchd.  If we can't use that in the sandbox, we'll have to remove the test portion of the formula.  I'm fine with that; we have CI that runs our integration tests on multiple platforms before we deem a revision good enough to push downstream from us.

DomT4:
It uses OS X's sandbox utility to ensure no process inside `test do` can *write* outside of the `testpath`. In this case it's trying to write a file outside of the `testpath` and being denied permission to do so. It fails earlier if the `plist` isn't already installed.

As you explain `watchman` creates the `plist` at runtime, which ties our hands here a bit. For various reasons I don't particularly want to set the precedent of allowing our `test do` blocks to write permanent `plists` on the system and set them running.

DomT4:
We may have to scale the test back here a bit, I guess. I'd like to keep something around, even a version check if necessary.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/86acff193799fca66c2f5ffc80195cab29fc9e15. Thank you @wez!

PR #43976, opened 2015-09-15
[WIP] Add Travis CI configuration.

mikemcquaid:
Just testing Travis for our main PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,23 @@
+# Force OS X.
+language: objective-c
+os: osx
+env: OSX=10.10
+osx_image: xcode7
+rvm:
+  - system
+matrix:
+  include:
+    - env: OSX=10.9
+      osx_image: beta-xcode6.2
+      rvm: system
+
+before_install:
+  - sudo rsync -az "$TRAVIS_BUILD_DIR/" /usr/local/
+  - export TRAVIS_BUILD_DIR="/usr/local"
+  - cd /usr/local
+  - if [ -f ".git/shallow" ]; then git fetch --unshallow; fi
+  - git reset --hard $TRAVIS_COMMIT
+  - git clean -qxdff
+
+script:
+  - brew test-bot
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -17,6 +17,7 @@
 # --tap=<tap>:     Use the git repository of the given tap
 # --dry-run:       Just print commands, don't run them.
 # --fail-fast:     Immediately exit on a failing step.
+# --verbose:       Print out all logs in realtime
 #
 # --ci-master:           Shortcut for Homebrew master branch CI options.
 # --ci-pr:               Shortcut for Homebrew pull request CI options.
@@ -91,13 +92,31 @@ module Homebrew
 
     def puts_command
       cmd = @command.join(" ")
-      print "#{Tty.blue}==>#{Tty.white} #{cmd}#{Tty.reset}"
+      cmd_line = "#{Tty.blue}==>#{Tty.white} #{cmd}#{Tty.reset}"
+      if ENV["TRAVIS"]
+        @travis_timer_id = rand(2**32).to_s(16)
+        puts "travis_fold:start:#{@command.join(".")}"
+        puts "travis_time:start:#{@travis_timer_id}"
+        puts cmd_line
+        return
+      end
+      print cmd_line
       tabs = (80 - "PASSED".length + 1 - cmd.length) / 8
       tabs.times { print "\t" }
       $stdout.flush
     end
 
     def puts_result
+      if ENV["TRAVIS"]
+        cmd = @command.join(" ")
+        puts "#{Tty.send status_colour}==> #{cmd}: #{status_upcase}#{Tty.reset}"
+        travis_start_time = @start_time.to_i*1000000000
+        travis_end_time = @end_time.to_i*1000000000
+        travis_duration = travis_end_time - travis_start_time
+        puts "travis_fold:end:#{@command.join(".")}"
+        puts "travis_time:end:#{@travis_timer_id},start=#{travis_start_time},finish=#{travis_end_time},duration=#{travis_duration}"
+        return
+      end
       puts " #{Tty.send status_colour}#{status_upcase}#{Tty.reset}"
     end
 
@@ -105,19 +124,25 @@ module Homebrew
       @output && !@output.empty?
     end
 
+    def time
+      @end_time - @start_time
+    end
+
     def run
+      @start_time = Time.now
+
       puts_command
       if ARGV.include? "--dry-run"
-        puts
+        @end_time = Time.now
         @status = :passed
+        puts_result
         return
       end
 
       verbose = ARGV.verbose?
-      puts if verbose
+      puts if verbose && !ENV["TRAVIS"]
       @output = ""
       working_dir = Pathname.new(@command.first == "git" ? @repository : Dir.pwd)
-      start_time = Time.now
       read, write = IO.pipe
 
       begin
@@ -138,7 +163,7 @@ module Homebrew
       end
 
       Process.wait(pid)
-      @time = Time.now - start_time
+      @end_time = Time.now
       @status = $?.success? ? :passed : :failed
       puts_result
 
@@ -269,12 +294,17 @@ module Homebrew
       @category = __method__
       @start_branch = current_branch
 
+      travis_pr = ENV["TRAVIS_PULL_REQUEST"] && ENV["TRAVIS_PULL_REQUEST"] != "false"
+
       # Use Jenkins environment variables if present.
       if no_args? && ENV["GIT_PREVIOUS_COMMIT"] && ENV["GIT_COMMIT"] \
          && !ENV["ghprbPullLink"]
         diff_start_sha1 = shorten_revision ENV["GIT_PREVIOUS_COMMIT"]
         diff_end_sha1 = shorten_revision ENV["GIT_COMMIT"]
         brew_update
+      elsif ENV["TRAVIS_COMMIT_RANGE"]
+        diff_start_sha1, diff_end_sha1 = ENV["TRAVIS_COMMIT_RANGE"].split "..."
+        diff_end_sha1 = ENV["TRAVIS_COMMIT"] if travis_pr
       elsif @hash
         diff_start_sha1 = current_sha1
         brew_update
@@ -286,6 +316,9 @@ module Homebrew
       # Handle Jenkins pull request builder plugin.
       if ENV["ghprbPullLink"]
         @url = ENV["ghprbPullLink"]
+        @hash = nil
+      elsif travis_pr
+        @url = "https://github.com/#{ENV["TRAVIS_REPO_SLUG"]}/pull/#{ENV["TRAVIS_PULL_REQUEST"]}"
         @hash = nil
       end
 
@@ -301,6 +334,9 @@ module Homebrew
         diff_start_sha1 = "#{@hash}^"
         diff_end_sha1 = @hash
         @name = @hash
+      elsif ENV["TRAVIS_PULL_REQUEST"] && ENV["TRAVIS_PULL_REQUEST"] != "false"
+        @short_url = @url.gsub("https://github.com/", "")
+        @name = "#{@short_url}-#{diff_end_sha1}"
       elsif @url
         diff_start_sha1 = current_sha1
         test "git", "checkout", diff_start_sha1
@@ -323,7 +359,7 @@ module Homebrew
       FileUtils.mkdir_p @log_root
 
       return unless diff_start_sha1 != diff_end_sha1
-      return if @url && !steps.last.passed?
+      return if @url && steps.last && !steps.last.passed?
 
       if @tap
         formula_path = %w[Formula HomebrewFormula].find { |dir| (@repository/dir).directory? } || ""
@@ -369,7 +405,7 @@ module Homebrew
     def setup
       @category = __method__
       return if ARGV.include? "--skip-setup"
-      test "brew", "doctor"
+      test "brew", "doctor" unless ENV["TRAVIS"]
       test "brew", "--env"
       test "brew", "config"
     end
@@ -610,7 +646,8 @@ module Homebrew
 
       checkout_args << @start_branch
 
-      if ARGV.include?("--cleanup") || @url || @hash
+      if @start_branch && !@start_branch.empty? && \
+         (ARGV.include?("--cleanup") || @url || @hash)
         test "git", "checkout", *checkout_args
       end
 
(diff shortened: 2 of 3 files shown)


xu-cheng:
Travis CI has its own copy of Homebrew. And I think it would be best if we can run test on the clone copy.  At the other hand, we need to test formula with prefix of `/user/local`.  Therefore, can or should we do something like this:
```
rm /usr/local/bin/brew
ln -sf $PWD/bin/brew /usr/local/bin/brew
mkdir /usr/local/Cellar
```

mikemcquaid:
@xu-cheng Yep, need to work that out. I'm wondering if there's a way to coax Travis into checking out into a certain directory.

xu-cheng:
Another thing I jus found is Travis CI supports custom log folding. We can use it to fold log inside brew test-bot. Refs:
* https://github.com/travis-ci/travis-ci/issues/2285#issuecomment-42724719
* https://github.com/macbre/travis-fold

mikemcquaid:
@xu-cheng Great idea.

.travis.yml
```diff
@@ -0,0 +1,19 @@
+# Force OS X.
+language: objective-c
+os: osx
+osx_image: xcode7 # 10.10
+matrix:
+  include:
+    - osx_image: beta-xcode6.2 # 10.9
+
+rvm:
+  - rvm use system
```
xu-cheng: This should be just:
```
rvm: 
   - system
```

.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+
+install:
+  - rvm use system
+  - cd /usr/local; git fetch --unshallow
+  - cd /usr/local; [ "$TRAVIS_PULL_REQUEST" != "false" ] && git fetch origin +refs/pull/$TRAVIS_PULL_REQUEST/merge
+  - cd /usr/local; git reset --hard $TRAVIS_COMMIT
```
xu-cheng: Personally, I think the symlink solution(like below) is better than git manipulation. It's easier, more robust and with better performance.

```
install:
   - rm /usr/local/bin/brew
   - ln -sf $PWD/bin/brew /usr/local/bin/brew
   - mkdir -p /usr/local/Cellar
```

It replaces `/usr/local/bin/brew` to a symlink to the brew file in the test repo. Doing so, Homebrew will use travis's repo as library while preserving  `/usr/local/` as prefix.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -91,6 +92,10 @@ def failed?
 
     def puts_command
       cmd = @command.join(" ")
+      if ENV["TRAVIS"]
+        puts cmd
+        puts "travis_fold:start:#{@command.join(".")}"
```
xu-cheng: We should add a `return` here.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -98,6 +103,10 @@ def puts_command
     end
 
     def puts_result
+      if ENV["TRAVIS"]
+        puts "travis_fold:end:#{@command.join(".")}"
+        return
```
xu-cheng: The `return` here is unnecessary and it will prevent the status result being printed.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -287,6 +298,9 @@ def brew_update
       if ENV["ghprbPullLink"]
         @url = ENV["ghprbPullLink"]
         @hash = nil
+      elsif ENV["TRAVIS_PULL_REQUEST"] && ENV["TRAVIS_PULL_REQUEST"] != "false"
+        @url = "https://github.com/#{ENV["TRAVIS_REPO_SLUG"]}/pull/#{ENV["TRAVIS_PULL_REQUEST"]}"
+        @hash = nil
```
xu-cheng: Do you intent to use Travis CI to test PR? I think it won't help much as we cannot use it generate bottles. Anyway your call.

mikemcquaid: Why not?

.travis.yml
mikemcquaid: Yep.

.travis.yml
mikemcquaid: I don't agree. If we do anything it should be to symlink the `.git` directory. The symlinked `brew` setup is not what most of our users do.

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: That was intentional.

mikemcquaid: (but changing now)

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -91,6 +92,10 @@ def failed?
 
     def puts_command
       cmd = @command.join(" ")
+      if ENV["TRAVIS"]
+        puts cmd
```
xu-cheng: Looks like we don't need `puts cmd` here. https://travis-ci.org/Homebrew/homebrew/jobs/80807853

Library/Homebrew/cmd/test-bot.rb
xu-cheng: Because we cannot upload the bottle files. [Travis doesn't support dependent build yet](https://github.com/travis-ci/travis-ci/issues/249).

mikemcquaid: We don't need a dependent build, the upload can be done in the job: http://docs.travis-ci.com/user/deployment/bintray/

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -91,6 +92,13 @@ def failed?
...
+        @travis_timer_id = rand(2**32).to_s(16)
+        puts "travis_time:start:$travis_timer_id:#{@travis_timer_id}"
+        puts "travis_fold:start:#{@command.join(".")}"
+        return
+      end
       print "#{Tty.blue}==>#{Tty.white} #{cmd}#{Tty.reset}"
```
xu-cheng: I think we should use `print "#{Tty.blue}==>#{Tty.white} #{cmd}#{Tty.reset}"` which will have nicer output instead of `puts cmd`

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -91,6 +92,13 @@ def failed?
 
     def puts_command
       cmd = @command.join(" ")
+      if ENV["TRAVIS"]
+        puts cmd
+        @travis_timer_id = rand(2**32).to_s(16)
```
xu-cheng: Should we use the checksum of commands instead of random value which may reproduce the same value.

Library/Homebrew/cmd/test-bot.rb
xu-cheng: We DO a dependent build. Because for security reason, tokens cannot be used in PR. http://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: I don't think so, this is doing the same thing the official Travis script does.

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: This is only used for the travis inline command so I think it's more consistent as-is. At least the `==>` looked pretty weird.

Library/Homebrew/cmd/test-bot.rb
xu-cheng: :cool:

Library/Homebrew/cmd/test-bot.rb
xu-cheng: I don't think `==>` looks weird. Currently it output strange empty line. https://travis-ci.org/Homebrew/homebrew/jobs/80815392.
While the old one https://travis-ci.org/Homebrew/homebrew/jobs/80807853 looks much better.

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: That's only for secure environment variables. We can still do deploys from PRs, we'll just have to be more clever with how we use the environment variables e.g. restrict them only to certain command steps rather than using Travis's encryption feature.

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: Still a work in progress.

.travis.yml
xu-cheng: I'm OK to symlink `.git` as well. My thought is we should reduce necessary operations, so it would be more robust (less likely break due to git error) and with better performance(less IO fetch).

Library/Homebrew/cmd/test-bot.rb
xu-cheng: >  we'll just have to be more clever with how we use the environment variables e.g. restrict them only to certain command steps rather than using Travis's encryption feature.

Sadly, that's impossible. You should note the whole codebase + `.travis.yml` is writable during PR.

Library/Homebrew/cmd/test-bot.rb
xu-cheng: :bow:

Library/Homebrew/cmd/test-bot.rb
xu-cheng: To secure the token, you need to store them somewhere else and make sure the code is read-only. Hence, we need dependent build.

mikemcquaid: I'm sure we can work something out here. I certainly don't think it's impossible, even if we only provide access to a staging area that we verify manually before moving.

mikemcquaid: Or alternatively we have the actual bottle deployment happen on `master` builds only, somehow.

mikemcquaid:
@xu-cheng Looks like this is green now. Thoughts on merging?

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -91,13 +92,31 @@ def failed?
...
       cmd = @command.join(" ")
-      print "#{Tty.blue}==>#{Tty.white} #{cmd}#{Tty.reset}"
+      cmd_line = "#{Tty.blue}==>#{Tty.white} #{cmd}#{Tty.reset}"
+      if ENV["TRAVIS"]
+        @travis_timer_id = rand(2**32).to_s(16)
+        puts "travis_fold:start:#{@command.join(".")}"
+        puts "travis_time:start:#{@travis_timer_id}"
+        puts cmd_line
+        return
+      end
```
xu-cheng: A style nit, I would prefer `if else` then `if return`.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -91,13 +92,31 @@ def failed?
...
+      print cmd_line
       tabs = (80 - "PASSED".length + 1 - cmd.length) / 8
       tabs.times { print "\t" }
       $stdout.flush
     end
 
     def puts_result
+      if ENV["TRAVIS"]
+        cmd = @command.join(" ")
+        puts "#{Tty.send status_colour}==> #{cmd}: #{status_upcase}#{Tty.reset}"
```
xu-cheng: Shouldn't we print result outside of fold?

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -91,13 +92,31 @@ def failed?
...
+        travis_start_time = @start_time.to_i*1000000000
+        travis_end_time = @end_time.to_i*1000000000
+        travis_duration = travis_end_time - travis_start_time
+        puts "travis_fold:end:#{@command.join(".")}"
+        puts "travis_time:end:#{@travis_timer_id},start=#{travis_start_time},finish=#{travis_end_time},duration=#{travis_duration}"
+        return
```
xu-cheng: Again, can we use `if else end` here?

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+rvm:
+  - system
+matrix:
+  include:
+    - env: OSX=10.9
+      osx_image: beta-xcode6.2
+      rvm: system
+
+before_install:
+  - sudo rsync -az "$TRAVIS_BUILD_DIR/" /usr/local/
```
xu-cheng: I wonder will `sudo` cause any permission problem?

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+  - export TRAVIS_BUILD_DIR="/usr/local"
```
xu-cheng: Where will this variable be used?

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+  - cd /usr/local
+  - if [ -f ".git/shallow" ]; then git fetch --unshallow; fi
```
xu-cheng: Why do we need to unshallow, Travis itself will fetch with depth in 50 which I think should be enough.

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: `puts` vs `print` here.

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: I prefer early `return` :wink:

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: I prefer early `return` :wink:

.travis.yml
mikemcquaid: The opposite; ensures permissions are the same.

.travis.yml
mikemcquaid: It might not be, just was recommended to be by the Travis CI support person. I suspect it's defining stuff for future build steps?

.travis.yml
mikemcquaid: Bottle revisions.

Library/Homebrew/cmd/test-bot.rb
xu-cheng: I usually use `early return` in nested or complex conditions. To me, using it in simple binary conditions makes the code inbalanace or asymmetric. Anyway, your call.

Library/Homebrew/cmd/test-bot.rb
xu-cheng: However, travis still hides the result. See https://travis-ci.org/Homebrew/homebrew/jobs/80865031

.travis.yml
xu-cheng: :+1:

.travis.yml
xu-cheng: Oh, I almost forget this part. But if we just use Travis to test master branch for the time being, we can remove this to save some time.

.travis.yml
xu-cheng: :cool:

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: Ah, I see what you mean. Yeh, good idea.

.travis.yml
mikemcquaid: Yeh, we can always remove it if necessary (or have it done just by `test-bot` for the bottle job), maybe. Will leave it as-is for now

PR #43991, opened 2015-09-16
pincaster: add openssl dependency

apjanke:
Fixes a build failure on 10.11 where `pincaster` fails to find OpenSSL, like several other formula.

When I first made this fix and tested it, it would then fail due to not being able to find the YAJL library. Though when I tested it today (on a machine without YAJL installed), it succeeded. Something about my other test box, which I had previously used to do the "try installing everything", may have had some residue that caused it to spuriously detect YAJL's presence.


DomT4:
See #43945. Just waiting for the CI to get done with its backlog & then I'll deal with the test failure there.

PR #44005, opened 2015-09-16
update-test: use git clone --local instead of copy

xu-cheng:
The .git directory in BrewTestBot is huge(1GB) which results to poor
performance of `brew update-test`. Therefore, let's use `git clone
--local` here which will create hard symlink to boost performance.


mikemcquaid:
:+1:

PR #44027, opened 2015-09-16
added customsshd daemon

xfreebird:
It was created to allow running commands through ssh in user UI session.
It is started by launchd when user logs in.
Useful for running iOS simulator tests when connected through ssh.
The default SSHD daemon does not have access the user UI session, hence failing the tests.


mikemcquaid:
This is a bit too niche and new for us for now. I'd suggest running your own tap. Sorry!

xfreebird:
No problem, already did. 
Thanks !

PR #44037, opened 2015-09-16
ratfor (new formula)

FranklinChen:


Library/Formula/ratfor.rb
```diff
@@ -0,0 +1,15 @@
+class Ratfor < Formula
+  version "1.02"
+  desc "Rational Fortran"
+  homepage "http://www.dgate.org/ratfor/"
+  url "http://www.dgate.org/ratfor/ratfor.tar.gz"
```
DomT4: We don't really use static URLs in the core. It needs to be a versioned link that won't silently break on us.

Library/Formula/ratfor.rb
```diff
@@ -0,0 +1,15 @@
...
+  sha256 "daf2971df48c3b3908c6788c4c6b3cdfb4eaad21ec819eee70a060956736ea1c"
+
+  depends_on "gcc"
```
DomT4: For build?

Library/Formula/ratfor.rb
FranklinChen: It's the only URL provided.

Library/Formula/ratfor.rb
FranklinChen: No, for use of the Fortran code generated. I guess that should be noted rather than a requirement for installation.

FranklinChen: "make check" uses gfortran from gcc.

Library/Formula/ratfor.rb
DomT4: You'll need to ask upstream for a versioned one, or host a tap for this yourself. Static URLs are bad UX, sadly.

Library/Formula/ratfor.rb
DomT4: `depends_on :fortran` then?

dunn:
A few audit failures:
```
==> audit problems
ratfor:
 * `url` (line 5) should be put before `version` (line 2)
 * Stable: version 1.02 is redundant with version scanned from URL
```
Upshot is you can remove the `version` line entirely, and we'd like a test that checks basic functionality.

FranklinChen:
How do I move "make check" from the install phase to the test block?

dunn:
Afraid that won't work; we run the test in a separate temporary directory so it doesn't have access to the source code.  Is there a command that can be run non-interactively and won't take too long?  CMake is a good example: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/cmake.rb#L125-L128

DomT4:
Closing in favour of #44358.

PR #44046, opened 2015-09-16
whirr: fix download URL

apjanke:
The old URL is downloading an interactive HTML page instead of the tarball and `brew install` is getting an SHA mismatch.


bfontaine:

DomT4:
Apache have killed the project off [apparently](http://attic.apache.org/projects/whirr.html). Will merge this though, we can support it for as long as it keeps working.

Merged in https://github.com/Homebrew/homebrew/commit/596b9343c1cc9918583e539d9c66f005245387f1. Thanks @apjanke!

PR #44054, opened 2015-09-16
Add librados and rados utility from Ceph

bertjwregeer:
This adds a new formula that builds from head (there is no stable version that supports OS X yet) the librados library as well as the rados utility. This allows development of Ceph applications using OS X.

Unfortunately it installs a python package unconditionally (ceph-detect-init), and there is no way to disable it, so I am forcing it to depend on the Homebrew python because otherwise it will attempt to install it in the wrong location (specifically, the OS site-packages).


bertjwregeer:
Sorry, just noticed that this did not do what I expected it to do :-(.

PR #44055, opened 2015-09-17
tig: install system tigrc as sample

### LINKED ISSUE #43958: Tig: etc/tigrc not updated during upgrade from 2.0.2 to 2.1.1
Since [Tig version 2.0](https://github.com/jonas/tig/blob/master/NEWS.adoc#tig-20), the system-wide `tigrc` file is where all default options are defined. When packaged files destined for `/usr/local/etc` are not automatically updated when a package is upgraded to a new version, it leads to broken and unexpected behaviour as reported in jonas/tig#407 where a default setting is missing when running Tig version 2.1.1 because it was added in Tig version 2.0.3 and thus not in the `tigrc` install for Tig version 2.0.2.

My proposed fix is to package Tig without a `/usr/local/etc/tigrc` so that it falls back to use the internal (and always updated) `tigrc` copy. Alternatively, a `tigrc.sample` can be included in either `/usr/local/etc` or `/usr/local/Cellar/tig/x.x.x/etc` allowing a system admin to rename if system-wide options need to be overridden. If this is acceptable I suggest that I first add a `--disable-tigrc-install` option to Tig's `configure` script after which upgrading the formulae should be trivial.

jonas:
Package tig without installing the default system-wide tigrc to
/usr/local/etc so that it falls back to use the internal (and always
updated) tigrc copy. This avoids problems when new options are added as
seen in jonas/tig#407.

As a caveat, users are offered the option to install the system-wide
tigrc from a tigrc sample.

Fixes #43958
Fixes jonas/tig#407

### DIFF
--- a/Library/Formula/tig.rb
+++ b/Library/Formula/tig.rb
@@ -43,10 +43,21 @@ class Tig < Formula
   def install
     system "./autogen.sh" if build.head?
     system "./configure", "--prefix=#{prefix}", "--sysconfdir=#{etc}"
-    system "make", "install"
+    system "make"
+    # Ensure the configured `sysconfdir` is used during runtime by
+    # installing in a separate step.
+    system "make", "install", "sysconfdir=#{prefix/"share/tig/examples"}"
     system "make install-doc-man" if build.with? "docs"
     bash_completion.install "contrib/tig-completion.bash"
     zsh_completion.install "contrib/tig-completion.zsh" => "_tig"
     cp "#{bash_completion}/tig-completion.bash", zsh_completion
+  end
+
+  def caveats; <<-EOS.undent
+    A sample of the default configuration has been installed to:
+      #{prefix/"share/tig/examples/tigrc"}
+    to override the system-wide default configuration, copy the sample to:
+      #{etc/"tigrc"}
+    EOS
   end
 end


jonas:
Trace of installation:

```
[20:02:13]fonseca:/usr/local#tig-install-tigrc-as-sample > git show --no-patch --oneline
ef2783b tig: install system tigrc as sample
[20:02:17]fonseca:/usr/local#tig-install-tigrc-as-sample > brew install tig --HEAD
==> Cloning https://github.com/jonas/tig.git
Cloning into '/Library/Caches/Homebrew/tig--git'...
remote: Counting objects: 253, done.
remote: Compressing objects: 100% (229/229), done.
remote: Total 253 (delta 52), reused 89 (delta 20), pack-reused 0
Receiving objects: 100% (253/253), 569.58 KiB | 0 bytes/s, done.
Resolving deltas: 100% (52/52), done.
Checking connectivity... done.
==> ./autogen.sh
==> ./configure --prefix=/usr/local/Cellar/tig/HEAD --sysconfdir=/usr/local/etc
==> make install sysconfdir=/usr/local/Cellar/tig/HEAD/share/tig/examples
==> Caveats
A sample of the default configuration has been installed to:
  /usr/local/Cellar/tig/HEAD/share/tig/examples/tigrc
to override the system-wide default configuration, copy the sample to:
  /usr/local/etc/tigrc

Bash completion has been installed to:
  /usr/local/etc/bash_completion.d

zsh completion has been installed to:
  /usr/local/share/zsh/site-functions
==> Summary
🍺  /usr/local/Cellar/tig/HEAD: 9 files, 388K, built in 14 seconds
[20:03:00]fonseca:/usr/local#tig-install-tigrc-as-sample > find /usr/local/etc/ -name tigrc
[20:04:56]fonseca:/usr/local#tig-install-tigrc-as-sample > /usr/local/bin/tig --version
tig version 2.1.1-g0ad3327-dirty
```

Library/Formula/tig.rb
```diff
@@ -43,10 +43,18 @@ class Tig < Formula
   def install
     system "./autogen.sh" if build.head?
     system "./configure", "--prefix=#{prefix}", "--sysconfdir=#{etc}"
-    system "make", "install"
+    system "make", "install", "sysconfdir=#{prefix/"share/tig/examples"}"
```
DomT4: Won't this make it rely on that directory for the config permanently, rather than just installing the sample?

jonas: Yes, you are right. `make install` will have to be done in a separate step. I'll try to find time to update the patch against the latest version and see if that fixes the build failure.

DomT4: If you just want to install an example configuration you could do:

```ruby
pkgshare.install "/path/to/example/config"
```
After `make install` has executed?

bfontaine:
Ping?

Library/Formula/tig.rb
jonas: The problem is that `make install` will copy `tigrc` to `sysconfdir`. Tig does not currently provide a way to disable installation of `tigrc`, but to simplify the Homebrew formula it might make sense to add it.

jonas:
I've rebased the patch on top of current `master` and split the build and installation into two steps so that the `sysconfdir` variable is only overridden when `tigrc` is copied to the example directory.

Let me know if you prefer that I add [an option to not install tigrc](https://github.com/Homebrew/homebrew/pull/44055#discussion_r48541196) during `make install`.

bfontaine:
Thanks! I don’t think such option would be necessary given that we install `tigrc` in an example directory.

mikemcquaid:

PR #44057, opened 2015-09-17
nuttcp: remove mirror with incompatible file type

apjanke:
Reverts 46e8c91bc8a2a789772b1a940d7b1c90ade0d575, which added the mirror, but with a .gz instead of a .bz2 file, so the SHAs differ.

Now that the main site is up, `brew install` fails because the SHA is for the mirrored .gz instead of the primary .bz2. (Mirror only has .gz, main only has .bz2.) I downloaded both tarballs and verified that their contents are in fact the same.


DomT4:
@BrewTestBot test this please

dunn:
😱

dunn:
Thanks for catching that, merged in d53c4e5386f11bb80d4be99cd5370c9382661bbd.

PR #44058, opened 2015-09-17
Fix brew update for taps with non-master branches

### LINKED ISSUE #43865: brew update fails on taps without a master branch
Hi, I made a new repo for my homebrew formulae and have run into an issue where I cannot update my tap because there is no branch labeled "master" as part of the repo. Some investigation into this found the problem lies [here](https://github.com/Homebrew/homebrew/blob/12fa55ffb258225ff85cb2b1827b2e1175decaaa/Library/Homebrew/cmd/update.rb#L197-L199). I ran `brew update --debug --verbose` and if you look at the log file [here](https://gist.github.com/samdmarshall/c104b7f4c7869ac244b1#file-brew-update-failure-log-L29-L32) it appears that regardless of the current branch the tap is on, it will try to switch to master. I checked the issues list and found the reasoning for this in (#43472). I don't propose changing that behavior but I think there needs to be some handling around the case that there can be repos without a master branch, or at the very least updated documentation to indicate that having a branch labeled "master" is a requirement as part of the tap documentation [here](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/brew-tap.md). Thanks!

mistydemeo:
The updater code currently assumes that the default branch of all taps is named `master` and hardcodes that. This breaks updating taps which use a different default branch name; `brew tap` is able to clone those repositories, but users will get "Error: Failed to update tap" when trying to run `brew update`.

Fixes #43865.

### DIFF
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -187,6 +187,15 @@ class Updater
       @stashed = true
     end
 
+    # The upstream repository's default branch may not be master;
+    # check refs/remotes/origin/HEAD to see what the default
+    # origin branch name is, and use that. If not set, fall back to "master".
+    begin
+      @upstream_branch = `git symbolic-ref refs/remotes/origin/HEAD 2>/dev/null`
+      @upstream_branch = @upstream_branch.chomp.sub('refs/remotes/origin/', '')
+    rescue ErrorDuringExecution
+      @upstream_branch = "master"
+    end
 
     begin
       @initial_branch = `git symbolic-ref --short HEAD 2>/dev/null`.chomp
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/update.rb
```diff
@@ -187,6 +187,14 @@ def pull!(options = {})
       @stashed = true
     end
 
+    # The upstream repository's default branch may not be master;
+    # check refs/remotes/origin/HEAD to see what the default
+    # origin branch name is, and use that. If not set, fall back to "master".
+    begin
+      @upstream_branch = `git symbolic-ref refs/remotes/origin/HEAD 2>/dev/null`.chomp.sub('refs/remotes/origin/', '')
```
mikemcquaid: Could you maybe split this into multiple lines?

mikemcquaid:
:+1:

PR #44067, opened 2015-09-17
texinfo: install reference card

afh:
Texinfo comes with reference card in pdf and tex format, that users may find very helpful.
I know I do :smile:


xu-cheng:

afh:
Thanks for merging.

PR #44073, opened 2015-09-17
rgxg 0.1 (new formula)

jrwittmann:
Second attempt to add rgxg to the brew repository (see also https://github.com/Homebrew/homebrew/pull/20219).

### DIFF
--- a/Library/Formula/rgxg.rb
+++ b/Library/Formula/rgxg.rb
@@ -1,0 +1,15 @@
+class Rgxg < Formula
+  desc "rgxg (ReGular eXpression Generator)"
+  homepage "http://rgxg.sourceforge.net"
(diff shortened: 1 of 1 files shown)


Library/Formula/rgxg.rb
```diff
@@ -0,0 +1,15 @@
+class Rgxg < Formula
+  desc "rgxg (ReGular eXpression Generator)"
```
DomT4: Don't need to use the formula name inside the `desc`. It'll be automatically prepended when people see it.

DomT4:
Bit concerned this hasn't been updated since the initial release, but it made it into Debian & Ubuntu so that's good enough for us in this situation.

Merged in https://github.com/Homebrew/homebrew/commit/fdb742a94fd7ce7b712071124189fd8646a1692e. Thank you for your contribution to Homebrew @jrwittmann!

PR #44074, opened 2015-09-17
WebSphere Liberty Profile

trajano:


Library/Formula/websphere-liberty-profile.rb
```diff
@@ -0,0 +1,70 @@
...
+  homepage "https://developer.ibm.com/wasdev/websphere-liberty/"
+  url "https://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/wasdev/downloads/wlp/8.5.5.7/wlp-javaee7-8.5.5.7.zip"
+  sha256 "77f710c93cfd45661782d9895d5a17caa751a4186003753044e06b730d380bba"
+
+  depends_on :java => "1.7+"
+
+  def install
+    rm_rf Dir["bin/*.bat"]
+    rm_rf "usr"
+    rmdir "usr"
```
DomT4: You shouldn't need both `rm_rf` and `rmdir` for the same folder, surely?

Library/Formula/websphere-liberty-profile.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    binaries = %w[binaryLog collective ddlGen installUtility securityUtility serverSchemaGen client configUtility featureManager isadc productInfo server]
```
DomT4: It'd be nice to split this onto two lines to make the formula more readable. You can still do that in the array.

Library/Formula/websphere-liberty-profile.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    binaries.each do |name|
+      inreplace "bin/#{name}" do |s|
+        s.gsub! %r{WLP_INSTALL_DIR=`unset CDPATH; cd \"\$dirname/\.\./?\" && pwd`}, "WLP_INSTALL_DIR=#{libexec}"
+      end
+    end
+
+    libexec.install Dir["*"]
+
+    # do not use bin.install because IDE integration may require the path to the scripts be in the same place.
```
DomT4: Don't particularly need this comment.

Library/Formula/websphere-liberty-profile.rb
```diff
@@ -0,0 +1,70 @@
...
+  def post_install
+    # create a link for the usr folder which contains the files that should be persisted between re-installs
+    (var/"websphere-liberty-profile/servers").mkpath
+    (var/"websphere-liberty-profile/shared/apps").mkpath
+    (var/"websphere-liberty-profile/shared/config").mkpath
+    (var/"websphere-liberty-profile/shared/resources").mkpath
+    ln_s "#{var}/websphere-liberty-profile", "#{libexec}/usr"
+  end
+
+  def caveats; <<-EOS.undent
```
DomT4: Unless these caveats are specific to Homebrew, we don't really need them.

Library/Formula/websphere-liberty-profile.rb
```diff
@@ -0,0 +1,70 @@
...
+      inreplace "bin/#{name}" do |s|
+        s.gsub! %r{WLP_INSTALL_DIR=`unset CDPATH; cd \"\$dirname/\.\./?\" && pwd`}, "WLP_INSTALL_DIR=#{libexec}"
+      end
+    end
+
+    libexec.install Dir["*"]
+
+    # do not use bin.install because IDE integration may require the path to the scripts be in the same place.
+    binaries.each do |name|
+      bin.install_symlink "#{libexec}/bin/#{name}"
```
DomT4: I'm a tad worried about installing binaries simply called "server" and "client" into the bin. It'd be nice to prefix those names somehow so they aren't very generic.

Library/Formula/websphere-liberty-profile.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  test do
+    # Ideally this would be following the steps noted in caveats
+    #
+    # system bin/"server", "create", "testServer"
+    # system bin/"server", "start", "testServer"
+    # system bin/"server", "stop", "testServer"
+    # rm_rf "#{libexec}/usr/servers/testServer"
+    #
+    # However, that would fail on --sandbox mode because the data is in var
```
DomT4: Is there no way to temporarily override the data directory?

Library/Formula/websphere-liberty-profile.rb
trajano: You would think so. But without the Ensor afterwards it seems to keep on keeping the usr folder.

Library/Formula/websphere-liberty-profile.rb
trajano: I think it is useful to note. Otherwise why not just use bin.install. Which would've been much simpler.

Library/Formula/websphere-liberty-profile.rb
trajano: What's the standard for that. Those are the names they used.

Library/Formula/websphere-liberty-profile.rb
trajano: I'm welcome for someone else to try it out. I just wrote this primarily so I have an easier way of getting the program installed. 

Mind you elastic search does the same thing.

Library/Formula/websphere-liberty-profile.rb
DomT4: We fairly strictly either refuse to put generic-named binaries in `bin`, or we symlink them with a new name, such as `websphere-liberty-server` and such.

Library/Formula/websphere-liberty-profile.rb
DomT4: That we're using `install_symlink` is likely to be enough explanation.

Library/Formula/websphere-liberty-profile.rb
trajano: So just change the name when we install right?  
May I use `wlp-server` and `wlp-client` it's not WebSphere Liberty but WebSphere Liberty Profile.

trajano:
Anyway I will hold off on this until https://github.com/Homebrew/homebrew/pull/43950 gets merged because I am not sure how to keep two formulas active without a lot of hand holding with git rebase switching the two commits around.

Library/Formula/websphere-liberty-profile.rb
trajano: Seems to work without it, I have removed the line. Will update the PR after #43950 is merged

bfontaine:
Anything new here?

DomT4:
Closing as dead, given the other PR is closed. Please feel free to resubmit both with comments addressed if desired. Thank for your submissions to Homebrew!

PR #44076, opened 2015-09-17
glm 0.9.7.1

cHoco:


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/5ab90b645adb24ebaa598e0992788d8492d3bff5. Thank you for your contribution to Homebrew @cHoco!

jellyshen:
This version(0.9.7.1) of glm installed from brew doesn't install FindGLM.cmake or glmConfig.cmake. It is very kind If you can modify the formula to install these file.

DomT4:
Did the previous versions?

jellyshen:
I don't know whether the previous version install these files because I don't know how to install the old versions, but I found a issue [#28403](https://github.com/Homebrew/homebrew/issues/28403) about this.

When i install glfm from brew, the glfmConfig.cmake is installed in /usr/local/lib/cmake/glfw directory, so i think glm could  install these files in /usr/local/lib/cmake/glm directory.

cHoco:
It was included in the previous versions, but with gym 0.9.7.1 developers decided to remove old FindGLM.cmake script and use new glmConfig.cmake glmTargets.cmake scripts. Old formula just copied whole src directory in include, now we need to run make to generate the config files. https://github.com/Homebrew/homebrew/issues/44701 should fix the issue

PR #44077, opened 2015-09-17
Ceph (new formula)

bertjwregeer:
Add a new Ceph formula. Unfortunately the latest release from Ceph.com does not yet compile on OS X (I'm hoping the next point release will), so it's head only for now, hopefully that restriction can get removed in the near future.


Library/Formula/ceph.rb
```diff
@@ -0,0 +1,60 @@
...
+  depends_on "openssl"
+  depends_on "snappy"
+  depends_on "cryptopp"
+  depends_on "boost"
+  depends_on "leveldb"
+
+  # Unfortunately it installs a Python script unconditionally, we depend on the
+  # Homebrew python so that when that script gets installed it gets installed
+  # into the Homebrew python site-packages rather than failing to install it in
+  # the system python site-packages.
```
DomT4: We should patch this and submit that patch upstream rather than relying on Homebrew's Python unconditionally.

Library/Formula/ceph.rb
```diff
@@ -0,0 +1,60 @@
...
+                          "--without-babeltrace",
+                          "--prefix=#{prefix}"
+    system "make"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/rados", "--help"
+    system "#{bin}/librados-config --version"
+    system "#{bin}/librados-config --vernum"
```
DomT4: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. e.g. write out a `relx.config` and do something with it. Thanks!

bertjwregeer: No, not unless you have a full Ceph cluster and a proper configuration file. All of these are tools that require a remote server to communicate with. My main goal was to get librados, so that I could develop my own applications against Ceph.

Library/Formula/ceph.rb
bertjwregeer: I am not familiar enough with autotools or this massive code base to even attempt to patch this, let alone understand what I need to modify to fix this. Hence the work-around.

DomT4: I'm afraid using Homebrew's Python by default isn't a workaround we'd likely be happy to take.

Python is a significant dependency to force people to use and consequently we have a policy of using the system Python in the core as heavily as possible.

Library/Formula/ceph.rb
DomT4: I see. In which case, one of the above tests is fine, unless there's some situation where `--version` would pass a basic test but `--vernum` wouldn't.

bertjwregeer: You are right, removing one wouldn't hurt anything. It's the only two flags that command takes, so I dropped it in. Will fix it up in a second.

Library/Formula/ceph.rb
bertjwregeer: The amount of people interested in Ceph will be extremely low on OS X (as in, I and the guy porting it to OS X are probably the only two people that even think about it :P), if this won't be accepted I'll drop it into my tap and call it a day.

DomT4: I'll let @tdsmith be judge & jury on this, but I suspect if you want it in the core it'll have to build against the system Python.

tdsmith: I bet `inreplace "ceph/src/ceph-detect-init/Makefile.am", "python setup.py install $$root $$options", "python setup.py install --prefix #{lib}/python2.7/site-packages --single-version-externally-managed --record=installed.txt"` will do ya.

bertjwregeer: I'll drop that in tomorrow morning and take a look.

mikemcquaid: > The amount of people interested in Ceph will be extremely low on OS X (as in, I and the guy porting it to OS X are probably the only two people that even think about it :P), if this won't be accepted I'll drop it into my tap and call it a day.

That probably seems like a good reason to keep it in your tap.

bertjwregeer: @tdsmith That line worked :-). Updated the patchset.

@mikemcquaid Let me know if that is what the decision is. Ultimately I think it would be nice to have it in tree, but that may come down the line.

mikemcquaid:
As it needs patching and is niche by your admission: passing on this, sorry. I'd suggest creating your own tap to host this. Sorry!

bertjwregeer:
bertjwregeer/homebrew-mine... Done.

PR #44091, opened 2015-09-18
shadowsocks-libev 2.3.3

blackgear:


dunn:
Is there any chance of adding a quick test while you're at it? Thanks!

blackgear:
Test method has been added. This test will start a server listen `127.0.0.1:9998`, then start a client listen `127.0.0.1:9999` as a socks5 proxy and send data through `127.0.0.1:9998` as  server side, finally use `curl` to test if this tunnel works correctly. It will test both ss-server and ss-local to ensure this two most importance parts works.

blackgear:
Seems something wrong with your test-bot....
`brew tests` failed but `brew test --verbose shadowsocks-libev` passed.

DomT4:
@BrewTestBot test this please

DomT4:
Despite the naming similarity, `brew tests` and `brew test` serve different functions. The latter is strictly related to that formula `test do` block, whereas the former checks over Homebrew's core code to check the PR hasn't broken that. In this case, a prior commit caused `brew tests` to fail.

dunn:
Merged in 54d5e53697cfb8ce48fc196d523ab6e89f931fc3, thanks so much!

PR #44092, opened 2015-09-18
cccc 3.1.4 (new formula)

kbinani:


Library/Formula/cccc.rb
```diff
@@ -0,0 +1,47 @@
+class Cccc < Formula
+  desc "C and C++ Code Counter"
+  homepage "http://sourceforge.net/projects/cccc/"
+  url "https://downloads.sourceforge.net/project/cccc/cccc/3.1.4/cccc-3.1.4.tar.gz"
+  sha256 "27b3eca9a83a66799332363a80cc4bcd6db9869eddbda1a59a25cdace3ff4dbf"
+  patch :DATA
```
dunn: Has this been submitted upstream?

Library/Formula/cccc.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    # these targets have dependencies, but nothing mentioned explicitly in makefile.
+    system "make", "pccts"
+    system "make", "cccc"
+    system "make", "test"
+    system "make", "-e", "install", "INSTDIR=#{bin}"
+  end
+
+  test do
+    system "which", "cccc"
```
dunn: We'd really like a test that runs the program, to check basic functionality. Thanks!

Library/Formula/cccc.rb
kbinani: Reported to upstream: https://sourceforge.net/p/cccc/bugs/55/

Library/Formula/cccc.rb
kbinani: Thanks for the comments!  I've upgraded the test contents.

Library/Formula/cccc.rb
```diff
@@ -0,0 +1,58 @@
...
+  desc "C and C++ Code Counter"
+  homepage "http://sourceforge.net/projects/cccc/"
+  url "https://downloads.sourceforge.net/project/cccc/cccc/3.1.4/cccc-3.1.4.tar.gz"
+  sha256 "27b3eca9a83a66799332363a80cc4bcd6db9869eddbda1a59a25cdace3ff4dbf"
+
+  # reported to upstream: https://sourceforge.net/p/cccc/bugs/55/
+  patch :DATA
+
+  def install
+    # "make cccc" will fail if parallelized.
```
DomT4: We need to report this upstream as well.

Library/Formula/cccc.rb
```diff
@@ -0,0 +1,58 @@
+class Cccc < Formula
+  desc "C and C++ Code Counter"
+  homepage "http://sourceforge.net/projects/cccc/"
+  url "https://downloads.sourceforge.net/project/cccc/cccc/3.1.4/cccc-3.1.4.tar.gz"
+  sha256 "27b3eca9a83a66799332363a80cc4bcd6db9869eddbda1a59a25cdace3ff4dbf"
+
+  # reported to upstream: https://sourceforge.net/p/cccc/bugs/55/
+  patch :DATA
```
DomT4: Just to be a bit neater it might be nice to turn these into an `inreplace` in `def install`, i.e.

```ruby
def install
  # Bug report link
  inreplace "cccc_tbl.cc", "erase(value_iterator);", "map_t::erase(value_iterator);"
  inreplace "makefile", 'su root -c "make -f install.mak"', "make -f install.mak"

  # Bug report link
  ENV.deparallelize
end
```

Library/Formula/cccc.rb
kbinani: Reported to upstream: https://sourceforge.net/p/cccc/bugs/56/
I didn't report about inreplace part

```
inreplace "makefile", 'su root -c "make -f install.mak"', "make -f install.mak"
```
because this is just a workaround for brew. Is this okey?

Library/Formula/cccc.rb
kbinani: Removed `patch :DATA`. Thank you for the suggestion!
Also checked that the `inreplace` doesn't affect other part of the 'cccc_tbl.cc' file.

Library/Formula/cccc.rb
dunn: That seems like a more general issue to me; the makefile shouldn't require root for anyone.

DomT4: Aye, ideally we can ask upstream about root. If they turn around and tell us they're not changing it that's their choice, but ideally in these situations `su`/`sudo` invocation should be left to user choice rather than hardcoded like this.

kbinani: Also reported to upstream: https://sourceforge.net/p/cccc/bugs/57/

Library/Formula/cccc.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    # "make cccc" will fail if parallelized.
+    # reported to upstream: https://sourceforge.net/p/cccc/bugs/56/
+    ENV.deparallelize
+
+    # these targets have dependencies, but nothing mentioned explicitly in makefile.
+    system "make", "pccts"
+    system "make", "cccc"
+    system "make", "test"
+    system "make", "-e", "install", "INSTDIR=#{bin}"
```
bfontaine: Does it not work without `-e`?

kbinani: Sorry, `-e` is not necessary here. I'll remove it.

Library/Formula/cccc.rb
bfontaine: Could you please add this link as a comment above `patch :DATA`? Thanks!

Library/Formula/cccc.rb
```diff
@@ -0,0 +1,47 @@
+class Cccc < Formula
+  desc "C and C++ Code Counter"
+  homepage "http://sourceforge.net/projects/cccc/"
+  url "https://downloads.sourceforge.net/project/cccc/cccc/3.1.4/cccc-3.1.4.tar.gz"
+  sha256 "27b3eca9a83a66799332363a80cc4bcd6db9869eddbda1a59a25cdace3ff4dbf"
+  patch :DATA
+
+  def install
+    # "make cccc" will fail if parallelized.
+    ENV.deparallelize
```
mikemcquaid: Can you please submit this issue to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Library/Formula/cccc.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    # these targets have dependencies, but nothing mentioned explicitly in makefile.
+    system "make", "pccts"
+    system "make", "cccc"
+    system "make", "test"
+    system "make", "install", "INSTDIR=#{bin}"
+  end
+
+  test do
+    system "which", "cccc"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikemcquaid:
Unfortunately it doesn't look like this software is maintained any more.

mikemcquaid:
(which means we probably shouldn't add it to Homebrew)

kbinani:
I had created 3 reports to upstream about this PR:
http://sourceforge.net/p/cccc/bugs/55/
http://sourceforge.net/p/cccc/bugs/56/
http://sourceforge.net/p/cccc/bugs/57/

But actually, there is no response from owners. So I'll giving up this PR.

mikemcquaid:
Yeh, I'm sorry @kbinani, it sucks when projects are abandoned. Thanks for your great work here and I hope you'll contribute to Homebrew again.

denji:
@kbinani https://github.com/AlDanial/cloc

PR #44095, opened 2015-09-18
yafc: Adds patch for mode parsing

gabriel-kaam:
There is a bug in the yafc project (v1.3.6) that happens when using the **chmod** command on a ssh connection .
This bug is fixed by this commit : https://github.com/sebastinas/yafc/commit/a16c1c985f3d4e2b676231f4773358927041d8e9 , but not realeased yet .

This PR added the patch to the yafc formula so people don't have to wait for the next yafc release .


gabriel-kaam:
This PR needs https://github.com/Homebrew/homebrew/pull/44094 to build ( sorry should have made all the changes in a single commit )

PR #44096, opened 2015-09-18
gradle 2.7

lnr0626:
Updated to new version of gradle.


lnr0626:
whoops - didn't see that there's already a PR for this. Sorry bout that.

PR #44101, opened 2015-09-18
vala 0.30.0

tschoonj:


Library/Formula/vala.rb
```diff
@@ -10,10 +10,11 @@ class Vala < Formula
     sha256 "b55053adbebaf5d9d016002fb8d2a26b72398649e5a54c5621a68a9e62a31e62" => :mountain_lion
   end
 
-  devel do
-    url "https://download.gnome.org/sources/vala/0.29/vala-0.29.3.tar.xz"
-    sha256 "687f89c404cec528bc6c8e80d80146be0a42aa30f60475fcaefad37bd12949fe"
-  end
+  # currently there is no development release available
```
DomT4: Please just remove the `devel` blocks in this situation, Thanks Tom!

tschoonj: If you don't mind I would like to keep them: it will make it clear that this is a formula that (usually) has a devel version available. Otherwise this may be forgotten.

Also it saves a bit on typing when the next devel release gets added :-)

DomT4: I'm afraid we very deliberately kick stuff out of formulae if it isn't being used :/. Commented out blocks tend to encourage people to do the same thing with other elements on updates, like our bottle blocks, sadly.

tschoonj:
Travis-CI failed for some reason but it's ok for Jenkins..

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/dd1e4b3197e8748b93855cce10055a4293aaeae1 & https://github.com/Homebrew/homebrew/commit/405622277337d71d99391a0c636009bf40770587. Thank you again Tom.

PR #44104, opened 2015-09-18
test-bot: better fold id for travis

xu-cheng:
* Only use first two word in command to keep id short
* Append unique number in the id.

### DIFF
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -80,8 +80,10 @@ module Homebrew
 
     def puts_command
       if ENV["TRAVIS"]
+        @@travis_step_num ||= 0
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> Only use first two word in command to keep id short

Makes sense :+1: 



> Append unique number in the id.

I don't object, just curious why?

xu-cheng:
> > Append unique number in the id.
> I don't object, just curious why?

I found travis will merge two fold together if they share same id.

mikemcquaid:
> I found travis will merge two folds together if they share same id.

Ah, ok. Do we not just have to do that for names that are the same rather than for all steps?

xu-cheng:
>  Do we not just have to do that for names that are the same rather than for all steps?

It's simple to do for all steps. And it can show us building order in case the log rendered incorrect.

mikemcquaid:
OK.

PR #44112, opened 2015-09-18
New formula for 'trend'

hawson:
This is a pull request for the 'trend' program.
Homepage:  http://www.thregr.org/~wavexx/software/trend/

The test cases are sub-optimal, but there isn't a clean way to automate a test yet.  The program runs "forever", unless a user manually exits the program.  The "--help" option returns a non-zero exit code, as does running the binary without arguments.

### DIFF
--- a/Library/Formula/trend.rb
+++ b/Library/Formula/trend.rb
@@ -16,7 +16,6 @@ class Trend < Formula
 
     Dir.chdir '..'
     man1.install 'trend.1'
-  
     doc.install 'AUTHORS', 'NEWS', 'THANKS', 'COPYING', 'README', 'TODO'
 
   end
--- a/Library/Formula/trend.rb
+++ b/Library/Formula/trend.rb
@@ -21,6 +21,17 @@ class Trend < Formula
   end
 
   test do
-    system "echo {1..5} {4..2} {3..10} | -FD - 20x1 src/trend"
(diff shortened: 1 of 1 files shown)


Library/Formula/trend.rb
```diff
@@ -0,0 +1,37 @@
+class Trend < Formula
+  desc "a general-purpose, efficient trend graph"
+  homepage "http://www.thregr.org/~wavexx/software/trend/"
+  url "http://www.thregr.org/~wavexx/software/trend/releases/trend-1.2.tar.gz"
+  sha256 "40b0f58101801c9373d2b1d990eb327a4582e9a96e4eb2b43faf87346f9dd64f"
+
+  depends_on :x11
+  depends_on 'freeglut'
```
mistydemeo: This formula's in the homebrew/x11 tap, not the main repository, so you need to refer to it by its full name: `homebrew/x11/freeglut`

Maybe it makes more sense for this to go to that repo?

hawson: So two problems then?

1. the pull request should be against the homebrew-x11 repo
1. the dependency should be 'homebrew/x11/freeglut'

DomT4:
Aye. Please resubmit this to `homebrew/x11`. In that tap you can just call the dependency as `depends_on "freeglut"` without needing to specify the full tap path. Thanks!

PR #44113, opened 2015-09-18
uhd 003_009_001

tsmithhisler:
Update uhd version to 003.009.001


bfontaine:
Thank you for your contribution to Homebrew @tsmithhisler; we appreciate it! :smile:

PR #44120, opened 2015-09-18
rust 1.3.0

dtrebbien:


dtrebbien:
Both Travis CI builds failed with "The job exceeded the maximum time limit for jobs, and has been terminated."

stouset:
Rust takes a long time to compile. How would one circumvent that?

DomT4:
There will be a Jenkins CI build done soonish which doesn't have such a limited timeout.

mikemcquaid:
Also if there's any way to speed up the build that would be handy too.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/aea0f1dd93220b6a28db8e79ccedf05d451c1ba8. Thank you for your contribution to Homebrew @dtrebbien; we appreciate it! :smiley_cat:

dtrebbien:
Just to follow up on reducing build time, I asked on \#rust-internals IRC and was told that build time is a known issue, but there is currently no master plan for fixing it.

One suggestion was to use an external LLVM rather than having rust build its own LLVM.  Testing this approach out this morning, I see that the build time is actually slightly longer on my MacBook Pro when using Homebrew's LLVM (i.e. external LLVM):
32.0 minutes with in-tree LLVM (https://github.com/rust-lang/llvm) versus 32.3, 33.7 minutes using Homebrew's LLVM 3.6.2.

As noted at rust-lang/llvm#21, Rust's LLVM has some additional optimizations.

dtrebbien:
PR #44138 adds a `--with-llvm` option

PR #44126, opened 2015-09-18
Yafc fix url plus chmod bug

gabriel-kaam:
This PR includes 2 fixes : 
===================

Seems like the url to download the tar.gz has changed from http://www.yafc-ftp.com/upload/yafc-1.3.6.tar.xz to http://www.yafc-ftp.com/downloads/yafc-1.3.6.tar.xz

----

There is a bug in the yafc project (v1.3.6) that happens when using the chmod command on a ssh connection .
This bug is fixed by this commit : sebastinas/yafc@a16c1c9 , but not realeased yet .

This PR added the patch to the yafc formula so people don't have to wait for the next yafc release .

### DIFF
--- a/Library/Formula/yafc.rb
+++ b/Library/Formula/yafc.rb
@@ -1,7 +1,7 @@
 class Yafc < Formula
   desc "Command-line FTP client"
   homepage "http://www.yafc-ftp.com/"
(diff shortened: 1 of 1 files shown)


mikemcquaid:

PR #44135, opened 2015-09-19
update: fix test for non-master update support

UniqMartin:
Commit a71d4a9b (PR #44058) brought support for updating repositories with a default branch different from master. This fixes the tests broken by that commit. Unfortunately, without this fix `brew tests` continues to be broken and consequently all bot builds fail.

### DIFF
--- a/Library/Homebrew/test/test_updater.rb
+++ b/Library/Homebrew/test/test_updater.rb
@@ -58,6 +58,7 @@ class UpdaterTests < Homebrew::TestCase
     FormulaVersions.stubs(:new).returns(stub(:formula_at_revision => "2.0"))
     @updater.diff = fixture(fixture_name)
     @updater.in_repo_expect("git diff --quiet", true)
+    @updater.in_repo_expect("git symbolic-ref refs/remotes/origin/HEAD 2>/dev/null", "refs/remotes/origin/master")
     @updater.in_repo_expect("git symbolic-ref --short HEAD 2>/dev/null", "master")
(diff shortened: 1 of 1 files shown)


UniqMartin:
@mistydemeo @mikemcquaid This seems rather important, thus the mention to get your attention. (And because you authored/reviewed the referenced commit/PR.)

----

*Is there a better way to get extra attention from Homebrew maintainers in cases like these? Or should I just be patient and the PR will be noticed and acted upon in due time? Really not sure how to best handle this without annoying people. Suggestions and/or documentation pointers are very welcome.*

DomT4:
I've pinged them in the team chat to glance at this if they can. It's a Saturday though, so they may be unavailable. This looks alright to me, but semi reluctant to jump over another maintainers' code without their review.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/80a90b7e7cd585e7f20ff9ce19448fe90be3f26b. It fixes the problem locally, and it passed the bot here when nothing else has been since the prior commit.

If Misty/Mike have a more desired fix later, feel free to revert. This looks alright to me but `¯\_(ツ)_/¯`.

UniqMartin:
Could you also reschedule the Jenkins and Travis builds of the affected PRs? Judging by the Jenkins logs 44125, 44130, and 44132 failed solely because of this. (Other PRs from that time frame also had other test failures.) — Thanks a lot @DomT4 for handling this!

DomT4:
Aye. Going back through them one by one at the moment to avoid giving the CI a panic attack.

PR #44138, opened 2015-09-19
rust: Add --with-llvm option

dtrebbien:
--with-llvm allows Homebrew's LLVM to be used instead of Rust's LLVM.

### DIFF
--- a/Library/Formula/rust.rb
+++ b/Library/Formula/rust.rb
@@ -31,8 +31,11 @@ class Rust < Formula
     sha256 "04f68ef26b4ca636b35be1d17ad2b641727bbc10cec52582a9e3a64b5784dbeb" => :mavericks
   end
 
+  option "with-llvm", "Build with brewed LLVM. By default, Rust's LLVM will be used."
+
   depends_on "cmake" => :build
   depends_on "pkg-config" => :build
+  depends_on "llvm" => :optional
   depends_on "openssl"
 
   # According to the official readme, GCC 4.7+ is required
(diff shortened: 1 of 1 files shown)


DomT4:
I've kicked this off our CI for now. We don't test non-default options so the changes in this PR wouldn't actually be reviewed.

Library/Formula/rust.rb
```diff
@@ -31,8 +31,11 @@ class Rust < Formula
     sha256 "04f68ef26b4ca636b35be1d17ad2b641727bbc10cec52582a9e3a64b5784dbeb" => :mavericks
   end
 
+  option "with-llvm", "Build with brewed LLVM"
```
dunn: Should we call this `with-brewed-llvm` (like the git formula does), to make sure users don't think that otherwise somehow their Rust won't have LLVM at all?

dtrebbien: Yes, that sounds much more clear.  In the new commit the option is renamed and a better explanation of the option is added.

mikemcquaid: I think actually `with-llvm` makes more sense and we should move more formulae to use that approach.

DomT4: Yeah. We can always make the description as it was, to be explicit about the brewed element. I really dislike `if build.with?` style dependencies.

dunn: Oh right, that way we can just do `depends_on "llvm" => :optional`.  Sounds good.

dtrebbien: I have now uploaded a new commit that changes the option name back to `--with-llvm`, and keeps the more descriptive description.

dtrebbien:
ping

DomT4:
Thanks for the nudge. Merged in https://github.com/Homebrew/homebrew/commit/4cd4135c5b8c9924217af3ad0875ba5b0cd880d6. Thank you for your contribution to Homebrew @dtrebbien! :smiley_cat:

PR #42452, opened 2015-08-04
Qt 5 Fix for El Capitan

jaeyeun97:
Created a fix for installing Qt 5 in El Capitan.

In addition to the previous fix already added to the formula, there needs another one to fix the path issue to the dynamic libraries. The new Xcode 7 does not have the .dylibs in the SDKROOT, and that needed to be redirected to /usr/lib.

I do not know how to contribute to the Qt project itself right now (I am looking into it). I would update when this patch goes upstream.

PR #44143, opened 2015-09-19
qt5: fix various build and runtime errors

UniqMartin:
As discussed in #42452. All addressed build errors are somehow related to Xcode 7 (Clang version, OS X SDK, …), so these fixes are needed for both Yosemite and El Capitan. Fortunately, most of this stuff will be in Qt 5.5.1 as far as I can tell.

Thanks to @DomT4, @mikemcquaid, and @swallat for preparing the patches used here.

### DIFF
--- a/Library/Formula/qt5.rb
+++ b/Library/Formula/qt5.rb
@@ -30,14 +30,35 @@ class Qt5 < Formula
     # when this formula is updated to 5.5.1
     patch :DATA
 
-    # Upstream commit to fix the fatal build error on OS X El Capitan.
+    # Build error: Fix build with clang 3.7.
     # https://codereview.qt-project.org/#/c/121545/
     # Should land in the 5.5.1 release.
-    if MacOS.version >= :el_capitan
-      patch do
-        url "https://raw.githubusercontent.com/DomT4/scripts/2107043e8/Homebrew_Resources/Qt5/qt5_el_capitan.diff"
-        sha256 "bd8fd054247ec730f60778e210d58cba613265e5df04ec93f4110421fb03b14a"
-      end
+    patch do
+      url "https://raw.githubusercontent.com/DomT4/scripts/2107043e8/Homebrew_Resources/Qt5/qt5_el_capitan.diff"
+      sha256 "bd8fd054247ec730f60778e210d58cba613265e5df04ec93f4110421fb03b14a"
+    end
+
+    # Build error: Fix library paths with Xcode 7 for QtWebEngine.
+    # https://codereview.qt-project.org/#/c/122729/
+    patch do
+      url "https://raw.githubusercontent.com/Homebrew/patches/2fcc1f8ec1df1c90785f4fa6632cebac68772fa9/qt5/el-capitan-2.diff"
+      sha256 "b8f04efd047eeed7cfd15b029ece20b5fe3c0960b74f7a5cb98bd36475463227"
+    end
+
+    # Build error: Fix CGEventCreateMouseEvent use with 10.11 SDK.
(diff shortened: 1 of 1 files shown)


swallat:
Is this timeout normal, or what could cause this build error?

DomT4:
Just ignore Travis in this case.

swallat:
Not enough disk space or RAM on the Jenkins server?

mikemcquaid:
May be fixed by https://github.com/Homebrew/homebrew/pull/44217.

UniqMartin:
@mikemcquaid Would it make sense to reschedule the test builds for this PR, now that some infrastructure changes have landed? Would be great to have a somewhat working Qt 5 in Homebrew again …

mikemcquaid:
@UniqMartin Yup. Can you just force-push this branch?

xu-cheng:
For travis, a rebase may be required.

mikemcquaid:
Yes, good idea :

UniqMartin:
Okay, rebased and force-pushed. Hope it works …

UniqMartin:
Yay! It worked. :tada:

DomT4:
Is this a bad time to say I was going to suggest re-fixing the `pkg-config` files that currently break at least three other Homebrew formulae?

I accepted a patch previously (Might have been yours actually) but reverted it because it wasn't clear Qt were going to accept it anytime soon & Qt source code is an achy-breaky nightmare to mess with without upstream's acceptance. A relatively simple `inreplace` of the impacted `.pc` files to remove the stray `'` in `post_install` could do the job nicely (Tested this the other night).

mikemcquaid:
@DomT4 Let's merge this as-is for now but then I definitely think it's worth merging that patch if it fixes things.

DomT4:
Aye. I was considering dumping the `inreplace` in `post_install` to avoid putting it back through the CI.

mikemcquaid:
I think fixing it with a new uploaded bottle would be preferable.

mikemcquaid:
Thanks for the great work here @UniqMartin!

DomT4:
Alright, I'll nudge it through the CI at some point overnight when things are quiet. Do we want to re-adopt the patch or `inreplace`?

Agreed, Thanks for sticking with this @UniqMartin. Hopefully will help reduce the number of Qt5 issues we've been seeing.

mikemcquaid:
@DomT4 I reckon re-adopt the patch (and ideally submit it to Qt too).

UniqMartin:
> @DomT4 I reckon re-adopt the patch (and ideally submit it to Qt too).

@mikemcquaid I guess I should finally go through the pains of submitting this patch. I contributed it a while back (#42105, [QTBUG-47162](https://bugreports.qt.io/browse/QTBUG-47162)), but it was reverted because it wasn't submitted to Qt's Gerrit. I had only put it on Qt's issue tracker in the hopes that it will be picked up by a Qt maintainer, but sadly no one seems to read or care about it there …  Thus as of now the issue isn't fixed at all in Qt's 5.5/5.5.1/5.6 branches.

mikemcquaid:
Yeh, there's no point in filing things in their Jira unfortunately, the Gerrit is the only way to go.

UniqMartin:
True, unfortunately. I was hoping it would be easier to contribute a trivial two-line change to Qt. Let's see if I manage to go through their submission process later tonight.

DomT4:
Will leave it to you then if you're happy to deal with the process, Thanks.

UniqMartin:
Here's the follow-up with the `pkg-config` fix: #44410

PR #44038, opened 2015-09-16
libusb 1.0.20

fran6co:

PR #44145, opened 2015-09-19
rkflashtool: Fix encoding issue in the test for mavericks

fran6co:
For some reason the encoding of cat is not the same in Maverkics => https://github.com/Homebrew/homebrew/pull/44038

### DIFF
--- a/Library/Formula/rkflashtool.rb
+++ b/Library/Formula/rkflashtool.rb
@@ -33,6 +33,8 @@ class Rkflashtool < Formula
   test do
     (testpath/"input.file").write "ABCD"
     system bin/"rkcrc", "input.file", "output.file"
-    assert_equal "ABCD\264\366\a\t", `cat output.file`
+    result = shell_output("cat output.file")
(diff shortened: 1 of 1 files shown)


DomT4:
@mistydemeo Am I remembering correctly that Ruby 1.8 doesn't have the `force_encoding` method? Should we couch this in another check so it doesn't bite <ML users?

fran6co:
This would be only for the tests, an easy fix would be adding an OS check only for Mavericks.

DomT4:
[This](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/abduco.rb#L21-L23) is the sort of syntax we've used elsewhere. Let's mimic that here.

fran6co:
Done

fran6co:
@BrewTestBot test this please

fran6co:
It's passing all the checks now @DomT4

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/a6a93e20691e88aad512d6c0000b2144dbfc9ee1. Thank you for your contribution to Homebrew @fran6co!

Issue #44110, opened 2015-09-18
Future Node Icu4c Discussion


bcomnes:
So there is this longstanding issue: https://github.com/Homebrew/homebrew/issues/36681

> So the issue is that the node executable links against libstdc++ because of an archaic deployment target setting.

> It would be cool if node would link against libc++, but that will require someone to convince upstream.

```
bret-mbr:~ bret$ otool -L `which node`
/usr/local/bin/node:
	/usr/local/opt/icu4c/lib/libicui18n.55.dylib (compatibility version 55.0.0, current version 55.1.0)
	/usr/local/opt/icu4c/lib/libicuuc.55.dylib (compatibility version 55.0.0, current version 55.1.0)
	/usr/local/opt/icu4c/lib/libicudata.55.1.dylib (compatibility version 55.0.0, current version 55.1.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1153.18.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
	/usr/lib/libstdc++.6.dylib (compatibility version 7.0.0, current version 104.1.0)
	/usr/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 283.0.0)
```

Even in Node 4+, they are building against libstdc++ which presents a potential issue with the homebrew Icu4c which links against libc++.  While there hasn't been a huge amount of direct evidence that this is  a major issue, it still warrants the considerations of the following options:

1) Stage Icu4c as a resource and staticly build along with node itself. https://github.com/chrmoritz/homebrew/commit/a669d8c7ec4b33ff7f8b0b5e0f184b50c620367e cc @chrmoritz

2) Link against homebrews Icu4c as a dependency https://github.com/bcomnes/homebrew/commit/309e9b4f61053a864698f82f7318829a1abcd2d8

Both seem to work fine.  Long term we can see if we can encourage linking against libc++ upstream but it would be nice to figure something out for now.

cc @DomT4 @mikemcquaid

bcomnes:
After thinking more on this, staging a resource seems to make the most sense currently.

- It lets us include full icu4c support by default
- Keeps homebrew's node in line with the distributed binary
- We can always change to a dependency model once the outstanding issue is resolved upstream.

chrmoritz:
[...]

PR #43973, opened 2015-09-15
Update Node.js to 4.0.0 (Alt)

bcomnes:
The [other PR](https://github.com/Homebrew/homebrew/pull/43714) still needed lots of work, and even went so far as to remove the entire npm staging routine.  

This, I believe, is a minimal subset of changes to the existing formula to make node.js 4+ work later this week once npm 2.14.3 is shipping without any floating patches.  

Please let me know if this PR is undesired or making too much noise.

This should also provide a good testing ground for figuring out which pieces of the formula can be simplified or removed.

Issue #36681, opened 2015-02-09
node and icu4c Yosemite bottles have different C++ standard libraries


MerelyAPseudonym:
(Either that, or there's a misleading warning)

```none
joshuatilles@MBP0206 ~> brew install node
==> Installing node dependency: icu4c
==> Downloading https://downloads.sf.net/project/machomebrew/Bottles/icu4c-54.1.
Already downloaded: /Library/Caches/Homebrew/icu4c-54.1.yosemite.bottle.1.tar.gz
==> Pouring icu4c-54.1.yosemite.bottle.1.tar.gz
==> Caveats
This formula is keg-only, which means it was not symlinked into /usr/local.

Mac OS X already provides this software and installing another version in
parallel can cause all kinds of trouble.

OS X provides libicucore.dylib (but nothing else).

Generally there are no consequences of this for you. If you build your
own software and it requires this formula, you'll need to add to your
build variables:

    LDFLAGS:  -L/usr/local/opt/icu4c/lib
    CPPFLAGS: -I/usr/local/opt/icu4c/include

==> Summary
🍺  /usr/local/Cellar/icu4c/54.1: 242 files, 65M
==> Installing node
==> Downloading https://downloads.sf.net/project/machomebrew/Bottles/node-0.12.0
Already downloaded: /Library/Caches/Homebrew/node-0.12.0.yosemite.bottle.tar.gz
==> Pouring node-0.12.0.yosemite.bottle.tar.gz
Warning: node dependency icu4c was built with a different C++ standard
library (libc++ from clang). This may cause problems at runtime.
==> Caveats
If you update npm itself, do NOT use the npm update command.
The upstream-recommended way to update npm is:
  npm install -g npm@latest

Bash completion has been installed to:
  /usr/local/etc/bash_completion.d
==> Summary
🍺  /usr/local/Cellar/node/0.12.0: 2196 files, 26M
joshuatilles@MBP0206 ~>
```

I'd expect that if I'm letting Homebrew handle the dependencies of a formula for me, it shouldn't do anything that it has to warn me about.

DomT4:
This is a known thing. I haven't found any issues with the setup in weeks of testing, so I don't think it's a big deal FWIW. It's not a misleading warning as such, It's a warning that something may go wrong, not that it will go wrong :).

[...]

PR #44147, opened 2015-09-19
node: add support for statically linking icu4c

chrmoritz:
This PR changes the Node.js formula's behavior from optionally dynamically linking against a brewed `icu4c` to statically linking to a `resource` staged `icu4c` by default to avoid issues with different c++ standard libraries, see #36681. A `resource` is needed, because otherwise the configure script would download `icu4c` itself (after passing an additional configure argument), which is against the formula guidelines.

For previous discussion see #44110 and https://github.com/Homebrew/homebrew/pull/43973#issuecomment-140591176 (and the following comments).

By default Node.js will now be build with 'small-icu' support, which matches the current distributed Node.js binaries. This means the full `Intl` API is supported, but only the English locale is bundled to reduce file size (see also [nodes Intl/icu docs](https://github.com/nodejs/node#intl-ecma-402-support)). For getting support other locales you can either build `--with-full-icu` or load additional locales at runtime as [described here](https://github.com/nodejs/node-v0.x-archive/wiki/Intl#using-and-customizing-the-small-icu-build).

A test for testing `Intl` support was added too.

### DIFF
--- a/Library/Formula/node.rb
+++ b/Library/Formula/node.rb
@@ -14,16 +14,14 @@ class Node < Formula
   option "with-debug", "Build with debugger hooks"
   option "without-npm", "npm will not be installed"
   option "without-completion", "npm bash completion will not be installed"
+  option "with-full-icu", "Build with full-icu (all locales) instead of small-icu (English only)"
 
   deprecated_option "enable-debug" => "with-debug"
+  deprecated_option "with-icu4c" => "with-full-icu"
 
   depends_on :python => :build if MacOS.version <= :snow_leopard
   depends_on "pkg-config" => :build
   depends_on "openssl" => :optional
-
-  # https://github.com/nodejs/node-v0.x-archive/issues/7919
-  # https://github.com/Homebrew/homebrew/issues/36681
-  depends_on "icu4c" => :optional
 
   fails_with :llvm do
     build 2326
@@ -34,11 +32,24 @@ class Node < Formula
     sha256 "c8b602de5d51f956aa8f9c34d89be38b2df3b7c25ff6588030eb8224b070db27"
   end
 
+  resource "icu4c" do
+    url "https://ssl.icu-project.org/files/icu4c/55.1/icu4c-55_1-src.tgz"
+    mirror "https://fossies.org/linux/misc/icu4c-55_1-src.tgz"
+    version "55.1"
+    sha256 "e16b22cbefdd354bec114541f7849a12f8fc2015320ca5282ee4fd787571457b"
+  end
+
   def install
     args = %W[--prefix=#{prefix} --without-npm]
     args << "--debug" if build.with? "debug"
-    args << "--with-intl=system-icu" if build.with? "icu4c"
     args << "--shared-openssl" if build.with? "openssl"
+    if build.with? "full-icu"
+      args << "--with-intl=full-icu"
+    else
+      args << "--with-intl=small-icu"
+    end
+
+    resource("icu4c").stage buildpath/"deps/icu"
 
     system "./configure", *args
     system "make", "install"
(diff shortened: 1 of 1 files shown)


Library/Formula/node.rb
```diff
@@ -14,17 +14,15 @@ class Node < Formula
   option "with-debug", "Build with debugger hooks"
   option "without-npm", "npm will not be installed"
   option "without-completion", "npm bash completion will not be installed"
+  option "with-full-icu", "Build with full-icu (all locales) instead of small-icu (English only)"
 
   deprecated_option "enable-debug" => "with-debug"
+  deprecated_option "with-icu4c" => "with-full-icu"
```
chrmoritz: Not sure if we should keep this or not:
* argument for keeping this: no side effect for someone who has previously build node `--with-icu4c` and is relying on non English locales (through he could add other locales at runtime to a `small-icu` build as mentioned above)
* argument against keeping this: someone who needs only the English local and has previously build node `--with-icu4c` would have an unnecessary big node installation and would have to build it from source instead of downloading a smaller bottle

chrmoritz:
I think it would be good to backport this to the iojs (and maybe node 0.12) formula too, once this is merged. What's the current plan for the iojs formula since it's now in maintenance? Should it stay in the main repo or should it be moved to node-versions to the other older node releases currently in LTS/maintance.

Also npm 3.x is now stable. What is the plan for this? Upgrading to is asap or stay with the npm 2.x release line for the Node.js 4.x releases and upgrade to npm 3.x with the Node.js 5.0 release to match upstream? I'm personally +0 for waiting for the Node.js 5.0 release for consistency with upstream releases, although I'm currently using npm 3 already.

bcomnes:
It was my understanding that the goal was to stay consistent with the latest node installers, so that would mean waiting till the installer ships npm3 before the formula here has it.  People who know what they are doing are free to `npm i npm@latest -g` which only takes a moment.

As for backporting this to iojs, I can put together some proposal PRs that would share the staging code between the two formula so that there is less duplication.

chrmoritz:
The approve of this PR is to handle `icu4c` consistent with the latest node installers, but in the past the npm in the `node` formula was not alway consistent with upstream releases (thats why it downloads the npm sources as a resource instead of using the npm sources shipped with node). 
Thats why I asking for the maintainers opinions about how to handle npm 3 in homebrew. 
As I'm mentioned above I'm too slightly in favor of waiting for the Node.js 5 release before bumping npm to 3.x.

Porting this to the `iojs` formula should be really straight forward. We just have to use this formula and replace all references to nodejs 4 with the current iojs and add a conflict with node.

DomT4:
About to go catch some :zzz:, but short and sweet answers:
* I'm fine with the option being deprecated as such. If people were using `icu4c` before it seems reasonable they'd want the full ICU experience.
* I'm fine with backporting it to `iojs` although I'd like to move `iojs` to Versions at this point or in the near future, ideally.
* I don't really care about staying in line with the upstream `node` installations and what they ship in terms of `npm`, but I think we should hold off on `npm3` until it does ship with a Node release in this case, unless some security problem arises that won't be fixed in `npm2`.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/ab0a4efcc03b9fbf582314c19a7b118770021b2e. Thank you for your contribution to Homebrew @chrmoritz!

srl295:
Late to the party, but FYI you can pass `--with-icu-source=<path-to-source>.zip` as a parameter to node's `configure`

chrmoritz:
From reading your [configure script](https://github.com/nodejs/node/blob/v4.1.2/configure#L932-L965) the only thing `--with-icu-source=<path-to-source>.zip` would do is to extract the contents of the icu.zip to `deps/icu`. This is exactly the same thing `resource("icu4c").stage buildpath/"deps/icu"` is doing with the only difference thats it's done on homebrews side in ruby. Please correct me if I missed another feature of the `with-icu-source` parameter.

Regarding the ICU version: I think the plan is to stay aligned with the upstream ICU version (bump the ICU resource to 56.1 with the node 4.2 upgrade).
BTW: The homebrew policy to do all resource downloading and checksum checking on homebrews side is the reason why ICU is a resource in this formula.

srl295:
You didn't miss any features. All sounds good.

PR #35839, opened 2015-01-13
poco: added support for --universal flag

cboynton-cweil:
Some of my projects require both 32-bit and 64-bit libraries for Poco. I added support to compile both library architectures and to lipo them together. This only occurs if passing the "--universal" flag.

PR #44148, opened 2015-09-19
poco: 1.6.1, use CMake, allow universal build

mikedld:
This patch bumps version to 1.6.1, adds HEAD, and switches build system to CMake. The latter also allows to build fat binaries and thus close pull request #35839.

Main reason for the change was the need in CMake export definitions, provided by CMake-based builds and installed to lib/cmake/Poco.

Couple of notes:
* I'm unsure whether `cxx11` option has any effect (didn't observe one)
* due to defect in install_name_tool (in at least Xcode 6.4 and 7.0) which leads to fat binaries corruption, fixing install names fails. I see other formulae which use cmake and provide universal option as well, not sure whether that's okay.

### DIFF
--- a/Library/Formula/poco.rb
+++ b/Library/Formula/poco.rb
@@ -1,8 +1,10 @@
 class Poco < Formula
   desc "POCO C++ Libraries"
   homepage "http://pocoproject.org/"
-  url "http://pocoproject.org/releases/poco-1.6.0/poco-1.6.0-all.tar.gz"
-  sha256 "ed1be29ee413141269e7ccee861b11a2992a9f70072dbb28bec31ad432d71cab"
+  url "http://pocoproject.org/releases/poco-1.6.1/poco-1.6.1-all.tar.gz"
+  sha256 "88c88ff0916228d3a011b86c486033dab75e62a27429b5d35736d43527cf5b44"
+
+  head "https://github.com/pocoproject/poco.git", :branch => "develop"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


mikedld:
Hang on, I may need to adjust the change (re HEAD build) if/when https://github.com/pocoproject/poco/pull/946 gets merged.

mikedld:
Updated (POCO pull request merged).

Library/Formula/poco.rb
```diff
@@ -12,18 +14,29 @@ class Poco < Formula
...
-                          "--no-tests"
-    system "make", "install", "CC=#{ENV.cc}", "CXX=#{ENV.cxx}"
+    args = ["-DENABLE_DATA_MYSQL=OFF", "-DENABLE_DATA_ODBC=OFF"]
+
+    if build.universal?
+      ENV.universal_binary
+      args << "-DCMAKE_OSX_ARCHITECTURES=#{Hardware::CPU.universal_archs.as_cmake_arch_flags}"
+    end
+
+    # Fix Foundation library version
```
DomT4: Let's write this as:

```ruby
if build.stable?
  inreplace "Foundation/CMakeLists.txt", "VERSION ${PROJECT_VERSION}", "VERSION ${SHARED_LIBRARY_VERSION}"
end
```
Is the need to do this known upstream?

Library/Formula/poco.rb
```diff
@@ -12,18 +14,29 @@ class Poco < Formula
...
+    inreplace "Foundation/CMakeLists.txt",
+              "VERSION ${PROJECT_VERSION}",
+              "VERSION ${SHARED_LIBRARY_VERSION}" if !build.head?
+
+    mktemp do
```
DomT4: `mkdir "build" do` is usually enough here, but not a huge deal.

Library/Formula/poco.rb
```diff
@@ -12,18 +14,29 @@ class Poco < Formula
...
+      system "cmake", buildpath, *args, *std_cmake_args
```
DomT4: Rather than `*args, *std_cmake_args` make the `std_cmake_args` part of the args, with something like:

```ruby
args = std_cmake_args
args << "-DENABLE_DATA_MYSQL=OFF" << "-DENABLE_DATA_ODBC=OFF"
```

Library/Formula/poco.rb
mikedld: That's what I did first, but there's already "build" directory in the source tree root, so I decided not to invent new name like "_build" or anything. Any other suggestions in this case?

Library/Formula/poco.rb
mikedld: Yes, already merged to develop branch in the aforementioned pull request.

mikedld:
Will do the review changes in a moment. Any thought on the notes in the very first comment?

Library/Formula/poco.rb
DomT4: Huh. Unusual. Alright, this is probably fine then.

mikemcquaid: `mkdir b`

mikedld:
Second by popularity after "build" is "macbuild", so I'm gonna use this directory name instead of `mktemp`.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/2c67be6eb6565b900fca199a952f2b7c674afb75. Thank you for your contribution to Homebrew @mikedld!

PR #44149, opened 2015-09-19
Pixz 1.0.4

justinmayer:


Library/Formula/pixz.rb
```diff
@@ -17,17 +18,23 @@ class Pixz < Formula
...
+      system "a2x", "--doctype", "manpage", "--format", "manpage", "src/pixz.1.asciidoc"
+      man1.install "src/pixz.1"
     end
   end
 
   test do
-    system "#{bin}/pixz"
+    ENV["LC_ALL"] = "en_US.UTF-8"
+    testfile = testpath/"file.txt"
+    filecontent = "foo"
```
bfontaine: I don’t think a variable is necessary here since it’s used only once.

justinmayer:
@bfontaine: I eliminated the variable. All good to merge this?

Library/Formula/pixz.rb
```diff
@@ -17,17 +18,22 @@ class Pixz < Formula
   end
 
   def install
+    ENV["PKG_CONFIG_PATH"] = "#{HOMEBREW_PREFIX}/lib/pkgconfig:#{HOMEBREW_PREFIX}/opt/libarchive/lib/pkgconfig"
```
DomT4: Is there a specific need to override the whole `PKG_CONFIG_PATH`?

Would rather do something like:

```ruby
ENV.prepend_path "PKG_CONFIG_PATH", Formula["libarchive"].opt_lib/"pkgconfig"
```
If it's having trouble finding that for whatever reason.

justinmayer: I will take a look at that. Anything else? Would prefer to make any additional changes all at once.

DomT4: Nothing else that needs fixing/looking at jumps out at me.

justinmayer:
Thanks for reviewing this so quickly. I made the `ENV.prepend_path` change, which appears to do the trick. Clear to merge?

DomT4:
Need to wait for it to trundle through our Jenkins CI process, but once it has, it looks fine to me.

dunn:
Merged in 5197df3d32c4403173af827fdfae4ae3d20dce26, thanks so much!

PR #44151, opened 2015-09-19
shards 0.4.0 (new formula)

genki:
Dependency manager for the Crystal language

### DIFF
--- a/Library/Formula/shards.rb
+++ b/Library/Formula/shards.rb
@@ -1,0 +1,14 @@
+class Shards < Formula
+  desc "Dependency manager for the Crystal language"
+  homepage "https://github.com/ysbaddaden/shards"
+  url "https://github.com/ysbaddaden/shards/archive/v0.4.0.tar.gz"
+  version "0.4.0"
+  sha256 "fff0f3c6562023fd279c80becf3683d2ac922ff097dda12f544f51f4b383ae5c"
+
+  depends_on "manastech/crystal/crystal-lang"
+
+  def install
+    system "make"
+    bin.install "bin/shards"
+  end
+end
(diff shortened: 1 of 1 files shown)


Library/Formula/shards.rb
```diff
@@ -0,0 +1,18 @@
...
+  depends_on "manastech/crystal/crystal-lang"
+  depends_on "libyaml"
+
+  def install
+    system "make"
+    bin.install "bin/shards"
+  end
+
+  test do
+    assert(`#{bin/"shards"} --version`.chomp == "Shards 0.4.0 (2015-09-19)")
```
dunn: This should be written as 
```ruby
assert_equal "Shards 0.4.0 (2015-09-19)", shell_output("#{bin/"shards"} --version").chomp
```

genki: Thanks, I've fixed.

Library/Formula/shards.rb
```diff
@@ -0,0 +1,19 @@
+class Shards < Formula
+  desc "Dependency manager for the Crystal language"
+  homepage "https://github.com/ysbaddaden/shards"
+  url "https://github.com/ysbaddaden/shards/archive/v0.4.0.tar.gz"
+  sha256 "fff0f3c6562023fd279c80becf3683d2ac922ff097dda12f544f51f4b383ae5c"
+
+  depends_on "manastech/crystal/crystal-lang"
```
DomT4: You can't have mandatory cross-tap dependencies inside core Homebrew formula, and we won't accept optional cross-tap dependencies for a tap that Homebrew doesn't have supervision over. If it needs this dependency, you'd be better hosting this in a tap.

genki: ok. I understood.

bfontaine: Crystal has been added in Homebrew; you can use `depends_on "crystal-lang"` now :wink:

bfontaine:
Ping?

bfontaine:

PR #44162, opened 2015-09-20
libspatialite 4.3.0a

nijikon:


DomT4:
The `a` versioning is kind of confusing here, usually indicated an alpha release but upstream has it tagged as `current version is 4.3.0a` so I guess it's stable. Merged in https://github.com/Homebrew/homebrew/commit/826e6ad1d230d4ff14ed323fc38b18d03a265e45. Thank you once again :fireworks:

PR #44164, opened 2015-09-20
gpac: explicitly disable pulseaudio dep to avoid linkage error.

geoff-codes:
`brew install pulseaudio mp4box` should be sufficient to reproduce.
In keeping with the "minimal" approach of the formula, I thought this was the simplest way to take care of it.

On a side note, this formula clearly installs as 'gpac.0.5.1-DEV', not v5.2; that single, solitary "release" is most certainly an anomaly, and the homepage basically that they do not *do* stable releases; using the nightlies is explicitly encouraged. So maybe a candidate for head-only...

### DIFF
--- a/Library/Formula/gpac.rb
+++ b/Library/Formula/gpac.rb
@@ -36,9 +36,11 @@ class Gpac < Formula
   depends_on "openjpeg" => :optional
 
   def install
-    args = ["--disable-wx",
+    # If pulseaudio is linked, the script detects it, but fails to link -lpulse.
+    args = ["--disable-pulseaudio",
+            "--disable-wx",
             "--prefix=#{prefix}",
-            "--mandir=#{man}"]
+            "--mandir=#{man}",]
 
     if build.with? "x11"
       # gpac build system is barely functional


Library/Formula/gpac.rb
```diff
@@ -38,7 +38,10 @@ class Gpac < Formula
   def install
     args = ["--disable-wx",
             "--prefix=#{prefix}",
-            "--mandir=#{man}"]
+            "--mandir=#{man}",]
+
+    # If pulseaudio is installed and linked, the script detects it, but fails to link -lpulse.
+    args << "--disable-pulseaudio"
```
mikemcquaid: Just add it to the array above, thanks.

geoff-codes:
@mikemcquaid Sorry about the delay; done.

Library/Formula/gpac.rb
```diff
@@ -36,9 +36,11 @@ class Gpac < Formula
   depends_on "openjpeg" => :optional
 
   def install
-    args = ["--disable-wx",
+    # If pulseaudio is linked, the script detects it, but fails to link -lpulse.
+    args = ["--disable-pulseaudio",
+            "--disable-wx",
             "--prefix=#{prefix}",
-            "--mandir=#{man}"]
+            "--mandir=#{man}",]
```
mikemcquaid: No point adding a comma here unless the `]` is on another line.

Library/Formula/gpac.rb
```diff
@@ -36,9 +36,11 @@ class Gpac < Formula
   depends_on "openjpeg" => :optional
 
   def install
-    args = ["--disable-wx",
+    # If pulseaudio is linked, the script detects it, but fails to link -lpulse.
```
mikemcquaid: Would be good to add just as a second argument to make the diff nicer. Normally wouldn't care but just given the comment about the bit below.

Library/Formula/gpac.rb
geoff-codes: Note that I added this because,
```
Put a comma after the last item of a multiline array.
```
is now raised with  `brew audit --strict` without this trailing comma. Introduced with ec3ead0a.

So, remove comma and push, even though it fails audit? Or roll back ec3ead0a? Or...?

mikemcquaid: Weird. I'd stick the `]` on a new line. Not `audit --strict` only needs to pass for new formulae.

tdsmith:

PR #44173, opened 2015-09-20
coccinelle 1.0.2

nijikon:
How do install a ocaml resource? We need to add here findlib in order to build.


beurdouche:
@nijikon, I am having exactly the same problem with FStar (#44341).

PR #44181, opened 2015-09-20
xlslib: use https homepage

vszakats:
It's redirected to http if not logged it, but https is used already in 12 other formulae in same scenario.


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/19e74b74b1455faedfd8827427ad138fb1c26b49. Thanks again!

PR #44186, opened 2015-09-20
swiftgen 0.4.4 (new formula)

AliSoftware:
See also previous PR attempt for older version 0.4.1 in #42983 which was waiting for Xcode 7.0

`brew audit --strict swiftgen` + `brew test swiftgen` run locally and seems ok

### DIFF
--- a/Library/Formula/swiftgen.rb
+++ b/Library/Formula/swiftgen.rb
@@ -1,0 +1,41 @@
+class Swiftgen < Formula
+  desc "Collection of Swift tools to generate Swift code"
+  homepage "https://github.com/AliSoftware/SwiftGen"
+  url "https://github.com/AliSoftware/SwiftGen/archive/0.4.4.tar.gz"
+  sha256 "429e50b33d1cd3e34fa81b5becff865ce5663284380e91602783bfa109c96b83"
+  head "https://github.com/AliSoftware/SwiftGen.git"
+
+  depends_on :xcode => "7.0"
+
+  def install
+    # Install all the binaries
+    ENV["SWIFTGEN_VERSION"] = "#{version} (via Homebrew)"
+    rake "install[#{bin},/]"
+
+    # Install the various test fixtures in pkgshare needed for `brew test`
+    pkgshare.install(
+      "Tests/Assets/fixtures/Images.xcassets",
+      "Tests/Colors/fixtures/colors.txt",
+      "Tests/L10n/fixtures/Localizable.strings",
+      "Tests/Storyboard/fixtures/Message.storyboard",
+      "Tests/Assets/expected/FileDefaults.swift.out" => "assets.swift",
+      "Tests/Colors/expected/FileDefaults.swift.out" => "colors.swift",
+      "Tests/L10n/expected/FileWithDefaults.swift.out" => "l10n.swift",
+      "Tests/Storyboard/expected/MessageWithDefaults.swift.out" => "storyboard.swift",
+    )
+  end
+
+  test do
+    output = shell_output("#{bin}/swiftgen-assets #{pkgshare}/Images.xcassets").strip
+    assert_equal output, (pkgshare/"assets.swift").read.strip, "swiftgen-assets failed"
+
+    output = shell_output("#{bin}/swiftgen-colors #{pkgshare}/colors.txt").strip
+    assert_equal output, (pkgshare/"colors.swift").read.strip, "swiftgen-colors failed"
+
+    output = shell_output("#{bin}/swiftgen-l10n #{pkgshare}/Localizable.strings").strip
+    assert_equal output, (pkgshare/"l10n.swift").read.strip, "swiftgen-l10n failed"
+
+    output = shell_output("#{bin}/swiftgen-storyboard #{pkgshare}/Message.storyboard").strip
+    assert_equal output, (pkgshare/"storyboard.swift").read.strip, "swiftgen-storyboard failed"
+  end
+end


Library/Formula/swiftgen.rb
```diff
@@ -0,0 +1,41 @@
...
+      "Tests/Colors/expected/FileDefaults.swift.out" => "colors.swift",
+      "Tests/L10n/fixtures/Localizable.strings" => "Localizable.strings",
+      "Tests/L10n/expected/FileWithDefaults.swift.out" => "l10n.swift",
+      "Tests/Storyboard/fixtures/Message.storyboard" => "Message.storyboard",
+      "Tests/Storyboard/expected/MessageWithDefaults.swift.out" => "storyboard.swift",
+    )
+  end
+
+  test do
+    output = `#{bin}/swiftgen-assets #{pkgshare}/Images.xcassets`.strip
```
DomT4: Can we use `shell_output()` or `pipe_output()` rather than backtics at all?

Library/Formula/swiftgen.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    # Install the various test fixtures in pkgshare needed for `brew test`
+    pkgshare.install(
+      "Tests/Assets/fixtures/Images.xcassets" => "Images.xcassets",
+      "Tests/Assets/expected/FileDefaults.swift.out" => "assets.swift",
+      "Tests/Colors/fixtures/colors.txt" => "colors.txt",
+      "Tests/Colors/expected/FileDefaults.swift.out" => "colors.swift",
+      "Tests/L10n/fixtures/Localizable.strings" => "Localizable.strings",
+      "Tests/L10n/expected/FileWithDefaults.swift.out" => "l10n.swift",
+      "Tests/Storyboard/fixtures/Message.storyboard" => "Message.storyboard",
```
DomT4: If you're not changing the name you don't need the `=> "Message.storyboard"` bit.

Library/Formula/swiftgen.rb
```diff
@@ -0,0 +1,41 @@
...
+  def install
+    # Install all the binaries
+    ENV["SWIFTGEN_VERSION"] = "#{version} (via Homebrew)"
+    rake "install[#{bin},/]"
+
+    # Install the various test fixtures in pkgshare needed for `brew test`
+    pkgshare.install(
+      "Tests/Assets/fixtures/Images.xcassets" => "Images.xcassets",
+      "Tests/Assets/expected/FileDefaults.swift.out" => "assets.swift",
+      "Tests/Colors/fixtures/colors.txt" => "colors.txt",
```
DomT4: Same comment here as below re `=>`.

Library/Formula/swiftgen.rb
```diff
@@ -0,0 +1,41 @@
...
+  depends_on :xcode => "7.0"
+
+  def install
+    # Install all the binaries
+    ENV["SWIFTGEN_VERSION"] = "#{version} (via Homebrew)"
+    rake "install[#{bin},/]"
+
+    # Install the various test fixtures in pkgshare needed for `brew test`
+    pkgshare.install(
+      "Tests/Assets/fixtures/Images.xcassets" => "Images.xcassets",
```
DomT4: Same comment here as below re `=>`.

Library/Formula/swiftgen.rb
```diff
@@ -0,0 +1,41 @@
...
+      "Tests/Assets/expected/FileDefaults.swift.out" => "assets.swift",
+      "Tests/Colors/fixtures/colors.txt" => "colors.txt",
+      "Tests/Colors/expected/FileDefaults.swift.out" => "colors.swift",
+      "Tests/L10n/fixtures/Localizable.strings" => "Localizable.strings",
```
DomT4: Same comment here as below re `=>`.

Library/Formula/swiftgen.rb
AliSoftware: I want to ensure that the file is at the root of `pkgshare`, and the relative path is flatten, isn't it necessary in such case?

DomT4: It won't install the whole folder tree, no, just the requested/expected file itself.

Library/Formula/swiftgen.rb
AliSoftware: I have no problem changing it to use `shell_output()`, but fmi what's wrong with backticks?

DomT4: We dislike using them, unless all else fails. @mikemcquaid probably knows whether that's stylistic or functional.

Library/Formula/swiftgen.rb
AliSoftware: Ok great.

I didn't know we could mix keys with and without `=>` in a ruby hash. Are you sure that rule can be applied if some of my files have new names and some don't?

For example when trying in `pry` I have a syntax error:
```
[1] pry(main)> h = { 'a' => 'AA', 'b' => 'BBB', 'c', 'd' => 'DDD' }
SyntaxError: unexpected =>, expecting end-of-input
h = { 'a' => 'AA', 'b' => 'BBB', 'c', 'd' => 'DDD' }
```

Or maybe in such cases you prefer that I use `=> nil` for non-renamed files?

DomT4: Works fine here:

```ruby
    pkgshare.install(
      "Tests/Assets/fixtures/Images.xcassets",
      "Tests/Colors/fixtures/colors.txt",
      "Tests/L10n/fixtures/Localizable.strings",
      "Tests/Storyboard/fixtures/Message.storyboard",
      "Tests/Assets/expected/FileDefaults.swift.out" => "assets.swift",
      "Tests/Colors/expected/FileDefaults.swift.out" => "colors.swift",
      "Tests/L10n/expected/FileWithDefaults.swift.out" => "l10n.swift",
      "Tests/Storyboard/expected/MessageWithDefaults.swift.out" => "storyboard.swift",
    )
```

```
~> ls -1 /usr/local/opt/swiftgen/share/swiftgen
Images.xcassets
Localizable.strings
Message.storyboard
assets.swift
colors.swift
colors.txt
l10n.swift
storyboard.swift
```

Library/Formula/swiftgen.rb
AliSoftware: Ok just discovered `shell_output` is a function of Homebrew and not standard Ruby, that's why I didn't know about it.

As seen in Homebrew's source, it uses backticks internally, but also prints the command before and assert the exit status after, that's probably why you prefer to use that wrapper instead of using backticks directly in formulas. Will change it in a sec.

Library/Formula/swiftgen.rb
AliSoftware: Ah, that's because of how your `pkgsare.install` is implemented in Ruby — and it only works if you group the keys having a value in the end. _(in practice this passes 5 arguments to `pkgshare.install`, the first 4 ones being the un-renamed path strings, the last one being the hash with 4 keys and values; and the implementation is done in a way that can handle it)_

AliSoftware:
@DomT4 all fixed, thx for your feedback!

Library/Formula/swiftgen.rb
DomT4: Heh, and now you know which parts of the Homebrew source code I haven't really paid attention to in the last year :wink:. It is indeed a customised function; we have quite a few tweaks that make life easier for us & contributors.

mikemcquaid: Yeh, we also prefer it because we split arguments meaning it's harder to inject arbitrary commands.

AliSoftware: @mikemcquaid What do you mean by "we split arguments"? I don't see any argument splitting in [this implementation](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/formula_assertions.rb#L27-L32), am I missing something?

mikemcquaid: @AliSoftware my mistake, I was confused about another proposed implementation.

DomT4:
I'm happy enough. Merged in https://github.com/Homebrew/homebrew/commit/90822cbf436ade1223a811755c33462872adfbf5. Thank you for your contribution to Homebrew @AliSoftware!

PR #44192, opened 2015-09-20
aide 0.15.1 (new formula)

jrwittmann:


Library/Formula/aide.rb
```diff
@@ -0,0 +1,27 @@
+class Aide < Formula
+  desc "File and directory integrity checker"
+  homepage "http://aide.sourceforge.net"
+  url "https://downloads.sourceforge.net/project/aide/devel/0.16a2/aide-0.16a2.tar.gz"
```
DomT4: Please use the stable version rather than the development version. You can add the development version in a `devel do` block if you'd like to.

Library/Formula/aide.rb
```diff
@@ -0,0 +1,27 @@
...
+  sha256 "b52451816bc85409ea09dc612e32823336f78438afd28248c252912ea8b91b87"
+  head "http://git.code.sf.net/p/aide/code.git"
+
+  depends_on "automake" => :build
```
DomT4: If these aren't required for building outside of the `head` they should be tucked inside a `head do` block.

Library/Formula/aide.rb
```diff
@@ -0,0 +1,27 @@
...
+                          "--disable-static",
+                          "--with-curl",
+                          "--with-zlib",
+                          "--prefix=#{prefix}"
+
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/aide", "--version"
```
DomT4: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. e.g. write out a `relx.config` and do something with it. Thanks!

jrwittmann:
@DomT4
Thanks for the feedback! I’ve implemented your suggestions to the formula.

Concerning the test block:
Without a user-defined config file there isn’t much left I can test. So I’ve tried to provide an example config file but unfortunately that hasn’t worked out yet but I’ll keep trying.

Library/Formula/aide.rb
```diff
@@ -0,0 +1,38 @@
+class Aide < Formula
+  desc "File and directory integrity checker"
+  homepage "http://aide.sourceforge.net"
+
+  stable do
```
DomT4: You don't need the `stable do` block. You're fine to do:

```ruby
desc "Example"
homepage "https://example.com"
sha256 "13550350a8681c84c861aac2e5b440161c2b33a3e4f302ac680ca5b686de48de"

devel do
  url "https://example.com/exampleb.tar.gz"
  sha256 "2554c5173b60d394e3880bcc7eafa03d1f2f2295e2786c86599ffb2806c088f6"
end

head do
  url "https://example.com/example.git"

end
```
For example.

DomT4:
Highlighted another nit but otherwise looks better. Thanks for making the changes!

> Without a user-defined config file there isn’t much left I can test. So I’ve tried to provide an example config file but unfortunately that hasn’t worked out yet but I’ll keep trying.

You can only write to the testpath in the test, per our sandbox. If it helps, the `HOME` environmental variable is set to the testpath for the duration of the test, and a lot of things will take arguments as to where the config file is. So you could do something like:
```ruby
(testpath/"example.conf").write <<-EOS.undent
  variable: example
  variable2: example2
EOS

system bin/"example", "-c", "example.conf"
```
Or such.

Library/Formula/aide.rb
```diff
@@ -0,0 +1,38 @@
...
+    sha256 "b52451816bc85409ea09dc612e32823336f78438afd28248c252912ea8b91b87"
+  end
+
+  head do
+    url "http://git.code.sf.net/p/aide/code.git"
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
+  end
+
+  depends_on "libgcrypt" => :build
```
DomT4: Looks like this is runtime as well:

```
==> /usr/local/Cellar/aide/0.15.1/bin/aide --version
dyld: Library not loaded: /usr/local/lib/libgcrypt.20.dylib
  Referenced from: /usr/local/Cellar/aide/0.15.1/bin/aide
  Reason: image not found
```

jrwittmann:
I’m a complete newbie to this so thank you very much for your help!

Regarding the dyld runtime problem I’m not sure if this is the right way to deal with it:

``` ruby
test do
  ENV.append “DYLD_LIBRARY_PATH”, “#{lib}”
  […]
end
```

DomT4:
> Regarding the dyld runtime problem I’m not sure if this is the right way to deal with it:

Afraid not. It just means the `dylib` it is expecting to find isn't available, which is happening because our CI removes `:build` dependencies after install and before testing to check those `:build` dependencies aren't actually runtime. If you make the dependency for `libgcrypt` just:
```ruby
depends_on "libgcrypt"
```
It should pass.

jrwittmann:
Thanks again, now I get it. That fix is much easier than I thought.

DomT4:
We want to add a `"--sysconfdir=#{etc}"` line to `./configure` here as well. It ensures that it expects the config files to be placed in `HOMEBREW_PREFIX/"etc"` and that way people's configurations won't get wiped out every update/reinstall/etc.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/b40bd9f42afcc48d083a5003597eb18a26c04709. Thank you for your contribution to Homebrew @jrwittmann!

PR #44201, opened 2015-09-21
Upgrade SSL to 1.0_2d

LaurentFough:
Upgrade OpenSSL requirement to 1.0_2d

### DIFF
--- a/Library/Formula/sslyze.rb
+++ b/Library/Formula/sslyze.rb
@@ -22,8 +22,8 @@ class Sslyze < Formula
   end
 
   resource "openssl" do
-    url "https://www.openssl.org/source/old/1.0.2/openssl-1.0.2a.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Will this work & not cause runtime/build problems @nabla-c0d3?

nabla-c0d3:
It will cause build problems as setup.py expects 1.0.2a. I am about to release Nassl and SSLyze 0.12 which will support OpenSSL 1.0.2.d so let's wait until that happens.

DomT4:
Cool. Thanks for the reply @nabla-c0d3 :+1:. Closing here, but happy to take an update PR when the new version is cut if you're still interested then @LaurentFough.

LaurentFough:
ditto @nabla-c0d3: Thanks.
Will wait on the PR.

PR #44206, opened 2015-09-21
Update npm to v3.3.3

### LINKED ISSUE #44202: Update NPM to 3.0
Last week NPM 3.0 was released. Any chance we can get the Node formula to use that version instead of the 2.0? I would create a PR myself but I don't think my knowledge about this is enough.

TrevorSayre:
Update npm to the [latest stable](https://docs.npmjs.com/) v3.3.3
Fixes #44202

### DIFF
--- a/Library/Formula/node.rb
+++ b/Library/Formula/node.rb
@@ -29,8 +29,8 @@ class Node < Formula
   end
 
   resource "npm" do
-    url "https://registry.npmjs.org/npm/-/npm-2.14.4.tgz"
-    sha256 "c8b602de5d51f956aa8f9c34d89be38b2df3b7c25ff6588030eb8224b070db27"
+    url "https://registry.npmjs.org/npm/-/npm-3.3.3.tgz"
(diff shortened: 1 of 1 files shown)


vinkla:
:+1:

DomT4:
Whilst I'm happy to jump over upstream Node for minor releases, I'm not going to enforce the jump to major on every single Homebrew/Node user until [Node upstream do the same thing](https://github.com/nodejs/node/pull/2958).

I'll dump a reasonable sum of money on `npm3` breaking some obscure package somewhere & having to deal with the whole "Homebrew breaks Node" fracas again :smile_cat:.

If people want to `npm i npm@latest -g` locally by all means nobody's going to stop you, but we're going to track Node on which *major* `npm` version to use, with the caveat that if a security issue pops up in `npm2` and upstream doesn't want to fix it, we make the jump.

For the sake of maintenance both here and upstream at Node, I'd like Homebrew to mimic the end-user experience between us and their binary installer as much as possible.

PR #44209, opened 2015-09-21
test-bot: auto truncate logs on travis

xu-cheng:
I'm not sure whether 3MB is a good line. Any suggestion?

cc @mikemcquaid @tdsmith

### DIFF
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -142,6 +142,9 @@ module Homebrew
       end
 
       verbose = ARGV.verbose?
+      travis = !ENV["TRAVIS"].nil?
+      @@log_length ||= 0
+
       @output = ""
       working_dir = Pathname.new(@command.first == "git" ? @repository : Dir.pwd)
       read, write = IO.pipe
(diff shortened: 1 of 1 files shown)


mikemcquaid:
The problem here is if the job is building more than one thing with long logs then this will exceed the limit. I was thinking that it might be worth increasing the number of lines we print out here: https://github.com/Homebrew/homebrew/blob/22092d86a3225ba0fea35c060ae8a362bb016dd0/Library/Homebrew/formula.rb#L1308 (5 is rarely enough to be useful) and maybe allow it to be bumped to even ~250 or something and just use that instead of `--verbose`. Relatedly `--verbose` seemed to cause some issues with building Boost 1.58 on El Capitan so the bottle had to be built without. I'm wondering if we get rid of it entirely from CI; it's almost never necessary to have more than a couple of hundred lines. Note that this all applies to Jenkins and even individual users compilation too.

xu-cheng:
That seems to be more reasonable. Also, we may create an env variable to adjust it dynamically, so a) set different value for CI and end-users, b) adjust it for specific formula building

mikemcquaid:
Yep, makes sense. As a starting point adding the environment variable and just using that for CI and removing `--verbose` from the `test-bot` seems like a good first step. Might even speed up Jenkins builds, too.

PR #44217, opened 2015-09-21
More failure logs

mikemcquaid:
- formula: increase fail log lines, allow config. The default is almost never useful. 15 seems like a good medium as it'll not fill a 80x24 default but provides a bit more context. Also allow it to be overriden for developers and `test-bot`.
- test-bot: configure log lines instead of verbose. The `--verbose` seems to cause some issues with compiling software like `boost` under Xcode 7.

CC @xu-cheng @tdsmith

### DIFF
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -18,7 +18,6 @@
 # --dry-run:       Just print commands, don't run them.
 # --fail-fast:     Immediately exit on a failing step.
 # --verbose:       Print out all logs in realtime
-# --no-verbose-install: Don't run `brew install` with `--verbose`.
 #
 # --ci-master:           Shortcut for Homebrew master branch CI options.
 # --ci-pr:               Shortcut for Homebrew pull request CI options.
@@ -532,7 +531,6 @@ module Homebrew
       test "brew", "fetch", "--retry", *formula_fetch_options
       test "brew", "uninstall", "--force", canonical_formula_name if formula.installed?
       install_args = []
-      install_args << "--verbose" unless ARGV.include? "--no-verbose-install"
       install_args << "--build-bottle" unless ARGV.include? "--no-bottle"
       install_args << "--HEAD" if ARGV.include? "--HEAD"
 
@@ -849,6 +847,7 @@ module Homebrew
     ENV["HOMEBREW_DEVELOPER"] = "1"
     ENV["HOMEBREW_SANDBOX"] = "1"
     ENV["HOMEBREW_NO_EMOJI"] = "1"
+    ENV["HOMEBREW_FAIL_LOG_LINES"] = "150"
     ARGV << "--verbose" if ENV["TRAVIS"]
 
     if ARGV.include?("--ci-master") || ARGV.include?("--ci-pr") \
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -141,9 +141,11 @@ module Homebrew
       end
 
       verbose = ARGV.verbose?
+      travis = !ENV["TRAVIS"].nil?
       @output = ""
       working_dir = Pathname.new(@command.first == "git" ? @repository : Dir.pwd)
       read, write = IO.pipe
+      print_thread = nil
 
       begin
         pid = fork do
@@ -154,6 +156,14 @@ module Homebrew
           working_dir.cd { exec(*@command) }
         end
         write.close
+        if travis
+          print_thread = Thread.new do
+            while true do
+              sleep 60
+              printf "."
+            end
+          end
+        end
         while line = read.gets
           puts line if verbose
           @output += line
@@ -163,6 +173,10 @@ module Homebrew
       end
 
       Process.wait(pid)
+      if travis
+        print_thread.exit
+        puts
+      end
       @end_time = Time.now
       @status = $?.success? ? :passed : :failed
       puts_result
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
   - export TRAVIS_BUILD_DIR="/usr/local"
   - sudo chown -R $USER /usr/local
   - cd /usr/local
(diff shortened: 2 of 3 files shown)


xu-cheng:
Some suggestions:
* We may choose a better name than `HOMEBREW_INSTALL_FAIL_LOG_LINES` as it's used in `install`, `postinstall` and `test`.
* We should document it in manpage.

Otherwise, :+1:

mikemcquaid:
> We may choose a better name than HOMEBREW_INSTALL_FAIL_LOG_LINES as it's used in install, postinstall and test.

Any suggestions? `HOMEBREW_FAIL_LOG_LINES`?

mikemcquaid:
> We should document it in manpage.

I think with new things like this we should hold off documenting until we've decided how we want the functionality to work.

mikemcquaid:
Went for `HOMEBREW_FAIL_LOG_LINES`.

tdsmith:
> The default is almost never useful. 15 seems like a good medium as it'll not fill a 80x24 default but provides a bit more context

I remember someone suggesting we default to zero and print a message about ~/Library/Logs/Homebrew and/or gist-logs instead; this would be an opportunity to do that. I'm happy either way.

UniqMartin:
> I remember someone suggesting we default to zero and print a message about ~/Library/Logs/Homebrew and/or gist-logs instead; this would be an opportunity to do that.

:+1: In my experience the error that causes the build failure is almost always so far up, that printing 15 lines wouldn't have helped either. In most cases this is due to build parallelism and make recursion.

mikemcquaid:
> I remember someone suggesting we default to zero and print a message about ~/Library/Logs/Homebrew and/or gist-logs instead; this would be an opportunity to do that. I'm happy either way.

Personally I almost never find the output of `gist-logs` more useful than `brew install -v` but maybe I'm unusual, there.

> In my experience the error that causes the build failure is almost always so far up, that printing 15 lines wouldn't have helped either. In most cases this is due to build parallelism and make recursion.

I'm open to it being more than 15 lines but I do think 15 lines is definitely no worse than 5.

UniqMartin:
> I'm open to it being more than 15 lines but I do think 15 lines is definitely no worse than 5.

No doubt 15 lines is an improvement over 5 and may be sufficient in simple cases. I think the default shouldn't be much higher considering Homebrew tends to have rather terse console output. (The different limit of 150 for the bot sounds like a good idea, too.)

I believe printing `From #{logfn}:` directly above those lines would be useful to those wishing to investigate locally. That way one could instantly use the path to have a closer look at the log. (And if the limit happens to be set to zero, maybe print `See #{logfn} for details.`)

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -532,7 +531,6 @@ def formula(formula_name)
       test "brew", "fetch", "--retry", *formula_fetch_options
       test "brew", "uninstall", "--force", canonical_formula_name if formula.installed?
       install_args = []
-      install_args << "--verbose" unless ARGV.include? "--no-verbose-install"
```
xu-cheng: I almost forget that we still need to print something for travis. Otherwise, it will be killed if no output in long time.

mikemcquaid: Yep. Mind if I steal your `...` approach from your PR?

xu-cheng: Go ahead.

mikemcquaid: @xu-cheng Done.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -155,6 +156,14 @@ def run
...
         end
         write.close
+        if travis
+          print_thread = Thread.new do
+            while true do
+              printf "."
+              sleep 60
+            end
+          end
+        end
```
xu-cheng: I don't think we should do this for every test. We only need it for `brew install`.

xu-cheng: How about:
```ruby
if travis
  wait_period  = 10
  begin
    s = read.read_nonblock(1)
  rescue IO::WaitReadable
    unless Process.waitpid(pid, Process::WNOHANG)
      print "."
      sleep wait_period
      wait_period = 60
      retry
    end
  else
    @output += s
    print s
  end
end
while line = read.gets
  puts line if verbose
 @output += line
end
```

mikemcquaid: What's the advantage to that? Seems a lot more complex. Having a e.g. one minute wait seems to make sense though (could just reverse the sleep and print). Creating/destroying a thread is easy enough to do.

xu-cheng: Ok, please ignore my proposed implementation above. I was mistaking assuming `brew install` won't print anything until error. 

What I'm thinking I s I don't like print random dot if there is a output. And using thread to print dot could hang up Travis if building is truly freeze. 

How about we implementing this inside `Formula#system`? So it will only print dot when it should be.

mikemcquaid: Putting it in `Formula#system` makes sense to me. I think printing a dot every minute is probably a good idea for there anyway as we get people posting quite often that they think the process has frozen when it hasn't. Would you mind having a go at that (I suspect you'll disagree with my implementation more than I'll disagree with yours :wink:)

tdsmith:
Here's an example of a travis build failing because the bottle relocation check gave too much output: https://travis-ci.org/Homebrew/homebrew/jobs/81536337

mikemcquaid:
@tdsmith Thanks for the heads up, I think I'll display a maximum number of results. I think we could/should make a tracking issue working out everything like this that needs tweaks.

Library/Homebrew/cmd/test-bot.rb
xu-cheng: #44264

mikemcquaid:
Closing in favour of https://github.com/Homebrew/homebrew/pull/44264.

@tdsmith Done in https://github.com/Homebrew/homebrew/pull/44266

PR #44219, opened 2015-09-21
Allow building libepoxy as universal.

waywardmonkeys:
This allows building libepoxy as a universal binary. This is useful when trying to build other things as universal that depend upon this.


bfontaine:
Thank you for your contribution @waywardmonkeys; we appreciate it! :smile:

PR #44221, opened 2015-09-21
glib 2.46.0

tschoonj:


tschoonj:
@DomT4 @mikemcquaid 

the tests appear to fail for Yosemite and Mavericks when testing fourstore due to some writing operation not permitted in the sandbox. Any idea how I can solve this?

In the meantime, could you merge this one in already? My PR #44235 depends on this one (should have started with one large PR immediately...)

Thanks!

tschoonj:
Thanks!

PR #44231, opened 2015-09-21
gdk-pixbuf 2.32.0

tschoonj:


tschoonj:
don't merge this one in yet, I found some changes in this version that are affecting other formulas adversely...

tschoonj:
problems fixed

Library/Formula/librsvg.rb
```diff
@@ -3,9 +3,9 @@ class Librsvg < Formula
   homepage "https://live.gnome.org/LibRsvg"
   url "https://download.gnome.org/sources/librsvg/2.40/librsvg-2.40.10.tar.xz"
   sha256 "965c807438ce90b204e930ff80c92eba1606a2f6fd5ccfd09335c99896dd3479"
+  revision 1
 
   bottle do
-    sha256 "05e36f46fa553bef32ce12fac9b2aedb29ef0ada303a76417da3d7be42615d46" => :el_capitan
```
DomT4: You can leave it this time, but don't worry about the bottles. We're building for Mav/Yos/Cap these days so when bottles are pulled ML will vanish & the others will move down in the list one place. It doesn't leave "dead" bottles in place.

Library/Formula/gdk-pixbuf.rb
```diff
@@ -21,6 +21,9 @@ class GdkPixbuf < Formula
   depends_on "libpng"
   depends_on "gobject-introspection"
 
+  # disable the default relocatable library setting
+  patch :DATA
```
DomT4: Would be perhaps nice to make this an `inreplace` if we're going to have to carry it indefinitely.

tschoonj: I am going to report this to the developer as I am starting to consider it as a bug: he makes the following assumption in configure.ac:

````
Enable replacing the build-time prefix in the loaders
cache with the installation prefix on this machine
for relocatable packages such as Windows and OS X
applications and linux bundles
````

Now IMHO this should certainly not be the default behavior on Mac OS X as most people that will install gdk-pixbuf will do it via Homebrew or MacPorts where this is certainly not desirable. The problem is that the developer did not turn this into an option, but instead hardcoded this :-(

Comment (DomT4) [inline Library/Formula/gdk-pixbuf.rb]: Alright. I don't mind carrying a short-term patch, but anything reasonably simple like this that can be `inreplace`'d should be if it's going to stick around for a while. The less in-formula patches the better; they've a habit of breaking :cry:

**Comment by tschoonj** [comment]
@DomT4 any idea what is causing `xz` not to be found?

````
       ==> Installing dependencies for goffice: libgsf, libcroco, librsvg, pcre
==> Installing goffice dependency: libgsf
==> Downloading https://homebrew.bintray.com/bottles/libgsf-1.14.34.yosemite.bottle.tar.gz
Already downloaded: /Library/Caches/Homebrew/libgsf-1.14.34.yosemite.bottle.tar.gz
==> Pouring libgsf-1.14.34.yosemite.bottle.tar.gz
/usr/local/Cellar/libgsf/1.14.34: 152 files, 2.7M
==> Installing goffice dependency: libcroco
==> Downloading https://homebrew.bintray.com/bottles/libcroco-0.6.8.yosemite.bottle.tar.gz
Already downloaded: /Library/Caches/Homebrew/libcroco-0.6.8.yosemite.bottle.tar.gz
==> Pouring libcroco-0.6.8.yosemite.bottle.tar.gz
/usr/local/Cellar/libcroco/0.6.8: 76 files, 2.0M
==> Installing goffice dependency: librsvg
==> Downloading https://homebrew.bintray.com/bottles/librsvg-2.40.10_1.yosemite.bottle.tar.gz
Error: Failed to download resource "librsvg"
Download failed: https://homebrew.bintray.com/bottles/librsvg-2.40.10_1.yosemite.bottle.tar.gz
==> Using the sandbox
Error: /usr/local/opt/xz not present or broken
Please reinstall xz. Sorry :(
````

**Comment by tschoonj** [comment]
@mikemcquaid Could you have a look at this? Is this the buildbot behaving badly or is there some deeper underlying cause?

Thanks

**Comment by DomT4** [comment]
You aren't being ignored :wink:. Were just discussing this in the team chat. It's an internal Homebrew error that we haven't yet been able to resolve.

**Comment by tschoonj** [comment]
Ok :smile:

**Comment by staticfloat** [comment]
May I ask why you don't want a relocatable bottle?  Unless I'm mistaken in understanding what "relocatable" means in this case, you almost always want a relocatable binary so that users can install the bottle in places other than `/usr/local`.

**Comment by DomT4** [comment]
@staticfloat I'm not quite sure what you mean? What triggered the question about relocatable bottles here?

**Comment by staticfloat** [comment]
The patch that states its purpose is to disable the following:

    Enable replacing the build-time prefix in the loaders
    cache with the installation prefix on this machine
    for relocatable packages such as Windows and OS X
    applications and linux bundles

**Comment by DomT4** [comment]
I don't know if that'll impact the bottles themselves (A lot of GTK related stuff hardcodes the prefix in a billion places), but it's an interesting point though. I'll let Tom counter.

**Comment by staticfloat** [comment]
The reason I have an interest is that I maintain [a tap](https://github.com/staticfloat/homebrew-juliadeps) that hosts bottles for the [Julia language](http://julialang.org), where we package things like `gtk+3` and force them to be relocatable through patches and environment variables that we [can set on the julia side](https://github.com/JuliaLang/Gtk.jl/blob/64caf82d735a69b8f54da8ad24ed4d0cd9b4e69c/deps/build.jl#L50-L51).  To that end, anything that increases the likelihood of packages being relocatable decreases the amount of work we have to do; and increases overall flexibility.  :)

File: Library/Formula/gdk-pixbuf.rb (hunk 2 above)
Comment (mikemcquaid) [inline Library/Formula/gdk-pixbuf.rb]: What's the actual downside for us?

Comment (mikemcquaid) [inline Library/Formula/gdk-pixbuf.rb]: (of leaving this unpatched).

**Comment by tschoonj** [comment]
Good morning all

First of all, I would like to say that my patch ensures that gdk-pixbuf continues to behave in exactly the same way as the preceding versions did. The relocatable feature was added only in this release.

The problem at hand can be simple stated as follows: Turning relocatable on for gdk-pixbuf would break every single package that relies on it.

To understand why, all you need to do is look at the full path that is used by gdk-pixbuf to determine the location of `loaders.cache` when relocation is on:

````
<path-to-executable>/../lib/gdk-pixbuf-2.0/2.10.0/loaders.cache
````

So in other words, the path is relative to the location in the filesystem of the executable that uses `gdk-pixbuf`, either through direct linking, or through plug-ins.

In other words, since path-to-executable is usually somewhere in `#{prefix}/bin` (prefix being unique for a particular version of a particular formula) this never works, resulting in a long list of error messages and no icons in your app (if it doesn't crash on startup!).

Interestingly, even MacPorts has put a [patch](https://trac.macports.org/browser/trunk/dports/graphics/gdk-pixbuf2/files/patch-configure.ac.diff) in place that turns off relocations by default, while giving an option to turn it on. You may wonder why they did this, since after all in their usage case this path-to-executable thing would actually since they work with a single installation prefix (everything in `/opt/local`).

The reason that they do put this measure in place is to avoid breaking python packages using gtk3 through pygobject: in this case the path-to-executable is the python executable, which is actually hidden somewhere deep inside the Python Framework. For instance if run `meld` from homebrew-x11, a GNOME package written in python using gtk3, when gdk-pixbuf 2.32.0 is compiled without my patch I get a huge list of errors immediately, with this one first:

````
(meld:33141): GdkPixbuf-WARNING **: Cannot open pixbuf loader module file '/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/../lib/gdk-pixbuf-2.0/2.10.0/loaders.cache': No such file or directory

This likely means that your installation is broken.
Try running the command
  gdk-pixbuf-query-loaders > /usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/../lib/gdk-pixbuf-2.0/2.10.0/loaders.cache
to make things work again for the time being.
````

Regardless of which python version you use (Mac OS X provided or Homebrew installed), this will clearly never work, which is why you need the patch.

It gets even worse: if you turn on relocation, and you plan on copying your gdk-pixbuf installation into an app bundle you want to distribute it will probably still not work since the convention on Mac OS X is to put such files in `Contents/Resources`, not in `Contents`, which would be required for the relocation to work.

It is true that turning off relocatable will hardcode the path to loaders.cache in the gdk-pixbuf making the bottle useless for anyone with `brew --prefix` != `/usr/local`, but that should not be a big problem since gtk+3 does this too, and this formula is the main reason why people install gdk-pixbuf in the first place.
I think that if you have a non-default Homebrew prefix, you should never use bottles anyway as it is guaranteed to break things.

The only way to have a relocatable gdk-pixbuf that avoids the aforementioned problems would be to ask users to set `GDK_PIXBUF_MODULE_FILE ` (and possible also `GDK_PIXBUF_MODULEDIR`) in their bashrc, since this environment variable takes precedence in locating loaders.cache. But I really feel we should avoid this, and rely on hardcoding instead.

By the way the caveat in the gdk-pixbuf formula is useless: you don't need to set this environment variable at all in order to have it working properly: this is accomplished by the line:

````
ENV.append_to_cflags "-DGDK_PIXBUF_LIBDIR=\\\"#{HOMEBREW_PREFIX}/lib\\\""
````

which ensures that both loaders.cache and the modules themselves will be found properly.

@staticfloat for the same reason I believe that your PR #44262 is not useful, as I recommend to remove the caveat altogether.

Perhaps I could introduce MacPort's patch here, making it relocatable an option (though off by default)?

**Comment by mikemcquaid** [comment]
Thanks for the details explanation. My main concern with patches like this is just that they are submitted upstream. If upstream refuses to accept them then (unlike MacPorts) it's more important for us to figure out how to run unpatched.

**Comment by tschoonj** [comment]
I will file a bug over this on bugzilla, the way it is now implemented makes gdk-pixbuf completely useless for the vast majority of Homebrew users (and also a lot of MacPorts/Fink users).

**Comment by tschoonj** [comment]
Here is the upstream bug: https://bugzilla.gnome.org/show_bug.cgi?id=755526

**Comment by staticfloat** [comment]
Thank you for the thorough explanation @tschoonj.  Your use case is different than mine, but I agree with you that this relocatability isn't what we want.

**Comment by tschoonj** [comment]
So would it be alright if I introduced the patch from MacPorts, add relocation as an option, and modify the caveat message saying that it only applies in some cases when relocation is on?

**Comment by mikemcquaid** [comment]
@tschoonj If the patch is submitted upstream: yep.

**Comment by tschoonj** [comment]
Well all I did was filing a bug report: no guarantee they will actually merge it in :smile: 

Will rewrite the formula later tonight and push it for rebuild

**Comment by mikemcquaid** [comment]
Thanks @tschoonj!

File: Library/Formula/gdk-pixbuf.rb [hunk 3]
```diff
@@ -51,17 +68,21 @@ def install
 
   def post_install
     # Change the version directory below with any future update
-    ENV["GDK_PIXBUF_MODULEDIR"]="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-2.0/2.10.0/loaders"
+    if build.with?("relocations")
+      ENV["GDK_PIXBUF_MODULE_FILE"]="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache"
```
staticfloat: Note that with `GDK_PIXBUF_MODULE_FILE` set to the linked path, running `gdk-pixbuf-query-loaders` like this will delete the symlink and replace it with a regular file, [as reported here](https://github.com/Homebrew/homebrew/pull/44262#discussion_r40289868).  This causes an upgrade or an `unlink`/`link` to throw an error, as the regular file is not removed after an `unlink` or `rm` of a formula.  My solution is to set `GDK_PIXBUF_MODULE_FILE` to point at the `#{lib}/` path, rather than `#{HOMEBREW_PREFIX}/lib` path for this particular operation.

Library/Formula/gdk-pixbuf.rb
```diff
@@ -51,17 +68,21 @@ def install
...
+      ENV["GDK_PIXBUF_MODULEDIR"]="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-2.0/2.10.0/loaders"
+    end
     system "#{bin}/gdk-pixbuf-query-loaders", "--update-cache"
   end
 
   def caveats; <<-EOS.undent
     Programs that require this module need to set the environment variable
       export GDK_PIXBUF_MODULEDIR="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-2.0/2.10.0/loaders"
-    If you need to manually update the query loader cache, set GDK_PIXBUF_MODULEDIR then run
+      export GDK_PIXBUF_MODULE_FILE="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache"
```
staticfloat: Same note here as above.

Issue #41085, opened 2015-06-25
Python fails to build core extension modules on Xcode-only 10.11


mikestecker:
I know I'm running a beta, but wanted to report this anyway.

Python fails during the postinstall phase:
```
==> /usr/local/Cellar/python/2.7.10/bin/python -s setup.py --no-user-cfg install --force --verbose --install-scripts=/usr/local/Cellar/python/2.7.10/
  File "/usr/local/Cellar/python/2.7.10/libexec/setuptools/setuptools/command/bdist_egg.py", line 463, in make_zipfile
    z = zipfile.ZipFile(zip_filename, mode, compression=compression)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/zipfile.py", line 736, in __init__
    "Compression requires the (missing) zlib module"
RuntimeError: Compression requires the (missing) lib module
```

I have the latest Xcode beta installed:
```
/Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk
xcode-select version 2342
```

Here is the output of running `brew gist-logs python` https://gist.github.com/dc7662dba10b070d2336

[edit @tdsmith Sep 14 2015: Installing the Command Line Tools with `xcode-select --install` should let you work around this. I would like this to work without the CLT package but won't have time to investigate for a bit; community efforts are welcome!]

shaneburrell:
Same here.

tdsmith:
What command did you use to install Python? What's `xcode-select -p` say? The Python build process is failing to build any of the extension modules that depend on the OS X SDK.

shaneburrell:
first brew reinstall python then uninstall and install

command output = /Applications/Xcode-beta.app/Contents/Developer

tdsmith:
@shaneburrell, can you paste logs with `brew gist-logs python` to be sure we're looking at the same problem?

Does switching the active Xcode configuration to the Command Line Tools (`xcode-select -s /Library/Developer/CommandLineTools`) help?

shaneburrell:
I only have Xcode 7 installed.

gist-log to follow in a min or two.

tdsmith:
It looks like python's setup.py is failing to detect the libraries and is refusing to even attempt to build them. It would be nice to be able to send a patch for this upstream.

shaneburrell:
[...]

PR #44245, opened 2015-09-22
Convince cpythons to recognize .tbd stubs

tdsmith:
Applies the patches I sent upstream. Allows building Python stdlib (and other) extension modules on Xcode 7-only systems.

Ref #41085.


mikemcquaid:
:+1: but worth thinking about what we'll do if upstream doesn't accept them.

Issue #44197, opened 2015-09-21
Formula commonmark fails to build with Python 3.5


zmwangx:
Actually, `make` works fine, but `make test` fails, due to some incompatibility with Python 3.5 in the tests (or perhaps something deeper, but I doubt it — I haven't looked too deep into the issue, and I only have a partial fix at the moment). This is clearly an upstream issue, and I have reported this upstream: https://github.com/jgm/cmark/issues/83.

This affects installing on 10.11, since there's no bottle for it. Installing on earlier systems works fine, since the formula was built against Python 3.4 when it was last bottled.

Btw, for future reference, I've never been clear on this: when there's a failure due to an upstream issue (that is already reported upstream), should I report in Homebrew's issue tracker also? That might help with troubleshooting, but that might also pollute the issue tracker with stuff for which nothing can be done here.

tdsmith:
I think we can reasonably choose not to run `make test` unless this will be fixed soon; what do you think?

I'm not clear on when it's useful to report upstream bugs here either; there's probably not a bright line.  Many upstream bugs are actionable in some way. The most useful upstream-bug tickets are if there are upstream commits or sane-looking pull requests that we can apply as patches between releases. This report strikes me as useful since it's suddenly broken because of a change we (in this case: I) recently made to another package. Stuff that's visibly broken to a lot of Homebrew users is good to have a ticket for, since it'll be reported anyway.

zmwangx:
> I think we can reasonably choose not to run `make test` unless this will be fixed soon; what do you think?

No objection, provided that someone — probably I — will remember to add it back when the issue is fixed; it's still good to have in my opinion. I have no idea of when this will be fixed.

Submitted #44246.

tdsmith:
I'll close this for now; thanks for reporting this and getting it fixed!

zmwangx:
No problem.

PR #44246, opened 2015-09-22
commonmark: disable make test during build

zmwangx:
`make test` fails due to incompatibility with CPython 3.5 (https://github.com/jgm/cmark/issues/83). Disable it per discussion in https://github.com/Homebrew/homebrew/issues/44197.


tdsmith:
Pulled; thanks for the PR!

PR #44250, opened 2015-09-22
llvm: add more components

aw1621107:
Compiler runtime libraries, regression tests, polly optimizer, libcxxabi, and a bit of
sort-of-organization.

From my testing all the stable stuff except `--with-sanitizers` will build fine. The sanitizer
error looks to me to be an upstream issue, so with any luck it'll eventually disappear
with the next update.

More info/reasoning for the changes can be found [here](https://github.com/aw1621107/homebrew/commits/llvm-updates); didn't want to create
a huge first comment.

--HEAD builds are still shaky, but as far as I know it's something to deal with on the
llvm side, if I can ever get someone's attention...


Library/Formula/llvm.rb
```diff
@@ -116,18 +152,39 @@ def install
...
     ENV.libcxx if ENV.compiler == :clang
 
-    if build.with?("lldb") && build.without?("clang")
-      raise "Building LLDB needs Clang support library."
+    # compiler-rt has some iOS simulator features that require i386 symbols
+    # I'm assuming the rest of clang needs support too for 32-bit compilation
+    # to work correctly, but if not, perhaps universal binaries could be
+    # limited to compiler-rt. llvm makes this somewhat easier because compiler-rt
+    # can almost be treated as an entirely different build from llvm.
+    ENV.permit_arch_flags
```
mikemcquaid: I'd rather limit this to specific cases where needed.

Library/Formula/llvm.rb
```diff
@@ -116,18 +152,39 @@ def install
...
+
+    prep = lambda do |resource_name, path, required|
+      if build.with?(resource_name) &&
+         (required.nil? || build.with?(required))
+        (buildpath/path).install resource(resource_name)
+      elsif build.with?(resource_name) && build.without?(required)
+        raise resource_name + " requires " + required + " to build."
+      end
```
mikemcquaid: I have no idea what's going on here. Please just inline these instead of using a lambda.

Library/Formula/llvm.rb
```diff
@@ -116,18 +152,39 @@ def install
...
+    if build.with?("libcxx")
+      (buildpath/"projects/libcxxabi").install resource("libcxxabi")
     end
 
-    (buildpath/"tools/lld").install resource("lld") if build.with? "lld"
-    (buildpath/"tools/lldb").install resource("lldb") if build.with? "lldb"
+    if build.with?("tests") &&
+       build.with?("sanitizers") &&
+       build.without?("libcxx")
+      raise "Building and running sanitizer tests requires libc++"
```
mikemcquaid: Use `odie`

Library/Formula/llvm.rb
aw1621107: I thought that that flag was required because some of the sanitizers that are built for iOS simulator were being built 64-bit only, while the simulator uses both 64 bit and 32 bit, depending on the particular model being simulated. Trying again with a different tweak...

aw1621107: Ran with/without ENV.permit_arch_flags. Without it, the build fails about here:
```
Scanning dependencies of target clang_rt.builtins_osx
/Applications/Xcode-beta.app/Contents/Developer/usr/bin/make -f projects/compiler-rt/lib/builtins/CMakeFiles/clang_rt.builtins_osx.dir/build.make projects/compiler-rt/lib/builtins/CMakeFiles/clang_rt.builtins_osx.dir/build
Generating libclang_rt.builtins_osx.a
cd /tmp/llvm20150923-33559-mwb0c7/projects/compiler-rt/lib/builtins && lipo -output /tmp/llvm20150923-33559-mwb0c7/./lib/clang/3.8.0/lib/darwin/libclang_rt.builtins_osx.a -create -arch i386 /tmp/llvm20150923-33559-mwb0c7/./lib/clang/3.8.0/lib/darwin/libclang_rt.builtins_i386_osx.a -arch x86_64 /tmp/llvm20150923-33559-mwb0c7/./lib/clang/3.8.0/lib/darwin/libclang_rt.builtins_x86_64_osx.a
fatal error: /Applications/Xcode-beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: specifed architecture type (i386) for file (/tmp/llvm20150923-33559-mwb0c7/./lib/clang/3.8.0/lib/darwin/libclang_rt.builtins_i386_osx.a) does not match its cputype (16777223) and cpusubtype (3) (should be cputype (7) and cpusubtype (3))
make[2]: *** [projects/compiler-rt/lib/builtins/libclang_rt.builtins_osx.a] Error 1
make[1]: *** [projects/compiler-rt/lib/builtins/CMakeFiles/clang_rt.builtins_osx.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```
With, it passes that point
```
Generating libclang_rt.builtins_osx.a
cd /tmp/llvm20150923-45191-1glhilv/projects/compiler-rt/lib/builtins && lipo -output /tmp/llvm20150923-45191-1glhilv/./lib/clang/3.8.0/lib/darwin/libclang_rt.builtins_osx.a -create -arch i386 /tmp/llvm20150923-45191-1glhilv/./lib/clang/3.8.0/lib/darwin/libclang_rt.builtins_i386_osx.a -arch x86_64 /tmp/llvm20150923-45191-1glhilv/./lib/clang/3.8.0/lib/darwin/libclang_rt.builtins_x86_64_osx.a
Built target clang_rt.builtins_osx
```
But fails later on
```
cd /tmp/llvm20150923-45191-1glhilv/projects/compiler-rt/lib/asan && /usr/local/Library/ENV/4.3/clang++   -DASAN_DYNAMIC=1 -DASAN_HAS_EXCEPTIONS=1 -D_DEBUG -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -I/tmp/llvm20150923-45191-1glhilv/projects/compiler-rt/lib/asan -I/tmp/llvm20150923-45191-1wmw25h/projects/compiler-rt/lib/asan -I/tmp/llvm20150923-45191-1glhilv/include -I/tmp/llvm20150923-45191-1wmw25h/include -I/tmp/llvm20150923-45191-1wmw25h/projects/compiler-rt/lib/asan/..  -fPIC -fvisibility-inlines-hidden -Wall -W -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -pedantic -Wno-long-long -Wcovered-switch-default -Wnon-virtual-dtor -std=c++11 -Wall -std=c++11 -arch i386 -arch x86_64   -UNDEBUG   -fPIC -fvisibility-inlines-hidden -Wall -W -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -pedantic -Wno-long-long -Wcovered-switch-default -Wnon-virtual-dtor -std=c++11 -Wall -std=c++11 -stdlib=libc++ -mios-simulator-version-min=7.0 -isysroot /Applications/Xcode-beta.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator9.1.sdk -fPIC -fno-builtin -fno-exceptions -fomit-frame-pointer -funwind-tables -fno-stack-protector -fvisibility=hidden -fno-function-sections -fno-lto -O3 -gline-tables-only -Wno-gnu -Wno-variadic-macros -Wno-c99-extensions -Wno-non-virtual-dtor -fno-rtti -ftls-model=initial-exec -o CMakeFiles/RTAsan_dynamic.iossim.dir/asan_win.cc.o -c /tmp/llvm20150923-45191-1wmw25h/projects/compiler-rt/lib/asan/asan_win.cc
Undefined symbols for architecture i386:
  "_sigaltstack$UNIX2003", referenced from:
      __sanitizer::SetAlternateSignalStack() in sanitizer_posix_libcdep.cc.o
      __sanitizer::UnsetAlternateSignalStack() in sanitizer_posix_libcdep.cc.o
  "_sleep$UNIX2003", referenced from:
      __sanitizer::SleepForSeconds(int) in sanitizer_posix_libcdep.cc.o
  "_usleep$UNIX2003", referenced from:
      __sanitizer::SleepForMillis(int) in sanitizer_posix_libcdep.cc.o
  "_waitpid$UNIX2003", referenced from:
      __sanitizer::SymbolizerProcess::StartSymbolizerSubprocess() in sanitizer_symbolizer_posix_libcdep.cc.o
ld: symbol(s) not found for architecture i386
make[2]: *** [lib/clang/3.8.0/lib/darwin/libclang_rt.ubsan_iossim_dynamic.dylib] Error 1
```

Manual builds outside of homebrew complete, though, so maybe ENV.permit_arch_flags is addressing a symptom rather than a cause...

mikemcquaid: It doesn't seem to be needed without specific options used, right?

aw1621107: True. I'll move it.

Library/Formula/llvm.rb
```diff
@@ -116,18 +152,66 @@ def install
...
+      else
+        odie "clang's extra tools must be built with clang"
+      end
+    end
+
+    if build.with? "libcxx"
       (buildpath/"projects/libcxx").install resource("libcxx")
-      (buildpath/"tools/clang").install resource("clang")
-      (buildpath/"tools/clang/tools/extra").install resource("clang-tools-extra")
+      (buildpath/"projects/libcxxabi").install resource("libcxxabi")
```
manphiz: The "libcxxabi" bit is actually no-op due to build script of libcxx only link to system libc++abi. Also it makes more sense to actually link to system libc++abi, otherwise there may be potentially interoperability issue when upstream libc++abi is not 100% incompatible with the system one, which could lead to obscure errors at run-time.

The actual case that does need libc++abi is on Snow Leopard, which didn't ship with libc++abi. However I guess it is not officially supported by now.

aw1621107: Huh, didn't know that. I'll get rid of that then. Thanks!

aw1621107:
Everything now compiles with #44312 and a LLVM patch that I just submitted.

Should the `lldb_codesign` issue be put as a caveat? Or should a way be found to add the login keychain at compile time?

aw1621107:
Stopped working... Will have to try something else

mikemcquaid:
https://github.com/Homebrew/homebrew/pull/44312 was merged.

aw1621107:
Alright, only thing left is getting this keychain thing sorted out...

aw1621107:
Well hopefully that's it. Build worked on my machine at least.

Library/Formula/llvm.rb
```diff
@@ -78,17 +92,25 @@ class Llvm < Formula
...
+    resource "sanitizers" do
+      url "http://llvm.org/git/compiler-rt.git"
     end
   end
 
   option :universal
-  option "with-clang", "Build Clang support library"
+  option "with-clang", "Build the Clang compiler and support libraries"
+  option "with-extra-tools", "Build extra tools for Clang"
+  option "with-libcxx", "Build the libc++ standard library"
```
xu-cheng: Shouldn't we build libc++ unconditionally for clang support? Without it, clang++ will be broken or useless, right?

Library/Formula/llvm.rb
```diff
@@ -153,9 +237,6 @@ def install
     end
 
     if build.with? "clang"
-      system "make", "-C", "projects/libcxx", "install",
-        "DSTROOT=#{prefix}", "SYMROOT=#{buildpath}/projects/libcxx"
-
```
xu-cheng: Why this can be removed?

Library/Formula/llvm.rb
```diff
@@ -116,18 +142,73 @@ def install
...
+
+        # Building lldb requires a code signing certificate.
+        # The instructions provided by llvm creates this certificate in the
+        # user's login keychain. Unfortunately, the login keychain is not in
+        # the search path in a superenv build. The following three lines add
+        # the login keychain to ~/Library/Preferences/com.apple.security.plist,
+        # which adds it to the superenv keychain search path.
+        mkdir_p "#{ENV["HOME"]}/Library/Preferences"
+        username = `whoami`.strip! # Has ending newline otherwise
+        system "security", "list-keychains", "-d", "user", "-s", "/Users/#{username}/Library/Keychains/login.keychain"
```
xu-cheng: You can use `ENV["USER"]` instead of `whoami`. Also can you test the option locally with sandbox enabled, as `brew install llvm --with-clang --with-lldb --sandbox -v`. I want to make sure it doesn't write file to place not permitted. Thanks.

Library/Formula/llvm.rb
manphiz: LLVM uses system libc++ by default. It also needs extra step to use its own libc++. Homebrew/versions/llvm* have instructions about how to enable it in caveat.

xu-cheng: :cool:

aw1621107: The clang [get started](http://clang.llvm.org/get_started.html) page says that libcxx is "only required to build and run Compiler-RT tests on OS X, optional otherwise".

Then again, the documentation between the various projects is a bit shaky, but if the clang guys are fine building without it I think we're good

Library/Formula/llvm.rb
aw1621107: LLVM automatically detects the presence of libcxx, so I thought there isn't a need to make it manually.

Library/Formula/llvm.rb
aw1621107: *That* is the answer I was looking for. Thank you!

The sandbox build completed just fine, too.

Library/Formula/llvm.rb
aw1621107: @manphiz Wait... Shouldn't building it anyways still be an option for people who might cross-compile? Homebrew/OS X may not use it itself, but I'd guess linking against a bleeding-edge libc++ would work better with the matching libc++abi...

DomT4: Technically Snow Leopard is still [supported](https://github.com/Homebrew/homebrew/blob/master/Library/brew.rb#L60-L65), albeit almost entirely by one maintainer these days.

aw1621107: I was thinking it could also be useful for cross-compiling to *nix or something, but libc++abi is only supported by OS X at the moment, and the webpage doesn't mention anything about ports to other platforms... Worth leaving in?

DomT4: I have no objections to leaving it in, personally. Unless it's causing harm I'd be tempted to leave it there until we have time to poke around a bit more when/why it's needed.

aw1621107: Alright, reverted.

aw1621107:
Uhhh.... Hmmm. Looks like I botched the merge. Only file I actually wanted to change was llvm.rb. Should be up to date with the current file now.

DomT4:
Yeah, It's not going through the CI cleanly at the moment because of the rebase/merge.

aw1621107:
I can open a new pull request if that would help

DomT4:
Might not be a bad idea at this point. Newer PR would perhaps get more maintainer eyes over it as well, since this PR is a little buried now.

PR #44253, opened 2015-09-22
Upgrades MacVim to snapshot 78 (based on 7.4.769)

emilyst:
This updates an existing formula to use the newest available snapshot,
number 78, which is based on Vim 7.4.769 and so contains approximately
700 new patches missing in snapshot 77.

### DIFF
--- a/Library/Formula/macvim.rb
+++ b/Library/Formula/macvim.rb
@@ -15,9 +15,9 @@ end
 class Macvim < Formula
   desc "GUI for vim, made for OS X"
   homepage "https://code.google.com/p/macvim/"
-  url "https://github.com/macvim-dev/macvim/archive/snapshot-77.tar.gz"
-  version "7.4-77"
-  sha256 "6b7f4b48ecef4a00dca17efef551884fcea1aa9897005497d31f52da7304bc5f"
+  url "https://github.com/macvim-dev/macvim/archive/snapshot-78.tar.gz"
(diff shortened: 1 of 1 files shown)


chdiza:
I'm not so sure putting Snapshot 78 in Homebrew core is a good idea.  It was made in error (as the maintainer freely admits), and half of it's Mac-specific commits were immediately reverted because they contained a bad GUI regression affecting all new MacVim windows.  The next two commits after Snapshot 78 (which are the last two in the repo, as of right now) fix the GUI issue.

77 is comically out of date, I know, but I suggest users brew from HEAD instead; that way one gets all of the Vim patch upgrades but doesn't get a broken GUI.

emilyst:
> It was made in error (as the maintainer freely admits), and half of it's Mac-specific commits were immediately reverted because they contained a bad GUI regression affecting all new MacVim windows.

If that's the case, I accept your judgment. I didn't know.

> 77 is comically out of date, I know, but I suggest users brew from HEAD instead; that way one gets all of the Vim patch upgrades but doesn't get a broken GUI.

I'll do that. Sad I won't have a *stable* snapshot.

chdiza:
@emilyst Two things.  First, HEAD is no longer out of date.  Second, there will be a snapshot coming very soon.  The maintainers are just ironing out the kinks of automated snapshot-making.

PR #44259, opened 2015-09-23
datomic 0.9.5302

rhendric:


Library/Formula/datomic.rb
```diff
@@ -1,9 +1,9 @@
 class Datomic < Formula
   desc "Database separating transactions, storage and queries"
   homepage "http://www.datomic.com/"
-  url "https://my.datomic.com/downloads/free/0.9.5206"
-  version "0.9.5206"
-  sha256 "b5b406d301b3e3fd4ff61d3e20946fb67fc5adae269bc2ee3bbca0819e02de3c"
+  version "0.9.5302"
+  url "https://my.datomic.com/downloads/free/#{version}"
```
DomT4: Please don't do this. We never inline version variables into URLs in Homebrew.

rhendric: Sorry, didn't know! (It's quite common in homebrew-cask.) Maybe this rule should be mentioned in the formula cookbook?

DomT4: No worries, the tone was intended as friendly, however badly tone comes across in text. Indeed, The Cask are fans of the inlined version, but we don't use it.

We're slowly moving towards proper API docs which will help. The Cookbook is a bit of a behemoth. If you wanted to file a PR though we'd certainly be happy to review it :)

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/4e079f7e0319a2a8833c5b3ec416097f35a45532. Thank you for your contribution to Homebrew @rhendric; we appreciate it! :smiley_cat:

PR #44262, opened 2015-09-23
Set GDK_PIXBUF_MODULE_FILE for gdk-pixbuf

staticfloat:
Without this variable set in addition to GDK_PIXBUF_MODULEDIR, the proper cache file does not get updated in the event of Homebrew installations in places other than `/usr/local/`

### DIFF
--- a/Library/Formula/gdk-pixbuf.rb
+++ b/Library/Formula/gdk-pixbuf.rb
@@ -52,12 +52,14 @@ class GdkPixbuf < Formula
   def post_install
     # Change the version directory below with any future update
     ENV["GDK_PIXBUF_MODULEDIR"]="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-2.0/2.10.0/loaders"
+    ENV["GDK_PIXBUF_MODULE_FILE"]="#{lib}/gdk-pixbuf-2.0/2.10.0/loaders.cache"
     system "#{bin}/gdk-pixbuf-query-loaders", "--update-cache"
   end
 
   def caveats; <<-EOS.undent
-    Programs that require this module need to set the environment variable
+    Programs that require this module need to set the environment variables:
       export GDK_PIXBUF_MODULEDIR="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-2.0/2.10.0/loaders"
+      export GDK_PIXBUF_MODULE_FILE="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache"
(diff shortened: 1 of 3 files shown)


staticfloat:
Sorry for the terrible branch name, submitting patch from my browser.

Library/Formula/gdk-pixbuf.rb
```diff
@@ -52,12 +52,14 @@ def install
   def post_install
     # Change the version directory below with any future update
     ENV["GDK_PIXBUF_MODULEDIR"]="#{HOMEBREW_PREFIX}/lib/gdk-pixbuf-2.0/2.10.0/loaders"
+    ENV["GDK_PIXBUF_MODULE_FILE"]="#{lib}/gdk-pixbuf-2.0/2.10.0/loaders.cache"
```
xu-cheng: This should be `HOMEBREW_PREFIX/lib`.

staticfloat: I don't think it should be; I ran into problems when putting it there; performing `brew rm` would not clear out the non-symlink file in `#{HOMEBREW_PREFIX}/lib`, and so subsequent `brew install gdk-pixbuf` operations would fail.  With this setup, the file is placed within the formula's Cellar, and is then symlinked into the global namespace, which works out better for everyone, as far as I can tell.

xu-cheng: If it won't affect other formulae, that would be OK. But you still need to manually make symlink. (brew link happens before post_install). Also please use `make_symlink` instead of `ln`, so it will create symlink with related path.  

BTW, it would be great if you check other formulae depend on this one, especially those who regenerate cache in their own post install. Thanks.

tschoonj:
I have filed a PR for the latest gdk-pixbuf release #44231. In this version there have been some changes that could possible affect your issue. Can you check if your PR is still necessary after mine is merged in (which will be today hopefully)

Library/Formula/gdk-pixbuf.rb
staticfloat: Interestingly enough, the bottle does contain a `loaders.cache`; so that symlink does get created.  By setting `GDK_PIXBUF_MODULE_FILE` as I have it up there, the original file gets destroyed and recreated by `gdk-pixbuf-query-loaders`, rather than the symlink getting destroyed and then recreated as a normal file.  I have updated two dependent formulae; `librsvg` passes its tests with this modification, I'm not sure about `xplanetfx`.

staticfloat:
Closing this in favor of the discussion in https://github.com/Homebrew/homebrew/issues/44231

PR #44264, opened 2015-09-23
More failure logs

xu-cheng:
cc @mikemcquaid

### DIFF
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -530,7 +530,7 @@ module Homebrew
       formula_fetch_options << canonical_formula_name
       test "brew", "fetch", "--retry", *formula_fetch_options
       test "brew", "uninstall", "--force", canonical_formula_name if formula.installed?
-      install_args = []
+      install_args = ["--verbose"]
       install_args << "--build-bottle" unless ARGV.include? "--no-bottle"
       install_args << "--HEAD" if ARGV.include? "--HEAD"
 
@@ -848,7 +848,11 @@ module Homebrew
     ENV["HOMEBREW_SANDBOX"] = "1"
     ENV["HOMEBREW_NO_EMOJI"] = "1"
     ENV["HOMEBREW_FAIL_LOG_LINES"] = "150"
-    ARGV << "--verbose" if ENV["TRAVIS"]
+
+    if ENV["TRAVIS"]
+      ARGV << "--verbose"
+      ENV["HOMEBREW_VERBOSE_USING_DOTS"] = "1"
+    end
 
     if ARGV.include?("--ci-master") || ARGV.include?("--ci-pr") \
        || ARGV.include?("--ci-testing")
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -154,9 +154,12 @@ module Homebrew
           working_dir.cd { exec(*@command) }
         end
         write.close
-        while line = read.gets
-          puts line if verbose
-          @output += line
+        while buf = read.read(1)
+          if verbose
+            print buf
+            $stdout.flush
+          end
+          @output << buf
         end
       ensure
         read.close
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -18,7 +18,6 @@
 # --dry-run:       Just print commands, don't run them.
 # --fail-fast:     Immediately exit on a failing step.
 # --verbose:       Print out all logs in realtime
-# --no-verbose-install: Don't run `brew install` with `--verbose`.
 #
 # --ci-master:           Shortcut for Homebrew master branch CI options.
 # --ci-pr:               Shortcut for Homebrew pull request CI options.
@@ -532,7 +531,6 @@ module Homebrew
       test "brew", "fetch", "--retry", *formula_fetch_options
       test "brew", "uninstall", "--force", canonical_formula_name if formula.installed?
       install_args = []
-      install_args << "--verbose" unless ARGV.include? "--no-verbose-install"
       install_args << "--build-bottle" unless ARGV.include? "--no-bottle"
       install_args << "--HEAD" if ARGV.include? "--HEAD"
 
@@ -849,6 +847,7 @@ module Homebrew
     ENV["HOMEBREW_DEVELOPER"] = "1"
     ENV["HOMEBREW_SANDBOX"] = "1"
     ENV["HOMEBREW_NO_EMOJI"] = "1"
+    ENV["HOMEBREW_FAIL_LOG_LINES"] = "150"
     ARGV << "--verbose" if ENV["TRAVIS"]
 
     if ARGV.include?("--ci-master") || ARGV.include?("--ci-pr") \
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1304,8 +1304,14 @@ class Formula
       $stdout.flush
 
       unless $?.success?
+        log_lines = ENV["HOMEBREW_FAIL_LOG_LINES"]
+        log_lines ||= "15"
+
         log.flush
-        Kernel.system "/usr/bin/tail", "-n", "5", logfn unless verbose
+        unless verbose
+          puts "Last #{log_lines} lines from #{logfn}:"
+          Kernel.system "/usr/bin/tail", "-n", log_lines, logfn
+        end
         log.puts
 
         require "cmd/config"
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1255,6 +1255,8 @@ class Formula
   # system "make", "install"</pre>
   def system(cmd, *args)
     verbose = ARGV.verbose?
+    verbose_using_dots = !ENV["HOMEBREW_VERBOSE_USING_DOTS"].nil?
+
     # remove "boring" arguments so that the important ones are more likely to
     # be shown considering that we trim long ohai lines to the terminal width
     pretty_args = args.dup
@@ -1288,9 +1290,23 @@ class Formula
           end
           wr.close
 
-          while buf = rd.gets
-            log.puts buf
-            puts buf
+          if verbose_using_dots
+            last_dot = Time.at(0)
+            while buf = rd.gets
+              log.puts buf
+              # make sure dots printed with interval of at least 1 min.
+              if (Time.now - last_dot) > 60
+                print "."
+                $stdout.flush
+                last_dot = Time.now
+              end
+            end
+            puts
+          else
+            while buf = rd.gets
+              log.puts buf
+              puts buf
+            end
           end
         ensure
           rd.close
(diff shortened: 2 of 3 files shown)


Library/Homebrew/formula.rb
```diff
@@ -1288,9 +1290,18 @@ def system(cmd, *args)
...
 
-          while buf = rd.gets
-            log.puts buf
-            puts buf
+          if verbose
+            while buf = rd.gets
+              log.puts buf
+              puts buf
+            end
+          else # travis
```
mikemcquaid: I think we should do this always when `--verbose` isn't passed, actually.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -155,9 +154,12 @@ def run
...
         write.close
-        while line = read.gets
-          puts line if verbose
-          @output += line
+        while buf = read.read(32)
+          if verbose
+            print buf
+            $stdout.flush
+          end
+          @output << buf
```
mikemcquaid: What's the reason/advantage of this change?

Library/Homebrew/formula.rb
```diff
@@ -1288,9 +1290,18 @@ def system(cmd, *args)
...
+            while buf = rd.gets
+              log.puts buf
+              print "."
```
mikemcquaid: It's unclear to me when this will print; after every new write to the output? That seems like it's going to print way too much; it should probably have a timer and print e.g. no more than once a minute.

xu-cheng: Currently, it will print a dot for every line of logs. If you want, we can make it every N lines or every N bytes.

>  it should probably have a timer and print e.g. no more than once a minute.

I kinda like dot printing corresponding real log. So we can see wether the build is really freezing.

Library/Homebrew/formula.rb
mikemcquaid: e.g. don't limit it just to travis. I think it would help avoid confusion for long builds like `boost` (as long as we keep the dot count sensible).

Library/Homebrew/formula.rb
mikemcquaid: Yeh, I like the idea of handling build freezing too. I think that printing every line is a good idea but will probably print too many dots. So, instead of printing a dot every 1 minute it prints a dot on every line unless there was already a dot printed in the last minute. How does that sound?

Library/Homebrew/formula.rb
xu-cheng: In that case, we need to introduce a new flag, `--semi-verbose` or something.

mikemcquaid: I don't agree. Users are confused with the current behaviour and I think this would be an improvement. If you want you can make it `HOMEBREW_DEVELOPER` only for now but I do think it's a sensible change to make.

Library/Homebrew/cmd/test-bot.rb
xu-cheng: The original `read.gets` will only return for every new line. When we print dots, they will be all in one single line, hence won't be printed out.

Library/Homebrew/formula.rb
mikemcquaid: I really don't like the idea of making `system` behave differently for just one CI system.

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: Ah, I see. Would be good to indicate what the `32` is by splitting it out into a variable. Also, it seems like `32` is probably more bytes than we want if we want to print every dot.

Library/Homebrew/formula.rb
xu-cheng: > I really don't like the idea of making system behave differently for just one CI system.

That's exactly what I'm talking about. We use a flag here and set this flag based on CI system in `test-bot`.

Library/Homebrew/formula.rb
xu-cheng: I don't like the idea of timer too much, it complicates the implementations. Can we just print dot based number of lines?

mikemcquaid: Have you tried this with e.g. `boost`? It seems like it'd produce a vast number of dots.

mikemcquaid: The timer is sensible for Travis because Travis itself is operating with a timer.

xu-cheng: OK, you have convinced me on this.

mikemcquaid: Note I don't mind changing the logic in `test-bot` to be travis-specific or move the timer there.

mikemcquaid: :+1:

mikemcquaid: I'm trying `boost` right now to see what it looks like.

Library/Homebrew/cmd/test-bot.rb
xu-cheng: We cannot make the variable too low as it will harm the performance.

Library/Homebrew/formula.rb
mikemcquaid:
```
==> brew install --build-bottle boost
Warning: You are using OS X 10.11.
We do not provide support for this pre-release version.
You may encounter build failures or other breakage.
==> Using the sandbox
==> Downloading https://downloads.sourceforge.net/project/boost/boost/1.58.0/boost_1_58_0.tar.bz2
Already downloaded: /Library/Caches/Homebrew/boost-1.58.0.tar.bz2
==> ./bootstrap.sh --prefix=/usr/local/Cellar/boost/1.58.0 --libdir=/usr/local/Cellar/boost/1.58.0/lib --without-icu --without-libraries=python,mpi
......................
==> ./b2 headers
..........
==> ./b2 --prefix=/usr/local/Cellar/boost/1.58.0 --libdir=/usr/local/Cellar/boost/1.58.0/lib -d2 -j8 --layout=tagged --user-config=user-config.jam install threading=multi,single link=shared,static
...........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................^C
```

:laughing:

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: We could make it lower just for Travis. If we're printing dots based on time output it seems to make sense to ensure that we output every dot otherwise they will e.g. just appear in groups of `16`.

xu-cheng:
Updated.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -155,9 +154,12 @@ def run
           working_dir.cd { exec(*@command) }
         end
         write.close
-        while line = read.gets
-          puts line if verbose
-          @output += line
+        while buf = read.read(8)
```
mikemcquaid: This is 8 bytes (rather than bits) I believe so it's still gonna only print every 8th `.` unfortunately.

Library/Homebrew/formula.rb
```diff
@@ -1288,9 +1290,23 @@ def system(cmd, *args)
...
 
-          while buf = rd.gets
-            log.puts buf
-            puts buf
+          if verbose_using_dots
+            last_dot = Time.at(0)
+            while buf = rd.gets
+              log.puts buf
+              # make sure dots printed with interval of at least 30 sec.
+              if Time.now - last_dot > 30
```
mikemcquaid: Does this maybe need to be `(Time.now - last_dot) > 30`? I can never remember the operator precedence.

mikemcquaid: Also, maybe 1 minute rather than 30 seconds?

Library/Homebrew/cmd/test-bot.rb
mikemcquaid: I really don't think it's worth worrying about performance here given it's a blocking IO operation in a tight loop and only in our CI.

mikemcquaid:
Other than the byte count and operator question: :+1:

xu-cheng:
Ok, addressed your feedback.

I also did a small benchmark, using `ruby -e puts("."*10000)` as the test case. It shows `read(1)` will have some slight performance issuing when handling high speed output.
```
Comparison:
                gets:        8.4 i/s
              read 8:        8.4 i/s - 1.00x slower
              read 1:        6.9 i/s - 1.22x slower
```

mikemcquaid:
@xu-cheng That seems worth it, to me, given that the actual impact on compilation performance will be minimal and its a relatively small difference.

xu-cheng:
I agree with you as such high speed output is kinda rare.

Anything else before :ship: it?

mikemcquaid:
:shipit:!

PR #44266, opened 2015-09-23
bottle: output maximum number of strings matches.

mikemcquaid:
`brew bottle --verbose` outputs all the matches for a string found in the files inside the Cellar. Instead of outputting all of these just output the first 100.

CC @xu-cheng @tdsmith

### DIFF
--- a/Library/Homebrew/cmd/bottle.rb
+++ b/Library/Homebrew/cmd/bottle.rb
@@ -32,6 +32,8 @@ BOTTLE_ERB = <<-EOS
   end
 EOS
 
+MAXIMUM_STRING_MATCHES = 100
+
 module Homebrew
   def keg_contains(string, keg, ignores)
     @put_string_exists_header, @put_filenames = nil
@@ -72,6 +74,8 @@ module Homebrew
         end
       end
 
+      text_matches = []
+
       # Use strings to search through the file for each string
       Utils.popen_read("strings", "-t", "x", "-", file.to_s) do |io|
         until io.eof?
@@ -82,11 +86,18 @@ module Homebrew
           next if linked_libraries.include? match # Don't bother reporting a string if it was found by otool
 
           result = true
+          text_matches << [match, offset]
+        end
+      end
 
-          if ARGV.verbose?
-            print_filename string, file
-            puts " #{Tty.gray}-->#{Tty.reset} match '#{match}' at offset #{Tty.em}0x#{offset}#{Tty.reset}"
-          end
+      if ARGV.verbose? && text_matches.any?
+        print_filename string, file
+        text_matches.first(MAXIMUM_STRING_MATCHES).each do |match, offset|
+          puts " #{Tty.gray}-->#{Tty.reset} match '#{match}' at offset #{Tty.em}0x#{offset}#{Tty.reset}"
+        end
+
+        if text_matches.size > MAXIMUM_STRING_MATCHES
+          puts "Only the first #{MAXIMUM_STRING_MATCHES} matches were output"
         end
       end
     end


Library/Homebrew/cmd/bottle.rb
```diff
@@ -84,10 +87,15 @@ def print_filename(string, filename)
...
+            strings_matches += 1
+            next if strings_matches > MAXIMUM_STRING_MATCHES
             print_filename string, file
             puts " #{Tty.gray}-->#{Tty.reset} match '#{match}' at offset #{Tty.em}0x#{offset}#{Tty.reset}"
           end
         end
+        if strings_matches > MAXIMUM_STRING_MATCHES
+          puts "Only the first #{MAXIMUM_STRING_MATCHES} matches were output"
+        end
       end
```
xu-cheng: Strange logic, why not `if ...  else ... end`? Also there are other types of string matching, such as `lib`. So how about rewriting `print_filename` directly:

```diff
diff --git a/Library/Homebrew/cmd/bottle.rb b/Library/Homebrew/cmd/bottle.rb
index 7774a56..bc7d0c1 100644
--- a/Library/Homebrew/cmd/bottle.rb
+++ b/Library/Homebrew/cmd/bottle.rb
@@ -31,12 +31,13 @@ BOTTLE_ERB = <<-EOS
     <% end %>
   end
 EOS
+MAXIMUM_STRING_MATCHES = 100
 
 module Homebrew
   def keg_contains(string, keg, ignores)
     @put_string_exists_header, @put_filenames = nil
 
-    def print_filename(string, filename)
+    def print_filename(string, filename, matches)
       unless @put_string_exists_header
         opoo "String '#{string}' still exists in these files:"
         @put_string_exists_header = true
@@ -47,6 +48,14 @@ module Homebrew
         puts "#{Tty.red}#{filename}#{Tty.reset}"
         @put_filenames << filename
       end
+
+      matches.first(MAXIMUM_STRING_MATCHES).each do |match|
+        puts " #{Tty.gray}-->#{Tty.reset} #{match}"
+      end
+
+      if matches.size > MAXIMUM_STRING_MATCHES
+        puts "Only the first #{MAXIMUM_STRING_MATCHES} matches were output"
+      end
     end
 
     result = false
@@ -65,11 +74,9 @@ module Homebrew
         linked_libraries = []
       end
 
-      if ARGV.verbose?
-        print_filename(string, file) if linked_libraries.any?
-        linked_libraries.each do |lib|
-          puts " #{Tty.gray}-->#{Tty.reset} links to #{lib}"
-        end
+      if ARGV.verbose? && !linked_libraries.empty?
+        matches = linked_libraries.map { |lib| "links to #{lib}" }
+        print_filename string, file, matches
       end
 
       # Use strings to search through the file for each string
@@ -84,8 +91,7 @@ module Homebrew
           result = true
 
           if ARGV.verbose?
-            print_filename string, file
-            puts " #{Tty.gray}-->#{Tty.reset} match '#{match}' at offset #{Tty.em}0x#{offset}#{Tty.reset}"
+            print_filename string, file, "match '#{match}' at offset #{Tty.em}0x#{offset}#{Tty.reset}"
           end
         end
       end
```

mikemcquaid: I'd rather all the linked_libraries were always printed. The problem is only the huge number of string matches. I'd be open to making this filenames instead (but even then there may be huge numbers of results).

xu-cheng: Ok, but I still find it's weird to have `if strings_matches > MAXIMUM_STRING_MATCHES` outside of `if ARGV.verbose?`. How about:
```ruby
text_matches = []
Utils.popen_read("strings", "-t", "x", "-", file.to_s) do |io|
 ...
 text_matches << [match, offset]
end
if text_matches.any? && ARGV.verbose?
  print_filename string, file
  text_matches.first(MAXIMUM_STRING_MATCHES).each do |match, offset|
    puts " #{Tty.gray}-->#{Tty.reset} match '#{match}' at offset #{Tty.em}0x#{offset}#{Tty.reset}"
  end
  if text_matches.size > MAXIMUM_STRING_MATCHES
    puts "Only the first #{MAXIMUM_STRING_MATCHES} matches were output"
  end
end
```

mikemcquaid: Cool, makes sense :+1:

mikemcquaid:
@xu-cheng look OK?

xu-cheng:
:+1:

PR #44268, opened 2015-09-23
docbook: sandbox related modifications

tschoonj:
per our observations in Homebrew/homebrew-x11#129
the install block contained a lot of operations that were dealing
with #{etc}/xml/catalog, outside the keg. These operations have now
been moved into a post_install block.

### DIFF
--- a/Library/Formula/docbook-xsl.rb
+++ b/Library/Formula/docbook-xsl.rb
@@ -3,6 +3,7 @@ class DocbookXsl < Formula
   homepage "http://docbook.sourceforge.net/"
   url "https://downloads.sourceforge.net/project/docbook/docbook-xsl/1.78.1/docbook-xsl-1.78.1.tar.bz2"
   sha256 "c98f7296ab5c8ccd2e0bc07634976a37f50847df2d8a59bdb1e157664700b467"
+  revision 1
 
   bottle do
     cellar :any_skip_relocation
@@ -21,6 +22,7 @@ class DocbookXsl < Formula
   end
 
   def install
+    ENV["XML_CATALOG_FILES"] = "#{etc}/xml/catalog"
     doc_files = %w[AUTHORS BUGS COPYING NEWS README RELEASE-NOTES.txt TODO VERSION VERSION.xsl]
     xsl_files = %w[assembly catalog.xml common docsrc eclipse epub epub3 extensions
                    fo highlighting html htmlhelp images javahelp lib log manpages
@@ -35,6 +37,7 @@ class DocbookXsl < Formula
   end
 
   def post_install
+    ENV["XML_CATALOG_FILES"] = "#{etc}/xml/catalog"
     [prefix/"docbook-xsl/catalog.xml", prefix/"docbook-xsl-ns/catalog.xml"].each do |catalog|
       system "xmlcatalog", "--noout", "--del", "file://#{catalog}", "#{etc}/xml/catalog"
       system "xmlcatalog", "--noout", "--add", "nextCatalog", "", "file://#{catalog}", "#{etc}/xml/catalog"
--- a/Library/Formula/docbook.rb
+++ b/Library/Formula/docbook.rb
@@ -12,8 +12,6 @@ class Docbook < Formula
     sha256 "5d563a04a11cee14fd7c52a6b4a85d397b019bf2f5cc96005e2d17dac4ad7231" => :mavericks
   end
 
-  depends_on "libxml2"
-
   resource "xml412" do
     url "http://www.docbook.org/xml/4.1.2/docbkx412.zip"
     sha256 "30f0644064e0ea71751438251940b1431f46acada814a062870f486c772e7772"
@@ -52,7 +50,7 @@ class Docbook < Formula
     # by other formulas to be removed
     unless File.file?("#{etc}/xml/catalog")
       ENV["XML_CATALOG_FILES"] = "#{etc}/xml/catalog"
-      system "#{Formula["libxml2"].opt_bin}/xmlcatalog", "--noout", "--create", "catalog"
+      system "xmlcatalog", "--noout", "--create", "catalog"
       (etc/"xml").install "catalog"
     end
 
@@ -78,9 +76,9 @@ class Docbook < Formula
     %w[4.2 4.1.2 4.3 4.4 4.5 5.0].each do |version|
       catalog = prefix/"docbook/xml/#{version}/catalog.xml"
 
-      system "#{Formula["libxml2"].opt_bin}/xmlcatalog", "--noout", "--del",
+      system "xmlcatalog", "--noout", "--del",
              "file://#{catalog}", "#{etc}/xml/catalog"
-      system "#{Formula["libxml2"].opt_bin}/xmlcatalog", "--noout", "--add", "nextCatalog",
+      system "xmlcatalog", "--noout", "--add", "nextCatalog",
              "", "file://#{catalog}", "#{etc}/xml/catalog"
     end
   end
--- a/Library/Formula/docbook.rb
+++ b/Library/Formula/docbook.rb
@@ -46,14 +46,6 @@ class Docbook < Formula
   def install
     (etc/"xml").mkpath
 
-    # only create catalog file if it doesn't exist already to avoid content added
-    # by other formulas to be removed
-    unless File.file?("#{etc}/xml/catalog")
-      ENV["XML_CATALOG_FILES"] = "#{etc}/xml/catalog"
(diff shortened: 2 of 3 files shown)


Library/Formula/docbook.rb
```diff
@@ -43,9 +43,6 @@ class Docbook < Formula
   end
 
   def install
-    (etc/"xml").mkpath
-    system "xmlcatalog", "--noout", "--create", "#{etc}/xml/catalog"
```
mikemcquaid: Might be nice to still create this in install inside `buildpath` and then use `etc.install`

tschoonj: If I would do that, wouldn't I still have the sandbox problem as I am creating a file outside of the keg during `install`?

mikemcquaid: `etc.install` is definitely fine.

DomT4:
Failed on El Cap for the same reason as before:

```
checking for xslt-config... /usr/bin/xslt-config
checking which XML catalog to use... /usr/local/etc/xml/catalog
checking for DocBook XML DTD... configure: error: not found. Make sure you have the DocBook DTD installed and ensure that it is registered in /usr/local/etc/xml/catalog.
==> Formula
Path: /usr/local/Library/Formula/scrollkeeper.rb
```

tschoonj:
I am not sure what is going on exactly but I has to do with libxml2: If I force the use of Homebrew's libxml2 then it fails on all three with exactly the same error.

It looks like the post_install block in docbook writes only the last nextCatalog line for some reason. Maybe the behavior of xmlcatalog has changed in the last version (which I think ships with El Capitan) regarding updating a file.

Cannot reproduce it locally though on my Yosemite with Homebrew's libxml2

tschoonj:
Is there any way to make `post_install` for docbook verbose on the buildbots? I would like to print the contents of the catalog file after every invocation of xmlcatalog.

Library/Formula/docbook.rb
```diff
@@ -4,11 +4,14 @@ class Docbook < Formula
...
   bottle do
-    revision 1
-    sha1 "b6666858a43379c827ba9a62df987bcbcfffc6f3" => :yosemite
-    sha1 "ac3a230baa6c7a2d9accada9d956a566f3520151" => :mavericks
-    sha1 "8cbf8d315805395f2ca483161daa2c8bc3c4e19c" => :mountain_lion
+    cellar :any_skip_relocation
+    revision 2
+    sha256 "a3ce9c187fefb20f949095c1c4ae4e3a9dc1887a17d7322c85c10633b92b3fb9" => :el_capitan
+    sha256 "d722e3791b31c9ab92a5f47d89a8f530accd628898dd10551a1185986c4eb08f" => :yosemite
+    sha256 "5d563a04a11cee14fd7c52a6b4a85d397b019bf2f5cc96005e2d17dac4ad7231" => :mavericks
```
xu-cheng: I think you need to rebase first.

mikemcquaid:
@tschoonj You could temporarily just use `ohai` or similar.

tschoonj:
what I need to do is get the output of:

````ruby
system "cat", "#{HOMEBREW_PREFIX}/etc/xml/catalog"
````

and see it printed in the Jenkins console

**Comment by mikemcquaid** [comment]
```
ohai `cat #{HOMEBREW_PREFIX}/etc/xml/catalog`
```

**Comment by tschoonj** [comment]
Thanks, let's see if the output is useful...

**Comment by tschoonj** [comment]
None of the `ohai` statements made it into the console output :cry: 

Anything else I could do?

File: Library/Formula/docbook.rb [hunk 3]
```diff
@@ -45,7 +47,14 @@ class Docbook < Formula
 
   def install
     (etc/"xml").mkpath
-    system "xmlcatalog", "--noout", "--create", "#{etc}/xml/catalog"
+
+    # only create catalog file if it doesn't exist already to avoid content added
+    # by other formulas to be removed
+    unless File.file?("#{etc}/xml/catalog")
+      ENV["XML_CATALOG_FILES"] = "#{etc}/xml/catalog"
+      system "#{Formula["libxml2"].opt_bin}/xmlcatalog", "--noout", "--create", "catalog"
```
DomT4: Does using the system `xmlcatalog` not work at all, on any platform?

tschoonj:
I have started a build removing the libxml2 dependency. Let's see if it works...

DomT4:
:+1:. If there's a problem on one platform I'm happy for us to use our `libxml` there temporarily, but I'd rather keep the dependency scoped to that problematic platform if that's necessary.

tschoonj:
It works! Can you pull this one in quickly? A lot of formulas depend on this one...

Thanks!

DomT4:
I'm getting further locally now, but building `gtk-doc` I get a newer error along the same lines:

```
checking for XML catalog... /usr/local/etc/xml/catalog
checking for xmlcatalog... /usr/local/opt/libxml2/bin/xmlcatalog
checking for DocBook XML DTD V4.3 in XML catalog... found
checking for DocBook XSL Stylesheets in XML catalog... not found
configure: error: could not find DocBook XSL Stylesheets in XML catalog
```

tschoonj:
I think docbook-xsl and scrollkeeper also need the `XML_CATALOG_FILES` variable in their formulas to work properly.

Try with this new commit.

DomT4:
```
checking for XML catalog... /usr/local/etc/xml/catalog
checking for xmlcatalog... /usr/local/opt/libxml2/bin/xmlcatalog
checking for DocBook XML DTD V4.3 in XML catalog... found
checking for DocBook XSL Stylesheets in XML catalog... found
==> Summary
🍺  /usr/local/Cellar/gtk-doc/1.24: 71 files, 2.7M, built in 37 seconds
```
:+1:

Library/Formula/docbook-xsl.rb
```diff
@@ -3,6 +3,7 @@ class DocbookXsl < Formula
   homepage "http://docbook.sourceforge.net/"
   url "https://downloads.sourceforge.net/project/docbook/docbook-xsl/1.78.1/docbook-xsl-1.78.1.tar.bz2"
   sha256 "c98f7296ab5c8ccd2e0bc07634976a37f50847df2d8a59bdb1e157664700b467"
+  revision 2
```
DomT4: Should be `revision 1` rather than `2`. 0 is the default but unwritten.

Library/Formula/scrollkeeper.rb
```diff
@@ -3,6 +3,7 @@ class Scrollkeeper < Formula
   homepage "http://scrollkeeper.sourceforge.net/"
   url "https://downloads.sourceforge.net/project/scrollkeeper/scrollkeeper/0.3.14/scrollkeeper-0.3.14.tar.gz"
   sha256 "4a0bd3c3a2c5eca6caf2133a504036665485d3d729a16fc60e013e1b58e7ddad"
+  revision 2
```
DomT4: Same here.

tschoonj:
sorry about that, had to do with my trial and error work on this formula last week

Library/Formula/docbook.rb
```diff
@@ -45,7 +45,14 @@ class Docbook < Formula
 
   def install
     (etc/"xml").mkpath
-    system "xmlcatalog", "--noout", "--create", "#{etc}/xml/catalog"
+
+    # only create catalog file if it doesn't exist already to avoid content added
+    # by other formulas to be removed
+    unless File.file?("#{etc}/xml/catalog")
```
xu-cheng: We can't relay on this. Especially for bottles. If this is really needed, it should be put in `post_install`.

DomT4: Potentially worth noting is that if this file does exist locally, the bottle installation won't overwrite it. Installs as `catalog.default` instead.

tschoonj:
so do I change the Formula or is it good as it is?

DomT4:
If it can be moved to `post_install` with the other "compile" steps, let's go for it. Might as well.

tschoonj:
pushed

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/fb6e21985c5347559cf4d4f5ddeec7b9ebb9b28e & https://github.com/Homebrew/homebrew/commit/87ce13db890037e444b6c68f0e5870eae4914190 & https://github.com/Homebrew/homebrew/commit/7d2f6c9f22114b9005ffdcba04c93adf5079882a & https://github.com/Homebrew/homebrew/commit/0aea7ab2a71aebd073851ca279b076bb1bd12a39.

In an ideal world there would be some squashing there, but not a huge deal this time. Will go restart the impacted builds elsewhere shortly. Thanks for getting this sorted Tom.

PR #37206, opened 2015-02-26
docker-compose 1.1.0

kriswill:
Fig renamed to "Docker Compose", version bump to 1.1.0

PR #44271, opened 2015-09-23
fig 1.4.2

Aigeruth:


xu-cheng:
Close in favor of #44273 

Aigeruth:
@xu-cheng I created my PR based on #37206. Has the renaming policy changed?

xu-cheng:
@Aigeruth We only add support for formula renaming recently. If you are interested, [this](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Rename-A-Formula.md) is how we handle it now.

Aigeruth:
@xu-cheng Thank you!

PR #44273, opened 2015-09-23
docker-compose: renamed from fig

alex:

PR #44284, opened 2015-09-23
docker-compose 1.4.2

ryanfb:


DomT4:
See #44273. Thanks for the submission though!

PR #44285, opened 2015-09-23
juju 1.24.6

sinzui:


DomT4:
It doesn't understand El Cap apparently. This may need reporting upstream:

```
==> /usr/local/Cellar/juju/1.24.6/bin/juju version
panic: osVersion reported an error: unknown series ""
```

sinzui:
This is https://bugs.launchpad.net/juju-core/+bug/1465317 where I predicted the failure. The Juju engineers are rethinking os versions to ensure Juju continues  to work after an upgrade. I think we need to press them for a hack as was done for yosemite.

DomT4:
@sinzui Is there any room to patch this ourselves and submit that upstream?

sinzui:
This issue looks be fixed in a future version. I considering a patch solution.

DomT4:
If upstream have applied or will accept a patch, we'll be happy to take that. Otherwise we'd need to halt people installing this on El Cap until the next release, which is an option, but not a great one.

sinzui:
I added a patch to register El Capitan to the supported versions. I will propose this patch for Juju 1.24.7 just in case it is ever released. Future juju 1.25 does not require OS versions to be registered.

DomT4:
Thanks! Is there an upstream patch submission or something we can link to to track?

sinzui:
The upstream pull request is at https://github.com/juju/juju/pull/3388 and approved for merge. It may be merged later today.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/013ef6fd30223cf0cc9725b524386ae185944be0. Thank you for creating the patch upstream & your contribution to Homebrew!

PR #44294, opened 2015-09-24
fig: 1.4.2

ruimarinho:
Reference: https://github.com/docker/compose/releases/tag/1.4.2


DomT4:
See #44273. Thanks for the submission though!

PR #44296, opened 2015-09-24
minidlna: fix broken sample conf

DomT4:
It seems unlikely users actually want their config which we instruct them to use to read:
```
friendly_name=Mac DLNA Server
media_dir=/private/tmp/minidlna20150920-86661-1e1sfik/minidlna-1.1.5/.brew_home/.config/minidlna/media
db_dir=/private/tmp/minidlna20150920-86661-1e1sfik/minidlna-1.1.5/.brew_home/.config/minidlna/cache
log_dir=/private/tmp/minidlna20150920-86661-1e1sfik/minidlna-1.1.5/.brew_home/.config/minidlna
```
:smile_cat:


mikemcquaid:
:+1:

PR #44300, opened 2015-09-24
Add the Apache Flink formula

tanbamboo:
platform for scalable batch and stream data processing.

http://flink.apache.org

### DIFF
--- a/Library/Formula/flink.rb
+++ b/Library/Formula/flink.rb
@@ -1,0 +1,18 @@
+class Flink < Formula
+  desc "Platform for scalable batch and stream data processing"
+  homepage "https://flink.apache.org"
+  head "https://github.com/apache/flink.git"
+  url "https://www.apache.org/dyn/closer.cgi?path=flink/flink-0.9.0/flink-0.9.0-bin-hadoop2.tgz"
+  version "0.9.0"
+  sha256 "1edf33d154b863dd108ba618ed3b7dce3fce6c116b45368aef164b47779b65f9"
+
+  def install
+    rm_f Dir["bin/*.cmd"]
+    rm_f Dir["bin/*.bat"]
+    libexec.install Dir["*"]
+    bin.write_exec_script Dir["#{libexec}/bin/*"]
+  end
+
+  test do
+  end
+end
--- a/Library/Formula/flink.rb
+++ b/Library/Formula/flink.rb
@@ -15,7 +15,7 @@ class Flink < Formula
 
   test do
     system "#{bin}/start-local.sh"
-    system "#{bin}/flink list"
+    system "#{bin}/flink" "list"
     system "#{bin}/stop-local.sh"
   end
 end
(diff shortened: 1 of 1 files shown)


Library/Formula/flink.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def install
+    rm_f Dir["bin/*.cmd"]
+    rm_f Dir["bin/*.bat"]
+    libexec.install Dir["*"]
+    bin.write_exec_script Dir["#{libexec}/bin/*"]
+  end
+
+  test do
+  end
```
bfontaine: Please add a test.

Library/Formula/flink.rb
```diff
@@ -0,0 +1,18 @@
+class Flink < Formula
+  desc "Platform for scalable batch and stream data processing"
+  homepage "https://flink.apache.org"
+  head "https://github.com/apache/flink.git"
+  url "https://www.apache.org/dyn/closer.cgi?path=flink/flink-0.9.0/flink-0.9.0-bin-hadoop2.tgz"
```
bfontaine: Can’t we build from source?

tanbamboo: sorry, i am not familiar with brew formula, just copy from apache-spark and modify to apache flink.

Library/Formula/flink.rb
```diff
@@ -0,0 +1,18 @@
+class Flink < Formula
+  desc "Platform for scalable batch and stream data processing"
+  homepage "https://flink.apache.org"
+  head "https://github.com/apache/flink.git"
```
DomT4: `head` should be below the `sha256` line.

Library/Formula/flink.rb
DomT4: It's probably fine given it's a Java thing. Presumably does need a `:java` dependency though?

Library/Formula/flink.rb
DomT4: Ping on @bfontaine's comment here.

tanbamboo: @bfontaine @DomT4 have add a test to running flink in local mode.

tanbamboo: @bfontaine @DomT4 seems can't access /usr/local/Cellar/flink/0.9.0/libexec/bin/../log, maybe need to change the default configuration?

Library/Formula/flink.rb
```diff
@@ -0,0 +1,21 @@
...
+  def install
+    rm_f Dir["bin/*.cmd"]
+    rm_f Dir["bin/*.bat"]
+    libexec.install Dir["*"]
+    bin.write_exec_script Dir["#{libexec}/bin/*"]
+  end
+
+  test do
+    system "#{bin}/start-local.sh"
+    system "#{bin}/flink list"
```
bfontaine: Please use `system "#{bin}/flink", "list"`

Library/Formula/flink.rb
DomT4: That'll be a sandboxing issue, but it should be trying to write logs and persistent similar files to `HOMEBREW_PREFIX/var` rather than `libexec`. Maybe we can do something with an `inreplace` or similar.

DomT4:
Does this need a Java dep at all?

bfontaine:
Ping?

PR #44302, opened 2015-09-24
travis: slight tweak

xu-cheng:


.travis.yml
```diff
@@ -26,7 +27,11 @@ before_install:
   - git clean -qxdff
 
 script:
-  - brew test-bot
+  - if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then
```
mikemcquaid: Can we push this logic to the `test-bot` itself?

mikemcquaid:
:+1: on the other changes

.travis.yml
xu-cheng: Done.

Thought, I feel like we should sort out these logics (but I have no clue what purpose for each branches are) https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/test-bot.rb#L302-L355

mikemcquaid: Yeh, agreed. Basically they are so `test-bot` can use environment variables, a specified commit, a PR URL, a formula etc. as input.

mikemcquaid:
:shipit: when :green_apple:

Issue #44288, opened 2015-09-23
add cleanup option to brew update


EmilVarona:
I had noticed that running `brew upgrade` does not remove outdated formulae. It might be useful to have an optional switch when running `brew upgrade` to perform this rather than relying on a separate call to `brew cleanup`.

bfontaine:
Would you mind doing a pull-request? This should help and I’d be happy to answer any question you might have: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md

rakkesh:
@bfontaine could you kindly share your thoughts on this approach.
```rb
require "cmd/cleanup"
   ....
    outdated.each do |f|
      upgrade_formula(f)
      cleanup_formula(f) if ARGV.include? '--cleanup' 
    end
```

bfontaine:
@rakkesh It seems great, see my comment on the PR.

DomT4:
PR was merged.

PR #44305, opened 2015-09-24
brew upgrade: provides an option `--cleanup`

rakkesh:
PR addresses issue https://github.com/Homebrew/homebrew/issues/44288. If this is an acceptable approach, I will update the docs as well.

### DIFF
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -1,5 +1,6 @@
 require "cmd/install"
 require "cmd/outdated"
+require "cmd/cleanup"
 
 module Homebrew
   def upgrade
@@ -41,7 +42,10 @@ module Homebrew
       puts pinned.map { |f| "#{f.full_name} #{f.pkg_version}" } * ", "
     end
 
-    outdated.each { |f| upgrade_formula(f) }
+    outdated.each do |f|
+      upgrade_formula(f)
+      cleanup_formula(f) if ARGV.include?("--cleanup") && f.installed?
+    end
   end
 
   def upgrade_pinned?
--- a/Library/Homebrew/manpages/brew.1.md
+++ b/Library/Homebrew/manpages/brew.1.md
@@ -476,10 +476,12 @@ Note that these flags should only appear after a command.
 
     If `--rebase` is specified then `git pull --rebase` is used.
 
-  * `upgrade [install-options]` [<formulae>]:
+  * `upgrade [install-options] [--cleanup]` [<formulae>]:
     Upgrade outdated, unpinned brews.
 
     Options for the `install` command are also valid here.
+
+    If `--cleanup` is specified then remove previously installed <formula> version(s).
 
     If <formulae> are given, upgrade only the specified brews (but do so even
     if they are pinned; see `pin`, `unpin`).
--- a/share/doc/homebrew/brew.1.html
+++ b/share/doc/homebrew/brew.1.html
@@ -116,6 +116,14 @@ instead of the intersection.</p>
 <p>By default, <code>deps</code> shows dependencies for <var>formulae</var>. To skip the <code>:build</code>
 type dependencies, pass <code>--skip-build</code>. Similarly, pass <code>--skip-optional</code>
 to skip <code>:optional</code> dependencies.</p></li>
+<li><p><code>desc</code> <var>formula</var>:
+Display <var>formula</var>'s name and one-line description.</p></li>
+<li><p><code>desc [-s|-n|-d] &lt;pattern></code>:
+Search both name and description (<code>-s</code>), just the names (<code>-n</code>), or just  the
+descriptions (<code>-d</code>) for <code>&lt;pattern></code>. <code>&lt;pattern></code> is by default interpreted
+as a literal string; if flanked by slashes, it is instead interpreted as a
+regular expression. Formula descriptions are cached; the cache is created on
+the first search, making that search slower than subsequent ones.</p></li>
 <li><p><code>diy [--name=&lt;name>] [--version=&lt;version>]</code>:
 Automatically determine the installation prefix for non-Homebrew software.</p>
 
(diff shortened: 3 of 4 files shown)


mikemcquaid:
Makes sense to me but will let other maintainers give thoughts.

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -41,7 +42,10 @@ def upgrade
       puts pinned.map { |f| "#{f.full_name} #{f.pkg_version}" } * ", "
     end
 
-    outdated.each { |f| upgrade_formula(f) }
+    outdated.each do |f|
+      upgrade_formula(f)
+      cleanup_formula(f) if ARGV.include? "--cleanup"
```
bfontaine: We might want to cache this value, e.g.:

```rb
cleanup = ARGV.include? "--cleanup"
outdated.each do |f|
  upgrade_formula(f)
  cleanup_formula(f) if cleanup
end
```

That’s probably unnecessary optimization thought.

bfontaine:
:+1:

xu-cheng:
Would be great to update manpage at the same time.

bfontaine:
To update the manpage, edit the [`Library/Homebrew/manpages/brew.1.md`](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/manpages/brew.1.md) file and run `brew man`.

Library/Homebrew/cmd/upgrade.rb
mikemcquaid: Yeh, I think that's overkill personally.

rakkesh:
Thank you all for your reviews. Let me know, if anything else needs to be done. 
By the way, should I squash these changes into a single commit?

bfontaine:
@rakkesh Yes that’d be great :+1:

Library/Homebrew/manpages/brew.1.md
```diff
@@ -476,7 +476,7 @@ Note that these flags should only appear after a command.
 
     If `--rebase` is specified then `git pull --rebase` is used.
 
-  * `upgrade [install-options]` [<formulae>]:
+  * `upgrade [install-options] --cleanup` [<formulae>]:
```
bfontaine: it should be between brackets since it‘s optional: `[--cleanup]`.

rakkesh:
@bfontaine It seems, I did not follow the format used in `man brew` with the option, my bad. Also, the option `--cleanup` be mentioned ahead of the `formulae` description. I'll make the changes now.

EmilVarona:
Thanks all and @rakkesh for making this happen.

aknuds1:
Thanks guys, was missing this.

Library/Homebrew/manpages/brew.1.md
```diff
@@ -476,11 +476,13 @@ Note that these flags should only appear after a command.
 
     If `--rebase` is specified then `git pull --rebase` is used.
 
-  * `upgrade [install-options]` [<formulae>]:
+  * `upgrade [install-options] [--cleanup]` [<formulae>]:
     Upgrade outdated, unpinned brews.
 
     Options for the `install` command are also valid here.
 
+    If `--cleanup` is passed, remove existing <formula> version(s) in Cellar.
```
bfontaine: I’d say “remove the previous version(s) of <formula>” or something like that to be clear that only older versions are removed.

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -41,7 +42,10 @@ def upgrade
       puts pinned.map { |f| "#{f.full_name} #{f.pkg_version}" } * ", "
     end
 
-    outdated.each { |f| upgrade_formula(f) }
+    outdated.each do |f|
+      upgrade_formula(f)
```
xu-cheng: We may need to tweak `upgrade_formula` to make it return success or failure. So we only cleanup formula which succeeds in upgrading.

Library/Homebrew/manpages/brew.1.md
rakkesh: I agree, that would be clear to understand.

Library/Homebrew/cmd/upgrade.rb
rakkesh: If an upgrade fails, it would re-link the existing formula, and anyway, `cleaup_formula` checks for an eligible version to remove.

rakkesh:
I apologize for the delay in changing the text in `brew man`. I am going to do that now but wanted to check with you all about the text - ```If `--cleanup` is passed, remove previous <formula> version(s) in Cellar, if a <formula> is successfully upgraded.```. Is this clear enough to convey the `option` functionality?

rakkesh:
The above text is not entirely true. As an example, if we have 2 older versions of node installed, e.g., `node-4.0.0` and `node-4.1.0` and upgrading to `node-4.1.1`, even if the upgrade fails, `cleanup_formula` will remove `node-4.0.0` but `node-4.1.0` will still be in place and linked. Does this work everyone or do we call `cleanup_formula` only if a "tweaked" `upgrade_formula` returns success, as suggested by @xu-cheng. 

Anyway, I am going to stick with the text what @bfontaine has suggested - ```If `--cleanup` is passed, remove previous <formula> version(s) in Cellar```.

bfontaine:
I think you could even remove “in Cellar”: “… remove previous <formula> version(s)”.

> I apologize for the delay in changing the text in `brew man`.

No worries :smile:

Library/Homebrew/manpages/brew.1.md
DomT4: I don't know if it's worth mentioning that if you've an old version linked it won't be cleaned up. We may already mention it elsewhere, so possibly redundant.

Library/Homebrew/cmd/upgrade.rb
xu-cheng: I just found out that you don't have to tweak `upgrade_formula`. Instead you can tweak the next line to:
```ruby
cleanup_formula(f) if ARGV.include?("--cleanup") && f.installed?
```

rakkesh: I don't see any benefit of adding the clause `&& f.installed?`, as a [check](https://github.com/Homebrew/homebrew/blob/88ed0ee3f6f903bde4ddd674196e2a9afdc8019f/Library/Homebrew/cmd/upgrade.rb#L17-L22) is done prior to calling `upgrade_formula`. So, `if ARGV.include?("--cleanup") && f.installed?` would always be true.

xu-cheng: `f.installed?` will only return true when the **latest** version is installed which means upgrading is success.

rakkesh: @xu-cheng I had missed your point earlier, I get it know. 👍🏽 I'll update the commit with your suggestion. Thanks for the reviews the code .

xu-cheng:

PR #43658, opened 2015-09-07
Adding miller (mlr) formula

natacado:
Note that there is another open pull request for Miller (https://github.com/Homebrew/homebrew/pull/43099) that's also open; this one pulls in a much more recent version of miller, passes brew audit, and has a decent test case.

PR #44307, opened 2015-09-24
miller 2.2.1 (new formula)

DomT4:
Replaces the seemingly dead #43658 & #43099.

### DIFF
--- a/Library/Formula/miller.rb
+++ b/Library/Formula/miller.rb
@@ -1,0 +1,33 @@
+class Miller < Formula
+  desc "Like sed, awk, cut, join & sort for name-indexed data such as CSV"
+  homepage "https://github.com/johnkerl/miller"
+  url "https://github.com/johnkerl/miller/releases/download/v2.2.1/mlr-2.2.1.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/miller.rb
```diff
@@ -0,0 +1,33 @@
...
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.csv").write <<-EOS.undent
+      a,b,c
+      1,2,3
+      4,5,6
+    EOS
+    output = pipe_output("#{bin}/mlr --csv --rs lf cut -f a test.csv")
```
nbeyer: According to the latest docs, the `--rs` argument is supposed to be ignored for `--csv` argument. There may be a bug and it's not being ignored, but if that gets fixed, the tests could be made more resilient by using `--csvlite` or putting CRLF line endings in the CSV fixtures.

DomT4: Good shout on using `--csvlite`. Will roll that & cut the `--rs` call.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/566036b16e043cd2d1fac8d521451dae7fe566da :tada:.

PR #44308, opened 2015-09-24
gws 0.1.8

frodeaa:
the formula adds patch to the gws script to avoid bug introduced in 0.1.8 (https://github.com/StreakyCobra/gws/issues/17)

the patch updates the 'PATH' with GNU's 'sed' and 'cut'

### DIFF
--- a/Library/Formula/gws.rb
+++ b/Library/Formula/gws.rb
@@ -1,16 +1,37 @@
 class Gws < Formula
   homepage "https://streakycobra.github.io/gws/"
   desc "Manage workspaces composed of git repositories"
-  url "https://github.com/StreakyCobra/gws/archive/0.1.7.tar.gz"
-  sha256 "36c25392f5cb93a39441a1973842249a62b605a2a7e369a040c78450683697d0"
+  url "https://github.com/StreakyCobra/gws/archive/0.1.8.tar.gz"
+  sha256 "c240601b0adcc5ae402199217fc0d5fd6775ada69919a860ad3c1c4b16805e63"
 
   depends_on "bash"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
I'd like to see a pull-request submitted upstream letting this use the OS X utilities as well (so we can apply that and remove the dependencies).

frodeaa:
Hopefully the bug should be fixed in the next version and we can remove the GNU dependencied

Library/Formula/gws.rb
```diff
@@ -1,18 +1,39 @@
...
   depends_on "bash"
+  depends_on "coreutils" # see bug https://github.com/StreakyCobra/gws/issues/17
+  depends_on "gnu-sed" # see bug https://github.com/StreakyCobra/gws/issues/17
 
   def install
     bin.install "src/gws"
 
+    # Add GNU's 'sed' and 'cut' to PATH to fix bug
+    inreplace "#{bin}/gws" do |s|
+      s.sub! /VERSION="0.1.8"/, "VERSION=\"0.1.8\"\nexport PATH=#{gws_gnu_path}"
```
dunn: We might hold off until https://github.com/StreakyCobra/gws/issues/17 is resolved, but an easier way to do this is with a wrapper script, like git-ftp does: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/git-ftp.rb#L29-L33

bfontaine:
Ping?

DomT4:
Closing as dead. Please feel free to resubmit as fresh with the comments above addressed and audit passing, if you'd like to. Thanks for the submission to Homebrew!

PR #44315, opened 2015-09-25
shellcheck 0.4.1

DomT4:


mikemcquaid:
:+1: and fixed audit in 4f01131.

DomT4:
The one PR I didn't audit before submitting. Should have known that'd bite me. Thanks for fixing!

PR #44325, opened 2015-09-25
Formula installer bug fix

xu-cheng:
This should help to solve https://github.com/Homebrew/homebrew/pull/44231#issuecomment-142518526

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -56,6 +56,10 @@ class FormulaInstaller
     @pour_failed   = false
   end
 
+  def skip_deps_check?
+    ignore_deps?
+  end
+
   # When no build tools are available and build flags are passed through ARGV,
   # it's necessary to interrupt the user before any sort of installation
   # can proceed. Only invoked when the user has no developer tools.
@@ -97,7 +101,7 @@ class FormulaInstaller
   end
 
   def prelude
-    verify_deps_exist unless ignore_deps?
+    verify_deps_exist unless skip_deps_check?
     lock
     check_install_sanity
   end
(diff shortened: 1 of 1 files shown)


Library/Homebrew/formula_installer.rb
```diff
@@ -97,7 +97,7 @@ def install_bottle_for?(dep, build)
   end
 
   def prelude
-    verify_deps_exist unless ignore_deps?
+    verify_deps_exist unless ignore_deps? || is_a?(DependencyInstaller)
```
mikemcquaid: Mind making this a `if`? I always find `unless ... or` hard to parse.

mikemcquaid: And same below. I wonder if you want a dedicated method for them all?

mikemcquaid:
Nice work :+1:

Library/Homebrew/formula_installer.rb
xu-cheng: What method name should we call it? How about `skip_deps_check?`? So we will have:
```ruby
class FormulaInstaller
  def skip_deps_check?
    ignore_deps?
  end
end

class DependencyInstaller < FormulaInstaller
  def skip_deps_check?
    true
  end
end
```

mikemcquaid: Yeh, something like that looks great :+1:

xu-cheng:
Updated, also fixed a new bug: prefix is not cleanup when pouring bottle failed.

mikemcquaid:
Looking good :+1:

xu-cheng:
@mikemcquaid Travis seems to be confused on which commits to test. Instead of testing this PR, it's testing the formula updated in current time period  https://travis-ci.org/Homebrew/homebrew/jobs/82146619

mikemcquaid:
@xu-cheng you can look at the:
```
name: Homebrew/homebrew/pull/44325-ff232e29f2758c887b56dc331999e6b737ad8565
url: https://github.com/Homebrew/homebrew/pull/44325
hash: 
diff_start_sha1: 628bde1ca9ad50715b7e394ac0381e7e40e22a35
diff_end_sha1: ff232e29f2758c887b56dc331999e6b737ad8565
```

debugging information which should help debug that.

xu-cheng:
Will looking into it tomorrow.

Issue #43862, opened 2015-09-12
mtr needs root and the colors are wrong


5c0tt:
the homebrew mtr has two problems.  The first, all of a sudden in one of the updates, it takes over the colors of your shell.  It used to inherit the colors of your shell, or go to a nice black screen if you had no color theme set up.

It also did not used to need root.  Or perhaps I should say, I used to run the non homebrew version, and hand installed it.  Which I have done again.  The binary is setuid root do you don't have to `sudo mtr`.  Is it adjacent homebrew policy to setuid?

At the very least, there is a bug in how it deals with color.

5c0tt:
I found that this is also an issue in source 0.85 is ok, is started in 0.86, and MacPorts has added this patch to their system: https://github.com/traviscross/mtr/commit/63a1f1493bfbaf7e55eb7e20b3791fc8b14cf92d

Can someone who knows how add this patch as well?

mikemcquaid:
@5c0tt Homebrew doesn't `setuid` and Homebrew can't even if we did as it isn't run as root. The colour bug seems to be an upstream bug. Is that commit in the upstream repository? If so, we can add it as a `patch do`. Could you try and open a pull request? This should help and I'm happy to walk you through anything else: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged

Thanks!

5c0tt:
I just installed 0.85 and the color bug is not in it. I don't know if the patch is upstream.  Mac Ports has an ability to apply a patch to source from their ports files, which I assume are a lot like the ruby files that brew uses.  In this case, I just uninstalled mtr from brew and installed version 0.85 by hand.

Hoever, I am pretty green to git, and don't know all the terms. I don't know if that is an upstream repo, nor do I know how to find out, I have the link that i found in google, and that is it, just the same data as you.

It looks like brew can take a patch the same way, so the patch code is there, but I would have no idea how to open a pull request.

I see you linked to instructions, which I will try to follow. Though this is about 5 lines, it may be simpler for you to just make the pull request, pop the code change in, and merge it with brew, or whatever it is that happens.

I will post back here if I make any progress, it will be later this evening.

mikemcquaid:
You can apply `https://github.com/traviscross/mtr/commit/63a1f1.patch` in a `patch do` block.

[...]

PR #44328, opened 2015-09-25
mtr: use-default-colors

rakkesh:
Closes issue - https://github.com/Homebrew/homebrew/issues/43862


rakkesh:
Moved `depends_on "automake" => :build` from `head` block as it is required after applying the patch.

PR #44329, opened 2015-09-25
migrator: link Cellar and opt before the keg.

### LINKED ISSUE #44306: Formula renames break `outdated` and `upgrade`?
```
tim@rocketman:~$ brew update
Updated Homebrew from 6072e88b to e12b24d8.
...
==> Renamed Formulae
fig -> docker-compose

tim@rocketman:~$ brew outdated
Error: No such file or directory - /usr/local/Cellar/docker-compose

tim@rocketman:~$ brew upgrade
Error: No such file or directory - /usr/local/Cellar/docker-compose

tim@rocketman:~$ ls /usr/local/Cellar/fig
1.3.3 1.4.0 1.4.1
```

What should have happened here? One potential complicating factor is that I haven't run `brew update` in a while since I use `git pull` a lot.

mikemcquaid:
When there's absolute symlinks in a linked directory in the `keg`
(e.g. `bin`)that point to the `Cellar` or `opt` then linking the `keg`
will fail before the `Cellar` or `opt` has been linked.

CC @vladshablinsky @tdsmith @xucheng

Closes #44306.

### DIFF
--- a/Library/Homebrew/migrator.rb
+++ b/Library/Homebrew/migrator.rb
@@ -163,9 +163,9 @@ class Migrator
       unlink_oldname
       move_to_new_directory
       repin
-      link_newname unless old_linked_keg.nil?
-      link_oldname_opt
       link_oldname_cellar
+      link_oldname_opt
+      link_newname unless old_linked_keg.nil?
       update_tabs
     rescue Interrupt
       ignore_interrupts { backup_oldname }


vladshablinsky:
:+1: I like the solution of the bug. Didn't think about the case with absolute symlinks inside kegs. I Tried to reproduce the behaviour -- I have relative ones.

mikemcquaid:
@vladshablinsky Are there any other issues with reordering these e.g. will these symlinks be left over if there's a failure?

mikemcquaid:
@vladshablinsky You can reproduce by doing `git checkout ec0da216ad7150812489425ba476c7024a2433ea`

vladshablinsky:
I think no. Here is what we do in `backup_oldname`
```ruby
  def backup_oldname
    unlink_oldname_opt
    unlink_oldname_cellar
    backup_oldname_cellar
    backup_old_tabs
    # ... 
  end
```

so we delete all these symlink anyway.

mikemcquaid:
Great, thanks!

PR #44333, opened 2015-09-25
passenger 5.0.20

FooBarWidget:


DomT4:
Test is failing:

```
Error: passenger: failed
<"/usr/local/Cellar/passenger/5.0.20/libexec/lib"> expected but was
<"/usr/local/Cellar/passenger/5.0.20/libexec/src/ruby_supportlib">.
```

FooBarWidget:
Oh oops, forgot to update the test. Will fix that.

FooBarWidget:
Fixed.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/b487a94a648aee4b5437f15eb9983f548a4fd44e. Thank you for your contribution to Homebrew @FooBarWidget!

lhoBas:
Heads up: the changed location.ini file path will cause issues for existing nginx + passenger installations using homebrew. I've updated the caveat in nginx.rb to report the new path.

*Users with existing installations will have to make a change to their* `/usr/local/etc/nginx/nginx.conf`:
`passenger_root` should be set to `/usr/local/opt/passenger/libexec/src/ruby_supportlib/phusion_passenger/locations.ini` starting with passenger 5.0.20.

(If your homebrew installation is not in `/usr/local`, your path will differ.)

FooBarWidget:
Maybe I should make a symlink at the old location for backward compatibility? What do you think?

PR #42662, opened 2015-08-08
cmd/search: bold installed formulae

dunn:
`brew search emacs`, with homebrew/emacs tapped:
```
real    0m2.586s
user    0m1.138s
sys     0m0.121s
```
Previously:
```
real    0m2.831s
user    0m1.385s
sys     0m0.087s
```

`brew search s`:
```
real    0m2.260s
user    0m1.201s
sys     0m0.114s
```

Previously:
```
real    0m5.953s
user    0m4.827s
sys     0m0.226s
```
I'm not sure what's responsible for this improvement. 💁

Closes #35696.

PR #44335, opened 2015-09-25
update: highlight installed formulae in report

UniqMartin:
I think it's nice to be able to see instantly which of the updated formulae are installed on the user's system. (Those are highlighted with an appended asterisk.) It's a bit like `brew outdated`, only it doesn't bother you with all the formulae you haven't updated yet while waiting for *that one awesome update*.

### DIFF
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -411,10 +411,33 @@ class Report
 
   def dump_formula_report(key, title)
     formula = select_formula(key)
-    formula.map! { |oldname, newname| "#{oldname} -> #{newname}" } if key == :R
     unless formula.empty?
+      # Determine list item indices of installed formulae.
+      formula_installed_index = formula.each_index.select do |index|
+        name, newname = formula[index]
+        installed?(name) || (newname && installed?(newname))
+      end
+
+      # Format list items of renamed formulae.
+      if key == :R
+        formula.map! { |oldname, newname| "#{oldname} -> #{newname}" }
+      end
+
+      # Append suffix '(installed)' to list items of installed formulae.
+      formula_installed_index.each do |index|
+        formula[index] += " (installed)"
+      end
+
+      # Fetch list items of installed formulae for highlighting.
+      formula_installed = formula.values_at(*formula_installed_index)
+
+      # Dump formula list.
       ohai title
-      puts_columns formula
+      puts_columns(formula, formula_installed)
     end
+  end
+
+  def installed?(formula)
+    (HOMEBREW_CELLAR/formula.split("/").last).directory?
   end
 end


Library/Homebrew/cmd/update.rb
```diff
@@ -398,10 +398,26 @@ def select_formula(key)
...
+    formula.
+      select { |oldname, newname| installed?(oldname) || installed?(newname) }.
+      map { |oldname, newname| "#{oldname} -> #{newname}" }
+  end
+
+  def installed?(formula)
+    f = Formulary.factory(formula)
+    f.installed?
+  rescue
+    false
```
xu-cheng: Let's use `(HOMEBREW_CELLAR/formula).directory?` which won't load formula hence better performance.

Library/Homebrew/cmd/update.rb
```diff
@@ -398,10 +398,26 @@ def select_formula(key)
 
   def dump_formula_report(key, title)
     formula = select_formula(key)
+    formula_installed = select_installed_formula(key, formula)
     formula.map! { |oldname, newname| "#{oldname} -> #{newname}" } if key == :R
     unless formula.empty?
       ohai title
-      puts_columns formula
+      puts_columns(formula, formula_installed)
```
xu-cheng: You may want to update `puts_columns` as well. Currently, it only appends a `*` for highlight entires. You may want to check it using [`Tty`](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/utils.rb#L11-L71)

Library/Homebrew/cmd/update.rb
UniqMartin: Yes, seems wasteful to load the entire formula for such a simple check. So something like `(HOMEBREW_CELLAR / formula.split("/", 3).last).directory?` to also get it right for taps?

xu-cheng: :+1:

Library/Homebrew/cmd/update.rb
UniqMartin: Have you been reading my mind? ;) I'd like to have `Tty` color support in `puts_columns`, but that's not a trivial change because `puts_columns` is using an external command (`pr`) for the multi-column output. And `pr` doesn't play nicely with ANSI escape sequences.

I have a working prototype and would prefer to contribute it in a separate PR, because it's completely independent of this change.

Library/Homebrew/cmd/update.rb
```diff
@@ -398,10 +398,26 @@ def select_formula(key)
...
     end
   end
+
+  def select_installed_formula(key, formula)
+    return formula.select { |name| installed?(name) } unless key == :R
+
+    formula.
+      select { |oldname, newname| installed?(oldname) || installed?(newname) }.
+      map { |oldname, newname| "#{oldname} -> #{newname}" }
+  end
```
xu-cheng: I'm wondering can we get ride of this method entirely, like:
```ruby
def dump_formula_report(key, title)
  formula = select_formula(key)
  formula_installed = formula.select do |name, newname|
    installed?(name) || (newname && installed?(newname))
  end
  unless formula.empty?
    ohai title
    if key == :R
      formula.map! { |oldname, newname| "#{oldname} -> #{newname}" }
      formula_installed.map! { |oldname, newname| "#{oldname} -> #{newname}" }
    end
    puts_columns(formula, formula_installed)
  end
end
```

Library/Homebrew/cmd/update.rb
xu-cheng: > contribute it in a separate PR 

Good idea.

xu-cheng:
FYI, we use `foo (installed)` notation in `brew search`. It would be nice we make the format consistence throughout all brew commands. So in the end, this is a question on how should we display the result. Let's see what other maintainers' opinion. cc @dunn as I remember you have experimented such style problems before.

Library/Homebrew/cmd/update.rb
UniqMartin: Thanks for this thorough feedback! Works for me and it is wonderfully concise. May I “steal” this code and make it part of my PR next time I push?

UniqMartin:
> FYI, we use `foo (installed)` notation in `brew search`. It would be nice we make the format consistence throughout all brew commands. So in the end, this is a question on how should we display the result.

I wasn't aware of how `brew search` formats its output, but I agree that we should be consistent here and an appended `(installed)` is certainly a lot more descriptive than an asterisk.

dunn:
I initially appended ` (installed)` and also bolded installed entries in the search results list, but there were problems with the columns display that I couldn't figure out how to fix: https://github.com/Homebrew/homebrew/pull/42662#issuecomment-129078548

UniqMartin:
@dunn Ultimately, I'd like to see the installed formulae to be also highlighted (e.g. bolded), but that requires changes to `puts_columns`. I started #44343 to lay the foundation for that. Once that one is merged, I'd like to implement highlighting on top of that. If interested, have a look. :)

Here's what the final result could look like (hacked together first version):
Highlighted items in multi-column output.

UniqMartin:
Just pushed a revised commit that incorporates the feedback. I added comments to make things clearer, but I could imagine this is seen as too verbose. The code works as is, but if/when #44343 lands, output will be so much nicer. :sparkles: Thus, it wouldn't be unreasonable to wait a little bit.

Library/Homebrew/cmd/update.rb
```diff
@@ -411,10 +411,33 @@ def select_formula(key)
 
   def dump_formula_report(key, title)
     formula = select_formula(key)
-    formula.map! { |oldname, newname| "#{oldname} -> #{newname}" } if key == :R
     unless formula.empty?
+      # Determine list item indices of installed formulae.
+      formula_installed = formula.each_index.select do |index|
```
xu-cheng: I think `formula_installed_index` may be a better name here.

Library/Homebrew/cmd/update.rb
```diff
@@ -411,10 +411,33 @@ def select_formula(key)
...
+
+      # Dump formula list.
       ohai title
-      puts_columns formula
+      puts_columns(formula, formula_installed)
     end
   end
+
+  def installed?(formula)
+    (HOMEBREW_CELLAR / formula.split("/", 3).last).directory?
```
xu-cheng: `(HOMEBREW_CELLAR/formula.split("/").last).directory?`

xu-cheng:
LGTM.

Library/Homebrew/cmd/update.rb
UniqMartin: I fully agree and will change this shortly.

Library/Homebrew/cmd/update.rb
UniqMartin: My intention was to make the behavior of `split` a bit more robust and I have to admit, that I'm still having a hard time leaving out spaces around binary operators (though here it's a operation on path). But I'll happily change this as suggested.

UniqMartin:
Thanks again for the feedback, @xu-cheng! I just included your latest suggestions and force pushed again.

UniqMartin:
#44343 has landed. I think this PR is ready to follow suit.

mikemcquaid:
Thanks again @UniqMartin!

PR #44337, opened 2015-09-25
Adding brotli formula

joeyhoer:
Adding [brotli](https://github.com/google/brotli) formula.

### DIFF
--- a/Library/Formula/brotli.rb
+++ b/Library/Formula/brotli.rb
@@ -1,8 +1,9 @@
 class Brotli < Formula
+  desc "Generic-purpose lossless compression algorithm"
   homepage "https://github.com/google/brotli"
   url "https://github.com/google/brotli/archive/v0.2.0.tar.gz"
-  sha256 "634d1089ee21b35e0ec5066cb5e44dd097e04e679e1e8c50bffa2b0dc77c2c29"
   version "0.1.0"
+  sha256 "634d1089ee21b35e0ec5066cb5e44dd097e04e679e1e8c50bffa2b0dc77c2c29"
 
   head "https://github.com/google/brotli.git"
 
--- a/Library/Formula/brotli.rb
+++ b/Library/Formula/brotli.rb
@@ -1,0 +1,22 @@
+class Brotli < Formula
+  homepage "https://github.com/google/brotli"
+  url "https://github.com/google/brotli/archive/v0.2.0.tar.gz"
+  sha256 "634d1089ee21b35e0ec5066cb5e44dd097e04e679e1e8c50bffa2b0dc77c2c29"
+  version "0.1.0"
+
+  head "https://github.com/google/brotli.git"
+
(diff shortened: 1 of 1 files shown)


Library/Formula/brotli.rb
```diff
@@ -0,0 +1,22 @@
+class Brotli < Formula
+  homepage "https://github.com/google/brotli"
+  url "https://github.com/google/brotli/archive/v0.2.0.tar.gz"
+  sha256 "634d1089ee21b35e0ec5066cb5e44dd097e04e679e1e8c50bffa2b0dc77c2c29"
+  version "0.1.0"
```
DomT4: Why `0.1.0`? The URL is `0.2.0`?

Library/Formula/brotli.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  head "https://github.com/google/brotli.git"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install "python/bro.py" => "bro"
```
DomT4: If this has a binary called `bro` you'll need to add the following line one empty line underneath the `depends_on`:

```ruby
conflicts_with "bro", :because => "Both install a `bro` binary"
```

joeyhoer:
I've made the change noted above.

DomT4:
Thanks. What's the situation with the version differences?

Library/Formula/brotli.rb
bfontaine: @DomT4 We would also need a similar line in the `bro` formulae:

```rb
conflicts_with "brotli", :because => "Both install a `bro` binary"
```

DomT4: Aye. Forgot about that briefly; I use conflicts mostly from Versions against core so got used to it being a one-way thing :smile_cat:.

dunn:
```
==> audit problems
brotli:
 * `version` (line 5) should be put before `checksum` (line 4)
 * Formula should have a desc (Description).
```

joeyhoer:
Despite the release being tagged 0.2.0, the tool reports 0.1.0 as the version.

DomT4:
That's probably worth asking upstream about to find out which version is "true".

DomT4:
Any chance you've managed to kick the version issue upstream yet so we know which version to use here?

joeyhoer:
Head was (updated 2-10-2015 to be v0.3.0)[https://github.com/google/brotli/pull/193#issuecomment-147414958] with no corresponding release. I expect that there will be some changes coming soon?

nitram509:
:+1: for this PR.
What is missing to merge it?

bfontaine:
Ping re Mike’s comments?

bfontaine:

PR #44339, opened 2015-09-25
Update docker-compose.rb

ryanmt:
Given the time since the name change, I think removing this explicit aliasing from the recipe will have less consequences.

For me, I prefer to handle aliases myself.


mikemcquaid:
Works for me :+1:

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/801d9a7032a81a99af4d91f4afb21f57e1f5f36c. Thank you for your contribution to Homebrew @ryanmt!

PR #44341, opened 2015-09-25
New formula : FStar 0.9.0

beurdouche:
New formula for the FStar compiler in stable version 0.9.0

### DIFF
--- a/Library/Formula/fstar.rb
+++ b/Library/Formula/fstar.rb
@@ -1,0 +1,94 @@
+class Fstar < Formula
+  desc "F*, an ML-like language with a type system for program verification"
+  homepage "http://fstar.gforge.inria.fr/"
+  url "https://github.com/FStarLang/FStar.git",
+    :tag => "v0.9.0",
+    :revision => "f24457213f4007302873f6d42eb2cf8094fa45d0"
+  head "https://github.com/FStarLang/FStar.git"
+
+  depends_on "mono"   => :build
+  depends_on "fsharp" => :build
+  depends_on "opam"   => :build
+  depends_on "ocaml"  => :recommended
+  depends_on "homebrew/science/z3"
+
+  resource "ocamlfind" do
+    url "http://download.camlcity.org/download/findlib-1.5.5.tar.gz"
+    sha256 "aafaba4f7453c38347ff5269c6fd4f4c243ae2bceeeb5e10b9dab89329905946"
+  end
+
+  resource "batteries" do
+    url "https://github.com/ocaml-batteries-team/batteries-included/archive/v2.3.1.tar.gz"
+    sha256 "df778b90fcdb26288d9d92a86e51dd75d6bb7c6e41888c748c7508e8ea58b1d4"
+  end
+
+  def install
+    ENV.deparallelize
+
+    opamroot = buildpath/"opamroot"
+    ENV["OPAMROOT"] = opamroot
+    ENV["OPAMYES"] = "1"
+    system "opam", "init", "--no-setup"
+    archives = opamroot/"repo/default/archives"
+    modules = []
+    resources.each do |r|
+      r.verify_download_integrity(r.fetch)
+      original_name = File.basename(r.url)
+      cp r.cached_download, archives/original_name
+      modules << "#{r.name}=#{r.version}"
+    end
+    system "opam", "install", *modules
+
+    system "make", "-C", "src/"
+    system "opam", "config", "exec", "--",
+    "make", "-C", "src/ocaml-output/"
+
+    bin.install "src/ocaml-output/fstar.exe"
+    prefix.install "README.md"
+
+    (libexec/"stdlib").install Dir["lib/*"]
+    (libexec/"contrib").install Dir["contrib/*"]
+    (libexec/"examples").install Dir["examples/*"]
+    (libexec/"tutorial").install Dir["doc/tutorial/*"]
+    (libexec/"src").install Dir["src/*"]
+    (libexec/"licenses").install "LICENSE-fsharp.txt", Dir["3rdparty/licenses/*"]
+
+    prefix.install_symlink libexec/"stdlib"
+    prefix.install_symlink libexec/"contrib"
+    prefix.install_symlink libexec/"examples"
+    prefix.install_symlink libexec/"tutorial"
+    prefix.install_symlink libexec/"src"
+    prefix.install_symlink libexec/"licenses"
+  end
+
+  def caveats; <<-EOS.undent
+    F* standard library is available in #{prefix}/stdlib:
+    - alias fstar='fstar.exe --include #{prefix}/stdlib --prims prims.fst'
+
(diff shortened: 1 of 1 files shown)


dunn:
At some point it's trying to install directly into `/usr/local`, instead of into the Cellar prefix:
```
==> make install
mkdir -p "/usr/local/opt/ocaml/bin"
mkdir -p "/usr/local/man"
mkdir: /usr/local/man: Operation not permitted
```
http://bot.brew.sh/job/Homebrew%20Pull%20Requests/33773/version=yosemite/testReport/junit/brew-test-bot/yosemite/install_fstar/

Library/Formula/fstar.rb
```diff
@@ -0,0 +1,69 @@
...
+    url "http://download.camlcity.org/download/findlib-1.5.5.tar.gz"
+    sha256 "aafaba4f7453c38347ff5269c6fd4f4c243ae2bceeeb5e10b9dab89329905946"
+  end
+
+  resource "batteries" do
+    url "https://github.com/ocaml-batteries-team/batteries-included/archive/v2.3.1.tar.gz"
+    sha256 "df778b90fcdb26288d9d92a86e51dd75d6bb7c6e41888c748c7508e8ea58b1d4"
+  end
+
+  def ocaml_install_package(*)
```
DomT4: Just in-line this. It's not huge enough or used often enough to need its own block.

```ruby
resource("ocamlfind").stage do
  system "./configure"
  system "make", "all"
  system "make", "install"
end

resource("batteries").stage do
  system "make", "all"
  system "make", "install"
end
```

Library/Formula/fstar.rb
```diff
@@ -0,0 +1,69 @@
+class Fstar < Formula
+  desc "F*, an ML-like language with a type system for program verification"
+  homepage "http://fstar.gforge.inria.fr/"
+  url "https://github.com/FStarLang/FStar.git", :tag => "v0.9.0",
```
DomT4: Please push `:tag` and `:revision` onto their own lines, i.e.

```ruby
url "https://ducks.duck.com/ducks.git",
     :tag => "v1.0.0",
     :revision => "a3a9181efef3824046cc0a9f55c7a7c37f102a02"
```

Library/Formula/fstar.rb
```diff
@@ -0,0 +1,69 @@
...
+                                                :revision => "f24457213f4007302873f6d42eb2cf8094fa45d0"
+  head "https://github.com/FStarLang/FStar.git", :branch => "master"
```
DomT4: Don't need to specify `master`. We assume that to be the upstream default.

Library/Formula/fstar.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  depends_on "mono"
+  depends_on "fsharp"
+  depends_on "ocaml"
+  depends_on "homebrew/science/z3"
```
DomT4: Can't have mandatory cross-tap dependencies in the core. You'll either need to make this optional, vendor it into `libexec/"vendor"` or such, or submit this to the Science tap.

Library/Formula/fstar.rb
```diff
@@ -0,0 +1,69 @@
...
+      ocaml_install_package
+    end
+
+    system "make", "-C", "src/"
+    system "make", "-C", "src/ocaml-output/"
+
+    bin.install "src/ocaml-output/fstar.exe"
+    prefix.install "README.md"
+
+    (prefix/"stdlib").install Dir["lib/*"]
```
DomT4: For the sake of keeping the `prefix` relatively clean and easy to glance through, we'd prefer to dump all of this into `libexec` and then symlink back what is necessary. You don't need to specify installing the `LICENSE` into the prefix, that should happen automatically per our detection code.

Library/Formula/fstar.rb
```diff
@@ -0,0 +1,69 @@
...
+    F* code can be extracted to OCaml code.
+    To compile generated OCaml code, you have to install the
+    package 'batteries' from the Opam package manager:
+    - brew install opam
+    - opam install batteries
+    EOS
+  end
+
+  test do
+    system "#{bin}/fstar.exe", "--admit_fsi", "FStar.Set", "#{prefix}/stdlib/set.fsi", "#{prefix}/stdlib/heap.fst", "#{prefix}/stdlib/st.fst", "#{prefix}/stdlib/all.fst", "#{prefix}/stdlib/list.fst", "#{prefix}/stdlib/string.fst", "#{prefix}/stdlib/int32.fst", "#{prefix}/stdlib/unit1.fst", "#{prefix}/stdlib/unit2.fst", "#{prefix}/stdlib/testset.fst"
```
DomT4: Please break this onto multiple lines to keep the formula readable.

beurdouche:
@dunn @DomT4, I am having trouble with OCaml resources here and I am not the only one (#44173)
Is there a way to have Opam packages automatically fetched as dependencies ?

bfontaine:
@beurdouche we need to be explicit about the resources we need; so we can cache them and verify them using checksums.

beurdouche:
@bfontaine, understandable. I just found out that #42405 pulled OCaml resources and was merged so I suppose that this strategy was ok with you ?

bfontaine:
@beurdouche as long as you use `resource do` blocks to download external resources it’s fine.

beurdouche:
@bfontaine, I fixed the formula that now passes the Travis build for 10.10 and 10.09. 
It fails for 10.11 because the aspcud bottle can't be fetched while it is registed in the formula : 
https://github.com/Homebrew/homebrew/commit/fae4fcf703f457c45c2b1a4cdd038bc9319a74f7

Library/Formula/fstar.rb
```diff
@@ -0,0 +1,94 @@
...
+  url "https://github.com/FStarLang/FStar.git",
+    :tag => "v0.9.0",
+    :revision => "f24457213f4007302873f6d42eb2cf8094fa45d0"
+  head "https://github.com/FStarLang/FStar.git"
+
+  depends_on "mono"   => :build
+  depends_on "fsharp" => :build
+  depends_on "opam"   => :build
+  depends_on "ocaml"  => :recommended
+  depends_on "homebrew/science/z3"
```
bfontaine: Core formulae can’t have hard dependencies in taps. We need to have this optional or move this formula in the `homebrew/science` tap.

beurdouche:
@bfontaine, sure I understand ! Should I open an issue about the non-existent `aspcud` bottle for el_capitan ? As soon as the bottle problem is fixed, I will move the formula to `homebrew/science`.
Any other feedback on the formula will be highly appreciated =)
Thanks for the help everyone !

DomT4:
> Should I open an issue about the non-existent aspcud bottle for el_capitan ? 

Removed the dead bottle.

> As soon as the bottle problem is fixed, I will move the formula to homebrew/science.

Closing per this. Thank you for the submission though; we appreciate that & look forwards to seeing this in Science :smiley_cat:.

PR #44343, opened 2015-09-25
utils: re-implement and improve column-wise output

UniqMartin:
I'd like to work on `puts_columns` as I believe decoupling this from an external tool will open up new possibilities without significant performance loss or increase in complexity.

***Please continue reading in the [3rd comment from 2015-10-03](#issuecomment-145188663). The remainder of this initial commit does no longer apply, but is retained in case someone already read it.***

----

The implementation currently doesn't try to faithfully recreate the previous output. In fact, these are some (major) differences:

- Items ordered line by line from left to right (was column by column from top to bottom).
- No pagination (was 55 lines per “page” irrespective of terminal height).
- Full justification even for single-line output (was flush left).
- Reduced right margin (was always at least two spaces).

----

Pagination and item order are probably worth some discussion. Here are some arguments:

### Current: Top to Bottom, With Pagination

- Columns are easier to follow visually.
- Scrolling up and down is inconvenient (for columns taller than terminal height).
- Pagination is awkward as there is no indication when a new “page” starts (only affects long listings with more than 55 lines).
- Disagrees with left-to-right reading direction in English.
- Slightly harder to implement.

Example layout on a hypothetical 9-column terminal (`|` = boundaries, `.` = blanks):

```
|a..d..g..|
|b..e.....|
|c..f.....|
```

### Proposed: Left to Right, Without Pagination

- Lines are slightly harder to follow visually.
- No pagination, no back-and-forth scrolling while scanning the listing from front to back.
- Agrees with left-to-right reading direction in English.
- Easier to implement.

Example layout on a hypothetical 9-column terminal (`|` = boundaries, `.` = blanks):

```
|a...b...c|
|d...e...f|
|g........|
```

----

Looking forward to your feedback!

*Some context: My motivation for re-implementing this in pure Ruby is that I'd like to be able to highlight individual items of the listing using `Tty`'s escape codes. This isn't possible with the current approach using an external command, because it does not have support for escape codes, effectively destroying the column layout as soon as those are used. Since highlighting is an entirely new feature, I'd like to postpone the implementation until this PR gets merged. Ultimately, #44335 will benefit from this highlighting.*

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -260,17 +260,34 @@ def puts_columns(items, star_items = [])
     items = items.map { |item| star_items.include?(item) ? "#{item}*" : item }
   end
 
-  if $stdout.tty?
-    # determine the best width to display for different console sizes
-    console_width = `/bin/stty size`.chomp.split(" ").last.to_i
-    console_width = 80 if console_width <= 0
-    max_len = items.reduce(0) { |max, item| l = item.length ; l > max ? l : max }
-    optimal_col_width = (console_width.to_f / (max_len + 2).to_f).floor
-    cols = optimal_col_width > 1 ? optimal_col_width : 1
-
-    IO.popen("/usr/bin/pr -#{cols} -t -w#{console_width}", "w") { |io| io.puts(items) }
-  else
+  unless $stdout.tty?
     puts items
+    return
+  end
+
+  # TTY case: If possible, output using multiple columns.
+  console_width = Tty.width
+  console_width = 80 if console_width <= 0
+  max_len = items.max_by(&:length).length
+  col_gap = 2 # number of spaces between columns
+  gap_str = " " * col_gap
+  cols = (console_width + col_gap) / (max_len + col_gap)
+  cols = 1 if cols < 1
+  rows = (items.size + cols - 1) / cols
+  cols = (items.size + rows - 1) / rows # avoid empty trailing columns
+
+  if cols >= 2
+    col_width = (console_width + col_gap) / cols - col_gap
+    items = items.map { |item| item.ljust(col_width) }
+  end
+
+  if cols == 1
+    puts items
+  else
+    rows.times do |row_index|
+      item_indices_for_row = row_index.step(items.size - 1, rows).to_a
+      puts items.values_at(*item_indices_for_row).join(gap_str)
+    end
   end
 end
 
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -42,6 +42,10 @@ class Tty
       bold 30
     end
 
+    def highlight
+      bold 43
+    end
+
     def width
       `/usr/bin/tput cols`.strip.to_i
     end
@@ -253,12 +257,8 @@ def curl(*args)
   safe_system curl, *args
 end
 
-def puts_columns(items, star_items = [])
+def puts_columns(items, highlight = [])
   return if items.empty?
-
-  if star_items && star_items.any?
-    items = items.map { |item| star_items.include?(item) ? "#{item}*" : item }
-  end
 
   unless $stdout.tty?
     puts items
@@ -276,9 +276,18 @@ def puts_columns(items, star_items = [])
   rows = (items.size + cols - 1) / cols
   cols = (items.size + rows - 1) / rows # avoid empty trailing columns
 
+  plain_item_lengths = items.map(&:length) if cols >= 2
+  if highlight && highlight.any?
+    items = items.map do |item|
+      highlight.include?(item) ? "#{Tty.highlight}#{item}#{Tty.reset}" : item
+    end
+  end
+
   if cols >= 2
     col_width = (console_width + col_gap) / cols - col_gap
-    items = items.map { |item| item.ljust(col_width) }
+    items = items.each_with_index.map do |item, index|
+      item + "".ljust(col_width - plain_item_lengths[index])
+    end
   end
 
   if cols == 1


UniqMartin:
Thinking a bit more about it, I could imagine that the pagination done by `pr` wasn't actually intended and went unnoticed. If top to bottom turns out to be the preferred layout, I'd suggest to implement it without pagination, resembling the output of `ls`. (Though this will still require scrolling back and forth to scan the entire output of a long list in its original order.)

UniqMartin:
There are two commits in this PR:

## 1: Reimplementation of Multi-Column Output

(Please review this first, as this is the primary topic of this PR. Ideally, I'd like to have a consensus here before tweaking the second part.)

The current implementation doesn't fully recreate the original output. This is intentional. Differences are:

- No pagination (was 55 lines per “page” irrespective of terminal height; default of `pr` command).
- Full justification even for single-line output (was flush left).
- Reduced right margin (was always at least two spaces).
- If environment variable`HOMEBREW_PUTS_COLUMNS_ROW_MAJOR` is set, put items in row-major order (new alternative). If unset, use column-major order (as before).
  - *Note:* I implemented both variants for testing purposes. I will remove one of them once there's a consensus on the preferred item arrangement.

## 2: Highlighting of Individual Items

(This part could also be moved into a separate PR once the first part gets accepted.)

Provided to make the code structure of the first part clearer; plus it conveys the bigger picture. The ultimate goal is to allow for highlighting of individual items in the multi-column listing. @xu-cheng suggested to implement this highlighting feature in a [diff comment](https://github.com/Homebrew/homebrew/pull/44335#discussion_r40449204) in #44335. I imagine @dunn might also be interested, according to comments [here](https://github.com/Homebrew/homebrew/pull/42662#issuecomment-129078548) and [here](https://github.com/Homebrew/homebrew/pull/44335#issuecomment-143476101). I wouldn't be surprised if there are even more uses for this feature.

I changed the interpretation of the second parameter of `puts_columns` for this. With this change it shouldn't be used to convey important information, as it will be only rendered on a TTY. However, this should be no problem, as there's currently no code in Homebrew core using this second parameter.

UniqMartin:
Here's a screenshot that shows how highlighting could be used to enrich the output of `brew update`:

Highlighting of installed formulae in formula update report.

Library/Homebrew/utils.rb
```diff
@@ -253,24 +257,54 @@ def curl(*args)
...
-    IO.popen("/usr/bin/pr -#{cols} -t -w#{console_width}", "w") { |io| io.puts(items) }
+  # TTY case: If possible, output using multiple columns.
+  console_width = Tty.width
+  console_width = 80 if console_width <= 0
+  max_len = items.max_by(&:length).length
+  col_gap = 2 # number of spaces between columns
+  gap_str = " " * col_gap
+  cols = (console_width + col_gap) / (max_len + col_gap)
+  cols = 1 if cols < 1
+  if ENV["HOMEBREW_PUTS_COLUMNS_ROW_MAJOR"]
```
xu-cheng: What's this for?

xu-cheng:
I wonder do you have any benchmark to show new implementation's performance vs old one.

cc @Homebrew/owners for advice

Library/Homebrew/utils.rb
UniqMartin: Are you referring to `ENV["HOMEBREW_PUTS_COLUMNS_ROW_MAJOR"]`? If yes: It was easy enough to implement both [row-major and column-major order](https://en.wikipedia.org/wiki/Row-major_order) in the output, so I figured I make it easy for people to test both variants. To quote from [my previous comment](#issuecomment-145188663):

> - If environment variable `HOMEBREW_PUTS_COLUMNS_ROW_MAJOR` is set, put items in row-major order (new alternative). If unset, use column-major order (as before).
>   - *Note:* I implemented both variants for testing purposes. I will remove one of them once there's a consensus on the preferred item arrangement.

UniqMartin:
> I wonder do you have any benchmark to show new implementation's performance vs old one.

I don't. It shouldn't be too difficult to benchmark this, though. I'll see what I can do and will report back.

However, my gut feeling is that it doesn't really matter, because almost always, the cost to generate the data to be output via `puts_columns` will be significantly higher (`brew update` probably being the most obvious example), making the performance of `puts_columns` itself negligible.

UniqMartin:
> It shouldn't be too difficult to benchmark this, though. I'll see what I can do and will report back.

Here are benchmark results using `Benchmark.measure`. The external command that generated the results, the JSON test data, and outputs (on a terminal with 120 columns) can be found [in a gist](https://gist.github.com/UniqMartin/859f635c2d9513442958).

The two test files `data-{3,4}.json` are more or less what would be reported by `brew update` if the last update was 1000/10000 commits ago. (Note for comparison with the times below, that it took 22/136 seconds to generate these files.)

I ran every command three times and picked the best result. The output is user/system/total CPU time and (in parentheses) elapsed real time.

Old (non-native) code:

```
$ brew benchmark-multicol dump-X.json # on branch master
  0.010000   0.010000   0.060000 (  0.055237) - dump-3, non-native
  0.010000   0.010000   0.070000 (  0.073264) - dump-4, non-native
```

New (native) code, column-major order:

```
$ brew benchmark-multicol --highlight dump-X.json # on PR branch
  0.000000   0.000000   0.020000 (  0.020293) - dump-3, column-major, highlighted
  0.020000   0.010000   0.050000 (  0.052954) - dump-4, column-major, highlighted
```

New (native) code, row-major order:

```
$ brew benchmark-multicol --rows --highlight dump-X.json # on PR branch
  0.000000   0.000000   0.020000 (  0.019649) - dump-3, row-major, highlighted
  0.020000   0.010000   0.050000 (  0.052012) - dump-4, row-major, highlighted
```

mikemcquaid:
This is sounding good to me so far. Can you suggest some commands I can run to view the difference in output? @xu-cheng any more thoughts? Thanks!

xu-cheng:
This PR should be merged alongside with #44335.

LGTM, except we should decide what layout we should use.

UniqMartin:
> Can you suggest some commands I can run to view the difference in output?

Try (prefix any of them with `HOMEBREW_PUTS_COLUMNS_ROW_MAJOR=1` for the row-major layout):

- `brew commands`,
- `brew list --full-name` (the only variant that uses `puts_columns`),
- `brew search`, `brew search pi` (or similar),
- `brew update` (probably easier to test with `brew benchmark-multicol` [from above](#issuecomment-145880361)), and
- `brew uses libpng` (or similar).

You can also use `brew irb` and then invoke `puts_columns` with pretty much any array of strings you can come up with.

> LGTM, except we should decide what layout we should use.

Indeed! For what it's worth, I'm now pretty much in favor of the (default) column-major layout.

mikemcquaid:
Implementation and local testing look good. I also prefer the default column-major layout. Can you kill the `HOMEBREW_PUTS_COLUMNS_ROW_MAJOR` stuff and I'll merge? Thanks!

UniqMartin:
Here we go. Just ripped out the `HOMEBREW_PUTS_COLUMNS_ROW_MAJOR` stuff, leaving column-major layout as the only option.

mikemcquaid:
Thanks @UniqMartin :heart_eyes:

PR #44354, opened 2015-09-26
gnuradio: fix --with-documentation

metacollin:
Fixed ruby error preventing formula from building when --with-documentation was used, and added a patch to fix wx not being found on systems with multiple versions installed.

### DIFF
--- a/Library/Formula/gnuradio.rb
+++ b/Library/Formula/gnuradio.rb
@@ -19,6 +19,14 @@ class Gnuradio < Formula
     sha256 "9e1c612f0f4063d387d85517cc420f050f49c7903d36fab45b72e8d828549e3c"
   end
 
+  # The wxpython dependency isn't properly found on multiversion 
+  # python systems.  This patches gr-wxgui's CMakeLists.txt file 
+  # to correctly import wxversion and select the brewed wxpython package.
+  patch do
+    url "https://gist.githubusercontent.com/metacollin/d6c7854a01e0e9094a60/raw/7569850f4a87d67bbb31d83a4bc20445cf0df61d/wx.patch"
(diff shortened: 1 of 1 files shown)


Library/Formula/gnuradio.rb
```diff
@@ -19,6 +19,14 @@ class Gnuradio < Formula
     sha256 "9e1c612f0f4063d387d85517cc420f050f49c7903d36fab45b72e8d828549e3c"
   end
 
+  # The wxpython dependency isn't properly found on multiversion 
+  # python systems.  This patches gr-wxgui's CMakeLists.txt file 
+  # to correctly import wxversion and select the brewed wxpython package.
```
dunn: Can you submit this patch upstream? Thanks!

Library/Formula/gnuradio.rb
```diff
@@ -138,7 +147,7 @@ def install
     ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
 
     res = %w[Markdown Cheetah lxml numpy]
-    res << %w[sphinx sphinx_rtd_theme alabaster babel docutils pygments
+    res += %w[sphinx sphinx_rtd_theme alabaster babel docutils pygments
```
bfontaine: Please leave `<<` here, it modifies `res` in-place while `+=` creates a new array then assign it to `res`.

jiegec: Nope. That means appending the whole list as a item to that list. My fault.

bfontaine: Right.

bfontaine:
Ping re Alex’ comment?

DomT4:
This should be resolved now. Please `brew update` and let us know if you run into any issues, Thanks!

PR #44360, opened 2015-09-26
v8: Fixed SHA256 for --build-from-source

dongcarl:
running `brew install v8 --build-from-source` used to result in a SHA256 mismatch, now it doesn't.

### DIFF
--- a/Library/Formula/v8.rb
+++ b/Library/Formula/v8.rb
@@ -4,7 +4,7 @@ class V8 < Formula
   desc "Google's JavaScript engine"
   homepage "https://code.google.com/p/v8/"
   url "https://github.com/v8/v8-git-mirror/archive/4.5.103.35.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Can we grab someone upstream to verify this checksum change is legit? Usually means they retagged it or pushed something to the branch, but we like to check for obvious reasons.

dongcarl:
@DomT4 agreed. the hash is for https://github.com/v8/v8-git-mirror/archive/4.5.103.35.tar.gz btw

DomT4:
Has there been any contact/confirmation with upstream on the reason for the checksum change here?

DomT4:
Oh whoops. Merged my own version of this after verifying the SHA1s of the commit references on the Github mirror and on the upstream official Git. Forgot to close this evidently. Thanks for the submission! :heart:

PR #44361, opened 2015-09-26
Add formula for StackExchange/blackbox v1.20150730

booi:


booi:
Hi all, this is to add a formula for https://github.com/StackExchange/blackbox, a set of scripts that uses GnuPG to encrypt secrets for safe storage within Git (or any other SCMs). 

Unfortunately, I couldn't figure out a simple way to test this as it looks at the user's ~/.gnupg directory for keys.

booi:
Closing this request, some new files were not added.

PR #44367, opened 2015-09-26
gdk-pixbuf 2.32.1

tschoonj:
see discussion in #44231 and #44262 

relocations patch was accepted upstream.
incorporated @staticfloat fixes in post_install and caveat blocks


staticfloat:
Thanks for seeing this through @tschoonj; all my concerns are satisfied with this latest updated.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/16f16d323fde80ae31f3c15e11a73f8d90211163. Thanks again @tschoonj!

PR #44370, opened 2015-09-26
weechat: added '-DCMAKE_BUILD_TYPE=Debug' argument

Drewshg312:
added debug option because of this issue: https://github.com/weechat/weechat/issues/527
seems like this formulae crashes pretty often so default debug option would be beneficial…

### DIFF
--- a/Library/Formula/weechat.rb
+++ b/Library/Formula/weechat.rb
@@ -16,6 +16,7 @@ class Weechat < Formula
   option "with-perl", "Build the perl module"
   option "with-ruby", "Build the ruby module"
   option "with-curl", "Build with brewed curl"
+  option "with-debug", "Build with debug information"
 
   depends_on "cmake" => :build
   depends_on "gnutls"
@@ -29,6 +30,10 @@ class Weechat < Formula
 
   def install
     args = std_cmake_args
+    if build.with? "debug"
+      args -= %w[-DCMAKE_BUILD_TYPE=Release]
+      args << "-DCMAKE_BUILD_TYPE=Debug"
+    end
 
     args << "-DENABLE_LUA=OFF" if build.without? "lua"
     args << "-DENABLE_PERL=OFF" if build.without? "perl"


DomT4:
We're not going to make the debug build the default. Most people don't want it or need it. Will consider an *option* to enable the debug build, but you'd need to also override the `Release` argument from `std_cmake_args` in-formula.

Drewshg312:
Thank you! But could you be more detailed about how to "override the `Release` argument from `std_cmake_args` in-formula"? I don't really see any  `Release` arguments there… But maybe you are right and that's why my `lldb --core` trace is not informative… this will help to fix crashes of weechat formula on osx (because of guil plugin bug)... Please help. Thanx

DomT4:
> But could you be more detailed about how to "override the Release argument from std_cmake_args in-formula"?

Something like:

```ruby
option "with-debug", "Build with debug information"

def install
  if build.with? "debug"
    args = std_cmake_args - %w[-DCMAKE_BUILD_TYPE=Release]
    args << "-DCMAKE_BUILD_TYPE=Debug"
  else
    args = std_cmake_args
  end
```

Would likely work. The `std_cmake_args` are set from [here](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/formula.rb#L925-L935), which is why you don't see them in-formula.

apjanke:
This looks good now. Could you squash this to a single commit so we don't have the unwanted code from the initial attempt in our commit history after merging?

Drewshg312:
Done... sort of...

Library/Formula/weechat.rb
```diff
@@ -28,7 +29,12 @@ class Weechat < Formula
   depends_on "curl" => :optional
 
   def install
-    args = std_cmake_args
```
mikemcquaid: Could leave this as is, get rid of the else below and just use `-=` to remove it.

apjanke: Ping... want to make this change, and we can get to merging it?

Drewshg312: yes, I just have hard time using git. Need some help squashing stuff... trying to do it right now...

apjanke: `git rebase -i master` should do the trick for squashing commits. Do "pick" for the first commit and "fixup" for subsequent ones. Or if you have uncommitted changes, just `git commit --amend` will let you include them in the prior commit.

Drewshg312:
Should be good… please tell me if it's still not the way it supposed to be. Thanks.

apjanke:
Looks pretty good. As a style issue, I would prefer

```
args -= %w[-DCMAKE_BUILD_TYPE=Release]
```

instead of

```
args = std_cmake_args - %w[-DCMAKE_BUILD_TYPE=Release]
```

here. It has the same effect in this particular case, but if a formula has multiple options, each should be incrementally changing `args` instead of "resetting" it against `std_cmake_args`. Doing the `args -= ...` avoids possible logic bugs when inserting or reordering code in the formula. (For example, if the lua and perl binding stuff got moved up before the debug stuff, it would break them.) Looks a little tidier, too.

bfontaine:

bfontaine:
Thank you for your contribution to Homebrew, @Drewshg312; we appreciate it! :tada:

PR #44373, opened 2015-09-27
zurl 1.4.10

jkarneges:


jkarneges:
```
clang++ -c -pipe -stdlib=libc++ -mmacosx-version-min=10.7 -O2 -arch x86_64 -fPIC -Wall -W -DJDNS_STATIC -DNO_IRISNET -DUSE_CURL -DQT_NO_DEBUG -DQT_NETWORK_LIB -DQT_CORE_LIB -DQT_SHARED -I/usr/local/Cellar/qt/4.8.7/mkspecs/unsupported/macx-clang-libc++ -I. -I/usr/local/Cellar/qt/4.8.7/lib/QtCore.framework/Versions/4/Headers -I/usr/local/Cellar/qt/4.8.7/lib/QtCore.framework/Versions/4/Headers -I/usr/local/Cellar/qt/4.8.7/lib/QtNetwork.framework/Versions/4/Headers -I/usr/local/Cellar/qt/4.8.7/lib/QtNetwork.framework/Versions/4/Headers -I/usr/local/Cellar/qt/4.8.7/include -I/usr/local/Cellar/zeromq/4.1.3/include -I/usr/local/Cellar/qt/4.8.7/include -I/usr/local/Cellar/qjson/0.8.1/include -I../../jdns/include/jdns -I../../qzmq/src -I../../common -I../../../src -I_moc -F/usr/local/Cellar/qt/4.8.7/lib -o _obj/websocket.o ../../websocket.cpp
./../websocket.cpp:324:2: error: unknown type name 'QSslSocket'
        QSslSocket *sock;
        ^
../../websocket.cpp:1033:34: error: use of undeclared identifier 'QSslError'
        void sock_sslErrors(const QList<QSslError> &errors)
```

The CI system picked up an impossible error on 10.11. It seems there was no trouble including the QSslSocket header (which maps to qsslsocket.h), yet the symbols inside aren't recognized. If you look in qsslsocket.h you'll see the definitions are protected by an `#ifndef QT_NO_OPENSSL` conditional, but this define is not present in the above compiler arguments.

Also is El Capitan compatibility a requirement for a merge?

DomT4:
> Also is El Capitan compatibility a requirement for a merge?

Nope. Just add `depends_on MaximumMacOSRequirement => :yosemite` to the formula to block El Cap users being able to install it.

jkarneges:
Well before we block it out I figured I'd see if I could fix openssl support on El Capitan. See the separate Qt patch.

However, it occurs to me that since zurl also depends on libcurl, it is possible that it'll end up linking to two different copies of openssl. The system one through libcurl, and the homebrew one through qt. It might be that zurl should depend on the homebrew curl package (which could in turn be made to link against the homebrew openssl package) on El Capitan, to ensure that only one openssl library is used at runtime.

DomT4:
The system `curl` hasn't had OpenSSL linkage since Mountain Lion. Apple builds it against SecureTransport.

jkarneges:
Rebased to trigger a build. Looks like the jenkins queue has been stuck for some hours though.

jkarneges:
there we go :)

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/651715b0e56f495588c875909b52f7735aaf1498. Thank you for another contribution to Homebrew @jkarneges! :smiley_cat:

PR #44376, opened 2015-09-27
ENV#userpaths!: ensure Superenv.bin and opt path come at first

xu-cheng:
This ensures that binaries in /usr/local/bin would not overwrite
superenv.

### DIFF
--- a/Library/Homebrew/extend/ENV/shared.rb
+++ b/Library/Homebrew/extend/ENV/shared.rb
@@ -190,10 +190,16 @@ module SharedEnvExtension
 
   # @private
   def userpaths!
-    paths = ORIGINAL_PATHS.map { |p| p.realpath.to_s rescue nil } - %w[/usr/X11/bin /opt/X11/bin]
-    self["PATH"] = paths.unshift(*self["PATH"].split(File::PATH_SEPARATOR)).uniq.join(File::PATH_SEPARATOR)
+    paths = self["PATH"].split(File::PATH_SEPARATOR)
+    # put Superenv.bin and opt path at the first
+    new_paths = paths.select { |p| p.start_with?("#{HOMEBREW_REPOSITORY}/Library/ENV") || p.start_with?("#{HOMEBREW_PREFIX}/opt") }
     # XXX hot fix to prefer brewed stuff (e.g. python) over /usr/bin.
-    prepend_path "PATH", HOMEBREW_PREFIX/"bin"
+    new_paths << "#{HOMEBREW_PREFIX}/bin"
+    # reset of self["PATH"]
+    new_paths += paths
+    # user paths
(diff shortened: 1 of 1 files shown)


tdsmith:
It looks like, assuming this runs after superenv's path manipulations, this changes the behavior of userpaths by adding opt_bin paths calculated from depends_ons to PATH ahead of any user paths; previously, the opt_bin paths were filtered out of PATH entirely.

I don't have an opinion about which is more correct; I'm wondering if there's a scenario we want to avoid where we inherit a dependency on some Homebrew package foo from the recursive dependencies of the package we're trying to build, when we really wanted to build against our user's foo instead.

I'm thinking we probably don't care.

:+1: on this PR if that sounds right to you.

tdsmith:
No, I'm wrong, as usual; I had no idea what `unshift` did.

Library/Homebrew/extend/ENV/shared.rb
```diff
@@ -190,10 +190,16 @@ def ncurses_define
...
   # @private
   def userpaths!
-    paths = ORIGINAL_PATHS.map { |p| p.realpath.to_s rescue nil } - %w[/usr/X11/bin /opt/X11/bin]
-    self["PATH"] = paths.unshift(*self["PATH"].split(File::PATH_SEPARATOR)).uniq.join(File::PATH_SEPARATOR)
+    paths = self["PATH"].split(File::PATH_SEPARATOR)
+    # put Superenv.bin and opt path at the first
+    new_paths = paths.select { |p| p.start_with?("#{HOMEBREW_REPOSITORY}/Library/ENV") || p.start_with?("#{HOMEBREW_PREFIX}/opt") }
     # XXX hot fix to prefer brewed stuff (e.g. python) over /usr/bin.
-    prepend_path "PATH", HOMEBREW_PREFIX/"bin"
+    new_paths << "#{HOMEBREW_PREFIX}/bin"
```
tdsmith: Is this still necessary? If the Homebrew dependencies we collected are already at the front of `PATH`, we should respect the ordering of `ORIGINAL_PATHS` for anything else.

xu-cheng:
> I had no idea what unshift did.

Me neither. Every time,  I need to fire an interactive shell to double check its function.

Library/Homebrew/extend/ENV/shared.rb
xu-cheng: I'm not sure. But it seems to be no harm to add this anyway.

xu-cheng: Rethink a bit. I think this is required. For example, `depends_on :python` will collect `/usr/local/bin/` instead of `/usr/local/opt/python/bin` while here I only ensure opt path coming at first. So by adding `/usr/local/bin`, the brew python will be chosen over system python.

PR #44379, opened 2015-09-27
More Travis Fixes

xu-cheng:
See commit messages for details.

cc @mikemcquaid

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,14 +16,14 @@ matrix:
       rvm: system
 
 before_install:
+  - export TRAVIS_COMMIT=$(git rev-parse --verify -q HEAD)
+  - cd /usr/local
   - sudo rm -rf /usr/local/.git/refs /usr/local/.git/packed-refs
   - sudo rsync -az "$TRAVIS_BUILD_DIR/" /usr/local/
   - export TRAVIS_BUILD_DIR="/usr/local"
   - sudo chown -R $USER /usr/local
-  - cd /usr/local
   - env | grep TRAVIS_
   - if [ -f ".git/shallow" ]; then travis_retry git fetch --unshallow; fi
-  - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then git reset --hard $TRAVIS_COMMIT || travis_retry git fetch origin +refs/pull/$TRAVIS_PULL_REQUEST/merge:; fi
   - git reset --hard $TRAVIS_COMMIT
   - git clean -qxdff
 
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -318,6 +318,7 @@ module Homebrew
       # Use Travis CI Git variables for master or branch jobs.
       elsif ENV["TRAVIS_COMMIT_RANGE"]
         diff_start_sha1, diff_end_sha1 = ENV["TRAVIS_COMMIT_RANGE"].split "..."
+        diff_start_sha1 = git("merge-base", diff_start_sha1, diff_end_sha1).strip
       # Otherwise just use the current SHA-1 (which may be overriden later)
       else
         diff_end_sha1 = diff_start_sha1 = current_sha1


Library/Homebrew/cmd/test-bot.rb
```diff
@@ -280,7 +280,7 @@ def single_commit?(start_revision, end_revision)
       def diff_formulae(start_revision, end_revision, path, filter)
         git(
           "diff-tree", "-r", "--name-only", "--diff-filter=#{filter}",
-          start_revision, end_revision, "--", path
+          "#{start_revision}..#{end_revision}", "--", path
```
xu-cheng: I don't know why but experiment shows this notation only working for `git log` instead of `git diff-tree`.

xu-cheng: Should fix now by using `git merge-base`.

tdsmith:
Great catches. In the event that a PR is force-pushed and updated from commit `a` to commit `b` between the launch of a build and the launch of a worker, does Travis post the build status to `a` or `b`? If we're seeing this, I guess Travis sets TRAVIS_COMMIT to `a`, checks out `b`, and then pushes the CI status to commit `b`?

xu-cheng:
>  I guess Travis sets TRAVIS_COMMIT to a, checks out b, and then pushes the CI status to commit b?

I think Travis would still to push the status to commit a(not 100% sure).

xu-cheng:
If it helps, this is a real case:
https://github.com/Homebrew/homebrew/pull/44335
https://travis-ci.org/Homebrew/homebrew/jobs/82179567

xu-cheng:
I think the real problem is not PR's commit changed, but the merge commit is the one changed. In the case above, the CI status is pushed to the commit 9e6ec84. While travis failed to checkout 71b1496b861288d82f023b0edc21e69e68351461, because the merge commit has been changed to 23fe87dbbe5fa6ef8c434bba89ae61a848e78377

.travis.yml
```diff
@@ -16,14 +16,14 @@ matrix:
       rvm: system
 
 before_install:
+  - export TRAVIS_COMMIT=$(git rev-parse --verify -q HEAD)
```
mikemcquaid: Why are you overriding the existing value?

.travis.yml
```diff
@@ -16,14 +16,14 @@ matrix:
...
+  - cd /usr/local
   - sudo rm -rf /usr/local/.git/refs /usr/local/.git/packed-refs
   - sudo rsync -az "$TRAVIS_BUILD_DIR/" /usr/local/
   - export TRAVIS_BUILD_DIR="/usr/local"
   - sudo chown -R $USER /usr/local
-  - cd /usr/local
```
mikemcquaid: Why move this? If there's no real reason: that's fine. Just checking I haven't missed something.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -318,6 +318,7 @@ def brew_update
       # Use Travis CI Git variables for master or branch jobs.
       elsif ENV["TRAVIS_COMMIT_RANGE"]
         diff_start_sha1, diff_end_sha1 = ENV["TRAVIS_COMMIT_RANGE"].split "..."
+        diff_start_sha1 = git("merge-base", diff_start_sha1, diff_end_sha1).strip
```
mikemcquaid: :cool:. As discussed would be great to add this to Jenkins too.

mikemcquaid: Or perhaps it'll work unconditionally?

xu-cheng: Yes, I agree that we should make it unconditionally

.travis.yml
xu-cheng: No real reason. :smile:

.travis.yml
xu-cheng: Whenever we push new commits into master, Github will auto update `+refs/pull/$TRAVIS_PULL_REQUEST/merge` to make PR merge into latest master commit. As result, Travis will download the new merge commit but fail to update `TRAVIS_COMMIT`. This is why we see `Cannot parse object TRAVIS_COMMIT` error, and it won't be solved by refetching.

So I override the value to point it to the `FETCH_HEAD` which is also the commit we want to test.

mikemcquaid: Cool, nice work. Maybe worth filing as a bug report to Travis somehow?

xu-cheng: Travis's issue is in https://github.com/travis-ci/travis-ci/issues/2666

PR #44385, opened 2015-09-27
mobile-shell: use official repository URL

rnh:
https://mosh.mit.edu/ refers to https://github.com/mobile-shell/mosh as the official repository, and while https://github.com/keithw/mosh redirects to https://github.com/mobile-shell/mosh using it here can cause some confusion.


tdsmith:

PR #44387, opened 2015-09-27
wine: move warning to stable only

LinusU:
The bug the warning referred to has since been fixed and is only still present in the stable version.


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/a1b23d16a299ca1534629e0c2f96e8dc024e293c. Thank you @LinusU!

PR #44388, opened 2015-09-27
qt: use homebrew openssl on el capitan

jkarneges:
Currently, Qt is compiled without openssl support on El Capitan since the headers are missing from the Mac SDK.

### DIFF
--- a/Library/Formula/qt.rb
+++ b/Library/Formula/qt.rb
@@ -31,6 +31,7 @@ class Qt < Formula
   option "with-docs", "Build documentation"
   option "with-developer", "Build and link with developer options"
 
+  depends_on "openssl"
   depends_on "d-bus" => :optional
   depends_on "mysql" => :optional
   depends_on "postgresql" => :optional
@@ -57,6 +58,10 @@ class Qt < Formula
         args << "unsupported/macx-clang"
       end
     end
+
+    args << "-openssl-linked"
+    args << "-I" << Formula["openssl"].opt_include
+    args << "-L" << Formula["openssl"].opt_lib
 
     args << "-plugin-sql-mysql" if build.with? "mysql"
     args << "-plugin-sql-psql" if build.with? "postgresql"


tdsmith:
If this works, we should do it on all platforms.

Library/Formula/qt.rb
```diff
@@ -58,6 +59,12 @@ def install
       end
     end
 
+    if MacOS.version >= :el_capitan
+      args << "-openssl-linked"
+      args << "-I" << "/usr/local/opt/openssl/include"
```
tdsmith: `Formula["openssl"].opt_include` -- we don't know that we're installed to /usr/local.

Library/Formula/qt.rb
```diff
@@ -58,6 +59,12 @@ def install
...
+      args << "-L" << "/usr/local/opt/openssl/lib"
```
tdsmith: similarly, `Formula["openssl"].opt_lib`

jkarneges:
Thanks for the tips. Patch updated.

jkarneges:
"The job exceeded the maxmimum time limit for jobs, and has been terminated." Heh. :)

(also typo in that error message)

tdsmith:
Eh, that's Travis's message.

tdsmith:
Travis is doing something wrong here; Jenkins should give good feedback but it'll take a while.

jkarneges:
Ah, I didn't realize you run it through both systems.

My local build of qt finally finished and it seems to work. I was able to build an application against it (zurl) and confirm ssl support works.

tdsmith:
Okay, great! Can you remove the `if el_capitan` logic? We should do this on all platforms.

jkarneges:
Done.

DomT4:
This hasn't worked before, which is why we've had to do things like blocking PhantomJS from building on El Cap, because `qt` loves to prioritise `sysroot` which means it scoops up the system libraries in `/usr/lib` but will try and use our headers in `HOMEBREW_PREFIX/opt/openssl/include` and that conflict has broken compile multiple times.

Perhaps they've fixed it upstream. We'll see I guess.

jkarneges:
One of the important parameters is -openssl-linked which tells qt to explicitly link to openssl. Qt's default behavior is to search for libssl manually at runtime, in which case it would likely find the version in /usr/lib and produce undesired effects.

I confirmed after compilation using `otool -L` that QtNetwork is linked to files in `/usr/local/opt/openssl/lib`.

DomT4:
Yes. That's the same argument we pass to PhantomJS as [well](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/phantomjs.rb#L31) that hasn't taken kindly to an openssl dependency in the past. If it works, cool. Extremely happy for my worries to prove misguided.

Mike used to be very firmly against this on the basis most Qt applications didn't make network calls, but he isn't around to check up with & his views may have changed in the last 9 months.

tdsmith:
This builds fine for me after poisoning my system OpenSSL headers (on 10.10, Xcode 7 + CLT) with `#error` directives so I'm pretty confident the bottles are built correctly.

Just because there's some question about what this does in the presence of sysroot directives (it's not clear to me if this refers to passing -sysroot to clang to reference the OS X SDK headers on Xcode-only systems, or an unrelated Qt build concept), I'll test this tomorrow on a 10.9 Xcode-only VM.

It sounds like Dominyk would like to see a test added to the qt formula that exercises QtNetwork. Building this should be fine: https://gist.github.com/tdsmith/f55e7e69ae174b5b5a03

tdsmith:
-- I don't think the test needs to be added in this PR; let's err on the side of not queueing another build. ;) I'll test this on 10.9 myself and add the test code to the formula when I pull this.

jkarneges:
Cool I see the builds finally finished. El Capitan succeeded, but Mavericks and Yosemite show a failure on codequery. Is this brew running tests on reverse dependencies?

tdsmith:
Yes; the codequery failure is probably because the pyqt bottles were bad, which I fixed earlier tonight.

DomT4:
> It sounds like Dominyk would like to see a test added to the qt formula that exercises QtNetwork. Building this should be fine: https://gist.github.com/tdsmith/f55e7e69ae174b5b5a03

This would be great. If we can get that done & passing, I'm happy enough.

jkarneges:
Anything I can do to help this along? The simple QtNetwork test works for me on 10.11.

tdsmith:
Oops, thanks for the ping. Testing now; should land this today.

DomT4:
Some thoughts:

* If we don't revision this, will people who try to build `zurl` (for example) from source hit build failures?
* Do we care about "infecting" Qt's license with OpenSSL's license by default, which Qt's website warns about?
* The bottles didn't build here apparently (?) so it needs to go back through the CI :cry:.
* Adding a test to verify there is actual SSL functionality would be nice. It doesn't have to be anything major:

```c
#include <QtCore>
#include <QDebug>
#include <QSslSocket>

int main(int argc, char *argv[])
{
    QCoreApplication app(argc, argv);
    qDebug() << "Qt has SSL available:" << QSslSocket::supportsSsl();
    QCoreApplication::quit();
}
```

And then:

```ruby
test do
  currently existing test

  (testpath/"testssl.cpp").write <<-EOS.undent
    above pasted example
  EOS

  (testpath/"testssl.pro").write <<-EOS.undent
    Tim's pro example
  EOS

  system bin/"qmake", "testssl.pro"
  system "make"
  assert_match /Qt has SSL available: true/, shell_output("./testssl")
end
```

That test fails on the current Qt, passes on the new one.

DomT4:
Guess my points are moot. Just noticed the commit tree, heh.

tdsmith:
probably, maybe, they totally did build though?, and the test uses TLS though I didn't test that it fails correctly on old qt

DomT4:
Just confirmed locally that unless people reinstall `qt` manually `zurl` will fail to build from source :/.

> they totally did build though

You managed to pull them, so apparently so :smile_cat:. Downstream build was marked as not started, which threw me off.

> the test uses TLS though 

I was just thinking about keeping something inline for the test rather than having to fetch a resource, which we try to avoid. Not a huge thing.

jkarneges:
Indeed, `zurl` assumes `qt` is built with SSL support and will fail to compile otherwise. That's not really a `qt` issue though. Bumping the `qt` revision so that `zurl` can explicitly reference it would may help, if that's possible.

jkarneges:
Also I wouldn't worry about license infection. For better or for worse this is a very common setup (pretty much all Linux distributions build Qt with OpenSSL, and so did Homebrew prior to El Cap), and so it's something Qt app developers have to be mindful of.

tdsmith:
I prefer that tests pull resources instead of using heredocs if there's more than a few lines of a foreign language involved. Who's we? ;)

I guess it makes sense to revision this; I'll queue a PR.

tdsmith:
(Why didn't this auto-close?)

DomT4:
> Who's we? ;)

The other maintainers who have complained at me including external resources for testing before now :p.

jkarneges:
I just wanted to report that the new qt package works for me on 10.10 and 10.11. In both cases it drags in the homebrew openssl, and I'm able to compile zurl.

How do I refer to the latest qt in a formula? I'm not sure how to include a minimum version with `depends_on`.

DomT4:
Just keep it at `depends_on "qt"`. Should be enough.

tdsmith:
What Dominyk said; a dependency on the latest version is implicit and it's not possible to depend on other versions.

dakcarto:
Hmm, why is this not optional? So, the workaround for a lack of headers is to force all users (regardless of OS version) to have Qt linked directly to Homebrew's OpenSSL? **YIKES!**

Possible other solutions:

* Depend on the `openssl` formula, but only with regards to includes, then add either a qmake rpath, or add it directly to the QNetwork module, leaving the compile option at `-openssl` (see my comment in #34964).
* Download a resource of the headers for the appropriate OpenSSL version shipped with El Capitan.

I'm in favor of the latter, so the `qt` formula can be used a dependencies for other built open source software, e.g. my are of focus is QGIS.

With regards to linking to OpenSSL, even Qt doesn't do this for their binary distros for export limitation reasons, so why would Homebrew? (see [Qt's Secure Sockets Layer (SSL) Classes doc](http://doc.qt.io/qt-4.8/ssl.html)) Obviously, Homebrew links to `openssl` in a bunch of other bottles already, and distributes OpenSSL binaries directly, but are the export restrictions this not still a concern?

Basically, anyone using Homebrew's Qt for dependencies in a bundled app will _now_ be hit with export limitations. [Here's how Entought has to manage that](https://www.enthought.com/licenses/export-restrictions/) (no Homebrew-based) and [how OpenSSL views the current export scenario](http://openssl.org/docs/fips/UserGuide-2.0.pdf) (see Appendix F)).

jkarneges:
It was linked either way. dlopen or "direct" is considered the same from the standpoint of a derived work (talk to someone from Debian about this).

Export laws are an interesting point, but if Mac OS X comes with OpenSSL, then isn't the point a bit moot? Regardless of whether a package links to the system OpenSSL or Homebrew OpenSSL, the user ends up with packages linked to OpenSSL. Or is it possible to have a Mac installation with OpenSSL absent, and we need to cover that case?

tdsmith:
>  So, the workaround for a lack of headers is to force all users (regardless of OS version) to have Qt linked directly to Homebrew's OpenSSL? YIKES!

No, we would have done this earlier for better TLS support if it had worked.

Homebrew prioritizes supporting end-users, not developers redistributing software. Developers redistributing libraries built with Homebrew do that at their own risk (and can edit formulas and build their own versions to satisfy their needs).

DomT4:
> Download a resource of the headers for the appropriate OpenSSL version shipped with El Capitan.

We're not going to get into the business of regularly resourcing OpenSSL headers exclusively for an OpenSSL version that's barely supported by Apple or OpenSSL these days and formally dies in 3 months time. Happy for options to be considered a bit more, but you couldn't pay me enough to want to do this particular dance.

dakcarto:
@jkarneges, @tdsmith, @DomT4 

You all make good points, especially 'formally dies in 3 months time'. Can't argue with that. Since Apple has moved away from OpenSSL, this makes sense.

Regarding the method of linking. Maybe follow the recommend way described in [SSL Classes doc](http://doc.qt.io/qt-4.8/ssl.html)?

```ruby
ENV["OPENSSL_LIBS"] = "-L#{Formula["openssl"].opt_lib} -lssl -lcrypto"
```
instead of
```ruby
args << "-L" << Formula["openssl"].opt_lib
```

I've compiled this and the linking shows up only in `QtNetwork.framework/QtNetwork`, which is appropriate.

DomT4:
> ENV["OPENSSL_LIBS"] = "-L#{Formula["openssl"].opt_lib} -lssl -lcrypto"

If this works across the CI and doesn't cause header detection problems on <10.11 I'd be happy to switch to this method, personally.

jkarneges:
@dakcarto No objection to this. The result should end up being the same.

tdsmith:
Adding a directory to the linker search path doesn't actually link anything, though; these should be equivalent.

PR #44189, opened 2015-09-20
salt 2015.8.0

nijikon:

PR #44395, opened 2015-09-28
saltstack 2015.8.0

ianbrandt:
This is my first attempt at updating a Homebrew formula.  All feedback welcomed.

The latest formula for Swig reportedly includes the fix for M2Crypto compatibility (swig/swig#344).  I'm not clear on whether this means Swig should still be installed as a resource, or should be reverted to a dependency?

### DIFF
--- a/Library/Formula/saltstack.rb
+++ b/Library/Formula/saltstack.rb
@@ -5,10 +5,9 @@ class Saltstack < Formula
   # (URLs starting with https://github.com/saltstack/salt/releases/download)
   # github tag archives will report wrong version number
   # https://github.com/Homebrew/homebrew/issues/43493
-  url "https://github.com/saltstack/salt/releases/download/v2015.5.5/salt-2015.5.5.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
See #44189. As far as I can see, the original error reported upstream is still present when built against newer `swig` versions. It doesn't flag up in the CI testing as it has to be triggered by messing with `PYTHONPATH`s and such, although I am tempted to do something about that.

DomT4:
If you did something like this in the `test do` you'd see the failure occur:

```ruby
test do
  ENV.prepend_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
  system "python", "-c", "import M2Crypto"

  system "#{bin}/salt", "--version"
end
```

ianbrandt:
@DomT4 Thanks for the feedback.  I completely missed #44189.  I didn't realize pull requests are filtered by default when searching issues.

I've reverted to Swig 3.0.4 and added the suggested test case.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/f8890ebe0352f99423296e5e8bb2278b1f435b1f. Thank you for your contribution to Homebrew @ianbrandt!

PR #44400, opened 2015-09-28
ruby-build 20150928

mislav:


bfontaine:
Thank you for your contribution @mislav! :tada:

PR #44401, opened 2015-09-28
nanopb-generator: use https urls

vszakats:


bfontaine:
Thank you again @vszakats! :smile:

PR #44402, opened 2015-09-28
languagetool: use https homepage

vszakats:


bfontaine:
Thank you @vszakats! :smile:

PR #44408, opened 2015-09-28
nccmp 1.7.4.0 (new formula)

rsmz:
Passes audit and `make check` during installation.

### DIFF
--- a/Library/Formula/nccmp.rb
+++ b/Library/Formula/nccmp.rb
@@ -1,0 +1,21 @@
+class Nccmp < Formula
+  desc "nccmp compares 2 NetCDF files bitwise, semantically or with a tolerance."
+  homepage "http://nccmp.sourceforge.net/"
+  url "https://downloads.sourceforge.net/projects/nccmp/files/nccmp-1.7.4.0.tar.gz"
+  sha256 "7760b65e54f6b07ee764fe37a71eaa1710f4c6440c6fa5ee05796912865748fb"
+
+  depends_on 'netcdf'
+
+  def install
+    ENV.deparallelize
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "check"
+    system "make", "install"
+  end
+
+  test do
+    system "true"
+  end
+end
--- a/Library/Formula/nccmp.rb
+++ b/Library/Formula/nccmp.rb
@@ -1,21 +1,24 @@
 class Nccmp < Formula
-  desc "nccmp compares 2 NetCDF files bitwise, semantically or with a tolerance."
+  desc "Compare two NetCDF files bitwise, semantically or with a tolerance."
   homepage "http://nccmp.sourceforge.net/"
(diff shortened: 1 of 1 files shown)


Library/Formula/nccmp.rb
```diff
@@ -0,0 +1,21 @@
+class Nccmp < Formula
+  desc "nccmp compares 2 NetCDF files bitwise, semantically or with a tolerance."
```
DomT4: The name will automatically prepend the `desc` output, so you don't need to start with `nccmp` here.

Library/Formula/nccmp.rb
```diff
@@ -0,0 +1,21 @@
...
+  homepage "http://nccmp.sourceforge.net/"
+  url "https://downloads.sourceforge.net/projects/nccmp/files/nccmp-1.7.4.0.tar.gz"
+  sha256 "7760b65e54f6b07ee764fe37a71eaa1710f4c6440c6fa5ee05796912865748fb"
+
+  depends_on 'netcdf'
```
DomT4: Double quotes everywhere please.

Library/Formula/nccmp.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def install
+    ENV.deparallelize
```
DomT4: Please report this upstream somewhere & link to that bug report above this line. We shouldn't need to deparallelize for build success.

Library/Formula/nccmp.rb
```diff
@@ -0,0 +1,21 @@
...
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "check"
+    system "make", "install"
+  end
+
+  test do
+    system "true"
```
DomT4: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

rsmz:
Please abandon previous formula and use new commit for formula nccmp 1.7.4.1, which uses your suggestions. Thanks.

Library/Formula/nccmp.rb
```diff
@@ -0,0 +1,24 @@
+class Nccmp < Formula
+  desc "Compare two NetCDF files bitwise, semantically or with a tolerance."
+  homepage "http://nccmp.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/nccmp/nccmp-1.7.4.1.tar.gz"
+  sha256 "985b50505dd7517710c5699cb9ee1028c6d75b06a17649bb911d838c4908dc69"
+
+  depends_on "netcdf"
```
DomT4: Ah. You can't have mandatory cross-tap dependencies in the core. This will either need to be:
```ruby
depends_on "homebrew/science/netcdf" => :optional
```
Or this formula will need submitting to the Science tap.

rsmz: Good point!  I will add to Science tap and make pull request there. Closing this request.

PR #44420, opened 2015-09-29
lcm 1.2.1 (new formula)

bluesquall:
Add formula for Lightweight Communications and Marshalling ([LCM]),
starting with release 1.2.1 from github.

[LCM]: https://lcm-proj.github.io/

### DIFF
--- a/Library/Formula/lcm.rb
+++ b/Library/Formula/lcm.rb
@@ -1,0 +1,35 @@
+class Lcm < Formula
+  desc "libraries and tools for message passing and data marshalling"
+  homepage "https://lcm-proj.github.io/"
+  url "https://github.com/lcm-proj/lcm/releases/download/v1.2.1/lcm-1.2.1.zip"
+  sha256 "55b959409fc1c0cfa705209b679dc4b3043b68f14078e192b65c1b6021c37ea6"
+
+  depends_on "pkg-config" => :build
+  depends_on "glib"
+  depends_on :java => :recommended
+  depends_on "python3" => :optional
+
+  def install
+    # Make the directory for shared jars (if it doesn't already exist).
+    # This way, we should get:
+    #   /usr/local/share/java/lcm.jar -> /usr/local/Cellar/lcm/1.2.1/share/java/lcm.jar
+    # instead of:
+    #   /usr/local/share/java -> /usr/local/Cellar/lcm/1.2.1/share/java
+    # and hopefully avoid conflicts with any other forumlae that install
+    # shared jar files.
+    mkpath "/usr/local/share/java"
+
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "unzip -p /Library/Caches/Homebrew/lcm-1.2.1.zip lcm-1.2.1/examples/types/example_t.lcm > example_t.lcm"
+    system "#{bin}/lcm-gen", "-c", "example_t.lcm"
+    puts `ls -lah`
+    # If you really want to test, look at the output of the header:
+    # puts File.read("exlcm_example_t.h")
+  end
+end


Library/Formula/lcm.rb
```diff
@@ -0,0 +1,35 @@
+class Lcm < Formula
+  desc "libraries and tools for message passing and data marshalling"
+  homepage "https://lcm-proj.github.io/"
+  url "https://github.com/lcm-proj/lcm/releases/download/v1.2.1/lcm-1.2.1.zip"
+  sha256 "55b959409fc1c0cfa705209b679dc4b3043b68f14078e192b65c1b6021c37ea6"
+
+  depends_on "pkg-config" => :build
+  depends_on "glib"
+  depends_on :java => :recommended
+  depends_on "python3" => :optional
```
DomT4: Should be `:python3` rather than `"python3"`. The Pythons are "special" dependencies in terms of how we handle them.

Library/Formula/lcm.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def install
+    # Make the directory for shared jars (if it doesn't already exist).
+    # This way, we should get:
+    #   /usr/local/share/java/lcm.jar -> /usr/local/Cellar/lcm/1.2.1/share/java/lcm.jar
+    # instead of:
+    #   /usr/local/share/java -> /usr/local/Cellar/lcm/1.2.1/share/java
+    # and hopefully avoid conflicts with any other forumlae that install
+    # shared jar files.
+    mkpath "/usr/local/share/java"
```
DomT4: You'll be barred from doing this in `def install`, I'm afraid. You could do it in `post_install`, which has less restrictions on file interaction outside the prefix.

Library/Formula/lcm.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "unzip -p /Library/Caches/Homebrew/lcm-1.2.1.zip lcm-1.2.1/examples/types/example_t.lcm > example_t.lcm"
```
DomT4: I'm afraid you can't count on this still being cached. If you need examples I'd recommend doing something like:

```ruby
pkgshare.install "examples/types/example_t.lcm"
```
And then either point to that in the test or copy & paste it to the `testpath` and use it there.

Library/Formula/lcm.rb
```diff
@@ -0,0 +1,35 @@
...
+    system "#{bin}/lcm-gen", "-c", "example_t.lcm"
+    puts `ls -lah`
```
DomT4: Don't `puts`. If you need to check a file exists you can use `assert File.exist?("filename")`

Library/Formula/lcm.rb
dunn: Also, use `HOMEBREW_PREFIX` instead of `/usr/local`; they're not always the same.

tdsmith: This should happen in the `brew link` code and not in the formula.

tdsmith: Here: https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/keg.rb#L293

Library/Formula/lcm.rb
bluesquall: Thanks. I saw that for python2, but missed the note for python3.

Library/Formula/lcm.rb
bluesquall: The `.lcm` file is only needed for the test, not for the installation, so I was trying to avoid copying everything over.

I'll revise it to simply generate a minimal example `.lcm` within the test section. If it is better practice to include all the examples, please let me know.

Library/Formula/lcm.rb
bluesquall: I was trying to provide some positive feedback that the files are in fact generated. Your suggestion is great. Thanks.

Library/Formula/lcm.rb
bluesquall: Might actually be a good idea to call this out explicitly in the [formula cookbook](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md).

Library/Formula/lcm.rb
bluesquall: Thanks for that example.

I added this because I noticed that homebrew was linking the whole `share/java` folder from this formula to `HOMEBREW_PREFIX/share/java`, and I would expect that to cause problems for any other formula that puts something in `share/java`.

From your example, it looks like *the right way* to do this is to add `java` to [`SHARE_PATHS` in `keg.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/keg.rb#L74). I'll submit a separate pull request for that.

If that isn't what you're hinting at, please suggest an alternative. Maybe you know of a way to poll the existing formulae to find out if any others are writing to `share/java`...

Library/Formula/lcm.rb
DomT4: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md#python-declarations-1

Library/Formula/lcm.rb
DomT4: In that case it would install just the one file, rather than the recursive directory. If you can write a minimal `.lcm` file for testing in the `test do` block, that's great too.

PR #44425, opened 2015-09-29
rtorrent/libtorrent: remove additional build dependencies

tzudot:


DomT4:
Could you possibly write up `test do` blocks for them at all, if that's possible? We'd love an easier way to track whether or not the two formulae actually run as expected without leaning on users to report faults, and a functionality test is a major step in catching errors seen previously before users hit them.

tzudot:
Yeah, I'll look into writing tests for these formulae.

DomT4:
Great! If we can get functional tests, more than happy to merge this.

bfontaine:
Ping?

tzudot:
Yeah, I have been busy with work. I could successfully automate a test for `rtorrent` locally using a shell script but there are dependencies which I am not sure about on how to apply in a `Homebrew` test which can be executed on `travis-ci.org`. 

`$ cat test_rtorrent.sh`
```sh
#!/bin/sh

echo system.method.set_key = event.download.finished,check_download,"execute=killall,-15,rtorrent" > ~/.rtorrent.rc
rtorrent ~/Downloads/Emacs\ Icons/Emacs.icns.modern.torrent
[[ -f "Emacs.icns.modern" ]] && echo success || echo fail
```
```sh
$ ./test_rtorrent.sh
success
```
ATM, I am unable to come up with a solution for these concerns - 
- [ ] AFAIK, `rtorrent` read from `~/.rtorrent.rc` and no other config file which can be passed on the command line.
- [ ] A torrent file (or a URL) is required to test `rtorrent` functionality.

From what I have checked there is an option `-o` which can be passed on to `rtorrent` for setting configuration but it didn't succeed while trying to set `system.method.set_key=event.download.finished,check_download,"execute=killall,-15,rtorrent"`. 
It works for simple config, e.g., `rtorrent -n -o ratio.min.set=400`.

tzudot:
As for the second concern in my previous comment, in the test, I had used a `torrent` file downloaded from my `aws` `s3` account. So, if we can arrange for an `aws` `s3` URL, a `torrent` file can be fetched from same the URL by appending `?torrent`.  

e.g., `rtorrent https://s3.amazonaws.com/<bucket>/<key>?torrent`.

bfontaine:
> It works for simple config, e.g., `rtorrent -n -o ratio.min.set=400`.

Let’s do that, it’s simpler. `$HOME` is set to the testpath during tests; so we shouldn’t have issues with `rtorrent` trying to deal with the `~/.rtorrent.rc` file.

tzudot:
@bfontaine Good to know that `$HOME` is set to the `testpath` while executing a test on `travis-ci.org`. 

However, thanks to @KannOx on `##rtorrent` irc channel for answering the question on how to import a configuration file provided on the command line. I have tested the following, and it works.
```sh
$ rtorrent -n -o import=/path/to/rtorrent.rc
```

I guess, it is better to use a configuration file provided from command line. 

As for testing the functionality of `rtorrent`, I need to look into a better approach than downloading a file from `aws` `s3`. @pyroscope on `##rtorrent` has made some suggestions on how to proceed, I'll investigate.

bfontaine:
@tzudot it’s ok to only test the config; we don’t want to rely on an external URL that might go down in the future.

tzudot:
Please note, as `rtorrent` is an `ncurses` base application, I am scheduling to send `SIGTERM` to make `rtorrent` exit after the configuration is loaded, as it seems to respect `SIGTERM`. If anyone else can suggest an alternative, I'll modify the test accordingly.

```rb
test do
  (testpath/"rtorrent.rc").write <<-EOF.undent
    schedule = close_rtorrent,1,1,"execute={killall,-15,rtorrent}"
  EOF
  system bin/"rtorrent", "-n", "-o", "import=rtorrent.rc"
end
```
I'll update the PR later today.

DomT4:
> Good to know that $HOME is set to the testpath while executing a test on travis-ci.org.

In case anyone wonders on this, we override `HOME` in `test do` and `def install` regardless of whether the build/test is done on CI or not. It's intentional sandboxing.

DomT4:
Closing in favour of #45016, which seems to include these changes as well.

PR #44428, opened 2015-09-29
Fix Salt default paths

ianbrandt:
With the current formula an attempt to use Salt without sudo and without configuration changes results in issues related to its default paths.  E.g.:

```
$ salt-call --local grains.ls
WARNING: CONFIG '/etc/salt' directory does not exist.
[ERROR   ] Could not cache minion ID: [Errno 2] No such file or directory: '/etc/salt/minion_id'
[WARNING ] Failed to open log file, do you have permission to write to /var/log/salt/minion?
[WARNING ] Although 'dmesg' was found in path, the current user cannot execute it. Grains output might not be accurate.
[ERROR   ] An un-handled exception was caught by salt's global exception handler:
OSError: [Errno 13] Permission denied: '/var/cache'
```

With the exception of the 'dmesg' issue this pull request attempts to address these warnings and errors by:

* Changing Salt's default root dir from '/' to 'HOMEBREW_PREFIX' to better align with Homebrew's '/usr/local' and "sudo is bad" conventions.
* Changing the configuration directory from 'etc/saltstack' to Salt's default 'etc/salt'.
* Ensuring the existence of 'var/log/salt' to fix log writing errors.
* Adding a 'salt-call' test case to validate all of the above.

Some questions:

* Does anyone think this is the wrong thing to do?  Presumably one could configure and sudo Salt for system-wide configuration management, but that doesn't seem like the Homebrew way to me.
* Does the attempt to migrate existing configuration at #L129-135 even make sense, particularly with regard to bottled installation?
* Is anything else needed for a proper formula revision other than #L11?
* Could any of the Ruby code be more idiomatic?
* What do I not know that I don't know?

Thanks!

### DIFF
--- a/Library/Formula/saltstack.rb
+++ b/Library/Formula/saltstack.rb
@@ -8,6 +8,7 @@ class Saltstack < Formula
   url "https://github.com/saltstack/salt/releases/download/v2015.8.0/salt-2015.8.0.tar.gz"
   sha256 "71e1cb2eb1d4b30f3247f5590c00a2089190b8f9a90c9330dc9a65fae517ec9b"
   head "https://github.com/saltstack/salt.git", :branch => "develop", :shallow => false
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


tdsmith:
This looks reasonable, thanks. Configuration should be migrated in a `post_install` method and not in `install` since `install` doesn't run on bottle pours. I'm not sure that migration is necessary if it never worked out of the box; we won't replace files that already exist in etc. If users are explicitly pointing to a configuration file now, that should keep working, I guess? I don't know salt very well.

From Travis feedback it looks like /usr/local/var/cache needs to exist as well. I think we prefer using Pathname methods like `(var/"cache").mkpath` over e.g. `mkdir_p` but either works.

ianbrandt:
Reverted to read-only test cases pending resolution of pull request #44458.

ianbrandt:
@tdsmith I believe this pull request is as good as I can think to make it with read-only testing.  Unless of course there is any further feedback I'd say this is ready to be merged.  If read-write testing becomes available I can submit another pull request to take advantage of that.

Library/Formula/saltstack.rb
```diff
@@ -141,5 +150,22 @@ def caveats; <<-EOS.undent
...
+    assertDirectory var/"log/salt"
+    assertDirectory var/"run/salt"
   end
+
+  def assertDirectory(directory)
+
+    assert directory.exist?, "#{directory} must exist"
+    assert directory.directory?, "#{directory} must be a directory"
+    assert directory.writable?, "#{directory} must be writable"
+  end
```
bfontaine: Please don’t make a method just for the test. Use a block and loop over directories instead:
```rb
[some_dir1, some_dir2, some_dir3].each do |d|
  assert d.exist?, "..."
  assert d.directory?, "..."
  assert d.writable?, "..."
end
```

mikemcquaid: :+1:

Library/Formula/saltstack.rb
```diff
@@ -141,5 +150,22 @@ def caveats; <<-EOS.undent
     system "python", "-c", "import M2Crypto"
 
     system "#{bin}/salt", "--version"
+
+    # Requires write access to not be blocked by '--sandbox'
```
mikemcquaid: That's on by default so: it will be blocked so a test that assumes it won't be will be broken.

apjanke:

PR #44434, opened 2015-09-29
Prepare for El Capitan release

xu-cheng:
cc @Homebrew/owners

### DIFF
--- a/Library/Homebrew/cmd/doctor.rb
+++ b/Library/Homebrew/cmd/doctor.rb
@@ -239,7 +239,7 @@ class Checks
   end
 
   def check_for_unsupported_osx
-    if !ARGV.homebrew_developer? && MacOS.version >= "10.11" then <<-EOS.undent
+    if !ARGV.homebrew_developer? && MacOS.version >= "10.12" then <<-EOS.undent
     You are using OS X #{MacOS.version}.
     We do not provide support for this pre-release version.
     You may encounter build failures or other breakage.
--- a/Library/Homebrew/cmd/doctor.rb
+++ b/Library/Homebrew/cmd/doctor.rb
@@ -424,10 +424,18 @@ class Checks
     __check_subdir_access "share/man"
   end
 
+  def check_access_homebrew_repository
+    unless HOMEBREW_REPOSITORY.writable_real? then <<-EOS.undent
+      The #{HOMEBREW_REPOSITORY} is not writable.
+      You should probably `chown` #{HOMEBREW_REPOSITORY}
+    EOS
+    end
+  end
+
   def check_access_usr_local
     return unless HOMEBREW_PREFIX.to_s == "/usr/local"
 
-    unless File.writable_real?("/usr/local") then <<-EOS.undent
+    unless HOMEBREW_PREFIX.writable_real? then <<-EOS.undent
     The /usr/local directory is not writable.
     Even if this directory was writable when you installed Homebrew, other
     software may change permissions on this directory. Some versions of the
@@ -435,6 +443,7 @@ class Checks
 
     You should probably change the ownership and permissions of /usr/local
     back to your user account.
+      sudo chown $(whoami):admin /usr/local
     EOS
     end
   end
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -1,4 +1,5 @@
 require "cmd/tap"
+require "cmd/doctor"
 require "formula_versions"
 require "migrator"
 require "formulary"
@@ -11,6 +12,16 @@ module Homebrew
         This command updates brew itself, and does not take formula names.
         Use `brew upgrade <formula>`.
       EOS
+    end
+
+    # check permissions
+    checks = Checks.new
+    %w[
+      check_access_usr_local
+      check_access_homebrew_repository
+    ].each do |check|
+      out = checks.send(check)
+      odie out unless out.nil?
     end
 
     # ensure git is installed


Library/Homebrew/cmd/doctor.rb
```diff
@@ -424,10 +424,18 @@ def check_access_share_man
...
+      You should probably `chown` #{HOMEBREW_REPOSITORY}
+    EOS
+    end
+  end
+
   def check_access_usr_local
     return unless HOMEBREW_PREFIX.to_s == "/usr/local"
 
-    unless File.writable_real?("/usr/local") then <<-EOS.undent
+    unless HOMEBREW_PREFIX.writable_real? then <<-EOS.undent
```
bfontaine: We should edit the line below as well:
```
The /usr/local directory is not writable.
```

Should become:

```
The #{HOMEBREW_PREFIX} directory is not writable.
```

xu-cheng: While this method only checks when HOMEBREW_PREFIX == /usr/local. so it's the same.

bfontaine: Right.

Library/Homebrew/cmd/doctor.rb
```diff
@@ -424,10 +424,18 @@ def check_access_share_man
     __check_subdir_access "share/man"
   end
 
+  def check_access_homebrew_repository
+    unless HOMEBREW_REPOSITORY.writable_real? then <<-EOS.undent
+      The #{HOMEBREW_REPOSITORY} is not writable.
+      You should probably `chown` #{HOMEBREW_REPOSITORY}
```
DomT4: Don't know if it's worth advising `sudo` here here as well?

Library/Homebrew/cmd/doctor.rb
```diff
@@ -424,10 +424,18 @@ def check_access_share_man
     __check_subdir_access "share/man"
   end
 
+  def check_access_homebrew_repository
+    unless HOMEBREW_REPOSITORY.writable_real? then <<-EOS.undent
```
DomT4: I presume there's not some edge case where it's possible for someone to run `brew doctor` when `HOMEBREW_REPOSITORY` doesn't exist and see this?

Library/Homebrew/cmd/doctor.rb
```diff
@@ -435,6 +443,7 @@ def check_access_usr_local
 
     You should probably change the ownership and permissions of /usr/local
     back to your user account.
+      sudo chown $(whoami):admin /usr/local
```
DomT4: `sudo chown -R $(whoami):admin /usr/local` maybe? It'll do the top directory and everything in it that way, rather than just the top level.

DomT4:
Thanks for chasing this down Xu. Super appreciate you taking the time here :bow:.

Library/Homebrew/cmd/doctor.rb
xu-cheng: That won't happen because that means homebrew isn't installed.

DomT4: :+1:.

Library/Homebrew/cmd/doctor.rb
xu-cheng: I'm not sure we should do this. `/usr/local` is not exclusive for homebrew. We should respect other softwares living there who choose to use root as default permission.

Library/Homebrew/cmd/doctor.rb
xu-cheng: "You should probably `chown`" is used multiple times in `doctor`. I think we should be consistent. 

Also, there are much point to mention sudo, unless we show complete command.

Library/Homebrew/cmd/doctor.rb
DomT4: I'm happy to try it without the `-R` but I fear we'll end up putting it in at some point. There's a lot of potential trip-points on a directory sharing non-root & root stuff.

xu-cheng: How about creating `brew repair-permission` command which will only repair files belong to homebrew.

Library/Homebrew/cmd/doctor.rb
DomT4: Alright. I guess I'll try and be trusting that if a `chown` fails users will know to try again with `sudo`.

Library/Homebrew/cmd/doctor.rb
DomT4: Sure. If it's feasible to know in a time-effective way everything `brew` owns I'd be happy with that. The command would end up being something like:

```
sudo chown $(whoami):admin /usr/local && brew repair-permissions
```
On a permission borked El Cap install in that case, I'd presume?

xu-cheng: I think `brew repair-permissions` should be enough as it should include every thing need to be repaired.

DomT4: Can `/usr/local` itself still be owned by `root:wheel` and be completely usable if `brew` owns everything in it?

xu-cheng: `/usr/local` itself is still better to be owned by admin. So it would complain permission problem when we add new files(like `.travis.yml` we just added recently) in the root.

DomT4: I get this locally on a test directory:

```
sudo chown root:wheel ~/Downloads/brewtest
ls -l ~/Downloads/brewtest

-rw-r--r-- 1 Dominyk admin  3161 Sep 29 15:20 CODEOFCONDUCT.md
-rw-r--r-- 1 Dominyk admin  1103 Sep 29 15:20 CONTRIBUTING.md
-rw-r--r-- 1 Dominyk admin  1241 Sep 29 15:20 LICENSE.txt
drwxr-xr-x 9 Dominyk admin   306 Sep 29 15:20 Library
-rw-r--r-- 1 Dominyk admin  2134 Sep 29 15:20 README.md
-rw-r--r-- 1 Dominyk admin 23801 Sep 29 15:20 SUPPORTERS.md
drwxr-xr-x 3 Dominyk admin   102 Sep 29 15:20 bin

brew install pkg-config
Error: Cannot write to /Users/Dominyk/Downloads/brewtest
```

xu-cheng: That message comes from check inside `brew install` which is correct behavior. Though, we may improve the message, mentioning chown or something.

xu-cheng: If you want to test this PR, you should run `brew update`

DomT4: I also get this if I `git pull` inside a `root:wheel` brew directory:

```
error: unable to unlink old '.gitignore' (Permission denied)
error: unable to unlink old '.travis.yml' (Permission denied)
error: unable to unlink old '.yardopts' (Permission denied)
error: unable to unlink old 'CODEOFCONDUCT.md' (Permission denied)
error: unable to unlink old 'CONTRIBUTING.md' (Permission denied)
error: unable to unlink old 'LICENSE.txt' (Permission denied)
error: unable to unlink old 'README.md' (Permission denied)
error: unable to unlink old 'SUPPORTERS.md' (Permission denied)
fatal: Could not reset index file to revision 'origin/master'.
```

Everything inside that directory is `Dominyk:admin` but the directory itself is `root:wheel` as per what happens to many people on El Cap upgrade. I'm just wary on the above point that `brew repair-permissions` wouldn't be able to use `sudo` to repair the `/usr/local` permissions and sometimes repairing the permissions inside the directory isn't enough.

xu-cheng: You should see warning message when running `brew update`(instead of manually `git pull`). 

As for `brew repair-permission`(though it hasn't been implemented), it should repair permissions for these files:
* `HOMEBREW_REPOSITORY` directory
* Every files in `git ls-tree`
* Every files and directories in `HOMEBREW_LIBRARY`
* `HOMEBREW_PREFIX` directory
* `HOMEBREW_PREFIX/opt` directory. 
* All directories and their subdirectories of `Keg::TOP_LEVEL_DIRECTORIES`
* `Language::Python.homebrew_site_package`
* Every files and directories in `HOMEBREW_CELLAR`
* `HOMEBREW_LOGS`
* `HOMEBREW_CACHE`
* `HOMEBREW_TEMP`

asparagui:
:fireworks:

Library/Homebrew/cmd/doctor.rb
DomT4: Given the deadline on this, If you want to merge this and discuss the potential of `brew repair-permission` over the next few days/weeks/whatever I'd be happy. I don't know if we'll have time to bash out & be happy with a new implementation before El Cap starts becoming available later today.

xu-cheng: Ok, I will merge this now along with `-R` flag in the instruction. And will replace it when we implemented `brew repair-permission` with more fine grained file inspection.

Issue #44320, opened 2015-09-25
Travis CI issues tracking


mikemcquaid:
I'd like us to be able to use Travis CI for all non-bottle builds in future (and maybe one day all bottle builds too). In the short-term this means fixing Travis so the only failures are legitimate timeouts and the only missing features are bottle uploads.

@Homebrew/owners please add any weird Travis issues you see for **any new builds started after now**  as comments in this issue and I'll try and triage them into specific issues and then fix them.

Once these are all fixed I'll switch our `master` build over to Travis CI and then move Jenkins so it requires a `@BrewTestBot test this please` to actually run jobs on Jenkins. This means Jenkins will be far less loaded and all builds that don't rely on a bottle being uploaded (e.g. audit failures, pulls of formulae without bottles) can be done without getting Jenkins involved. It'll also mean, due to Travis's parallel builds, that we can get much quicker feedback to users.

Thanks!

tdsmith:
It looks like linking after install is generating output that can overwhelm the pretty log display for e.g. boost: https://travis-ci.org/Homebrew/homebrew/jobs/82676343

dpo:
I'm very excited about this. I hope this is transferrable to taps when ready. FWIW Science's current solution is here: https://github.com/dpo/homebrew-science/blob/travis-osx/.travis.yml

xu-cheng:
Sorry accidentally closed this.

mikemcquaid:
@Homebrew/maintainers So, Travis is looking pretty reliable and faster now. I think it's probably good enough that we can maybe use it for our `master` branch feedback instead of Jenkins. In PRs I'm not seeing any non-intermittent/timeout errors that are causing Travis to fail and Jenkins to pass. I'm wondering what else other maintainers would want to see before we can disable Jenkins for PRs by default unless `@BrewTestBot test this please` is passed.

DomT4:
> I think it's probably good enough that we can maybe use it for our master branch feedback instead of Jenkins. 

:+1: :+1: :+1:  :+1:.

mikemcquaid:
I've added email notifications on failure and got it posting to the mailing list. It'll only email the relevant committer with the failure so seems :+1: to add to me.

mikemcquaid:
As soon as I receive verification that these builds are running and notifying people I'll disable the Jenkins ones.

scpeters:
A semi-related idea I've been tinkering with is using Linux machines for homebrew CI with linuxbrew:

[...]

PR #44440, opened 2015-09-29
ObserverPathnameExtension: only puts first 100 operations

xu-cheng:
Fixes https://github.com/Homebrew/homebrew/issues/44320#issuecomment-143951973

cc @tdsmith


xu-cheng:
I wonder should I enforce this unconditionally or only in Travis?

PR #44449, opened 2015-09-29
Add local bin folder to path for interactive shell

kausters:
This lets you use your Homebrew-installed and linked binaries when running commands such as `brew install --interactive --git $formula` to, for example, quickly open the package directory in Atom (using `atom .`).

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -120,6 +120,8 @@ def interactive_shell(f = nil)
     FileUtils.touch "#{ENV["HOME"]}/.zshrc"
   end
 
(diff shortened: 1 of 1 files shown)


DomT4:
I don't particularly want to go down this path. We deliberately kick `bin` out of the `$PATH` under the default env to stop accidental linkage, and we already provide a workaround for this by allowing users to pass `--env=std` to the command.

kausters:
I see. I initially thought perhaps Homebrew could have `--env=std` as the default in interactive mode, but it seems like a very bad idea to have the environment change between interactive/non-interactive mode.

DomT4:
Yes. People drop into the interactive shell for various debugging tasks. We try and keep away from `--env=std` for a host of reasons, but it is there for people to use locally if the need arises.

PR #44450, opened 2015-09-29
command: fail early if no command is given

bfontaine:
I’m not sure the message is clear enough. Without this line the command aborts in the following way:

```
$ brew command
Error: Unknown command:
```

### DIFF
--- a/Library/Homebrew/cmd/command.rb
+++ b/Library/Homebrew/cmd/command.rb
@@ -1,5 +1,6 @@
 module Homebrew
   def command
+    abort "This command requires a command argument" if ARGV.empty?
     cmd = ARGV.first
     cmd = HOMEBREW_INTERNAL_COMMAND_ALIASES.fetch(cmd, cmd)
 


Library/Homebrew/cmd/command.rb
```diff
@@ -1,5 +1,6 @@
 module Homebrew
   def command
+    abort "This command requires a command argument" if ARGV.named.empty?
     cmd = ARGV.first
```
xu-cheng: I think it should be `abort "..." unless cmd` after this line. The major reason is there are internal commands start with `-`.

xu-cheng: Except this edge case. Good catch. :+1:

bfontaine: Should I change `ARGV.named.empty?` into `ARGV.empty?` then?

Library/Homebrew/cmd/command.rb
```diff
@@ -1,5 +1,6 @@
 module Homebrew
   def command
+    abort "This command requires a command argument" if ARGV.empty?
```
bfontaine: @xu-cheng Is this ok?

xu-cheng: :+1:

PR #44455, opened 2015-09-29
lcm 1.3.0 (new formula)

bluesquall:
Add formula for Lightweight Communications and Marshalling ([LCM]),
starting with release 1.2.1 from github.

With a few modifications to the formula after feedback to pull request 44420.

[LCM]: https://lcm-proj.github.io/

### DIFF
--- a/Library/Formula/lcm.rb
+++ b/Library/Formula/lcm.rb
@@ -1,0 +1,65 @@
+class Lcm < Formula
+  desc "libraries and tools for message passing and data marshalling"
+  homepage "https://lcm-proj.github.io/"
+  url "https://github.com/lcm-proj/lcm/releases/download/v1.3.0/lcm-1.3.0.zip"
+  sha256 "23bce19a4a6dc5df444d811f57796e14b4b0d1886cb3c219f90b72f328f2184e"
+
+  head do
+    url "https://github.com/lcm-proj/lcm.git"
+
+    depends_on "xz" => :build
+    depends_on "libtool" => :build
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
+  end
+
+  depends_on "pkg-config" => :build
+  depends_on "glib"
+  depends_on :java => :recommended
+  depends_on :python => :optional
+  depends_on :python3 => :optional
+
+  def install
+    ENV.java_cache
+
+    if build.head?
+      system "./bootstrap.sh"
+    else
+      # This deparallelize setting can be removed after an upstream release
+      # that includes the revised makefile for the java part of LCM.
+      #
+      # (see https://github.com/lcm-proj/lcm/pull/48)
+      #
+      # Note that the pull request has been merged with the upstream master,
+      # so it will be included in the next release of LCM.
+      ENV.deparallelize
+    end
+
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"example_t.lcm").write <<-EOS.undent
+      package exlcm;
+
+      struct example_t
+      {
+          int64_t timestamp;
+          double position[3];
+          string name;
+      }
+    EOS
+    system "#{bin}/lcm-gen", "-c", "example_t.lcm"
+    assert(File.exist?("exlcm_example_t.h"), "lcm-gen did not generate C header file")
+    assert(File.exist?("exlcm_example_t.c"), "lcm-gen did not generate C source file")
+    system "#{bin}/lcm-gen", "-x", "example_t.lcm"
+    assert(File.exist?("exlcm/example_t.hpp"), "lcm-gen did not generate C++ header file")
+    if build.with? "java"
+      system "#{bin}/lcm-gen", "-j", "example_t.lcm"
+      assert(File.exist?("exlcm/example_t.java"), "lcm-gen did not generate java file")
+    end
+  end
+end


bluesquall:
Looks like there are some problems with OSX 10.9 and 10.11 that seem to be related to java. Is there a way to mark this formula as not working with those versions?

Library/Formula/lcm.rb
```diff
@@ -0,0 +1,26 @@
+class Lcm < Formula
+  desc "libraries and tools for message passing and data marshalling"
+  homepage "https://lcm-proj.github.io/"
+  url "https://github.com/lcm-proj/lcm/releases/download/v1.2.1/lcm-1.2.1.zip"
+  sha256 "55b959409fc1c0cfa705209b679dc4b3043b68f14078e192b65c1b6021c37ea6"
+
+  depends_on "pkg-config" => :build
+  depends_on "glib"
+  depends_on :java => :recommended
+  depends_on :python => :recommended
```
DomT4: Should be:

```ruby
depends_on :python => :recommended if MacOS.version <= :snow_leopard
```
That way we don't enforce the use of Homebrew's Python on everyone who wants to install this formula from our binary packages.

Library/Formula/lcm.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def install
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    File.write("example_t.lcm", "package exlcm;\n\nstruct example_t\n{\n    int64_t timestamp;\n    double position[3];\n    string name;\n}")
```
DomT4: Use `(testpath/"example_t.lcm").write` instead of `File`.

bluesquall:
Think I pushed the squashed changes correctly so that I don't need to close this and open a new pull request. Please let me know if I didn't.

dunn:
Yep, you got it! The build is still pending, though.

DomT4:
@BrewTestBot test this please

DomT4:
It's dying on Yosemite [for some reason](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/34594/version=yosemite/testReport/junit/brew-test-bot/yosemite/install_lcm/) :confused:

bluesquall:
It looks like the [Travis Cl build](https://travis-ci.org/Homebrew/homebrew/jobs/82814405) is only failing in `brew audit --strict --online` due to some API rate limit: `Error: GitHub API rate limit exceeded for 208.78.110.199.`

And the formula still installs on my 10.10 (Yosemite) machine just fine. I do see these java compiler notes:
```
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
```
but not the error above them that is in your [Jenkins test report](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/34594/version=yosemite/testReport/junit/brew-test-bot/yosemite/install_lcm/):
```
java.io.IOException: META-INF : could not create directory
	at sun.tools.jar.Main.extractFile(Main.java:995)
	at sun.tools.jar.Main.extract(Main.java:933)
	at sun.tools.jar.Main.run(Main.java:309)
	at sun.tools.jar.Main.main(Main.java:1233)
make[1]: *** [build/org/apache/xmlgraphics/java2d/TextHandler.class] Error 1
make[1]: *** Waiting for unfinished jobs....
```

AFAIK, this machine has the standard java installation:
screen shot 2015-10-13 at 11 44 57

Let me know how else I can help.

bluesquall:
Sounds kinda like [this bug](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=4067481), but that is really old, and marked as fixed.

Otherwise, perhaps it is a [permissions issue](http://stackoverflow.com/questions/18052460/error-while-trying-to-extract-jar-file) with your sandbox?

dunn:
Run `brew test --sandbox lcm` to use the sandbox when testing locally.

DomT4:
Try passing `ENV["_JAVA_OPTIONS"] = "-Duser.home=#{ENV["HOME"]}"` at the top of `def install`. Maybe that'll help.

bfontaine:
@DomT4 is this underscore before `JAVA_OPTIONS` intended?

DomT4:
>  is this underscore before JAVA_OPTIONS intended?

Yup. Java is a strange language.

bluesquall:
Running `brew test --sandbox lcm` on my local machine *without any changes* still works.

Adding `ENV["_JAVA_OPTIONS"] = "-Duser.home=#{ENV["HOME"]}"` doesn't break the install on my machine, but I can't verify that it will have any effect in your environment. I'd rather not add it if it isn't needed.

Also note that `brew test` passes on Travis CI, while the install itself is failing on your Jenkins bot.

What is the difference in environments between the Travis CI machine and the Jenkins machine?

dunn:
You might have to uninstall and `brew install --sandbox` to replicate the error.

bluesquall:
That did it -- I could replicate the failure and your suggestion on the java options fixed it. New squashed commits pushed.

bfontaine:

bfontaine:
The build is failing: https://travis-ci.org/Homebrew/homebrew/jobs/85654349
```
...snip...
Picked up _JAVA_OPTIONS: -Duser.home=/private/tmp/lcm20151016-4759-uu5rim/lcm-1.2.1/.brew_home
Picked up _JAVA_OPTIONS: -Duser.home=/private/tmp/lcm20151016-4759-uu5rim/lcm-1.2.1/.brew_home
Picked up _JAVA_OPTIONS: -Duser.home=/private/tmp/lcm20151016-4759-uu5rim/lcm-1.2.1/.brew_home
java.io.IOException: META-INF : could not create directory
	at sun.tools.jar.Main.extractFile(Main.java:1050)
	at sun.tools.jar.Main.extract(Main.java:981)
	at sun.tools.jar.Main.run(Main.java:311)
	at sun.tools.jar.Main.main(Main.java:1288)
make[1]: *** [build/org/apache/xmlgraphics/java2d/TextHandler.class] Error 1
make[1]: *** Waiting for unfinished jobs....
warning: [options] bootstrap class path not set in conjunction with -source 1.5
warning: [options] source value 1.5 is obsolete and will be removed in a future release
warning: [options] target value 1.5 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
4 warnings
echo timestamp > classnoinst.stamp
```

bluesquall:
Looks like something with java being unable to create a directory in the
sandbox...

apjanke:
> >  is this underscore before JAVA_OPTIONS intended?
>
> Yup. Java is a strange language.

The leading underscore means it's a private variable not intended as part of the public API. It's specific to the Sun/Oracle JVM implementation and to some degree part of its internals (like the `-X` or `-XX` command line options, but privater), and not part of the Java Language or JVM specs.

apjanke:
Would you mind rebasing this on the current `master`? It doesn't have any conflicts, but `.gitignore` and some of its dependencies have had updates since the original submission, making this a little hard to work with locally.

DomT4:
> The leading underscore means it's a private variable not intended as part of the public API.

My comment was more tongue-in-cheek than anything; I wasn't "dissing" Java :stuck_out_tongue_winking_eye:.

Library/Formula/lcm.rb
```diff
@@ -0,0 +1,27 @@
...
+  sha256 "55b959409fc1c0cfa705209b679dc4b3043b68f14078e192b65c1b6021c37ea6"
+
+  depends_on "pkg-config" => :build
+  depends_on "glib"
+  depends_on :java => :recommended
+  depends_on :python => :recommended if MacOS.version <= :snow_leopard
+  depends_on :python3 => :optional
+
+  def install
+    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{ENV["HOME"]}"
```
DomT4: You can use `ENV.java_cache` here now, after you've rebased against an up-to-date master.

bluesquall: Sounds good. I'll try to fit this in over lunch today.

bfontaine:
@bluesquall Yes it’s better to squash commits. Use `lcm 1.3.0 (new formula)`. Thanks!

bluesquall:
Only fails on XCode beta and OSX 10.9, same kind of failure as before. I'm perfectly comfortable just saying that the formula doesn't apply to those versions, if that is ok with Homebrew.
```
Picked up _JAVA_OPTIONS: -Duser.home=/Library/Caches/Homebrew/java_cache
java.io.IOException: META-INF : could not create directory
	at sun.tools.jar.Main.extractFile(Main.java:995)
	at sun.tools.jar.Main.extract(Main.java:933)
	at sun.tools.jar.Main.run(Main.java:309)
	at sun.tools.jar.Main.main(Main.java:1233)
make[1]: *** [build/org/apache/xmlgraphics/java2d/TextHandler.class] Error 1
make[1]: *** Waiting for unfinished jobs....
```

mikemcquaid:
@bluesquall yup, that's fine :+1:. You can mark it with `depends_on :macos => :yosemite`

bluesquall:
Probably can't manage squashing that change from the phone. I'll try it
later this morning.

bluesquall:
Interesting that the Travis-CI job still queued the OSX-10.9 build. Will it complete as passed when it hits that `depends_on` clause?

bluesquall:
We're back to the same java failure, but now it is showing up for Yosemite:
```
code/src/info/monitorenter/util/math/IntegerMutable.java jchart2d-code/src/info/monitorenter/util/math/MathUtil.java
Picked up _JAVA_OPTIONS: -Duser.home=/Library/Caches/Homebrew/java_cache
Picked up _JAVA_OPTIONS: -Duser.home=/Library/Caches/Homebrew/java_cache
Picked up _JAVA_OPTIONS: -Duser.home=/Library/Caches/Homebrew/java_cache
java.io.IOException: META-INF : could not create directory
	at sun.tools.jar.Main.extractFile(Main.java:1050)
	at sun.tools.jar.Main.extract(Main.java:981)
	at sun.tools.jar.Main.run(Main.java:311)
	at sun.tools.jar.Main.main(Main.java:1288)
make[1]: *** [build/com/jidesoft/comparator/AlphanumComparator.class] Error 1
make[1]: *** Waiting for unfinished jobs....
```
Why would there be permissions issues if it is running in the sandbox?

bluesquall:
Is there a way to request testbot and Travis-CI builds on this formula, but using my github fork of the upstream project instead of the release zipfile?

I'd like to check whether changes I'm making to the upstream `Makefile.am` for the java part of this project are fixing the problem in the homebrew sandbox, but I still haven't been able to replicate the problem on my machine in the first place.

Perhaps I should just keep pushing updates to this PR, but the build is passing right now and it seems I should separate this somehow.

bluesquall:
Added the `head` section to the formula, so that you can build from my fork of LCM on github while testing these java-related changes for upstream. Is there a way to tell the build to test using `--HEAD`?

bluesquall:
Looks like there's not a way to mark dependencies only for `head` and still pass `brew audit`. And there's some unaccounted dependency for `automake` on `xz` that shows up in the sandbox.

How would you suggest I proceed?

DomT4:
> Looks like there's not a way to mark dependencies only for head and still pass brew audit.

```ruby
head do
  url ""

  depends_on "blah"
  depends_on "blah2"
end
```
...

bluesquall:
Sorry to keep poking holes here, but if I add the dependencies for building from the git repo in a `head do` block, and include the ` => :build` modifier, it seems homebrew doesn't resolve them in a sandbox.

For example, with:
```ruby
  head do
    url "https://github.com/bluesquall/lcm.git", :branch => "simplify-java-Makefile.am"

    depends_on "xz" => :build
  end
```
running:
```shell
brew install --sandbox --verbose --build-bottle --HEAD --with-java lcm
```
produces the error:
```shell
HEAD is now at 60c4754 Simplify lcm-java/Makefile.am
==> ./bootstrap.sh
running "autoreconf -i"...
./bootstrap.sh: line 4: autoreconf: command not found
==> Formula
Path: /usr/local/Library/Formula/lcm.rb
...
Error: lcm HEAD did not build
```
Where `autoreconf` should be installed from formula `autoconf`, which is a dependency of `automake`.

For the time being, I've simply removed the `=> :build` modifier from those dependencies. But that means the formula won't pass `brew audit --strict`.

dunn:
> Where autoreconf should be installed from formula autoconf, which is a dependency of automake.

Even though it's installed (since it's a dependency, it's not added to the `superenv` unless you explicity add `depends_on "autoconf" => :build`.

bluesquall:
Sure enough. And now I'm finally getting the same error as the Travis-CI build with META-INF. Think that has something to do with the way LCM is repacking jar files in the build process.

Out of curiosity, why isn't `autoreconf` included in `superenv` when `automake` is?

bluesquall:
In case anyone stumbles across this PR while searching for similar error messages:

The root cause was that the makefile was extracting two jar files into the same working directory with the rest of the build so that all the classes in those two extensions could be packaged with the new classes into a single file -- `lcm.jar`. Each extraction was defined as a separate target for make. Sometimes (but not always) those targets would be built at the same time with a parallel build using multiple cores or threads. Both jar files had a META-INF directory. And even though permissions were fine in the working directory, both jobs cannot extract to the same folder at the same time.

The solution was to make both extraction operations part of the same target for make. In this case, we didn't need the classes extracted until right before packing them back up.

The interim solution for a working homebrew formula while waiting for a new release upstream is to include `ENV.deparallelize` in the `def install` block.

Thanks for the help & patience @DomT4 @dunn @bfontaine @mikemcquaid @apjanke

Library/Formula/lcm.rb
```diff
@@ -0,0 +1,48 @@
+class Lcm < Formula
+  desc "libraries and tools for message passing and data marshalling"
+  homepage "https://lcm-proj.github.io/"
+  url "https://github.com/lcm-proj/lcm/releases/download/v1.3.0/lcm-1.3.0.zip"
+  sha256 "23bce19a4a6dc5df444d811f57796e14b4b0d1886cb3c219f90b72f328f2184e"
+
+  head do
+    # This url can be changed to "https://github.com/lcm-proj/lcm.git"
+    # after PR #48 on lcm is accepted.
+    url "https://github.com/bluesquall/lcm.git", :branch => "simplify-java-Makefile.am"
```
mikemcquaid: We need to just omit `head` until this is merged.

Library/Formula/lcm.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    depends_on "xz" => :build
+    depends_on "libtool" => :build
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
+  end
+
+  depends_on "pkg-config" => :build
+  depends_on "glib"
+  depends_on :java => :recommended
```
mikemcquaid: I'd rather this is `:optional` as it's a pretty heavy dependency.

Library/Formula/lcm.rb
```diff
@@ -0,0 +1,48 @@
...
+  def install
+    ENV.java_cache
+
+    if build.head?
+      system "./bootstrap.sh"
+    else
+      # This else statement can be removed after an upstream release that
+      # includes the revised makefile for the java part of LCM.
+      # (See PR #48 at https://github.com/lcm-proj/lcm)
+      ENV.deparallelize
```
mikemcquaid: Can you link to the PR itself directly?

Library/Formula/lcm.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"example_t.lcm").write("package exlcm;\n\nstruct example_t\n{\n    int64_t timestamp;\n    double position[3];\n    string name;\n}")
```
mikemcquaid: Use `EOS.undent` rather than `\n`s.

Library/Formula/lcm.rb
bluesquall: - Java is required for two of the main tools that LCM provides: `lcm-spy` and `lcm-logplayer-gui`
- Java support is "strongly recommended" in the build instructions for LCM.

That's why I went through the headache of making the formula work with `depends_on :java => :recommended`.

Library/Formula/lcm.rb
bluesquall: Will do in next commit. Didn't realize how short URLs are for github PRs, been on Bitbucket for work recently.

https://github.com/lcm-proj/lcm/pull/48

Library/Formula/lcm.rb
bluesquall: I'd rather include the upstream head in the formula, and just apply ENV.deparallelize to both for the time being, if that's acceptable. It is useful to be able to optionally build from the git repo, and wouldn't be the default, in any case.

Library/Formula/lcm.rb
mikemcquaid: I still think it's worth being optional but Java isn't provided on OS X by default, is a pretty heavy dependency and cannot be installed by Homebrew alone.

Library/Formula/lcm.rb
mikemcquaid: That's fine, yeh.

Library/Formula/lcm.rb
bluesquall: My point is that the majority of people using LCM will expect the main tools to be installed by default (without needing to specify `--with-java`) and the few installing LCM who don't already have java and don't want those tools shouldn't have a problem specifying `--without-java`.

bluesquall:
FYI - I'm finally finding examples for some of your suggestions in [`example-formula.rb`](https://github.com/syhw/homebrew/blob/master/Library/Contributions/example-formula.rb) on a fork of homebrew that is apparently not up to date.

For what it's worth, putting that back into `HEAD` and linking it from the cookbook would be useful to people like me.

Library/Formula/lcm.rb
mikemcquaid: I'd personally still rather but I'll let other maintainers express their views.

Library/Formula/lcm.rb
```diff
@@ -0,0 +1,69 @@
...
+    system "#{bin}/lcm-gen", "-x", "example_t.lcm"
+    assert(File.exist?("exlcm/example_t.hpp"), "lcm-gen did not generate C++ header file")
+    if build.with? "java"
+      system "#{bin}/lcm-gen", "-j", "example_t.lcm"
+      assert(File.exist?("exlcm/example_t.java"), "lcm-gen did not generate java file")
+    end
+    if (build.with? "python") || (build.with? "python3")
+      system "#{bin}/lcm-gen", "-p", "example_t.lcm"
+      assert(File.exist?("exlcm/example_t.py"), "lcm-gen did not generate python file")
+    end
```
mikemcquaid: Can remove this; we don't test non-default options so it won't be run.

bluesquall: Those tests all run (and pass) when you do `brew test lcm` after `brew
install lcm --with-python`.

mikemcquaid: Our CI will never do that if it's optional though so it's better they are removed.

bluesquall: Sometimes I use `brew check foo` after installing, like I used to run `make
check` for other packages.

But if your policy is to only include tests that Travis-CI will run, we can
remove the python test.

DomT4: We'd prefer the non-default test to be removed at the present time, aye, Thanks!

bluesquall:
Python tests removed.

bluesquall:
Travis-CI failing now after *removing* tests, when it was passing before. This looks unrelated to the actual formula:
```
Error: SSL_read:: sslv3 alert bad record mac
```
Please tell me how to re-start the test.

DomT4:
Ignore Travis here. It's not as important as Jenkins.

bluesquall:
Ok. Please let me know what else you need before merging this, then.

Library/Formula/lcm.rb
DomT4: If anyone else wants to chime in on Mike's comment please do so in the next day or so. Thanks!

apjanke: I'd lean toward `:recommended`, agreeing with bluesquall's reasoning. Weighing in that Java is heavyweight, but probably likely to be already installed among LCM's main userbase, or not objectionable, so `:optional` would inconvenience most of its users and help few (those LCM users who don't have or want Java installed).

Especially since [upstream's OS X-specific build instructions](http://lcm-proj.github.io/build_instructions.html#build_osx) flat-out say to install Java, unconditionally, even though it's just "strongly recommended" in the dependency listing. If we do `:optional`, then our default installation of `lcm` deviates from upstream's standard OS X installation.

Also because the process to fix a "wrong" default installation is different: For non-Java LCM users, default installation will fail and the error message will tell them how to fix it, and then they just have to rerun with `--without-java`. For LCM users who want all the tools, default install will succeed silently but will be missing some stuff and they'll have to do some research to figure out why and how to fix it.

But this is complicated by another factor: the build boxes. The test bots on which bottles are built have the JDK installed. So the `:optional` `:java` requirement will be recognized as being met, and the bottles are built with Java support in them. So if it's `:optional`, a user can do `brew install lcm` and install it from a bottle and they will get Java support – they can get a bottle install with Java support as long as they remember to *not* specify `--with-java`. A user who doesn't have Java installed and wants to actually install LCM without Java support built will have to know to do `brew install -s lcm`, (and `brew install --without-java` won't install an actual non-Java build without `-s`) even though Java is `:optional` so you'd think that would be the default behavior. This may not matter in practice because Java-based stuff is usually done dynamically, so maybe the parts that require Java will just get turned off or be broken if you try to run them. So in practice most people will get Java support anyway, but this relies on weird edge case behavior and our build bot setup and is confusing.

You can see this in the build logs. The testing output has the `-j` Java test, so you can see that `build_with? "java"` was in effect.

```
==> /usr/local/Cellar/lcm/1.3.0/bin/lcm-gen -j example_t.lcm
```

So, maybe `:recommended` is also useful because then the bottles will line up with the formula definition, `--without-java` will always work, and the effects of `brew install lcm` options will be easier to understand. And the behavior won't change if we change our test bot configuration.

apjanke: Or is that just the expected behavior of `:optional` `:java` dependencies, and the build and run time aspects of the dependency are understood to be independent? Still weird that it means we end up with a bottle that does have Java support but you have to know not to explicitly ask for `--with-java`. If this is the case, then `:optional` makes sense because it means that most everybody gets Java support anyway.

Do you know what the failure mode of LCM is if a Java-enabled build is installed on a machine without Java? Is that just the same as doing a non-Java build, and you can just install Java without reinstalling LCM to enable the Java-dependent bits?

mikemcquaid: > So the :optional :java requirement will be recognized as being met, and the bottles are built with Java support in them

This shouldn't happen (and if it does it's a bug we can fix it) unless you specifically pass `--with-java`. That said, given your arguments above I'm :ok: with just making this `:recommended` :+1:

apjanke: Oh yeah. I misinterpreted the build output; I forgot that it's currently doing `:recommended`.

DomT4:
Alright, merging then. Thanks for the additional Java review Andrew!

Merged in https://github.com/Homebrew/homebrew/commit/fb7ff1dc625c5aeb6610ff3ef2166241ade81415. Thank you for another contribution to Homebrew @bluesquall; we appreciate it! :smiley_cat:

PR #44456, opened 2015-09-29
Add java to SHARE_PATHS

bluesquall:
Add `java` to `SHARE_PATHS` so that `HOMEBREW_PREFIX/share/java` is a folder with symlinks in it, rather than a symlink to a folder for a specific formula.

This way we avoid conflicts if multiple formulas put `jar` files in the standard location `HOMEBREW_PREFIX/share/java`.

See also: [pull request 44420](https://github.com/Homebrew/homebrew/pull/44420)


xu-cheng:

PR #44458, opened 2015-09-29
allow tests to write to var

tdsmith:
Fixes tests for things like salt that expect to be able to write to
prefix/var on launch.

### DIFF
--- a/Library/Homebrew/cmd/test.rb
+++ b/Library/Homebrew/cmd/test.rb
@@ -50,6 +50,7 @@ module Homebrew
             sandbox.allow_write_temp_and_cache
             sandbox.allow_write_log(f)
             sandbox.allow_write_xcode
+            sandbox.allow_write_path(HOMEBREW_PREFIX/"var")
             sandbox.exec(*args)
           else
             exec(*args)


DomT4:
As far as I'm aware, we've been deliberately trying to avoid letting things write test artifacts to `var` because they tend to create pids and logs and such that the user didn't actually create themselves :confused:. We've been feeding manual overrides & configs to places where this has happened so far. 

Will let Xu comment, he wrote the implementation :smile_cat:.

tdsmith:
We never run `test` on a user's behalf and I consider it predictable that `test` will launch the installed service so I don't feel that we need to avoid that.

DomT4:
We have asked users to run `brew test` as part of debugging at times though. I've also found it quite useful locally to be able to test various formulae without spamming up my `var` folder.

I don't know. Everything else on that list is temporary; logs, `/tmp` or cache directories. There's nothing designed to stick around and this feels like a drift from that policy. As hinted above though, if Xu is happy, I don't mind enough to disagree `¯\_(ツ)_/¯`.

tdsmith:
Talking to @ianbrandt on IRC, it sounds like salt also wants to be able to write state information to `prefix/etc` at runtime. On some level this is distasteful but I don't see that we have an interest in preventing it; I'd rather we have a real test.

xu-cheng:
I would like to avoid writing to `var` or `etc`. Because it may alter or interfere users' config, database etc.

xu-cheng:
Also, by allowing to write to `var` and `etc`, it would not be easy to distinguish whether we are writing innocent log and pid or unintentionally changing users' files. Which as result will diminish the isolated function offered by sandbox.

tdsmith:
I would rather rely on writing tests that behave well instead of relying on the sandbox to protect us. This configuration keeps us from testing real software in a useful way. It seems totally reasonable for an application developer to assume they can write state information or logs to var and I don't think it's a problem for the test to touch those files. The situation in etc is not as nice but I think the risk of overwriting somebody's configuration is small.

ianbrandt:
Just for some context my pull request that lead to this one is #44428.  The issue I ran into is that the current SaltStack formula only tests `salt --version`, and this proves to be a poor indicator of whether Salt is actually installed and configured to be usable as is.  In an attempt to start really validating the formula I'm trying to add a `salt-call --local grains.ls` test case, which simply asks Salt to list the local system information.  This is conceptually a read-only local operation, but it invokes Salt's core service code.  That entails being able to write log, pid, cache, and cached configuration information to 'var/log/salt', 'var/run/salt', 'var/cache/salt' and 'etc/salt' respectively.  Otherwise the command errors out.  The way Salt is written I can't think of any meaningful test case that wouldn't require this write access.  It's worth noting that what's being written here is just what would be written by any normal invocation of Salt.  I believe it very unlikely that this would do anything to upset an existing installation.

I realize anecdote isn't data, but for what it's worth my use case as a Homebrew user has always been to just install, update and uninstall packages.  I never even thought to run `brew test` on my local machine for anything.  Actually I didn't even know it existed before I tried learning about formula development.  It's not in the `brew` usage doc, nor the [FAQ](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/FAQ.md) or [Tips N' Tricks](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Tips-N'-Tricks.md) docs online, and I never needed to read the `man` page in any detail before now.

Truth be told I'd strongly prefer formula be thoroughly tested before being made available to me, so I can continue having no use for `brew test`.  I'd think that would be a win for formula developers too--less back and forth with bug reports.  For client-server and service apps like SaltStack I don't see how comprehensive testing could be done without write access.  If anything it seems to me that the `--sandbox` option (which by the way isn't documented in the `man` page or anywhere else I can find) is backwards.  I would think it should be off in the CI environment to allow for thorough formula testing there, and on by default to protect end-users should they try to run `brew test` locally without anticipating the consequences.

Perhaps what is needed is to separate tests into two suites: those that are read-only, and those that are read-write.  In other words the classic [unit vs integration/functional test](http://www.artima.com/weblogs/viewpost.jsp?thread=126923) separation.  Another idea that comes to mind is some sort of chroot or copy-on-write union filesystem approach to facilitate non-destructive read-write tests, though that seems like it might prove to be overkill.

xu-cheng:
I think we can allow test to write paths like `var/run`, `var/cache` and `var/log`. But I don't think we should allowing it to write to `var` and `etc` in general.

> I would rather rely on writing tests that behave well instead of relying on the sandbox to protect us. 

However, it's hard to tell what a formula will do to the files in `var` and `etc`. It may create innocent temp file or mess with users config files. You cannot distinguish each other simply by checking the formula, because we don't know what happens underneath.

> We never run test on a user's behalf

But users still should be able to run `brew test` by themselves to check the integrity of their installation without messing their config files.

@ianbrandt  Can the path of `etc` be passed explicitly to `salt`? Frankly, in most case, manual overrides & configs work smoothly for other formulae.

xu-cheng:
@tdsmith maybe let's allow to write to `var/run`, `var/cache` and `var/log` first? We can revisit this if there are demands on writing to `var` or `etc` in general.

PR #44467, opened 2015-09-30
Remove MacVTKITKPythonBottles from doc

iMichka:
Bottles for VTK and ITK python are no more provided.

ITK can be wrapped with Python since version 4.8, so this tap
is no more needed. Using the vanilla homebrew VTK and ITK formulas is
the way to go.


dunn:
Thanks for helping us keep everything up-to-date! Merged in 8ea675d6f7f66be06334907ccae74ab73cbced7f.

PR #44469, opened 2015-09-30
otto: add 0.1.1 (new formula)

chrisfu:
Knocked up a quick formula to build Otto, which was recently announced by HashiCorp as a replacement for Vagrant. Based this on the pre-existing formula for Vault. Audited and confirmed working in Yosimite.

Thanks.

### DIFF
--- a/Library/Formula/otto.rb
+++ b/Library/Formula/otto.rb
@@ -76,6 +76,6 @@ class Otto < Formula
   end
 
   test do
-    system "#{bin}/otto", "status"
+    system "#{bin}/otto", "--version"
   end
 end
--- a/Library/Formula/otto.rb
+++ b/Library/Formula/otto.rb
@@ -44,7 +44,7 @@ class Otto < Formula
   end
 
   def install
-    contents = Dir["{*,.git,.gitignore,.travis.yml}"]
+    contents = Dir["{*,.git,.gitignore}"]
     gopath = buildpath/"gopath"
     (gopath/"src/github.com/hashicorp/otto").install contents
 
--- a/Library/Formula/otto.rb
+++ b/Library/Formula/otto.rb
@@ -1,0 +1,81 @@
+require "language/go"
+
+class Otto < Formula
+  desc "Development and deployment system designed as the successor to Vagrant"
+  homepage "https://ottoproject.io"
+  url "https://github.com/hashicorp/otto.git",
+      :tag => "v0.1.1",
+      :revision => "257771b197d5a52bfb5c26205872969c328fca91"
+
+  head "https://github.com/hashicorp/otto.git"
+
+  depends_on "go" => :build
+
+  go_resource "github.com/tools/godep" do
+    url "https://github.com/tools/godep.git",
+        :revision => "2c6ee5e071cf01a41359e33502b7fe79a5801b9e"
+  end
+
+  go_resource "github.com/jteeuwen/go-bindata" do
+    url "https://github.com/jteeuwen/go-bindata.git",
+        :revision => "bfe36d3254337b7cc18024805dfab2106613abdf"
+  end
+
+  # godep's dependencies
+  go_resource "github.com/kr/fs" do
+    url "https://github.com/kr/fs.git",
+        :revision => "2788f0dbd16903de03cb8186e5c7d97b69ad387b"
+  end
+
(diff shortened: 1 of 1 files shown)


chrisfu:
Looks as if this is an effective pass now. The free Github API limit caused the CI fail.

Library/Formula/otto.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  # gox dependency
+  go_resource "github.com/mitchellh/iochan" do
+    url "https://github.com/mitchellh/iochan.git",
+        :revision => "87b45ffd0e9581375c491fef3d32130bb15c5bd7"
+  end
+
+  def install
+    contents = Dir["{*,.git,.gitignore,.travis.yml}"]
```
bfontaine: Why do we need the `.travis.yml` file?

chrisfu:
@bfontaine Hmm, are these legitimate failures on my part, or is this indeed effectively a pass?

bfontaine:
@chrisfu No it’s an issue on the bot; I’ll re-trigger it.

bfontaine:

Library/Formula/otto.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    cd gopath/"src/github.com/hashicorp/otto" do
+      system "make", "dev"
+      bin.install "bin/otto"
+    end
+  end
+
+  test do
+    system "#{bin}/otto", "--version"
```
dunn: Any chance of a test that checks the functionality a bit more?

chrisfu:
@dunn I'd imagine so. However, as it's a new tool that does things quite differently to Vagrant, I (and anyone else interested in this) would need to have a think about how best to test it without spawning an entire development environment. :)

It's something that I'm giving some though to, but I don't think the fairly weak test should be any sort of barrier to merging for now.

@bfontaine Any change of triggering another build please? A build failing/passing seems like roulette sometimes, heh.

dunn:
Don't worry about the Travis build failing, as long as Jenkins passes.

AndrewDryga:
+1

dunn:
Looking closer at the build process, it uses `go get` (https://github.com/hashicorp/otto/blob/master/scripts/build.sh#L24) without an accompanying `Godeps.json`.  That means it's just pulling the HEAD of each dependency, right?

We try to avoid that, which is why we have the `go_resource` method.  I know it's a hassle, but could the (recursive) dependencies be specified in the formula?  Alternatively you could ask the Otto developers to add a `Godeps.json` with dependency revisions.

bfontaine:
@dunn `go get` does pull the HEAD of each dependency; but the `Godeps.json` file is related to `godep`, a Go vendoring tool, and not to `go` itself. That means running `go get ./...` with or without a `Godeps.json` file does the same thing.
We could either ask upstream for specific revisions or just use whatever the revision currently is for each dependency and put that in `go_resource` blocks.

dunn:
Oh, gotcha. Is the Godep we're vendoring used by otto at runtime, then, rather than being used to build otto?

bfontaine:
No, `Godep` is only used at build-time; Go uses static linking by default.

maxnordlund:
How is this going?

bfontaine:
We need `go_resource`s for the external dependencies per @dunn’s comment. Right now the `build.sh` fetches external dependencies; we must put them in `go_resource`s.

The following should work:

1. Run `brew fetch otto`
2. Disable your Internet access
3. Run `brew install otto`

Right now it fails:
```
…snip…
==> make dev
find . -type f -name '.DS_Store' -delete
go generate ./...
==> Getting dependencies...
# cd .; git clone https://github.com/hashicorp/go-checkpoint /private/tmp/otto20151021-53973-n08so6/gopath/src/github.com/hashicorp/go-checkpoint
Cloning into '/private/tmp/otto20151021-53973-n08so6/gopath/src/github.com/hashicorp/go-checkpoint'...
fatal: unable to access 'https://github.com/hashicorp/go-checkpoint/': Could not resolve host: github.com
package github.com/hashicorp/go-checkpoint: exit status 128
…more errors…
```

DomT4:
Closing in favour of #45525, which seems more or less complete. Thanks for the submission nonetheless @chrisfu! :heart:

PR #44472, opened 2015-09-30
Add DNAnexus utilities formula

andyshinn:
The build for `gnureadline` uses a script to check which `-arch` LLVM can build with. It loops over each arch to see which fail. The superenv process was removing the `-arch` flags from this command thus making every arch pass. Then during the actual it would fail since ppc and ppc64 were not valid arches on that machine.

To get around this problem I found anther formula using `ENV.permit_arch_flags` and set it. But if there is a better way to solve this I'm all ears.

### DIFF
--- a/Library/Formula/dxpy.rb
+++ b/Library/Formula/dxpy.rb
@@ -1,0 +1,95 @@
+class Dxpy < Formula
+  desc "DNAnexus toolkit utilities and platform API bindings for Python"
+  homepage "https://github.com/dnanexus/dx-toolkit"
+  url "https://pypi.python.org/packages/source/d/dxpy/dxpy-0.162.1.tar.gz"
+  sha256 "095bd85bccecc1cc8b5fb1322fe773f27fe963e69d0f9f397df740c3a2523b24"
+
+  resource "futures" do
+    url "https://pypi.python.org/packages/source/f/futures/futures-3.0.3.tar.gz"
+    sha256 "2fe2342bb4fe8b8e217f0d21b5921cbe5408bf966d9f92025e707e881b198bed"
+  end
+
+  resource "ws4py" do
+    url "https://pypi.python.org/packages/source/w/ws4py/ws4py-0.3.2.tar.gz"
+    sha256 "48a4e005496a60081f74ca130ce55603ff87e1507483535acf902b94761bda8b"
+  end
+
+  resource "python-dateutil" do
+    url "https://pypi.python.org/packages/source/p/python-dateutil/python-dateutil-2.3.tar.gz"
+    sha256 "2db67d8832f19332908b4b9644865ced34087919702140862093e347e95730e4"
+  end
+
+  resource "python-magic" do
+    url "https://pypi.python.org/packages/source/p/python-magic/python-magic-0.4.6.tar.gz"
+    sha256 "903d3d3c676e2b1244892954e2bbbe27871a633385a9bfe81f1a81a7032df2fe"
+  end
+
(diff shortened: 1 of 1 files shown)


DomT4:
Looks like you're missing a Python resource here:

```
==> /usr/local/Cellar/dxpy/0.162.1/bin/dx --version
  File "/usr/local/Cellar/dxpy/0.162.1/libexec/bin/dx", line 5, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 2603, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 666, in require
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 565, in resolve
    raise DistributionNotFound(req)  # XXX put more info here
pkg_resources.DistributionNotFound: six
Error: dxpy: failed
<0> expected but was
<1>.
```

andyshinn:
Oops, it wasn't in their `requirements.txt` :confused:. Updated.

Library/Formula/dxpy.rb
```diff
@@ -0,0 +1,87 @@
...
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    system "python", *Language::Python.setup_install_args(libexec)
+
+    bin.install Dir["#{libexec}/bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    assert_match "dx v0.162.1",
+      shell_output("#{bin}/dx --version")
```
bfontaine: Can’t we write a test that does something more substantial than e.g. `--help`/`--version`?

andyshinn: I'll contact DNAnexus and see if there is some sort of API call that can be requested which doesn't require authentication. To my knowledge, all the commands require auth to the API. Is an acceptable test to try running a command and just seeing that it fails with a 401?

andyshinn:
Added a more appropriate test. I've been using the direct raw formula link for a couple people now and it has been working well. Would love to get this merged.

Library/Formula/dxpy.rb
```diff
@@ -0,0 +1,94 @@
...
+    sha256 "e24052411fc4fbd1f672635537c3fc2330d9481b18c0317695b46259512c91d5"
+  end
+
+  resource "gnureadline" do
+    url "https://pypi.python.org/packages/source/g/gnureadline/gnureadline-6.3.3.tar.gz"
+    sha256 "a259b038f4b625b07e6206bbc060baa5489ca17c798df3f9507875f2bf980cbe"
+  end
+
+  def install
+    ENV.permit_arch_flags
```
mikemcquaid: Can you add a comment with the justification you posted in the PR body here? Just so we can keep track of it in future. Thanks!

mikemcquaid:
One nit otherwise :+1:

Library/Formula/dxpy.rb
andyshinn: Added.

DomT4:
Can this work with Python's less than version 2.7?

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `dxpy 0.162.1 (new formula)`.

@DomT4 Added Python 2.7 dep in 93e7f58.

PR #44477, opened 2015-09-30
audit: fix GitHub repository audit

dlh:
In the data returned from the GitHub API, `watchers_count` is actually the number of stars.

See https://developer.github.com/v3/activity/starring/#starring-vs-watching


xu-cheng:

PR #44488, opened 2015-10-01
Remove broken dockutil dependency

keith:
plistlib has been removed from pypi. As far as I can tell, the default
OS X python ships with plistlib, so this isn't necessary. I wouldn't be
surprised if this wasn't the case for some of the older versions
supported by this formula.

I'm not sure what the best way to deal with this case is, since this might have broken older operating systems, but those are also already broken. Let me know how we should handle this case!


tdsmith:

keith:
<3 that's @tdsmith

PR #44494, opened 2015-10-01
Replace GNU patch with git apply for applying patches

### LINKED ISSUE #44436: Formula not creating symlink from patch diff
I'm editing a Homebrew formula and trying to symlink a certain file into the source directory, to be used when compiling. The link is specified in a diff, as a DATAPatch.

When applying the diff by Git itself, it creates the symlink as it should. However, when it's done via Homebrew, it creates a new file with just the path of the source file inside. This makes me think it's a bug with Homebrew since applying by Git works.

Here is the diff (note the `new file mode 120000` which indicates it's a symlink):
```
diff --git a/wordexp.h b/wordexp.h
new file mode 120000
index 0000000..1b6e119
--- /dev/null
+++ b/wordexp.h
@@ -0,0 +1 @@
+/usr/include/wordexp.h
\ No newline at end of file
```

kausters:
This PR replaces the default patching tool used by Homebrew from GNU patch to Git.

Since Git is the [preferred tool](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/build.rb#L121) for getting diffs when editing, it would make sense to apply these diffs using Git as well. Additionally, as I read it, the version of GNU patch that comes with OS X (version 2.5.8 as of OS X 10.11) does not fully support the the Git diff format, only supporting "[most features](http://savannah.gnu.org/forum/forum.php?forum_id=7361)" of it in version 2.7.

As a personal thing, this fixes #44436 by letting me define a symlink as a patch, before the `--interactive phase` of `brew install` kicks in.

I ran this PR through `brew tests` and it passed, don't know how else to test it apart from real-world usage, where I haven't had any problems. Also I'm not sure whether `ensure_git_installed!` is really necessary.

### DIFF
--- a/Library/Homebrew/patch.rb
+++ b/Library/Homebrew/patch.rb
@@ -64,9 +64,12 @@ class EmbeddedPatch
   end
 
   def apply
+    Utils.ensure_git_installed!
+
     data = contents.gsub("HOMEBREW_PREFIX", HOMEBREW_PREFIX)
-    cmd = "/usr/bin/patch"
-    args = %W[-g 0 -f -#{strip}]
+    cmd = "git"
+    args = %W[apply -p #{strip[1..-1]}]
+
     IO.popen("#{cmd} #{args.join(" ")}", "w") { |p| p.write(data) }
     raise ErrorDuringExecution.new(cmd, args) unless $?.success?
   end


Library/Homebrew/patch.rb
```diff
@@ -64,9 +64,12 @@ def contents
   end
 
   def apply
+    Utils.ensure_git_installed!
+
     data = contents.gsub("HOMEBREW_PREFIX", HOMEBREW_PREFIX)
-    cmd = "/usr/bin/patch"
-    args = %W[-g 0 -f -#{strip}]
+    cmd = "/usr/bin/git apply"
```
dunn: `cmd` should be `/usr/bin/git`, and `apply` should be the first item in `args`.

kausters:
@dunn Thanks, fixed and pushed.

Library/Homebrew/patch.rb
```diff
@@ -64,9 +64,12 @@ def contents
   end
 
   def apply
+    Utils.ensure_git_installed!
+
     data = contents.gsub("HOMEBREW_PREFIX", HOMEBREW_PREFIX)
-    cmd = "/usr/bin/patch"
-    args = %W[-g 0 -f -#{strip}]
+    cmd = "/usr/bin/git"
```
mistydemeo: We actually can't guarantee that git is at `/usr/bin/git`; prior to Xcode 4, Apple didn't ship it.

DomT4: Believe we also permit whatever `git` happens to fall in the `$PATH` everywhere else, so would be nice to be consistent here.

kausters: So I should probably change it to just `git` instead of `/usr/bin/git`?

xu-cheng:
Sorry, but we cannot just replace `patch` to `git apply` because it will break patches whose format aren't `git diff`.(e.g. `malbolge` formula)

kausters:
Hmm, maybe we could have a solution where, if `git apply` fails (like with `malbolge`), it falls back to `patch`?

DomT4:
Passing on this for now. Thanks for the submission though! :heart:

mikemcquaid:
Sorry, missed this the first time as I was on holiday. I do think it's worth considering because Git's binary patches cannot currently be applied by Homebrew which has caused me pain a few times in the past.

DomT4:
Apologies; presumed from the lack of comments/progress here and Xu's objection further up this was dead.

mikemcquaid:
No worries @DomT4, I didn't notice until it was closed. Just wanted to have a bit more discussion on the idea (even if we want to fiddle with the implementation)

bfontaine:
Maybe we could add an option to patches similarly to what we use for URLs (e.g. `:using => FooStrategy`)?

mikemcquaid:
> Maybe we could add an option to patches similarly to what we use for URLs (e.g. :using => FooStrategy)? Something that would say “this patch needs git apply” or “that one needs patch”.

Yeh, this is a good idea :+1:

MikeMcQuaid:
Passing on this as-is but we'd accept another PR that made this optional instead of compulsory.

MikeMcQuaid:
(another PR on https://github.com/homebrew/brew)

PR #44498, opened 2015-10-01
charm-tools 1.7.1

marcoceppi:
This updates charm-tools to 1.7.1, a patch release which addresses several minor bugs in the charm generation logic.

Furthermore, as of 1.7.0, charm-tools requires pip as a dependency on the system, but homebrew recommends users do so via `easy_install` instead of `brew install python`. I'd like to know how best to update the formula to make sure pip is installed in a way that charm-tools can access it. Should a call out to easy_install in the install block be added, or is there another way to achieve this?

### DIFF
--- a/Library/Formula/charm-tools.rb
+++ b/Library/Formula/charm-tools.rb
@@ -1,8 +1,8 @@
 class CharmTools < Formula
   desc "Tools for authoring and maintaining juju charms"
   homepage "https://launchpad.net/charm-tools"
-  url "https://launchpad.net/charm-tools/1.7/1.7.0/+download/charm-tools-1.7.0.tar.gz"
-  sha256 "6bc12d24460b366e12176538692d5b29c3697f5c8a98b525a05fa5ec7b04e042"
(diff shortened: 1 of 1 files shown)


DomT4:
> I'd like to know how best to update the formula to make sure pip is installed in a way that charm-tools can access it. 

Add a `pip` resource probably:

```ruby
  resource "pip" do
    url "https://pypi.python.org/packages/source/p/pip/pip-7.1.2.tar.gz"
    sha256 "ca047986f0528cfa975a14fb9f7f106271d4e0c3fe1ddced6c1db2e7ae57a477"
  end
```

[This is a good full example](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/jenkins-job-builder.rb) of the process.

marcoceppi:
Thank you for the examples, I've updated the formula to reflect this addition!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/ba49bd4026e37e6d0484c1b2bc0238c5203888b8. Thank you for your contribution to Homebrew @marcoceppi!

PR #44504, opened 2015-10-01
zabbix: add foreground mode using patch

gips0n:
[launchd requires daemons to run in foreground mode](https://developer.apple.com/library/mac/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLaunchdJobs.html#//apple_ref/doc/uid/10000172i-SW7-104302), but zabbix still [didn't support it](https://support.zabbix.com/browse/ZBXNEXT-611), so we can add it using patch from Boris Manojlovic, which is already used it in [opensuse](https://build.opensuse.org/package/show/server:monitoring:zabbix/zabbix24)

### DIFF
--- a/Library/Formula/zabbix.rb
+++ b/Library/Formula/zabbix.rb
@@ -12,6 +12,7 @@ class Zabbix < Formula
 
   option "with-mysql", "Use Zabbix Server with MySQL library instead PostgreSQL."
   option "without-server-proxy", "Install only the Zabbix Agent without Server and Proxy."
+  option "with-foreground-mode-patch", "Install Zabbix with foreground patch https://support.zabbix.com/browse/ZBXNEXT-611"
 
   deprecated_option "agent-only" => "without-server-proxy"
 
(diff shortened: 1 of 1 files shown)


gips0n:
Tested on OS X 10.10 and OS X 10.11, works fine.
N.B. Used patch doesn't support logs redirection to stdout/stderr, everything will be logged into file defined by LogFile option

gips0n:
@DomT4 can you (or somebody else) review it, please?

mikemcquaid:
Has this patch been accepted and merged upstream? Otherwise we'll need to pass on it, sorry!

gips0n:
No, it's not merged, you can find details on [issue page](https://support.zabbix.com/browse/ZBXNEXT-611). But:
* it's must-have patch for OS X
* patch developer still supporting it and patch is available for 2.2, 2.4 and even for 3.0.0 alpha 3 release

DomT4:
What functionality is lost without the patch? The ability to have `launchctl` manage it, or?

gips0n:
yes, you're correct - [the ability to manage zabbix by launchd](https://developer.apple.com/library/mac/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLaunchdJobs.html#//apple_ref/doc/uid/10000172i-SW7-104302)

apjanke:
Upstream have incorporated this functionality in 3.0, per https://support.zabbix.com/browse/ZBXNEXT-611. But looks like it'll never officially make it in to 2.x. What does that mean in terms of Homebrew policy?

mikemcquaid:
I'd rather wait for a backport of the production change, applying it to only `devel do` if necessary and not making it an option.

apjanke:
Closing this out as won't-accept. We'll wait until upstream backports it or releases a stable 3.0.

Which is now! 3.0.2 [is out](https://sourceforge.net/projects/zabbix/files/ZABBIX%20Latest%20Stable/3.0.2/). If you want to submit a PR to the new [`homebrew/core` repo](https://github.com/homebrew/homebrew-core) for zabbix 3.0.2, we should be able to take that.

PR #44512, opened 2015-10-01
Convert jsoncpp formula to use cmake

d235j:
This simplifies the formula greatly. Additionally, a cmake build file and a pkgconfig file are automatically provided.


dunn:
Thanks so much! Merged in 2fb1582f289673f9ff76148c25bfbd507b455cbb.

PR #44515, opened 2015-10-02
zanata-cli 3.7.4 (new formula)

Fryguy:
~~I'm not sure on the name.  The resultant bin file is `zanata-cli`, but the RPM package for Fedora is called `zanata-client`.~~  UPDATE: Using `zanata-client`

---
Some more info:

"Zanata is a web-based system for translators, content creators and developers to manage localisation projects."
Some notable projects using Zanata are [Fedora](https://fedoraproject.org/wiki/L10N) and [OpenStack](https://translate.openstack.org/).

http://zanata.org/
https://github.com/zanata/
https://github.com/zanata/zanata-client

### DIFF
--- a/Library/Formula/zanata-client.rb
+++ b/Library/Formula/zanata-client.rb
@@ -1,0 +1,18 @@
+class ZanataClient < Formula
+  desc "Zanata translation system command-line client"
+  homepage "http://zanata.org/"
+  url "https://search.maven.org/remotecontent?filepath=org/zanata/zanata-cli/3.7.4/zanata-cli-3.7.4-dist.tar.gz"
+  sha256 "4424322bffb81f5185f87d1f11c7e92cde504c736695dbb4ced5854e49672037"
+
+  depends_on :java => "1.7+"
+
+  def install
+    libexec.install Dir["*"]
+    (bin/"zanata-cli").write_env_script libexec/"bin/zanata-cli", Language::Java.java_home_env("1.7+")
+    bash_completion.install libexec/"bin/zanata-cli-completion"
+  end
+
+  test do
+    assert_match /Zanata Java command-line client/, shell_output("zanata-cli --help")
+  end
+end


DomT4:
> but the RPM package for Fedora is called zanata-client.

If other packagers are calling it `zanata-client` let's go for that.

Library/Formula/zanata-cli.rb
```diff
@@ -0,0 +1,19 @@
...
+  depends_on :java => "1.7+"
+
+  def install
+    libexec.install Dir["*"]
+    bin.install_symlink "#{libexec}/bin/zanata-cli"
+    bash_completion.install "#{libexec}/bin/zanata-cli-completion"
+  end
+
+  test do
+    %x(JAVA_HOME=\`/usr/libexec/java_home\` zanata-cli --help) =~
```
DomT4: If you need to set `JAVA_HOME` here do it with `ENV["JAVA_HOME"]` rather than this way, but it may not be necessary. For the actual check you can do:

```ruby
assert_match /Zanata Java command-line client/, shell_output("#{bin}/zanata-cli --help")
```

Fryguy:
Thanks @DomT4 .  Updated as you suggested.  I also added a caveat about `JAVA_HOME` being required for `zanata-cli`, and how one can add it to their profile.  Let me know if the wording is good for you.

dunn:
Just one audit failure:
```
==> audit problems
zanata-client:
 * `caveats method` (line 20) should be put before `test block` (line 15)
```

Fryguy:
Doh! I forgot to audit.  Thanks for the reminder :)

Fryguy:
Updated to fix the auditing.

Library/Formula/zanata-client.rb
```diff
@@ -0,0 +1,29 @@
...
+    bash_completion.install "#{libexec}/bin/zanata-cli-completion"
+  end
+
+  def caveats
+    <<-EOF.undent
+    The JAVA_HOME environment variable must be set in order to use zanata-cli.
+    If you do not already have this value in your profile, you can put the
+    following in your profile:
+
+      JAVA_HOME=`/usr/libexec/java_home`
```
DomT4: We can actually do this for users, with something like:

```ruby
(bin/"zanata-cli").write_env_script libexec/"bin/zanata-cli", Language::Java.java_home_env
```

bfontaine:
Ping?

Library/Formula/zanata-client.rb
```diff
@@ -0,0 +1,29 @@
+class ZanataClient < Formula
+  desc "Zanata translation system command-line client"
+  homepage "http://zanata.org/"
+  url "http://search.maven.org/remotecontent?filepath=org/zanata/zanata-cli/3.7.4/zanata-cli-3.7.4-dist.tar.gz"
```
bfontaine: You can use `https://` for this `url`.

Fryguy:
Thanks for all the help, everyone!

@DomT4 Updated to remove the caveat and instead use write_env_script
@bfontaine Updated to use the https URL.

Library/Formula/zanata-client.rb
```diff
@@ -0,0 +1,18 @@
...
+  desc "Zanata translation system command-line client"
+  homepage "http://zanata.org/"
+  url "https://search.maven.org/remotecontent?filepath=org/zanata/zanata-cli/3.7.4/zanata-cli-3.7.4-dist.tar.gz"
+  sha256 "4424322bffb81f5185f87d1f11c7e92cde504c736695dbb4ced5854e49672037"
+
+  depends_on :java => "1.7+"
+
+  def install
+    libexec.install Dir["*"]
+    (bin/"zanata-cli").write_env_script libexec/"bin/zanata-cli", Language::Java.java_home_env
```
DomT4: Should this be `Language::Java.java_home_env("1.7+")` @xu-cheng?

xu-cheng: Yes, it should be.

bfontaine:
Ping on @DomT4’s comment above.

Library/Formula/zanata-client.rb
Fryguy: Thanks, I'll change it.  Out of curiosity, would it be possible for the `Language::Java.java_home_env` bit to just automatically determine that versioning based on the `depends_on` line above?  It feels kind of weird to have to specify the dependency twice.

Fryguy:
Updated based on @DomT4 's suggestion.

Library/Formula/zanata-client.rb
bfontaine: @Fryguy `Language::Java.java_home_env` doesn’t know in which environment it gets called so it can’t automatically determine that versioning. I guess we could add a method on `Formula` that wraps this call and use the correct versioning if we wanted to.

Library/Formula/zanata-client.rb
```diff
@@ -0,0 +1,18 @@
...
+  depends_on :java => "1.7+"
+
+  def install
+    libexec.install Dir["*"]
+    (bin/"zanata-cli").write_env_script libexec/"bin/zanata-cli", Language::Java.java_home_env("1.7+")
+    bash_completion.install libexec/"bin/zanata-cli-completion"
+  end
+
+  test do
+    assert_match /Zanata Java command-line client/, shell_output("zanata-cli --help")
```
bfontaine: Use `"#{bin}/zanata-cli --help"` instead of `"zanata-cli --help"`.

DomT4:
@BrewTestBot test this please

DomT4:
Fixed the last nit in post.

Merged in https://github.com/Homebrew/homebrew/commit/837ad4161e740e8f1a7c0b525c20f29bcc68dfa7. Thank you for your contribution to Homebrew @Fryguy; we appreciate it! :smiley_cat:

Fryguy:
Thanks @DomT4!  I had been away on business for a week, and hadn't seen the last comment.  I appreciate the added fix.

PR #44526, opened 2015-10-02
Fix crash in gtk+

dlech:
There are several invalid object checks in the quartz backend that
cause a crash on OS X 10.11. This crash can be reproduced by running
gtk-demo and double-clicking on "Clipboard" in the list.

The patch is an adaption of a patch for gtk+3 from an upstream
developer.

### DIFF
--- a/Library/Formula/gtk+.rb
+++ b/Library/Formula/gtk+.rb
@@ -3,7 +3,7 @@ class Gtkx < Formula
   homepage "http://gtk.org/"
   url "https://download.gnome.org/sources/gtk+/2.24/gtk+-2.24.28.tar.xz"
   sha256 "b2c6441e98bc5232e5f9bba6965075dcf580a8726398f7374d39f90b88ed4656"
-  revision 2
+  revision 3
 
   bottle do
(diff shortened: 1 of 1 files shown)


DomT4:
CC @tschoonj

tschoonj:
I don't run El Cap yet so cannot really the bug but it sure looks legit

One thing though: seeing as you are pulling in my big gnome 3.18.0 branch now which include gtk+3 3.18.0, the revision number in this patch will need to be set to .

tschoonj:
I filed a PR ( #44583 ) for gtk+3 3.18.1. Could you check if your patch is already included? Thanks!

dlech:
This patch/pull request is for `gtk+`, not `gtk+3`, so it is still needed regardless of what happens to `gtk+3`.

There is a similar patch for `gtk+3`, but it is not included in `3.18.1` (it was applied to the upstream master yesterday, but after the point release, so it should be included in `3.18.2` whenever that is released).

tschoonj:
Oh missed that important piece of information completely :smile:

DomT4:
We've all done that before :smile_cat:.

Merged in https://github.com/Homebrew/homebrew/commit/ecaf88ac6312174cb8c84c2e9df92843f61e6382. Thank you for your contribution to Homebrew @dlech!

PR #44527, opened 2015-10-02
ffts 0.7: update dead link

chadwackerman:


Library/Formula/ffts.rb
```diff
@@ -2,8 +2,9 @@ class Ffts < Formula
   desc "C library that computes the discrete Fourier transform"
   homepage "http://anthonix.com/ffts/"
   head "https://github.com/anthonix/ffts.git"
-  url "http://anthonix.com/ffts/releases/ffts-0.7.tar.gz"
-  sha256 "6362e498e5aa241661cfe9060153076b78e300a5d3c365997ba6ac2f637df3ff"
+  url "https://github.com/anthonix/ffts/archive/fa1780c68593762b1e4bdbc46d83912db3eba27a.tar.gz"
```
DomT4: We're going to need an actual tag or tarball here. Can you find out upstream why the releases vanished?

DomT4: The homepage is also dead & there hasn't been an upstream commit in 18+ months, so this may end up as a bit of a rabbit-hole.

We could move it to `homebrew/homebrew-head-only` and remove the release for now.

chadwackerman: Sorry -- that's all I got. Code matches, my tests pass. Ideally the sha256 would too, but there are additional non-source files in the GitHub tarball.

apjanke:
If someone can find a copy of the original file (with matching shasum), we could just host it on our bintray.

No luck for me with Internet Archive for http://anthonix.com. They have the main pages, but not the downloads. But maybe there's another copy floating around out there?

mikemcquaid:
Sounds like it's dead and should be boneyarded.

apjanke:
Probably. It was hard for me to find any activity related to it in the past couple years by searching Google.

bfontaine:
I tried searching too, with no luck: the author wrote a paper on the library [here](http://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=6557535) but it points to an [almost empty page](http://www.cs.waikato.ac.nz/~ablake/ffts/) which itself points to the GitHub repo.

DomT4:
Nothing else uses it, hasn't been an upstream commit in nearly 2 years, Andrew's upstream request a month ago has gone unnoticed despite a direct ping. Let's kill it off, IMO.

PR #44539, opened 2015-10-02
ola: with-ftdi option

Hazardius:
Option in OLA recipe for adding dependencies to build it with FTDI USB plugin.

### DIFF
--- a/Library/Formula/ola.rb
+++ b/Library/Formula/ola.rb
@@ -19,6 +19,7 @@ class Ola < Formula
   end
 
   option :universal
+  option "with-ftdi", "Install FTDI USB plugin for OLA."
 
   depends_on "pkg-config" => :build
   depends_on "cppunit"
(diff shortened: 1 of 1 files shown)


bfontaine:
How common is this need?

Hazardius:
I'm not fully aware about needs of community, but I think FTDI devices are quite popular these days (partially because they're really cheap) and this change makes it much easier to use f.e. OpenDMX USB 2.0 with OLA.

I needed OLA with FTDI plugin myself and after figuring out how to do that with brew recipe I asked about feedback on [open-lighting google mailing group](https://groups.google.com/forum/#!msg/open-lighting/7ZZc6sN0yHE/Ug6MI7msCgAJ) what do they think about this option. They proposed to send it and that's how I've got here. :)

I hope it'll be included in homebrew in this or better manner since I'm using it a lot recently.

bfontaine:
Thanks. Don’t we need to add an `--enable-something` option? Or is it automatically picked by the build?

Hazardius:
Yes - if there's no libftdi amongst dependencies - it disables installing FTDI plugin.

bfontaine:
Thank you @Hazardius! :tada:

PR #44543, opened 2015-10-02
Allow tapping github urls

adzenith:
Allows one to write:
```
brew tap https://github.com/user/repo
```
whereas previously one had to write:
```
brew tap user/repo
```
Makes copy/pasting from the address bar just a little bit easier.

### DIFF
--- a/Library/Homebrew/tap_constants.rb
+++ b/Library/Homebrew/tap_constants.rb
@@ -1,5 +1,7 @@
+# allows tap names to start with https://www.github.com/, which is then stripped off
+HOMEBREW_TAP_GITHUB_URL_PREFIX = %r{(?:(?:(?:https?://)?(?:www\.)?)?github\.com/)?}
 # match expressions when taps are given as ARGS, e.g. someuser/sometap
-HOMEBREW_TAP_ARGS_REGEX = %r{^([\w-]+)/(homebrew-)?([\w-]+)$}
+HOMEBREW_TAP_ARGS_REGEX = %r{^#{HOMEBREW_TAP_GITHUB_URL_PREFIX}([\w-]+)/(homebrew-)?([\w-]+)$}
 # match taps' formulae, e.g. someuser/sometap/someformula
 HOMEBREW_TAP_FORMULA_REGEX = %r{^([\w-]+)/([\w-]+)/([\w+-.]+)$}
 # match core's formulae, e.g. homebrew/homebrew/someformula


Library/Homebrew/tap_constants.rb
```diff
@@ -1,9 +1,11 @@
+# allows tap names to start with https://www.github.com/, which is then stripped off
+HOMEBREW_TAP_URL_REGEX = %r{(?:(?:(?:https?://)?(?:www\.)?)?github\.com/)?}
```
bfontaine: Something like `github_prefix` might be more explicit here.

adzenith: What about HOMEBREW_TAP_GITHUB_URL_PREFIX? This file is `require`d in global.rb, so I wanted to keep the HOMEBREW_TAP_ prefix on the name.

Library/Homebrew/tap_constants.rb
```diff
@@ -1,9 +1,11 @@
...
 # match expressions when taps are given as ARGS, e.g. someuser/sometap
-HOMEBREW_TAP_ARGS_REGEX = %r{^([\w-]+)/(homebrew-)?([\w-]+)$}
+HOMEBREW_TAP_ARGS_REGEX = %r{^#{HOMEBREW_TAP_URL_REGEX}([\w-]+)/(homebrew-)?([\w-]+)$}
 # match taps' formulae, e.g. someuser/sometap/someformula
-HOMEBREW_TAP_FORMULA_REGEX = %r{^([\w-]+)/([\w-]+)/([\w+-.]+)$}
+HOMEBREW_TAP_FORMULA_REGEX = %r{^#{HOMEBREW_TAP_URL_REGEX}([\w-]+)/([\w-]+)/([\w+-.]+)$}
```
xu-cheng: You should not change this, otherwise it will break many internal codes.

Library/Homebrew/tap_constants.rb
```diff
@@ -1,9 +1,11 @@
...
 # match core's formulae, e.g. homebrew/homebrew/someformula
-HOMEBREW_CORE_FORMULA_REGEX = %r{^homebrew/homebrew/([\w+-.]+)$}i
+HOMEBREW_CORE_FORMULA_REGEX = %r{^#{HOMEBREW_TAP_URL_REGEX}homebrew/homebrew/([\w+-.]+)$}i
```
xu-cheng: Same here, which is a regex unrelated to tap.

adzenith: Oh, really? It had `TAP` in the name so I assumed it was only used by `tap`.

Library/Homebrew/tap_constants.rb
adzenith: Do you know which ones it will break? I glanced through the code and it seemed like everything would still work.

xu-cheng:
I wonder do we really need this. I know it can be convenient but it can cause some confusion. For example, if I run `brew tap https://github.com/a/b`, what I really get is `https://github.com/a/homebrew-b`.

Library/Homebrew/tap_constants.rb
xu-cheng: We have defined tap formula's full-qualified name as `user/repo/formula` since we begin to support tap. And such format is used almost everywhere in the core and some external commands. Therefore, we cannot change it.

Library/Homebrew/tap_constants.rb
xu-cheng: There is not `Tap` in this variable name. :wink:

adzenith:
I think it only works the other way: if you type `brew tap https://github.com/a/homebrew-b`, then it shows up as having tapped `a/b`.

adzenith:
How's this look?

xu-cheng:
Sorry, but I think we probably should pass this, because this change can cause confusing:
* If I run `brew tap https://github.com/a/b`, it will actually tap `https://github.com/a/homebrew-b`. 
* It gets more confusing if someone runs `brew tap https://github.com/a/b <another URL>` as it will create a tap `a/b` using a completely different URL. 

Unless another maintainers object, I'm gonna close this. Thanks for your PR, we appreciate your effort and contribution.

adzenith:
Both good points! Thanks for your consideration.

PR #44545, opened 2015-10-02
wordnet: Fix build, add test

dholm:


DomT4:
What was the prior failure seen here?

Library/Formula/wordnet.rb
```diff
@@ -22,8 +20,15 @@ def install
...
-                          "--mandir=#{man}"
+                          "--mandir=#{man}",
+                          "--with-tcl=#{MacOS.sdk_path}/usr/lib",
+                          "--with-tk=#{MacOS.sdk_path}/usr/lib"
     ENV.deparallelize
     system "make", "install"
   end
+
+  test do
+    output = `#{bin}/wn homebrew -synsn`
```
DomT4: Use `shell_output` or `pipe_output` if at all possible, Thanks.

dholm:
@DomT4 Because it couldn't find TCL it didn't build any of the binaries and only installed the dictionaries.

Library/Formula/wordnet.rb
dholm: Fixed

DomT4:
It may be the case that on CLT-only installations this doesn't resolve the issue, but we'll deal with that hurdle if someone reports it.

Merged in https://github.com/Homebrew/homebrew/commit/1a40d5085077d9ba60c583bbc80000c316d314a9. Thank you for another contribution to Homebrew @dholm! :smiley_cat:

PR #44546, opened 2015-10-02
libfabric 1.1.1 (new formula)

sungeunchoi:
OpenFabrics libfabric is an open source, community-developed library
designed to provide low-latency interfaces to high-performance fabric
hardware.  Support on OS-X is via the sockets "provider", which is
intended to be a development platform for middleware clients such as
MPICH, OpenMPI, OpenSHMEM and Chapel.

### DIFF
--- a/Library/Formula/libfabric.rb
+++ b/Library/Formula/libfabric.rb
@@ -1,0 +1,17 @@
+class Libfabric < Formula
+  desc "OpenFabrics libfabric"
+  homepage "https://ofiwg.github.io/libfabric/"
+  url "http://downloads.openfabrics.org/downloads/ofi/libfabric-1.1.1.tar.bz2"
+  sha256 "228c50d1f9595009a026bb2293f372a04d89ecb3b0f7fcde3905476dbf49fc95"
+
+  def install
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "#(bin}/fi_info"
+  end
+end


dunn:
A few audit faliures:
```
==> audit problems
libfabric:
 * Please use https:// for http://ofiwg.github.io/libfabric/
 * "#{prefix}/bin" should be "#{bin}"
```

sungeunchoi:
@dunn Fixed audit problems.  Thanks!

Library/Formula/libfabric.rb
```diff
@@ -0,0 +1,18 @@
+class Libfabric < Formula
+  desc "OpenFabrics libfabric"
+  homepage "https://ofiwg.github.io/libfabric/"
+  url "http://downloads.openfabrics.org/downloads/ofi/libfabric-1.1.1.tar.bz2"
+  sha256 "228c50d1f9595009a026bb2293f372a04d89ecb3b0f7fcde3905476dbf49fc95"
+  head "https://github.com/ofiwg/libfabric"
```
dunn: This should have `.git` at the end: `head "https://github.com/ofiwg/libfabric.git"`. Sorry I didn't catch that the first time around.

sungeunchoi:
@dunn Fixed head URL.

Library/Formula/libfabric.rb
```diff
@@ -0,0 +1,18 @@
+class Libfabric < Formula
+  desc "OpenFabrics libfabric"
+  homepage "https://ofiwg.github.io/libfabric/"
+  url "http://downloads.openfabrics.org/downloads/ofi/libfabric-1.1.1.tar.bz2"
+  sha256 "228c50d1f9595009a026bb2293f372a04d89ecb3b0f7fcde3905476dbf49fc95"
+  head "https://github.com/ofiwg/libfabric.git"
```
DomT4: You'll need to either remove the `head` or fix it. It won't currently build because they use the autotools to generate the `configure` script so the call to `./configure` further below will just die on `HEAD`.

Just needs something like:

```ruby
head do
  url "https://github.com/ofiwg/libfabric.git"

  depends_on "libtool"  => :build
  end
```

And then:

```ruby
def install
```

And so on.

sungeunchoi:
@DomT4 I think we won't be wanting use head, so I removed it.  Thanks.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/458582ab0d5dc95de31a515e857493191c58d5a9. Thank you for your contribution to Homebrew @sungeunchoi; we appreciate it! :smiley_cat:

PR #44552, opened 2015-10-03
ssdb 1.9.2

felixbuenemann:
* Update to 1.9.2
* Patch and install ssdb-cli
* Separate data directory for ssdb slave
* Fix ssdb-ins.sh script

### DIFF
--- a/Library/Formula/ssdb.rb
+++ b/Library/Formula/ssdb.rb
@@ -1,8 +1,8 @@
 class Ssdb < Formula
   desc "NoSQL database supporting many data structures: Redis alternative"
   homepage "http://ssdb.io/"
-  url "https://github.com/ideawu/ssdb/archive/1.8.2.tar.gz"
-  sha256 "2fe10689a0e1e9e9ca67acbe9e0fa3607810dca90dcc9f0813b2661cc6d4e826"
+  url "https://github.com/ideawu/ssdb/archive/1.9.2.tar.gz"
+  sha256 "9387ebeaf24f4e3355967ba2459b59be683f75d6423f408ce4cefed2596b4736"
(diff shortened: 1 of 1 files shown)


felixbuenemann:
Swapped back some lines, to make the diff more readable.

felixbuenemann:
Hmm, the ssdb-ins.sh is still broken, I think it doesn't make sense to include it, because it's basically a SysV init script. I only kept it because it was in the previous release.

dunn:
Have these patches been reported upstream?

felixbuenemann:
I don't really think it makes sense, because the patches adjust for the changes in directory structure done  by the formula itself. The ssdb-ins.sh could probably be changed upstream to be able to work with both relative and absolute pidfile paths, but that still leaves the difference in the bin path.

dunn:
Since there's a `make` step, shouldn't that be configurable?

felixbuenemann:
The ssdb-ins.sh file is not generated by make.

felixbuenemann:
@dunn I agree that the amount of patching required to get this package to work fine with homebrews directory structure is a bit unhealthy and I'd be preferable to have the upstream build system improved to accommodate for that, but I really don't have the time to do that right now. So maybe we could live with the patches for the time being?

mikemcquaid:
@dunn Doesn't look too bad to me but feel free to disagree.

dunn:
@BrewTestBot test this please

dunn:
Right-o. Merged in 779726ab031c9dcaae782329f68e3e8a69459176, thank you @felixbuenemann!

felixbuenemann:
Thanks, Alex!

PR #44553, opened 2015-10-03
nut 2.7.3: fix usb option

reitermarkus:
Signed-off-by: Markus Reiter <me@reitermark.us>

### DIFF
--- a/Library/Formula/nut.rb
+++ b/Library/Formula/nut.rb
@@ -55,7 +55,7 @@ class Nut < Formula
             "--without-wrap"
            ]
     args << (build.with?("serial") ? "--with-serial" : "--without-serial")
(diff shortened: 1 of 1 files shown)


dunn:
Since `depends_on "libusb-compat" => :recommended` is still present, this is going to create two options.  We should just fix the second line to read:
```ruby
args << (build.with?("libusb-compat") ? "--with-usb" : "--without-usb")
```

DomT4:
If we're changing an option name it should be `deprecated` onto the new name as well.

reitermarkus:
Ok, I reverted the first line and only changed the second line.

dunn:
Thanks! Merged in 86fc06d2a060c55650375e3761dd2cd9086948eb.

PR #44554, opened 2015-10-03
Remove redundant chown commands.

CodeGnome:
`chown -R` should be sufficient. There is no obvious reason to chown the top-level directory as a separate command.


tdsmith:
You're right; thanks for the PR!

DomT4:
Previous chit-chat on this for context: https://github.com/Homebrew/homebrew/commit/e2556afc411cf0f6a4c1c0789d8defe51226bb11#commitcomment-13340795

PR #44551, opened 2015-10-03
wine flicker patch

dholm:
Re-add the patch for screen flickering since the issue is unresolved.

PR #44555, opened 2015-10-03
wine: add back vital patch

LinusU:
When wine was upgraded to `1.7.52` someone dropped a patch because it didn't apply cleanly. For some reason that got accepted without requiring any reasoning to why the patch was dropped.

This adds the patch back in and bumps the revision.


LinusU:
Closing this in favour of this one: #44551

PR #44563, opened 2015-10-03
stlink: support --HEAD

erkki:


Library/Formula/stlink.rb
```diff
@@ -4,6 +4,10 @@ class Stlink < Formula
   url "https://github.com/texane/stlink/archive/1.1.0.tar.gz"
   sha256 "3ac4dfcf1da0da40a1b71a8789ff0f1e7d978ea0222158bebd2de916c550682c"
 
+  head do
```
dunn: This can be a single line: `head "https://github.com/texane/stlink.git"` (with double-quotes, please!)

We only need the block style if there are head-specific dependencies or patches.

erkki: Sure thing, pushed, thanks for the quick feedback :)

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/132ba4855744d6f19591d93651a0e98aaa995acb. Thank you for your contribution to Homebrew @erkki; we appreciate it! :smiley_cat:

PR #44565, opened 2015-10-03
libass: 0.13.0

pigoz:
Removed fontconfig since this version introduces CoreText support for
font lookup.

Also made Harfbuzz recommended because it's recommended by the upstream.
Libass can use Harfbuzz for text layouting instead of Freetype, and has
an internal high perfomance rasterizer. Using Harfbuzz fixes many
problems with asian and arabic scripts, as well as providing better
support for opentype features like ligatures, kerning, etc.
Freetype is still needed as a dependency to parse the actual font files.

**Note:** I did not change the bottle SHAs. No idea if I am supposed to when updating a formula.

### DIFF
--- a/Library/Formula/libass.rb
+++ b/Library/Formula/libass.rb
@@ -1,8 +1,8 @@
 class Libass < Formula
   desc "Subtitle renderer for the ASS/SSA subtitle format"
   homepage "https://github.com/libass/libass"
-  url "https://github.com/libass/libass/releases/download/0.12.3/libass-0.12.3.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
> Note: I did not change the bottle SHAs. No idea if I am supposed to when updating a formula.

Can leave those. They get automatically regenerated by our CI.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/61d79b8d929001c6805e420b637dded326aaea33. Thank you for your contribution to Homebrew @pigoz; we appreciate it! :smiley_cat:

PR #44254, opened 2015-09-22
leiningen 2.5.3

brunchboy:
Upgrade Leiningen to the latest 2.5.3. Fixes a regression with respect to GPG integration.

PR #44566, opened 2015-10-03
pmd: Use unzip instead of tar to extract

rwhogg:
The PMD formula previously used tar to extract the zip archive. This works
fine on Macs, because BSD's tar can handle zip archives. GNU tar can't, so we
now use unzip instead.

This fixes https://github.com/Homebrew/linuxbrew/issues/579.

### DIFF
--- a/Library/Formula/pmd.rb
+++ b/Library/Formula/pmd.rb
@@ -7,12 +7,13 @@ class Pmd < Formula
   depends_on "maven" => :build
 
   def install
+    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{ENV["HOME"]}"
     system "mvn", "clean", "package"
 
     doc.install "LICENSE", "NOTICE", "README.md"
 
     # The mvn package target produces a .zip with all the jars needed for PMD
-    safe_system "tar", "-xf", buildpath/"pmd-dist/target/pmd-bin-#{version}.zip"
+    safe_system "unzip", buildpath/"pmd-dist/target/pmd-bin-#{version}.zip"
     libexec.install "pmd-bin-#{version}/bin", "pmd-bin-#{version}/lib"
 
     bin.install_symlink "#{libexec}/bin/run.sh" => "pmd"


tdsmith:
Build is failing because of a sandbox issue; maven wants to write to the user's home directory and doesn't respect $HOME.

The fix is to set `ENV["_JAVA_OPTIONS"] = "-Duser.home=#{ENV["HOME"]}”`, per https://github.com/Homebrew/homebrew/pull/44254#issuecomment-142701938.

I think Homebrew should do this for every install operation; I'll look into fixing up our environment code.

Library/Formula/pmd.rb
```diff
@@ -25,6 +25,7 @@ def caveats; <<-EOS.undent
   end
 
   test do
+    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{ENV["HOME"]}”
```
tdsmith: this would need to happen in the `install` method, not here, but I'd rather not work around this in the formula. Thanks for the update though!

DomT4:
> I think Homebrew should do this for every install operation; I'll look into fixing up our environment code.

It's currently known to break a few modules, which is the hold up.

tdsmith:
Can you elaborate on that at #44567, Dominyk?

Library/Formula/pmd.rb
rwhogg: Sorry. What do you recommend I do then?

a) Work around the problem for now by moving this into the `install` function
b) Remove this line entirely and just let CI be broken for now
c) Maintain my own fork until the problem @DomT4 mentioned is fixed

Sorry, I'm new to Homebrew development (didn't even know there was even a Linux version until last week!) so I'm not sure about the protocol here.

tdsmith: I wasn't sure, sorry. I think `a` is the best answer for now. Sorry to send you back and forth.

rwhogg: No matter @tdsmith . I'll add it back in the `install` function.

And don't worry about sending me back and forth! GitHub literally ignored a pull request I sent for weeks until it became out-of-date, at which point I just closed it. A little back-and-forth is far better than that.

Library/Formula/pmd.rb
```diff
@@ -7,12 +7,13 @@ class Pmd < Formula
   depends_on "maven" => :build
 
   def install
+    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{ENV["HOME"]}”
```
DomT4: This has been closed with the wrong form of `"`. You've used `”` rather than `"`, which upsets Ruby :smile:

rwhogg: Oops, think I might have copy-pasted directly from a Markdown display. My bad!

Will fix.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/68b082bb26aa730779ee6f7e5a2bdf6195fbfcb0. Thank you for your contribution to Homebrew @rwhogg; we appreciate it :smiley_cat:

PR #44567, opened 2015-10-03
set Java user.home for builds

tdsmith:
Java doesn't respect $HOME. Maven attempts to write to the home
directory during builds, so Maven builds will fail unless we set the
user.home property or configure it not to use a local package cache.

This feels kinda gross to me but I think this is the only place to do this because this is the soonest we learn what HOME will be.

I'd also like to do this for the test environment; there's a setup_test_home that I'd like to repurpose for `install` as well and maybe this can be combined with that.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1381,12 +1381,15 @@ class Formula
       mkdir_p env_home
 
       old_home, ENV["HOME"] = ENV["HOME"], env_home
+      old_java_options = ENV["_JAVA_OPTIONS"]
+      ENV.append "_JAVA_OPTIONS", "-Duser.home=#{env_home}"
 
       begin
         yield
       ensure
         @buildpath = nil
         ENV["HOME"] = old_home
+        ENV["_JAVA_OPTIONS"] = old_java_options
       end
     end
   end


DomT4:
Re request for comment, see [this bug report](https://issues.apache.org/jira/browse/MESOS-3482). Some Maven modules break against it, in particular `maven-javadoc-plugin` which seems to be quite a common one. We're hacking around this in [a not entirely pretty way](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mesos.rb#L83-L94) in the `mesos` formula at the moment.

I haven't done much testing beyond that yet because of time & studies & El Cap fixes and such, but I'd be surprised if that was the only breakage point.

Library/Homebrew/formula.rb
```diff
@@ -1381,12 +1381,15 @@ def stage
       mkdir_p env_home
 
       old_home, ENV["HOME"] = ENV["HOME"], env_home
+      old_java_options = ENV["_JAVA_OPTIONS"]
+      ENV.append "_JAVA_OPTIONS", "-Duser.home=#{env_home}"
```
DomT4: Don't know if we want to `prepend` here in case the user has the same local variable set?

tdsmith: I assumed appending would overwrite an earlier definition, which I assume is what we prefer.

DomT4: Wouldn't appending end up passing something like:

```
_JAVA_OPTIONS = -Duser.home=/Users/Dave/JavaIsFun -Duser.home=#{env_home}
```
With `append`? Guess there's an easy enough way to find out.

tdsmith: Yep.

tdsmith:
Thanks for the background; to my eye that's clearly a maven-plugins bug and we should file against maven-plugins if we haven't yet.

tdsmith:
Opened https://issues.apache.org/jira/browse/MJAVADOC-435

Library/Homebrew/formula.rb
DomT4: Mildly amusingly, `prepend` breaks where `append` works. Turns out Java will obey whichever of the same variable is passed last rather than first `¯\_(ツ)_/¯`.

tdsmith:
No progress upstream.

dunn:
Should this be done in `java_requirement.rb` instead? https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements/java_requirement.rb#L16-L24

DomT4:
Not yet, sadly. It doesn't work for every formula, and can actually cause breakage. Needs to be tested each time.

DomT4:
Eventually, possibly. Depends on whether or not the Java requirement is always invoked and available. In some ways I prefer it in `formula.rb` because it's silent and doesn't require users to ever be involved in ensuring it's being called correctly.

We should probably think of an answer though because eventually we may end up with overrides for a bunch of other languages too :smile:.

MikeMcQuaid:
Passing on this as-is until upstream sort their 💩 out.

PR #44568, opened 2015-10-03
qt: bump revision for new OpenSSL linkage

tdsmith:
Per discussion in #44388.

### DIFF
--- a/Library/Formula/qt.rb
+++ b/Library/Formula/qt.rb
@@ -1,6 +1,7 @@
 class Qt < Formula
   desc "Cross-platform application and UI framework"
   homepage "https://www.qt.io/"
+  revision 1
 
   stable do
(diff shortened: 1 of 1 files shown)


tdsmith:
Oof, the test is bad:
```
==> /private/tmp/qt20151003-63754-1ti2kiv/qtnetwork-test 2>&1
Error: qt: failed
invalid byte sequence in US-ASCII
```

tdsmith:
Dunno why codequery is still broken; it works locally for me :/

jkarneges:
It seems unlikely the problem would be in qt, unless the issue can be traced to some kind of openssl loading error (since nothing else was changed). Can the patch still be merged?

tdsmith:
Oh! Yes, sorry, I got halfway through pulling this and then got distracted.

tdsmith:
Revision landed as 18e7a2bb32.

PR #44569, opened 2015-10-04
zsh: add HEAD support

zmwangx:
I needed to test the master branch of Zsh, and turned out that it's a non-trivial task, so I figured I would submit this to make life easier for future testers.

Unfortunately, building man pages on master requires [`yodl`](https://fbb-git.github.io/yodl/), which in turn depends on [`icmake`](https://fbb-git.github.io/icmake/), both of which are not available from either OS X or Homebrew, and a bit nontrivial to build too. Release tarballs bypass this problem by shipping with precompiled `.1` files. Therefore, I've disabled the `install.man` target for `HEAD` and other related targets (`install.info` and `install.runhelp`). Hopefully this isn't too bad.

P.S. Looks like [zchee/homebrew-zsh](https://github.com/zchee/homebrew-zsh) has implemented formulae for `yodl` and `icmake`. I haven't tested them.

### DIFF
--- a/Library/Formula/zsh.rb
+++ b/Library/Formula/zsh.rb
@@ -11,6 +11,11 @@ class Zsh < Formula
     sha256 "932fe97487753363d3ddd683918210367ec29104e700001bbf5cd18c2f4d59fa" => :mavericks
   end
 
+  head do
+    url "git://git.code.sf.net/p/zsh/code"
+    depends_on "autoconf" => :build
+  end
+
   option "without-etcdir", "Disable the reading of Zsh rc files in /etc"
 
   deprecated_option "disable-etcdir" => "without-etcdir"
@@ -19,6 +24,8 @@ class Zsh < Formula
   depends_on "pcre"
 
   def install
+    system "Util/preconfig" if build.head?
+
     args = %W[
       --prefix=#{prefix}
       --enable-fndir=#{share}/zsh/functions
@@ -46,8 +53,14 @@ class Zsh < Formula
     inreplace ["Makefile", "Src/Makefile"],
       "$(libdir)/$(tzsh)/$(VERSION)", "$(libdir)"
 
-    system "make", "install"
-    system "make", "install.info"
+    if build.head?
+      # disable target install.man, because the required yodl comes neither with OS X nor Homebrew
+      # also disable install.runhelp and install.info because they would also fail or have no effect
+      system "make", "install.bin", "install.modules", "install.fns"
+    else
+      system "make", "install"
+      system "make", "install.info"
+    end
   end
 
   def caveats; <<-EOS.undent


Library/Formula/zsh.rb
```diff
@@ -17,8 +18,11 @@ class Zsh < Formula
 
   depends_on "gdbm"
   depends_on "pcre"
+  depends_on "autoconf" => :build if build.head?
```
DomT4: Please use a `head do` block instead here.

```ruby
head do
  url "git://git.code.sf.net/p/zsh/code"

end
```

zmwangx: Okay, so where should I put it? Before `bottle do` or after all other `depends_on`?

DomT4: After `bottle do` but before `depends_on`.

zmwangx: Fixed.

apjanke:
Thanks for the pointer @zmwangx! I've been working on my own icmake and yodl formulas since I want to work with zsh dev on OS X. Going to try his out instead, and/or see if upstream would be willing to port them to BSD.

zmwangx:
I just had a closer look at zchee/homebrew-zsh and it's no good. https://github.com/zchee/homebrew-zsh/blob/master/yodl.rb#L14-L15: hard-coded path and `icmake` in `/usr/bin`??? Anyway, his formulae are quite helpful with understanding the build system.

Personally I wouldn't bother, because (1) it's too weird a build system; (2) I find having slightly outdated man pages fine. I always go to the HTML manual anyway: http://zsh.sourceforge.net/Doc/Release/ (by the way we could add `install.html` to this formula; I didn't bother to change).

apjanke:
I was able to tweak them a little and get them building, and passing `brew audit`. See https://github.com/apjanke/homebrew-zsh/tree/apjanke-work. Haven't tested them with the `zsh` source yet; feel free to do so and send me bug reports if you have time, since my goal is to support that.

I'll send my tweaks along to zchee with a PR.

apjanke:
I went ahead and merged this in as-is, since yodl is unlikely to make it in to Homebrew any time soon.

Thank you for your contribution to Homebrew!

apjanke:
Also, `icmake` and `yodl` are on their way in to Homebrew core now: #45178, #45546. Once they're in we'll be able to revise this to build the docs for HEAD.

zmwangx:
:+1:

zchee:
@zmwangx @apjanke Sorry, I did not manage the PR.
My PR is exist `.dSYM` file. Needs patch.

BTW, correct https://github.com/apjanke/homebrew-zsh/blob/apjanke-work/yodl.rb ?
If true, I will rebase and re-commit.

I respect @apjanke opinion.

PR #44584, opened 2015-10-04
gxemul 0.6.0.1

dholm:


dholm:
I have posted the patch to the gxemul-devel mailing list but since I'm not subscribed its awaiting moderation.

Library/Formula/gxemul.rb
```diff
@@ -0,0 +1,86 @@
+class Gxemul < Formula
+  desc "Framework for full-system computer architecture emulation."
+  homepage "http://gxemul.sourceforge.net"
+  url "http://gxemul.sourceforge.net/src/gxemul-0.6.0.1.tar.gz"
+  sha256 "0790823d8d7f8848c98318ace829f7a0159a5f9b79d80bd8c367fb17014dcda9"
+
+  depends_on :x11
```
DomT4: If this has a hard `x11` dependency it should be filed at https://github.com/Homebrew/homebrew-x11

dholm: Ok, refiled.

PR #44587, opened 2015-10-04
freetype 2.6.1

lembacon:


DomT4:
Patch is failing. Perhaps partly or completely no longer needed?

```
==> Patching
/usr/bin/patch -g 0 -f -p1 -i /private/tmp/freetype--patch20151004-26315-1xgr500/patch
patching file builds/unix/ftconfig.in
Hunk #1 FAILED at 365.
1 out of 1 hunk FAILED -- saving rejects to file builds/unix/ftconfig.in.rej
can't find file to patch at input line 30
```

tschoonj:
Keep in mind that every time `freetype` is updated, a lot of formulas need revision bumps too: the pkg-config files of depending formulas often contain full paths to the include and lib folders of the keg.

Example `fontconfig.pc`:

````
Name: Fontconfig
Description: Font configuration and customization library
Version: 2.11.1
Requires:  freetype2
Requires.private: 
Libs: -L${libdir} -lfontconfig
Libs.private: -lexpat -L/usr/local/Cellar/freetype/2.6_1/lib -lfreetype   
Cflags: -I${includedir}  -I/usr/local/Cellar/freetype/2.6_1/include/freetype2
````

On my installation, the following formulas are affected:

`ImageMagick`, `fontconfig`, `gdlib` but I am sure there are more

**Comment by lembacon** [comment]
@DomT4 The patch has been removed via https://github.com/lembacon/homebrew/commit/1f4edc1ca595ce35c2b135295ca8a576456d57f7. :-)

**Comment by lembacon** [comment]
@tschoonj I have bumped the revisions for all formulae which have freetype as their required or recommended dependecy via https://github.com/lembacon/homebrew/commit/47e0eac519a27f28849618a921ce58951f0111e5.

**Comment by tschoonj** [comment]
This is way too many revision bumps :smile: 

Most pkg-config packages actually include the freetype dependency through its pkg-config name, so those won't need a revision bump at all. Also it's not because a Formula depends on freetype that the dependency will also make it into the pkg-config file it may provide.

I gladly admit that it's quite a lot of work to actually determine the right Formulas for this. 

@DomT4 @mikemcquaid do you have any idea how to do this easily: grep all the pkg-config files of all Formulas that depend on freetype, including those depending recursively, and see which ones contain a full path to the keg? 

I don't want to install them all, it's several hundreds :smile:

**Comment by bfontaine** [comment]
This branch needs rebasing on top of `master`.

**Comment by lembacon** [comment]
@bfontaine Rebased. :smiley_cat:

**Comment by jiegec** [comment]
@tschoonj Maybe we can create a brew-depend command and list all packages depending on freetype. Well that build may exceed time limit...

**Comment by tschoonj** [comment]
That already exists and it's called `brew uses --recursive freetype`

But the formulas that need a revision bump are only a small subset of the output obtained with this command. We would actually need to look into the pkg-config files they (may) install.

**Comment by jiegec** [comment]
@tschoonj Oh i forgot that command. Thanks. Maybe we can tag the fontconfig with :pkg-config-path-depend or else and bottle again when upgrading freetype. This path problem is checked when bottling but just ignored.

**Comment by bfontaine** [comment]

**Comment by apjanke** [comment]
Test in CI is failing: looks like`brew test pstoedit` is being run twice, and its dependency `ghostscript` is being uninstalled before the second time it is tested.

screen shot 2016-01-30 at 11 24 51 pm

**Comment by apjanke** [comment]
Previous version bumps to `freetype` (#35377, #40603, #34825) have not included revision bumps to formulae which have full-path pkgconfig references to it, so those dependent formulae are probably already broken, if their formulae haven't been updated for other reasons since. (The formulae themselves are correct, but their bottles and installations of them done before the `freetype` version bump will have incorrect paths. If I understand this correctly.) For example, `simple-tiles`, when poured from bottle, is currently referencing freetype 2.5.5 and 2.5.3_1. (Its `glib`, `pixman`, `pango`, and `cairo` references are also outdated.)

```
Cflags: -I${includedir} -I/usr/local/Cellar/libpng/1.6.17/include/libpng16 -I/usr/local/Cellar/freetype/2.5.5/include/freetype2 -I/usr/local/Cellar/fontconfig/2.11.1/include -I/usr/local/Cellar/freetype/2.5.3_1/include/freetype2 -I/usr/local/Cellar/pixman/0.32.6/include/pixman-1 -I/usr/local/Cellar/glib/2.44.1/include/glib-2.0 -I/usr/local/Cellar/glib/2.44.1/lib/glib-2.0/include -I/usr/local/opt/gettext/include -I/usr/local/Cellar/cairo/1.14.2/include/cairo -I/usr/local/Cellar/pango/1.36.8/include/pango-1.0 -I/opt/X11/include -I/usr/local/Cellar/gdal/1.11.2_1/include
```

You can detect the hard-path pkgconfig dependencies by going through all the bottles for all the formulae, extracting their `lib/pkgconfig` subdirectory, and grepping them for a versioned `freetype` path reference. [Here is a script](https://gist.github.com/apjanke/3db63e22f794c475a2f5) to do that programmatically.

I ran it and found these dependencies.

```
fontconfig
gd
imagemagick
libbluray
simple-tiles
```

Freetype 2.6.2 is already out. I've made a PR for it, including revision bumps on those dependents. #48666

This detection using bottles is rather slow. Maybe for formulae like this, we should just include a comment with a list of dependent formulae that need to be revision-bumped along with them? And then do detection like this every so often when the landscape changes.

@tschoonj: 
> Also it's not because a Formula depends on freetype that the dependency will also make it into the pkg-config file it may provide.

Does this mean we should actually do this pkgconfig reference detection on all formulae, not just the ones in `brew uses --recursive freetype`?

tschoonj:
No, you did things right. Thanks for writing that script: looks really useful!

DomT4:
Let's close in favour of @apjanke's PR then. Thank you nonetheless for your efforts here @lembacon!

PR #44592, opened 2015-10-04
Nomad 0.1.0 (new formula)

amochtar:


Library/Formula/nomad.rb
```diff
@@ -0,0 +1,54 @@
...
+  homepage "https://www.nomadproject.io"
+  url "https://github.com/hashicorp/nomad.git",
+    :tag => "v0.1.0",
+    :revision => "520763c0715ee88b6571db840e62fab186d7fe59"
+
+  head "https://github.com/hashicorp/nomad.git"
+
+  depends_on "go" => :build
+
+  # explicit install of shirou/gopsutil/cpu to work around error message:
```
dunn: This comment should be above the installation on line 32, right?

amochtar: Yes, that makes sense :)

dunn:
Thank you for contributing! Merged in 0b9278be0f9842fd91f6f7442c3ba90cee4890cd.

PR #44593, opened 2015-10-04
unshield 1.3

bfontaine:
I can’t find a sample `cab` file to test the functionality.

### DIFF
--- a/Library/Formula/unshield.rb
+++ b/Library/Formula/unshield.rb
@@ -1,18 +1,18 @@
 class Unshield < Formula
   desc "Extract files from InstallShield cabinet files"
   homepage "https://github.com/twogood/unshield"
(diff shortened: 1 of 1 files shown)


DomT4:
There's openssl linkage, apparently `¯\_(ツ)_/¯`

bfontaine:
Right, thanks @DomT4 :smile:

DomT4:
Slightly frustratingly the only thing they're using OpenSSL for is to do MD5 hashing. They have an internal function available for this, but apparently you only get to activate that if OpenSSL isn't available along the `$PATH` (Hence the El Cap pass on the last build).

This is the sort of thing in an ideal world people would have moved over to CommonCrypto on OS X for by now :cry:.

PR #44596, opened 2015-10-04
liblinear 2.10

bfontaine:
I’m not sure which version we should use here. Upstream’s website use `2.1` and their Git tag is `210`; The previous version was `2.01` with a `201` tag. We parse both `2.01` and `2.1` as a 2.1 version; I used `2.10` as a compromise between the website version and the tag.

### DIFF
--- a/Library/Formula/liblinear.rb
+++ b/Library/Formula/liblinear.rb
@@ -1,8 +1,9 @@
 class Liblinear < Formula
   desc "Library for large linear classification"
   homepage "https://www.csie.ntu.edu.tw/~cjlin/liblinear/"
-  url "https://www.csie.ntu.edu.tw/~cjlin/liblinear/oldfiles/liblinear-2.01.tar.gz"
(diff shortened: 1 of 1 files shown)


bfontaine:
cc @DomT4 or @xu-cheng for a thought on the version thing.

DomT4:
Was the only issue with the `version` on `2.1` that it didn't mesh with the git tag or were there other issues?

bfontaine:
@DomT4 The previous version was `2.01`, which we parse the same as `2.1`. If we remove the `version` here users won’t get an upgraded formula. Maybe a `revision 1` is the solution here?

DomT4:
@bfontaine Thanks. We really need to look at our version detection at some point :confused:.

`revision 1` seems to get the outdated flag locally so that would work with the new URL and no explicit version, but I'm fine if you want to specify a manual one here as well. Both solutions are about as good as each other.

bfontaine:
Thanks @DomT4. Both are not perfect solutions but I think the explicit `2.10` version might be better for the future so this is what I pulled.

Issue #44018, opened 2015-09-16
uwsgi failed to build on 10.11


apjanke:
https://gist.github.com/958d206bd0c643dbf8fe

```

python
uwsgiconfig.py
--build
brew

core/socket.c:713:28: error: use of undeclared identifier 'SOL_TCP'
                if (setsockopt(serverfd, SOL_TCP, TCP_FASTOPEN, (const void *) &uwsgi.tcp_fast_open, sizeof(int)) < 0) {
                                         ^
1 error generated.

```

brooneysn:
I got this to work in 10.11 by installing the command line tools for 10.10 - Newest version.

bartTC:
Thank you @brooneysn this made it working. I downloaded the latest, stable command line tools from the [developer site](https://developer.apple.com/downloads/).

    "Command Line Tools OS X 10.10 for Xcode 7"	"Sep 16, 2015"

toabi:
Manual "reinstall" of command line tools also fixed this for me.

alexanderfrey:
does not work for me. I'm on capitan, reinstalled xcode-select --install   and still brew install uwsgi does not work. Any recommendations ? Best

brooneysn:
It worked by installing the command line tools:

​

apjanke:
Does anyone know what exactly happens when you install the "Command Line Tools OS X 10.10 for Xcode 7" on a 10.11 system? What are the implications for other builds?

apjanke:
Reported upstream: unbit/uwsgi#1056.

apjanke:
@alexanderfrey: ` xcode-select --install` isn't sufficient, they're talking about putting specifically the 10.10 Command Line Tools on your 10.11 system.

xrmx:
In the meantime a new stable version is the release you can cherry pick b608eb1772641d525bfde268fe9d6d8d0d5efde7 from uwsgi master.

```
commit b608eb1772641d525bfde268fe9d6d8d0d5efde7
Author: Ragnar Rova <ragnar.rova@gmail.com>
Date:   Wed Jul 15 00:17:46 2015 +0200

    OSX 10.11 supports TCP_FASTOPEN
    
    But does not define SOL_TCP.
    IPPROTO_TCP is the equivalent.

diff --git a/core/socket.c b/core/socket.c
index 8f40043..f75bea9 100644
--- a/core/socket.c
+++ b/core/socket.c
[...]

PR #44602, opened 2015-10-05
uwsgi: use --verbose to get detailed build logs

apjanke:
By default, uwsgi's build script suppresses output of individual commands. This uses --verbose to get output more like a normal build process.

Also fixes some `brew audit` complaints.


apjanke:
The test-bot failures are El Cap failures due to known issue #44018, which is fixed in upstream HEAD. I'd like to have this merged anyway, since it makes diagnosing the El Cap and future failures easier. (I had to do it locally to get a good upstream bug report for #44018.)

tdsmith:
Works for me; pulled. Thanks!

PR #44612, opened 2015-10-05
freeimage: patches for gcc 5.0

davydden:


DomT4:
We'd want to make that patch an external patch rather than an inline one. Way too much chance a patch of that size breaks somewhere. Has this been submitted upstream?

davydden:
just submitted it upstream https://sourceforge.net/p/freeimage/patches/110/

davydden:
The author replied that he fixed it, but i guess until the next `freeimage` release we won't have it (i do not see any public git repo for freeimage and i don't know how to get the patch our ot CVS of SourceForge: http://freeimage.cvs.sourceforge.net/viewvc/freeimage/FreeImage/Source/LibWebP/).

davydden:
@DomT4 i would suggest merge this until the next release. If you don't want to have the patch inline I can put it in a gist. Let me know.

DomT4:
Yeah, we'd probably want the patch submitted to https://github.com/Homebrew/patches and then that used instead of the inline one.

davydden:
i will update the link when/if the PR to `patches` is merged...

davydden:
@DomT4 it's good to merge.

Library/Formula/freeimage.rb
```diff
@@ -26,6 +26,13 @@ class Freeimage < Formula
 
   patch :DATA
 
+  # fix GCC 5.0
+  patch do
+    url "https://raw.githubusercontent.com/Homebrew/patches/master/freeimage/gcc5.diff"
```
DomT4: Can you use a commit reference here in case we ever delete the patch from master or something? Thanks.

davydden: you mean this `https://raw.githubusercontent.com/Homebrew/patches/f2b4fb191277d6124450c06a7664e9b24c91ffae/freeimage/gcc5.diff` ?

DomT4: You don't need to use the full SHA1, git will accept 7/8 digits of that just fine. Otherwise :+1:.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/e0db996a891ece32e95a0628b5a6b84099b723bd. Thank you for your contribution to Homebrew @davydden!

PR #44614, opened 2015-10-05
(formula) dockutil: no plistlib needed because it is integrated in python

Fab1n:
dockutil is not installable because the URL in "resource" is not available. plistlib is now integrated in python as far as I know and so it should not be needed as dependency. 
I don't know what happens, if a user uses an older version of OSX or Python. 
How should this be handled within homebrew formulas?


DomT4:
Fixed by https://github.com/Homebrew/homebrew/commit/d93b57e2acd645b67cbe7ae2dfddf73b02494e41. Thanks for the submission though! :heart:

PR #41662, opened 2015-07-13
audit: recommend use of assert_match.

mikemcquaid:
It produces significantly easier output to debug than doing `assert` and `.include?` (which just shows if it passed or failed).

PR #44618, opened 2015-10-05
audit fixes: A 04: apollo - avidemux

apjanke:
audit fixes for: apollo aqbanking artifactory assimp atk atomicparsley avian avidemux

### DIFF
--- a/Library/Formula/apollo.rb
+++ b/Library/Formula/apollo.rb
@@ -5,8 +5,11 @@ class Apollo < Formula
   version "1.7.1"
   sha256 "74577339a1843995a5128d14c68b21fb8f229d80d8ce1341dd3134f250ab689d"
 
-  option "no-bdb", "Install without bdb store support"
-  option "no-mqtt", "Install without MQTT protocol support"
+  deprecated_option "no-bdb" => "without-bdb"
+  deprecated_option "no-mqtt" => "without-mqtt"
+
+  option "without-bdb", "Install without bdb store support"
+  option "without-mqtt", "Install without MQTT protocol support"
 
   # http://www.oracle.com/technetwork/database/berkeleydb/overview/index-093405.html
   resource "bdb-je" do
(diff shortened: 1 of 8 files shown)


DomT4:
Although we'd usually be hesitant to take *just* audit fixes I don't mind them so much at the moment as an opportunity to generate bottles for El Cap or for `build.include` style options. It's preferable to stick to one commit per formula though if possible, in terms of tracking breakage if there turns out to be some.

Merged in https://github.com/Homebrew/homebrew/commit/8a8743b1023f3ad158eb2ef95f9f33a7bf605aa3. Thank you for another contribution to Homebrew @apjanke :smiley_cat:

apjanke:
Oh, sorry! I thought that [from the assert_match change](https://github.com/Homebrew/homebrew/pull/41662#issuecomment-140925692), you were looking to get cleanup PRs. I have... uhh... a few of these queued up, in batches of 5-10. (I'm up to "G".) Would you rather have these submitted / broken up into single-formula audit-only PRs / just set aside until other changes make formula updates urgent / something else... ? No worries if it's "we'd rather defer those changes until later"; I was just doing this to scratch my own "no-warnings" itch, and get practice with git and Ruby.

(I can see now how these multi-formula commits turn into a hassle real fast, esp with the CI/test-bot workload.)

DomT4:
> I have... uhh... a few of these queued up, in batches of 5-10. (I'm up to "G".) 

:smile_cat:. I'm fine with dealing with them at the moment because it lets us deal with El Cap bottles, etc but normally we'd want to handle something that matters more than a small style nit. Things like option renames to the `with/without` syntax are great, as are getting rid of `assert.include?`. Things that matter less are sticking commas at the end of arrays and similar small nits.

The CI situation is er, fluid, but as long as you keep an eye on the build queue I don't have a major problem on that front right at the moment.

apjanke:
Okay. How about I submit what I have done now (while being sympathetic to the CI server load), and then do the rest of the formulae just for deprecated `with/without` and `assert_match` and `404`s, but ignoring code-formatting things like spacing, commas, multi-arg `system` calls, and the like?

PR #44624, opened 2015-10-05
argp-standalone: fetch from svn

heavyk:
trac is down. this fetches from svn similar to 4fe7455


DomT4:
Trac is alive again, so closing. Am working on importing all the trac patches though and storing them in Homebrew's Patches repo. Trac has been down quite a lot recently, sadly.

Thanks for the submission on this though! Appreciated.

heavyk:
then why not put the svn one as a mirror for now?

Issue #44606, opened 2015-10-05
`audit --strict` output divided by topic, not formula


apjanke:
The `brew audit --strict` check includes a call to `brew style` to check formula style too. The style check is done as one big `style` call on all the formulae, before the regular audit checks on each of them. So the style output comes first, other audit results later, and a formula which has both style and regular audit problems will have its results in two places in the output, with other formulae intervening.

I think the results would be more readable if the style check was done for each formula in the main loop, so all the results for a single formula were contiguous in the output. Would you consider a PR to do this? Or is there a reason for breaking all the style results out separately?

#### head + head do  ####

Also, as an aside, should it be okay for a formula to have both a `head` line and a `head do` block? `agda` does this, and the ordering required by `brew audit` results in them being split up. Maybe having both `head` and `head do` should be another audit complaint?

```
  head "https://github.com/agda/agda.git", :branch => "master"

  bottle do
    sha256 "2dc343203159551613fa7346a93d3e2931064026dccab6c30b68a8490845643b" => :el_capitan
    sha256 "b5dfadfde1b45ca3d70f2ac812e98a6d72caf1d840d4cb9c4eee95e81d74ff0b" => :yosemite
    sha256 "4513aecea2fbf5d1a682e04e4cccfd795d48836bcf1b16f79c24fa3bd6482854" => :mavericks
    sha256 "d0e17808ee896cdedc499b010b67961dc3db3c430c501faef16c4b4482238447" => :mountain_lion
  end

  head do
    resource "stdlib" do
      url "https://github.com/agda/agda-stdlib.git", :branch => "master"
    end
  end
```

bfontaine:
> should it be okay for a formula to have both a `head` line and a `head do` block?

No, it should have only a `head do` block.

apjanke:
Want to add a test in `audit` for it?

```
$ brew audit --strict agda
==> brew style agda

1 file inspected, no offenses detected

==> audit problems
agda:
 * `head` (line 20) should be put before `bottle block` (line 13)
 * `head block` (line 34) should be put before `option` (line 22)
 * Should not have both `head` and `head do`

Error: 3 problems in 1 formula
```


https://github.com/apjanke/homebrew/tree/two-heads-are-for-mutants

bfontaine:
Yes that’d be great. Nice branch name :smile:

apjanke:
Okay, does that other branch look ready to turn in to a PR?

bfontaine:
[...]

PR #44626, opened 2015-10-05
audit: add check for both 'head' and 'head do' being present

apjanke:
Follows up discussion in #44606


bfontaine:
@xu-cheng any comment here before I merge this?

xu-cheng:
:+1:

bfontaine:
Thank you @apjanke! :tada:

PR #44605, opened 2015-10-05
audit fixes: A: 01: ack - apache-forrest

apjanke:
audit fixes in: ack, acpica, aeskeyfind, aircrack-ng, alure, amap, apache-forrest

PR #44627, opened 2015-10-05
apache-forrest: add Java dep, audit fixes

apjanke:
Addresses breakage in #44605

### DIFF
--- a/Library/Formula/apache-forrest.rb
+++ b/Library/Formula/apache-forrest.rb
@@ -4,6 +4,8 @@ class ApacheForrest < Formula
   url "https://www.apache.org/dyn/closer.cgi?path=forrest/apache-forrest-0.9-sources.tar.gz"
   sha256 "c6ac758db2eb0d4d91bd1733bbbc2dec4fdb33603895c464bcb47a34490fb64d"
 
+  depends_on :java
+
   resource "deps" do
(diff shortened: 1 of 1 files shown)


Library/Formula/apache-forrest.rb
```diff
@@ -4,6 +4,8 @@ class ApacheForrest < Formula
   url "https://www.apache.org/dyn/closer.cgi?path=forrest/apache-forrest-0.9-sources.tar.gz"
   sha256 "c6ac758db2eb0d4d91bd1733bbbc2dec4fdb33603895c464bcb47a34490fb64d"
 
+  depends_on :java
```
DomT4: Any idea if they want a specific minimum version?

apjanke: Java 1.5; they're pretty basic. Given that (AFAIK) nothing pre-1.6 is available on OS X, do we need to note that?

DomT4: Na. Wouldn't bother for 1.5.

apjanke: All right. I think (imho) this is good to go as is if the brew-test-bot tests pass.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/653663d0a323871b6b54c0a0e1d7f7247a53c71d. Thank you for another contribution @apjanke! :star2:

PR #44629, opened 2015-10-05
Fix incorrect patch hash for pygobject 2.28.6

basilfx:
With current `master`, this Formula will not install due to a changed hash of a patch. I modified the hash, and the Formula compiles and installs without any problems.

### DIFF
--- a/Library/Formula/pygobject.rb
+++ b/Library/Formula/pygobject.rb
@@ -19,8 +19,8 @@ class Pygobject < Formula
 
   # https://bugzilla.gnome.org/show_bug.cgi?id=668522
   patch do
-    url "https://git.gnome.org/browse/pygobject/patch/gio/gio-types.defs?id=42d01f060c5d764baa881d13c103d68897163a49"
-    sha256 "7bb60636a9731afd030820090062688a6b53af22c276d89d6af8db264d76edcc"
+    url "https://raw.githubusercontent.com/Homebrew/patches/master/pygobject/patch-enum-types.diff"
+    sha256 "99a39c730f9af499db88684e2898a588fdae9cd20eef70675a28c2ddb004cb19"
   end
 
   def install


asparagui:
Any clue why the hash changed?

basilfx:
I cannot clarify why this hash changed (the patch looks OK and does not produce any compile errors). Furthermore, the issue hasn't changed recently.

I do compile this under 10.11 (which has no bottled formula yet), so I think that is why this issue appeared.

dunn:
Hashes from cgit seem to change sometimes, so this is probably benign. But maybe we should add this to Homebrew/patches so it doesn't happen again.

basilfx:
I could also add the patch to the Formula. It's a small patch, but I don't know what is the preferred method.

DomT4:
Aye, let's get the patch submitted to https://github.com/Homebrew/patches instead and then use that URL in the `patch do` block.

basilfx:
I've submitted a [patch](https://github.com/Homebrew/patches/pull/2)

bfontaine:
(merged)

basilfx:
PR Updated. I've added the new URL, verified both hashes and tested installation.

Library/Formula/pygobject.rb
```diff
@@ -19,8 +19,9 @@ class Pygobject < Formula
 
   # https://bugzilla.gnome.org/show_bug.cgi?id=668522
   patch do
-    url "https://git.gnome.org/browse/pygobject/patch/gio/gio-types.defs?id=42d01f060c5d764baa881d13c103d68897163a49"
-    sha256 "7bb60636a9731afd030820090062688a6b53af22c276d89d6af8db264d76edcc"
+    url "https://raw.githubusercontent.com/Homebrew/patches/master/pygobject/patch-enum-types.diff"
+    mirror "https://git.gnome.org/browse/pygobject/patch/gio/gio-types.defs?id=42d01f060c5d764baa881d13c103d68897163a49"
```
DomT4: It's probably worth just removing this line at this point to be honest.

bfontaine: :+1:

basilfx:
I removed the mirror URL, so it should be OK now.

bfontaine:
Thank you for your contribution @basilfx; we appreciate it! :tada:

PR #44632, opened 2015-10-05
libass: add option with-fontconfig

zmwangx:
libass 0.13.0 (https://github.com/libass/libass/releases/tag/0.13.0) introduced CoreText support and turned it on by default on OS X. This is good news for some (no more fontconfig caching nightmare, which is especially confusing for users new to fontconfig) but a heavy blow to the rest of us who are heavily invested in fontconfig, and at the time of writing I just don't see the same level of customizability as fontconfig when CoreText support is enabled.

Therefore, this commit attempts to optionally restore the fontconfig legacy for the users concerned by adding a `--without-coretext` option.

### DIFF
--- a/Library/Formula/libass.rb
+++ b/Library/Formula/libass.rb
@@ -11,15 +11,23 @@ class Libass < Formula
     sha256 "7303bf3c119bc80d850f8e16ac1212e70224c83df3f8a3c0bf2adb0d378a6e80" => :mavericks
   end
 
+  option "with-fontconfig", "Disable CoreText backend in favor of the more traditional fontconfig"
+
   depends_on "pkg-config" => :build
   depends_on "yasm" => :build
 
   depends_on "freetype"
   depends_on "fribidi"
   depends_on "harfbuzz" => :recommended
+  depends_on "fontconfig" => :optional
 
   def install
-    system "./configure", "--disable-dependency-tracking", "--prefix=#{prefix}"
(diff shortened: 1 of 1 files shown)


Library/Formula/libass.rb
```diff
@@ -11,15 +11,23 @@ class Libass < Formula
...
 
+  option "without-coretext", "Disable CoreText backend in favor of the more traditional fontconfig"
+
   depends_on "pkg-config" => :build
   depends_on "yasm" => :build
 
   depends_on "freetype"
   depends_on "fribidi"
   depends_on "harfbuzz" => :recommended
+  depends_on "fontconfig" if build.without? "coretext"
```
DomT4: Let's keep the option as `with-fontconfig` and then a similar explanation. We're trying very hard to set fire to `if build.with?` style `depends_on` and I don't really want a new one added. Fontconfig can move to just being `:optional`.

zmwangx:
Changed `without-coretext` to `with-fontconfig` per @DomT4 's request: https://github.com/Homebrew/homebrew/pull/44632#discussion_r41181086.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/bc4ddf70bf946a6cb8ef28f574eba09530da9949. Thank you for your contribution to Homebrew @zmwangx! :smiley_cat:

PR #44640, opened 2015-10-05
libpng: Add command to set up HEAD build config

aw1621107:
libpng requires ./autogen.sh to be run if the source is retrieved from
the git repository because that repo does not include the config files.
The script fails, though, due to it thinking the installed autotools is
incompatible. Turns out running the autoreconf command manually works
fine, so set up formula to run that step

### DIFF
--- a/Library/Formula/libpng.rb
+++ b/Library/Formula/libpng.rb
@@ -28,6 +28,7 @@ class Libpng < Formula
 
   def install
     ENV.universal_binary if build.universal?
+    system "autoreconf", "--force", "--install" if head?
     system "./configure", "--disable-dependency-tracking",
                           "--disable-silent-rules",
                           "--prefix=#{prefix}"


dunn:
What's the incompatibility? Is it something that the developer should fix?

aw1621107:
That's a good question; the error message didn't really help much, even in `--interactive`. Or maybe I missed it; I'll check again. I sort of assumed it's something the script didn't like about superenv, because the entering the command it called manually worked just fine.

Library/Formula/libpng.rb
```diff
@@ -28,6 +28,7 @@ class Libpng < Formula
 
   def install
     ENV.universal_binary if build.universal?
+    system "autoreconf", "--force", "--install" if head?
```
DomT4: You can use `system "autoreconf", "-fvi"` rather than explicitly writing it out. Should be `if build.head?` as well.

aw1621107:
Yeah, this might be on the dev side... Running the script manually with some extra prints shows that the script doesn't see the files it generated. Thing is that every file it says is missing is there, so idk what's going on

Library/Formula/libpng.rb
aw1621107: I normally use the longer names for commands that might not be seen as frequently, but if the short ones are preferred I can make that change.

I wasn't too sure about the head? part because I didn't know where that and/or build was defined. I'll change that

DomT4: The shorter `-fvi` command is used frequently throughout Homebrew where required, so it's nice to stick to the standard there.

aw1621107: Ah, ok. I'll make that change then.

PR #44651, opened 2015-10-05
Encourage https for bottle hosting

scpeters:
I assume that we want to encourage people hosting their own bottles to serve them using `https`. This PR has a trivial change in the example. We may consider adding some text to the `root_url` description with a discussion of security.

### DIFF
--- a/share/doc/homebrew/Bottles.md
+++ b/share/doc/homebrew/Bottles.md
@@ -32,7 +32,7 @@ end
 A full example:
 ```ruby
 bottle do
-  root_url "http://mikemcquaid.com"
+  root_url "https://example.com"
   prefix "/opt/homebrew"
   cellar "/opt/homebrew/Cellar"
   revision 4


DomT4:
We'd probably want to adopt a sample domain that actually supports SSL/TLS. Mike's site doesn't.

scpeters:
I noticed that, but it's not really a functional example anyway since there aren't any bottles hosted there. Also, the link isn't clickable [when I view it on github](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Bottles.md).

I'm happy to change it; those are just the reasons I didn't bother in the first place.

Options:
* https://homebrew.bintray.com doesn't display anything interesting on its homepage
* brew.sh doesn't support https
* https://github.com ?

bfontaine:
You could use https://example.com.

scpeters:
Good call @bfontaine; I changed it to https://example.com and squashed

xu-cheng:

PR #44656, opened 2015-10-06
libzip 1.0.1

tomyun:


DomT4:
Looks like the `dylib` version number might have changed, in which case, revision all the `brew uses libzip`.

tomyun:
Here is what `brew uses libzip` gives:

- [x] `ebook-tools` : `revision` updated
- [x] `ideviceinstaller` : `revision` updated
- [ ] `homebrew/fuse/fuse-zip` : PR for version bump is open (Homebrew/homebrew-fuse#10)
- [ ] `homebrew/games/fceux` : has no real dep on `libzip`. I'll open a PR for fixing this.
- [ ] `homebrew/science/libsigrokdecode` : new version (0.3.1) available. I'll open a PR for version bump.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/70601f0063b00a44aa5bfe8b1f95a850aff2d717. Thanks  for another contribution to Homebrew @tomyun!

Due to the way tap builds work this needed to be merged before tap builds could build against the new `libzip`, so you should be fine to file revision bumps in the taps now.

PR #44667, opened 2015-10-06
recommend python -m pip

tdsmith:
This invocation is more verbose but more deterministic; the `pip` script
can be overwritten when pip is upgraded under python3, which creates
confusion.

This aligns with the intent-to-deprecate discussion for the pip scripts
in e.g. pypa/pip issue 3164.

### DIFF
--- a/Library/Formula/python.rb
+++ b/Library/Formula/python.rb
@@ -315,10 +315,10 @@ class Python < Formula
 
   def caveats; <<-EOS.undent
     Pip and setuptools have been installed. To update them
-      pip install --upgrade pip setuptools
+      python -m pip install --upgrade pip setuptools
 
     You can install Python packages with
-      pip install <package>
+      python -m pip install <package>
 
     They will install into the site-package directory
       #{site_packages}
--- a/Library/Formula/python3.rb
+++ b/Library/Formula/python3.rb
@@ -306,10 +306,10 @@ class Python3 < Formula
   def caveats
     text = <<-EOS.undent
       Pip and setuptools have been installed. To update them
-        pip3 install --upgrade pip setuptools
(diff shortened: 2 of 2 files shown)


jiegec:
+1. pip2 is ok too.

xu-cheng:
Why not recommending `pip2` and `pip3`?

tdsmith:
Recommending `pip2` would be an improvement.

A reason to prefer the more explicit `-m` syntax is to avoid ambiguity in a situation where pip2 has a shebang pointing to a python interpreter which is not the same as the python in PATH.

One way this may occur is if a user installs pip against the system python (`sudo /usr/bin/python -m ensure_pip`) after installing Homebrew python, since system python also writes its scripts to /usr/local/bin.

xu-cheng:
Personally I wouldn't worry about that because users may also put `/usr/bin` ahead of `/usr/local/bin` to cause system Python being picked up. On the other hand, if users choose to install pip on system Python, we may assume they want to use system Python and have the ability to manage system Python and brew Python at the same time?

tdsmith:
The advantage of `python -m pip install foo` (for some package `foo` providing a module `foo`) is that it should always let you run `python -c "import foo"` afterwards. For the case you describe: If system python is first in PATH and Homebrew python has just been installed and someone tries to install something with pip, they will get Homebrew's pip. Unless they've configured their system python to search the /usr/local site-packages, they will not be able to use the module they installed the next time they run `python, which can be confusing.

Users get all kinds of advice of all kinds of quality so I don't know that we can assume that any particular configuration is intentional.

xu-cheng:
:cool: chief of :snake:

tdsmith:
I think seeing this alternative invocation will only confuse people.

PR #44661, opened 2015-10-06
ttfautohint-1.4

lemzwerg:
Regarding #44608, thanks for help!  However, I managed to completely screw up the repository, so I started from scratch to clean up everything, this time doing it in a clean way hopefully :-)

PR #44671, opened 2015-10-06
ttfautohint: 1.4

moyogo:


DomT4:
Beaten to it by #44661 I'm afraid, which has just been merged. Thanks for the submission though! :heart:

moyogo:
I shouldn’t have looked for it :-)
Thanks!

PR #44673, opened 2015-10-06
gobject-introspection: depend on python

rhendric:
... for all older OS X versions that don't ship with Python >=2.7

### DIFF
--- a/Library/Formula/gobject-introspection.rb
+++ b/Library/Formula/gobject-introspection.rb
@@ -16,7 +16,7 @@ class GobjectIntrospection < Formula
   depends_on "glib"
   depends_on "cairo"
   depends_on "libffi"
-  depends_on "python" if MacOS.version == :mavericks
+  depends_on "python" if MacOS.version <= :mavericks
 
   resource "tutorial" do
     url "https://gist.github.com/7a0023656ccfe309337a.git",


DomT4:
Snow Leopard introduced Python 2.7, not Mavericks. I don't think it actually *requires* Python 2.7, but we had to insert the dep on Mavericks because the Mavericks Python had some issues. Unless those issues are also present on ML and below, we don't really want to do this.

I could be smoking the wrong pipe here, so CC @tschoonj @tdsmith

rhendric:
Hmm. Well, I'm on Snow Leopard, my `/usr/bin/python -V` reports `Python 2.6.1`, and `./configure` failed on my machine complaining of a lack of Python >=2.7 until I changed this. Figured that was the issue for Mavericks as well but that was just a guess on my part—I would be okay with a `if MacOS.version == :mavericks || MacOS.version <= :snow_leopard`, if that's better.

tschoonj:
Is there any word on how the latest gobject-introspection compiles/runs on Mountain Lion and Lion?

DomT4:
Hmm. Tim removed the Python dep in https://github.com/Homebrew/homebrew/commit/c6731ddd0d28af9bb38d0ebadd290017b8269c1a. Has it been broken since then, or do the newer versions insert the 2.7 minimum requirement?

rhendric:
I think the latter; I last built gobject-introspection on May 20, after that commit but before 1.46.

BigBlueHat:
This changed worked for me just now on Snow Leopard, fwiw.

:+1:

DomT4:
Would like @tdsmith to sign off on this, but I guess we need to do something here if Python 2.7 is now mandated.

tdsmith:
This is fine with me (though I continue to consider the mavericks failure curious and upsetting!).

PR #44679, opened 2015-10-06
libass: add HEAD

AirPort:


Library/Formula/libass.rb
```diff
@@ -25,6 +33,7 @@ def install
     args = %W[--disable-dependency-tracking --prefix=#{prefix}]
     args << "--disable-coretext" if build.with? "fontconfig"
 
+    system "autoreconf -i" if build.head?
```
bfontaine: Please use `system "autoreconf", "-i"` here.

xu-cheng:

PR #44681, opened 2015-10-06
gstreamer: patch header file issue

rhendric:
Reported upstream as https://bugzilla.gnome.org/show_bug.cgi?id=756136

I tried unpatched gstreamer-1.6.0 with both `--cc=gcc-5` and `--cc=llvm-gcc`, and got the same error both times. Waiting on upstream to respond, but figured it couldn't hurt to propose the PR and get early feedback.

### DIFF
--- a/Library/Formula/gstreamer.rb
+++ b/Library/Formula/gstreamer.rb
@@ -24,6 +24,11 @@ class Gstreamer < Formula
   depends_on "glib"
   depends_on "bison"
 
+  # Fix header file issue (exact OS versions affected unknown; first noticed on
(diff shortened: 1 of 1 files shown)


DomT4:
Apologies for the delay on this; was waiting to see how upstream handled it, but they seem to have merged your patch whole so will pull this.

Merged in https://github.com/Homebrew/homebrew/commit/576d6f97c265450c0c765538f2a713b67d9b97eb. Thank you for your contribution to Homebrew @rhendric; we appreciate it! :smiley_cat:

PR #44682, opened 2015-10-06
truncate, a simple c program that creates empty files of user specified size

artw:
Shame that OSX does not have truncate from BSD, creating files with dd is ineffective. Found this on MacPorts. Took the latest version.

### DIFF
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -473,6 +473,7 @@ module ObserverPathnameExtension
 
     def reset_counts!
       @n = @d = 0
+      @put_verbose_trimmed_warning = false
     end
 
     def total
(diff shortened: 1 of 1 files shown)


Library/Formula/truncate.rb
```diff
@@ -0,0 +1,21 @@
...
+  url "http://www.vanheusden.com/truncate/truncate-0.8.tgz"
+  sha256 "165914d643351324d69096427d3ce5d3ab13aed3f7c990a71ba3bd0c6f278f00"
+
+  def install
+    inreplace "truncate.c" do |s|
+      s.gsub! '#include "error.h"', ""
+      s.gsub! "off64_t", "off_t"
+      s.gsub! "stat64", "stat"
+      s.gsub! "ftruncate64", "ftruncate"
+    end
```
bfontaine: What are these modifications for? Is there a bug to fix? We should use a patch since those don’t seem to be strictly related to Homebrew.

Library/Formula/truncate.rb
```diff
@@ -0,0 +1,21 @@
...
+    system "make"
+    bin.install "truncate"
+    man1.install "truncate.1"
+  end
+  test do
+    system "truncate", "-s", "1k", "testfile"
```
bfontaine: Please use `"#{bin}/truncate"`

Library/Formula/truncate.rb
```diff
@@ -0,0 +1,21 @@
+class Truncate < Formula
+  desc "This program truncates a file to a given size"
```
bfontaine: You could shorten this: “`truncate a file to a given size`”.

Library/Formula/truncate.rb
artw: error.h does not come with the program, but it compiles without it fine
OSX off_t, stat and ftruncate are already natively 64bit and these *64 are either missing or deprecated in 10.11.
I made a patch at first, but it is quite large, since there is a lot of these calls. So I though the less code in formula the better

Library/Formula/truncate.rb
```diff
@@ -0,0 +1,21 @@
...
+  desc "truncates a file to a given size"
+  homepage "http://www.vanheusden.com/truncate"
+  url "http://www.vanheusden.com/truncate/truncate-0.8.tgz"
+  sha256 "165914d643351324d69096427d3ce5d3ab13aed3f7c990a71ba3bd0c6f278f00"
+
+  def install
+    inreplace "truncate.c" do |s|
+      s.gsub! '#include "error.h"', ""    # file is missing from distribution
+      s.gsub! "off64_t", "off_t"          ##
+      s.gsub! "stat64", "stat"            ## These syscalls are 64 bit in OSX
```
bfontaine: Please remove the trailing whitespace at the end of these line.

Library/Formula/truncate.rb
bfontaine: Could we use something like this?

```c
#ifdef __APPLE__
#define off64_t off_t
#define stat64 stat
#define ftruncate64 ftruncate
#endif
```

PR #44684, opened 2015-10-06
open-mpi: Set --HEAD to make release build

aw1621107:
open-mpi from git configs a debug build by default. See HACKING in the
repo root for details

### DIFF
--- a/Library/Formula/open-mpi.rb
+++ b/Library/Formula/open-mpi.rb
@@ -41,6 +41,7 @@ class OpenMpi < Formula
       --enable-ipv6
       --with-libevent=#{Formula["libevent"].opt_prefix}
       --with-sge
+      --with-platform=optimized
     ]
     args << "--disable-mpi-fortran" if build.without? "fortran"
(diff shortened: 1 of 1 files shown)


DomT4:
Is this something we should only be passing when the `head` build is requested or?

aw1621107:
Whoops, you're right. Fixing...

Library/Formula/open-mpi.rb
```diff
@@ -41,6 +41,7 @@ def install
       --enable-ipv6
       --with-libevent=#{Formula["libevent"].opt_prefix}
       --with-sge
+      --with-platform=optimized if build.head?
```
dunn: This is adding `--with-platform-optimized`, `if`, and `build.head?` as three separate arguments.  Instead add this below:
```ruby
args << `--with-platform-optimized` if build.head?

aw1621107: Whoops, didn't realize that. Thanks!

dunn:
Thanks! Merged in 96fe7d1818822f497de21ea5a98ad3c271bcbbb0.

PR #44691, opened 2015-10-07
audit fixes: B 02: bittwist - blockhash

apjanke:
audit fixes for: bittwist blahtexml blazeblogger blockhash


apjanke:
Removed bigloo because it requires dependency changes to fix the audit failures.

xu-cheng:

PR #44695, opened 2015-10-07
(WIP) ddar: fix python shebang location, add test, fix audit

apjanke:
WIP. Please do not merge yet. I'm mainly PRing it so it can go through test-bot and CI.

Compare with other ddar WIP,  #43358

### DIFF
--- a/Library/Formula/ddar.rb
+++ b/Library/Formula/ddar.rb
@@ -8,8 +8,8 @@ class Ddar < Formula
   head "https://github.com/basak/ddar.git"
 
   depends_on "xmltoman" => :build
-  depends_on :python if MacOS.version <= :snow_leopard
-  depends_on "protobuf" => "with-python"
+  depends_on :python
+  depends_on "protobuf"
 
   def install
     system "make", "-f", "Makefile.prep", "pydist"
@@ -20,5 +20,9 @@ class Ddar < Formula
 
     bin.env_script_all_files(libexec+"bin", :PYTHONPATH => ENV["PYTHONPATH"])
     man1.install Dir["*.1"]
+  end
+
+  test do
+    assert_match "ddar: store multiple files efficiently", shell_output("#{bin}/ddar --help")
   end
 end


Library/Formula/ddar.rb
```diff
@@ -21,4 +24,8 @@ def install
     bin.env_script_all_files(libexec+"bin", :PYTHONPATH => ENV["PYTHONPATH"])
     man1.install Dir["*.1"]
   end
+
+  test do
+    assert_match "ddar: store multiple files efficiently", shell_output("ddar --help")
```
bfontaine: It should be `"#{bin}/ddar --help"`.

apjanke: Fixed.

Library/Formula/ddar.rb
```diff
@@ -8,11 +8,14 @@ class Ddar < Formula
   head "https://github.com/basak/ddar.git"
 
   depends_on "xmltoman" => :build
-  depends_on :python if MacOS.version <= :snow_leopard
-  depends_on "protobuf" => "with-python"
+  depends_on :python
```
DomT4: Why are we mandating our Python here now?

tdsmith:
We should install this to libexec and write wrapper scripts to add protobuf's modules to site-packages I guess. If we need to touch a faux __init__.py for the google namespace package while installing protobuf, we can; it should be harmless.

Vendoring the protobuf package sounds potentially nicer, actually.

Sorry, this all kinda sucks.

Distutils should actually already be rewriting the shebang so I don't think that inreplace is doing anything.

Library/Formula/ddar.rb
apjanke: Does `:python` actually mandate the Homebrew one? I thought the system Python would satisfy `:python` on OS X versions where it's installed. If not, then I just made a mistake here.

apjanke: Anyway, the main issue is that `ddar` depends on python modules installed by itself and its dependencies. If we don't reference `:python`, then it gets installed against the system Python, but the modules are in `/usr/local`. So it will break with a "module not found" error when you actually run it. And if you have the Homebrew Python installed, neither the Caveats nor `brew doctor` will give you the warning about having to add a user `homebrew.pth` to pick up those modules, and you'll get the issue jiegec was discussing [here](https://github.com/Homebrew/homebrew/pull/44621#discussion_r41247046).

apjanke:
You're right: if I remove that inreplace, then the shebang becomes `#!/usr/local/opt/python/bin/python2.7`. I'll remove the `inreplace`.

With this patch, all the ddar stuff does get installed to libexec, and a wrapper script to /usr/local/bin. But I don't think that's sufficient in itself.

I don't think we can successfully add protobuf and other brewed modules to site-packages with a wrapper script at run time. Some of those paths are defined using `.pth` files, which are only processed in directories which are in Python's site path based on their location in the filesystem. Adding dirs to `PYTHONPATH` only gets those particular directories on the path, and `.pth` files in them are not followed. E.g. the protobuf modules get on the path via `/usr/local/lib/python2.7/site-packages/homebrew-protobuf.pth`. So, from what I can tell, that `site-packages` dir needs to either be the primary `site-packages` dir for the `python` you're running it under, or included via another `.pth` file in one of the primary `site-packages` dirs.

What does "vendoring the protobuf package" mean?

Library/Formula/ddar.rb
DomT4: Tim might throw me into something if I get this wrong yet again, but:

* `depends_on :python if MacOS.version <= :snow_leopard` = Don't use Homebrew's Python unless building for Snow Leopard or below.
* `depends_on :python` = Use the first Python in the `$PATH` for `--build-from-source` but bottles will be built against Homebrew's Python.
* Lack of `depends_on :python` but still uses Python = Always use the system. Software is compatible with Python 2.6 on old platforms.

Due to our `default_formula` mechanism, `depends_on :python` builds the bottle against Homebrew's Python and consequently forces the installation of our Python when the bottle is used.

apjanke: One other aspect I'd like to discuss here: `ddar` is not a library for Python developers. It is a command line utility which happens to be implemented in Python. More like `tar` or `gzip`, and less like `pyqt` or `protobuf`'s Python bindings. Its use of python is an implementation detail.

Maybe Homebrew's policy about these things – CLI utilities with Python internals – should be different from libraries with Python bindings or dependencies. Maybe we should be more aggressive about doing what's needed to make it "just work" without the user having to make Python-specific configuration changes to their account (which could have side effects on other programs). Depending on brewed Python, where we can guarantee `/usr/lib/.../site-python` inclusion of its deps, would be one way to do that, though it feels pretty heavyweight. As is, the other option is giving the user a somewhat-obscure warning about Python modules and paths... if they don't already know Python, that seems like a burden.

I don't really know Python and I'm a Homebrew newbie, so I'm definitely open to advice and other views here. But I think `ddar` might be a case where we should try to make it "just work" instead of be efficient.

apjanke: And, apologies if I'm making this difficult. This python stuff is complicated, and I clearly don't understand it or the Homebrew-Python interaction as well as folks who've been here longer.

DomT4: Tim will be able to talk more about the upsides/downsides (if he has time & desire to), but Python really isn't a small dependency and Homebrew was built on the principle of using the system tools available. That's developed over time to insert caveats about using system tools available *providing* they are secure and not chronically neglected (see deprecating Apple's OpenSSL), but I'm not sure Python falls under that bracket.

Tim's the Python guru though; I trust his thoughts on these various discussions.

apjanke:
Here's an idea: would it be possible to wrap the program in a *python* wrapper script instead of a shell script, which checked the paths and called `site.addsitedir` to add the homebrew site-packages path if necessary, before running the main program in the same `python` instance? That way we maybe could add the `site-packages` directory at run time in a way that respected the `.pth` files in it.

PR #44696, opened 2015-10-07
bigloo: use brewed openssl, audit fixes

apjanke:


Library/Formula/bigloo.rb
```diff
@@ -45,4 +48,8 @@ def install
     manpages = %w[bgldepend bglmake bglpp bgltags bglafile bgljfile bglmco bglprof]
     manpages.each { |m| man1.install "manuals/#{m}.man" => "#{m}.1" }
   end
+
+  test do
+    assert_match "Bigloo (4", shell_output("bigloo --help")
```
xu-cheng: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

apjanke: Sure, but it may take me a couple days to find time to do so. Want to wait on merging this until I can add it?

DomT4:
It looks like there's an update available here:

> The source file distribution that can be used for Unix systems and Win32/Cygwin is:
> bigloo4.2a.tar.gz

apjanke:
The 4.2a distribution is failing to compile for me on 10.9.5; may require formula tweaks.

DomT4:
Will take this as-is to resolve the OpenSSL situation, but if you wanted to chase it in at some point with an improved test and an attempt at the version bump we certainly wouldn't object :wink:.

Merged in https://github.com/Homebrew/homebrew/commit/89ee43edaa63d3b656e83828f3a3b6beab3a1492. Thanks again @apjanke!

PR #44697, opened 2015-10-07
libtomcrypt: add head

apjanke:
The build is failing under 10.10+ (#43906), and seems to be somewhat fixed in upstream head. The `--HEAD` build still fails when done under Homebrew, but having this in the formula should help test it out and maybe give feedback to upstream before the next release.


apjanke:
Build failures here are expected and known (#43906); adding HEAD just makes it easier to test the upcoming versions.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/2f0017a4160d4742685b3dbaf6822f60dcb240f0. Thank you fellow maintainer @apjanke! :wink:

PR #44700, opened 2015-10-07
add formula 'racer'

baishuai:
racer is a Rust Code Completion utility. It is widely used in rust programmer.

### DIFF
--- a/Library/Formula/racer.rb
+++ b/Library/Formula/racer.rb
@@ -7,8 +7,6 @@ class Racer < Formula
   head "https://github.com/phildawes/racer.git"
 
   option "without-src", "Do not get rust source for racer from www.rust-lang.org"
-  option "with-beta-src", "Get beta version of rust source"
-  option "with-nightly-src", "Get nightly version of rust source"
   depends_on "rust" => :build
 
   resource "rustsrc" do
@@ -16,28 +14,12 @@ class Racer < Formula
     sha256 "ea02d7bc9e7de5b8be3fe6b37ea9b2bd823f9a532c8e4c47d02f37f24ffa3126"
   end
 
-  resource "rustsrc-beta" do
-    url "https://static.rust-lang.org/dist/rustc-beta-src.tar.gz"
-    sha256 "e393758a541b1e1859c17deb54a548eb076ab9a7901210ddb47b033aa2d16e79"
-  end
-
-  resource "rustsrc-nightly" do
-    url "https://static.rust-lang.org/dist/rustc-nightly-src.tar.gz"
-    sha256 "be004a627cff79e7eb3ef17b2b855d6a5ee2258246c56cd6812b900ddf5ee79c"
-  end
-
   def install
     if build.with?("src")
-      rustsource = resource("rustsrc")
-      if build.with?("beta-src")
-        rustsource = resource("rustsrc-beta")
-      elsif build.with?("nightly-src")
-        rustsource = resource("rustsrc-nightly")
-      end
-      rustsource.stage do
+      resource("rustsrc").stage do
         rm_rf "src/llvm"
         rm_rf "src/test"
-        (share/"rustsrc").install Dir["./src/*"]
+        (share/"rust-#{resource("rustsrc").version}").install Dir["./src/*"]
       end
     end
 
@@ -47,16 +29,9 @@ class Racer < Formula
 
   def caveats
     if build.with?("src")
-      rustsource = resource("rustsrc")
-      if build.with?("beta-src")
-        rustsource = resource("rustsrc-beta")
-      elsif build.with?("nightly-src")
-        rustsource = resource("rustsrc-nightly")
-      end
       <<-EOS.undent
       racer in installed at #{opt_prefix}/libexec/racer.
-      Rust source with version:#{rustsource.version} is installed at #{opt_prefix}/share/rustsrc.
-      You should export RUST_SRC_PATH=#{opt_prefix}/share/rustsrc.
+      Rust source (version:#{resource("rustsrc").version}) is installed at #{opt_prefix}/share/rust-#{resource("rustsrc").version}
       EOS
     else
       <<-EOS.undent
@@ -66,7 +41,7 @@ class Racer < Formula
   end
 
   test do
-    ENV["RUST_SRC_PATH"] = "#{opt_prefix}/share/rustsrc"
+    ENV["RUST_SRC_PATH"] = "#{opt_prefix}/share/rust-#{resource("rustsrc").version}"
     assert_match /^MATCH BufReader/, shell_output("#{libexec}/racer complete std::io::B")
   end
 end
--- a/Library/Formula/racer.rb
+++ b/Library/Formula/racer.rb
@@ -1,0 +1,45 @@
+class Racer < Formula
+  desc "Rust Code Completion utility"
+  homepage "https://github.com/phildawes/racer"
+  url "https://github.com/phildawes/racer/archive/v1.0.0.tar.gz"
+  sha256 "78895296ed688eeccbaf7745235f0fc503407bfa718f53583a4dcc9e1246b7f5"
+
+  keg_only "You should set path in your special editor which use racer"
+
+  option "without-src", "Do not get rust source for racer from www.rust-lang.org"
+
+  depends_on "rust" => :build
+
+  resource "rustsrc" do
+    url "https://static.rust-lang.org/dist/rustc-1.3.0-src.tar.gz"
+    sha256 "ea02d7bc9e7de5b8be3fe6b37ea9b2bd823f9a532c8e4c47d02f37f24ffa3126"
+  end
+
+  def install
+    if build.with?("src")
+      resource("rustsrc").stage do
+        rm_rf "./src/llvm/test/"
+        (prefix/"rust-src").install Dir["./src/*"]
+      end
+    end
+
+    system "cargo", "build", "--release"
+    bin.install "target/release/racer"
+  end
+
+  def caveats; <<-EOS.undent
+    Rust source is installed at #{opt_prefix}/rust-src by default.
+    You should export RUST_SRC_PATH=#{opt_prefix}/rust-src or set
+    rust source path in your special editor which use racer
+    EOS
+  end
+
+  test do
+    if build.with?("src")
+      ENV["RUST_SRC_PATH"] = "#{opt_prefix}/rust-src"
+      assert_match /^MATCH BufReader/, shell_output("#{bin}/racer complete std::io::B")
+    else
+      assert_match /^RUST_SRC_PATH/, shell_output("#{bin}/racer complete std::io::B", 1)
+    end
+  end
+end
--- a/Library/Formula/racer.rb
+++ b/Library/Formula/racer.rb
@@ -1,10 +1,15 @@
 class Racer < Formula
   desc "Rust Code Completion utility"
   homepage "https://github.com/phildawes/racer"
-  url "https://github.com/phildawes/racer/archive/v1.0.0.tar.gz"
-  sha256 "78895296ed688eeccbaf7745235f0fc503407bfa718f53583a4dcc9e1246b7f5"
 
-  keg_only "You should set path in your special editor which use racer"
+  stable do
+    url "https://github.com/phildawes/racer/archive/v1.0.0.tar.gz"
+    sha256 "78895296ed688eeccbaf7745235f0fc503407bfa718f53583a4dcc9e1246b7f5"
+  end
+
+  head do
+    url "https://github.com/phildawes/racer.git"
+  end
 
   option "without-src", "Do not get rust source for racer from www.rust-lang.org"
 
@@ -18,24 +23,31 @@ class Racer < Formula
   def install
     if build.with?("src")
       resource("rustsrc").stage do
-        rm_rf "./src/llvm/test/"
-        (prefix/"rust-src").install Dir["./src/*"]
+        Dir.glob(["src/llvm/*", "src/test/*"]) do |f|
+          if File.directory?(f)
+            rm_r(f)
+          else
+            File.delete(f)
+          end
+        end
+        (share/"rustsrc").install Dir["./src/*"]
       end
     end
 
     system "cargo", "build", "--release"
-    bin.install "target/release/racer"
+    libexec.install "target/release/racer"
   end
 
   def caveats; <<-EOS.undent
-    Rust source is installed at #{opt_prefix}/rust-src by default.
-    You should export RUST_SRC_PATH=#{opt_prefix}/rust-src or set
+    racer in installed at #{opt_prefix}/libexec/racer, set it to path if needed.
+    Rust source is installed at #{opt_prefix}/share/rustsrc by default.
+    You should export RUST_SRC_PATH=#{opt_prefix}/share/rustsrc or set
     rust source path in your special editor which use racer
     EOS
   end
 
   test do
-    ENV["RUST_SRC_PATH"] = "#{opt_prefix}/rust-src"
-    assert_match /^MATCH BufReader/, shell_output("#{bin}/racer complete std::io::B")
+    ENV["RUST_SRC_PATH"] = "#{opt_prefix}/share/rustsrc"
+    assert_match /^MATCH BufReader/, shell_output("#{libexec}/racer complete std::io::B")
   end
 end
--- a/Library/Formula/racer.rb
+++ b/Library/Formula/racer.rb
@@ -7,7 +7,8 @@ class Racer < Formula
   head "https://github.com/phildawes/racer.git"
 
   option "without-src", "Do not get rust source for racer from www.rust-lang.org"
-
+  option "with-beta-src", "Get beta version of rust source"
+  option "with-nightly-src", "Get nightly version of rust source"
   depends_on "rust" => :build
 
   resource "rustsrc" do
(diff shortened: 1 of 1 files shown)


Library/Formula/racer.rb
```diff
@@ -0,0 +1,20 @@
...
+  depends_on "rust" => :build
+
+  def install
+    system "cargo", "build", "--release"
+    bin.mkpath
+    bin.install "target/release/racer"
+  end
+
+  test do
+    assert_match "RUST_SRC_PATH", `#{bin}/racer complete std::io::B`
```
bfontaine: Use `shell_output("#{bin}/racer complete std::io::B")` instead of `` `#{bin}/racer complete std::io::B` ``.

Library/Formula/racer.rb
```diff
@@ -0,0 +1,20 @@
...
+  url "https://github.com/phildawes/racer/archive/v1.0.0.tar.gz"
+  sha256 "78895296ed688eeccbaf7745235f0fc503407bfa718f53583a4dcc9e1246b7f5"
+
+  keg_only "You should set path in your special editor which use racer"
+
+  depends_on "rust" => :build
+
+  def install
+    system "cargo", "build", "--release"
+    bin.mkpath
```
bfontaine: You can remove this line, `bin` will be created by `bin.install`.

Library/Formula/racer.rb
```diff
@@ -0,0 +1,44 @@
...
+      resource("rustsrc").stage do
+        (prefix/"rustsrc").install Dir["./src/*"]
+      end
+    end
+  end
+
+  def caveats; <<-EOS.undent
+    Rust source is installed at /usr/local/opt/racer/rustsrc/src by default.
+    You should export RUST_SRC_PATH=/usr/local/opt/racer/rustsrc/src or set
+    rust cource path in your special editor which use racer
```
bfontaine: Typo: `cource path` -> `source path`

Library/Formula/racer.rb
```diff
@@ -0,0 +1,44 @@
...
+    EOS
+  end
+
+  test do
+    if build.with?("src")
+      ENV["RUST_SRC_PATH"] = "#{rustsrc}/src"
+      assert_match /^MATCH BufReader/, shell_output("#{bin}/racer complete std::io::B")
+    else
+      assert true
```
bfontaine: Please add a real test here :smile:

Library/Formula/racer.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    if build.with?("src")
+      resource("rustsrc").stage do
+        (prefix/"rustsrc").install Dir["./src/*"]
+      end
+    end
+  end
+
+  def caveats; <<-EOS.undent
+    Rust source is installed at /usr/local/opt/racer/rustsrc/src by default.
```
DomT4: Homebrew is not always in `/usr/local`. This should be `"#{opt_prefix}/racer/rustsrc/src"`

Library/Formula/racer.rb
baishuai: In this branch, #{bin}/racer return 1 but not 0. So shell_output will fail.
Is there any other method I can use? Or maybe shuold use 
`` `#{bin}/racer complete std::io::B` `` instead?

Library/Formula/racer.rb
baishuai: Get it. Thanks

Library/Formula/racer.rb
DomT4:
```ruby
assert_match /something/, shell_output("#{bin}/racer complete std::io::", 1)
```

baishuai:
@bfontaine Hi, the Travis CI build failed because of 

```
 Updating registry `https://github.com/rust-lang/crates.io-index`
 Downloading toml v0.1.21
Unable to get packages from source
```

What can I do to fix it? Thanks.

bfontaine:
@baishuai Don’t worry about the Travis build here; only the “default” one matters.

Library/Formula/racer.rb
```diff
@@ -0,0 +1,45 @@
+class Racer < Formula
+  desc "Rust Code Completion utility"
+  homepage "https://github.com/phildawes/racer"
+  url "https://github.com/phildawes/racer/archive/v1.0.0.tar.gz"
+  sha256 "78895296ed688eeccbaf7745235f0fc503407bfa718f53583a4dcc9e1246b7f5"
+
+  keg_only "You should set path in your special editor which use racer"
+
+  option "without-src", "Do not get rust source for racer from www.rust-lang.org"
```
bfontaine: How common is this need?

Library/Formula/racer.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def caveats; <<-EOS.undent
+    Rust source is installed at #{opt_prefix}/rust-src by default.
+    You should export RUST_SRC_PATH=#{opt_prefix}/rust-src or set
+    rust source path in your special editor which use racer
+    EOS
+  end
+
+  test do
+    if build.with?("src")
```
DomT4: We only test default options, so if `src` is the default option you don't need the non-`src` test.

Library/Formula/racer.rb
baishuai: Racer need the rust source code to work properly. It will search the rust source code and generate the code completion. But `rust` formula don't offer the rust source code. Users must download rust source code to use `racer`. 
So, it is necessary to add rust source code here.

bfontaine: Right, I thought having the `rust` formula installed would be sufficient but it’s not.

Library/Formula/racer.rb
```diff
@@ -0,0 +1,41 @@
+class Racer < Formula
+  desc "Rust Code Completion utility"
+  homepage "https://github.com/phildawes/racer"
+  url "https://github.com/phildawes/racer/archive/v1.0.0.tar.gz"
+  sha256 "78895296ed688eeccbaf7745235f0fc503407bfa718f53583a4dcc9e1246b7f5"
+
```
alkis: How about adding:

```
head "https://github.com/phildawes/racer.git"
```

here? 1.0.0 is 100+ commits behind head and head is the recommended way of installing it.

bfontaine: :+1:

CamJN:
```
class Racer < Formula
  desc "Rust Code Completion utility"
  homepage "https://github.com/phildawes/racer"
  url "https://github.com/phildawes/racer/archive/v1.0.0.tar.gz"
  sha256 "78895296ed688eeccbaf7745235f0fc503407bfa718f53583a4dcc9e1246b7f5"

  depends_on "rust"

  def install
    system "cargo", "build", "--release"
    bin.mkpath
    bin.install "./target/release/racer"
    (share/"rust_src").mkpath
    resource("rust_source").stage do
      Dir.glob(["src/llvm/*", "src/test/*", "src/librustdoc/*", "src/etc/snapshot.pyc"]) do |f|
        if File.directory?(f)
          rm_r(f)
        else
          File.delete(f)
        end
      end
      (share/"rust_src").install "src"
    end
  end

  resource "rust_source" do
    url "https://static.rust-lang.org/dist/rustc-1.3.0-src.tar.gz"
    sha256 "ea02d7bc9e7de5b8be3fe6b37ea9b2bd823f9a532c8e4c47d02f37f24ffa3126"
  end

  def caveats
    "You must set RUST_SRC_PATH=#{HOMEBREW_PREFIX}/share/rust_src/src/ in your ENV."
  end

  test do
    ENV["RUST_SRC_PATH"] = (share/"rust_src/src")
    system "#{bin}/racer", "complete", "std::io::B"
  end
end
```

Library/Formula/racer.rb
```diff
@@ -0,0 +1,41 @@
...
+  keg_only "You should set path in your special editor which use racer"
```
bfontaine: Since `racer` is never called directly it’s probably better to install it in `opt_prefix` instead of `bin` and remove the `keg_only` clause.

DomT4: :+1:

baishuai: Does installing in `opt_prefix` not link the executable file to `/usr/local/bin`?

bfontaine: It’s described here: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#variables-for-directory-locations

No it doesn’t link into `/usr/local/bin`.

baishuai: When installing `racer` to `opt_prefix` and remove `key_only`, I got error
```sh
➜  Desktop  brew install racer
==> Downloading https://github.com/phildawes/racer/archive/v1.0.0.tar.gz
Already downloaded: /Library/Caches/Homebrew/racer-1.0.0.tar.gz
==> Downloading https://static.rust-lang.org/dist/rustc-1.3.0-src.tar.gz
Already downloaded: /Library/Caches/Homebrew/racer--rustsrc-1.3.0.tar.gz
==> cargo build --release
Error: An unexpected error occurred during the `brew link` step
Directory not empty - /usr/local/opt/racer
Error: Directory not empty - /usr/local/opt/racer
```
Installing to `opt_prefix` create directory  `/usr/local/opt/racer` and install `racer` to it. And finally when link `/usr/local/Cellar/racer/1.0.0` to `opt_prefix`, it would be wrong because directory `/usr/local/opt/racer` not empty

DomT4: Use `prefix.install "target/release/racer"` rather than `opt.prefix` install.

alkis: Wouldn't ```libexec``` be more appropriate than ```opt_prefix```? This is a binary that is not called directly but by other programs (editors): http://www.linuxbase.org/betaspecs/fhs/fhs/ch04s07.html

bfontaine: @DomT4 Wouldn’t installing under `prefix` cause issues on upgrade with editors that rely on the binary path?

DomT4: `prefix` == `opt_prefix`, eventually. Once linked after install `/usr/local/opt/racer/racer` will really be located at `/usr/local/Cellar/racer/someversion/racer`. `opt_prefix` is just a symlink.

The trick would be getting text editors (or people) to set `/usr/local/opt/racer/racer` as the expected tool so it doesn't break across updates.

> Wouldn't libexec be more appropriate than opt_prefix?

I don't really mind either way. If it works and you prefer `libexec`, feel free. At that point `/usr/local/opt/racer/libexec/racer` will => `/usr/local/Cellar/racer/someversion/libexec/racer`.

Library/Formula/racer.rb
```diff
@@ -0,0 +1,43 @@
...
+        (share/"rustsrc").install Dir["./src/*"]
+      end
+    end
+
+    system "cargo", "build", "--release"
+    libexec.install "target/release/racer"
+  end
+
+  def caveats; <<-EOS.undent
+    racer in installed at #{opt_prefix}/libexec/racer, set it to your editor if needed.
```
alkis: Let's remove everything after the comma.

Library/Formula/racer.rb
```diff
@@ -0,0 +1,43 @@
...
+    Rust source is installed at #{opt_prefix}/share/rustsrc by default.
```
alkis: Can you specify the version here? Perhaps using ```#{resource(rustsrc).version}``` would do it? I would word it like this:

```
You should set RUST_SRC_PATH to the Rust source code. Rust #{resource(rustsrc).version} is installed at #{opt_prefix}/share/rustsrc for your convenience. If completion for another version of Rust is required, it can be installed in any directory and RUST_SRC_PATH should be set accordingly.
```

bfontaine: @alkis Why do you want the version? It’ll need to be hardcoded; `resource`s don’t have versions.

alkis: @bfontaine I would like the version to be shown so that when rust-lang releases a new version people know that the sources are no longer valid and they need to point somewhere else.

BTW I thought that resource has version: http://www.rubydoc.info/github/Homebrew/homebrew/master/Resource#version-instance_method. Maybe I am reading the docs incorrectly?

bfontaine: My bad; resources _do_ have versions.

Library/Formula/racer.rb
```diff
@@ -0,0 +1,72 @@
...
+    sha256 "e393758a541b1e1859c17deb54a548eb076ab9a7901210ddb47b033aa2d16e79"
+  end
+
+  resource "rustsrc-nightly" do
+    url "https://static.rust-lang.org/dist/rustc-nightly-src.tar.gz"
+    sha256 "be004a627cff79e7eb3ef17b2b855d6a5ee2258246c56cd6812b900ddf5ee79c"
+  end
+
+  def install
+    if build.with?("src")
```
alkis: If you want to support different versions this is not the right way to do it. The same formula (racer) should work for most people. Let's see what options we have:

* stable: this is a must, we expect most people to use this
* best: some people use beta and in particular alongside with stable, so we must support both at the same time. It is being cut every week so you would want to update the formula quite often to make this work
* nightly: a lot of devs use nightly. But nightly is pushed every day so unless this formula is updated every day the sha won't match so the resource will be useless

Given the above my suggestion would be to support stable and at most beta. You should make them two different options so that people can install none, one, or both of them.

alkis: Actually come to think of it more I think we should support only released versions: 1.0, 1.1, etc.

Library/Formula/racer.rb
```diff
@@ -0,0 +1,72 @@
...
+      rustsource = resource("rustsrc")
+      if build.with?("beta-src")
+        rustsource = resource("rustsrc-beta")
+      elsif build.with?("nightly-src")
+        rustsource = resource("rustsrc-nightly")
+      end
+      rustsource.stage do
+        rm_rf "src/llvm"
+        rm_rf "src/test"
+        (share/"rustsrc").install Dir["./src/*"]
```
alkis: FWIW sources should install into {opt_prefix}/src/rust-{version}

Library/Formula/racer.rb
baishuai: This commit is really extreme. I'm agree with yours.

Library/Formula/racer.rb
```diff
@@ -0,0 +1,47 @@
...
+  homepage "https://github.com/phildawes/racer"
+
+  url "https://github.com/phildawes/racer/archive/v1.0.0.tar.gz"
+  sha256 "78895296ed688eeccbaf7745235f0fc503407bfa718f53583a4dcc9e1246b7f5"
+  head "https://github.com/phildawes/racer.git"
+
+  option "without-src", "Do not get rust source for racer from www.rust-lang.org"
+  depends_on "rust" => :build
+
+  resource "rustsrc" do
```
alkis: What do you think about naming this rust-1.3? This is so that you can have rust-1.4 when it is (soon) released.

If you change this to rust-1.3, it would make more sense to install the source into ```#{opt_prefix}/src/rust-1.3```.

baishuai: You means when 1.4 released，this formula should have both 1.3 and 1.4 source code？

baishuai: @alkis You means when 1.4 released，this formula should have both 1.3 and 1.4 source code？

alkis: Yes. Each release should be behind a build option:

with-1.3
with-1.4

etc

alkis:
Why was this closed?

baishuai:
@alkis Sorry, I made a mistake about git.

bfontaine:

alkis:
Now that cargo install is available (cargo is rust's build tool) this formula is much less important. Since racer can be installed through that, I think a formula for rust-src would be ideal here, since racer needs the source to give completions. What do you think?

mikemcquaid:
@alkis I think it's still work including this formulae as it installs binaries and doesn't require Rust or knowledge of another build tool.

Library/Formula/racer.rb
```diff
@@ -0,0 +1,79 @@
...
+  homepage "https://github.com/phildawes/racer"
+
+  url "https://github.com/phildawes/racer/archive/v1.0.0.tar.gz"
+  sha256 "78895296ed688eeccbaf7745235f0fc503407bfa718f53583a4dcc9e1246b7f5"
+  head "https://github.com/phildawes/racer.git"
+
+  option "with-1.4", "Get 1.4 version of rust source"
+  option "with-1.3", "Get 1.3 version of rust source"
+  option "with-1.2", "Get 1.2 version of rust source"
+  option "without-rust", "Disable any version of rust source"
```
mikemcquaid: How would I know as a user which of these I should pick?

Library/Formula/racer.rb
```diff
@@ -0,0 +1,79 @@
...
+    if build.with?("rust")
+      <<-EOS.undent
+      racer in installed at #{opt_prefix}/libexec/racer.
+      Rust source is installed at #{opt_prefix}/share/rust-{version}.
+      EOS
+    else
+      <<-EOS.undent
+      racer in installed at #{opt_prefix}/libexec/racer.
+      EOS
+    end
```
mikemcquaid: Doesn't need to be an `if`/`else`; instead use a string variable and add to it.

baishuai:
Since rust 1.5, we can use `cargo install` to install a rust binary. so I think this formula is not necessary any more.
As for the rust source file needed by racer, maybe it should be a install option in rust formula.

PR #44701, opened 2015-10-07
glm 0.9.7.1 + cmake targets config

cHoco:
Fixes regression introduced by https://github.com/Homebrew/homebrew/pull/44076 that was caused by the new build scripts of glm.


xu-cheng:

PR #44709, opened 2015-10-07
add Formula#aliases

### LINKED ISSUE #44331: `brew info --json=v1 --installed` should also return all aliases.
Will make fixing e.g. https://github.com/Homebrew/homebrew-bundle/issues/105 much easier.

xu-cheng:
Fixes #44331.


mikemcquaid:
Looks great :+1:

PR #44713, opened 2015-10-07
uwsgi: use secure url

vszakats:


bfontaine:
Thank you @vszakats! :smile:

PR #44721, opened 2015-10-07
Reimplement #pretty_duration

koraktor:
Also reuse this improved implementation in the GitHub rate limit
errors.

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -138,9 +138,18 @@ def pretty_uninstalled(f)
 end
 
 def pretty_duration(s)
-  return "2 seconds" if s < 3 # avoids the plural problem ;)
-  return "#{s.to_i} seconds" if s < 120
-  "%.1f minutes" % (s/60.0)
+  s = s.to_i
+  res = ""
+
+  if s > 59
+    m = s / 60
+    s %= 60
+    res = "#{m} minute#{plural m}"
+    return res if s == 0
+    res << " "
+  end
+
+  res + "#{s} second#{plural s}"
 end
 
 def plural(n, s = "s")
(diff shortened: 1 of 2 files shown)


Library/Homebrew/utils.rb
```diff
@@ -101,9 +101,18 @@ def odie(error)
...
 def pretty_duration(s)
-  return "2 seconds" if s < 3 # avoids the plural problem ;)
-  return "#{s.to_i} seconds" if s < 120
-  "%.1f minutes" % (s/60)
+  s = s.to_i
+  res = ""
+
+  if s > 59
+    m = s / 60
+    s = s % 60
```
bfontaine: You can use `s %= 60` here.

Library/Homebrew/utils.rb
```diff
@@ -101,9 +101,18 @@ def odie(error)
...
+    res = "%d minute%s" % [ m, plural(m) ]
```
bfontaine: Please use `"#{m} minute#{plural(m)}"` instead of the Python-like formatting.

Library/Homebrew/utils.rb
koraktor: Thanks for the hint.

Library/Homebrew/utils.rb
koraktor: Is there a reason for this style decision? String formatting isn't mentioned in GitHub's style guide and IMHO interpolation is less readable in this case.

bfontaine: We use String interpolation everywhere so it’s just a matter of consistency. Also, `rubocop` flag this: `Favor format over String#%`.

koraktor: Thanks for the clarification. I updated my patch accordingly.

bfontaine:
Please rebase this on top of `master`.

koraktor:
No problem. Done.

bfontaine:
Could you please update the tests in `Library/Homebrew/test/test_utils.rb` so the test suite pass? It stills think `pretty_duration(1)` should return `"2 seconds"`.

koraktor:
Sure, sorry about that. Fixed.

bfontaine:
Thank you for your contribution to Homebrew @koraktor; we appreciate it! :tada:

PR #44726, opened 2015-10-07
Qscintilla2 2.9

### LINKED ISSUE #41464: qscintilla2 fails to build on 10.11: sip files generating qt5 API, but only qt4 installed
When generating the python bindings, the sip files all die with error:

```
./sipQsciQsciLexerCSharp.cpp:194:43: error: no viable conversion from 'const QMetaMethod' to 'const char *'
        QsciLexerCSharp::disconnectNotify(a0);
                                          ^~
/Volumes/Xcode/Brew/Cellar/qt/4.8.7/lib/QtCore.framework/Versions/4/Headers/qobject.h:291:47: note: passing argument to parameter 'signal' here
    virtual void disconnectNotify(const char *signal);
                                              ^
./sipQsciQsciLexerCSharp.cpp:212:40: error: no viable conversion from 'const QMetaMethod' to 'const char *'
        QsciLexerCSharp::connectNotify(a0);
                                       ^~
/Volumes/Xcode/Brew/Cellar/qt/4.8.7/lib/QtCore.framework/Versions/4/Headers/qobject.h:290:44: note: passing argument to parameter 'signal' here
    virtual void connectNotify(const char *signal);
                                           ^
```

QT4 used the char* argument for these methods, QT5 switched it to use QMetaMethod instead.
So it looks like sip is generating QT5 calls even though I only have QT4 installed?  Not sure if this is a bug/issue with sip, or something in the build configuration for qscintilla...

apjanke:
Depends on (and includes) #44725
Fixes #41464

### DIFF
--- a/Library/Formula/pyqt.rb
+++ b/Library/Formula/pyqt.rb
@@ -1,9 +1,8 @@
 class Pyqt < Formula
   desc "Python bindings for Qt"
   homepage "http://www.riverbankcomputing.co.uk/software/pyqt"
-  url "https://downloads.sf.net/project/pyqt/PyQt4/PyQt-4.11.3/PyQt-mac-gpl-4.11.3.tar.gz"
-  sha256 "8b8bb3a2ef8b7368710e0bc59d6e94e1f513f7dbf10a3aaa3154f7b848c88b4d"
-  revision 1
+  url "https://downloads.sf.net/project/pyqt/PyQt4/PyQt-4.11.4/PyQt-mac-gpl-4.11.4.tar.gz"
+  sha256 "f178ba12a814191df8e9f87fb95c11084a0addc827604f1a18a82944225ed918"
 
   bottle do
     sha256 "45a0bd946b7eca2af72aa3e56bda87ad90ef54bda91f0aea05c0ea2d50d6e8c4" => :el_capitan
--- a/Library/Formula/qscintilla2.rb
+++ b/Library/Formula/qscintilla2.rb
@@ -1,8 +1,8 @@
 class Qscintilla2 < Formula
   desc "Port to Qt of the Scintilla editing component"
   homepage "http://www.riverbankcomputing.co.uk/software/qscintilla/intro"
-  url "https://downloads.sf.net/project/pyqt/QScintilla2/QScintilla-2.8.4/QScintilla-gpl-2.8.4.tar.gz"
-  sha256 "9b7b2d7440cc39736bbe937b853506b3bd218af3b79095d4f710cccb0fabe80f"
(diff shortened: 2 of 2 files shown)


apjanke:
Test bot failure:

```
==> make install
cp -f "libqscintillaplugin.dylib" "/usr/local/Cellar/qscintilla2/2.9/plugins/designer/libqscintillaplugin.dylib"
Error: Operation not permitted - (/usr/local/Cellar/qscintilla2/2.9/plugins/designer/libqscintillaplugin.dylib, /usr/local/opt/qt/plugins/designer/libqscintillaplugin.dylib)
```

No idea what's going on here.

DomT4:
Looks like it's trying to copy the `dylib` into the `qt` prefix, which is forbidden under sandboxing. We'll need to convince it to keep itself within its own prefix.

apjanke:
Looks like it's [the `qscintilla2` formula itself](https://github.com/Homebrew/homebrew/blob/8658e27d7012cd7326ad150b34f8b0c1d18ba829/Library/Formula/qscintilla2.rb#L80-L82) that's adding that file under the `qt` prefix.

```
      # symlink Qt Designer plugin (note: not removed on qscintilla2 formula uninstall)
      ln_sf prefix/"plugins/designer/libqscintillaplugin.dylib",
            Formula["qt"].opt_prefix/"plugins/designer/"
```

This was introduced by #26152, to get the designer plugin working. What do you want to do here? Is there a shared `qt` plugin location yet? Or just let this pass?

mikemcquaid:
> Is there a shared qt plugin location yet?

There isn't but if you wanted to create one that'd be :sparkles:

apjanke:
Don't get your hopes up too much. I took a brief look at it, and it seems doable, but boy that's a lot of hours of compiling stuff to test it out. :)

apjanke:
For Qt plugins: do we need to support using Qt plugins from un-linked kegs?

mikemcquaid:
@apjanke No, I don't think so.

McPo:
Ill just add that if somebody wants to temporarily resolve this issue for themselves.

https://www.quora.com/How-do-I-turn-off-the-rootless-in-OS-X-El-Capitan-10-11

schoeps:
Hi, is someone still working on this issue? I just thought about adding qt5 support to Qscintilla2....

mikemcquaid:
@schoeps Feel free to open a new PR!

UniqMartin:
The issue of the shared Qt plug-in path has been resolved in #47245 and the current `qscintilla2` formula has been adapted accordingly. This PR should be fine now once it's rebased.

mikemcquaid:
@apjanke gentle poke on this 😉

apjanke:
Superseded by Homebrew/homebrew-core#7; let's see how this works.

PR #44730, opened 2015-10-08
Add jack support for cmus

danielcotton:
In some cases, people may prefer to use jack instead of libao as an output plugin.


bfontaine:
Thanks a lot @danielcotton! :tada:

PR #44732, opened 2015-10-08
pyenv-virtualenv v20151006

yyuu:
https://github.com/yyuu/pyenv-virtualenv/releases/tag/v20151006.


xu-cheng:

PR #44738, opened 2015-10-08
truncate 0.8: truncate a file to a given size

artw:
remade needed changes as a patch as Baptiste Fontaine proposed.

### DIFF
--- a/Library/Formula/truncate.rb
+++ b/Library/Formula/truncate.rb
@@ -1,0 +1,36 @@
+class Truncate < Formula
+  desc "truncates a file to a given size"
+  homepage "http://www.vanheusden.com/truncate"
+  url "http://www.vanheusden.com/truncate/truncate-0.8.tgz"
+  sha256 "165914d643351324d69096427d3ce5d3ab13aed3f7c990a71ba3bd0c6f278f00"
+
+  patch :DATA # remove missing file; rename syscalls for OSX which are natively 64bit
+
+  def install
+    system "make"
+    bin.install "truncate"
+    man1.install "truncate.1"
+  end
+
+  test do
+    system "#{bin}/truncate", "-s", "1k", "testfile"
+  end
+end
+__END__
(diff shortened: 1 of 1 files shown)


Library/Formula/truncate.rb
```diff
@@ -0,0 +1,34 @@
...
+  url "http://www.vanheusden.com/truncate/truncate-0.8.tgz"
+  sha256 "165914d643351324d69096427d3ce5d3ab13aed3f7c990a71ba3bd0c6f278f00"
+  patch :DATA # remove missing file; rename syscalls for OSX which are natively 64bit
+
+  def install
+    system "make"
+    bin.install "truncate"
+    man1.install "truncate.1"
+  end
+  test do
```
bfontaine: Please add a newline between `end` and `test do`.

Library/Formula/truncate.rb
```diff
@@ -0,0 +1,34 @@
+class Truncate < Formula
+  desc "truncates a file to a given size"
+  homepage "http://www.vanheusden.com/truncate"
+  url "http://www.vanheusden.com/truncate/truncate-0.8.tgz"
+  sha256 "165914d643351324d69096427d3ce5d3ab13aed3f7c990a71ba3bd0c6f278f00"
+  patch :DATA # remove missing file; rename syscalls for OSX which are natively 64bit
```
DomT4: Please leave an empty line between the initial block and any patching. The need to patch for build success also needs to be reported upstream.

Library/Formula/truncate.rb
```diff
@@ -0,0 +1,36 @@
+class Truncate < Formula
+  desc "truncates a file to a given size"
+  homepage "http://www.vanheusden.com/truncate"
+  url "http://www.vanheusden.com/truncate/truncate-0.8.tgz"
```
bfontaine: You can use `https://` for both of these links.

PR #44739, opened 2015-10-08
Make "pgroonga" 0.9.0 to be optional resource in postgresql formula

cosmo0920:
Revised https://github.com/Homebrew/homebrew/pull/41174 with `optional resource` technique.

### DIFF
--- a/Library/Formula/postgresql.rb
+++ b/Library/Formula/postgresql.rb
@@ -39,8 +39,10 @@ class Postgresql < Formula
   depends_on "readline"
   depends_on "libxml2" if MacOS.version <= :leopard # Leopard libxml is too old
   depends_on :python => :optional
-  depends_on "groonga" => :build if build.with? "pgroonga"
-  depends_on "pkg-config" => :build if build.with? "pgroonga"
+  if build.with? "pgroonga"
+    depends_on "groonga" => :build
+    depends_on "pkg-config" => :build
+  end
 
   conflicts_with "postgres-xc",
     :because => "postgresql and postgres-xc install the same binaries."
--- a/Library/Formula/postgresql.rb
+++ b/Library/Formula/postgresql.rb
@@ -96,7 +96,7 @@ class Postgresql < Formula
     # Now build and install pgroonga as optional resource
     if build.with? "pgroonga"
       resource("pgroonga").stage do
-        ENV.append_path "PATH", "#{Formula["postgresql"].bin}"
+        ENV.append_path "PATH", Formula["postgresql"].bin
         system "make"
         system "make", "install"
       end
--- a/Library/Formula/postgresql.rb
+++ b/Library/Formula/postgresql.rb
@@ -96,7 +96,7 @@ class Postgresql < Formula
     # Now build and install pgroonga as optional resource
     if build.with? "pgroonga"
       resource("pgroonga").stage do
-        ENV.append_path "PATH", Formula["postgresql"].bin
+        ENV.append_path "PATH", bin
         system "make"
         system "make", "install"
       end
--- a/Library/Formula/postgresql.rb
+++ b/Library/Formula/postgresql.rb
@@ -24,15 +24,23 @@ class Postgresql < Formula
   option "without-perl", "Build without Perl support"
   option "without-tcl", "Build without Tcl support"
   option "with-dtrace", "Build with DTrace support"
+  option "with-pgroonga", "Build with the PGroonga postgresql extension."
 
   deprecated_option "no-perl" => "without-perl"
   deprecated_option "no-tcl" => "without-tcl"
   deprecated_option "enable-dtrace" => "with-dtrace"
 
+  resource "pgroonga" do
+    url "http://packages.groonga.org/source/pgroonga/pgroonga-0.9.0.tar.gz"
+    sha256 "846b89d20c847bf54103978e5234fab0fbf95eafdcfffb35eac5273a26357f51"
+  end
+
   depends_on "openssl"
   depends_on "readline"
   depends_on "libxml2" if MacOS.version <= :leopard # Leopard libxml is too old
   depends_on :python => :optional
+  depends_on "groonga" => :build if build.with? "pgroonga"
+  depends_on "pkg-config" => :build if build.with? "pgroonga"
 
   conflicts_with "postgres-xc",
     :because => "postgresql and postgres-xc install the same binaries."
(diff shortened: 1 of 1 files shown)


cosmo0920:
This PR should solve sandbox issue, but test-bot CI reports tests failure.
How do I fix them?

Library/Formula/postgresql.rb
```diff
@@ -24,15 +24,23 @@ class Postgresql < Formula
...
+  resource "pgroonga" do
+    url "http://packages.groonga.org/source/pgroonga/pgroonga-0.9.0.tar.gz"
+    sha256 "846b89d20c847bf54103978e5234fab0fbf95eafdcfffb35eac5273a26357f51"
+  end
+
   depends_on "openssl"
   depends_on "readline"
   depends_on "libxml2" if MacOS.version <= :leopard # Leopard libxml is too old
   depends_on :python => :optional
+  depends_on "groonga" => :build if build.with? "pgroonga"
```
DomT4: If two variables are the same block them:

```ruby
if build.with? "pgroonga"
  depends_on "groonga" => :build
end
```

Library/Formula/postgresql.rb
```diff
@@ -24,15 +24,23 @@ class Postgresql < Formula
   option "without-perl", "Build without Perl support"
   option "without-tcl", "Build without Tcl support"
   option "with-dtrace", "Build with DTrace support"
+  option "with-pgroonga", "Build with the PGroonga postgresql extension."
 
   deprecated_option "no-perl" => "without-perl"
   deprecated_option "no-tcl" => "without-tcl"
   deprecated_option "enable-dtrace" => "with-dtrace"
 
+  resource "pgroonga" do
```
DomT4: Nit but resources should go beneath dependencies.

Library/Formula/postgresql.rb
```diff
@@ -82,6 +90,15 @@ def install
 
     system "./configure", *args
     system "make", "install-world"
+
+    # Now build and install pgroonga as optional resource
+    if build.with? "pgroonga"
+      resource("pgroonga").stage do
+        ENV.append_path "PATH", "#{Formula["postgresql"].bin}"
```
DomT4: Doesn't need the surrounding `""` here. Should work as `ENV.append_path "PATH", Formula["postgresql"].bin`

DomT4:
> This PR should solve sandbox issue, but test-bot CI reports tests failure.
How do I fix them?

On El Capitan it's using the deprecated system OpenSSL instead of ours. It probably needs something like `--with-openssl=#{Formula["openssl"].opt_bin}.` Check the `configure` switches and see what's available on that.

The other failure is a sandbox failure elsewhere, with the test trying to use a configuration file we no longer permit:

```
Testing pgpool-ii
==> Using the sandbox
==> /usr/local/Cellar/pgpool-ii/3.4.2/bin/pg_md5 --md5auth pool_passwd --config-file /usr/local/etc/pgpool.conf.sample
ERROR: pid 81709: initializing pool password, failed to open file:"/usr/local/etc/pool_passwd"
Error: pgpool-ii: failed
Failed executing: /usr/local/Cellar/pgpool-ii/3.4.2/bin/pg_md5 --md5auth pool_passwd --config-file /usr/local/etc/pgpool.conf.sample
```

cosmo0920:
> Check the configure switches and see what's available on that.

`--with-openssl` option is available, but it does not nothing passed there.
I fixed to pass `Formula["openssl"].opt_prefix` into this option.

Library/Formula/postgresql.rb
```diff
@@ -82,6 +93,15 @@ def install
 
     system "./configure", *args
     system "make", "install-world"
+
+    # Now build and install pgroonga as optional resource
+    if build.with? "pgroonga"
+      resource("pgroonga").stage do
+        ENV.append_path "PATH", Formula["postgresql"].bin
```
xu-cheng: It should be just `bin` here. You don't need `Formula["postgresql"]` to point itself.

xu-cheng:
Can you add a test for this option? Something like:
```ruby
test do
  ...
  if build.with? "pgroonga"
    ...
  end
end
```

Also our CI won't test non default option. So you should test this locally.

DomT4:
> Also our CI won't test non default option. 

I'm not sure it's worth adding tests for non-default options. We've refused to do this several times over the last couple months even because it opens the floodgates to having a test element for every option even though we'd never use it. Has our position changed here?

xu-cheng:
> We've refused to do this several times over the last couple months

I was unaware this. Please ignore my comment on adding new test. But still this option need to be tested locally.

DomT4:
I'm fine if we want to have that team discussion; just didn't know if I'd missed something somewhere :)

cosmo0920:
>>    Check the configure switches and see what's available on that.

> --with-openssl option is available, but it does not nothing passed there.
> I fixed to pass Formula["openssl"].opt_prefix into this option.

Hmm, `--with-openssl` does not accept arguments.... :/

Library/Formula/postgresql.rb
cosmo0920: I've confirmed that building `pgroonga` succeeded using `bin` here.

DomT4:
@BrewTestBot test this please

DomT4:
This is passing the CI now. The OpenSSL issue got fixed with the version bumps & changes merged yesterday. How does `pgroonga` know where to install, out of interest? I noticed there's no prefix argument being passed anywhere.

cosmo0920:
> How does pgroonga know where to install, out of interest? I noticed there's no prefix argument being passed anywhere.

Ah..., yes. Is this not suitable for Homebrew's policy?
`PGroonga`'s makefile calculates suitable installation location.
In this situation, `pgroonga.so` is installed in `$(brew -prefix)/Cellar/postgresql/<version>/lib` and `pgroonga.sql` is installed in  `$(brew -prefix)/Cellar/postgresql/<version>/share/postgresql/extension/`.

DomT4:
It's fine if that's the case. Just wanted to check it was installing into the prefix and not a generic `/usr/local` directory or such.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/022faf82010f19e8eeac36ba4079ce266f4b2b8d. Thank you for another contribution to Homebrew @cosmo0920 :star2:

cosmo0920:
Thanks! I'm appreciate to make to be pgroonga optional resource in postgresql formula!

PR #44740, opened 2015-10-08
cdrdao: update two links in comments

vszakats:


bfontaine:
Thank you @vszakats :smile:

PR #44417, opened 2015-09-28
autocode: rename from crystal

ctate:
Hi - we have renamed our project as `autocode` to resolve conflicts with `crystal` which should be used for crystal-lang (manastech/crystal) instead.

PR #44742, opened 2015-10-08
Add formula crystal (0.8.0)

waj:
Making use of the released `crystal` name at #44417 (Thanks @ctate!)

### DIFF
--- a/Library/Formula/crystal-lang.rb
+++ b/Library/Formula/crystal-lang.rb
@@ -1,0 +1,61 @@
+class CrystalLang < Formula
+  desc "Fast and statically typed, compiled language with Ruby-like syntax"
+  homepage "http://crystal-lang.org/"
+  url "https://github.com/manastech/crystal/archive/0.8.0.tar.gz"
+  sha256 "986a000bb2eded22e446fd55c543062770ec4000e28791f0b07f63fcee37b245"
+  head "https://github.com/manastech/crystal.git"
+
+  resource "boot" do
+    url "https://github.com/manastech/crystal/releases/download/0.8.0/crystal-0.8.0-1-darwin-x86_64.tar.gz"
+    sha256 "5a16826145a846da3548e875cf104bdfc04c35cd6628cf66487de1bfbe9c5faf"
+  end
+
+  resource "shards" do
+    url "https://github.com/ysbaddaden/shards/archive/v0.5.1.tar.gz"
+    sha256 "95916792766e42e3b005b144190c3f88d5cb8bcbaaddeb8fa8ced8bac1ef424d"
+  end
+
+  option "without-release", "Do not build the compiler in release mode"
+  option "without-shards", "Do not include `shards` dependency manager"
+
+  depends_on "libevent"
+  depends_on "libpcl"
+  depends_on "bdw-gc"
+  depends_on "llvm" => :build
+  depends_on "libyaml" if build.with?("shards")
+
+  def install
+    (buildpath/"boot").install resource("boot")
+
+    if build.head?
+      ENV["CRYSTAL_CONFIG_VERSION"] = `git rev-parse --short HEAD`.strip
+    else
+      ENV["CRYSTAL_CONFIG_VERSION"] = version
+    end
+
+    ENV["CRYSTAL_CONFIG_PATH"] = prefix/"src:libs"
+    ENV.append_path "PATH", "boot/bin"
+
+    if build.with? "release"
+      system "make", "crystal", "release=true"
+    else
+      system "make", "llvm_ext"
+      (buildpath/".build").mkpath
+      system "bin/crystal", "build", "-o", ".build/crystal", "src/compiler/crystal.cr"
+    end
+
+    if build.with? "shards"
+      resource("shards").stage do
+        system buildpath/"bin/crystal", "build", "-o", buildpath/".build/shards", "src/shards.cr"
+      end
+      bin.install ".build/shards"
+    end
+
+    bin.install ".build/crystal"
+    prefix.install "src"
+  end
+
+  test do
+    system "#{bin}/crystal", "eval", "puts 1"
+  end
+end


Library/Formula/crystal.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+    end
+
+    bin.install ".build/crystal"
+    bin.install ".build/shards" if build.with?("shards")
+    prefix.install "src"
+  end
+
+  def test
+    system "crystal", "eval", "puts 1"
```
bfontaine: Please use `"#{bin}/crystal"` instead of `"crystal"`.

DomT4:
@xu-cheng I thought we couldn't use the old names at this point as they became symlinks? Was that fixed?

Library/Formula/crystal.rb
waj: Done

xu-cheng:
@DomT4 if users have installed the formula using the oldname, they won't be able to install new formula(although we need to add some check in `brew install` to warn this situation) unless they uninstall old formula. For other users, it will be ok. 

Since there will be some adjustments for core logic required, let's hold off merging this PR for the moment.

ctate:
@xu-cheng Anything I can help with? If you let me know the exact behavior, I can submit a PR.

waj:
@xu-cheng Is there any planned timeframe for that adjustment? Should I wait? I could otherwise name this formula `crystal-lang` and rename it to `crystal` in the future when the logic change is done.
Is there any open issue I can follow to know when such change is done and I can rename the formula back?

will:
Awesome, looking forward to this, thanks @waj

DomT4:
`crystal-lang` seems fine to me until we decide how to handle freeing up old names properly, unless Xu has a magical fix in mind already, which he may well do :smile:.

waj:
It's a pity that actually @ctate asked exactly about this before sending the PR (https://github.com/Homebrew/homebrew/pull/44417#issuecomment-144293862) but now neither can use the name.

I can wait if you say that the problem will be fixed in a reasonable time and we avoid another rename in the future.

mikemcquaid:
We can't promise a timeframe as we're all volunteers working in our spare time. I do recommend just waiting, though.

waj:
@mikemcquaid I totally understand and I don't expect any promise. Just saying that if someone is currently working in the fix then it might be better to wait. Otherwise I'll rename the formula and leave `crystal` unused.

mikemcquaid:
No-one is working on it right now but it's a planned feature for future. Perhaps renaming this at a later stage is a good idea.

waj:
Ok, I just renamed the formula then. Thanks!

waj:
The build is failing but I think it's not my fault :innocent:
Anything else needed to get this through?

Library/Formula/crystal-lang.rb
```diff
@@ -0,0 +1,61 @@
...
+        system buildpath/"bin/crystal", "build", "-o", buildpath/".build/shards", "src/shards.cr"
+      end
+    end
+
+    bin.install ".build/crystal"
+    bin.install ".build/shards" if build.with?("shards")
+    prefix.install "src"
+  end
+
+  def test
```
bfontaine: Please use `test do` instead of `def test`.

Library/Formula/crystal-lang.rb
```diff
@@ -0,0 +1,61 @@
+class CrystalLang < Formula
+  homepage "http://crystal-lang.org/"
```
bfontaine: Please add a short description above this line using `desc "..."`.

Library/Formula/crystal-lang.rb
```diff
@@ -0,0 +1,61 @@
...
+    sha256 "95916792766e42e3b005b144190c3f88d5cb8bcbaaddeb8fa8ced8bac1ef424d"
+  end
+
+  depends_on "libevent"
+  depends_on "libpcl"
+  depends_on "bdw-gc"
+  depends_on "llvm" => :build
+
+  option "without-release", "Do not build the compiler in release mode"
+  option "without-shards", "Do not include `shards` dependency manager"
```
bfontaine: Please move the `option`s before the `depends_on` block.

waj:
@bfontaine Done

Library/Formula/crystal-lang.rb
```diff
@@ -0,0 +1,61 @@
+class CrystalLang < Formula
+  desc "Efficient, type safe and statically compiled language with Ruby inspired syntax"
```
bfontaine: Please shorten this description. “`crystal-lang: <desc>` should fit under 80 chars.

```
==> audit problems
crystal-lang:
Length is calculated as crystal-lang + desc. (currently 93)
```

waj: Done!

waj:
@bfontaine Anything else?

Library/Formula/crystal-lang.rb
```diff
@@ -0,0 +1,61 @@
...
+    if build.head?
+      ENV["CRYSTAL_CONFIG_VERSION"] = `git rev-parse --short HEAD`.strip
+    else
+      ENV["CRYSTAL_CONFIG_VERSION"] = version
+    end
+
+    ENV["CRYSTAL_CONFIG_PATH"] = prefix/"src:libs"
+    ENV.append_path "PATH", "boot/bin"
+
+    if build.with?("release")
```
bfontaine: You can drop the parentheses: `build.with? "release"`.

Library/Formula/crystal-lang.rb
```diff
@@ -0,0 +1,61 @@
+class CrystalLang < Formula
+  desc "Fast and statically typed, compiled language with Ruby like syntax"
```
bfontaine: `Ruby-like`

Library/Formula/crystal-lang.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    ENV["CRYSTAL_CONFIG_PATH"] = prefix/"src:libs"
+    ENV.append_path "PATH", "boot/bin"
+
+    if build.with?("release")
+      system "make", "crystal", "release=true"
+    else
+      system "make", "llvm_ext"
+      mkdir ".build"
```
bfontaine: Use `(buildpath/".build").mkpath`

Library/Formula/crystal-lang.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    if build.with?("shards")
+      resource("shards").stage do
+        system buildpath/"bin/crystal", "build", "-o", buildpath/".build/shards", "src/shards.cr"
+      end
+    end
+
+    bin.install ".build/crystal"
+    bin.install ".build/shards" if build.with?("shards")
```
bfontaine: You can move this in the `if build.with?("shards")` above.

Library/Formula/crystal-lang.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    if build.with?("release")
+      system "make", "crystal", "release=true"
+    else
+      system "make", "llvm_ext"
+      mkdir ".build"
+      system "bin/crystal", "build", "-o", ".build/crystal", "src/compiler/crystal.cr"
+    end
+
+    if build.with?("shards")
```
bfontaine: You can drop the parentheses here too.

bfontaine:
@waj Some style nitpicking but otherwise we’re good :smile:

waj:
@bfontaine Done!

mikemcquaid:

For future reference the preferred commit message format for new formulae is `crystal-lang 0.8.0 (new formula)`.

PR #44750, opened 2015-10-08
svtplay-dl 0.20.2015.10.08

spaam:
new version of svtplay-dl :tada:

### DIFF
--- a/Library/Formula/svtplay-dl.rb
+++ b/Library/Formula/svtplay-dl.rb
@@ -1,12 +1,23 @@
 class SvtplayDl < Formula
   desc "Download videos from http://svtplay.se"
   homepage "https://github.com/spaam/svtplay-dl"
-  url "https://github.com/spaam/svtplay-dl/archive/0.10.2015.08.24.tar.gz"
-  sha256 "0dcd2c6862ad0ed00a2b782f6484f005431de3229360e1797810f7adf16aedd9"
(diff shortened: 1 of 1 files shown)


Library/Formula/svtplay-dl.rb
```diff
@@ -1,12 +1,23 @@
...
+  end
+
   def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
+    resources.each do |r|
+      r.stage do
+        system "python", *Language::Python.setup_install_args(libexec/"vendor")
+      end
+    end
     bin.install "svtplay-dl"
```
DomT4: Presumably we need to env script this so it can actually find the `requests` vendored?

spaam: is it better to use?
```
    ENV.prepend_create_path "PYTHONPATH", libexec+"lib/python2.7/site-packages"

    resources.each do |r|
      r.stage { system "python", *Language::Python.setup_install_args(libexec) }
    end
```

DomT4: You should be able to do something like:

```ruby
(libexec/"bin").install "svtplay-dl"
```

DomT4:
Closing as dead. Please feel free to resubmit as fresh with the comments above addressed and audit passing, if you'd like to. Thanks for the submission to Homebrew!

PR #44756, opened 2015-10-09
pius 2.2.1

tresni:
Get latest version and add support for HEAD to make it easy to test the latest

### DIFF
--- a/Library/Formula/pius.rb
+++ b/Library/Formula/pius.rb
@@ -11,7 +11,7 @@ class Pius < Formula
     ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
     # Replace hardcoded gpg path (WONTFIX):
     # https://sourceforge.net/p/pgpius/bugs/12/
-    inreplace "libpius/constants.py", /\/usr\/bin\/gpg2?/, "#{HOMEBREW_PREFIX}/bin/gpg"
+    inreplace "libpius/constants.py", %r{/usr/bin/gpg2?}, "#{HOMEBREW_PREFIX}/bin/gpg"
     system "python", *Language::Python.setup_install_args(libexec)
     bin.install Dir["#{libexec}/bin/*"]
     bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
--- a/Library/Formula/pius.rb
+++ b/Library/Formula/pius.rb
@@ -11,6 +11,9 @@ class Pius < Formula
     ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
     # Replace hardcoded gpg path (WONTFIX):
     # https://sourceforge.net/p/pgpius/bugs/12/
+    # According to the author, the next version of pius should ONLY support gpg2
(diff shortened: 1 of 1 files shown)


Library/Formula/pius.rb
```diff
@@ -10,7 +11,7 @@ def install
     ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
     # Replace hardcoded gpg path (WONTFIX):
     # https://sourceforge.net/p/pgpius/bugs/12/
-    inreplace "libpius/constants.py", "/usr/bin/gpg", "#{HOMEBREW_PREFIX}/bin/gpg"
+    inreplace "libpius/constants.py", /\/usr\/bin\/gpg2?/, "#{HOMEBREW_PREFIX}/bin/gpg"
```
bfontaine: Use `%r{/usr/bin/gpg2?}` to avoid escaping backslashes.

Library/Formula/pius.rb
```diff
@@ -10,7 +11,7 @@ def install
     ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
     # Replace hardcoded gpg path (WONTFIX):
     # https://sourceforge.net/p/pgpius/bugs/12/
-    inreplace "libpius/constants.py", "/usr/bin/gpg", "#{HOMEBREW_PREFIX}/bin/gpg"
+    inreplace "libpius/constants.py", %r{/usr/bin/gpg2?}, "#{HOMEBREW_PREFIX}/bin/gpg"
```
DomT4: This may not always be correct, since if you have `gpg2` installed but not `gpg` it won't force the latter installation and there won't be a `gpg` binary in the `$PATH`. We may need to invent some logic to handle this situation. Thoughts @xu-cheng?

tresni: Given the package actually depends on `gpg` and not `gpg2`, shouldn't this always point to the `gpg` binary?  Currently pius works with both `gpg` and `gpg2`.  The next release will be `gpg2` only at which time I'd say change the depends and just update the regex.

All that happened is that the author changed the default.  The correct binary can always be given via CLI option as well.

DomT4: It's a special dependency that can be [satisfied](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/requirements.rb#L69-L74) by having either `gpg` or `gpg2` in the `$PATH`.

I guess if `gpg2` becomes the default soon I'm fine with this as-is. Perhaps it's worth updating the comment above the `inreplace` to reference this discussion.

tresni: I can update the PR to include that. Just an FYI that I will be without internet for a week so unable to respond to additional requests till then should they come up.

DomT4: That'd be great, thanks. It looks more or less fine if we get the comment added, but thank you for the heads up on being away - Helps us not worry a PR has gone dead if you stop responding for a while.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/6f67e679d479a59f3a0fa9a443123b62a269177a. Thank you for your contribution to Homebrew @tresni; we appreciate it! :smiley_cat:

PR #44758, opened 2015-10-09
audit fixes: C 02: cdb - chisel

apjanke:


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/2a0946fe5ca21e9d29862aa6417fe41858388e52. Thank you once more :wink: 

Had to revert one of the changes in https://github.com/Homebrew/homebrew/commit/0708ef94ad51f3bdf8d9ea95bba30de947ec236f as Ruby 1.8 which Mountain Lion and below use doesn't support that syntax tweak.

apjanke:
Too bad we can't disable that style check. Looks like rubocop's configuration granularity is too coarse to disable that without also disabling the other trailing-comma checks.

DomT4:
We'll be moving to Ruby 2.x exclusively in the semi-near-future, so supporting ye olde syntax won't be an issue at that point and we can start to set fire to all the little compatibility dances we've added over time.

Issue #43864, opened 2015-09-12
brew desc -d returns wrong results (brew debug)


denji:
Problem when searching by description (`-d`), it show the debug-mode homebrew
```
desc  [-s|-n|-d]
      <pattern>:
      Search both name and description (-s),
      just the names (-n),
      or just the descriptions (-d) for <pattern>.
```

`brew desc -d pgloader` – debug mode brew-ruby, but search description is not working
`brew desc -s pgloader` – searches for substring in name formula too (unnecessary results)

xu-cheng:
We probably need to choose a better flag than `--description`.

apjanke:
> brew desc -d pgloader – debug mode brew-ruby, but search description is not working

The description search does seem to be working. This particular search produces no output because the string "pgloader" doesn't appear in any formula descriptions.

```
$ brew desc -d pgloader
$ brew desc -d PostgreSQL | grep pgloader
pgloader: Data loading tool for PostgreSQL
```


About the debugging output:

It looks like this is because the Homebrew extension `ARGV.flag?` logic allows collisions for single-letter option names. The `-d` looks like both `--description` and `--debug` if it's asking with `flag?`.

`formulary.rb` [checks for a debug option with `ARGV.debug?`](https://github.com/Homebrew/homebrew/blob/433aa72b1ecfea67146bb29541ed53d23f351c9f/Library/Homebrew/formulary.rb#L81)

```
 STDERR.puts "#{$0} (#{self.class.name}): loading #{path}" if ARGV.debug?
```

which in turn [uses `flag?`](https://github.com/Homebrew/homebrew/blob/433aa72b1ecfea67146bb29541ed53d23f351c9f/Library/Homebrew/extend/ARGV.rb#L103-L105)

```
  def debug?
    flag?("--debug") || !ENV["HOMEBREW_DEBUG"].nil?
  end
```

which in turn [uses a substring](https://github.com/Homebrew/homebrew/blob/433aa72b1ecfea67146bb29541ed53d23f351c9f/Library/Homebrew/extend/ARGV.rb#L183-L185) to automatically recognize the single-letter form of the option.

```
  def flag?(flag)
    options_only.include?(flag) || switch?(flag[2, 1])
  end
```

So a `-d` in any command that uses the formulary classes will be seen as `--debug` in addition to whatever other meaning it has, and turn on debugging output.

[...]

PR #44762, opened 2015-10-09
manpage: add --debug/-d to main brew synopsis

apjanke:
Adds `--debug`|`-d` to the main `brew` synopsis, establishing that it may be recognized for all commands (which it is), and establishing that other commands shouldn't use `-d` for other meanings.

Follows up discussion in #43864.

### DIFF
--- a/Library/Homebrew/manpages/brew.1.md
+++ b/Library/Homebrew/manpages/brew.1.md
@@ -4,7 +4,7 @@ brew(1) -- The missing package manager for OS X
 ## SYNOPSIS
 
 `brew` `--version`<br>
-`brew` <command> [`--verbose`|`-v`] [<options>] [<formula>] ...
+`brew` <command> [`--verbose`|`-v`] [`--debug`|`-d`] [<options>] [<formula>] ...
 
 ## DESCRIPTION
 
@@ -15,7 +15,10 @@ didn't include with OS X.
 
 For the full command list, see the [COMMANDS][] section.
 
-With `--verbose` or `-v`, many commands print extra debugging information. Note that these flags should only appear after a command.
+With `--verbose` or `-v`, many commands print extra debugging information. 
+With `--debug` or `-d`, additional debugging output and behaviors are enabled.
+Details are documented under the individual commands below.
+Note that these flags should only appear after a command.
 
   * `install` <formula>:
     Install <formula>.
--- a/share/doc/homebrew/brew.1.html
+++ b/share/doc/homebrew/brew.1.html
@@ -7,7 +7,7 @@
 <h2 id="SYNOPSIS">SYNOPSIS</h2>
 
 <p><code>brew</code> <code>--version</code><br />
-<code>brew</code> <var>command</var> [<code>--verbose</code>|<code>-v</code>] [<var>options</var>] [<var>formula</var>] ...</p>
+<code>brew</code> <var>command</var> [<code>--verbose</code>|<code>-v</code>] [<code>--debug</code>|<code>-d</code>] [<var>options</var>] [<var>formula</var>] ...</p>
 
 <h2 id="DESCRIPTION">DESCRIPTION</h2>
 
@@ -18,7 +18,10 @@ didn't include with OS X.</p>
 
 <p>For the full command list, see the <a href="#COMMANDS" title="COMMANDS" data-bare-link="true">COMMANDS</a> section.</p>
 
-<p>With <code>--verbose</code> or <code>-v</code>, many commands print extra debugging information. Note that these flags should only appear after a command.</p>
+<p>With <code>--verbose</code> or <code>-v</code>, many commands print extra debugging information.
+With <code>--debug</code> or <code>-d</code>, additional debugging output and behaviors are enabled.
+Details are documented under the individual commands below.
+Note that these flags should only appear after a command.</p>
 
 <dl>
 <dt><code>install</code> <var>formula</var></dt><dd><p>Install <var>formula</var>.</p></dd>
--- a/share/man/man1/brew.1
+++ b/share/man/man1/brew.1
@@ -10,7 +10,7 @@
 \fBbrew\fR \fB\-\-version\fR
 .
 .br
-\fBbrew\fR \fIcommand\fR [\fB\-\-verbose\fR|\fB\-v\fR] [\fIoptions\fR] [\fIformula\fR] \.\.\.
+\fBbrew\fR \fIcommand\fR [\fB\-\-verbose\fR|\fB\-v\fR] [\fB\-\-debug\fR|\fB\-d\fR] [\fIoptions\fR] [\fIformula\fR] \.\.\.
 .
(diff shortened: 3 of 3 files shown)


Library/Homebrew/manpages/brew.1.md
```diff
@@ -15,8 +15,7 @@ didn't include with OS X.
 
 For the full command list, see the COMMANDS section.
 
-With `--verbose` or `-v`, many commands print extra debugging information.
-Note that these flags should only appear after a command.
+With `--verbose` or `-v`, many commands print extra debugging information. With `--debug` or `-d`, additional debugging output and behaviors are enabled. Details are documented under the individual commands below. Note that these flags should only appear after a command.
```
DomT4: Since we're explicitly talking about the `debug` now I'd be tempted to change the `verbose` explanation to note it includes full build output, etc rather than necessarily "debug" information. Unsure. Open to other maintainer thoughts.

UniqMartin: I agree it would be good to draw a clearer distinction between what `--verbose` enables and what `--debug` does. Also, can we stay with one sentence per line in here? This will make diffs more useful and doesn't negatively impact the rendered output.

UniqMartin:
Very minor nit, but the `-` of `-d` is missing in the commit message. Aside from this and my code comment, I'm very much in favor of this PR.

Library/Homebrew/manpages/brew.1.md
apjanke: Maybe the item for `install` could have a note about `--verbose` doing full build output, since that's the behavior for that particular subcommand?

The rest of the man page isn't one sentence per line; it's manually hard-wrapped at about 80 characters per line. I reformatted this section to be the same way.

apjanke:
Fixed the `-d` in the comment and updated formatting for the code.

Library/Homebrew/manpages/brew.1.md
UniqMartin: My bad. I hadn't taken into account the rest of the file when complaining about the formatting.

UniqMartin:
You haven't addressed @DomT4's and my main point though, better distinction between `--verbose` and `--debug` in the text. Where's the difference between “print extra debugging information” and “additional debugging output” and won't it confuse people more than not knowing about one of the options?

Take this with a grain of salt please, as I also don't fully understand/know to what extent these flags affect all of the various commands, maybe except for `brew install`, where they seem to be most useful and worth documenting, and `brew update`.

mikemcquaid:
I actually think it's worth omitting this text from the documentation as it's a developer-focused flag, mentioned in the manpage and we rarely if ever tell end-users to run it.

apjanke:
Rebased to pick up the changes from other recent documentation changes.

Making one last attempt to keep this open:

man pages are supposed to be pretty comprehensive. As a Unix user, I'd expect them to be the primary reference for a command (unless I'm on a silly GNU system that's pushing me to use `info`), so I think *all* the options a command takes should be mentioned in it. Even if they're hardly described, and users need to go elsewhere or to support forums to track it down. At least they know it's there and it might do something.

> Where's the difference between “print extra debugging information” and “additional debugging output”

I don't even know the difference myself, and I don't think there really is a general guideline that's laid out. Individual parts of the commands just seem to pick one or both of these and decide what to do for them.

> I actually think it's worth omitting this text from the documentation

Do you mean we should omit the "With `--debug` or `-d`, additional debugging" verbiage from this man page section? Or just not write about it more in additional documentation?

UniqMartin:
> man pages are supposed to be pretty comprehensive. As a Unix user, I'd expect them to be the primary reference for a command (unless I'm on a silly GNU system that's pushing me to use `info`), so I think *all* the options a command takes should be mentioned in it.

That's also my line of thinking. I'd actually go as far as saying that *all* our commands (even the ones only targeted towards maintainers) should be completely documented in our man page. This will also benefit new maintainers (and old maintainers that use a command/option infrequently; it's a shame to have to dig through the source to find the options).

By now I have formed the following opinion about `--debug` and `--verbose`:

- They should be removed from the `SYNOPSIS` section. I think they are not important enough for most users to be mentioned there.
- They should be removed from the `ESSENTIAL COMMANDS` section. They are really not that essential and not the first thing a new user has to learn about.
- I suggest to move them either to the beginning of the `COMMANDS` section or to create a separate section that precedes that one (maybe call it `COMMON OPTIONS`?).

**About the wording for `--debug` vs. `--verbose`:**

Maybe adjust it to say something along the lines of “`--debug` enables additional debugging output and behaviors” and “`--verbose` makes the output of most commands more verbose without altering their behavior”. I'm not sure if it is like that in all cases, but I think that's how it should be. So let's document it like that and fix commands that violate this description, once we notice.

mikemcquaid:
> That's also my line of thinking. I'd actually go as far as saying that all our commands (even the ones only targeted towards maintainers) should be completely documented in our man page.

I can't agree with this. I really don't want to point users towards stuff like `test-bot --cleanup` which will basically destroy their Homebrew installation.

UniqMartin:
> I really don't want to point users towards stuff like test-bot --cleanup which will basically destroy their Homebrew installation.

Sorry, but I don't get how this is supposed to protect people from doing something stupid. Not teaching people about (safe) sex, as we probably all know, also doesn't help to prevent unwanted pregnancies and sexually transmitted diseases. Also, commands like `rm -rf /` are documented while being much more dangerous. Furthermore, there are ways to deal with this perceived danger (some combination of):

- State that this is a command to be used by developers/maintainers only.
- Put its documentation in a separate `DEVELOPER COMMANDS` section.
- Refuse to do anything if such a command is invoked without `HOMEBREW_DEVELOPER` being set.

mikemcquaid:
> Not teaching people about (safe) sex, as we probably all know, also doesn't help to prevent unwanted pregnancies and sexually transmitted diseases.

Bad analogy; people want to and will do that anyway if you don't teach them. 99.9% (if not more) of Homebrew's users (hell, probably even the majority of our maintainers) will never want to run `brew test-bot`. I don't see adding end-user documentation for any `dev-cmd` as being necessary or even helpful.

UniqMartin:
My point wasn't that documentation will solve all problems, but that teaching people has more positive impact than hiding stuff from them. But I'll stop here, since my arguments seem to be less convincing than I thought they were.

apjanke:
> I can't agree with this. I really don't want to point users towards stuff like `test-bot --cleanup` which will basically destroy their Homebrew installation.

I'm finding this persuasive. Since everything in `brew` is stuck under one command, it doesn't fit so well in to the traditional Unix pattern of "focused commands, with one exhaustive man page per command." It's more like `perl`, with its sprawling functionality that gets split up in to different topics like `perlrun` / `perlref` / `perlplan9`, or `git`, which has a subcommand structure similar to `brew`, and does one man page per subcommand. 

Let's just close this out for now, and maybe revisit it after I'm more familiar with the entire set of Homebrew doco.

PR #44765, opened 2015-10-09
brew irb to consider --noverbose flag

scpeters:
Don't print console messages if `--noverbose` is supplied to `brew irb`. This flag was chosen since it is an existing `irb` command-line argument.

Before this PR:

~~~
$ echo "puts 'boost'.f.stable.url" | brew irb --noverbose --noecho
https://downloads.sourceforge.net/project/boost/boost/1.58.0/boost_1_58_0.tar.bz2

~~~

After this PR:

~~~
$ echo "puts 'boost'.f.stable.url" | brew irb --noverbose --noecho
https://downloads.sourceforge.net/project/boost/boost/1.58.0/boost_1_58_0.tar.bz2

~~~

### DIFF
--- a/Library/Homebrew/cmd/irb.rb
+++ b/Library/Homebrew/cmd/irb.rb
@@ -21,8 +21,10 @@ module Homebrew
       puts ":lua.f.methods - 1.methods"
       puts ":mpd.f.recursive_dependencies.reject(&:installed?)"
     else
-      ohai "Interactive Homebrew Shell"
-      puts "Example commands available with: brew irb --examples"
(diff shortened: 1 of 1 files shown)


xu-cheng:
Sorry, but I don't think you should use `brew irb` as a ruby interpreter other than interactive shell.

If you need such function, you can `brew tap homebrew/dev-tools` and use `brew ruby` command, e.g.
`brew ruby -e "puts 'boost'.f.stable.url"`

scpeters:
Thanks, I didn't know that already existed. That's even better!

PR #44772, opened 2015-10-09
postgresql 9.4.5

damireh:
http://www.postgresql.org/docs/9.4/static/release-9-4-5.html


ojab:
I've added this patch to PR #44785, I assume that this PR should be closed.

damireh:
@ojab thanks! will close

PR #44773, opened 2015-10-09
test: allow writing to certain var directories

xu-cheng:
Closes #44458.


DomT4:
Slightly bleh about loosening the sandbox on this, but it's better than allowing access to the whole of `var`.

xu-cheng:
Yes, I think this is a middle ground to let formula author be happier to write test case without messing around users' config files in `var` and `etc` in general.

Issue #40863, opened 2015-06-18
Proposal: display caveats in `brew info` according to how the formula has been installed.


alexeymuranov:
I find the meaning of `caveats` method confusing when it is used by `brew info`.

I see that conditionals like `build.with?` are often used in `caveats` method, and it makes sense if `caveats` generates post-install information.  However, in this case it is strange that `caveats` are also displayed for a formula that has not yet been installed.  What is the point of this?  Should the user run `brew info` with all possible combinations of options 2<sup>N</sup> times to learn all possible caveats, or should he or she know exactly the options with which to install before running `brew info` (and before `brew install`)?

IMO, this is even more confusing when running `brew info` for a formula that is installed.  In this case the (poor) user should not forget to provide exactly the same options that were given to `brew install`, otherwise the (unsuspecting) user will see the wrong caveats.

I would suggest to make `brew info` not display any caveats if the formula has not been installed, and otherwise display the caveats for the options that were actually used at installation.

What do you think?

P.S. Some documentation of the semantics of `caveats` would be helpful.

P.P.S.  IMO, it is strange anyway to use the same options with `brew install` and with `brew info`.

mikemcquaid:
> I would suggest to make brew info not display any caveats if the formula has not been installed, and otherwise display the caveats for the options that were actually used at installation.

I think that's a good idea.

alexeymuranov:
The same applies to `brew test` and friends, i think.

alexeymuranov:
A change probably must be made in how `build.include?`, `build.with?`, `build.without?` behave.  For example, they can raise an error if the formula is not installed and not being installed, and otherwise tell about how the formula was actually installed.

jacknagel:
`brew test` already respects installed options.

Changing the behavior of the BuildOptions class is not appropriate. You can see how we handle it for `brew test` here. https://github.com/Homebrew/homebrew/blob/39dc3bba1210678fdc5ce5687b898f97bbdb9cba/Library/Homebrew/formula.rb#L792

chdiza:
I don't think this is a good idea.  Sometimes the caveats for an uninstalled formula tell me something that makes me decide "I don't think I'm going to install that now", or "I had better back up my such-and-such before I try this formula".

alexeymuranov:
How do you know which caveats of an uninstalled formula to look for?  Currently they will be different depending on the options you give to `brew info`.

jacknagel:
[...]

PR #44779, opened 2015-10-09
caveats: show information based on build options

### LINKED ISSUE #44729: Installing --without-multilib and no OpenMP support
Hello,

Apologies in advance as I'm not very experienced with installing the compiler and if I'm posting in the wrong place. I've gone through all the steps to install gcc through brew with -without-multilib, but when I check the info of the gcc compiler it still says it's been installed with multilib and that OpenMP support might not work.

Essentially this is what I've done:

brew reinstall gcc --without-multilib

and then:

brew info gcc

shows this:

gcc: stable 5.2.0 (bottled)
GNU compiler collection
https://gcc.gnu.org
/usr/local/Cellar/gcc/5.2.0 (1285 files, 215M) *
  Built from source with: --without-multilib
From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/gcc.rb
==> Dependencies
Required: gmp ✔, libmpc ✔, mpfr ✔, isl ✔
==> Options
--with-all-languages
	Enable all compilers and languages, except Ada
--with-java
	Build the gcj compiler
--with-jit
	Build the jit compiler
--with-nls
	Build with native language support (localization)
--without-fortran
	Build without the gfortran compiler
--without-multilib
	Build without multilib support
==> Caveats
GCC has been built with multilib support. Notably, OpenMP may not work:
  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60670
If you need OpenMP support you may want to
  brew reinstall gcc --without-multilib

As some background, I'm on Mac OS Mavericks and have an application that I want to build that uses OpenMP and currently stumbles as it can't find <omp.h> (I'm building with gcc and -fopenmp is specified as flag). Since apparently Clang that ships with Xcode doesn't support it I decided to install gcc from reading some SO posts.

I'm not quite sure this is my problem but not sure why it says that it has been built with multilib support when I specifically installed it with the --without-multilib parameter.

Any help would be greatly appreciated.

Thank you,
Pedro

xu-cheng:
Fixes #44729.

### DIFF
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -7,8 +7,13 @@ class Caveats
 
   def caveats
     caveats = []
-    s = f.caveats.to_s
-    caveats << s.chomp + "\n" if s.length > 0
+    begin
+      build, f.build = f.build, Tab.for_formula(f)
+      s = f.caveats.to_s
+      caveats << s.chomp + "\n" if s.length > 0
+    ensure
+      f.build = build
+    end
     caveats << keg_only_text
     caveats << bash_completion_caveats
     caveats << zsh_completion_caveats


DomT4:
Also fixes #40863, I'd presume.

mikemcquaid:
:clap:

Issue #44652, opened 2015-10-06
Promote z3 from homebrew-science


acfoltzer:
Right now the Z3 SMT solver is only available through [homebrew-science](https://github.com/Homebrew/homebrew-science/blob/master/z3.rb). Other solvers like [CVC4](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/cvc4.rb) are available on mainline homebrew, and SMT solvers generally are being used more as dependencies for [other software](http://goto.ucsd.edu/~rjhala/liquid/haskell/blog/about/) and [programming languages](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/cryptol.rb). It would be nice to have more than just CVC4 to choose from without tapping homebrew-science.

It looks like this was proposed a couple years ago in #16188 and #21509, but at the time it was more difficult to build and was not available under an open source license. Hopefully there are no longer any blockers to including it in the main repo.

DomT4:
CC @Homebrew/science for thoughts/objections/cake/etc.

iml:
No objections from me.

sjackman:
Fine by me.

dpo:
We have only one dep on `z3` in `science`, so I guess it's ok.

DomT4:
@Homebrew/owners Any objections on moving `z3` *to* the Core?

dunn:
:ok_woman:

dunn:
That's "OK (good)", not "I have objection".

bfontaine:
:ok_hand:

DomT4:
@acfoltzer Go ahead and file PRs in Science to remove & here at Homebrew/homebrew to add :)

PR #44795, opened 2015-10-09
add z3 per #44652

acfoltzer:
see discussion in #44652

### DIFF
--- a/Library/Formula/z3.rb
+++ b/Library/Formula/z3.rb
@@ -1,0 +1,26 @@
+class Z3 < Formula
+  desc "High-performance theorem prover"
+  homepage "https://github.com/Z3Prover/z3"
+  url "https://github.com/Z3Prover/z3/archive/z3-4.4.0.tar.gz"
+  sha256 "65b72f9eb0af50949e504b47080fb3fc95f11c435633041d9a534473f3142cba"
+  head "https://github.com/Z3Prover/z3.git"
+  revision 1
+
+  def install
+    inreplace "scripts/mk_util.py", "dist-packages", "site-packages"
+    system "python", "scripts/mk_make.py", "--prefix=#{prefix}"
+
+    cd "build" do
+      system "make"
+      system "make", "install"
+    end
+
+    pkgshare.install "examples"
+  end
+
+  test do
+    system ENV.cc, "-I#{include}", "-L#{lib}", "-lz3",
+           pkgshare/"examples/c/test_capi.c", "-o", testpath/"test"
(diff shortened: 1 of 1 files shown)


Library/Formula/z3.rb
```diff
@@ -0,0 +1,33 @@
+class Z3 < Formula
+  desc "A high-performance theorem prover"
+  homepage "https://github.com/Z3Prover/z3"
+  url "https://github.com/Z3Prover/z3/archive/z3-4.4.0.tar.gz"
+  sha256 "65b72f9eb0af50949e504b47080fb3fc95f11c435633041d9a534473f3142cba"
+  head "https://github.com/Z3Prover/z3.git"
+  revision 1
```
DomT4: Cut this line. It'll be treated as a new formula essentially.

Library/Formula/z3.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  bottle do
```
DomT4: And kill this block. We'll add new bottles.

Library/Formula/z3.rb
```diff
@@ -0,0 +1,33 @@
+class Z3 < Formula
+  desc "A high-performance theorem prover"
```
DomT4: Per the audit check, this should be `desc "High-performance theorem prover"`

Library/Formula/z3.rb
xu-cheng: @DomT4 we cannot reset the revision here, otherwise users who already installed this formula will see formula's version gets downgraded.

DomT4: I was thinking about this earlier for a bit & I think you're right. It'd be nice to codify tap => core migration more officially if we're going to start seeing this sort of thing more often.

acfoltzer:
revised the branch per your comments; thanks!

beurdouche:
CI failure does not seem to be your fault...
"GitHub API rate limit exceeded" : I think you have to push it again

mikemcquaid:
It's OK, the default CI has passed.

mikemcquaid:
Will let @DomT4 pull if he's :ok:

DomT4:
Pulling now.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/4dbd577fec9ee5001c3755b03bfa3dea0dabe56b. Thank you for your contribution to Homebrew @acfoltzer! :smiley_cat:

acfoltzer:
Many thanks!

PR #44798, opened 2015-10-09
Simplified CxxStdlib

koraktor:
This drops the polymorphism introduced in e1f97e260ca320141c9c48158814aff895e76486 which isn’t really used here.

/cc @jacknagel

### DIFF
--- a/Library/Homebrew/cxxstdlib.rb
+++ b/Library/Homebrew/cxxstdlib.rb
@@ -16,20 +16,16 @@ class CxxStdlib
     if type && ![:libstdcxx, :libcxx].include?(type)
       raise ArgumentError, "Invalid C++ stdlib type: #{type}"
     end
-    klass = GNU_GCC_REGEXP === compiler.to_s ? GnuStdlib : AppleStdlib
-    klass.new(type, compiler)
+    new(type, compiler)
   end
 
   def self.check_compatibility(formula, deps, keg, compiler)
     return if formula.skip_cxxstdlib_check?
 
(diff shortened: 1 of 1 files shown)


jacknagel:
I don't work on Homebrew anymore, but the rationale for using polymorphism here is that we only have to perform the regex match once, at instantiation time, rather than on each comparison. It's purely a design choice. Why change it?

Library/Homebrew/cxxstdlib.rb
```diff
@@ -16,20 +16,16 @@ def self.create(type, compiler)
...
     stdlib = create(keg.detect_cxx_stdlibs.first, compiler)
-
-    begin
-      stdlib.check_dependencies(formula, deps)
-    rescue CompatibilityError => e
-      opoo e.message
-    end
+    stdlib.check_dependencies(formula, deps)
+  rescue CompatibilityError
+    opoo $!.message
```
jacknagel: Why change the error handling? The new version is less correct (the scope of the rescue has changed). What is the point of switching from `=> e` to `$?`?).

koraktor:
IMHO it’s harder to understand code that uses two different subclasses just to represent `true` / `false`. If the performance is really an issue here, there could be a boolean `@apple_compiler` variable instead.

DomT4:
Going to pass on this for now. If Jack isn't enthusiastic about the changed syntax around compiler handling you'll struggle to find another maintainer who is. Unless there's a significant performance issue here not sure what the absolute upside would be, and we've been slowly getting API documentation sorted to make the code clearer. Thanks for the submission though! :heart:

Issue #44048, opened 2015-09-16
xaric failed to build on 10.11


apjanke:
https://gist.github.com/2f35fa2a661c3b75fd71

Can't find OpenSSL, even though it already `depends_on "openssl"`.

```
checking for getpgid... yes
checking whether getpgrp requires zero arguments... yes
checking for OpenSSL... configure: error: Cannot find ssl libraries
```

DomT4:
Fixed by #44810.

PR #44802, opened 2015-10-10
Fixes Xaric config issue with openssl

Moisan:
This should fix [issue 44048](https://github.com/Homebrew/homebrew/issues/44048). The configure script needed a path to openssl opt directory and the only way to pass it was to assign it to an undocumented variable used only for this check.


DomT4:
You've got a `hbase` update that's already been merged in here, which may not be intentional.

Moisan:
Wow sorry, created my branch from an earlier branch instead of master. I will simply delete this PR and recreate a new one with a clean branch.

PR #44807, opened 2015-10-10
dnsperf: use secure homepage url and follow redir

vszakats:


DomT4:
:boat:'ed in https://github.com/Homebrew/homebrew/commit/e1b133396ec55332ee853c6f872135916bef975d. Thanks again!

PR #44810, opened 2015-10-10
Fix xaric configure with a path to openssl

Moisan:
This should fix [issue 44048](https://github.com/Homebrew/homebrew/issues/44048). The configure script needed a path to openssl opt directory and the only way to pass it was to assign it to an undocumented variable used only for this check.

### DIFF
--- a/Library/Formula/xaric.rb
+++ b/Library/Formula/xaric.rb
@@ -15,12 +15,12 @@ class Xaric < Formula
 
   def install
     system "./configure", "--disable-dependency-tracking",
-                          "--prefix=#{prefix}"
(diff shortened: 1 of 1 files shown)


DomT4:
We should report the need to use this upstream. `withval` is generally supposed to be tied to another variable, i.e. `--with-ssl=` rather than being called itself. They seem to have a Github repository available.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/0e255f3ab614e93f5217ce330e5532eae58bb6be. Thank you for reporting this upstream & your contribution to Homebrew @Moisan! :smiley_cat:

PR #44314, opened 2015-09-25
unar 1.9.1

santazhang:
Update `unar` to 1.9.1

Problem is, the source code archive `unar1.9.1_src.zip` has a path prefix "The Unarchiver" in it, while HEAD checkout does not. I have to add

```ruby
cd build.head? ? "." : "./The Unarchiver" do
```

to deal with it.

PR #44817, opened 2015-10-11
unar 1.9.1

jiegec:


jiegec:
Well, this is already done in https://github.com/Homebrew/homebrew/pull/44314. Why it works here but no in https://github.com/Homebrew/homebrew/pull/44314? The change is almost the same.

PR #44835, opened 2015-10-11
arangodb 2.7.0

fceller:
Updated to ArangoDB 2.7.0

### DIFF
--- a/Library/Formula/arangodb.rb
+++ b/Library/Formula/arangodb.rb
@@ -1,8 +1,8 @@
 class Arangodb < Formula
   desc "Universal open-source database with a flexible data model"
   homepage "https://www.arangodb.com/"
-  url "https://www.arangodb.com/repositories/Source/ArangoDB-2.6.9.tar.gz"
-  sha256 "1a56212eb52c92a2942969af6ac4a92eb4ef60be80dc6eb3868efb48c8edc515"
+  url "https://www.arangodb.com/repositories/Source/ArangoDB-2.7.0.tar.gz"
+  sha256 "67fc1b0cb75e9ebf9857b30f54805283461161611865c8b343adcec8bb6060ed"
 
   head "https://github.com/arangodb/arangodb.git", :branch => "unstable"
 
@@ -14,8 +14,19 @@ class Arangodb < Formula
 
   depends_on "go" => :build
   depends_on "openssl"
+  depends_on "readline"
 
   needs :cxx11
+
+  # GCC 5 no longer supports the --stdlib options. This patch checks
+  # if the option is available or not. It will be part of ArangoDB 2.7.1
+  # in which case the patch can be removed again
+  patch :DATA
+
+  fails_with :clang do
+    build 600
+    cause "Fails with compile errors"
+  end
 
   def install
     # clang on 10.8 will still try to build against libstdc++,
@@ -27,7 +38,6 @@ class Arangodb < Formula
       --disable-dependency-tracking
       --prefix=#{prefix}
       --disable-relative
-      --enable-mruby
       --datadir=#{share}
       --localstatedir=#{var}
     ]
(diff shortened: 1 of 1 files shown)


DomT4:
It looks like you need to our a dependency to our `readline`, based on your configure script output :wink: 

```
checking READLINE support... readline
checking READLINE version... 0x0402.hex
configure: error: Please install readline 6 from brew. If you have a non-standard brew location, use "--with-readline".
```

fceller:
@DomT4 yes, I missed that. Added dependency.

DomT4:
That got it passing on Yosemite & El Cap. Died on the [Mavericks build](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/34609/version=mavericks/testReport/junit/brew-test-bot/mavericks/install_arangodb/) though :cry:.

fceller:
It seems that the CLANG is not suitable on Mavericks. 

@DomT4 Do you know if there is GCC 4.9 available for Mavericks?

DomT4:
@fceller If you can build using GCC 5 it should be as simple as adding a:

```ruby
fails_with :clang do
  build 600
  cause "Causes compile failure"
end
```

If you specifically need `gcc49` it's more of an issue because we'd have to add `fails_with` blocks for both `clang` and `gcc5` which would remove the ability to bottle Mavericks' installation.

fceller:
@DomT4 thanks, will try. I do not know if it compiles with GCC 5 (in they earlier version of GCC 5, there were also some internal compiler errors).

fceller:
@DomT4 tried 

```
 26   fails_with :gcc5 do
 27     cause "V8 fails to compile"
 28   end
```

locally, but it still uses GCC5.

DomT4:
Syntax is `fails_with :gcc => "5"` :)

fceller:
@DomT4 perfect, that did the trick

DomT4:
Apologies for the delay here. Want to test the build manually on a Mavericks VM to make sure we both don't get swamped with bug reports when this is merged. Will get that done today.

fceller:
@DomT4 No worries. Unfortunately I do not have a Mavericks VM, so I cannot test it myself.

DomT4:
Afraid it [died still](https://gist.github.com/DomT4/a0430dfd139d2f7e93b7) :cry:.

fceller:
@DomT4 

Strange, du you understand, what the BUILD BOT ist doing? It ends with

    arangodb cannot be built with any available compilers.
      brew install gcc

It seems that it has downloaded gcc-5.x

    ==> Downloading https://homebrew.bintray.com/bottles/gcc-5.2.0.mavericks.bottle.2.tar.gz
    
    ==> Pouring gcc-5.2.0.mavericks.bottle.2.tar.gz

Do I need to create a "depends_on" for gcc4.9?

DomT4:
> arangodb cannot be built with any available compilers To install this formula, you may need to:
  brew install gcc

Heh. This is a case of a bad message. It's actually saying `Install a compatible gcc please` but well, it doesn't say that. 

> It seems that it has downloaded gcc-5.x

This is also an error on our side. It downloads Homebrew's `gcc` and then goes "Oh wait, I can't use this one" and throws the error message at the top. We know this is kind of confusing and it's on the list of things we need to look at. Apologies :confused:

fceller:
@DomT4 any idea how I can fix this? Is there a way to download gcc49 as a depends_on?

mikemcquaid:
We need to not upgrade until this builds with Clang, I'm afraid. We can apply a patch (if upstreamed) if necessary.

DomT4:
> any idea how I can fix this? Is there a way to download gcc49 as a depends_on?

We really need to look at the GCC 5 failure mechanism but since it only impacts two/three other formulae at this point in time the amount of investment it takes to resolve/test weighed against the impact of making two/three formula a bit clearer on how to install makes it not a massive priority compared to other bug fixes/etc.

If you download and install `homebrew/versions/gcc49` though you'll see as in the gist logs that there's build failure there as well. GCC-4.9 is the compiler I used on the Mavericks VM to generate the gist-logs linked further up.

If it's interesting, since the 4th of this month 33 bottle downloads have been from Mavericks, 81 from Yosemite & 135 from El Capitan.

mikemcquaid:
@DomT4 I dunno, in my take if stuff doesn't build against Clang or GCC 5: boneyard.

fceller:
@mikemcquaid V8 seems to build fine with the newer clang, but not with Maverick (older versions of V8 did). I'll try to figure out, what the problem with GCC 5 is. That seems better than trying to support on outdated clang.

mikemcquaid:
@fceller Good ideas, thanks!

fceller:
@mikemcquaid @DomT4 I've added a patch for GCC 5 - I tests it with Yosemite, but cannot test it with Mavericks.

fceller:
@mikemcquaid The maverick build stopped with

```
==> Downloading https://www.arangodb.com/repositories/Source/ArangoDB-2.7.0.tar.gz



The build has been terminated
```

while the others were able to download the file. I have no idea why.

mikemcquaid:
You can ignore the Travis failure.

Library/Formula/arangodb.rb
```diff
@@ -14,9 +14,17 @@ class Arangodb < Formula
 
   depends_on "go" => :build
   depends_on "openssl"
+  depends_on "readline"
 
   needs :cxx11
 
+  patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

fceller: It has already been submitted and should be available in the next release (ArangoDB 2.7.1), I will add a comment

DomT4:
A bit nervous about including a patch that long in the formula as an `__END__` as inline patches are very breakage prone, but it passed the CI so will pull and see how it goes.

Merged in https://github.com/Homebrew/homebrew/commit/12246e335ecf341e5d6541ba6e7eb1bbe640c0d1. Thank you for fixing this upstream so quickly @fceller. Appreciated!

fceller:
@DomT4 hopefully there will a ArangoDB 2.7.1 next week. Than we can remove the patch.

PR #44842, opened 2015-10-12
Update repo to v1.12.32

Realignist:
I update repo to v1.12.32.

### DIFF
--- a/Library/Formula/repo.rb
+++ b/Library/Formula/repo.rb
@@ -2,6 +2,7 @@ class Repo < Formula
   desc "Repository tool for Android development"
   homepage "https://source.android.com/source/developing.html"
   url "https://raw.githubusercontent.com/android/tools_repo/v1.12.32/repo"
+  version "1.22"
   sha256 "9907c36d63cf5222d73ab270fee5249dc84f6b4580cda03ddd7ac309f11b289e"
   revision 1
 
--- a/Library/Formula/repo.rb
+++ b/Library/Formula/repo.rb
@@ -1,9 +1,9 @@
 class Repo < Formula
   desc "Repository tool for Android development"
   homepage "https://source.android.com/source/developing.html"
-  url "https://raw.githubusercontent.com/android/tools_repo/v1.12.21/repo"
-  version "1.21"
-  sha256 "44d1ac8b948e6154a495469dad05870dc6182712c0610da8d255b5471dd5bcb9"
+  url "https://raw.githubusercontent.com/android/tools_repo/v1.12.32/repo"
+  version "1.12.32"
+  sha256 "9907c36d63cf5222d73ab270fee5249dc84f6b4580cda03ddd7ac309f11b289e"
   revision 1
 
   def install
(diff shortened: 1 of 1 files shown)


Library/Formula/repo.rb
```diff
@@ -1,9 +1,9 @@
 class Repo < Formula
   desc "Repository tool for Android development"
   homepage "https://source.android.com/source/developing.html"
-  url "https://raw.githubusercontent.com/android/tools_repo/v1.12.21/repo"
-  version "1.21"
-  sha256 "44d1ac8b948e6154a495469dad05870dc6182712c0610da8d255b5471dd5bcb9"
+  url "https://raw.githubusercontent.com/android/tools_repo/v1.12.32/repo"
+  version "1.12.32"
+  sha256 "9907c36d63cf5222d73ab270fee5249dc84f6b4580cda03ddd7ac309f11b289e"
   revision 1
```
bfontaine: Please remove this `revision` line.

bfontaine:
We might have an issue with the version here; Homebrew will assume `1.12.32` < `1.21`.

Realignist:
@bfontaine I just removed the version of this file.

Library/Formula/repo.rb
```diff
@@ -1,9 +1,8 @@
 class Repo < Formula
   desc "Repository tool for Android development"
   homepage "https://source.android.com/source/developing.html"
-  url "https://raw.githubusercontent.com/android/tools_repo/v1.12.21/repo"
-  version "1.21"
-  sha256 "44d1ac8b948e6154a495469dad05870dc6182712c0610da8d255b5471dd5bcb9"
+  url "https://raw.githubusercontent.com/android/tools_repo/v1.12.32/repo"
+  sha256 "9907c36d63cf5222d73ab270fee5249dc84f6b4580cda03ddd7ac309f11b289e"
   revision 1
```
bfontaine: Please remove this `revision` line as well.

bfontaine:
@Realignist Please leave it; Homebrew can’t parse it from the URL here. The problem we have here is that the previous version was “`1.21`” and the new one is “`1.12.32`”; Homebrew will assume that the new version is not higher than the previous one and won’t upgrade it.
We might want to keep with the [original version numbering](https://github.com/Homebrew/homebrew/pull/23107/files) and use `version "1.22"` to avoid the issue.

Realignist:
Okay, I understand.

Library/Formula/repo.rb
```diff
@@ -1,9 +1,9 @@
 class Repo < Formula
   desc "Repository tool for Android development"
   homepage "https://source.android.com/source/developing.html"
-  url "https://raw.githubusercontent.com/android/tools_repo/v1.12.21/repo"
-  version "1.21"
-  sha256 "44d1ac8b948e6154a495469dad05870dc6182712c0610da8d255b5471dd5bcb9"
+  url "https://raw.githubusercontent.com/android/tools_repo/v1.12.32/repo"
+  version "1.22"
+  sha256 "9907c36d63cf5222d73ab270fee5249dc84f6b4580cda03ddd7ac309f11b289e"
   revision 1
```
bfontaine: Please remove the line with `revision 1`. Revisions are reset when we upgrade the version; so this line is not necessary anymore.

Realignist:
Sorry for late.

bfontaine:
Thank you for your contribution @Realignist; we appreciate it! :tada:

PR #44843, opened 2015-10-12
Let the user know what exactly was already tapped

srn:
This will make it easier to debug scripts that might do multiple taps etc.

Before:

```sh
❯ brew tap homebrew/services
Warning: Already tapped!
```

After:

```
❯ brew tap homebrew/services
Warning: homebrew/services already tapped!
```

### DIFF
--- a/Library/Homebrew/cmd/tap.rb
+++ b/Library/Homebrew/cmd/tap.rb
@@ -15,7 +15,7 @@ module Homebrew
     else
       user, repo = tap_args
       clone_target = ARGV.named[1]
-      opoo "#{user}/#{repo} already tapped" unless install_tap(user, repo, clone_target)
+      opoo "#{user}/#{repo} already tapped!" unless install_tap(user, repo, clone_target)
     end
   end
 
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/tap.rb
```diff
@@ -15,7 +15,7 @@ def tap
     else
       user, repo = tap_args
       clone_target = ARGV.named[1]
-      opoo "Already tapped!" unless install_tap(user, repo, clone_target)
+      opoo "#{user}/#{repo} already tapped" unless install_tap(user, repo, clone_target)
```
DomT4: It'd be nice to keep the `!` or even a `.` at the end of the sentence, but that nit aside I'm a big :+1: on this change, personally.

bfontaine: :+1: too.

dunn:
Looks good to me.

bfontaine:
Thanks a lot @srn! :tada:

PR #44844, opened 2015-10-12
Add formula for wiki

walle:
Got a request for packaging my tool with homebrew (https://github.com/walle/wiki/issues/10)

The tool is called wiki (https://github.com/walle/wiki) and built with Go.

I hope I got everything right in the formula. I have run `brew audit wiki`, `brew tests` and `brew install wiki && brew test wiki` without any issues.

I have also run `brew audit --strict wiki` and fixed all issues.

Please tell me if there is anything else I have to fix.

### DIFF
--- a/Library/Formula/wiki.rb
+++ b/Library/Formula/wiki.rb
@@ -1,0 +1,32 @@
+require "language/go"
+
+class Wiki < Formula
+  desc "Fetch summaries from MediaWiki wikis, like Wikipedia"
+  homepage "https://github.com/walle/wiki"
+  url "https://github.com/walle/wiki/archive/1.3.0.tar.gz"
+  sha256 "c12edcaed5c9d5e69fc43e77713a68948a399017467d248ba59367b5d458a9e6"
+
+  go_resource "github.com/mattn/go-colorable" do
+    url "https://github.com/mattn/go-colorable.git",
+      :revision => "40e4aedc8fabf8c23e040057540867186712faa5"
+  end
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    wikipath = buildpath/"src/github.com/walle/wiki"
+    wikipath.install Dir["{*,.git}"]
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    cd "src/github.com/walle/wiki" do
+      system "go", "build", "-o", "build/wiki", "cmd/wiki/main.go"
+      bin.install "build/wiki"
+      man1.install "_doc/wiki.1"
+    end
+  end
+
+  test do
+    assert_match "Read more: https://en.wikipedia.org/wiki/Go", shell_output("#{bin}/wiki golang")
+  end
+end


Library/Formula/wiki.rb
```diff
@@ -0,0 +1,32 @@
+require "language/go"
+
+class Wiki < Formula
+  desc "Fetch summaries from MediaWiki wikis, like Wikipedia"
+  homepage "https://github.com/walle/wiki"
+  url "https://github.com/walle/wikiarchive/1.3.0.tar.gz"
```
bfontaine: You missed a slash here: `wikiarchive` → `wiki/archive`.

Library/Formula/wiki.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    cd "src/github.com/walle/wiki" do
+      system "make"
+      bin.install "build/wiki"
+      man1.install "_doc/wiki.1"
+    end
+  end
+
+  test do
+    system bin/"wiki", "-version"
```
bfontaine: Can’t we write a test that does something more substantial than e.g. `-help`/`-version`?

Library/Formula/wiki.rb
walle: Thanks, I have updated the pull request.

Library/Formula/wiki.rb
walle: Do you have an example? 

I have this test for the tool, would it be possible to convert it to a ruby test somehow?

```bash
# Test that standard usage prints a link to page
OUTPUT="$($BIN golang)"
STATUS=$?
if [[ $STATUS -ne 0 ]]; then
  fail 'Did not get success exit code'
  exit 1
fi
echo "$OUTPUT" | grep -q "Read more: https://en.wikipedia.org/wiki/Go"
if [ $? -ne 0 ];then
  fail 'Standard usage did not output link to page'
fi
pass
```

bfontaine: Something like this should work:

```rb
assert_match "Read more: https://en.wikipedia.org/wiki/Go", shell_output("#{bin}/wiki golang")
```

walle: Thanks! I have updated the test.

DomT4:
The only remaining failure is this:
```
==> audit problems
wiki:
```

Given the upstream popularity of this project, even with being very new, I'm open to some flexibility on the requirement there unless other maintainers are :-1: on the idea.

bfontaine:
> Given the upstream popularity of this project, even with being very new, I'm open to some flexibility on the requirement

I’m :+1: on that because it’s popular _and_ we have a stable release.

walle:
I noticed one thing, and will update the pull request.

The formula uses `make [build]`, which does a `go get` for a dependency. I think it might be better to do a `go build` instead. Will update now.

walle:
I have updated the pull request now.

DomT4:
Nobody screamed murder so merged in https://github.com/Homebrew/homebrew/commit/63260c4ef310b9390465b526b5d60a1ea7f40727. Thank you for your contribution to Homebrew @walle; we appreciate it! :smiley_cat:

walle:
Thanks, that is great!

Have a nice day!

PR #44848, opened 2015-10-12
Blas dependency

davydden:
this is a result of a big disscussion about [linuxbrew on a cluster](https://github.com/Homebrew/linuxbrew/issues/504) with @Rombur @dpo @tjhei @jppelteret @dctrud @sjackman  and two PRs: [original by @tdsmith](https://github.com/Homebrew/linuxbrew/pull/511) and my further work [here](https://github.com/Homebrew/linuxbrew/pull/592).

By introducing `:blas` dependency we:

1. Avoid all `depends_on openblas => :optional`. If someone wants to compile everything with openblas, it's one place to set this;
2. Sometimes we need to use `veclibfort` to wrap `-framework Accelerate` to use single precision or complex valued BLAS Fortran routines. It's is now also done automatically inside the dependency. So in many formulae we will be able cut down something like `if OSX? use veclibfort else if "openblas" use openblas else use "-llapack -lblas"`.
3. As far as blas is concered, formulae in Homebrew-science will have the same code be it OS-X or Linuxbrew. 
4. Finally, the whole infrastructure is ready to use BLAS natively provided by computer vendors or other alternatives such as Atlas and most notably **Math-Kernel-Library (MKL)**, which recently [became free](https://software.intel.com/en-us/articles/free_mkl) and hopefully will be soon available for OSX. 

Summary of `:blas` usage cases which were tested on  a) **CentOS cluster + MKL**; b) **Ubuntu 14.04 + Openblas**;  b) **Ubuntu 14.04 + system blas** are:
1. library names (space/semicolon separated) + dir: [Hypre](https://github.com/davydden/homebrew-science/blob/mkl_dealii/hypre.rb#L39-L45), [Trilinos](https://github.com/davydden/homebrew-science/blob/mkl_dealii/trilinos.rb#L115-L120)
2. ldflags : [Superlu_dist](https://github.com/davydden/homebrew-science/blob/mkl_dealii/superlu_dist.rb#L31), [scalapack](https://github.com/davydden/homebrew-science/blob/mkl_dealii/scalapack.rb#L28-L29), [mumps](https://github.com/davydden/homebrew-science/blob/mkl_dealii/mumps.rb#L107-L108), [p4est](https://github.com/davydden/homebrew-science/blob/mkl_dealii/p4est.rb#L32)
3. full path (space/semicolon separated) : [Petsc](https://github.com/davydden/homebrew-science/blob/mkl_dealii/petsc.rb#L92-L93), [deal.II](https://github.com/davydden/homebrew-science/blob/mkl_dealii/dealii.rb#L58-L62)

@tdsmith [proposed](https://github.com/Homebrew/linuxbrew/pull/511#issuecomment-140774803) to name it `:blas_lapack`, which is cleaner. I have not changed the name yet as there are some static functions of `BlasRequirement` class. One can also discuss the name of the tag (currently `:fortran`), perhaps something like `:fortran_single` (single precision) will be more descriptive. But this are minor changes easy to make if desired. I am open to add further in-code comments if needed.

ps. 

here is a nice summary article about BLAS and LAPACK http://glennklockwood.blogspot.de/2014/02/quantum-espresso-compiling-and-choice.html

>BLAS is the lowest-level library and provides subroutines that do basic vector operations.  Netlib provides a reference implementation of BLAS written in Fortran, but the big idea behind BLAS is to allow hardware vendors to provide highly tuned versions of the BLAS subroutines that obviate the need for application developers to worry about optimizing their linear algebra for every possible computer architecture on which the application might run.  This motivation is also what gave rise to the MPI standard, but unlike MPI, BLAS is not an actual standard.
>
>LAPACK builds upon BLAS and provides higher-level matrix operations such as diagonalization (i.e., solving for eigenvectors and eigenvalues) and inversion.  BLAS and LAPACK seem to be bundled together when actually implemented (e.g., IBM ESSL and Intel MKL both provide both optimized BLAS and LAPACK), but they provide two distinct layers of abstracting the mathematical complexity away from application developers.
>

### DIFF
--- a/Library/Homebrew/requirements/blas_requirement.rb
+++ b/Library/Homebrew/requirements/blas_requirement.rb
@@ -1,0 +1,47 @@
+require "requirement"
+
+class BlasRequirement < Requirement
+  fatal true
+  default_formula "openblas"
+
+  # This ensures that HOMEBREW_BLAS_CFLAGS and HOMEBREW_BLAS_LDFLAGS
+  # are always set. It does _not_ add them to CFLAGS or LDFLAGS; that
+  # should happen inside the formula.
+  env do
+    if @satisfied_result
+      ENV["HOMEBREW_BLAS_CFLAGS"] ||= ""
+      ENV["HOMEBREW_BLAS_LDFLAGS"] ||= "-lblas -llapack"
+    else
+      ENV["HOMEBREW_BLAS_CFLAGS"] = "-I#{Formula["openblas"].opt_include}"
+      ENV["HOMEBREW_BLAS_LDFLAGS"] = "-L#{Formula["openblas"].opt_lib} -lopenblas"
+    end
+  end
+
+  satisfy :build_env => true do
+    cflags = ENV["HOMEBREW_BLAS_CFLAGS"] || ""
+    ldflags = ENV["HOMEBREW_BLAS_LDFLAGS"] || "-lblas -llapack"
+    success = nil
+    Dir.mktmpdir do |tmpdir|
+      tmpdir = Pathname.new tmpdir
+      (tmpdir/"blastest.c").write <<-EOS.undent
+        double cblas_ddot(const int, const double*, const int, const double*, const int);
+        int main() {
+          double x[] = {1.0, 2.0, 3.0}, y[] = {4.0, 5.0, 6.0};
+          cblas_ddot(3, x, 1, y, 1);
+          return 0;
+        }
+      EOS
+      success = system "#{ENV["CC"]} #{cflags} #{tmpdir}/blastest.c -o #{tmpdir}/blastest #{ldflags}",
+                :err => "/dev/null"
+    end
+    if !success
+      opoo "BLAS not configured"
+      puts <<-EOS.undent
+        Falling back to brewed openblas. If you prefer to use a system BLAS,
+        please set HOMEBREW_BLAS_CFLAGS and HOMEBREW_BLAS_LDFLAGS to correct
+        values.
+      EOS
+    end
+    success
+  end
+end
--- a/Library/Homebrew/requirements/blas_requirement.rb
+++ b/Library/Homebrew/requirements/blas_requirement.rb
@@ -25,13 +25,27 @@ class BlasRequirement < Requirement
     end
   end
 
+  def self.ldflags(blas_lib,blas_names)
+    res  = blas_lib != "" ? "-L#{blas_lib} " : ""
+    res += blas_names.split(";").map { |word| "-l#{word}" }.join(" ")
+    return res
+  end
+
+  def self.cflags(blas_inc)
+    return blas_inc != "" ? "-I#{blas_inc}"  : ""
+  end
+
+  def self.full_path(blas_lib,blas_names,separator)
+    exten = (OS.mac?) ? "dylib" : "so"
+    return blas_names.split(";").map { |word| "#{blas_lib}/lib#{word}.#{exten}" }.join(separator)
+  end
+
   satisfy :build_env => true do
     blas_names = ENV["HOMEBREW_BLASLAPACK_NAMES"] || "blas;lapack"
     blas_lib   = ENV["HOMEBREW_BLASLAPACK_LIB"]   || ""
     blas_inc   = ENV["HOMEBREW_BLASLAPACK_INC"]   || ""
-    cflags     = blas_inc != "" ? "-I#{blas_inc}"  : ""
-    ldflags    = blas_lib != "" ? "-L#{blas_lib} " : ""
-    ldflags   += blas_names.split(";").map { |word| "-l#{word}" }.join(" ")
+    cflags     = BlasRequirement.cflags(blas_inc)
+    ldflags    = BlasRequirement.ldflags(blas_lib,blas_names)
     # MKL BLAS may want to link against libpthread (e.g. pthread_mutex_trylock)
     # and we most likely need basic math (atan2, sin, etc) from libm
     # Adding both won't make much harm:
--- a/Library/Homebrew/requirements/blas_requirement.rb
+++ b/Library/Homebrew/requirements/blas_requirement.rb
@@ -10,14 +10,29 @@ class BlasRequirement < Requirement
   # On Linux we can always fallback to openblas
   default_formula "openblas"   unless OS.mac?
 
+  def initialize(tags = [])
+    # if we are on OSX and need fortran and veclibfort is installed
+    # by default try it. Otherwise do standard "blas;lapack"
+    if tags.include?(:fortran) && OS.mac? && Formula["veclibfort"].installed?
+      @default_names = "veclibfort"
+      @default_lib   = "#{Formula["veclibfort"].opt_lib}"
+      @default_inc   = "#{Formula["veclibfort"].opt_include}"
+    else
+      @default_names = "blas;lapack"
+      @default_lib   = ""
+      @default_inc   = ""
+    end
+    super(tags)
+  end
+
   # This ensures that HOMEBREW_BLASLAPACK_NAMES, HOMEBREW_BLASLAPACK_LIB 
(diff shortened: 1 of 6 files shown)


Library/Homebrew/requirements/blas_requirement.rb
```diff
@@ -0,0 +1,124 @@
+require "requirement"
+
+class BlasRequirement < Requirement
+  fatal true
+  # on OSX -lblas and -llapack should work OOB.
+  # the only case when test should fail is when we
+  # need BLAS with single precision or complex with Fortran =>
+  # use veclibfort
+  default_formula "veclibfort" if OS.mac?
```
DomT4: You're going to hit some issues with this as it doesn't know what `veclibfort` is because the CIs work off the assumption not everyone is going to have `homebrew/science` tapped. Same situation with `openblas`.

Until something happens on that the CI isn't going to be able to give you a huge amount of feedback I'm afraid.

davydden: do you want me to temporary add `openblas` and `veclibfort` to this PR?

davydden:
btw, i don't think you need those `blastest_xxx` formulae, they are here only to illustrate how the dependency work (in addition to those real-life examples i tried on `Homebrew/science`). I can wipe them out when squashing...

Library/Homebrew/requirements/blas_requirement.rb
dpo: Is there such a thing as per-tap requirements? I.e., can taps extend Homebrew?

mikemcquaid: Also: Homebrew's core code assumes it's a Mac so this should be unconditional.

mikemcquaid:
I think this should maybe live in Homebrew science as a requirement in there instead of Homebrew core.

davydden:
But `Homebrew science` is not a fork of `Homebrew`? How can one add a `:blas` requirement there? 

Additionally there must be some formulae in Homebrew core which use `-framework Accelerate` (link against system's BLAS/Lapack). It would make sense to use this dependency there so that it works automatically on `Linuxbrew` without extra hacking.

mikemcquaid:
> But Homebrew science is not a fork of Homebrew? How can one add a :blas requirement there?

It can be added to individual formulae or `require`d in formulae and made a separate file.

> Additionally there must be some formulae in Homebrew core which use -framework Accelerate (link against system's BLAS/Lapack). It would make sense to use this dependency there so that it works automatically on Linuxbrew without extra hacking.

That code should live in Linuxbrew, not Homebrew.

davydden:
> It can be added to individual formulae or required in formulae and made a separate file.

i mean how can one add `depends_on :blas` requirement to Homebrew-sicence? The whole point is to incapsulate it nicely and beautifully inside and use throughout in each formula.

davydden:
p.s. above i like to usage in homebrew-science not because it applies to homebrew science only, but because i test what i am familiar with and what use on day-to-day basis for research.

davydden:
I suppose one does not see immediately which formulae in the main Homebrew repository link against system's blas because the configuration scripts do not specify blas explicitly. So most likely it's automatically detected where need. For example, looking at [viennacl](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/viennacl.rb#L21) one could guess that it does link against system's blas. Another one which certainly use blas is [grassrb](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/grass.rb#L62).

mikemcquaid:
Yep, we don't specify dependencies for things provided by the system.

davydden:
btw, just checked with [hashdist](https://github.com/hashdist/hashdist), it seems they often have `openblas` hardcoded as dependencies, e.g. in [trilnos](https://github.com/hashdist/hashstack/blob/master/pkgs/trilinos.yaml#L35). 

My naïve view to this PR is that it makes Homebrew (`Homebrew core`, `Homebrew science` and `Linuxbrew`) more attractive and flexible (see argumentation in the first post).

xu-cheng:
Sorry being busy lately to check this PR. 

I'm OK to put this inside Homebrew as it offers an abstract on blas dependency. In the mean time, I think can improve the implementation by using Superenv(which is also consistent with other requirement). Instead of creating `ldflags/cflags` methods. You should update `ENV.cflags` and `ENV.ldflags` directly inside `env` block. You can see [`JavaRequirement`](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/requirements/java_requirement.rb#L10-L20) for example.

davydden:
@xu-cheng I would keep static functions as there are usage cases where `ldflags` for blas/lapack are provided in a different way, see for example [scalapack](https://github.com/davydden/homebrew-science/blob/mkl_dealii/scalapack.rb#L27-L28), [superlu_dist](https://github.com/davydden/homebrew-science/blob/mkl_dealii/superlu_dist.rb#L31-L37), [mumps](https://github.com/davydden/homebrew-science/blob/mkl_dealii/mumps.rb#L107-L108) and [p4est](https://github.com/davydden/homebrew-science/blob/mkl_dealii/p4est.rb#L32). I think whether or not a given software use environmen flags `ENV.cflags` and can pick up BLAS/LAPACK from there is very much individual. Some, perhaps, would fail to configure even before compiling if you do not provide BLAS and LAPACK in the form of expected input.

Perhaps in the light if this complexity it is better to avoid adding to `cflags` and `ldflags` so that there is no confusion on how to use the dependency.

xu-cheng:
You can use `ENV.ldflags` in formula instead of `BlasRequirement.ldflags`. Let's keep formula code simple.

davydden:
> You can use ENV.ldflags in formula instead of BlasRequirement.ldflags. Let's keep formula code simple.

sure, but what if there is something more in `ENV.ldflags`, added from other dependencies?

davydden:
maybe `ENV.blas_ldflags`?

davydden:
together with `ENV.blas_names` (an array), `ENV.blas_lib_dir`, `ENV.blas_inc_dir` and `ENV.blas_full_path_to_libs` (an array). See the different usage scenarios in the first post.

mikemcquaid:
> I'm OK to put this inside Homebrew as it offers an abstract on blas dependency. In the mean time, I think we can improve the implementation by using Superenv(which is also consistent with other requirement). Instead of creating ldflags/cflags methods. You should update ENV.cflags and ENV.ldflags directly inside env block. You can see JavaRequirement for example.

@xu-cheng Which formulae in Homebrew core would need this?

xu-cheng:
You are right. It seems that this is most used in `science` and `python` taps.

davydden:
It can't go to linuxbre because this can be used both on mac and linux. It cant go to homebre-science because it's not a fork of homebrew. The only place this flexibility can be placed is homebrew core. I do not see any alternatives to the implementation.

Fine, u dont plan to use it much in formulae within the core, i understand tgat choice. But it is super usefull in dependent projects (linuxbre/homebre science).

mikemcquaid:
> It cant go to homebre-science because it's not a fork of homebrew.

It can but just not as `depends_on :blas => :fortran`. It'd need to be `require`d manually and then `depends_on BlasRequirement => :fortran`

davydden:
U mean to create a formula "BlasRequirement" ? If so, how can one do the same type of tests and fallback to optional dependency if tests fail?

davydden:
If it stays as a class derived from requirement, how can one make homebrew see it?

mikemcquaid:
I've repeated that several times.

dpo:
> It can but just not as depends_on :blas => :fortran. It'd need to be required manually and then depends_on BlasRequirement => :fortran

That wouldn't be so bad. It seems like a good way to stress test it and improve it over time.

davydden:
What i was asking is something like:
Write require_relative "blas_requirement" in target formula, assuming that blas_requirement.rb is in the Homebrew-science tap. 

But i figured it out already, will continue with another PR to Homebrew-science...

PR #44854, opened 2015-10-12
puddletag: Bugfix.

chincheta0815:


Library/Formula/puddletag.rb
```diff
@@ -41,6 +41,12 @@ def install
     end
 
     cp_r buildpath/"source/.", buildpath
+    inreplace buildpath/"puddlestuff/puddleobjects.py" do |s|
+      s.gsub! /QPalette.Window/, "pal.window()"
```
bfontaine: Why is this necessary?

Library/Formula/puddletag.rb
```diff
@@ -41,6 +41,12 @@ def install
...
+    end
+    inreplace buildpath/"puddlestuff/tagmodel.py" do |s|
+      s.gsub! /amarok -p/, "open -a iTunes"
```
bfontaine: You should raise this issue to upstream; they could use the correct command depending on the platform (`from sys import platform` then test if e.g. `platform == "darwin"` then use iTunes).

Library/Formula/puddletag.rb
chincheta0815: The puddletag version 1.0.5 (which is the actual tagged one in the upstream github repo) does only support an older version of pyqt (4.11.3). Since the pyqt formula got an update, this inreplace is necessary for puddletag in version 1.0.5 to work (for compatibility).

The master for puddletag in the github includes exactly this fix.

Once puddletag will get a new version tag, this line can hopefully be deleted.

If there is a method to install a specific version of pyqt via brew please tell me.

Library/Formula/puddletag.rb
chincheta0815: For the second inreplace (itunes) I will ask them.
How about the first?

Library/Formula/puddletag.rb
bfontaine: Thanks. Please use upstream’s patch with a short comment saying why it’s necessary and that we can remove it at the next upgrade. Add this block above `def install`:

```rb
# Upstream commit to fix an issue with PyQT 4.11.4. Remove on next version.
patch do
  url "https://github.com/keithgg/puddletag/commit/489acd2ee62eb5fbff95f8220dc8958c14871931.patch"
  sha256 "fce0cfce4d4477cde4827a0a4d3ef74fbabf630ada2d0cf035cf155a17c37a68"
end
```

chincheta0815:
All requested changes are made:
- patch do added.
- inreplace added to upstream.

Both issues are already implemented into upstream master, but not yet versioned. Will be fixed in formula with next upstream release.

bfontaine:
Sorry, just one missing thing: could you add a line at the top of the formula, just below the checksum, that say `revision 1`. It’ll force Homebrew to re-generate the bottles and provide the fix to everyone.

The formula should then look like this:

```rb
  url "..."
  sha256 "..."
  revision 1

  head "..."
```

chincheta0815:
I added the 'revision 1' statement.

Just for my understanding: When I change the formula without changing the upstream version (meaning I just add patches etc.) I set a revision which I have to increment until there is a version bump of the upstream software.

After a version change in the upstream software I have to delete the revision statement (and start incementing a revision again in case of bugfixes...).

Is that right?

mikemcquaid:
Yup!

chincheta0815:
Okay.
Thanks for the info. I will remind that with the next version update.
For now I hope the formula revision is ready for a merge now.

bfontaine:
Yep. Thank you for your contribution to Homebrew @chincheta0815; we appreciate it! :smile:

PR #44855, opened 2015-10-12
pgbouncer: use secure homepage url

vszakats:


bfontaine:
Thanks again @vszakats!

PR #44856, opened 2015-10-12
innoextract: use secure urls

vszakats:


bfontaine:
Thanks again @vszakats!

PR #44866, opened 2015-10-12
at-spi2-core 2.18.1

tschoonj:


bfontaine:
Thank you again @tschoonj! :smile:

PR #44867, opened 2015-10-12
at-spi2-atk 2.18.1

tschoonj:


bfontaine:
Thank you for your contribution @tschoonj; we really appreciate it! :tada:

PR #44873, opened 2015-10-12
emp 0.10.0 (new formula)

zabawaba99:
Add [empire](https://github.com/remind101/empire) cli tool

### DIFF
--- a/Library/Formula/emp.rb
+++ b/Library/Formula/emp.rb
@@ -1,0 +1,49 @@
+require "language/go"
+
+class Emp < Formula
+  desc "CLI for Empire."
+  homepage "https://github.com/remind101/empire"
+  url "https://github.com/remind101/empire/archive/v0.10.0.tar.gz"
+  sha256 "0f49543e25d44a05522f8e2cbb4afbb5b08153767620e304e371e13b8f574508"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    (buildpath/"src/github.com/remind101/").mkpath
+    ln_s buildpath, buildpath/"src/github.com/remind101/empire"
+
+    system "go", "build", "-o", bin/"emp", "./src/github.com/remind101/empire/cmd/emp"
+  end
+
+  test do
+    require "webrick"
+    require "utils/json"
+
+    server = WEBrick::HTTPServer.new :Port => 8035
+    server.mount_proc "/apps/foo/releases" do |_req, res|
+      resp = {
+        "created_at" => "2015-10-12T0:00:00.00000000-00:00",
+        "description" => "my awesome release",
+        "id" => "v1",
+        "user" => {
+          "id" => "zab",
+          "email" => "zab@waba.com",
+        },
+        "version" => 1,
+      }
+      res.body = Utils::JSON.dump([resp])
+    end
+
+    Thread.new { server.start }
+
+    begin
+      ENV["EMPIRE_API_URL"] = "http://127.0.0.1:8035"
+      assert_match /v1  zab  Oct 1(1|2|3) \d\d:00  my awesome release/,
+        shell_output("#{bin}/emp releases -a foo").strip
+    ensure
+      server.shutdown
+    end
+  end
+end


Library/Formula/emp.rb
```diff
@@ -0,0 +1,27 @@
...
+    emppath = buildpath/"src/github.com/remind101/emp"
+    emppath.install Dir["{*,.git}"]
+
+    cd emppath do
+      system "go", "build", "-o", bin/"emp"
+    end
+  end
+
+  test do
+    assert_equal "#{version}",
```
zabawaba99: I know that according to the [testing guidelines](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#add-a-test-to-the-formula) one should try and avoid testing the formula with `--version` however this CLI typically operates against an HTTP server. 

If it's preferred, I could write a mock http server that responds to the CLI's requests and test a "simple" route such as listing apps.

bfontaine: It’s ok to fetch small resources for testing, see e.g. `wiki.rb`, which we’ll merge very soon: https://github.com/Homebrew/homebrew/pull/44844/files#diff-b1eede01b0006877d01481b90c0e93e1R29

zabawaba99: What do you think about something like this https://gist.github.com/zabawaba99/fceca5a1cff178738ac9 ?

bfontaine: Shorter version: https://gist.github.com/bfontaine/64c9c82a156da46d7bb6

zabawaba99: If I run the tests twice, it fails because the port is still bound. 

```
~ brew test emp
Testing emp
31824
==> go run /private/tmp/emp20151013-31812-17nf07w/test-server.go
==> EMPIRE_API_URL=http://127.0.0.1:8035 /usr/local/Cellar/emp/0.9.1/bin/emp releases -a foo
~ lsof | grep 8035
test-serv 31829 stevenberlanga    3u    IPv6 0xc5f2310168a54eef         0t0      TCP *:8035 (LISTEN)
~ ps aux | grep 31829
stevenberlanga  31871   0.0  0.0  2434836    768 s001  R+   10:00AM   0:00.00 grep 31829
stevenberlanga  31829   0.0  0.0 573394348   3548 s001  S     9:58AM   0:00.01 /var/folders/vp/d2lk3_ms5dvb4y69wdy4z5800000gn/T/go-build411514785/command-line-arguments/_obj/exe/test-server
```

I tried using [Process.spawn](http://ruby-doc.org/core-2.2.0/Process.html#method-c-spawn) instead but still ran into the same issue

xu-cheng: Noted you should not use `go` to create HTTP server during test, because `go` is a build dependence for this formula and may not be available during the test. 

You could consider to use `webrick` which is a standard ruby library.

bfontaine: Thanks; let’s keep it simple and fallback on your simple `version` test.

Library/Formula/emp.rb
```diff
@@ -0,0 +1,27 @@
...
+      shell_output("EMPIRE_API_URL=http://foo #{bin}/emp version").strip
```
bfontaine: Does `EMPIRE_API_URL` need to be defined in order to print the version?

zabawaba99: Aye. I opened up an issue against the project.

Library/Formula/emp.rb
bfontaine: Ah, right @xu-cheng.

zabawaba99: Fair point. Looking into webrick now.

zabawaba99: Updated the tests using a webrick server. Thanks for the help

bfontaine:

Library/Formula/emp.rb
```diff
@@ -0,0 +1,41 @@
+require "language/go"
+
+class Emp < Formula
+  desc "CLI for Empire."
+  homepage "https://github.com/remind101/emp"
+  url "https://github.com/remind101/emp/archive/v0.9.1.tar.gz"
+  version "0.9.1"
```
bfontaine: You can remove this line; Homebrew will extract the version from the URL.

Library/Formula/emp.rb
```diff
@@ -0,0 +1,41 @@
...
+    emppath.install Dir["{*,.git}"]
+
+    cd emppath do
+      system "go", "build", "-o", bin/"emp"
+    end
+  end
+
+  test do
+    require "webrick"
+    require "json"
```
bfontaine: You probably don’t need to `require "json"` here.

Library/Formula/emp.rb
zabawaba99: I forgot about that. I removed the line.

Library/Formula/emp.rb
zabawaba99: You're right. I think I had it there because I was trying to do something different and ended up going with a simpler approach. I removed the `require`

Library/Formula/emp.rb
```diff
@@ -0,0 +1,39 @@
...
+    server = WEBrick::HTTPServer.new :Port => 8035
+    server.mount_proc "/apps/foo/releases" do |_req, res|
+      res.body = '[{"created_at":"2015-10-12T18:49:16.69829901-04:00","description":"my awesome release","id":"v1","user":{"id":"zab","email":"zab@waba.com"},"version":1}]'
+    end
+
+    Thread.new { server.start }
+
+    begin
+      assert_equal "v1  zab  Oct 12 18:49  my awesome release",
+        shell_output("EMPIRE_API_URL=http://127.0.0.1:8035 #{bin}/emp releases -a foo").strip
```
bfontaine: Use `ENV["EMPIRE_API_URL"] = "http://127.0.0.1:8035"` to set an environment variable.

zabawaba99: Done

Library/Formula/emp.rb
```diff
@@ -0,0 +1,40 @@
...
+    server = WEBrick::HTTPServer.new :Port => 8035
+    server.mount_proc "/apps/foo/releases" do |_req, res|
+      res.body = '[{"created_at":"2015-10-12T00:00:00.00000000-04:00","description":"my awesome release","id":"v1","user":{"id":"zab","email":"zab@waba.com"},"version":1}]'
+    end
+
+    Thread.new { server.start }
+
+    begin
+      ENV["EMPIRE_API_URL"]="http://127.0.0.1:8035"
+      assert_equal "v1  zab  Oct 12 00:00  my awesome release",
```
dunn: :scream_cat: 
```
<"v1  zab  Oct 12 00:00  my awesome release"> expected but was
<"v1  zab  Oct 11 21:00  my awesome release">.

Comment (bfontaine) [inline Library/Formula/emp.rb]: Maybe the test result depends on the timezone?

Comment (zabawaba99) [inline Library/Formula/emp.rb]: I believe the `emp` tool prints the time out in whatever timezone you are in. What about doing an assert match with a regex like:

```
/v1  zab  Oct 1(1|2|3) \d\d:00  my awesome release/
```

That way it doesn't matter what timezone you are in.

Comment (dunn) [inline Library/Formula/emp.rb]: Sounds good!

Comment (zabawaba99) [inline Library/Formula/emp.rb]: I pushed the code. Seems like the tests are passing now.

File: Library/Formula/emp.rb [hunk 7]
```diff
@@ -0,0 +1,40 @@
...
+  url "https://github.com/remind101/emp/archive/v0.9.1.tar.gz"
+  sha256 "9619ffd25856988d92d39b31d9e7ed6edacbd85543ff6e60300954df7c16135d"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    emppath = buildpath/"src/github.com/remind101/emp"
+    emppath.install Dir["{*,.git}"]
```
mikemcquaid: Can you do `ln_s buildpath, buildpath/"src/github.com/remind101/emp"` instead (like most of our other formulae)? Thanks!

zabawaba99: Would you mind giving me a spot on how the change would be exactly? I'm not too sure how the change changes the build process. I made the change and the formula fails to install.

mikemcquaid: Sounds like a bug that would be worth reporting upstream?

bfontaine:

bfontaine:
Seems like upstream moved the project under https://github.com/remind101/empire in the meantime and that the 0.10.0 release is out: https://github.com/remind101/empire/releases/tag/v0.10.0

zabawaba99:
I just took a look at the new release version. It look like they also moved away from Godeps. Do you want to close this PR and I'll create a new one with the updated formula?

bfontaine:
@zabawaba99 You can update this PR by pushing new commits and we can rename it as well.

zabawaba99:
@bfontaine will do.

zabawaba99:
@bfontaine It took a bit, but I was finally able to get this to work. However, I was having some issues that I could use some help on.

1 - This project use's [Go's Experimental Vendoring](https://blog.gopheracademy.com/advent-2015/vendor-folder/). I tried building with go1.6 as well as setting the `GO15VENDOREXPERIMENT` and it would not pick up the dependencies in the vendor directory.

2 - My workaround to this issue was to just to copy the vendor directory into the `GOPATH` directory. Is there a cleaner way to do this?

DomT4:
> 1 - This project use's Go's Experimental Vendoring. I tried building with go1.6 as well as setting the GO15VENDOREXPERIMENT and it would not pick up the dependencies in the vendor directory.

Note that the vendor experiment is enabled by default since Go 1.6.

zabawaba99:
> Note that the vendor experiment is enabled by default since Go 1.6.

I initially tried building with 1.5 and the variable and then moved to 1.6 without it and both resulted the same.

bfontaine:
I was able to build it using `"./src/github.com/remind101/empire/cmd/emp"` instead of `"./cmd/emp"` in the `system "go", "build"` step.

zabawaba99:
Was able to reproduce locally as well. Thanks! Last question would be, since this requires a minimum go version of 1.5, how would I specify that formula? I tried doing
`go => ["1.5+", :build]` that I got for some other formulas but that downloaded the latest go version every time.

dunn:
`depends_on "go"` sets a hard dependency on Homebrew's Go, which is 1.6, so you don't need to worry about specifying a minimum version.

zabawaba99:
Just pushed an updated version. Thanks for all the help on this formula.

Library/Formula/emp.rb
```diff
@@ -0,0 +1,39 @@
...
+  sha256 "0f49543e25d44a05522f8e2cbb4afbb5b08153767620e304e371e13b8f574508"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    mkdir_p buildpath/"src/github.com/remind101/"
+    ln_s buildpath, buildpath/"src/github.com/remind101/empire"
+    Language::Go.stage_deps resources, buildpath/"src"
```
DomT4: Don't seem to be any resources to stage here?

Library/Formula/emp.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    system "go", "build", "-o", bin/"emp", "./src/github.com/remind101/empire/cmd/emp"
+  end
+
+  test do
+    require "webrick"
+
+    server = WEBrick::HTTPServer.new :Port => 8035
+    server.mount_proc "/apps/foo/releases" do |_req, res|
+      res.body = '[{"created_at":"2015-10-12T0:00:00.00000000-00:00","description":"my awesome release","id":"v1","user":{"id":"zab","email":"zab@waba.com"},"version":1}]'
```
DomT4: Any way we can break this onto two lines so it doesn't stretch the formula out so much?

Library/Formula/emp.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    Thread.new { server.start }
+
+    begin
+      ENV["EMPIRE_API_URL"]="http://127.0.0.1:8035"
```
DomT4: `ENV["EMPIRE_API_URL"] = "http://127.0.0.1:8035"` is a little more readable.

Library/Formula/emp.rb
zabawaba99: Sure thing. What do you think of something like

```
res.body = [{
        "created_at" => "2015-10-12T0:00:00.00000000-00:00",
        "description" => "my awesome release",
        "id" => "v1",
        "user" => {
          "id" => "zab",
          "email" => "zab@waba.com",
        },
        "version" => 1,
      }].to_json
```

DomT4: If it works :+1:.

Library/Formula/emp.rb
zabawaba99: Removed

Library/Formula/emp.rb
zabawaba99: :+1:

Library/Formula/emp.rb
zabawaba99: Just pushed the change

Library/Formula/emp.rb
dunn: Looks like it's still there?

zabawaba99: I added it again because I changed the test as per https://github.com/Homebrew/homebrew/pull/44873#discussion_r53568529 . If I remember correctly, this comment was made because `to_json` was not being used in the tests.

zabawaba99:
It seems the [test](https://travis-ci.org/Homebrew/homebrew/jobs/110775021#L485) are failing because it can't load the `json` gem. Should I not be using it in the tests?

Library/Formula/emp.rb
DomT4: If you need it you should be able to use `require "utils/json"` which should work on the older Ruby version.

zabawaba99: Thanks. I pushed the change.

dunn:
Just some minor audit failures now:
```
== /usr/local/Library/Formula/emp.rb ==
C: 27:  9: Use 2 spaces for indentation in a hash, relative to the first position after the preceding left parenthesis.
C: 35:  7: Indent the right brace the same as the first position after the preceding left parenthesis.

1 file inspected, 2 offenses detected

zabawaba99:
> Just some minor audit failures now:

I had to update the `.rubocop.yml` locally in order to see the issues. Just pushed a fix.

mikemcquaid:

zabawaba99:
I know this PR was a long one. Thanks for all your help in fixing it up!

PR #44874, opened 2015-10-12
memcached: Add user flag to plist template

kabel:
This allows `brew services` to launch memcached from the root user. It is safely ignored if run as a non-privileged user.

### DIFF
--- a/Library/Formula/memcached.rb
+++ b/Library/Formula/memcached.rb
@@ -47,6 +47,8 @@ class Memcached < Formula
         <string>#{opt_bin}/memcached</string>
         <string>-l</string>
         <string>localhost</string>
+        <string>-u</string>
+        <string>nobody</string>
       </array>
       <key>RunAtLoad</key>
       <true/>


kabel:
Bump. Any chance this PR can get some love?

DomT4:
@BrewTestBot test this please

mikemcquaid:
I think this should probably be done in your local configuration as this plist is designed for the non-root case.

kabel:
That's fair, but this would make it work in both cases.

mikemcquaid:
@kabel I think I'd rather avoid the formula expanding for that case. I'll let other maintainers chip in, though.

DomT4:
I don't have a particular objection here really.

apjanke:
We've enhanced brew services and plist support since this was submitted. Is this still relevant?

kabel:
It is still relevant, but it doesn't seem like there's any interest in supporting this use-case, so I'll close.

PR #44875, opened 2015-10-12
watchman 3.9.0

wez:
Tested via:

```
brew install watchman
brew audit watchman
brew test watchman
```

Note: I'll update this to point to the final 3.9.0 tag after we see the results from the brew CI build.

I'm using an API introduced in 10.10 so I wouldn't be surprised if we see a failure on the older releases.
Assuming that it builds fine on all systems, I'll add the tag and update.


bfontaine:
Thank you for your contribution @wez; we appreciate it! :smile:

PR #44876, opened 2015-10-13
git-imerge: allow completion (for HEAD builds)

rhunter:
git-imerge has provided bash-completion for a over a year now, but the project hasn't made a numbered version release in that time.

We can take advantage of it in the mean time (and drop the condition when the next version of git-imerge is formally released).


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/5080c9b2ecaeaba97e8ff6a81e778917705f2a6b. Thank you for your contribution to Homebrew @rhunter! :smiley_cat:

PR #44877, opened 2015-10-13
lcov 1.12

nsnutella:
Since update to xcode 7 do not works with `lcov` 1.11, he have updated `lcov` to the latest version so we can test coverage to our project.

A new patch data has created and applied to the files `Makefile` and `install.sh`.

### DIFF
--- a/Library/Formula/lcov.rb
+++ b/Library/Formula/lcov.rb
@@ -1,8 +1,8 @@
 class Lcov < Formula
   desc "Graphical front-end for GCC's coverage testing tool (gcov)"
   homepage "http://ltp.sourceforge.net/coverage/lcov.php"
-  url "https://downloads.sourceforge.net/ltp/lcov-1.11.tar.gz"
-  sha256 "c282de8d678ecbfda32ce4b5c85fc02f77c2a39a062f068bd8e774d29ddc9bf8"
+  url "https://downloads.sourceforge.net/ltp/lcov-1.12.tar.gz"
+  sha256 "b474e49c6c962754063b3be97a757a2ba9e7a455f0aea612863bf67e9b8b8ea7"
 
   head "https://github.com/linux-test-project/lcov.git"
 
@@ -19,14 +19,14 @@ class Lcov < Formula
   def install
     inreplace %w[bin/genhtml bin/geninfo bin/lcov],
       "/etc/lcovrc", "#{prefix}/etc/lcovrc"
-    system "make", "PREFIX=#{prefix}", "install"
+    system "make", "PREFIX=#{prefix}", "BIN_DIR=#{bin}", "MAN_DIR=#{man}", "install"
   end
 end
 
 __END__
---- lcov-1.8/bin/install.sh~	2010-01-29 19:14:46.000000000 +0900
-+++ lcov-1.8/bin/install.sh	2010-04-16 21:40:57.000000000 +0900
-@@ -34,7 +34,8 @@
+--- lcov-1.12/bin/install.sh
++++ lcov-1.12/bin/install.sh
+@@ -34,7 +34,8 @@ do_install()
    local TARGET=$2
    local PARAMS=$3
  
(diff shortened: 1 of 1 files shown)


bfontaine:
Could you please add a short comment above `patch :DATA` explaining why the patch is necessary as well as a link to the upstream issue/submission? Thanks!

nsnutella:
Hi @bfontaine,

Of course.
So in the `Makefile` i add a `BREW_DIR` variable to be used in the others `DIR` variables.
In the `install.sh` file since the `install` binary on OS X is the BSD version and not the GNU one, this doesn't supports `install -p -d`, so i split it in a `mkdir` plus a `install`.

About the upstream issue/submission, i'm not understanding what u'r requesting.

I will add the comments in the file and commit the changes

Many thanks :smiley:

bfontaine:
> About the upstream issue/submission, i'm not understanding what u'r requesting.

You should report the issue to the `lcov` developers and add a link to this report as a comment above `patch :DATA`; so we know when reading the formula file why this patch is here.

nsnutella:
I think this is not a issue on `lcov`, because the master version of this formula also patches it in almost the same way.

If you look at lcov.rb in the master, it has the `patch :DATA` to solve this "problem".

I think the lcov doesn't installs out of box in os x, is developed to work on linux. Even on linux, they just provide support for rpm based systems, i think.
http://ltp.sourceforge.net/coverage/lcov.php

https://github.com/Homebrew/homebrew/blob/edcf1d119c4ca9d79d7147a684b7d74767cbb1f6/Library/Formula/lcov.rb
This is the master version of lcov formula, and as you can see it also has a patch to the makefile and install.sh

So i think is is not a bug.

bfontaine:
The second part of the patch could be fixed by upstream using `PREFIX ?= ...` instead of `PREFIX := ...`.

bfontaine:
I was able to install without the second part of the patch by using:
```
system "make", "PREFIX=#{prefix}", "BIN_DIR=#{bin}", "MAN_DIR=#{man}", install
```

nsnutella:
Hum strange, i'm not able to install without patch the `Makefile`
I've tried that, but i get the error:

```
==> Downloading from http://netcologne.dl.sourceforge.net/project/ltp/Coverage%20Analysis/LCOV-1.12/lcov-1.12.tar.gz
==> Patching
patching file bin/install.sh
Error: stack level too deep
Please report this bug:
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/forwardable.rb:174
```

bfontaine:
> Error: stack level too deep
This seems like an issue with the formula; could you post the formula you tried?

nsnutella:
Oh sorry, i totally forgot, i got it working, i was missing " on install.
This is my final version that is working:

```
class Lcov < Formula
  desc "Graphical front-end for GCC's coverage testing tool (gcov)"
  homepage "http://ltp.sourceforge.net/coverage/lcov.php"
  url "https://downloads.sourceforge.net/ltp/lcov-1.12.tar.gz"
  sha256 "b474e49c6c962754063b3be97a757a2ba9e7a455f0aea612863bf67e9b8b8ea7"

  head "https://github.com/linux-test-project/lcov.git"

  bottle do
    cellar :any_skip_relocation
    sha256 "922c6785490b66abc36ca3189c7a8d3f1c28b7aa952936ae327ca1f99b2ac606" => :el_capitan
    sha1 "edad7ab819deb6032734442ea88c343e6779f60c" => :mavericks
    sha1 "61cb990e2928ad0a1b29e131790790994b5a95d6" => :mountain_lion
    sha1 "b417a6801535d439dc26200213e63a4536989d11" => :lion
  end

  patch :DATA

  def install
    inreplace %w[bin/genhtml bin/geninfo bin/lcov],
      "/etc/lcovrc", "#{prefix}/etc/lcovrc"
    system "make", "PREFIX=#{prefix}", "BIN_DIR=#{bin}", "MAN_DIR=#{man}", "install"
  end
end

__END__
--- lcov-1.12/bin/install.sh
+++ lcov-1.12/bin/install.sh
@@ -34,7 +34,8 @@ do_install()
   local TARGET=$2
   local PARAMS=$3
 
-  install -p -D $PARAMS $SOURCE $TARGET
+  mkdir -p `dirname $TARGET`
+  install -p $PARAMS $SOURCE $TARGET
 }

```

bfontaine:
Thank you for your contribution @nsnutella; we appreciate it! :tada:

PR #44806, opened 2015-10-10
Update fltk.rb to support HEAD for OS X El Capitan

jmtatsch:
Added HEAD option for building from svn in order to build with OS X El Capitan

PR #44880, opened 2015-10-13
fltk: add devel build.

dpo:
Closes #44806.

### DIFF
--- a/Library/Formula/fltk.rb
+++ b/Library/Formula/fltk.rb
@@ -1,8 +1,27 @@
 class Fltk < Formula
   desc "Cross-platform C++ GUI toolkit"
   homepage "http://www.fltk.org/"
-  url "https://fossies.org/linux/misc/fltk-1.3.3-source.tar.gz"
-  sha256 "f8398d98d7221d40e77bc7b19e761adaf2f1ef8bb0c30eceb7beb4f2273d0d97"
+
+  stable do
+    url "https://fossies.org/linux/misc/fltk-1.3.3-source.tar.gz"
+    sha256 "f8398d98d7221d40e77bc7b19e761adaf2f1ef8bb0c30eceb7beb4f2273d0d97"
+
+    # Fltk 1.3.4 include support for El Capitan. Remove on update.
+    depends_on MaximumMacOSRequirement => :yosemite
+
+    # Fixes issue with -lpng not found.
+    # Based on: https://trac.macports.org/browser/trunk/dports/aqua/fltk/files/patch-src-Makefile.diff
+    patch :DATA
+  end
+
+  devel do
+    url "http://fltk.org/pub/fltk/snapshots/fltk-1.3.x-r10866.tar.gz"
+    sha256 "4fd4911a1da99c2fa1e6cc0c985b3b8645a7e954802230338fb513ae880ff2cc"
+    version "1.3.3-r10866" # convince brew that this is newer than stable
+
+    depends_on "autoconf" => :build
+    depends_on "autogen" => :build
+  end
 
   bottle do
     sha1 "33c75cce41deadbfe54bdcc22ae91d17d3ecc782" => :mavericks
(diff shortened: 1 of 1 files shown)


dpo:
```
==> brew audit fltk
==> FAILED
Error: 1 problem in 1 formula
fltk:
 * devel version 1.3.x-r10866 is older than stable version 1.3.3
```

bfontaine:
Is `1.3.x-r10866` supposed to be more recent than `1.3.3`?

dpo:
Don't think so. That's just how the FLTK folks versioned their development snapshot.

bfontaine:
As the audit shown we don’t allow devel versions that are older than the stable ones.

dpo:
It's not really older. It's clearly newer because 10.11 is supported. That is the point of this PR.

bfontaine:
Yeah, just checked upstream; the 1.3.3 version was released in November, 2014 while the 1.3.x-r10866 one was released two weeks ago.

You might use `version "1.3.3-r10866"` to ensure Hombrew think it’s a newer release.

dpo:
`1.3.4-r10866`, no?

bfontaine:
`1.3.4-r10866` > `1.3.4` so it won’t work when we’ll update to `1.3.4`.

Library/Formula/fltk.rb
```diff
@@ -1,8 +1,26 @@
...
   desc "Cross-platform C++ GUI toolkit"
   homepage "http://www.fltk.org/"
-  url "https://fossies.org/linux/misc/fltk-1.3.3-source.tar.gz"
-  sha256 "f8398d98d7221d40e77bc7b19e761adaf2f1ef8bb0c30eceb7beb4f2273d0d97"
+  revision 1
+
+  stable do
+    url "https://fossies.org/linux/misc/fltk-1.3.3-source.tar.gz"
+    sha256 "f8398d98d7221d40e77bc7b19e761adaf2f1ef8bb0c30eceb7beb4f2273d0d97"
+    depends_on MaximumMacOSRequirement => :yosemite
```
bfontaine: Please keep the original comment above this line:
`# Fltk 1.3.4 include support for El Capitan. Remove on update.`

Library/Formula/fltk.rb
```diff
@@ -14,20 +32,15 @@ class Fltk < Formula
...
   end
 
-  # Fixes issue with -lpng not found.
-  # Based on: https://trac.macports.org/browser/trunk/dports/aqua/fltk/files/patch-src-Makefile.diff
-  patch :DATA
-
   def install
     ENV.universal_binary if build.universal?
+
+    system "NOCONFIGURE=1 ./autogen.sh" if build.devel?
```
bfontaine: Please export the env variable separatly:
```rb
if build.devel?
  ENV["NOCONFIGURE"] = "1"
  system "./autogen.sh"
end
```

dpo: That's 4 lines instead of 1. Why is it necessary?

bfontaine: `system "foo bar"` runs in a shell while `system "foo", "bar"` doesn’t. We thus prefer the later.

dpo:
Done.

Library/Formula/fltk.rb
```diff
@@ -1,8 +1,28 @@
...
+
+    # Fixes issue with -lpng not found.
+    # Based on: https://trac.macports.org/browser/trunk/dports/aqua/fltk/files/patch-src-Makefile.diff
+    patch :DATA
+  end
+
+  devel do
+    url "http://fltk.org/pub/fltk/snapshots/fltk-1.3.x-r10866.tar.gz"
+    sha256 "4fd4911a1da99c2fa1e6cc0c985b3b8645a7e954802230338fb513ae880ff2cc"
+    version "1.3.3-r10866" # convince brew that this is older than stable
```
bfontaine: This is *newer* than stable :wink:

Library/Formula/fltk.rb
```diff
@@ -1,8 +1,28 @@
 class Fltk < Formula
   desc "Cross-platform C++ GUI toolkit"
   homepage "http://www.fltk.org/"
-  url "https://fossies.org/linux/misc/fltk-1.3.3-source.tar.gz"
-  sha256 "f8398d98d7221d40e77bc7b19e761adaf2f1ef8bb0c30eceb7beb4f2273d0d97"
+  revision 1
```
bfontaine: You don’t need a revision since you didn’t modify the stable version.

dpo:
Sorry ;-). This is the PR that never ends.

Library/Formula/fltk.rb
```diff
@@ -14,20 +33,18 @@ class Fltk < Formula
...
 
-  # Fixes issue with -lpng not found.
-  # Based on: https://trac.macports.org/browser/trunk/dports/aqua/fltk/files/patch-src-Makefile.diff
-  patch :DATA
-
   def install
     ENV.universal_binary if build.universal?
+
+    if build.devel?
+      ENV["NOCONFIGURE"] = 1
```
bfontaine: Use `"1"` instead of `1` here. The build fails otherwise.

dpo:
It was so nice and easy with the one-liner ;-).

dpo:
@BrewTestBot test this please

dpo:
All good now. Thanks!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/004f13435c91c96ab512e226dfa4783c14e727ea. Thank you for pushing through with this @dpo :smiley_cat:.

dpo:
Thank you.

PR #44884, opened 2015-10-13
mailhog 0.1.8 (new formula)

maijs:
[MailHog](https://github.com/mailhog/MailHog) is a Go-based email testing tool for developers.

This formula builds and installs MailHog version 0.1.8.

### DIFF
--- a/Library/Formula/mailhog.rb
+++ b/Library/Formula/mailhog.rb
@@ -155,12 +155,17 @@ class Mailhog < Formula
   end
 
   test do
-    # Test for following default MailHog output:
-    # 1970/01/01 00:00:01 Using in-memory storage
-    # 1970/01/01 00:00:01 [SMTP] Binding to address: 0.0.0.0:1025
-    # MailHog starts process, which needs to be terminated to avoid
-    # indefinite runtime.
-    output = shell_output("#{bin}/MailHog 2>&1 > /dev/null & sleep 3; kill $! | head -n 2")
-    assert_match /Binding to address/, output
+    pid = fork do
+      exec "#{bin}/MailHog"
+    end
+    sleep 2
+
+    begin
+      output = shell_output("curl -s http://localhost:8025")
+      assert_match /<title>MailHog<\/title>/, output
+    ensure
+      Process.kill("SIGINT", pid)
+      Process.wait(pid)
+    end
   end
 end
--- a/Library/Formula/mailhog.rb
+++ b/Library/Formula/mailhog.rb
@@ -130,7 +130,7 @@ class Mailhog < Formula
     Language::Go.stage_deps resources, buildpath/"src"
     system "go", "build", "-o", "MailHog"
     bin.install "MailHog"
-    
+
     puts <<-EOS.undent
     To use MailHog as a sendmail replacement for PHP, add the following line to php.ini file:
       sendmail_path = #{HOMEBREW_PREFIX}/bin/MailHog sendmail test@test
@@ -165,7 +165,7 @@ class Mailhog < Formula
     # Test for following default MailHog ouput:
     # 1970/01/01 00:00:01 Using in-memory storage
     # 1970/01/01 00:00:01 [SMTP] Binding to address: 0.0.0.0:1025
-    output = shell_output("#{bin}/MailHog 2>&1 | head -n 2")
-    assert_match %r{Binding to address:}, output
+    output = shell_output("#{bin}/MailHog 2>&1 > /dev/null | head -n 2")
+    assert_match /Binding to address/, output
   end
 end
--- a/Library/Formula/mailhog.rb
+++ b/Library/Formula/mailhog.rb
@@ -162,7 +162,7 @@ class Mailhog < Formula
 
     begin
       output = shell_output("curl -s http://localhost:8025")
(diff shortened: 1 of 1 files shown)


Library/Formula/mailhog.rb
```diff
@@ -0,0 +1,171 @@
...
+      <string>#{var}/log/mailhog.log</string>
+      <key>StandardOutPath</key>
+      <string>#{var}/log/mailhog.log</string>
+    </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    output = shell_output("#{bin}/MailHog -h 2>&1 | head -n 1")
```
bfontaine: Can’t we write a better test than e.g. `-h`/`-help`/`-version`?

Library/Formula/mailhog.rb
```diff
@@ -0,0 +1,171 @@
...
+    bin.install "MailHog"
+  end
+
+  def caveats; <<-EOS.undent
+    To use MailHog as a sendmail replacement for PHP, add the
+    following line to php.ini file:
+
+      sendmail_path = #{HOMEBREW_PREFIX}/bin/MailHog sendmail test@test
+    EOS
+  end
```
bfontaine: We shouldn’t include this unless this is Homebrew-specific.

Library/Formula/mailhog.rb
maijs: Valid point. Given that `MailHog` outputs debug messages when run directly, test should check against that.

Library/Formula/mailhog.rb
maijs: It seems that there's no `caveats` equivalent for info messages, will use regular `puts` in `def install` to inform users about sendmail replacement option.

Library/Formula/mailhog.rb
```diff
@@ -0,0 +1,171 @@
...
+
+    ENV["GOPATH"] = buildpath
+    Language::Go.stage_deps resources, buildpath/"src"
+    system "go", "build", "-o", "MailHog"
+    bin.install "MailHog"
+
+    puts <<-EOS.undent
+    To use MailHog as a sendmail replacement for PHP, add the following line to php.ini file:
+      sendmail_path = #{HOMEBREW_PREFIX}/bin/MailHog sendmail test@test
+    EOS
```
bfontaine: Please don’t print this. We use caveats for Homebrew- or OSX-specific message; this is a general message that should be in the software documentation instead.

maijs: MailHog project page [has information](https://github.com/mailhog/MailHog#sendmail) about sendmail replacement, hence message removal from `def install`.

Library/Formula/mailhog.rb
```diff
@@ -0,0 +1,168 @@
...
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+    <dict>
+      <key>KeepAlive</key>
+      <true/>
+      <key>Label</key>
+      <string>#{plist_name}</string>
+      <key>ProgramArguments</key>
+      <array>
+        <string>#{HOMEBREW_PREFIX}/bin/MailHog</string>
```
DomT4: This should be `opt_bin` rather than `#{HOMEBREW_PREFIX}/bin/`

maijs: @DomT4 Thanks, noted.

maijs:
Is there anything else that needs to be done for this formula to be merged?

Library/Formula/mailhog.rb
```diff
@@ -0,0 +1,166 @@
...
+      :revision => "8fd9a4b75098b2125fe442e48a3ffbf738254e13"
+  end
+
+  go_resource "labix.org/v2/mgo" do
+    url "https://github.com/go-mgo/mgo.git",
+      :revision => "2e26580ebcca4eca6533535f2dd062cd6bff44ac"
+  end
+
+  def install
+    mkdir_p "#{buildpath}/src/github.com/mailhog/"
```
bfontaine: Use `(buildpath/"src/github.com/mailhog/").mkpath`

Library/Formula/mailhog.rb
```diff
@@ -0,0 +1,166 @@
...
+      <key>StandardErrorPath</key>
+      <string>#{var}/log/mailhog.log</string>
+      <key>StandardOutPath</key>
+      <string>#{var}/log/mailhog.log</string>
+    </dict>
+    </plist>
+    EOS
+  end
+
+  test do
```
DomT4: Can we use the [mpd](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mpd.rb#L129-L141) style test here instead?

maijs: @DomT4 Indeed, utilizing `Process.kill` is a leaner approach.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/4e4cab4556bcf1e43dd9fcca511bf0536351bcb2. Thank you for your contribution to Homebrew @maijs; we appreciate it! :smiley_cat:

PR #44888, opened 2015-10-13
Resurrect owamp from boneyard by including the missing patch inline

candlerb:
See #43361 and Homebrew/homebrew-boneyard#66

### DIFF
--- a/Library/Formula/owamp.rb
+++ b/Library/Formula/owamp.rb
@@ -1,0 +1,49 @@
+class Owamp < Formula
+  desc "Implementation of the One-Way Active Measurement Protocol"
+  homepage "http://www.internet2.edu/performance/owamp/"
+  url "http://software.internet2.edu/sources/owamp/owamp-3.4-10.tar.gz"
+  sha256 "059f0ab99b2b3d4addde91a68e6e3641c85ce3ae43b85fe9435841d950ee2fb3"
+
+  bottle do
+    cellar :any
+    sha1 "e1746058ddd62ec75ec1b62be837e22c3527c37a" => :yosemite
+    sha1 "45429f2d582a54caa0979b9aa533f6dff1c74ec1" => :mavericks
+    sha1 "1020642e1ca36fd067828ed344799023aca8ff09" => :mountain_lion
+  end
+
+  depends_on "i2util"
+
+  # Fix to prevent tests hanging under certain circumstances.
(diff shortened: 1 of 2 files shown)


bfontaine:
Please also edit `tap_migrations.rb` to remove this line: https://github.com/Homebrew/homebrew/blob/51be1a80a8dfadfdc3b91e5dc51f16512dd9fc3c/Library/Homebrew/tap_migrations.rb#L160

bfontaine:
How were you able to retrieve the missing patch? :open_mouth:

DomT4:
This will need to meet the standards in `brew audit --strict --online` now, as it'll be treated like a new formula by the CI.

candlerb:
> How were you able to retrieve the missing patch? :open_mouth:

From `/Library/Caches/Homebrew/owamp--patch-9e5588d57b357f438ae1a785a713f0deaea5a5ba.patch` on a machine where I'd built it previously

candlerb:
> This will need to meet the standards in brew audit --strict --online now, as it'll be treated like a new formula by the CI.

Done, except I don't know how to get rid of the following error:

~~~
==> audit problems
owamp:
 * Formula name conflicts with homebrew/boneyard/owamp
~~~

This is after editing `tap_migrations.rb` as per @bfontaine

DomT4:
You won't be able to get rid of that until it is removed from the boneyard. If that's the only failure, we can ignore it.

candlerb:
It's the only error now. I fixed two others:

~~~
 * `require "formula"` is now unnecessary
~~~

Library/Formula/owamp.rb
```diff
@@ -0,0 +1,49 @@
+class Owamp < Formula
+  desc "Implementation of the One-Way Active Measurement Protocol"
+  homepage "http://www.internet2.edu/performance/owamp/"
```
DomT4: It looks like this more or less redirects to `http://software.internet2.edu/owamp/` now, so perhaps worth using that?

DomT4:
Changed my mind on the notation. Fixed it after merge rather than redoing the whole CI dance.

Merged in https://github.com/Homebrew/homebrew/commit/0d4bfc41df578b66879a921951bb0efca6be43dd. Thank you for your contribution to Homebrew @candlerb! :smiley_cat:

Issue #42222, opened 2015-07-28
make install from tap process more self-documenting


anentropic:
I love homebrew and it's my first stop for installing commandline stuff on Mac

I just tried to do this:

```
brew install xld
Error: No available formula for xld
==> Searching formulae...
==> Searching taps...
Caskroom/cask/xld
```

ok great, `brew install` went ahead and looked in the taps for me... and it seems to have found a match

however at this stage I have no idea what to do to proceed - not great

I kind of think maybe brew should have just gone ahead and installed it from the tap it found.  But I can see it hasn't done that.

(TLDR: I think at this point brew should definitely at least tell me how to install this formula from the tap)

So, I vaguely remember the taps are like some extra repos of formulae. I think I 'tapped' one in the past, maybe that's what I have to do here:

```
$ brew tap Caskroom/cask/xld
Error: Invalid tap name
$ brew tap caskroom/cask
Warning: Already tapped!
```

hmm ok

```
brew install xld
Error: No available formula for xld
==> Searching formulae...
==> Searching taps...
Caskroom/cask/xld
```

oh yeah, I already tried that. er, was it `brew install --tap something` ...? I can't remember

at this point I resort to the google and, though I get satisfaction shortly after, I feel sad in the meantime

dunn:
This is a good point. Some searches will yield multiple formula from different taps, so the output might have to be processed a bit—grouped by tap, maybe with a heading `brew tap Caskroom/cask` above each.

anentropic:
yes, I just really wanted something to say "hey we found this, you can install it by `brew install Caskroom/cask/xld`"

so simple and yet not quite obvious

mikemcquaid:
What about `==> Searching taps... (install with brew install)` or something?

bfontaine:
We could do something like what [homebrew/command-not-found](https://github.com/Homebrew/homebrew-command-not-found) does, e.g.:

```
==> Searching formulae...
==> Searching taps...
Caskroom/cask/xld

Use: brew install <selected formula>
```

mikemcquaid:
@bfontaine works for me :+1:

DomT4:
:+1:. I was going to suggest something like `Maybe you want brew install homebrew/dupes/make` or similar, so @bfontaine's suggestion works nicely.

anentropic:
I will just say, in this example:

```
==> Searching formulae...
==> Searching taps...
Caskroom/cask/xld

Use: brew install <selected formula>
```

[...]

PR #44892, opened 2015-10-13
Surface more feedback when attempting to install a formula that doesn't exist

ericandrewlewis:


ericandrewlewis:
See #42222

Library/Homebrew/cmd/install.rb
```diff
@@ -98,10 +98,27 @@ def install
...
+
         ohai "Searching taps..."
-        puts_columns(search_taps(query))
+        taps_search_results = search_taps(query)
+        if !taps_search_results.any?
+          ofail 'No formulae found in taps.'
+        else
+          puts 'These formulae were found in taps:'
+          puts_columns(taps_search_results)
+          puts "If you meant to install one of them, run `brew install <formula>`"
```
dunn: One point of confusion was how to install formulae from taps.  Maybe we could use the first result as an example? Something like
```ruby
puts "If you meant to install one of them, run (for example) `brew install #{taps_search_results.first}`"

Comment (mikemcquaid) [inline Library/Homebrew/cmd/install.rb]: Yeh, this makes sense to me :+1:

Comment (mikemcquaid) [inline Library/Homebrew/cmd/install.rb]: Or just `brew install <user/tap/formula>`?

File: Library/Homebrew/cmd/install.rb [hunk 2]
```diff
@@ -98,10 +98,27 @@ def install
...
         ofail e.message
         query = query_regexp(e.name)
-        ohai "Searching formulae..."
-        puts_columns(search_formulae(query))
+
+        ohai "Searching for similarly named formulae..."
+        formulae_search_results = search_formulae(query)
+
+        if !formulae_search_results.any?
+          ofail 'No similarly named formulae found.'
```
bfontaine: Please use double quotes everywhere.

Library/Homebrew/cmd/install.rb
```diff
@@ -98,10 +98,27 @@ def install
       else
         ofail e.message
         query = query_regexp(e.name)
-        ohai "Searching formulae..."
-        puts_columns(search_formulae(query))
+
+        ohai "Searching for similarly named formulae..."
+        formulae_search_results = search_formulae(query)
+
+        if !formulae_search_results.any?
```
bfontaine: Use `unless x` instead of `if !x`.

Library/Homebrew/cmd/install.rb
```diff
@@ -98,10 +98,27 @@ def install
...
+          ofail 'No similarly named formulae found.'
+        else
+          puts 'These similarly named formulae were found:'
+          puts_columns(formulae_search_results)
+          puts "If you meant to install one of them, run `brew install <formula>`"
```
bfontaine: We could make a special case for when there’s only one result, e.g. “``Run `brew install <formula>` to install it``”.

mikemcquaid: :+1:

ericandrewlewis:
Added changes based on feedback from @bfontaine and @mikemcquaid in ce671ae

Library/Homebrew/cmd/install.rb
```diff
@@ -98,10 +98,36 @@ def install
...
+        when 0
+          ofail "No similarly named formulae found."
+        when 1
+          puts "This similarly named formula was found:"
+          puts_columns(formulae_search_results)
+          puts "If you meant to install it, run `brew install #{formulae_search_results.first}`"
+        else
+          puts "These similarly named formulae were found:"
+          puts_columns(formulae_search_results)
+          puts "If you meant to install one of them, run (for example) `brew install #{formulae_search_results.first}`"
```
mikemcquaid: Might be worth checking this still looks nice when wrapped at 80 columns (default terminal size) and otherwise add newlines. Could also use `e.g.` instead of `(for example)`

mikemcquaid:
Looks great so far, really like where this is going :clap:

Library/Homebrew/cmd/install.rb
ericandrewlewis: Here's what it looks like at 80x24.

screen shot 2015-10-17 at 12 29 30 pm

Using `e.g.` reads a bit confusing to me:

```bash
If you meant to install one of them, run e.g. `brew install homebrew/x11/atari++`
```

If the intent here is to keep the line brief, I think we can drop the "if you meant" bit:

```bash
To install one of them, run (for example) `brew install homebrew/x11/atari++`
```

bfontaine: :+1: to drop “if you meant”

ericandrewlewis:
Updated in 7b9e1e2 for brevity and to fit well into an 80 character wide terminal based on feedback

mikemcquaid:
Still gonna get wrapping here on long formulae names (particularly from taps) e.g.
```
To install one of them, run (for example) `brew install homebrew/versions/allegr
o4`
```

Maybe instead enforce a newline and 2-character indent like:
```
To install one of them run (for example):
  brew install homebrew/versions/allegro4
```

ericandrewlewis:
@mikemcquaid Dig it, the output looks great and feels more novice-user friendly without the backtick. Updated in 15e9624

I see Travis builds are failing on this but that looks to be an issue with Travis, GitHub, or The Internet™:

```
Cloning into 'Homebrew/homebrew'...
fatal: unable to access 'https://github.com/Homebrew/homebrew.git/': Could not resolve host: github.com
The command "eval git clone --depth=50 https://github.com/Homebrew/homebrew.git Homebrew/homebrew" failed. Retrying, 2 of 3.
```

mikemcquaid:
Merged! Amazing work here @ericandrewlewis; please continue to point out usability issues like this in Homebrew and keep submitting great PRs :clap:

ericandrewlewis:
:confetti_ball: thanks, will do!

Issue #43472, opened 2015-09-01
cmd/update: changed behaviour


vladshablinsky:
https://github.com/Homebrew/homebrew/commit/3b98a75802024d40dc7b3b125d33fe6644e12e7a#diff-278960b31d25036b2a5e7e041556f671

Noticed these changes only today, I think the previous behaviour was pretty comfortable for working with formulae not on the master branch. I think, when you're not on the master branch, it's very likely that you make some development stuff and it is comfortable for you to get the changes you have on your not master branch.

It might be useful to have a flag or something to include that changes from not-master branch to report, because now we don't have anything from not-master branch. Am I the only one who found that behaviour comfortable? Before this commit it was very easy to test something related with adding/deleting/renaming formulae: you could just make a new branch, do whatever you want with a formula and then update from that branch.

I think if you don't need to have stuff from not master branch in the report, you can just update from master branch.

UniqMartin:
Some information can be found in #42074, but since I'm more or less responsible for that change, let me provide a bit more context.

From my point of view `brew update` is basically doing only one thing, no matter the current branch: It updates the master branch. Therefore, it should also generate a report based on what it has done, i.e. the changes to the master branch only.

I think the use case you have, shouldn't be handled by `brew update`. If I understand you correctly, what you'd like to have is something like `git diff master..current-branch` but with the reporting of `brew update`. I believe something like `brew report master..current-branch` would be cool to have and it should only do the comparison, but shouldn't fetch changes and shouldn't touch the master branch at all. What do you think?

> I think if you don't need to have stuff from not master branch in the report, you can just update from master branch.

Unfortunately, this would not really work for me. I tend to work in a branch called `dev` that if basically `master` with a number of branches merged into it (work in progress, customizations that will never be accepted upstream, …). Because at least one of those branches tweaks the update process, I have to be in `dev` before running `brew update` for my customizations to have an effect. Does that make sense?

vladshablinsky:
Hm, everything you said really has sense. I agree, that update updates only master so should include diff from what it really updates. So the current behaviour is OK, but I think something can be added.

Sometimes we need to do something with formulae and we want update to operate on that changed formulae. Since we can't change upstream Homebrew repository add/delete/rename formula there for fun/testing we need to have another Homebrew repository locally (let's call it homebrew-2) and have remote origin set to this local Homebrew copy. So the process of simulating formula rename looks like:

1. Go to `homebrew-2` repo.
[...]

PR #44893, opened 2015-10-13
update: allow updating from local branch

vladshablinsky:
Allows updating from local branch. For example you can change some formulae on your local `branch_name` and test things like formula-renames, report format and other stuff that is performed on update when some formulae get changed.

The option for this `update from branch` behaviour is `--from-branch`, so if you want to check whether migration occurs on brew update for renamed formula on your branch, you can use `brew update --from-branch`.

Solves https://github.com/Homebrew/homebrew/issues/43472.

### DIFF
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -216,6 +216,20 @@ class Updater
       @initial_branch = ""
     end
 
+    # Used for testing purposes, e.g., for testing formula migration after
+    # renaming it in the currently checked-out branch. To test run
+    # "brew update --simulate-from-current-branch"
+    if ARGV.include?("--simulate-from-current-branch")
+      @initial_revision = `git rev-parse -q --verify #{@upstream_branch}`.chomp
+      @current_revision = read_current_revision
+      begin
+        safe_system "git", "merge-base", "--is-ancestor", @initial_revision, @current_revision
+      rescue ErrorDuringExecution
+        odie "Your HEAD is not a descendant of '#{@upstream_branch}'."
+      end
+      return
+    end
+
     if @initial_branch != @upstream_branch && !@initial_branch.empty?
       safe_system "git", "checkout", @upstream_branch, *quiet
     end


Library/Homebrew/cmd/update.rb
```diff
@@ -216,6 +216,20 @@ def pull!(options = {})
       @initial_branch = ""
     end
 
+    # used for testing purpose, for example for testing formula migrations when
+    # some formula gets renamed on your branch and you test migrations by
+    # running 'brew update --from-branch"
+    if ARGV.include?("--from-branch")
+      @initial_revision = `git rev-parse -q --verify master`.chomp
```
UniqMartin: Maybe replace the hard-coded `master` with `#{@upstream_branch}` so it won't fail for the rare cases without a master branch (see #43865)? Same for the `odie` message below.

Library/Homebrew/cmd/update.rb
```diff
@@ -216,6 +216,20 @@ def pull!(options = {})
       @initial_branch = ""
     end
 
+    # used for testing purpose, for example for testing formula migrations when
+    # some formula gets renamed on your branch and you test migrations by
+    # running 'brew update --from-branch"
```
UniqMartin: I'd rephrase the comment as follows to make it a bit clearer (I'm not a native speaker): “Used for testing purposes, e.g., for testing formula migrations in the currently checked-out branch. To test, run `brew update --from-branch`.”

UniqMartin:
I'm a bit conflicted about the option name `--from-branch`, because in my mind it's an update *from* master *to* the local branch. What do you think about `--to-current-branch` (my favorite), `--to-local-branch`, or `--to-HEAD` instead?

vladshablinsky:
>I'm a bit conflicted about the option name --from-branch, because in my mind it's an update from master to the local branch. What do you think about --to-current-branch (my favorite), --to-local-branch, or --to-HEAD instead?

I think I disagree here. When we get updates from the upstream we don't say "to upstream". I think something similar we can see here i.e. we get the diff from the current head and use it in report.

In the independent case when we update just from master to origin/master "to" have more sense I think. But in "--from-branch" I mean get the chages `from` it, but not update from some commit `to` it. @UniqMartin what do you think. Doesn't it have sense?

UniqMartin:
> What do you think? Doesn't it make sense?

Thanks! Now I understand why you named it like that and it makes sense. It feels right to say “get the updates *from* a certain branch” or “get the updates *from* upstream”. I'm happy with any of the suggested option names, including `--from-branch`.

mikemcquaid:
I wonder about the argument naming too as it wasn't obvious to me what it was doing. Maybe `--simulate-from-current-branch` or something as it's not actually doing the update?

vladshablinsky:
@mikemcquaid are you ok with `--simulate-from-current-branch`? It describes the purpose more clearly. It's pretty long though.

mikemcquaid:
@vladshablinsky Works for me :+1:

PR #44895, opened 2015-10-13
Clean whitespace

aarzee:
Remove leading newlines; replace lines containing only whitespace with empty lines; replace multiple trailing newlines with a single newline.

This PR was created semiautomatically.


mikemcquaid:
Have you checked all these formulae still install with whitespace removed from their patches?

DomT4:
Is this not extremely likely to break a lot of the patches?

aarzee:
No, but I can do it now. Give me a few minutes and I’ll get back with an answer.

aarzee:
This is going to take a while, but so far nothing's breaking. It would be quite helpful if there were specific formulae which I should be checking; should I only check those with patches?

aarzee:
I realized that doing `brew install [...] --build-from-source` was also going to build each package's dependencies from source, so I'm currently doing `brew install [...] --only-dependencies`. It looks like I'm about halfway through the packages (this probably has little to no bearing on how much longer the dependency installation will take), and once that's done I'm going to do the first command again. FWIW, I did start doing the first command before I had this realization, and it got through the first 6 packages without any failed installations.

aarzee:
The fontconfig formula broke. https://ghostbin.com/paste/576fg
This is really odd as every single formula before that one installed fine.

dunn:
What's the rationale for this PR?

mikemcquaid:
I'm going to pass on this PR. We're not really interested in changes like this and it has merge conflicts now. Sorry!

PR #44898, opened 2015-10-13
z80dasm: use https homepage

vszakats:


DomT4:
Maintainer note: Let's leave this PR to completion so we can pull new bottles here (Doesn't currently have any).

dunn:
```
z80dasm:

vszakats:
@dunn: In this series of commits the single goal was to sync up homepages (and other related urls) to the newly updated download urls to https following @bfontaine's commits. It means I haven't run audit checks this time, nor did I make an attempt to fix all the many potential, but unrelated formula issues. I'm not sure if this is accepted from a non-maintainer, but that's what I had the capacity for.

DomT4:
It'd be helpful to us if you audited everything prior to submission and fixed nits that will actively fail the CI. There aren't many in terms of just `brew audit` at this point - The `assert include` situation will be the most popular and also one of the easiest to resolve generally.

Fixed it after merge here. Merged in https://github.com/Homebrew/homebrew/commit/f5a9f7b7ef65ff1d5e2f9627d7e85749ff91f6f9. Thank you @vszakats :smiley:

PR #44901, opened 2015-10-13
mecab-unidic: follow redirect + use https homepage

vszakats:

PR #44902, opened 2015-10-13
libpuzzle: follow homepage redirect + use https

vszakats:


DomT4:
Maintainer note: Let's leave this PR to completion so we can pull new bottles here (Doesn't currently have an El Cap one).

vszakats:
@DomT4: Thanks, that makes sense. But, in case I have some time, I'd prefer to close this along with https://github.com/Homebrew/homebrew/pull/44901 and https://github.com/Homebrew/homebrew/pull/44898 and take this time to fix my fork repo, then resubmit along with 50 more, similar modifications in a single PR. Does this sound okay?

DomT4:
> then resubmit them along with 50 more, similar modifications in a single PR. Does this sound okay?

Please don't, for now. Our CI is having a laughably busy day and big PRs cause a lot of problems at the moment. We'd take them in a significantly smaller number per PR, ideally.

vszakats:
Okay, np! Would 5 per PR seem reasonable?

DomT4:
Yup!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/8c27b4d250e4be9029d78bc0bb2af487ef282c28. Thanks @vszakats!

PR #44909, opened 2015-10-13
yank 0.4 (new formula)

dongcarl:
First time making a formula... Please tell me if there's anything I can improve on! :smile:

### DIFF
--- a/Library/Formula/yank.rb
+++ b/Library/Formula/yank.rb
@@ -9,8 +9,6 @@ class Yank < Formula
   end
 
   test do
-    system "#{bin}/yank", "-v"
-    system "man", "yank"
     (testpath/"test").write <<-EOS.undent
       #!/usr/bin/expect -f
       spawn sh
--- a/Library/Formula/yank.rb
+++ b/Library/Formula/yank.rb
@@ -11,5 +11,23 @@ class Yank < Formula
   test do
     system "#{bin}/yank", "-v"
     system "man", "yank"
+    (testpath/"test").write <<-EOS.undent
+      #!/usr/bin/expect -f
+      spawn sh
+      send "echo key=value | yank -d = | cat"
+      send "\r"
+      send "\016"
+      send "\r"
+      expect -exact "value"
+      send "echo key=value | yank -d = | cat"
+      send "\r"
+      send "\016"
+      send "\020"
+      send "\r"
+      expect -exact "key"
+      send "exit\r"
+    EOS
+    (testpath/"test").chmod 0755
+    system "./test"
   end
 end
--- a/Library/Formula/yank.rb
+++ b/Library/Formula/yank.rb
@@ -1,8 +1,8 @@
 class Yank < Formula
   desc "Yank terminal output to clipboard"
   homepage "https://github.com/mptre/yank"
-  url "https://github.com/mptre/yank/archive/v0.4.0.tar.gz"
-  sha256 "1adf1b07f7cb9401daeed7a05bad492db8ed77ead4728d9b45f541d56bc2e8c5"
+  url "https://github.com/mptre/yank/archive/v0.4.1.tar.gz"
+  sha256 "cbd6b3c9d580fa619f3cba2c53b18d26674d587333976fe75b402f76ea01c4e5"
 
   def install
     system "make", "install", "PREFIX=#{prefix}", "YANKCMD=pbcopy"
--- a/Library/Formula/yank.rb
+++ b/Library/Formula/yank.rb
@@ -2,7 +2,6 @@ class Yank < Formula
   desc "Yank terminal output to clipboard"
   homepage "https://github.com/mptre/yank"
   url "https://github.com/mptre/yank/archive/v0.4.0.tar.gz"
-  # version "0.4.0"
   sha256 "1adf1b07f7cb9401daeed7a05bad492db8ed77ead4728d9b45f541d56bc2e8c5"
 
   def install
--- a/Library/Formula/yank.rb
+++ b/Library/Formula/yank.rb
@@ -5,9 +5,7 @@ class Yank < Formula
   sha256 "1adf1b07f7cb9401daeed7a05bad492db8ed77ead4728d9b45f541d56bc2e8c5"
 
   def install
-    ENV["PREFIX"] = prefix
-    ENV["YANKCMD"] = "pbcopy"
-    system "make", "install"
+    system "make", "install", "PREFIX=#{prefix}", "YANKCMD=pbcopy"
   end
 
   test do
--- a/Library/Formula/yank.rb
+++ b/Library/Formula/yank.rb
@@ -1,0 +1,18 @@
+class Yank < Formula
+  desc "Yank terminal output to clipboard"
+  homepage "https://github.com/mptre/yank"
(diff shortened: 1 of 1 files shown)


Library/Formula/yank.rb
```diff
@@ -0,0 +1,18 @@
+class Yank < Formula
+  desc "Yank terminal output to clipboard"
+  homepage "https://github.com/mptre/yank"
+  url "https://github.com/mptre/yank/archive/v0.4.0.tar.gz"
+  # version "0.4.0"
```
bfontaine: You can remove this comment.

Library/Formula/yank.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def install
+    ENV["PREFIX"] = prefix
+    ENV["YANKCMD"] = "pbcopy"
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/yank", "-v"
+    system "man", "yank"
```
bfontaine: We need a test for the actual functionality of the software; `man yank` doesn’t :wink: 
Could you please write a test that does something more substantial than e.g. `-v`/`-h`? Thanks!

Library/Formula/yank.rb
```diff
@@ -0,0 +1,18 @@
...
+  desc "Yank terminal output to clipboard"
+  homepage "https://github.com/mptre/yank"
+  url "https://github.com/mptre/yank/archive/v0.4.0.tar.gz"
+  # version "0.4.0"
+  sha256 "1adf1b07f7cb9401daeed7a05bad492db8ed77ead4728d9b45f541d56bc2e8c5"
+
+  def install
+    ENV["PREFIX"] = prefix
+    ENV["YANKCMD"] = "pbcopy"
+    system "make", "install"
```
bfontaine: You can give both `YANKCMD` and `PREFIX` on the command-line: `system "make", "install", "PREFIX=#{prefix}", "YANKCMD=pbcopy"`.

dongcarl: Of course! Why is this preferred style-wise or functionality-wise over what I've done? I started out with `system "env", "PREFIX=#{prefix}", "YANKCMD=pbcopy", "make", "install"` which `brew audit` told me I should use `ENV` instead of `env`

Library/Formula/yank.rb
dongcarl: @bfontaine Since the program is interactive, the only way of testing functionality would be if I could simulate key presses in the test. How would I go about doing that? Would simulating key presses be safe?

Library/Formula/yank.rb
dongcarl: Okay!

Library/Formula/yank.rb
bfontaine: It’s shorter and slightly more efficient to give them directly to `make`.

dongcarl: Done.

Library/Formula/yank.rb
bfontaine: We were able to test similar formulae using `expect`, like [`scrypt.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/scrypt.rb#L21-L38). See this to understand how it works and how to use `autoexpect` to generate these scripts: http://www.admin-magazine.com/Articles/Automating-with-Expect-Scripts

If you don’t find any way to test this we’ll accept the `yank -v` test.

exalted:
I want I want I want `yank` :+1: :wink:

Library/Formula/yank.rb
dongcarl: @bfontaine for some reason I'm unable to install `homebrew/dupes/expect`, see issue here: https://github.com/Homebrew/homebrew-dupes/issues/518

mikemcquaid: @dongcarl It's provided by OS X in `/usr/bin/expect`

dongcarl: @mikemcquaid according to @bfontaine the OS X version doesn't have autoexpect, I've just sent in a patch fixing `homebrew/dupes/expect`

mikemcquaid: We can't rely on `dupes` here unfortunately.

mikemcquaid: Well, we can't depend on it but if you can use that to generate something we can use inline: :+1:

dongcarl: @mikemcquaid @bfontaine a really poor attempt at making a test to test functionality by going over all environmental keys and seeing if <Control-N> then <Enter> will output the correct value. (<Control-N> is \016 in octal apparently) It doesn't work as of now because somehow the `#!/usr/bin/expect -f` line is simply ignored by Ruby. https://gist.github.com/dongcarl/136f19f7c4ecc348d87c

bfontaine: @mikemcquaid `autoexpect` is only used to generate a script we then put in the test; it’s not directly used in the test.

mikemcquaid: :+1:

dongcarl:
@bfontaine tests seem to be okay. Two questions:

1. Why aren't assertions required in the test block? I don't see them in any other tests on homebrew formulas either
2. If the expect script fails, it currently just hangs until someone Ctrl-C's it, how can we fix this?

dongcarl:
@bfontaine It's been a good lesson in writing `expect` scripts. Everything should now work and my previous two questions have been addressed. Please merge after verifying.

bfontaine:
@dongcarl Great! Just remove the first two lines of the test and we’ll be good to merge :smile: 
```rb
  system "#{bin}/yank", "-v"
  system "man", "yank"
```

> Why aren't assertions required in the test block? I don't see them in any other tests on homebrew formulas either

What do you mean? Some formulae have assertions, e.g. [`git.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/git.rb#L153).

dongcarl:
@bfontaine Done. What I meant was I'm confused as to why we issue system calls but don't check the exit code of system calls, just curious I guess.

bfontaine:
@dongcarl The `system` we’re using here is a wrapper around Ruby’s `system` which automatically check the exit code :wink:

dongcarl:
@bfontaine absolutely amazing... Btw why did we want to remove the test for `man yank`? Wouldn't we want to make sure the man page was installed and configured correctly?

stephencelis:
0.4.1 is out. Can we update the formula before merging? :smile:

dongcarl:
@stephencelis Done.

bfontaine:
@dongcarl We don’t want to test everything; manpages are usually installed correctly.

bfontaine:
Thank you @dongcarl for your (first) contribution to Homebrew; we appreciate it! :tada:

dongcarl:
@bfontaine looks like 0.5.0 is out... What is the conventional way of making an upgrade? Should I delete `yank.rb` and do another `brew create`? If I `brew edit yank` should I just remove the bottle section and update the rest accordingly?

bfontaine:
@dongcarl `brew edit yank`; keep the bottle section and update the rest accordingly. Our CI bot will update the bottle section when we’ll pull the PR.

PR #44914, opened 2015-10-13
thrift 0.9.3

tonyghita:


tonyghita:
Seems like the Travis CI build fails because we've uninstalled `thrift`, which is a requirement of `osquery` (which depends on thrift). Any thoughts on how to proceed? cc @DomT4 @adzenith @vszakats

tonyghita:
Ahh I think I understand... the tests checks if the new version breaks any dependencies... and this change happens to break `osquery`

mikemcquaid:
@tonyghita Yup! `osquery` (and any other `brew uses thrift`) probably needs a `revision 1` added or the `revision` increased if it already exists.

tonyghita:
@mikemcquaid I've bumped the revision on `osquery`, and now I'm seeing a different errors around bottles

mikemcquaid:
@tonyghita You can ignore the Travis failure, that was my screwup yesterday. The Jenkins build is :green_heart: so pulling. Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

tonyghita:
Hey, thanks for the kind words! Glad I could help

PR #44915, opened 2015-10-13
update homepage urls to https (1/10)

vszakats:
* also update related misc urls in comments and resources


DomT4:
Merged in a bunch of commits. Thanks again! :smile:

vszakats:
Thank you :)

PR #44921, opened 2015-10-13
update homepage urls to https (5/10)

vszakats:


mikemcquaid:
`mp3info` download is failing.

mikemcquaid:
Thanks @vszakats!

PR #44923, opened 2015-10-13
update homepage urls to https (3/10)

vszakats:
* also update related misc urls in comments and resources


mikemcquaid:
Thanks @vszakats!

PR #44925, opened 2015-10-13
brew doctor: per formula doctor blocks

josh:
**Feature idea**

The gist is formulas themselves can provide `doctor do` blocks (similar to `test do`) that are included in `brew doctor`.

Since this is more of a debugging feature, I'm wondering how useful this would be to the homebrew maintainers? Are there enough use cases for certain problematic formulas that'd it be useful to have some custom doctor logic?

Could this also cleanup existing brew doctor checks? Could some existing formula specific checks be organized into the relevant formula itself?

--
To: @mikemcquaid

### DIFF
--- a/Library/Homebrew/cmd/doctor.rb
+++ b/Library/Homebrew/cmd/doctor.rb
@@ -1257,6 +1257,15 @@ class Checks
     s
   end
 
+  # define custom checks provided by installed formulas
+  Formula.installed.each do |f|
+    if f.respond_to?(:doctor)
+      define_method "check_#{f.name}" do
+        f.doctor
+      end
+    end
+  end
+
   def all
     methods.map(&:to_s).grep(/^check_/)
   end
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1860,6 +1860,10 @@ class Formula
       define_method(:test, &block)
     end
 
+    def doctor(&block)
+      define_method(:doctor, &block)
+    end
+
     # @private
     def link_overwrite(*paths)
       paths.flatten!


mikemcquaid:
Interesting idea but I'd love to see what other @Homebrew/maintainers think.

xu-cheng:
What exactly do you want to put inside formula's `doctor do` block? Do you have a specific example? Why isn't it covered by `test do`?

From what I'm understanding, you seems to ask for a `brew test-all-installed-formula` command.

bfontaine:
Interesting idea; but I have the same concerns than Xu: what need does it fulfill?

DomT4:
Yeah. Had this thought last night; "Where & how would we use this".

One thing that comes to mind is selectively ignoring the OpenSSL warning, but I don't want to because I pretty strongly believe people should be nagging upstream's to provide a way to build against a more secure OpenSSL or even statically bundling their own.

Curious to know what problems you're resolving here in your use case to help me understand this a little more.

mikemcquaid:
I know @josh's use-case but I'll let him explain it.

josh:
> What exactly do you want to put inside formula's doctor do block? Do you have a specific example? Why isn't it covered by test do?
> 
> From what I'm understanding, you seems to ask for a brew test-all-installed-formula command.

The `test do` block seems like its more for verifying the keg itself was built correctly. Its something that has to work on the homebrew CI system and usually asserts very isolated things.

Many of the existing doctor checks look for things outside the kegs themselves or incompatibilities between kegs and something in the user's environment. None of these will be likely issues in the isolated CI environment.

I think many of the formulas that provide some sort of initial `etc/`, `var/` or launchagent configuration or have some kind of installation caveat section might be able to provide some useful automated doctor check.

For a specific example, the [tuntap formula](https://github.com/Homebrew/homebrew/blob/a6611f9f57786533dae793f9a0ea08800470ed92/Library/Formula/tuntap.rb#L36-L61) could verify all the extensions can been copied in place and they have the correct permissions.

mikemcquaid:
@josh Maybe worth explaining what you did to me specifically with the `launchdns` formula?

bfontaine:
Interesting; so the use-case is formulae that need specific steps done by the user after the installation.

josh:
> Maybe worth explaining what you did to me specifically with the launchdns formula?

Yeah, launchdns has a similar post install caveat section.

https://github.com/Homebrew/homebrew/blob/c14b644356113963d5111d709db2d4684217eed9/Library/Formula/launchdns.rb#L29-L33

Theres some cases where the linking gets screwed up or the homebrew etc config directory wasn't initialized correctly.

mikemcquaid:
To take a step back this maybe points towards having a `brew userinstall launchdns` or something that you can also run with `sudo` for some packages to do this type of caveat installation manually but with a single command.

josh:
> To take a step back this maybe points towards having a brew user install launchdns

Something like that would be nice to have as well. You'd have to figure out how `useruninstall` works, how upgrading works, can you just upgrade the keg or do you need to reinit config as well.

Issue #33448, opened 2014-10-21
mysql failed to build on 10.10


keewek:
https://gist.github.com/keewek/5e05a1968e3d205c0ab7#file-02-make-L346-L414

htonll detection failure, same as in https://github.com/Homebrew/homebrew/issues/33439

RandomDSdevel:
@jacknagel:  I've read [that bug](http://bugs.mysql.com/bug.php?id=73405)'s comments, and I _don't_ think that its resolution is stalled by 'red tape.'  In the second-to-last comment, Sveta Smirnova specifically stated the following:  

```
Thank you for the report.

I can verify this as feature request only unless you provide example of any OS which is supported by MySQL officially and has htonll defined as a macro.
```

So, the problem is that Ryan Schmidt, the individual who reported the bug and submitted some patches for it, did not put enough details about this bug into either his initial report or his later comments.  Maybe somebody could drop an info dump about this on the bug's upstream page?

RandomDSdevel:
Oops, that was thread necromancy…the bug still isn't fixed, though, and _I'd_ like to have Homebrew build MySQL with `memcached` support before it installs it, _too_.  (Just sayin'.)  Oh, well; guess I'll have to keep my eye on this issue and reinstall MySQL when it gets fixed.

jacknagel:
Personally, I don't have the time or interest to pursue the issue, but if you're interested, by all means, try to help move the upstream issue along.

RandomDSdevel:
OK, I might as well look into doing so, but it's not that high on _my_ priority list, either…

Cam:
Is there another way to build MySQL using Homebrew? This thing has me stopped in my tracks, and I can't seem to find any valuable info anywhere on workarounds or solutions :(

mikemcquaid:
@Cam Build without `memcached` support.

Cam:
Thanks @mikemcquaid. I figured ou that memchache wasn't needed eventually. Cheers!

RandomDSdevel:
Ah, good!  I'll try a `brew reinstall mysql --with-memcached` once #44928 gets resolved and pulled into `master`, then.

Issue #42935, opened 2015-08-14
no memcached support in most recent MySQL brew port


GETandSELECT:
see http://stackoverflow.com/questions/31407630/os-x-and-brew-install-innodb-mysql-memcached-plugin-for-nosql

innodb_memcached_config.sql not available https://dev.mysql.com/doc/refman/5.6/en/innodb-memcached-internals.html

mikemcquaid:
Can you provide more details about this and what you're expecting?

jeffjohnson9046:
I ran into this same issue yesterday:

### Steps to Reproduce
1. Uninstall existing MySQL installation: ```brew uninstall mysql```
2. Make sure ```homebrew``` is up to date: ```brew update```
3. Verify that the ```libevent``` dependency is installed: ```brew install libevent``` returns ```Warning: libevent-2.0.22``` already installed
4. Attempt to install MySQL, including the ```memcached``` plugin:  ```brew install --with-memcached mysql```

**NOTE:** The ```brew install --enable-memcached mysql``` also produces the same result as described in **Actual Result** below, but the output on the console indicates that ```--enable-memcached``` has been deprecated and replaced with ```--with-memcached```.

### Expected/Desired Result
MySQL is installed with the components necessary to install and configure the MySQL memcached plugin.

### Actual Result
Observe the following result after executing ```brew install --with-memcached mysql```:
```
==> Downloading https://cdn.mysql.com/Downloads/MySQL-5.6/mysql-5.6.26.tar.gz
Already downloaded: /Library/Caches/Homebrew/mysql-5.6.26.tar.gz
==> cmake . -DCMAKE_INSTALL_PREFIX=/usr/local/Cellar/mysql/5.6.26 -DCMAKE_FIND_FRAMEWORK=LAST -DCMAKE_VERBOSE_MAKEFILE=ON -DMYSQL_DATADIR=/usr/local/var/mysql -DINSTALL_INCLUDEDIR=i
==> make
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libstrings.a(longlong2str.c.o) has no symbols
/usr/bin/ranlib libstrings.a
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libstrings.a(longlong2str.c.o) has no symbols
[ 10%] Built target strings
make: *** [all] Error 2

[...]

Issue #33439, opened 2014-10-21
percona-server failed to build on 10.10


keewek:
https://gist.github.com/keewek/1fa14b2b1f4bd1e4d8f4

robrecord:
Sorry guys, can I bump this? How do I help get this sorted? Haven't been able to build persona server for some time now.

jacknagel:
It's an upstream bug (shared by all mysql derivatives). Either bump an upstream bug report or build without memcached support. Sorry :/

robrecord:
Okay, thank you - I will do both! And will maybe check out virtualisation for my dev environment.

PR #44887, opened 2015-10-13
Upstream patch applied for memcached plugin support

ehoner:
This applies the patch mentioned in the mysql bug tracker - http://bugs.mysql.com/bug.php?id=73405.

Should enable build support for memcached plugin. Successful build on El Capitan with:
brew install mysql --with-archive-storage-engine --with-local-infile --with-blackhole-storage-engine --with-memcached --with-tests

Fixes:
https://github.com/Homebrew/homebrew/issues/33448. 

Potentially related to:
https://github.com/Homebrew/homebrew/issues/42935
https://github.com/Homebrew/homebrew/issues/33439

PR #44928, opened 2015-10-14
mysql: apply upstream patch for memcached plugin support

ehoner:
This applies the patch mentioned in the mysql bug tracker - http://bugs.mysql.com/bug.php?id=73405.

Should enable build support for memcached plugin. Successful build on El Capitan with:
brew install mysql --with-archive-storage-engine --with-local-infile --with-blackhole-storage-engine --with-memcached --with-tests

Fixes:
https://github.com/Homebrew/homebrew/issues/33448

Potentially related to:
https://github.com/Homebrew/homebrew/issues/42935
https://github.com/Homebrew/homebrew/issues/33439

### DIFF
--- a/Library/Formula/mytop.rb
+++ b/Library/Formula/mytop.rb
@@ -15,8 +15,8 @@ class Mytop < Formula
   depends_on "openssl"
 
   resource "DBD::mysql" do
-    url "https://search.cpan.org/CPAN/authors/id/C/CA/CAPTTOFU/DBD-mysql-4.027.tar.gz"
-    mirror "https://search.mcpan.org/CPAN/authors/id/C/CA/CAPTTOFU/DBD-mysql-4.027.tar.gz"
+    url "http://search.cpan.org/CPAN/authors/id/C/CA/CAPTTOFU/DBD-mysql-4.027.tar.gz"
+    mirror "https://cpan.metacpan.org/authors/id/C/CA/CAPTTOFU/DBD-mysql-4.027.tar.gz"
     sha256 "2e00f1eb5822aa097b310203ff611dd635f9d000db9df7ce1e8b0493c082be41"
   end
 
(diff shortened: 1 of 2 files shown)


ehoner:
This is the updated version of pull request https://github.com/Homebrew/homebrew/pull/44887

ehoner:
The error in the build is for version 10.10 and 10.9 on package mytop - not part of the change here. The 10.11 version passes, but does not include the mytop package. What do I need to fix ?

Library/Formula/mytop.rb
```diff
@@ -15,8 +15,8 @@ class Mytop < Formula
   depends_on "openssl"
 
   resource "DBD::mysql" do
-    url "http://search.cpan.org/CPAN/authors/id/C/CA/CAPTTOFU/DBD-mysql-4.027.tar.gz"
-    mirror "http://search.mcpan.org/CPAN/authors/id/C/CA/CAPTTOFU/DBD-mysql-4.027.tar.gz"
+    url "https://search.cpan.org/CPAN/authors/id/C/CA/CAPTTOFU/DBD-mysql-4.027.tar.gz"
```
mikemcquaid: This URL is invalid.

ehoner:
The reason this fails is because the mytop test - a dependency for the pull request to update mysql - cannot succeed in the absence of a running instance of mysql. I have updated the test to work properly when installed locally, but mysql does not have permission to run in the sandbox.

Is there a way to start mysql successfully in the sandbox so the test for mytop can pass?

apjanke:
Ugh, setting up `mysqld` and `mytop` to run inside the sandbox would be a good amount of work. it's probably something we should do, but not for a little tool like `mytop` which looks like it's on its way to the boneyard anyway.

Let's just change the `mytop` test to this trivial one and be done with it.

```
  test do
    safe_system("man mytop > /dev/null")
  end
```

MikeMcQuaid:
We'll consider accepting this PR on https://github.com/homebrew/brew in future when this patch is merged upstream. Thanks!

PR #44936, opened 2015-10-14
test-bot: make master builds faster.

mikemcquaid:
Just do a few things rather than doing e.g. a full compile.

CC @xu-cheng @DomT4 who I'd talked about this with.
CC https://github.com/Homebrew/homebrew/issues/44320

### DIFF
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -10,7 +10,6 @@
 # --skip-homebrew: Don't check Homebrew's files and tests are all valid.
 # --junit:         Generate a JUnit XML test results file.
 # --email:         Generate an email subject file.
-# --no-bottle:     Run brew install without --build-bottle
 # --keep-old:      Run brew bottle --keep-old to build new bottles for a single platform.
 # --HEAD:          Run brew install with --HEAD
 # --local:         Ask Homebrew to write verbose logs under ./logs/ and set HOME to ./home/
@@ -18,6 +17,7 @@
 # --dry-run:       Just print commands, don't run them.
 # --fail-fast:     Immediately exit on a failing step.
 # --verbose:       Print out all logs in realtime
+# --fast:          Don't install any packages but run e.g. audit anyway.
 #
 # --ci-master:           Shortcut for Homebrew master branch CI options.
 # --ci-pr:               Shortcut for Homebrew pull request CI options.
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/test-bot.rb
```diff
@@ -437,10 +438,35 @@ def formula(formula_name)
...
+      if ARGV.include? "--fast"
+        formula_fetch_options = []
+        formula_fetch_options << "--force" if ARGV.include? "--cleanup"
+        formula_fetch_options << canonical_formula_name
+        test "brew", "fetch", "--retry", "--build-bottle",
+             *formula_fetch_options
+
+        if formula.bottled?
+          test "brew", "fetch", "--retry", *formula_fetch_options
+          install_args = ["--verbose", canonical_formula_name]
```
bfontaine: Any reason to define this as a variable even if it’s used only once?

mikemcquaid: Nope! It was used twice in an earlier iteration. I'll fix when I pull (to avoid another build)

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -437,10 +438,35 @@ def formula(formula_name)
         formula_name
       end
 
-      test "brew", "uses", canonical_formula_name
-
       formula = Formulary.factory(canonical_formula_name)
 
+      if ARGV.include? "--fast"
```
xu-cheng: Can we put all of this in a standalone method like `fast_formula_test` or something and use it in [here](https://github.com/mikemcquaid/homebrew/blob/test-bot-master-fast/Library/Homebrew/cmd/test-bot.rb#L765-L767) like:
```ruby
if ARGV.include? "--fast"
  formulae.each { |f| fast_formula_test(f) }
else
  formulae.each { |f| formula(f) }
end
```

Personally, I found long method with early return on different function is hard to follow.

xu-cheng:
There are two thing you need to handle even in fast test.
* Handle the untapped deps
* Handle the conflict formulae

mikemcquaid:
> There are two thing you need to handle even in fast test.

I think I'll just change the approach then so some steps are skipped on a fast test.

mikemcquaid:
Have changed the approach so `--fast` just excludes some stuff instead.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -883,7 +894,7 @@ def sanitize_ARGV_and_ENV
       ARGV << "--junit" << "--local"
     end
     if ARGV.include? "--ci-master"
-      ARGV << "--no-bottle" << "--email"
+      ARGV << "--no-bottle" << "--email" << "--fast"
```
xu-cheng: Do we still need `--no-bottle` flag? seems to be replaced by `--fast`.

mikemcquaid: Good cal, getting rid of it.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -547,35 +547,41 @@ def formula(formula_name)
...
-        test "brew", "install", *install_args
+        if !ARGV.include?("--fast") || formula_bottled
+          test "brew", "install", "--only-dependencies", *install_args unless dependencies.empty?
+          test "brew", "install", *install_args
+          install_passed = steps.last.passed?
+        else
+          install_passed = false
+        end
       end
-      install_passed = steps.last.passed?
```
bfontaine: Just caught this: moving `install_passed` inside the `run_as_not_developer` block broke the `if install_passed` check; the variable is not defined anymore outside of the block. Fixed in a6611f9f.

mikemcquaid: Whoops, thanks.

PR #44940, opened 2015-10-14
Add auditing as necessary step before submitting new formulae to "Formula-Cookbook.md"

maijs:
Travis CI runs `brew audit [formula]` check with `--strict` option which results in more scrupulous test. Formula creators should be aware of this fact and make sure `brew audit --strict [formula]` returns no errors before they create pull request.


mikemcquaid:
That's only necessary for new formulae in which case `--online` should also be used.

maijs:
I concur. `--online` option added.

mikemcquaid:
Sorry, to clarify: `--strict --online` are only required for new formulae submissions, not all formulae updates.

maijs:
@mikemcquaid, thanks for clarification. Moved the instructions to `Formula-Cookbook.md` which addresses new formulae specifically.

mikemcquaid:

PR #44946, opened 2015-10-14
FormulaValidationError: include full_name

bfontaine:
This would make debugging easier.

You can try with the following:
```
$ brew install trash
$ vim trash.rb # edit the trash.rb formula to remove the version from the URL
$ brew cleanup trash
```

Before:
```
Error: invalid attribute: version (nil)
Please report this bug:
/usr/local/Library/Homebrew/formula.rb:195:in `validate_attributes!'
…snip…
```

After:
```
Error: invalid attribute for formula 'trash': version (nil)
Please report this bug:
/usr/local/Library/Homebrew/formula.rb:195:in `validate_attributes!'
…snip…
```

Don’t forget to checkout/edit the `trash.rb` formula to get the version back after the test.

### DIFF
--- a/Library/Homebrew/exceptions.rb
+++ b/Library/Homebrew/exceptions.rb
@@ -23,11 +23,12 @@ class NoSuchKegError < RuntimeError
 end
 
 class FormulaValidationError < StandardError
-  attr_reader :attr
+  attr_reader :attr, :formula
 
-  def initialize(attr, value)
+  def initialize(formula, attr, value)
     @attr = attr
-    super "invalid attribute: #{attr} (#{value.inspect})"
+    @formula = formula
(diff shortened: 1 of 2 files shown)


mikemcquaid:
:+1:

Library/Homebrew/exceptions.rb
```diff
@@ -25,9 +25,9 @@ def initialize(name)
 class FormulaValidationError < StandardError
   attr_reader :attr
 
-  def initialize(attr, value)
+  def initialize(full_name, attr, value)
```
xu-cheng: It may be worth to just call this variable as `formula` and store it as instance variable.

bfontaine: Done.

PR #44952, opened 2015-10-14
Update google-sparsehash to latest upstream release

rdwampler:
Upstream released v 2.0.3 so we no longer need to apply the previous patches. Also updated the links to the new homepage.


bfontaine:
Thank you for your contribution to Homebrew @rdwampler; we appreciate it! :tada: 

One tiny note for the future: the preferred commit message format for version bumps is “`google-sparsehash 2.0.3`” i.e. “`<formula> <version>`”.

Issue #44150, opened 2015-09-19
Poppler fails to install with Qt5 optional bindings


skalee:
Following fails ([logs in gist](https://gist.github.com/anonymous/03d1b54cf313eef1075e)):

    brew install poppler --with-qt5

However this one surprisingly succeeds ([logs in gist](https://gist.github.com/anonymous/410676aff5ad19926d5f)):

    brew install poppler --with-qt --with-qt5

Thanks for your attention (and hopefully fixes, or maybe caveats entry is enough?). Pretty odd, isn't it?

skalee:
Though I don't see `poppler-qt5.h` anywhere neither `/usr/local/Cellar/poppler/0.35.0/include/poppler/qt5` whereas `poppler-qt4.h` is present in `/usr/local/Cellar/poppler/0.35.0/include/poppler/qt4`…

skalee:
1. Only the most recent Poppler versions: 0.35.0 and 0.36.0 are affected.
2. Installation succeeds for Poppler 0.34.0.
3. As far as I understand the logs, the problem is related to tests run after compilation.

DomT4:
The problem is `qt5`. It has ridiculously broken `pkg-config` files, which `poppler` uses during compile. Poppler is a victim rather than a suspect in this breakage, sadly.

skalee:
My previous comment was invalid. Poppler 0.34.0 does not detect Qt5 and disables the corresponding wrapper, even though `----enable-poppler-qt5` was passed:

```
checking for POPPLER_QT5... yes
checking for POPPLER_QT5_TEST... yes
checking for moc... moc
checking for Qt5 moc... no
checking for moc-qt5... no
checking for Qt5 moc-qt5... checking for qtchooser... no
…
Building poppler with support for:
  qt5 wrapper:        no
```

On the other hand, Poppler 0.35.0 detects Qt correctly, it just fails to brew:
```
checking for POPPLER_QT4... no
checking for POPPLER_QT5... yes
checking for POPPLER_QT5_TEST... yes
checking for moc... moc
checking for Qt5 moc... moc
…
Building poppler with support for:
  qt5 wrapper:        yes
```

skalee:
@DomT4 Thanks a lot.

DomT4:
[...]

PR #44954, opened 2015-10-14
poppler: refuse to build with incompatible options

UniqMartin:
Previously, if both `--with-qt` and `--with-qt5` were passed, the latter was silently ignored. This sometimes caused confusion because the Qt 5 bindings were missing. The build now aborts in such cases and prints a helpful error message.

This PR is motivated by #44150. The reasons for using `odie` instead of something non-fatal are:

- A message that does not interrupt the build will very likely stay unnoticed. The resulting absence of the requested Qt 5 bindings will confuse the user.
- There are already formulae that prefer to fail instead of partially ignoring some arguments, e.g., `pyqt`, `pyqt5`, `sip`.

### DIFF
--- a/Library/Formula/poppler.rb
+++ b/Library/Formula/poppler.rb
@@ -52,7 +52,9 @@ class Poppler < Formula
       --enable-introspection=yes
     ]
 
-    if build.with? "qt"
+    if build.with?("qt") && build.with?("qt5")
+      raise "poppler: --with-qt and --with-qt5 cannot be used at the same time"
+    elsif build.with? "qt"
       args << "--enable-poppler-qt4"
     elsif build.with? "qt5"
       args << "--enable-poppler-qt5"


UniqMartin:
Bump. :wave:

Library/Formula/poppler.rb
```diff
@@ -52,7 +52,9 @@ def install
       --enable-introspection=yes
     ]
 
-    if build.with? "qt"
+    if build.with?("qt") && build.with?("qt5")
+      odie "poppler: --with-qt and --with-qt5 cannot be used at the same time"
```
xu-cheng: Please use `raise` instead of `odie`.

UniqMartin: Thanks, will change that! Just out of curiosity: What is the difference? Because the result is (as far as I can tell) the same in both cases.

xu-cheng: I think the reason is mostly about consistence. If you look for a reason other than consistence, I think there are some small differences. `odie` will quit program immediately, while `raise` is something you can rescue.

UniqMartin: I see. Would it be good to have an audit test that checks for uses of `odie` and suggests to use `raise` instead? Or would that be overkill? I picked `odie` because it was the first thing that came to my mind and I saw it used in multiple formulae for the exact same purpose (to fail on incompatible options) …

xu-cheng: >  Would it be good to have an audit test that checks for uses of odie and suggests to use raise instead? 

Good idea if you can submit a PR.

mikemcquaid: I don't think there's an advantage of `raise` over `odie` if the error output to users is the same. I don't think it's really necessary to be consistent but if we do I'd prefer we did on `odie` which is documented.

xu-cheng: @mikemcquaid about `odie` vs `raise`.

> I think the reason is mostly about consistence. If you look for a reason other than consistence, I think there are some small differences. odie will quit program immediately, while raise is something you can rescue.

mikemcquaid: Yeh, I read that. Whether it's rescuable or not is an implementation detail; we're not rescuing it in this case.

UniqMartin: Most use cases for either `odie` or `raise` in formulae that I came across were telling the user about conflicting options. Maybe it's worth having a specialized method that is only available to formulae?

mikemcquaid: Maybe could just have `Formula#odie` throw an exception if we really wanted. Would rather not add more to the DSL unless really necessary (and eventually we'll solve the "conflicting options" stuff in a more elegant way, I expect)

UniqMartin: I'd be happy with such a solution.

PR #44955, opened 2015-10-14
Upgrade repo launcher command to 1.22

grahamc:
The old android source on github is no longer maintained, so switch to the official git repository.

### DIFF
--- a/Library/Formula/repo.rb
+++ b/Library/Formula/repo.rb
@@ -1,9 +1,10 @@
 class Repo < Formula
   desc "Repository tool for Android development"
   homepage "https://source.android.com/source/developing.html"
-  url "https://raw.githubusercontent.com/android/tools_repo/v1.12.32/repo"
+  url "https://gerrit.googlesource.com/git-repo.git",
+      :tag => "v1.12.32",
+      :revision => "745b4ad660f8050045b521c4e15b7d3ac0b3d70e"
   version "1.22"
-  sha256 "9907c36d63cf5222d73ab270fee5249dc84f6b4580cda03ddd7ac309f11b289e"
 
   def install
     bin.install "repo"


Library/Formula/repo.rb
```diff
@@ -1,9 +1,8 @@
 class Repo < Formula
   desc "Repository tool for Android development"
   homepage "https://source.android.com/source/developing.html"
-  url "https://raw.githubusercontent.com/android/tools_repo/v1.12.21/repo"
-  version "1.21"
-  sha256 "44d1ac8b948e6154a495469dad05870dc6182712c0610da8d255b5471dd5bcb9"
+  url "https://gerrit.googlesource.com/git-repo", :ref => 'v1.12.32', :using => :git
```
DomT4: You can stick `.git` on the end of the `googlesource` URL; I haven't come across a GoogleSource URL that doesn't support that yet. Likewise the syntax should be:

```ruby
url "https://gerrit.googlesource.com/git-repo",
     :tag => "v1.12.32",
     :revision => "CommitSHA1"
```

grahamc: Ah, cheers! Fixed, thank you.

Library/Formula/repo.rb
```diff
@@ -1,9 +1,11 @@
 class Repo < Formula
   desc "Repository tool for Android development"
   homepage "https://source.android.com/source/developing.html"
-  url "https://raw.githubusercontent.com/android/tools_repo/v1.12.21/repo"
-  version "1.21"
-  sha256 "44d1ac8b948e6154a495469dad05870dc6182712c0610da8d255b5471dd5bcb9"
+  url "https://gerrit.googlesource.com/git-repo.git",
+      :ref => 'v1.12.32',
```
DomT4: `tag` rather than `ref` :)

grahamc: Ok. I think I got it this time :| Thank you!

grahamc:
I see the build failed, but I don't understand why.

DomT4:
That's our bad. Will be fixed shortly.

bfontaine:
I just merged #44842. We can still keep this PR to use the new URL but you’ll need to rebase it on top of `master`.

grahamc:
OK, I just pushed an update rebasing against master.

grahamc:
@DomT4 @bfontaine OK I rebased onto master to fix it up, tests have run... except for 'default', which I have no idea hat that is / why it hasn't ever finished :)

DomT4:
`default` is our regular CI that produces our binary packages (bottles) for each formula. Isn't actually necessary here because there's no advantage to a binary package for a formula that literally only installs one binary with no compile stage, but *normally* we'd wait on `default` to build those. And it hasn't started yet because the last day or so has been wildly busy in terms of submissions to Homebrew :smile_cat:.

bfontaine:
Thank you for your contribution @grahamc; we appreciate it! :tada:

grahamc:
Woohoo! Thank you!

PR #44959, opened 2015-10-14
qemu: Add mirror link.

DoomHammer:
It may come in handy in case of main server outage.

### DIFF
--- a/Library/Formula/qemu.rb
+++ b/Library/Formula/qemu.rb
@@ -2,6 +2,7 @@ class Qemu < Formula
   desc "x86 and PowerPC Emulator"
   homepage "http://wiki.qemu.org"
   url "http://wiki.qemu-project.org/download/qemu-2.4.0.1.tar.bz2"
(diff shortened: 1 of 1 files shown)


Library/Formula/qemu.rb
```diff
@@ -2,6 +2,7 @@ class Qemu < Formula
   desc "x86 and PowerPC Emulator"
   homepage "http://wiki.qemu.org"
   url "http://wiki.qemu-project.org/download/qemu-2.4.0.1.tar.bz2"
+  mirror "http://ftp.osuosl.org/pub/blfs/conglomeration/qemu/qemu-2.4.0.tar.bz2"
```
bfontaine: Shouldn’t it be `qemu-2.4.0.1.tar.bz2`?

DoomHammer: You bet it should. Thanks for catching it!

mikemcquaid:

PR #44963, opened 2015-10-14
audit fixes: C 05: cpmtools - cryptopp

apjanke:


mikemcquaid:
@apjanke I know it's a total pain but can you split all these audit PRs into one commit per formula? Also: just for future reference we probably don't really need audit-only fixes (but I'll apply the ones you've already done) Thanks!

DomT4:
> Also: just for future reference we probably don't really need audit-only fixes

Kind of "my bad" on this advice; I said we normally wouldn't take audit only fixes but I was alright with them at the moment as it gave us a chance to get El Cap bottles done and kill off old SHA1 bottles.

DomT4:
Merged wholesale in https://github.com/Homebrew/homebrew/commit/28782ca94a920752967f33a0e2b7fc5e8a11cbdf here but let's try and get one commit per formula done in future. It's useful on `git blame` if things break and `git revert` and so on.

Thanks again @apjanke :smiley:.

apjanke:
Given the workload it's placing on the CI and people, I'm just going to hold off indefinitely on the remaining audit fixes and just handle them when other formula updates/issues come up.

mikemcquaid:
@apjanke Now you have commit access: you can commit them directly :)

DomT4:
If you do a lot at once though you may wish to cancel the resulting `Homebrew` job on the CI unless things are quiet :smile_cat:

mikemcquaid:
@DomT4 @apjanke Should be alright now that I've made it much faster (basically it will no longer try to build things from source).

apjanke:
All right. I'm going to drop another one in and see how the build time goes.

(I'm new enough to be real nervous about committing anything directly yet, though this might be a good benign opportunity to practice the workflow.)

DomT4:
@mikemcquaid :star2: :beer: 

@apjanke Has anyone/any documentation let you know about the magic of `brew pull` yet?

PR #44965, opened 2015-10-14
audit fixes: C 04: cloog - cowsay

apjanke:


apjanke:

PR #44966, opened 2015-10-14
mpd: Allow optional libmad dependency

werkshy:
I have been unable to stream from internet radio stations (any mp3 streams I have tried, such as "http://relay2.hirschmilch.de:5000/psytrance.mp3") due to a crash in ffmpeg:

<pre>
Oct 14 16:46 : ffmpeg: detected input format 'mp3' (MP2/3 (MPEG audio layer 2/3))
mpd(24330,0x112aad000) malloc: *** error for object 0x112aaabe0: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
</pre>

This pull request enables an optional dependency on libmad as a workaround.

### DIFF
--- a/Library/Formula/mpd.rb
+++ b/Library/Formula/mpd.rb
@@ -61,6 +61,7 @@ class Mpd < Formula
   depends_on "opus" => :optional        # Opus support
   depends_on "libvorbis" => :optional
   depends_on "libnfs" => :optional
+  depends_on "mad" => :optional
 
   def install
     # mpd specifies -std=gnu++0x, but clang appears to try to build
@@ -82,7 +83,7 @@ class Mpd < Formula
       --disable-libwrap
     ]
 
-    args << "--disable-mad"
+    args << "--disable-mad" if build.without? "mad"
     args << "--disable-curl" if MacOS.version <= :leopard
 
     args << "--enable-zzip" if build.with? "libzzip"


Library/Formula/mpd.rb
```diff
@@ -30,6 +30,7 @@ class Mpd < Formula
   option "with-libvorbis", "Build with vorbis support (for Ogg encoding)"
   option "with-yajl", "Build with yajl support (for playing from soundcloud)"
   option "with-opus", "Build with opus support (for Opus encoding and decoding)"
+  option "with-mad", "Build with libmad support"
```
bfontaine: You can leave out this line if you don’t provide an explanation beside “build with libmad support”; it’ll be automatically added with the `depends_on "libmad" => :optional`.

werkshy: Thanks, fixed that (and fixed libmad->mad suggestion from 'brew audit')

Library/Formula/mpd.rb
```diff
@@ -61,6 +61,7 @@ class Mpd < Formula
   depends_on "opus" => :optional        # Opus support
   depends_on "libvorbis" => :optional
   depends_on "libnfs" => :optional
+  depends_on "mad" => :optional
```
bfontaine: Shouldn’t that be `:recommended` to keep the current behavior? `:optional` means it’s not included by default while `:recommended` means it’s always included unless you use `--without-mad`?

werkshy: The current behavior is to not depend on mad at all. I am trying to make it optionally available, is this not the right way to do that?

I believe the original intent in disabling the dependency on mad was to simplify the build requirements since ffmpeg handles mp3. I have not been able to diagnose my crashes (the same streams work in mplayer with ffmpeg as the decoder) so this seems like a useful workaround for me and maybe for others.

bfontaine: Yeah I read too quickly, sorry :smile:

bfontaine:
Thank you for your contribution @werkshy! :tada:

PR #44968, opened 2015-10-14
swiftgen 0.5.0

AliSoftware:
Simple version bump with new features

Notes:
* The `rake install` command as changed and now takes the directories where to install `$bin` and `$lib` (Frameworks)
* The fixtures have been reorganized in the source repo, and as a consequence it's now easier to install them in `pkgshare` (hence the big diff in the formula there)

### DIFF
--- a/Library/Formula/swiftgen.rb
+++ b/Library/Formula/swiftgen.rb
@@ -1,8 +1,8 @@
 class Swiftgen < Formula
   desc "Collection of Swift tools to generate Swift code"
   homepage "https://github.com/AliSoftware/SwiftGen"
-  url "https://github.com/AliSoftware/SwiftGen/archive/0.4.4.tar.gz"
-  sha256 "429e50b33d1cd3e34fa81b5becff865ce5663284380e91602783bfa109c96b83"
+  url "https://github.com/AliSoftware/SwiftGen/archive/0.5.0.tar.gz"
+  sha256 "555f190f2ffef940eebd80a926eeb05d3d0de573412028c5bd2184e2b9542929"
   head "https://github.com/AliSoftware/SwiftGen.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


Library/Formula/swiftgen.rb
```diff
@@ -14,33 +14,34 @@ class Swiftgen < Formula
...
-    pkgshare.install "Tests/Assets/expected/FileDefaults.swift.out" => "assets.swift"
-    pkgshare.install "Tests/Colors/expected/FileDefaults.swift.out" => "colors.swift"
-    pkgshare.install "Tests/L10n/expected/FileWithDefaults.swift.out" => "l10n.swift"
-    pkgshare.install "Tests/Storyboard/expected/MessageWithDefaults.swift.out" => "storyboard.swift"
   end
 
   test do
-    output = shell_output("#{bin}/swiftgen-assets #{pkgshare}/Images.xcassets").strip
-    assert_equal output, (pkgshare/"assets.swift").read.strip, "swiftgen-assets failed"
+    shell_output("#{bin}/swiftgen --version")
```
bfontaine: Use `system "#{bin}/swiftgen", "--version"` if you don’t use the output.

AliSoftware: Indeed I don't use the output, but I want all the other checks that Homebrew makes with `shell_output`, especially `chai` the command + ensuring the `exitstatus` is `0`, as per previous recommandation [here](https://github.com/Homebrew/homebrew/pull/44186#discussion_r39935952), which `system` doesn't, right?

bfontaine: `system` works exactly like `shell_output` except that it doesn’t capture the output; so that’s exactly what you’re looking for here :wink: 

It’s a wrapper, not the bare-bone Ruby `system`: https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/formula.rb#L1309

AliSoftware: Ha! Didn't know it was a wrapper too, thx for the info :wink: 

Just fixed it!

bfontaine:
Thank you for your contribution @AliSoftware; we appreciate it! :tada:

AliSoftware:
This installation seems to fail for some users because BrewBot brewed a bottle on your own servers — on which Xcode is installed in `/Applications/Xcode.app` and so is where my binary is searching for Swift dylibs when running — but on some users' machines (esp. on a CircleCI VM) `Xcode.app` is installed elsewhere and my binary can't find the Swift dylibs.

See https://github.com/AliSoftware/SwiftGen/issues/35#issuecomment-148491187

Any suggestion on how to solve this? Maybe add a post-script step where I add an `@rpath` to my binary (`install_name_tool -add_rpath …`) so it adds the path to the local installation of Xcode in the dylibs search path? Should I open a dedicated issue about this?

mikemcquaid:
@AliSoftware We probably need to add a `def pour_bottle` method which checks for the Xcode location before pouring the bottle. Could you try and open a pull request? This should help and I'm happy to walk you through anything else: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged

Thanks!

mikemcquaid:
@AliSoftware I see you already did! Thanks: https://github.com/Homebrew/homebrew/pull/45008

AliSoftware:
Yep indeed, thx!

PR #44972, opened 2015-10-14
nv - Lightweight utility to load context specific environment variables.

jcouture:


mikemcquaid:
I'm afraid this project isn't yet notable or old enough for Homebrew, sorry! I recommend creating a tap instead. Thanks!

bfontaine:
Some docs on the tap creation: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md

PR #44974, opened 2015-10-15
brew: check xcode license status

DomT4:
Code stolen right out of the INSTALL script.

Check for user agreement of the Xcode license before permitting any other brew usage to continue. This prevents the situation where people are instructed to "please re-run as root via sudo" on brew commands. The check can only fail when Xcode is installed & the active developer dir.

It turns out to be quicker to just run the command than checking the OS & if Xcode is installed first. Some benchmarks:

**WITH CHECK:**
* brew  0.16s user 0.04s system 99% cpu 0.201 total
* brew  0.17s user 0.06s system 45% cpu 0.491 total
* brew  0.16s user 0.04s system 87% cpu 0.235 total
* brew  0.16s user 0.04s system 99% cpu 0.204 total
* brew  0.15s user 0.04s system 98% cpu 0.200 total

**NO CHECK:**
* brew  0.15s user 0.04s system 73% cpu 0.256 total
* brew  0.15s user 0.04s system 98% cpu 0.192 total
* brew  0.14s user 0.04s system 90% cpu 0.199 total
* brew  0.14s user 0.04s system 99% cpu 0.180 total
* brew  0.15s user 0.04s system 99% cpu 0.186 total

**WITH `if OS.mac? && MacOS::Xcode.installed?` precondition:**
* brew  0.19s user 0.05s system 97% cpu 0.247 total
* brew  0.18s user 0.05s system 98% cpu 0.238 total
* brew  0.18s user 0.05s system 97% cpu 0.240 total
* brew  0.20s user 0.05s system 91% cpu 0.277 total
* brew  0.18s user 0.05s system 98% cpu 0.239 total

### DIFF
--- a/Library/brew.rb
+++ b/Library/brew.rb
@@ -25,18 +25,30 @@ elsif ARGV.first == "-v"
   exit 0 if ARGV.length == 1
 end
 
-# Check for bad xcode-select before anything else, because `doctor` and
-# many other things will hang
-# Note that this bug was fixed in 10.9
-if OS.mac? && MacOS.version < :mavericks && MacOS.active_developer_dir == "/"
-  odie <<-EOS.undent
-  Your xcode-select path is currently set to '/'.
-  This causes the `xcrun` tool to hang, and can render Homebrew unusable.
-  If you are using Xcode, you should:
-    sudo xcode-select -switch /Applications/Xcode.app
-  Otherwise, you should:
-    sudo rm -rf /usr/share/xcode-select
-  EOS
+# Check for bad xcode-select before other checks, because `doctor` and
+# many other things will hang. Note that this bug was fixed in 10.9
+if OS.mac?
+  if MacOS.version < :mavericks && MacOS.active_developer_dir == "/"
+    odie <<-EOS.undent
+    Your xcode-select path is currently set to '/'.
+    This causes the `xcrun` tool to hang, and can render Homebrew unusable.
+    If you are using Xcode, you should:
+      sudo xcode-select -switch /Applications/Xcode.app
+    Otherwise, you should:
+      sudo rm -rf /usr/share/xcode-select
+    EOS
+  end
+
+  # Check for user agreement of the Xcode license before permitting
+  # any other brew usage to continue. This prevents the situation where
+  # people are instructed to "please re-run as root via sudo" on brew commands.
+  # The check can only fail when Xcode is installed & the active developer dir.
+  if MacOS::Xcode.installed? && `/usr/bin/xcrun clang 2>&1` =~ /license/ && !$?.success?
+    odie <<-EOS.undent
+      You have not agreed to the Xcode license. Please resolve this by running:
+        sudo xcodebuild -license
+    EOS
+  end
 end
 
 case HOMEBREW_PREFIX.to_s
@@ -45,7 +57,7 @@ when "/", "/usr"
   abort "Cowardly refusing to continue at this prefix: #{HOMEBREW_PREFIX}"
 end
 
-if OS.mac? and MacOS.version < "10.6"
+if OS.mac? && MacOS.version < "10.6"
   abort <<-EOABORT.undent
     Homebrew requires Snow Leopard or higher. For Tiger and Leopard support, see:
     https://github.com/mistydemeo/tigerbrew
@@ -88,7 +100,7 @@ begin
   sudo_check = %w[ install link pin unpin upgrade ]
 
   if sudo_check.include? cmd
-    if Process.uid.zero? and not File.stat(HOMEBREW_BREW_FILE).uid.zero?
+    if Process.uid.zero? && !File.stat(HOMEBREW_BREW_FILE).uid.zero?
       raise <<-EOS.undent
         Cowardly refusing to `sudo brew #{cmd}`
         You can use brew with sudo, but only if the brew executable is owned by root.


DomT4:
CC @mikemcquaid for review & @xu-cheng for a considerably better way to do this :stuck_out_tongue_winking_eye:.

Library/brew.rb
```diff
@@ -39,13 +38,24 @@
...
+# Check for user agreement of the Xcode license before permitting
+# any other brew usage to continue. This prevents the situation where
+# people are instructed to "please re-run as root via sudo" on brew commands.
+# The check can only fail when Xcode is installed & the active developer dir.
+if `/usr/bin/xcrun clang 2>&1` =~ /license/ && !$?.success?
+  odie <<-EOS.undent
+    You have not agreed to the Xcode license. Please resolve this by running:
+      sudo xcodebuild -license
+  EOS
+end
```
xu-cheng: Can we put it outside of `brew.rb`? I think we should put it in `brew doctor` and use the check in `brew install`.

Another question is how will it behave for non-Xcode and non-CLT installation?

mikemcquaid: Yeh, non-Xcode is the concern for me. This does seem to be the same code as in the installer, though, so should probably be done there too.

DomT4: This happens:
```
xcode-select: note: no developer tools were found at '/Applications/Xcode.app', requesting install. Choose an option in the dialog to download the command line developer tools.
```
We could throw it inside an Xcode existing block though, easily enough, at the cost of a little extra time. 

I deliberately didn't put it in `brew doctor` or another subcommand because if you haven't agreed to the license this happens:
```
Vagrants-Mac:local vagrant$ brew doctor
Agreeing to the Xcode/iOS license requires admin privileges, please re-run as root via sudo.
Error: Failure while executing: /usr/bin/otool -L /usr/bin/install_name_tool
```
```
Vagrants-Mac:local vagrant$ brew install -fs libtool
Agreeing to the Xcode/iOS license requires admin privileges, please re-run as root via sudo.
Error: Failure while executing: /usr/bin/otool -L /usr/bin/install_name_tool
```
Which implies the `brew` command should be run as root, and we've seen a lot of people hitting this and being confused by it lately, which is why I kind of wanted to put it in `brew` so they literally couldn't run a brew command without handling the license situation first.

mikemcquaid: Yeh, Xcode existing block makes sense to me :+1:

mikemcquaid: Does that happen with all commands or just `doctor`? Would be good to maybe run with `--debug` and work out the backtrace that's trying to run that and throwing the exception there instead.

DomT4: > Does that happen with all commands or just doctor? 

You can run `info`, `desc`, `audit` (but not `style`), `cat`, `cleanup`, `create`, `edit`, `deps`, `fetch`, `irb`, `leaves`, `link`, `linkapps`, `list`, `prune`, `readall`, `pin` and `unpin`, `search` and `uses`.

`install`, `doctor` both fail, so does `update` (very loudly), `upgrade` fails, as does `tap`, as does `test-bot` relatively obviously, `pull` fails messily when it throws the license message & fails to apply the patch. `brew tests` breaks trying to compile the gems with the license failure.  

`config` partially fails:
```
Xcode: 7.0.1
CLT: N/A
Agreeing to the Xcode/iOS license requires admin privileges, please re-run as root via sudo.
Clang: N/A
X11: N/A
System Ruby: 2.0.0-p645
```
`gist-logs` partially fails:
```
Vagrants-Mac:~ vagrant$ brew gist-logs pkg-config
Agreeing to the Xcode/iOS license requires admin privileges, please re-run as root via sudo.
https://gist.github.com/2382bae6d49a1491be4d
```

mikemcquaid: Running any of those with `--debug` should help you get a backtrace. Can you paste that here so we can figure out the best place to put this call in the stack? Probably doesn't need to be in `brew.rb` in that case.

DomT4: Failure point seems to be [this](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/mach.rb#L166-L172) for `doctor` at least.

DomT4:
```
Vagrants-Mac:~ vagrant$ brew doctor --debug


Agreeing to the Xcode/iOS license requires admin privileges, please re-run as root via sudo.


Error: Failure while executing: /usr/bin/otool -L /usr/bin/install_name_tool
/usr/local/Library/Homebrew/mach.rb:170:in `parse_otool_L_output'
/usr/local/Library/Homebrew/mach.rb:163:in `initialize'
/usr/local/Library/Homebrew/mach.rb:187:in `new'
/usr/local/Library/Homebrew/mach.rb:187:in `mach_metadata'
/usr/local/Library/Homebrew/mach.rb:197:in `dynamically_linked_libraries'
/usr/local/Library/Homebrew/cmd/doctor.rb:391:in `check_for_bad_install_name_tool'
/usr/local/Library/Homebrew/cmd/doctor.rb:1294:in `block in doctor'
/usr/local/Library/Homebrew/cmd/doctor.rb:1292:in `each'
/usr/local/Library/Homebrew/cmd/doctor.rb:1292:in `doctor'
/usr/local/Library/brew.rb:133:in `<main>'
```

DomT4: Tap fails like this:
```
Error: Failure while executing: git clone https://github.com/Homebrew/homebrew-completions /usr/local/Library/Taps/homebrew/homebrew-completions --depth=1
/usr/local/Library/Homebrew/utils.rb:228:in `safe_system'
/usr/local/Library/Homebrew/cmd/tap.rb:34:in `install_tap'
/usr/local/Library/Homebrew/cmd/tap.rb:18:in `tap'
/usr/local/Library/brew.rb:133:in `<main>'
```

DomT4: Those two points seem to be the failures:
* Sending any calls to the system `git` prior to the license check.
* The `mach` block.

mikemcquaid: Yeh, guess it does need to live in `brew.rb` if even `git` is going to fail otherwise. :+1: as-is from me.

DomT4: You can't even run stuff like `git status` without agreeing to the license alas. As soon as you invoke `git` it license shouts at you.

I'll add the Xcode check in at least so we're not throwing the CLT dialogue at people who have neither Xcode or CLT installed.

xu-cheng: `update` and `tap` may relate to git. We can add some check in https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/utils/git.rb#L9

`install` and `upgrade` should belong to `doctor` check https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/install.rb#L135-L147

For `doctor` check, we can bypass some check if license check fails.

xu-cheng: `gist-log` may fail because it also calls `brew doctor`.

mikemcquaid: @xu-cheng I think at that point we may as well make the check global.

xu-cheng: @mikemcquaid I think we still need some check for `git`, especially if we want to support non-Xcode/CLT installation?

Maybe is a good idea to put the logic inside `OS::Mac::Xcode#accepted_license?`

mikemcquaid: I think given the code immediately above is very similar it makes sense to have it in `brew.rb`

DomT4: @xu-cheng If you want to play with a few ideas I don't mind if you steal the PR. I'll add the Xcode check in here and we can either move ahead with this PR or whatever ideas you've in mind after the CI has trundled back over this.

xu-cheng: I'm fine to add it in `brew.rb`. You should also enforce the condition with `OS.mac?`.

DomT4: Aye. `if OS.mac? && MacOS::Xcode.installed?` was the original check there; It was just 0.03 seconds slower so I left it out initially. Didn't occur to me calling `clang` with no CLT/Xcode installed would throw the CLT dialogue, despite having seen it a thousand times :smile_cat:

mikemcquaid: I'd maybe share the `OS.mac?` with the section immediately above it.

DomT4: Pushed :+1:.

Benchmarks:
* brew  0.18s user 0.05s system 97% cpu 0.233 total
* brew  0.18s user 0.05s system 97% cpu 0.229 total
* brew  0.17s user 0.05s system 98% cpu 0.226 total
* brew  0.18s user 0.05s system 98% cpu 0.233 total
* brew  0.18s user 0.05s system 98% cpu 0.230 total

Sharing the check knocked a tiny bit of the average there, which is a nice upside.

Library/brew.rb
```diff
@@ -25,18 +25,30 @@
...
-  odie <<-EOS.undent
-  Your xcode-select path is currently set to '/'.
-  This causes the `xcrun` tool to hang, and can render Homebrew unusable.
-  If you are using Xcode, you should:
-    sudo xcode-select -switch /Applications/Xcode.app
-  Otherwise, you should:
-    sudo rm -rf /usr/share/xcode-select
-  EOS
+# Check for bad xcode-select before other checks, because `doctor` and
+# many other things will hang. Note that this bug was fixed in 10.9
```
UniqMartin: Shouldn't this comment be inside the `if OS.mac?` block, right before `if MacOS.version < :mavericks && MacOS.active_developer_dir == "/"`? Otherwise, it gives the impression of applying to both of the following two checks.

mikemcquaid: :+1:

DomT4: Yes, but I can fix that on merge after the CI does its thing.

mikemcquaid: :shipit:

PR #44976, opened 2015-10-15
v8 4.6.85.28

DomT4:


mikemcquaid:
Travis failure looks legit.

DomT4:
Oh, er, yeah, not good: `[1]    13261 segmentation fault  v8 -e 'print("Hello World!")'`

DomT4:
Have built this (and the new version) pretty much every way V8 can be built, and it segfaults the same way every time, so going to poke upstream.

DomT4:
2 hours of poking later, identified the root cause, reported it upstream, stuck a slight hack in the formula which should work. Works locally, at least.

DomT4:
@jacobsa Any idea what's going on here with `gjstest`? The error may well be with the new V8 which has been all kinds of fun to try and get sorted and building/testing without failure.

The backtrace is:

```
Process 70702 launched: '/usr/local/bin/gjstest' (x86_64)
Process 70702 stopped
* thread #1: tid = 0xf370f, 0x00000001006d0d85 libv8.dylib`v8::internal::NativesCollection<(v8::internal::NativeType)0>::GetBuiltinsCount() + 11, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xc)
    frame #0: 0x00000001006d0d85 libv8.dylib`v8::internal::NativesCollection<(v8::internal::NativeType)0>::GetBuiltinsCount() + 11
libv8.dylib`v8::internal::NativesCollection<(v8::internal::NativeType)0>::GetBuiltinsCount:
->  0x1006d0d85 <+11>: movl   0xc(%rax), %eax
    0x1006d0d88 <+14>: popq   %rbp
    0x1006d0d89 <+15>: retq

libv8.dylib`v8::internal::NativesCollection<(v8::internal::NativeType)0>::GetDebuggerCount:
    0x1006d0d8a <+0>:  pushq  %rbp
(lldb) bt
* thread #1: tid = 0xf370f, 0x00000001006d0d85 libv8.dylib`v8::internal::NativesCollection<(v8::internal::NativeType)0>::GetBuiltinsCount() + 11, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xc)
  * frame #0: 0x00000001006d0d85 libv8.dylib`v8::internal::NativesCollection<(v8::internal::NativeType)0>::GetBuiltinsCount() + 11
    frame #1: 0x000000010047817d libv8.dylib`v8::internal::Heap::CreateInitialObjects() + 4439
    frame #2: 0x000000010047e454 libv8.dylib`v8::internal::Heap::CreateHeapObjects() + 34
    frame #3: 0x0000000100528330 libv8.dylib`v8::internal::Isolate::Init(v8::internal::Deserializer*) + 1244
    frame #4: 0x0000000100323189 libv8.dylib`v8::Isolate::New(v8::Isolate::CreateParams const&) + 313
    frame #5: 0x0000000100008f22 gjstest`gjstest::CreateIsolate() + 196
    frame #6: 0x000000010000e7dd gjstest`gjstest::RunTests(gjstest::NamedScripts const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >*) + 168
    frame #7: 0x0000000100011320 gjstest`main + 1940
    frame #8: 0x00007fff9a5995ad libdyld.dylib`start + 1
    frame #9: 0x00007fff9a5995ad libdyld.dylib`start + 1
```

DomT4:
Apparently the previously suspected V8 "bug" is actually a new intended thing when building with the `snapshot` enabled. Didn't seem to be in the documentation I read through but upstream confirmed on the bug tracker.

jacobsa:
Sorry, I'm not sure. Can you give me a link to the issue you reported, and ideally instructions for reproducing what you're seeing, including getting it into a debugger?

For what it's worth, v8 4.6.85.24 appears to work fine with gjstest in Google's source tree. Also for what it's worth, I share your frustration about how much of a pain in the ass v8 is to work with in terms of updates and build process. Ugh.

DomT4:
It's in the formula: https://code.google.com/p/v8/issues/detail?id=4511. Deliberate new feature apparently, in regards to needing to retain the `.bin` files in the same folder as the produced binaries.

In terms of the debugger, I've just been checking out my `v8` branch, doing `brew install v8 -fs && brew install gjstest` and then loading `lldb` with `/usr/local/bin/gjstest` as the `target`. You can also use `gdb` but it needs codesigning on OS X and that's a bit of a hassle.

jacobsa:
Thanks. So I'm not sure: is there anything gjstest-specific here? (I.e. is there anything I need to do?)

DomT4:
I'm not sure. V8 is running and testing fine now, but perhaps the test isn't doing enough compared to `gjstest`.

jacobsa:
Okay, I can reproduce and will try to take a look today (but am dealing with a sick child, so time is spotty). Apologies in advance if this winds up being a gjstest bug.

By the way, what does `-fs` do?

jacobsa:
It looks like the crash is [here](https://github.com/v8/v8/blob/3b00385a93215ce2a2af737fbea95309b40f08d9/src/snapshot/natives-external.cc#L203) in `NativesCollection<CORE>::GetBuiltinsCount`, where `NativesHolder<CORE>::set` has never been called, so the `holder_` variable is null and `get` returns null.

It looks like `set` is supposed to be called [here](https://github.com/v8/v8/blob/3b00385a93215ce2a2af737fbea95309b40f08d9/src/snapshot/natives-external.cc#L159), in `ReadNatives`, which reads from `natives_blob_`. That in turn is set by [`SetNativesFromFile`](https://github.com/v8/v8/blob/3b00385a93215ce2a2af737fbea95309b40f08d9/src/snapshot/natives-external.cc#L173), which is documented as

> Set the Natives (library sources) blob, as generated by js2c + the build system.

So this winds up looking a lot like the [issue](https://code.google.com/p/v8/issues/detail?id=4511) you reported (and therefore maybe you already knew all of the above).

Now I'm stuck, because I have no idea what this file is, whether it is installed, how it is supposed to be located, etc. It seems like v8 is simply not intended to be used as a pre-built library...

jacobsa:
v8.h does imply [here](https://github.com/v8/v8/blob/3b00385a93215ce2a2af737fbea95309b40f08d9/include/v8.h#L5978-L5979) that this behavior is optional, and the data can be compiled in:

> By default the startup data is linked into the V8 library, in which case this function is not meaningful.

It [looks like](https://github.com/v8/v8/blob/3b00385a93215ce2a2af737fbea95309b40f08d9/src/v8.cc#L129-L133) this is controlled with `V8_USE_EXTERNAL_STARTUP_DATA`?

DomT4:
> By the way, what does -fs do?

I meant `-s`, which is just shorthand for `--build-from-source`. `-f` is shorthand for `--force`.

I noticed you had a bit of a discussion with upstream on the changes made. Do you have any recommendations here?

jacobsa:
It [sounds like](https://code.google.com/p/v8/issues/detail?id=4511#c3) you want to set `GYP_DEFINES="v8_use_external_startup_data=0"`. I have no idea how to do that though, in the context of homebrew or otherwise. Does that mean anything to you?

DomT4:
Yeah we can set that easily enough. Does that negate the need for the hack we have in place now around keeping the `.bin` files or does it just resolve the build issues we've seen on formula which depend on `v8`? Or both, or neither :smile_cat:.

jacobsa:
It sounds like it means that at least one (both?) `.bin` files is not needed (or even produced?). From examining the v8 code I would guess that it would solve the gjstest issue, but I can't say for sure.

DomT4:
Alright, I'll poke around a bit. Thanks for talking to upstream about this.

DomT4:
Your suggestion seems to work :+1:.

Updated to the latest V8 as well, since a new stable Chrome landed yesterday with a minor V8 bump. `gjstest`'s test from bottle breaks, but if you build it from source it passes so I've bumped the revision there.

PR #44982, opened 2015-10-15
test-bot: merge formula's bottle before test

xu-cheng:
Fixes the bug described in https://github.com/Homebrew/homebrew-science/pull/2866#issuecomment-148298146

### DIFF
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -590,6 +590,11 @@ module Homebrew
           if bottle_step.passed? && bottle_step.has_output?
             bottle_filename =
               bottle_step.output.gsub(/.*(\.\/\S+#{bottle_native_regex}).*/m, '\1')
+            bottle_rb_filename = bottle_filename.gsub(/\.(\d+\.)?tar\.gz$/, ".rb")
+            bottle_merge_args = ["--merge", "--write", "--no-commit", bottle_rb_filename]
+            bottle_merge_args << "--tap=#{@tap}" if @tap
+            bottle_merge_args << "--keep-old" if ARGV.include? "--keep-old"
+            test "brew", "bottle", *bottle_merge_args
             test "brew", "uninstall", "--force", canonical_formula_name
             if unchanged_build_dependencies.any?
               test "brew", "uninstall", "--force", *unchanged_build_dependencies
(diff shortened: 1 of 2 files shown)


mikemcquaid:
:+1:

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -590,6 +590,11 @@ def formula(formula_name)
           if bottle_step.passed? && bottle_step.has_output?
             bottle_filename =
               bottle_step.output.gsub(/.*(\.\/\S+#{bottle_native_regex}).*/m, '\1')
+            bottle_rb_filename = bottle_filename.gsub(/\.(\d+\.)?tar\.gz$/, ".rb")
+            bottle_merge_args = ["--merge", "--write", "--no-commit", bottle_rb_filename]
```
xu-cheng: I'm thinking we may add `--root-url=file://#{Dir.pwd}` to speed up the build process.

xu-cheng: @mikemcquaid Any comment on this idea?

xu-cheng: Ignore me, `--root-url` isn't used in merge process.

mikemcquaid: :+1:

mikemcquaid: (as in :+1:: it's not used)

xu-cheng: But we can make it use it if we want :wink: 

Therefore, it may help to speed up process when building multiple formulae depend on each other. For example, gcc.

mikemcquaid: Given they are cached locally and these machines all have absurdly fast internet connections I'm not convinced it'll make much of a difference.

xu-cheng: I'm not talking avoid re-download but avoid rebuilding from source. From what I'm seeing, some formula will be built twice or more when we process PR that bumps revisions. The root cause is [we uninstall the formula after test it](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/test-bot.rb#L621). So if another formula depends on the previous formula, it will be rebuilt.

mikemcquaid: Ah, I see. Yeh, I guess we could so something smarter there (or just not uninstall it and leave that for `git clean -xdf` to handle)

PR #39706, opened 2015-05-13
Add formula for mg3a

paaguti:
mg3a is a version of mg that supports unicode and other nice features.
Good micro-clone for emacs.

PR #44983, opened 2015-10-15
mg: add test

bfontaine:
This passes locally; if it works on the CI we could use it for #39706 too.

### DIFF
--- a/Library/Formula/mg.rb
+++ b/Library/Formula/mg.rb
@@ -18,4 +18,18 @@ class Mg < Formula
     doc.install "tutorial"
     man1.install "mg.1"
   end
+
+  test do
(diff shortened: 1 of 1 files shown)


bfontaine:
Nope :disappointed:

Library/Formula/mg.rb
```diff
@@ -18,4 +18,18 @@ def install
     doc.install "tutorial"
     man1.install "mg.1"
   end
+
+  test do
+    (testpath/"command.sh").write <<-EOS.undent
+      #!/usr/local/bin/expect -f
```
DomT4: Shouldn't this be `/usr/bin/expect`?

bfontaine: Right; this one is Homebrew’s.

PR #44988, opened 2015-10-15
MacOS: full_version added

bfontaine:
cc @mikemcquaid @xu-cheng for review.

### DIFF
--- a/Library/Homebrew/global.rb
+++ b/Library/Homebrew/global.rb
@@ -26,12 +26,12 @@ else
 end
 RUBY_BIN = RUBY_PATH.dirname
 
+MACOS_FULL_VERSION = MacOS.full_version
+MACOS_VERSION = MacOS.version
+
 if RUBY_PLATFORM =~ /darwin/
-  MACOS_FULL_VERSION = `/usr/bin/sw_vers -productVersion`.chomp
-  MACOS_VERSION = MACOS_FULL_VERSION[/10\.\d+/]
   OS_VERSION = "OS X #{MACOS_FULL_VERSION}"
 else
-  MACOS_FULL_VERSION = MACOS_VERSION = "0"
   OS_VERSION = RUBY_PLATFORM
 end
 
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -12,7 +12,19 @@ module OS
     # This can be compared to numerics, strings, or symbols
     # using the standard Ruby Comparable methods.
     def version
-      @version ||= Version.new(MACOS_VERSION)
+      @version ||= Version.new(full_version.to_s[/10\.\d+/])
+    end
+
+    # This can be compared to numerics, strings, or symbols
+    # using the standard Ruby Comparable methods.
+    def full_version
+      @full_version ||= Version.new(
+        if RUBY_PLATFORM =~ /darwin/
+          `/usr/bin/sw_vers -productVersion`.chomp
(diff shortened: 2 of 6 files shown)


mikemcquaid:
:+1:

xu-cheng:
:+1: 

Also about style problem, should we consider to enforce `OS::Mac` over `MacOS`? According to [the code comment](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/os/mac.rb#L10), `MacOS` is only for compatibility purpose.

Library/Homebrew/os/mac.rb
```diff
@@ -12,7 +12,19 @@ module Mac
...
     def version
-      @version ||= Version.new(MACOS_VERSION)
+      @version ||= Version.new(full_version.to_s[/10\.\d+/])
+    end
+
+    # This can be compared to numerics, strings, or symbols
+    # using the standard Ruby Comparable methods.
+    def full_version
+      @full_version ||= Version.new(
+        if RUBY_PLATFORM =~ /darwin/
```
xu-cheng: I think  it may be better to check OS type in `global.rb`. After all, it doesn't make sense to check OS type in `OS::Mac#full_version`.

Library/Homebrew/global.rb
```diff
@@ -26,12 +26,12 @@
...
+MACOS_VERSION = MacOS.version
+
 if RUBY_PLATFORM =~ /darwin/
-  MACOS_FULL_VERSION = `/usr/bin/sw_vers -productVersion`.chomp
-  MACOS_VERSION = MACOS_FULL_VERSION[/10\.\d+/]
   OS_VERSION = "OS X #{MACOS_FULL_VERSION}"
 else
-  MACOS_FULL_VERSION = MACOS_VERSION = "0"
   OS_VERSION = RUBY_PLATFORM
 end
```
xu-cheng: Let's make it:
```ruby
if OS.mac?
  MACOS_FULL_VERSION = OS::Mac.full_version.to_s
  MACOS_VERSION = OS::Mac.version.to_s
  OS_VERSION = "OS X #{MACOS_FULL_VERSION}"
else
  MACOS_FULL_VERSION = MACOS_VERSION = "0"
  OS_VERSION = RUBY_PLATFORM
end
```

Library/Homebrew/os/mac.rb
bfontaine: With the `||=` it’ll only check it once. I think we should keep OS-related code under `os/` but that’s just a matter of style.

xu-cheng: I agree that we should keep os relate code in `os`. It just feels weird to have OS detect code in `OS::Mac`, which should means OS detection is finished. 

How about put them outside of `global.rb` into `Library/os.rb` as
```ruby 
if OS.mac?
  require "os/mac"
  ::MACOS_FULL_VERSION = OS::Mac.full_version.to_s
  ::MACOS_VERSION = OS::Mac.version.to_s
  ::OS_VERSION = "OS X #{MACOS_FULL_VERSION}"
  ISSUES_URL = ...
...

bfontaine: > It just feels weird to have OS detect code in `OS::Mac`, which should means OS detection is finished.

Right, I didn’t think of this.

bfontaine:
(I killed the Jenkins build)

bfontaine:
Closing in favor of #45021.

Issue #44797, opened 2015-10-09
Promote FStar from Homebrew/Science


beurdouche:
@DomT4, @dpo, @bfontaine, @dunn 
Hello Hello !
Last week, we decided to move FStar from the Core to the Science tap (#44341) because it was dependant on Z3. =)
Now Z3 was promoted to Core #44652 and all similar programming languages like Rust, Coq are in Core.

Do you mind if we move FStar back to Core ? =)

DomT4:
If you can get @Homebrew/science and another one/two of @Homebrew/owners to sign off on it, yes, feel free. No objections personally. You'd need to wait on the `z3` formula to be merged here, I'd presume.

dpo:
Apparently
* `z3` has `fstar` as unique dependent
* `fstar` has no dependent
* `cvc4` has `cryptol` as unique dependent, which is a cryptography package
* `cryptol` has no dependent.

So `z3 -> fstar` and `cvc4 -> cryptol` are two pretty small dependency graphs.

Whether `cryptol` is a "scientific" package is debatable but the home page says

> A Cryptol implementation of an algorithm resembles its mathematical specification more closely than an implementation in a general purpose language.

So either way, they should probably all be in core, or all in science.

On a side note:
```
$ brew deps fstar
Error: No available formula for fsharp 
```

beurdouche:
@Dpo : thanks, it is fixed already locally and ready to push as soon as a decision is reached here. =)

bfontaine:
I don’t have any hard objection; but why do you want to move formulae from homebrew/science to the core?

beurdouche:
I submitted it in core in the first place because I believe that such a programming language is a more general usage than a domain specific tools. I do not hesitate to submit scientific tools to `science` anyway (see `proverif`, `cryptoverif`, soon `flextls`) but I don't believe FStar should be in `science`, for this reason and because of it's increasing use. I am not opposed to F* staying in `science` but I would really prefer having it where similar languages and tools are ie. in core.

bfontaine:
Thank you for the explanation. I’m :+1: here.

beurdouche:
Should I proceed ?

bfontaine:
Yep.

DomT4:
Closing in favour of #44989.

PR #44989, opened 2015-10-15
New formula : FStar 0.9.1.1 (#44797)

beurdouche:


Library/Formula/fstar.rb
```diff
@@ -0,0 +1,92 @@
...
+      modules << "#{r.name}=#{r.version}"
+    end
+    system "opam", "install", *modules
+
+    system "make", "-C", "src/"
+    system "opam", "config", "exec", "--",
+    "make", "-C", "src/ocaml-output/"
+
+    bin.install "src/ocaml-output/fstar.exe"
+    prefix.install "README.md"
```
DomT4: Is this not happening automatically?

Library/Formula/fstar.rb
```diff
@@ -0,0 +1,92 @@
...
+    sha256 "aafaba4f7453c38347ff5269c6fd4f4c243ae2bceeeb5e10b9dab89329905946"
+  end
+
+  resource "batteries" do
+    url "https://github.com/ocaml-batteries-team/batteries-included/archive/v2.3.1.tar.gz"
+    sha256 "df778b90fcdb26288d9d92a86e51dd75d6bb7c6e41888c748c7508e8ea58b1d4"
+  end
+
+  def install
+    ENV.deparallelize
```
DomT4: Can we report this upstream somewhere? It shouldn't be necessary ideally.

Library/Formula/fstar.rb
```diff
@@ -0,0 +1,92 @@
+class Fstar < Formula
+  desc "Language with a type system for program verification"
+  homepage "https://www.fstar-lang.org/"
+  url "https://github.com/FStarLang/FStar.git",
+    :tag => "v0.9.1.1",
+    :revision => "c41ea17778ff2fef4e75d56604ce1804e1985f33"
+  head "https://github.com/FStarLang/FStar.git"
+
+  depends_on "mono"   => :build
```
DomT4: Don't worry about equal spacing around the `=>`. I heavily prefer out of sync lines over random spacing, personally. One space is fine. `depends_on "mono" => :build` and such.

Library/Formula/fstar.rb
beurdouche: It's not from F* but from Opam. I'll make sure it gets reported and reference it here.

DomT4: Ah of course. Remember this from elsewhere now. Thanks for kicking it to the right upstream. It may be worth adding a comment in-formula that the line isn't an `fstar` problem.

beurdouche:
@DomT4, done =)

beurdouche:
Oups, mistaken clic on close instead of comment...

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/939fd91fd8a64a7f081a5cbbff72060799583461. Thank you for your contribution to Homebrew @beurdouche! :smiley_cat:

beurdouche:
@DomT4, Thanks to you =)
> Merged in 939fd91 <https://github.com/Homebrew/homebrew/commit/939fd91fd8a64a7f081a5cbbff72060799583461>. Thank you for your contribution to Homebrew @beurdouche <https://github.com/beurdouche>! 
>

Library/Formula/fstar.rb
```diff
@@ -0,0 +1,91 @@
...
+    prefix.install_symlink libexec/"contrib"
+    prefix.install_symlink libexec/"examples"
+    prefix.install_symlink libexec/"tutorial"
+    prefix.install_symlink libexec/"src"
+    prefix.install_symlink libexec/"licenses"
+  end
+
+  def caveats; <<-EOS.undent
+    F* standard library is available in #{prefix}/stdlib:
+    - alias fstar='fstar.exe --include #{prefix}/stdlib --prims prims.fst'
```
apjanke: Are there cases where an `fstar` user wouldn't include the `stdlib`? If this is needed for basic operation, maybe the formula should include an `fstar` wrapper script instead of asking the user to set up an alias. The alias as written will break upon version upgrades of `fstar`, too.

beurdouche: > Are there cases where an fstar user wouldn't include the stdlib? If this is needed for basic operation, maybe the formula should include an fstar wrapper script instead of asking the user to set up an alias. The alias as written will break upon version upgrades of fstar, too.
> 
Yes, you are right. I didn’t realize that…
I’ll think about it and try to fix that when updating the formula (soon). Thanks ! B.

PR #44991, opened 2015-10-15
openvdb 3.1

jehutymax:
Version update and added the vdb_test target explictly since it's not part of the install target.

### DIFF
--- a/Library/Formula/openvdb.rb
+++ b/Library/Formula/openvdb.rb
@@ -1,8 +1,8 @@
 class Openvdb < Formula
   desc "Sparse volume processing toolkit"
   homepage "http://www.openvdb.org/"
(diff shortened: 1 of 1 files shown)


Library/Formula/openvdb.rb
```diff
@@ -95,6 +95,10 @@ def install
 
     cd "openvdb" do
       system "make", "install", *args
+      if build.with? "tests"
+        system "make", "vdb_test", *args
+        cp "#{buildpath}/openvdb/vdb_test", "#{bin}"
```
bfontaine: Use `bin.install "vdb_test"` here instead.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/22694340a0ac77fd36d811788314ade0e7f5ed2a. Thank you for your contribution to Homebrew @jehutymax; we appreciate it! :smiley_cat:

PR #44995, opened 2015-10-15
tap-info: tweak output for two edge cases

UniqMartin:
Output of `brew tap-info repo/user` for a tap with neither commands nor formulae:

- Old: `repo/user: unpinned, `
- New: `repo/user: unpinned, no formulae/commands`

Output of `brew tap-info repo/user` for a tap with both commands and formulae:

- Old: `repo/user: unpinned, 1337 formulae 42 commands `
- New: `repo/user: unpinned, 1337 formulae, 42 commands`


mikemcquaid:
Thanks again @UniqMartin!

PR #44999, opened 2015-10-15
commands: ensure external commands are executable

UniqMartin:
For consistency with `brew command` and the logic in `brew.sh` (both use `which` to find/validate an external command), we need to filter files that are not executable.

Otherwise `brew commands` and thus bash completion will offer commands that will produce an error when attempting to use them.

### DIFF
--- a/Library/Homebrew/cmd/commands.rb
+++ b/Library/Homebrew/cmd/commands.rb
@@ -38,6 +38,7 @@ module Homebrew
 
   def external_commands
     paths.flat_map { |p| Dir["#{p}/brew-*"] }.
+      select { |f| File.executable?(f) }.
       map { |f| File.basename(f, ".rb")[5..-1] }.
       reject { |f| f =~ /\./ }.
       sort


mikemcquaid:
:+1: very good. I do wonder if we can combine the logic between these two somehow, though?

UniqMartin:
I'm not entirely sure about what “these two” refers to. Are you talking about some code duplication between [brew.rb#L132-L144](https://github.com/Homebrew/homebrew/blob/31db1e66c71a2378c847070d58f143af4f39b068/Library/brew.rb#L132-L144) and [command.rb#L7-L15](https://github.com/Homebrew/homebrew/blob/31db1e66c71a2378c847070d58f143af4f39b068/Library/Homebrew/cmd/command.rb#L7-L15)?

mikemcquaid:
> Are you talking about some code duplication between brew.rb#L132-L144 and command.rb#L7-L15?

Yup!

UniqMartin:
I guess it could be done. I can give it a try and see if the result is more elegant than the current solution, of which I'm not entirely sure yet. But surely this will be stuff for a different PR or would you like to continue the discussion here?

mikemcquaid:
Yup, will merge this as-is for now, thanks!

PR #45000, opened 2015-10-15
haskell-stack v0.1.6.0

borsboom:
Hope you don't mind my submitting this PR.  I'm the release manager for Stack and noticed this here, so figured I might as well update it right away.

Side note: I'm thinking it might be better to use the sdist from Hackage (http://hackage.haskell.org/package/stack-0.1.6.0/stack-0.1.6.0.tar.gz), or just `cabal install stack-0.1.6.0` directly, since the .cabal file on Hackage will have more precise version bounds which reduces the risk of bugs being introduced due to using the wrong Hackage dependency versions.

Ping @superlukas @fboyer @bjpbakker

### DIFF
--- a/Library/Formula/haskell-stack.rb
+++ b/Library/Formula/haskell-stack.rb
@@ -5,8 +5,8 @@ class HaskellStack < Formula
 
   desc "The Haskell Tool Stack"
   homepage "https://www.stackage.org"
-  url "https://github.com/commercialhaskell/stack/archive/v0.1.6.0.tar.gz"
-  sha256 "71014893d9a01f33c8e7400357813f610cc264a011d5e6af3ca8f9b598940d36"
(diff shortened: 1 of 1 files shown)


superlukas:
Using the tarball from Hackage is a great idea. I think the Homebrew peeps may have a problem with us using cabal-install directly. See [Acceptable Formulae](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Acceptable-Formulae.md#we-dont-like-install-scripts-that-download-things).

borsboom:
Alright, updated the PR to use the Hackage sdist.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `haskell-stack 0.1.6.0`.

PR #45001, opened 2015-10-15
Replacement of cctools for dylib relocation with ruby_macho (see: #40721, #41021)

woodruffw:
CCing @mikemcquaid, @mistydemeo

This PR aims to replace the system of hooking in the `cctools` formula to perform dylib relocation on systems without developer tools with a pure-Ruby implementation of a Mach-O parser. The ultimate goal is to have Homebrew be able to perform dylib operations like rpath modification and metadata retrieval (for normal and "Fat" binaries) with no dependency on external utilities (`otool`, `install_name_tool`) whatsoever.

Previous related work: #40721, #41021

Since my Mach-O parser is still incomplete, I'm going to organize these changes so that they are only activated when an environment variable is set (`$HOMEBREW_RUBY_MACHO`).

Current progress:
- [x] Branch code so that ruby_macho is only used when `$HOMEBREW_RUBY_MACHO` is set.
- [x] Include `ruby_macho` in `vendor/`
- [x] Fill in branched stubs from step 1 with `ruby_macho` calls.
- [x] Remove branches, prune old code depending on execution of `otool`, `install_name_tool`.
- [x] Either remove old system of relocation entirely or quarantine it with an environment check.

Best,
William

### DIFF
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -70,16 +70,6 @@ class Keg
     end
   end
 
-  def change_dylib_id(id, file)
-    puts "Changing dylib ID of #{file}\n  from #{file.dylib_id}\n    to #{id}" if ARGV.debug?
-    install_name_tool("-id", id, file)
-  end
-
-  def change_install_name(old, new, file)
-    puts "Changing install name in #{file}\n  from #{old}\n    to #{new}" if ARGV.debug?
-    install_name_tool("-change", old, new, file)
-  end
-
   # Detects the C++ dynamic libraries in place, scanning the dynamic links
   # of the files within the keg.
   # Note that this doesn't attempt to distinguish between libstdc++ versions,
@@ -108,16 +98,6 @@ class Keg
         yield file if hardlinks.add? file.stat.ino
       end
     end
-  end
-
-  def install_name_tool(*args)
-    @require_install_name_tool = true
-    tool = MacOS.install_name_tool
-    system(tool, *args) || raise(ErrorDuringExecution.new(tool, args))
-  end
-
-  def require_install_name_tool?
-    !!@require_install_name_tool
   end
 
   # If file is a dylib or bundle itself, look for the dylib named by
--- a/Library/Homebrew/os/mac/pathname.rb
+++ b/Library/Homebrew/os/mac/pathname.rb
@@ -1,5 +1,9 @@
-require "os/mac/mach"
-
 class Pathname
-  include MachO
+  if ENV["HOMEBREW_RUBY_MACHO"]
+    require "os/mac/ruby_mach"
+    include RubyMachO
+  else
+    require "os/mac/cctools_mach"
+    include CctoolsMachO
+  end
 end
--- a/Library/Homebrew/os/mac/ruby_mach.rb
+++ b/Library/Homebrew/os/mac/ruby_mach.rb
@@ -1,0 +1,103 @@
+require "vendor/macho/macho"
+require "os/mac/architecture_list"
+
+module RubyMachO
+  # @private
+  def macho
+    @macho ||= begin
+      MachO.open(to_s)
+    end
+  end
+
+  # @private
+  def mach_data
+    @mach_data ||= begin
+      machos = []
+      mach_data = []
+
+      if MachO.fat_magic?(macho.magic)
+        machos = macho.machos
+      else
+        machos << macho
+      end
+
+      machos.each do |m|
+        arch = case m.cputype
+        when "CPU_TYPE_I386" then :i386
+        when "CPU_TYPE_X86_64" then :x86_64
+        when "CPU_TYPE_POWERPC" then :ppc7400
+        when "CPU_TYPE_POWERPC64" then :ppc64
+        else :dunno
+        end
+
+        type = case m.filetype
+        when "MH_EXECUTE" then :executable
+        when "MH_DYLIB" then :dylib
+        when "MH_BUNDLE" then :bundle
+        else :dunno
+        end
+
+        mach_data << { :arch => arch, :type => type }
+      end
+
+      mach_data
+    rescue
+      []
+    end
+  end
+
+  def archs
+    mach_data.map { |m| m.fetch :arch }.extend(ArchitectureListExtension)
+  end
+
+  def arch
+    case archs.length
+    when 0 then :dunno
+    when 1 then archs.first
+    else :universal
+    end
+  end
+
+  def universal?
+    arch == :universal
+  end
+
+  def i386?
+    arch == :i386
+  end
+
+  def x86_64?
+    arch == :x86_64
+  end
+
+  def ppc7400?
+    arch == :ppc7400
+  end
+
+  def ppc64?
+    arch == :ppc64
+  end
+
+  # @private
+  def dylib?
+    mach_data.any? { |m| m.fetch(:type) == :dylib }
+  end
+
+  # @private
+  def mach_o_executable?
+    mach_data.any? { |m| m.fetch(:type) == :executable }
+  end
+
+  # @private
+  def mach_o_bundle?
+    mach_data.any? { |m| m.fetch(:type) == :bundle }
+  end
+
+  def dynamically_linked_libraries
+    macho.linked_dylibs
+  end
+
+  def dylib_id
+    macho.dylib_id
+  end
+end
--- a/Library/Homebrew/vendor/README.md
+++ b/Library/Homebrew/vendor/README.md
@@ -1,0 +1,53 @@
+Vendored Dependencies
+=====================
+
+* [okjson](https://github.com/kr/okjson), version 43.
+
+* [ruby-macho](https://github.com/woodruffw/ruby-macho), version 0.2.2.
+
+## Licenses:
+
+### okjson
+
+> Copyright 2011, 2012 Keith Rarick
+>
+> Permission is hereby granted, free of charge, to any person obtaining a copy
+> of this software and associated documentation files (the "Software"), to deal
+> in the Software without restriction, including without limitation the rights
+> to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+> copies of the Software, and to permit persons to whom the Software is
+> furnished to do so, subject to the following conditions:
+>
+> The above copyright notice and this permission notice shall be included in
+> all copies or substantial portions of the Software.
+>
+> THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+> IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+> FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+> AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+> LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+> OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+> THE SOFTWARE.
+
+### ruby-macho
+
+> The MIT License
+> Copyright (c) 2015 William Woodruff <william @ tuffbizz.com>
+>
+> Permission is hereby granted, free of charge, to any person obtaining a copy
+> of this software and associated documentation files (the "Software"), to deal
+> in the Software without restriction, including without limitation the rights
+> to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+> copies of the Software, and to permit persons to whom the Software is
+> furnished to do so, subject to the following conditions:
+>
+> The above copyright notice and this permission notice shall be included in
+> all copies or substantial portions of the Software.
+>
+> THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+> IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+> FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+> AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+> LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+> OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+> THE SOFTWARE.
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -4,6 +4,7 @@ require "os/mac/xcode"
 require "os/mac/xquartz"
 require "os/mac/pathname"
 require "os/mac/sdk"
+require "os/mac/keg"
 
 module OS
   module Mac
--- a/Library/Homebrew/os/mac/architecture_list.rb
+++ b/Library/Homebrew/os/mac/architecture_list.rb
@@ -1,0 +1,52 @@
+module ArchitectureListExtension
+  # @private
+  def fat?
+    length > 1
+  end
+
+  # @private
+  def intel_universal?
+    intersects_all?(Hardware::CPU::INTEL_32BIT_ARCHS, Hardware::CPU::INTEL_64BIT_ARCHS)
+  end
+
+  # @private
+  def ppc_universal?
+    intersects_all?(Hardware::CPU::PPC_32BIT_ARCHS, Hardware::CPU::PPC_64BIT_ARCHS)
+  end
+
+  # Old-style 32-bit PPC/Intel universal, e.g. ppc7400 and i386
+  # @private
+  def cross_universal?
+    intersects_all?(Hardware::CPU::PPC_32BIT_ARCHS, Hardware::CPU::INTEL_32BIT_ARCHS)
+  end
+
+  # @private
+  def universal?
+    intel_universal? || ppc_universal? || cross_universal?
+  end
+
+  def ppc?
+    (Hardware::CPU::PPC_32BIT_ARCHS+Hardware::CPU::PPC_64BIT_ARCHS).any? { |a| self.include? a }
+  end
+
+  # @private
+  def remove_ppc!
+    (Hardware::CPU::PPC_32BIT_ARCHS+Hardware::CPU::PPC_64BIT_ARCHS).each { |a| delete a }
+  end
+
+  def as_arch_flags
+    collect { |a| "-arch #{a}" }.join(" ")
+  end
+
+  def as_cmake_arch_flags
+    join(";")
+  end
+
+  protected
+
+  def intersects_all?(*set)
+    set.all? do |archset|
+      archset.any? { |a| self.include? a }
+    end
+  end
+end
--- a/Library/Homebrew/os/mac/cctools_keg.rb
+++ b/Library/Homebrew/os/mac/cctools_keg.rb
@@ -1,0 +1,21 @@
+module CctoolsKeg
+  def install_name_tool(*args)
+    @require_install_name_tool = true
+    tool = MacOS.install_name_tool
+    system(tool, *args) || raise(ErrorDuringExecution.new(tool, args))
+  end
+
+  def require_install_name_tool?
+    !!@require_install_name_tool
+  end
+
+  def change_dylib_id(id, file)
+    puts "Changing dylib ID of #{file}\n  from #{file.dylib_id}\n    to #{id}" if ARGV.debug?
+    install_name_tool("-id", id, file)
+  end
+
+  def change_install_name(old, new, file)
+    puts "Changing install name in #{file}\n  from #{old}\n    to #{new}" if ARGV.debug?
+    install_name_tool("-change", old, new, file)
+  end
+end
--- a/Library/Homebrew/os/mac/cctools_mach.rb
+++ b/Library/Homebrew/os/mac/cctools_mach.rb
@@ -1,57 +1,6 @@
-module ArchitectureListExtension
-  # @private
-  def fat?
-    length > 1
-  end
+require "os/mac/architecture_list"
 
-  # @private
-  def intel_universal?
-    intersects_all?(Hardware::CPU::INTEL_32BIT_ARCHS, Hardware::CPU::INTEL_64BIT_ARCHS)
-  end
-
-  # @private
-  def ppc_universal?
-    intersects_all?(Hardware::CPU::PPC_32BIT_ARCHS, Hardware::CPU::PPC_64BIT_ARCHS)
-  end
-
-  # Old-style 32-bit PPC/Intel universal, e.g. ppc7400 and i386
-  # @private
-  def cross_universal?
-    intersects_all?(Hardware::CPU::PPC_32BIT_ARCHS, Hardware::CPU::INTEL_32BIT_ARCHS)
-  end
-
-  # @private
-  def universal?
-    intel_universal? || ppc_universal? || cross_universal?
-  end
-
-  def ppc?
-    (Hardware::CPU::PPC_32BIT_ARCHS+Hardware::CPU::PPC_64BIT_ARCHS).any? { |a| self.include? a }
-  end
-
-  # @private
-  def remove_ppc!
-    (Hardware::CPU::PPC_32BIT_ARCHS+Hardware::CPU::PPC_64BIT_ARCHS).each { |a| delete a }
-  end
-
-  def as_arch_flags
-    collect { |a| "-arch #{a}" }.join(" ")
-  end
-
-  def as_cmake_arch_flags
-    join(";")
-  end
-
-  protected
-
-  def intersects_all?(*set)
-    set.all? do |archset|
-      archset.any? { |a| self.include? a }
-    end
-  end
-end
-
-module MachO
+module CctoolsMachO
   # @private
   OTOOL_RX = /\t(.*) \(compatibility version (?:\d+\.)*\d+, current version (?:\d+\.)*\d+\)/
 
@@ -77,8 +26,6 @@ module MachO
         end
       when 0xcefaedfe, 0xcffaedfe, 0xfeedface, 0xfeedfacf # Single arch
         offsets << 0
-      when 0x7f454c46 # ELF
-        mach_data << { :arch => :x86_64, :type => :executable }
       else
         raise "Not a Mach-O binary."
       end
--- a/Library/Homebrew/os/mac/hardware.rb
+++ b/Library/Homebrew/os/mac/hardware.rb
@@ -1,4 +1,4 @@
-require "os/mac/mach"
+require "os/mac/pathname"
 
 module MacCPUs
   OPTIMIZATION_FLAGS = {
--- a/Library/Homebrew/os/mac/keg.rb
+++ b/Library/Homebrew/os/mac/keg.rb
@@ -1,0 +1,10 @@
+class Keg
+  if ENV["HOMEBREW_RUBY_MACHO"]
+    require "os/mac/ruby_keg"
+    include RubyKeg
+  else
+    require "os/mac/cctools_keg"
+    include CctoolsKeg
+  end
+end
+
--- a/Library/Homebrew/os/mac/ruby_keg.rb
+++ b/Library/Homebrew/os/mac/ruby_keg.rb
@@ -1,0 +1,19 @@
+require "vendor/macho/macho"
+
+module RubyKeg
+  def change_dylib_id(id, file)
+    @require_install_name_tool = true
+    puts "Changing dylib ID of #{file}\n  from #{file.dylib_id}\n    to #{id}" if ARGV.debug?
+    MachO::Tools.change_dylib_id(file, id)
+  end
+
+  def change_install_name(old, new, file)
+    @require_install_name_tool = true
+    puts "Changing install name in #{file}\n  from #{old}\n    to #{new}" if ARGV.debug?
+    MachO::Tools.change_install_name(file, old, new)
+  end
+
+  def require_install_name_tool?
+    !!@require_install_name_tool
+  end
+end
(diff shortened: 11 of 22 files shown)


mikemcquaid:
Thanks for this!

woodruffw:
No problem! I'll move the parser code into `vendor` tonight.

DomT4:
It's a nit, but it looks like your editor has indented with tabs instead of spaces throughout. Presume not intentional.

woodruffw:
Yeah, unintentional. Want me to convert to Homebrew's spacing style?

DomT4:
Ideally, if you don't mind :). We use the two space separator generally & it's nice to keep consistency unless it causes breakage, etc.

woodruffw:
Not a problem! I use tabs in my own projects, so I'll convert these and set something up to correct the indentation whenever I update the vendored copies.

woodruffw:
I see that `parse_otool_L_output` in *mach.rb* takes the `path` attribute and stores it in `ENV["HOMEBREW_MACH_O_FILE"]` - is there a particular reason for this or can I skip setting/unsetting this part of the environment and just use `path` directly?

mikemcquaid:
> ENV["HOMEBREW_MACH_O_FILE"] - is there a particular reason for this or can I skip setting/unsetting this part of the environment and just use path directly?

I'm not 100% sure what the reasoning is but let's leave it as-is for now and explore more later.

Library/Homebrew/mach.rb
```diff
@@ -160,7 +162,19 @@ class Metadata
...
+    def dylib_id_and_dylibs
+      if ENV["HOMEBREW_RUBY_MACHO"]
+        file = MachO.open(path.expand_path.to_s)
+        id = file.dylib_id
+        dylibs = file.linked_dylibs
+      else
+        id, dylibs = parse_otool_L_output
+      end
+
+      return id, dylibs
```
mikemcquaid: Can just be `id, dylibs` due to Ruby's implicit return.

Library/Homebrew/mach.rb
```diff
@@ -160,7 +162,19 @@ class Metadata
...
       @path = path
-      @dylib_id, @dylibs = parse_otool_L_output
+      @dylib_id, @dylibs = dylib_id_and_dylibs
+    end
+
+    def dylib_id_and_dylibs
+      if ENV["HOMEBREW_RUBY_MACHO"]
+        file = MachO.open(path.expand_path.to_s)
+        id = file.dylib_id
+        dylibs = file.linked_dylibs
```
mikemcquaid: Could just be `file.dylib_id, file.linked_dylibs`

Library/Homebrew/mach.rb
```diff
@@ -160,7 +162,19 @@ class Metadata
...
+      else
+        id, dylibs = parse_otool_L_output
```
mikemcquaid: Could just be `parse_otool_L_output`

Library/Homebrew/mach.rb
mikemcquaid: Actually if you do the two comments above you can kill this line completely.

mikemcquaid:
Might be nice to note somewhere where it's been vendored from and where upstream is?

mikemcquaid:
Looking good so far :+1:

woodruffw:
Noting that it's vendored sounds like a good idea. Any thoughts on where I should make note of that?

mikemcquaid:
Could add a README in the `vendor` directory? Bonus points if you can work out where `okjson` is from and the version.

UniqMartin:
If I had a wish: Can we also get a facility for listing, altering, adding, and deleting of RPATHs (the `LC_RPATH` load commands) once all the other stuff works? Maybe as an optional item on your TO DO list? It's not currently needed, but I suspect it will be very useful.

`install_name_tool` supports these manipulations via `-{,add_,delete_}rpath`. Listing can be done with `otool -l`, though having to scan that output for `LC_RPATH` commands is a bit annoying.

Great work by the way! I'm watching with big interest. :+1:

woodruffw:
@mikemcquaid Sure! Looks like `okjson` is from here, by the way: https://github.com/kr/okjson
I'll document both in the README.

@UniqMartin Thank you! RPATH modification is definitely on the TODO list. They're represented using `lc_str` offsets just like install names and dylib IDs, so the logic for manipulating them should be similar.

woodruffw:
++//////////////////////

mikemcquaid:
@woodruffw Let me know when you consider this ready, thanks!

Library/Homebrew/vendor/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+* okjson is a pure-Ruby JSON parser written by
+[Keith Rarick](https://github.com/kr). Its upstream repository can be found
+[here](https://github.com/kr/okjson).
+
+* ruby-macho is a pure-Ruby Mach-O parser written by
+[William Woodruff](https://github.com/woodruffw). Its upstream repository can
+be found [here](https://github.com/woodruffw/ruby-macho).
+
+## Licenses:
```
mikemcquaid: Could probably just link these to the respective upstream versions.

Library/Homebrew/vendor/README.md
```diff
@@ -0,0 +1,56 @@
+Vendored Dependencies
+=====================
+
+* okjson is a pure-Ruby JSON parser written by
+[Keith Rarick](https://github.com/kr). Its upstream repository can be found
+[here](https://github.com/kr/okjson).
+
+* ruby-macho is a pure-Ruby Mach-O parser written by
+[William Woodruff](https://github.com/woodruffw). Its upstream repository can
+be found [here](https://github.com/woodruffw/ruby-macho).
```
mikemcquaid: I wonder if it's worth noting the revisions we've vendored?

DomT4: :+1: IMO.

woodruffw: So something like this instead?

```diff
diff --git a/Library/Homebrew/vendor/README.md b/Library/Homebrew/vendor/README.md
index 2403e05..cd7beb8 100644
--- a/Library/Homebrew/vendor/README.md
+++ b/Library/Homebrew/vendor/README.md
@@ -1,13 +1,9 @@
 Vendored Dependencies
 =====================
 
-* okjson is a pure-Ruby JSON parser written by
-[Keith Rarick](https://github.com/kr). Its upstream repository can be found
-[here](https://github.com/kr/okjson).
+* [okjson](https://github.com/kr/okjson), version 43.
 
-* ruby-macho is a pure-Ruby Mach-O parser written by
-[William Woodruff](https://github.com/woodruffw). Its upstream repository can
-be found [here](https://github.com/woodruffw/ruby-macho).
+* [ruby-macho](https://github.com/woodruffw/ruby-macho), version 0.0.7.
 
 ## Licenses:
```

DomT4: I'm happy with the existing language if Mike is. My thoughts were just to include a version or even commit reference alongside the current language just to be clear at what point it was vendored from.

mikemcquaid: That looks amazing.

woodruffw: Alright!
By the way, okjson is now on revision 45. It looks like a few numerical parsing issues were fixed in the last few revisions, so it might be worth updating the vendored version at some point.

xu-cheng:
I wonder whether it will be a great idea to transform ruby_macho repo to Homebrew organization? After all, if we go down this road, this library has to be maintained.

Another thing I'm interesting is how the performance differentiate between original binaries and ruby library.

woodruffw:
I intend to maintain `ruby_macho` for the forseeable future. I'm not opposed to putting it under the Homebrew umbrella for the sake of organization and future maintenance, however. 

In terms of performance, I haven't done any substantial benchmarking. `ruby_macho` currently loads the files it operates on into buffers in memory, which is definitely an allocation hog but also makes operating on the actual mach-o data very efficient (everything ends up being an index-bound operation). The ultimate performance will almost definitely be worse than `libmacho`, but it shouldn't be so bad as to cause noticeable slowdowns in Homebrew's execution (considering how much time Homebrew spends forking and executing subprocesses).

woodruffw:
Pinging @xu-cheng and @mikemcquaid:

Are there historical/compatibility reasons for the `mach_data` method in `/os/mac/mach.rb` also expecting ELF headers?

``` ruby
when 0x7f454c46 # ELF
    mach_data << { :arch => :x86_64, :type => :executable }
```

I'm rewriting this method to use `ruby-macho` instead, and I'm wondering if this is something that I can remove safely.

mikemcquaid:
> Are there historical/compatibility reasons for the mach_data method in /os/mac/mach.rb also expecting ELF headers?

I'm not sure, I'm afraid. `git log` may tell you more.

Library/Homebrew/keg_relocate.rb
```diff
@@ -70,12 +72,22 @@ def relocate_text_files(old_prefix, new_prefix, old_cellar, new_cellar)
...
   def change_dylib_id(id, file)
     puts "Changing dylib ID of #{file}\n  from #{file.dylib_id}\n    to #{id}" if ARGV.debug?
-    install_name_tool("-id", id, file)
+
+    if ENV["HOMEBREW_RUBY_MACHO"]
+      MachO::Tools.change_dylib_id(file, id)
+    else
+      install_name_tool("-id", id, file)
+    end
   end
```
mikemcquaid: Was just thinking it might be nice to use `Module`s here and `include` them based on the environment variable.

woodruffw: As in defining `change_dylib_id` separately in two modules and including the correct one at the top of `Keg` and other classes that require Mach-O operations?

woodruffw:
Pinging @sjackman with a question about 076fcd11:
Are your changes to `mach.rb` (now in `os/mac/mach.rb`) still relevant to your work on Linux support?
I'm working on replacing this logic with calls to a new Mach-O parser, and I'd just like to make sure I'm not clobbering anybody else's work/setup with the changes.

woodruffw:
I've changed the logic in *os/mac/pathname.rb* so that the proper implementation (ruby or cctools-based) is `include`d based on the `HOMEBREW_RUBY_MACHO` variable. This is slightly more verbose than what I had originally, but it has the added benefit of modularity.

Should I do something similar for the logic in *keg_relocate.rb*?

mikemcquaid:
> I've changed the logic in os/mac/pathname.rb so that the proper implementation (ruby or cctools-based) is included based on the HOMEBREW_RUBY_MACHO variable. This is slightly more verbose than what I had originally, but it has the added benefit of modularity.

It looks great :+1: 



> Should I do something similar for the logic in keg_relocate.rb?

That's be cool. Additionally it'd be great to maybe put all your new Ruby code in a file by itself?

woodruffw:
Sounds good to me!
Since both `CctoolsMacho` and `RubyMachO` use `ArchitectureListExtension`, how does splitting them into their own files and `require`-ing both into *os/mac/mach.rb* sound?

mikemcquaid:
> Since both CctoolsMacho and RubyMachO use ArchitectureListExtension, how does splitting them into their own files and require-ing both into os/mac/mach.rb sound?

:+1:

sjackman:
@woodruffw Sorry for the delay in my reply, William. I'm on vacation in Thailand until the new year.

The ELF support was added for Linuxbrew. Feel free to remove it if it makes your life easier. That feature can live in the Linuxbrew branch. @danabrand

woodruffw:
@sjackman No worries, thanks for the response!

woodruffw:
@mikemcquaid Now that *mach.rb* is under *os/mac/*, should we consider moving *keg_relocate.rb* as well? It's similarly platform-dependent.

Alternatively, when I break methods like `fix_install_names` and `relocate_install_names` out into their own implementation-specific modules, I could just put those under *os/mac/* and keep *keg_relocate.rb* where it is.

Library/Homebrew/keg_relocate.rb
mikemcquaid: Yup :+1:

mikemcquaid:
> @mikemcquaid Now that mach.rb is under os/mac/, should we consider moving keg_relocate.rb as well? It's similarly platform-dependent.

It does have a bunch of non-platform-dependent bits e.g. the relocating of text files. Technically those aren't relocations though so could go to another class and the mach stuff could go under `os/mac`, yeh.

> Alternatively, when I break methods like fix_install_names and relocate_install_names out into their own implementation-specific modules, I could just put those under os/mac/ and keep keg_relocate.rb where it is.

That works to me as well, feel free to pick :+1:

woodruffw:
@mikemcquaid As of 7d500e2, the ruby-macho replacement calls should be fully functional. From here, I can either get rid of the `cctools` based calls altogether (maybe in another PR?), or reverse the environment check so that the old code is reverted to in the presence of `$HOMEBREW_CCTOOLS` or something similar. 

I can also open another PR to get rid of the cctools-hooking code I added during the GSoC, although it might be worth preserving as part of the fallback system above.

Library/Homebrew/os/mac/cctools_relocate.rb
```diff
@@ -0,0 +1,31 @@
+module CctoolsRelocate
+	def install_name_tool(*args)
```
mikemcquaid: Ident needs tweaked in this file.

mikemcquaid:
Loving the look of this :heart_eyes: 



> @mikemcquaid As of 7d500e2, the ruby-macho replacement calls should be fully functional. From here, I can either get rid of the cctools based calls altogether (maybe in another PR?), or reverse the environment check so that the old code is reverted to in the presence of $HOMEBREW_CCTOOLS or something similar.

What we probably want to do for an incremental rollout is as follows:

1. default to the existing code and check to verify this refactoring hasn't broken anything
2. make CI and Homebrew developers use the new code for a while (a few weeks) to make sure it hasn't broken anything
3. flip the defaults so people can opt-in to the old code for a few weeks
4. delete all the old code and cctools stuff generally.

Sounds good?

UniqMartin:
Sounds very reasonable to me. :+1:

*Just out of curiosity:* How will this PR be merged into `master`? Will the entire commit history be retained and will this be one of those rare case where there will be a true merge commit? Or is this going to be (partially) squashed and rebased?

I have implemented a Mach-O parser in the past (sadly not publicly available) and thus have some experience with this topic. If that is of any value, I could offer a thorough review of the ready-to-be-merged state of this PR before it is actually merged. Sorry I couldn't chime in more regularly to comment on intermediate states of this PR.

mikemcquaid:
> Just out of curiosity: How will this PR be merged into master? Will the entire commit history be retained and will this be one of those rare case where there will be a true merge commit? Or is this going to be (partially) squashed and rebased?

We'll squash it down to one or a few commits before merging.



> I have implemented a Mach-O parser in the past (sadly not publicly available) and thus have some experience with this topic. If that is of any value, I could offer a thorough review of the ready-to-be-merged state of this PR before it is actually merged. Sorry I couldn't chime in more regularly to comment on intermediate states of this PR.

That would be fantastic.

Library/Homebrew/os/mac/ruby_relocate.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def change_install_name(old, new, file)
+    puts "Changing install name in #{file}\n  from #{old}\n    to #{new}" if ARGV.debug?
+    MachO::Tools.change_install_name(file, old, new)
+  end
+
+  # only here until old relocation code is removed
+  def require_install_name_tool?
+    true
```
xu-cheng: This is incorrect. `require_install_name_tool?` is used to properly tag bottle with `:any` or `:any_skip_relocation`. It doesn't concern on which relocation tool is used.

Library/Homebrew/keg_relocate.rb
```diff
@@ -174,16 +162,6 @@ def find_dylib(bad_name)
...
 
-  def mach_o_files
-    mach_o_files = []
-    path.find do |pn|
-      next if pn.symlink? || pn.directory?
-      mach_o_files << pn if pn.dylib? || pn.mach_o_bundle? || pn.mach_o_executable?
-    end
-
-    mach_o_files
-  end
```
xu-cheng: Why relocate this method whose logic doesn't concern with low level mach library.

Library/Homebrew/keg_relocate.rb
```diff
@@ -1,4 +1,12 @@
+require "os/mac/relocate"
```
xu-cheng: If possible, I would like to have `require "os/mac/..."` only exist inside `os` folder to have a better OS abstraction. So I suggest to create an `os/mac/keg.rb` file similar to `os/mac/pathname.rb` file.

xu-cheng:
Loving to have this :+1:. Some general comments:

* We may want to run the code against `brew style` to unify ruby code style.
* Maybe we should maintain the library code inside Homebrew organization(of course keeping @woodruffw as project owner) and sync the code to core using `git subtree`?

mikemcquaid:
> We may want to run the code against brew style to unify ruby code style.

On the non-vendored Homebrew code: I agree.

> Maybe we should maintain the library code inside Homebrew organization(of course keeping @woodruffw as project owner) and sync the code to core using git subtree?

This would be great if @woodruffw was willing. If you'd rather keep this project on your own: that's OK too.

UniqMartin:
> That would be fantastic.

Just to clarify: I can review the Mach-O bits. I will also try to spot potential issues on the Homebrew side, but since I'm still not very familiar with the inner workings and assumptions made, that part will be a lot less useful. And the review probably won't happen this week (too busy with other stuff).

Library/Homebrew/keg_relocate.rb
woodruffw: I had to move this because of some strangeness with how the original `mach_data` worked. Instead of skirting around non-mach-o files, it raises an exception and then later rescues that exception to return an empty array. This messes with the `ruby-macho` calls, as they expect the array to be nonempty.

This could be fixed by guarding the usage of methods like `Pathname#dylib?` with a file sanity check, but until then I just decided to preserve the behavior that the original `mach_data` produced.

Library/Homebrew/keg_relocate.rb
woodruffw: Right now the pertinent code is in `os/mac/{ruby_,cctools,,}relocate`, does renaming all of these to "keg" instead of "relocate" and correcting the `require`s sound good to you?

Library/Homebrew/keg_relocate.rb
xu-cheng: > This could be fixed by guarding the usage of methods like Pathname#dylib? with a file sanity check

Let's do this. It's strange to let a non dylib file raise exception instead of return false when calling `dylib?`.

Library/Homebrew/keg_relocate.rb
xu-cheng: Yeah, sounds good to rename `os/mac/relocate.rb` to `os/mac/keg.rb`. The whole idea is trying to keep os related codes inside `os` folder.

woodruffw:
@UniqMartin
I would greatly appreciate your review of `ruby-macho` - just tell me anything I can do to help!

> Maybe we should maintain the library code inside Homebrew organization(of course keeping
> @woodruffw as project owner) and sync the code to core using git subtree?

I'm open to moving `ruby-macho` into the Homebrew organization - it would make bringing new features/updates to the parser a much simpler process :smile: 

> What we probably want to do for an incremental rollout is as follows:

> 1. default to the existing code and check to verify this refactoring hasn't broken anything
> 2. make CI and Homebrew developers use the new code for a while (a few weeks) to make sure it hasn't broken anything
> 3. flip the defaults so people can opt-in to the old code for a few weeks
> 4. delete all the old code and cctools stuff generally.

This sounds great to me!

Library/Homebrew/os/mac/ruby_relocate.rb
woodruffw: Yeah, I misunderstood the semantics of `require_install_name_tool?`. I'll fix this in a moment.

xu-cheng: Yeah, we may need a better name, such `Keg#require_relocation?`

woodruffw: 47376b4 contains the preliminary change on this. I think `Keg#require_relocation?` is a much better name, so I can change both files to use that if you'd like that as well.

Library/Homebrew/keg_relocate.rb
woodruffw: Is `require` smart enough/properly loaded to search `os/mac` before the top source directory? I'm about to rename these, but I think I'll still have to explicitly `require "os/mac/relocate"` at the top of `keg_relocate.rb`

xu-cheng: You need to put require statement inside `os/mac.rb`

woodruffw: That makes sense - thank you!

Library/Homebrew/keg_relocate.rb
woodruffw: ce315ae should take care of this!

mikemcquaid:
@woodruffw This is looking pretty good but it's hard to review. Would there be any chance of squashing down the commits into: standalone core changes, `ruby-macho` import, porting to use `ruby-macho` optionally. CC @UniqMartin for thoughts on the approach.

woodruffw:
Certainly! I'll squash down right now.

Library/Homebrew/os/mac/pathname.rb
```diff
@@ -1,5 +1,9 @@
 require "os/mac/mach"
 
 class Pathname
-  include MachO
+  if ENV["HOMEBREW_RUBY_MACHO"]
+    include RubyMachO
+  else
+    include CctoolsMachO
+  end
```
xu-cheng: One last nit: I think it's better to put `require "os/mac/ruby_mach"` and `require "os/mac/cctools_mach"` here inside the condition block and remove `Library/Homebrew/os/mac/mach.rb` file. The reason is we should only load one of the libraries instead of both.

woodruffw: That makes a lot of sense, fixed!
I also made `os/mac/keg` `require` its library conditionally in a similar fashion.

mikemcquaid:
@woodruffw Sorry for pedantry but I'd love to see this squashed to three commits: first standalone core changes that don't need `ruby-macho`, next just the import of ruby-macho without any hooking up, finally all the porting/changes to use ruby-macho optionally.

UniqMartin:
Splitting the commits as suggested by @mikemcquaid makes sense to me, too. Sorry for the long time without any activity here, @woodruffw. I guess I still owe you the review that I promised to do …

woodruffw:
@mikemcquaid I don't think I made any core changes that don't involve `ruby-macho`, so I could squash it all the way down to two - is that okay?

@UniqMartin No worries, I've also been inactive here for a while.

mikemcquaid:
@woodruffw Sure. I guess I mean: if there's any core changes that you _could_ make without `ruby-macho` having them as a first commit would be good.

woodruffw:
@mikemcquaid I suppose I could split the removal of the ELF header case out into its own commit, but it is partially related to the split of *mach.rb* into *cctools_mach.rb* and *ruby_mach.rb*.
Thoughts?

mikemcquaid:
@woodruffw Yep, that would be ideal; basically all the renames/splitting/moving in a commit before the MachO stuff is imported.

woodruffw:
That should do it.
4b95994 is the ELF pruning.
765eb38 is branching of all the relocation code and inclusion of `ruby-macho` components.
d2871bc is the actual `ruby-macho` code dump.

mikemcquaid:
@woodruffw One nit on the https://github.com/Homebrew/homebrew/commit/4b95994e989c886b04c7098f0447ed1295888333 and can you tweak the ordering so that https://github.com/Homebrew/homebrew/commit/d2871bc61480ba351476936a529f9f8749818e0f comes before https://github.com/Homebrew/homebrew/commit/765eb38733749ebebe21fb3eb05a015613a305ac (so tests on any individual commit would still pass). After that: :+1: from me.

woodruffw:
That should do it :tada:

mikemcquaid:
Thanks @woodruffw! 

@Homebrew/maintainers any last thoughts before I merge this? After merging if there's no regressions for a few days I'll enable the variable on the bots (and for myself) for testing, then `HOMEBREW_DEVELOPER`s and then unconditionally.

bfontaine:
It’d be great to have a few unit tests for the stuff outside of `vendor/`.

woodruffw:
@bfontaine Do you have any particular tests in mind? When `$HOMBREW_RUBY_MACHO` is set, the current mach-o/relocation tests should automatically switch it testing it instead.
I can make that manual if you'd like.

bfontaine:
Oh right, so it’ll be tested when we’ll set it for the bot :+1:

woodruffw:
Yep!

woodruffw:
@UniqMartin I've brought in your changes as 0.2.2, thank you!

apjanke:
Nice! I'm no Mach-O expert, but I'm comfortable with the work history here.

Minor procedural nit: Is the license for this one of the well-known open source licenses? The README contains the text of the license, but I didn't see a license name anywhere. Would be nice to include the name if it's a standard one; I don't know the licenses well enough to recognize them from the license text.

woodruffw:
@apjanke Oops, that's a mistake. It's the MIT license.

I'll fix it right now!

apjanke:
Cool. :+1: to merge by me.

Library/Homebrew/os/mac/ruby_mach.rb
```diff
@@ -0,0 +1,101 @@
...
+
+      if MachO.fat_magic?(macho.magic)
+        machos = macho.machos
+      else
+        machos << macho
+      end
+
+      machos.each do |m|
+        arch = case m.cputype
+        when "CPU_TYPE_X86" then :i386
```
UniqMartin: This never matches because the string returned by `cputype` is `"CPU_TYPE_I386"` in this case, effectively breaking some functionality for IA32 (x86, whatever one likes to call it). See also woodruffw/ruby-macho#5.

Library/Homebrew/os/mac/ruby_mach.rb
```diff
@@ -0,0 +1,101 @@
...
+        when "CPU_TYPE_X86_64" then :x86_64
+        when "CPU_TYPE_POWERPC" then :ppc7400
+        when "CPU_TYPE_POWERPC64" then :ppc64
+        end
```
UniqMartin: I would suggest to insert a line `else :dunno` before this one (like in `cctools_mach.rb`) to avoid returning `nil` instead of a symbol, as that will make breakage due to new/unknown architectures easier to spot.

Library/Homebrew/os/mac/ruby_mach.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        type = case m.filetype
+        when "MH_EXECUTE" then :executable
+        when "MH_DYLIB" then :dylib
+        when "MH_BUNDLE" then :bundle
+        end
```
UniqMartin: I would suggest to insert a line `else :dunno` before this one (like in `cctools_mach.rb`) to avoid returning `nil` instead of a symbol, as that will make breakage due to an unrecognized/unexpected Mach-O file type easier to spot.

Library/Homebrew/os/mac/ruby_mach.rb
woodruffw: Yes, silly mistake on my part. Fixing it now.

Library/Homebrew/os/mac/ruby_mach.rb
woodruffw: Good idea, adding it.

Library/Homebrew/os/mac/ruby_mach.rb
woodruffw: Ditto here.

mikemcquaid:
Did some local testing and looked good with and without `HOMEBREW_RUBY_MACHO` for installing and bottling `wget`. Did some commit message rewriting (http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html is a good guide) and pushed. I'll enable `HOMEBREW_RUBY_MACHO` on the bots if there's no issues in about a week, enable for all `HOMEBREW_DEVELOPER`s after that and everyone after that. Nice work @woodruffw!

UniqMartin:
Indeed, great work here! I only had limited time for testing (or I'm just being slow :blush:), but as far as I can tell the support for x86 and x86_64 is fairly robust. My testing focused on reading stuff, so can't say anything yet about the parts that modify Mach-O files.

Not sure how much we care about legacy architectures (Misty certainly does). Support for PowerPC is theoretically there, but it doesn't work in practice. I have to conduct a bit more testing, but I'm confident I can submit something for discussion to the upstream repository very soon and given how quickly @woodruffw has picked up my other fixes/suggestions, this will surely be resolved soon.

*If anyone is eager to test:* Try building more complex stuff like `gcc` and `qt5` and a few formulae with universal binaries (x86/x86_64 only), preferably keep around the previous build and compare results with `otool -L` etc. (It's also a good idea to build with `-d` to see what should have been rewritten.)

woodruffw:
@mikemcquaid Awesome!
(Thanks for the commit message link - I'll format future commits more regularly).

By the way, are we still entertaining the idea of bringing `ruby-macho` under the Homebrew project umbrella?

woodruffw:
@UniqMartin Thank you!
Your contributions to `ruby-macho` were (and are) greatly appreciated.

I put PPC support on the back burner when I began writing it, so there are a few assumptions baked in that need to be changed (mostly assumptions about endianness and `FORMAT` constants in load commands).

In the mean time, I'll be running tests as well and expanding the library's test suite.

mikemcquaid:
> By the way, are we still entertaining the idea of bringing ruby-macho under the Homebrew project umbrella?

@woodruffw if you're up for that: yes! I can add you to the organisation and let you transfer it here, if you like?

woodruffw:
@mikemcquaid Sounds good to me!

mikemcquaid:
@woodruffw You've been invited. You should be able to move the repo over and then I'll make you an admin.

woodruffw:
Hmm, It looks like I need admin rights to move a repository over.

mikemcquaid:
@woodruffw You should be able to move it to the org as-is? If not, I'll sort it tomorrow, let me know.

woodruffw:
@mikemcquaid I get a little pop-up:



I think we could also execute the transfer by transferring the repo to you first (or any other admin), then transferring it to Homebrew.

mikemcquaid:
@woodruffw Because I'm GitHub staff I just moved it manually for you and made you an admin.

woodruffw:
Awesome. One of the perks of the job :smile:

PR #45002, opened 2015-10-15
new formula: sqliteodbc 0.9992

Habbie:


Library/Formula/sqliteodbc.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  depends_on "sqlite"
+  depends_on "unixodbc"
+
+  def install
+    lib.mkdir
+    system "./configure", "--prefix=#{prefix}"
+    system "make"
+    system "make", "install"
+    File.rename "#{lib}/libsqlite3odbc-#{version}.dylib", "#{lib}/libsqlite3odbc-#{version}.so"
```
DomT4: OS X uses `dylib`s rather than shared object files. Why the change here?

Habbie: I'm following the lead of psqlodbc, mysql-connector-odbc, freetds here, that all also use .so. In general I get the impression that libraries that are linked on startup get .dylib, but 'plugins' get .so. If you disagree, I'm happy to remove that change of course, the library works fine with either extension (I tested).

mikemcquaid: I don't think it's worth renaming unless there's actual an issue. If you think it should be named differently it might be worth talking to upstream.

Habbie: I have contacted upstream, will see what they say. How would you feel about a symlink instead of a rename, so ODBC users at least have consistency in their config - and also things won't break if upstream fixes it?

Habbie:
I have pushed a second commit that replaces the rename with a single symlink; please let me know what you think and I'll squash the right version back into one commit. Upstream has not gotten back on the `.so`/`.dylib` issue yet.

Habbie:
mavericks and yosemite failed with ` * The homepage is not reachable (curl exit code 56)` - can you please retest?

mikemcquaid:
@BrewTestBot test this please

Library/Formula/sqliteodbc.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  depends_on "sqlite"
+  depends_on "unixodbc"
+
+  def install
+    lib.mkdir
+    system "./configure", "--prefix=#{prefix}"
+    system "make"
+    system "make", "install"
+    File.symlink "#{lib}/libsqlite3odbc.dylib", "#{lib}/libsqlite3odbc.so"
```
DomT4: Can we use `lib.install_symlink "#{lib}/libsqlite3odbc.dylib" => "libsqlite3odbc.so"` rather than throwing out to `File` directly? Thanks.

Library/Formula/sqliteodbc.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  test do
+    output = `#{Formula["unixodbc"].bin}/dltest #{lib}/libsqlite3odbc.so`
```
DomT4: Can we use `shell_output` or `pipe_output` here? That would check exit status as well without the addition of the third line.

Library/Formula/sqliteodbc.rb
Habbie: doing this

Library/Formula/sqliteodbc.rb
Habbie: of course - copied this code from the psqlodbc brew without looking too closely. Want me to fix psqlodbc in a separate PR?

DomT4: If it works there as well, sure, feel free!

Habbie:
`lib.install_symlink` and `shell_output` pushed!

Habbie:
Not to sound impatient, just making sure: is there anything I can/should do to move this PR along?

DomT4:
Will take a look at this sometime over the next couple hours. Just trying to resolve some El Cap build issues with other formulae at the moment.

Habbie:
Definitely more important, good luck!

On October 23, 2015 5:20:12 PM GMT+02:00, Dominyk Tiller <notifications@github.com> wrote:
>Will take a look at this sometime over the next couple hours. Just
>trying to resolve some El Cap build issues with other formulae at the
>moment.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/Homebrew/homebrew/pull/45002#issuecomment-150605814

-- 
Sent from my Android phone with K-9 Mail. Please excuse my brevity.

DomT4:
Apologies for the delay, headache got the better of me.

Merged in https://github.com/Homebrew/homebrew/commit/0ee7593cdd11a24686f3d4f5ab23bf73761543a5. Thank you for another contribution to Homebrew @Habbie! :smiley:

Habbie:
Please don't apologize, we are all volunteers here. Thank you for merging!

PR #44293, opened 2015-09-23
Add sift formula

mwean:
Grep-like tool https://sift-tool.org/

PR #45006, opened 2015-10-15
Add sift formula

mwean:
Continuing from #44293


mikemcquaid:

For future reference the preferred commit message format for new formulae is `sift 0.3.4 (new formula)`.

PR #45019, opened 2015-10-16
downloads: Fix checkout on initial clone

### LINKED ISSUE #44828: go_resource statements do not check out the specified revision to $GOPATH directory
I have a formula that has `go_resource` statements with URLs pointing to github.com repos with specific revisions.

    go_resource "labix.org/v2/mgo" do
      url "https://github.com/go-mgo/mgo.git",
        :revision => "2e26580ebcca4eca6533535f2dd062cd6bff44ac"
    end

When `Language::Go.stage_deps resources, buildpath/"src"` command stages the repos into `$GOPATH` folder, the repos are checked out to the tip of the tree instead of revision specified in the formula. This breaks the building of Go binaries.

Try running `brew install cayley -v` and you will see the breakage as the formula points to an older revision of `https://github.com/go-mgo/mgo.git` but it's getting the latest commit instead.

My setup: MacOS 10.11 with homebrewed `go-1.5.1`.

### LINKED ISSUE #44981: problem with :revision keyword
Put [this formula] (https://raw.githubusercontent.com/lemzwerg/homebrew/29196533fa50af4a90e875c8ca554e1638127639/Library/Formula/ttfautohint-1.3.56.rb) into `Library/Formula`, which is a slight variation of the already present `ttfautohint.rb` file.  Then say

    brew install ttfautohint-1.3.56

`ttfautohint --version` should contain 1.3.56.  However, instead of picking up the selected git revision as specified with the `url` keyword, homebrew uses git HEAD instead.  If I follow the above command with

      brew reinstall ttfautohint-1.3.56

I finally get the right version.

In other words, it seems that Homebrew doesn't handle the `:revision` keyword correctly if the git repository to be accessed is not yet cloned.

apjanke:
Also moves the "Checking out" output to where checkouts actually happen, to avoid spurious checkout announcements.

I've tested this with some formulae that use `:revision` keywords covering both `:git` and `:hg` downloads: `cayley`, `gost`, `packer`, `colortail`.

Fixes #44981
Fixes #44828

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -135,10 +135,6 @@ class VCSDownloadStrategy < AbstractDownloadStrategy
     end
   end
 
-  def stage
-    ohai "Checking out #{@ref_type} #{@ref}" if @ref_type && @ref
-  end
-
   def cached_location
     @clone
   end
@@ -503,7 +499,10 @@ class SubversionDownloadStrategy < VCSDownloadStrategy
     args = ["svn", svncommand]
     args << url unless target.directory?
     args << target
-    args << "-r" << revision if revision
+    if revision
+      ohai "Checking out #{ref}"
+      args << "-r" << revision
+    end
     args << "--ignore-externals" if ignore_externals
     quiet_safe_system(*args)
   end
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Great work here @apjanke. I'm :+1: when CI passes.

apjanke:
Thanks!

mikemcquaid:
Thanks @apjanke!

PR #45021, opened 2015-10-16
OS related code cleanup

xu-cheng:
cc @bfontaine @mikemcquaid @woodruffw


Library/Homebrew/os/mac/mach.rb
```diff
@@ -202,3 +199,7 @@ def dylib_id
     mach_metadata.dylib_id
   end
 end
+
+class Pathname
+  include(MachO)
+end
```
mikemcquaid: I think it's better this lives in `Library/Homebrew/extend/pathname.rb` or `Library/Homebrew/extend/pathname` somewhere.

xu-cheng: Actually, I move this from `Library/Homebrew/extend/pathname.rb`. What I'm thinking is to put OS related code inside `os` folder.

mikemcquaid: Yeh, I prefer all monkeypatching happens in `Library/Homebrew/extend/` somewhere.

xu-cheng: How about `Library/Homebrew/os/mac/extend/` or something?

If we still like to put it in original place, may be as `include(MachO) if OS.mac?`?

If you don't like either of these, I'm happy to keep original form. (Just make sure on the same page for style issue)

mikemcquaid: How about `Library/Homebrew/extend/os/mac/pathname`?

xu-cheng: :+1:  will update PR shortly.

xu-cheng: Updated.

Library/Homebrew/os/mac/mach.rb
```diff
@@ -164,11 +164,10 @@ def initialize(path)
     end
 
     def parse_otool_L_output
-      ENV["HOMEBREW_MACH_O_FILE"] = path.expand_path.to_s
```
mikemcquaid: I'm hesitant to remove this environment variable unless you're 100% sure there's no reason it was being used before.

mikemcquaid:
One comment otherwise :+1:

Library/Homebrew/os/mac/mach.rb
xu-cheng: I'm 100% sure this is used to escape path. See https://github.com/Homebrew/homebrew/blob/8f9854f81f8064c1428a05efc8d9754e92e259a7/Library/Homebrew/mach.rb#L153

mikemcquaid: Ok. :+1:

PR #45023, opened 2015-10-16
gist-logs: Friendlier titles and display for log gists

apjanke:
Here's a suggestion for making the gists produced by `gist-log` easier to work with.

Adds a summary file and description to get more informative displays on gist.github.com. Adds `--with-hostname` to include host name in summary, for differentiating different machines if you have several.

The summary file starts with "#" so it sorts asciibetically before other files in the gist, and is thus taken by the gist webpage as the display title to use for the gist, and the summary text displayed for it when it's in a list of multiple gists.

Adds a description to the gist, which is used as the title of the page when the gist is viewed on gist.github.com. This may also come in handy if some new page displays a denser list of gists that uses just the text descriptions, instead of an excerpt from the first file.

This makes it easier to identify Homebrew log gists when working with more than one. For example, when you're displaying the list of "my gists", or have several bug reports open in browser windows at once. As it stands now, the effective title of the gists is always something like `01.configure` or `01.cmake`, and not the name of the formula in question. The name of the formula may not even appear in the first several lines of that file, so when looking at the gist preview, you can't necessarily tell what it is.

Old:

screen shot 2015-10-16 at 5 49 59 am

With this change, the formula name is in the gist's "title" and the summary is always in the preview text when listing multiple gists. And it shows up in the tab title too, once you open them.

screen shot 2015-10-16 at 5 51 08 am

This PR adds `gist-logs` option `--with-hostname` which includes the host name in that new summary, which is useful if, like me, you run Homebrew on several different machines with different configurations. I left it off by default, because it has some mild potential privacy implications, so it should probably be opt-in.

### DIFF
--- a/Library/Homebrew/cmd/gist-logs.rb
+++ b/Library/Homebrew/cmd/gist-logs.rb
@@ -3,25 +3,36 @@ require "cmd/config"
 require "net/http"
 require "net/https"
 require "stringio"
+require "socket"
 
 module Homebrew
   def gistify_logs(f)
     files = load_logs(f.logs)
+    build_time = f.logs.ctime
+    timestamp = build_time.strftime("%Y-%m-%d_%H-%M-%S")
 
     s = StringIO.new
     Homebrew.dump_verbose_config(s)
-    files["config.out"] = { :content => s.string }
-    files["doctor.out"] = { :content => `brew doctor 2>&1` }
+    # Dummy summary file, asciibetically first, to control display title of gist
+    files["# #{f.name} - #{timestamp}.txt"] = { :content => brief_build_info(f) }
+    files["00.config.out"] = { :content => s.string }
+    files["00.doctor.out"] = { :content => `brew doctor 2>&1` }
     unless f.core_formula?
       tap = <<-EOS.undent
         Formula: #{f.name}
         Tap: #{f.tap}
         Path: #{f.path}
       EOS
-      files["tap.out"] = { :content => tap }
+      files["00.tap.out"] = { :content => tap }
     end
 
-    url = create_gist(files)
+    # Description formatted to work well as page title when viewing gist
+    if f.core_formula?
+      descr = "#{f.name} on #{OS_VERSION} - Homebrew build logs"
+    else
+      descr = "#{f.name} (#{f.full_name}) on #{OS_VERSION} - Homebrew build logs"
+    end
+    url = create_gist(files, descr)
 
     if ARGV.include?("--new-issue") || ARGV.switch?("n")
       auth = :AUTH_TOKEN
@@ -38,6 +49,19 @@ module Homebrew
     end
 
     puts url if url
+  end
+
+  def brief_build_info(f)
+    build_time_str = f.logs.ctime.strftime("%Y-%m-%d %H:%M:%S")
+    s = <<-EOS.undent
+      Homebrew build logs for #{f.full_name} on #{OS_VERSION}
+    EOS
+    if ARGV.include?("--with-hostname")
+      hostname = Socket.gethostname
+      s << "Host: #{hostname}\n"
+    end
+    s << "Build date: #{build_time_str}\n"
+    s
   end
 
   # Hack for ruby < 1.9.3
@@ -68,8 +92,8 @@ module Homebrew
     logs
   end
 
-  def create_gist(files)
-    post("/gists", "public" => true, "files" => files)["html_url"]
+  def create_gist(files, descr)
+    post("/gists", { "public" => true, "files" => files, "description" => descr })["html_url"]
   end
 
   def new_issue(repo, title, body, auth)


apjanke:
Don't know why this test failed, and on 10.9 only. It passes on my 10.9 box with this PR's changes.

```
VersionParsingTests#test_pathname_version [/usr/local/Library/Homebrew/test/test_versions.rb:88]:

--- expected
+++ actual
@@ -1,6 +1,3 @@
 #<Version:0xXXXXXX
- @tokens=
-  [#<Version::NumericToken 0>,
-   #<Version::NumericToken 1>,
-   #<Version::NumericToken 9>],
- @version="0.1.9">
+ @tokens=[#<Version::StringToken "r4">, #<Version::AlphaToken "a">],
+ @version="r4a">


```

Library/Homebrew/cmd/gist-logs.rb
```diff
@@ -21,7 +26,9 @@ def gistify_logs(f)
       files["tap.out"] = { :content => tap }
     end
 
-    url = create_gist(files)
+    # Description formatted to work well as page title when viewing gist
+    descr = "#{f.name} on #{MACOS_FULL_VERSION} - Homebrew build failure "
```
bfontaine: We should probably use `build logs` instead of `build failure` because `brew gist-logs` can be called on a successful build.

mikemcquaid:
Nice, I really like this approach :+1:. It'd also be really handy to have `doctor` and `config` output at the top rather than the bottom.

Library/Homebrew/cmd/gist-logs.rb
DomT4: :+1:, Agree with Baptiste. Use `gist-logs` quite a bit to compare my successful build against someone's unsuccessful build as it's easier than opening 7 files side-by-side in a text editor heh.

DomT4:
Works nicely with Ruby 1.8. This approach is great, a significant improvement. Agree with `hostname` being left as an opt-in thing.

apjanke:
@mikemcquaid: Thanks! Easy peasy: put "00." on `config` and `doctor`, and they sort to the top after the summary and before the main build logs. https://gist.github.com/apjanke/19e0fcb1ec43d5f872fc

Added that, along with removing the possibly-spurious "failure" mentions.

Library/Homebrew/cmd/gist-logs.rb
apjanke: I just woke up with that bugging me too. Wording changed.

luzpaz:
:+1: This is a great feature and will help me navigate my gist archive so much better! Thanks @apjanke

Library/Homebrew/cmd/gist-logs.rb
```diff
@@ -21,7 +26,9 @@ def gistify_logs(f)
       files["tap.out"] = { :content => tap }
     end
 
-    url = create_gist(files)
+    # Description formatted to work well as page title when viewing gist
+    descr = "#{f.name} on #{MACOS_FULL_VERSION} - Homebrew build logs"
```
xu-cheng: I suggest:
* `f.full_name` instead of `f.name`.
* `OS_VERSION` instead of `MACOS_FULL_VERSION`.

Library/Homebrew/cmd/gist-logs.rb
```diff
@@ -40,6 +47,19 @@ def gistify_logs(f)
     puts url if url
   end
 
+  def brief_build_info(f)
+    build_time_str = f.logs.ctime.strftime("%Y-%m-%d %H:%M:%S")
+    hostname = Socket.gethostname
```
xu-cheng: Maybe put this line inside if branch below. Also, I wonder do we really need an extra option?

Library/Homebrew/cmd/gist-logs.rb
```diff
@@ -40,6 +47,19 @@ def gistify_logs(f)
...
+    s = <<-EOS.undent
+      Homebrew build logs for #{f.name} on #{MACOS_FULL_VERSION}
```
xu-cheng: Similar as above:
* `f.full_name` instead of `f.name`.
* `OS_VERSION` instead of `MACOS_FULL_VERSION`.

Library/Homebrew/cmd/gist-logs.rb
```diff
@@ -21,7 +26,9 @@ def gistify_logs(f)
       files["tap.out"] = { :content => tap }
```
xu-cheng: Maybe want to rename it as `00.tap.out` as well.

Library/Homebrew/cmd/gist-logs.rb
apjanke: Agreed on these and the `00.tap.out`. Amending to do so.

Library/Homebrew/cmd/gist-logs.rb
apjanke: Moving it inside the `if`. It was just leftover here from when I originally made it unconditional.

I think we really ought to have an option or environment variable here. It's not huge, but exposing the hostname is potentially personally identifying information, and for bug reports which may have security implications, an attack surface. These gists are public. And `gist-logs` has been around for a while and it hasn't exposed info like this in the past. So I think we should make it explicitly opt-on.

Conversely, it is not a big deal to leave it out IMHO: including the host name is only useful for users who run Homebrew on several different machines and need to distinguish between them in their bug reports. (e.g. I have hosts like `foo-yos`, `foo-elcap`, `foo-no-CLI` for different configurations that are quickly identifiable by name.) That sounds like an advanced use case.

apjanke:
I've amended this with xu-cheng's suggestions for better tap-related info. Non-core formulae will display their tap-qualified names, and `tap.out` is moved up top with `00.config.out` and `00.doctor.out` as `00.tap.out`.

Library/Homebrew/cmd/gist-logs.rb
DomT4: >  And gist-logs has been around for a while and it hasn't exposed info like this in the past. So I think we should make it explicitly opt-on.

Agree on this FWIW.

apjanke:
Okay, I fixed up the description/tab title for non-core formulae, and it's looking good now IMHO.

That's everything I had yet to do on this PR. I think it's ready to merge if the CI tests pass.

screen shot 2015-11-26 at 10 15 29 am - annotated

mikemcquaid:
:+1: to merge, would love to have this years ago :grinning:

apjanke:
Away it goes! Merged.

Issue #44998, opened 2015-10-15
qt5: bug fixes and formula improvements (discussion)


UniqMartin:
Qt 5.5.1 has landed, the bump commit is ready, and the local test build is currently running. Fortunately, we will be able to drop most of the patches needed for 5.5.0.

The reason this is not yet a PR is that I'd like to use the opportunity to “sneak in” a few improvements to the formula that are unrelated to the version bump, without burdening the test bot. Please have a look at <https://github.com/UniqMartin/homebrew/compare/master...qt5-cleanup>. Every commit corresponds to one isolated change and the motivation for that change is given in the commit message. I'd appreciate any feedback.

I'd like to create a PR for the version bump (including the improvements) once we have agreed on a subset of the above commits. Is this approach okay with everyone?

CC @DomT4 as you seem to be the primary Homebrew maintainer dealing with this formula.

mikemcquaid:
Looks great and really appreciate the way you've approached this. Made one comment.

UniqMartin:
Thanks for the quick feedback! So far, my minimal local build has completed successfully. If that's okay, I'll wait with the PR until tomorrow. By then, my full local build will be finished (hopefully without errors) and others will have had a chance to comment here, too.

DomT4:
> CC @DomT4 as you seem to be the primary Homebrew maintainer dealing with this formula.

I'm happy not to be though :smile_cat: :trollface:.

I'd almost be tempted to turn WebKit off by default as that's a *ridiculous* chunk of build time and something most people likely don't use, but probably too radical a step to make suddenly. It'd be nice to check if they've resolved any of the outstanding `rpath` problems upstream and whether we can cut that option but I don't have high hopes.

All the proposed changes look fine to me. If you ever feel a particularly burning passion for Qt it'd be useful to see what we can transition across from using Qt to Qt5, as the former is basically dead upstream now. Appreciate you working on the Qt5 formula & your contributions to Homebrew lately more broadly.

mikemcquaid:
> I'd almost be tempted to turn WebKit off by default as that's a ridiculous chunk of build time and something most people likely don't use, but probably too radical a step to make suddenly. It'd be nice to check if they've resolved any of the outstanding rpath problems upstream and whether we can cut that option but I don't have high hopes.

Nah, people definitely use it.

noplay:
I agree people use webkit and Qt team push is usage more and more.

[...]

PR #45026, opened 2015-10-16
qt5 5.5.1

UniqMartin:
Following the discussion in #44998, here's the version bump and a few improvements to the formula itself.

### DIFF
--- a/Library/Formula/qt5.rb
+++ b/Library/Formula/qt5.rb
@@ -117,6 +117,8 @@ class Qt5 < Formula
       args << "-L#{dbus_opt}/lib"
       args << "-ldbus-1"
       args << "-dbus-linked"
+    else
+      args << "-no-dbus"
     end
 
     if build.with? "oci"
--- a/Library/Formula/qt5.rb
+++ b/Library/Formula/qt5.rb
@@ -15,48 +15,17 @@ class Qt5 < Formula
   desc "Version 5 of the Qt framework"
   homepage "https://www.qt.io/"
   head "https://code.qt.io/qt/qt5.git", :branch => "5.5", :shallow => false
-  revision 1
 
   stable do
-    url "https://download.qt.io/official_releases/qt/5.5/5.5.0/single/qt-everywhere-opensource-src-5.5.0.tar.xz"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Nice :+1:

luzpaz:
brew bot reports: 
```ERROR: Publisher 'Publish JUnit test result report' failed: No test report files were found. Configuration error?```

DomT4:
The failure is slightly above that line; It hit the timeout of 70 minutes or so. We'll reschedule with a longer timeout when the CI isn't quite so rammed.

DomT4:
@BrewTestBot test this please

UniqMartin:
The build succeeded. :sunglasses: 

However, if we can wait committing this for a day or two, I might have an elegant fix that also solves the `brew install qt5 --with-docs` issue (confirmed) and the broken `macdeployqt` (unconfirmed, have yet to do some testing). If that turns out to be the case, we could spare people a revision bump with the ensuing rebuild or bottle download.

mikemcquaid:
Decided to pull as-is because when I was a Qt developer I wouldn't have wanted to wait :grinning:. Looking forward to the follow-up PR(s) though. Nice work @UniqMartin and thanks!

DomT4:
@UniqMartin Both the Qt and Qt5 formulae install the same `.app` bundles with the same names and overwrite each other on `brew linkapps`. It might be an idea to tweak the `qt5` in-formula mechanism so those app bundles are appended with a `5` and don't overwrite the `qt` formula's apps, similar to how we do for [Python3](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/python3.rb#L168-L169).

Would file a PR myself but you seem to be working on the two remaining `qt5` issues and it's nice to spare the CI multiple Qt5 builds very quickly, heh. Let me know if you have no immediate plans and I'll push it though :smiley:

UniqMartin:
Will keep that in mind. If you've already made that change locally, push it to your clone and I'll cherry-pick it into the branch that will eventually become my next Qt 5 PR.

DomT4:
Haven't pushed it yet. You're more than welcome to include it in one of your PRs; not fussed about whether or not I get commit credit for the thought :)

UniqMartin:
Alright. :)

PR #45030, opened 2015-10-16
utils: improve highlighting in multi-column output

### LINKED ISSUE #45028: `brew update` color output is unreadable
I know this sounds very bikeshed-y, but a recent update (it might have been #44335) colorizes the output of `brew update` with highlighting of installed formulas. In my terminal (and perhaps others using a dark background, unlike the white background used in the PR) this displays as white text on a yellow background, which is unreadable:

screen shot 2015-10-16 at 8 51 07 am

May I suggest adding an option along the lines of HOMEBREW_NO_EMOJI to disable this?

UniqMartin:
The yellow background made the text unreadable for light-on-dark color schemes. ~~Fix by explicitly setting the foreground color to black.~~ Fix by making the highlighted text just bold.

Fixes #45028.

This fixes the issue of poor contrast for all users with a sane terminal color scheme (tested locally with quite a bunch). It doesn't solve the problem that everyone has a different favorite color. Text on colored background instead of bolded/colored text on default background was chosen for greater visibility.

Let the bikeshedding begin. :wink:

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -43,7 +43,7 @@ class Tty
     end
 
     def highlight
-      bold 43
+      bold 39
     end
 
     def width


mikemcquaid:
I think just doing bold text would be nice, personally.

UniqMartin:
I'm not happy with plain bold for the stated reason, but ultimately it's a question of preference. However, unless someone else suddenly chimes in, it'll be just bold & done (just pushed again).

mikemcquaid:
Have pushed as-is for now. If other maintainers disagree: please bikeshed here :grinning:.

dunn:
My bike is shedding
screenshot 2015-10-16 18 01 06
(Solarized Light on iTerm in tmux)

White on blue looks a bit better, on Solarized dark/light plus the builtin iTerm and Terminal.app color schemes:
```diff
diff --git a/Library/Homebrew/utils.rb b/Library/Homebrew/utils.rb
index 528bedd..329d73a 100644
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -43,7 +43,7 @@ class Tty
     end

     def highlight
-      bold 43
+      bold "37;44"
     end

     def width
```

mikemcquaid:
There's nowhere else in Homebrew that sets coloured backgrounds, as far as I can tell, right?

UniqMartin:
> There's nowhere else in Homebrew that sets coloured backgrounds, as far as I can tell, right?

Should be true unless some code is generating escape codes manually, instead of relying on the `Tty` class. All styles provided by `Tty` leave the background color alone.

dunn:
Oh, somehow I never saw the result of 52c546361a0d80df09c72cbc572e93a22af68f75 and thought the yellow background was still being used.  Just bold is fine with me.

PR #45037, opened 2015-10-16
wine 1.7.53 (devel)

uyjulian:


uyjulian:
Seems like I should merge first, then commit, not the other way around.

Library/Formula/wine.rb
```diff
@@ -5,7 +5,6 @@
 class Wine < Formula
   desc "Wine Is Not an Emulator"
   homepage "https://www.winehq.org/"
-  revision 1
```
DomT4: This line shouldn't get removed on devel version bumps. Only stable ones.

uyjulian: you sure? 02eefc7c2443a1bd45a71619eb21e882b44ed68d

dholm: We only added it because the patch was removed when the devel version was bumped last time.

DomT4: Nonetheless it applies to both stable and devel versions, so the version for stable is:
```
https://homebrew.bintray.com/bottles/wine-1.6.2_1.el_capitan.bottle.tar.gz
```
If we remove the revision bump and then have to add it again for some issue later stable users won't get that revision because they'll already be on it.

DomT4:
Mike merged an alternative PR in https://github.com/Homebrew/homebrew/commit/8ef29e6c9ffa2fabe6b5573cbbff7ec5034032c7 this morning, so closing here. Thanks for the submission nonetheless!

PR #45059, opened 2015-10-17
git-pr: initial formula

ikalnitsky:
A git command for fetching GitHub’s pull requests to local copy.


bfontaine:
This seems too new and not popular enough to be included in the core; we’re going to pass on this sorry!

However you can maintain the formula in your own tap ([doc here](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md)) and feel free to re-submit it in the future when it’ll be more popular. Thanks! :smile:

PR #45062, opened 2015-10-17
cmake: use https in head and a comment

vszakats:


bfontaine:
Thank you @vszakats! :tada:

PR #45066, opened 2015-10-17
task: bumped HEAD to 2.5.0 branch

JensErat:


dunn:
Travis caught a few audit failures:
```
==> brew audit task
task:
 * Use spaces instead of tabs for indentation
 * Use spaces instead of tabs for indentation
```

Library/Formula/task.rb
```diff
@@ -14,6 +14,8 @@ class Task < Formula
 
   depends_on "cmake" => :build
   depends_on "gnutls" => :optional
+	# GnuTLS is not optional any more starting with 2.5.0
```
DomT4: The indent is wrong here. Used tabs instead of spaces. If it's mandatory in the head you'll need to do something like:

```ruby
stable do
  url "https://taskwarrior.org/download/task-2.4.4.tar.gz"
  sha256 "7ff406414e0be480f91981831507ac255297aab33d8246f98dbfd2b1b2df8e3b"

  depends_on "gnutls" => :optional
end

head do
  url "https://git.tasktools.org/scm/tm/task.git", :branch => "2.5.0", :shallow => false

  depends_on "gnutls"
end
```

JensErat:
I was sure I fixed this, seems at the wrong place (using Linux and the linuxbrew audit, so had to shuffle around the Formula before running the audit...). To bad there's nothing like `brew audit /path/to/out/of/tree/formula.rb`.

Thank you for the explanation on different dependencies for HEAD and release, I adjusted the commit to your proposal and force-pushed to not clutter with intermediate commits.

DomT4:
>  To bad there's nothing like brew audit /path/to/out/of/tree/formula.rb.

You can `brew audit /Users/Bob/Downloads/formula.rb` and such, if that's what you're asking? :)

JensErat:
> You can `brew audit /Users/Bob/Downloads/formula.rb` and such, if that's what you're asking? :)

Indeed, using absolute paths it works fine, using relative it does not:

    $ brew audit `readlink -f Library/Formula/task.rb`
    $ echo $?
    0
    $ brew audit Library/Formula/task.rb 
    Error: No available formula for library/formula/task.rb
    $ echo $?
    1

I guess the problem is discriminating between local files and casks... But I'm fine with absolute paths.

bfontaine:
Use `./Library/Formula/task.rb` instead of `Library/Formula/task.rb`. The latter is the tap syntax.

JensErat:
The patch itself should be fine now, or has still something to be changed?

bfontaine:
Yes it’s fine; thank you for your contribution @JensErat! :tada:

PR #45068, opened 2015-10-17
swiftgen 0.5.1

AliSoftware:
Fix more installation issues (see AliSoftware/SwiftGen#38)

Now the `rake install` executed by Homebrew will also copy the Swift dylibs over when installing `swiftgen`. This way, it won't depend on the user's `Xcode.app` install path anymore, as SwiftGen will use its own Swift dylibs.

* This ensure that `swiftgen` won't break if the Xcode installed at `/Applications/Xcode.app` is not the same version as the one used when BrewBot brew the bottle
* This allows to use the bottle for everyone, as the Swift dylibs used to build SwiftGen will now be provided with the bottle.

\cc @esttorhe

### DIFF
--- a/Library/Formula/swiftgen.rb
+++ b/Library/Formula/swiftgen.rb
@@ -1,21 +1,14 @@
 class Swiftgen < Formula
   desc "Collection of Swift tools to generate Swift code"
   homepage "https://github.com/AliSoftware/SwiftGen"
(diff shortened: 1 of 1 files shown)


bfontaine:
Thank you a lot for your contribution to Homebrew @AliSoftware! :smile:

PR #45070, opened 2015-10-17
youtube-dl 2015.10.16

dstftw:


bfontaine:
Thank you for your contribution to Homebrew @dstftw; we appreciate it! :tada:

PR #45072, opened 2015-10-17
gtk+ add HEAD

tschoonj:
Since the gtk+ people are taking their time to release a new update,
I figured it may be a good idea to add support for HEAD...

### DIFF
--- a/Library/Formula/gtk+.rb
+++ b/Library/Formula/gtk+.rb
@@ -1,14 +1,33 @@
 class Gtkx < Formula
   desc "GUI toolkit"
   homepage "http://gtk.org/"
-  url "https://download.gnome.org/sources/gtk+/2.24/gtk+-2.24.28.tar.xz"
-  sha256 "b2c6441e98bc5232e5f9bba6965075dcf580a8726398f7374d39f90b88ed4656"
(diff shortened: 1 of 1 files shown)


Library/Formula/gtk+.rb
```diff
@@ -49,10 +61,13 @@ def install
             "--disable-glibtest",
             "--enable-introspection=yes",
             "--with-gdktarget=quartz",
-            "--disable-visibility"]
+            "--disable-visibility",]
 
     args << "--enable-quartz-relocation" if build.with?("quartz-relocation")
 
+    inreplace "autogen.sh", "libtoolize", "glibtoolize" if build.head?
```
DomT4: You can put these in the same block:
```ruby
if build.head?
  inreplace "autogen.sh", "libtoolize", "glibtoolize"
  ENV["NOCONFIGURE"] = "yes"
  system "./autogen.sh"
end
```

tschoonj:
done! Thanks for the tip

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/0772cb6a2c25589ef9914698082509fe518f596d. Thanks Tom!

PR #45077, opened 2015-10-17
pandoc 1.15.1.1

adunning:
Added note on bash completion support (new in this version).

### DIFF
--- a/Library/Formula/pandoc.rb
+++ b/Library/Formula/pandoc.rb
@@ -5,8 +5,8 @@ class Pandoc < Formula
 
   desc "Swiss-army knife of markup format conversion"
   homepage "http://pandoc.org"
-  url "https://hackage.haskell.org/package/pandoc-1.15.0.6/pandoc-1.15.0.6.tar.gz"
-  sha256 "d950968142526d5a765af7dfb26b47c3f60d4f883aa9d9eab668614179f1ed46"
+  url "https://hackage.haskell.org/package/pandoc-1.15.1/pandoc-1.15.1.tar.gz"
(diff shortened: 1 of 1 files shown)


adunning:
Sorry, @dunn, #45073 hadn't been created yet when I looked earlier! Should I close this one, or do we want to keep the note on bash completion support?

dunn:
Oh cool, thanks for noticing that addition! Let's go with this PR.

Since `pandoc --bash-completion` generates a completion script, how about we install it for the user?  Something like this should work, after `cabal_clean_lib`:
```ruby
(bash_completion/"pandoc").write shell_output("#{bin}/pandoc --bash-completion")
```

adunning:
That's an excellent idea. I won't be able to get to this until Monday, so if you want to throw that into your version, that would be brilliant; but if you don't have the time, I can take care of it later.

adunning:
I've updated this for 1.15.1.1. I had to modify the line to install the bash completion script slightly, since `shell_output` doesn't work in that context, but it works otherwise; thanks for the pointer!

dunn:
The test failure was the fault of the bot:
```
[ 4 of 18] Compiling Text.TeXMath.Readers.MathML.EntityMap ( src/Text/TeXMath/Readers/MathML/EntityMap.hs, dist/dist-sandbox-8039fe57/build/Text/TeXMath/Readers/MathML/EntityMap.o )
/var/folders/03/jqjv4qms2b18_hwt2dcgsghc0000gp/T/ghc89119_0/ghc_21.s: hPutBuf: resource exhausted (No space left on device)
cabal: Error: some packages failed to install:
texmath-0.8.4 failed during the building phase. The exception was:
ExitFailure 1

xu-cheng:

PR #45078, opened 2015-10-17
mitmproxy: build on Xcode-only systems

### LINKED ISSUE #44556: mitmproxy failed to build on 10.11
Installing mitmproxy with ```brew install mitmproxy --with-protobuf --with-cssutils --with-pyamf``` fails, as it is using the system Python installation (which is missing ffi.h) even if homebrew Python is also installed.

Gist logs at https://gist.github.com/anonymous/e9d07ea7c55f33c7f2ed

Thank you!

tdsmith:
Should fix #44556.


mikemcquaid:
:+1:

PR #45058, opened 2015-10-17
wine-1.7.53 devel

samrussell:
Travis-ci passes for OSX 10.10 (others in progress): https://travis-ci.org/samrussell/homebrew

I can confirm that flicker patch works just like on 1.7.52

Also includes compressed texture patch to fix games like Black and White etc

PR #45082, opened 2015-10-17
update wine 1.7.53

alexchandel:


DomT4:
Two PRs for the same version bump is enough, let alone three :smile_cat:. Thanks for the willingness to contribute to Homebrew though! :heart:.

Ref: #45058 #45037

PR #45086, opened 2015-10-18
Remove "instal" alias for bash-completion

anemator:
I didn't see "instal" used anywhere, but I suppose it is (or was) there for a reason. Code review appreciated. Thanks!

### DIFF
--- a/Library/Contributions/brew_bash_completion.sh
+++ b/Library/Contributions/brew_bash_completion.sh
@@ -600,7 +600,10 @@ _brew ()
     done
 
     if [[ $i -eq $COMP_CWORD ]]; then
-        __brewcomp "$(brew commands --quiet --include-aliases)"
+        # Do not auto-complete "instal" abbreviation for "install" command.
+        # Prefix newline to prevent not checking the first command.
+        local cmds=$'\n'"$(brew commands --quiet --include-aliases)"
+        __brewcomp "${cmds/$'\n'instal$'\n'/$'\n'}"
         return
     fi
 


Library/Homebrew/global.rb
```diff
@@ -62,7 +62,6 @@ module Homebrew
   "-S" => "search",
   "up" => "update",
   "ln" => "link",
-  "instal" => "install", # gem does the same
```
bfontaine: Please leave it here. It’s a handy shortcut for people typing too fast: `brew instal git` is thus an alias for `brew install git`.

anemator:
How about just removing it from the list of words to auto-complete then, but leave it accessible to those who wish to type out "instal"?

bfontaine:
@anemator Yeah, that’s what I had in mind.

Library/Contributions/brew_bash_completion.sh
```diff
@@ -48,6 +48,7 @@ __brewcomp ()
 
     for s in $1; do
         __brewcomp_words_include "$s" && continue
+        [[ $s == instal ]] && continue
```
bfontaine: Could you please add a comment above this line explaining what’s this line does? It’s hard to understand it if you don’t know the context.

Library/Contributions/brew_bash_completion.sh
```diff
@@ -48,6 +48,8 @@ __brewcomp ()
 
     for s in $1; do
         __brewcomp_words_include "$s" && continue
+        # do not auto-complete "instal" abbreviation for "install"
+        [[ $s == instal ]] && continue
         list="$list$s$sep"
```
UniqMartin: Sorry, but this is totally the wrong place to do this. `__brewcomp` is used to generate completions for all kinds of arguments, not just the command argument. E.g., if there was a formula named `instal`, it would never be completed.

Instead, the filtering should be done on the list obtained from `brew commands`. May I suggest:
```patch
--- i/Library/Contributions/brew_bash_completion.sh
+++ w/Library/Contributions/brew_bash_completion.sh
@@ -579,7 +579,7 @@ _brew_uses ()

 _brew ()
 {
-    local i=1 cmd
+    local i=1 cmd cmds

     # find the subcommand
     while [[ $i -lt $COMP_CWORD ]]; do
@@ -600,7 +600,8 @@ _brew ()
     done

     if [[ $i -eq $COMP_CWORD ]]; then
-        __brewcomp "$(brew commands --quiet --include-aliases)"
+        cmds="$(brew commands --quiet --include-aliases)"
+        __brewcomp "${cmds/$'\n'instal$'\n'/$'\n'}"
         return
     fi
```

(This assumes that `instal` will never be the first item in the list.)

UniqMartin:
Thank you for pursuing this! It has bothered me for a long time, but I never came around to fixing this.

Library/Contributions/brew_bash_completion.sh
anemator: Good catch, thanks! I see the issue you're talking about it. Would you mind explaining that syntax for representing newline though? I looked through the docs, but couldn't find much.

anemator:
Definitely, thanks for helping!

Library/Contributions/brew_bash_completion.sh
```diff
@@ -600,7 +600,10 @@ _brew ()
     done
 
     if [[ $i -eq $COMP_CWORD ]]; then
-        __brewcomp "$(brew commands --quiet --include-aliases)"
+        # Do not auto-complete "instal" abbreviation for "install" command.
+        # Prefix newline to prevent not checking the first command.
+        local cmds="\n$(brew commands --quiet --include-aliases)"
```
UniqMartin: The `\n` is not interpreted as you seem to expect. In fact, now the first entry in the list is a literal `\n--cache` instead of `--cache` (just `echo` the string on your command line to check). To get an interpretation like in C (and below) use:

```
$'\n'"$(brew commands --quiet --include-aliases)"
```

Library/Contributions/brew_bash_completion.sh
```diff
@@ -600,7 +600,10 @@ _brew ()
...
+        __brewcomp "${cmds//$'\n'instal$'\n'/$'\n'}"
```
UniqMartin: This is really just a minor detail and won't be noticeable performance-wise, but there should be no need to replace all instances in the string, because there should be only one line containing `instal`. To replace only one match:

```
"${cmds/$'\n'instal$'\n'/$'\n'}"
```

Library/Contributions/brew_bash_completion.sh
UniqMartin: Have a look at the `bash` man page in the [QUOTING section](http://manpages.ubuntu.com/manpages/vivid/en/man1/bash.1.html#contenttoc11). The relevant stuff starts with “Words of the form $'string' are treated specially.”

anemator: Excellent, Thanks!

UniqMartin:
Saw your question about the `$'\n'` syntax only after complaining a little more. But answered it now. :)

Library/Contributions/brew_bash_completion.sh
anemator: Yeah didn't realize that it wouldn't expand, since the behavior was as expected even for ```--cache```. Thanks!

Library/Contributions/brew_bash_completion.sh
anemator: I added that just in case and since as you mentioned performance impact will be negligible in this case.

anemator:
No worries, just happy to have an answer.

UniqMartin:
As far as I'm concerned, it's ready now. :+1:

Library/Contributions/brew_bash_completion.sh
```diff
@@ -600,7 +600,10 @@ _brew ()
     done
 
     if [[ $i -eq $COMP_CWORD ]]; then
-        __brewcomp "$(brew commands --quiet --include-aliases)"
+        # Do not auto-complete "instal" abbreviation for "install" command.
+        # Prefix newline to prevent not checking the first command.
+        local cmds=$'\n'"$(brew commands --quiet --include-aliases)"
+        __brewcomp "${cmds/$'\n'instal$'\n'/$'\n'}"
```
bfontaine: Can’t we do something like `__brewcomp "$(brew commands --quiet --include-aliases | sed 's/ instal / /')"`?

UniqMartin: We could, but invoking another external command like `sed` is much more expensive that doing a string replacement in pure Bash. (We already have to call `brew` to get the list of commands.)

bfontaine: Thanks; that makes sense.

bfontaine:
Yeah sorry for the delay; thank you both! :smile:

PR #45112, opened 2015-10-18
brew: tighten sudo checks

DomT4:
`reinstall` can currently be executed under sudo and has every chunk of problems running install under sudo does when `brew` isn't root owned.

Likewise with `postinstall` which we increasingly use to create directories and move/install files elsewhere. Using sudo there both masks deeper issues and makes uninstall troublesome.

### DIFF
--- a/Library/brew.rb
+++ b/Library/brew.rb
@@ -97,7 +97,7 @@ begin
 
   cmd = HOMEBREW_INTERNAL_COMMAND_ALIASES.fetch(cmd, cmd)
 
-  sudo_check = %w[ install link pin unpin upgrade ]
+  sudo_check = %w[ install reinstall postinstall link pin unpin upgrade ]
 
   if sudo_check.include? cmd
     if Process.uid.zero? && !File.stat(HOMEBREW_BREW_FILE).uid.zero?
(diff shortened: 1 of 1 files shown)


Library/brew.rb
```diff
@@ -97,7 +97,7 @@ def require?(path)
 
   cmd = HOMEBREW_INTERNAL_COMMAND_ALIASES.fetch(cmd, cmd)
 
-  sudo_check = %w[ install link pin unpin upgrade ]
+  sudo_check = %w[ install reinstall postinstall link pin unpin upgrade ]
```
mikemcquaid: I'd be tempted to add: `update create migrate tap switch pin`

DomT4: `pin` is already in there. Agree about the rest; all would/could create or symlink files/folders that would then fail to remove without another application of `sudo`.

mikemcquaid: > all would/could create or symlink files/folders that would then fail to remove without another application of sudo.

Exactly my thoughts :+1:

DomT4: If nobody objects by the time I start to tread towards sleep late tonight I'll commit this with the extra suggested commands.

dunn:
:+1:

PR #45116, opened 2015-10-18
ginac 1.6.5

jgarbs:


xu-cheng:

PR #45121, opened 2015-10-18
bacula-fd launchd configuration

### LINKED ISSUE #45004: bacula-fd launchd configuration
Hello,

the following is a working launchd configuration, tested on OSX 10.10 with Bacula 5 & 7, which may be installed to ```/Library/LaunchDaemons/homebrew.mxcl.bacula-fd.plist```.

```
<?xml version="0.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <dict>
    <key>Label</key>
    <string>homebrew.mxcl.bacula-fd</string>
    <key>ProgramArguments</key>
    <array>
      <string>/usr/local/bin/bacula-fd</string>
      <!-- make sure this doesn't contain any spaces -->
      <string>-c/usr/local/etc/bacula-fd.conf</string>
      <string>-f</string>
    </array>
    <key>RunAtLoad</key>
    <true/>
    <key>UserName</key>
    <string>root</string>
  </dict>
</plist>
```

I also wanted to have mentioned previous bacula contributors for comments and corrections @jasperla @tgvaughan @ewindisch

sts:
Adding bacula launchd plist configuration, closes #45004

### DIFF
--- a/Library/Formula/bacula-fd.rb
+++ b/Library/Formula/bacula-fd.rb
@@ -36,4 +36,27 @@ class BaculaFd < Formula
     # Create the working directory:
     (var + "lib/bacula").mkpath
   end
+
+  plist_options :startup => true, :manual => "bacula-fd"
+
+  def plist; <<-EOS.undent
+    <?xml version="0.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+      <dict>
+        <key>Label</key>
+        <string>#{plist_name}</string>
+        <key>RunAtLoad</key>
+        <true/>
+        <key>ProgramArguments</key>
+        <array>
+          <string>#{opt_bin}/bacula-fd</string>
+          <string>-f</string>
+        </array>
+        <key>UserName</key>
+        <string>root</string>
+      </dict>
+    </plist>
+  EOS
+  end
 end


Library/Formula/bacula-fd.rb
```diff
@@ -13,6 +13,29 @@ class BaculaFd < Formula
   depends_on "readline"
   depends_on "openssl"
 
+  def plist; <<-EOS.undent
```
DomT4: This should be underneath `def install`.

Library/Formula/bacula-fd.rb
```diff
@@ -13,6 +13,29 @@ class BaculaFd < Formula
...
+    <?xml version="0.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+      <dict>
+        <key>Label</key>
+        <string>#{plist_name}</string>
+        <key>RunAtLoad</key>
+        <true/>
+        <key>UserName</key>
+        <string>root</string>
```
DomT4: Can the tool not run without root access?

Library/Formula/bacula-fd.rb
DomT4: You'll probably want a line like `plist_options :manual => "bacula-fd"` as well, depending on whether it must be root or not.

Library/Formula/bacula-fd.rb
```diff
@@ -13,6 +13,29 @@ class BaculaFd < Formula
...
+        <key>ProgramArguments</key>
+        <array>
+          <string>#{opt_bin}/bacula-fd</string>
+          <!-- make sure parameter/argument don't contain any spaces! -->
+          <string>-c/usr/local/etc/bacula-fd.conf</string>
```
DomT4: Homebrew won't always be in `/usr/local`. That prefix is our default, but many users choose to install elsewhere.

sts: Thank you, I must have overlooked this one. I tested it, it seems to be the default location for the config file, so I will remove this option.

Library/Formula/bacula-fd.rb
sts: Yes it can, the question is how much sense it would make to run a backup agent without root access. I added a comment around the user key/value, so one is able to enable it.

Library/Formula/bacula-fd.rb
sts: I added the plist_options call and moved the method below install.

dunn:
Most plists are "ready-to-go", so maybe we should add a quick caveat here about editing the plist to run as root?

DomT4:
If it probably should be run under root we can do:

```ruby
plist_options :startup => true, :manual => "bacula-fd"
```
That'll place the plist in `/Library/LaunchDaemons`. Given being non-root removes functionality here I'd be fine with simplifying the plist to target that use-case and letting folks without root access make the necessary adjustments themselves.

sts:
+1 for DomT4

apjanke:
Ping... are you going to do that `:startup => true` bit? Would also be nice to squash this to a single commit for tidy history.

sts:
Rebased.

mikemcquaid:

For future reference the preferred commit message format is `bacula-fd: add launchd plist configuration.`.

Issue #44211, opened 2015-09-21
CZMQ: --with-libsodium isn't recognized


paddor:
According to https://github.com/zeromq/czmq/issues/1107#issuecomment-141929930, that option is a relic. I guess it can be removed?

dunn:
Yeah, looks like it.  We'd be very grateful for a pull request! Take a look at https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md and let us know if you have questions.

DomT4:
Looks like it should be flipped around to be:

```ruby
args << "--without-libsodium" if build.without? "libsodium"
```

dunn:
Oh, I figured with superenv it would just fail to find it.

DomT4:
Sadly `superenv` doesn't *yet* protect `include` or `lib`. It only punts `bin` out of the `$PATH`. I presume if it is now the upstream default, it could well be picked up on local installs when `libsodium` is already present.

Issue #44210, opened 2015-09-21
czmq --HEAD: `make check` fails


paddor:
When trying to install the latest version of CZMQ with the command `brew install -v czmq --HEAD --with-libsodium`, `make check` fails. See https://gist.github.com/paddor/4a64778ae3b77f54bc3a.

Checking out CZMQ manually and running the same command sequence as run by Homebrew, it doesn't fail.

I'm running OSX 10.10.5 and tried with both ZMQ stable and `--HEAD` (which installs without problems). Same outcome.

dunn:
I'm getting the same error with `--HEAD` and without.  Looking into it.

paddor:
Okay, thanks @dunn. What about `--HEAD`? It doesn't happen if I install it manually (without Homebrew) on my machine.

DomT4:
The failure is:

```
 * zconfig: OK
 * zdigest: OK
Assertion failed: (subdir), function s_posix_populate_entry, file src/zdir.c, line 93.
 * zdir: make[2]: *** [check-local] Abort trap: 6
make[1]: *** [check-am] Error 2
make: *** [check-recursive] Error 1
```

paddor:
Yeah, same here. But do you know why? @dunn

dunn:
I don't know why, and I wasn't able to build it outside Homebrew to compare. Hopefully I'll have time to try again this weekend.

paddor:
Strange. I can compile and `make check` it without any issues, if I do it manually, using the following commands:

```
git clone git@github.com:zeromq/czmq.git && cd czmq
./autogen.sh
./configure --disable-dependency-tracking --prefix=/usr/local/Cellar/czmq/HEAD
make
make check
```

What's the error on your machine? @dunn

dunn:
Oh, I just needed to set `XML_CATALOG_FILES`.  Now it builds and `make check` passes outside of Homebrew for me as well.

paddor:
What does that environment variable have to do with compiling CZMQ? And to what value did you have to set it?

paddor:
And the big question is still: Why does it fail with Homebrew?

dunn:
I set it to `/usr/local/etc/xml/catalog`; which is often required when running docbook or asciidoc in the Homebrew environment.

No clue why it's failing at `make check`, though.

samrussell:
[...]

PR #45130, opened 2015-10-19
fix bugs #44211 and #44210

samrussell:
- `make test` now passes
- libsodium dependency handled correctly
- removed reference to "makecert" as zeromq now calls this "zmakecert"


mikemcquaid:

For future reference the preferred commit message format is `czmq: fix some issues.`.

Issue #45089, opened 2015-10-18
List outdated formulae separately in `brew update`


Alhadis:
Recent updates have attempted to highlight outdated brews to users by [adding coloured highlights](https://github.com/Homebrew/homebrew/pull/44335), which [didn't play too well](https://github.com/Homebrew/homebrew/issues/45028) with some terminal colour schemes.

Currently, any installed formulae that're affected by `brew update` are displayed in bold with ` (installed)` appended. While a lot easier on the eyes than white-against-yellow, it adds clutter to the feedback - and some users probably aren't that quick to notice the bold letters.

I suggest simply listing affected formulae after the updates:
![1](https://cloud.githubusercontent.com/assets/2346707/10563252/8ce53528-75ce-11e5-8ca1-053fcaa7bfc7.png)
Ignore what's listed in the example, I had to improvise with makeshift feedback, since all my brews are currently up-to-date. Heh.

Thoughts?

mikemcquaid:
Can you confirm what you've changed here, I'm not totally sure. Thanks!

Alhadis:
Well, there's the addition of this at the bottom:
```
==> Outdated Formulae
emscripten (1.34.0 < 1.34.12)
clojurescript (1.7.13 < 1.7.145)
  2 installed formulae require updates
```
Instead of relying on something like bold text to draw a user's attention to outdated modules, it's probably better to list them explicitly, as though the user themselves had typed in `brew outdated` after updating. :)

mikemcquaid:
Basically just running `brew outdated` does make sense to me.

bfontaine:
What does the output look like for revision bumps?

Also I’m not sure about the wording of “2 installed formulae require updates”; they don’t *require* updates per se; you’re free to keep the version you have forever.

Alhadis:
Haha, true, I think "`can be upgraded`" would drill it down nicely. :)

As for revision bumps, well... if the version is the same, you can drop the version diff altogether:

```
==> Affected Formulae
emscripten (1.34.0 < 1.34.12)
clojurescript
  1 installed formula can be upgraded
```

Now I'm thinking that `Affected Formulae` might be a more fitting title, since not every formulae that's affected will be outdated (given the possibility of revision bumps).

UniqMartin:
> Basically just running `brew outdated` does make sense to me.

[...]

PR #45131, opened 2015-10-19
Use `(installed)` and emoji ticks consistently.

mikemcquaid:
This feels more consistent with what we're doing elsewhere now. I can't decide whether this should also be bold. It does feel like there should be a utility method somewhere that let's me pass a formula and receive back either the formula name or the formula name with highlight and `(installed)` if it's already installed.

CC @UniqMartin @Alhadis

### DIFF
--- a/Library/Homebrew/cmd/search.rb
+++ b/Library/Homebrew/cmd/search.rb
@@ -42,8 +42,7 @@ module Homebrew
       query = ARGV.first
       rx = query_regexp(query)
       local_results = search_formulae(rx)
-      local_results_installed = local_results.select { |f| f.end_with? "(installed)" }
-      puts_columns(local_results, local_results_installed)
+      puts_columns(local_results)
       tap_results = search_taps(rx)
       puts_columns(tap_results)
 
@@ -157,7 +156,7 @@ module Homebrew
       if aliases.include?(name) && results.include?(canonical_full_name)
         next
       elsif (HOMEBREW_CELLAR/canonical_name).directory?
-        "#{name} (installed)"
+        pretty_installed(name)
       else
         name
       end
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -455,30 +455,20 @@ class Report
   end
 
   def dump_formula_report(key, title)
-    formula = select_formula(key)
-    unless formula.empty?
-      # Determine list item indices of installed formulae.
-      formula_installed_index = formula.each_index.select do |index|
-        name, newname = formula[index]
-        installed?(name) || (newname && installed?(newname))
-      end
-
-      # Format list items of renamed formulae.
+    formula = select_formula(key).map do |name, new_name|
+      # Format list items of renamed formulae
       if key == :R
-        formula.map! { |oldname, newname| "#{oldname} -> #{newname}" }
+        new_name = pretty_installed(new_name) if installed?(name)
+        "#{name} -> #{new_name}"
+      else
+        installed?(name) ? pretty_installed(name) : name
       end
+    end
 
-      # Append suffix '(installed)' to list items of installed formulae.
-      formula_installed_index.each do |index|
-        formula[index] += " (installed)"
-      end
-
-      # Fetch list items of installed formulae for highlighting.
-      formula_installed = formula.values_at(*formula_installed_index)
-
+    unless formula.empty?
       # Dump formula list.
       ohai title
-      puts_columns(formula, formula_installed)
+      puts_columns(formula)
     end
   end
 
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -9,6 +9,20 @@ require "open-uri"
 
 class Tty
   class << self
+    def tick
+      # necessary for 1.8.7 unicode handling since many installs are on 1.8.7
+      @tick ||= ["2714".hex].pack("U*")
+    end
+
+    def cross
+      # necessary for 1.8.7 unicode handling since many installs are on 1.8.7
+      @cross ||= ["2718".hex].pack("U*")
+    end
+
+    def strip_ansi(string)
+      string.gsub(/\033\[\d+(;\d+)*m/, "")
+    end
+
     def blue
       bold 34
     end
@@ -101,6 +115,26 @@ end
 def odie(error)
   onoe error
   exit 1
+end
+
+def pretty_installed(f)
+  if !$stdout.tty?
+    "#{f}"
+  elsif ENV["HOMEBREW_NO_EMOJI"]
+    "#{Tty.highlight}#{Tty.green}#{f} (installed)#{Tty.reset}"
+  else
+    "#{Tty.highlight}#{f} #{Tty.green}#{Tty.tick}#{Tty.reset}"
+  end
+end
+
+def pretty_uninstalled(f)
+  if !$stdout.tty?
+    "#{f}"
+  elsif ENV["HOMEBREW_NO_EMOJI"]
+    "#{Tty.red}#{f} (uninstalled)#{Tty.reset}"
+  else
+    "#{f} #{Tty.red}#{Tty.cross}#{Tty.reset}"
+  end
 end
 
 def pretty_duration(s)
@@ -267,7 +301,7 @@ def curl(*args)
   safe_system curl, *args
 end
 
-def puts_columns(items, highlight = [])
+def puts_columns(items)
   return if items.empty?
 
   unless $stdout.tty?
@@ -278,20 +312,14 @@ def puts_columns(items, highlight = [])
   # TTY case: If possible, output using multiple columns.
   console_width = Tty.width
   console_width = 80 if console_width <= 0
-  max_len = items.max_by(&:length).length
+  plain_item_lengths = items.map { |s| Tty.strip_ansi(s).length }
+  max_len = plain_item_lengths.max
   col_gap = 2 # number of spaces between columns
   gap_str = " " * col_gap
   cols = (console_width + col_gap) / (max_len + col_gap)
   cols = 1 if cols < 1
   rows = (items.size + cols - 1) / cols
   cols = (items.size + rows - 1) / rows # avoid empty trailing columns
-
-  plain_item_lengths = items.map(&:length) if cols >= 2
-  if highlight && highlight.any?
-    items = items.map do |item|
-      highlight.include?(item) ? "#{Tty.highlight}#{item}#{Tty.reset}" : item
-    end
-  end
 
   if cols >= 2
     col_width = (console_width + col_gap) / cols - col_gap
--- a/Library/Homebrew/cmd/info.rb
+++ b/Library/Homebrew/cmd/info.rb
@@ -147,24 +147,8 @@ module Homebrew
   end
 
   def decorate_dependencies(dependencies)
-    # necessary for 1.8.7 unicode handling since many installs are on 1.8.7
-    tick = ["2714".hex].pack("U*")
-    cross = ["2718".hex].pack("U*")
-
     deps_status = dependencies.collect do |dep|
-      if dep.installed?
-        color = Tty.green
-        symbol = tick
-      else
-        color = Tty.red
-        symbol = cross
-      end
-      if ENV["HOMEBREW_NO_EMOJI"]
-        colored_dep = "#{color}#{dep}"
-      else
-        colored_dep = "#{dep} #{color}#{symbol}"
-      end
-      "#{colored_dep}#{Tty.reset}"
+      dep.installed? ? pretty_installed(dep) : pretty_uninstalled(dep)
     end
     deps_status * ", "
   end


Alhadis:
Would this affect the tick's presence in `brew info`? Because I can't really see that looking too readable...

screen shot 2015-10-19 at 10 42 03 pm

I think ` (installed)` should really be reserved for instances where a formulae's details is being queried directly, and it's the subject of a command (which it's currently doing now if you run `brew info`). It's a little too verbose to comfortably wedge beside formula names willy-nilly...

bfontaine:
Sorry wrong button

mikemcquaid:
@Alhadis Yes, it would replace the tick.

Alhadis:
Erm, if somebody wanted to install something very dependency-heavy (such as `ffmpeg` or `imagemagick`), they'd be reduced to reading the results very carefully in order to ascertain what they lack (or could install to improve their brew's capabilities):



Being able to visually grep brews based on coloured glyphs is a huge asset, and part of what makes Homebrew's feedback so friendly and minimalist.

xu-cheng:
Personally, I prefer the old emoji :smile:

mikemcquaid:
@xu-cheng :grinning: I'm open to whatever, I just think we should be consistent.

Alhadis:
Then I humbly suggest we append ticks to the names of installed formulae in the feedback of `brew update`/`brew search`, et al.

Copying a visual I [prepared earlier](https://github.com/Homebrew/homebrew/issues/45089#issuecomment-149190905):
f3a84a0a-76ac-11e5-8ffd-40a72301bf50

There's no ambiguity as to what the highlighting means if there's a line at the bottom making the connection. :) Taking this approach keeps this concise and consistent without sacrificing what it communicates to the reader.

bfontaine:
That might sound nitpicky but having a tick *after* installed names then *before* the “can be updated” line confuse me :smile:

UniqMartin:
I also think we should be as consistent as possible, but within reason.

`brew info` is a case where the emoji works way better and makes the list (especially if it's long) actually readable. Appending `(installed)` almost everywhere wouldn't be as legible. More importantly, I think this is a case where it's obvious from the context what the symbols mean.

`brew search` and `brew update` are different in that some symbol next to a formula wouldn't be as obvious. I'm still in favor of keeping the `(installed)` there. Or if we really go for a symbol like the tick mark, there would have to be a legend saying what it means.

Alhadis:
> That might sound nitpicky but having a tick after installed names then before the “can be updated” line confuses me

Well, we could use this instead:
8a29b6f2-769f-11e5-81f9-cd650995cf0a

>  Or if we really go for a symbol like the tick mark, there would have to be a legend saying what it means.

Yup, that's what I suggested in my second-last comment. The one-liner at the bottom which combines the icon's meaning *and* a useful summary.

xu-cheng:
For the consistence part, we may consider to remove `(install)` for `brew search <query>` if not in `stty`. This thing is: (a) we don't append `(install)` for `brew search` (no argument)  (b) appending for `(install)` could be bad for script which tries to call `brew search`

mikemcquaid:
@xu-cheng Doesn't `brew search` also output all the `==>` stuff to stdout? If so I think it's mostly moot.

> brew info is a case where the emoji works way better and makes the list (especially if it's long) actually readable. Appending (installed) almost everywhere wouldn't be as legible. More importantly, I think this is a case where it's obvious from the context what the symbols mean.

It's not obvious to everyone; I've had people express surprise about that before.

UniqMartin:
> (a) we don't append `(install)` for `brew search` (no argument)

We should also be appending `(installed)` in this case for consistency. Until now I wasn't even aware that search works without arguments. :) Might be a good idea to drop these suffixes if the output isn't going to a TTY, but not sure if that would create more confusion …

xu-cheng:
> Doesn't brew search also output all the ==> stuff to stdout?

No, it doesn't print `==>` at least not for non stty env.

> It's not obvious to everyone; I've had people express surprise about that before.

I think it's kinda obvious while also quite compact.

> We should also be appending (installed) in this case for consistency. 

We cannot just do this, because at least `brew search` is used in shell completion to list all formulae.

Alhadis:
Even if it's not obvious from context, the bit at the bottom should elucidate. Ahem.

UniqMartin:
> It's not obvious to everyone; I've had people express surprise about that before.

We could drop the emoji, but keep the color, *and* append `(installed)` as suggested in this PR. That way it would be less compact, but still pretty easy to scan due to color coding and more obvious:

----

`brew info` without emoji but with color

----

> We cannot just do this, because at least `brew search` is used in shell completion to list all formulae.

We definitely should drop the suffix for the non-TTY case. Probably also for the case with search query and for other commands that append `(installed)`.

Alhadis:
Uh, that looks way too busy to me, *and* it's overusing colour (nowhere else are brew names displayed in different colours... right?)

Here we are spoon-feeding text to users because a minority didn't make the connection between a tick and its implied meaning... :(

xu-cheng:
> We could drop the emoji, but keep the color, and append (installed) as suggested in this PR. That way it would be less compact, but still pretty easy to scan due to color coding and more obvious:

I still think too many `(install)` are redundant. Referring to your `octave` example, the UI would not be good if you have all of deps installed. Now I'm considering appending emoji for `search/update`.

Since it's a UI/UX problem, we may better to get more opinion among the users before rushing into any decision.

> We definitely should drop the suffix for the non-TTY case. Probably also for the case with search query and for other commands that append (installed).

Agree

UniqMartin:
Whatever we decide: There were some implied suggestions of using underlining for highlighting. Please don't! It's a mode of highlighting carried over from pen and paper, where there are very few alternatives. On a computer screen there is color, boldface, italics, and probably more. The former don't hurt legibility if used sparingly, but underlining always does. (That much I learned when I dabbled with typography. Just think for a second why Markdown has no syntax for underlining.)

Alhadis:
Already replied to that [here](https://github.com/Homebrew/homebrew/issues/45089#issuecomment-149217907), which I'll copy+paste here for the sake of not splitting discussion:

===

Markdown has no syntax for a lot of things, mate, let's not go down that road... xD

On a computer screen there is colour, boldface, italics, and probably more.
Yes, but in a terminal emulator, the range of formatting choices becomes a lot more limited than what you're free to use in Photoshop or in CSS. Underlining has the benefit of clarity: not every typeface looks good rendered in bold...

UniqMartin:
> Already replied to that here, which I'll copy+paste here for the sake of not splitting discussion:

Sorry for cross-positing this. Maybe we should keep *that* discussion in the other thread. Or in a separate issue …

bfontaine:
Beware of colors; we don’t all have the same terminal setup: some colors might be ok on a white background but not on a dark, blue or [pink](http://www.idiotinside.com/static/uploads/2015/10/if-else.png) one.

mikemcquaid:
Agreed. You also can't rely on colours exclusively as some people are colour-blind.

Alhadis:
*raises hand* WAI AYE. Colour-blind Geordie right here.

And yeah, which is why there's also the benefit of a tick adding extra clarification. :-) Even if colour's disabled, the shape of the icon is unmistakable.

DomT4:
I like the emoji tick on `info` FWIW. It's clean, doesn't take up a ton of space, and obvious what it means. Agree `(installed)` is more appropriate for `update` and consistency is important but I'm a bit meh about removing the emoji tick here.

> Beware of colors; we don’t all have the same terminal setup: some colors might be ok on a white background but not on a dark

I use bright green on black like the world's most stereotypical TV hacker, and colorisation does pretty much nothing in that context.

mikemcquaid:
https://twitter.com/nautilebleu/status/656095637155500033 is interesting as it shows they think the highlight is showing outdated packages rather than all the installed ones. These packages are not necessarily outdated.

I'm increasingly coming around to just outputting `brew outdated` at the end. If people don't want to update: they can use `brew pin`.

xu-cheng:
>  These packages are not necessarily outdated.

@mikemcquaid They are outdated. Because we only show formula with pkg_version changed in update report.

mikemcquaid:
@xu-cheng I forgot about that change as it's relatively recent :grinning:. I still think the `outdated` output is the most helpful/obvious.

Alhadis:
> I still think the outdated output is the most helpful/obvious.

Aye, damn well seconded. :)

xu-cheng:
> I still think the outdated output is the most helpful/obvious.

FYI, in oh-my-zsh, they already created an alias for this. https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/brew/brew.plugin.zsh#L2

mikemcquaid:
`update` is already verbose. I don't think we should start adding options for such things unless multiple users complain. Options complicate the code and testing.

mikemcquaid:
But it sounds like we're reaching a consensus that `outdated` should be output after `brew update`. I still think we should aim for `info` to be more consistent with other parts of Homebrew, though.

xu-cheng:
> Options complicate the code and testing.

Right.

Another thing I'm worry about is will printing outdated slow down the already slowing `brew update`.

mikemcquaid:
Anyone in here have vast numbers of packages installed and want to benchmark? I don't have many and  it's always super-quick for me.

@xu-cheng Another option is doing some of these calculations while we e.g. wait for other network calls to complete

Alhadis:
Depends what you consider vast, a Scotsman's idea of a large number of kegs is probably different to most. ;-)

Jokes aside, my count's probably not big enough:

    213 kegs, 99241 files, 4.3G

If it is, let me know. :)

xu-cheng:
> I still think we should aim for info to be more consistent with other parts of Homebrew, though.

Personally, I really like to preserve current emoji.

> Anyone in here have vast numbers of packages installed and want to benchmark? 

I think @bfontaine has.

> Another option is doing some of these calculations while we e.g. wait for other network calls to complete

Right, the solution would be multiple threads in update. @UniqMartin any interesting to take over the work on #40745 ?

mikemcquaid:
@Alhadis That looks good.

mikemcquaid:
> Personally, I really like to preserve current emoji.

In which case: shall we consider using the emoji more widely?

xu-cheng:
> In which case: shall we consider using the emoji more widely?

I like the idea.

Alhadis:
> That looks good.

Alright, well, here's the time result, with the speed of `brew update` included for comparison:

```
phoenix:Desktop johngardner$ time brew update
Updated Homebrew from fd2b1ff1 to 6849444f.
No changes to formulae.

real	0m31.080s
user	0m1.724s
sys	0m1.755s
phoenix:Desktop johngardner$ time brew outdated

real	0m0.690s
user	0m0.611s
sys	0m0.077s
```

UniqMartin:
If the “outdated” section would be in fact the version bumps for installed formulae pulled in by the update, then that information is relatively cheap. The update is already looking at the version numbers of formulae to determine if it counts as an updated formula or just as a minor change. Carrying that over into a section “Installed formulae that were updated during this run of `brew update`” (that obviously isn't going to be the title, just wanted to be clear) would be relatively easy.

If the plan is to just stick the output of `brew outdated` under the output of `brew update`, I'm still against this [as explained here](https://github.com/Homebrew/homebrew/issues/45089#issuecomment-149028198). (Short version: Those two commands have less in common than it seems.)

> In which case: shall we consider using the emoji more widely?

I'm alright with that. But we should make sure that there's still a decent rendering without emoji, as some poor souls have to SSH from less capable OSes into their OS X machines. And also it's going to be interesting to support this more involved formatting in the multi-column output …

> @UniqMartin any interesting to take over the work on #40745?

Interested I may be. But that's not something I'm willing to work on right now, sorry.

Alhadis:
> Carrying that over into a section “Installed formulae that were updated during this run of brew update” (that obviously isn't going to be the title, just wanted to be clear) would be relatively easy.

That was actually what I had in mind when I originally suggested it. It's not duplicating the output of `brew outdated`, its more making a summary of what it's collecting and comparing it with what the user has.

mikemcquaid:
@UniqMartin I disagree that `outdated` would become long and annoying in that case, personally.

bfontaine:
@xu-cheng I killed a lot of packages recently because my 128GB SSD doesn’t have much space left :cry: 
```
244 kegs, 89295 files, 4,4G
```

UniqMartin:
> **EDIT:** If I'm becoming annoying, please let me know, haha. ^^ Bit hard to tell if I'm wanted/needed here. :<

No! If I left this impression, I'm sorry. My intention was just to be as clear as possible, because I had the feeling that different people were talking about different things when referring to the “outdated” section.

> I disagree that outdated would become long and annoying in that case, personally.

I'm not opposing the experiment. I'm pretty sure people will tell us, if they feel annoyed once this part is added. :stuck_out_tongue_winking_eye: It might even be healthy to remind some people that a `brew upgrade` once in a while isn't going to hurt.

DomT4:
> Anyone in here have vast numbers of packages installed and want to benchmark?

I have a 7GB Cellar :smile:. Praise the TB HD. What do folks want benching?

> In which case: shall we consider using the emoji more widely?

:+1: from me.

Alhadis:
> No! If I left this impression, I'm sorry.

Haha, no, it's not that. :D It's that I feel like a third wheel in this conversation; bit like that bloke who keeps trying to talk over others without knowing he's no longer needed, haha.

> In which case: shall we consider using the emoji more widely?
:+1: from me too. :)

`++votes;`

mikemcquaid:
It seems we've probably agreed on consistently using the green emoji when emojis are enabled and `(installed)` when they are not. Does that sound reasonable to everyone?

DomT4:
WFM.

dunn:
:+1:

UniqMartin:
Works for me.

Alhadis:
WAI AYE! :+1:

mikemcquaid:
Have updated this to make this behaviour consistent. Let me know what you think.

Library/Homebrew/cmd/search.rb
```diff
@@ -157,7 +157,7 @@ def search_formulae(rx)
       if aliases.include?(name) && results.include?(canonical_full_name)
         next
       elsif (HOMEBREW_CELLAR/canonical_name).directory?
-        "#{name} (installed)"
+        pretty_installed(name)
```
UniqMartin: Sorry to be this blunt, but this change completely ignores the way the result of `search_formulae` is used by the caller:

1. It relies on the `(installed)` suffix to be there to know which items to highlight in the output.
2. The list is passed to `puts_columns` for multi-column output and this function can only deal with plain text without any embedded formatting (but `pretty_installed` adds formatting). That's why `puts_columns` has a second argument to list the items that should be highlighted and does the highlighting internally. Unless `puts_columns` is extended to understand a bare minimum of ANSI escape codes, it won't be able to create a proper multi-column layout with this kind of input.

Library/Homebrew/cmd/update.rb
```diff
@@ -453,8 +453,8 @@ def dump_formula_report(key, title)
       end
 
       # Append suffix '(installed)' to list items of installed formulae.
-      formula_installed_index.each do |index|
-        formula[index] += " (installed)"
+      formula_installed_index.map! do |index|
+        pretty_installed(formula[index])
       end
```
UniqMartin: I'm not sure which thought went into this, but it doesn't make sense to me to transform the indices of the installed formula into the formatted text and then to use this text as indices to access the `formula` array as is done in the following code line. The thing that would make the most sense to me would be to have something like:

```ruby
formula_installed_index.each do |index|
  formula[index] = pretty_installed(formula[index])
end
```

But even then, the code below would require adjustment and `puts_columns` wouldn't be able to create a proper multi-column layout due to the preformatted list items (like explained in my other comment).

Library/Homebrew/utils.rb
```diff
@@ -103,6 +103,26 @@ def odie(error)
   exit 1
 end
 
+def pretty_installed(f)
+  # necessary for 1.8.7 unicode handling since many installs are on 1.8.7
+  tick = ["2714".hex].pack("U*")
```
UniqMartin: Maybe move this into the `else` clause as it is only needed there?

Library/Homebrew/utils.rb
```diff
@@ -103,6 +103,26 @@ def odie(error)
...
+  if ENV["HOMEBREW_NO_EMOJI"]
+    "#{Tty.green}#{f} (installed)#{Tty.reset}"
+  else
+    "#{f} #{Tty.green}#{tick}#{Tty.reset}"
+  end
+end
+
+def pretty_uninstalled(f)
+  # necessary for 1.8.7 unicode handling since many installs are on 1.8.7
+  cross = ["2718".hex].pack("U*")
```
UniqMartin: Same as for the `tick` variable above.

UniqMartin:
> Have updated this to make this behaviour consistent. Let me know what you think.

I'm okay with the general code structure and the intended formatting. But the code isn't doing what it is supposed to do for both `brew seach` and `brew update`, so currently :-1: from me.

Library/Homebrew/utils.rb
xu-cheng: I would suggest to make it a const, so it will only be computed once.

mikemcquaid:
> I'm okay with the general code structure and the intended formatting. But the code isn't doing what it is supposed to do for both brew seach and brew update, so currently  from me.

Sorry, I should have been clearer that this code wasn't tested yet but just wanted to people's feedback on the approach rather than how it actually works. It seems people are broadly in favour of the general approach of how we use ticks or installed so I'll get it actually working now :wink:

Alhadis:
Hrm, it still seems to be displaying the old <code>name <b>(installed)</b></code> cue... or is that because it's not been merged/approved yet?

mikemcquaid:
This has not been merged yet, yeh.

mikemcquaid:
This has been updated to address the various comments here. In either this or a future PR we could perhaps have a `puts_pretty_formulae_columns` which takes in a list of formulae and outputs them in columns prettified only if they are installed.

Library/Homebrew/cmd/search.rb
UniqMartin: The above no longer applies, obviously. As for the use of `search_formulae` in `brew install`, I guess it's better to sort that out in #45707.

Library/Homebrew/utils.rb
```diff
@@ -7,6 +7,9 @@
 require "utils/git"
 require "open-uri"
 
+HOMEBREW_UNICODE_TICK = ["2714".hex].pack("U*")
+HOMEBREW_UNICODE_CROSS = ["2718".hex].pack("U*")
```
UniqMartin: Could the `Tty` class be a better place to host these constants? We could remove the `HOMEBREW_` prefix then.

Library/Homebrew/utils.rb
```diff
@@ -103,6 +106,28 @@ def odie(error)
   exit 1
 end
 
+def pretty_installed(f)
+  if !$stdout.tty?
+    "#{f}"
+  # necessary for 1.8.7 unicode handling since many installs are on 1.8.7
```
UniqMartin: I think this comment belongs above the constants, not here.

Library/Homebrew/utils.rb
```diff
@@ -103,6 +106,28 @@ def odie(error)
...
+    "#{Tty.highlight}#{Tty.green}#{f} (installed)#{Tty.reset}"
+  else
+    "#{Tty.highlight}#{f} #{Tty.green}#{HOMEBREW_UNICODE_TICK}#{Tty.reset}"
+  end
+end
+
+def pretty_uninstalled(f)
+  if !$stdout.tty?
+    "#{f}"
+  # necessary for 1.8.7 unicode handling since many installs are on 1.8.7
```
UniqMartin: Same for this comment.

Library/Homebrew/utils.rb
```diff
@@ -267,7 +292,11 @@ def curl(*args)
   safe_system curl, *args
 end
 
-def puts_columns(items, highlight = [])
+def output_length string
```
UniqMartin: I think a better place for this method would be the `Tty` class. Maybe also rename it to `plain_length`? And haven't we settled on always putting the arguments in parentheses on method definition?

One more thought: Maybe let the method return the string with ANSI escapes removed? It would be the responsibility of the caller to call `length` on the result, but that would make this function a lot more useful in case we ever want to get plain text from formatted strings. If we can agree on that, an even better name might be `strip_ansi`.

Library/Homebrew/utils.rb
```diff
@@ -267,7 +292,11 @@ def curl(*args)
...
+  string.gsub(/\033\[(\d;)?\d+m/, "").length
```
UniqMartin: I think we should make the regexp `/\033\[\d+(;\d+)*m/`. This will handle all cases I can come up with correctly, even those we do not use that currently, e.g. `"\033[1;37;41m"` (bold, foreground color, background color) or `"\e[32;1m"` (foreground color and bold swapped, still valid).

Library/Homebrew/utils.rb
```diff
@@ -278,7 +307,7 @@ def puts_columns(items, highlight = [])
   # TTY case: If possible, output using multiple columns.
   console_width = Tty.width
   console_width = 80 if console_width <= 0
-  max_len = items.max_by(&:length).length
+  max_len = output_length(items.max_by(&method(:output_length)))
```
UniqMartin: This looks a bit unwieldy. If we are running everything through `output_length` anyway, we can also move

```
plain_item_lengths = items.map(&method(:output_length))
```

here (from a few lines below) and then proceed to “compute” `max_len` from that.

Library/Homebrew/utils.rb
```diff
@@ -286,12 +315,7 @@ def puts_columns(items, highlight = [])
   rows = (items.size + cols - 1) / cols
   cols = (items.size + rows - 1) / rows # avoid empty trailing columns
 
-  plain_item_lengths = items.map(&:length) if cols >= 2
-  if highlight && highlight.any?
-    items = items.map do |item|
-      highlight.include?(item) ? "#{Tty.highlight}#{item}#{Tty.reset}" : item
-    end
-  end
+  plain_item_lengths = items.map(&method(:output_length)) if cols >= 2
```
UniqMartin: Is there any benefit over `items.map { |s| output_length(s) }` that, to me, seems easier to grasp?

UniqMartin:
I posted a bunch of comments, but I like the direction this has taken (both code- and output-wise). :+1:

> In either this or a future PR we could perhaps have a `puts_pretty_formulae_columns` which takes in a list of formulae and outputs them in columns prettified only if they are installed.

I think this can wait. Which commands would use that and benefit from it?

UniqMartin:
*Examples (excepts, no full output) for those not willing to test the code themselves:*

`$ brew report 51a477de..353e3573` (like `brew update`):

update-emoji

`$ HOMEBREW_NO_EMOJI= brew report 51a477de..353e3573` (like `brew update`):

update-no-emoji

`$ brew info petsc`:

info-emoji

`$ HOMEBREW_NO_EMOJI= brew info petsc`:

info-no-emoji

mikemcquaid:
@UniqMartin I've addressed all your comments. Take another look.

Library/Homebrew/utils.rb
```diff
@@ -278,7 +312,7 @@ def puts_columns(items, highlight = [])
   # TTY case: If possible, output using multiple columns.
   console_width = Tty.width
   console_width = 80 if console_width <= 0
-  max_len = items.max_by(&:length).length
+  max_len = items.max_by { |s| Tty.strip_ansi(s).length }.length
```
UniqMartin: I'd rather not run everything through `Tty.strip_ansi` twice. Why not put

```ruby
plain_item_lengths = items.map { |s| Tty.strip_ansi(s).length }
max_len = plain_item_lengths.max
```

here and …

Library/Homebrew/utils.rb
```diff
@@ -286,12 +320,7 @@ def puts_columns(items, highlight = [])
   rows = (items.size + cols - 1) / cols
   cols = (items.size + rows - 1) / rows # avoid empty trailing columns
 
-  plain_item_lengths = items.map(&:length) if cols >= 2
-  if highlight && highlight.any?
-    items = items.map do |item|
-      highlight.include?(item) ? "#{Tty.highlight}#{item}#{Tty.reset}" : item
-    end
-  end
+  plain_item_lengths = items.map { |s| Tty.strip_ansi(s).length } if cols >= 2
```
UniqMartin: … and remove this line?

UniqMartin:
> I've addressed all your comments.

I think `brew style` disagrees on that for some newly introduced code :wink: and then there are two additional comments. Other than that, I'm happy with what I'm seeing.

mikemcquaid:
@UniqMartin Cheers for the comments. Final fixups done and then if you :+1: I'll merge. Thanks!

Library/Homebrew/utils.rb
```diff
@@ -9,6 +9,20 @@
...
   class << self
+    def tick
+      # necessary for 1.8.7 unicode handling since many installs are on 1.8.7
+      @tick ||= ["2714".hex].pack("U*")
+    end
+
+    def cross
+      # necessary for 1.8.7 unicode handling since many installs are on 1.8.7
+      @cross ||= ["2718".hex].pack("U*")
+    end
```
xu-cheng: Why not use constant directly?

Library/Homebrew/cmd/update.rb
```diff
@@ -442,28 +442,23 @@ def select_formula(key)
...
+      # Make installed formulae more visible/pretty.
+      formula.map! do |name, newname|
+        name = pretty_installed(name) if installed?(name)
+        newname = pretty_installed(newname) if (newname && installed?(newname))
+        [name, newname]
       end
 
       # Format list items of renamed formulae.
       if key == :R
         formula.map! { |oldname, newname| "#{oldname} -> #{newname}" }
```
xu-cheng: Maybe we should highlight here. So we will have `old name -> new name (installed)` rather than `old name (installed) -> new name (installed)`

Library/Homebrew/utils.rb
mikemcquaid: @UniqMartin suggested this fit more nicely in `Tty` and I think it fits better with the rest of the methods here (each of which is also a function but could be a constant).

Library/Homebrew/cmd/update.rb
mikemcquaid: I think it's nicer to indicate they were both installed as we're leaving around those symlinks for the `old_name`

xu-cheng: That's not defently true. Since migration happens later and won't guarantee to be success.

Library/Homebrew/cmd/update.rb
```diff
@@ -442,28 +442,23 @@ def select_formula(key)
...
     formula = select_formula(key)
     unless formula.empty?
-      # Determine list item indices of installed formulae.
-      formula_installed_index = formula.each_index.select do |index|
-        name, newname = formula[index]
-        installed?(name) || (newname && installed?(newname))
+      # Make installed formulae more visible/pretty.
+      formula.map! do |name, newname|
+        name = pretty_installed(name) if installed?(name)
+        newname = pretty_installed(newname) if (newname && installed?(newname))
```
xu-cheng: Since this happens before migration, only one of them be tagged as installed.

mikemcquaid:
> `$ brew report 51a477de..353e3573` (like `brew update`):

@UniqMartin it'd be great to make that a `dev-cmd`.

Library/Homebrew/cmd/update.rb
mikemcquaid: Good point; fixed to your version.

Library/Homebrew/utils.rb
UniqMartin: Just to be clear, I originally suggested to move the constants into the class. But I definitely prefer the look of `Tty.cross` over `Tty::UNICODE_CROSS`, even if it may come at a certain penalty (I'd guess it is marginally slower).

UniqMartin:
> Cheers for the comments. Final fixups done and then if you :+1: I'll merge. Thanks!

@mikemcquaid I'm a bit sad this was merged so quickly, giving me no chance to react. But since everything I really wanted to see addressed has been addressed, I shall not complain too loudly.

>> `$ brew report 51a477de..353e3573` (like `brew update`):
> 
> it'd be great to make that a `dev-cmd`.

I'll try to publish it soon in [my tools tap](https://github.com/UniqMartin/homebrew-tools) where I intended to put it. And then everyone can have a look and we can move it, given sufficient demand and if having one more tap is too inconvenient. Deal? :smiley_cat:

mikemcquaid:
> @mikemcquaid I'm a bit sad this was merged so quickly, giving me no chance to react. But since everything I really wanted to see addressed has been addressed, I shall not complain too loudly.

Apologies. In future if you just let me know if you'd rather I wait for e.g. another pass.

> I'll try to publish it soon in my tools tap where I intended to put it. And then everyone can have a look and we can move it, given sufficient demand and if having one more tap is too inconvenient. Deal?

If you can publish stuff like this in https://github.com/homebrew/homebrew-dev-tools it'd be great!

PR #45132, opened 2015-10-19
libuecc 5 (new formula)

apfohl:
This PR adds a new formular to Homebrew for `libuecc`.

`libuecc` is an essential dependency for projects like `fastd`[1] and `ecdsautils`[2]. These projects are very important to operate gateways for the Freifunk community[3]. In order to use these projects on OS X it is necessary to have this library installed. It would be nice to see it in Homebrew for a convenient installation.

- [1] https://projects.universe-factory.net/projects/fastd/
- [2] https://github.com/tcatm/ecdsautils
- [3] http://freifunk.net/en/

### DIFF
--- a/Library/Formula/libuecc.rb
+++ b/Library/Formula/libuecc.rb
@@ -1,0 +1,32 @@
+class Libuecc < Formula
+  desc "Very small Elliptic Curve Cryptography library"
+  homepage "http://git.universe-factory.net/libuecc/"
+  url "http://git.universe-factory.net/libuecc/snapshot/libuecc-5.tar"
+  sha256 "5f4104e70e48f077f92395e6652d9a139e3fdbcc4dc51113ddc955bf2a82542a"
+
+  head "git://git.universe-factory.net/libuecc"
+
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", ".", *std_cmake_args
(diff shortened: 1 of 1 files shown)


Library/Formula/libuecc.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      int main(void)
+      {
+          ecc_int256_t secret;
+          ecc_25519_gf_sanitize_secret(&secret, &secret);
+
+          return EXIT_SUCCESS;
+      }
+    EOS
+    args = `pkg-config --cflags --libs libuecc`.split
```
bfontaine: Please use `shell_output("pkg-config --cflags --libs libuecc")` instead of `` `pkg-config --cflags --libs libuecc` ``.

bfontaine: In fact you can’t rely on `pkg-config` being installed so you’ll have to write the flags directly in the test.

apfohl: All right, will improve that.

apfohl:
Done. I fixed the small remaining issues.

dunn:
Awesome! Merged in e44bd3d460e98ba29b12b05aff78fc0cbf9a946a.

PR #45134, opened 2015-10-19
Provide better link when rate limit exceeded error is thrown

pstadler:
The form at https://github.com/settings/tokens/new can be prefilled using query parameters. This is purely for convenience.

Example Error with improved link:
```
Error: GitHub API rate limit exceeded for 212.51.144.2. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
Try again in 47 minutes 1 seconds, or create an personal access token:
  https://github.com/settings/tokens/new?scopes=&description=Homebrew+on+Sledgehammer.local
and then set the token as: HOMEBREW_GITHUB_API_TOKEN
```

Result:

screen shot 2015-10-19 at 14 55 50

### DIFF
--- a/Library/Homebrew/manpages/brew.1.md
+++ b/Library/Homebrew/manpages/brew.1.md
@@ -636,6 +636,8 @@ can take several different forms:
     <https://developer.github.com/v3/#rate-limiting> for more information.
     Homebrew uses the GitHub API for features such as `brew search`.
 
+    *NOTE*: Homebrew doesn't require permissions for any of the scopes.
+
   * HOMEBREW\_LOGS:
     If set, Homebrew will use the given directory to store log files.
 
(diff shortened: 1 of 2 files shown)


bfontaine:
Great idea!

mikemcquaid:
I think I'd rather the description is just set to "Homebrew" but otherwise :+1:

pstadler:
As one should not expose API tokens to the public, it's better to have one for each machine and therefore it makes sense to have the host name in there :+1:

mikemcquaid:
It actually doesn't matter that much if you expose an API token that has only public read access to the public (so GitHub won't automatically revoke them when they discovers them committed to public repos). I also think it's reasonable to e.g. use 1Password to share a token between machines.

xu-cheng:
I suggest we enable gist scope by default, otherwise users wouldn't be able to run `brew gist-logs`.(although this could be considered as a bug in Homebrew).

bfontaine:
You can keep “Homebrew” for the title then let users append whatever they want to it :+1:

pstadler:
I force pushed the changes you asked for.

mikemcquaid:

PR #45135, opened 2015-10-19
gst-plugins-good: patch to compile on < Lion

rhendric:
Patch accepted [upstream](http://cgit.freedesktop.org/gstreamer/gst-plugins-good/commit/?id=fc203a4bd7eb1cecc0e17bcb7ec67e0672806867).

### DIFF
--- a/Library/Formula/gst-plugins-good.rb
+++ b/Library/Formula/gst-plugins-good.rb
@@ -7,6 +7,16 @@ class GstPluginsGood < Formula
     sha256 "a0915639595305e48884656e22b16fda7c6892aa02cdb3eb43e23dab6e6b81fa"
 
     depends_on "check" => :optional
+
+    if MacOS.version < :lion
+      # Snow Leopard and below don't have strnlen()
+      # https://bugzilla.gnome.org/show_bug.cgi?id=756154
+      # http://cgit.freedesktop.org/gstreamer/gst-plugins-good/commit/?id=fc203a4bd7eb1cecc0e17bcb7ec67e0672806867
+      patch do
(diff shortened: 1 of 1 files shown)


Library/Formula/gst-plugins-good.rb
```diff
@@ -50,6 +50,15 @@ class GstPluginsGood < Formula
 
   depends_on "libogg" if build.with? "flac"
 
+  if MacOS.version < :lion
+    # Snow Leopard and below don't have strnlen()
+    # https://bugzilla.gnome.org/show_bug.cgi?id=756154
+    patch do
+      url "http://cgit.freedesktop.org/gstreamer/gst-plugins-good/patch/?id=fc203a4bd7eb1cecc0e17bcb7ec67e0672806867"
```
DomT4: Could you possibly submit this to https://github.com/Homebrew/patches and use that URL? We've found `cgit` URLs tend to churn the checksum every now and again and then we have issues verifying with upstream the change wasn't malicious.

DomT4: If it's been accepted upstream you'll want to tuck this inside `stable do` further up as well.

rhendric: It's also available from the [Bugzilla](https://bugzilla.gnome.org/attachment.cgi?id=313403&action=diff&context=patch&collapsed=&headers=1&format=raw); would Homebrew/patches still be preferable?

DomT4: Bugzilla should be fine in this case then. That one shouldn't move around in terms of SHA.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/080d7efbd480fffc20003d088ee2f8744e5c38fb. Thank you for another contribution to Homebrew @rhendric! :smiley_cat:

Issue #45114, opened 2015-10-18
qt5 5.5.1 breaks PyQt5 (QtDBus)


The-Compiler:
Since the Qt5 update to 5.5.1, it seems PyQt5 is broken.

When doing `import PyQt5.QtGui`, I get:

```
ImportError: dlopen(.../lib/python3.5/site-packages/PyQt5/QtGui.so, 2): Library not loaded: /usr/local/opt/qt5/lib/QtDBus.framework/Versions/5/QtDBus
  Referenced from: .../lib/python3.5/site-packages/PyQt5/QtGui.so
  Reason: image not found
```

I'd guess this is because of 0c6557f48665221574685da8cae019e6ee3e47b8?

/cc @UniqMartin @mikemcquaid

UniqMartin:
Yes, this is quite likely because of the mentioned commit. Can you rebuild `pyqt5` (or build it from source, if you used the bottle) and see if that fixes the error (it should)?

I'll also have a look. Maybe that's not the only formula that needs a revision bump.

The-Compiler:
The build seems to fail:

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -headerpad_max_install_names -undefined dynamic_lookup -install_name /usr/local/Cellar/pyqt5/5.5_1/lib/python3.5/site-packages/PyQt5
/QtGui.so -framework QtPrintSupport -framework QtDBus -framework QtWidgets -Wl,-syslibroot,/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk -mmacosx-version-min=10.10 -Wl,-rpat
h,/usr/local/Cellar/qt5/5.5.1/lib -single_module -dynamiclib -o libQtGui.dylib sipQtGuicmodule.o sipQtGuiQAbstractOpenGLFunctions.o sipQtGuiQAbstractTextDocumentLayout.o sipQtGuiQAbstractTextDocumentLayoutPaintContext.o sipQtGu
iQAbstractTextDocumentLayoutSelection.o sipQtGuiQActionEvent.o sipQtGuiQBackingStore.o sipQtGuiQBitmap.o sipQtGuiQBrush.o sipQtGuiQClipboard.o sipQtGuiQCloseEvent.o sipQtGuiQColor.o sipQtGuiQConicalGradient.o sipQtGuiQContextMe
[...]

PR #45145, opened 2015-10-19
pyqt5: work around a hard-coded 'd-bus' dependency

UniqMartin:
Fix for problem uncovered and discussed in #45114.


mikemcquaid:
:+1:

mikemcquaid:
Thanks @UniqMartin!

PR #45146, opened 2015-10-19
Bottle Required Language

DomT4:
CC @mikemcquaid @xu-cheng

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -172,11 +172,12 @@ class FormulaAuditor
       [/^  bottle do/,                     "bottle block"],
       [/^  devel do/,                      "devel block"],
       [/^  head do/,                       "head block"],
+      [/^  bottle (:unneeded|:disable)/,   "bottle modifier"],
       [/^  option/,                        "option"],
       [/^  depends_on/,                    "depends_on"],
       [/^  def install/,                   "install method"],
       [/^  def caveats/,                   "caveats method"],
-      [/^  test do/,                       "test block"]
+      [/^  test do/,                       "test block"],
     ]
 
     present = component_list.map do |regex, name|
@@ -192,6 +193,9 @@ class FormulaAuditor
     present.map!(&:last)
     if present.include?("head") && present.include?("head block")
       problem "Should not have both `head` and `head do`"
+    end
+    if present.include?("bottle modifier") && present.include?("bottle block")
+      problem "Should not have `bottle :unneeded/:disable` and `bottle do`"
     end
   end
 
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1592,26 +1592,34 @@ class Formula
     # @!attribute [w] bottle
     # Adds a {.bottle} {SoftwareSpec}.
     # This provides a pre-built binary package built by the Homebrew maintainers for you.
-    # It will be installed automatically if there is a binary package for your platform and you haven't passed or previously used any options on this formula.
+    # It will be installed automatically if there is a binary package for your platform
+    # and you haven't passed or previously used any options on this formula.
+    #
     # If you maintain your own repository, you can add your own bottle links.
     # https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Bottles.md
     # You can ignore this block entirely if submitting to Homebrew/Homebrew, It'll be
     # handled for you by the Brew Test Bot.
     #
     # <pre>bottle do
-    #   root_url "http://mikemcquaid.com" # Optional root to calculate bottle URLs
+    #   root_url "https://example.com" # Optional root to calculate bottle URLs
     #   prefix "/opt/homebrew" # Optional HOMEBREW_PREFIX in which the bottles were built.
     #   cellar "/opt/homebrew/Cellar" # Optional HOMEBREW_CELLAR in which the bottles were built.
     #   revision 1 # Making the old bottle outdated without bumping the version/revision of the formula.
-    #   sha256 "4355a46b19d348dc2f57c046f8ef63d4538ebb936000f3c9ee954a27460dd865" => :yosemite
-    #   sha256 "53c234e5e8472b6ac51c1ae1cab3fe06fad053beb8ebfd8977b010655bfdd3c3" => :mavericks
-    #   sha256 "1121cfccd5913f0a63fec40a6ffd44ea64f9dc135c66634ba001d10bcf4302a2" => :mountain_lion
+    #   sha256 "4355a46b19d348dc2f57c046f8ef63d4538ebb936000f3c9ee954a27460dd865" => :el_capitan
+    #   sha256 "53c234e5e8472b6ac51c1ae1cab3fe06fad053beb8ebfd8977b010655bfdd3c3" => :yosemite
+    #   sha256 "1121cfccd5913f0a63fec40a6ffd44ea64f9dc135c66634ba001d10bcf4302a2" => :mavericks
     # end</pre>
     #
-    # For formulae which don't require compiling, you can tag them with:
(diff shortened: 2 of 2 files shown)


DomT4:
I'll look at the `pull` situation in another PR. Decided to get the very simple stuff out the way first.

xu-cheng:
:+1:  Also for `audit`, you can add a check to prevent bottle block and bottle method coexisting, like what we do with head block and head filed.

Library/Homebrew/formula.rb
```diff
@@ -1599,19 +1599,22 @@ def mirror(val)
...
+    #   sha256 "4355a46b19d348dc2f57c046f8ef63d4538ebb936000f3c9ee954a27460dd865" => :el_capitan
+    #   sha256 "53c234e5e8472b6ac51c1ae1cab3fe06fad053beb8ebfd8977b010655bfdd3c3" => :yosemite
+    #   sha256 "1121cfccd5913f0a63fec40a6ffd44ea64f9dc135c66634ba001d10bcf4302a2" => :mavericks
     # end</pre>
     #
-    # For formulae which don't require compiling, you can tag them with:
+    # Only formulae where the upstream URL breaks or moves frequently, require compile or have a reasonable amount of patches/resources should be bottled.
+    # Formulae which do not meet the above requirements should not be bottled.
+    #
+    # Formulae which should not be bottled & can be installed without Xcode/CLT installation should be tagged with:
```
xu-cheng: May be explicitly specifying not compiling is required.

DomT4: I wasn't sure of the entire scope of not needing Xcode/CLT because I *think* you'd need at least the CLT to use the system `git` for patch applying?

mikemcquaid: It doesn't use `git` for applying but `patch`.

DomT4: Oh, I thought there was `git` in there somewhere, apologies :smile:. So without Xcode/CLT should just be changed to `does not require any compile` or such?

xu-cheng:
@DomT4 FYI, you can add `bottle do` and `bottle modifier` coexisting check [here](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/audit.rb#L192-L195)

Library/Homebrew/formula.rb
```diff
@@ -1599,19 +1599,22 @@ def mirror(val)
     # handled for you by the Brew Test Bot.
     #
     # <pre>bottle do
-    #   root_url "http://mikemcquaid.com" # Optional root to calculate bottle URLs
+    #   root_url "https://example.com" # Optional root to calculate bottle URLs
```
mikemcquaid: aw! :wink:

Library/Homebrew/formula.rb
mikemcquaid: Yeh :+1:

Library/Homebrew/formula.rb
```diff
@@ -1599,19 +1599,22 @@ def mirror(val)
...
-    #   sha256 "4355a46b19d348dc2f57c046f8ef63d4538ebb936000f3c9ee954a27460dd865" => :yosemite
-    #   sha256 "53c234e5e8472b6ac51c1ae1cab3fe06fad053beb8ebfd8977b010655bfdd3c3" => :mavericks
-    #   sha256 "1121cfccd5913f0a63fec40a6ffd44ea64f9dc135c66634ba001d10bcf4302a2" => :mountain_lion
+    #   sha256 "4355a46b19d348dc2f57c046f8ef63d4538ebb936000f3c9ee954a27460dd865" => :el_capitan
+    #   sha256 "53c234e5e8472b6ac51c1ae1cab3fe06fad053beb8ebfd8977b010655bfdd3c3" => :yosemite
+    #   sha256 "1121cfccd5913f0a63fec40a6ffd44ea64f9dc135c66634ba001d10bcf4302a2" => :mavericks
     # end</pre>
     #
-    # For formulae which don't require compiling, you can tag them with:
+    # Only formulae where the upstream URL breaks or moves frequently, require compile or have a reasonable amount of patches/resources should be bottled.
```
bfontaine: You might want to wrap it at 80 chars max.

Library/Homebrew/formula.rb
DomT4: Made the change locally last night :). Just wrapping up the audit discussion from last night's team chat where I was being stupid.

Library/Homebrew/formula.rb
DomT4: The length of [this line](https://github.com/DomT4/homebrew/blob/botdoc/Library/Homebrew/formula.rb#L1595) just above made me suspect we didn't bother on API comments, but I'm happy to if it makes readability a little easier.

Library/Homebrew/formula.rb
DomT4: You were replaced in the other bottle example we keep lying around the other week. Just mimicking that change, promise :wink:

DomT4:
> @DomT4 FYI, you can add bottle do and bottle modifier coexisting check here

Thanks! I completely scrolled past that and missed it about 20 times last night :smile_cat:.

DomT4:
* Tweaked the bottle explanation to wrap at ~80 characters.
* Added the audit check for the dual modifier/bottle block error.

This should be good to ship unless folks have wording tweaks/etc in mind.

Library/Homebrew/cmd/audit.rb
```diff
@@ -193,6 +194,9 @@ def audit_file
     if present.include?("head") && present.include?("head block")
       problem "Should not have both `head` and `head do`"
     end
+    if present.include?("bottle modifier") && present.include?("bottle block")
```
bfontaine: This is not related to *this* PR per se but we should probably use symbols instead of strings for the components, e.g. `:bottle_modifier`/`:test_block`/etc. It’d speed up the lookup in `present`. Or maybe use a `set`.

Library/Homebrew/cmd/audit.rb
```diff
@@ -193,6 +194,9 @@ def audit_file
...
+      problem "Should not have a bottle modifier & bottle block"
```
bfontaine: Let’s use “`and a bottle block`” instead of “`& bottle block`”.

Library/Homebrew/formula.rb
```diff
@@ -1592,26 +1592,34 @@ def mirror(val)
...
+    # Only formulae where the upstream URL breaks or moves frequently, require compile
+    # or have a reasonable amount of patches/resources should be bottled.
+    # Formulae which do not meet the above requirements should not be bottled.
+    #
+    # Formulae which should not be bottled & can be installed without any compile
+    # required should be tagged with:
     # <pre>bottle :unneeded</pre>
     #
-    # To disable bottle for other reasons.
+    # Otherwise formulae which do not meet the above requirements & should not be
```
bfontaine: Could we use “`and`” instead of “`&`” here and above? :smile:

Library/Homebrew/cmd/audit.rb
xu-cheng: Probably want to have consistent format with <code>Should not have both \`head\` and \`head do\`</code> in the above

DomT4: Done.

Library/Homebrew/formula.rb
DomT4: Done :wink:

Library/Homebrew/cmd/audit.rb
DomT4: I left this be for now, but I'm happy to go make changes in this PR if there's consensus on the path to take.

DomT4:
Merged this so we can move onto bigger changes around the bottle handling (like working on `pull`, etc) but if anyone reads over the language and decides it sounds weird, etc feel free to change it further.

Appreciate the review Xu, Baptiste & Mike :heart:

PR #45148, opened 2015-10-19
Elasticsearch 1.7.3

MerelyAPseudonym:
By the way, Elastic actually [provides a SHA1 checksum](https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.7.3.tar.gz.sha1.txt). So, if the policy is to prefer official checksums over our own, let me know and I’ll use theirs in the PR.

### DIFF
--- a/Library/Formula/elasticsearch.rb
+++ b/Library/Formula/elasticsearch.rb
@@ -1,8 +1,8 @@
 class Elasticsearch < Formula
   desc "Distributed real-time search & analytics engine for the cloud"
   homepage "https://www.elastic.co/products/elasticsearch"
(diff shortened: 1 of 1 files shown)


DomT4:
> So, if the policy is to prefer official checksums over our own, let me know and I’ll use theirs in the PR.

We use purely SHA256 these days and are actively trying to encourage upstreams to do the same thing.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/bb80a10b02c65bfdf0a2e873aa3dca4404def263. Thank you for another contribution to Homebrew @MerelyAPseudonym! :smiley_cat:

PR #45149, opened 2015-10-19
glslang 3.0 (new formula)

andystanton:
This formula adds the Khronos glsl validator.

Project URL: https://www.khronos.org/opengles/sdk/tools/Reference-Compiler/
Project sources: https://github.com/KhronosGroup/glslang

### DIFF
--- a/Library/Formula/glslang.rb
+++ b/Library/Formula/glslang.rb
@@ -1,0 +1,40 @@
+class Glslang < Formula
+  desc "glslang is an OpenGL and OpenGL ES shader front end and validator."
+  homepage "https://www.khronos.org/opengles/sdk/tools/Reference-Compiler/"
+  url "https://github.com/KhronosGroup/glslang/archive/3.0.tar.gz"
+  sha256 "91653d09a90440a0bc35aa490d0c44973501257577451d4c445b2df5e78d118c"
+
+  depends_on "cmake" => :build
+
+  def install
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args
+      system "make"
+      system "make", "install"
+
+      # Version 3.0 of glslang does not respect the overridden CMAKE_INSTALL_PREFIX. This has
+      # been fixed in master [1] so when that is released, the manual install commands should
+      # be removed.
+      #
+      # 1. https://github.com/KhronosGroup/glslang/commit/4cbf748b133aef3e2532b9970d7365304347117a
+      bin.install Dir["install/bin/*"]
+      lib.install Dir["install/lib/*"]
+    end
+  end
+
+  test do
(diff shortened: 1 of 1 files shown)


Library/Formula/glslang.rb
```diff
@@ -0,0 +1,21 @@
...
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args
+      system "make"
+      system "make", "install"
+      bin.install "StandAlone/glslangValidator"
+    end
+  end
+
+  test do
+    system "#{bin}/glslangValidator", "-v"
```
bfontaine: Can’t we write a test that does something more substantial than e.g. `-h`/`-v`?

Library/Formula/glslang.rb
```diff
@@ -0,0 +1,21 @@
...
+  sha256 "91653d09a90440a0bc35aa490d0c44973501257577451d4c445b2df5e78d118c"
+
+  depends_on "cmake" => :build
+
+  def install
+    mkdir "build" do
+      system "cmake", "..", *std_cmake_args
+      system "make"
+      system "make", "install"
+      bin.install "StandAlone/glslangValidator"
```
bfontaine: Isn’t this installed by the `make install` step?

andystanton: @bfontaine The ```CMAKE_INSTALL_PREFIX``` variable isn't overridden in this release of glslang, but has been fixed for future releases: https://github.com/KhronosGroup/glslang/commit/4cbf748b133aef3e2532b9970d7365304347117a

Library/Formula/glslang.rb
andystanton: @bfontaine I've added two sample files that are bundled with glslang and run glslangValidator against these as a more robust test that the application is working: https://github.com/andystanton/homebrew/commit/150b3e9e23ee40a628131d06928a4996e54c6c0a

I've put these in ```Library/Homebrew/test/fixtures``` and attempted to follow the ```test.*``` convention. Is this an acceptable place for test fixtures?

andystanton:
Thanks for the tip @bfontaine. I've made that change and also added a comment in the formula explaining why there is a manual install step, and that it should be removed in the next version.

Library/Formula/glslang.rb
DomT4: Since the test fixtures are tiny and BSD licensed let's just inline them directly into `test do` with `(testpath/"something.extension").write <<-EOS.undent` and so on. You can add a small attribution if you'd like to.

andystanton: @DomT4 done. As the fixtures are so simple I rewrote them from scratch.

mikemcquaid:

PR #45152, opened 2015-10-19
added configure option to build ODBC driver as well

aseidlitz:
current formula doesn't build ODBC driver. This patch adds config option to do it. Tested on El Captain

### DIFF
--- a/Library/Formula/mdbtools.rb
+++ b/Library/Formula/mdbtools.rb
@@ -17,7 +17,7 @@ class Mdbtools < Formula
   def install
     ENV.deparallelize
 
-    args = ["--prefix=#{prefix}"]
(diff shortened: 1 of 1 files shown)


Library/Formula/mdbtools.rb
```diff
@@ -17,7 +17,7 @@ class Mdbtools < Formula
   def install
     ENV.deparallelize
 
-    args = ["--prefix=#{prefix}"]
+    args = ["--prefix=#{prefix}","--with-unixodbc=/usr/local"]
```
bfontaine: Please add a space after the comma. If this `/usr/local` refers to Homebrew’s prefix please use `"--with-unixodbc=#{HOMEBREW_PREFIX}"` instead.

aseidlitz: Thanks, will do.

PR #45156, opened 2015-10-19
Update passenger

FooBarWidget:
Use 'passenger-config compile-nginx-engine' instead of the deprecated 'rake webhelper' task.


mikemcquaid:
Thanks @FooBarWidget!

PR #45160, opened 2015-10-19
crosstool-ng should set full path for gmake

ndimiduk:
crosstool-ng installs the script `bin/ct-ng` with a shbang, using exactly
what's provided to configure for `--with-make`. When `--with-make` is
provided to this formula, that results in a shbang line of

    #!gmake -rf

which is not not a valid interpreter

    $ ct-ng version
    -bash: /usr/local/bin/ct-ng: gmake: bad interpreter: No such file or directory

### DIFF
--- a/Library/Formula/crosstool-ng.rb
+++ b/Library/Formula/crosstool-ng.rb
@@ -42,7 +42,7 @@ class CrosstoolNg < Formula
 
     args << "--with-grep=ggrep" if build.with? "grep"
 
-    args << "--with-make=gmake" if build.with? "make"
+    args << "--with-make=#{Formula["make"].opt_bin}/gmake" if build.with? "make"
 
     args << "CFLAGS=-std=gnu89"
 


ndimiduk:
I'm not sure why the formula's test doesn't catch this error. Perhaps exit codes are not checked?

DomT4:
> I'm not sure why the formula's test doesn't catch this error. 

We don't check non-default options on the CI because on a lot of formulae it would *ridiculously increase* the amount of build and testing time required per PR.

Library/Formula/crosstool-ng.rb
```diff
@@ -42,7 +42,7 @@ def install
 
     args << "--with-grep=ggrep" if build.with? "grep"
 
-    args << "--with-make=gmake" if build.with? "make"
+    args << "--with-make=#{HOMEBREW_PREFIX}/bin/gmake" if build.with? "make"
```
DomT4: Rather than setting this hard on the `HOMEBREW_PREFIX` which will break if someone unlinks `gmake`, it may be worth using `#{Formula["make"].opt_bin}/gmake` instead.

ndimiduk: Sure, but the "make" formula comes from the dupes tap; does it need to be fully qualified, as in `Formula["homebrew/dupes/make"]` ?

DomT4: Shouldn't do. The line will only be "activated" if someone passes `--with-make`, during which `dupes` will be tapped if it isn't already and it'll be a known formula. The full path for the dep is only necessary because otherwise `brew` doesn't know what `make` formula you're asking it to go find.

ndimiduk:
Updated the formula, things work locally.

DomT4:
Thanks! Will pull when it has passed the CI tagged `default`.

mikemcquaid:

For future reference the preferred commit message format is `crosstool-ng: set full path for gmake`.

ndimiduk:
Thanks for the reviews and merge guys!

Issue #45140, opened 2015-10-19
Formulae found in multiple taps even after brew tap-pin


davydden:
subj. Happens for dependencies inside formulae. For example, if if my tap contains `formula1` which depends on `formula2` and the same two are there in other tap. Then `brew info frmula2` is fine (and will show it according to my version), whereas `brew info formula1` will give 
```
==> Dependencies
Error: Formulae found in multiple taps:
```

 Is it the expected behaviour or is it a bug?

davydden:
i see that 
```
Do note that pinned taps are prioritized only when the formula name is directly given by you. i.e., it will not influence formulae automatically installed as dependencies.
```
So the question is: if do `brew install formula2` this should take this formula from the prioritized tap. Can I still `brew install formula1` from my (prioritized) tap or will it fail the same way as `brew info` fail?

davydden:
to further specify the needs, i am trying to customise a couple of formulae from `homebrew-science` and still keep it for other formulae. That's why i need both tapped.

davydden:
how about this fix
```
diff --git a/Library/Homebrew/formulary.rb b/Library/Homebrew/formulary.rb
index 7577edb..f5f6a92 100644
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -287,6 +287,16 @@ class Formulary
     if possible_alias.file?
       return AliasLoader.new(possible_alias)
     end
+
+    # before looking for formula in Taps, check PinnedTaps:
+    possible_pinned_tap_formulae = tap_paths(ref, Dir["#{HOMEBREW_LIBRARY}/PinnedTaps/*/*/"]).map(&:realpath)
+    if possible_pinned_tap_formulae.size > 1
+      raise TapFormulaAmbiguityError.new(ref, possible_pinned_tap_formulae)
+    elsif possible_pinned_tap_formulae.size == 1
+      path = possible_pinned_tap_formulae.first.resolved_path
+      name = path.basename(".rb").to_s
+      return FormulaLoader.new(name, path)
+    end

     possible_tap_formulae = tap_paths(ref)
     if possible_tap_formulae.size > 1
```
?

xu-cheng:
This is a known issue which will eventually be resolved by https://github.com/Homebrew/homebrew/pull/43535. But since there are still lots work to do, it may take some time.

davydden:
[...]

PR #45162, opened 2015-10-19
search for a formula in PinnedTaps before trying Taps

davydden:
fixes https://github.com/Homebrew/homebrew/issues/45140


xu-cheng:
Sorry, but we deliberately don't want this. `brew tap-pin` is only used as a short cut when passing formula in ARGV. For dependencies resolution, formula should be determined by formula author instead of user's config. Some information you may find interesting: https://github.com/Homebrew/homebrew/pull/40991#issuecomment-116948161

PR #45164, opened 2015-10-19
Fix header file issue on Mountain Lion

mjago:
Fix header file issue on Mountain Lion (10.8.5). Same fix as for Snow Leopard worked for me.
https://bugzilla.gnome.org/show_bug.cgi?id=756136

### DIFF
--- a/Library/Formula/gstreamer.rb
+++ b/Library/Formula/gstreamer.rb
@@ -28,6 +28,7 @@ class Gstreamer < Formula
   # Snow Leopard)
   # https://bugzilla.gnome.org/show_bug.cgi?id=756136
   patch :DATA if MacOS.version <= :snow_leopard
+  patch :DATA if MacOS.version <= :mountain_lion
 
   def install
     args = %W[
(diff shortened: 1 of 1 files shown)


Library/Formula/gstreamer.rb
```diff
@@ -28,6 +28,7 @@ class Gstreamer < Formula
   # Snow Leopard)
   # https://bugzilla.gnome.org/show_bug.cgi?id=756136
   patch :DATA if MacOS.version <= :snow_leopard
+  patch :DATA if MacOS.version <= :mountain_lion
```
DomT4: You could just make the above line `patch :DATA if MacOS.version <= :mountain_lion`.

DomT4: i.e. the latter replaces the former, so it might as well actually replace the former.

mjago:
done

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/a29bfacbd190c404c8b3bd1fa5c40915227361b2. Thank you for your contribution to Homebrew @mjago; we appreciate it! :smiley_cat:

PR #45167, opened 2015-10-20
bitcoin-core 0.11.1 (new formula)

yumitsu:
Cross-platform bitcoin client.


DomT4:
Some prior discussion about Bitcoin in #28914. The Bitcoin team have also in the past expressed some hostility to their software being bottled, which it would be in the core. I'm not sure there's a huge upside here if there's been prior reluctance from other maintainers the past *and* the Bitcoin team has expressed reluctance. I'm open to thoughts from others on this though.

yumitsu:
@DomT4 I agree, but I still think we need to have a way to build it without additional tools (e.g. `bitcoin-qt`) for development purposes.

mistydemeo:
The [homebrew-bitcoin)(https://github.com/WyseNynja/homebrew-bitcoin) tap has a regularly maintained bitcoin formula which builds without Qt by default.

DomT4:
Unless there's a compelling reason not to, I'd rather leave it to them. They seem to have adapted a couple of my Bitcoin-related formulae so I have complete faith in them :stuck_out_tongue_winking_eye:.

If there's a real appetite from Bitcoin to see their stuff in Homebrew core I'd consider it but I don't see the upside of rubbing upstream the wrong way if there's a good tap available already. Closing for now, but per the above, the door isn't shut forever.

PR #45173, opened 2015-10-20
linkapps: stop linking .app bundles from 'bin/'

UniqMartin:
`Keg#app_installed?` only checks the formula prefix and `libexec/` for .app bundles to determine if a formula provides any. This is used by `Caveats#app_caveats` to generate an appropriate message.

For consistency, `brew linkapps` should use the same set of directories.

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -253,8 +253,13 @@ class Keg
     Dir["#{path}/lib/python2.7/site-packages/*.pth"].any?
   end
 
+  def apps
+    app_prefix = optlinked? ? opt_record : path
+    Pathname.glob("#{app_prefix}/{,libexec/}*.app")
+  end
+
   def app_installed?
-    Dir["#{path}/{,libexec/}*.app"].any?
+    !apps.empty?
   end
 
   def elisp_installed?
--- a/Library/Homebrew/cmd/linkapps.rb
+++ b/Library/Homebrew/cmd/linkapps.rb
@@ -23,11 +23,9 @@ module Homebrew
     end
 
     kegs.each do |keg|
-      keg = keg.opt_record if keg.optlinked?
-      Dir["#{keg}/*.app", "#{keg}/bin/*.app", "#{keg}/libexec/*.app"].each do |app|
+      keg.apps.each do |app|
         puts "Linking #{app} to #{target_dir}."
-        app_name = File.basename(app)
-        target = "#{target_dir}/#{app_name}"
+        target = "#{target_dir}/#{app.basename}"
 
         if File.exist?(target) && !File.symlink?(target)
           onoe "#{target} already exists, skipping."


mikemcquaid:
> For consistency, brew linkapps should use the same set of directories.

Better yet: unify the two uses into a single function.

UniqMartin:
> Better yet: unify the two uses into a single function.

I tend to agree, but I feel that's easier said than done. `Keg#app_installed?` only ever has to check the keg directory, but `brew linkapps` also goes to the trouble of checking if the formula is linked in `opt`, and if that's the case, makes this the destination of the symlinks.

Regardless, maybe something like this in `Keg` could work:

```ruby
def app_list(args = {})
  app_prefix = (args[:prefer_opt] && optlinked?) ? opt_record : path
  Dir["#{app_prefix}/{,libexec/}*.app"]
end

def app_installed?
  !app_list.empty?
end
```

And `linkapps` would use `keg.app_list(:prefer_opt => true)`. What do you think?

mikemcquaid:
@UniqMartin Works for me.

UniqMartin:
Great. :smile: Just pushed this variant.

Library/Homebrew/keg.rb
```diff
@@ -253,8 +253,13 @@ def python_pth_files_installed?
     Dir["#{path}/lib/python2.7/site-packages/*.pth"].any?
   end
 
+  def app_list(args = {})
```
xu-cheng: I would call it `apps`. :winked:

Library/Homebrew/keg.rb
```diff
@@ -253,8 +253,13 @@ def python_pth_files_installed?
...
+    app_prefix = (args[:prefer_opt] && optlinked?) ? opt_record : path
```
xu-cheng: Why do you need this args. I think we would always prefer opt rights?

Library/Homebrew/keg.rb
UniqMartin: Alright. Works for me.

Library/Homebrew/keg.rb
UniqMartin: Good question. I guess I wanted to stay as faithful as possible to the original implementation. But in reality `Keg#app_installed?` does not care which paths it is looking at. It makes sense to get rid of the argument and `args[:prefer_opt] &&` entirely.

Any opinion on `Dir[]` vs. `Pathname.glob()`? The latter seems to be more popular nowadays.

xu-cheng: > Any opinion on Dir[] vs. Pathname.glob()? The latter seems to be more popular nowadays.

I would prefer `Pathname.glob` because we prefer to using `pathname` over `string` for files.

UniqMartin:
@xu-cheng Thanks for your feedback! Here's the next iteration.

xu-cheng:
Thanks @UniqMartin

Issue #45165, opened 2015-10-19
`brew uninstall` doesn't unlink apps


mrkishi:
Is this intentional? It didn't feel very intuitive. I might forget to `brew unlinkapps` again in the future.

mikemcquaid:
It's not intentional and should probably be changed.

UniqMartin:
Closing in favor of #46542.

PR #45174, opened 2015-10-20
unlinkapps: avoid deleting too many symlinks

UniqMartin:
Use `start_with?` to make sure the symlink actually points into one of the Homebrew directories (depending on given arguments). Previously, only a substring match was used, which would also remove a symlink to a hypothetical `/opt/unrelated/usr/local/opt/Unrelated.app`. Even if unlikely to occur, altering stuff unrelated to Homebrew is bad.

Furthermore, make sure to always use a trailing slash with directories. Otherwise, e.g., `brew unlinkapps qt` will unlink .app bundles of both `qt` and `qt5` if both are installed and `brew linkapps qt qt5` was issued before. (Please ignore that `qt` and `qt5` offer a conflicting set of .app bundles. This will have to be addressed elsewhere.)

### DIFF
--- a/Library/Homebrew/cmd/unlinkapps.rb
+++ b/Library/Homebrew/cmd/unlinkapps.rb
@@ -25,9 +25,9 @@ module Homebrew
 
   def should_unlink?(file)
     if ARGV.named.empty?
-      file.match(HOMEBREW_CELLAR) || file.match("#{HOMEBREW_PREFIX}/opt")
+      file.start_with?("#{HOMEBREW_CELLAR}/", "#{HOMEBREW_PREFIX}/opt/")
     else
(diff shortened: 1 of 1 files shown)


mikemcquaid:
> Use start_with? to make sure the symlink actually points into one of the Homebrew directories (depending on given arguments). Previously, only a substring match was used, which would also remove a symlink to a hypothetical /opt/unrelated/usr/local/opt/Unrelated.app. Even if unlikely to occur, altering stuff unrelated to Homebrew is bad.

:+1: 

If you fancied tackling https://github.com/Homebrew/homebrew/issues/45165 that'd be :heart: too.

mikemcquaid:
Thanks @UniqMartin!

UniqMartin:
I'll have a look at #45165. No promises though …

Those `linkapps` and `unlinkapps` fixes were never intended. They were born as a byproduct while working on something completely different. :wink:

PR #45175, opened 2015-10-20
linkapps/unlinkapps: improve bash completion

UniqMartin:


xu-cheng:
Is possible to improve zsh completion as well :smile: ?

UniqMartin:
> Is possible to improve zsh completion as well :smile: ?

I guess I did. :) I'm no Z shell guru, so I'm not fully confident that my changes are without a flaw …

xu-cheng:
Thanks @UniqMartin

PR #45178, opened 2015-10-20
icmake 7.23.02

bfontaine:
The GitHub repo has no stars/forks because it was recently imported here. This is a 13-years-old project with 29 releases and [Debian](https://packages.debian.org/search?keywords=icmake&searchon=names&suite=all&section=all), [Ubuntu](http://packages.ubuntu.com/search?keywords=icmake&searchon=names&suite=all&section=all), and [MacPorts](https://trac.macports.org/browser/trunk/dports/devel/icmake/Portfile) packages.

### DIFF
--- a/Library/Formula/icmake.rb
+++ b/Library/Formula/icmake.rb
@@ -1,0 +1,42 @@
+class Icmake < Formula
+  desc "Make utility using a C-line grammar"
+  homepage "https://fbb-git.github.io/icmake/"
+  url "https://github.com/fbb-git/icmake/archive/7.23.02.tar.gz"
+  sha256 "f030086d0630c2c1755519556b9cf88b8495bc8debf36b257e0c69d3b199c46b"
+
+  depends_on "gnu-sed" => :build
+
+  def install
+    ENV.prepend_path "PATH", Formula["gnu-sed"].libexec/"gnubin"
+
+    # override the existing file
+    (buildpath/"icmake/INSTALL.im").open("w") do |f|
+      f.write <<-EOS.undent
+        #define BINDIR      "#{bin}"
+        #define SKELDIR     "#{share}"
+        #define MANDIR      "#{man}"
+        // not a typo; the install script puts binaries under LIBDIR/
+        #define LIBDIR      "#{bin}"
+        #define CONFDIR     "#{etc}"
+        #define DOCDIR      "#{doc}"
+        #define DOCDOCDIR   "#{doc}"
+      EOS
+    end
+
+    cd "icmake" do
+      system "./icm_bootstrap", "/"
+      system "./icm_install", "all", "/"
+    end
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

bfontaine:

bfontaine:

DomT4:
Seems like the CI might be ignoring us on the requests.

DomT4:
@BrewTestBot test this please

DomT4:
Web UI restart did the job. It's in the queue now.

bfontaine:
Thanks @DomT4.

Library/Formula/icmake.rb
```diff
@@ -0,0 +1,42 @@
+class Icmake < Formula
+  desc "Make utility using a C-line grammar"
+  homepage "https://fbb-git.github.io/icmake/"
+  url "https://github.com/fbb-git/icmake/archive/7.23.02.tar.gz"
+  sha256 "f030086d0630c2c1755519556b9cf88b8495bc8debf36b257e0c69d3b199c46b"
+
+  depends_on "gnu-sed" => :build
```
apjanke: This might be a run time dependency, too. Parts of `icmake` shell out to `sed` when you run it, and they may use GNU-specific behavior. When fiddling with 7.22, I had to [repoint it at `gsed`](https://github.com/apjanke/homebrew-zsh/blob/ff79b989ec00717b8aa53cf21d94efd7a9d34ab3/icmake.rb#L24-L27) to get it to work for building `yodl`.

bfontaine: Fixed, thanks.

apjanke:
Cool.

In addition to making `gnu-sed` a full-time dependency, I think we'll need to either make additional patches to the code, or wrap all its commands in env scripts that pull in `gnu-sed`'s `libexec/gnubin` to the front of the `$PATH`. Internally, `icmake` shells out to plain `sed`. It may seem to run and succeed, but it may silently produce incorrect output due to differences between BSD and GNU `sed`. Aside from the command line switches and defaulting with respect to Extended regexes, there are some edge cases in which their interpretation of patterns seems to just differ. In particular, with multi-line patterns which have backslashes stacked up at line endings. (This is based on [my previous attempts](https://github.com/apjanke/homebrew-zsh/tree/ff79b989ec00717b8aa53cf21d94efd7a9d34ab3) to get icmake and yodl working on OS X.)

I'll take a stab at revising this and the pending `yodl` PR to address this. I'm personally invested in getting this working, because I'd like to do `zsh` development on OS X, and that requires `yodl` for building its documentation.

Wrapping the commands to use `gnu-sed` sounds like a cleaner approach to me. Thoughts?

bfontaine:
Yeah I think that’s a good solution.

PR #45179, opened 2015-10-20
update and add https to riverbankcomputing urls

vszakats:


DomT4:
Merged in 4 commits :smile:. Thanks again @vszakats!

PR #45187, opened 2015-10-20
Add libstatgrab formula

adamflott:


Library/Formula/libstatgrab.rb
```diff
@@ -0,0 +1,18 @@
+class Libstatgrab < Formula
+  desc "provides cross platform access to statistics about the system."
```
dunn: "Provides" should be capitalized, thanks!

bfontaine: Isn’t this “`cross-platform`” instead of “`cross platform`”?

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/e152b3aabdc2e8dabca4630be3cc111d8eef323b. Thank you for your contribution to Homebrew @adamflott; we appreciate it! :smiley_cat:.

Future nit: New formulae commit messages should be written as something like:
```
coolformula 1.1.2 (new formula)
```
Thanks!

PR #45194, opened 2015-10-20
ammonite-repl 0.4.8 (new formula)

andystanton:
This formula adds the Ammonite Scala REPL.

Project URL: https://lihaoyi.github.io/Ammonite/#Ammonite-REPL
Project sources: https://github.com/lihaoyi/Ammonite

### DIFF
--- a/Library/Formula/ammonite-repl.rb
+++ b/Library/Formula/ammonite-repl.rb
@@ -1,0 +1,20 @@
+class AmmoniteRepl < Formula
+  desc "Ammonite is a cleanroom re-implementation of the Scala REPL"
+  homepage "https://lihaoyi.github.io/Ammonite/#Ammonite-REPL"
+  url "https://github.com/lihaoyi/Ammonite/archive/0.4.8.tar.gz"
+  sha256 "5ba76b6dad8e4397e5163112fd6ae2c4a35ea52a0e612b6f8ac774e896d9aa87"
+
+  depends_on :java => "1.6+"
+  depends_on "sbt" => :build
+
+  def install
+    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{ENV["HOME"]}"
+    system "sbt", "repl/assembly"
+    bin.install "repl/target/scala-2.11/ammonite-repl-0.4.8-2.11.7" => "amm"
+  end
+
+  test do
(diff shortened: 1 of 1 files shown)


andystanton:
It looks like the build fails because sbt can't create its .sbt path in the user's home directory: https://travis-ci.org/Homebrew/homebrew/jobs/86488418#L338

Any tips on an alternative approach?

DomT4:
> It looks like the Travis build fails because sbt can't create its .sbt path in the user's home directory

The CI builds are sandboxed these days, in eventual preparation for rolling that out to users by default. You can put `ENV["_JAVA_OPTIONS"] = "-Duser.home=#{ENV["HOME"]}"` at the top of `def install`, which should resolve this.

andystanton:
That worked, thanks @DomT4

Library/Formula/ammonite-repl.rb
```diff
@@ -0,0 +1,20 @@
+class AmmoniteRepl < Formula
+  desc "Ammonite is a cleanroom re-implementation of the Scala REPL"
+  homepage "https://lihaoyi.github.io/Ammonite/#Ammonite-REPL"
+  url "https://github.com/lihaoyi/Ammonite/archive/0.4.8.tar.gz"
+  sha256 "5ba76b6dad8e4397e5163112fd6ae2c4a35ea52a0e612b6f8ac774e896d9aa87"
+
+  depends_on :java
```
bfontaine: Does it need a specific version?

andystanton: @bfontaine yes it does. Scala 2.11 requires Java 6.

andystanton:
Changed test so that it verifies output rather than just that the command could be executed:

```ruby
system "#{bin}/amm", "-c", 'println("hello world!")'
```

becomes

```ruby
assert_equal "hello world!", shell_output("#{bin}/amm -c 'print(\"hello world!\")'")
```

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/82e478a5b89f46f32bbd3b8a602dd78e4ba44cc7. Thank you for your contribution to Homebrew @andystanton; we appreciate it! :smiley_cat:

PR #45206, opened 2015-10-21
neo4j 2.2.6 + neo4j 2.3.0-RC1 (devel)

mindrones:
After this one I'll make a new PR with the new stable (2.3.0), I've done this one to keep the complete versions history.

### DIFF
--- a/Library/Formula/neo4j.rb
+++ b/Library/Formula/neo4j.rb
@@ -1,14 +1,14 @@
 class Neo4j < Formula
   desc "Robust (fully ACID) transactional property graph database"
   homepage "http://neo4j.com"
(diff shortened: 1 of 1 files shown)


bfontaine:
Can’t we directly get the 2.3.0? It’ll save time (both for maintainers and the CI) to have one PR instead of two.

mindrones:
But what if someone wants to use 2.2.6?

bfontaine:
@mindrones Homebrew doesn’t allows you to choose the version you want to install; it’s always the latest available. You might want to submit the 2.2.6 version to [Homebrew/versions](https://github.com/Homebrew/homebrew-versions) instead.

mindrones:
@bfontaine OK thanks, here it is #45208 (closing this one instead)

PR #45219, opened 2015-10-21
trafficserver: fix default paths

### LINKED ISSUE #45190: Traffic Server formula produces binaries that 'work' but have default paths hard coded into the bottle instead of the Homebrew path.
The default configuration of Traffic Server is heavily dependant on the hard coded paths in the binary. At the moment that's being set to the bottle path rather than the homebrew path, which means it looks for and tries to create files in the bottle. Obviously this doesn't work very well on an actual system.

The test clause in the formula doesn't spot this, because it breaks traffic_cop's monitoring, so it doesn't know that traffic_server has failed to start and incorrectly reports traffic_server as running.

DomT4:
Closes #45190.

### DIFF
--- a/Library/Formula/trafficserver.rb
+++ b/Library/Formula/trafficserver.rb
@@ -23,6 +23,7 @@ class Trafficserver < Formula
 
   depends_on "openssl"
   depends_on "pcre"
+
   if build.with? "spdy"
     depends_on "spdylay"
     depends_on "pkg-config" => :build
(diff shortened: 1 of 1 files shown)


Library/Formula/trafficserver.rb
```diff
@@ -39,17 +40,26 @@ def install
...
+    (var/"log/trafficserver").mkpath
+    (var/"trafficserver").mkpath
+
+    args = %W[
+      --prefix=#{prefix}
+      --mandir=#{man}
+      --localstatedir=#{var}
+      --sysconfdir=#{etc}
+      --with-openssl=#{Formula["openssl"].opt_prefix}
+      --with-user=#{ENV["USER"]}
```
xu-cheng: Is the user variable hard coded? we may want to set it during post_install for bottle.

DomT4: I did an `ack` through the installation directory and couldn't *find* any mentions of the user `brew` or (locally) `Dominyk` so I'm not entirely sure where this is getting pulled in. The only place `USER` turns up is here:

```
bin/trafficserver
178:	chown $USER:admin $TS_BASE/usr/local/Cellar/trafficserver/6.0.0/var/trafficserver
```

DomT4: (That's from the current bottles. The `var` path is obviously different after this PR :smile:)

PR #45226, opened 2015-10-21
gst-plugins-good: disable buggy osxvideosink

rhendric:


DomT4:
I suspect this will need applying to lower OS X versions as well, but will see if we get bug reports first. Merged in https://github.com/Homebrew/homebrew/commit/68b31513989399c5ea60977b9f04e11a9da18b37. Thank you once more @rhendric! :smiley_cat:

PR #45236, opened 2015-10-22
arm-none-eabi-gcc 20150921 (new formula)

nitsky:


mikemcquaid:
This is a better fit for a tap. Sorry!

bfontaine:
Here are the docs: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md

adzenith:
Tap already exists, if anybody comes here later: https://github.com/nitsky/homebrew-stm32

adzenith:
@mikemcquaid: was this rejected because it's too niche? This compiler can be used for any embedded ARM core, like some Arduinos, quadcopters, electronics, etc. I recommended to @nitsky that he try to get it merged after I spent a while trying to figure out which of the various implementations of this formula worked and were up to date.
Thanks!

mikemcquaid:
I just think a tap is a better fit for this. It doesn't look like a stable release, doesn't really have a homepage, doesn't pass `brew audit --strict --online` and it's unclear what I'd use it for (and how I'd know that).

adzenith:
Gotcha! All valid reasons. :) Thanks for your consideration!

kaidokert:
Good candidate for a cask : https://launchpad.net/gcc-arm-embedded

PR #45237, opened 2015-10-22
libcdr 0.1.1

luzpaz:
Changed the target compression from tar.bz2 to tar.xz to save 94kb. 
Also future changes will keep the tar.xz

### DIFF
--- a/Library/Formula/libcdr.rb
+++ b/Library/Formula/libcdr.rb
@@ -1,8 +1,8 @@
 class Libcdr < Formula
   desc "C++ library to parse the file format of CorelDRAW documents"
   homepage "https://wiki.documentfoundation.org/DLP/Libraries/libcdr"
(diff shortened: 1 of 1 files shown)


bfontaine:
> Changed the target compression from tar.bz2 to tar.xz to save 94kb.

Does it really matter since the formula is bottled?

mikemcquaid:
It also adds a dependency that's not provided by OS X. Please change it back, thanks.

luzpaz:
Other Document Liberation of the librevenge suite use .xz and therefor will already have installed xz dependency. There is a probability that people will also install some of these:
```brew cat libetonyek | grep .xz```
Same for libmspub, libvisio, libpagemaker,

mikemcquaid:
That's fine either way, just don't think the size factor makes a big difference and generally we prefer to avoid `xz` unless necessary. Thanks!

luzpaz:
Ok. No problem. Cancelling PR.

mikemcquaid:
Thanks!

PR #45242, opened 2015-10-22
dfmt 0.4.1

John-Colvin:
dfmt is a tool for formatting D source code. It's well maintained and in the long term is likely to become part of the main DMD distribution (At some point I will split all the constituent parts of `dmd.rb` out into separate packages, so don't worry, this formula won't become obsolete).

No, there isn't a `make install`.

### DIFF
--- a/Library/Formula/dfmt.rb
+++ b/Library/Formula/dfmt.rb
@@ -1,0 +1,34 @@
+class Dfmt < Formula
+  desc "Formatter for D source code"
+  homepage "https://github.com/Hackerpilot/dfmt"
+  url "https://github.com/Hackerpilot/dfmt.git",
+      :tag => "v0.4.1",
+      :revision => "5ce683bab21598db69e399768ecd143caa57ef34"
+  head "https://github.com/Hackerpilot/dfmt.git", :shallow => false
+
+  depends_on "dmd" => :build
+
+  def install
+    system "make"
+    bin.install "bin/dfmt"
+  end
+
+  test do
+    (testpath/"test.d").write <<-EOS.undent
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Looks good to me :+1:

John-Colvin:
The tester appears to have had a hiccough...

mikemcquaid:
@BrewTestBot test this please

Library/Formula/dfmt.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    void main()
+    {
+        writeln("Hello, world without explicit compilations");
+    }
+    EOS
+
+    system "#{bin}/dfmt", "-i", "test.d"
+
+    diff "test.d", "test_expected.d"
```
DomT4: What's the purpose of this line, out of interest? Just wondering since we're not checking for anything here, and `diff` as a command isn't likely to ever fail?

John-Colvin: I had assumed that the result of diff would be checked. Fixed now.

Library/Formula/dfmt.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    void main()
+    {
+        writeln("Hello, world without explicit compilations!");
+    }
+    EOS
+
+    system "#{bin}/dfmt", "-i", "test.d"
+
+    system "diff", "test.d", "test_expected.d"
```
bfontaine: Instead of writing an expected result then `diff`-ing it you can put it in a variable (e.g. `test_expected`) then use `assert_equal test_expected, (testpath/"test.d).read`.

bfontaine:
Thank you for your contribution to Homebrew @John-Colvin; we appreciate it! :tada:

PR #45244, opened 2015-10-22
poppler 0.37.0

edporras:
The following break after upgrade because of missing libpoppler.so. Included formula revision bump in commit to force rebuild:

- [x] diff-pdf
- [x] pdf2htmlex

As above, also fails so a reinstall is needed. Not sure how to include in commit. 
- [ ] pdftoipe (using --HEAD homebrew/head-only/pdftoipe)

Don't break after poppler upgrade and build-from-source suceeds:
- [x] pdf2svg
- [x] pdfgrep
- [x] xournal (using --with-poppler option)
- [x] inkscape
- [x] pdf-tools


DomT4:
> pdftoipe (using --HEAD homebrew/head-only/pdftoipe)

Can't do anything about `head-only`. People will need to rebuild manually.

Merged in https://github.com/Homebrew/homebrew/commit/0ea6c4381e8f49b45f3922c001dfed2ab62ecd6f. Thank you once more @edporras! :smiley_cat:

PR #45246, opened 2015-10-22
Passenger: fix conf message

CamJN:
The locations.ini file got moved, but the recommended apache conf still referenced the old location.


mikemcquaid:
CC @FooBarWidget for thoughts

FooBarWidget:
Yes that looks fine. Thanks for spotting this mistake.

mikemcquaid:

For future reference the preferred commit message format is `passenger: fix caveats.`.

PR #45247, opened 2015-10-22
Mesos 0.25.0

wendigo:
Looks it's working without futher changes :)


mikemcquaid:

For future reference the preferred commit message format for simple version updates is `mesos 0.25.0`.

PR #45249, opened 2015-10-22
psqlodbc: use shell_output, see discussion in #45002

Habbie:


mikemcquaid:
Thanks for this, @Habbie!

PR #45250, opened 2015-10-22
openssl: add 'enable-ssl-trace' option

dconnolly:
The enable-ssl-trace build flag allows the -trace option to be passed to
openssl s_client, which shows verbose, human-readable trace output of
protocol messages. https://www.openssl.org/docs/manmaster/apps/s_client.html

### DIFF
--- a/Library/Formula/openssl.rb
+++ b/Library/Formula/openssl.rb
@@ -16,6 +16,7 @@ class Openssl < Formula
 
   option :universal
   option "without-check", "Skip build-time tests (not recommended)"
+  option "with-trace", "Enable s_client support for the -trace option"
 
   depends_on "makedepend" => :build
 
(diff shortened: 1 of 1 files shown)


Library/Formula/openssl.rb
```diff
@@ -16,6 +16,7 @@ class Openssl < Formula
 
   option :universal
   option "without-check", "Skip build-time tests (not recommended)"
+  option "with-trace", "Build with enable-ssl-trace"
```
bfontaine: Would you mind using a more understandable text for people who are not familiar with the `./configure` options? Something like “Enable support for the -trace option on s_client” for example.

dconnolly: Sure thing. :+1:

dconnolly:
Updated option description, squashed.

dunn:
Since we're testing everything in a sandbox now, a bunch of dependents are failing: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/35262/#showFailuresLink

If you wanted to fix those up that would be _super_, otherwise we'll get to them when we can.

tdsmith:
AFAICT this isn't in the [1.0.2 branch](https://www.openssl.org/docs/man1.0.2/apps/s_client.html) but we could do this for 1.1.0. Let me know if I'm wrong!

PR #45252, opened 2015-10-22
ldc 0.16.0

John-Colvin:
We're finally free of that ugly beta-as-stable situation.


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/776514efca35375aac63369f379d910362fc05c6. Thanks @John-Colvin!

PR #45257, opened 2015-10-22
os: centralize check for pre-release OS X versions

UniqMartin:
Provide `OS::Mac.prerelease?` for pre-release checks and use it where appropriate. This should simplify updating the test once a new OS X release lands.

This also fixes a bug in `BuildError#dump`, where an empty warning message was printed on El Capitan after a failed from-source build, because the check there and the one in `check_for_unsupported_osx` were out of sync.

(I replaced `MacOS` with `OS::Mac` in the affected lines, because `OS::Mac` in `os/mac.rb` has this line:
```
::MacOS = self # compatibility
```
Not sure though, if that's true, as `MacOS` seems much more widely used.)

### DIFF
--- a/Library/Homebrew/exceptions.rb
+++ b/Library/Homebrew/exceptions.rb
@@ -251,7 +251,7 @@ class BuildError < RuntimeError
       puts issues.map { |i| "#{i["title"]} #{i["html_url"]}" }.join("\n")
     end
 
-    if MacOS.version >= "10.11"
+    if OS::Mac.prerelease?
       require "cmd/doctor"
       opoo Checks.new.check_for_unsupported_osx
     end
--- a/Library/Homebrew/os/mac/xcode.rb
+++ b/Library/Homebrew/os/mac/xcode.rb
@@ -18,7 +18,7 @@ module OS
         when "10.11" then "7.0.1"
         else
           # Default to newest known version of Xcode for unreleased OSX versions.
-          if MacOS.version > "10.11"
+          if OS::Mac.prerelease?
             "7.0.1"
           else
             raise "OS X '#{MacOS.version}' is invalid"
--- a/Library/Homebrew/cmd/doctor.rb
+++ b/Library/Homebrew/cmd/doctor.rb
@@ -240,7 +240,7 @@ class Checks
   end
 
   def check_for_unsupported_osx
-    if !ARGV.homebrew_developer? && MacOS.version >= "10.12" then <<-EOS.undent
+    if !ARGV.homebrew_developer? && OS::Mac.prerelease? then <<-EOS.undent
     You are using OS X #{MacOS.version}.
     We do not provide support for this pre-release version.
     You may encounter build failures or other breakages.
@@ -259,8 +259,7 @@ class Checks
       end
     end
 
-    # TODO: bump version when new OS is released
-    if MacOS.version >= "10.12"
+    if OS::Mac.prerelease?
       def check_xcode_up_to_date
         if MacOS::Xcode.installed? && MacOS::Xcode.outdated?
           <<-EOS.undent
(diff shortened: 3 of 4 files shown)


Library/Homebrew/exceptions.rb
```diff
@@ -251,7 +251,7 @@ def dump
       puts issues.map { |i| "#{i["title"]} #{i["html_url"]}" }.join("\n")
     end
 
-    if MacOS.version >= "10.11"
+    if OS::Mac.prerelease?
       require "cmd/doctor"
       opoo Checks.new.check_for_unsupported_osx
     end
```
xu-cheng: Maybe we should do this to avoid duplicated check.
```ruby
unsupported_osx = Checks.new.check_for_unsupported_osx
opoo unsupported_ox if unsupported_osx
```

xu-cheng:
Good catch :+1: 

> Not sure though, if that's true, as MacOS seems much more widely used.)

We used to use `MacOS` so it's used more widely. We probably need to change all of them to `OS::Mac` at least for core code.

Library/Homebrew/exceptions.rb
UniqMartin: You mean require `cmd/doctor` and run `Checks#check_for_unsupported_osx` unconditionally after a failed build? I guess I agree that the micro-optimization done here is not needed after a possibly long build and it's even cleaner to leave the check entirely to the `check_*` methods.

UniqMartin:
> We used to use MacOS so it's used more widely. We probably need to change all of them to OS::Mac at least for core code.

Would a PR that does just that be acceptable? It's an easy enough change, but like the automated code style update done not too long ago it would disturb the Git history quite a bit. And it would certainly touch a lot of files and break a few pending PRs.

xu-cheng:
> Would a PR that does just that be acceptable?

That would be great.

>   but like the automated code style update done not too long ago it would disturb the Git history quite a bit

It isn't too hard to git blame recursively  :smile:

Library/Homebrew/exceptions.rb
xu-cheng: Yep, run it unconditionally to avoid duplicated logic in checking.

Library/Homebrew/os/mac/xcode.rb
```diff
@@ -18,7 +18,7 @@ def latest_version
         when "10.11" then "7.0.1"
```
mikemcquaid: While you're in here these can be bumped to 7.1.0.

mikemcquaid:
Good idea :+1:, thanks!

Library/Homebrew/os/mac/xcode.rb
UniqMartin: Will do. :)

mikemcquaid:
Thanks @UniqMartin!

UniqMartin:
That was almost too quick. :dash: :wink:

I had no time to address @xu-cheng's [comment](https://github.com/Homebrew/homebrew/pull/45257#discussion_r42838867). Guess I'll do that in a separate PR then. :smiley:

xu-cheng:
> I had no time to address @xu-cheng's comment. Guess I'll do that in a separate PR then. 

I have took care of it in 96c0409c8605bb2f0724f591b8bfffb10c278121 :wink:

mikemcquaid:
> I had no time to address @xu-cheng's comment. Guess I'll do that in a separate PR then

Sorry, missed that!

mikemcquaid:
Thanks @xu-cheng!

UniqMartin:
It's addressed so that's good. :+1: FWIW, here's my version of @xu-cheng suggestion I was just about to push (it avoids one reference to the variable):
```ruby
require "cmd/doctor"
if (unsupported_osx_message = Checks.new.check_for_unsupported_osx)
  opoo unsupported_osx_message
end
```

UniqMartin:
> I have took care of it in 96c0409 :wink:

I'm soooo slow: :snail: :joy:

PR #45259, opened 2015-10-22
Allow apr-util to be built with additional libs

kabel:
These additional libs are necessary to allow HTTPD to be built with LDAP
and driver supported DBD.

### DIFF
--- a/Library/Formula/apr-util.rb
+++ b/Library/Formula/apr-util.rb
@@ -19,6 +19,11 @@ class AprUtil < Formula
   depends_on "apr"
   depends_on "openssl"
   depends_on "postgresql" => :optional
+  depends_on "mysql" => :optional
+  depends_on "freetds" => :optional
+  depends_on "unixodbc" => :optional
+  depends_on "sqlite" => :optional
+  depends_on "homebrew/dupes/openldap" => :optional
 
   def install
     ENV.universal_binary if build.universal?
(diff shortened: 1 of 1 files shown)


DomT4:
Have all these options been checked against some of the stuff Homebrew uses `apr` for when the Command Line Tools aren't installed? *(We fallback on non-CLT installations to using our `apr` package unless folks pour bottles for the impacted formulae due to the packaging nightmare that is Xcode/CLT Apr splitting via Apple)*.

kabel:
You'll have to help me out a little bit. I'm not familiar with the esoteric details of non-CLT installations. I figured that because all of these are optional build flags using them would automatically assume you could build from source. I have only built/tested locally.

DomT4:
The problem isn't with the `apr-util` package and non-CLT installations itself; the problem occurs when you don't install the Command Line Tools package and then try to install something like `subversion` which checks for `apr` things during configure and breaks sans-CLT thanks to an Apple packaging quirk.

Essentially my question is more or less if `apr-util` is built with some/all of these options on a non-CLT system does `brew install svn` still work.

kabel:
Good question. I'm not quite sure the best way to test that. How would somebody install `apr-util` with these options _without_ CTL?

DomT4:
Well, unless you have a non-CLT system to play with, the easiest way to check would be a relatively simple:

```
sudo mv /Library/Developer/CommandLineTools ~/Downloads
brew install apr-util --with-mysql --with-whatever --with-whatever
brew install subversion
```

And then afterwards to stick the CLT back where it should be:

```
sudo mv ~/Downloads/CommandLineTools /Library/Developer/
```

kabel:
Surprisingly, I do have a non-CLT system to play with. A quick test of just `--with-mysql --with-sqlite --with-openldap` on my El Capitan machine shows no problem after installing subversion from source/bottle.

DomT4:
Alright, will take it then. Merged in https://github.com/Homebrew/homebrew/commit/5a09a4467f9b4372480c81ec162f9705192e268a. Thanks for checking here, and thank you for the contribution to Homebrew :star2:

PR #45264, opened 2015-10-23
audit: warn about unknown bottle modifiers

bfontaine:
This prevents typos like `bottle :uneeded` or `bottle :disable`.

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -463,6 +463,12 @@ class FormulaAuditor
     end
   end
 
+  def audit_bottle_spec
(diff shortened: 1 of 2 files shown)


xu-cheng:
Instead of checking it line by line. We can do this:
```ruby
class BottleDisableReason
  SUPPORTED_TYPES = [:unneeded, :disabled]

  def valid?
    SUPPORTED_TYPES.include? @type
  end
end

# use it in audit
if formula.bottle_disabled? && !formula.bottle_disable_reason.valid?
  problem "Unrecognized bottle modifier"
end
```

:+1: on this feature

bfontaine:
Thanks @xu-cheng; I edited the PR.

mikemcquaid:
:+1:

PR #45269, opened 2015-10-23
moreutils: Don't conflict with parallel when building without it

amarshall:
Previously, the following would occur:

```
$ brew install parallel
$ brew install moreutils --without-parallel
==> Installing moreutils with --without-parallel
Error: Cannot install moreutils because conflicting formulae are installed.

  parallel: because Both install a 'parallel' executable.
```

This fixes the erroneous failure.

### DIFF
--- a/Library/Formula/moreutils.rb
+++ b/Library/Formula/moreutils.rb
@@ -19,7 +19,9 @@ class Moreutils < Formula
 
   depends_on "docbook-xsl" => :build
 
-  conflicts_with "parallel", :because => "Both install a 'parallel' executable."
+  if build.with? "parallel"
+    conflicts_with "parallel", :because => "Both install a 'parallel' executable."
+  end
   conflicts_with "task-spooler", :because => "Both install a 'ts' executable."
 
   resource "Time::Duration" do


DomT4:
We deliberately reverted this before because our conflict resolution isn't supposed to be this fine-grained yet and there were subtle issues around the failures it threw. Ref: https://github.com/Homebrew/homebrew/commit/55ee343824d0db27a950db170fc903a9566850b0

amarshall:
Actually, this doesn’t entirely work. Since now a `brew upgrade parallel` would fail because it has a conflict with moreutils as well. I’m not sure what the best way is to handle it on that side, since I’m unsure of how to get the flags `parallel` was built with. Though maybe it makes sense to just see if there is a `parallel` binary in `$(brew prefix)/bin`?

amarshall:
@DomT4 Yea… *sigh*

DomT4:
We've basically left the door open for someone willing to come along and fix the conflict resolution, but edge cases like this are well, extremely few, and that makes it a low priority on the list of things to work on long-term.

But yeah, you stumbled pretty much right onto the other side of this problem in the `upgrade` situation. You can tab for options used and use that as a conditional around each conflict, but it gets increasingly dancey.

DomT4:
Passing on this as a known issue. As said above we're more than happy to take a look at a PR if folks want to play with the conflict resolution itself. Thanks for the submission nonetheless!

PR #45270, opened 2015-10-23
El Cap Various

DomT4:
Closing o'bunch of issues.

### DIFF
--- a/Library/Formula/racket.rb
+++ b/Library/Formula/racket.rb
@@ -11,12 +11,23 @@ class Racket < Formula
     sha256 "d10ec37ab262c32ce23cbc253cd52a3b381e285a848b073048a5f593a9446b13" => :mountain_lion
   end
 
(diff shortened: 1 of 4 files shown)


mikemcquaid:
:+1:

DomT4:
`openh264` failed on Travis, which is interesting. Would guess either Travis doesn't come with the CLT installed, which probably isn't very likely, or the problem is actually `nasm` on El Cap as the build succeeded locally using a `nasm` I've had around since Yosemite.

DomT4:
Looks like a `nasm` bug. My `nasm` from Yosemite fixes the build & their current developmental release candidate fixes whatever the problem was; I'll ask how close we are to that being cut or whether trying to identify a fix and backporting it is worthwhile. I've pulled `openh264` out for now until I can post on `nasm`'s bug tracker tonight.

PR #45282, opened 2015-10-23
docker-machine: patch for virtualbox driver

rhendric:
Required for VirtualBox functionality on OS X <10.8; see comment.

### DIFF
--- a/Library/Formula/docker-machine.rb
+++ b/Library/Formula/docker-machine.rb
@@ -41,6 +41,18 @@ class DockerMachine < Formula
     url "https://github.com/pmezard/go-difflib.git", :revision => "f78a839676152fd9f4863704f5d516195c18fc14"
   end
 
+  # The latest release of VirtualBox requires Mountain Lion, but VirtualBox is
+  # still maintaining a 4.3 line which supports pre-Mountain Lion. However, the
+  # absence of a particular flag at VM creation time results, in 4.3, in a VM
+  # that can't boot. Upstream doesn't want to add this because they only
+  # officially support the latest VirtualBox release and this flag causes
+  # problems (I think?) in that line. But if you're on Lion or below, you can't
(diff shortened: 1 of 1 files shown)


DomT4:
I'm a wee bit reluctant to accept this given upstream closed your Issue there with a "wontfix" status at the moment. We have introduced backwards compatibility in a way upstream hasn't in a few formula, so I'll leave this open for other maintainers to ponder a bit, if they wish to.

mikemcquaid:
Agreed, sorry.

rhendric:
Wait, so where's the utility in turning this down? I understand upstream not wanting to introduce some kind of ad-hoc version checking for this given a general policy of ‘use the latest’—it adds a dimension to the problem that for them is already complex due to multiple OSes, and if and when they decide to tackle that they want to do it systematically. But Homebrew already thinks along this dimension, has a mechanism for addressing this class of problem; the PR uses the mechanism in just about the simplest possible way, and changes the status of the software for pre-10.8 users from ‘broken’ to ‘usable’—what disutility is present here that is greater than silently giving those users a broken build? (At the _very least_, there should be a warning printed that points to this thread!)

mikemcquaid:
In short: if you can't convince upstream of the merit of this change then that's the problem. Homebrew respecting upstream's decisions is unusual amongst package managers but a principle I'm proud of.

rhendric:
Respectfully, I'm not convinced that applying that principle in this case is worth it, but I won't pick a fight with you about your principles. :smiley: Would you accept a PR that issues a warning message on pre-10.8? I just don't want the next user to come across this to have to redo the work I did to diagnose it—there are, it seems, a _lot_ of potential reasons why `docker-machine` might just hang.

mikemcquaid:
A caveat or `depends_on` for the OS X version would be :+1:

PR #45287, opened 2015-10-24
Agda 2.4.2.4

mietek:
Updates Agda to 2.4.2.4.

cc @DomT4 @nijikon 

Initially, I also proposed the following:

> Changes the installation location of the Agda standard library from `prefix/agda-stdlib` to `prefix/lib`.

> Adds a dependency on GHC, in order to support compiling programs to native code using the MAlonzo backend.  Installing GHC can be disabled using the `--without-malonzo` option.

> Removes the `--with-malonzo-ffi` option.

> Ensures the tests using the MAlonzo backend are run only when GHC is installed.

> Ensures the test using the standard library's FFI bindings for the MAlonzo backend is run properly.

> Changes the dependency on Emacs from optional to recommended.

> Fixes compilation of the Emacs mode by patching the code to add a dependency on the `fontset` Emacs package.  The patch is intended to be removed once https://github.com/agda/agda/pull/1700 is merged.

> Adds a caveats section explaining how to set up the Emacs mode, and how to reference the standard library.

### DIFF
--- a/Library/Formula/agda.rb
+++ b/Library/Formula/agda.rb
@@ -5,10 +5,8 @@ class Agda < Formula
 
   desc "Dependently typed functional programming language"
   homepage "http://wiki.portal.chalmers.se/agda/"
-  url "https://hackage.haskell.org/package/Agda-2.4.2.3/Agda-2.4.2.3.tar.gz"
-  mirror "https://github.com/agda/agda/archive/2.4.2.3.tar.gz"
-  sha256 "bc6def45e32498f51863d67acfbe048c039d630c6a36761ed27e99a5f68d7b27"
-  revision 2
+  url "https://github.com/agda/agda/archive/2.4.2.4.tar.gz"
+  sha256 "0147f8a1395a69bee1e7a452682094e45c83126233f9864544b8a14f956ce8c3"
 
   bottle do
     sha256 "2dc343203159551613fa7346a93d3e2931064026dccab6c30b68a8490845643b" => :el_capitan


Library/Formula/agda.rb
```diff
@@ -20,10 +18,13 @@ class Agda < Formula
   head "https://github.com/agda/agda.git", :branch => "master"
 
   option "without-stdlib", "Don't install the Agda standard library"
-  option "with-malonzo-ffi",
-    "Include the MAlonzo backend's FFI (depends on the standard library)"
+  option "without-ghc", "Don't install GHC (needed for the MAlonzo backend)"
 
-  depends_on "ghc" => :build
+  if build.with? "ghc"
+    depends_on "ghc"
```
mikemcquaid: Just do `depends on "ghc" => [:build, :optional]` instead of the `if`/`else`.

Library/Formula/agda.rb
```diff
@@ -101,11 +104,13 @@ module simple-test where
       snoc [] x = x :: []
       snoc (x :: xs) y = x :: (snoc xs y)
     EOS
-    system bin/"agda", "-c", "--no-main", "--safe", test_file_path
     system bin/"agda", "--js", "--safe", test_file_path
+    if which("ghc")
```
mikemcquaid: Should instead be `if build.with? "ghc"`

Library/Formula/agda.rb
```diff
@@ -101,11 +104,13 @@ module simple-test where
...
+      system bin/"agda", "-c", "--no-main", "--safe", test_file_path
+    end
 
     # typecheck, compile, and run a program that uses the standard library
-    if build.with?("stdlib") && build.with?("malonzo-ffi")
+    if build.with? "stdlib" && which("ghc")
```
mikemcquaid: `&& build.with?("ghc")`

Library/Formula/agda.rb
mietek: By using `which`, I’m hoping to cover the case where the user has GHC installed using other methods than Homebrew.

Library/Formula/agda.rb
mietek: Same here.

Library/Formula/agda.rb
mietek: OK, thanks.

mietek: It appears your suggested change doesn’t have the same meaning. Reverting.

mietek:
The failures are due to Agda taking more than 10 minutes to build:
https://travis-ci.org/Homebrew/homebrew/jobs/87183441
https://travis-ci.org/Homebrew/homebrew/jobs/87183443

One build succeeded:
https://travis-ci.org/Homebrew/homebrew/jobs/87183442

Library/Formula/agda.rb
mikemcquaid: I'm not sure I understand when `ghc` is used at runtime?

Library/Formula/agda.rb
mikemcquaid: We don't want to support that case, thanks.

Library/Formula/agda.rb
mietek: The failures quoted above do not happen at runtime. In these failures, `cabal sandbox init` fails due to `ghc` not being available at _build_ time.

Library/Formula/agda.rb
mietek: OK, changed.

mietek:
@mikemcquaid: Clarified changes and commit message. Try again?

mietek:
It seems that the fourth test is only just now getting exercised… Let’s see if I can get it to pass.

Library/Formula/agda.rb
```diff
@@ -51,38 +56,81 @@ def install
...
+
+          Add the following option to Agda2 Program Args:
+            --ghc-flag=-i#{lib}/ffi
+          EOS
+        end
+      end
+    end
+
+    s
+  end
```
mikemcquaid: We don't want as many caveats as this. Strip it down to the bare minimum that is Homebrew-specific, thanks.

Library/Formula/agda.rb
```diff
@@ -20,14 +18,17 @@ class Agda < Formula
   head "https://github.com/agda/agda.git", :branch => "master"
 
   option "without-stdlib", "Don't install the Agda standard library"
-  option "with-malonzo-ffi",
-    "Include the MAlonzo backend's FFI (depends on the standard library)"
+  option "without-malonzo", "Don't install GHC (needed for the MAlonzo backend)"
```
mikemcquaid: Please don't change this option name or the flip the default.

Library/Formula/agda.rb
```diff
@@ -20,14 +18,17 @@ class Agda < Formula
...
+  if build.with? "malonzo"
+    depends_on "ghc"
+  else
+    depends_on "ghc" => :build
+  end
   depends_on "cabal-install" => :build
 
   depends_on "gmp"
-  depends_on "emacs" => :optional
+  depends_on "emacs" => :recommended
```
mikemcquaid: Please leave the default as it was.

Library/Formula/agda.rb
```diff
@@ -41,6 +42,10 @@ class Agda < Formula
     url "https://github.com/agda/agda-stdlib.git", :branch => "2.4.2.4"
   end
 
+  # fix compilation of the included Emacs mode
+  # to be removed once https://github.com/agda/agda/pull/1700 is merged
+  patch :DATA
```
mikemcquaid: Is this only needed when `build.with? "emacs"`?

Library/Formula/agda.rb
```diff
@@ -51,38 +56,81 @@ def install
     cabal_clean_lib
 
     if build.with? "stdlib"
-      resource("stdlib").stage prefix/"agda-stdlib"
+      resource("stdlib").stage prefix/"lib"
```
mikemcquaid: Is this not going to cause issues for other formulae/user installations that rely on the existing path?

mikemcquaid:
Please try to not keep force-pushing this until you've addressed all comments and confirmed it's working locally. Otherwise it puts real pressure on our CI system. Thanks!

Library/Formula/agda.rb
mietek: No. All that is conditional on `build.with? "emacs"` is running `agda-mode compile`. The user can run this command manually, having installed, say, a GUI version of Emacs, and the command will fail without the patch.

Library/Formula/agda.rb
mietek: Agda is a research language, with new releases introducing version changes. I’m not aware of any other formulae depending on Agda, and I don’t think there’s many user installations of Homebrewed Agda. Considering that prior to this patch, the Emacs mode didn’t work, the MAlonzo-based tests didn’t pass, and the formula wasn’t updated in the month following the [release of 2.4.2.4](http://wiki.portal.chalmers.se/agda/pmwiki.php).

Library/Formula/agda.rb
mietek: Why? The default was wrong, and cause the problem with the Emacs mode to go unnoticed for a long time, since it wasn’t exercised by CI.

The Agda developers assume [the user is working in Emacs](http://wiki.portal.chalmers.se/agda/pmwiki.php?n=Main.QuickGuideToEditingTypeCheckingAndCompilingAgdaCode):

> Agda programs are commonly edited using Emacs. To edit a module (assuming you have installed Agda and its Emacs mode properly), start Emacs and open a file ending in .agda. Programs are developed interactively, which means that one can type check code which is not yet complete: if a question mark (?) is used as a placeholder for an expression, and the buffer is then checked, Agda will replace the question mark with a “hole” which can be filled in later.

mietek:
Perhaps the improvements I proposed will be easier to understand as individual PRs.

Restricting the scope of this PR to update-only.

mietek:
@mikemcquaid: This PR is ready for review. The test failures are to be expected. I have described their causes in #45318, and I will provide a solution in a separate PR.

Library/Formula/agda.rb
mietek: This option does not work, and moreover, it _should_ not work, as I describe in #45318.

Library/Formula/agda.rb
mietek: OK.

mikemcquaid:
> @mikemcquaid: This PR is ready for review. The test failures are to be expected. I have described their causes in #45318, and I will provide a solution in a separate PR.

Could you consider removing the failing test for now? I prefer not to merge when it's not passing (as I'll get an email from Travis complaining).

mikemcquaid:
Let's merge https://github.com/Homebrew/homebrew/pull/45327 and this at the same time, thanks!

DomT4:
Was merged as part of https://github.com/Homebrew/homebrew/commit/48fdabed98e5a201ed26ab47a47364a50a6be33b.

PR #45294, opened 2015-10-24
Add formula for gdub

emanuelez:
This PR add a formula for gdub, a small script that makes it easier to work with gradle projects

### DIFF
--- a/Library/Formula/gdub.rb
+++ b/Library/Formula/gdub.rb
@@ -1,0 +1,20 @@
+class Gdub < Formula
+  desc "Gdub is a Gradle wrapper wrapper"
+  homepage "http://www.gdub.rocks"
+  url "https://github.com/dougborg/gdub/archive/v0.1.0.tar.gz"
+  sha256 "ddf2572cc67b8df3293b1707720c6ef09d6caf73227fa869a73b16239df959c3"
+
+  depends_on "gradle"
+
+  def install
+    bin.install "bin/gw"
+  end
+
+  test do
+    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{testpath}"
+    Dir.chdir(testpath)
+    system "gradle", "init"
+    Dir.chdir("#{testpath}/gradle")
+    system "gw", "tasks"
+  end
+end


Library/Formula/gdub.rb
```diff
@@ -0,0 +1,12 @@
+class Gdub < Formula
+  desc "Gdub is a Gradle wrapper wrapper"
+  homepage "http://gdub.rocks"
```
DomT4: This is a 404.

Library/Formula/gdub.rb
```diff
@@ -0,0 +1,12 @@
...
+  url "https://github.com/dougborg/gdub/archive/v0.1.0.tar.gz"
+  sha256 "ddf2572cc67b8df3293b1707720c6ef09d6caf73227fa869a73b16239df959c3"
+
+  depends_on "gradle"
+
+  def install
+    bin.install "bin/gw"
+  end
```
DomT4: Please add a test :). See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

emanuelez: Yeah I tried that but with no success. The purpose of this shell script is to be able to run the gradlew script from any folder of the project as opposed to being able to do it only from the root folder. I don't see a way to write a test without having to setup a whole gradle project. I tried with a `gradle init` and then use that, but the testing environment did not seem to like that.

mikemcquaid: We'll need a test, I'm afraid. Using `gradle init` sounds like a good approach.

DomT4: > I tried with a gradle init and then use that, but the testing environment did not seem to like that.

What was the printout on that?

emanuelez: from here: https://travis-ci.org/Homebrew/homebrew/jobs/87228679

```
==> gradle init
.
Last 150 lines from /Users/travis/Library/Logs/Homebrew/gdub/01.gradle:
2015-10-24 19:03:37 +0000
gradle
init
FAILURE: Build failed with an exception.
* What went wrong:
Could not create service of type ScriptPluginFactory using BuildScopeServices.createScriptObjectConfigurerFactory().
* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
```

DomT4: Might need `ENV["_JAVA_OPTIONS"] = "-Duser.home=#{testpath}"` at the top of `test do` given it's throwing out to `gradle` and may be trying to write places it shouldn't be.

mikemcquaid: @DomT4 I wonder if we should just set that by default?

DomT4: We can't yet, sadly, as much as I'd like us to. See #44567.

emanuelez: It looks like it worked

DomT4:
I tweaked the `test do` block a little bit for style and removing an unnecessary bit (You're already in the `testpath` when that stage is initialised). Otherwise, merged untouched in https://github.com/Homebrew/homebrew/commit/91cafddc250221b125fec3ebabc2e0c8a9a536ab.

Thank you for your contribution to Homebrew @emanuelez; we appreciate it! :smiley_cat:

PR #45299, opened 2015-10-24
macvim: Fix building on Yosemite with Xcode 7.1

chdiza:
This is a temporary fix until upstream decide on a fix.


chdiza:
See https://github.com/macvim-dev/macvim/issues/98

Without this, a MacVim built on Yosemite using Xcode 7.1 will produce an .app that can only be used on El Capitan(!).

Over on the MacVim repo, we are considering how to fix this issue for real.  In the meantime, this PR provides a workaround.

mikemcquaid:
Thanks @chdiza!

PR #45302, opened 2015-10-25
cat: raise unless object is formulae

### LINKED ISSUE #44630: Cat not working when package is not installed
To inspect what happens if I am to install a certain package, I tried to run `brew cat Caskroom/cask/retro-virtual-machine`, but I get the following error message:
```
Error: undefined method `path' for nil:NilClass
Please report this bug:
/usr/local/Library/Homebrew/cmd/cat.rb:9:in `cat'
/usr/local/Library/brew.rb:133:in `<main>'
```

Similarly, `brew desc <package>` returns nothing if the package is not installed yet.

I think both commands should at least produce some informative output (or maybe I'm doing something wrong: I just installed homebrew yesterday).

### LINKED ISSUE #45300: brew cat of cask gives error: undefined method path
pb3:~ jhawk$ brew cat Caskroom/cask/mysqlworkbench 
Error: undefined method `path' for nil:NilClass
Please report this bug:
/usr/local/Library/Homebrew/cmd/cat.rb:9:in `cat'
/usr/local/Library/brew.rb:146:in `<main>'

I imagine I should tap the cask first, but there should be a better error message.

DomT4:
Until a point in time when the Cask/Homebrew codebases are harmonised a little more it probably makes sense to refuse to `cat` Casks. Right now the Homebrew codebase is only minimally aware of what a Cask is.

Fixes #45300.
Fixes #44630.

### DIFF
--- a/Library/Homebrew/cmd/cat.rb
+++ b/Library/Homebrew/cmd/cat.rb
@@ -3,9 +3,11 @@ module Homebrew
     # do not "fix" this to support multiple arguments, the output would be
     # unparsable, if the user wants to cat multiple formula they can call
     # brew cat multiple times.
+    formulae = ARGV.formulae
+    raise FormulaUnspecifiedError if formulae.empty?
+    raise "`brew cat` doesn't support multiple arguments" if formulae.size > 1
 
-    raise FormulaUnspecifiedError if ARGV.named.empty?
     cd HOMEBREW_REPOSITORY
-    exec "cat", ARGV.formulae.first.path, *ARGV.options_only
+    exec "cat", formulae.first.path, *ARGV.options_only
   end
 end


mikemcquaid:
:+1:

Library/Homebrew/cmd/cat.rb
```diff
@@ -3,8 +3,9 @@ def cat
     # do not "fix" this to support multiple arguments, the output would be
     # unparsable, if the user wants to cat multiple formula they can call
     # brew cat multiple times.
+    formulae = ARGV.formulae
+    raise FormulaUnspecifiedError if formulae.empty?
 
-    raise FormulaUnspecifiedError if ARGV.named.empty?
     cd HOMEBREW_REPOSITORY
     exec "cat", ARGV.formulae.first.path, *ARGV.options_only
```
bfontaine: You could replace `ARGV.formulae` with `formulae` here now.

Library/Homebrew/cmd/cat.rb
```diff
@@ -3,8 +3,9 @@ def cat
     # do not "fix" this to support multiple arguments, the output would be
     # unparsable, if the user wants to cat multiple formula they can call
     # brew cat multiple times.
+    formulae = ARGV.formulae
+    raise FormulaUnspecifiedError if formulae.empty?
```
xu-cheng: Maybe it's a good idea to add `raise "brew cat doesn't support multiple arguments" if formulae.size > 1`

Library/Homebrew/cmd/cat.rb
DomT4: Whoops. Good spot.

Library/Homebrew/cmd/cat.rb
DomT4: Sure.

DomT4:
Pushed the suggested changes.

mikemcquaid:
:+1:

PR #45324, opened 2015-10-26
antlr: add classpath to antlr4 wrapper script

### LINKED ISSUE #45292: Antlr isn't ready to use after brew install
Logs: https://gist.github.com/anonymous/d2418a95612900f3315e

Should probably add a message saying to add the jar to your CLASSPATH (or do it automatically if you can)

dunn:
Fixes #45292.

### DIFF
--- a/Library/Formula/antlr.rb
+++ b/Library/Formula/antlr.rb
@@ -8,8 +8,13 @@ class Antlr < Formula
 
   def install
     prefix.install "antlr-#{version}-complete.jar"
-    bin.write_jar_script prefix/"antlr-#{version}-complete.jar", "antlr4"
-    (bin+"grun").write <<-EOS.undent
+
+    (bin/"antlr4").write <<-EOS.undent
+      #!/bin/bash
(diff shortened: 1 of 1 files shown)


xu-cheng:
I'm not sure that we should add this.

Looks to me that users have to set `classpath` anyway. If I removed `classpath` setting in the [test](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/antlr.rb#L31-L32), it will fail.

*P.S.* I don't use java much and have little knowledge on this matter.

mikemcquaid:
I think we probably should add it to the wrapper script.

dunn:
Should we have the wrapper script set `CLASSPATH` as an environment variable instead?

mikemcquaid:
> Should we have the wrapper script set CLASSPATH as an environment variable instead?

If it works :+1:

dunn:
@vermiculus do you foresee any issue using `CLASSPATH` instead?

vermiculus:
As far as I understand, they have the same effect.

dunn:
Oops, pushed to the wrong remote.

dunn:
Ugh, lazy mistake.  Fix in #45411.

PR #45325, opened 2015-10-26
commands: ensure internal commands are files

xu-cheng:
`Pathname#children(with_directory = false)` doesn't filter directories,
instead it returns path with basename.


mikemcquaid:
:+1:

Issue #45322, opened 2015-10-26
Agda: Broken --with-emacs option


mietek:
cc @mikemcquaid @DomT4 @nijikon

The Agda formula offers a broken `--with-emacs` option.

Agda is [intended to be used](http://wiki.portal.chalmers.se/agda/pmwiki.php?n=Main.QuickGuideToEditingTypeCheckingAndCompilingAgdaCode) via the included Emacs mode, but the Agda formula declares Emacs as an optional dependency.  This is the **wrong default**, and it has caused the following problems with the Emacs mode to go unnoticed for a long time, as the related code was **not run in CI**.

Agda includes two commands associated with the Emacs mode:

1. `agda-mode setup`, which adds the mode to the user’s `~/.emacs`; and
2. `agda-mode compile`, which uses Emacs to compile the mode’s source code.

The Agda formula attempts to run both commands.

Running the first command **does not work**, because it alters the the `.emacs` inside the temporary `.brew_home` directory.

Running the second command **does not work**, because there appears to be an upstream issue (https://github.com/agda/agda/issues/1699, [PR agda/agda#1700](https://github.com/agda/agda/pull/1700)):

```
...
In toplevel form:
/usr/local/Cellar/agda/2.4.2.4/share/x86_64-osx-ghc-7.10.2/Agda-2.4.2.4/emacs-mode/agda2-mode.el:1701:1:Error: the function `create-fontset-from-fontset-spec' is not known to be defined.
Unable to compile the following Emacs Lisp files:
  /usr/local/Cellar/agda/2.4.2.4/share/x86_64-osx-ghc-7.10.2/Agda-2.4.2.4/emacs-mode/agda2-mode.el
```

mietek:
Proposed solution: https://github.com/mietek/homebrew/commit/07dd8b596f1ad04cf3d4d0acebb2979bc5e2f85e

mietek:
Proposed solution: https://github.com/mietek/homebrew/commit/74cc401023c1d2f4d731d00e72602fc3f59f11ee (without revision number)

mietek:
Proposed solution: https://github.com/mietek/homebrew/commit/74cc401023c1d2f4d731d00e72602fc3f59f11ee

mietek:
[PR agda/agda#1700](https://github.com/agda/agda/pull/1700) has been merged.

Issue #45321, opened 2015-10-26
Agda: Broken --with-malonzo-ffi option


mietek:
cc @mikemcquaid @DomT4 @nijikon

The Agda formula offers a broken `--with-malonzo-ffi` option.

This option is [described](https://github.com/Homebrew/homebrew/blob/d79c83a831198679d94b0ecc15b1cdc1437280b1/Library/Formula/agda.rb#L24) to "Include the MAlonzo backend's FFI (depends on the standard library)".  However, the `ffi` Cabal package is included in the Agda standard library **whether this option is set or not**.

The code conditional on this option is [described](https://github.com/Homebrew/homebrew/blob/d79c83a831198679d94b0ecc15b1cdc1437280b1/Library/Formula/agda.rb#L66) to "install the standard library's FFI bindings for the MAlonzo backend".  However, the code **does not work**, as `cabal install --user` results in the FFI bindings being installed inside the temporary `.brew_home` directory.  This is one of the causes of Agda test failures (#45318).

mietek:
Proposed solution: https://github.com/mietek/homebrew/commit/396d5f42d18f674250ed993771be84afc19881f0

mietek:
Proposed solution: https://github.com/mietek/homebrew/commit/79a159299096a6ffde94b706184952c7d2180772 (with correct paths in caveats)

mikemcquaid:
Maybe we should just enable this by default?

mietek:
Proposed solution: https://github.com/mietek/homebrew/commit/79a159299096a6ffde94b706184952c7d2180772

mietek:
Updated solution: https://github.com/mietek/homebrew/commit/9338e465332be24cf0b042c1174625879087a9db

Issue #45320, opened 2015-10-26
Agda: Installation of unnecessary files


mietek:
cc @mikemcquaid @DomT4 @nijikon

The Agda formula installs over 200M of unnecessary files.

When installing the Agda standard library, as part of the default configuration, we get 231M:
```
$ cd /usr/local/Cellar/agda/2.4.2.4/agda-stdlib

$ du -sch .cabal bin dist
225M	.cabal
2.6M	bin
3.4M	dist
231M	total
```

Adding the `--with-malonzo-ffi` option, we get 243M:
```
$ du -sch .cabal bin dist ffi/dist
225M	.cabal
2.6M	bin
3.4M	dist
 12M	ffi/dist
243M	total
```

The Agda formula installs the standard library’s helper tools into `agda-stdlib/bin`.  This is **unnecessary**, as these tools are only used as part of the build process, in order to generate the standard library's bytecode, documentation, and vim highlighting files.

The `.cabal`, `dist`, and `ffi/dist` directories are also **unnecessary**, as they contain build products left over from the build process.

mietek:
The solution for #45321 gets rid of the `ffi/dist` directory.

Proposed solution for getting rid of the remaining directories: https://github.com/mietek/homebrew/commit/8278ffc988589d5fe3db846e2647385c9018a9ad

mikemcquaid:
Sounds good :+1:

DomT4:
FYI although I've dealt with bits of the `agda` formulae before I really don't use Haskell/Cabal and there's not a lot of knowledge I can impart on these discussions, alas.

mietek:
Same solution after rebase: https://github.com/mietek/homebrew/commit/60c6e0107ac2f8982d5e12dd14b7dcace9bcacb2

Issue #45318, opened 2015-10-25
Agda: Test failures


mietek:
cc @mikemcquaid @DomT4 @nijikon 

There are [four tests](https://github.com/Homebrew/homebrew/blob/d79c83a831198679d94b0ecc15b1cdc1437280b1/Library/Formula/agda.rb#L86-L123) for the Agda formula, two of which exhibit issues:

1. the Agda self-test, which passes in CI;
2. compiling `simple-test.agda` to native code using the MAlonzo backend, which **fails in CI**;
3. compiling `simple-test.agda` using the JS backend, which passes in CI; and
4. compiling `stdlib-test.agda` to native code using the MAlonzo backend, conditional on `build.with? "stdlib"` and `build.with? "malonzo-ffi"`, which is **not run in CI**, and **fails locally**.

### Problem 1.

```
...
Compiling simple-test in /private/tmp/agda20151024-15554-48b16p/simple-test.agdai to /private/tmp/agda20151024-15554-48b16p/MAlonzo/Code/QsimpleZ45Ztest.hs
Calling: ghc -O -Werror -i/private/tmp/agda20151024-15554-48b16p /private/tmp/agda20151024-15554-48b16p/MAlonzo/Code/QsimpleZ45Ztest.hs --make -fwarn-incomplete-patterns -fno-warn-overlapping-patterns
ghc: createProcess: runInteractiveProcess: exec: does not exist (No
such file or directory)
```
([Job #2664.2](https://travis-ci.org/Homebrew/homebrew/jobs/87148415#L843-L849))

Agda supports compiling programs using three backends: the default backend, the MAlonzo backend, and the JS backend.  All three backends are part of the Agda core and do not need to be installed separately.  However, in order for the MAlonzo backend to work, GHC must be available at runtime.

The Agda formula declares GHC as a build dependency only, which results in the MAlonzo backend not working — unless GHC is already installed.  This is the **cause of failure for test 2**, and the **_first_ cause of failure for test 4**.

### Problem 2.

```
...
Compiling stdlib-test in /private/tmp/agda20151024-15682-1wjoo39/stdlib-test.agdai to /private/tmp/agda20151024-15682-1wjoo39/MAlonzo/Code/QstdlibZ45Ztest.hs
[...]

PR #45327, opened 2015-10-26
Agda: Combo fix

mietek:
Fixes a number of issues with the Agda formula.  To be merged after #45287.

### Broken --with-emacs option (#45322)

Changes the dependency on Emacs from optional to recommended.

Fixes compilation of the Emacs mode by patching the code to add a dependency on the `fontset` Emacs package.  The patch is intended to be removed once agda/agda#1700 is merged.

### Broken --with-malonzo-ffi option (#45321)

Removes the `--with-malonzo-ffi` option, as the FFI bindings for the MAlonzo backend are part of the Agda standard library.

Improves the attempt to compile and install the FFI bindings by using a dedicated GHC package database, and cleaning up leftover build products.

Ensures the test using the FFI bindings references them properly.

Adds a caveats section explaining how to reference the standard library.

### Test failures (#45318)

Adds a dependency on GHC, in order to support compiling programs to native code using the MAlonzo backend.  Installing GHC can be disabled using the `--without-malonzo` option.

Ensures the tests using the MAlonzo backend are run only when GHC is installed.

Ensures the test using the standard library's FFI bindings for the MAlonzo backend is run properly.

### Installation of unnecessary files (#45320)

Installs the standard library's helper tools within the Cabal sandbox already used as part of the build process.

Improves the standard library build process by cleaning up leftover build products.

### Bonus

Satisfies `brew audit agda --strict --online`.

Fixes the standard library’s version and revision.

### DIFF
--- a/Library/Formula/agda.rb
+++ b/Library/Formula/agda.rb
@@ -15,7 +15,16 @@ class Agda < Formula
     sha256 "d0e17808ee896cdedc499b010b67961dc3db3c430c501faef16c4b4482238447" => :mountain_lion
   end
 
-  head "https://github.com/agda/agda.git", :branch => "master"
+  head do
+    url "https://github.com/agda/agda.git", :branch => "master"
+    resource "stdlib" do
+      url "https://github.com/agda/agda-stdlib.git", :branch => "master"
+    end
+  end
+
+  resource "stdlib" do
+    url "https://github.com/agda/agda-stdlib.git", :branch => "2.4.2.4"
+  end
 
   option "without-stdlib", "Don't install the Agda standard library"
   option "without-malonzo", "Disable the MAlonzo backend"
@@ -31,16 +40,6 @@ class Agda < Formula
   depends_on "emacs" => :recommended
 
   setup_ghc_compilers
-
-  head do
-    resource "stdlib" do
-      url "https://github.com/agda/agda-stdlib.git", :branch => "master"
-    end
-  end
-
-  resource "stdlib" do
-    url "https://github.com/agda/agda-stdlib.git", :branch => "2.4.2.4"
-  end
 
   # fix compilation of the included Emacs mode
   # to be removed once https://github.com/agda/agda/pull/1700 is merged
--- a/Library/Formula/agda.rb
+++ b/Library/Formula/agda.rb
@@ -25,7 +25,7 @@ class Agda < Formula
   depends_on "cabal-install" => :build
 
   depends_on "gmp"
-  depends_on "emacs" => :optional
+  depends_on "emacs" => :recommended
 
   setup_ghc_compilers
 
@@ -38,6 +38,10 @@ class Agda < Formula
   resource "stdlib" do
     url "https://github.com/agda/agda-stdlib.git", :branch => "2.4.2.4"
   end
+
+  # fix compilation of the included Emacs mode
+  # to be removed once https://github.com/agda/agda/pull/1700 is merged
+  patch :DATA
 
   def install
     # install Agda core
@@ -74,9 +78,8 @@ class Agda < Formula
       end
     end
 
-    # byte-compile and install Agda's included emacs mode
+    # compile the included Emacs mode
     if build.with? "emacs"
-      system bin/"agda-mode", "setup"
       system bin/"agda-mode", "compile"
     end
   end
@@ -119,3 +122,17 @@ class Agda < Formula
     end
   end
 end
+
+__END__
+diff --git a/src/data/emacs-mode/agda2-mode.el b/src/data/emacs-mode/agda2-mode.el
+index 04604ee..f6b3122 100644
+--- a/src/data/emacs-mode/agda2-mode.el
++++ b/src/data/emacs-mode/agda2-mode.el
+@@ -20,6 +20,7 @@ Note that the same version of the Agda executable must be used.")
+ (require 'time-date)
+ (require 'eri)
+ (require 'annotation)
++(require 'fontset)
+ (require 'agda-input)
+ (require 'agda2)
+ (require 'agda2-highlight)
--- a/Library/Formula/agda.rb
+++ b/Library/Formula/agda.rb
@@ -58,14 +58,14 @@ class Agda < Formula
     if build.with? "stdlib"
       resource("stdlib").stage prefix/"agda-stdlib"
 
-      # install the standard library's helper tools
+      # generate the standard library's bytecode
       cd prefix/"agda-stdlib" do
         cabal_sandbox do
           cabal_install "--only-dependencies"
-          cabal_install "--prefix=#{prefix/"agda-stdlib"}"
-          system prefix/"agda-stdlib"/"bin"/"GenerateEverything"
+          cabal_install
+          system "GenerateEverything"
         end
-        cabal_clean_lib
+        rm_rf [".cabal", "dist"]
       end
 
       # install the standard library's FFI bindings for the MAlonzo backend
--- a/Library/Formula/agda.rb
+++ b/Library/Formula/agda.rb
@@ -5,10 +5,8 @@ class Agda < Formula
 
   desc "Dependently typed functional programming language"
   homepage "http://wiki.portal.chalmers.se/agda/"
-  url "https://hackage.haskell.org/package/Agda-2.4.2.3/Agda-2.4.2.3.tar.gz"
-  mirror "https://github.com/agda/agda/archive/2.4.2.3.tar.gz"
-  sha256 "bc6def45e32498f51863d67acfbe048c039d630c6a36761ed27e99a5f68d7b27"
-  revision 2
+  url "https://github.com/agda/agda/archive/2.4.2.4.tar.gz"
+  sha256 "0147f8a1395a69bee1e7a452682094e45c83126233f9864544b8a14f956ce8c3"
 
   bottle do
     sha256 "2dc343203159551613fa7346a93d3e2931064026dccab6c30b68a8490845643b" => :el_capitan
--- a/Library/Formula/agda.rb
+++ b/Library/Formula/agda.rb
@@ -23,7 +23,9 @@ class Agda < Formula
   end
 
   resource "stdlib" do
-    url "https://github.com/agda/agda-stdlib.git", :branch => "2.4.2.4"
+    url "https://github.com/agda/agda-stdlib.git",
+        :tag => "v0.11",
+        :revision => "8602c29a7627eb001344cf50e6b74f880fb6bf18"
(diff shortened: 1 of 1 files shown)


Library/Formula/agda.rb
```diff
@@ -17,29 +15,38 @@ class Agda < Formula
...
-  option "with-malonzo-ffi",
-    "Include the MAlonzo backend's FFI (depends on the standard library)"
+  option "without-malonzo", "Disable the MAlonzo backend"
 
-  depends_on "ghc" => :build
+  if build.with? "malonzo"
+    depends_on "ghc"
+  else
+    depends_on "ghc" => :build
+  end
```
mikemcquaid: Maybe it's just worth removing this option altogether and always using `ghc`?

Library/Formula/agda.rb
```diff
@@ -51,38 +58,61 @@ def install
     cabal_clean_lib
 
     if build.with? "stdlib"
-      resource("stdlib").stage prefix/"agda-stdlib"
+      resource("stdlib").stage lib
```
mikemcquaid: Probably the reason we didn't use `lib` before is everything in there will be linked into `/usr/local/lib`. That's not to say we shouldn't change it but just worth considering.

Library/Formula/agda.rb
```diff
@@ -51,38 +58,61 @@ def install
...
+  def caveats
+    s = ""
+
+    if build.with? "emacs"
+      s += <<-EOS.undent
+
+      To add the Agda Emacs mode to your ~/.emacs file, run:
+        agda-mode setup
+      EOS
+    end
```
mikemcquaid: You can remove this as it's not Homebrew-specific.

Library/Formula/agda.rb
```diff
@@ -101,11 +131,13 @@ module simple-test where
       snoc [] x = x :: []
       snoc (x :: xs) y = x :: (snoc xs y)
     EOS
-    system bin/"agda", "-c", "--no-main", "--safe", test_file_path
+    if build.with? "malonzo"
+      system bin/"agda", "-c", "--no-main", "--safe", test_file_path
+    end
```
mikemcquaid: The `test do` can assume the default options. That said, if this is useful for your testing: :+1:

Library/Formula/agda.rb
mikemcquaid: I'm just tempted to remove options unless there's a really different use-case as they tend to bit-rot from lack of CI.

mikemcquaid:
Great explanations here :+1:

Library/Formula/agda.rb
mikemcquaid: Yeh, so this is the reason: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/35374/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_agda/

These files need to be deleted or moved somewhere other than `lib`.

mikemcquaid:
Test still seems to be failing, unfortunately: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/35374/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/test_agda/

Library/Formula/agda.rb
mietek: Because `--with-malonzo` is the default, we have CI exercising all the tests.  This option disables a well-defined piece of rarely-used functionality, and doesn’t introduce anything that isn’t being tested.

Library/Formula/agda.rb
mietek: While some of the source files could possibly be deleted, there is a fine line between what is a useful, and what isn’t. I don’t want to introduce a brittle dependency on a particular version of the standard library, which is still evolving.

I thought `skip_clean "lib"` would prevent this complaint. I must be misunderstanding its purpose. Can you clarify?

Library/Formula/agda.rb
mietek: I don’t understand your comment. I’ve left the contents of the tests as I’ve found them. Can you clarify?

Library/Formula/agda.rb
mietek: I would really rather leave this in.  As I understand, it _is_ Homebre policy not to make modifications to the contents of the user’s home directory; otherwise, this command would’ve been executed automatically, as indeed it was earlier part of the formula.  Once again, Agda is expected to be used via Emacs, as I describe in #45322.

Library/Formula/agda.rb
```diff
@@ -17,29 +15,38 @@ class Agda < Formula
...
-  head "https://github.com/agda/agda.git", :branch => "master"
+  head do
+    url "https://github.com/agda/agda.git", :branch => "master"
+    resource "stdlib" do
+      url "https://github.com/agda/agda-stdlib.git", :branch => "master"
+    end
+  end
+
+  resource "stdlib" do
+    url "https://github.com/agda/agda-stdlib.git", :branch => "2.4.2.4"
```
dunn: For stable resources, we'd like a `:revision` as well, so it can't change and break future builds.

mietek: I understand.  However, the Agda standard library is explicitly [described as experimental](https://github.com/agda/agda-stdlib/blob/master/CHANGELOG#L7-L8).  Today, the `2.4.2.4` branch has [received changes](https://github.com/agda/agda-stdlib/commits/2.4.2.4) backported from `master`.

mietek: We might get a stable version soon, though: https://github.com/agda/agda-stdlib/issues/74

mietek:
The last outstanding issue is, how to prevent `brew audit` from complaining that [“non-libraries were installed”](http://bot.brew.sh/job/Homebrew%20Pull%20Requests/35374/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_agda/).  This is currently causing a build failure, as @mikemcquaid [pointed out](https://github.com/Homebrew/homebrew/pull/45327#discussion-diff-42966168).

I believe that these files ought to be left alone.  Some of these files are generated during the standard library installation process and expected to be there.  Some of the other files are sources, which occupy hardly any space — the entire standard library is 46M, binaries included — and are useful to keep for reference.  Finally, the standard library is expected to change, and so selectively deleting its contents will be prone to breakage.

I thought adding [`skip_clean "lib"`](https://github.com/mietek/homebrew/commit/92fda971728eb61c438501d2c8fad5acdf49548f#diff-687c498e17fb0d2795d0c15293c1a837R50) would silence the complaint.  Evidently, it hasn’t helped.  Any hints?

Library/Formula/agda.rb
mikemcquaid: `skip_clean` stops files being deleted but does not stop them being linked. The only way to do that is to put them in a different path than `lib`.

Library/Formula/agda.rb
mikemcquaid: It's OK, you can ignore that.

Library/Formula/agda.rb
mikemcquaid: My point is: this would still need to be added by e.g. `apt-get` on Debian and they don't show messages so we don't need to either.

mikemcquaid: The more caveats we add the less any of them get read.

mikemcquaid:
> The last outstanding issue is, how to prevent brew audit from complaining that “non-libraries were installed”. This is currently causing a CI failure, as @mikemcquaid pointed out.

They can't be put in `lib` in that case. The options are delete them or use another directory. An alternative in some of these cases is install the entire application under `libexec` and then you can have `libexec/lib` which won't be linked and just created wrapper scripts or symlinks into `bin`.

Library/Formula/agda.rb
mietek: I see; thanks.

Library/Formula/agda.rb
mietek: OK, removing.

mietek:
Thanks for the suggestion.  Deleting the files deemed to be “non-libraries” is not an option, because some of them are required to be present for the standard library to work.

Leaving the standard library as it was, in `agda-stdlib`.

mietek:
All three jobs timed out.  Retrying.

mietek:
One of the tests is still failing, but in CI only, as I describe in https://github.com/Homebrew/homebrew/issues/45318#issuecomment-151575825.

Is `/usr/local` read-only in CI?

mietek:
Ping.

Library/Formula/agda.rb
mietek: …and we got it: [agda/agda-stdlib#v0.11](https://github.com/agda/agda-stdlib/tree/v0.11)

mikemcquaid:
> Is /usr/local read-only in CI during testing?

No. You may be able to reproduce the issue locally with `brew test --verbose --sandbox agda`

mietek:
Thanks.  I have now reproduced the issue:
```
Compilation error:
/usr/local/Cellar/agda/2.4.2.4/agda-stdlib/ffi/IO/FFI.hi: openBinaryFile: permission denied (Operation not permitted)

==> Sandbox log
Nov  1 02:49:18 sandboxd[303]: ghc(88786) deny file-write-create /usr/local/Cellar/agda/2.4.2.4/agda-stdlib/ffi/IO/FFI.hi
```
Why doesn’t Homebrew permit this operation? Is this by design?

Library/Formula/agda.rb
```diff
@@ -17,29 +15,37 @@ class Agda < Formula
...
+  if build.with? "malonzo"
+    depends_on "ghc"
+  else
+    depends_on "ghc" => :build
+  end
   depends_on "cabal-install" => :build
 
   depends_on "gmp"
-  depends_on "emacs" => :optional
+  depends_on "emacs" => :recommended
```
dunn: While we're at it, this can be switched to `depends_on :emacs`, and if there's a minimum version that can be specified too with
```ruby
depends_on "emacs" => ["24.1", :recommended]

mikemcquaid:
> Why doesn’t Homebrew permit this operation? Is this by design?

I'm not sure CC @xu-cheng for sandbox help.

Library/Formula/agda.rb
mietek: Thanks, but I don’t know if there is a minimum version.

mietek:
Finally, ladies and gentlemen, we are [green](https://travis-ci.org/Homebrew/homebrew/jobs/88603410).

mikemcquaid:
Thanks for your contribution to Homebrew, patience and hard work @mietek! Without people like you submitting PRs we couldn't run this project. You rock!

PR #45333, opened 2015-10-26
inspircd: use the 2.0 branch as head instead of master

SaberUK:
This updates the InspIRCd formula to use the 2.0 branch when `--HEAD` is specified.

The reason for this is that I will soon be merging in several changes into InspIRCd master (the development branch) which will (unintentionally) break the formula completely. Rather than have to file several updates to fix it I believe it is a more sensible move to switch it to a branch which is not receiving breaking changes.


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/db1a4e24863629063a3979b047bd3fed95f6eb6f. Thanks @SaberUK! :smiley_cat:

Issue #45310, opened 2015-10-25
brew create missing "name" option


hryamzik:
Brew create uses a first part of string splitted by a 'minus' sign, so this:

```bash
brew create https://github.com/fboender/ansible-cmdb.git
```

Returns an error:

```
Error: /usr/local/Library/Formula/ansible.rb already exists
```

Indeed it exists! But ansible-cmbd doesn't....

bfontaine:
You can use `brew create --set-name https://github.com/fboender/ansible-cmdb.git ansible-cmdb`.

dunn:
This isn't an issue with the name scanner, though?

bfontaine:
@dunn I guess so. The issue is that `Version.parse` assumes `cmdb` is the version and thus the `FormulaCreator` thinks the first part is the name: https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/create.rb#L83

dunn:
Maybe when there's a `.git` at the end we could get it to assume it's a `head` url?

bfontaine:
Yeah good idea :smile:

dunn:
:cool: I'm not too familiar with that code so I might not get to this right away.

bfontaine:
Will try something tomorrow.

hryamzik:
@bfontaine ```brew create https://github.com/fboender/ansible-cmdb.git --set-name ansible-cmdb``` did the trick, thanks!

This option is no mentioned nor in [FAQ](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/FAQ.md), nor in [Formula Cookbook](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md) nor in ```brew create --help``` though...

PR #45335, opened 2015-10-26
cookbook: mention --set-name for 'brew create'

bfontaine:
It’s mentioned in the manpage but not in the cookbook. See #45310.


DomT4:
:+1:

PR #45336, opened 2015-10-26
create: accept HEAD urls

### LINKED ISSUE #45310: brew create missing "name" option

bfontaine:
Fixes #45310.

### DIFF
--- a/Library/Homebrew/cmd/create.rb
+++ b/Library/Homebrew/cmd/create.rb
@@ -78,10 +78,16 @@ class FormulaCreator
     @url = url
     path = Pathname.new(url)
     if @name.nil?
-      %r{github.com/\S+/(\S+)/archive/}.match url
-      @name ||= $1
-      /(.*?)[-_.]?#{path.version}/.match path.basename
-      @name ||= $1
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1: although I do wonder if people necessarily mean `head` in that case or if they are just intending to provide the homepage?

bfontaine:
@mikemcquaid If it ends with `.git` it’s probably that they mean `head`; it takes more effort to copy `github.com/a/b.git` than `github.com/a/b`.

bfontaine:
We could pre-fill the `homepage` field though.

mikemcquaid:
> @mikemcquaid If it ends with .git it’s probably that they mean head; it takes more effort to copy github.com/a/b.git than github.com/a/b.

:+1:

PR #45342, opened 2015-10-26
Ignore QT dependencies for wireshark if not requested

nils-ohlmeier:
Wireshark configure automatically pulls in the QT dependency as long as you don't explicitly tell it to ignore the QT libs on the system. This removed the QT dependency if QT has not been request by one of the --with-qt options.

### DIFF
--- a/Library/Formula/wireshark.rb
+++ b/Library/Formula/wireshark.rb
@@ -71,8 +71,11 @@ class Wireshark < Formula
 
     args << "--disable-wireshark" if no_gui
     args << "--disable-gtktest" if build.without?("gtk+3") && build.without?("gtk+")
-    args << "--with-qt" if build.with?("qt") || build.with?("qt5")
-    args << "--with-qt=no" if build.without?("qt") && build.without?("qt5")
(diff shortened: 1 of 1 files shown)


Library/Formula/wireshark.rb
```diff
@@ -72,6 +72,7 @@ def install
     args << "--disable-wireshark" if no_gui
     args << "--disable-gtktest" if build.without?("gtk+3") && build.without?("gtk+")
     args << "--with-qt" if build.with?("qt") || build.with?("qt5")
```
DomT4: Can make this:

```ruby
if build.with?("qt") || build.with?("qt5")
  args << "--with-qt"
else
  args << "--with-qt=no"
end
```

nils-ohlmeier: Good point. Thank you. Updated accordingly.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/9d427101f7adf21b63af7afd41fef007b5ec20fa. Thank you for your contribution to Homebrew @nils-ohlmeier; we appreciate it! :smiley_cat:

PR #45344, opened 2015-10-26
ansible-cmdb 1.6 (new formula)

hryamzik:


dunn:
Looks like the makefile is trying to write directly to `/usr/local/lib` instead of to the Cellar: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/35392/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/install_ansible_cmdb/

Library/Formula/ansible-cmdb.rb
```diff
@@ -0,0 +1,24 @@
...
+  homepage "https://github.com/fboender/ansible-cmdb"
+  url "https://github.com/fboender/ansible-cmdb/releases/download/1.6/ansible-cmdb-1.6.zip"
+  version "1.6"
+  sha256 "a201207d1a8b8929f07bd28be0ca3f1224e9ecf296756b73d938b6a41bdde80e"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on "libyaml"
+  
+  def install
+    system "make", "install"
```
dunn: The solution is probably just to pass a `prefix` argument to make.

hryamzik:
@dunn so does it have to be patched?

dunn:
If the Makefile doesn't allow the install prefix to be customized, then yeah.

Library/Formula/ansible-cmdb.rb
```diff
@@ -0,0 +1,24 @@
+class AnsibleCmdb < Formula
+  desc "Ansible-cmdb takes the output of Ansible's fact gathering and converts it into a static HTML overview page containing system configuration information."
+  homepage "https://github.com/fboender/ansible-cmdb"
+  url "https://github.com/fboender/ansible-cmdb/releases/download/1.6/ansible-cmdb-1.6.zip"
+  version "1.6"
```
DomT4: Suspect this is redundant. We can handle reading simple URLs for version numbers.

hryamzik:
@DomT4 you mean version could be omitted or used as a variable?

hryamzik:
```
 * Non-libraries were installed to "/usr/local/Cellar/ansible-cmdb/1.6/lib"
Installing non-libraries to "lib" is discouraged.
The offending files are:
```

Ok, I'll ```inreplace``` this. Is there anything else that has to be fixed?

mikemcquaid:
@hryamzik `brew audit` lists a few other issues: https://travis-ci.org/Homebrew/homebrew/jobs/87652927#L382

hryamzik:
@mikemcquaid got it, thanks, now cleaning up locally.

```
ansible-cmdb:
 * Don't need to interpolate "prefix" with inreplace
 * Don't need to interpolate "bin" with inreplace
```

Any suggestions on this? I don't see any other way to replace full paths.

mikemcquaid:
@hryamzik That's telling you that `"#{prefix}"` is the same as `prefix`

hryamzik:
@mikemcquaid indeed, thanks!

Library/Formula/ansible-cmdb.rb
```diff
@@ -0,0 +1,26 @@
...
+  desc "Generates static HTML overview page from Ansible facts"
+  homepage "https://github.com/fboender/ansible-cmdb"
+  url "https://github.com/fboender/ansible-cmdb/releases/download/1.6/ansible-cmdb-1.6.zip"
+  sha256 "a201207d1a8b8929f07bd28be0ca3f1224e9ecf296756b73d938b6a41bdde80e"
+
+  depends_on :python if MacOS.version <= :snow_leopard
+  depends_on "libyaml"
+
+  def install
+    Dir.mkdir "#{bin}"
```
DomT4: Can just do `bin.mkpath` here.

Library/Formula/ansible-cmdb.rb
```diff
@@ -0,0 +1,26 @@
...
+    inreplace "Makefile", "/usr/local/lib/${PROG}", prefix
```
DomT4: If we're changing the same file multiple times please do:

```ruby
inreplace "Makefile" do |s|
  s.gsub! "/usr/local/lib/${PROG}", prefix
  s.gsub! "/usr/local/bin", bin
end
```
etc.

Library/Formula/ansible-cmdb.rb
```diff
@@ -0,0 +1,26 @@
...
+    inreplace "Makefile", "/usr/local/bin", bin
+    inreplace "ansible-cmdb", "(bin_dir, '..', 'lib', 'ansible-cmdb')", "(bin_dir, '..')"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"hosts").write( \
```
DomT4: Can you try to use `.write <<-EOS.undent` style syntax here? The `command(` multiple line syntax can cause big problems on older Ruby versions we still attempt to support.

hryamzik:
@DomT4 thanks, done.

apjanke:
Not supporting `PREFIX` in a Makefile sounds worthy of an upstream bug report; that's pretty basic.

Looks like the audit problems are all fixed now. Could you squash this to a single commit to avoid cluttering up Homebrew's git history?

Also, where did the "6.3" in the issue title come from? Looks like this program is v 1.6. (1.8 is available now, too.)

mikemcquaid:
> Not supporting PREFIX in a Makefile sounds worthy of an upstream bug report; that's pretty basic.

:+1:

hryamzik:
@apjanke, all fixed, I'll take a look at 1.8 a bit later and will make a new PR.

apjanke:
Good to go. Merged. Thanks for your contribution to Homebrew!

Issue #44628, opened 2015-10-05
Yap install failed with Brew version 0.9.5 on OSX 10.11


andrewcropper:
https://gist.github.com/cb4d18112768ac8bfc1b

Terminal output:

==> ./configure --enable-tabling --enable-depth-limit --enable-coroutining --ena
==> make
Last 15 lines from /Users/andrew/Library/Logs/Homebrew/yap/02.make:
./SolverTypes.h:122:20: note:   'Clause_new'
     friend Clause* Clause_new(const V& ps, bool learnt = false) {
                    ^
./Solver.C: In member function 'lbool Solver::search(int, int)':
./Solver.C:602:59: error: 'Clause_new' was not declared in this scope
                 Clause* c = Clause_new(learnt_clause, true);
                                                           ^
./Solver.C:602:59: note: suggested alternative:
In file included from ./Solver.h:29:0,
                 from ./Solver.C:20:
./SolverTypes.h:122:20: note:   'Clause_new'
     friend Clause* Clause_new(const V& ps, bool learnt = false) {
                    ^
make[1]: *** [Solver.o] Error 1
make: *** [all] Error 2

asparagui:
See also #43729, #41558.

DomT4:
Closing in favour of #41558.

Issue #41558, opened 2015-07-09
Yap failed to build on 10.10


bfontaine:
This is a follow-up to #35936. I don’t know if this is related to #35859.

```
$ brew install yap
==> Downloading http://www.dcc.fc.up.pt/~vsc/Yap/yap-6.2.2.tar.gz
Already downloaded: /Library/Caches/Homebrew/yap-6.2.2.tar.gz
==> ./configure --enable-tabling --enable-depth-limit --enable-coroutining --enable-threads --enable-pthread-locking --with-gmp=/usr/local/opt/gmp --with-readline=/usr/local/opt/
==> make
      _p_trie_usage in tries.o
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
make[1]: *** [tries.dylib] Error 1
make: *** [all] Error 2
```

==> [gist logs](https://gist.github.com/67cf3182037fb9e26088) <==

rdp:
10.10 gcc 5.2.0 I get

$ brew install yap
==> Downloading http://www.dcc.fc.up.pt/~vsc/Yap/yap-6.2.2.tar.gz
######################################################################## 100.0%
==> ./configure --enable-tabling --enable-depth-limit --enable-coroutining --enable-threads --enable-pthread-locking --with-gmp=/usr/local/opt/gmp --with-readline=/usr/local/opt/readlin

==> make
Last 15 lines from /Users/packrd/Library/Logs/Homebrew/yap/02.make:
      _p_trie_to_list in tries.o
  "_trie_traverse_cont", referenced from:
      _p_trie_traverse_cont in tries.o
  "_trie_traverse_first", referenced from:
      _p_trie_traverse_first in tries.o
  "_trie_traverse_init", referenced from:
      _p_trie_traverse_init in tries.o
  "_trie_traverse_next", referenced from:
      _p_trie_traverse_next in tries.o
  "_trie_usage", referenced from:
      _p_trie_usage in tries.o
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
make[1]: *** [tries.dylib] Error 1
make: *** [all] Error 2

apjanke:
Same error on 10.9.5 and 10.11.1 for me.

apjanke:
Looks like an incompatibility with our GCC 5? Surprising, since GCC seems to be their main target. Though maybe not, since their last release was Oct-28-2010.

[...]

PR #45357, opened 2015-10-26
yap: use secure urls

vszakats:


vszakats:
It seems this formula had already been broken (due to C compilation errors).

vszakats:
Breakage unrelated to present updates.

mikemcquaid:
@vszakats Would you mind trying to fix them?

vszakats:
@mikemcquaid Spent an hour on it, found out at least 3 different fatal build issues: 
- int size detection failure - in Jenkins results (OK locally)
- undeclared `Clause_new()` error - in local build and also #44628
- link failure - in Travis CI results and also #41558
- plus an immense amount of warnings

I'm afraid I won't be able to fix any of these, sorry :(

vszakats:
Another (different) problem report here, regarding Mountain Lion: https://sourceforge.net/p/yap/discussion/131077/thread/04971fbb/

DomT4:
I *suspect* this might build against GCC-4.9.

vszakats:
@DomT4 That seems like a safe bet. Is there any way to force that in the formula?

vszakats:
Closing this one due to the too many problems with this formula.

PR #45362, opened 2015-10-26
qtpass 1.0.3 (new formula)

annejan:
For people that prefer source builds over casks, unfortunately currently (always) links with Qt 4.8

### DIFF
--- a/Library/Formula/qtpass.rb
+++ b/Library/Formula/qtpass.rb
@@ -1,0 +1,21 @@
+class Qtpass < Formula
+  desc "GUI for pass, the standard unix password manager."
+  homepage "https://qtpass.org/"
+  url "https://github.com/IJHack/qtpass/archive/v1.0.3.tar.gz"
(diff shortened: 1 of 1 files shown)


annejan:
Interesting how it fails with `ld: library not found for -lcrt1.10.6.o`
Is that a common thing?

DomT4:
> Is that a common thing?

Nope. Can't recall seeing that one before. It should be in `/usr/lib`.

mikemcquaid:
What's the advantage over the Cask?

annejan:
Technically speaking there is only a disadvantage currently

annejan:
Technically there is only a disadvantage currently of it linking to Qt4.

I thought having a source distribution would be nicer, but I think I'll leave it for now..

Might revisit the idea at a later time, when I've fixed the `crt1` issue (and possibly Qt5 is more common, because it looks so much nicer).

For the time being I'll keep producing the binary installers with every release.

Feel free to close this pullrequest.

mikemcquaid:
The binary installers are preferable, I think, particularly as we are moving Homebrew and Cask closer. Sorry and thanks!

PR #45366, opened 2015-10-27
desk 0.1.2

cades:
This PR add a formula for desk, a lightweight workspace manager for the shell.

I know this project is very young, but since it's appear on [hacker news](https://news.ycombinator.com/item?id=10445461), it has [394 stars](https://github.com/jamesob/desk)!

### DIFF
--- a/Library/Formula/desk.rb
+++ b/Library/Formula/desk.rb
@@ -1,0 +1,15 @@
+class Desk < Formula
+  desc "Lightweight workspace manager for the shell"
+  homepage "https://github.com/jamesob/desk"
+  url "https://github.com/jamesob/desk/archive/v0.1.2.tar.gz"
+  sha256 "db6dfb80cbb012adf925a8002ce43e5e021b947917de4992bad06fc3cab16d23"
+
+  def install
+    bin.install "desk"
+  end
+
+  test do
+    system "#{bin}/desk", "help"
+    system "#{bin}/desk", "version"
+  end
+end


Library/Formula/desk.rb
```diff
@@ -0,0 +1,15 @@
...
+  url "https://github.com/jamesob/desk/archive/v0.1.2.tar.gz"
+  sha256 "db6dfb80cbb012adf925a8002ce43e5e021b947917de4992bad06fc3cab16d23"
+
+  def install
+    bin.install "desk"
+  end
+
+  test do
+    system "#{bin}/desk", "help"
+    system "#{bin}/desk", "version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikemcquaid:
Want to give this repo at least a few more days before we add it.

Library/Formula/desk.rb
DomT4: The `$HOME` env is set to the testpath here, so you might be able to do:

```ruby
test do
  system bin/"desk", "init"
  assert File.exist?(".desk")
end
```
Or such, perhaps.

mikemcquaid: :+1:

cades: I tried, but it hangs and timeout. seems to related to mktemp. Should I explicitly set HOMEBREW_TEMP in formula file?

mikemcquaid: Does it hang locally? `mktemp` shouldn't make a difference.

jamesob:
@cades may want to bump to `v0.2.0`: https://github.com/jamesob/desk/releases/tag/v0.2.0

jamesob:
May need to bump this to 0.3.1 :)

mikemcquaid:
@jamesob Want to try and open your own PR? Closing this due to lack of response.

PR #45371, opened 2015-10-27
qbs 1.4.3

jakepetroules:
Also add HEAD support.

### DIFF
--- a/Library/Formula/qbs.rb
+++ b/Library/Formula/qbs.rb
@@ -1,8 +1,9 @@
 class Qbs < Formula
   desc "Qt Build Suite"
   homepage "https://wiki.qt.io/Qt_Build_Suite"
-  url "https://download.qt.io/official_releases/qbs/1.4.2/qbs-src-1.4.2.tar.gz"
-  sha256 "b9d36118c3ae0f7d4df6bf7239a0a0163c0340b701d00191fa5f832cef341ce5"
+  head "https://code.qt.io/qt-labs/qbs.git"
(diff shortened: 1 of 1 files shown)


Library/Formula/qbs.rb
```diff
@@ -1,8 +1,10 @@
 class Qbs < Formula
   desc "Qt Build Suite"
   homepage "https://wiki.qt.io/Qt_Build_Suite"
-  url "https://download.qt.io/official_releases/qbs/1.4.2/qbs-src-1.4.2.tar.gz"
-  sha256 "b9d36118c3ae0f7d4df6bf7239a0a0163c0340b701d00191fa5f832cef341ce5"
+  version "1.4.3"
```
jakepetroules: So basically, according to the build bot, I can't specify `version` to reduce redundancy, because it's already the same as what's determined from the URL which is constructed FROM THIS?

How can I specify the version without typing it twice? This is annoying.

jakepetroules: So, this is bug #32540. I will work around it for now, I suppose...

DomT4: The version not being accepted inline in the URL is not a bug. It's an intentional choice we've made before.

We prefer full URLs for the sake of readability, making changes more obvious, etc. We pluck version out of the URL automatically and that can be used throughout the formula with the version syntax.

DomT4: Side-note: You're not really de-duplicating something if you replace two version numbers with one explicit version definition and two version variables. You've gone from two things to three to support the same thing.

PR #45373, opened 2015-10-27
update: use git stash silently.

mikemcquaid:
There's been a few issues where users have been confused about these errors. They may have modified stuff but we probably don't care about telling them that unless we're debugging other issues.


DomT4:
You beat me to doing something about this (It's on the list of things I've started but hadn't yet finished).

I started down the route of stashing unconditionally with no `pop` unless the `HOMEBREW_DEVELOPER` env was set, but this is less aggressive & probably likely to rub less people the wrong way.

PR #45375, opened 2015-10-27
leafnode: fix runtime issues

### LINKED ISSUE #45255: leafnode - no such user: news
I've moved over from macports to homebrew, and am having an issue with leafnode.

I've installed it, but it won't run, saying "no such user: news".

The install process on macports creates this user, but it seems brew does not, or have I missed something? How do I correctly do this if I'm supposed to do it manually?

(El Capitan, new install, have fully updated homebrew)

DomT4:
Closes #45255.


dunn:
<3

PR #45382, opened 2015-10-27
pyqt5 5.5.1

UniqMartin:
Also deprecate `--enable-debug` in favor of `--with-debug` to address a strict audit error.


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/3b6ebea3c8e90543b89f548121d03a8f6f664884. Thanks once more @UniqMartin! :star2:

PR #45388, opened 2015-10-27
brew install --auto added

bfontaine:
This option disables the `--HEAD` and `--devel` checks for HEAD-only and devel-only formulae. This allows one to run `brew install --auto foo bar qux` without worrying about `--HEAD` or `--devel`.

For example if `a` has a stable and HEAD block; `b` has only a HEAD block and `c` only have a `devel` block; one needs to do the following to install all of them:
```
$ brew install a
$ brew install --HEAD b
$ brew install --devel c
```

With this change one could use the following command:
```
$ brew install --auto a b c
```

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -43,23 +43,23 @@ module Homebrew
       FormulaInstaller.prevent_build_flags unless MacOS.has_apple_developer_tools?
 
       ARGV.formulae.each do |f|
-        # head-only without --HEAD is an error
-        if !ARGV.build_head? && f.stable.nil? && f.devel.nil?
+        # head-only without --HEAD nor --auto is an error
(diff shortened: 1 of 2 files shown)


xu-cheng:
Personally, I'm not too much favor on adding new flag, considering there are already too many.

In fact, if such feature is demanded, we can just remove these flag checks, everything else will work just the same.

DomT4:
IMO brew-bundle is pretty much perfect for this sort of thing, without us needing to stack up more core code. Not massively against but not sure bundle isn't already useful here?

bfontaine:
@DomT4 my use-case here was how to install the whole Emacs tap without having to give `--HEAD` to head-only formulae without giving it to not-head-only ones. I’m not sure how brew-bundle can help here.

If this is not a feature people would use I can still write a custom command in my personal tap.

mikemcquaid:
Agreed with @DomT4 and @xu-cheng.

bfontaine:
Thanks everyone for the feedback; I’ll use a custom command :smile:

PR #45389, opened 2015-10-27
added new stable relase (2.5.0) of taskwarrior

plattenschieber:
released on 2015/10/21

### DIFF
--- a/Library/Formula/task.rb
+++ b/Library/Formula/task.rb
@@ -3,8 +3,8 @@ class Task < Formula
   homepage "https://taskwarrior.org/"
 
   stable do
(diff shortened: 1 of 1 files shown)


mikemcquaid:
`install` and `audit` are failing: https://travis-ci.org/Homebrew/homebrew/jobs/87689571#L331. `devel` block needs removed and the `depends_on "gnutls"` added.

chdiza:
The ChangeLog indicates there's a way to disable the "feature" that requires `gnutls`.  Ideally our formula would have an option to thus disable it.

DomT4:
Looks like this version bump & @chdiza's proposed change were merged in other PRs. Closing here. Thanks for the submission to Homebrew!

PR #45396, opened 2015-10-27
Boot 2.4.0

MerelyAPseudonym:
N.B. in addition to updating the version of Boot, I added a caveat.

### DIFF
--- a/Library/Formula/boot-clj.rb
+++ b/Library/Formula/boot-clj.rb
@@ -13,6 +13,16 @@ class BootClj < Formula
     bin.install "boot.sh" => "boot"
   end
 
+  def caveats; <<-EOS.undent
+    Note that only the Boot wrapper was updated! To get the updated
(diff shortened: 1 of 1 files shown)


mikemcquaid:
You can squash out the `2.3.0` commit.

Library/Formula/boot-clj.rb
```diff
@@ -13,6 +13,16 @@ def install
     bin.install "boot.sh" => "boot"
   end
 
+  def caveats; <<-EOS.undent
+    Note that only the Boot wrapper was updated! To get the updated
+    Boot *library* (i.e., “the real Boot”), you must run:
+
+        boot -u
```
mikemcquaid: Where is the library stored?

DomT4:
Looks like we've merged in more recent version updates recently, so closing here. Thanks for the submission to Homebrew!

PR #45399, opened 2015-10-27
Formula for Carina CLI

rgbkrk:
Client for [Carina](https://getcarina.com/). This is my first formula.

One awkwardness here: the Makefile for `carina` set the version using `git describe --tags --dirty='-dev'`. When the homebrew build is going on, the repository ends up in a dirty state (dotfiles missing, not sure why). This formula gets the right tag directly, based on trusting that homebrew has checked out the right tag.

UPDATE: Looks like I'll be back in two weeks since the newer repo has only existed for 15 days. This CLI used to be a part of the go bindings before being broken out separately, so it's existed for longer but its not a big deal.

### DIFF
--- a/Library/Formula/carina.rb
+++ b/Library/Formula/carina.rb
@@ -1,0 +1,27 @@
+class Carina < Formula
+  desc "Work with Swarm clusters on Carina"
+  homepage "https://github.com/getcarina/carina"
+  url "https://github.com/getcarina/carina.git",
+        :tag => "v0.8.0",
+        :revision => "094a0d2cfb24e245015f980127b9492889ed86f9"
+  head "https://github.com/getcarina/carina.git"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    carinapath = buildpath/"src/github.com/getcarina/carina"
+    carinapath.install Dir["{*,.git}"]
+
+    cd carinapath do
+      system "make", "get-deps"
+      system "make", "carina", "VERSION=#{version}"
+      bin.install "carina"
+    end
+  end
+
+  test do
+    system "#{bin}/carina", "--version"
+  end
+end


Library/Formula/carina.rb
```diff
@@ -0,0 +1,30 @@
...
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+
+    carinapath = buildpath/"src/github.com/getcarina/carina"
+    carinapath.install Dir["{*,.git}"]
+
+    cd carinapath do
+      version = `git describe --tags 2> /dev/null`.strip
```
mikemcquaid: You can get rid of this variable; it's already provided by the formula.

Library/Formula/carina.rb
```diff
@@ -0,0 +1,30 @@
...
+      commit = `git rev-parse --verify HEAD`.strip
+
+      system "make", "get-deps"
+      system "make", "carina", "COMMIT=#{commit}", "VERSION=#{version}"
+      bin.install "carina"
+    end
+  end
+
+  test do
+    system "#{bin}/carina", "--version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/carina.rb
rgbkrk: Awesome, thanks! I dug too deep into `build` and other bits trying to get what was set for `tag` and `revision` with no luck.

Ended up making this much cleaner to have the commit detected in `carina`s Makefile as well.

Library/Formula/carina.rb
rgbkrk: This comes from looking at other formulas that are go binaries, including `docker`.

All the commands except for `--version`, `--help` and `--bash-completion` reach out to a remote service. Since this is a statically linked go binary, as long as it has been placed it's in a good spot.

I'm not completely sure what else I can test on the outside without hitting a running server.

mikemcquaid: > I'm not completely sure what else I can test on the outside without hitting a running server.

If you can't spin up servers locally or anything: no worries.

DomT4:
> UPDATE: Looks like I'll be back in two weeks since the newer repo has only existed for 15 days. This CLI used to be a part of the go bindings before being broken out separately, so it's existed for longer but its not a big deal.

Given this, I'm not too fussed about that particular requirement if everything else is passing and other maintainers are happy. There's a difference between a new project and an existing project being reorganised, IMO.

mikemcquaid:
> There's a difference between a new project and an existing project being reorganised, IMO.

Agreed :+1:

mikemcquaid:

For future reference the preferred commit message format for new formulae is `carina 0.8.0 (new formula)`.

rgbkrk:
You all rock! Thanks for the :shipit:!

> For future reference the preferred commit message format for new formulae is carina 0.8.0 (new formula).

Whoops! Next time I'll make sure to adhere to that.

Issue #45393, opened 2015-10-27
libgdiplus: Needed for complete mono installation on os x


haf:
In a 'Xamarin' installation, we have this:

``` bash
$ locate libgdiplus.dylib
/Applications/Xamarin Studio.app/Contents/Resources/lib/monodevelop/AddIns/Xamarin.Sketches/XamarinInteractiveAgentsMac.app/Contents/MonoBundle/libgdiplus.dylib
/Applications/Xamarin.iOS Build Host.app/Contents/MonoBundle/libgdiplus.dylib
/Library/Frameworks/Mono.framework/Versions/3.12.1/lib/libgdiplus.dylib
... etc
```

But not for the homebrew installation.

The mono recipe should depend on libgdiplus: https://github.com/mono/libgdiplus

mikemcquaid:
> The mono recipe should depend on libgdiplus: https://github.com/mono/libgdiplus

Could you try and open a pull request? This should help and I'm happy to walk you through anything else: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged

Thanks!

haf:
Yes, but I'm not sure I can pull it all the way (pun intended)...

haf:
```
brew audit libgdiplus
Error: Version value must be a string
Please report this bug:
    https://git.io/brew-troubleshooting
/usr/local/Library/Homebrew/version.rb:186:in `initialize'
/usr/local/Library/Homebrew/version.rb:176:in `new'
/usr/local/Library/Homebrew/version.rb:176:in `detect'
/usr/local/Library/Homebrew/cmd/audit.rb:966:in `audit_version'
/usr/local/Library/Homebrew/cmd/audit.rb:952:in `audit'
/usr/local/Library/Homebrew/cmd/audit.rb:515:in `block in audit_specs'
/usr/local/Library/Homebrew/cmd/audit.rb:512:in `each'
/usr/local/Library/Homebrew/cmd/audit.rb:512:in `audit_specs'
/usr/local/Library/Homebrew/cmd/audit.rb:904:in `audit'
/usr/local/Library/Homebrew/cmd/audit.rb:58:in `block in audit'
/usr/local/Library/Homebrew/cmd/audit.rb:56:in `each'
/usr/local/Library/Homebrew/cmd/audit.rb:56:in `audit'
[...]

PR #45404, opened 2015-10-27
first take at libgdiplus

haf:
See https://github.com/Homebrew/homebrew/issues/45393

Works with http://accord-framework.net/ for a number of tests... However; http://sourceforge.net/p/giflib/code/ci/ce02f85e6ed1e1455a4aea9ffa68a3e2b8f8a4b1/ and https://github.com/rainycape/magick/issues/21 seem to hit us. It's moved to line 117 here http://sourceforge.net/p/giflib/code/ci/ce02f85e6ed1e1455a4aea9ffa68a3e2b8f8a4b1/tree/util/rgb2gif.c

This might be a better solution for https://github.com/OpenRA/OpenRA/issues/6413

### DIFF
--- a/Library/Formula/libgdiplus.rb
+++ b/Library/Formula/libgdiplus.rb
@@ -1,16 +1,16 @@
 class Libgdiplus < Formula
   desc "C-based implementation of the GDI+ API for Mono"
   homepage "http://www.mono-project.com/docs/gui/libgdiplus/"
+  version "3.12"
   url "https://github.com/mono/libgdiplus/tarball/3b284ab28cb8737f9d14dabfedc6903655c66a7f"
   head "https://github.com/mono/libgdiplus.git"
   sha256 "c7b6e68f4f4ef62e1f7551769c7f0b87e7debd52311123a0264d23cf7ac9aee8"
-  version "3.12" # latest tag = 3.12, plus two patches.
+  patch :DATA
 
   # it depends on mono to link to System.Drawing
   depends_on "libexif"
   depends_on "glib"
   depends_on "cairo"
-  depends_on :x11 => :build
   depends_on "automake" => :build
   depends_on "libtool" => :build
   depends_on "autoconf" => :build
@@ -19,9 +19,10 @@ class Libgdiplus < Formula
   # in that recipe.
 
   def install
-    system "CPPFLAGS=-I/opt/X11/include ./autogen.sh"
+    system "./autogen.sh"
     system "./configure", "--disable-dependency-tracking",
                           "--disable-silent-rules",
+                          "--disable-tests",
                           "--prefix=#{prefix}"
     system "make", "install"
   end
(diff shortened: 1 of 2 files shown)


Library/Formula/libgdiplus.rb
```diff
@@ -0,0 +1,32 @@
+class Libgdiplus < Formula
+  desc "C-based implementation of the GDI+ API for Mono"
+  homepage "http://www.mono-project.com/docs/gui/libgdiplus/"
+  url "https://github.com/mono/libgdiplus/tarball/79ae6cdb90163ba5eb601f8691ea9b7558d25371"
+  head "https://github.com/mono/libgdiplus.git", using: :git, tag: "master"
+  sha256 "0b84f8dd36e4e6df0877d66c816e7ef19892b52a552b1eed18e039bfb52d4de1"
+  version "3.12" # latest tag = 3.12, plus two patches.
+
+  # it depends on mono to link to System.Drawing
+  depends_on :x11
```
mikemcquaid: We can't have this depend on X11, unfortunately; we can't assume users have it installed.

haf:
@mikemcquaid how about when compiling like now?

haf:
Another question: libgdiplus would prefer a pre 4.2 version of giflib – can I specify that somewhere?

mikemcquaid:
Unfortunately not; it needs to use the newest version of `glib` or vendor it as a `resource`.

haf:
Well, OK, I'd say this is more with regards to the libgdiplus project.

@directhex – could you use line the QuantizeBuffer function from http://sourceforge.net/p/giflib/code/ci/ce02f85e6ed1e1455a4aea9ffa68a3e2b8f8a4b1/ (a bit down in the `util/rgb2gif.c` file) instead of from the `lib/gif_lib.h` file?

haf:
@directhex – actually it seems that an existing PR solves this issue: https://github.com/mono/libgdiplus/pull/32/files#diff-7c18b451e587fd18f4ded5380d98b15dR855

Library/Formula/libgdiplus.rb
```diff
@@ -0,0 +1,32 @@
+class Libgdiplus < Formula
+  desc "C-based implementation of the GDI+ API for Mono"
+  homepage "http://www.mono-project.com/docs/gui/libgdiplus/"
+  url "https://github.com/mono/libgdiplus/tarball/79ae6cdb90163ba5eb601f8691ea9b7558d25371"
+  head "https://github.com/mono/libgdiplus.git", :using => :git, :tag => "master"
```
DomT4: Need nothing after `head "https://github.com/mono/libgdiplus.git"` here. Everything else is the assumed default.

Library/Formula/libgdiplus.rb
```diff
@@ -0,0 +1,32 @@
+class Libgdiplus < Formula
+  desc "C-based implementation of the GDI+ API for Mono"
+  homepage "http://www.mono-project.com/docs/gui/libgdiplus/"
+  url "https://github.com/mono/libgdiplus/tarball/79ae6cdb90163ba5eb601f8691ea9b7558d25371"
```
DomT4: Going to need a real release on this.

Library/Formula/libgdiplus.rb
```diff
@@ -0,0 +1,32 @@
...
+  head "https://github.com/mono/libgdiplus.git", :using => :git, :tag => "master"
+  sha256 "0b84f8dd36e4e6df0877d66c816e7ef19892b52a552b1eed18e039bfb52d4de1"
+  version "3.12" # latest tag = 3.12, plus two patches.
+
+  # it depends on mono to link to System.Drawing
+  depends_on "mono"
+  depends_on "libexif"
+  depends_on "glib"
+  depends_on "cairo"
+  depends_on :x11 => :build
```
DomT4: What about `x11` is it using at build that it wouldn't use at runtime?

Mailaender:
Yeah, we have to patch this in @openSUSE for months now. I upstreamed it, adapted it according to the code review. No idea why https://github.com/mono/libgdiplus/pull/32 hasn't been merged yet. Shouldn't be controversial.

Library/Formula/libgdiplus.rb
haf: There's no 'real release' to be had afaik.

Library/Formula/libgdiplus.rb
haf: To quote @mikemcquaid "We can't have this depend on X11, unfortunately; we can't assume users have it installed."

Definitely needed to build, I don't know if it's needed to run, but probably it's needed for some things and not for others as my testing has shown.

You tell me what to do.

haf:
Ready to merge? The build is failing on some internal error in the audit command and the PR discussed above was successfully merged.

Library/Formula/libgdiplus.rb
haf: Fixed.

Library/Formula/libgdiplus.rb
```diff
@@ -0,0 +1,32 @@
+class Libgdiplus < Formula
+  desc "C-based implementation of the GDI+ API for Mono"
+  homepage "http://www.mono-project.com/docs/gui/libgdiplus/"
+  url "https://github.com/mono/libgdiplus/tarball/3b284ab28cb8737f9d14dabfedc6903655c66a7f"
```
mikemcquaid: We'll need a stable release before we can add this to Homebrew.

Library/Formula/libgdiplus.rb
mikemcquaid: We should try to patch it so it's not needed at build-time or add it to `homebrew/x11`

Library/Formula/libgdiplus.rb
```diff
@@ -0,0 +1,32 @@
...
+  head "https://github.com/mono/libgdiplus.git"
+  sha256 "c7b6e68f4f4ef62e1f7551769c7f0b87e7debd52311123a0264d23cf7ac9aee8"
+  version "3.12" # latest tag = 3.12, plus two patches.
+
+  # it depends on mono to link to System.Drawing
+  depends_on "mono"
```
xu-cheng: I think instead of making this depend on mono, it should be installed as a recommended resource inside mono. Thus, it will have same behavior as upstream installer.

haf: Does that mean it gets installed *after* mono?

Library/Formula/libgdiplus.rb
haf: You mean you need a stable location to download the bytes from, because this code is both stable and released. How do you prefer it hosted and what precisely are your requirements to allow it to be included in homebrew?

Library/Formula/libgdiplus.rb
haf: Since this is a core part of mono, it should preferably be as easy to install as mono is. Since adding taps (or even finding them) isn't as easy, that probably excludes x11. On the other hand, even if it's slightly harder to install, unless you're doing the work for the patching (or someone else around here who wants to do it?), then it may be worth the extra hardship to install to add it to a x11 tap to get it up there. The rest can be solved with docs.

Library/Formula/libgdiplus.rb
mikemcquaid: We need a stable, tagged release.

Library/Formula/libgdiplus.rb
mikemcquaid: It would currently require it to manually be installed.

haf: That's a great idea. Let's do it like that. I remove this line and you update the mono recipe with the line to make it a resource?

Library/Formula/libgdiplus.rb
haf: So you need a git tag in the linked repo? Or you can take a tag in any cloned repo?

Library/Formula/libgdiplus.rb
haf: I can update the mono recipe if that helps to get this done quicker, btw.

Library/Formula/libgdiplus.rb
mikemcquaid: Our documentation is clear. 

> So you need a git tag in the linked repo?

Yes.




> Are you saying you need the official mono sponsor, Xamarin, to tag it?

Yes.

Library/Formula/libgdiplus.rb
mikemcquaid: @haf Yes, that would be preferable. You'd still need to address the X11 issue, though.

mikemcquaid:
> Ready to merge?

No, it needs to not rely on X11. Additionally (repeated above from @xu-cheng) I think it would be better if you could make this a resource on the Mono formula after it no longer relies on X11 and has an official upstream tag.

Library/Formula/libgdiplus.rb
haf: Thanks.

@directhex / Jo, Could you please tag the currently merged state in the git repository? Thank you. :)

haf:
It relies on X11, that's how the software is built, and I've already followed @xu-cheng 's advice and removed the dependency.

The software doesn't drop dependencies because you declare it must, Mike. Could you explain your thinking?

dunn:
We just don't allow software with hard X11 dependencies in the core repository. We have a [homebrew-x11](https://github.com/Homebrew/homebrew-x11) tap for that.

haf:
So let us merge to the tap instead then.

haf:
But when you say hard, what do you mean? 99%of the usage of this module is not dependent on x11, and works fine without it.

rakkesh:
@haf `depends_on :x11 => :build` is a hard dependency. If it was, `depends_on :x11 => :optional`, it would have been accepted in `Homebrew` core repository. 

As @dunn has mentioned, you can open a PR over at https://github.com/Homebrew/homebrew-x11/pulls.

haf:
It seems there are ifdefs removing the need to use X11:

`./src/graphics.c:#ifdef CAIRO_HAS_XLIB_SURFACE` and
`./src/gdiplus-private.h:	#ifdef CAIRO_HAS_XLIB_SURFACE`. However, there's an explicit check in ./configure: `configure: error: "Failed to compile with X11/Xlib.h include.  You must fix your compiler paths"`; can you avoid the check and the code?

rakkesh:
You could use a patch to test it but I am not sure if this going to be accepted unless you can justify.

Remove the lines for `X11` check from `configure.ac`, generate a diff and apply it in the formula.

haf:
If I do it it should be accepted, otherwise I just should not do it. I am time constrained. I was asked to create a PR for this lib, but the discussion here is not encouraging.

rakkesh:
I am sorry for digressing on this thread. Please follow @dunn's comment above, and make a PR in homebrew-x11 tap. That should be resolved quicker.

haf:
I'm just going to leave this here with you guys and check back next weekend. Feel free to do what you want with it, I have the code which covers my own use-case.

mikemcquaid:
> You could use a patch to test it but I am not sure if this going to be accepted unless you can justify.

The justification is: OS X does not ship with X11 and this library doesn't seem to need X11 at runtime.

haf:
So that means you'll a patch to the library that removes the check for x11.h?

haf:
So I've created a patch from https://github.com/mono/libgdiplus/pull/35.diff that seems to make it all compile fine without x11. However, the diff file doesn't work, and I haven't been google my way to a working diff file. The archaic binary talks about fuzz and chunks, but doesn't tell me what's wrong.

haf:
ping @mikemcquaid

dunn:
Inline patches are sometimes a bit fussy; try a `patch do` block as in https://github.com/Homebrew/homebrew/blob/master/Library/Formula/bash.rb#L16-L19

mikemcquaid:
A few audit failures here: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/35834/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_libgdiplus___strict___online/

With the patch either try a `patch do` block and if that doesn't work you'll probably need to manually patch the source as perhaps the upstream patch doesn't apply to this stable version.

bfontaine:
Ping?

haf:
I just don't know how to "manually patch the source"... I don't normally work with patches.

bfontaine:
Run `brew install -i --git libgdiplus`; edit the source to fix the issues; run `git diff | pbcopy`; exit the shell. In the formula, at its very end add `__END__` then paste the patch just after it. Add `patch :DATA` at the beginning of the formula to apply it. Tell me if that’s not clear/you need help.

bfontaine:
Anything new here?

haf:
It's slightly cyclic for us, what we work on. We're moving towards graphics again in a week or two and then we're going to finish this PR. Sorry about the time taken, but we have not forgotten it! =)

DomT4:
How's this going?

haf:
I had an intern try to do the patches like people here want but he didn't succeed either.

bfontaine:
Gentle ping :wink:

apjanke:

haf:
Instead of closing, how about you fix the patch? You seem to know how to do it.

PR #45410, opened 2015-10-28
Getting rid of weird newline characters in anttweakbar

hlin117:
For some reason, my homebrew was having a lot of trouble `brew up`ing because anttweakbar.rb had windows newline characters.

This is to get rid of those windows newline characters. If it seems that it's only been me who has been affected, it's okay to close this PR.

### DIFF
--- a/Library/Formula/anttweakbar.rb
+++ b/Library/Formula/anttweakbar.rb
@@ -1,5 +1,5 @@
 class Anttweakbar < Formula
-  desc "C/C++ library for adding GUIs to OpenGL apps"
+  desc "C/C++ library for adding GUIs 往 OpenGL apps"
   homepage "http://www.antisphere.com/Wiki/tools:anttweakbar"
   url "https://downloads.sourceforge.net/project/anttweakbar/AntTweakBar_116.zip"
   version "1.16"


Library/Formula/anttweakbar.rb
```diff
@@ -1,5 +1,5 @@
 class Anttweakbar < Formula
-  desc "C/C++ library for adding GUIs to OpenGL apps"
+  desc "C/C++ library for adding GUIs 往 OpenGL apps"
```
dunn: Looks like you've introduced an unexpected new character yourself!

hlin117: I don't even own a windows computer >=\ (And this is homebrew, for macs of course.) You sure they weren't there before?

hlin117: No, the audit bot found that the newline characters were introduced into the commit. I am so perplexed now.

Sorry for the PR.

hlin117: I'm seriously lost though. I performed `git reset --hard origin/master`, which added the newline characters.... So not sure what's happening here, honestly.

hlin117:
How can I have homebrew uninstall this ruby file? I don't know what package this ruby file is associated with.

dunn:
Oh, `Library/Formula/anttweakbar.rb` is part of Homebrew, I just meant you accidentally added that `往`.

PR #45441, opened 2015-10-29
ngrok 2.0.19

sheharyarn:
The `ngrok` formula hasn't been updated in a while. The author hasn't open-sourced the latest version, so instead of building it manually, now we have to download the pre-compiled binary and place it.

### DIFF
--- a/Library/Formula/ngrok.rb
+++ b/Library/Formula/ngrok.rb
@@ -1,25 +1,13 @@
 class Ngrok < Formula
-  desc "Expose localhost to the internet and capture traffic for replay"
   homepage "https://ngrok.com"
-  head "https://github.com/inconshreveable/ngrok.git"
-  url "https://github.com/inconshreveable/ngrok/archive/1.7.1.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
We're not going to include a release that isn't open-source.

dunn:
Yeah, we'll have to stay on the 1.x.x branch for now.  Is 2.x.x going to be open-sourced at some point?

sheharyarn:
I don't think that 2.x.x is going to be open sourced any time soon.

bfontaine:
Upstream’s [README](https://github.com/inconshreveable/ngrok#ngrok-20) states that “The code for ngrok 2.0 is not *yet* open source” (emphasis mine). One option would be to move this to binary-only (see https://github.com/Homebrew/homebrew-binary/pull/262).

DomT4:
Added some notations on updating in https://github.com/Homebrew/homebrew/commit/82354ac22ffc80e2951500001d210bf60ccb51b7.

I'm happy to keep the 1.x branch here but the proprietary release would need to be submitted elsewhere - Binary if we accept proprietary things there or the `caskroom/cask`.

PR #45454, opened 2015-10-29
libmatroska 1.4.4, remove patch linking libebml

jamesyc:
Patch in older formula was for 1.4.2, which was merged in 1.4.3. Upstream patch (Matroska-Org/libmatroska@9466bf5) should no longer be needed.

### DIFF
--- a/Library/Formula/libmatroska.rb
+++ b/Library/Formula/libmatroska.rb
@@ -3,13 +3,9 @@ class Libmatroska < Formula
   homepage "http://www.matroska.org/"
 
   stable do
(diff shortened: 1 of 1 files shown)


jamesyc:
Travis CI check is currently failing because libmatroska needs libebml 1.3.3 but the current version on homebrew is 1.3.1. 

... I submitted a pull request to update libebml to 1.3.3 a few minutes before this, so once that works, this should catch the right dependency and work too.

bfontaine:
Thank you for your contribution to Homebrew @jamesyc; we appreciate it! :tada:

bfontaine:
That was merged in 322ca8e6.

PR #45467, opened 2015-10-29
task: Add option --without-sync

chdiza:
This allows users who don't use the sync feature to avoid having
the dependency on gnutls.

### DIFF
--- a/Library/Formula/task.rb
+++ b/Library/Formula/task.rb
@@ -5,8 +5,6 @@ class Task < Formula
   stable do
     url "https://taskwarrior.org/download/task-2.5.0.tar.gz"
     sha256 "4d8e67415a6993108c11b8eeef99b76a991af11b22874adbb7ae367e09334636"
-
-    depends_on "gnutls"
   end
 
   bottle do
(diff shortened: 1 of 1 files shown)


chdiza:
I probably got the conditionality of the dep all awkward.  Feel free to tell me what the right thing is.

Library/Formula/task.rb
```diff
@@ -17,14 +15,17 @@ class Task < Formula
...
   head do
     url "https://git.tasktools.org/scm/tm/task.git", :branch => "2.5.1", :shallow => false
-
-    depends_on "gnutls"
   end
 
+  option "without-sync", "Disable sync support"
+
   depends_on "cmake" => :build
+  depends_on "gnutls" if !build.without? "sync"
```
bfontaine: use `if build.with? "sync"`.

mikemcquaid: Just do `:optional` here and make `without-gnutls` note that it disables sync support.

chdiza:
The default should be to build with `gnutls`.  Doesn't that mean the `depends_on` should be `:recommended`, not `:optional`?

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/0fd477375e05757a94231a0c4c27e294b33040cb. Thank you @chdiza!

PR #45478, opened 2015-10-29
Tippecanoe 1.3.1

ericfischer:
* Tile generation is multithreaded to take advantage of multiple CPUs
* More compact data representation reduces memory usage and improves speed
* Polygon clipping uses [Clipper](http://www.angusj.com/delphi/clipper/documentation/Docs/_Body.htm)
  and makes sure interior and exterior rings are distinguished by winding order
* Individual GeoJSON features can specify their own minzoom and maxzoom
* New `tile-join` utility can add new properties from a CSV file to an existing tileset
* Feature coalescing, line-reversing, and reordering by attribute are now options, not defaults
* Output of `decode` utility is now in GeoJSON format
* Tile generation with a minzoom spends less time on unused lower zoom levels
* Bare geometries without a Feature wrapper are accepted
* Default tile resolution is 4096 units at all zooms since renderers assume it

### DIFF
--- a/Library/Formula/tippecanoe.rb
+++ b/Library/Formula/tippecanoe.rb
@@ -1,8 +1,8 @@
 class Tippecanoe < Formula
   desc "Build vector tilesets from collections of GeoJSON features"
   homepage "https://github.com/mapbox/tippecanoe"
-  url "https://github.com/mapbox/tippecanoe/archive/v1.2.0.tar.gz"
-  sha256 "237510a1a92a8626407c29c8d8047e73bbb22c1a8af8f9d1b8931d994c8fac2d"
+  url "https://github.com/mapbox/tippecanoe/archive/v1.3.1.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Looks like it's trying to ignore the `prefix` argument, perhaps? We link things ourselves; It's not necessary for projects to try to install to a top level prefix instead.

```
==> make install PREFIX=/usr/local/Cellar/tippecanoe/1.3.1
mkdir -p /usr/local/Cellar/tippecanoe/1.3.1/bin
cp tippecanoe /usr/local/Cellar/tippecanoe/1.3.1/bin/tippecanoe
cp man/tippecanoe.1 /usr/local/share/man/man1/
cp: /usr/local/share/man/man1/tippecanoe.1: Operation not permitted
make: *** [install] Error 1
```

ericfischer:
Oops, that was sloppy of me. Thanks. I'll fix it.

ericfischer:
Thanks. I think it should be fixed now.

bfontaine:
Thank you for your contribution to Homebrew @ericfischer! :tada:

PR #45481, opened 2015-10-29
Enable glib static libraries by default

jeroenooms:
LTDR: could we please include glib static libraries with the bottles? :dog:

Full story: the R software needs to statically link `libcairo` and dependencies for building redistributable addon packages. However currently `cairo` has one dependency that does not include static libraries by default: `glib`. Therefore the R addon package that was compiled on the build-server might not work for users that do not have brew and glib installed.

I know I can reinstall glib `--with-static`, however it would be super nice if static libraries automatically get included with the bottles. Is there any good argument against making this the default, as for most c libraries?

### DIFF
--- a/Library/Formula/glib.rb
+++ b/Library/Formula/glib.rb
@@ -3,6 +3,7 @@ class Glib < Formula
   homepage "https://developer.gnome.org/glib/"
   url "https://download.gnome.org/sources/glib/2.46/glib-2.46.1.tar.xz"
   sha256 "5a1f03b952ebc3a7e9f612b8724f70898183e31503db329b4f15d07163c8fdfb"
+  revision 1
 
   bottle do
     sha256 "d9a33e02f6765853ccaa85b1a221773256d61057f0755677b50634270c0c8b30" => :el_capitan
@@ -12,7 +13,6 @@ class Glib < Formula
 
   option :universal
   option "with-test", "Build a debug build and run tests. NOTE: Not all tests succeed yet"
-  option "with-static", "Build glib with a static archive."
 
   deprecated_option "test" => "with-test"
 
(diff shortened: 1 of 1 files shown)


Library/Formula/glib.rb
```diff
@@ -3,6 +3,7 @@ class Glib < Formula
   homepage "https://developer.gnome.org/glib/"
   url "https://download.gnome.org/sources/glib/2.46/glib-2.46.1.tar.xz"
   sha256 "5a1f03b952ebc3a7e9f612b8724f70898183e31503db329b4f15d07163c8fdfb"
+  revision 1
```
DomT4: I'd rather the people who need the static libraries reinstall `glib` via `brew` themselves rather than making every single Homebrew user who has `glib` installed rebuild personally, unless this is likely to hit a particularly vast number of people.

jeroenooms: What is the argument for this? Saving disk space?

DomT4: The bottles only work on this formula for `/usr/local` Homebrew users, and for 10.9-10.11. With a `revision` you'd be forcing every non-/usr/local and everyone on <10.9 to rebuild `glib` from source for a non-essential issue. We try to limit how often that sort of thing happens, because it's a pain.

mikemcquaid: Agreed with @DomT4.

mikemcquaid:
My only concern with this is that it ends up breaking software that relied on just the shared libraries being available.

mikemcquaid:
> Is there any good argument against making this the default, as for most c libraries?

Do you have a citation for this?

jeroenooms:
> Do you have a citation for this?

No, just my impression from inspecting other formulas and `ls /usr/local/lib/*.a`.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `glib: enable static libraries by default.`.

PR #45486, opened 2015-10-29
node 5.0.0

chrmoritz:
This PR updates node to the [latest stable release 5.0.0](https://nodejs.org/en/blog/release/v5.0.0/). 
It also updates npm from 2.14.7 to the latest 3.3.9 (because npm was upgraded to 3.3.x in the official node binaries too) and some test cleanup.

### DIFF
--- a/Library/Formula/node.rb
+++ b/Library/Formula/node.rb
@@ -1,8 +1,8 @@
 class Node < Formula
   desc "Platform built on the V8 JavaScript runtime to build network applications"
   homepage "https://nodejs.org/"
-  url "https://nodejs.org/dist/v4.2.1/node-v4.2.1.tar.gz"
-  sha256 "8861b9f4c3b4db380fcda19a710c0430c3d62d03ee176c64db63eef95a672663"
+  url "https://nodejs.org/dist/v5.0.0/node-v5.0.0.tar.gz"
+  sha256 "698d9662067ae6a20a2586e5c09659735fc0050769a0d8f76f979189ceaccdf4"
   head "https://github.com/nodejs/node.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


Library/Formula/node.rb
```diff
@@ -32,12 +32,13 @@ class Node < Formula
...
   resource "npm" do
-    url "https://registry.npmjs.org/npm/-/npm-2.14.7.tgz"
-    sha256 "54f3a5195a1e6b52fc954749e81f8abf1f7c8ca912b90ecf2b33581800d0f6a5"
+    url "https://registry.npmjs.org/npm/-/npm-3.3.9.tgz"
+    sha256 "b406bfc670045c2b92432aff419fe22ff28fd439bb0ef4faa14fd6f16bda0c22"
   end
 
   resource "icu4c" do
     url "https://ssl.icu-project.org/files/icu4c/56.1/icu4c-56_1-src.tgz"
+    mirror "http://ftp.mirrorservice.org/sites/download.qt-project.org/development_releases/prebuilt/icu/src/icu4c-56_1-src.tgz"
```
DomT4: You can use `https://` for the `mirrorservice`.

DomT4:
You beat me to this :smile_cat:. If anyone wants to file the 4.x branch in `homebrew/versions` as `node-lts` or something I'd welcome that.

chrmoritz:
Why not just `node4` for the `homebrew/versions` formula name? As LTS last for 18 months a `node-lts` can get confusing in a year when `node6` enters LTS but `node4` is still in LTS for 6 months.

BTW: changed the mirror to `https`

DomT4:
> Why not just node4 for the homebrew/versions formula name? 

We already have `node04`, `node06` & `node08` and `node4` at a glance doesn't necessarily appear newer. It also seems likely someone will install `node04` or `node4` expecting the other. Maybe `node4-lts` or something.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/b99e20502f3c00fa308e6aba1f003aa2b32e60ae. Thank you again @chrmoritz! :beer:

killswitch:
I don't see a confusion in `node04` and `node4`, but I agree with @chrmoritz on `node-lts` confusion when 6 comes out.

DomT4:
I'm personally a fan of `node4-lts`, as it covers all sides of the issue. Shows it has long-term support, shows is the 4x branch, people are less likely to mistype it on the keyboard and end up with the ancient `node04` or vice versa, etc.

stejmurphy:
node-lts sounds fair. 
When node 6 hits, the node-lts branch should migrate to this updated version and if need be, there could be a place for node-lts-old which would serve up 4 until it's no longer maintained and then migrate to 6 ..and then to the next.

killswitch:
ehh, not a fan of that. What if I install `node-lts` wanting 4, and then 4 is no longer supported but I still want to continue to use it, but I run `brew update && brew upgrade` because I want to keep everything else updated, I should still see `v4.x` when I run `node -v` not `v6.x` We should associate the actual version numbers with these versions.

Maybe introduce `node-v4` to curb confusion with `node04` and continue pumping out `node-v5`, `node-v6` etc.

DomT4:
I don't mind `node-v4`, I don't mind `node4-lts`. Let's agree on *something* and get it into Versions :smile_cat:.

PR #45489, opened 2015-10-29
rust 1.4.0

alex:


DomT4:
Punting this off the CI queue for a few minutes whilst I bump the timeout.

DomT4:
Back in the queue.

alex:
thanks!

dtrebbien:
@alex: What version of Mac OS X are you running?  I tried out similar changes on 10.11.1 'El Capitan' and I get an error:

<pre>
note: Undefined symbols for architecture x86_64:
  "_EVP_aes_256_ctr", referenced from:
      _libssh2_crypt_method_aes256_ctr in liblibssh2_sys-3d2e39bd460070f9.rlib(crypt.c.o)
     (maybe you meant: __libssh2_EVP_aes_256_ctr)
  "_EVP_aes_192_ctr", referenced from:
      _libssh2_crypt_method_aes192_ctr in liblibssh2_sys-3d2e39bd460070f9.rlib(crypt.c.o)
     (maybe you meant: __libssh2_EVP_aes_192_ctr)
  "_EVP_aes_128_ctr", referenced from:
      _libssh2_crypt_method_aes128_ctr in liblibssh2_sys-3d2e39bd460070f9.rlib(crypt.c.o)
     (maybe you meant: __libssh2_EVP_aes_128_ctr)
</pre>

alex:
@dtrebbien Not sure what would cause that, it looks green on CI

dtrebbien:
Hmmm.  I tried your patch exactly and I still get the same errors.

DomT4:
The only thing I can think of here is that @dtrebbien is building outside `/usr/local` where we've seen some issues reported upstream before, such as https://github.com/rust-lang/rust/issues/29133.

dtrebbien:
I am running the `brew install` command within `~/Documents/Projects/build`.  Is that what you mean by "building outside `/usr/local`"?

Homebrew is installed to `/usr/local`.

DomT4:
@dtrebbien Can you generate a `brew gist-logs` for this version of `rust` then? Thanks.

dtrebbien:
Here are gist logs:  https://gist.github.com/dtrebbien/accc7954b27be6f7acb6

When building via `brew gist-logs`, there is a different error:

<pre>
src/openssl_shim.c:1:10: fatal error: 'openssl/hmac.h' file not found
#include <openssl/hmac.h>
         ^
1 error generated.
</pre>

It appears that `brew gist-logs` sets up the build environment slightly differently.  Normally adding a dependency on openssl causes Homebrew to add `/usr/local/opt/openssl/lib/pkgconfig` to `PKG_CONFIG_PATH`.  However, as you can see, the `PKG_CONFIG_PATH` was not set up correctly by `brew gist-logs`.

dtrebbien:
Never mind my comment regarding `brew gist-logs`.  Those logs were from a different build with other changes.

dtrebbien:
https://gist.github.com/b695ca96abe2f888d9fa

dtrebbien:
In my case, the Rust libssh2-sys library is building its own copy of libssh2 using CMake.  The libssh2 CMake build script checks for the existence of the EVP_aes_128_ctr() function in OpenSSL's libcrypto.  If found, then libssh2 will use that function; otherwise, libssh2 provides an implementation.  CMake is checking the Homebrew version of OpenSSL (`/usr/local/Cellar/openssl/1.0.2d_1/lib/libcrypto.dylib`), and therefore the function existence check succeeds.  However, when `cargo` is linked, the library search paths are set up such that `/usr/lib` is searched before `/usr/local/Cellar/openssl/1.0.2d_1/lib`.  Because `/usr/lib/libcrypto.dylib` is missing the EVP_aes_128_ctr() function, the missing symbol errors occur.

Can others check their `/usr/lib/libcrypto.dylib` for the `_EVP_aes_128_ctr` symbol:

<pre>
nm -a /usr/lib/libcrypto.dylib
</pre>

I have not modified this file on my system, so I am not sure why the `cargo` link step fails on my system, but succeeds on other 'El Capitan' systems.

A fix for my system is to add a dependency on libssh2 to the formula.  This causes the Rust libssh2-sys library **not** to build its own copy of libssh2:

    depends_on "libssh2"

mikemcquaid:
> A fix for my system is to add a dependency on libssh2 to the formula. This causes the Rust libssh2-sys library not to build its own copy of libssh2:

This seems OK :+1:

DomT4:
> I have not modified this file on my system, so I am not sure why the cargo link step fails on my system, but succeeds on other 'El Capitan' systems.

You don't have the CLT installed, is the obvious detail. All the CIs ship both Xcode and CLT, so it would pass there *if* that was the case.

dtrebbien:
Hi @alex, did you want to push an amended commit with the extra `depends_on "libssh2"` line?

alex:
I'm not sure I understand why rust needs to depend on `libssh2`, it doesn't use it all, that's an external crate as far as I can tell?

dtrebbien:
`cargo` appears to have a dependency on the [libssh2-sys Rust package](https://crates.io/crates/libssh2-sys).  If libssh2 is not installed (findable via pkg-config), then libssh2-sys builds libssh2 from source using CMake.  The libssh2 built by libssh2-sys is meant to be linked with the Homebrew OpenSSL libs, but what can happen is the older-version Mac OS X OpenSSL libs are linked instead, causing missing symbol errors.

Adding a dependency on libssh2 to the formula just makes sure that the Homebrew OpenSSL libs are used.

alex:
Ahhh, ok, it's cargo itself. I'll add `libssh2` as a dependency then. Thanks for clarifiying.

DomT4:
Would you mind squashing this and rebasing it against the master? Several of us have tried to pull it locally but we're getting git error messages :cry:.

alex:
@DomT4 done

DomT4:
That did the trick. Merged in https://github.com/Homebrew/homebrew/commit/517aac3f84699829fc1c4d2d053d679ee5f76a14.

Thank you for another contribution to Homebrew @alex; we appreciate it! :smiley_cat:

PR #45490, opened 2015-10-29
librest 0.7.93

muellermartin:
librest (sometimes just called rest) is a rather small library to access
RESTful APIs as a part of the GNOME project.

After issues #31480 and #31501 weren't successful in adding librest to
Homebrew. Here is my try ;)

The test is rather useless (doesn't do anything useful) but should
compile if the library is installed correctly and works.

### DIFF
--- a/Library/Formula/librest.rb
+++ b/Library/Formula/librest.rb
@@ -32,8 +32,20 @@ class Librest < Formula
         return EXIT_SUCCESS;
       }
     EOS
+    glib = Formula["glib"]
+    libsoup = Formula["libsoup"]
     flags = (ENV.cflags || "").split + (ENV.ldflags || "").split
-    flags += `pkg-config --cflags --libs rest-0.7`.split
+    flags += %W[
+      -I#{libsoup.opt_include}/libsoup-2.4
+      -I#{glib.opt_include}/glib-2.0
+      -I#{glib.opt_lib}/glib-2.0/include
+      -I#{include}/rest-0.7
+      -L#{libsoup.opt_lib}
+      -L#{glib.opt_lib}
+      -L#{lib}
+      -lrest-0.7
+      -lgobject-2.0
+    ]
     system ENV.cc, "test.c", "-o", "test", *flags
     system "./test"
   end
--- a/Library/Formula/librest.rb
+++ b/Library/Formula/librest.rb
@@ -1,5 +1,5 @@
 class Librest < Formula
-  desc "librest is a library access web services that claim to be RESTful"
+  desc "Library to access RESTful web services"
   homepage "https://wiki.gnome.org/Projects/Librest"
   url "https://download.gnome.org/sources/rest/0.7/rest-0.7.93.tar.xz"
   sha256 "c710644455340a44ddc005c645c466f05c0d779993138ea21a62c6082108b216"
--- a/Library/Formula/librest.rb
+++ b/Library/Formula/librest.rb
@@ -1,0 +1,40 @@
+class Librest < Formula
+  desc "librest is a library access web services that claim to be RESTful"
+  homepage "https://wiki.gnome.org/Projects/Librest"
+  url "https://download.gnome.org/sources/rest/0.7/rest-0.7.93.tar.xz"
+  sha256 "c710644455340a44ddc005c645c466f05c0d779993138ea21a62c6082108b216"
+
+  depends_on "pkg-config" => :build
+  depends_on "glib"
+  depends_on "libsoup"
+  depends_on "gobject-introspection"
+
+  def install
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}",
+                          "--without-gnome",
+                          "--without-ca-certificates",
+                          "--enable-introspection=yes"
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.c").write <<-EOS.undent
+      #include <stdlib.h>
+      #include <rest/rest-proxy.h>
+
+      int main(int argc, char *argv[]) {
+        RestProxy *proxy = rest_proxy_new("http://localhost", FALSE);
+
+        g_object_unref(proxy);
+
+        return EXIT_SUCCESS;
+      }
+    EOS
+    flags = (ENV.cflags || "").split + (ENV.ldflags || "").split
+    flags += `pkg-config --cflags --libs rest-0.7`.split
+    system ENV.cc, "test.c", "-o", "test", *flags
+    system "./test"
+  end
+end


Library/Formula/librest.rb
```diff
@@ -0,0 +1,40 @@
...
+      int main(int argc, char *argv[]) {
+        RestProxy *proxy = rest_proxy_new("http://localhost", FALSE);
+
+        g_object_unref(proxy);
+
+        return EXIT_SUCCESS;
+      }
+    EOS
+    flags = (ENV.cflags || "").split + (ENV.ldflags || "").split
+    flags += `pkg-config --cflags --libs rest-0.7`.split
```
bfontaine: We can’t rely on `pkg-config` in the test; we’ll need to refer explicitly to the formula. See [`tinyxml2.rb`](https://github.com/Homebrew/homebrew/blob/73028c892c7a56b7b2130c0413075898f01c2ef8/Library/Formula/tinyxml2.rb#L31) for an example.

Library/Formula/librest.rb
```diff
@@ -0,0 +1,40 @@
+class Librest < Formula
+  desc "librest is a library access web services that claim to be RESTful"
```
bfontaine: > a library **to** access

Please also remove “librest is a”: `desc "Library to access …"`. We could also shorten `web services that claim to be RESTful` to `RESTful Web services`.

Library/Formula/librest.rb
muellermartin: I thought I could shorten the code, but I'll change that. Thanks for the info!

Library/Formula/librest.rb
muellermartin: I copied most of the text from the website, but I can live with your suggestion, so I'll change that, too ;)

Library/Formula/librest.rb
muellermartin: Just as a side question: Why can't the test rely on pkg-config? The Formula depends on it anyway, so shouldn't it be available then?

DomT4: It's a `build` dependency. We remove those in testing to make sure people aren't mislabelling dependencies.

bfontaine: Also since it’s a `build` dependency one is free to remove it after installing `librest` and still expect `brew test librest` to succeed.

muellermartin: Hm, that sounds reasonable. Thanks for clarifying! The Formula itself is a just a library which also provides a `.pc` file, but surely it can be used without using it.
Just so I understand it: If I would declare the `pkg-config` dependency as not `build`-only, could I use `pkg-config` in the test then?

muellermartin:
The CI test builds seem to fail because `brew audit librest` results in an error which claims `0.7.93 is a development release`.
This error seemed odd to me, because I could not see why this would be a development release. I'm no frequent contributor to Homebrew, so I tried to look it up because I thought I forgot sth. in the Formula, but searching the web revealed the cause in the [source code of `brew audit`](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/audit.rb#L546): When the source is obtained from the GNOME project, it is assumed that odd minor versions are development versions. I don't know if that is also true for this library, as I'm not involved in the project (but I could ask the devs, if they see it as stable…).
Looking at other package managers, most of them declare this version as stable:
- Gentoo: https://packages.gentoo.org/packages/net-libs/rest
- Ubuntu: http://packages.ubuntu.com/wily/librest-0.7-0 and http://packages.ubuntu.com/wily/librest-dev
- Arch: https://www.archlinux.org/packages/extra/x86_64/rest/
- Fedora: https://apps.fedoraproject.org/packages/rest/overview/
- Macports: https://trac.macports.org/browser/trunk/dports/gnome/rest/Portfile

These are options I suggest to fix this:
- ignore the test (if that's possible)
- change the Homebrew code (e.g. by not failing hard)
- ask the devs if they reflect the stable versioning by changing the version number
- chose an older lower "stable" version number for this Formula

DomT4:
GNOME packages tend to follow the pattern of stable = even, devel = odd, hence the audit check. If that isn't the case here, we can ignore it. It does apply for about 90% of the GNOME projects we ship.

muellermartin:
@DomT4 Okay, that confirmed my assumption made from the source code. This odd/even versioning system is even described on Wikipedia so it's a thing ;)

Judging all the package managers I searched for (lib)rest most, if not all of them, consider this version stable. So I guess Homebrew would just be another one then.

And thanks for the quick reply!

mikemcquaid:
@DomT4 @tschoonj I think that audit rule may not apply when it's less than `1.0.0`? If that sounds right then maybe I should adjust it that way.

tschoonj:
Hmmm.. this looks very much like a development release to me. The 93 is a clear indicator here: this precedes a future 0.8.0 release. Perhaps this should be added to Homebrew-devel while we wait for a stable release?

It's a bit of a weird gnome project: the downloads are not hosted at download.gnome.org like all the other ones. But it does have its own gnome page at https://wiki.gnome.org/Projects/Librest so it looks legit.

muellermartin:
@tschoonj Of course it is hosted at download.gnome.org: https://download.gnome.org/sources/rest/0.7/
As I pointed out 100% of the other package managers use this version as stable. But I'll try to reach out to the devs to get a statement from them...

tschoonj:
Sorry, didn't see that as I was looking for librest in the list...

muellermartin:
@tschoonj No problem, it's indeed confusing that this lib is sometimes referred to as librest and other times it's called only rest (e.g. Gentoo's package is net-libs/rest).

mikemcquaid:
> But I'll try to reach out to the devs to get a statement from them...

Let me know when you find out :+1: 



> @DomT4 @tschoonj I think that audit rule may not apply when it's less than 1.0.0? If that sounds right then maybe I should adjust it that way.

@tschoonj What would your thoughts be on adjusting our `audit` that way?

tschoonj:
The audit rules are difficult to generalize for GNOME packages, especially when below 1.0.0. This one seems to follow the most commonly used one for development releases: minor number is odd, and the last one indicates that the developers are close to releasing the next stable one (with even minor number).

Perhaps the audit should just remain the way it is, and deal with packages that fail the audit on a case by case basis?

mikemcquaid:
@tschoonj The issue I have with that is that then we'll get failed CI notifications every time we merge a bump. To me: if we can't be sure under 1.0 that sounds like a good reason to remove the audit below then. Thanks for the feedback!

mikemcquaid:

For future reference the preferred commit message format for new formulae is `librest 0.7.93 (new formula)`.

mikemcquaid:
Updated the audit rule in a6269bd/

PR #45491, opened 2015-10-29
elasticsearch 2.0

brainix:


Library/Formula/elasticsearch.rb
```diff
@@ -47,12 +47,11 @@ def install
     # Set up Elasticsearch for local development:
     inreplace "#{prefix}/config/elasticsearch.yml" do |s|
       # 1. Give the cluster a unique name
-      s.gsub!(/#\s*cluster\.name\: elasticsearch/, "cluster.name: #{cluster_name}")
+      s.gsub!(/#\s*cluster\.name\: my-application/, "cluster.name: #{cluster_name}")
 
       # 2. Configure paths
       s.sub!(%r{#\s*path\.data: /path/to.+$}, "path.data: #{var}/elasticsearch/")
       s.sub!(%r{#\s*path\.logs: /path/to.+$}, "path.logs: #{var}/log/elasticsearch/")
-      s.sub!(%r{#\s*path\.plugins: /path/to.+$}, "path.plugins: #{var}/lib/elasticsearch/plugins")
```
bfontaine: Where are plugins installed now?

brainix: Good catch, @bfontaine, thank you.  The example config file no longer has a line to customize the plugin path, so I added the line to the end of the `brew` customized config file.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `elasticsearch 2.0.0`.

mikemcquaid:
This doesn't seem to work correctly so I reverted it: https://github.com/Homebrew/homebrew/issues/45593

mikemcquaid:
We should also probably improve the `test do` so this can't happen again.

brainix:
Thanks for the heads up.  I'll work on writing more thorough tests and resubmit!

I believe that there are some permissions issues - my local environment may be different from others.  The formula works for me locally, but once you reopened this PR, I noticed that it's broken on my work computer.  Good catch, and apologies.

jasontedor:
@brainix Thanks for lending a hand towards getting Elasticsearch 2.0.0 working with Homebrew. Unfortunately, there are some underlying issues that need to be worked out as discussed in #45644.

brainix:
@jasontedor Thank you!  Yes, this formula quickly escalated beyond my depth of understanding of Homebrew.  I'll close this PR to avoid confusion, and the discussion can happen on #45644.

Issue #42191, opened 2015-07-28
qt5 5.5.0 errors in macdeployqt


fzwoch:
Possibly configuration error for qt5 5.5.0? Upgrading to 5.5.0 broke the "macdeployqt" application for me:

```
Log: Adding framework:
Log: Framework name "QtCore.framework" 
 Framework directory "/usr/local/Cellar/qt5/5.5.0_1/lib/" 
 Framework path "" 
 Binary directory "" 
 Binary name "" 
 Binary path "" 
 Version "" 
 Install name "/usr/local/Cellar/qt5/5.5.0_1/lib/" 
 Deployed install name "" 
 Source file Path "" 
 Framework Destination Directory (relative to bundle) "" 
 Binary Destination Directory (relative to bundle) "" 
```

This looks suspicious to me. With the official install from qt.io It looks more like this:

```
Log: Framework name "QtGui.framework" 
 Framework directory "/Users/fzwoch/Qt5.5.0/5.5/clang_64/lib/" 
 Framework path "/Users/fzwoch/Qt5.5.0/5.5/clang_64/lib/QtGui.framework" 
 Binary directory "Versions/5" 
 Binary name "QtGui" 
 Binary path "/Versions/5/QtGui" 
 Version "5" 
 Install name "/Users/fzwoch/Qt5.5.0/5.5/clang_64/lib/QtGui.framework/Versions/5/QtGui" 
 Deployed install name "@executable_path/../Frameworks/QtGui.framework/Versions/5/QtGui" 
 Source file Path "/Users/fzwoch/Qt5.5.0/5.5/clang_64/lib/QtGui.framework/Versions/5/QtGui" 
 Framework Destination Directory (relative to bundle) "Contents/Frameworks/QtGui.framework" 
 Binary Destination Directory (relative to bundle) "Contents/Frameworks/QtGui.framework/Versions/5" 
```

There are several of these observations. Further down they lead to wrong/invalid otool commands:

```
Log: Using install_name_tool:
Log:  in "client.app/Contents/Frameworks/QtGui.framework/Versions/5/QtGui"
Log:  change reference "/usr/local/Cellar/qt5/5.5.0_1/lib/"
Log:  to ""
```

Which finally results in many of these errors:

```
ERROR: no file at ""
```

I'm not sure whether this is a bug or did I a miss an additional configuration step?

DomT4:
Please `brew update && brew reinstall qt5`, Apologies.

[...]

PR #45494, opened 2015-10-30
keg_relocate: fix relocation of frameworks

UniqMartin:
When fixing references to regular dylibs, it is sufficient to search for a file with the same base name, e.g., `libpoppler.56.dylib`. However, if the broken reference is to a framework, we also have to take into account preceding path components to find a suitable match. Framework references (according to the `dyld` man page) come in two flavors:

- `XXX.framework/Versions/YYY/XXX` (with version)
- `XXX.framework/XXX` (without version)

The change here is to detect these patterns and to make sure that the fixed library reference has the same suffix as the broken one.

Prior to this fix, a broken framework reference to, e.g., `QtXml.framework/Versions/5/QtXml` (if originating in a sister framework) would have been rewritten to `<qt5-keg>/lib/QtXml.framework/QtXml`. In practice, this mostly works, but is technically incorrect and thus creates problems like in #42191. With this fix, the framework reference is correctly rewritten to `<qt5-keg>/lib/QtXml.framework/Versions/5/QtXml`.

### DIFF
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -130,7 +130,7 @@ class Keg
       "@loader_path/#{bad_name}"
     elsif file.mach_o_executable? && (lib + bad_name).exist?
       "#{lib}/#{bad_name}"
-    elsif (abs_name = find_dylib(Pathname.new(bad_name).basename)) && abs_name.exist?
+    elsif (abs_name = find_dylib(bad_name)) && abs_name.exist?
       abs_name.to_s
     else
       opoo "Could not fix #{bad_name} in #{file}"
@@ -156,8 +156,22 @@ class Keg
     opt_record.join(relative_dirname, basename).to_s
   end
 
-  def find_dylib(name)
-    lib.find { |pn| break pn if pn.basename == name } if lib.directory?
+  # Matches framework references like `XXX.framework/Versions/YYY/XXX` and
+  # `XXX.framework/XXX`, both with or without a slash-delimited prefix.
+  FRAMEWORK_RX = %r{(?:^|/)(([^/]+)\.framework/(?:Versions/[^/]+/)?\2)$}.freeze
+
+  def find_dylib_suffix_from(bad_name)
+    if (framework = bad_name.match(FRAMEWORK_RX))
+      framework[1]
+    else
+      File.basename(bad_name)
+    end
+  end
+
+  def find_dylib(bad_name)
+    return unless lib.directory?
+    suffix = "/#{find_dylib_suffix_from(bad_name)}"
+    lib.find { |pn| break pn if pn.to_s.end_with?(suffix) }
   end
 
   def mach_o_files


mikemcquaid:
Makes sense to me but would like to see @xu-cheng and/or @mistydemeo :+1: first.

UniqMartin:
Would love to see this merged, but definitely not without a proper review from a maintainer. It's a delicate change, even if I'm pretty confident and it hasn't broken anything for me in the past two weeks.

Would putting a `maintainer feedback` label on this PR help?

DomT4:
If a direct ping didn't work the label may well go unnoticed as well, but we can try.

mikemcquaid:
I suspect it will. @UniqMartin can you provide your test cases so I can verify this works as expected locally for me too? After that I'm happy to just merge it.

UniqMartin:
> can you provide your test cases so I can verify this works as expected locally for me too?

The last paragraph of the initial comment hints at my test case. It is the `qt5` formula, but (!) *without the fix that landed today*. To see the effect of this change one would have to do something like this:

```
cd $(brew --prefix)

git checkout -b framework-relocation-test 6e43e9a3
brew remove qt5
brew install -d qt5 --without-webengine --without-webkit | tee before.log

git merge UniqMartin/framework-relocation
brew remove qt5
brew install -d qt5 --without-webengine --without-webkit | tee after.log
```

Compare the `brew install` outputs, especially the output generated by `Keg#change_install_name` (`Changing install name in […]` and following lines).

mikemcquaid:
@UniqMartin I'm sorry that I keep making you repeat yourself but: why is this not needed for the current Qt5 formula? Are there any other formulae this applies to? Thanks!

UniqMartin:
The reason is that the logic I'm modifying here only applies to dylibs/frameworks built with a relative install name (dylib ID). Furthermore, to see the issue, there must be multiple frameworks in the same formula that depend on each other. In the simplest case, e.g., `A.framework` and `B.framework` are both provided by the same formula and `A.framework` has a dependency on `B.framework`.

But with the fix that went into `qt5` today, those frameworks are now built with an absolute dylib ID (as is the norm on OS X), so `Keg#fix_install_names` won't even touch them: [lines 11–12 in keg_relocate.rb](https://github.com/Homebrew/homebrew/blob/3363d246d03707cc150446e5f90b9be22a769e0d/Library/Homebrew/keg_relocate.rb#L11-L12).

Does that make some sense?

> Are there any other formulae this applies to?

Not that I'm aware of. But I'm only seeing and using a very small portion of the available formulae.

mikemcquaid:
Cool, thanks for the explanation. Last question: do you see any way this could break existing formulae or any assumptions they have? If not: I trust your judgement and will just merge this.

UniqMartin:
> Last question: do you see any way this could break existing formulae or any assumptions they have?

I'm pretty confident this won't break anything. And it would have prevented #42191 if it was already in place when Qt 5.5.0 came around.

mikemcquaid:
Great, thanks @UniqMartin!

UniqMartin:
Thanks for your confidence in my work and for taking the time to handle this yourself! :heart: Feel free to direct any issues that smell like they might have been caused by this my way.

mikemcquaid:
Will do! Thanks for explaining :heart:

DomT4:
Heads up: Older Rubies *really* don't like this:

```
~> ~/Downloads/brew-1.8 readall
Error: /usr/local/Library/Homebrew/keg_relocate.rb:7: undefined (?...) sequence: /(?:^|\/)(?<suffix>(?<name>[^\/]+)\.framework\/(?:Versions\/[^\/]+\/)?\k<name>)$/
Please report this bug:
/usr/local/Library/Homebrew/keg.rb:2:in `require'
/usr/local/Library/Homebrew/keg.rb:2
/usr/local/Library/Homebrew/formula_pin.rb:1:in `require'
/usr/local/Library/Homebrew/formula_pin.rb:1
/usr/local/Library/Homebrew/formula.rb:3:in `require'
/usr/local/Library/Homebrew/formula.rb:3
/usr/local/Library/Homebrew/cmd/readall.rb:6:in `require'
/usr/local/Library/Homebrew/cmd/readall.rb:6
/usr/local/Library/brew.rb:72:in `require'
/usr/local/Library/brew.rb:72:in `require?'
/usr/local/Library/brew.rb:123
```

Is the cause of #45814.

mikemcquaid:
Reverted in 98c044a. Gonna get a Travis 1.8 build going now.

UniqMartin:
Sorry about that. I never worked with Ruby 1.8 and wasn't expecting it to fail because of this. :sob:

mikemcquaid:
It's OK @UniqMartin, we really need to fix the CI around this, it's not your fault.

xu-cheng:
Sorry being late. 

Since it's reverted, can we reopen this PR(or create a new one) to keep track the issue?

mikemcquaid:
If @UniqMartin undeletes the branch I believe I can reopen it. Otherwise: a new PR would be great (but I want to hold off merging until we have 1.8 CI).

UniqMartin:
My intention was to post this as a new PR once the 1.8 CI is ready (I'm keeping an eye on that). If restoring the branch and reopening this PR is the better solution, let's do that (I've no experience in this regard). In any case, I'll make sure this fix won't be lost.

UniqMartin:
@mikemcquaid Just restored the branch, if you want to reopen this.

mikemcquaid:
Thanks @UniqMartin!

UniqMartin:
Alright, now it's using 1.8.7-compatible captures and back references in the regular expression. (I also moved the regexp constant closer to where it is used.) I think it's now ready to be merged for good. :pray:

mikemcquaid:
Thanks @UniqMartin!

PR #45495, opened 2015-10-30
emacs_requirement: set $EMACS

dunn:
The Emacs shell sets $EMACS to "t" for detection purposes, but it causes
builds to fail when they attempt to call Emacs using the variable.

Fixes Homebrew/homebrew-emacs#30.


mikemcquaid:
:+1:

PR #45496, opened 2015-10-30
postgresql: update sha256 of pgroonga

niku:
I ask to @kou who mantainer of pgroonga via twitter.

(In Japanese)
https://twitter.com/niku_name/status/659909397615763457 

He saids that:
> I updated new package to same URL, and I tried to revert it, but I have already lost previous package.


bfontaine:
Thanks a lot @niku! :tada:

PR #45501, opened 2015-10-30
JavaRequirement: check satisfaction directly

xu-cheng:
It made less sense to call a method `java_version` when it returns
boolean value.


mikemcquaid:
:+1:

PR #45403, opened 2015-10-27
docker-machine update to 0.5.0

vlesierse:
See issue #45352 

Docker Machine 0.5.0 has been released and this PR can be merged

PR #45503, opened 2015-10-30
docker-machine: update the installation process of Docker Machine v0.5

skatsuta:
Docker Machine v0.5 has simplified its build process using Makefile.


skatsuta:
Sorry, dup #45403

PR #45512, opened 2015-10-30
libmill 1.0; import from dev-only

bfontaine:
See https://github.com/Homebrew/homebrew-devel-only/pull/35.


bfontaine:
The build fails because of the name conflict.

bfontaine:

PR #45517, opened 2015-10-30
postgrest 0.2.12.0 (new formula)

xDAGRONx:
Add formula for PostgREST, which serves up a RESTful API from any
existing PostgreSQL database.

### DIFF
--- a/Library/Formula/postgrest.rb
+++ b/Library/Formula/postgrest.rb
@@ -1,0 +1,48 @@
+require "language/haskell"
+require "net/http"
+
+class Postgrest < Formula
+  include Language::Haskell::Cabal
+
+  desc "Serves a fully RESTful API from any existing PostgreSQL database."
+  homepage "https://github.com/begriffs/postgrest"
+  url "https://github.com/begriffs/postgrest/archive/v0.2.12.0.tar.gz"
+  sha256 "9f7277720b947b06eb53ac0a54686eb437253d417695bc756220e703532a725a"
+
+  depends_on "ghc" => :build
+  depends_on "cabal-install" => :build
+  depends_on "postgresql"
+
+  setup_ghc_compilers
+
+  def install
+    install_cabal_package
+  end
+
+  test do
+    pg_bin  = Formula["postgresql"].bin
+    pg_port = 55561
+    pg_user = "postgrest_test_user"
+    test_db = "test_postgrest_formula"
+
+    system "#{pg_bin}/initdb", "-D", testpath/test_db,
+      "--auth=trust", "--username=#{pg_user}"
+
+    system "#{pg_bin}/pg_ctl", "-D", testpath/test_db, "-l",
+      testpath/"#{test_db}.log", "-w", "-o", %("-p #{pg_port}"), "start"
+
+    begin
+      system "#{pg_bin}/createdb", "-w", "-p", pg_port, "-U", pg_user, test_db
+      pid = Process.spawn("postgrest -d #{test_db} -P #{pg_port} " \
+                          "-U #{pg_user} -a #{pg_user} -p 55560",
+                          :out => "/dev/null", :err => "/dev/null",)
+      sleep(5) # Wait for the server to start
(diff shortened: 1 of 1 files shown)


Library/Formula/postgrest.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  setup_ghc_compilers
+
+  def install
+    cabal_sandbox do
+      cabal_install "--only-dependencies"
+      cabal_install "--prefix=#{prefix}"
+    end
+
+    cabal_clean_lib
```
bfontaine: You can replace this whole code with `install_cabal_package`:

```ruby
def install
  install_cabal_package
end
```

Library/Formula/postgrest.rb
```diff
@@ -0,0 +1,29 @@
+require "language/haskell"
+
+class Postgrest < Formula
+  include Language::Haskell::Cabal
+
+  desc "PostgREST serves a RESTful API from any existing PostgreSQL database."
```
bfontaine: Please don’t start the description with the formula name.

Library/Formula/postgrest.rb
```diff
@@ -0,0 +1,24 @@
...
+  depends_on "postgresql"
+
+  setup_ghc_compilers
+
+  def install
+    install_cabal_package
+  end
+
+  test do
+    system "#{bin}/postgrest", "--help"
```
dunn: Any chance of a test that checks more of the functionality? `cmake.rb` and `when.rb` are good examples of tests.  Thanks!

xDAGRONx: Unfortunately, I'm not sure of an easy way to write a better test. The tool requires access to a Postgres database. I thought about using the default database for testing, but if a user modifies that database, the tests would likely fail.

dunn: If you're feeling ambitious, you could try to copy some of the [`madlib.rb`](https://github.com/Homebrew/homebrew-science/blob/master/madlib.rb#L88-L103) test from the Science repository; but if it doesn't end up working this is probably OK.

mikemcquaid: Can it not create a postgres database in a temporary directory? It does depend on postgres, after all.

xDAGRONx: I can implement something like the madlib.rb test. The biggest problem there is that if someone requires a password for the ENV["USER"] postgres user, the test will fail.

mikemcquaid: Creating a new database in the temporary directory with `initdb` shouldn't require a password, right?

xDAGRONx: Maybe. But the postgrest tool requires a user (and password, if the user account requires one) to connect to the database.

mikemcquaid: Ok. I _think_ that's set by the `initdb` call?

xDAGRONx: I'm not sure. I just pushed up a revision using a test similar to the madlib.rb example. I can change it if there is a better way.

Library/Formula/postgrest.rb
```diff
@@ -0,0 +1,51 @@
...
+    test_db = "test_postgrest_formula"
+
+    # Start PostgreSQL server if it is not already running
+    if File.exist? "#{var}/postgres/postmaster.pid"
+      pg_running = true
+    else
+      pg_running = false
+      system "launchctl", "load", "-w", pg_plist
+      sleep(5) # Wait for the server to start
+    end
```
mikemcquaid: Would be good to start a dedicated server on a dedicated port rather than piggy-backing on the global version (which may be used for other things)?

xDAGRONx: I'm not sure I see the disadvantage of piggy-backing on the global version. If I start a dedicated server on a dedicated port, I would have to add logic to find an open port, or risk colliding with something already running on that port (which I'm already doing with the postgrest server).

mikemcquaid:

PR #45531, opened 2015-10-30
ne 3.0.1 (new formula)

alehed:
This adds a formula for the ne text editor. Ne is a lightweight
text editor with a focus on performance and little bandwidth use.
Apart from that it is designed to be approachable yet extensible
enough to keep an advanced user productive.
Development is located at https://github.com/vigna/ne.

### DIFF
--- a/Library/Formula/ne.rb
+++ b/Library/Formula/ne.rb
@@ -1,0 +1,19 @@
+class Ne < Formula
+  desc "The nice editor"
+  homepage "http://ne.di.unimi.it"
+  url "http://ne.di.unimi.it/ne-3.0.1.tar.gz"
+  sha256 "92b646dd2ba64052e62deaa4239373821050a03e1b7d09d203ce04f2adfbd0e4"
+
+  def install
+    system "make"
+    system "make", "PREFIX=#{prefix}", "install"
+  end
+
+  test do
+    ENV["TERM"] = "xterm"
+    (testpath/"test.txt").write("This is a test document.\n")
+    (testpath/"test_ne").write("GotoLine 2\nInsertString line 2\nExit\n")
+    system "script", "-q", "/dev/null", "#{bin}/ne", "--macro", "#{(testpath/"test_ne")}", "#{(testpath/"test.txt")}"
+    assert_equal "This is a test document.\nline 2", File.read("#{testpath}/test.txt")
+  end
+end


Library/Formula/ne.rb
```diff
@@ -0,0 +1,17 @@
+class Ne < Formula
+  desc "ne, the nice editor"
+  homepage "http://ne.di.unimi.it"
+  url "http://ne.di.unimi.it/ne-3.0.1.tar.gz"
+  sha256 "92b646dd2ba64052e62deaa4239373821050a03e1b7d09d203ce04f2adfbd0e4"
+
+  bottle :unneeded
```
dunn: Why shouldn't it be bottled?

Library/Formula/ne.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def install
+    system "make"
+    system "make", "PREFIX=#{prefix}", "install"
+  end
+
+  test do
+    system "#{bin}/ne", "--help"
```
dunn: We'd love a test that checks a bit more of the functionality!

Library/Formula/ne.rb
```diff
@@ -0,0 +1,17 @@
+class Ne < Formula
+  desc "ne, the nice editor"
```
DomT4: Please don't start formula descriptions with the formula name. That happens automatically.

alehed:
This is my first formula so bear with me.
Found a way to test the core editing functionality of the editor. I picked up the bottle :unneeded from some other formula and I didn't see any use in providing a precompiled keg if it's not hosted anywhere.

alehed:
It seems like the files could not be created on travis. I'll have a look at it afterwards

alehed:
For now I can't determine what the problem is exactly. The block is caused by the editor asking for user feedback which happens every time it can't write to the file or there file was written by another application in the meantime.

DomT4:
Please be careful how many pushes you make here; each push is ran through our CI and doing so many is quite resource intensive.

You can try adding `ENV["TERM"] = "xterm"` to the top of `test do`, which has helped in similar errors before.

alehed:
Ok, I'll give this one more try and then I forfeit.

DomT4:
I can't get it to fail locally, so I do suspect `ENV["TERM"] = "xterm"` will help.

alehed:
Unfortunately ne has no command line options that avoid opening the editor except --help. The only thing that seems possible to me is different file permissions on travis, because it works just fine on my machine.

hussain0ip:
hi guys 
i need some help for installing android stuido

hussain0ip:
please speek with me on tiwtter @thesnipergreat

alehed:
Looks like you're in the wrong thread

Library/Formula/ne.rb
```diff
@@ -0,0 +1,19 @@
...
+  def install
+    system "make"
+    system "make", "PREFIX=#{prefix}", "install"
+  end
+
+  test do
+    ENV["TERM"] = "xterm"
+    (testpath/"test.txt").write("This is a test document.\n")
+    (testpath/"test_ne").write("GotoLine 2\nInsertString line 2\nExit\n")
+    system "#{bin}/ne", "--macro", "#{(testpath/"test_ne")}", "#{(testpath/"test.txt")}"
```
dunn: What may work is changing this line to:
```ruby
system "script", "-q", "/dev/null", "#{bin}/ne", "--macro", "#{(testpath/"test_ne")}", "#{(testpath/"test.txt")}"
```
to trick `ne` into thinking input is coming from a tty.

alehed:
The CI part passed! Thanks a lot @dunn @DomT4 for the support, couldn't have done it on my own!

Library/Formula/ne.rb
```diff
@@ -0,0 +1,19 @@
+class Ne < Formula
+  desc "The nice editor"
+  homepage "http://ne.di.unimi.it"
+  url "http://ne.di.unimi.it/ne-3.0.1.tar.gz"
+  sha256 "92b646dd2ba64052e62deaa4239373821050a03e1b7d09d203ce04f2adfbd0e4"
+
```
dunn: Couldn't hurt to add:
```ruby
head "https://github.com/vigna/ne.git"
```
as well

dunn: Oh, building from the repository requires LaTeX for the documentation. Nevermind, then.

mikemcquaid:

PR #45534, opened 2015-10-31
rust-multirust conflict

InPermutation:
> [Multirust](https://github.com/brson/multirust) is a tool for managing multiple installations of the Rust toolchain. It replaces the standard Rust toolchain with components that dynamically choose between alternate implementations based on context.

Both formulae fail with a link error, naturally, if the other one is installed, because they are both trying to link programs named `rustc`, `rustdoc`, `cargo`, `rust-lldb`, `rust-gdb`. They are mutually exclusive packages.


mikemcquaid:

PR #45535, opened 2015-10-31
doc: be consistent when referring to official taps

UniqMartin:
Always use the `<user/repo>` format as used by the `brew tap` command. Also fixed a link to `homebrew/head-only` and sorted a list in `FAQ.md` in alphabetical order.


mikemcquaid:
Thanks @UniqMartin!

PR #45548, opened 2015-10-31
watchman 4.1.0

wez:
In addition to various improvements in the watchman service, we also introduced two new python applications in this release, so the pythonic bits bear some scrutiny.

Tested via:

```
brew install watchman
brew audit watchman
brew test watchman
```

Previously we were shipping python bindings for the watchman client, but doing this blissfully unaware of the information in https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md, which yields issues such as https://github.com/facebook/watchman/issues/184#issuecomment-152599517

This change embraces the idea that the `watchman` brew formula is about applications and that folks that want bindings can `pip install pywatchman`.

To do this we need to disable the automatic python detection in the watchman configure script and just use the homebrew python procedures from the linked guide to manage installing the scripts.

### DIFF
--- a/Library/Formula/watchman.rb
+++ b/Library/Formula/watchman.rb
@@ -1,8 +1,8 @@
 class Watchman < Formula
   desc "Watch files and take action when they change"
   homepage "https://github.com/facebook/watchman"
-  url "https://github.com/facebook/watchman/archive/v3.9.0.tar.gz"
-  sha256 "1739cd2d6846cc688b12911c37406fae5601d76c0d11f3da957c2b7273941221"
+  url "https://github.com/facebook/watchman/archive/v4.1.0.tar.gz"
+  sha256 "5bc579475a8a26f5e1af58abbf848a7c3067524b9be448f98feba9e455284eeb"
   head "https://github.com/facebook/watchman.git"
 
   bottle do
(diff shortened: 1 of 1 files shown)


Library/Formula/watchman.rb
```diff
@@ -11,18 +11,29 @@ class Watchman < Formula
...
+                          "--with-pcre",
+                          # we'll do the homebrew specific python
+                          # installation below
+                          "--without-python"
     system "make"
     system "make", "install"
+
+    # Homebrew specific python application installation
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    Dir.chdir "python"
```
DomT4: You should be able to do:

```ruby
  cd "python" do
    system "python", *Language::Python.setup_install_args(libexec)
  end
  bin.install Dir[libexec/"bin/*"]
end
```

wez:
let's try ;-)  Updated and will let CI prove it

DomT4:
CI seems happy. :ship:'ed in https://github.com/Homebrew/homebrew/commit/fa709b2b33e1e2e12cebbe16a1f29d93941812ca. Thank you once more @wez! :smiley:

PR #45551, opened 2015-10-31
ponyc 0.2.1

colindean:
* Moves URL to Github since releases.ponylang.org is no longer updated
* Changes test to build stdlib instead of builtin, because builtin is not meant to be built directly by the compiler according to https://github.com/CausalityLtd/ponyc/issues/335


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/349fdbd8b00ab2bc2464ac3648dca7268c956438. Thank you for another contribution to Homebrew @colindean! :smiley_cat:

colindean:
Thanks @DomT4 :+1:

PR #45557, opened 2015-10-31
macvim: Snapshot 79

chdiza:


chdiza:
Among other more important things, the problem of building on Yosemite under Xcode 7.1 has been resolved, so we can delete the workaround for it.

PR #45558, opened 2015-10-31
teensy_loader_cli: add HEAD patch to avoid libusb access error on Mac OS

rixtox:
According to https://github.com/PaulStoffregen/teensy_loader_cli/issues/11
it's suggested to compile with `OS=LINUX` and include libusb, and due to an issue described in
https://github.com/PaulStoffregen/teensy_loader_cli/blob/master/teensy_loader_cli.c#L246-L248
libusb behave differently on Mac OS. Thus the patch comments out the lines as described to fix the problem.

### DIFF
--- a/Library/Formula/teensy_loader_cli.rb
+++ b/Library/Formula/teensy_loader_cli.rb
@@ -1,13 +1,21 @@
 class TeensyLoaderCli < Formula
   desc "Command-line integration for Teensy USB development boards"
   homepage "https://www.pjrc.com/teensy/loader_cli.html"
-  head "https://github.com/PaulStoffregen/teensy_loader_cli.git"
+  depends_on "libusb-compat"
+
   url "https://www.pjrc.com/teensy/teensy_loader_cli.2.1.zip"
   sha256 "dafd040d6748b52e0d4a01846d4136f3354ca27ddc36a55ed00d0a0af0902d46"
 
+  head do
+    url "https://github.com/PaulStoffregen/teensy_loader_cli.git"
+    patch do
+      url "https://github.com/rixtox/teensy_loader_cli/commit/747f3a249722eb07b494b2db8c7f872e603448f3.diff"
+      sha256 "be47f352d6f54a88db0bc8a88062f9ec470cab2c675655cfec14a9c238492677"
+    end
+  end
+
   def install
-    ENV["OS"] = "MACOSX"
-    ENV["SDK"] = MacOS.sdk_path || "/"
+    ENV["OS"] = "LINUX"
     system "make"
     bin.install "teensy_loader_cli"
   end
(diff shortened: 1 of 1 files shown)


Library/Formula/teensy_loader_cli.rb
```diff
@@ -1,9 +1,10 @@
 class TeensyLoaderCli < Formula
   desc "Command-line integration for Teensy USB development boards"
   homepage "https://www.pjrc.com/teensy/loader_cli.html"
+  depends_on "libusb-compat"
```
dunn: This should be down below the `head` block.

rixtox: but to compile 2.1 with `OS=LINUX` also require libusb

dunn: Oh, not in the head block, just below it and above the install block.  Otherwise `brew audit --strict teensy_loader_cli` will complain.

rixtox: got it. fixed

Library/Formula/teensy_loader_cli.rb
```diff
@@ -12,9 +12,18 @@ class TeensyLoaderCli < Formula
     sha256 "dcd10140babb4d2937ce376c89e9c24a2e8046d2cabdad2cfdbc2542afa14471" => :mavericks
   end
 
+  head do
+    url "https://github.com/PaulStoffregen/teensy_loader_cli.git"
+    patch do
```
DomT4: Please submit this patch upstream (or at least a bug report) and use that link. We don't particularly want to use a commit from an unofficial repo.

bfontaine:
Ping?

rixtox:
@bfontaine Let's wait for response from upstream. https://github.com/PaulStoffregen/teensy_loader_cli/pull/19

rixtox:
As [responded from the upstream](https://github.com/PaulStoffregen/teensy_loader_cli/pull/19#issuecomment-169309996). The patch is just a quick and dirty workaround for the new Mac OS X version, and is using the Linux build codes. Merging the patch to the upstream will break the Linux code. Thus I suggest merge this patch in Homebrew and wait for any solid progress from the upstream to fix the native Mac OS X build.

bfontaine:
@rixtox I’m afraid we won’t accept a patch if upstream doesn’t want to apply it. Use [`inreplace`](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#inreplace) instead.

mikemcquaid:
> @rixtox I’m afraid we won’t accept a patch if upstream doesn’t want to apply it. Use inreplace instead.

Even in that case: I'd rather we didn't add new features that rely on patches (`inreplace` or otherwise) that can't be merged upstream

mikemcquaid:
Fixed in #48252.

PR #45561, opened 2015-11-01
download_strategy: silence curl on Travis CI.

### LINKED ISSUE #44446: qt5: not working properly when installed from within Travis CI
Until a couple of days ago, I could install qt5 within [Travis CI](https://travis-ci.org/) without any problem. However, today, it just keeps failing. I have noticed that a new version of the formula has been released. So, could there be a 'bug' in it that breaks Travis CI?

Basically, Travis CI complains about the log length having exceeded the limit of 4 MB. Checking the Travis CI output when installing qt5, I can see that the 'progress bar' doesn't progress 'normally' and, in fact, flashes all over the place. I am not sure what is going on under the hood, but it's certainly causing Travis CI to 'fail'.

My 'fix' currently consists of redirecting the output to ```/dev/null```, but that's clearly not something I want to have to do...

mikemcquaid:
Use one of the Travis CI default environment variables:
http://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

CC @xu-cheng @scpeters @dunn @agarny

Closes #44446.
Closes https://github.com/travis-ci/travis-ci/issues/4936.

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -251,7 +251,7 @@ def curl(*args)
 
   args = [flags, HOMEBREW_USER_AGENT, *args]
   args << "--verbose" if ENV["HOMEBREW_CURL_VERBOSE"]
-  args << "--silent" unless $stdout.tty?
+  args << "--silent" if !$stdout.tty? || ENV["TRAVIS"]
 
   safe_system curl, *args
 end


xu-cheng:
This logic may be more suitable to put in `curl` method [here](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/utils.rb#L249-L254).

Also I still would like to support a flag (such as `--curl-silent`) and env (such as `HOMEBREW_CURL_SILENT`). The main reason is there are other CI systems than Travis and other cases this function is useful. For the same reason, it makes more sense to check `CI` rather than `TRAVIS`.

mikemcquaid:
Agreed on the placement.

I'm not sure I see the advantage of it being a user option; we've not had any requests for the functionality but just for it to work in Travis. Equally, we've not had problems or reports with the output in Jenkins or other CI systems.

I'm really :-1: on adding new environment variables all over the place to allow configuration when users haven't requested that functionality. It makes it harder to test code and debug issues.

xu-cheng:
> I'm not sure I see the advantage of it being a user option; we've not had any requests for the functionality but just for it to work in Travis. Equally, we've not had problems or reports with the output in Jenkins or other CI systems.

IMO, there isn't much point to fix problem in only one CI system and ignore others. Because this is definitely kinda problem suffered by all CI systems(Some other system like Jenkins may not fail, but it would still create nasty log).

> I'm really :-1: on adding new environment variables all over the place to allow configuration when users haven't requested that functionality. It makes it harder to test code and debug issues.

I agree this in general. But we already have `HOMEBREW_CURL_VERBOSE`. It's kinda fair to support `HOMEBREW_CURL_SILENT` as well. And I don't see it will make harder to test or debug for this specific feature.

mikemcquaid:
> IMO, there isn't much point to fix problem in only one CI system and ignore others.
> Because this is definitely kinda problem suffered by all CI systems

Can you name other CI systems that have problems (i.e. those that use `tty`s and have progress bar issues)? I don't agree it's necessarily a general problem. If we receive reports that it is: that's easy to fix.

> I agree this in general. But we already have HOMEBREW_CURL_VERBOSE. It's kinda fair to support HOMEBREW_CURL_SILENT as well.

I don't see the use-case. `HOMEBREW_CURL_VERBOSE` is used for debugging.

xu-cheng:
> those that use ttys and have progress bar issues

Oh, I miss this part. How about using `CI` instead of `TRAVIS`? This will solve current problem as well as covering possible other case.

mikemcquaid:
I don't think it does; it will cause a behaviour change on CI systems that don't necessarily have issues. We'll agree to disagree here; I'm fixing a specific bug with a fix for that bug. If people report issues with e.g. CircleCI I'm :+1: on making it more general.

xu-cheng:
>  it will cause a behaviour change on CI systems that don't necessarily have issues. 

Just FYI, I don't think this will happen. Either it doesn't use `ttys` in which case nothing changes. Or it does use `ttys` and output logs in which case it requires such fix.

> We'll agree to disagree here

fine by me.

>  I'm :+1: on making it more general.

Personally, I don't like to put CI specific code in core (except `test-bot.rb`). Anyway, it's just a personal favor.

scpeters:
Thanks!

PR #45563, opened 2015-11-01
pull: handle unneeded and disabled bottles.

mikemcquaid:
CC @xu-cheng @bfontaine @DomT4 for review. I've used it a bit and it works well. Would be good to merge ASAP so I can use it.

### DIFF
--- a/Library/Homebrew/cmd/pull.rb
+++ b/Library/Homebrew/cmd/pull.rb
@@ -128,8 +128,17 @@ module Homebrew
         end
       end
 
-      unless ARGV.include? "--bottle"
-        changed_formulae.each do |f|
+      fetch_bottles = false
+      changed_formulae.each do |f|
+        if ARGV.include? "--bottle"
+          if f.bottle
+            fetch_bottles = true
+          elsif f.bottle_disabled?
+            ohai "#{f}: skipping disabled bottle: #{f.bottle_disable_reason}"
+          elsif f.bottle_unneeded?
+            ohai "#{f}: skipping unneeded bottle."
+          end
+        else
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/pull.rb
```diff
@@ -128,8 +128,12 @@ def pull
         end
       end
 
-      unless ARGV.include? "--bottle"
-        changed_formulae.each do |f|
+      fetch_bottle = false
+      changed_formulae.each do |f|
+        if ARGV.include? "--bottle"
+          fetch_bottle ||= !!f.bottle
```
dunn: Since it's explicitly set to false, won't these conditional assignments always be ignored?

mikemcquaid: Nope; `||=` sets it if the value is currently falsey.

dunn: TIL!

mikemcquaid: It's equivalent of `fetch_bottle = fetch_bottle || !!f.bottle`

DomT4:
I'd be tempted to shout at us that the bottle was disabled for x formula in the pull, but I'm not that fussed about it if you're against. Happy otherwise.

mikemcquaid:
@DomT4 Updated. Let me know what you think.

xu-cheng:
:+1:

bfontaine:
I was sleeping at this time but :+1: :wink:

DomT4:
Thanks @mikemcquaid, the changed the language is :100:. Appreciate the tweak!

PR #45566, opened 2015-11-01
El Capitan and Homebrew: update instructions.

### LINKED ISSUE #45387: Obsolete El Capitan System Integrity Protection (SIP) Instructions
The Homebrew installer (https://raw.githubusercontent.com/Homebrew/install/master/install) currently does the following check:

```
"abort <<-EOABORT if !File.directory?(HOMEBREW_PREFIX) && macos_version >= "10.11"
OS X 10.11 & above require an additional step to write into #{HOMEBREW_PREFIX}.
Please follow the steps described here and then run the install script again:
https://git.io/vnC7w
EOABORT"
```

While these directions may have been correct for some range of the El Capitan developer betas, they seem to be incorrect now.

I removed the /usr/local dir, upgraded 10.11.0 -> 10.11.1, and had no trouble running sudo mkdir /usr/local && sudo chflags norestricted /usr/local && sudo chown -R $(whoami):admin /usr/local

There was no need to disable SIP.

mikemcquaid:
These seem to not apply for everyone on 10.11.0 any more (as explained in #45387).

CC @DomT4 @ilovezfs @CamJN

Closes #45387.

### DIFF
--- a/share/doc/homebrew/El_Capitan_and_Homebrew.md
+++ b/share/doc/homebrew/El_Capitan_and_Homebrew.md
@@ -2,9 +2,9 @@
 
 Part of the OS X 10.11/El Capitan changes is something called [System Integrity Protection](https://en.wikipedia.org/wiki/System_Integrity_Protection) or "SIP".
 
-SIP prevents you from writing to many system directories such as `/usr`, `/System` & `/bin`, regardless of whether or not you are root. The Apple keynote is [here](https://developer.apple.com/videos/wwdc/2015/?id=706) if you'd like to learn more.
+SIP prevents you from writing to many system directories such as `/usr`, `/System` & `/bin`, regardless of whether or not you are root. The Apple keynote is [here](https://developer.apple.com/videos/wwdc/2015/?id=706) if you'd like to learn more. As noted in the keynote, Apple is leaving `/usr/local` open for developers to use, so Homebrew can still be used as expected.
 
-One of the implications of SIP is that you cannot simply create `/usr/local` if you have removed it. However, as noted in the keynote, Apple is leaving `/usr/local` open for developers to use, so Homebrew can still be used as expected.
+One of the implications of SIP was that you could not simply create `/usr/local` if you had removed it. This issue was fixed with the `com.apple.pkg.SystemIntegrityProtectionConfig.14U2076` update.
 
 **If you haven't installed Homebrew in `/usr/local` or another system-protected directory, this document does not apply to you.**
 
@@ -17,6 +17,13 @@ sudo chown -R $(whoami):admin /usr/local
 ```
 
 ## If `/usr/local` does not exist:
+First, try to create `/usr/local` the normal way:
+
+```bash
+  sudo mkdir /usr/local && sudo chflags norestricted /usr/local && sudo chown -R $(whoami):admin /usr/local
+```
+
+If you see permission issues instead try:
 
 * Reboot into Recovery mode (Hold Cmd+R on boot) & access the Terminal.
 * In that terminal run:


share/doc/homebrew/El_Capitan_and_Homebrew.md
```diff
@@ -4,7 +4,7 @@ Part of the OS X 10.11/El Capitan changes is something called [System Integrity
 
 SIP prevents you from writing to many system directories such as `/usr`, `/System` & `/bin`, regardless of whether or not you are root. The Apple keynote is [here](https://developer.apple.com/videos/wwdc/2015/?id=706) if you'd like to learn more.
 
-One of the implications of SIP is that you cannot simply create `/usr/local` if you have removed it. However, as noted in the keynote, Apple is leaving `/usr/local` open for developers to use, so Homebrew can still be used as expected.
+One of the implications of SIP was that you could not simply create `/usr/local` if you had removed it. This issue was fixed with the OS X 10.11.0 release. However, as noted in the keynote, Apple is leaving `/usr/local` open for developers to use, so Homebrew can still be used as expected.
```
DomT4: Dom is a nitpicker alert: Is `10.11.0` a thing, technically? It's just `10.11` right?

mikemcquaid: Yeh, can maybe be `10.11 stable release`?

DomT4:
I'd like to know in which circumstances this is applicable and isn't. The other thread as thrown up as many questions as answers, as grateful as I am for the digging there. I'm fine with suggesting people try it first though and then falling back on the disable if they can't run the command - There's no obvious downside to them at least trying the simple way first.

share/doc/homebrew/El_Capitan_and_Homebrew.md
DomT4: Works for me. I don't know if the change was snuck in as part of the final release or as a security tweak afterwards, but the difference isn't important to people who will read this document.

mikemcquaid:
> I'd like to know in which circumstances this is applicable and isn't. The other thread has thrown up as many questions as answers, as grateful as I am for the digging there. I'm fine with suggesting people try it first though and then falling back on the disable if they can't run the command - There's no obvious downside to them at least trying the simple way first.

It now seems it should only apply to OS X 10.11 if you installed particular betas and upgraded from them rather than doing a clean install. It seems it also doesn't apply to anyone who has a 10.11.1 install. At this point I'm hoping to avoid scaring people doing fresh 10.11 installs.

share/doc/homebrew/El_Capitan_and_Homebrew.md
ilovezfs: This is actually false. 10.11.0 full installer and 10.11.1 full installer both install version 10.0 of the compatibility bundle. Maybe 10.11.2 will come with 12.0 preinstalled, but that is not yet the case.

At least for now, [com.apple.pkg.SystemIntegrityProtectionConfig.14U2076] (http://swcdn.apple.com/content/downloads/53/38/031-40358/y7sn2l7yvfxb73myuiufqin8fp2sqyetvl/SystemIntegrityProtectionConfig.pkg) must have been installed one or another, regardless of whether it happened automatically or manually.

mikemcquaid: Good to know, thanks. Will update accordingly.

ilovezfs:
@mikemcquaid @DomT4 That had been conjecture on my part, and has been disproven in the other thread. It is quite clear any 10.11.0 or 10.11.1 system that has com.apple.pkg.SystemIntegrityProtectionConfig.14U2076 will not have the problem.

There are circumstances under which software updates (including this particular one) can fail, but that is far beyond Homebrew's scope, and has no correlation with prior installed betas or developer version nor with clean installs vs. upgrade installs.

mikemcquaid:
Thanks for the feedback everyone. Merged given the extra information given here.

PR #45567, opened 2015-11-01
Stricter elisp handling

dunn:
Require that the subdirectory in site-lisp match the formula name
exactly.  This lets us provide better information in the caveats and
will make it easier for helper methods to write to the correct
location (as in in Homebrew/homebrew-emacs#13).

Also define `elisp` and `opt_elisp`. If this part is too niche I can just do it in the Emacs tap.

### DIFF
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -143,12 +143,7 @@ class Caveats
     if keg && keg.elisp_installed?
       <<-EOS.undent
         Emacs Lisp files have been installed to:
-        #{HOMEBREW_PREFIX}/share/emacs/site-lisp/
-
-        Add the following to your init file to have packages installed by
-        Homebrew added to your load-path:
-        (let ((default-directory "#{HOMEBREW_PREFIX}/share/emacs/site-lisp/"))
-          (normal-top-level-add-subdirs-to-load-path))
+        #{HOMEBREW_PREFIX}/share/emacs/site-lisp/#{f.name}
       EOS
     end
   end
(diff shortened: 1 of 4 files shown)


dunn:
cc @MerelyAPseudonym

mikemcquaid:
:+1:

Library/Homebrew/caveats.rb
```diff
@@ -143,12 +143,7 @@ def elisp_caveats
     if keg && keg.elisp_installed?
       <<-EOS.undent
         Emacs Lisp files have been installed to:
-        #{HOMEBREW_PREFIX}/share/emacs/site-lisp/
-
-        Add the following to your init file to have packages installed by
```
dunn: Removing this since it's not really Homebrew-specific; it was there to help with the transition from putting everything directly in `site-lisp/` to using subdirectories.

dunn:
Travis failure:
```
==> brew tests
ERROR:  Could not find a valid gem 'bundler' (>= 0), here is why:
          Unable to download data from https://rubygems.org/ - Errno::ETIMEDOUT: Operation timed out - connect(2) (https://rubygems.org/latest_specs.4.8.gz)

**Comment by mikemcquaid** [comment]
Try to retry the build.

File: Library/Homebrew/keg.rb [hunk 2]
```diff
@@ -263,7 +263,7 @@ def app_installed?
   end
 
   def elisp_installed?
-    Dir["#{path}/share/emacs/site-lisp/**/*.el"].any?
+    (path/"share/emacs/site-lisp"/name).children.any? { |f| %w[.el .elc].include? f.extname }
```
dunn: Tweaked to work if only compiled files are installed.

PR #45568, opened 2015-11-01
class-dump: cap stable on Mavericks, fix head.

### LINKED ISSUE #42384: class-dump failed to build on 10.11
https://gist.github.com/4cd00f2a8da4b4bf8b6c

DomT4:
This formula is problematic. If upstream isn't able to cut a new release soon we may end up boneyarding it.

Closes #42384.


mikemcquaid:
:+1:

DomT4:
Will fail the audit for obvious reasons. Willing to give it a *bit* of time given it uses `libcrypto` rather than `libssl` *and* upstream have mostly fixed the problem pending merge of Andrew's PR and a new release but if there's nothing by the end of the year I'd be tempted to set the formula on :fire:.

PR #45301, opened 2015-10-24
Kotlin 1.0.0 beta 1038

fboldog:

PR #45579, opened 2015-11-01
Upgrade Kotlin to 1.0.0-beta-1038

windoze:
Upgrade Kotlin to 1.0.0-beta-1038

### DIFF
--- a/Library/Formula/kotlin-compiler.rb
+++ b/Library/Formula/kotlin-compiler.rb
@@ -1,8 +1,9 @@
 class KotlinCompiler < Formula
   desc "Statically typed programming language for the JVM"
   homepage "https://kotlinlang.org/"
-  url "https://github.com/JetBrains/kotlin/releases/download/build-0.14.449/kotlin-compiler-0.14.449.zip"
-  sha256 "bbb55c84669b6d920dc89beab811b26dfbdbeb95d71e1d6d3384fbfbcd48ec70"
+  url "https://github.com/JetBrains/kotlin/releases/download/build-1.0.0-beta-1038/kotlin-compiler-1.0.0-beta-1038.zip"
+  sha256 "53e1516eb95274ac58b5445d1d8a82d265166d7c5038a878395589604f35275c"
+  version "1.0.0-beta-1038"
 
   bottle :unneeded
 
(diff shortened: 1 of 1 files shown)


Library/Formula/kotlin-compiler.rb
```diff
@@ -1,8 +1,9 @@
 class KotlinCompiler < Formula
   desc "Statically typed programming language for the JVM"
   homepage "https://kotlinlang.org/"
-  url "https://github.com/JetBrains/kotlin/releases/download/build-0.14.449/kotlin-compiler-0.14.449.zip"
-  sha256 "bbb55c84669b6d920dc89beab811b26dfbdbeb95d71e1d6d3384fbfbcd48ec70"
+  url "https://github.com/JetBrains/kotlin/releases/download/build-1.0.0-beta-1038/kotlin-compiler-1.0.0-beta-1038.zip"
+  sha256 "53e1516eb95274ac58b5445d1d8a82d265166d7c5038a878395589604f35275c"
+  version "1.0.0-beta-1038"
```
mikemcquaid: This should be a stable release. You could add this version in a `devel do` block instead.

DomT4:
Closes https://github.com/Homebrew/homebrew/pull/45301, if fixed & merged.

Library/Formula/kotlin-compiler.rb
windoze: Indeed, but 1.0.0-beta-1038 is exactly the successor of M14(0.14.449), I don't see reason to mark one milestone build as "stable" and another one as "devel".

DomT4: If upstream considers it stable they should label it so. Otherwise it should be in a `devel` block, given the `beta` tag.

windoze: That means M14 should also be tagged as "devel", there is no "stable" Kotlin exists so far.

DomT4:
@mikemcquaid OP may actually be right here. The [website](http://kotlinlang.org/) is pointing people towards the `1.0.0-beta-1038` release :confused:.

mikemcquaid:
Yeh, I guess we should just upgrade as-is. Annoying though!

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `kotlin-compiler 0.14.449`.

PR #45584, opened 2015-11-01
libressl: add vuln fix from upstream

DomT4:
They pushed out a 2.2.4 release to fix the problem but no 2.3.1 release yet. Have given it a couple of weeks but not overly happy to leave it unpatched much longer. Gentoo have already [applied the change](https://github.com/gentoo/libressl/commit/9ebb9135d8e8f864dd9504cb0cc777c67a74a53b).


mikemcquaid:
:+1:

PR #45585, opened 2015-11-01
mutt: gettext dep, optional libidn

dunn:
Currently a "fatal error" is being thrown during `make` if `gettext` isn't available, which for some reason doesn't stop the build: https://gist.github.com/dunn/2e653faa14ea5e0ab538#file-02-make-L30-L32

Mutt is distributed with a vendored gettext so this seems like a legitimate dependency: https://bitbucket.org/mutt/mutt/src/17991330c086d6b0de56ff091e449b220e633063/INSTALL?at=default&fileviewer=file-view-default#INSTALL-121:126

### DIFF
--- a/Library/Formula/mutt.rb
+++ b/Library/Formula/mutt.rb
@@ -48,8 +48,10 @@ class Mutt < Formula
 
   depends_on "openssl"
   depends_on "tokyo-cabinet"
-  depends_on "s-lang" => :optional
+  depends_on "gettext" => :optional
   depends_on "gpgme" => :optional
+  depends_on "libidn" => :optional
+  depends_on "s-lang" => :optional
 
   # original source for this went missing, patch sourced from Arch at
   # https://aur.archlinux.org/packages/mutt-ignore-thread/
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

chdiza:
Did you try adding the `--with-included-gettext` flag to the configure arguments?  It didn't look like it in the gist.

chdiza:
Also, it seems to me that if that flag for some reason doesn't help, then rather than saddle mutt with a heavy dep, we could (a) get upstream to fix this for OS X, or even better, (b) pass `--disable-NLS` among the configure arguments.

chdiza:
Whoops, I called (b) above the "even better" option, when (a) is the one that's even better.

dunn:
> Did you try adding the --with-included-gettext flag to the configure arguments? It didn't look like it in the gist.

I didn't; I didn't think about the size of the dependency.  But I tried it and get the same error, and I don't see the header included in the stable tarball. 

Are we really better off with `--disable-NLS`?

chdiza:
> Are we really better off with --disable-NLS?

As a default, overridable with `--with-nls` or some such, yes.

dunn:
Alright. I might try to get in touch with upstream first.

dunn:
http://dev.mutt.org/trac/ticket/3789

dunn:
Switched to @chdiza's suggested modification until I hear from upstream; will merge unless anyone objects.

PR #45591, opened 2015-11-01
libsodium 1.0.6

DomT4:


DomT4:
Colour me an idiot - I missed the dylib version change locally.

mikemcquaid:
:+1: when green.

DomT4:
Bumped all the things, also wrote `minisign` a proper test since the one it had was terrible. Should pass now, presuming we don't ram the timeout. Shouldn't do.

PR #45595, opened 2015-11-02
influxdb: Fix Too Many Open Files Error

felixbuenemann:
This raises the softlimit for open files in the plist from the MacOS X default for launchd agents of 256 open files to 10240 (which is the default limit returned by `sysctl kern.maxfilesperproc`).

**For comparison:**
The default softlimit for the login shell is 7168 files (see `ulimit -n` and `launchctl limit` for launchd).

Without this change influxdb runs out of open file handles very quickly and stops accepting connections, even during moderate use.


felixbuenemann:
Is it correct to bump the revision for this change?

mikemcquaid:

PR #45597, opened 2015-11-02
agda: fix HEAD build, install Emacs Lisp

dunn:
I should've caught this in the original PR, but this way I get to use the new `elisp` helper so ¯\\_(シ)_/¯

### DIFF
--- a/Library/Formula/agda.rb
+++ b/Library/Formula/agda.rb
@@ -2,11 +2,17 @@ require "language/haskell"
 
 class Agda < Formula
   include Language::Haskell::Cabal
-
   desc "Dependently typed functional programming language"
   homepage "http://wiki.portal.chalmers.se/agda/"
-  url "https://github.com/agda/agda/archive/2.4.2.4.tar.gz"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

mietek:
Should the `resource` section be moved into the `stable` section, as well?

dunn:
I just checked and `brew install --HEAD adga` uses the resource defined in the `head do` block, so it's not breaking anything.

Library/Formula/agda.rb
```diff
@@ -94,6 +96,7 @@ def install
     # compile the included Emacs mode
     if build.with? "emacs"
       system bin/"agda-mode", "compile"
+      elisp.install Dir["#{share}/*/Agda-#{version}/emacs-mode/*"]
```
mietek: Has this change been tested?

dunn: I tested it locally, yeah.

mietek: What does Emacs tell you when you visit an Agda source file?

Issue #45607, opened 2015-11-02
Possible way to handle sandbox issues for Postgres's plugins


xu-cheng:
As we can seen in https://github.com/Homebrew/homebrew/pull/41962 and many others PRs, all of Postgres's plugins are broken under sandbox. Moreover, this means all of them are broken during `upgrade/unlink/link/switch` etc.

Considering the amount of plugins for Postgres, vending all of them will soon become unscalable. However, until it's fixed/supported by upstream (See https://github.com/Homebrew/homebrew/issues/10247), Postgres is inherently hostile to Homebrew-style sandboxing where several components are symlinked into a common prefix.

Since there isn't any perfect solution, we may will just accept some hacking middle ground. AFAIK, NixOS handles this by copying all of binaries directly to common prefix, hence breaking its symlink sandbox as well. We may take some similar approach:
* Compile Postgres as usual.
* Move all of binaries in `prefix/bin` to `prefix/libexec/bin-backup`.
* Hard link binaries `prefix/libexec/bin-backup` to `HOMEBREW_PREFIX/bin` during `post_install`.

Clearly, it's still breaking our symlink system. But at least, it can work under sandbox.

Any objection/suggestion/commments? OR should we just vendor all of them inside one mega formula?

cc @mikemcquaid @DomT4

mikemcquaid:
> Any objection/suggestion/commments? OR should we just vendor all of them inside one mega formula?

I think we should do that.

xu-cheng:
Found a new way to fix this, see https://github.com/Homebrew/homebrew/pull/45611#issuecomment-154383042

PR #45611, opened 2015-11-02
Postgres Sandbox Fix

xu-cheng:
See #45607 for explanation.

cc @DomT4 @cosmo0920

### DIFF
--- a/Library/Formula/pgroonga.rb
+++ b/Library/Formula/pgroonga.rb
@@ -1,0 +1,36 @@
+class Pgroonga < Formula
+  desc "PostgreSQL plugin to use Groonga as index"
+  homepage "https://pgroonga.github.io/"
+  url "http://packages.groonga.org/source/pgroonga/pgroonga-1.0.0.tar.gz"
+  sha256 "854d66bdc79e7cfceb29ae8706556a3ad1c8b32ac4535939d296966e3b19fd3e"
+
+  depends_on "pkg-config" => :build
+  depends_on "groonga"
+  depends_on "postgresql"
+
+  def install
+    system "make"
+    mkdir "stage"
+    system "make", "install", "DESTDIR=#{buildpath}/stage"
+
+    lib.install Dir["stage/**/lib/*"]
+    (share/"postgresql/extension").install Dir["stage/**/share/postgresql/extension/*"]
+  end
+
+  test do
+    pg_bin = Formula["postgresql"].opt_bin
+    pg_port = "55561"
+    system "#{pg_bin}/initdb", testpath/"test"
+    pid = fork { exec "#{pg_bin}/postgres", "-D", testpath/"test", "-p", pg_port }
+
+    begin
+      sleep 2
+      system "#{pg_bin}/createdb", "-p", pg_port
+      system "#{pg_bin}/psql", "-p", pg_port, "--command", "CREATE DATABASE test;"
+      system "#{pg_bin}/psql", "-p", pg_port, "-d", "test", "--command", "CREATE EXTENSION pgroonga;"
+    ensure
+      Process.kill 9, pid
+      Process.wait pid
+    end
+  end
+end
--- a/Library/Formula/postgresql.rb
+++ b/Library/Formula/postgresql.rb
@@ -1,6 +1,7 @@
 class Postgresql < Formula
   desc "Object-relational database system"
   homepage "https://www.postgresql.org/"
+  revision 1
 
   stable do
     url "https://ftp.postgresql.org/pub/source/v9.4.5/postgresql-9.4.5.tar.bz2"
@@ -50,7 +51,9 @@ class Postgresql < Formula
     args = %W[
       --disable-debug
       --prefix=#{prefix}
-      --datadir=#{share}/#{name}
+      --datadir=#{HOMEBREW_PREFIX}/share/postgresql
+      --libdir=#{HOMEBREW_PREFIX}/lib
+      --sysconfdir=#{etc}
       --docdir=#{doc}
       --enable-thread-safety
       --with-bonjour
@@ -83,7 +86,10 @@ class Postgresql < Formula
     end
 
     system "./configure", *args
-    system "make", "install-world"
+    system "make"
+    system "make", "install-world", "datadir=#{pkgshare}",
+                                    "libdir=#{lib}",
+                                    "pkglibdir=#{lib}/postgresql"
   end
 
   def post_install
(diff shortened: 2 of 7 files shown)


Library/Formula/postgresql.rb
```diff
@@ -96,18 +86,17 @@ def install
...
-      end
-    end
+    # move all of binaries to libexec/bin
+    (libexec/"bin").install Dir["#{bin}/*"]
   end
 
   def post_install
+    # hardlink binaries in libexec/bin to HOMEBREW_PREFIX
+    # https://github.com/Homebrew/homebrew/issues/45607
+    ln Dir["#{libexec}/bin/*"], HOMEBREW_PREFIX/"bin", :force => true
```
mikemcquaid: Does this need to be done for all binaries or is there a subset we can use for e.g. knowing the extensions path?

Library/Formula/pgroonga.rb
```diff
@@ -0,0 +1,39 @@
...
+  url "http://packages.groonga.org/source/pgroonga/pgroonga-1.0.0.tar.gz"
+  sha256 "854d66bdc79e7cfceb29ae8706556a3ad1c8b32ac4535939d296966e3b19fd3e"
+
+  depends_on "pkg-config" => :build
+  depends_on "groonga"
+  depends_on "postgresql"
+
+  def install
+    # To find HOMEBREW_PREFIX/bin/pg_config
+    ENV.prepend_path "PATH", HOMEBREW_PREFIX/"bin"
```
DomT4: Appreciate it's an important part of your mechanism here, but I'm quite uncomfortable with adding the entire `HOMEBREW_PREFIX` bin back into the `$PATH`. It'd be fine for producing bottles but on local machines where folks have other bits and pieces installed the chances of scooping something out of the `$PATH` that shouldn't be scooped is very high :confused:.

Would using the path to Postgresql's `libexec/bin` not do the trick at all? I presume not, if you've chosen to use `HOMEBREW_PREFIX` here.

Library/Formula/pgroonga.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  test do
+    system "#{HOMEBREW_PREFIX}/bin/initdb", testpath/"test"
+    pid = fork { exec "#{HOMEBREW_PREFIX}/bin/postgres", "-D", testpath/"test" }
+
+    begin
+      sleep 2
+      system "#{HOMEBREW_PREFIX}/bin/createdb"
+      system "#{HOMEBREW_PREFIX}/bin/psql", "--command", "CREATE DATABASE pgroonga_test;"
+      system "#{HOMEBREW_PREFIX}/bin/psql", "-d", "pgroonga_test", "--command", "CREATE EXTENSION pgroonga;"
```
cosmo0920: These commands seems to be good test case.
Great! :smile_cat:

Library/Formula/postgresql.rb
xu-cheng: > Does this need to be done for all binaries

Yes, they all share same code on resolving lib path.

Library/Formula/pgroonga.rb
xu-cheng: Unfortunately, we have to use binary in `HOMEBREW_PREFIX` due to its way of resolving lib path.

cosmo0920: I have no objection to use `HOMEBREW_PREFIX` for now.

DomT4: I still think it's a bad idea and should be avoided at most costs. If they need to share a common permanent prefix I'd almost rather dump `postgresql` related binaries and plugins in `HOMEBREW_PREFIX/var/postgresql_bin/*` and then hardlink or copy over them with the newer versions `postinstall` as necessary.

People would have to add that directory to their `$PATH` but I think it's less likely to randomly break than circumventing a fairly big part of superenv.

xu-cheng: You can't do this either. In which case, binaries won't be able to find supported libraries and other resources files. And also pretty much breaking every Postgres users' system. 

Currently, there is less you can do regarding it.

DomT4: Welp. I'm starting to feel more sympathetic to just having a ridiculously big `postgresql` formula, or even doing something like allowing formula which define a `postgresql` dependency to write to a specific prefixes inside the `postgresql` keg :/.

xu-cheng: The problem for using one mega formula is it doesn't scale well considering so many plugins out there. At the same time, users won't get upgrading notice for each plugin formulae. And final but not least, I feel a little uncomfortable to mix the package with multiple different sources and authors. 

About allowing to prefix, I would really like to avoid it. It's very fragile. For example, whenever Postgres got upgraded or user doing reinstall, all of plugins have to been manually reinstalled. Deep down, current problem has little thing to do with sandbox, as it has already been broken in too many ways.

mikemcquaid: Does this have to be done for every binary or can it be done for a subset?

> even doing something like allowing formula which define a postgresql dependency to write to a specific prefixes inside the postgresql keg :/.

This breaks on postgres upgrades.

mikemcquaid: > The problem for using one mega formula is it doesn't scale well considering so many plugins out there. At the same time, users won't get upgrading notice for each plugin formulae.

This also applies to `nginx`. Perhaps we're better to just bundle the most popular plugins (e.g. those already in core and stuff like `postgis`) and move the rest to a tap.

xu-cheng: > This also applies to nginx. Perhaps we're better to just bundle the most popular plugins (e.g. those already in core and stuff like postgis) and move the rest to a tap.

While, `nginx` doesn't have any problem on prefix sandbox. Currently we have two choice:
* Bundle everything in one formula
* Like this PR, hacking using hard link so plugins can be as individual formulae.

So should I merge this PR?

DomT4: Still think bypassing superenv on the bin is going to cause a bunch of problems for end users, but if you're happy to deal with those if they come up I won't poke you about it yet again.

mikemcquaid: I don't agree. I think we should bundle everything in one formulae if we can't convince `postgresql` to somehow use a shared plugin directory in `share` or `var`.

xu-cheng: @DomT4 I may find a way to work around superenv issue.

cosmo0920: IMO, I like dividing formulae instead of all in one formulae.
Because optional resourced extensions are not handled during `upgrade/unlink/link/switch`.

In fact, I've confused that `PGroonga` does not to be handled when upgrading postgresql.

xu-cheng:
Updated to a new approach: Step 1, make `libexec/share/postgresql` as its sharedir. Step 2, symlink `HOMEBREW_PREFIX/share/postgresql` back to `libexec/share/postgresql`. Problem solved.

mikemcquaid:
Looks great. If CI is :green_apple:: :shipit:

ivar:
thanks for your working fixing the @xu-cheng - much appreciated!

PR #45617, opened 2015-11-02
jenv: change order of caveats

hugocf:
Suggest to set the JENV_ROOT *before* setting the shims, so the `init` doesn‘t create the `~/.jenv` folder in the first place.

### DIFF
--- a/Library/Formula/jenv.rb
+++ b/Library/Formula/jenv.rb
@@ -13,11 +13,11 @@ class Jenv < Formula
    end
 
   def caveats; <<-EOS.undent
-     To enable shims and autocompletion add to your profile:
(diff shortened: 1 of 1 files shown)


Library/Formula/jenv.rb
```diff
@@ -13,11 +13,11 @@ def install
    end
 
   def caveats; <<-EOS.undent
-     To enable shims and autocompletion add to your profile:
-       if which jenv > /dev/null; then eval "$(jenv init -)"; fi
-
      To use Homebrew's directories rather than ~/.jenv add to your profile:
        export JENV_ROOT=#{var}/jenv
+
```
DomT4: There be whitespace on this line.

mikemcquaid:

PR #45626, opened 2015-11-02
confuse 2.8

fhunleth:
Bump the version of libconfuse and also update the website URL. The nongnu.org site hasn't been updated in almost 2 years, and the libconfuse maintainers are exclusively using github now.


dunn:
We'd 💙 if you could add a quick test while you're at it!

fhunleth:
Good point. I just added a simple test like you suggested and also a compile-time call to "make check". It turns out that "make check" is pretty quick for libconfuse, so I don't think that it will slow any builds down.

mikemcquaid:

PR #45627, opened 2015-11-02
cjdns: Update to 17.1

cjdelisle:
Sorry about the update just after the other one, somebody made a PR which broke part of permissions dropping and I accidentally pulled it into 17.


dunn:
Thanks for keeping Homebrew up-to-date! Merged in a20c9b082d376494e50b443db2e6f06ed5a2bb81.

PR #45633, opened 2015-11-03
unicorn 0.9

shaunwheelhouse:
I could't find a formula for [Unicorn Engine](http://unicorn-engine.org) so I wrote one. I'm not sure how stable the underlying build system's interface is and I didn't bolt the unit and regression tests on, they're coupled with the project build and I couldn't be bothered separating them.

How do you like to handle switches for debug builds? At the moment I'm using --with-debug.

### DIFF
--- a/Library/Formula/unicorn.rb
+++ b/Library/Formula/unicorn.rb
@@ -5,34 +5,21 @@ class Unicorn < Formula
   sha256 "1ca03b1c8f6360335567b528210713461e839d47c4eb7c676ba3aa4f72b8cf10"
   head "https://github.com/unicorn-engine/unicorn.git"
 
+  option "with-all", "Build with support for ARM64, Motorola 64k, PowerPC and "\
+    "SPARC"
   option "with-debug", "Create a debug build"
-  option "without-x86", "Build without x86 support"
-  option "without-x86_64", "Build without x86_64 support"
-  option "without-static", "Don't build static libraries"
-  option "with-all", "Build with complete architectural support"
-  option "with-arm", "Build with ARM support"
-  option "with-aarch64", "Build with ARM64 support"
-  option "with-m68k", "Build with Motorola 68000 support"
-  option "with-mips", "Build with MIPS support"
-  option "with-ppc", "Build with PowerPC support"
-  option "with-shared", "Build with shared library support"
-  option "with-sparc", "Build with SPARC support"
+  option "with-shared", "Build shared libraries"
 
   depends_on "capstone"
   depends_on "glib"
   depends_on "pkg-config" => :build
-  depends_on "python"
 
   def install
-    archs = []
-    %w[x86 x86_64 arm aarch64 m64k mips ppc sparc].each do |arch|
-      if build.with?("all") || build.with?(arch)
-        archs << arch
-      end
-    end
+    archs  = %w[x86 x86_64 arm mips]
+    archs += %w[aarch64 m64k ppc sparc] if build.with?("all")
     ENV["PREFIX"] = prefix
     ENV["UNICORN_ARCHS"] = archs.join " "
-    %w[static shared debug].each do |condition|
+    %w[shared debug].each do |condition|
       ENV["UNICORN_#{condition.upcase}"] = build.with?(condition) ? "yes" : "no"
     end
     system "make", "install"
--- a/Library/Formula/unicorn.rb
+++ b/Library/Formula/unicorn.rb
@@ -1,16 +1,11 @@
 class Unicorn < Formula
-  desc "A lightweight multi-platform, multi-architecture CPU emulation framework"
-  head "https://github.com/unicorn-engine/unicorn.git"
+  desc "Lightweight multi-architecture CPU emulation framework"
+  homepage "http://www.unicorn-engine.org"
   url "https://github.com/unicorn-engine/unicorn/archive/0.9.tar.gz"
   sha256 "1ca03b1c8f6360335567b528210713461e839d47c4eb7c676ba3aa4f72b8cf10"
-  homepage "http://www.unicorn-engine.org"
+  head "https://github.com/unicorn-engine/unicorn.git"
 
-  depends_on 'capstone'
-  depends_on 'glib'
-  depends_on 'pkg-config' => :build
-  depends_on 'python'
-
-  option "debug", "Create a debug build"
+  option "with-debug", "Create a debug build"
   option "without-x86", "Build without x86 support"
   option "without-x86_64", "Build without x86_64 support"
   option "without-static", "Don't build static libraries"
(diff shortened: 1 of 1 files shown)


Library/Formula/unicorn.rb
```diff
@@ -0,0 +1,88 @@
...
+  depends_on "glib"
+  depends_on "pkg-config" => :build
+  depends_on "python"
+
+  def install
+    archs = []
+    ["x86", "x86_64"].each do |arch|
+      archs << arch if build.with? arch
+    end
+    ["arm", "aarch64", "m64k", "mips", "ppc", "sparc"].each do |arch|
```
bfontaine: Use `%w[]` here and above.

Library/Formula/unicorn.rb
```diff
@@ -0,0 +1,88 @@
...
+      if (build.with? "all") || (build.with? arch)
+        archs << arch
+      end
+    end
+    ENV["PREFIX"] = prefix
+    ENV["UNICORN_ARCHS"] = archs.join " "
+    ["static", "shared", "debug"].each do |condition|
```
bfontaine: Use `%w[]` here too.

Library/Formula/unicorn.rb
```diff
@@ -0,0 +1,88 @@
...
+        uc_reg_write(uc, UC_X86_REG_EDX, &r_edx);
+        err = uc_emu_start(uc, ADDRESS, ADDRESS + sizeof(X86_CODE32) - 1, 0, 0);
+        if (err)
+          return -1;
+        uc_close(uc);
+        puts("Emulation complete.");
+        return 0;
+      }
+    EOS
+    system "cc", *(`pkg-config --libs glib-2.0`.split), "-o", \
```
bfontaine: * Use `ENV.cc` instead of `"cc"`
* You can’t rely on `pkg-config` in the test. See [`tinyxml2.rb`](https://github.com/Homebrew/homebrew/blob/38acad93df2f36d48b808c28698c03942b858122/Library/Formula/tinyxml2.rb#L31) for an example.

Library/Formula/unicorn.rb
```diff
@@ -0,0 +1,88 @@
...
+    ENV["UNICORN_ARCHS"] = archs.join " "
+    ["static", "shared", "debug"].each do |condition|
+      ENV["UNICORN_#{condition.upcase}"] = build.with?(condition) ? "yes" : "no"
+    end
+    system "make", "install"
+  end
+
+  test do
+    source = "test1.c"
+    executable = "test1"
```
bfontaine: Use `source = testpath/"test1.c"` and `executable = testpath/"test1"` since you never use them without this prefix.

Library/Formula/unicorn.rb
```diff
@@ -0,0 +1,88 @@
...
+  depends_on "pkg-config" => :build
+  depends_on "python"
+
+  def install
+    archs = []
+    ["x86", "x86_64"].each do |arch|
+      archs << arch if build.with? arch
+    end
+    ["arm", "aarch64", "m64k", "mips", "ppc", "sparc"].each do |arch|
+      if (build.with? "all") || (build.with? arch)
```
bfontaine: Use `if build.with?("all") || build.with?(arch)`.

shaunwheelhouse:
Thanks, I made the changes. The build command's still not perfect but neither is the test, I'll put something better together another day. Let me know if there's anything else.

Library/Formula/unicorn.rb
```diff
@@ -0,0 +1,93 @@
...
+  option "without-x86_64", "Build without x86_64 support"
+  option "without-static", "Don't build static libraries"
+  option "with-all", "Build with complete architectural support"
+  option "with-arm", "Build with ARM support"
+  option "with-aarch64", "Build with ARM64 support"
+  option "with-m68k", "Build with Motorola 68000 support"
+  option "with-mips", "Build with MIPS support"
+  option "with-ppc", "Build with PowerPC support"
+  option "with-shared", "Build with shared library support"
+  option "with-sparc", "Build with SPARC support"
```
mikemcquaid: I don't think we need all these options. Which do you personally need/use?

mikemcquaid: Definitely don't need the `without`s here. Maybe just have `with-all`?

Library/Formula/unicorn.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  depends_on "capstone"
+  depends_on "glib"
+  depends_on "pkg-config" => :build
+  depends_on "python"
```
mikemcquaid: Can this not use the system Python?

shaunwheelhouse:
@mikemcquaid Maybe I got a bit carried away with that. My rationale was just that different people may want different combinations of architectural support and that x86/x86-64 made for reasonable defaults on OS X. If it's too verbose we could speculate that x86, x86-64, ARM and MIPS *might* be representative of the most common configurations and throw all of the rest under --with-all, but that's probably not fair.

As for Python... my path was broken at the time and for some reason I forgot I already had it. I'll update it.

Library/Formula/unicorn.rb
```diff
@@ -0,0 +1,80 @@
...
+  desc "Lightweight multi-architecture CPU emulation framework"
+  homepage "http://www.unicorn-engine.org"
+  url "https://github.com/unicorn-engine/unicorn/archive/0.9.tar.gz"
+  sha256 "1ca03b1c8f6360335567b528210713461e839d47c4eb7c676ba3aa4f72b8cf10"
+  head "https://github.com/unicorn-engine/unicorn.git"
+
+  option "with-all", "Build with support for ARM64, Motorola 64k, PowerPC and "\
+    "SPARC"
+  option "with-debug", "Create a debug build"
+  option "with-shared", "Build shared libraries"
```
mikemcquaid: Why would you not want these by default?

Library/Formula/unicorn.rb
```diff
@@ -0,0 +1,80 @@
...
+  depends_on "pkg-config" => :build
+
+  def install
+    archs  = %w[x86 x86_64 arm mips]
+    archs += %w[aarch64 m64k ppc sparc] if build.with?("all")
+    ENV["PREFIX"] = prefix
+    ENV["UNICORN_ARCHS"] = archs.join " "
+    %w[shared debug].each do |condition|
+      ENV["UNICORN_#{condition.upcase}"] = build.with?(condition) ? "yes" : "no"
+    end
```
mikemcquaid: just hard-coded these values; makes them easier to follow.

mikemcquaid: And use an `if/else` rather than a ternary.

Library/Formula/unicorn.rb
```diff
@@ -0,0 +1,80 @@
...
+    system "make", "install"
+  end
+
+  test do
+    source = testpath/"test1.c"
+    executable = testpath/"test1"
```
mikemcquaid: Just hard-code this value; will be easier to follow.

shaunwheelhouse:
@mikemcquaid Don't mind me, I'm just used to compiling for the minimum of what I need. Updated.

Library/Formula/unicorn.rb
```diff
@@ -0,0 +1,78 @@
+class Unicorn < Formula
+  desc "Lightweight multi-architecture CPU emulation framework"
+  homepage "http://www.unicorn-engine.org"
+  url "https://github.com/unicorn-engine/unicorn/archive/0.9.tar.gz"
+  sha256 "1ca03b1c8f6360335567b528210713461e839d47c4eb7c676ba3aa4f72b8cf10"
+  head "https://github.com/unicorn-engine/unicorn.git"
+
+  option "with-all", "Build with support for ARM64, Motorola 64k, PowerPC and "\
+    "SPARC"
```
mikemcquaid: I'd probably just put this all on one line.

mikemcquaid: Actually: never mind.

mikemcquaid:

For future reference the preferred commit message format for new formulae is `unicorn 0.9 (new formula)`.

aquynh:
thank you for doing this Brew package! i wanted to submit, but could not find time.

@shaunwheelhouse: do you want me to credit you in https://github.com/unicorn-engine/unicorn/blob/master/CREDITS.TXT?

shaunwheelhouse:
@mikemcquaid No worries! I'll keep that in mind next time.

@aquynh No, thank you! Unicorn is awesome. If it's no trouble that would be great, thanks. :smile:

I'll try and write a better formula test in a couple of days that does some conditional feature checks. If you have any suggestions or preferences in that respect you're welcome to email me.

aquynh:
@shaunwheelhouse see https://github.com/unicorn-engine/unicorn/commit/2fe1b31d51a9df03bb9dcf0893aa746ea9b1ed0d, thanks.

also, if you need any change in Unicorn code to make it easier to maintain this formula, please feel free to send PR for that.

shaunwheelhouse:
@aquynh Thanks again, likewise.

Library/Formula/unicorn.rb
```diff
@@ -0,0 +1,78 @@
...
+  option "with-debug", "Create a debug build"
+
+  depends_on "capstone"
```
aquynh: why does this depends on Capstone??

shaunwheelhouse:
@aquynh asdfasdfasdf

I should have been sleeping instead of tinkering. That and Python, what was I thinking?

PR #45634, opened 2015-11-03
macvim: Update homepage from Google Code to Github

jamer:
The old homepage is just a 301 Moved Permanently to here.


bfontaine:
Thank you @jamer! :smile:

PR #45636, opened 2015-11-03
Add Formula ssllabs-scan

rdvn:
This tool is a command-line client for the SSL Labs APIs, designed for automated and/or bulk testing.

### DIFF
--- a/Library/Formula/ssllabs-scan.rb
+++ b/Library/Formula/ssllabs-scan.rb
@@ -13,6 +13,6 @@ class SsllabsScan < Formula
   end
 
   test do
-    system "#{bin}/ssllabs-scan", "-version"
+    system "#{bin}/ssllabs-scan", "-grade", "-quiet", "-usecache", "ssllabs.com"
   end
 end
--- a/Library/Formula/ssllabs-scan.rb
+++ b/Library/Formula/ssllabs-scan.rb
@@ -1,0 +1,18 @@
+class SsllabsScan < Formula
+  desc "This tool is a command-line client for the SSL Labs APIs."
+  homepage "https://github.com/ssllabs/ssllabs-scan/"
+  url "https://github.com/ssllabs/ssllabs-scan/archive/v1.2.0.tar.gz"
+  sha256 "c8becd57a4aa65f5c4c32824e392d8373ae169ff055192107e9411032dfcd017"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    system "make", "build"
+    bin.install "ssllabs-scan"
+  end
+
+  test do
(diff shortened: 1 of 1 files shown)


Library/Formula/ssllabs-scan.rb
```diff
@@ -0,0 +1,18 @@
...
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    system "make", "build"
+    bin.install "ssllabs-scan"
+  end
+
+  test do
+    system "#{bin}/ssllabs-scan", "-version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

rdvn: Writing test could be a little bit fishy, because this is just simple layer to interact with remote API. Everything except ```--help``` and ```--version``` is dependend on remote API calls. What is the correct way to write test in this case?

mikemcquaid: It's OK to depend on remote API calls if they can be interacted with anonymously. If not: let me know and no worries.

rdvn:
I've modified tests to actualy interact with API.

DomT4:
I don't know if we should include a caveat here about if you've installed this formula you've agreed to upstream's [Terms of Use](https://www.ssllabs.com/downloads/Qualys_SSL_Labs_Terms_of_Use.pdf)? It's not *entirely* permissive.

From the README:

> Before you use this tool please review the terms and conditions, which can be found here: https://www.ssllabs.com/about/terms.html

mikemcquaid:
> I don't know if we should include a caveat here about if you've installed this formula you've agreed to upstream's Terms of Use? It's not entirely permissive.

Yeh, I agree :+1:

DomT4:
@rdvn Would you mind adding a caveat? Something like:

```ruby
def caveats; <<-EOS.undent
  By installing this package you agree to the Terms and Conditions defined by Qualys.
  You can find the terms and conditions at this link:
     https://www.ssllabs.com/about/terms.html

  If you do not agree with those you should uninstall the formula.
EOS
end
```

Should go between the `def install` and `test do` blocks.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/e3f3e9c5874aaa995ab447c1100561a68b2e072f. Thank you for your contribution to Homebrew @rdvn; we appreciate it! :smiley_cat:.

Just a tiny note for future; new formula submissions should have their commit message styled:

```
ssllabs-scan 1.2.0 (new formula)
```
Or similar. Thanks!

PR #45637, opened 2015-11-03
test-bot: hard link generated bottle file to HOMEBREW_CACHE

xu-cheng:


mikemcquaid:
If we do this I think it's worth deleting it afterwards on cleanup too.

xu-cheng:
There is already [cleanup code](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/test-bot.rb#L1020-Lundefined) for CI. Do you want to cleanup these files for users as well?

mikemcquaid:
> There is already cleanup code for CI. Do you want to cleanup these files for users as well(which is somehow covered by brew cleanup)?

That's not being run in CI unless the option is passed? As-is this could cause issues if e.g. a future PR tries to download that bottle as a dependency (e.g. without `--force`) and then checksum may be incorrect depending on the bottle upload ordering.

xu-cheng:
Updated.

Library/Homebrew/cmd/test-bot.rb
```diff
@@ -1017,7 +1018,14 @@ def test_bot
...
     end
   ensure
-    HOMEBREW_CACHE.children.each(&:rmtree) if ARGV.include? "--clean-cache"
+    if ARGV.include? "--clean-cache"
+      HOMEBREW_CACHE.children.each(&:rmtree)
+    else
+      Dir.glob("*.bottle*.tar.gz") do |bottle_file|
+        FileUtils.rm_f HOMEBREW_CACHE/bottle_file
+      end
+    end
```
mikemcquaid: I don't think we want this to delete all bottles in this case? That will mean it redownloads bottles used as dependencies that haven't been updated every time. Feels nicer to store and delete the specific file.

xu-cheng: This will only delete new generated bottles. Noticed that it's scanning on current directory not HOMEBREW_CACHE.

mikemcquaid: Aha, I see! Nice work :+1:

mikemcquaid:
:+1:

PR #45638, opened 2015-11-03
Fix error in PostGIS SQL update expression

emilbayes:


Library/Formula/postgis.rb
```diff
@@ -142,7 +142,7 @@ def caveats
       To create a spatially-enabled database, see the documentation:
         http://postgis.net/docs/manual-2.1/postgis_installation.html#create_new_db_extensions
       If you are currently using PostGIS 2.0+, you can go the soft upgrade path:
-        ALTER EXTENSION postgis UPDATE TO "2.1.5";
+        ALTER EXTENSION postgis UPDATE TO "2.1.7";
```
dunn: If you change this line to
```ruby
ALTER EXTENSION postgis UPDATE TO "#{version}";
```
it'll stay up-to-date automatically!

mikemcquaid: 👍

PR #45641, opened 2015-11-03
Adding bugzilla 5.0.1 formula

swarnava:


Library/Formula/bugzilla.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
+    system "make", "install"
+  end
+
+  test do
+    system "false"
```
dunn: We'll need a test; just something to check basic functionality.  [`etcd.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/etcd.rb#L31-L49) and [`consul.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/consul.rb#L188-L194) have good examples of server-based tests.

Library/Formula/bugzilla.rb
```diff
@@ -0,0 +1,22 @@
+class Bugzilla < Formula
+  desc "Bugzilla is server software designed to help you manage software development."
+  homepage "https://www.bugzilla.org"
+  url "https://ftp.mozilla.org/pub/mozilla.org/webtools/bugzilla-5.0.1.tar.gz"
+  version "5.0.1"
```
dunn: Homebrew should be able to scan the version from the URL, so this shouldn't be necessary.

Library/Formula/bugzilla.rb
```diff
@@ -0,0 +1,22 @@
+class Bugzilla < Formula
+  desc "Bugzilla is server software designed to help you manage software development."
```
dunn: The name will be attached to the description so this can just be 
```ruby
desc "Server software designed to help you manage software development."
```

Library/Formula/bugzilla.rb
```diff
@@ -0,0 +1,22 @@
...
+  homepage "https://www.bugzilla.org"
+  url "https://ftp.mozilla.org/pub/mozilla.org/webtools/bugzilla-5.0.1.tar.gz"
+  version "5.0.1"
+  sha256 "f0f85bf405984770d65f60f5a828ea4d173acaca4aa85e4e58b4e88b1ae186a5"
+
+  depends_on :x11
```
dunn: If this dependency can't be made optional, then I'm afraid it'll need to go in our [homebrew/x11](https://github.com/Homebrew/homebrew-x11) tap.

Library/Formula/bugzilla.rb
```diff
@@ -0,0 +1,22 @@
...
+  desc "Bugzilla is server software designed to help you manage software development."
+  homepage "https://www.bugzilla.org"
+  url "https://ftp.mozilla.org/pub/mozilla.org/webtools/bugzilla-5.0.1.tar.gz"
+  version "5.0.1"
+  sha256 "f0f85bf405984770d65f60f5a828ea4d173acaca4aa85e4e58b4e88b1ae186a5"
+
+  depends_on :x11 
+
+  def install
+
```
bfontaine: Please remove this empty line.

Library/Formula/bugzilla.rb
bfontaine: Please also remove the trailing space.

apjanke:
This formula seems to have some deeper issues. There is no `configure` in the Bugzilla distribution. It has a perl-based build system where you run `perl install-module.pl` and `perl checksetup.pl` to build it, and then you manually edit the config file and copy the whole thing somewhere and point a web server like Apache at it. A formula for bugzilla would probably need extensive customization to get running smoothly under homebrew.

Library/Formula/bugzilla.rb
apjanke: Bugzilla has no dependency on `X11`. That `depends_on :x11` may just be left over from an example formula this was based on.

DomT4:

Issue #45432, opened 2015-10-28
brew cleanup -n


Ivoah:
It would be nice if `brew cleanup -n` showed the total size of all files that would be removed at the end of the output.

chdiza:
I have a hunch this would unacceptably slow it down.  It's already been slowed-down by printing a size for each individual item.

Ivoah:
Wouldn't it just have to add all the sizes together? I can't imagine that would take any appreciable amount of time.

bfontaine:
We could review a pull-request for that.

DomT4:
Closing in favour of #45642, Thanks!

PR #45642, opened 2015-11-03
brew cleanup -n: show total disk space which would be cleared

rakkesh:
Addresses the feature request in issue https://github.com/Homebrew/homebrew/issues/45432.

### DIFF
--- a/Library/Homebrew/cmd/cleanup.rb
+++ b/Library/Homebrew/cmd/cleanup.rb
@@ -2,8 +2,15 @@ require "formula"
 require "keg"
 require "bottles"
 require "thread"
+require "utils"
 
 module Homebrew
+  @@disk_cleanup_size = 0
+
+  def update_disk_cleanup_size(path_size)
+    @@disk_cleanup_size += path_size
+  end
+
   def cleanup
     if ARGV.named.empty?
       cleanup_cellar
@@ -15,6 +22,15 @@ module Homebrew
       end
     else
       ARGV.resolved_formulae.each { |f| cleanup_formula(f) }
+    end
+
+    if @@disk_cleanup_size > 0
+      disk_space = disk_usage_readable(@@disk_cleanup_size)
+      if ARGV.dry_run?
+        ohai "This operation would free approximately #{disk_space} of disk space."
+      else
+        ohai "This operation has freed approximately #{disk_space} of disk space."
+      end
     end
   end
 
@@ -108,6 +124,7 @@ module Homebrew
       puts "Removing: #{path}... (#{path.abv})"
       yield
     end
+    update_disk_cleanup_size(path.disk_usage)
   end
 
   def cleanup_lockfiles
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -1,10 +1,50 @@
 require "pathname"
 require "resource"
 require "metafiles"
+require "utils"
+
+module DiskUsageExtension
+  def disk_usage
+    return @disk_usage if @disk_usage
+    compute_disk_usage
+    @disk_usage
+  end
+
+  def file_count
+    return @file_count if @file_count
+    compute_disk_usage
+    @file_count
+  end
+
+  def abv
+    out = ""
+    compute_disk_usage
+    out << "#{number_readable(@file_count)} files, " if @file_count > 1
+    out << "#{disk_usage_readable(@disk_usage)}"
+  end
+
+  private
+
+  def compute_disk_usage
+    if self.directory?
+      @file_count, @disk_usage = [0, 0]
+      self.find.each do |f|
+        if !File.directory?(f) and !f.to_s.match(/\.DS_Store/)
+          @file_count += 1
+          @disk_usage += File.size(f)
+        end
+      end
+    else
+      @file_count, @disk_usage = [1, self.size]
+    end
+  end
+end
 
 # Homebrew extends Ruby's `Pathname` to make our code more readable.
 # @see http://ruby-doc.org/stdlib-1.8.7/libdoc/pathname/rdoc/Pathname.html  Ruby's Pathname API
 class Pathname
+  include DiskUsageExtension
+
   # @private
   BOTTLE_EXTNAME_RX = /(\.[a-z0-9_]+\.bottle\.(\d+\.)?tar\.gz)$/
 
@@ -408,17 +448,6 @@ class Pathname
       filename.chmod 0644
       install(filename)
     end
-  end
-
-  # @private
-  def abv
-    out = ""
-    n = Utils.popen_read("find", expand_path.to_s, "-type", "f", "!", "-name", ".DS_Store").split("\n").size
-    out << "#{n} files, " if n > 1
-    size = Utils.popen_read("/usr/bin/du", "-hs", expand_path.to_s).split("\t")[0]
-    size ||= "0B"
-    out << size.strip
-    out
   end
 
   # We redefine these private methods in order to add the /o modifier to
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -139,6 +139,14 @@ class Keg
     path.abv
   end
 
+  def disk_usage
+    path.disk_usage
+  end
+
+  def file_count
+    path.file_count
+  end
+
   def directory?
     path.directory?
   end
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -548,3 +548,32 @@ module GitHub
     open(uri) { |json| json["private"] }
   end
 end
+
+def disk_usage_readable(size_in_bytes)
+  len = size_in_bytes.to_s.length
+  case
+  when len > 9
+    sym, unit = ["G", 1_073_741_824]
+  when len > 6
+    sym, unit = ["M", 1_048_576]
+  when len > 3
+    sym, unit = ["K", 1_024]
+  else
+    sym, unit = ["B", 1]
+  end
+
+  num = "%.1f" % [size_in_bytes.to_f / unit]
+  # check whether the rounded value has a zero after decimal point,
+  # if true, then display just the integer value.
+  if num.split(".").last.to_i == 0
+    "%d#{sym}" % num.to_i
+  else
+    "#{num}#{sym}"
+  end
+end
+
+def number_readable(number)
+  numstr = number.to_i.to_s
+  (numstr.size - 3).step(1, -3) { |i| numstr.insert(i, ",") }
+  numstr
+end


rakkesh:
I have found this useful, so, tried to make minimal changes to the code to achieve the result.

@bfontaine @xu-cheng please review and provide your feedback.

bfontaine:
I’m :+1: on the feature but I wonder if there’s way to avoid all those parsing operations for each path.

xu-cheng:
Agree that we should reimplement `abv` feature(probably in pure ruby) rather than using parsing.

rakkesh:
I agree, parsing the output of `du -hs` from `path.abv` adds an overhead. On my machine, the time difference of running `brew cleanup -n` with this code is about `0.26 seconds` for about 50 files/directories which total upto `847.3M`.

As per your suggestion, I'll try to implement the calculation of disk size in ruby for `pathname.abv`.

UniqMartin:
Good feature! :+1: An implementation that doesn't require parsing output of external commands, which is prone to break, would be great.

I think it would be nice to show the summary line both for the dry run as well as for the actual cleanup case (with appropriate change of wording, e.g. `This will free up […]` vs. `This has freed up […]`).

rakkesh:
Using `Find.find` and `File.size` in Ruby, the process of counting files, and calculating the size of a directory has certainly improved (> 2x) on my machine.

Snippet from [gist](https://gist.github.com/rakkesh/b4419d38265185b47bd3) referred below,
```rb
Find.find(@path).each do |f|
  if !File.directory?(f) or !f.match(/#{@exclude.join('|')}/)
    @file_count += 1
    @size_in_bytes += File.size(f)
  end
end
```

I am no Ruby programmer but I have tried to come up with an [implementation](https://gist.github.com/rakkesh/b4419d38265185b47bd3). Let me know, if this is any good for general use in `Homebrew`.

mikemcquaid:
@rakkesh Adding that to this PR to replace `Pathname.abv` would be great! We can help you with the Ruby there.

rakkesh:
@mikemcquaid Thank you. Updated the PR with the changes as discussed above by all the participants. Let me know, if anything else needs to be changed.

Library/Homebrew/utils/diskusage.rb
```diff
@@ -0,0 +1,58 @@
...
+        sym, pow = ['M', 2]
+      when len > 3
+        sym, pow = ['K', 1]
+      else
+        sym, pow = ['B', 0]
+      end
+      "%.1f#{sym}" % [number.to_f / (1024 ** pow)]
+    end
+
+    def self.count_readable(number)
```
rakkesh: I guess, this can be done in a better way. I give it a shot again if it is useful at all.

Library/Homebrew/utils/diskusage.rb
```diff
@@ -0,0 +1,58 @@
...
+      len = count.to_s.length
```
rakkesh: ah, typo!

Library/Homebrew/utils/diskusage.rb
```diff
@@ -0,0 +1,58 @@
...
+      when len > 9
+        sym, pow = ['G', 3]
+      when len > 6
+        sym, pow = ['M', 2]
+      when len > 3
+        sym, pow = ['K', 1]
+      else
+        sym, pow = ['B', 0]
+      end
+      "%.1f#{sym}" % [number.to_f / (1024 ** pow)]
```
bfontaine: It might be worth avoiding this `1024 ** pow` and use the full values above, e.g. `sym, unit = ['G', 1_073_741_824]`.

rakkesh: 👍🏽 I'll change it.

Library/Homebrew/utils/diskusage.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def inspect
+      "#{@path} (#{@file_count} files, #{DiskUsage.disk_space_readable(@size_in_bytes)})"
+    end
+
+    def self.disk_space_readable(number)
+      len = number.to_s.length
+      case
+      when len > 9
+        sym, pow = ['G', 3]
```
DomT4: Super nit, but please stick to double quotes here. We've pretty much eradicated single quotes in the core code.

Library/Homebrew/cmd/cleanup.rb
```diff
@@ -12,6 +19,9 @@ def cleanup
       unless ARGV.dry_run?
         cleanup_lockfiles
         rm_DS_Store
+        puts "==> This operation has cleared #{Utils::DiskUsage.disk_space_readable(@@cleanup_size_in_bytes)} of disk space."
```
bfontaine: Use `ohai "my string"` instead of `puts "==> my string"`.

Library/Homebrew/utils/diskusage.rb
```diff
@@ -0,0 +1,58 @@
...
+        Find.find(@path).each do |f|
+          if !File.directory?(f) and !f.match(/#{@exclude.join('|')}/)
+            @file_count += 1
+            @size_in_bytes += File.size(f)
+          end
+        end
+      rescue Errno::ENOENT
+        raise "DiskUsage: No such path `#{@path}'"
+      rescue
+        raise "DiskUsage: error in calculating disk space!"
```
bfontaine: Don’t use an empty `rescue`; the caller will rescue the exception.

Library/Homebrew/utils/diskusage.rb
```diff
@@ -0,0 +1,57 @@
+require "find"
+
+module Utils
+  class DiskUsage
```
mikemcquaid: I don't think this needs to be a separate class? Could this be part of `Pathname.abv`?

rakkesh: I don't think I follow your thought of replacing `Pathname.abv`, please explain. 

As I see, `Pathname.abv` is just a display method but we need `Fixnum` (disk usage) values for all the `path`s which are to be cleared, so that, a total can be calculated without parsing output of `Pathname.abv`. I'll check, and try to modify `extend/pathname.rb` to accommodate this shortcoming.

mikemcquaid: Yeh, I think we could have a shared method that both `Pathname.abv` and these methods could call. I also think we probably could keep it as a short method inside `utils.rb` rather than requiring a separate class.

rakkesh: Yes, we could have a common method in `utils.rb` but we'll end up calling that method twice - `pathname.abv` for displaying, and in `cleanup.rb` to get a `disk space usage` value. Or just once if we have a display-cum-update method in `cleanup.rb` without invoking `pathname.abv`.

Here is another [attempt](https://github.com/rakkesh/homebrew/commit/81db726f7dd592667e60ca24099afe5f8a8386ab) to integrate better into `Homebrew` core. I have made changes to `Pathname` to hold `directory size` value, which gets updated when `abv` is invoked but I wonder if this can be done differently.

mikemcquaid: > Yes, we could have a common method in utils.rb but we'll end up calling that method twice - pathname.abv for displaying, and in cleanup.rb to get a disk space usage value.

That sounds good :+1:

rakkesh: Wouldn't that create an overhead of calculating disk space twice?

mikemcquaid: You could make a cache in the method using an instance or class variable.

rakkesh: > You could make a cache in the method using an instance or class variable.

Change available on `brew-cleanup` branch in my fork of `Homebrew`. You can view it [here](https://github.com/rakkesh/homebrew/commit/81db726f7dd592667e60ca24099afe5f8a8386ab).

Library/Homebrew/cmd/cleanup.rb
```diff
@@ -2,8 +2,15 @@
 require "keg"
 require "bottles"
 require "thread"
+require "utils/diskusage"
```
xu-cheng: Maybe put this line inside `Library/Homebrew/utils.rb`

Library/Homebrew/cmd/cleanup.rb
```diff
@@ -102,12 +118,14 @@ def cleanup_cache
   end
 
   def cleanup_path(path)
+    du = Utils::DiskUsage.new(path.to_s)
     if ARGV.dry_run?
-      puts "Would remove: #{path} (#{path.abv})"
+      puts "Would remove: #{du.to_s}"
```
xu-cheng: `#{du}` should be enough. `to_s` is unnecessary.

Library/Homebrew/utils/diskusage.rb
```diff
@@ -0,0 +1,57 @@
...
+    attr_reader :size_in_bytes, :file_count
+
+    def initialize(path, excludes = [])
+      @size_in_bytes = 0
+      @file_count = 0
+      @path = path
+      @excludes = [".DS_Store"] + excludes
+      begin
+        Find.find(@path).each do |f|
+          if !File.directory?(f) and !f.match(/#{@excludes.join("|")}/)
```
xu-cheng: It's better to put `/#{@excludes.join("|")}/` outside of loop for better performance.

I would also be interesting to see some benchmark between ruby version with original shellout approach.

Library/Homebrew/utils/diskusage.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    def to_s
+      out = "#{@path} ("
+      out << "#{number_readable(@file_count)} files, " if @file_count > 1
+      out << "#{disk_space_readable(@size_in_bytes)})"
+    end
+
+    def self.disk_space_readable(size_in_bytes)
+      len = size_in_bytes.to_s.length
```
xu-cheng: Please avoid such string converting. You should compare Fixnum directly.

rakkesh:
@mikemcquaid Thanks for the feedback. As it was pointed out [earlier](https://github.com/Homebrew/homebrew/pull/45642#discussion_r44203059) whether we need a separate class `DiskUsage` in this PR, I have attempted to integrate disk usage into the existing `Pathname` class  here, https://github.com/rakkesh/homebrew/commit/81db726f7dd592667e60ca24099afe5f8a8386ab but has few minor problems. May be you guys can fix it.

Anyway, I'll go ahead make changes as per the latest feedback.

Library/Homebrew/utils/diskusage.rb
```diff
@@ -0,0 +1,57 @@
...
+        sym, unit = ["M", 1_048_576]
+      when len > 3
+        sym, unit = ["K", 1_024]
+      else
+        sym, unit = ["B", 1]
+      end
+      "%.1f#{sym}" % [size_in_bytes.to_f / unit]
+    end
+
+    def disk_space_readable(size_in_bytes)
```
xu-cheng: You can avoid such method duplication

 Instead let's just put `disk_space_readable` and `number_readable` outside of `DiskUsage`. Maybe just put them inside `utils.rb` as standalone methods.

xu-cheng:
> As it was pointed out earlier whether we need a separate class DiskUsage in this PR

I suggest to fix it using below structure.
```ruby
module DiskUsageExtend
  def disk_usage
    @disk_usage ||= compute_disk_usage[0]
  end

  def files_count
     @files_count ||= compute_disk_usage[1]
  end

  def abv
    # call disk_usage and files_count methods.
  end
  
  private

  def compute_disk_usage
    # return tuple (disk_usage, files_count) 
  end
end

class Pathname
  include DiskUsageExtend
end
```

Library/Homebrew/utils/diskusage.rb
```diff
@@ -0,0 +1,57 @@
...
+      self.class.disk_space_readable(size_in_bytes)
+    end
+
+    def self.number_readable(number)
+      number.to_s.reverse.split("").each_slice(3).to_a.map {|x| x.reverse.join}.reverse.join(",")
```
xu-cheng: `number.to_s.chars.to_a.reverse.each_slice(3).map(&:join).join(",").reverse`

UniqMartin: Quite some heavy lifting for what should be a simple string operation. I'd suggest:

```ruby
str = number.to_i.to_s
(str.size - 3).step(1, -3) { |i| str.insert(i, ",") }
str
```

rakkesh: 👍🏽 @UniqMartin that certainly is much faster.

rakkesh:
@xu-cheng your detailed feedback is quite helpful but I am confused on where to add this, `module DiskUsageExtend`, in `extend/pathname.rb` or in `utils.rb` and include in `class Pathname`?

xu-cheng:
You can put it `extend/pathname.rb` or separate file. Whatever you like. And maybe calling it as   `DiskUsageExtension` is better?

apjanke:
It looks like this is ready to go now. Any more comments/requests before we merge it?

Library/Homebrew/extend/pathname.rb
```diff
@@ -1,10 +1,50 @@
...
+    if self.directory?
+      @file_count, @disk_usage = [0, 0]
+      self.find.each do |f|
+        if !File.directory?(f) and !f.to_s.match(/\.DS_Store/)
+          @file_count += 1
+          @disk_usage += File.size(f)
+        end
+      end
+    else
+      @file_count, @disk_usage = [1, self.size]
```
xu-cheng: It would be better to split it to two statements, i.e.
```ruby
@file_count = 1
@disk_usage = self.size
```

xu-cheng:
:+1: LGTM, but we may want to have test case before we merge it.

mikemcquaid:
Merged (with @xu-cheng's last comment). Thanks @rakkesh!

@xu-cheng I wanted to merge this as-is as it's a non-maintainer contribution that predates our more recent guidelines.

mikemcquaid:
Doh; because this also predated the Ruby 1.8 check it caused test failures (fixed in 082fcd1).

xu-cheng:
I added more improvement in 91ed156f4b701cc8c8ef1ad3dbcd8995098bb3b7. (Sorry for missing them at the beginning)

rakkesh:
@xu-cheng @mikemcquaid nice work; good to see this is finally merged along with improvements. 🙏

adzenith:
Just saw the size message when I cleaned up and came here to say thanks! :clap:

PR #45644, opened 2015-11-03
elasticsearch 2.0.0

jasontedor:
This commit addresses several issues that exist in the Elasticsearch
formula.

1. Upgrades formula to install 2.0.0, the latest stable release of
Elasticsearch
2. Changes `HEAD` builds to require Java 8 (cf.
elastic/elasticsearch@11314336b6f5bf45fdb69a3d7c11ce29386535af)
3. Changes `HEAD` builds to use gradle for assembling (due to
Elasticsearch build system change, cf.
elastic/elasticsearch@c86100f636e9bddf9fd578c52dff3722b0ccd488)
4. Stop moving lib to libexec (due to security enhancements in
Elasticsearch; cf.
elastic/elasticsearch@052cf1446f9caea16454e0db5860e8e694c61119)
5. Keep plugins under Elasticsearch home
6. Fix startup command
7. Remove stack size setting (cf.
elastic/elasticsearch@09b4d0e097fe5b40c7c1b4469ed76629cc9860b9)

Closes elastic/elasticsearch#14424

### DIFF
--- a/Library/Formula/elasticsearch.rb
+++ b/Library/Formula/elasticsearch.rb
@@ -1,14 +1,15 @@
 class Elasticsearch < Formula
   desc "Distributed real-time search & analytics engine for the cloud"
   homepage "https://www.elastic.co/products/elasticsearch"
-  url "https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.7.3.tar.gz"
-  sha256 "af517611493374cfb2daa8897ae17e63e2efea4d0377d316baa351c1776a2bca"
+  url "https://download.elasticsearch.org/elasticsearch/release/org/elasticsearch/distribution/tar/elasticsearch/2.0.0/elasticsearch-2.0.0.tar.gz"
+  sha256 "b25f13f615337c2072964fd9fc5c7250f8a2a983b22198daf93548285d5d16df"
 
   bottle :unneeded
 
   head do
     url "https://github.com/elasticsearch/elasticsearch.git"
-    depends_on "maven" => :build
+    depends_on :java => "1.8"
+    depends_on "gradle" => :build
   end
 
   depends_on :java => "1.7+"
@@ -20,80 +21,66 @@ class Elasticsearch < Formula
   def install
     if build.head?
       # Build the package from source
-      system "mvn", "clean", "package", "-DskipTests"
+      system "gradle", "clean", "assemble"
       # Extract the package to the current directory
-      system "tar", "--strip", "1", "-xzf", "target/releases/elasticsearch-*.tar.gz"
+      targz = Dir["distribution/tar/build/distributions/elasticsearch-*.tar.gz"].first
+      system "tar", "--strip-components=1", "-xf", targz
     end
 
     # Remove Windows files
     rm_f Dir["bin/*.bat"]
     rm_f Dir["bin/*.exe"]
 
-    # Move libraries to `libexec` directory
-    libexec.install Dir["lib/*.jar"]
-    (libexec/"sigar").install Dir["lib/sigar/*.{jar,dylib}"]
-
     # Install everything else into package directory
-    prefix.install Dir["*"]
-
-    # Remove unnecessary files
-    rm_f Dir["#{lib}/sigar/*"]
-    if build.head?
-      rm_rf "#{prefix}/pom.xml"
-      rm_rf "#{prefix}/src/"
-      rm_rf "#{prefix}/target/"
-    end
+    libexec.install "bin", "config", "lib"
 
     # Set up Elasticsearch for local development:
-    inreplace "#{prefix}/config/elasticsearch.yml" do |s|
+    inreplace "#{libexec}/config/elasticsearch.yml" do |s|
       # 1. Give the cluster a unique name
-      s.gsub!(/#\s*cluster\.name\: elasticsearch/, "cluster.name: #{cluster_name}")
+      s.gsub!(/#\s*cluster\.name\: .*/, "cluster.name: #{cluster_name}")
 
       # 2. Configure paths
       s.sub!(%r{#\s*path\.data: /path/to.+$}, "path.data: #{var}/elasticsearch/")
       s.sub!(%r{#\s*path\.logs: /path/to.+$}, "path.logs: #{var}/log/elasticsearch/")
-      s.sub!(%r{#\s*path\.plugins: /path/to.+$}, "path.plugins: #{var}/lib/elasticsearch/plugins")
-
-      # 3. Bind to loopback IP for laptops roaming different networks
-      s.gsub!(/#\s*network\.host\: [^\n]+/, "network.host: 127.0.0.1")
     end
 
-    inreplace "#{bin}/elasticsearch.in.sh" do |s|
+    inreplace "#{libexec}/bin/elasticsearch.in.sh" do |s|
       # Configure ES_HOME
-      s.sub!(%r{#\!/bin/sh\n}, "#!/bin/sh\n\nES_HOME=#{prefix}")
-      # Configure ES_CLASSPATH paths to use libexec instead of lib
-      s.gsub!(%r{ES_HOME/lib/}, "ES_HOME/libexec/")
+      s.sub!(%r{#\!/bin/sh\n}, "#!/bin/sh\n\nES_HOME=#{libexec}")
     end
 
-    inreplace "#{bin}/plugin" do |s|
+    inreplace "#{libexec}/bin/plugin" do |s|
       # Add the proper ES_CLASSPATH configuration
-      s.sub!(/SCRIPT="\$0"/, %(SCRIPT="$0"\nES_CLASSPATH=#{libexec}))
+      s.sub!(/SCRIPT="\$0"/, %(SCRIPT="$0"\nES_CLASSPATH=#{libexec}/lib))
       # Replace paths to use libexec instead of lib
       s.gsub!(%r{\$ES_HOME/lib/}, "$ES_CLASSPATH/")
     end
 
     # Move config files into etc
-    (etc/"elasticsearch").install Dir[prefix/"config/*"]
-    (prefix/"config").rmtree
+    (etc/"elasticsearch").install Dir[libexec/"config/*"]
+    (etc/"elasticsearch/scripts").mkdir unless File.exists?(etc/"elasticsearch/scripts")
+    (libexec/"config").rmtree
+
+    bin.write_exec_script Dir[libexec/"bin/*"]
   end
 
   def post_install
     # Make sure runtime directories exist
     (var/"elasticsearch/#{cluster_name}").mkpath
     (var/"log/elasticsearch").mkpath
-    (var/"lib/elasticsearch/plugins").mkpath
-    ln_s etc/"elasticsearch", prefix/"config"
+    ln_s etc/"elasticsearch", libexec/"config"
+    (libexec/"plugins").mkdir
   end
 
   def caveats; <<-EOS.undent
     Data:    #{var}/elasticsearch/#{cluster_name}/
     Logs:    #{var}/log/elasticsearch/#{cluster_name}.log
-    Plugins: #{var}/lib/elasticsearch/plugins/
+    Plugins: #{libexec}/plugins/
     Config:  #{etc}/elasticsearch/
     EOS
   end
 
-  plist_options :manual => "elasticsearch --config=#{HOMEBREW_PREFIX}/opt/elasticsearch/config/elasticsearch.yml"
+  plist_options :manual => "elasticsearch"
 
   def plist; <<-EOS.undent
       <?xml version="1.0" encoding="UTF-8"?>
@@ -107,12 +94,9 @@ class Elasticsearch < Formula
           <key>ProgramArguments</key>
           <array>
             <string>#{HOMEBREW_PREFIX}/bin/elasticsearch</string>
-            <string>--config=#{etc}/elasticsearch/elasticsearch.yml</string>
           </array>
           <key>EnvironmentVariables</key>
           <dict>
-            <key>ES_JAVA_OPTS</key>
-            <string>-Xss200000</string>
           </dict>
           <key>RunAtLoad</key>
           <true/>
@@ -128,6 +112,14 @@ class Elasticsearch < Formula
   end
 
   test do
-    system "#{bin}/plugin", "--list"
+    system "#{bin}/plugin", "list"
+    pid = "#{testpath}/pid"
+    begin
+      system "#{bin}/elasticsearch", "-d", "-p", pid, "--path.data", testpath
+      sleep 10
+      system "curl", "-XGET", "localhost:9200/"
+    ensure
+      Process.kill(9, File.read(pid).to_i)
+    end
   end
 end


jasontedor:
That `brew audit` is failing is known to me. The issue is that Homebrew is requiring the jar files to be in `libexec`. However, [enhancements](https://github.com/elastic/elasticsearch/commit/052cf1446f9caea16454e0db5860e8e694c61119) in the Elasticsearch security manager require that the jar files be under `lib`. Together we'll need to find a way to work around this.

Library/Formula/elasticsearch.rb
```diff
@@ -128,6 +106,6 @@ def plist; <<-EOS.undent
   end
 
   test do
-    system "#{bin}/plugin", "--list"
+    system "#{bin}/plugin", "list"
```
mikemcquaid: Can we add another test that actually starts elasticsearch and checks it's working with `curl`?

mikemcquaid:
We can't install the `.jar` files to `lib`. We need to work out another location to install them to (i.e. `libexec`).

Library/Formula/elasticsearch.rb
jasontedor: Sure thing. I've pushed such a test in commit 28d4d999a84b00878dafee56579829963592b911.

Library/Formula/elasticsearch.rb
```diff
@@ -128,6 +103,17 @@ def plist; <<-EOS.undent
...
+    require 'tempfile'
+    file = Tempfile.new('homebrew-elasticsearch-test')
+    file.close
```
mikemcquaid: You can actually just write a file to the `testpath` which is in a temporary directory already.

Library/Formula/elasticsearch.rb
mikemcquaid: Nice :+1:

jasontedor:
@mikemcquaid For Elasticsearch 2.0.0, no such change is possible. The jar files must be in `lib`, no exceptions; Elasticsearch 2.0.0 will intentionally fail on startup if the jar files are moved to any other location (even with an `ES_CLASSPATH` configuration change). This is part of the aforementioned security enhancements that have been made to lock down Elasticsearch installations.

Looking beyond Elasticsearch 2.0.0, it is extremely unlikely that a change will be made to this policy. Elasticsearch now runs under the Java Security Manager and has a very aggressive security policy to make the system harder to exploit. Part of this includes explicitly enforcing the jar files to be in `lib` as part of the dynamic generation of file permissions (cf. [`o.e.e.Environment`](https://github.com/elastic/elasticsearch/blob/v2.0.0/core/src/main/java/org/elasticsearch/env/Environment.java#L159) and [`o.e.b.Security`](https://github.com/elastic/elasticsearch/blob/v2.0.0/core/src/main/java/org/elasticsearch/bootstrap/Security.java#L159)). It is extremely unlikely that this will be made user configurable.

Can you help me better understand the reason that Homebrew has this requirement for jar files, especially since this deviates from the standard file system hierarchy (`lib` is binary dependencies, `libexec` is for helper binaries)?

Library/Formula/elasticsearch.rb
jasontedor: Awesome, thanks. Added in 8e3c7208e429b540de7463c50a3543be6789ebb7.

mikemcquaid:
Note Debian also doesn't install these to `lib`: https://packages.debian.org/sid/all/libelasticsearch1.7-java/filelist

The message is fairly clear: https://github.com/Homebrew/homebrew/blob/ceab5e292c0420eaae02e3a4f9d84d4cc5d81aec/Library/Homebrew/formula_cellar_checks.rb#L41-L55

We symlink these files to `lib` from the Cellar. We cannot symlink all these jars because if other formula install the same files with the same names then they conflict. I don't think it's fair for Elasticsearch to expect us to change our policies just for them.

avinassh:
Is there any temp fix for the time being? I have installed elasticsearch using homebrew and the version is 2.0 (I don't know how, since [formula](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/elasticsearch.rb) still says 1.7). Anyways for the time being if you can tell me how do I get the v2 working would be great.

    $elasticsearch

    java.io.IOException: Resource not found: "org/joda/time/tz/data/ZoneInfoMap" ClassLoader: sun.misc.Launcher$AppClassLoader@43c0ae76
        at org.joda.time.tz.ZoneInfoProvider.openResource(ZoneInfoProvider.java:210)
        at org.joda.time.tz.ZoneInfoProvider.<init>(ZoneInfoProvider.java:127)
        at org.joda.time.tz.ZoneInfoProvider.<init>(ZoneInfoProvider.java:86)
        at org.joda.time.DateTimeZone.getDefaultProvider(DateTimeZone.java:514)
        at org.joda.time.DateTimeZone.getProvider(DateTimeZone.java:413)
        at org.joda.time.DateTimeZone.forID(DateTimeZone.java:216)
        at org.joda.time.DateTimeZone.getDefault(DateTimeZone.java:151)
        at org.joda.time.chrono.ISOChronology.getInstance(ISOChronology.java:79)
        at org.joda.time.DateTimeUtils.getChronology(DateTimeUtils.java:266)
        at org.joda.time.format.DateTimeFormatter.selectChronology(DateTimeFormatter.java:968)
        at org.joda.time.format.DateTimeFormatter.printTo(DateTimeFormatter.java:672)
        at org.joda.time.format.DateTimeFormatter.printTo(DateTimeFormatter.java:560)
        at org.joda.time.format.DateTimeFormatter.print(DateTimeFormatter.java:644)
        at org.elasticsearch.Build.<clinit>(Build.java:51)
        at org.elasticsearch.node.Node.<init>(Node.java:135)
        at org.elasticsearch.node.NodeBuilder.build(NodeBuilder.java:145)
        at org.elasticsearch.bootstrap.Bootstrap.setup(Bootstrap.java:170)
        at org.elasticsearch.bootstrap.Bootstrap.init(Bootstrap.java:270)
        at org.elasticsearch.bootstrap.Elasticsearch.main(Elasticsearch.java:35)
    [2015-11-04 13:20:26,431][INFO ][node                     ] [Abigail Brand] version[2.0.0], pid[3867], build[de54438/2015-10-22T08:09:48Z]
    [2015-11-04 13:20:26,433][INFO ][node                     ] [Abigail Brand] initializing ...
    Exception in thread "main" java.lang.IllegalStateException: failed to load bundle [] due to jar hell
    Likely root cause: java.security.AccessControlException: access denied ("java.io.FilePermission" "/usr/local/Cellar/elasticsearch/2.0.0/libexec/antlr-runtime-3.5.jar" "read")
        at java.security.AccessControlContext.checkPermission(AccessControlContext.java:372)
        at java.security.AccessController.checkPermission(AccessController.java:559)
        at java.lang.SecurityManager.checkPermission(SecurityManager.java:549)
        at java.lang.SecurityManager.checkRead(SecurityManager.java:888)
        at java.util.zip.ZipFile.<init>(ZipFile.java:206)
        at java.util.zip.ZipFile.<init>(ZipFile.java:145)
        at java.util.jar.JarFile.<init>(JarFile.java:154)
        at java.util.jar.JarFile.<init>(JarFile.java:91)
        at org.elasticsearch.bootstrap.JarHell.checkJarHell(JarHell.java:173)
        at org.elasticsearch.plugins.PluginsService.loadBundles(PluginsService.java:340)
        at org.elasticsearch.plugins.PluginsService.<init>(PluginsService.java:113)
        at org.elasticsearch.node.Node.<init>(Node.java:144)
        at org.elasticsearch.node.NodeBuilder.build(NodeBuilder.java:145)
        at org.elasticsearch.bootstrap.Bootstrap.setup(Bootstrap.java:170)
        at org.elasticsearch.bootstrap.Bootstrap.init(Bootstrap.java:270)
        at org.elasticsearch.bootstrap.Elasticsearch.main(Elasticsearch.java:35)
    Refer to the log for complete error details.

rmuir:
> We symlink these files to lib from the Cellar. We cannot symlink all these jars because if other formula install the same files with the same names then they conflict. I don't think it's fair for Elasticsearch to expect us to change our policies just for them.

We aren't changing to libexec. I don't think its fair for homebrew to expect us to change our code just for them.

jasontedor:
> Note Debian also doesn't install these to lib: https://packages.debian.org/sid/all/libelasticsearch1.7-java/filelist

1. That's a pre-2.0.0 version (1.7) of Elasticsearch but the requirement was not put in place until version 2.0.0 of Elasticsearch. 
2. That's only the libelasticsearch package, which itself depends on many other lib* packages (which all get installed in `/usr/share/java`).
3. The officially [supported](https://www.elastic.co/downloads/elasticsearch) Debian package for Elasticsearch 2.0.0 puts all of the jars in `lib` (`/usr/share/elasticsearch/lib`).

> The message is fairly clear: https://github.com/Homebrew/homebrew/blob/ceab5e292c0420eaae02e3a4f9d84d4cc5d81aec/Library/Homebrew/formula_cellar_checks.rb#L41-L55

I [already](https://github.com/Homebrew/homebrew/pull/45644#issuecomment-153466711) understand that.

> We symlink these files to lib from the Cellar. We cannot symlink all these jars because if other formula install the same files with the same names then they conflict.

These days, almost all jars come with versions in their names so the conflicts that you speak of are effectively nonexistent. It only becomes conflict if someone tries to add all of `/usr/local/lib` to their classpath which would be a horrible practice. No installed package should do that, and no developer should be doing that. Symlinking all the jars from all the installed packages into one location is just setting up for [trouble](https://en.wikipedia.org/wiki/Java_Classloader#JAR_hell).

Maybe the answer is as simple as not symlinking jars from the package installed `lib` folders?

> I don't think it's fair for Elasticsearch to expect us to change our policies just for them.

Elastic hasn't asked for anything. The way I see the conversation right now is merely exploring the issue and seeing what options, if any, fall out.

catalandres:
@avinassh Have you tried `brew edit elasticsearch` and then do the edits in the diff corresponding to the commit? I did that and ES compiled and started without complaining. It's a dirty hack and I have not verified that the software is 100% operational.

avinassh:
@catalandres Worked! Thanks. I didn't think of it earlier.

mikemcquaid:
> We aren't changing to libexec. I don't think its fair for homebrew to expect us to change our code just for them.

You'll note I'm not submitting a PR or issue to your project asking you to change your code. I'm happy for Homebrew users to remain on Elasticsearch 1.x until this is resolved.

> The officially supported Debian package for Elasticsearch 2.0.0 puts all of the jars in lib (/usr/share/elasticsearch/lib).

That's not the package Debian includes. Similarly, Elasticsearch could distribute their own Homebrew formula which does not follow our `audit` rules. Like the Debian package it won't be installed if you `brew install elasticsearch` without adding an external repository, though.

> These days, almost all jars come with versions in their names so the conflicts that you speak of are effectively nonexistent. And having different versions of the same artifacts in /usr/local/lib only becomes a conflict if someone tries to add all of /usr/local/lib to their classpath or multiple versions of the same artifact, both of which would are bad practices. 

That's not the case because of how Homebrew works. The contents of `Cellar/elasticsearch/2.0.0/lib` are always symlinked into `lib`. If another package also wants to symlink the same files to `lib`: there is a conflict. We have a `conflict_with` DSL we use to indicate this but it just means some software can't be installed at the same time. We resolve this by installing things to `libexec`, as we have done with Elasticsearch in the past.

> Symlinking all the jars from all the installed packages into one location is just setting up for trouble.

That's why we don't allow jars in `lib`. That's exactly how Homebrew works and has always worked.

> What are the reasons for doing this? Maybe the answer is as simple as not symlinking jars from the package-installed lib folders?

That's not simple. It would require fundamentally changing the way Homebrew installs software. This would potentially break other software that implicitly relies on the way Homebrew currently installs software, require us to write and test a bunch of code and discuss how best to deal with the new problems that would result.

jasontedor:
> You'll note I'm not submitting a PR or issue to your project asking you to change your code.

That's how open source works.

> I'm happy for Homebrew users to remain on Elasticsearch 1.x until this is resolved.

I don't understand this as that doesn't seem in the best interest of Homebrew's users.

> That's not the package Debian includes.

What does that matter?

> Elasticsearch could distribute their own Homebrew formula which does not follow our `audit` rules. Like the Debian package it won't be installed if you `brew install elasticsearch` without adding an external repository, though.

That's the most likely outcome of this, as well as opening a pull request to remove the Elasticsearch formula from the homebrew repository since it's currently broken for latest stable and head and not fixable given the conversation here.

> That's why we don't allow jars in `lib`. That's exactly how Homebrew works and has always worked.

You don't allow them because symlinking would be setting up for trouble. So just don't symlink them?

> That's not simple. It would require fundamentally changing the way Homebrew installs software. This would potentially break other software that implicitly relies on the way Homebrew currently installs software, require us to write and test a bunch of code and discuss how best to deal with the new problems that would result.

How is it not simple to exclude jars from being symlinked if they are in `lib`? It breaks nothing that exists today because they aren't allowed in `lib` today.

Thank you for taking the time to review the pull request and your comments on adding a test. Sincerely, this has been helpful and enlightening.

jasontedor:
> Is there any temp fix for the time being?

@avinassh Yup, as @catalandres mentioned (thanks!), you can apply the changes in this pull request locally. That will get you up and running for now. We will very likely be establishing an official Elasticsearch [tap](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/brew-tap.md) going forward. Whatever the outcome, you can track the status of this on elastic/elasticsearch#14424.

mikemcquaid:
>> You'll note I'm not submitting a PR or issue to your project asking you to change your code.
>
> That's how open source works.

It's worth noting here that Homebrew is exclusively run by volunteers in their free time and that's not the case for Elastic.

>> I'm happy for Homebrew users to remain on Elasticsearch 1.x until this is resolved.
>
> I don't understand this as that doesn't seem in the best interest of Homebrew's users.

With respect, I've been a Homebrew maintainer for just under 7 years and I think I've got a pretty good grasp of what's in our users best interests. To merge this as-is will cause additional issues. None of our `audit` rules are there for no reason, they have all been added to prevent previous issues from recurring.

>> That's not the package Debian includes.
> 
> What does that matter?

My point there was that it's unclear you've convinced other package managers to package Elasticsearch in the way you want us to (e.g. put all the jars in `lib`).

>> Elasticsearch could distribute their own Homebrew formula which does not follow our audit rules. Like the Debian package it won't be installed if you brew install elasticsearch without adding an external repository, though.
> That's the most likely outcome of this, as well as opening a pull request to remove the Elasticsearch formula from the homebrew repository since it's currently broken for latest stable and HEAD and not fixable given the conversation here.

It's not currently broken, it's just not up-to-date. It is fixable but requires us to find us a middle solution together.

> That's why we don't allow jars in lib. That's exactly how Homebrew works and has always worked.
> You don't allow them because symlinking would be setting up for trouble. So just don't symlink them?
> 
> That's not simple. It would require fundamentally changing the way Homebrew installs software. This would potentially break other software that implicitly relies on the way Homebrew currently installs software, require us to write and test a bunch of code and discuss how best to deal with the new problems that would result.
> How is it not simple to exclude jars from being symlinked if they are in lib? It breaks nothing that exists today because they aren't allowed in lib today.

> How is it not simple to exclude jars from being symlinked if they are in lib? It breaks nothing that exists today because they aren't allowed in lib today.

Technically the code is simple to just special-case this one thing. I hope you appreciate that with >3000 formulae writing making these type changes based on one formula quickly becomes unsustainable.

I'm still not sure I completely understand what the requirement is here with `lib`. Would it be possible to use symlinks and/or a wrapper script and keep installing to `libexec`? If not, can you explain why this is not possible so we can try and work out another solution?

jasontedor:
> With respect, I've been a Homebrew maintainer for just under 7 years and I think I've got a pretty good grasp of what's in our users best interests.

I don't doubt that for a second. I've been a Homebrew user for a long time, and is the only package manager that I have installed on my systems; that should tell you the level of respect and appreciation that I have for the work of the Homebrew maintainers including yourself.

Saying that

> I'm happy for Homebrew users to remain on Elasticsearch 1.x until this is resolved.

when prior to your recent comment it was appearing that we weren't going to get anywhere you were effectively banishing Homebrew users to a rapidly-aging version of Elasticsearch that will soon be out of support.

> My point there was that it's unclear you've convinced other package managers to package Elasticsearch in the way you want us to (e.g. put all the jars in `lib`).

First, we have our own official packages that we support. Second, the package that you were pointing to is an old version (1.7) of Elasticsearch that didn't have the requirement that jars be in `lib`. If an attempt were to be made today by an outside organization (and to be clear, Elastic welcomes such community involvement), it would also have to observe the `lib` requirement because it will not work otherwise.

> It's not currently broken, it's just not up-to-date. It is fixable but requires us to find us a middle solution together.

It is de facto broken for `HEAD`, and effectively broken since it can't be brought up to latest stable without breaking.

> It is fixable but requires us to find us a middle solution together.

That's all I've wanted all along. Do you have any proposals?

> Technically the code is simple to just special-case this one thing. I hope you appreciate that with >3000 formulae writing making these type changes based on one formula quickly becomes unsustainable.

Your concern is conflicting jars from symlinks in `/usr/local/lib` to package-installed jars in package `lib` folders. I'm of the mindset that there isn't a good reason to be doing this anyway, Making that change doesn't break anything that exists today (since it is impossible for there to be any jars there right now). I'm not proposing that Elasticsearch be special-cased, I'm proposing that jars just not be symlinked in `/usr/local/lib`. I think this avoids all the problems that concern you, and it's a good practice anyway.

> I'm still not sure I completely understand what the requirement is here with `lib`. 

It's part of our integration with the Java Security Manager. We are going to great lengths to minimize Elasticsearch exploit vectors. In this particular case, with filesystem protections in place and requiring that jars be in `lib`, we remove some exploit vectors. At Elasticsearch startup, we create a special set of permissions for only the jars in the `lib` folder.

> Would it be possible to use symlinks and/or a wrapper script and keep installing to `libexec`?

The only place that such a symlink could come from is `lib`, no other symlink would work. But that still fails the `brew audit`, doesn't avoid the issues that concern you, and we are possibly going to remove the ability for `lib` to be a symlink anyway.

> If not, can you explain why this is not possible so we can try and work out another solution?

I hope I clarified this, but if not, let me know? Thanks again for your time.

mikemcquaid:
> The only place that such a symlink could come from is lib, no other symlink would work. But that still fails the brew audit, doesn't avoid the issues that concern you, and we are possibly going to remove the ability for lib to be a symlink anyway.

What would happen if we installed the libraries to `libexec/lib` and make `libexec` the home (thanks @xu-cheng)? We could then install everything in there and use wrapper scripts and/or symlinks just into `bin`.

DomT4:
Applying this diff to your formula works for me, locally, with the test you wrote and also starting/stopping the `elasticsearch` server.

```diff
diff --git a/Library/Formula/elasticsearch.rb b/Library/Formula/elasticsearch.rb
index d1b16a2..38583f7 100644
--- a/Library/Formula/elasticsearch.rb
+++ b/Library/Formula/elasticsearch.rb
@@ -32,12 +32,10 @@ class Elasticsearch < Formula
     rm_f Dir["bin/*.exe"]
 
     # Install everything else into package directory
-    prefix.install "bin"
-    prefix.install "config"
-    prefix.install "lib"
+    libexec.install "bin", "config", "lib"
 
     # Set up Elasticsearch for local development:
-    inreplace "#{prefix}/config/elasticsearch.yml" do |s|
+    inreplace "#{libexec}/config/elasticsearch.yml" do |s|
       # 1. Give the cluster a unique name
       s.gsub!(/#\s*cluster\.name\: .*/, "cluster.name: #{cluster_name}")
 
@@ -46,21 +44,24 @@ class Elasticsearch < Formula
       s.sub!(%r{#\s*path\.logs: /path/to.+$}, "path.logs: #{var}/log/elasticsearch/")
     end
 
-    inreplace "#{bin}/elasticsearch.in.sh" do |s|
+    inreplace "#{libexec}/bin/elasticsearch.in.sh" do |s|
       # Configure ES_HOME
-      s.sub!(%r{#\!/bin/sh\n}, "#!/bin/sh\n\nES_HOME=#{prefix}")
+      s.sub!(%r{#\!/bin/sh\n}, "#!/bin/sh\n\nES_HOME=#{libexec}")
     end
 
     # Move config files into etc
-    (etc/"elasticsearch").install Dir[prefix/"config/*"]
-    (prefix/"config").rmtree
+    (etc/"elasticsearch").install Dir[libexec/"config/*"]
+    (libexec/"config").rmtree
+
+    # Exec script executables back to top-level bin
+    bin.write_exec_script Dir[libexec/"bin/*"]
   end
 
   def post_install
     # Make sure runtime directories exist
     (var/"elasticsearch/#{cluster_name}").mkpath
     (var/"log/elasticsearch").mkpath
-    ln_s etc/"elasticsearch", prefix/"config"
+    ln_s etc/"elasticsearch", libexec/"config"
   end
 
   def caveats; <<-EOS.undent
```

jasontedor:
Thanks for the suggestion @mikemcquaid. It's hacky, it's goes further against the standard filesystem hierarchy, but it might work. I'm still working through the implications locally (plugins do not work right now).

jasontedor:
@DomT4 Yes, I have the same working locally for Elasticsearch server, but plugins do not work at the moment. I've almost worked through the implications and a fix.

DomT4:
> Yes, I have the same working locally for Elasticsearch server, but plugins do not work at the moment.

Can you provide an example of how we'd discover this breakage and what the breakage specifically is? Elasticsearch is not something I use incredibly regularly myself.

Curious on what that breakage is given presumably `elasticsearch` doesn't care what prefix you put it in, as long as the home is set correctly and all the elements remain inside the same prefix untouched? As I said I don't use it much, so feel free to ELI5.

jasontedor:
@mikemcquaid @DomT4 I've pushed b3d24e555b725ea3a558f7302da22f69a7b05383 incorporating what we've discussed. This pull request passes `brew audit` now.

mikemcquaid:
@jasontedor :clap: that's great, thanks.

mikemcquaid:
It looks like the test is failing but should be relatively easy to fix: https://travis-ci.org/Homebrew/homebrew/jobs/89335368#L313

You can test locally with `brew test --sandbox --verbose elasticsearch`

jasontedor:
> You can test locally with `brew test --sandbox --verbose elasticsearch`

Didn't know that, thank you. But I think I figured out one issue from the CI logs and pushed 5ee397865bbef1b8b697ad0e1078e707777fb87c to address but it looks like there will be more. I'll iterate locally. Thanks again for your time and help.

jasontedor:
@mikemcquaid With 1b4c8d042087719bf43cffb81583dd9910e31182 sandboxed tests are passing locally. :smile:

jasontedor:
@mikemcquaid With green on the CI now, is this ready for the rebase/squash dance into a final single commit for integration into master?

mikemcquaid:
@jasontedor That would be fantastic, thanks. Thanks for all your work here and apologies for not recognising a better solution earlier.

jasontedor:
@mikemcquaid Done and done. Thanks again.

jasontedor:
@mikemcquaid It occurred to me earlier this morning that if Homebrew is okay with installing into subdirectories and with the `bin.write_exec_script` trick, we can just install Elasticsearch components into `home` instead of `libexec`. I've pushed 69fa59298e301bf8a18b60ebfdc56b8127441f33. Let me know what you think and I'll squash one last time before this is integrated into master?

mikemcquaid:
> @mikemcquaid It occurred to me earlier this morning that if Homebrew is okay with installing into subdirectories and with the bin.write_exec_script trick, we can just install Elasticsearch components into home instead of libexec. I've pushed 69fa592. Let me know what you think and I'll squash one last time before this is integrated into master?

@jasontedor We'd probably prefer `libexec` just because it's ended up being a Homebrew convention so it's more obvious to maintainers why that's been done. If you could go back to `libexec` and squash I'll merge this ASAP.

jasontedor:
> We'd probably prefer `libexec` just because it's ended up being a Homebrew convention so it's more obvious to maintainers why that's been done.

Understood.

> If you could go back to libexec and squash I'll merge this ASAP.

Done. Thanks a lot!

mikemcquaid:
Pushed! Thanks for your contribution to Homebrew, @jasontedor, and all your hard work here! Without people like you submitting PRs we couldn't run this project. You rock! Thanks also to @DomT4 for the suggested patch.

Just so you don't :suspect:: I also added a revision here so anyone on the previous (broken) 2.0.0 version we shipped will see an upgrade to the new `2.0.0_1` version.

Issue #45648, opened 2015-11-03
chicken incorrectly selects 32-bit build env, results in build error


turanct:
The error is `apply-hack.x86.S:35:suffix or operands invalid for 'call'` and is [described in the official README](http://code.call-cc.org/cgi-bin/gitweb.cgi?p=chicken-core.git;a=blob;f=README;h=2b8b2901284a368628d81be6636424ccbc3cf4a7;hb=b25963183b63530eda36ddca14466baffdff01ee#l463), as is the fix for it.

It boils down to manually setting an argument to the make command. If i knew how to do this in brew, i would create a pull request, so ping me with a howto and i'll do it.

For the time being i fixed it by just changing the [formula file](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/chicken.rb) on my machine with the necessary argument hardcoded.

turanct:
in the mean time, i found out how to do it, working on a PR right now so we can do `brew install --with-force-64-bit chicken` if the error occurs

DomT4:
Closing in favour of #45654, Thanks!

PR #45654, opened 2015-11-03
Chicken: Add an optional arg to allow forcing 64 bit build ENV

turanct:
Chicken has [a known error](http://code.call-cc.org/cgi-bin/gitweb.cgi?p=chicken-core.git;a=blob;f=README;h=2b8b2901284a368628d81be6636424ccbc3cf4a7;hb=b25963183b63530eda36ddca14466baffdff01ee#l463) that it will not compile without the right
make option on some older OSX versions. This commit adds the make option
as a build option, so people encountering the error, can rectify it.

Read more: issue #45648

### DIFF
--- a/Library/Formula/chicken.rb
+++ b/Library/Formula/chicken.rb
@@ -12,6 +12,10 @@ class Chicken < Formula
     sha256 "1f46226c58b1b7cd92f4d0c68cc0583e2c5ec4c2cce60b53193afca9a5d8be19" => :mountain_lion
   end
 
+  # Sometimes chicken detects a 32-bit environment by mistake, causing errors,
+  # see https://github.com/Homebrew/homebrew/issues/45648
+  option "with-force-64-bit", "Build with forced 64-bit build environment"
+
   def install
     ENV.deparallelize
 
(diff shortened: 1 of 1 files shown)


DomT4:
Is there a reason you wouldn't want a 64-bit build on a 64-bit machine by default?

turanct:
@DomT4 not all systems are 64bit. The makefile for chicken tries to determine if the system is 64bit or not, and fails for some specific macs built around 2009 (from what i understood of the original issue on their site), which is also why it failed with me. The fix is to specify it manually. This would of course be very easy to do automatically if brew knew for sure if the system was 64bit or not, but i don't know if we can do that?

DomT4:
> This would of course be very easy to do automatically if brew knew for sure if the system was 64bit or not, but i don't know if we can do that?

`if MacOS.prefer_64_bit?`

turanct:
@DomT4 thanks! sorry, i'm new to creating/editing brew Formulae

I tested it, and it works as expected, it now just works out of the box with `brew install chicken`, so I added it to this pull request. thanks again!

DomT4:
No worries! More than happy to answer if you have questions or such. We all had our first Homebrew contribution at some point, and mine was a mess from what I remember, heh.

Merged in https://github.com/Homebrew/homebrew/commit/9edee4fdf7b9a6545318540cfab29430b73dc121. Thank you for your contribution to Homebrew @turanct; we appreciate it! :smiley_cat:

PR #45658, opened 2015-11-03
maximum_macos_requirement: reword message

DomT4:
*Whole bunch* of confusion about the purpose of the MMOSR, largely with people believing we're adding artificial caps ourselves rather than having them enforced on us by upstream issues.

New wording looks like:

```
ld64: This formula either does not compile or function as expected on newer
OS X versions than Snow Leopard due to an upstream issue.

You can try to install the formula with `--HEAD` but one may not be
defined or the issue may still be present upstream.
```

### DIFF
--- a/Library/Homebrew/requirements/maximum_macos_requirement.rb
+++ b/Library/Homebrew/requirements/maximum_macos_requirement.rb
@@ -12,7 +12,11 @@ class MaximumMacOSRequirement < Requirement
 
   def message
     <<-EOS.undent
-      OS X #{@version.pretty_name} or older is required.
+      This formula either does not compile or function as expected on newer
+      OS X versions than #{@version.pretty_name} due to an upstream incompatibility.
+
+      You can try to install the formula with `--HEAD` but one may not be
+      defined or the issue may still be present upstream.
     EOS
   end
 end


Library/Homebrew/requirements/maximum_macos_requirement.rb
```diff
@@ -12,7 +12,11 @@ def initialize(tags)
 
   def message
     <<-EOS.undent
-      OS X #{@version.pretty_name} or older is required.
+      This formula either does not compile or function as expected on newer
```
bfontaine: Shouldn’t that be “`nor`” instead of “`or`” here?

DomT4: My former English teacher will shoot me out of a cannon for this, but I couldn't remember if the standard negative `nor` / `or` usage applied in the context where one may not be valid, i.e. some formula break on compile but some formula compile fine but break at runtime. It's not universally the case that a formula would both break at compile *and* at runtime.

DomT4: I think `or` is typically used in an `either`, and `nor` in `neither`?

* Dom found neither Mike nor Misty.
* Dom found either Mike or Misty.

Rather than

* Dom found either Mike nor Misty.
* Dom found neither Mike or Misty.

Maybe.

Library/Homebrew/requirements/maximum_macos_requirement.rb
```diff
@@ -12,7 +12,11 @@ def initialize(tags)
...
+      OS X versions than #{@version.pretty_name} due to an upstream incompatibility.
```
mikemcquaid: `on OS X versions newer than`

Library/Homebrew/requirements/maximum_macos_requirement.rb
```diff
@@ -12,7 +12,11 @@ def initialize(tags)
...
+
+      You can try to install the formula with `--HEAD` but one may not be
+      defined or the issue may still be present upstream.
```
mikemcquaid: I'm kinda meh on this given how broken `--HEAD` often is. I'd just omit this line.

DomT4: Happy to if you'd prefer not to keep it. I added it because we deliberately don't MMOSR `head`s in the impacted formulae to give people who complain loudly something to try.

Library/Homebrew/requirements/maximum_macos_requirement.rb
DomT4: Will change.

DomT4:
Made the language changes in https://github.com/Homebrew/homebrew/commit/ceab5e292c0420eaae02e3a4f9d84d4cc5d81aec when I merged.

Library/Homebrew/requirements/maximum_macos_requirement.rb
bfontaine: Yeah. I’m not a native speaker but here is how I understand the sentence:
> This formula either (does not compile) or (function as expected)

Moving the “either” sounds weird:
> This formula does not either compile or function as expected

Again I’m not a native speaker and @tdsmith might have a more informed opinion.

Issue #45616, opened 2015-11-02
EmacsRequirement error on brew update


daanemanz:
When running `brew update` I'm encountering the following error:
```
Error: Specify a version for EmacsRequirement
```

Version:
```
$ brew --version
0.9.5 (git revision dc5a; last commit 2015-10-22)
```

I'm seeing this both on Travis CI and on my local machine.

bfontaine:
What does `brew update --verbose` say?

daanemanz:
@bfontaine nothing but the cloning of repositories; it just hangs there:

```
$ brew update --verbose
==> git symbolic-ref --short HEAD 2>/dev/null
master
==> git rev-parse -q --verify HEAD
dc5a7e797bc1d6411699090fef6d86d7292a94cb
git config core.autocrlf false
git pull --ff --no-rebase origin refs/heads/master:refs/remotes/origin/master
remote: Counting objects: 3896, done.
remote: Compressing objects: 100% (1107/1107), done.
remote: Total 3896 (delta 3069), reused 3600 (delta 2777), pack-reused 0
Receiving objects: 100% (3896/3896), 537.78 KiB | 0 bytes/s, done.
Resolving deltas: 100% (3069/3069), completed with 501 local objects.
From https://github.com/Homebrew/homebrew
   dc5a7e7..18c483f  master     -> origin/master
Updating dc5a7e7..18c483f
Fast-forward
 .travis.yml                                        |   5 +-
<snippety snip>
...
</snip>
 520 files changed, 3529 insertions(+), 1696 deletions(-)
 create mode 100644 Library/Formula/carina.rb
 create mode 100644 Library/Formula/dfmt.rb
 create mode 100644 Library/Formula/gdub.rb
 create mode 100644 Library/Formula/glslang.rb
 create mode 100644 Library/Formula/libmill.rb
 create mode 100644 Library/Formula/librest.rb
 create mode 100644 Library/Formula/libsoundio.rb
 create mode 100644 Library/Formula/libstatgrab.rb
 create mode 100644 Library/Formula/mailhog.rb
 create mode 100644 Library/Formula/ne.rb
 create mode 100644 Library/Formula/otto.rb
 create mode 100644 Library/Formula/shairport-sync.rb
 create mode 100644 Library/Formula/sqliteodbc.rb
 create mode 100644 Library/Formula/swagger-codegen.rb
 create mode 100644 Library/Formula/tailor.rb
[...]

PR #45676, opened 2015-11-04
update: always rescue formula loading

xu-cheng:
Short term fix

### DIFF
--- a/Library/Homebrew/cmd/update.rb
+++ b/Library/Homebrew/cmd/update.rb
@@ -108,7 +108,8 @@ module Homebrew
 
       begin
         f = Formulary.factory("#{user}/#{repo}/#{newname}")
-      rescue FormulaUnavailableError, *FormulaVersions::IGNORED_EXCEPTIONS
+      # short term fix to prevent situation like https://github.com/Homebrew/homebrew/issues/45616
+      rescue Exception
       end
 
       next unless f
@@ -314,7 +315,8 @@ class Updater
             end
             old_version = FormulaVersions.new(formula).formula_at_revision(@initial_revision, &:pkg_version)
             next if new_version == old_version
-          rescue FormulaUnavailableError, *FormulaVersions::IGNORED_EXCEPTIONS => e
+          # short term fix to prevent situation like https://github.com/Homebrew/homebrew/issues/45616
+          rescue Exception => e
             onoe e if ARGV.homebrew_developer?
           end
           map[:M] << file


bfontaine:
You might want to add a comment above these lines explaining it’s a short-term fix.

UniqMartin:
I'd be curious to know what this is supposed to fix and what the long-term solution will be.

xu-cheng:
By short term fix, I mean we should probably reimplement `brew update` rather than this is a fix which will be reverted soon.

xu-cheng:
@bfontaine @UniqMartin see #45616

bfontaine:
@xu-cheng Yeah I’m aware of this issue :wink: but it might be worth adding a link to it as a comment above these `rescue Exception` just in case someone reads this code and wonder why we’re rescuing all Exceptions.

UniqMartin:
I wasn't aware, so thanks for the issue link! I agree that there should be a comment mentioning the issue number and that it's a short-term fix, though.

xu-cheng:
Updated

UniqMartin:
LGTM :+1:

bfontaine:
Thank you @xu-cheng :heart:

Library/Homebrew/cmd/update.rb
```diff
@@ -108,7 +108,8 @@ def update
 
       begin
         f = Formulary.factory("#{user}/#{repo}/#{newname}")
-      rescue FormulaUnavailableError, *FormulaVersions::IGNORED_EXCEPTIONS
+      # short term fix to prevent situation like https://github.com/Homebrew/homebrew/issues/45616
+      rescue Exception
```
mikemcquaid: Should probably also handle `RuntimeError` with just a `rescue` if you go this route?

mikemcquaid:
The problem with this is if there's an error there's no indication to the user anything has happened and they will just get incorrect output. The longer-term fix, I think, is not ever to load formula code in the `update` process but instead shell out to a new process _after_ update.

Library/Homebrew/cmd/update.rb
xu-cheng: `Exception` will cover them all.

xu-cheng:
I'm exploring the long term fix. No ETA though.

UniqMartin:
Definitely agree on the long-term fix. `brew update` should really be a two-step process that first updates all the Git repositories and then spawns a new process that runs the updated code to generate the report and to do the renames, migrations, etc.

I would also be interested in working on that, but I think this will be in a few weeks at the earliest. Would it make sense to create an issue now to coordinate this and to keep everyone interested up-to-date? Whoever happens to start working on this first could post there to save others from duplicating work.

**As for detecting these kinds of problems with `brew update-test`:**

I haven't looked at the code, but I believe it tests the update `origin/master` :arrow_right: `origin/master+PR`. If so, would it be possible to add additional test cases like

- `origin/master` from a week ago :arrow_right: `origin/master+PR`
- `origin/master` from a month ago :arrow_right: `origin/master+PR`
- `origin/master` from three months ago :arrow_right: `origin/master+PR`

without too much effort?

mikemcquaid:
I'm OK with having a tracking issue and will probably be able to start in the next week or so. I don't think additional testing really helps us here; by the time we detect it it's already too late. I'd actually like to remove the `update-test` as it's relatively slow and I think it's lulled us into a false sense of security.

> I'm exploring the long term fix. No ETA though.

If you disagree with the approaches we've said here, please say so. I'm a little frustrated that this fix was pushed through without much time for review while I and other maintainers were asleep.

xu-cheng:
> I'm exploring the long term fix. No ETA though.

I mean I'm exploring the approach we talk about, i.e. reimplement it using bash.  

> I'm a little frustrated that this fix was pushed through without much time for review while I and other maintainers were asleep.

Sorry about that.

mikemcquaid:
> I mean I'm exploring the approach we talk about, i.e. reimplement it using bash.

Great, thanks :+1: 

> Sorry about that.

That's ok :heart:. I understand why it happened and I've done the same in the past but I just figured it was better to say if I was a little :angry: rather than you not know. :bow:

Issue #45667, opened 2015-11-04
Please modify swatchbooker formula to not prompt for 'pip install pillow


luzpaz:
https://github.com/Homebrew/homebrew/blob/master/Library/Formula/swatchbooker.rb
When running the above formula I get prompted to install pillow via pip. That's an inconvenience

mikemcquaid:
Could you try and open a pull request? This should help and I'm happy to walk you through anything else: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged

Thanks!

luzpaz:
CC @olivierberten

DomT4:
Closing in favour of #45682, Thanks!

luzpaz:
please offer guidance thanks

PR #45682, opened 2015-11-04
swatchbooker w/o pip install pillow

luzpaz:
as per https://github.com/Homebrew/homebrew/issues/45667

### DIFF
--- a/Library/Formula/swatchbooker.rb
+++ b/Library/Formula/swatchbooker.rb
@@ -4,10 +4,14 @@ class Swatchbooker < Formula
   url "https://launchpad.net/swatchbooker/trunk/0.7.3/+download/SwatchBooker-0.7.3.tar.gz"
   sha256 "c0c8bf038156337f1eebdf6f7c99f5b7a8e8f9a332cd625c57269ddc8ba18eb7"
 
+  resource "Pillow" do
+    url "https://pypi.python.org/packages/source/P/Pillow/Pillow-2.9.0.zip"
+    sha256 "d1db8dfed77547076037d589b598e04f2cbc1a7835d3d3f137bf20c8994854d5"
+  end
+
   depends_on :python
   depends_on "little-cms" => "with-python"
   depends_on "pyqt"
-  depends_on "homebrew/python/pillow"
 
   patch :DATA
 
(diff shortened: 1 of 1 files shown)


Library/Formula/swatchbooker.rb
```diff
@@ -5,9 +5,9 @@ class Swatchbooker < Formula
   sha256 "c0c8bf038156337f1eebdf6f7c99f5b7a8e8f9a332cd625c57269ddc8ba18eb7"
 
   depends_on :python
-  depends_on "pillow" => [:python, "PIL"]
   depends_on "little-cms" => "with-python"
   depends_on "pyqt"
+  depends_on "homebrew/python/pillow"
```
bfontaine: We don’t allow core formulae to depend on tap ones. If this is a mandatory requirement we must move this formula to the Python tap.

DomT4: It should be fine to vendor. We [do so](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/mitmproxy.rb#L93-L96) in `mitmproxy` without any obviously horrific problems. Possible the instruction to use `pip` in this formula previously is a relatively old relic.

luzpaz:
how do i make the install block work ?

mikemcquaid:
Check out `ansible.rb`, that should be helpful.

luzpaz:
From [Cookbook](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#specifying-gems-python-modules-go-projects-etc-as-dependencies):  
> [homebrew-pypi-poet](https://github.com/tdsmith/homebrew-pypi-poet) can help you generate resource stanzas for the dependencies of your Python application  

Will test this out

luzpaz:
@DomT4 this is also a little out of my comfort zone. Would you be willing to assist ?

Library/Formula/swatchbooker.rb
```diff
@@ -17,10 +21,16 @@ def install
...
     end
 
-    system "python", "setup.py", "install", "--prefix=#{prefix}"
-    bin.env_script_all_files(libexec+"bin", :PYTHONPATH => ENV["PYTHONPATH"])
-    chmod 0755, libexec/"bin/swatchbooker"
-  end
+    ENV["PYTHONPATH"] = libexec/"vendor/lib/python2.7/site-packages"
+
+    resource("Pillow").stage
+      system "python", *Language::Python.setup_install_args(libexec/"vendor")
```
bfontaine: Wrap this line in a block and give it to the line above; it should look like this:
```rb
resource("Pillow").stage do
end
```

`Resource#stage` takes either a directory or a block. In the former case it’ll unpack (read: uncompress) in it. In the latter it’ll create a temporary directory, `cd` and unpack in it, call the block, then `cd` back to where it was previously. Here we’re using a block so that unpacking the resource doesn’t mess with the buildpath.

This is what [`ansible.rb`](https://github.com/Homebrew/homebrew/blob/8e50348c825f9322d783105295d0c8c61be37896/Library/Formula/ansible.rb#L551-L553) does except that it uses `resources` to access all resources instead of retrieving each resource with its name using `resource("<name>")`.

bfontaine:
Ping?

luzpaz:
Help needed on this... please

ilovezfs:
@lupaz Does this help

https://github.com/Homebrew/homebrew-core/blob/master/Formula/mitmproxy.rb#L215-L223

apjanke:

PR #45683, opened 2015-11-04
Add /etc/shells caveat to zsh formula

herrbischoff:
As a result of a discussion held [here](https://github.com/herrbischoff/awesome-osx-command-line/pull/58#issuecomment-153714473), adding the caveat will advise users to add zsh to `/etc/shells` in order to use it as a login shell. Analoguous to the [bash formula](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/bash.rb#L44-L48).

### DIFF
--- a/Library/Formula/zsh.rb
+++ b/Library/Formula/zsh.rb
@@ -51,6 +51,8 @@ class Zsh < Formula
   end
 
   def caveats; <<-EOS.undent
+    In order to use this build of zsh as your login shell,
+    it must be added to /etc/shells.
     Add the following to your zshrc to access the online help:
       unalias run-help
       autoload run-help


DomT4:
Seems fine to me. Merged in https://github.com/Homebrew/homebrew/commit/f8691d004dd2baa867b913d15276eaeacc48b84a.

Thank you for your contribution to Homebrew @herrbischoff; we appreciate it! :smiley_cat:

herrbischoff:
Thanks! Just out of curiosity: why do you close pull requests instead of merging it? While the commit appears to remain in place, the "closed" marker just looks kind of bad in the profile. ;) There's probably a good reason for working this way and I'd like to understand.

DomT4:
We accept a bunch of commits every day & adding merge commits into that is both messy & makes it harder to people to find interesting things running git log locally, essentially. You still get commit credit, even with the big closed sign. If you're interested in the specific mechanism check out Library/Homebrew/cmd/pull.rb inside your Homebrew prefix :).

mikemcquaid:
> Thanks! You're welcome. Just out of curiosity: why do you close pull requests instead of merging it? While the commit appears to remain in place, the "closed" marker just looks kind of bad in the profile. ;) There's probably a good reason for working this way and I'd like to understand.

Basically we use a rebase rather than merge-based branch workflow.

herrbischoff:
Thanks! Could you give me a pointer to a basic workflow of Git commands to achieve this? I took a look at `pull.rb` but there's a lot going on in there.

mikemcquaid:
@herrbischoff It's kind of involved I'm afraid, if you're struggling to understand (and I hate to do this) I wrote a chapter in my book about it: https://www.manning.com/books/git-in-practice

herrbischoff:
Alright, thanks anyway. I'll take a look at that book, maybe it's for me.

PR #45688, opened 2015-11-04
Update docker-compose stable to 1.5.0

kudos:
1.5.0 went stable yesterday. Stable and HEAD are equivalent in dependencies right now.

### DIFF
--- a/Library/Formula/docker-compose.rb
+++ b/Library/Formula/docker-compose.rb
@@ -3,18 +3,8 @@ class DockerCompose < Formula
   homepage "https://docs.docker.com/compose/"
 
   stable do
-    url "https://github.com/docker/compose/archive/1.4.2.tar.gz"
-    sha256 "cc11f8281f0cf99fcb5502edb6e0d49caca26f4a11570b8ad68943bd3a97dd5c"
-
-    resource "docker-py" do
-      url "https://pypi.python.org/packages/source/d/docker-py/docker-py-1.3.1.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/docker-compose.rb
```diff
@@ -62,6 +27,31 @@ class DockerCompose < Formula
...
+    sha256 "36673ac378feed3daa5956276a829699056523d7961027911f064b52255ead41"
+  end
+
+  resource "functools32" do
+    url "https://pypi.python.org/packages/source/f/functools32/functools32-3.2.3-2.tar.gz"
+    sha256 "f6253dfbe0538ad2e387bd8fdfd9293c925d63553f5813c4e587745416501e6d"
+  end
+
   resource "pyyaml" do
     url "https://pypi.python.org/packages/source/P/PyYAML/PyYAML-3.11.tar.gz"
```
robsonpeixoto: Wrong PyYAML version. Must be 3.10
https://github.com/docker/compose/blob/master/requirements.txt#L1

robsonpeixoto:
The six version should be `1.7.3`.
https://github.com/docker/compose/blob/master/requirements.txt#L8

kudos:
They should all match now. For some reason the resources previously in `head do` were newer than the requirements file called for. I guess whoever originally added them just used the latest of each?

robsonpeixoto:
To be honest, I have no idea. But I guess the same, @kudos.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/c32f04bccfe4ba95a088010cf4174adda3e5dc5c. Thank you for your contribution to Homebrew @kudos; we appreciate it! :smiley_cat:

PR #45690, opened 2015-11-04
llvm: Add option to build Llvm dylib

zhihan:
This will enable the python bindings, which uses ctypes to load the libLLVM.dylib. 

Otherwise the python biding tries to load llvm dylib and errors out.

### DIFF
--- a/Library/Formula/llvm.rb
+++ b/Library/Formula/llvm.rb
@@ -88,7 +88,7 @@ class Llvm < Formula
   option "with-lld", "Build LLD linker"
   option "with-lldb", "Build LLDB debugger"
   option "with-rtti", "Build with C++ RTTI"
-  option "with-llvm-dylib", "Build shared library (libLLVM.dylib)" 
+  option "with-llvm-dylib", "Build shared library (libLLVM.dylib)"
   option "with-python", "Build Python bindings against Homebrew Python"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Is this perhaps worth enabling by default?

mikemcquaid:
CC @tdsmith for Python binding thoughts.

DomT4:
It's on by default in all the Versions formulae. Doesn't seem to have caused any harm at least.

zhihan:
I think the travis build failure is time out. But I cannot rerun it.

DomT4:
Ignore Travis on this one. As long as the one named `default` keeps passing, that's fine.

zhihan:
By the way, this is not the only issue of the python binding. The compiled library is libLLVM.dylib but the python binding looks for libLLVM-<version>.dylib. I also looked into a linux build it is named the same way. The python binding mentions 3.4, maybe this is changed from 3.4 to 3.6.2? 

Anyway, adding this option will make sure to build libLLVM.dylib.

zhihan:
What is the next step for me to do? Please let me know.

mikemcquaid:
Sorry, to check: is the Python binding working or not with this option? If it's not working: you should fix that too. If it is working: you can make this unconditional without an option. Thanks!

PR #45693, opened 2015-11-04
docker-swarm 1.0.0

pwnall:
I changed the `url` block to match the format in the docker formula. I hope that's OK.


bfontaine:
Is this necessary? If not we should use the tarball.

pwnall:
@bfontaine It's not necessary. I switched back to the tarball.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/7c21a3ecee9c2abf3b0db349e139f1a2cec0a08d. Thank you for another contribution to Homebrew @pwnall! :smiley_cat:

pwnall:
@DomT4 @bfontaine Thank you for the quick turnaround! :heart:

Issue #45655, opened 2015-11-03
superenv default - documentation bug?


dwalling:
I had the common "Segmentation fault: 11" issue with Python (trying to install boost-python for caffe). The Common Issues page says: "Check that `which python` points to the python you want" and "Homebrew formulae use a brewed Python if available or, if not so, they use whatever python is in your PATH."

On my system that is only true with --env=std. By default superenv ignores both PATH and the brewed Python and calls /usr/bin/python. Googling for help led me down many irrelevant paths. I got it working by hacking ENV["PATH"] in the Formula and then found the better solution:

```
brew remove boost-python
brew install --build-from-source --env=std boost-python
```

Is this the intended behavior? Should I take a crack at updating that Common Issues section?

https://gist.github.com/anonymous/247cd00d8bac6f5c72c0

mikemcquaid:
> Is this the intended behavior? 

Yep. Homebrew will always use the system Python unless `env=std` or `depends_on :python` is used. In the latter case it'll use the first Python in your PATH (unless it's <2.7 or you're installing from a bottle).

> Should I take a crack at updating that Common Issues section?

That'd be :heart_eyes:

DomT4:
Will let Tim/Mike review the PR you submitted in #45704, but closing the issue in favour of that. Thanks!

PR #45704, opened 2015-11-05
Update Common-Issues.md python seg fault section per #45655

dwalling:
These instructions worked for me, did I miss anything?

### DIFF
--- a/share/doc/homebrew/Common-Issues.md
+++ b/share/doc/homebrew/Common-Issues.md
@@ -106,23 +106,23 @@ may have been built with a different arch (e.g. Apple's python is still not a
 pure 64-bit build). Other things can go wrong, too. Welcome to the dirty
 underworld of C.
 
-To solve this, you should remove the problematic formula with those python
-bindings and all of its dependencies.
+To solve this first remove the problematic formula with those python
+bindings and all of its dependencies. `brew rm <problematic formula>` may be sufficient.
+If not:
 
   - `brew rm $(brew deps --installed <problematic_formula>)`
   - `brew rm <problematic_formula>`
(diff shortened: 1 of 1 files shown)


share/doc/homebrew/Common-Issues.md
```diff
@@ -106,23 +106,23 @@ may have been built with a different arch (e.g. Apple's python is still not a
...
-You can basically use any Python (2.x) for the bindings homebrew provides, but
-you can't mix.  Homebrew formulae use a brewed Python if available or, if not
-so, they use whatever `python` is in your `PATH`.
+
+Then reinstall with the correct python. By default Homebrew uses 
+--env:[super](Formula-Cookbook.md#superenv-notes), which ignores your `PATH` and
+finds /usr/bin/python. This is probably not what you want.
+  - Check that `which python` points to the python you want (edit your `PATH` if
+    not).
+  - `brew install --env:std <problematic_formula>`
```
mikemcquaid: We don't want to recommend `env:std` as it's now deprecated.

dwalling: Ok. What's the right way to specify which python to use? If it's TBD how about keeping this change with a deprecation warning and link to the issue about deprecating env:std?

I think at least this sentence should be changed, it took me a long time to figure out it wasn't true: "Homebrew formulae use a brewed Python if available or, if not so, they use whatever `python` is in your `PATH`."

mikemcquaid: It's not user-customisable unless a formula already has `depends_on :python` in which case it will use the Python in your `PATH`. Definitely agree this should be changed :+1:

PR #45707, opened 2015-11-05
cmd/install: skip installed formulae in search

dunn:
To avoid this:
```
🌰  brew install nex
Error: No available formula with the name "nex"
==> Searching for similarly named formulae...
These similarly named formulae were found:
git-annex (installed)                  homebrew/science/nexusformat           openexr (installed)
homebrew/science/nextflow              nexus                                  winexe
To install one of them, run (for example):
  brew install git-annex (installed)
==> Searching taps...
This formula was found in a tap:
Caskroom/cask/git-annex
To install it, run:
  brew install Caskroom/cask/git-annex
```
Should work whether it's ` (installed)` or the Unicode checkmark.

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -100,7 +100,8 @@ module Homebrew
         query = query_regexp(e.name)
 
         ohai "Searching for similarly named formulae..."
-        formulae_search_results = search_formulae(query)
+        formulae_search_results = search_formulae(query, :installed => false)
+
         case formulae_search_results.length
         when 0
           ofail "No similarly named formulae found."
--- a/Library/Homebrew/cmd/search.rb
+++ b/Library/Homebrew/cmd/search.rb
@@ -41,7 +41,7 @@ module Homebrew
     else
       query = ARGV.first
       rx = query_regexp(query)
-      local_results = search_formulae(rx)
+      local_results = search_formulae(rx, :decoration => :text)
       local_results_installed = local_results.select { |f| f.end_with? "(installed)" }
       puts_columns(local_results, local_results_installed)
       tap_results = search_taps(rx)
@@ -141,26 +141,42 @@ module Homebrew
     []
   end
 
-  def search_formulae(rx)
+  # @param [Regexp] rx
+  # @param [Hash] options
+  # @option options [Boolean] :installed If true, include installed formulae
+  # @option options [Symbol] :decoration
+  #
+  # @return [Array<String>]
+  def search_formulae(rx, options = {})
     aliases = Formula.alias_full_names
     results = (Formula.full_names+aliases).grep(rx).sort
 
     results.map do |name|
       begin
         formula = Formulary.factory(name)
+        next if formula.installed? && !options.fetch(:installed, true)
+
         canonical_name = formula.name
         canonical_full_name = formula.full_name
-      rescue
+      rescue FormulaUnavailableError
         canonical_name = canonical_full_name = name
       end
       # Ignore aliases from results when the full name was also found
-      if aliases.include?(name) && results.include?(canonical_full_name)
-        next
-      elsif (HOMEBREW_CELLAR/canonical_name).directory?
-        "#{name} (installed)"
-      else
-        name
+      next if aliases.include?(name) && results.include?(canonical_full_name)
+
+      decorate = options.fetch(:decoration, nil)
+      next name unless decorate
+
+      if (HOMEBREW_CELLAR/canonical_name).directory?
+        if decorate == :text
+          tag = " (installed)"
+        elsif decorate == :unicode
+          tag = " #{["2714".hex].pack("U*")}"
+        end
+      elsif decorate == :unicode
+        tag = " #{["2718".hex].pack("U*")}"
       end
+      "#{name}#{tag}"
     end.compact
   end
 end


bfontaine:
Shouldn’t we avoid telling the user how to install it if it’s already installed?

UniqMartin:
Maybe a cleaner approach would be to add options to `search_formulae` and `search_taps`, one that allows to skip installed formulae and one that disables formatting the results; then call them with these options?

dunn:
> Shouldn’t we avoid telling the user how to install it if it’s already installed?

Yep, I was tired and didn't realize that this is perfectly possible.

@UniqMartin oh yeah, maybe.  I'll take a look.

mikemcquaid:
> Shouldn’t we avoid telling the user how to install it if it’s already installed?

Agreed. Would also be good to have a consistent way to print out formulae lists that can have e.g. an `options` array passed to enable/disable columns, `(installed)`, filtering based on whether they are installed or not.

dunn:
Is this the sort of approach you had in mind?

I didn't add a separate option for toggling `(installed)` decoration because I couldn't think of a use-case where you want `:installed => true` but `:decorations => false`. Am I forgetting an obvious one?

Library/Homebrew/cmd/install.rb
```diff
@@ -100,7 +100,8 @@ def install
         query = query_regexp(e.name)
 
         ohai "Searching for similarly named formulae..."
-        formulae_search_results = search_formulae(query)
+        formulae_search_results = search_formulae(query, { :installed => false })
```
UniqMartin: I think its better and more customary to drop the braces and rely on Ruby to collect those “named arguments” into a hash and pass that as the last argument.

Library/Homebrew/cmd/search.rb
```diff
@@ -141,13 +141,20 @@ def search_tap(user, repo, rx)
...
   end
 
-  def search_formulae(rx)
+  # @param [Regexp] rx
+  # @param [Hash] options
+  # @option options [Boolean] :installed If true, include installed formulae
+  #
+  # @return [Array<String>]
+  def search_formulae(rx, options = { :installed => true })
     aliases = Formula.alias_full_names
```
UniqMartin: It's better to use an empty hash here and use `Hash#fetch` to deal with defaults. See [cmd/tap.rb](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/tap.rb#L62-L67) for an example.

Library/Homebrew/cmd/search.rb
```diff
@@ -141,13 +141,20 @@ def search_tap(user, repo, rx)
...
     results = (Formula.full_names+aliases).grep(rx).sort
 
     results.map do |name|
       begin
         formula = Formulary.factory(name)
+        next if formula.installed? && !options[:installed]
```
UniqMartin: … and `options[:installed]` would become `options.fetch(:installed, true)`.

UniqMartin:
> Is this the sort of approach you had in mind?

Yes, that's exactly how I envisioned it. Pointed out a few nits in the code, but nothing that detracts from the approach. And I think it's great that you also added some documentation. :heart_eyes:

Library/Homebrew/cmd/install.rb
mikemcquaid: You can just do `search_formulae(query, :installed => false)` I believe.

Library/Homebrew/cmd/search.rb
mikemcquaid: Agreed.

Library/Homebrew/cmd/search.rb
mikemcquaid: Agreed.

Library/Homebrew/cmd/search.rb
```diff
@@ -141,13 +141,20 @@ def search_tap(user, repo, rx)
     []
   end
 
-  def search_formulae(rx)
+  # @param [Regexp] rx
+  # @param [Hash] options
+  # @option options [Boolean] :installed If true, include installed formulae
```
xu-cheng: Maybe we can add an option(`:highlight_installed`), it can be used to turn off  highlight for non-stty cases.

xu-cheng: OR handle highlight automatically based on `tty`.

dunn: Isn't that already happening, at least with `puts_columns`? https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/utils.rb#L259-L265

xu-cheng: By not highlight, I mean not appending `(installed)`. Also current highlight function inside `puts_columns` is kinda of inefficient, for its unnecessary item replacement.

UniqMartin: Yes, to have an option to suppress the `(installed)` suffix would be useful to make the output of some commands more useful when used programmatically. I don't necessarily think this has to be part of this PR though. Once the option handling is in place, it's easy to add an additional option (and the code that uses it) in a separate PR.

dunn:
Here's a possible implementation.  Decoration is off by default, so that might be good enough for handling non-tty cases.

dunn:
The output before:
```
🌰  brew install nex
Error: No available formula with the name "nex"
==> Searching for similarly named formulae...
These similarly named formulae were found:
git-annex (installed)                  homebrew/science/nexusformat           openexr (installed)
homebrew/science/nextflow              nexus                                  winexe
To install one of them, run (for example):
  brew install git-annex (installed)
==> Searching taps...
This formula was found in a tap:
Caskroom/cask/git-annex
To install it, run:
  brew install Caskroom/cask/git-annex
```
Output after:
```
🌰  brew install nex
Error: No available formula with the name "nex"
==> Searching for similarly named formulae...
These similarly named formulae were found:
homebrew/science/nextflow                                 nexus
homebrew/science/nexusformat                              winexe
To install one of them, run (for example):
  brew install homebrew/science/nextflow
==> Searching taps...
This formula was found in a tap:
Caskroom/cask/git-annex
To install it, run:
  brew install Caskroom/cask/git-annex
```

And it would also make it easier to change to/from text/unicode decorations, not that this example is a good idea:
```
🌰  brew search nex
git-annex ✔                            homebrew/science/nexusformat ✘         openexr ✔
homebrew/science/nextflow ✘            nexus ✘                                winexe ✘
Caskroom/cask/git-annex
```

dunn:
@xu-cheng where were you thinking of placing the constants? Just in `global.rb`?

bfontaine:
You’ll need to rebase this on top of `master`.

dunn:
Sorry for letting this get stale.

I started updating it, but realized I don't think it's a good idea anymore.  If a user runs `brew install <something>` which isn't a formula, a search is run, and if the only result is something they already have installed, they will get `ofail "No similarly named formulae found."`  If they don't know that they have that formula installed (maybe that's what they're trying to install with `brew install <something>`), this will be very misleading.

PR #45708, opened 2015-11-05
jsoncpp: add universal option

d235j:
Adds a universal option to jsoncpp.


DomT4:
I presume these are all tested as universal builds?

d235j:
Yes — however https://github.com/Homebrew/homebrew/pull/45712 is necessary for it to work as expected.

PR #45709, opened 2015-11-05
ilmbase: add universal option

d235j:
Adds a universal option to ilmbase.


DomT4:
I presume these are all tested as universal builds?

d235j:
Yes, this has been tested.

PR #45712, opened 2015-11-05
cmake: don't hardcode bitness in .cmake files

d235j:
CMake normally hardcodes architecture in the .cmake files. For formulas that use a library's .cmake file for config, this causes build failure when the library is built as universal.
This is a direct port of https://trac.macports.org/ticket/41835 from MacPorts.

### DIFF
--- a/Library/Formula/cmake.rb
+++ b/Library/Formula/cmake.rb
@@ -13,6 +13,12 @@ class Cmake < Formula
     sha256 "cb0c73778120fcd899df10b1a06d9fbe1a1778645f252ac498bce1d0bd3ec63c" => :mavericks
   end
 
+  patch do
+    # skip bitness check (see https://cmake.org/Bug/view.php?id=15835)
+    url "https://gist.githubusercontent.com/d235j/695144126c2b9b8d5eec/raw/9c6f0a752c978ec35bd2ea9b423318b40120e78b/cmake-skip-archcheck.diff"
+    sha256 "ce0d53a362ac217680b11bef669d5992bc469a06e267e71075b1192f9ac88330"
+  end
+
   option "without-docs", "Don't build man pages"
   option "with-completion", "Install Bash completion (Has potential problems with system bash)"
 


DomT4:
Do you know if this ever got reported to CMake? If you can't find it on an upstream bug tracker somewhere can you submit the patch to them (Feel free to point to the original link if desired) and then include that link in the formula?

We don't really do the thing where we create patches ourselves and then don't pass them upstream for consideration and carry them forever regardless.

d235j:
This has not yet been reported to CMake. I'm in the process of putting together a somewhat better patch — ideally the .cmake file would be able to match against all architectures passed with -DCMAKE_OSX_ARCHITECTURES.

Unfortunately this patch is necessary for universal CMake formulas to work properly, if it is a library and another program is linking against it. Otherwise, CMake will fail due to unsupported arch.

d235j:
I have cleaned up the patch and submitted it upstream at https://cmake.org/Bug/view.php?id=15835.

mikemcquaid:
I'd like to see some upstream movement on this before we merge it. Thanks for submitting it though!

d235j:
I agree, but again, this bug completely breaks compilation of a program as non-universal against a CMake-compiled library built as universal, and there doesn't appear to be a better workaround, unless you can think of one...

mikemcquaid:
I agree but it's as-yet unclear if this patch will have any other negative effects. As it's fixing a CMake rather than a Homebrew-specific bug I'd rather wait for movement there, thanks!

mistydemeo:
> I agree but it's as-yet unclear if this patch will have any other negative effects.

As a data point, Macports has had a version of this patch for about two years without negative side effects.

Looks like there's some initial discussion on the cmake tracker, but if getting a patch in takes *too* long I wouldn't mind adopting some form of interim patch. They acknowledge there's an issue, at least.

mikemcquaid:
> As a data point, Macports has had a version of this patch for about two years without negative side effects.

It's disappointing that it wasn't submitted upstream :cry:. I'd still like to give it a little longer until CMake is happy with (or ideally merge) the patch.

d235j:
Any chance this can be reviewed? It seems there's no progress upstream, and this breaks universal builds with CMake.

mikemcquaid:
I don't want to accept this until upstream do, sorry.

mikemcquaid:
@d235j If you ask upstream more questions or email them directly you may have better luck writing a patch they might accept.

bfontaine:
Anything new here?

mikemcquaid:
Passing on this as-is, sorry.

d235j:
@mikemcquaid if this pull request is not acceptable, then an issue should probably be opened since this breaks CMake-built libraries that are built with --universal — thus rendering the --universal option useless for such libraries.

mikemcquaid:
@d235j We'll review a PR to remove the relevant options if you provide a reproducible test case, thanks.

PR #45715, opened 2015-11-05
kafka 0.8.2.2

esamson:


DomT4:
I suspect you need a `ENV["_JAVA_OPTIONS"] = "-Duser.home=#{testpath}"` line at the top of `def install` here.

Library/Formula/kafka.rb
```diff
@@ -26,6 +26,8 @@ def pour_bottle?
   end
 
   def install
+    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{testpath}"
```
dunn: Oops; in the install block it should be `buildpath` rather than `testpath`.

DomT4: Oops. That was my bad advice, apologies @esamson!

DomT4:
@BrewTestBot test this please

Library/Formula/kafka.rb
```diff
@@ -26,6 +26,8 @@ def pour_bottle?
   end
 
   def install
+    ENV["_JAVA_OPTIONS"] = "-Duser.home=#{buildpath}"
```
mikemcquaid: @DomT4 I'm wondering if we should just set this globally in all formulae?

mikemcquaid:

Library/Formula/kafka.rb
DomT4: Still waiting on the problem we discussed in the other thread weeks ago and Tim reported upstream :cry:

mikemcquaid: Sorry, not sure if I saw that. Could you link me when you get a chance?

DomT4: https://github.com/Homebrew/homebrew/pull/44567
https://issues.apache.org/jira/browse/MJAVADOC-435

mikemcquaid: Bah. Cheers!

PR #45721, opened 2015-11-05
add pkg-config to luabind

ajshort:


Library/Formula/luabind.rb
```diff
@@ -58,6 +58,22 @@ def install
     end
     args << "--prefix=#{prefix}"
     system "bjam", *args
+
+    (lib/"pkgconfig/luabind.pc").write pc_file
+  end
+
+  def pc_file; <<-EOS.undent
```
dunn: Since this is just used once it doesn't really need to be a method; may as well just do 
```ruby 
(lib/"pkgconfig/luabind.pc").write <<-EOS.undent
  prefix=#{HOMEBREW_PREFIX}
```

DomT4: Don't mind it this way personally. It's consistent with our [Luas](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/lua.rb#L114-L135).

DomT4:
Curious about the justification here. Do other package managers ship `pkg-config` files or did upstream previously ship one and no longer does or? Just want to understand why there's a need for a manual one now.

Library/Formula/luabind.rb
ajshort: Yep, this pattern is used through most other recipes. I actually asked someone to change it last time I just put the string in the install method.

ajshort:
AFAIK luabind never shipped with a pkg-config file. I personally find it very useful for discovering libraries. There are a number of other packages which already manually add a pkg-config file:

* mad
* squirrel
* mstch
* libid3tag
* tinyxml
* nss
* a few more

DomT4:
Checked Debian. Turns out they ship a `.pc` file, so I'm fine with this.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/aaf39feecc5f8f433113344e99f76bf502fff063. Thank you for another contribution to Homebrew @ajshort! :smiley_cat:

PR #45730, opened 2015-11-05
mpich 3.2rc2 (devel)

smseo:
Updated the mpich formula to use the latest develop branch of MPICH, mpich-3.2rc2.

### DIFF
--- a/Library/Formula/mpich.rb
+++ b/Library/Formula/mpich.rb
@@ -13,8 +13,8 @@ class Mpich < Formula
   end
 
   devel do
-    url "https://www.mpich.org/static/downloads/3.2b4/mpich-3.2b4.tar.gz"
-    sha256 "4fecce31b02095643a093aa01900fc1c6dec7690259c7459cc423f0ed10e949b"
+    url "https://www.mpich.org/static/downloads/3.2rc2/mpich-3.2rc2.tar.gz"
+    sha256 "00ab9725e6552f7c4f299a5f792d429d219e9de465e72478d69372cf86b9f8a5"
   end
 
   head do


s2mcom:
The Travis CI build failed because no fortran compiler was found. It was a configure failure.

mikemcquaid:
Hmm, not sure why the `depends_on :fortran` isn't working.

s2mcom:
Any suggestions to resolve this issue?

mikemcquaid:
I'm not sure. @xu-cheng @DomT4 @tdsmith @mistydemeo may know more about Fortran stuff.

xu-cheng:
Maybe `ENV.fortran if build.with? "fortran"` will help.

DomT4:
No idea I'm afraid. I try to avoid Fortran stuff.

apjanke:
I think the issue is that the `depends_on :fortran => :recommended` won't cause a fortran compiler to be installed; it just detects what is available on the system. In this normal case, it needs to have `gcc` installed since `gfortran` is supplied by it.

If I change it to an unconditional `depends_on :fortran`, then `brew install mpich` will automatically install `gcc`.

Sounds like the `default_formula` requirement resolution may need to be reviewed for `:recommended` dependencies.

Here's what I think is happening, based on adding some logging statements to `formula_installer.rb`. That `:recommended` `:fortran` requirement gets converted to a dependency on `gcc`. But then later on [in `formula_installer#expand_dependencies`](https://github.com/Homebrew/homebrew/blob/3126196ca45beb822d64ae4dd4962850bd0bb576/Library/Homebrew/formula_installer.rb#L325-L342), it checks that `gcc` dependency against `build.without?` and it returns false, so the installer skips it.

```
      if (dep.optional? || dep.recommended?) && build.without?(dep)
        puts "Pruning unused optional dependency #{dep}"
        Dependency.prune
```

mikemcquaid:
@apjanke Nice work finding the bug :+1:

smseo:
Thanks @apjanke.
@mikemcquaid I can modify `depends_on :fortran => :recommended` to `depends_on :fortran`, but I'm not sure it's a right fix. Is it better to wait until the bug above is resolved?

mikemcquaid:
> Is it better to wait until the bug above is resolved?

Yup, thanks!

smseo:
OK. Thanks!

zbeekman:
So, what's the status here? I'm a bit lost.

mikemcquaid:
> Here's what I think is happening, based on adding some logging statements to formula_installer.rb. That :recommended :fortran requirement gets converted to a dependency on gcc. But then later on in formula_installer#expand_dependencies, it checks that gcc dependency against build.without? and it returns false, so the installer skips it.

This is the issue here. `build.without?` needs fixed to handle this case and/or we should check it twice to check based on the actual requirement name that was expanded to the dependency.

mikemcquaid:
The dependency issue is fixed in https://github.com/Homebrew/homebrew/issues/46916. If you rebase this pull request it should be sorted.

smseo:
Thanks @mikemcquaid. I've rebased, but I think this commit is not meaningful because #47001 has a patch for MPICH 3.2 (stable release) and MPICH 3.2rc2 is not a devel version anymore. What do you think about closing this pull request?

UniqMartin:
@smseo If you could work with the author of #47001 and help out there, that would be awesome! I agree that working on this devel version doesn't make much sense anymore. Nonetheless, thanks for your work here and sorry for the issue that stalled this PR!

(Please let me know, if I misread your comment and thus closed this PR too early.)

UniqMartin:
@smseo Sorry, but when I wrote my last comment I didn't realize that other PR was basically done.

smseo:
@UniqMartin No problem. And, thanks for closing this request. I also saw other PR was done.

PR #45732, opened 2015-11-05
add a test to cvc4 to expose current bug

acfoltzer:
This breaks this formula, but the formula shouldn't currently be passing its tests as it produces broken binaries on El Capitan. The bug is known by the CVC4 team, and will be fixed in their 1.5 release which is scheduled for early this month.

### DIFF
--- a/Library/Formula/cvc4.rb
+++ b/Library/Formula/cvc4.rb
@@ -45,5 +45,15 @@ class Cvc4 < Formula
     EOS
     result = shell_output "#{bin}/cvc4 #{(testpath/"simple.cvc")}"
     assert_match /valid/, result
+    (testpath/"simple.smt").write <<-EOS.undent
+      (set-option :produce-models true)
+      (set-logic QF_BV)
+      (define-fun s_2 () Bool false)
+      (define-fun s_1 () Bool true)
+      (assert (not s_1))
+      (check-sat)
+    EOS
+    result = shell_output "#{bin}/cvc4 --lang smt #{(testpath/"simple.smt")}"
+    assert_match /unsat/, result
   end
 end


DomT4:
Looks like the problem is Xcode 7 rather than El Cap specifically, given this borked on Yosemite and El Cap?

acfoltzer:
Must be; I don't have a Yosemite machine handy for testing.

apjanke:
Is this the upstream bug report for it?

http://cvc4.cs.nyu.edu/bugs/show_bug.cgi?id=675

acfoltzer:
Yeah, that looks like the same bug, and it looks like their 1.5 release date has slipped :(

tdsmith:
Still no 1.5 release; what do you think makes sense here, @acfoltzer? Is there a patch we can apply?

acfoltzer:
Sadly I've not heard word of any progress on the release.

apjanke:
No cvc 1.5 release and no action on this bug. Looks like we're packaging broken software here, and upstream is moribund.

I don't know how popular cvc4 is and whether boneyarding is merited. But I think we should go ahead and merge this test so it's at least obvious that it's broken.

mikemcquaid:
I don't know how I feel about merging a broken test if we have no workaround for people.

apjanke:
It's segfaulting, and six months without a fix. I think that's "broken" by Homebrew standards. How about we boneyard it, and merge the test on its way to the boneyard as a reminder of why it was boneyarded?

mikemcquaid:
@apjanke 👍

DomT4:
If we boneyard this we'll need to boneyard `civl` as well, FYI.

apjanke:
https://github.com/Homebrew/homebrew-core/pull/909

acfoltzer:
As the original submitter of this formula, I'm aligned with boneyarding. Due to the instability we've moved over to `z3` for our SMT dependency for `cryptol`, so it won't be an impact there.

I took a brief look at the documentation for `civl`, and it sounds like `cvc4` is only an optional dependency that could similarly be dropped.

apjanke:
Cool. I'll take a look at `civl` too and see if we can drop the `cvc4` dep instead of boneyarding it.

acfoltzer:
FWIW the relevant section of the `civl` manual is here ([pdf](http://vsl.cis.udel.edu/lib/sw/civl/civl-manual.pdf)):
image

apjanke:
Yup. The README in the civl distribution agrees: it says "Install at least one of the theorem provers below" and then lists CVC3, CVC4, and Z3. Looks like it can run with just Z3, as your software can.

I've changed the PR to drop civl's cvc4 dependency instead of boneyarding it.

apjanke:
I merged in the test as part of https://github.com/Homebrew/homebrew-core/pull/909, which has been merged.

sfsiegel:
I am one of the CIVL developers.  It is true civl can work without cvc4, but it will work better with it.  Besides, CVC4 is a great tool for many reasons.  I hope the issue with it can be fixed.

DomT4:
Once upstream release CVC4 1.5, which allegedly will fix the problems, we can reinstate the formula and dependency as desired. Just waiting on upstream here before we can explore that, sadly.

apjanke:
I've gotten another request by email for CVC4 to be reinstated, too. As soon as one of us has time, let's ping CVC4 upstream about the 1.5 release schedule, or maybe an interim 1.4.1 release to backport the fix if it can be done and 1.5 is still a ways away..

And if there's enough demand, maybe we should just put 1.4 back in circulation; maybe as an optional component with a caveat. It's a bug and it's a crasher, but seems like people were still getting use out of it; I don't know how pervasive the effect of the bug is.

@sfsiegel Do you have any insight on the expected release schedule for CVC4 1.5, or how extensive the effect of this crash case is?

acfoltzer:
When I was first investigating this, it seemed like it was only an issue when using SMT-LIB input, rather than the CVC4-native input format. If people are getting use out of that still, it might be worth keeping around, but many applications like Cryptol only use SMT-LIB and will fail unexpectedly.

AlshammariA:
I will keep cvc4 team updated about this issue, so they can fix it or at least has a certain time of cvc4 1.5 release. I highly respect Homebrew rules regarding some packages that have bugs like cvc4. I think if we can have a caveat about this issue, so all users and even cvc4 team can track this ticket as long is it active. As a CIVL user, even though I can work without cvc4, it is still important for CIVL user for good performance. Thank you @apjanke and @acfoltzer for pointing out this bug and for your other suggestions. I already emailed cvc4 and provided to them this request link for participating

mikemcquaid:
I'm OK with whatever but people passionate about cvc4 need to be the people addressing these concerns with pull requests. Thanks!

barrettcw:
Hi all - I'm one of the project leaders for CVC4.  For various reasons, a 1.5 release has been delayed several times, but I would be happy to provide an intermediate release for your purposes.  The current master is quite stable.  Let me consult with the team and see if we can't get that to you right away.

apjanke:
That would be ideal. Thanks @barrettcw!

barrettcw:
Hi all - just submitted a pull request to boneyard with an updated cvc4 formula.  Sorry it took so long.  Hope this solves the issues for now.  Feel free to contact me in the future with any cvc4 issues.

PR #45742, opened 2015-11-06
Updated formula for 'premake'.

georgewhite5:
The premake formula now uses the downloads available on the website.


georgewhite5:
I've also added a `premake5` formula.

PR #45744, opened 2015-11-06
premake 4.4-beta5, 5.0.0-alpha6 (devel)

georgewhite5:
The premake formula now uses the downloads available on their website.


Library/Formula/premake5.rb
```diff
@@ -0,0 +1,18 @@
+class Premake5 < Formula
+  desc "Write once, build anywhere Lua-based build system"
+  homepage "https://premake.github.io/"
+  url "https://github.com/premake/premake-core/releases/download/v5.0.0-alpha6/premake-5.0.0-alpha6-src.zip"
+  sha256 "9c13372699d25824cba1c16a0483507a6a28903e2556ffb148b288c189403aee"
+
+  bottle do
```
DomT4: Don't need this block.

Library/Formula/premake.rb
```diff
@@ -1,8 +1,8 @@
 class Premake < Formula
-  desc "Premake is a build script generator"
-  homepage "http://industriousone.com/premake"
-  url "https://downloads.sourceforge.net/project/premake/Premake/4.3/premake-4.3-src.zip"
-  sha256 "36536490f8928d8ecde135da80cd8b751ea5bebe50cabba5c0de49cd41cb2780"
+  desc "Write once, build anywhere Lua-based build system"
+  homepage "https://premake.github.io/"
+  url "https://downloads.sourceforge.net/project/premake/Premake/4.4/premake-4.4-beta5-src.zip"
```
DomT4: We don't support unstable versions as the primary URL in the core.

Library/Formula/premake5.rb
```diff
@@ -0,0 +1,18 @@
+class Premake5 < Formula
```
DomT4: We don't really support side-by-side versions in the core, unless it's something like Python which is extremely popular, or Lua which has remaining formula deps to resolve before the old version can be booted to Versions.

mikemcquaid: Agreed, just update the current version to 5.

georgewhite5: Can premake5 execute premake4.lua files? Version 5 completely revamps the structure of the tool, @mikemcquaid @DomT4

Library/Formula/premake.rb
georgewhite5: This is the latest version of premake, @DomT4. I don't understand why they call it a beta, but this is the only version of premake4 they offer on their website.

Library/Formula/premake5.rb
DomT4: No idea. Don't use it enough to know the answer to that question. Essentially, whichever is going to be less useful *for now* should live in Versions.

Library/Formula/premake.rb
DomT4: It might be worth checking on that upstream and asking why it's been labelled a beta.

Library/Formula/premake5.rb
mikemcquaid: Basically unless this breaks Homebrew-stuff using `premake`: we should package the latest version.

Library/Formula/premake.rb
apjanke: Looks like development on 4.x has basically stalled out in favor of 5.x, and no real 4.4 release has been made yet. That https://premake.github.io is a pretty thin landing page that they're still working on filling out.

https://github.com/premake/premake-4.x/wiki/Development_Roadmap

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!searchin/premake-development/4.4/premake-development/i1uA1Wk6zYM/kbp9q4Awu70J

If you really need access to the 4.4 prereleases, you could add it as a `devel` option.

georgewhite5: @apjanke Premake 5 is more important for me.

bfontaine:
Ping?

georgewhite5:
@bfontaine This is the current situation: many programs use premake5, but premake5 is alpha software. premake4 is available in the repositories, but does not support premake5 scripts. Releases are no longer available on their website.

bfontaine:
@georgewhite5 If upstream doesn’t provide a stable version for premake 5 nor 4.4 I’m afraid we can’t update the stable URL for this formula. Maybe the solution is to update the `devel` block to use premake 5?

georgewhite5:
This sounds OK to me, but some users may require premake5 and premake4 to co-exist. I'm only using premake5 because I'm using it in my own code.

georgewhite5:
@bfontaine @DomT4 If you visit the download page on Premake's GitHub Pages website, it claims that Premake 4.4 (beta) is 'the latest in Premake's 4.x stable branch'.

apjanke:
@BrewTestBot test this please

apjanke:
The CI run is failing, but in the "procedural" stage.

```
	Test Result (2 failures / +2)

    brew-test-bot.mavericks.pull --clean --legacy https://github.com/Homebrew/legacy-homebrew/pull/45744
    brew-test-bot.mavericks.readall --aliases
```

I'm going to try to just migrate this to the new homebrew/core and see how it goes there.

PR #45747, opened 2015-11-06
carthage 0.10

dlackty:
See also #43884.
This should close Carthage/Carthage#769 as well.

PR #45748, opened 2015-11-06
Carthage 0.10

uny:
Lively discussion is [here](https://github.com/Homebrew/homebrew/pull/43884#issuecomment-154289599), but I would like this version soon via Homebrew.


uny:
Same for https://github.com/Homebrew/homebrew/pull/45747.

PR #45769, opened 2015-11-07
Debian various.

DomT4:
* Bottle generation for newer OS X versions
* Moving away from `https://mirrors.kernel.org/debian` as it uses a SHA1 certificate chain which causes it to no longer be shown as secure in Google Chrome, which looks more worrying than it technically is but is an easy enough thing to fix.
* Adding fallback mirrors.
* Improving/Adding tests.

All these are <2 minute compiles, hence combining 9 of them in the same PR. All built and tested locally first, so should be fine.

### DIFF
--- a/Library/Formula/logcheck.rb
+++ b/Library/Formula/logcheck.rb
@@ -1,8 +1,8 @@
 class Logcheck < Formula
   desc "Mail anomalies in the system logfiles to the administrator"
   homepage "https://logcheck.alioth.debian.org/"
-  url "https://mirrors.kernel.org/debian/pool/main/l/logcheck/logcheck_1.3.17.tar.xz"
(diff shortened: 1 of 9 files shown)


mikemcquaid:
:+1:

DomT4:
The `urlview` test works locally, works on Travis, breaks on Jenkins.

```ruby
  test do
    (testpath/"urls").write "https://github.com/Homebrew/homebrew"
    io = IO.popen("#{bin}/urlview urls")
    sleep 1
    Process.kill("SIGINT", io.pid)
    Process.wait(io.pid)
    assert_match %r{https://github.com/Homebrew/homebrew}, io.read
  end
```

Will pull it out of the formula for now.

PR #45771, opened 2015-11-07
Various.

DomT4:
Added/Improved tests. Bunch o' minor fixes. All built and tested locally.

Mostly killing off some old SHA1 bottles.


DomT4:
Heh. Should have guessed a network test would die on the CI. Will pull that out.

PR #45773, opened 2015-11-07
auto install certain external commands

xu-cheng:
* Replace `install_tap` with `Tap#install`.
* Auto install certain external commands, therefore it will behavior like they are pre-installed.

cc @vitorgalvao

### DIFF
--- a/Library/Homebrew/cmd/tap.rb
+++ b/Library/Homebrew/cmd/tap.rb
@@ -1,5 +1,4 @@
 require "tap"
-require "descriptions"
 
 module Homebrew
   def tap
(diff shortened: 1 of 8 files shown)


Library/brew.rb
```diff
@@ -152,6 +152,28 @@ def require?(path)
...
+    end
+
+    if possible_tap && !possible_tap.installed?
+      possible_tap.install
+
+      if cmd == "cask"
+        require "cmd/install"
+        brew_cask = Formulary.factory("brew-cask")
+        Homebrew.install_formula(brew_cask)
+      end
```
xu-cheng: It would be prefect, if `cask` is a tap external commands rather than standalone formula.

mikemcquaid: Agreed. One of us should probably submit a PR to cask to do that for them.

Library/Homebrew/cmd/tap.rb
```diff
@@ -14,48 +13,22 @@ def tap
...
     else
       user, repo = tap_args
-      clone_target = ARGV.named[1]
-      opoo "#{user}/#{repo} already tapped!" unless install_tap(user, repo, clone_target)
+      tap = Tap.fetch(user, repo)
+      tap.install(:clone_target => ARGV.named[1], :full_clone => ARGV.include?("--full"))
     end
   end
 
+  # @deprecated this method will be removed in the future, if no external commands use it.
```
mikemcquaid: Maybe print a warning inside the method saying that?

mikemcquaid: Oh, you already did :grinning:

Library/brew.rb
```diff
@@ -152,6 +152,28 @@ def require?(path)
...
   else
+    require "tap"
+    possible_tap = case cmd
+    when *%w[brewdle brewdler bundle bundler]
+      Tap.fetch("Homebrew", "bundle")
+    when "cask"
+      Tap.fetch("caskroom", "cask")
+    when "services"
+      Tap.fetch("Homebrew", "services")
+    end
```
mikemcquaid: Very nice.

mikemcquaid:
Love this :heart_eyes:

vitorgalvao:
So as I understand it, as soon as someone `brew cask {{whatever}}`, homebrew will take care of all the dirty work of tapping and whatnot. Brilliant! I like this very much. [One less thing to worry about](https://github.com/caskroom/homebrew-cask/pull/15142/files).

Pinging the rest of @caskroom.

xu-cheng:
> So as I understand it, as soon as someone brew cask {{whatever}}, homebrew will take care of all the dirty work of tapping and whatnot. 

That's right. 

Also as mentioned in above command, it will offer better user experience if cask is a tap external command rather than standalone formula.

jawshooah:
This is great! One step closer to full integration :smile:

We would very much appreciate any help in transitioning from a formula to an external command.

victorpopkov:
Amazing! Definitely makes the life simpler. Really like this feature. :+1:

fanquake:
👍🏼

PR #45775, opened 2015-11-07
nodebrew 0.9.2

lion-man44:
Hi.

Update nodebrew 0.9.2 version

### DIFF
--- a/Library/Formula/nodebrew.rb
+++ b/Library/Formula/nodebrew.rb
@@ -1,8 +1,8 @@
 class Nodebrew < Formula
   desc "Node.js version manager"
   homepage "https://github.com/hokaccha/nodebrew"
-  url "https://github.com/hokaccha/nodebrew/archive/v0.8.1.tar.gz"
-  sha256 "c976a879954dfb5dd657f1e93934ec182ffe4aa4e91ee2be0f7ece91cd25af6f"
+  url "https://github.com/hokaccha/nodebrew/archive/v0.9.2.tar.gz"
+  sha256 "ebd9df28fd9a65bec3cbab6e115300f63a012cba80a27cc8d34ab8ccf9c1d7c1"
   head "https://github.com/hokaccha/nodebrew.git"
 
   def install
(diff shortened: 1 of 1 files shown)


Library/Formula/nodebrew.rb
```diff
@@ -22,6 +22,6 @@ def caveats; <<-EOS.undent
   end
 
   test do
-    assert shell_output("#{bin}/nodebrew ls-remote").include?("v0.10.0")
+    assert_match version.to_s, shell_output("#{bin}/nodebrew --version | egrep \"nodebrew \\d\"")
```
bfontaine: This ` | egrep \"nodebrew \\d\"` shouldn’t be necessary here.

xu-cheng: Also, why replace the test? The old test `nodebrew ls-remote` is preferred than simple `--version` test.

DomT4: Maybe the audit complained about `assert...include`?

```ruby
assert_match /v0.10.0/, shell_output("#{bin}/nodebrew ls-remote")
```
Should work on that though.

lion-man44: @DomT4 
> Maybe the audit complained about assert...include?

Yes.

> Should work on that though.

It's good.
But, what is `v0.10.0`?

DomT4: > But, what is v0.10.0?

Don't use Nodebrew personally but the naming of `ls-remote` would suggest its purpose is to list all the Nodes available for installation, and back in 2013 or so there was a Node `v0.10.0` release which this presumably matches.

lion-man44: Oh, sorry.
Typo what -> why.

Thanks for advice.
Code is fixed.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/f75a13fa099200501c2a13c5e01180e5120682f4. Thank you for your contribution to Homebrew @lion-man44; we appreciate it! :smiley_cat:

PR #45778, opened 2015-11-07
iproute2mac v1.1.0 (new formula)

brona:
New formula for iproute2mac CLI tool as discussed in #32944.

### DIFF
--- a/Library/Formula/iproute2mac.rb
+++ b/Library/Formula/iproute2mac.rb
@@ -1,0 +1,21 @@
+class Iproute2mac < Formula
+  desc "CLI wrapper for basic network utilities on Mac OS X - ip command"
+  homepage "https://github.com/brona/iproute2mac"
+  url "http://github.com/brona/iproute2mac/archive/v1.1.0.tar.gz"
+  sha256 "3149dc23d5487e837646f50572d512a31799c74cf21288609887cbf8e465d4e6"
+
+  depends_on :python
+
+  def install
+    bin.install "src/ip.py" => "ip"
+  end
+
+  test do
+    pipe_output("#{bin}/ip route")
+    assert $?.success?, "Calling route failed"
+    pipe_output("#{bin}/ip address")
+    assert $?.success?, "Calling ifconfig failed"
+    pipe_output("#{bin}/ip neigh")
+    assert $?.success?, "Calling arp failed"
+  end
+end


Library/Formula/iproute2mac.rb
```diff
@@ -0,0 +1,21 @@
...
+    bin.install "src/ip.py" => "ip"
+  end
+
+  test do
+    pipe_output("#{bin}/ip route")
+    assert $?.success?, "Calling route failed"
+    pipe_output("#{bin}/ip address")
+    assert $?.success?, "Calling ifconfig failed"
+    pipe_output("#{bin}/ip neigh")
+    assert $?.success?, "Calling arp failed"
```
bfontaine: You can remove all these assertions and use `system "#{bin}/ip", "route"` instead of `pipe_output("#{bin}/ip route")` and the same for all `pipe_output` calls.

The `system` used here is a Homebrew wrapper that already takes care of checking `$?.success?`.

brona: Fixed in 23e03e9, was not really sure what system() does. Thanks for advice.

Library/Formula/iproute2mac.rb
```diff
@@ -0,0 +1,18 @@
...
+  depends_on :python
+
+  def install
+    bin.install "src/ip.py" => "ip"
+  end
+
+  test do
+    system("#{bin}/ip", "route")
+    system("#{bin}/ip", "address")
+    system("#{bin}/ip", "neigh")
```
bfontaine: Small nit: you can drop the parentheses here.

Library/Formula/iproute2mac.rb
```diff
@@ -0,0 +1,18 @@
+class Iproute2mac < Formula
+  desc "CLI wrapper for basic network utilities on Mac OS X - ip command"
+  homepage "https://github.com/brona/iproute2mac"
+  url "http://github.com/brona/iproute2mac/archive/v1.1.0.tar.gz"
+  sha256 "3149dc23d5487e837646f50572d512a31799c74cf21288609887cbf8e465d4e6"
+
+  depends_on :python
```
bfontaine: @tdsmith Is `if MacOS.version <= :snow_leopard` needed here?

DomT4: I expect so, otherwise it'll be bottled against our Python.

brona: Added if `MacOS.version <= :snow_leopard` in  0a71f28 according to https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md#python-declarations

Is there anything else I can do to help push this forward?

PR #45784, opened 2015-11-07
Not everything under /usr/local is owned by brew

gammamatrix:
If you have MySQL installed from http://www.mysql.com/downloads/, running a recursive chmod will mess with your database.

MySQL installs too

    /usr/local/mysql-5.6.23-osx10.8-x86_64/

and sym links to

    /usr/local/mysql


mikemcquaid:
Passing on this for now. We have plans to improve this situation (a `brew repair-permissions` command) or similar) but I don't want to add a list of special cases to here for now, sorry! Thanks!

gammamatrix:
I understand that. When it is updated, it would be proper to not say `chmod -R`.

mikemcquaid:
Thanks!

PR #45785, opened 2015-11-07
anjuta 3.18.2

tschoonj:


mikemcquaid:
Failing on 10.10/10.9: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/36119/

tschoonj:
For some reason the Homebrew python is required on these two platforms when running itstool, which already has a dependency on `:python`. Strange...

DomT4:
`itstool` has an unconditional `:python` dep which is probably the cause of that. CC @tdsmith in case he knows the background of that dep.

tschoonj:
Same issue as in #45845. @tdsmith could you shine some light on this?

Thanks

mikemcquaid:
@BrewTestBot test this please

tschoonj:
Excellent! Both look ready now

mikemcquaid:
Thanks @tschoonj!

PR #45786, opened 2015-11-07
2ping 3.0.1 (new formula)

rfinnie:
2ping is a bi-directional ping utility. It uses 3-way pings (akin to TCP
SYN, SYN/ACK, ACK) and after-the-fact state comparison between a 2ping
listener and a 2ping client to determine which direction packet loss
occurs.

New formula twoping.rb with a 2ping alias.

### DIFF
--- a/Library/Formula/twoping.rb
+++ b/Library/Formula/twoping.rb
@@ -1,0 +1,19 @@
+class Twoping < Formula
+  desc "Ping utility to determine directional packet loss"
+  homepage "http://www.finnie.org/software/2ping/"
+  url "http://www.finnie.org/software/2ping/2ping-3.0.1.tar.gz"
+  sha256 "d6997cd1680151e6f7d5e60137d45cd41bf385d26029878afdaaf5dc4f63dcc4"
+
+  def install
+    ENV.prepend_create_path "PYTHONPATH", libexec/"lib/python2.7/site-packages"
+    system "python", *Language::Python.setup_install_args(libexec)
+    man1.install "doc/2ping.1"
+    man1.install_symlink "2ping.1" => "2ping6.1"
+    bin.install Dir["#{libexec}/bin/*"]
+    bin.env_script_all_files(libexec/"bin", :PYTHONPATH => ENV["PYTHONPATH"])
+  end
+
+  test do
+    system bin/"2ping", "-c", "5", "test.2ping.net"
+  end
+end
--- a/Library/Aliases/2ping
+++ b/Library/Aliases/2ping
@@ -1,0 +1,1 @@
+../Formula/twoping.rb


Library/Formula/twoping.rb
```diff
@@ -0,0 +1,18 @@
+class Twoping < Formula
+  desc "Ping utility to determine directional packet loss"
+  homepage "http://www.finnie.org/software/2ping/"
+  url "http://www.finnie.org/software/2ping/2ping-3.0.1.tar.gz"
+  sha256 "d6997cd1680151e6f7d5e60137d45cd41bf385d26029878afdaaf5dc4f63dcc4"
+
+  depends_on "Python"
```
mistydemeo: Looks like this works with Python 2.6 and 2.7, so you should be able to rely on the version of Python that comes with the OS.

rfinnie: Correct, it'll work out of the box with system OS X Python 2.7 (or any 2.6/2.7).  Looks like Leopard was 2.5, Snow Leopard was 2.6, so should I replace that with:

```
depends_on :python if MacOS.version <= :leopard
```

(Modification of example from Python-for-Formula-Authors.md which suggests "<= :snow_leopard" for 2.7-only programs.)

rfinnie: Actually, 2ping is an easy-install program which relies on the installed twoping module (not a single script).  With the Python formula removed, it warns about site-packages during install:

```
ryan@ndnd:~$ brew install 2ping
==> Downloading http://www.finnie.org/software/2ping/2ping-3.0.1.tar.gz
Already downloaded: /Library/Caches/Homebrew/twoping-3.0.1.tar.gz
==> python -c import setuptools... --no-user-cfg install --prefix=/usr/local/Cellar/twoping/3.0.1 --single-version-externally-managed --record=installed.txt
==> Caveats
Python modules have been installed and Homebrew's site-packages is not
in your Python sys.path, so you will not be able to import the modules
this formula installed. If you plan to develop with these modules,
please run:
  mkdir -p /Users/ryan/Library/Python/2.7/lib/python/site-packages
  echo 'import site; site.addsitedir("/usr/local/lib/python2.7/site-packages")' >> /Users/ryan/Library/Python/2.7/lib/python/site-packages/homebrew.pth
==> Summary
🍺  /usr/local/Cellar/twoping/3.0.1: 27 files, 244K, built in 2 seconds
```

and 2ping fails:

```
ryan@ndnd:~$ 2ping
  File "/usr/local/bin/2ping", line 5, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 2793, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 673, in require
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 576, in resolve
    raise DistributionNotFound(req)
pkg_resources.DistributionNotFound: 2ping==3.0.1
```

The stated workaround works but users can't be expected to do that, correct?  Does that mean homebrew-managed python is effectively required?

rfinnie:
Full disclaimer: self-submitted project, but 2ping should fulfill the requirements under Acceptable Formulae:

 * Mature project (5 years old)
 * Known and used; high Debian popcon: https://qa.debian.org/popcon.php?package=2ping

Library/Formula/twoping.rb
mikemcquaid: You can just remove it altogether, actually.

mikemcquaid: (the `depends_on`)

Library/Formula/twoping.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def install
+    system "python", *Language::Python.setup_install_args(prefix)
+    man1.install "doc/2ping.1"
+    man1.install_symlink "2ping.1" => "2ping6.1"
+  end
+
+  test do
+    assert_equal "#{version}", shell_output("#{bin}/2ping -V 2>&1").chomp
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

rfinnie: 2ping is a network utility.  It includes both the client and listener (server) components, so it would be possible to set up a listener and then have the client test against it and then stop the listener, but my Ruby experience is limited and I wouldn't know how to handle the concurrency (unless you know of an existing formula which does something similar).

Alternatively, test.2ping.net is a permanent listener which the client could test against, but that means relying on an external service to test.  I'm not sure how acceptable that is w/r/t CI, policy, etc.

Another alternative is simply e.g. "2ping --listen --deadline=5", which would start up the listener, do nothing for 5 seconds and exit, which is probably better than nothing.

Library/Formula/twoping.rb
rfinnie: Removing the depends_on seems to be no different than the condition depends_on; without the python formula installed it still installs the egg to the local site-packages, which the system python cannot find without the manual workaround specified in the Caveats.

Library/Formula/twoping.rb
mikemcquaid: > Alternatively, test.2ping.net is a permanent listener which the client could test against, but that means relying on an external service to test. I'm not sure how acceptable that is w/r/t CI, policy, etc.

That's acceptable :+1:

Library/Formula/twoping.rb
mikemcquaid: See `ansible.rb` for an example of using a `PYTHONPATH` wrapper for this case.

rfinnie:
Thanks @mikemcquaid.  Removed `depends_on "Python"` and successfully running against system Python, and test now tests against test.2ping.net.

Library/Formula/twoping.rb
mistydemeo: Homebrew doesn't support any OS older than 10.6, so don't worry about supporting Leopard here. I can put work into that in Tigerbrew (the Homebrew fork for older Macs).

mikemcquaid:

PR #45787, opened 2015-11-07
llvm 3.6.2 Link to Homebrew Python.

Sean1708:
llvm currently links to the system Python, this causes the interpreter to crash when importing the lldb-python bindings in Homebrew Python. This commit links llvm to the Homebrew Python instead, fixing this.

### DIFF
--- a/Library/Formula/llvm.rb
+++ b/Library/Formula/llvm.rb
@@ -135,6 +135,15 @@ class Llvm < Formula
 
     args << "-DLLVM_ENABLE_RTTI=On" if build.with? "rtti"
 
+    if build.with? "python"
+      # LLVM links against `/usr/bin/python`, instead check `python-config` for
+      # the user's preferred python installation.
+      # Ref: https://llvm.org/bugs/show_bug.cgi?id=25446
+      python_prefix = `python-config --exec-prefix`.strip
+      args << "-DPYTHON_INCLUDE_DIR='" + python_prefix + "/include/python2.7'"
+      args << "-DPYTHON_LIBRARY='" + python_prefix + "/lib/libpython2.7.dylib'"
+    end
+
     if build.with? "assertions"
       args << "-DLLVM_ENABLE_ASSERTIONS=On"
     else


mikemcquaid:
Seems reasonable to me but CC @tdsmith just in case.

Library/Formula/llvm.rb
```diff
@@ -94,7 +94,7 @@ class Llvm < Formula
   deprecated_option "rtti" => "with-rtti"
   deprecated_option "disable-assertions" => "without-assertions"
 
-  if MacOS.version <= :snow_leopard
+  if MacOS.version <= :snow_leopard || build.with?("python")
```
DomT4: Is this not triggered automatically if you `brew install llvm --with-python`? That invokes the `depends_on :python` dep below, no?

Sean1708: Hmmm, it doesn't on my machine.

DomT4: Hmm. The logic should work like this AFAIK:

* MacOS = Snow Leopard or below? Yes => Use Homebrew Python. No => Use System Python.
* MacOS = Lion or above? Yes => Requested Python build? No => Use system. Yes => Use Homebrew's.

The second is conditional on Homebrew's Python being in your `$PATH`.

Maybe I'm wrong.

Sean1708: So it turns out that using `depends_on :python` _doesn't_ bring Python in as a dependency but `depends_on "python"` _does_. Is this a bug in Homebrew or in the Formula?

DomT4: `depends_on :python` is [special](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/requirements/python_requirement.rb).

Sean1708: Ok that makes sense, I'll fix put that part back to how it was originally.

tdsmith:
This is a workaround for this particular situation but anything `import`able should never link to any python library.

Since this is off by default, depending on :python is fine for now, but you should invoke `python-config` to learn these paths instead of assuming that Homebrew python is installed (:python can be satisfied by system python, Homebrew python, or a python.org python).

Sean1708:
The problem is that LLVM links to `/usr/bin/python` unless you pass `-DPYTHON_INCLUDE_DIR` and `-DPYTHON_LIBRARY` to `cmake`.

Is there a way to check whether the user has Homebrew Python?

Or is there a better way to solve this problem in general?

tdsmith:
Yes; it's an LLVM bug that LLVM links against Python.

Instead of checking whether a user has Homebrew python installed, you should invoke `python-config` and use its output to construct paths, like ``library = `python-config --exec-prefix`.strip + "/lib/libpython2.7.dylib"``.

tdsmith:
upstream bug: https://llvm.org/bugs/show_bug.cgi?id=25446

Sean1708:
Ok I've fixed it to use `python-config` instead, but if this is an upstream bug maybe it's best to just not bother and wait 'til LLVM fix it?

tdsmith:
No progress upstream; merging this seems fine if we aren't doing this yet.

tdsmith:
It looks like we always build Python bindings; by default against system but we allow switching to Homebrew Python by passing `--with-python`.

PR #45789, opened 2015-11-08
Add flif 0.1 (new formula)

yumitsu:
FLIF is a lossless image format based on MANIAC (Meta-Adaptive Near-zero Integer Arithmetic Coding) compression.

### DIFF
--- a/Library/Formula/flif.rb
+++ b/Library/Formula/flif.rb
@@ -1,0 +1,24 @@
+class Flif < Formula
+  desc "Free Lossless Image Format"
+  homepage "http://flif.info"
+  url "https://github.com/FLIF-hub/FLIF/archive/v0.1-alpha.tar.gz"
+  version "0.1-alpha"
+  sha256 "b15229e07b43758bb993aa912a08d66f62e078e3f4964b13ffe7a13b6b3ecd07"
+
+  head "https://github.com/FLIF-hub/FLIF.git"
+
+  depends_on "pkg-config" => :build
+  depends_on "libpng"
+
+  def install
+    system "make"
+
+    bin.mkpath
+    bin.install "flif"
+    include.install %w[flif.h flif_config.h]
+  end
+
+  test do
+    system bin/"flif", "--help"
+  end
+end


DomT4:
Can you please squash the 7 commits back into 1 here? Thanks.

yumitsu:
@DomT4 Done.

Library/Formula/flif.rb
```diff
@@ -0,0 +1,24 @@
...
+  def install
+    system "make"
+
+    bin.mkpath
+    bin.install "flif"
+    include.install %w[flif.h flif_config.h]
+  end
+
+  test do
+    system bin/"flif", "--help"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/flif.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  depends_on "pkg-config" => :build
+  depends_on "libpng"
+
+  def install
+    system "make"
+
+    bin.mkpath
+    bin.install "flif"
+    include.install %w[flif.h flif_config.h]
```
mikemcquaid: Is there a `make install` we can use?

Library/Formula/flif.rb
```diff
@@ -0,0 +1,24 @@
+class Flif < Formula
+  desc "Free Lossless Image Format"
+  homepage "http://flif.info"
+  url "https://github.com/FLIF-hub/FLIF/archive/v0.1-alpha.tar.gz"
+  version "0.1-alpha"
```
mikemcquaid: Unless there's a stable release this should probably live in `homebrew/homebrew-devel-only`. I'm happy to review it here until then, though.

Library/Formula/flif.rb
yumitsu: @mikemcquaid Thanks for tip, I'll take a look.

Library/Formula/flif.rb
yumitsu: @mikemcquaid Unfortunately, no.

Library/Formula/flif.rb
yumitsu: @mikemcquaid OK, I will make a PR to `homebrew-devel-only` after we finish with this review.

DomT4:
Please submit this to `devel-only`, Thanks!

PR #45792, opened 2015-11-08
groonga: add HEAD support

cosmo0920:
Some of Groonga users want to try new features as soon as possible.
It's reasonable to support build with `--HEAD` option.
How do you think about it?

### DIFF
--- a/Library/Formula/groonga.rb
+++ b/Library/Formula/groonga.rb
@@ -39,12 +39,14 @@ class Groonga < Formula
       --without-libstemmer
     ]
 
-    args << "--with-ruby" if build.head?
     args << "--enable-benchmark" if build.with? "benchmark"
     args << "--with-mecab" if build.with? "mecab"
     args << "--with-lz4" if build.with? "lz4"
 
-    system "./autogen.sh" if build.head?
+    if build.head?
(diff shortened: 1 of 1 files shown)


Library/Formula/groonga.rb
```diff
@@ -32,10 +39,12 @@ def install
       --without-libstemmer
     ]
 
+    args << "--with-ruby" if build.head?
     args << "--enable-benchmark" if build.with? "benchmark"
     args << "--with-mecab" if build.with? "mecab"
     args << "--with-lz4" if build.with? "lz4"
 
+    system "./autogen.sh" if build.head?
```
DomT4: You could put this in one block :)

```ruby
if build.head?
 args << "--with-ruby"
 system "./autogen.sh"
end
```

cosmo0920: Thanks, I've fixed it!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/6e43e9a35174000aac3d2275a561a0f43f98bb2a. Thank you once more @cosmo0920! :smiley_cat:

cosmo0920:
Thank you for your advice and quick merging! :yum:

PR #45793, opened 2015-11-08
qt5: various improvements and an important fix

UniqMartin:
See #44998 for the discussion.


DomT4:
Merged. Thank you for sticking with this and working with Qt to get various things submitted, which isn't the easiest process in the world. Appreciated :heart:

PR #45796, opened 2015-11-08
codeship-jet v0.11.5 (new formula) 

pigri:
https://codeship.com/documentation/docker/release-notes/

### DIFF
--- a/Library/Formula/codeship-jet.rb
+++ b/Library/Formula/codeship-jet.rb
@@ -1,0 +1,18 @@
+class CodeshipJet < Formula
+  desc "Running Codeship Locally for Development"
+  homepage "https://codeship.com"
+  url "https://s3.amazonaws.com/codeship-jet-releases/0.11.5/jet-darwin_amd64_0.11.5.tar.gz"
+  version "0.11.5"
+  sha256 "b844855284a078b4dce8418b76cdabbbf1428ce67772ab3e32bdfc3a40d0cb4a"
+
+  bottle :unneeded
+
+  def install
+    bin.install "jet"
+  end
+
+  test do
+    system "jet", "version"
+    system "jet", "help"
+  end
+end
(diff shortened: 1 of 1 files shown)


DomT4:
Is this open-source? I can't find any sort of licensing reference on upstream's website.

Library/Formula/codeship-jet.rb
```diff
@@ -0,0 +1,18 @@
+class CodeshipJet < Formula
+  desc "Running Codeship Locally for Development"
+  homepage "https://codeship.com"
+  url "https://s3.amazonaws.com/codeship-jet-homebrew/0.11.5/jet-darwin_amd64.tar.gz"
```
DomT4: Can we not build from source at all? Otherwise this will presumably need to be limited to 64 bit platforms only.

Library/Formula/codeship-jet.rb
```diff
@@ -0,0 +1,18 @@
...
+  sha256 "b844855284a078b4dce8418b76cdabbbf1428ce67772ab3e32bdfc3a40d0cb4a"
+
+  bottle :unneeded
+
+  def install
+    bin.install "jet"
+  end
+
+  test do
+    system "jet", "version"
```
DomT4: Please use `system bin/"jet"` and such, rather than just `system "jet"`.

Library/Formula/codeship-jet.rb
pigri: Just now is only 64 bit platform support. Licensing problem is answer tomorrow. I fixed a bin problem.

DomT4: Alright. If it isn't permissively licensed you could always submit it to the [Caskroom](https://github.com/caskroom/homebrew-cask/).

mikemcquaid: We need to build this from source, sorry.

DomT4:
Any update here at all?

PR #45801, opened 2015-11-08
valgrind: build on Xcode-only systems

tamird:
This partially reverts commit a41a44debd69cadc1f7490968ae1312676376c4.

I'm not sure why this stuff was removed to begin with - valgrind doesn't build on my machine without it.

### DIFF
--- a/Library/Formula/valgrind.rb
+++ b/Library/Formula/valgrind.rb
@@ -39,6 +39,12 @@ class Valgrind < Formula
     end
 
     system "./autogen.sh" if build.head?
+
+    # Look for headers in the SDK on Xcode-only systems: https://bugs.kde.org/show_bug.cgi?id=295084
+    unless MacOS::CLT.installed?
+      inreplace "coregrind/Makefile.in", %r{(?<=\s)(?=/usr/include/mach/)}, MacOS.sdk_path.to_s
+    end
+
     system "./configure", *args
     system "make"
     system "make", "install"


DomT4:
Patch no longer applies cleanly, which may be why it got culled by the other contributor:

```
==> Verifying valgrind--patch-ff8e8186af6985f9ffb8dca8fbea944e807d1caf144f320dca0f995d0417db7f.diff checksum
==> Patching
/usr/bin/patch -g 0 -f -p1 -i /private/tmp/valgrind--patch20151108-29390-1hmudvn/valgrind-sdk-paths-Makefile-in.diff
patching file coregrind/Makefile.in
Hunk #1 FAILED at 105.
Hunk #2 succeeded at 8474 (offset 591 lines).
1 out of 2 hunks FAILED -- saving rejects to file coregrind/Makefile.in.rej
```

tamird:
Ah, I didn't test with non-head. Maybe we need different patches now. I'll look into it shortly.

tamird:
@DomT4 alright, this works a bit better, I hope.

tamird:
@DomT4 thoughts?

DomT4:
I don't know if it's worth wrapping the inreplace in a `unless MacOS::CLT.installed?` block, just to save that operation happening needlessly on with-CLT builds.

DomT4:
Looks fine other than that ponder though - Much cleaner than patching.

tamird:
Hm, okay; happy to make that change.

tamird:
Done.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/6d21fefcecc034a6d3875625b2892a1a5225547f. Thank you @tamird!  :smiley:

tamird:
:confetti_ball:

DomT4:
Going to need you to look at this again I'm afraid :cry:. It broke for Ruby 1.8, which is everyone running lower than Mavericks.

Error is:

```
/usr/local/Library/Formula/valgrind.rb:45: undefined (?...) sequence: /(?<=\s)(?=\/usr\/include\/mach\/)/
```

tamird:
Oh dear. OK, so 1.8.7 doesn't support lookbehind.

PR #45802, opened 2015-11-08
arx 0.2.1 (new formula)

msetek:
The arx tool captures the parameters of an execution and encodes them as
an executable, making for easy, consistent transfer and repetition of a
particular run.

### DIFF
--- a/Library/Formula/arx.rb
+++ b/Library/Formula/arx.rb
@@ -1,0 +1,28 @@
+class Arx < Formula
+  desc "Bundles files and programs for easy transfer and repeatable execution"
+  homepage "https://github.com/solidsnack/arx"
+  url "https://github.com/solidsnack/arx/archive/0.2.1.tar.gz"
+  sha256 "c3e91e4b9f72353061b0a104ee752da88770d695ef6329eca0f1049a56e9c110"
+
+  depends_on "ghc" => :build
+  depends_on "cabal-install" => :build
+
+  def install
+    system "cabal", "sandbox", "init"
+    system "cabal", "update"
+    system "cabal", "install", "--only-dependencies"
+    system "make"
+
+    tag = `./bin/dist tag`.chomp
+    bin.install "tmp/dist/arx-#{tag}/arx" => "arx"
+  end
+
+  test do
+    testscript = (testpath/"testing.sh")
+
+    testscript.write shell_output("#{bin}/arx tmpx // echo 'testing'")
+    testscript.chmod 0555
+
+    assert_match /testing/, shell_output("./testing.sh")
+  end
+end


Library/Formula/arx.rb
```diff
@@ -0,0 +1,24 @@
...
+  def install
+    system "cabal", "sandbox", "init"
+    system "cabal", "update"
+    system "cabal", "install", "--only-dependencies"
+    system "make"
+    bin.install "tmp/dist/arx-0.2.1-amd64-osx/arx" => "arx"
+  end
+
+  test do
+    arx_output = `arx tmpx // echo 'testing'`
```
DomT4: Please use `shell_output` or `pipe_output` here, Thanks.

Library/Formula/arx.rb
```diff
@@ -0,0 +1,24 @@
...
+    File.write("testing.sh", arx_output)
```
DomT4: Should be able to do `(testpath/"testing.sh").write arx_output` here rather than throwing out to `File` directly, I believe.

Library/Formula/arx.rb
```diff
@@ -0,0 +1,24 @@
...
+    res_output = `sh testing.sh`
```
DomT4: You can use something like:

```ruby
assert_match /testing/, shell_output("./testing.sh")
```
Here, which will also check the exit code for you without needing the explicit line below.

Library/Formula/arx.rb
```diff
@@ -0,0 +1,24 @@
+class Arx < Formula
+  desc "bundles files and programs for easy transfer and repeatable execution"
```
DomT4: Tiny nit: `desc`s should start with capital letters.

Library/Formula/arx.rb
```diff
@@ -0,0 +1,26 @@
...
+    system "cabal", "update"
+    system "cabal", "install", "--only-dependencies"
+    system "make"
+    bin.install "tmp/dist/arx-0.2.1-amd64-osx/arx" => "arx"
+  end
+
+  test do
+    testscript = (testpath/"testing.sh")
+
+    testscript.write shell_output("arx tmpx // echo 'testing'")
```
bfontaine: Please use `#{bin}/arx` instead of `arx` here.

Library/Formula/arx.rb
```diff
@@ -0,0 +1,26 @@
...
+    testscript.chmod(0555)
```
bfontaine: You can drop the parentheses.

Library/Formula/arx.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  depends_on "ghc" => :build
+  depends_on "cabal-install" => :build
+
+  def install
+    system "cabal", "sandbox", "init"
+    system "cabal", "update"
+    system "cabal", "install", "--only-dependencies"
+    system "make"
+    bin.install "tmp/dist/arx-0.2.1-amd64-osx/arx" => "arx"
```
DomT4: Will this always be `amd64`? Do GHC/Cabal not support the old 32-bit platforms? (Can't remember off the top of my head).

msetek:
I don't know if GHC/Cabal works or doesn't on old 32-bit platforms. I'm now using the same method that the Makefile uses to determine the platform-specific path to the binary, so it should work in any case.

I tried to use pipe_output, but that doesn't seem to work inside the install method, hence the backticks.
I'm open to suggestions on better ways to capture the output here.

DomT4:
> I'm now using the same method that the Makefile uses to determine the platform-specific path to the binary, so it should work in any case.

Do you know what the possible arches it can throw out on OS X are? We could do something like:

```ruby
arch = MacOS.prefer_64_bit? ? "amd64" : "386"
bin.install "tmp/dist/arx-#{arch}/arx" => "arx"
```
Presuming it uses `386` rather than `x86`.

>  tried to use pipe_output, but that doesn't seem to work inside the install method, hence the backticks.

Yeah, I believe it's test exclusive. Both methods check exit code, which is useful in testing, but you don't necessarily want things to die during install just because the command throws a bad exit code for some reason.

msetek:
I've tried to find out what names ghc uses for the various architectures it supports. I _think_ that it uses the names that gnu autotools uses (i386 and x86_64). But the script that the Makefile uses (./bin/dist)  actually translates "x86_64" into "amd64" (but doesn't do anything with other architecture names). Also, the pathname that the Makefile uses also includes the version number and a suffix that it creates based on the output of "uname -a" ("-ubuntu" or "-osx").

So I think that what you're proposing would work, but I feel it's safer to have the script from arx just tell us the path to the binary rather than assuming that "i386" is correct.

mikemcquaid:
> So I think that what you're proposing would work, but I feel it's safer to have the script from arx just tell us the path to the binary rather than assuming that "i386" is correct.

Makes sense to me :+1:

mikemcquaid:
@BrewTestBot test this please

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/1e267843ba1192867af269eac755b89e05137614. Thank you for your contribution to Homebrew @msetek; we appreciate it! :smiley_cat:

PR #45805, opened 2015-11-08
cdrtools 3.02a01 (devel)

fbrosson:
Add devel, simplify url (drop "projects" and "alpha") and switch to http since SourceForge redirects https to http.
Add test for the extraction feature of isoinfo.

### DIFF
--- a/Library/Formula/cdrtools.rb
+++ b/Library/Formula/cdrtools.rb
@@ -1,10 +1,18 @@
 class Cdrtools < Formula
   desc "CD/DVD/Blu-ray premastering and recording software"
   homepage "http://cdrecord.org/"
+  revision 1
 
   stable do
     url "https://downloads.sourceforge.net/project/cdrtools/cdrtools-3.01.tar.bz2"
+    mirror "https://www.mirrorservice.org/sites/downloads.sourceforge.net/c/cd/cdrtools/cdrtools-3.01.tar.bz2"
+    mirror "https://fossies.org/linux/misc/cdrtools-3.01.tar.bz2"
     sha256 "ed282eb6276c4154ce6a0b5dee0bdb81940d0cbbfc7d03f769c4735ef5f5860f"
+
+    patch do
+      url "https://downloads.sourceforge.net/project/cdrtools/cdrtools-3.01-fix-20151126-mkisofs-isoinfo.patch"
+      sha256 "4e07a2be599c0b910ab3401744cec417dbdabf30ea867ee59030a7ad1906498b"
+    end
   end
 
   bottle do
@@ -14,12 +22,28 @@ class Cdrtools < Formula
     sha256 "8f4e80ac5e512312227029504b52d39085c80b5f7d1f06d0911aaf0eaca1b6a5" => :mountain_lion
   end
 
+  devel do
+    url "https://downloads.sourceforge.net/project/cdrtools/alpha/cdrtools-3.02a02.tar.bz2"
+    mirror "https://fossies.org/linux/misc/cdrtools-3.02a02.tar.bz2"
+    sha256 "b5c33d6cfbe265806f24f365bdb885dfe35194ef716f4b6f809b4377ec159c05"
+
+    patch do
+      url "https://downloads.sourceforge.net/project/cdrtools/alpha/cdrtools-3.02a02-fix-20151126-mkisofs.patch"
+      sha256 "ae7eb217a4f4b1dd8399899282306fc75aaa3b62b269e7f189448657e3944ac6"
+    end
+  end
+
   depends_on "smake" => :build
 
   conflicts_with "dvdrtools",
     :because => "both dvdrtools and cdrtools install binaries by the same name"
 
   def install
+    # Speed-up the build by skipping the compilation of the profiled libraries.
+    # This could be done by dropping each occurence of *_p.mk from the definition
+    # of MK_FILES in every lib*/Makefile. But it is much easier to just remove all
+    # lib*/*_p.mk files. The latter method produces warnings but works fine.
+    rm_f Dir["lib*/*_p.mk"]
     system "smake", "INS_BASE=#{prefix}", "INS_RBASE=#{prefix}", "install"
     # cdrtools tries to install some generic smake headers, libraries and
     # manpages, which conflict with the copies installed by smake itself
@@ -27,7 +51,6 @@ class Cdrtools < Formula
     %w[libschily.a libdeflt.a libfind.a].each do |file|
       (lib/file).unlink
     end
-    (lib/"profiled").rmtree
     man5.rmtree
   end
 
@@ -35,8 +58,13 @@ class Cdrtools < Formula
     system "#{bin}/cdrecord", "-version"
     system "#{bin}/cdda2wav", "-version"
     date = shell_output("date")
-    (testpath/"testfile.txt").write(date)
-    system "#{bin}/mkisofs", "-r", "-o", "test.iso", "testfile.txt"
+    mkdir "subdir" do
+      (testpath/"subdir/testfile.txt").write(date)
+      system "#{bin}/mkisofs", "-r", "-o", "../test.iso", "."
+    end
     assert (testpath/"test.iso").exist?
+    system "#{bin}/isoinfo", "-R", "-i", "test.iso", "-X"
+    assert (testpath/"testfile.txt").exist?
+    assert_equal date, File.read("testfile.txt")
   end
 end


Library/Formula/cdrtools.rb
```diff
@@ -3,7 +3,7 @@ class Cdrtools < Formula
   homepage "http://cdrecord.org/"
 
   stable do
-    url "https://downloads.sourceforge.net/project/cdrtools/cdrtools-3.01.tar.bz2"
+    url "http://downloads.sourceforge.net/cdrtools/cdrtools-3.01.tar.bz2"
```
DomT4: Please stick to the style of URL already used in `stable`, Thanks.

DomT4:
Unusual permission error:

```
==> /usr/local/Cellar/cdrtools/3.01/bin/mkisofs -r -o test.iso testfile.txt
Total translation table size: 0
Total rockridge attributes bytes: 277
Total directory bytes: 0
Path table size(bytes): 10
Max brk space used 0
176 extents written (0 MB)
==> /usr/local/Cellar/cdrtools/3.01/bin/isoinfo -R -i test.iso -X
isoinfo: Permission denied. Cannot create 'testfile.txt'.
Error: cdrtools: failed
```

fbrosson:
I think isoinfo is not being called from within the testpath directory.
If someone knows how to fix this, I'll be happy to add the isoinfo call back.
Gonna squash my commits.
Thanks

Library/Formula/cdrtools.rb
fbrosson: I've reverted the https -> http change. Should I also evert the removal of "/project/" and "/alpha/" ?
"curl --head" shows that SourceForge responds with nice redirects which allow to just provide the name of the tarball without having to indicate the exact path (e.g. /alpha/") as long as we don't specify "/projects/". Are you OK for the simplified URLs?

apjanke:
I think your test may have been correctly written and is actually exposing a problem with the `isoinfo` program. It's throwing a segmentation fault when reading that `test.iso` file, and it seems to be doing a `chmod u-w .` as a side effect when the `-X` option is used. When I look at the directory left over from the failed test, it is non-writable. And I can reproduce it with a manually created directory on the command line.

```
[~/tmp]
$ mkdir cdrtools-test
[~/tmp]
$ cd cdrtools-test
[~/tmp/cdrtools-test]
$ isoinfo -R -i /private/tmp/cdrtools20151108-22029-1pvw354/test.iso
[1]    22816 segmentation fault  isoinfo -R -i /private/tmp/cdrtools20151108-22029-1pvw354/test.iso
[✘ ~/tmp/cdrtools-test]
$ ls -lad .
drwxr-xr-x 2 janke staff 68 Nov  8 18:28 .
[~/tmp/cdrtools-test]
$ isoinfo -R -i /private/tmp/cdrtools20151108-22029-1pvw354/test.iso -X
isoinfo: Permission denied. Cannot create 'testfile.txt'.
[1]    22827 segmentation fault  isoinfo -R -i /private/tmp/cdrtools20151108-22029-1pvw354/test.iso -X
[✘ ~/tmp/cdrtools-test]
$ ls -lad .
dr-xr-xr-x 2 janke staff 68 Nov  8 18:21 .
```

fbrosson:
Thanks @apjanke for your message. I haven't tried isoinfo on OS X yet, but on a Linux box it does not have the side effet of applting "chmod u-w" to the current directory but only to the extracted files. It does not even do a "chmod -w" on directories that exist in the iso image.
But maybe isoinfo behaves differenty on Linux than it does on OS X.

I'm case someone would like to troubleshoot the test bloc, here's what it looked like before I reverted (and squashed) my recent enhancement attempts:
```
  test do
    system "#{bin}/cdrecord", "-version"
    system "#{bin}/cdda2wav", "-version"
    date = shell_output("date")
    (testpath/"testfile.txt").write(date)
    system "#{bin}/mkisofs", "-r", "-o", "test.iso", "testfile.txt"
    assert (testpath/"test.iso").exist?
    rm (testpath/"testfile.txt")
    system "#{bin}/isoinfo", "-R", "-i", "test.iso", "-X"
    assert (testpath/"testfile.txt").exist?
    open("testfile.txt") do |f|
      testfile_content = f.read(100)
      assert_equal testfile_content, date
    end
  end
```
BTW, couldn't the write permission issue be triggered by one of the two commands beteen the calls to mkisofs and isoinfo, that is, the first assert and the rm? Because at the time mkisofs is called, the current directory is writtable, whereas when (or after?) isoinfo is called, it is not writtable anymore.
In other words, is it be possible that a "chmod -w" is being launched as a side effect of the first a:ssert or the rm command?

@DomT4, shall I try a slightly modified test bloc? I'd like to remove the first assert and the rm command, and replace them with:
```
    system "mv", "testfile.txt", "testfile.txt.orig"
```
BTW, Is there a better way to call mv?

Thank you all for your help.

apjanke:
Here's about what I'd do. And you can stick a `ls -lad .` right before the `isoinfo` call to pinpoint where the permission error is happening with `brew test -v cdrtools`.

```
  test do
    system "#{bin}/cdrecord", "-version"
    system "#{bin}/cdda2wav", "-version"
    date = shell_output("date")
    (testpath/"testfile.txt").write(date)
    system "#{bin}/mkisofs", "-r", "-o", "test.iso", "testfile.txt"
    assert (testpath/"test.iso").exist?
    mv "testfile.txt", "testfile.txt.orig"
    system "ls -lad ."
    system "#{bin}/isoinfo", "-R", "-i", "test.iso", "-X"
    assert (testpath/"testfile.txt").exist?
    assert_equal date, File.read("testfile.txt")
  end
```

It's writable right up until the `isoinfo` call.

```
==> ls -lad .
drwx------  5 janke  wheel  170 Nov  8 21:40 .
==> /usr/local/Cellar/cdrtools/3.01/bin/isoinfo -R -i test.iso -X
isoinfo: Permission denied. Cannot create 'testfile.txt'.
```

apjanke:
I poked around the `isoinfo` source a bit. Seems like some pointer or offset issue is causing its internal data structures to get corrupted. There's one value (`todo_idr`) which is initialized to NULL and as far as I can tell never assigned to during the run, but is non-NULL when it comes time to do the output. Maybe something about the compiler or libraries used on Linux avoids this issue.

Should be reported upstream, I think.

fbrosson:
Thanks @apjanke for your hints and code snippets.
Thanks @DomT4 for the "upstream issue" label.
The patch I added fixes the isoinfo -X issue on OS X 10.5 Leopard (without Homebrew), but I can't see why the call to isoinfo is failing with the Homebrew build.
I'll try to dig this. Any extra help is welcome!
BTW, I already reported the issue to the upstrean author.

fbrosson:
Hi @DomT4 and @apjanke.
I found what was going on. (See the message just above and the commit message.)
As cdrtools 3.02a02 —released yesterday— has a workaround for the bug in the clang optimizer (which was failling to produce a correct runtime for isoinfo) and fixes the write-permission bug in isoinfo, I have updated this PR to use 3.02a02 instead of 3.02a01.
The patch for isoinfo in stable is a quick hack. Should I replace it by a resource with steps to replace the isoinfo.c in stable by the one in 3.02a02? This would allow to keep the optimizations for isoinfo in both stable and devel, instead of only in devel.
BTW, @DomT4, now that all issues are fixed, could you remove the "upstream issue" flag?
Regarding the title of this PR, is it possible to replace 3.02a01 by 3.02a02? Or is there a better thing to do?
Thanks!

Library/Formula/cdrtools.rb
```diff
@@ -14,12 +17,29 @@ class Cdrtools < Formula
...
 
   def install
+    # Speed-up the build by changing the definition of MK_FILES in each lib*/Makefile
+    # from "MK_FILES= NAME.mk NAME_p.mk $(NO_SHL)shlNAME.mk"
+    # to   "MK_FILES= NAME.mk           $(NO_SHL)shlNAME.mk"
+    # NAME can be any of libcdrdeflt libdeflt libedc libfile libfind libhfs librscg
+    #                    libscg libscgcmd libschily libsiconv mdigest paranoia
+    # This way the build will skip the *_p.mk makefiles and, consequently, also skip
+    # the creation of the profiled libs.
+    chmod "u+w", Dir["lib*/Makefile"], :verbose => true
```
DomT4: Why the wide `chmod` here?

fbrosson: Oops, I added the `chmod u+w` because I was not sure whether `inreplace` would accept to modify read-only files... and later forgot to test the formula without it. Do you know if `inreplace` works on read-only files? (If it does I"ll remove the `chmod`, otherwise I'll just remove the `verbose` option.) Remark: 12 out of the 13 files to be modified by `inreplace` have `444` permissions.

fbrosson:
I also improved the regexp for the `inreplace` and added an extra one to cope with a different `MK_FILES` pattern in `libedc/Makefile`: it has 2 *_p.mk files which were not all being discarded. This is now fixed, and the bottles do not include any profiled libs. I think the formula is working as expected now. BTW, I tried to make the `inreplace` parameters as generic as possible, so that it won't be broken in case white-spaces diffs appear or new libs are added with the default ` MK_FILES` layout.

Library/Formula/cdrtools.rb
```diff
@@ -4,7 +4,10 @@ class Cdrtools < Formula
 
   stable do
     url "https://downloads.sourceforge.net/project/cdrtools/cdrtools-3.01.tar.bz2"
+    mirror "https://www.mirrorservice.org/sites/downloads.sourceforge.net/c/cd/cdrtools/cdrtools-3.01.tar.bz2"
+    mirror "https://fossies.org/linux/misc/cdrtools-3.01.tar.bz2"
     sha256 "ed282eb6276c4154ce6a0b5dee0bdb81940d0cbbfc7d03f769c4735ef5f5860f"
+    patch :p0, :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Library/Formula/cdrtools.rb
```diff
@@ -14,12 +17,23 @@ class Cdrtools < Formula
...
 
   conflicts_with "dvdrtools",
     :because => "both dvdrtools and cdrtools install binaries by the same name"
 
   def install
+    # Speed-up the build by skipping the compilation of the profiled libraries.
+    # This could be done by dropping each occurence of *_p.mk from the definition
+    # of MK_FILES in every lib*/Makefile. But it is much easier to just remove all
+    # lib*/*_p.mk files. The latter method produces warnings but works fine.
+    rm_f Dir["lib*/*_p.mk"]
```
mikemcquaid: :+1:

Library/Formula/cdrtools.rb
```diff
@@ -27,7 +41,8 @@ def install
     %w[libschily.a libdeflt.a libfind.a].each do |file|
       (lib/file).unlink
     end
-    (lib/"profiled").rmtree
+    # We do not need to call (lib/"profiled").rmtree anymore thanks to
+    # the removal of all lib*/*_p.mk before compilation.
```
mikemcquaid: You can remove this comment.

mikemcquaid: Instead put it in a commit message.

Library/Formula/cdrtools.rb
fbrosson: The patch for `isoinfo.c` fixes the bug in 3.01, but the one for `isoinfo.mk` is just a workaround for a bug in **`clang`**. This means we would still need to embed a patch in the formula. Now, since the loop that was triggering the bug in `clang` has been rewritten in 3.02a02 in such a way that `clang` does not produce wrong code anymore, it would be easy to get rid of the patch by simply adding a resource (to the 3.02a02 tarball) and steps (in a `if build.stable?` bloc) to overwrite `isoinfo.c` by a copy of the one in 3.02a02. I already know how to do this. May I change the formula for that? Or would you prefer to have an external patch for isoinfo.c and an embedded one in the formula?

mikemcquaid: Sorry, I don't understand what `3.02a02` is, is it a stable version of this software? We just need to have patches submitted upstream where there are bugs upstream or a comment explaining in detail why that isn't necessary. Thanks!

fbrosson: `3.02a02` is a preview of the next stable release, `3.02`. The bug in `isoinfo.c` is a write-permission bug and only affects the extraction feature of isoinfo when it is run by non-root users. The other bug, which is due to `clang`, is critical because it makes `isoinfo`  segfault. So we really need both halfs of the patch: the first half to fix the write-permission bug; and the second half to workaround the bug in `clang`.

fbrosson:
Hi @mikemcquaid and thanks for your feedback! I think I need to move `revision` 9 lines up, just below `homepage`. I will also replace the patch by an external patch (as you requested yesterday), because the author told me there was another bug (related to directory relocation) which causes `isoinfo` (in stable) to ignore files and directories when paths are only present in the image in a notation that is not understood by the `isoinfo` shipped in release `3.01`.

Library/Formula/cdrtools.rb
```diff
@@ -4,8 +4,12 @@ class Cdrtools < Formula
 
   stable do
     url "https://downloads.sourceforge.net/project/cdrtools/cdrtools-3.01.tar.bz2"
+    mirror "https://www.mirrorservice.org/sites/downloads.sourceforge.net/c/cd/cdrtools/cdrtools-3.01.tar.bz2"
+    mirror "https://fossies.org/linux/misc/cdrtools-3.01.tar.bz2"
     sha256 "ed282eb6276c4154ce6a0b5dee0bdb81940d0cbbfc7d03f769c4735ef5f5860f"
+    patch :DATA
```
mikemcquaid: Again: can you please submit this patch to the upstream developers of this project or, if it's already upstream, link to the upstream commits/bugs and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

mikemcquaid:
@fbrosson Ok, thanks!

Library/Formula/cdrtools.rb
DomT4: Also please leave a space between the checksum and any patch lines, Thanks.

Library/Formula/cdrtools.rb
```diff
@@ -4,8 +4,12 @@ class Cdrtools < Formula
...
   end
+  revision 1
```
DomT4: This should be at the top of the formula, underneath the `homepage` line in this case.

fbrosson:
A new bug was found yesterday in mkisofs and fixed by the author in schilytools 2015-11-26. I wrote a one-line description of each bug in the commit message. All  known bugs in stable and devel are addressed by the (external) patches. I think the formula is OK now. Thanks @DomT4 and @mikemcquaid for your kind guidance.

mikemcquaid:

apjanke:
:+1:

PR #45808, opened 2015-11-08
lua: tweak include directory and install name

UniqMartin:
This addresses two (minor) issues with the `lua` formula.

The more important one is that `liblua.5.2.dylib` has an install name of `HOMEBREW_PREFIX/lib/liblua.5.2.dylib`, but we nowadays prefer `HOMEBREW_PREFIX/opt/lua/lib/liblua.5.2.dylib`. This way, dependent formulae continue to work if `lua` is unlinked.

### DIFF
--- a/Library/Formula/lua.rb
+++ b/Library/Formula/lua.rb
@@ -58,6 +58,9 @@ class Lua < Formula
   def install
     ENV.universal_binary if build.universal?
 
+    # Subtitute formula prefix in `src/Makefile` for install name (dylib ID).
+    inreplace "src/Makefile", "@LUA_PREFIX@", prefix
+
     # Use our CC/CFLAGS to compile.
     inreplace "src/Makefile" do |s|
       s.remove_make_var! "CC"
@@ -114,7 +117,7 @@ class Lua < Formula
   def pc_file; <<-EOS.undent
     V= 5.2
     R= 5.2.4
-    prefix=#{HOMEBREW_PREFIX}
+    prefix=#{prefix}
     INSTALL_BIN= ${prefix}/bin
     INSTALL_INC= ${prefix}/include
     INSTALL_LIB= ${prefix}/lib
@@ -197,7 +200,7 @@ index 8c9ee67..7f92407 100644
  $(LUA_A): $(BASE_O)
 -	$(AR) $@ $(BASE_O)
 -	$(RANLIB) $@
-+	$(CC) -dynamiclib -install_name HOMEBREW_PREFIX/lib/liblua.5.2.dylib \
++	$(CC) -dynamiclib -install_name @LUA_PREFIX@/lib/liblua.5.2.dylib \
 +		-compatibility_version 5.2 -current_version 5.2.4 \
 +		-o liblua.5.2.4.dylib $^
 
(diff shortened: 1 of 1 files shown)


Library/Formula/lua.rb
```diff
@@ -114,7 +117,7 @@ def install
   def pc_file; <<-EOS.undent
     V= 5.2
     R= 5.2.4
-    prefix=#{HOMEBREW_PREFIX}
+    prefix=#{prefix}
```
mikemcquaid: Maybe `opt_prefix` is preferable?

Library/Formula/lua.rb
```diff
@@ -58,6 +58,9 @@ def pour_bottle?
   def install
     ENV.universal_binary if build.universal?
 
+    # Subtitute formula prefix in `src/Makefile` for install name (dylib ID).
+    inreplace "src/Makefile", "@LUA_PREFIX@", prefix
```
mikemcquaid: Not sure if `opt_prefix` is better here but worth thinking about?

mikemcquaid:
A few nits but otherwise I'm :+1: when @DomT4 is.

Library/Formula/lua.rb
DomT4: Probably. I was quite hard against the idea of rewriting `pkg-config` files to point to `opt_xyz` instead of `xyz` but coming around to the merit of it.

DomT4:
Seems fine to me. I get the feeling we deliberately used `HOMEBREW_PREFIX` for some reason in the past, but I can't recall what it was, so it may well have been just the patch issue. If it passes the CI I'm happy.

Library/Formula/lua.rb
UniqMartin: My motivation for using `prefix` instead of `opt_prefix` was consistency with basically every other formula that exists today, as the value in the `.pc` file usually comes from `./configure --prefix=`. (I remembered the discussion about not wanting to mass-rewrite `.pc` files to `opt`.)

Library/Formula/lua.rb
UniqMartin: Almost the same reasoning like above. All other formulae also produce dylibs with a dylib ID that points into `prefix`. The rewrite to `opt_prefix` happens later in `Keg#fix_install_names` (the `opt`-based dylib ID is constructed [here](https://github.com/Homebrew/homebrew/blob/6e5d803878b2ea35b2243ce4fa1192e7d3b5e8a8/Library/Homebrew/keg_relocate.rb#L151-L157)).

UniqMartin:
I don't really care whether it will be `opt_prefix` or `prefix`, as long as it moves away from `HOMEBREW_PREFIX`. :wink: The code comments explain why a favored `prefix`.

Library/Formula/lua.rb
DomT4: Yeah. I don't really have a problem with `prefix` but I am wary we're one of the few package managers off the top of my head to set `pkg-config` paths pointing at something that isn't the top-level directory, thanks to our unusual symlink from the Cellar structure.

I think there's probably a discussion to ponder a bit more there, but I don't mind `prefix` staying as `prefix` until we make a wider decision.

UniqMartin: I guess I'm glad to have rekindled this discussion?!? :smiling_imp:

mikemcquaid: Yeh, `prefix` sounds good for now but it's maybe worth considering this in future, yeh.

mikemcquaid:
Thanks @UniqMartin!

PR #45816, opened 2015-11-08
travis/test-bot: add Ruby 1.8 test.

mikemcquaid:
This should make it easier to verify we don't accidentally merge pull requests that aren't valid Ruby 1.8 (breaking OS X 10.8 and below).

CC @UniqMartin @xu-cheng @DomT4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
       os: osx
       osx_image: xcode7
       rvm: system
-    - env: OSX=10.9
+    - env: OSX=10.9 HOMEBREW_RUBY=1.8.7
       os: osx
       osx_image: beta-xcode6.2
       rvm: system
@@ -21,11 +21,19 @@ before_install:
   - sudo rm -rf /usr/local/.git/refs /usr/local/.git/packed-refs
   - sudo rsync -az "$TRAVIS_BUILD_DIR/" /usr/local/
   - export TRAVIS_BUILD_DIR="/usr/local"
-  - sudo chown -R $USER /usr/local
-  - env | grep TRAVIS_
-  - if [ -f ".git/shallow" ]; then travis_retry git fetch --unshallow; fi
+  - if [ -f ".git/shallow" ]; then
+      travis_retry git fetch --unshallow;
+    fi
   - git reset --hard $TRAVIS_COMMIT
   - git clean -qxdff
+
+install:
+  - if [ "$HOMEBREW_RUBY" = "1.8.7" ]; then
+      brew install homebrew/versions/ruby187;
+      export PATH="/usr/local/opt/ruby187/bin:$PATH";
+      export HOMEBREW_RUBY_PATH="/usr/local/opt/ruby187/bin/ruby";
+    fi
+  - export HOMEBREW_DEVELOPER="1"
 
 script:
   - brew test-bot
--- a/bin/brew
+++ b/bin/brew
@@ -1,6 +1,5 @@
 #!/bin/sh
-
-chdir () {
+chdir() {
   cd "$@" >/dev/null
 }
 
@@ -10,9 +9,9 @@ export HOMEBREW_BREW_FILE="$BREW_FILE_DIRECTORY/${0##*/}"
 BREW_SYMLINK=$(readlink "$0")
 if [ -n "$BREW_SYMLINK" ]
 then
-    BREW_SYMLINK_DIRECTORY=$(dirname "$BREW_SYMLINK")
-    BREW_FILE_DIRECTORY=$(chdir "$BREW_FILE_DIRECTORY" &&
-                          chdir "$BREW_SYMLINK_DIRECTORY" && pwd -P)
+  BREW_SYMLINK_DIRECTORY=$(dirname "$BREW_SYMLINK")
+  BREW_FILE_DIRECTORY=$(chdir "$BREW_FILE_DIRECTORY" &&
+                        chdir "$BREW_SYMLINK_DIRECTORY" && pwd -P)
 fi
 
 BREW_LIBRARY_DIRECTORY=$(chdir "$BREW_FILE_DIRECTORY"/../Library && pwd -P)
@@ -22,10 +21,18 @@ BREW_LIBRARY_DIRECTORY=$(chdir "$BREW_FILE_DIRECTORY"/../Library && pwd -P)
 unset GEM_HOME
 unset GEM_PATH
 
-BREW_SYSTEM=$(uname -s | tr "[:upper:]" "[:lower:]")
-if [ "$BREW_SYSTEM" = "darwin" ]
+if [ -z "$HOMEBREW_DEVELOPER" ]
 then
-    exec "$BREW_LIBRARY_DIRECTORY/brew.rb" "$@"
-else
-    exec ruby -W0 "$BREW_LIBRARY_DIRECTORY/brew.rb" "$@"
+  unset HOMEBREW_RUBY_PATH
 fi
+
+if [ -z "$HOMEBREW_RUBY_PATH" ]; then
+  BREW_SYSTEM=$(uname -s | tr "[:upper:]" "[:lower:]")
+  if [ "$BREW_SYSTEM" = "darwin" ]; then
+    HOMEBREW_RUBY_PATH="/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby"
+  else
+    HOMEBREW_RUBY_PATH="ruby"
+  fi
+fi
+
+exec "$HOMEBREW_RUBY_PATH" -W0 "$BREW_LIBRARY_DIRECTORY/brew.rb" "$@"
--- a/Library/Homebrew/cmd/test-bot.rb
+++ b/Library/Homebrew/cmd/test-bot.rb
@@ -632,7 +632,7 @@ module Homebrew
       else
         test "brew", "tests", "--no-compat"
         readall_args = ["--aliases"]
-        readall_args << "--syntax" if MacOS.version >= :mavericks
+        readall_args << "--syntax" if RUBY_VERSION.split(".").first.to_i >= 2
         test "brew", "readall", *readall_args
         test "brew", "update-test"
       end


xu-cheng:
In fact, I hope we can move to Ruby 2.0 instead of supporting Ruby 1.8.

Library/Homebrew/os.rb
```diff
@@ -1,6 +1,6 @@
 module OS
   def self.mac?
-    /darwin/i === RUBY_PLATFORM
+    /darwin/i === RUBY_PLATFORM || ENV["HOMEBREW_EMULATE_MAC"]
```
xu-cheng: We don't really need this. It should works fine without it. 

In fact, as a design guidance, we should allow homebrew to run under Linux for metadata processing(e.g. `brew info --json=v1`) without any environment.

.travis.yml
```diff
@@ -14,18 +14,37 @@ matrix:
...
+      cd /usr/local;
+      sudo rm -rf /usr/local/.git/refs /usr/local/.git/packed-refs;
+      sudo rsync -az "$TRAVIS_BUILD_DIR/" /usr/local/;
+      export TRAVIS_BUILD_DIR="/usr/local";
+      sudo chown -R $USER /usr/local;
+    elif [ -n "$RUBY" ]; then
+      export HOMEBREW_OLD_RUBY_TEST="1";
+    elif [ -n "$LINUX" ]; then
+      export HOMEBREW_OLD_RUBY_TEST="1";
+      export HOMEBREW_EMULATE_MAC="1";
```
xu-cheng: If possible, I'd prefer to set env using travis `env` DSL. The main reason is if we keep each command as a entry, Travis will fold them nicely one by one(instead of one big block).

UniqMartin:
> In fact, I hope we can move to Ruby 2.0 instead of supporting Ruby 1.8.

I'd love if this was the case, I'm just not sure how realistic this is. It makes sense as 10.8 is no longer a supported OS X release, but apparently that doesn't stop people from using it (or much older OS X versions). It's gonna be hard to sell this decision to them, but that's a decision not to be made by me. :wink:

Library/Homebrew/os.rb
UniqMartin: I think what you are saying is a goal Homebrew could and should strive for. But I think this still requires a lot of work, as not everything in a formula is purely declarative. One example I can think of is dependencies being declared (or not) based on the currently running OS X release by using a suffixed `if` clause. There are probably many other things that require thought and redesign …

DomT4:
> In fact, I hope we can move to Ruby 2.0 instead of supporting Ruby 1.8.

Would love to do this, but adding an audit check for invalid <2.0 syntax is a lot quicker and easier a fix short-term.

Even in Homebrew we'd need to work out the best delivery system for a modern Ruby, persuade users to actually adopt using it and get our `#!` calls to use it, and then Misty is going to be forced to act quicker than she may wish to to do a similar thing in Tigerbrew or let potentially breaking syntax errors pile up. It's something quite tricky to rush, compared to borrowing Travis for a further check.

xu-cheng:
Currently, https://github.com/phusion/traveling-ruby offers one of most promised Ruby interrupter. It's fully static linked, so can be used in any location without any dependency requirement. And it's quite compact. You can download from [here](http://traveling-ruby.s3-us-west-2.amazonaws.com/list.html) to test it.  Therefore, it would be helpful if someone with old version VM to test it locally.  And probably adopted the similar techniques to PowerPC for 10.4. Luckily, this project offers full code for compiling such customed interpreter.  

If we can fix interpreter part, we should then work on new version of update system. And how to migrate such change in the process. 

As for `#!`, we don't really that. Ruby interpreter selection should all be done by bash inside `bin/brew`.

Library/Homebrew/os.rb
xu-cheng: Yes, it is still not perfect. But it's currently workable for most of meta data. I know some users invoking brew info command inside docker to get bottle URL list for unofficial mirror.

mikemcquaid:
I'm all for us installing a newer Ruby but I think this is essential in the mean time to ensure we don't break stuff on old Rubies without realising it.

Library/Homebrew/os.rb
mikemcquaid: Try running current `master` on Linux and see how broken it is. This PR is still work-in-progress while I work out whether emulating Mac or actually using a Mac Travis worker is going to be easier.

UniqMartin:
I'm not sure how practical the following is with Travis CI, but I simply installed Ruby 1.8.7 on OS X 10.11 and made `brew` use that. To me this seems to be a much simpler solution that getting everything to work on Linux. Here's what I did in more detail:

First, I applied the following patch, which allows to override the `ruby` used by Homebrew:

```patch
diff --git a/bin/brew b/bin/brew
index 20a13131..7b2b13d7 100755
--- a/bin/brew
+++ b/bin/brew
@@ -23,9 +23,10 @@ unset GEM_HOME
 unset GEM_PATH

 BREW_SYSTEM=$(uname -s | tr "[:upper:]" "[:lower:]")
-if [ "$BREW_SYSTEM" = "darwin" ]
+if [ "$BREW_SYSTEM" = "darwin" ] && [ -z "$HOMEBREW_RUBY" ]
 then
     exec "$BREW_LIBRARY_DIRECTORY/brew.rb" "$@"
 else
-    exec ruby -W0 "$BREW_LIBRARY_DIRECTORY/brew.rb" "$@"
+    [ -z "$HOMEBREW_RUBY" ] && HOMEBREW_RUBY=ruby
+    exec "$HOMEBREW_RUBY" -W0 "$BREW_LIBRARY_DIRECTORY/brew.rb" "$@"
 fi
```

I then installed `ruby187` from the `homebrew/versions` tap, put it in `PATH` and set `HOMEBREW_RUBY`:

```shell
brew install homebrew/versions/ruby187
export PATH="$(brew --prefix)/opt/ruby187/bin:$PATH"
export HOMEBREW_RUBY=ruby
```

To be able to run `brew tests` I also installed RubyGems from <https://rubygems.org/pages/download>.

Afterwards, the following stuff usually done by the CI worked and used Ruby 1.8.7:

```
brew --env
brew config
brew tests
brew tests --no-compat
brew readall --aliases --syntax
```

I have zero experience with Travis CI, thus cannot try and do this myself, but maybe this is the fastest way to make this work and move on?

*Note: To make this more practical, it might be a good idea to build an El Capitan bottle for the `ruby187` formula and to vendor RubyGems directly into the formula.*

DomT4:
Xu has a [script](https://github.com/xu-cheng/dotfiles/blob/master/home/bin/brew-1.8) which does something similar.

xu-cheng:
> but I think this is essential in the mean time to ensure we don't break stuff on old Rubies without realising it.

I have no objection in this PR in general.

Library/Homebrew/os.rb
xu-cheng: I just realize it's broken now. I will add on how to at least make `brew info --json=v1` workable on Linux to my TODO list.

mikemcquaid:
> I have zero experience with Travis CI, thus cannot try and do this myself, but maybe this is the fastest way to make this work and move on?

This seems like the best approach. I think it'd be worth just adding it to the beginning (or end) of an existing Travis CI job. Good idea, thanks!

mikemcquaid:
So, the current status is that the Homebrew Ruby 1.8.7 doesn't seem to install `rubygems` which means the tests are failing.

UniqMartin:
> I think it'd be worth just adding it to the beginning (or end) of an existing Travis CI job.

Wouldn't it be easier yet if it was a separate job that focused on just that? Or are there reasons (maybe on the Travis CI side) I'm not aware of?

> So, the current status is that the Homebrew Ruby 1.8.7 doesn't seem to install rubygems which means the tests are failing.

Yes. I tried to hint at that in my longish comment. But as I see you already started to work on that in `homebrew/versions`.

.travis.yml
```diff
@@ -21,12 +21,22 @@ before_install:
...
   - git clean -qxdff
 
+install:
+  - if [ "$HOMEBREW_OLD_RUBY_TEST" = "1.8.7" ]; then
+      brew install --build-bottle homebrew/versions/ruby187;
+      export PATH="/usr/local/opt/ruby187/bin:$PATH";
+      tar xf https://rubygems.org/rubygems/rubygems-2.5.0.tgz;
+      cd rubygems-2.5.0;
+      ruby setup.rb;
+      cd /usr/local
```
UniqMartin: There's a `;` missing here, but I guess you know already and instead focus on bringing `homebrew/versions/ruby187` in a more useful shape.

mikemcquaid:
> Wouldn't it be easier yet if it was a separate job that focused on just that? Or are there reasons (maybe on the Travis CI side) I'm not aware of?

If I use a separate OS X worker it'll limit all our other OS X jobs.

DomT4:
> persuade users to actually adopt using it

In regards to this and answering my own question there, we could just bundle it as part of `brew update` in `Library/Homebrew/Vendor` presuming the Ruby license is compliant with BSD.

MacPorts bundles their own interpreter so people can't really angry rant too much, theoretically. It'll make the first `brew update` a bit chunkier, but afterwards we'd only need to update whenever OS X did the same.

xu-cheng:
> So, the current status is that the Homebrew Ruby 1.8.7 doesn't seem to install rubygems which means the tests are failing.

Looks like you need to backport rubygem config from `ruby20` formula to `ruby187` formula.

.travis.yml
```diff
@@ -21,12 +21,22 @@ before_install:
...
-  - if [ -f ".git/shallow" ]; then travis_retry git fetch --unshallow; fi
+  - if [ -f ".git/shallow" ]; then
+      travis_retry git fetch --unshallow;
+    fi
   - git reset --hard $TRAVIS_COMMIT
   - git clean -qxdff
 
+install:
+  - if [ "$HOMEBREW_OLD_RUBY_TEST" = "1.8.7" ]; then
+      brew install --build-bottle homebrew/versions/ruby187;
```
xu-cheng: You probably want to remove `--build-bottle` to use bottle.

bin/brew
```diff
@@ -23,7 +23,7 @@ unset GEM_HOME
 unset GEM_PATH
 
 BREW_SYSTEM=$(uname -s | tr "[:upper:]" "[:lower:]")
-if [ "$BREW_SYSTEM" = "darwin" ]
+if [ "$BREW_SYSTEM" = "darwin" ] && [ -z "$HOMEBREW_OLD_RUBY_TEST" ]
```
xu-cheng: As a more generate approach, how about using `HOMEBREW_RUBY_PATH` like below?
```
if [ -n "$HOMEBREW_RUBY_PATH" ]; then
  exec "$HOMEBREW_RUBY_PATH" -W0 "$BREW_LIBRARY_DIRECTORY/brew.rb" "$@"
elif [ "$BREW_SYSTEM" = "darwin" ]; then
  exec "$BREW_LIBRARY_DIRECTORY/brew.rb" "$@"
else
  exec ruby -W0 "$BREW_LIBRARY_DIRECTORY/brew.rb" "$@"
fi
```

mikemcquaid:
> Looks like you need to backport rubygem config from ruby20 formula to ruby187 formula.

I don't think so; I don't know if we want them sharing the global gem directory. I also couldn't get it to work as expected and I'm not convinced it's worth the effort, honestly.

.travis.yml
mikemcquaid: Pretty sure the bottle is broken as build from source it's installing outside the sandbox

bin/brew
mikemcquaid: I'd rather not have a general approach; I don't want to encourage this behaviour.

xu-cheng:
> I don't know if we want them sharing the global gem directory

The gem directory should have version in its path, so it should be OK. I will take a look.

bin/brew
xu-cheng: I was thinking use it when we move to ruby 2.x only

mikemcquaid: That's cool, something we can do later. Unrelated but I think while we move to Ruby 2 we should also consider deprecating <=10.8 support (and/or moving to Tigerbrew)

mikemcquaid:
Looks like this is :green_heart: (thanks to @UniqMartin) you can see the last job I did before the rebase is using Ruby 1.8 and all tests are passing: https://travis-ci.org/Homebrew/homebrew/jobs/91776346#L218. I think using Ruby 1.8 on our oldest OS X is probably the best strategy for now; it saves us adding another job and slowing things down.

Any objections to merging this @DomT4 @xu-cheng?

bin/brew
```diff
@@ -23,7 +23,7 @@ unset GEM_HOME
 unset GEM_PATH
 
 BREW_SYSTEM=$(uname -s | tr "[:upper:]" "[:lower:]")
-if [ "$BREW_SYSTEM" = "darwin" ]
+if [ "$BREW_SYSTEM" = "darwin" ] && [ "$HOMEBREW_OLD_RUBY_TEST" != "1.8.7" ]
```
xu-cheng: Maybe just whether the env exists rather than its content. After all, its content is never used.

xu-cheng:
One comment. Everything else are all good.

xu-cheng:
Rethink a bit, I think we should introduce `HOMEBREW_RUBY_PATH ` variable for following reasons:
* This allows us to choose brew ruby interpreter without messing with `PATH` variable. Using `PATH` is fragile as it can a) interfere formula building (especially for formula building using Stdenv or `depends_on :ruby`) b) be interfered by other processes. If `PATH` is reset, we wouldn't know what ruby we are testing.
* It allows us to choose ruby for Superenv. Currently(even after this PR) superenv scripts are still being interpreted by system ruby. While we clearly want to check ruby 1.8 for superenv scripts  as well.
* It will benefit moving to ruby 2.0 in the future. All we have to do is reassign `HOMEBREW_RUBY_PATH ` variable in the initial script.

Therefore, I suggest we do something like below:
```bash
if [ -z "$HOMEBREW_RUBY_PATH" ]; then
  if [ "$BREW_SYSTEM" = "darwin" ]; then
    export HOMEBREW_RUBY_PATH=/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby
  else
    export HOMEBREW_RUBY_PATH=ruby
  fi
fi

exec "$HOMEBREW_RUBY_PATH" -W0 "$BREW_LIBRARY_DIRECTORY/brew.rb" "$@"
```

mikemcquaid:
> And for Superenv scripts, say cc, we can rename it to cc.rb and create a new cc script:

This approach would also be useful in e.g. unsetting `RUBYLIB`.

I've implemented your approach, @xu-cheng, and will merge once I've confirmed this is work as expected. Thanks, as usual, for a smart approach!

.travis.yml
```diff
@@ -10,7 +10,7 @@ matrix:
       os: osx
       osx_image: xcode7
       rvm: system
-    - env: OSX=10.9
+    - env: OSX=10.9 RUBY=1.8.7
```
xu-cheng: You probably need to set HOMEBREW_DEVELOPER as well.

mikemcquaid: :+1:

mikemcquaid:
Merged. Thanks for the help @xu-cheng, @UniqMartin, @DomT4!

DomT4:
Thanks for getting this sorted Mike. Will be good to stop stumbling into 1.8 syntax errors.

PR #45818, opened 2015-11-08
gnupg url and mirror updates

vszakats:
* replace ftp with secure protocol
* sync mirrors across gnupg projects
* replace `mirrors.kernel.org` with `mirrorservice.org`
* consistent use of `gnupg.org` (for URLs) vs. `www.gnupg.org` (for homepages)
* remove insecure mirrors
* homepage updates and cleanups
* minor sync of build options formatting

### DIFF
--- a/Library/Formula/libksba.rb
+++ b/Library/Formula/libksba.rb
@@ -1,7 +1,7 @@
 class Libksba < Formula
   desc "X.509 and CMS library"
-  homepage "https://www.gnupg.org/related_software/libksba/index.en.html"
-  url "ftp://ftp.gnupg.org/gcrypt/libksba/libksba-1.3.3.tar.bz2"
+  homepage "https://www.gnupg.org/related_software/libksba/"
+  url "https://gnupg.org/ftp/gcrypt/libksba/libksba-1.3.3.tar.bz2"
   mirror "https://www.mirrorservice.org/sites/ftp.gnupg.org/gcrypt/libksba/libksba-1.3.3.tar.bz2"
   sha256 "0c7f5ffe34d0414f6951d9880a46fcc2985c487f7c36369b9f11ad41131c7786"
 
@@ -16,8 +16,9 @@ class Libksba < Formula
   depends_on "libgpg-error"
 
   def install
-    system "./configure", "--disable-dependency-tracking", "--prefix=#{prefix}",
-                          "--disable-silent-rules"
+    system "./configure", "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}"
     system "make", "install"
   end
 
--- a/Library/Formula/dirmngr.rb
+++ b/Library/Formula/dirmngr.rb
@@ -1,8 +1,8 @@
 class Dirmngr < Formula
   desc "Server for managing certificate revocation lists"
-  homepage "https://www.gnupg.org"
+  homepage "https://www.gnupg.org/"
   url "https://gnupg.org/ftp/gcrypt/dirmngr/dirmngr-1.1.1.tar.bz2"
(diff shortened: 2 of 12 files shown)


mikemcquaid:
Could you take a look at the CI failures? Thanks!

vszakats:
It fails in `dirmngr` which was not updated in this PR, and just passed tests successfully an hour ago in a separate [PR](https://github.com/Homebrew/homebrew/pull/45817).

```
==> dirmngr-client --help
dyld: Library not loaded: /usr/local/lib/libassuan.0.dylib
  Referenced from: /usr/local/bin/dirmngr-client
  Reason: image not found
Error: dirmngr: failed
Failed executing: dirmngr-client --help
```

Library/Formula/libksba.rb
```diff
@@ -1,8 +1,8 @@
 class Libksba < Formula
   desc "X.509 and CMS library"
   homepage "https://www.gnupg.org/related_software/libksba/index.en.html"
-  url "ftp://ftp.gnupg.org/gcrypt/libksba/libksba-1.3.3.tar.bz2"
```
DomT4: Please keep some form of official upstream URL in place here.

vszakats: Tthat was a mistake, thanks for noticing.

vszakats: Fixed! (Is there a way to squash two non-consecutive commits?)

mikemcquaid: `git rebase --interactive` and cut-and-paste the lines to a different order.

DomT4: Yeah. If you `git rebase -i HEAD~12` and then just drag/drop your commits into the order desired and squash as per normal.

DomT4: Oops, Mike beat me to it :smile_cat:

vszakats: Thanks to both of you! Got it the way you described 👍🏻

vszakats: Did some `gnupg.org` (URLs) vs. `www.gnupg.org` (homepages) inconsistency fixup. They may be further synced by using the latter for URLs as well.

vszakats:
Question: Shouldn't be all additional insecure mirrors removed? The canonical and `mirrorservice.org` URLs would still remain. (Having an insecure mirror can be exploited in theory.)

DomT4:
Feel free to kill anything that isn't SSL/TLS, IMO, presuming we have one good `url` and one good `mirror`.

vszakats:
Cool, will do!

vszakats:
Done.

vszakats:
What does the new result `UNSTABLE` mean?
```
==> brew cleanup --prune=7
Recording test results
Build step 'Publish JUnit test result report' changed build result to UNSTABLE
Archiving artifacts
Started calculate disk usage of build
Finished Calculation of disk usage of build in 0 seconds
Started calculate disk usage of workspace
Finished Calculation of disk usage of workspace in 0 seconds
[WS-CLEANUP] Deleting project workspace...[WS-CLEANUP] done
Finished: UNSTABLE
```

DomT4:
It's not new. Just doesn't appear very often. Official definition from Jenkins is:

> A build is unstable if it was built successfully and one or more publishers report it unstable. For example if the JUnit publisher is configured and a test fails then the build will be marked unstable.

DomT4:
In Homebrew terms, you can usually read `UNSTABLE` as *"Something broke that wasn't my fault and can probably be ignored by the maintainers"*, to be honest.

vszakats:
Thanks @DomT4. I'll make some more minor cleanups, but still nothing that should cause any change in the actual build processes.

vszakats:
Okay, finished with that, too.

DomT4:
Merged in a whole bunch of commits. Thank you @vszakats :smiley_cat:

vszakats:
Thank you so much @DomT4!

PR #45820, opened 2015-11-08
freealut: restore trailing space in embedded patch

vszakats:
after 2c4e697863ede6525ee7da7685bf23d64c96fbf6

### DIFF
--- a/Library/Formula/freealut.rb
+++ b/Library/Formula/freealut.rb
@@ -33,6 +33,6 @@ index 2b26d6d..4001db1 100644
 -AC_SEARCH_LIBS([alGetError], [openal32 openal])
 +# Use Mac OS X frameworks
 +LIBS="$LIBS -framework IOKit -framework OpenAL"
-
+ 
  ################################################################################
  # Checks for header files.


mikemcquaid:
@vszakats Does it fail to build without this?

vszakats:
Haven't checked this specific case, but in general trailing spaces are significant in diffs, so it's better to not lose them. Since the update doesn't mention fixing the patch, I assumed they were accidentally deleted (by the text editor).

mikemcquaid:
Would you mind checking? Thanks!

vszakats:
It worked in my test without that space, and it also works with it.

Notice that another trailing space in line 30 was kept there, even though 
the patch still works without that one too.

So, I'd think either both should be kept or both deleted. Preferably the first, 
as that is what the `diff` tool generates.

mikemcquaid:
I'm happy to remove both if it works without both.

vszakats:
Hm, I'd rather close this one, it just doesn't make sense to work more on something that is generally incorrect. (I've had to fix accidental trailing spaces several times in my own patches)

vszakats:
(that is: unless we know what is the exact rule for trailing space requirements in diffs, I think it's not safe to mess with them even if it works in my specific test case. My past experience is definitely not to mess with them ever.)

vszakats:
(I wish diffs would not require trailing spaces ever, though. ATM I need to use separate editor for them, because the text editor used for source code is set to automatically remove them. Rant: off.)

PR #45821, opened 2015-11-08
iat: use secure homepage

vszakats:
(it will be redirected to insecure if not logged-in)


mikemcquaid:
Thanks again @vszakats!

PR #45822, opened 2015-11-08
libwbxml: use secure homepage

vszakats:
(it will be redirected to insecure if not logged-in)


mikemcquaid:
Could you take a look at the CI failures? Thanks!

vszakats:
It fails in CMake. Unrelated to this patch. Since I had never successfully used CMake yet, I'd rather close this one or leave it to an expert.

DomT4:
Closing in favour of #45834, which includes this commit.

PR #45825, opened 2015-11-09
update: always reset to origin/master.

mikemcquaid:
Do a reset to `origin/master` and then stash but don't pop the stash after running update (unless you were on a branch). This may be mildly more annoying for Homebrew developers but means it's easier for our users who don't understand Git (and particularly when they don't understand that every tap is a separate Git repository).

CC @DomT4 @xu-cheng who I talked about this with


xu-cheng:
LGTM, except need to update test accordingly.

Library/Homebrew/cmd/update.rb
```diff
@@ -263,27 +263,43 @@ def pull!(options = {})
     @current_revision = read_current_revision
 
     if @initial_branch != "master" && !@initial_branch.empty?
```
UniqMartin: Now that you're touching this piece of code, shouldn't that be `@initial_branch != @upstream_branch`?

Library/Homebrew/cmd/update.rb
```diff
@@ -241,12 +241,12 @@ def pull!(options = {})
       return
     end
 
+    @initial_revision = read_current_revision
```
UniqMartin: I feel this is quite a big semantic change. Previously, you've got a report for the changes pulled into `master` (or whatever the upstream branch is). By moving this here, you'll get a diff between your current working branch and whatever commit happens to be `master` after the update. Since I'm the one who moved that to where it is now in #42074, I'm not exactly happy to see this.

Library/Homebrew/cmd/update.rb
```diff
@@ -227,6 +218,15 @@ def pull!(options = {})
       @initial_branch = ""
     end
 
+    unless system "git", "diff", "--quiet"
```
UniqMartin: Now that this is touched, I think a safer way to check for a pristine working copy is to use `git status --untracked-files=all --porcelain` (as is also done in `brew doctor`, see [here](https://github.com/Homebrew/homebrew/blob/ad6f9d1bb856499f8166468c9d96192c44b00f9b/Library/Homebrew/cmd/doctor.rb#L1073)). The current approach will fail to detect staged, but not yet committed changes, and untracked files.

UniqMartin:
Can we maybe go for a more defensive and less destructive approach (lots of `--force` in there)? What about checking if the working copy is clean and whether `master` can be fast-forwarded? And if this is not the case tell the user that they either should know what they are doing (the changes were intended) or provide a link to documentation that explains how to return the working copy to a pristine state.

This is just an idea and not a fully fleshed out yet. What do you think?

mikemcquaid:
> Can we maybe go for a more defensive and less destructive approach (lots of --force in there)? What about checking if the working copy is clean and whether master can be fast-forwarded? And if this is not the case tell the user that they either should know what they are doing (the changes were intended) or provide a link to documentation that explains how to return the working copy to a pristine state.

I specifically don't want to require documentation for this. The case where users want to have a non-pristine install means they have already interacted with it intentionally with Git. This should be the non-default case as as-is it causes a bunch of issues for less Git-savvy users.

mikemcquaid:
To be clear though: the `reset` and `stash` mean that all user changes will be stored for retrieval if desired.

Library/Homebrew/cmd/update.rb
mikemcquaid: :+1:

Library/Homebrew/cmd/update.rb
mikemcquaid: I'll investigate.

Library/Homebrew/cmd/update.rb
mikemcquaid: Good idea :+1:

Library/Homebrew/cmd/update.rb
UniqMartin: Thanks! Without this being addressed, I believe running `brew update`, the report it generates, and the code that acts on this report will become mostly useless to me (as I'm almost always on some branch different from `master`).

My (possibly wrong) expectation is that `brew update` should fetch upstream changes and then only report and act on them, disregarding the stuff that might be going on in the personal branch that I'm currently on.

UniqMartin:
Thanks! Now I better understand the purpose of this change and I'm in favor of your solution. I hope you'll be able to address my code comment about the possible semantic change before merging this.

mikemcquaid:
> Thanks! Now I better understand the purpose of this change and I'm in favor of your solution. I hope you'll be able to address my code comment about the possible semantic change before merging this.

Yup, I'll take a look.

mikemcquaid:
@UniqMartin Updated, can you take another look?

Library/Homebrew/cmd/update.rb
```diff
@@ -228,6 +219,16 @@ def pull!(options = {})
       @initial_branch = ""
     end
 
+    unless `git status --untracked-files=all --porcelain 2>/dev/null`.chomp.empty?
+      if ARGV.verbose?
+        puts "Stashing uncommitted changes to #{repository}."
+        system "git", "status", "--short", "--untracked-files=all"
+      end
+      safe_system "git", "stash", "save", "--include-untracked", *@quiet_args
+      safe_system "git", "reset", "--hard"
```
UniqMartin: This command always produces output, which might be confusing to most users (that rarely run `brew update` with the `--verbose` option). The output is something along the lines of:

```
HEAD is now at <sha1> <commit-message>
```

UniqMartin: Just to clarify, it does so for repositories with local modifications. Otherwise this line obviously wouldn't be executed.

Also, isn't this line a no-op and can be omitted as the immediately preceding `git stash save` already returns the working directory to a clean state? Or am I overlooking a case where that's not true?

Library/Homebrew/cmd/update.rb
```diff
@@ -263,18 +264,29 @@ def pull!(options = {})
...
+      system "git", "status", "--short", "--untracked-files"
     end
+    @stashed = false
+  end
+
+  def pop_stash_message
+    return unless @stashed
+    puts "To restore the stashed changes to #{repository} run:"
+    puts "  `cd #{repository} && git stash pop`"
+    @stashed = false
```
UniqMartin: I understand why this is here, but I'm afraid that users won't notice this message (as it will be printed above the formula update report) and then they will complain that Homebrew destroyed their uncommitted local changes.

But let's keep it like this and if someone complains, we can tell them to `git stash pop`, and if too many people complain, we can still address this later.

UniqMartin:
I just tested it locally and failed to break it. It also always behaved like I expected it to be. I have left two comments about minor issues that could be addressed (but don't necessarily have to), but otherwise :+1:. (Of course, the tests need to be adapted.)

Library/Homebrew/cmd/update.rb
mikemcquaid: :+1:

Library/Homebrew/cmd/update.rb
mikemcquaid: It will usually be a no-op but it will help resolve issues with e.g. bad/mid merges.

mikemcquaid: Added a `--quiet`

mikemcquaid:
@UniqMartin Addressed your comments and fixed the tests.

Library/Homebrew/cmd/update.rb
UniqMartin: I think the `--quiet` went in the wrong place. And it might be better to use `*@quiet_args` like in the line above. The command itself will be echoed in verbose mode, so there's no harm in not silencing its output.

Library/Homebrew/cmd/update.rb
```diff
@@ -139,7 +139,7 @@ def git_init_if_necessary
       safe_system "git", "config", "remote.origin.url", "https://github.com/Homebrew/homebrew.git"
       safe_system "git", "config", "remote.origin.fetch", "+refs/heads/*:refs/remotes/origin/*"
       safe_system "git", "fetch", "origin"
-      safe_system "git", "reset", "--hard", "origin/master"
+      safe_system "git", "reset", "--quiet", "--hard", "origin/master"
```
UniqMartin: Here's the misplaced `--quiet`.

mikemcquaid: Bleurg. Fixed, pushed.

PR #45828, opened 2015-11-09
app-engine-python: Add remaining executable symlinks to PATH.

n8henrie:
The Google App Engine GUI symlinks 16 additional executables to the
PATH. The existing formula missed a handful of these, causing
dev_appserver to raise an error when attempting to import them (e.g.
`import wrapper_util.py`). This simply adds the remaining executables to
the list of those to be linked.


bfontaine:
It’s probably worth to bump the revision here. Please add `revision 1` just below the `sha256` at the top of the formula.

n8henrie:
Updated with revision

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/aeb7871b27a7a7d824b602035aa2cfec83c916eb. Thank you for your contribution to Homebrew @n8henrie; we appreciate it! :smiley_cat:

PR #45835, opened 2015-11-09
keg_relocate: remove options from Keg#dylib_id_for

UniqMartin:
No longer used since commit 3b0cbe6a56d9133941482a9e2d033ad86cdb5e79.


xu-cheng:
It would be great to update the method argument list as well.

xu-cheng:
Oh, I just saw you did it. :+1:

xu-cheng:
Thanks again.

PR #45837, opened 2015-11-09
audit: respect umask in formula file mode check

UniqMartin:
My impression is that Homebrew's philosophy is to respect the user's `umask`, i.e. there is no `brew doctor` check for a non-standard `umask`. But having a non-standard `umask` like `0077` will always create a spurious error when running `brew audit --strict` locally. This change fixes this.


mikemcquaid:
Nice. Can you split into more variables (or add comments) to make it clearer what's going on here? I'm stupid but it's not 100% clear to me. Thanks!

UniqMartin:
Updated with code comment and slightly tweaked variable names. Do you think it's clear enough now?

mikemcquaid:
Looks good. Last thing: can you give an example of the spurious error you were seeing before? Thanks!

UniqMartin:
> Last thing: can you give an example of the spurious error you were seeing before?

Sure. Here's the output of `brew audit qt5` before the change (umask is `0077`, `qt5.rb` has mode `600`):

```
qt5:
 * Incorrect file permissions: chmod 644 /opt/homebrew/Library/Formula/qt5.rb

Error: 1 problem in 1 formula
```

After the change, there's no output for the same command. Removing, e.g., the user-writable bit will still yield an output that now also shows the current mode of the file:

```
qt5:
 * Incorrect file permissions (400): chmod 600 /opt/homebrew/Library/Formula/qt5.rb

Error: 1 problem in 1 formula
```

UniqMartin:
Please don't merge just yet. I discovered a stupid and confusing mistake in my code comment. Will push an update shortly.

UniqMartin:
> Please don't merge just yet. I discovered a stupid and confusing mistake […]

Fixed and pushed. Now it should be ready. :)

PR #45842, opened 2015-11-09
Add new command to fix Homebrew's permissions.

mikemcquaid:
This involves some (nasty) code sharing with the install script which is a well-tested version of permissions fixing and is fairly conservative in what it fixes.

Also see https://github.com/Homebrew/install/pull/29.

CC @xu-cheng @bfontaine @DomT4 for thoughts.

### DIFF
--- a/Library/Homebrew/cmd/fix-permissions.rb
+++ b/Library/Homebrew/cmd/fix-permissions.rb
@@ -1,0 +1,117 @@
+require "extend/ENV"
+
+module Homebrew
+  def fix_permissions
+    fix_permissions_shared(:wait => false)
+    system "/usr/bin/sudo", "-k"
+  end
+
+  ### !!!!!!!!!!
+  ### EVERYTHING between these markers is shared (and should be kept in sync)
+  ### with Homebrew/install/install
+  ### !!!!!!!!!!
+  def sudo *args
+    ohai "/usr/bin/sudo #{args.join ' '}"
+    system "/usr/bin/sudo", *args
+  end
+
+  def fix_permissions_shared(options={})
+    wait = options.fetch(:wait, true)
+
+    def chmod?(d)
+      File.directory?(d) && !(File.readable?(d) && File.writable?(d) && File.executable?(d))
+    end
+
+    def chown?(d)
+      File.directory?(d) && !File.owned?(d)
+    end
+
+    def chgrp?(d)
+      File.directory?(d) && !File.grpowned?(d)
+    end
+
+    def getc
+      system "/bin/stty raw -echo"
+      if STDIN.respond_to?(:getbyte)
+        STDIN.getbyte
+      else
+        STDIN.getc
+      end
+    ensure
+      system "/bin/stty -raw echo"
+    end
+
+    def wait_for_user
+      puts
+      puts "Press RETURN to continue or any other key to abort"
+      c = getc
+      # we test for \r and \n because some stuff does \r instead
+      abort unless c == 13 or c == 10
+    end
+
+    paths = %w(
+      .
+      bin
+      etc
+      include
+      lib
+      lib/pkgconfig
+      Library
+      sbin
+      share
+      var
+      var/log
+      share/locale
+      share/man
+      share/man/man1
+      share/man/man2
+      share/man/man3
+      share/man/man4
+      share/man/man5
+      share/man/man6
+      share/man/man7
+      share/man/man8
+      share/info
+      share/doc
+      share/aclocal
+    ).map { |d| File.join(HOMEBREW_PREFIX, d) }
+    chmods = paths.select { |d| chmod?(d) }
+    chowns = paths.select { |d| chown?(d) }
+    chgrps = paths.select { |d| chgrp?(d) }
+
+    unless chmods.empty?
+      ohai "The following directories will be made group writable:"
+      puts(*chmods)
+    end
+    unless chowns.empty?
+      ohai "The following directories will have their owner set to #{Tty.underline 39}#{ENV['USER']}#{Tty.reset}:"
+      puts(*chowns)
+    end
+    unless chgrps.empty?
+      ohai "The following directories will have their group set to #{Tty.underline 39}admin#{Tty.reset}:"
+      puts(*chgrps)
+    end
+
+    wait_for_user if wait && STDIN.tty?
+
+    if File.directory? HOMEBREW_PREFIX
+      sudo "/bin/chmod", "g+rwx", *chmods unless chmods.empty?
+      sudo "/usr/sbin/chown", ENV['USER'], *chowns unless chowns.empty?
+      sudo "/usr/bin/chgrp", "admin", *chgrps unless chgrps.empty?
+    else
+      sudo "/bin/mkdir", HOMEBREW_PREFIX
+      sudo "/bin/chmod", "g+rwx", HOMEBREW_PREFIX
+      # the group is set to wheel by default for some reason
+      sudo "/usr/sbin/chown", "#{ENV['USER']}:admin", HOMEBREW_PREFIX
+    end
+
+    sudo "/bin/mkdir", HOMEBREW_CACHE unless File.directory? HOMEBREW_CACHE
+    sudo "/bin/chmod", "g+rwx", HOMEBREW_CACHE if chmod? HOMEBREW_CACHE
+    sudo "/usr/sbin/chown", ENV['USER'], HOMEBREW_CACHE if chown? HOMEBREW_CACHE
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cmd/fix-permissions.rb
```diff
@@ -0,0 +1,117 @@
...
+    def chown?(d)
+      File.directory?(d) && !File.owned?(d)
+    end
+
+    def chgrp?(d)
+      File.directory?(d) && !File.grpowned?(d)
+    end
+
+    def getc
+      system "/bin/stty raw -echo"
```
bfontaine: Any reason to use `system "/bin/stty raw -echo"` instead of `system "/bin/stty", "raw", "-echo"`?

Library/Homebrew/cmd/fix-permissions.rb
```diff
@@ -0,0 +1,117 @@
...
+    end
+
+    sudo "/bin/mkdir", HOMEBREW_CACHE unless File.directory? HOMEBREW_CACHE
+    sudo "/bin/chmod", "g+rwx", HOMEBREW_CACHE if chmod? HOMEBREW_CACHE
+    sudo "/usr/sbin/chown", ENV['USER'], HOMEBREW_CACHE if chown? HOMEBREW_CACHE
+    sudo "/usr/bin/chgrp", "admin", HOMEBREW_CACHE if chgrp? HOMEBREW_CACHE
+  end
+  ### !!!!!!!!!!
+  ### EVERYTHING between these markers is shared (and should be kept in sync)
+  ### with Library/Homebrew/cmd/fix-permissions.rb
```
bfontaine: This should be “`with Homebrew/install/install`”.

DomT4:
You went about this an fairly different way to what I had in mind (Not that I have any issues with your method, sharing code makes sense).

Few of the things I had in mind to do differently:

* Don't use Homebrew methods at all, so command could be run independently of a functioning installation. Doesn't really matter. Chances of someone breaking their Homebrew so badly they couldn't execute *any* Homebrew code are slim.

* Don't automatically use `sudo`, so non-admin users could fix permission problems in their local installations/etc with the command as well. Was thinking we could rescue if permissions wouldn't fix sans sudo and advise running the command itself under `sudo` at that point.

* Dry-run automatically and require people to pass `--confirm` or something rather than asking them for permission beforehand. The differences between us there are pretty much entirely stylistic rather than substance.

As said, don't have a problem with anything you've written here. Just wondered if there's anything useful in my general thoughts on how I'd approach this.

xu-cheng:
Part of my mind is it would be great if we can unify the code in fix permission and mission check in doctor. So we only need to maintain one set of rules. 

Secondly, we need to fix permission for Homebrew code files as well.

Library/Homebrew/cmd/fix-permissions.rb
```diff
@@ -0,0 +1,117 @@
...
+      system "/bin/stty -raw echo"
+    end
+
+    def wait_for_user
+      puts
+      puts "Press RETURN to continue or any other key to abort"
+      c = getc
+      # we test for \r and \n because some stuff does \r instead
+      abort unless c == 13 or c == 10
+    end
```
xu-cheng: Maybe we should avoid interactive prompt and use flag instead. As bonus, we can use this command inside Travis CI instead of `sudo chown -R`.

Library/Homebrew/cmd/fix-permissions.rb
mikemcquaid: :+1:

Library/Homebrew/cmd/fix-permissions.rb
mikemcquaid: Nope!

Library/Homebrew/cmd/fix-permissions.rb
mikemcquaid: We can probably remove this method altogether as it's not actually called.

mikemcquaid:
> Don't use Homebrew methods at all, so command could be run independently of a functioning installation. Doesn't really matter. Chances of someone breaking their Homebrew so badly they couldn't execute any Homebrew code are slim.

I was thinking of something similar. I also considered if we should do a `git reset --hard origin/master` and similar.

> Part of my mind is it would be great if we can unify the code in fix permission and permission check in doctor. So we only need to maintain one set of rules.

Good point.

> Secondly, we need to fix permission for Homebrew code files as well.

Given this has been sufficient for the installer I'd like to see if it's enough for now.

DomT4:
> I also considered if we should do a git reset --hard origin/master and similar.

:+1:

xu-cheng:
> Given this has been sufficient for the installer I'd like to see if it's enough for now. 

Installer is different as the homebrew code doesn't exist at all in that period of time.

xu-cheng:
> I also considered if we should do a git reset --hard origin/master and similar.

To be frankly, this behavior can be little counterintuitive

DomT4:
> To be frankly, this behavior can be little counterintuitive with command name (i.e. fix-permissions)

To an extent, sure, but if you upgrade and haven't updated Homebrew you can pretty much expect to run into all kinds of oddities like build failures we've since fixed, warnings that we don't support this OS X version because it's pre-release (had this quite a bit recently), etc.

I agree it doesn't really fit with the command name, but I'm not sure what to do on that since people already ignore the outdated please update message in `doctor` persistently.

I don't know. I'd like to update by default (without upgrading, etc) but we've had a lot of pushback on that from other maintainers in the past.

mikemcquaid:
We can change the command name if need be.

xu-cheng:
> We can change the command name if need be.

:+1:

Library/Homebrew/utils.rb
```diff
@@ -53,12 +53,6 @@ def truncate(str)
       str.to_s[0, width - 4]
     end
 
-    private
-
-    def color(n)
-      escape "0;#{n}"
-    end
-
```
UniqMartin: Do we really want to make `Tty.bold` and `Tty.underline` public methods? My feeling is they are okay as private helpers, but too cryptic for public consumption. The only use in this PR seems to be `Tty.underline 39`, but there already exists `Tty.em` for this.

Library/Homebrew/cmd/fix-permissions.rb
```diff
@@ -0,0 +1,117 @@
...
+  def fix_permissions
+    fix_permissions_shared(:wait => false)
+    system "/usr/bin/sudo", "-k"
+  end
+
+  ### !!!!!!!!!!
+  ### EVERYTHING between these markers is shared (and should be kept in sync)
+  ### with Homebrew/install/install
+  ### !!!!!!!!!!
+  def sudo *args
```
UniqMartin: Shouldn't this be `def sudo(*args)` nowadays?

Library/Homebrew/cmd/fix-permissions.rb
```diff
@@ -0,0 +1,117 @@
...
+    ohai "/usr/bin/sudo #{args.join ' '}"
+    system "/usr/bin/sudo", *args
+  end
+
+  def fix_permissions_shared(options={})
```
UniqMartin: Just a very minor nit. Add whitespace: `options = {}`?

Library/Homebrew/cmd/fix-permissions.rb
```diff
@@ -0,0 +1,117 @@
...
+    ensure
+      system "/bin/stty -raw echo"
+    end
+
+    def wait_for_user
+      puts
+      puts "Press RETURN to continue or any other key to abort"
+      c = getc
+      # we test for \r and \n because some stuff does \r instead
+      abort unless c == 13 or c == 10
```
UniqMartin: `||` instead of `or`? And maybe more stuff `brew style` would complain about? (Don't know if it's useful to point out all the style inconsistencies I happen to see.)

mikemcquaid: This stuff is all copy-pasted from the installer.

UniqMartin: Sorry, didn't want to detract from the actual discussion. I was just wondering whether we want to ignore style issues for new code (at least new in this repository), want to adjust the installer, or do something completely different. Wanted to raise this issue before the merge happens.

UniqMartin:
> We can change the command name if need be.

If it will do more than just fix permissions: yes, please! How about `brew repair`? Or maybe `brew cure` or `brew heal` (something one would do after getting a diagnosis from `brew doctor`)?

mikemcquaid:
Closing in favour of https://github.com/Homebrew/homebrew/issues/45934.

PR #45843, opened 2015-11-09
Make `brew test` use sandbox by default

mikemcquaid:
As discussed with @xu-cheng and @DomT4.


DomT4:
Part of me wants to not even give people a way of opting out of the sandbox, but I guess we should, at least for now. LGTM.

xu-cheng:
:+1:

PR #45844, opened 2015-11-09
Shibboleth SP

kbenbour:
Shibboleth Service Provider 2.5.5 based on the MacPorts portfile here: https://trac.macports.org/browser/trunk/dports/security/shibboleth/Portfile

as well as installation guide here:
https://wiki.shibboleth.net/confluence/display/SHIB2/Installation

Change to xml-tooling-c formula is required to allow xml-tooling-c to be built with curl with openssl support; shibboleth-sp otherwise cannot be built.

### DIFF
--- a/Library/Formula/shibboleth-sp.rb
+++ b/Library/Formula/shibboleth-sp.rb
@@ -30,7 +30,7 @@ class ShibbolethSp < Formula
       "--disable-debug",
       "--disable-dependency-tracking",
       "--disable-silent-rules",
-      "--prefix=#{prefix}"
+      "--prefix=#{prefix}",
     ]
     args << "--with-xmltooling=#{Formula["xml-tooling-c"].opt_prefix}"
     args << "--with-saml=#{Formula["opensaml"].opt_prefix}"
@@ -96,7 +96,7 @@ class ShibbolethSp < Formula
     unless File.exist? "#{prefix}/var/run/shibboleth"
       (prefix/"var/run/shibboleth/").mkpath
     end
-    unless File.exists? "#{prefix}/var/cache/shibboleth"
+    unless File.exist? "#{prefix}/var/cache/shibboleth"
       (prefix/"var/cache/shibboleth").mkpath
     end
   end
--- a/Library/Formula/shibboleth-sp.rb
+++ b/Library/Formula/shibboleth-sp.rb
@@ -4,9 +4,6 @@ class ShibbolethSp < Formula
   url "http://shibboleth.net/downloads/service-provider/latest/shibboleth-sp-2.5.5.tar.gz"
   sha256 "30da36e0bba2ce4606a9effc37c05cd110dafdd6d3141468c4aa0f57ce4d96ce"
 
-  option "with-homebrew-httpd22", "Use Homebrew Apache httpd 2.2"
-  option "with-homebrew-httpd24", "Use Homebrew Apache httpd 2.4"
-
   depends_on "curl" => "with-openssl"
   depends_on "opensaml"
   depends_on "xml-tooling-c" => "with-openssl"
@@ -14,16 +11,13 @@ class ShibbolethSp < Formula
   depends_on "xml-security-c"
   depends_on "log4shib"
   depends_on "boost"
-  depends_on "homebrew/apache/httpd22" if build.with? "homebrew-httpd22"
-  depends_on "homebrew/apache/httpd24" if build.with? "homebrew-httpd24"
-
-  skip_clean "var/run/shibboleth"
-  skip_clean "var/cache/shibboleth"
+  depends_on "homebrew/apache/httpd22" => :optional
+  depends_on "homebrew/apache/httpd24" => :optional
 
   def apache_configdir
-    if build.with? "homebrew-httpd22"
+    if build.with? "httpd22"
       "#{etc}/apache2/2.2"
-    elsif build.with? "homebrew-httpd24"
+    elsif build.with? "httpd24"
       "#{etc}/apache2/2.4"
     else
       "/etc/apache2"
@@ -48,8 +42,6 @@ class ShibbolethSp < Formula
     end
     system "./configure", *args
     system "make", "install"
-    (prefix/"var/run/shibboleth/").mkpath
-    (prefix/"var/cache/shibboleth").mkpath
   end
 
   def plist; <<-EOS.undent
@@ -98,6 +90,15 @@ class ShibbolethSp < Formula
         https://wiki.shibboleth.net/confluence/display/EDS10/3.1+Configuring+the+Service+Provider
     EOS
     s
+  end
+
+  def post_install
+    unless File.exist? "#{prefix}/var/run/shibboleth"
+      (prefix/"var/run/shibboleth/").mkpath
+    end
+    unless File.exists? "#{prefix}/var/cache/shibboleth"
+      (prefix/"var/cache/shibboleth").mkpath
+    end
   end
 
   test do
--- a/Library/Formula/shibboleth-sp.rb
+++ b/Library/Formula/shibboleth-sp.rb
@@ -26,17 +26,18 @@ class ShibbolethSp < Formula
     elsif build.with? "homebrew-httpd24"
       "#{etc}/apache2/2.4"
     else
-       "/etc/apache2"
+      "/etc/apache2"
     end
   end
 
   def install
     ENV.O2
-    args = []
-    args << "--disable-debug"
-    args << "--disable-dependency-tracking"
-    args << "--disable-silent-rules"
-    args << "--prefix=#{prefix}"
+    args = [
+      "--disable-debug",
+      "--disable-dependency-tracking",
+      "--disable-silent-rules",
+      "--prefix=#{prefix}"
+    ]
     args << "--with-xmltooling=#{Formula["xml-tooling-c"].opt_prefix}"
     args << "--with-saml=#{Formula["opensaml"].opt_prefix}"
     args << "--with-boost=#{Formula["boost"].opt_prefix}"
@@ -44,7 +45,7 @@ class ShibbolethSp < Formula
     args << "--with-xmlsec=#{Formula["xml-security-c"].opt_prefix}"
     args << "LDFLAGS=-L#{Formula["curl"].opt_prefix}/curl/lib"
     args << "CPPFLAGS=-I#{Formula["curl"].opt_prefix}/curl/include"
-    args << "DYLD_LIBRARY_PATH=#{prefix}/lib"
+    args << "DYLD_LIBRARY_PATH=#{lib}"
     if build.with? "homebrew-httpd22"
       args << "--enable-apache-22"
     elsif build.with? "homebrew-httpd24"
@@ -85,9 +86,10 @@ class ShibbolethSp < Formula
     s += <<-EOS.undent
     You must manually edit #{apache_configdir}/httpd.conf to include
     EOS
-    mod = "mod_shib_24.so"
     if build.with? "homebrew-httpd22"
       mod = "mod_shib_22.so"
+    else
+      mod = "mod_shib_24.so"
     end
 
     s += <<-EOS.undent
@@ -95,10 +97,10 @@ class ShibbolethSp < Formula
     EOS
 
     s+= <<-EOS.undent
-    You must also manually configure
-      #{opt_prefix}/shibboleth-sp/etc/shibboleth/shibboleth2.xml
-    as per your own requirements. For more information please see
-      https://wiki.shibboleth.net/confluence/display/EDS10/3.1+Configuring+the+Service+Provider
+      You must also manually configure
+        #{opt_prefix}/shibboleth-sp/etc/shibboleth/shibboleth2.xml
+      as per your own requirements. For more information please see
+        https://wiki.shibboleth.net/confluence/display/EDS10/3.1+Configuring+the+Service+Provider
     EOS
     s
   end
--- a/Library/Formula/shibboleth-sp.rb
+++ b/Library/Formula/shibboleth-sp.rb
@@ -35,9 +35,9 @@ class ShibbolethSp < Formula
     args << "--with-xmltooling=#{Formula["xml-tooling-c"].opt_prefix}"
     args << "--with-saml=#{Formula["opensaml"].opt_prefix}"
     args << "DYLD_LIBRARY_PATH=#{lib}"
-    if build.with? "homebrew-httpd22"
+    if build.with? "httpd22"
       args << "--enable-apache-22"
-    elsif build.with? "homebrew-httpd24"
+    elsif build.with? "httpd24"
       args << "--enable-apache-24"
     end
     system "./configure", *args
@@ -73,7 +73,7 @@ class ShibbolethSp < Formula
     s += <<-EOS.undent
       You must manually edit #{apache_configdir}/httpd.conf to include
     EOS
-    if build.with? "homebrew-httpd22"
+    if build.with? "httpd22"
       mod = "mod_shib_22.so"
     else
       mod = "mod_shib_24.so"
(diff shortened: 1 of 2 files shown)


Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,109 @@
...
+    args << "--disable-silent-rules"
+    args << "--prefix=#{prefix}"
+    args << "--with-xmltooling=#{Formula["xml-tooling-c"].opt_prefix}"
+    args << "--with-saml=#{Formula["opensaml"].opt_prefix}"
+    args << "--with-boost=#{Formula["boost"].opt_prefix}"
+    args << "--with-xerces=#{Formula["xerces-c"].opt_prefix}"
+    args << "--with-xmlsec=#{Formula["xml-security-c"].opt_prefix}"
+    args << "LDFLAGS=-L#{Formula["curl"].opt_prefix}/curl/lib"
+    args << "CPPFLAGS=-I#{Formula["curl"].opt_prefix}/curl/include"
+    args << "DYLD_LIBRARY_PATH=#{prefix}/lib"
```
bfontaine: Use `#{lib}` instead of `#{prefix}/lib`.

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,109 @@
...
+  skip_clean "var/run/shibboleth"
+  skip_clean "var/cache/shibboleth"
+
+  def apache_configdir
+    if build.with? "homebrew-httpd22"
+      "#{etc}/apache2/2.2"
+    elsif build.with? "homebrew-httpd24"
+      "#{etc}/apache2/2.4"
+    else
+       "/etc/apache2"
```
bfontaine: The indentation is slightly off here.

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,109 @@
...
+    args << "--disable-debug"
+    args << "--disable-dependency-tracking"
+    args << "--disable-silent-rules"
+    args << "--prefix=#{prefix}"
+    args << "--with-xmltooling=#{Formula["xml-tooling-c"].opt_prefix}"
+    args << "--with-saml=#{Formula["opensaml"].opt_prefix}"
+    args << "--with-boost=#{Formula["boost"].opt_prefix}"
+    args << "--with-xerces=#{Formula["xerces-c"].opt_prefix}"
+    args << "--with-xmlsec=#{Formula["xml-security-c"].opt_prefix}"
+    args << "LDFLAGS=-L#{Formula["curl"].opt_prefix}/curl/lib"
```
DomT4: Is it mandatory to use a Homebrewed curl built against OpenSSL? It can't use the system curl?

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,109 @@
...
+      "#{etc}/apache2/2.4"
+    else
+       "/etc/apache2"
+    end
+  end
+
+  def install
+    ENV.O2
+    args = []
+    args << "--disable-debug"
```
DomT4: These four arguments can be part of the original `args` block. It's no more stylistically pleasing to have a giant row of `args <<` lines than it is to have a slightly larger initial `args = %W[` block.

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,109 @@
...
+  depends_on "opensaml"
+  depends_on "xml-tooling-c" => "with-openssl"
+  depends_on "xerces-c"
+  depends_on "xml-security-c"
+  depends_on "log4shib"
+  depends_on "boost"
+  depends_on "homebrew/apache/httpd22" if build.with? "homebrew-httpd22"
+  depends_on "homebrew/apache/httpd24" if build.with? "homebrew-httpd24"
+
+  skip_clean "var/run/shibboleth"
```
DomT4: These should go inside `HOMEBREW_PREFIX/var`, if at all possible. `skip_clean` is something of a hack.

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,109 @@
...
+
+  depends_on "curl" => "with-openssl"
+  depends_on "opensaml"
+  depends_on "xml-tooling-c" => "with-openssl"
+  depends_on "xerces-c"
+  depends_on "xml-security-c"
+  depends_on "log4shib"
+  depends_on "boost"
+  depends_on "homebrew/apache/httpd22" if build.with? "homebrew-httpd22"
+  depends_on "homebrew/apache/httpd24" if build.with? "homebrew-httpd24"
```
mikemcquaid: Let's hold off on this options for now.

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,109 @@
...
+      "#{etc}/apache2/2.2"
+    elsif build.with? "homebrew-httpd24"
+      "#{etc}/apache2/2.4"
+    else
+       "/etc/apache2"
+    end
+  end
+
+  def install
+    ENV.O2
```
mikemcquaid: What does the default break?

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,109 @@
...
+      <true/>
+    </dict>
+    </plist>
+    EOS
+  end
+
+  def caveats
+    s = ""
+    s += <<-EOS.undent
+    You must manually edit #{apache_configdir}/httpd.conf to include
```
mikemcquaid: Can you indent this by two chars (and same below)

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,109 @@
...
+    EOS
+    mod = "mod_shib_24.so"
+    if build.with? "homebrew-httpd22"
+      mod = "mod_shib_22.so"
+    end
```
mikemcquaid: Use an `else` here for readability.

Library/Formula/shibboleth-sp.rb
kbenbour: make install step breaks if you remove `ENV.O2`

Library/Formula/shibboleth-sp.rb
kbenbour: Could we leave those options in? If you remove the option to use a brewed httpd then the user is forced to configure the system apache to use shibboleth. For many users this may not be a possibility and/or they don't want to muck around with the system apache.

Library/Formula/shibboleth-sp.rb
kbenbour: I don't think this is something I can get around. When shibboleth first starts it needs var/cache/shibboleth and var/run/shibboleth to exist so it can copy some files to cache and create a pid file in run. I believe (I may be wrong about this though) that shibboleth looks for these directories with a path relative to the shibd executable. I know that using skip clean is a working solution at the current moment, and if I can figure out a way to get rid of it I certainly will.

Library/Formula/shibboleth-sp.rb
kbenbour: Yes it is mandatory. shibd will not work unless it compiles against curl libraries built with OpenSSL.

DomT4: That's a touch frustrating. You don't need the `LDFLAGS` or `CPPFLAGS` regardless; Homebrew handles that automatically for `keg_only` dependencies, which `curl` is.

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,111 @@
...
+    ENV.O2
+    args = [
+      "--disable-debug",
+      "--disable-dependency-tracking",
+      "--disable-silent-rules",
+      "--prefix=#{prefix}"
+    ]
+    args << "--with-xmltooling=#{Formula["xml-tooling-c"].opt_prefix}"
+    args << "--with-saml=#{Formula["opensaml"].opt_prefix}"
+    args << "--with-boost=#{Formula["boost"].opt_prefix}"
```
DomT4: It should be able to find a lot of thees without prodding? The ones linked into `/usr/local` and not duplicated by the system should be the first in the default search paths.

kbenbour: I was able to remove the prodding lines except for xml-tooling-c and opensaml. Without passing these in, the configure script can't find them.

Library/Formula/shibboleth-sp.rb
kbenbour: I removed the LDFLAGS and CPPFLAGS in the most recent commits

Library/Formula/shibboleth-sp.rb
mikemcquaid: Create them in `post_install`.

Library/Formula/shibboleth-sp.rb
mikemcquaid: Can you file an upstream issue about this?

Library/Formula/shibboleth-sp.rb
mikemcquaid: Ok, if you leave them in use `=> :optional` instead of `build.with?` to generate options. Also, it's not obvious which one I would pick and why.

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,106 @@
...
+      "--prefix=#{prefix}"
+    ]
+    args << "--with-xmltooling=#{Formula["xml-tooling-c"].opt_prefix}"
+    args << "--with-saml=#{Formula["opensaml"].opt_prefix}"
+    args << "DYLD_LIBRARY_PATH=#{lib}"
+    if build.with? "homebrew-httpd22"
+      args << "--enable-apache-22"
+    elsif build.with? "homebrew-httpd24"
+      args << "--enable-apache-24"
+    end
```
mikemcquaid: Does this link against Apache or just generate modules that can be used with that version?

kbenbour: Just generates the modules for those specific versions (as far as I'm aware).

Library/Formula/shibboleth-sp.rb
kbenbour: Do you mean to the author/maintainer of the shibboleth source?

Library/Formula/shibboleth-sp.rb
mikemcquaid: In that case it's not strictly a dependency, right?

Library/Formula/shibboleth-sp.rb
mikemcquaid: Yep, thanks.

Library/Formula/shibboleth-sp.rb
kbenbour: You're correct to say it is not strictly a dependency. But without this, the caveat section instructing the user to place reference to the shib library in their httpd.conf file wouldn't exist and I think it would make the whole process even more confusing than it already is by removing that.

Library/Formula/shibboleth-sp.rb
kbenbour: This is the developers choice, and im almost 100% sure he will not do anything to accommodate anything other than -O2.

mikemcquaid: Could you add a comment explaining that (and a source for it being the developers choice)? Thanks!

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,107 @@
...
+
+  depends_on "curl" => "with-openssl"
+  depends_on "opensaml"
+  depends_on "xml-tooling-c" => "with-openssl"
+  depends_on "xerces-c"
+  depends_on "xml-security-c"
+  depends_on "log4shib"
+  depends_on "boost"
+  depends_on "homebrew/apache/httpd22" => :optional
+  depends_on "homebrew/apache/httpd24" => :optional
```
mikemcquaid: Maybe just default to building for 2.4 by default and add building against 2.2 as an option. We don't need the `depends_on` here as it's not doing anything with these files and `httpd24` matches the system version (at least on 10.10).

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,107 @@
...
+    if build.with? "httpd22"
+      args << "--enable-apache-22"
+    elsif build.with? "httpd24"
+      args << "--enable-apache-24"
+    end
+    system "./configure", *args
+    system "make", "install"
+  end
+
+  def plist; <<-EOS.undent
```
mikemcquaid: Please add a `plist_options` line.

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,107 @@
...
+      <true/>
+    </dict>
+    </plist>
+    EOS
+  end
+
+  def caveats
+    s = ""
+    s += <<-EOS.undent
+      You must manually edit #{apache_configdir}/httpd.conf to include
```
mikemcquaid: Can just be `httpd.conf` and kill the `apache_configdir` function.

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,107 @@
...
+      mod = "mod_shib_24.so"
+    end
+
+    s += <<-EOS.undent
+      LoadModule mod_shib #{lib}/shibboleth/#{mod}
+    EOS
+
+    s+= <<-EOS.undent
+      You must also manually configure
+        #{opt_prefix}/shibboleth-sp/etc/shibboleth/shibboleth2.xml
```
mikemcquaid: Can this file be installed into `etc` with `etc.install`? Otherwise it'll be removed on every upgrade.

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,107 @@
...
+      as per your own requirements. For more information please see
+        https://wiki.shibboleth.net/confluence/display/EDS10/3.1+Configuring+the+Service+Provider
+    EOS
+    s
+  end
+
+  def post_install
+    unless File.exist? "#{prefix}/var/run/shibboleth"
```
mikemcquaid: Don't need the `unless` here; it's a no-op without it.

kbenbour: Removed these.

Library/Formula/shibboleth-sp.rb
kbenbour: Added the sysconfdir argument to the configure step so it installs shibboleth2.xml to $(brew --prefix)/etc/shibboleth

Library/Formula/shibboleth-sp.rb
kbenbour: Fixed as asked.

Library/Formula/shibboleth-sp.rb
kbenbour: Added.

Library/Formula/shibboleth-sp.rb
kbenbour: Got rid of the depends_on and replaced with an option to build 2.2 mod_shib lib.

DomT4:

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,96 @@
...
+
+  def caveats
+    s = ""
+    s += <<-EOS.undent
+      You must manually edit httpd.conf to include
+    EOS
+    if build.with? "apache-22"
+      mod = "mod_shib_22.so"
+    else
+      mod = "mod_shib_24.so"
```
kbenbour: Fixed this in the next commit

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,96 @@
...
+      <key>KeepAlive</key>
+      <true/>
+    </dict>
+    </plist>
+    EOS
+  end
+
+  def caveats
+    s = ""
+    s += <<-EOS.undent
```
kbenbour: Fixed this in the next commit

Library/Formula/shibboleth-sp.rb
```diff
@@ -0,0 +1,96 @@
...
+    if build.with? "apache-22"
+      args << "--enable-apache-22"
+    else
+      args << "--enable-apache-24"
+    end
+    system "./configure", *args
+    system "make", "install"
+  end
+
+  plist_options :startup => true
```
kbenbour: Fixed this in next commit

PR #45848, opened 2015-11-09
Add unix style default configuration file

tjamet:
Inspired from the unix `/etc/default` or `/etc/sysconfig` custom parameters sourcing for system daemons, this PR adds the same mechanism to brew.

The main motivation for this PR is to be able to register the github token without having to expose it to other programs launched from the terminal.
More over, sourcing a script makes it easier to get this token encrypted through a third party tool such as pass (http://www.passwordstore.org/ | https://github.com/stuartsierra/password-store)

### DIFF
--- a/bin/brew
+++ b/bin/brew
@@ -17,6 +17,20 @@ fi
 
 BREW_LIBRARY_DIRECTORY=$(chdir "$BREW_FILE_DIRECTORY"/../Library && pwd -P)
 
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Sorry, having no configuration file is a design decision we made a long time ago and want to maintain. Thanks for the PR anyway!

PR #45849, opened 2015-11-09
ruby 1.8: address syntax warnings in Homebrew core

UniqMartin:
I figured we might want to address these before turning on the Ruby 1.8 CI. All of them are reported as warnings when running `brew readall --aliases --syntax` with Ruby 1.8.7.

CC @mikemcquaid @DomT4

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -600,7 +600,7 @@ class FormulaAuditor
       problem "\"Formula.factory(name)\" is deprecated in favor of \"Formula[name]\""
     end
 
-    if text =~ /system "npm", "install"/ && text !~ %r[opt_libexec}/npm/bin]
+    if text =~ /system "npm", "install"/ && text !~ %r[opt_libexec\}/npm/bin]
       need_npm = "\#{Formula[\"node\"].opt_libexec\}/npm/bin"
       problem <<-EOS.undent
        Please add ENV.prepend_path \"PATH\", \"#{need_npm}"\ to def install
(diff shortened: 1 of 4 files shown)


DomT4:
If it's useful to other maintainers for reference:

```
/usr/local/Library/Homebrew/cmd/audit.rb:603: warning: regexp has `}' without escape
/usr/local/Library/Homebrew/test/test_formula.rb:36: warning: don't put space before argument parentheses
/usr/local/Library/Homebrew/test/test_migrator.rb:187: warning: (...) interpreted as grouped expression
/usr/local/Library/Homebrew/test/test_pathname.rb:219: warning: (...) interpreted as grouped expression
Error: problem in /usr/local/Library/Formula/valgrind.rb
compile error
/usr/local/Library/Formula/valgrind.rb:45: undefined (?...) sequence: /(?<=\s)(?=\/usr\/include\/mach\/)/
```

DomT4:
Looks like I need to go revert https://github.com/Homebrew/homebrew/commit/6d21fefcecc034a6d3875625b2892a1a5225547f as well.

UniqMartin:
Thanks for providing the output! I'll try to remember to add it next time I submit something similar.

DomT4:
No worries at all! I don't think any of us are going to miss Ruby 1.8 when support for it is gone, heh.

DomT4:
I'm fine with this but I'll let Mike or @xu-cheng sign off on it, just in case we're less fussed about warnings than we are fatal errors.

xu-cheng:
Thanks @UniqMartin :tada:

PR #45850, opened 2015-11-10
saldl v31 (new formula)

robertgzr:
https://github.com/saldl/saldl


Library/Formula/saldl.rb
```diff
@@ -0,0 +1,39 @@
+class Saldl < Formula
+  desc "CLI downloader optimized for speed and early preview."
+  homepage "https://saldl.github.io"
+  url "https://github.com/saldl/saldl/archive/v31.tar.gz"
+  # version "31"
```
dunn: You can scratch this comment, thanks!

Library/Formula/saldl.rb
```diff
@@ -0,0 +1,39 @@
...
+    # head uses git describe to acquire a version
+    args << "--saldl-version=v#{version}" unless build.head?
+
+    system "./waf", "configure", *args
+    system "./waf", "build"
+    system "./waf", "install"
+  end
+
+  test do
+    system "saldl", "--version"
```
dunn: We'd love a more comprehensive test.  [`when.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/when.rb) and [`scons.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/scons.rb) are good examples of application tests.

Library/Formula/saldl.rb
```diff
@@ -0,0 +1,39 @@
...
+  sha256 "146a980ae2109a391c7a8ab6a1c525458db9bf18f3f0477731c9b876630078b4"
+
+  head do
+    # git describe does not work with shallow clones
+    url "https://github.com/saldl/saldl.git", :shallow => false
+  end
+
+  depends_on "pkg-config" => :build
+  depends_on "asciidoc" => :build
+  depends_on "curl"
```
DomT4: Why can't it use the system curl?

Library/Formula/saldl.rb
MoSal: `saldl --version` actually does a little bit more than just printing version info. It calls a function in libcurl to quire some basic information. I think that's a sufficient **basic** test.

Library/Formula/saldl.rb
MoSal: saldl requires `libcurl >= 7.42`. This is checked at compile time and run time.

DomT4: You can do `depends_on "curl" if MacOS <= :mavericks` then. Yosemite and El Cap ship a new enough `curl`.

Library/Formula/saldl.rb
```diff
@@ -0,0 +1,39 @@
...
+  depends_on "libevent"
+
+  def install
+    # a2x/asciidoc needs this to build the man page successfully
+    ENV["XML_CATALOG_FILES"] = "#{etc}/xml/catalog"
+
+    args = [
+      "--libcurl-cflags=-I#{HOMEBREW_PREFIX}/opt/curl/include",
```
DomT4: Don't pass these two lines. Homebrew's core code handles this for `keg_only` formulae.

Library/Formula/saldl.rb
DomT4: We love functionality tests more than basic tests. In this case it looks extremely possible to download a file and assert the file exists in the testpath.

Library/Formula/saldl.rb
MoSal: I don't know about El Cap, but Yosemite ships 7.37.1 which is not new enough. And I'd rather keep following upstream releases.

DomT4: OS X 10.10.5 includes an update to `7.43.0`, which is new enough. People should have updated by now. We have a policy on using the system utilities where possible; since Yosemite's latest release and El Capitan's latest release ship a supported `curl`, we'll want to use that.

Library/Formula/saldl.rb
MoSal: Confirmed.
Those lines are not needed.

Library/Formula/saldl.rb
MoSal: Test added.

@robertgzr Please update the formula with the new changes I made.

robertgzr:
@DomT4 
I don't get why the checks fail? I did `brew audit --strict --online` without any errors.

MoSal:
@robertgzr 
The build is failing in Yosemite and El Cap due to the lack of pkg-config support for system libcurl.

This should be fixed now.
Sorry for not catching this earlier.

robertgzr:
Alright.

Now I messed up the commit message^^

Library/Formula/saldl.rb
```diff
@@ -0,0 +1,43 @@
...
+    ENV["XML_CATALOG_FILES"] = "#{etc}/xml/catalog"
+
+    args = [
+      "--prefix=#{prefix}"
+    ]
+
+    # head uses git describe to acquire a version
+    args << "--saldl-version=v#{version}" unless build.head?
+
+    # pkg-config is not supported with system libcurl
```
DomT4: This shouldn't be true. We [write](https://github.com/Homebrew/homebrew/tree/master/Library/ENV/pkgconfig/10.10) `pkg-config` files for a few system utilities that would normally ship with them.

Library/Formula/saldl.rb
```diff
@@ -0,0 +1,43 @@
...
+  depends_on "libevent"
+
+  # curl >= 7.42 is required
+  depends_on "curl" if MacOS.version <= :mavericks
+
+  def install
+    # a2x/asciidoc needs this to build the man page successfully
+    ENV["XML_CATALOG_FILES"] = "#{etc}/xml/catalog"
+
+    args = [
```
DomT4: Just do `args = ["--prefix=#{prefix}"]` rather than having a block for one line.

Library/Formula/saldl.rb
MoSal: @DomT4 
With arbitrary versions?
This check would fail with that `.pc` file:
```
pkg-config --atleast-version=7.42 libcurl
```

DomT4: That would be because the `pkg-config` files we write aren't up-to-date. I'll look into it.

MoSal:
@robertgzr 
Can you submit the latest formula.
Thank you in advance.

robertgzr:
@MoSal 
Seems like it really was the newline.

Library/Formula/saldl.rb
```diff
@@ -0,0 +1,47 @@
+class Saldl < Formula
+  desc "CLI downloader optimized for speed and early preview."
+  homepage "https://saldl.github.io"
+  url "https://github.com/saldl/saldl/archive/v31.tar.gz"
+  sha256 "146a980ae2109a391c7a8ab6a1c525458db9bf18f3f0477731c9b876630078b4"
+
+  head do
+    # git describe does not work with shallow clones
```
DomT4: Don't need this comment.

Library/Formula/saldl.rb
```diff
@@ -0,0 +1,47 @@
...
+    url "https://github.com/saldl/saldl.git", :shallow => false
+  end
+
+  depends_on "pkg-config" => :build
+  depends_on "asciidoc" => :build
+  # Install DTDs locally for a faster and more reliable build result
```
DomT4: Don't need this comment.

Library/Formula/saldl.rb
```diff
@@ -0,0 +1,47 @@
...
+  depends_on "docbook-xsl" => :build
+
+  depends_on "libevent"
+
+  # Add option to use keg_only curl formula instead of the version provided by OSX
```
DomT4: Would prefer not to provide this option personally, unless wider demand crops up over time.

MoSal: I would prefer it if this option stays, unless you consider it a blocker.

1- I would like to provide instructions that don't require manually editing formulas for HTTP/2 support.

2- Setting a default protocol is a library feature that appeared in `7.45`. The default protocol is set to `HTTPS` if the required `libcurl` version is available.

3- OSX will always lag behind upstream releases. And each release comes with a set of fixes and features that I will probably make use of. `HTTPS proxy` support is an important feature that is expected to be introduced in the `7.47` release.

DomT4: In regards to 3, the flipside of that is because there are so many different (and untested by CI) ways you can build `curl` you are *much, much more* likely to encounter breakage using our `curl` when tied into other formulae than you would the system's.

If you really want to keep the dep, feel free, but let's make it less bulky:

```ruby
if MacOS.version <= :mavericks
  depends_on "curl"
else
  depends_on "curl" => :optional
end
```
And then drop the option.

MoSal:
@robertgzr 
Please update the formula again.
Hopefully this will be accepted.
Thank you for your cooperation.

MoSal:
The error in the failing test:

```
==> brew style saldl

ERROR:  Error installing rubocop:

	rainbow requires Ruby version >= 1.9.2.

Error: Failure while executing: gem install --no-ri --no-rdoc --user-install rubocop -v 0.35.1

==> FAILED
```

This has nothing to do with the formula or saldl.

Library/Formula/saldl.rb
```diff
@@ -0,0 +1,42 @@
+class Saldl < Formula
+  desc "CLI downloader optimized for speed and early preview."
+  homepage "https://saldl.github.io"
+  url "https://github.com/saldl/saldl/archive/v31.tar.gz"
+  sha256 "146a980ae2109a391c7a8ab6a1c525458db9bf18f3f0477731c9b876630078b4"
+
+  head do
```
DomT4: This doesn't need a block. Can just be:

```ruby
sha256 ""
head "https://github.com/saldl/saldl.git", :shallow => false
```
You only need a block if you have head-only deps or patching, etc.

Library/Formula/saldl.rb
```diff
@@ -0,0 +1,42 @@
...
+    url "https://github.com/saldl/saldl.git", :shallow => false
+  end
+
+  depends_on "pkg-config" => :build
+  depends_on "asciidoc" => :build
+  depends_on "docbook-xsl" => :build
+
```
DomT4: Don't need this empty line.

DomT4:
> This has nothing to do with the formula or saldl.

The Travis failure can be ignored here, yes.

MoSal:
@robertgzr 
Can you update the formula again?
Thank you.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `saldl 31 (new formula)`.

PR #45851, opened 2015-11-10
valgrind: build on Xcode-only systems

tamird:
This time, don't use lookbehind which Ruby 1.8.7 doesn't have.

@DomT4 cc #45801

### DIFF
--- a/Library/Formula/valgrind.rb
+++ b/Library/Formula/valgrind.rb
@@ -39,6 +39,12 @@ class Valgrind < Formula
     end
 
     system "./autogen.sh" if build.head?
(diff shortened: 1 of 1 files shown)


DomT4:
Seems fine to me. Local Ruby 1.8.x check isn't throwing a giant tantrum this time, so.

Merged in https://github.com/Homebrew/homebrew/commit/9938e5209d25cb137f08bf421478d3936ced3822. Thanks for the quick follow up and fix @tamird :star2:

tamird:
Let's hope it sticks!

PR #45855, opened 2015-11-10
add .editorconfig

xu-cheng:
See http://editorconfig.org

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,32 @@
+# http://editorconfig.org
+
+[Library/{brew.rb,Homebrew/**.rb}]
+charset = utf-8
+end_of_line = lf
+indent_style = space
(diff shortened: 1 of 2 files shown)


mikemcquaid:
I'd probably rather we didn't add this.

xu-cheng:
The idea is to let formula author's editors auto pick up this config file. So we don't need to manually point out style problem(mainly indent size etc). 

But I'm OK to both include or not include such change.

mikemcquaid:
Yeh, I guess my problem is that it requires a plugin for most editors, no?

xu-cheng:
While hand of editors support this without plugin. Additional I saw many OSS projects(e.g. Ruby) includes this.

But I guess I will pass this.

PR #45875, opened 2015-11-11
apt-cacher-ng: fix 0.8.6 missing installed files

ericvw:
Currently, there are files missing needed for configuration and to view
the statistics reporting and configuration web pages.  This change
updates the formula to utilize `make install` to ensure all essential
files are present.

Other non-essential items to note are:
* The out-of-source build setup is removed to simplify the formula and
  because the benefit in the context of Homebrew builds is negligible.
* Minor cosmetic changes (i.e., including parenthesis for method calls),
  which were picked up by linting for code clarity.

### DIFF
--- a/Library/Formula/apt-cacher-ng.rb
+++ b/Library/Formula/apt-cacher-ng.rb
@@ -30,19 +30,14 @@ class AptCacherNg < Formula
     (var/"spool/apt-cacher-ng").mkpath
     (var/"log").mkpath
 
-    mkdir "build" do
-      system "cmake", "..", *std_cmake_args
-      system "make", "apt-cacher-ng"
+    inreplace "conf/acng.conf.in" do |s|
+      s.gsub!(/^CacheDir: .*/, "CacheDir: #{var}/spool/apt-cacher-ng")
+      s.gsub!(/^LogDir: .*/, "LogDir: #{var}/log")
     end
 
-    inreplace "build/conf/acng.conf" do |s|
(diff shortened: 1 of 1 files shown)


DomT4:
> This change updates the formula to utilize make install to ensure all essential files are present.

I'm 99% sure we're not using `make install` deliberately and for good reasons. What those reasons are off the top of my head I can't remember, but this was an intentional choice at some point.

ericvw:
> I'm 99% sure we're not using `make install` deliberately and for good reasons.

Is this for all Homebrew formulae or specifically for `apt-cacher-ng`?

> What those reasons are off the top of my head I can't remember, but this was an intentional choice at some point.

So how should we proceed forward?  I may be naive, but it seems that utilizing `make install` ensures everything is staged properly in `/usr/local/Cellar/apt-cacher-ng` to then be symlink'd to the respective places under `/usr/local` versus having to duplicate what `make install` does in the formula itself, which may be difficult to maintain as the underlying package evolves what is necessary to install (or not install).

In making this change, I went with the path of least resistance :).  I'll dig around to see if I can find some context with respect to `make install`.

ericvw:
According to `git log -p -- apt-cacher-ng.rb`, it appears that `make install` never existed for `apt-cacher-ng` formula before; the files were manually installed explicitly using the formula facilities provided.

mikemcquaid:
@ericvw A PR to use `make install` would be great.

mikemcquaid:
> @ericvw A PR to use make install would be great.

(I thought this was an issue! Reviewing now)

mikemcquaid:

PR #45881, opened 2015-11-11
bazel 0.1.1 (new formula)

kashif:


kashif:
Travis CI failing due to error:

```
...
The Bazel installer must have write access to /usr/local/etc!
...
```

Can the travis user not write to `/usr/local/etc/` ?

Library/Formula/bazel.rb
```diff
@@ -0,0 +1,23 @@
+class Bazel < Formula
+  desc "Google's own build tool"
+  homepage "http://bazel.io/"
+  url "https://github.com/bazelbuild/bazel/releases/download/0.1.1/bazel-0.1.1-installer-darwin-x86_64.sh"
```
mikemcquaid: This needs to be built from source.

Library/Formula/bazel.rb
```diff
@@ -0,0 +1,23 @@
+class Bazel < Formula
+  desc "Google's own build tool"
+  homepage "http://bazel.io/"
```
mikemcquaid: Can this be an HTTPS URL?

Library/Formula/bazel.rb
```diff
@@ -0,0 +1,23 @@
...
+  def install
+    chmod 0755, "./bazel-0.1.1-installer-darwin-x86_64.sh"
+    system "./bazel-0.1.1-installer-darwin-x86_64.sh", "--prefix=#{libexec}",
+                                                       "--bazelrc=#{etc}"
+
+    bin.install_symlink Dir["#{libexec}/bin/*"]
+  end
+
+  test do
+    system "#{bin}/bazel", "version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Library/Formula/bazel.rb
```diff
@@ -0,0 +1,23 @@
+class Bazel < Formula
+  desc "Google's own build tool"
+  homepage "http://bazel.io/"
+  url "https://github.com/bazelbuild/bazel/releases/download/0.1.1/bazel-0.1.1-installer-darwin-x86_64.sh"
+  version "0.1.1"
+  sha256 "2d34e19a47067f8ab793c634146a117552666120c690630d0c9ab0039aafb48a"
+
+  bottle :unneeded
+
+  depends_on :java
```
mikemcquaid: Any particular version?

Library/Formula/bazel.rb
kashif: no its only http at the moment :cry:

Library/Formula/bazel.rb
kashif: @mikemcquaid building from source seems to be a bit confusing on where the built files etc. are... I could look into it, but the binary version I think is now working with better tests... could you kindly have a look?

mikemcquaid: We don't support binaries outside `homebrew/binary` or `Caskroom/cask`

damienmg:
Thank you for doing that!

Please ping me if you need any help from the Bazel team with that :+1:

kashif:
thanks @damienmg could you advise me how to compile it from source, in that which folders and files need to be in `libexec` and how to generate the `bazelrc` file?

damienmg:
./compile.sh should generate a binary in output/bazel and a base_workspace in base_workspace.

The installation is simply put the bazel binary somewhere (and bazel.bazelrc next to it) and the base_workspace somewhere else (actually these are just symlinks, you want to do a deep copy here). The bazel.bazelrc would then looks like:

```
build --package_path=%workspace%:/path/to/base_workspace
query --package_path=%workspace%:/path/to/base_workspace
fetch --package_path=%workspace%:/path/to/base_workspace
```

damienmg:
Another easy way to get everything packaged in a tar is using the rule we do for debian package:
https://github.com/bazelbuild/bazel/blob/master/scripts/packages/BUILD#L128

Basically you do:
```
./compile.sh
./output/bazel --bazelrc=/dev/null --nomaster_bazelrc build \
      --embed_label=${version} --stamp   //scripts/packages:debian-data || exit $?
```

And you get everything that should be shiped on the file system in bazel-bin/scripts/packages/debian-data.tar.gz

The bazel rc file is made for the install path on debian. We could probably make a custom target for you if that would make your life easier.

kashif:
thanks @damienmg! @mikemcquaid can you kindly have a look now?

kashif:
@mikemcquaid do you know how to do something like `cp -RL ...` via the ruby `cp_r` command?

Library/Formula/bazel.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  depends_on :java => "1.8+"
+
+  def install
+    system "./compile.sh"
+
+    (prefix/"base_workspace").mkdir
+    cp_r Dir["base_workspace/*"], (prefix/"base_workspace"), :dereference_root => true
+    bin.install "output/bazel" => "bazel"
+    (prefix+"etc/bazel.bazelrc").write <<-EOS
```
DomT4: Can you write this directly to `(etc/"bazel/bazel.bazelrc")` instead? We prefer to keep configuration files in the top-level `etc` so they don't get pulped on updates/reinstalls/accidental removals/etc.

Library/Formula/bazel.rb
```diff
@@ -0,0 +1,53 @@
...
+build --package_path=%workspace%:#{prefix}/base_workspace
```
DomT4: Nit, but can this be indented two spaces after the initial bracket above? You can use `EOS.undent`. Just looks a bit neater.

kashif:
@DomT4 should be ready now

kashif:
@damienmg ah ok at the moment what i am doing is making the user copy the bazelrc to their home as per the caveat instructions...


I would have to look through the code to figure out where to change the default location of the bazelrc file... can you tell me?

damienmg:
Sure, here:
https://github.com/bazelbuild/bazel/blob/master/src/main/cpp/blaze_startup_options.cc#L177

kashif:
@mikemcquaid @DomT4 should be ready now

Library/Formula/bazel.rb
```diff
@@ -0,0 +1,63 @@
+class Bazel < Formula
+  desc "Google's own build tool"
+  homepage "http://bazel.io/"
+  url "https://github.com/bazelbuild/bazel/archive/0.1.1.tar.gz"
+  sha256 "49d11d467cf9e32dea618727198592577fbe76ff2e59217c53e3515ddf61cd95"
+
+  depends_on :java => "1.8+"
+
+  # Replace the default system wide rc path to /usr/local/etc/bazel/bazel.bazelrc.
+  patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

mikemcquaid: Or, if it'll never be accepted, use `inreplace` instead of `patch`.

mikemcquaid: Arguably this should be configurable by `compile.sh` arguments.

kashif:
@mikemcquaid ok can you kindly have a look now?

mikemcquaid:

PR #45882, opened 2015-11-11
Reflect move of cloc to github in recipe

ruurd:
Second attempt to correct the cloc formula reflecting the move to github.


mikemcquaid:

For future reference the preferred commit message format is `cloc: move to GitHub.`.

ruurd:
I do what I can. I would be lost without homebrew. Really.

PR #45890, opened 2015-11-11
cc: disable RUBYOPT variable.

mikemcquaid:
If this was set at build-time (i.e. by the Ruby 1.8.7 buildsystem) then `cc` could be passed arguments and fail to run.

### DIFF
--- a/Library/ENV/4.3/cc
+++ b/Library/ENV/4.3/cc
@@ -1,4 +1,4 @@
-#!/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby -W0
+#!/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby -W0 --disable-rubyopt
 
 $:.unshift Dir["/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/lib/ruby/{1.8,2.0.0}"].first
 require "pathname"


UniqMartin:
Won't this break for users of Ruby 1.8.7 as `--disable-rubyopt` is not a recognized option in that version?

mikemcquaid:
@UniqMartin 
fuuu

mikemcquaid:
Reverted in b1a1f8c.

UniqMartin:
Ruby 1.8(.7) – everybody's darling. :stuck_out_tongue_winking_eye:

DomT4:
Burn it. Burn Ruby 1.8.x to the ground :fire: :sob:

UniqMartin:
The bold move would be to stop supporting OS X releases Apple no longer supports, thus freeing us from the pain that Ruby 1.8.x is. We could also get rid of a lot of legacy code. Is any of the maintainers even using OS X 10.8 or older?

mikemcquaid:
I agree that we should do that. The hard thing is how we actually disable old OS X versions in that way.

DomT4:
Misty has Tigerbrew. We did discuss moving anything <10.9 to Tigerbrew in the past, but obviously that's a huge move in terms of migrating people over and dumping a lot on Misty suddenly.

PR #45893, opened 2015-11-11
magit: move to homebrew/emacs

dunn:
If people object it can remain in core, but I think this makes sense.


DomT4:
Seems to not be used by anything else, which makes moving a bit easier. :+1: from me.

mikemcquaid:
:+1:

PR #45899, opened 2015-11-11
apcupsd 3.14.13 (new formula)

reitermarkus:


dunn:
It's trying to write directly to `/Applications`: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/36339/version=yosemite/testReport/junit/brew-test-bot/yosemite/install_apcupsd/

Any way to change that?

reitermarkus:
Where should it install instead? To `prefix`?

dunn:
Yeah, app bundles should be installed to `prefix` or `libexec`.

Library/Formula/apcupsd.rb
```diff
@@ -0,0 +1,105 @@
...
+  depends_on "libusb-compat"
+
+  if build.with?("cgi")
+    depends_on "gd"
+  end
+
+  def install
+    cd "src/apcagent" do
+      # Install "apcagent" to prefix instead of /Applications
+      inreplace "Makefile", "Applications", "#{prefix}".gsub(/^\//, "")
```
reitermarkus: What's the best way to write this without getting audit problems?

reitermarkus:
Ok, everything's fixed now.

reitermarkus:
Anything else I need to change or can this be merged?

Library/Formula/apcupsd.rb
```diff
@@ -0,0 +1,107 @@
...
+  option "with-test", "Compile TEST driver code"
+  option "without-pcnet", "No PCNET driver"
+  option "with-modbus-usb", "Compile MODBUS/USB driver code"
+  option "without-modbus", "No MODBUS driver"
+  option "with-apcagent", "Build menubar app for Mac OS X"
+  option "with-install-distdir", "Install distribution"
+
+  depends_on "libusb-compat"
+
+  if build.with?("cgi")
```
DomT4: This doesn't need a block. `depends_on "gd" if build.with?("cgi")` is fine.

DomT4: Technically speaking, `depends_on "gd" if build.with? "cgi"` is fine as well :wink:

Library/Formula/apcupsd.rb
```diff
@@ -0,0 +1,107 @@
...
+
+      inreplace "apcupsd-start", "/sbin", "#{HOMEBREW_PREFIX}/sbin"
+      inreplace "apcupsd-start", "/etc/apcupsd", sysconfdir
+
+      inreplace "apcupsd-uninstall", "/Library/StartupItems/apcupsd", "#{prefix}.split('/')[0..-2].join('/')"
+      inreplace "apcupsd-uninstall", "/sbin", sbin
+      inreplace "apcupsd-uninstall", "/etc/apcupsd", sysconfdir
+      inreplace "apcupsd-uninstall", "/usr/share/man", man
+
+      inreplace "org.apcupsd.apcupsd.plist", "/sbin", "#{HOMEBREW_PREFIX}/sbin"
```
DomT4: Use `opt_sbin` instead of `#{HOMEBREW_PREFIX}/sbin`.

Library/Formula/apcupsd.rb
```diff
@@ -0,0 +1,107 @@
...
+  option "without-dumb", "No dumb UPS driver"
+  option "with-usb", "Compile USB UPS code"
+  option "without-net", "No network driver for slaves"
+  option "without-snmp", "No SNMP driver"
+  option "with-test", "Compile TEST driver code"
+  option "without-pcnet", "No PCNET driver"
+  option "with-modbus-usb", "Compile MODBUS/USB driver code"
+  option "without-modbus", "No MODBUS driver"
+  option "with-apcagent", "Build menubar app for Mac OS X"
+  option "with-install-distdir", "Install distribution"
```
DomT4: I'd really like to narrow down the amount of options here. Ideally key stuff that more than a few people will use.

Non-mandatory options don't get tested by our CI, so the more we include, the more likely breakage is to sneak through undetected and bite end users.

reitermarkus:
Ok, I have removed the `--disable` options, I think it doesn't hurt building them even if you don't need those options.

Also removed the `apcagent` option, because on Darwin this can't be disabled.
The `all` option fails because of the `GAPCMON` dependency and apparently `install-distdir` is enabled by default.

Library/Formula/apcupsd.rb
```diff
@@ -0,0 +1,84 @@
...
+    <<-EOS.undent
+      The apcupsd configuration files are located here:
+        #{etc}/apcupsd
+
+      In order for apcupsd to work, the apcupsd kernel extension
+      must be installed by the root user:
+
+        sudo cp -pR #{kext_prefix}/ApcupsdDummy.kext /System/Library/Extensions/
+        sudo chown -R root:wheel /System/Library/Extensions/ApcupsdDummy.kext
+        sudo touch /System/Library/Extensions/
```
reitermarkus: Are you sure it won't work on 10.10? Isn't El Capitan the first version to make the signature check mandatory? Because I have it working on a 10.10 server without problems.

mikemcquaid: https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/requirements/unsigned_kext_requirement.rb

Maybe you disabled it? Regardless, we're not accepting any more kext formulae as a result, I'm afraid.

reitermarkus: I didn't disable it. I read https://github.com/Homebrew/homebrew/issues/31164, could it be because it is included in `AppleKextExcludeList.kext`?

mikemcquaid: I'm not sure CC @DomT4 who knows more about this.

reitermarkus: I also found this: https://forums.developer.apple.com/thread/19553
Unfortunately, there's no response.

It would be interesting to know if in El Capitan `AppleKextExcludeList.kext` is ignored or not.

DomT4: In 10.10 you could disable it by turning off the unsigned kext check. In 10.11 you have to *partially* disable SIP to achieve the same thing. You can copy things into the correct location with `sudo` but OS X will ignore them if unsigned. I believe the exclusion list is ignored for 10.11, but could be wrong; there's not a lot of official documentation on that available yet.

Library/Formula/apcupsd.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      #{if MacOS.version >= :el_capitan then <<-EOS.undent
+            Note: On OS X El Capitan, System Integrity Protection (SIP) prevents
+            this kernel extension from loading. It is needed in order for apcupsd
+            to interface with UPSes connected via USB. Without this extension you
+            will have to unplug and plug the USB cable in again after each reboot.
+            This can only be circumvented by partially disabling SIP, which can
+            only be done from the Recovery partition:
+
+              csrutil enable --without kext
```
DomT4: I *really* don't want anything in Homebrew formulae that tells people to even partially disable SIP. I'd rather just include the disclaimer about needing to unplug/plug back in.

reitermarkus:
Is it Travis or the formula that is failing?

reitermarkus:
Ok, I removed the part about SIP and added a disclaimer.

Library/Formula/apcupsd.rb
```diff
@@ -0,0 +1,106 @@
+class Apcupsd < Formula
+  desc "Daemon for controlling APC UPSes"
+  homepage "http://www.apcupsd.org"
+  url "https://downloads.sourceforge.net/project/apcupsd/apcupsd%20-%20Stable/3.14.13/apcupsd-3.14.13.tar.gz"
+  sha256 "57ecbde01d0448bf8c4dbfe0ad016724ae66ab98adf2de955bf2be553c5d03f9"
+
+  option "with-usb", "Build with support for USB UPSes."
```
mikemcquaid: Maybe just worth enabling this by default?

Library/Formula/apcupsd.rb
```diff
@@ -0,0 +1,106 @@
...
+
+  depends_on "gd"
+  depends_on "libusb-compat"
```
mikemcquaid: Is this only required if using usb support? If so, let's make it `=> :optional` and rename the `with-usb` option to `with-libusb-compat`

Library/Formula/apcupsd.rb
```diff
@@ -0,0 +1,106 @@
...
+
+  def install
+    cd "src/apcagent" do
+      # Install apcagent.app to `prefix` instead of /Applications.
+      inreplace "Makefile", "Applications", "#{"#{prefix}".gsub(%r{/^\//}, "")}"
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Library/Formula/apcupsd.rb
```diff
@@ -0,0 +1,106 @@
...
+
+      # Tell the launch script and LaunchDaemon to use the appropriate sbin and config directories.
+      inreplace "apcupsd-start", "/sbin", opt_sbin
+      inreplace "apcupsd-start", "/etc/apcupsd", sysconfdir
+      inreplace "org.apcupsd.apcupsd.plist", "/sbin", opt_sbin
+      inreplace "org.apcupsd.apcupsd.plist", "/etc/apcupsd", sysconfdir
+
+      # Custom uninstaller is not needed as everything is installed to `prefix`.
+      inreplace "Makefile", /.*apcupsd-uninstall.*/, ""
+    end
```
mikemcquaid: Can you please submit these patches to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Library/Formula/apcupsd.rb
```diff
@@ -0,0 +1,106 @@
...
+    args << "--enable-usb" << "--enable-modbus-usb" if build.with? "usb"
+
+    system "./configure", *args
+    system "make", "install"
+  end
+
+  def caveats
+    s = <<-EOS.undent
+      The #{name} configuration files are located here:
+        #{etc}/apcupsd
```
mikemcquaid: We don't need this comment.

Library/Formula/apcupsd.rb
```diff
@@ -0,0 +1,106 @@
...
+
+        EOS
+      else
+        s += <<-EOS.undent
+          For #{name} to be able to communicate with UPSes connected via USB,
+          the #{name} kernel extension must be installed by the root user:
+
+            sudo cp -pR #{kext_prefix}/ApcupsdDummy.kext /System/Library/Extensions/
+            sudo chown -R root:wheel /System/Library/Extensions/ApcupsdDummy.kext
+            sudo touch /System/Library/Extensions/
```
mikemcquaid: We'll need to verify if this works on 10.11 with SIP enabled.

Library/Formula/apcupsd.rb
reitermarkus: The upstream developer told me this needs to be hardcoded for the creation of the `.pkg` installer, at least for now. The same goes for the lines below as these are also part of the `.pkg`.

Library/Formula/apcupsd.rb
reitermarkus: It doesn't, hence the if statement.

Library/Formula/apcupsd.rb
reitermarkus: The default location for the configuration files is `/etc/apcupsd` so I thought it would be good to tell the user that they are not installed in their default location.

mikemcquaid: That applies to every Homebrew package (none install outside `brew --prefix`) so can be omitted.

reitermarkus: That makes sense. =)

mikemcquaid:
Can you suggest how I can test that the kernel module loads correctly on OS X 10.11? Thanks!

reitermarkus:
I don't know of a way without an actual APC UPS to test with, but here are my test results:

On my Mac mini Server (OS X 10.10), it works without any problems (SIP enabled).

On my iMac (OS X 10.11):

  * SIP disabled: works, but only if you unplug the USB cable and plug it in again
  * SIP enabled: same as with SIP disabled

So it seems as if the kext whitelist is still adhered to by El Capitan.

Just to make sure, I tried if it would work without the kext (It didn't.), so the kext is working, but only partially. Apparently my assumption that SIP is responsible for the kext not working was wrong.

mikemcquaid:

PR #45902, opened 2015-11-11
use secure URLs for maven components

vszakats:


DomT4:
Be wary with Maven stuff you will hit sandbox issues in places, and we may ask you to fix 'em :wink:

vszakats:
Thanks for the pre-warning :) I may stand not much chance to fix any of those, but I'm staying prepared for the red flags (and hoping for getting enough clues to make the edits as/if needed).

vszakats:
Looks like a lucky case, it's all green :)

DomT4:
Colour me slightly amazed :smile:. Maven stuff keeps trying to break free from the sandbox when it downloads additional modules.

Merged. Thanks for catching these @vszakats!

Issue #45859, opened 2015-11-10
Upgrading dnscrypt should not overwrite dnscrypt-resolvers.csv with outdated version


notDavid:
After running `brew upgrade` i [noticed](https://github.com/Homebrew/homebrew/commit/7ab2b46b15e50ab1eedb5152c7432417eb5b9013#commitcomment-14288704) all of my configured DNS servers stopped working. I suspect what caused this is the 'brew upgrade` overwriting the up-to-date dnscrypt-resolvers.csv on my system with an old outdated version in [dnscrypt-proxy-1.6.0.tar.bz2](https://download.dnscrypt.org/dnscrypt-proxy/dnscrypt-proxy-1.6.0.tar.bz2).

I think what we would want to happen instead is:
- Don't overwrite dnscrypt-resolvers.csv when upgrading, if it is already present,
or even better:
- Fetch the latest up-to-date version from https://github.com/jedisct1/dnscrypt-proxy/blob/master/dnscrypt-resolvers.csv

mikemcquaid:
Could you try and open a pull request? It should probably be using `etc.install` to stop configuration getting replaced. This should help and I'm happy to walk you through anything else: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged

Thanks!

notDavid:
@mikemcquaid sure, however i'm not sure how, some observations;

i just tried manually compiling/installing dnscrypt-proxy via the [standard procedure](https://github.com/jedisct1/dnscrypt-proxy/#installation) , so without Homebrew, and dnscrypt-resolvers.csv also gets overwritten when running `make install`. The last line of the 'make install' output is: `/usr/bin/install -c -m 644 dnscrypt-resolvers.csv minisign.pub '/usr/local/share/dnscrypt-proxy'`

If i see correct `etc.install` would install the config file in `/usr/local/etc/` while Homebrew currently installs the file in `/usr/local/share/dnscrypt-proxy/`.

I have seen [pkgshare.install](http://www.rubydoc.info/github/Homebrew/homebrew/master/Formula%3Apkgshare) , should i be using that? If i understand correct, it will then overwrite the config file (which was already overwritten by 'make install' with an outdated version anyway), but at least now it will be a working/fresh one...?

notDavid:
[...]

PR #45904, opened 2015-11-11
dnscrypt-proxy: fetch latest/working version of resolvers

notDavid:
The dnscrypt-resolvers.csv in the latest release/download is (almost) always out of date, so after installing you end up with a not working dnscrypt-proxy. This fetches the latest up-to-date version from https://github.com/jedisct1/dnscrypt-proxy/blob/master/dnscrypt-resolvers.csv

### DIFF
--- a/Library/Formula/dnscrypt-proxy.rb
+++ b/Library/Formula/dnscrypt-proxy.rb
@@ -40,11 +40,8 @@ class DnscryptProxy < Formula
 
     system "./configure", *args
     system "make", "install"
-  end
 
-  def post_install
-    if build.with? "minisign"
-      (bin/"dnscrypt-update-resolvers").write <<-EOS.undent
+    (bin/"dnscrypt-update-resolvers").write <<-EOS.undent
         #!/bin/sh
         RESOLVERS_UPDATES_BASE_URL=https://download.dnscrypt.org/dnscrypt-proxy
         RESOLVERS_LIST_BASE_DIR=#{pkgshare}
@@ -63,6 +60,10 @@ class DnscryptProxy < Formula
           ${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv
       EOS
       chmod 0775, bin/"dnscrypt-update-resolvers"
+  end
+
+  def post_install
+    if build.with? "minisign"
       system bin/"dnscrypt-update-resolvers"
     end
   end
--- a/Library/Formula/dnscrypt-proxy.rb
+++ b/Library/Formula/dnscrypt-proxy.rb
@@ -41,25 +41,27 @@ class DnscryptProxy < Formula
     system "./configure", *args
     system "make", "install"
 
-    (bin/"dnscrypt-update-resolvers").write <<-EOS.undent
-        #!/bin/sh
-        RESOLVERS_UPDATES_BASE_URL=https://download.dnscrypt.org/dnscrypt-proxy
-        RESOLVERS_LIST_BASE_DIR=#{pkgshare}
-        RESOLVERS_LIST_PUBLIC_KEY="RWQf6LRCGA9i53mlYecO4IzT51TGPpvWucNSCh1CBM0QTaLn73Y7GFO3"
+    if build.with? "minisign"
+      (bin/"dnscrypt-update-resolvers").write <<-EOS.undent
+          #!/bin/sh
+          RESOLVERS_UPDATES_BASE_URL=https://download.dnscrypt.org/dnscrypt-proxy
+          RESOLVERS_LIST_BASE_DIR=#{pkgshare}
+          RESOLVERS_LIST_PUBLIC_KEY="RWQf6LRCGA9i53mlYecO4IzT51TGPpvWucNSCh1CBM0QTaLn73Y7GFO3"
 
-        curl -L --max-redirs 5 -4 -m 30 --connect-timeout 30 -s \
-          "${RESOLVERS_UPDATES_BASE_URL}/dnscrypt-resolvers.csv" > \
-          "${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.tmp" && \
-        curl -L --max-redirs 5 -4 -m 30 --connect-timeout 30 -s \
-          "${RESOLVERS_UPDATES_BASE_URL}/dnscrypt-resolvers.csv.minisig" > \
-          "${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.minisig" && \
-        minisign -Vm ${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.tmp \
-          -x "${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.minisig" \
-          -P "$RESOLVERS_LIST_PUBLIC_KEY" -q && \
-        mv -f ${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.tmp \
-          ${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv
-      EOS
-      chmod 0775, bin/"dnscrypt-update-resolvers"
+          curl -L --max-redirs 5 -4 -m 30 --connect-timeout 30 -s \
+            "${RESOLVERS_UPDATES_BASE_URL}/dnscrypt-resolvers.csv" > \
+            "${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.tmp" && \
+          curl -L --max-redirs 5 -4 -m 30 --connect-timeout 30 -s \
+            "${RESOLVERS_UPDATES_BASE_URL}/dnscrypt-resolvers.csv.minisig" > \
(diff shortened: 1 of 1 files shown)


DomT4:
Self Ref: https://github.com/Homebrew/homebrew/issues/45859#issuecomment-155728401

notDavid:
Ok, so perhaps it's an option to include [this script](https://github.com/jedisct1/dnscrypt-proxy/issues/314#issuecomment-157347007) which also checks the file signature of the downloaded resolver file - and run it after installing or updating?

I tried, and added the script, but haven't had any luck yet to actually *run* the script during install. If i run it with `system libexec/"update-resolvers.sh"` i get [this error](https://gist.github.com/notDavid/e42806444fd6d69889f2#file-homebrew-dnscrypt-proxy-error-L16) 

Anyways let me know what you guys think of this... :-)

DomT4:
> Ok, so perhaps it's an option to include this script which also checks the file signature of the downloaded resolver file - and run it after installing or updating?

Heh, I was just about to propose writing an update script for users.

> If i run it with system libexec/"update-resolvers.sh" i get this error 

You'd need to do `brew install minisign`.

Library/Formula/dnscrypt-proxy.rb
```diff
@@ -18,6 +18,7 @@ class DnscryptProxy < Formula
     depends_on "autoconf" => :build
     depends_on "automake" => :build
     depends_on "libtool" => :build
+    depends_on "minisign" => :build
```
DomT4: Why would this only be `head` (or `:build`)? Would the script not work for currently stable versions?

Library/Formula/dnscrypt-proxy.rb
```diff
@@ -26,6 +27,26 @@ class DnscryptProxy < Formula
 
   depends_on "libsodium"
 
+  def updateresolvers; <<-EOS.undent
```
DomT4: I'd rather inline this in `install` rather than give it a whole `def`.

Library/Formula/dnscrypt-proxy.rb
```diff
@@ -39,6 +60,10 @@ def install
 
     system "./configure", *args
     system "make", "install"
+
+    (libexec/"update-resolvers.sh").write updateresolvers
+    chmod 0774, libexec/"update-resolvers.sh"
+    bin.write_exec_script libexec/"update-resolvers.sh"
```
DomT4: I'd probably drop the `.sh` here so people don't have to remember that part. Plenty of shell-only executables that aren't called without it.

Library/Formula/dnscrypt-proxy.rb
```diff
@@ -26,6 +27,26 @@ class DnscryptProxy < Formula
 
   depends_on "libsodium"
 
+  def updateresolvers; <<-EOS.undent
+    #!/bin/sh
+    RESOLVERS_UPDATES_BASE_URL=https://download.dnscrypt.org/dnscrypt-proxy
+    RESOLVERS_LIST_BASE_DIR=#{share}/dnscrypt-proxy
```
DomT4: This would just be `pkgshare`, which is shorthand for `#{share}/dnscrypt-proxy`.

notDavid:
> You'd need to do brew install minisign.

It's already installed (with Homebrew), but apparently it can't be found during the installation for some reason?
If i open a shell when the error occurs, [i can't find minisign because it's not in the $PATH](https://gist.github.com/notDavid/b2ea340649f1543258c5#file-gistfile1-txt-L30). However, in a default terminal window minisign works, and also the script runs fine...

DomT4:
> If i open a shell when the error occurs, i can't find minisign because it's not in the $PATH.

We remodel the `$PATH` during our compiles to stop random cross-contamination issues. We re-add each dependency's bin folder to the `$PATH` during build. It's not happening for you I suspect because you've left the `minisign` dependency inside the `head do` block and you're not doing a `head` install.

notDavid:
@DomT4 Ah that makes sense, thanks for pointing me in the right direction... again :)

Library/Formula/dnscrypt-proxy.rb
```diff
@@ -25,6 +25,7 @@ class DnscryptProxy < Formula
   deprecated_option "plugins" => "with-plugins"
 
   depends_on "libsodium"
+  depends_on "minisign"
```
DomT4: I'd probably make this `=> :recommended` so people who aren't interested in using the script can opt-out of the dep.

Library/Formula/dnscrypt-proxy.rb
```diff
@@ -39,6 +40,27 @@ def install
...
+        "${RESOLVERS_UPDATES_BASE_URL}/dnscrypt-resolvers.csv.minisig" > \
+        "${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.minisig" && \
+      minisign -Vm ${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.tmp \
+        -x "${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.minisig" \
+        -P "$RESOLVERS_LIST_PUBLIC_KEY" -q && \
+      mv -f ${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.tmp \
+        ${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv
+    EOS
+    chmod 0774, libexec/"update-resolvers.sh"
+    system libexec/"update-resolvers.sh"
```
DomT4: This should be run `post-install` so the bottles don't hold an outdated version of the resolvers, and only if people choose to build with `minisign`.

Library/Formula/dnscrypt-proxy.rb
```diff
@@ -41,6 +42,31 @@ def install
...
+        curl -L --max-redirs 5 -4 -m 30 --connect-timeout 30 -s \
+          "${RESOLVERS_UPDATES_BASE_URL}/dnscrypt-resolvers.csv.minisig" > \
+          "${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.minisig" && \
+        minisign -Vm ${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.tmp \
+          -x "${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.minisig" \
+          -P "$RESOLVERS_LIST_PUBLIC_KEY" -q && \
+        mv -f ${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv.tmp \
+          ${RESOLVERS_LIST_BASE_DIR}/dnscrypt-resolvers.csv
+      EOS
+      chmod 0774, libexec/"update-resolvers.sh"
```
mikemcquaid: Go for `755`

mikemcquaid: Sorry, `775`

Library/Formula/dnscrypt-proxy.rb
```diff
@@ -53,6 +79,9 @@ def caveats; <<-EOS.undent
     resolver. If you would like to change these settings, you will have to edit
     the plist file (e.g., --resolver-address, --provider-name, --provider-key, etc.)
 
+    If at some point the resolver file gets outdated, it can be updated to the
+    latest version by running: #{libexec}/update-resolvers
```
mikemcquaid: Looks like this should be `#{libexec}/update-resolvers.sh`. Might be nice to name this `dnscrypt-update-resolvers` and but it in `bin`?

notDavid:
@mikemcquaid Okay, done.

So if i see correctly the build is failing on Mavericks only. Not sure why yet though..

```
/usr/local/Cellar/dnscrypt-proxy/1.6.0_2/bin/dnscrypt-update-resolvers

Warning: The post-install step did not complete successfully
You can try again using `brew postinstall dnscrypt-proxy`
```

DomT4:
Looks like a false positive.

Library/Formula/dnscrypt-proxy.rb
```diff
@@ -41,6 +42,31 @@ def install
     system "make", "install"
   end
 
+  def post_install
+    if build.with? "minisign"
+      (bin/"dnscrypt-update-resolvers").write <<-EOS.undent
```
DomT4: You should write this script inside `def install` with the `chmod`, and then just execute it `postinstall` unless they've build without `minisig`. Otherwise you're going to run into things like this:

```
Error: Will not overwrite /usr/local/Cellar/dnscrypt-proxy/1.6.0_2/bin/dnscrypt-update-resolvers
```

Library/Formula/dnscrypt-proxy.rb
```diff
@@ -39,6 +40,32 @@ def install
 
     system "./configure", *args
     system "make", "install"
+
+    (bin/"dnscrypt-update-resolvers").write <<-EOS.undent
```
DomT4: Sorry, I was unclear. This should still be wrapped in a `if build.with? "minisign"` block.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/ba3d74ab663445902d45584a858f616cfd4dae49. Thank you for your contribution to Homebrew @notDavid; we appreciate it! :smiley_cat:

PR #45906, opened 2015-11-11
cabal-install: Allow clang for Lion and above.

### LINKED ISSUE #45385: brew install cabal-install fails due to mtl package
https://gist.github.com/0f98fb63d0c7f23a489e

I suspect I'm using the wrong compiler, since two options are not recognized:
-Wno-invalid-pp-token
-Wno-unicode

DanLipsitt:
The GHC formula has allowed llvm for Lion and above since commit
b160b16 in April. There are problems building with GCC on Mountain Lion,
and this patch resolves them. In addition, the configuration now matches
ghc.rb as claimed in the inline comments again.

Fixes #45385.

### DIFF
--- a/Library/Formula/cabal-install.rb
+++ b/Library/Formula/cabal-install.rb
@@ -13,7 +13,7 @@ class CabalInstall < Formula
 
   depends_on "ghc"
 
-  fails_with :clang if MacOS.version < :mavericks # Same as ghc.rb
+  fails_with :clang if MacOS.version < :lion # Same as ghc.rb
 
   def install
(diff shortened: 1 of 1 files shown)


DanLipsitt:
Is there a way to trigger a Mountain Lion bottle build?

DomT4:
Afraid not; we no longer have a CI hooked up for Mountain Lion. We always have CIs for the latest 3 stable OS X releases.

DanLipsitt:
OK. Well this is tested on my 10.8.5 machine.

DomT4:
If you had a log of that, that'd be great.

PR #45912, opened 2015-11-12
use secure cpan urls and minor related updates

vszakats:
* use secure cpan URLs as primary `url`s
* move insecure cpan URLs as `mirror`s
* cleanup cpan `http://www.cpan.org/` URLs to `http://search.cpan.org/CPAN/` uniformly

### DIFF
--- a/Library/Formula/clipsafe.rb
+++ b/Library/Formula/clipsafe.rb
@@ -7,20 +7,20 @@ class Clipsafe < Formula
   depends_on :macos => :mountain_lion
 
   resource "Crypt::Twofish" do
-    url "http://search.cpan.org/CPAN/authors/id/A/AM/AMS/Crypt-Twofish-2.17.tar.gz"
-    mirror "https://cpan.metacpan.org/authors/id/A/AM/AMS/Crypt-Twofish-2.17.tar.gz"
+    url "https://cpan.metacpan.org/authors/id/A/AM/AMS/Crypt-Twofish-2.17.tar.gz"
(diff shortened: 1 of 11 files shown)


DomT4:
FWIW I'm entirely happy to use `metacpan` as the primary resource URL given it is secure and the other isn't.

DomT4:
But leave `innotop` & `mytop` out of the PR please; will fix those in the MySQL PR I have open which I really don't want to do significant rebasing of :laughing:.

vszakats:
Got it! I actually wanted to inquire you whether to keep those insecure cpan URLs. I'm also happy to drop them. I may replace `innotop` and `mytop` with some http removal, if that's okay.

DomT4:
> I actually wanted to inquire you whether to keep those insecure cpan URLs.

I don't mind them as fallback mirrors, but I'm happy with people using the secure alternative instead as the primary URL.

vszakats:
I've put back the insecure one as mirrors, but I'd argue this isn't very good from a security POV. The secure URL can be easily forced to fall back to the insecure mirror and defeat the point of https altogether.

vszakats:
Also replaced `http://www.cpan.org/` with `http://search.cpan.org/CPAN/`.

DomT4:
> but I'd argue this isn't very good from a security POV

I'm aware. At some point the no insecure redirect environmental variable will block the downgrade.

DomT4:
Pulled so many bottles here :laughing:. Was going to kick it off the CI and just merge it, but most of the formulae here needed new bottles either for SHA256 or El Cap, so I let it run.

Merged in a whole bunch of commits. Thank you @vszakats :smiley:

vszakats:
Thank you @DomT4!

PR #45917, opened 2015-11-12
newlisp: make readline a recommended dependency

rakkesh:


DomT4:
Can you look into why it was mandatory?

rakkesh:
Probably because `makefile_darwin` states - "needs readline library and headerfiles installed (X tools)". 

READLINE wrapper is provided by `libedit (editline)` which `newlisp` can use on Mac OS X. All line editing functionalities work fine in `newlisp` repl.

rakkesh:
If `newlisp` is built with `libedit`, line editing functionalities are working fine but custom `.inputrc` has no effect, e.g., `set blink-matching-paren on` in `.inputrc` doesn't work in `repl`. Homebrew owners need to  take a call on whether to merge or reject this PR. I don't mind building `newlisp` from source. Thanks.

DomT4:
Maybe `=> :recommended` then? That way you could still `brew install newlisp --without-readline` yourself but not remove some functionality by default.

rakkesh:
@DomT4 Thanks for the tip. Updated the PR accordingly.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/e3b652651d21915c2749f7aecca6cdd8a2622da5. Thank you for another contribution to Homebrew @rakkesh! :smiley_cat:

PR #45932, opened 2015-11-12
gedit 3.18.2

tschoonj:


tschoonj:
Shall I just make update the `itstool` to ensure it depends on the brewed python on Mavericks?

DomT4:
The `:python` dep has been around for [about two years](https://github.com/Homebrew/homebrew/commit/8e38767b6440942e4cb6838df273333040375cce) now, but I can't find any explanation why it isn't just using the system's by default.

mikemcquaid:
> The :python dep has been around for about two years now, but I can't find any explanation why it isn't just using the system's by default.

Let's just remove it, in that case.

tschoonj:
Ok, I will file a PR over this

mikemcquaid:
@tschoonj Should be able to rebase this now and (hopefully) it'll work.

DomT4:
Mavericks is being a... Mavericks, again. Was this rebased against the master? I presume so? :confused: 

```
env: /usr/local/opt/python/bin/python: No such file or directory
make[2]: *** [gedit/Gedit-3.0.gir] Error 127
make[1]: *** [install-recursive] Error 1
make: *** [install] Error 2
```

tschoonj:
Well it looks like a different issue this time since the failure is related to gobject-introspection, which has the following line in its formula:

````ruby
depends_on :python if MacOS.version <= :mavericks
````

This may be related to a recent change by @xu-cheng in aeb6e8e. I think the dependency to the Homebrew python was necessary to avoid a build failure on Mavericks (I submitted a PR over this I think).

xu-cheng:
Are we sure that it fails because the lack of homebrew Python? Or it's the bad shebang need to be fix?

tschoonj:
Any suggestions on how to fix this? This could turn out to be a showstopper for several formulas

PR #45934, opened 2015-11-12
[WIP] add fix-permissions

xu-cheng:
Another take of #45842.

cc @mikemcquaid @bfontaine  @DomT4 @UniqMartin

### DIFF
--- a/Library/Homebrew/cmd/fix-permissions.rb
+++ b/Library/Homebrew/cmd/fix-permissions.rb
@@ -1,0 +1,15 @@
+require "permission_checker"
+
+module Homebrew
+  def fix_permissions
+    checker = PermissionChecker.new
+    if ARGV.dry_run?
+      return if checker.check!
+      opoo checker.report
+      puts "To fix permissions, run `brew fix-permissions`"
+      Homebrew.failed = true
+    else
+      checker.fix!
+    end
+  end
+end
--- a/Library/Homebrew/cmd/doctor.rb
+++ b/Library/Homebrew/cmd/doctor.rb
@@ -396,64 +396,6 @@ class Checks
     end
   end
 
-  def __check_subdir_access(base)
-    target = HOMEBREW_PREFIX+base
-    return unless target.exist?
-
-    cant_read = []
-
-    target.find do |d|
-      next unless d.directory?
-      cant_read << d unless d.writable_real?
-    end
-
-    return if cant_read.empty?
-    inject_file_list cant_read.sort, <<-EOS.undent
-      Some directories in #{target} aren't writable.
-      This can happen if you "sudo make install" software that isn't managed
-      by Homebrew. If a brew tries to add locale information to one of these
-      directories, then the install will fail during the link step.
-
-      You should probably `sudo chown -R $(whoami)` them:
-    EOS
-  end
-
-  def check_access_share_locale
-    __check_subdir_access "share/locale"
-  end
-
-  def check_access_share_man
-    __check_subdir_access "share/man"
-  end
-
-  def check_access_homebrew_repository
-    unless HOMEBREW_REPOSITORY.writable_real? then <<-EOS.undent
-      The #{HOMEBREW_REPOSITORY} is not writable.
-
-      You should probably change the ownership and permissions of #{HOMEBREW_REPOSITORY}
-      back to your user account.
-        sudo chown -R $(whoami) #{HOMEBREW_REPOSITORY}
-    EOS
-    end
-  end
-
-  def check_access_usr_local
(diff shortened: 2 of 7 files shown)


Library/Homebrew/cmd/fix-permissions.rb
```diff
@@ -0,0 +1,126 @@
...
+        opoo "#{HOMEBREW_CACHE} doesn't exist."
+      end
+
+      puts "To fix permissions, run `brew fix-permissions --force`"
+      Homebrew.failed = true
+    end
+  end
+end
+
+class PermissionChecker
```
xu-cheng: This class will be copied to install script.

Library/brew.rb
```diff
@@ -98,7 +98,7 @@ def require?(path)
   cmd = HOMEBREW_INTERNAL_COMMAND_ALIASES.fetch(cmd, cmd)
 
   sudo_check = %w[ install reinstall postinstall link pin unpin
-                   update upgrade create migrate tap switch ]
+                   update upgrade create migrate tap switch fix-permissions ]
```
mikemcquaid: Seems like maybe we should allow this and just default to `--force` if you're using `sudo`?

mikemcquaid:
Seems pretty reasonable. I'd like to see how the `install` script looks as presumably the YAML will need embedded in that script too. Maybe also worth doing the `git checkout -f master; git reset --hard origin/master` too.

DomT4:
This looks great Xu. Moving away from the constant repetition of `sudo chown -R blah` in `doctor` is a very welcome fix.

Library/Homebrew/cmd/fix-permissions.rb
```diff
@@ -0,0 +1,142 @@
...
+    load_from_yaml (HOMEBREW_LIBRARY_PATH/"permission_checklist.yml").read
+  end
+
+  # only used in install script
+  def self.load_from_remote
+    url = "https://raw.githubusercontent.com/Homebrew/homebrew/master/Library/Homebrew/permission_checklist.yml"
+    remote_yaml = `curl -fsSL ${url}`
+    raise "Fail to fetch remote resource: #{url}" if remote_yaml.empty?
+    load_from_yaml remote_yaml
+  end
```
xu-cheng: > I'd like to see how the install script looks as presumably the YAML will need embedded in that script too. 

@mikemcquaid This is how install script will load yaml file. So we only need to maintain one checklist.

Library/brew.rb
xu-cheng: I'm not sure whether running `sudo brew fix-permissions` would cause any user mismatch problem(especially in multiuser system).

mikemcquaid: Ok. Might be worth checking and/or having a more specific message?

xu-cheng: Did some local tests. I think current `sudo_check` behavior is exactly what we want.
```
$ ruby -e 'puts File.owned?("/usr/local/bin/brew")'
true

$ sudo ruby -e 'puts File.owned?("/usr/local/bin/brew")'
false

$ ruby -e 'puts ENV["USER"]'
xucheng

$ sudo ruby -e 'puts ENV["USER"]'
root
```

mikemcquaid: :+1:

xu-cheng:
I completely rewrote this PR. tests are required.

Remaining TODO list:
* [ ] write test case.
* [ ] test this against all travis images to check possible false positive result during `brew update`.

mikemcquaid:
Looks good. Would it be worth adding unification with `install` to the TODO list? With `brew update` we could use the `fetch` and `merge`/`rebase` separation to check the permissions for the actually updated files.

apjanke:
This looks like something we should still pull in. Want to migrate it to homebrew/brew to keep the discussion live?

xu-cheng:
This is still on my TODO list.

ilovezfs:
I'm not thrilled with the idea of having a built-in command that runs sudo. Maybe this should be an external utility the user has to download separately like the install script?

mikemcquaid:
@ilovezfs I think it's fine if we signal loudly enough and (for a change) perhaps prompt for confirmation.

mikemcquaid:
@xu-cheng I think it would be good to have this sooner rather than later (e.g. pre-1.0) so could you consider either passing this on to someone else or finishing it up? Thanks.

xu-cheng:
I'm kinda busy on other matters. So if anyone want to work on this, it would be fine with me. Otherwise, it will just be on my TODO list for the time being.

mikemcquaid:
@xu-cheng Thanks!

MikeMcQuaid:
Can this be ported over to the Homebrew/brew repo when it's ready to 🚢? Thanks!

PR #45939, opened 2015-11-12
keybase 1.0.0

patrickxb:
Switching to production release of new keybase CLI


patrickxb:
Please disregard this PR.  Going to use a tap.

PR #45940, opened 2015-11-12
qt5: raise minimum OS X requirement to 10.8

UniqMartin:
Qt 5.5.1 has been reported to fail to build on OS X 10.7 with the latest available Xcode for that OS X release. There is no easy workaround.

Also fail early if user tries to specify a compiler different from Apple Clang, as that is the only one currently supported and the Qt build system doesn't exactly facilitate switching to other compilers (except for the even more outdated Apple GCC).

*This is in response to [this comment](https://github.com/Homebrew/homebrew/issues/44998#issuecomment-156093215) and what followed it. It is probably safe to abort the CI build after the PR has passed the first few tests and just merge this. Or the CI timeout needs to be bumped.*

CC @mikemcquaid

### DIFF
--- a/Library/Formula/qt5.rb
+++ b/Library/Formula/qt5.rb
@@ -62,9 +62,11 @@ class Qt5 < Formula
 
   deprecated_option "qtdbus" => "with-d-bus"
 
-  # Snow Leopard is untested and support has been removed in 5.4
-  # https://qt.gitorious.org/qt/qtbase/commit/5be81925d7be19dd0f1022c3cfaa9c88624b1f08
-  depends_on :macos => :lion
+  # OS X 10.7 Lion is still supported in Qt 5.5, but is no longer a reference
+  # configuration and thus untested in practice. Builds on OS X 10.7 have been
+  # reported to fail: <https://github.com/Homebrew/homebrew/issues/45284>.
+  depends_on :macos => :mountain_lion
+
   depends_on "d-bus" => :optional
   depends_on :mysql => :optional
   depends_on :xcode => :build


Library/Formula/qt5.rb
```diff
@@ -72,6 +74,13 @@ class Qt5 < Formula
   depends_on OracleHomeVarRequirement if build.with? "oci"
 
   def install
+    # Save users some frustration by telling them that `--cc=gcc-5` or similar
+    # won't work. It won't be picked up by the Qt build system, there is no easy
+    # way to select an arbitrary compiler, and it's not supported anyway.
+    if ENV.compiler != :clang
```
DomT4: You could `fails_with` anything that isn't Clang, rather than raising, maybe?

mikemcquaid: :+1:

UniqMartin: I don't think I can do this. If so, how do I express that it fails with everything but Clang?

If I say `fails_with :gcc` it will mean Apple GCC 4.2. If I say `fails_with :gcc => "5"` it will apply to the non-Apple GCC, but only to a specific series. And I haven't even talked about other compilers, mostly because the only other ones recognized by `CompilerFailure` are `:gcc_4_0` and `:llvm` and these are becoming increasingly irrelevant.

DomT4:
```ruby
fails_with :gcc
fails_with :llvm
fails_with :gcc_4_0
fails_with :gcc => "5"
("4.3".."4.9").each do |n|
  fails_with :gcc => n
end
```

Something like that would likely work.

UniqMartin: Can you explain why this is better? To me it feels more repetitive, unlike the `raise` version it doesn't tell the user what's wrong, and it will break (or require adjustment) once GCC 6 is released.

xu-cheng: We probably need a `fails_without` DSL.

> Can you explain why this is better?

`fails_with` has different semantical meaning than `raise` inside `install`. For example, it should be able to be recognized in pre-install. (Though, current implementation isn't good enough)

DomT4: > We probably need a fails_without DSL.

Agreed.

UniqMartin: I see and agree that it is desirable to perform this check as early as possible, certainly before unpacking the source archive and starting to build. Maybe it would be good to have a more general system for checking preconditions, but maybe “requirements” are just that and provide sufficient coverage, now that I think a bit more about it.

So lacking a `fails_without`, maybe a cleaner solution would be `depends_on AppleClangRequirement` with the following implementation. What do you think?

```ruby
class AppleClangRequirement < Requirement
  fatal true
  satisfy { ENV["HOMEBREW_CC"] == "clang" }

  def message
    "Apple Clang is the only compiler officially supported by Qt 5."
  end
end
```

DomT4: Custom requirements in the core to save us fixing core bugs soon are a bit bleh, but yeah, it'd work.

UniqMartin: My intention was to put it into `qt5.rb`. Wouldn't be the first requirement to live in a formula. It would be great to find a cleaner solution directly in core, but I don't want to put that on my personal TO DO list right now …

DomT4: Apologies for the bad wording. That's what I get for trying to do a bunch of things at the same time, heh. I meant to say "Custom requirements in core formula to save us fixing core bugs...".

mikemcquaid: I guess we don't technically need the repeated `fails_with` unless we know they actually fail.

UniqMartin: > I guess we don't technically need the repeated `fails_with` unless we know they actually fail.

An actual failure is not the point here. The point is, whatever the user decides to specify via `--cc` is mostly going to be ignored by the Qt build system. As stated in my code comment above I want to save users some frustration if they try to build with a different compiler than Apple Clang and then find out their choice was silently ignored.

I came to this realization and believe it to be true after looking at Qt's `mkspecs` and after some personal experiments. Notice how `??.make.cc` doesn't exist or is very small in a Qt build as it doesn't use the compiler wrappers provided by `superenv`.

UniqMartin: My last comment sounds more rude than I was hoping it would. Sorry!

I'd be alright without this part and just raising the minimum OS X requirement (the primary goal of this PR). I could revise this part once a better solution in core is available (either implemented by me or someone else). How does that sound?

mikemcquaid: That's OK, don't worry. I think probably `fails_with :gcc => "5"` with a decent enough message will be enough for now.

> I came to this realization and believe it to be true after looking at Qt's mkspecs and after some personal experiments. Notice how ??.make.cc doesn't exist or is very small in a Qt build as it doesn't use the compiler wrappers provided by superenv.

When you get bored making Qt respect superenv seems like a good wee project.

UniqMartin: > I think probably `fails_with :gcc => "5"` with a decent enough message will be enough for now.

When you talk about the message you mean the code comment I would put inside the formula code above the `fails_with` line, aren't you?

Also, some bad news I guess: The restrictions implied by `fails_with` aren't checked if a compiler is manually selected via `--cc`. However, they are checked if the compiler is selected via the `HOMEBREW_CC` environment variable or auto-selected. Is this a bug or by design? (The [relevant code](https://github.com/Homebrew/homebrew/blob/06cb002431c05417454d2fd2283d722144d24b5e/Library/Homebrew/extend/ENV/shared.rb#L151-L163) is in `extend/ENV/shared.rb`. Observe how these cases are treated differently.)

> When you get bored making Qt respect superenv seems like a good wee project.

I'm not sure if this is a feasible project. A lot of compiler details seem to be baked in into those `mkspecs` and I can easily imagine things to break if `superenv` is forced on the Qt build system.

mikemcquaid: > Is this a bug or by design?

It's perhaps by design but I'd happily see that "design" changed.



> I'm not sure if this is a feasible project. A lot of compiler details seem to be baked in into those mkspecs and I can easily imagine things to break if superenv is forced on the Qt build system.

I'd be surprised if this breaks that badly. I suspect it's a full vs relative path thing.

UniqMartin:
Thanks for all the feedback! I just pushed an updated version with what I believe to be a good intermediate solution. If you disagree, let's drop the part about the compiler for now and just raise the minimum OS X requirement to 10.8.

That said, I hear all of the suggestions about how to improve `fails_with` and related things in core and have put them on my TO DO list. I'll try to work on them in the near future.

*And as before, it should be safe to kill the CI job before it wastes too much time.*

mikemcquaid:
@UniqMartin I'd probably just rather omit the Clang addition as-is because otherwise people may copy it into other formulae and it's not the "right" way of doing it, as mentioned. Let's either leave it as-is, add a `fails_with` for the latest GCC version or `fails_with` for them all. As this isn't actually a problem users have run into I'm not that convinced about the need to fix it.

UniqMartin:
I agree and have removed the Clang bits. Let's keep this simple. (As usual, feel free to kill the CI run.)

mikemcquaid:
Cheers @UniqMartin!

PR #45945, opened 2015-11-12
qpm v0.10.0 (new formula)

jrbarron:
qpm is a package manager designed for managing dependencies of Qt/QML (https://www.qt.io) applications.

### DIFF
--- a/Library/Formula/qpm.rb
+++ b/Library/Formula/qpm.rb
@@ -1,0 +1,48 @@
+require "language/go"
+
+class Qpm < Formula
+  desc "Package manager for Qt applications"
+  homepage "https://www.qpm.io"
+  url "https://github.com/Cutehacks/qpm.git",
+      :tag => "v0.10.0",
+      :revision => "b0dfc0f26e009ff778cc25ebb8b166fa460ea998"
+
+  depends_on "go" => :build
+
+  go_resource "github.com/golang/protobuf" do
+    url "https://github.com/golang/protobuf.git",
+        :revision => "d3d78384b82d449651d2435ed329d70f7c48aa56"
+  end
+
+  go_resource "github.com/howeyc/gopass" do
+    url "https://github.com/howeyc/gopass.git",
+        :revision => "10b54de414cc9693221d5ff2ae14fd2fbf1b0ac1"
+  end
+
+  go_resource "golang.org/x/crypto" do
+    url "https://go.googlesource.com/crypto.git",
+        :revision => "575fdbe86e5dd89229707ebec0575ce7d088a4a6"
+  end
+
+  go_resource "golang.org/x/net" do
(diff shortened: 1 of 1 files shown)


Library/Formula/qpm.rb
```diff
@@ -0,0 +1,18 @@
+class Qpm < Formula
+  desc "Package manager for Qt applications"
+  homepage "https://www.qpm.io"
+  url "https://github.com/Cutehacks/qpm.git", :tag => "v0.10.0",
```
DomT4: Would you mind putting the `tag` on its own line? Thanks. Trying to keep URL/tag/revision lines more or less in sync so people can just glance down a formula rather than quite far across.

Library/Formula/qpm.rb
```diff
@@ -0,0 +1,19 @@
...
+  homepage "https://www.qpm.io"
+  url "https://github.com/Cutehacks/qpm.git",
+      :tag => "v0.10.0",
+      :revision => "b0dfc0f26e009ff778cc25ebb8b166fa460ea998"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    system "go", "install", "qpm.io/qpm"
```
DomT4: We prefer to vendor submodules in ourselves using `go_resource`s and then use `go build` instead of `go install`.

Seems to be pulling a lot in:

```
Submodule 'src/github.com/golang/protobuf' (https://github.com/golang/protobuf.git) registered for path 'src/github.com/golang/protobuf'
Submodule 'src/github.com/howeyc/gopass' (https://github.com/howeyc/gopass) registered for path 'src/github.com/howeyc/gopass'
Submodule 'src/golang.org/x/crypto' (https://go.googlesource.com/crypto) registered for path 'src/golang.org/x/crypto'
Submodule 'src/golang.org/x/net' (https://go.googlesource.com/net) registered for path 'src/golang.org/x/net'
Submodule 'src/google.golang.org/grpc' (https://github.com/grpc/grpc-go) registered for path 'src/google.golang.org/grpc'
Cloning into 'src/github.com/golang/protobuf'...
```

jrbarron:
@mikemcquaid I've updated the formula to install a package now, but it is failing in Travis due to a failing assertion about a file existing. Running `brew test --verbose qpm` works fine locally and strangely enough, when I added a system call to `ls -al` to the test before the assertion, it passed in Travis. Any ideas?

DomT4:
Seems to be a transient issue that doesn't consistently reproduce, even on Travis. Will ignore it for now; can always review as it passes through the CI each time.

Merged in https://github.com/Homebrew/homebrew/commit/c34ee713598563a2d32563f3b90da93de2b13965. Thank you for your contribution to Homebrew @jrbarron; we appreciate it! :smiley_cat:

PR #45952, opened 2015-11-13
Use `version` to dynamically update proper directory paths in gdk-pixbuf

staticfloat:
This makes it so that we don't have to remember to update all the versioned file paths in this file.


dunn:
The problem is that loaders directory doesn't keep up with the version number—it's still at 2.10.0.  We could rename it during the installation, but then we'd also have to change every formula using it that currently hardcodes `2.10.0`.

staticfloat:
Ah, I think I must have screwed something up on my end to see a different version number.  Thanks!

PR #45960, opened 2015-11-13
mat 0.5.4 (import from boneyard)

bfontaine:
Upstream fixed most of the issues we had when building for OS X.

See https://github.com/Homebrew/homebrew-boneyard/pull/73

### DIFF
--- a/Library/Formula/mat.rb
+++ b/Library/Formula/mat.rb
@@ -1,0 +1,83 @@
+class Mat < Formula
+  desc "Metadata anonymization toolkit"
+  homepage "https://mat.boum.org/"
+  url "https://mat.boum.org/files/mat-0.5.4.tar.xz"
+  sha256 "a928cb2d5ebcafec4563b552096436771598376f8b4dded86a769c278c1314d1"
+
+  bottle do
+    cellar :any
+    sha1 "3476e92b91bddb20d386c35c997662fe08805da6" => :yosemite
+    sha1 "2f30838e07c7c342c01afa3a7faaa5cde129b4ab" => :mavericks
+    sha1 "810fe4dfe335c5dc8331734b1ea9abbe207b446d" => :mountain_lion
+  end
+
+  depends_on :python => :optional
+  depends_on "coreutils"
+  depends_on "poppler"
+  depends_on "pygobject3"
+  depends_on "exiftool" => :optional
+  depends_on "gettext" => :build
+  depends_on "intltool" => :build
+
+  resource "hachoir-core" do
+    url "https://pypi.python.org/packages/source/h/hachoir-core/hachoir-core-1.3.3.tar.gz"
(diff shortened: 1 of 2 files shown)


Library/Formula/mat.rb
```diff
@@ -0,0 +1,106 @@
...
+  depends_on :python => :optional
+  depends_on "coreutils"
+  depends_on "poppler"
+  depends_on "intltool" => :build
+  depends_on "gettext" => :build
+
+  depends_on "exiftool" => :optional
+
+  # mat imports the "gi" python module which is not available here.
+  patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks! :wink:

Library/Formula/mat.rb
```diff
@@ -0,0 +1,106 @@
...
+  sha256 "a928cb2d5ebcafec4563b552096436771598376f8b4dded86a769c278c1314d1"
+
+  bottle do
+    cellar :any
+    sha1 "3476e92b91bddb20d386c35c997662fe08805da6" => :yosemite
+    sha1 "2f30838e07c7c342c01afa3a7faaa5cde129b4ab" => :mavericks
+    sha1 "810fe4dfe335c5dc8331734b1ea9abbe207b446d" => :mountain_lion
+  end
+
+  option "with-massive-audio", "Enable massive audio format support"
```
mikemcquaid: Maybe worth just doing this b default?

bfontaine: Yeah that’s just one more resource.

Library/Formula/mat.rb
bfontaine: IIRC we reported it to upstream; I’ll try to find that on the mailing-list.

bfontaine: Just checked; we reported a few issues but not that one.

bfontaine:
Seems like the `gi` missing import could be fixed by another dependency but I can’t figure which one. I tried `pygobject` and `pygtk` without success.

DomT4:
As far as I remember, `pygobject3` provides the `gi` import.

DomT4:
(Which despite the name, does actually provide bindings for Python 2.7 by default)

bfontaine:
Wow thank you @DomT4; it was that. I was able to remove the patch but I now get a warning. Let’s say that’s a progress :wink: 

```
/usr/local/Cellar/mat/0.5.4/lib/python2.7/site-packages/libmat/strippers.py:31: PyGIWarning: Poppler was imported without specifying a version first. Use gi.require_version('Poppler', '0.18') before import to ensure that the right version gets loaded.
  from gi.repository import Poppler
```

mikemcquaid:
LGTM :+1:

PR #45961, opened 2015-11-13
psql2csv 0.4 (new formula)

fphilipe:
Description: Run a query in psql and output the result as CSV.  
Homepage: https://github.com/fphilipe/psql2csv

This is a thin wrapper around psql to run a single query and print the output as CSV to STDOUT.

Blog post with explanation: http://phili.pe/posts/postgresql-on-the-command-line/#extracting-data


Library/Formula/psql2csv.rb
```diff
@@ -0,0 +1,16 @@
...
+  sha256 "da009e8fbba669c32489beea0c5caf4c0a4184030403e8027a5820d8d8ebc536"
+
+  depends_on "postgresql"
+
+  def install
+    bin.install "psql2csv"
+  end
+
+  test do
+    system bin/"psql2csv", "--help"
```
bfontaine: Could you write a test for the actual functionality? You could use e.g. `--dry-run` to print the query.

fphilipe: Fixed, thanks!

fphilipe:
For some reason the test fails:

```
Testing psql2csv
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20151113-5296-1es0wtg.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Formula/psql2csv.rb --verbose
==> /usr/local/Cellar/psql2csv/0.6/bin/psql2csv --dry-run "SELECT 1"
Error: psql2csv: failed
<"COPY (SELECT 1) TO STDOUT WITH (FORMAT csv, HEADER true, DELIMITER ',', ENCODING 'UTF8')"> expected but was
<"COPY () TO STDOUT WITH (FORMAT csv, HEADER true, DELIMITER ',', ENCODING 'UTF8')">.
```

It looks like it's related to *sed* on [this line](https://github.com/fphilipe/psql2csv/blob/v0.6/psql2csv#L71). Has anyone experienced something like this before? Locally everything works as expected. Tried running with gnu sed, which also worked. Do the machines on travis have a different sed?

DomT4:
OS X ships BSD `sed` rather than GNU `sed`.

fphilipe:
@DomT4 Thanks. I knew that and in fact, I tested on both. But for some reason on travis it fails. Changed so that `psql2csv` uses `set -e`. Maybe this gives some insight...

fphilipe:
Fixed. The problem I had boils down to this: psql2csv checks whether STDIN is open or not. Locally everything worked as expected, but on travis it was failing.

Locally, it ran like this:

    $ if [ -t 0 ]; then echo 'closed'; else echo 'open'; fi
    closed

On travis, the command was part of a script that happens to have STDIN open, basically this:

    sh <<< "if [ -t 0 ]; then echo 'closed'; else echo 'open'; fi"
    open

PR #45965, opened 2015-11-13
tap_utils added

bfontaine:
The goal here is to move common tap functions out of `cmd/`.

cc @xu-cheng @mikemcquaid

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -1,7 +1,7 @@
 require "blacklist"
 require "cmd/doctor"
 require "cmd/search"
-require "cmd/tap"
+require "tap_utils"
 require "formula_installer"
 require "hardware"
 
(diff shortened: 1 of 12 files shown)


bfontaine:
I wonder if we should move `tap_args` on `ARGV`.

xu-cheng:
I think we can(and maybe should) keep `install_tap` as it is. It is only used to keep backwards compatibility for external command. 

As for other two methods, we can make them class method of `Tap`.

bfontaine:
@xu-cheng You’re right; I’ll keep it there.

xu-cheng:
How about adding the remaining methods into `Tap` class or at least in `tap.rb`.  Maybe I'm too picky, but `require "tap"` seems to be better than `require "tap_utils"`

bfontaine:
@xu-cheng Yes I’d be happy if we could fit these methods elsewhere instead of creating a new file.

bfontaine:
I’m closing this for now and will reopen another PR with clean changes that take into account the recent tap changes.

PR #45966, opened 2015-11-13
Common build environment methods moved from cmd/env.rb to build_environment.rb

bfontaine:
This is similar to #45965 but for `cmd/env`. ~~The other PR used `tap_utils.rb` because `tap.rb` is already taken. I used `env.rb` here because it was free.~~

### DIFF
--- a/Library/Homebrew/build_environment.rb
+++ b/Library/Homebrew/build_environment.rb
@@ -30,3 +30,34 @@ module BuildEnvironmentDSL
     @env.merge(settings)
   end
 end
+
+module Homebrew
+  def build_env_keys(env)
+    %w[
+      CC CXX LD OBJC OBJCXX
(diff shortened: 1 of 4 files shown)


xu-cheng:
Similar to last PR, we can avoid creating a new module/file by incorporating it into existing `env/share.rb` or `build_environment.rb`

bfontaine:
Thank you @xu-cheng; I edited the PR.

xu-cheng:
I wonder whether it's better to put it in `extend/ENV/shared.rb` as `ENV`'s method. i.e. `ENV#build_env_keys` and `ENV#dump_build_env`.

bfontaine:
I don’t know. cc @mikemcquaid for opinion here.

mikemcquaid:
I'm cool as-is :+1:

PR #45967, opened 2015-11-13
arangodb 2.7.1

fceller:
Upgraded to ArangoDB 2.7.1, removed patch for GCC 5 because it is now available in upstream


apjanke:
The test failure here is just a timed-out download.

```
==> brew fetch --retry --build-bottle arangodb

==> Downloading https://www.arangodb.com/repositories/Source/ArangoDB-2.7.1.tar.gz

```

We can ignore that, since Jenkins still passed.

apjanke:

PR #45968, opened 2015-11-13
keybase 1.0.0

patrickxb:
We have a new command line client.  It has many new features and is written in Go instead of Node.  This is our first public release of the new command line client.

### DIFF
--- a/Library/Formula/keybase.rb
+++ b/Library/Formula/keybase.rb
@@ -1,33 +1,26 @@
 class Keybase < Formula
   desc "Command-line interface to Keybase.io"
   homepage "https://keybase.io/"
-  url "https://github.com/keybase/node-client/archive/v0.8.23.tar.gz"
-  sha256 "20e7e7e7a125ddb3b0e7023c3dcaf12f12c7b183822592b164ca11dd1e6f9a25"
-  head "https://github.com/keybase/node-client.git"
+  url "https://github.com/keybase/client/archive/v1.0.0-47.tar.gz"
+  sha256 "a026fa0df54909989a50d607c0ac942857aa4ca7de00d5adef8ef30f403a082d"
+  version "1.0.0-47"
 
-  bottle :unneeded
+  head "https://github.com/keybase/client.git"
 
-  depends_on "node"
-  depends_on :gpg
+  depends_on "go" => :build
 
   def install
-    # remove self-update command
-    # https://github.com/keybase/keybase-issues/issues/1477
-    rm "lib/command/update.js"
-    inreplace "lib/command/all.js", '"update", ', ""
-    inreplace "lib/req.js", "keybase-installer", "brew update && brew upgrade keybase"
+    ENV["GOPATH"] = buildpath
+    ENV["GOBIN"] = buildpath
+    ENV["GO15VENDOREXPERIMENT"] = "1"
+    (buildpath/"src/github.com/keybase/client/").install "go"
 
-    libexec.install Dir["*"]
-    (bin/"keybase").write <<-EOS.undent
-      #!/bin/sh
-      export KEYBASE_BIN="#{bin}/keybase"
-      exec "#{Formula["node"].opt_bin}/node" "#{libexec}/bin/main.js" "$@"
-    EOS
+    system "go", "build", "-a", "-tags", "production brew", "github.com/keybase/client/go/keybase"
+
+    bin.install "keybase"
   end
 
   test do
-    # Keybase requires a valid GPG keychain to be set up. Fetch Homebrew's pubkey.
-    system "gpg", "--keyserver", "pgp.mit.edu", "--recv-keys", "0xE33A3D3CCE59E297"
-    system "#{bin}/keybase", "id", "homebrew"
+    system "#{bin}/keybase", "-standalone", "id", "homebrew"
   end
 end


Library/Formula/keybase.rb
```diff
@@ -1,33 +1,33 @@
...
   desc "Command-line interface to Keybase.io"
   homepage "https://keybase.io/"
-  url "https://github.com/keybase/node-client/archive/v0.8.23.tar.gz"
-  sha256 "20e7e7e7a125ddb3b0e7023c3dcaf12f12c7b183822592b164ca11dd1e6f9a25"
-  head "https://github.com/keybase/node-client.git"
+  url "https://github.com/keybase/client/archive/v1.0.0-47.tar.gz"
+  sha256 "a026fa0df54909989a50d607c0ac942857aa4ca7de00d5adef8ef30f403a082d"
+
+  head "https://github.com/keybase/client.git"
+  version "1.0.0-47"
```
bfontaine: Please move this line just below the `url` one.

Library/Formula/keybase.rb
```diff
@@ -1,33 +1,33 @@
...
-    libexec.install Dir["*"]
-    (bin/"keybase").write <<-EOS.undent
-      #!/bin/sh
-      export KEYBASE_BIN="#{bin}/keybase"
-      exec "#{Formula["node"].opt_bin}/node" "#{libexec}/bin/main.js" "$@"
-    EOS
+    ENV["GOPATH"] = buildpath
+    ENV["GOBIN"] = buildpath
+    ENV["GO15VENDOREXPERIMENT"] = "1"
+    mkdir_p "src/github.com/keybase/client/"
```
bfontaine: Use `(buildpath/"src/github.com/keybase/client/").mkpath`

Library/Formula/keybase.rb
```diff
@@ -1,33 +1,33 @@
...
+    mv "go", "src/github.com/keybase/client/"
+
+    system "go", "build", "-a", "-tags", "production brew", "github.com/keybase/client/go/keybase"
```
bfontaine: Why is there a `brew` tag here?

Library/Formula/keybase.rb
```diff
@@ -1,33 +1,33 @@
...
-    (bin/"keybase").write <<-EOS.undent
-      #!/bin/sh
-      export KEYBASE_BIN="#{bin}/keybase"
-      exec "#{Formula["node"].opt_bin}/node" "#{libexec}/bin/main.js" "$@"
-    EOS
+    ENV["GOPATH"] = buildpath
+    ENV["GOBIN"] = buildpath
+    ENV["GO15VENDOREXPERIMENT"] = "1"
+    mkdir_p "src/github.com/keybase/client/"
+    mv "go", "src/github.com/keybase/client/"
```
bfontaine: Even better that the `mkpath` suggestion above: you can merge these two lines into:
```rb
(buildpath/"src/github.com/keybase/client/").install "go"
```

Library/Formula/keybase.rb
```diff
@@ -1,33 +1,33 @@
...
+  url "https://github.com/keybase/client/archive/v1.0.0-47.tar.gz"
+  sha256 "a026fa0df54909989a50d607c0ac942857aa4ca7de00d5adef8ef30f403a082d"
+
+  head "https://github.com/keybase/client.git"
+  version "1.0.0-47"
 
   bottle :unneeded
 
-  depends_on "node"
-  depends_on :gpg
```
xu-cheng: Just to make sure keybase doesn't require gpg anymore?

Library/Formula/keybase.rb
```diff
@@ -1,33 +1,33 @@
...
+
+  def post_install
+    system "#{opt_bin}/keybase", "launchd", "restart", "homebrew.mxcl.keybase"
   end
 
   test do
-    # Keybase requires a valid GPG keychain to be set up. Fetch Homebrew's pubkey.
-    system "gpg", "--keyserver", "pgp.mit.edu", "--recv-keys", "0xE33A3D3CCE59E297"
-    system "#{bin}/keybase", "id", "homebrew"
+    system "#{bin}/keybase", "help"
```
xu-cheng: We prefer more substantial test, e.g. original test.

Library/Formula/keybase.rb
```diff
@@ -1,33 +1,33 @@
...
+    mv "go", "src/github.com/keybase/client/"
+
+    system "go", "build", "-a", "-tags", "production brew", "github.com/keybase/client/go/keybase"
+
+    bin.install "keybase"
+  end
+
+  def post_install
+    system "#{opt_bin}/keybase", "launchd", "restart", "homebrew.mxcl.keybase"
   end
```
xu-cheng: This violates our sandbox policy. You should use `plist_options` and `plist` method. Check `mysql.rb` for example.

Library/Formula/keybase.rb
patrickxb: it's just for our help system so it can output the correct upgrade instructions if they are using homebrew.

Library/Formula/keybase.rb
patrickxb: Fixed.

Library/Formula/keybase.rb
patrickxb: Fixed.  Thanks!

Library/Formula/keybase.rb
patrickxb: No, it's not a requirement.

Library/Formula/keybase.rb
patrickxb: I understand you don't want writes to ~/Library/LaunchAgents.  

But just to clarify, all this is does is restart a previously running keybase service.  It doesn't change anything in ~/Library/LaunchAgents.  Is that not allowed either?

xu-cheng: But you have to create a plist in the first place, am I right? We require user's intervention to write any file outside of homebrew prefix. 

You can put the plist file in `prefix`, homebrew will auto pick it up to show instruction in caveats.

patrickxb: No, it works if no plist file is there.  

It's just to make sure that if they were running version 1.0.0 of the service that after `brew upgrade` to 1.0.1 it will restart the service.

patrickxb: Running `keybase [some command]` will auto-install a plist file if it doesn't exist.

Library/Formula/keybase.rb
patrickxb: Ok, I made it like equivalent to then one that existed before in b9b59a8c2894d50128252f270a925c92b1c45543.

Library/Formula/keybase.rb
DomT4: What are you using instead of GPG for things like message encryption, etc, out of curiosity?

Library/Formula/keybase.rb
DomT4: > It's just to make sure that if they were running version 1.0.0 of the service that after brew upgrade to 1.0.1 it will restart the service.

We print caveats telling users to do this with `launchctl` if there's a `plist` in the installation directory.

Library/Formula/keybase.rb
```diff
@@ -1,33 +1,32 @@
 class Keybase < Formula
   desc "Command-line interface to Keybase.io"
   homepage "https://keybase.io/"
-  url "https://github.com/keybase/node-client/archive/v0.8.23.tar.gz"
```
DomT4: The old client (`node-client`) is still referenced on the download page, Is that being updated? Just don't want to confuse people who expect one thing from the website and end up with another from us, ideally.

patrickxb: As soon as the hombrew PR gets accepted, we are flipping the switch on the website to new instructions.

Library/Formula/keybase.rb
patrickxb: We built pgp commands into the client now using golang.org/x/crypto/openpgp.  

So you can do `keybase pgp encrypt` and it doesn't farm out to gpg.

DomT4: Interesting, Thanks!

Library/Formula/keybase.rb
DomT4: :+1:

Library/Formula/keybase.rb
patrickxb: We've had issues during beta testing where users failed to do this, so we built in support for it.

DomT4: I don't have any influence over what you do upstream, but feel like we as a package manager have no right to automatically stop/start/install processes on behalf of the user in 99% of cases without that explicit consent.

Library/Formula/keybase.rb
```diff
@@ -1,33 +1,32 @@
...
-      #!/bin/sh
-      export KEYBASE_BIN="#{bin}/keybase"
-      exec "#{Formula["node"].opt_bin}/node" "#{libexec}/bin/main.js" "$@"
-    EOS
+    ENV["GOPATH"] = buildpath
+    ENV["GOBIN"] = buildpath
+    ENV["GO15VENDOREXPERIMENT"] = "1"
+    (buildpath/"src/github.com/keybase/client/").install "go"
+
+    system "go", "build", "-a", "-tags", "production brew", "github.com/keybase/client/go/keybase"
```
bfontaine: Why is there a `brew` tag here?

Library/Formula/keybase.rb
patrickxb: Ok, will remove it.

DomT4: Thanks. If you want to add a `caveat` nudging people to use that command, feel free. Just uncomfortable with us doing things like that automatically.

patrickxb: Removed in a9cf21ea5402d31fc9d4352ef5e6a575abfa5038

Library/Formula/keybase.rb
patrickxb: It's for our upgrade help for users.  It gives them homebrew-specific upgrade instructions.  As opposed to linux/windows users.

mistydemeo:
Looks like postinstall is failing on 10.9; does this work with pre-10.10 launchd commands?

patrickxb:
I just removed post_install completely.

Library/Formula/keybase.rb
bfontaine: Nice, thanks!

Library/Formula/keybase.rb
```diff
@@ -1,33 +1,28 @@
...
-  url "https://github.com/keybase/node-client/archive/v0.8.23.tar.gz"
-  sha256 "20e7e7e7a125ddb3b0e7023c3dcaf12f12c7b183822592b164ca11dd1e6f9a25"
-  head "https://github.com/keybase/node-client.git"
+  url "https://github.com/keybase/client/archive/v1.0.0-47.tar.gz"
+  sha256 "a026fa0df54909989a50d607c0ac942857aa4ca7de00d5adef8ef30f403a082d"
+  version "1.0.0-47"
+
+  head "https://github.com/keybase/client.git"
 
   bottle :unneeded
```
xu-cheng: @patrickxb Final problem before we can merge this. You need to remove this line so we can create bottles for new version keybase.

xu-cheng:
Also can you squash the commits please? Thanks.

Library/Formula/keybase.rb
patrickxb: @xu-cheng We were hoping to not bottle for security reasons.  Can we leave that line in?  Or is there a different way to turn off bottling?  Thanks.

mikemcquaid: Why would not using a bottle improve security?

xu-cheng: It's kinda bummer as users will have to install `go` formula to install keybase. And it won't be able to work on non-Xcode/CLT system. So I hope you can revisit this decision. 

If you have to turn it off, please use `bottle :disable, "your reason"`.

mikemcquaid: I don't want to include this with bottles disabled unless there's a reason I agree with.

patrickxb: Ok, I removed that line.

patrickxb:
commits squashed.  Thanks!

Library/Formula/keybase.rb
DomT4: I'm happy for someone to fund an independent audit of our CI security, if anyone's really up for that, but otherwise the insinuation Homebrew bottles are questionable is something I'm reluctant to endorse. We take CI security seriously and welcome reports on any discovered vulnerabilities at security@brew.sh.

If you remove the bottle from this formula, users will only need to install `go`, and *most* will use a bottle to do that, so you're not really minimising a perceived threat all that much.

patrickxb: Ok.  We weren't insinuating anything about your CI security.  Sorry if that was the message you received.

In an ideal world, it would be nice to be able to sign bottles.

mikemcquaid: No worries. We may add GPG signing eventually but we do check against a SHA256 hash.

mikemcquaid:

PR #45970, opened 2015-11-13
Update 10.11 pkgconfig files

DomT4:
The 10.10 ones can be done as well, if we're okay with expecting people to have updated to 10.10.5?


MoSal:
> The 10.10 ones can be done as well, if we're okay with expecting people to have updated to 10.10.5?

So, should formulas assume yosemite=10.10.5 or not?

DomT4:
@mikemcquaid @xu-cheng Thoughts on whether or not it's safe to assume OS X Yosemite users have updated to 10.10.5 in our `pkg-config` files?

mikemcquaid:
I think that's fine, yeh.

PR #45976, opened 2015-11-13
emacs: devel is now emacs-25

danieroux:
emacs-25 has been created as the devel branch: http://lists.gnu.org/archive/html/emacs-devel/2015-11/msg01326.html


dunn:
Thanks so much! Merged in ba883cba79c42e66e8403c4592fcdf2266c576b7.

PR #45977, opened 2015-11-13
yap: don't use GCC 5, which breaks

### LINKED ISSUE #41558: Yap failed to build on 10.10

apjanke:
Fixes #41558

### DIFF
--- a/Library/Formula/yap.rb
+++ b/Library/Formula/yap.rb
@@ -3,16 +3,19 @@ class Yap < Formula
   homepage "https://www.dcc.fc.up.pt/~vsc/Yap/"
   url "https://www.dcc.fc.up.pt/~vsc/Yap/yap-6.2.2.tar.gz"
   sha256 "f15f8382104443319a5883eafce5f52f4143b526c7f1cd88d19c1f63fc06d750"
-
-  bottle :disable, "Temporary workaround"
+  revision 1
 
   devel do
     url "https://www.dcc.fc.up.pt/~vsc/Yap/yap-6.3.3.tar.gz"
     sha256 "aee3b449b1669af07a8291ce6e7fb0a9b35e1343e2ab96fadb1a37552397fa78"
   end
 
+  bottle :disable, "Fails with various compilers"
+
   depends_on "gmp"
   depends_on "readline"
+
+  fails_with :gcc => "5"
 
   fails_with :clang do
     cause "uses variable-length arrays in structs, which will never be supported by clang"


mistydemeo:
Looks like this doesn't build with certain other GCCs, either. I'm testing to see what it does build with.

apjanke:
GCC 4.9 is the only one I can report "Yes" on, though their mailing list seems to indicate 4.2 should work, too.

mistydemeo:
What version of OS X are you on? I can't get it to work with 4.2, 4.8 or 4.9 on 10.11.

apjanke:
10.9.5. I've also got a 10.11 machine; pulling that out to test too...

apjanke:
I was using the system GCC 4.9 on 10.9, not the homebrew-versions `gcc49`. Would that make a difference to things?

mistydemeo:
Xcode doesn't include GCC 4.9 - Apple never shipped a version past 4.2. Xcode 6.2 (which is probably what you have?) doesn't actually include any GCC at all, only clang. We've got a version table [in the docs](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Xcode.md) for reference.

apjanke:
Oops! My mistake; it was the Homebrewed `gcc49` formula. I didn't recall ever installing it on this machine. I have Xcode 5.0.2 and 6.2 on that box.

Worked for me on 10.11.1 with `gcc49`. [Here are logs from the build.](https://gist.github.com/apjanke/1e2fa513410294246914)

apjanke:
Should we mark some additional GCCs as not working?

mikemcquaid:
> Should we mark some additional GCCs as not working?

:+1:, good idea.

DomT4:
What's the status here? Almost submitted a fresh PR before stumbled across it :blush:.

mikemcquaid:
:+1: to just merging this for now and iterating

DomT4:
May be worth rebasing on master. I added a `bottle :disable` line the other night to workaround an issue whilst I looked into it. As it turned out, the issue had been fixed by this PR :smile:.

Still won't be able to pull bottles, but can at least update the disable reasoning.

apjanke:
Rebased on master, leaving the `bottle :disable` in and updating the comment. Look right?

DomT4:
LGTM :smiley:

PR #45268, opened 2015-10-23
Adds Devd Formula

gcoonrod:
Adds Formula for [devd](https://github.com/cortesi/devd), a lightweight http server for developers.

PR #45978, opened 2015-11-13
devd v0.3 (new formula)

kad:
devd is a small cross-platform, self-contained but yet
very powerfull local webserver for developers.

Replaces old #45268

### DIFF
--- a/Library/Formula/devd.rb
+++ b/Library/Formula/devd.rb
@@ -1,0 +1,140 @@
+require "language/go"
+
+class Devd < Formula
+  desc "Local webserver for developers"
+  homepage "https://github.com/cortesi/devd"
+  url "https://github.com/cortesi/devd/archive/v0.3.tar.gz"
+  sha256 "e806421a4de6572eb3196b215a86a3b72bcd0f5a12956224d191e47663f9c4ab"
+
+  depends_on "go" => :build
+
+  go_resource "github.com/GeertJohan/go.rice" do
+    url "https://github.com/GeertJohan/go.rice.git", :revision => "ada95a01c963696fb73320ee662195af68be81ae"
+  end
+
+  # go.rice dependencies
+  go_resource "github.com/daaku/go.zipexe" do
+    url "https://github.com/daaku/go.zipexe.git", :revision => "a5fe2436ffcb3236e175e5149162b41cd28bd27d"
+  end
+
+  go_resource "github.com/kardianos/osext" do
+    url "https://github.com/kardianos/osext.git", :revision => "6e7f843663477789fac7c02def0d0909e969b4e5"
+  end
+
+  go_resource "github.com/GeertJohan/go.incremental" do
+    url "https://github.com/GeertJohan/go.incremental.git", :revision => "92fd0ce4a694213e8b3dfd2d39b16e51d26d0fbf"
+  end
+
+  go_resource "github.com/akavel/rsrc" do
+    url "https://github.com/akavel/rsrc.git", :revision => "ba14da1f827188454a4591717fff29999010887f"
+  end
+
+  go_resource "github.com/jessevdk/go-flags" do
+    url "https://github.com/jessevdk/go-flags.git", :revision => "fc93116606d0a71d7e9de0ad5734fdb4b8eae834"
+  end
+
+  # devd dependencies
+  go_resource "github.com/bmatcuk/doublestar" do
+    # v1.0.1
+    url "https://github.com/bmatcuk/doublestar.git", :revision => "4f612bd6c10e2ef68e2ea50aabc50c3681bbac86"
+  end
+
+  go_resource "github.com/dustin/go-humanize" do
+    url "https://github.com/dustin/go-humanize.git", :revision => "64dbdae0d393b7d71480a6dace78456396b55286"
+  end
+
+  go_resource "github.com/fatih/color" do
+    url "https://github.com/fatih/color.git", :revision => "9aae6aaa22315390f03959adca2c4d395b02fcef"
+  end
+
+  go_resource "github.com/goji/httpauth" do
+    url "https://github.com/goji/httpauth.git", :revision => "c1b2bcd8769bd15cc56751223fd4b9f45ca987ca"
+  end
+
+  go_resource "github.com/gorilla/websocket" do
+    url "https://github.com/gorilla/websocket.git", :revision => "361d4c0ffd78338ebe0a9e6320cdbe115d7dc026"
+  end
+
+  go_resource "github.com/juju/ratelimit" do
+    url "https://github.com/juju/ratelimit.git", :revision => "772f5c38e468398c4511514f4f6aa9a4185bc0a0"
+  end
+
+  go_resource "github.com/mitchellh/go-homedir" do
+    url "https://github.com/mitchellh/go-homedir.git", :revision => "d682a8f0cf139663a984ff12528da460ca963de9"
+  end
+
+  go_resource "github.com/rjeczalik/notify" do
+    url "https://github.com/rjeczalik/notify.git", :revision => "1869adb163fffce8fb5b8755379d1042cdb4c4f8"
+  end
+
+  go_resource "github.com/toqueteos/webbrowser" do
+    # v1.0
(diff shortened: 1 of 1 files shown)


cortesi:
See further discussion [here](https://github.com/cortesi/devd/issues/3). The previous submission came when the project was 3 days old, and was entirely understandably declined. We're now at our 3rd release , with 1700 stars on Github, so hopefully we've crossed the notability threshold. :)

Library/Formula/devd.rb
```diff
@@ -0,0 +1,132 @@
+require "language/go"
+
+class Devd < Formula
+  desc "devd: a local webserver for developers"
```
DomT4: Don't include the formula name at the start of the description. Happens automatically.

Library/Formula/devd.rb
```diff
@@ -0,0 +1,132 @@
...
+
+  go_resource "github.com/toqueteos/webbrowser" do
+    # v1.0
+    url "https://github.com/toqueteos/webbrowser.git", :revision => "21fc9f95c83442fd164094666f7cb4f9fdd56cd6"
+  end
+
+  go_resource "github.com/alecthomas/template" do
+    url "https://github.com/alecthomas/template.git", :revision => "b867cc6ab45cece8143cfcc6fc9c77cf3f2c23c0"
+  end
+  go_resource "github.com/alecthomas/units" do
```
DomT4: Please leave an empty line between the resources. Helps with readability, a bit.

Library/Formula/devd.rb
```diff
@@ -0,0 +1,132 @@
...
+  go_resource "gopkg.in/alecthomas/kingpin.v2" do
+    # v2.4.1
+    url "https://github.com/alecthomas/kingpin.git", :revision => "95529ad11b3c862a5b828a2142b9e50db579cf2c"
+  end
+
+  def install
+    ENV["GOOS"] = "darwin"
+    ENV["GOARCH"] = MacOS.prefer_64_bit? ? "amd64" : "386"
+    ENV["GOPATH"] = buildpath
+    ENV.prepend_create_path "PATH", buildpath/"bin"
```
DomT4: Just `ENV.prepend_path "PATH", buildpath/"bin"`. If there's not already a `$PATH` variable something's gone horribly wrong and the build will have likely failed before even getting to this point :smile_cat:. Shouldn't need to create one.

Library/Formula/devd.rb
```diff
@@ -0,0 +1,132 @@
...
+
+    mkdir_p buildpath/"src/github.com/cortesi/"
+    ln_sf buildpath, buildpath/"src/github.com/cortesi/devd"
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    system "go", "install", "github.com/GeertJohan/go.rice/rice"
+
+    # system "./build", "single"
```
DomT4: Why the commented out line here?

Library/Formula/devd.rb
```diff
@@ -0,0 +1,132 @@
...
+    system "rice", "embed-go"
+    cd "#{buildpath}/livereload" do
+      system "rice", "embed-go"
+    end
+    system "go", "build", "-o", "#{bin}/devd", "./cmd/devd"
+  end
+
+  test do
+    begin
+      io = IO.popen("#{bin}/devd /tmp")
```
DomT4: What is this creating in `/tmp`?

Library/Formula/devd.rb
kad: ok. will fix.
Do you prefer second commit or updating this one ?

Library/Formula/devd.rb
kad: my bad. missed in long list of dependencies :(, will fix.

Library/Formula/devd.rb
kad: make sense. will fix.

Library/Formula/devd.rb
kad: It mostly note for future packagers. in HEAD, build script is already changed, so something that I do in formula for 0.3 release, in future would be simplified with calling official build script. Will probably add comment about that.

Library/Formula/devd.rb
kad: it need directory to serve content from. practically docroot in terms of normal http servers.
I had two safe choices, "." or "/tmp". I thought /tmp would be preferable, but can change to ".", doesn't really matter, as test doesn't depend on content of served dir.

Library/Formula/devd.rb
kad: BTW, it doesn't work with just prepend_path:

```
HEAD is now at 95529ad Use gopkg import to avoid duplicate fetch.
==> go install github.com/GeertJohan/go.rice/rice
==> rice embed-go
Failed to execute: rice
/usr/local/Library/Homebrew/debrew.rb:11:in `raise'
BuildError: Failed executing: rice embed-go
1. raise
2. ignore
3. backtrace
4. irb
5. shell
Choose an action: 5
When you exit this shell, you will return to the menu.
bash-3.2$ echo $PATH
/usr/local/Library/ENV/4.3:/usr/local/opt/go/bin:/usr/local/opt/git/bin:/usr/bin:/bin:/usr/sbin:/sbin
bash-3.2$
```

to use ENV.prepend_path, directory specified as argument must exist.
So, there are two options:
1. move that line after rice tool installed, and change it to ENV.prepend_path
2. let ENV.prepend_create_path to create that directory and add it to PATH before rice tool installed, like it's now.

DomT4: > 1. move that line after rice tool installed, and change it to ENV.prepend_path

This is fine.

Library/Formula/devd.rb
```diff
@@ -0,0 +1,139 @@
...
+    ENV["GOOS"] = "darwin"
+    ENV["GOARCH"] = MacOS.prefer_64_bit? ? "amd64" : "386"
+    ENV["GOPATH"] = buildpath
+    ENV.prepend_create_path "PATH", buildpath/"bin"
+
+    mkdir_p buildpath/"src/github.com/cortesi/"
+    ln_sf buildpath, buildpath/"src/github.com/cortesi/devd"
+    Language::Go.stage_deps resources, buildpath/"src"
+
+    system "go", "install", "github.com/GeertJohan/go.rice/rice"
```
DomT4: Doesn't `go install` like this fetch stuff remotely rather than via our `go_resource`s?

Library/Formula/devd.rb
```diff
@@ -0,0 +1,139 @@
...
+    cd "#{buildpath}/livereload" do
+      system "rice", "embed-go"
+    end
+    system "go", "build", "-o", "#{bin}/devd", "./cmd/devd"
+    doc.install "README.md"
+  end
+
+  test do
+    begin
+      io = IO.popen("#{bin}/devd .")
```
DomT4: Use `#{testpath}` rather than `.` just to be explicit about it.

Library/Formula/devd.rb
kad: At least it is not like that in my test environment. If I remove something from go_resources, I'm getting following error:

```
==> go install -v -x github.com/GeertJohan/go.rice/rice
Last 15 lines from /Users/kad/Library/Logs/Homebrew/devd/01.go:
2015-11-16 11:06:43 +0200

go
install
-v
-x
github.com/GeertJohan/go.rice/rice

src/github.com/GeertJohan/go.rice/rice/append.go:15:2: cannot find package "github.com/daaku/go.zipexe" in any of:
	/usr/local/Cellar/go/1.5.1/libexec/src/github.com/daaku/go.zipexe (from $GOROOT)
	/private/tmp/devd20151116-4786-1f8uapp/devd-0.3/src/github.com/daaku/go.zipexe (from $GOPATH)
```

Library/Formula/devd.rb
kad: updated.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `devd 0.3 (new formula)` and please don't sign-off your own commits.

kad:
Thanks. will skip sign-off :) just habit from another projects.

Issue #43329, opened 2015-08-27
brew install --HEAD ghostscript fails


jerzyk:
command fails with an error:
```
make: *** No rule to make target `src/png.h', needed by `obj/xpspng.o'.  Stop.
make: *** Waiting for unfinished jobs....
```

I suppose this is a mixup in using system vs local libpng (autogen.sh finds system wide libpng):
```
checking for local zlib source... yes
checking for local png library source... no
checking for png_create_write_struct in -lpng... yes
checking png.h usability... yes
checking png.h presence... yes
```

Logs:
     /.../Library/Logs/Homebrew/ghostscript/01.autogen.sh
     /.../Library/Logs/Homebrew/ghostscript/01.autogen.sh.cc
     /.../Library/Logs/Homebrew/ghostscript/02.make
     /.../Library/Logs/Homebrew/ghostscript/02.make.cc
     /.../Library/Logs/Homebrew/ghostscript/config.log

uploaded to: https://gist.github.com/a0cfb6f3ab1bd4e3429b

dunn:
Good catch.  Letting it build the vendored libpng (instead of shunting it out of the way here https://github.com/Homebrew/homebrew/blob/master/Library/Formula/ghostscript.rb#L69) avoids this, but I'll see if there's a fix that allows us to keep using our libpng.

dunn:
So the short-term fix is `inreplace "xps/xps.mak", "$(PNGSRCDIR)$(D)png.h", ""`.

The long-term fix is getting the GS build process to officially allow system versions of libpng etc., instead of the vendored ones.  I don't know how long that will take, so I'm leaning towards letting it have its vendored deps instead of patching Makefiles.

dunn:
If we do go the patchy route, might as well do zlib too: https://github.com/Homebrew/homebrew/compare/Homebrew:master...dunn:ghostscript

And considering we're already pushing the vendored files around, this isn't much more invasive.

jerzyk:
but, there may be an issue when brewed version will not be supported (yet) by gs

dunn:
Forgot to close this via commit; should be fixed now!

PR #45979, opened 2015-11-14
ghostscript 9.18

dunn:
Replacing the vendored dependencies with our own was causing problems; see https://github.com/Homebrew/homebrew/issues/43329

### DIFF
--- a/Library/Formula/ghostscript.rb
+++ b/Library/Formula/ghostscript.rb
@@ -3,10 +3,13 @@ class Ghostscript < Formula
   homepage "http://www.ghostscript.com/"
 
   stable do
-    url "http://downloads.ghostscript.com/public/ghostscript-9.16.tar.gz"
-    sha256 "746d77280cca8afdd3d4c2c1389e332ed9b0605bd107bcaae1d761b061d1a68d"
+    url "http://downloads.ghostscript.com/public/ghostscript-9.18.tar.gz"
+    sha256 "5fc93079749a250be5404c465943850e3ed5ffbc0d5c07e10c7c5ee8afbbdb1b"
 
-    patch :DATA # Uncomment OS X-specific make vars
+    patch do
+      url "https://github.com/Homebrew/patches/raw/master/ghostscript/bug-696301_gserrors.h.patch"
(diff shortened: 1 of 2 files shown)


dunn:
Forgot to check revisions but looks like libspectre isn't compatible yet anyway: https://gist.github.com/dunn/e2bb64b63db6de7ec04f#file-02-make-L16-L30

dunn:
https://bugs.freedesktop.org/show_bug.cgi?id=92637

dunn:
Yeah, but it also needs patching.  Should we use the Gentoo one (https://gitweb.gentoo.org/repo/gentoo.git/tree/app-text/libspectre/files/libspectre-0.2.7-gs918.patch?id=8d34151aea08aa04cca35e1f5da59445af0440d0) or wait for upstream?

mikemcquaid:
Happy to use a patch that's been submitted upstream (as usual) :+1:

dunn:
OK, the change to using all the vendored deps is a separate commit now, so it can be reverted if it does cause trouble.

Library/Formula/ghostscript.rb
```diff
@@ -27,25 +22,16 @@ class Ghostscript < Formula
...
-  depends_on "jpeg"
-  depends_on "libtiff"
-  depends_on "jbig2dec"
-  depends_on "little-cms2"
-  depends_on "libpng"
-  depends_on :x11 => ["2.7.2", :optional]
   depends_on "djvulibre" if build.with? "djvu"
-  depends_on "freetype"
+  depends_on "little-cms2"
+  depends_on :x11 => :optional
```
dunn: The 2.7.2 requirement was from almost three years ago (#16924) and I'm guessing it's not necessary?

dunn:
Finally replicated that build failure locally. Maybe I still had GS 9.16 installed and that's where it was finding `gserrors.h`.

dunn:
Nevermind, even weirder: we're only installing that header when `--HEAD`.

dunn:
Aha http://git.ghostscript.com/?p=ghostpdl.git;a=commit;h=feafe5e540a0545ec5d28f3f66bb542056bba495

PR #45992, opened 2015-11-14
gobject-introspection: restore support for non-libtool builds

tschoonj:
pulls in upstream committed patch from https://bugzilla.gnome.org/show_bug.cgi?id=625195

I am working on creating a formula for webkitgtk and this requires gobject-introspection to be fixed with this patch...


DomT4:
> I am working on creating a formula for webkitgtk

I tried that a while back for something, if I'm remembering the name correctly. I wish you less endless frustration.

Merged in https://github.com/Homebrew/homebrew/commit/0ae2be4e9a7e644e85a532764530f3bdc242dbc2. Thanks Tom!

tschoonj:
Actually I got webkitgtk 2.10.4 compiling properly against gtk+3 with quartz backend. The only problem is that it doesn't actually work :cry: as I keep getting a segfault when running a basic example...

I filed a bugreport on the webkit bugzilla, hope they can help me...

Thanks for merging in the PR!

PR #46002, opened 2015-11-15
twoping: Install launchd plist

rfinnie:
2ping includes a listener component, so install a plist to allow for
starting the listener on startup.  The plist runs the listener as
"nobody" as the process requires no special permissions.


xu-cheng:

PR #46009, opened 2015-11-15
Update jags.rb to latest version 4.0.1

carsten-j:


DomT4:
The `revision 1` line needs removing as well, Thanks!

PR #46014, opened 2015-11-15
Add an option to build MC without NLS

arabesc:


arabesc:
I don't know why it still complains on whitespaces.

arabesc:
Currently I have russian localization in the MC by default on my system. But I would like to have english there because it's more laconic, there are no so many awkward reductions. And I like english as the main language in the tools.
Why are you asking? This change doesn't break anything and it's backward compatible.

mikemcquaid:
@arabesc We don't add options without understanding the reason why. I've added it though.

For future reference the preferred commit message format is `mc: add without-nls option.`.

PR #46019, opened 2015-11-15
update lnav to v0.8.0

tstack:


Library/Formula/lnav.rb
```diff
@@ -21,6 +21,7 @@ class Lnav < Formula
 
   depends_on "readline"
   depends_on "pcre"
+  depends_on "curl"
```
DomT4: Why can't this continue using the system `curl`?

tstack: I don't think the system curl has sftp support, which is useful in this context.

DomT4: You are correct, I believe, it doesn't. Our standard curl doesn't either though. You'd have to specify `curl => "with-libssh2`.

Unless it's feature critical I'd rather make it optional though and let people invoke it as desired.

PR #46021, opened 2015-11-15
fpc: Generate a default configuration file

### LINKED ISSUE #45783: freepascal: Missing unit folder in default search path
I noticed that fpc installed from homebrew does not add the default unit folder in its search path. For example compiling something that uses the GL unit (or any other unit) will return "unit GL not found" despite being compiled and installed.

The current solution is to add something like
``` 
    fpcversion = `fpc -iW`.chomp
    fpcpath = "#{Formula["fpc"].opt_lib}/fpc/"
    args << "-DCMAKE_Pascal_FLAGS=-Fu" + fpcpath + fpcversion + "/units/x86_64-darwin/*/"
```
however this is somewhat fragile and cumbersome when using outside homebrew.

Would it be possible to add the default search path (`/usr/local/lib/fpc/2.6.4/units/x86_64-darwin/` and its subfolders) while installing fpc?

If you need a quick sample feel free to use https://github.com/Homebrew/homebrew-games/blob/master/hedgewars.rb and remove the above line to see its effect.

kodabb:
The file provides a set of default paths and options, which help in
unit search paths (equivalent to C search paths) so that formulas won't
need to specify them when needed. All options and variables can be
overridden at any time on the command line.

Fix #45783.

references
http://www.freepascal.org/docs-html/3.0.0/user/usersu10.html
http://wiki.freepascal.org/Installing_Lazarus#STEP_.234:_Create_fpc.cfg_file

### DIFF
--- a/Library/Formula/fpc.rb
+++ b/Library/Formula/fpc.rb
@@ -28,11 +28,15 @@ class Fpc < Formula
 
     # Prevent non-executable audit warning
     rm_f Dir[bin/"*.rsj"]
+
+    # Generate a default fpc.cfg to set up unit search paths
+    system "#{bin}/fpcmkcfg", "-p", "-d", "basepath=#{lib}/fpc/#{version}", "-o", "#{prefix}/etc/fpc.cfg"
   end
 
   test do
     hello = <<-EOS.undent
       program Hello;
+      uses GL;
       begin
         writeln('Hello Homebrew')
       end.


Library/Formula/fpc.rb
```diff
@@ -25,11 +25,15 @@ def install
     system "make", "install", "PP=#{fpc_compiler}", "PREFIX=#{prefix}"
 
     bin.install_symlink lib/"#{name}/#{version}/ppcx64"
+
+    # generate a default fpc.cfg to facilitate search paths
+    system "./utils/fpcmkcfg/fpcmkcfg", "-p", "-d", "basepath=#{Formula["fpc"].opt_lib}/fpc/" + `./compiler/ppc1 -iW`.chomp, "-o", "#{Formula["fpc"].opt_prefix}/etc/fpc.cfg"
```
dunn: During installation we only allow writing within the `/usr/local/Cellar/formula/version/` prefix, so you'll need to use `Formula["fpc"].lib` and `Formula["fpc"].prefix` here, rather than the `opt` versions.

dunn: Actually since this _is_ the `fpc` formula, you can just use `lib` and `prefix`.

kodabb: I'm not sure I follow. To recap, in order to write fpc.cfg you need to find wherever the final fpc is going to be, then go above one level and then cd to etc. Since fpc could be placed in different places the basepath has to be adjusted every time, and you need to specify it at command line.

So, would this be the right approach? `system "./utils/fpcmkcfg/fpcmkcfg", "-p", "-d", "basepath=#{prefix}/lib/fpc/" + `./compiler/ppc1 -iW`.chomp, "-o", "#{prefix}/etc/fpc.cfg"`

dunn: Yeah, except `basepath=#{prefix}/lib/fpc/` can be `basepath=#{lib}/fpc/`.

kodabb:
updated :)

Library/Formula/fpc.rb
```diff
@@ -25,11 +25,15 @@ def install
     system "make", "install", "PP=#{fpc_compiler}", "PREFIX=#{prefix}"
 
     bin.install_symlink lib/"#{name}/#{version}/ppcx64"
+
+    # generate a default fpc.cfg to facilitate search paths
+    system "./utils/fpcmkcfg/fpcmkcfg", "-p", "-d", "basepath=#{lib}/fpc/" + `./compiler/ppc1 -iW.chomp`, "-o", "#{prefix}/etc/fpc.cfg"
```
mikemcquaid:
```ruby
`./compiler/ppc1 -iW.chomp`
```

should be

```
`./compiler/ppc1 -iW`.chomp
```

Also, please pull out the `ppc1` into a separate variable and use `#{variable}` instead of `+`. Thanks!

kodabb:
@mikemcquaid done

dunn:
The test is failing on 10.10 and 10.11. I wonder if they haven't updated it to handle these versions:
```
==> /usr/local/Cellar/fpc/2.6.4/bin/fpc hello.pas
Free Pascal Compiler version 2.6.4 [2015/11/17] for x86_64
Copyright (c) 1993-2014 by Florian Klaempfl and others
Error: Invalid value for MACOSX_DEPLOYMENT_TARGET environment variable: 10.10
```
```
==> /usr/local/Cellar/fpc/2.6.4/bin/fpc hello.pas
Free Pascal Compiler version 2.6.4 [2015/11/17] for x86_64
Copyright (c) 1993-2014 by Florian Klaempfl and others
Error: Invalid value for MACOSX_DEPLOYMENT_TARGET environment variable: 10.11

**Comment by kodabb** [comment]
@dunn I don't know, I develop on 10.9, but looks unrelated to my patch

**Comment by dunn** [comment]
Yeah, you might have to pester the developers about it :/

**Comment by kodabb** [comment]
Sorry I don't thinkg I can help you there, I don't have any mean to test it and I don't want to add too much stuff to a oneliner patch. Feel free to open a bug here http://bugs.freepascal.org/, but please don't stop this PR due to an unrelated bug.

**Comment by mikemcquaid** [comment]
It's not related an unrelated bug; it's meaning the formula doesn't work on the last two versions of OS X. You have an interest in the formula and as Homebrew is maintained by the community you're probably a good person to fix this. If you'd rather not: that's OK but this formula may be removed from Homebrew.

**Comment by kodabb** [comment]
Since the formula didn't get deleted, and I changed operating systems in the meantime, I updated the formula to work with fpc 3.0

File: Library/Formula/fpc.rb [hunk 3]
```diff
@@ -28,11 +28,18 @@ def install
 
     # Prevent non-executable audit warning
     rm_f Dir[bin/"*.rsj"]
+
+    # Generate a default fpc.cfg to facilitate search paths
+    fpcversion = `./compiler/ppc1 -iW`.chomp
```
DomT4: Will this generate the same output on every system? It won't change between Bob's Mac and Alice's Mac, if both are installed in `/usr/local`?

Library/Formula/fpc.rb
```diff
@@ -28,11 +28,18 @@ def install
...
+    fpccpu     = `./compiler/ppc1 -iTP`.chomp
+    fpcos      = `./compiler/ppc1 -iTO`.chomp
+    system "./utils/fpcmkcfg/bin/#{fpccpu}-#{fpcos}/fpcmkcfg", "-p", "-d", "basepath=#{lib}/fpc/#{fpcversion}", "-o", "#{prefix}/etc/fpc.cfg"
```
tomyun: I think these 4 lines can be squashed into `system #{bin}/fpcmkcfg", ...` since `fpcmkcfg` should be already built and installed in `bin` at this point.

kodabb: good tip, it'll address @DomT4 concerns too

Library/Formula/fpc.rb
kodabb: I think so, but @tomyun suggested a way I like more

kodabb:
formula updated

Library/Formula/fpc.rb
```diff
@@ -28,11 +28,16 @@ def install
 
     # Prevent non-executable audit warning
     rm_f Dir[bin/"*.rsj"]
+
+    # Generate a default fpc.cfg to facilitate search paths
+    fpcversion = `#{bin}/fpc -iW`.chomp
+    system "#{bin}/fpcmkcfg", "-p", "-d", "basepath=#{lib}/fpc/#{fpcversion}", "-o", "#{prefix}/etc/fpc.cfg"
```
tomyun: Probably we can just use `#{version}` instead of `#{fpcversion}` as they should match.

kodabb:
Updated with additional comments from @tomyun

kodabb:
all is good now?

Library/Formula/fpc.rb
```diff
@@ -28,11 +28,15 @@ def install
 
     # Prevent non-executable audit warning
     rm_f Dir[bin/"*.rsj"]
+
+    # Generate a default fpc.cfg to set up unit search paths
+    system "#{bin}/fpcmkcfg", "-p", "-d", "basepath=#{lib}/fpc/#{version}", "-o", "#{prefix}/etc/fpc.cfg"
```
mikemcquaid: You should use `#{etc}` here instead or generate it into `buildpath` and then use `etc.install`.

kodabb:
updated with @mikemcquaid's comment

Library/Formula/fpc.rb
```diff
@@ -28,11 +28,15 @@ def install
...
     rm_f Dir[bin/"*.rsj"]
+
+    # Generate a default fpc.cfg to set up unit search paths
+    system "#{bin}/fpcmkcfg", "-p", "-d", "basepath=#{lib}/fpc/#{version}", "-o", "#{etc}/fpc.cfg"
   end
 
   test do
     hello = <<-EOS.undent
       program Hello;
+      uses GL;
```
mikemcquaid: Why did you add this? Looks like it makes the test fail: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/41972/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/test_fpc/

kodabb: it's to make sure paths are correctly set up, an unused unit will be simply ignored

```
 $ brew test --verbose fpc
Testing fpc
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160218-35257-4zttbi.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Formula/fpc.rb --verbose
==> /usr/local/Cellar/fpc/3.0.0/bin/fpc hello.pas
Free Pascal Compiler version 3.0.0 [2016/02/15] for x86_64
Copyright (c) 1993-2015 by Florian Klaempfl and others
Target OS: Darwin for x86_64
Compiling hello.pas
Assembling (pipe) hello.s
Linking hello
5 lines compiled, 0.1 sec
```

mikemcquaid: Ok. The `test do` will need something added to it to setup the paths I guess.

kodabb: I don't think so, the whole point of this additional step is that fpc will read #{etc}/fpc.cfg and won't need any additional path

in fact it works for me locally (the output i posted is from my shell), and I'm quite sure I saw a green merged button a few revisions ago.

Maybe the use of #{etc} is not appropriate in this case, and instead it should use #{prefix}/etc?

mikemcquaid: Do you have environment variables set that we need to in the test, perhaps?

kodabb: umh I don't think it needs any

http://www.freepascal.org/docs-html/user/usersu10.html

mikemcquaid: Well, we can't merge this until the test is passing on the bot.

kodabb: ok i see the problem, #{etc} is the wrong directory for this file: it is installed in /usr/local/etc/ but the binary is just a symlink elsewhere, so even if the compiler is in /usr/local/bin/, it can't find the configuration file in the directory above.

It needs to be #{prefix}/etc, which normally expands to /usr/local/Cellar/fpc/3.0.0/etc, I'll update the PR shortly

mikemcquaid: No, sorry, `#{prefix}/etc` is the wrong location, we'll need to fix this up so the configuration file is written into `/usr/local/etc` as otherwise it'll be destroyed on reinstall and upgrades (which sucks for configuration files).

kodabb: i consider that a feature, it *needs* to be destroyed so that new defaults can be installed
any user configuring custom defaults can just add a fpc.cfg to their home directory

mikemcquaid: Ok, so this file should never be modified by users? If so: `prefix` does seem reasonable after all.

kodabb: That is my understanding. According to the following snipped (from the url above) users have at least two additional paths for a customized cfg file, before the default one is loaded. The binary installer also does something like this, installs a fpc.cfg to ../etc (wrt the fpc binary) and then lets users create a custom cfg for their needs.

```
Unless you specify the -n (see page 96) option, the compiler will look for a configuration file fpc.cfg in the following places:
    Under unix (such as linux)
        The current directory.
        Your home directory, it looks for .fpc.cfg.
        The directory specified in the environment variable PPC_CONFIG_PATH, and if it is not set, it will look in the etc directory above the compiler directory. (For instance, if the compiler is in /usr/local/bin, it will look in /usr/local/etc)
        The directory /etc.
```

mikemcquaid:

kodabb:
thanks for your patience and insightful reviews

PR #45914, opened 2015-11-12
use secure protocol for sourceforge.net homepages

vszakats:
* it will be redirected by SourceForge to cleartext unless logged in
* follow some URL redirects
* update some moved homepages

PR #46026, opened 2015-11-16
librevenge: use secure homepage, fix test failure

vszakats:


dunn:
What are the revision bumps for?

DomT4:
Weird Boost CI issues, if I remember correctly from the other PR.

vszakats:
Ref: https://github.com/Homebrew/homebrew/pull/45914#issuecomment-156412624

DomT4:
Looks like it's still failing. Weird. Have kind of a busy day today but I'll try and take a look locally this PM.

vszakats:
It didn't resolve the problem, just opened this one to continue where the issue was abandoned in another [PR](https://github.com/Homebrew/homebrew/pull/45914).

vszakats:
👍🏻

DomT4:
Oh. Kind of an obvious failure that I should have spotted really. All the failing tests seem to be looking for `Boost` because it is present at compile-time, but because it is marked as a `:build` dep we remove it postinstall to check `:build` deps aren't mislabelled, and that causes the test to fail on looking for Boost.

It looks like Boost became runtime-expected here, at some point, probably because we're passing ` ENV["LDFLAGS"] = "-lboost_system-mt"`, I suspect.

vszakats:
@DomT4 You were spot on &mdash; after applying your fix, it passes without any problem.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/8191b84a0f233d26e9aeb3f88685bfd4e1229939. Thanks @vszakats!

PR #46031, opened 2015-11-16
sphinx-doc 1.3.1 (new formula)

xu-cheng:


xu-cheng:
Should I make `sphinx-doc` key_only to avoid conflicting with user's installation from `pip`(especially consider they may want to use sphinx with custom plugins).

/cc @tdsmith @DomT4

DomT4:
> Should I make sphinx-doc key_only to avoid conflicting with user's installation from pip

Tim's the expert but I would, aye. I suspect it's common enough that we'd see Issues submitted to us otherwise.

I'd be tempted to add `sphinx-doc` to the `audit` build dep check as well, much like `libtool` et al.

xu-cheng:
Updated to tag it as keg_only and add audit rule.

PR #46036, opened 2015-11-16
openal-soft 1.17.0

ojab:


Library/Formula/openal-soft.rb
```diff
@@ -5,13 +5,8 @@ class OpenalSoft < Formula
   revision 1
```
xu-cheng: You need to remove `revision 1`. And it would be great if you can remove redundant stable block to following form. Thanks.
```ruby
desc ".."
homepage ".."
url ".."
sha256 ".."
head ".."
```

Library/Formula/openal-soft.rb
```diff
@@ -40,10 +31,7 @@ class OpenalSoft < Formula
   def install
     ENV.universal_binary if build.universal?
 
-    # Please don't reenable example building. See:
```
DomT4: I'd be surprised if this wasn't still broken. Please do:

```
brew install sdl2 sdl_sound
```
And then install this build locally. Let us know if that passes.

ojab: Already tried, builds fine here (El Capitan).

ojab: Hmm, looks like my test was somehow broken, got build error this time. 
Rolled back this change.

xu-cheng:

PR #46037, opened 2015-11-16
improve brew info

xu-cheng:
* Show pinned and keg-only as attributes, where pinned
     version is shown as well.
* Use brackets instead of parentheses to distinguish formula
     attributes(e.g. keg-only) with spec attributes(e.g. bottled)
* Don't show blank line for formula without homepage.
* Don't show duplicated keg-only message which will be shown later
     in caveats.
* Underline urls.
* Remove unnecessary github_info return value check, which is always
  existed.

### DIFF
--- a/Library/Homebrew/cmd/info.rb
+++ b/Library/Homebrew/cmd/info.rb
@@ -105,18 +105,13 @@ module Homebrew
 
     specs << "HEAD" if f.head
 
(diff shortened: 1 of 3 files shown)


mikemcquaid:
Can you list some sample output? Thanks!

xu-cheng:
> Can you list some sample output? Thanks!

Here you are(Left: before, Right: after)

mikemcquaid:
Looks good :+1:

PR #46038, opened 2015-11-16
strongswan 5.3.4

tobiasbrunner:


dunn:
Mind adding a test while you're at it? We'd be very grateful!

tobiasbrunner:
Not that easy to add a "good" test, we can't really test real VPN connections after all (would also require running the executables as root). I added a simple test for some executables that should have been built (running them with `--version`). Another thing I added is calling `make check` before `make install`, which runs the unit tests.

PR #46042, opened 2015-11-16
Add python3 installation option to z3

cHoco:


Library/Formula/z3.rb
```diff
@@ -6,6 +6,9 @@ class Z3 < Formula
   head "https://github.com/Z3Prover/z3.git"
   revision 1
 
+  depends_on :python => :recommended
```
DomT4: If you're going to make this recommended it needs to use the system Python by default.

```ruby
depends_on :python => :recommended if MacOS.version <= :snow_leopard
```

mikemcquaid:
`brew test` is failing, can you take a look?

cHoco:
Does the 'Language::Python.each_python' iterator include system python?
It was failing because the the iteration inside that do block was never executed. Is moving it out the right solution?

DomT4:
> Does the 'Language::Python.each_python' iterator include system python?

Should do. You can see `hashpump` which [uses the system](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/hashpump.rb#L34-L36) Python by default.

You appear to be replicating the install block above your formula changes here; I think you should just be able to replace it?

cHoco:
Mmmm, there's something really strange happening here.

If I add 
```  
option "without-python", "Build without python 2 support"
```
to the formula then the block includes system python, otherwise it doesn't.
Don't know if it is a bug just wanted to point it out. 
Either way I added to the formula since it makes sense in this case. 

Also I made python3 recommended since I think it makes sense to install python bindings for all versions on system, but I'll change it back if you prefer to make it optional.
Finally I'm not familiar with bottles and formulas, so by adding python3 and making it recommended does the bootle need to be regenerated? If yes, do I have to change something?

Thank you for assisting me with this, it's really appreciated.

DomT4:
> to the formula then the block includes system python, otherwise it doesn't.

We seem to have temporarily lost Tim who is our Python guru, but I believe he has said previously this is correct.

> Also I made python3 recommended since I think it makes sense to install python bindings for all versions on system, but I'll change it back if you prefer to make it optional.

It should *probably* be optional otherwise people will have to install our `Python3` formula, which is kind of a heavy dep to get installed automatically for optional extra bindings.

> If yes, do I have to change something?

Bottles regen automatically. No contributor participation required there :).

cHoco:
> It should probably be optional otherwise people will have to install our Python3 formula, which is kind of a heavy dep to get installed automatically for optional extra bindings.

Yeah that's what I thought :)
I changed it back to optional, should be good to go now I hope! 
Again always a pleasure contributing to homebrew :D Great work!

DomT4:
Seems fine to me. Merged in https://github.com/Homebrew/homebrew/commit/6ad6376dea5d32164e61d5245f6320605fee6b08 :smiley:.

Thank you for another Homebrew contribution & the positive attitude @cHoco; people like you make maintaining a pleasure.

PR #46043, opened 2015-11-16
move more deprecated methods to compat folder

xu-cheng:


mikemcquaid:
`Stdenv` should maybe be marked both private and deprecated. It is deprecated.

mikemcquaid:
Otherwise :+1:

xu-cheng:
> Stdenv should maybe be marked both private and deprecated. It is deprecated.

Isn't still using heavily in `test`?

mikemcquaid:
Yeh, it should be deprecated for non-test usage.

xu-cheng:
Currently it's only used in test while it's tagged as private. I don't think we need to worry the wrong use case. Also, it feels weird to me to deprecate codes we are still actively using.

DomT4:
>  Also, it feels weird to me to deprecate codes we are still actively using.

```ruby
# deprecated... soon... hopefully
# private
```
:wink:

xu-cheng:
Thanks for advice. However, in current context, we are unlikely to deprecate `Stdenv`. If I recall correctly, we deliberately want to test the formula in `Stdenv` to best simulate user env.

DomT4:
Well, from Mike's wording above, `stdenv` is already deprecated in *most* aspects?

Given all the problems with `stdenv` I'm happy enough seeing it tagged as deprecated with the eventual idea that we iron out enough `superenv` bugs to have everything using that, even if `stdenv` sticks around exclusively for testing formulae.

xu-cheng:
> the eventual idea that we iron out enough superenv bugs to have everything using that, even if stdenv sticks around exclusively for testing formulae.

This is already the case now. I don't see there is any problem in current `Stdenv/Superenv` use case. 

On the contrary, tagging it as deprecate can be quite misleading as it feels like we will completely remove `Stdenv` code eventually. As you can see in this PR, I'm moving all of *deprecate* codes into `compat` folder and offer option to use Homebrew without loading any deprecate code. 

Overall, what I want to express is deprecate has unique semantically meaning. And we shouldn't missive it.

mikemcquaid:
> On the contrary, tagging it as deprecate can be quite misleading as it feels like we will completely remove Stdenv code eventually.

We should probably do that. Instead of using `Stdenv` we should be just shelling out in test.

DomT4:
> This is already the case now.

Well, apart from the formula we've allowed to keep `:std`? There's some pretty major exceptions in there.

xu-cheng:
> We should probably do that. Instead of using Stdenv we should be just shelling out in test.

Good idea. I think we can still keep Stdenv but remove all compiling related codes. so Stdenv will be a basic utility extend from ruby's `ENV`.

Added the note to reflect current discussion in PR. Let me know whether there is anything need to be changed before merging it.

mikemcquaid:
> Good idea. I think we can still keep Stdenv but remove all compiling related codes. so Stdenv will be a basic utility extend from ruby's ENV.

I'd rather introduce a new class and change the name, honestly.

xu-cheng:
> I'd rather introduce a new class and change the name, honestly.

Anyhow that should belong to future PR.

mikemcquaid:
:ok: :+1:

PR #46059, opened 2015-11-17
update consul-template to 0.11.1

grepnull:
updated to newest version of consul-template. The build scripts have
changed which required some changes to the formula.

### DIFF
--- a/Library/Formula/consul-template.rb
+++ b/Library/Formula/consul-template.rb
@@ -4,8 +4,8 @@ class ConsulTemplate < Formula
   desc "Generic template rendering and notifications with Consul"
   homepage "https://github.com/hashicorp/consul-template"
   url "https://github.com/hashicorp/consul-template.git",
-      :tag => "v0.10.0",
-      :revision => "14cf110504f3089a3c96a36126bd6d5fe39df37a"
+      :tag => "v0.11.1",
+      :revision => "134aecc73e607f09986370600232ecb8f44d0940"
 
   head "https://github.com/hashicorp/consul-template.git"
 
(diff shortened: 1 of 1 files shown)


grepnull:
Let me know if there's anything else you need here. I'd really like to get this merged! Thanks!

DomT4:
Why the need for so many less dependencies? Is there still no remote fetching happening during compile?

grepnull:
The build script is grabbing all the dependencies it needs, it just needs gox to bootstrap.

DomT4:
Is that a yes or a no on remote fetching not specified in the formula file, for clarity?

grepnull:
Sorry, but I guess I don't really understand your question.

grepnull:
I can re-add the resources if that helps. I simply don't really understand why they'd be needed or what you mean by "remote fetching happening during compile". I've only done a few homebrew updates and just getting my feet wet with Go as well.

If you could help me out with this merge, I'd really appreciate it. Thanks!

dunn:
The Makefile is running `go get`: https://github.com/hashicorp/consul-template/blob/master/scripts/build.sh#L23-L24

We might have to run the `gox` command manually.

grepnull:
@dunn, I don't think that's necessary. `make bin` is calling that `build.sh` script and so it is grabbing all the compile dependencies it needs.

I've been using this locally for the past week with no problem. The tests all pass in Travis here (and there's an actual real, albeit simple, check in the formula that uses the built binary).

I also tested by running this PR's template: 
```
brew install https://raw.githubusercontent.com/grepnull/homebrew/consul-template-0.11.1/Library/Formula/consul-template.rb
``` 
on another mac of mine that never even had Go installed and it installed Go and then built fine and passed the test.

As I understand Go, these are compile-time dependencies, not run-time dependencies so they shouldn't be needed to use the built binary.

If I'm mistaken in any of this, please let me know.

Thanks again.

dunn:
We don't like letting build scripts download dependencies because it's a security risk (we can't verify the checksum) and we can't cache them, so if a build fails the user has to download them all again.

grepnull:
Got it. I can put the dependencies back then, call `gox` and update the PR.

dunn:
Thanks so much!

grepnull:
Hopefully good to merge now... let me know if there are any other changes you want. Thanks!

dunn:
Looks great, thanks so much! Merged in e648ed1b1a717a1e9dc8d85c5f4e533c656bf5d2.

PR #46063, opened 2015-11-17
pgloader fix checksums

ojab:


ojab:
oh, well. Checksums for freshly downloaded pkgs are changed back to the previous values :/

PR #46067, opened 2015-11-17
shairport 1.1.1 (import from head-only)

bfontaine:
See https://github.com/Homebrew/homebrew-head-only/pull/184


bfontaine:
The failure is due to the name conflict with the head-only formula.

bfontaine:

PR #46068, opened 2015-11-17
tcpstat 0.1 (import from head-only)

bfontaine:
See https://github.com/Homebrew/homebrew-head-only/pull/185

### DIFF
--- a/Library/Formula/tcpstat.rb
+++ b/Library/Formula/tcpstat.rb
@@ -1,0 +1,25 @@
+class Tcpstat < Formula
+  desc "Active TCP connections monitoring tool"
+  homepage "https://github.com/jtt/tcpstat"
+  url "https://github.com/jtt/tcpstat/archive/rel-0-1.tar.gz"
+  version "0.1"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

Library/Formula/tcpstat.rb
```diff
@@ -0,0 +1,25 @@
...
+  homepage "https://github.com/jtt/tcpstat"
+  url "https://github.com/jtt/tcpstat/archive/rel-0-1.tar.gz"
+  version "0.1"
+  sha256 "366a221950759015378775862a7499aaf727a3a9de67b15463b0991c2362fdaf"
+
+  head "https://github.com/jtt/tcpstat.git"
+
+  def install
+    system "make"
+    bin.install "tcpstat"
```
mikemcquaid: The usual `make install` question :grinning:

bfontaine:

Library/Formula/tcpstat.rb
bfontaine: Yep tried it but it doesn’t work (uses a wrong `install` option).

PR #46072, opened 2015-11-17
unar 1.9.1

santazhang:
Resubmitting formula for unar 1.9.1 done in pull request #44314.

I think #44314 keeps failing because the CI servers failed to cleanup generated files?

### DIFF
--- a/Library/Formula/unar.rb
+++ b/Library/Formula/unar.rb
@@ -1,11 +1,11 @@
 class Unar < Formula
   desc "RAR archive command-line tools"
   homepage "http://unarchiver.c3.cx/commandline"
-  url "https://theunarchiver.googlecode.com/files/unar1.8.1_src.zip"
-  version "1.8.1"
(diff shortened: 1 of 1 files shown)


DomT4:
I'm kind of tempted to pull this either without bottles or without a Mavericks bottle. Our Mavericks CI apparently has some issues with the build that don't reproduce for people locally or on Travis :confounded:. Thoughts here on the bottle situation @mikemcquaid @xu-cheng?

chdiza:
Perhaps some of the problem is that the formula, ever since it landed in HB way back when, is full of unneeded complication.  That library is just for building `unar/lsar` with, it's not to be installed on a user's machine.

mikemcquaid:
:+1: to pulling without a Mavericks bottle.

DomT4:
@mikemcquaid Just for fun, this works perfectly fine if you `brew install unar -s` on our Mavericks VM :confused:.

Merged in https://github.com/Homebrew/homebrew/commit/771c66cd5a6d746cf02700d2ad53948a0ccb6ded. Thank you for your contribution to Homebrew @santazhang; we appreciate it! :smiley_cat:

PR #46073, opened 2015-11-17
Update libcurl pkgconfig file for 10.10

MoSal:
This matches the version shipped with 10.10.5.

 The list of supported features/protocols was obtained
 using `curl-config`:

``` bash
 features="`curl-config --features`"; echo "${features//$'\n'/ }"
 protocols="`curl-config --protocols`"; echo "${protocols//$'\n'/ }"
```

 `Libs.private` line removed as it only concerns static linking.

Signed-off-by: Mohammad AlSaleh <CE.Mohammad.AlSaleh@gmail.com>

### DIFF
--- a/Library/ENV/pkgconfig/10.10/libcurl.pc
+++ b/Library/ENV/pkgconfig/10.10/libcurl.pc
@@ -27,13 +27,13 @@ prefix=/usr
 exec_prefix=${prefix}
 libdir=${exec_prefix}/lib
 includedir=${prefix}/include
-supported_protocols="DICT FILE FTP FTPS GOPHER HTTP HTTPS IMAP IMAPS LDAP LDAPS POP3 POP3S RTSP SMTP SMTPS TELNET TFTP"
-supported_features="SSL IPv6 libz AsynchDNS NTLM NTLM_WB GSS-API"
+supported_protocols="DICT FILE FTP FTPS GOPHER HTTP HTTPS IMAP IMAPS LDAP LDAPS POP3 POP3S RTSP SMB SMBS SMTP SMTPS TELNET TFTP"
+supported_features="SSL IPv6 UnixSockets libz AsynchDNS GSS-API SPNEGO Kerberos NTLM NTLM_WB"
 
 Name: libcurl
 URL: http://curl.haxx.se/
 Description: Library to transfer files with ftp, http, etc.
-Version: 7.30.0
+Version: 7.43.0
 Libs: -L${libdir} -lcurl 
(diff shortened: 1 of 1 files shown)


mikemcquaid:
CC @DomT4

DomT4:
> Libs.private line removed as it only concerns static linking.

I'd kind of rather keep this line. The `pkg-config` files we supply are intended (IMO) to match as-close-to-upstream as possible if Apple actually installed the generated `pkg-config` files.

MoSal:
@DomT4 
I'm wondering why do you think a generated `.pc` file would include that line.
The output of `curl-config --static-libs` is empty.

I can add it back if you insist.

DomT4:
@MoSal Check the Homebrewed `curl` package:

```
prefix=/usr/local/Cellar/curl/7.45.0
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
supported_protocols="DICT FILE FTP FTPS GOPHER HTTP HTTPS IMAP IMAPS LDAP LDAPS POP3 POP3S RTSP SCP SFTP SMB SMBS SMTP SMTPS TELNET TFTP"
supported_features="SSL IPv6 UnixSockets libz AsynchDNS IDN GSS-API SPNEGO Kerberos NTLM NTLM_WB TLS-SRP HTTP2"

Name: libcurl
URL: http://curl.haxx.se/
Description: Library to transfer files with ftp, http, etc.
Version: 7.45.0
Libs: -L${libdir} -lcurl
Libs.private: -lcares -lnghttp2 -lidn -lssh2 -lssh2 -lssl -lcrypto -lssl -lcrypto -lgssapi_krb5 -lresolv -lldap -lz
Cflags: -I${includedir}
```

MoSal:
@DomT4 
Installing the bottle. I have:
```
Libs.private: -lldap -lz
```

And `curl-config --static-libs` output:

```
/usr/local/Cellar/curl/7.45.0/lib/libcurl.a -framework CoreFoundation -framework Security -lldap -lz
```

DomT4:
Yes. It all rather depends on how you build `curl`. As you can tell, I use a rather custom and feature-heavy `curl`.

I'd like us to keep the line, personally.

MoSal:
@DomT4 
Okay.
I added `Libs.private` back.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/48f69a96274dc11b569ed83e10514c6b5d4a1ed0. Thank you @MoSal!

PR #46076, opened 2015-11-17
zzz 1 (imported from head-only)

bfontaine:
See https://github.com/Homebrew/homebrew-head-only/pull/188

### DIFF
--- a/Library/Formula/zzz.rb
+++ b/Library/Formula/zzz.rb
@@ -1,0 +1,20 @@
+class Zzz < Formula
+  desc "Command-line tool to put Macs to sleep"
+  homepage "https://github.com/Orc/Zzz"
+  url "https://github.com/Orc/Zzz/archive/v1.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/zzz.rb
```diff
@@ -0,0 +1,14 @@
+class Zzz < Formula
+  desc "Command-line tool to put Macs to sleep"
+  homepage "https://github.com/Orc/Zzz"
+  url "https://github.com/Orc/Zzz/archive/v1.tar.gz"
+  sha256 "8c8958b65a74ab1081ce1a950af6d360166828bdb383d71cc8fe37ddb1702576"
+
+  head "https://github.com/Orc/Zzz.git"
+
+  # No test is possible: this has no --help or --version, it just
+  # sleeps the Mac instantly.
```
mikemcquaid: Would be good to even have a `File.exist?` so it passes `brew audit --strict`

bfontaine: Done. I also added `bottle :unneeded` because it takes only 2 seconds to compile.

PR #46080, opened 2015-11-17
simavr: intitial formula

jedahan:
simavr is super useful, when you want to program an arduino on a plane or in the subway

### DIFF
--- a/Library/Formula/simavr.rb
+++ b/Library/Formula/simavr.rb
@@ -1,0 +1,24 @@
+class Simavr < Formula
+  desc "a lean, mean and hackable AVR simulator for linux & OSX "
+  homepage "https://github.com/buserror/simavr"
+  url "https://github.com/buserror/simavr/archive/v1.2.tar.gz"
+  sha256 "c7f645b323d74247409b50ea6915f1f324a7ee2a799ef107613626801defc070"
+
+  depends_on "osx-cross/avr/avr-gcc"
+  depends_on "libelf"
+
+  head do
+    url "https://github.com/buserror/simavr.git"
+
+    patch :p1 do
+      url "https://patch-diff.githubusercontent.com/raw/buserror/simavr/pull/147.diff"
+      sha256 "b2322ef7d1aee1f7564cc1cb6fcb92079680fda2ab7b4f0c3c74b6454649c982"
+    end
+  end
+
+  def install
+    ENV.append_path "PATH", "#{ENV["HOMEBREW_PREFIX"]}/bin"
+    system "make", "install", "DESTDIR=#{prefix}", "RELEASE=1"
+    prefix.install "examples"
+  end
+end


Library/Formula/simavr.rb
```diff
@@ -0,0 +1,24 @@
+class Simavr < Formula
+  desc "a lean, mean and hackable AVR simulator for linux & OSX "
+  homepage "https://github.com/buserror/simavr"
+  url "https://github.com/buserror/simavr/archive/v1.2.tar.gz"
+  sha256 "c7f645b323d74247409b50ea6915f1f324a7ee2a799ef107613626801defc070"
+
+  depends_on "osx-cross/avr/avr-gcc"
```
DomT4: Can't have mandatory cross-tap dependencies in the core I'm afraid. This would either need to be vendored or live outside the core.

Library/Formula/simavr.rb
```diff
@@ -0,0 +1,24 @@
...
+    url "https://github.com/buserror/simavr.git"
+
+    patch :p1 do
+      url "https://patch-diff.githubusercontent.com/raw/buserror/simavr/pull/147.diff"
+      sha256 "b2322ef7d1aee1f7564cc1cb6fcb92079680fda2ab7b4f0c3c74b6454649c982"
+    end
+  end
+
+  def install
+    ENV.append_path "PATH", "#{ENV["HOMEBREW_PREFIX"]}/bin"
```
DomT4: Please don't manually re-add `HOMEBREW_PREFIX/"bin"` back to the `$PATH` manually. We keep it excluded for good reason.

Library/Formula/simavr.rb
jedahan: What does vendored mean?

Library/Formula/simavr.rb
jedahan: So if a package's Makefile calls brew, are we supposed to patch the Makefile in the formula? I can do that, if that's what is recommended.

Library/Formula/simavr.rb
DomT4:
```ruby
resource "example" do
  sha256 "13550350a8681c84c861aac2e5b440161c2b33a3e4f302ac680ca5b686de48de"
end

def install
  resource("example").stage do
    system "./buildscript", "--prefix=#{libexec}/example", "--other-options"
    system "make", "install"
  end

  ENV.prepend_path "PATH", libexec/"example/bin" # Presuming you're installing executables
  the rest of your normal installation here.
end
```

DomT4: You essentially build the dependency as a separate object rather than a dependency and if you need it at runtime install it into the formula's Cellar. Like an `.app` bundle to some extent.

Library/Formula/simavr.rb
DomT4: What is it calling `brew` to find out?

jedahan: In a [patch I submitted](https://github.com/buserror/simavr/pull/147/files) (which might totally be terrible), I have the makefile call `brew --prefix` to find out where to link against avr-libc and libelf. I can try changing that to be `${HOMEBREW_PREFIX:-/usr/local/}` and see if that works...

DomT4: Ah. Yeah, that's tricky to handle in a way. An ideal Makefile would allow people to pass variables, such as:

`make PREFIX="path" LIBELF="path" AVR_LIBC="path"`

Only falling back on automatic checks if those aren't passed. That's probably the best route to go down in terms of patching.

Library/Formula/simavr.rb
jedahan: Hmm that seems messy, maybe this should be submitted to https://github.com/osx-cross/homebrew-avr ?

DomT4: You might well find that an easier path, or hosting this in a tap yourself, aye. The latter is easier than it probably sounds; tap creation and maintenance is a fairly pleasant process. There's some documentation [here](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md) if you're interested in that option.

Library/Formula/simavr.rb
jedahan: Added a HOMEBREW_PREFIX option, since the Makefile is already 'build system aware'. It would be better and more flexible to use LIBELF and AVR_LIBC, but i'll see what the maintainer wants first. Thanks for clarifying everything!

jedahan:
Submitted to homebrew-avr in https://github.com/osx-cross/homebrew-avr/pull/12 , as that tap is required anyway

PR #46081, opened 2015-11-17
adding shml formula

jmervine:
Hello,

I'm (re)submitting [shml v1.0.4](https://github.com/MaxCDN/shml/tree/1.0.4). This is my first brew formula, so please let me know if you need me to change anything.

Cheers!
J

(cc: @jdorfman)

### DIFF
--- a/Library/Formula/shml.rb
+++ b/Library/Formula/shml.rb
@@ -6,7 +6,7 @@ class Shml < Formula
 
   def install
     bin.install "shml.sh"
-    ln_s "#{bin}/shml.sh", "#{bin}/shml"
+    bin.install_symlink "foo.sh" => "foo"
   end
 
   test do
--- a/Library/Formula/shml.rb
+++ b/Library/Formula/shml.rb
@@ -10,7 +10,10 @@ class Shml < Formula
   end
 
   test do
-    assert_match "1.0.4", pipe_output("#{bin}/shml", "-v")
-    assert_match "1.0.4", pipe_output("#{bin}/shml.sh", "-v")
+    ["shml", "shml.sh"].each do |cmd|
+      result = shell_output("#{bin}/#{cmd} -v")
+      result.force_encoding("UTF-8") if result.respond_to?(:force_encoding)
+      assert_match "1.0.4", result
+    end
   end
 end
--- a/Library/Formula/shml.rb
+++ b/Library/Formula/shml.rb
@@ -1,0 +1,16 @@
+class Shml < Formula
+  desc "Style Framework for The Terminal"
+  homepage "http://shml.xyz"
+  url "https://github.com/MaxCDN/shml/archive/1.0.4.tar.gz"
+  sha256 "583fefbbf2adf45d989b052696be0f20f695903878558a4d2e9f74255b91ed51"
+
+  def install
+    bin.install "shml.sh"
+    ln_s "#{bin}/shml.sh", "#{bin}/shml"
+  end
+
+  test do
+    assert_match "1.0.4", pipe_output("#{bin}/shml", "-v")
+    assert_match "1.0.4", pipe_output("#{bin}/shml.sh", "-v")
+  end
+end
--- a/Library/Formula/shml.rb
+++ b/Library/Formula/shml.rb
@@ -6,7 +6,7 @@ class Shml < Formula
 
   def install
     bin.install "shml.sh"
-    bin.install_symlink "foo.sh" => "foo"
+    bin.install_symlink bin/"shml.sh" => "shml"
   end
 
   test do


jdorfman:
:+1:

Library/Formula/shml.rb
```diff
@@ -0,0 +1,16 @@
+class Shml < Formula
+  desc "Style Framework for The Terminal"
+  homepage "http://shml.xyz"
+  url "https://github.com/MaxCDN/shml/archive/1.0.4.tar.gz"
+  sha256 "583fefbbf2adf45d989b052696be0f20f695903878558a4d2e9f74255b91ed51"
+
+  def install
+    bin.install "shml.sh"
```
DomT4: Is there any harm in installing the binary directly as `shml` without the `.sh`?

jmervine: No, not really. I do want both executables available as typical use-case is to source this executable, and the more common pattern is `source shml.sh`. That said, I can either install both or install it as `shml` and then generate a symlink for `shml.sh`, if that's preferred.

I couldn't find any clear documentation on using `bin.install` and/or `bin.install_symlink`. I would assume that `bin.install "shml.sh" => "shml"` works?

jdorfman: @DomT4 not at all.  The NPM install symlinks it e.g.: 
`/usr/local/bin/shml -> /usr/local/lib/node_modules/shml/shml.sh`

jmervine:
@DomT4 How would I reference a binary that's is in the repo root? `bin.install` is looking for "bin/<exec>", would `install` and `install_symlink` alone work?

Library/Formula/shml.rb
DomT4: > I would assume that bin.install "shml.sh" => "shml" works?

Yup.

> I do want both executables available as typical use-case is to source this executable, and the more common pattern is source shml.sh.

In this case perhaps:

```ruby
libexec.install "shml.sh"
bin.install_symlink libexec/"shml.sh" => "shml"
```
:question:

jmervine: Yes, it currently does. That's actually something I was going to fix in the next release... supporting both `shml` and `shml.sh`.

jmervine: It's not a hard fast requirement. More of a personal preference for usability. I can certainly omit `shml.sh` if it's a Homebrew deal-break. :)

Library/Formula/shml.rb
```diff
@@ -0,0 +1,16 @@
+class Shml < Formula
+  desc "Style Framework for The Terminal"
+  homepage "http://shml.xyz"
+  url "https://github.com/MaxCDN/shml/archive/1.0.4.tar.gz"
+  sha256 "583fefbbf2adf45d989b052696be0f20f695903878558a4d2e9f74255b91ed51"
+
+  def install
+    install "shml.sh"
```
DomT4: If you want to leave it in the prefix you're looking for `prefix.install` here. `install` by itself won't do anything, I believe.

Library/Formula/shml.rb
```diff
@@ -0,0 +1,16 @@
...
+    install "shml.sh" => "shml"
```
DomT4: Likewise if this needs to be executed sometimes rather than just sourced you'll want this symlink in the `bin` I presume?

Library/Formula/shml.rb
DomT4: >  I can certainly omit shml.sh if it's a Homebrew deal-breaker. :)

Not at all. Having both is fine. Just wondered what the set up you were after was :)

jmervine: Oh, that's what `libexec` is.... okay... thanks.

DomT4: Some people like to put scripts that need sourcing into the `prefix`, others prefer `libexec`. I have no strong preference either way. If it needs to live in the `bin` as well you'll want to create a symlink there.

The syntax would be:
```ruby
libexec.install "shml.sh"
bin.install_symlink libexec/"shml.sh" => "shml"
```
And then replace `libexec` with `prefix` in that example, if you so wish to.

Library/Formula/shml.rb
jmervine: I don't know why I'm having so much trouble with visualizing how this works. So, if I'm understanding you correctly, if I want to install the file (without symlinking it) as both `shml` and `shml.sh` it should be...

```ruby
libexec.install "shml.sh" => bin/"shml.sh"
libexec.install "shml.sh" => bin/"shml"
```

DomT4: You can't install the same file twice, because `.install` is essentially our helper for the Ruby version of `mv`.

The easiest way to deal with that sort of thing is to install the original somewhere, like `libexec.install "shml.sh"` and then symlink it. In shell terms:

```ruby
libexec.install "shml.sh"
```
would be more or less:
```bash
mv buildpath/shml.sh HOMEBREW_CELLAR/shml/version/libexec
```

And:
```ruby
bin.install_symlink libexec/"shml.sh" => "shml"
```
would be more or less:
```bash
ln -s HOMEBREW_CELLAR/shml/version/libexec/shml.sh HOMEBREW_CELLAR/shml/version/bin/shml
```

jmervine: Okay, thank you for the direction. I'll give that a shot.

Library/Formula/shml.rb
jmervine: @DomT4 After some mucking around locally, this is what I'm looking for (assuming CI finally passes): https://github.com/jmervine/homebrew/commit/8baaa57961ff8b029d5cdae6c9236d40a49f9b62. 

I want both `shml.sh` and `shml` to be executable and to be able to be sourced in a subsequent bash script. 

Thanks again for your help.

DomT4: That would work, yes. Having two almost-identically named executables in the top-level `bin` feels unnecessary, but if no other maintainer objects I won't either.

DomT4:
New error:

```
==> /usr/local/Cellar/shml/1.0.4/bin/shml
Error: shml: failed
invalid byte sequence in US-ASCII
```
Can probably be solved by adding either:
```ruby
ENV["LC_ALL"] = "en_US.UTF-8"
```
To the top of `test do` or if that fails something like [this](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/abduco.rb#L20-L23).

Library/Formula/shml.rb
jmervine: I understand the point on two almost identical bins. But this is really something that is sourced as a framework, more than executed. It needs to be in the path to properly source within bash/zsh/etc. Recommended use is `source shml`, but as more common practice is to source `.sh` files. As such we really prefer to support both.

mikemcquaid: Can it be run as an executable instead of being sourced?

jmervine: Yes. At this time though it's main purpose of being executed is to provide usage and examples. However, in future releases we're expecting the executable functionality to expand.

mikemcquaid: Ok, thanks!

mikemcquaid:

For future reference the preferred commit message format for new formulae is `shml 1.0.4 (new formula)`.

PR #46091, opened 2015-11-17
subversion: Fix build with PERL_MM_OPT set

felixbuenemann:
Without this change, the build aborts in Makefile.pl with the following error:
> Only one of PREFIX or INSTALL_BASE can be given. Not both.

PERL_MB_OPT and PERL_MM_OPT usually get set from the shell's profile, if the user configured the CPAN shell to install modules to the users HOME.

For details see the similar issue #45091.

### DIFF
--- a/Library/Formula/subversion.rb
+++ b/Library/Formula/subversion.rb
@@ -75,6 +75,11 @@ class Subversion < Formula
       inreplace "SConstruct", "unique=1", "unique=0"
 
       ENV.universal_binary if build.universal?
+
+      # Fix perl bindings Makefile.pl failing with:
+      # Only one of PREFIX or INSTALL_BASE can be given.  Not both.
+      ENV.delete "PERL_MM_OPT"
+
       # scons ignores our compiler and flags unless explicitly passed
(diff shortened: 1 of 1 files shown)


felixbuenemann:
The failure on Travis CI with OSX=10.10 seems to be due to a vulnerability in the shipped apache version:
> configure: error: Apache httpd 2.4.10 MMN 20120211.36 has broken auth (CVE-2015-3184)

DomT4:
Travis needs to update the Xcode/CLT on Yosemite, apparently.

jabenninghoff:
According to the Apache [security advisory](https://subversion.apache.org/security/CVE-2015-3184-advisory.txt), the fix requires httpd to be upgraded to 2.4.16. Mavericks OSX=10.9 doesn't use a vulnerable version of httpd (2.4.0 - 2.4.12) so it builds fine.

DomT4:
Yeah, the problem is Yosemite on Travis, which hasn't had the latest CLT/Xcode update applied there yet.

Merged in https://github.com/Homebrew/homebrew/commit/adaf83b57042c734a490a9cf2049bdf222b35af5. Thank you for another Homebrew contribution @felixbuenemann! :smiley_cat:

PR #46093, opened 2015-11-17
git: Fix git-svn --with-brewed-subversion lib path

felixbuenemann:
On newer systems perl installs modules into the standard `lib/perl5/site_perl` hierarchy, instead of the Mac OS X specific `Library/Perl`. This fix makes sure that both old and new directories are considered.

Without this fix `git-svn` is broken on Mac OS X 10.11.2 Beta (15C47a) with the system built-in `/usr/bin/perl` 5.18.2.

I'm not sure which perl version switched the directory structure, so I'm keeping the old lib path.

### DIFF
--- a/Library/Formula/git.rb
+++ b/Library/Formula/git.rb
@@ -53,7 +53,10 @@ class Git < Formula
     perl_version = /\d\.\d+/.match(`perl --version`)
 
     if build.with? "brewed-svn"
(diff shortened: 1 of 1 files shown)


felixbuenemann:
Should I bump the revision for this fix?

mikemcquaid:

PR #46094, opened 2015-11-17
add formula for iMatix GSL

paddor:
Just a new formula for @imatix's GSL https://github.com/imatix/gsl.

This makes it possible to install it using `brew install --HEAD imatix-gsl`.


DomT4:
We don't accept `head-only` formulae in the core I'm afraid. They'd need to have a release to be accepted here.

paddor:
I've just created a Homebrew tap for zproject and GSL: [paddor/homebrew-zproject](https://github.com/paddor/homebrew-zproject)

@sappo @jemc @hintjens I can transfer ownership of this repository to the ZMQ organisation, if you want.

PR #46097, opened 2015-11-17
gst-python: link against brew python if available

cbenhagen:


DomT4:
Not going to do this by default; It'll force everyone to install Homebrew's Python to use this formula. You could make it `:optional`, if you really want to and @tdsmith doesn't feel upset by it.

Library/Formula/gst-python.rb
```diff
@@ -3,6 +3,7 @@ class GstPython < Formula
   homepage "http://gstreamer.freedesktop.org/modules/gst-python.html"
   url "http://gstreamer.freedesktop.org/src/gst-python/gst-python-1.6.1.tar.xz"
   sha256 "3cbe332e18fd2eaf23ddeee96c414f79ac1edc2f7d990582fa0ec5f977bd04f1"
+  revision 1
```
DomT4: Now the change below is optional, you don't need the revision bump.

cbenhagen:
Oh sorry that was not the intended behaviour. Fixed and rebased it.

cbenhagen:
https://travis-ci.org/Homebrew/homebrew/jobs/91688277#L1178 does not seem related to this PR. But anyway. Does someone have an idea why this happens on travis but not on my systems running 10.10 and 10.11?

@tdsmith Can you live with that optional dependency? Or is there a better way for a user to build gst-python against Homebrew's Python?

bfontaine:
@cbenhagen Install with the `--sandbox` option to reproduce the error.

cbenhagen:
I am not sure how to fix this sandbox issue but i will add an issue to track that.

What do you think about disabling the bottle and dropping the `:optional`? This way it always gets built for the first python in PATH.

cbenhagen:
@DomT4 i rebased this your recent commit 4b498186bd0c1d2759d6ebfb19eaa52dab84b743. So as expected it does build again.

Any thoughts about how we could manage this to work for Homebrew and System Python out of the box? Are there alternatives to disabling the bottle? Otherwise it would just silently build against system Python if the user does pass the --with-python option. Which is not expected and will eventually break in strange ways.

tdsmith:
Dumb question, since I don't understand this stack very well: what is gst-python and what does it do? Does it let you call python from gstreamer or does it let you control gstreamer from external Python code or both or neither?

cbenhagen:
Its complementing the GStreamer Python bindings provided by python-gi. So it lets you control GStreamer from Python.

DomT4:
> What do you think about disabling the bottle and dropping the :optional? 

This isn't going to happen I'm afraid. Our aim is to bottle everything in Homebrew that requires compile, and that includes this. The solution in this PR would seem to be the best one for the use case you're requesting.

cbenhagen:
But how can we prevent that the user gets a corrupt system by accident then?

Maybe we should have two bottles then? One for system and one for Homebrew Python?

mikemcquaid:
> Maybe we should have two bottles then? One for system and one for Homebrew Python?

We're not doing that, unfortunately, bottles are only built against the default options.

cbenhagen:
So what can we do? If dropping the bottle is no option.. what about a way to force a build --with-python when a homebrewed Python is installed? I feel bad for other users who run into issues because of this.

mikemcquaid:
Can you elaborate on your issues/

tdsmith:
Yeah, it'd be interesting to see an example of something that fails.

cbenhagen:
This might not be the best example. I will try to come up with a simpler one later.

In my case i was using PyInstaller to package a simple demo app which is using GStreamer and gst-python. So when gst-python is linked against system python the resulting app crashes. While it works as expected when gst-python was linked against the homebrewed python.

cbenhagen:
Somehow i am having a hard time to trigger this issue without PyInstaller now. But using python modules which are built against system python together with ones that are not just screams for issues no?

tdsmith:
It's probably a little more complicated. Looking at the files distributed
with gst-python, the importable module in site-packages is not linked to
any python framework (and does not need to be). The gstreamer plugin is
linked to the system python; it needs to have an explicit link to a python
framework in order to embed the interpreter. I don't know how those are
conflicting in practice or how PyInstaller is contributing. If you can
share the PyInstaller bundle I'd be curious to take a look.

cbenhagen:
@tdsmith https://www.dropbox.com/s/i7elddpyx4rlcxb/testapp.zip?dl=1

This is not cleaned up and does include a Kivy gui as the simpler GStreamer only example does not work anymore for some other reason. The only difference between the two app bundles is that for one i had gst-python installed `--with-python`


This is the traceback from the one linked against the wrong Python version:

```
  File "/usr/local/bin/../Cellar/python/2.7.10_2/bin/../Frameworks/Python.framework/Versions/2.7/lib/python2.7/site.py", line 548, in <module>
    main()
  File "/usr/local/bin/../Cellar/python/2.7.10_2/bin/../Frameworks/Python.framework/Versions/2.7/lib/python2.7/site.py", line 530, in main
    known_paths = addusersitepackages(known_paths)
  File "/usr/local/bin/../Cellar/python/2.7.10_2/bin/../Frameworks/Python.framework/Versions/2.7/lib/python2.7/site.py", line 266, in addusersitepackages
    user_site = getusersitepackages()
  File "/usr/local/bin/../Cellar/python/2.7.10_2/bin/../Frameworks/Python.framework/Versions/2.7/lib/python2.7/site.py", line 241, in getusersitepackages
    user_base = getuserbase() # this will also set USER_BASE
  File "/usr/local/bin/../Cellar/python/2.7.10_2/bin/../Frameworks/Python.framework/Versions/2.7/lib/python2.7/site.py", line 231, in getuserbase
    USER_BASE = get_config_var('userbase')
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/sysconfig.py", line 520, in get_config_var
    return get_config_vars().get(name)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/sysconfig.py", line 453, in get_config_vars
    import re
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/re.py", line 108, in <module>
    import _locale
SystemError: dynamic module not initialized properly
```

The packaged app looks like this:

```Python
#!/usr/bin/env python
import time
from kivy.app import App
from kivy.core.audio import SoundLoader
from kivy.uix.button import Button

def play(*args):
    sound = SoundLoader.load('http://upload.wikimedia.org/wikipedia/commons/c/c8/Example.ogg')
    if sound:
        print("Sound found at %s" % sound.source)
        print("Sound is %.3f seconds" % sound.length)
        sound.play()

class TestApp(App):
    def build(self):
        return Button(text='Hello OGG World', on_release=play)

TestApp().run()
```

tdsmith:
Here's the different binding:

```diff
 app_name.app/Contents/MacOS/gst_plugins/libgstpythonplugin.so:
 	@loader_path/../libgstpythonplugin.so (compatibility version 0.0.0, current version 0.0.0)
-	@loader_path/../Python (compatibility version 2.7.0, current version 2.7.0)
+	/System/Library/Frameworks/Python.framework/Versions/2.7/Python (compatibility version 2.7.0, current version 2.7.10)
 	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1225.1.1)
 	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1255.1.0)
 	@loader_path/../libgstreamer-1.0.0.dylib (compatibility version 602.0.0, current version 602.0.0)
```

MikeMcQuaid:
Passing on this as-is but @tdsmith will address this in some form.

PR #46099, opened 2015-11-17
codemod 20151117 (import from head-only)

bfontaine:
See https://github.com/Homebrew/homebrew-head-only/pull/189


kastiglione:
:+1:

PR #46108, opened 2015-11-18
Update to Maven 3.3.9

dadoonet:
I also changed `mirror` as it gives for now a `404` error but it's may be only because it takes some time to sync?

### DIFF
--- a/Library/Formula/maven.rb
+++ b/Library/Formula/maven.rb
@@ -1,9 +1,9 @@
 class Maven < Formula
   desc "Java-based project management"
   homepage "https://maven.apache.org/"
-  url "https://www.apache.org/dyn/closer.cgi?path=maven/maven-3/3.3.3/binaries/apache-maven-3.3.3-bin.tar.gz"
-  mirror "https://archive.apache.org/dist/maven/maven-3/3.3.3/binaries/apache-maven-3.3.3-bin.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
Apologies, I merged #46113 because I'm working from newest to oldest through open PRs rather than vice versa. Will try to double check next time; we have a policy of trying to merge whoever filed the PR first. Thank you for the submission to Homebrew nonetheless @dadoonet.

dadoonet:
The funny thing is that @jasontedor and me are working in the same team :p

DomT4:
You should probably feel feel to berate him for ignoring a team-mate then :wink:

jasontedor:
Apologies @dadoonet, I did not see this.

@DomT4 Out of curiosity, I was under the impression from #25962 that `brew search maven` would me tell if there was an open pull request for the Maven formula. I ran that this morning before opening #46113 and it didn't show any open pull requests?

DomT4:
> Out of curiosity, I was under the impression from #25962 that brew search maven would me tell if there was an open pull request for the Maven formula. 

It seems to only do so if `brew search` doesn't already return a result:

```
~> brew search getdns
getdns
```

```
~> brew search getdns1
No formula found for "getdns1".
==> Searching pull requests...
```

jasontedor:
> It seems to only do so if brew search doesn't already return a result

Thanks. I will keep that in mind in the future.

Issue #46083, opened 2015-11-17
Kibana 4.2.1


jmagnusson:
Wanted to create a PR for an updated version of Kibana, but ran into some trouble as they seem to have changed their build task (there is no longer any `dist` directory and they are no longer using bower for example). The changes I made to the formula can be seen here:

```diff
diff --git a/Library/Formula/kibana.rb b/Library/Formula/kibana.rb
index 44a71e0..e715c06 100644
--- a/Library/Formula/kibana.rb
+++ b/Library/Formula/kibana.rb
@@ -1,8 +1,8 @@
 class Kibana < Formula
   desc "Visualization tool for elasticsearch"
   homepage "https://www.elastic.co/products/kibana"
-  url "https://github.com/elastic/kibana/archive/v4.1.1.tar.gz"
-  sha256 "3f91e99e20e82d4e84ec141007822fea8f9454c71595551f9348ea2609c98284"
+  url "https://github.com/elastic/kibana/archive/v4.2.1.tar.gz"
+  sha256 "f69eb7d4e0447222a9e383b95562b384212e9070361400d8d722eda278ade56f"
   head "https://github.com/elastic/kibana.git"

   bottle do
@@ -20,15 +20,14 @@ class Kibana < Formula

     system "npm", "install"
     system "npm", "install", "grunt-cli"
-    system "npm", "install", "bower"
-    system "./node_modules/.bin/bower", "install"
     system "./node_modules/.bin/grunt", "build", "--force"

-    dist_dir = buildpath/"build/dist/kibana"
+    dist_dir = buildpath/"build/kibana"

     rm_f dist_dir/"bin/*.bat"

     prefix.install dist_dir/"src"
     (etc/"kibana").mkpath
     (var/"lib/kibana/plugins").mkpath
```

It installs successfully, but when I try to run `/usr/local/bin/kibana` I get the following error:

```bash
module.js:339
    throw err;
    ^

Error: Cannot find module 'lodash'
    at Function.Module._resolveFilename (module.js:337:15)
    at Function.Module._load (module.js:287:25)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/usr/local/Cellar/kibana/4.2.1/src/optimize/babelOptions.js:3:17)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
[...]

PR #46124, opened 2015-11-18
Kibana 4.2.1

jmagnusson:
WIP. Can’t seem to find node modules when running kibana after install.
See #46083 for more info.


DomT4:
Once we've resolved whatever we're going to do with #46139 I'll take a look at this for you.

afinne:
Newest Kibana would be nice to get. I don't know if it just me, or if the version of elasticsearch you get with brew is not compatible with the version of kibana...

jmagnusson:
Closed in favor of #46124.

PR #46127, opened 2015-11-18
urbano 0.3

iarenzana:
Command-line Urban Dictionary.

### DIFF
--- a/Library/Formula/urbano.rb
+++ b/Library/Formula/urbano.rb
@@ -1,0 +1,41 @@
+require "language/go"
+
+class Urbano < Formula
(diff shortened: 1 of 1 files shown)


dunn:
```
==> audit problems
urbano:
 * Stable: version 0.3 is redundant with version scanned from URL
 * Description should use "command-line" instead of "command line"
 * Description shouldn't include the formula name
```
We might sit on this for a little bit, just to make sure it has some traction.  Accepting new software adds to our maintenance burden so we want to make sure we don't end up with broken or abandoned software in a few months.

mikemcquaid:
Passing on this for now due to reasons above.

PR #46137, opened 2015-11-18
choose 0.1.0 (import from head-only)

bfontaine:
Regarding the test the `script -q /dev/null` trick doesn’t work nor any `expect` script I’ve tried (I asked a [question on SO](http://stackoverflow.com/q/33791936/735926) about that).

See https://github.com/Homebrew/homebrew-head-only/pull/192

### DIFF
--- a/Library/Formula/choose.rb
+++ b/Library/Formula/choose.rb
@@ -1,0 +1,32 @@
+class Choose < Formula
+  desc "Make choices on the command-line"
+  homepage "https://github.com/geier/choose"
+  url "https://github.com/geier/choose/archive/v0.1.0.tar.gz"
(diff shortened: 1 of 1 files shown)


dunn:
It's awful but it works:
```ruby
  test do
    assert_match /^homebrew/, shell_output("echo homebrew | script -q /dev/null #{bin}/choose")
  end

bfontaine:
@dunn Interestingly your solution works but not `pipe_output("script -q /dev/null #{bin}/choose", "homebrew")`.

bfontaine:
Thinking about that I the difference is that `shell_output` does run in a shell while `pipe_output` doesn’t.

bfontaine:
It stills timeout on Mavericks & Yosemite :/

bfontaine:
I reverted to the existing test for now.

PR #46139, opened 2015-11-18
node: stage npm directly to libexec

### LINKED ISSUE #46131: npm: `npm outdated -g` errors - specific to Homebrew npm installation
```shell
$ npm outdated -g

$ npm outdated
```

ericvw:
`npm`'s installation uses `npm`, itself, to install `npm` at a
particular version.  In doing so, it also makes the implicit assumption
that the tarball is unpackaged at `$PREFIX/lib/node_modules/npm`.
However, Homebrew conducts its builds under `/tmp`, which `npm` will
record incorrect paths relative to where the installation of `npm` was
executed in `package.json` for caching.

This change stages the `npm` tarball directly under `libexec` without
needeing to execute `make install`, since `npm` can function completely
when untar'd to its desired location.

Closes #46131

### DIFF
--- a/Library/Formula/node.rb
+++ b/Library/Formula/node.rb
@@ -63,21 +63,11 @@ class Node < Formula
     system "make", "install"
 
     if build.with? "npm"
-      resource("npm").stage buildpath/"npm_install"
-
-      # make sure npm can find node
-      ENV.prepend_path "PATH", bin
-      # set log level temporarily for npm's `make install`
-      ENV["NPM_CONFIG_LOGLEVEL"] = "verbose"
-
-      cd buildpath/"npm_install" do
-        system "./configure", "--prefix=#{libexec}/npm"
-        system "make", "install"
-      end
+      resource("npm").stage "#{libexec}/npm"
 
       if build.with? "completion"
         bash_completion.install \
-          buildpath/"npm_install/lib/utils/completion.sh" => "npm"
+          "#{libexec}/npm/lib/utils/completion.sh" => "npm"
       end
     end
   end
@@ -91,7 +81,7 @@ class Node < Formula
     # Kill npm but preserve all other modules across node updates/upgrades.
     rm_rf node_modules/"npm"
 
-    cp_r libexec/"npm/lib/node_modules/npm", node_modules
+    cp_r libexec/"npm", node_modules
     # This symlink doesn't hop into homebrew_prefix/bin automatically so
     # remove it and make our own. This is a small consequence of our bottle
     # npm make install workaround. All other installs **do** symlink to
@@ -105,7 +95,7 @@ class Node < Formula
       # Dirs must exist first: https://github.com/Homebrew/homebrew/issues/35969
       mkdir_p HOMEBREW_PREFIX/"share/man/#{man}"
       rm_f Dir[HOMEBREW_PREFIX/"share/man/#{man}/{npm.,npm-,npmrc.}*"]
-      ln_sf Dir[libexec/"npm/lib/node_modules/npm/man/#{man}/npm*"], HOMEBREW_PREFIX/"share/man/#{man}"
+      ln_sf Dir[libexec/"npm/man/#{man}/npm*"], HOMEBREW_PREFIX/"share/man/#{man}"
     end
 
     npm_root = node_modules/"npm"


ericvw:
cc @DomT4

DomT4:
I'll double check a few things locally, but provisionally if it works I don't have a problem with the change.

ericvw:
No worries.  Happy to make any adjustments necessary.

DomT4:
Line 98 of your version of the Node formula here needs changing to match the new location. Currently breaks `postinstall`:

```
~> brew postinstall node -v
Error: unknown file type: /usr/local/Cellar/node/5.1.0/libexec/npm/lib/node_modules/npm
```

It's also breaking some of our other Node using formulae, which will need to be addressed in a way that doesn't involve dumping the main `HOMEBREW_PREFIX/bin` back into the `$PATH` for those compiles. I'm not entirely sure why the error here is being thrown, but we'll need to investigate it:

```
==> Verifying skinny-2.0.0.tar.gz checksum
tar xf /usr/local/var/homebrew/cache/skinny-2.0.0.tar.gz
==> npm install yo
module.js:340
    throw err;
    ^

Error: Cannot find module '/usr/local/opt/node/libexec/npm/bin/node_modules/npm/bin/npm-cli.js'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:289:25)
    at Function.Module.runMain (module.js:457:10)
    at startup (node.js:138:18)
    at node.js:974:3
```

ericvw:
@DomT4, I am a bit confused.  It seems like Travis CI and build checks appear OK.  I am unable to replicate:
```shell
$ brew postinstall node -v
$
```

Can you walk me through exactly how you are producing the error you are seeing?

ericvw:
Interesting, no matter what path I change line 98 to, I don't receive an error.  Is this due a difference in environment?

ericvw:
It appears that `FileUtils.ln_sf` does not error when there is an invalid entry for the src list - I am using Ruby version `ruby 2.0.0p645 (2015-04-13 revision 50299) [universal.x86_64-darwin15]`.

```shell
$ touch foo.txt
$ mkdir onelevel
$ irb
irb(main):001:0> require 'fileutils'
=> true
irb(main):002:0> FileUtils.ln_sf(Dir['/tmp/bogus.txt'], './onelevel')
=> []
irb(main):003:0> FileUtils.ln_sf(Dir['/tmp/foo.txt'], './onelevel')
=> ["/tmp/foo.txt"]
irb(main):004:0>^D
$ $ ls -l onelevel/
total 4
lrwxr-xr-x 1 eric wheel 12 Nov 19 11:43 foo.txt -> /tmp/foo.txt
```

DomT4:
Apologies. That was my late night screw up. I forgot to merge this locally before reinstalling `node` via the new bottle generated via the CI, so it threw an error because we were trying to symlink something that no longer existed. The line referenced does need updating in this PR though.

Also still need to work out why other Homebrew formula can no longer build against our `npm` here. That was tested with a from-source build after this PR was merged. That problem seems to be happening because the `npm` executable in `libexec/npm/bin/npm` reads:

```
NPM_CLI_JS="$basedir/node_modules/npm/bin/npm-cli.js"
```
And the `basedir` when `npm` is just copied rather than installed in this way is `libexec/npm/bin`.

ericvw:
I agree and will update the line 89 for the correct path.  However, I am concerned that despite it being wrong, that `brew postinstall -v node` gives no indication of error.

How can I go about testing other Homebrew formulae that builds against `npm`?  Happy to help out where I can.

ericvw:
Pushed an amended commit to address the symlink'd `man` path.

DomT4:
>  However, I am concerned that despite it being wrong, that brew postinstall -v node gives no indication of error.

Because we're throwing out to FileUtils there there's not a lot we can do, unless you want to raise the issue with Ruby themselves than trying to symlink something that doesn't exist should probably error. If something throws an error code we'll pick up on it I believe, but presumably that operation isn't.

> How can I go about testing other Homebrew formulae that builds against npm?

`brew install`
* skinny
* tegh
* autocode
* kibana

I've only tested `skinny` so far because it's a reasonable download and the `npm` operation is the first thing it does, but I expect them all to fail identically because of the `npm` script language in the `bin` without install.

ericvw:
Closing in favor of #46186.

PR #46142, opened 2015-11-19
theora: use secure homepage, add head

vszakats:


Library/Formula/theora.rb
```diff
@@ -1,8 +1,9 @@
 class Theora < Formula
   desc "Open video compression format"
-  homepage "http://www.theora.org/"
+  homepage "https://www.theora.org/"
   url "http://downloads.xiph.org/releases/theora/libtheora-1.1.1.tar.bz2"
   sha256 "b6ae1ee2fa3d42ac489287d3ec34c5885730b1296f0801ae577a35193d3affbc"
+  head "https://git.xiph.org/theora.git"
```
DomT4: If you want to add a `head` line you'll need to put it in a `head do` block with autotool deps and then add a `system "./autogen.sh if build.head?` line to `def install`. There's no pre-generated `configure` script in the upstream repo.

vszakats: Thank you! Will do!

vszakats:
Done! (I hope I got it right.)

DomT4:
Looks perfect! Thanks for the quick follow-up!

Merged in https://github.com/Homebrew/homebrew/commit/fdeb375ecb535228f31a0a17b0934c3b369539f3.

PR #46146, opened 2015-11-19
autocode: use secure url in test

vszakats:


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/80c2ff1dbac44e62e42557ca12252f73eeef21e3. Your talent for spotting even minor insecure URLs is both impressive and slightly terrifying :wink:. Thanks!

vszakats:
`grep` is to blame :)

DomT4:
I'm a fan of `ack` personally for mass huntings. Extremely handy tool.

vszakats:
Yeah, tried to give `ack` a chance but the muscle memory didn't moved on just yet. It is planned along some other mid-term retooling.

vszakats:
I'm going to give `ack` another go. It's so much faster and works without much manual tweaking.

DomT4:
Despite being targeted towards programmers it's useful pretty widely. I used it the other day to find a quote I was looking for in my study notes, spread out across a bunch of word documents, where I couldn't remember which file it was in.

vszakats:
I hope to remember this next time I need it. [ In theory, Spotlight should be able to do that :) El Capitan does search somewhat better than previous iterations. ]

DomT4:
I use "Alfred" as well, but `ack` has proved pretty useful for finding snippets of text. Don't ever really use the Spotlight in terms of GUI; use `mdfind` from the CLI a lot though.

vszakats:
`mdfind` noted, thanks! I use Spotlight only as an app starter really.

PR #46147, opened 2015-11-19
curlpp: use secure url in test

vszakats:


DomT4:
These PRs give me a nice excuse to pull new bottles to replace SHA1 and/or generate El Cap ones, heh.

This went in as https://github.com/Homebrew/homebrew/commit/9bd24ab6bccad0d8375bad81f8442f4a47d301a8 :star2:

vszakats:
Moving to sha256 bottles is indeed a nice "fall-out" of these.

DomT4:
It's part of a wider move permanently away from SHA1 across the ecosystem.

vszakats:
Your collective efforts are much appreciated, including this move. Homebrew is an exemplary project.

Just for the fun of it, do you have any consideration already for the next secure hash &mdash; beyond `sha256`? Maybe `blake2` or `sha3`?

DomT4:
My personal end goal, providing there's full team consent and the ecosystem hops on board the train, is by the end of 2017 to have eliminated SHA1 support and to provide optional GPG verification of upstream tarballs where they're signed and perhaps more importantly, our bottles.

> Just for the fun of it, do you have any consideration already for the next secure hash — beyond SHA256? 

We're more or less stuck to tracking the Digest module of Ruby, to avoid having to shell out to external utilities which is time and resource heavy (relatively speaking). I think BLAKE2 is a really interesting (and quick) hash, but I suspect SHA3 will become more of an industry standard and it'll be adopted widely sooner, so that's likely the direction we'll go in.

Schneier is still big on SHA2 last time I heard, and obviously Google et al. have started pushing SHA256 recently as well for cryptographic signing, etc, so hopefully it remains secure for a long time to come and if we decide to migrate over to a newer hash at some point it can be done on a nice steady deadline.

DomT4:
Er, end of 2016, even. Dom forgets what year it is.

vszakats:
:) - GPG totally escaped me, it'd be the perfect solution for these purposes, albeit not trivial to implement. As for SHA3 vs. BLAKE2, that's my sentiment as well. Thanks for the insights to Homebrew's plans. And I seriously hope SHA256 will hold up for a number of years.

DomT4:
GPG is kind of a pain to work with, but at the same time it hasn't been widely replaced by anything yet. There's things like `signify` and `minisign` out there but people actually using those are even smaller a niche than us GPG users are.

Fingers crossed about SHA256. I'd like to get to a point where we can make these changes on a reasonable timescale rather than so suddenly.

vszakats:
I did actually implement `minisign`-like signing to replace the heavy OpenSSL dependency (in private apps). The new NaCl crypto is real compact and fast, too &mdash; really good to work with. The pain with key-pairs is still there, but that's a given. It already helps that you can work with short key _strings_ instead of relatively large key _files_.

Me too, though having a reserve/future algo is not a bad idea at any given time, if for nothing else, to keep all the Homebrew logic for multiple hash support in an up-to-date state.

DomT4:
Yeah I use `minisign` personally from time to time, and the OS X port of `signify` (Although less since `minisign` became available since it's compatible I believe). The crypto behind both is great, and certainly relatively readable. If adoption takes off they could both be useful as alternatives to the fairly heavy GPG, but people tend to stick to what they know.

PR #46153, opened 2015-11-19
fix minor nits around Google URLs

vszakats:


DomT4:
Merged "All the things". Thanks again @vszakats!

Issue #46114, opened 2015-11-18
MySQL 5.7.9 installation note hidden


sander3:
Hi,

During the post-install of MySQL 5.7.9 a temporary password is generated for root@localhost. But because Brew doesn't show this temporary password you won't be able to login.

renatomoya:
I'm having the same issue. Just installed MySQL 5.7.9 and I can't log in as root because I don't know where the password is.

DomT4:
Welp. You can get around this for now by removing the existing database in `HOMEBREW_PREFIX/var/mysql` generated by `postinstall` (Providing one didn't already exist - We don't overwrite existing databases), and then run `brew postinstall mysql -v` to see the output.

Will look at something that doesn't require user intervention.

sander3:
cat ~/Library/Logs/Homebrew/mysql/01.mysqld maybe?

DomT4:
If you're installing a bottle that would work. If building from source it'd be a lot further down the log number.

renatomoya:
`cat ~/Library/Logs/Homebrew/mysql/01.mysqld` works.

I re-initialized mysql server like this:

```bash
rm -rf /usr/local/var/mysql
mysqld --initialize-insecure
```

This initializes mysql server with root account with an empty password.

sander3:
Yeah I did that too. But it doesn't solve the problem that the user isn't notified about the user credentials.

This should do the trick:
```
  def post_install
    # Make sure the datadir exists
    datadir.mkpath
    unless (datadir/"mysql/user.frm").exist?
      ENV["TMPDIR"] = nil
      system bin/"mysqld", "--initialize-insecure", "--user=#{ENV["USER"]}",
        "--basedir=#{prefix}", "--datadir=#{datadir}", "--tmpdir=/tmp"
    end
  end

  def caveats; <<-EOS.undent
    A "/etc/my.cnf" from another install may interfere with a Homebrew-built
    server starting up correctly.
    To connect:
        mysql -u root
    Please don't forget to change the root user's password.
    EOS
  end
```

DomT4:
I'd rather not routinely leave people's `mysql` databases unprotected unless there's absolutely no other way to avoid it.

sander3:
You could run ```mysql_secure_installation```.

DomT4:
Discussing a workaround with the team. If nobody on the team objects I'll push it to the formula later today.

mikemcquaid:
[...]

PR #46159, opened 2015-11-19
mysql: default to no root password (like 5.6).

mikemcquaid:
Point the users to the relevant command so they can secure it more if they wish.

CC @xu-cheng https://github.com/Homebrew/homebrew/issues/46114


mikemcquaid:
Discussed in Slack. Happy to iterate on the caveats.

PR #46162, opened 2015-11-19
Skip pouring from a bottle if --cc is passed

### LINKED ISSUE #46046: Build from source should be the default behavior if --cc option is passed
I needed to build [Boost](http://www.boost.org) using Homebrew GCC (i.e. `gcc-5` and `g++-5`).
`brew install boost --cc=gcc-5` results in boost being installed from the bottle:

<pre>
$ brew install boost --cc=gcc-5
==> Downloading https://homebrew.bintray.com/bottles/boost-1.59.0.el_capitan.bot
==> Pouring boost-1.59.0.el_capitan.bottle.1.tar.gz
🍺  /usr/local/Cellar/boost/1.59.0: 10928 files, 454M
</pre>

As a work-around, I can additionally pass the `--build-from-source` option, but I think that building from source should be the default behavior if the `--cc` option is passed.

dtrebbien:
A formula should be built from source by default if the --cc option is
passed to specify a particular compiler.

Added a test to test_formula_installer: test_not_poured_from_bottle_when_compiler_specified

Modified test_formula_installer to assert that the formula was not poured
from a bottle. Similarly modified test_formula_installer_bottle to assert
that the formula *was* installed from a bottle.

Added an install method to the TestballBottle formula (the same as the
Testball formula's install method) so that the TestballBottle formula can
be "built from source".

Fixes #46046 - Build from source should be the default behavior if --cc
option is passed

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -80,6 +80,7 @@ class FormulaInstaller
     bottle = formula.bottle
     return true  if force_bottle? && bottle
     return false if build_from_source? || build_bottle? || interactive?
+    return false if ARGV.cc
     return false unless options.empty?
     return false if formula.bottle_disabled?
     return true  if formula.local_bottle_path
--- a/Library/Homebrew/test/test_formula_installer.rb
+++ b/Library/Homebrew/test/test_formula_installer.rb
@@ -3,7 +3,9 @@ require "formula"
 require "compat/formula_specialties"
 require "formula_installer"
 require "keg"
+require "tab"
 require "testball"
+require "testball_bottle"
 
 class InstallTests < Homebrew::TestCase
   def temporary_install(formula)
@@ -18,6 +20,8 @@ class InstallTests < Homebrew::TestCase
     assert_predicate formula, :installed?
 
     begin
+      refute_predicate Tab.for_keg(keg), :poured_from_bottle
+
       yield formula
     ensure
       keg.unlink
@@ -65,6 +69,21 @@ class InstallTests < Homebrew::TestCase
 
     temporary_install(formula) do |f|
       assert_predicate f, :installed?
+    end
+  end
+
+  def test_not_poured_from_bottle_when_compiler_specified
+    assert_nil ARGV.cc
+
+    cc_arg = "--cc=llvm-gcc"
+    ARGV << cc_arg
+    begin
+      temporary_install(TestballBottle.new) do |f|
+        tab = Tab.for_formula(f)
+        assert_equal "llvm", tab.compiler
+      end
+    ensure
+      ARGV.delete_if { |x| x == cc_arg }
     end
   end
 end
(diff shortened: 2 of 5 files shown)


DomT4:
CC @xu-cheng :smile:.

tdsmith:
What should happen if HOMEBREW_CC is defined?

mikemcquaid:
@tdsmith Personally I'd like it if we deprecated/removed that in favour of the command-line options. "Build everything with GCC" is a recipe for hard-to-debug errors.

xu-cheng:
Per PR, LGTM :+1:

Per `HOMEBREW_CC` issue, I don't think it's harm to keep it. After all it should be only used when users understand what they are doing.

mikemcquaid:
> Per HOMEBREW_CC issue, I don't think it's harm to keep it. After all it should be only used when users understand what they are doing.

I'm afraid I can't agree here.

xu-cheng:
> I'm afraid I can't agree here.

Wo, I was guessing you won't agree it. IMO, we shouldn't block advanced users hacking Homebrew, so I like to keep it. But that to be said, I won't be against for removing it.

mikemcquaid:
> Wo, I was guessing you won't agree it. IMO, we shouldn't block advanced users hacking Homebrew, so I like to keep it. But that to be said, I won't be against for removing it.

I agree with the advanced users bit; I just think as-is it can encourage some random blog post to say "set `HOMEBREW_CC` to use GCC to make things faster/better/more open-source".

dtrebbien:
> What should happen if HOMEBREW_CC is defined?

I have pushed an amended commit which checks `|| ENV["HOMEBREW_CC"]` as well.

> I agree with the advanced users bit; I just think as-is it can encourage some random blog post to say "set `HOMEBREW_CC` to use GCC to make things faster/better/more open-source".

This seems like something that would happen.  For example, you know Internet Explorer's "Do you want to allow this webpage to access your Clipboard?" dialog?  Well, a common recommendation to "disable" this popup is simply to allow script access to the clipboard, thus disabling an important security protection!  :worried:  See `https://techjourney.net/disable-allow-this-webpage-to-access-your-clipboard-pop-up-warning-message-in-ie7/` for an example of this and the several "thank you" comments.

What about adding a warning to `brew doctor` if `HOMEBREW_CC` is set?

mikemcquaid:
> What about adding a warning to brew doctor if HOMEBREW_CC is set?

:+1: good idea.

> I have pushed an amended commit which checks || ENV["HOMEBREW_CC"] as well.

I'll let other maintainers chime in but I'd probably rather we didn't handle this case.

DomT4:
> I just think as-is it can encourage some random blog post to say "set HOMEBREW_CC to use GCC to make things faster/better/more open-source".

Agree pretty heavily with this.

xu-cheng:
> > I have pushed an amended commit which checks || ENV["HOMEBREW_CC"] as well.

> I'll let other maintainers chime in but I'd probably rather we didn't handle this case

Take a look at [codes for compiler resolution](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/extend/ENV/shared.rb#L151-L163), we handle differently for `--cc` and `HOMEBREW_CC`. So I think we probably want to skip handling this case until we sorted it out internally.

dtrebbien:
> I have pushed an amended commit which checks `|| ENV["HOMEBREW_CC"]` as well.

This is now reverted.

mikemcquaid:

PR #46163, opened 2015-11-19
dromeaudio 0.3.0 (import from head-only)

bfontaine:
See https://github.com/Homebrew/homebrew-head-only/pull/193

### DIFF
--- a/Library/Formula/dromeaudio.rb
+++ b/Library/Formula/dromeaudio.rb
@@ -1,0 +1,21 @@
+class Dromeaudio < Formula
+  desc "Small C++ audio manipulation and playback library"
+  homepage "https://github.com/joshb/dromeaudio/"
+  url "https://github.com/joshb/DromeAudio/archive/v0.3.0.tar.gz"
+  sha256 "d226fa3f16d8a41aeea2d0a32178ca15519aebfa109bc6eee36669fa7f7c6b83"
+
+  head "https://github.com/joshb/dromeaudio.git"
+
+  depends_on "cmake" => :build
+
+  def install
+    # install FindDromeAudio.cmake under share/cmake/Modules/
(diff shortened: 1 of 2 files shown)


bfontaine:
```
-- Installing: /usr/local/Cellar/cmake/3.4.0/share/cmake/Modules/FindDromeAudio.cmake
CMake Error at share/cmake_install.cmake:39 (file):
  file INSTALL cannot copy file
  "/tmp/dromeaudio20151119-84455-vo7cqk/DromeAudio-0.3.0/share/FindDromeAudio.cmake"
  to
  "/usr/local/Cellar/cmake/3.4.0/share/cmake/Modules/FindDromeAudio.cmake".
Call Stack (most recent call first):
  cmake_install.cmake:35 (include)
```

mikemcquaid:
@bfontaine Yeh, it'll need fixed up to not try and install that into CMake's Cellar but instead under `HOMEBREW_PREFIX/share/cmake/Modules`

bfontaine:
@mikemcquaid Installing in this location breaks the sandbox. Should I install under `prefix` then either `cp` or symlink the CMake file in `postinstall`?

With `--sandbox`:
```
CMake Error at share/cmake_install.cmake:39 (file):
  file INSTALL cannot copy file
  "/tmp/dromeaudio20151124-80542-1979tka/DromeAudio-0.3.0/share/FindDromeAudio.cmake"
  to "/usr/local/share/cmake/Modules/FindDromeAudio.cmake".
Call Stack (most recent call first):
  cmake_install.cmake:35 (include)
```

mikemcquaid:
> @mikemcquaid Installing in this location breaks the sandbox. Should I install under prefix then either cp or symlink the CMake file in postinstall?

Sorry, yeh, maybe this should be installed inside the prefix/share/cmake/Modules and `brew link`d automatically?

bfontaine:
I need to fix the test but otherwise it works :+1:

PR #46165, opened 2015-11-19
Adding fuzzy-choose 1.0

sdegutis:
This should meet the requirements as helpfully laid out by the kind and generous @DomT4.

### DIFF
--- a/Library/Formula/fuzzy-choose.rb
+++ b/Library/Formula/fuzzy-choose.rb
@@ -1,0 +1,17 @@
+class FuzzyChoose < Formula
+  desc "Hybrid string fuzzy matcher that uses both std{in,out} and a native GUI"
+  homepage "https://github.com/sdegutis/choose"
+  url "https://github.com/sdegutis/choose/archive/1.0.tar.gz"
+  sha256 "b1d16c6e143e2a9e9b306cd169ce54535689321d8f016308ff26c82c3d2931bf"
+
+  depends_on :xcode => :build
+
+  def install
+    xcodebuild "SDKROOT=", "SYMROOT=build"
+    bin.install "build/Release/choose"
+  end
+
+  test do
+    system "#{bin}/choose", "-h"
+  end
+end


Library/Formula/fuzzy-choose.rb
```diff
@@ -0,0 +1,17 @@
+class FuzzyChoose < Formula
+  desc "Hybrid string fuzzy matcher that uses both std{in,out} and a native GUI"
```
bfontaine: Please use a shorter description. “`fuzzy-choose: <desc>`” should fit in 80 chars.

bfontaine:
Could you please add ``conflicts_with "choose", :because => "both install a `choose` binary"`` as well as the corresponding line in the `choose` formula? [See an example here](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/ganglia.rb#L21)

Library/Formula/fuzzy-choose.rb
sdegutis: @bfontaine Actually I worded it the way I do so it would fit in a "Show HN" post title, which has an 80-char limit (including the "Show HN: " part). So currently it already is 71 characters. `echo -n "Hybrid string fuzzy matcher that uses both std{in,out} and a native GUI" | wc -c` => `71`

sdegutis:
@bfontaine At this point I think I'll just try renaming the whole project as well as the binary's name. Is there a way I can search all of Homebrew's repos where it would conflict? That way I can just search them to find a name that isn't already taken. When I tried `brew search choose` it came up empty with no results, which is why I named it choose in the first place. So is there a more "full" search function I can use that will ensure this issue doesn't come up again?

bfontaine:
@sdegutis Homebrew doesn’t support searching binaries; you can however look at this file which lists all the binaries of all the formulae from the core and the official taps: https://github.com/Homebrew/homebrew-command-not-found/blob/master/executables.txt

sdegutis:
@bfontaine Thanks! Found a good name. Will rename the repo (and executable) later and send a new PR.

DomT4:
Thanks for the PR! Feel free to ping in the new one when filed if you need anything there.

PR #46178, opened 2015-11-19
audit: check position of 'revision' and 'keg_only'

UniqMartin:
I'd like to establish a standard position for these two attributes in a formula. From my survey of existing formulae, this seems to be the most widely used and most reasonable location for them.


DomT4:
Discussion point: I actually prefer to bump the head line deeper if we revision, i.e.

```diff
 url "https://example.com/example.tar.gz"
 sha256 "example"
-head  "https://example.com/example.git"
+revision 1

+head "https://example.com/example.git"
```

UniqMartin:
That means without a revision, the `head` line stays attached to the top block, and with a revision it goes underneath it and separated by an empty line? Or what is your mental model here?

DomT4:
Yeah. My mental process on that is that `revision` bumps apply to `stable` and `devel`; if you `brew upgrade` or `brew upgrade --devel` you'll get that revision bump applied to your local install.

`brew upgrade --HEAD` does nothing at all, and revisions are ignored for it. IMO it makes sense to create separation in that situation between the revisioned and unrevisioned elements.

The complication is when there's also a `devel do` block since at that point we don't move `devel do` above the `revision` line. That's relatively rare though - there are 56 core formula with devel blocks and over 1000 with `head`s.

UniqMartin:
Thanks for explaining your reasons! You've got me convinced and I'll happily change it, but will wait to solicit more feedback. Personally, I'm not very passionate about any particular order, as long as it's not counterintuitive. But I'd really love to see it standardized.

DomT4:
I've tagged this with the feedback tag in the hope it, well, gets some wider feedback :smile:.

bfontaine:
I’m :+1: with @DomT4.

UniqMartin:
Just pushed an update with the improved position for `revision`.

DomT4:
Relatively minor core change, so I'll take Baptiste and I as enough authority to merge :smile:.

Merged in https://github.com/Homebrew/homebrew/commit/df89889e41d1c4120e6a995c6da2f6521a4e1e10. Thanks again @UniqMartin!

UniqMartin:
Thanks! :smile:

PR #46179, opened 2015-11-19
describe commit message format in CONTRIBUTING

jasonkarns:
Each time I open new pull requests to bump formula versions, I need to lookup the desired commit message/pr title format. I end up on the wiki or docs and then contributing, and finally following the link from the contributing doc. As the contributing doc is the quickest thing to review from an in-progress pull request, I think it would be best served to describe the desired commit message format front and center in the contributing doc. At least, as an intermittent contributor, that's where I think it's most helpful.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,11 @@ Please read:
 
 * [Code of Conduct](https://github.com/Homebrew/homebrew/blob/master/CODEOFCONDUCT.md#code-of-conduct)
 
+Commit message format for simple version updates:
+
+* `"FORMULA_NAME NEW_VERSION"`, e.g. `"source-highlight 3.1.8"`
+* devel version bumps should have the commit message marked with addtional (devel) suffix like `"nginx 1.9.1 (devel)"`
+
 To contribute code please read:
 
 * [Formula API](http://www.rubydoc.info/github/Homebrew/homebrew/master/frames)


CONTRIBUTING.md
```diff
@@ -14,6 +14,11 @@ Please read:
 
 * [Code of Conduct](https://github.com/Homebrew/homebrew/blob/master/CODEOFCONDUCT.md#code-of-conduct)
 
+Commit message format for simple version updates:
+
+* `"FORMULA_NAME NEW_VERSION"`, e.g. `"source-highlight 3.1.8"`
+* devel version bumps should have the commit message marked with addtional (devel) suffix like `"nginx 1.9.1 (devel)"`
+
```
UniqMartin: Maybe remove the straight double quotes, as the messages are already quoted in backticks?

UniqMartin:
I like the idea. :+1: Would you mind extending the list a bit more? I was thinking of something like:

- New version of existing formula: `<formula> <version>`
  - Development version: `<formula> <version> (devel)`
- New formula: `<formula> <version> (new formula)`
- Formula changes without version bump: `<formula>: <summary of changes>`

Feel free to tweak the exact wording and enrich with examples.

DomT4:
FWIW, We do go into detail in the rather bluntly named document [here](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md) and 99% of contributors ignore that entirely. Wherever we put it, I'm increasingly sure it'll get ignored.

UniqMartin:
> FWIW, We do go into detail in the rather bluntly named document [here](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md) and […]

Maybe we should also update that document and make the expected commit message format stand out a bit more, e.g., by putting it in a bullet list. Explicitly listing the new/tweaked formula case wouldn't hurt, as these are common enough. Also bullet point 5 (Make a separate commit for each changed formula) feels like a more natural place for that than point 7 when everything's already committed and pushed.

In the end, we won't be able to make every contributor read that. That's just human nature I guess. But for those who are willing to read, I think we can make it a bit more accessible.

DomT4:
If people want to make the documentation a little prettier I certainly don't think anyone'll object. Just don't underestimate how much people do not enjoy reading documentation, essentially.

What works for one may not work for others, and someone else may then go *"Well you put the same language here but I'd personally prefer to read it in this document so let's copy it there as well"* and then we end up updating one document but forgetting the others. That's my concern on this sort of thing.

UniqMartin:
I can see how this kind of duplication can be bad and agree. The solution to this could be a separate document that explains the commit message format and that both mentioned documents would link to. It could be very brief in its first paragraph, just listing the various typical cases, and then go into more detail for those who are not sure yet and enjoy reading it.

Let's see what the author of this PR thinks about our discussion and if that's out of scope for him.

jasonkarns:
I posit that it's not the clarity of prettiness of the documentation that needs changed, but rather the *location*. As a user, I'm quite frequently used to seeing `CONTRIBUTING.md` docs, and github obviously does a great job making this forefront when submitting a PR. However, let's review what we see when we view CONTRIBUTING.md: a list of links for further reading. The *last* thing I'm going to do as I'm about to open a PR is delve through 8 more documents linked from the contributing doc. I should be able to get the majority of the information I need right from that view. In short, it all needs simplified. Having those links is great. But they need to be for further reading. And most of that information isn't critical *at the time of submitting a PR*. (Which is right when the CONTRIBUTING doc is brought to the fore.) 

I agree that duplicating this information is bad; and to remedy that I think majority of the information needs to be summarized and taken in context for a user who has a PR in draft and is about to hit submit. At that point, what is the most important thing that would help the maintainers for the contributor to know? And make *that* front and center in the CONTRIBUTING doc. Everything else that's less important (or used by a smaller subset of people) can be hidden behind Further Reading links.

DomT4:
FWIW that our documentation is a mess has been discussed numerous times at considerable length by the other maintainers & I.

The hold-up on moving forwards with that has been more or less agreeing on an approach to documentation (Particularly since we're making an effort to create better API documentation) and then someone having the time and patience to rewrite it and go through a lengthy review process with a bunch of changes and minor discussions.

jasonkarns:
Understood. Please don't consider this PR or my comments to be a critique
on the team or anything.

I especially understand that, especially with regards to documentation,
multiple small-step PRs can likely be a detractor and not actually improve
things on the whole. What's probably needed (as has likely been discussed)
is a larger scale rewrite with a plan/roadmap. This feels like the kind of
thing where numerous small PRs may be optimizing in the small, but not
helping in the large.

I can work on this PR a bit to see what I can do. But I also understand if
you find it better to work on a general game plan to attack the entire
documentation problem before accepting smaller changes like this.

DomT4:
Oh no worries at all. It was almost a qualification on my earlier comment which wasn't particularly intended as a critique of this PR alone; just letting you know really that we're aware of the wider issues and trying to come to some agreement on how to fix them.

It's always good to hear from contributors what's actually useful and what is never read, what is missed, what people don't care about, etc. The feedback is great in terms of keeping it in mind every time the documentation issue gets discussed.

mikemcquaid:
My concern with this is that there's a bunch of other stuff that we could argue should be in here. I'm wondering if we should add more in here or try and make it clear you don't have to read all linked documents (perhaps with a couple of words summary for why you'd read each one).

jasonkarns:
I'm reminded of:
http://thecodelesscode.com/case/215?utm_content=buffer8103a&utm_medium=social&utm_source=twitter.com&utm_campaign=buffer

I think whatever can be done to make it simpler, will make it more likely
to be effective and useful.

jasonkarns:
I'm reminded of the [parable of documentation](http://thecodelesscode.com/case/215?utm_content=buffer8103a&utm_medium=social&utm_source=twitter.com&utm_campaign=buffer)

I think whatever can be done to make it simpler, will make it more likely
to be effective and useful.

DomT4:
Going to close in favour of https://github.com/Homebrew/homebrew/pull/46618, which wraps in a whole bunch of documentation changes including ones similar to this. Thank you for caring enough about documentation to actually file PRs against it though @jasonkarns :heart:

mikemcquaid:
Yes thanks @jasonkarns; this provoked me into spending 4 :airplane: journeys rereading and fixing our documentation with what you'd written here in mind.

PR #46180, opened 2015-11-19
New formula: pldebugger.

lifepillar:
This is not meant to be accepted as is. I open this pull request to get some feedback and see if there is any chance to ever get it accepted.

- pldebugger has a stable 1.0 version (not tagged as such, though), but no packaged release. Is the formula acceptable if the source is a Git tag?

- The formula installs files in `$HOMEBREW/share/postgresql/extension`. They persist there after `brew remove pldebugger` (I think that `pgtap` is a similar case). An alternative might be to include pldebugger in the `postgresql` formula (presumably, as an optional resource). What do you think?

- Assuming that this formula might be ok, how should I change the test for automated testing?

- In case this doesn't get accepted, is homebrew-head-only a viable alternative?

### DIFF
--- a/Library/Formula/pldebugger.rb
+++ b/Library/Formula/pldebugger.rb
@@ -1,0 +1,41 @@
+class Pldebugger < Formula
+  desc "PL/pgSQL debugger server-side code"
+  homepage "http://git.postgresql.org/gitweb/"
+  url "http://git.postgresql.org/git/pldebugger.git",
+      :tag => "REL-9_5_0",
+      :revision => "85d7b3b2821301e182d5974d9e6f353d7a241eff"
+  version "1.0" # See default_version field in pldbgapi.control
+  sha256 "2bb8e27aa8f8434a4861fdbc70adb9cb4b47e1dfe472910d62d6042cb80a2ee1"
+
+  head "git://git.postgresql.org/git/pldebugger.git"
+
+  depends_on "postgresql"
+
+  def install
+    ENV["USE_PGXS"] = "1"
+    pg_config = "#{Formula["postgresql"].opt_bin}/pg_config"
+    system "make", "PG_CONFIG=#{pg_config}"
+    mkdir "stage"
+    system "make", "DESTDIR=#{buildpath}/stage", "PG_CONFIG=#{pg_config}", "install"
+    lib.install Dir["stage/**/lib/*"]
+    (doc/"postgresql/extension").install Dir["stage/**/share/doc/postgresql/extension/*"]
+    (share/"postgresql/extension").install Dir["stage/**/share/postgresql/extension/*"]
+  end
+
+  test do
+    pg_bin = Formula["postgresql"].opt_bin
+    pg_port = "55561"
+    system "#{pg_bin}/initdb", testpath/"test"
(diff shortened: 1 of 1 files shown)


mistydemeo:
> pldebugger has a stable 1.0 version (not tagged as such, though), but no packaged release. Is the formula acceptable if the source is a Git tag?

A git tag is definitely acceptable! However, we prefer not to invent arbitrary version numbers. Is there a source for "1.0" being an official version number for this tag? If not, please omit the version number and the tag will be used as the version.

mistydemeo:
> The formula installs files in $HOMEBREW/share/postgresql/extension. They persist there after brew remove pldebugger (I think that pgtap is a similar case). An alternative might be to include pldebugger in the postgresql formula (presumably, as an optional resource). What do you think?

Formulae can't install into another formula's prefix, or into `HOMEBREW_PREFIX`. Will postgres find these files if they're installed into pldebugger's prefix and symlinked into that location within the `HOMEBREW_PREFIX`?

lifepillar:
>Is there a source for "1.0" being an official version number for this tag?

See `default_version` in `pldbgapi.control` (and [§35.15.1](http://www.postgresql.org/docs/current/static/extend-extensions.html) in PostgresSQL's manual). Also, the installed files have a `--1.0.sql` suffix.

>Will postgres find these files if they're installed into pldebugger's prefix and symlinked into that location within the HOMEBREW_PREFIX?

I think so, because other formulae (e.g., `postgis`) do that (I should have checked before!).

Ok, so I will update the formula based on your feedback.

mistydemeo:
> See default_version in pldbgapi.control (and §35.15.1 in PostgresSQL's manual). Also, the installed files have a --1.0.sql suffix.

Okay! Can you add a comment to that effect so people know how to find the version when it gets bumped in the future?

Library/Formula/pldebugger.rb
```diff
@@ -0,0 +1,34 @@
+class Pldebugger < Formula
+  desc "PL/pgSQL debugger server-side code"
+  homepage "http://git.postgresql.org/gitweb/"
+  url "http://git.postgresql.org/git/pldebugger.git", :tag => "AS92_UPD-REL-9_2_2_4"
```
DomT4: We need a `revision` line here as well, the commit hash.

Library/Formula/pldebugger.rb
```diff
@@ -0,0 +1,34 @@
...
+  version "1.0"
+  sha256 "2bb8e27aa8f8434a4861fdbc70adb9cb4b47e1dfe472910d62d6042cb80a2ee1"
+
+  head "git://git.postgresql.org/git/pldebugger.git", :branch => "master"
```
DomT4: We assume `master` is the default. Doesn't need specifying.

mistydemeo:
Looks like it's failing to build - mind looking into it?

lifepillar:
PostgreSQL >= 9.3 needs the `HEAD` version of this formula. I have contacted the author to ask whether the current head may be tagged.

lifepillar:
The author has added a new tag and I have updated the formula accordingly. I don't understand the two CI failures left, though: they do not seem to depend on the formula.

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `pldebugger 1.0 (new formula)`.

PR #46185, opened 2015-11-19
utils: native rewrite of install_gem_setup_path!

UniqMartin:
Setting the environment variable `HOMEBREW_RUBY_PATH` allows one to run Homebrew with a non-standard Ruby, even one that is not in `PATH`. This creates the problem that the `gem` in `PATH` might not be the right one to call, possibly leading to confusing results as the wrong RubyGems installation is queried/manipulated.

[This comment](https://github.com/Homebrew/homebrew/pull/45816#issuecomment-157685795) very nicely explains why relying on `PATH` can be problematic. But currently, it has to be set as is done in [line 33 of .travis.yml](https://github.com/Homebrew/homebrew/blob/16747a5285e2728e534f9045084f2abf4218e6bc/.travis.yml#L33). Otherwise things will break:

```
$ with-ruby 1.8 brew tests
>> Setting HOMEBREW_RUBY_PATH to '/opt/homebrew/opt/ruby187/bin/ruby'.
Error: The 'bundler' gem is installed but couldn't find 'bundler' in the PATH:
~/.gem/ruby/1.8/bin:[…]
```

(Note how it claims `bundler` is installed, because it's asking the wrong `gem`.) This PR addresses this particular problem, making it possible to leave `PATH` unaltered and only adjust `HOMEBREW_RUBY_PATH`:

```
$ with-ruby 1.8 brew tests
>> Setting HOMEBREW_RUBY_PATH to '/opt/homebrew/opt/ruby187/bin/ruby'.
==> Installing missing or updating outdated 'bundler' gem
Successfully installed bundler-1.10.6
1 gem installed
Fetching gem metadata from https://rubygems.org/.........
Fetching version metadata from https://rubygems.org/..
Installing rake 10.4.2
Installing docile 1.1.5
Installing json 1.8.3 with native extensions
Installing metaclass 0.0.4
Installing minitest 5.7.0
Installing mocha 1.1.0
Installing simplecov-html 0.10.0
Installing simplecov 0.10.0
Using bundler 1.10.6
Bundle complete! 4 Gemfile dependencies, 9 gems now installed.
Bundled gems are installed into ./vendor/bundle.
Run options: --seed 59757
[…]
```

I tested this successfully on my local machine (OS X 10.11.1) with Ruby 1.8.7, 1.9.3, 2.0.0, 2.1.7, and 2.2.3 as provided by core and `homebrew/versions`. I cleaned out the respective `.gem` directories (except for the one used by the system Ruby) before running the test.

CC @xu-cheng @mikemcquaid

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -190,12 +190,23 @@ module Homebrew
     require "rubygems"
     ENV["PATH"] = "#{Gem.user_dir}/bin:#{ENV["PATH"]}"
 
-    args = [gem]
-    args << "-v" << version if version
+    if Gem::Specification.find_all_by_name(gem, version).empty?
+      ohai "Installing or updating '#{gem}' gem"
+      install_args = %W[--no-ri --no-rdoc --user-install #{gem}]
+      install_args << "--version" << version if version
 
-    unless quiet_system "gem", "list", "--installed", *args
-      safe_system "gem", "install", "--no-ri", "--no-rdoc",
-                                    "--user-install", *args
+      # Do `gem install [...]` without having to spawn a separate process or
+      # having to find the right `gem` binary for the running Ruby interpreter.
+      require "rubygems/commands/install_command"
+      install_cmd = Gem::Commands::InstallCommand.new
+      install_cmd.handle_options(install_args)
+      exit_code = 1 # Should not matter as `install_cmd.execute` always throws.
+      begin
+        install_cmd.execute
+      rescue Gem::SystemExitException => e
+        exit_code = e.exit_code
+      end
+      odie "Failed to install/update the '#{gem}' gem." if exit_code != 0
     end
 
     unless which executable


Library/Homebrew/utils.rb
```diff
@@ -188,14 +188,32 @@ def self.homebrew_version_string
 
   def self.install_gem_setup_path!(gem, version = nil, executable = gem)
     require "rubygems"
-    ENV["PATH"] = "#{Gem.user_dir}/bin:#{ENV["PATH"]}"
 
-    args = [gem]
-    args << "-v" << version if version
+    if Gem::Specification.find_all_by_name(gem, version).empty?
+      ohai "Installing missing or updating outdated '#{gem}' gem"
```
dunn: I think it reads better as just `"Installing or updating '#{gem}' gem"`.

xu-cheng:
:+1: LGTM

Library/Homebrew/utils.rb
```diff
@@ -188,14 +188,32 @@ def self.homebrew_version_string
...
-    unless quiet_system "gem", "list", "--installed", *args
-      safe_system "gem", "install", "--no-ri", "--no-rdoc",
-                                    "--user-install", *args
+    # Make sure RubyGems `bin` directory is in `PATH` and is before all others.
+    path = ENV["PATH"].split(File::PATH_SEPARATOR)
+    unless path.first && path.first == "#{Gem.user_dir}/bin"
+      path.delete("#{Gem.user_dir}/bin")
+      path.unshift("#{Gem.user_dir}/bin")
+      ENV["PATH"] = path.join(File::PATH_SEPARATOR)
     end
```
mikemcquaid: I don't see the advantage here over unconditionally prepending?

Library/Homebrew/utils.rb
mikemcquaid: This is gonna be a lot more noisy if it's printed for every gem. I think we can perhaps just omit this?

mikemcquaid:
A few nits otherwise :+1:

Library/Homebrew/utils.rb
UniqMartin: > I think it reads better as just `"Installing or updating '#{gem}' gem"`.

Agreed.

> This is gonna be a lot more noisy if it's printed for every gem.

I found it genuinely useful to see why my invocation of `brew <command>` is stalling before the first output from `gem install` hits my terminal. And it is only one line for the originally requested gem, not for every dependency that results from that. (See the longer example output from my original comment. The ensuing output of `gem install` is much noisier.) Finally, the worst case scenario is that we get three additional lines of output, because there are only three places in core that call this method and request a gem (commands `man`, `style`, and `tests` requesting one of `ronn`, `rubocop`, and `bundler`). Does that sound convincing?

mikemcquaid: I'm convinced :+1:

Library/Homebrew/utils.rb
UniqMartin: The idea is to avoid duplicating the path or prepending it multiple times, if the method is called repeatedly during a single `brew` run (only a precaution, not currently happening). If we agree on the advantage, a more concise version of this code would be:

```ruby
gem_bin = "#{Gem.user_dir}/bin"
path = ENV["PATH"].split(File::PATH_SEPARATOR).reject { |p| p == gem_bin }
ENV["PATH"] = ([gem_bin] + path).join(File::PATH_SEPARATOR)
```

mikemcquaid:
Would be good to benchmark this with `brew prof` to see how the performance differs: https://github.com/Homebrew/homebrew-dev-tools/blob/master/cmd/brew-prof.rb

Library/Homebrew/utils.rb
mikemcquaid: I'm not sure I see the problem by multiple prepends if we're not outputting the PATH anywhere? I guess I prefer simple code (more code == more bugs)

UniqMartin: That's true, multiple instances don't create any problem I'm aware of. I only changed this because I didn't like the thought of having duplicates in there. I'm fine with going back to:

```ruby
ENV["PATH"] = "#{Gem.user_dir}/bin#{File::PATH_SEPARATOR}#{ENV["PATH"]}"
```

mikemcquaid:
@UniqMartin Thanks!

Library/Homebrew/utils.rb
UniqMartin: My primary reason still stands (output before stalling with network action). But I have to admit that the verbosity in my example is due to me having moved `ENV["PATH"] =` (now back to the top) and mostly due to `bundler` (invoked by `brew tests`). Just in case you'd like to reconsider.

More representative output with profiling information is forthcoming.

UniqMartin:
Thanks for the feedback! I pushed an update that hopefully incorporates it all.

> Would be good to benchmark this with `brew prof` to see how the performance differs.

I'm not sure how useful that information is as most of the time will be spent downloading/building gems or, if the gem is already present, running external commands provided by these gems.

Here's a summary (total time, [gist with details](https://gist.github.com/UniqMartin/c46059be716e554fd620); *=including download/build time):

| What                                    | Old     | New     |
|-----------------------------------------|---------|---------|
| `brew prof man` with empty `~/.gem`     | 75.22s* | 85.52s* |
| `brew prof man` with populated `~/.gem` |  1.31s  |  1.13s  |

Diff of output for old/new version with empty `~/.gem`:

```diff
 $ brew prof man
+==> Installing or updating 'ruby-prof' gem
 Fetching: ruby-prof-0.15.8.gem (100%)
 Building native extensions.  This could take a while...
 Successfully installed ruby-prof-0.15.8
 1 gem installed
+==> Installing or updating 'ronn' gem
 Fetching: hpricot-0.8.6.gem (100%)
 Building native extensions.  This could take a while...
 Successfully installed hpricot-0.8.6
 Fetching: rdiscount-2.1.8.gem (100%)
 Building native extensions.  This could take a while...
 Successfully installed rdiscount-2.1.8
 Fetching: mustache-1.0.2.gem (100%)
 Successfully installed mustache-1.0.2
 Fetching: ronn-0.7.3.gem (100%)
 Successfully installed ronn-0.7.3
 4 gems installed
 Writing HTML fragments to /opt/homebrew/share/doc/homebrew
 Writing manpages to /opt/homebrew/share/man/man1
```

Output for old/new version with populated `~/.gem`:

```diff
 $ brew prof man
 Writing HTML fragments to /opt/homebrew/share/doc/homebrew
 Writing manpages to /opt/homebrew/share/man/man1
```

UniqMartin:
Here are some (in my opinion) more useful numbers for the common case that the requested gem is already present. It very nicely shows the overhead of running `gem list --installed`.

Old:

```
$ brew ruby -rbenchmark -e '3.times { puts Benchmark.measure { Homebrew.install_gem_setup_path! "bundler" } }'
  0.000000   0.010000   0.180000 (  0.181835)
  0.000000   0.000000   0.180000 (  0.188135)
  0.000000   0.000000   0.200000 (  0.194010)
```

New:

```
$ brew ruby -rbenchmark -e '3.times { puts Benchmark.measure { Homebrew.install_gem_setup_path! "bundler" } }'
  0.020000   0.000000   0.020000 (  0.025470)
  0.000000   0.000000   0.000000 (  0.000124)
  0.000000   0.000000   0.000000 (  0.000115)
```

mikemcquaid:
Nice, definitely a win here :+1:

mikemcquaid:
Thanks @UniqMartin!

UniqMartin:
> Nice, definitely a win here :+1:

Yes, indeed. Though the main point is still fixing an issue that would surface once `HOMEBREW_RUBY_PATH` is set to a `ruby` binary not in `PATH` without adapting `PATH` …

mikemcquaid:
Yup :+1:

xu-cheng:
I think we can remove `PATH` tweaking for ruby 1.8 in `.travis.yml` now, right?

mikemcquaid:
@xu-cheng Should be able to, yep.

ilovezfs:
It turns out this PR is the cause of #48264
git reverting it leads to a successful run without further intervention in a new, empty brew installation on 10.8.
https://gist.github.com/ilovezfs/99bf57567f0de96f5627

UniqMartin:
> It turns out this PR is the cause of #48264

That might be true, but you are citing an issue that affects an unsupported OS X release. Also, this PR wasn't done to annoy users of legacy OS X releases, but to solve a problem in non-standard setups on non-legacy OS X releases. I agree that retaining backwards compatibility where feasible is a good goal, but if that hinders progress or makes the code unreadable because of all the compatibility workarounds, then maybe it's best to leave a few users behind.

> git reverting it leads to a successful run without further intervention in a new, empty brew installation on 10.8.

I'd be happier if you weren't suggesting a simple revert without understanding its implications (sorry, maybe I'm wrong and you do understand but are just trolling). If you can come up with a fix that solves the issue with older RubyGems versions without regressing on the improvements this PR made, I'm happy to review it. (Personally, I'm more interested in advancing Homebrew for a majority of its users instead of keeping it alive on OS X releases from the distant past.)

ilovezfs:
1. Making original reviewers aware of subsequent regressions arising from the PR on the PR itself is neither an indictment of the PR nor a demand or expectation of a fix.
2. There seems to have been some doubt about whether this PR was or was not the cause: https://github.com/Homebrew/homebrew/commit/8f25e002fefce53af2225060bb5de2e59549e49d#commitcomment-15535020
3. "without understanding its implications" ... I'll just assume that was some kind of attempt at humor.
4. I'm in no need of a fix, but you also objected to the other workaround I mentioned, so I thought I'd point out an alternative that I now know to have been effective at solving the problem, as well.

UniqMartin:
@ilovezfs I'm sorry if there has been a misunderstanding; I guess I've interpreted too much into your words. The way you phrased your comment implied to me that this PR was to blame and reverting the related commit would fix the issue. There was no word that indicated to me that you understood why this PR had happened in the first place and why it could be a bad idea to revert it.

> I'm in no need of a fix, but you also objected to the other workaround I mentioned, so […]

I guess with that you're referring to setting `HOMEBREW_RUBY_PATH` and `HOMEBREW_DEVELOPER`? I don't think I was objecting to this, but merely pointing out some caveats to be aware of when doing that.

ilovezfs:
@UniqMartin I think we're on the same page now. I was in no way suggesting the PR be removed from master – it absolutely should not be.

PR #46186, opened 2015-11-20
node: use bundled npm dependency during install

### LINKED ISSUE #46131: npm: `npm outdated -g` errors - specific to Homebrew npm installation

ericvw:
This simplifies the installation of `npm` when installed with `node`.
Note that the existing behavior is still maintained where `npm` modules
are properly installed under `/usr/local/lib/node_modules` and will
remain stable across upgrades.

Closes #46131

Figured out a more elegant solution compared to #46139.

### DIFF
--- a/Library/Formula/node.rb
+++ b/Library/Formula/node.rb
@@ -32,14 +32,6 @@ class Node < Formula
     fails_with :gcc => n
   end
 
-  # We track major/minor from upstream Node releases.
-  # We will accept *important* npm patch releases when necessary.
-  # https://github.com/Homebrew/homebrew/pull/46098#issuecomment-157802319
-  resource "npm" do
-    url "https://registry.npmjs.org/npm/-/npm-3.3.12.tgz"
-    sha256 "09475d7096731d93c0aacd7dfe58794d67c52ee6562675aee6c1f734ddba8158"
-  end
-
   resource "icu4c" do
     url "https://ssl.icu-project.org/files/icu4c/56.1/icu4c-56_1-src.tgz"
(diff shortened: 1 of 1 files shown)


ericvw:
cc @DomT4

ericvw:
The trick is where you you write out the `npmrc`.

DomT4:
I'm sure you're starting to contemplate throwing me into a wall at this point but we don't want to go down this route either :cry:.

There have been times when `npm` have pushed a security release and we've been able to fold that into our formula in less than an hour and bump everyone's installs to resolve the problem. This way we'd be relying on Node to cut a new release, which is more difficult for them as they have more "stuff" to get through before they can get to that point.

The current formula works for 99.9% of `npm` operation, internally and externally. I have little desire to make radical changes to a stable formula. We agreed with `npm` a long time ago the tarball solution worked best for us and them, and barring the `outdated -g` problem we need to resolve, it has proven stable and reliable.

IMO there has to be a simpler fix than reinventing the wheel here.

ericvw:
No worries.  I am equally trying to figure a simpler fix and ensuring that what changes I propose keep in line with Hombrew's philosophy and history :).

DomT4:
Sorry, I know this is all frustrating. Pretty much every aspect of the Node formula these days was born out of historical discussions and experience on top of that, and it's been more stable in the last year than it has ever been before.

Genuinely am trying to find common ground here, but I appreciate it can be annoying to work on something and have it shot down repeatedly because of circumstances/history beyond your control.

Did you get a chance at all to try out the hack of just copying the old `package.json` pre-install over to replacing the new one after the `make install` step? It's a hack but curious whether that works at all. I'll try and play with it a bit later today if not.

ericvw:
Closing in favor of PR #46195.

PR #46190, opened 2015-11-20
adding nuget formula

blairham:


Library/Formula/nuget.rb
```diff
@@ -0,0 +1,17 @@
+class Nuget < Formula
+  desc "NuGet is the package manager for the Microsoft development platform including .NET"
+  homepage "https://wwww.nuget.org/"
```
mistydemeo: Typo?

Library/Formula/nuget.rb
```diff
@@ -0,0 +1,17 @@
...
+  url "https://s3.amazonaws.com/homebrew-nuget/nuget-3.2.0.10516.tar.gz"
```
mistydemeo: What distinguishes the homebrew package from the others?

Library/Formula/nuget.rb
```diff
@@ -0,0 +1,17 @@
...
+  depends_on "mono"
+
+  def install
+    bin.install Dir["*"]
+  end
+
+  test do
+    system "#{bin}/nuget"
+  end
+end
```
mistydemeo: Newline at end of file, please.

Library/Formula/nuget.rb
```diff
@@ -0,0 +1,17 @@
...
+  sha256 "1a76e705a3253d020c5ee8a92816f9c10b8604226eab7fcd196ae29e8bfacbda"
+
+  depends_on "mono"
+
+  def install
+    bin.install Dir["*"]
+  end
+
+  test do
+    system "#{bin}/nuget"
```
mistydemeo: Is there any actual functionality you can test?

DomT4:
Any update on an answer to Misty's question? Thanks!

> What distinguishes the homebrew package from the others?

blairham:
I'm not 100% sure how to answer this, but I will attempt. This package enables you to use NuGet on OSX. It mainly distributes Nuget.exe along with a bash script that passes it through mono.

mikemcquaid:
> I'm not 100% sure how to answer this, but I will attempt. This package enables you to use NuGet on OSX. It mainly distributes Nuget.exe along with a bash script that passes it through mono.

We need to build this from source instead.

blairham:
Good luck with that it's a windows app.

mikemcquaid:
@blairham The homebrew/binary tap and Homebrew Cask may accept it. It's unclear to me that this is an official release supported by `nuget`, though so I'd be concerned about support issues.

blairham:
Ok I'll look into that, this was mainly a way to distribute the bash script to make it work on OS X. I'll also look and see if the code will compile on mono, I have serious doubts it will but I'm willing to attempt some day.

mikemcquaid:
@blairham This is the type of thing where unless the Bash script is trivial enough to be included in the formula (e.g. <10 lines) then we'd like to see upstream distribute it and provide official support. Otherwise we're not really installing `nuget` but `nuget with a Bash wrapper that someone wrote".

blairham:
Understood, then this will never fly as the whole point of the bash script is to make it work on OS X as it just wraps it around mono. I'll just drop it from here

darkn3rd:
http://www.codedistillers.com/?p=99

PR #46195, opened 2015-11-20
node: copy vanilla package.json for npm install

### LINKED ISSUE #46131: npm: `npm outdated -g` errors - specific to Homebrew npm installation

ericvw:
When installing `npm` via the `node` formula, `npm`'s `package.json`
remembers relative paths to where `make install` was invoked.  This
change copys back over the vanilla `package.json` so that `npm -g`
commands behave correctly.

Closes #46131

Hopefully this is better than ##46139 and #46139.

### DIFF
--- a/Library/Formula/node.rb
+++ b/Library/Formula/node.rb
@@ -73,6 +73,10 @@ class Node < Formula
       cd buildpath/"npm_install" do
         system "./configure", "--prefix=#{libexec}/npm"
         system "make", "install"
+        # `package.json` has relative paths to the npm_install directory.
+        # This copies back over the vanilla `package.json` that is expected.
+        # https://github.com/Homebrew/homebrew/issues/46131#issuecomment-157845008
+        cp buildpath/"npm_install/package.json", libexec/"npm/lib/node_modules/npm"
       end
 
       if build.with? "completion"


ericvw:
cc @DomT4

DomT4:
Will poke around with this tonight with fingers crossed it resolves the problem without introducing any new ones, heh. Thanks for submitting the PR :heart:.

ericvw:
"Third time's a charm" - hopefully :).

Library/Formula/node.rb
```diff
@@ -73,6 +73,10 @@ def install
       cd buildpath/"npm_install" do
         system "./configure", "--prefix=#{libexec}/npm"
         system "make", "install"
+        # `package.json` has relative paths to the npm_install directory.
+        # This copies back over the vanilla `package.json` that is expected.
+        # https://github.com/Homebrew/homebrew/issues/46131#issuecomment-157845008
+        cp buildpath/"npm_install/package.json", libexec/"npm/lib/node_modules/npm"
```
mikemcquaid: My concern with this: what if new things are introduced into this file by newer versions of `npm`? Would it not be possible to `inreplace` the relative paths to absolute paths instead?

ericvw: If new things are introduced to the file, this would copy `package.json` from the tarball to the installation location.  Maybe I am missing something, but I think `cp`'ing the newer `package.json` addresses the concern.

shaunstanislaus:
@ericvw  
i am experiencing what you experienced. 
` shaunstanislaus@Master  ~  npm update -g
npm ERR! addLocal Could not install /private/tmp/node20151123-39483-1rq8gak/node-v5.1.0/npm_install`

how do i update to the latest npm package?

Library/Formula/node.rb
DomT4: I think the `cp` is fine, more or less. The problem doesn't go away if you fix the paths. I'll run some tests later today, providing I don't wake up feeling :mask: again.

DomT4:
`npm install -g npm@latest`, and then you can just `npm update -g` to your heart's content after that.

shaunstanislaus:
@DomT4 
i did as you told.

it could only be upgraded to 3.4.1 and not 5.1.0, any idea why?

DomT4:
`3.4.1` is the latest stable [npm](https://github.com/npm/npm/releases) release. `5.1.0` is the Node version, not the `npm` version.

shaunstanislaus:
@DomT4 , okay noted. so there is no way to use the node version?

Library/Formula/node.rb
mikemcquaid: Ah, yes, sorry, I see now. Yeh, this approach seems :+1: to me but will let @DomT4 give the final :ok:.

DomT4: Poking around a bit it looks like the problem is that if you install `npm` from a local destination, `npm` expects that local destination to stick around indefinitely and act as a sort of install/update base for `npm`. I'm not convinced in some way this isn't an edge-case bug that upstream should look at.

The hardcoded paths themselves don't seem to be a huge problem - They exist if you `npm install npm@latest -g` as well, albeit from whoever packaged that `npm` release upstream rather than a local user. Ref: https://github.com/Homebrew/homebrew/issues/46131#issuecomment-157855837

I'm happy here. I can't make it break via normal usage, or inside brew usage, and it resolves the presented issue. I'd like `npm` to take a look at this sort of setup process upstream, but I'm happy to merge this as-is.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/2b40691c5d31a3397d7eef067087db5dffd0b218. Thank you for your persistence and understanding on this @ericvw; I appreciate that, and your contribution to Homebrew.

PR #46196, opened 2015-11-20
groonga: Add `--with-suggest-plugin` option

cosmo0920:
Some advanced Groonga users says that "we want to use `suggest` plugin in OS X".

How do you think about it?


DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/a1f2b7ce14e09a9c4e093b33290e747b4a41a201. Thank you once more @cosmo0920! :smiley_cat:

cosmo0920:
Thanks!

PR #46210, opened 2015-11-20
travis: remove ruby path tweak

xu-cheng:


xu-cheng:
@UniqMartin any idea on ci failure?

UniqMartin:
This might be due to the `#!/usr/bin/env ruby` shebang in `Library/Homebrew/test/vendor/bundle/ruby/1.8/bin/rake`, that I'm also seeing locally. I'm not sure yet what causes this line to be written, though. And I don't understand yet why that doesn't cause any issues when testing locally.

UniqMartin:
> And I don't understand yet why that doesn't cause any issues when testing locally.

After wiping `Library/Homebrew/test/vendor/bundle` and `~/.gem` I see the same issue locally. It has something to do with what shebangs are generated by RubyGems and/or Bundler. Currently, we get `#!/usr/bin/env ruby` and this breaks if the matching Ruby binary is not in `PATH`.

I won't have much time this weekend, so feel free to ponder this issue yourself, if you like. Otherwise I'll investigate how to change this behavior next week …

xu-cheng:
@UniqMartin recently I may not have time around to further debug this. And it doesn't work nicely with my rbenv settings. So I will close PR. Feel free to work on a new PR yourself.

UniqMartin:
Alright, I'll revisit this issue in a PR of my own as soon as I find the time. Thanks for giving it a try!

xu-cheng:
Thanks

UniqMartin:
Should be addressed by #47694.

PR #46211, opened 2015-11-20
minor code improvements

humzashah:
+ replaced some extra `maps`
+ reduced some lines by replacing standard method definitions with `define_method`
+ made a `begin/rescue/else` block more readable by moving the single-line `if` condition in the end and changing it to a guard clause

### DIFF
--- a/Library/Homebrew/dependable.rb
+++ b/Library/Homebrew/dependable.rb
@@ -3,12 +3,20 @@ require "options"
 module Dependable
   RESERVED_TAGS = [:build, :optional, :recommended, :run]
 
-  RESERVED_TAGS.each do |tag|
-    method_name = "#{tag}?"
+  def build?
+    tags.include? :build
+  end
 
-    define_method method_name do
-      tags.include? tag
-    end
+  def optional?
+    tags.include? :optional
+  end
+
+  def recommended?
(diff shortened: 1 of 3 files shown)


Library/Homebrew/dependable.rb
```diff
@@ -3,20 +3,12 @@
...
-  end
+  RESERVED_TAGS.each do |tag|
+    method_name = "#{tag}?"
 
-  def run?
-    tags.include? :run
+    define_method method_name do
+      tags.include? tag
+    end
   end
```
bfontaine: The issue I have with these dynamically-generated methods is that it breaks autocomplete; make them less discoverable (no documentation); and prevent any static analysis: you don’t know if a method simply doesn’t exist or will be created in the future.

dunn: Yeah, I agree.

humzashah: I see. Personally, I add a `def method_name` on top to make it searchable. But it's a non-issue since it doesn't introduced improvements per se. Reverted `define_method`

mikemcquaid:
Seems fine to me :+1:

mikemcquaid:

PR #46213, opened 2015-11-20
charm-tools 1.9.1

marcoceppi:
This is a feature and bug fix release, deprecating 1.8.0

### DIFF
--- a/Library/Formula/charm-tools.rb
+++ b/Library/Formula/charm-tools.rb
@@ -1,8 +1,8 @@
 class CharmTools < Formula
   desc "Tools for authoring and maintaining juju charms"
   homepage "https://github.com/juju/charm-tools"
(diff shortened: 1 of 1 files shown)


marcoceppi:
Welp, not sure how to get at the pip that this installed, but `pip install -r requirements.txt` needs to run as well in this release.

dunn:
We usually add python dependencies as `resources`, as in [`sphinx-doc.rb`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/sphinx-doc.rb).  That way we can check the SHA and cache them.  Is that not possible here?

DomT4:
Should be fine to do it our usual way, looking at the upstream code.

marcoceppi:
It's easier to just patch `setup.py` and do a new upstream release than sort this out. See #46273

PR #46214, opened 2015-11-20
heroku-toolbelt: Remove outdated caveat

grantovich:
The Heroku Toolbelt has replaced Foreman with `heroku local`, which internally uses Forego but does not require any separate installation. See: https://devcenter.heroku.com/articles/heroku-local

### DIFF
--- a/Library/Formula/heroku-toolbelt.rb
+++ b/Library/Formula/heroku-toolbelt.rb
@@ -18,14 +18,6 @@ class HerokuToolbelt < Formula
     bin.write_exec_script libexec/"bin/heroku"
   end
 
-  def caveats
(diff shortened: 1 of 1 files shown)


grantovich:
CI failed on the Ruby 1.8.7 build, which makes sense because the Toolbelt only supports Ruby 1.9+ (and this is already expressed in the formula). Other builds are passing.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/f290e96e2ac2c43c80b24d5444539895f28bdc21. Thank you for your contribution to Homebrew @grantovich; we appreciate it! :smiley_cat:

PR #46218, opened 2015-11-20
qjackctl 0.4.1

david0:


Library/Formula/qjackctl.rb
```diff
@@ -0,0 +1,62 @@
...
+  desc "simple Qt application to control the JACK sound server daemon"
+  homepage "http://qjackctl.sourceforge.net"
+  url "https://downloads.sourceforge.net/qjackctl/qjackctl-0.4.1.tar.gz"
+  sha256 "98024e1ea9f55ac16c290feda051d6494b2261e83f6a918f0fa583b509e2bfd1"
+
+  depends_on "qt"
+  depends_on "jack"
+  depends_on :x11
+
+  patch :DATA
```
dunn: Has this patch been submitted upstream?

Library/Formula/qjackctl.rb
```diff
@@ -0,0 +1,62 @@
+class Qjackctl < Formula
+  desc "simple Qt application to control the JACK sound server daemon"
+  homepage "http://qjackctl.sourceforge.net"
+  url "https://downloads.sourceforge.net/qjackctl/qjackctl-0.4.1.tar.gz"
+  sha256 "98024e1ea9f55ac16c290feda051d6494b2261e83f6a918f0fa583b509e2bfd1"
+
+  depends_on "qt"
+  depends_on "jack"
+  depends_on :x11
```
dunn: If this is a mandatory dependency, I'm afraid it'll have to go in our homebrew/x11 tap.

dunn:
We'll need a test also, thanks!

david0:
I added another patch to remove the X11 dependency and added a test case.
All patches are already contained in upstream.

Im not sure whats going wrong at travis, but it looks like a problem with the travis setup in general.

Library/Formula/qjackctl.rb
```diff
@@ -0,0 +1,118 @@
...
+  desc "simple Qt application to control the JACK sound server daemon"
+  homepage "http://qjackctl.sourceforge.net"
+  url "https://downloads.sourceforge.net/qjackctl/qjackctl-0.4.1.tar.gz"
+  sha256 "98024e1ea9f55ac16c290feda051d6494b2261e83f6a918f0fa583b509e2bfd1"
+
+  depends_on "autoconf" => :build
+  depends_on "qt"
+  depends_on "jack"
+
+  patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Library/Formula/qjackctl.rb
```diff
@@ -0,0 +1,118 @@
...
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--enable-qt4",
+                          "--disable-dbus",
+                          "--disable-xunique",
+                          "--prefix=#{prefix}"
+
+    system "make"
+    prefix.install "src/qjackctl.app"
+    man1.install "qjackctl.1"
```
mikemcquaid: Is there a `make install` you can use?

david0: There is a make install, but it is not of much help: it installs the app to ``/usr/local/Cellar/qjackctl/0.4.1/bin/qjackctl.app/`` which is not what we want I guess.
Also when using make install, I don't know how to make homebrew show up the ``Run `brew linkapps qjackctl` to symlink these to /Applications.``-message

dunn: `linkapps` only works when the app bundle is installed in `libexec` or `prefix`.

Library/Formula/qjackctl.rb
david0: Well this patch is a compilation of different commits that are already contained in upstream. The respective commit hashes are already contained in the patch. I will add additional comments why a backport is necessary from my point of view.

david0:
I updated the formula once again to use make install and added comments to justificate the paches.

Library/Formula/qjackctl.rb
```diff
@@ -0,0 +1,135 @@
+class Qjackctl < Formula
+  desc "simple Qt application to control the JACK sound server daemon"
+  homepage "http://qjackctl.sourceforge.net"
+  url "https://downloads.sourceforge.net/qjackctl/qjackctl-0.4.1.tar.gz"
+  sha256 "98024e1ea9f55ac16c290feda051d6494b2261e83f6a918f0fa583b509e2bfd1"
+  head "http://git.code.sf.net/p/qjackctl/code", :using=>:git
+
+  stable do
+    patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Library/Formula/qjackctl.rb
```diff
@@ -0,0 +1,135 @@
...
+                          "--disable-xunique",
+                          "--prefix=#{prefix}"
+
+    system "make", "install"
+    mv "#{bin}/qjackctl.app", prefix
+    bin.install_symlink "#{prefix}/qjackctl.app/Contents/MacOS/qjackctl"
+  end
+
+  test do
+    Kernel.system "qjackctl", "--version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

Also, instead of using `Kernel.system` use `shell_output` with `1` as the second parameter to verify the exit code and verify the command output with `assert_match`

Library/Formula/qjackctl.rb
david0: No its not possible to submit this patch, I'm sorry. All changesets in this patch are contained in upstream, as i already said before. Their commit hashes are contained in the respective part and can be used to verify this change in upstream.  
And there are also comments explaining their necessity (in the patch data part)

mikemcquaid: In that case add a link to the upstream changesets and a Ruby comment above here explaining why the patch is needed, thanks.

david0:
I updated the test case and added some comments based on the review.

Library/Formula/qjackctl.rb
```diff
@@ -0,0 +1,142 @@
...
+  desc "simple Qt application to control the JACK sound server daemon"
+  homepage "http://qjackctl.sourceforge.net"
+  url "https://downloads.sourceforge.net/qjackctl/qjackctl-0.4.1.tar.gz"
+  sha256 "98024e1ea9f55ac16c290feda051d6494b2261e83f6a918f0fa583b509e2bfd1"
+  head "http://git.code.sf.net/p/qjackctl/code", :using=>:git
+
+  stable do
+    # fixes compile error with getDeviceUIDFromID and combo box device selection is general
+    # also fixes linking without X11
+    patch :DATA
```
mikemcquaid: Please add the links here too.

david0: Done

mikemcquaid:

PR #46219, opened 2015-11-20
haskell-stack 0.1.8.0

borsboom:
Since Stack no longer uses the `ignore` package, the dependence on libpcre is removed.

### DIFF
--- a/Library/Formula/haskell-stack.rb
+++ b/Library/Formula/haskell-stack.rb
@@ -5,8 +5,8 @@ class HaskellStack < Formula
 
   desc "The Haskell Tool Stack"
   homepage "https://www.stackage.org"
-  url "https://hackage.haskell.org/package/stack-0.1.6.0/stack-0.1.6.0.tar.gz"
-  sha256 "a47ffc204b9caef8281d1e5daebc21bc9d4d2414ed695dc10d32fcca4d81978d"
+  url "https://hackage.haskell.org/package/stack-0.1.8.0/stack-0.1.8.0.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/haskell-stack.rb
```diff
@@ -4,9 +4,9 @@ class HaskellStack < Formula
   include Language::Haskell::Cabal
 
   desc "The Haskell Tool Stack"
-  homepage "https://www.stackage.org"
-  url "https://hackage.haskell.org/package/stack-0.1.6.0/stack-0.1.6.0.tar.gz"
-  sha256 "a47ffc204b9caef8281d1e5daebc21bc9d4d2414ed695dc10d32fcca4d81978d"
+  homepage "https://github.com/commercialhaskell/stack"
```
DomT4: Why the change here? Upstream's Github repo still seem to point to the existing one as the homepage?

borsboom: That was a historical detail (an earlier incarnation of stack was called
stackage-build before the two projects were separated), and as of now no
longer does.  The `stack.cabal` file has pointed to
https://github.com/commercialhaskell/stack for a long time, as do the
`.deb`s and `.rpms`.

>

DomT4: Cool. Thanks for the clarification!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/5f8969f3696a076946ac68191f31bb1f640e4ce3. Thank you for another Homebrew contribution @borsboom! :smiley_cat:

PR #46220, opened 2015-11-20
wine 1.8-rc1（devel）

uyjulian:


uyjulian:
all fine on my side √

DomT4:
Why the move away from Sourceforge here? Is that an upstream thing?

uyjulian:
wine switched to dl.winehq.org in the release notes: https://www.winehq.org/announce/1.8-rc1

DomT4:
Cool, don't mind the change away if that's an upstream thing. Thanks for restoring the `https` upstream link!

uyjulian:
It seems like there is still new versions published on the sourceforge. Oh, well I'll include that as a mirror anyway.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/2bbfd03937b8cd2c612da64508f52443938422ba. Thank you for another Homebrew contribution @uyjulian! :smiley_cat:

PR #46224, opened 2015-11-20
New formula: dockviz 0.3

waj:


Library/Formula/dockviz.rb
```diff
@@ -0,0 +1,19 @@
...
+  desc "Visualizing docker data"
+  homepage "https://github.com/justone/dockviz"
+  url "https://github.com/justone/dockviz.git", :tag => "v0.3", :revision => "ae855d6a5a540c3563948e7035f9d19cab17f5b0"
+  head "https://github.com/justone/dockviz.git"
+
+  depends_on "go" => :build
+
+  def install
+    ENV["GOPATH"] = buildpath
+    system "go", "get", "github.com/fsouza/go-dockerclient"
```
DomT4: Please use `go_resource`s rather than `go get`, Thanks. We pretty heavily frown on remote fetching happening once we're in the `def install` execution process.

There's a bunch of examples available by searching for `go_resource` if [this one](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/consul-template.rb) doesn't fit your purpose :)

Library/Formula/dockviz.rb
```diff
@@ -0,0 +1,19 @@
+class Dockviz < Formula
+  desc "Visualizing docker data"
+  homepage "https://github.com/justone/dockviz"
+  url "https://github.com/justone/dockviz.git", :tag => "v0.3", :revision => "ae855d6a5a540c3563948e7035f9d19cab17f5b0"
```
DomT4: Please leave `tag` and `revision` on their own lines. Helps readability a bit, Thanks!

waj:
@DomT4 I just updated the formula after your comments. Thanks!

Library/Formula/dockviz.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def install
+    ENV["GOPATH"] = buildpath
+    Language::Go.stage_deps resources, buildpath/"src"
+    system "go", "build", "-o", bin/"dockviz"
+  end
+
+  test do
+    system "#{bin}/dockviz", "--version"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

waj: This tool can be used to visualize docker containers or images, but not necessarily running on the local machine, that's why I didn't add a dependency with the "docker" formula. And I don't see any way of testing that the installation was successfull other than testing to a real Docker host.

mikemcquaid:

PR #46225, opened 2015-11-20
saltstack 2015.8.1

mpolichette:
Hey guys,

I'm updating saltstack on homebrew to version `2015.8.1` to keep brew fresh and to solve an issue with the local salt grain on OS X being reported as proxy instead of MacOS


dunn:
Thank you! Merged in 1686b6307f6e923929c6203d565988bcc7500a37.

PR #46226, opened 2015-11-20
openvpn: only depend on tuntap on pre-yosemite os-x

lacombar:
Signed-off-by: Arnaud Lacombe <lacombar@gmail.com>

### DIFF
--- a/Library/Formula/openvpn.rb
+++ b/Library/Formula/openvpn.rb
@@ -12,7 +12,7 @@ class Openvpn < Formula
   end
 
   depends_on "lzo"
-  depends_on :tuntap
+  depends_on :tuntap if MacOS.version < :yosemite
   depends_on "openssl"
 
   def install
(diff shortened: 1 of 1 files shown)


Library/Formula/openvpn.rb
```diff
@@ -42,11 +45,17 @@ def install
     rm "#{share}/doc/openvpn/README.polarssl"
   end
 
-  def caveats; <<-EOS.undent
-    If you have installed the Tuntap dependency as a source package you will
-    need to follow the instructions found in `brew info tuntap`. If you have
-    installed the binary Tuntap package, no further action is necessary.
+  def caveats
+    if MacOS.version < :yosemite
+      puts <<-EOS.undent
```
DomT4: You can use `s` instead of `puts`. See [Node](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/node.rb#L116-L136) for an example.

Library/Formula/openvpn.rb
```diff
@@ -42,11 +45,17 @@ def install
     rm "#{share}/doc/openvpn/README.polarssl"
   end
 
-  def caveats; <<-EOS.undent
-    If you have installed the Tuntap dependency as a source package you will
-    need to follow the instructions found in `brew info tuntap`. If you have
-    installed the binary Tuntap package, no further action is necessary.
+  def caveats
+    if MacOS.version < :yosemite
```
mistydemeo: Instead of `puts`, assign these to a string and append to it:

```ruby
def caveats
  s = ""

  if MacOS.version < :yosemite
    s += <<-EOS.undent
    If you have installed the Tuntap dependency...
    EOS
  end

  s += <<-EOS.undent
  For OpenVPN to work as a server...
  EOS
end

File: Library/Formula/openvpn.rb [hunk 3]
```diff
@@ -12,9 +12,12 @@ class Openvpn < Formula
   end
 
   depends_on "lzo"
-  depends_on :tuntap
   depends_on "openssl"
 
+  if MacOS.version < :yosemite
```
mistydemeo: Can make this a one-liner:

```ruby
depends_on :tuntap if MacOS.version < :yosemite
```

lacombar: will do.

PR #46232, opened 2015-11-20
boot2docker: obsoleted by docker-machine

kad:
Boot2docker-cli obsoleted by docker-machine. no sense to keep that formula.


DomT4:
5200 people have downloaded bottles for the obsolete `boot2docker` in the last 30 days :wink:. I'm semi-reluctant to kill off a formula still being used heavily by people, but I'll let other maintainers discuss it.

mikemcquaid:
The download count is good enough reason for me.

kad:
ok. however, official site recommends to switch to docker-machine asap: see notes on last release of boot2docker-cli: https://github.com/boot2docker/boot2docker-cli/releases

potentially we need to look at counter for docker-machine, as it might be the case where people downloading something that was found in old docs, tried to used and found out that it's not usable with recent versions of docker client. Also, maybe alias boot2docker -> docker-machine would be useful.

Or maybe there are other ways to gradually deprecate that formula ?

bfontaine:
I personally don’t think we should make such choice for our users; some of them may have good reasons to continue using `boot2docker` instead of `docker-machine`. We could maybe put something in the caveats but we usually reserve these for Homebrew-specific issues.

DomT4:
Yeah, I'd be happy enough to see something in the caveats about upstream recommending using xyz instead, but till that download number drops significantly don't really want to force the decision on end users who may still have a legitimate use case.

mikemcquaid:
> Yeah, I'd be happy enough to see something in the caveats about upstream recommending using xyz instead, but till that download number drops significantly don't really want to force the decision on end users who may still have a legitimate use case.

Agreed :+1:

Issue #44036, opened 2015-09-16
vera++ failed to build on 10.11: failed resource download


apjanke:
https://gist.github.com/cc16bcd79a947dd89a07

It's getting a SHA mismatch for one of the resources it downloads.

```
-- downloading...
     src='https://github.com/verateam/luabind/archive/vera-1.3.0.zip'
     dst='/tmp/vera++20150916-1700-1qegp0c/vera++-1.3.0/src/luabind-prefix/src/vera-1.3.0.zip'
     timeout='none'
-- [download 0% complete]
[...snip...]
-- downloading... done
CMake Error at luabind-stamp/verify-luabind.cmake:27 (message):
  error: MD5 hash of

    /tmp/vera++20150916-1700-1qegp0c/vera++-1.3.0/src/luabind-prefix/src/vera-1.3.0.zip

  does not match expected value

    expected: c5780d1ca562183d64be496159ded07f
      actual: 3a812ff2de5d5befdd43e80d8422f1bd



make[2]: *** [src/luabind-prefix/src/luabind-stamp/luabind-download] Error 1
make[1]: *** [src/CMakeFiles/luabind.dir/all] Error 2
make: *** [all] Error 2
```

DomT4:
It's fetching external code in the compile process? :eyes:.

PR #46234, opened 2015-11-21
vera++: fix build error in 1.3.0

kersson:
The error in #44036 was fixed in the vera source in commit [887350b](https://bitbucket.org/verateam/vera/commits/887350b1cb5ca35af940abaded33f85596d26104).  I applied this patch to v1.3.0 source and it built fine.

### DIFF
--- a/Library/Formula/vera++.rb
+++ b/Library/Formula/vera++.rb
@@ -19,9 +19,60 @@ class Veraxx < Formula
     sha256 "122a15e33a54265d62a6894974ca2f0a8f6ff98742cf8e6152d310cc23099400"
   end
 
+  # Custom-built boost, lua, and luabind are used by the build scripts
+  resource "boost" do
+    url "https://downloads.sourceforge.net/project/boost/boost/1.56.0/boost_1_56_0.tar.bz2"
+    sha256 "134732acaf3a6e7eba85988118d943f0fa6b7f0850f65131fff89823ad30ff1d"
+  end
+
+ resource "lua" do
+   url "https://github.com/LuaDist/lua/archive/5.2.3.tar.gz"
+   sha256 "c8aa2c74e8f31861cea8f030ece6b6cb18974477bd1e9e1db4c01aee8f18f5b6"
+ end
+
+ resource "luabind" do
+   url "https://github.com/verateam/luabind/archive/vera-1.3.0.tar.gz"
+   sha256 "7d93908b7d978e44ebe5dfad6624e6daa033f284a5f24013f37cac162a18f71a"
+ end
+
   def install
-    # don't use system lua to avoid a dependency on lua 5.1
-    system "cmake", ".", "-DVERA_USE_SYSTEM_LUA:BOOL=OFF", *std_cmake_args
+    resource("boost").stage do
+      system "./bootstrap.sh", "--prefix=#{buildpath}/3rdParty",
+             "--with-libraries=filesystem,system,program_options,regex,wave,python"
+      system "./b2", "install", "threading=multi", "link=static", "warnings=off",
+             "cxxflags=-DBOOST_WAVE_SUPPORT_MS_EXTENSIONS=1", "-s NO_BZIP2=1"
+    end
+
+    resource("lua").stage do
+      args = std_cmake_args
+      args << "-DBUILD_SHARED_LIBS:BOOL=OFF"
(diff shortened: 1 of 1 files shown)


DomT4:
I don't really want to fix *just that* here. I'd also want to fix the fact their compile script remotely fetches things during the compile process which makes it impossible to cache, and then verifies them with an *MD5sum* of all things. Remote fetching during compile is a bad idea.

I'd like to go down the route of turning off the remote fetching (It looks like there's an option in the script you've patched there) and vendoring the things it wants to use as `resources` and installing into either the `buildpath` or `libexec` and getting it to use those libraries.

That way, it's still using the libraries it wants to use, but all the downloading is done prior to compile starting and verified against a hash known to not be completely exploitable and broken.

kersson:
Sounds good.  Sorry for not doing that first.  New to contributing and didn't thoroughly read the [Acceptable Formula section](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Acceptable-Formulae.md#we-dont-like-install-scripts-that-download-things) on the matter.

The three libraries that it depends on that it currently downloads are `lua`, `luabind` and `boost`, all of which are provided by `brew` so I first tried just adding them as dependencies.  However, it turns out that due to a [difference in behavior](https://bitbucket.org/verateam/vera/issues/73/ruledump-test-fails-on-rhel7-when-built-in) in `boost_wave` between unix and Windows, Boost needs to be compiled using [different compile flags](https://bitbucket.org/verateam/vera/src/3cd00f6f6d6102b09e070bc59e2c35d9808a9dcb/src/boost.cmake?at=master&fileviewer=file-view-default#boost.cmake-85).

So my current plan is to add the libraries as `resources` as you suggested.  However, there is no `cmake` argument to allow me to specify a local path to the `lua`, `luabind`, and `boost` source archives.  The URLs are hard-coded into the cmake scripts ([here](https://bitbucket.org/verateam/vera/src/69f28c5a505294e29a3e95ed0998192fa18b0f08/src/lua.cmake?at=maint&fileviewer=file-view-default#lua.cmake-32), [here](https://bitbucket.org/verateam/vera/src/69f28c5a505294e29a3e95ed0998192fa18b0f08/src/lua.cmake?at=maint&fileviewer=file-view-default#lua.cmake-67), and [here](https://bitbucket.org/verateam/vera/src/69f28c5a505294e29a3e95ed0998192fa18b0f08/src/boost.cmake?at=maint&fileviewer=file-view-default#boost.cmake-74)), so I'll have to patch the scripts to add `cmake` arguments to pass the source file paths in.

Let me know if you see any issues with this plan.  Thanks!

DomT4:
It looks like you can just pass `-DVERA_USE_SYSTEM_LUA=ON` and `-DVERA_USE_SYSTEM_BOOST=ON` to the `cmake_args`? That'd negate the need to patch.

You'd add the resources essentially, and then install them into `libexec/vendor` or such, and then prepend the $PATH/CFLAGS/whatever other variables are necessary in the formula and it should find the vendored "system" versions and use them.

DomT4:
[Example of that resource handling here if useful](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/libbtbb.rb#L36-L46).

kersson:
Well, I was trying to avoid having to duplicate the building of the three libraries in the formula, and instead just tell `cmake` where to find a local copy of the sources (instead of downloading them), but that requires patching the `cmake` scripts.

Either way seems messy.  Sounds like having the formula build the libraries instead is the way to go here?  I'm happy to do whatever is more canonical.

DomT4:
Essentially the issue with patching is that we ask for patches to be submitted upstream. If you wanted to patch and then submit that upstream with an explanation of remote fetching being a pain for package management feel free to go down that route. Either method works and is acceptable IMO, so go for whatever you're comfortable with.

kersson:
Here's another thought.  Since `lua`, `luabind`, and `boost` are all provided by `brew`, is there any way for me to leverage those formulas and have them built with the custom flags `vera++` needs, have them installed into `buildpath` or any other temporary location, so the `vera++` build script can static link against them?

DomT4:
Not in any sort of practical manner, I'm afraid.

kersson:
How's it look?

mikemcquaid:

DomT4:
Nice work here @kersson :heart:

kersson:
Thanks guys, and thanks for the assistance @DomT4. Homebrew is awesome. Happy that I could give back.

PR #46235, opened 2015-11-21
gmp 6.1.0

dunn:
Not available on the mirrors yet ¯\\(´o｀)/¯

### DIFF
--- a/Library/Formula/gmp.rb
+++ b/Library/Formula/gmp.rb
@@ -1,10 +1,8 @@
 class Gmp < Formula
   desc "GNU multiple precision arithmetic library"
   homepage "https://gmplib.org/"
-  url "http://ftpmirror.gnu.org/gmp/gmp-6.0.0a.tar.bz2"
-  mirror "https://gmplib.org/download/gmp/gmp-6.0.0a.tar.bz2"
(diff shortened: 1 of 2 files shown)


dunn:
john-jumbo _really_ wants to write to `~`; this is the only way I could figure out how to stop it, smh.  Other proposals very welcome.

DomT4:
Yeah, it uses an alternative language-specific method to determine `$HOME` that doesn't involve actually checking the `$HOME` env, much like Java is prone to doing. I'd probably just ditch the test for now, rather than forcing an unnecessary rebuild on users.

dunn:
Why is the 10.9 test installing things from Versions? https://travis-ci.org/Homebrew/homebrew/jobs/92458524

DomT4:
We tap `versions` to get the special Ruby installation, so I guess the CI just builds against anything it can find that uses the formula as a dep, maybe?

dunn:
Oh, makes sense.

DomT4:
Tried putting it back through Travis but looks like it's still using the travis file at the time of commit. The Travis build of the `.travis` configuration change passed fine though, so presume this would be alright.

Up to you whether you want to rebase and let Travis poke it again or just merge off the Jenkins build being green.

dunn:
May as well, thanks for figuring out the fix!

Issue #45812, opened 2015-11-08
[Formula Request] gsl 2.0


ambs:
Hello,

Not sure how you want to deal with libgsl 2.0. Breaking the current formula doesn't seem like a good choice, so probably create a new one named gsl2? looking to the libgsl 1.6 formula, it should be easy to produce a new formula in case it compiles cleanly. I can try doing it, although my experience is quite limited.

mikemcquaid:
Generally we try and migrate all the current formulae and then add it when it's all ported over. If you need it before that then adding `libgsl2` to `homebrew/versions` is probably a good idea. Thanks!

ambs:
Thank. Will try!

tschoonj:
Personally I think we should create a new formula `gsl2` and add it to `core` right away. It won't take long before formulas become compatible with this new version and they shouldn't depend on a formula in `versions`. All I hope is that the GSL developers provided support for separate `gsl` and `gsl2` installations...

DomT4:
> and they shouldn't depend on a formula in versions

We don't accept mandatory cross-tap dependencies in the core, so it's more a case of *won't* than *shouldn't* :smile:.

tschoonj:
Had a quick look at `gsl` 2.0 and it looks bad. They provided absolutely no support for simultaneous installations of 1.16 and 2.0: they use the same pkg-config file and they didn't even bother to get their libtool version numbering right for an API breakage, even though the right thing to do would have been renaming the libtool library.

I have a feeling there will be an update soon, as Linux distro packagers will quickly complain about this situation: they will need to support both versions for a long time to come (as do we), which right now can be achieved only after heavy patching.

mikemcquaid:
Sounds like it's worth just holding off for now, then.

DomT4:
Feel free to file a PR in versions with a `keg_only` addition of the new `gsl` 2.0. We can look again at the core when it wouldn't break things here.

PR #46237, opened 2015-11-21
gsl 2.1

dunn:


tschoonj:
see #45812

This new version is a disaster IMHO and will break probably every single formula that depends on it.

dunn:
Oh yeah, forgot about that. Thanks!

bfontaine:
@tschoonj #45812 is about gsl 2.0; is gsl 2.1 affected by these issues too?

tschoonj:
yep it is

If you really want to provide support for this release, I suggest you create a formula `gsl2`, and explicitly mark it as conflicting with the original `gsl`

PR #46243, opened 2015-11-21
Docs: deprecate homebrew/head-only

bfontaine:


share/doc/homebrew/Interesting-Taps-&-Branches.md
```diff
@@ -22,7 +22,7 @@ Homebrew has the capability to add (and remove) multiple taps to your local inst
 
 *   [homebrew/games](https://github.com/Homebrew/homebrew-games): Game or gaming-emulation related formulae.
 
-*   [homebrew/head-only](https://github.com/Homebrew/homebrew-head-only): A tap for brews that only have unstable, unreleased versions.
+*   [homebrew/head-only](https://github.com/Homebrew/homebrew-head-only): A tap for brews that only have unstable, unreleased versions. This tap is deprecated and doesn’t accept new formulae.
```
xu-cheng: You may want to boldface the deprecate part.

xu-cheng:
Apart of document, I think we should remove `head-only` from `official_taps.rb`

mikemcquaid:
> Apart of document, I think we should remove head-only from official_taps.rb

:+1:

share/doc/homebrew/Acceptable-Formulae.md
```diff
@@ -66,9 +66,8 @@ Formulae in the core repository must have a stable version tagged by
 the upstream project. Tarballs are preferred to git checkouts, and
 tarballs should include the version in the filename whenever possible.
 
-Software that does not provide a stable, tagged version, or had guidance to
-always install the most recent version, should be put in
-[homebrew/head-only](https://github.com/Homebrew/homebrew-head-only) or [homebrew/devel-only](https://github.com/Homebrew/homebrew-devel-only).
+Software that does not provide a stable, tagged version should be put in
```
mikemcquaid: Might want to say `Software that only provides a development/beta, tagged version should be put in devel-only. Software without a tagged version cannot be accepted by Homebrew.`

DomT4: I'd be tempted to add a reason there as well, such as it providing terrible UX and quickly becoming outdated.

Probably won't stop many people trying to get `head-only` stuff into the core again, but gives us justification to point to in the docs if it happens.

bfontaine: What about this?

> Software that only provides a development/beta, tagged version should be put in
> devel-only. We don’t accept software without a tagged version because it
> provides a terrible UX and quickly becomes outdated.

DomT4: Sounds good. I'd be tempted to just make `development/beta` into `unstable`, but I don't think it makes much difference.

On the terrible UX I'd probably give reasons, partially so if we all get hit by a truck the next set of maintainers know what we thought about fixing, and partially because someone's bound to ask.

Just something like there being no automatic update mechanism for `head-only` formulae which makes them very quickly outdated, they regularly break due to upstream changes, we can't provide bottles for them, etc, maybe.

bfontaine: Changed to:

> We don’t accept software without a tagged version because it regularly break due to upstream changes; we can’t provide [bottles](Bottles.md) for them; and we don’t have an automatic update mechanism for `head-only` formulae which makes them very quickly outdated.

DomT4: World's smallest nit, but someone's going to point it out if I don't - break should be pluralised in the first sentence.

Otherwise looks perfect to me. Thanks Baptiste.

dunn: Or "they regularly break" 🙃

bfontaine: I went with “they regularly break”  to be consistent with the whole sentence. Thank you both; don’t hesitate if something sounds strange or if there’s a typo; I’m open to even the smallest nits :wink:

share/doc/homebrew/Interesting-Taps-&-Branches.md
bfontaine: Thank you; Done.

Issue #46101, opened 2015-11-18
Build GraphicsMagick with quantum-depth 16 by default


carandraug:
The default configure option to build GraphicsMagick is quantum depth 8. This limits GraphicsMagick to always read images with that bitdepth. This can be quite limiting since GraphicsMagick is then used by scientific applications where 16bit images are very common.

There is a very long discussion on a debian bug report [#557879](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=557879) about the subject. I will only quote [one the comments](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=557879#32) by Bob Friesenhahn, the maintainer of GraphicsMagick:

>> Bob Friesenhahn, the upstream GraphicsMagick maintainer, also recommends 
>> building GraphicsMagick with a 16-bit quantum depth [1]:
>
> It has been a while since 2004, but today that is still my 
> recommendation if (almost) all the capabilities are needed.  Of course 
> there is a memory penalty, which becomes more severe for people 
> dealing with high-resolution monochrome files like faxes.
> 
> Today's computers come with quite a lot more RAM and people have 
> become used to dealing with raw digital camera files and 16-bit/sample 
> TIFF files.
> 
>>> If you were to offer just one set of binaries,
>>> the 16bit version would be a better offering
>>> since it is clearly superior for some tasks
>>> and there is more assurance that quality won't
>>> be lost.
>>
>> Bob also mentioned the 16-bit build is the one he spends most time optimizing 
>> [2].
>
> That is true and is a reason that the performance of the 16-bit build 
> does not lag as much behind the 8-bit build as it used to do.  In the 
> GraphicsMagick 1.2 timeframe, considerable effort was put into being 
> able to import and export from the 16-bit samples with the best 
> performance.

Can the default in homebrew be changed? I can submit a patch but need to know first if that's not a problem.

DomT4:
Don't have a problem with the default being changed personally.

carandraug:
Does b78da5c416c7ba8e7ccd7312b4c9e285ce3d5a57 looks acceptable then?

DomT4:
Oh, actually, looking at the formula it looks like there is *no default* currently?

```ruby
    if build.with? "quantum-depth-32"
      quantum_depth = 32
    elsif build.with? "quantum-depth-16"
      quantum_depth = 16
    elsif build.with? "quantum-depth-8"
      quantum_depth = 8
    end

[...]

PR #46248, opened 2015-11-21
graphicsmagick: set default quantum-depth to 16 (issue #46101)

carandraug:
See original discussion on the original issue #46101

### DIFF
--- a/Library/Formula/graphicsmagick.rb
+++ b/Library/Formula/graphicsmagick.rb
@@ -12,7 +12,7 @@ class Graphicsmagick < Formula
   end
 
   option "with-quantum-depth-8", "Compile with a quantum depth of 8 bit"
-  option "with-quantum-depth-16", "Compile with a quantum depth of 16 bit"
+  option "with-quantum-depth-16", "Compile with a quantum depth of 16 bit (default)"
   option "with-quantum-depth-32", "Compile with a quantum depth of 32 bit"
   option "without-magick-plus-plus", "disable build/install of Magick++"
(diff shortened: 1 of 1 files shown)


schoeps:
Now the user could specify depth-8, depth-16 and depth-32 at the same time without an error message. That's not ideal. At least a check for conflict should be made. Some other formulae solve this problem such that you have to disable the default setting manually "--without-quantum-depth-16 --with-quantum-depth-32"  (e.g. I did this for gmsh: https://github.com/Homebrew/homebrew-science/blob/master/gmsh.rb, line 33f)

tdsmith:
As a microscopist I'm quite sympathetic; I agree with @schoeps. The dependents are mostly in homebrew/science, so queueing a PR there (to be rebuilt after this is pulled here) would be reasonable:

```
$ brew uses graphicsmagick
homebrew/science/bioformats-cpp
homebrew/science/dgtal
homebrew/science/gnudatalanguage
homebrew/science/octave
homebrew/science/vips
homebrew/science/vips7
homebrew/x11/xastir
```

DomT4:
Closing as dead. Please feel free to resubmit a fresh PR as desired with Tim's comment in mind. Thanks for your submission to Homebrew!

carandraug:
@DomT4 Tim's comment was to note other recipes that they will have to update once this gets accepted. But this has not been accepted yet. Will it be if I inform `homebrew/science` and `homebrew/x11` of this?

Note that this pull request came after issues on `homebrew/science` and is why @schoeps is commenting here.

PR #46250, opened 2015-11-21
perl: disable make test on El Capitan

### LINKED ISSUE #41716: perl --with-tests fails to build on 10.11
[gist](https://gist.github.com/aw1621107/07b72067fc236ef0af5d)

Tests seem to be looking for a library in the perl install before install is linked? Failed both from a normal install and reinstall. Library was confirmed to exist in the location the test was expecting it on the reinstall, strangely enough...

DomT4:
SIP kills the ability for `make test` to pass. Have reported the
issue upstream and included a link in the formula.

Closes #41716.

### DIFF
--- a/Library/Formula/perl.rb
+++ b/Library/Formula/perl.rb
@@ -20,14 +20,14 @@ class Perl < Formula
   option "with-tests", "Build and run the test suite"
 
   def install
-    args = [
-      "-des",
-      "-Dprefix=#{prefix}",
-      "-Dman1dir=#{man1}",
-      "-Dman3dir=#{man3}",
-      "-Duseshrplib",
-      "-Duselargefiles",
-      "-Dusethreads"
+    args = %W[
+      -des
+      -Dprefix=#{prefix}
+      -Dman1dir=#{man1}
+      -Dman3dir=#{man3}
+      -Duseshrplib
+      -Duselargefiles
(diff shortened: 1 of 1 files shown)


Library/Formula/perl.rb
```diff
@@ -35,7 +35,15 @@ def install
...
     system "./Configure", *args
     system "make"
-    system "make", "test" if build.with?("tests") || build.bottle?
+
+    # OS X El Capitan's SIP feature prevents DYLD_LIBRARY_PATH from being passed to child
+    # processes, which causes the make test step to fail.
+    # https://rt.perl.org/Ticket/Display.html?id=126706
+    # https://github.com/Homebrew/homebrew/issues/41716
+    if MacOS.version < :el_capitan
+      system "make", "test" if build.with?("tests") || build.bottle?
```
bfontaine: You could combine both `if`s here.

DomT4: Just wanted to try and keep line length reasonable really.

```ruby
system "make", "test" if MacOS.version < :el_capitan && build.with?("tests") || build.bottle?
```
Felt a bit hefty.

DomT4: Will merge as-is, but if it bothers you I won't sulk if you want to change it :smile:.

PR #45895, opened 2015-11-11
evince 3.18.2

tschoonj:

PR #45845, opened 2015-11-09
gucharmap 3.18.2

tschoonj:

PR #46266, opened 2015-11-22
itstool: removed :python dependency

tschoonj:


DomT4:
Probably still needs it on `snow_leopard` and below.

DomT4:
Also possibly worth bumping the revision so we don't end up with some users still having it linked to Homebrew's Python and some to the system, etc. Sorry for not thinking of that earlier :sweat:.

tschoonj:
@DomT4 done!

DomT4:
Part of me is slightly worried this is somehow going to find a new and interesting way to blow up in our faces, but fingers crossed all goes smoothly. It passed the CI at least.

Merged in https://github.com/Homebrew/homebrew/commit/ad561e85536c588295b6bac52d4853d1501e9d4d. Thanks once more Tom!

tschoonj:
Thanks for merging it in.

To see if it works properly, could you restart my remaining PRs? #45785 #45845 #45932 #45895 

Thanks!

PR #46269, opened 2015-11-22
bin/brew: improve code quality and optimize

UniqMartin:
One motivation is code quality improvements (obviously according to what I believe to be good shell code). The other goal is to improve the startup time (the contribution from the shell wrapper). Thus, to benchmark the improvement in startup time, I replaced the last line `exec "$HOMEBREW_RUBY_PATH" -W0 "$BREW_LIBRARY_DIRECTORY/brew.rb" "$@"` with

```
echo "$HOMEBREW_BREW_FILE"
echo "$HOMEBREW_RUBY_PATH"
```

and then recorded the timings for four consecutive invocations of `time brew`. (I was delighted to see an actually measurable improvement.)

Before:

```
real   0m0.018s  0m0.015s  0m0.014s  0m0.013s
user   0m0.007s  0m0.007s  0m0.007s  0m0.007s
sys    0m0.011s  0m0.010s  0m0.009s  0m0.009s
```

After:

```
real   0m0.009s  0m0.009s  0m0.010s  0m0.010s
user   0m0.004s  0m0.004s  0m0.004s  0m0.004s
sys    0m0.005s  0m0.006s  0m0.006s  0m0.005s
```

**Reviewers:** Please look at the individual commits, to see meaningful diffs and read the commit messages that contain the rationale for every change. I'm happy with dropping cosmetic changes, though I hope my arguments are convincing and we can get the full package in. :)

### DIFF
--- a/bin/brew
+++ b/bin/brew
@@ -3,18 +3,18 @@ chdir() {
   cd "$@" >/dev/null
 }
 
-BREW_FILE_DIRECTORY=$(chdir "${0%/*}" && pwd -P)
+BREW_FILE_DIRECTORY="$(chdir "${0%/*}" && pwd -P)"
 export HOMEBREW_BREW_FILE="$BREW_FILE_DIRECTORY/${0##*/}"
 
-BREW_SYMLINK=$(readlink "$0")
+BREW_SYMLINK="$(readlink "$0")"
 if [ -n "$BREW_SYMLINK" ]
 then
-  BREW_SYMLINK_DIRECTORY=$(dirname "$BREW_SYMLINK")
-  BREW_FILE_DIRECTORY=$(chdir "$BREW_FILE_DIRECTORY" &&
-                        chdir "$BREW_SYMLINK_DIRECTORY" && pwd -P)
+  BREW_SYMLINK_DIRECTORY="$(dirname "$BREW_SYMLINK")"
+  BREW_FILE_DIRECTORY="$(chdir "$BREW_FILE_DIRECTORY" &&
+                         chdir "$BREW_SYMLINK_DIRECTORY" && pwd -P)"
 fi
 
-BREW_LIBRARY_DIRECTORY=$(chdir "$BREW_FILE_DIRECTORY"/../Library && pwd -P)
+BREW_LIBRARY_DIRECTORY="$(chdir "$BREW_FILE_DIRECTORY"/../Library && pwd -P)"
 
 # Users may have these set, pointing the system Ruby
 # at non-system gem paths
@@ -28,7 +28,7 @@ fi
 
 if [ -z "$HOMEBREW_RUBY_PATH" ]
 then
-  BREW_SYSTEM=$(uname -s | tr "[:upper:]" "[:lower:]")
+  BREW_SYSTEM="$(uname -s | tr "[:upper:]" "[:lower:]")"
   if [ "$BREW_SYSTEM" = "darwin" ]
   then
     export HOMEBREW_RUBY_PATH="/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby"
--- a/bin/brew
+++ b/bin/brew
@@ -4,7 +4,7 @@ chdir() {
 }
 
 BREW_FILE_DIRECTORY="$(chdir "${0%/*}" && pwd -P)"
-export HOMEBREW_BREW_FILE="$BREW_FILE_DIRECTORY/${0##*/}"
+HOMEBREW_BREW_FILE="$BREW_FILE_DIRECTORY/${0##*/}"
 
 BREW_SYMLINK="$(readlink "$0")"
 if [ -n "$BREW_SYMLINK" ]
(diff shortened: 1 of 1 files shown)


bfontaine:
Looks great! :+1: 

By the way, do we have any reason not to use `if [ "$(uname -s)" = "Darwin" ]` instead of the existing code below?
```sh
 BREW_SYSTEM="$(uname -s)"
  if [ "$BREW_SYSTEM" = "Darwin" ]
```

UniqMartin:
Previously, it would have resulted in a pretty long and convoluted line. But now your suggestion makes perfect sense and there's no technical reason not to do it. Since I'm touching both lines in the commit anyway, I should simply merge them. Thanks for the hint! (I have pushed an update incorporating this.)

mikemcquaid:
All makes sense to me :+1:

UniqMartin:
@xu-cheng I went with your suggestion in https://github.com/UniqMartin/homebrew/commit/8cd4da6b6ac2055a3735f7f4c097a4c84a9598af as I couldn't find a way to break it even with very crazy symlinking of both files and directories. Hence the update I just pushed.

**RFC:** I found another opportunity to optimize (with a measurable impact) by completely getting rid of `uname`. I haven't pushed that one yet. If we assume that `/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby` only exists on OS X, we can use this to drive our decision in the shell wrapper (instead of basing our decision on `uname -s`). Thus, we could replace

```shell
if [ "$(uname -s)" = "Darwin" ]
then
  HOMEBREW_RUBY_PATH="/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby"
else
  HOMEBREW_RUBY_PATH="$(which ruby)"
fi
```

with

```shell
HOMEBREW_RUBY_PATH="/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby"
if [ ! -x "$HOMEBREW_RUBY_PATH" ]
then
  HOMEBREW_RUBY_PATH="$(which ruby)"
fi
```

This can only break if Apple decides to drop Ruby or to relocate it to a completely different location, but even then there would be a fallback to `which ruby`. Is this a good idea or too radical and obscure?

mikemcquaid:
> This can only break if Apple decides to drop Ruby or to relocate it to a completely different location, but even then there would be a fallback to which ruby. Is this a good idea or too radical and obscure?

The thing that's more likely to break this is users deleting that.

xu-cheng:
> I went with your suggestion in UniqMartin@8cd4da6 as I couldn't find a way to break it even with very crazy symlinking of both files and directories. Hence the update I just pushed.

I did some test myself. I can confirm you are right.

> This can only break if Apple decides to drop Ruby or to relocate it to a completely different location, but even then there would be a fallback to which ruby. Is this a good idea or too radical and obscure?

It feels too hacky to me. And personally, I'm not in too much favor on such non-intuitive logic. Also FYI, this logic will be dramatic changed for future ruby 2.0 move. It will be looked like something like this:

```bash
if [ "$(uname -s)" = "Darwin" ]
then
  if [ -d "/System/Library/Frameworks/Ruby.framework/Versions/1.8"]
  then
    HOMEBREW_RUBY_PATH="<path to vendor ruby>"
  else
    HOMEBREW_RUBY_PATH="/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby"
  fi
else
  HOMEBREW_RUBY_PATH="$(which ruby)"
fi
```

xu-cheng:
Rethink a bit. I found @UniqMartin's proposal does have one advantage. That is it's more resistant to any possible Apple changes. So we can prevent homebrew break similar to last time when Apple moved default ruby to 2.0.

mikemcquaid:
> Rethink a bit. I found @UniqMartin's proposal does have one advantage. That is it's more resistant to any possible Apple changes. So we can prevent homebrew break similar to last time when Apple moved default ruby to 2.0.

What if Apple moves the `Current` symlink to e.g. Ruby 3 which isn't backwards compatible?

xu-cheng:
> What if Apple moves the Current symlink to e.g. Ruby 3 which isn't backwards compatible?

Homebrew will be broken with or without @UniqMartin's proposed change.

mikemcquaid:
Sorry, I meant the other way around: if it's removed by users and their `ruby` in their PATH is not a compatible version. I think it's better that we error in that case.

Hopefully all this stuff will be handled by a Bash `brew update` wrapper in future so people can at least always run `brew update` to get onto the latest version.

mikemcquaid:
To be clear though: I do :+1: the current version and would like it to be merged.

UniqMartin:
Thanks for the great discussion! Let's get this in as-is and revisit the issue of picking a Ruby at a later stage, e.g., when we actually move to Ruby 2.0 for all users.

> Sorry, I meant the other way around: if it's removed by users and their `ruby` in their PATH is not a compatible version. I think it's better that we error in that case.

This could be somewhat mitigated by having a wrapper on the Ruby side as well. What I mean by this is that `Library/brew.rb` would only perform the bare minimum of checks to be able to give the user a helpful error message (maybe only check environment, Ruby version, and existence of crucial files). Only after performing those checks it would `require` a file that contains what is currently found in `Library/brew.rb`.

This wrapper would be implemented with forward and backward compatibility in mind (and thoroughly tested, too), making it run on as many Rubies as is reasonably possible. And I hope Ruby 3 won't be such a radical change as to break it.

PR #46273, opened 2015-11-23
charm-tools 1.9.2

marcoceppi:
This is a feature and bug fix release, deprecating 1.8.0 and 1.9.1

### DIFF
--- a/Library/Formula/charm-tools.rb
+++ b/Library/Formula/charm-tools.rb
@@ -1,8 +1,8 @@
 class CharmTools < Formula
   desc "Tools for authoring and maintaining juju charms"
   homepage "https://github.com/juju/charm-tools"
-  url "https://github.com/juju/charm-tools/releases/download/v1.8.0/charm-tools-1.8.0.tar.gz"
-  sha256 "e72e788f069fcf4c2db7435db71ca8feb3dd8c6fd63c4b7e591b4064d76f4dbb"
+  url "https://github.com/juju/charm-tools/releases/download/v1.9.2/charm-tools-1.9.2.tar.gz"
+  sha256 "2a1c8b0066d3f6c527d6b8d273c7def20800f7539d4c82488c84d97f92364161"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


marcoceppi:
Travis seems to be experiencing a networking hiccup, hence the failed test.

mistydemeo:
@BrewTestBot test this please

marcoceppi:
I can push a non-changing commit to the branch to kick travis again if it helps push this through. @BrewTestBot and 2/3 travis tests signed off on this.

DomT4:
Wanted to raise a point from the other thread you closed earlier. The Travis build reveals remote fetching of undeclared Python modules during compile, and as much was mentioned in https://github.com/Homebrew/homebrew/pull/46213.

I'm a touch confused by https://github.com/juju/charm-tools/commit/2a70207ee8d0a2cdd4a07c9c4c5d8936f5da7aa0 - Homebrew is capable of handling Python fetching and installing into the correct place; we do so regularly for a whole bunch of formulae. `jrnl` is the one we point to in the Python documentation.

I'm not sure what is stopping the use of our usual resource model here? We don't permit `pip install` directly because it's fairly bad UX for the end user (If something goes wrong, they have to fetch a bunch of resources again afresh rather than just using cached ones; the resource version used can differ between the bottles and source installs which makes debugging more troublesome, etc etc) but we permit resources.

The end result for users is the same as doing `pip install` with a changed `PYTHONPATH`.

marcoceppi:
The fact that it's tedious, python and setup.py already manages the dependency chain of requirements for this project. To breakdown each one - and each one's dependencies, would mean I'd be defining upwards of 30+ resources in this file. Seems weird to re-implement that here with homebrew.

The forumula as it stands today was the recommended way to do this when I first set out to package charm-tools for OSX users (#23161). Having homebrew run `python setup.py` with the prefix has been working out great for the past two years. As I found out in #46213 I can't just define the resource, I have to define the resource _and_ it's dependencies as resources. Python already had a way to do such a thing.

mikemcquaid:
> The fact that it's tedious and cumbersome to do this as a resource in homebrew. Python and setup.py already manages the dependency chain of requirements for this project. As I found out in #46213 I can't just define the resource, I have to define the resource and it's dependencies as resources. To breakdown each one - and each one's dependencies, would mean I'd be defining upwards of 30+ resources in this file. Seems weird to re-implement that here with homebrew.

It would be good if you could do that, thanks. It means users don't have to download things at build-time which opens them up to potentially nasty vulnerabilities unless there's checksum validation done there as strongly as ours. It also means none of those resources will be cached. We'd really appreciate it if you could do things our way here as we try and improve our tooling to automate this. Thanks!

marcoceppi:
Python package manager does checksum validation and all traffic over https. I can understand how defining resources for software which don't come with robust package management but this seems like overkill.

I care about my users but I've held up a release for 3 days already trying to get this into homebrew, it was easier to package with debian - now that's a metric ;) I will try to get this, but it's very difficult for me to test without first uploading it to your repo. Are there docs on how I can build and test this locally?

DomT4:
There's documentation on [Python handling here](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md).

You may particularly enjoy Tim's tool for resource block creation pretty much automatically: [Link](https://pypi.python.org/pypi/homebrew-pypi-poet).

marcoceppi:
I still think this is wrong. So that I can reference in the future:

```
virtualenv /tmp/homebrew-resources
. /tmp/homebrew-resources/bin/active
pip install charm-tools homebrew-pypi-poet
poet -r charm-tools > /tmp/resources
cat /tmp/foo | pastebinit
for i in `cat /tmp/foo | grep "resource " | awk -F'"' '{ print $2}'`; do echo -n "$i "; done | /tmp/resources-loop
```

PR #46279, opened 2015-11-23
bazel: add version information to the binary

damienmg:
The previous recipe was missing the version in the compiled binary
and `bazel version` resulted in a `HEAD (@non-git)` version which
makes hard for Bazel developers to track issues.

Instead uses `#{version}-homebrew` (for instances `0.1.1-homebrew`
for `0.1.1` release).

### DIFF
--- a/Library/Formula/bazel.rb
+++ b/Library/Formula/bazel.rb
@@ -20,6 +20,8 @@ class Bazel < Formula
       "/etc/bazel.bazelrc",
       "#{etc}/bazel/bazel.bazelrc"
 
+    ENV["EMBED_LABEL"] = "#{version}-homebrew"
+
     system "./compile.sh"
 
     (prefix/"base_workspace").mkdir


damienmg:
Should this wait for 0.1.2 version bump (should happens this week)?

Should this be tested? The guideline says we shouldn't test `bazel version`...

Library/Formula/bazel.rb
```diff
@@ -20,6 +20,8 @@ def install
       "/etc/bazel.bazelrc",
       "#{etc}/bazel/bazel.bazelrc"
 
+    ENV["EMBED_LABEL"] = "#{version}-homebrew"
```
bfontaine: Shouldn’t the version be set by the build script?

damienmg: The build script extract the version from the environment. Our CI set the version depending on the git tag. The compile.sh script knows nothing about the version label.

damienmg: Also compile.sh knows nothing about where it was built on. Here we mark the binary as being built by homebrew.

bfontaine: Makes sense, thanks.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/725d18f3677436d872a2cffe4d93580bbdc31c73. Thank you for your contribution to Homebrew @damienmg; we appreciate it! :smiley_cat:

PR #46283, opened 2015-11-23
brew-log: warn if shallow clone

bfontaine:
Example output:
```
$ brew log rtmp-nginx-module
Warning: The git repository is a shallow clone.
Use `git fetch --unshallow` to get the full repository.
commit 2887af6f8034fd3d5efe51c114c9a7a8d404d896
Author: Denis Denisov <denji0k@gmail.com>
Date:   Thu Jul 23 19:21:06 2015 +0300

    notice-nginx-module: use new patch #101
```

cc @mikemcquaid

### DIFF
--- a/Library/Homebrew/cmd/log.rb
+++ b/Library/Homebrew/cmd/log.rb
@@ -4,11 +4,25 @@ module Homebrew
   def log
     if ARGV.named.empty?
       cd HOMEBREW_REPOSITORY
-      exec "git", "log", *ARGV.options_only
+      git_log
     else
       path = Formulary.path(ARGV.named.first)
       cd path.dirname # supports taps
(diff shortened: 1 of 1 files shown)


mikemcquaid:
I guess rather than a shallow clone warning it's worth thinking about why we're warning them and what about?

bfontaine:
@mikemcquaid I don’t get you.

mikemcquaid:
Sorry, I mean: why print this warning? What is the problem with shallow clones. You told me in Slack but it might be worth thinking about communicating something like that instead e.g. `The git repository is a shallow clone therefore the filtering may be incorrect` or something similar.

bfontaine:
Oh right that makes sense. I updated the PR.

mikemcquaid:
:+1: :shipit:

PR #46286, opened 2015-11-23
Add note to remove `revision` on formula update

gunnaraasen:
Adds a small note that `revision` should also be removed when updating a formula for a version bump.

### DIFF
--- a/share/doc/homebrew/Formula-Cookbook.md
+++ b/share/doc/homebrew/Formula-Cookbook.md
@@ -909,7 +909,7 @@ Homebrew provides two Formula methods for launchd plist files. `plist_name` will
 
 ## Updating formulae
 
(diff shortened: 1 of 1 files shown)


DomT4:
Possibly worth mentioning only revision lines *outside* the bottle block need removing on updates.

mikemcquaid:
> Possibly worth mentioning only revision lines outside the bottle block need removing on updates.

Agreed. Thanks for the docs update :sparkles:!

DomT4:
Made the `bottle do` tweak in a follow-up commit. Merged in https://github.com/Homebrew/homebrew/commit/ed90e7365029bc67c05bae32900f18d37e7ab5b8.

Thank you for your contribution to Homebrew @gunnaraasen! :smiley_cat:

gunnaraasen:
Thanks @DomT4!

PR #46287, opened 2015-11-23
entr: add property list file for raising maxfiles

eradman:
This change recommends a changes to the default system limits when
`entr` is installed. This solves a common problem users have; e.g.:

https://bitbucket.org/eradman/entr/issues/28/a-way-to-watch-a-large-code-base

### DIFF
--- a/Library/Formula/entr.rb
+++ b/Library/Formula/entr.rb
@@ -68,8 +68,6 @@ class Entr < Formula
     touch testpath/"test.1"
     fork do
       sleep 0.5
-  plist_options :manual => ""
-
       touch testpath/"test.2"
     end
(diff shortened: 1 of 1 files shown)


eradman:
One more change is needed: because of the `plist` file Homebrew appends "To have launchd start entr ..." to the caveats. `entr` is not a daemon, it's a command-line utility. How can we suppress this auto-generated message?

mikemcquaid:
I'd rather this workaround be documented/output but `entr` itself as it's not Homebrew-specific. Additionally I'm not sure the `plist` method is a good fit for this use-case.

eradman:
That makes sense--I will add documentation `entr` on this subject.

PR #46302, opened 2015-11-24
konoha: style cleanup

DomT4:
Mostly a test of the special dependency problem we've been seeing in places, but I'll merge the style changes w/bottles if it actually passes.

### DIFF
--- a/Library/Formula/konoha.rb
+++ b/Library/Formula/konoha.rb
@@ -4,9 +4,15 @@ class Konoha < Formula
   url "https://github.com/konoha-project/konoha3/archive/v0.1.tar.gz"
   sha256 "e7d222808029515fe229b0ce1c4e84d0a35b59fce8603124a8df1aeba06114d3"
 
-  head "https://github.com/konoha-project/konoha3.git"
+  head do
+    url "https://github.com/konoha-project/konoha3.git"
 
-  option "tests", "Verify the build with make test (1 min)"
+    depends_on "openssl"
+  end
+
+  option "with-test", "Verify the build with make test (May currently fail)"
+
+  deprecated_option "tests" => "with-test"
 
   depends_on "cmake" => :build
   depends_on :mpi => [:cc, :cxx]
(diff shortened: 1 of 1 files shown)


mikemcquaid:
:+1:

mikemcquaid:
Looks like the tests are failing.

mikemcquaid:
The in-`install` tests, that is.

Library/Formula/konoha.rb
```diff
@@ -14,14 +15,13 @@ class Konoha < Formula
...
+  depends_on :python if MacOS.version <= :snow_leopard # for python glue code
 
   def install
-    args = std_cmake_args + [".."]
     mkdir "build" do
-      system "cmake", *args
+      system "cmake", "..", *std_cmake_args
       system "make"
-      system "make", "test" if build.include? "tests"
+      system "make", "test" if build.with?("tests") || build.bottle?
```
xu-cheng: I find behavior caused by `build.with?("tests") || build.bottle?` can be confusing without reading the code. Why not just enable test by default?

xu-cheng: In more general point, I think we should always enable test by default. The only exception I can think of is that test is buggy or take enormous long time and it's not security or crypto formula.

DomT4:
> Looks like the tests are failing.

Yeah. I did a bad thing and didn't test the PR before I pushed it because I expected it to die on the special dep and for us to deal with that. As it happened, it passed the special dep and then failed the `make check` :sweat:.

Library/Formula/konoha.rb
DomT4: I think all bottled builds should have test/check done on them, but I think it has more use to packagers than end users. Not against making it default all over the place barring special cases, but I think it's more useful to us than most people.

mikemcquaid:
> Yeah. I did a bad thing and didn't test the PR before I pushed it because I expected it to die on the special dep and for us to deal with that. As it happened, it passed the special dep and then failed the make check .

That's OK to do! That's what CI is for!

DomT4:
I normally test them locally, but really wasn't expecting this to get that far into the build so didn't bother, heh. Slightly annoyed it *did* get that far into the build because the special dep issue remains elusive.

apjanke:
It's passing now. Is this good to merge?

mikemcquaid:
:+1:

DomT4:
Heh, reported the failure upstream the other night, wrote a working test, pushed the branch again to trigger a new build... and then forgot all about the PR :sweat:. Merged.

PR #46308, opened 2015-11-24
parallel 20151122

zmwangx:
https://lists.gnu.org/archive/html/parallel/2015-11/msg00001.html.


dunn:
Thanks so much! Merged in 19168b61f83eff5229adeb9601e688d19990aa0b.

PR #46310, opened 2015-11-24
harfbuzz: 1.1.0

moyogo:
with patch from https://github.com/behdad/harfbuzz/issues/183

### DIFF
--- a/Library/Formula/harfbuzz.rb
+++ b/Library/Formula/harfbuzz.rb
@@ -1,8 +1,8 @@
 class Harfbuzz < Formula
   desc "OpenType text shaping engine"
   homepage "https://wiki.freedesktop.org/www/Software/HarfBuzz/"
-  url "http://www.freedesktop.org/software/harfbuzz/release/harfbuzz-1.0.4.tar.bz2"
-  sha256 "b030373457e7c00d3a7920f15e6fcd35defac3c4e44cd14ed85869030df74381"
+  url "http://www.freedesktop.org/software/harfbuzz/release/harfbuzz-1.1.0.tar.bz2"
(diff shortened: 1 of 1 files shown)


Library/Formula/harfbuzz.rb
```diff
@@ -34,6 +34,10 @@ class Harfbuzz < Formula
     sha256 "9535d35dab9e002963eef56757c46881f6b3d3b27db24eefcc80929781856c77"
   end
 
+  def patches
```
mikemcquaid: Use `patch do` instead.

Library/Formula/harfbuzz.rb
```diff
@@ -34,6 +34,10 @@ class Harfbuzz < Formula
...
+    DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

moyogo: Is it OK if the comment points to the issue upstream that has a patch in a comment?
```ruby
  # See https://github.com/behdad/harfbuzz/issues/183
  patch do
    DATA
  end
```

mikemcquaid: Given it's a GitHub repository: submitting a pull request would be good.

moyogo: Ah I see, `patch do` is really cool. I'll use that.

moyogo:
@mikemcquaid There’s now a pull request upstream behdad/harfbuzz#184 and `patch do` is used.

moyogo:
Closing since harfbuzz 1.1.1 just got released.

PR #46311, opened 2015-11-24
Add Nuxeo 7.10 formula

akervern:
Hi,

Here is a new formula to easily install Nuxeo.
This formula makes Nuxeo up and running with the basic configuration, and allow user to customise it in separate files.

Regards,

### DIFF
--- a/Library/Formula/nuxeo.rb
+++ b/Library/Formula/nuxeo.rb
@@ -33,6 +33,9 @@ class Nuxeo < Formula
       s.gsub! /#nuxeo\.pid\.dir.*/, "nuxeo.pid.dir=#{var}/run/nuxeo"
     end
     etc.install "#{libexec}/bin/nuxeo.conf"
+    (var/"log/nuxeo").mkpath
+    (var/"lib/nuxeo/data").mkpath
+    (var/"run/nuxeo").mkpath
   end
 
   def caveats; <<-EOS.undent
--- a/Library/Formula/nuxeo.rb
+++ b/Library/Formula/nuxeo.rb
@@ -1,0 +1,74 @@
+class Nuxeo < Formula
+  desc "Nuxeo Platform - Enterprise Content Management"
+  homepage "https://nuxeo.github.io/"
+  url "http://community.nuxeo.com/static/releases/nuxeo-7.10/nuxeo-cap-7.10-tomcat.zip"
+  version "7.10"
+  sha256 "9e95fb4e3dd0fdac49395ae8b429c6cc7db3c621ecd4d0dff74ea706a2aba723"
+
+  depends_on "poppler" => :recommended
+  depends_on "pdftohtml" => :optional
+  depends_on "imagemagick"
+  depends_on "ghostscript"
+  depends_on "ufraw"
+  depends_on "libwpd"
+  depends_on "ffmpeg" => :optional
+
+  def install
+    libexec.install Dir["#{buildpath}/*"]
+
+    bin.mkpath
+    libexec.mkpath
+
+    # link binary
+    bin.install_symlink "#{libexec}/bin/nuxeoctl"
+    # Inreplace Homebrew' paths to be independant from env variables
+    inreplace "#{libexec}/bin/nuxeoctl" do |s|
+      s.gsub! /\${NUXEO_HOME:=.*$/, '${NUXEO_HOME:=$(cd "$(dirname `PRG=$0; while [ -h "$PRG" ] ;do ls=\\\\`ls -ld "$PRG"\\\\`; link=\\\\`expr "$ls" : ".*-> \\\\(.*\)\$"\\\\`; if expr "$link" : "/.*" &> /dev/null; then PRG="$link"; else PRG="\\\\`dirname "$PRG"\\\\`/$link"; fi; done; echo $PRG`)"/..; pwd -P)}'
+      s.gsub! /\${NUXEO_CONF:=.*$/, "${NUXEO_CONF:=\"#{etc}/nuxeo.conf\"}"
+    end
+
+    # Initiatialize var dir
+    var_nuxeo = "#{var}/nuxeo"
+    unless Dir.exist?(var_nuxeo)
+      mkdir_p(var_nuxeo)
+    end
+
+    # Move data folder if not exists yet
+    unless Dir.exist?("#{var_nuxeo}/data")
+      mv(libexec/"nxserver/data", "#{var_nuxeo}/data")
+    end
+
+    unless File.exist?("#{etc}/nuxeo.conf")
+      # Change default path to Homebrew ones
+      inreplace "#{libexec}/bin/nuxeo.conf" do |s|
+        s.gsub! /#nuxeo\.log\.dir.*/, "nuxeo.log.dir=#{var_nuxeo}/log"
+        s.gsub! /#nuxeo\.data\.dir.*/, "nuxeo.data.dir=#{var_nuxeo}/data"
+        s.gsub! /#nuxeo\.tmp\.dir.*/, "nuxeo.tmp.dir=/tmp/nuxeo"
+      end
+
+      cp("#{libexec}/bin/nuxeo.conf", "#{etc}/nuxeo.conf")
+    end
+  end
+
+  def caveats; <<-EOS.undent
+    Nuxeo installed.
+
+    Configuration:
+      Data folder: #{var}/nuxeo/data
+      Log folder: #{var}/nuxeo/log
+      nuxeo.conf file: #{etc}/nuxeo.conf
+
+    Start Nuxeo:
+      $ nuxeoctl start
+  EOS
+  end
+
+  test do
+    # Copy configuration file to test path, due to some automatic writes on it.
+    cp("#{etc}/nuxeo.conf", "#{testpath}/nuxeo.conf")
+    ENV['NUXEO_CONF'] = "#{testpath}/nuxeo.conf"
+
+    # Ensure nuxeoctl is running fine
+    system "#{bin}/nuxeoctl", "showconf"
+  end
+end
--- a/Library/Formula/nuxeo.rb
+++ b/Library/Formula/nuxeo.rb
@@ -1,7 +1,7 @@
 class Nuxeo < Formula
-  desc "Nuxeo Platform - Enterprise Content Management"
+  desc "Enterprise Content Management"
   homepage "https://nuxeo.github.io/"
-  url "http://community.nuxeo.com/static/releases/nuxeo-7.10/nuxeo-cap-7.10-tomcat.zip"
+  url "http://cdn.nuxeo.com/nuxeo-7.10/nuxeo-cap-7.10-tomcat.zip"
   version "7.10"
   sha256 "9e95fb4e3dd0fdac49395ae8b429c6cc7db3c621ecd4d0dff74ea706a2aba723"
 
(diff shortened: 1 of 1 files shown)


Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,74 @@
...
+    libexec.install Dir["#{buildpath}/*"]
+
+    bin.mkpath
+    libexec.mkpath
+
+    # link binary
+    bin.install_symlink "#{libexec}/bin/nuxeoctl"
+    # Inreplace Homebrew' paths to be independant from env variables
+    inreplace "#{libexec}/bin/nuxeoctl" do |s|
+      s.gsub! /\${NUXEO_HOME:=.*$/, '${NUXEO_HOME:=$(cd "$(dirname `PRG=$0; while [ -h "$PRG" ] ;do ls=\\\\`ls -ld "$PRG"\\\\`; link=\\\\`expr "$ls" : ".*-> \\\\(.*\)\$"\\\\`; if expr "$link" : "/.*" &> /dev/null; then PRG="$link"; else PRG="\\\\`dirname "$PRG"\\\\`/$link"; fi; done; echo $PRG`)"/..; pwd -P)}'
```
mikemcquaid: This is quite a heavy thing for an `inreplace`. Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,74 @@
...
+      s.gsub! /\${NUXEO_CONF:=.*$/, "${NUXEO_CONF:=\"#{etc}/nuxeo.conf\"}"
+    end
+
+    # Initiatialize var dir
+    var_nuxeo = "#{var}/nuxeo"
+    unless Dir.exist?(var_nuxeo)
+      mkdir_p(var_nuxeo)
+    end
```
mikemcquaid: `(var/"nuxeo").mkpath` can replace all these lines above.

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,74 @@
...
+    unless File.exist?("#{etc}/nuxeo.conf")
+      # Change default path to Homebrew ones
+      inreplace "#{libexec}/bin/nuxeo.conf" do |s|
+        s.gsub! /#nuxeo\.log\.dir.*/, "nuxeo.log.dir=#{var_nuxeo}/log"
+        s.gsub! /#nuxeo\.data\.dir.*/, "nuxeo.data.dir=#{var_nuxeo}/data"
+        s.gsub! /#nuxeo\.tmp\.dir.*/, "nuxeo.tmp.dir=/tmp/nuxeo"
+      end
+
+      cp("#{libexec}/bin/nuxeo.conf", "#{etc}/nuxeo.conf")
+    end
```
mikemcquaid: Use `etc.install` instead of checking if the file exists.

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    Configuration:
+      Data folder: #{var}/nuxeo/data
+      Log folder: #{var}/nuxeo/log
+      nuxeo.conf file: #{etc}/nuxeo.conf
+
+    Start Nuxeo:
+      $ nuxeoctl start
+  EOS
+  end
```
mikemcquaid: Don't think we need any of these caveats.

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,74 @@
...
+    # Initiatialize var dir
+    var_nuxeo = "#{var}/nuxeo"
+    unless Dir.exist?(var_nuxeo)
+      mkdir_p(var_nuxeo)
+    end
+
+    # Move data folder if not exists yet
+    unless Dir.exist?("#{var_nuxeo}/data")
+      mv(libexec/"nxserver/data", "#{var_nuxeo}/data")
+    end
```
mikemcquaid: Do this `postinstall` instead and check for a file existing rather than just the directory.

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,74 @@
...
+  depends_on "libwpd"
+  depends_on "ffmpeg" => :optional
+
+  def install
+    libexec.install Dir["#{buildpath}/*"]
+
+    bin.mkpath
+    libexec.mkpath
+
+    # link binary
```
mikemcquaid: Don't need these comments; the code is self-explanatory.

Library/Formula/nuxeo.rb
akervern: The request is already done on the Nuxeo side; but will only be available on the next release (see https://jira.nuxeo.com/browse/NXP-18418).
I'm adding a comment to ensure it's clear for everyone.

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,74 @@
+class Nuxeo < Formula
+  desc "Nuxeo Platform - Enterprise Content Management"
+  homepage "https://nuxeo.github.io/"
+  url "http://community.nuxeo.com/static/releases/nuxeo-7.10/nuxeo-cap-7.10-tomcat.zip"
```
jcarsique: Use http://cdn.nuxeo.com/nuxeo-7.10/nuxeo-cap-7.10-tomcat.zip instead (CDN).

Library/Formula/nuxeo.rb
mikemcquaid: Ok, cool.

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,74 @@
...
+  EOS
+  end
+
+  test do
+    # Copy configuration file to test path, due to some automatic writes on it.
+    cp("#{etc}/nuxeo.conf", "#{testpath}/nuxeo.conf")
+    ENV['NUXEO_CONF'] = "#{testpath}/nuxeo.conf"
+
+    # Ensure nuxeoctl is running fine
+    system "#{bin}/nuxeoctl", "showconf"
```
jcarsique: `nuxeoctl status`?

akervern: Nope can't use it. As the return code is not 0 when the server is not running.

mikemcquaid: `shell_output` can verify a non-`0` exit code.

Library/Formula/nuxeo.rb
akervern: Moved to post_install. But I'd prefer to still check the directory as the sub files / sub directories depend on the user configuration and we want to keep this folder between upgrades.

mikemcquaid: I mean check for a file you're installing in that directory in case the user decides to delete the contents of that directory and expects a `reinstall` to put those files back.

akervern:
Can someone trigger a new travis build? The failed one is due to robocop gem installation.
See: https://travis-ci.org/Homebrew/homebrew/jobs/92965770

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,56 @@
...
+  depends_on "ghostscript"
+  depends_on "ufraw"
+  depends_on "libwpd"
+  depends_on "exiftool"
+  depends_on "ffmpeg" => :optional
+
+  def install
+    libexec.install Dir["#{buildpath}/*"]
+
+    bin.mkpath
```
bfontaine: You can drop this line, `install_symlink` will create `bin`.

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,56 @@
...
+      s.gsub! /#nuxeo\.data\.dir.*/, "nuxeo.data.dir=#{var}/lib/nuxeo/data"
+      s.gsub! /#nuxeo\.tmp\.dir.*/, "nuxeo.tmp.dir=/tmp/nuxeo"
+      s.gsub! /#nuxeo\.pid\.dir.*/, "nuxeo.pid.dir=#{var}/run/nuxeo"
+    end
+    etc.install "#{libexec}/bin/nuxeo.conf"
+  end
+
+  def caveats; <<-EOS.undent
+    You need to edit #{etc}/nuxeo.conf file to configure manually the server.
+    In case of upgrade, use 'nuxeoctl mp-list' against the old version then 'nuxeoctl mp-set' on the new version to reinstall packages.
```
bfontaine: Small nit here: could you please add a newline somewhere so the text is not wider than 80 characters? Thanks!

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,56 @@
...
+    EOS
+  end
+
+  test do
+    # Copy configuration file to test path, due to some automatic writes on it.
+    cp("#{etc}/nuxeo.conf", "#{testpath}/nuxeo.conf")
```
bfontaine: You can drop the parentheses here.

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    # TODO: https://jira.nuxeo.com/browse/NXBT-966 Use a wrapper
+    inreplace "#{libexec}/bin/nuxeoctl" do |s|
+      # See https://jira.nuxeo.com/browse/NXP-18418 nuxeoctl symlink is not correctly resolved and breaks NUXEO_HOME resolution.
+      s.gsub! /\${NUXEO_HOME:=.*$/, '${NUXEO_HOME:=$(cd "$(dirname `PRG=$0; while [ -h "$PRG" ] ;do ls=\\\\`ls -ld "$PRG"\\\\`; link=\\\\`expr "$ls" : ".*-> \\\\(.*\)\$"\\\\`; if expr "$link" : "/.*" &> /dev/null; then PRG="$link"; else PRG="\\\\`dirname "$PRG"\\\\`/$link"; fi; done; echo $PRG`)"/..; pwd -P)}'
+      s.gsub! /\${NUXEO_CONF:=.*$/, "${NUXEO_CONF:=\"#{etc}/nuxeo.conf\"}"
+    end
+
+    inreplace "#{libexec}/bin/nuxeo.conf" do |s|
+      s.gsub! /#nuxeo\.log\.dir.*/, "nuxeo.log.dir=#{var}/log/nuxeo"
```
DomT4: You probably want to `.mkpath` these elsewhere in `def install` so you know they'll definitely exist?

akervern: The `nuxeoctl` script do it at startup time.

DomT4: Presumably user owned and not root owned or anything?

akervern: Added those .mkpath to ensure paths rights.

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,59 @@
...
+    (var/"lib/nuxeo/data").mkpath
+    (var/"run/nuxeo").mkpath
+  end
+
+  def caveats; <<-EOS.undent
+    You need to edit #{etc}/nuxeo.conf file to configure manually the server.
+    In case of upgrade, use 'nuxeoctl mp-list' against the old version then
+    'nuxeoctl mp-set' on the new version to reinstall packages.
+    EOS
+  end
```
mikemcquaid: Is any of this Homebrew-specific? If not, please remove it.

jcarsique: I'd say it's Homebrew-specific and worth to mention it: other packaging systems differently manage the upgrade.

mikemcquaid: I think you can remove the `etc` bit. With the upgrade: where is the old/new version?

akervern: In the Nuxeo way; we keep the same nuxeo.conf file between upgrades. It contains each configuration entries and are versions independent. That's why the file is never override in the formula.
Otherwise, i can remove the `etc`, but as some of our users love to manually edit this configuration file; and Homebrew's `etc` directory might be unknown for the user; I'd prefer to let it here.

mikemcquaid: Sorry, I don't understand. You say `use 'nuxeoctl mp-list' against the old version` but I don't understand what that means. What command would I run?

akervern: In fact, Nuxeo allow users to install some plugins (aka Nuxeo Packages) and they can't be stored outside of the keg. So, I was not able to install them in a version independant path.

When the user will update Nuxeo using the formula upgrade, I guess he'll want to hold them. As we didn't figured out anything to handle version upgrade inside the Formula; we thought it was useful to give some guidelines to the user. 

So the `nuxeoctl mp-list` cmd is just a parameter to the nuxeo statup script that list the installed packages; then the other one (`nuxeoctl mp-set`) is to initialize a new nuxeo from a list of packages. I can add a link to the documentation to make it clearer if you want.

jcarsique: @akervern this can be indeed improved.
1) is there a command which would run the mp-list against the previous version? For instance with the proper NUXEO_HOME environment variable.
2) the list could be saved in a file before upgrade. Then output and point to the file after upgrade.
3) the list could be loaded after upgrade. For instance using the `installAfterRestart.log` file and the automatic mp-init command. Though I'm not sure that last step is wanted. You should follow the strategy used by the Debian package.

akervern: Thanks @jcarsique, good ideas. I think i'll handle a more precise command like you thought on the first point. As 2 or 3 were better and hidden for the user; but not sure how I can handle them without any upgrade entry point.

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  test do
+    # Copy configuration file to test path, due to some automatic writes on it.
+    cp "#{etc}/nuxeo.conf", "#{testpath}/nuxeo.conf"
+    inreplace "#{testpath}/nuxeo.conf" do |s|
+      s.gsub! /#{var}/, testpath
+    end
+    ENV["NUXEO_CONF"] = "#{testpath}/nuxeo.conf"
```
mikemcquaid: Does this always need set at runtime? If so, we should use wrapper scripts to set it for th euser.

Library/Formula/nuxeo.rb
mikemcquaid: > In fact, Nuxeo allow users to install some plugins (aka Nuxeo Packages) and they can't be stored outside of the keg. So, I was not able to install them in a version independant path.

Is it not possible to do this with symlinking paths elsewhere? It'd be good if we can do this.

jcarsique: You mean to store the Nuxeo packages outside of the server path? Packages content is tied to the server version, only their list will be *optionally* kept among upgrades (optionally because some may become deprecated, useless/redundant or conflicting).
Experience highlighted it was error-prone not to store the plugins inside the server.

mikemcquaid: It's really a really bad user experience to require all plugins be reinstall on a reinstall and/or upgrade. Again, I'd suggest using symlinking here which works for most software which has this limitation.

Library/Formula/nuxeo.rb
jcarsique: Is this the same suggestion as in the above comment `# TODO: https://jira.nuxeo.com/browse/NXBT-966 Use a wrapper`? 
If so, your recommendations are welcome:
- could that improvement be delayed to the next version of the current formula?
- do you agree it does not sound very good, nor easy to read or maintain, to make the formula generate that wrapper on the fly?
- we plan to distribute such a wrapper along with the next Nuxeo versions but for the already released versions, is it ok to make the formula download the wrapper from the almost same url as the main ZIP?

mikemcquaid: > could that improvement be delayed to the next version of the current formula?

I'd rather if it wasn't; it's a poor user experience as-is.



> do you agree it does not sound very good, nor easy to read or maintain, to make the formula generate that wrapper on the fly?

We have tooling to do this easily for you e.g. see `ansible.rb`).



> we plan to distribute such a wrapper along with the next Nuxeo versions but for the already released versions, is it ok to make the formula download the wrapper from the almost same url as the main ZIP?

If you'd like to officially distribute a wrapper it would be good if it was in the main zip.

bfontaine:
Ping re Mike’s comments?

jcarsique:
@akervern was not available last week. He will answer asap.

akervern:
Hi, I apologize for the long time without news

So, thanks for the pointer about `write_env_script` method; it's exactly what we need to prevent the regex. And, not sure to understand why, but it fixes also the need to update paths for the test.
About the upgrade addons; I'll handle on the Nuxeo side something to make a transparent upgrade when differents versions of Nuxeo are sharing the same installed packages cache / reference. But for now, with the symlink there is no blocker issue as the upgraded version will run fine, but there is still the need to execute the command specified in the caveats to upgrade / reinstall packages.

Thanks,

akervern:
Hi there, I still apologize for the lack of responses; but do you have any new remarks?

mikemcquaid:
@akervern Waiting for the existing ones to be addressed.

Library/Formula/nuxeo.rb
akervern: I used the wrapper for the nominal way.
But for testing, it is still needed to ensure everything is done in the sandbox, cause' our startup script is forcing a write on the configuration file to fix all possible CRLF issues... As Java do not like them...

mikemcquaid: :ok: :cool:

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    # TODO: Remove grep as soon as warn are correctly written on stderr (https://jira.nuxeo.com/browse/NXP-18724)
+    assert_equal "#{testpath}/nuxeo.conf", shell_output("#{libexec}/bin/nuxeoctl config -q --get nuxeo.conf | grep -v WARN").strip
+    assert_equal "#{libexec}", shell_output("#{libexec}/bin/nuxeoctl config -q --get nuxeo.home | grep -v WARN").strip
```
mikemcquaid: Would be better to use `assert_match` for these e.g. `assert_match(/.*WARN.*#{libexec}/, shell_output(...`

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,57 @@
...
+      s.gsub! /#nuxeo\.log\.dir.*/, "nuxeo.log.dir=#{var}/log/nuxeo"
+      s.gsub! /#nuxeo\.data\.dir.*/, "nuxeo.data.dir=#{var}/lib/nuxeo/data"
+      s.gsub! /#nuxeo\.tmp\.dir.*/, "nuxeo.tmp.dir=/tmp/nuxeo"
+      s.gsub! /#nuxeo\.pid\.dir.*/, "nuxeo.pid.dir=#{var}/run/nuxeo"
+    end
+    etc.install "#{libexec}/bin/nuxeo.conf"
+    (var/"log/nuxeo").mkpath
+    (var/"lib/nuxeo/data").mkpath
+    (var/"run/nuxeo").mkpath
+    (var/"cache/nuxeo/packages").mkpath
```
mikemcquaid: Put all the `mkpath`s in a `post_install` for binary packaging, thanks!

Library/Formula/nuxeo.rb
```diff
@@ -0,0 +1,57 @@
...
+  def install
+    libexec.install Dir["#{buildpath}/*"]
+
+    (bin/"nuxeoctl").write_env_script "#{libexec}/bin/nuxeoctl",
+      :NUXEO_HOME => "#{libexec}", :NUXEO_CONF => "#{etc}/nuxeo.conf"
+
+    inreplace "#{libexec}/bin/nuxeo.conf" do |s|
+      s.gsub! /#nuxeo\.log\.dir.*/, "nuxeo.log.dir=#{var}/log/nuxeo"
+      s.gsub! /#nuxeo\.data\.dir.*/, "nuxeo.data.dir=#{var}/lib/nuxeo/data"
+      s.gsub! /#nuxeo\.tmp\.dir.*/, "nuxeo.tmp.dir=/tmp/nuxeo"
```
mikemcquaid: I wonder if it's worth putting this in `var` too? Not sure `/tmp` will ever be cleaned.

mikemcquaid:
A few final comments. Almost there!

Library/Formula/nuxeo.rb
akervern: We wondering about it too. But as it's used for generating files that are mainly used once, and to ease any behaviors about cleaning this folder as it could grow... We prefer to let it in the system `/tmp` by default in the formula.

akervern:
Finger crossed :)

Library/Formula/nuxeo.rb
mikemcquaid: > to ease any behaviors about cleaning this folder as it could grow

That's my point: `/tmp` won't be cleaned automatically.

akervern: Not sure to understand why `/var` could be better in that case.

mikemcquaid: I guess at least it's a more obvious location where other Homebrew data is stored rather than growing indefinitely in `/tmp`

akervern: I see. In that case, I'll let the default folder as there is no need to share this folder between all versions.

akervern:
Can someone trigger a new build? Please? :)

bfontaine:

mikemcquaid:

For future reference the preferred commit message format for simple version updates is `nuxeo 7.10 (new formula)`.

PR #46322, opened 2015-11-24
Fix llvm --HEAD

quantumsteve:
This fixes an error installing the head version of llvm. The location of scan-build and scan-view have changed.


mikemcquaid:

For future reference the preferred commit message format is `llvm: fix head install.`.

PR #46332, opened 2015-11-24
gsar 1.21 (new formula)

vszakats:


Library/Formula/gsar.rb
```diff
@@ -0,0 +1,16 @@
+class Gsar < Formula
+  desc "General Search And Replace on files"
+  homepage "http://tjaberg.com/"
+  url "http://tjaberg.com/gsar121.zip"
+  sha256 "05fb9583c970aba4eb0ffae2763d7482b0697c65fda1632a247a0153d7db65a9"
+  version "1.21"
```
bfontaine: Please move this above the checksum.

Library/Formula/gsar.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def install
+    system "make"
+    bin.install "gsar"
```
bfontaine: Is there no `make install` we could use?

vszakats: No, unfortunately it's missing from the Makefile.

Library/Formula/gsar.rb
vszakats: Yes, got those, and the whitespace problems as well. Thanks!

bfontaine:
Is it a popular project? It’s hard to find anything on Google.

vszakats:
I don't have any information on it's popularity. I've been using it for 10+ years on the Windows platform and I've recently found references for it on Stackoverflow. It's a very practical and fast simple search and replace tool for those occasions when `ack` and `sed` is "too much". Having this on OS X also helps making the shift to the platform without having to rewrite much existing script logic.

bfontaine:
Thank you @vszakats! :smile:

PR #46363, opened 2015-11-25
gsar: let it be bottled

vszakats:
Ref: https://github.com/Homebrew/homebrew/commit/7e82d8587121bdc4f01ec806b0eab97b8dd344d6#commitcomment-14620042


bfontaine:
Thank you @vszakats! :wink:

PR #46364, opened 2015-11-25
hugo v0.15

joneskoo:
go_resources generated with:

  homebrew-go-resources github.com/spf13/hugo

Replaced manually maintained resources with output from script. Seems like a lot of work to do it any other way than by script?

gen autocomplete syntax had changed.

This is my first change to a go formula…

### DIFF
--- a/Library/Formula/hugo.rb
+++ b/Library/Formula/hugo.rb
@@ -3,8 +3,8 @@ require "language/go"
 class Hugo < Formula
   desc "Configurable static site generator"
   homepage "https://gohugo.io/"
-  url "https://github.com/spf13/hugo/archive/v0.14.tar.gz"
-  sha256 "67e0fda342e6a35490ce03f5535e30cf04bc06775c47bacf416d96d83cce4535"
+  url "https://github.com/spf13/hugo/archive/v0.15.tar.gz"
+  sha256 "417106a9a52fd1740bcb8a0102684da13da389b9a126eb1829b61cfdf3d86b63"
 
   bottle do
     cellar :any_skip_relocation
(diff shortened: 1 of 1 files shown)


Library/Formula/hugo.rb
```diff
@@ -186,7 +178,7 @@ def install
 
     system "go", "build", "-o", bin/"hugo", "main.go"
 
-    system bin/"hugo", "genautocomplete", "--completionfile=#{buildpath}/hugo.sh"
+    system bin/"hugo", "gen", "autocomplete", "--completionfile=#{buildpath}/hugo.sh"
```
dunn: The release notes say `hugo gen` can create manpages too. Could you add a second command for that? Thanks so much!

joneskoo:
@dunn Added man pages. Unfortunately doesn't look like the command supports target dir, but luckily brew is flexible!

https://github.com/spf13/hugo/blob/3d09de891072985e6c59f899f09132a76d44f7cd/commands/genman.go#L19

dunn:
Thanks so much; merged in fbca49bba5e9f478a9dec4e5dd84e8c73e79912d.  If you feel like pestering the developers to make `gen man` more flexible, that'd be awesome too!

anthonyfok:
You are right!  I thought I did add a configurable destination man page output directory in my pull request https://github.com/spf13/hugo/pull/1602, but it turns out that I really didn't!  No wonder nothing shows up with `hugo gen man --help`!  (Oops!)

Thank you for working around that in the Homebrew recipe(?)/script.  I will get that fix in soon to go in for v0.16.

Cheers,
Anthony

PR #46369, opened 2015-11-25
BuildOptions#with{,out}?: support multiple args

bfontaine:
This would allow this:

```rb
something if build.without? "python", "python3"
```

Right now we have to use the following:

```rb
something if build.without?("python") && build.without?("python3")
```

Another possible implementation would be to accept an array: `build.without? %w[python python3]`.

### DIFF
--- a/Library/Homebrew/build_options.rb
+++ b/Library/Homebrew/build_options.rb
@@ -23,22 +23,24 @@ class BuildOptions
   # else
   #   args << "--with-example1"
   # end</pre>
-  def with?(val)
-    name = val.respond_to?(:option_name) ? val.option_name : val
+  def with?(*vals)
+    vals.all? do |val|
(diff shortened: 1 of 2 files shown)


UniqMartin:
I think in cases like these, the current and more verbose syntax is much clearer to someone reading the formula. If `with?` and `without?` start to allow for multiple arguments or arrays, I wouldn't intuitively know whether that means `any?` or `all?` and I'd argue that, depending on the circumstances, both interpretations might make sense.

I'd be more in support if we created new methods and named them something like `with_any_of?`, `with_all_of?`, and `with_none_of?`. That would be more self-evident. I'm a bit skeptical whether there are enough use cases to warrant the existence of these methods, to be honest.

bfontaine:
> I think in cases like these, the current and more verbose syntax is much clearer to someone reading the formula.

Yeah you’re right I was thinking about this when writing the tests.

PR #46370, opened 2015-11-25
fpc 3.0.0

tomyun:


Library/Formula/fpc.rb
```diff
@@ -25,6 +25,9 @@ def install
     system "make", "install", "PP=#{fpc_compiler}", "PREFIX=#{prefix}"
 
     bin.install_symlink lib/"#{name}/#{version}/ppcx64"
+
+    # Prevent non-executable audit warning
+    rm_f Dir[bin/"*.rsj"]
```
DomT4: Are these files not needed by anything?

tomyun: To my understanding, they are only useful for localization later, but not required for execution. There were only two files for `makeskel` and `ptop`.

mikemcquaid:

PR #46374, opened 2015-11-26
icu4c 56.1

DomT4:


mikemcquaid:
:+1: when :green_heart:

DomT4:
Aborted last night because the test I wrote that works fine locally seems to hang indefinitely on the CI. I'll just remove the test block there and put it back in a follow up PR, rather than trying to diagnose in a PR that's going to take 2+ hours to build each time. Functionality is fine locally against the new `icu4c`.

DomT4:
CC @Homebrew/science - You may need to do some revisions on your `icu4c`-using formulae.

PR #46378, opened 2015-11-26
gptfdisk 1.0.1

dunn:


Library/Formula/gptfdisk.rb
```diff
@@ -13,10 +12,15 @@ class Gptfdisk < Formula
     sha256 "95593d9ce977a9529b11c9de8ee1089e56c67d76a642e13bdac31097aa5c7f69" => :mountain_lion
   end
 
+  # Due to SIP
+  depends_on MaximumMacOSRequirement => :yosemite
```
dunn: Detailed on the homepage: http://www.rodsbooks.com/gdisk/

Library/Formula/gptfdisk.rb
```diff
@@ -13,10 +12,15 @@ class Gptfdisk < Formula
...
+
   depends_on "popt"
   depends_on "icu4c"
 
   def install
+    inreplace "Makefile.mac", "/opt/local/lib/libncurses.a", "/usr/lib/libncurses.dylib"
```
dunn: I doubt this is worth reporting upstream since we'll probably have to boneyard the whole thing soon.

Library/Formula/gptfdisk.rb
mikemcquaid: I'd be tempted to go straight to the boneyard.

dunn: OK!

PR #46386, opened 2015-11-26
GitRequirement: use Utils.git_available?

xu-cheng:
`which("git")` will return incorrect result for OS X without Xcode/CLT
installed, where `/usr/bin/git` is a stub.


mikemcquaid:
:+1:

UniqMartin:
:+1:

PR #46396, opened 2015-11-26
updating archey-osx to 1.5.2

obihann:


Library/Formula/archey.rb
```diff
@@ -1,8 +1,9 @@
 class Archey < Formula
   desc "Archey script for OS X"
+  version "1.5.2"
```
DomT4: Won't need this line. We can read simple versions from URLs.

obihann: Thanks for the info, I've removed the line now.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/0e20c9a820a76770645979587fbcbb3bf6fbbcda. Thank you for your contribution to Homebrew, and the very cool Archey OS X clone @obihann.

PR #46400, opened 2015-11-26
Homebrew Bundle speedup changes

mikemcquaid:
I'm working on speeding up `brew bundle check` so as part of that I'm needing to make a few core changes. These are:

- Make `brew info --json=v1` also output outdated and pinned formulae.
- Make `tab` use a in-memory cache to avoid unnecessary rereading the same `Tab` multiple times per process.

CC @xu-cheng

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -927,6 +927,44 @@ class Formula
   end
 
   # @private
+  def outdated_versions
+    @outdated_versions ||= begin
+      all_versions = []
+      older_or_same_tap_versions = []
+
+      if oldname && !rack.exist? && (dir = HOMEBREW_CELLAR/oldname).directory? &&
+        !dir.subdirs.empty? && tap == Tab.for_keg(dir.subdirs.first).tap
+        raise Migrator::MigrationNeededError.new(f)
+      end
+
+      rack.subdirs.each do |keg_dir|
+        keg = Keg.new keg_dir
+        version = keg.version
+        all_versions << version
+        older_version = pkg_version <= version
+
+        tab_tap = Tab.for_keg(keg).tap
+        if tab_tap.nil? || tab_tap == tap || older_version
+          older_or_same_tap_versions << version
+        end
+      end
+
+      if older_or_same_tap_versions.all? { |v| pkg_version > v }
+        all_versions
+      else
+        []
+      end
+    end
+  end
+
+  # @private
+  def outdated?
+    outdated_versions.any?
+  rescue Migrator::MigrationNeededError
+    true
+  end
+
+  # @private
   def pinnable?
     @pin.pinnable?
   end
@@ -1202,6 +1240,8 @@ class Formula
       "revision" => revision,
       "installed" => [],
       "linked_keg" => (linked_keg.resolved_path.basename.to_s if linked_keg.exist?),
+      "pinned" => pinned?,
+      "outdated" => outdated?,
       "keg_only" => keg_only?,
       "dependencies" => deps.map(&:name).uniq,
       "conflicts_with" => conflicts.map(&:name),
--- a/Library/Homebrew/tab.rb
+++ b/Library/Homebrew/tab.rb
@@ -9,6 +9,11 @@ require "utils/json"
 # `Tab.create`.
 class Tab < OpenStruct
   FILENAME = "INSTALL_RECEIPT.json"
+  CACHE = {}
+
+  def self.clear_cache
+    CACHE.clear
+  end
 
   def self.create(formula, compiler, stdlib, build)
     attributes = {
@@ -32,7 +37,7 @@ class Tab < OpenStruct
   end
 
   def self.from_file(path)
-    from_file_content(File.read(path), path)
+    CACHE.fetch(path) { |p| CACHE[p] = from_file_content(File.read(p), p) }
   end
 
   def self.from_file_content(content, path)
@@ -217,6 +222,7 @@ class Tab < OpenStruct
   end
 
   def write
+    CACHE[tabfile] = self
     tabfile.atomic_write(to_json)
   end
 
--- a/Library/Homebrew/cmd/outdated.rb
+++ b/Library/Homebrew/cmd/outdated.rb
@@ -13,41 +13,12 @@ module Homebrew
     Homebrew.failed = ARGV.resolved_formulae.any? && outdated.any?
   end
 
-  def outdated_brews(formulae)
-    formulae.map do |f|
-      all_versions = []
-      older_or_same_tap_versions = []
-
-      if f.oldname && !f.rack.exist? && (dir = HOMEBREW_CELLAR/f.oldname).directory? &&
-        !dir.subdirs.empty? && f.tap == Tab.for_keg(dir.subdirs.first).tap
-        raise Migrator::MigrationNeededError.new(f)
-      end
-
-      f.rack.subdirs.each do |keg_dir|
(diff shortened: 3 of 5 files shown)


Library/Homebrew/tab.rb
```diff
@@ -35,30 +35,39 @@ def self.from_file(path)
...
-    attributes["tabfile"] = path
-    attributes["source"] ||= {}
-
-    tapped_from = attributes["tapped_from"]
-    unless tapped_from.nil? || tapped_from == "path or URL"
-      attributes["source"]["tap"] = attributes.delete("tapped_from")
-    end
+  def self.write_file_content(content, path)
+    from_file_content(content, path, :refresh_cache => :true)
+    path.atomic_write(content)
```
xu-cheng: Sorry this seems to be incorrect. When I need to write new tab file, I shouldn't read old tab file.

mikemcquaid: This doesn't read the old tab file as `self.from_file_content` is passed the new path/content and reads from the passed content rather than from disk.

Library/Homebrew/tab.rb
```diff
@@ -35,30 +35,39 @@ def self.from_file(path)
...
+  end
+
+  def self.from_file_content(content, path, options={})
```
xu-cheng: I think you may accidentally overcomplicate this. You can catch the object directly in `from_file`.

Library/Homebrew/cmd/info.rb
```diff
@@ -54,7 +55,12 @@ def print_json
     else
       ARGV.formulae
     end
-    json = ff.map(&:to_hash)
+    outdated_formulae = outdated_brews(ff)
+    json = ff.map do |f|
+      hash = f.to_hash
+      hash["outdated"] = outdated_formulae.include?(f)
+      hash
+    end
```
xu-cheng: I think we should do a deep refactor here by introducing `Formula#outdated?` method.

Library/Homebrew/tab.rb
xu-cheng: I don't think you need to refresh cache as you are editing the object inside cache directly.

mikemcquaid: You need to update the contents of the cache with what we're about to write to the file so future reads get what has been written.

Library/Homebrew/tab.rb
mikemcquaid: It needs to be cached on both read and write to ensure consistency and to avoid writing something whose contents we already know.

Library/Homebrew/cmd/info.rb
mikemcquaid: Yeh, agreed and then include it in the `hash`.

Library/Homebrew/tab.rb
xu-cheng: I think this is all you need to cache the tab object.

```diff
diff --git a/Library/Homebrew/tab.rb b/Library/Homebrew/tab.rb
index f2dceb7..677e607 100644
--- a/Library/Homebrew/tab.rb
+++ b/Library/Homebrew/tab.rb
@@ -9,6 +9,7 @@ require "utils/json"
 # `Tab.create`.
 class Tab < OpenStruct
   FILENAME = "INSTALL_RECEIPT.json"
+  CACHE = {}
 
   def self.create(formula, compiler, stdlib, build)
     attributes = {
@@ -32,7 +33,7 @@ class Tab < OpenStruct
   end
 
   def self.from_file(path)
-    from_file_content(File.read(path), path)
+    CACHE.fetch(path) { |p| CACHE[p] = from_file_content(File.read(p), p) }
   end
 
   def self.from_file_content(content, path)
@@ -217,6 +218,7 @@ class Tab < OpenStruct
   end
 
   def write
+    CACHE[tabfile] = self
     tabfile.atomic_write(to_json)
   end
 
```

mikemcquaid:
@xu-cheng Updated and ready for rereview.

Library/Homebrew/formula.rb
```diff
@@ -927,6 +927,40 @@ def unlock
...
+    end
+
+    rack.subdirs.each do |keg_dir|
+      keg = Keg.new keg_dir
+      version = keg.version
+      all_versions << version
+      older_version = pkg_version <= version
+
+      tap = Tab.for_keg(keg).tap
+      if tap.nil? || tap == tap || older_version
```
xu-cheng: `tap == tap`?

You need a new variable here.

mikemcquaid: Hah, yes, whoops.

Library/Homebrew/formula.rb
```diff
@@ -927,6 +927,40 @@ def unlock
...
+        older_or_same_tap_versions << version
+      end
+    end
+
+    return [] unless older_or_same_tap_versions.all? { |v| pkg_version > v }
+
+    all_versions
```
xu-cheng: `Formula#outdated_versions` will return `all_versions`?

Maybe we should make them two methods, i.e. `Formula#alll_versions` and `Formula#outdated_versions`.

mikemcquaid: It'll only return all versions if they are all outdated.

Library/Homebrew/formula.rb
```diff
@@ -927,6 +927,40 @@ def unlock
...
+  def outdated_versions
+    all_versions = []
+    older_or_same_tap_versions = []
+
+    if oldname && !rack.exist? && (dir = HOMEBREW_CELLAR/oldname).directory? &&
+      !dir.subdirs.empty? && tap == Tab.for_keg(dir.subdirs.first).tap
+      raise Migrator::MigrationNeededError.new(f)
+    end
+
+    rack.subdirs.each do |keg_dir|
```
xu-cheng: If you want, this PR offers us a chance to do more deep refactoring. e.g.
* make  `rack.subdirs` to `Formula#keg_dirs`
* introduce `Formula#kegs`

So we can avoid more duplicated logics.

mikemcquaid: I'd probably rather keep it as-is for now.

mikemcquaid: (I want to get this merged :soon:)

Library/Homebrew/formula.rb
xu-cheng: I think make them two methods can be benefit as:
* it's more easy to understand.
* `Formula#all_installed_versions` could be used somewhere else.

mikemcquaid: It could be but it's not currently being used elsewhere. It'll also need to loop through the subdirs twice. This logic is more or less as-is from `outdated.rb` and I think that's good as it will behave the same.

xu-cheng: Maybe you should ignore my last comment. I see why they are need to be in the same methods

mikemcquaid: :+1:. Anything else blocking this being merged?

Library/Homebrew/formula.rb
```diff
@@ -927,6 +927,40 @@ def unlock
...
+    end
+
+    return [] unless older_or_same_tap_versions.all? { |v| pkg_version > v }
+
+    all_versions
+  end
+
+  # @private
+  def outdated?
+    outdated_versions.any?
```
xu-cheng: One last thing: `outdated_versions`'s result need to be cached. Otherwise it will be computed twice.

xu-cheng:
Except my last comment, :+1:.

I will handle more refactoring in new PR.

PR #46401, opened 2015-11-26
open-scene-graph: new GitHub page, add devel and head

apjanke:
Upstream confirmed on mailing list that the GitHub repo is now the primary
source location.
http://forum.openscenegraph.org/viewtopic.php?p=65771#65771

CI tests for this will fail due to other outstanding issues.


apjanke:
Closing this to avoid spurious test jobs. I'll resubmit this and the other `open-scene-graph` PRs as a combined PR that can pass the tests.

apjanke:
Revised to remove the `devel` block, since it's not building correctly.

PR #46407, opened 2015-11-26
wget: add metalink/gpgme support

DomT4:
Had a Twitter request from upstream, as well as from antbryan.

### DIFF
--- a/Library/Formula/wget.rb
+++ b/Library/Formula/wget.rb
@@ -1,6 +1,5 @@
 # NOTE: Configure will fail if using awk 20110810 from dupes.
 # Upstream issue: https://savannah.gnu.org/bugs/index.php?37063
-
 class Wget < Formula
   desc "Internet file retriever"
   homepage "https://www.gnu.org/software/wget/"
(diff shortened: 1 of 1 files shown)


antbryan:
aha, thank you sir!

it was the

````
````
that I didn't have, that was messing me up. thanks a bunch & have a good one. :)

DomT4:
Yeah, turns out the only way `wget` wants to detect `libmetalink` is via `pkg-config`. Since it tries to use `pkg-config` for the build regardless, making it an unconditional build dependency was the simplest way to resolve that.

All seems to work locally from testing. Apologies for the delay on the reply to you - I've been moving Macs which didn't go as smoothly as hoped, heh.

antbryan:
is that worth contacting upstream about? (do wget's other dependencies behave that way, or are they detected another way?)

oh, please don't apologize, I appreciate the help!

DomT4:
You *can* pass direct paths to `wget` for some other dependencies; it accepts `--with-libssl-prefix` and `--with-gpgme-prefix`. Looks like the rest are either searched for by `pkg-config` or searched for in default locations.

I'm not quite sure whether they've moving towards or away from being able to pass direct paths to dependencies prefix, but if upstream are interested being able to pass a direct path rather than throwing out to `pkg-config` would be an upside.

PR #46408, opened 2015-11-26
Fix typo: hyphen-minus instead of en dash in command

zmwangx:
An en dash slipped into the command `brew –-cache -s erlang`... So if an unsuspecting reader (that is I) copies it into the terminal, he will get an `Error: Unknown command: –-cache`.


dunn:
That was me 😰 good catch!

zmwangx:
Since en dash looks just like a hyphen-minus in 10pt Monaco, I wondered for a while (until I `xxd`ed at last) why it didn't work, as it's also documented in the man page...

dunn:
Yeah, having my editor "intelligently" insert en/em dashes and curly quotes is maybe a bad idea.

dunn:
Merged in b51d0bc7f7b3575227538f84151b3fa4318f2774, thanks again!

PR #46411, opened 2015-11-26
Add dcadec formula

rcombs:


Library/Formula/dcadec.rb
```diff
@@ -0,0 +1,20 @@
+class Dcadec < Formula
+  desc "DTS Coherent Acoustics decoder with support for HD extensions"
+  homepage "https://github.com/foo86/dcadec"
+  url "https://github.com/foo86/dcadec/archive/v0.1.0.tar.gz"
+  sha256 "52a68a63f79ae50e5e5fd9c631a715e9feeed5d2b4a8442f0b9208692012a43f"
+  head "https://github.com/foo86/dcadec.git"
+
+  def install
+    system "make", "all"
+    unless File.exist?("test/samples/README")
```
mikemcquaid: When would this file exist/not exist?

Library/Formula/dcadec.rb
```diff
@@ -0,0 +1,20 @@
...
+      system "git", "clone", "https://github.com/foo86/dcadec-samples.git", "test/samples"
+    end
+    system "make", "check"
+    system "make", "PREFIX=#{prefix}", "install"
+  end
+
+  test do
+    system "#{bin}/dcadec", "-h"
```
mikemcquaid: Would be good to actual encode a sample file.

Library/Formula/dcadec.rb
```diff
@@ -0,0 +1,20 @@
...
+  desc "DTS Coherent Acoustics decoder with support for HD extensions"
+  homepage "https://github.com/foo86/dcadec"
+  url "https://github.com/foo86/dcadec/archive/v0.1.0.tar.gz"
+  sha256 "52a68a63f79ae50e5e5fd9c631a715e9feeed5d2b4a8442f0b9208692012a43f"
+  head "https://github.com/foo86/dcadec.git"
+
+  def install
+    system "make", "all"
+    unless File.exist?("test/samples/README")
+      system "git", "clone", "https://github.com/foo86/dcadec-samples.git", "test/samples"
```
mikemcquaid: If you do this it should use a `resource` instead of `git clone`

Library/Formula/dcadec.rb
rcombs: It's in the submodule, and doesn't exist when downloading the tarball.

Library/Formula/dcadec.rb
rcombs: How is that done for a git repo?

Library/Formula/dcadec.rb
rcombs: `make check` does that (relying on the sample files being present).

Library/Formula/dcadec.rb
mikemcquaid: In that case it'll never exist. If you use `https://github.com/foo86/dcadec.git` as the URL and provide `:revision` and `:tag` it'll download the submodules for you.

Library/Formula/dcadec.rb
mikemcquaid: This isn't a test suite but a verification that the formula was installed and compiled properly. Please install some sample files and run the command here too. Thanks!

Library/Formula/dcadec.rb
rcombs: Does that lock `--HEAD` to always installing that revision, or just indicate which one the tar.gz comes from?

mikemcquaid: Nope, `--HEAD` is still unlocked but will share the repository. Homebrew will just check it out to the right revision.

Library/Formula/dcadec.rb
rcombs: I'm not sure how best to go about this. The package's build system has a built-in way to decode a number of provided samples (from that submodule URL) and check that the output was correct. I could duplicate that test logic in Ruby, but  it'd probably be easier to just checkout the repo and run `make check`. Would that be preferred over having `make check` run in the install step? (It'd probably be convenient to have some way of specifying that there's a test suite within the repo)

Library/Formula/dcadec.rb
rcombs: So does that replace the `.tar.gz` URL, then?

Library/Formula/dcadec.rb
mikemcquaid: It's sufficient to just `.install` a sample file and run with `system` which will make sure it didn't e.g. crash.

rcombs: Alright, should I remove the `make check` and submodule, then?

mikemcquaid: It's fine to keep them too; think of it as unit vs integration tets.

Library/Formula/dcadec.rb
bfontaine: Use:

```rb
resource "some-name" do
  url "https://github.com/foo86/dcadec-samples.git"
end
```

mikemcquaid: Except in this case it's a submodule so not needed.

bfontaine:

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/f0c4fcf9216ee9fd8a64b0787078bb55703f40fb. Thank you for your contribution to Homebrew @rcombs; we appreciate it! :smiley_cat:

PR #46412, opened 2015-11-27
Remove brew-any-tap and brew-desc from external commands docs

zmwangx:
The functionality of both external commands have been merged into core (#40326 and #39911).


mikemcquaid:

For future reference the preferred commit message format is `External-Commands: remove brew-any-tap, brew-desc.`.

PR #46413, opened 2015-11-27
Add mpv formula

rcombs:


zmwangx:
`mpv` has it's own official tap: [`mpv-player/mpv`](https://github.com/mpv-player/homebrew-mpv/). Isn't it better to respect the developers' choice, if they prefer to host their formulae on their own? Or have you contacted them regarding the import? I'd consider it not so cool if you just grab [someone else's code](https://github.com/mpv-player/homebrew-mpv/blob/master/Formula/mpv.rb) without notice and attribution.

pigoz:
We have been discussing moving moving the `mpv` formula to the main homebrew repository and closing down the tap.

There are currently some pain points we have to address with vapoursynth integration. The patch from @rcombs removes it, but leaves the `python3` dependency and all the `PYTHONPATH` related code that  is only needed for vapoursynth. So the code should be improved.

Vapoursynth has always been problematic, and it's main use cases within mpv are now handled by the player itself: we have a GLSL debanding shader that can be used instead of f3kdb, and better frame timing and interpolation which makes mvtools a lot less desirable. For these reason I propose we remove it and all related formulas from the tap repository. If no angry uses come poking at us for the removal within a few weeks, we can contribute this new stripped down formula of mpv to homebrew and close down the tap.

mikemcquaid:
> If no angry uses come poking at us for the removal within a few weeks, we can contribute this new stripped down formula of mpv to homebrew and close down the tap.

Sounds good :+1:

Cellane:
Angry user reporting! (Not really)

I was just wondering – and I don't know if this is the proper place for this kind of discussion, but I don't mind being redirected to a more appropriate place – about the “better frame timing and interpolation which makes mvtools a lot less desirable” part of @pigoz's comment.

Am I to understand that `mpv` is now capable of  motion interpolation on its own, without the need for `vapoursynth` and `mvtools`? `mpv` has been my player of choice ever since I learned how to use these two tools to interpolate anime to a more watchable frame rate, and I actually very much prefer the effect interpolating to 60 FPS has even on traditional movies and TV shows. 

If there's a way to do this in `mpv` on its own now, then I suppose everything is in order. But if `vapoursynth` and `mvtools` support was dropped and the aforementioned functionality is/will be lost, then I hope you could reconsider this decision.

(I just wish I didn't do `brew update` today, sigh.)

pigoz:
@Cellane we have improved the `:interpolation` suboption for `vo=opengl-hq` and added `video-sync=display-resample`. It's no motion interpolation though. I'll look into adding `vapoursynth` support back.

Cellane:
@pigoz Yey! That's amazing, thank you! For the record, I honestly don't think it matters (to the end-users like me) whether the `mpv`, `vapoursynth` and `mvtools` formulae live in their separate repository and must be tapped before using them, or if they live in the central repository. If the second option is faster and easier to maintain, then I guess that would be the way to go… right?

By the way, there seem to be… well, probably not a lot of people, but still [some who seem to be using this setup as well](https://www.reddit.com/r/osx/comments/38c98i/60_fps_svp_thing_playback_on_os_x/). If that helps :smiley:

Argon-:
I just wanted to add that I greatly appreciate the revert of the removal of VS. I do not only use it occasionally for mvtools based motion interpolation I also use it for "live" tests of stuff on video all the time which is just so convenient.

pigoz:
I managed to clean up the formulae at https://github.com/mpv-player/homebrew-mpv, simplify the python related code, and update the VapourSynth formula.

I will wait again a couple of weeks to get some testing from users, and send a PR with all the formulae from that repo.

Cellane:
Thanks for the information! I updated & upgraded my Homebrew packages (for the first time in month) and after renewing the /usr/local/Cellar/vapoursynth/29/lib/vapoursynthlibmvtools.dylib -> /usr/local/Cellar/mvtools/HEAD/lib/libmvtools.dylib symlink, everything seems to be working perfectly.

(Speaking of the symlink… would it be difficult for the `vapoursynth` formula to create it automatically at the end of the installation if it finds out that `mvtools` is also installed? Just a suggestion, if it's difficult to implement, then please forget it – it's easy enough to renew the symlink after new Vapousynth release manually.)

zmwangx:
@Cellane Of course it's possible to link `libmvtools.dylib` to `$(brew --prefix vapoursynth)/lib/vapoursynth` in `libmvtools`'s `install` block, but I question its legitimacy. It is my understanding that VapourSynth can load plugin from anywhere with the `path` argument of [`std.LoadPlugin`](http://www.vapoursynth.com/doc/functions/loadplugin.html#std.LoadPlugin). Sounds more like a problem of your script.

pigoz:
Exactly, I use `core.std.LoadPlugin(path="/usr/local/lib/libmvtools.dylib")` in my scripts, because relying  on the autoloading is too bothersome.

Cellane:
Ah! I didn't know I can load the plugin from within the script itself, I always did it using the symlink. Apologies for silly suggestion then!

Library/Formula/mpv.rb
```diff
@@ -0,0 +1,72 @@
...
+  depends_on "jpeg"        => :recommended
+  depends_on "little-cms2" => :recommended
+  depends_on "lua"         => :recommended
+  depends_on "youtube-dl"  => :recommended
+
+  depends_on "libcaca"     => :optional
+  depends_on "libdvdread"  => :optional
+  depends_on "libdvdnav"   => :optional
+  depends_on "libbluray"   => :optional
+  depends_on "libaacs"     => :optional
```
mikemcquaid: Don't bother doing all the alignment here and above; we don't really do this with our in-house style.

Library/Formula/mpv.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  WAF_VERSION = "waf-1.8.12"
+  WAF_SHA256  = "01bf2beab2106d1558800c8709bc2c8e496d3da4a2ca343fe091f22fca60c98b"
+
+  resource "waf" do
+    url "https://waf.io/#{WAF_VERSION}"
+    sha256 WAF_SHA256
```
mikemcquaid: Just hard-code these values.

Library/Formula/mpv.rb
```diff
@@ -0,0 +1,72 @@
...
+      ENV.prepend_create_path "PYTHONPATH", buildpath/"lib/python2.7/site-packages"
+      resource("docutils").stage do
+        system "python", "setup.py", "install", "--prefix=#{buildpath}"
+      end
+      ENV["RST2MAN"] = buildpath/"bin/rst2man.py"
+    else
+      args << "--disable-manpage-build"
+    end
+
+    buildpath.install resource("waf").files(WAF_VERSION => "waf")
```
mikemcquaid: You can use `Dir["...*"]` here instead of `WAF_VERSION`

Library/Formula/mpv.rb
```diff
@@ -0,0 +1,72 @@
+class Mpv < Formula
+  desc "Video player based on MPlayer/mplayer2"
+  homepage "https://mpv.io"
+  url "https://github.com/mpv-player/mpv/archive/v0.13.0.tar.gz"
+  sha256 "1372704fd8f5701ef7d60f347fd15fe536e23ef148279bf4a4415e72896912db"
+  head "https://github.com/mpv-player/mpv.git"
+
+  option "with-libmpv",          "Build shared libray."
+  option "without-docs",         "Disable building man page."
```
mikemcquaid: Maybe just worth removing this option unless it takes a really long time.

Library/Formula/mpv.rb
```diff
@@ -0,0 +1,72 @@
+class Mpv < Formula
+  desc "Video player based on MPlayer/mplayer2"
+  homepage "https://mpv.io"
+  url "https://github.com/mpv-player/mpv/archive/v0.13.0.tar.gz"
+  sha256 "1372704fd8f5701ef7d60f347fd15fe536e23ef148279bf4a4415e72896912db"
+  head "https://github.com/mpv-player/mpv.git"
+
+  option "with-libmpv",          "Build shared libray."
```
mikemcquaid: `with-shared` would be a more consistent name. Also `library` rather than `libray`.

Library/Formula/mpv.rb
```diff
@@ -0,0 +1,72 @@
...
+  option "without-docs",         "Disable building man page."
+  option "without-bundle",       "Disable compilation of a Mac OS X Application bundle."
+  option "without-zsh-comp",     "Install without zsh completion"
```
mikemcquaid: Just remove this option.

Library/Formula/mpv.rb
```diff
@@ -0,0 +1,72 @@
+class Mpv < Formula
+  desc "Video player based on MPlayer/mplayer2"
+  homepage "https://mpv.io"
+  url "https://github.com/mpv-player/mpv/archive/v0.13.0.tar.gz"
+  sha256 "1372704fd8f5701ef7d60f347fd15fe536e23ef148279bf4a4415e72896912db"
+  head "https://github.com/mpv-player/mpv.git"
+
+  option "with-libmpv",          "Build shared libray."
+  option "without-docs",         "Disable building man page."
+  option "without-bundle",       "Disable compilation of a Mac OS X Application bundle."
```
mikemcquaid: Default this to off; we prefer to avoid application bundles.

Library/Formula/mpv.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    if build.with? "bundle"
+      ohai "creating a OS X Application bundle"
+      system "python", "TOOLS/osxbundle.py", "build/mpv"
+      prefix.install "build/mpv.app"
+    end
+  end
+
+  test do
+    system "#{bin}/mpv"
```
mikemcquaid: Would be good if you could play one of the test fixtures e.g. `test_fixtures("test.wav")`

zmwangx:
@mikemcquaid If you read the discussion above (especially https://github.com/Homebrew/homebrew/pull/46413#issuecomment-166148475), you'll realize that the `mpv` formula is currently being cleaned up (or rather, already cleaned up) by @pigoz at https://github.com/mpv-player/homebrew-mpv/blob/master/Formula/mpv.rb, and some of the aforementioned problems are already gone. So it's kind of a waste of time to improve this PR. Maybe this one should be closed and let's just wait for @pigoz to submit a new one when he thinks it's ready?

mikemcquaid:
> Maybe this one should be closed and let's just wait for @pigoz to submit a new one when he thinks it's ready?

Sounds good :+1:

Library/Formula/mpv.rb
pigoz: How does `Dir["...*"]` work? I tried it locally but does not seem to work.

mikemcquaid: grep other formula for `Dir[` to see the usage.

PR #46414, opened 2015-11-27
pazpar2: add test

DomT4:
Follow up to https://github.com/Homebrew/homebrew/pull/46374.

The improved test hung for 30+ minutes in that PR, so was removed. Re-adding that here with minor changes, and hopefully won't time out the CI this time.

### DIFF
--- a/Library/Formula/pazpar2.rb
+++ b/Library/Formula/pazpar2.rb
@@ -20,4 +20,28 @@ class Pazpar2 < Formula
                           "--prefix=#{prefix}"
     system "make", "install"
   end
+
+  test do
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Looks like it timed out again :sob:

DomT4:
Womp. It passes locally every time, but fails consistently on the CI. I'll try writing something else.

DomT4:
I have no idea here. This seems to pretty much the simplest test available, it's not doing anything elaborate in the slightest, and it passes locally within seconds every time. I'm at a loss, so closing for now.

PR #46432, opened 2015-11-27
camlp4: bump to 4.02.3+6 to keep in sync with OCaml version

avsm:
Although the distribution here is the same as 4.02.2+6, it must
be recompiled with the version of OCaml that is installed.
OCaml is now 4.02.3 in the tree, and @dsheets reports a failure
due to having a camlp4 that was not recompiled after the OCaml
compiler was updated.

### DIFF
--- a/Library/Formula/camlp4.rb
+++ b/Library/Formula/camlp4.rb
@@ -2,10 +2,9 @@ class Camlp4 < Formula
   desc "Tool to write extensible parsers in OCaml"
   homepage "https://github.com/ocaml/camlp4"
   url "https://github.com/ocaml/camlp4/archive/4.02+6.tar.gz"
-  version "4.02.2+6"
+  version "4.02.3+6"
   sha256 "820c35b69fdff3225bda6045fabffe5d7c54dda00fb157444ac8bda5e1778d45"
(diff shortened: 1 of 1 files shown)


Library/Formula/camlp4.rb
```diff
@@ -2,7 +2,7 @@ class Camlp4 < Formula
   desc "Tool to write extensible parsers in OCaml"
   homepage "https://github.com/ocaml/camlp4"
   url "https://github.com/ocaml/camlp4/archive/4.02+6.tar.gz"
-  version "4.02.2+6"
+  version "4.02.3+6"
   sha256 "820c35b69fdff3225bda6045fabffe5d7c54dda00fb157444ac8bda5e1778d45"
   head "https://github.com/ocaml/camlp4.git", :branch => "trunk"
   revision 1
```
DomT4: Need to kill `revision` lines outside the bottle block on update, Thanks!

avsm: Pushed a fix, thanks!

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/060bcd935cb0b3ccf889f6a9a1ab4add17ab67b6. Thank you for another Homebrew contribution @avsm! :smiley_cat:

PR #46433, opened 2015-11-27
Add missing require to formula.rb

ipsi:
Missing require statement breaks outdated checks in formula.rb (and hence it seems to break Chef as well).


mikemcquaid:

PR #46434, opened 2015-11-27
upscaledb 2.1.12 (rename)

dunn:
`install_name_tool` is failing to fix the vendored libuv; I've only dealt with this problem once before and I forget how to fix it, so cc @mistydemeo and @UniqMartin I guess.

### DIFF
--- a/Library/Formula/upscaledb.rb
+++ b/Library/Formula/upscaledb.rb
@@ -1,15 +1,8 @@
-class Hamsterdb < Formula
+class Upscaledb < Formula
   desc "Database for embedded devices"
-  homepage "http://hamsterdb.com"
-  url "http://files.hamsterdb.com/dl/hamsterdb-2.1.10.tar.gz"
-  sha256 "8fced5607847e234133b2baa379e7ff3d763ffa51546b27f1e3c5f1aef33493e"
-
-  head do
-    url "https://github.com/cruppstahl/hamsterdb.git"
-    depends_on "automake" => :build
-    depends_on "autoconf" => :build
-    depends_on "libtool" => :build
-  end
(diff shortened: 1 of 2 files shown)


dunn:
https://travis-ci.org/Homebrew/homebrew/jobs/93571224#L744-L748

UniqMartin:
The problem is that `libuv.dylib` is built without a dylib ID. Consequently, when `upscaledb` is built, there's no path that can be recorded in the binaries that depend on `libuv.dylib`. This would be normally fixed by the logic in `keg_relocate.rb`, but because `libuv.dylib` is not in `lib/`, it cannot be found and thus the reference to the library not fixed. Hence the warning.

The cleanest way to fix this is to build `libuv.dylib` with a dylib ID (install name) already embedded. The way to do this is (it seems) to add the argument `SO_LDFLAGS=-Wl,-install_name,#{libexec}/libuv/lib/libuv.dylib` to the `make libuv.dylib` invocation. To be more precise:

```patch
diff --git i/Library/Formula/upscaledb.rb w/Library/Formula/upscaledb.rb
index 10b97b49..8d982f40 100644
--- i/Library/Formula/upscaledb.rb
+++ w/Library/Formula/upscaledb.rb
@@ -57,7 +57,7 @@ class Upscaledb < Formula

     if build.with? "remote"
       resource("libuv").stage do
-        system "make", "libuv.dylib"
+        system "make", "libuv.dylib", "SO_LDFLAGS=-Wl,-install_name,#{libexec}/libuv/lib/libuv.dylib"
         (libexec/"libuv/lib").install "libuv.dylib"
         (libexec/"libuv").install "include"
       end
```

dunn:
:ok_woman:  thanks!

PR #46437, opened 2015-11-27
audit fixes: F 02: findutils - fmdiff

apjanke:


apjanke:
Sneaking in the last of my old audit fixes during the Thanksgiving lull.

PR #46448, opened 2015-11-28
CUDA requirement

dpo:
Usage: `depends_on :cuda`, or `depends_on :cuda => "7.5"`. Within `test` methods, one can also use `ENV.cuda`, similarly to `ENV.fortran`.

### DIFF
--- a/Library/Homebrew/extend/ENV/shared.rb
+++ b/Library/Homebrew/extend/ENV/shared.rb
@@ -202,6 +202,12 @@ module SharedEnvExtension
     self["PATH"] = new_paths.uniq.join(File::PATH_SEPARATOR)
   end
 
+  def cuda
+    nvcc = which("nvcc")
+    self["CUDACC"] = nvcc
+    prepend_create_path "DYLD_LIBRARY_PATH", nvcc.dirname.parent/"lib"
+  end
+
   def fortran
     flags = []
 
--- a/Library/Homebrew/dependency_collector.rb
+++ b/Library/Homebrew/dependency_collector.rb
@@ -114,6 +114,7 @@ class DependencyCollector
     when :ant        then ant_dep(spec, tags)
     when :apr        then AprRequirement.new(tags)
     when :emacs      then EmacsRequirement.new(tags)
+    when :cuda       then CudaRequirement.new(tags)
     # Tiger's ld is too old to properly link some software
     when :ld64       then LD64Dependency.new if MacOS.version < :leopard
     when :python2
--- a/Library/Homebrew/requirements.rb
+++ b/Library/Homebrew/requirements.rb
@@ -13,6 +13,7 @@ require "requirements/tuntap_requirement"
 require "requirements/unsigned_kext_requirement"
 require "requirements/x11_requirement"
 require "requirements/emacs_requirement"
+require "requirements/cuda_requirement"
 
 class XcodeRequirement < Requirement
   fatal true
--- a/Library/Homebrew/requirements/cuda_requirement.rb
+++ b/Library/Homebrew/requirements/cuda_requirement.rb
@@ -1,0 +1,37 @@
+class CudaRequirement < Requirement
+  fatal true
+  cask "cuda"
+
+  def initialize(tags)
+    @version = tags.shift if /\d+\.*\d*/ === tags.first
+    super
+  end
+
+  satisfy :build_env => false do
+    next false unless which "nvcc"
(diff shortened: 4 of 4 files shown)


mikemcquaid:
I think this should probably live as a `Requirement` in the formula and/or science until we have a few that use it. We don't want to add a new `requirement` in core for a single formula in a tap.

dpo:
It's lived in homebrew/science for a while already:
- https://github.com/Homebrew/homebrew-science/blob/master/beagle.rb
- https://github.com/Homebrew/homebrew-science/blob/master/pcl.rb

Other formulas optionally assume that `nvcc` is available (they should say really `depends_on :cuda => :optional`):

- https://github.com/Homebrew/homebrew-science/blob/master/opencv.rb#L103
- https://github.com/Homebrew/homebrew-science/blob/master/opencv3.rb#L132

I'd like to write formulas for a number of other cuda-related libraries. I thought this requirement would be better off here to avoid code duplication there. Using `:cuda` like in this PR is infinitely more convenient than importing a `Requirement` on a formula-by-formula basis.

mikemcquaid:
Could this Requirement not live in homebrew-science itself? Are there any core formulae that would use it?

dpo:
It could live in homebrew-science, thought it would be slightly less convenient. I don't see core formulae using gpu computing at the moment, but I only did a cursory search. Something I didn't find how to do with a tap-specific requirement is [this](https://github.com/Homebrew/homebrew-science/pull/3064/files#diff-a701addc1a725323f9cda3848822b8b8R18). We'll want to test gpu code, and so have access to the appropriate environment (as in `ENV.fortran`). Is there a way to achieve that?

Library/Homebrew/extend/ENV/shared.rb
```diff
@@ -202,6 +202,11 @@ def userpaths!
     self["PATH"] = new_paths.uniq.join(File::PATH_SEPARATOR)
   end
 
+  def cuda
+    self["CUDACC"] = which("nvcc")
+    prepend_create_path "DYLD_LIBRARY_PATH", which("nvcc").dirname.parent/"lib"
```
bfontaine: You could reuse `self["CUDACC"]` here to avoid another call to `which("nvcc")`.

mikemcquaid:
> Something I didn't find how to do with a tap-specific requirement is this.

You can set the environment inside a requirement (and probably should be in this case too).

mikemcquaid:
See e.g. https://github.com/Homebrew/homebrew/blob/1eba76d0778e5055775ff9bf7f458318410a550b/Library/Homebrew/requirements/emacs_requirement.rb#L17-L20

dpo:
@mikemcquaid I think [I do](https://github.com/Homebrew/homebrew/pull/46448/files#diff-b2b0e7df3d8d1d3dd38d2baf9abed2d9R20), but that environment goes away when `install` exits. So it's not available in `test`. Or did I misunderstand?

Library/Homebrew/extend/ENV/shared.rb
dpo: Yes, thanks!

mikemcquaid:
That's somewhat expected, yeh. You need to manually change the environment in tests.

dpo:
Yes, that's one of the reasons why it's convenient to have this kind of requirement in core, but I do understand the concern.

mikemcquaid:
That's a bit of a feature rather than a bug in tests; the test code should make it clear what users have to do to achieve the same behaviour.

dpo:
Agreed. I wasn't suggesting it's a bug.

tdsmith:
I think this isn't likely to be useful in core and could be maintained in homebrew-science; `requires "requirements/cuda.rb"`seems like an okay amount of overhead.

dpo:
That's too bad. I'll leave it in science then. A few questions:
1. is there a "standard" way to implement requirements in a tap? Several science formulas have a requirement coded up in the same file as the formula but it would be best to factor them out.
2. is there still a way to say `needs :cuda` or `depends_on :cuda` for a requirement in a tap?
3. how about `ENV.cuda`?

tdsmith:
Homebrew-php does some slightly wacky things but they do have Requirements: https://github.com/Homebrew/homebrew-php/tree/master/Requirements

I don't think you can get symbol dependencies in a tap.

Because Ruby, you can probably reopen any namespace you want in the body of the requirement Ruby file!

PR #46457, opened 2015-11-28
Khal 0.7.0

benjaminfrank:


Library/Formula/khal.rb
```diff
@@ -89,14 +89,15 @@ class Khal < Formula
   end
 
   def install
-    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python2.7/site-packages"
+    version = Language::Python.major_minor_version "python"
+    ENV.prepend_create_path "PYTHONPATH", libexec/"vendor/lib/python#{version}/site-packages"
```
mikemcquaid: What does this change fix?

benjaminfrank: Nothing really.
I had some strange runtime error, so I wanted to try if the same error comes up with python3 (it did as it was a config issue on my end).
however, python2 support might be dropped in future, so i copied this snipped from my vdirsyncher.rb which uses python3.

mikemcquaid: Let's just leave it as-is for now as it only supports a single Python version.

benjaminfrank: How can I support multiple python versions? The tool runs under both and something is (slightly) better than python3?

DomT4: [This sort of thing](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/z3.rb).

benjaminfrank: Nice thanks. I'll push a python2 version for now. maybe next release I'll put in both.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/e3ea87bbb264ef1989fe81b9d40a433c992c94f5. Thank you for another Homebrew contribution @benjaminfrank! :smiley_cat:

PR #46458, opened 2015-11-28
Update mupdf tools url

adlermedrado:
The URL was incorrect on Formula, so I need to update it pointing to the right place.

### DIFF
--- a/Library/Formula/mupdf-tools.rb
+++ b/Library/Formula/mupdf-tools.rb
@@ -1,7 +1,7 @@
 class MupdfTools < Formula
   desc "Lightweight PDF and XPS viewer"
   homepage "http://mupdf.com"
-  url "http://mupdf.com/downloads/archive/mupdf-1.7a-source.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
It looks like the URL was moved because there's a new version. You could do the version bump instead,  if you'd like to.

mikemcquaid:
> It looks like the URL was moved because there's a new version. You could do the version bump instead, if you'd like to.

That'd be good :+1:

adlermedrado:
Sure, will do.

DomT4:
Thanks!

tomafro:
I've bumped the version here: https://github.com/Homebrew/homebrew/pull/46669

chdiza:
We can in future avoid the disappearing older versions problem by switching the base of the URL to `http://mupdf.com/downloads/archive/`, where the newest versions AND the older ones are all stored.  This way, when we're caught between the time of new release and the time where someone *realizes* there's a new release, the formula will still work.

Issue #46422, opened 2015-11-27
Problem using openmpi with fortran


PierreMarchand20:
Hello,

I was using openmpi with fortran without any problem until a few weeks ago when I upgraded openmpi (brew upgrade). Since then, I have the following message when I use mpif90, mpif70 and mpifort : 

No underlying compiler was specified in the wrapper compiler data file
(e.g., mpicc-wrapper-data.txt)

So I don't know why it has lost the compiler and I have tried to uninstall openmpi et reinstall it. I certainly don't use the option --without-fortran ! Any ideas ?

zbeekman:
Yes, I'm experiencing this same problem trying to submit a formula that requires MPI and Fortran. I seem to be able to build things OK locally, but on the test-bot, without re-triggering an OpenMPI build in a very ugly hack, I'm getting the same error. See PR #46547

dpo:
See #46461.

PierreMarchand20:
The issue has been fixed with the last update (see #46461 )

PR #46461, opened 2015-11-28
open-mpi: bump revision to rebottle.

dpo:
We're having lots of MPI-related issues over at homebrew-science. It turns out the current bottles are incomplete:
```
$ mpif77 -v
--------------------------------------------------------------------------
No underlying compiler was specified in the wrapper compiler data file
(e.g., mpicc-wrapper-data.txt)
--------------------------------------------------------------------------
```
so many formulae fail to build. Building from source solves the issue, so I'm hoping re-bottling is the solution.

### DIFF
--- a/Library/Formula/open-mpi.rb
+++ b/Library/Formula/open-mpi.rb
@@ -3,6 +3,7 @@ class OpenMpi < Formula
   homepage "https://www.open-mpi.org/"
   url "https://www.open-mpi.org/software/ompi/v1.10/downloads/openmpi-1.10.1.tar.bz2"
   sha256 "7919ecde15962bab2e26d01d5f5f4ead6696bbcacb504b8560f2e3a152bfe492"
+  revision 1
 
   bottle do
     sha256 "257a75a3227ddc0b5e7c970a97c76cadeda417a259c54efbe25df955e0efe32b" => :el_capitan
@@ -83,5 +84,19 @@ class OpenMpi < Formula
     system "#{bin}/mpicc", "hello.c", "-o", "hello"
     system "./hello"
     system "#{bin}/mpirun", "-np", "4", "./hello"
+    (testpath/"hellof.f90").write <<-EOS.undent
+      program hello
+      include 'mpif.h'
+      integer rank, size, ierror, tag, status(MPI_STATUS_SIZE)
+      call MPI_INIT(ierror)
+      call MPI_COMM_SIZE(MPI_COMM_WORLD, size, ierror)
+      call MPI_COMM_RANK(MPI_COMM_WORLD, rank, ierror)
+      print*, 'node', rank, ': Hello Fortran world'
+      call MPI_FINALIZE(ierror)
+      end
+    EOS
+    system "#{bin}/mpif90", "hellof.f90", "-o", "hellof"
+    system "./hellof"
+    system "#{bin}/mpirun", "-np", "4", "./hellof"
   end
 end


dpo:
The new test is exposing the same issue on the bot. It works when I build locally. What could be going on here?

DomT4:
If you're using Fortran in the test you may need to pass `ENV.fortran`. If you're not, not entirely sure. 

Apologies for not getting back to the Science team publicly on the issues around special dependencies - I raised the issue to the rest of the maintainer team in chat the other day and we discussed it a little bit. It's a bit of mystery as to why the CI falls apart on things we can't replicate locally.

dpo:
It's not clear why you'd need `ENV.fortran` when running `mpif90` because the link to `gfortran` should be recorded inside the MPI installation. But I'll try inserting a `ENV.f90` into the test.

dpo:
Actually it fails in exactly the same way.

DomT4:
Well, I didn't promise that it would fix the problem, just wanted to narrow that down as a variable.

Any thoughts @apjanke? I thought I saw you comment on a special dependency issue the other day? Apologies if mistaken.

dpo:
No, no, I understand. I think I'll revert the `ENV.fortran` as it seems unnecessary when calling `mpif90`.

mikemcquaid:
Test is still failing.

dpo:
Yes, it was failing before too, but the problem is elsewhere. Somehow the Fortran compiler isn't being "remembered" by `mpif77` and `mpif90`.

apjanke:
Yeah, this looks like it could be similar to that issue in #45730.

```
  depends_on :fortran => :recommended
```

When `:fortran` is only a `:recommended` dependency, it will not cause `gcc` to be automatically installed. And the test bots / CI machines don't have `gcc` installed by default, IIRC, so `gfortran` won't exist on those machines. So this could be causing `open-mpi` to end up getting bottled without Fortran support, even though it's `:recommended` and should be included by default.

This is a bug in `brew`'s dependency resolution for `:recommended` requirements, not an issue in the individual formula.

Here's build logs from when I uninstalled GCC from one of my test machines and did a `brew install --build-from-source open-mpi`. With that, `brew test open-mpi` had the same failures as you're seeing here. Installing `gcc` and re-installing `open-mpi` fixed it.
https://gist.github.com/apjanke/e90b54d043653810cc58

Looks like #46422 is the same issue, too.

As a workaround, you could make `:fortran` an unconditional dependency. Doesn't seem like that would be much of a burden for users, since the main users of `open-mpi` would be developers, who may well have `gcc` installed regardless, right?

Also, if we're bumping `revision` just to get a new bottle, can that `revision` go inside the `bottle` block instead of being applied to the whole formula?

wence-:
FWIW, I'm seeing this too.  Additionally, I can't make the `build-from-source` option to work either.  Inside the build environment, gfortran is not available (even after installing gcc).

Steps to reproduce:

```
$ env | grep FC
$ ls -l $(which gfortran)
lrwxr-xr-x  1 xxxxxx  admin  32  1 Dec 14:25 /usr/local/bin/gfortran -> ../Cellar/gcc/5.2.0/bin/gfortran
$ brew install -i --ignore-dependencies --build-from-source openmpi --with-fortran
==> Downloading https://www.open-mpi.org/software/ompi/v1.10/downloads/openmpi-1
Already downloaded: /Library/Caches/Homebrew/open-mpi-1.10.1.tar.bz2
==> Entering interactive mode
Type `exit' to return and finalize the installation
Install to this prefix: /usr/local/Cellar/open-mpi/1.10.1
bash-3.2$ gfortran --version
bash: gfortran: command not found
bash-3.2$ echo $PATH
/usr/local/Library/ENV/4.3:/usr/bin:/bin:/usr/sbin:/sbin
bash-3.2$ ls -l ${PATH/:*/}
total 376
-rwxr-xr-x  1 xxxxxxxx  admin   156 23 Nov  2013 ant
-rwxr-xr-x  1 xxxxxxxx  admin   362 24 Jul  2013 apr-1-config
-rwxr-xr-x  1 xxxxxxxx  admin    77 24 Jan  2015 bsdmake
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 c++ -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 c89 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 c99 -> cc
-rwxr-xr-x  1 xxxxxxxx  admin  8258  3 Nov 16:31 cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 clang -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 clang++ -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 cpp -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 g++ -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 g++-4.2 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 g++-4.3 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 g++-4.4 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 g++-4.5 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 g++-4.6 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 g++-4.7 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 g++-4.8 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 g++-4.9 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul 10:44 g++-5 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 gcc -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 gcc-4.2 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 gcc-4.3 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 gcc-4.4 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 gcc-4.5 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 gcc-4.6 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 gcc-4.7 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 gcc-4.8 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 23 Nov  2013 gcc-4.9 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul 10:44 gcc-5 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin    10  8 Oct  2014 git -> ../scm/git
lrwxr-xr-x  1 xxxxxxxx  admin     4 24 Jul  2013 gmake -> make
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 i686-apple-darwin11-llvm-g++-4.2 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 i686-apple-darwin11-llvm-gcc-4.2 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 ld -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 llvm-g++ -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 llvm-g++-4.2 -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 llvm-gcc -> cc
lrwxr-xr-x  1 xxxxxxxx  admin     2 24 Jul  2013 llvm-gcc-4.2 -> cc
-rwxr-xr-x  1 xxxxxxxx  admin    74 10 Jan  2014 make
-rwxr-xr-x  1 xxxxxxxx  admin    96 24 Jul  2013 mig
-rwxr-xr-x  1 xxxxxxxx  admin   154 24 Jan  2015 pod2man
-rwxr-xr-x  1 xxxxxxxx  admin   183 24 Jul  2013 sed
lrwxr-xr-x  1 xxxxxxxx  admin    10  8 Oct  2014 svn -> ../scm/svn
-rwxr-xr-x  1 xxxxxxxx  admin  1157 16 Jul  2014 xcrun
```

So inside the build shell, gfortran is unavailable even though the gcc package is installed.  I'm not sure how to fix this, I don't really understand how this SuperEnv stuff works.

dpo:
Does it work with `--env=std`?

wence-:
Yes, because then the containing environment (including PATH) is in place and gfortran is found.  However, this is not really a "fix", right?

dpo:
It's not a "fix" but it points in the direction of superenv.

wence-:
OK, scrap that, passing ignore-dependencies means that the `depends-on :fortran` line in the formula is never executed and FC is never passed in to the environment.  After which sequence of events, I fall back to agreeing with apjanke's last comment.  The lack of a required :fortran dependency means that if gcc is not installed, even if I explicitly say "--with-fortran", fortran support is not compiled in.

apjanke:
Yeah. I think you can see the difference in behavior by linking and unlinking `gcc` before doing that `install -i`.

```
$ brew unlink gcc
Unlinking /usr/local/Cellar/gcc/5.2.0... 30 symlinks removed
[~]
$ brew install -i  --build-from-source openmpi --with-fortran
Warning: The sandbox cannot be used in debug or interactive mode.
==> Downloading https://www.open-mpi.org/software/ompi/v1.10/downloads/openmpi-1.10.1.tar.bz2
Already downloaded: /Library/Caches/Homebrew/open-mpi-1.10.1.tar.bz2
==> Entering interactive mode
Type `exit' to return and finalize the installation
Install to this prefix: /usr/local/Cellar/open-mpi/1.10.1
eilonwy% which gfortran
gfortran not found
eilonwy% exit
Aborting due to non-zero exit status
[✘ ~]
$ brew link gcc
Linking /usr/local/Cellar/gcc/5.2.0... 30 symlinks created
[~]
$ brew install -i  --build-from-source openmpi --with-fortran
Warning: The sandbox cannot be used in debug or interactive mode.
==> Using Homebrew-provided fortran compiler.
This may be changed by setting the FC environment variable.
==> Downloading https://www.open-mpi.org/software/ompi/v1.10/downloads/openmpi-1.10.1.tar.bz2
Already downloaded: /Library/Caches/Homebrew/open-mpi-1.10.1.tar.bz2
==> Entering interactive mode
Type `exit' to return and finalize the installation
Install to this prefix: /usr/local/Cellar/open-mpi/1.10.1
eilonwy% which gfortran
/usr/local/bin/gfortran
eilonwy%
```

It finds the `gfortran` binary if `gcc` is linked, but doesn't attempt to install anything if it isn't, even without `--ignore-dependencies`.

zbeekman:
Well, I'm sad that this issue has yet to be resolved, however, I am super happy that my issue over at PR #46547 seems to be deeper than just my poor understanding of Ruby/Homebrew!

I'm not sure how helpful this is, but if I add:

```ruby
depends_on :mip => [:cc, :fc, "with-fortran"]
```

This will trigger a rebuild of OpenMPI on the test bot. It has the effect of causing all tests to pass on Jenkins and one to pass on Travis, while the other two time out.

apjanke:
> depends_on :mip => [:cc, :fc, "with-fortran"]

I think this is basically the same thing as removing the `:recommended` from the `depends_on :fortran => :recommended`. It's another way of specifying an unconditional fortran dependency, just indirectly in this case.

zbeekman:
I just performed another experiment... it seems that this is a necessary but not sufficient condition to get the build to work. It seems I also need to try to force the OpenMPI build with a brewed GCC... still poking around.

zbeekman:
FYI, This seems to pass, at least if Travis-CI doesn't time-out first. It takes FOREVER to build OpenMPI, though: https://github.com/Homebrew/homebrew/pull/46624

mikemcquaid:
This is the same issue mentioned in https://github.com/Homebrew/homebrew/pull/46461

michoski:
We are having the same issue, and it has been persisting for some time.   We are actually having some sort of linking problem with gcc as well.   I think after install the /usr/local/bin/gcc should be linked, but on our latest build, it does not get linked; rather 'which gcc' gives /usr/bin/gcc.  And of course, we are getting the same fortran problems as well, with mpif90 being absent.

We actually have tried older brews, and they are not working now either.  Can I ask: what is the recommended cleanest way to get a fully functioning version running?  I mean, are we at just uninstalling homebrew and building everything locally?

mikemcquaid:
> Can I ask: what is the recommended cleanest way to get a fully functioning version running?

Of what, sorry?

michoski:
Of versions of gcc and open-mpi in homebrew that work and link correctly?

mikemcquaid:
> Of versions of gcc and open-mpi in homebrew that work and link correctly?

Open MPI has some current issues. GCC does link correctly; it's installed as `/usr/local/bin/gcc-5`.

michoski:
For us, gcc-5 installs correctly but does not link.  Tried installing and uninstalling multiple times.  The native version is what 'which gcc' gives:
/usr/bin/gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 7.0.0 (clang-700.0.72)
Target: x86_64-apple-darwin14.5.0
Thread model: posix

mikemcquaid:
@michoski Please reread my message: `which gcc` will not return it but `which gcc-5` will.

michoski:
I realize that.  So this is strange behavior.  It use to install, in older versions of homebrew, as just gcc (not with an appended version number).  Having multiple versions of gcc installed is not in my opinion very helpful, for trying to build packages, such as petsc, that look automatically for these things in the environment.  It just easily leads to conflicts and mismatching.  Is there some reason for it?

We have things working with just aliasing the gcc-5 back to gcc, but aren't sure this will iron out all the possible problems this could cause.  Do you have a suggestion for how to make gcc-5 appear as native everywhere?  Is this enough?

I originally suspected this might have something to do with the opem-mpi problem actually: "No underlying compiler was specified in the wrapper compiler data file"

zbeekman:
> I originally suspected this might have something to do with the opem-mpi problem actually: "No underlying compiler was specified in the wrapper compiler data file"

I have doubts that this is related, but I'm no expert.

If you really want to use  GCC-5 to build EVERYTHING built by Homebrew, you can try something like

```bash
export HOMEBREW_CC=gcc-5
export HOMEBREW_CXX=g++-5
export HOMEBREW_FC=gfortran-5
```

in `.bashrc` or similar

mikemcquaid:
@zbeekman Please don't recommend that, it's a bad idea so I removed it from the message. Thanks!

michoski:
But fortran is not included in the native gcc install for OSX, I don't think.  So anything, like open-mpi, that might be looking for gfortran, is going to find a big empty space there ...?  You have to look for gfortran-5, or some such instead?  I mean, I can set these env variables, but why is this not already being done?   I was under the impression homebrew was trying to be fairly well-integrated.

zbeekman:
Sorry, it wasn't meant to be a recommendation per-se.

@michoski disregard the second part of my comment please...

UniqMartin:
@michoski There's no such thing as a native GCC install on OS X. A long time ago there was, before Clang (on top of LLVM) became Apple's only native compiler. Now and back then there was no Fortran support. Even today there's still a binary named `gcc`, but as you can see from the `gcc --version` output that's not really GCC but Clang instead. It's only there for ancient/stubborn build systems that expect to find a `gcc` binary on OS X, so mostly a compatibility measure.

*A few more words about GCC/GFortran as provided by Homebrew:* Homebrew tries to work with the native tools as much as possible, that's why it doesn't shadow `gcc` and `g++` even with the `gcc` formula installed. As you correctly noticed, you can access them using a version suffix via `gcc-5` and `g++-5`. Unlike for C/C++, there is no native Fortran compiler on OS X, so Homebrew fills that gap by providing both `gfortran-5` and `gfortran`. Comparable to many Linux distributions, Homebrew is perfectly able to install multiple GCC major releases in parallel and there are valid use cases for this (though they are rare). Providing multiple GCC installations with a version suffix allows for peaceful coexistence and is common practice, also in Linux land. (I hope that helped to clarify the situation.)

However, all of this is completely unrelated to the issue this PR tries to solve. If you still believe there's an issue with how Homebrew distributes GCC, please open a new issue. Thank you!

UniqMartin:
I believe I found the underlying issue. Solving it properly might be tricky, though. #46680 shows a workaround. It's not a great solution, but not horribly bad either. @dpo Feel free to pick the relevant two lines from my test PR. I “stole” everything else from you (and from this PR) anyway.

Here's an attempt to briefly explain what happens:

- `depend_on :fortran => :recommended` is a requirement and at some stage in `brew install --only-dependencies` it gets transformed into a regular dependency, that is more or less equivalent to `depend_on "gcc" => :recommended`.
- Like every other optional/recommended dependency, it is now matched against the build options. The option known to the `open-mpi` formula is still either `--with-fortran` or `--without-fortran`, but because the dependency is now named `gcc` the check in `brew install --only-dependencies` looks for `--with-gcc` or `--without-gcc` instead, both of which are unknown.
- Consequently, the `gcc` dependency is pruned and there's no `gfortran` binary during the build.

(The [travis CI log](https://travis-ci.org/Homebrew/homebrew/jobs/93876575#L286) also shows this discrepancy between the `brew install --only-dependencies` output and the `brew fetch` invocation above this, that is supposed to fetch the dependencies.)

mikemcquaid:
The dependency issue is fixed in https://github.com/Homebrew/homebrew/issues/46916. If you rebase this pull request it should be sorted.

zbeekman:
:tada: Thanks to all the folks at Homebrew for getting this sorted.

dpo:
Many thanks! PR rebased.

dpo:
There's an error on Travis/Mavericks: https://travis-ci.org/Homebrew/homebrew/jobs/97037667#L817. Not sure what it's due to.

dpo:
Same error on the bot, Mavericks only. Any ideas?
```
==> Verifying open-mpi-1.10.1.tar.bz2 checksum
tar xf /Library/Caches/Homebrew/open-mpi-1.10.1.tar.bz2
Error: wrong number of arguments (2 for 1)
```

UniqMartin:
There was a hiccup with our infrastructure that is now fixed, so the build finally finished successfully. Thanks everyone for enduring the long wait while we (mostly Mike) sorted that out!

Merged in 57466d992187c120f4a2cf1882454ec2f9388ff3. Thanks for another much appreciated contribution to Homebrew, @dpo! :tada:

zbeekman:
:bow:

dpo:
Many thanks to all! Fixing this is major. We really appreciate all your hard work.

dpo:
Well I'm sorry to report that it seems the trouble aren't entirely resolved. I just tried to rebuild https://github.com/Homebrew/homebrew-science/pull/3098 and https://github.com/Homebrew/homebrew-science/pull/3002. The first says

```
The Open MPI wrapper compiler was unable to find the specified compiler
/usr/local/bin/gfortran in your PATH.

Note that this compiler was either specified at configure time or in
one of several possible environment variables.
```
and the second
```
checking for mpicc... no
checking for hcc... no
checking for mpcc... no
checking for mpcc_r... no
checking for mpxlc... no
checking for cmpicc... no
checking for MPI_Init... no
checking for MPI_Init in -lmpi... no
checking for MPI_Init in -lmpich... no
configure: error: could not find mpi library for --enable-mpi
```

UniqMartin:
> Well I'm sorry to report that it seems the trouble aren't entirely resolved.

The good news is, we know (see #47040). The bad news is, it's not fixed yet. I'll try to prioritize my work on this issue as it seems to be even more widespread than I thought.

To use `abinit` as an example. Current behavior on a fresh Homebrew installation:

```
$ brew install -v abinit
==> Installing abinit from homebrew/science
==> Installing dependencies for homebrew/science/abinit: sphinx-doc, cmake, veclibfort, scalapack, fftw, libxc, szip, hdf5, netcdf, etsf_io, gsl
==> Installing homebrew/science/abinit dependency: sphinx-doc
[…]
```

Behavior with the fix from #47040 applied:

```
$ brew install -v abinit
==> Installing abinit from homebrew/science
==> Installing dependencies for homebrew/science/abinit: xz, gmp, mpfr, libmpc, isl, gcc, pkg-config, makedepend, openssl, libevent, open-mpi, sphinx-doc, cmake, veclibfort, scalapack, fftw, libxc, szip, hdf5, netcdf, etsf_io, gsl
==> Installing homebrew/science/abinit dependency: xz
```

Notice the massively different (and certainly more complete) list of dependencies.

UniqMartin:
@dpo I just merged #47040 which hopefully fixes the missed dependencies in those failing builds from homebrew-science. Could you give them another try and report back here? Thanks!

PR #46466, opened 2015-11-29
Handle uninstalled packages in outdated_versions

### LINKED ISSUE #46463: `brew info --json=v1 $pkg` broken when $pkg not installed
When running `brew info --json=v1 $pkg`, valid json is not returned if the package is not installed:

```
$ brew info --json=v1 par
Error: No such file or directory - /usr/local/Cellar/par
```

Whereas if I omit the `--json=v1` argument I get:

```
$ brew info par
par: stable 1.52 (bottled)
Paragraph reflow for email
http://www.nicemice.net/par/
Conflicts with: rancid
Not installed
From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/par.rb
```

Is this intended?  If so, what is the best (machine-parseable) way to get information about formulae from `brew`?

hivehand:
This should resolve #46463, which arises from the newly-added `outdated_versions` method triggering an error when asked to process a non-installed formula. The fix just implements @mistydemeo’s suggestion of having `outdated_versions` exit early, returning an empty array, if `rack.exist?` returns false.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -938,6 +938,8 @@ class Formula
         raise Migrator::MigrationNeededError.new(self)
       end
 
+      return [] unless rack.exist?
+
       rack.subdirs.each do |keg_dir|
         keg = Keg.new keg_dir
         version = keg.version


xu-cheng:
I intend to fix this in #46468

mikemcquaid:
Thanks @xu-cheng, @hivehand!

xu-cheng:
Fixed by #46468. Thanks for PR @hivehand.

hivehand:
Thanks for the official fix, @xu-cheng!

PR #46475, opened 2015-11-29
pcre 8.38

vszakats:


dunn:
Test getting blocked by the sandbox; not sure why El Cap isn't: http://bot.brew.sh/job/Homebrew%20Pull%20Requests/37366/#showFailuresLink

vszakats:
Even worse, it's a test of a dependency. This `pcre` update is a purely bugfix release (there won't be any more non-bugfix releases for the `8.*` series), so it's not supposed to create much build trouble in itself.

mikemcquaid:
Thanks @vszakats!

mikemcquaid:
@dunn It'll be that El Cap doesn't have a bottle for that reverse dep so doesn't run the test.

PR #46479, opened 2015-11-29
V8 - added i18n support via use_system_ucu

kynx:
This PR adds support for ECMAScript 420 internationalisation to V8, using brew icu4c.

```
brew install --with-icu4c v8
```

I had to patch `v8.gyp` so the build process finds icu under /usr/local/opt instead of /usr/local. I raised this [upstream](https://groups.google.com/a/chromium.org/forum/#!topic/chromium-packagers/pNpOJ0wh5c8) - maybe we can get this in so a patch isn't required. What fun gyp is :(=

### DIFF
--- a/Library/Formula/v8.rb
+++ b/Library/Formula/v8.rb
@@ -14,6 +14,7 @@ class V8 < Formula
   end
 
   option "with-readline", "Use readline instead of libedit"
+  option "with-icu4c", "Use icu4c for l18n support"
 
   # not building on Snow Leopard:
   # https://github.com/Homebrew/homebrew/issues/21426
@@ -21,6 +22,10 @@ class V8 < Formula
 
   depends_on :python => :build # gyp doesn't run under 2.6 or lower
   depends_on "readline" => :optional
+
+  if build.with? "icu4c"
+    depends_on "icu4c"
+  end
 
   needs :cxx11
 
@@ -55,6 +60,12 @@ class V8 < Formula
         :revision => "9855a87157778d39b95eccfb201a9dc90f6d61c6"
   end
 
+  if build.with? "icu4c"
+    # Patch tools/gyp/v8.gyp so it finds icu4c headers and .dylibs
+    # See https://groups.google.com/a/chromium.org/forum/#!topic/chromium-packagers/pNpOJ0wh5c8
+    patch :DATA
+  end
+
   def install
     # Bully GYP into correctly linking with c++11
     ENV.cxx11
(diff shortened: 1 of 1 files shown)


DomT4:
To be honest I'm going to be reluctant to include this until that patch is actually accepted upstream; `v8` is *very* prone to breaking across updates and I don't want to chuck potentially more fuel on that fire.

Library/Formula/v8.rb
```diff
@@ -14,6 +14,7 @@ class V8 < Formula
   end
 
   option "with-readline", "Use readline instead of libedit"
+  option "with-icu4c", "Use icu4c for l18n support"
```
DomT4: Can probably drop this line. It's fairly obvious what you'd use an `icu4c` dep for.

Library/Formula/v8.rb
```diff
@@ -22,6 +23,10 @@ class V8 < Formula
   depends_on :python => :build # gyp doesn't run under 2.6 or lower
   depends_on "readline" => :optional
 
+  if build.with? "icu4c"
+    depends_on "icu4c"
```
DomT4: Just `depends_on "icu4c" => :optional`

Library/Formula/v8.rb
```diff
@@ -76,8 +87,18 @@ def install
     (buildpath/"testing/gtest").install resource("gtest")
     (buildpath/"tools/clang").install resource("clang")
 
+    i18nsupport="i18nsupport=off"
+    if build.with? "icu4c"
+      i18nsupport="i18nsupport=on"
+      
+      # Patch needs icu-config on our path
+      icu4c = Formula["icu4c"]
+      ENV.append "PATH", icu4c.opt_bin
```
DomT4: This will happen automatically inside `brew`; We deliberately put `keg_only` dependencies into the `$PATH` when they're invoked.

mikemcquaid:
> To be honest I'm going to be reluctant to include this until that patch is actually accepted upstream; v8 is very prone to breaking across updates and I don't want to chuck potentially more fuel on that fire.

Agreed :+1:

kynx:
Totally understand your reluctance! I thought this was going to be a doddle, then I met all them gyps ;)

But i18n, even if static linked, would be great. Didnt work for me, but you ever tried that?

DomT4:
We vendor an `icu` element as part of the `DEPS` because `v8` throws a tantrum otherwise; If that would do the job and doesn't cause breakage, I'm not opposed to us actually linking that in.

kynx:
Yeah, that would be great! 

I'll keep pestering the google guys - the use_system_icu stuff is documented and _should_ work, just doesn't when icu isn't in a standard location. The patch is going to need a bit of work to ensure it doesn't break other builds, but I'm confident I'll get there in the end.

FWIW I'm keen on this so I can get formatjs working inside php-v8js.

kynx:
Upstream have pointed me in the direction of an [icu.gyp](https://chromium.googlesource.com/chromium/src/build/+/master/linux/unbundle/icu.gyp) ubuntu use when building chromium. It wouldn't work here OOTB, but it's undoubtably a less invasive way of doing it (+ would save downloading the icu source).

I'm closing this and will re-open a new PR if I make any progress.

PR #46489, opened 2015-11-29
htpdate: boneyard

dunn:
New versions require `sys/timex.h`


mikemcquaid:
I don't see a reason to necessarily remove it if the current version works. Would be good to file an upstream issue and see if they are receptive to making it portable again.

dunn:
> I don't see a reason to necessarily remove it if the current version works 

OK, I wasn't sure what the policy was for out-of-date formulae that can't be upgraded.

> Would be good to file an upstream issue and see if they are receptive to making it portable again.

Will do.

dunn:
Developer doesn't have access to OSX so he's not going to try to make it portable.  So I guess it stays at 0.91.

PR #46491, opened 2015-11-29
htmlcleaner 2.15

dunn:


Library/Formula/htmlcleaner.rb
```diff
@@ -1,14 +1,15 @@
...
 
-  bottle :unneeded
+  depends_on "maven" => :build
 
   def install
-    libexec.install "htmlcleaner-#{version}.jar"
-    bin.write_jar_script libexec/"htmlcleaner-#{version}.jar", "htmlcleaner"
+    system "mvn", "clean", "package", "-Duser.home=#{buildpath}"
+    libexec.install Dir["target/htmlcleaner-*-SNAPSHOT.jar"]
+    bin.write_jar_script Dir["#{libexec}/htmlcleaner-*-SNAPSHOT.jar"].first, "htmlcleaner"
```
dunn: This `Dir` silliness is because the `.jar` built is called `htmlcleaner-2.16-SNAPSHOT.jar`.  I assume this is just because they made the zipball after updating the version strings.

mikemcquaid: That's fine :+1:

PR #46498, opened 2015-11-30
icarus-verilog 10.0

dunn:
Dunno why it jumped from 0.9.7 to 10.0 but hey.


dunn:
The single Jenkins failure sortof looked like a race condition, reported at https://github.com/steveicarus/iverilog/issues/85

PR #46510, opened 2015-11-30
shark: migrate from boneyard to science

bfontaine:
See https://github.com/Homebrew/homebrew-boneyard/pull/78 and https://github.com/Homebrew/homebrew-science/pull/3074.


xu-cheng:
This formula may be more suitable to submit to science tap.

bfontaine:
Good idea!

PR #46512, opened 2015-11-30
configuration file for rethinkdb

meehow:
Hi,
Rethinkdb http admin console listens by default on port 8080, what can cause a conflict with other tools (like node-inspector) which are even not checking if port is already in use.
Right now port can be configured in plist file, but it would be easier and more transparent to have a configuration file with available config options.

### DIFF
--- a/Library/Formula/rethinkdb.rb
+++ b/Library/Formula/rethinkdb.rb
@@ -32,6 +32,81 @@ class Rethinkdb < Formula
     system "make", "install-osx"
 
     (var/"log/rethinkdb").mkpath
+
+    (buildpath+"rethinkdb.conf").write rethinkdb_conf
+    etc.install "rethinkdb.conf"
+  end
+
+  def rethinkdb_conf; <<-EOS.undent
+    directory=#{var}/rethinkdb
+
+    ### Network options
+
+    ## Address of local interfaces to listen on when accepting connections
+    ## May be 'all' or an IP address, loopback addresses are enabled by default
+    ## Default: all local addresses
+    bind=127.0.0.1
+
+    ## Address that other rethinkdb instances will use to connect to this server.
+    ## It can be specified multiple times
+    # canonical-address=
+
+    ## The port for rethinkdb protocol for client drivers
+    ## Default: 28015 + port-offset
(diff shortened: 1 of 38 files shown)


bfontaine:
Does upstream not provide a similar example config file?

meehow:
It does, but it doesn't set `directory`. I also changed `bind` to `127.0.0.1` because many homebrew daemons are configured this way by default.

meehow:
Travis just timed out...
```
==> make

..........................

The build has been terminated
```

Library/Formula/rethinkdb.rb
```diff
@@ -32,6 +32,81 @@ def install
     system "make", "install-osx"
 
     (var/"log/rethinkdb").mkpath
+
+    (buildpath+"rethinkdb.conf").write rethinkdb_conf
+    etc.install "rethinkdb.conf"
```
mikemcquaid: Does `rethinkdb` have a sample configuration we can copy/modify rather than storing it all in the formula?

Library/Formula/rethinkdb.rb
```diff
@@ -32,6 +32,81 @@ def install
...
+    # io-threads=64
+
+    ## Enable direct I/O
+    # direct-io
+
+    ### Meta
+
+    ## The name for this server (as will appear in the metadata).
+    ## If not specified, it will be randomly chosen from a short list of names.
+    # server-name=server1
```
mikemcquaid: We should remove everything commented out except actual comments for uncommented lines from this file.

mikemcquaid:
> It does, but it doesn't set directory. I also changed bind to 127.0.0.1 because many homebrew daemons are configured this way by default.

Sorry, I missed that before. It'd be better to install the upstream file and add the directory line to it and `inreplace` the bind. That'll be more concise and future proof.

meehow:
@mikemcquaid Agree! I didn't know abut `inreplace`. Thanks for advice.
I checked RethinkDB documentation and behavior more carefully and it looks like manipulating `bind` isn't event needed because it listens on localhost and nothing else by default.

meehow:
I will create another pull request ;)

PR #46515, opened 2015-11-30
yank 0.5.0

dongcarl:
Caveat: running `brew audit yank --strict --online` gives this
```
==> brew style yank

==> audit problems
yank:
 * Description shouldn't include the formula name

Error: 1 problem in 1 formula
```
I'm guessing that the audit rule "Description shouldn't include the formula name" was added to prevent people from writing things like "\<formula\> does blah blah blah." However I think this formula warrants an exception doesn't it?

### DIFF
--- a/Library/Formula/yank.rb
+++ b/Library/Formula/yank.rb
@@ -1,8 +1,8 @@
 class Yank < Formula
   desc "Yank terminal output to clipboard"
   homepage "https://github.com/mptre/yank"
-  url "https://github.com/mptre/yank/archive/v0.4.1.tar.gz"
-  sha256 "cbd6b3c9d580fa619f3cba2c53b18d26674d587333976fe75b402f76ea01c4e5"
+  url "https://github.com/mptre/yank/archive/v0.5.0.tar.gz"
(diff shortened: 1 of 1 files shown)


DomT4:
> I'm guessing that the audit rule "Description shouldn't include the formula name" was added to prevent people from writing things like "<formula> does blah blah blah."

Nope. The audit rule was added because we already include the name by default on `desc` calls:

```
brew desc valgrind
valgrind: Dynamic analysis tools (memory, debug, profiling)
```

And `brew info` includes the formula name:

```
valgrind: stable 3.11.0 (bottled), HEAD
Dynamic analysis tools (memory, debug, profiling)
http://www.valgrind.org/
Not installed
From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/valgrind.rb
```

dongcarl:
@DomT4 I see! However I'm using the word "Yank" as a verb in the description so hopefully it'll be approved 😄

DomT4:
Heh, I'll let it go in this circumstance given that.

Merged in https://github.com/Homebrew/homebrew/commit/fdfe971cc77a8002a9f04fd071a2c41eb3b03f35. Thank you for another Homebrew contribution @dongcarl! :smiley_cat:

PR #46516, opened 2015-11-30
valgrind: Fix the `inreplace' for systems without CLT

dtrebbien:
This fixes:
<pre>
Making all in coregrind
make[2]: *** No rule to make target `&lt;', needed by `m_mach/mach_vmUser.c'.  Stop.
</pre>

See also PRs #45851 and #45801.

### DIFF
--- a/Library/Formula/valgrind.rb
+++ b/Library/Formula/valgrind.rb
@@ -42,7 +42,7 @@ class Valgrind < Formula
 
     # Look for headers in the SDK on Xcode-only systems: https://bugs.kde.org/show_bug.cgi?id=295084
     unless MacOS::CLT.installed?
-      inreplace "coregrind/Makefile.in", %r{(\s)(?=/usr/include/mach/)}, '<\1>'+MacOS.sdk_path.to_s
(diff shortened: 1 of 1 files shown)


dtrebbien:
CC @DomT4 @tamird

tamird:
Does this work correctly with Ruby 1.8.7? I was following those docs when I wrote this.

DomT4:
If it's passing Travis it *should* be okay. Our Travis 10.9 bot now [runs under](https://travis-ci.org/Homebrew/homebrew/jobs/93984812) Ruby 1.8.7 to check these things.

dtrebbien:
I am not sure because I have Ruby 2.0.0p645.  However, I have only modified the replacement string.  `'<\1>'` was replaced with `'\1'`.

tamird:
Yeah, this is currently busted and this change fixes it. LGTM

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/06f832095f51efc8309088014caaf937bcbeab06. Thank you for another Homebrew contribution @dtrebbien & Thank you @tamird for the quick review!

PR #46522, opened 2015-12-01
Update zanata-client to 3.8.1

Fryguy:
cc @lukebrooker

### DIFF
--- a/Library/Formula/zanata-client.rb
+++ b/Library/Formula/zanata-client.rb
@@ -1,8 +1,8 @@
 class ZanataClient < Formula
   desc "Zanata translation system command-line client"
   homepage "http://zanata.org/"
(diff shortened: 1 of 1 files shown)


lukebrooker:
@Fryguy Thanks again for this. I was going to update this when Zanata 3.8 was released, but it's fine to do it now.

There is one more change needed, which is to depend on Java 1.8+ instead of 1.7+.

Fryguy:
Updated with the 1.8+ change.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/193d9e783a20358a0a9f64c031bb48cb8298521d. Thank you for another Homebrew contribution @Fryguy! :smiley_cat:

PR #46534, opened 2015-12-01
yank: Fix issue with terminal output

elyscape:
This provides a workaround for mptre/yank#22.

### DIFF
--- a/Library/Formula/yank.rb
+++ b/Library/Formula/yank.rb
@@ -11,6 +11,8 @@ class Yank < Formula
     sha256 "523fa3f42ba361f22e1f65da4e04567501e2a54b38b00074ab79e9ce3870a03a" => :mavericks
   end
 
+  patch :DATA
+
   def install
     system "make", "install", "PREFIX=#{prefix}", "YANKCMD=pbcopy"
   end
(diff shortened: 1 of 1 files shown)


DomT4:
It looks like upstream are proposing a different solution here. What's the status on that?

elyscape:
Still in the evaluation phase. We could skip this PR if you'd like, though, as it looks like there's going to be a solution in the next release. Your call.

DomT4:
How soon is the next release, and how severe is the issue?

elyscape:
No idea how soon the next release will be but the issue does make it so that it doesn't display correctly. Every time you move, it reprints all the options again below the old ones, instead of printing them over the old ones.

DomT4:
Alright. I'm happy enough to take it whilst you talk to upstream about a more permanent solution, but can you edit the formula to include a link to the upstream bug submission and include a short one or two line summary as to what the patch fixes.

You can't assume any future contributors will come back to find this issue, sadly.

elyscape:
Sure. I'll get that in probably tomorrow.

elyscape:
How's this look, @DomT4?

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/1848c1c1d03995d137dc4be0a2b40ee0120c7d67. Thank you for another Homebrew contribution @elyscape! :smiley_cat:

elyscape:
You're very welcome!

PR #46537, opened 2015-12-01
centralize the logic of handling `homebrew-` in Tap.fetch

xu-cheng:


Library/Homebrew/tap.rb
```diff
@@ -17,6 +17,9 @@ def self.clear_cache
   end
 
   def self.fetch(user, repo)
+    # we special case homebrew so users don't have to shift in a terminal
+    user = "Homebrew" if user == "homebrew"
+    repo = repo[9..-1] if repo.start_with? "homebrew-"
```
bfontaine: Wouldn’t something like this be clearer? `repo.sub!(/^homebrew-/, "")`

xu-cheng: I found regex's performance is far worse than substring. And it's a hot-spot here. Some benchmark:

```ruby
require "benchmark/ips"

def regex(input)
  input.sub(/^homebrew-/, "")
end

def substr(input)
  input = input[9..-1] if input.start_with? "homebrew-"
  input
end

def test(method)
  %w[
    homebrew-foo
    bar
  ].each { |s| send(method, s) }
end

Benchmark.ips do |x|
  x.report("use regex") { test :regex }
  x.report("use substr") { test :substr}
  x.compare!
end
```

While yields following result:
```
Calculating -------------------------------------
           use regex    30.447k i/100ms
          use substr    59.048k i/100ms
-------------------------------------------------
           use regex    395.658k (± 5.6%) i/s -      1.979M
          use substr    971.412k (± 5.3%) i/s -      4.901M

Comparison:
          use substr:   971411.6 i/s
           use regex:   395658.0 i/s - 2.46x slower
```

bfontaine: Thanks!

UniqMartin: Could the following be a worthwhile compromise between performance and legibility?

```ruby
class String
  def strip_prefix(prefix)
    start_with?(prefix) ? self[prefix.length..-1] : self
  end
end
```

It's also more reusable. On my machine, this gives:

```
Calculating -------------------------------------
           use regex    18.635k i/100ms
          use substr    37.992k i/100ms
           use strip    33.656k i/100ms
-------------------------------------------------
           use regex    272.187k (± 3.8%) i/s -      1.360M
          use substr    705.938k (± 4.5%) i/s -      3.533M
           use strip    582.990k (± 5.1%) i/s -      2.928M

Comparison:
          use substr:   705938.1 i/s
           use strip:   582990.3 i/s - 1.21x slower
           use regex:   272186.9 i/s - 2.59x slower
```

xu-cheng: I like the idea which can also be used in other place of codebase.

xu-cheng:
Update to use `String#strip_prefix`

UniqMartin:
I like this cleanup a lot. No more `sub`/`gsub` with a mix of string and (anchored) regexp arguments.

Thus big :+1: from me.

bfontaine:
Yep I’m :+1: here for the same reason.

DomT4:
It's good with me.

mikemcquaid:
Great work :+1:

PR #46538, opened 2015-12-01
.gitignore: ignore IDEA RubyMine project files

apjanke:
Folks,

What do you think about adding `.idea` to the `.gitignore` like this, so the RubyMine IDE can be more easily used with the local Homebrew installation? I'm experimenting with using it instead of a plain text editor.

You can still use RubyMine without this in the ignore by creating an empty project at a separate location and then adding `/usr/local/Library` as a separate content root in the project definition. But that's a little unwieldy imho. And adding `.idea` seems benign: we'd never want one checked in to the repo anyway, since it's IDE-specific and contains local per-user settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@
 /Library/PinnedKegs
 /Library/Taps
 /Library/PinnedTaps
+# RubyMine project files:
+.idea


bfontaine:
You might want to configure a global `.gitignore` file. Put the following in your `~/.gitconfig`:

```gitconfig
[core]
	excludesfile = ~/.cvsignore
```

Then put `.idea` in `~/.cvsignore`. It’ll be ignored in *all* your git projects.

xu-cheng:
Agree that you probably need to use global gitignore file. Also this shouldn't be necessary considering our gitignore file is [a whitelist that ignores everything in the first line](https://github.com/Homebrew/homebrew/blob/master/.gitignore#L1).

apjanke:
> You might want to configure a global .gitignore file. 

I'd rather not use a global gitignore file for this, because for some smaller projects, I do want to check in the IDE project definition files. Either because I'm collaborating with a smaller team that's specifically adopted it, or the IDE project is effectively part of the build system, or I just want to sync multiple copies of the project for myself. Not a big deal for RubyMine specifically, but the `.idea` project directory name is also the same one used by JetBrains IntelliJ IDEA, a Java IDE, which I would want to track (at least partially) in some cases.

> Also this shouldn't be necessary considering our gitignore file is a whitelist that ignores everything in the first line.

The `/*` whitelist only ignores files at the top `/lusr/local` level, I think. When I tried this out, I also put the `.idea` at `/usr/local/Library`, because I wanted the project to pick up just the Homebrew code, and not all the files from installed formulae. It showed up as an untracked file there.

apjanke:
Oh - that could also be that `/Library/` [is itself whitelisted](https://github.com/Homebrew/homebrew/blob/2fd089afa3eef07c18c6326af2e25bc8a303e5c0/.gitignore#L5), and that's where I'd want to drop the project definition under.

mikemcquaid:
I'm fairly :-1: on this just because it's got the potential to increase to everyone's favourite editor but don't feel super strongly.

apjanke:
Closing this out as WONTFIX. Turns out it's no big deal in practice to create a RubyMine project with multiple "Content Roots" so the `.idea` project definition directory lives outside `/usr/local` (or whatever your Homebrew location is), and just has a reference to it. You just need to be careful that any new files you create go in `/usr/local` if you want them to get checked in. This avoids polluting `/usr/local`, and does not require changes to the user's global `.gitignore` configuration.

PR #46539, opened 2015-12-01
Add formula for fifechan 0.1.2

stubb:
Adds a formula for the fifechan library.


Library/Formula/fifechan.rb
```diff
@@ -0,0 +1,31 @@
+class Fifechan < Formula
+  desc "Is a C++ GUI library designed for games."
```
bfontaine: You can remove “`Is a`”.

dunn:
If this is _only_ for games, I wonder if we should have it in our Games tap instead.

mikemcquaid:
> If this is only for games, I wonder if we should have it in our Games tap instead.

Agreed :+1:.

stubb:
What do you mean by "Games tab"?

stubb:
Nervermind. Figured it out https://github.com/Homebrew/homebrew-games/pull/457

PR #46547, opened 2015-12-01
OpenCoarrays 1.3.2 (new formula)

zbeekman:
OpenCoarrays is an open source runtime library and toolchain for implementing the features of Coarray Fortran (CAF). It is designed to be the main CAF backend for GCC's GFortran, and to be an open source implementation that may be adopted by other compiler vendors, and provides alternate implementations for end users.

The main supported backend is currently MPI, although a more experimental GASNet backend exists as well. MPICH or MVAPICH is preferred to OpenMPI.

### DIFF
--- a/Library/Formula/opencoarrays.rb
+++ b/Library/Formula/opencoarrays.rb
@@ -1,0 +1,67 @@
+class Opencoarrays < Formula
+  desc "open-source coarray Fortran ABI, API, and compiler wrapper"
+  homepage "http://opencoarrays.org"
+  url "https://github.com/sourceryinstitute/opencoarrays/releases/download/1.3.2/OpenCoarrays-1.3.2.tar.gz"
+  sha256 "391937a5d24962bd93ecec2402545a3e9d6c7a43987082c7de7dee3b74151ec3"
+
+  head "https://github.com/sourceryinstitute/opencoarrays.git"
+
+  option "without-test", "Skip build time tests (requires properly configured MPICH and firewall settings)"
+
+  depends_on :fortran
+  depends_on :mpi => [:cc, :f90]
+  depends_on "cmake" => :build
+
+  # GCC >= 5.2 required; no other Homebrew supported compilers are compatible
+  # but the mpicc compiler may be clang (which is ABI compatible with GCC)
+  fails_with :llvm
+  fails_with :gcc
+  fails_with :gcc_4_0
+  ("4.3".."4.9").each do |series|
+    fails_with :gcc => series
+  end
+  fails_with :gcc => "5.1"
+
+  def install
+    mkdir "build" do
+      ENV["FC"] = ENV["MPIFC"]
+      ENV["CC"] = ENV["MPICC"]
+      cmake_args = std_cmake_args
+      system "cmake", "..", *cmake_args
+      system "make"
+      # A bug in GFortran <= 5.2 causes the "co_reduce" test to fail; find
+      # which version of gfortran is being used, and skip the test if needed.
+      gfortran_version = /\d\.\d+/.match(`mpif90 --version`)
+      test_excludes = []
+      test_excludes = ["-E", "co_reduce"] if gfortran_version.to_s < "5.3"
+      system "ctest", *test_excludes if build.with? "test"
+      system "make", "install"
+    end
+  end
+
+  test do
+    ENV.fortran
+    (testpath/"tally.f90").write <<-EOS.undent
+      program main
+        use iso_c_binding, only : c_int
+        use iso_fortran_env, only : error_unit
+        implicit none
+        integer(c_int) :: tally
+        tally = this_image() ! this image's contribution
+        call co_sum(tally)
+        verify: block
+          integer(c_int) :: image
+          if (tally/=sum([(image,image=1,num_images())])) then
+             write(error_unit,'(a,i5)') "Incorrect tally on image ",this_image()
+             error stop 2
+          end if
+        end block verify
+        ! Wait for all images to pass the test
+        sync all
+        if (this_image()==1) write(*,*) "Test passed"
+      end program
+    EOS
+    system "#{bin}/caf", "tally.f90", "-o", "tally"
+    system "#{bin}/cafrun", "-np", "3", "./tally"
+  end
+end


zbeekman:
trying to enforce use of gcc with `fails_with` blocks... seeing if this helps fix the issue...

zbeekman:
added more rigorous check

zbeekman:
Alright, just waiting on some responses on #46551 and from the folks at OpenCoarrays/SourceryInstitute about whether or not test 6 failing is a known issue or not. Test 18, "co_reduce" is a [known issue with GCC 5.2](https://github.com/sourceryinstitute/opencoarrays/blob/master/STATUS.md#compiler-issues-gnu)

zbeekman:
If tests pass, please _*DON'T*_ merge without answering #46551

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,73 @@
+class Opencoarrays < Formula
+  desc "Coarray Fortran runtime library and toolchain, free & open source"
+  homepage "http://opencoarrays.org"
+  url "https://github.com/sourceryinstitute/opencoarrays/releases/download/1.2.0/opencoarrays-1.2.0.tar.gz"
+  sha256 "7d0c3f2dd842b32b2a24ac5fb923bc026d541d9e037205162b39a3d381d9fa83"
+
+  head "https://github.com/sourceryinstitute/opencoarrays.git"
+
+  option "without-test", "Run build time tests (requires properly configured MPICH and firewall settings)"
```
UniqMartin: You might want to make the option description match the `without`, e.g. `Skip build time tests (…)`.

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  depends_on :fortran
+  depends_on "gcc"
```
UniqMartin: There is no need to explicitly depend on `gcc`. As far as I can tell, the build system only uses the MPI wrappers anyway. Also, `depends_on :fortran` in practice already implies a `gcc` dependency.

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,73 @@
...
+  depends_on :mpi => [:cc, :f90, "with-fortran"]
```
UniqMartin: `"with-fortran"` should be removed.

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  fails_with :clang do
+    cause "Only compatible with GCC >= 5.1"
+  end
+
+  fails_with :llvm do
+    cause "Only compatible with GCC >= 5.1"
+  end
+
+  fails_with :gcc => 5.0 do
```
UniqMartin: This syntax requires a string on the right-hand side of `=>`, not a float. Also, there was never a GCC 5.0 release. You might want to consolidate this and the preceding statements based on the suggestions made in #46551.

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,73 @@
...
+    cause "GCC 5.1 or greater is required"
+  end
+
+  def install
+    mkdir "build" do
+      ENV["FC"] = HOMEBREW_PREFIX + "bin/mpif90"
+      ENV["CC"] = HOMEBREW_PREFIX + "bin/mpicc"
```
UniqMartin: There should be no need to specify the full path. Using `mpif90` and `mpicc` should suffice. Ideally, these won't be needed at all and the build system picks up the environment variables `MPICC` and `MPIFC` that are automatically set by `depends_on :mpi => [:cc, :f90]`.

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,73 @@
...
+      system "cmake", "..", *std_cmake_args
+      system "make"
+      system "ctest", "-E", "co_reduce|allocate_as_barrier_proc" if build.with? "test"
```
UniqMartin: It would be good to have a comment with an explanation why these tests are excluded as well as references to the upstream issues (if any).

zbeekman: @UniqMartin good point. This raises a question:

Is it possible to determine the version of GCC being used? Is it possible to specify a preference for GCC >= 5.3? The `co_reduce` test failure is a known issue with GCC/GFortran 5.2 and a patch has been submitted and will be in the 5.3 and 6.x development branch.

The other test is the subject of my inquiry and I'll report back when I find the answer to my question: sourceryinstitute/opencoarrays#28

Library/Formula/opencoarrays.rb
zbeekman: @UniqMartin I can confirm that [the build system requires these variables to be set](https://travis-ci.org/Homebrew/homebrew/jobs/94325955#L578).

I am unsure of whether this is a deficiency with the opencoarrays build system, a CMake bug/deficiency, or a Homebrew idiosyncrasy, but it appears to be necessary to tell CMake to use `mpicc` and `mpif90` as the C and Fortran compilers.

zbeekman:
@UniqMartin it seems that one of your suggested changes has [broken the build](https://travis-ci.org/Homebrew/homebrew/jobs/94416190#L558). CMake is complaining that mpif90 is broken. I have some ideas to try to fix this, by adding back in some stuff.... I'll give it a whirl.

zbeekman:
The two failing Travis-CI tests timed out. I was watching the logs as they ran, and it seemed like they just stalled and stopped updating at some point. I wonder if Travis is having issues, or if resubmitting the jobs will get them to pass.

As far as I'm concerned this Formula seems to be working to my liking, and pending further comments from @homebrew team members, is ready to be merged.

zbeekman:
attempting to fix problems and speed build using full path to brewed `mpif90` and `mpicc`

zbeekman:
Seems like my struggles are caused by #46422 and #46461

zbeekman:
Alright, as far as I can tell, this formula is good to go, but blocked by 
#46422 and #46461

zbeekman:
I'm breaking out further experimentation into a new PR to allow me to test with the test bot. Above comment still applies.

UniqMartin:
@zbeekman Thank you for researching this further! Having a quick look, I guess we have to address the outstanding MPI-related issues here at Homebrew before we can move forward with your PR.

Please bear with me, as I'm a bit short on time right now. I'll definitely have a look and see what I can do and will respond to your questions as soon as possible.

zbeekman:
@UniqMartin I understand completely. All I ask of you, is that at some point you look over my latest iteration of the Formula, for any additional outstanding issues. No rush.

I think I have resolved all the other questions which I had.

UniqMartin:
> I understand completely. All I ask of you, is that at some point you look over my latest iteration of the Formula, for any additional outstanding issues. No rush.

Thanks for understanding! I will for sure. And I don't want to rush this, as I'm a relatively new maintainer and your formula submission is not exactly trivial.

zbeekman:
fair... it uses some very new GCC features, and anytime you have MPI and Fortran it complicates things.

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,85 @@
...
+  fails_with :gcc_4_0
+  ("4.3".."4.9").each do |series|
+    fails_with :gcc => series
+  end
+  fails_with :gcc => "5.1"
+
+  def install
+    mkdir "build" do
+      ENV["FC"] = "mpif90"
+      ENV["CC"] = "mpicc"
```
mikemcquaid: Can you rely on both these variables being these exact values or are there other ways it could have `:mpi` satisfied?

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,85 @@
...
+      # Release build type in std_cmake_args can break the build due to GCC bugs
+      # at certain optimization levels. See sourceryinstitute/opencoarrays#28
```
mikemcquaid: Can you make this a proper URL? Also, this should be patched upstream.

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,85 @@
...
+      # for further details
+      cmake_args = std_cmake_args
+      cmake_args.delete("-DCMAKE_BUILD_TYPE=Release")
+      system "cmake", "..", *cmake_args
+      system "make"
+      # see sourceryinstitute/opencoarrays/blob/master/STATUS.md#compiler-issues-gnu
```
mikemcquaid: Please explain so people don't have to read this, thanks.

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,85 @@
...
+      gfortran_version = /\d\.\d+/.match(`mpif90 --version`)
+      test_excludes = ["-E", "co_reduce"] if gfortran_version.to_s < "5.3"
+      system "ctest", *test_excludes if build.with? "test"
```
mikemcquaid: Will this `test_excludes` not be `nil` in some cases?

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+  end
+
+  def caveats; <<-EOS.undent
+    If the firewall is enabled, it is normal to receive pop-up messages asking
+    permission to accept incomming connections due to MPI during the build, unless
+    `--without-test` is specified. Additionally `brew test opencoarrays` will also
+    cause these firewall notifications. They may be safely ignored, or they may be
+    prevented by momentarily disabling the Mac OS X firewall under security
+    settings.
```
mikemcquaid: Remove this caveat; it applies to most OS X software in Homebrew.

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    GCC >= 5.2 is required to build and use OpenCoarrays. Additionally MPICH is
+    prefered to OpenMPI.
```
mikemcquaid: Don't need this caveat, it's specified above.

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    For a list of known issues please see:
+    https://github.com/sourceryinstitute/opencoarrays/blob/master/STATUS.md#known-issues
```
mikemcquaid: Don't need this caveat; it's not Homebrew specific.

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,85 @@
...
+             error stop 2
+          end if
+        end block verify
+        ! Wait for all images to pass the test
+        sync all
+        if (this_image()==1) write(*,*) "Test passed"
+      end program
+    EOS
+    system "#{bin}/caf", "tally.f90", "-o", "tally"
+    system "#{bin}/cafrun", "-np", "3", "./tally" # non-zero exit indicates failure
```
mikemcquaid: Don't need `# non-zero exit indicates failure`; that applies to almost all Unix software.

Library/Formula/opencoarrays.rb
zbeekman: I'm not sure I completely understand your question, @mikemcquaid. AFAICT, both OpenMPI and MPICH installed via homebrew provide these.

Is there a more robust/generic way to get the f90 and cc MPI wrapper scripts?

Library/Formula/opencoarrays.rb
zbeekman: Sure, I can make this a proper URL.

When you say patched upstream, which upstream do you mean? It seems that there are workarounds baked into OCA's CMakeLists.txt to work around GCC bugs. It is my understanding that without these GCC bugs, this wouldn't be an issue. So a forthcoming GCC might resolve this issue.

I can try to submit a PR upstream to OpenCoarrays... the OCA guys have been a bit swamped lately, so it may take a while to resolve. Is there a chance that this formula PR could be accepted with the `std_cmake_args.delete('-DCMAKE_BUILD_TYPE=Release')` hack or is this a show stopper in your opinion?

Library/Formula/opencoarrays.rb
zbeekman: sure thing.

Library/Formula/opencoarrays.rb
zbeekman: yes, when GFortran is version 5.3 or greater. (GFortran 5.3 was released today)

Library/Formula/opencoarrays.rb
zbeekman: fair... it was mostly a reminder to myself, since I was having a lot of trouble with `assert_match / Test passed./ shell.output("#{bin}/cafrun -np 3 ./tall")` on one of the OS X images (10.9? I can't remember, but this has  been an outstanding issue for some time.) I'll change it.

Library/Formula/opencoarrays.rb
zbeekman: I'll kill the entire caveats block then...

Library/Formula/opencoarrays.rb
mikemcquaid: I think we can include it but it'd be nicer if there was another way as it may have adverse other effects.

mikemcquaid: But if the upstream issue spells this out: :+1:

Library/Formula/opencoarrays.rb
mikemcquaid: Doing `*` on `nil` will fail so needs to be set to an empty array by default and then make the `if` add to it.

Library/Formula/opencoarrays.rb
mikemcquaid: Great, thanks.

Library/Formula/opencoarrays.rb
mikemcquaid: Thanks!

Library/Formula/opencoarrays.rb
zbeekman: OK, looking at what you, @mikemcquaid said, and what @UniqMartin said above, what if I changed these lines to:

```ruby
   ENV["CC"] = ENV["MPICC"]
   ENV["FC"] = ENV["MPIFC"]
```

Library/Formula/opencoarrays.rb
zbeekman: ah, got it... interesting, in my testing (manually changing/setting values) no problems arose due to that issue. TBH, I'm pretty new to Ruby (mostly homebrew formulae) and don't really understand what `*` is for... At any rate I can change it.

Library/Formula/opencoarrays.rb
zbeekman: The intended install methodology from the upstream is not to specify a build type, although I filed an issue with them and they seemed interested in getting the compilation and flags to be set correctly with all build types.... I'll try to work up a PR for them.

zbeekman:
I'm not going to push the latest iteration until we conclude the discussion on the diffs, especially this one: https://github.com/Homebrew/homebrew/pull/46547#discussion-diff-46712441

Library/Formula/opencoarrays.rb
UniqMartin: Yes, these are the variables implicitly set by `depends_on :mpi => [:cc, :f90]`, so this is slightly better than hard-coding `mpicc` and `mpif90`. Ideally though, the OpenCoarrays build system would be taught to pick up `MPICC` and `MPIFC` and use them directly (if possible). Is upstream aware of the issue?

zbeekman: I think there is a FindMPI CMake module that should be able to deal with reconstructing the compiler and linker flags... maybe if I find time, I'll try to submit a patch upstream to remove the need for these lines... right now without them the build fails.

Library/Formula/opencoarrays.rb
UniqMartin: `*` is the splat operator in this instance. It will convert the items of an array into individual arguments to be passed to the `system` method. It has other uses, too: [quick overview](https://endofline.wordpress.com/2011/01/21/the-strange-ruby-splat/).

Library/Formula/opencoarrays.rb
zbeekman: @UniqMartin: As far as I know, CMake will always pick the compiler first thing at configure time, and only from the FC and CC variables... So I don't know if this is really an upstream issue, or a CMake issue, but, as I alluded to above, it also does have other MPI aware capabilities... I'll look into this further, and make upstream aware, but I don't have enough experience building MPI packages with CMake to know what is "the right way to do it"™

Library/Formula/opencoarrays.rb
zbeekman: Thanks for the explanation @UniqMartin! So should I `test_excludes = Array.new` or `test_excludes = []` or `test_excludes = ""` or does it not matter?

Library/Formula/opencoarrays.rb
UniqMartin: > […], but I don't have enough experience building MPI packages with CMake to know what is "the right way to do it"™

Neither do I, so I can't tell you what the right way is in this constellation, but only what would be desirable given how other build systems deal with MPI. We're not asking for an immediate and all-encompassing solution, but if upstream is aware of the issue and open to your future contribution, this will benefit everyone who tries to package this piece of software; and it will lower everyone's maintenance burden and need for ugly workarounds.

Library/Formula/opencoarrays.rb
UniqMartin: `Array.new` is unnecessarily verbose and `""` is a string, not an array. The line starting with `test_excludes = ` should be replaced by:

```ruby
test_excludes = []
test_excludes << "-E" << "co_reduce" if gfortran_version.to_s < "5.3"
```

Library/Formula/opencoarrays.rb
mikemcquaid: > ENV["CC"] = ENV["MPICC"]
> ENV["FC"] = ENV["MPIFC"]

That's be better, yep.

Library/Formula/opencoarrays.rb
UniqMartin: > ah, got it... interesting, in my testing (manually changing/setting values) no problems arose due to that issue.

Just to clarify: That's because newer Ruby versions (OS X 10.9+) do the same thing for both `*[]` and `*nil`. But we try not to break things for Ruby 1.8 users and there the behavior is different.

zbeekman:
OK, I'll push my latest iteration, and poke around upstream regarding the `-DCMAKE_BUILD_TYPE=Release` issue and the compiler specification issue, unless anyone things that there are more points in our discussion that need to be addressed. I'll hold off pushing for a few hours...

UniqMartin:
Thanks for all the work you put into this and for working with upstream to try to resolve the outstanding issues! I'm sure this will benefit everyone in the long run.

> I'll hold off pushing for a few hours...

Would you mind keeping an eye on #46461? That issue isn't resolved yet, unfortunately, so pushing before this happens will only create additional pressure on our CI infrastructure without much benefit. (Please correct me, if I'm overlooking something.) And sneaking the `"with-fortran"` back in will make it work, but will also force the CI to additionally build `open-mpi` and this is not the solution we are looking for. I know that waiting is frustrating, but there's only so many hours in a day and we're not exactly overstaffed …

zbeekman:
fair, I'll hold off pending resolution of #46461 (I also think that #46553 should probably be resolved first too...)

Library/Formula/opencoarrays.rb
zbeekman: Updates:
 - As discussed, this is a GFortran bug work-around/issue. The status can be followed here: https://groups.google.com/forum/m/#!topic/opencoarrays/IOx0RFLJEr0 which is a conversation between GFortran/GCC contributors/maintainers
 - I have submitted a PR Sourceryinstitute/opencoarrays#43 to put some logic in the build system to detect if the code can be compiled safely with optimization flags.

zbeekman:
OK, it sounds like the OpenMPI/Homebrew-Fortran bug has been resolved.... I'm working with OpenCoarrays now to address the GFortran bug work around (that necessitated removing the `-DCMAKE_BUILD_TYPE=Release` flag from `std_cmake_args` and to avoid passing `mpif90` and `mpicc` directly as FC and CC to Cmake.

So please hold off on merging this. Once the upstream changes have been made, I'll modify the Formula here, which should trigger a rebuild. Then you can suggest any further changes that need to be made or merge the Formula.

UniqMartin:
No worries, we're not going to merge this as long as we don't have a working build on our CI.

Thanks for working with upstream on the remaining issues! Once you feel things are ready, simply push your changes and let us know.

zbeekman:
yup will do. And @mikemcquaid I cannot express my gratitude enough that you took the bull by the horns and got the MPI issue resolved!

mikemcquaid:
@zbeekman My pleasure, thanks for the kind words.

zbeekman:
I just wanted to drop by and say this PR isn't dead. I've been busy, and am working on some upstream stuff (getting rid of the requirement to pass FC=mpif90 and also some build/install layout stuff)

I think we hope to get a minor release out on Friday or shortly thereafter, if we can swing it, and hopefully I can include all the stuff I want to fix/improve before finalizing this PR. Then, I can work on this, and use the latest release.

zbeekman:
I have made some minor updates to the formula and have been working on cleaning some stuff up with the upstream.

Upstream we are still working on:

 1. We are reexamining some of the installation logic, and where things get put to make sure that this is in line with the usual GNU/Homebrew install directory convention
 2. We are looking into using CMake's `FindMPI` module rather than passing the compilers as `mpif90` and `mpicc` and incorporating the added benefits of this module into our MPI compiler wrapper script and MPI runtime wrapper script.

If these tests pass, we could merge this into Homebrew, with the caveat that the install directory structure may need to be have some tweaks done to it. I'm happy to wait too.

zbeekman:
CC/FYI: @sck-at-ucy There are some tweaks that I'm working on in the build system and hopefully I should finish them soonish, then the folks at Homebrew will give a final :+1: or :-1: on if they will merge.

sck-at-ucy:
@zbeekman thanks for all the kind help getting started

Library/Formula/opencoarrays.rb
```diff
@@ -0,0 +1,67 @@
...
+  fails_with :gcc => "5.1"
+
+  def install
+    mkdir "build" do
+      ENV["FC"] = ENV["MPIFC"]
+      ENV["CC"] = ENV["MPICC"]
+      cmake_args = std_cmake_args
+      system "cmake", "..", *cmake_args
+      system "make"
+      # A bug in GFortran <= 5.2 causes the "co_reduce" test to fail; find
```
mikemcquaid: I'd rather see a requirement here that requires the use of gfortran > 5.2.

mikemcquaid:
One more comment: otherwise LGTM :+1:

Library/Formula/opencoarrays.rb
zbeekman: > I'd rather see a requirement here that requires the use of gfortran > 5.2.

I would too.... but how? Or did you mean that I should update the comment to say that it requires the use of gfortran > 5.2?

mikemcquaid: Add a new `Requirement` class that basically has this logic or add it to the current `MPIRequirement` class (see `JavaRequirement` for an example there).

zbeekman: @mikemcquaid OK, I'll ack for `JavaRequirement` in the Formula subdir... Just FYI, my ruby is limited to my experience contributing to Homebrew, so I may ping you (or maybe someone else) if I get stuck.... Any links you can provide to relavent source code would be *highly* appreciated. :bow:

mikemcquaid: Sure thing, feel free to ping!

zbeekman:
If I still want to finish this and get it merged do I need to port it over to the new repo? (And rebase, obvi...)

UniqMartin:
It's been a while since the PR has been updated, so the bottles that were built back then are gone by now. The easiest would be to open a new PR in [`homebrew/core`](https://github.com/Homebrew/homebrew-core) and to link from there to this PR for reference. (And if you could summarize the status of this PR in the first comment, that would be great.) Luckily, there's not much to port, as this is a new formula, so this should be just a matter of copying the formula file and creating a new commit for it.

zbeekman:
@UniqMartin sounds good. There are a few more details to attend to... Also, the project is changing a lot especially the build/install system and dependencies. So far all backwards compatible, but with my limited time I'm waiting to address a few remaining issues until things settle down a little bit. GCC 5.4 added support for coarray fortran "events" and GCC 6.1 isn't too far off, so I may also wait to see if OpenCoarrays decides to require the latest bleeding edge GCC as of 5.4 or 6.1.

In addition, OpenCoarrays 2.x is theoretical not far away, which will have source to source translation capabilities allowing a user to write CAF code and translate it to run on any recent-ish fortran compiler. This may also help ease the tendendency to make the bleeding edge GCC a hard prereq.

MikeMcQuaid:
Passing on this as-is, please reopen on https://github.com/homebrew/homebrew-core if desired.

PR #46548, opened 2015-12-01
platypus 5.0

chdiza:


chdiza:
I don't understand why the bots failed.  Their logs claim that the directory where some built resources should be doesn't exist.  It works locally :(

DomT4:
Looks like a compiler problem?
```
/tmp/platypus20151201-57494-1313grb/Shared/ScriptAnalyser.m:132:19: error: expected '>'
+ (NSArray<NSArray*> *)interpreterSuffixes {
                  ^
/tmp/platypus20151201-57494-1313grb/Shared/ScriptAnalyser.m:132:20: error: expected ')'
+ (NSArray<NSArray*> *)interpreterSuffixes {
                   ^
/tmp/platypus20151201-57494-1313grb/Shared/ScriptAnalyser.m:132:3: note: to match this '('
+ (NSArray<NSArray*> *)interpreterSuffixes {
  ^
/tmp/platypus20151201-57494-1313grb/Shared/ScriptAnalyser.m:101:15: warning: comparison of constant 'NSNotFound' (9223372036854775807) with expression of type 'int' is always false [-Wtautological-constant-out-of-range-compare]
    if (index == NSNotFound) {
        ~~~~~ ^  ~~~~~~~~~~
```

Maybe no longer supports Mavericks `clang`?

chdiza:
It's supposed to work on Lion and up.

DomT4:
It looks like it's built with the latest Xcode though:
> Built with XCode 7 and now requires Mac OS X 10.7 or later and a 64-bit Intel system

So some breakage might have slipped through on older build tools?

chdiza:
> So some breakage might have slipped through on older build tools?

Maybe.  But that doesn't explain the El_Cap and Yosemite bot failures.

DomT4:
On Travis? No idea. Looks like a Travis quirk, to be honest, given it didn't reproduce on Jenkins despite the directories Travis complains about not existing on the Jenkins VMs and presumably would still generate a sandbox violation.

mikemcquaid:
Let's retry Jenkins.

mikemcquaid:
@BrewTestBot test this please

DomT4:
Still no test failure on Jenkins, just the Mavericks build issue.

chdiza:
I don't have a Mavericks machine with which to debug.

mikemcquaid:
@chdiza You can use the Jenkins output and trial-and-error.

chdiza:
> You can use the Jenkins output

Where is the Jenkins output?  If I click on "Details" under "All checks have failed", on the line that doesn't mention  "Travis", I can't see any output.  I see some sort of authorization prompt.  Is this a bug? Why would I need grant some entity access to my GitHub account just to view a build log?

mikemcquaid:
@chdiza It's just checking what organisations you are in so it knows whether to make you an administrator.

mikemcquaid:
It does not give access to your GitHub account beyond that.

chdiza:
Upstream [says](https://github.com/sveinbjornt/Platypus/issues/56#issuecomment-163799382) that although compiled binaries will work down all the way to 10.7, *building* `platypus` requires Xcode7 or higher.

So if Mavericks is the only system on which the CI is failing, I'll add a guard to prevent `brew` from even attempting to build unless we're on Yosemite or up.

Is there a parallel sort of guard for specific versions of Xcode?  Or should I just do it for OS?

DomT4:
You can do `depends_on :xcode => ["7.0", :build]` or similar.

chdiza:
OK, I inserted an Xcode version dep.  Should I also add a `:macos` version dep?  Like `depends_on :macos => [:yosemite, :build]`?  If we don't, someone with Yosemite but Xcode 6.X will get a build error.  OTOH, their doctor will bug them to install Xcode 7.

chdiza:
Also, are we going to end up with a Mavericks bottle for this?  Upstream says the latest version will work on Mavericks, we just can't build it on Mavericks.

DomT4:
> If we don't, someone with Yosemite but Xcode 6.X will get a build error. 

Well, if upstream require Xcode 7 to be certain of build success, presumably we want it to fail in that case?

> Also, are we going to end up with a Mavericks bottle for this?

No, because the CI ship with the most recent Xcode release for that specific version. Mavericks will be on 6.2 or 6.4 or similar.

chdiza:
> Well, if upstream require Xcode 7 to be certain of build success, presumably we want it to fail in that case?

Yes, I confused myself.  I was thinking of the case where someone had Xcode 7 but not Yosemite or greater, which is not a possible case.  (Or if it is possible, they have bigger problems than a failed `platypus` :). )

chdiza:
I have no idea why the bots are failing now.

DomT4:
Yeah, keeps failing. Weirdly it wasn't failing when we didn't feed it a specific Xcode dependency :confused:.

chdiza:
I don't have time to try it right now, but later I'll switch the Xcode dep to 7.1 and see if that makes a difference.

DomT4:
Doing a rebuild. Might have been my screw up in forgetting to open the new Xcode apps after doing the CI update the other night.

DomT4:
Yeah, the latest error was my screw up. Forgot sometimes after doing Xcode updates you have to actually open the Xcode app to placate OS X.

Merged in https://github.com/Homebrew/homebrew/commit/5e383f0e1dc9696abc74211ddf4cf845be53e0b2. Thanks again @chdiza!

chdiza:
Heh, after all that, I just heard from upstream and he just made Mavericks buildable in the master branch (though not in the release).

DomT4:
Nice! Guess you could always modify the head line into a block and specify that only needs Xcode 6+, if you wanted to.

Issue #46542, opened 2015-12-01
Uninstalling a package with a linked app doesn't remove the link


grahamc:
Expected behavior: `brew uninstall emacs` would clean up and remove the app link.

Actual behavior:

```
➜  ~  brew uninstall emacs
Uninstalling /usr/local/Cellar/emacs/24.5... (3923 files, 118M)
```

however `/Applications/emacs.app` still existed.

UniqMartin:
One reason this doesn't happen automatically is because Homebrew tries not to touch stuff outside of its prefix (`/usr/local` by default), unless explicitly asked by the user. Hence, `brew linkapps` is also a separate manual step.

That said, I agree it would be desirable to check if uninstalling a formula breaks any of the symbolic links in `/Applications` and/or `~/Applications` and to notify the user about this (with a suggestion on how to resolve this). Would the following (not currently implemented) work for you?

```
$ brew uninstall emacs
Uninstalling /usr/local/Cellar/emacs/24.5... (3923 files, 118M)

The following linked applications are no longer installed:
  /Applications/emacs.app
Run `brew unlinkapps --uninstalled` to remove these symlinks.
```

grahamc:
Yes, I think that notice would be helpful.

DomT4:
In this case I'd be happy for it to happen automatically on `brew rm`, to be honest. I don't think it's outside our mandate to prune symlinks that we know belong inside our Cellar and no longer exist.

UniqMartin:
Yes, I guess it would be safe to do this automatically. I'll look into implementing this behavior.

DomT4:
You could probably throw out to `brew unlinkapps emacs` (for example) unconditionally in `brew uninstall` as a really simple solution. `brew unlinkapps` doesn't fail hard if an `.app` doesn't actually exist for that package.

Feel free to get elegant as desired though :smile:.

MikeMcQuaid:
We're closing out all issues relating to features that no-one is implementing in the short term in favour of pull requests.

PR #46549, opened 2015-12-01
(un)linkapps: modernize, prune: remove broken app symlinks

UniqMartin:
The first two commits are all about using `Pathname` as much as possible (bringing the code up to current Homebrew standards), reducing duplication, and avoiding external commands in favor of native methods.

The third (and last) commit implements `brew unlinkapps --prune` and allows to remove broken symlinks from `/Applications` and `~/Applications` if they were previously created with `brew linkapps`. This part also provides `Homebrew.unlinkapps_prune` that could be used from `cmd/uninstall.rb` et al. to implement automatic pruning of broken symlinks as suggested in #46542 (will be addressed in a separate PR).

### DIFF
--- a/Library/Homebrew/cmd/linkapps.rb
+++ b/Library/Homebrew/cmd/linkapps.rb
@@ -1,12 +1,11 @@
-# Links any Applications (.app) found in installed prefixes to /Applications
 require "keg"
 require "formula"
 
 module Homebrew
   def linkapps
-    target_dir = ARGV.include?("--local") ? File.expand_path("~/Applications") : "/Applications"
+    target_dir = linkapps_target(:local => ARGV.include?("--local"))
 
-    unless File.exist? target_dir
+    unless target_dir.directory?
       opoo "#{target_dir} does not exist, stopping."
       puts "Run `mkdir #{target_dir}` first."
       exit 1
@@ -16,24 +15,40 @@ module Homebrew
       kegs = Formula.racks.map do |rack|
         keg = rack.subdirs.map { |d| Keg.new(d) }
         next if keg.empty?
-        keg.detect(&:linked?) || keg.max { |a, b| a.version <=> b.version }
+        keg.detect(&:linked?) || keg.max_by(&:version)
       end
     else
       kegs = ARGV.kegs
     end
 
+    link_count = 0
     kegs.each do |keg|
       keg.apps.each do |app|
-        puts "Linking #{app} to #{target_dir}."
-        target = "#{target_dir}/#{app.basename}"
+        puts "Linking: #{app}"
+        target_app = target_dir/app.basename
 
-        if File.exist?(target) && !File.symlink?(target)
-          onoe "#{target} already exists, skipping."
+        if target_app.exist? && !target_app.symlink?
+          onoe "#{target_app} already exists, skipping."
           next
         end
 
-        system "ln", "-sf", app, target_dir
+        # We cannot use `install_symlink` because we want an absolute link.
+        FileUtils.ln_sf(app, target_dir)
+        link_count += 1
       end
     end
+
+    if link_count.zero?
+      puts "No apps linked to #{target_dir}" if ARGV.verbose?
+    else
+      puts "Linked #{link_count} app#{plural(link_count)} to #{target_dir}"
+    end
+  end
+
+  private
+
+  def linkapps_target(opts = {})
+    local = opts.fetch(:local, false)
+    Pathname.new(local ? "~/Applications" : "/Applications").expand_path
   end
 end
--- a/Library/Homebrew/cmd/prune.rb
+++ b/Library/Homebrew/cmd/prune.rb
@@ -1,5 +1,6 @@
 require "keg"
 require "cmd/tap"
+require "cmd/unlinkapps"
 
 module Homebrew
   def prune
@@ -47,5 +48,7 @@ module Homebrew
       print "and #{d} directories " if d > 0
       puts "from #{HOMEBREW_PREFIX}"
     end unless ARGV.dry_run?
+
+    unlinkapps_prune(:dry_run => ARGV.dry_run?, :quiet => true)
   end
 end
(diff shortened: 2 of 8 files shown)


Library/Homebrew/cmd/linkapps.rb
```diff
@@ -1,22 +1,21 @@
...
 
     if ARGV.named.empty?
-      kegs = Formula.racks.map do |rack|
-        keg = rack.subdirs.map { |d| Keg.new(d) }
-        next if keg.empty?
-        keg.detect(&:linked?) || keg.max { |a, b| a.version <=> b.version }
+      kegs = Formula.installed.map do |formula|
+        formula_kegs = formula.installed_kegs
+        next if formula_kegs.empty?
+        formula_kegs.detect(&:linked?) || formula_kegs.max_by(&:version)
```
xu-cheng: We shouldn't use `Formula.installed` here. The thing is commands like `link/unlink/switch/uninstall/linkapp` should work fine with DIY installation, i.e. no formula required. So let's use rack directly instead of trying to load formula.

UniqMartin: Thanks for pointing this out! I'll fix it shortly. DIY seems like such an esoteric feature that I tend to always forget about it (and I never used it).

xu-cheng: Beside the complete DIY installation(i.e. manually compiling while set prefix to cellar), another common situation is `brew install <URL/path to formula>` following with cache purge. Therefore, formula file will also be gone.

mikemcquaid:
Looks good to me :+1: 

> This part also provides Homebrew.unlinkapps_prune that could be used from cmd/uninstall.rb et al. to implement automatic pruning of broken symlinks as suggested in #46542 (will be addressed in a separate PR).

That would be great. I was going to take a look but you're already closer to it :wink:

UniqMartin:
I addressed the problem raised by @xu-cheng (thanks!) with the nice side-effect of being much faster now. Nothing else changed in what I just pushed.

> That would be great. I was going to take a look but you're already closer to it :wink:

I intended to do this anyway and as this issue of stale symlinks has resurfaced again in #46542, I decided to address it sooner than later. :smile_cat:

UniqMartin:
Updated to address a merge conflict. It's :white_check_mark: again.

Library/Homebrew/manpages/brew.1.md
```diff
@@ -451,6 +451,10 @@ With `--verbose` or `-v`, many commands print extra debugging information. Note
 
     If no <formulae> are provided, all linked app will be removed.
 
+  * `unlinkapps --prune`:
+    Removes links created by `brew linkapps` that are no longer valid, e.g.,
+    because a formula was removed or apps provided by a formula have changed.
```
mikemcquaid: Wondering if this behaviour should just be done by `brew prune`?

UniqMartin: This sounds like a better place for this functionality; thanks for pointing it out! Do you mean as default behavior for `brew prune` or as a separate mode of operation, e.g. `brew prune --apps`?

mikemcquaid: Default behaviour; I don't see a reason why you'd _not_ want to do this.

UniqMartin: Agreed; I'll update the code to move the new functionality into `brew prune`.

UniqMartin:
Relocated pruning of broken app symlinks into `brew prune` as discussed. Also added `--dry-run` option to `brew unlinkapps` (many destructive commands have this option, also for symmetry with `brew prune`) and partially rewrote the code from the first two commits.

Library/Homebrew/cmd/prune.rb
```diff
@@ -47,5 +48,7 @@ def prune
       print "and #{d} directories " if d > 0
       puts "from #{HOMEBREW_PREFIX}"
     end unless ARGV.dry_run?
+
+    unlinkapps_prune(:dry_run => ARGV.dry_run?, :quiet => true)
```
mikemcquaid: While you're here I wonder if it's worth just adding to `uninstall` too? No worries if you'd rather leave it for another PR.

UniqMartin: I'd prefer to do that in another PR as I expect some discussion. The reason is there are multiple places from where this likely needs to be called, though `uninstall` is the most obvious one.

Library/Homebrew/cmd/prune.rb
```diff
@@ -1,5 +1,6 @@
 require "keg"
 require "cmd/tap"
+require "cmd/unlinkapps"
```
mikemcquaid: I don't feel super-strongly about this but we're trying to avoid `cmd/` includes now so if you're feeling extra lovely (can be a future PR) pulling these methods into another module would be :heart:

mikemcquaid:
A few nits but nothing preventing this being merged if you'd rather do the rest in another PR :+1:

Library/Homebrew/cmd/prune.rb
UniqMartin: You are right and I think it makes more sense to address this here in this PR (big chunks have been rewritten anyway) instead of deferring the code move to a new PR. I'll reshuffle things soon. Thanks for the soft nudge. :heart_eyes_cat:

Library/Homebrew/cmd/prune.rb
mikemcquaid: :+1:

Library/Homebrew/cmd/prune.rb
mikemcquaid: :+1:

mikemcquaid:
@UniqMartin thoughts on merging as-is?

UniqMartin:
I was hoping to revamp things directly in here to avoid cross `cmd/` requirements, but if you're still fine with it, I'll rebase to resolve the merge conflict and then merge as-is, leaving the other cleanup for later.

mikemcquaid:
@UniqMartin Sounds good 👍

UniqMartin:
Thanks (also for the nudge)! Test failure should be fixed, too. I'll ship as soon as the bot is happy.

UniqMartin:
Closing in favor of Homebrew/brew#15 as this no longer cleanly applies in the new repository.

PR #46555, opened 2015-12-01
Ace: add a test

Moisan:
Add a test to the `ace` formula.

### DIFF
--- a/Library/Formula/ace.rb
+++ b/Library/Formula/ace.rb
@@ -33,10 +33,7 @@ class Ace < Formula
                    "static_libs=0",
                    "install"
 
-    cd "examples" do
-    system "make"
-    end
-
+    system "make", "-C", "examples"
     pkgshare.install "examples"
   end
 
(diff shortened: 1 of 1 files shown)


Library/Formula/ace.rb
```diff
@@ -32,5 +32,16 @@ def install
                    "shared_libs=1",
                    "static_libs=0",
                    "install"
+
+    cd "examples" do
+    system "make"
```
DomT4: This needs an indent, or you should be able to do something like `system "make", "-C", "examples"`

DomT4: Sorry, I was possibly unclear there. Just in case, meant:

```ruby
system "make", "-C", "examples"
pkgshare.install "examples"
```

Rather than:
```ruby
cd "examples" do
  system "make", "-C", "examples"
end
pkgshare.install "examples"
```

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/d40f41dd0b674de8fcdcf7bfa2e7a3e9c8c4e3f7. Thank you for another Homebrew contribution @Moisan! :smiley_cat:

PR #46559, opened 2015-12-02
giflib 5.1.1

jakepetroules:
This is the latest version. Also, update the description to that which is used by the giflib SourceForge website.

### DIFF
--- a/Library/Formula/giflib.rb
+++ b/Library/Formula/giflib.rb
@@ -1,8 +1,8 @@
 class Giflib < Formula
-  desc "GIF library using patented LZW algorithm"
+  desc "A library and utilities for processing GIFs"
   homepage "http://giflib.sourceforge.net/"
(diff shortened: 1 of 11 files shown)


DomT4:
It looks like more than revisions are going to be needed here. Failing to compile pretty badly even from source. Most likely need patches either grabbed from or submitted to various upstreams and applied to the broken projects.

Library/Formula/giflib.rb
```diff
@@ -1,8 +1,8 @@
 class Giflib < Formula
-  desc "GIF library using patented LZW algorithm"
+  desc "A library and utilities for processing GIFs"
```
dunn: Don't need the article here. Just "Library and utilities for processing GIFs"

dunn:
Yeah, we've left it at 4.x.x because so many dependents haven't been made compatible with 5.x.x.

Issue #46520, opened 2015-11-30
New version of gsl


nilsbecker:
Hi, gsl 2.1 is released (the formula is still at 1.16)

bfontaine:
gsl 2.1 might cause various issues, see #46237 and #45812.

nilsbecker:
thanks, good to know, was not aware.

PR #46569, opened 2015-12-02
gsl 2.1 (update formular)

ge11232002:
Update for gsl-2.1, released on 11 November 2015.


tschoonj:
see #46520 (and others...)

PR #46570, opened 2015-12-02
diffoscope 42 (new formula).

mikemcquaid:
diffoscope will try to get to the bottom of what makes files or directories different. It will recursively unpack archives of many kinds and transform various binary formats into more human readable form to compare them. It can compare two tarballs, ISO images, or PDF just as easily.


mikemcquaid:
Merged (and patch upstreamed).

PR #46571, opened 2015-12-02
update resourced setuptools to 18.7.1

felixonmars:


DomT4:
Want @tdsmith to sign off on this if possible, and we'll likely want this PR to be broken into individual PRs because the `pypy` builds alone take 90 minutes *each*.

tdsmith:
No objections but is there a reason it's important to update to this release?

felixonmars:
It's not important, I'm just bumping because it's outdated (too much).

DomT4:
Will let Tim decide on necessity here.

DomT4:
Passing on this for now. We'll handle non-urgent updates gradually rather than in bulk. Thank you for the submission though @felixonmars :heart:.

felixonmars:
Would you be interested if I split the commit? :P

DomT4:
Not unless Tim is particularly interested, or there are some kind of vulnerabilities in the older versions. Several of these formulae are huge compiles and would chew through CI time with glee. It's Tim's call.

xu-cheng:
@DomT4 I don't think we should pass this because it's not urgent. Some of them are outdated to much.

@felixonmars It would be nice to split the commit. e.g. one PR for one formula.

DomT4:
@xu-cheng I'm not sure on the upsides of setting aside ~5 hours of CI time for a bump that doesn't gain anything end users actually *need*, especially since we won't be adding formulae revisions and most users won't see that bump until the next update anyway.

xu-cheng:
@DomT4 but it will benefit users who install these formulae later :wink: Things like these rarely got updated. It's nice to have someone in community to improve them, no matter the gain.

>  ~5 hours of CI time 

We are not in backlog now. So I won't worry about that. Especially, if the commits are split by several PRs.

PR #46575, opened 2015-12-02
android-sdk: always install build/platform tools.

mikemcquaid:
Useful for #42545 but also these are generally used by all `android-sdk` users anyway.

### DIFF
--- a/Library/Formula/android-sdk.rb
+++ b/Library/Formula/android-sdk.rb
@@ -90,6 +90,9 @@ class AndroidSdk < Formula
       decoded_file.write Base64.decode64(File.read("adb.bash"))
       bash_completion.install decoded_file
     end
+
+    # automatically install platform and build tools
+    system "echo y | bash #{bin}/android --verbose update sdk --no-ui --all --filter platform-tools,build-tools-#{build_tools_version}"
   end
 
   def caveats; <<-EOS.undent


Library/Formula/android-sdk.rb
```diff
@@ -90,6 +90,9 @@ def install
       decoded_file.write Base64.decode64(File.read("adb.bash"))
       bash_completion.install decoded_file
     end
+
+    # automatically install platform and build tools
+    system "echo y | bash #{bin}/android --verbose update sdk --no-ui --all --filter platform-tools,build-tools-23.0.1"
```
bfontaine: Can’t we do something with `pipe_output` instead of this `echo y | …` here?

xu-cheng: Also may want to use `build_tools_version` instead of hard coded `23.0.1`.

mikemcquaid: Yeh, meant to hard code. We don't have `pipe_output` here and it's essentially doing the same thing (but doesn't display incremental output which makes debugging hard). Also, `yes |` doesn't work.

mikemcquaid:
Also removed the comments/caveats that are now fixed in cf3a857 and bumped revision in 396b9d3.

PR #46577, opened 2015-12-02
Enhance caveats section for Fish formulae

skalee:
Fish tab completions require manual rebuild. That has been added to the Caveats section. Furthermore, short note about stock web-based configuration manager `fish_config` has been included.

No post-install hook because of this commit: 8a108bed154ae79a766d59949f0e753484525a76 by @geoff-codes. There's no explanation why demon restart has been removed from that hook, so I decided to take the defensive approach as well.

### DIFF
--- a/Library/Formula/fish.rb
+++ b/Library/Formula/fish.rb
@@ -36,9 +36,16 @@ class Fish < Formula
       chsh -s #{HOMEBREW_PREFIX}/bin/fish
     to make fish your default shell.
 
-    If you are upgrading from an older version of fish, you should now run:
+    And optionally:
+      fish_config
+    to review configuration.
+
+    If you are upgrading from an older version of fish, you should:
+
+    1. Terminate the outdated fish daemon:
       killall fishd
-    to terminate the outdated fish daemon.
+    2. Update tab completions:
+      fish_update_completions
     EOS
   end
 


geoff-codes:
Note that `fishd` is not being _restarted_; it has been removed.
See discussion re: removal of post-install hook in https://github.com/Homebrew/homebrew/pull/41621#issuecomment-120731018, https://github.com/Homebrew/homebrew/pull/41631, et al. 

Otherwise, no qualms here except for the bit about running `fish_update_completions`.
As noted here  https://github.com/Homebrew/homebrew/pull/41631#discussion_r3442274, this is done automatically.

Library/Formula/fish.rb
```diff
@@ -36,9 +36,16 @@ def caveats; <<-EOS.undent
       chsh -s #{HOMEBREW_PREFIX}/bin/fish
     to make fish your default shell.
 
-    If you are upgrading from an older version of fish, you should now run:
+    And optionally:
+      fish_config
```
dunn: Is this specific to Homebrew? We try to avoid including general advice since it can become outdated and the more information in `caveats`, the less users actually read.

Library/Formula/fish.rb
```diff
@@ -36,9 +36,16 @@ def caveats; <<-EOS.undent
...
+    to review configuration.
```
mikemcquaid: Please remove this optionally section, it's not necessary.

Library/Formula/fish.rb
```diff
@@ -36,9 +36,16 @@ def caveats; <<-EOS.undent
...
+
+    If you are upgrading from an older version of fish, you should:
+
+    1. Terminate the outdated fish daemon:
       killall fishd
```
mikemcquaid: Mind just changing this to `killall fishd; fish_update_completions` to get it on one line

Library/Formula/fish.rb
geoff-codes: > Is this specific to Homebrew?

No; in fact, fish on first run prints a greeting,
```
Welcome to fish, the friendly interactive shell.
Type help for instructions on how to use fish.
```
Which really is more that sufficient to cover this.

Library/Formula/fish.rb
geoff-codes: @mikemcquaid
Respectfully, I have ask "why?". We [nuked `pkill fishd`](https://github.com/Homebrew/homebrew/commit/8a108bed154ae79a766d59949f0e753484525a76#diff-f45e0a4db63db6d7ac3617f42ccf6192L32) with the 2.2.0 release; there's really no harm in having a zombie `fishd` (a few hundred K of memory, max) until the next time the user reboots; and since the daemon has been eliminated, it won't reoccur after that.

Again, no offense intended, I fail to see any good reason for merging this PR; it just adds 'caveats'  bloat, IMHO, as none of the steps outlined are "required" by any means.

geoff-codes:
@skalee Can you cite your assertion,
> Fish tab completions require manual rebuild

?

---- 
I similarly once tried to add an initial `fish_update_completions` to the formula, and the (`fish` owners) [strongly opposed](https://github.com/Homebrew/homebrew/pull/41631#discussion_r34422741). Per @DomT4 it was dropped:

Quoting from the line notes in the PR linked above:
> I don't think this will do what you want, and also I would recommend against it - may create problems in multiuser environments, for example. fish runs fish_update_completions in the background on first launch anyway.

I'm definitely :-1: on this PR.

geoff-codes:
I'd also like to point out (since I seem to stand accused), that

> 8a108be by @geoff-codes. There's no explanation why demon restart has been removed from that hook

... is untrue. The explanation/documentation is in the line notes of the linked PR. As far as I'm aware, this is still acceptable "documentation" for changes make Homebrew PRs.

I'll admit that GitHub's 'hiding' of line notes on "outdated diffs" is irksome, but I think including documentation of every change made in the commit log is impractical.

Library/Formula/fish.rb
mikemcquaid: The `killall fishd` is already here.

mikemcquaid:
I think we'll pass on this PR, thanks.

PR #46587, opened 2015-12-02
bottle: make bottle checksums reproducible.

mikemcquaid:
I've set all the variable data to versions that are dependent on the latest formula commit date and various modification dates also to the latest formula commit date. With this if you rerun `brew bottle` multiple times in a row you will see the same checksum even if you have `brew reinstall`ed (as long as upstream does not hardcode e.g. the build date).

I debugged this with diffoscope and worked on this as part of the Athens 2015 reproducible builds workshop: https://reproducible-builds.org/events/athens2015/

CC @DomT4, @xu-cheng and @UniqMartin for thoughts here (I know you'll have some) :wink:

### DIFF
--- a/Library/Homebrew/cmd/bottle.rb
+++ b/Library/Homebrew/cmd/bottle.rb
@@ -174,6 +174,9 @@ module Homebrew
     filename = Bottle::Filename.create(f, bottle_tag, bottle_revision)
     bottle_path = Pathname.pwd/filename
 
+    tar_filename = filename.to_s.sub(/.gz$/, "")
+    tar_path = Pathname.pwd/tar_filename
+
     prefix = HOMEBREW_PREFIX.to_s
     cellar = HOMEBREW_CELLAR.to_s
 
@@ -183,7 +186,13 @@ module Homebrew
     relocatable = false
     skip_relocation = false
 
+    formula_source_time = f.stable.stage do
+      Pathname.pwd.to_enum(:find).map(&:mtime).max
+    end
+
     keg.lock do
+      original_tab = nil
+
       begin
         keg.relocate_install_names prefix, Keg::PREFIX_PLACEHOLDER,
           cellar, Keg::CELLAR_PLACEHOLDER
@@ -192,10 +201,25 @@ module Homebrew
 
         keg.delete_pyc_files!
 
+        tab = Tab.for_keg(keg)
+        original_tab = tab.dup
+        empty_tab = Tab.empty
+        tab["poured_from_bottle"] = empty_tab["poured_from_bottle"]
+        tab["HEAD"] = empty_tab["HEAD"]
+        tab["time"] = empty_tab["time"]
+        tab.write
+
+        keg.find {|k| File.utime(File.atime(k), formula_source_time, k) }
+
         cd cellar do
+          safe_system "tar", "cf", tar_path, "#{f.name}/#{f.pkg_version}"
+          File.utime(File.atime(tar_path), formula_source_time, tar_path)
+          relocatable_tar_path = "#{f}-bottle.tar"
+          mv tar_path, relocatable_tar_path
           # Use gzip, faster to compress than bzip2, faster to uncompress than bzip2
           # or an uncompressed tarball (and more bandwidth friendly).
-          safe_system "tar", "czf", bottle_path, "#{f.name}/#{f.pkg_version}"
+          safe_system "gzip", "-f", relocatable_tar_path
+          mv "#{relocatable_tar_path}.gz", bottle_path
         end
 
         if bottle_path.size > 1*1024*1024
@@ -222,6 +246,7 @@ module Homebrew
         raise
       ensure
         ignore_interrupts do
+          original_tab.write
           keg.relocate_install_names Keg::PREFIX_PLACEHOLDER, prefix,
             Keg::CELLAR_PLACEHOLDER, cellar
         end
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -741,6 +741,8 @@ class FormulaInstaller
 
     tab.tap = formula.tap
     tab.poured_from_bottle = true
+    tab.time = Time.now.to_i
+    tab.head = Homebrew.git_head
     tab.write
   end
 


Library/Homebrew/cmd/bottle.rb
```diff
@@ -192,10 +199,19 @@ def bottle_formula(f)
 
         keg.delete_pyc_files!
 
+        tab = Tab.for_keg(keg)
+        tab["HEAD"] = formula_last_commit_hash
+        tab["time"] = formula_last_commit_time.to_i
+        tab.write
```
xu-cheng: I don't think you should manipulate tab file here for following reasons:
* It's more intuitively to write tab file in formula installer rather than bottle command.
* if you want to introduce new time attribute, it's better to do it in `tab.rb`. So it will affect to all tab file as opposite to only affect bottles. 
* We already have a HEAD attribute here now which is set to be the Homebrew head commit during compilation. So we should change the existing logic instead of simpiling overwriting it to a value with different meaning. 
* Using formula last commit is not enough, because:
  * Tap commit for tap formulae is certainly  required. 
  * Tap commits for all tap dependencies are required as different version of  dependency will create different binary files.

Library/Homebrew/cmd/bottle.rb
```diff
@@ -183,6 +186,10 @@ def bottle_formula(f)
     relocatable = false
     skip_relocation = false
 
+    formula_last_commit_hash, formula_last_commit_time = \
+      *Utils.popen_read("git", "log", "--pretty=%H%cd", "--", f.path).split(" ")
+    formula_last_commit_time = Time.parse(formula_last_commit_time)
```
xu-cheng: As explained below, I think such logic should be more suitable to put in formula installer. While, here, we can read the time directly from tab file.

Library/Homebrew/cmd/bottle.rb
xu-cheng: Simply abandoning current HEAD attribute, which stores the information about what core code is used to build, can be a little tricky. The reason is we may introduce new core code which will generate different content. Therefore, core code version matters in reproducible bottles. 

Therefore, I suggest to keep current HEAD attribute as it is. In all, I think we should design tab file as such:
```
HEAD: homebrew core commit used for build. 
taps: a dict to store tap name and their
         corresponding commits. Where we should  
         enumerate all taps for formula or direct
         dependences. 
time: formula last commit time. 
```

This also means we should create new command or `install` mode to checkout corresponding commit for reproducible verification

mikemcquaid: It's debatable for me whether `HEAD` or `time` are useful in the tab file at all; I've never seen them used in debugging and they aren't used anywhere in the code. If we really want to keep these attributes as-is they are we need to create a separate tab file that lives outside the bottle (and therefore also possible the Cellar) otherwise you cannot make a reproducible build if you have your Homebrew core commit at a different version (which is an unnecessary limitation). Another option would be not installing the tab files in bottles or not just not including `HEAD` or `time` in bottles.

mikemcquaid: > The reason is we may introduce new core code which will generate different content. Therefore, core code version matters in reproducible bottles.

Also: this is fine for reproducibility as we can debug this relatively easily or store this information in e.g. the bottle commit/DSL. The problem with including it in the tap is that the bottle SHA256 changes unnecessarily every time the core HEAD changes.

mikemcquaid: @xu-cheng Thoughts?

UniqMartin: I think capturing the commit hash and commit time at this stage basically gets us nowhere. Due to our contribution model, both will have changed by the time the commit ends up in our `master` and the information in the tab inside the bottle won't help to find or verify the commit to work with.

UniqMartin:
My understanding of reproducible builds is that there are basically two components to them:

**1. Capture enough information to be able to reconstruct the build setup**

This will have to be at least:

- The OS X version we are building on, including any installed security updates and other variations that have an influence.
- The Xcode and/or CLT version we are building with.
- The commit hash (public, i.e. pushed to `master`) of core code that will drive the build, particularly all of `superenv`, but also other bits and pieces that have an influence.
- The commit hash (public, i.e. pushed to `master`) of the formula or (if not public yet) just a dump of the formula code and, if applicable, the tap name.
  - The same for all other formulae accessed via `Formula["<dependency>"]`, as these might have changed since the dependency was bottled.
- The environment variables that are going to influence the build.
- The exact version and build of all non-Homebrew dependencies, e.g. XQuartz.
- Recursively, the same information for all dependencies of the formula (those should be bottles).

**2. Eliminate all sources of nondeterminism in the build process**

The [documentation on the Reproducible Builds web site](https://reproducible-builds.org/docs/) is a fantastic resource and lists a lot of things that will need to be addressed. Things, that are specific to OS X and that come to mind:

- Mach-O files contain a UUID load command that is inserted by the linker. The contained UUID changes with every (?) build.

**Closing remarks**

I think reproducible builds are a great goal to strive for and if anyone finds enough time to work on that, it's definitively a worthwhile effort.

I also believe we first have to analyze the process from PR to bottle (including the build process) for a bunch of formulae of increasing complexity to understand what information we have to retain to be able to reconstruct the build setup. After this is done, we can start making changes to core to support this.

xu-cheng:
First, sorry about the late commend.

I do think that we have to store information like core code `HEAD` and even more like @UniqMartin suggests. The reason is that the major goal for reproducible builds is to be able to verify the integrity of a bottle file. This includes any current and historical bottle files. To achieve that, we have to store all the information related to compilation. Otherwise a so call *reproducible* bottle will suddenly become unreproducible if some unsaved env changes, and this will make reproducible builds useless. 

> otherwise you cannot make a reproducible build if you have your Homebrew core commit at a different version

That's not the cause. Because to verify a reproducible build, the first thing is to restore the original build environment which would include check commits of core and other associated taps.

Mike, if I guess correctly, you want to use reproducible builds to avoid unnecessary bottle regenerations, right? If so, I do understand such need, but that is not the top goal for reproducible builds. And beside, to solve this particular problem, we can compare two bottle internally by skipping the tab files.

> I think capturing the commit hash and commit time at this stage basically gets us nowhere. Due to our contribution model, both will have changed by the time the commit ends up in our master and the information in the tab inside the bottle won't help to find or verify the commit to work with.

I completely agree. I haven't yet thought out a way to handle commit hash change. We probably need to start using merge to preserve the commit number instead of `brew pull`. But I do have a solution on time. Instead of setting every files time to formula's last commit, I think it's better to set to the time of the source code. i.e. time inside source code's tarball or last commit date of source code for formulae using SCM.

I also agree the rest of @UniqMartin's comment. I think reproducible builds is an urgent need to secure our bottle infrastructure. And I think we need to talk and finalize more details. Therefore, it's unlikely to achieve it in one PR.

mikemcquaid:
> 1. Capture enough information to be able to reconstruct the build setup
> 
> This will have to be at least:

Well, there's not really a least or most here. There will be other things that can affect the build process beyond these things and it's unclear if all of these things will definitely affect the output bytes. I do agree that this is a good  list though.

> The OS X version we are building on, including any installed security updates and other variations that have an influence.
> The Xcode and/or CLT version we are building with.
> The commit hash (public, i.e. pushed to master) of core code that will drive the build, particularly all of superenv, but also other bits and pieces that have an influence.

This may be better using a tree hash rather than a commit hash.

> The commit hash (public, i.e. pushed to master) of the formula or (if not public yet) just a dump of the formula code and, if applicable, the tap name.

This may be better using a file contents hash rather than a commit hash.

> The same for all other formulae accessed via Formula["<dependency>"], as these might have changed since the dependency was bottled.

This may be better using a file contents hash rather than a commit hash.

> The environment variables that are going to influence the build.

Read: all environment variables :wink:.

> 2. Eliminate all sources of nondeterminism in the build process
> 
> The documentation on the Reproducible Builds web site is a fantastic resource and lists a lot of things that will need to be addressed. Things, that are specific to OS X and that come to mind:
> 
> Mach-O files contain a UUID load command that is inserted by the linker. The contained UUID changes with every (?) build.

This doesn't match my experience. With this PR I was able to `brew reinstall --build-from-source wget` and get an identical checksum.

> The reason is that the major goal for reproducible builds is to be able to verify the integrity of a bottle file. This includes any current and historical bottle files. To achieve that, we have to store all the information related to compilation. Otherwise a so call reproducible bottle will suddenly become unreproducible if some unsaved env changes, and this will make reproducible builds useless.

That's "a goal" not "the goal". The thing that kept coming up with the reproducible builds summit is there is no single goal for this stuff, it's about trying to get as much use from it as you can.

Even if we store all the information above and more: there will be times when users cannot reproduce builds; this came up repeatedly at the Reproducible Builds event as being something that's inevitable. This does not make them "useless". Even this PR as-is is useful: before it reproducible builds are impossible for all formulae, after it reproducible builds are possible for some formulae.

> That's not the cause. Because to verify a reproducible build, the first thing is to restore the original build environment which would include check commits of core and other associated taps.

That's not necessarily necessary. It may be necessary but it's not a hard requirement.

> Mike, if I guess correctly, you want to use reproducible builds to avoid unnecessary bottle regenerations, right? If so, I do understand such need, but that is not the top goal for reproducible builds. And beside, to solve this particular problem, we can compare two bottle internally by skipping the tab files.

Again, there's not a clear "top goal". I also think generating identical checksums is more useful than having to exclude files.

> I completely agree. I haven't yet found out a way to handle commit hash change. We probably need to start using merge to preserve the commit number instead of brew pull. But I do have a solution on time. Instead of setting every files time to formula's last commit, I think it's better to set to the time of the source code. i.e. time inside source code's tarball or last commit date of source code for formulae using SCM.

We're not going to start using merge. I think we can tweak the timestamp we use; using the latest timestamp from inside the source code tarball or checkout feels like it could be a nice middle ground.

> I also agree the rest of @UniqMartin's comments. I think reproducible build is an urgent need to secure our bottle infrastructure. And I think we need to talk and finalize more details. Therefore, it's unlikely to achieve it in one PR.

It's not urgent because it wasn't urgent last week.

We're not going to achieve it all in one PR: I agree. I spent most of last week at the Reproducible Builds event talking to people across many package managers about these issues and working and learning about solutions. Most of them seemed to agree that incremental progress is better than trying to try and solve this all in a single PR. I do think that this PR gets us closer than we were before and I'd like to know what in it you/@UniqMartin consider blocking and what can be iterated on.

What I think would probably be a helpful inclusion in this PR is either having two Tab files: one of which is inside the bottle and one of which is uploaded and stored outside the bottle or having the Tab file start always living outside bottles.

mikemcquaid:
> I do think that this PR gets us closer than we were before and I'd like to know what in it you/@UniqMartin consider blocking and what can be iterated on.

Just want to quote/highlight this as I'd like to know what items block this PR from being merged. Thanks!

xu-cheng:
> Just want to quote/highlight this as I'd like to know what items block this PR from being merged. Thanks!

In my option, we should focus on timestamp in this PR. As discussed, the preferred timestamp would be from the source code instead of formulae.

As for other, could we please remove the code on tweaking the tab file in this PR? The reasons are (a) we don't yet finalize what data should be stored; (b) the content of tab file is part of our public API which I tend to keep compatibility. On the other head, the `HEAD` attribute would be used in some other on-going work, e.g formula rename detection for installation. So it would be nice to keep it as it is until we sort out.

mikemcquaid:
> In my option, we should focus on timestamp in this PR. As discussed, the preferred timestamp would be from the source code instead of formulae.

I want to resolve both the timestamp and Tab time/commit issue in this PR.

> the content of tab file is part of our public API which I tend to keep compatibility.

We've never stated this is a public API and we've changed the format a few times ourselves with no user complaints ever. I'm extremely dubious that the `time` has ever been useful (particularly given the Cellar and/or bottle timestamps communicate the same information).

> On the other head, the HEAD attribute would be used in some other on-going work, e.g formula rename detection for installation.

We need to work out a way of keeping it outside the bottle, I think. Would you be opposed to having it be a separate metadata file that's uploaded alongside the bottle to Bintray? As stated earlier: information like this should be useful in reconstructing a reproducible bottle if you are unable to do so by default but as-is it's stopping a reproducible bottle from being created if you run `git commit` or `brew update` which is incorrect; the bottle may be dependent on that change in commit but it's definitely not guaranteed.

mikemcquaid:
> Would you be opposed to having it be a separate metadata file that's uploaded alongside the bottle to Bintray?

I also think this would be useful in allowing us to dump more build information like compilers etc. used during a build and could also be used in `brew gist-logs`

xu-cheng:
> I want to resolve both the timestamp and Tab time/commit issue in this PR.

Then let's not rush on merging it. At least, without solving all the problems.

> We've never stated this is a public API 

What I mean is it's part of file system.

> and we've changed the format a few times ourselves with no user complaints ever. 

But we have code in the core to handle old tab file.

> I'm extremely dubious that the time has ever been useful (particularly given the Cellar and/or bottle timestamps communicate the same information).

I agree this. I think `brew bottle` should just read timestamp from Cellar, assuming it has been set correctly during build.

> We need to work out a way of keeping it outside the bottle, I think. Would you be opposed to having it be a separate metadata file that's uploaded alongside the bottle to Binary?

I have no objection on storing it in separate location. However, I do want to strength
* Please keep tab file content unified regardless bottles and building from source. It's bad smell to have attribute with same name to have different meanings.
* Please keep `HEAD` attribute as it is. For one, it will be used in other features(i.e. formula rename); And more importantly, it's better to use a different name then overwriting it. Otherwise, you won't be able to distinguish them apart on old bottle files.

As for PR, the logic of extracting commit should be changed due to our `brew pull` mechanism. The reason has been well explained  by @UniqMartin

mikemcquaid:
> Then let's not rush on merging it. At least, without solving all the problems.

Sure.

> But we have code in the core to handle old tab file.

Yup, so as long as we handle old tab files we can change the format/meaning.

> I agree this. I think brew bottle should just read timestamp from Cellar, assuming it has been set correctly during build.

I prefer your idea of using the timestamp from the source file.

> Please keep tab file content unified regardless bottles and building from source. It's bad smell to have attribute with same name to have different meanings.

:+1: 

> Please keep HEAD attribute as it is. For one, it will be used in other features(i.e. formula rename); And more importantly, it's better to use a different name then overwriting it. Otherwise, you won't be able to distinguish them apart on old bottle files.

Sure. As long as they lives outside the bottle file I'm happy to keep all attributes as-is.

So, the changes I see to do here:
- [ ] use newest timestamp from the source package for the bottle file and any files in the `Cellar` that are newer than that
- [ ] pull `HEAD`/`time` attributes out to a separate file that's uploaded alongside the bottle tarball (although another crazy option here would be including the `HEAD` inside our actual bottle block)
- [ ] keep tab file the same for building from source and bottling

Sound good?

UniqMartin:
Sorry for the lack of feedback from my side. I'm afraid I have to withdraw from this discussion due to lack of time, though the topic interests me. (But I think you are starting to converge on a good first step. :+1:)

neverpanic:
On the last day of the summit, Thomas Klausner has written a small utility to simplify finding the newest timestamp of a source package: https://github.com/0-wiz-0/findnewest. This may be helpful for your first action item.

Note that this will not necessarily help you with source code fetched from version control systems, because version control systems do not necessarily set file mtimes.

mikemcquaid:
Hey @neverpanic, thanks for that :zap:.

mikemcquaid:
I've updated this based on my understanding of the results of the discussion.

Library/Homebrew/cmd/bottle.rb
```diff
@@ -183,6 +186,15 @@ def bottle_formula(f)
...
     skip_relocation = false
 
+    formula_source_time = f.stable.stage do
+      newest_mtime = nil
+      Pathname(Dir.pwd).find do |path|
+        newest_mtime ||= path.mtime
+        newest_mtime = path.mtime if path.mtime > newest_mtime
+      end
+      newest_mtime
+    end
```
UniqMartin: I was going to suggest

```
formula_source_time = Pathname(Dir.pwd).find.map(&:mtime).max
```
but then I realized that we are still stuck with Ruby 1.8.7. :sob: Could the (more compatible, more ugly)

```
formula_source_time = Pathname(Dir.pwd).to_enum(:find).map(&:mtime).max
```

still be a good enough and succinct replacement for this entire block?

UniqMartin: Oh, and why not replace `Pathname(Dir.pwd)` with `Pathname.pwd`?

Library/Homebrew/cmd/bottle.rb
```diff
@@ -192,10 +204,24 @@ def bottle_formula(f)
 
         keg.delete_pyc_files!
 
+        tab = Tab.for_keg(keg)
+        empty_tab = Tab.empty
+        tab["poured_from_bottle"] = empty_tab["poured_from_bottle"]
```
xu-cheng: This seems to be unnecessary.

Library/Homebrew/cmd/bottle.rb
mikemcquaid: It could be, yeh, good call.

Library/Homebrew/cmd/bottle.rb
mikemcquaid: It's necessary sometimes as it doesn't affect the output whether it's been poured from a bottle or not before (re)bottling  but it does affect the checksum.

xu-cheng: Sorry, but this attribute should always be false from the keg created by `brew install --build-bottle`. If it's true, I would suspect the keg has been altered and should not be used to create bottle.

Library/Homebrew/cmd/bottle.rb
```diff
@@ -192,10 +204,24 @@ def bottle_formula(f)
...
+        tab["HEAD"] = empty_tab["HEAD"]
+        tab["time"] = empty_tab["time"]
+        tab.write
```
xu-cheng: We probably need to restore these information back to tab file in an `ensure` block.  So we could run `brew bottle` multiple times(considering we will store these info somewhere else).

mikemcquaid: :+1:

Library/Homebrew/cmd/bottle.rb
mikemcquaid: If you rebottle a poured bottle the checksum should remain the same but this stops that being the case.

xu-cheng: > If you rebottle a poured bottle 

I don't think you could do this? For example, files in `etc` and `var` will be lost.

Library/Homebrew/cmd/bottle.rb
xu-cheng: For performance(to avoid unnecessary stage) and consistency(whether bottle or not), it would be better to put these logic in `formula_installer`. And here we could just read time from any file in keg(except tab file).

Library/Homebrew/cmd/bottle.rb
mikemcquaid: If that's the case: the checksum will change. As mentioned before: I'd rather make checksum changes the result of contents changing rather than metadata.

Library/Homebrew/cmd/bottle.rb
mikemcquaid: I'd rather the performance penalty be in `bottle` rather than `install` as there's a lot more `install`s happening than `bottle`s.

mikemcquaid: I'm not convinced the consistency of timestamps matters for non-bottles.

Library/Homebrew/cmd/bottle.rb
xu-cheng: Oh, I have no objection on logic here. But I do have a more general question: should we support rebottling a poured keg at all? To me, it feels a bad use case, considering bottles are likely useless.

mikemcquaid: I don't think we should particularly support it but it's something I find useful for local testing/debugging.

xu-cheng: > it's something I find useful for local testing/debugging.

:cool:

Library/Homebrew/cmd/bottle.rb
```diff
@@ -192,10 +204,24 @@ def bottle_formula(f)
...
+
+        keg.find {|k| File.utime(File.atime(k), formula_source_time, k) }
```
UniqMartin: If I'm not mistaken, this changes the modification time of the linked source file of a symbolic link instead of the link itself, thus messes up the `tar` archive meta data. Maybe worth looking into, as symlinks are relatively common?

UniqMartin:
I tried it locally with building and rebuilding `xz` and it mostly worked as advertised. Aside from some `tar` archive meta data, there were differences in the static library`liblzma.a` (but that's just another archive format that saves time stamps and other meta data, certainly outside the scope of this PR).

I think this is a good first step and almost ready to be merged.

Library/Homebrew/cmd/bottle.rb
mikemcquaid: I'll check this out in a future PR, if that's :ok:

Library/Homebrew/cmd/bottle.rb
xu-cheng: > I'd rather the performance penalty be in bottle rather than install as there's a lot more installs happening than bottles.

There isn't much penalty on install as we have to stage anyway in `install`. The only possible penalty would be scanning the file. But I highly doubt that would be a hotspot comparing to stage itself.

More importantly, this would mean network is required for bottle, which may be something we want to avoid. Especially for formulae who are huge. 

Finally, git doesn't store timestamp. So we will get different timestamp for different stage for formulae using `GitDownloadStrategy`. In such case, we should use git commit date as the timestamp. And we probably need to check other download strategies as well. Therefore, it may be more help to introduce `AbstractDownloadStrategy#newest_mtime` method. If so, putting logic in `formula_installer` make more sense.

mikemcquaid: > More importantly, this would mean network is required for bottle, which may be something we want to avoid. Especially for formulae who are huge.

If you've done `brew install --build-bottle` before this you'll have a cached download so shouldn't need the network.



> Finally, git doesn't store timestamp. So we will get different timestamp for different stage for formulae using GitDownloadStrategy. In such case, we should use git commit date as the timestamp. And we probably need to check other download strategies as well. Therefore, it may be more help to introduce AbstractDownloadStrategy#newest_mtime method. If so, putting logic in formula_installer make more sense.

Yup, that's another edge-case to be addressed (in a future PR if that's :ok:)

xu-cheng: > If you've done brew install --build-bottle before this you'll have a cached download so shouldn't need the network.

I guess what I'm hoping to make it work without network even if I purge the bottle.

> Yup, that's another edge-case to be addressed (in a future PR if that's :ok:)

I'm OK to merge this PR as it is. But I would certainly hope to put these logics in  `formula_installer` in long term due to the reasons I already present.

Library/Homebrew/cmd/bottle.rb
UniqMartin: Yes. It doesn't seem to be easy to do (at least directly from Ruby) and it might turn out it's more practical to manipulate the created `tar` archive.

Library/Homebrew/cmd/bottle.rb
xu-cheng: Also FYI, if we support `Tab#write` without timestamp change, we can simplify the logic even further. But I guess you want to consider this in future.

Library/Homebrew/cmd/bottle.rb
xu-cheng: We can shell out using `touch` for symlinks.

mikemcquaid:
@neverpanic I meant to ask: does `libfaketime` work on 10.11 for you? I wasn't sure if it was a Homebrew bug or if SIP was messing with it.

mikemcquaid:
@Homebrew/maintainers now that this is shipped: watch out for new bottle blocks where all the checksums are the same. That's a possibility now and it means we can consider providing a single bottle for multiple platforms.

neverpanic:
@mikemcquaid Never tried `libfaketime`, but as with any use of `DYLD_INSERT_LIBRARIES` on 10.11, SIP will prevent it. We've seen the problem in MacPorts and our approach is to copy the binary if it's under SIP and run the copy instead. That's of course additional effort that would have to be implemented in libfaketime as well… so fiddling with `DYLD_*` is basically dead on 10.11.

mikemcquaid:
@neverpanic Thanks!

PR #46592, opened 2015-12-02
formula: delete git-encrypt

maletor:
This package is deprecated and no longer used.


DomT4:
Should be "migrated" to Homebrew/boneyard, rather than deleted outright. See https://github.com/Homebrew/homebrew/pull/46489/files for an example.

maletor:
@DomT4, updated and [added related pull request](https://github.com/Homebrew/homebrew-boneyard/pull/79).

bfontaine:
I don’t think we should remove it from the core if it still builds. The boneyard is for formulae that don’t build anymore.

DomT4:
Normally I'd agree, but the author force pushed over the source code and added a note to the README that essentially hints it may not be secure:

> Please consider using AGWA/git-crypt instead. AGWA's tool has a proven security and is much better than this tool.

maletor:
Right, this tool is insecure and unsupported by most other package managers.

bfontaine:
Thanks; it makes sense.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/74dbae0ec95d756ec78c1f1fae852e95317eb6af. Thank you for your contribution to Homebrew @maletor; we appreciate it! :smiley_cat:

PR #46349, opened 2015-11-25
elasticsearch 2.1.0

jasontedor:
This pull request updates the Elasticsearch formula from version 2.0.0 to version
2.1.0, the latest stable version of Elasticsearch as of 2015-11-24.

Closes #46343.

PR #46593, opened 2015-12-02
kibana-4.3.0

jasontedor:
This pull request updates the Kibana formula from version 4.1.1 to version
4.3.0, the latest stable version of Kibana as of 2015-11-24.

Relates #46349

### DIFF
--- a/Library/Formula/kibana.rb
+++ b/Library/Formula/kibana.rb
@@ -1,48 +1,32 @@
 class Kibana < Formula
-  desc "Visualization tool for elasticsearch"
+  desc "Analytics and search dashboard for Elasticsearch"
   homepage "https://www.elastic.co/products/kibana"
-  url "https://github.com/elastic/kibana/archive/v4.1.1.tar.gz"
-  sha256 "3f91e99e20e82d4e84ec141007822fea8f9454c71595551f9348ea2609c98284"
+  url "https://download.elastic.co/kibana/kibana/kibana-4.3.0-darwin-x64.tar.gz"
+  sha256 "b9952879be618b179275a0a0fb6ef9121f4f6d8784f70fb9aa059a05e3991c9d"
+  version "4.3.0"
   head "https://github.com/elastic/kibana.git"
 
-  bottle do
-    cellar :any_skip_relocation
-    sha256 "a63101c190529aa0798ff7dabe7e7ee3fe8a6a04a126d324eb58fa8343e3a67d" => :el_capitan
-    sha256 "f8768f21442e9ce85b3c9b3da58d8e73c176d7bf656b636d15ab582074f14991" => :yosemite
-    sha256 "0bcea2ae726c5a175660c6350f4d60ee07a364d1574c90a2bd691ee25123afbb" => :mavericks
-    sha256 "5d102f006b25dae70bd62c99d9787abe10111460bc2119e5e5ceb412d6335dae" => :mountain_lion
-  end
-
-  depends_on "node"
+  bottle :unneeded
 
   def install
-    ENV.prepend_path "PATH", "#{Formula["node"].opt_libexec}/npm/bin"
+    rm_f Dir["bin/*.bat"]
+    prefix.install Dir["*"]
 
-    system "npm", "install"
-    system "npm", "install", "grunt-cli"
-    system "npm", "install", "bower"
-    system "./node_modules/.bin/bower", "install"
-    system "./node_modules/.bin/grunt", "build", "--force"
-
-    dist_dir = buildpath/"build/dist/kibana"
-
-    rm_f dist_dir/"bin/*.bat"
-
-    prefix.install dist_dir/"src"
-    (etc/"kibana").mkpath
-    (var/"lib/kibana/plugins").mkpath
-
-    (etc/"kibana").install dist_dir/"config/kibana.yml" unless (etc/"kibana/kibana.yml").exist?
-
-    # point to our node
-    inreplace dist_dir/"bin/kibana" do |s|
-      s.sub! /^NODE=.*$/, "NODE=#{Formula["node"].opt_bin}/node"
+    inreplace "#{prefix}/config/kibana.yml" do |s|
+      s.sub!(%r{/var/run/kibana.pid}, "/usr/local/var/run/kibana.pid")
     end
+    (etc/"kibana").install Dir[prefix/"config/kibana.yml"] unless (etc/"kibana/kibana.yml").exist?
+  end
 
-    bin.install dist_dir/"bin/kibana"
+  def post_install
+    ln_s etc/"kibana", prefix/"config"
 
-    (prefix/"config").install_symlink etc/"kibana/kibana.yml"
-    prefix.install_symlink var/"lib/kibana/plugins"
+    (var/"lib/kibana").mkpath
+    rm_f var/"lib/kibana/node_modules"
+    ln_s prefix/"node_modules", var/"lib/kibana/node_modules"
+
+    (var/"lib/kibana/installedPlugins").mkpath
+    ln_s var/"lib/kibana/installedPlugins", prefix/"installedPlugins"
   end
 
   plist_options :manual => "kibana"
@@ -65,6 +49,7 @@ class Kibana < Formula
   end
 
   test do
+    ENV["BABEL_CACHE_PATH"] = testpath/".babelcache.json"
     assert_match /#{version}/, shell_output("#{bin}/kibana -V")
   end
 end
--- a/Library/Formula/kibana.rb
+++ b/Library/Formula/kibana.rb
@@ -34,6 +34,11 @@ class Kibana < Formula
       s.gsub!(%r{('(linux-x64|linux-x86|windows)',?(?!;))}, "// \\1")
     end
 
+    # do not build zip package
+    inreplace buildpath/"tasks/build/archives.js" do |s|
+      s.gsub!(%r{(await exec\('zip'.*)}, "// \\1")
+    end
+
     ENV.prepend_path "PATH", prefix/"libexec/node/bin"
     system "npm", "install"
     system "npm", "run", "build"
--- a/Library/Formula/kibana.rb
+++ b/Library/Formula/kibana.rb
@@ -24,6 +24,11 @@ class Kibana < Formula
       system "make", "install"
     end
 
+    # do not download binary installs of Node.js
+    inreplace buildpath/"tasks/build/index.js" do |s|
+      s.gsub!(%r{('_build:downloadNodeBuilds:\w+',)}, "// \\1")
+    end
+
     ENV.prepend_path "PATH", prefix/"libexec/node/bin"
     system "npm", "install"
     system "npm", "run", "build"
--- a/Library/Formula/kibana.rb
+++ b/Library/Formula/kibana.rb
@@ -29,6 +29,11 @@ class Kibana < Formula
       s.gsub!(%r{('_build:downloadNodeBuilds:\w+',)}, "// \\1")
     end
 
+    # do not build packages for other platforms
+    inreplace buildpath/"tasks/config/platforms.js" do |s|
+      s.gsub!(%r{('(linux-x64|linux-x86|windows)',?(?!;))}, "// \\1")
+    end
+
     ENV.prepend_path "PATH", prefix/"libexec/node/bin"
     system "npm", "install"
     system "npm", "run", "build"
--- a/Library/Formula/kibana.rb
+++ b/Library/Formula/kibana.rb
@@ -25,19 +25,13 @@ class Kibana < Formula
     end
 
     # do not download binary installs of Node.js
-    inreplace buildpath/"tasks/build/index.js" do |s|
-      s.gsub!(%r{('_build:downloadNodeBuilds:\w+',)}, "// \\1")
-    end
+    inreplace buildpath/"tasks/build/index.js", %r{('_build:downloadNodeBuilds:\w+',)}, "// \\1"
 
     # do not build packages for other platforms
-    inreplace buildpath/"tasks/config/platforms.js" do |s|
-      s.gsub!(%r{('(linux-x64|linux-x86|windows)',?(?!;))}, "// \\1")
-    end
+    inreplace buildpath/"tasks/config/platforms.js", %r{('(linux-x64|linux-x86|windows)',?(?!;))}, "// \\1"
 
     # do not build zip package
-    inreplace buildpath/"tasks/build/archives.js" do |s|
-      s.gsub!(%r{(await exec\('zip'.*)}, "// \\1")
-    end
+    inreplace buildpath/"tasks/build/archives.js", %r{(await exec\('zip'.*)}, "// \\1"
 
     ENV.prepend_path "PATH", prefix/"libexec/node/bin"
     system "npm", "install"
@@ -52,12 +46,8 @@ class Kibana < Formula
   end
 
   def post_install
-    inreplace "#{bin}/kibana" do |s|
-      s.sub!(%r{/node/bin/node}, "/libexec/node/bin/node")
-    end
-    inreplace "#{prefix}/config/kibana.yml" do |s|
-      s.sub!(%r{/var/run/kibana.pid}, "/usr/local/var/run/kibana.pid")
-    end
+    inreplace "#{bin}/kibana", %r{/node/bin/node}, "/libexec/node/bin/node"
+    inreplace "#{prefix}/config/kibana.yml", %{/var/run/kibana.pid}, "/usr/local/var/run/kibana.pid"
     (etc/"kibana").install prefix/"config/kibana.yml" unless (etc/"kibana/kibana.yml").exist?
     rm_rf prefix/"config"
     ln_s etc/"kibana", prefix/"config"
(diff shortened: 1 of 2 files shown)


c9n:
:+1: works

Library/Formula/kibana.rb
```diff
@@ -1,48 +1,32 @@
 class Kibana < Formula
-  desc "Visualization tool for elasticsearch"
+  desc "Analytics and search dashboard for Elasticsearch"
   homepage "https://www.elastic.co/products/kibana"
-  url "https://github.com/elastic/kibana/archive/v4.1.1.tar.gz"
-  sha256 "3f91e99e20e82d4e84ec141007822fea8f9454c71595551f9348ea2609c98284"
+  url "https://download.elastic.co/kibana/kibana/kibana-4.3.0-darwin-x64.tar.gz"
```
DomT4: This looks like a binary release, rather than a source release?

c9n: @DomT4 It is binary release.

DomT4: We'd prefer to build from source, per Homebrew submission rules. (Alternatively this could be shipped off to `homebrew/binary` or something).

c9n: @DomT4 It takes a long time to build `kibana` from source. Really hurts. I do agree moving it to `homebrew/binary`.

DomT4: Won't be a huge problem for most of our users though, because we produce our own binaries for the three most recent OS X versions.

c9n: Is here a doc about `how to produce binaries on bintray for homebrew`? I tried but haven’t found.

DomT4: On our side? Happens automatically. The CI bumps out bottles for every formula build unless we tell it we don't need 'em.

c9n: @DomT4 Find this: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Brew-Test-Bot.md

DomT4: It's in the middle of being updated. Some our docs have been a little neglected. Did you have specific questions/concerns on the process?

c9n: I want to contribute, so I am curious about how it really works.

DomT4: Alright. Let us know if we can help there. The bottle generation logic for the CI specifically is stored in `Library/Homebrew/cmd/test-bot.rb` if you want to poke around locally.

c9n: Thanks very much. I will read the codes.

jasontedor: @DomT4 Thanks for the feedback here. I'll probably be looking at modifying this pull request to use a source release rather than a binary distribution but I might not get to it until next week.

DomT4: Thanks! No worries, give us/me a shout when the change is made so we spot it.

jasontedor:
> Thanks! No worries, give us/me a shout when the change is made so we spot it.

@DomT4 I've spent a fair amount of time exploring some options here. Right now, the root problem is that Kibana is tested and supported against Node.js 0.12.7 (for version 4.3.0). While there is an open issue to get to an LTS release in elastic/kibana#5541, that's not slated for the 4.x line of Kibana.

Looking past that, I tried to get Kibana running against the latest Node.js, but it turns out we run into nodejs/node#3606 (actually nodejs/node#3922 but that issue is just marked as a duplicate of nodejs/node#3606).

The two proceedings paragraphs combined strongly suggest that it would be preferable to use Node.js 0.12.7 for the time being. That presents several immediate options.

1. Use the bundled Node.js that comes with the binary distribution (as submitted with this pull request) with the downside of not being a source distribution.
2. Depend on the versioned formula node012 with the downside of conflicting with existing formula node creating a mess for many users.
3. Depend on node using npm to install [n](https://github.com/tj/n) to install Node.js 0.12.7 with the downside being this ultimately lands us back to depending on a binary install of Node.js.
4. Depend on the nvm formula to install Node.js 0.12.7 with the same downside of depending on a binary install of Node.js.
5. During the install of the kibana formula, download and compile Node.js 0.12.7 from source and install it alongside kibana (effectively how the binary distributions are packaged to begin with).

So we have three options that depend on a binary install, one that creates ugly conflicts with the node formula, and one that is going to duplicate the existing node012 formula to avoid conflicting with the node formula.

But before any firm conclusions are reached, I have a few questions. Is there an option that I'm not seeing? Is there an easy way to manage the conflicts in option 2. that doesn't create a mess for users? In particular, is there a way to just install the node012 formula inside the Kibana install?

mikemcquaid:
> 5. During the install of the kibana formula, download and compile Node.js 0.12.7 from source and install it alongside kibana (effectively how the binary distributions are packaged to begin with).

This seems like the nicest option to me in terms of avoiding side-effects and being consistent with how we handle other formulae. One question: is that version of Node just needed at build time or also needed at run time? Thanks!

jasontedor:
> This seems like the nicest option to me in terms of avoiding side-effects and being consistent with how we handle other formulae. 

Okay, I'll work on a version of the pull request that does this, but probably won't get to it until early next week.

> One question: is that version of Node just needed at build time or also needed at run time?

It's needed at runtime.

mikemcquaid:
@jasontedor Sounds good, thanks!

DomT4:
> This seems like the nicest option to me in terms of avoiding side-effects and being consistent with how we handle other formulae.

:+1:. Should be dumped into `libexec` and only the `kibana` binary linked into the top-level prefix, though. I presume if they live inside the same "real" prefix and you use an exec script that shouldn't be problematic (?).

jasontedor:
> I presume if they live inside the same "real" prefix and you use an exec script that shouldn't be problematic (?).

Yes. I don't foresee that being a problem. :)

jasontedor:
@mikemcquaid @DomT4 I've pushed b85a0f3e329404fc74fde47b00f21510a874426a that builds Node.js from source alongside Kibana. One hacky thing I don't like is a change to the [audit](https://github.com/jasontedor/homebrew/commit/b85a0f3e329404fc74fde47b00f21510a874426a#diff-fab7ad24732f89faa5deb3bb1f253a52R613) logic to exclude the kibana formula from the npm check so any suggestions there would be appreciated.

Library/Formula/kibana.rb
```diff
@@ -13,36 +12,58 @@ class Kibana < Formula
...
 
-    system "npm", "install"
-    system "npm", "install", "grunt-cli"
-    system "npm", "install", "bower"
-    system "./node_modules/.bin/bower", "install"
-    system "./node_modules/.bin/grunt", "build", "--force"
+    # do not download binary installs of Node.js
+    inreplace buildpath/"tasks/build/index.js" do |s|
+      s.gsub!(%r{('_build:downloadNodeBuilds:\w+',)}, "// \\1")
+    end
```
mikemcquaid: Can just use the `inreplace` function form when doing a single replacement in a file.

Library/Formula/kibana.rb
```diff
@@ -13,36 +12,58 @@ class Kibana < Formula
...
 
-    prefix.install dist_dir/"src"
-    (etc/"kibana").mkpath
-    (var/"lib/kibana/plugins").mkpath
+    ENV.prepend_path "PATH", prefix/"libexec/node/bin"
+    system "npm", "install"
+    system "npm", "run", "build"
+    mkdir "tar"
+    system "tar", "--strip-components", "1", "-xf", Dir["target/kibana-*-darwin-x64.tar.gz"].first, "-C", "tar"
```
mikemcquaid: Maybe work using `Dir["target/kibana-*.tar.gz"]` as the glob just to avoiding definitely breaking 32-bit or Linuxbrew builds.

Library/Formula/kibana.rb
jasontedor: @mikemcquaid Sorry if I'm misunderstanding, there are two replacements here: one for the `_build:downloadNodeBuilds:start` task, and one for the `_build:downloadNodeBuilds:finish` task.

Library/Formula/kibana.rb
```diff
@@ -13,36 +12,58 @@ class Kibana < Formula
...
+    # do not download binary installs of Node.js
+    inreplace buildpath/"tasks/build/index.js" do |s|
+      s.gsub!(%r{('_build:downloadNodeBuilds:\w+',)}, "// \\1")
+    end
 
-    dist_dir = buildpath/"build/dist/kibana"
+    # do not build packages for other platforms
+    inreplace buildpath/"tasks/config/platforms.js" do |s|
+      s.gsub!(%r{('(linux-x64|linux-x86|windows)',?(?!;))}, "// \\1")
+    end
```
mikemcquaid: Can just use the `inreplace` function form when doing a single replacement in a file.

Library/Formula/kibana.rb
```diff
@@ -13,36 +12,58 @@ class Kibana < Formula
...
 
-    rm_f dist_dir/"bin/*.bat"
+    # do not build zip package
+    inreplace buildpath/"tasks/build/archives.js" do |s|
+      s.gsub!(%r{(await exec\('zip'.*)}, "// \\1")
+    end
```
mikemcquaid: Can just use the `inreplace` function form when doing a single replacement in a file.

Library/Formula/kibana.rb
```diff
@@ -13,36 +12,58 @@ class Kibana < Formula
...
 
-    prefix.install dist_dir/"src"
-    (etc/"kibana").mkpath
-    (var/"lib/kibana/plugins").mkpath
+    ENV.prepend_path "PATH", prefix/"libexec/node/bin"
+    system "npm", "install"
+    system "npm", "run", "build"
+    mkdir "tar"
```
mikemcquaid: Could use `mkdir "tar" do` and a block and mean that you don't need to prefix `tar` into the commands below.

Library/Formula/kibana.rb
```diff
@@ -13,36 +12,58 @@ class Kibana < Formula
...
+    system "tar", "--strip-components", "1", "-xf", Dir["target/kibana-*-darwin-x64.tar.gz"].first, "-C", "tar"
 
-    (etc/"kibana").install dist_dir/"config/kibana.yml" unless (etc/"kibana/kibana.yml").exist?
+    rm_f Dir["tar/bin/*.bat"]
+    ["bin", "config", "node_modules", "optimize", "package.json", "src", "webpackShims"].each do |s|
+      prefix.install "tar/#{s}"
+      end
```
mikemcquaid: This is misaligned but would be nice to just pass these as multiple arguments to `prefix.install` after using `mkdir do` above.

Library/Formula/kibana.rb
```diff
@@ -13,36 +12,58 @@ class Kibana < Formula
...
+  end
 
-    # point to our node
-    inreplace dist_dir/"bin/kibana" do |s|
-      s.sub! /^NODE=.*$/, "NODE=#{Formula["node"].opt_bin}/node"
+  def post_install
+    inreplace "#{bin}/kibana" do |s|
+      s.sub!(%r{/node/bin/node}, "/libexec/node/bin/node")
     end
```
mikemcquaid: Can just use the `inreplace` function form when doing a single replacement in a file.

Library/Formula/kibana.rb
```diff
@@ -13,36 +12,58 @@ class Kibana < Formula
...
+    inreplace "#{prefix}/config/kibana.yml" do |s|
+      s.sub!(%r{/var/run/kibana.pid}, "/usr/local/var/run/kibana.pid")
+    end
```
mikemcquaid: Can just use the `inreplace` function form when doing a single replacement in a file.

Library/Formula/kibana.rb
```diff
@@ -13,36 +12,58 @@ class Kibana < Formula
...
+    (etc/"kibana").install prefix/"config/kibana.yml" unless (etc/"kibana/kibana.yml").exist?
```
mikemcquaid: Doing an `etc.install` in `install` will prevent overwriting and then perhaps doing just a conditional `inreplace` in `post_install` if needed.

Library/Homebrew/cmd/audit.rb
```diff
@@ -610,7 +610,7 @@ def audit_text
       problem "\"Formula.factory(name)\" is deprecated in favor of \"Formula[name]\""
     end
 
-    if text =~ /system "npm", "install"/ && text !~ %r[opt_libexec\}/npm/bin]
+    if text =~ /system "npm", "install"/ && text !~ %r[opt_libexec\}/npm/bin] && text !~ %r[kibana]
```
DomT4: Might just be able to do something like:
```ruby
if text =~ /system "npm", "install"/ && text !~ %r[opt_libexec\}/npm/bin]
  return if formula.name.start_with?("kibana")
```
Or something similar. Just working off the top of my head there though, please test first :smile:.

mikemcquaid: probably preferable to use `formula.name == "kibana"` instead

DomT4: :+1:

mikemcquaid: @DomT4 I was thinking the same but the `return` here would mean other audits wouldn't run, unfortunately.

DomT4: Oh, I thought you could return *inside* a block and it'd just skip onto the next block rather than bailing entirely? :cry:

Library/Formula/kibana.rb
jasontedor: Thanks. Addressed in dfc3bc7c9fc4fc6b201a918ef147a45f3480b5c4.

Library/Formula/kibana.rb
jasontedor: Addressed in dfc3bc7c9fc4fc6b201a918ef147a45f3480b5c4.

Library/Formula/kibana.rb
jasontedor: Addressed in dfc3bc7c9fc4fc6b201a918ef147a45f3480b5c4.

Library/Homebrew/cmd/audit.rb
mikemcquaid: That'd be `next` and it's only in blocks so e.g. `do` or `{ .. }`

Library/Formula/kibana.rb
jasontedor: Addressed in dfc3bc7c9fc4fc6b201a918ef147a45f3480b5c4.

Library/Formula/kibana.rb
jasontedor: Addressed in dfc3bc7c9fc4fc6b201a918ef147a45f3480b5c4.

Library/Homebrew/cmd/audit.rb
DomT4: Blurgh. My bad. Thought the same functionality worked with `return` as long as it was nested. That's a pain.

jasontedor: Addressed in d1141f3fbc402cd73b61b758ae048cf4bb2d8165.

DomT4: Maybe
```ruby
 if text =~ /system "npm", "install"/ && text !~ %r[opt_libexec\}/npm/bin] && formula.name != "kibana"
```
Or something. Just in case the word `kibana` suddenly becomes commonplace or anything.

DomT4: Ah, you beat me to the idea. Thanks @jasontedor :smiley:

Library/Formula/kibana.rb
jasontedor: That's nicer; addressed in 92f9774491fd721e10f70b66c140de2eb9f4a5fe.

Library/Formula/kibana.rb
jasontedor: Addressed in 92f9774491fd721e10f70b66c140de2eb9f4a5fe.

Library/Formula/kibana.rb
jasontedor: I pushed ed5b6c1d037da1178f3a21114510aa3f6fe57139.

Library/Formula/kibana.rb
jasontedor: Hmm...I wonder why that's better? This feels like a post install task to me (but I'm still coming up to speed on Homebrew idioms).

jasontedor:
@mikemcquaid Thanks for the thorough review; I've addressed or responded to all of your comments and think this is ready for another review cycle?

Library/Formula/kibana.rb
mikemcquaid: @jasontedor In handles rewriting and writing out a default version in a slightly cleverer/more consistent way so it's preferable, if you don't mind.

mikemcquaid:
@jasontedor One final comment then :+1: from me!

Library/Formula/kibana.rb
jasontedor: Got it, thanks! I pushed 854bd6e8d8cf24b57b3ccc3c1db6a2e9b8eaf658.

jasontedor:
> One final comment then :+1: from me!

@mikemcquaid I've addressed this final comment, but also pushed some caveats in 8be0508c79333882bf5621a28095a51605a84519. I think we're ready now contingent on the CI builds?

mikemcquaid:
Thanks for the great work @jasontedor!

jasontedor:
Thanks again @mikemcquaid and @DomT4!

PR #46594, opened 2015-12-02
Disable AESNI on OSX <= 10.7 for pandoc dependency

frickler01:
Pandoc depends on cryptonite, which fails on OSX <= 10.7,
since the system compiler doesn't understand the '-maes' option.
See: https://hackage.haskell.org/package/cryptonite

Unfortunately the changed code - as opposed to running `--interactive` and typing the exact same command - does not work yet and results in

    ==> cabal install --jobs=2 --constraint="cryptonite -support_aesni"
    cabal: expected a package name followed by a constraint, which is either a version range, 'installed', 'source' or flags

Maybe someone more proficient in the fine art of brewing can help out?

PS: Otherwise building works with an up to date GCC, but you need to have one (not necessarily the case) and type something like

    brew install pandoc --cc=gcc-5

Maybe that could be taken care of in the formula somehow.

### DIFF
--- a/Library/Formula/pandoc.rb
+++ b/Library/Formula/pandoc.rb
@@ -25,10 +25,7 @@ class Pandoc < Formula
   def install
     cabal_sandbox do
       args = []
-      if MacOS.version <= :lion
-        # disable AESNI for OSX <= 10.7
-        args << "--constraint=cryptonite -support_aesni"
-      end
+      args << "--constraint=cryptonite -support_aesni" if MacOS.version <= :lion
       cabal_install "--only-dependencies", *args
       cabal_install "--prefix=#{prefix}"
     end
--- a/Library/Formula/pandoc.rb
+++ b/Library/Formula/pandoc.rb
@@ -24,7 +24,12 @@ class Pandoc < Formula
 
   def install
     cabal_sandbox do
-      cabal_install "--only-dependencies"
+      args = []
+      if MacOS.version <= :lion
+        # disable AESNI for OSX <= 10.7
+        args << "--constraint=cryptonite -support_aesni"
+      end
+      cabal_install "--only-dependencies", *args
       cabal_install "--prefix=#{prefix}"
     end
     cabal_clean_lib


Library/Formula/pandoc.rb
```diff
@@ -24,7 +24,12 @@ class Pandoc < Formula
 
   def install
     cabal_sandbox do
-      cabal_install "--only-dependencies"
+      args = []
+      if MacOS.version <= :lion
+        # disable AESNI for OSX <= 10.7
+        args << "--constraint='cryptonite -support_aesni'"
```
mistydemeo: Try removing the quotes. These arguments aren't being passed to a shell, so shell splitting/quoting rules don't apply.

frickler01: That was it, now working. Thanks a lot!

How would we optionally build with a current GCC? With the new check, even `--cc=gcc-5` would not use AESNI. Is there a way to catch the cc flag if passed and then see if the version is good enough?

Library/Formula/pandoc.rb
```diff
@@ -24,7 +24,12 @@ class Pandoc < Formula
 
   def install
     cabal_sandbox do
-      cabal_install "--only-dependencies"
+      args = []
+      if MacOS.version <= :lion
+        # disable AESNI for OSX <= 10.7
+        args << "--constraint=cryptonite -support_aesni"
```
mikemcquaid: Can do this in one line: `args << "--constraint=cryptonite" << "-support_aesni" if MacOS.version <= :lion` and skip the comment as it's self-explanatory from the code.

frickler01: Done, it needs to be one argument though.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/128f031a5620a896eeb6a0fa1949e00303377f22. Thank you for your contribution to Homebrew @frickler01; we appreciate it! :smiley_cat:

PR #46596, opened 2015-12-02
Improve libbpg formula

bilhackmac:
- Remove deps to x265 since x265 is included in source
- Allow to disable new default encoder (x265)
- Allow to enable old encode (jctvc)

### DIFF
--- a/Library/Formula/libbpg.rb
+++ b/Library/Formula/libbpg.rb
@@ -16,7 +16,7 @@ class Libbpg < Formula
   option "without-x265", "Disable built-in x265 encoder - Multi threaded, faster but produce bigger file"
 
   depends_on "cmake" => :build
-  depends_on "yasm" => :build if !build.without? "x265"
+  depends_on "yasm" => :build if build.with? "x265"
   depends_on "libpng"
   depends_on "jpeg"
 
@@ -40,7 +40,7 @@ class Libbpg < Formula
       #system "#{bin}/bpgenc", "-e", "jctvc", test_fixtures("test.jpg")
     end
 
-    if !build.without? "x265"
+    if build.with? "x265"
       system "#{bin}/bpgenc", "-e", "x265", test_fixtures("test.png")
       # Unable to test 1x1 jpg
       #system "#{bin}/bpgenc", "-e", "x265", test_fixtures("test.jpg")
--- a/Library/Formula/libbpg.rb
+++ b/Library/Formula/libbpg.rb
@@ -3,6 +3,7 @@ class Libbpg < Formula
   homepage "http://bellard.org/bpg/"
   url "http://bellard.org/bpg/libbpg-0.9.6.tar.gz"
   sha256 "2800777d88a77fd64a4a9036b131f021a5bda8304e1dbe7996dd466567fb484e"
+  revision 1
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


Library/Formula/libbpg.rb
```diff
@@ -11,24 +12,38 @@ class Libbpg < Formula
...
+      system "#{bin}/bpgenc", "-e", "jctvc", test_fixtures("test.png")
+      # Unable to test 1x1 jpg
+      #system "#{bin}/bpgenc", "-e", "jctvc", test_fixtures("test.jpg")
+    end
+
+    if build.with? "x265"
+      system "#{bin}/bpgenc", "-e", "x265", test_fixtures("test.png")
+      # Unable to test 1x1 jpg
+      #system "#{bin}/bpgenc", "-e", "x265", test_fixtures("test.jpg")
+    end
```
bfontaine: We should only test the default case here.

Library/Formula/libbpg.rb
```diff
@@ -3,6 +3,7 @@ class Libbpg < Formula
   homepage "http://bellard.org/bpg/"
   url "http://bellard.org/bpg/libbpg-0.9.6.tar.gz"
   sha256 "2800777d88a77fd64a4a9036b131f021a5bda8304e1dbe7996dd466567fb484e"
+  revision 1
```
bfontaine: Is this change really worth re-building the formula?

bilhackmac: I don't really know, I thought that if formula is updated without version update, `revision` was needed.
I can remove it, but I think it can warn users that default encoder change from `jctvc` to `x265`, this is one of major change between `0.9.5` and `0.9.6`

Library/Formula/libbpg.rb
bilhackmac: In this case, if you want only old (and best) encoder `--with-jctvc --without-x265`, test will always fail. If it's ok for you, I'll remove these cases.

Library/Formula/libbpg.rb
mikemcquaid: I don't think it's needed.

mikemcquaid: (so you can remove this line)

Library/Formula/libbpg.rb
```diff
@@ -11,24 +12,38 @@ class Libbpg < Formula
     sha256 "7b8d1585ee9e2de010abfab1aaab419d052cf35d82704587c2f75947d4fc8ee5" => :mavericks
   end
 
-  option "with-x265", "Enable x265 encoder"
-  option "without-jctvc", "Disable built-in JCTVC encoder"
+  option "with-jctvc", "Enable built-in JCTVC encoder - Mono threaded, slower but produce smaller file"
+  option "without-x265", "Disable built-in x265 encoder - Multi threaded, faster but produce bigger file"
```
mikemcquaid: Which of these is the upstream default? We should default to what they do.

Library/Formula/libbpg.rb
mikemcquaid: That's fine, the bot will only test the default case.

Library/Formula/libbpg.rb
```diff
@@ -11,24 +12,38 @@ class Libbpg < Formula
...
-  depends_on "x265" => :optional
 
   def install
     bin.mkpath
-    args = []
-    args << "USE_X265=y" if build.with? "x265"
-    args << "USE_JCTVC=" if build.without? "jctvc"
-    system "make", "install", "prefix=#{prefix}", "CONFIG_APPLE=y", *args
+
+    makeArgs = []
```
mikemcquaid: Call this `args` instead of `makeArgs` to be consistent with other formulae.

Library/Formula/libbpg.rb
```diff
@@ -11,24 +12,38 @@ class Libbpg < Formula
...
+    makeArgs << "install"
+    makeArgs << "prefix=#{prefix}"
+    makeArgs << "CONFIG_APPLE=y"
```
mikemcquaid: As these aren't conditional they can be left on the `system` line like before.

Library/Formula/libbpg.rb
bilhackmac: As I say above, it's one of major change between version `0.9.5` and `0.9.6`.
With `0.9.5`, `jctvc ` was the default encoder and `x265` is optional (depends on `x265` formula).
With `0.9.6`, `x265` is the default encoder (do not depends on `x265` formula anymore) and `jctvc` is optional.

Library/Formula/libbpg.rb
```diff
@@ -11,24 +11,27 @@ class Libbpg < Formula
...
+    args << "USE_JCTVC=y" if build.with? "jctvc"
+    args << "USE_X265=" if build.without? "x265"
+
     system "make", "install", "prefix=#{prefix}", "CONFIG_APPLE=y", *args
   end
 
   test do
     system "#{bin}/bpgenc", test_fixtures("test.png")
+    # Unable to test 1x1 jpg
+    #system "#{bin}/bpgenc", test_fixtures("test.jpg")
```
mikemcquaid: Why not? If so, please remove the comment.

Library/Formula/libbpg.rb
mikemcquaid: Ok, thanks.

Library/Formula/libbpg.rb
bilhackmac: I don't know why… It's work perfectly with a 8x8 jpg, but it's not available in fixtures.

mikemcquaid:

For future reference the preferred commit message format is `libbpg: make x265 the default encoder.`.

bilhackmac:
You're welcome, and thanks for the advices

PR #46598, opened 2015-12-02
docker-machine 0.5.2

davidolrik:


DomT4:
Looks like there's potentially a new dep or something?

```
cmd/machine.go:9:2: cannot find package "github.com/codegangsta/cli" in any of:
	/usr/local/Cellar/go/1.5.1/libexec/src/github.com/codegangsta/cli (from $GOROOT)
	/private/tmp/docker-machine20151202-90860-13ip54z/machine-0.5.2/src/github.com/codegangsta/cli (from $GOPATH)
```

davidolrik:
No the dependencies are all there in the dist tar in a vendor directory, but I can't get go to pick them up.

I tried setting GO15VENDOREXPERIMENT=1 to no avail.

DomT4:
Hmm. Is this a new change in `0.5.2` or was it there for `0.5.1` as well?

davidolrik:
An included vendor directory was also in `0.5.1`, but `github.com/codegangsta/cli` is new in `0.5.2`.

davidolrik:
I symlinked my way out of the missing dependency, but I now get this error when trying to build it:

```
[djo@marklar machine-0.5.2] $ make build
go build -o /Users/davidolrik/tmp/machine-0.5.2/bin/docker-machine  -tags "" -ldflags "-X `go list ./version`.GitCommit=`git rev-parse --short HEAD 2>/dev/null` -w -s"  cmd/machine.go
# command-line-arguments
cmd/machine.go:76: cannot use commands.Commands (type []"github.com/docker/machine/vendor/github.com/codegangsta/cli".Command) as type []"github.com/codegangsta/cli".Command in assignment
make: *** [/Users/davidolrik/tmp/machine-0.5.2/bin/docker-machine] Error 2
```

davidolrik:
The build needs to be performed from within `src/github.com/docker/machine`, as we are basically faking [Local Go 1.5 development environment](https://github.com/docker/machine/blob/master/CONTRIBUTING.md#local-go-15-development-environment) to do the build.

Library/Formula/docker-machine.rb
```diff
@@ -22,7 +22,9 @@ def install
     mkdir_p buildpath/"src/github.com/docker/"
     ln_sf buildpath, buildpath/"src/github.com/docker/machine"
 
+    ENV["PWD"] = buildpath/"src/github.com/docker/machine/"
```
mikemcquaid: Change directory to here instead with a `cd "src/github.com/docker/machine/" do` block.

Library/Formula/docker-machine.rb
```diff
@@ -22,7 +22,9 @@ def install
...
     system "make", "build"
+
```
mikemcquaid: Please don't add whitespace unnecessarily, thanks.

Library/Formula/docker-machine.rb
davidolrik: I initially did that like so:

```ruby
Dir.chdir(buildpath/"src/github.com/docker/machine/") do
  system "make", "build"
end
```

But is doesn't work as the destination directory is a symlink to the `buildpath`, meaning the current path looks the same to `make` both inside and outside the block.

mikemcquaid: Sounds like it'll need reported and patched upstream to fix this issue; that `PWD` works and `cd` doesn't is pretty much just luck.

davidolrik: Actually it's not luck, $PWD is a shell build in and setting it *will* change your path.

Try this in your shell:

    export PWD=/

But I agree that `Dir.chdir` should not resolve the symlinked path to the absolute path.

davidolrik: Ruby upstream are not likely to fix this, as it has been this way for ages.
`docker-machine` upstream are even less likely to fix it, as this is a problem we created in this formular by creating the symlink from within the extracted tar ball.

What we want to do is this (from the [docs](https://github.com/docker/machine/blob/master/CONTRIBUTING.md#local-go-15-development-environment)):
```bash
mkdir docker-machine
cd docker-machine
export GOPATH="$PWD"
go get github.com/docker/machine
cd src/github.com/docker/machine
```

The formular emulates this by creating a symlink `src/github.com/docker/machine` and cd'ing into it.

Is there a way to manually unpack the tar ball? That would let us create the directory `go get` creates, `cd` into it an build it just as `go get` does without the need for `symlink` shenanigans.

Library/Formula/docker-machine.rb
```diff
@@ -22,9 +22,13 @@ def install
     mkdir_p buildpath/"src/github.com/docker/"
     ln_sf buildpath, buildpath/"src/github.com/docker/machine"
 
-    system "make", "build"
-    bin.install Dir["bin/*"]
+    Dir.chdir(buildpath/"src/github.com/docker/machine/") do
+      # Hack to force ruby chdir into the symlinked path
+      ENV["PWD"] = buildpath/"src/github.com/docker/machine/"
+      system "make", "build"
+    end
```
mikemcquaid: Can you please submit this patch for this  to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

davidolrik: Not sure which upstream you mean I should submit a patch to - this is only a problem because the formular creates a symlink from `buildpath` to `buildpath/"src/github.com/docker/machine"` and then tries to build in there.

Ruby upstream are not likely to fix this, as it has been this way for ages.
`docker-machine` upstream are even less likely to fix it, as this is a problem we created in this formular by creating the symlink from within the extracted tar ball.

What we want to do is this (from the [docs](https://github.com/docker/machine/blob/master/CONTRIBUTING.md#local-go-15-development-environment)):
```bash
mkdir docker-machine
cd docker-machine
export GOPATH="$PWD"
go get github.com/docker/machine
cd src/github.com/docker/machine
```

The formular emulates this by creating a symlink `src/github.com/docker/machine` and cd'ing into it.

Is there a way to manually unpack the tar ball? That would let us create the directory `go get` creates, `cd` into it an build it just as `go get` does without the need for `symlink` shenanigans.

mikemcquaid: You can delete the symlink if necessary. We need to not do a `go get` though; we can't have it downloading stuff in `install`. It should be submitted to `docker-machine` upstream because every other Go package works fine without this hack.

davidolrik: I think the hack is needed here due to the use of the `vendor` dependency feature.

    export GO15VENDOREXPERIMENT=1

`docker-machine` is the *only* Go package in Homebrew, that is using that, and at the same time using `make` for building.

There is one other package, namely `keybase`, using `GO15VENDOREXPERIMENT`, but it uses `go build`.

mikemcquaid: I'm happy to tweak anything else in this formula you think we should. Go wild :grinning:

fiunchinho:
Any news on this one? Thanks!

mikemcquaid:
@fiunchinho Waiting on comments to be addressed. Feel free to open your own PR if you'd rather?

PR #46605, opened 2015-12-03
postgresql 9.5beta2 (devel)

DocOtak:


Library/Formula/postgresql.rb
```diff
@@ -1,7 +1,6 @@
 class Postgresql < Formula
   desc "Object-relational database system"
   homepage "https://www.postgresql.org/"
-  revision 2
```
DomT4: The revision doesn't get removed on `devel` version updates, only stable ones.

DomT4: I know the documentation merged the other day isn't explicit about that; I'll fix that in the morning.

DocOtak:
I've updated the pull to only change the devel source/sha.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/32e08d49dd971d9ea4fa2cf9b1ec62def46323f7. Thank you for your contribution to Homebrew @DocOtak; we appreciate it! :smiley_cat:

PR #46606, opened 2015-12-03
bash-preexec 0.2.2 (new formula)

shpoont:
preexec and precmd functions for Bash just like Zsh

### DIFF
--- a/Library/Formula/bash-preexec.rb
+++ b/Library/Formula/bash-preexec.rb
@@ -1,0 +1,27 @@
+class BashPreexec < Formula
+  desc "preexec and precmd functions for Bash just like Zsh."
+  homepage "https://github.com/rcaloras/bash-preexec"
+  url "https://github.com/rcaloras/bash-preexec/archive/0.2.2.tar.gz"
+  sha256 "68e748308754437bad0f948cde8dd3a903ad53d919b4e0ce7b18e012895e6fa8"
+
+  head "https://github.com/rcaloras/bash-preexec.git"
+
+  bottle :unneeded
+
+  def install
+    (prefix/"etc/profile.d").install "bash-preexec.sh"
+  end
+
+  def caveats; <<-EOS.undent
+    Add the following line to your bash profile (e.g. ~/.bashrc, ~/.profile, or ~/.bash_profile)
+
+      [[ -f $(brew --prefix)/etc/profile.d/bash-preexec.sh ]] && . $(brew --prefix)/etc/profile.d/bash-preexec.sh
+
+    EOS
+  end
+
+  test do
+    # Just testing that the file is installed
+    assert File.exist?("#{prefix}/etc/profile.d/bash-preexec.sh")
+  end
+end


Library/Formula/bash-preexec.rb
```diff
@@ -0,0 +1,22 @@
...
+  head "https://github.com/rcaloras/bash-preexec.git"
+
+  bottle :unneeded
+
+  def install
+    (prefix/"etc/profile.d").install "bash-preexec.sh"
+  end
+
+  def caveats; <<-EOS.undent
+    Add the following line to your bash profile (e.g. ~/.bashrc, ~/.profile, or ~/.bash_profile)
```
DomT4: You can actually use `#{shell_profile}` here and it'll output the appropriate configuration file.

Library/Formula/bash-preexec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+        [[ -f $(brew --prefix)/etc/profile.d/bash-preexec.sh ]] && source $(brew --prefix)/etc/profile.d/bash-preexec.sh
```
DomT4: Indent looks a bit too much here? Should be two spaces after the start of the above line rather than four.

Library/Formula/bash-preexec.rb
shpoont: This tool is suitable for bash only

Library/Formula/bash-preexec.rb
shpoont: Fixed

Library/Formula/bash-preexec.rb
DomT4: Apologies - I glanced over the line and completely missed they were all bash related configuration files.

Library/Formula/bash-preexec.rb
```diff
@@ -0,0 +1,27 @@
...
+    Add the following line to your bash profile (e.g. ~/.bashrc, ~/.profile, or ~/.bash_profile)
+
+      [[ -f $(brew --prefix)/etc/profile.d/bash-preexec.sh ]] && source $(brew --prefix)/etc/profile.d/bash-preexec.sh
+
+    EOS
+  end
+
+  test do
+    # Just testing that the file is installed
+    shell_output("[[ -f #{prefix}/etc/profile.d/bash-preexec.sh ]]")
```
DomT4: If we're just doing a file check here let's do `assert File.exist?("#{prefix}/etc/profile.d/bash-preexec.sh")`. It'll break a bit more obviously if something is wrong.

shpoont: Yes, you are right! Fixed.

Library/Formula/bash-preexec.rb
```diff
@@ -0,0 +1,27 @@
...
+  bottle :unneeded
+
+  def install
+    (prefix/"etc/profile.d").install "bash-preexec.sh"
+  end
+
+  def caveats; <<-EOS.undent
+    Add the following line to your bash profile (e.g. ~/.bashrc, ~/.profile, or ~/.bash_profile)
+
+      [[ -f $(brew --prefix)/etc/profile.d/bash-preexec.sh ]] && source $(brew --prefix)/etc/profile.d/bash-preexec.sh
```
mikemcquaid: Can do `.` instead of `source`, right?

shpoont: You can, they are exactly the same.

shpoont: You want me to change it?

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/3c81f48b174375db5cd50a7669305bcefe429aad. Thank you for your contribution to Homebrew @shpoont; we appreciate it! :smiley_cat:

shpoont:
np :)

PR #46608, opened 2015-12-03
metabase 0.13.0 (new formula)

tlrobinson:
This commit contains a sort of hacky way of getting `npm` on the `PATH`, which is required for building from source (only for `--HEAD`), otherwise we get `./bin/build: line 22: npm: command not found`. If there's a better way I'd love to know about it.

### DIFF
--- a/Library/Formula/metabase.rb
+++ b/Library/Formula/metabase.rb
@@ -1,0 +1,61 @@
+class Metabase < Formula
+  desc "Business intelligence report server"
+  homepage "http://www.metabase.com/"
+  url "http://downloads.metabase.com/v0.12.1/metabase.jar"
+  version "0.12.1"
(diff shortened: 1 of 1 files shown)


Library/Formula/metabase.rb
```diff
@@ -0,0 +1,63 @@
...
+    depends_on "leiningen" => :build
+  end
+
+  bottle :unneeded
+
+  depends_on :java => "1.7+"
+
+  def install
+    if build.head?
+      # HACK: get npm on the PATH
```
DomT4: We already have a mechanism for this:

```ruby
ENV.prepend_path "PATH", "#{Formula["node"].opt_libexec}/npm/bin"
```

tlrobinson: Perfect! That's what I was hoping for.

mikemcquaid:

Issue #46553, opened 2015-12-01
Issue with `:mpi` dependency


dpo:
This may be related to #46461 or not, but there is another issue with the `:mpi` dependency:
```
$ brew uninstall open-mpi
Uninstalling /usr/local/Cellar/open-mpi/1.10.1_1... (783 files, 22M)
$ brew uninstall mpich
Error: No such keg: /usr/local/Cellar/mpich
$ brew install mpich
==> Downloading https://homebrew.bintray.com/bottles/mpich-3.1.4_1.mavericks.bottle.tar.gz
Already downloaded: /Library/Caches/Homebrew/mpich-3.1.4_1.mavericks.bottle.tar.gz
==> Pouring mpich-3.1.4_1.mavericks.bottle.tar.gz
🍺  /usr/local/Cellar/mpich/3.1.4_1: 850 files, 16M

$ brew install -v parmetis  # a formula that depends_on :mpi
==> Installing parmetis from homebrew/homebrew-science
==> Installing dependencies for homebrew/science/parmetis: open-mpi
==> Installing homebrew/science/parmetis dependency: open-mpi
Error: Cannot install open-mpi because conflicting formulae are installed.

  mpich: because both install mpi__ compiler wrappers

Please `brew unlink mpich` before continuing.

Unlinking removes a formula's symlinks from /usr/local. You can
link the formula again after the install finishes. You can --force this
install, but the build may fail or cause obscure side-effects in the
resulting software.
```
So `mpich` is no longer recognized as satisfying the `:mpi` dependency.

dpo:
ps: This is independent of whether `mpich` was installed from a bottle or built from source.

mikemcquaid:
Could you try sticking some debugging code in https://github.com/Homebrew/homebrew/blob/f7f8243b6bfc4cdc835be7f24253610befc355ea/Library/Homebrew/requirements/mpi_requirement.rb#L40-L51 and see which part is failing to match.

dpo:
I had already done that, and the strange thing is that `satisfy` returns `true`. I'm trying to trace the execution with a debugger, but not familiar enough with the core of Homebrew to know what's going on. However, here's a potential clue. If I change the `:mpi` dependency to a `:build` dependency, all is well.

UniqMartin:
[...]

PR #46618, opened 2015-12-03
Overhaul, simplify and cleanup documentation.

mikemcquaid:
Remove duplication, link to the API documentation more often, tweak wording, add `@UniqMartin` as a maintainer, note `@jacknagel`'s and `@adamv`'s significant past contributions to Homebrew, delete some outdated or unneeded documentation, add some missing `Formula` API` entries and simplify/improve `CONTRIBUTING.md`.

Closes #46179.


bfontaine:
:metal:

.gitignore
```diff
@@ -11,6 +11,10 @@
 !/bin
 /bin/*
 !/bin/brew
+!/share
+/share/*
+!/share/doc
+/share/doc/*
```
xu-cheng: Why this is necessary? If it's indeed required, you will have to update corresponding  logic in uninstall script: https://github.com/Homebrew/install/blob/master/uninstall#L107-L118

mikemcquaid: Try to make a change to something in `share/doc/homebrew` and `git add` it and see if it needs `-f`. If it doesn't: could just me my setup that's weird.

share/doc/homebrew/Formula-Cookbook.md
```diff
@@ -1,101 +1,50 @@
...
-`$HOMEBREW_REPOSITORY/Library/Formula/foo.rb`
+```shell
+brew create https://example.com/foo-0.1.tar.gz
+```
 
-And opens it in your `$EDITOR`. It'll look like:
+This creates `/usr/local/Library/Formula/foo.rb` and opens it in your `$EDITOR`. It'll look something like:
 
 ```ruby
 class Foo < Formula
```
DomT4: Tiny nit: The example ruby block here is wrong. Should be desc => homepage => url => sha256, not url => homepage => sha256.

share/doc/homebrew/Formula-Cookbook.md
```diff
@@ -107,39 +56,38 @@ class Foo < Formula
 
   def install
     system "./configure", "--prefix=#{prefix}", "--disable-debug", "--disable-dependency-tracking"
-#   system "cmake", ".", *std_cmake_args
+    # system "cmake", ".", *std_cmake_args
```
DomT4: Are we agreed on this now? We've been going back and forth on whether or not to do `"cmake", *std_cmake_args` or `"cmake", ".", *std_cmake_args` for in-tree builds for a while.

share/doc/homebrew/Formula-Cookbook.md
```diff
@@ -107,39 +56,38 @@ class Foo < Formula
...
+## Fill in the [`homepage`](http://www.rubydoc.info/github/Homebrew/homebrew/master/Formula#homepage%3D-class_method)
 
-**We don’t accept formulae without homepages!**
+**We don’t accept formulae without [`homepage`](http://www.rubydoc.info/github/Homebrew/homebrew/master/Formula#homepage%3D-class_method)s!**
 
-SSL/TLS (https) homepage is preferred, if one is available.
+A SSL/TLS (https) [`homepage`](http://www.rubydoc.info/github/Homebrew/homebrew/master/Formula#homepage%3D-class_method) is preferred, if one is available.
 
-Homebrew now has a description field (`desc`). Try and summarize this from the homepage.
+Try to summarize from the [`homepage`](http://www.rubydoc.info/github/Homebrew/homebrew/master/Formula#homepage%3D-class_method) what the formula does in the [`desc`](http://www.rubydoc.info/github/Homebrew/homebrew/master/Formula#desc%3D-class_method)ription.
```
DomT4: Maybe worth mentioning the `desc` automatically outputs the formula name and consequently it doesn't need including? Dunno. It's audit flagged at least if we omit it here.

share/doc/homebrew/Formula-Cookbook.md
mikemcquaid: I don't mind either way, actually. I'll just update this to whatever `brew create` does.

share/doc/homebrew/Formula-Cookbook.md
mikemcquaid: Aye, good idea.

share/doc/homebrew/Formula-Cookbook.md
DomT4: `brew create` currently outputs with the dot included.

share/doc/homebrew/Formula-Cookbook.md
```diff
@@ -588,24 +507,11 @@ class Foo < Formula
...
-
-```ruby
-devel do
-  url "https://foo.com/foo-0.1.tar.gz"
-  sha256 "85cc828a96735bdafcf29eb6291ca91bac846579bcef7308536e0c875d6c81d7"
-end
-```
-
-You can test if the "devel" spec is in use with `build.devel?`.
+Formulae that only have [`head`](http://www.rubydoc.info/github/Homebrew/homebrew/master/Formula#head-class_method) versions should be submitted to [homebrew/headonly](https://github.com/Homebrew/homebrew-headonly) instead of Homebrew/homebrew.
```
DomT4: Guess we should kill this block now with the ongoing deprecation of head-only?

share/man/man1/brew.1
```diff
@@ -554,10 +554,6 @@ HOMEBREW_BROWSER
 If set, uses this setting as the browser when opening project homepages, instead of the OS default browser\.
 .
 .TP
-HOMEBREW_BUILD_FROM_SOURCE
```
DomT4: Are we officially deprecating this option entirely, just the variable, or just no longer documenting/encouraging it?

CONTRIBUTING.md
```diff
@@ -1,25 +1,24 @@
...
-* [Acceptable Formulae](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Acceptable-Formulae.md#acceptable-formulae)
-* [Ruby Style Guide](https://github.com/styleguide/ruby)
-* [How To Open a Homebrew Pull Request (and get it merged)](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged)
+To contribute a fix to the `foo` formula:
+
+- `brew edit foo` and make edits, leave the [`bottle`](http://www.rubydoc.info/github/Homebrew/homebrew/master/Formula#bottle-class_method) as-is, `brew install foo`, `git commit` with message formatted `foo: fix <insert details>.`, [open a pull request](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged) and fix any failing tests.
+
+To report a security vulnerability:
+
+- [Email security@brew.sh](mailto:security@brew.sh) (a private mailing list) with detailed reproduction instructions.
```
DomT4: It'd be nice to include a link to our GPG key here, to match what's in the README.

share/man/man1/brew.1
mikemcquaid: I'd like if we deprecated it but I can see it being useful. I'd just like to no longer document/encourage it as it's never really A Good Idea.

xu-cheng: I'd like to keep it which is a legitimate feature. I know there are some group of people use it.

mikemcquaid: We can keep the feature, I'd rather we didn't encourage it. I don't think the use-case is valid and we've already had issues from people using it.

chdiza: > I don't think the use-case is valid 

What is not valid about wanting to build from source?

xu-cheng: > I'd rather we didn't encourage it.

Personally, I think we should still document. The only thing should be undocumented should be either private or deprecated. Which neither of them apply to this feature. 

>  I don't think the use-case is valid 

I disagree. I think self brewing is one of base features homebrew should offer. Because there are some good reasons to do this, e.g. people who think highly of security than average users would like to build everything from source. Or users who have bad network connection to binary CDN would also like such feature. 

> we've already had issues from people using it.

If users have issues on simply building from source. Then I think this is a bug to solve rather to hide.

mikemcquaid: Building from source does not get you higher security, it gets you lower. We do not use a sandbox so you are executing arbitrary buildsystem code on your machine as an admin user. Building from source is significantly less reliable and significantly slower as it relies heavily on the user's external environment. If you look at our issues list: they are almost all building from source issues.

Look, I'm not saying we remove `--build-from-source` as a flag. I'm saying telling Homebrew "I never want your binary packages" is not something we should support or document. I've just come back from a week long meeting with maintainers from Debian, MacPorts, FreeBSD, NetBSD and various other Linux distros. Many of them were amazed why we allow the amount of options we do (e.g. build from source, pick your compiler, pick your prefix, optional dependencies, install from HEAD). The more options and entropy: the worse the user experience for the majority of users in favour of a small, vocal minority. That is a bad way of doing software. Obviously anyone is welcome to help out and fix it but I don't think any individuals have enough time to fix this and this is why most software is distributed as in binary form.

xu-cheng: > Building from source does not get you higher security, it gets you lower. We do not use a sandbox so you are executing arbitrary buildsystem code on your machine as an admin user. 

Security comes from many aspects. The code execution problem can be migrated by vm. What I mean here that building from source is more security than using third party binaries is focusing one of the aspects: there is no perfect way to verify whether a binary file is comprised. Even with reproducible bottles is not perfect. 

> The more options and entropy: the worse the user experience for the majority of users in favour of a small, vocal minority. That is a bad way of doing software. 

I think it's naive to think everyone have the same need and we should offer an unify UX for all them. What we should do is making the default UX to favor the majority need. At the same time, we should individuals custom their UX based on their own. And this is including (a) keep such feature unless it's too expensive to maintain (b) well document the optional features (which by nature means it would be used in only small group)

chdiza:
> Many of them were amazed why we allow the amount of options we do (e.g. build from source, pick your compiler, pick your prefix, optional dependencies, install from HEAD)

But those things are the the whole point of Homebrew :smile:  Otherwise Homebrew is just a rewrite of MacPorts in Ruby.

share/man/man1/brew.1
mikemcquaid: > Security comes from many aspects. The code execution problem can be migrated by vm.

We don't provide a VM. I'd be OK with `BUILD_FROM_SOURCE` needing a VM though, but I don't think that's what you're suggesting. :wink:

> What I mean here that building from source is more security than using third party binaries is focusing one of the aspects: there is no perfect way to verify whether a binary file is comprised. Even with reproducible bottles is not perfect.

With reproducible builds: it's as perfect as `sha256` is. If you don't trust our bottle `sha256`: why would you trust our formula files?

> I think it's naive to think everyone have the same need and we should offer an unify UX for all them. What we should do is making the default UX to favor the majority need. At the same time, we should individuals custom their UX based on their own. And this is including (a) keep such feature unless it's too expensive to maintain (b) well document the optional features (which by nature means it would be used in only small group)

Allow customisation of this sort makes software worse. Look through our issue tracker and tell me honestly if all users were using binaries we would have more issues rather than less. I don't mean to pull rank but having worked on Homebrew for a long time now: binary packages have solved more support issues than any other individual feature.

chdiza: > Allow customisation of this sort makes software worse. Look through our issue tracker and tell me honestly if all users were using binaries we would have more issues rather than less.

The "number of issues" is not the only relevant consideration in judging what Homebrew should do.

mikemcquaid: @chdiza It's a strong indication of the quality of your software. No-one likes having issues. If you cannot offer a feature that will work consistently: you should not offer the feature to users but instead let it be installable through some other system (e.g. a tap).

xu-cheng: > With reproducible builds: it's as perfect as sha256 is.

Reproducible builds are not perfect for at least two reasons. (a) no all of bottles can be reproducible. (b) it's not resistance to Ken Thomposn Hack, aka, compiler backdoor. (unless you can make build reproducible cross compiler). Apologies on over sensitive on security related issue. Part of the reason is I study cryptography and security as my PhD research field. 

> Allow customisation of this sort makes software worse. Look through our issue tracker and tell me honestly if all users were using binaries we would have more issues rather than less. I don't mean to pull rank but having worked on Homebrew for a long time now: binary packages have solved more support issues than any other individual feature.

This doesn't contradict to my statement. On the contrary is consistent to my logic. That we should support marjor need as default, but keep custom behavior friendly which will only apply to minority people and come with lower support.

mikemcquaid: Fair enough on the security issue. The summit last week would disagree that not everything can be made reproducible but I'll stay away as it's not important here :wink:.



> That we should support marjor need as default, but keep custom behavior friendly which will only apply to minority people and come with lower support.

Why does it need to be documented in the manpage?

xu-cheng: > Why does it need to be documented in the manpage?

Make it friendly to people who need such custom to find it. It's like that there is a custom option which will have lower support than default one. But if you need it, here it is.  :wink:

mikemcquaid: I'd be OK with leaving it if if we provide a string saying "we do not support this configuration, please do not file issues if you use it" or something similar.

xu-cheng: We can introduce "not recommend" tags to our document.

UniqMartin:
Good discussion so far and I'll make sure to chime in soon … :wink:

> […] add `@UniqMartin` as a maintainer, note `@jacknagel`'s and `@adamv`'s significant past contributions to Homebrew […]

I don't think it's a good thing to conflate these small changes with this major documentation overhaul. IMO, they should have a separate PR or at least a separate commit.

share/man/man1/brew.1
mikemcquaid: Would you be OK with asking people to not file issues and/or saying that they are unsupported?

xu-cheng: It's impossible to stop people filing issues. But we can establish some regulation that when a unsupported feature relate issue is filed, we will:
* if user has PR to fix it, we will review it. 
* if it's extrem easy to fix for us, we will fix it. 
* if it's not easy to fix but be related to more substantial supported feature, it will be treated as normal issue. 
* if non of above applied, we will tag wontfix, close it and encourage user to fix on their own if they really need it.

chdiza: > Would you be OK with asking people to not file issues and/or saying that they are unsupported?

Would such a note only apply to the shellvar? The flag `--build-from-source` shouldn't be unsupported.

share/doc/homebrew/Formula-Cookbook.md
bfontaine: Yeah we should.

share/man/man1/brew.1
mikemcquaid: @chdiza We haven't decided yet but I think it should probably be unsupported unless you're asked to use it by a maintainer.

xu-cheng: FYI, it's at least used by science tap maintainers.

mikemcquaid: What for, specifically?

DomT4: Potentially worth noting is that some stuff does perform better if you compile from source locally, particularly crypto related things. There are also some things we don't bottle features of because can't absolutely guarantee they'll be supported identically across various CPUs, such as `aesni`.

I think there's a legitimate use case for from source builds, but I'm fine with us scrapping the "Here's an option to always build from source unconditionally" notation in the manpage at least.

xu-cheng: If I recall correctly, @dpo once told me that he always builds from source. I don't know why.

dpo: I used to *always* build from source. Now I *often* build from source, either for performance reasons, or to avoid bottling pitfalls such as https://github.com/Homebrew/homebrew/issues/46553. Another situation we often encounter in homebrew-science is that `hdf5` will get bumped to a new version, and a plethora of dependents will need to be bumped in the appropriate order (so the bot doesn't time out and builds don't fail). Being able to build from source is quite crucial.

.gitignore
UniqMartin: I can confirm this observation. With these rules added, `git status` also starts showing untracked files in the documentation directory again.

share/man/man1/brew.1
UniqMartin: For what it's worth, I'd like to keep it documented, but in a separate section that clearly states “If you use this feature, you are mostly on your own. Only use if you are comfortable debugging any arising issues yourself.” or something similar. Generally, I'd prefer if more stuff was documented, e.g. commands targeted at maintainers only, instead of trying to hide stuff from users.

(Of course there's no point in having that documentation if nobody is willing to maintain it, but I think it could be feasible to require at least an update to the man page whenever a flag, command, or environment variable is introduced/changed in core code.)

share/doc/homebrew/Formula-Cookbook.md
```diff
@@ -345,79 +275,60 @@ Thus, if you change the name of the class, you must also rename the file. Filena
...
 * explain the commit throughly
 
-At Homebrew, we like to put the name of the formula up front like so: "foobar 7.3 (new formula)".
+At Homebrew, we like to put the name of the formula up front like so: `foobar 7.3 (new formula)`.
 This may seem crazy short, but you’ll find that forcing yourself to summarise the commit encourages you to be atomic and concise. If you can’t summarise it in 50-80 characters, you’re probably trying to commit two commits as one. For a more thorough explanation, please read Tim Pope’s excellent blog post, [A Note About Git Commit Messages](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
 
-The preferred commit message format for simple version updates is "foobar 7.3".
-
-Ensure you reference any relevant GitHub issue `#12345` in the commit message. Homebrew’s history is the first thing future contributors will look to when trying to understand the current state of formulae they’re interested in.
+The preferred commit message format for simple version updates is `foobar 7.3` and for fixes is `foobar: fix flibble matrix.`.
```
UniqMartin: Judging from the `git log` on `Library/Homebrew` you're more or less the only maintainer to end commit subjects with a full stop. We don't have to regulate commit messages to that level of detail, but maybe it is relevant to how we choose our examples?

share/doc/homebrew/Formula-Cookbook.md
```diff
@@ -588,24 +507,11 @@ class Foo < Formula
...
 
 ## Compiler selection
 
-Sometimes a package fails to build when using a certain compiler. Since recent Xcode no longer includes a GCC compiler, we cannot simply force the use of GCC. Instead, the correct way to declare this is the `fails_with` DSL method. A properly constructed `fails_with` block documents the latest compiler build version known to cause compilation to fail, and the cause of the failure. For example:
+Sometimes a package fails to build when using a certain compiler. Since recent Xcodes no longer include a GCC compiler we cannot simply force the use of GCC. Instead, the correct way to declare this is the [`fails_with` DSL method](http://www.rubydoc.info/github/Homebrew/homebrew/master/Formula#fails_with-class_method). A properly constructed [`fails_with`](http://www.rubydoc.info/github/Homebrew/homebrew/master/Formula#fails_with-class_method) block documents the latest compiler build version known to cause compilation to fail, and the cause of the failure. For example:
```
UniqMartin: Why are we asking people to document the cause like this, if it is never used (`cause` is a no-op)? Wouldn't a code comment be sufficient or are there plans to resurrect `cause` and to make use of it?

UniqMartin:
I spotted several typos and minor quirks while reading the entire diff, but I guess this is not what you are looking for at this stage of the PR. I (mostly) like the direction this is taking and applaud the immense effort you already put into this! :bow:

mikemcquaid:
> I spotted several typos and minor quirks while reading the entire diff, but I guess this is not what you are looking for at this stage of the PR. I (mostly) like the direction this is taking and applaud the immense effort you already put into this!

Please feel free to point out typos and quirks (particularly if they are mine rather than there already).

share/doc/homebrew/Formula-Cookbook.md
```diff
@@ -345,79 +275,60 @@ Thus, if you change the name of the class, you must also rename the file. Filena
...
-*   The keg is symlinked into Homebrew’s prefix
-*   Caveats are displayed
-
+*   Keep merge commits out of the pull request
 
 # Convenience Tools
-
 ## Messaging
 
 Three commands are provided for displaying informational messages to the user:
```
UniqMartin: As discussed elsewhere, let's replace `onoe` with `odie`:

> Why should anyone use `onoe`? Right now the only difference is that it's printed to `stderr` and has a different color. Maybe `onoe` should be removed from the documentation and replaced with `odie`, also noting that it will abort the installation.

The example underneath should be also removed or replaced with a better one. Maybe something like:

```ruby
if build.with?("qt") && build.with("qt5")
  odie "Options --with-qt and --with-qt5 are mutually exclusive."
end
system "make", "install"
```

.gitignore
mikemcquaid: I'd love if our `.gitignore` could be less weird. If anyone feels like doing so ever: feel free.

mikemcquaid:
All coments addressed and conflicts resolved.

.gitignore
xu-cheng: FWIW, `git add -u` works fine for me without this gitignore change.

share/doc/homebrew/Formula-Cookbook.md
xu-cheng: > cause is a no-op

We may want to store such information in the future. Beside, I think it's a nice DSL which can encourage/remind authors to document the cause then normal comment(which feels more optional).

.gitignore
mikemcquaid: What about without the `-u`?

share/doc/homebrew/Formula-Cookbook.md
mikemcquaid: Fixed.

share/doc/homebrew/Formula-Cookbook.md
mikemcquaid: Will leave as-is for now.

share/doc/homebrew/Formula-Cookbook.md
mikemcquaid: Yeh, I don't think the full stop matters particularly either way.

share/doc/homebrew/Formula-Cookbook.md
mikemcquaid: Done.

.gitignore
xu-cheng: Works fine with me as well.

share/doc/homebrew/Formula-Cookbook.md
UniqMartin: … including the error I accidentally embedded into my example (missing `?` after 2nd `with`) and the sentence above still references `onoe`. I don't think it's urgent to fix this, though. If you don't object, I'll try to go through the entire documentation and collect post-overhaul fixes to submit them as a new PR.

mikemcquaid: New PR would be :+1: :heart:

PR #46622, opened 2015-12-03
Update Kettle to 6.0

danieljimenez:


Library/Formula/kettle.rb
```diff
@@ -55,7 +55,6 @@ def plist; <<-EOS.undent
 
   test do
     ENV["_JAVA_OPTIONS"] = "-Duser.home=#{testpath}"
-    system "#{bin}/pdikitchen", "-file=#{libexec}/samples/jobs/Slowly\ Changing\ Dimension/create\ -\ populate\ -\ update\ slowly\ changing\ dimension.kjb", "-level=RowLevel"
```
DomT4: Why is this being removed?

danieljimenez: It no longer passes, even though the install of PDI kettle 6 works perfectly. The problem is with that job they ship with the 6.0 package.

DomT4:
Merged in https://github.com/Homebrew/homebrew/commit/af6b06e8fbfd2179217921d7fdc9c2ab7f7c6daf. Thank you for your contribution to Homebrew @danieljimenez; we appreciate it! :smiley_cat:

PR #46624, opened 2015-12-03
mpi-experiments

zbeekman:
I'm just using some PRs to trigger builds on the test bot to try to debug the problems with OpenMPI. This is not a real PR and **SHOULD NOT** be merged.


DomT4:
I've chucked some tags on it just in case any of us miss the commit message and explanation :smile:.

zbeekman:
I'm gonna close this, I think I've learned as much as I can from these...

PR #46625, opened 2015-12-03
openssl 1.0.2e

alex:


alex:
Not sure if it's avialable on all the mirrors yet, but I assume so long as the primary stays up it's not a big deal if they lag behind?

alex:
Tests appear to be failing, it's not obvious to me why, bug in upstream packaging?

reaperhulk:
The OpenSSL developers are aware that there's a bug in the `test` make target. I'm not clear on when they expect to release a fix though.

DomT4:
Yeah, don't sweat the mirrors. I'll update Homebrew's in a second so it has at least one functioning one if necessary. Ignoring the test failures on the few OpenSSL-using formulae because I know about them already and they aren't directly related to OpenSSL but rather little quirks.

Merged in https://github.com/Homebrew/homebrew/commit/e9660e557609ad475fdbe63250bd64030aae1e98. Thank you for another Homebrew contribution @alex! :smiley_cat:

PR #46630, opened 2015-12-03
ENV: warn_if_universal_binaries_not_supported

rhendric:
Raise an error when attempting to create universal binaries with a GCC
that can't do so.

Without this, Homebrew will just silently create 32-bit only binaries when using a GNU GCC.

(Is `homebrew_cc` the right thing to test here? Is there a different way I should have done this? `ENV` seems like an odd place to put this check but given that `ENV.universal_binary` is already here it seemed minimally disruptive. Very open to feedback.)

### DIFF
--- a/Library/Homebrew/extend/ENV/shared.rb
+++ b/Library/Homebrew/extend/ENV/shared.rb
@@ -306,4 +306,10 @@ module SharedEnvExtension
       end
     end
   end
+
+  def check_for_compiler_universal_support
+    if homebrew_cc =~ GNU_GCC_REGEXP
+      raise "Non-Apple GCC can't build universal binaries"
+    end      
+  end
 end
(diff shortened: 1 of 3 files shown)


Library/Homebrew/extend/ENV/shared.rb
```diff
@@ -306,4 +306,10 @@ def fetch_compiler(value, source)
       end
     end
   end
+
+  def warn_if_universal_binaries_not_supported
```
mikemcquaid: Maybe name this `check_for_compiler_universal_support` as it's throwing an exception rather than warning.

bfontaine:
Ping re Mike’s comment.

rhendric:
Renamed!

mikemcquaid:
:+1:

bfontaine:
Thank you for your contribution to Homebrew @rhendric; we appreciate it! :tada:

PR #46636, opened 2015-12-03
jack formula uses new berkeley-db5 formula

### LINKED ISSUE #46628: jack formula has berkeley DB v6 as a dependency, but requires v5
Any good reason for this?

If not I can send a pull request to update the jack formula and make a new one for berkeley-db5

hellominotaur:
I want to resolve #46628 with these two commits, one being the creation of a new formula 'berkeley-db5', the other being a modification to the jack formula to use it as a dependency.

### DIFF
--- a/Library/Formula/berkeley-db5.rb
+++ b/Library/Formula/berkeley-db5.rb
@@ -1,0 +1,70 @@
+class BerkeleyDb5 < Formula
+  desc "High performance key/value database"
+  homepage "https://www.oracle.com/technology/products/berkeley-db/index.html"
+  url "http://download.oracle.com/berkeley-db/db-5.3.28.tar.gz"
+  sha256 "e0a992d740709892e81f9d93f06daf305cf73fb81b545afe72478043172c3628"
+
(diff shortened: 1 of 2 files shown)


DomT4:
I really don't want a *third* BDB in the core if we can at all help it. The existing two are prone to causing trouble as-is.

If it's not going to be used by anything else in the core and it absolutely needs it, it should be vendored in rather than made globally available via new formula.

hellominotaur:
I'm unsure what you mean by 'vendoring in' in this context, could you explain?

mikemcquaid:
@hellominotaur Use a `resource` to install it directly into `libexec` for the formula that needs it.

hellominotaur:
Well this is embarrassing, about midway through applying Berkeley DB as a resource  it appears as if `jackd` no longer makes the complaint about `libdb-5.3.dylib` not being on my system when installed through master. Closing this PR for now, but I'll do my bit to keep an eye open if anyone else has the same issue with jack, on the offchance I've just done something funny to my system.

PR #46640, opened 2015-12-03
catimg v2.2.1

posva:
Hi,

I created a pull request #32209 one year ago when the project was too young. Now that catimg has evolved and community has shown some support, I think it would be a great addition to homebrew.

**This is a new formula**

### DIFF
--- a/Library/Formula/catimg.rb
+++ b/Library/Formula/catimg.rb
@@ -1,0 +1,18 @@
+class Catimg < Formula
+  desc "Insanely fast image printing in your terminal"
+  homepage "https://github.com/posva/catimg"
+  url "https://github.com/posva/catimg/archive/v2.2.1.tar.gz"
+  sha256 "eb76d3baeb5e7382d8839d7d4351794166c0b6e8d777ffe0087a3401f907e991"
+  head "https://github.com/posva/catimg.git"
+
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/catimg", test_fixtures("test.png")
+  end
+end


posva:
I amended the commit to fix style errors

Library/Formula/catimg.rb
```diff
@@ -0,0 +1,21 @@
...
+  sha256 "eb76d3baeb5e7382d8839d7d4351794166c0b6e8d777ffe0087a3401f907e991"
+
+  head do
+    url "https://github.com/posva/catimg.git"
+  end
+
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", "-D", "CMAKE_BUILD_TYPE=Release", ".", *std_cmake_args
```
DomT4: We pass `CMAKE_BUILD_TYPE` as part of the `*std_cmake_args`.

Library/Formula/catimg.rb
```diff
@@ -0,0 +1,21 @@
+class Catimg < Formula
+  desc "Insanely fast image printing in your terminal"
+  homepage "https://github.com/posva/catimg"
+  url "https://github.com/posva/catimg/archive/v2.2.1.tar.gz"
+  sha256 "eb76d3baeb5e7382d8839d7d4351794166c0b6e8d777ffe0087a3401f907e991"
+
+  head do
+    url "https://github.com/posva/catimg.git"
```
DomT4: If you don't need any extra dependencies or anything here you're good to just tack `head "https://example.git"` just under the `sha256` rather than in its own block.

posva:
@DomT4 Thanks, is it fine like that?

Library/Formula/catimg.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    system "#{bin}/catimg", "-h"
```
mikemcquaid: Can this test be modified to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

mikemcquaid: i.e. actually print an image

posva:
@mikemcquaid I added an image with base64 and printed it. Tell me if I should add more tests

Library/Formula/catimg.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"file.b").write("iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8/9hAAABKElEQVR4XmP8//8/AyMj4/94BgYGRSBmB2INJMwIlIhHEwPJMwIVNwDFGRnaGf5jVQBURYw44/XrDP/BJuLQwMnJwcDFxYUTM0ZERPxfsYKdAattuJyPZBlLuD47Q70+AwM2V1xnICzOsqECoghkAExDcjIzw3sREYbALi4GUECBCFAAVzaguhSknvE/Q/t/RiKcihIzyF7YsJ6BgRinIocRTH13NzMDy9OnvAzvgaF8EYgZNEGpgAETxDNgBWVAl7Cc/izNoPEZEg7X20kPTMb/19f/Z8SRBjZvZgbHPwsLCwMbGxuYbWCAGpCM1xmuQxISmQHJAgpE5ChEdw2hAGYJ1maGJ1NGIUGSUyRjfPv//xroOZDIjAROvf+v/8cZiMBszshAAAAAxMRgwhvL0coAAAAASUVORK5CYII=")
```
dunn: We have test fixtures that you can use.  Unless you need the PNG to have specific properties, you can do `system "#{bin}/catimg", test_fixtures("test.png")` here instead.

posva:
@dunn Thanks! I used that image instead

apjanke:

posva:
You're very welcome!

PR #46645, opened 2015-12-04
CMake 3.4.1

RandomDSdevel:
This commit updates Homebrew's CMake formula in order to reflect the fact that Kitware has released CMake v3.4.1 to the public.  
<hr />
Something weird seemed to happen with CMake's dependencies during my build attempt, though, so I'm not sure if this is ready to merge or not.  I forgot to log my installation run, though, so I can't provide you guys with any specifics (I _think_ I saw ICU4C and one other thing in the list of problem dependencies, but it all flashed by so fast that I'm not entirely sure…)


xu-cheng:

PR #46646, opened 2015-12-04
eg 1.01 (new formula)

vszakats:


Library/Formula/eg.rb
```diff
@@ -0,0 +1,83 @@
+class Eg < Formula
+  desc "Expert Guide. Norton Guide Reader For GNU/Linux"
+  homepage "http://www.davep.org/norton-guides/"
+  url "http://www.davep.org/norton-guides/eg-1.00.tar.gz"
+  sha256 "e985b2abd160c5f65bea661de800f0a83f0bfbaca54e5cbdc2e738dfbbdb164e"
+
+  depends_on "s-lang"
+
+  patch :DATA
```
DomT4: Need an upstream submission to link to here and a wee explanation of what is being fixed, Thanks!

Library/Formula/eg.rb
```diff
@@ -0,0 +1,83 @@
...
+      egdraw.c
+      eggetfld.c
+      eghelp.c
+      eglib.c
+      egmenu.c
+      egnavgte.c
+      egregex.c
+      egscreen.c
+      egsigs.c], "<slang/slang.h>", "<slang.h>"
+    inreplace "eglib.c", "/usr/share/", "#{HOMEBREW_PREFIX}/share/"
```
DomT4: Likewise, this is something we should be able to set in the `make` stage rather than inreplacing every time, so ideally poke upstream on this as well.

Library/Formula/eg.rb
vszakats: This is a tool that hasn't been updated in the last 15 years, so before we ping the author, do you have an idea why `clang`/`gcc` doesn't want to compile those escaped newlines in the original source? Granted it's quite a peculiar syntax, but it _should be_ valid C.

```
gcc -Wall -s -O2   -c -o eg.o eg.c
eg.c:177:13: error: expected ')'
            "\
            ^
eg.c:176:11: note: to match this '('
    printf( "Expert Guide Version " EG_VERSION "\n\n"
          ^
eg.c:181:37: error: use of undeclared identifier 'you'
     This program is free software; you can redistribute it and/or modify
                                    ^
eg.c:183:36: error: use of undeclared identifier 'either'
     the Free Software Foundation; either version 2 of the license, or.
                                   ^
eg.c:187:32: error: use of undeclared identifier 'without'
     but WITHOUT ANY WARRANTY; without even the implied warranty of
                               ^
eg.c:192:34: error: expected '(' after 'if'
     along with this program; if not, write to the Free Software
                                 ^
5 errors generated.
make: *** [eg.o] Error 1
```

vszakats:
Hi Dave (@davep),

In trying to bring your `eg` tool to OS X via Homebrew package manager, I've stumbled into two minor issues:
- S-Lang headers need to be referred to as <slang.h> vs. <slang/slang.h> in current code. Looking up the s-lang documentation, it appears that the former might be the canonical method.
- The escaped newline to form a multiline string seems to be breaking `gcc`, which is just a wrapper for `clang` since a few OS iterations. Would you mind adopting the patch attached to this PR for a future `eg` release?

Aside of these technicalities, which are already fixed in this PR, the tools works nicely still.

davep:
I'm not entirely sure what it is you're needing me to do (although as someone who recently acquired an iMac and has started using `homebrew` I'm happy to help get `eg` working on OS X if possible).

Library/Formula/eg.rb
mikemcquaid: > This is a tool that hasn't been updated in the last 15 years

In that case we're unlikely to accept this.

vszakats:
Thanks for jumping and happy that you got a Mac :) @DomT4 as a Homebrew maintainer asked me above to escalate these patches to you (aka "upstream"). The goal is to avoid maintaining patches for 3rd party code inside Homebrew formulae for indefinite periods. Ideally, and if deemed valid, it'd be nice if these modifications could be incorporated into a future release of your code.

Library/Formula/eg.rb
vszakats: Explanations added.

davep:
I've pulled the code into github at https://github.com/davep/eg and have started tidying it up. Currently going so far as compiling without too many issues but linking is failing right now (likely an environment thing -- it simply can't find the S-Lang library, but I need to get back to other things for now).

vszakats:
Great! I'm going to add a `head` to the formula. To link S-Lang you'll need to add the option `-L/usr/local/lib` (and have `s-lang` brew formula installed of course). You may also need to remove the space between `-o` and `eg` in the link command. The Homebrew build system does some clever magic here, so none of those tricks are needed when building under the brew environment.

Library/Formula/eg.rb
```diff
@@ -0,0 +1,92 @@
...
+  url "http://www.davep.org/norton-guides/eg-1.00.tar.gz"
+  sha256 "e985b2abd160c5f65bea661de800f0a83f0bfbaca54e5cbdc2e738dfbbdb164e"
+  head "https://github.com/davep/eg.git"
+
+  depends_on "s-lang"
+
+  # clang complains about escaped newlines used in the source
+  # to form a multiline string. The patch converts said string
+  # to a more common C syntax.
+  patch :DATA
```
mikemcquaid: Can you please submit this patch to the upstream developers of this project and add a link to the upstream patch submission and explanation of why the patch is needed in a comment in the formula file. Thanks!

vszakats: @mikemcquaid We're in talks (under this PR) with @davep, the author of this tool to resolve this.

Do you need more explanation/reasoning for the patch than I've already added, or would that be sufficient? I can think of copying the exact compiler error, if that's desired.

mikemcquaid: Yeh, I saw, that's cool. Were he not around I'd have wanted to see an actual upstream submission link.

vszakats: @mikemcquaid @DomT4 In fact Dave kindly fixed both issues, and some more already in upstream `head`:
- https://github.com/davep/eg/commit/f724fd6b402544f38b06f5a7f3f934dc77807899
- https://github.com/davep/eg/commit/2c1463accba1babb822f9093c5ad3f838d706960

Library/Formula/eg.rb
```diff
@@ -0,0 +1,96 @@
...
+  sha256 "e985b2abd160c5f65bea661de800f0a83f0bfbaca54e5cbdc2e738dfbbdb164e"
+  head "https://github.com/davep/eg.git"
+
+  depends_on "s-lang"
+
+  # clang complains about escaped newlines used in the source
+  # to form a multiline string. The patch converts said string
+  # to a more common C syntax.
+  # Upstream patch:
+  #   https://github.com/davep/eg/commit/2c1463accba1babb822f9093c5ad3f838d706960
```
mikemcquaid: Nice to use `https://github.com/davep/eg/commit/2c1463.patch` if it works?

vszakats: Yes (it should), I'll be testing it tomorrow.

vszakats:
@mikemcquaid @DomT4 Could you give me some guidance on how to copy `default-guide/eg.ng` to `#{HOMEBREW_PREFIX}/share/norton-guides/`? (or rather copying it to Cellar then symlink it to the final target).

davep:
I'm not sure exactly how `homebrew` works in terms of making packages (something I should fix) but, in case it's important/helpful, `eg` will now always live at https://github.com/davep/eg -- I've also made a new release and the archive file (which I saw mentioned in the build file here somewhere) is now https://github.com/davep/eg/archive/eg-v1.01.tar.gz

Hope this helps.

vszakats:
@davep It definitely does, thank you very much! I'm going to update this PR to use 1.01 and the new URLs, this also means we won't need those patches in the formula.

vszakats:
These audit failures hit:
```
```
I hope an exception can be made here by maintainers from these GitHub repo specific rules, given that it's a project with a long past that just has been moved to GitHub from it former home.

bfontaine:
> Could you give me some guidance on how to copy `default-guide/eg.ng` to `#{HOMEBREW_PREFIX}/share/norton-guides/`? (or rather copying it to Cellar then symlink it to the final target).

Is there any reason it needs to be there instead of `#{share}/norton-guides/`?

> I hope an exception can be made here by maintainers from these GitHub repo specific rules, given that it's a project with a long past that just has been *moved* to GitHub from it former home.

Yep.

Library/Formula/eg.rb
```diff
@@ -0,0 +1,25 @@
...
+  sha256 "bb7a2af37c8d5d07f785466692f21561257ff99106d2cb91db13ba2e946ff13b"
+  head "https://github.com/davep/eg.git"
+
+  depends_on "s-lang"
+
+  def install
+    inreplace "eglib.c", "/usr/share/", "#{HOMEBREW_PREFIX}/share/"
+    system "make"
+    bin.install "eg"
+    man1.install "eg.1"
```
bfontaine: Does `make install` not work here?

vszakats:
@bfontaine Only the fact that I wasn't aware of `#{share}` :) of course it should be there then.

Library/Formula/eg.rb
```diff
@@ -0,0 +1,25 @@
...
+    # TODO: copy 'default-guide/eg.ng' to '#{HOMEBREW_PREFIX}/share/norton-guides/'
+  end
+
+  test do
+    # It will return a non-zero exit code when called with any option
+    # except a filename, but will return success even if the file
+    # doesn't exist, and we're exploiting this here.
+    ENV["TERM"] = "xterm"
+    system "eg", "not_here.ng"
```
bfontaine: Couldn’t we write a dummy `.ng` file or use `default-guide/eg.ng` (once we’ll find a way to install it) here?

Library/Formula/eg.rb
vszakats: It failed when testing locally, but I'm quite new in build testing so maybe it was just me.

vszakats: Let's see anyway.

Library/Formula/eg.rb
vszakats: If we use an existing `.ng`, the UI will pop up and won't exit without user intervention, so I believe we can't.

Library/Formula/eg.rb
vszakats: No dice:
```
==> make install
/usr/bin/install eg /usr/bin
install: /usr/bin/eg: Operation not permitted
make: *** [install] Error 71
```

Library/Formula/eg.rb
vszakats: Clarified the comment in this regard.

vszakats:
@bfontaine Cheers for the exception!

bfontaine:
@vszakats You can find all those variables here: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#variables-for-directory-locations

Library/Formula/eg.rb
bfontaine: Looking at the Makefile the following might work:

```rb
bin.mkpath
man1.mkpath
system "make", "install", "BINDIR=#{bin}", "MANDIR=#{man}", "NGDIR=#{share}/norton-guides"
```

vszakats:
Yes, thank you, I've seen it, and I can now remember why I discarded the idea: `#{share}` is pointing inside the Cellar, while `eg` will look for `.ng` files inside `/usr/local/share/norton-guides`, which is patched to be `#{HOMEBREW_PREFIX}/share/` using an `inreplace`. It can be patched otherwise, I'm unsure what is the best option here.

bfontaine:
Yeah we need to find a location outside of the Cellar so that files are not lost on version upgrades. Maybe `#{etc}/norton-guides`? cc @mikemcquaid for an opinion here.

vszakats:
Updated install steps didn't succeed:
```
==> make install BINDIR=/usr/local/Cellar/eg/1.01/bin MANDIR=/usr/local/Cellar/eg/1.01/share/man NGDIR=/usr/local/Cellar/eg/1.01/share/norton-guides
/usr/bin/install eg /usr/local/Cellar/eg/1.01/bin
install: eg: No such file or directory
make: *** [install] Error 71
```

bfontaine:
You need to do it in two steps:
```rb
system "make"
system "make", "install", # ...
```

vszakats:
Sweet. Committed above and another correction.

Library/Formula/eg.rb
```diff
@@ -0,0 +1,25 @@
...
+    system "make"
+    system "make", "install", "BINDIR=#{bin}", "MANDIR=#{man}", "NGDIR=#{etc}/norton-guides"
+  end
+
+  test do
+    # It will return a non-zero exit code when called with any option
+    # except a filename, but will return success if the file doesn't
+    # exist, without popping into the UI - we're exploiting this here.
+    ENV["TERM"] = "xterm"
+    system "eg", "not_here.ng"
```
DomT4: Please use the full path here, i.e. `system bin/"eg", "not_here.ng"`

vszakats:
Corrected!

mikemcquaid:

For future reference the preferred commit message format for new formula is `eg 1.01 (new formula)` as a single squashed commit.

vszakats:
I'm squashing happily (and also as a routine by now), just didn't know we reached the end of the necessary updates, sorry! :o

vszakats:
(IOW, merge came as a bit of - nice - surprise &mdash; many thanks for it, I'm now going to install `eg` and start using it :)

PR #46647, opened 2015-12-04
libdecnumber

alkis:
libdecnumber 3.68 (new formula)

ANSI C implementation of General Decimal Arithmetic

### DIFF
--- a/Library/Formula/libdecnumber.rb
+++ b/Library/Formula/libdecnumber.rb
@@ -1,0 +1,41 @@
+class Libdecnumber < Formula
+  desc "ANSI C implementation of General Decimal Arithmetic"
+  homepage "http://speleotrove.com/decimal/#decNumber"
+  url "http://download.icu-project.org/files/decNumber/decNumber-icu-368.zip"
+  version "3.68"
+  sha256 "14ec2cf30b58758493a7661b78b80abfb281652b61a425b85cda83173518fe25"
+  resource "cmakelists" do
+    url "https://raw.githubusercontent.com/alkis/decnumber-cmake/master/CMakeLists.txt"
+    sha256 "eb8bb964b5382e5cc9ea6530585f2567b19ffc280d1b78ee2ebb8151c6cb9b9a"
+  end
+
+  depends_on "cmake" => :build
+
+  def install
+    resource("cmakelists").stage { mv("CMakeLists.txt", @buildpath) }
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+
+  test do
+    (testpath/"test.c").write <<-EOS.undent
+      #include "decQuad.h"
+      #include <stdio.h>
+      int main(int argc, char *argv[]) {
+        decQuad a, b;
+        decContext set;
+        char string[DECQUAD_String];
+        decContextTestEndian(0);
+        decContextDefault(&set, DEC_INIT_DECQUAD);
+        decQuadFromString(&a, "1.123", &set);
+        decQuadFromString(&b, "3.321", &set);
+        decQuadAdd(&a, &a, &b, &set);
+        decQuadToString(&a, string);
+        puts(string);
+     }
+    EOS
+    system ENV.cc, "-I#{include}", "-L#{lib}", "-ldecnumber",
+           testpath/"test.c", "-o", testpath/"test"
+    system "./test"
+  end
+end


Library/Formula/libdecnumber.rb
```diff
@@ -0,0 +1,41 @@
+class Libdecnumber < Formula
+  desc "ANSI C implementation of General Decimal Arithmetic"
+  homepage "http://speleotrove.com/decimal/#decNumber"
+  url "http://download.icu-project.org/files/decNumber/decNumber-icu-368.zip"
+  version "3.68"
+  sha256 "14ec2cf30b58758493a7661b78b80abfb281652b61a425b85cda83173518fe25"
+  resource "cmakelists" do
```
DomT4: Why are we having to fetch and use a private `cmakelists`?

alkis: Because the distribution has no Makefile. It is just headers and sources.

DomT4: So what is the CMakeList being used for? What is being installed that couldn't be installed with something like:

```ruby
include.install Dir["*.h"]
```
Or such?

alkis: The CMakeList is being used to build static and shared libraries. It is also used to install the libraries and also the headers. I could install the headers with include.install but that would be useless without the static and shared libraries.

DomT4: We could install those with `lib.install` as well. We're not likely to accept a build process not supported by upstream because if it breaks and we upstream it we'll get slightly perplexed looks. If you want to use `cmake` it should be supported by them.

alkis: There is *no* Makefile in the distribution. How do you suggest I should build the libraries?

DomT4: Manually, if necessary. If upstream are making it this difficult to build it, it's hard not to wonder if they intend it to be packaged by anyone in *this manner* rather than embedded in other sources.

alkis: I don't think I understand what you are trying to achieve here. I wrote the
CMakeLists. That's a script in a language made specifically to build stuff.
Yet you suggest I write the same thing in Ruby? I would likely get that
wrong so i really don't see the advantage.

I am not sure why the developer did not provide a build file. It is likely
because this is a very low level library mostly used by compilers so the
couple of compilers that use it, incorporate the sources in their code and
use their own build system to build it.

DomT4: I'm trying to achieve us being able to support this package without actually becoming the upstream maintainers or adding in unsupported, unknown-by-upstream build steps that if we get bug reports on we'll have to handle ourselves.

You should submit the CMakeList to upstream and try to talk them into using it, rather than working around upstream's lack of convenient build process on a package-manager-by-package-manager basis.

alkis: Ok lets take this one point at a time:

* unsupported, unknown-by-upstream build steps. You suggestion to write this in Ruby does not change anything here, except that we would do these steps in an unsuitable language. In addition I do not agree on the unsupported claim. If there are no build steps whatsoever, any build steps are supported. It seems that you are suggesting the opposite, that is none are supported. So you claim this library is unbuildable and unusable, which is not very constructive.
* lack of convenient build process: upstream provides **no** build process. It offers instructions on how to build the examples in readme.txt and no instructions in how to build a shared or static library. Because of this we as the owners of this formula are in fact the owners of the build process. This means writing our own build scripts. Choosing Ruby over CMake for build scripts is IMO unwise.

I can try to sell the build script upstream but given how infrequently this library is updated (stable since 2010) I wouldn't want to hold my breath for a new release. That also reminds me that in [errata](http://speleotrove.com/decimal/decnumerr.html) there is a fix (in text) for a bug in the latest release. I am sure you will agree that chances of a CMakeLists making it into any release are slim.

alkis: BTW would you be OK if I inline the CMakeLists.txt in the formula itself? Would this address any of your concerns?

mikemcquaid: If this software isn't maintained and doesn't have a buildsystem: it shouldn't be included in Homebrew. If you'd like it to be more included you need to convince upstream to accept a buildsystem.

alkis: The software is stable for >7 years now. It does lack proper release engineering practices but that does not mean it is useless.

It would be nice if you can document this new requirement of "no build system" == "no homebrew" for the benefits of others that might fall in the same trap as I did.

Feel free to close this pull request as I won't be pursuing it further given this new set of requirements.

mikemcquaid: We did: https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Acceptable-Formulae.md#niche-or-self-submitted-stuff

PR #46655, opened 2015-12-04
gnu-units: Fix Python 2 path for units_cur

### LINKED ISSUE #46527: gnu-units: gunits_cur uses wrong Python 2 path
Running `gunits_cur` (or `units_cur` if the `gnu-units` formula was installed with the `--with-default-names` flag) gives the following error:

```
zsh: /usr/local/bin/units_cur: bad interpreter: /usr/bin/python2: no such file or directory
```

The reason appears to be the first line of the script (`less $(where gunits_cur)`), which says:

```zsh
#!/usr/bin/python2
```

But `python2` is located inside `$(brew --prefix)/bin`, i.e. `/usr/local/bin` by default.

fenhl:
Fixes #46527. The `gunits_cur` script still errors for me, but that seems to be an upstream issue.

### DIFF
--- a/Library/Formula/gnu-units.rb
+++ b/Library/Formula/gnu-units.rb
@@ -28,6 +28,11 @@ class GnuUnits < Formula
   patch :DATA
 
   def install
+    # OSX doesn't provide 'python2', but on some Linux distributions
+    # 'python' is an alias for python3 so this won't be changed
+    # upstream
+    inreplace "units_cur", "#!/usr/bin/python2", "#!/usr/bin/env python"
+
     args = ["--prefix=#{prefix}"]
     args << "--program-prefix=g" if build.without? "default-names"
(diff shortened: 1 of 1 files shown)


Library/Formula/gnu-units.rb
```diff
@@ -28,6 +28,11 @@ class GnuUnits < Formula
   patch :DATA
 
   def install
+    # OSX doesn't provide 'python2', but on some Linux distributions
```
DomT4: We can shrink these notes a bit. Something like `OS X does not provide a `python2` executable` or something. It's a relatively well known issue that doesn't need too much explaining.

dunn:
Merged in c721e89cabce46a588c03d2f364ba133632ddb3b, thanks!

Issue #46562, opened 2015-12-02
GNUTLS 3.4.x


rakkesh:
GnuTLS has a new release, `3.4.7`, and `3.4.x` is marked as stable. 3.4.x has API/ABI changes from the previous stable releases `3.3.x`. So, every formula which is dependent on `gnutls` needs to be revised. 
What is the process of making such an update?

By the way, I have successfully built `emacs`, `wireshark` and `qemu` with `gnutls 3.4.7`.

rakkesh:
Formulae from `Homebrew` core which depend on GnuTLS

**Formula: Dependency type**
- [ ] bitlbee: required
- [x] emacs: optional
- [ ] freediameter: required
- [ ] git-annex: required
- [ ] glib-networking: required
- [ ] gloox: optional
- [ ] gnu-smalltalk: required
- [ ] gst-plugins-bad: optional
- [ ] gwenhywfar: required
- [ ] hamsterdb: required
- [ ] inspircd: optional
- [ ] kbtin: required
- [ ] knot: required
- [ ] libav: optional
- [ ] libinfinity: required
- [ ] libsoup: required
- [ ] libvirt: required
- [ ] libxmlsec1: recommended
- [ ] loudmouth: required
- [x] mailutils: required
- [ ] minbif: required
- [ ] ncdc: required
- [ ] net6: required
- [ ] ntopng: build
- [ ] openconnect: required
- [ ] pianobar: required
- [ ] pianod: required
- [ ] pidgin: required
- [ ] profanity: required
- [x] qemu: required
- [ ] task: required
- [ ] taskd: required
- [ ] ucommon: required
- [ ] vpnc: required
- [ ] vte3: required
- [ ] weechat: required
- [x] wireshark: required

_Note: I have built the ones with the check mark._

DomT4:
This could get a bit complicated. To date the new GnuTLS branch has required the new Nettle [branch](https://github.com/Homebrew/homebrew-versions/blob/master/gnutls34.rb#L19) as well, so you'd have to factor in the breaking `nettle` release as well into the checks here.

rakkesh:
Thanks @DomT4. Updated the list in my previous comment.

rakkesh:
`lsh` failed, gist logs - https://gist.github.com/6c666e9b48ffb5d2f2c7

rakkesh:
@DomT4 The check list is complete now. 

Bug reports:
`lsh`: reported bug[1713] on their `bugzilla`.
[...]

PR #46660, opened 2015-12-04
Update nettle to revision 3.1

abique:


DomT4:
Sorry, not going to be able to do this yet. Please track https://github.com/Homebrew/homebrew/issues/46562 for progress.

You can also `brew install homebrew/versions/gnutls34` if you need that urgently, and the same with `nettle3`.

PR #46664, opened 2015-12-04
rubberband 1.8.1 (new formula)

david0:
Originally from http://tuohela.net/irc/rubberband.rb
Modified by mtolly to include libsndfile dependency,
and to install pkg-config file
plus tests, updates and fixes from myself

### DIFF
--- a/Library/Formula/rubberband.rb
+++ b/Library/Formula/rubberband.rb
@@ -1,0 +1,37 @@
+class Rubberband < Formula
+  desc "audio time stretcher tool and library"
+  homepage "http://breakfastquay.com/rubberband/"
+  url "http://code.breakfastquay.com/attachments/download/34/rubberband-1.8.1.tar.bz2"
+  sha256 "ff0c63b0b5ce41f937a8a3bc560f27918c5fe0b90c6bc1cb70829b86ada82b75"
+  head "https://bitbucket.org/breakfastquay/rubberband/", :using => :hg
+
+  stable do
+    # replace vecLib.h by Accelerate.h
+    # already fixed in upstream:
+    # https://bitbucket.org/breakfastquay/rubberband/commits/cb02b7ed1500f0c06c0ffd196921c812dbcf6888
+    # https://bitbucket.org/breakfastquay/rubberband/commits/9e32f693c6122b656a0df63bc77e6a96d6ba213d
+    patch :p1 do
+      url "http://tuohela.net/irc/rubberband-1.8.1-yosemite.diff"
+      sha1 "76ea7cac0fc0ab99b38081176375ef7c34be678f"
+    end
+  end
+
+  depends_on "pkg-config" => :build
+  depends_on "libsamplerate"
+  depends_on "libsndfile"
+
+  def install
+    system "make", "-f", "Makefile.osx"
+    bin.install "bin/rubberband"
+    lib.install "lib/librubberband.dylib"
+    include.install "rubberband"
+
+    cp "rubberband.pc.in", "rubberband.pc"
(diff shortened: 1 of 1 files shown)


Library/Formula/rubberband.rb
```diff
@@ -0,0 +1,52 @@
+# Originally from http://tuohela.net/irc/rubberband.rb
+# Modified by mtolly to include libsndfile dependency,
+# and to install pkg-config file.
```
mikemcquaid: You can remove these comments.

Library/Formula/rubberband.rb
```diff
@@ -0,0 +1,52 @@
...
+  depends_on "pkg-config" => :build
+  depends_on "libsamplerate"
+  depends_on "libsndfile"
+
+  def install
+    system "make", "-f", "Makefile.osx"
+    bin.install "bin/rubberband"
+    lib.install "lib/librubberband.dylib"
+    include.install "rubberband"
+    (lib/"pkgconfig/rubberband.pc").write pc_file
```
mikemcquaid: I don't think we should include this if upstream doesn't provide it.

Library/Formula/rubberband.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  depends_on "pkg-config" => :build
+  depends_on "libsamplerate"
+  depends_on "libsndfile"
+
+  def install
+    system "make", "-f", "Makefile.osx"
+    bin.install "bin/rubberband"
+    lib.install "lib/librubberband.dylib"
+    include.install "rubberband"
```
mikemcquaid: No `make install` we can use?

Library/Formula/rubberband.rb
```diff
@@ -0,0 +1,52 @@
...
+    Version: #{version}
+    Description:
+    Libs: -L${libdir} -lrubberband
+    Cflags: -I${includedir}
+    EOS
+  end
+
+  test do
+    assert_match /Pass 2: Processing.../, shell_output("rubberband -t2 " + test_fixtures("test.wav") + " out.wav 2>&1")
+    rm "out.wav"
```
mikemcquaid: don't need to remove this, it's cleaned up automatically.

Library/Formula/rubberband.rb
```diff
@@ -0,0 +1,52 @@
...
+    Name: rubberband
+    Version: #{version}
+    Description:
+    Libs: -L${libdir} -lrubberband
+    Cflags: -I${includedir}
+    EOS
+  end
+
+  test do
+    assert_match /Pass 2: Processing.../, shell_output("rubberband -t2 " + test_fixtures("test.wav") + " out.wav 2>&1")
```
mikemcquaid: Use `rubberband -t2 #{test_fixtures("test.wav")} out.wav` and don't redirect stderr if you can and instead match on some `stdout`.

Library/Formula/rubberband.rb
david0: It looks like it is provided for linux (autoconf), but not using the OS X makefile. 
But now that I think of it, I could use the included rubberband.pc.in and post-process that. Would that be ok?

Library/Formula/rubberband.rb
david0: No, there is one for linux (autoconf) but none in `Makefile.osx` unfortunately

Library/Formula/rubberband.rb
david0: all the output is on stderr, thats why im redirecting to match it.

mikemcquaid:

PR #46666, opened 2015-12-04
cern-ndiff 5.02.08 (new formula)

Eothred:
The CERN ndiff tool is a specialiced diff tool
for comparing numerical differences in (large) text files.
There are no testing defined for the binary at the moment,
which is why the test is empty in this recipe.
The tool is primarily used for testing of the Mad-X software,
but generally useful for people that need to do advanced numerical
difference comparison.

### DIFF
--- a/Library/Formula/cern-ndiff.rb
+++ b/Library/Formula/cern-ndiff.rb
@@ -13,4 +13,11 @@ class CernNdiff < Formula
     system "cmake", ".", *std_cmake_args
     system "make", "install"
   end
+
+  test do
+    (testpath/"lhs.txt").write("0.0 2e-3 0.003")
+    (testpath/"rhs.txt").write("1e-7 0.002 0.003")
+    (testpath/"test.cfg").write("*   * abs=1e-6")
+    system "#{bin}/numdiff", "lhs.txt", "rhs.txt", "test.cfg"
+  end
 end
--- a/Library/Formula/cern-ndiff.rb
+++ b/Library/Formula/cern-ndiff.rb
@@ -1,4 +1,3 @@
-
 class CernNdiff < Formula
   desc "Numerical diff tool"
   # Note: ndiff is a sub-project of Mad-X at the moment..
--- a/Library/Formula/cern-ndiff.rb
+++ b/Library/Formula/cern-ndiff.rb
@@ -1,0 +1,16 @@
+
+class CernNdiff < Formula
+  desc "Numerical diff tool"
+  # Note: ndiff is a sub-project of Mad-X at the moment..
+  homepage "http://cern.ch/mad"
+  url "http://svn.cern.ch/guest/madx/tags/5.02.08/madX/tools/numdiff"
+  version "5.02.08"
+  head "http://svn.cern.ch/guest/madx/trunk/madX/tools/numdiff"
+
+  depends_on "cmake" => :build
+
+  def install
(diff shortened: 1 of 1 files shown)


Library/Formula/cern-ndiff.rb
```diff
@@ -0,0 +1,16 @@
+
```
mikemcquaid: Remove this empty line.

Library/Formula/cern-ndiff.rb
```diff
@@ -0,0 +1,16 @@
...
+class CernNdiff < Formula
+  desc "Numerical diff tool"
+  # Note: ndiff is a sub-project of Mad-X at the moment..
+  homepage "http://cern.ch/mad"
```
mikemcquaid: Need a homepage that refers to the tool.

Library/Formula/cern-ndiff.rb
```diff
@@ -0,0 +1,16 @@
...
+  version "5.02.08"
+  head "http://svn.cern.ch/guest/madx/trunk/madX/tools/numdiff"
+
+  depends_on "cmake" => :build
+
+  def install
+    system "cmake", ".", *std_cmake_args
+    system "make", "install"
+  end
+end
```
mikemcquaid: Can a test be added to do something more substantial than e.g. `--version` or `--help`? See `cmake.rb` for an example of an application formula with a good test and `tinyxml2.rb` for an example of a library formula with a good test. Thanks!

dunn:
Should it be `cern-numdiff.rb`? From the URL it looks like the software is 'numdiff' rather than 'ndiff'.

Eothred:
Hi,

The tool has the official name 'ndiff'. It was previously called 'numdiff' but renamed 'ndiff' because other software had already used the name 'numdiff' I think. Yes, the binary is still called 'numdiff'.

The manual is here (but pdf only accessible to CERN users unfortunately): http://cds.cern.ch/record/1574594
I have asked the author about why it is restricted. He has previously stated the code is under GPL so I am not sure why they would limit access to the documentation.

The ndiff tool is mentioned in the "resources" tab on the homepage currently given: http://mad.web.cern.ch/mad/www/resources.html
But it does not have a separate homepage yet. I still thought the 'Mad-X' was the most appropriate homepage to define, since it does not have it's separate homepage yet.

Eothred:
Previously I thought the test part of the ruby script should only be used in case a test suite was available (e.g. ctest). I now understood that you basically just want to check against basic segfaults/missing libraries etc, so I updated with a very basic one line comparison.

Thanks for giving me quick feedback btw, appreciate it! :)

Eothred:
Aah, the ndiff manual is accessible on the homepage if you click on the 'documentation tab', I did not notice that before. [direct link](http://cern.ch/madx/madX/tools/numdiff/doc/CERN-ACC-NOTE-2013-0005.pdf)

That same manual is also available in the svn repository that is downloaded, under doc/CERN-ACC-NOTE-2013-0005.pdf

Library/Formula/cern-ndiff.rb
Eothred: Done, thanks for pointing it out.

bfontaine:
Thank you @Eothred for your contribution to Homebrew; we appreciate it! :tada:

bfontaine:

PR #41807, opened 2015-07-17
scala 2.12.0-M3 (devel)

AtkinsChang:
scala 2.12.0-M2, java version requirement to 1.8+.

The release note of 2.12.0-M2:
```
Beginning with 2.12.0-M2, the Scala 2.12 series targets Java 8. Programs written in Scala 2.12, including the Scala 2.12 compiler, can only be executed on Java 8 or newer.
```

PR #46668, opened 2015-12-04
scala 2.12.0-M3 (devel)

dpoggi:


dpoggi:
Errr... brew test works fine locally, looks like it's missing scala-library from org.scala-lang in the Jenkins environment... but whyyyyy...

mikemcquaid:
Does `brew test --devel --verbose --sandbox scala` pass? You can see there's a sandbox write issue in the test: https://travis-ci.org/Homebrew/homebrew/jobs/94924895

dpoggi:
screen shot 2015-12-04 at 2 24 12 pm

mikemcquaid:
Not sure what's going on, I'm afraid, but we'll need to fix this issue before it can be merge :sob:

dunn:
See also #41807 :crying_cat_face:

DomT4:
Passes here as well, on a machine that has never had `scala` installed before:
```
 ~> brew test -v scala --devel
Testing scala
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20151227-75071-xke1b4.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Formula/scala.rb -v --devel --devel
==> /usr/local/Cellar/scala/2.12.0-M3/bin/scala /private/tmp/scala20151227-75072-9s7z4k/hello.scala
```

I'd maybe try setting `ENV.java_cache` in `test do` here as a complete hunch, but eh.

dpoggi:
Figured this one out. You can replicate the result of the test locally by installing devel, uninstalling it, then installing stable - it will produce the same error. Running a Scala program as a script (`scala Foo.scala`) starts a Scala CompileServer process in the background. When the tests run on Travis, stable is installed, the tests are run, then stable is uninstalled, devel is installed, and the tests are run again. The problem is, the stable version of CompileServer is still running in the background, so when the tests are run for devel, the newer version of the "scala" command-line tool tries to talk to an older version of the CompileServer, and this error happens. Let's see if this fixes it...

dpoggi:
Wooooooo! That's a pass. @mikemcquaid, I think this is finally ready for merge.

DomT4:
Great sleuthing on chasing the error down here!

Merged in https://github.com/Homebrew/homebrew/commit/bc5581d22e62dca4e5d22b2bc024bf3d0d72c385. Thank you for another contribution to Homebrew @dpoggi! :smiley_cat:

PR #44064, opened 2015-09-17
sleuthkit 4.2.0

anarchivist:
* Adds patch where JNI bindings sample compilation fails (see sleuthkit/sleuthkit#510)
* Changes command to build JNI bindings
* Adds test

PR #46672, opened 2015-12-04
sleuthkit 4.2.0

theopolis:
1. Bump sleuthkit to 4.2.0, prefer to ~~build~~ download from Github/source of truth.
2. Add tests and restrict JNI bindings during default build/distribution.

Also, this closes #44064.

### DIFF
--- a/Library/Formula/sleuthkit.rb
+++ b/Library/Formula/sleuthkit.rb
@@ -3,14 +3,12 @@ class Sleuthkit < Formula
   homepage "http://www.sleuthkit.org/"
 
   stable do
-    url "https://downloads.sourceforge.net/project/sleuthkit/sleuthkit/4.1.3/sleuthkit-4.1.3.tar.gz"
-    sha256 "67f9d2a31a8884d58698d6122fc1a1bfa9bf238582bde2b49228ec9b899f0327"
+    url "https://github.com/sleuthkit/sleuthkit/archive/sleuthkit-4.2.0.tar.gz"
+    sha256 "d71414134c9f8ce8e193150dd478c063173ee7f3b01f8a2a5b18c09aaa956ba7"
+  end
 
-    # Upstream fix for https://github.com/sleuthkit/sleuthkit/issues/345
-    patch do
-      url "https://github.com/sleuthkit/sleuthkit/commit/39c62d6d169f8723c821ca7decdb8e124e126782.diff"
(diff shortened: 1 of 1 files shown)


Library/Formula/sleuthkit.rb
```diff
@@ -1,17 +1,9 @@
...
-    sha256 "67f9d2a31a8884d58698d6122fc1a1bfa9bf238582bde2b49228ec9b899f0327"
-
-    # Upstream fix for https://github.com/sleuthkit/sleuthkit/issues/345
-    patch do
-      url "https://github.com/sleuthkit/sleuthkit/commit/39c62d6d169f8723c821ca7decdb8e124e126782.diff"
-      sha256 "f9419d6665a89df5625487dd50c16c12d1680477797917f8ec9182db55df4f7f"
-    end
-  end
+  url "https://github.com/sleuthkit/sleuthkit.git",
+    :tag => "sleuthkit-4.2.0",
```
dunn: Any reason we can't use a tarball?

theopolis: Building from a commit hash IMO is more honest, auditable, and debuggable. That said, I don't hold that opinion very strongly. :)

mikemcquaid: We prefer using a tarball as it's easier to checksum and cache.

theopolis: I'm confused! :confused: 

Isn't the commit hash that you request added alongside `tag` a checksum. Not sure about the caching, in almost all cases folks install using the bottles. When building a bottle locally/from source my clients always pull from the URL listed in the formula.

mikemcquaid: It is a Git checksum which means it's SHA-1 rather than SHA-256 (which is more secure). Additionally it's faster to `curl` a tarball than `git fetch` a repository.

theopolis: Eh, I'm still not convinced a tarball is a better option. Caching != not speed/efficiency of a download for a mostly-exceptional case (building from source); a repo and commit hash feels much better! But now that you mention GIT's binding to sha1, though that feels like a reach, it is nice to start moving everything to sha256.

Anyways, I'll update to a tarball because you prefer it, always happy to oblige.

mikemcquaid: Thanks!

apjanke:

PR #46675, opened 2015-12-04
add mediaconch formula

dericed:
based on the similar media-info formula (which installs mediainfo)


mikemcquaid:

For future reference the preferred commit message format for new formulae is `mediaconch 15.11 (new formula)`.

PR #46676, opened 2015-12-04
man page: finer-grained markup

### LINKED ISSUE #44769: manpage formatting with finer-grained markup
Hi, Homebrew folks.

I took a stab at [reformatting the `brew` manpage](https://github.com/apjanke/homebrew/tree/manpage-formatting) with finer-grained markup to make the syntax more clear, better distinguishing placeholders and syntax structure indicators. E.g. using the ronn <...> markup for "variables", leaving the [ ] | ( ) elements un-marked-up, and making sure all literal text was marked up with \`...\`.

Before:

```
`brew` --version<br>
`brew` command [--verbose|-v] [options] [formula] ...
```

After:

```
`brew` `--version`<br>
`brew` <command> [`--verbose`|`-v`] [<options>] [<formula>] ...
```

This ends up with a nicer-looking and more readable man page output, I think.

Before:
screen shot 2015-10-09 at 3 06 54 am
After:
screen shot 2015-10-09 at 3 06 06 am
Before:
screen shot 2015-10-09 at 3 07 13 am
After:
screen shot 2015-10-09 at 3 06 44 am

But the web page looks worse. The italic proportional and roman monospace don't flow well together.

Before:
screen shot 2015-10-09 at 3 10 52 am
After:
screen shot 2015-10-09 at 3 10 59 am
screen shot 2015-10-09 at 3 11 21 am

Ronn doesn't seem to have a way of indicating a span of "code" that should be set in monospace in the web page output that can also contain bold and underline/italic characters to differentiate literal text and placeholders. Like this, where the italicization of "func" is "nested" inside monospacing to indicate replaceable code.

screen shot 2015-10-09 at 5 39 27 am

Not an issue with the man page output, since everything is monospaced there.

Thoughts on this? Would it be worth reformatting to get better markup in the man page, and let the web version be a bit uglier, since it's a secondary target? Or should we stick with marking up all the code with \`...\` and using internal `<...>` that the users should read as placeholder indicators, instead of being turned in to underlines?

apjanke:
Closes #44769

### DIFF
--- a/Library/Homebrew/manpages/brew.1.md
+++ b/Library/Homebrew/manpages/brew.1.md
@@ -3,8 +3,8 @@ brew(1) -- The missing package manager for OS X
 
 ## SYNOPSIS
 
-`brew` --version<br>
-`brew` command [--verbose|-v] [options] [formula] ...
+`brew` `--version`<br>
+`brew` <command> [`--verbose`|`-v`] [<options>] [<formula>] ...
 
 ## DESCRIPTION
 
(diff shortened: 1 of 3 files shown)


UniqMartin:
:+1:

PR #46677, opened 2015-12-04
vmtouch 1.0.2 (new formula)

maligree:
The previous formula never got in (#9009), so here's 1.0.2.

First time submitting a formula, so I may have missed a thing here or there.

### DIFF
--- a/Library/Formula/vmtouch.rb
+++ b/Library/Formula/vmtouch.rb
@@ -1,0 +1,15 @@
+class Vmtouch < Formula
+  desc "Portable file system cache diagnostics and control"
+  homepage "http://hoytech.com/vmtouch"
+  url "https://github.com/hoytech/vmtouch/archive/vmtouch-1.0.2.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/vmtouch.rb
```diff
@@ -0,0 +1,15 @@
...
+  url "https://github.com/hoytech/vmtouch/archive/vmtouch-1.0.2.tar.gz"
+  sha256 "48d41e5b805a0fcac46c12feeb4650e4b9def44b75d528d7c66c2f8b3a747c39"
+  head "https://github.com/hoytech/vmtouch.git"
+
+  def install
+    system "make", "install", "PREFIX=#{prefix}", "MANDIR=#{man}"
+  end
+
+  test do
+    system "#{bin}/vmtouch", "#{bin}/vmtouch"
```
DomT4: What's the test doing here, out of curiosity?

maligree: In its simplest form, vmtouch reports the cached pages of a file. The test
passes the vmtouch binary as the subject file. If the call doesn't error
out, we seem to have a working install.

mikemcquaid:

PR #46679, opened 2015-12-04
sslscan 1.11.0 (new formula).

dsXLII:
This is my first Homebrew formula, so I expect harsh criticism. Formula adds the 'sslscan' CLI tool for quickly checking SSL/TLS on a remote site. (This is not the same as ssllabs-scan.)

### DIFF
--- a/Library/Formula/sslscan.rb
+++ b/Library/Formula/sslscan.rb
@@ -1,0 +1,19 @@
+class Sslscan < Formula
+  desc "Test SSL/TLS enabled services to discover supported cipher suites."
+  homepage "https://github.com/rbsec/sslscan"
+  url "https://github.com/rbsec/sslscan/archive/1.11.0-rbsec.tar.gz"
+  version "1.11.0"
+  sha256 "698dbf01b9af29aaddf6bce466f568762d852f7b71936861191a3b18d9dda6a5"
+  head "https://github.com/rbsec/sslscan.git"
+
+  depends_on "openssl"
+
+  def install
+    system "make"
+    bin.install "sslscan"
+  end
+
+  test do
+    system "#{bin}/sslscan", "google.com"
+  end
+end


DomT4:
Has this stopped breaking impressively if built against an OpenSSL that doesn't supply SSLv2 or relatively soon SSLv3? We previously axed it because of that: https://github.com/Homebrew/homebrew/pull/39880

Looks like this is perhaps a fork of the original (seemingly dead) project?

Library/Formula/sslscan.rb
```diff
@@ -0,0 +1,23 @@
+class Sslscan < Formula
+  desc "Test SSL/TLS enabled services to discover supported cipher suites."
+  homepage "https://github.com/rbsec/sslscan"
+  url "https://github.com/rbsec/sslscan/archive/1.11.0-rbsec.tar.gz"
+  version "1.11.0"
+  sha256 "698dbf01b9af29aaddf6bce466f568762d852f7b71936861191a3b18d9dda6a5"
+  head "https://github.com/rbsec/sslscan.git"
+
+  bottle :unneeded
```
DomT4: You would need a bottle here. Anything that has a build step needs bottling.

Library/Formula/sslscan.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  depends_on "openssl" => :build
+  depends_on "openssl"
+
+  def install
+    system "make"
```
DomT4: Presumably you shouldn't need both `make install` and `bin.install "sslscan"`. You'd also want to pass a prefix argument here if you intend to use `make`, since most things default to installing in `/usr` or `/usr/local`.

Library/Formula/sslscan.rb
```diff
@@ -0,0 +1,23 @@
...
+  homepage "https://github.com/rbsec/sslscan"
+  url "https://github.com/rbsec/sslscan/archive/1.11.0-rbsec.tar.gz"
+  version "1.11.0"
+  sha256 "698dbf01b9af29aaddf6bce466f568762d852f7b71936861191a3b18d9dda6a5"
+  head "https://github.com/rbsec/sslscan.git"
+
+  bottle :unneeded
+
+  depends_on "openssl" => :build
+  depends_on "openssl"
```
DomT4: A runtime dep is both runtime and buildtime, so don't need the `=> :build` dependency here.

dsXLII:
It builds, and runs just fine. If you use the Homebrew version of OpenSSL you lose support for SSLv2, but it just displays a friendly "hey there's no SSLv2 support" message and otherwise appears to work just fine.

(Its Makefile optionally downloads and compiles its own version of OpenSSL and statically links it, but at least for my purposes linking against Homebrew's version is just fine. Also, you have to go to their HEAD for that to work on MacOS; it's broken on their current release version.)

I've updated my fork with your suggestions.

mikemcquaid:

PR #46680, opened 2015-12-04
[WIP] open-mpi: fix Fortran support

UniqMartin:
**PLEASE DON'T MERGE!** Some parts shamelessly stolen from #46461.

### DIFF
--- a/Library/Formula/open-mpi.rb
+++ b/Library/Formula/open-mpi.rb
@@ -3,6 +3,7 @@ class OpenMpi < Formula
   homepage "https://www.open-mpi.org/"
   url "https://www.open-mpi.org/software/ompi/v1.10/downloads/openmpi-1.10.1.tar.bz2"
   sha256 "7919ecde15962bab2e26d01d5f5f4ead6696bbcacb504b8560f2e3a152bfe492"
+  revision 1
 
   bottle do
     sha256 "257a75a3227ddc0b5e7c970a97c76cadeda417a259c54efbe25df955e0efe32b" => :el_capitan
@@ -20,6 +21,7 @@ class OpenMpi < Formula
   deprecated_option "disable-fortran" => "without-fortran"
   deprecated_option "enable-mpi-thread-multiple" => "with-mpi-thread-multiple"
 
+  option "without-fortran", "Build without Fortran support"
   option "with-mpi-thread-multiple", "Enable MPI_THREAD_MULTIPLE"
   option :cxx11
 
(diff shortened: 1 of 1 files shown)


Library/Formula/open-mpi.rb
```diff
@@ -27,7 +29,7 @@ class OpenMpi < Formula
   conflicts_with "lcdf-typetools", :because => "both install same set of binaries."
 
   depends_on :java => :build
-  depends_on :fortran => :recommended
+  depends_on :fortran if build.with?("fortran")
```
xu-cheng: I would to fix the problem in the core since we repeatedly discourage others to do this.

mikemcquaid: Agreed :+1:

UniqMartin: I completely agree! As I said in the other PR, this is just a semi-ugly workaround and not the real solution. I myself would like to see all of those conditional `depends_on` eliminated (those within an `if`/`unless` block or with a trailing `if`/`unless` modifier). But fixing the issue here and eradicating conditional `depends_on` requires an in-depth understanding of Homebrew's dependency resolution logic – something I don't have (yet).

UniqMartin: This is really just a trick to move the evaluation to the formula, before the requirement has been transformed to a dependency with a different name, making the option handling in `brew install` fail to connect the `--with{,out}-fortran` (or its absence) with the `gcc` dependency.

This PR is really only here to check with the CI if the workaround has the desired effect.

xu-cheng:
Looked around a little. This bug is originally from [`Dependency.merge_repeat`](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/dependency.rb#L121-L130). In which, the crucial  `option_name` is abandoned. In order to solve that, I think we may need to introduce `option_name_list` variable as one dependency would have multiple option names after merging. I will look further for solution when time is free.

UniqMartin:
@xu-cheng If you can, please do. I agree that we need a proper solution to this issue eventually. For now, the dependency code is way to complicated for me to hack on (and to important to the project at large for me to break it).

zbeekman:
Any chance #46553 is also caused by the same bug in Dependency.merge_repeat?

UniqMartin:
@zbeekman Could be, but my gut feeling is there's something else going on there. I hope I'll be able to debug the mentioned issue soon (maybe later today or tomorrow) if no one else beats me to it.

